PR #184, opened 2026-02-18
feat: add self-update tracking fields to Group struct

alltheseas:
## Summary

- Adds `needs_self_update: bool` and `last_self_update_at: Option<Timestamp>` to the `Group` struct, enabling MDK to track post-join self-update obligations (MIP-02) and periodic key rotation staleness (MIP-00) at the storage layer
- `merge_pending_commit()` inspects the `StagedCommit` to detect pure self-update commits before merging, then conditionally clears the flag and records the timestamp — non-self-update commits leave both fields untouched
- Adds `groups_needing_self_update(threshold_secs)` query as a default trait method on `GroupStorage`, exposed via public API and UniFFI bindings

## Motivation

During MIPs compliance review for both pika and Vector, we identified that self-update tracking state (`pending_self_updates`, rotation timestamps) was being duplicated at the app layer. These are protocol-level requirements that belong in MDK so all consumers (pika, Vector, marmot-ts, future clients) get correct behavior for free.

## Changes

- **`mdk-storage-traits`**: Two new fields on `Group`, `groups_needing_self_update()` default trait method, test coverage
- **`mdk-core`**: `accept_welcome()` sets `needs_self_update = true`; `merge_pending_commit()` detects self-updates via `StagedCommit` API; `create_group()` initializes defaults; public API for query
- **`mdk-sqlite-storage`**: V004 migration adding columns, save/load support
- **`mdk-memory-storage`**: Test struct updates
- **`mdk-uniffi`**: Fields + query exposed to Swift/Kotlin

## Test plan

- [x] All 814 existing tests pass (struct literals updated across 9 files)
- [x] New `test_groups_needing_self_update` covers 5 cases: post-join flag, stale rotation, fresh rotation, inactive exclusion, creator-no-obligation
- [x] Runs against both SQLite and memory storage backends

This PR adds structured self-update tracking to groups and a query to find groups that need self-updates, implementing MIP-02 post-join obligations and MIP-00 periodic-rotation staleness checks. It replaces the previous two-field approach with a single SelfUpdateState enum, persists it in storage (SQLite migration), and exposes the query through core APIs and UniFFI bindings.

**What changed**
- Introduces a public SelfUpdateState enum (NotRequired, Required, CompletedAt(Timestamp)) and replaces needs_self_update + last_self_update_at with Group.self_update_state across crates.  
- Adds Group::groups_needing_self_update(threshold_secs) in mdk-core that delegates to storage to return active groups requiring self-update.  
- Implements a default groups_needing_self_update(threshold_secs) on the GroupStorage trait in mdk-storage-traits that filters active groups by SelfUpdateState::Required or CompletedAt older than threshold.  
- Updates merge_pending_commit() to detect pure self-update commits and set self_update_state = CompletedAt(now) on successful merges, leaving non-self-update commits unchanged.  
- Sets self_update_state = Required in accept_welcome() (post-join obligation) and initializes new groups with NotRequired in create_group().  
- Adds a SQLite migration V004 adding last_self_update_at INTEGER and maps it to SelfUpdateState in mdk-sqlite-storage row mapping and persistence.  
- Updates mdk-memory-storage implementations and tests to include the new self_update_state field.  
- Exposes self_update_state and groups_needing_self_update via UniFFI (mdk-uniffi) with stringified state values and a hex-encoded group ID result for the query.

**Protocol changes**
- Implements MIP-02 post-join self-update obligation by marking joined groups Required until a self-update commit is merged.  
- Supports MIP-00 periodic rotation staleness checks by recording CompletedAt timestamps and allowing threshold-based stale detection.

**API surface**
- Breaking change: public Group representation changed in mdk-storage-traits (and downstream crates) by replacing needs_self_update: bool and last_self_update_at: Option<Timestamp> with self_update_state: SelfUpdateState.  
- New public API: groups_needing_self_update(threshold_secs: u64) added to MDK/GroupStorage surfaces (default trait impl in mdk-storage-traits, core wrapper in mdk-core).  
- Storage schema change: SQLite migration V004 adds last_self_update_at INTEGER with semantics NULL = no obligation, 0 = required, >0 = unix timestamp of last rotation.  
- UniFFI/FFI change: UniFFI Group now includes a self_update_state string and a new groups_needing_self_update() binding returning Vec<String> group IDs.

**Testing**
- Tests added: test_groups_needing_self_update covers post-join flag, stale rotation, fresh rotation, inactive exclusion, and creator-no-obligation cases.  
- Existing tests updated across mdk-core, mdk-sqlite-storage, and mdk-memory-storage to initialize and roundtrip the new self_update_state field, including serde roundtrip and defaulting tests.  
- All storage backends (SQLite and memory) run the new cross-storage test and pass in CI per the PR description.


alltheseas (commented):
Thanks for the review. Addressed the inline comments in 717458a:

- **db.rs `unwrap_or`** → now uses `?` consistently (test helper schema updated too)
- **group_tests.rs inline import** → moved to module scope
- **`#[serde(default)]`** → added to both new fields for backward-compat deserialization

For the items I'm leaving as-is:

- **Group Debug redaction**: Pre-existing `derive(Debug)` on the struct, not introduced by this PR. `image_key`/`image_nonce` are already wrapped in `Secret<>` which handles redaction. Worth a follow-up but out of scope here.
- **SQL partial index**: `groups_needing_self_update()` is a default trait method that calls `all_groups()` and filters in Rust — there's no SQL query to index for. If/when someone adds a SQL-native override, the index would make sense then.

alltheseas (commented):
Thanks for the review. Addressed the inline comments in 717458a:

- **db.rs unwrap_or** — now uses `?` consistently (test helper schema updated too)
- **group_tests.rs inline import** — moved to module scope
- **serde(default)** — added to both new fields for backward-compat deserialization

For the items I'm leaving as-is:

- **Group Debug redaction**: Pre-existing `derive(Debug)` on the struct, not introduced by this PR. `image_key`/`image_nonce` are already wrapped in `Secret<>` which handles redaction. Worth a follow-up but out of scope here.
- **SQL partial index**: `groups_needing_self_update()` is a default trait method that calls `all_groups()` and filters in Rust — there's no SQL query to index for. If/when someone adds a SQL-native override, the index would make sense then.

PR #186, opened 2026-02-18
fix(welcomes): make client tag optional in welcome validation per MIP-02

### LINKED ISSUE #180: Welcome validation should not require client tag
## Summary

The Welcome event validation in `welcomes.rs` (`validate_welcome_event`) requires 4 tags including a `client` tag. However, the MIP-02 spec only defines 3 required tags for Welcome events:

- `e` (KeyPackage event reference)
- `relays` (relay URLs)
- `encoding` (content encoding format)

The `client` tag is an **optional** informational field, not a required one. Requiring it means that a spec-compliant implementation that omits the `client` tag would have its Welcome events rejected by MDK.

## Current Behavior

```rust
// welcomes.rs validate_welcome_event
if tags.len() < 4 {
    return Err(Error::InvalidWelcomeMessage);
}
// ...
if !has_client {
    return Err(Error::InvalidWelcomeMessage);
}
```

## Expected Behavior

- Minimum tag count should be 3 (relays + e + encoding)
- `client` tag presence should not be validated (it's optional/informational)
- Validation should only enforce the 3 required tags per MIP-02

## Impact

Any third-party implementation following the MIP-02 spec that omits the `client` tag would be unable to send Welcome events to MDK-based clients.

## References

- MIP-02 Welcome Event Structure: https://github.com/marmot-protocol/marmot/blob/master/02.md#welcome-event-structure

erskingardner:
## Summary

- Make the `client` tag optional in welcome event validation, per MIP-02 spec
- Previously, any spec-compliant third-party implementation that omitted the `client` tag would have its Welcome events rejected by MDK

## Changes

- **`validate_welcome_event`**: Reduced minimum tag count from 4 to 3, removed `client` tag presence check
- **Tests**: Updated existing tests to not depend on `client` tag, added `test_welcome_validation_accepts_missing_client_tag` to verify welcomes without client tag are accepted
- **CHANGELOG**: Added entry under `## Unreleased > ### Fixed`

## Notes

MDK still *includes* the `client` tag when creating welcome events — the change only affects validation of incoming events.

Closes #180

This PR relaxes Welcome event validation to comply with MIP-02 by making the informational `client` tag optional and lowering the minimum required tag count, while enforcing that welcome content encoding must be "base64". The change improves interoperability with third-party implementations that omit the `client` tag without changing MDK's own welcome creation behavior.

**What changed**:
- validate_welcome_event now requires at least 3 tags (relays, e, encoding) instead of 4 and no longer requires the `client` tag (crates/mdk-core).
- The `client` tag remains supported but is optional; if present its value must be non-empty (crates/mdk-core).
- Encoding tag validation was tightened to accept only "base64" and rejects other values (crates/mdk-core).
- Tests were updated to stop depending on the `client` tag and a new test (test_welcome_validation_accepts_missing_client_tag) was added (crates/mdk-core tests).
- CHANGELOG updated to document that welcome validation no longer requires the `client` tag (crates/mdk-core).

**Security impact**:
- Encoding handling now requires explicit base64, reducing ambiguity in content decoding and preventing encoding-downgrade or ambiguity attacks.
- No changes to cryptographic primitives, key handling, SQLCipher configuration, or file permissions.

**Protocol changes**:
- Welcome event validation now conforms to MIP-02 by treating the `client` tag as optional and enforcing presence of `relays`, `e` (KeyPackage event reference), and `encoding` tags.
- Tag order remains unenforced to maximize interoperability.

**API surface**:
- No breaking changes to public APIs, public types, traits, storage schemas, or UniFFI/FFI boundaries.

**Testing**:
- Existing tests were modified and a new test added to verify acceptance of spec-compliant welcomes without the `client` tag and to reflect the stricter encoding validation.

### DIFF
--- a/crates/mdk-core/src/welcomes.rs
+++ b/crates/mdk-core/src/welcomes.rs
@@ -137,6 +137,13 @@ where
                         has_event_ref = true;
                     }
                 }
+                TagKind::Client => {
+                    // Client tag is optional, but if present its value must be non-empty
+                    match tag.content() {
+                        Some(value) if !value.is_empty() => {}
+                        _ => return Err(Error::InvalidWelcomeMessage),
+                    }
+                }
                 TagKind::Custom(name) if name.as_ref() == "encoding" => {
                     // Validate encoding value is "base64"
                     if let Some(encoding_value) = tag.content() {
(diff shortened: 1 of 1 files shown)


mubarakcoded (approved):
This looks good to me. Good defensive choice preventing implementations from sending ["client", ""], which is technically present but meaningless.

PR #196, opened 2026-02-23
Expose clear_pending_commit in UniFFI bindings

dannym-arx:
This PR exposes the `clear_pending_commit()` method through the UniFFI bindings, allowing foreign language clients to roll back uncommitted pending MLS commits. This is a recovery mechanism for failed relay publishes that would otherwise permanently block group operations.

**What changed**:
- Added `clear_pending_commit(mls_group_id: String)` public method to the `Mdk` struct in the mdk-uniffi crate, which accepts a hex-encoded MLS group ID and rolls back to the pre-commit state
- Updated mdk-uniffi CHANGELOG.md to document the new method

**API surface**:
- New public method added to mdk-uniffi's `Mdk` type that mirrors the existing `merge_pending_commit()` flow but performs a rollback instead of merge
- The method parses the hex-encoded group ID and delegates to the underlying mdk-core implementation with error handling

### DIFF
--- a/crates/mdk-uniffi/CHANGELOG.md
+++ b/crates/mdk-uniffi/CHANGELOG.md
@@ -31,6 +31,8 @@
 
 ### Added
 
+- **`clear_pending_commit` method**: Added `clear_pending_commit(group_id)` to allow callers to roll back an uncommitted pending MLS commit. ([#196](https://github.com/marmot-protocol/mdk/pull/196))
+
 ### Fixed
 
 ### Removed


erskingardner (approved):
All good, unless you want to add a test, but given it's the thinnest wrapper possible, I'm fine with this.

dannym-arx:
I don't think we need a test personally, the useful tests are in the PR introducing clear_pending_commit. Any test added here would just be silly imo

Issue #220, opened 2026-03-06
Optimize library size if possible


SubatomicPlanets:
I use the UniFFI Kotlin bindings, and when I analyze my APK, I can see that libmdk_uniffi.so file uses 6.3MB (plus another 5.2MB for armv7. This is the largest library in my app). Other libraries like rust-nostr use configuration code like this to optimize space:
```
[profile.release]
opt-level = 'z'     # Optimize for size
lto = true          # Enable Link Time Optimization
codegen-units = 1   # Reduce the number of codegen units to increase optimizations
panic = "abort"     # Abort on panic
strip = true        # Full strip the binary. A debug binary will be used for generating UniFFI foreign languages!
```
(copied directly from [rust-nostr](https://github.com/rust-nostr/nostr-sdk-ffi/blob/master/.cargo/config.toml))

I don't know how much something like this would help, but the smaller the better!

PR #221, opened 2026-03-08
perf: optimize release binary size for Android/iOS bindings

dannym-arx:
marmot

Add a `[profile.release]` section to the workspace `Cargo.toml` with size-optimizing flags — `opt-level = "z"`, fat LTO, single codegen unit, `panic = "abort"`, and symbol stripping — bringing the marmot colony's `libmdk_uniffi.so` on a serious diet. Expected reduction of ~40-60% in `.so` size for Android/iOS bindings.

Closes https://github.com/marmot-protocol/mdk/issues/220

This PR adds a workspace-level size-optimized Cargo release profile and updates UniFFI binding build steps to use it, reducing the libmdk_uniffi native library size (reported ~40% reduction) to shrink Android/iOS APK footprints without changing Rust source or public APIs. The change is build-time only and focused on UniFFI artifacts and packaging.

What changed:
- Added [profile.release-size] to the workspace Cargo.toml (inherits = "release", opt-level = "z", lto = true, codegen-units = 1, panic = "abort", strip = true) to optimize for binary size.
- Updated justfile build targets and binding generation steps to use cargo build --profile release-size and to reference target/release-size artifact paths for Android and iOS.
- Documented the new release-size profile and measured size savings in crates/mdk-uniffi/CHANGELOG.md.
- Affected crate: mdk-uniffi only; no source code or public APIs modified.

Security impact:
- No security-sensitive changes detected; no cryptographic, key handling, identity, SQLCipher, or file-permission changes.

Protocol changes:
- None.

API surface:
- No breaking changes to public APIs, no new public APIs, and no UniFFI/FFI boundary or storage schema changes.

Testing:
- No tests were added or modified.

### DIFF
--- a/justfile
+++ b/justfile
@@ -310,9 +310,9 @@ gen-binding lang: (_build-uniffi "false" "false")
     @echo "Generating {{lang}} bindings..."
     cd crates/mdk-uniffi && cargo run --bin uniffi-bindgen generate \
         -l {{lang}} \
-        --library ../../target/release/{{lib_filename}} \
+        --library ../../target/release-size/{{lib_filename}} \
         --out-dir bindings/{{lang}}
-    cp target/release/{{lib_filename}} crates/mdk-uniffi/bindings/{{lang}}/{{lib_filename}}
+    cp target/release-size/{{lib_filename}} crates/mdk-uniffi/bindings/{{lang}}/{{lib_filename}}
     @echo "✓ Bindings generated in crates/mdk-uniffi/bindings/{{lang}}/"
 
 gen-binding-kotlin: (_build-uniffi "true") (gen-binding "kotlin")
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -69,3 +69,10 @@ tracing-subscriber = "0.3"
 
 # Testing
 tempfile = "3.20"
+
+[profile.release]
+opt-level = "z"      # Optimize for size
+lto = true           # Link-time optimization across all crate boundaries
+codegen-units = 1    # Single codegen unit (required to maximize LTO effectiveness)
+panic = "abort"      # No unwinding machinery (safe at UniFFI FFI boundary)
```
erskingardner: I like the goal here, and I agree with `opt-level = "z"`, `lto = true`, and `codegen-units = 1` all make sense to me.

My concern is with the two settings that change operational behavior more than just size:

- `panic = "abort"` means any unexpected Rust panic in release will hard-abort the process instead of unwinding. In our Flutter / whitenoise-rs world that likely turns a Rust bug into a native app crash, so I don’t think we should call this universally “safe at the UniFFI FFI boundary” without being explicit that this is the behavior we want.
- `strip = true` helps size, but it also makes production crash diagnosis and symbolication harder unless we’re preserving symbols separately in CI.

Also, this is a workspace-global `[profile.release]`, but the rationale in the PR is specifically about Android/iOS bindings. So this changes all release builds in the workspace, not just mdk-uniffi.

My preference would be one of:
1. keep `opt-level = "z"`, `lto = true`, `codegen-units = 1` and drop `panic = "abort"` / `strip = true`, or
2. scope this optimization to `mdk-uniffi` packaging rather than the whole workspace, if that’s possible.

jgmontoya (approved):

PR #225, opened 2026-03-10
fix(mdk-core): atomically strip admin status when removing members

mubarakcoded:
## Summary
- `remove_members()` now uses `CommitBuilder` to bundle removal proposals with a `GroupContextExtensions` update in a single MLS commit
- When any removed members are in the admin list, the admin list is atomically updated within the same epoch — no separate follow-up commit needed
- When no removed members are admins, behavior is unchanged (no extension proposal added)

Closes marmot-protocol/whitenoise-rs#514

## Test plan
- [x] `test_remove_members_strips_admin_status` — removes an admin member, verifies admin list is updated atomically
- [x] `test_remove_non_admin_member_preserves_admin_list` — removes a non-admin member, verifies admin list is unchanged
- [x] `just fmt`, `just lint`, `just test` all pass

This PR makes remove_members() atomic: when removing members who are also group admins, the admin list is updated in the same MLS commit so a removed-and-readded member cannot retain prior admin privileges. It centralizes processed-message tracking and tightens error handling around removals (including rejecting self-removal).

**What changed**:
- remove_members() (crates/mdk-core) now uses CommitBuilder to include removal proposals and, when needed, a GroupContextExtensions proposal in a single MLS commit so admin-list updates are atomic.  
- A private helper track_processed_message was added to centralize creation and storage of ProcessedMessage records and replace repeated inline persistence.  
- remove_members now rejects self-removal and explicitly validates admin-extension updates (including an empty-admin-set check) and surfaces related errors instead of silently proceeding.  
- Commit-building in remove_members handles PSK loading, optional GroupContextExtensions construction, and consistent post-commit processed-message tracking.  
- Tests in crates/mdk-core were added/updated to cover atomic admin stripping, non-admin removals, re-add regression, multi-admin removals, self-removal rejection, and processed-message-state tracking.

**Protocol changes**:
- Uses MLS GroupContextExtensions to atomically update group metadata (admin list) alongside membership removals (aligns with MLS extension proposal semantics per RFC 9420).

**API surface**:
- No public API signature or storage-schema changes; internal behavior of MDK::remove_members changed and UniFFI/FFI boundaries are unaffected.

**Testing**:
- New/extended tests in crates/mdk-core include test_remove_members_strips_admin_status, test_remove_non_admin_member_preserves_admin_list, and related admin-permission and processed-message-tracking tests; formatting, linting, and test suites pass.

### DIFF
--- a/crates/mdk-core/src/groups.rs
+++ b/crates/mdk-core/src/groups.rs
@@ -2761,6 +2761,51 @@ mod tests {
     }
 
     #[test]
+    fn test_leave_group_records_processed_state() {
+        let creator_mdk = create_test_mdk();
+        let (creator, initial_members, admins) = create_test_group_members();
+        let creator_pk = creator.public_key();
+
+        let mut key_package_events = Vec::new();
+        for m in &initial_members {
+            key_package_events.push(create_key_package_event(&creator_mdk, m));
+        }
+
+        let create_result = creator_mdk
+            .create_group(
+                &creator_pk,
+                key_package_events,
+                create_nostr_group_config_data(admins),
+            )
+            .expect("Failed to create group");
+
+        let group_id = &create_result.group.mls_group_id.clone();
+        creator_mdk
+            .merge_pending_commit(group_id)
+            .expect("Failed to merge pending commit");
+
+        let leave_result = creator_mdk
+            .leave_group(group_id)
+            .expect("Failed to leave group");
+
+        // Verify the processed message was recorded with Processed state
+        // (not ProcessedCommit), since leave_group creates a proposal.
+        let processed = creator_mdk
+            .storage()
+            .find_processed_message_by_event_id(&leave_result.evolution_event.id)
+            .expect("Failed to query processed message")
+            .expect("ProcessedMessage should exist");
+
+        assert_eq!(
+            processed.state,
+            message_types::ProcessedMessageState::Processed,
+            "leave_group should record Processed state, not ProcessedCommit"
+        );
+        assert_eq!(processed.wrapper_event_id, leave_result.evolution_event.id);
+        assert!(processed.failure_reason.is_none());
+    }
+
+    #[test]
     fn test_self_update_success() {
         let creator_mdk = create_test_mdk();
         let (creator, initial_members, admins) = create_test_group_members();


crates/mdk-core/src/groups.rs
```diff
@@ -2467,6 +2515,135 @@ mod tests {
...
+            .expect("Group should exist");
+
+        let expected_admins: BTreeSet<PublicKey> = std::iter::once(creator_pk).collect();
+        assert_eq!(
+            group_after.admin_pubkeys, expected_admins,
+            "Admin set should contain only the creator after removing admin member"
+        );
+    }
+
+    #[test]
```
erskingardner: Both new tests remove a single member. The admin-stripping logic iterates over `pubkeys` and calls `remove_admin` for each, so it should handle bulk removals correctly — but there's no test that exercises removing two admins in the same `remove_members` call. A quick third test case would close this gap.

crates/mdk-core/src/groups.rs
```diff
@@ -850,10 +855,53 @@ where
...
+            None
+        };
+
+        // Build a single commit containing removal proposals and, if needed,
+        // a GroupContextExtensions proposal to update the admin list.
+        let builder = mls_group
+            .commit_builder()
+            .propose_removals(leaf_indices.iter().cloned());
+
+        let builder = match updated_extensions {
```
erskingardner:
```rust
let builder = mls_group
    .commit_builder()
    .propose_removals(leaf_indices.iter().cloned());

let builder = match updated_extensions {
    Some(ext) => builder
        .propose_group_context_extensions(ext)
        .map_err(|e| Error::Group(e.to_string()))?,
    None => builder,
};
```

Rebinding `builder` with `let builder = match ...` is idiomatic Rust and works, but an alternative that avoids the `None => builder` arm (and the implicit type-unification requirement) is to call `propose_group_context_extensions` conditionally before chaining:

```rust
let commit_bundle = {
    let mut b = mls_group
        .commit_builder()
        .propose_removals(leaf_indices.iter().cloned());
    if let Some(ext) = updated_extensions {
        b = b.propose_group_context_extensions(ext)
            .map_err(|e| Error::Group(e.to_string()))?;
    }
    b
};
```

Not a blocking issue — the current code compiles and reads clearly — but the explicit `if let` form avoids `match` with a passthrough `None` arm.

PR #253, opened 2026-04-09
feat: add disappearing message duration to MLS group extension

### LINKED ISSUE #278: Add support for disappearing messages settings for group
We need to add a field to the nostr group data extension for disappearing chat settings. 

- Only admins should be able to change this
- We need getters and setters for the public API
- We also need to update the Rust bridge with method for the @parres-hq/white-noise-flutter app

dannym-arx:
marmot

## Known issues that will not be fixed this PR

- Media files are not deleted on the blossom server (there's no bud for expiration afaict), a solution would be sending a delete request directly on expiration, but that's for another pr
- Media files are kept in media_cache, again something for another pr 

## Summary

Adds `disappearing_message_duration_secs: Option<u64>` across the full MDK stack, giving clients the plumbing to set per-group message expiration timers.

- `None` means messages persist forever (disabled)
- `Some(n)` means messages expire `n` seconds after creation
- `Some(0)` is rejected at both `create_group` and `update_group_data`

## What changed

**mdk-core (MLS extension v3):** New field on `NostrGroupDataExtension` with TLS wire serialization. Old v2 groups deserialize with `None` (backward compatible). `NostrGroupDataUpdate` builder gets a `.disappearing_message_duration_secs()` setter. Validation rejects zero-duration at group creation and update.

**mdk-storage-traits:** New `Option<u64>` field on the `Group` struct.

**mdk-uniffi:** New field on both the `Group` record and `GroupDataUpdate` record, with conversion wiring.

**mdk-sqlite-storage:** V006 migration adds the column. `save_group` persists it, `row_to_group` reads it.

Part of https://github.com/marmot-protocol/whitenoise-rs/issues/278

⚠️ Security-sensitive changes: SQLite PRAGMA secure_delete = ON was enabled so deleted rows (including expired/removed messages) are overwritten on disk.

This PR adds per-group disappearing-message configuration across the MDK stack, allowing groups to set a message expiration in seconds (None = persist forever; Some(n) = expire after n seconds; Some(0) is rejected at creation/update). It bumps the Nostr MLS group-data extension to v3 to TLS-serialize the new field while remaining backward-compatible with v1/v2 groups, and it introduces local deletion APIs and storage support to remove expired or explicit messages (media cleanup on blossom is a known unaddressed issue).

What changed:
- mdk-core: Introduced `disappearing_message_duration_secs: Option<u64>` on `NostrGroupDataExtension` (extension version 2→3), extended `NostrGroupConfigData::new(...)` and `NostrGroupDataUpdate` (tri-state update support via a setter), validate/reject zero durations on create/update, auto-insert or replace NIP-40 expiration tags when a group has a duration, propagate the field through welcome/creation/update/sync flows, and added MDK local deletion APIs `delete_message`, `delete_messages_before_timestamp`, and `delete_processed_messages_for_group`.
- mdk-storage-traits: Added `disappearing_message_duration_secs: Option<u64>` to the `Group` struct and added required `MessageStorage` methods `delete_message`, `delete_messages_before_timestamp`, and `delete_processed_messages_for_group`.
- mdk-uniffi: Exposed the disappearing-message tri-state (`DisappearingMessageDuration`) and added `disappearing_message_duration_secs` to the `Group` record and `GroupDataUpdate` at the FFI boundary, and exported the deletion APIs over UniFFI.
- mdk-sqlite-storage: Added migration V006 to add `groups.disappearing_message_duration_secs INTEGER` (NULL = disabled), updated `save_group`/`row_to_group` to persist/read the field, implemented message deletion operations, and enabled `PRAGMA secure_delete = ON` during DB initialization.
- mdk-memory-storage: Implemented the new deletion APIs in-memory and updated test fixtures and helpers to include the new `Group` field.

Security impact:
- Enabled SQLite `PRAGMA secure_delete = ON` to reduce forensic recoverability of deleted rows by overwriting deleted content on disk.
- No changes to cryptographic primitives, key derivation, nonce usage, identity binding, or key storage semantics.
- No new error-message exposures of sensitive identifiers were introduced.

Protocol changes:
- MLS/Nostr: `NostrGroupDataExtension` wire format bumped to version 3 to include `disappearing_message_duration_secs` while v1/v2 groups deserialize the field as None (backward-compatible).
- Nostr NIP-40: MDK now auto-adds or replaces NIP-40 expiration tags on outer wrapper (kind:445) events when a group's disappearing-message duration is set.

API surface:
- Breaking changes: Rust code constructing `Group` literals must include the new `disappearing_message_duration_secs: Option<u64>` field.
- New/changed APIs: `NostrGroupConfigData::new(...)` gained a `disappearing_message_duration_secs` parameter, `NostrGroupDataUpdate` gained a `.disappearing_message_duration_secs()` setter (tri-state at FFI), the `MessageStorage` trait gained three deletion methods, MDK added corresponding public deletion methods, and UniFFI records and FFI methods were extended to surface the disappearing-message tri-state and deletion APIs.
- Storage schema: SQLite migration V006 adds a nullable `disappearing_message_duration_secs` column to `groups`.

Testing:
- Tests were added and/or updated across crates to cover constructor/update validation (rejecting zero), create/update behavior, auto-expiration tag application, and the new deletion APIs, and many test fixtures were updated to initialize `disappearing_message_duration_secs` to `None`.


dannym-arx:
@coderabbitai full review

crates/mdk-sqlite-storage/migrations/V006__add_disappearing_message_duration.sql
```diff
@@ -0,0 +1,4 @@
+-- Add disappearing message duration to groups table.
+-- NULL means disabled (messages persist forever).
+-- Non-null positive integer means messages expire after that many seconds.
+ALTER TABLE groups ADD COLUMN disappearing_message_duration_secs INTEGER;
```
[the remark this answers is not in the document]
dannym-arx: I mean... play stupid games win stupid prizes... adding a check there means recreating the groups table, which in a way is fine, but it's not worth it. We can just model out this behavior in the rust side

dannym-arx:
@coderabbit full review

crates/mdk-core/src/messages/create.rs
```diff
@@ -132,6 +137,22 @@ where
...
+                let expiration =
+                    Timestamp::from(rumor.created_at.as_secs().saturating_add(duration_secs));
+                let expiration_tag = EventTag::expiration(expiration);
+                let mut all_tags = tags.unwrap_or_default();
+                all_tags.retain(|t| !t.is_expiration());
+                all_tags.push(expiration_tag);
+                Some(all_tags)
+            }
+            None => tags,
+        };
```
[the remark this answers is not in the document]
dannym-arx: I think I'm gonna keep this. because I think there should be a difference between per event expiry and per group expiry

crates/mdk-core/src/extension/types.rs
```diff
@@ -156,13 +213,29 @@ impl NostrGroupDataExtension {
...
+            Err(_v3_err) => {
+                let (v1v2, remainder) =
+                    TlsNostrGroupDataExtensionV1V2::tls_deserialize_bytes(bytes)?;
+                if !remainder.is_empty() {
+                    return Err(Error::ExtensionFormatError(
+                        "Trailing bytes in NostrGroupDataExtension".to_string(),
+                    ));
+                }
+                Self::from_raw(v1v2.into_v3())
+            }
```
jgmontoya: This logic seems fragile, I think we should just use the version bytes (first two bytes), otherwise a v3 payload that is corrupt could pass as a v1/v2 payload and obscure the problem

crates/mdk-sqlite-storage/src/groups.rs
```diff
@@ -249,7 +250,8 @@ impl GroupStorage for MdkSqliteStorage {
                     &last_message_processed_at,
                     &(group.epoch as i64),
                     group.state.as_str(),
-                    &last_self_update_at
+                    &last_self_update_at,
+                    &group.disappearing_message_duration_secs.and_then(|d| i64::try_from(d).ok())
```
jgmontoya: Should we perhaps add a comment saying that values larger than `i64::MAX` will be considered as `None` and won't disappear after the whatever many hundreds of billions of years that is? 😅

crates/mdk-memory-storage/src/messages.rs
```diff
@@ -295,6 +295,107 @@ impl MessageStorage for MdkMemoryStorage {
...
+            };
+
+        // Also scan messages_cache for orphaned entries (LRU divergence)
+        let orphaned: Vec<EventId> = inner
+            .messages_cache
+            .iter()
+            .filter(|(_, msg)| &msg.mls_group_id == group_id && msg.created_at < before)
+            .map(|(eid, _)| *eid)
+            .collect();
+
```
jgmontoya, line 351: Having to do this is very annoying. I think we should add an issue that addresses the divergence of both caches so that they stay in-sync at source and we can rely upon them without having to double-check in multiple places.

As an idea: we could perhaps make `messages_cache` a secondary index, mapping to `GroupId` rather than a duplicated `Message`, and make the per-group one the "primary" index. A stale pointer would still be a miss. But maybe we can drop the `messages_cache` altogether

crates/mdk-sqlite-storage/src/groups.rs
dannym-arx: I mean... should we? lol I don't think it matters tbh

crates/mdk-memory-storage/src/messages.rs
dannym-arx: I raised a separate issue for this, and will address it in a follow up pr: https://github.com/marmot-protocol/mdk/issues/255

dannym-arx:
closing this, splitting it into 3 self contained PRs as we discussed internally

PR #29, opened 2025-11-29
Implement hex-to-base64 encoding migration for KeyPackage and Welcome

mubarakcoded:
This implements a 3-phase migration strategy to reduce payload sizes by ~33%:

Phase 1 (Current):
- Add dual-format reading support (both hex and base64)
- Keep writing hex by default (use_base64_encoding: false)
- Backward and forward compatible

Phase 2 (After 3 months):
- Switch to base64 encoding via config flag
- Still read both formats during transition

Phase 3 (After 6 months - Optional):
- Remove hex support entirely

Changes:
- Add MdkConfig struct with use_base64_encoding flag (default: false)
- Implement smart format detection based on character set
  - Hex: only [0-9a-fA-F]
  - Base64: [A-Za-z0-9+/=]
- Add conditional encoding in key package and welcome creation
- Update builder pattern to support configuration

This ensures a smooth migration that handles the 3-month key package lifetime without breaking compatibility.

## Summary by CodeRabbit

* **New Features**
  * Builder-style MDK configuration with a toggle for base64 vs hex encoding; runtime-selectable encoding for outgoing payloads and encoding-aware decoding for incoming welcomes/key packages (encoding indicated via tags).

* **Bug Fixes**
  * Clearer, encoding-aware decoding errors and persisted failure state when decoding fails.

* **Public API**
  * Updated public returns and builder/config exposure to support encoding-tagged payloads and variable-length tags.

* **Tests**
  * Expanded tests for cross-format decoding, invalid inputs, padding/edge cases, and size differences.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #256, opened 2026-04-14
Fix receiver-side admin depletion validation

erskingardner:
## Summary
- Add pre-merge validation that receiver-processed commits leave at least one active admin.
- Validate the staged post-commit group-data extension against effective membership, including removals, self-removals, and additions.
- Add a defensive storage-sync guard to prevent persisting adminless MLS group metadata.
- Add regressions for malicious admin-authored extension updates and adminless metadata sync.

## Testing
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p mdk-core`
- `cargo clippy -p mdk-core --all-targets -- -D warnings`
- `cargo test -p mdk-core --no-default-features test_admin_extension_update_cannot_deplete_all_admins -- --nocapture`

Fixes https://github.com/marmot-protocol/marmot-security/issues/29

This PR prevents receiver-side acceptance or storage of MLS group state that would leave a group with no active admins by validating the post-commit effective admin set (including extension-only updates) and adding a defensive storage-sync guard. It adds a small state_validation module with utilities that compute live and post-commit member identities and enforces an "at least one active admin" invariant during commit processing and metadata sync, plus regression tests covering both extension-based and removal-based attack vectors.

What changed:
- Added a new internal module crates/mdk-core/src/state_validation.rs providing post-commit and post-departure member-identity derivation and admin-invariant checks (member_pubkeys, post_commit_member_pubkeys, validate_admin_depletion, validate_active_admins(_in_group), validate_admin_invariant_after_commit).
- Update validate_commit_authorization to call validate_admin_invariant_after_commit for admin senders and for SelfRemove cases, moving from removal-count heuristics to post-commit admin-set validation.
- Added a defensive guard in sync_group_metadata_from_mls to refuse persisting group metadata when the effective admin set is empty.
- Introduced regression tests: test_admin_extension_update_cannot_deplete_all_admins and test_sync_group_metadata_rejects_adminless_group_data that assert malicious admin-authored extension updates and adminless metadata sync are rejected.
- Minor visibility change: pubkey_from_credential widened to pub(crate) to permit use by the new module.
- Updated CHANGELOG with an Unreleased → Fixed entry documenting rejection of adminless group states.

Security impact:
- ⚠️ Security-sensitive changes: strengthens identity-binding/credential validation by enforcing that at least one admin identity is still represented by live members after commits and before persisting metadata.
- No cryptographic algorithm, key derivation, nonce, or HKDF context changes were introduced.
- No changes to SQLCipher, file permissions, or storage encryption configuration are present.
- Error messages are generic for the admin-depletion case ("Would leave group with no admins") and do not leak member identities.

Protocol changes:
- MLS handling: commit/merge-time validation now checks the staged post-commit admin set (including admins from group-data extensions and add/remove proposals) to ensure the group is not orphaned, tightening receiver-side validation to complement creating-side guards.
- Nostr integration: admin identity determination remains based on Nostr PublicKey extracted from member credentials; the PR ensures admin_pubkeys extension and member identities are reconciled post-commit.

API surface:
- No breaking public API changes to exported crates; changes are internal to mdk-core (new pub(crate) methods and a module).
- Storage schema, UniFFI/FFI boundaries, and public types remain unchanged.

Testing:
- Added focused regression tests covering admin-extension-only depletion and adminless metadata sync rejection, and included test commands used by the author (cargo fmt/check, cargo test -p mdk-core, cargo clippy -p mdk-core, targeted cargo test invocation).

### DIFF
--- a/crates/mdk-core/src/groups.rs
+++ b/crates/mdk-core/src/groups.rs
@@ -199,7 +199,10 @@ where
     Storage: MdkStorageProvider,
 {
     /// Extracts a public key from an MLS credential.
-    fn pubkey_from_credential(&self, credential: &Credential) -> Result<PublicKey, Error> {
+    pub(crate) fn pubkey_from_credential(
+        &self,
+        credential: &Credential,
+    ) -> Result<PublicKey, Error> {
         let basic = BasicCredential::try_from(credential.clone())?;
         self.parse_credential_identity(basic.identity())
     }
@@ -501,16 +504,7 @@ where
     /// * `Err(Error)` - If the group is not found or there is an error accessing member data
     pub fn get_members(&self, group_id: &GroupId) -> Result<BTreeSet<PublicKey>, Error> {
         let group = self.load_mls_group(group_id)?.ok_or(Error::GroupNotFound)?;
-
-        // Store members in a variable to extend its lifetime
-        let mut members = group.members();
-        members.try_fold(BTreeSet::new(), |mut acc, m| {
-            let credentials: BasicCredential = BasicCredential::try_from(m.credential)?;
-            let identity_bytes: &[u8] = credentials.identity();
-            let public_key = self.parse_credential_identity(identity_bytes)?;
-            acc.insert(public_key);
-            Ok(acc)
-        })
+        self.member_pubkeys(&group)
     }
 
     /// Returns the local member's current MLS leaf index for a group.
--- a/crates/mdk-core/src/groups.rs
+++ b/crates/mdk-core/src/groups.rs
@@ -1777,8 +1777,9 @@ where
         let mls_group = self.load_mls_group(group_id)?.ok_or(Error::GroupNotFound)?;
         let mut stored_group = self.get_group(group_id)?.ok_or(Error::GroupNotFound)?;
 
-        // Validate the mandatory group-data extension FIRST before making any state changes
-        // This ensures we don't update stored_group if the extension is missing, invalid, or unsupported
+        // Validate the mandatory group-data extension FIRST before making any state changes.
+        // Receiver-side commits are checked before merge; this guard covers local merge
+        // and storage-sync paths so invalid MLS metadata is not copied into storage.
         let group_data = NostrGroupDataExtension::from_group(&mls_group)?;
         self.validate_active_admins_in_group(&mls_group, &group_data.admins)?;
         // Only after successful validation, update epoch and metadata from MLS group
@@ -3590,6 +3591,8 @@ mod tests {
             .merge_pending_commit(&creator_mdk.provider)
             .expect("OpenMLS should merge the adminless extension update");
 
+        // This guard does not roll back the mutated MLS group. It prevents the
+        // invalid extension from being copied into stored group metadata.
         let result = creator_mdk.sync_group_metadata_from_mls(&group_id);
 
         assert!(
--- a/crates/mdk-core/src/groups.rs
+++ b/crates/mdk-core/src/groups.rs
@@ -1786,6 +1786,7 @@ where
         // Validate the mandatory group-data extension FIRST before making any state changes
         // This ensures we don't update stored_group if the extension is missing, invalid, or unsupported
         let group_data = NostrGroupDataExtension::from_group(&mls_group)?;
+        self.validate_active_admins_in_group(&mls_group, &group_data.admins)?;
         // Only after successful validation, update epoch and metadata from MLS group
         stored_group.epoch = mls_group.epoch().as_u64();
 
@@ -3556,6 +3557,61 @@ mod tests {
             initial_stored_group.last_message_at
         );
         assert_eq!(synced_stored_group.state, initial_stored_group.state);
+    }
+
+    #[test]
+    fn test_sync_group_metadata_rejects_adminless_group_data() {
+        let creator_mdk = create_test_mdk();
+        let (creator, initial_members, admins) = create_test_group_members();
+        let group_id = create_test_group(&creator_mdk, &creator, &initial_members, &admins);
+
+        let initial_stored_group = creator_mdk
+            .get_group(&group_id)
+            .expect("Failed to get initial stored group")
+            .expect("Stored group should exist");
+
+        let mut mls_group = creator_mdk
+            .load_mls_group(&group_id)
+            .expect("Failed to load MLS group")
+            .expect("MLS group should exist");
+        let mut group_data =
+            NostrGroupDataExtension::from_group(&mls_group).expect("Group data should parse");
+        group_data.admins.clear();
+
+        let extension =
+            super::MDK::<MdkMemoryStorage>::get_unknown_extension_from_group_data(&group_data)
+                .expect("Failed to build adminless group-data extension");
+        let mut extensions = mls_group.extensions().clone();
+        extensions
+            .add_or_replace(extension)
+            .expect("Failed to replace group-data extension");
+
+        let signature_keypair = creator_mdk
+            .load_mls_signer(&mls_group)
+            .expect("Failed to load signer");
+        let (_message_out, _, _) = mls_group
+            .update_group_context_extensions(&creator_mdk.provider, extensions, &signature_keypair)
+            .expect("OpenMLS should allow the adminless extension update");
+        mls_group
+            .merge_pending_commit(&creator_mdk.provider)
+            .expect("OpenMLS should merge the adminless extension update");
+
+        let result = creator_mdk.sync_group_metadata_from_mls(&group_id);
+
+        assert!(
+            matches!(result, Err(crate::Error::Group(ref msg)) if msg.contains("no admins")),
+            "sync should reject adminless group data, got: {:?}",
+            result
+        );
+
+        let stored_group = creator_mdk
+            .get_group(&group_id)
+            .expect("Failed to reload stored group")
+            .expect("Stored group should still exist");
+        assert_eq!(
+            stored_group.admin_pubkeys, initial_stored_group.admin_pubkeys,
+            "Rejected sync should not alter stored admins"
+        );
     }
 
     #[test]
--- a/crates/mdk-core/src/groups.rs
+++ b/crates/mdk-core/src/groups.rs
@@ -492,7 +492,7 @@ where
             .map_err(|e| Error::Group(e.to_string()))
     }
 
-    /// Gets the public keys of all members in an MLS group
+    /// Gets the Nostr identities represented by all live members in an MLS group
     ///
     /// # Arguments
     ///
@@ -500,11 +500,11 @@ where
     ///
     /// # Returns
     ///
-    /// * `Ok(BTreeSet<PublicKey>)` - Set of member public keys
+    /// * `Ok(BTreeSet<PublicKey>)` - Set of represented member identities
     /// * `Err(Error)` - If the group is not found or there is an error accessing member data
     pub fn get_members(&self, group_id: &GroupId) -> Result<BTreeSet<PublicKey>, Error> {
         let group = self.load_mls_group(group_id)?.ok_or(Error::GroupNotFound)?;
-        self.member_pubkeys(&group)
+        self.live_member_identities(&group)
     }
 
     /// Returns the local member's current MLS leaf index for a group.
--- a/crates/mdk-core/src/messages/commit.rs
+++ b/crates/mdk-core/src/messages/commit.rs
@@ -190,7 +190,10 @@ mod tests {
     use mdk_storage_traits::groups::GroupStorage;
     use mdk_storage_traits::groups::types as group_types;
     use nostr::{EventBuilder, EventId, Keys, Kind};
+    use openmls::prelude::{Extension, UnknownExtension};
+    use tls_codec::Serialize as TlsSerialize;
 
+    use crate::extension::NostrGroupDataExtension;
     use crate::messages::MessageProcessingResult;
     use crate::test_util::*;
     use crate::tests::create_test_mdk;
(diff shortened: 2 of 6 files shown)


crates/mdk-core/src/messages/commit.rs
```diff
@@ -796,6 +799,98 @@ mod tests {
         );
     }
 
+    /// Test that an admin-authored commit cannot empty the admin set.
+    ///
+    /// This bypasses MDK's creating-side group-data guard and uses OpenMLS
+    /// directly to create the receiver-side shape reported in issue #29.
```
dannym-arx: perhaps this should either not say issue 29 or say https://github.com/marmot-protocol/marmot-security/issues/29 because a reader would be confused and think this is https://github.com/marmot-protocol/mdk/pull/29 which is of course a completely different thing

dannym-arx (changes requested):
I don't understand what this PR is solving.

We already have a function to check for admin depletion 

https://github.com/marmot-protocol/mdk/blob/8a8d06c19d617394b0e268f5f9adfad34db3cdf6/crates/mdk-core/src/messages/validation.rs#L236

It's even being called here on validation.rs:476

if for some reason that function is not enough (can't see why it wouldn't be), then it should be extended rather than create another one that does the same thing

crates/mdk-core/src/messages/validation.rs
```diff
@@ -270,6 +273,87 @@ where
...
+        Err(Error::Group("Would leave group with no admins".to_string()))
+    }
+
+    fn departing_leaf_indices(staged_commit: &StagedCommit) -> Vec<LeafNodeIndex> {
+        let mut departing_leaf_indices = Vec::new();
+
+        for queued in staged_commit.queued_proposals() {
+            match queued.proposal() {
+                Proposal::Remove(remove) => departing_leaf_indices.push(remove.removed()),
+                Proposal::SelfRemove => {
```
mubarakcoded: The SelfRemove arm is doing real work here, but there's no regression test for it. If a last admin self-removes, does this actually catch it end-to-end?

crates/mdk-core/src/messages/validation.rs
```diff
@@ -270,6 +273,87 @@ where
...
+    ) -> Result<()> {
+        for member in mls_group.members() {
+            let basic_cred = BasicCredential::try_from(member.credential)?;
+            let pubkey = self.parse_credential_identity(basic_cred.identity())?;
+            if admins.contains(&pubkey) {
+                return Ok(());
+            }
+        }
+
+        Err(Error::Group("Would leave group with no admins".to_string()))
```
jgmontoya: I think we have a bit of an inconsistency with when errors become `Unprocessable` vs when they're propagated as `Err`. This one for example becomes a `MessageProcessingResult::Unprocessable` while for others we have the specific error mapped (for example `Error::CommitFromNonAdmin`)

crates/mdk-core/src/groups.rs
```diff
@@ -1786,6 +1786,7 @@ where
         // Validate the mandatory group-data extension FIRST before making any state changes
         // This ensures we don't update stored_group if the extension is missing, invalid, or unsupported
         let group_data = NostrGroupDataExtension::from_group(&mls_group)?;
+        self.validate_active_admins_in_group(&mls_group, &group_data.admins)?;
```
jgmontoya, line 1784: Is this the right place to validate this? I think it could perhaps be validated before merging the commit, no?

crates/mdk-core/src/messages/validation.rs
```diff
@@ -270,6 +273,87 @@ where
...
+    /// The admin set is checked against live MLS members so stale admin entries
+    /// do not satisfy the invariant.
+    pub(crate) fn validate_active_admins_in_group(
+        &self,
+        mls_group: &MlsGroup,
+        admins: &BTreeSet<PublicKey>,
+    ) -> Result<()> {
+        for member in mls_group.members() {
+            let basic_cred = BasicCredential::try_from(member.credential)?;
+            let pubkey = self.parse_credential_identity(basic_cred.identity())?;
```
jgmontoya: I think this is repeated multiple times, we should probably extract it to a method, something like 

```rust
fn member_pubkey(&self, member: &Member) -> Result<PublicKey> {
    let basic_cred = BasicCredential::try_from(member.credential.clone())?;
    self.parse_credential_identity(basic_cred.identity())
}
```

crates/mdk-core/src/messages/validation.rs
```diff
@@ -270,6 +273,87 @@ where
...
+            if staged_group_data.admins.contains(&pubkey) {
+                return Ok(());
+            }
+        }
+
+        Err(Error::Group("Would leave group with no admins".to_string()))
+    }
+
+    fn departing_leaf_indices(staged_commit: &StagedCommit) -> Vec<LeafNodeIndex> {
```
jgmontoya: This should return a set (`BTreeSet<LeafNodeIndex>`) so that the .contains checks we're doing on `validate_post_commit_admin_invariant` become faster (it's also semantically closer to what we want: the set of departures)

crates/mdk-core/src/messages/validation.rs
```diff
@@ -389,6 +474,7 @@ where
                         .collect();
 
                     self.validate_admin_depletion(mls_group, &departing_leaves)?;
+                    self.validate_post_commit_admin_invariant(mls_group, staged_commit)?;
```
jgmontoya: Is this call necessary? I think this is a non-admin path so it seems that self-remove wouldn't really affect this invariant

crates/mdk-core/src/messages/commit.rs
```diff
@@ -796,6 +799,98 @@ mod tests {
...
+
+        let bob_group = bob_mdk
+            .get_group(&group_id)
+            .expect("Failed to get Bob's group")
+            .expect("Bob's group should exist");
+        assert_eq!(
+            bob_group.admin_pubkeys,
+            admins.into_iter().collect(),
+            "Rejected commit should not alter Bob's stored admin set"
+        );
```
jgmontoya, line 900: Should we also assert that the epoch didn't advance for Bob? Not really sure if needed but leaving it out there...

jgmontoya (commented):
I agree with @dannym-arx that this seems to be duplicating an already existing check (`validate_admin_depletion`), and digging deeper, it seems that the duplication is worse. From my agent:

> The admin-invariant predicate — _"the intersection of `group_data.admins` with live member pubkeys is non-empty"_ — is already computed **four separate times** in the codebase in **four different styles**, and none of them match:
> 
> | Site                               | File / line                                        | Style                                                                     |
> | ---------------------------------- | -------------------------------------------------- | ------------------------------------------------------------------------- |
> | Receiver-side SelfRemove rejection | `validation.rs:239` `validate_admin_depletion`     | Dual-count comparison over departing / active admins                      |
> | Creator-side admin-set update      | `groups.rs:1866` `prune_and_validate_admin_update` | `BTreeSet::filter → is_empty`                                             |
> | Self-demotion guard                | `groups.rs:1527-1548` (inside `update_group_data`) | Nested `filter + any + and_then + is_some_and` closure                    |
> | Member-pubkey extraction           | `groups.rs:502` `get_members`                      | `try_fold` over `BasicCredential::try_from` + `parse_credential_identity` |
> 
> This PR adds a **fifth** implementation in a **fifth** style (`validate_post_commit_admin_invariant`). The spec rule is one sentence; the code reads as five unrelated utilities that happen to touch the same concept.

<details>
<summary>
And this is the suggestion the agent made about this:
</summary>

The right primitive is already in the tree. **`groups.rs:202`** already defines

```rust
fn pubkey_from_credential(&self, credential: &Credential) -> Result<PublicKey, Error> {
    let basic = BasicCredential::try_from(credential.clone())?;
    self.parse_credential_identity(basic.identity())
}
```

It's module-private (plain `fn`), which is why `validation.rs` hasn't used it — but a single visibility bump to `pub(crate) fn` makes it accessible from every validator. And `get_members` at `groups.rs:502` already manually reimplements exactly the body that should be built on top of this helper.

**The replacement I'd write.** Two primitives in `groups.rs`, right next to the existing `pubkey_from_credential`:

```rust
/// All current member pubkeys of an already-loaded MLS group.
pub(crate) fn member_pubkeys(
    &self,
    mls_group: &MlsGroup,
) -> Result<BTreeSet<PublicKey>, Error> {
    mls_group
        .members()
        .map(|m| self.pubkey_from_credential(&m.credential))
        .collect()
}

/// The member pubkey set that would exist after applying `staged_commit`.
pub(crate) fn post_commit_member_pubkeys(
    &self,
    mls_group: &MlsGroup,
    staged_commit: &StagedCommit,
) -> Result<BTreeSet<PublicKey>, Error> {
    let departing: BTreeSet<LeafNodeIndex> = staged_commit
        .queued_proposals()
        .filter_map(|q| match q.proposal() {
            Proposal::Remove(r) => Some(r.removed()),
            Proposal::SelfRemove => match q.sender() {
                Sender::Member(idx) => Some(*idx),
                _ => None,
            },
            _ => None,
        })
        .collect();

    let mut pubkeys: BTreeSet<PublicKey> = mls_group
        .members()
        .filter(|m| !departing.contains(&m.index))
        .map(|m| self.pubkey_from_credential(&m.credential))
        .collect::<Result<_, _>>()?;

    for add in staged_commit.add_proposals() {
        pubkeys.insert(
            self.pubkey_from_credential(
                add.add_proposal().key_package().leaf_node().credential(),
            )?,
        );
    }

    Ok(pubkeys)
}
```

Then the entire PR's new validation logic collapses to **one** function in `validation.rs`:

```rust
/// Rejects a commit whose post-commit admin set is disjoint from the
/// post-commit membership. Per MIP-03.
pub(super) fn validate_admin_invariant_after_commit(
    &self,
    mls_group: &MlsGroup,
    staged_commit: &StagedCommit,
) -> Result<()> {
    let post_admins =
        NostrGroupDataExtension::from_group_context(staged_commit.group_context())?.admins;
    let post_members = self.post_commit_member_pubkeys(mls_group, staged_commit)?;

    if post_admins.is_disjoint(&post_members) {
        return Err(Error::AdminDepletion);
    }
    Ok(())
}
```

That is the whole check. The MIP-03 sentence — _"the resulting admin_pubkeys set would be empty"_ — becomes a single-line set-theoretic expression: `post_admins.is_disjoint(&post_members)`. A future reader gets the invariant at a glance with zero plumbing to parse.

**Call sites become one line each.** In `validate_commit_authorization`:

```rust
// admin-sender branch: replaces the current dual call
if sender_is_admin {
    self.validate_admin_invariant_after_commit(mls_group, staged_commit)?;
    return Ok(());
}

// non-admin SelfRemove-only branch: replaces BOTH validate_admin_depletion
// AND validate_post_commit_admin_invariant
self.validate_admin_invariant_after_commit(mls_group, staged_commit)?;
```

In `sync_group_metadata_from_mls` (if the defence-in-depth guard is kept at all — see Issue #2):

```rust
let members = self.member_pubkeys(&mls_group)?;
if group_data.admins.is_disjoint(&members) {
    return Err(Error::AdminDepletion);
}
```

**What gets deleted.** `validate_admin_depletion` (~35 lines), `validate_active_admins_in_group` (~15 lines), `validate_post_commit_admin_invariant` (~37 lines), `departing_leaf_indices` (~17 lines — inlined into `post_commit_member_pubkeys`), the `departing_leaves` `filter_map` boilerplate at `validation.rs:468-474`, and all four duplicated `BasicCredential::try_from → parse_credential_identity` blocks in the new code.

**Line-count estimate.** PR as-is: +239 / −1. Replacement: roughly +95 / −38. Same receiver-side behaviour, broader invariant coverage (admin-authored `GroupContextExtensions` updates and SelfRemove-depletion are now caught by the _same_ check via the _same_ code path), and the whole admin invariant is stated once in the language of sets.

**Edge cases I traced against the replacement.**

- `Remove` + `Add` of the same pubkey (rejoin with fresh key package): `BTreeSet` deduplicates. ✓
- Pure `GroupContextExtensions` update (the marmot-security#29 attack): no departures, no adds; `post_admins` == the cleared set; disjointness check fires. ✓
- SelfRemove-only commit from the last admin: `add_proposals` empty; `post_admins` unchanged from pre-commit; `post_members` excludes the departing admin; disjoint → rejected. ✓
- Admin removes another admin where both exist: `post_admins` unchanged, `post_members` excludes one; remaining admin is in both sets → accepted. ✓
- Admin simultaneously removes the last existing admin and adds a new admin: the added pubkey appears in `post_members`; whether it appears in `post_admins` depends on whether the commit also updated the group-data extension. If yes, accepted (owner swap). If no, rejected (dangling commit that would leave an admin entry pointing at no-one). Both are correct. ✓

**Bonus refactors this unlocks** (mention them in the PR description, don't include them in the same PR — keep the security fix focused):

1. `get_members` at `groups.rs:502` collapses to three lines:

   ```rust
   pub fn get_members(&self, group_id: &GroupId) -> Result<BTreeSet<PublicKey>, Error> {
       let group = self.load_mls_group(group_id)?.ok_or(Error::GroupNotFound)?;
       self.member_pubkeys(&group)
   }
   ```

2. The self-demotion guard at `groups.rs:1527-1548` (fourteen lines of nested closure) collapses to:

   ```rust
   let members = self.member_pubkeys(&mls_group)?;
   if group_data.admins.intersection(&members).count() <= 1 {
       return Err(Error::Group(
           "Cannot self-demote: last active admin. \
            Designate another admin first using update_group_data.".into(),
       ));
   }
   ```

3. The duplicated `BasicCredential::try_from → parse_credential_identity` pattern — which appears ~15 times across `validation.rs`, `groups.rs`, and `messages/proposal.rs` — becomes `self.pubkey_from_credential(&credential)` everywhere. Issue #3 collapses to a find-and-replace.
</details>

erskingardner:
> I don't understand what this PR is solving.

@dannym-arx I think the existing `validate_admin_depletion` helps for the "admins are removed as members" case, but it doesn’t catch the extension-only case from the report. A malicious admin can commit `admin_pubkeys = []` without removing anyone, so there are no departing leaves for that helper to count. This PR is trying to validate the post-commit effective admin state, not just current admins minus removals.

That said, I agree the shape is confusing. I can refactor/rename this so it’s clearer how it differs from the existing helper instead of looking like duplicate theatre.

erskingardner:
Just seeing @jgmontoya's comments (damn Github is going downhill)... going to take a closer look at all of it and then come back to this.

erskingardner:
Update after the validation review: I added a small `state_validation` module and moved only the active-admin/member-state pieces this PR needs into it: live member pubkey extraction, post-departure/post-commit member sets, and the shared “admins must intersect live members” check.

I’m intentionally keeping the bigger cleanup out of this PR. Follow-up should move the rest of proposal/commit/group-evolution validation into this module in smaller passes, including the identity checks and error-shape consistency.

dannym-arx (approved):

mubarakcoded (approved):

PR #257, opened 2026-04-14
Fix memory storage message payload bounds

erskingardner:
## Summary
- Enforce SQLite-matching size limits in `mdk-memory-storage` for message content, serialized tags JSON, and serialized event JSON.
- Reject oversized messages before they are inserted into memory caches.
- Document the fixed message payload limits and add regression coverage for each rejected payload type.

## Testing
- `cargo test -p mdk-memory-storage save_message_rejects_oversized -- --nocapture`
- `cargo test -p mdk-memory-storage`
- `cargo clippy -p mdk-memory-storage --all-targets --all-features -- -D warnings`
- `cargo doc -p mdk-memory-storage --no-deps`
- `cargo fmt --check`
- `just check`

Fixes https://github.com/marmot-protocol/marmot-security/issues/10

### DIFF
--- a/crates/mdk-memory-storage/CHANGELOG.md
+++ b/crates/mdk-memory-storage/CHANGELOG.md
@@ -17,6 +17,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


dannym-arx (approved):

jgmontoya (approved):
LGTM!

PR #258, opened 2026-04-14
feat: extension v3 wire format + disappearing_message_secs schema (1/3)

dannym-arx:
marmot

## Step 1 of 3: splitting #253

This PR lays the wire-format and storage groundwork for disappearing messages without adding any business logic. A marmot architect drafting blueprints before the builders arrive.

### What changed

**Extension v3 wire format.** `NostrGroupDataExtension` gains a `disappearing_message_secs` field serialized as a variable-length `Vec<u8>` (empty = `None`, 8 bytes = `Some(u64)`). The version header bumps from 2 to 3. Deserialization dispatches on the first two bytes: v1/v2 payloads decode through `TlsNostrGroupDataExtensionV1V2` and promote to v3 with the field defaulting to `None`. A dedicated test covers the adversarial case where a v3 header wraps a v1/v2-sized body.

**Storage schema.** V006 migration adds a nullable `disappearing_message_secs INTEGER` column to the `groups` table, and `row_to_group` / `save_group` read and write it. `Group` in `mdk-storage-traits` carries the new field.

**Config structs.** `NostrGroupConfigData` and `NostrGroupDataUpdate` both gain the field. `NostrGroupDataUpdate` uses `Option<Option<u64>>` where the outer option means "was it specified?" and the inner means "enabled or disabled?". The value flows through `create_group`, `update_group_data`, `sync_group_metadata`, and welcome processing.

**Boilerplate.** Every existing `Group` construction site across tests, examples, and storage crates adds `disappearing_message_secs: None`.

### What this PR does NOT add

- No validation (zero-duration rejection comes in PR 2)
- No NIP-40 expiration tag logic (PR 2)
- No message deletion methods (PR 3)
- No `PRAGMA secure_delete` (PR 3)
- No UniFFI binding changes beyond the `create_group` call site fix

### Files touched (24)

| Area | Files | Lines |
|------|-------|-------|
| Extension v3 | `extension/types.rs` | +327 |
| Storage traits | `groups/types.rs`, `test_utils.rs` | +8 |
| SQLite | `V006` migration, `db.rs`, `groups.rs` | +35 |
| Core structs | `groups.rs` | +16 |
| Boilerplate None | 14 files across crates | +33 |

This PR implements groundwork for disappearing messages by introducing a v3 NostrGroupDataExtension wire format (bumping the extension version 2→3), adding persistent storage for an optional disappearing_message_secs duration, and threading the new field through group creation, updates, syncing, and welcome processing; it deliberately does not implement expiration enforcement, message deletion, or broader UniFFI API surface changes. TLS serialization/deserialization is updated to accept v3 payloads while promoting legacy v1/v2 payloads to v3 with disappearing_message_secs = None and to robustly reject malformed inputs.

**What changed**:
- Added disappearing_message_secs: Option<u64> to NostrGroupDataExtension, NostrGroupConfigData, NostrGroupDataUpdate, and the public Group struct (affects mdk-core, mdk-storage-traits).
- Bumped NostrGroupDataExtension::CURRENT_VERSION from 2 → 3 and extended the v3 TLS representation to encode disappearing_message_secs as empty Vec (None) or an 8-byte big-endian u64 (mdk-core).
- Reworked TLS deserialization to read a big-endian u16 version from the first two bytes, dispatch to the v3 parser for version ≥3, or parse legacy v1/v2 payloads and promote them to v3 with disappearing_message_secs = None (mdk-core).
- Added validation that non-empty disappearing_message_secs bytes must be exactly 8 bytes and that zero-valued durations are rejected as invalid during parsing (mdk-core).
- Introduced SQLite migration V006 adding a nullable disappearing_message_secs INTEGER column to groups and updated row_to_group, save_group, and upsert logic to read/write the new column (mdk-sqlite-storage).
- Propagated the new field through create_group, update_group_data, sync_group_metadata, welcome processing, and updated all Group construction sites in tests and examples to include disappearing_message_secs: None (mdk-core, mdk-memory-storage, mdk-sqlite-storage, mdk-storage-traits, mdk-uniffi).

**Security impact**:
- No cryptographic algorithm, key derivation, nonce/HKDF context, identity-binding, key handling/storage, error-message leakage, SQLCipher configuration, PRAGMA secure_delete, or file-permission changes are introduced by this PR.

**Protocol changes**:
- Nostr group-data extension wire format advanced to v3 to carry disappearing_message_secs, deserialization now rejects malformed v3 headers lacking a proper v3 body, and legacy v1/v2 payloads are still supported via promotion (Nostr extension behavior change).

**API surface**:
- Breaking changes: NostrGroupDataExtension::CURRENT_VERSION changed to 3 and NostrGroupConfigData::new(...) now requires an additional disappearing_message_secs: Option<u64> argument (mdk-core).
- Public storage-traits change: Group now includes disappearing_message_secs: Option<u64>, requiring all Group constructors to supply the field (mdk-storage-traits).
- Storage schema change: V006 migration adds groups.disappearing_message_secs INTEGER (nullable) to SQLite (mdk-sqlite-storage).
- UniFFI: create_group call-site updated to pass disappearing_message_secs: None; no other UniFFI surface changes.

**Testing**:
- Added unit tests covering v3 round-trip serialization/deserialization, legacy v1/v2 promotion, rejection of zero-valued durations, and rejection of a v3 header paired with a v1/v2-shaped body (mdk-core).
- Updated numerous tests and examples to set disappearing_message_secs: None in Group literals and added SQLite tests verifying persistence, upsert updates, and clearing of the new column (mdk-sqlite-storage, mdk-memory-storage, mdk-core).

### DIFF
--- a/crates/mdk-core/src/extension/types.rs
+++ b/crates/mdk-core/src/extension/types.rs
@@ -10,13 +10,14 @@ use nostr::{PublicKey, RelayUrl};
 use openmls::extensions::{Extension, ExtensionType};
 use openmls::group::{GroupContext, MlsGroup};
 use tls_codec::{
-    DeserializeBytes, TlsDeserialize, TlsDeserializeBytes, TlsSerialize, TlsSerializeBytes, TlsSize,
+    DeserializeBytes, Serialize as TlsSerializeTrait, TlsDeserialize, TlsDeserializeBytes,
+    TlsSerialize, TlsSerializeBytes, TlsSize,
 };
 
 use crate::constant::NOSTR_GROUP_DATA_EXTENSION_TYPE;
 use crate::error::Error;
 
-/// TLS-serializable representation of Nostr Group Data Extension.
+/// TLS-serializable representation of Nostr Group Data Extension (v3+).
 ///
 /// This struct is used exclusively for TLS codec serialization/deserialization
 /// when the extension is transmitted over the MLS protocol. It uses `Vec<u8>`
@@ -47,13 +48,62 @@ pub(crate) struct TlsNostrGroupDataExtension {
     pub image_key: Vec<u8>,        // Use Vec<u8> to allow empty for None
     pub image_nonce: Vec<u8>,      // Use Vec<u8> to allow empty for None
     pub image_upload_key: Vec<u8>, // Use Vec<u8> to allow empty for None (v2 only)
+    pub disappearing_message_secs: Vec<u8>, // Use Vec<u8>: empty for None, 8 bytes for Some(u64) (v3 only)
+}
+
+/// TLS-serializable representation for v1/v2 payloads (before disappearing messages).
+///
+/// Legacy groups serialized without the `disappearing_message_secs` field
+/// use this struct for deserialization. The missing field is synthesized as an empty
+/// Vec (mapping to `None`) when converting to `TlsNostrGroupDataExtension`.
+#[derive(
+    Clone,
+    PartialEq,
+    Eq,
+    TlsSerialize,
+    TlsDeserialize,
+    TlsSerializeBytes,
+    TlsDeserializeBytes,
+    TlsSize,
+)]
+struct TlsNostrGroupDataExtensionV1V2 {
+    pub version: u16,
+    pub nostr_group_id: [u8; 32],
+    pub name: Vec<u8>,
+    pub description: Vec<u8>,
+    pub admin_pubkeys: Vec<[u8; 32]>,
+    pub relays: Vec<Vec<u8>>,
+    pub image_hash: Vec<u8>,
+    pub image_key: Vec<u8>,
+    pub image_nonce: Vec<u8>,
+    pub image_upload_key: Vec<u8>,
+}
+
+impl TlsNostrGroupDataExtensionV1V2 {
+    /// Promote to the v3 struct with the missing field defaulted to empty.
+    fn into_v3(self) -> TlsNostrGroupDataExtension {
+        TlsNostrGroupDataExtension {
+            version: self.version,
+            nostr_group_id: self.nostr_group_id,
+            name: self.name,
+            description: self.description,
+            admin_pubkeys: self.admin_pubkeys,
+            relays: self.relays,
+            image_hash: self.image_hash,
+            image_key: self.image_key,
+            image_nonce: self.image_nonce,
+            image_upload_key: self.image_upload_key,
+            disappearing_message_secs: Vec::new(),
+        }
+    }
 }
 
 /// This is an MLS Group Context extension used to store the group's name,
 /// description, ID, admin identities, image URL, and image encryption key.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct NostrGroupDataExtension {
-    /// Extension format version (current: 2)
+    /// Extension format version (current: 3)
+    /// Version 3: Adds disappearing_message_secs field
     /// Version 2: image_key field contains image_seed, image_upload_key contains upload_seed
     /// Version 1: image_key field contains encryption key directly (deprecated)
     pub version: u16,
@@ -84,6 +134,11 @@ pub struct NostrGroupDataExtension {
     /// In v2, the upload keypair is derived from this seed (cryptographically independent from image_key).
     /// In v1, the upload keypair was derived from image_key (now deprecated).
     pub image_upload_key: Option<[u8; 32]>,
+    /// Disappearing message duration in seconds (v3 only)
+    ///
+    /// - `None`: Messages persist forever (disabled)
+    /// - `Some(n)`: Messages expire `n` seconds after creation (`n > 0`)
+    pub disappearing_message_secs: Option<u64>,
 }
 
 impl NostrGroupDataExtension {
@@ -91,9 +146,10 @@ impl NostrGroupDataExtension {
     pub const EXTENSION_TYPE: u16 = NOSTR_GROUP_DATA_EXTENSION_TYPE;
 
     /// Current extension format version (MIP-01)
+    /// Version 3: Adds disappearing_message_secs field
     /// Version 2: Uses image_seed (stored in image_key field) with HKDF derivation
     /// Version 1: Uses image_key directly as encryption key (deprecated)
-    pub const CURRENT_VERSION: u16 = 2;
+    pub const CURRENT_VERSION: u16 = 3;
 
     /// Creates a new NostrGroupDataExtension with the given parameters.
     ///
@@ -119,6 +175,7 @@ impl NostrGroupDataExtension {
         image_key: Option<[u8; 32]>,
         image_nonce: Option<[u8; 12]>,
         image_upload_key: Option<[u8; 32]>,
+        disappearing_message_secs: Option<u64>,
     ) -> Self
     where
         T1: Into<String>,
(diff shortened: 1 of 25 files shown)


crates/mdk-sqlite-storage/src/groups.rs
```diff
@@ -249,7 +250,8 @@ impl GroupStorage for MdkSqliteStorage {
                     &last_message_processed_at,
                     &(group.epoch as i64),
                     group.state.as_str(),
-                    &last_self_update_at
+                    &last_self_update_at,
+                    &group.disappearing_message_secs.and_then(|d| i64::try_from(d).ok())
```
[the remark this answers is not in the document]
dannym-arx: lol doesn't matter

crates/mdk-core/src/extension/types.rs
```diff
@@ -156,17 +213,38 @@ impl NostrGroupDataExtension {
...
+        } else {
+            let (v1v2, remainder) = TlsNostrGroupDataExtensionV1V2::tls_deserialize_bytes(bytes)?;
+            if !remainder.is_empty() {
+                return Err(Error::ExtensionFormatError(
+                    "Trailing bytes in NostrGroupDataExtension".to_string(),
+                ));
+            }
+            v1v2.into_v3()
+        };
+        Self::from_raw(raw)
```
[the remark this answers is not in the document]
dannym-arx: Rather than branching as_raw() to emit a legacy body shape, the simpler fix is to always write version: CURRENT_VERSION in as_raw(). The serialized body is always v3-shaped, so the version header should match. This auto-upgrades legacy extensions on the next group update. Committing that fix

crates/mdk-core/src/extension/types.rs
```diff
@@ -416,10 +521,10 @@ impl NostrGroupDataExtension {
         {
             tracing::warn!(
                 target: "mdk_core::extension::types",
-                "Migrating from v1 to v2 without new image_seed and image_upload_seed - existing image_key will be treated as seed, which may cause issues since v1 image_key is a direct encryption key, not a seed"
+                "Migrating from v1 to v3 without new image_seed and image_upload_seed - existing image_key will be treated as seed, which may cause issues since v1 image_key is a direct encryption key, not a seed"
```
erskingardner: should we even allow people to try and do this? wouldn't it be better to just have two methods, one for v1->v2 and another for v2->v3?

dannym-arx: maybe? there's no new logic happening here, I'm just renaming it to v3 for consistency, but there's no extra migration, you can migrate from v1 to v3 and from v2 to v3 without a problem, because there's no new data that needs to be created

crates/mdk-core/src/extension/types.rs
```diff
@@ -476,7 +582,9 @@ impl NostrGroupDataExtension {
 
     pub(crate) fn as_raw(&self) -> TlsNostrGroupDataExtension {
         TlsNostrGroupDataExtension {
-            version: self.version,
+            // Always write the current version — the body shape is always v3
+            // (includes disappearing_message_secs), so the header must match.
+            version: Self::CURRENT_VERSION,
```
erskingardner: I don't think that we have backwards compatibility here. When an older client hits a new version that it doesn't understand it doesn't know how to read only the bytes it knows about it, it'll be a hard fail. 

> Answer: Old clients will hard-fail, not gracefully degrade.
> The tracing::warn! block in master that says "Continue processing with forward compatibility - unknown fields will be ignored" is aspirational, not real. It only helps if a future version bumps version but keeps the field layout byte-identical. The moment a new field is appended (as PR #258 does), the old parser hits its own Trailing bytes in NostrGroupDataExtension check and returns Error::ExtensionFormatError.
> 
> Concretely, once a v3-serializing client performs a commit:
> 
> Every v1/v2 peer fails from_group_context / from_group → can't process welcomes, commits, or inspect the group.
> This is true even when disappearing_message_secs = None, because the empty Vec<u8> still emits a length prefix on the wire (no empty-case shortcut).
> The PR authors implicitly acknowledge this: they added TlsNostrGroupDataExtensionV1V2 + a version-dispatch at the top of the new deserialize_bytes so new clients can read old data. But old clients have no symmetric escape hatch — they don't know to stop reading early.
> 
> This amplifies my review point #2: silently upgrading every commit to v3 is a hard-break for unupgraded peers, not a soft warning. Options:
> 
> Gate the version bump: only write v3 when disappearing_message_secs.is_some(); keep v2 wire format otherwise.
> Accept the break and coordinate a flag-day (document it loudly, dual-publish period, etc.).
> Redesign so "unknown trailing fields" are actually ignorable (e.g., length-prefixed envelope around the struct) — larger change.
> The current "forward compatibility" comment in master should probably be deleted or rewritten; it's been giving false confidence.

crates/mdk-core/src/extension/types.rs
```diff
@@ -494,13 +602,18 @@ impl NostrGroupDataExtension {
             image_upload_key: self
                 .image_upload_key
                 .map_or_else(Vec::new, |key| key.to_vec()),
+            disappearing_message_secs: self
+                .disappearing_message_secs
+                .filter(|&d| d != 0)
```
erskingardner: > ```rust
> // In new():
> let disappearing_message_secs = disappearing_message_secs.filter(|&d| d != 0);
> 
> // In as_raw():
> disappearing_message_secs: self
>     .disappearing_message_secs
>     .filter(|&d| d != 0)
>     .map_or_else(Vec::new, |d| d.to_be_bytes().to_vec()),
> ```
> 
> Zero normalization is applied in two different places: in the constructor and again in `as_raw()`. If a `NostrGroupDataExtension` struct is constructed directly (not via `new()`) with `disappearing_message_secs: Some(0)`, the struct field will hold `Some(0)` until `as_raw()` normalizes it away. This creates a window where `extension.disappearing_message_secs == Some(0)` is observable in the struct but would serialize as `None`. The `from_raw()` path also rejects zero with an error, which means a round-trip of a manually constructed zero-value would be an error on read-back but was silently accepted on write. Pick one authoritative enforcement point: `from_raw()` should be the read-side gate; `new()` and any public setter should be the write-side gate; `as_raw()` should not need to re-check.

crates/mdk-core/src/extension/types.rs
```diff
@@ -395,43 +508,26 @@ impl NostrGroupDataExtension {
...
+    /// // Migrate extension from v1 to v2
+    /// extension.migrate_v1_to_v2(
+    ///     new_hash,
+    ///     v2_prepared.image_key,
+    ///     v2_prepared.image_nonce,
+    ///     v2_prepared.upload_seed,
     /// );
     /// ```
-    pub fn migrate_to_v2(
+    pub fn migrate_v1_to_v2(
```
erskingardner: I don't know if this is really worth addressing but it's a breaking change.

> The public method `migrate_to_v2` is renamed to `migrate_v1_to_v2` and its signature is changed from accepting `Option<T>` arguments to required `T` arguments. This is a breaking change for any downstream callers. If the method is intended as public API, a `#[deprecated]` redirect from the old name should be left for at least one release. If the method has no external callers, tighten visibility to `pub(crate)`. The PR description says "no UniFFI binding changes beyond the `create_group` call site fix", which suggests this is internal — but it is marked `pub`.

erskingardner (commented):
One small comment

crates/mdk-core/src/extension/types.rs
dannym-arx: I guess I'll change it to pub(crate) because nobody should call this

erskingardner (approved):

PR #261, opened 2026-04-17
fix(mdk-core): compute group RequiredCapabilities as LCD of invitee capabilities to unblock mixed-version invites (whitenoise-rs#749)

jgmontoya:
This PR changes how group RequiredCapabilities are computed so invitee capabilities are intersected (LCD) at group creation, unblocking mixed-version invites that were previously forced to require SelfRemove; it also updates leave/self-update behavior to honor per-group capabilities, removes now-unused helpers/constants, refreshes rotated-leaf capability advertising, and adds tests and a legacy KeyPackage test helper to validate mixed/legacy scenarios. These changes focus on capability/extension handling and test utilities rather than cryptographic primitives or storage of secrets.

**What changed**:
- create_group now pre-parses invitee KeyPackages and computes the group's RequiredCapabilities.proposal_types as the least-common-denominator intersection of SUPPORTED_PROPOSALS with each invitee's advertised proposals, embedding that as the RequiredCapabilities extension (crates/mdk-core).  
- leave_group / try_self_remove now consult the group's RequiredCapabilities and fall back to an admin-committable legacy Remove when ProposalType::SelfRemove is not included, preserving the existing PURE_CIPHERTEXT wire-format fallback (crates/mdk-core).  
- self_update now uses MDK::capabilities() for rotated LeafNodeParameters so rotated leaves advertise the MDK capability set rather than cloning the old leaf capabilities (crates/mdk-core).  
- Removed the dead GROUP_CONTEXT_REQUIRED_PROPOSALS constant and the crate-private MDK::required_capabilities_extension() helper, moving responsibility to per-call-site LCD computation (crates/mdk-core).  
- Added create_legacy_key_package_event test helper to emit capability-poor legacy KeyPackage events (kind:443) for tests and expanded unit tests covering LCD-based RequiredCapabilities, leave/self-remove fallback, self_update behavior, ciphertext/exporter-secret persistence, and mixed/legacy regressions (crates/mdk-core).  
- Minor internal iterator/sorting cleanups in mdk-memory-storage (prune/exporter-secret key iteration, processed message iteration, pending-welcomes sort) with no behavioral change (crates/mdk-memory-storage).

**Security impact**:
- No changes to cryptographic primitives, key derivation, AEAD, HKDF contexts, SQLCipher configuration, file permissions, or secret storage semantics.  
- Rotated leaves now advertise the current MDK capabilities (protocol behavior change) but this does not expose secret material or change key management.

**Protocol changes**:
- Group RequiredCapabilities semantics now reflect the LCD of invitee-advertised proposals, affecting MLS admission/extension behavior (notably whether SelfRemove is required) and interacting with OpenMLS admission checks and MIP-03-related compatibility scenarios.  
- Nostr integration: test tooling added to emit legacy kind:443 KeyPackage events to model pre-MIP-03 clients; no production wire-format changes beyond testing helpers.

**API surface**:
- Removed public constant GROUP_CONTEXT_REQUIRED_PROPOSALS from crates/mdk-core/src/constant.rs.  
- Removed crate-private method MDK::required_capabilities_extension() from crates/mdk-core/src/lib.rs.  
- Added public test helper fn create_legacy_key_package_event<Storage>(mdk: &MDK<Storage>, member_keys: &Keys) -> Event in crates/mdk-core/src/test_util.rs.  
- No storage schema, UniFFI/FFI boundary, or other public API breaking changes beyond the above removals/addition.

**Testing**:
- Added and expanded unit tests in crates/mdk-core for RequiredCapabilities computation at group creation, leave/self-remove fallback under capability constraints, self_update capability refresh, ciphertext/exporter-secret persistence checks, and mixed/legacy regression scenarios, and introduced a legacy KeyPackage event test helper.

### DIFF
--- a/crates/mdk-core/CHANGELOG.md
+++ b/crates/mdk-core/CHANGELOG.md
@@ -39,6 +39,8 @@
 - New groups now use `MIXED_CIPHERTEXT` wire format policy (outgoing: ciphertext, incoming: mixed) instead of the OpenMLS default `PURE_CIPHERTEXT`. Regular messages remain `PrivateMessage`; the mixed incoming policy is required to accept `PublicMessage` SelfRemove proposals from departing members.
 - `leave_group` now sends a SelfRemove proposal (MLS Extensions draft, type `0x000a`) instead of a Remove proposal. Falls back to Remove for legacy groups created with `PURE_CIPHERTEXT`. SelfRemove proposals are auto-committed by any member, removing the requirement for an admin to be online.
 - Non-admin members can now create SelfRemove-only Commits in addition to self-update Commits. These two commit types must not be combined.
+- **`create_group` now computes the group's `RequiredCapabilities` as a least-common-denominator (LCD) intersection of `SUPPORTED_PROPOSALS` with every invitee's advertised proposals**, instead of requiring `SelfRemove` unconditionally. An all-modern invitee set still produces `RequiredCapabilities = [SelfRemove]`; any legacy invitee strips `SelfRemove` from the required set. Empty-invitee (single-member / "message to self") groups stay permissive with `RequiredCapabilities = []` so a later `add_members` call can admit legacy key packages. This unblocks two call patterns that previously failed OpenMLS's admission check with `Error::Group("Proposals are not acceptable")`: `create_group` with at least one legacy invitee, and `add_members` of a legacy invitee into a group that was created mixed (`RequiredCapabilities = []`). Resolves [whitenoise-rs #749](https://github.com/marmot-protocol/whitenoise-rs/issues/749) via [#261](https://github.com/marmot-protocol/mdk/pull/261).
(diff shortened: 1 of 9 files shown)


dannym-arx (approved):
It's fine for now, but I think we should consider some mitigation for the following scenario in the future, because LCD may have some issues... 

Say MDK v0.10 adds something like `ForceEpochRotation` to SUPPORTED_PROPOSALS because we discover that without it, a compromised member can stall epoch advancement. Now:

1. Alice (v0.10) creates a group with Bob (v0.10) and Charlie (v0.8, legacy)
2. LCD strips ForceEpochRotation from RequiredCapabilities
3. Alice has no idea the group is missing a security guarantee she expected

Essentially an adversary can choose to publish a capability-poor KP to affect a group, what are your thoughts on this?

mubarakcoded (approved):

PR #270, opened 2026-05-04
[codex] Encode SQLite keys for keyring storage

### LINKED ISSUE #268: Encode SQLite DB keys before storing in keyring
## Summary

`mdk-sqlite-storage` currently stores database encryption keys as raw 32-byte secrets in `keyring-core`. That works for backends that accept arbitrary binary data, but it is not portable across all Secret Service implementations. In particular, KWallet-backed Secret Service is documented/known to require UTF-8/string-compatible secret payloads, so storing random 32-byte key material can fail or behave incorrectly on KDE Linux desktops.

This surfaced while updating White Noise to use the Secret Service keyring backend on Linux. White Noise can choose the platform keyring store, but MDK owns the MLS SQLite DB encryption key format and read/write path, so the durable fix should live here.

## Current behavior

In `crates/mdk-sqlite-storage/src/encryption.rs`:

- `EncryptionConfig::generate()` creates 32 random bytes.
- `EncryptionConfig::key()` exposes `&[u8; 32]`.

In `crates/mdk-sqlite-storage/src/keyring.rs`:

- `get_or_create_db_key()` stores the raw key with `entry.set_secret(config.key())`.
- `get_db_key()` reads bytes with `entry.get_secret()` and passes them directly to `EncryptionConfig::from_slice(&secret)`.
- `EncryptionConfig::from_slice()` accepts only exactly 32 bytes.

That means the keyring payload is currently the raw binary DB key.

## Proposed fix

Store MDK SQLite DB keys in a backend-safe encoded representation instead of raw bytes. Base64 is probably the simplest choice.

Suggested shape:

- Add helper(s) near the keyring storage path, for example:
  - `encode_db_key_for_keyring(config: &EncryptionConfig) -> Vec<u8>` or `String`
  - `decode_db_key_from_keyring(secret: &[u8]) -> Result<EncryptionConfig, Error>`
- On new key creation, store the encoded form with `entry.set_secret(...)`.
- On read, decode the new format before constructing `EncryptionConfig`.

A version/prefix would make future migrations safer, e.g. store `mdk-sqlite-key-v1:<base64>` rather than bare base64. Bare base64 is possible too, but prefixing removes ambiguity with legacy payloads.

## Required migration behavior

This must preserve existing installs that already have raw 32-byte keys in the keyring.

Recommended read path:

1. `entry.get_secret()`.
2. Try to parse the new encoded format first.
   - If using a prefix, check for `mdk-sqlite-key-v1:` and base64-decode the suffix.
   - Validate that decoded key material is exactly 32 bytes.
3. If the payload is not in the new format and `secret.len() == 32`, treat it as the legacy raw-key format.
4. For a legacy raw key:
   - Build `EncryptionConfig::from_slice(&secret)`.
   - Rewrite the same entry using the new encoded format.
   - Return the config only if the rewrite succeeds, or deliberately document/log if we choose best-effort rewrite.
5. If the payload is neither recognized encoded format nor exactly 32 raw bytes, return the existing invalid-key/keyring error path.

Important: do not generate a new DB key when an existing invalid/unrecognized payload is present. Generating a new key would make an existing encrypted DB unreadable.

## Tests to add

Please add tests covering at least:

- New keys are stored in encoded form, not raw 32-byte form.
- Encoded keys read back to the original `EncryptionConfig`.
- Legacy raw 32-byte entries are read successfully and migrated/re-written to encoded form.
- Invalid encoded payloads fail clearly and do not generate a replacement key.
- Existing short/long invalid payload tests still behave correctly, adjusted for the new parser.

If possible, include a small fake/mock keyring store test so we can assert the exact stored bytes after migration without depending on a real platform keychain.

## Notes

This is especially relevant for Linux Secret Service/KWallet, but storing printable/UTF-8-safe keyring payloads is a safer invariant for all keyring backends.

erskingardner:
## Summary

- Store new `mdk-sqlite-storage` database keyring payloads as UTF-8-safe `mdk-sqlite-key-v1:<base64>` values instead of raw 32-byte keys.
- Read the encoded format first, while preserving legacy raw 32-byte payload compatibility.
- Migrate legacy raw keyring entries to the encoded format after successful reads, without generating replacement keys for invalid existing payloads.
- Add the `mdk-sqlite-storage` changelog entry for this compatibility fix.

Fixes #268.

## Validation

- `cargo fmt`
- `cargo test -p mdk-sqlite-storage keyring --no-default-features`
- `just precommit`

⚠️ Security-sensitive changes

This PR changes how SQLite database encryption keys are stored in the keyring: raw 32-byte binary keys are replaced by a versioned, UTF-8-safe prefixed base64 encoding to ensure compatibility with string-only keyring backends, while preserving compatibility with and migrating legacy raw entries. The implementation prefers the new prefixed format when reading, falls back to legacy 32-byte payloads, and rewrites successful legacy reads into the encoded format without regenerating keys for unrecognized/invalid payloads.

**What changed**:
- Database encryption keys stored in the keyring are now encoded as prefixed base64 strings ("mdk-sqlite-key-v1:<base64>") instead of raw 32-byte secrets in the mdk-sqlite-storage crate.
- Added helper functions to encode/decode the DB key payload and a save_db_key helper that writes the encoded payload via keyring-core.
- get_db_key now attempts to decode the prefixed encoded payload first, falls back to legacy raw 32-byte payloads, and migrates legacy entries by re-saving them in the encoded format after successful reads.
- get_or_create_db_key uses the new save helper when generating and storing new keys (with an in-process lock to avoid races).
- Added base64 dependency to crates/mdk-sqlite-storage/Cargo.toml and updated CHANGELOG.md documenting the fix.

**Security impact**:
- Key storage format changed from binary to a versioned, prefixed base64 encoding to ensure UTF-8 safety across keyring backends, reducing platform compatibility issues that could prevent secure storage.
- Legacy 32-byte raw keys are accepted and automatically migrated to the encoded format after successful reads, reducing persistence of raw binary key material in keyrings.
- Invalid or unrecognized payloads (e.g., malformed base64, wrong lengths) return clear keyring errors and do not trigger key regeneration, preventing accidental replacement of keys that could render encrypted databases unreadable.
- No changes to cryptographic algorithms, key derivation, SQLCipher configuration, or file permission logic.

**API surface**:
- No breaking changes to public function signatures or exported types; changes are internal to mdk-sqlite-storage's keyring integration.
- No UniFFI/FFI or storage schema changes.

**Testing**:
- Tests added/updated in mdk-sqlite-storage to cover: new encoded payload storage, reading encoded payloads, migration of legacy raw 32-byte entries to encoded form, failure behavior for invalid encoded payloads (no replacement), and other legacy-length handling; tests use the crate's mock keyring store for deterministic assertions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1928,6 +1928,7 @@ dependencies = [
 name = "mdk-sqlite-storage"
 version = "0.7.1"
 dependencies = [
+ "base64",
  "getrandom 0.4.2",
  "hex",
  "keyring-core",
--- a/crates/mdk-sqlite-storage/CHANGELOG.md
+++ b/crates/mdk-sqlite-storage/CHANGELOG.md
@@ -41,6 +41,7 @@
 
 ### Fixed
 
+- Stored new SQLite database keyring payloads as prefixed base64 while continuing to read and migrate legacy raw 32-byte keyring entries. ([#270](https://github.com/marmot-protocol/mdk/pull/270))
 - Added runtime SQLCipher verification and a post-migration file header check so encrypted storage fails closed if linked against a SQLite provider without active SQLCipher support. ([#252](https://github.com/marmot-protocol/mdk/pull/252))
 - Treated only `:memory:` and empty paths as special SQLite paths, so colon-prefixed filenames still receive secure permissions and encrypted database verification. ([#252](https://github.com/marmot-protocol/mdk/pull/252))
 
--- a/crates/mdk-sqlite-storage/Cargo.toml
+++ b/crates/mdk-sqlite-storage/Cargo.toml
@@ -17,6 +17,7 @@ default = []
 test-utils = []
 
 [dependencies]
+base64.workspace = true
 getrandom.workspace = true
 hex = { workspace = true, features = ["std"] }
 keyring-core.workspace = true
--- a/crates/mdk-sqlite-storage/src/keyring.rs
+++ b/crates/mdk-sqlite-storage/src/keyring.rs
@@ -48,6 +48,7 @@ use crate::error::Error;
 /// simultaneously, only one will generate and store it.
 static KEY_GENERATION_LOCK: OnceLock<Mutex<()>> = OnceLock::new();
 
+// v1 stores the raw 32-byte SQLCipher key as RFC 4648 base64.
 const DB_KEY_KEYRING_PREFIX: &str = "mdk-sqlite-key-v1:";
 
 enum DecodedDbKey {
@@ -59,6 +60,8 @@ fn encode_db_key_for_keyring(config: &EncryptionConfig) -> Vec<u8> {
     format!("{DB_KEY_KEYRING_PREFIX}{}", BASE64.encode(config.key())).into_bytes()
 }
 
+/// Decode current prefixed payloads first; non-UTF-8 payloads and UTF-8 payloads
+/// without the prefix are both handled by the legacy raw-key path.
 fn decode_db_key_from_keyring(secret: &[u8]) -> Result<DecodedDbKey, Error> {
     match std::str::from_utf8(secret) {
         Ok(secret_text) => match secret_text.strip_prefix(DB_KEY_KEYRING_PREFIX) {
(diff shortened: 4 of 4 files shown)


erskingardner:
@coderabbitai full review please

dannym-arx (approved):

PR #306, opened 2026-05-19
feat: validation + NIP-40 expiration + UniFFI surface for disappearing messages (2/3)

dannym-arx:
marmot

Step 2 of 3: splitting #253. Part 1 (#258) landed the v3 wire format and storage column. This PR adds validation, the NIP-40 expiration tag, and the UniFFI surface. Part 3 handles deletion.

## What changed

**Validation.** `create_group` and `update_group_data` reject `Some(0)` with `Error::Group`. Callers pass `None` (or `Some(None)` on the update path) to disable. The extension constructor keeps its silent zero filter as a backstop; surfaced errors live at the public API where callers can act on them.

**NIP-40 expiration on outer wrappers.** `build_message_event` builds every outer kind:445 event: messages, proposals, commits. When the group has a `disappearing_message_secs` setting, the wrapper now carries an `expiration` tag. If the caller also supplies one (say, an ephemeral location burst), the earlier of the two wins. Group sets a ceiling; caller can ask for shorter but never longer. Both `created_at` and the expiration math share one `Timestamp::now()` snapshot so they cannot drift.

**UniFFI surface.** The `Group` record now exposes `disappearing_message_secs: Option<UInt64>`. On `GroupDataUpdate`, the field is `Option<Option<UInt64>>` (outer = specified, inner = enabled). For `create_group`, the binding takes a new `disappearing_message_secs` parameter. Mobile clients can read and write the duration through the binding now that the field is plumbed on both sides.

## What this PR does NOT add

- Local message deletion methods (Part 3)
- UniFFI deletion exports (Part 3)
- `PRAGMA secure_delete=ON` (Part 3)

## Files touched (6)

| Area | Net lines |
|---|---|
| `mdk-core/src/groups.rs` | +197 (logic + 4 tests) |
| `mdk-core/src/messages/create.rs` | +144 (4 tests) |
| `mdk-uniffi/src/lib.rs` | +32 |
| `mdk-core/src/extension/types.rs` | +5 / -2 |
| Changelogs | +6 |
| Total | +376 / -8 |

8 new tests pass. Full workspace suite (458 tests) green. Precommit (fmt, clippy, docs, cargo audit) passes on stable and MSRV.

A second batch of marmots, this one teaching messages how to stop existing on schedule.

This PR (step 2 of 3 for disappearing messages) adds validation for disappearing-message configuration, ensures outer wrapper events receive deterministic NIP-40 expiration tags when a group configures disappearing messages, and exposes the disappearing-message setting through the UniFFI surface so mobile clients can read/write it.

What changed:
- Validation: mdk-core create_group and update_group_data now reject Some(0) (and Some(Some(0)) on updates); callers must use None (or Some(None) on update) to disable disappearing messages.
- Expiration tagging: build_message_event now attaches a NIP-40 expiration tag to outer kind:445 wrapper events when the group has disappearing_message_secs, selecting the earlier of a caller-supplied expiration and the group-derived expiration, and omitting the tag when neither applies.
- Timestamp consistency: created_at and expiration calculations use a single Timestamp::now() snapshot to prevent drift and ensure deterministic expiration math.
- UniFFI surface: mdk-uniffi Mdk::create_group gains disappearing_message_secs: Option<UInt64>; GroupDataUpdate uses Option<Option<UInt64>> to represent unchanged/disable/set semantics; Group record exposes disappearing_message_secs: Option<UInt64>.
- Files/crates affected: mdk-core (groups, messages/create, extension/types, changelog) and mdk-uniffi (lib.rs, changelog); storage crates (mdk-sqlite-storage, mdk-memory-storage, mdk-storage-traits) were not modified.

Security impact:
⚠️ Security-sensitive changes
- Cryptographic operations: build_message_event snapshots now-then derives the MLS exporter secret for event encryption and generates per-event ephemeral signing keys as part of constructing outer wrapper events, so reviewers should audit exporter context strings and key handling.
- No changes to SQLCipher, file permissions, or deletion/secure-delete behavior in this PR (deletion handled in Part 3).
- Public API now rejects ambiguous zero encodings to avoid silent misconfiguration that could leak policy expectations.

Protocol changes:
- Nostr: automatic addition of NIP-40 (expiration) tags to outer wrapper (kind:445) events when a group's disappearing_message_secs is set, following NIP-40 semantics and preferring the earliest expiration between caller and group.
- MLS: no changes to MLS algorithms or protocol behavior, but code derives the MLS exporter secret during event creation to support encryption of the outer wrapper.

API surface:
- Breaking/unavoidable API change: UniFFI method signature Mdk::create_group(...) now accepts disappearing_message_secs: Option<u64>.
- New/changed API surface: GroupDataUpdate gains disappearing_message_secs: Option<Option<UInt64>> and Group record exposes disappearing_message_secs: Option<UInt64>.
- No storage schema changes in this PR; storage deletion and secure-deletion APIs are planned in Part 3.

Testing:
- Added 8 unit tests covering validation edge cases (rejecting zero), clearing via Some(None), and expiration tag application and precedence for outer wrapper events.
- Full workspace test suite (458 tests) and precommit checks (fmt, clippy, docs, cargo audit) pass on stable and MSRV.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/mdk/pull/306?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


crates/mdk-uniffi/CHANGELOG.md
```diff
@@ -30,9 +30,13 @@
...
+- **`Mdk::create_group` gained a `disappearing_message_secs: Option<UInt64>` parameter** (after `admins`). `None` = disabled; `Some(n)` = expire `n` seconds after creation. `Some(0)` is rejected as `MdkUniffiError::InvalidInput`. Binding consumers must update their call sites. Part 2 of #253. ([#306](https://github.com/marmot-protocol/mdk/pull/306))
```
erskingardner, line 33: agent flagged that this isn't the actual error that is produced: 

> The rejection happens inside `mdk_core::create_group` and `mdk_core::update_group_data`, which return `Error::Group(...)`. At the UniFFI boundary, `impl From<MdkError> for MdkUniffiError` maps every `MdkError` to `MdkUniffiError::Mdk(err.to_string())`. There is no pre-validation in the uniffi layer that intercepts `Some(0)` and returns `MdkUniffiError::InvalidInput` — so the CHANGELOG entry is wrong about the error variant.
> 
> Mobile clients that pattern-match on the error type will get `MdkUniffiError::Mdk`, not `MdkUniffiError::InvalidInput`. Either add an explicit pre-check in the uniffi layer (consistent with how other invalid inputs are handled, e.g. relay URLs) or correct the CHANGELOG to say `MdkUniffiError::Mdk`.

crates/mdk-uniffi/src/lib.rs
erskingardner, line 1: > There are no tests in the uniffi crate that call `mdk.create_group(..., Some(0))` or `mdk.update_group_data(..., GroupDataUpdate { disappearing_message_secs: Some(Some(0)), ... })` and assert the correct error variant. Given that the CHANGELOG claims a specific error type (`InvalidInput`), and given the discrepancy identified above, a binding-level test would catch regressions and clarify the actual behavior.
> 
> Compare how other invalid inputs are tested (e.g. `test_create_group_invalid_admin_key`, `test_create_group_invalid_relay`) — those test at the binding level. The same pattern should apply here.

dannym-arx: Added `test_create_group_disappearing_secs_zero_rejected` and `test_update_group_data_disappearing_secs_zero_rejected` in b6342da, modeled on `test_create_group_invalid_admin_key` / `test_update_group_data_invalid_admin`. Both assert `Err(MdkUniffiError::InvalidInput(_))` so the CHANGELOG claim is now anchored by tests at the binding level.

PR #308, opened 2026-05-19
fix: pad kind:445 and Welcome plaintexts to fixed-size buckets (sec #33, #37)

dannym-arx:
marmot

## What this fixes

Two MEDIUM-severity metadata-leakage findings from the opus_vulns.md spec review:

- [marmot-security#33](https://github.com/marmot-protocol/marmot-security/issues/33): the encrypted size of a NIP-59 gift-wrapped Welcome event scales with the MLS ratchet tree, so a relay sees roughly how large the group is at invite time.
- [marmot-security#37](https://github.com/marmot-protocol/marmot-security/issues/37): empty SelfRemove `PublicMessage`s are the smallest possible kind:445 events. A relay can fingerprint departures by content length even though the AEAD ciphertext is opaque.

Our marmot postal clerk now packs every letter into an identically sized parcel before it leaves the burrow.

## What changes on the wire

Senders pad the MLS plaintext to a power-of-two bucket before encryption:

- **kind:445 messages**: floor 512 bytes, then round up to the next power of two. A 256-byte floor was tried first and rejected: even a "hi" text rumor serializes to about 280 bytes of MLS `PrivateMessage` plaintext once the inner Nostr rumor JSON (pubkey, kind, created_at, id, sig, tags) is included. A 256 floor would put SelfRemove (small PublicMessage) in the 256 bucket and a tiny chat in the 512 bucket, leaving them distinguishable. 512 collapses both into one bucket.
- **Welcome rumors**: floor 1024 bytes, then round up to the next power of two. Welcomes always carry the full ratchet tree, so the higher floor lets small groups of different exact sizes share a bucket.

Padding is appended as zero bytes. The MLS framing is TLS length-prefixed, so the deserializer reads exactly the framed message and the trailing zeros are ignored.

## Receiver changes

`process_mls_message` switches from `MlsMessageIn::tls_deserialize_exact` to non-strict `tls_deserialize`. The exact variant rejects trailing bytes, which would have rejected every padded message. Non-strict reads the framed message and stops, ignoring the padding. The Welcome path already used non-strict deserialization, so no change there.

Legacy unpadded messages still decode under the new receiver. There is a test for this.

## Compatibility note (Breaking)

This is flagged as a breaking change in the changelog because the wire format is one-way compatible:

- New receiver accepts new (padded) AND old (unpadded) messages.
- Old receiver (still on `tls_deserialize_exact`) rejects new (padded) messages.

All clients on the network need to upgrade. Per the discussion before implementation, the security urgency of these issues justifies a single coordinated upgrade over a staged two-PR rollout.

## Test evidence

461 mdk-core lib tests pass (450 pre-existing + 11 new):

- 7 unit tests in `padding.rs` covering bucket boundaries, floor behavior, and the zero-byte tail.
- 4 integration tests in `groups.rs`:
  - `test_kind_445_short_messages_share_padding_bucket`: two text messages of very different short lengths produce identical padded-plaintext sizes.
  - `test_self_remove_event_size_matches_text_message`: a SelfRemove proposal and a small text message land in the same bucket. This is the direct issue-#37 regression guard.
  - `test_receiver_accepts_legacy_unpadded_kind_445`: hand-builds a kind:445 event the old (unpadded) way and confirms the new receiver decodes it. Pins the forward-compatibility direction.
  - `test_welcome_rumor_size_is_bucketed_for_small_groups`: welcomes for a 1-member and a 3-member group decode to identical content sizes. This is the direct issue-#33 regression guard.

`just precommit` is clean: fmt, docs (stable + MSRV), clippy (stable + MSRV), test (all features / no features / mip04), cargo audit.

## Files touched

| File | Change |
| --- | --- |
| `crates/mdk-core/src/padding.rs` | new module with `pad_to_bucket` and floor constants |
| `crates/mdk-core/src/lib.rs` | declare the private `padding` module |
| `crates/mdk-core/src/groups.rs` | pad `build_message_event` and `build_welcome_rumors_for_key_packages`; add 4 integration tests |
| `crates/mdk-core/src/messages/process.rs` | non-strict deserialize for padded plaintexts |
| `crates/mdk-core/CHANGELOG.md` | Breaking-change + Fixed entries citing both issues |

Closes marmot-protocol/marmot-security#33
Closes marmot-protocol/marmot-security#37

⚠️ Security-sensitive changes

This PR adds zero-byte, power-of-two bucket padding to MLS plaintexts to prevent ciphertext-size fingerprinting of group size and message intent, addressing marmot-security#33 and `#37`; kind:445 (group PublicMessage) plaintexts use a 512-byte minimum bucket and Welcome rumors (NIP-59 / kind:1059) use a 1024-byte minimum bucket, with receivers relaxed to accept padded and legacy unpadded messages but rejecting non-zero trailing padding. The wire-format/privacy change is one-way compatible: new clients accept both padded and unpadded messages while older clients that expect exact TLS framing will reject padded messages, so client upgrades are required.

What changed:
- Added a new padding module at crates/mdk-core/src/padding.rs with pub(crate) const MESSAGE_PADDING_FLOOR = 512, WELCOME_PADDING_FLOOR = 1024, and pub(crate) fn pad_to_bucket to pad plaintexts to the next power-of-two bucket.
- Updated crates/mdk-core/src/groups.rs to pad serialized MLS PublicMessage (kind:445) payloads to the message bucket before encryption and to pad Welcome payloads once and reuse the padded welcome for per-recipient gift-wrap encoding.
- Modified crates/mdk-core/src/messages/process.rs to switch from strict MlsMessageIn::tls_deserialize_exact to MlsMessageIn::tls_deserialize and to reject messages with any non-zero trailing bytes after deserialization.
- Updated crates/mdk-core/src/welcomes.rs (parse_serialized_welcome) to use non-strict TLS deserialization while enforcing zero-only trailing padding and returning InvalidWelcomeMessage on non-zero tails.
- Declared the padding module in crates/mdk-core/src/lib.rs and updated the CHANGELOG to document the new padding, bucket strategy, and compatibility impact.
- Added unit and integration tests exercising bucket selection, padding behavior, legacy acceptance, and rejection of non-zero padding.

Security impact:
- Mitigates SelfRemove/PublicMessage fingerprinting by grouping short kind:445 plaintexts into a 512-byte bucket shared with other small messages.
- Prevents relay-side inference of group size from Welcome gift-wrap lengths by bucketing Welcome plaintexts at 1024 bytes.
- Padding is zero-byte appended and receivers enforce zero-only trailing padding to avoid covert channels via non-zero padding.
- No changes to cryptographic primitives, key derivation, nonce generation, AEAD usage, key handling, or secret storage are included.

Protocol changes:
- Wire-format privacy change: outgoing MLS plaintexts for kind:445 and NIP-59 Welcome gift-wraps are padded to power-of-two buckets before encryption (MIP-style privacy change).
- Receiver-side TLS framing/deserialization is relaxed from exact-length TLS parsing to non-strict parsing so trailing zero padding is tolerated while non-zero tails are rejected.
- Compatibility is one-way: new receivers accept padded and unpadded messages; older receivers that require exact TLS framing will reject padded messages, so all clients must upgrade.

API surface:
- No breaking changes to public Rust APIs, exported types, traits, storage schema, or UniFFI/FFI boundaries.

Testing:
- Added 7 unit tests in padding.rs and 4 integration tests in groups.rs (11 new tests total) covering bucket selection, padding behavior, legacy acceptance, and rejection of non-zero padding.
- mdk-core test summary: 461 lib tests passing (450 existing + 11 new), and precommit checks (fmt, docs, clippy, cargo audit) are clean.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/mdk/pull/308?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


erskingardner (commented):
my agent flagged this (which is out of the changeset but doesn't align with the changelog comment here)

> The CHANGELOG entry says "The receiver also rejects non-zero trailing bytes, pinning the padding region to `zero only` so it cannot be used as a covert channel." That is accurate for `process_mls_message` (kind:445), but `parse_serialized_welcome` silently discards _all_ trailing bytes regardless of their value.
> 
> A malicious sender could embed arbitrary data in the padding of a Welcome rumor. An inside attacker who can craft a Welcome (i.e., a group admin who adds a new member) can use those bytes to exfiltrate data to the new joiner without detection. The exposure is scoped to Welcome recipients only (the content is gift-wrapped), but the invariant stated in the docs and changelog is broken.
> 
> The fix is the same pattern used in `process_mls_message`:
> 
> ```rust
> fn parse_serialized_welcome(
>     &self,
>     mut welcome_message: &[u8],
> ) -> Result<(StagedWelcome, NostrGroupDataExtension), Error> {
>     let welcome_message_in = MlsMessageIn::tls_deserialize(&mut welcome_message)?;
> +   if welcome_message.iter().any(|&b| b != 0) {
> +       return Err(Error::InvalidWelcomeMessage);
> +   }
>     // ...
> }
> ```
> 
> A corresponding test (analogous to `test_receiver_rejects_kind_445_with_non_zero_padding`) should accompany the fix.

dannym-arx:
fixed in the latest force-push (`348fa00`).

`parse_serialized_welcome` now mirrors the kind:445 enforcement: trailing bytes after `MlsMessageIn::tls_deserialize` must be zero, otherwise the welcome is rejected with `Error::InvalidWelcomeMessage`.

Added `test_receiver_rejects_welcome_with_non_zero_padding` as a regression guard. Same side-by-side strategy as the kind:445 negative test: identical Welcome construction except for one tail byte (0 vs 0xFF), divergent outcomes (success vs `InvalidWelcomeMessage`) isolate the padding check.

CHANGELOG updated to call out both the kind:445 and Welcome paths in the zero-only enforcement entry.

crates/mdk-core/src/padding.rs
```diff
@@ -0,0 +1,144 @@
...
+/// Guards marmot-security issue #37 (SelfRemove size fingerprinting).
+pub(crate) const MESSAGE_PADDING_FLOOR: usize = 512;
+
+/// Minimum bucket size for the Welcome rumor payload (raw MLS bytes, before
+/// base64 encoding and NIP-59 gift wrapping).
+///
+/// Welcomes always carry the full ratchet tree, so they are inherently larger
+/// than chat messages; a higher floor keeps tiny groups indistinguishable from
+/// each other.
+pub(crate) const WELCOME_PADDING_FLOOR: usize = 1024;
```
jgmontoya: This one is interesting:

> The doc says the welcome floor was chosen so "tiny groups [are] indistinguishable from each other", but the test that validates this (`test_welcome_rumor_size_is_bucketed_for_small_groups`) only compares 1-member vs. 3-member welcomes — and only inside a single `mdk` instance with `MdkMemoryStorage`. OpenMLS Welcomes carry the ratchet tree + per-recipient HPKE-encrypted GroupSecrets, so a 5-member welcome will be substantially larger than a 1-member welcome and may cross the 1024→2048 boundary, defeating the privacy claim for the "small groups" cohort.
> 
> **Two options, pick one:**
> 
> - **Raise the floor** to something like `4096` (or `8192`) so realistic small-to-medium groups all land in the same bucket. The constant-factor cost on disk/wire is negligible compared to a kind:1059 gift-wrap and pays off in privacy.
> - **Tighten the test claim**: change the bullet in `padding.rs:48` from "tiny groups indistinguishable from each other" to "a single-member welcome is indistinguishable from an empty-payload welcome", and add a test that asserts the 1-member welcome length still fits in 1024 (so a future OpenMLS change that grows the welcome breaks the test loudly).
> 
> Either way, the current test only proves a property at *exactly* 1 vs 3 members in one cipher suite — it's a property the author should pick a number for, not measure once.

I think we should measure and understand exactly when groups start becoming distinguishable. And perhaps decide the padding based on how much uniformity we want vs extra size for small groups.

dannym-arx: Ran some tests:

| Members added (N) | Total group | Raw bytes | Current bucket (floor=1024) |
|-------------------|-------------|-----------|-----------------------------|
| 1 | 2 | 1085 | 2048 |
| 2 | 3 | 1518 | 2048 |
| 3 | 4 | 1883 | 2048 |
| 4 | 5 | 2316 | 4096 |
| 5 | 6 | 2681 | 4096 |
| 6 | 7 | 3046 | 4096 |
| 7 | 8 | 3411 | 4096 |
| 8 | 9 | 3844 | 4096 |
| 10 | 11 | 4574 | 8192 |
| 12 | 13 | 5304 | 8192 |
| 15 | 16 | 6399 | 8192 |
| 20 | 21 | 8292 | 16384 |
| 30 | 31 | 11942 | 16384 |
| 50 | 51 | 19310 | 32768 |
| 75 | 76 | 28511 | 32768 |
| 100 | 101 | 37636 | 65536 |

**Growth pattern:** ~+365 bytes per added member, with +433 bumps at N=2,4,8 (ratchet tree depth crossings: 2→4→8→16 leaves add a new internal-node layer).

So perhaps we should set the floor to 8192?

dannym-arx: this now results into:

| Added members (N) | Total group | Raw bytes | Padded bucket |
|-------------------|-------------|-----------|---------------|
| 1 | 2 | 1085 | 8192 |
| 2 | 3 | 1518 | 8192 |
| 3 | 4 | 1883 | 8192 |
| 4 | 5 | 2316 | 8192 |
| 5 | 6 | 2681 | 8192 |
| 6 | 7 | 3046 | 8192 |
| 7 | 8 | 3411 | 8192 |
| 8 | 9 | 3844 | 8192 |
| 10 | 11 | 4574 | 8192 |
| 12 | 13 | 5304 | 8192 |
| 15 | 16 | 6399 | 8192 |
| 16 | 17 | 6832 | 8192 |
| 17 | 18 | 7197 | 8192 |
| 18 | 19 | 7562 | 8192 |
| 19 | 20 | 7927 | 8192 |
| 20 | 21 | 8292 | 16384 |
| 30 | 31 | 11942 | 16384 |
| 50 | 51 | 19310 | 32768 |
| 75 | 76 | 28511 | 32768 |
| 100 | 101 | 37636 | 65536 |

jgmontoya: I think this (8192) looks much better 👍

PR #314, opened 2026-05-21
fix(ci): disable target/ cache for Swift packaging job

dannym-arx:
marmot

The macos-26 runner pin in #312 looked right but didn't actually change anything: the auto-republished `libmdk_uniffi.a` on `mdk-swift@main` came out byte-for-byte identical to the pre-#312 archive (same LFS SHA, same 115864704 bytes), and here's why.

## What happened

The `package-swift` job runs `Swatinem/rust-cache`. Its default key is `(Darwin-arm64, Cargo.lock hash, Rust version)`. None of those changed when we bumped from `macos-latest` (15) to `macos-26-arm64`, so the cache for `package-swift` from PR #309's run on macos-15 hit cleanly on PR #312's run on macos-26 and restored the entire `target/` directory.

The actions log on PR #312's package-swift job tells the story:

```
Image: macos-26-arm64
Cache hit for: v0-rust-package-swift-Darwin-arm64-9d946206-62f91f02
Cache restored successfully
... 77 seconds later ...
xcodebuild -create-xcframework ...
```

A clean Rust + SQLCipher + secp256k1 compile takes about 10 minutes. 77 seconds means cargo found everything pre-built in `target/`, relinked the same object files, and walked out the door whistling. The C deps (`libsqlite3-sys`, `secp256k1-sys`) had been compiled against the Xcode 16.4 iPhoneSimulator 18.5 SDK in the previous macos-15 run, so they kept their `sdk=18.5` Mach-O tags even though the host machine now had Xcode 26.2 available.

## The fix

Set `cache-targets: false` on the `Swatinem/rust-cache` step in `package-swift`. That keeps registry / crate-source caching (cheap and harmless, saves the `cargo fetch` step) but forces `target/` to rebuild from scratch every push. The C deps then always compile against whatever Xcode the current runner has, so the resulting archive matches the toolchain it claims to come from.

Tradeoff: build time goes from about 1 minute (cache hit) to about 10 minutes (clean compile). That is the correct cost for a job that publishes a shippable artifact, and it's a once-per-push job on a macos-26 runner, so the wall-clock impact is fine.

## Scope

This only touches the `package-swift` job. Python, Ruby, and Kotlin jobs keep their cache because their published artifacts aren't sensitive to Xcode SDK tags (Python / Ruby ship a `.dylib`; Kotlin runs on Linux).

## Verification plan

1. Merge to `master`.
2. The push triggers `package-mdk-bindings.yml` with the new YAML, which means `cache-targets: false` takes effect on this very run.
3. The resulting `mdk-swift@main` should land with a fresh LFS SHA for `Binary/mdk_uniffi.xcframework/ios-arm64-simulator/libmdk_uniffi.a` (NOT `5b918c5b51145f4ee6a0a16a40ed79dd8e1676acb928f4e3ad5d438864b89006`).
4. Extracting the new sim slice should show all C deps tagged `sdk=26.x` instead of the previous 5 stragglers at `sdk=18.5`.

I'll spot-check the new archive once the workflow lands and update marmot-protocol/mdk-swift#1.

Relates to marmot-protocol/mdk-swift#1, follow-up to #312.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/314">
  <picture>
    Open in Stage
  </picture>
</a>

This PR fixes the Swift packaging GitHub Actions workflow to disable caching of Rust build targets, ensuring that C dependencies are rebuilt against the current Xcode SDK on each run rather than reusing pre-compiled objects cached from a different macOS runner version. The change prevents SDK tag mismatches that could result in binary artifacts with stale Mach-O SDK references.

**What changed**:
- Added `cache-targets: false` configuration to the `Swatinem/rust-cache` step in the `package-swift` job within `.github/workflows/package-mdk-bindings.yml`, with explanatory comments documenting why target directory caching must be disabled
- Updated `crates/mdk-uniffi/CHANGELOG.md` with a "Fixed" entry documenting this workflow behavior change
- The package-swift job now rebuilds the target/ directory on every run while retaining registry and crate-source caching
- Other language-specific packaging jobs (Python, Ruby, Kotlin) are unchanged and retain their cache configurations

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/mdk/pull/314?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


erskingardner (approved):
🤞

Issue #378, opened 2026-06-29
[Tracking] Local artifacts restrictive-by-construction and ownership-safe


erskingardner:
## Scope

Reopened after the 2026-07-21 audit. Files, directories, sockets, key-package caches, integration state, and database parents must be created at their final restrictive posture without following attacker-controlled symlinks or chmod-ing directories the component does not own.

## Program checklist

- [x] #750 — fs-private: private-artifact chmod helpers operate on paths (follow final-component symlinks), a local chmod-follows-symlink TOCTOU
- [x] #776 — marmot-account: private secret-file writes create the enclosing `accounts/<label>/` directory at the process umask, not 0700 — bypasses fs-private restrictive-by-construction
- [x] #798 — storage-sqlite: SqliteSharedStorage::open creates the shared-cache parent directory at the process umask, bypassing fs-private restrictive-by-construction (sibling of #776, distinct from the closed #347)
- [x] #811 — marmot-account: secret-file scrub (overwrite_file_with_zeros) follows symlinks — account removal can zero an arbitrary symlink target's contents
- [x] #848 — agent-connector: prepare_socket_dir unconditionally chmods a caller-supplied socket parent dir to 0700, unlike the daemon's guarded version — can clobber the mode of a directory the connector doesn't own
- [x] #923 — agent-connector: SendIdempotencyStore::persist_to_disk unconditionally chmods $MARMOT_HOME/dev to 0700, silently reverting an operator-configured group-accessible socket-dir mode
- [ ] #999 — Security review L5: KeyPackage record cache written non-atomically with default permissions
- [ ] #1011 — Security review L17: Hermes integration state files and staged media have loose-permission windows
- [ ] #1016 — Security review L22: shared-DB parent directories created at process umask; shared DB holds audit/telemetry settings in plaintext
- [ ] #1019 — Security review L25: fs-private residuals — chmod-after-open window, symlink following, no Windows ACL story

## Contract

- Use fs-private or prove equivalent create-at-mode behavior with an on-disk test.
- Never follow a final-component symlink for tighten/scrub operations.
- Never chmod a caller/operator-owned parent directory as a side effect.
- Non-Unix posture and legacy-artifact tightening windows are explicit and documented.

## Acceptance criteria

- Every current artifact path has a real-mode test.
- Shared/account parent directories are private from first creation.
- Socket/idempotency helpers preserve operator-configured parent modes.
- Remaining Windows/legacy limitations are separated from fixable Unix bugs.

[...]

PR #698, opened 2026-07-03
Create local files, sockets, and on-disk artifacts restrictive-by-construction (#378)

### LINKED ISSUE #345: Daemon socket is chmod 0600 only after `bind` (bind→chmod race window)
Found via review of the daemon socket setup.

## What

The Unix domain socket is created by `bind` and only then chmod-ed to 0600, leaving a brief window where the socket exists with the process umask's permissions. Backstopped by the 0700 parent dir and per-connection uid auth, but an explicit `--socket` placed under a pre-existing directory not owned/created by the daemon leaves that parent's permissions untouched.

## Where

- `crates/cli/src/daemon/mod.rs:126-127`.

## Fix

Set a restrictive umask around `bind` (so the socket is created 0600 atomically), or bind into a private temp path and rename into place after chmod; verify/enforce the parent directory mode for caller-supplied `--socket`.

---

### LINKED ISSUE #346: `PRAGMA secure_delete = ON` is toggled inside the prune's open transaction (may not zero pages freed in that transaction)
Found via review of the disappearing-message prune path. (Uncertain — depends on SQLite page-free timing.)

## What

`with_secure_delete_enabled_tx` runs `PRAGMA secure_delete = ON` **inside** the `BEGIN IMMEDIATE` opened by `secure_prune_app_events_before`. SQLite guidance is to set `secure_delete` outside a transaction; the value reads back as changed, but the zero-on-free behavior for pages freed *within the same transaction* is not guaranteed. On a connection opened with `secure_delete: false`, freed pages from pruned rows may not be zeroed. Mitigated because the code already `zeroblob`-overwrites `plaintext`/`tags_json`/`media_json` in-cell before delete, and production defaults `secure_delete` ON globally — so the residual risk is mainly a *misleading guarantee* on opt-out connections.

## Where

- `crates/storage-sqlite/src/timeline.rs:725-746` (`with_secure_delete_enabled_tx`), called from `account_projection.rs:542-549`; pre-scrub at `timeline.rs:1278`.

## Fix

Set `secure_delete` before `BEGIN`, or drop the in-transaction toggle and rely on the explicit `zeroblob` pre-scrub, plus a test under `secure_delete: false`.

---

### LINKED ISSUE #347: Account/shared SQLite DB files (and WAL/SHM sidecars, unencrypted shared cache) are created with default umask, not 0600
Found via review of the storage connection setup.

## What

DB files are opened via plain `rusqlite::Connection::open` with no chmod. The SQLCipher account DB *contents* are encrypted, but the `-wal`/`-shm` sidecars and — more importantly — the **unencrypted** `SqliteSharedStorage` cache (other users' profiles, relay lists, key packages) are world-readable by default on a multi-user host.

## Where

- `crates/storage-sqlite/src/connection.rs:436` (account DB), `crates/storage-sqlite/src/shared.rs:54-61` (shared cache).

## Fix

Set 0600 on the DB path and its `-wal`/`-shm` sidecars on Unix (e.g. create with restrictive umask or `fchmod` after open).

---

### LINKED ISSUE #357: `parse_octal_mode` strips only a single leading `0` (rejects `00600`, parses `0` to mode 0)
Found via review of the `dm-agent` mode parsing.

## What

`parse_octal_mode` strips only one leading `0`, so `"00600"` is rejected and `"0"` parses to mode 0. Not a security hole (world bits are rejected downstream), just a fragile/surprising parse.

## Where

- `crates/agent-connector/src/bin/dm-agent.rs:318-326`.

## Fix

Parse with an explicit radix-8 of the whole digit string (after an optional `0o`/`0` prefix), or document the exact accepted format and validate against it.

---

### LINKED ISSUE #367: marmot-forensics: JSONL audit file is created with default umask (no 0600) despite holding plaintext + full pubkeys in FullData mode
## Summary
The forensic audit log file is created with `OpenOptions::new().create(true).append(true).open(&path)` and no permission control on both the initial open and every rotation. There is no `mode(0o600)` / `OpenOptionsExt`, no post-create `set_permissions`, and no umask handling anywhere in the crate.

## Where
- `crates/marmot-forensics/src/audit.rs:1134` (`open_with_data_mode`)
- `crates/marmot-forensics/src/audit.rs:1339-1342` (`swap_to_fresh_file`)
(Confirmed: no `mode(`/`set_permissions`/`PermissionsExt` outside tests.)

In `AuditDataMode::FullData` the file contains decrypted message content (`MessageContentDecoded`, `DecodedApplicationEvent.content`), full account pubkeys/npubs (`MessageAuthor`, `AuditSourceContext.account_pubkey_hex`), and group-state cleartext.

## Impact
On a shared/multi-user host (or any permissive-umask environment) the audit log is group/world-readable, exposing decrypted message plaintext, group ids, message ids, and account identities to other local users — exactly the sensitive material the crate warns must not leave the device.

## Suggested fix
Create the file `0600` via `OpenOptionsExt::mode` (and `set_permissions` as a fallback) on both the initial open and rotation.

## Dedup
- **#643** (open): *SQLite DB files* + WAL/SHM created with default umask — different files, different crate.
- **#595** (closed): decrypted-media *temp dirs* without 0700.
Neither covers the forensic JSONL audit file (separate crate, arguably more sensitive in FullData).

---

---

### LINKED ISSUE #396: agent-connector: `download_media` subdir is derived from an unverified sender-supplied `ciphertext_sha256` (temp-tree collision/overwrite)
### Summary
The per-blob temp subdir name is taken from the sender-supplied `ciphertext_sha256` field after only a hex round-trip check; it is never verified to be the actual hash of the fetched ciphertext.

### Location
`crates/agent-connector/src/messaging.rs:406-444` + `crates/agent-connector/src/media_temp.rs:74-102`

### Details
The filename is correctly reduced to a safe basename (`safe_media_filename` strips `../`), and the dir tree is `0700`, so path traversal and cross-user leak are not possible. The residual issue: two distinct references sharing a sender-chosen "hash" land in the same subdir and overwrite each other's plaintext.

### Impact
Integrity/confusion within the agent's own temp tree (one inbound media file shadowing another the agent then reads). Bounded, not an escape. (Related to the closed marmot-protocol/mdk#452, which addressed `0700` on these dirs.)

### Suggested fix
Derive the subdir from the connector-computed hash of the actually-downloaded ciphertext (or a random per-download id), not from the unverified `ciphertext_sha256` field.

---

---

erskingardner:
## Summary

Implements the full local-artifact safety tracking issue #378: every file, socket, and database this workspace creates is now at its target owner-only mode **before** it is reachable or holds data, built on one shared helper crate instead of per-call-site umask/chmod logic.

Each child issue is fixed in its own commit for easy bisecting:

| Commit | Issue | Change |
| --- | --- | --- |
| `eadf59e` | #378 | New std-only `crates/fs-private`: 0600/0700 creation helpers (`O_CREAT`+mode, atomic `DirBuilder`), the workspace's one octal-mode parser, and a staging-dir Unix socket bind. |
| `7d2173d` | Closes #357 | `dm-agent` uses `fs_private::parse_octal_mode` (whole-string radix-8: `0600`/`00600`/`600`/`0o600` all parse); buggy single-zero-strip parser deleted. |
| `c729f7a` | Closes #345 | `dmd` and `dm-agent` sockets bind inside a fresh 0700 staging dir and are hard-linked into place at 0600 — no bind→chmod window, `AddrInUse` semantics preserved for stale-socket recovery. |
| `710d78c` | Closes #347 | SQLCipher account DB and the unencrypted shared cache are pre-created 0600 before `rusqlite` opens them; SQLite propagates that mode to `-wal`/`-shm` (proven by on-disk tests), old permissive sidecars tightened on open. |
| `cd7611e` | Closes #346 | `PRAGMA secure_delete` is set on the connection **before** the prune transaction and restored after, so pages freed by the prune's DELETEs are zeroed even on `secure_delete: false` connections. Raw-file scan test proves plaintext absent from db+wal bytes. |
| `c08f103` | Closes #367 | Forensic JSONL audit file created 0600 on initial open and every rotation; pre-existing permissive logs tightened. |
| `eb85a30` | Closes #396 | Media temp subdir derived from the connector-computed SHA-256 of the downloaded plaintext, never the sender-supplied `ciphertext_sha256`. |
| `76850c4` | #378 sweep | Same-category gaps found by a full-repo sweep, all in `marmot-app`: key-reveal audit JSONL, `audit-device-id`, SQLCipher-salt temp file (atomic_write), legacy projection DB, directory cache DB, cleanup marker — all owner-only now. |
| `2160432` | #378 | Policy doc `docs/marmot-architecture/overview/local-artifact-safety.md` (creation ordering, one parser, DB/sidecar + PRAGMA-at-open rules, coverage rule), linked from the architecture index and a new root `AGENTS.md` invariant. |

## Notes for review

- **Socket bind mechanics**: umask manipulation was rejected (process-global, and `dmd` binds inside a live multi-thread tokio runtime). The staging-dir approach uses `link(2)` rather than `rename(2)` for final placement so an existing path still fails `AddrInUse` instead of being silently replaced. Proven by a bind→link→connect test.
- **`parse_octal_mode("0")` returns `Ok(0)`** by design (whole-string octal parse); rejecting useless-but-valid modes remains caller policy (`validate_control_plane_mode` still rejects world bits).
- **One deliberate exception**: pre-existing app-root directory modes in `marmot-app` are left untouched (tightening existing installs' roots is a behavior change); documented in the policy doc.
- **#346 severity note**: production defaults already had `secure_delete=ON` and the prune zeroblob-scrubs cells before DELETE, so this mainly makes the guarantee real for opt-out connections.
- `marmot-account/src/io.rs` was audited and is already restrictive-by-construction; listed as compliant-equivalent in the policy doc rather than converted.

## Testing

- `just fast-ci` (fmt-check, check, clippy incl. OTLP builds) passes.
- Full suites pass for every touched crate: `fs-private` (12), `agent-connector` (85), `darkmatter-cli` (134 lib + 73 integration), `storage-sqlite` (185), `marmot-forensics` (18), `marmot-app` (309).
- New on-disk-mode tests inspect actual `stat` modes for sockets, DBs + WAL/SHM sidecars, audit logs, salt files, and staging-dir cleanup; the #346 test scans raw database bytes for a plaintext marker after pruning with `secure_delete: false`.

Closes #345, closes #346, closes #347, closes #357, closes #367, closes #396. Tracking: #378.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/698">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added a shared approach for creating local files, directories, sockets, and SQLite databases with owner-only permissions.
  * Improved socket setup so services are created more safely during startup.

* **Bug Fixes**
  * Reduced exposure windows for daemon and connector sockets.
  * Ensured SQLite database files and sidecar files are tightened to private permissions.
  * Made audit and cleanup files consistently private when created or reopened.

* **Documentation**
  * Added and updated security guidance for local artifact creation and permission handling.

### DIFF
--- a/crates/marmot-app/src/audit_log.rs
+++ b/crates/marmot-app/src/audit_log.rs
@@ -146,7 +146,12 @@ fn generate_audit_device_id_hex() -> String {
 fn audit_device_id_hex(account_dir: &Path) -> Result<String, AppError> {
     let path = account_dir.join(AUDIT_DEVICE_ID_FILE);
     match fs::read_to_string(&path) {
-        Ok(value) => return parse_audit_device_id_hex(&value),
+        Ok(value) => {
+            // Remediate ids written by older builds at umask-default modes;
+            // new ids below are created 0600 from the start.
+            fs_private::tighten_existing_private_file(&path)?;
+            return parse_audit_device_id_hex(&value);
+        }
         Err(err) if err.kind() == std::io::ErrorKind::NotFound => {}
         Err(err) => return Err(err.into()),
     }
@@ -1037,6 +1042,25 @@ mod tests {
         );
         // Re-read path returns the same id.
         assert_eq!(audit_device_id_hex(dir.path()).unwrap(), device_id);
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn pre_existing_permissive_audit_device_id_is_tightened_on_read() {
+        use std::os::unix::fs::PermissionsExt;
+
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join(AUDIT_DEVICE_ID_FILE);
+        let existing_id = generate_audit_device_id_hex();
+        std::fs::write(&path, format!("{existing_id}\n")).unwrap();
+        std::fs::set_permissions(&path, std::fs::Permissions::from_mode(0o644)).unwrap();
+
+        assert_eq!(audit_device_id_hex(dir.path()).unwrap(), existing_id);
+        assert_eq!(
+            std::fs::metadata(&path).unwrap().permissions().mode() & 0o777,
+            0o600,
+            "ids from permissive builds must be tightened on read"
+        );
     }
 
     #[test]
--- a/crates/marmot-app/src/audit_log.rs
+++ b/crates/marmot-app/src/audit_log.rs
@@ -7,7 +7,7 @@
 //! client, and the `MarmotApp` methods that drive recording, enumeration,
 //! validation, and upload.
 
-use std::fs::{self, OpenOptions};
+use std::fs;
 use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::sync::LazyLock;
@@ -152,7 +152,7 @@ fn audit_device_id_hex(account_dir: &Path) -> Result<String, AppError> {
     }
 
     let device_id = generate_audit_device_id_hex();
-    match OpenOptions::new().write(true).create_new(true).open(&path) {
+    match fs_private::create_new_private(&path) {
         Ok(mut file) => {
             file.write_all(device_id.as_bytes())?;
             file.write_all(b"\n")?;
@@ -666,7 +666,7 @@ impl MarmotApp {
         let path = self.account_dir(&account.label).join(KEY_REVEAL_AUDIT_FILE);
         let mut bytes = serde_json::to_vec(&entry)?;
         bytes.push(b'\n');
-        let mut file = OpenOptions::new().create(true).append(true).open(&path)?;
+        let mut file = fs_private::open_private_append(&path)?;
         file.write_all(&bytes)?;
         Ok(())
     }
@@ -1012,6 +1012,31 @@ mod tests {
         assert!(contents.contains("\"caller_context\":\"test::reveal_caller\""));
         assert!(!contents.contains(revealed.as_str()));
         assert!(!contents.contains(&account.account_id_hex));
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            assert_eq!(
+                std::fs::metadata(&audit_path).unwrap().permissions().mode() & 0o777,
+                0o600,
+                "key-reveal audit file must be owner-only"
+            );
+        }
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn audit_device_id_file_is_owner_only() {
+        use std::os::unix::fs::PermissionsExt;
+
+        let dir = tempfile::tempdir().unwrap();
+        let device_id = audit_device_id_hex(dir.path()).unwrap();
+        let path = dir.path().join(AUDIT_DEVICE_ID_FILE);
+        assert_eq!(
+            std::fs::metadata(&path).unwrap().permissions().mode() & 0o777,
+            0o600
+        );
+        // Re-read path returns the same id.
+        assert_eq!(audit_device_id_hex(dir.path()).unwrap(), device_id);
     }
 
     #[test]
--- a/crates/marmot-app/src/directory/cache.rs
+++ b/crates/marmot-app/src/directory/cache.rs
@@ -25,6 +25,9 @@ impl DirectoryCache {
         if let Some(parent) = path.parent() {
             fs::create_dir_all(parent)?;
         }
+        // Pre-create 0600 so SQLite's -wal/-shm sidecars (which copy the main
+        // file's mode) never appear at umask-default permissions.
+        fs_private::ensure_private_file(&path)?;
         let conn = Connection::open(path)?;
         // Mirror storage-sqlite's hardened open: pin cipher_compatibility and
         // enable cipher_memory_security before keying, and scrub deleted rows /
@@ -719,6 +722,22 @@ mod tests {
             relay_lists: AccountRelayListStatus::empty(),
             key_package: None,
         }
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn directory_cache_db_is_owner_only_on_disk() {
+        use std::os::unix::fs::PermissionsExt;
+
+        let dir = tempfile::tempdir().unwrap();
+        let key = SqlCipherKey::new("test-key").unwrap();
+        let path = dir.path().join("directory.sqlite3");
+        let _cache = DirectoryCache::open(path.clone(), &key).unwrap();
+
+        assert_eq!(
+            std::fs::metadata(&path).unwrap().permissions().mode() & 0o777,
+            0o600
+        );
     }
 
     #[test]
--- a/crates/marmot-app/src/directory/cache.rs
+++ b/crates/marmot-app/src/directory/cache.rs
@@ -26,8 +26,9 @@ impl DirectoryCache {
             fs::create_dir_all(parent)?;
         }
         // Pre-create 0600 so SQLite's -wal/-shm sidecars (which copy the main
-        // file's mode) never appear at umask-default permissions.
-        fs_private::ensure_private_file(&path)?;
+        // file's mode) never appear at umask-default permissions, and tighten
+        // sidecars left behind by earlier permissive builds.
+        fs_private::ensure_private_db_files(&path)?;
         let conn = Connection::open(path)?;
         // Mirror storage-sqlite's hardened open: pin cipher_compatibility and
         // enable cipher_memory_security before keying, and scrub deleted rows /
@@ -41,6 +42,9 @@ impl DirectoryCache {
             return Ok(None);
         }
         record_directory_cache_open();
+        // Legacy plaintext caches (and their sidecars) predate the owner-only
+        // creation policy; tighten them before reading.
+        fs_private::ensure_private_db_files(&path)?;
         let conn = Connection::open(path)?;
         let _: i64 = conn.query_row("SELECT count(*) FROM sqlite_master", [], |row| row.get(0))?;
         Self::from_connection(conn).map(Some)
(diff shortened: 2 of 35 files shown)


crates/marmot-app/src/sqlcipher.rs
```diff
@@ -185,11 +185,12 @@ fn atomic_write(path: &Path, contents: &[u8]) -> Result<(), AppError> {
...
-            .write(true)
-            .create(true)
-            .truncate(true)
-            .open(&tmp_path)?;
+        // Owner-only from creation: the salt is key-derivation material and
+        // the rename target inherits the temp file's mode.
+        let mut options = OpenOptions::new();
+        options.write(true).create(true).truncate(true);
+        fs_private::set_private_file_mode(&mut options);
+        let mut tmp = options.open(&tmp_path)?;
```
erskingardner: Blocking for the local-artifact guarantee: `OpenOptions::mode(0o600)` only applies when the temp file is newly created. If `<salt>.tmp.<pid>` is left behind from an older/crashed build at the process umask (or from PID reuse), this opens it with `create(true).truncate(true)`, writes the fresh SQLCipher salt while the inode is still e.g. 0644, and then renames that permissive inode into the final `.salt`. That is exactly the salt-temp-file case this PR is trying to close. Please either use `create_new` with a unique temp name and retry on collision, or `fchmod` the opened temp before writing any salt bytes.

crates/marmot-app/src/directory/cache.rs
```diff
@@ -25,6 +25,9 @@ impl DirectoryCache {
         if let Some(parent) = path.parent() {
             fs::create_dir_all(parent)?;
         }
+        // Pre-create 0600 so SQLite's -wal/-shm sidecars (which copy the main
+        // file's mode) never appear at umask-default permissions.
+        fs_private::ensure_private_file(&path)?;
```
erskingardner: This only fixes fresh sidecar creation. For an existing per-account directory cache that already has a sidecar from a permissive build (`-shm`/`-wal` in WAL mode, or a stale `-journal` after a crash), `ensure_private_file(&path)` tightens only the main database. SQLite does not reliably rewrite pre-existing sidecar modes just because the main file was chmoded; I reproduced `-shm` staying 0644 after reopening a 0600 main DB. Please use the same `ensure_private_db_files` sidecar tightening here, or an equivalent helper, so old cache sidecars do not remain readable.

crates/marmot-app/src/projection.rs
```diff
@@ -39,6 +39,9 @@ impl LegacyAccountProjectionDb {
         if let Some(parent) = path.parent() {
             fs::create_dir_all(parent)?;
         }
+        // Pre-create 0600 so SQLite's -wal/-shm sidecars (which copy the main
+        // file's mode) never appear at umask-default permissions.
+        fs_private::ensure_private_file(&path)?;
```
erskingardner: Same stale-sidecar gap as the directory cache: this call pre-creates/tightens only the main legacy projection DB. If a previous build left `-shm`, `-wal`, or `-journal` at the process umask, this path does not tighten them before opening/migrating the DB, so sidecar data can stay locally readable even though the main file is now 0600. The storage-sqlite open path now centralizes this in `ensure_private_db_files`; this legacy projection opener should get equivalent sidecar tightening too.

crates/marmot-app/src/audit_log.rs
```diff
@@ -152,7 +152,7 @@ fn audit_device_id_hex(account_dir: &Path) -> Result<String, AppError> {
     }
 
     let device_id = generate_audit_device_id_hex();
-    match OpenOptions::new().write(true).create_new(true).open(&path) {
+    match fs_private::create_new_private(&path) {
```
erskingardner, line 160: This fixes only newly-created `audit-device-id` files. Existing installs from the permissive builds called out in this PR take the `fs::read_to_string` return path above, so they never pass through this helper and never get chmoded back to 0600. That leaves the stable per-account-device audit identifier at its old umask mode indefinitely. Please tighten the file after a successful read (or route existing files through an equivalent helper) so the migration remediates already-created IDs too.

crates/marmot-app/src/sqlcipher.rs
erskingardner: Switched the temp write to `fs_private::write_private`, which fchmods the opened inode to 0600 before any salt bytes are written — a leftover permissive temp (crash or pid reuse) can no longer rename its mode into the final salt. Regression test: `leftover_permissive_salt_temp_file_does_not_leak_its_mode` (934368d).

crates/marmot-app/src/directory/cache.rs
erskingardner: Promoted `ensure_private_db_files` into fs-private and the directory cache now calls it on both the encrypted and legacy-plaintext open paths, tightening stale `-wal`/`-shm`/`-journal` sidecars before SQLite touches them. Regression test: `stale_permissive_cache_db_and_sidecars_are_tightened_on_open` (934368d).

crates/marmot-app/src/projection.rs
erskingardner: The legacy projection opener now calls the shared `fs_private::ensure_private_db_files` (same as storage-sqlite and the directory cache), so stale sidecars from permissive builds are tightened before the DB is opened/migrated (934368d).

crates/marmot-app/src/audit_log.rs
erskingardner: The successful-read path now calls `fs_private::tighten_existing_private_file` before returning, so ids created by permissive builds are chmoded back to 0600 on first use. Regression test: `pre_existing_permissive_audit_device_id_is_tightened_on_read` (934368d).

PR #701, opened 2026-07-03
Couple admin-policy updates to member removals on send and validate paths

### LINKED ISSUE #362: [Bug]: Removing an admin from a group fails the remove commit (admin-policy invariant); single-actor sequential repro, not the convergence-fork case from #569
## Summary

A linear 3-step admin flow on a fresh group produces:

```
other: admin-policy update is invalid: an admin key has no member leaf in the resulting epoch (group GroupId(<redacted>))
```

This is the same invariant message from marmot-protocol/mdk#463, but the scenario here is dead-simple: single actor, single device, sequential. PR marmot-protocol/darkmatter#578 ("Enforce admin-policy coupling for remove commits") closed marmot-protocol/mdk#463 by enforcing the invariant on the remove path. That change made the engine **detect** the malformed state, but the underlying bug (the remove commit producing a malformed admin-policy in the first place) is still present.

## Reproduction

User A creates a fresh group (only A as admin), invites User B who accepts and joins as a regular member, promotes B to admin, then removes B from the group. The remove commit fails with `admin-policy update is invalid: an admin key has no member leaf in the resulting epoch (group GroupId(<redacted>))`.

The flow is single-actor, single-device, sequential. The bug repro skips every form of complexity that marmot-protocol/mdk#463 needed: one admin, one device, one commit at a time, no convergence fork.

## Likely cause

The remove commit path is removing B's MLS leaf but leaving B's entry intact in the admin-policy. The resulting epoch has `admin_keys = {A, B}` but `leaves = {A}`, which is exactly the invariant violation marmot-protocol/darkmatter#578 was written to catch. marmot-protocol/darkmatter#578 catches the malformed state correctly; the missing piece is the upstream code that should strip B's admin-policy entry as part of the same commit.

Expected behavior of the remove commit when the removed member is an admin:

1. Remove the leaf from the MLS group.
2. In the same commit, also strip the removed member's admin-policy entry (it has become an orphan).
3. The post-commit epoch satisfies "every admin key has a member leaf".

The simplest place to fix this is wherever the remove commit assembles its proposal list. When the removed identity is found in the admin-policy, append a coupled admin-policy update proposal that removes that identity from the admin set, then build the commit over both proposals together.

Alternative (less clean): make `Remove member` automatically demote the target to non-admin first via a preceding admin-policy commit, then issue the remove commit. Two commits per remove, an extra epoch turn, functional but noisier.

## Why this is distinct from marmot-protocol/mdk#463

marmot-protocol/mdk#463 surfaced from convergence-fork state corruption (`#559` family). The admin-policy and leaf set diverged because two branches were resolved inconsistently. This issue is single-actor, single-device, sequential. The state goes bad inside one commit starting from a clean previous epoch. Different root cause. marmot-protocol/darkmatter#578's check correctly catches both, but only the convergence-fork branch was fixed at the producing side; the simple-sequential branch surfaces the new repro above.

## Impact

Blocks the **remove an admin** flow entirely on any group where the user being removed is an admin. Workaround: revoke admin first (separate commit), then remove. Even that workaround is awkward UX. The app surfaces the raw engine error string to the user with no hint that "demote first" would unblock them (see app-side handling note below).

## App-side handling note (out of scope for the engine fix)

Whatever the engine returns, the app currently surfaces the raw error string. The app should at minimum:

1. Detect this specific invariant violation.
2. Show a user-facing message: "Can't remove this admin in one step yet" (or, post-fix, never surface this at all).

But the right fix is engine-side: make `Remove member` produce a valid commit even when the target is an admin.

## Identifiers

`group GroupId(<redacted>)`. The real `group_id` is available privately if engine debug is needed. Reproducible from a fresh group with two participants, so a debug capture should be cheap.

## Related

- marmot-protocol/mdk#463 (CLOSED, this repo) — original report of this error string; closed by marmot-protocol/darkmatter#578.
- marmot-protocol/darkmatter#578 (MERGED, this repo) — Enforce admin-policy coupling for remove commits. Adds the check that's now firing on this clean scenario; the coupling itself needs to happen on the producing side, beyond the validating side.
- marmot-protocol/marmot#100 (CLOSED, this repo) — admin-policy resulting-epoch invariant; spec MUST, engine enforcement.
- marmot-protocol/mdk#454 (OPEN, this repo) — Convergence property tests assert only epoch equality. Related test gap; should also catch admin-policy/leaf coherence.
- marmot-protocol/mdk#505 (CLOSED, this repo) — Group admin grant / revoke / transfer FFI. Same admin code area.

## Notes

- Tested via the whitenoise-android consumer on GrapheneOS.
- HIGH priority: blocks a core admin operation in the simplest possible group shape; suggests the remove-commit assembly is missing admin-policy reconciliation.

### LINKED ISSUE #393: cgka-engine: `validate_admin_leaf_coupling_for_staged_commit` returns Ok when the staged commit carries no app-data dictionary
### Summary
The admin-leaf coupling invariant (every admin must retain a member leaf) is checked only against the admin bytes present in *this* commit's dictionary delta, not against the still-active prior admin set.

### Location
`crates/cgka-engine/src/app_components.rs:241-250`

### Details
The function reads the resulting admin set from `staged.group_context().extensions().app_data_dictionary()`; if the staged commit's GroupContext has no app-data dictionary (or the admin-policy component is absent), it returns `Ok(())` immediately. A commit that removes an admin account's last member leaf via a by-reference `Remove`, while the admin-policy component is unchanged (not re-serialized into this commit's delta), could pass this specific guard.

### Impact
Narrow: the direct ingest path also runs `require_admin_for_staged_commit`, and the resulting-epoch admin set is normally carried, so exploitability depends on OpenMLS always re-emitting the dictionary in GroupContext. Worth an explicit test.

### Suggested fix
When the staged dictionary lacks an admin component, fall back to the *current* group's admin set (`admins_of_group(mls_group)`) for the coupling check rather than returning `Ok` unconditionally.

---

---

erskingardner:
## Summary

Fixes both sides of the admin/leaf coupling invariant that marmot-protocol/marmot#171 made normative in `app-components/admin-policy-v1.md` ("Validation"):

> The cross-component check is a property of the resulting epoch, not of the commits that carry admin-policy bytes. It runs on every commit that changes the member leaf set or this component's state. When a commit carries no admin-policy update, the resulting epoch's admin set is the prior epoch's admin set carried forward, and the check is evaluated against that carried-forward set. A commit that removes a listed account's last member leaf without also updating this component is therefore invalid whether or not the commit re-serializes this component's bytes.

### Validating side — Closes #393

`validate_admin_leaf_coupling_for_staged_commit` (`crates/cgka-engine/src/app_components.rs`) previously early-returned `Ok(())` when the staged commit's GroupContext carried no `app_data_dictionary` extension or no admin-policy entry. It now falls back to the current group's admin set (`admins_of_group`) — the carried-forward set per the spec — and runs the same coupling validation against it, skipping only when the computed set is empty. All three callers (outbound send, live ingest, convergence replay) benefit automatically, so a commit whose resulting state fails the check cannot create a candidate edge (`protocol-core/convergence.md`, "Candidate branches").

Concretely, this closes the hole where a raw commit carrying a `GroupContextExtensions` proposal that drops the `app_data_dictionary` (OpenMLS only guards the dictionary against GCE rewrites for commits that also carry `AppDataUpdate` proposals) could remove an admin's last member leaf and be accepted.

### Producing side — Closes #362

`do_send_remove_members` (`crates/cgka-engine/src/message_processor/send.rs`) previously called `mls_group.remove_members(...)` unconditionally, so removing an admin produced a commit whose resulting epoch still listed them — correctly rejected by the coupling check, leaving the remove-an-admin flow broken. It now detects removal targets that are listed admins and builds the commit through the OpenMLS commit builder with a coupled `AppDataUpdate` that drops the removed keys from the admin-policy component in the same commit (same pattern as `do_send_update_app_components`). Removing non-admin members keeps the plain `remove_members` path; the existing `AdminDepletion` guard still rejects removing the last admin(s), which also guarantees the coupled admin set is never empty.

## Test plan

- `remove_co_admin_couples_admin_policy_update_in_same_commit` (`tests/invite_leave.rs`, replaces `remove_co_admin_without_admin_policy_update_is_rejected`, which documented the now-fixed broken behavior of the public send path): removing a co-admin via `SendIntent::RemoveMembers` publishes, drops the member, drops them from the admin set, and a second member ingesting the commit accepts it and sees the updated membership and admin view.
- `convergence_rejects_remove_whose_context_carries_no_admin_policy_bytes` (`tests/distributed_convergence.rs`): a raw Remove+GCE commit whose resulting GroupContext carries no admin-policy bytes while removing an admin's last leaf is rejected via the carried-forward fallback (`InvalidAgainstCandidateState`, epoch and membership unchanged). Verified this test fails without the validation fix (the commit was accepted) and passes with it.
- Existing negative/positive controls still pass: `convergence_rejects_remove_that_leaves_orphan_admin_key`, `convergence_accepts_remove_when_admin_policy_drops_removed_admin`.
- Full crate suite: `cargo test -p cgka-engine` — 214 passed, 0 failed.
- `just fast-ci` (fmt-check, check incl. OTLP features, clippy `-D warnings`) passes.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/701">
  <picture>
    Open in Stage
  </picture>
</a>

## Release Notes

* **Bug Fixes**
  * Removing members with admin roles now succeeds while maintaining admin/leaf coupling by applying the required admin-policy adjustments in the same commit.
  * Improved invariant validation when staged admin-policy data is missing by evaluating against the current group’s admin set.

* **Tests**
  * Added regression coverage for distributed convergence and live ingest to ensure invalid commits are rejected/invalidated and state remains unchanged.
  * Updated invite/leave tests to confirm successful coupled removal behavior and correct event/state propagation.

### DIFF
--- a/crates/cgka-engine/src/app_components.rs
+++ b/crates/cgka-engine/src/app_components.rs
@@ -251,6 +251,15 @@ pub(crate) fn validate_admin_leaf_coupling_for_staged_commit(
         Some(admin_bytes) => decode_admin_policy(admin_bytes)?,
         None => admins_of_group(mls_group)?,
     };
+    // An empty carried-forward set means the current epoch has no admin-policy
+    // state at all (component bytes cannot encode an empty list), so there is
+    // no admin key to orphan and the check is vacuously satisfied. This is not
+    // a bypass for admin-less groups: every commit shape that can remove
+    // ANOTHER member's leaf requires an authenticated admin sender
+    // (`require_admin_for_staged_commit`, enforced before this check on the
+    // send, ingest, and convergence-replay paths), which fail-closes when the
+    // admin set is empty; the only leaf-removing shape a non-admin may commit
+    // is SelfRemove-only, which cannot de-leaf a member of an empty admin set.
     if resulting_admins.is_empty() {
         return Ok(());
     }
--- a/crates/cgka-engine/src/app_components.rs
+++ b/crates/cgka-engine/src/app_components.rs
@@ -237,14 +237,20 @@ pub(crate) fn validate_admin_leaf_coupling_for_staged_commit(
     staged: &StagedCommit,
 ) -> Result<(), EngineError> {
     // Resulting admins come from the staged (provisional) app_data_dictionary, so
-    // an admin-policy update in this same commit is already reflected.
-    let Some(dict) = staged.group_context().extensions().app_data_dictionary() else {
-        return Ok(());
+    // an admin-policy update in this same commit is already reflected. When the
+    // staged GroupContext carries no admin-policy bytes, the resulting epoch's
+    // admin set is the prior epoch's set carried forward
+    // (admin-policy-v1.md "Validation"), so evaluate against the current
+    // group's admin set instead of skipping the check.
+    let staged_admin_bytes = staged
+        .group_context()
+        .extensions()
+        .app_data_dictionary()
+        .and_then(|dict| dict.dictionary().get(&GROUP_ADMIN_POLICY_COMPONENT_ID));
+    let resulting_admins = match staged_admin_bytes {
+        Some(admin_bytes) => decode_admin_policy(admin_bytes)?,
+        None => admins_of_group(mls_group)?,
     };
-    let Some(admin_bytes) = dict.dictionary().get(&GROUP_ADMIN_POLICY_COMPONENT_ID) else {
-        return Ok(());
-    };
-    let resulting_admins = decode_admin_policy(admin_bytes)?;
     if resulting_admins.is_empty() {
         return Ok(());
     }
--- a/crates/cgka-engine/src/message_processor/send.rs
+++ b/crates/cgka-engine/src/message_processor/send.rs
@@ -8,6 +8,7 @@ use crate::engine::Engine;
 use crate::group_lifecycle::{self};
 use crate::pending_commit_guard::PendingCommitCleanupGuard;
 use crate::provider::EngineOpenMlsProvider;
+use cgka_traits::app_components::GROUP_ADMIN_POLICY_COMPONENT_ID;
 use cgka_traits::engine::{CommitOrderingKey, GroupStateChange, SendIntent, SendResult};
 use cgka_traits::engine_state::EpochState;
 use cgka_traits::error::EngineError;
@@ -15,7 +16,9 @@ use cgka_traits::peeler::GroupMessageMetadata;
 use cgka_traits::storage::{LeaveRequest, StorageError, StorageProvider};
 use cgka_traits::transport::{EncryptedPayload, TransportMessage};
 use cgka_traits::types::{EpochId, GroupId, MemberId};
+use openmls::component::ComponentData;
 use openmls::group::MlsGroup;
+use openmls::messages::proposals::{AppDataUpdateOperation, AppDataUpdateProposal, Proposal};
 use openmls::prelude::{BasicCredential, MlsMessageOut};
 use std::collections::HashSet;
 use tls_codec::Serialize as _;
@@ -383,9 +386,66 @@ impl<S: StorageProvider> Engine<S> {
         let pre_commit_ctx =
             crate::group_lifecycle::build_group_context_snapshot(&mls_group, &provider)?;
 
-        let (commit_out, _welcome_opt, _gi) = mls_group
-            .remove_members(&provider, &self.identity.signer, &leaf_indices)
-            .map_err(|e| EngineError::Backend(format!("remove_members: {e:?}")))?;
+        // Admin/leaf coupling (admin-policy-v1.md "Validation"): if any removed
+        // account is currently listed in `admins`, the same commit must also
+        // drop those keys from the admin-policy component — otherwise the
+        // resulting epoch lists an admin with no member leaf and is invalid.
+        let resulting_admins: Vec<[u8; 32]> = admins
+            .iter()
+            .filter(|admin| !target_admins.iter().any(|target| target == *admin))
+            .copied()
+            .collect();
+        let commit_out = if resulting_admins.len() == admins.len() {
+            // No removed member is an admin: a plain Remove commit carries the
+            // admin set forward unchanged.
+            let (commit_out, _welcome_opt, _gi) = mls_group
+                .remove_members(&provider, &self.identity.signer, &leaf_indices)
+                .map_err(|e| EngineError::Backend(format!("remove_members: {e:?}")))?;
+            commit_out
+        } else {
+            // The AdminDepletion guard above ensures resulting_admins is
+            // non-empty here.
+            let admin_update = Proposal::AppDataUpdate(Box::new(AppDataUpdateProposal::update(
+                GROUP_ADMIN_POLICY_COMPONENT_ID,
+                crate::app_components::encode_admin_policy(&resulting_admins)?,
+            )));
+            let mut builder = mls_group
+                .commit_builder()
+                .propose_removals(leaf_indices)
+                .add_proposal(admin_update)
+                .load_psks(
+                    <EngineOpenMlsProvider<'_, S> as openmls_traits::OpenMlsProvider>::storage(
+                        &provider,
+                    ),
+                )
+                .map_err(|e| EngineError::Backend(format!("load_psks: {e:?}")))?;
+            let mut app_data = builder.app_data_dictionary_updater();
+            for proposal in builder.app_data_update_proposals() {
+                if let AppDataUpdateOperation::Update(data) = proposal.operation() {
+                    app_data.set(ComponentData::from_parts(
+                        proposal.component_id(),
+                        data.clone(),
+                    ));
+                }
+            }
+            builder.with_app_data_dictionary_updates(app_data.changes());
+            let commit_bundle = builder
+                .build(
+                    <EngineOpenMlsProvider<'_, S> as openmls_traits::OpenMlsProvider>::rand(
+                        &provider,
+                    ),
+                    <EngineOpenMlsProvider<'_, S> as openmls_traits::OpenMlsProvider>::crypto(
+                        &provider,
+                    ),
+                    &self.identity.signer,
+                    |_| true,
+                )
+                .map_err(|e| EngineError::Backend(format!("remove_members build: {e:?}")))?
+                .stage_commit(&provider)
+                .map_err(|e| EngineError::Backend(format!("remove_members stage: {e:?}")))?;
+            let (commit_out, _welcome_opt, _gi) = commit_bundle.into_contents();
+            commit_out
+        };
 
         let staged_commit = mls_group
             .pending_commit()
--- a/crates/cgka-engine/src/message_processor/send.rs
+++ b/crates/cgka-engine/src/message_processor/send.rs
@@ -16,9 +16,8 @@ use cgka_traits::peeler::GroupMessageMetadata;
 use cgka_traits::storage::{LeaveRequest, StorageError, StorageProvider};
 use cgka_traits::transport::{EncryptedPayload, TransportMessage};
 use cgka_traits::types::{EpochId, GroupId, MemberId};
-use openmls::component::ComponentData;
 use openmls::group::MlsGroup;
-use openmls::messages::proposals::{AppDataUpdateOperation, AppDataUpdateProposal, Proposal};
+use openmls::messages::proposals::{AppDataUpdateProposal, Proposal};
 use openmls::prelude::{BasicCredential, MlsMessageOut};
 use std::collections::HashSet;
 use tls_codec::Serialize as _;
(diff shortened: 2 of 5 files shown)


crates/cgka-engine/src/message_processor/send.rs
```diff
@@ -383,9 +386,66 @@ impl<S: StorageProvider> Engine<S> {
...
-        let (commit_out, _welcome_opt, _gi) = mls_group
-            .remove_members(&provider, &self.identity.signer, &leaf_indices)
-            .map_err(|e| EngineError::Backend(format!("remove_members: {e:?}")))?;
+        // Admin/leaf coupling (admin-policy-v1.md "Validation"): if any removed
+        // account is currently listed in `admins`, the same commit must also
+        // drop those keys from the admin-policy component — otherwise the
+        // resulting epoch lists an admin with no member leaf and is invalid.
+        let resulting_admins: Vec<[u8; 32]> = admins
+            .iter()
+            .filter(|admin| !target_admins.iter().any(|target| target == *admin))
```
erskingardner: Admin keys are dropped whenever the removed account appears in `target_admins`, without checking whether any of that account's leaves survive in the resulting epoch.

That is safe **today** because this function always collects *all* leaf indices for each targeted `MemberId` (account-level removal), but the producing-side coupling is implicitly tied to that API contract rather than the spec wording ("last member leaf"). If a future leaf-scoped removal entry point is added, this filter would silently over-prune admins even when the account still has another device in the group.

Consider either (a) a comment tying this to account-level removal, or (b) computing `resulting_admins` from the staged membership delta (same model as `validate_admin_leaf_coupling_for_staged_commit`) so the two sides stay aligned.

crates/cgka-engine/src/message_processor/send.rs
```diff
@@ -383,9 +386,66 @@ impl<S: StorageProvider> Engine<S> {
...
+            commit_out
+        } else {
+            // The AdminDepletion guard above ensures resulting_admins is
+            // non-empty here.
+            let admin_update = Proposal::AppDataUpdate(Box::new(AppDataUpdateProposal::update(
+                GROUP_ADMIN_POLICY_COMPONENT_ID,
+                crate::app_components::encode_admin_policy(&resulting_admins)?,
+            )));
+            let mut builder = mls_group
+                .commit_builder()
```
erskingardner: This ~35-line commit-builder block is nearly identical to `update_group_data.rs` (`commit_builder` -> `app_data_dictionary_updater` -> `with_app_data_dictionary_updates` -> `build` -> `stage_commit`).

Not a functional bug in this PR, but it is a drift magnet: any fix to PSK loading, proposal handling, or dictionary merge semantics will need to land in two places. Worth extracting a shared helper (even a private `fn` on `Engine`) before a third callsite appears.

crates/cgka-engine/src/app_components.rs
```diff
@@ -237,14 +237,20 @@ pub(crate) fn validate_admin_leaf_coupling_for_staged_commit(
...
+        .group_context()
+        .extensions()
+        .app_data_dictionary()
+        .and_then(|dict| dict.dictionary().get(&GROUP_ADMIN_POLICY_COMPONENT_ID));
+    let resulting_admins = match staged_admin_bytes {
+        Some(admin_bytes) => decode_admin_policy(admin_bytes)?,
+        None => admins_of_group(mls_group)?,
     };
-    let resulting_admins = decode_admin_policy(admin_bytes)?;
     if resulting_admins.is_empty() {
```
erskingardner, line 263: Pre-existing behavior, but the new fallback makes this path more salient: when the staged context has no admin-policy bytes we fall back to `admins_of_group`, and if that returns `[]` we skip the cross-component check entirely.

Can a membership-changing commit land on a group with no admin-policy component in the current epoch (e.g. after a prior GCE stripped the dictionary)? If yes, a commit that removes the last admin's leaf while the carried-forward set is empty would bypass this guard. If groups without the component cannot exist in practice, a one-line comment would help future readers.

crates/cgka-engine/tests/distributed_convergence.rs
```diff
@@ -675,6 +752,109 @@ async fn convergence_rejects_remove_that_leaves_orphan_admin_key() {
     );
 }
 
+#[tokio::test]
+async fn convergence_rejects_remove_whose_context_carries_no_admin_policy_bytes() {
```
erskingardner, line 756: This regression only exercises stored convergence replay (`buffer` + `converge_stored_openmls_messages`). The #393 fix also applies on the live ingest path (`message_processor/ingest.rs` calls the same validator).

Same function, so low risk, but a sibling test that calls `carol.ingest(invalid_remove)` and expects `EngineError` / `MessageState::Failed` would pin the hot path explicitly and catch accidental divergence if ingest and projection ever diverge again.

erskingardner (commented):
## Adversarial review summary

The core fix looks correct: the validation fallback to `admins_of_group` matches the spec's carried-forward semantics, and the producing-side coupled `Remove + AppDataUpdate` path mirrors the existing `update_group_data` commit-builder flow. Tests cover the previously broken send path and the GCE/drop-dictionary attack vector on convergence replay.

I don't see a correctness blocker in the diff, but a few latent risks and coverage gaps below are worth addressing or documenting before merge.

crates/cgka-engine/src/message_processor/send.rs
```diff
@@ -383,9 +386,66 @@ impl<S: StorageProvider> Engine<S> {
...
+            // No removed member is an admin: a plain Remove commit carries the
+            // admin set forward unchanged.
+            let (commit_out, _welcome_opt, _gi) = mls_group
+                .remove_members(&provider, &self.identity.signer, &leaf_indices)
+                .map_err(|e| EngineError::Backend(format!("remove_members: {e:?}")))?;
+            commit_out
+        } else {
+            // The AdminDepletion guard above ensures resulting_admins is
+            // non-empty here.
+            let admin_update = Proposal::AppDataUpdate(Box::new(AppDataUpdateProposal::update(
```
erskingardner, line 423: This path now stages an admin-policy update, but the local publish-confirmation events below still only queue `MemberRemoved` changes from `unique_targets`.

That makes the sender diverge from receivers: inbound ingest snapshots `before_admins` / `after_admins` and emits `AdminRemoved` for the same commit, and `UpdateAppComponents` already queues `admin_changes(&before, &after)` for locally authored admin updates. The author of a co-admin removal would only get the membership system row/audit event, while every receiver gets both membership and admin-policy rows.

Can we append `group_state_changes::admin_changes(&admins, &resulting_admins)` to `pending_state_changes` whenever this branch builds the coupled `AppDataUpdate`?

crates/cgka-engine/src/message_processor/send.rs
erskingardner: Fixed in b7cea5e — took option (b). `resulting_admins` is now derived from the staged membership delta: collect the leaves this commit removes, walk `mls_group.members()` to build the set of accounts with at least one surviving leaf, and keep only admins present in that set. This is the same model `validate_admin_leaf_coupling_for_staged_commit` uses, so an admin key is dropped exactly when none of its account's leaves survives, independent of how targets were expanded to leaf indices. The AdminDepletion guard still guarantees a non-empty coupled set (an untargeted admin always keeps its leaves).

crates/cgka-engine/src/message_processor/send.rs
erskingardner: Extracted in b7cea5e: `Engine::stage_commit_with_app_data_updates(mls_group, provider, removals, proposals, context)` in update_group_data.rs owns the commit_builder → dictionary-updater → with_app_data_dictionary_updates → build → stage_commit sequence. Both callsites (UpdateAppComponents and the coupled removal path) now use it; the `context` arg keeps error messages attributable to the calling intent. upgrade.rs has a similar-looking block but a different shape (GCE + AppDataUpdate mixed), so I left it out of this extraction.

crates/cgka-engine/tests/distributed_convergence.rs
erskingardner: Added in b7cea5e: `live_ingest_rejects_remove_whose_context_carries_no_admin_policy_bytes` pushes the same invalid commit through the public `ingest` entry point (no pre-buffering), asserts the outcome is never `Processed`, drives convergence to quiescence, and checks epoch/membership/admin view are untouched and the message lands `EpochInvalidated`. Verified it fails against the pre-fix validator (commit was accepted) and passes with it. Note on the truly direct staged-commit seam at ingest.rs:649: inbound commits with `msg_epoch >= current_epoch` always route into convergence (`commit_should_enter_convergence`), so a fresh current-epoch commit can't hit that seam from the public API outside fork-recovery replay — both seams call the same shared validator, and the new test pins the reachable hot path.

crates/cgka-engine/src/app_components.rs
erskingardner: Investigated and documented in b7cea5e (comment at the early return). Findings: (1) engine-created groups always write the component when it's in the negotiated required set — creator is implicitly admin, and the encoding cannot represent an empty list (`encode_admin_policy`/`decode_admin_policy` both reject it), so an empty computed set only means the component is absent; (2) `validate_app_component_remove` rejects AppDataUpdate::Remove of required components on ingest/replay; (3) a GCE-only commit CAN strip the dictionary today — OpenMLS's draft-08 guard only applies to commits that also carry AppDataUpdate proposals — but the resulting admin-less group is not a coupling bypass: every commit shape that removes ANOTHER member's leaf requires an authenticated admin sender (`require_admin_for_staged_commit`, enforced before this check on all three call paths), which fail-closes on an empty admin set, and SelfRemove-only can't orphan a member of an empty set. So the two-commit sequence (strip, then remove) dies at the admin gate. The GCE-strip itself is a separate required-components-persistence gap (admin-authored group freeze / DoS, no key orphaned) — I've filed it as a follow-up task rather than widening this PR.

crates/cgka-engine/src/message_processor/send.rs
erskingardner: Fixed in b7cea5e: the pending state changes now append `group_state_changes::admin_changes(&admins, &resulting_admins)` (attributed to self, matching the UpdateAppComponents path) alongside the `MemberRemoved` rows. Since `resulting_admins ⊆ admins` this yields only `AdminRemoved` entries, and it's empty on the plain (non-coupled) branch, so uncoupled removes are unchanged. The invite_leave test now drains the author's events after confirm and asserts BOTH MemberRemoved and AdminRemoved for the removed co-admin.

crates/cgka-engine/src/update_group_data.rs
```diff
@@ -357,4 +333,55 @@ impl<S: StorageProvider> Engine<S> {
...
+        let mut app_data = builder.app_data_dictionary_updater();
+        for proposal in builder.app_data_update_proposals() {
+            if let AppDataUpdateOperation::Update(data) = proposal.operation() {
+                app_data.set(ComponentData::from_parts(
+                    proposal.component_id(),
+                    data.clone(),
+                ));
+            }
+        }
+        builder.with_app_data_dictionary_updates(app_data.changes());
```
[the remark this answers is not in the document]
erskingardner: Fixed in 5f77f49. Verified first: the Update-only loop was inherited verbatim from the original `do_send_update_app_components` block (pre-existing, not introduced by the extraction), and no current caller can produce a Remove — every send site constructs `AppDataUpdateProposal::update` and `SendIntent::UpdateAppComponents` only carries update bytes. Since the receiver-side `process_commit_with_app_data_updates` applies both operations and the helper is now shared infrastructure, it now mirrors Remove into the dictionary updater AND runs the same `validate_app_component_remove` guard recipients enforce (checked before the builder takes its mutable borrow), so a future caller can't stage a commit whose local dictionary diverges from what recipients compute — or stage a required-component removal recipients would reject. No test exercises the Remove arm: the helper is `pub(crate)` and no public intent surfaces a Remove yet, so it's defensive parity; the receiver-side Remove path keeps its existing coverage.

PR #702, opened 2026-07-03
Withdraw losing committer's group-state notifications via explicit invalidation event (#363)

### LINKED ISSUE #363: [Bug]: Concurrent rename: losing committer's local timeline shows their losing UpdateGroupData commit as a successful system message; production surface of the convergence weakness #600 fixed simulator-side
## Summary

When two admins concurrently rename a group, the losing committer's local timeline renders **their own losing rename as a successful system message**, immediately after the winning rename's system message. The actual canonical group state holds the winner's name. The losing client's history therefore claims an action took effect that never did.

This is the production-side surface of the convergence weakness fixed simulator-side by marmot-protocol/mdk#447 (group_name convergence comparison) and partially tracked by marmot-protocol/mdk#454 (member-set / payload-set convergence assertions): the engine routes a losing UpdateGroupData commit to the application as if it landed, and the app surfaces it as a confirmed system message.

## Reproduction

1. Group G has two admins, A and B, both online, both on the conversation screen.
2. A renames G → "New name from A". Commit submitted. MLS commit lands and propagates.
3. Concurrently (within the same epoch window, before A's commit reaches B), B renames G → "New name from B". Commit submitted.
4. Convergence resolves. A's commit wins. The canonical group_name is "New name from A".
5. Observed on A's device:
   - Timeline shows one system message: "A renamed the group to 'New name from A'."
   - Group header shows "New name from A". Correct.
6. Observed on B's device:
   - Timeline shows **two** system messages, in order:
     1. "A renamed the group to 'New name from A'."
     2. "B renamed the group to 'New name from B'." (this is the lie)
   - Group header shows "New name from A" (correctly applies the winning commit).
   - The second system message claims B's rename succeeded, but the group's actual name is A's. B's rename had no effect on canonical state.

## Why this is a bug

The losing committer's history asserts a successful action that the canonical state contradicts. From B's perspective:

- B intended to rename the group. The app reports the rename as completed (system message present, timestamped, attributed to B).
- The canonical group state holds a different name (A's winning value).
- B's only path to detecting the divergence is reading the header back and noticing the name string disagrees with the system message immediately above.

This sits in the same family of trust failures as marmot-protocol/mdk#468 (rendered membership diverging from MLS state) and marmot-protocol/mdk#463 (admin/leaf invariant violation after fork): the projected state asserts something the canonical state disproves, with the disagreement hidden.

## Likely cause

When B's losing UpdateGroupData commit is processed, the engine appears to:

1. Apply B's commit locally as if it landed.
2. Emit a system-message event to the application layer ("B renamed the group to X").
3. Later, on convergence resolution, replace the canonical group_name with A's winning value.
4. **Without** retracting / rolling back the system-message event emitted in step 2, and **without** emitting any "your rename was superseded" event.

The application layer faithfully renders the original event, producing the lying system message.

Two possible fix sides:

1. **Engine side:** during convergence resolution, when a losing UpdateGroupData commit is dropped, also emit a retraction event for any system-message events derived from it. The app already has the rendering for "rename happened"; it just needs the symmetric "rename retracted" event to mark the row as superseded (or remove it).
2. **App side:** if the engine cannot easily retract emitted events post-hoc, the app must observe the canonical group_name after each commit lands and reconcile its system-message history. Any rename-system-message whose target string disagrees with the current canonical name should be marked "(superseded)" or removed.

Engine-side is correct. The app shouldn't have to maintain a separate reconciliation pass over system messages it already trusted the engine to emit correctly.

## Expected behavior

On B's device after convergence:

- Either: the second system message is removed entirely (B's losing commit produces no visible artifact, only the winner's message remains).
- Or: the second system message is rendered with a clear "(superseded)" marker and visually de-emphasized, so B understands their rename lost the race.

The first option is simpler and matches how most chat apps handle concurrent edits (only the winning edit appears). The second option preserves B's intent in their own history, useful for explainability ("my rename happened but was overridden"), but adds rendering complexity.

## Test plan

- A and B both rename concurrently → both devices show exactly one rename system message (the winner) and the same canonical name in the header. Either no losing-rename artifact, or one clearly marked superseded.
- A renames, then B renames after A's commit has already converged → both devices show two ordered rename messages, both accurate (sequential, no fork).
- Three admins concurrently rename → only the canonical winner appears in everyone's timeline (or losers appear marked superseded).
- Same scenario with avatar change instead of name change (GroupData updates beyond just name) → same correctness applies.
- Same scenario with concurrent admin-policy update and rename → no cross-contamination of which "lost" message gets retracted.

## Identifiers

Available privately if engine debug is needed. Repro is cheap: two admin devices, simultaneous rename.

## Related

- marmot-protocol/mdk#468 (CLOSED, this repo) — Rendered membership disagrees with MLS state after convergence fork. Same family of "projected state lies about canonical state".
- marmot-protocol/mdk#463 (CLOSED, this repo) — admin-policy invariant violation after convergence fork.
- marmot-protocol/mdk#447 (CLOSED, this repo) — Simulator oracle now compares `group_name` for ClientConvergence. Test-side gap closed; this issue is the production surface that gap was hiding.
- marmot-protocol/mdk#454 (OPEN, this repo) — Convergence property tests don't compare member set / payload set. Same test-gap family; should include system-message-event equivalence as another dimension worth asserting.
- marmot-protocol/mdk#632 (CLOSED, this repo) — Past-epoch competing commit within rewind horizon dropped inline as Failed instead of routed to convergence → permanent fork. Adjacent code path; the "dropped inline" handling here is the symmetric place where the losing commit's emitted events need cleanup.

## Notes

- Tested via the whitenoise-android consumer on GrapheneOS.
- HIGH priority: the timeline lies about an action's outcome; trust / correctness issue equivalent to the marmot-protocol/mdk#468 / marmot-protocol/mdk#463 family.

erskingardner:
## Problem

Two admins concurrently rename a group; convergence picks A's commit. B's device correctly shows A's name in the header, but B's timeline still shows "B renamed the group to X" as a successful system message: the engine emitted `GroupStateChanged` for B's own commit at `confirm_published` and never retracted it when branch selection superseded that commit.

Spec basis: marmot-protocol/marmot#171 (merged) — `protocol-core/convergence.md` "Applying the selected branch" and `protocol-core/inbound-processing.md` "Application-visible output". When branch selection supersedes a commit the client previously applied — **including the client's own published and confirmed commit** — the client MUST emit a group-state-change invalidation naming the superseded commit, withdrawing every state notification attributed to it. Conformance is the resulting view: settled notifications == exactly those derivable from accepted commits on the selected branch.

## Design

**New event** (`crates/traits/src/engine.rs`):

```rust
GroupEvent::GroupStateInvalidated {
    group_id: GroupId,
    epoch: EpochId,                     // source epoch of the fork the commit lost
    invalidated_commit_id: MessageId,   // same id space as GroupStateChanged.origin_commit_id
    reason: GroupStateInvalidationReason, // SupersededByBranchSelection
}
```

**Attribution: `origin_commit_id` (MessageId), not `commit_digest`.** `GroupStateChanged` already carries `origin_commit_id` as the local attribution key, and both supersession seams already hold the exact id used at emit time:

- Direct staged-commit seam: the fork-recovery incumbent record's `storage_id` is the same id `confirm_published` stamps onto the committer's own `GroupStateChanged` events (both come from the same `CommitRecoveryRecord`), and the same id `ForkRecovered` reports.
- Stored-convergence seam: inbound commits are keyed by the content-derived dedup id on both the emit path (`push_group_state_change(origin_commit_id = msg.id)`) and the drop path (`dropped_messages`).

So the MessageId deterministically identifies the superseded commit across emit and invalidate **per device**, which is all the spec's implementation-defined notification surface requires. No new commit_digest plumbing was needed; the digest already participates via `CommitOrderingKey` for selection, not attribution.

**Emit sites** (both supersession seams):

1. `crates/cgka-engine/src/message_processor/ingest.rs` — alongside `ForkRecovered` when a same-epoch candidate wins over the previously applied incumbent. This is the issue's exact scenario: the losing committer's own published-and-confirmed commit is the incumbent (`we_committed_from` routes the sibling commit to this seam, not convergence).
2. `crates/cgka-engine/src/distributed_convergence.rs::emit_rolled_back_commits` — alongside `CommitRolledBack` for every commit dropped `InvalidAgainstCandidateState`. Idempotent for commits this client never applied (no row carries the id), same argument as the existing `CommitRolledBack` emission.

The commit-level events (`ForkRecovered` / `CommitRolledBack`) are kept unchanged; `GroupStateInvalidated` is the explicit, seam-independent state-notification withdrawal signal.

**Attribution audit (scoping item 3):** every commit-derived `GroupStateChanged` emitter populates `origin_commit_id` — direct seam (`Some(msg.id)`), own-commit confirm (`Some(recovery record id)`), and single-commit convergence passes. The one remaining `None` emitter is a multi-commit convergence pass, where the previous-tip → selected-tip diff is the *combined* effect of several accepted commits and per-commit attribution is inherently ambiguous (pre-existing, documented on the field). Those unattributed rows describe accepted-branch canonical state, not losing-branch claims, so they don't reproduce this bug; making them per-commit attributable would require per-commit replay diffing and is left out per the smallest-design guidance.

**App + FFI:**

- `marmot-app`: event routed like `CommitRolledBack` (`event_routing.rs`, `groups.rs::event_group_id`, reaches the subscriber surface via `observe_event`); `client/sync.rs` invalidates every kind-1210 row whose `origin_commit_id` matches (multi-row, idempotent alongside the existing `ForkRecovered`/`CommitRolledBack` handlers).
- `marmot-uniffi`: `GroupEventKindFfi::GroupStateInvalidated { epoch, invalidated_commit_id_hex, reason }` with the stable low-cardinality tag `superseded_by_branch_selection`. Bindings are derived artifacts regenerated by the platform scripts per `crates/marmot-uniffi/README.md`; only the Rust side is in-repo.

## Deliberately skipped

- **Conformance-simulator oracle**: adding an invalidation `OracleBehavior` fans out into the serialized vector observation schema (`ClientObservation`/`ClientEventCounts`), evidence counting, per-stimulus expectation tables, and stored fixture regeneration. The engine-level tests below are the required coverage; the oracle extension is left for a follow-up. Existing simulator tests pass unchanged (the new event is not extracted into any observation, so fixtures stay valid).
- **Multi-commit convergence-pass attribution** (see audit note above).

## Test plan

- `concurrent_rename_withdraws_losing_committers_state_notification` (crates/cgka-engine/tests/update_group_data.rs) — the issue scenario: both admins rename in the same epoch window, both commits delivered to both. Asserts: canonical name converges; the losing committer emits exactly one `GroupStateInvalidated` whose `invalidated_commit_id` equals the `origin_commit_id` of its earlier `GroupStateChanged { GroupRenamed }`; the winner withdraws nothing and classifies the losing commit stale; and the **resulting view** on the losing device (notifications minus withdrawals) is exactly the winner's rename attributed to the winning commit.
- `sequential_renames_emit_two_notifications_without_invalidation` — sequential control: two ordered renames, both survive, zero invalidations on either device.
- `concurrent_rename_and_retention_change_withdraw_only_superseded_commit` — cross-contamination guard: rename racing a message-retention change; only the superseded commit's notification is withdrawn, the winner's survives with its attribution.
- `convergence_rollback_emits_commit_rolled_back_for_losing_branch` (tests/distributed_convergence.rs) — extended: the stored-convergence seam emits `GroupStateInvalidated` naming the losing commit (and never the accepted commit).
- Traits snapshot for the new event variant.
- Suites: `cargo test -p cgka-traits -p cgka-engine -p marmot-app -p marmot-uniffi -p cgka-conformance-simulator` all green; `just fast-ci` (fmt-check, check incl. OTLP features, clippy) green.

Closes #363

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/702">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added explicit **group-state invalidation** events when an applied branch is superseded, including a standardized reason and the superseded commit reference.
  * Extended event propagation so invalidations update timelines and related app views.
  * Updated the client/FFI event stream to deliver the new invalidation event consistently.

* **Bug Fixes**
  * Improved rollback and fork-recovery behavior so superseded (losing-branch) notifications are withdrawn while winning-branch notifications remain.

* **Tests**
  * Added regression coverage for concurrent admin renames, mixed-change isolation, and durable/convergence replay correctness.
  * Added timeline invalidation no-op behavior for already-invalidated rows.

### DIFF
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -14,7 +14,9 @@ use crate::openmls_projection::{
 };
 use std::sync::atomic::{AtomicU64, Ordering};
 
-use cgka_traits::engine::{AppMessageInvalidationReason, GroupEvent, GroupStateChange};
+use cgka_traits::engine::{
+    AppMessageInvalidationReason, GroupEvent, GroupStateChange, GroupStateInvalidationReason,
+};
 use cgka_traits::message::{MessageRecord, MessageState, StoredMessagePayload};
 use cgka_traits::storage::{StorageError, StorageProvider};
 use cgka_traits::transport::TransportMessage;
@@ -693,6 +695,13 @@ impl<S: StorageProvider> Engine<S> {
     /// and the app-side tombstone is a no-op when no row carries the commit id,
     /// so emitting here is idempotent and safe even for commits this client
     /// never applied.
+    ///
+    /// Each rolled-back commit also gets a [`GroupEvent::GroupStateInvalidated`]:
+    /// the spec-required explicit withdrawal of every state notification
+    /// attributed to the superseded commit (convergence.md "Applying the
+    /// selected branch"). This covers the client's own published-and-confirmed
+    /// commit when it loses branch selection through the stored-convergence
+    /// seam, mirroring the direct seam's `ForkRecovered` pairing.
     fn emit_rolled_back_commits(
         &mut self,
         group_id: &GroupId,
@@ -707,10 +716,29 @@ impl<S: StorageProvider> Engine<S> {
             {
                 continue;
             }
+            let invalidated_commit_id = message_id_from_hex(&dropped.message_id)?;
+            // The stored record's epoch is the commit's source epoch (the fork
+            // it lost). Best-effort: fall back to the selected branch's fork
+            // epoch when the record is gone — the id-matched withdrawal is what
+            // conformance requires; the epoch is presentation metadata.
+            let epoch = self
+                .storage
+                .get_message(&invalidated_commit_id)
+                .map(|record| record.epoch)
+                .unwrap_or(EpochId(
+                    result.selected_fork_epoch.unwrap_or(result.previous_tip),
+                ));
             self.events_buf.push_back(GroupEvent::CommitRolledBack {
                 group_id: group_id.clone(),
-                invalidated_commit_id: message_id_from_hex(&dropped.message_id)?,
+                invalidated_commit_id: invalidated_commit_id.clone(),
             });
+            self.events_buf
+                .push_back(GroupEvent::GroupStateInvalidated {
+                    group_id: group_id.clone(),
+                    epoch,
+                    invalidated_commit_id,
+                    reason: GroupStateInvalidationReason::SupersededByBranchSelection,
+                });
         }
         Ok(())
     }
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -435,6 +435,7 @@ impl<S: StorageProvider> Engine<S> {
         self.emit_application_replay_events(group_id, &observations);
         self.emit_invalidated_app_events(group_id, &result)?;
         self.emit_rolled_back_commits(group_id, &result)?;
+        self.emit_superseded_processed_commits(group_id, &result)?;
 
         // A selected branch reached the canonical state (Settled): the run is
         // applied/stable. With no selected branch the pass simply settled with
@@ -718,16 +719,18 @@ impl<S: StorageProvider> Engine<S> {
             }
             let invalidated_commit_id = message_id_from_hex(&dropped.message_id)?;
             // The stored record's epoch is the commit's source epoch (the fork
-            // it lost). Best-effort: fall back to the selected branch's fork
-            // epoch when the record is gone — the id-matched withdrawal is what
-            // conformance requires; the epoch is presentation metadata.
-            let epoch = self
-                .storage
-                .get_message(&invalidated_commit_id)
-                .map(|record| record.epoch)
-                .unwrap_or(EpochId(
-                    result.selected_fork_epoch.unwrap_or(result.previous_tip),
-                ));
+            // it lost). A missing record falls back to the selected branch's
+            // fork epoch — the id-matched withdrawal is what conformance
+            // requires; the epoch is presentation metadata. Any other storage
+            // failure propagates: masking a locked/corrupt backend here would
+            // emit a fabricated epoch and hide the failure from the caller.
+            let epoch = match self.storage.get_message(&invalidated_commit_id) {
+                Ok(record) => record.epoch,
+                Err(StorageError::NotFound) => {
+                    EpochId(result.selected_fork_epoch.unwrap_or(result.previous_tip))
+                }
+                Err(e) => return Err(OpenMlsProjectionError::Storage(format!("{e:?}"))),
+            };
             self.events_buf.push_back(GroupEvent::CommitRolledBack {
                 group_id: group_id.clone(),
                 invalidated_commit_id: invalidated_commit_id.clone(),
@@ -737,6 +740,100 @@ impl<S: StorageProvider> Engine<S> {
                     group_id: group_id.clone(),
                     epoch,
                     invalidated_commit_id,
+                    reason: GroupStateInvalidationReason::SupersededByBranchSelection,
+                });
+        }
+        Ok(())
+    }
+
+    /// Withdraw every commit this device PREVIOUSLY APPLIED (stored state
+    /// `Processed`) that a convergence apply left off the selected branch.
+    ///
+    /// The canonicalization drop set only covers commits the candidate BFS
+    /// could materialize. A device's OWN published-and-confirmed commit is not
+    /// replayable through `process_message` (MLS cannot process own messages),
+    /// so when a reorg supersedes it — e.g. after a restart cleared the
+    /// in-memory `committed_from` guard and routed a same-epoch sibling into
+    /// stored convergence — the own commit gets no disposition at all: no
+    /// `CommitRolledBack`, no withdrawal, and the confirm-time
+    /// `GroupStateChanged` rows survive as the issue #363 lie.
+    ///
+    /// Spec (convergence.md "Applying the selected branch"): branch selection
+    /// superseding "a commit the client previously applied — including the
+    /// client's own published and confirmed commit" MUST withdraw the state
+    /// notifications attributed to it. `Processed` is exactly "previously
+    /// applied"; not on the accepted path at or above the selected fork epoch
+    /// is exactly "superseded". Commits the drop set already covered are
+    /// skipped so each supersession is reported once.
+    fn emit_superseded_processed_commits(
+        &mut self,
+        group_id: &GroupId,
+        result: &CanonicalizationResult,
+    ) -> Result<(), OpenMlsProjectionError> {
+        if result.selected_tip.is_none() {
+            return Ok(());
+        }
+        let Some(fork_epoch) = result.selected_fork_epoch else {
+            return Ok(());
+        };
+        let accepted: std::collections::BTreeSet<&str> =
+            result.accepted_commits.iter().map(String::as_str).collect();
+        let records = self
+            .storage
+            .list_messages(group_id, EpochId(0))
+            .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?;
+        for record in records {
+            if record.state != MessageState::Processed || record.epoch.0 < fork_epoch {
+                continue;
+            }
+            let payload = StoredMessagePayload::decode(&record.payload)
+                .map_err(|e| OpenMlsProjectionError::Serialize(format!("{e:?}")))?;
+            let Some(message) = payload.as_openmls_wire() else {
+                continue;
+            };
+            if project_mls_message(&message.payload)?.kind != OpenMlsContentKind::Commit {
+                continue;
+            }
+            let record_id_hex = hex::encode(record.id.as_slice());
+            // The same MLS bytes can be keyed by the wrap-time transport id
+            // (own sent commits) or the content dedup id (inbound commits), so
+            // acceptance is checked under both aliases before declaring the
+            // record superseded.
+            let content_id_hex = hex::encode(
+                crate::message_processor::content_dedup_id(&message.payload).as_slice(),
+            );
+            if accepted.contains(record_id_hex.as_str())
+                || accepted.contains(content_id_hex.as_str())
+            {
+                continue;
+            }
+            if result
+                .dropped_messages
+                .iter()
+                .any(|dropped| dropped.message_id == record_id_hex)
+            {
+                continue;
+            }
+            self.storage
+                .update_message_state(&record.id, MessageState::EpochInvalidated)
+                .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?;
+            self.audit_group(
+                group_id,
+                crate::audit_helpers::message_state_changed_event(
+                    record_id_hex,
+                    MessageState::EpochInvalidated,
+                    "superseded_processed_commit",
+                ),
+            );
+            self.events_buf.push_back(GroupEvent::CommitRolledBack {
+                group_id: group_id.clone(),
+                invalidated_commit_id: record.id.clone(),
+            });
+            self.events_buf
+                .push_back(GroupEvent::GroupStateInvalidated {
+                    group_id: group_id.clone(),
+                    epoch: record.epoch,
+                    invalidated_commit_id: record.id,
                     reason: GroupStateInvalidationReason::SupersededByBranchSelection,
                 });
         }
--- a/crates/cgka-engine/tests/update_group_data.rs
+++ b/crates/cgka-engine/tests/update_group_data.rs
@@ -426,6 +426,113 @@ async fn create_pair() -> (
     (alice, bob, gid)
 }
 
+/// Like [`create_pair`] but with storage handles and BOTH members as group
+/// admins, so either side can publish an admin-gated `UpdateGroupData` /
+/// `UpdateAppComponents` commit (the concurrent-committer fork scenarios).
+async fn create_admin_pair_with_storage() -> (
+    Engine<SqliteAccountStorage>,
+    SqliteAccountStorage,
+    Engine<SqliteAccountStorage>,
+    SqliteAccountStorage,
+    GroupId,
+) {
+    let (mut alice, alice_storage) = build_with_storage(b"alice");
+    let (mut bob, bob_storage) = build_with_storage(b"bob");
+    let bob_kp = bob.fresh_key_package().await.unwrap();
+    let (gid, create) = alice
+        .create_group(CreateGroupRequest {
+            name: "original".into(),
+            description: "orig description".into(),
+            members: vec![bob_kp],
+            required_features: vec![],
+            app_components: vec![],
+            initial_admins: vec![bob.self_id()],
+        })
+        .await
+        .unwrap();
+    let (pending, welcomes) = match create {
+        SendResult::GroupCreated { pending, welcomes } => (pending, welcomes),
+        _ => unreachable!(),
+    };
+    alice.confirm_published(pending).await.unwrap();
+    let welcome = welcomes.into_iter().next().unwrap();
+    bob.join_welcome(welcome).await.unwrap();
+    (alice, alice_storage, bob, bob_storage, gid)
+}
+
+fn route_to_group(msg: &TransportMessage, gid: &GroupId) -> TransportMessage {
+    TransportMessage {
+        envelope: TransportEnvelope::GroupMessage {
+            transport_group_id: gid.as_slice().to_vec(),
+        },
+        ..msg.clone()
+    }
+}
+
+/// The `(change, origin_commit_id)` of every `GroupStateChanged` in `events`.
+/// Asserts each commit-derived notification carries an attribution — without
+/// it the notification could never be withdrawn by origin commit.
+fn attributed_state_changes(
+    events: &[cgka_traits::engine::GroupEvent],
+) -> Vec<(cgka_traits::engine::GroupStateChange, MessageId)> {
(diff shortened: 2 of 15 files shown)


crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -707,10 +716,29 @@ impl<S: StorageProvider> Engine<S> {
...
+            // it lost). Best-effort: fall back to the selected branch's fork
+            // epoch when the record is gone — the id-matched withdrawal is what
+            // conformance requires; the epoch is presentation metadata.
+            let epoch = self
+                .storage
+                .get_message(&invalidated_commit_id)
+                .map(|record| record.epoch)
+                .unwrap_or(EpochId(
+                    result.selected_fork_epoch.unwrap_or(result.previous_tip),
+                ));
```
erskingardner: Using `unwrap_or` here treats every storage failure as if the row were absent. `get_message` can return transient/backend/serialization errors, and most engine call sites only ignore `StorageError::NotFound` while propagating the rest. With this code a locked/corrupt DB during convergence would still emit a successful invalidation with a fallback epoch and the convergence run would continue, hiding the storage failure. Please match on `get_message` and only fallback for `NotFound`; propagate other errors as `OpenMlsProjectionError::Storage`.

crates/marmot-app/src/client/sync.rs
```diff
@@ -488,6 +488,26 @@ impl AppClient {
             {
                 summary.projection_updates.push(projection_update);
             }
+            // Explicit state-notification withdrawal (convergence.md "Applying
```
erskingardner: This is not actually idempotent with the preceding `ForkRecovered` / `CommitRolledBack` handlers. The engine now emits `GroupStateInvalidated` immediately after those events, and `invalidate_app_events_by_origin_commit` updates all matching rows unconditionally and returns a projection update even when they were already invalidated. So one rollback now produces duplicate timeline updates and overwrites the reason from `LosingBranch` to `SupersededByBranchSelection`. I'd either route app tombstoning only from the explicit state-invalidation event, or make the storage helper no-op for already-invalidated rows before relying on this idempotence.

erskingardner (commented):
Adversarial pass found two issues worth fixing before merge.

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -707,10 +716,29 @@ impl<S: StorageProvider> Engine<S> {
             {
                 continue;
             }
+            let invalidated_commit_id = message_id_from_hex(&dropped.message_id)?;
```
erskingardner, line 740: **Cross-seam `origin_commit_id` mismatch in production Nostr (pre-existing, but now duplicated).**

The PR description says `invalidated_commit_id` matches `GroupStateChanged.origin_commit_id`. That holds *within* each seam:
- **Direct / fork recovery:** both use fork-recovery `storage_id` (= `wrapped.id`, the peeler-assigned transport `MessageId`)
- **Stored convergence:** both use the content-derived dedup id (`SHA-256(mls_bytes)` from `dropped.message_id`)

In production Nostr those identifiers differ — see `persist_sent_openmls_content_marker`, which exists because `wrapped.id` (Nostr event id) ≠ `content_dedup_id(mls_bytes)`.

`confirm_published` stamps kind-1210 rows with the **transport** id. If a locally confirmed commit is later dropped by a **stored-convergence reorg** (emitting `CommitRolledBack` / this new `GroupStateInvalidated` with the **content** id, without a direct-seam `ForkRecovered`), `invalidate_timeline_origin_commit` is a no-op and the stale system row survives. That gap predates this PR (`CommitRolledBack` had the same id space), but the new event inherits it.

Issue #363 is fixed on the direct seam because fork recovery uses the same transport id end-to-end. Consider either stamping `origin_commit_id` with `content_dedup_id` at confirm time, or teaching invalidation to resolve both ids for the same MLS bytes.

crates/cgka-engine/tests/update_group_data.rs
```diff
@@ -1035,6 +1142,417 @@ async fn convergence_emits_attributed_message_retention_change_events() {
...
+/// attributed to the superseded commit via an explicit
+/// `GroupStateInvalidated` naming that commit — including the client's own
+/// published-and-confirmed commit.
+#[tokio::test]
+async fn concurrent_rename_withdraws_losing_committers_state_notification() {
+    use cgka_traits::engine::{GroupStateChange, GroupStateInvalidationReason};
+    use cgka_traits::ingest::{IngestOutcome, StaleReason};
+
+    let (mut alice, alice_storage, mut bob, bob_storage, gid) =
+        create_admin_pair_with_storage().await;
```
erskingardner: **Regression tests collapse transport and content ids.**

`MockPeeler` sets `TransportMessage.id = hash_id(ciphertext)`, so `content_id(msg)` equals the fork-recovery `storage_id` and equals inbound `msg.id`. All three new concurrent-supersession tests therefore pass without exercising the production split where Nostr event ids ≠ `SHA-256(mls_bytes)`.

Recommend a variant (ephemeral transport id per wrap, content dedup unchanged) and assert that direct-seam invalidation still tombstones the `confirm_published` row. Without that, the cross-seam gap in `emit_rolled_back_commits` stays untested.

crates/marmot-app/src/client/sync.rs
```diff
@@ -488,6 +488,26 @@ impl AppClient {
...
+            // the selected branch"): branch selection superseded a previously
+            // applied commit — including this account's own published-and-
+            // confirmed commit — so every kind-1210 system row stamped with its
+            // origin commit id is invalidated. Idempotent alongside the
+            // `ForkRecovered` / `CommitRolledBack` handlers above, which fire
+            // for the same commit on their respective seams.
+            if let cgka_traits::engine::GroupEvent::GroupStateInvalidated {
```
erskingardner: **No app-layer regression test for this handler.**

Engine tests assert event emission and resulting-view logic, but nothing in `marmot-app` verifies that `GroupStateInvalidated` (or the paired `ForkRecovered` path) actually invalidates persisted kind-1210 rows via `invalidate_timeline_origin_commit`. A routing/projection wiring mistake would pass CI today. Even a focused unit test with a fake storage backend asserting `origin_commit_id` tombstoning would close the loop.

erskingardner (commented):
## Adversarial review

Solid fix for the #363 direct-seam scenario (concurrent rename → `ForkRecovered` + explicit `GroupStateInvalidated`, matching `origin_commit_id` from `confirm_published`). Engine integration coverage is strong.

I left inline notes on two correctness gaps that the new tests do not exercise, plus a missing app-layer regression test. (CodeRabbit already flagged the `unwrap_or` storage swallow and the non-idempotent triple tombstone in `sync.rs` — I agree with both.)

crates/cgka-engine/src/distributed_convergence.rs
erskingardner: Fixed in ed4975c: `emit_rolled_back_commits` now matches on the error — the fork-epoch fallback applies only to `StorageError::NotFound`; every other storage failure propagates as `OpenMlsProjectionError::Storage`, so a locked/corrupt backend fails the convergence pass instead of emitting an invalidation with a fabricated epoch.

crates/marmot-app/src/client/sync.rs
erskingardner: Fixed in ed4975c, both halves of your either/or:

1. Routing: origin-commit tombstoning now fires ONLY from `GroupStateInvalidated` — the duplicate calls in the `ForkRecovered`/`CommitRolledBack` handlers are removed. The engine pairs the withdrawal event with both rollback seams (the direct-seam `ForkRecovered` site and every convergence emission path, including the new superseded-`Processed`-commit pass), so no rollback path emits without it; the engine tests assert the pairing on each seam.
2. Storage: `invalidate_app_events_by_origin_commit` now filters `AND invalidated = 0` in both the SELECT and the UPDATE, so a replayed withdrawal is a projection no-op and never rewrites the recorded reason (pinned by a storage test and by the new marmot-app test).

One rollback now produces exactly one projection update with one reason.

crates/marmot-app/src/client/sync.rs
```diff
@@ -488,6 +488,26 @@ impl AppClient {
...
+                ..
+            } = event
+                && let Some(projection_update) = self.app.invalidate_timeline_origin_commit(
+                    &self.state.label,
+                    &hex::encode(invalidated_commit_id.as_slice()),
+                    &format!("{reason:?}"),
+                )?
+            {
+                summary.projection_updates.push(projection_update);
+            }
```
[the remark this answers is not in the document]
erskingardner: Addressed in ed4975c together with the maintainer thread above: `GroupStateInvalidated` is now the single origin-commit tombstoning trigger (duplicate `ForkRecovered`/`CommitRolledBack` dispatches removed), and the storage helper skips already-invalidated rows so replays are no-ops. The repeated hex-encode/invalidate/push branches collapsed into one shared dispatch, `MarmotApp::projection_update_for_invalidation_event`, which the sync loop calls once per event — covering the dedup suggestion as a side effect.

crates/traits/src/engine.rs
```diff
@@ -195,6 +195,19 @@ pub enum AppMessageInvalidationReason {
     UndecryptableInCanonicalState,
 }
 
+/// Why the state notifications attributed to a commit were withdrawn
+/// (spec `protocol-core/convergence.md` "Applying the selected branch"). The
+/// state-notification counterpart of [`AppMessageInvalidationReason`], carried
+/// by [`GroupEvent::GroupStateInvalidated`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
erskingardner: Fixed in ed4975c: `#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]`.

crates/cgka-engine/src/distributed_convergence.rs
erskingardner: You were right that this path was broken, and the new regression test proved it — but the failure mechanism turned out to be one level deeper than an id mismatch.

**What the ids actually do (verified with a production-shaped peeler that mints a fresh transport id per wrap):** the ids are consistent per device on every path that can name a commit, because the own sent commit's stored wire record keeps `wrapped.id` as its record/inner id (`persist_openmls_wire_message`), and that record — not the content-id marker row, which is RawTransport and excluded from the candidate graph — is what enters stored convergence. So a drop of the own commit would carry `wrapped.id`, matching the confirm-time stamp; inbound commits are content-id keyed on both their emit and drop paths. `persist_sent_openmls_content_marker` exists for OwnEcho classification, not attribution.

**The real gap:** the own commit never got dropped at all. MLS cannot `process_message` its own commit, so the candidate BFS fails to materialize the own branch (`Materialized(None)`), the commit receives no disposition, `emit_rolled_back_commits` never sees it, and the confirm-time rows survive — exactly the stale-row outcome you predicted, reachable after a restart clears the in-memory `committed_from` guard and routes the same-epoch sibling into stored convergence instead of the direct seam.

**Fix (ed4975c):** a new `emit_superseded_processed_commits` pass after each convergence apply implements the spec sentence directly — every stored `Processed` commit ("previously applied") at/above the selected fork epoch that is not on the accepted path is marked `EpochInvalidated` and withdrawn via `CommitRolledBack` + `GroupStateInvalidated`. Acceptance is checked under both id aliases (record id and content dedup id) so the transport/content split can never cause a false supersession, and it dedupes against the canonicalization drop set so materializable losing commits are still reported exactly once. No re-stamping of `origin_commit_id` was needed: attribution stays per-device-consistent as-is, which the new tests pin.

crates/cgka-engine/tests/update_group_data.rs
erskingardner: Added in ed4975c: `EphemeralIdPeeler` mints a fresh transport id on every wrap (asserted distinct from `SHA-256(mls_bytes)`), and the concurrent-rename scenario now runs under both peelers via a shared body (`concurrent_rename_withdrawal_survives_distinct_transport_and_content_ids`). The stronger pin is `rebuilt_engine_convergence_withdraws_own_confirmed_rename_by_stamped_origin`: the ordering-key loser confirms its rename, restarts (dropping the in-memory `committed_from` guard), and receives the winning sibling through stored convergence — the withdrawal must name the wrap-time transport id its rows were stamped with. That test fails without the new `emit_superseded_processed_commits` engine pass (the own commit previously got no disposition at all — see the thread on distributed_convergence.rs) and passes with it.

crates/marmot-app/src/client/sync.rs
erskingardner: Added in ed4975c: the sync loop's invalidation dispatch is extracted into `MarmotApp::projection_update_for_invalidation_event` (unit-testable without transport plumbing), and `group_state_invalidated_event_tombstones_origin_commit_system_rows` in marmot-app drives it against real SQLCipher-backed storage: two persisted kind-1210 system rows stamped with different origin commits, a `GroupStateInvalidated` event tombstones exactly the stamped row with the withdrawal reason, the other row stays live, a replayed event is a projection no-op, and `CommitRolledBack` no longer dispatches (the explicit withdrawal event is authoritative).

erskingardner:
Context: this PR (#363) is an early instance of the convergence-seam contract now tracked in #707 — withdrawing a superseded commit's state notifications is exactly the kind of invariant that must hold identically on the direct and stored-convergence/replay seams.

PR #703, opened 2026-07-03
Surface self-eviction instead of swallowing UseAfterEviction (SelfEvicted / realizing removal)

### LINKED ISSUE #376: Silent eviction: removed member is never notified when the removal Commit isn't applied (UseAfterEviction swallowed)
## Summary

When an admin removes a member of a group, the removed member on a **separate device** is never told it was evicted: the conversation stays fully usable (composer enabled, group reads as active) and `groupMembers()` still returns self. The eviction is invisible.

This is the engine-side dependency behind the app-layer composer-disable work — the UI gate is already correct; it simply never receives an eviction signal across devices.

## Reproduction

- Admin (device 1) and member B (device 2, separate engine) are in group G.
- Admin removes B.
- **Same device** (B is a second account on device 1): switching to B correctly shows the removed state — the removal Commit is applied locally, so the notify path fires.
- **Cross device** (B on device 2): B's group state never changes. B keeps showing the group as active and (per its own engine) still a member. No removal event, no group-state change.

Confirmed on-device: the removed account kept receiving `NEW_LAST_MESSAGE` in the group after removal — i.e. its core never advanced past / applied the removal Commit, and nothing surfaced the eviction.

## Expected (per spec)

`data_flows.md` "Add/Remove Members" mandates the removed member's flow:

```
A->>R: Publish Commit (445)
R->>N: Deliver to removed member
N->>N: Process Commit, Realize removed, Delete group state
```

The removed member is supposed to realize it was removed and tear down group state.

## Root cause

There are two receive-side outcomes, and **only one notifies the app**:

1. **Removal Commit is processed & merged** → roster diff sees self in `removed` → `GroupStateChanged { MemberRemoved }` is emitted → the app marks self-membership and disables the composer.
   `crates/cgka-engine/src/message_processor/ingest.rs:815-851`

2. **Removal Commit is missed, and the member instead hits a later post-eviction message** → OpenMLS returns `UseAfterEviction` → the engine marks the message `Failed` and returns `Stale { PeelFailed }`, **emitting no event**:

   ```rust
   // crates/cgka-engine/src/message_processor/ingest.rs:560
   Err(ProcessMessageError::GroupStateError(MlsGroupStateError::UseAfterEviction)) => {
       self.update_stored_message_state(&msg.id, MessageState::Failed)?;
       return Ok(IngestOutcome::Stale { reason: StaleReason::PeelFailed });
   }
   ```

`UseAfterEviction` is OpenMLS authoritatively stating "you have been evicted from this group." The engine discards that signal. Outcome marmot-protocol/mdk#694 only fires if the removed client applies the *specific* removal Commit; when relay timing / ordering means it doesn't apply that Commit and only later sees a post-eviction message, the eviction is swallowed and the client is never told. There is no fallback that treats "I can no longer process this group because I'm evicted" as a membership change.

The publish side is fine: the Commit routes to the group's relays (`route_wrapped_group_message` → `group_routes` from `state.groups`, `crates/marmot-account/src/routing.rs:113`), which the removed member still subscribes to.

## Proposed fix

At `ingest.rs:560`, on `UseAfterEviction` for a group, surface a self-eviction `GroupStateChange` (mark self removed / group inactive) before returning `Stale`, instead of silently dropping it. That gives the app the eviction signal even when the original removal Commit was never applied — the cross-device reality — turning silent eviction into an observed one.

Add a regression test: a member that misses the removal Commit and then ingests a post-eviction message must emit a self-removal `GroupStateChange` (and the group must read as inactive), not a silent `Stale`.

## Notes

- Forward secrecy itself is intact (the evicted client cannot decrypt new-epoch content); this is a missing **eviction notification / state teardown**, not a confidentiality break.
- A secondary, murkier question (why the removal Commit sometimes never applies on the second device — relay delivery vs. epoch ordering/buffering) likely also contributes and warrants a relay-level look, but the swallowed `UseAfterEviction` is a self-contained, reproducible gap on its own.

erskingardner:
## Summary

Fixes the silent-eviction gap from #376: when the local retained canonical state records the member's **own removal** — OpenMLS group state `Inactive`, surfaced as `MlsGroupStateError::UseAfterEviction` — the engine classified later group input as generic `Stale { PeelFailed }`, emitted no event, and never marked the group inactive. A cross-device removed member kept seeing the group as fully active forever.

This implements the merged spec behavior from marmot-protocol/marmot#171:

- **`foundation/errors.md` — `SelfEvicted` outcome** (disposition `stale`, category `stale_epoch`): terminal for the group on that client; attaches to later input for a group whose retained canonical state records the local member's own removal.
- **`protocol-core/member-departure.md` — "Realizing removal"**: a state-derived, idempotent obligation. Whenever retained canonical state records the local member's removal and the local copy is not yet marked removed, the client MUST emit a self-removed state notification and mark the copy removed. The removed copy is retained inactive, not deleted. Failure to decrypt alone is NOT evidence of removal.
- **`protocol-core/inbound-processing.md`**: processing such input must surface the removal as a state notification instead of failing silently.

## What changed

- **`crates/traits/src/ingest.rs`** — new `StaleReason::SelfEvicted` variant (unit variant; group scoping travels elsewhere, matching the existing style).
- **`crates/traits/src/group.rs`** — new `Group.removed: bool` (`#[serde(default)]`, so pre-existing serialized records read back as `removed: false`). This is the "local copy is marked removed" realization marker: the record and history are retained, the group is just never presented as active again. No prior inactive/removed representation existed, so this is the smallest coherent one.
- **`crates/cgka-engine/src/message_processor/ingest.rs`** — both authenticated-eviction sites now realize instead of swallowing:
  - the `!mls_group.is_active()` front gate (the persisted OpenMLS `Inactive` state — the site all later input actually hits), and
  - the `UseAfterEviction` arm from `process_message` (kept as the defense-in-depth handler bound to the OpenMLS signal itself).

  Both call the new `realize_self_eviction`: if the copy is not yet marked removed, emit `GroupStateChanged { change: MemberRemoved { member: self }, actor: None }` via the existing `push_group_state_change` machinery, mark the record removed, clear any leave-request state; then return `Stale { SelfEvicted }`. Already-realized copies return `SelfEvicted` without re-emitting.
- **Marker coupling at the paths that already notify** — the direct commit-apply seam (roster diff) and the convergence reorg diff set `removed = true` in the same pass that emits the self-departure notification, so the marker and the notification never diverge and realized removals never double-notify.
- **`crates/cgka-engine/src/audit_helpers.rs`** — `stale_reason_str` maps the new reason to the stable low-cardinality string `"self_evicted"` for forensic audit rows.
- **Consumers** — `cgka-session`, `marmot-account`, and the conformance simulator match only on outcome kind (`Processed`/`Buffered`/`Stale`), so `SelfEvicted` correctly flows as `Stale` and never enters convergence buffering; verified no reason-specific branches exist. `IngestOutcome` is not exposed through `marmot-uniffi` (verified — no FFI change).
- Mechanical `removed: false` additions at `Group` literal construction sites (engine group lifecycle, storage/simulator/account test fixtures) and updated `cgka-traits` insta snapshots.

## Guarded semantics

Only authenticated evidence triggers realization: the OpenMLS group state records the eviction (set exclusively by merging a commit that removed our own leaf). Generic decrypt failures / missed commits keep their existing missing-history behavior (peel-deferred / convergence-buffered) — covered by a dedicated guard test.

## Test plan

New tests in `crates/cgka-engine/tests/invite_leave.rs`:

- `removed_member_applying_removal_commit_marks_local_copy_removed` — applying the removal commit emits the departure and durably marks the copy removed.
- `post_eviction_message_realizes_self_removal_and_returns_self_evicted` — **the #376 regression**: a copy whose canonical state records eviction but which never realized it (silent-eviction state) ingests a later post-eviction message → outcome is `Stale { SelfEvicted }`, a self-removed `GroupStateChanged` is emitted, and the record reads removed.
- `second_post_eviction_message_is_self_evicted_without_duplicate_notification` — idempotence: repeated post-eviction input keeps classifying `SelfEvicted` with no duplicate notification.
- `missed_removal_commit_without_evidence_is_not_self_evicted` — guard: a member that merely missed the removal commit (no authenticated evidence) stays buffered/repair, is not marked removed, and gets no fabricated notification.

Ran locally:

- `just fast-ci` (fmt-check, workspace check incl. OTLP builds, clippy) — clean
- `cargo test -p cgka-traits` (66 + snapshot suites) — pass
- `cargo test -p cgka-engine` (full suite incl. the 4 new tests) — pass
- `cargo test -p cgka-session`, `-p marmot-account`, `-p storage-sqlite`, `-p marmot-app`, `-p cgka-conformance-simulator` — pass

Closes #376

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/703">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added a persisted “removed” marker to local group records to better represent self-eviction.
  * Ingestion now classifies post-eviction traffic with a dedicated **SelfEvicted** stale outcome.
* **Bug Fixes**
  * Self-eviction realization is now idempotent and reliably reconciles persisted roster/state.
  * Outbound sends and queued intent drains are blocked/cleared for removed groups to prevent repeated failures.
* **Tests**
  * Expanded self-eviction, queue-purge, and snapshot coverage, including updated CLI failure assertions.

### DIFF
--- a/crates/cgka-engine/src/message_processor/ingest.rs
+++ b/crates/cgka-engine/src/message_processor/ingest.rs
@@ -1303,6 +1303,19 @@ impl<S: StorageProvider> Engine<S> {
     /// is kept inactive, never deleted). Callers must only invoke this on
     /// authenticated evidence of eviction (the local MLS group state records
     /// our removal), never on a bare decrypt failure.
+    ///
+    /// Ordering: the notification is enqueued BEFORE the durable marker write,
+    /// so a crash between the two re-runs realization on the next
+    /// `SelfEvicted` input (at worst a duplicate notification, absorbed by the
+    /// app's canonical-row-id upsert) instead of leaving a durable marker that
+    /// suppresses a notification nobody observed. A window this ordering
+    /// cannot close remains: `events_buf` is in-memory, so a process death
+    /// after the marker write but before the caller drains and persists the
+    /// event still loses it. That at-most-once property is shared by every
+    /// `GroupStateChanged` emission in the engine today (the commit-apply seam
+    /// has the same shape for all roster/admin/profile rows); making the event
+    /// channel durable — or reconciling on session open from `Group::removed`
+    /// at the app layer — is a systemic follow-up, not realization-specific.
     pub(crate) fn realize_self_eviction(
         &mut self,
         group_id: &GroupId,
@@ -1316,9 +1329,14 @@ impl<S: StorageProvider> Engine<S> {
         if group.removed {
             return Ok(());
         }
-        group.removed = true;
-        self.storage.put_group(&group)?;
-        self.clear_leave_request_state(group_id)?;
+        // OpenMLS's `Inactive` state does not record WHY our leaf left the
+        // tree. The durable leave request is local, authenticated intent to
+        // leave, so when one is pending the departure being realized is our
+        // own leave: attribute it as `MemberLeft` (actor = self), matching the
+        // direct seam's SelfRemove classification. Without one the removal is
+        // attributed as involuntary with an unknown actor — the accepted
+        // degradation when the removing commit itself was never applied.
+        let had_leave_request = self.load_leave_request_state(group_id)?.is_some();
         // Privacy-safe breadcrumb: aggregate signal only, no group/member ids
         // (observability.md).
         tracing::info!(
@@ -1326,18 +1344,27 @@ impl<S: StorageProvider> Engine<S> {
             method = "realize_self_eviction",
             "realized own removal for a group whose canonical state records eviction"
         );
-        // Actor unknown: the removing commit was not (or could not be)
-        // attributed here — realization derives from retained state, not from
-        // applying the specific removal commit.
-        self.push_group_state_change(
-            group_id,
-            epoch,
-            None,
-            GroupStateChange::MemberRemoved {
-                member: self.identity.self_id().clone(),
-            },
-            None,
-        );
+        let member = self.identity.self_id().clone();
+        let (actor, change) = if had_leave_request {
+            (
+                Some(member.clone()),
+                GroupStateChange::MemberLeft { member },
+            )
+        } else {
+            (None, GroupStateChange::MemberRemoved { member })
+        };
+        self.push_group_state_change(group_id, epoch, actor, change, None);
+        // Marker + roster reconciliation in one durable write: a removed copy
+        // must not keep presenting self as a member, so roster-gated callers
+        // (`members()`, hydrate restore checks, backfill) cannot disagree with
+        // the terminal marker. The seam and convergence paths already write
+        // the post-removal roster; this reconciles the pathological copy whose
+        // roster was never mirrored.
+        group.removed = true;
+        let self_id = self.identity.self_id().clone();
+        group.members.retain(|member| member.id != self_id);
+        self.storage.put_group(&group)?;
+        self.clear_leave_request_state(group_id)?;
         Ok(())
     }
 
--- a/crates/cgka-engine/src/message_processor/ingest.rs
+++ b/crates/cgka-engine/src/message_processor/ingest.rs
@@ -854,6 +854,10 @@ impl<S: StorageProvider> Engine<S> {
                         .any(|member| member == self.identity.self_id())
                     {
                         self.clear_leave_request_state(&group_id)?;
+                        // The copy just became removed: purge queued outbound
+                        // intents so later drains do not re-fail them forever
+                        // against the removed-copy send gate.
+                        self.discard_queued_outbound_intents_for_removed_group(&group_id)?;
                     } else if after_ids.contains(self.identity.self_id()) {
                         if self.load_leave_request_state(&group_id)?.is_some() {
                             // A SelfRemove proposal is valid only in its
@@ -1364,6 +1368,11 @@ impl<S: StorageProvider> Engine<S> {
         let self_id = self.identity.self_id().clone();
         group.members.retain(|member| member.id != self_id);
         self.storage.put_group(&group)?;
+        // A removed copy must never publish: drop any outbound intents that
+        // were durably queued before the removal was realized, instead of
+        // leaving them to re-fail through the removed-copy send gate on every
+        // later drain.
+        self.discard_queued_outbound_intents_for_removed_group(group_id)?;
         // Deliberately LAST, after the marker write — not before it like the
         // convergence path (which has no attribution read). The notification
         // is already enqueued above, so a failure here cannot lose it; it only
--- a/crates/cgka-engine/src/message_processor/ingest.rs
+++ b/crates/cgka-engine/src/message_processor/ingest.rs
@@ -127,14 +127,23 @@ impl<S: StorageProvider> Engine<S> {
             // group returns to Stable.
             let current_epoch = EpochId(mls_group.epoch().as_u64());
             if !mls_group.is_active() {
+                // The persisted OpenMLS group state is `Inactive`: retained
+                // canonical state records our own removal. This is
+                // authenticated evidence (a merged commit removed our leaf),
+                // not a decrypt failure, so later input for this group is
+                // `SelfEvicted` and must perform "realizing removal" when not
+                // already done (member-departure.md), instead of classifying
+                // as ordinary stale traffic while the group still presents as
+                // active.
                 self.persist_transport_message(
                     msg,
                     &group_id,
                     current_epoch,
                     MessageState::Failed,
                 )?;
+                self.realize_self_eviction(&group_id, current_epoch)?;
                 return Ok(IngestOutcome::Stale {
-                    reason: StaleReason::PeelFailed,
+                    reason: StaleReason::SelfEvicted,
                 });
             }
             if !self.epoch_manager.can_ingest(&group_id) {
@@ -558,9 +567,19 @@ impl<S: StorageProvider> Engine<S> {
                     });
                 }
                 Err(ProcessMessageError::GroupStateError(MlsGroupStateError::UseAfterEviction)) => {
+                    // OpenMLS states authoritatively that our own leaf was
+                    // removed (the group state records eviction). Surface the
+                    // removal instead of silently dropping the message: this
+                    // is the `SelfEvicted` outcome, and processing it must
+                    // perform "realizing removal" when not already done
+                    // (member-departure.md). Normally the `is_active` gate
+                    // above classifies this first; this arm keeps the
+                    // realization obligation attached to the authenticated
+                    // OpenMLS signal itself.
                     self.update_stored_message_state(&msg.id, MessageState::Failed)?;
+                    self.realize_self_eviction(&group_id, current_epoch)?;
                     return Ok(IngestOutcome::Stale {
-                        reason: StaleReason::PeelFailed,
+                        reason: StaleReason::SelfEvicted,
                     });
                 }
                 Err(e) => {
@@ -766,6 +785,19 @@ impl<S: StorageProvider> Engine<S> {
                         crate::group_lifecycle::mirror_app_components_into_record(
                             &mls_group, &mut g,
                         );
+                        // This commit removed our own leaf: mark the local
+                        // copy removed (member-departure.md, "Realizing
+                        // removal") in the same record write. The departure
+                        // notification for self is emitted with the roster
+                        // diff below, so the marker and the notification stay
+                        // coupled and later `SelfEvicted` input does not
+                        // re-emit it.
+                        if removed
+                            .iter()
+                            .any(|member| member == self.identity.self_id())
+                        {
+                            g.removed = true;
+                        }
                         self.storage.put_group(&g)?;
                     }
                     // #740 rotation: a peer's applied commit may have changed the
@@ -1261,6 +1293,52 @@ impl<S: StorageProvider> Engine<S> {
         });
         pending_commit_guard.disarm();
         Ok(true)
+    }
+
+    /// Realize our own removal from `group_id` (member-departure.md,
+    /// "Realizing removal"): emit the self-removed state notification and mark
+    /// the local group copy removed. State-derived and idempotent — when the
+    /// copy is already marked removed the notification was already surfaced,
+    /// so this is a no-op. The record and its history are retained (the copy
+    /// is kept inactive, never deleted). Callers must only invoke this on
+    /// authenticated evidence of eviction (the local MLS group state records
+    /// our removal), never on a bare decrypt failure.
+    pub(crate) fn realize_self_eviction(
+        &mut self,
+        group_id: &GroupId,
+        epoch: EpochId,
+    ) -> Result<(), EngineError> {
+        let mut group = match self.storage.get_group(group_id) {
+            Ok(group) => group,
+            Err(StorageError::NotFound) => return Ok(()),
+            Err(err) => return Err(EngineError::Storage(err)),
+        };
+        if group.removed {
+            return Ok(());
+        }
+        group.removed = true;
+        self.storage.put_group(&group)?;
+        self.clear_leave_request_state(group_id)?;
+        // Privacy-safe breadcrumb: aggregate signal only, no group/member ids
+        // (observability.md).
+        tracing::info!(
+            target: "cgka_engine::message_processor",
+            method = "realize_self_eviction",
+            "realized own removal for a group whose canonical state records eviction"
+        );
+        // Actor unknown: the removing commit was not (or could not be)
+        // attributed here — realization derives from retained state, not from
+        // applying the specific removal commit.
+        self.push_group_state_change(
+            group_id,
+            epoch,
+            None,
+            GroupStateChange::MemberRemoved {
+                member: self.identity.self_id().clone(),
+            },
+            None,
+        );
+        Ok(())
     }
 
     fn mark_raw_transport_message_failed_if_deferred(
--- a/crates/cgka-engine/src/message_processor/ingest.rs
+++ b/crates/cgka-engine/src/message_processor/ingest.rs
@@ -1364,6 +1364,16 @@ impl<S: StorageProvider> Engine<S> {
         let self_id = self.identity.self_id().clone();
         group.members.retain(|member| member.id != self_id);
         self.storage.put_group(&group)?;
+        // Deliberately LAST, after the marker write — not before it like the
+        // convergence path (which has no attribution read). The notification
+        // is already enqueued above, so a failure here cannot lose it; it only
+        // leaves a stale leave-request row, which the removed-copy send gate
+        // supersedes and re-join cleans up. Clearing BEFORE the marker write
+        // would be worse: a failed marker write would then retry against
+        // already-cleared leave state and emit a differently-attributed
+        // duplicate (`MemberRemoved` after an original `MemberLeft`), i.e. a
+        // second visible row. With this order any retry duplicate is
+        // byte-identical and collapses in the app's canonical-row-id upsert.
         self.clear_leave_request_state(group_id)?;
         Ok(())
     }
--- a/crates/traits/src/group.rs
+++ b/crates/traits/src/group.rs
@@ -19,6 +19,18 @@ pub struct Group {
     pub epoch: EpochId,
     pub members: Vec<Member>,
     pub required_capabilities: GroupCapabilities,
+    /// The local copy of this group is marked removed: retained canonical
+    /// state records the local member's own removal (spec
+    /// `protocol-core/member-departure.md`, "Realizing removal"). The record
+    /// is retained inactive — history may be kept, but the group must not be
+    /// presented as active and nothing may be sent or published to it. This
+    /// flag is the idempotence marker for the realization obligation: it is
+    /// set together with the self-removed state notification, so later input
+    /// classified `SelfEvicted` does not re-emit the notification. Terminal
+    /// for the group on this client. Defaults to `false` for records
+    /// persisted before this field existed.
+    #[serde(default)]
+    pub removed: bool,
 }
 
 /// One member of a group, as storage sees it.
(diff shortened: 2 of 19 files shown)


erskingardner (commented):
## Adversarial review

Solid fix for the #376 silent-eviction gap: authenticated inactive MLS state now maps to `SelfEvicted`, realization is idempotent, and the guard test correctly keeps bare decrypt failures out of this path.

I left inline notes on a few durability / terminal-semantics gaps (marker-before-event ordering, `MemberRemoved` vs voluntary leave UX, stale `members` coexisting with `removed`, and outbound send not consulting the new flag). None block the core classification fix, but they're worth addressing or tracking before this is the only recovery path in production.

erskingardner (commented):
## Adversarial review

Solid fix for the #376 silent-eviction gap: authenticated inactive MLS state now maps to `SelfEvicted`, realization is idempotent, and the guard test correctly keeps bare decrypt failures out of this path.

I left inline notes on a few durability / terminal-semantics gaps (marker-before-event ordering, `MemberRemoved` vs voluntary leave UX, stale `members` coexisting with `removed`, and outbound send not consulting the new flag). None block the core classification fix, but they are worth addressing or tracking before this is the only recovery path in production.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -1263,6 +1295,52 @@ impl<S: StorageProvider> Engine<S> {
...
+            Ok(group) => group,
+            Err(StorageError::NotFound) => return Ok(()),
+            Err(err) => return Err(EngineError::Storage(err)),
+        };
+        if group.removed {
+            return Ok(());
+        }
+        group.removed = true;
+        self.storage.put_group(&group)?;
+        self.clear_leave_request_state(group_id)?;
```
erskingardner: **Durability ordering vs idempotence:** `put_group(removed = true)` happens before `push_group_state_change`. If the process dies after the durable write but before the caller drains `events_buf`, a restart sees `removed = true` and later `SelfEvicted` input will skip re-notification. That makes the marker a suppression flag, not just an idempotence guard.

Same pattern exists on the commit-apply seam, and `cgka-session::ingest` drains in-process today, so the window is small — but any future async effect pipeline would turn this into a silent UI miss. Consider either (a) persisting the marker only after the app projection acknowledges the departure, or (b) on hydrate, if `group.removed` and OpenMLS `!is_active()`, re-emit the notification once.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -1263,6 +1295,52 @@ impl<S: StorageProvider> Engine<S> {
...
+            "realized own removal for a group whose canonical state records eviction"
+        );
+        // Actor unknown: the removing commit was not (or could not be)
+        // attributed here — realization derives from retained state, not from
+        // applying the specific removal commit.
+        self.push_group_state_change(
+            group_id,
+            epoch,
+            None,
+            GroupStateChange::MemberRemoved {
```
erskingardner: **Departure semantics for cross-device self-remove:** Realization always emits `MemberRemoved { self }` with `actor: None`. In `marmot-app`, that maps to `SelfMembership::Removed` (involuntary eviction), not `MemberLeft` / `SelfMembership::Left`. For a user who self-removed on device A, device B will label the chat as removed-by-someone rather than you-left.

That may match the spec actor-unknown bucket, but it is a product-visible misclassification worth an explicit decision (or a dedicated change variant).

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -1263,6 +1295,52 @@ impl<S: StorageProvider> Engine<S> {
...
+    ) -> Result<(), EngineError> {
+        let mut group = match self.storage.get_group(group_id) {
+            Ok(group) => group,
+            Err(StorageError::NotFound) => return Ok(()),
+            Err(err) => return Err(EngineError::Storage(err)),
+        };
+        if group.removed {
+            return Ok(());
+        }
+        group.removed = true;
```
erskingardner: **`Group.removed` does not reconcile `members`:** Realization sets the flag but does not remove `self` from `group.members` or refresh from OpenMLS. After realization you can end up with `removed = true` while `members()` still lists the local account (the regression test simulates exactly this).

Downstream code that gates on roster membership (e.g. `leave_request_to_restore_on_hydrate`, backfill keyed off `runtime.members()`) may disagree with the terminal `removed` marker. Consider stripping self from `members` in the same write.

crates/cgka-engine/src/message_processor/ingest.rs
erskingardner: **Outbound terminal semantics:** `clear_leave_request_state` drops the leave send gate. `do_send` only checks `has_leave_send_gate`, not `group.removed` or OpenMLS inactive state (`send.rs::do_send_app_message`). Eviction likely fails at `create_message` today, but as an opaque `Backend` error rather than deterministic terminal `InvalidTransition`.

Since this PR introduces a durable terminal marker, consider having `do_send` (and hydrate) treat `group.removed` like the leave gate.

crates/traits/src/group.rs
```diff
@@ -19,6 +19,18 @@ pub struct Group {
...
+    /// state records the local member's own removal (spec
+    /// `protocol-core/member-departure.md`, "Realizing removal"). The record
+    /// is retained inactive — history may be kept, but the group must not be
+    /// presented as active and nothing may be sent or published to it. This
+    /// flag is the idempotence marker for the realization obligation: it is
+    /// set together with the self-removed state notification, so later input
+    /// classified `SelfEvicted` does not re-emit the notification. Terminal
+    /// for the group on this client. Defaults to `false` for records
+    /// persisted before this field existed.
+    #[serde(default)]
```
erskingardner, line 32: **No upper-layer consumer yet:** `Group.removed` is written in three engine paths but nothing above `cgka-engine` reads it — terminal UX still depends entirely on the ephemeral `GroupStateChanged` event reaching `marmot-app` projection. If intentional for this PR, a follow-up to wire hydration/backfill off `removed` would close the restart half of #376.

crates/cgka-engine/src/message_processor/ingest.rs
erskingardner: Fixed the part that's fixable at this layer, documented the rest (5aa7b19).

Reordered: `realize_self_eviction` now enqueues the notification **before** the durable marker write. A crash between the two now re-runs realization on the next `SelfEvicted` input — at worst a duplicate notification, which the app's projection absorbs (kind-1210 rows key on `group_system_canonical_id` and `record_app_event` upserts on `ON CONFLICT(group_id_hex, message_id_hex)`), instead of a durable marker suppressing a notification nobody observed. Fail-open to re-notify rather than fail-silent.

On the two options: (a) an app-ack is the only complete fix but there is no ack channel today; (b) hydrate re-emit is not safe as proposed — the marker is also set by the seam/convergence paths, whose original notifications carry a *different* canonical identity (attributed actor, or `MemberLeft`), so a blind `MemberRemoved/actor:None` re-emit on every restart would add a spurious duplicate row for every healthy removed group, and the engine cannot know which identity was originally emitted.

The residual window (process death after the marker write but before the caller persists the drained event) is documented in the helper doc. Worth being precise about its scope: `events_buf` is in-memory and drained synchronously inside the same `ingest_delivery` call, and this at-most-once property is shared by **every** `GroupStateChanged` emission today — the commit-apply seam has the identical shape for all roster/admin/profile rows (merge is durable, events are not). So the complete fix is systemic (durable event outbox, or app-side reconciliation on session open reading `Group.removed` — see thread on `group.rs`), and I'd rather track it as a follow-up than bolt a realization-only ack into this PR.

crates/cgka-engine/src/message_processor/ingest.rs
erskingardner: Implemented the cheap improvement you suggested (5aa7b19): `realize_self_eviction` now reads the durable leave-request state **before** clearing it. A pending leave request is authenticated local intent to leave, so realization emits `MemberLeft { self }` with `actor: Some(self)` — matching the direct seam's SelfRemove classification — and only falls back to `MemberRemoved { self }, actor: None` when no leave intent exists. Test: `realization_with_pending_leave_request_attributes_member_left`.

The residual limitation is now documented in code: OpenMLS's `Inactive` state does not record *why* the leaf left the tree, so a member who self-removed on device A and whose device B has no local leave request (the cross-device case where the leave was initiated elsewhere) still realizes as involuntary `MemberRemoved` with unknown actor. That's accepted degraded attribution — the same path-dependence the harness already documents for convergence reorgs (`emits_departure_of` accepts either variant for exactly this reason). Distinguishing it fully would require the removal commit's proposal contents, which is the thing realization by definition never saw.

crates/cgka-engine/src/message_processor/ingest.rs
erskingardner: Fixed (5aa7b19): realization now strips self from `group.members` in the same `put_group` write that sets the marker, so roster-gated callers can never observe `removed = true` with self still in the roster. Assertion added to the regression test.

Checked the other two marker sites as suggested — both already write the post-removal roster in the same pass: the direct seam sets `g.members = after_members` from the merged tree, and the convergence path's record is rewritten from `final_members` by `update_group_record_from_replay` before the marker block runs. So the realization path was the only one that could diverge.

crates/cgka-engine/src/message_processor/ingest.rs
erskingardner: Fixed (5aa7b19): a copy marked removed is now a terminal outbound gate. `do_send` rejects **every** intent kind (including `Leave` — there is nothing left to leave) with a deterministic `EngineError::InvalidTransition { from: "Removed", reason: "local group copy is marked removed (self-evicted)" }`, checked before the queueing branch so nothing gets queued either. The same check sits at the top of `do_send_ready`, so queued-intent drains for a group that became removed after queueing hit the same terminal error instead of an opaque `UseAfterEviction` backend error.

Hydrate needs no change: the gate reads the record directly, and the marker's lifecycle is already correct across restart — it only clears through an authenticated re-join (`do_join_welcome` rebuilds the record with `removed: false` after `clear_live_openmls_group`), which is exactly when sends should become legal again.

Test: `send_after_realized_eviction_is_rejected_terminally` (AppMessage and Leave both → `InvalidTransition`, not `Backend`).

crates/traits/src/group.rs
erskingardner: Partially narrowed in 5aa7b19, remainder is a deliberate follow-up.

What reads `removed` now, inside the engine: (1) the realization idempotence check; (2) the terminal outbound gate in `do_send` / `do_send_ready` (every intent kind → deterministic `InvalidTransition`); (3) roster reconciliation guarantees the flag and `members` can't disagree.

What still doesn't read it: anything above cgka-engine. Terminal *UX on a running client* does not depend on the flag — it flows through the (now reliably emitted) `GroupStateChanged` event, which marmot-app already projects to a durable kind-1210 row, so restart UX is covered by that persisted row in the normal case. The real remaining gap is the one you named: marmot-app's group listing/open path should read `Group.removed` on session open as a state-derived backstop (which would also close the residual crash window from the durability thread — a lost event no longer matters if the app reconciles from the marker). That's app-layer surface (projection + runtime + likely uniffi exposure) and belongs in its own PR; I'd suggest tracking it as a follow-up issue: "marmot-app: surface Group.removed on session open (state-derived eviction backstop)". Happy to file it.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -1263,6 +1295,52 @@ impl<S: StorageProvider> Engine<S> {
...
+        // applying the specific removal commit.
+        self.push_group_state_change(
+            group_id,
+            epoch,
+            None,
+            GroupStateChange::MemberRemoved {
+                member: self.identity.self_id().clone(),
+            },
+            None,
+        );
```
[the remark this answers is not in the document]
erskingardner: Verified against the current code — the specific stranding this describes was already removed by 5aa7b19, and the remaining order is deliberate (now documented in-code in 21e0dd2).

Current order in `realize_self_eviction`:
1. `load_leave_request_state` (fallible **read** — required *before* emitting, it decides `MemberLeft` vs `MemberRemoved` attribution)
2. `push_group_state_change` (in-memory, infallible)
3. `put_group(removed = true)` (durable marker + roster reconciliation)
4. `clear_leave_request_state`

So the failure mode claimed here — marker durably set while the notification was never pushed — cannot occur: the push precedes the marker write, and no fallible step sits between them. Walking the actual failure points:
- **(1) fails:** nothing emitted, nothing marked → clean re-realization on next `SelfEvicted` input.
- **(3) fails:** notification already enqueued, marker unset, leave state intact → retry re-reads the *same* leave state and emits a byte-identical event (same canonical row id → app upsert collapses it), then retries the marker. Nothing lost.
- **(4) fails:** notification already enqueued *and* marker set → nothing lost; only a stale leave-request row remains, which the removed-copy send gate (checked before the leave gate) supersedes and re-join cleans up.

Why not mirror the convergence path's clear-before-mark order: that path has no attribution read. Here, clearing before the marker write would mean a failed marker write retries against already-cleared leave state and emits a **differently-attributed** duplicate (`MemberRemoved`/no actor after an original `MemberLeft`/self) — a different canonical row id, i.e. a second visible row. Keeping the clear last makes any retry duplicate identical and idempotent. Added a comment at the clear site so this ordering doesn't get 'fixed' later.

erskingardner:
CI note: the "CLI e2e (mock relay)" failure in `three_user_message_lifecycle_covers_invite_remove_and_later_delivery` was the intended consequence of the review-round outbound gate — a removed member's send now fails with the deterministic `invalid_transition` ("local group copy is marked removed (self-evicted)") instead of the old opaque `engine_error`. Updated the e2e expectation to assert the new code + message (the engine fix is unchanged), recorded the JSON error-shape change in the CLI `Unreleased` changelog, and verified the full CLI test binary (73/73) plus `just fast-ci` locally. A workspace-wide grep found no other suite asserting the old `engine_error` shape for post-eviction sends. (21e0dd2)

jgmontoya (commented):
I think it looks good overall.

My agent flagged this:

> ### Correctness/Operational: queued outbound intents on a realized-removed group retry forever and are never cleaned up
> 
> **crates/cgka-engine/src/message_processor/mod.rs:104-127** (`converge_and_drain_queued_outbound_intents`)
> 
> > ```rust
> > let result = self.do_send_ready(record.intent.clone()).await?;
> > self.storage.delete_queued_outbound_intent(&record.id)?;
> > ```
> 
> Once `realize_self_eviction` marks a group's local copy `removed`, `do_send_ready` (send.rs:23-40, added by
> this PR) unconditionally returns `Err(EngineError::InvalidTransition(...))` for that group. If any
> `SendIntent` was already durably queued for that group (via `queue_outbound_intent`, e.g. a message sent while
> the group was mid-convergence, just before the removal was realized), this loop's `?` propagates that error
> immediately — before `delete_queued_outbound_intent` runs — so the stuck record is never drained and never
> removed from storage.
> 
> Tracing the caller chain: `CgkaEngine::advance_convergence` → `crates/marmot-app/src/client/sync.rs:289`
> (`advance_convergence_after_runtime_sync`) → `crates/marmot-app/src/runtime/account_worker.rs:507-530`. On
> `Err`, the worker calls `scheduled_convergence.schedule_retry_groups([...])`, which re-arms the timer with
> backoff and retries — forever, since the underlying condition (`group.removed == true`) is permanent by
> design ("Terminal for the group on this client"). The practical effect is an indefinite, low-frequency retry
> loop for that account that will always fail and always call `publish_app_runtime_account_error(...)`, for the
> lifetime of the account, for any group that happened to have a queued send at the moment its self-eviction was
> realized.
> 
> Given this PR's whole point is making self-eviction a clean, terminal state, it seems worth also draining
> (and discarding, or converting to a terminal `SendResult`/dropped-intent event) any `list_queued_outbound_intents`
> rows for the group inside `realize_self_eviction`, rather than leaving them to be discovered lazily and
> perpetually re-failed by the outbound drain path.
>

erskingardner:
Re: @jgmontoya's review — confirmed real and fixed in ec1e9fa. A durably queued outbound intent (send accepted mid-convergence just before the removal was realized) was a poison pill: `converge_and_drain_queued_outbound_intents` propagated the removed-copy `InvalidTransition` from `do_send_ready` *before* `delete_queued_outbound_intent`, so the record was never drained, and the app-layer scheduler (`schedule_retry_groups`) would retry the permanent failure forever.

Fixed on both sides:
- **Marker-site purge** (your suggestion, applied to *all three* sites that set `removed = true`, not just realization): `realize_self_eviction`, the commit-apply seam's self-removal branch, and the convergence reorg diff now discard every queued outbound intent for the group at the moment the copy becomes removed.
- **Drain-side terminal handling** (defense in depth for any ordering the purge misses): the drain checks `group_record_is_removed` *before dispatching*, discards any remaining queued records, and returns nothing to drain — so only the terminal removed state discards, and genuine transient send errors on healthy groups stay retryable (no error-shape matching involved).

Discard is silent toward the app — the self-removed `GroupStateChanged` notification already carries the user-facing signal and the spec says the removed copy MUST NOT publish — but each dropped intent leaves a forensic `Rejection` audit row (`queued_outbound_intent_discarded_group_removed`) plus an aggregate privacy-safe trace line.

Regression tests (`crates/cgka-engine/tests/invite_leave.rs`): `applying_removal_commit_purges_queued_outbound_intents` (queue → removal commit applies → queue empty) and `drain_on_removed_copy_discards_queued_intents_without_error` (realization-side purge, plus an intent queued *after* the marker → drain returns `Ok([])`, publishes nothing, queue empty). Full `cgka-engine` suite, `cgka-session`/`marmot-account`/`marmot-app`/simulator suites, and `just fast-ci` all green locally.

erskingardner:
Context: this PR (#376) is an early instance of the convergence-seam contract now tracked in #707 — realizing self-removal regardless of which seam delivered the input is one of the per-seam invariants #707 wants enforced at a shared chokepoint.

PR #704, opened 2026-07-03
Reject commits whose resulting GroupContext strips required app components

erskingardner:
## What

Adds a staged-commit validation, `validate_required_component_retention_for_staged_commit` (`crates/cgka-engine/src/app_components.rs`), that rejects — pre-merge — any commit whose resulting GroupContext:

- drops the `app_data_dictionary` extension entirely,
- drops the engine-owned `app_components` entry, or
- drops the state entry of any component in the group's current required set.

## Why

OpenMLS's draft-08 guard (`validate_app_data_update_proposals_and_group_context`) only validates the dictionary when the commit **also** carries `AppDataUpdate` proposals, and the engine's `validate_app_component_remove` only runs for `AppDataUpdate::Remove` operations. So an admin-authored **GroupContextExtensions-only** commit could replace the extensions with a set that omits the `app_data_dictionary` (or individual required entries) and merge cleanly.

The strip cannot orphan an admin key (an absent admin-policy component is the empty admin set — see the empty-set handling in `validate_admin_leaf_coupling_for_staged_commit` and the #701 review discussion), but after merging, the group becomes admin-less and frozen for every admin-required operation: a denial-of-service / consistency gap rather than a coupling bypass.

The rule uses the **current** epoch's required set, mirroring `validate_app_component_remove`: a component only becomes droppable after a prior commit removes it from the required list via an `AppDataUpdate` to `app_components`.

## Enforcement points

- `message_processor/ingest.rs` — direct staged-commit seam, alongside the admin-leaf-coupling check.
- `openmls_projection.rs` (`process_openmls_messages_inner`) — the convergence/replay path, returning `InvalidCommit`. This is the path that actually handles inbound same-epoch commits, since `ingest` routes commits with `msg_epoch >= current_epoch` through convergence.
- `message_processor/send.rs` — outbound remove-members path (defense-in-depth, same placement as the coupling check).
- `upgrade.rs` — the one outbound path that authors GCE commits, self-checked after staging.

## Tests

Three tests in `crates/cgka-engine/tests/distributed_convergence.rs`, built on a new `raw_group_context_extensions_strip_commit` helper (a raw OpenMLS `update_group_context_extensions` commit with no member changes):

- `ingest_rejects_group_context_commit_stripping_app_data_dictionary`
- `ingest_rejects_group_context_commit_dropping_required_component_entry` (dictionary kept, required admin-policy entry omitted)
- `convergence_rejects_group_context_commit_stripping_app_data_dictionary` — the pinned regression: via stored convergence the strip commit is dropped as `InvalidAgainstCandidateState`, epoch and admin set unchanged, message state `EpochInvalidated`.

Verified the tests pin the regression by temporarily disabling the projection-site check: all three then fail with the strip commit **accepted** (the pre-fix behavior).

## Notes for reviewers

- The ingest-path tests assert `IngestOutcome::Stale` + unchanged epoch/admins rather than `Err`: same-epoch commits route through convergence inside the input window, so the terminal disposition isn't persisted until the window closes. The pinned invariant is that the strip commit is never `Processed`.
- `cargo test -p cgka-engine` (18 test binaries) and `just fast-ci` both pass.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/704">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Added an app-component integrity check to reject staged commits that try to remove or tamper required app-component state (including the app data dictionary and mandatory entries) unless the changes are made through the allowed app data update operations.
  * Applied the validation across ingest, stored convergence/replay, send/remove-members flows, and upgrade staging to ensure failures don’t advance the epoch.

* **Tests**
  * Extended distributed convergence tests with GroupContextExtensions tampering scenarios (dictionary strip, required entry removal, and prohibited byte rewrites), verifying consistent rejection/invalidations and unchanged epoch/admin keys.

### DIFF
--- a/crates/cgka-engine/src/app_components.rs
+++ b/crates/cgka-engine/src/app_components.rs
@@ -16,9 +16,9 @@ use cgka_traits::error::EngineError;
 use cgka_traits::types::{GroupId, MemberId};
 use openmls::extensions::{AppDataDictionary, AppDataDictionaryExtension, Extension};
 use openmls::group::{MlsGroup, StagedCommit};
-use openmls::messages::proposals::Proposal;
+use openmls::messages::proposals::{AppDataUpdateOperation, Proposal};
 use openmls::prelude::{BasicCredential, LeafNode, Sender};
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) struct InitialComponentState {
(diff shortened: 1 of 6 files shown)


crates/cgka-engine/src/app_components.rs
```diff
@@ -297,6 +297,54 @@ pub(crate) fn validate_admin_leaf_coupling_for_staged_commit(
...
+    let resulting = resulting.dictionary();
+    // The negotiated required-component list is engine-owned and can never be
+    // dropped (mirrors `validate_app_component_remove`); required components
+    // must keep whatever state entry the current epoch carries. A component
+    // becomes droppable only after a prior commit removes it from the required
+    // list via an `AppDataUpdate` to `app_components`.
+    let mut protected = required_app_components_of_group(mls_group)?.ids;
+    protected.insert(APP_COMPONENTS_COMPONENT_ID);
+    for component_id in protected {
+        if current.get(&component_id).is_some() && resulting.get(&component_id).is_none() {
```
erskingardner: This still only protects presence, not contents. A raw `GroupContextExtensions` commit can keep `GROUP_ADMIN_POLICY_COMPONENT_ID` and the other required component IDs present while rewriting their bytes in `app_data_dictionary`, with no `AppDataUpdate` proposals. The OpenMLS draft-08 guard returns early when there are no app-data proposals, and our `process_commit_with_app_data_updates` component validation only runs for `AppDataUpdate` operations, so this still bypasses the component validators and can mutate or corrupt required state directly (for example, change admin-policy bytes as long as the resulting admins have leaves, or replace profile/routing/retention bytes with invalid data).

I think this check needs to reject direct GCE changes to protected entries, or otherwise validate/equality-check entries that were not changed through an `AppDataUpdate`, plus add a regression where the required entry is present but has different bytes.

erskingardner: Fixed in 24c09e2: renamed to `validate_app_component_integrity_for_staged_commit` and added an attribution rule — every dictionary entry that changes vs. the current epoch (added, rewritten, or removed) must match one of the commit's own `AppDataUpdate` operations, so a GCE cannot rewrite protected (or any) entries directly; added ingest + convergence regressions where the admin-policy entry is present but rewritten to a different well-formed admin set.

erskingardner:
Context: related to tracking issue #707 — rejecting commits whose GroupContext strips required components is another 'a validation that must run on *every* commit path' instance (here, closing the GroupContextExtensions-only bypass of the AppDataUpdate-gated checks).

PR #723, opened 2026-07-04
Materialize own confirmed commits as pre-validated convergence candidates (#706)

### LINKED ISSUE #706: Own published-and-confirmed commit can't materialize in stored convergence after restart — wrong branch selected; mutual committer restart forks the group permanently
### Summary

A device's OWN published-and-confirmed commit can never materialize as a candidate branch in stored convergence, because MLS refuses to process a device's own commit through `process_message`. While the engine process that published the commit stays alive, the in-memory `epoch_manager.committed_from` guard routes same-epoch siblings onto the direct fork-recovery seam, which handles the race correctly. But `committed_from` is populated only at send time and never rehydrated — so after an engine restart, a late-arriving same-epoch sibling routes into stored convergence, where the own commit's branch silently drops out of branch selection:

- `crates/cgka-engine/src/openmls_projection.rs::build_stored_openmls_candidate_paths` probes each candidate path by replaying it from the retained anchor; the own commit fails replay (`process_message` returns an error for own commits), the probe returns `Materialized(None)`, and the branch is pruned with no disposition.
- `crates/cgka-engine/src/epoch_manager.rs::committed_from` — in-memory only, populated at `begin_pending` / `record_committed_from`, never rebuilt from storage at hydrate.

### Consequences

1. **Restarted winner reorged off the canonical branch.** If the restarted device's own commit should WIN the deterministic same-epoch ordering (source epoch, priority, authenticated committer, digest — `CommitOrderingKey` / `convergence.rs::compare_scores`), the losing sibling is nevertheless the only candidate that materializes, so branch selection picks it. The device withdraws its own (canonical!) state notifications and diverges from every peer that resolved the same fork correctly on the direct seam.
2. **Permanent group fork when both concurrent committers restart.** Each device sees exactly one materializable candidate — the OTHER device's branch — so each deterministically reorgs onto the other's commit and the group forks permanently (A ends on B's rename, B on A's). Observed empirically during #702 work in a discarded test variant; now reproduced by `mutually_rebuilt_engines_converge_on_same_branch_after_concurrent_renames` in `crates/cgka-engine/tests/update_group_data.rs` (fails without the fix with exactly the swapped-branch fork).

PR #702 (issue #363) added `emit_superseded_processed_commits` so state notifications correctly track whatever the engine applies — but that only makes the *notifications* honest about the wrong branch; the selection-vs-own-commit problem itself remained.

### Root cause

Branch selection is only deterministic across replicas if every replica can score the same candidate set. The own commit is the one branch a replica can never rebuild by replay (MLS design: own commits are merged via `merge_pending_commit`, never `process_message`), and nothing durable carried its ordering metadata (priority, authenticated committer, consumed proposal refs) or a way to realize its post-state during candidate materialization.

### Fix (implemented)

Teach candidate materialization to treat own `Processed` commits as pre-validated branches, using durable state only:

1. **Confirm-time stamp** (`crates/traits/src/message.rs`): new `StoredMessagePayload::OwnCommitWire { message, stamp }` variant carrying `OwnCommitConvergenceStamp { committer, priority, consumed_proposal_refs }`. `do_confirm_published` captures the stamp while the staged commit is still attached and enriches the stored wire record in the same durable write that marks it `Processed`.
2. **Anchor rollforward during materialization** (`openmls_projection.rs::process_openmls_messages_inner`): when a replayed candidate path reaches a stamped own commit and every commit replayed before it was canonical (`Processed`), the group is rolled forward to the retained anchor snapshot at the commit's resulting epoch — the exact post-merge state the confirm path snapshots (`publish.rs` retains anchors at both the pre-commit and post-merge epochs) — and its `CommitStaged` observation is synthesized from the stamp. Any failure (missing anchor, epoch mismatch, non-canonical prefix) prunes the branch exactly as before, so this is strictly additive.
3. **Already-applied prefix skip at apply** (`openmls_projection.rs::apply_openmls_canonicalization_result`): the selected branch's leading run of `Processed` commits below the live tip is the live canonical chain; it is excluded from the apply replay (an own commit there cannot be re-processed) and the apply start epoch moves past it. When the own branch wins, apply is a no-op on group state; when it loses, the existing rollback-to-anchor + replay path is unchanged.

With both branches materialized, selection and withdrawal compose correctly with #702: a losing own branch is dropped `InvalidAgainstCandidateState` and withdrawn via `emit_rolled_back_commits` (same stamped origin id); `emit_superseded_processed_commits` stays as the backstop for own commits that never materialized (e.g. pre-stamp records).

### Tests

In `crates/cgka-engine/tests/update_group_data.rs`, modeled on the existing restart-loser variant (`rebuilt_engine_convergence_withdraws_own_confirmed_rename_by_stamped_origin`):

- `rebuilt_engine_convergence_keeps_own_confirmed_rename_when_it_wins_selection` — restart-WINNER variant; without the fix the sibling is wrongly selected.
- `mutually_rebuilt_engines_converge_on_same_branch_after_concurrent_renames` — mutual-restart variant; without the fix the group forks permanently.

Both fail without the fix and pass with it; the full `cgka-engine` suite, `cgka-traits`, `storage-sqlite`, and `cgka-conformance-simulator` suites pass.

### Residual edges (out of scope here, noting for the record)

- **Sent-but-unconfirmed own commits**: after restart, `pending_commit_recovery` clears the surviving MLS pending commit, so a published-but-never-confirmed own commit still cannot materialize (no stamp, no post-merge anchor). Its record stays `Sent` and the branch prunes as before.
- **Pre-fix records**: own commits confirmed before the stamp existed have no `OwnCommitWire` payload and keep the old (pruned) behavior; #702's `emit_superseded_processed_commits` continues to cover their withdrawals.
- **Accepted proposals below a skipped prefix**: if the selected branch is the own branch and an accepted proposal targets an epoch below the apply-start (possible only with an unresolved same-epoch sibling plus a pending fork-epoch proposal), its apply replay can fail with a wrong-epoch error; the apply snapshot rolls back cleanly. Epoch-bound proposals below the selected tip can never be consumed by future commits, so the right long-term treatment is a terminal disposition for them.

### Context

Discovered while fixing #363 (PR #702, branch `fix/363-state-change-invalidation`); the fix builds on that branch (the withdrawal paths compose).

erskingardner:
Fixes #706.

A device's own published-and-confirmed commit cannot be replayed through OpenMLS `process_message`, so after an engine restart cleared the in-memory `committed_from` guard, its branch silently dropped out of stored-convergence branch selection: a losing same-epoch sibling was selected even when the own commit's authenticated ordering key should win, and when both concurrent committers restarted, each deterministically reorged onto the other's branch — a permanent group fork.

### What this does

1. **Confirm-time stamp** (`crates/traits/src/message.rs`, `crates/cgka-engine/src/publish.rs`): new `StoredMessagePayload::OwnCommitWire { message, stamp }` variant carrying `OwnCommitConvergenceStamp { committer, priority, consumed_proposal_refs }`. `do_confirm_published` captures it while the staged commit is still attached and enriches the stored wire record in the same durable write that marks it `Processed`. Convergence replay still reads the record as a plain wire row; the stamp rides alongside.
2. **Anchor rollforward during candidate materialization** (`crates/cgka-engine/src/openmls_projection.rs`): when a replayed candidate path reaches a stamped own commit and every commit replayed before it was canonical (`Processed`), the group is rolled forward to the retained anchor snapshot at the commit's resulting epoch — the exact post-merge state the confirm path retains — and its `CommitStaged` observation is synthesized from the stamp. Any failure (missing anchor, epoch mismatch, non-canonical prefix) prunes the branch exactly as before, so the change is strictly additive.
3. **Already-applied prefix skip at apply**: the selected branch's leading run of `Processed` commits below the live tip is the live canonical chain; it is excluded from the apply replay and the apply-start epoch moves past it. An own-branch win applies without re-processing own commits (no-op on group state); an own-branch loss keeps the existing rollback-to-anchor + replay path.

With both branches materialized, selection and withdrawal compose with #702: a losing own branch is dropped `InvalidAgainstCandidateState` and withdrawn via `emit_rolled_back_commits` under the same confirm-time stamped origin id; `emit_superseded_processed_commits` remains the backstop for own commits that never materialized (pre-stamp records).

### Tests

- `rebuilt_engine_convergence_keeps_own_confirmed_rename_when_it_wins_selection` — restart-WINNER variant: without the fix the losing sibling is wrongly selected and the device reorgs off its own canonical branch.
- `mutually_rebuilt_engines_converge_on_same_branch_after_concurrent_renames` — mutual-restart variant: without the fix both devices swap onto each other's branch (verified: disabling the stamp collection reproduces the fork with exactly that failure).
- JSON shape snapshot for the new payload variant; `publish_lifecycle` fault-injection extended to cover the enriched `Processed` write.

`just fast-ci` green; `cgka-engine`, `cgka-traits`, `storage-sqlite`, and `cgka-conformance-simulator` suites pass on this branch rebased onto current master (includes #701/#702/#703/#704).

### Residual edges (documented in #706)

Sent-but-never-confirmed own commits still prune (no stamp, no post-merge anchor); pre-fix records keep the old behavior with #702's backstop; an accepted fork-epoch proposal below a skipped prefix can fail apply-replay cleanly (snapshot rolls back).

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/723">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Stored convergence now preserves metadata for confirmed own commits, improving how restarted devices reconcile their latest changes.
  * Confirmed commits can be replayed more reliably from local storage without being reprocessed incorrectly.

* **Bug Fixes**
  * Fixed branch selection issues after restart so a device is less likely to switch away from its own confirmed commit.
  * Improved handling of persisted message processing to avoid duplicate replay of already-applied commits.
  * Strengthened retry behavior for transient write failures during publish confirmation.

### DIFF
--- a/crates/cgka-engine/src/openmls_projection.rs
+++ b/crates/cgka-engine/src/openmls_projection.rs
@@ -213,6 +213,7 @@ struct StoredOpenMlsCanonicalizationWork {
     policy: CanonicalizationPolicy,
     now_ms: u64,
     replay_start_epoch: u64,
+    own_commits: PrevalidatedOwnCommits,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -309,6 +310,107 @@ impl From<StorageError> for OpenMlsProjectionError {
     }
 }
 
+/// Own published-and-confirmed commits usable as pre-validated candidate-path
+/// segments during stored-convergence materialization.
+///
+/// MLS cannot process a device's own commit through `process_message`, so a
+/// candidate path containing one cannot be materialized by replay alone: after
+/// a restart clears the in-memory `committed_from` guard, the own commit's
+/// branch would silently drop out of branch selection and the device could be
+/// reorged onto a losing sibling (or, with two restarted committers, the group
+/// could fork permanently). Instead, an own `Processed` commit whose ordering
+/// stamp was persisted at confirm time is realized by rolling the group
+/// forward to the retained anchor snapshot at its resulting epoch — the exact
+/// post-merge state the commit produced — and synthesizing its
+/// `CommitStaged` observation from the stamp.
+#[derive(Clone, Debug, Default)]
+pub(crate) struct PrevalidatedOwnCommits {
+    /// Confirm-stamped own `Processed` commits, keyed by wire-bytes digest.
+    by_digest: BTreeMap<[u8; 32], cgka_traits::message::OwnCommitConvergenceStamp>,
+    /// Digests of every `Processed` commit (own and others') — the canonical
+    /// chain. An own commit is pre-validated only while the replayed path
+    /// prefix stays canonical: it was created from the canonical state at its
+    /// source epoch, so on any diverging prefix it must NOT apply.
+    canonical_digests: BTreeSet<[u8; 32]>,
+}
+
+impl PrevalidatedOwnCommits {
+    fn insert_canonical(&mut self, digest: [u8; 32]) {
+        self.canonical_digests.insert(digest);
+    }
+
+    fn insert_stamped(
+        &mut self,
+        digest: [u8; 32],
+        stamp: cgka_traits::message::OwnCommitConvergenceStamp,
+    ) {
+        self.by_digest.insert(digest, stamp);
+    }
+
+    fn stamp(&self, digest: &[u8; 32]) -> Option<&cgka_traits::message::OwnCommitConvergenceStamp> {
+        self.by_digest.get(digest)
+    }
+
+    fn is_canonical(&self, digest: &[u8; 32]) -> bool {
+        self.canonical_digests.contains(digest)
+    }
+}
+
+/// Build the convergence ordering stamp for a staged local commit, captured
+/// while the staged commit is still attached (confirm time). See
+/// [`cgka_traits::message::OwnCommitConvergenceStamp`].
+pub(crate) fn own_commit_stamp(
+    staged: &openmls::group::StagedCommit,
+    committer: MemberId,
+) -> Result<cgka_traits::message::OwnCommitConvergenceStamp, cgka_traits::error::EngineError> {
+    let priority = crate::app_components::commit_ordering_priority_for_staged(staged);
+    let mut consumed_proposal_refs = staged
+        .queued_proposals()
+        .map(|proposal| tls_hex(proposal.proposal_reference_ref()))
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|e| cgka_traits::error::EngineError::Serialize(format!("{e}")))?;
+    consumed_proposal_refs.sort();
+    consumed_proposal_refs.dedup();
+    Ok(cgka_traits::message::OwnCommitConvergenceStamp {
+        committer,
+        priority,
+        consumed_proposal_refs,
+    })
+}
+
+/// Mark the origin commit record `Processed` and, when a confirm-time stamp is
+/// available, enrich its stored wire payload to `OwnCommitWire` in the same
+/// write, so stored convergence can later rebuild this commit's branch as a
+/// pre-validated candidate.
+pub(crate) fn stamp_processed_own_commit_record<S: StorageProvider>(
+    storage: &S,
+    message_id: &MessageId,
+    stamp: Option<cgka_traits::message::OwnCommitConvergenceStamp>,
+) -> Result<(), cgka_traits::error::EngineError> {
+    let Some(stamp) = stamp else {
+        storage.update_message_state(message_id, MessageState::Processed)?;
+        return Ok(());
+    };
+    let mut record = storage.get_message(message_id)?;
+    let payload = StoredMessagePayload::decode(&record.payload)
+        .map_err(|e| cgka_traits::error::EngineError::Serialize(format!("{e:?}")))?;
+    let payload = match payload {
+        StoredMessagePayload::OpenMlsWire(message)
+        | StoredMessagePayload::OwnCommitWire { message, .. } => {
+            StoredMessagePayload::own_commit_wire(message, stamp)
+        }
+        // Raw-transport rows never enter the OpenMLS candidate graph; a plain
+        // state update preserves their shape.
+        other @ StoredMessagePayload::RawTransport(_) => other,
+    };
+    record.state = MessageState::Processed;
+    record.payload = payload
+        .encode()
+        .map_err(|e| cgka_traits::error::EngineError::Serialize(format!("{e:?}")))?;
+    storage.put_message(&record)?;
+    Ok(())
+}
+
 pub fn project_mls_message(
     bytes: &[u8],
 ) -> Result<OpenMlsMessageProjection, OpenMlsProjectionError> {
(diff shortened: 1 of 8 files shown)


crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1138,13 +1356,19 @@ fn apply_openmls_canonicalization_result_inner<S: StorageProvider>(
 fn replay_messages_for_canonicalization_result<S: StorageProvider>(
     storage: &S,
     result: &CanonicalizationResult,
+    applied_prefix: &BTreeSet<String>,
 ) -> Result<Vec<TransportMessage>, OpenMlsProjectionError> {
     let mut replay_messages = Vec::new();
     let mut seen = BTreeSet::new();
     for hex_message_id in result
         .accepted_proposals
```
erskingardner, line 1369: This still replays proposals that were consumed by commits in the skipped `applied_prefix`. In the branch this PR is fixing, if the already-applied own commit consumed a fork-epoch proposal, `applied_prefix` filters the commit out and `apply_start_epoch` stays at the live tip, but this loop still puts that old proposal into `replay_messages` before the unskipped suffix. OpenMLS will then process a proposal whose source epoch is below the live group epoch and reject the apply, rolling back the convergence pass.

I think this needs to filter `accepted_proposals` to only proposals consumed by the unskipped commit suffix, or else start replay from the earliest accepted proposal/commit source epoch instead of the first unskipped commit. A regression test where the winning restarted own commit consumes a pending proposal would catch it.

erskingardner: Good catch — confirmed, and it was worse than a one-off rejected pass: the proposal stayed a pending input, so every retry re-selected the own branch and hit the same WrongEpoch apply failure — a permanent livelock for the group.

Fixed in 795dace by filtering the replay set: an accepted proposal whose record epoch sits below the apply-start epoch can only have been consumed by a commit in the skipped prefix (MLS proposals are epoch-scoped, so no unskipped suffix commit at a later epoch could consume it). Its effect is already inside the live state, and its `Processed` disposition still persists — which is what removes it from the convergence input set and closes the retry loop. When the prefix is empty (normal rollback path) the apply-start epoch is the fork epoch and no accepted proposal sits below it, so that path is unchanged.

Regression test as suggested: `rebuilt_winner_applies_own_selfremove_commit_without_replaying_consumed_proposal` — both remaining members race the SelfRemove auto-commit (each consuming the leaver's proposal by reference), the ordering-key winner restarts with the consumed proposal re-opened as a pending input, and converges over the losing sibling. Without the filter it fails with exactly the apply failure you described; with it, the pass settles, the own branch survives, and the proposal record returns to `Processed`.

One note from building the test: a welcome-joined member whose only own commit is the auto-commit holds no retained anchor at the fork epoch, because the auto-commit send path projects the group record to the post-merge epoch before confirm's pre-commit retention reads it — so historical replay reports `MissingRetainedAnchor` and the group goes `Unrecoverable`. That's a pre-existing retention quirk orthogonal to this PR (the test sidesteps it with a settled rename round first); happy to file it separately.

PR #724, opened 2026-07-04
Clear removed marker when branch selection supersedes the local member's removal

erskingardner:
## What

Handles the #702 × #703 interaction left out of scope during the #702 merge: a commit that removes the local member's own leaf is applied (realizing removal — `Group.removed` set, send gate closed, group presented as removed), and that commit later **loses branch selection**. The withdrawal machinery from #702 already retracts the self-removed notification; this PR pins the marker semantics so the device does not stay self-quarantined for a group it canonically still belongs to.

- `distributed_convergence::emit_convergence_events`: when the selected canonical branch records the local member's membership, a surviving `removed` marker is cleared — the explicit, state-derived inverse of `realize_self_eviction`. Outbound intents purged at realization stay purged; the app re-issues sends against the reopened gate.
- Doc updates in `engine.rs` / `traits::Group`: the marker is terminal only while the removal remains canonical; it clears via authenticated re-join **or** supersession of the removal that set it.
- CLI CHANGELOG (Unreleased/Fixed): sends work again on such a group instead of failing `invalid_transition` forever.

## Spec basis

member-departure.md defines removal evidence as "an accepted commit **on the selected canonical branch** … recorded in retained canonical state" and makes realization state-derived; convergence.md ("Applying the selected branch") requires that a change that lost branch selection "MUST NOT remain visible to the application as a completed change". A superseded removal was never canonically applied, so clearing is spec-required, not a violation of the "terminal for that group copy" rule — that rule presumes the removal stays canonical. Since the terminal sentence can be misread as unconditional, a spec clarification is filed as marmot-protocol/marmot#220.

## What a reviewer should know

**The marker could not actually survive a convergence reorg on master.** The reorg apply rolls back to the retained anchor before replaying the winner, and SQLite group snapshots capture the whole group record — so the anchor restore already reset `removed: false` incidentally (see `snapshot_rollback_restores_group_messages_queue_caps_and_openmls_group_state`). The resulting view was spec-correct, but only as an undocumented side effect of snapshot mechanics; a future change narrowing snapshot scope (plausible, if someone deliberately preserves the "terminal" marker across rollbacks) would silently reintroduce the quarantine bug. This PR makes the invariant explicit in code, docs, and tests.

## Tests

- `superseded_self_removal_clears_removed_marker_and_restores_send` — the previously untested scenario end-to-end: Carol applies a peer commit removing her own leaf (marker set, send rejected with the `Removed` invalid-transition, self-removed notification emitted), then the removal loses a same-epoch fork to a rename through stored convergence. Asserts the full resulting view: withdrawal names the superseded removal in the stamped id space, `CommitRolledBack` fires, no `ForkRecovered`, marker cleared, membership + the winner's rename presented, roster-correction `MemberAdded` emitted, and the previously rejected send succeeds. Passes with or without the code change — it pins the behavior against regressions in either mechanism (anchor restore or explicit clear).
- `convergence_apply_clears_removed_marker_without_canonical_evidence` — covers the explicit reconciliation branch directly (marker forced on the record, mirroring the existing pathological-copy tests in `invite_leave.rs`); fails without the fix.

Verified: both new tests, full `cargo test -p cgka-engine`, `cargo test -p cgka-traits`, `just fast-ci`, `cargo fmt --check`.

Related: the open post-restart own-commit branch-selection investigation is how this scenario becomes reachable in practice.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/724">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed groups now clear their “removed” marker when later convergence indicates the device is still an active member.
  * App sending eligibility is restored after convergence reorg/branch selection supersedes a prior self-removal state.
  * Superseded self-removal no longer leaves the group copy incorrectly self-quarantined.
* **Tests**
  * Added regression tests covering removed-marker reconciliation across stored-convergence apply/reorg edge cases.
* **Documentation**
  * Updated the changelog and expanded “removed” marker documentation to clarify all clearing scenarios.

### DIFF
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -632,6 +632,40 @@ impl<S: StorageProvider> Engine<S> {
             );
         }
         if current_ids.contains(self.identity.self_id()) {
+            // The selected canonical branch records our membership, so a
+            // `removed` marker on the record is inconsistent with canonical
+            // state and must clear: a previously applied removal that lost
+            // branch selection "MUST NOT remain visible to the application as
+            // a completed change" (convergence.md, "Applying the selected
+            // branch"), and member-departure.md's "terminal for that group
+            // copy" rule presumes removal evidence on the selected canonical
+            // branch (clarification tracked in marmot-protocol/marmot#220).
+            // A reorg normally already restored the pre-removal
+            // record from the retained anchor before the replay (the anchor
+            // predates the marker write), so this is the explicit
+            // state-derived inverse of `realize_self_eviction`: it reconciles
+            // the pathological copy whose marker survived without canonical
+            // evidence, and keeps the send-gate posture derivable from
+            // canonical state alone. Outbound intents purged at realization
+            // stay purged — the app re-issues sends against the reopened
+            // send gate.
+            let mut group = self
+                .storage
+                .get_group(group_id)
+                .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?;
+            if group.removed {
+                group.removed = false;
+                self.storage
+                    .put_group(&group)
+                    .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?;
+                // Privacy-safe breadcrumb: aggregate signal only, no
+                // group/member ids (observability.md).
+                tracing::info!(
+                    target: "cgka_engine::distributed_convergence",
+                    method = "emit_convergence_events",
+                    "cleared removed marker: selected canonical branch records local membership"
+                );
+            }
             if self
                 .load_leave_request_state(group_id)
                 .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -568,6 +568,10 @@ impl<S: StorageProvider> Engine<S> {
                 );
             }
         }
+        // Captured before `current_group.members` is consumed below: gates the
+        // removed-marker reconciliation so the common (not-removed) path pays
+        // no extra storage read.
+        let group_record_was_removed = current_group.removed;
         let previous_ids: HashSet<MemberId> = previous_members
             .iter()
             .map(|member| member.id.clone())
@@ -648,23 +652,28 @@ impl<S: StorageProvider> Engine<S> {
             // evidence, and keeps the send-gate posture derivable from
             // canonical state alone. Outbound intents purged at realization
             // stay purged — the app re-issues sends against the reopened
-            // send gate.
-            let mut group = self
-                .storage
-                .get_group(group_id)
-                .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?;
-            if group.removed {
-                group.removed = false;
-                self.storage
-                    .put_group(&group)
+            // send gate. Gated on the record snapshot loaded above so the
+            // common (not-removed) path re-reads nothing; the re-fetch on the
+            // rare marked path re-checks under the fresh record before
+            // writing.
+            if group_record_was_removed {
+                let mut group = self
+                    .storage
+                    .get_group(group_id)
                     .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?;
-                // Privacy-safe breadcrumb: aggregate signal only, no
-                // group/member ids (observability.md).
-                tracing::info!(
-                    target: "cgka_engine::distributed_convergence",
-                    method = "emit_convergence_events",
-                    "cleared removed marker: selected canonical branch records local membership"
-                );
+                if group.removed {
+                    group.removed = false;
+                    self.storage
+                        .put_group(&group)
+                        .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?;
+                    // Privacy-safe breadcrumb: aggregate signal only, no
+                    // group/member ids (observability.md).
+                    tracing::info!(
+                        target: "cgka_engine::distributed_convergence",
+                        method = "emit_convergence_events",
+                        "cleared removed marker: selected canonical branch records local membership"
+                    );
+                }
             }
             if self
                 .load_leave_request_state(group_id)
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -1,4 +1,25 @@
 //! Engine entry point for stored-message distributed convergence.
+//!
+//! # Removed-marker lifecycle
+//!
+//! `Group.removed` realizes the local member's own removal
+//! (member-departure.md, "Realizing removal"): it is set when an accepted
+//! commit on the selected canonical branch removes our leaf — on the direct
+//! seam, via `realize_self_eviction`, or in `emit_convergence_events` here —
+//! and it clears in exactly two ways: an authenticated re-join
+//! (`do_join_welcome`), or branch selection superseding the removal that set
+//! it. A superseded removal was never canonically applied, so it "MUST NOT
+//! remain visible to the application as a completed change" (convergence.md,
+//! "Applying the selected branch"); member-departure.md's "terminal for that
+//! group copy" rule presumes the removal stays canonical (clarification
+//! tracked in marmot-protocol/marmot#220). A supersession reorg normally
+//! restores the pre-removal record wholesale from the retained anchor (the
+//! anchor predates the marker write); the explicit reconciliation in
+//! `emit_convergence_events` is the state-derived inverse of
+//! `realize_self_eviction` for markers no anchor restore can see, and it
+//! clears only when the canonical roster AND the live MLS state both record
+//! our active membership. Outbound intents purged at realization stay purged
+//! — the app re-issues sends against the reopened send gate.
 
 use std::collections::HashSet;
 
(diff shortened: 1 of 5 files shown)


crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -632,6 +636,45 @@ impl<S: StorageProvider> Engine<S> {
             );
         }
         if current_ids.contains(self.identity.self_id()) {
```
erskingardner, line 691: **Predicate trusts Marmot `Group.members`, not MLS.**

The guard keys off `current_ids.contains(self)` from the record loaded at the top of this function. That is usually correct immediately post-replay, but the send gate is safety-critical and this is the inverse of `realize_self_eviction`, which strips `members` from the record.

If a pathological copy ever has `removed = true` with a **stripped** roster while the replayed `MlsGroup` still carries our leaf (or the inverse — stale `members` listing self while MLS does not), this branch either skips the heal or clears the marker incorrectly.

Consider belt-and-suspenders: also require `mls_group.own_leaf_node().is_some()` (you already have `reorg_component_snapshot` loading MLS state in this function) before clearing. That keeps the gate derivable from canonical MLS state, matching the comment's stated goal.

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -632,6 +636,45 @@ impl<S: StorageProvider> Engine<S> {
...
+            // send gate. Gated on the record snapshot loaded above so the
+            // common (not-removed) path re-reads nothing; the re-fetch on the
+            // rare marked path re-checks under the fresh record before
+            // writing.
+            if group_record_was_removed {
+                let mut group = self
+                    .storage
+                    .get_group(group_id)
+                    .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?;
+                if group.removed {
```
erskingardner, line 704: **Not a full inverse of `realize_self_eviction`.**

Setting the marker via the direct seam / `realize_self_eviction` also strips self from `Group.members` and purges queued outbound intents. This path only flips `removed` back to `false` (intentionally leaving purged intents purged — noted in the comment).

That is fine when roster replay already restored membership, but the pathological test only forces `removed = true` while leaving `members` intact. Worth a sibling test that sets `removed = true` **and** strips self from `members` (mirroring post-`realize_self_eviction` shape) then forward-applies a commit — to prove the guard heals the send gate even when the record/roster was fully evicted-shaped before replay refreshed it.

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -632,6 +636,45 @@ impl<S: StorageProvider> Engine<S> {
             );
         }
         if current_ids.contains(self.identity.self_id()) {
+            // The selected canonical branch records our membership, so a
```
erskingardner, line 692: Nit: ~20 lines of spec rationale inline here will drift. Most of this is already captured well in the PR description and `traits::Group` doc update — consider trimming to 2–3 lines plus a pointer to marmot-protocol/marmot#220, and moving the longer rationale to a module-level rustdoc paragraph on the removed-marker lifecycle.

crates/cgka-engine/tests/distributed_convergence.rs
```diff
@@ -1612,6 +1612,331 @@ async fn convergence_rollback_emits_commit_rolled_back_for_losing_branch() {
...
+/// so it "MUST NOT remain visible to the application as a completed change"
+/// (convergence.md, "Applying the selected branch") — member-departure.md's
+/// terminal-marker rule presumes removal evidence "on the selected canonical
+/// branch". This test pins the whole resulting view of that supersession:
+/// the withdrawal names the superseded removal in the id space the
+/// self-removed notification was stamped with, the copy stops
+/// self-quarantining (marker cleared, membership and the winner's rename
+/// presented), and the send gate reopens.
+#[tokio::test]
+async fn superseded_self_removal_clears_removed_marker_and_restores_send() {
```
erskingardner, line 1628: **This test does not exercise the new guard** (as the PR description acknowledges): anchor restore already resets `removed` before `emit_convergence_events` runs, so it passes with or without this diff.

That makes it a valuable regression pin for snapshot semantics, but a weak pin for *this* code path. Options:
1. Assert the new `tracing::info!(… "cleared removed marker" …)` breadcrumb fires on reorg (privacy-safe, no ids).
2. Add a variant that forces `removed = true` **after** the reorg anchor restore but **before** event emission (if you can hook that), so the explicit guard is the only heal.

Without one of these, CI green on this test alone would not catch a revert of the new block.

crates/cgka-engine/tests/distributed_convergence.rs
```diff
@@ -1612,6 +1612,331 @@ async fn convergence_rollback_emits_commit_rolled_back_for_losing_branch() {
...
+/// State-derived inverse of `realize_self_eviction`: a `removed` marker that
+/// survives WITHOUT canonical evidence — the selected canonical branch
+/// records our membership — is reconciled by the next convergence apply. The
+/// marker is forced directly on the record to simulate the pathological
+/// copy: a real supersession reorg already restores the pre-removal record
+/// from the retained anchor (covered by
+/// `superseded_self_removal_clears_removed_marker_and_restores_send`); this
+/// pins the explicit guard for a marker no anchor restore can see.
+#[tokio::test]
+async fn convergence_apply_clears_removed_marker_without_canonical_evidence() {
```
erskingardner, line 1855: Good — this is the test that actually fails without the fix.

One gap: it simulates a marker with membership still present, but not the post-`realize_self_eviction` shape (`removed = true`, self stripped from `members`). A combined pathological setup would increase confidence that clearing the marker and reopening the send gate still works when the record looked fully evicted before replay.

erskingardner (commented):
## Adversarial review summary

Solid fix for a real spec gap: making the removed-marker clear explicit rather than relying on anchor-restore side effects is the right defensive move, and `convergence_apply_clears_removed_marker_without_canonical_evidence` is the test that actually gates the new code.

I don't see a correctness bug in the convergence guard itself — the mutual exclusion with the removal loop in the same pass looks sound, and the `group_record_was_removed` snapshot + re-fetch pattern is a reasonable optimization.

Main gaps are **parity** (direct seam still snapshot-only), **coverage** (the big E2E test passes without this diff), and **restart** (hydration never reconciles a stale marker). Inline notes below.

Overall: **approve direction**, with follow-ups worth tracking before/soon after merge.

erskingardner:
Two additional adversarial notes that don't map to changed lines in this diff:

### 1. Direct-seam parity gap (`message_processor/ingest.rs`)

The post-`ForkRecovered` block (~889–902) handles leave-request restoration when `after_ids.contains(self)` but **never explicitly clears `Group.removed`**. Same story as pre-fix convergence: snapshot rollback to the incumbent's pre-commit anchor normally restores `removed = false`, but that is incidental — the direct seam sets `g.removed = true` on self-removal (~810–815) without a symmetric clear on supersession.

If snapshot scope ever narrows (the scenario this PR explicitly guards against on the convergence side), the direct seam would reintroduce the quarantine bug for clients that applied a self-removal via ingest rather than stored convergence. Worth either mirroring this guard there or documenting the shared snapshot dependency in one place.

### 2. Session restart / hydration

`hydrate_one_stored_group` never reconciles `Group.removed` against MLS membership at open. This fix only runs on the next `converge_stored_openmls_messages` pass. The PR description already flags the post-restart branch-selection path as the practical trigger — until that lands, a device that persists a pathological `removed = true` and restarts stays send-gated until convergence runs again. Not necessarily in scope for #724, but worth a tracked follow-up (hydrate-time reconcile or a startup convergence sweep).

crates/cgka-engine/src/distributed_convergence.rs
erskingardner: Added `self_leaf_is_active_in_mls` — the clear now also requires the live MlsGroup to load, be active, and carry our identity on its own leaf (fail-closed on any missing/unreadable state), so the heal is derivable from canonical MLS state, not the record roster alone.

crates/cgka-engine/src/distributed_convergence.rs
erskingardner: Added `convergence_apply_heals_fully_evicted_shaped_record`: removed = true AND self stripped from members (the post-realize_self_eviction shape), forward apply rebuilds the roster from replayed MLS state, the guard clears the marker, and the send gate reopens. Fails without the guard (verified against pre-PR source).

crates/cgka-engine/src/distributed_convergence.rs
erskingardner: Trimmed to ~6 lines pointing at a new module-level "Removed-marker lifecycle" rustdoc (which carries the full rationale + marmot-protocol/marmot#220 pointer).

crates/cgka-engine/tests/distributed_convergence.rs
erskingardner: The guard intentionally never fires on a reorg — the anchor restore precedes emit_convergence_events, so asserting the breadcrumb there would encode wrong behavior, and there's no seam to inject a marker between restore and emission. The guard-only path is pinned by the two healing tests (both fail without the block, verified against pre-PR source); this e2e test deliberately pins the supersession resulting view against regressions in either mechanism (snapshot scope or the guard).

crates/cgka-engine/tests/distributed_convergence.rs
erskingardner: Added the combined-shape sibling: `convergence_apply_heals_fully_evicted_shaped_record` (removed = true + self stripped from members), asserting marker cleared, roster rebuilt from replayed MLS state, roster-correction MemberAdded, and a successful send after the heal.

PR #726, opened 2026-07-04
Close out #707: mirror ingest invariants on every seam, enforce quarantine, bound the deferred-peel lifecycle

### LINKED ISSUE #339: Engines attempt to decrypt pre-membership messages, causing large processing backlogs
## Summary

When an engine joins an existing group, it may receive historical encrypted events created before that member joined.

The engine currently appears to attempt to peel or decrypt many of these events even though the member was not part of the group at the corresponding epochs and therefore cannot decrypt them by design.

This produces large numbers of `peel_failed_no_snapshot` results and may create a significant local processing backlog.

## Problem

Pre-membership application messages are permanently undecryptable for a newly joined member.

They should not be handled in the same way as delayed messages from epochs in which the engine was a valid group member.

The engine currently does not appear to distinguish clearly between:

### Pre-membership events

- The engine was not a member at the event epoch.
- Decryption is impossible by design.
- The event should not be retried.

### Recoverable historical events

- The engine was a member at the event epoch.
- Historical state or a retained snapshot may be required.
- Recovery or retry may be appropriate.

## Observed behavior

After joining or catching up, an engine can attempt to process a large historical event backlog.

Many events produce:

```text
peel_failed_no_snapshot
```

The current telemetry does not show whether each failed event:

- predates the engine’s membership,
- belongs to an epoch where the engine was a member,
- is a duplicate,
- is permanently undecryptable,
- or is still eligible for recovery.

## Suspected impact

A large number of unnecessary peel attempts may:

- consume CPU and battery,
- increase memory pressure,
- create database contention,
- increase queue depth,
- delay processing of current events,
- delay message display,
- slow reconnect and backfill completion,
- produce repeated retries that can never succeed,
- obscure real recovery failures in telemetry,
- amplify an existing delivery backlog.

This may create a feedback loop:

```text
Delayed delivery
→ large historical backfill
→ unnecessary peel attempts
→ local processing slowdown
→ slower catch-up of current events
```

The logs do not yet prove that this backlog causes transport delays, but it is a plausible local performance amplifier.

## Expected behavior

Before attempting an expensive peel operation, the engine should determine whether the event falls within the engine’s valid membership history.

### Pre-membership event

```text
event_epoch < member_join_epoch
→ classify as pre-membership
→ do not peel
→ do not retry
→ optionally retain minimal metadata for deduplication
```

### Valid historical event

```text
engine was a member at event_epoch
→ locate current or retained epoch state
→ attempt peel
→ retry only when recovery is possible
```

### Duplicate or already processed event

```text
event ID already classified or processed
→ deduplicate before peel
→ do not repeat expensive work
```

## Proposed changes

- Record the engine’s join epoch or valid membership interval.
- Determine membership eligibility before peeling an event.
- Skip application messages from before the engine joined.
- Deduplicate events before any cryptographic processing.
- Mark permanently undecryptable events as terminal.
- Prevent retries for pre-membership events.
- Separate pre-membership skips from missing-snapshot failures.
- Prioritize current-epoch events over irrelevant historical backlog.
- Add queue-depth and processing-duration telemetry.

## Telemetry requirements

For every skipped or failed peel, expose:

- event ID
- event epoch
- engine join epoch
- whether the engine was a member at that epoch
- classification
- snapshot expected
- snapshot found or missing
- retryable or terminal
- retry count
- queue wait time
- peel processing duration

Suggested classifications:

- `pre_membership_event`
- `valid_history_snapshot_missing`
- `duplicate_event`
- `already_processed`
- `retry_pending`
- `permanently_undecryptable`

## Acceptance criteria

- Pre-membership application messages are identified before peel.
- Pre-membership messages produce no cryptographic peel attempt.
- Pre-membership messages are never retried.
- Duplicate events are deduplicated before peel.
- Valid historical events still use retained-state recovery.
- Current events are not blocked behind irrelevant historical events.
- Telemetry clearly distinguishes skipped pre-membership events from genuine recovery failures.
- A large pre-membership backfill does not cause an unbounded processing queue or significant UI slowdown.

### LINKED ISSUE #364: Hydration quarantine is not enforced on the ingest / convergence seams — an unhealthy group keeps processing commits and silently re-activates
## Summary
`quarantined_groups` is consulted only at session-open hydration and in `retry_hydrate_quarantined_group`. None of the live data-path seams check it, so a group quarantined as unhealthy (e.g. `MemberValidationFailed`) keeps ingesting inbound commits and can silently un-quarantine itself.

## Where
- `do_ingest` never inspects `quarantined_groups` — `crates/cgka-engine/src/message_processor/mod.rs:37-84`.
- `ingest_group_message` independently does `MlsGroup::load(...)`; for a `MemberValidationFailed` quarantine the OpenMLS state is intact, so the load succeeds — `crates/cgka-engine/src/message_processor/ingest.rs:84-151`.
- The epoch-state gate does not stop it: a quarantined group has no `epoch_manager` entry, and `can_ingest` treats absence as ingestible:
```rust
pub(crate) fn can_ingest(&self, group_id: &GroupId) -> bool {
    self.states.get(group_id).is_none_or(|s| s.can_ingest())  // None => true
}
```
- Ingest then merges the commit, writes the group record, and calls `self.epoch_manager.set_stable(...)` (`ingest.rs:747`) — re-establishing as "live" a group the engine had excluded, while `quarantined_groups` still lists it. `converge_stored_openmls_messages` similarly falls back to `previous_group.epoch` when `epoch_manager` has no entry and proceeds — `crates/cgka-engine/src/distributed_convergence.rs:184+`.

## Impact
A group flagged unhealthy (e.g. a leaf that failed account-identity-proof/credential validation) keeps processing commits against the very state validation rejected, and is partially un-quarantined out-of-band (group record + `epoch_manager` updated, but the quarantine map and any app-side "vanished" UI remain stale). The intended fail-safe ("freeze until verified repair") is not enforced on the data path.

## Suggested fix
Gate `do_ingest` / `converge_stored_openmls_messages` on `quarantined_groups` (defer/skip and surface a structured "quarantined" disposition), and require an explicit repair to clear quarantine before `set_stable` can re-activate.

## Dedup
- **#426** (closed): app-side *recovery UX* for quarantined groups — does not enforce quarantine on the ingest/converge path. No filed issue covers this gap.

---

---

### LINKED ISSUE #365: Quarantined group `members()` returns the full roster while `epoch()` returns `UnknownGroup` — contract says both vanish
## Summary
The documented quarantine contract (`crates/cgka-engine/src/engine.rs:1011-1016`) says a hydration-quarantined group "is not in the live roster (`epoch`/`members` return `UnknownGroup`) and otherwise vanishes from the account." The two accessors disagree.

## Where
- `epoch()` reads `epoch_manager`, which never got `set_stable` for a quarantined group, so it correctly returns `UnknownGroup` — `crates/cgka-engine/src/engine.rs:1641-1649`.
- `members()` → `do_members()` reads the **durable Marmot record**, which is never deleted on quarantine, so it returns a fully-populated roster — `crates/cgka-engine/src/group_lifecycle.rs:623-631`:
```rust
pub(crate) fn do_members(&self, group_id: &GroupId) -> Result<Vec<Member>, EngineError> {
    let group = self.storage.get_group(group_id)?;   // still present for a quarantined group
    Ok(group.members)
}
```
A `MemberValidationFailed` quarantine (`engine.rs:683-687`) is raised *after* the group record loads fine, so the roster is fully populated.

## Impact
API/contract violation and inconsistent state: a caller trusting the documented invariant ("`members` returns `UnknownGroup`" ⇒ group is live) renders/acts on the roster of a group the engine declared "vanished," while `epoch()` simultaneously reports it unknown. The roster of a vanished group also leaks to the application against the stated intent.

## Suggested fix
Have `do_members` (and any other durable-record accessor) check `quarantined_groups` and return `UnknownGroup`, matching `epoch()`.

## Dedup
- **#426** (closed): recovery UX, not the accessor asymmetry. No filed issue covers this.

---

---

### LINKED ISSUE #369: cgka-engine: `persist_stored_message_payload` skips the write on a (group, epoch, state) match without comparing the payload variant
## Summary
The idempotency short-circuit compares only `group_id`, `epoch`, and `state` — not the encoded payload — so a second persist of the same message id at the same epoch/state with a *different* `StoredMessagePayload` variant is silently dropped, leaving the wrong variant stored.

## Where — `crates/cgka-engine/src/message_processor/store.rs:185-225`
```rust
if previous.as_ref().is_some_and(|record| {
    record.group_id == *group_id && record.epoch == epoch && record.state == state
}) {
    return Ok(());   // keeps the OLD payload even if a different one was supplied
}
```

## Impact
If the same id is persisted twice at the same epoch/state but as a different variant (e.g. a `raw_transport` record then an `openmls_wire` record both landing at `MessageState::Failed`), the second write is dropped. A later reader using `as_raw_transport` (`replay_buffered_messages` / `retry_deferred_peels`) or `as_openmls_wire` (convergence) sees the wrong shape and skips the record — quietly losing the message from one processing path. Narrow (most colliding persists differ in `state`), but the guard is weaker than intended.

## Suggested fix
Include a payload-variant/byte comparison in the short-circuit, or overwrite when the variant differs.

## Dedup
Unrelated to marmot-protocol/mdk#528 (torn-write transaction) or marmot-protocol/marmot#126/#356 (dedup-id derivation). No related issue found.

---

---

### LINKED ISSUE #383: cgka-engine: convergence/replay path can emit `MessageReceived` with an empty, unauthenticated sender (S3 regression)
### Summary
The convergence/replay seam does not mirror the audit fix **S3**, which hardened the *direct* ingest path so it never surfaces an application message whose leaf credential cannot be resolved. The reorg path can still emit `GroupEvent::MessageReceived` with a blank, unauthenticated `MemberId`.

### Location
- `crates/cgka-engine/src/openmls_projection.rs:1490-1508` (the `ApplicationMessage` arm of `process_openmls_messages_inner`)
- Emitter: `crates/cgka-engine/src/distributed_convergence.rs:606-628`

### Details
When `sender_member_id(processed.sender(), &mls_group)` returns `None`, `sender` becomes `vec[]` (lines 1396–1400). The commit arm immediately above (line **1442**) has an explicit guard:
```rust
if sender_id.is_none() {
    return Err(OpenMlsProjectionError::Replay("commit has no authenticated member sender".into()));
}
```
The application-message arm has **no** such guard. The only gate is `app_payload_is_valid_for_sender(&payload, &sender)`.

Verified: `app_payload_is_valid_for_sender(payload, &[])` calls `event.validate_sender(&hex::encode(&[]))` = `validate_sender("")`, and `validate_sender` (`crates/traits/src/app_event.rs:185`) returns `Ok` iff `event.pubkey == ""`. So a crafted `MarmotAppEvent` whose `pubkey` field is the empty string, delivered through the convergence/reorg seam with an MLS sender that does not resolve to a member leaf, is surfaced to the application with an empty, unauthenticated author.

The direct ingest path (`message_processor/ingest.rs:575-594`) was *explicitly* changed under audit item **S3** to mark such messages `Failed` rather than emit an empty-sender `MessageReceived`. This is the same defect reachable through a different seam.

### Impact
Silent attribution loss / spoofable author on out-of-order (buffered/replayed) application messages. Reachability is narrow (requires an app message whose sender does not resolve *and* `event.pubkey == ""`), but it is a regression of a deliberate security fix at a trust boundary.

### Suggested fix
In the `ApplicationMessage` arm, when `sender_id.is_none()` push `Ignored` instead of `ApplicationProcessed` (mirror `ingest.rs`), or refuse to emit `MessageReceived` for an empty sender in `emit_application_replay_events`. Consider collapsing the two parallel validators in `app_payload.rs` (`validate_app_payload_for_sender` typed vs `app_payload_is_valid_for_sender` bool) into one so both seams share one code path.

---

---

### LINKED ISSUE #394: cgka-engine: `unreachable!` in the fork-recovery missing-snapshot branch is attacker-reachable if the invariant ever breaks
### Summary
A defensive `unreachable!()` sits in an attacker-reachable ingest path; it is correct today but a future change to either of two separate lookups turns it into a remote-input-triggered panic (DoS).

### Location
`crates/cgka-engine/src/message_processor/ingest.rs:486-489`

### Details
When `recovery_snapshot_name_for_fork` returns `None`, the code calls `resolve_fork_candidate` and `unreachable!()`s on the `IncumbentWins` / `CandidateWins` arms, relying on the implicit invariant that both methods read the same `incumbents` map. The coupling is implicit across two `self.fork_recovery` calls.

### Suggested fix
Replace `unreachable!` with a fail-closed `Err(EngineError::Backend(...))` so a future invariant break degrades to a handled error rather than a panic on an attacker-delivered same-epoch fork commit.

---

---

### LINKED ISSUE #707: [Tracking] Convergence input lifecycle, seam parity, and restart recovery
## Scope

Reopened after the 2026-07-21 full-backlog audit. The direct-ingest, stored-convergence, replay, fork-recovery, and Welcome seams must classify the same input the same way, persist terminal/retryable state consistently, and rebuild every gating state at hydrate.

## Program checklist

- [x] #962 — Security review C1: never-validating commits get no terminal disposition — one crafted commit permanently freezes all outbound group traffic
- [x] #963 — Security review C2: stale proposals poison every candidate-path probe — permanent silent convergence freeze
- [x] #967 — Security review H4: crafted gift-wrapped welcome hard-errors the engine and livelocks the account worker's transport loop
- [x] #971 — Security review M2: Unrecoverable state is in-memory only — restart silently clears it without verified repair
- [x] #976 — Security review M7: post-peel MLS deserialization failure escapes as Err — poison-pill redelivery
- [x] #983 — Security review M14: restart does not seed convergence scheduling from durable inputs — queued sends wait for fresh traffic
- [x] #906 — cgka-engine: flood-cap-dropped undecryptable messages are marked `seen`, so identical-id transport redelivery (the documented recovery path) never recovers them
- [x] #806 — marmot-app: scheduled-convergence settlement timer has no maximum deferral deadline — sustained buffered-convergence input can starve a group's convergence pass indefinitely
- [x] #885 — cgka-engine: a failed/rolled-back commit-producing send leaves a stale `committed_from` guard, degrading a later legitimate same-epoch fork into a fail-closed quarantine
- [x] #932 — cgka-engine: reserve_peel_deferred_slot increments the in-memory flood-cap counter before the durable persist — a persist failure leaks a cap slot for the rest of the session
- [x] #1021 — Security review L27: malformed (not just duplicate) input not extending the settlement quiescence window is untested

## Contract

- Classify before an input can affect quiescence or the outbound gate.
- Give every retained input an explicit terminal, retryable, or expiring lifecycle.
- Persist and rehydrate liveness-affecting state, including Unrecoverable and durable convergence work.
- A malformed or permanently invalid input must not livelock a worker or poison every future pass.

## Acceptance criteria

- C1/C2 regression tests fail before the fix and pass after it.
- Invalid, duplicate, stale, and terminal input is excluded from quiescence and send gating.
- Restart tests prove durable work is re-driven and unrecoverable state is not silently cleared.
- Direct and replay seams share the same classification helpers and disposition vocabulary.

## Operating rule

erskingardner:
Closes the remaining open children of tracking issue #707 (one convergence contract: every application-visible invariant enforced identically on the direct-ingest, stored-convergence/replay, and fork-recovery seams). Four commits, one per cluster.

Fixes #383
Fixes #394
Fixes #369
Fixes #364
Fixes #365
Fixes #339
Fixes #707

### Commit 1 — seam parity (#383)

The replay seam could emit `MessageReceived` with an empty, unauthenticated sender (S3 regression): the ApplicationMessage arm defaulted an unresolvable MLS sender to empty bytes and validated the inner event against `hex("")`, so a crafted `MarmotAppEvent` with `pubkey: ""` surfaced with a blank author. The per-seam validator pairs are collapsed into the shared chokepoints (`identity::member_id_of_sender`, typed `app_payload::validate_app_payload_for_sender` — which now rejects an empty `MemberId` outright), the replay arm pushes `Ignored` for anything that fails attribution (terminal via existing canonicalization), and `emit_application_replay_events` skips empty senders as a backstop. The mirror-on-all-seams contract is now documented as a hard invariant in `crates/cgka-engine/AGENTS.md`.

### Commit 2 — fail-closed hardening (#394, #369)

- The fork-recovery missing-snapshot branch no longer panics via `unreachable!()` on an invariant break reachable from attacker-delivered same-epoch fork commits; both arms share one fail-closed helper (`detect_fork` + audit + `EpochInvalidated` + typed `ForkedEpoch`).
- `persist_stored_message_payload`'s idempotency short-circuit now compares the encoded payload bytes, so a same-`(group, epoch, state)` persist under a different `StoredMessagePayload` variant overwrites instead of silently keeping the old variant. Verified independent of the #723 `OwnCommitWire` enrichment (which writes via `put_message` directly).

### Commit 3 — quarantine enforcement (#364, #365)

`quarantined_groups` is now enforced on every live surface through one chokepoint (`Engine::ensure_group_live`): all durable/MLS accessors return `UnknownGroup` (matching `epoch()` and the documented "vanish" contract — including `group_context()`, which previously exported a quarantined group's exporter secrets); `send`, convergence, and queued-intent drains refuse to run; inbound input is retained durably (`PeelDeferred`) and classified with the new `StaleReason::Quarantined`; nothing can `set_stable` a quarantined group out of band. Repair via `retry_hydrate_quarantined_group` or an authenticated re-join welcome clears the quarantine and replays the retained input.

### Commit 4 — deferred-peel lifecycle (#339, full scope)

`PeelDeferred` rows were retried forever at O(rows × snapshots) per convergence pass and grew the durable store without bound under a re-wrap flood (raw transport ids are attacker-controllable). The lifecycle is now: context-fingerprint gating (retry only when the (epoch, snapshot-set) peel context actually changes), a per-row retry budget → terminal `permanently_undecryptable`, a per-group cap of 256 retained rows (overflow dropped unpersisted), bounded 64-row sweeps so backlog never starves current events, and post-peel pre-membership classification via the new `Group.join_epoch` (`pre_membership_event` vs `valid_history_snapshot_missing`). One `MessageDisposition` taxonomy supplies the audit reason tags; `MessageStateChanged` gains optional `retry_count`/`sweeps_waited`, and each sweep emits an aggregate privacy-safe tracing line with queue depth and duration.

Note on #339's "identify pre-membership *before* peel" criterion: the transport wrapper carries no epoch, so pre-peel classification is physically impossible; the fix achieves the criterion's intent (no unbounded retries, no cryptographic amplification, terminal classification) via the gate + budget + post-peel classification.

### Verification

`just fast-ci` green; full workspace suite green (1992 tests, 0 failed); clippy clean across the workspace. New coverage: `app_payload` unit tests + `convergence_app_message_with_unresolvable_sender_emits_no_event_and_lands_terminal` (#383), `same_key_persist_with_different_payload_variant_overwrites` (#369), five quarantine-enforcement tests incl. the #707 acceptance-criterion test `quarantined_group_rejects_valid_inbound_commit_on_do_ingest` (#364/#365), and six lifecycle tests in the new `tests/deferred_peel_lifecycle.rs` (#339) with `send_preflight_retries_deferred_peels_after_convergence_apply` unmodified.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/726">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added hydration-quarantine handling: quarantined groups are blocked until recovered via retry hydration or authenticated re-join, with replay resuming afterward.
  * Enhanced deferred-peel lifecycle for more reliable retry/recovery, including bounded retries and improved convergence behavior.
* **Bug Fixes**
  * Empty or unresolvable senders are now rejected consistently, preventing invalid application events from being accepted or surfaced in group events.
  * Improved message storage idempotency so updated payload variants are preserved correctly.
  * Added more accurate group join-epoch metadata to improve join, replay, and recovery consistency.

### DIFF
--- a/crates/cgka-conformance-simulator/tests/openmls_replay_probe.rs
+++ b/crates/cgka-conformance-simulator/tests/openmls_replay_probe.rs
@@ -1432,6 +1432,7 @@ fn dummy_group(group_id: GroupId) -> Group {
         }],
         required_capabilities: GroupCapabilities::default(),
         removed: false,
+        join_epoch: EpochId(0),
     }
 }
 
--- a/crates/cgka-engine/AGENTS.md
+++ b/crates/cgka-engine/AGENTS.md
@@ -212,8 +212,12 @@ in `tests/`; this section exists so a future contributor can grep for the rule t
 - **Item:** **S3** Unattributable application messages
   - **What changed:** `message_processor::ingest_group_message` no longer emits `MessageReceived` with an empty
     `MemberId` for senders whose leaf credential cannot be resolved. The message is marked `Failed`; a typed sender
-    variant on `GroupEvent` would be a larger API change deferred to v0.2.
-  - **Test:** implicit (no event leak — covered by existing tests)
+    variant on `GroupEvent` would be a larger API change deferred to v0.2. Mirrored on the stored-convergence/replay
+    seam (mdk#383): the replay ApplicationMessage arm pushes `Ignored` for an unresolvable sender, both seams share
+    `app_payload::validate_app_payload_for_sender` (which rejects an empty `MemberId` outright), and
+    `emit_application_replay_events` skips empty-sender observations as a backstop.
+  - **Test:** `app_payload::tests` (empty-sender and pubkey-mismatch rejection);
+    `tests/distributed_convergence.rs::convergence_app_message_with_unresolvable_sender_emits_no_event_and_lands_terminal`
 
 - **Item:** **Sm1** Atomic state-machine transitions
   - **What changed:** `EpochManager::confirm_publish` / `rollback_publish` / `begin_pending` clone the prior
@@ -315,6 +319,19 @@ shapes. Tests: `tests/fork_detection.rs` plus the harness `deliberate_fork_via_h
 
 ## Conventions in this crate
 
+- **Mirror every ingest invariant on every inbound seam.** An inbound MLS message reaches application-visible state
+  through three seams — **direct ingest** (`message_processor/ingest.rs::ingest_group_message`),
+  **stored-convergence/replay** (`openmls_projection.rs::process_openmls_messages_inner`, materialization and apply),
+  and **fork recovery** (the `WrongEpoch` branch in `message_processor/ingest.rs` + `fork_recovery.rs`). Every
+  sender-authentication, admin/identity-proof, app-payload, and component-retention check MUST run identically on all
+  of them, through the *same* shared helper — never a seam-local re-implementation. Shared chokepoints:
+  `identity::member_id_of_sender` (MLS `Sender` → validated `MemberId`) and
+  `app_payload::validate_app_payload_for_sender` (payload + `&MemberId` → validated `MarmotAppEvent`; rejects an empty
+  id). An application message whose sender cannot resolve to a validated member id is never surfaced as
+  `MessageReceived` and never accepted into canonical state on any seam (direct: `Failed`; replay: `Ignored` →
+  terminal disposition). Fork-recovery paths fail closed with typed errors (`EngineError::ForkedEpoch` / `Backend`) —
+  never `unreachable!`/`panic!` — on attacker-influenced input. When you add a guard to one seam, add it to the shared
+  helper (or all seams) and extend the parity tests; a guard that exists on one seam only is a bug (see mdk#707).
 - **Only `EpochManager` may construct non-`Stable` `EpochState` variants.** This is enforced by visibility — the
   variants' fields are private. Don't add a public constructor for `Recovering` etc. somewhere else.
 - **No Nostr library/SDK dependency.** These crates do not depend on any Nostr crate and use no Nostr SDK types. They
--- a/crates/cgka-engine/AGENTS.md
+++ b/crates/cgka-engine/AGENTS.md
@@ -332,6 +332,16 @@ shapes. Tests: `tests/fork_detection.rs` plus the harness `deliberate_fork_via_h
   terminal disposition). Fork-recovery paths fail closed with typed errors (`EngineError::ForkedEpoch` / `Backend`) —
   never `unreachable!`/`panic!` — on attacker-influenced input. When you add a guard to one seam, add it to the shared
   helper (or all seams) and extend the parity tests; a guard that exists on one seam only is a bug (see mdk#707).
+- **Hydration quarantine is enforced through `Engine::ensure_group_live`.** A quarantined group vanishes from every
+  live surface (mdk#364 / #365): every public accessor that reads durable or MLS group state (`members`,
+  `group_record`, `group_context`, `admin_pubkeys`, `app_component`, `feature_status`, capability queries, the
+  safe-export family, `own_leaf_index`) calls `ensure_group_live` first and returns `UnknownGroup`; `do_send` and
+  `converge_and_drain_queued_outbound_intents` refuse to run; `ingest_group_message` retains inbound input as
+  `PeelDeferred` and classifies it `Stale { reason: Quarantined }`; `converge_stored_openmls_messages` reports a
+  `Blocked` run without touching state; `retry_deferred_peels` skips the group. When you add a new accessor or data
+  path that reads group state, add the gate — a path that bypasses it can silently un-quarantine a group via
+  `set_stable`. Quarantine clears only through `retry_hydrate_quarantined_group` or an authenticated re-join welcome,
+  both of which schedule retained input for replay.
 - **Only `EpochManager` may construct non-`Stable` `EpochState` variants.** This is enforced by visibility — the
   variants' fields are private. Don't add a public constructor for `Recovering` etc. somewhere else.
 - **No Nostr library/SDK dependency.** These crates do not depend on any Nostr crate and use no Nostr SDK types. They
--- a/crates/cgka-engine/src/app_payload.rs
+++ b/crates/cgka-engine/src/app_payload.rs
@@ -1,9 +1,19 @@
 use cgka_traits::{EngineError, MarmotAppEvent, MemberId};
 
+/// Shared application-payload sender validation for every inbound seam
+/// (direct ingest, stored-convergence/replay). An application message is
+/// surfaced only when its inner event's author matches the MLS-authenticated
+/// sender; an empty sender is rejected outright so an unattributable message
+/// can never validate, regardless of what the inner event claims.
 pub(crate) fn validate_app_payload_for_sender(
     payload: &[u8],
     sender: &MemberId,
 ) -> Result<MarmotAppEvent, EngineError> {
+    if sender.as_slice().is_empty() {
+        return Err(EngineError::InvalidAppMessagePayload(
+            "application message has no authenticated member sender".into(),
+        ));
+    }
     let event = MarmotAppEvent::decode(payload)
         .map_err(|err| EngineError::InvalidAppMessagePayload(err.to_string()))?;
     let sender_hex = hex::encode(sender.as_slice());
@@ -13,10 +23,47 @@ pub(crate) fn validate_app_payload_for_sender(
     Ok(event)
 }
 
-pub(crate) fn app_payload_is_valid_for_sender(payload: &[u8], sender: &[u8]) -> bool {
-    let event = match MarmotAppEvent::decode(payload) {
-        Ok(event) => event,
-        Err(_) => return false,
-    };
-    event.validate_sender(&hex::encode(sender)).is_ok()
+#[cfg(test)]
+mod tests {
+    use super::validate_app_payload_for_sender;
+    use cgka_traits::{EngineError, MarmotAppEvent, MemberId};
+
+    fn payload_from(pubkey: &str) -> Vec<u8> {
+        MarmotAppEvent::new(pubkey, 1, 9, vec![], "hello")
+            .encode()
+            .expect("encode app event")
+    }
+
+    #[test]
+    fn empty_sender_is_rejected_even_when_event_pubkey_is_empty() {
+        // Regression for the S3 replay-seam gap (#383): an event whose
+        // `pubkey` is the empty string must not validate against an empty
+        // (unresolvable) MLS sender.
+        let payload = payload_from("");
+        let result = validate_app_payload_for_sender(&payload, &MemberId::new(Vec::new()));
+        assert!(matches!(
+            result,
+            Err(EngineError::InvalidAppMessagePayload(_))
+        ));
+    }
+
+    #[test]
+    fn pubkey_mismatch_is_rejected() {
+        let sender = MemberId::new(vec![0x11; 32]);
+        let payload = payload_from(&hex::encode([0x22; 32]));
+        let result = validate_app_payload_for_sender(&payload, &sender);
+        assert!(matches!(
+            result,
+            Err(EngineError::InvalidAppMessagePayload(_))
+        ));
+    }
+
+    #[test]
+    fn matching_sender_validates() {
+        let sender = MemberId::new(vec![0x11; 32]);
+        let payload = payload_from(&hex::encode([0x11; 32]));
+        let event =
+            validate_app_payload_for_sender(&payload, &sender).expect("matching sender validates");
+        assert_eq!(event.content, "hello");
+    }
 }
--- a/crates/cgka-engine/src/audit_helpers.rs
+++ b/crates/cgka-engine/src/audit_helpers.rs
@@ -453,6 +453,8 @@ pub(crate) fn message_state_changed_event(
         new_state: message_state_str(state).to_string(),
         epoch: None,
         reason: reason.to_string(),
+        retry_count: None,
+        sweeps_waited: None,
     }
 }
 
@@ -470,6 +472,31 @@ pub(crate) fn message_state_transition_event(
         new_state: message_state_str(state).to_string(),
         epoch: epoch.map(|epoch| epoch.0),
         reason: reason.to_string(),
+        retry_count: None,
+        sweeps_waited: None,
+    }
+}
+
+/// Terminal transition of a deferred-peel row, carrying the lifecycle's
+/// per-row telemetry: how many re-peel attempts it consumed and how many
+/// retry sweeps it waited between first attempt and this transition
+/// (mdk#339).
+pub(crate) fn deferred_peel_terminal_event(
+    msg_id_hex: MessageRefHex,
+    epoch: Option<EpochId>,
+    reason: &str,
+    retry_count: u64,
+    sweeps_waited: u64,
+) -> AuditEventKind {
+    AuditEventKind::MessageStateChanged {
+        msg_id: msg_id_hex,
+        artifact_kind: None,
+        previous_state: Some(message_state_str(MessageState::PeelDeferred).to_string()),
+        new_state: message_state_str(MessageState::Failed).to_string(),
+        epoch: epoch.map(|epoch| epoch.0),
+        reason: reason.to_string(),
+        retry_count: Some(retry_count),
+        sweeps_waited: Some(sweeps_waited),
     }
 }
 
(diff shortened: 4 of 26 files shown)


erskingardner:
Adversarial review finding 1: `crates/cgka-engine/src/message_processor/mod.rs:793-815`

Including `MessageState::PeelDeferred` in `replay_buffered_messages` makes the re-join recovery path replay quarantine-retained inputs, but the success path still falls through `Ok(_) => {}`. If a retained raw row is successfully applied or terminally classified during `do_join_welcome` replay, the content-derived row is updated, but the original raw replay-buffer row remains durable `PeelDeferred` and its cap slot is never released. That leaves stale quarantine input in the retry lifecycle and can permanently consume the per-group cap after a re-join recovery.

Please mirror the retirement logic from `retry_deferred_peels`: when `was_peel_deferred` and the replay result is no longer buffered/quarantined, mark the raw row processed or otherwise terminal and call `note_peel_deferred_row_retired`.

erskingardner:
Adversarial review finding 2: `crates/cgka-engine/src/message_processor/mod.rs:535-564`

`retry_deferred_peels` leaves every `StaleReason::PeelFailed` result in `PeelDeferred`, but `ingest_group_message` uses `PeelFailed` for more than "still cannot peel". It also returns `PeelFailed` after the raw row has successfully peeled and the content-derived message is terminally rejected, for example invalid/unattributable application payloads. In those cases this retry loop consumes an attempt, leaves the raw row durable as `PeelDeferred`, records no progress, and can arm the unchanged-context gate, so terminal input remains retained until some unrelated context change or eventual budget exhaustion.

That undermines the new lifecycle promise that post-peel terminal classifications leave the deferred queue. Please distinguish "peel still failed" from "peeled and terminally rejected" here, or make the terminal `PeelFailed` paths mark/retire the raw deferred row the same way the pre-membership/too-distant paths already call `mark_raw_transport_message_failed_if_deferred`.

erskingardner:
Addressed the CodeRabbit review in ffaee58:

- **Misplaced doc comment (engine.rs, Minor)** — fixed. `set_recorder`'s doc block was split from it by the inserted `set_deferred_peel_retry_budget`; moved it back so each method carries its own doc.
- **Duplicate pre-membership condition (ingest.rs)** — fixed. Replaced `group_join_epoch` with a shared `msg_is_pre_membership` predicate used by both the pre-processing short-circuit and the `too_distant_in_the_past` refinement, so the join-epoch bound can't diverge.
- **Unthrottled `Rejection` audit under flood (ingest.rs)** — fixed. The flood-cap `Rejection` is now audited once per cap-full episode (re-armed via `note_peel_deferred_row_retired` once the backlog drops below the cap), so an attacker-paced flood past the cap no longer emits one audit write per rejected message. Applied to both the deferral-cap and quarantine-cap paths.
- **Duplicate test helpers (`route`/`welcome_for`/`evolution`) — skipped, with reason.** `tests/support/mod.rs` is compiled into 17 integration-test binaries and has no `#![allow(dead_code)]`, so shared-but-unused helpers would fail the `-D warnings` gate in the binaries that don't use them. The duplication is contained to test scaffolding, so I left the local copies in place rather than regress CI.

`just fast-ci` green; full `cgka-engine` suite green.

erskingardner:
Addressed both adversarial-review findings in 9701fa8. Both were real per-group cap-slot leaks — a raw `PeelDeferred` row kept alive (holding a cap slot) after the message was actually resolved.

**Finding 2 (retry seam) — fixed at the root.** `ingest_group_message` returns `StaleReason::PeelFailed` both when a peel genuinely still fails *and* when the row peeled successfully but the content was terminally rejected (unattributable sender, invalid app payload, non-MLS body, welcome-in-group-envelope, stale-epoch-no-snapshot). Each of those terminal-after-peel sites now calls `mark_raw_transport_message_failed_if_deferred` — exactly as the pre-membership / too-distant paths already do — so the raw row is marked `Failed` and released from the retry lifecycle (a no-op on the direct path). With that in place, `retry_deferred_peels`'s `PeelFailed` no-op arm is correct: by the time it sees `PeelFailed`, a peeled-and-terminal row is already retired, and a genuinely-still-deferred row is correctly left in place.

**Finding 1 (replay seam) — fixed.** `replay_buffered_messages`'s `Ok(_) => {}` arm now mirrors `retry_deferred_peels`: a `was_peel_deferred` row is retired (marked `Processed` + `note_peel_deferred_row_retired`) on apply/buffer and on terminal reclassification, and is left in place only for still-un-peelable `PeelFailed` or `Quarantined`. Non-deferred (`Created`/`Retryable`) rows keep their prior behavior.

Regression tests:
- `deferred_row_terminally_rejected_after_peel_leaves_the_deferred_queue` — a forged unattributable payload deferred at a future epoch, then re-peeled after catch-up and rejected, ends `Failed` (not stuck `PeelDeferred`).
- `rejoin_welcome_replays_and_retires_quarantine_retained_input` — input retained while quarantined and replayed by `do_join_welcome` leaves zero `PeelDeferred` rows.

`cgka-engine` suite green; clippy + fmt clean.

PR #727, opened 2026-07-04
Enforce strict tag extraction and post-verification id trust at the Nostr ingest boundary

### LINKED ISSUE #336: Gift-wrap recipient `p` tag is read with first-match `tag_value` instead of single-tag enforcement, unlike the strict group `h`-tag path
## Summary
kind:445 group events require exactly one `h` tag (`single_tag_value`), but kind:1059 gift wraps read the recipient `p` tag with first-match `tag_value`, so a malformed multi-`p` wrap is routed by an arbitrary tag rather than rejected. Decryption is still key-gated (not a spoofing bypass).

## Severity
**LOW** — event-validation (boundary asymmetry)

## Location
`crates/transport-nostr-peeler/src/event.rs:42-47 (to_transport_message gift-wrap arm) and peeler.rs:427 (ensure_welcome_routing_matches)`

## Details
For kind-445 group events the peeler requires exactly one `h` tag via single_tag_value (event.rs:38, rejecting duplicates). For kind-1059 gift wraps the recipient `p` tag is read with tag_value (first match) both at route mapping (event.rs:43) and in ensure_welcome_routing_matches (peeler.rs:427). A gift wrap carrying multiple `p` tags is silently routed by its first one with no rejection, an asymmetry with the group path's strict single-tag rule. Decryption is still gated by the recipient private key (extract_rumor), so this is not directly exploitable for spoofing, but the inconsistent strictness means a malformed multi-`p` wrap is accepted for routing and burns a delivery/peel attempt rather than being rejected at the boundary as the group path would do.

## Impact
Inconsistent boundary validation: malformed multi-recipient gift wraps are accepted and routed by an arbitrary (first) p tag rather than rejected, diverging from the strict single-tag rule applied to group messages. No decryption-bypass because the NIP-59 unwrap still requires the recipient key; impact is wasted delivery work and a validation gap if downstream ever trusts the routed recipient without re-checking.

## Suggested fix
Use single_tag_value(RECIPIENT_TAG) (or otherwise reject events with more than one `p` tag) in the gift-wrap arm of to_transport_message and in ensure_welcome_routing_matches, matching the group h-tag strictness.

---

### LINKED ISSUE #350: Welcome rumor `e`/`relays` tags read with first-match instead of single-tag enforcement
Found via review of the Nostr peeler welcome path. Related to marmot-protocol/mdk#336 / marmot-protocol/mdk#543.

## What

The group `h`-tag path is strict (`single_tag_value`), but the welcome rumor's `e` (KeyPackage id) and `relays` tags accept duplicates and silently take the first match. The sender is authenticated and the engine re-validates the Welcome, so this is hardening — but it is a third loose-tag instance on the authenticated-join path after the outer `p` tag (#606) and the KeyPackage id-list (#363).

## Where

- `crates/transport-nostr-peeler/src/peeler.rs:255-264` (`rumor_tag_value`/`rumor_tag_values`), helpers `:368-388`.

## Fix

Require exactly one `e` and one `relays` tag (reject duplicates), matching the strict `h`-tag path.

---

### LINKED ISSUE #351: Inbound event's self-reported `id` is trusted as the transport `MessageId` before id/signature verification (telemetry/audit poisoning)
Found via review of the Nostr transport adapter ingest/telemetry path.

## What

`to_transport_message` copies the event's self-reported `id` into `TransportMessage.id` (used for routing metrics, cross-relay telemetry keying, and forensic `wire_id`) without checking it matches the event hash; signature/id verification only happens later in the peeler, and never on the `observe_relay_event` telemetry path. A forged `id` on an otherwise-valid envelope poisons telemetry/audit. No decrypt/auth break (the engine re-verifies and marmot-protocol/marmot#126 already mandates a non-event-id dedup), hence LOW.

## Where

- `crates/transport-nostr-peeler/src/event.rs:32-64`; consumed in `crates/transport-nostr-adapter/src/lib.rs:381-429` and `:440-450`.

## Fix

Verify the event id in `to_transport_message` (or key diagnostics/telemetry on a post-verification value).

---

### LINKED ISSUE #392: transport-nostr-peeler: welcome rumor `relays` tag accepted unvalidated (count / length / scheme)
### Summary
`peel_welcome` only checks the `relays` tag is non-empty with no empty entries; the relay strings (count, length, URL scheme) are attacker-controlled and surfaced unvalidated.

### Location
`crates/transport-nostr-peeler/src/peeler.rs:258-264`; `rumor_tag_values` at `:378-388`.

### Details
The relay list comes from inside a NIP-59 rumor authored by the welcoming party. The seal/sig checks authenticate *who* sent it but not *what* the relay list contains. A malicious/compromised inviter can embed an arbitrarily large list of arbitrary strings (e.g. thousands of long non-`wss://` URLs) passed up as `WelcomeMetadata`.

### Impact
Potential SSRF / amplification if downstream layers attempt to connect to or persist these as relay endpoints.

### Suggested fix
Bound the number of relays and per-entry length, and validate each is a well-formed `ws://`/`wss://` relay URL before surfacing; reject otherwise. Related tracking issues: marmot-protocol/marmot#74, marmot-protocol/marmot#77.

---

---

### LINKED ISSUE #709: [Tracking] Strict Nostr ingress, identifier boundaries, and authenticated attribution
## Scope

Reopened after the 2026-07-21 audit. Nostr parsing and routing must enforce canonical shapes, keep transport routing identifiers distinct from opaque MLS group ids, normalize values once, and surface only attribution established by the authenticated inner protocol.

## Program checklist

- [x] #719 — transport-nostr-peeler: kind-445 group `h` tag decoded with unbounded `decode_hex` — missing the 32-byte length check every sibling id enforces
- [ ] #788 — transport-nostr-peeler: `PeeledMessage.group_id` is populated with the 32-byte transport routing id wrapped in a `GroupId`, conflating the two id spaces the invariant separates (latent)
- [ ] #795 — transport-nostr-adapter: relay sync-timing telemetry + initial-sync gate key the same relay under two different RelayIndex values (verbatim start endpoint vs RelayUrl-canonicalized inbound endpoint) — EOSE/first-event silently dropped, `subscription_synced` never reaches true
- [x] #815 — marmot-app: encrypted-media download integrity check compares ciphertext_sha256 case-sensitively — an uppercase-hex reference makes valid media permanently undownloadable
- [ ] #850 — traits: decode_nostr_routing_v1 passes usize::MAX as the relay-vector cap and enforces no relay-count bound, unlike every sibling app-component decoder (missing explicit cap; defense-in-depth/consistency)
- [x] #914 — traits: decode_components_list accepts non-canonical (unsorted) component-id order, unlike its strict-canonical sibling decoders
- [x] #951 — marmot-app: validate_fingerprint accepts uppercase hex, diverging from its lowercase-only sibling gossip validators — an uppercase sha256 fingerprint is persisted verbatim and diverges from strict peers
- [x] #953 — marmot-uniffi: endpoints() helper does not dedup (or trim) despite its doc contract, diverging from the sibling ffi_discovery_relays — duplicate relay endpoints flow into NIP-65 publication
- [x] #1006 — Security review L12: GroupJoined.welcomer is the NIP-59 seal author, not the MLS-validated welcome sender — invite attribution spoofing

## Contract

- Parse-time limits precede decoding/allocation.
- Canonical encodings and identifier namespaces are explicit types, not wrapper reuse.
- Endpoint/index normalization has one representation on start and inbound paths.
- Outer transport authors/timestamps never masquerade as MLS-authenticated attribution.

## Acceptance criteria

- Routing-id vs GroupId misuse is impossible at the public boundary.
- Relay/index and endpoint normalization is shared and tested with non-canonical inputs.
- Welcome attribution comes from the validated inner sender or is explicitly unknown.
- Decoder canonicality/length rules are consistent across app components and FFI.

## Operating rule

erskingardner:
## Summary

Lands the unified boundary contract from #709: **every routing-significant tag on a Nostr ingest path is extracted with single-tag enforcement and content-validated, and no self-reported identifier is trusted before verification.** The strict helper already guarded the kind-445 `h` tag; this brings the `p`, `e`, and `relays` tags and the event id up to the same contract instead of closing the gaps one tag at a time.

## Changes

**Strict single-tag extraction (#336, #350)**
- The kind-1059 gift-wrap recipient `p` tag now uses `single_tag_value` at route mapping (`event.rs to_transport_message`) and at peel (`peeler.rs ensure_welcome_routing_matches`); a multi-`p` wrap is rejected, not routed by an arbitrary first tag.
- The welcome rumor `e` (KeyPackage id) and `relays` tags are extracted with new `rumor_single_tag_*` helpers that reject duplicate tags; the loose first-match `rumor_tag_value`/`rumor_tag_values` helpers are gone.

**Welcome `relays` content validation (#392)**
- `validate_welcome_relays` bounds the count (`MAX_WELCOME_RELAYS = 16`, parity with `marmot-app`'s `MAX_RELAY_ENDPOINTS_PER_ROUTE`), bounds each entry (`MAX_WELCOME_RELAY_URL_LEN = 512`), and requires well-formed ws/wss URLs via `RelayUrl::parse`.
- Applied on peel (attacker-controlled input fails closed as `Malformed`) and on wrap (`WrapFailed` — the peeler never emits a rumor it would reject on ingest).

**Post-verification id trust (#351)**
- `to_transport_message` recomputes the NIP-01 event id from the event's own fields (new `computed_id()`, reusing the existing `pre_signing_id`) and rejects a mismatched self-reported id before it becomes `TransportMessage.id`.
- Both adapter ingest paths go through this conversion, so a forged id can no longer reach routing, cross-relay telemetry keying, or the forensic `wire_id` — including the `observe_relay_event` telemetry-only path, which previously never verified at all. Signature verification stays at peel time.

**Contract documentation**
- `crates/transport-nostr-peeler/AGENTS.md` gains a "Boundary validation contract" section so a future tag wired up with a loose helper is a visible contract violation.

## Tests

- Multi-`p` gift wraps rejected at route mapping and at peel (routing check runs before signature verification, so a tampered wrap dies at the boundary).
- Duplicate `e` / duplicate `relays` rumor tags rejected.
- Oversized relay lists (17+), over-long entries, `https://` scheme, garbage URLs, and empty lists rejected at peel; same violations rejected at wrap.
- Forged-id events: rejected at route mapping; `handle_relay_event` fails closed with no delivery; `observe_relay_event` records zero telemetry state.
- `computed_id` parity with SDK-signed event ids locked by a test including JSON-escaped content; all existing wrap→peel round trips exercise it implicitly.
- Fabricated-id test fixtures across `transport-nostr-adapter`, `marmot-app`, and the peeler now compute real ids (distinguishing markers moved into content so ids stay distinct).

## Verification

- `cargo test -p transport-nostr-peeler` (28), `-p transport-nostr-adapter` (+ `--features sdk`), `-p marmot-app` (305+), `-p cgka-session` — all green.
- `just fast-ci` passes.

Closes #336
Closes #350
Closes #392
Closes #351
Closes #709

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/727">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for inbound Nostr events and welcome messages, rejecting forged event IDs, duplicate or missing required tags, and invalid relay data.
  * Improved routing checks so message IDs are verified against recomputed event hashes before being accepted.
  * Tightened recipient handling for gift-wrapped messages to prevent ambiguous tag parsing.
* **Tests**
  * Added and updated regression coverage for forged IDs, duplicate tags, relay validation, and hash computation consistency.

### DIFF
--- a/crates/transport-nostr-peeler/src/event.rs
+++ b/crates/transport-nostr-peeler/src/event.rs
@@ -30,6 +30,15 @@ pub struct NostrTransportEvent {
 impl NostrTransportEvent {
     /// Convert a Nostr event into the raw transport message the engine ingests.
     pub fn to_transport_message(&self) -> Result<TransportMessage, NostrPeelerError> {
+        // #709/#351 — the resulting `TransportMessage.id` keys routing metrics,
+        // telemetry, and the forensic `wire_id`, so bind it to the event hash
+        // here rather than trusting the self-reported id. Signature
+        // verification still happens at peel time.
+        if !self.id.eq_ignore_ascii_case(&self.computed_id()) {
+            return Err(NostrPeelerError::Malformed(
+                "event id does not match event hash".into(),
+            ));
+        }
         let id = MessageId::new(decode_hex_exact("event id", &self.id, 32)?);
         let envelope = match self.kind {
             KIND_MARMOT_GROUP_MESSAGE => {
@@ -39,9 +48,7 @@ impl NostrTransportEvent {
                 }
             }
             KIND_NIP59_GIFT_WRAP => {
-                let recipient = self
-                    .tag_value(RECIPIENT_TAG)
-                    .ok_or_else(|| NostrPeelerError::MissingTag(RECIPIENT_TAG.into()))?;
+                let recipient = self.single_tag_value(RECIPIENT_TAG)?;
                 TransportEnvelope::Welcome {
                     recipient: MemberId::new(decode_hex_exact("recipient p tag", recipient, 32)?),
                 }
@@ -135,6 +142,19 @@ impl NostrTransportEvent {
         }
     }
 
+    /// NIP-01 event id (lowercase hex sha256 of the canonical serialization)
+    /// computed from this event's own fields, independent of the self-reported
+    /// `id` field.
+    pub fn computed_id(&self) -> String {
+        pre_signing_id(
+            &self.pubkey,
+            self.created_at,
+            self.kind,
+            &self.tags,
+            &self.content,
+        )
+    }
+
     /// Build an unsigned local Nostr DTO and precompute the event id for the
     /// supplied unsigned event core.
     pub fn new_unsigned(
@@ -212,8 +232,8 @@ mod tests {
 
     #[test]
     fn kind_445_event_maps_to_group_transport_message() {
-        let event = NostrTransportEvent {
-            id: "11".repeat(32),
+        let mut event = NostrTransportEvent {
+            id: String::new(),
             pubkey: "22".repeat(32),
             created_at: 1_700_000_000,
             kind: KIND_MARMOT_GROUP_MESSAGE,
@@ -221,10 +241,14 @@ mod tests {
             content: "encrypted body".into(),
             sig: None,
         };
+        event.id = event.computed_id();
 
         let msg = event.to_transport_message().expect("event maps");
 
-        assert_eq!(msg.id.as_slice(), vec![0x11; 32].as_slice());
+        assert_eq!(
+            msg.id.as_slice(),
+            hex::decode(&event.id).unwrap().as_slice()
+        );
         assert_eq!(msg.timestamp.0, 1_700_000_000);
         assert_eq!(msg.source.0, NOSTR_SOURCE);
         // The peeler does not extract `e` causal-dependency tags (kind 445 carries
@@ -273,8 +297,8 @@ mod tests {
 
     #[test]
     fn kind_1059_route_mapping_defers_signature_verification_to_peeling() {
-        let event = NostrTransportEvent {
-            id: "33".repeat(32),
+        let mut event = NostrTransportEvent {
+            id: String::new(),
             pubkey: "44".repeat(32),
             created_at: 1_700_000_001,
             kind: KIND_NIP59_GIFT_WRAP,
@@ -282,6 +306,7 @@ mod tests {
             content: "gift wrap body".into(),
             sig: None,
         };
+        event.id = event.computed_id();
 
         let msg = event
             .to_transport_message()
@@ -293,5 +318,73 @@ mod tests {
                 recipient: MemberId::new(vec![0x55; 32]),
             }
         );
+    }
+
+    #[test]
+    fn route_mapping_rejects_forged_event_id() {
+        // #351 — `TransportMessage.id` keys routing/telemetry/forensics, so a
+        // self-reported id that does not match the event hash fails closed.
+        let event = NostrTransportEvent {
+            id: "33".repeat(32),
+            pubkey: "22".repeat(32),
+            created_at: 1_700_000_000,
+            kind: KIND_MARMOT_GROUP_MESSAGE,
+            tags: vec![vec!["h".into(), "aa55".into()]],
+            content: "encrypted body".into(),
+            sig: None,
+        };
+        assert_ne!(event.id, event.computed_id());
+
+        assert!(matches!(
+            event.to_transport_message(),
+            Err(NostrPeelerError::Malformed(_))
+        ));
+    }
+
+    #[test]
+    fn route_mapping_rejects_duplicate_recipient_tags() {
+        // #336 — gift-wrap `p` extraction is strict single-tag, matching the
+        // kind-445 `h` path; a multi-`p` wrap is rejected, not first-matched.
+        let mut event = NostrTransportEvent {
+            id: String::new(),
+            pubkey: "44".repeat(32),
+            created_at: 1_700_000_001,
+            kind: KIND_NIP59_GIFT_WRAP,
+            tags: vec![
+                vec!["p".into(), "55".repeat(32)],
+                vec!["p".into(), "66".repeat(32)],
+            ],
+            content: "gift wrap body".into(),
+            sig: None,
+        };
+        event.id = event.computed_id();
+
+        assert!(matches!(
+            event.to_transport_message(),
+            Err(NostrPeelerError::Malformed(_))
+        ));
+    }
+
+    #[test]
+    fn computed_id_matches_sdk_signed_event_id() {
+        // `to_transport_message` verifies self-reported ids against
+        // `computed_id`, so the local NIP-01 id computation must agree with the
+        // Nostr SDK's — including for content that needs JSON escaping.
+        let content = "line\nbreak \"quote\" back\\slash tab\t unicode ✨ control \u{1}";
+        let signed = nostr::EventBuilder::new(
+            nostr::Kind::Custom(KIND_MARMOT_GROUP_MESSAGE as u16),
+            content,
+        )
+        .tags([nostr::Tag::custom(
+            nostr::TagKind::custom("h"),
+            [hex::encode([0x99; 32])],
+        )])
+        .sign_with_keys(&nostr::Keys::generate())
+        .expect("sign kind-445");
+
+        let dto = NostrTransportEvent::from_nostr_event(&signed).unwrap();
+
+        assert_eq!(dto.id, dto.computed_id());
+        assert!(dto.to_transport_message().is_ok());
     }
 }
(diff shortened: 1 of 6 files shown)


crates/transport-nostr-peeler/src/event.rs
```diff
@@ -30,6 +30,15 @@ pub struct NostrTransportEvent {
...
+        // here rather than trusting the self-reported id. Signature
+        // verification still happens at peel time.
+        if !self.id.eq_ignore_ascii_case(&self.computed_id()) {
+            return Err(NostrPeelerError::Malformed(
+                "event id does not match event hash".into(),
+            ));
+        }
         let id = MessageId::new(decode_hex_exact("event id", &self.id, 32)?);
         let envelope = match self.kind {
             KIND_MARMOT_GROUP_MESSAGE => {
```
erskingardner, line 44: This kind-445 `h` path still accepts any hex length because the value is passed to `decode_hex` below. The protocol text defines `h` as the lowercase hex encoding of the 32-byte `nostr_group_id`, and this PR's new boundary contract says ids are hex/length-checked, but a signed event with `h = "aa55"` still routes into `TransportEnvelope::GroupMessage` (the updated unit test currently locks that in). Since this is routing-significant boundary input, this should use `decode_hex_exact(..., 32)` or otherwise reject non-32-byte group route ids.

erskingardner (commented):
Found two boundary-validation gaps that look worth fixing before this lands. Both are in the exact contract this PR is trying to make explicit.

crates/transport-nostr-peeler/src/event.rs
```diff
@@ -39,9 +48,7 @@ impl NostrTransportEvent {
                 }
             }
             KIND_NIP59_GIFT_WRAP => {
-                let recipient = self
-                    .tag_value(RECIPIENT_TAG)
-                    .ok_or_else(|| NostrPeelerError::MissingTag(RECIPIENT_TAG.into()))?;
```
erskingardner, line 44: The duplicate-`p` hardening is still bypassable with a malformed duplicate tag. `single_tag_value()` delegates to `tag_values()`, and `tag_values()` drops matching tags that do not have `tag[1]`; an event containing both `["p"]` and `["p", <valid recipient>]` is therefore treated as having exactly one `p` value and routes successfully. That violates the new "exactly one tag" contract for routing-significant tags (and the same helper is used for `h`). The helper should count matching tag occurrences first, then validate the single tag has the expected value.

crates/transport-nostr-peeler/src/event.rs
erskingardner: Fixed in bf0fbef — the `h` value now goes through `decode_hex_exact("group h tag", .., 32)` at route mapping, and `ensure_group_routing_matches` applies the same check on the peel-side re-parse. The unit test that locked in `h = "aa55"` now uses a 32-byte route id, and a new `route_mapping_rejects_non_32_byte_group_route_id` regression test drives the short-id rejection.

crates/transport-nostr-peeler/src/event.rs
erskingardner: Fixed in bf0fbef — `single_tag_value` now counts tag *occurrences* (any tag whose first element matches the name) before extracting the value, so `["p"]` + `["p", <valid>]` is rejected as a duplicate, and a single valueless `["p"]` is `Malformed` ("tag has no value") rather than silently dropped. The `h` path inherits the same fix, and the rumor-side `rumor_single_tag_*` helpers already counted occurrences this way. New `single_tag_enforcement_counts_valueless_duplicate_tags` test covers all three cases.

erskingardner:
Review fixes landed in bf0fbef (32-byte `h` route ids; occurrence-counting single-tag enforcement).

On the two CodeRabbit nitpicks, deferring both:
- **Shared relay-limit constant**: `marmot-app` depends on `transport-nostr-peeler`, not the reverse, so the peeler can't import `MAX_RELAY_ENDPOINTS_PER_ROUTE`; hoisting a shared constant into `traits` would couple the welcome-rumor bound to the relay-route bound, which are separate policies that happen to agree today. The parity is documented at both sites.
- **Unifying `rumor_single_tag` with `single_tag_value`**: the two operate on different tag representations (`Vec<Vec<String>>` DTO vs `nostr::Tag` on the unwrapped rumor) and different error types; abstracting over both costs more than the ~15 duplicated lines. Both now enforce the same occurrence-counting semantics, locked by tests on each side.

Issue #710, opened 2026-07-03
[Tracking] QUIC preview correctness, crypto, and broker availability


erskingardner:
## Scope

Expanded after the 2026-07-21 audit. The direct and broker preview transports remain advisory, but must preserve nonce uniqueness, forward compatibility, multi-subscriber behavior, bounded connection/room lifetimes, and deadline-backed reads/writes.

## Program checklist

- [ ] #730 — transport-quic-stream: direct-path client binds 127.0.0.1, so a direct sender can only reach loopback receivers
- [x] #742 — closed as not applicable: accepting a mid-stream first sequence would violate adopted gap/transcript semantics
- [ ] #758 — transport-quic-broker: unauthenticated subscribers can squat all room slots (hold streams open) and lock out publishers with RoomLimitExceeded
- [x] #767 — transport-quic-broker: publish re-encodes each retained record solely to measure its size (double serialization on the replay path)
- [ ] #779 — transport-quic-broker: subscribe-fanout write path has no write deadline — a stalled (non-reading) subscriber pins the handler task and holds its stream + connection permits
- [x] #787 — transport-quic-stream: agent-text-stream record AEAD has no per-session salt — re-publishing to the same anchor within an epoch reuses the ChaCha20-Poly1305 (key, nonce) pair
- [ ] #797 — transport-quic-broker: publish hot path runs a full O(max_rooms) room sweep + total-backlog recompute under the single global BrokerStateInner mutex on every forwarded record
- [x] #816 — transport-quic-stream/broker: an unknown record_type is fatal on decode, defeating documented ignore-unknown degradation and making stream_record_text's unknown-type arm dead code
- [ ] #843 — transport-quic-stream: direct receiver accepts a single connection with no accept loop — a connect-first peer consumes the only accept slot and denies the legitimate sender
- [ ] #865 — transport-quic-broker: a connection that completes the handshake but opens no stream pins its connection permit indefinitely (server keepalive defeats the idle timeout) — connection-slot exhaustion DoS
- [ ] #899 — transport-quic-broker: under global backlog byte-pressure, `publish` evicts from the currently-publishing room — discards the active stream's freshest replay records while stale quiet-room data survives
- [ ] #942 — transport-quic-stream/broker: preview sender/publisher record writes have no write deadline — an untrusted receiver/broker that stops reading pins the sender task forever (keepalive defeats the idle timeout)
[...]

Issue #334, opened 2026-06-24
Broker per-publish-stream record/byte limits are hardcoded `default()`, silently truncating legitimate long live previews


erskingardner:
## Summary
`handle_publish_stream` builds `AgentTextStreamReceiveAccumulator::default()` (4096 records / ~1 MiB), not derived from group policy or broker config; exceeding it finishes the room mid-response. The bound also counts ciphertext, not plaintext.

## Severity
**LOW** — bug (availability cliff)

## Location
`crates/transport-quic-broker/src/handlers.rs:72, crates/transport-quic-broker/src/handlers.rs:86-92`

## Details
handle_publish_stream constructs `AgentTextStreamReceiveAccumulator::default()` (handlers.rs:72), which applies AGENT_TEXT_STREAM_DEFAULT_MAX_RECORDS (4096) and AGENT_TEXT_STREAM_DEFAULT_MAX_PLAINTEXT_BYTES (1 MiB). These limits are NOT derived from the group's AgentTextStreamQuicPolicyV1 and are unrelated to the broker's configured backlog budgets. When a publisher exceeds 4096 records or ~1 MiB of cumulative frame bytes in a single stream, observe() returns a limit error (handlers.rs:90), the publish loop terminates, and the room is finished — cutting off the live preview mid-response. Additionally, the accumulator observes the wire record whose plaintext_frame is actually ciphertext (the broker never decrypts), so the 1 MiB "plaintext" bound is really a 1 MiB ciphertext bound. A long agent turn (many small deltas or a >1 MiB streamed answer) realistically hits this and silently kills the preview for the rest of the response, with the composer latching live_error.

## Impact
Long or chatty agent turns have their live preview terminated at 4096 records / ~1 MiB of forwarded bytes regardless of operator backlog configuration or group policy, degrading the feature for exactly the long-running responses it targets. It is an availability/correctness cliff, not a memory-safety issue (the cap does bound memory).

## Suggested fix
Drive the publish-stream accumulator from operator/broker config (or omit per-stream record/byte caps and rely on the existing backlog-byte, room, and stream-count budgets plus QUIC idle reaping), and document that the value counts ciphertext bytes, not plaintext.

---
*Filed by an automated, meticulous code review of `master` (scope: bugs, security, performance, refactor). The cited `file:line` locations were verified against source. Checked against existing open/closed issues for duplicates before filing; dedup notes included where a related issue exists.*

erskingardner:
[...]

PR #729, opened 2026-07-04
Share one hardening profile between the direct and broker QUIC preview transports (#710)

### LINKED ISSUE #332: QUIC preview broker enables unbounded TLS 0-RTT early data, making publish control frames network-replayable (no app-layer anti-replay)
## Summary
The broker sets `max_early_data_size = u32::MAX` (server) and `enable_early_data = true` (client). TLS 1.3 0-RTT is replayable, and the broker has no application-layer anti-replay, so captured publish control frames can be replayed to create/feed rooms and burn budgets. The direct-path stream crate deliberately omits early data.

## Severity
**MEDIUM** — security (replay / DoS)

## Location
`crates/transport-quic-broker/src/tls.rs:82, crates/transport-quic-broker/src/tls.rs:141`

## Details
broker_server_config sets `crypto.max_early_data_size = u32::MAX` (line 82) and client_endpoint sets `crypto.enable_early_data = true` (line 141), enabling QUIC/TLS1.3 0-RTT early data on every broker connection. TLS 0-RTT data carries no forward-secrecy and, by the TLS 1.3 design, is replayable by a passive network attacker who captures the first flight. The broker performs no application-layer anti-replay: a publish stream's control envelope (QuicBrokerControlEnvelopeV1) and the first record frame(s) can be sent as early data and replayed. Because the broker is an unauthenticated, memory-only pub/sub that routes purely by (stream_id, start_event_id) carried in the replayable control frame, a replayed early-data flight lets an off-path attacker (re)create/feed a room and consume room/stream/backlog budget without completing a fresh handshake. The direct-path stream crate (transport-quic-stream/src/tls.rs) deliberately does NOT enable early data, so this is a broker-only divergence. The replayed records are AEAD-sealed and seq-deduped by subscribers, so message injection into the decrypted transcript is not the concern; resource/room creation and amplification via cheap replay is.

## Impact
A passive network adversary can replay captured 0-RTT flights to spin up or feed broker rooms and burn the max_rooms / per-connection-stream / backlog budgets with no handshake cost, aiding a memory/room-exhaustion DoS against the public broker. No anti-replay or idempotency layer mitigates it.

Verified: `transport-quic-broker/src/tls.rs:82` (`max_early_data_size = u32::MAX`) and `:141` (`enable_early_data = true`); `grep early_data` in `transport-quic-stream/src/tls.rs` returns nothing (direct path does not enable it).

## Suggested fix
Disable 0-RTT on the broker: drop `crypto.enable_early_data = true` on the client and set `max_early_data_size = 0` (the default) on the server, OR gate early data behind an application-level anti-replay token / single-use nonce in the control envelope. Match the direct-path stream crate, which already omits early data.

---

### LINKED ISSUE #354: Broker client connect (publish and subscribe) has no application-level connect timeout
Found via review of the QUIC preview broker client. MEDIUM in combination with marmot-protocol/mdk#331.

## What

`endpoint.connect(...).await` has no `tokio::time::timeout` wrapper and `tls::client_endpoint` installs no client `TransportConfig`, so only quinn defaults bound a hung connect. In particular the **subscribe** path (the receiver-side broker watch flagged by marmot-protocol/mdk#331 as connecting to a sender-controlled `quic://` candidate) has no abort/timeout, unlike the publish/compose side.

## Where

- `crates/transport-quic-broker/src/client.rs:88-92` and `:260-263`.

## Fix

Wrap both `connect().await` calls in `tokio::time::timeout` and set a bounded client `max_idle_timeout` in the client `TransportConfig`.

---

### LINKED ISSUE #368: transport-quic-stream: direct QUIC receiver has no timeout on accept/`accept_uni` and installs no `TransportConfig` — unauthenticated peer can pin a receive task/fd pre-stream
## Summary
The direct QUIC receiver awaits the connection handshake and `accept_uni()` with **no application-level deadline**, and `configure_server` installs **no `TransportConfig`** (no idle timeout / keep-alive / stream caps), unlike the broker. A peer that completes the handshake (no client auth required) but never opens the unidirectional stream parks the receive task — and its UDP endpoint/fd — indefinitely.

## Where
```rust
// crates/transport-quic-stream/src/receive.rs:64-72
let incoming = self.endpoint.accept().await.ok_or(QuicTextStreamError::EndpointClosed)?;
let connection = incoming.await?;          // no timeout
let mut recv = connection.accept_uni().await?;  // no timeout
```
`configure_server` (`crates/transport-quic-stream/src/tls.rs:16-37`) never calls `server_config.transport_config(...)`, so it relies entirely on quinn defaults — contrast the broker's `broker_transport_config` (`max_idle_timeout`, `keep_alive_interval`, concurrent-stream caps). (Confirmed: no `timeout`/`Duration`/`deadline` in `receive.rs`.)

## Impact
A single peer pins a receive task + fd by stalling between handshake and stream open; the only backstop is quinn's default idle timeout, which an operator cannot tune. Also permits many concurrent streams up to the default cap. The marmot-protocol/mdk#432/#682 read-loop bounds do not cover this pre-stream phase.

## Suggested fix
Wrap `incoming.await` and `accept_uni().await` in an explicit timeout, and install a `TransportConfig` on the direct server (idle timeout, keep-alive, `max_concurrent_uni_streams`) mirroring the broker.

## Dedup
- **#658 / marmot-protocol/mdk#420** (open): the *record read loop* (dedup-before-observe, per-record deadline) — different primitive (stall is before any `read_record`).
- **#384** (closed): direct path missing ALPN — different config surface.

---

---

### LINKED ISSUE #370: transport-quic-stream / broker: a non-UTF-8 frame on a text-bearing record type is stream-fatal, discarding the whole received preview
## Summary
`stream_record_text` returns `Err(Utf8)` for any `TextDelta`/`Status`/`ProgressDelta`/`Checkpoint` record whose decrypted frame is not valid UTF-8, and that error propagates out of the receive loop via `?`, tearing down the entire stream and discarding every accumulated chunk.

## Where
- `crates/transport-quic-stream/src/receive.rs:113` and `:156-168` (`stream_record_text`).
- Same pattern in `crates/transport-quic-broker/src/client.rs:326` and `:364-376`.
```rust
let frame_text = stream_record_text(&record)?;   // '?' aborts the whole receive
```

## Impact
A single malformed (authentic-but-non-UTF-8) frame — or, on the broker subscribe path, an injected frame from a forged publisher (#651) at the right `seq` — tears down a whole live preview instead of skipping/flagging the one bad frame. For an advisory, provisional preview transport, a per-frame decode failure should not be terminal.

## Suggested fix
Treat a per-frame UTF-8 decode failure as a skippable/lossy frame (or render a placeholder) rather than a stream-fatal error.

## Dedup
- **#658/#661/#682** (read-loop limits/timeouts) and **#651** (forged records via sequencing) are different failure modes (valid sequencing + valid AEAD, invalid text bytes). No related issue found.

---

---

### LINKED ISSUE #371: transport-quic-broker: control-frame reader allows a ~66 KB pre-auth allocation for a frame whose only valid size is ~155 bytes
## Summary
`read_control_frame` reads the first frame on every (unauthenticated, post-TLS) broker stream using `MAX_FRAME_SIZE` (≈ 66 KB) as the length cap, then allocates `vec[0u8; len]` and reads `len` bytes before structurally rejecting anything malformed — but a valid control envelope is at most ~155 bytes.

## Where
- `crates/transport-quic-broker/src/frame.rs:22-30` (`read_control_frame` → `read_bytes_frame`).
- Cap from `crates/transport-quic-broker/src/protocol.rs:42-43` (`MAX_FRAME_SIZE` = `AGENT_TEXT_STREAM_MAX_PLAINTEXT_FRAME_LEN + 1024`).

A valid `QuicBrokerControlEnvelopeV1` is ≤ ~155 bytes (21-byte protocol string + 1 type + ≤64 stream_id + ≤64 start_event_id + varints). A peer can declare a 66 KB control frame on each of up to `max_streams_per_connection` × `max_connections` (default 64 × 256) streams.

## Impact
Bounded but unnecessary pre-auth heap pressure (~66 KB × stream count) and wasted reads before the cheap structural reject.

## Suggested fix
Introduce a dedicated `MAX_CONTROL_FRAME_LEN` (~256 bytes) and reject oversized control frames at the length check.

## Dedup
- **#212** (closed): the *agent-control* crate's *unbounded* pre-size-check buffer — this broker path is bounded but mis-sized. No broker-control issue found.

---

---

### LINKED ISSUE #389: transport-quic-broker: entire retained backlog (up to 64 MiB) is deep-cloned while holding the global broker mutex
### Summary
`subscribe` clones the room's whole retained backlog while holding the single mutex that guards every room, so a subscribe to a full-backlog room can stall all publish/forward progress broker-wide.

### Location
`crates/transport-quic-broker/src/state.rs:126-146` (`subscribe`)

### Details
```rust
let backlog: Vec<_> = room.backlog.iter().map(|e| e.record.clone()).collect();
```
runs while holding `self.inner` (`Mutex<BrokerStateInner>`, the lock for all rooms). With a non-zero `replay_ttl`, backlog can hold up to `max_backlog` (1024) records / `max_backlog_bytes` (64 MiB). Every `publish`, `subscribe`, `unsubscribe`, and cleanup contends on this one mutex.

### Impact
A subscriber joining a full-backlog room copies up to 64 MiB under the lock, blocking publish/forward across **every** room for the duration of the copy. A remote peer can repeatedly subscribe to a busy retained room to induce broker-wide head-of-line blocking. (Distinct from marmot-protocol/mdk#372, which is a backlog-byte accounting bug on the same path.)

### Suggested fix
Take a cheap snapshot under the lock (store records as `Arc`, hand out `Arc<[Record]>`, or move the `VecDeque` clone off-lock) so the large copy happens after the mutex is released.

---

---

### LINKED ISSUE #391: transport-quic-broker: receive-side record/byte caps are forced onto the publish stream, truncating legitimate long previews
### Summary
`handle_publish_stream` enforces the receiver default caps (`max_records = 4096`, `max_plaintext_bytes = 1 MiB`) on the publisher, cutting off legitimate long agent previews mid-stream.

### Location
`crates/transport-quic-broker/src/handlers.rs:72,86-92`; defaults at `crates/transport-quic-stream/src/limits.rs:31-32`.

### Details
```rust
let mut limit_state = AgentTextStreamReceiveAccumulator::default();
while let Some(record) = read_record_frame(&mut recv, None, MAX_FRAME_SIZE).await? {
    if record.stream_id != key.stream_id { return Err(QuicBrokerError::MixedStreamIds); }
    limit_state.observe(&record)?;   // <-- receiver-default caps applied to the publisher
    state.publish(&key, record).await?;
}
```
These are *receiver* defaults meant to bound what a subscriber accepts, not what a broker forwards. When a publisher exceeds either, `observe` returns `Err`, the loop ends, and the room is closed.

### Impact
A legitimate long-running agent preview producing more than 4096 records or more than 1 MiB of cumulative deltas is silently terminated at the broker mid-stream — exactly the "live preview dies for the rest of the response" failure the surrounding comment elsewhere says it is trying to avoid. There is no operator knob to raise it (unlike subscriber paths, which take `limits`).

### Suggested fix
Make publish-side limits configurable via `QuicBrokerConfig` (or derive from group policy), separate from the receiver default; or cap only per-record size and rely on `max_connections` / `max_rooms` / idle-timeout for aggregate bounds.

---

---

erskingardner:
## Summary

Implements tracking issue #710 as one bundle: the direct (`transport-quic-stream`) and broker (`transport-quic-broker`) QUIC preview transports now derive transport config, early-data policy, connect/read deadlines, frame caps, and the per-frame decode policy from **one shared hardening profile** (`transport_quic_stream::hardening`), and receive-side limits are never imposed on the publish/forward role.

Closes #332
Closes #354
Closes #368
Closes #370
Closes #371
Closes #389
Closes #391

(#334 was closed manually as a duplicate of #391.)

## What changed

**New shared module `transport-quic-stream/src/hardening.rs`**
- `QuicPreviewTransportProfile` builds the `TransportConfig` for every endpoint role: direct server (idle 30s / keepalive 10s / uni=1 / bidi=0), broker server (operator-tunable stream caps), and all clients (no peer-initiated streams). Broker public defaults alias the shared constants; `MAX_FRAME_SIZE` is unified as `QUIC_PREVIEW_MAX_FRAME_LEN`.
- `connect_with_timeout` bounds every preview dial (connect + TLS handshake) to `QUIC_PREVIEW_CONNECT_TIMEOUT` (15s, zero disables).
- Module docs pin the early-data policy: 0-RTT stays off everywhere (no app-layer anti-replay exists); tests in both crates enforce it.

**Broker (`transport-quic-broker`)**
- #332 — 0-RTT disabled: removed `max_early_data_size = u32::MAX` (server) and `enable_early_data = true` (client); rustls config construction factored into testable helpers.
- #354 — both client dials (`BrokerTextPublisher::connect`, subscribe) use `connect_with_timeout`; the client endpoint installs the shared client transport profile.
- #371 — control frames capped at `MAX_CONTROL_FRAME_LEN` (256 B; a maximal valid envelope is 155 B) instead of ~66 KB pre-auth.
- Comment gap A — `read_bytes_frame` wraps the WHOLE frame (length prefix + body) in one deadline; per-`read()` windows no longer restart on each dribbled byte.
- #391 — publish streams get forward-role limits from `QuicBrokerConfig` (`publish_max_records` = 65536, `publish_max_plaintext_bytes` = 64 MiB; new `marmot-quic-broker` flags, echoed in startup output), never the subscriber-sized receive defaults.
- Publish record reads now carry the shared 120s quiet-gap deadline (`RECORD_QUIET_GAP_DEADLINE`), matching the direct path and subscriber loop — this deliberately reverses the documented "no per-record deadline on publish" decision (maintainer-approved): QUIC keepalives sustain an alive-but-wedged publisher forever, and the transport contract already assumes ≤120s gaps on the other two paths. The server also bounds the TLS handshake in its accept task.
- #389 — backlog + subscriber queues carry `Arc<AgentTextStreamRecordV1>`; a subscribe snapshots a full backlog with pointer clones under the global mutex, and fan-out stops deep-cloning per subscriber.

**Direct path (`transport-quic-stream`)**
- #368 residual — `endpoint.accept()` is bounded by a new `accept_timeout` receive limit (default 300s, `0` disables — generous because `wn stream receive` is a manual out-of-band rendezvous) with a distinct `AcceptTimeout` error; the handshake falls under the setup `read_timeout`.
- Comment gap B — the client endpoint installs the shared client transport profile and the outbound dial is bounded.

**Both paths**
- #370 — one shared, lossy `stream_record_text`: a non-UTF-8 frame on a text-bearing record renders as U+FFFD replacement characters instead of tearing down the whole preview. Transcript hashing covers raw bytes, so sender/receiver transcript verification is unaffected. The dead `Utf8` variants are removed from both error enums (API-breaking for the enums; workspace is `publish = false` and no external matches exist).

## Acceptance criteria → tests

| Criterion (#710) | Test |
| --- | --- |
| Single shared profile; equivalent idle/keepalive/stream bounds on both servers | `shared_transport_profiles_pin_liveness_and_stream_bounds`, `direct_server_config_sets_quic_liveness_backstop` (stream) + `broker_and_direct_transports_share_hardening_profile` (broker) |
| Long publish (>4096 records) forwarded without closing the room; limits from config | `broker_forwards_long_publish_past_receive_defaults_without_closing_room`, `broker_publish_forward_limits_come_from_config`, `broker_config_rejects_zero_resource_limits` |
| Non-UTF-8 frame skipped (lossy), not stream-fatal, on both paths | `receiver_replaces_invalid_utf8_frame_instead_of_aborting_stream`, `non_utf8_frame_decodes_lossily_per_record` (stream) + `subscriber_replaces_invalid_utf8_frame_instead_of_aborting_stream` (broker) |
| 0-RTT disabled; bounded connects; control frames ≤256 B; `accept()` bounded | `direct_server_and_client_configs_disable_early_data`, `broker_server_and_client_configs_disable_early_data`, `connect_with_timeout_bounds_dial_to_unresponsive_peer`, `receiver_times_out_when_no_sender_dials`, `receiver_disables_accept_deadline_at_zero`, `max_control_frame_len_covers_maximal_valid_envelope`, `broker_rejects_control_frames_above_control_cap` |
| Whole-frame deadline (dribble hardening) | `broker_applies_one_deadline_across_dribbled_length_prefix` |
| Subscribe to a full backlog doesn't copy under the global mutex | `broker_subscribe_snapshots_backlog_by_reference` (`Arc::ptr_eq`, deterministic) |

Also adapted: `broker_applies_quiet_gap_not_handshake_deadline_to_authenticated_publisher` (renamed; the old comment/test documented the intentionally-unbounded publish read).

## Verification

- `cargo test -p transport-quic-stream` (27) and `-p transport-quic-broker` (44 unit + binary/envelope integration) — all green
- Ripple: `cargo test -p marmot-app -p agent-stream-compose -p agent-connector` — all green; `cargo check --all-targets` on wn-cli + conformance simulator — clean
- `just fast-ci` — passes

## Docs

- `crates/cli/CHANGELOG.md` (Unreleased: Security / Fixed / Changed)
- `docs/quic-broker-deployment.md` — new operator flags
- `docs/marmot-architecture/runtime-state-bounds.md` — publish forward bounds + handshake deadline rows
- `crates/transport-quic-stream/AGENTS.md` — `src/hardening.rs` key-file entry

After merge: post the criterion→test mapping on #710 and close it.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/729">
  <picture>
    Open in Stage
  </picture>
</a>


crates/transport-quic-broker/src/handlers.rs
```diff
@@ -69,21 +94,23 @@ async fn handle_publish_stream(
     }
     let key = control.key();
     state.wait_for_subscriber(&key).await?;
-    let mut limit_state = AgentTextStreamReceiveAccumulator::default();
+    // Forward-role limits come from the broker config: the subscriber-sized
+    // receive defaults would truncate a legitimate long preview mid-stream and
+    // close the room. Each subscriber still enforces its own receive limits.
+    let mut limit_state = policy.publish_limits.accumulator();
```
erskingardner, line 108: This limit is documented and surfaced as `publish_max_plaintext_bytes`, but the broker applies it before any subscriber decrypts the record. On brokered encrypted streams `BrokerTextPublisher` calls `encrypt_record`, which replaces `record.plaintext_frame` with ciphertext (`plaintext + 16-byte AEAD tag`) before the broker sees it. `AgentTextStreamReceiveAccumulator::observe` therefore counts ciphertext bytes here, so an operator-configured 64 MiB plaintext budget can trip early by 16 bytes per forwarded record. Either this needs to be renamed/documented as a wire/ciphertext payload limit, or the limit needs enough AEAD overhead allowance to preserve the advertised plaintext budget for encrypted previews.

erskingardner (commented):
I did a full adversarial pass over the QUIC hardening changes. I found one issue around the new broker publish byte limit on encrypted streams; leaving it inline.

crates/transport-quic-broker/src/handlers.rs
erskingardner: Good catch — confirmed: the publish handler observes records pre-decrypt (the broker never decrypts), so the accumulator was counting ciphertext (plaintext + 16-byte AEAD tag per record) against a budget named "plaintext". Only the publish path had this; the subscriber observes after `decrypt_record`, so its budget genuinely counts plaintext.

Fixed in 944a991 by renaming to the honest wire semantics rather than padding the budget — the broker is policy-blind and can't know whether frames are encrypted, so an AEAD allowance would over-allow unencrypted streams:

- `publish_max_plaintext_bytes` → `publish_max_frame_bytes` (config field, `DEFAULT_BROKER_PUBLISH_MAX_FRAME_BYTES`, `--publish-max-frame-bytes`, startup output, `EmptyPublishFrameByteLimit`), documented as counting record frame bytes as carried on the wire: ciphertext for encrypted previews, plaintext for unencrypted ones.
- `PublishForwardLimits::accumulator()` now carries a comment explaining why the accumulator's `max_plaintext_bytes` field counts wire bytes at this call site.
- New test `broker_publish_frame_byte_limit_counts_wire_bytes_for_encrypted_records` pins the semantics: five 10-byte plaintext chunks encrypt to 26-byte wire frames; against a 100-byte budget the fourth record trips it (4×26 = 104 > 100), where a plaintext-counting budget (50 bytes total) would have passed all five.

Docs/CHANGELOG updated to match.

Issue #428, opened 2026-06-28
Every inbound message rewrites the sender's directory record across all caches and fires an un-coalesced full directory-subscription relay re-sync (write amplification + relay churn)


erskingardner:
## Summary
On the receive hot path, each delivered message unconditionally rewrites the sender's directory record (shared storage + every per-account cache) and queues a full directory-subscription relay re-sync — with no "is the sender already known and unchanged?" check and no debounce/coalescing.

## Location
- Call site: `crates/marmot-app/src/client/sync.rs:387` — `self.app.remember_directory_message_sender(&message)?;` per delivered message in `observe_account_device_effects`.
- `crates/marmot-app/src/directory/methods.rs:895-913` — `save_directory_entry_with_reason` **unconditionally**:
```rust
let proposed_entry = self.hydrate_directory_record(entry.clone())?;
let shared_storage = self.shared_storage()?;
let shared_entry = shared_storage.public_directory_user(&proposed_entry.account_id_hex)? ... ;
let entry = select_newer_directory_entry(...);
shared_storage.put_public_directory_user(&public_directory_user_record(&entry)?)?;  // write
for cache in self.directory_caches()? { cache.put_with_reason(&entry, reason)?; }   // write all caches
self.request_directory_sync_rebuild();                                              // full relay re-sync
```
- `crates/marmot-app/src/directory/sync.rs:120-199` — each `Rebuild` runs `run_directory_sync_once`, which rebuilds the whole directory sync plan and **re-syncs all directory user subscriptions against relays** (network I/O).

## Impact
A burst of inbound traffic (an active group, or a peer flooding messages) causes, per message: redundant shared-storage + N-cache writes plus a queued full directory subscription relay re-subscribe. The bounded channel caps queue depth, but each queued item is a full relay re-subscribe, so steady inbound load keeps the directory worker continuously re-subscribing — relay churn and wasted I/O proportional to inbound message **volume**, not to actual directory changes.

## Suggested fix
- Skip the write + rebuild when the sender record is already cached and unchanged (compare before writing).
- Coalesce/debounce rebuild requests in `DirectorySyncHandle` (collapse pending `Rebuild`s into one).

## Dedup note
[...]

PR #734, opened 2026-07-04
Validate-before-mutate discipline: close the #711 torn-write bundle

### LINKED ISSUE #333: Auto-commit stages the projected post-merge group record before `begin_pending`, leaving a torn record if an intervening step fails
## Summary
`stage_auto_commit_for_queued_proposal` persists the group record at epoch N+1 (members dropped) before the fallible `commit_priority`/`begin_pending` steps; the cleanup guard rolls back only the OpenMLS commit, not the record.

## Severity
**LOW** — bug (torn write, narrow window)

## Location
`crates/cgka-engine/src/message_processor/ingest.rs:1147-1170`

## Details
In `stage_auto_commit_for_queued_proposal`, after `commit_to_pending_proposals` stages the OpenMLS commit, the code mutates and persists the Marmot group record to the projected post-merge state BEFORE the state-machine transition runs:

```rust
let new_epoch = EpochId(pre_commit_epoch.0.saturating_add(1));
if let Ok(mut g) = self.storage.get_group(group_id) {
    g.epoch = new_epoch;
    g.members.retain(|member| !auto_removed.iter().any(|id| id == &member.id));
    self.storage.put_group(&g)?;            // <-- record advanced to N+1, members dropped
}
let commit_priority = mls_group.pending_commit().map(...).ok_or_else(|| EngineError::Backend(...))?;  // fallible
...
self.epoch_manager.begin_pending(...)?;     // fallible
```

The `PendingCommitCleanupGuard` armed earlier (line 1109-1114) only clears the OpenMLS pending commit and releases the recovery snapshot on early return — it does NOT restore the group record. So if `commit_priority` resolves to `None` (Backend error) or `begin_pending` returns `Err`, the function returns early: the OpenMLS group is rolled back to `pre_commit_epoch` by the guard, but the persisted group record stays at `new_epoch` with the auto-removed member(s) already dropped. The record's epoch then disagrees with both the MLS epoch and the epoch_manager state. Unlike the invite/remove/upgrade send paths (which run `put_group` only after staging but share the same guard limitation), this path also drops members, widening the divergence to the rendered membership list.

## Impact
A torn group record (epoch N+1 / member removed in storage while MLS + state-machine are at epoch N) on a mid-stage failure, matching the membership-vs-MLS mismatch class of mdk#559. Practical trigger probability is low because `is_stable` is checked at ingest.rs:1081 before this block (so `begin_pending` from Stable succeeds) and a pending commit was just produced (so `commit_priority` is `Some`), but the failure window is real and the guard does not cover the record write.

## Suggested fix
Defer the projected `put_group` until after `begin_pending` succeeds (as the publish-confirm path does), or extend `PendingCommitCleanupGuard` to also snapshot/restore the Marmot group record, or wrap the put_group + begin_pending sequence so a failure rolls the record back to its pre-stage projection.

## Dedup note
Same membership-vs-MLS mismatch class as marmot-protocol/mdk#468, but a distinct code path and a narrow failure window.

---

### LINKED ISSUE #337: `sync_account_groups` mutates relay subscriptions before updating routing state, with no rollback on partial failure
## Summary
subscribe/unsubscribe are applied to relays before `AdapterState` is updated; a mid-loop unsubscribe failure returns Err with relays half-reconciled while routing state still reflects the old group set, so the routing table and relay subscriptions can diverge until a later fully-successful sync.

## Severity
**LOW** — concurrency/correctness

## Location
`crates/transport-nostr-adapter/src/lib.rs:525-541 (sync_account_groups)`

## Details
sync_account_groups computes (to_add, to_remove), then `subscribe_all(to_add)` and a serial loop of `unsubscribe(to_remove)`, and only afterward (line 538-540) updates AdapterState via `sync_groups`. If `subscribe_all` succeeds but an `unsubscribe` in the loop fails, the function returns Err before `sync_groups` runs: the new subscriptions are already registered on relays, some stale subscriptions are removed and some are not, yet AdapterState still reflects the OLD group set. A subsequent retry re-diffs against the stale state and re-issues already-registered adds while the half-removed set stays inconsistent. The routing table (which gates inbound delivery) and the relay-side subscription set can therefore diverge, and the metrics counters (subscriptions_created/removed) never get updated for the work that did land.

## Impact
After a transient relay error mid-sync, the adapter can hold a routing table that disagrees with the relays' actual subscriptions: inbound group events may be dropped (no matching route) or stale group subscriptions keep delivering. Self-heals only on a later fully-successful sync. Limited blast radius (requires a relay-side unsubscribe failure) hence LOW.

## Suggested fix
Update AdapterState to the desired group set even on partial relay failure (the routing table should reflect intent), or apply the diff transactionally: record state first, then reconcile relays, and on relay error re-enqueue the failed unsubscribe rather than abandoning the state update.

---

### LINKED ISSUE #338: Read-marker leak on failed send: the local projection is invalidated but the read marker it advanced pre-publish is left pointing at the invalidated message
## Summary
On an outbound chat send the read marker is advanced pre-publish; on a hard publish failure the timeline row is invalidated but the marker is not rolled back, which can transiently mask an inbound message that arrived in the window or desync the unread badge until the next read.

## Severity
**LOW** — correctness (self-healing)

## Location
`crates/marmot-app/src/client/mod.rs:1022-1098 (send_app_event_with_local_projection); record_local_app_event_projection in client/projection.rs:20-61`

## Details
On an outbound chat send, record_local_app_event_projection is called BEFORE publish (source_message_id_hex=None). For MARMOT_APP_EVENT_KIND_CHAT it both records the app event AND calls mark_timeline_message_read(group, event.id), advancing the per-group read marker to this message id. If the subsequent sync/publish fails and pending state did not confirm (a plain AppMessage send has no pending resolution, so fail_if_publish_failed returns a hard error), the code invalidates the timeline app event via invalidate_timeline_app_event("local_publish_failed") but does not roll the read marker back. The marker was moved to a message that is now invalidated/absent.

## Impact
After a failed send, the group's read marker can point at the just-invalidated outgoing message. Depending on how the chat-list unread aggregate compares the marker against surviving rows, this can mask a genuinely-unread inbound message that arrived between the read-marker advance and the failure, or otherwise desync the unread badge until the next successful read. The window is small and self-heals on the next successful mark-read, so user impact is minor, but it is a real read-state inconsistency on the failure path that the explicit projection-retract was meant to undo cleanly.

## Suggested fix
On the publish-failure branch, after invalidate_timeline_app_event, recompute/re-clamp the read marker for the group (e.g. re-derive it from the surviving timeline tail) rather than leaving it advanced to the invalidated id; or defer the mark_timeline_message_read of the local projection until after publish succeeds (the post-publish second record_local_app_event_projection already runs the read marker).

## Dedup note
Related to the unread-count area of marmot-protocol/mdk#529 but a distinct failed-send code path.

---

### LINKED ISSUE #352: Welcome-delivery failure after a confirmed commit is downgraded to a flat string failure with no re-delivery handle
Found via review of the account group-evolution publish path.

## What

The anti-rollback logic correctly keeps a commit once any endpoint accepts it, but subsequent welcome publishes only record into `output.failures` as a string — no pending/retry context. An added member can therefore be left unjoinable with no structured signal to re-deliver just the welcome.

## Where

- `crates/marmot-account/src/runtime.rs:503-535` (`publish_group_evolution`), `:466-501` (`publish_group_created`).

## Why it matters

Correctness/operability (same spirit as marmot-protocol/mdk#469/#217): a confirmed membership change with an undelivered welcome has no first-class re-delivery path.

## Fix

Surface per-welcome failures with recipient/pending context so the runtime can re-deliver the welcome without re-committing.

---

### LINKED ISSUE #358: `JsonlRecorder::set_data_mode` advances `data_mode` before the file swap, with no rollback on swap failure (audit record loss + broken one-mode-per-file invariant)
Found via review of the forensic audit recorder rotation path.

## What

`set_data_mode` assigns `inner.data_mode = new_mode` **before** calling `swap_to_fresh_file?`. `swap_to_fresh_file` unlinks the old file before opening the new one. If the `open` fails, the function returns the error with `data_mode` already changed, the old file already deleted, the stale unlinked fd still in `inner.writer`, and no `AuditDataModeChanged` row written.

## Where

- `crates/marmot-forensics/src/audit.rs:1296-1297` (mode assigned), `:1322-1348` (`swap_to_fresh_file`: unlink at `:1331`, open at `:1339`).

## Why it matters

After a mid-rotation IO error the recorder silently stamps `new_mode` and keeps writing to a dead inode — subsequent audit records are lost, and the "one unambiguous data mode per file" invariant is broken. Edge case (requires an IO failure during rotation), hence LOW, but it is a silent forensic-integrity failure.

## Suggested fix

Swap the file first and only assign `inner.data_mode = new_mode` after the new file opens successfully; or open-before-unlink so a failed open leaves the original file/fd intact.

---

### LINKED ISSUE #366: agent-connector: `StreamFinalize` validation rejection destroys the session and compose task *before* validating — finalize becomes unretryable and the durable final is never sent
## Summary
`stream_finalize_response` removes the session from the store and tells the compose task to `Finish` (making it exit) **before** validating the client-supplied `final_text` / `transcript_hash_hex` / `chunk_count`. On any mismatch it returns an error with the session already gone, so the canonical durable `finish_agent_text_stream` is never called and the agent cannot re-issue `StreamFinalize`.

## Where — `crates/agent-connector/src/stream.rs:185-251`
```rust
let session = self.streams.remove(&stream_id_hex)?;          // (1) session removed
session.tx.send(StreamComposeCommand::Finish { respond })...; // (2) compose task exits
let report = response.await...?;
if report.text != final_text { return Err(...); }            // (3) too late: state already gone
if actual_transcript_hash_hex != Some(expected) { return Err(...); }
if report.chunk_count != chunk_count { return Err(...); }
let (_payload, summary) = self.runtime.finish_agent_text_stream(...).await?;  // never reached on mismatch
```

## Impact
A transient client/connector accounting disagreement (off-by-one chunk count, hash/text mismatch after a dropped live append) **permanently strands** the stream: the live preview was published but the durable MLS final message is never sent, and there is no recovery path short of starting a new stream. A misbehaving/prompt-injected gateway can deterministically trigger this to leave previews with no durable final.

## Suggested fix
Validate the report **before** removing the session / sending `Finish`; only finalize-and-remove after validation passes, keeping the session alive (and retryable) on mismatch.

## Dedup
- **#519** (closed): Hermes *client* drops the durable reply when the connector rejects finalize — client-side behavior on rejection.
- **#215** (closed): a stalled broker blocks finalize.
Both differ: this is a connector-side ordering bug that makes the rejection itself irrecoverable by destroying server state before the validation that produced it.

---

---

### LINKED ISSUE #375: marmot-account: `required_acks == 0` ('best effort') confirms a commit/welcome with zero accepted endpoints and reports no failure
## Summary
`TransportPublishReport::met_required_acks()` is `accepted_count() >= required_acks`, so with the trait-documented `required_acks == 0` ("best effort") it returns `true` even when **every** endpoint failed and none accepted. The account runtime then `confirm_published`s the commit/welcome and pushes no `PublishFailure`.

## Where
- `crates/traits/src/transport_adapter.rs:192-194` (`met_required_acks`) and `:126-127` (the `0` = best-effort contract).
- Consumed in `crates/marmot-account/src/runtime.rs` (`publish_one` ~`:637`, `publish_pending` ~`:665`, `publish_group_created` ~`:450/487`, `publish_group_evolution` ~`:513`).
```rust
pub fn met_required_acks(&self) -> bool { self.accepted_count() >= self.required_acks }  // 0 >= 0 == true
```

## Impact
A misconfigured/degraded routing policy that yields `required_acks == 0` produces a **confirmed-but-never-published** commit/welcome — the local device advances its epoch / confirms a membership change that reached no relay, with no failure surfaced. This is the inverse of the marmot-protocol/mdk#509/#499 class (which fixed *accepted-but-treated-as-failed*). The shipping default is `1` (`AppRoutingState`), and `required_acks` is only ever a stored config constant (never derived to 0), so this is a **latent footgun / missing guard**, not a live default-config bug.

## Suggested fix
Treat zero accepted endpoints as a failure regardless of `required_acks` (or require `met_required_acks` to also satisfy `accepted_count() >= 1` when any endpoints were attempted), and surface a `PublishFailure`.

## Dedup
- **#483 / marmot-protocol/mdk#503** (closed): rollback of *relay-accepted* messages — opposite direction. No issue covers the zero-ack confirm path.

---

---

### LINKED ISSUE #711: [Tracking] Atomic, compensating, and restart-reconcilable multi-step state changes
## Scope

Reopened after the 2026-07-21 audit. The existing multi-step-state contract remains correct; new paths have bypassed it across engine, projection, adapter, connector, account-home, and push orchestration.

## Program checklist

- [ ] #819 — marmot-account: remove_account propagates post-commit errors (?) — aborts the wipe after the account is already gone, skipping tombstone secret-scrub and cleanup
- [x] #820 — cli: finish_stream_compose validation guards orphan a dead compose session — the transcript becomes un-finishable and un-cleanable, defeating the #366 retry guarantee
- [x] #822 — marmot-account: local-signing import writes the keychain secret before the account record — a crash in between orphans a keychain entry that blocks re-import with no API recovery
- [ ] #823 — agent-connector: stream_begin_response publishes the durable StreamStart before crypto/session setup — a failure after publish orphans the start with no final and no compensation
- [x] #824 — cgka-engine: do_send_update_group_data mirrors the projected group record BEFORE begin_pending, deviating from the #333 project-after-staging atomicity rule (latent torn record)
- [ ] #847 — cli daemon: open_stream_compose publishes the durable stream-start marker before deriving stream crypto — a crypto failure orphans the start with no compose session and no compensation
- [x] #869 — cli/daemon: run_server leaks the bound Unix socket and pid file when post-bind relay resolution fails (uncompensated startup side effects)
- [ ] #907 — agent-connector: a relay publish failure during `start()` aborts the whole daemon before it serves any connection (and leaks the bound socket)
- [ ] #908 — agent-connector: create_account_response orphans a durably-created account when the follow-on key-package publish fails; non-idempotent retries mint fresh orphans
- [ ] #909 — marmot-app: LeaveGroup / DeclineGroupInvite worker handlers never flush the own-group projection updates they queue — live timeline stays stale and the update is later emitted mis-timed
- [ ] #910 — transport-nostr-adapter: sync_account_groups issues relay REQs for newly-added groups BEFORE registering their routes, so stored catch-up history can be dropped as unroutable
- [x] #921 — cgka-engine: do_send_update_group_data sources the carried-forward profile field via `.ok()...unwrap_or_default()`, blanking the unchanged field on a swallowed storage read error
- [x] #929 — Push registration gossip failures are reported as success and have no durable per-group retry
- [ ] #931 — marmot-app: three account-worker handlers strand queued own-group kind-1210 projection updates — mis-timed live emission (sibling of #909)
- [ ] #943 — marmot-app: directory subscription sync is non-atomic with the SDK — a partial subscribe_with_id_to failure leaves subscriptions live but their validation filter never committed, silently dropping their events
- [x] #966 — Security review H3: inbound delivery is engine-durable and dedup-marked before app projection — publish/storage error in between silently drops the message
- [x] #974 — Security review M5: queued outbound intent is deleted before the publish it gates is confirmed
- [x] #978 — Security review M9: post-mutation catch-up failure misreports a committed, externally visible group operation as an FFI error
- [x] #984 — Security review M15: create_group tears between engine-confirmed group and projection roster — engine-only ghost groups
- [x] #1003 — Security review L9: adapter reports Ok with zero acceptances when required_acks == 0
- [x] #1005 — Security review L11: activate_account leaves half-active state when subscription issuance fails
- [x] #1020 — Security review L26: publish-before-apply is only tested for total publish failure — no partial-accept race scenarios

## Post-audit additions (2026-07-23)

- [ ] #1036 — marmot-app: a mid-batch ingest error in sync_sdk_relay discards the whole batch SyncSummary — already-projected earlier deliveries are never emitted as live events
- [x] #1039 — agent-connector: send_final / stream_finalize idempotency has an in-flight TOCTOU — two concurrent same-key+same-body requests both send, double-posting an unrecallable message
- [x] #1088 — closed as an accepted upstream-blocked limitation; Hermes still does not provide stable cross-attempt delivery identity

## Contract

- Validate before mutation where possible.
- Otherwise persist intent and make retry idempotent, or compensate every applied step.
- Never turn a committed/exposed mutation into an API error that invites duplicate retry.
- Reconciliation must compare actual durable/external state, not merely replay the last call.

## Acceptance criteria

- Engine-durable but projection-missing deliveries are repaired automatically.
- Account/group/stream creation and removal have tested failure injection after every step.
- Subscription/activation partial failures either roll back or remain safely retryable.
- Publish outcomes require real exposure and preserve partial-success truth.

## Operating rule

erskingardner:
## Summary

Fixes all seven open child issues of the #711 tracking bundle — multi-step operations that advanced durable or observable state before the fallible step that justified it, with no compensation on failure — and adds the workspace convention doc #711 asks for. One commit per issue.

| Commit | Issue | Fix |
| --- | --- | --- |
| `f6792b6` | #375 | `met_required_acks` requires at least one accepted endpoint; `required_acks == 0` relaxes the threshold, never the existence requirement. Zero-ack publishes now surface a `PublishFailure` and roll back unexposed work. |
| `0267d1e` | #352 | Post-confirm welcome misses surface as structured `WelcomeDeliveryFailure { message_id, recipient, group_id, reason }`; new `AccountDeviceRuntime::redeliver_welcome` re-publishes the stored wrapped welcome without re-committing. |
| `6e5a418` | #338 | The chat read marker advances only in the post-publish success projection; a failed send can no longer leave it pointing at an invalidated own message or silently clear inbound unreads. Safe with no compensation needed: the unread aggregate excludes the sender's own messages. |
| `c93242f` | #333 | Auto-commit staging projects the group record only after `commit_priority`/`begin_pending`; a failed `put_group` is compensated in-memory via `rollback_publish`. `PendingCommitCleanupGuard` also removes the stored SelfRemove proposal (left behind, OpenMLS 0.8.1 panics when a later `remove_members` filters it). |
| `081b6ab` | #337 | `sync_account_groups` records intent first (routing table updated before relay teardown); failed unsubscribes queue in `AdapterState` and retry on the next sync, filtered against live routes so a re-added group's identical subscription id is never torn down. New `unsubscribe_retries_pending` gauge. |
| `3bc00fd` | #358 | Audit-file swap is stage-then-rename (atomic replace, 0600 preserved); the data mode is assigned only after the swap succeeds. A mid-rotation failure leaves the original file, fd, seq, session id, and mode untouched. |
| `f5adbe3` | #366 | Stream finalize validation moved inside the compose task's `Finish` handling (atomic with teardown, no snapshot-then-finish race): a mismatch responds `Err` with zero side effects, the session stays alive, and finalize is retryable. |
| `ac8e7f3` | #711 | New `docs/marmot-architecture/overview/multi-step-state-changes.md` + repo-invariant bullet and index entries. |

## Acceptance criteria → regression tests

Every #711 acceptance criterion is pinned by a failure-injection test:

- #366 → `finish_with_mismatched_expectation_keeps_session_alive_and_retryable` (agent-stream-compose), `connector_socket_finalize_mismatch_keeps_stream_session_retryable` (agent-connector, socket-level)
- #333 → `auto_commit_record_write_failure_leaves_no_torn_group_record` (new `tests/auto_commit_atomicity.rs`, `FaultStorage` one-shot `put_group` fault)
- #358 → `set_data_mode_failure_leaves_mode_file_and_recording_intact`, `rotate_failure_keeps_recording_to_original_file`, `set_data_mode_recovers_from_stale_staged_swap_file`
- #337 → `sync_with_failed_unsubscribe_returns_ok_and_routes_reflect_intent`, `failed_unsubscribe_is_retried_and_drained_on_next_sync`, `pending_unsubscribe_for_readded_group_is_discarded_not_replayed`, `deactivate_account_clears_pending_unsubscribes`
- #338 → `failed_send_keeps_read_marker_and_inbound_unread` (relay shutdown mid-test)
- #375 → unit tests in `transport_adapter.rs` + `create_group_with_best_effort_acks_rolls_back_when_nothing_accepted`
- #352 → extended partial-exposure/welcome-failure runtime tests incl. a full `redeliver_welcome` round-trip and a commit-ack-miss misclassification guard

## Verification

- `just fast-ci` clean (fmt, check, clippy incl. OTLP builds)
- Targeted suites all green: cgka-traits, cgka-engine (incl. full integration battery), cgka-session, marmot-account, marmot-app (305 lib + 65 relay_runtime), transport-nostr-adapter (incl. `--features sdk`), marmot-forensics, agent-stream-compose, agent-connector, wn-cli, cgka-conformance-simulator (incl. the repo-wide tracing audit)

## Follow-ups noted, out of scope

- `do_send_invite` / `do_send_remove_members` / `do_send_update_app_components` share #333's put-group-before-begin_pending shape; their windows are equally unreachable in-process today, and the same mechanical reorder can follow under the new convention.
- `crates/cli/src/daemon/stream_workers.rs` `finish_stream_compose` (~line 592) comments claim a retryability the compose task doesn't provide (it exits after answering `Finish`) — a #366-class claim/behavior mismatch in the CLI daemon worth its own issue.
- If `finish_agent_text_stream` fails *after* a validated compose finish, the durable final is still unretryable (pre-existing; needs an idempotent runtime-side finish).

Closes #366
Closes #333
Closes #358
Closes #337
Closes #338
Closes #375
Closes #352
Closes #711

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/734">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added pending welcome delivery listing and welcome redelivery APIs across the app and runtime.
  * Improved stream finalization workflow to support safe retries after composition completes.
* **Bug Fixes**
  * Preserved local read/unread state when a send fails.
  * Stream finalization mismatches and marker publish failures no longer tear down sessions, allowing subsequent retries to succeed.
  * Improved robustness of audit log rotation and relay-side unsubscribe retries.
* **Documentation**
  * Added multi-step “no-torn-writes” guidance for safer state updates.

### DIFF
--- a/crates/agent-connector/src/stream.rs
+++ b/crates/agent-connector/src/stream.rs
@@ -17,7 +17,7 @@ use crate::quic::{
     broker_trust_for_candidate, first_quic_candidate, parse_quic_candidate,
     resolve_quic_candidate_addr,
 };
-use crate::stream_session::ActiveStreamSession;
+use crate::stream_session::{ActiveStreamSession, FinalizedStream};
 use crate::validation::{normalize_hex, transcript_hash_from_hex, unix_now_seconds};
 use crate::{
     AgentConnector, STREAM_COMPOSE_CHANNEL_DEPTH, STREAM_COMPOSE_CHUNK_BYTES,
@@ -120,6 +120,7 @@ impl AgentConnector {
                 cancel_tx,
                 abort: handle.abort_handle(),
                 last_activity: Instant::now(),
+                finalized: None,
             },
         );
         Ok(AgentControlResponse::StreamBegun {
@@ -204,6 +205,32 @@ impl AgentConnector {
         // its teardown. On mismatch the session stays registered and the
         // compose task keeps running, so finalize is retryable (#366).
         let session = self.streams.get(&stream_id_hex)?;
+
+        // Retry fast-path: a prior finalize already validated the transcript and
+        // the compose task exited, but the durable finish below failed. The
+        // compose task is gone, so re-attempt the durable finish directly from
+        // the retained transcript, rejecting a retry that disagrees with what
+        // was frozen (a finalize cannot change after the transcript is sealed).
+        if let Some(finalized) = &session.finalized {
+            if finalized.final_text != final_text
+                || finalized.transcript_hash != transcript_hash
+                || finalized.chunk_count != chunk_count
+            {
+                return Err(ConnectorError::Stream(
+                    "stream finalize does not match the already-finalized transcript".into(),
+                ));
+            }
+            return self
+                .finish_finalized_stream(
+                    &stream_id_hex,
+                    &session,
+                    final_text,
+                    transcript_hash,
+                    chunk_count,
+                )
+                .await;
+        }
+
         let (respond, response) = oneshot::channel();
         if session
             .tx
@@ -241,18 +268,49 @@ impl AgentConnector {
         // running and the session stays registered, so do nothing else here
         // and the agent can append again and/or retry the finalize.
         report.map_err(ConnectorError::Stream)?;
-        let _ = self.streams.remove_if_same(&stream_id_hex, &session);
-        // If the runtime finish below fails, the compose task has already
-        // exited and the durable final remains unretryable — pre-existing
-        // gap, out of scope for #366.
+        // Validation passed and the compose task has now exited. Freeze the
+        // transcript on the still-registered session so a durable-finish
+        // failure below leaves a retryable finalize (the compose task can no
+        // longer be consulted).
+        self.streams.mark_finalized(
+            &stream_id_hex,
+            &session,
+            FinalizedStream {
+                final_text: final_text.clone(),
+                transcript_hash,
+                chunk_count,
+            },
+        );
+        self.finish_finalized_stream(
+            &stream_id_hex,
+            &session,
+            final_text,
+            transcript_hash,
+            chunk_count,
+        )
+        .await
+    }
+
+    /// Publish the durable stream final for an already-validated finalize, and
+    /// remove the session only once that publish succeeds. On failure the
+    /// session stays registered with its frozen transcript so a re-issued
+    /// `StreamFinalize` retries this step (#366).
+    async fn finish_finalized_stream(
+        &self,
+        stream_id_hex: &str,
+        session: &ActiveStreamSession,
+        final_text: String,
+        transcript_hash: [u8; 32],
+        chunk_count: u64,
+    ) -> Result<AgentControlResponse, ConnectorError> {
         let (_payload, summary) = self
             .runtime
             .finish_agent_text_stream(
                 &session.account_label,
                 &session.group_id,
                 AgentTextStreamFinishRequest {
-                    stream_id: session.stream_id,
-                    start_event_id: session.start_message_id_hex,
+                    stream_id: session.stream_id.clone(),
+                    start_event_id: session.start_message_id_hex.clone(),
                     final_text_or_reference: final_text,
                     transcript_hash,
                     chunk_count,
@@ -260,8 +318,10 @@ impl AgentConnector {
                 },
             )
             .await?;
+        // Durable final published: it is now safe to drop the session.
+        let _ = self.streams.remove_if_same(stream_id_hex, session);
         Ok(AgentControlResponse::StreamFinalized {
-            stream_id_hex,
+            stream_id_hex: stream_id_hex.to_owned(),
             message_ids_hex: summary.message_ids,
         })
     }
--- a/crates/agent-connector/src/stream.rs
+++ b/crates/agent-connector/src/stream.rs
@@ -3,7 +3,9 @@
 use std::time::Instant;
 
 use agent_control::AgentControlResponse;
-use agent_stream_compose::{StreamComposeCommand, StreamComposeReport, run_stream_compose_session};
+use agent_stream_compose::{
+    StreamComposeCommand, StreamComposeReport, StreamFinishExpectation, run_stream_compose_session,
+};
 use cgka_traits::{GroupId, MessageId};
 use marmot_app::AgentTextStreamFinishRequest;
 use tokio::sync::mpsc::error::TrySendError;
@@ -196,45 +198,53 @@ impl AgentConnector {
         chunk_count: u64,
     ) -> Result<AgentControlResponse, ConnectorError> {
         let stream_id_hex = normalize_hex(stream_id_hex)?;
-        let session = self.streams.remove(&stream_id_hex)?;
+        let transcript_hash = transcript_hash_from_hex(transcript_hash_hex)?;
+        // Clone (not remove) the session: the final text/hash/chunk-count
+        // expectation is validated inside the compose task, atomically with
+        // its teardown. On mismatch the session stays registered and the
+        // compose task keeps running, so finalize is retryable (#366).
+        let session = self.streams.get(&stream_id_hex)?;
         let (respond, response) = oneshot::channel();
         if session
             .tx
-            .send(StreamComposeCommand::Finish { respond })
+            .send(StreamComposeCommand::Finish {
+                expected: Some(StreamFinishExpectation {
+                    final_text: final_text.clone(),
+                    transcript_hash_hex: hex::encode(transcript_hash),
+                    chunk_count,
+                }),
+                respond,
+            })
             .await
             .is_err()
         {
+            // The compose task is gone: drop the dead registration (only if
+            // it is still this session) and reclaim the task, preserving the
+            // previous remove-then-abort cleanup semantics.
+            let _ = self.streams.remove_if_same(&stream_id_hex, &session);
             session.abort.abort();
             return Err(ConnectorError::Stream(
                 "stream compose session is closed".into(),
             ));
         }
-        let report = response
-            .await
-            .map_err(|err| ConnectorError::Stream(err.to_string()))?
-            .map_err(ConnectorError::Stream)?;
-        if report.text != final_text {
-            return Err(ConnectorError::Stream(
-                "stream final text does not match appended transcript".into(),
-            ));
-        }
-        let transcript_hash = transcript_hash_from_hex(transcript_hash_hex)?;
-        let expected_transcript_hash_hex = hex::encode(transcript_hash);
-        let actual_transcript_hash_hex = report
-            .transcript_hash
-            .as_deref()
-            .map(normalize_hex)
-            .transpose()?;
-        if actual_transcript_hash_hex.as_deref() != Some(expected_transcript_hash_hex.as_str()) {
-            return Err(ConnectorError::Stream(
-                "stream final transcript hash does not match appended transcript".into(),
-            ));
-        }
-        if report.chunk_count != chunk_count {
-            return Err(ConnectorError::Stream(
-                "stream final chunk count does not match appended transcript".into(),
-            ));
-        }
+        let report = match response.await {
+            Ok(report) => report,
+            Err(err) => {
+                // The compose task died mid-command: same cleanup as the
+                // send failure above.
+                let _ = self.streams.remove_if_same(&stream_id_hex, &session);
+                session.abort.abort();
+                return Err(ConnectorError::Stream(err.to_string()));
+            }
+        };
+        // An `Err` report is a validation mismatch: the compose task is still
+        // running and the session stays registered, so do nothing else here
+        // and the agent can append again and/or retry the finalize.
+        report.map_err(ConnectorError::Stream)?;
+        let _ = self.streams.remove_if_same(&stream_id_hex, &session);
+        // If the runtime finish below fails, the compose task has already
+        // exited and the durable final remains unretryable — pre-existing
+        // gap, out of scope for #366.
         let (_payload, summary) = self
             .runtime
             .finish_agent_text_stream(
--- a/crates/agent-connector/src/stream.rs
+++ b/crates/agent-connector/src/stream.rs
@@ -271,8 +271,12 @@ impl AgentConnector {
         // Validation passed and the compose task has now exited. Freeze the
         // transcript on the still-registered session so a durable-finish
         // failure below leaves a retryable finalize (the compose task can no
-        // longer be consulted).
-        self.streams.mark_finalized(
+        // longer be consulted). If the freeze does not land — the session was
+        // superseded by a same-stream-id replacement between our `get` and now
+        // — this clone is stale: proceeding would publish under a lost retry
+        // handle, so bail and let the agent re-finalize against the live
+        // session instead.
+        if !self.streams.mark_finalized(
             &stream_id_hex,
             &session,
             FinalizedStream {
@@ -280,7 +284,11 @@ impl AgentConnector {
                 transcript_hash,
                 chunk_count,
             },
-        );
+        ) {
+            return Err(ConnectorError::Stream(
+                "stream session was superseded during finalize".into(),
+            ));
+        }
         self.finish_finalized_stream(
             &stream_id_hex,
             &session,
--- a/crates/agent-connector/src/stream_session.rs
+++ b/crates/agent-connector/src/stream_session.rs
@@ -322,6 +322,23 @@ pub(crate) struct ActiveStreamSession {
     pub(crate) cancel_tx: mpsc::Sender<()>,
     pub(crate) abort: tokio::task::AbortHandle,
     pub(crate) last_activity: Instant,
+    /// Set once the compose task has validated the finalize expectation and
+    /// exited. The compose task is then gone, but the durable
+    /// `finish_agent_text_stream` publish that follows can still fail; the
+    /// retained transcript lets a re-issued `StreamFinalize` retry that publish
+    /// without the (dead) compose task, so a transient error never strands the
+    /// stream with a live preview and no durable final (#366).
+    pub(crate) finalized: Option<FinalizedStream>,
+}
+
+/// Frozen finalize inputs retained after the compose task exits, so the durable
+/// finish step is retryable and a retry that disagrees with the frozen
+/// transcript is rejected.
+#[derive(Clone)]
+pub(crate) struct FinalizedStream {
+    pub(crate) final_text: String,
+    pub(crate) transcript_hash: [u8; 32],
+    pub(crate) chunk_count: u64,
 }
 
 impl StreamSessionStore {
@@ -371,6 +388,25 @@ impl StreamSessionStore {
         match sessions.get(stream_id_hex) {
             Some(entry) if entry.tx.same_channel(&session.tx) => sessions.remove(stream_id_hex),
             _ => None,
+        }
+    }
+
+    /// Freeze the finalize inputs on the stored entry (only when it is still the
+    /// same session) so a durable-finish failure can be retried without the
+    /// compose task, which has exited by this point. Also refreshes activity so
+    /// the idle sweep does not reclaim a session mid-finalize.
+    pub(crate) fn mark_finalized(
+        &self,
+        stream_id_hex: &str,
+        session: &ActiveStreamSession,
+        finalized: FinalizedStream,
+    ) {
+        let mut sessions = self.sessions.lock().expect("stream session lock poisoned");
+        if let Some(entry) = sessions.get_mut(stream_id_hex)
+            && entry.tx.same_channel(&session.tx)
+        {
+            entry.finalized = Some(finalized);
+            entry.last_activity = Instant::now();
         }
     }
 
--- a/crates/agent-connector/src/stream_session.rs
+++ b/crates/agent-connector/src/stream_session.rs
@@ -355,6 +355,25 @@ impl StreamSessionStore {
         Ok(session.clone())
     }
 
+    /// Remove the entry for `stream_id_hex` only when it is still the same
+    /// session as `session` (identified by command-channel identity).
+    ///
+    /// With a get-first finalize flow, an unconditional removal could tear
+    /// down a same-stream-id replacement session inserted concurrently
+    /// between the `get` and the removal; the channel-identity check makes
+    /// the removal a no-op in that case.
+    pub(crate) fn remove_if_same(
+        &self,
+        stream_id_hex: &str,
+        session: &ActiveStreamSession,
+    ) -> Option<ActiveStreamSession> {
+        let mut sessions = self.sessions.lock().expect("stream session lock poisoned");
+        match sessions.get(stream_id_hex) {
+            Some(entry) if entry.tx.same_channel(&session.tx) => sessions.remove(stream_id_hex),
+            _ => None,
+        }
+    }
+
     pub(crate) fn remove(
         &self,
         stream_id_hex: &str,
--- a/crates/agent-connector/src/stream_session.rs
+++ b/crates/agent-connector/src/stream_session.rs
@@ -391,22 +391,30 @@ impl StreamSessionStore {
         }
     }
 
-    /// Freeze the finalize inputs on the stored entry (only when it is still the
-    /// same session) so a durable-finish failure can be retried without the
-    /// compose task, which has exited by this point. Also refreshes activity so
-    /// the idle sweep does not reclaim a session mid-finalize.
+    /// Freeze the finalize inputs on the stored entry so a durable-finish failure
+    /// can be retried without the compose task, which has exited by this point.
+    ///
+    /// Returns `true` only when the freeze landed on the still-current session
+    /// (matched by command-channel identity). It returns `false` when the entry
+    /// was removed or replaced by a same-stream-id session between the caller's
+    /// `get` and this call: in that case the caller's session is stale and must
+    /// NOT proceed to the durable publish, or a failure there would leave no
+    /// retry handle (`finalized` unset) and strand the stream (#366).
+    #[must_use]
     pub(crate) fn mark_finalized(
         &self,
         stream_id_hex: &str,
         session: &ActiveStreamSession,
         finalized: FinalizedStream,
-    ) {
+    ) -> bool {
         let mut sessions = self.sessions.lock().expect("stream session lock poisoned");
-        if let Some(entry) = sessions.get_mut(stream_id_hex)
-            && entry.tx.same_channel(&session.tx)
-        {
-            entry.finalized = Some(finalized);
-            entry.last_activity = Instant::now();
+        match sessions.get_mut(stream_id_hex) {
+            Some(entry) if entry.tx.same_channel(&session.tx) => {
+                entry.finalized = Some(finalized);
+                entry.last_activity = Instant::now();
+                true
+            }
+            _ => false,
         }
     }
 
@@ -431,12 +439,21 @@ impl StreamSessionStore {
     /// session otherwise keeps the compose task, its `mpsc::Sender`, the accumulated
     /// transcript, and (when broker connect succeeded) a dedicated quinn `Endpoint`
     /// UDP socket plus a live keep-alive'd QUIC connection alive forever.
+    ///
+    /// A session whose transcript has been finalized is NEVER swept: its compose
+    /// task has already exited and the frozen transcript is the only handle for
+    /// retrying a durable finish that failed. Sweeping it would recreate the
+    /// exact #366 failure mode (a published live preview with no durable final
+    /// and no way to retry), so a finalized session lives until its durable
+    /// finish succeeds or the connector restarts.
     pub(crate) fn sweep_idle(&self, max_idle: Duration) -> usize {
         let now = Instant::now();
         let mut sessions = self.sessions.lock().expect("stream session lock poisoned");
         let stale: Vec<String> = sessions
             .iter()
-            .filter(|(_, session)| now.duration_since(session.last_activity) >= max_idle)
+            .filter(|(_, session)| {
+                session.finalized.is_none() && now.duration_since(session.last_activity) >= max_idle
+            })
             .map(|(stream_id_hex, _)| stream_id_hex.clone())
             .collect();
         for stream_id_hex in &stale {
--- a/crates/cgka-engine/src/message_processor/ingest.rs
+++ b/crates/cgka-engine/src/message_processor/ingest.rs
@@ -1332,6 +1332,7 @@ impl<S: StorageProvider> Engine<S> {
         }
 
         let provider = EngineOpenMlsProvider::<S>::new(&self.crypto, self.storage.mls_storage());
+        let stored_proposal_ref = queued.proposal_reference_ref().clone();
         mls_group
             .store_pending_proposal(
                 <EngineOpenMlsProvider<'_, S> as openmls_traits::OpenMlsProvider>::storage(
@@ -1344,6 +1345,7 @@ impl<S: StorageProvider> Engine<S> {
         let pre_commit_epoch = EpochId(mls_group.epoch().as_u64());
         let mut pending_commit_guard =
             PendingCommitCleanupGuard::arm(&self.storage, &provider, group_id.clone());
+        pending_commit_guard.set_stored_proposal_ref(stored_proposal_ref);
         let recovery_snapshot =
             self.fork_recovery
                 .create_snapshot(&self.storage, group_id, pre_commit_epoch)?;
@@ -1381,13 +1383,6 @@ impl<S: StorageProvider> Engine<S> {
         )?;
 
         let new_epoch = EpochId(pre_commit_epoch.0.saturating_add(1));
-        if let Ok(mut g) = self.storage.get_group(group_id) {
-            g.epoch = new_epoch;
-            g.members
-                .retain(|member| !auto_removed.iter().any(|id| id == &member.id));
-            self.storage.put_group(&g)?;
-        }
-
         let commit_priority = mls_group
             .pending_commit()
             .map(crate::app_components::commit_ordering_priority_for_staged)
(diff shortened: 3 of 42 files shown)


crates/marmot-account/src/runtime.rs
```diff
@@ -722,6 +828,11 @@ pub struct AccountDeviceEffects {
     pub pending_convergence: Vec<GroupId>,
     pub reports: Vec<TransportPublishReport>,
     pub failures: Vec<PublishFailure>,
+    /// Welcomes whose publish failed after their commit/create was already
+    /// confirmed. Unlike `failures`, each entry carries the recipient and
+    /// group so the caller can re-deliver the stored welcome via
+    /// [`AccountDeviceRuntime::redeliver_welcome`] without re-committing.
+    pub welcome_failures: Vec<WelcomeDeliveryFailure>,
```
erskingardner, line 835: This adds a first-class repair handle, but every product-facing caller currently drops it. `AppClient::create_group` / invite flows call `fail_if_publish_failed(&effects)?` and then collapse the result to `GroupId`/`SendSummary`; `send_summary_from_effects`, runtime worker responses, CLI, and UniFFI do not carry `welcome_failures`, and there is no app/runtime command that calls `redeliver_welcome`. So a confirmed-but-partial welcome publish still returns success to the app while losing the only message id needed to repair the unjoinable member. Please either persist/queue these failures or thread them through the app/runtime surface together with a redelivery entry point.

crates/cli/src/daemon/stream_workers.rs
```diff
@@ -601,7 +601,10 @@ pub(crate) async fn finish_stream_compose(
     };
     let (respond, response) = oneshot::channel();
     if tx
-        .send(StreamComposeCommand::Finish { respond })
+        .send(StreamComposeCommand::Finish {
```
erskingardner, line 638: This still consumes the compose task before the durable finish-marker publish. `Finish` makes `run_stream_compose_session` send its final report and exit; if `run_hosted_stream_marker_cli_json` fails below, leaving the `StreamComposeSession` in `workers` only preserves a closed `tx`. The next retry/append will hit `stream compose session is closed`, so the transcript is not actually retryable. The existing regression test only checks that the map entry remains, which misses the dead sender. Please either defer `Finish`/teardown until after the marker publishes, or keep enough finalized transcript state to retry the marker without talking to the exited compose task.

crates/agent-connector/src/stream.rs
```diff
@@ -190,45 +192,53 @@ impl AgentConnector {
...
+                let _ = self.streams.remove_if_same(&stream_id_hex, &session);
+                session.abort.abort();
+                return Err(ConnectorError::Stream(err.to_string()));
+            }
+        };
+        // An `Err` report is a validation mismatch: the compose task is still
+        // running and the session stays registered, so do nothing else here
+        // and the agent can append again and/or retry the finalize.
+        report.map_err(ConnectorError::Stream)?;
+        let _ = self.streams.remove_if_same(&stream_id_hex, &session);
```
erskingardner: Same retryability hole on the connector path: after a matching `Finish`, the compose task has exited and this removes the session before `runtime.finish_agent_text_stream` publishes the durable final. A transient runtime/publish error then returns `stream_error` to the agent with no live session left to retry or append to. The new in-task validation fixes transcript mismatches, but finalization can still strand the stream on the next fallible step; the session should survive until the durable finish succeeds, or the finalized report needs to be retained for retry.

erskingardner: Fixed in f2ead89. The session now survives until the durable finish succeeds. After the compose task validates and exits, the frozen transcript (final_text/hash/chunk_count) is retained on the still-registered session via `mark_finalized`; the session is removed only once `finish_agent_text_stream` succeeds. A transient runtime/publish error leaves the session registered, so a re-issued `StreamFinalize` retries the durable publish directly from the frozen transcript without the (dead) compose task — a retry that disagrees with the frozen transcript is rejected. New white-box test `connector_finalize_retries_durable_finish_without_compose_task` aborts the compose task, then proves both the mismatch rejection and a matching retry that finalizes without it.

crates/cli/src/daemon/stream_workers.rs
erskingardner: Fixed in a7c7eb9. You're right that the retained `tx` was dead. The final `StreamComposeReport` is now frozen on the session (`finalized`) after `Finish`; a re-issued `stream finish` takes a fast-path that re-runs the marker publish directly from the frozen report (via the new `finish_stream_marker_from_report` helper) instead of sending `Finish` to the closed channel, and the session is dropped only once the marker publishes. The regression test now also asserts a second finish attempt does NOT surface "stream compose session is closed" — i.e. it uses the frozen report, not the exited compose task.

crates/marmot-account/src/runtime.rs
erskingardner: Fixed in cc048ff by persisting the failures (your first suggested option), so the repair handle is never lost regardless of what the immediate caller does with the return value. Confirmed create/invite now record each undelivered welcome into a new durable `app_pending_welcome_delivery` table (keyed by the stored welcome's MLS message id + recipient + group). The app surface exposes them: `AppClient` and `MarmotAppRuntime` gain `pending_welcome_deliveries` (list) and `redeliver_welcome` (re-publishes from the engine's stored welcome bytes, no re-commit; clears the record on success, keeps it on repeated failure), wired through the account-worker command channel. This decouples the fix from the `create_group`/`invite`/`SendSummary` signatures (which is why I went with persist over threading them through 14 call sites + UniFFI). The CLI/UniFFI command surface is a noted follow-up; the durable record already ensures the member is repairable across the call return and a restart.

erskingardner (commented):
## Adversarial review summary

This is a strong, well-tested bundle — the core reorderings/compensations look correct, prior review threads on connector/CLI finalize retry and welcome persistence appear addressed, and the failure-injection tests are genuinely probing the widest windows.

I still see a few **#366-class retryability gaps** and one **contract mismatch** worth fixing or documenting before merge:

| Severity | Finding |
| --- | --- |
| **Should fix** | Finalized stream sessions are still eligible for the 5-minute idle sweeper; frozen transcripts are in-memory only (sweep/restart = stranded stream) |
| **Should fix** | `mark_finalized` can silently no-op if the map entry was replaced; finalize still proceeds to the fallible durable publish with no frozen retry state |
| **Consider** | CLI daemon finishes with `expected: None` while the connector validates — asymmetric safety surface |
| **Consider** | `redeliver_welcome` success predicate checks only `failures`, not `welcome_failures` / ack reports |
| **Nit** | `mark_finalized` doc claims idle sweep won't reclaim mid-finalize, but only bumps `last_activity` once |

Detailed comments inline. Happy to re-review after fixes.

crates/agent-connector/src/stream_session.rs
```diff
@@ -355,6 +372,44 @@ impl StreamSessionStore {
...
+        match sessions.get(stream_id_hex) {
+            Some(entry) if entry.tx.same_channel(&session.tx) => sessions.remove(stream_id_hex),
+            _ => None,
+        }
+    }
+
+    /// Freeze the finalize inputs on the stored entry (only when it is still the
+    /// same session) so a durable-finish failure can be retried without the
+    /// compose task, which has exited by this point. Also refreshes activity so
+    /// the idle sweep does not reclaim a session mid-finalize.
```
erskingardner: **Idle sweeper can still destroy a finalized-but-unpublished session.**

This comment says the idle sweep won't reclaim a session mid-finalize, but `sweep_idle` only keys off `last_activity` and never checks `finalized.is_some()`. `mark_finalized` bumps activity once; if the durable `finish_agent_text_stream` publish fails and the agent doesn't retry within `STREAM_SESSION_IDLE_TIMEOUT` (300s), the sweeper removes the session and the frozen transcript is gone — same #366 failure mode (live preview, no durable final, no retry handle).

Consider exempting `finalized.is_some()` entries from sweep entirely (or refreshing activity on every failed durable-finish attempt / heartbeat). A regression test that sweeps a finalized session and asserts retry still works would pin this.

crates/agent-connector/src/stream.rs
```diff
@@ -190,62 +193,129 @@ impl AgentConnector {
...
+        };
+        // An `Err` report is a validation mismatch: the compose task is still
+        // running and the session stays registered, so do nothing else here
+        // and the agent can append again and/or retry the finalize.
+        report.map_err(ConnectorError::Stream)?;
+        // Validation passed and the compose task has now exited. Freeze the
+        // transcript on the still-registered session so a durable-finish
+        // failure below leaves a retryable finalize (the compose task can no
+        // longer be consulted).
+        self.streams.mark_finalized(
```
erskingardner: **`mark_finalized` can silently no-op, but finalize proceeds anyway.**

If the map entry was replaced between `get` and `mark_finalized` (same stream id, different channel), the freeze is skipped yet `finish_finalized_stream` still runs. On a transient publish failure the agent gets no retry fast-path (`session.finalized` is still `None`) and the next attempt re-sends `Finish` to a dead compose task → "stream compose session is closed".

Either treat a no-op freeze as a hard error before attempting the durable publish, or have `finish_finalized_stream` verify `finalized` landed (re-read from the store) and bail if not.

crates/cli/src/daemon/stream_workers.rs
```diff
@@ -601,7 +627,10 @@ pub(crate) async fn finish_stream_compose(
...
+            expected: None,
```
erskingardner, line 639: **Asymmetric validation vs the connector path.**

The connector sends `StreamFinishExpectation` and rejects transcript mismatches before teardown; the CLI daemon still sends `Finish { expected: None }`, so it never validates that the composed transcript matches what the caller thinks it is finishing. If the daemon's local view ever diverges from the compose task (bug, partial crash, manual tampering), it can freeze/publish the wrong final without the guardrail the connector now has.

Worth threading the same expectation through the CLI finish path, or documenting why the daemon is trusted to skip it.

crates/marmot-app/src/client/mod.rs
```diff
@@ -1982,6 +1988,69 @@ impl AppClient {
...
+    /// from the copy the engine stored at wrap time — no re-commit (mdk#352).
+    /// Clears the pending record only once the welcome is accepted; a repeated
+    /// failure leaves it queued for a later retry.
+    pub async fn redeliver_welcome(
+        &mut self,
+        message_id_hex: &str,
+    ) -> Result<SendSummary, AppError> {
+        let message_id = cgka_traits::MessageId::new(hex::decode(message_id_hex)?);
+        let effects = self.runtime.redeliver_welcome(&message_id).await?;
+        if !effects.failures.is_empty() {
```
erskingardner: **Success predicate is narrower than the failure modes this PR introduces.**

Clearing the pending row only checks `effects.failures.is_empty()`. The new `welcome_failures` vector (and `TransportPublishReport::met_required_acks`) are the canonical signals for "still undelivered". Today `publish_one` keeps those in sync, but this is exactly the kind of drift the #375 ack fix was about — I'd assert `welcome_failures.is_empty()` (and/or `reports.last().met_required_acks()`) before deleting the durable pending record so a future refactor can't clear the queue while the welcome is still unreachable.

crates/marmot-app/src/client/mod.rs
```diff
@@ -261,6 +261,7 @@ impl AppClient {
             )
             .await?;
         fail_if_publish_failed(&effects)?;
```
erskingardner, line 268: **Confirmed-but-partial welcomes still return bare success to callers.**

`fail_if_publish_failed` correctly soft-passes confirmed partials (#428), and `record_welcome_delivery_failures` now persists repair handles — good. But `create_group` still returns `Ok(GroupId)` with no indication that a member is unjoinable until someone polls `pending_welcome_deliveries`. That's operable for a daemon/runtime that knows to poll, but any UI/CLI/UniFFI caller that only checks the return type will treat this as fully succeeded. Consider surfacing a warning/flag in the return type or emitting a structured app event when welcomes are queued.

crates/marmot-account/src/runtime.rs
erskingardner: Addressed in cc048ffd — record_welcome_delivery_failures persists to app_pending_welcome_delivery, and AppClient / MarmotAppRuntime now expose pending_welcome_deliveries + redeliver_welcome through the account worker. Remaining gap (also noted in my review): create_group/invite still return bare success without signaling queued welcomes to callers that do not poll.

crates/agent-connector/src/stream_session.rs
erskingardner: Fixed in 204954c. Finalized sessions are now exempt from `sweep_idle` entirely — a finalized session is awaiting a durable-finish retry, not idle, so the sweep skips any entry with `finalized.is_some()` regardless of `last_activity`. It lives until its durable finish succeeds or the connector restarts. New regression test `stream_session_sweep_spares_finalized_session_despite_idle` inserts a finalized session with last_activity an hour past the timeout and asserts the sweep spares it.

crates/agent-connector/src/stream.rs
erskingardner: Fixed in 204954c. `mark_finalized` now returns whether the freeze landed (`#[must_use] -> bool`): it lands only when the entry is still the same session by channel identity. `stream_finalize_response` checks the result and, if the freeze did not land (entry replaced/removed between our get and now), returns a 'stream session was superseded during finalize' error instead of proceeding to the durable publish under a lost retry handle — the agent re-finalizes against the live session.

crates/cli/src/daemon/stream_workers.rs
erskingardner: Documented (1027346) rather than changed, because the daemon has no independent transcript to validate: unlike the connector — where a remote agent supplies its own final_text/hash/chunk over the control protocol, which can disagree with the compose task — `finish_stream_compose` derives the final marker *entirely* from the compose task's own report. The compose task is the single source of truth here, so there is nothing to cross-check `expected` against. Added a comment at the `Finish { expected: None }` site explaining this.

crates/marmot-app/src/client/mod.rs
erskingardner: Fixed in 8490c27. `redeliver_welcome` now clears the durable pending record only when all three 'still undelivered' signals agree it landed: `effects.failures.is_empty() && effects.welcome_failures.is_empty() && reports.last().is_some_and(|r| r.met_required_acks())`. So a future refactor that drifts one signal from another (the #375 class) can no longer clear the queue while the welcome is unreachable. On a not-delivered result it returns a publish error and leaves the record queued.

crates/marmot-app/src/client/mod.rs
erskingardner: Addressed in 8490c27 with the app-event option you suggested. Confirmed create/invite now emit `MarmotAppEvent::WelcomeDeliveryPending { account_id_hex, account_label, group_id, message_id_hex, recipient_hex }` per queued welcome, with a matching UniFFI `MarmotEventFfi::WelcomeDeliveryPending` variant. So a UI/CLI/UniFFI subscriber on the runtime event stream learns a member is unjoinable without polling `pending_welcome_deliveries`, and the event carries the `message_id_hex` handle for `redeliver_welcome`. (I kept `create_group` returning `GroupId` to avoid the 14-caller + UniFFI churn; the event + durable record together cover the signal.)

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -1404,6 +1280,44 @@ impl<S: StorageProvider> Engine<S> {
...
+                        "stable",
+                        pre_commit_epoch,
+                        "auto_commit_stage_failed",
+                        Some(pending_ref),
+                        Some("group_evolution"),
+                    ),
+                );
+                return Err(err.into());
+            }
+        }
```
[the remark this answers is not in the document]
erskingardner: Fixed in eab92c8. A failed `get_group` now unwinds through the same compensation as a failed `put_group`: read-then-project is one fallible unit (`get_group(...).and_then(|g| { ...; put_group(&g) })`), and any error runs `rollback_publish(pending_ref)` + the `auto_commit_stage_failed` audit event and returns `Err`. No pending auto-publish is left without its group-record projection.

crates/marmot-forensics/src/audit.rs
```diff
@@ -1416,29 +1410,39 @@ impl ForensicRecorder for JsonlRecorder {
...
             Err(err) => return Err(err),
         }
-        // Open a brand-new owner-only file at the same path and swap it in.
+        // The handle is write-only rather than append — equivalent for a
+        // brand-new empty file with a single `BufWriter`.
+        let file = fs_private::create_new_private(&staged)?;
+        // Atomic replace: the live path always names either the old complete
+        // file or the fresh empty one, and rename preserves the staged 0600
+        // mode. Recorder state advances only after the rename succeeds.
+        if let Err(err) = std::fs::rename(&staged, &self.path) {
```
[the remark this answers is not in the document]
erskingardner: Not changed — documented instead (1027346). The crate targets Unix only: `fs_private` (and this recorder's owner-only 0600 file model) is Unix-mode based, and CI runs on ubuntu-latest and macos-latest, no Windows. On Unix, POSIX `rename` atomically replaces an existing destination, which is exactly what this swap relies on. The suggested delete-then-rename fallback would reintroduce the non-atomic window (#358) that the swap exists to remove. Added a comment at the rename explaining the Unix assumption.

crates/transport-nostr-adapter/src/lib.rs
```diff
@@ -553,23 +570,55 @@ impl TransportAdapter for NostrTransportAdapter {
...
+            state.record_subscription_starts(&to_add, now_ms);
+            state.sync_groups(sync, to_add.len());
+            state.queue_pending_unsubscribes(to_remove);
+            state.take_drainable_unsubscribes()
+        };
+
+        let mut confirmed = 0_usize;
+        let mut requeue = Vec::new();
+        for subscription in drainable {
+            match self.relay_client.unsubscribe(subscription.clone()).await {
```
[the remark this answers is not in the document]
erskingardner: Fixed in de0f1a2 by serializing the drain/update boundary, as you suggested. `activate_account` / `deactivate_account` / `sync_account_groups` now hold a dedicated async `subscription_lock` for their whole duration, so their relay subscribe/unsubscribe calls — which run outside the state `RwLock` — cannot interleave. A concurrent sync can no longer re-add a subscription id in the window between another sync's live-route filter and its unsubscribe. The `RwLock` still guards the delivery/routing hot path independently. New test `concurrent_group_syncs_do_not_overlap_relay_unsubscribes` runs two syncs concurrently and asserts the relay never sees two unsubscribes in flight at once.

crates/agent-connector/src/stream.rs
```diff
@@ -196,62 +193,129 @@ impl AgentConnector {
...
+            },
+        );
+        self.finish_finalized_stream(
+            &stream_id_hex,
+            &session,
+            final_text,
+            transcript_hash,
+            chunk_count,
+        )
+        .await
```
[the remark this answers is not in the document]
erskingardner: Addressed in 204954c. `mark_finalized` now returns whether the freeze landed and `stream_finalize_response` bails (superseded error) rather than publishing under a lost retry handle if a concurrent finalize replaced/removed the entry first. Combined with the sweep exemption for finalized sessions, a second finalize can no longer strand the stream: either it finds the frozen snapshot (retry fast-path) or it is rejected before the durable publish.

PR #735, opened 2026-07-04
Standardize wire-boundary validation fixes

### LINKED ISSUE #349: `build_key` concatenates label+key with no delimiter (undelimited-concatenation collision hazard, same class as fixed #158)
Found via review of the value-store key derivation. (Uncertain — not exploitable with current labels.)

## What

`build_key` forms `storage_key = label || key || version` with no length prefix/delimiter between `label` and `key`. Not exploitable today (all labels are distinct fixed constants and none is a prefix of another), but it is exactly the undelimited-concatenation hazard that marmot-protocol/mdk#697 fixed for `epoch_key_pairs_id`, left in the generic path. A future label that is a prefix of an existing one would silently collide (`INSERT OR REPLACE` clobber).

## Where

- `crates/storage-sqlite/src/labels.rs:24-29`.

## Fix

Length-prefix the label (mirror the tuple encoding adopted in `epoch_key_pairs_id` for marmot-protocol/mdk#697).

---

### LINKED ISSUE #386: cgka-engine: inbound KeyPackages are never checked for lifetime/expiry
### Summary
The engine never imposes a max-age / `not_after` policy on inbound invitee KeyPackages. A stale or replayed KeyPackage fetched from a relay is accepted indefinitely.

### Location
- `crates/cgka-engine/src/key_package.rs:41-48` (`key_package_metadata`), `:158-184` (`parse_key_package`)
- Welcome path: `crates/cgka-engine/src/group_lifecycle.rs:480`

### Details
`kp_in.validate(&crypto, ProtocolVersion::Mls10)` only enforces a `Lifetime` extension *if the KeyPackage carries one*. The engine builds its own packages with `.mark_as_last_resort()` and no `Lifetime`, and there is no freshness gate anywhere in the credential-ingress path (grep confirms no `Lifetime` / `not_after` / `expir` handling in the engine identity code).

### Impact
Combined with last-resort reuse, an invitee can be added to a group on an arbitrarily old leaf, defeating the forward-secrecy intent of KeyPackage rotation, and a replayed KeyPackage remains usable indefinitely. Inbound *acceptance* policy is an engine concern even though refresh *scheduling* is a higher layer.

### Suggested fix
Add an explicit freshness gate in `parse_key_package` / `key_package_metadata`: reject KeyPackages whose `Lifetime` (if present) is expired, and define a policy for last-resort packages (e.g. require a `Lifetime` and cap max age) rather than silently accepting any age.

---

---

### LINKED ISSUE #398: marmot-uniffi: `group_id_from_hex` skips the 32-byte length check its sibling helpers enforce
### Summary
`group_id_from_hex` decodes hex and calls `GroupId::new(bytes)` with no length validation, while message-id helpers reject non-32-byte input.

### Location
`crates/marmot-uniffi/src/conversions/common.rs:30-36` (vs `crates/marmot-uniffi/src/lib.rs:93` for message ids).

### Impact
Correctness/consistency at the FFI edge (not a panic — `GroupId` is a bare `Vec<u8>`). A truncated/oversized id is silently accepted and surfaces later as `UnknownGroup` instead of a clear `InvalidHex`. Most group FFI commands build their id through this path.

### Suggested fix
Reject `bytes.len() != 32` with `MarmotKitError::InvalidHex`, matching `optional_message_id_hex`. Related tracking issue: marmot-protocol/marmot#77.

---

---

erskingardner:
## Summary

- length-delimit OpenMLS value-store keys and migrate/delete legacy concatenated rows
- reject non-32-byte UniFFI group id hex at the boundary
- enforce KeyPackage lifetime validity/range checks at parse/invite boundaries

## Verification

- cargo test -p storage-sqlite openmls_storage::value_store::tests -- --nocapture
- cargo test -p marmot-uniffi optional_group_id_hex_trims_and_canonicalizes -- --nocapture
- cargo test -p cgka-engine keypackage --test group_creation -- --nocapture
- just fast-ci

Closes #349
Closes #386
Closes #398
Refs marmot-protocol/marmot#77

Spec-side companion PR: marmot-protocol/marmot#236.

## Summary by CodeRabbit

* **New Features**
  * KeyPackage validation now enforces Marmot lifetime-range acceptance during transported KeyPackage parsing.
  * SQLite-backed OpenMLS value storage now uses a length-delimited key format and migrates legacy entries forward on write/delete.

* **Bug Fixes**
  * Group ID hex inputs are rejected unless they decode to exactly 32 bytes.
  * Group creation now rejects invitee KeyPackages with expired lifetimes or lifetime ranges exceeding Marmot limits.

* **Documentation & Tests**
  * Updated Marmot kind 30443 last-resort/caching behavior guidance and refresh/rotation documentation; added/expanded lifetime-validation test coverage.

### DIFF
--- a/crates/cgka-engine/src/key_package.rs
+++ b/crates/cgka-engine/src/key_package.rs
@@ -2,7 +2,7 @@
 //!
 //! Scope for 0.1.0: produce a fresh KeyPackage whose leaf capabilities are
 //! derived from the [`crate::FeatureRegistry`]. Transported KeyPackages are
-//! validated for OpenMLS lifetime validity and Marmot lifetime range policy;
+//! validated for OpenMLS lifetime validity and accepted lifetime range;
 //! refresh scheduling is handled above the engine.
 
 use crate::capabilities::leaf_capabilities;
@@ -12,11 +12,12 @@ use cgka_traits::engine::KeyPackage;
 use cgka_traits::error::EngineError;
 use cgka_traits::storage::StorageProvider;
 use openmls::prelude::{
-    Extensions, KeyPackage as MlsKeyPackage, MlsMessageBodyIn, MlsMessageIn, MlsMessageOut,
-    ProtocolVersion,
+    Extensions, KeyPackage as MlsKeyPackage, KeyPackageVerifyError, MlsMessageBodyIn, MlsMessageIn,
+    MlsMessageOut, ProtocolVersion,
 };
 use openmls_rust_crypto::RustCrypto;
 use openmls_traits::OpenMlsProvider as _;
+use openmls_traits::crypto::OpenMlsCrypto;
 use tls_codec::{Deserialize as _, Serialize as _};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -39,10 +40,7 @@ pub fn key_package_metadata(kp: &KeyPackage) -> Result<KeyPackageMetadata, Engin
         }
     };
     let crypto = RustCrypto::default();
-    let key_package = kp_in
-        .validate(&crypto, ProtocolVersion::Mls10)
-        .map_err(|e| EngineError::Backend(format!("key_package validate: {e:?}")))?;
-    validate_key_package_lifetime_policy(&key_package)?;
+    let key_package = validate_key_package(kp_in, &crypto)?;
     let member_id = crate::identity::validated_member_id_of_leaf(key_package.leaf_node())?;
     crate::account_identity_proof::validate_leaf_account_identity_proof(
         key_package.leaf_node(),
@@ -71,10 +69,7 @@ pub fn is_last_resort_key_package(kp: &KeyPackage) -> Result<bool, EngineError> 
         }
     };
     let crypto = RustCrypto::default();
-    let key_package = kp_in
-        .validate(&crypto, ProtocolVersion::Mls10)
-        .map_err(|e| EngineError::Backend(format!("key_package validate: {e:?}")))?;
-    validate_key_package_lifetime_policy(&key_package)?;
+    let key_package = validate_key_package(kp_in, &crypto)?;
     crate::identity::validated_member_id_of_leaf(key_package.leaf_node())?;
     crate::account_identity_proof::validate_leaf_account_identity_proof(
         key_package.leaf_node(),
@@ -140,9 +135,7 @@ impl<S: StorageProvider> Engine<S> {
             }
         };
         let provider = EngineOpenMlsProvider::<S>::new(&self.crypto, self.storage.mls_storage());
-        let key_package = kp_in
-            .validate(provider.crypto(), ProtocolVersion::Mls10)
-            .map_err(|e| EngineError::Backend(format!("key_package validate: {e:?}")))?;
+        let key_package = validate_key_package(kp_in, provider.crypto())?;
         let hash_ref = key_package
             .hash_ref(provider.crypto())
             .map_err(|e| EngineError::Backend(format!("key_package ref: {e:?}")))?;
@@ -172,10 +165,7 @@ impl<S: StorageProvider> Engine<S> {
         };
 
         let provider = EngineOpenMlsProvider::<S>::new(&self.crypto, self.storage.mls_storage());
-        let key_package = kp_in
-            .validate(provider.crypto(), ProtocolVersion::Mls10)
-            .map_err(|e| EngineError::Backend(format!("key_package validate: {e:?}")))?;
-        validate_key_package_lifetime_policy(&key_package)?;
+        let key_package = validate_key_package(kp_in, provider.crypto())?;
         // foundation/key-packages.md: reject a KeyPackage whose credential
         // identity is not a valid Marmot account identity. This single gate
         // covers both the create-group and invite invitee paths.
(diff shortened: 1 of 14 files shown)


erskingardner:
@coderabbitai full review

crates/cgka-engine/src/key_package.rs
```diff
@@ -183,3 +187,15 @@ impl<S: StorageProvider> Engine<S> {
...
 }
+
+fn validate_key_package_lifetime_policy(key_package: &MlsKeyPackage) -> Result<(), EngineError> {
+    let lifetime = key_package.life_time();
+    if !lifetime.has_acceptable_range() {
+        return Err(EngineError::Backend(format!(
+            "key_package lifetime range exceeds Marmot maximum: not_before={} not_after={}",
+            lifetime.not_before(),
+            lifetime.not_after()
+        )));
```
erskingardner: **Two issues in this error string.**

**1. "Marmot maximum" mischaracterizes an OpenMLS-owned constant.** `Lifetime::has_acceptable_range()` checks against `MAX_LEAF_NODE_LIFETIME_RANGE_SECONDS` — a private, hardcoded constant inside the `openmls` crate (≈3 months), not a Marmot-configured value. Nothing in this diff defines a Marmot maximum. The text `"exceeds Marmot maximum"` will send the next reader hunting for a Marmot policy constant to tune when this is purely deferring to OpenMLS's built-in ValSem gate. Suggest something like `"key_package lifetime range exceeds the accepted maximum (openmls ValSem check)"`. Same framing should be softened in `crates/cgka-engine/AGENTS.md:66` ("lifetime/range acceptance policy") and the module doc at the top of this file ("Marmot lifetime range policy").

**2. Privacy: raw timestamps in `Display` reach the forensic audit `detail`.** This error's `to_string()` flows verbatim into `AuditEventKind::HumanAction { detail, .. }` via `audit_helpers::engine_error_detail` → `engine.rs` (`detail: engine_error_detail(err)`). `docs/marmot-architecture/overview/observability.md` says error `Display` should stay free of raw values and logs should carry `error_kind` only. Every other `EngineError::Backend` constructor in this crate wraps an opaque library error via `{e:?}`; this is the first to intentionally compose two raw `u64` domain values (`not_before`/`not_after`) into the message. The typed-variant fix below resolves this cleanly (keep the timestamps off `Display`, or omit them from the audit `detail` mapping).

crates/cgka-engine/src/key_package.rs
```diff
@@ -183,3 +187,15 @@ impl<S: StorageProvider> Engine<S> {
         Ok(key_package)
     }
 }
+
+fn validate_key_package_lifetime_policy(key_package: &MlsKeyPackage) -> Result<(), EngineError> {
```
erskingardner, line 204: **Design: reuse of the stringly-typed `Backend` variant instead of a typed error.** The sibling wire-boundary rejections in this same file use purpose-built variants — `InvalidCredentialIdentity(_)`, `InvalidAccountIdentityProof(_)` — both of which are branched on downstream in `audit_helpers.rs::engine_error_kind` (mapped to dedicated forensic reason codes) and in the conformance simulator. Because this check has no typed variant, any caller wanting to distinguish "excessive lifetime range" from a generic library failure (e.g. to decide silently-rotate vs. hard-error, or to emit a precise `error_kind`) is forced into substring matching — as the PR's own tests already are. A variant like `InvalidKeyPackageLifetime { not_before, not_after }` would fall through the audit-mapping pattern already established for the siblings here, make the two test assertions robust to wording changes, and (per the comment below) let `Display` stay free of the raw timestamps.

crates/cgka-engine/tests/group_creation.rs
```diff
@@ -534,6 +575,64 @@ async fn create_group_rejects_invitee_keypackage_with_mismatched_account_identit
...
+            description: "".into(),
+            members: vec![excessive_kp],
+            required_features: vec![],
+            app_components: vec![],
+            initial_admins: vec![],
+        })
+        .await
+        .expect_err("create_group must reject an over-long invitee KeyPackage");
+    assert!(
+        matches!(err, EngineError::Backend(ref message) if message.contains("lifetime range exceeds Marmot maximum")),
```
erskingardner: This assertion (and the `message.contains("InvalidLifetime")` one at the expired test above) substring-matches free error text purely because there's no typed variant to `matches!` on — so a benign reword of the error string in `validate_key_package_lifetime_policy` silently breaks the test. A typed `EngineError::InvalidKeyPackageLifetime { .. }` variant (see the key_package.rs comment) would make both assertions structural.

Minor observation, not a defect: the expired-KP path here is actually rejected by OpenMLS's own `validate()` (hence `"InvalidLifetime"`), *before* reaching the new range check — so the new `validate_key_package_lifetime_policy` only adds the too-long-range gate. Worth confirming that's the intended division of labor, since the function name/comments imply it owns expiry too.

erskingardner (commented):
Adversarial review — all three fixes are legitimate, well-scoped wire-boundary validation gaps, each with a targeted test, and the storage-migration logic (the bulk of the diff) holds up under close inspection: reads, writes, appends, removes, and deletes all fall back to and clean up the legacy key format under one transaction, with no torn-write or stale-legacy-row scenario I could construct. The two source-inspection guard tests are a cheap, durable way to keep the single-transaction invariant enforced by CI.

My findings all cluster on the one genuinely new engine surface — `validate_key_package_lifetime_policy` — and are about wording/typing/privacy rather than correctness. See the three inline comments. Two smaller notes:

- `crates/storage-sqlite/src/openmls_storage/labels.rs` — the `build_key_legacy` doc comment explaining the collision bug is excellent; consider echoing a one-line "why length-prefix" note above `build_key` itself so a reader skimming only the new format understands the motivation without scrolling to the legacy helper.
- `crates/cgka-engine/tests/group_creation.rs` — `key_package_with_account_identity_proof_and_lifetime` is a near-clone of `key_package_with_mismatched_account_identity_proof` differing only by the `.key_package_lifetime(...)` call; consider having one delegate to the other to keep the builder pipeline in one place as more fixtures accumulate.

Nice touch: the `group_id_from_hex` 32-byte check is a single fix point that covers ~40 FFI call sites automatically, and the `marmot-app` AGENTS.md/README wording updates stay in sync with the actual `publish_key_package` behavior.

PR #751, opened 2026-07-05
Stabilize agent connectors and fill CLI/TUI surfaces

erskingardner:
## Summary

- Stabilize `wn-agent` finalization with optional `stream_finalize` idempotency keys and matching Hermes/OpenClaw retry behavior.
- Bring Hermes and OpenClaw closer to parity for install/dev setup, generated helpers, deterministic connector E2Es, and onboarding docs.
- Fill CLI/TUI gaps with real backing behavior for invites, notifications, key deletion/listing, stream compose, chat mute/unmute, TUI image upload, and visible `wn sync`.
- Add app/storage backing for per-chat notification suppression and tighten CLI JSON error contracts.

## Verification

- `just fast-ci`
- `cargo test -p wn-cli`
- `cargo test -p agent-control -p agent-connector -p agent-stream-compose`
- `python3 -m unittest discover -s integrations/hermes/marmot/tests`
- `cd integrations/openclaw/marmot && pnpm typecheck && pnpm test`
- `integrations/hermes/marmot/tests/test_dev_scripts.sh`
- `integrations/openclaw/marmot/test/dev-scripts.sh`
- Hermes deterministic E2E via `scripts/hermes_marmot_deterministic_e2e.sh`
- Hermes real `wn-agent` connector E2E via `scripts/hermes_marmot_connector_e2e.sh`
- OpenClaw real `wn-agent` connector E2E via `scripts/openclaw_marmot_connector_e2e.sh`

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/751">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added chat mute/unmute controls, notification subscriptions, group invite actions, key-package deletion, and improved stream finalization reliability.
  * Added new dev helper commands and scripts for smoke tests and deterministic connector E2E runs.
  * Expanded TUI support for `/chat mute`, `/chat unmute`, and `/image`.

* **Bug Fixes**
  * Improved error messages and validation for daemon-required commands, mute durations, relay types, and confirmation flows.
  * Fixed notification delivery to respect per-chat mute settings and retry-safe finalize behavior.

### DIFF
--- a/crates/agent-connector/src/bin/wn-agent.rs
+++ b/crates/agent-connector/src/bin/wn-agent.rs
@@ -340,9 +340,9 @@ fn safe_error_message(err: &ConnectorError) -> String {
     match err {
         ConnectorError::Io(io) => {
             format!(
-                "startup failed code={} detail={}",
+                "startup failed code={} io_kind={:?}",
                 err.privacy_safe_code(),
-                io
+                io.kind()
             )
         }
         _ => format!("startup failed code={}", err.privacy_safe_code()),
--- a/crates/agent-connector/src/bin/wn-agent.rs
+++ b/crates/agent-connector/src/bin/wn-agent.rs
@@ -337,7 +337,16 @@ fn which_qrencode() -> Option<String> {
 }
 
 fn safe_error_message(err: &ConnectorError) -> String {
-    format!("startup failed code={}", err.privacy_safe_code())
+    match err {
+        ConnectorError::Io(io) => {
+            format!(
+                "startup failed code={} detail={}",
+                err.privacy_safe_code(),
+                io
+            )
+        }
+        _ => format!("startup failed code={}", err.privacy_safe_code()),
+    }
 }
 
 fn read_auth_token(path: Option<&PathBuf>) -> Result<Option<String>, String> {
--- a/crates/cli/src/commands/key_package.rs
+++ b/crates/cli/src/commands/key_package.rs
@@ -1,13 +1,12 @@
 //! `keys` (KeyPackage) command namespace handlers and output helpers.
 
 use marmot_account::AccountHome;
-use marmot_app::{FetchedKeyPackage, MarmotApp, MarmotAppRuntime};
+use marmot_app::{AccountKeyPackageRecord, FetchedKeyPackage, MarmotApp, MarmotAppRuntime};
 use serde_json::{Value, json};
 
 use crate::{
     CommandOutput, KeyPackageCommand, WnError, account_selector_or_default, ensure_local_signing,
     npub_for_account_id, parse_public_key, relay_endpoints, relay_lists_json, resolve_account,
-    unsupported_command,
 };
 
 pub(crate) async fn key_package_command(
@@ -30,22 +29,12 @@ pub(crate) async fn key_package_command_with_runtime(
     match command {
         KeyPackageCommand::List => {
             let account = resolve_account(account_home, account_flag)?;
-            let relay_lists =
-                app.account_relay_list_status_for_account_id(&account.account_id_hex)?;
-            let fetched = if relay_lists.nip65.relays.is_empty() {
-                None
-            } else {
-                Some(
-                    app.fetch_latest_key_package_for_account_id(
-                        &account.account_id_hex,
-                        relay_endpoints(relay_lists.nip65.relays.clone())?,
-                    )
-                    .await?,
-                )
-            };
-            let keys = fetched
+            let records = runtime
+                .account_key_packages(&account.label, Vec::new())
+                .await?;
+            let keys = records
                 .into_iter()
-                .map(key_package_fetch_json)
+                .map(account_key_package_record_json)
                 .collect::<Vec<_>>();
             Ok(CommandOutput {
                 plain: if keys.is_empty() {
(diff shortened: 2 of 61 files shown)


crates/agent-connector/src/bin/wn-agent.rs
```diff
@@ -337,7 +337,16 @@ fn which_qrencode() -> Option<String> {
 }
 
 fn safe_error_message(err: &ConnectorError) -> String {
-    format!("startup failed code={}", err.privacy_safe_code())
+    match err {
+        ConnectorError::Io(io) => {
+            format!(
+                "startup failed code={} detail={}",
```
erskingardner: **Blocking (privacy): this branch defeats the helper's own contract.**

Every other `ConnectorError` variant here goes through `privacy_safe_code()` — a fixed `&'static str` with no interpolated detail, by design (see `error.rs`, where `client_message()`/`code()` deliberately return static strings). This new `Io` branch is the sole exception: it `Display`-formats the raw `io::Error` into a message printed to stderr on every startup failure.

That's a concrete path-leak vector, not hypothetical — `crates/fs-private/src/lib.rs` constructs `io::Error`s whose text embeds `path.display()` (e.g. `"{path} already exists"`), and those flow into `ConnectorError::Io` via `#[from]` and can reach this branch during socket-path prep in `serve_socket`. A `--home`/socket path routinely embeds the OS username (`/Users/<user>/...`), exactly what the logging-privacy invariant keeps out of any "safe" surface. This also slips past both automated guards: the `tracing_audit` `DIRECT_OUTPUT_MACROS` check exempts `src/bin/`, and `FORBIDDEN_ERROR_INTERPOLATIONS` only scans `tracing::*!` macro bodies — a bare `format!` is invisible to both.

Severity is lower than a telemetry leak (local operator stderr), but the function is literally named `safe_error_message`. Drop the `detail=` field, or restrict it to `io.kind()` (an enum, not `Display`) if finer diagnosis is needed.

crates/cli/src/commands/key_package.rs
```diff
@@ -132,30 +121,104 @@ pub(crate) async fn key_package_command_with_runtime(
...
+            let mut accepted_relays = 0_usize;
+            for record in records.into_iter().filter(|record| record.relay) {
+                if deleted.iter().any(|deleted: &DeletedKeyPackage| {
+                    deleted.event_id == record.key_package_event_id
+                }) {
+                    continue;
+                }
+                let relays = relay_endpoints(record.source_relays.clone())?;
+                let accepted = runtime
+                    .delete_key_package(&account.label, &record.key_package_event_id, relays)
```
erskingardner, line 174: **Blocking (torn-write): `keys delete-all` aborts the batch on the first per-record failure and discards already-applied deletions.**

`delete_key_package` → `delete_key_package_event` publishes a kind-5 Nostr deletion requiring at least one relay ack; if a record's relay set is unreachable this hard-errors. The `?` here propagates immediately out of `key_package_command_with_runtime`, discarding the `deleted` vec accumulated so far. So if record 3 of 5 fails, the CLI reports a bare error and the caller never learns that records 1–2 were already deleted — irreversibly, since those deletion events are already published.

This repo already solves this exact class of problem: the `sign_out_and_wipe` path (`crates/marmot-app/src/runtime/mod.rs`) does best-effort per-relay KeyPackage cleanup and records failures in `SignOutOutcome::key_package_failures` for the same delete-many-KeyPackages operation. `keys delete-all` should follow the same aggregate-and-report shape (`deleted` + `failed: Vec<{event_id, error}>`) rather than losing partial progress on the first error. The existing test (`keys_delete_and_delete_all_use_runtime_relay_deletion`) only covers the all-succeed case, so it wouldn't catch this.

scripts/openclaw_marmot_dev_teardown.sh
```diff
@@ -20,18 +22,23 @@ if [ ! -d "$ROOT" ]; then
...
+    exit 0
+fi
+
+if [ "$FORCE" -ne 1 ]; then
+    printf 'Delete dev root %s? [y/N] ' "$ROOT"
+    read -r reply
+    case "$reply" in y|Y|yes|YES) ;; *) echo "aborted"; exit 1;; esac
+fi
+
 rm -rf "$ROOT"
```
erskingardner: **Dev-script parity regression (the header says "Mirrors the Hermes teardown" — but it doesn't, in two ways):**

1. **It never stops the background processes before `rm -rf`.** `openclaw_marmot_dev_setup.sh` generates `start-wn-agent.sh`/`start-openclaw-gateway.sh` (which write `wn-agent.pid`/`openclaw-gateway.pid`) and a `stop-dev-processes.sh`. Hermes's teardown calls `"$dev_root/stop-dev-processes.sh" || true` before deleting; this one goes straight to `rm -rf`, deleting the socket/home dir out from under a live `wn-agent`/gateway and orphaning the process with stale fds. The new `test/dev-scripts.sh` never starts a background process before teardown, so CI won't catch it.

2. **The unsafe-path guard is materially weaker than Hermes's.** This is a flat `case "$ROOT"` string match with no `realpath`/`cd && pwd` resolution and no "must be within a tmp dir" check. Hermes resolves the real path first, blocks far more system dirs (`/etc`,`/usr`,`/bin`,`/sbin`,`/System`,`/Library`,`$PWD`), and requires the resolved root live under `/tmp`, `/var/tmp`, `$TMPDIR`, or `~/.local/share`. A relative `--root`, a trailing slash, or a symlinked path could bypass this check in ways Hermes is hardened against — and `rm -rf` is the gated op. Worth pulling the Hermes `same_or_within`/`within_root` helpers over.

erskingardner (commented):
## Adversarial review

Solid, mechanically consistent PR. I compiled it, ran clippy/`fmt --check` (clean), and ran the PR's own named suites against the branch (`wn-cli` 77 passed; `agent-connector`/`agent-control`/`agent-stream-compose` 125 passed). The `stream_finalize` idempotency work reuses the existing hardened `SendIdempotencyStore` with a `stream_finalize:` key prefix (good infra reuse), Hermes/OpenClaw retry parity is genuinely matched and tested, and the chat-mute path validates group existence before mutating at every layer.

Three issues worth addressing before merge are left inline. Two are blocking (a privacy-invariant regression and a torn-write gap that deviates from a pattern this repo already solves correctly); one is a dev-script parity regression in the OpenClaw teardown.

### Non-blocking suggestions

- **`crates/marmot-app/src/notifications.rs`** — `chat_muted(...)?` now runs before the graceful `group()` lookup in `notification_update_from_message`. `chat_notification_settings` errors `UnknownGroup` for a missing group, whereas `group()` treats missing as `None`. Not a functional bug (both consumers catch `Err(_)`), but a routine race (message arriving just before/after its group is projected/left) will now emit a warning log where it previously silently no-op'd. Consider special-casing `AppError::UnknownGroup` alongside `NotificationsDisabled` in those match arms to avoid warning noise.
- **`crates/storage-sqlite/src/account_projection.rs`** — `chat_notification_settings_at` defaults an unmuted chat to `muted_until_ms: Some(0)` while `None` means "muted forever." It's correct and well-tested, but `Some(0)` as the "not muted" sentinel next to `None` = "forever" is surprising to read; a one-line comment at the default tuple would save the next reader a re-derivation. (Also a latent trap: a caller computing `now_ms <= 0` — clock rollback, a seconds/ms unit mixup via FFI — would read the unmuted default as muted.)
- **CHANGELOG** — the breaking JSON error-contract change (`unsupported_command` → `confirmation_required`/`invalid_mute_duration`/`invalid_relay_type`/etc.) is documented, good. Worth double-checking no downstream consumer (dev scripts, dashboards) still matches on the retired `"unsupported_command"` code string.
- **`crates/agent-connector/AGENTS.md`** — still describes `SendIdempotencyStore` only in terms of `send_final`; a one-line note that `stream_finalize` now shares it under a prefixed key would help future readers.

Issue #747, opened 2026-07-05
cgka-engine: directory-facing KeyPackage helpers validate the account-identity proof against DEFAULT_CIPHERSUITE instead of the KeyPackage's own ciphersuite


erskingardner:
**Severity:** LOW (latent; benign under single-ciphersuite deployment)
**Component:** `crates/cgka-engine`

## Summary

`key_package_metadata` and `is_last_resort_key_package` validate the leaf account-identity proof with a hardcoded `DEFAULT_CIPHERSUITE`, whereas the real invite path (`parse_key_package`) correctly uses the KeyPackage's own ciphersuite. `validate_leaf_account_identity_proof` strictly checks `proof.request.ciphersuite == u16::from(ciphersuite)` and `proof.request.signature_scheme == ciphersuite.signature_algorithm()`, so these two helpers will disagree with `parse_key_package` the moment a second ciphersuite is supported.

## Location

`crates/cgka-engine/src/key_package.rs:45-48` (`key_package_metadata`) and `:76-79` (`is_last_resort_key_package`).

## Details

```rust
crate::account_identity_proof::validate_leaf_account_identity_proof(
    key_package.leaf_node(),
    crate::DEFAULT_CIPHERSUITE,   // should be key_package.ciphersuite()
)?;
```

Harmless while the engine is single-ciphersuite. Once a second ciphersuite exists, a validly-proofed KeyPackage for the non-default ciphersuite is wrongly rejected by these transport-directory checks (or the checks disagree with `parse_key_package`), producing inconsistent accept/reject decisions for the same KeyPackage across the codebase.

## Suggested fix

Derive the ciphersuite from the validated KeyPackage (`key_package.ciphersuite()`) and pass that to `validate_leaf_account_identity_proof`, matching `parse_key_package`.

---
_Filed from a meticulous automated code-review pass; verified against source and deduplicated against existing open/closed issues._

Issue #746, opened 2026-07-05
cgka-engine: invitee KeyPackage omitting the admin-policy component silently negotiates it out of the created group (empty admin set, frozen membership)


erskingardner:
**Severity:** LOW (griefing / self-inflicted DoS; fail-closed, no privilege escalation)
**Component:** `crates/cgka-engine`

## Summary

`default_group_components()` includes `GROUP_ADMIN_POLICY_COMPONENT_ID`, but that id lives in `desired_components`, not in `required_caps.app_components`. The negotiated component set is intersected against every invitee's advertised leaf components, so an invitee whose KeyPackage leaf omits the admin-policy component id causes admin-policy to be dropped from the required set — the group is then created with **no admin-policy bytes and an empty admin set**.

## Location

`crates/cgka-engine/src/group_lifecycle.rs:102-124`. Component id at `crates/traits/src/app_components/mod.rs:100-104`.

## Details

```rust
let mut negotiated_components = desired_components.intersection(&self.supported_app_components);
for kp in &req.members {
    ...
    negotiated_components = negotiated_components.intersection(&had.app_components); // invitee can drop admin-policy
}
required_caps.app_components = negotiated_components;
```

With an empty admin set, `admins_of_group()` is empty, so `require_admin` fails closed for everyone including the creator: all admin-gated operations (add/remove members, GCE, app-data updates) are permanently rejected, and new joiners are rejected outright by `do_join_welcome` step 5d (`require_admin(welcome_sender)` on an empty admin set returns `NotGroupAdmin`). Membership changes for the group are frozen. Because it is fail-closed there is no escalation, only denial.

## Suggested fix

Treat the engine-owned admin-policy component as non-negotiable: after negotiation, assert `required_caps.app_components` still contains `GROUP_ADMIN_POLICY_COMPONENT_ID` (and the profile component), or reject any invitee lacking it with `MissingRequiredCapabilities`.

---
_Filed from a meticulous automated code-review pass; verified against source and deduplicated against existing open/closed issues._

Issue #737, opened 2026-07-05
cgka-engine: group creation and welcome-join skip the admin-leaf-coupling check enforced on every commit seam (phantom/pre-provisioned admins)


erskingardner:
**Severity:** MEDIUM
**Component:** `crates/cgka-engine`

## Summary

Every **commit** seam runs `validate_admin_leaf_coupling_for_staged_commit`, which rejects any epoch where an admin key has no corresponding member leaf (verified at `message_processor/send.rs:455`, `message_processor/ingest.rs:788`, `openmls_projection.rs:1854`). But the **initial group state** at creation and the **welcome-join acceptance** never run that check. This lets a group be created with admin keys that do not (yet) correspond to any member — a "phantom"/pre-provisioned admin that becomes active the instant a matching leaf appears, with no `AdminAdded` commit/event that other members would observe. The crate invariant is explicit: "a guard that exists on one seam only is a bug" (`crates/cgka-engine/AGENTS.md`).

## Location

- `crates/cgka-engine/src/group_lifecycle.rs:136-154` — `do_create_group` admin-set assembly.
- `crates/cgka-engine/src/group_lifecycle.rs:489-533` — `do_join_welcome` validation steps 5b–5d.

## Details

At creation the admin set is built from the creator plus `req.initial_admins`, with only a 32-byte length check (via `admin_pubkey_from_member_id`) and no coupling to a member leaf:

```rust
let mut admin_set: Vec<[u8; 32]> = vec[creator_pubkey];
for extra in &req.initial_admins {
    let pk = crate::app_components::admin_pubkey_from_member_id(extra)?; // length check only
    if !admin_set.contains(&pk) { admin_set.push(pk); }
}
```

`req.initial_admins` is independent of `req.members` (the invitees). `do_join_welcome` validates member credentials/proofs (5b), required capabilities (5c), and `require_admin(welcome_sender)` (5d), but never validates that the group's admin set is leaf-coupled.

## Failure scenario

[...]

Issue #728, opened 2026-07-04
cgka-engine: third sender resolver in app_components.rs bypasses validate_credential_identity


erskingardner:
### Summary

While collapsing the direct-vs-replay sender-resolver pairs in #726 (part of #707), a third, deliberately-looser sender resolver surfaced that the "mirror on all seams" contract does not currently cover.

### Location

`crates/cgka-engine/src/app_components.rs` — a private `member_id_of_sender` helper (distinct from `identity::member_id_of_sender`) used by the admin-self-remove policy check against a carried-forward admin set.

### Details

`identity::member_id_of_sender` (the shared chokepoint used by the direct-ingest and stored-convergence/replay seams as of #726) runs `validate_credential_identity` (32-byte x-only secp256k1 / BIP-340 check) before trusting a leaf's identity as a `MemberId`. The `app_components.rs` variant builds the id via a raw `BasicCredential::try_from` **without** that validation.

This is not a direct/replay pair, so it was intentionally left out of the #726 collapse to avoid changing admin-policy behavior for malformed identities. But it means a malformed credential identity that would be rejected at the two main seams could still be compared against the admin set on the self-remove policy path.

### Suggested action

Either converge it with `identity::member_id_of_sender` (strict validation) after confirming no admin-policy behavior regression for malformed identities, or document at the call site why it is deliberately looser.

### Context

Spun off from #707 / #726. Low priority (narrow reachability), but worth closing the loop so the "one sender resolver per authenticated ingress" intent holds everywhere.

PR #752, opened 2026-07-05
cgka-engine: bound the future side of the convergence send-gate (#736), scope superseded scan (#745)

### LINKED ISSUE #736: cgka-engine: unbounded future-epoch convergence input permanently gates all outbound sends (member-inflicted durable DoS)
**Severity:** HIGH
**Component:** `crates/cgka-engine`

## Summary

Any single group member can permanently disable another member's ability to send messages or evolve a group by broadcasting **one** crafted message. The convergence store buffers inbound commit/application messages keyed by an **attacker-chosen `source_epoch` with no upper bound**. A message forged with a far-future epoch (e.g. `2^63`) is stored as a durable `MessageState::Created` row that convergence never reaches, never materializes, and never assigns a terminal disposition. From then on `has_unresolved_convergence_inputs` reports the group as unsettled forever, so `do_send` queues every intent instead of sending and the queue never drains. The effect is durable (survives restart) and has no recovery path short of manual DB surgery.

## Root cause

The past side of the convergence horizon is bounded, but the future side is not.

1. **Buffering accepts any epoch.** `buffer_openmls_convergence_message` reads the epoch straight off the (plaintext, `PURE_PLAINTEXT_WIRE_FORMAT_POLICY`) MLS header via `project_mls_message` and stores the row with that epoch and no bound:

```rust
// crates/cgka-engine/src/distributed_convergence.rs:190-221
let projection = project_mls_message(&message.payload)?;
let Some(source_epoch) = projection.source_epoch else { return Ok(()); };
// ... no check that source_epoch is anywhere near the current tip ...
self.storage.put_message(&MessageRecord {
    id: content_id,
    group_id: group_id.clone(),
    epoch: EpochId(source_epoch),          // attacker-chosen, e.g. 2^63
    state: MessageState::Created,
    payload: /* ... */,
})?;
```

The ingest gate routes any commit with `msg_epoch >= current_epoch` (and any application message with `msg_epoch > current_epoch`) into this buffer **before** any signature/confirmation-tag validation:

```rust
// crates/cgka-engine/src/message_processor/ingest.rs:507-509
let commit_should_enter_convergence = if msg_content_type == ContentType::Commit {
    if msg_epoch >= current_epoch { true } else { /* bounded past-rewind branch */ }
```

`project_mls_message` reads `protocol.epoch()` off the wire header with **no cryptographic validation** (`crates/cgka-engine/src/openmls_projection.rs:~414-433`), so the epoch is fully attacker-controlled.

2. **The orphan is never resolved.** The candidate-path BFS only extends a path where `commit.source_epoch == path.tip_epoch`, starting from the current tip:

```rust
// crates/cgka-engine/src/openmls_projection.rs:936-937
for commit in &commits {
    if commit.source_epoch != path.tip_epoch || path.digests.contains(&commit.digest) { continue; }
```

A far-future epoch is unreachable from the tip, so the forged row is never probed, never materialized, never given a disposition. The only horizon check drops the **past** side (`source_epoch < retained_anchor_epoch`, `openmls_projection.rs:678`); there is no symmetric upper-bound arm.

3. **The send gate stays stuck forever.** `has_unresolved_convergence_inputs` lists messages at/after the anchor and returns `true` for any `Created`/`Retryable` commit or application row:

```rust
// crates/cgka-engine/src/message_processor/mod.rs:381-417
let records = self.storage.list_messages(group_id, EpochId(anchor))?;
for record in records {
    // ... Created | Retryable ...
    if matches!(projection.kind, OpenMlsContentKind::Commit | OpenMlsContentKind::Application) {
        return Ok(true);   // the forged 2^63 commit matches forever
    }
}
```

`advance_convergence_inputs_until_settled` (`mod.rs:340-365`) then returns `false` (converge settles with nothing to apply but leaves the orphan `Created`), so `do_send` queues intents and `converge_and_drain_queued_outbound_intents` drains nothing.

## Failure scenario

1. Attacker is a member at epoch 5 (so they hold the group exporter secret and can wrap a kind-445 event that peels for everyone).
2. Attacker crafts a plaintext MLS message with `content_type = Commit`, `epoch = 2^63`, wraps it at the group's real current epoch, and broadcasts it.
3. Every recipient buffers it as a durable `Created` convergence row at epoch `2^63`.
4. From then on, each victim's `do_send` sees `has_unresolved_convergence_inputs == true`, queues the intent, and never sends. The victim can no longer send app messages, invite, remove, leave, or update the group.
5. Restarting does not help (the row is durable); advancing the epoch does not help (`anchor = epoch − max_rewind` is always far below `2^63`). An application message with a forged far-future epoch produces the same permanent block via the `Retryable`/`UndecryptableInCanonicalState` path — the mdk#144 "wait for the commit that advances the epoch" behavior never terminates because that commit will never exist.

This is a whole-group DoS from a single insider, and the forged branch cannot be materialized into canonical state (candidate branches go through real OpenMLS `process_message`), so the impact is denial-of-service rather than state corruption — hence HIGH rather than CRITICAL.

## Suggested fix

Bound the future side the way the past side is already bounded. In `buffer_openmls_convergence_message` (and/or the ingest buffering gate), reject or terminally classify (`Failed`) any commit/app input whose `source_epoch` exceeds `current_tip + max_rewind_commits` — a legitimate future input can only be a small, contiguous number of epochs ahead of the tip. Give the horizon check at `openmls_projection.rs:678` a symmetric upper-bound arm, and/or age out convergence inputs that survive N passes without ever materializing.

## Related hardening (same gate)

`has_unresolved_convergence_inputs` also **fails closed to "unresolved" on any decode/projection error** (`crates/cgka-engine/src/message_processor/mod.rs:389-397`): a `Created`/`Retryable` row that fails `StoredMessagePayload::decode`, is not an openmls-wire payload, or fails `project_mls_message` makes the whole group report "unresolved" indefinitely, permanently gating sends with no recovery. Since convergence never assigns such a row a terminal disposition, the block is permanent. Consider treating a decode/projection failure here as "not a gating input" while separately quarantining/failing the offending row.

---
_Filed from a meticulous automated code-review pass; verified against source and deduplicated against existing open/closed issues (distinct from #706, #144, #363, #711, #734)._

### LINKED ISSUE #745: cgka-engine: emit_superseded_processed_commits rescans and re-decodes the group's entire message table on every convergence apply
**Severity:** LOW
**Component:** `crates/cgka-engine`

## Summary

Every convergence pass that selects a tip re-lists and re-decodes **all** stored messages for the group from `EpochId(0)`, decoding + projecting each to find previously-`Processed` commits at/above the fork epoch. On an active, long-lived group (agents, the `wn-agent` daemon) with a large retained history this is O(total messages) work per apply, repeated up to `MAX_CONVERGENCE_REPROCESSING_PASSES` (16) times per drain.

## Location

`crates/cgka-engine/src/distributed_convergence.rs:907-980` — `self.storage.list_messages(group_id, EpochId(0))` followed by per-row decode/project.

## Details

The function already ignores `record.epoch.0 < fork_epoch`, so scoping the storage query to `EpochId(fork_epoch)` (as `has_unresolved_convergence_inputs` scopes to the anchor) would skip decoding the vast majority of rows while producing an identical result set.

## Suggested fix

Pass `EpochId(fork_epoch)` (not `EpochId(0)`) to `list_messages` so the scan and per-row decode/project are bounded to the epochs that can actually contribute.

---

erskingardner:
## Summary

PR-A of a two-PR sweep closing the six `cgka-engine:` issues. This one covers the **convergence send-gate** cluster; the admin/identity-seam cluster (#728, #737, #746, #747) ships separately.

### #736 — unbounded future-epoch convergence input permanently gates all sends (HIGH, member-inflicted durable DoS)

The convergence send-gate (`has_unresolved_convergence_inputs`) bounded only the **past** side of the horizon (`>= anchor`) with **no future ceiling**, and **failed closed** on any decode/projection error. A single member could broadcast one crafted plaintext message with an attacker-chosen far-future epoch (e.g. `2^63`); every recipient buffered it as a durable `Created` row that convergence can never materialize (the candidate-path BFS only extends `source_epoch == tip_epoch`) and never assigns a terminal disposition. The gate then reported the group unsettled **forever**, so `do_send` queued every intent and the queue never drained — durable across restart, no recovery short of manual DB surgery.

**Fix** (symmetric to the existing past bound, reusing `max_rewind_commits`, `message_processor/mod.rs`):
- Add the missing **future ceiling**: a row whose projected `source_epoch > tip + max_rewind_commits` is not resolvable convergence work and does not gate. The row is **left in storage** (not dropped), so it gates again correctly once the tip advances into `[anchor, ceiling]`.
- **Fail open, not closed**: a row that fails decode / is not openmls-wire / fails projection no longer counts as unresolved work.

**Deliberately not added** — a terminal future-drop arm in the projection scan (the original plan's step 2). The candidate BFS legitimately chains a *contiguous* forward run of commits with no upper bound, and ingest buffers future-epoch inputs unbounded, so terminally dropping beyond-ceiling commits would discard data a lagging/out-of-order client still needs — a real convergence regression, unlike the past side which drops only already-applied history. The gate ceiling fixes the send-DoS without that risk.

**Follow-up (separate, lower severity):** unbounded future-epoch buffering at ingest is a storage/CPU-growth vector; bounding it safely needs transport re-delivery guarantees. I'll file it as its own issue.

### #745 — `emit_superseded_processed_commits` rescans the whole message table (LOW perf)

Listed from `EpochId(0)` and decoded/projected every row only to skip those below `fork_epoch`. Scoped the storage query to `EpochId(fork_epoch)` (`list_messages`' second arg is `at_or_after_epoch`) for an identical result set with bounded work.

## Tests

- `far_future_convergence_input_beyond_ceiling_does_not_gate_sends` — forged far-future input no longer queues sends; row stays retained.
- `undecodable_convergence_row_does_not_gate_sends` — garbage `Created` row fails open.
- Regressions green: `future_epoch_app_message_stays_retryable_until_commit_arrives`, `engine_queues_app_send_until_convergence_is_settled` (near-future within window still gates), `convergence_rollback_emits_commit_rolled_back_for_losing_branch`.

`just fast-ci` + `cargo test -p cgka-engine` pass.

Closes #736, closes #745.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/752">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where outgoing messages could be blocked by unusable convergence data, allowing sends to continue normally.
  * Improved handling of malformed or out-of-range stored records so they no longer stall messaging.
  * Reduced unnecessary convergence scanning, which may improve responsiveness in affected cases.

* **Chores**
  * Updated release notes to reflect the messaging reliability fix.

### DIFF
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -917,9 +917,13 @@ impl<S: StorageProvider> Engine<S> {
         };
         let accepted: std::collections::BTreeSet<&str> =
             result.accepted_commits.iter().map(String::as_str).collect();
+        // Scope the scan to `fork_epoch` onward (mdk#745): rows below the fork
+        // epoch are skipped anyway, so listing from `EpochId(0)` re-decoded and
+        // re-projected the group's entire retained history on every apply. The
+        // storage query filters `at_or_after_epoch`, yielding an identical set.
         let records = self
             .storage
-            .list_messages(group_id, EpochId(0))
+            .list_messages(group_id, EpochId(fork_epoch))
             .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?;
         for record in records {
             if record.state != MessageState::Processed || record.epoch.0 < fork_epoch {
--- a/crates/cgka-engine/src/message_processor/mod.rs
+++ b/crates/cgka-engine/src/message_processor/mod.rs
@@ -16,7 +16,7 @@ pub(crate) use ingest::avatar_component_snapshot;
 pub(crate) use send::merge_capabilities;
 
 use crate::engine::{Engine, ScheduledSelfRemoveAutoCommit};
-use crate::openmls_projection::{OpenMlsContentKind, project_mls_message};
+use crate::openmls_projection::{OpenMlsContentKind, decode_openmls_wire_projection};
 use cgka_traits::engine::{GroupEvent, GroupStateChange, SendIntent, SendResult};
 use cgka_traits::engine_state::EpochState;
 use cgka_traits::error::EngineError;
@@ -392,6 +392,7 @@ impl<S: StorageProvider> Engine<S> {
             Err(e) => return Err(EngineError::Storage(e)),
         };
         let records = self.storage.list_messages(group_id, EpochId(anchor))?;
+        let mut skipped_non_resolvable: usize = 0;
         for record in records {
             if !matches!(
                 record.state,
@@ -405,14 +406,12 @@ impl<S: StorageProvider> Engine<S> {
             // corrupt/garbage row permanently gate sends with no recovery path.
             // Such a row simply does not contribute to the send gate; the
             // convergence horizon (`openmls_projection`) is responsible for
-            // assigning it a terminal disposition.
-            let Ok(stored_payload) = StoredMessagePayload::decode(&record.payload) else {
-                continue;
-            };
-            let Some(message) = stored_payload.as_openmls_wire() else {
-                continue;
-            };
-            let Ok(projection) = project_mls_message(&message.payload) else {
+            // assigning it a terminal disposition. Count the skips so sustained
+            // abuse (an insider spraying undecodable rows) is visible in audits
+            // without spamming a log line per row (mdk#752 review).
+            let Some((_message, projection)) = decode_openmls_wire_projection(&record.payload)
+            else {
+                skipped_non_resolvable += 1;
                 continue;
             };
             // Beyond the future horizon: unreachable from the current tip, so
@@ -440,6 +439,18 @@ impl<S: StorageProvider> Engine<S> {
             ) {
                 return Ok(true);
             }
+        }
+        // Reached only when nothing gates: surface any fail-open skips so an
+        // insider spraying undecodable rows (which no longer wedges the gate but
+        // still costs a scan) is visible in forensic audits. Aggregate count
+        // only — no ids, epochs, or payloads (observability privacy invariant).
+        if skipped_non_resolvable > 0 {
+            tracing::debug!(
+                target: "cgka_engine::message_processor",
+                method = "has_unresolved_convergence_inputs",
+                skipped_non_resolvable,
+                "skipped non-resolvable convergence rows (fail-open)"
+            );
         }
         Ok(false)
     }
--- a/crates/cgka-engine/src/message_processor/mod.rs
+++ b/crates/cgka-engine/src/message_processor/mod.rs
@@ -365,15 +365,28 @@ impl<S: StorageProvider> Engine<S> {
     }
 
     fn has_unresolved_convergence_inputs(&self, group_id: &GroupId) -> Result<bool, EngineError> {
-        let anchor = match self.storage.get_group(group_id) {
+        // The convergence horizon is bounded on BOTH sides (mdk#736). The past
+        // side (`anchor`) drops inputs older than the retained-anchor window.
+        // The future side (`ceiling`) is symmetric: a convergence input more
+        // than `max_rewind_commits` epochs ahead of the current tip cannot chain
+        // from the tip yet (the candidate-path BFS in `openmls_projection` only
+        // extends `source_epoch == tip_epoch`), so it is not *resolvable
+        // convergence work* and must not gate outbound sends. Without the
+        // ceiling, a single member could forge one far-future-epoch (e.g. 2^63)
+        // plaintext message whose buffered `Created` row is never materialized
+        // and never given a terminal disposition, permanently gating every send
+        // for the whole group. The row is left in storage (not dropped here), so
+        // it gates again correctly once the tip advances into `[anchor, ceiling]`.
+        let (anchor, ceiling) = match self.storage.get_group(group_id) {
             Ok(group) => {
                 let policy = self
                     .convergence_policy_for_group(group_id)
                     .map_err(|e| EngineError::Backend(format!("load convergence policy: {e}")))?;
-                group
-                    .epoch
-                    .0
-                    .saturating_sub(policy.convergence.max_rewind_commits)
+                let rewind = policy.convergence.max_rewind_commits;
+                (
+                    group.epoch.0.saturating_sub(rewind),
+                    group.epoch.0.saturating_add(rewind),
+                )
             }
             Err(StorageError::NotFound) => return Ok(false),
             Err(e) => return Err(EngineError::Storage(e)),
@@ -386,15 +399,27 @@ impl<S: StorageProvider> Engine<S> {
             ) {
                 continue;
             }
+            // Fail OPEN, not closed (mdk#736): a row we cannot decode, is not an
+            // openmls-wire payload, or cannot be projected is NOT resolvable
+            // convergence work — treating it as "unresolved" would let a single
+            // corrupt/garbage row permanently gate sends with no recovery path.
+            // Such a row simply does not contribute to the send gate; the
+            // convergence horizon (`openmls_projection`) is responsible for
+            // assigning it a terminal disposition.
             let Ok(stored_payload) = StoredMessagePayload::decode(&record.payload) else {
-                return Ok(true);
+                continue;
             };
             let Some(message) = stored_payload.as_openmls_wire() else {
-                return Ok(true);
+                continue;
             };
             let Ok(projection) = project_mls_message(&message.payload) else {
-                return Ok(true);
+                continue;
             };
+            // Beyond the future horizon: unreachable from the current tip, so
+            // not gating work (see the ceiling rationale above).
+            if projection.source_epoch.is_some_and(|epoch| epoch > ceiling) {
+                continue;
+            }
             // A lone uncommitted Proposal does NOT make canonical state
             // ambiguous: commits are the consensus log; a proposal only takes
             // effect once a commit consumes it (convergence.md:7-8). The
--- a/crates/cgka-engine/tests/distributed_convergence.rs
+++ b/crates/cgka-engine/tests/distributed_convergence.rs
@@ -4397,6 +4397,17 @@ async fn far_future_convergence_input_beyond_ceiling_does_not_gate_sends() {
         .buffer_openmls_convergence_message(&group_id, far_future_msg.clone(), 1_000)
         .expect("far-future message buffered");
 
+    // Convergence is not perpetually unsettled on account of the beyond-ceiling
+    // row — this distinguishes "send gate fixed" from "convergence loop still
+    // wedged on the same forged input", which was the original failure mode.
+    assert!(
+        carol
+            .advance_convergence_inputs_until_settled(&group_id, 1_000_000)
+            .await
+            .unwrap(),
+        "beyond-ceiling row must not leave convergence perpetually unsettled"
+    );
+
     // The fix: the beyond-ceiling row does not gate, so Carol can still send.
     // Pre-fix this returned `SendResult::Queued` forever.
     let sent = carol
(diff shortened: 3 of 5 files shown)


crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -386,15 +399,27 @@ impl<S: StorageProvider> Engine<S> {
...
-                return Ok(true);
+                continue;
             };
             let Ok(projection) = project_mls_message(&message.payload) else {
-                return Ok(true);
+                continue;
             };
+            // Beyond the future horizon: unreachable from the current tip, so
+            // not gating work (see the ceiling rationale above).
+            if projection.source_epoch.is_some_and(|epoch| epoch > ceiling) {
```
erskingardner, line 419: **Residual perf DoS (non-blocking, follow-up):** The ceiling skip is correct for *gating semantics*, but `list_messages(group_id, EpochId(anchor))` still returns every `Created`/`Retryable` row from anchor onward — including arbitrarily many beyond-ceiling rows an insider can buffer. Each `do_send` / `advance_convergence_inputs_until_settled` pass decodes and projects them before skipping.

Send availability is fixed; CPU/storage growth is not. When ingest bounding lands, consider either an `epoch <= ceiling` storage filter for this scan or a dedicated index of gate-relevant rows so gate work stays proportional to the horizon window, not attacker-chosen backlog depth.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -386,15 +399,27 @@ impl<S: StorageProvider> Engine<S> {
...
+            // Fail OPEN, not closed (mdk#736): a row we cannot decode, is not an
+            // openmls-wire payload, or cannot be projected is NOT resolvable
+            // convergence work — treating it as "unresolved" would let a single
+            // corrupt/garbage row permanently gate sends with no recovery path.
+            // Such a row simply does not contribute to the send gate; the
+            // convergence horizon (`openmls_projection`) is responsible for
+            // assigning it a terminal disposition.
             let Ok(stored_payload) = StoredMessagePayload::decode(&record.payload) else {
-                return Ok(true);
+                continue;
```
erskingardner: **Fail-open observability (suggestion):** These three `continue` arms correctly prevent a single corrupt row from wedging the group. They are also completely silent.

An insider spraying valid-looking `Created` rows that fail decode/projection (or wrapping non-MLS bytes in a decodable envelope) would be invisible to operators while still paying the scan cost above.

A privacy-safe aggregate counter or rare `tracing::debug!` (e.g. `skipped_non_resolvable_rows += 1` per pass, no ids/payloads) would preserve fail-open semantics while making sustained abuse visible in forensic audits.

crates/cgka-engine/tests/distributed_convergence.rs
```diff
@@ -4322,6 +4322,177 @@ async fn engine_queues_app_send_until_convergence_is_settled() {
...
+    // (correctly) once the tip advances into `[anchor, ceiling]`.
+    assert_message_state(&carol_storage, &far_future_msg, MessageState::Created);
+}
+
+/// mdk#736 (related hardening): a `Created`/`Retryable` convergence row that
+/// cannot be decoded / is not an openmls-wire payload / fails projection is NOT
+/// resolvable convergence work and must fail OPEN (not gate sends). Before the
+/// fix, `has_unresolved_convergence_inputs` returned `true` on any such row,
+/// permanently gating sends with no recovery path.
+#[tokio::test]
```
erskingardner, line 4441: **Test coverage gap (minor):** This test exercises only the `StoredMessagePayload::decode` failure path. The production gate has two additional fail-open branches (`as_openmls_wire() == None`, `project_mls_message` error).

An adversary is more likely to craft a *decodable* stored envelope with non-MLS or malformed inner wire bytes than raw garbage. A sibling test (or table-driven variant) for those branches would lock in the full fail-open contract and prevent a partial revert.

crates/cgka-engine/tests/distributed_convergence.rs
```diff
@@ -4322,6 +4322,177 @@ async fn engine_queues_app_send_until_convergence_is_settled() {
...
+        .send(SendIntent::AppMessage {
+            group_id: group_id.clone(),
+            payload: app_payload_for(&carol, b"still able to send"),
+        })
+        .await
+        .unwrap();
+    assert!(
+        matches!(sent, SendResult::ApplicationMessage { .. }),
+        "beyond-ceiling convergence input must not gate the send, got {sent:?}"
+    );
```
erskingardner, line 4423: **Optional hardening:** You already assert immediate `send` succeeds and the row stays `Created`. Consider also asserting `advance_convergence_inputs_until_settled` does not report perpetual unsettled solely because of this beyond-ceiling row — that would distinguish "send gate fixed" from "convergence loop still wedged on the same input", which was part of the original failure mode.

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -917,9 +917,13 @@ impl<S: StorageProvider> Engine<S> {
...
         let accepted: std::collections::BTreeSet<&str> =
             result.accepted_commits.iter().map(String::as_str).collect();
+        // Scope the scan to `fork_epoch` onward (mdk#745): rows below the fork
+        // epoch are skipped anyway, so listing from `EpochId(0)` re-decoded and
+        // re-projected the group's entire retained history on every apply. The
+        // storage query filters `at_or_after_epoch`, yielding an identical set.
         let records = self
             .storage
-            .list_messages(group_id, EpochId(0))
+            .list_messages(group_id, EpochId(fork_epoch))
```
erskingardner, line 926: **#745 looks correct.** Scoping `list_messages` to `EpochId(fork_epoch)` matches the existing loop filter (`record.epoch.0 < fork_epoch` → skip) and should yield an identical supersession set with strictly less decode/projection work. No adversarial correctness concern here — the redundant epoch check in the loop is now defensive belt-and-suspenders, which is fine.

erskingardner (commented):
## Adversarial review (PR #752)

Overall this is a **sound fix** for the member-inflicted send-gate DoS (#736). The future ceiling mirrors the existing past bound without the convergence regression risk the PR body calls out (terminally dropping beyond-ceiling rows in the projection scan). Fail-open on decode/projection errors is the right trade for a gate whose old fail-closed behavior was itself the exploit. #745 is a straightforward, correctness-preserving perf win.

**Verdict:** Approve with non-blocking follow-ups noted inline.

### Attack surface closed
- Single far-future `Created`/`Retryable` row can no longer wedge `has_unresolved_convergence_inputs` forever.
- Garbage/corrupt rows no longer fail-closed the entire group.

### Residual adversarial vectors (acknowledged / follow-up)
- Unbounded ingest of beyond-ceiling rows still grows storage and makes every send-gate check O(n) over `Created`/`Retryable` rows — send availability is restored, resource exhaustion is not.
- Fail-open skips are silent; operators get no signal if an insider is spraying undecodable rows (availability preserved, observability gap).

### Tests
- `far_future_convergence_input_beyond_ceiling_does_not_gate_sends` is a faithful reproduction of the original attack.
- `undecodable_convergence_row_does_not_gate_sends` covers one of three fail-open branches; see inline suggestion for the other two.
- Existing near-future gating regressions (`engine_queues_app_send_until_convergence_is_settled`, `future_epoch_app_message_stays_retryable_until_commit_arrives`) appropriately stay green.

erskingardner:
Thanks — non-blocking follow-ups addressed in 735b7d4.

- **Fail-open helper (CodeRabbit):** extracted the `decode -> as_openmls_wire -> project` chain into `openmls_projection::decode_openmls_wire_projection` and used it in the send gate. Note the other two sites deliberately differ: `emit_superseded_processed_commits` and `buffer_*` **propagate** decode errors (a corrupt `Processed`/inbound row there is a real integrity signal), so they intentionally keep their own `?` chain rather than fail open — the helper's doc calls this out.
- **Fail-open observability:** the three skips were silent. The gate now counts non-resolvable rows per scan and emits one privacy-safe aggregate `tracing::debug!` (count only, no ids/epochs/payloads) on the not-gated path, so an insider spraying undecodable rows shows up in audits without a per-row log line.
- **Test coverage of the other two branches:** added `non_wire_and_unprojectable_convergence_rows_do_not_gate_sends` (a decodable non-openmls-wire row + an openmls-wire row whose inner bytes don't project).
- **Convergence-not-wedged:** the far-future test now also asserts `advance_convergence_inputs_until_settled` returns settled, distinguishing "send gate fixed" from "convergence loop still wedged".

On the **residual perf DoS** (beyond-ceiling rows still scanned/decoded per pass) — agreed, out of scope here; send availability is restored but resource growth isn't. It's the same root as the unbounded-ingest-buffering follow-up noted in the PR description; I'll file that as its own issue rather than fold an unbounded-buffering bound (which needs transport re-delivery guarantees) into this fix.

PR #753, opened 2026-07-05
cgka-engine: close the admin/identity seam-parity gaps (#728, #737, #746, #747)

### LINKED ISSUE #728: cgka-engine: third sender resolver in app_components.rs bypasses validate_credential_identity

### LINKED ISSUE #737: cgka-engine: group creation and welcome-join skip the admin-leaf-coupling check enforced on every commit seam (phantom/pre-provisioned admins)

### LINKED ISSUE #746: cgka-engine: invitee KeyPackage omitting the admin-policy component silently negotiates it out of the created group (empty admin set, frozen membership)

### LINKED ISSUE #747: cgka-engine: directory-facing KeyPackage helpers validate the account-identity proof against DEFAULT_CIPHERSUITE instead of the KeyPackage's own ciphersuite

erskingardner:
## Summary

PR-B of a two-PR sweep closing the six `cgka-engine:` issues (PR-A [#752](https://github.com/marmot-protocol/mdk/pull/752) covered the convergence send-gate). These four are **one meta-defect** — an invariant enforced at some seams but not all of them, the crate's own *"a guard that exists on one seam only is a bug"* rule (mdk#707). Where cheap, the fix collapses duplicated logic onto a single chokepoint so the next seam can't drift.

### #728 — third sender resolver bypassed identity validation
`app_components.rs` had a private `member_id_of_sender` that built a `MemberId` via raw `BasicCredential::try_from`, skipping the `validate_credential_identity` (x-only secp256k1) check the shared `identity::member_id_of_sender` runs on the direct-ingest and stored-convergence seams. **Deleted the copy**; the admin-self-remove policy path now resolves through the validating chokepoint. A malformed-identity sender resolves to `None` and is skipped — safe, since a malformed identity can never be in the validated admin set it would be compared against.

### #737 (MEDIUM) — creation & welcome-join skipped admin-leaf coupling
Every commit seam runs `validate_admin_leaf_coupling_*`, but `do_create_group` and `do_join_welcome` didn't — letting a group be created with a phantom/pre-provisioned admin (`initial_admins` is independent of `members`) that activates the instant a matching leaf appears, with no `AdminAdded` event other members observe. Extracted the set-membership core into **`reject_admins_without_member_accounts`** (one validator, three callers). Creation now x-only-validates each `initial_admins` entry and runs the coupling check against the projected creator+invitee accounts *before* any membership side effects; welcome-join runs it against the joined group.

### #746 — mandatory components negotiable out at creation
Profile + admin-policy lived in `desired_components` but not `required_caps`, so the per-invitee intersection let an invitee whose KeyPackage omitted them negotiate them out — creating a group with an **empty admin set and permanently frozen membership**. Treat `default_group_components()` as non-negotiable: reject an invitee lacking them up front (like a missing required capability), plus a post-negotiation assert.

### #747 (LOW/latent) — proof validated against the wrong ciphersuite
`key_package_metadata` / `is_last_resort_key_package` validated the account-identity proof against a hardcoded `DEFAULT_CIPHERSUITE`, while `parse_key_package` uses the KeyPackage's own ciphersuite — they'd disagree the moment a second ciphersuite exists. Derive it from the KeyPackage in both helpers.

## Tests

New: `create_group_rejects_uninvited_initial_admin` (+ `create_group_accepts_invited_initial_admin` control), `create_group_rejects_invitee_missing_mandatory_component`, `directory_key_package_helpers_validate_against_own_ciphersuite`.

Updated two existing tests that **encoded the #746 bug** (asserting admin-policy gets silently negotiated out) to assert the fixed non-negotiable behavior. Existing admin-self-remove tests cover the #728 valid-path parity.

`just fast-ci` + `cargo test -p cgka-engine` pass.

Closes #728, closes #737, closes #746, closes #747.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/753">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/crates/cgka-engine/src/app_components.rs
+++ b/crates/cgka-engine/src/app_components.rs
@@ -436,6 +436,18 @@ pub(crate) fn member_account_pubkeys(mls_group: &MlsGroup) -> BTreeSet<[u8; 32]>
 /// resolves it from the joined group. Keeping them on one function is why a
 /// phantom/pre-provisioned admin can no longer slip in at a seam that "forgot"
 /// the check (mdk#737).
+///
+/// Note the callers build `member_accounts` on two different bases, and that is
+/// intentional. Creation uses `member_id_of_key_package` /
+/// `validated_member_id_of_leaf` (secp256k1-validated) because it is the
+/// authoring path and controls exactly which invitees it admits. The
+/// `MlsGroup`-backed callers use [`member_account_pubkeys`] /
+/// `credential_account_pubkey` (32-byte length only) because every leaf already
+/// present in the tree passed credential validation at its own ingress seam
+/// (welcome-join step 5b `validate_member_credentials_and_account_proofs`, and
+/// add/commit ingest), so the tree cannot carry an unvalidated member leaf here.
+/// The admin set (`decode_admin_policy`) is likewise length-only, so both sides
+/// of the coupling comparison share a basis.
 pub(crate) fn reject_admins_without_member_accounts(
     admins: &[[u8; 32]],
     member_accounts: &BTreeSet<[u8; 32]>,
@@ -446,7 +458,7 @@ pub(crate) fn reject_admins_without_member_accounts(
     }
     if admins.iter().any(|admin| !member_accounts.contains(admin)) {
         return Err(EngineError::Other(format!(
-            "admin-policy update is invalid: an admin key has no member leaf (group {group_id:?})"
+            "admin key has no member leaf (group {group_id:?})"
         )));
     }
     Ok(())
@@ -465,17 +477,28 @@ fn reject_admin_self_remove_proposals(
         if !matches!(queued.proposal(), Proposal::SelfRemove) {
             continue;
         }
-        // Resolve the sender through the shared, validating chokepoint
-        // (mdk#728): `identity::member_id_of_sender` runs
-        // `validate_credential_identity` (x-only secp256k1) before trusting a
-        // leaf's identity as a `MemberId`, exactly as the direct-ingest and
-        // stored-convergence seams do. A malformed identity resolves to `None`
-        // and is skipped — safe, because a malformed identity can never appear
-        // in the validated admin set it would be compared against.
-        let Some(sender) = crate::identity::member_id_of_sender(queued.sender(), mls_group) else {
-            continue;
+        // Resolve the sender's account pubkey on the SAME (length-based) basis
+        // the admin set is decoded on (`decode_admin_policy` /
+        // `credential_account_pubkey`), NOT the secp256k1-validating
+        // `identity::member_id_of_sender` (mdk#728 review). The admin set is not
+        // curve-validated, so a leaf whose 32-byte identity equals a listed
+        // admin key but fails secp256k1 validation would resolve to `None` under
+        // the validating chokepoint and SKIP this guard — letting that admin
+        // self-remove in violation of MIP-03. Matching the admin-set basis keeps
+        // both sides of the comparison comparable. Fail CLOSED: a `SelfRemove`
+        // whose sender cannot be resolved to a member account is rejected, never
+        // waved through (mirrors the Sm7 auto-committer / fork-recovery posture).
+        let sender_pubkey = match queued.sender() {
+            Sender::Member(leaf_idx) => mls_group
+                .member_at(*leaf_idx)
+                .and_then(|member| credential_account_pubkey(member.credential)),
+            _ => None,
         };
-        let sender_pubkey = admin_pubkey_from_member_id(&sender)?;
+        let Some(sender_pubkey) = sender_pubkey else {
+            return Err(EngineError::AdminCannotSelfRemove {
+                group_id: group_id.clone(),
+            });
+        };
         if admins.iter().any(|admin| admin == &sender_pubkey) {
             return Err(EngineError::AdminCannotSelfRemove {
                 group_id: group_id.clone(),
--- a/crates/cgka-engine/src/app_components.rs
+++ b/crates/cgka-engine/src/app_components.rs
@@ -412,16 +412,39 @@ pub(crate) fn reject_admins_without_member_leaf(
     group_id: &GroupId,
     admins: &[[u8; 32]],
 ) -> Result<(), EngineError> {
-    if admins.is_empty() {
-        return Ok(());
-    }
+    reject_admins_without_member_accounts(admins, &member_account_pubkeys(mls_group), group_id)
+}
+
+/// Member account pubkeys currently in the group's ratchet tree.
+pub(crate) fn member_account_pubkeys(mls_group: &MlsGroup) -> BTreeSet<[u8; 32]> {
     let mut accounts: BTreeSet<[u8; 32]> = BTreeSet::new();
     for member in mls_group.members() {
         if let Some(pk) = credential_account_pubkey(member.credential) {
             accounts.insert(pk);
         }
     }
-    if admins.iter().any(|admin| !accounts.contains(admin)) {
+    accounts
+}
+
+/// Core admin-leaf-coupling check (admin-policy-v1.md "Validation") over an
+/// explicit set of member account pubkeys: every admin key MUST correspond to a
+/// member account. This is the single validator shared by every seam that
+/// establishes or mutates a group's admin set — outbound component updates and
+/// commit apply resolve the set from the live `MlsGroup`
+/// ([`reject_admins_without_member_leaf`]); group creation resolves it from the
+/// projected creator+invitee set (no `MlsGroup` exists yet); welcome-join
+/// resolves it from the joined group. Keeping them on one function is why a
+/// phantom/pre-provisioned admin can no longer slip in at a seam that "forgot"
+/// the check (mdk#737).
+pub(crate) fn reject_admins_without_member_accounts(
+    admins: &[[u8; 32]],
+    member_accounts: &BTreeSet<[u8; 32]>,
+    group_id: &GroupId,
+) -> Result<(), EngineError> {
+    if admins.is_empty() {
+        return Ok(());
+    }
+    if admins.iter().any(|admin| !member_accounts.contains(admin)) {
         return Err(EngineError::Other(format!(
             "admin-policy update is invalid: an admin key has no member leaf (group {group_id:?})"
         )));
@@ -442,7 +465,14 @@ fn reject_admin_self_remove_proposals(
         if !matches!(queued.proposal(), Proposal::SelfRemove) {
             continue;
         }
-        let Some(sender) = member_id_of_sender(queued.sender(), mls_group) else {
+        // Resolve the sender through the shared, validating chokepoint
+        // (mdk#728): `identity::member_id_of_sender` runs
+        // `validate_credential_identity` (x-only secp256k1) before trusting a
+        // leaf's identity as a `MemberId`, exactly as the direct-ingest and
+        // stored-convergence seams do. A malformed identity resolves to `None`
+        // and is skipped — safe, because a malformed identity can never appear
+        // in the validated admin set it would be compared against.
+        let Some(sender) = crate::identity::member_id_of_sender(queued.sender(), mls_group) else {
             continue;
         };
         let sender_pubkey = admin_pubkey_from_member_id(&sender)?;
@@ -453,17 +483,6 @@ fn reject_admin_self_remove_proposals(
         }
     }
     Ok(())
-}
-
-fn member_id_of_sender(sender: &Sender, group: &MlsGroup) -> Option<MemberId> {
-    match sender {
-        Sender::Member(leaf_idx) => {
-            let member = group.member_at(*leaf_idx)?;
-            let basic = BasicCredential::try_from(member.credential).ok()?;
-            Some(MemberId::new(basic.identity().to_vec()))
-        }
-        _ => None,
-    }
 }
 
 /// Does this staged commit require the sender to be a group admin?
--- a/crates/cgka-engine/src/group_lifecycle.rs
+++ b/crates/cgka-engine/src/group_lifecycle.rs
@@ -142,10 +142,12 @@ impl<S: StorageProvider> Engine<S> {
             parsed_kps.push(parsed);
         }
         required_caps.app_components = negotiated_components;
-        // Belt-and-suspenders (mdk#746): the engine-owned components survived
-        // negotiation. Guaranteed by the per-invitee guard above plus the
-        // engine's own support; the assert catches a future refactor of the
-        // negotiation flow that reintroduces the drop.
+        // Invariant check (mdk#746): the engine-owned components survived
+        // negotiation. The per-invitee guard above is the real runtime gate (it
+        // rejects any invitee lacking them in every build); this assertion just
+        // documents the post-condition and, because `cargo test` builds with
+        // debug assertions on, trips CI if a future negotiation refactor
+        // reintroduces the drop despite the guard.
         debug_assert!(
             mandatory_components
                 .missing_from(&required_caps.app_components)
@@ -163,6 +165,11 @@ impl<S: StorageProvider> Engine<S> {
         ])
         .map_err(|e| EngineError::Backend(format!("leaf extensions: {e:?}")))?;
 
+        // Validate the creator (implicit admin) on the SAME x-only secp256k1
+        // basis as the co-admins below (mdk#737 review), so no admin-set entry
+        // is accepted on length alone regardless of how the engine identity was
+        // constructed.
+        crate::identity::validate_credential_identity(self.identity.self_id().as_slice())?;
         let creator_pubkey =
             crate::app_components::admin_pubkey_from_member_id(self.identity.self_id())?;
         let mut admin_set: Vec<[u8; 32]> = vec![creator_pubkey];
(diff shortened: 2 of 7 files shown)


crates/cgka-engine/src/app_components.rs
```diff
@@ -442,7 +465,14 @@ fn reject_admin_self_remove_proposals(
...
-        let Some(sender) = member_id_of_sender(queued.sender(), mls_group) else {
+        // Resolve the sender through the shared, validating chokepoint
+        // (mdk#728): `identity::member_id_of_sender` runs
+        // `validate_credential_identity` (x-only secp256k1) before trusting a
+        // leaf's identity as a `MemberId`, exactly as the direct-ingest and
+        // stored-convergence seams do. A malformed identity resolves to `None`
+        // and is skipped — safe, because a malformed identity can never appear
+        // in the validated admin set it would be compared against.
+        let Some(sender) = crate::identity::member_id_of_sender(queued.sender(), mls_group) else {
             continue;
```
erskingardner: **Fail-open vs fail-closed asymmetry (#728)**

Skipping a `SelfRemove` when `member_id_of_sender` returns `None` is the opposite posture from Sm7/auto-committer, which *rejects* when the leaver's credential is malformed.

The comment argues a malformed identity "can never appear in the validated admin set", but `admins_of_group` → `decode_admin_policy` still accepts any sorted 32-byte keys with **no** `validate_credential_identity` pass. So the premise is only true for engine-authored creation, not for inbound Welcome/GroupContext bytes.

Concrete bypass sketch: admin set lists pubkey `K`; a member leaf carries 32 bytes equal to `K` that fail secp256k1 validation (off-curve / `[0;32]`). `credential_account_pubkey` counts them as a member for coupling; `member_id_of_sender` returns `None` here, so `AdminCannotSelfRemove` is **not** enforced and that listed admin can self-remove despite MIP-03.

Consider fail-closed: if a queued `SelfRemove` sender cannot be resolved to a validated `MemberId`, reject the staged commit (or at least reject when `credential_account_pubkey` on that leaf matches an admin key). Happy to be wrong if there's an invariant elsewhere that makes such leaves unreachable.

crates/cgka-engine/src/group_lifecycle.rs
```diff
@@ -531,6 +590,18 @@ impl<S: StorageProvider> Engine<S> {
         // admin-gated member additions (joining.md, admin-policy-v1.md).
         crate::app_components::require_admin(&mls_group, &group_id, &welcome_sender_id)?;
 
+        // 5e. Admin-leaf coupling on the joined group (mdk#737): reject a
+        // Welcome whose admin set lists an admin with no member leaf. The
+        // Welcome embeds the full ratchet tree, so every current member's leaf
+        // is present here — a phantom/pre-provisioned admin is detectable and
+        // must be rejected, mirroring the check every commit seam enforces so a
+        // joiner never silently accepts a group with a pre-provisioned admin.
+        crate::app_components::reject_admins_without_member_leaf(
```
erskingardner, line 606: **Missing regression test for the new welcome-join seam (#737)**

Creation gets `create_group_rejects_uninvited_initial_admin`, but the equally important join path added in 5e has no test. This is the path where a victim client could silently join a group that already embeds a phantom admin in signed group-context bytes.

Suggest mirroring the creation test: build a welcome (harness or low-level fixture) whose admin-policy lists a pubkey with no ratchet-tree leaf, assert `join_welcome` fails with the same admin-leaf-coupling error before any group record is persisted.

crates/cgka-engine/src/group_lifecycle.rs
```diff
@@ -137,11 +167,16 @@ impl<S: StorageProvider> Engine<S> {
             crate::app_components::admin_pubkey_from_member_id(self.identity.self_id())?;
```
erskingardner, line 174: **Validation asymmetry on the creator admin key (#737)**

`initial_admins` entries now run `validate_credential_identity`, but the implicit creator admin still only goes through `admin_pubkey_from_member_id` (length check). If the goal is "no 32-byte blobs in the admin set", the creator path is the odd one out.

Probably low risk in production (engine identity is validated at builder time), but it's inconsistent with the comment on the co-admin loop and leaves a seam if a test harness ever constructs an engine with a non-curve 32-byte `MemberId`.

crates/cgka-engine/src/app_components.rs
```diff
@@ -412,16 +412,39 @@ pub(crate) fn reject_admins_without_member_leaf(
...
-        return Ok(());
-    }
+    reject_admins_without_member_accounts(admins, &member_account_pubkeys(mls_group), group_id)
+}
+
+/// Member account pubkeys currently in the group's ratchet tree.
+pub(crate) fn member_account_pubkeys(mls_group: &MlsGroup) -> BTreeSet<[u8; 32]> {
     let mut accounts: BTreeSet<[u8; 32]> = BTreeSet::new();
     for member in mls_group.members() {
         if let Some(pk) = credential_account_pubkey(member.credential) {
```
erskingardner, line 422: **Strictness mismatch inside the new shared validator**

`reject_admins_without_member_accounts` is now the single chokepoint, but callers feed it member sets built two different ways:
- **Creation** (`group_lifecycle.rs`): projected invitees via `member_id_of_key_package` → `validated_member_id_of_leaf` (secp256k1 enforced).
- **Welcome-join / commit seams** (`member_account_pubkeys` here): `credential_account_pubkey` (32-byte length only).

That means creation is stricter than welcome-join for the *member* side of the coupling check. Worth either documenting as intentional (MLS tree may contain legacy leaves) or aligning on `validated_member_id` with a fallible loop so the chokepoint's inputs are comparable across seams.

crates/cgka-engine/src/app_components.rs
```diff
@@ -412,16 +412,39 @@ pub(crate) fn reject_admins_without_member_leaf(
...
+pub(crate) fn reject_admins_without_member_accounts(
+    admins: &[[u8; 32]],
+    member_accounts: &BTreeSet<[u8; 32]>,
+    group_id: &GroupId,
+) -> Result<(), EngineError> {
+    if admins.is_empty() {
+        return Ok(());
+    }
+    if admins.iter().any(|admin| !member_accounts.contains(admin)) {
         return Err(EngineError::Other(format!(
```
erskingardner, line 460: Nit: error text still says `"admin-policy update is invalid"` but this function is now also the creation and welcome-join failure path. Misleading when triaging client errors — consider a neutral message like `"admin key has no member leaf"` without implying an update operation.

crates/cgka-engine/src/group_lifecycle.rs
```diff
@@ -118,10 +121,37 @@ impl<S: StorageProvider> Engine<S> {
...
+            }
             negotiated_components = negotiated_components.intersection(&had.app_components);
             parsed_kps.push(parsed);
         }
         required_caps.app_components = negotiated_components;
+        // Belt-and-suspenders (mdk#746): the engine-owned components survived
+        // negotiation. Guaranteed by the per-invitee guard above plus the
+        // engine's own support; the assert catches a future refactor of the
+        // negotiation flow that reintroduces the drop.
+        debug_assert!(
```
erskingardner, line 151: The #746 belt-and-suspenders guard is a `debug_assert!`, so it evaporates in release builds. The per-invitee check above is the real gate (fine), but if this assert is meant to catch future negotiation refactors in CI, a `#[cfg(test)]` assertion or an integration test that would fail in release mode might be more honest than the comment implies.

crates/cgka-engine/tests/group_creation.rs
```diff
@@ -403,6 +403,97 @@ async fn create_group_with_three_members_happy_path() {
...
+            required_features: vec![],
+            app_components: vec![],
+            initial_admins: vec![mallory.self_id()],
+        })
+        .await
+        .expect_err("an uninvited initial admin must be rejected");
+    assert!(
+        matches!(err, EngineError::Other(ref msg) if msg.contains("no member leaf")),
+        "expected admin-leaf-coupling rejection, got {err:?}"
+    );
```
erskingardner, line 432: You added `create_group_rejects_uninvited_initial_admin` but no sibling test that `initial_admins` containing a 32-byte non-secp256k1 key (e.g. `[0xFF; 32]`) is rejected via `InvalidCredentialIdentity`. The new `validate_credential_identity` call on the co-admin path is untested.

erskingardner (commented):
## Adversarial review summary

Solid direction: collapsing admin-leaf coupling onto `reject_admins_without_member_accounts` and routing the third sender resolver through `identity::member_id_of_sender` matches the crate's seam-parity rule (#707). The #746 mandatory-component guard and #747 ciphersuite fix are clearly correct.

I found a few places where the fix is **fail-open** where neighboring seams are **fail-closed**, plus test gaps on the highest-risk new path (welcome-join). Details inline.

**Verdict:** Request changes on the fail-open SelfRemove skip and missing welcome-join regression test; the rest are consistency/nit follow-ups.

erskingardner:
Thanks for the adversarial pass — addressed in f7abadd.

**Blocking**
- **Fail-open SelfRemove skip (#728) — you're right, my change introduced the regression.** Routing the sender through the secp256k1-validating `member_id_of_sender` was wrong *for this seam* precisely because the admin set is decoded length-only, so an off-curve leaf whose bytes equal a listed admin key resolved to `None` and skipped the guard. Fixed by resolving the sender on the **same length-based basis as the admin set** (`credential_account_pubkey`) and **failing closed** when a `SelfRemove` sender can't be resolved to a member account (mirrors Sm7 / fork-recovery). Documented why this seam deliberately matches the admin-set basis. This is effectively issue #728's "document why it's deliberately looser" option — the collapse was the wrong call here.
- **Missing welcome-join test (#737).** Added `join_rejects_welcome_whose_admin_set_lists_a_phantom_admin` in `invite_leave.rs`, reusing the existing `welcome_from_fork_with_self_promoted_admin` fixture with `{alice, bob, mallory}` so it clears the 5d `require_admin` author check and reaches 5e on the phantom (mallory, valid key with no leaf). Asserts the coupling error before any group record is persisted.

**Non-blocking**
- **Creator-key symmetry (#737):** the creator identity now runs `validate_credential_identity` too, so no admin-set entry is accepted on length alone. New test `create_group_rejects_off_curve_initial_admin`.
- **Shared-validator strictness basis:** documented as intentional — creation validates invitees (authoring path); `MlsGroup`-backed callers use the length-only basis because every leaf already passed credential validation at its ingress seam, matching the length-only admin-set decode.
- **Error text:** now neutral (`"admin key has no member leaf"`).
- **#746 `debug_assert`:** clarified in the comment that it fires under `cargo test` (CI) since test builds enable debug assertions; the per-invitee guard remains the runtime gate in all builds.

PR #755, opened 2026-07-06
Add external signer account support

### LINKED ISSUE #382: Add external signer account support for Amber/NIP-55
## Summary

Add first-class external signer support to MDK so clients can create and use
accounts whose Nostr signing key lives outside the app, starting with Android
Amber / NIP-55.

## Current State

MDK is gaining the right account shape for external signer accounts:

- an account can be represented without storing a local Nostr secret
- clients can register callback-backed signing/encryption/decryption methods
- local-key accounts continue to work separately from external-signer accounts
- clients can re-register signer callbacks after app restart

The main blocker is the MLS account-identity proof.

The current proof flow asks the external signer to return a raw 64-byte BIP-340
Schnorr signature over a protocol-specific SHA-256 digest. That is not a Nostr
event. Amber / NIP-55 signers can sign canonical Nostr events, but they do not
provide a standard arbitrary raw-digest signing API.

This is not just a login edge case. The account-identity proof is on the core
group path: KeyPackage publication, group creation, joining, and leaf updates can
all require the account's Nostr key to prove the MLS leaf belongs to that
account. If this remains a raw prehash signing requirement, an Amber-backed
account would be able to start login but would not be able to reliably use
groups.

## Goal

Support a durable external-signer account mode where:

- the client stores public identity/account metadata only
- the private Nostr signing key remains in the external signer
- MDK can request required Nostr operations through client-provided callbacks
- clients can re-register callbacks after app/process restart
- account-identity proofs are compatible with standard event-signing external
  signers
- local `nsec` accounts continue working unchanged

## Required MDK Design Change

The account-identity proof should use a Nostr-event-signing pattern instead of a
raw-digest signing pattern.

Suggested shape:

1. MDK constructs a canonical, unpublished Nostr proof event from the same proof
   inputs: account Nostr pubkey, MLS leaf signature key, ciphersuite/signature
   scheme, domain/version, and any group/session context required by the proof.
2. MDK asks the external signer to sign that unsigned event through the normal
   `signEvent` callback.
3. MDK stores only the proof material needed at the MLS layer, for example the
   event signature and enough version/type data to reconstruct the unsigned
   event.
4. Verifiers reconstruct the same unsigned event from the MLS leaf/proof context
   and verify the signature against the account Nostr key.

This keeps the security property we need, while allowing Amber/NIP-55,
NIP-46-style signers, and other event-signing integrations to participate
without local key storage.

## Proposed UniFFI/Client Contract

External signer callback surface should include:

- `publicKey`
- `signEvent`
- `nip44Encrypt`
- `nip44Decrypt`
- `nip04Encrypt` / `nip04Decrypt` if still required by current runtime surfaces

The contract should avoid requiring arbitrary raw-digest signing for the normal
Amber/NIP-55 path.

MDK should also expose typed external-signer failures so clients can distinguish:

- signer unavailable / callback not registered
- signer public key mismatch
- user rejected or cancelled the signer prompt
- signer capability unsupported

## Acceptance Criteria

- MDK exposes a stable external signer account flow to clients.
- Clients can create/login an external-signer account using a public key.
- Clients can register/re-register callbacks for existing external-signer
  accounts.
- Runtime operations that require the account Nostr signer route through the
  external signer callbacks.
- Account-identity proof generation is compatible with external signers that can
  sign Nostr events but cannot sign arbitrary raw digests.
- Local `nsec` accounts continue to work unchanged.
- The contract is documented enough for Android, iOS, and desktop clients to
  implement consistently.

mubarakcoded:
## Summary

- add external signer account support for MDK accounts whose Nostr key is owned by the host client
- switch the account identity proof to an Amber-compatible canonical Nostr event signature
- route runtime signing paths through account signers instead of local secret loading where needed
- support external signer accounts across push owner proofs, notification triggers, media upload auth, relay/auth publishing, directory storage, and UniFFI
- document the client contract for Android/iOS/Linux integrations

## Notes

The account identity proof is now proof version 2 and signs a canonical, unpublished kind 450 Nostr event. This keeps the MLS leaf/account binding while allowing platform signers that only support Nostr event signing.

NIP-04 remains exposed on the external signer callback for legacy signer compatibility, but current MDK protocol paths do not require it.

## Verification

- `cargo fmt`
- `git diff --check`
- `cargo check -p marmot-app -p marmot-uniffi`
- `cargo test -p cgka-engine account_identity_proof`
- `cargo test -p marmot-app notifications`
- `cargo test -p marmot-app external_signer`
- `cargo test -p marmot-app media`
- `cargo test -p marmot-account external_signer`

Closes #382

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/755">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added host-provided external-signer accounts with login, registration, and restart continuity, including reversible sign-out behavior.
  * Account summaries now indicate whether signing is local or external.
  * External-signer signing is used for media uploads, push/gift-wrap publishing, directory publishing, and relay transport setup.
  * Added support for discovery relays during account setup, including new UniFFI bindings for external-signer login/registration.
* **Bug Fixes**
  * Upgraded account-identity proof to v2 event-based signing/verification (with stronger negative validation).
  * Added event-based push owner-proof verification with legacy fallback.
  * Improved external-signer error handling (unavailable, mismatch, rejected) and external signer signing eligibility.
* **Documentation**
  * Expanded docs for external-signer accounts, callback and error behavior.


crates/marmot-uniffi/src/external_signer.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+}
+
+impl AccountIdentityProofSigner for ExternalAccountSignerAdapter {
+    fn sign_account_identity_proof(
+        &self,
+        request: &AccountIdentityProofRequest,
+    ) -> Result<[u8; 64], String> {
+        let unsigned_event_json = request.proof_event_json()?;
+        let signed_event_json = self
```
erskingardner, line 182: I think this proof-signing path still needs one more pass before merge. The normal `NostrSigner::sign_event` FFI path above offloads the foreign callback with `spawn_blocking` and maps `ExternalSignerRejected` through `callback_signer_error`, but the account-identity proof path calls the foreign `sign_event` synchronously and collapses `MarmotKitError` into a string. That means a user-cancelled proof prompt during KeyPackage publication/group leaf work can surface as `InvalidAccountIdentityProof`/generic runtime instead of typed `ExternalSignerRejected`, and the callback may block a Tokio worker. Can we route this through the same blocking + typed-error path, or make the proof signer return a typed app/signer error instead of `String`?

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -2992,6 +3036,159 @@ impl AccountManager {
...
+        discovery_relays: Vec<TransportEndpoint>,
+    ) -> Option<AccountRelayListStatus> {
+        let status = match self
+            .app
+            .fetch_account_relay_list_status_for_account_id(account_id_hex, discovery_relays)
+            .await
+        {
+            Ok(status) => status,
+            Err(err) => {
+                tracing::debug!(
```
erskingardner, line 3180: This currently fails the tracing privacy audit (`production_tracing_calls_are_structured_and_privacy_safe`). The trace call needs an explicit `target` and `method`, and it cannot include `account_id_hex`/`account_id`; use aggregate/privacy-safe fields only, e.g. an error kind. Same issue appears in the profile preflight debug call below.

erskingardner (commented):
Left two comments I think are still worth addressing before this lands, given clean-break compatibility is okay and the near-term scope is Nostr signers.

crates/marmot-uniffi/src/external_signer.rs
```diff
@@ -0,0 +1,165 @@
...
+        payload: &'a str,
+    ) -> nostr::util::BoxedFuture<'a, Result<String, SignerError>> {
+        let signer = self.signer.clone();
+        Box::pin(async move {
+            signer
+                .nip44_decrypt(public_key.to_hex(), payload.to_owned())
+                .map_err(callback_signer_error)
+        })
+    }
+}
```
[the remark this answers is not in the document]
mubarakcoded: All NostrSigner callbacks are offloaded via tokio::task::spawn_blocking. The account-identity proof callback is synchronous but runs inside AccountDeviceSession::open under blocking_app_task, so it likewise never blocks an async worker; rejections are typed end-to-end (see the proof-path change).

crates/cgka-engine/src/account_identity_proof.rs
```diff
@@ -7,18 +7,23 @@
...
 /// Marmot custom LeafNode extension:
-/// `marmot.account-identity-proof.v1`.
+/// `marmot.account-identity-proof.v2`.
 pub const ACCOUNT_IDENTITY_PROOF_EXTENSION_TYPE: u16 = 0xF2F1;
+pub const ACCOUNT_IDENTITY_PROOF_EVENT_KIND: u16 = 450;
 
-const ACCOUNT_IDENTITY_PROOF_VERSION: u8 = 1;
-const ACCOUNT_IDENTITY_PROOF_DOMAIN: &[u8] = b"marmot.account-identity-proof.v1";
+const ACCOUNT_IDENTITY_PROOF_VERSION: u8 = 2;
+const ACCOUNT_IDENTITY_PROOF_DOMAIN: &str = "marmot.account-identity-proof.v2";
```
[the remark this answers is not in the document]
mubarakcoded: @coderabbitai this is an intentional clean break. v2 fully replaces v1 behind a version-guard discriminator, clean-break compatibility is acceptable here. There are no v1 proofs to migrate

crates/cgka-engine/src/account_identity_proof.rs
```diff
@@ -7,18 +7,23 @@
 
 use cgka_traits::error::EngineError;
 use cgka_traits::types::MemberId;
+use nostr::prelude::JsonUtil;
+use nostr::{
+    Event, EventBuilder, Kind, PublicKey, Tag, TagKind, Timestamp, UnsignedEvent,
+    secp256k1::schnorr::Signature,
+};
```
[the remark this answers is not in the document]
mubarakcoded: Valid layering concern — tracked as a follow-up (TODO added in the module). The engine should expose a neutral contract (canonical event bytes + event-id + 64-byte signature) with nostr EventBuilder/add_signature moved to the app/session adapter. Deferred for this PR because verification runs deep in the engine ingest paths (staged-commit + KeyPackage validation) and needs either a proof-message-computer callback threaded through or in-engine NIP-01 canonicalization to drop the SDK — not a clean lift, and outside this PR's must-fix scope.

crates/marmot-app/src/runtime/mod.rs
mubarakcoded: Fixed. Both preflight trace/debug calls now use explicit target/method and error_kind = err.privacy_safe_kind(), with no account_id_hex/raw identifiers. The forensics structured-tracing privacy audit passes.

crates/marmot-uniffi/src/external_signer.rs
mubarakcoded: Done. The proof callback now maps through callback_proof_signer_error, which carries a cancelled prompt as the EXTERNAL_SIGNER_REJECTED sentinel, external_signer_session_error recovers it from AccountDeviceSession::open's SessionError into AppError::ExternalSignerRejected, so a user-cancelled proof surfaces as the typed rejection instead of InvalidAccountIdentityProof. On blocking: the proof is signed inside AccountDeviceSession::open, which runs under blocking_app_task, so the synchronous foreign call executes on a blocking-pool thread and never parks a Tokio worker

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -2992,6 +3041,165 @@ impl AccountManager {
...
+        self.shared.lifecycle().ensure_running()?;
+        let signer_public_key = signer
+            .get_public_key()
+            .await
+            .map_err(crate::external_signer_public_key_error)?;
+        let account_id_hex = AccountHome::account_id_for_public_key(&public_key)?;
+        if signer_public_key.to_hex() != account_id_hex {
+            return Err(AppError::ExternalSignerMismatch);
+        }
+        let created_account = self.app.account_home().account(&account_id_hex).is_err();
```
erskingardner: I think this still leaves a torn account for the public/tracked-account upgrade path. `created_account` is false whenever a public account already exists, but `add_external_signer_account()` below mutates that record to `external_signing = true`. If relay-list setup or initial KeyPackage publication then fails, the error path skips rollback because `created_account == false`, leaving the original public account durably converted into an active external-signing account even though login failed and setup is incomplete. This is easy to hit with an existing tracked account plus missing/failed default relays. Can we distinguish "already-existing external account re-login" from "public account upgraded in this call" and either revert the flag or rollback the upgrade on setup failure?

crates/marmot-app/src/notifications.rs
```diff
@@ -540,30 +544,106 @@ fn push_digest(bytes: &[u8]) -> [u8; 32] {
...
+    signed
+        .verify()
+        .map_err(|err| AppError::Publish(format!("invalid push owner proof signature: {err}")))?;
+    Ok(hex::encode(signed.sig.serialize()))
+}
+
+/// Verify a Nostr event signature by `member_id_hex` over the canonical,
+/// unpublished owner-proof event. Returns false on malformed input rather than
+/// erroring, so a bad entry is dropped.
+fn verify_push_owner_sig(proof_event: UnsignedEvent, owner_sig_hex: &str) -> bool {
```
erskingardner, line 633: This hard-switches `owner_sig` verification to the new event-shaped proof, but the wire/storage contract still looks like the old one: same `marmot-push-v1` payload, same `owner_sig` field, and existing `group_push_tokens` rows already persist signatures produced over `SignedRecord` digests. That means a valid token gossip record from an older client, or a stored/transitively relayed kind 448 record created before this PR, will now be silently filtered out by `verify_push_gossip()`. If the push owner proof is intentionally breaking, I think the gossip version/proof version needs to change; otherwise this verifier should accept the legacy digest signature as a fallback while new clients start emitting event signatures.

erskingardner (changes requested):
Found two setup/compatibility issues that look worth fixing before merge. I did not re-raise the cgka-engine Nostr-layering concern because it is already tracked/deferred in this PR.

erskingardner (approved):

PR #771, opened 2026-07-06
marmot-uniffi: expose encrypted group avatars (download_group_image + image_hash_hex)

dannym-arx:
marmot

Group avatars set from the desktop client are committed as the encrypted `marmot.group.blossom.image.v1` component (0x8002). Over FFI, hosts could only reach the plain-URL avatar component (0x8007), so the Android app had no way to detect or download the encrypted image: mixed desktop-and-Android groups showed no avatar on Android.

Two changes:

- `Marmot::download_group_image(account_ref, group_id_hex)` fetches the group's Blossom blob, verifies the hash, decrypts it, and returns the raw image bytes.
- `AppGroupRecordFfi.image_hash_hex` carries the encrypted avatar's content hash, so hosts can detect presence and key their bitmap caches. Content addressing means a changed image is a new key, with no invalidation step.

Also fixes `kotlin-bindings.sh`: with `RUSTFLAGS="-C strip=symbols"` (as release pipelines set for the Android libraries) the host dylib loses the metadata uniffi-bindgen reads, and bindgen exits 0 while writing no Kotlin at all. The host build now pins RUSTFLAGS empty, and the script fails loudly when the binding file is missing.

The Android marmots can now hang the same portrait in their burrow that the desktop marmots framed.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/771">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added support for downloading a group’s Blossom avatar image directly in the app (returning decrypted raw image bytes when available).
* **Bug Fixes**
  * Improved binding-generation reliability by detecting missing UniFFI Kotlin output early and failing with a clear error.
  * Corrected group avatar metadata handling so empty Blossom image fields are treated as unavailable (and cache keys are only set when appropriate).

### DIFF
--- a/crates/marmot-uniffi/src/commands/group.rs
+++ b/crates/marmot-uniffi/src/commands/group.rs
@@ -403,6 +403,25 @@ impl Marmot {
         Ok(summary.into())
     }
 
+    /// Fetch and decrypt the group's encrypted Blossom avatar
+    /// (`marmot.group.blossom.image.v1`) into raw image bytes (PNG/JPEG/…).
+    /// Errors when the group has no Blossom image set. Presence and the
+    /// content hash (for caching) are on `AppGroupRecordFfi::image_hash_hex`;
+    /// when the group also carries a URL avatar, the URL takes precedence
+    /// for rendering.
+    pub async fn download_group_image(
+        &self,
+        account_ref: String,
+        group_id_hex: String,
+    ) -> Result<Vec<u8>, MarmotKitError> {
+        let group_id = group_id_from_hex(&group_id_hex)?;
+        let bytes = self
+            .runtime
+            .download_group_image(&account_ref, &group_id)
+            .await?;
+        Ok(bytes)
+    }
+
     /// Set (or clear, with `url = None`) the group's URL-based avatar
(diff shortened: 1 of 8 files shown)


crates/marmot-uniffi/src/commands/group.rs
```diff
@@ -403,6 +403,25 @@ impl Marmot {
         Ok(summary.into())
     }
 
+    /// Fetch and decrypt the group's encrypted Blossom avatar
+    /// (`marmot.group.blossom.image.v1`) into raw image bytes (PNG/JPEG/…).
+    /// Errors when the group has no Blossom image set. Presence and the
+    /// content hash (for caching) are on `AppGroupRecordFfi::image_hash_hex`;
+    /// when the group also carries a URL avatar, the URL takes precedence
+    /// for rendering.
+    pub async fn download_group_image(
```
erskingardner: I think that this needs to be clear in the method name that it's the blossom download (specific to that app component).

erskingardner (changes requested):

erskingardner (approved):

PR #772, opened 2026-07-06
[codex] Prepare wn-agent installer release 0.9.2

erskingardner:
## Summary

Prepare the `wn-agent-v0.9.2` release path for both Hermes and OpenClaw installer one-liners.

- Make the Hermes installer curl-pipe safe, guided/noninteractive, service-aware, and additive to existing Hermes connectors.
- Add matching OpenClaw one-liner behavior: install `wn-agent`, install/enable the OpenClaw plugin, bootstrap/reuse the agent home, accept `--allow-welcomer`, and patch only `channels.marmot`.
- Extend `wn-agent bootstrap` with `--allow-welcomer <npub-or-hex>` and include normalized allowlist state in JSON output.
- Include the Hermes config helper in release packaging and bump release-facing metadata to `0.9.2`.

## Validation

- `WN_AGENT_SHA=9.9.9 MARMOT_RELEASE_TAG=wn-agent-v9.9.9-test bash scripts/install-openclaw-marmot.sh --dry-run --yes`
- `WN_AGENT_SHA=9.9.9 MARMOT_RELEASE_TAG=wn-agent-v9.9.9-test bash -s -- --dry-run --yes < scripts/install-openclaw-marmot.sh`
- Hermes installer dry-run
- `integrations/openclaw/marmot/test/dev-scripts.sh`
- `integrations/hermes/marmot/tests/test_dev_scripts.sh`
- `python3 -m unittest discover -s integrations/hermes/marmot/tests`
- `cargo check -p agent-connector --bin wn-agent`
- `just fast-ci`

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/772">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added welcomer allowlist support to the agent bootstrap flow (including repeatable non-interactive installer inputs).
  * Updated Hermes/OpenClaw Marmot setup to patch gateway configuration with optional streaming controls and welcomer allowlist propagation.

* **Bug Fixes**
  * Improved installer/plugin packaging to ensure the correct plugin version is written and gateway configuration helpers are included.
  * Safer gateway/config updates with improved YAML handling, plus optional timestamped backups.

* **Documentation**
  * Refreshed WN Agent, Hermes, and OpenClaw install/release examples to version `0.9.2`, including updated non-interactive command snippets.

### DIFF
--- a/.github/workflows/wn-agent-binaries.yml
+++ b/.github/workflows/wn-agent-binaries.yml
@@ -194,6 +194,10 @@ jobs:
           mkdir -p "$bundle_dir"
           sed -e "s/^version:.*/version: $artifact_version/" \
             integrations/hermes/marmot/plugin.yaml > "$bundle_dir/plugin.yaml"
+          if ! grep -qx "version: $artifact_version" "$bundle_dir/plugin.yaml"; then
+            echo "error: packaged Hermes plugin metadata did not get version $artifact_version" >&2
+            exit 1
+          fi
           cp integrations/hermes/marmot/__init__.py "$bundle_dir/"
           cp integrations/hermes/marmot/adapter.py "$bundle_dir/"
           cp integrations/hermes/marmot/README.md "$bundle_dir/"
--- a/.github/workflows/wn-agent-binaries.yml
+++ b/.github/workflows/wn-agent-binaries.yml
@@ -192,10 +192,12 @@ jobs:
 
           rm -rf "$stage_parent" "$asset" "$asset.sha256"
           mkdir -p "$bundle_dir"
-          cp integrations/hermes/marmot/plugin.yaml "$bundle_dir/"
+          sed -e "s/^version:.*/version: $artifact_version/" \
+            integrations/hermes/marmot/plugin.yaml > "$bundle_dir/plugin.yaml"
           cp integrations/hermes/marmot/__init__.py "$bundle_dir/"
           cp integrations/hermes/marmot/adapter.py "$bundle_dir/"
           cp integrations/hermes/marmot/README.md "$bundle_dir/"
+          cp scripts/hermes_marmot_configure_gateway.py "$bundle_dir/configure_gateway.py"
 
           cat > "$bundle_dir/manifest.json" <<EOF
           {
@@ -208,6 +210,7 @@ jobs:
               "plugin.yaml",
               "__init__.py",
               "adapter.py",
+              "configure_gateway.py",
               "README.md",
               "manifest.json"
             ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,7 +39,7 @@ dependencies = [
 
 [[package]]
 name = "agent-connector"
-version = "0.9.1"
+version = "0.9.2"
 dependencies = [
  "agent-control",
  "agent-stream-compose",
@@ -66,7 +66,7 @@ dependencies = [
 
 [[package]]
 name = "agent-control"
-version = "0.9.1"
+version = "0.9.2"
 dependencies = [
  "serde",
  "serde_json",
@@ -76,7 +76,7 @@ dependencies = [
 
 [[package]]
 name = "agent-stream-compose"
-version = "0.9.1"
+version = "0.9.2"
 dependencies = [
  "cgka-traits",
  "hex",
@@ -712,7 +712,7 @@ checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
 name = "cgka-conformance-simulator"
-version = "0.9.1"
+version = "0.9.2"
 dependencies = [
  "async-trait",
  "cgka-engine",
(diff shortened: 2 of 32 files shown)


erskingardner:
## Adversarial review — installer release 0.9.2

Focused on the two `curl | bash` installers and the Rust/Python surface they drive. Overall the mechanics are solid (SHA256 verified, plist XML-escaped, welcomer refs normalized/deduped, existing plugin dir backed up rather than destroyed, config backups written, `--json` output stays clean because `write_bootstrap_output` early-returns before the new `Welcomer allowlist:` line, and only marmot-scoped config keys are touched). A few things I'd want addressed before tagging.

### 1. Service-install failure never falls back to the temp agent (moderate)

`install_user_service || start_temp_agent` is meant to fall back when the service manager can't bring up `wn-agent`. It effectively can't:

- The call is the left operand of `||`, so `set -e` is suspended for the entire function body.
- `install_macos_service` ends on `log "installed and started LaunchAgent: $label"`, and `install_linux_user_service` ends on `log "installed and started systemd user service"`. Both return `0` even when the preceding `launchctl bootstrap` / `systemctl --user enable --now` failed.
- So `install_user_service` returns non-zero only on the *explicit* `return 1` paths (non-Darwin/Linux, or `systemctl` binary absent). A genuine runtime failure of `launchctl`/`systemctl` returns `0`, the log falsely claims success, `start_temp_agent` is skipped, the socket never appears, and the run instead hard-fails ~15s later at `wn-agent bootstrap`.

Net effect: the fallback is dead code for the most likely failure mode, and the user sees a misleading "installed and started" line right before an unrelated-looking bootstrap timeout. Consider capturing the service-manager exit status explicitly (e.g. `run launchctl bootstrap ... || return 1`) so the `|| start_temp_agent` fallback actually engages. Same pattern in both scripts.

### 2. The no-arg default got much more invasive, and side effects precede validation (moderate)

Previously `curl … | bash` with no flags installed the binary + plugin and printed next steps. Now the default path also installs a persistent LaunchAgent/systemd unit, starts a background daemon, bootstraps a Nostr identity, and rewrites gateway config — and adds hard deps (`python3`; `node` for OpenClaw; `hermes`/`openclaw` now required even for a plain install). That's a big blast-radius change for a piped one-liner; at minimum it deserves a prominent call-out in the release notes.

More concretely, it inverts the repo's own validate-before-mutate discipline (`docs/marmot-architecture/overview/multi-step-state-changes.md`): `--allow-welcomer` values aren't validated until `wn-agent bootstrap` parses them, which runs *after* the binary is installed, the service is installed/enabled, and the agent is started. A malformed npub/hex aborts mid-run under `set -e`, leaving an enabled service and a running daemon but no gateway config. Validating the welcomer refs up front (before any install step) would keep a bad arg from leaving a half-configured, running service behind.

### 3. Headline behavior has no automated coverage (minor)

`test_dev_scripts.sh` / `dev-scripts.sh` only exercise `--dry-run --yes`, which returns before service install, bootstrap, and config patching. The service units, the `bootstrap.json` parse, and the config-patch (Python for Hermes, inline Node for OpenClaw) never run in CI. Given #1 and #2 live entirely on the non-dry-run path, some coverage there (even just asserting the generated plist/unit and the patched config JSON/YAML against fixtures) would be worth it.

### 4. Unmanaged temp agent is never tracked or torn down (minor)

In the Hermes script `start_temp_agent` sets `WN_AGENT_TEMP_PID="$!"` but nothing ever reads it; the OpenClaw script doesn't capture the PID at all. On `--no-service` (or the intended fallback path), a background `wn-agent` is left running with no PID record and no mention in `print_next_steps`, so the user has an unmanaged daemon they weren't told about and can't easily stop. Either drop the unused var or surface/track the process.

### 5. `ExecStart=%q` is shell-quoting, not systemd quoting (minor/edge)

The systemd unit builds `ExecStart` with bash `printf %q`. systemd unit files use their own quoting rules, not shell quoting. For the default paths (no spaces/special chars) `%q` is a no-op and works, but a `--home`/relay containing spaces or backslashes will be quoted in a way systemd parses differently. Low severity given defaults, but the two quoting schemes aren't interchangeable.

### 6. Nits

- `append_csv` declares `local value`/`local item` but not `_items`, so the array leaks to global scope in both scripts.
- User-facing installer changes (new default behavior, new flags, dropped `--bootstrap` semantics) aren't reflected in any changelog; `crates/cli/CHANGELOG.md` still shows `[Unreleased]` pinned at `0.9.0` while the workspace bumps to `0.9.2`.
- `--bootstrap` silently becoming a no-op alias is reasonable for back-compat, but anyone who scripted `--no-start-wn-agent` *without* `--bootstrap` previously got no bootstrap; now bootstrap is unconditional. Worth a note that old flag combinations changed meaning.

Nothing here is a data-loss/security regression — the concerns are reliability of the new default path (#1), an intentionally larger blast radius that's under-validated and under-tested (#2/#3), and cleanup/quoting polish (#4–#6).

PR #781, opened 2026-07-07
Make wn-opencode production ready

erskingardner:
## Summary

This evolves the opencode Marmot integration from the PR #773 prototype into a production-shaped harness alongside the Hermes and OpenClaw connector work.

- Moves the harness to `integrations/opencode/marmot` and splits the single-file prototype into focused Rust modules for config, control-plane IO, opencode execution, chunking, session state, repo picking, and bridge runtime.
- Adds a release installer for `wn-agent` + `wn-opencode`, service setup paths, release workflow packaging, and docs.
- Adds strict allowlist mirroring, privacy-safe tracing, private session persistence, per-group backpressure, reconnect/dedupe behavior, retryable durable sends, and 30KB byte-based reply chunking.
- Adds deterministic production-readiness coverage: unit tests, installer tests, tracing audit coverage, and a real `wn-agent` debug-control E2E with a fake opencode binary.

## Why

The original opencode PR was a useful proof of concept, but it was still too monolithic and under-tested to treat like the Hermes/OpenClaw connector integrations. This PR keeps opencode as a pure harness, not a gateway, while giving it the install, release, runtime, and validation posture we expect for production use.

## Validation

- `cargo test -p wn-opencode`
- `cargo clippy -p wn-opencode --all-targets -- -D warnings`
- `just opencode-dev-e2e-connector`
- `just tracing-audit`
- `just fast-ci`
- `git diff --check`

I also checked the locally installed opencode `1.17.14` CLI. The documented `run --format json` / `--session` surface matches the harness; the local provider route returned opencode JSON error events, which the harness now handles with privacy-safe summaries.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/781">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added the `wn-opencode` Marmot harness, release packaging/publishing, and `install-opencode-marmot.sh`, including new versioned bundle assets and a refreshed `wn-agent-latest` installer flow.
* **Bug Fixes**
  * Improved OpenClaw packaging tarball discovery with stricter failure when no archive is produced.
* **Documentation**
  * Expanded Opencode integration docs and installer/development guidance; updated Hermes/OpenClaw defaults and commands to use the latest installer flow.
* **Tests**
  * Added Opencode installer dry-run tests, optional e2e connector coverage, and expanded conformance source-audit checks.


erskingardner (commented):
## Adversarial review

Solid productionization overall — modular split, fs-private session store, allowlist mirroring, idempotent sends, chunking, and installer/E2E coverage all look good. I did find several correctness and privacy issues worth addressing before merge.

**Blockers / high:**
1. `ResyncRequired` is only logged; Hermes forces a reconnect so the subscription can reconcile after broadcast lag. Staying on the same stream after a resync signal can silently drop prompts.
2. Inbound dedupe runs before queue admission — a queue-full reject marks the message as seen without processing it; redelivery of the same `message_id_hex` is then dropped forever.
3. Opencode stderr (up to 4KB) is forwarded into Marmot replies, which can leak provider errors, paths, or secrets into the group chat and conflicts with the crate's privacy rules.

**Medium:** partial chunk send failures and opencode task join errors leave the user with no follow-up; persisted session `cwd` is not re-validated on load; multi-account setups silently pick the first account.

Inline comments have details and suggested fixes.

integrations/opencode/marmot/src/bridge.rs
```diff
@@ -0,0 +1,672 @@
...
+        }
+        AgentControlEvent::ResyncRequired { dropped_events, .. } => {
+            warn!(
+                target: TRACE_TARGET,
+                method = "dispatch_event",
+                event = "resync_required",
+                dropped_events,
+                "inbound event resync required"
+            );
+        }
```
erskingardner, line 284: **High — `ResyncRequired` should force reconnect, not just log.**

`agent-connector` emits this when broadcast lag exceeds the replay window or storage replay fails — dropped inbound messages are gone and the client is expected to reconcile (see `event_projection.rs` and Hermes `adapter.py::_handle_resync_required`, which clears caches and tears down the subscription).

Here the subscription stays open, so prompts that were dropped during lag are silently lost with no user-visible signal.

**Suggested fix:** treat `ResyncRequired` like a disconnect (return `DrainOutcome::Disconnected` from `dispatch_event`, or break the recv loop), and consider clearing `InboundDedupe` on reconnect so legitimately replayed ids aren't blocked after recovery.

integrations/opencode/marmot/src/bridge.rs
```diff
@@ -0,0 +1,672 @@
...
+                        &ctx_for_reply,
+                        &account_id_hex,
+                        &group_id_hex,
+                        &message_id_hex,
+                        "[wn-opencode] too many prompts are already queued for this group; try again shortly.",
+                        0,
+                    )
+                    .await;
+                });
+                return;
```
erskingardner: **High — dedupe before queue admission can permanently drop prompts.**

`dedupe.insert()` runs at L191–199, but `try_enter()` can fail at L200. In that case the message id is marked seen even though `handle_message` never runs.

If wn-agent later replays the same `message_id_hex` (small lag recovery path) or the busy reply at L209–217 fails to send, the prompt is silently dropped on redelivery.

**Suggested fix:** acquire the queue permit first, or roll back/remove the dedupe entry when rejecting for `queue_full`.

integrations/opencode/marmot/src/bridge.rs
```diff
@@ -0,0 +1,672 @@
...
+            && code != 0
+        {
+            message.push_str(&format!(" (exit {code})"));
+        }
+        if outcome.stderr.is_empty() {
+            message.push('.');
+        } else {
+            message.push_str(":\n");
+            message.push_str(&outcome.stderr);
+        }
```
erskingardner: **High — stderr is forwarded into Marmot group messages.**

When `chunk_count == 0`, up to 4096 bytes of opencode stderr are appended to the reply sent to the group. Provider/API errors often include paths, model names, or credential hints.

This contradicts `AGENTS.md` ("no local paths" / privacy-safe outputs) and can leak operational detail to every group member.

**Suggested fix:** keep stderr local (logs only, privacy-safe summary), or truncate/redact aggressively and gate behind a dev flag.

integrations/opencode/marmot/src/bridge.rs
```diff
@@ -0,0 +1,672 @@
...
+                    )
+                    .await
+                    {
+                        warn!(
+                            target: TRACE_TARGET,
+                            method = "send_final",
+                            error_kind = err.privacy_safe_kind(),
+                            "failed to send opencode reply chunk"
+                        );
+                    }
```
erskingardner, line 383: **Medium — partial chunk delivery fails open.**

If `send_reply` fails after earlier chunks succeeded, the code only emits a `warn!` and continues/finishes. The user sees a truncated answer with no indication that later chunks were lost.

Consider aborting the opencode run and sending a single follow-up error chunk (with a fresh idempotency key) when a mid-stream send exhausts retries.

integrations/opencode/marmot/src/bridge.rs
```diff
@@ -0,0 +1,672 @@
...
+        }
+        Err(err) => {
+            let err = HarnessError::from(err);
+            warn!(
+                target: TRACE_TARGET,
+                method = "opencode_run",
+                error_kind = err.privacy_safe_kind(),
+                "opencode task join failed"
+            );
+        }
```
erskingardner, line 448: **Medium — opencode task join failure is silent to the user.**

On `runner.await` `Err` (join/panic), this path logs and returns without calling `send_reply`. The inbound prompt gets no response.

Mirror the `Ok(Err(err))` branch and send a generic failure reply.

integrations/opencode/marmot/src/bridge.rs
```diff
@@ -0,0 +1,672 @@
...
+    }
+}
+
+async fn resolve_cwd_and_prompt(
+    ctx: &BridgeContext,
+    inbound: &InboundPrompt,
+    known_session: Option<&SessionRecord>,
+) -> Result<Option<(PathBuf, String)>> {
+    if let Some(record) = known_session {
+        return Ok(Some((record.cwd.clone(), inbound.text.clone())));
```
erskingardner: **Medium — persisted session cwd is trusted without re-validation.**

For known sessions, `record.cwd` from disk is passed straight to opencode. `resolve_repo()`'s `$HOME` containment check only runs on the first `/repo` picker message.

A tampered `sessions.json`, symlink swap, or directory removal after pinning could run opencode outside the intended repo.

**Suggested fix:** re-run the same canonicalize-under-`$HOME` check (or stat+is_dir) before each invocation.

integrations/opencode/marmot/src/bridge.rs
```diff
@@ -0,0 +1,672 @@
...
+    }
+    if accounts.len() > 1 {
+        warn!(
+            target: TRACE_TARGET,
+            method = "account_list",
+            count = accounts.len(),
+            "multiple accounts available; using first"
+        );
+    }
+    Ok(accounts[0].account_id_hex.clone())
```
erskingardner, line 603: **Medium — ambiguous multi-account selection.**

When `WN_OPENCODE_ACCOUNT_ID_HEX` is unset and wn-agent has multiple accounts, the harness warns and picks `accounts[0]`. On a shared machine this can bind to the wrong identity without failing fast.

Consider requiring an explicit account id whenever `accounts.len() > 1` (installer could persist it from bootstrap.json).

integrations/opencode/marmot/src/opencode.rs
```diff
@@ -0,0 +1,351 @@
...
+pub(crate) fn build_run_args(session_id: Option<&str>, prompt: &str) -> Vec<String> {
+    let mut args = vec!["run".to_owned(), "--format".to_owned(), "json".to_owned()];
+    if let Some(session_id) = session_id
+        && !session_id.is_empty()
+    {
+        args.push("--session".to_owned());
+        args.push(session_id.to_owned());
+    }
+    args.push("--".to_owned());
+    args.push(prompt.to_owned());
```
erskingardner, line 177: **Low — prompt content is visible in the process list.**

Passing the full prompt as a CLI argument after `--` means `ps` on the host can expose user prompts (and any secrets they contain) to other local users.

Worth documenting as a deployment constraint; stdin-based invocation would reduce exposure if opencode supports it.

integrations/opencode/marmot/src/opencode.rs
```diff
@@ -0,0 +1,351 @@
...
+    let mut buf = Vec::new();
+    let mut captured = String::new();
+    while let Ok(read) = reader.read_until(b'\n', &mut buf).await {
+        if read == 0 {
+            break;
+        }
+        if captured.len() < STDERR_CAPTURE_BYTES {
+            captured.push_str(&String::from_utf8_lossy(&buf));
+            if captured.len() > STDERR_CAPTURE_BYTES {
+                captured.truncate(STDERR_CAPTURE_BYTES);
```
erskingardner: **Bug: `truncate` can panic on a UTF-8 boundary.** `String::truncate` panics if the index is not a char boundary. `STDERR_CAPTURE_BYTES` (4096) is a fixed byte offset, and `captured` accumulates arbitrary opencode stderr — any multi-byte UTF-8 (accented chars, emoji, non-English provider/API error text) straddling byte 4096 panics this task. It's caught by `stderr_task.await?` → `HarnessError::Join`, so the process survives, but a normal opencode failure becomes a generic "task join failed" that discards the real stderr — defeating the diagnostics this function exists to provide.

Reuse the `floor_char_boundary` helper already added in `chunking.rs` (line 42) to find the largest boundary `<= STDERR_CAPTURE_BYTES` before truncating.

integrations/opencode/marmot/src/bridge.rs
```diff
@@ -0,0 +1,672 @@
...
+    match event {
+        AgentControlEvent::InboundMessage {
+            account_id_hex,
+            group_id_hex,
+            message_id_hex,
+            sender_account_id_hex,
+            text,
+            ..
+        } => {
+            if sender_account_id_hex == ctx.account_ref {
```
erskingardner: **Self-echo check isn't case-normalized like the allowlist check right below it.** This compares `sender_account_id_hex == ctx.account_ref` byte-for-byte, while line 181 normalizes `sender_account_id_hex.to_ascii_lowercase()` before the `allowed_senders` check (and those are normalized via `normalize_hex` at config-parse time). Hex ids are consistently lowercase today so this is latent, but if `sender_account_id_hex` ever arrives with different casing than `ctx.account_ref`, the harness fails to recognize its own echoed message and could prompt opencode with its own reply text. Cheap to route both sides through the same normalization.

integrations/opencode/marmot/src/store.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub(crate) fn set(&self, group_key: &str, record: SessionRecord) -> Result<()> {
+        let mut map = self.map.lock().expect("session store lock poisoned");
+        let mut next = map.clone();
+        next.insert(group_key.to_owned(), record);
+        self.write_snapshot(&next)?;
+        *map = next;
+        Ok(())
+    }
+
+    fn write_snapshot(&self, snapshot: &HashMap<String, SessionRecord>) -> Result<()> {
```
erskingardner: **Blocking fs I/O held under a `std::sync::Mutex` in async code.** `set`/`get` use a sync `Mutex` called from async (`bridge.rs` `finish_success` / `resolve_cwd_and_prompt`). Two effects: (1) `write_snapshot` runs blocking filesystem syscalls while holding the lock, blocking the tokio worker for the write/rename — fine at today's volume, a foot-gun if traffic grows; consider `spawn_blocking`. (2) `std::sync::Mutex` poisons on panic, so any panic while the lock is held (e.g. inside `write_snapshot`) permanently breaks session persistence for the rest of the process — every later `.lock().expect(...)` panics. A single point of unrecoverable failure for session continuity, which should degrade gracefully instead.

integrations/opencode/marmot/src/bridge.rs
```diff
@@ -0,0 +1,672 @@
...
+            inner: Mutex::new(HashMap::new()),
+        }
+    }
+
+    async fn try_enter(&self, group_ref: &str) -> Option<GroupPermit> {
+        let queue = {
+            let mut inner = self.inner.lock().await;
+            inner
+                .entry(group_ref.to_owned())
+                .or_insert_with(|| {
```
erskingardner: **Unbounded per-group queue map.** Every distinct `group_id` ever seen gets a permanent `Arc<GroupQueue>` (a `Mutex` + `Semaphore`) here for the life of the process — nothing evicts entries once a group's queue drains. `InboundDedupe` right next to it is deliberately bounded (`DEDUPE_LIMIT = 2048`, LRU eviction), so this reads as an oversight rather than intent. For a long-running "production-shaped" service it's a slow memory leak proportional to distinct-groups-over-lifetime. Worth evicting entries whose semaphore has no outstanding permits, similar to the dedupe cache.

erskingardner (commented):
Adversarial review (automated, via Claude Code). Overall the crate is well-structured and unusually well-tested for a harness integration — genuine e2e test spawning real `wn-agent` + a fake opencode binary, symlink-safe repo confinement in `resolve_repo`, prompt/flag isolation after `--`, idempotent retryable sends, and the tracing-privacy audit extended to cover the new crate. One real correctness bug plus three lower-severity robustness nits below.

erskingardner:
@coderabbitai full review

PR #782, opened 2026-07-07
Feat/repro loop

jgmontoya:
---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/782">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added an incident replay tool and CLI to classify exports, recover fork/convergence outcomes, synthesize conformance vectors, and accept only reproducible results.
  * Added new conformance vectors for committer- vs witness-based convergence selection, plus incident recovery scenarios.
  * Simulator now captures and lets scenarios assert “settled” convergence decision observations.
* **Bug Fixes**
  * Re-admits previously delivered app messages for witness scoring.
  * Improves replay robustness by treating a specific OpenMLS replay failure mode as non-fatal.
* **Documentation**
  * Added incident-replay adapter documentation and updated the repository vector/incident manifest.
* **Tests**
  * Expanded simulator and incident-replay end-to-end coverage, including convergence decision assertions and acceptance/recovery flows.

### DIFF
--- a/crates/incident-replay/src/convergence.rs
+++ b/crates/incident-replay/src/convergence.rs
@@ -8,11 +8,13 @@
 //!
 //! Two reproducible shapes are recovered: the **committer-decided** case (the
 //! selector reached the authenticated-committer tiebreak with no app-witness
-//! quorum) and the **witness-decided** case (an app-witness quorum won by greater
-//! effective depth) — the case that matches real convergence traffic. Anything
-//! else (a committer tiebreak that itself met a quorum, or a priority/digest/
-//! differing-depth rule) fail-closes rather than synthesize a vector that would
-//! silently assert the wrong outcome.
+//! quorum) and the **witness-decided** case (two equal-depth branches where an
+//! app-witness quorum's boost broke the tie) — the case that matches real
+//! convergence traffic. Anything else fail-closes rather than synthesize a
+//! vector that would silently assert the wrong outcome: a committer tiebreak that
+//! itself met a quorum, a witness winner whose branches were *not* at equal depth
+//! (so it won on raw commit depth, not the quorum boost), indistinguishable
+//! candidate branches, or a priority/digest rule.
 
 use crate::export::{AgentStateExport, ConvergenceCandidate, ConvergenceRule, EventKind};
 
@@ -57,6 +59,8 @@ pub enum ConvergenceRecoveryError {
     NoConvergenceDecision,
     #[error("expected exactly two contested branches, found {0}")]
     AmbiguousCandidates(usize),
+    #[error("the two candidates do not have distinct, non-empty branch ids")]
+    IndistinctCandidates,
     #[error("the convergence_decision recorded no decisive selector rule")]
     NoDecisiveRule,
     #[error(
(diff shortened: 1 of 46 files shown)


crates/incident-replay/src/convergence.rs
```diff
@@ -0,0 +1,169 @@
...
+    // - `effective_commit_depth` is witness-decided only with a *met* quorum (which
+    //   implies an app-witness score at the quorum threshold); without one the
+    //   winner won by raw commit depth (a differing-depth branch), which the
+    //   equal-depth witness vector can't reproduce. The rule is supported here — the
+    //   quorum status is what is not — so this is a quorum error, not an
+    //   `UnsupportedDecisiveRule`.
+    // - any other rule (priority, digest, valid-depth) has no v1 vector shape.
+    let kind = match (decisive, quorum_met) {
+        (TIP_COMMITTER_RULE, Some(false)) => ConvergenceDecisionKind::CommitterDecided,
+        (EFFECTIVE_COMMIT_DEPTH_RULE, Some(true)) => ConvergenceDecisionKind::WitnessDecided,
```
erskingardner: [P2] Confirm the witness quorum actually caused this effective-depth win before synthesizing a witness-decided vector. Right now any `effective_commit_depth` decision whose winner has `witness_quorum_met: true` maps to `WitnessDecided`, but the adapter does not parse the winner/loser `valid_commit_depth`, `effective_commit_depth`, or `app_witness_score`. A branch can win effective depth because it has more raw commit depth and also happen to have quorum; the synthesized equal-depth witness vector would then assert the wrong incident shape. Please fail closed unless the recorded scores prove the quorum boost/app witness score is the decisive factor, and add a regression for the raw-depth-plus-quorum case.

crates/incident-replay/src/export.rs
```diff
@@ -0,0 +1,193 @@
...
+    },
+    #[serde(other)]
+    Other,
+}
+
+/// One branch the convergence selector evaluated. Only the fields extraction
+/// reads are modelled; every other field the export carries is ignored.
+#[derive(Debug, Clone, Deserialize)]
+pub struct ConvergenceCandidate {
+    #[serde(default)]
```
erskingardner, line 120: [P2] Reject missing or duplicate branch IDs before recovery. Defaulting `branch_id` to `""` lets malformed candidate entries pass parsing, and `recover_convergence` only checks that there are exactly two candidates and that `selected_branch_id` matches one of them. An export with a missing or duplicate loser can still be treated as a reproducible two-branch incident. Please require two non-empty, distinct candidate branch IDs before synthesizing.

erskingardner (commented):
I focused this review on the convergence replay / witness recovery surface, especially the commit that re-admits already-delivered app messages as convergence witnesses. The engine-side witness-counting fix looks directionally sound to me; I left two inline comments on the incident-replay recovery path where I think the adapter should fail closed more strictly before synthesizing a vector.

crates/incident-replay/src/convergence.rs
jgmontoya: good catch, fixed by narrowing the accepted shape instead of just parsing the fields: witness-decided now requires the two branches at equal valid_commit_depth with the winner's quorum boost breaking the tie (winner/loser depths + scores are parsed now). Anything else fail-closes with a named error (WitnessQuorumUnconfirmed), so a raw-depth win that happens to hold quorum can't synthesize the wrong causal story. Doc-comment enumerates each rejected shape.

crates/incident-replay/src/export.rs
jgmontoya: done! recovery now requires two non-empty, distinct branch ids and fail-closes with IndistinctCandidates otherwise. Also noted in the module docs which fields the lenient parser is allowed to be lenient about (unknown/extra fields) vs the ones synthesis depends on (these).

mubarakcoded (approved):
LGTM! though I skimmed mostly through incident-replay crate

PR #789, opened 2026-07-08
Add marmot-c C ABI bindings for the Marmot runtime

### LINKED ISSUE #328: Add C bindings for the Marmot runtime
## Problem

The Marmot runtime is exposed via UniFFI today (Swift/iOS, Android via Kotlin). Consumers that can't pull in a UniFFI runtime — embedded targets, C/C++ apps, language bindings without UniFFI support (Zig, Nim, raw FFI from Go/Lua/etc.), and any FFI consumer that wants a stable C ABI — have no way to use the runtime directly.

The MDK family already includes `mdk-c` alongside the Kotlin/Swift/Ruby/Python/Zig/JS bindings, so a C surface is a known shape in this org. Dark Matter should expose the same.

## Proposed change

Add a C-binding crate (e.g. `crates/marmot-c` or `crates/dm-c`) that exposes the runtime through a stable C ABI:

- C header file generated from the Rust surface (cbindgen or hand-curated).
- `cdylib` + `staticlib` build targets.
- Coverage of the same surface UniFFI exposes today: account/session, group operations, message send/receive, agent text streams, storage primitives.
- Packaging notes for how downstream apps link the artifact (pkg-config or a minimal CMake snippet is enough to start).

## Notes

- Implementation choice (cbindgen vs handwritten `extern \"C\"` vs another tool) left to maintainers.
- Useful prior art in `marmot-protocol/mdk-c`.
- Related: existing UniFFI bindings under `crates/marmot-uniffi` (PRs marmot-protocol/mdk#24, marmot-protocol/mdk#25, marmot-protocol/mdk#70).

dannym-arx:
marmot

Marmots who can't pull in a UniFFI runtime had no way to talk to the Marmot runtime. This adds `marmot-c`: a stable C ABI over `marmot-uniffi`, so C/C++ and raw-FFI callers (Zig, Nim, Go, Odin, Lua, PHP) can drive it directly.

## What's in it

- `cdylib` + `staticlib` build, with a cbindgen-generated `include/marmot.h` checked into the repo and diff-gated in CI.
- Every runtime record and enum has a `#[repr(C)]` mirror built from the `marmot-uniffi` `...Ffi` types, so the C shape stays in lockstep with the Swift/Kotlin surface. Values cross as owned pointers freed only by their matching `marmot_*_free`; inputs are borrowed and never retained.
- Async runtime work runs on an embedded tokio runtime through blocking calls. Subscriptions offer blocking `*_next` reads or callback registration.
- Memory-safety scaffolding, because the point is to hand C a stable ABI and not a footgun: an `alloc-audit` test feature that proves every deep-free releases what it allocated, `catch_unwind` at every boundary so a panic never crosses into C, and a typed `MarmotStatus` code per error variant with a thread-local last-error channel.

## Verification

`c-smoke.sh` builds and runs seven worked example consumers, one per language, each driving the same slice of the runtime: lifecycle, markdown parsing, offline reads, the error taxonomy, and best-effort identity. The C, Zig, and Nim examples also walk the recursive Markdown tagged-union tree. Together they cover both linkage models (static archive and shared object) and both binding styles (header import and hand-declared decls). The C example runs under valgrind in CI.

Closes #328

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/789">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added official Marmot C bindings with a versioned Linux x86_64 distribution bundle (zip, manifest, checksums) published for `marmotc-v<version>` tag releases.
  * Expanded the C ABI surface (client lifecycle, accounts, groups, messaging, media, notifications, relays/telemetry, timeline access, and subscriptions), plus a pkg-config file and bundled `marmot.h`.
* **Tests**
  * Added C bindings CI smoke tests with header regeneration diff-gating and GCC/Clang runs, including Valgrind when available and multiple language consumers.
* **Documentation**
  * Added/updated Marmot C bindings docs and release guidance for the new `marmotc-v*` track.


crates/marmot-c/src/types/event.rs
```diff
@@ -0,0 +1,783 @@
...
+unsafe impl Send for MarmotEvent {}
+
+/// Free an event root returned by the events subscription. NULL is a no-op.
+///
+/// # Safety
+/// `event` must be NULL or an unfreed pointer returned by this library.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn marmot_event_free(event: *mut MarmotEvent) {
+    unsafe { free_boxed(event) };
+}
```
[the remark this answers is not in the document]
dannym-arx: it's not possible to make this not use unsafe, and we WANT things to crash on double free because that could result in memory going haywire

Datawav:

The C ABI is justified for native C/C++/Zig/Odin and raw-FFI consumers that UniFFI does not serve directly. Mirroring the existing UniFFI surface, centralizing allocation/free behavior, and generating the header with cbindgen is preferable to maintaining separate hand-written bindings per language. A narrower alternative would expose only a minimal C core, but would force downstream wrappers to recreate substantial domain conversion and subscription logic.

Actionable findings:

1. **High — Callback cancellation provides no safe point to reclaim `user_data`.** In `crates/marmot-c/src/subscriptions.rs`, `SubscriptionCore::clear` removes the task handle and calls non-blocking `abort()`. If the pump is awaiting its next item, cancellation prevents the terminal NULL callback; if a callback is executing, clear/free returns before it finishes. Consequently, the documented advice to observe the terminal callback or a quiescent flag cannot reliably establish that the task will never access `user_data` again. This can cause use-after-free in otherwise ordinary native cleanup. Provide an explicit join/quiescence API, or make the pump own a callback context whose destructor runs after termination. Test cancellation both while awaiting and during a callback.

2. **Medium — Subscription handles may be freed during a blocking operation without a documented concurrency invariant.** Each `*_next` obtains a reference through `sub_ref` and then may block indefinitely through `sub.core.block_next`; concurrent `*_subscription_free` immediately reconstructs and drops the same allocation. That invalidates the reference still used by the blocking call. Either make operation lifetime independent of the raw handle, such as through a refcounted core, or explicitly require in the generated header and README that free cannot overlap `next`, callback registration, clearing, or pagination on the same handle.

3. **Medium — Release tags are not verified against the crate version.** `.github/workflows/c-bindings.yml` validates only tag syntax, derives filenames and release metadata from that tag, and merely records the workspace version separately. Because `marmot-c` inherits `version.workspace`, a mistyped tag can publish binaries whose compiled/pkg-config version differs from the advertised release. Compare the parsed `marmotc-v…` version with the workspace/package version and fail before building or publishing if they differ.

_Advisory audit of head [`43e035a76385`](https://github.com/marmot-protocol/mdk/commit/43e035a76385f0deae0970e20b0f88a40fedfac0); no approval, merge, push, checkout, or PR-code execution was performed._

Datawav:
## Nightly deep audit

1. **High — Callback cancellation has no safe reclamation point for `user_data`.**  
   `SubscriptionCore::clear` removes the task handle and calls non-blocking `abort()` (`crates/marmot-c/src/subscriptions.rs:151-160`). If a callback is executing, both `clear_callback` and subscription destruction can return while it still dereferences `user_data`. If the pump is instead waiting in `next().await`, abort prevents the documented terminal NULL callback (`subscriptions.rs:215-230`). Consequently, the suggested synchronization at lines 26-29 cannot establish quiescence after cancellation. Reinstalling a callback immediately after `clear` can also overlap the new callback with the still-running old callback. Native callers must otherwise leak callback state or build lifetime tracking outside the API.

   The binding needs push subscriptions, but non-blocking cancellation alone is not a complete ownership design. Add an explicit completion/join operation, or accept an owned callback context plus destructor that runs exactly once after the pump terminates. A blocking `clear` is another alternative only if self-cancellation from inside callbacks is detected to avoid deadlock. Test cancellation while awaiting, during a callback, callback replacement, self-clear, and subscription destruction; assert both callback quiescence and exactly-once context reclamation.

2. **Medium — Release tags can publish binaries under a version different from the compiled ABI package.**  
   The workflow validates only the tag’s syntax (`.github/workflows/c-bindings.yml:32-39`). Packaging derives filenames, manifest `version`, release title, and release assets from that tag, while `workspace_version` is merely recorded separately (`:70-95`). For example, tagging the current `0.9.3` source as `marmotc-v1.0.0` would publish 0.9.3 libraries and pkg-config metadata inside a 1.0.0 archive and release. That undermines dependency resolution, ABI compatibility decisions, and rollback provenance.

   Fail before building unless the tag version equals the `marmot-c` package’s resolved Cargo version. Prefer `cargo metadata` over parsing the first root `version` line, since the crate currently inherits its version but may later stop doing so. A reusable validation script tested with matching, mismatching, prerelease, and build-metadata versions would also keep local release preparation consistent with CI.

3. **Low — Existing release assets are mutable under a stable version and checksum name.**  
   On a rerun or recreated tag, the release job uses `gh release upload --clobber` and rewrites the notes (`.github/workflows/c-bindings.yml:148-150`). The same versioned URL and `.sha256` filename can therefore resolve to different binaries later. Because these are prebuilt native crypto/runtime libraries, consumers who did not pin the original checksum have no durable convergence point, and mirrors may retain a mixture of old archives and new checksum files.

   Make published version assets immutable: fail if the release or asset already exists. If replacement is genuinely necessary, issue a new version or uniquely identify artifacts by commit and preserve prior checksums. A less desirable alternative is a manual, audited replacement workflow with provenance attestations; ordinary CI reruns should never clobber published assets.

Datawav:
Automation marker: repaired merge conflicts with `master` and pushed merge commit `2a18dd644a31e4ecf62357a38dbdf292de5b5fae` to `c-bindings`. Conflict resolution preserves both the new upstream CI/Justfile gates and the PR’s C-bindings smoke/header targets. Local verification: `cargo fmt --all --check`, `git diff --check`, and `just --list` passed. CI is re-running on the repaired head.

Datawav (commented):
## Automated review

Outcome: 3 actionable findings.

- High — Mixed subscription consumers race for the same stream. `subscriptions.rs:116-130` executes every blocking `next` directly, while `:138-148` reserves only the callback slot; `marmot_events_subscription_next` at `:289-301` and the callback pump at `:319-339` can therefore call `inner.next()` concurrently. A callback, blocking reader, or second blocking reader can nondeterministically consume an update intended for another consumer, causing missed UI updates or inconsistent state. Track consumption mode in `SubscriptionCore`: atomically reject callbacks while a blocking read is active, reject reads while a callback is installed, and serialize or reject concurrent reads; release reservations on every exit path.

- High — Caller-controlled Rust enum discriminants make malformed C input undefined behavior before validation. Exported functions accept `MarmotPushPlatform` and `MarmotNotificationWakeSource` directly (`commands/push.rs:53`; `commands/notification.rs:119`), while caller-owned structures contain `MarmotAuditDataMode` (`types/audit.rs:245,261-264`); their conversion matches in `types/push.rs:35-43`, `types/notification.rs:42-51`, and `types/audit.rs:37-44` occur only after Rust has received the enum value. C can supply an arbitrary integer, and an invalid Rust enum discriminant is UB that `ffi_guard` cannot catch. Expose fixed-width integer tags at the ABI boundary, validate them, then convert to internal enums; apply the same rule to caller-supplied tagged-union discriminants.

- High — Required output pointers are validated only after mutations execute. `commands/mod.rs:41-60,89-119` calls `write_out` inside `deliver`, `deliver_scalar`, and `deliver_string`, but wrappers first execute operations: `marmot_send_text` calls `block_on(send_text(...))` before `deliver` (`commands/message.rs:23-40`), and `marmot_create_group` similarly creates the group before checking `out_group_id_hex` (`commands/group.rs:35-55`). A null output can return `MARMOT_STATUS_NULL_POINTER` after a message was sent or a group created, encouraging a retry that duplicates the side effect. Preflight every required output pointer before invoking the runtime operation, preferably through a shared wrapper helper.

crates/marmot-c/src/subscriptions.rs
```diff
@@ -0,0 +1,1849 @@
...
+    timeout_ms: u32,
+    out_event: *mut *mut MarmotEvent,
+) -> MarmotStatus {
+    ffi_guard(|| {
+        let sub = match unsafe { sub_ref(sub) } {
+            Ok(s) => s,
+            Err(status) => return status,
+        };
+        let inner = sub.inner.clone();
+        let result = sub.core.block_next(timeout_ms, inner.next());
```
Datawav, line 300: `block_next` can poll `inner.next()` while the callback pump (or another blocking caller) polls the same subscription concurrently; `install` reserves only the callback slot and does not coordinate blocking reads. Consumers can therefore race and nondeterministically steal updates from each other. Track the consumption mode in `SubscriptionCore`: reject reads while a callback is installed, reject callback installation while a read is active, and serialize or reject concurrent reads, releasing reservations on every exit path.

crates/marmot-c/src/commands/push.rs
```diff
@@ -0,0 +1,127 @@
...
+/// # Safety
+/// `client` must be a live handle; `account_ref`, `raw_token`, and
+/// `server_pubkey_hex` valid strings; `relay_hint` a valid string or NULL;
+/// `out_registration` a valid pointer. Free the result with
+/// `marmot_push_registration_free`.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn marmot_upsert_push_registration(
+    client: *const MarmotClient,
+    account_ref: *const c_char,
+    platform: MarmotPushPlatform,
```
Datawav, line 55: Accepting `MarmotPushPlatform` directly from C makes malformed input undefined behavior before this function can validate it: C may pass an arbitrary integer, but invalid Rust enum discriminants are already UB on entry and cannot be caught by `ffi_guard`. Use a fixed-width integer tag at the ABI boundary, validate it, and only then convert to the internal enum. The same issue applies to the notification wake-source and caller-supplied audit data-mode discriminants.

crates/marmot-c/src/commands/message.rs
```diff
@@ -0,0 +1,327 @@
...
+    out_summary: *mut *mut MarmotSendSummary,
+) -> MarmotStatus {
+    ffi_guard(|| {
+        let client = try_arg!(unsafe { client_ref(client) });
+        let account_ref = try_arg!(unsafe { required_str(account_ref) });
+        let group_id_hex = try_arg!(unsafe { required_str(group_id_hex) });
+        let text = try_arg!(unsafe { required_str(text) });
+        unsafe {
+            deliver(
+                client.block_on(client.marmot.send_text(account_ref, group_id_hex, text)),
```
Datawav, line 36: The send operation executes before `deliver` validates `out_summary`. If the caller passes NULL, the message may still be committed/published and the function then returns `MARMOT_STATUS_NULL_POINTER`; retrying can duplicate the side effect. Validate every required output pointer before invoking a mutating runtime operation, preferably through a shared preflight helper. The group-creation wrapper has the same ordering problem.

crates/marmot-c/src/commands/push.rs
```diff
@@ -0,0 +1,136 @@
...
+        unsafe {
+            deliver(
+                client.block_on(client.marmot.upsert_push_registration(
+                    account_ref,
+                    platform.to_ffi(),
+                    raw_token,
+                    server_pubkey_hex,
+                    relay_hint,
+                )),
+                out_result,
```
Datawav, line 76: Validate and initialize `out_result` before calling `block_on`. As written, `deliver` checks the output pointer only after `upsert_push_registration` has executed, so a null pointer can return `MARMOT_STATUS_NULL_POINTER` after mutating registration state and encourage an unsafe retry.

crates/marmot-c/src/commands/push.rs
```diff
@@ -0,0 +1,136 @@
...
+    client: *const MarmotClient,
+    account_ref: *const c_char,
+    out_outcome: *mut *mut MarmotPushRegistrationShareOutcome,
+) -> MarmotStatus {
+    ffi_guard(|| {
+        let client = try_arg!(unsafe { client_ref(client) });
+        let account_ref = try_arg!(unsafe { required_str(account_ref) });
+        unsafe {
+            deliver(
+                client.block_on(client.marmot.clear_push_registration(account_ref)),
```
Datawav, line 97: Validate and initialize `out_outcome` before starting this operation. `deliver` checks the pointer only after `clear_push_registration` has executed, so a null pointer can report `MARMOT_STATUS_NULL_POINTER` even though the registration was already cleared, encouraging retries against changed state.

crates/marmot-c/src/types/media.rs
```diff
@@ -0,0 +1,753 @@
...
+            locators
+        };
+        Ok(MediaAttachmentReferenceFfi {
+            locators,
+            ciphertext_sha256: unsafe { required_str(self.ciphertext_sha256.cast_const()) }?,
+            plaintext_sha256: unsafe { required_str(self.plaintext_sha256.cast_const()) }?,
+            nonce_hex: unsafe { required_str(self.nonce_hex.cast_const()) }?,
+            file_name: unsafe { required_str(self.file_name.cast_const()) }?,
+            media_type: unsafe { required_str(self.media_type.cast_const()) }?,
+            version: self.version.to_ffi(),
```
Datawav, line 170: Do not match directly on a caller-populated Rust enum. C can provide an out-of-range discriminant for `MarmotEncryptedMediaVersion`, making `self.version` invalid before this conversion runs and causing undefined behavior. Store a fixed-width integer tag in the input structure, validate it fallibly, and only then construct `EncryptedMediaVersionFfi`.

Datawav (commented):
## Automated review

Outcome: both previously requested boundary-safety changes touched by this delta remain unresolved; the delta introduces two new affected paths.

Previously requested changes:
- Output-pointer preflight remains unresolved in the modified `marmot_upsert_push_registration`.
- Integer validation before constructing caller-controlled Rust enums remains unresolved for `MarmotPushPlatform`.

Newly introduced by this delta:
- High — `marmot_clear_push_registration` now performs the operation before validating its new required output pointer. `commands/push.rs:87-100` calls `block_on(clear_push_registration(...))` before `deliver` checks `out_outcome`. A null pointer can therefore return `MARMOT_STATUS_NULL_POINTER` after clearing has executed, encouraging retries against already-changed state. Validate and initialize `out_outcome` before entering `block_on`, then execute and deliver the result.
- High — `MarmotMediaAttachmentReference.version` is changed from a string to caller-controlled `MarmotEncryptedMediaVersion` (`types/media.rs:13-36,105-113`), and `to_ffi` matches that Rust enum directly at `:45-70,167`. An out-of-range C value creates an invalid Rust enum discriminant before validation, causing undefined behavior. Expose a fixed-width integer tag in the input structure, validate it fallibly, and only then construct `EncryptedMediaVersionFfi`.

Datawav:
CI remediation applied for the monitored head `e1bddc3069aea9e74e9050d83ea7d8e0b0dec518`.

- Failure: Rust clippy rejected two new unsafe C free functions because their docs lacked required `# Safety` sections (`crates/marmot-c/src/types/push.rs:158` and `:186`).
- Fix: added pointer ownership/nullability safety contracts and pushed commit `2eeb3a08` to `c-bindings`.
- Verification: `cargo clippy --all-targets --all-features -- -D warnings` passes locally.
- CI run: https://github.com/marmot-protocol/mdk/actions/runs/31603428046

Datawav (commented):
## Automated review

Outcome: this delta touches none of the previously requested changes and introduces no new actionable findings.

Datawav:
Remediated the stale failing CI at the current PR head by merging `master` into `c-bindings` and synchronizing the C event ABI after upstream removed `GroupEventKindFfi::ForkRecovered`.

Verification before push:
- `cargo check -p marmot-c --all-targets --all-features --locked`
- `cargo test -p marmot-c --features alloc-audit --locked` (96 passed)
- regenerated `crates/marmot-c/include/marmot.h` with cbindgen 0.29.2
- `git diff --check`

Fresh CI is now queued against `a15b4f5e2c139ca85edf2f72180d2080381ca4e1`.

Datawav:
CI remediation applied: regenerated `crates/marmot-c/include/marmot.h` with the workflow-pinned cbindgen 0.29.4 and pushed commit 4520f0c to `c-bindings`.

Root cause: the checked-in header was generated with cbindgen 0.29.2 while the `C bindings smoke` header diff gate installs 0.29.4. Local verification passed: 96 `marmot-c` alloc-audit tests, C11/C++17 header syntax checks, and `git diff --check`. GitHub CI has been retriggered for the new head.

crates/marmot-c/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+/// with `marmot_client_free`.
+pub struct MarmotClient {
+    pub(crate) runtime: tokio::runtime::Runtime,
+    pub(crate) marmot: Arc<Marmot>,
+}
+
+impl MarmotClient {
+    /// Run an async runtime call to completion on the embedded runtime.
+    pub(crate) fn block_on<F: Future>(&self, fut: F) -> F::Output {
+        self.runtime.block_on(fut)
```
Datawav, line 47: `Runtime::block_on` panics when called from a callback already running on this runtime. Subscription callbacks are invoked by spawned runtime tasks, so a callback that sends a reply or invokes another synchronous command returns `MARMOT_STATUS_PANIC_CAUGHT`. Detect runtime-thread reentry and use `block_in_place` with `Handle::block_on`, or dispatch the future and synchronously receive its result without nesting `Runtime::block_on`.

crates/marmot-c/src/commands/message.rs
Datawav, line 36: This performs `send_text` before `deliver` validates `out_summary`. If the caller passes NULL, the message can be published even though the API reports `NullPointer`, inviting a duplicate retry. The subscription `next` wrappers have the same ordering problem and can discard an item. Validate and initialize required output pointers before awaiting, mutating state, or dequeuing anything.

crates/marmot-c/src/commands/notification.rs
```diff
@@ -0,0 +1,135 @@
...
+/// and `error` fields.
+///
+/// # Safety
+/// `client` must be a live handle; `out_collection` a valid pointer. Free
+/// the result with `marmot_background_notification_collection_free`.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn marmot_collect_notifications_after_wake(
+    client: *const MarmotClient,
+    max_wait_ms: u32,
+    source: MarmotNotificationWakeSource,
```
Datawav, line 118: Accepting a `#[repr(C)]` Rust enum by value across a raw C ABI is unsound: foreign callers can supply an unknown discriminant, producing an invalid Rust value before `ffi_guard` can validate it. The other enum inputs and direct Rust `bool` parameters have the same issue for invalid discriminants or non-0/1 bytes. Accept fixed-width integers at the ABI boundary, validate their ranges, then convert to internal enums or booleans.

crates/marmot-c/src/subscriptions.rs
```diff
@@ -0,0 +1,1849 @@
...
+
+    /// Cancel the callback task, if any, at its next await point.
+    fn clear(&self) {
+        let task = self
+            .callback_task
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner)
+            .take();
+        if let Some(task) = task {
+            task.abort();
```
Datawav, line 159: `abort()` is non-blocking, so this teardown can return while the callback task is still invoking the callback with `user_data`. A caller that frees its context after clear/free returns can therefore trigger a use-after-free; retaining it indefinitely leaks memory. Provide synchronized cancellation that waits for callback completion, including callback-thread reentrancy handling, or define a guaranteed terminal callback after which reclamation is safe.

.github/workflows/c-bindings.yml
```diff
@@ -0,0 +1,156 @@
...
+          sudo apt-get install -y valgrind
+          ./crates/marmot-c/c-smoke.sh gcc clang
+
+      - name: Package bundle
+        shell: bash
+        run: |
+          set -euo pipefail
+          tag="$GITHUB_REF_NAME"
+          version="${tag#marmotc-v}"
+          workspace_version="$(sed -n 's/^version = "\(.*\)"/\1/p' Cargo.toml | head -n 1)"
```
Datawav, line 72: The tag-derived `version` is never compared with `workspace_version`. A `marmotc-vX.Y.Z` tag that disagrees with Cargo metadata will produce archive names, a manifest, and a release title advertising one version while Cargo and pkg-config advertise another, corrupting dependency and cache identity. Fail before building unless these versions match exactly.

Datawav (commented):
## Automated review

Outcome: five actionable defects remain in the new C ABI.

- High — Reentrant commands from subscription callbacks panic. `crates/marmot-c/src/lib.rs:44-65` implements synchronous commands with `Runtime::block_on`, while `crates/marmot-c/src/subscriptions.rs:204-231,320-340` invokes callbacks from tasks on that runtime. A callback that sends a reply or calls another normal command attempts nested `block_on` and receives `MARMOT_STATUS_PANIC_CAUGHT`, breaking callback-driven clients. Detect runtime-thread reentry and use `block_in_place` plus `Handle::block_on`, or dispatch to the runtime and synchronously receive the result without nesting `Runtime::block_on`.

- High — Required output pointers are validated after operations consume or mutate state. `crates/marmot-c/src/commands/message.rs:22-38` completes `send_text` before `crates/marmot-c/src/commands/mod.rs:40-58` checks `out_summary`; `crates/marmot-c/src/subscriptions.rs:174-198,289-303` similarly consumes the next subscription item before validating its destination. A null output can therefore publish a message while reporting failure, inviting duplicate retries, or irretrievably discard an event. Validate and initialize every required output pointer at wrapper entry, before awaiting, mutating, or dequeuing anything.

- High — Caller-supplied enums and booleans admit invalid Rust values across the raw ABI. `crates/marmot-c/src/types/notification.rs:18-28` and `types/push.rs:17-24` expose `#[repr(C)]` enums accepted by value at `commands/notification.rs:115-123` and `commands/push.rs:52-71`; Rust `bool` is likewise accepted directly at `commands/account.rs:114`, `chat_list.rs:25`, `notification.rs:52,80`, `group.rs:885`, `media.rs:204`, and `message.rs:312`. Raw or newer foreign bindings can supply an unknown discriminant or a non-0/1 byte, causing undefined behavior before `ffi_guard` can return an error. Accept fixed-width integers at the ABI boundary, validate their ranges explicitly, and convert only validated values to internal enums or booleans.

- High — Subscription teardown provides no safe callback-context reclamation boundary. `crates/marmot-c/src/subscriptions.rs:151-167` removes the handle and calls non-blocking `JoinHandle::abort`, while `:215-225` can still invoke the callback with `user_data` during the cancellation race. A caller freeing its context after clear/free returns can trigger use-after-free; keeping it forever leaks memory. Add a synchronized teardown that cancels and waits for callback completion, with callback-thread reentrancy handling, or emit a guaranteed terminal callback after which `user_data` may be reclaimed.

- Medium — Release tags can disagree with packaged version metadata. `.github/workflows/c-bindings.yml:67-96,126-156` derives archive names, manifest version, and release title from `marmotc-vX.Y.Z` but never compares that value with `workspace_version`; `crates/marmot-c/Cargo.toml:1-6`, `c-bindings.sh:35-36`, and `marmot-c.pc.in:8` package the workspace version. A mislabeled archive can consequently advertise one version while Cargo and pkg-config advertise another, corrupting dependency and cache identity. Fail before building unless the tag-derived version exactly matches the workspace/package version.

crates/marmot-c/src/lib.rs
Datawav, line 47: Synchronous commands invoked from inside a subscription callback always fail. `MarmotClient::block_on` unconditionally calls `Runtime::block_on`, but callback pumps are spawned as tasks on that same multi-thread runtime, so when a host calls any ordinary `marmot_*` command (e.g. `marmot_send_text`) from inside a callback — a natural auto-reply pattern the public docs don't prohibit — tokio panics on the nested `Runtime::block_on` from a runtime worker thread. `ffi_guard` converts that to `MARMOT_STATUS_PANIC_CAUGHT`, documented as "state may be inconsistent; treat as fatal", so a deterministic, always-reproducible callback-driven command failure is reported through the fatal-error channel. Since callback-driven consumers are a headline use case of this PR, the smallest sound fix is to detect runtime context and use `tokio::task::block_in_place(|| Handle::current().block_on(fut))` (or dispatch the future and synchronously await it), keeping direct `block_on` for foreign threads. A regression test should install a callback that invokes a harmless blocking command and assert its normal status instead of PanicCaught.

crates/marmot-c/src/commands/message.rs
Datawav, line 36: Side-effecting operations complete before required out-pointers are validated. Here `client.block_on(client.marmot.send_text(...))` runs to completion, and only afterward does `deliver`/`write_out` discover a NULL `out_summary`. A caller that passes NULL and interprets `MARMOT_STATUS_NULL_POINTER` as "nothing happened" may retry a non-idempotent send and duplicate an externally visible action. The same ordering appears in `marmot_create_group`, all react/reply/edit/delete/upload/mutation wrappers, and `marmot_client_new` itself; the sharpest variant is the blocking `*_next` functions, which consume an item from the stream before `deliver_next` checks the out-pointer — a NULL out silently discards an event that can never be re-read. `deliver_bytes` already preflights its out-pointers, proving the intended pattern exists but is applied inconsistently. Fix: validate every required out-pointer at wrapper entry via a shared `require_out` preflight helper before invoking the runtime operation, keeping the `deliver_*` checks as defense-in-depth. Regression test: mock invocation counter behind a send wrapper and a subscription `next`; call with NULL out, assert NULL_POINTER plus zero invocations / zero consumed items.

crates/marmot-c/src/commands/notification.rs
Datawav, line 118: Caller-supplied Rust `#[repr(C)]` enums crossing the ABI boundary admit invalid values, which is UB before `ffi_guard` can convert anything into an error. `MarmotNotificationWakeSource` is a fieldless Rust enum accepted here by value, and `to_ffi` matches on it infallibly — a caller not using the generated header (hand-declared decls in Zig/Nim/Go/Lua/PHP are explicitly in scope per the PR body) can supply a discriminant outside the valid set from an uninitialized field, wrong width, or a newer variant. The same applies to `MarmotPushPlatform` and to Rust `bool` parameters in seven exported functions (e.g. `marmot_sign_out`'s `delete_key_packages`, `include_archived`, `archived`, `has_limit`, two `enabled` flags, `insecure_local`), plus bool/enum fields in caller-supplied input structs read without validation. This is a soundness hole in exactly the "stable ABI, not a footgun" layer the PR promises, and defeats forward-compat enum handling. Fix: accept `uint32_t`/`uint8_t` at the boundary for all caller-supplied enums and booleans, validate explicitly, and return an InvalidArgument-style status for unknown values, keeping Rust enums internal. Regression test: drive each enum/bool-taking ABI with an out-of-range integer and assert a deterministic error with zero runtime operations.

crates/marmot-c/src/commands/account.rs
```diff
@@ -0,0 +1,559 @@
...
+/// NIP-09 deletions.
+///
+/// # Safety
+/// `client` must be a live handle; `account_ref` a valid string;
+/// `out_outcome` a valid pointer. Free with `marmot_sign_out_outcome_free`.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn marmot_sign_out(
+    client: *const MarmotClient,
+    account_ref: *const c_char,
+    delete_key_packages: bool,
```
Datawav, line 113: Rust `bool` parameters cross the FFI boundary directly, which is UB for any byte value other than 0x00/0x01 — before `ffi_guard` can convert anything into an error. `delete_key_packages: bool` here is one of seven exported functions taking a raw Rust `bool` (`include_archived`, `archived`, `has_limit`, two `enabled` flags, `insecure_local` in `marmot_watch_agent_text_stream`), and caller-supplied input structs also carry bool fields read without validation (`MarmotAuditLogSettings.enabled/data_mode`, `MarmotMediaUploadRequest.send`, `MarmotRelayTelemetrySettings.export_enabled`, `MarmotTimelineMessageQuery.has_*`). A caller with hand-declared bindings passing an uninitialized or wrong-width byte hits undefined behavior at function entry. Fix: accept `uint8_t` at the boundary for all caller-supplied booleans, validate explicitly, and return an error status for values other than 0/1; keep Rust `bool` internal. Regression test: drive each bool-taking ABI with an out-of-range byte (e.g. 0x02 or 0xFF) and assert a deterministic error with zero runtime operations.

crates/marmot-c/src/subscriptions.rs
Datawav, line 159: There is no deterministic way to reclaim callback `user_data`. `SubscriptionCore::clear` takes the pump's `JoinHandle` and calls non-blocking `abort()`; the pump may be between `next().await` resolving and entering the callback when the abort lands, and the terminal-NULL delivery the docs recommend as a reclamation signal is itself skippable by the same abort. So a caller freeing a subscription during shutdown has no API-provided point after which the callback provably cannot run, and freeing `user_data` after `clear`/`free` returns is a use-after-free race — even a caller-side "callback not currently active" check races with a pump holding an item it hasn't delivered yet. A documented hazard is not a safe boundary. Fix: add a synchronous teardown (`*_stop_and_wait`) that aborts and awaits the JoinHandle, with documented reentrancy detection (calling it from inside the callback returns an error rather than deadlocking), or guarantee the terminal-NULL callback is always delivered exactly once before teardown returns. Regression test: block inside a callback, request synchronized teardown from another thread, assert teardown returns only after the callback exits with no invocation afterward, then free `user_data` under valgrind/ASAN.

.github/workflows/c-bindings.yml
Datawav, line 72: A `marmotc-vX.Y.Z` tag can publish artifacts whose metadata advertises a different version. The stage dir, asset zip name, manifest `version`, and release title are all derived from the tag, while `workspace_version` is extracted independently here and never compared against it — yet the compiled crate version and the generated pkg-config `Version:` come from the workspace version (`version.workspace = true` in `crates/marmot-c/Cargo.toml`). Pushing `marmotc-v0.2.0` on a commit whose workspace version is 0.1.0 publishes `marmot-c-linux-x86_64-0.2.0.zip` whose `marmot-c.pc` and crate say 0.1.0, so package managers and downstream caches resolve under a wrong compatibility identity, and `--clobber` re-uploads bake the inconsistency in. (No runtime ABI-version symbol exists, so this is a packaging-identity mismatch, not a compiled-ABI mismatch.) Fix: fail the workflow before building unless the tag-derived version exactly equals the workspace version — the existing "Validate release tag" step is the natural home. Regression test: run the validation step with a mismatched tag/workspace pair and require nonzero exit before any artifact is staged.

Datawav (commented):
## Nightly deep review

All candidate findings are now verified against the patch. I have everything needed for the deep audit.

Deep audit — marmot-protocol/mdk PR #789 ("Add marmot-c C ABI bindings"), head 4520f0c7, 59 files (+24,117). Coverage complete; all candidates below were re-derived from the patches, not the advisory text. CI evidence at capture is incomplete ("Simulator and vectors" in progress), so no test-pass claim is implied.

FINDING 1 — High | crash/ABI contract: synchronous commands invoked from inside a subscription callback always fail
- Evidence: `MarmotClient::block_on` unconditionally calls `Runtime::block_on` (`src/lib.rs:44-48`). Callback pumps are spawned as tasks on that same multi-thread runtime (`src/subscriptions.rs:226-248`, installed via `core.install` at each `*_set_callback`). The runtime is built with `new_multi_thread().enable_all()` and default worker count (`lib.rs`, `marmot_client_new`).
- Trigger: a host calls any ordinary `marmot_*` command (e.g. `marmot_send_text`) from inside a subscription callback — a natural pattern (auto-reply on message event) that the public docs do not prohibit (module docs at `subscriptions.rs:1-35` discuss only user_data lifetime, not re-entry).
- Consequence: tokio panics on nested `Runtime::block_on` from a runtime worker thread; `ffi_guard` converts it to `MARMOT_STATUS_PANIC_CAUGHT` — documented as "state may be inconsistent; treat as fatal." So a deterministic, always-reproducible callback-driven command failure is reported through the fatal-error channel.
- Need/design: callback-driven consumers are a headline use case of the PR; forbidding re-entry silently would be a worse contract. Smallest sound fix: detect runtime context and use `tokio::task::block_in_place(|| Handle::current().block_on(fut))` (or dispatch the future to the runtime and synchronously await the result), keeping direct `block_on` for foreign threads. Alternative: document "no marmot_* calls from callbacks" — but that contradicts the advertised callback API and can't be enforced.
- Regression test: install a callback that invokes a harmless blocking command; assert the command's normal status instead of PanicCaught. Fails today.

FINDING 2 — High | retry safety/dataflow: side-effecting operations complete before required out-pointers are validated
- Evidence: `marmot_send_text` runs `client.block_on(client.marmot.send_text(...))` to completion, then `deliver`/`write_out` discovers a NULL out-pointer (`commands/message.rs:22-38`; `commands/mod.rs:40-58`, `lib.rs:82-90`). Same ordering in `marmot_create_group` (`commands/group.rs:29-53`), all react/reply/edit/delete/upload/mutation wrappers, and `marmot_client_new` itself (builds the runtime and client before `write_out(out_client, ...)`). Sharpest variant: blocking `*_next` first consumes the item from the stream, then `deliver_next` checks the out-pointer (`subscriptions.rs:174-198` + `:289-303`) — a NULL out silently discards an event that can never be re-read.
- Consequence: a caller that passes NULL and interprets `MARMOT_STATUS_NULL_POINTER` as "nothing happened" may retry a non-idempotent send/mutation and duplicate externally visible actions; for subscriptions the event is lost, not duplicated. `deliver_bytes` already preflights its out-pointers (`mod.rs:135-145`), proving the intended pattern exists but is applied inconsistently.
- Fix: validate every required out-pointer at wrapper entry (shared `require_out` preflight helper) before invoking the runtime operation; keep `deliver_*` checks as defense-in-depth.
- Regression test: a mock invocation counter behind a representative send wrapper and a subscription `next`; call with NULL out, assert NULL_POINTER plus zero invocations / zero consumed items.

FINDING 3 — High | memory safety/UB at ABI boundary: caller-supplied Rust `#[repr(C)]` enums and `bool`s admit invalid values from raw FFI
- Evidence: `MarmotNotificationWakeSource` and `MarmotPushPlatform` are fieldless Rust enums accepted by value (`types/notification.rs:18-28`, `commands/notification.rs:115-123`; `types/push.rs:17-24`, `commands/push.rs:52-71`). Rust `bool` parameters cross the boundary directly in 7 exported functions (`account.rs` sign_out `delete_key_packages`, `chat_list.rs` `include_archived`, `group.rs` `archived`, `media.rs`/`message.rs` `has_limit`, `notification.rs` two `enabled` flags, plus `insecure_local` in `marmot_watch_agent_text_stream`), and bool/enum fields appear in caller-supplied input structs whose `to_ffi` reads them without validation (`MarmotAuditLogSettings.enabled/data_mode`, `MarmotMediaUploadRequest.send`, `MarmotMediaAttachmentReference.version`, `MarmotTimelineMessageQuery.has_*`, `MarmotRelayTelemetrySettings.export_enabled`).
- Trigger: any caller not using the generated header — hand-declared decls in Zig/Nim/Go/Lua/PHP are explicitly in scope per the PR body — supplies a byte/value outside the valid set (uninitialized field, wrong width, newer variant).
- Consequence: invalid Rust enum discriminant or invalid `bool` is UB at function entry, before `ffi_guard` can convert anything into an error; this is a soundness hole in the exact "stable ABI, not a footgun" layer the PR promises, and it defeats forward-compat enum handling.
- Fix: accept `uint32_t`/`uint8_t` at the boundary for all caller-supplied enums and booleans, validate explicitly, return InvalidArgument-style status for unknown values; keep Rust enums internal. Output-only enums/bools are unaffected.
- Regression test: drive each enum/bool-taking ABI with an out-of-range integer and assert a deterministic error with zero runtime operations.

FINDING 4 — High | concurrency/lifecycle: no deterministic way to reclaim callback `user_data`
- Evidence: `SubscriptionCore::clear` takes the `JoinHandle` and calls non-blocking `abort()` (`subscriptions.rs:151-167`); the pump may be between `next().await` resolving and entering `callback` when abort lands (`:215-225`), and terminal-NULL delivery is itself skippable by abort. Free paths rely on the same mechanism (`Drop for SubscriptionCore`).
- Consequence: a caller freeing a subscription during shutdown has no API-provided point after which the callback provably cannot run; freeing `user_data` after `clear/free` returns is a use-after-free race. The docs state this honestly, but "documented hazard" ≠ "safe boundary": even a caller-side "callback not currently active" check races with a pump that holds an item and hasn't entered the callback yet. The advertised safe pattern (observe terminal NULL) is unreliable precisely because abort can suppress the terminal call.
- Fix: add a synchronous teardown (`*_stop_and_wait`) that aborts and awaits the JoinHandle — with documented reentrancy detection (calling it from inside the callback returns an error rather than deadlocking) — or guarantee the terminal-NULL callback is always delivered exactly once before teardown returns, giving a reliable reclamation signal. Rename the current non-blocking call to make its semantics unmistakable.
- Regression test: block inside a callback, request synchronized teardown from another thread, assert teardown returns only after the callback exits and no invocation happens afterward; then free `user_data` (valgrind/ASAN clean).

FINDING 5 — Medium | release identity: a `marmotc-vX.Y.Z` tag can publish artifacts whose metadata advertises a different version
- Evidence: the release workflow names the stage dir, asset zip, manifest `version`, and release title from the tag, and independently extracts `workspace_version` without comparing them (`.github/workflows/c-bindings.yml:70-96`, `:126-156`); the compiled crate version and the generated pkg-config `Version:` come from the workspace version (`crates/marmot-c/Cargo.toml` `version.workspace = true`; `c-bindings.sh:35-36`; `marmot-c.pc.in:8`).
- Consequence: pushing `marmotc-v0.2.0` on a commit whose workspace version is 0.1.0 publishes `marmot-c-linux-x86_64-0.2.0.zip` whose `marmot-c.pc` and crate say 0.1.0 — package managers and downstream caches resolve under a wrong compatibility identity, and `--clobber` re-uploads bake the inconsistency in. (Narrowed per cross-review: there is no runtime ABI-version symbol, so this is a packaging-identity mismatch, not a compiled-ABI mismatch; the `--clobber` note is orthogonal.)
- Fix: fail the workflow before building unless the tag-derived version exactly equals the workspace version (the existing "Validate release tag" step is the natural home).
- Regression test: run the validation step with a mismatched tag/workspace pair and require nonzero exit before any artifact is staged.

Discarded as speculative after patch reconciliation: double-free/leak claims (alloc-audit + `free_in_place` discipline checks out, and `deliver` frees the box on `write_out` failure); interior-NUL truncation (strings are stripped, not truncated); status-code renumbering (append-only, i32-fixed); "enums nested in output records" UB (output direction only); Send-impl unsoundness (mirrors own their allocations exclusively, justification holds); agent-stream Option-wrapping anomalies (pump/`deliver_next` types line up). The adversarial pass's "no findings" verdict is overridden by the four verified items above; its memory-model analysis is otherwise consistent with what I see.

Fork/convergence, durable-state, authz, timeouts: this layer delegates all of those to `marmot-uniffi`/runtime unchanged (bearer tokens flow in but are never echoed out — verified in the audit tracker config path); no new qualifying defect there beyond the boundary-safety items above.

Test gaps worth closing regardless: no behavioral test exercises callback→command re-entry (F1), pre-operation out-pointer validation (F2), invalid enum/bool rejection (F3), or teardown synchronization (F4) — the current smoke examples only walk the happy path, and `deliver_bytes_rejects_null_out` tests the post-computation helper, not the ordering defect.

dannym-arx:
Closing this PR and reopening a new one later.

PR #793, opened 2026-07-08
Bound the advisory directory steps during external signer login

mubarakcoded:
`login_external_signer` awaits two advisory steps inline — the existing-account directory preflight and the post-setup directory refresh. Both already proceed on error (their results are best-effort), but neither was bounded in time, so a stalled indexer could hold the whole login hostage.

This caps each at a 10s `ACCOUNT_SETUP_ADVISORY_WAIT`. Measured on an Android device via step instrumentation: cold-cache preflight ran 7.7s (comfortably inside the cap), warm-cache 1.6s; total external-signer login 5-11s. The anti-clobber ordering (preflight completes before relay-list setup) is preserved — a timeout behaves exactly like the existing error path.

Also fixes the Kotlin bindings script under macOS's default bash 3.2, where expanding an empty `FEATURE_ARGS` array with `set -u` aborts the build.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/793">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Login and account import now avoid long stalls by bounding directory discovery/refresh with an advisory time limit, keeping sign-in and setup responsive even when indexing or discovery is slow.

* **Chores**
  * Improved UniFFI/Kotlin and xcframework build scripts to only pass feature flags when provided, avoiding unnecessary empty feature arguments during compilation.

* **Tests**
  * Added coverage to ensure stalled discovery endpoints still complete promptly within the advisory cap.

### DIFF
--- a/crates/marmot-app/src/app_telemetry.rs
+++ b/crates/marmot-app/src/app_telemetry.rs
@@ -6,7 +6,7 @@
 //! key material.
 
 use std::sync::{Arc, Mutex};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use serde::{Deserialize, Serialize};
 use transport_nostr_adapter::{DurationHistogramSnapshot, HistogramBucket};
@@ -24,6 +24,7 @@ pub(crate) enum AppPerformanceOperation {
     AccountOpen,
     AccountCatchUp,
     AccountSync,
+    AccountSetupAdvisoryStep,
     OutboundMessageSend,
     GroupInviteMembers,
     GroupInviteKeyPackageLookup,
@@ -60,6 +61,7 @@ pub struct AppPerformanceSnapshot {
     pub account_open: AppPerformanceOperationSnapshot,
     pub account_catch_up: AppPerformanceOperationSnapshot,
     pub account_sync: AppPerformanceOperationSnapshot,
+    pub account_setup_advisory_step: AppPerformanceOperationSnapshot,
     pub outbound_message_send: AppPerformanceOperationSnapshot,
     #[serde(default)]
     pub group_invite_members: AppPerformanceOperationSnapshot,
@@ -100,6 +102,7 @@ struct AppPerformanceTelemetryInner {
     account_open: AppPerformanceOperationTelemetry,
     account_catch_up: AppPerformanceOperationTelemetry,
     account_sync: AppPerformanceOperationTelemetry,
+    account_setup_advisory_step: AppPerformanceOperationTelemetry,
     outbound_message_send: AppPerformanceOperationTelemetry,
     group_invite_members: AppPerformanceOperationTelemetry,
     group_invite_key_package_lookup: AppPerformanceOperationTelemetry,
@@ -211,6 +214,9 @@ impl AppPerformanceTelemetry {
                 inner.account_catch_up.record(duration, success);
             }
             AppPerformanceOperation::AccountSync => inner.account_sync.record(duration, success),
+            AppPerformanceOperation::AccountSetupAdvisoryStep => {
+                inner.account_setup_advisory_step.record(duration, success)
+            }
             AppPerformanceOperation::OutboundMessageSend => {
                 inner.outbound_message_send.record(duration, success);
             }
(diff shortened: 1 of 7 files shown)


crates/marmot-app/src/runtime/mod.rs
jgmontoya, line 1: This preflight call in `create_or_import_account` is the same unbounded advisory call the PR wraps in `login_external_signer`, and it's on the same "pre-existing identity" gate (`imports_private_key || reactivating_existing || !account.local_signing`). A stalled indexer will hang import/reactivation indefinitely, the exact bug this PR fixes, on a more common path. Should we wrap in timeout there as well?

crates/marmot-uniffi/kotlin-bindings.sh
jgmontoya, line 1: Could you check if the same bug is present and should be fixed in `crates/marmot-uniffi/xcframework.sh` ??

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3089,12 +3089,16 @@ impl AccountManager {
         // its profile from public indexers (its outbox metadata does not live on
         // the app's messaging relays). Runs before the relay-list setup so
         // discovery keeps its anti-clobber ordering.
-        let _ = self
-            .preflight_existing_account_directory(
+        // Discovery is advisory (its error path already proceeds without it),
+        // so a stalled indexer must not hold the whole login hostage.
+        let _ = timeout(
+            ACCOUNT_SETUP_ADVISORY_WAIT,
```
jgmontoya, line 3109: When `preflight_existing_account_directory` errors it logs a `tracing::debug!` with `error_kind`; when this new `timeout` fires, the future is dropped before that log runs, so a hit cap is invisible. Combined with the fact that a timed-out step is only ever backfilled by ambient directory sync (never re-attempted), there's no signal at all that it happened. Consider emitting a debug trace or an `AppPerformanceOperation` sample when the cap fires, so the 10s value can be validated against the fleet rather than the one device it was tuned on. Applies equally to the refresh timeout at (post-change) line 3131.

jgmontoya (commented):
I think it would be good if we had some test for this new behavior

crates/marmot-app/src/runtime/mod.rs
mubarakcoded: Good catch — it's the same gate and the more common path. `94ba288` wraps it with the same bound, and adds an integration test that imports an account whose discovery endpoint accepts TCP and never speaks: the import must complete (and succeed) well inside the cap.

crates/marmot-uniffi/kotlin-bindings.sh
mubarakcoded: Confirmed — same `"${FEATURE_ARGS[@]}"` expansion under `set -u` at three sites in `xcframework.sh`. Guarded in `1cf2104` with the same `${FEATURE_ARGS[@]+...}` idiom.

crates/marmot-app/src/runtime/mod.rs
mubarakcoded: Agreed on both counts. `94ba288` routes all three call sites through a shared `bounded_advisory_step` helper: a hit cap now emits a `tracing::debug!` with the step name and cap, and every run - completed or capped - records an `AppPerformanceOperation::AccountSetupAdvisoryStep` sample (capped = failure), exported alongside the other per-op metrics. So the 10s value is validatable from fleet telemetry, not just the device it was tuned on.

mubarakcoded:
Tests added in `94ba288`: two unit tests on the `bounded_advisory_step` helper (value passes through + success recorded; stalled future capped + failure recorded), and an end-to-end `relay_runtime` test where `create_or_import_account` runs against a stalled discovery endpoint and must complete within the cap. Full `marmot-app` suite: 394 passing.

crates/marmot-app/src/app_telemetry.rs
```diff
@@ -292,6 +299,45 @@ impl AppPerformanceTelemetry {
...
+            );
+            Some(value)
+        }
+        Err(_) => {
+            tracing::debug!(
+                target: "marmot_app::app_telemetry",
+                step,
+                cap_ms = cap.as_millis() as u64,
+                "advisory account-setup step hit its time cap"
+            );
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in `b4c5bc8` (crossed with this review — the tracing audit in CI caught the same thing). The call now carries `method = "bounded_advisory_step"` per the convention that `method` names the emitting function (matching `preflight_existing_account_directory`'s usage), with `step` kept as the variable dimension identifying which advisory step hit the cap, alongside `cap_ms`. The conformance tracing audit passes on the current head.

mubarakcoded:
Valid catch on the outside-diff finding: `create_or_import_account`'s trailing `refresh_user_directory_for_account_id` was still a bare await. Bounded through `bounded_advisory_step` (`import_directory_refresh`) with the same inputs — all four advisory directory awaits across login and import now go through the helper. Clippy clean; telemetry unit tests and the stalled-discovery integration test pass.

jgmontoya (approved):

PR #804, opened 2026-07-09
Feat/incident replay ndjson liveness

jgmontoya:
---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/804">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Incident replay now supports streaming Goggles `goggles-group-export/v1` NDJSON inputs (fail-closed) in addition to `agent-state.json`.
  * Added an epoch-divergence liveness quarantine gate that reports `WentDark` vs `ActiveWhileBehind`.
* **Bug Fixes**
  * Strengthened NDJSON stream validation (manifest/eof handling, completeness, strict protocol enforcement, and per-section count reconciliation).
  * Classification now uses additional exported epoch/liveness signals and improves a prior liveness mislabeling scenario.
* **Documentation**
  * Updated incident-replay adapter documentation and CLI help for both input formats and the new gate behavior.
* **Tests**
  * Added new JSON/NDJSON fixtures and expanded classification and stream-parse test coverage.
* **Chores**
  * Updated `.gitignore` to ignore local `.env`.

### DIFF
--- a/crates/incident-replay/src/classify.rs
+++ b/crates/incident-replay/src/classify.rs
@@ -4,8 +4,52 @@
 //! this verdict, so a healthy export yields zero vectors and a clean exit rather
 //! than a crash. Built incrementally, one rule per behaviour.
 
+use std::collections::BTreeMap;
+use std::fmt;
+
 use crate::export::{AgentStateExport, EventKind};
 use serde::Serialize;
+
+/// How many epochs an engine may trail the group's epoch high-water mark
+/// before it counts as left behind. The value is derived, not tuned. Lag 1 is
+/// the noise floor: every commit leaves every other engine one epoch behind
+/// until it propagates, so a snapshot routinely catches healthy engines there —
+/// indistinguishable from a commit in flight, and so carrying no information.
+/// Lag 2 is the smallest lag that *cannot* be one in-flight commit: the engine
+/// missed a commit and the group then produced another it also missed (e.g. two
+/// members invited back-to-back), so the gap provably outlived an inter-commit
+/// interval. Being event-anchored rather than clock-anchored, it means the same
+/// in a chatty group and a weekly one — the property that disqualified
+/// wall-clock staleness. Any higher threshold is the arbitrary one: "one
+/// survived commit is fine but two are not" has no structural answer, whereas 2
+/// is just the noise floor plus one. A commit burst can still park healthy
+/// engines at lag 2 for its propagation window — the claim bounds transients,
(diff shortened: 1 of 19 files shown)


crates/incident-replay/src/classify.rs
```diff
@@ -4,9 +4,25 @@
...
+use std::collections::BTreeMap;
+use std::fmt;
+
 use crate::export::{AgentStateExport, EventKind};
 use serde::Serialize;
 
+/// How many epochs an engine may trail the group's epoch high-water mark
+/// before it counts as left behind. One epoch behind is routine commit
+/// propagation; two or more means consecutive commits never arrived.
+const EPOCH_DIVERGENCE_MIN_LAG: u64 = 2;
```
dannym-arx, line 35: I'm not so sure about this, it's possible to be 2 epochs behind and still recover... e.g. 2 people get invited while you're offline, that's 2 epochs

jgmontoya: You're right, two behind is recoverable, and the comment's "consecutive commits never arrived" claim was wrong: a device offline through two quick invites drains both commits on reconnect and misses nothing. Fixed in the latest push.

The threshold itself survives, though, and for a structural reason rather than a tuned one. Lag 1 is the noise floor: every commit leaves every other engine one behind until it propagates, so a snapshot routinely catches healthy engines there, it carries no information. Lag 2 is the smallest lag that cannot be a single in-flight commit: the engine missed a commit and the group then produced another it also missed, so the gap provably outlived an inter-commit interval, event-anchored, so it means the same in a chatty group and a weekly one. Any higher cutoff is the arbitrary one ("one survived commit is fine but two are not" has no answer). I also backtested it against hourly slices of the real incident export as a falsification check: lag 1 fires on the noise floor, lag 3 loses half the incident, both stuck devices sat exactly two behind the tip, and 2 held.

For your scenario specifically: an offline-through-a-burst device shows up labeled went_dark on that one pull and self-resolves on the next (measured, half the engines showed exactly that overnight shape, and none was ever labeled active_while_behind). So a single quarantine reads as "re-pull," not "broken"; persistence across pulls is the signal. That's now spelled out in the constant docs and the crate AGENTS.md, along with why the softer arming variants were rejected, the incident's genuinely-stuck device read as dark in the export, so arming only on active evidence would have missed the real thing. Also added two tests pinning the flip side: untimed evidence never arms the gate.

dannym-arx (approved):

PR #812, opened 2026-07-10
Add draft persistence and profile website bindings

erskingardner:
## What changed

- add per-account SQLCipher composer draft persistence with list, load, upsert, and delete APIs
- preserve draft text, reply targets, ordered attachment bytes, and attachment presentation metadata
- expose draft records and commands through MarmotKit/UniFFI for Swift and Kotlin clients
- expose cached Nostr kind 0 website metadata through a read-only profile accessor
- bump the workspace to 0.9.4 and update conformance fixture version stamps
- add an Unreleased changelog entry and FFI validation coverage

## Why

White Noise clients need durable composer drafts at the shared encrypted core boundary, including attachment drafts, and profile surfaces need access to the cached website field without widening the writable profile record.

## Storage and compatibility

- migration 0026 creates draft and draft-attachment tables in each account's encrypted SQLCipher database
- draft rows are group-scoped and cascade when the owning group is deleted
- all new MarmotKit APIs are additive

## Validation

- `cargo test -p storage-sqlite` (197 passed)
- `cargo test -p marmot-uniffi` (51 unit + 20 smoke passed)
- `cargo test -p marmot-app` (318 unit + 79 integration passed)
- canonical vector fixture regeneration check
- `just fast-ci` including formatting, workspace checks, clippy, and OTLP feature builds
- Swift/iOS XCFramework generation
- Kotlin bindings plus Android JNI builds for arm64-v8a, armeabi-v7a, x86, and x86_64

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/812">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added encrypted composer draft storage with support for saving, editing, viewing, and deleting draft text, reply references, and media attachments.
  * Added access to cached profile website information.
* **Documentation**
  * Updated the changelog to document composer draft storage and profile website access.
* **Chores**
  * Updated the application and conformance release version to 0.9.4.
  * Refreshed conformance metadata for existing validation scenarios.


crates/storage-sqlite/src/message_drafts.rs
```diff
@@ -0,0 +1,306 @@
...
+    pub group_id_hex: String,
+    pub content: String,
+    pub reply_to_message_id_hex: Option<String>,
+    pub media_attachments: Vec<StoredMessageDraftAttachment>,
+    pub created_at_ms: i64,
+    pub updated_at_ms: i64,
+}
+
+impl SqliteAccountStorage {
+    pub fn message_drafts(&self) -> StorageResult<Vec<StoredMessageDraft>> {
```
erskingardner: `message_drafts()` loads the **full `plaintext` BLOB of every attachment across every draft** into memory in one shot, then moves each `Vec<u8>` through three record layers (`Stored* → MessageDraft* → *Ffi`) and marshals it over UniFFI. A draft *list* is almost always used to answer "which conversations have unsent drafts / show a preview" — it rarely needs raw media bytes. With several drafts each holding image/video attachments this is an O(all drafts × all media bytes) allocation on the hot path.

Suggest the list return metadata only (text, reply target, attachment count / total size / file names) and fetch `plaintext` lazily via `message_draft(group_id)` when a specific composer is actually restored.

crates/storage-sqlite/src/message_drafts.rs
```diff
@@ -0,0 +1,306 @@
...
+                .query_map([], |row| row.get::<_, String>(0))
+                .storage()?
+                .collect::<Result<Vec<_>, _>>()
+                .storage()?
+        };
+
+        group_ids
+            .into_iter()
+            .map(|group_id_hex| {
+                load_message_draft(&conn, &group_id_hex)?.ok_or_else(|| {
```
erskingardner: This error branch is unreachable. `self.lock()` at the top of `message_drafts()` holds the connection lock for the entire body, so no writer can delete a row between the `SELECT group_id_hex` and each `load_message_draft`. The `"message draft disappeared while loading"` case can't happen, which also means it can't be covered by a test. Either drop it (e.g. `filter_map` the `None` away) or add a comment explaining why the defensive branch is intentionally kept.

crates/storage-sqlite/src/migrations/0026_message_drafts.rs
```diff
@@ -0,0 +1,41 @@
...
+    attachment_id TEXT NOT NULL,
+    file_name TEXT NOT NULL,
+    media_type TEXT NOT NULL,
+    plaintext BLOB NOT NULL,
+    dim TEXT,
+    thumbhash TEXT,
+    duration_seconds REAL,
+    waveform_samples_json TEXT NOT NULL DEFAULT '[]',
+    PRIMARY KEY (group_id_hex, position),
+    UNIQUE (group_id_hex, attachment_id),
```
erskingardner, line 35: Two client-supplied inputs surface here as opaque `StorageError::Backend` strings rather than typed validation errors:

- Two attachments sharing an `attachment_id` trip this `UNIQUE (group_id_hex, attachment_id)` mid-transaction → the whole save fails with a raw `"UNIQUE constraint failed"`.
- A non-finite `waveform_samples` value (NaN/Infinity) makes `serde_json::to_string` in `save_message_draft` error out with a raw serde message (note `duration_seconds` is a plain REAL column and would accept it, so the two float paths are inconsistent).

Both `id` and `waveform_samples` come straight from the client over FFI. Consider validating/deduping at the boundary and returning a typed error so callers can distinguish "bad input" from a real storage failure.

crates/marmot-app/src/drafts.rs
```diff
@@ -0,0 +1,134 @@
...
+        &self,
+        account_ref: &str,
+        group_id_hex: &str,
+        content: &str,
+        reply_to_message_id_hex: Option<&str>,
+        media_attachments: Vec<MessageDraftAttachment>,
+    ) -> Result<MessageDraft, AppError> {
+        let account = self.account_home().account(account_ref)?;
+        self.ensure_account_state(&account.label)?;
+        self.group(&account.label, group_id_hex)?
```
erskingardner: Coverage gap. This `UnknownGroup` guard and the entire app→FFI happy path are untested. The only round-trip test lives in `storage-sqlite`; the uniffi `smoke.rs` test exercises **only** the invalid-hex and missing-account error branches, so `save_message_draft` never returns `Ok` under test and `From<MessageDraft> for MessageDraftFfi` is never actually constructed. For a repo that gates on coverage, please add a positive round-trip through the runtime/uniffi layer — save → list → load → delete, plus the `UnknownGroup` case — so the success conversions and this guard are exercised.

erskingardner (commented):
Adversarial review of the draft-persistence + profile-website work. Mechanics are solid — I verified `PRAGMA foreign_keys=ON` and `secure_delete=ON` at connection open, so the `ON DELETE CASCADE` chain (account_groups → message_drafts → message_draft_attachments) actually fires and deleted plaintext pages get zeroed. The UPSERT correctly preserves `created_at_ms` while bumping `updated_at_ms`, hex is canonicalized to lowercase before it hits the reply-to CHECK constraint, and everything is parameterized. No correctness bug that corrupts data.

My concerns are about **API shape at scale** and **test coverage**, plus a couple of robustness nits — inline. Two cross-cutting points:

1. **Drafts store full plaintext attachment bytes and never expire.** A draft is only removed on explicit delete or group deletion, and each attachment's entire decrypted body lives in the row as a BLOB. Combined with the list API (see inline on `message_drafts()`), the encrypted DB can grow without bound and a single list call can materialize every draft's media at once. Worth a size/count budget, or at least a documented expectation of caller-side pruning.

2. **The group-existence guard is TOCTOU-racy and only partially typed.** `save_message_draft` does `self.group(...).ok_or(UnknownGroup)` and then relies on the FK for the real enforcement. If the group is deleted between the check and the insert, the caller gets a raw `StorageError::Backend("FOREIGN KEY constraint failed")` instead of `UnknownGroup`. Low severity given the storage lock, but the two enforcement paths return different error types for the same condition.

Nothing here is blocking on its own; the coverage gap is the one I'd want addressed before merge given this repo gates on coverage.

PR #830, opened 2026-07-12
fix: replace opencode wall-clock timeout with idle timeout

### LINKED ISSUE #829: wn-opencode: replace wall-clock timeout with idle timeout
## Problem

`wn-opencode` today wraps the whole OpenCode stdout-reading loop in a single wall-clock `tokio::time::timeout(WN_OPENCODE_TIMEOUT_SECS, ...)` (`opencode.rs:99`, default 300s). This is a hard end-to-end ceiling on any single OpenCode invocation.

When the deadline expires:
- The child is SIGKILL'd (`opencode.rs:152`).
- The runner returns `HarnessError::OpencodeTimedOut`.
- The group sees: `[wn-opencode] opencode timed out before producing a complete response.`
- Session id is not persisted if this was the first prompt in a new group (`finish_success` is not called on the error path in `bridge.rs`).

The timer never resets. A prompt that streams useful text for 4:59 and keeps producing for another second gets killed and the operator sees "timed out" even though OpenCode was actively working.

## Root causes this masks

On a 5-minute clock, "timed out" can mean any of:

1. Real work taking longer than 5 minutes (large repo + tool calls + tests).
2. Model provider latency or rate-limiting.
3. Permission prompt pending in OpenCode (`wn-opencode` has no approval path today — the child sits idle until the deadline).
4. Long tool-only sequences that never emit `text` parts (the harness only wakes on `text` events).
5. Genuine wedge (deadlocked child, stuck HTTP).

Only (5) is what an "opencode is stuck" signal should catch. Cases (1)–(4) are current false positives.

## Proposed fix — modular, backwards-compatible

Replace the outer wall-clock timeout with a per-line idle timeout, and keep an optional total-cap for genuine wedges.

### Behavior changes

- **Idle timeout** (`WN_OPENCODE_IDLE_TIMEOUT_SECS`, default `120`): reset every time a stdout line is read (any JSON line, not only `text`). Fires only when OpenCode has been silent that long.
- **Total cap** (`WN_OPENCODE_TIMEOUT_SECS`, default raised to `3600`): kept as a safety ceiling. Old default (300) becomes obsolete; the new default is intentionally generous because idle timeout catches the "stuck" case.
- On idle-timeout expiry, the message text should distinguish the case: `[wn-opencode] opencode went silent for Ns without producing output; killing the invocation.` Total-cap expiry keeps the current message.

### Implementation shape

In `opencode.rs`, replace the current outer `timeout(...)` wrapping the whole `while let Some(line) = ...` loop with:

```rust
let mut idle = tokio::time::timeout(idle_deadline, lines.next_line());
let total = tokio::time::sleep(total_deadline);
tokio::pin!(total);
loop {
  tokio::select! {
    _ = &mut total => { child.kill().await?; return Err(HarnessError::OpencodeTimedOut); }
    line_result = &mut idle => {
      match line_result {
        Err(_elapsed) => { child.kill().await?; return Err(HarnessError::OpencodeIdle); }
        Ok(Ok(None)) => break, // EOF
        Ok(Ok(Some(line))) => { /* existing parse_event_line handling */ }
        Ok(Err(_)) => return Err(HarnessError::OpencodeStream),
      }
      idle = tokio::time::timeout(idle_deadline, lines.next_line());
    }
  }
}
```

### Session persistence on timeout

Secondary fix, same PR: persist the observed session id even on timeout, so a retry does not restart a fresh OpenCode session. `bridge.rs:412` matches on `HarnessError::OpencodeTimedOut` today; the `outcome.observed_session` is thrown away because `finish_success` is not called. Consider extracting session persistence out of `finish_success` and running it on any terminal outcome that saw a `step_start`.

### Config

- Add `WN_OPENCODE_IDLE_TIMEOUT_SECS` (default 120) to `config.rs` and the README env table.
- Raise `WN_OPENCODE_TIMEOUT_SECS` default from 300 to 3600.
- Document that the wall-clock cap is now a wedge-recovery ceiling, not a normal-work budget.

### Error taxonomy

Add `HarnessError::OpencodeIdle` alongside `OpencodeTimedOut`. Match it in `bridge.rs:412` with a distinct user-facing message.

## Non-goals

- No change to how prompts are dispatched (still `opencode run` argv).
- No dependency on the `opencode serve` pivot in #792. This fix stands alone and is compatible with the pivot; if #792 lands and moves observation to SSE, the same idle/total pair can apply to bus-event arrival.
- No progress messages surfaced to the group.

## Test plan

1. Unit-level: mock a stdout that emits a line every 90s for 20 minutes; verify no timeout under defaults (idle=120, total=3600).
2. Unit-level: mock a stdout that goes silent for 3 minutes; verify `OpencodeIdle` fires at ~120s and SIGKILLs the child.
3. Unit-level: mock a stdout that streams every 60s for over 60 minutes; verify total-cap fires at 3600s with existing `OpencodeTimedOut`.
4. Integration: existing `just opencode-dev-e2e-connector` should still pass with idle default; add a case for the silent-child path.
5. Session persistence: prove that a first-prompt idle timeout still records the observed session id so the next prompt continues instead of starting fresh.

## Files touched

- `integrations/opencode/marmot/src/opencode.rs` (loop shape, new error variant)
- `integrations/opencode/marmot/src/error.rs` (`OpencodeIdle` variant)
- `integrations/opencode/marmot/src/bridge.rs` (match on new variant, session persistence on error path)
- `integrations/opencode/marmot/src/config.rs` (new env)
- `integrations/opencode/marmot/README.md` (env table)

## Related

- #792 opencode: session UX v1 — product outline (later architectural pivot; this fix is intentionally independent so it can ship immediately).

agent-p1p:
Fixes #829

## Summary
- Replace the fixed five-minute stdout timeout with a 120-second idle timeout reset by every OpenCode stdout line.
- Retain a one-hour total safety cap across streaming, child exit, and stderr capture.
- Preserve newly observed OpenCode session IDs on idle, total-timeout, and stream failures without overwriting established sessions.
- Update environment defaults, installer output, documentation, and deterministic process fixtures.

## Verification
- `cargo test -p wn-opencode --locked` (40 passed, 1 ignored)
- 10 consecutive full `wn-opencode` test-suite runs passed
- `cargo clippy -p wn-opencode --all-targets --locked -- -D warnings`
- `just fast-ci`
- `integrations/opencode/marmot/tests/test_installer.sh`
- shell syntax checks for installer and mock fixture
- independent pre-commit review: passed

## Sensitive paths
None.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/830">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable idle detection for OpenCode runs, stopping executions that produce no output for the specified period.
  * Extended the overall execution safety limit to support longer-running tasks while ongoing output continues.
  * Improved session preservation when runs fail or time out.

* **Bug Fixes**
  * Ensured processes are properly cleaned up after timeout or stream failures.
  * Preserved sessions when output backpressure causes a run to fail.

### DIFF
--- a/integrations/opencode/marmot/Cargo.toml
+++ b/integrations/opencode/marmot/Cargo.toml
@@ -22,3 +22,4 @@ tracing-subscriber.workspace = true
 
 [dev-dependencies]
 tempfile.workspace = true
+tokio = { workspace = true, features = ["test-util"] }
--- a/integrations/opencode/marmot/README.md
+++ b/integrations/opencode/marmot/README.md
@@ -90,7 +90,8 @@ Configure with environment variables:
 | `WN_OPENCODE_ADMIN_HEX` | unset | Legacy alias for `WN_OPENCODE_ALLOWED_SENDERS_HEX` |
 | `WN_OPENCODE_ACCOUNT_ID_HEX` | first local account | Specific `wn-agent` account to use |
 | `WN_OPENCODE_BIN` | `opencode` | OpenCode binary or executable path |
-| `WN_OPENCODE_TIMEOUT_SECS` | `300` | Hard timeout for each OpenCode invocation |
+| `WN_OPENCODE_IDLE_TIMEOUT_SECS` | `120` | Kill the invocation when OpenCode produces no stdout line for this long |
+| `WN_OPENCODE_TIMEOUT_SECS` | `3600` | Generous total safety cap for wedge recovery; ongoing output resets only the idle timer |
 | `WN_OPENCODE_REQUEST_TIMEOUT_SECS` | `30` | Timeout for each control-socket request |
 | `WN_OPENCODE_MAX_REPLY_BYTES` | `30000` | UTF-8 byte limit for each durable Marmot reply chunk |
 | `WN_OPENCODE_MAX_PENDING_PER_GROUP` | `4` | Per-group in-flight/queued prompt cap |
--- a/integrations/opencode/marmot/src/bridge.rs
+++ b/integrations/opencode/marmot/src/bridge.rs
@@ -13,7 +13,7 @@ use crate::chunking::split_reply_chunks;
 use crate::config::{Config, dirs_home};
 use crate::control::ControlClient;
 use crate::error::{HarnessError, Result};
-use crate::opencode::{Invocation, Outcome, RunnerEvent};
+use crate::opencode::{Invocation, Outcome, RunFailure, RunnerEvent};
 use crate::repo_picker::{parse_repo_picker, resolve_repo, validate_session_cwd};
 use crate::store::{SessionRecord, SessionStore};
 
@@ -343,6 +343,7 @@ async fn handle_message(ctx: Arc<BridgeContext>, inbound: InboundPrompt, permit:
     let invocation = Invocation {
         bin: ctx.cfg.opencode_bin.clone(),
         timeout: ctx.cfg.opencode_timeout,
+        idle_timeout: ctx.cfg.opencode_idle_timeout,
         cwd: cwd.clone(),
         session_id,
         prompt,
@@ -402,7 +403,26 @@ async fn handle_message(ctx: Arc<BridgeContext>, inbound: InboundPrompt, permit:
             )
             .await;
         }
-        Ok(Err(err)) => {
+        Ok(Err(RunFailure {
+            error: err,
+            observed_session,
+        })) => {
+            if let Err(store_err) = persist_observed_session_if_unset(
+                &ctx.sessions,
+                &inbound.group_ref,
+                known_session.as_ref(),
+                cwd,
+                observed_session,
+            )
+            .await
+            {
+                warn!(
+                    target: TRACE_TARGET,
+                    method = "session_store",
+                    error_kind = store_err.privacy_safe_kind(),
+                    "failed to persist opencode session"
+                );
+            }
             warn!(
                 target: TRACE_TARGET,
                 method = "opencode_run",
@@ -410,20 +430,25 @@ async fn handle_message(ctx: Arc<BridgeContext>, inbound: InboundPrompt, permit:
                 "opencode invocation failed"
             );
             let text = match err {
+                HarnessError::OpencodeIdle => format!(
+                    "[wn-opencode] opencode went silent for {}s without producing output; killing the invocation.",
+                    ctx.cfg.opencode_idle_timeout.as_secs()
+                ),
                 HarnessError::OpencodeTimedOut => {
                     "[wn-opencode] opencode timed out before producing a complete response."
+                        .to_owned()
                 }
                 HarnessError::OpencodeSpawn => {
-                    "[wn-opencode] failed to start opencode; check WN_OPENCODE_BIN."
+                    "[wn-opencode] failed to start opencode; check WN_OPENCODE_BIN.".to_owned()
                 }
-                _ => "[wn-opencode] opencode failed while streaming its response.",
+                _ => "[wn-opencode] opencode failed while streaming its response.".to_owned(),
             };
             let _ = send_reply(
                 &ctx,
                 &inbound.account_ref,
                 &inbound.group_ref,
                 &inbound.message_ref,
-                text,
+                &text,
                 0,
             )
             .await;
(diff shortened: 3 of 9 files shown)


agent-p1p:
## Adversarial review

**Verdict: changes required** — 1 blocking finding, 2 suggestions.

### Blocking

1. **`integrations/opencode/marmot/src/opencode.rs:134-138,185-195` — stdout EOF bypasses the idle timeout and can wedge a group for the full one-hour cap.**

   `Ok(Ok(None))` breaks out of the idle-timed read loop. The subsequent `child.wait()` and stderr join remain protected only by `timeout_at(total_deadline, ...)`. If OpenCode closes stdout but remains alive, it is permanently unable to produce another stdout line, yet the default behavior waits up to 3600 seconds instead of declaring the invocation idle after 120 seconds. `handle_message` holds the per-group serial lock for this whole period.

   The new `run_total_cap_includes_child_wait_after_stdout_closes` test currently codifies this bug by making the total deadline shorter than the idle deadline and expecting `OpencodeTimedOut`. Please keep idle detection active through the post-EOF child wait (while retaining the total cap), and add the converse regression: idle deadline shorter than total, stdout closes while the child stays alive, expect `OpencodeIdle` promptly.

### Suggestions

1. **`opencode.rs:440-451` — make the central streaming regression fail under the old wall-clock implementation.** The fixture runs for about 750ms while `mock_invocation` gives it a 10s total timeout, so the test would also pass before this PR. Prefer paused Tokio time and exercise output beyond the old 300s wall clock but below the new 3600s cap, with every line less than 120s apart.

2. **`bridge.rs:406-454,870-907` — cover first-prompt idle persistence through the bridge error path.** The current bridge test obtains a backpressure-induced total timeout and then invokes `persist_observed_session_if_unset` directly. It does not prove that the `handle_message` idle-error arm persists the observed session or emits the distinct idle message. A focused helper-level test is sufficient; no broad harness abstraction needed.

### Out-of-scope follow-ups

- #831 tracks terminating OpenCode descendant tool processes, not only the immediate child PID.
- #832 tracks restarting already-active Linux systemd services after installer upgrades.

### Verification

- `cargo test -p wn-opencode --locked`: 40 passed, 1 ignored.
- `cargo clippy -p wn-opencode --all-targets --locked -- -D warnings`: passed.
- `cargo fmt --all --check`: passed.
- Installer test and shell syntax checks: passed.
- Commit signature and `agent-p1p` attribution verified.
- Sensitive paths touched: none.

agent-p1p:
Addressed the blocking EOF lifecycle finding in d733ea8.

- The idle deadline now remains active after stdout EOF through both child exit and stderr capture; EOF does not reset the remaining budget.
- Each actual stdout line still starts a fresh idle interval.
- Added regressions for a live child after EOF, delayed EOF near the idle deadline, long-running output beyond the old wall-clock cap, and run-failure session persistence/reply selection.

Verification:
- `cargo test -p wn-opencode --locked` — 42 passed, 1 ignored
- `cargo clippy -p wn-opencode --all-targets --locked -- -D warnings`
- `cargo fmt --all --check`
- fixture `bash -n`
- independent pre-commit review — passed

A repository-wide `just fast-ci` attempt reached the workspace OTLP clippy step, then the host ran out of disk; focused crate checks above are green and GitHub CI is running on the pushed commit.

agent-p1p:
## Re-review after EOF lifecycle fix and rebase

**Verdict: prior blocking finding resolved. No remaining blocking findings.**

I independently re-reviewed the fix on rebased head `e532707` rather than relying on the author's follow-up claim.

### Original blocker

- **Resolved:** the idle deadline no longer disappears after stdout EOF. `opencode.rs` now retains the current absolute idle deadline through `child.wait()` and stderr capture, while the outer `timeout_at(total_deadline, ...)` independently preserves the one-hour total cap.
- EOF does not reset the remaining idle budget.
- Every actual stdout line resets the idle deadline, including non-text JSON lines.
- Idle/total failures abort stderr capture and kill/reap the immediate child.

### Regression evidence

- `run_idle_timeout_fires_after_stdout_eof_with_live_child` proves idle wins when stdout closes and the child remains alive.
- `run_eof_keeps_remaining_idle_budget` proves EOF does not start a fresh idle window.
- `run_total_cap_includes_child_wait_after_stdout_closes` proves the total cap still wins when shorter.
- `stdout_lines_reset_idle_past_old_wall_clock_deadline` advances paused time beyond the old 300-second ceiling while keeping every gap below the idle limit.
- `run_failure_path_persists_first_session_and_selects_idle_reply` exercises the bridge failure helper used by the live error arm and verifies both first-session persistence and the distinct idle response.

### Verification on `e532707`

- Rebased cleanly onto current `master` `7f6fd44`; `HEAD~2` equals `origin/master`.
- `cargo test -p wn-opencode --locked`: 42 passed, 1 ignored.
- `cargo clippy -p wn-opencode --all-targets --locked -- -D warnings`: passed.
- `cargo fmt --all --check`: passed.
- Installer test and both shell syntax checks: passed.
- `just fast-ci`: passed, including workspace and OTLP-feature checks/clippy.
- Rebased GitHub CI: all jobs green.
- Both rebased commits have valid `agent-p1p` SSH signatures.
- No crypto, MLS, CGKA, key, trust, auth, or storage-sensitive paths changed.

The earlier blocking review is cleared. From a code-correctness and CI perspective, this is ready for maintainer approval/merge.

Issue #832, opened 2026-07-12
OpenCode installer: restart active systemd services after upgrade


agent-p1p:
## Problem

On Linux, rerunning `install-opencode-marmot.sh` overwrites the binaries and service units, runs `systemctl --user daemon-reload`, then calls `systemctl --user enable --now`. `enable --now` starts an inactive unit but does not restart an already-active one.

An existing `wn-opencode`/`wn-agent` process therefore keeps running the old executable and old environment after an apparent successful upgrade. This is visible in #830 because a rerun writes the new idle/total timeout environment while the active harness can continue using the old 300-second behavior until manual restart or reboot. The macOS path already bootouts and bootstraps its LaunchAgent.

## Expected behavior

After installing/updating and reloading units, restart active systemd user services (while preserving the explicit no-start options). Add an installer test that distinguishes a fresh enable/start from an active-service upgrade.

## Scope note

Found while reviewing #830. The upgrade lifecycle bug predates that PR and affects all installer-delivered changes.

PR #833, opened 2026-07-12
test(installer): exercise systemd service activation

agent-p1p:
## Summary
- rebase onto current `master` after the production restart behavior landed through #1043
- retain only the Linux execution fixture with mocked `systemctl` calls
- cover fresh installs, active-service upgrades, and `--no-start-wn-opencode`

## Test plan
- Linux container: `integrations/opencode/marmot/tests/test_installer.sh`
- `bash -n scripts/install-opencode-marmot.sh integrations/opencode/marmot/tests/test_installer.sh`
- `just fast-ci`

Follow-up regression coverage for #832; the production fix is already on `master`.

## Summary by CodeRabbit

* **Tests**
  * Added Linux integration coverage for installer service management.
  * Verified service behavior during initial installation, upgrades, and runs that disable starting the OpenCode service.
  * Added checks to confirm expected systemd actions occur and unintended actions are prevented.

### DIFF
--- a/integrations/opencode/marmot/tests/test_installer.sh
+++ b/integrations/opencode/marmot/tests/test_installer.sh
@@ -8,6 +8,152 @@ allow_hex="$(printf '11%.0s' {1..32})"
 [ -x "$installer" ]
 bash -n "$installer"
 
+assert_log_contains() {
+    local log_file="$1"
+    local expected="$2"
+    local line
+    while IFS= read -r line; do
+        [ "$line" != "$expected" ] || return 0
+    done <"$log_file"
+    echo "missing systemctl call: $expected" >&2
+    return 1
+}
+
+assert_log_excludes() {
+    local log_file="$1"
+    local unexpected="$2"
+    local line
+    while IFS= read -r line; do
+        if [ "$line" = "$unexpected" ]; then
+            echo "unexpected systemctl call: $unexpected" >&2
+            return 1
+        fi
+    done <"$log_file"
+}
+
+run_linux_service_case() {
+    local fixture_root="$1"
+    local active="$2"
+    local log_file="$3"
+    local mock_bin="$fixture_root/mock-bin"
+    shift 3
+
+    : >"$log_file"
+    SYSTEMCTL_ACTIVE="$active" \
+    SYSTEMCTL_LOG="$log_file" \
+    HOME="$fixture_root/home" \
+    MARMOT_HOME="$fixture_root/marmot-home" \
+    MARMOT_INSTALL_PREFIX="$fixture_root/install" \
+    PATH="$mock_bin:/usr/bin:/bin" \
+    WN_AGENT_SHA="9.9.9" \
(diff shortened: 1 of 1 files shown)


agent-p1p (commented):
Adversarial review of 8e1d4ef1eacf515939512c8f96c56812265dd6aa

BLOCKING:
- none

SUGGESTION:
- Dry-run output no longer describes the upgrade path: `scripts/install-opencode-marmot.sh:494,535` still says the installer would run `enable --now` unconditionally. A real active-service upgrade now runs `enable` plus `restart`. Since dry-run intentionally avoids the state probe, wording such as “enable/start, or restart if already active” would keep the preview accurate.
- Guard the explicit no-start behavior in the new active-service fixture: `integrations/opencode/marmot/tests/test_installer.sh:135-142` exercises the active default path, while `--no-start-wn-opencode` is covered only by dry-run text. An active fixture with `--no-start-wn-opencode` that asserts no `restart wn-opencode.service` call would lock in the PR’s preservation claim.

NITPICK:
- none

FOLLOW-UP:
- Filed #834 for the same pre-existing upgrade bug in the Hermes and OpenClaw installers. This is out of scope for #832/#833: https://github.com/marmot-protocol/mdk/issues/834

SUMMARY: The change is small and correct. After unit rewrite and `daemon-reload`, active `wn-agent` and `wn-opencode` user services take the `enable` + `restart` path and therefore load the new executable/environment; inactive services retain `enable --now`. Existing `--no-service`, `--no-start-wn-agent`, and `--no-start-wn-opencode` gates remain upstream. The deterministic mock uses exact command-line assertions to distinguish fresh and active upgrades. No security/privacy regressions or sensitive paths were found.

VERDICT: APPROVE
SENSITIVE_PATHS: none

Verified locally:
- `integrations/opencode/marmot/tests/test_installer.sh` — passed
- `bash -n scripts/install-opencode-marmot.sh integrations/opencode/marmot/tests/test_installer.sh` — passed

agent-p1p:
Addressed both adversarial-review suggestions in 9b59c8c: Linux dry-run output now describes enable/start versus active-service restart, and the active-service fixture verifies --no-start-wn-opencode performs no systemd probe, enable, or restart for wn-opencode. Local installer and shell-syntax checks pass.

agent-p1p:
@coderabbitai review

erskingardner:
I compared this head carefully with the implementation merged through #1043.

The production change is superseded: `master` has the same active-service branch (`systemctl --user enable` + `restart`) and the same inactive-service branch (`enable --now`) for both `wn-agent` and `wn-opencode`, while preserving `--no-start-wn-opencode`.

This PR is **not entirely redundant**, however. Its Linux fixture executes the installer with a mocked `systemctl` and asserts the fresh-install, active-upgrade, and no-start call sequences. The test on `master` currently checks the dry-run wording and greps for the branch in the script, but does not execute those branches. The PR’s CI is green and there are no unresolved review threads.

Leaving this open for now because the production commits should not be merged as-is, but the execution-level test is worth either rebasing into a test-only PR or porting separately before closing this head.

PR #840, opened 2026-07-12
fix(hermes/marmot): accept npub welcomer ids and stop wiping the allowlist

### LINKED ISSUE #837: Hermes plugin: npub welcomer allowlist entries never work and silently wipe the allowlist
## Summary

Configuring the welcomer allowlist with an **npub** (which the installer explicitly accepts — `--allow-welcomer npub1...`) never works, and worse, it **silently wipes** the allowlist on every gateway connect. The agent then declines all group welcomes and "leaves immediately."

## Root cause

- `install-hermes-marmot.sh` accepts and validates npub (`--allow-welcomer VALUE  Allow invites from this npub or hex pubkey`), and `hermes_marmot_configure_gateway.py` writes the value **verbatim** into `platforms.marmot.extra.welcomer_allowlist`.
- At runtime, `normalize_welcomer_id()` only does `str(entry).strip().lower().removeprefix("0x")` — it does **not** decode bech32 npub → hex.
- `sync_allowlist()` builds `want` = entries that match the 64-char hex regex. An npub fails that regex, so `want` becomes **empty**, and the reconcile loop then **removes every existing entry** from wn-agent's allowlist.

Net effect: with an npub configured, the welcomer list is emptied on each connect → all group welcomes are declined. Any manually added hex id is also wiped on the next restart.

## Repro

```
install-hermes-marmot.sh --yes --allow-welcomer npub1...
# then, from the phone, invite the agent to a group -> it leaves instantly
```

`allowlist_list` shows `welcomer_account_ids_hex: []` even though config has the npub. Switching the config value to the raw 64-char hex makes it persist and work.

## Suggested fix

Decode npub → hex at the single normalization choke point so config/env/installer npub inputs all work:

```python
def normalize_welcomer_id(entry: str | int) -> str:
    s = str(entry).strip()
    if s.lower().startswith("npub1"):
        decoded = _npub_to_hex(s)      # small pure-Python bech32 decode
        if decoded:
            return decoded
    return s.strip().lower().removeprefix("0x")
```

Plus a safety guard in `sync_allowlist()`: do **not** remove existing entries when `want` is empty but `desired` was non-empty (i.e. everything failed to parse) — and log dropped/unparseable welcomer entries instead of silently discarding them. PR to follow.

jooray:
Fixes #837.

`install-hermes-marmot.sh` documents `--allow-welcomer npub1...` and `hermes_marmot_configure_gateway.py` writes the value verbatim into `platforms.marmot.extra.welcomer_allowlist`. But at runtime `normalize_welcomer_id()` only did `str(entry).strip().lower().removeprefix("0x")` — it never decoded bech32 npub → hex. `sync_allowlist()` then filtered the npub out of its hex-only `want` set, so `want` came out **empty**, and the reconcile loop **removed every existing entry** from wn-agent's allowlist on each connect.

Net effect with an npub configured: the welcomer list is emptied on every connect, so the agent declines all group welcomes and "leaves immediately"; any manually added hex id is also wiped on the next restart.

**Changes**
- `normalize_welcomer_id()` decodes `npub1...` → hex via a self-contained pure-Python bech32 decoder (no new dependencies), so npub inputs from config/env/installer all work. Non-npub input is returned lowercased for the caller's existing hex check.
- `sync_allowlist()` no longer removes existing entries when *every* desired entry failed to parse (a genuinely empty desired list still clears the allowlist as before), and logs dropped/unparseable entries instead of discarding them silently.

**Tests** — added to `WelcomerAllowlistTests`: npub → hex normalization, npub-in-`desired` sync adds the decoded hex, and the no-wipe-on-parse-failure guard. Full `test_adapter.py` (85 tests) and `test_configure_gateway.py` pass.

Verified on a live deployment: switching the configured welcomer to hex (equivalent to decoding the npub) makes the allowlist persist across restarts and the agent accept group invites.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/840">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/integrations/hermes/marmot/adapter.py
+++ b/integrations/hermes/marmot/adapter.py
@@ -415,8 +415,88 @@ def group_state_change_sentence(change: str, detail: Optional[str] = None) -> st
     return "The group state changed."
 
 
+_BECH32_CHARSET = "qpzry9x8gf2tvdw0s3jn54khce6mua7l"
+
+
+def _bech32_polymod(values: list[int]) -> int:
+    generators = (0x3B6A57B2, 0x26508E6D, 0x1EA119FA, 0x3D4233DD, 0x2A1462B3)
+    chk = 1
+    for value in values:
+        top = chk >> 25
+        chk = ((chk & 0x1FFFFFF) << 5) ^ value
+        for i in range(5):
+            chk ^= generators[i] if ((top >> i) & 1) else 0
+    return chk
+
+
+def _bech32_decode(bech: str) -> Tuple[Optional[str], Optional[list[int]]]:
+    if any(ord(ch) < 33 or ord(ch) > 126 for ch in bech):
+        return None, None
+    if bech.lower() != bech and bech.upper() != bech:
+        return None, None
+    bech = bech.lower()
+    pos = bech.rfind("1")
+    if pos < 1 or pos + 7 > len(bech):
+        return None, None
+    if any(ch not in _BECH32_CHARSET for ch in bech[pos + 1:]):
+        return None, None
+    hrp = bech[:pos]
+    data = [_BECH32_CHARSET.find(ch) for ch in bech[pos + 1:]]
+    expanded = [ord(c) >> 5 for c in hrp] + [0] + [ord(c) & 31 for c in hrp]
+    if _bech32_polymod(expanded + data) != 1:
+        return None, None
+    return hrp, data[:-6]
+
+
+def _bech32_convertbits(data: list[int], frombits: int, tobits: int) -> Optional[list[int]]:
+    acc = 0
+    bits = 0
+    ret: list[int] = []
+    maxv = (1 << tobits) - 1
+    for value in data:
+        if value < 0 or (value >> frombits):
+            return None
+        acc = (acc << frombits) | value
+        bits += frombits
+        while bits >= tobits:
+            bits -= tobits
+            ret.append((acc >> bits) & maxv)
+    # npub packs a 32-byte payload evenly, so no padding bits may remain.
+    if bits >= frombits or ((acc << (tobits - bits)) & maxv):
+        return None
+    return ret
+
+
+def npub_to_hex(value: str) -> Optional[str]:
+    """Decode a bech32 ``npub1...`` public key into 64-char lowercase hex.
+
+    Returns ``None`` for anything that is not a well-formed npub (wrong hrp,
+    bad checksum, wrong length) so callers can fall back to treating the input
+    as a raw hex id.
+    """
+    hrp, data = _bech32_decode(value)
+    if hrp != "npub" or data is None:
+        return None
+    decoded = _bech32_convertbits(data, 5, 8)
+    if decoded is None or len(decoded) != 32:
+        return None
+    return bytes(decoded).hex()
+
+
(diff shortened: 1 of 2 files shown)


erskingardner (changes requested):
Thanks for this — the root-cause diagnosis is precise and the fix is well-scoped. I independently cross-checked the bech32 decoder against BIP-173 and the PR's test vector, and the npub↔hex round-trip is correct. The `sync_allowlist` guard also correctly distinguishes "every desired entry failed to parse" (preserve existing allowlist, fail safe) from "desired is genuinely empty" (still clears, matching prior behavior), which is the right default for something that gates who can join encrypted groups.

One blocking issue before this can land:

### Blocking: new warning logs raw pubkey/account-id values

`integrations/hermes/marmot/adapter.py` (~lines 517–524):

```python
if dropped:
    logger.warning(
        "Marmot welcomer allowlist: ignoring %d unparseable entr%s "
        "(expected 64-char hex or npub1...): %s",
        len(dropped),
        "y" if len(dropped) == 1 else "ies",
        ", ".join(dropped),
    )
```

`dropped` holds the raw configured entries — i.e. attempted npub/hex account ids. Logging them violates the privacy-logging invariant stated in root `AGENTS.md`, `integrations/AGENTS.md`, and `integrations/hermes/marmot/AGENTS.md` ("no account ids, ... pubkeys, ..." in logs). Every other log line in this file follows the count-only convention (e.g. the adjacent `added=%d removed=%d` debug log).

Because `_sync_welcomer_allowlist()` runs on every `connect()`, a persistently misconfigured entry re-logs the same pubkey material on every reconnect — a recurring leak into logs that often end up in support bundles or pasted into issues.

**Fix:** log the count only, or a truncated/hashed form if operators need to distinguish entries — never the raw value.

### Non-blocking suggestions

- In the `have` set construction (~lines 522–531), `normalize_welcomer_id(entry)` is still called twice per entry (once in the expression, once in the filter), while the `want` loop above was refactored to call it once via a local. Cheap either way, but worth matching for consistency.
- Decoder edge cases aren't directly unit-tested: a syntactically valid bech32 string with (a) a non-`npub` HRP (e.g. `nsec1...`), (b) a bad checksum, or (c) a payload that isn't 32 bytes after `convertbits`. These are exactly the cases `npub_to_hex`'s guards exist for — a couple of targeted tests on `npub_to_hex` directly would pin that behavior down. (The existing `npub1invalid` case fails on charset before checksum validation is reached.)
- `_bech32_decode` omits BIP-173's 90-char upper bound on total string length. Not exploitable here (config-driven input, linear polymod), but free spec conformance / defense-in-depth.

erskingardner:
Superseded by #1043. I compared this head against the merged `master` implementation rather than relying only on the closed target issue:

- `master` decodes valid `npub1...` values to the same 32-byte lowercase hex form.
- It preserves the existing allowlist when a non-empty configured set contains no valid entries.
- It has regression tests for both successful npub reconciliation and the no-wipe-on-invalid case.
- The merged implementation is slightly stricter: it bounds Bech32 input length and logs only an invalid-entry count rather than echoing untrusted configured values.

The public behavior and regression contract from this PR are therefore fully represented on `master`; there is no unique change left to merge.

PR #852, opened 2026-07-13
fix(media): use ciphertext-compatible Blossom defaults

### LINKED ISSUE #851: Encrypted media uploads fail with HTTP 415 against default Blossom endpoint
## Summary

Encrypted media uploads from both Android and iOS fail against the MDK default Blossom endpoint:

```
Send failed
details=blob store request failed: upload failed for all Blossom servers: server 1: upload returned HTTP 415
```

Both clients pass no per-request server override, so MDK uses the endpoint embedded in the group encrypted-media policy. New groups currently default to `https://blossom.primal.net`.

## Reproduction

1. Create or use a group whose `marmot.group.encrypted-media.v1` component contains the default Primal endpoint.
2. Send an image or another media attachment from Android or iOS.
3. MDK encrypts the plaintext with ChaCha20-Poly1305.
4. MDK uploads the ciphertext to `PUT https://blossom.primal.net/upload` as `application/octet-stream`.
5. Primal returns HTTP 415.

An authenticated upload probe reproduced the server response:

```
upload rejected: unsupported media type application/octet-stream
```

Changing only the declared header to `image/jpeg` does not resolve the failure because the uploaded body is encrypted ciphertext and the server identifies it as opaque binary.

## Root cause

MDK combines an opaque encrypted-media format with a sole default Blossom server that currently rejects opaque `application/octet-stream` uploads.

The client-visible diagnostic also drops the response body and optional Blossom reason header, leaving only the status code.

## Scope

- Select/configure default encrypted-media Blossom endpoints that accept opaque ciphertext.
- Avoid a single incompatible endpoint as the default for newly created groups.
- Preserve ordered endpoint failover.
- Include a migration path for existing group policies through `replace_encrypted_media_blob_endpoints`.
- Surface a privacy-safe server rejection reason or response body for failed uploads.
- Add tests for a 415 response, diagnostic preservation, and endpoint failover.
- Document the fact that changing the application default affects new groups only; existing groups retain their embedded endpoint policy.
- Confirm the shared group-image upload path is compatible because it also uploads encrypted bytes.

## Client impact

Android and iOS should continue delegating endpoint selection to MDK. They should only need updated MDK bindings/artifacts and, where required, an admin migration for existing groups.

mubarakcoded:
## Summary

- replace the Primal-only encrypted-media default with an ordered list of Blossom endpoints verified to accept opaque `application/octet-stream` ciphertext
- preserve bounded, privacy-filtered Blossom rejection reasons so client errors explain failures such as HTTP 415
- document that endpoint policy is signed group state and that existing groups require an admin migration

## Root cause

MDK encrypts media before upload, so the Blossom request body is intentionally indistinguishable from random bytes and is sent as `application/octet-stream`. The previous built-in endpoint, `https://blossom.primal.net`, rejects that payload with HTTP 415. Android and iOS both delegate endpoint selection to MDK/the group's encrypted-media component, so the shared fix belongs here.

## Existing groups

The defaults apply when creating new groups. Upgrading MDK does not rewrite signed state for existing groups; an active group admin must call `replace_encrypted_media_blob_endpoints` to migrate a group whose embedded endpoints reject encrypted blobs.

Group images carry no endpoint in group state: upload and fetch resolve against each build's compiled primary endpoint, so clients built with different defaults look for group images in different places during a rollout window. Re-setting the group image on a current build republishes it to the current primary endpoint. Carrying the endpoint in the group-image component is tracked in #905.

## Verification

- `cargo test -p marmot-app`: 323 unit, 11 audit-log, and 68 relay-runtime tests passed
- focused media suite: 36 passed
- endpoint policy integration regression: passed
- `just fast-ci`: formatting, naming gate, workspace checks, feature checks, and Clippy all passed
- authenticated live probes using 1 KiB cryptographically random payloads: each selected endpoint accepted `application/octet-stream`; successful probes were deleted immediately

Closes #851

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/852">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Encrypted media and encrypted group-image uploads now fall back in a defined order across compatible upload endpoints, improving reliability and avoiding unsupported HTTP 415 responses.
  - Upload failures now surface privacy-safe, bounded server rejection reasons when present, while stripping sensitive or overly detailed content.

- **Documentation**
  - Updated CLI and app docs to clarify encrypted-media upload encryption, required `application/octet-stream` handling, endpoint selection behavior, and how to migrate existing groups when defaults change.
  - Added changelog notes describing the new endpoint policy and group-image endpoint resolution differences.

### DIFF
--- a/crates/marmot-app/src/media/blossom.rs
+++ b/crates/marmot-app/src/media/blossom.rs
@@ -118,7 +118,7 @@ fn privacy_safe_server_reason(reason: &str) -> Option<String> {
         return None;
     }
     let lowercase = reason.to_ascii_lowercase();
-    if ["http://", "https://", "nostr:", "npub1", "nsec1"]
+    if ["://", "nostr:", "npub1", "nsec1"]
         .iter()
         .any(|needle| lowercase.contains(needle))
     {
--- a/crates/marmot-app/src/media/blossom.rs
+++ b/crates/marmot-app/src/media/blossom.rs
@@ -125,9 +125,11 @@ fn privacy_safe_server_reason(reason: &str) -> Option<String> {
         return None;
     }
     if reason.split_ascii_whitespace().any(|token| {
-        let token = token.trim_matches(|character: char| !character.is_ascii_alphanumeric());
-        (token.len() >= 32 && token.bytes().all(|byte| byte.is_ascii_hexdigit()))
-            || (token.len() >= 48 && token.bytes().all(|byte| byte.is_ascii_alphanumeric()))
+        // Collapse punctuation so hyphenated hashes and UUIDs still register
+        // as one long identifier.
+        let token: String = token.chars().filter(char::is_ascii_alphanumeric).collect();
+        token.len() >= 48
+            || (token.len() >= 32 && token.bytes().all(|byte| byte.is_ascii_hexdigit()))
     }) {
         return None;
     }
--- a/crates/marmot-app/src/media/blossom.rs
+++ b/crates/marmot-app/src/media/blossom.rs
@@ -12,6 +12,7 @@ use crate::{AppError, unix_now_seconds};
 
 const BLOSSOM_UPLOAD_AUTH_TTL: Duration = Duration::from_secs(10 * 60);
 const BLOSSOM_UPLOAD_CONTENT_TYPE: &str = "application/octet-stream";
+const MAX_BLOSSOM_ERROR_BODY_BYTES: u64 = 1024;
 const MEDIA_HTTP_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
 const MEDIA_HTTP_READ_TIMEOUT: Duration = Duration::from_secs(15);
 const MEDIA_HTTP_TOTAL_TIMEOUT: Duration = Duration::from_secs(60);
@@ -45,10 +46,7 @@ pub(crate) async fn upload_blossom_blob(
         .await
         .map_err(reqwest_blob_error)?;
     if !response.status().is_success() {
-        return Err(AppError::BlobStore(format!(
-            "upload returned HTTP {}",
-            response.status().as_u16()
-        )));
+        return Err(blossom_upload_status_error(response).await);
     }
     let descriptor = response
         .json::<BlossomBlobDescriptor>()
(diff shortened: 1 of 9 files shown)


dannym-arx (dismissed):

crates/marmot-app/src/media/blossom.rs
```diff
@@ -76,6 +74,66 @@ pub(crate) async fn upload_blossom_blob(
...
+    }
+    if reason.split_ascii_whitespace().any(|token| {
+        let token = token.trim_matches(|character: char| !character.is_ascii_alphanumeric());
+        (token.len() >= 32 && token.bytes().all(|byte| byte.is_ascii_hexdigit()))
+            || (token.len() >= 48 && token.bytes().all(|byte| byte.is_ascii_alphanumeric()))
+    }) {
+        return None;
+    }
+    Some(reason)
+}
```
jgmontoya, line 137: From my agent:

> The long-token check only catches a token that is *purely* hex or *purely* alphanumeric. Any punctuation embedded
> inside the token (hyphen, colon, slash that isn't `://`, etc.) defeats it, because `trim_matches` only strips
> matching characters from the two ends, not the middle, so the mixed token fails both `is_ascii_hexdigit()` and
> `is_ascii_alphanumeric()` "all" checks and is let through untouched.
> 
> Verified with the actual function body extracted and run standalone:
> 
> ```
> "duplicate blob 1111...1111 exists"          => None   (correctly filtered: pure-hex 64-char token)
> "duplicate-blob-1111...1111-exists"          => Some("duplicate-blob-1111...1111-exists")   <- hash leaks
> "uuid 123e4567-e89b-12d3-a456-426614174000 already used" => Some(...)                        <- UUID leaks
> ```
> 
> A malicious or compromised Blossom endpoint (which the client is already talking to — either the group's signed
> default or a user-supplied `--server`) can trivially defeat this specific defense-in-depth check by hyphenating or
> otherwise punctuating an embedded hash/UUID/identifier, exactly contradicting the function's own doc comment ("without
> allowing an untrusted response to inject blob hashes, pubkeys, URLs"). Suggest stripping internal punctuation before
> the alnum/hex run-length check (e.g. `token.chars().filter(char::is_ascii_alphanumeric).collect::<String>()`), or
> scanning for any maximal alphanumeric *run* within the token rather than requiring the whole token to qualify.

jgmontoya (commented):
From my agent:

> ### Correctness: changing the default Blossom endpoint reopens a cross-version compatibility gap for group images
> 
> **crates/marmot-app/src/media/group_image.rs:18, 81, 120** (unchanged by this PR, but directly affected by it)
> 
> ```rust
> /// ... the image is self-contained and content-addressed by `image_hash_hex` — no URL
> /// or file name is stored.
> ...
> let server = server.unwrap_or(DEFAULT_BLOSSOM_SERVER_URL);   // upload_group_image, line 81
> ...
> let server = server.unwrap_or(DEFAULT_BLOSSOM_SERVER_URL);   // fetch_group_image, line 120
> ```
> 
> Unlike encrypted media attachments (whose endpoints live in the signed, replicated
> `marmot.group.encrypted-media.v1` component and are read from `policy.default_blob_endpoints` on both upload and
> download — see `crates/marmot-app/src/client/mod.rs:1382-1397` and `:1441`), an encrypted **group image** stores no
> locator at all. Both `upload_group_image` and `fetch_group_image` fall back to the crate-level
> `DEFAULT_BLOSSOM_SERVER_URL` when no explicit server is passed, and `client/mod.rs` (`update_group_image`,
> `fetch_group_image_bytes` — see call sites at `client/mod.rs:1469` and `:1515`) never passes an explicit server, so
> in practice it's always the uploading/fetching client's own build-time default.
> 
> This PR changes that default from `https://blossom.primal.net` to `https://blossom.divine.video`. During any rollout
> window where some peers in a group are running an MDK build compiled with the old default and others with the new
> one (completely normal for staggered app-store rollouts), a group avatar uploaded by an "upgraded" peer will be
> unreachable to a "non-upgraded" peer (and vice versa), because the fetching side reconstructs the URL from its own
> compiled default rather than from any value carried in the signed component. There is also no test coverage
> anywhere in `marmot-app` for `upload_group_image`/`fetch_group_image` (no unit tests in `media/group_image.rs`, no
> integration coverage in `tests/relay_runtime.rs`), so this path isn't exercised at all by `just fast-ci`/`cargo test`.
> 
> This is a pre-existing architectural gap (group images were always implicitly relying on a fleet-wide constant), but
> this PR is the first change that actually flips the constant's value, which is exactly the condition that turns the
> gap into a real, user-visible regression risk. Worth at least a callout in the PR's "Existing groups" migration note
> (which currently only discusses the encrypted-media policy component, not group images), and ideally worth fixing
> properly — e.g. by carrying the upload endpoint in the (already MLS-protected) group-image component fields, or by
> giving group images the same ordered-fallback list treatment used for message attachments.

crates/marmot-app/src/media/blossom.rs
mubarakcoded: Fixed in fbca7eb

mubarakcoded:
> From my agent:
> 
> > ### Correctness: changing the default Blossom endpoint reopens a cross-version compatibility gap for group images

Agreed on the gap: group images carry no endpoint in group state, so both sides resolve against their own compiled default, and this PR is the first change that actually flips that constant. For this PR I've documented it rather than fixed it — the README, CLI changelog, and the "Existing groups" section above now call out that clients built with different defaults resolve group images against different endpoints, with re-setting the image on a current build as the interim heal. The proper fix (carry the endpoint in the MLS-protected group-image component, plus test coverage for the currently untested upload/fetch path) is filed as #905 — it touches signed component fields, so I'd rather not grow this PR's scope while it's pending re-approval. One practical note on the rollout window: builds with the old default were already unable to upload group images at all (the same 415 this PR fixes), so the exposure is effectively one-directional — old builds failing to fetch newly uploaded images — not a regression from a working state.

jgmontoya (approved):

PR #874, opened 2026-07-15
Honor group-admin moderation deletes in timeline ingest

### LINKED ISSUE #873: Timeline ingest ignores group-admin deletes of other members' messages (self-retraction only)
## Summary

`marmot-app`'s `delete_message` publishes a delete intent for any target, but the timeline ingest applies it only when the delete's author matches the target message's author (`crates/storage-sqlite/src/timeline.rs`, the `row.sender != delete.sender` guard). Group admins therefore cannot moderate: their delete publishes and tombstones the message locally (optimistically in clients), while every other member silently ignores it, and the deleter's own store resurrects the message on reprojection.

## Expected

A group admin/owner deleting another member's message should tombstone it for the whole group, matching the app-layer product rule ("Delete for everyone" available to group admins). Suggested acceptance rule at ingest: honor a delete when `delete.sender == row.sender` OR `delete.sender` was an admin of the group at the time the delete was committed (MLS group context / roster check), never for direct (two-member, unnamed) conversations.

## Context

- White Noise Android's unified deletion redesign ships a capability model with the admin branch gated behind `GROUP_MODERATION_DELETE_SUPPORTED = false` until this lands; flipping that one constant enables the already-tested UI/policy.
- Client parity issues referencing the same product rules: marmot-protocol/whitenoise-ios#691, marmot-protocol/whitenoise-mac#501.

## Notes

- Reaction deletes already have a forged-delete guard keyed on sender equality; message moderation needs an explicit admin-aware rule rather than dropping that guard.
- Restart/reprojection must preserve the moderated tombstone.

mubarakcoded:
Timeline ingest previously honored a delete only when its author matched the target message's author, so a group admin deleting another member's message tombstoned it locally while every other member silently ignored it — and the deleter's own store resurrected the message on reprojection.

## Design

The moderation verdict is evaluated once, at ingest, against the signed MLS group state, and persisted with the event:

- `app_events.moderation_grant` (migration 0027) is stamped true only for a delete whose authenticated sender is in the group's admin set (`admin_pubkeys`, mirrored from signed group state) and whose group is not a direct (two-member, unnamed) conversation. Lookup failures fail closed to self-retraction semantics.
- The timeline acceptance rule becomes: honor a delete when `delete.sender == row.sender` OR the delete carries the persisted grant. Reaction retraction keeps its sender-equality forged-delete guard — moderation authority applies to messages, not reaction removal.
- The verdict is frozen at the event's first record (the upsert keeps the stored value), so reprojection and relay echoes arriving after an admin-set change can never flip an honored tombstone.
- The grant is stamped on both chokepoints: inbound sync ingest and the local-send projection, so the moderator's own store keeps the tombstone across reprojection too.

## Verification

- `just fast-ci` green (fmt, naming gate, check, clippy, workspace-wide including OTLP builds).
- `cargo test -p storage-sqlite`: 213 passed, including new coverage for admin tombstone via incremental projection, out-of-order delete-before-target delivery, full timeline rebuild preserving the moderated tombstone, and the reaction guard staying sender-equality.
- `cargo test -p marmot-app`: 331 passed, including grant-helper cases (direct conversations never grant, whitespace-only names still count as unnamed, named two-member groups do grant, empty admin set grants nothing).

Closes #873

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/874">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added persisted moderation authorization for delete events, so authorized admins can tombstone other members’ messages consistently across local projections and timeline rebuilds.
* **Bug Fixes**
  * Prevented non-authorized deletes from being treated as moderation-granted, including for cross-sender delete cases.
  * Improved moderation-grant “freeze” and “refresh” behavior so grants stay stable when re-received but can be recomputed correctly.
* **Tests**
  * Expanded coverage for admin authorization rules, direct-conversation exclusions, out-of-order handling, full rebuilds, and moderation-grant lifecycle.

### DIFF
--- a/crates/marmot-app/src/client/projection.rs
+++ b/crates/marmot-app/src/client/projection.rs
@@ -4,7 +4,9 @@ use cgka_traits::app_components::{
     GROUP_BLOSSOM_IMAGE_COMPONENT_ID, GROUP_ENCRYPTED_MEDIA_COMPONENT_ID,
     GROUP_MESSAGE_RETENTION_COMPONENT_ID, NOSTR_ROUTING_COMPONENT_ID,
 };
-use cgka_traits::app_event::{MARMOT_APP_EVENT_KIND_CHAT, MarmotAppEvent as MarmotInnerEvent};
+use cgka_traits::app_event::{
+    MARMOT_APP_EVENT_KIND_CHAT, MARMOT_APP_EVENT_KIND_DELETE, MarmotAppEvent as MarmotInnerEvent,
+};
 
 use crate::groups::{EventGroupProjection, GroupConfirmationProjection, add_group};
 use crate::{
@@ -23,18 +25,24 @@ impl AppClient {
     /// post-publish success projection advances it.
     pub(crate) fn record_local_app_event_projection(
         &self,
-        group_id_hex: &str,
+        group_id: &GroupId,
         sender: &str,
         event: &MarmotInnerEvent,
         source_message_id_hex: Option<String>,
         source_epoch: Option<u64>,
         advance_read_marker: bool,
     ) -> Result<crate::AppProjectionUpdate, AppError> {
+        let group_id_hex = hex::encode(group_id.as_slice());
+        // Stamped on the sender's own store too, so a moderation delete of
+        // another member's message survives local reprojection instead of
+        // resurrecting the target.
+        let moderation_grant = event.kind == MARMOT_APP_EVENT_KIND_DELETE
+            && self.delete_moderation_grant(group_id, sender);
         let message_projection = AppMessageProjection {
             message_id_hex: event.id.clone(),
             source_message_id_hex,
             direction: "sent".to_owned(),
-            group_id_hex: group_id_hex.to_owned(),
+            group_id_hex: group_id_hex.clone(),
             sender: sender.to_owned(),
             plaintext: event.content.clone(),
             kind: event.kind,
@@ -44,6 +52,7 @@ impl AppClient {
             // Only synthesized kind-1210 system rows carry an origin commit;
             // ordinary sent app events do not.
             origin_commit_id: None,
+            moderation_grant,
         };
         let update = self
             .app
@@ -51,7 +60,7 @@ impl AppClient {
         if advance_read_marker && event.kind == MARMOT_APP_EVENT_KIND_CHAT {
             let read_marker =
                 self.app
-                    .mark_timeline_message_read(&self.state.label, group_id_hex, &event.id);
+                    .mark_timeline_message_read(&self.state.label, &group_id_hex, &event.id);
             if let Err(err) = read_marker {
                 let error_code = read_marker_error_code(&err);
                 tracing::warn!(
@@ -63,6 +72,18 @@ impl AppClient {
             }
         }
         Ok(update)
+    }
+
+    /// Fail-closed: a group or admin-policy lookup failure yields no grant, so
+    /// the delete degrades to self-retraction semantics.
+    pub(crate) fn delete_moderation_grant(&self, group_id: &GroupId, sender_hex: &str) -> bool {
+        let Ok(group) = self.runtime.group_record(group_id) else {
+            return false;
+        };
+        let Ok(admins) = self.runtime.admin_pubkeys(group_id) else {
+            return false;
+        };
+        crate::groups::delete_moderation_grant(&group, &admins, sender_hex)
     }
 
     pub(crate) fn state_group_record(&self, group_id: &GroupId) -> Option<AppGroupRecord> {
@@ -342,6 +363,7 @@ fn build_group_system_projection(
         tags: material.tags,
         source_epoch: Some(epoch),
         recorded_at: Some(recorded_at),
+        moderation_grant: false,
         // Non-unique link to the origin commit so a losing-branch rollback can
         // invalidate every row this commit synthesized (1:N).
         origin_commit_id,
--- a/crates/marmot-app/src/client/projection.rs
+++ b/crates/marmot-app/src/client/projection.rs
@@ -54,9 +54,20 @@ impl AppClient {
             origin_commit_id: None,
             moderation_grant,
         };
-        let update = self
-            .app
-            .record_account_app_event(&self.state.label, &message_projection)?;
+        // The reconciling post-publish projection (advance_read_marker) runs
+        // after group sync, so its recomputed moderation grant supersedes the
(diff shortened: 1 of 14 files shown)


jgmontoya:
@coderabbitai full review

crates/storage-sqlite/src/migrations/0027_app_event_moderation_grant.rs
```diff
@@ -0,0 +1,12 @@
+use crate::SqliteResultExt;
+use cgka_traits::storage::StorageResult;
+use rusqlite::Transaction;
+
+pub(crate) fn apply(tx: &Transaction<'_>) -> StorageResult<()> {
+    tx.execute_batch(
+        r#"
+ALTER TABLE app_events ADD COLUMN moderation_grant INTEGER NOT NULL DEFAULT 0;
```
jgmontoya, line 15: Should we properly backfill? I'm not sure it's worth it but would rather it be a deliberate decision

crates/marmot-app/src/groups.rs
jgmontoya, line 1: From my agent:

> ### Correctness/Security: "direct conversation" gate is derived from a mutable, admin-controlled field
> 
> **crates/marmot-app/src/groups.rs:176-183**
> 
> > ```rust
> > pub(crate) fn delete_moderation_grant(
> >     group: &Group,
> >     admins: &[[u8; 32]],
> >     sender_hex: &str,
> > ) -> bool {
> >     let direct = group.members.len() == 2 && group.name.trim().is_empty();
> >     !direct && admins.iter().any(|admin| hex::encode(admin) == sender_hex)
> > }
> > ```
> 
> The doc comment states the intent plainly: "direct chats stay self-retraction only, so neither peer can moderate the other." But "direct" here is inferred from `members.len() == 2 && name.trim().is_empty()` — and both halves of that are within the reach of whoever created the conversation:
> 
> - `crates/cgka-engine/src/group_lifecycle.rs:169-171` — the creator is *always* an implicit admin (`let mut admin_set: Vec<[u8; 32]> = vec![creator_pubkey];`), regardless of `initial_admins`.
> - There is no separate "start a DM" API. `create_group` (`crates/marmot-app/src/client/mod.rs:219`) is the only creation path for any conversation, 1:1 or group, and takes an arbitrary `name: &str` for any member count. `validate_group_profile` (`crates/marmot-app/src/groups.rs:1238`) only checks length, and the CLI (`crates/cli/src/commands/groups.rs:33-41`) passes the caller-supplied name straight through with no special-casing for a 2-member group.
> - Renaming a group's profile (`GROUP_PROFILE_COMPONENT_ID` via `do_send_update_group_data`) requires admin (`crate::app_components::require_admin` in `crates/cgka-engine/src/update_group_data.rs:100`) — but the creator already *is* admin, so they can rename at will.
> 
> Put together: Alice creates a "DM" with Bob. Whether by naming it at creation or renaming it moments later (she's already an admin), the conversation stops being "direct" per this heuristic. From that point, Alice's deletes of Bob's messages get `moderation_grant = true` and are honored on *every* client, including Bob's own — the exact "everyone else honors it" behavior this PR intentionally adds for admin moderation. Bob has no way to know that a name change flipped a private 1:1 into a moderatable one; there's no protocol-level "conversation kind" that's immutable from creation, and this crate has no way to enforce that downstream consumers (this repo doesn't contain the actual client apps) ever leave 1:1 conversations unnamed in the first place.
> 
> This isn't a hypothetical edge case reachable only through misuse — it's reachable through the only group-creation/rename API that exists in this codebase, by the party the guard is supposed to restrict, using ordinary operations (name a group, or rename it).
> 
> Suggest tracking a real, immutable "conversation kind" set once at creation time (e.g. a bit alongside `required_capabilities` or in the group-creation request) instead of inferring "direct" from mutable state. At minimum this should be called out as a known limitation rather than presented as "direct chats stay self-retraction only" without qualification.

crates/marmot-app/src/client/projection.rs
jgmontoya, line 1: From my agent:

> ### Design: current-admin-set evaluation + freeze-at-first-record can still diverge across devices
> 
> **crates/marmot-app/src/client/projection.rs:88-96**, **crates/storage-sqlite/src/timeline.rs:456-463**
> 
> `delete_moderation_grant` evaluates against whatever `admin_pubkeys` returns *right now* on this device (current signed group state), not the admin set as of the delete's epoch, and the verdict then freezes at first record. That means two devices that first observe the *same* delete event at different points in their own local sync/hydration timeline (e.g. one already reflects an admin-adding commit locally, the other hasn't caught up yet, or one ingests while the group is quarantined via `ensure_group_live`) can permanently disagree on whether it's honored — even after both devices are fully caught up. This is a milder recurrence of the exact cross-device inconsistency #873 was about. It's explicitly a documented trade-off in the PR description ("fail closed... frozen at first record"), so this may be intentional, but worth confirming it's an accepted risk given how much of this codebase's recent history (fork-recovery, quarantine, incident-replay) is about exactly this kind of transient per-device divergence.

crates/storage-sqlite/src/migrations/0027_app_event_moderation_grant.rs
mubarakcoded: No backfill, on purpose. Old rows default to false, which is what they were
before: those deletes were all self-retraction, so false is the right verdict
for them. Backfilling true would honor deletes no client ever applied.
Documented in b5761cb.

crates/marmot-app/src/groups.rs
mubarakcoded: Fair point. A real fix means an immutable conversation-kind set at creation,
which is a protocol change and has to match how the clients detect DMs, so I'd
rather not do it piecemeal here. For now, I dropped the overclaim and documented
the limitation (b5761cb), and I'll open a follow-up. If you want a stopgap, I can
treat any two-member chat as direct regardless of name, but that diverges from
the clients until the coordinated fix.

crates/marmot-app/src/client/projection.rs
mubarakcoded: Intentional. The grant is read from the current admin set and frozen at first
record, so devices that first see the delete at different sync points can
disagree. Cleaner fix is anchoring to the delete's epoch, noted as future work
in b5761cb. Happy to treat it as accepted for this PR unless you'd rather block
on it.

jgmontoya (dismissed):

dannym-arx (approved):

PR #875, opened 2026-07-15
feat: add Sonar-compatible sticker support

vincenzopalazzo:
## Summary
- integrate the pinned sonar-stickers SDK for NIP-30031 packs, kind-10031 installed lists, and exact kind-9 sticker references
- persist validated pack/install/outbox projections in the per-account encrypted SQLite database
- add bounded public relay and Blossom flows, Signal pack import, typed app/runtime projections, and UniFFI bindings

## Security and privacy
- exact hash, MIME, dimension, animation-frame, redirect, DNS/IP, and byte-size validation
- trusted Sonar/Signal URL origins only; relay hints are ignored
- Signal pack keys are zeroized and never persisted
- external-signer Signal import is explicitly gated to avoid mass signer prompts

## Verification
- cargo test -p marmot-app -p marmot-uniffi -p storage-sqlite
- cargo clippy -p marmot-app -p marmot-uniffi -p storage-sqlite --all-targets -- -D warnings

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/875">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added support for Sonar sticker packs, including browsing, synchronization, installation, removal, Signal imports, and asset downloads.
  - Added sticker sending across messaging, notifications, timelines, chat previews, and native integrations.
  - Added sticker-pack persistence and migration support.
- **Improvements**
  - Added clear sticker-specific errors and validation.
  - Added bounded media downloads and public event retrieval for sticker assets.


crates/storage-sqlite/src/chat_list.rs
```diff
@@ -881,8 +899,11 @@ fn chat_list_message_from_row(row: &rusqlite::Row<'_>) -> rusqlite::Result<ChatL
         sender_display_name: None,
         plaintext: row.get(2)?,
         kind: row.get::<_, i64>(3)?.try_into().unwrap_or_default(),
-        timeline_at: row.get::<_, i64>(4)?.try_into().unwrap_or_default(),
-        deleted: row.get::<_, i64>(5)? != 0,
+        tags: serde_json::from_str(&row.get::<_, String>(4)?).map_err(|err| {
+            rusqlite::Error::FromSqlConversionFailure(4, rusqlite::types::Type::Text, Box::new(err))
+        })?,
```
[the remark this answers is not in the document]
vincenzopalazzo: Verified against the schema and write paths: `message_timeline.tags_json` is created as `TEXT NOT NULL` in migration 0004, and the table did not exist before that migration, so there are no pre-schema rows with a nullable value. This mapper selects `message_timeline.tags_json`; migration 0026 only adds the separate nullable `chat_list_rows.last_message_tags_json` projection column. Every timeline insert serializes tags to non-null JSON. Reading this as `Option<String>` would mask an unreachable/corrupt state rather than provide legacy compatibility, so no code change is needed.

crates/storage-sqlite/src/timeline.rs
```diff
@@ -2410,21 +2415,24 @@ fn reply_preview_from_row(row: &rusqlite::Row<'_>) -> rusqlite::Result<TimelineR
         sender: row.get(1)?,
         plaintext: row.get(2)?,
         kind: row.get::<_, i64>(3)?.try_into().unwrap_or_default(),
-        media: optional_value_from_json(row.get::<_, Option<String>>(4)?).map_err(|err| {
+        tags: serde_json::from_str(&row.get::<_, String>(4)?).map_err(|err| {
             rusqlite::Error::FromSqlConversionFailure(4, rusqlite::types::Type::Text, Box::new(err))
         })?,
```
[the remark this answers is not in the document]
vincenzopalazzo: Verified against migration 0004: `message_timeline.tags_json` has been `TEXT NOT NULL` since the table was introduced, and all timeline write/projection paths persist serialized non-null tag JSON. No older version of this table can contain a legacy NULL row. This reply-preview mapper therefore correctly treats NULL/type mismatch or malformed JSON as storage corruption instead of silently converting it to an empty tag set. No code change is needed.

vincenzopalazzo:
Addressed the remaining non-inline CodeRabbit feedback in 9b8f3cb: sticker-pack searches now escape SQLite LIKE metacharacters (`%`, `_`, and `!`) and include a regression test proving literal semantics. The later Grok closure review also found and drove the cached/offline historical-reference fix in 61feace; its final verdict was **No findings**. Full `storage-sqlite` and `marmot-app` suites, clippy with warnings denied, and formatting are green.

erskingardner:
@vincenzopalazzo I think we need to do more protocol level work before we can get to a PR to add this to MDK. ideally we would have a NIP PR up on the NIPs repo. Do you want to do that work starting from the doc you have (https://github.com/hedwig-corp/bitchat-to-sonar/blob/162ac26c88b351dfa3f35d8f098c00a18ae46114/docs/SONAR-STICKERS.md#identifiers)? Or do you want to start something new?

vincenzopalazzo:
Hey @erskingardner talking with fiatjaf in DM he direct me to open a couple of PR on https://github.com/nostr-protocol/nips/pull/2410 and https://github.com/nostr-protocol/registry-of-kinds/pull/4 based on the docs that I have on sonar website

Are these enough? Or were you referring to something deeper?

Datawav:

## Final audit

### High — GIF frame geometry bypasses sticker dimension limits

`inspect_image` limits only the GIF logical-screen dimensions. In `inspect_gif`, each image descriptor increments the frame count but its left, top, width, and height fields are never parsed or validated (`crates/marmot-app/src/stickers.rs`, added lines 1330–1398). A GIF with a small logical screen can therefore declare a frame as large as 65,535×65,535 and pass validation, potentially causing excessive decoder allocation or denial of service.

Validate every frame’s nonzero dimensions and pixel count, and require its bounds to fit within the logical screen. Add adversarial tests for oversized and out-of-canvas descriptors.

### Medium — Pack links disclose user interest to attacker-selected relays

The parser accepts relay hints embedded in `naddr` values and Sonar URL query parameters (added lines 165–245). `fetch_sticker_pack` and `install_sticker_pack` pass those hints to `fetch_pack_into_storage`, which contacts them before configured account relays (added lines 434–468 and 658–697). Anyone sharing a crafted pack link can therefore cause a connection to their relay, exposing the client IP and exact pack being requested. This also conflicts with the PR’s stated privacy claim that relay hints are ignored.

Ignore untrusted link-provided hints and fetch through configured relays. If interoperability requires hints, an alternative is explicit user approval or a trusted-relay allowlist; relay sanitization alone does not address this privacy leak.

Typed sticker references, hash-pinned historical assets, encrypted per-account projections, and an outbox/rebase model address a legitimate need and fit the existing architecture. The principal alternative—resolving mutable pack state without preserving hash-exact asset history—would weaken message integrity, so the chosen persistence design is justified. Local Signal import is also reasonable, provided the image-validation boundary is completed as described above.

The advisory WebP frame-count concern is not actionable: `inspect_webp` returns the counted `ANMF` frames, and the shared `inspect_image` check rejects counts above 200. The pinned Git dependency is a maintenance-policy concern, but the supplied patch does not establish a concrete defect or security failure.

_Advisory audit of head [`e0e12be709a7`](https://github.com/marmot-protocol/mdk/commit/e0e12be709a7e7978440cd2000bd35535538b6f8); no approval, merge, push, checkout, or PR-code execution was performed._

erskingardner:
@vincenzopalazzo want to rebase this and add some details about the NIP in the comments? would love to get it into the apps.

Datawav:
## Nightly deep audit

1. **High — Cross-instance publication can silently discard a concurrent install/uninstall.**  
   `publish_pending_installed_list` reads the desired projection and signs a snapshot, performs an asynchronous relay publication, and then `commit_installed_sticker_publication` replaces the base and executes `DELETE FROM app_sticker_install_operations` without identifying which operations the event actually contains (`crates/marmot-app/src/stickers.rs`; `crates/storage-sqlite/src/stickers.rs`). The new mutex only serializes callers sharing one `MarmotApp`; it does not protect another app instance or process using the same account database. If instance A signs operation X, instance B enqueues Y while A awaits relay publication, and A then commits, Y is deleted despite never appearing in A’s event. The advertised durable outbox/rebase model therefore loses user intent under a valid fork/convergence race.  
   **Need/design:** Give operations immutable sequence IDs and record the publication’s covered high-water mark, then delete only covered operations in the commit transaction. Alternatively, create the snapshot/outbox row and associate its exact operation IDs atomically, clearing only those IDs after publication. A database-backed lease can serialize publishers, but it still needs crash expiry.  
   **Test gap:** Use two independently opened storage/app instances and pause publication between snapshot creation and commit; enqueue a second operation and verify it survives.

2. **High — Animated-image validation does not enforce the claimed decoder resource bounds.**  
   `inspect_webp` increments `frames` for every `ANMF` chunk but never rejects a count above `MAX_STICKER_ANIMATION_FRAMES`; the later shared check receives only the final saturating count, so ordinary counts above 200 reach it and are rejected—but only after scanning the whole file. More importantly, neither `inspect_webp` nor `inspect_gif` validates per-frame geometry. GIF image descriptors’ left/top/width/height fields are skipped, while WebP `ANMF` frame rectangle fields are ignored. Thus a file with a small validated canvas can declare a frame up to format limits, potentially inducing a much larger allocation in host decoders. APNG already validates each frame against its canvas, demonstrating the intended invariant.  
   **Need/design:** Parse every GIF image descriptor and WebP `ANMF` header; require nonzero bounded frame dimensions, checked coordinate arithmetic, containment within the logical canvas, and reject immediately after frame 200. Also bound WebP parsing to the declared RIFF extent rather than `bytes.len()`. Using a hardened metadata decoder is an alternative if it exposes frame rectangles without decoding pixels.  
   **Test gap:** Add oversized/out-of-canvas GIF and WebP frames, more than 200 `ANMF` chunks, integer-boundary coordinates, and data appended beyond the declared RIFF container.

3. **Medium — Untrusted links trigger attacker-selected relay connections and can pin a stale pack version.**  
   `parse_sticker_pack_input_with_relays` accepts relay hints from both `naddr` data and Sonar query parameters. `fetch_pack_into_storage` contacts those endpoints first and returns immediately when any valid matching event is found, without querying configured account relays. A crafted pack link therefore reveals the client IP and exact pack interest to arbitrary relay operators. It can also make a legitimately signed but stale event win locally by withholding the author’s newer version. Endpoint sanitization addresses SSRF-style destinations, not this privacy or convergence issue. This directly contradicts the PR description’s claim that relay hints are ignored.  
   **Need/design:** Ignore input-provided hints, matching the stated policy. If interoperability requires them, combine them with configured relays in one fetch and select the normal NIP-01 replacement winner across all responses; contacting new relays should require an explicit trust policy or user approval.  
   **Test gap:** Assert that parsing/opening attacker-controlled links never contacts their hints, or that configured and hinted responses are jointly converged before persistence.

4. **Medium — Public discovery and on-demand fetches create an unbounded durable database projection.**  
   `MAX_DISCOVERY_PACKS` limits a single response, not retained state. Every sync can persist up to 100 previously unseen coordinates through `replace_sticker_pack_if_newer`, and `fetch_sticker_pack`/received-reference resolution can add arbitrary additional packs. Neither migration nor storage code applies a global pack/asset quota or evicts old, uninstalled data. Because each retained pack has associated current and historical asset rows, relay churn or repeated crafted references can grow the encrypted account database indefinitely. The 100-pack installed-list limit does not bound this path.  
   **Need/design:** Preserve installed packs and assets referenced by retained messages, while applying a transactional count/byte quota and LRU/age eviction to disposable discovery entries. A simpler alternative is keeping discovery results ephemeral and persisting only installed or message-referenced packs.  
   **Test gap:** Run repeated disjoint discovery windows and on-demand fetches, then assert a fixed retained-state bound without removing installed or historically referenced assets.

Datawav:
Health remediation applied for stale head `e0e12be709a7e7978440cd2000bd35535538b6f8`.

- Merged current `marmot-protocol/mdk` `master` (`623892b8508774c42f334bf0ea83206a7af1c039`) into `codex/sonar-stickers`.
- Resolved 18 conflicts while preserving both the Sonar sticker feature and current upstream behavior; sticker migrations were renumbered to 0046/0047 to avoid collisions.
- New head: `356f06fa636ec7885bf9dc2f2a699391221efc45`.
- Verified: `cargo fmt --all -- --check`; `cargo check -p marmot-app -p marmot-uniffi -p storage-sqlite`; 18 targeted sticker tests passed (7 storage-sqlite, 11 marmot-app; marmot-uniffi compiled with 0 matching tests).
- GitHub now reports the PR mergeable; remaining blocked state is policy/check related, not a merge conflict.

crates/storage-sqlite/src/migrations/0046_sonar_stickers.rs
```diff
@@ -0,0 +1,72 @@
...
+    height INTEGER,
+    alt TEXT,
+    emoji TEXT,
+    position INTEGER NOT NULL,
+    PRIMARY KEY (pack_coordinate, shortcode),
+    FOREIGN KEY (pack_coordinate) REFERENCES app_sticker_packs(coordinate) ON DELETE CASCADE
+);
+
+CREATE UNIQUE INDEX idx_app_stickers_pack_hash
+    ON app_stickers (pack_coordinate, sha256);
```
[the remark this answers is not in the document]
vincenzopalazzo: Good catch -- dropped the UNIQUE constraint on `idx_app_stickers_pack_hash` so packs can reuse the same image across shortcodes. **Fixed in 301f445.** Added test `pack_can_reuse_one_image_across_multiple_shortcodes`.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1101,12 +1102,12 @@ fn rebuild_chat_list_row_for_group_tx(
...
             last_read_message_id_hex, last_read_timeline_at,
             conversation_created_at, activity_sort_at, retained_activity_sort_at,
-            updated_at, self_membership
+            updated_at, self_membership, last_message_tags_json
          )
          VALUES (
             ?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10,
             ?11, ?12, ?13, ?14, ?15, ?16, ?17, ?18, ?19, ?20,
-            ?21, ?22, ?23, ?24, ?25, ?26, ?27, 0, ?28, ?29
+            ?21, ?22, ?23, ?24, ?25, ?26, ?27, 0, ?28, ?29, ?30
```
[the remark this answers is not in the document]
vincenzopalazzo: Good catch -- added a structural completeness check in `chat_list_projection_complete_tx` that forces a rebuild for any row with `last_message_id_hex IS NOT NULL` but NULL `last_message_tags_json`. This catches existing rows left stale by migration 0046's nullable column. **Fixed in 301f445.** Added test `ensure_chat_list_rows_rebuilds_when_last_message_tags_json_is_null`.

Cargo.toml
```diff
@@ -119,6 +119,7 @@ cgka-traits = { path = "crates/traits" }
 cgka-engine = { path = "crates/cgka-engine" }
 cgka-session = { path = "crates/cgka-session" }
 storage-sqlite = { path = "crates/storage-sqlite" }
+sonar-stickers = { git = "https://github.com/hedwig-corp/bitchat-to-sonar", rev = "162ac26c88b351dfa3f35d8f098c00a18ae46114", features = ["nostr", "signal-import"] }
```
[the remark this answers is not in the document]
vincenzopalazzo: Not a bug -- the git dependency resolves correctly. Cargo finds `sonar-stickers v0.1.0` at the pinned rev even without a root `Cargo.toml` — the `core/Cargo.toml` workspace manifest exposes it as a member, and `cargo check -p marmot-app` succeeds. No code change needed.

crates/storage-sqlite/src/stickers.rs
```diff
@@ -0,0 +1,969 @@
...
+impl SqliteAccountStorage {
+    /// Atomically replace one addressable pack only when the incoming NIP-01
+    /// replacement wins. Newer `created_at` wins; equal timestamps choose the
+    /// lowest event id. Pack metadata and items change in one transaction so a
+    /// reader never observes a half-replaced pack.
+    pub fn replace_sticker_pack_if_newer(&self, pack: &StoredStickerPack) -> StorageResult<bool> {
+        self.connection.with_transaction(|| {
+            let conn = self.lock()?;
+            let existing = sticker_pack_version_tx(&conn, &pack.coordinate)?;
+            if !replacement_wins(existing.as_ref(), &pack.version) {
```
Datawav, line 70: Returning before the asset inserts means a valid older replacement event contributes none of its immutable sticker assets. If a newer pack version arrived first, messages referencing an asset that exists only in the older version can remain `StickerNotFound`; the fetch path also skips relay resolution whenever any current pack exists. Persist assets from every valid pack event independently of whether it wins replacement ordering, and allow bounded historical resolution when the exact `(coordinate, shortcode, hash)` is absent.

crates/marmot-app/src/stickers.rs
```diff
@@ -0,0 +1,1814 @@
...
+        let signer = context.signer.as_nostr_signer();
+        let mut stickers_by_id = HashMap::new();
+        let mut stickers = Vec::with_capacity(imported.stickers.len());
+        for imported_sticker in &imported.stickers {
+            validate_sticker_asset_size(&imported_sticker.bytes)?;
+            let inspected = inspect_image(&imported_sticker.bytes)?;
+            if sha256_hex(&imported_sticker.bytes) != imported_sticker.sha256 {
+                return Err(invalid_sticker("Signal sticker hash mismatch"));
+            }
+            let url = upload_blossom_blob(
```
Datawav, line 549: This starts a durable Blossom upload before the remaining stickers and final pack metadata have been validated. If a later image, hash, sticker constructor, address, or pack constructor fails, previously uploaded blobs are orphaned even though the import returns an error. Validate every asset and construct the complete pack/upload plan without side effects before the first upload; still handle partial failures during the upload/publication phase explicitly.

crates/marmot-app/src/stickers.rs
```diff
@@ -0,0 +1,1814 @@
...
+    } else if bytes.starts_with(b"RIFF") && bytes.get(8..12) == Some(b"WEBP") {
+        inspect_webp(bytes)?
+    } else {
+        return Err(invalid_sticker("unsupported sticker image format"));
+    };
+    if inspected.width == 0
+        || inspected.height == 0
+        || inspected.width > MAX_STICKER_DIMENSION
+        || inspected.height > MAX_STICKER_DIMENSION
+        || u64::from(inspected.width) * u64::from(inspected.height) > MAX_STICKER_PIXELS
```
Datawav, line 1231: These checks constrain only the reported canvas, not individual animation frames. The GIF parser skips descriptor offsets and dimensions, and the WebP parser only counts `ANMF` chunks, so zero-sized, overflowing, over-limit, or out-of-canvas frames can pass validation and trigger excessive allocation or decoder failures downstream. Parse and validate every GIF descriptor and WebP `ANMF` rectangle against the canvas and configured limits.

Datawav (commented):
## Automated review

Outcome: three actionable defects remain.

- Medium — Historical sticker assets can be lost. `crates/storage-sqlite/src/stickers.rs:69-71,111-155` returns when an older pack loses replacement ordering, before recording its immutable assets; both ingestion paths use this API (`crates/marmot-app/src/stickers.rs:713-743`). Exact-hash misses trigger relay resolution only when no current pack exists (`:332-355`). Consequently, an older message can permanently render `StickerNotFound` when a newer pack version arrived first. Record assets from every valid pack event independently of current-pack replacement, and perform bounded historical relay resolution whenever the exact `(coordinate, shortcode, hash)` asset is absent.

- Medium — Signal import publishes assets before deterministic validation completes. `crates/marmot-app/src/stickers.rs:543-557` uploads each plaintext sticker before validating later stickers; metadata and final pack construction follow at `:558-599`. A malformed later sticker or invalid pack therefore returns an error after earlier assets have been durably uploaded to Blossom. Validate all assets and construct a side-effect-free upload plan before the first upload; retain explicit handling for unavoidable partial failures during upload or later publication.

- Medium — Animated frame bounds are not validated. `crates/marmot-app/src/stickers.rs:1227-1238` limits only reported canvas dimensions; GIF parsing ignores descriptor offsets and dimensions at `:1346-1373`, while WebP parsing merely counts `ANMF` chunks at `:1427-1465`. Oversized, overflowing, or out-of-canvas frames can pass validation and cause downstream decoder failures or disproportionate allocation. Parse every GIF descriptor and WebP `ANMF` rectangle, rejecting zero-sized, overflowing, over-limit, or out-of-canvas frames.

Datawav:
PR health repair is blocked: this exact head conflicts with current `master` (`623892b8508774c42f334bf0ea83206a7af1c039`). I re-fetched the PR/base and verified the branch is still at `301f4450007f5eb741fbf8cfa2935e7d3985f1ef`, but the authenticated account has pull-only access to `vincenzopalazzo/mdk` (`push: false`), so I cannot safely merge and push the conflict resolution. The branch owner must merge current `master` into `codex/sonar-stickers` and resolve/push the conflicts, or grant this account push access to that fork branch.

Datawav (commented):
## Nightly deep review

Deep audit — `marmot-protocol/mdk#875` at `301f4450007f5eb741fbf8cfa2935e7d3985f1ef`

1. High — Relay-controlled discovery causes unbounded permanent database growth

Evidence:
- `sync_sticker_packs` fetches recent kind-30031 events and persists every valid pack, without requiring installation, an existing message reference, or prior user intent: `crates/marmot-app/src/stickers.rs:402-415,725-745`.
- `replace_sticker_pack_if_newer` permanently adds each distinct `(coordinate, shortcode, sha256)` to `app_sticker_assets`: `crates/storage-sqlite/src/stickers.rs:66-72,116-156`.
- `app_sticker_assets` retains every attacker-selected hash. Its only deletion path is cascading deletion of the parent pack, but this patch provides no pack deletion or history-pruning path: `crates/storage-sqlite/src/migrations/0047_sticker_asset_history.rs:7-20`.
- `MAX_DISCOVERY_PACKS` and per-pack limits bound one synchronization response, not cumulative storage. An attacker can continually publish fresh coordinates, use multiple authors, or publish winning replacements with fresh hashes.

Trigger and impact:
- Repeated synchronization against a relay carrying attacker-created valid packs causes the per-account encrypted SQLite database to grow indefinitely.
- The same account can be attacked repeatedly across launches; the in-memory mutation lock does not bound durable state.
- Eventually this can exhaust device storage or make migrations, backups, queries, and account startup increasingly expensive.

Why a design change is needed:
- A per-request limit cannot enforce a durable-state invariant.
- A per-author quota alone is insufficient because relay identities are cheap.
- Simply limiting history per coordinate conflicts with the requirement to resolve immutable references from old messages.

Required design:
- Separate bounded discovery metadata from durable, user-relevant state.
- Persist packs durably only when installed, explicitly fetched, or referenced by retained message history; otherwise keep discovery in a globally bounded LRU/cache.
- Add transactional global byte/row limits and garbage collection. Protect installed packs, pending outbox/install operations, current pack versions, and assets reachable from retained message references.
- Permit pruned historical assets to be recovered through bounded exact-reference fetching, as described in finding 2.

Alternatives:
- A strict global row/byte quota with deterministic eviction is acceptable if eviction preserves the protected roots above.
- Persisting all discovery remains viable only if it uses a separately bounded cache database that can be safely discarded.

Regression coverage:
- Synchronize more than the configured durable quota across fresh coordinates and replacement hashes; assert row and byte bounds remain enforced.
- Verify installed packs, pending publications, and assets referenced by retained messages survive eviction.
- Verify quota checks and ingestion occur in one transaction so concurrent fetches cannot overshoot the bound.

2. Medium — Historical sticker resolution fails after a newer pack version is stored

Evidence:
- `replace_sticker_pack_if_newer` returns before inserting assets whenever the incoming event loses NIP-01 replacement ordering: `crates/storage-sqlite/src/stickers.rs:69-71`.
- Historical asset insertion happens only while replacing the current pack: `crates/storage-sqlite/src/stickers.rs:111-156`.
- Both relay ingestion paths route every event through this replacement-only method: `crates/marmot-app/src/stickers.rs:713-743`.
- `fetch_sticker_asset` performs an on-demand relay fetch only when the entire pack is absent. If the current pack exists but the requested `(shortcode, hash)` does not, it immediately returns `StickerNotFound`: `crates/marmot-app/src/stickers.rs:332-355`.

Trigger and impact:
- A device receives the newest replacement event before an older valid version, or first encounters an old message after caching the current pack.
- The old event cannot contribute its immutable asset mapping, and the exact-hash miss does not trigger historical resolution.
- Valid older messages consequently lose their stickers even when relays return the signed historical event.

Why both paths must change:
- Ingesting assets from all events fixes out-of-order bulk synchronization but not an existing-pack exact-hash miss.
- Fetching on every miss without retaining historical mappings would repeatedly perform network work and still depend on relay ordering.

Required design:
- Split pack ingestion into two independently enforced operations:
  1. Insert validated immutable asset mappings from every accepted pack event using conflict-safe semantics.
  2. Replace `app_sticker_packs` and `app_stickers` only when `replacement_wins`.
- On an exact asset miss, perform a bounded coordinate-specific historical fetch even when a newer current pack exists, ingest all valid returned versions, and retry the exact lookup.
- Apply durable history bounds together with finding 1; old mappings should be retained when reachable and otherwise safely refetchable.

Alternatives:
- A dedicated historical-assets cache is suitable if exact references remain keyed by coordinate, shortcode, and plaintext hash.
- Storing full historical pack versions is unnecessary if only immutable asset metadata is required for rendering.

Regression coverage:
- Store a newer version first, then ingest an older version containing a different hash for the same shortcode. Assert the older exact reference resolves while the newer pack remains current.
- Start with only the newer version stored, request the old exact hash, return both versions from the relay, and assert bounded recovery succeeds.
- Exercise equal-timestamp event-ID ordering to ensure asset ingestion is independent of replacement convergence.

3. Medium — Signal import uploads plaintext assets before deterministic validation completes

Evidence:
- Each sticker is uploaded inside the same loop that validates subsequent stickers: `crates/marmot-app/src/stickers.rs:543-557`.
- Sticker metadata construction follows each upload, and cover selection, pack address construction, final `StickerPack::new`, signing, and event publication happen only after all uploads: `crates/marmot-app/src/stickers.rs:558-609`.

Trigger and impact:
- The first sticker is valid and uploaded, but a later sticker fails size, image, hash, or metadata validation.
- Final cover/pack construction or signing can also fail after every plaintext asset has been uploaded.
- The API returns an import failure while remotely durable Blossom content remains for a pack that was never successfully constructed or published. Retrying may repeat uploads.

Why restructuring is needed:
- Local validation failures are fully avoidable before the first irreversible network side effect.
- Moving validation earlier cannot make the complete distributed operation transactional: later upload, signing, or relay failures can still leave partial uploads.

Required design:
- First validate every asset, hash, image property, shortcode, metadata field, cover relationship, address, title/description, and final pack shape without network side effects.
- Produce a validated upload plan, then upload assets and substitute returned URLs into the final publication model.
- Explicitly define recovery for failures after uploads begin: idempotent content-addressed reuse, supported Blossom deletion/rollback, or durable resumable import state.

Alternatives:
- Content-addressed idempotent uploads can make retry reuse the same blobs, reducing duplicates, but they do not remove orphaned plaintext.
- A server supporting authenticated staging and commit/abort would provide a stronger publication boundary, at greater protocol complexity.

Regression coverage:
- Use a recording uploader or loopback Blossom server with a valid first sticker and malformed second sticker; assert validation fails with zero upload requests.
- Inject failure during a later upload, signing, and relay publication; verify the documented resume/cleanup behavior and that retries do not create uncontrolled additional state.

4. Medium — Animated GIF and WebP frame rectangles bypass dimension bounds

Evidence:
- The global validation applies limits only to dimensions returned by the format inspector: `crates/marmot-app/src/stickers.rs:1227-1238`.
- GIF parsing counts image descriptors but does not validate each descriptor’s left, top, width, or height fields: `crates/marmot-app/src/stickers.rs:1346-1373`.
- WebP parsing counts `ANMF` chunks without validating frame offsets and dimensions against the canvas: `crates/marmot-app/src/stickers.rs:1427-1465`.

Trigger and impact:
- A file declares an acceptable canvas but contains a zero-sized, oversized, overflowing, or out-of-canvas animation frame rectangle.
- It passes the patch’s validation despite the claimed exact dimension/frame enforcement.
- Downstream platform decoders may reject it, allocate disproportionate resources, or exhibit decoder-specific behavior, creating inconsistent rendering and avoidable resource risk.

Required design:
- Parse every GIF image descriptor and WebP `ANMF` frame rectangle.
- Reject zero dimensions, arithmetic overflow, dimensions above the configured per-axis/pixel limits, and rectangles extending beyond the declared canvas.
- Use checked arithmetic for all offset-plus-size and width-times-height calculations.

Alternatives:
- Fully decoding with a hardened image library under explicit memory, pixel, and frame budgets is safer than maintaining partial format parsers, but increases dependency and runtime cost.
- Retaining custom parsing is reasonable only with complete structural bounds for every frame-bearing chunk.

Regression coverage:
- Add GIF and WebP fixtures with oversized frames inside small canvases, out-of-canvas offsets, zero dimensions, maximum-value overflow, and boundary-valid rectangles.
- Assert malformed inputs are rejected before upload or rendering.

Reconciliation notes:
- The proposed Signal query-string credential leak is not supported by the supplied patch. The patch passes the URL to the pinned SDK, but no evidence shows that the SDK requests or logs that original URL; accepting a query is therefore a compatibility concern, not a demonstrated disclosure.
- Exact-reference send authorization is present: the outgoing sticker must already match a stored coordinate, shortcode, and plaintext hash.
- Per-account mutation locking and the installed-list outbox/rebase design address same-process ordering and stale snapshot convergence. No additional concrete fork/convergence defect was established from the supplied patches.
- `check_evidence.runs` is empty, so the bundle provides no successful build, test, clippy, migration, or concurrency-run evidence.

crates/marmot-app/src/stickers.rs
```diff
@@ -0,0 +1,1814 @@
...
+            Err(_) => continue,
+        };
+        if !seen.insert((
+            stored.coordinate.clone(),
+            stored.version.event_id_hex.clone(),
+        )) {
+            continue;
+        }
+        if storage.replace_sticker_pack_if_newer(&stored)? {
+            updated += 1;
```
Datawav, line 742: **High:** This persists every valid discovered pack, while `MAX_DISCOVERY_PACKS` only limits one synchronization response. Relays can continually supply fresh coordinates or replacement hashes, and `app_sticker_assets` has no global row/byte quota or pruning path, so the encrypted database can grow without bound across syncs and launches. Keep uninstalled discovery in a bounded cache, or enforce transactional global limits and deterministic GC while protecting installed packs, pending operations, current versions, and assets referenced by retained messages.

crates/marmot-app/src/stickers.rs
```diff
@@ -0,0 +1,1814 @@
...
+        let mut stickers_by_id = HashMap::new();
+        let mut stickers = Vec::with_capacity(imported.stickers.len());
+        for imported_sticker in &imported.stickers {
+            validate_sticker_asset_size(&imported_sticker.bytes)?;
+            let inspected = inspect_image(&imported_sticker.bytes)?;
+            if sha256_hex(&imported_sticker.bytes) != imported_sticker.sha256 {
+                return Err(invalid_sticker("Signal sticker hash mismatch"));
+            }
+            let url = upload_blossom_blob(
+                server,
```
Datawav, line 550: **Medium:** Uploading inside this loop creates an irreversible network side effect before validation of later stickers and before cover, address, final pack construction, and signing can succeed. A malformed later sticker therefore returns an import error after plaintext assets have already been uploaded, and retries may repeat the uploads. Validate the complete pack into a side-effect-free upload plan first, then upload; also define idempotent reuse, cleanup, or resumable state for failures after uploading begins.

crates/marmot-app/src/stickers.rs
```diff
@@ -0,0 +1,1814 @@
...
+                dimensions = Some(((bits & 0x3fff) + 1, ((bits >> 14) & 0x3fff) + 1));
+            }
+            b"VP8 " if length >= 10 && bytes.get(data + 3..data + 6) == Some(b"\x9d\x01\x2a") => {
+                dimensions = Some((
+                    u32::from(le_u16(bytes, data + 6)? & 0x3fff),
+                    u32::from(le_u16(bytes, data + 8)? & 0x3fff),
+                ));
+            }
+            b"ANMF" => frames = frames.saturating_add(1),
+            _ => {}
```
Datawav, line 1456: **Medium:** `ANMF` chunks are only counted; their frame offsets and dimensions are never parsed or checked against the declared canvas. The GIF path similarly skips image-descriptor rectangle fields. An animation can therefore advertise a small valid canvas while containing zero-sized, oversized, overflowing, or out-of-canvas frames. Parse every frame rectangle and use checked arithmetic to reject zero dimensions, per-axis/pixel-limit violations, overflow, and `offset + size` beyond the canvas.

Datawav:
The pull request currently contains unsigned merge commit `356f06fa636ec7885bf9dc2f2a699391221efc45` (`Merge upstream/master into codex/sonar-stickers`). GitHub reports `verification.verified=false` with reason `unsigned`.

Please rewrite or remove that merge commit so every commit in the pull request is signed by a GitHub-recognized key, then force-push the repaired branch with an exact lease. Afterward, verify the pull request commits through GitHub and require `commit.verification.verified=true` for the rewritten commit before review continues.

Datawav (changes requested):
Required CI fails at this head (301f4450), so the draft cannot merge as-is. Verifiable failures on the latest run:

1. `Rust clippy`: `error: unused import: Deserialize` at `crates/cgka-engine/src/engine.rs:46` (`use tls_codec::{Deserialize as _, Serialize as _};`) — the engine no longer uses `Deserialize` on that import after this branch's changes; drop it from the import.
2. `Rust tests (2/4)`: `thread 'failed-leave-push-compensation' has overflowed its stack` (SIGABRT, 529/1091 tests run before nextest cancels). The same harness-thread stack overflow appears on #1611's head and not on current master CI, so it may be a shared stale-base artifact — but it must reproduce green on a rebase before merge.
3. `Rust check`, `Required CI`, and `iOS account keychain clippy` fail transitively from the above.

The sticker feature scope (storage migrations 0046/0047, stickers module, timeline integration, UniFFI/C surfaces across 36 files) is large for one PR; given the 16 prior review rounds, please also rebase onto current master so reviewers see a clean diff against green CI.

Reviewed by Hermes (openai-codex/gpt-5.6-sol via review profile) — job-log verification of the failing checks at 301f4450 plus diff-stat review; no local build run.

PR #889, opened 2026-07-16
fix(cgka-engine): phantom committed_from from failed commit staging poisons fork detection

### LINKED ISSUE #885: cgka-engine: a failed/rolled-back commit-producing send leaves a stale `committed_from` guard, degrading a later legitimate same-epoch fork into a fail-closed quarantine
**Severity:** MEDIUM • **Category:** bug (correctness / availability)

**Primary site:** `crates/cgka-engine/src/message_processor/send.rs:145-146` (and the sibling commit-producing sends: remove-members `send.rs:467-468`, `upgrade.rs:161-162`, `update_group_data.rs:213-214`).
**Missing cleanup:** `crates/cgka-engine/src/epoch_manager.rs:224-244` (`rollback_publish`) and `crates/cgka-engine/src/pending_commit_guard.rs` (`PendingCommitCleanupGuard::drop`) — neither clears `committed_from`; the only removal is horizon-based `prune_committed_from_before` (`epoch_manager.rs:255-270`).
**Consumers of the stale guard:** `crates/cgka-engine/src/message_processor/ingest.rs:517` and `:592-609`.

## What

Every commit-producing send records the source epoch in the fork-detection guard **before** any fallible work:

```rust
// send.rs:145-146 (do_send_invite), before add_members / wrap / record_sent_openmls_message / begin_pending
self.epoch_manager
    .record_committed_from(&group_id, pre_commit_epoch);
```

`record_committed_from` inserts `pre_commit_epoch` into `committed_from` (`epoch_manager.rs:248-253`). Everything that follows it in the send is fallible: `add_members` (`send.rs:156`), the async `peeler.wrap_group_message` (`:167`), `record_sent_openmls_message` — a **durable write that can return `SQLITE_BUSY`** (`:179`), the welcome-wrap loop, `put_group`, and finally `begin_pending`. If any returns `Err`, the function returns early; `PendingCommitCleanupGuard::drop` clears the orphaned OpenMLS pending commit and releases the fork-recovery snapshot — **but nothing clears the `committed_from` entry.**

The same holds after a *successful* stage followed by a failed transport publish: `publish_failed → rollback_publish` (`epoch_manager.rs:224-244`) returns the group to `Stable{prior_epoch}` and removes the `pending` entry, but does **not** touch `committed_from`, and no incumbent recovery record is created (that only happens in `do_confirm_published` via `promote_pending_commit_for_recovery`). So after either failure, `committed_from(N)` is set with **no** retained fork snapshot and **no** incumbent recovery record for epoch `N`.

## Why it matters

`committed_from` is the routing guard between the two same-epoch-fork paths. With it left stale-true for epoch `N`:

1. **Clean convergence is blocked.** A legitimate past-epoch commit at `N` (`msg_epoch < current`) only enters stored convergence when `!we_committed_from(group, N)` (`ingest.rs:514-519`). Stale-true makes that `false`, so the message is *not* routed into convergence.
2. **It falls into fork-recovery, which has no snapshot.** The message then hits `process_message → WrongEpoch`. The branch at `ingest.rs:592-595` fires because `we_committed_from(&group_id, N)` is stale-true and `current > N`. `recovery_snapshot_name_for_fork(N)` returns `None` (no record was ever promoted), so `resolve_fork_candidate` returns `ForkResolution::MissingSnapshot` → `forked_epoch_fail_closed` (`ingest.rs:606-609`) → the group is quarantined to `Recovering` and the ingest returns `EngineError::ForkedEpoch`.

With a correctly-cleared `committed_from`, step 1 would instead route the commit into stored convergence (given a retained anchor snapshot at `N`) and resolve the reorg cleanly — exactly the case distributed convergence exists to handle.

## Concrete failure scenario

1. `do_send_invite` (or remove / upgrade / update_group_data) at epoch `N` fails after `record_committed_from(N)` — e.g. a transient `SQLITE_BUSY` on `record_sent_openmls_message`, a peeler error, or a subsequent transport `publish_failed`. `committed_from(N)` is now stale; the group stays `Stable{N}`; no incumbent recovery record and no retained fork snapshot for `N` exist.
2. Group history advances past `N` via other traffic (e.g. a peer's commit at `N` applied in-order → tip `N+1`).
3. A genuine late/reordered epoch-`N` fork sibling arrives while `current == N+1`. Because `we_committed_from(N)` is stale-true, it is blocked from convergence (step 1 above), takes the WrongEpoch fork branch, finds no snapshot, and quarantines the group.

The quarantine clears only through `retry_hydrate_quarantined_group` or an authenticated re-join welcome. The trigger is remotely influenceable: a peer who publishes a late/duplicate distinct epoch-`N` commit after the victim's own send fails converts a recoverable reorg into an availability-impacting quarantine. Transport publish failures (the `publish_failed → rollback_publish` path) are routine, making the stale-guard window common.

## Suggested fix

Clear the `committed_from` entry for `pre_commit_epoch` whenever a staged commit is discarded without being confirmed:
- in `do_publish_failed` / `rollback_publish` (after `forget_pending_commit_for_recovery`), and
- in `PendingCommitCleanupGuard::drop` after the orphaned pending commit is cleared,

guarding against clearing it when another live pending/incumbent still legitimately owns that epoch (e.g. a concurrent staged commit or an existing incumbent recovery record for `N`).

None of the 136 open issues mention `committed_from`, the send-failure / `publish_failed` rollback path, or the convergence-vs-fork-recovery mis-routing. Closest neighbors are distinct: **#394** turned a fork-recovery `unreachable!` into a fail-closed error (this is about *why the fail-closed path is wrongly reached* — a stale routing guard — not the panic-vs-error handling); **#824** is a torn projected *record* before `begin_pending` in `do_send_update_group_data` (unrelated to `committed_from`); **#805** is `do_join_welcome` MLS-orphan state. The fork issues #841/#801 are in the separate `incident-replay` crate.

dannym-arx:
marmot

Four send paths (invite, remove-members, capability upgrade, group-data update) recorded "we committed from this epoch" while commit staging could still fail. When staging failed, the cleanup guard cleared the staged OpenMLS commit, but the committed_from entry stayed behind. A marmot that later advanced past that epoch through a peer's commit (settled via convergence, so it holds no fork-recovery snapshot of its own) then mis-routed a legitimate same-epoch sibling commit: the stale entry blocked convergence entry, the WrongEpoch fork branch found no recovery snapshot, and ingest failed closed with ForkedEpoch. The group got stuck in Recovering and rejected further sends.

begin_pending already records committed_from atomically with the pending-publish transition, so the four manual calls were redundant on success and harmful on failure. This PR removes them and deletes the now-unused EpochManager::record_committed_from.

Tests:

- cgka-engine: failed_invite_staging_does_not_poison_fork_detection reproduces the chain (failed duplicate-member invite, peer-driven advance via convergence, sibling commit for the past epoch). It fails with ForkedEpoch on the pre-fix engine and passes with the fix.
- cgka-conformance-simulator: failed_invite_staging_does_not_poison_fork_detection_via_harness mirrors the scenario across five marmots on the bus, with a partition keeping the sibling committer at the old epoch. Adds HarnessClient::try_invite so a scenario can drive a deliberately failing invite; registered in SCENARIOS.md.

Verified with just fast-ci, cargo test -p cgka-engine, and cargo test -p cgka-conformance-simulator.

fixes #885 

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/889">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed failed invite staging/publishing no longer poisoning fork detection or leaving the group stuck in recovery.
  * Corrected rollback behavior so provisional epoch bookkeeping is removed when appropriate, while preserving existing committed-history state.
  * Improved invite handling to report underlying send failures gracefully rather than crashing or misclassifying outcomes.
  * Groups remain usable after a failed invite followed by a concurrent sibling commit.
* **Documentation**
  * Added a conformance scenario covering the failed-invite regression case.

### DIFF
--- a/crates/cgka-engine/src/message_processor/send.rs
+++ b/crates/cgka-engine/src/message_processor/send.rs
@@ -124,8 +124,14 @@ impl<S: StorageProvider> Engine<S> {
             parsed_kps.push(parsed);
         }
 
-        // Record the pre-commit epoch so a later same-epoch commit can be
-        // compared against this locally produced branch.
+        // The pre-commit epoch is the commit origin for fork detection.
+        // `committed_from` bookkeeping is deliberately NOT recorded here: it
+        // happens atomically inside `begin_pending` below, after staging
+        // succeeds. Recording it earlier left a phantom "we committed from
+        // this epoch" entry behind when staging failed (the cleanup guard
+        // clears the OpenMLS commit but nothing prunes `committed_from`),
+        // which later mis-routed legitimate sibling commits into
+        // fail-closed fork recovery.
         let pre_commit_epoch = EpochId(mls_group.epoch().as_u64());
         // Arm the cleanup guard before creating the snapshot so the snapshot is
         // released on early return / cancellation even before a pending commit
@@ -142,8 +148,6 @@ impl<S: StorageProvider> Engine<S> {
             pre_commit_epoch,
             "pre_invite_commit",
         );
-        self.epoch_manager
-            .record_committed_from(&group_id, pre_commit_epoch);
         let pre_commit_ctx =
             crate::group_lifecycle::build_group_context_snapshot(&mls_group, &provider)?;
 
@@ -464,8 +468,6 @@ impl<S: StorageProvider> Engine<S> {
         )?;
         let commit_priority =
             crate::app_components::commit_ordering_priority_for_staged(staged_commit);
-        self.epoch_manager
-            .record_committed_from(&group_id, pre_commit_epoch);
 
         let commit_bytes = commit_out
             .tls_serialize_detached()
--- a/crates/cgka-engine/src/message_processor/send.rs
+++ b/crates/cgka-engine/src/message_processor/send.rs
@@ -131,7 +131,9 @@ impl<S: StorageProvider> Engine<S> {
         // this epoch" entry behind when staging failed (the cleanup guard
         // clears the OpenMLS commit but nothing prunes `committed_from`),
         // which later mis-routed legitimate sibling commits into
-        // fail-closed fork recovery.
+        // fail-closed fork recovery. The post-staging counterpart —
+        // `publish_failed` after a successful `begin_pending` — is handled by
+        // `rollback_publish`, which removes the provisional entry again.
         let pre_commit_epoch = EpochId(mls_group.epoch().as_u64());
         // Arm the cleanup guard before creating the snapshot so the snapshot is
         // released on early return / cancellation even before a pending commit
--- a/crates/cgka-conformance-simulator/SCENARIOS.md
+++ b/crates/cgka-conformance-simulator/SCENARIOS.md
@@ -190,6 +190,17 @@ These are real simulator scenarios that are still tied to Rust harness details.
   set.
 - Reason: the test inspects recovery observations and exact branch ordering keys.
 
+### `failed_invite_staging_does_not_poison_fork_detection_via_harness`
+
+- Setup: Alice's invite fails during commit staging (duplicate-member KeyPackage). Alice then advances to epoch 2 by
+  settling Bob's commit through convergence, and a partitioned Dave later delivers a sibling commit from the same
+  source epoch.
+- Expected: the sibling commit is adjudicated deterministically (stale losing branch or reorg onto the winner) — never
+  a fail-closed `ForkedEpoch` — and Alice's group stays usable for a follow-up invite.
+- Reason: regression guard for the phantom `committed_from` bookkeeping bug (fixed by recording it only inside
+  `begin_pending`); mirrors the cgka-engine test `failed_invite_staging_does_not_poison_fork_detection` at the
+  multi-client bus level.
+
 ### `convergence-e2e-group-events/v1`
 
 - Setup: Alice creates a four-member group. Alice and Bob race invite commits and each branch has an app payload. Carol
--- a/crates/cgka-conformance-simulator/src/client.rs
+++ b/crates/cgka-conformance-simulator/src/client.rs
@@ -567,6 +567,17 @@ impl HarnessClient {
 
     /// Invite new members to the default group.
     pub async fn invite(&mut self, kps: Vec<KeyPackage>) -> PendingStateRef {
+        self.try_invite(kps).await.expect("send invite")
+    }
+
+    /// Invite new members to the default group, surfacing the engine error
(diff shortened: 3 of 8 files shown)


crates/cgka-engine/src/message_processor/send.rs
```diff
@@ -124,8 +124,14 @@ impl<S: StorageProvider> Engine<S> {
             parsed_kps.push(parsed);
         }
 
-        // Record the pre-commit epoch so a later same-epoch commit can be
-        // compared against this locally produced branch.
+        // The pre-commit epoch is the commit origin for fork detection.
+        // `committed_from` bookkeeping is deliberately NOT recorded here: it
+        // happens atomically inside `begin_pending` below, after staging
```
erskingardner, line 129: This fixes failures before `begin_pending`, but not a later `publish_failed`. `begin_pending` still inserts the source epoch into `committed_from`; rollback clears the staged commit and drops its recovery record/snapshot, but never removes that epoch. I reproduced: stage an invite → `publish_failed` → advance via a peer commit → ingest a late same-epoch sibling, which still returns `ForkedEpoch`. Please remove the rolled-back pending commit’s provisional `committed_from` ownership (without clearing any legitimate confirmed incumbent/other owner) and add a rollback regression test.

erskingardner (commented):
One correctness gap remains in the rollback path.

crates/cgka-engine/src/message_processor/send.rs
dannym-arx: Fixed in 9a0aeec. `rollback_publish` now removes the provisional `committed_from` entry that its `begin_pending` inserted. Ownership is tracked in `PendingMeta.owns_committed_from` (set from whether the `begin_pending` insert was new; a group holds at most one live pending, so a fresh insert is solely owned by it), an entry that already existed, e.g. a confirmed incumbent after a fork-recovery rewind to the same epoch, is left in place.

Tests added:
- `epoch_manager::tests::rollback_publish_removes_provisional_committed_from`
- `epoch_manager::tests::rollback_publish_preserves_confirmed_committed_from_incumbent`
- `fork_detection::publish_failed_rollback_does_not_poison_fork_detection`. your exact repro (stage invite → `publish_failed` → advance via peer commit → ingest late same-epoch sibling). Verified it fails with `ForkedEpoch` when the rollback pruning is disabled and passes with it.

erskingardner (approved):

PR #960, opened 2026-07-21
feat(cli): add --reply-to to messages send

jgmontoya:
Route replies through the runtime's existing reply_to_message, which emits the same q and e tags timeline ingest parses, so replies from the CLI are wire-identical to other Marmot clients. A reply to a parent that has not synced yet sends fine and hydrates its preview when the parent arrives.

A --reply-to placed after the message text would previously be swallowed as literal text by the hyphen-tolerant trailing argument; it now fails loudly with error code reply_to_after_message_text. JSON response shapes are unchanged; no new runtime API was needed.

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/960">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added `--reply-to <message-id>` to `wn messages send` and legacy `wn message send` for sending replies.
  * Replies keep the original-message link and show a hydrated reply preview when the parent message becomes available (even if it isn’t stored locally yet).
* **Bug Fixes**
  * Validates argument ordering: placing `--reply-to` after the message text now fails loudly with error code `reply_to_after_message_text` instead of being sent as body text.
  * Prevents sending messages containing a bare `--reply-to` / `--reply-to=<id>` token in the text.

### DIFF
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -31,6 +31,16 @@ versioning through the workspace version in the root `Cargo.toml`.
   `wn daemon start`). They are forwarded to the `daemon start` child and to `create-identity`/`login` account setup
   so a first run with no relay configuration can supply relays without dead-ending. Flag passthrough only; no JSON
   response shapes changed.
+- `wn messages send` (and the older `wn message send`) accept `--reply-to <message-id>` to send the text as a reply to
+  an existing message. The reply carries the same `q`/`e` reference tags other Marmot clients produce, so recipients'
+  timelines parse it into `reply_to_message_id` with a hydrated `reply_preview`. `--reply-to` is additive input only:
+  the JSON response shape is unchanged, pass the group with `--group` when replying, and a reply to a not-yet-synced
+  parent is still sent (its preview hydrates when the parent arrives). No new runtime API was needed; the CLI routes
+  reply sends through the existing `MarmotAppRuntime::reply_to_message`. Because the trailing message text is
+  hyphen-tolerant, a `--reply-to` placed *after* the text is read as literal text; rather than silently sending that
+  stray flag as part of the body, the send now fails loudly with error code `reply_to_after_message_text` (put
+  `--reply-to` before the text with `--group`). Tradeoff: message text that is itself exactly `--reply-to` can no
+  longer be sent.
 - MarmotKit/UniFFI now exposes encrypted per-account composer draft storage with metadata-only list, full load, upsert,
   and delete operations. Drafts retain their text, reply target, ordered attachment bytes, and attachment presentation
   metadata in the account's SQLCipher database; attachment bytes are loaded only for the selected draft.
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -37,10 +37,11 @@ versioning through the workspace version in the root `Cargo.toml`.
   the JSON response shape is unchanged, pass the group with `--group` when replying, and a reply to a not-yet-synced
   parent is still sent (its preview hydrates when the parent arrives). No new runtime API was needed; the CLI routes
   reply sends through the existing `MarmotAppRuntime::reply_to_message`. Because the trailing message text is
-  hyphen-tolerant, a `--reply-to` placed *after* the text is read as literal text; rather than silently sending that
-  stray flag as part of the body, the send now fails loudly with error code `reply_to_after_message_text` (put
-  `--reply-to` before the text with `--group`). Tradeoff: message text that is itself exactly `--reply-to` can no
-  longer be sent.
+  hyphen-tolerant, a `--reply-to` placed *after* the text (either spelling: `--reply-to <id>` or `--reply-to=<id>`) is
+  read as literal text; rather than silently sending that stray flag as part of the body, the send now fails loudly with
+  error code `reply_to_after_message_text` (put `--reply-to` before the text with `--group`). Tradeoff: the guard
+  rejects any message whose text contains a bare `--reply-to` or `--reply-to=<id>` token anywhere (e.g.
+  `hello --reply-to friend`), so such text can no longer be sent this way.
 - MarmotKit/UniFFI now exposes encrypted per-account composer draft storage with metadata-only list, full load, upsert,
   and delete operations. Drafts retain their text, reply target, ordered attachment bytes, and attachment presentation
   metadata in the account's SQLCipher database; attachment bytes are loaded only for the selected draft.
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -214,6 +214,7 @@ Message commands:
 ```sh
 wn --account <npub-or-hex> messages send <group-hex> "hello"
 wn --account <npub-or-hex> messages send --group <group-hex> "--text that starts with a dash"
+wn --account <npub-or-hex> messages send --group <group-hex> --reply-to <message-id> "replying to you"
 wn --account <npub-or-hex> messages list
 wn --account <npub-or-hex> messages list <group-hex> --limit 20
 wn --account <npub-or-hex> messages list <group-hex> --before <unix-seconds> --before-message-id <event-id>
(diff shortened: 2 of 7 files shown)


ghost:
Ready to review this PR? Stage has broken it down into 6 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Define error for misplaced reply-to flags](https://stagereview.app/marmot-protocol/mdk/pull/960/chapters/1) |
| 2 | [Add reply-to argument to CLI parser](https://stagereview.app/marmot-protocol/mdk/pull/960/chapters/2) |
| 3 | [Implement reply-to validation and routing](https://stagereview.app/marmot-protocol/mdk/pull/960/chapters/3) |
| 4 | [Verify reply-to behavior with unit tests](https://stagereview.app/marmot-protocol/mdk/pull/960/chapters/4) |
| 5 | [Add integration tests for message replies](https://stagereview.app/marmot-protocol/mdk/pull/960/chapters/5) |
| 6 | [Update documentation and changelog](https://stagereview.app/marmot-protocol/mdk/pull/960/chapters/6) |

<a href="https://stagereview.app/marmot-protocol/mdk/pull/960">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 01a8d1234185e80ff2c7d0dde444c7f165d6c47a on Jul 22, 2026 3:26pm UTC.</sub>

erskingardner:
## Adversarial review

Reviewed at head commit `b2966ee9`. Overall this is well-built — scoped to the CLI crate, no new runtime surface, and genuinely thorough tests — but there is one verified correctness bug that should be fixed before merging.

### 1. Bug (verified by reproduction): `--reply-to=<id>` equals-form bypasses the misplaced-flag guard

`reject_misplaced_reply_to` (`crates/cli/src/commands/messages.rs:51-58`) exists to fail loudly when `--reply-to` lands after the message text and gets absorbed as literal text. But it only matches the exact token `"--reply-to"`. clap's `--flag=value` spelling survives in the trailing args as a single token like `"--reply-to=PARENT"`, which never matches:

```
argv: wn messages send --group GROUP hello --reply-to=PARENT
parsed -> reply_to=None, args=["hello", "--reply-to=PARENT"]
reject_misplaced_reply_to(None, args) -> Ok(())
```

So that command sends successfully with literal body `"hello --reply-to=PARENT"` and no reply linkage — the exact silent footgun this guard was added to prevent, just via the `=` spelling. The positional-group form (`wn message send GROUP hello --reply-to=PARENT`) hits the same gap.

**Suggested fix:** also match the prefix form, e.g.

```rust
token == "--reply-to" || token.starts_with("--reply-to=")
```

plus a parse-lock/guard unit test for the equals form alongside the existing `message_send_stray_reply_to_in_text_is_rejected` so it can't regress silently.

### 2. Docs: the tradeoff note undersells the guard's actual scope

The sentence "message text that is itself exactly `--reply-to` can no longer be sent" appears in `crates/cli/CHANGELOG.md`, `crates/cli/README.md`, and the comment above the guard. It reads as if only a single-word message is affected, but the code rejects **any** message containing `--reply-to` as a standalone token anywhere — e.g. `wn messages send --group G hello --reply-to friend` is rejected today. That broader behavior is arguably correct for the footgun, but the docs should state it plainly (something like "a message containing the literal token `--reply-to` can no longer be sent as plain text") so a user who hits the rejection on a multi-word message isn't confused — and so the wording stays consistent with whichever fix is chosen for the `=`-form gap above.

### 3. Minor (non-blocking)

The `repair` hint key for `ReplyToAfterMessageText` in `crates/cli/src/error.rs` is `"reply"`, which is generic next to sibling repair keys (`"retry_with_relay"`, `"supply_both"`). Something like `"reorder"` or `"usage"` would fit the existing convention better.

### What's done well

- Reuses the existing `MarmotAppRuntime::reply_to_message` — no new runtime API, change stays scoped to the CLI crate.
- Strong test coverage: parse-lock unit tests documenting the exact clap quirk, and a real relay round-trip integration test asserting the wire-format `q` tag and timeline-side hydrated reply preview end-to-end.
- `messages_send_reply_to_unknown_parent_still_sends` explicitly covers the not-yet-synced-parent causal-ordering edge case.
- CHANGELOG and README updated together under `Unreleased`, per crate convention; JSON output shape kept identical for plain sends vs. replies, and that's tested.
- `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and all tests pass at head.

jgmontoya:
> ## Adversarial review
> 
> Reviewed at head commit `b2966ee9`. Overall this is well-built — scoped to the CLI crate, no new runtime surface, and genuinely thorough tests — but there is one verified correctness bug that should be fixed before merging.
> ### 1. Bug (verified by reproduction): `--reply-to=<id>` equals-form bypasses the misplaced-flag guard
> 
> `reject_misplaced_reply_to` (`crates/cli/src/commands/messages.rs:51-58`) exists to fail loudly when `--reply-to` lands after the message text and gets absorbed as literal text. But it only matches the exact token `"--reply-to"`. clap's `--flag=value` spelling survives in the trailing args as a single token like `"--reply-to=PARENT"`, which never matches:
> 
> ```
> argv: wn messages send --group GROUP hello --reply-to=PARENT
> parsed -> reply_to=None, args=["hello", "--reply-to=PARENT"]
> reject_misplaced_reply_to(None, args) -> Ok(())
> ```
> 
> So that command sends successfully with literal body `"hello --reply-to=PARENT"` and no reply linkage — the exact silent footgun this guard was added to prevent, just via the `=` spelling. The positional-group form (`wn message send GROUP hello --reply-to=PARENT`) hits the same gap.
> 
> **Suggested fix:** also match the prefix form, e.g.
> 
> ```rust
> token == "--reply-to" || token.starts_with("--reply-to=")
> ```
> 
> plus a parse-lock/guard unit test for the equals form alongside the existing `message_send_stray_reply_to_in_text_is_rejected` so it can't regress silently.
> ### 2. Docs: the tradeoff note undersells the guard's actual scope
> 
> The sentence "message text that is itself exactly `--reply-to` can no longer be sent" appears in `crates/cli/CHANGELOG.md`, `crates/cli/README.md`, and the comment above the guard. It reads as if only a single-word message is affected, but the code rejects **any** message containing `--reply-to` as a standalone token anywhere — e.g. `wn messages send --group G hello --reply-to friend` is rejected today. That broader behavior is arguably correct for the footgun, but the docs should state it plainly (something like "a message containing the literal token `--reply-to` can no longer be sent as plain text") so a user who hits the rejection on a multi-word message isn't confused — and so the wording stays consistent with whichever fix is chosen for the `=`-form gap above.
> ### 3. Minor (non-blocking)
> 
> The `repair` hint key for `ReplyToAfterMessageText` in `crates/cli/src/error.rs` is `"reply"`, which is generic next to sibling repair keys (`"retry_with_relay"`, `"supply_both"`). Something like `"reorder"` or `"usage"` would fit the existing convention better.
> ### What's done well
> 
>     * Reuses the existing `MarmotAppRuntime::reply_to_message` — no new runtime API, change stays scoped to the CLI crate.
> 
>     * Strong test coverage: parse-lock unit tests documenting the exact clap quirk, and a real relay round-trip integration test asserting the wire-format `q` tag and timeline-side hydrated reply preview end-to-end.
> 
>     * `messages_send_reply_to_unknown_parent_still_sends` explicitly covers the not-yet-synced-parent causal-ordering edge case.
> 
>     * CHANGELOG and README updated together under `Unreleased`, per crate convention; JSON output shape kept identical for plain sends vs. replies, and that's tested.
> 
>     * `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and all tests pass at head.
> 
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

All three points verified by reproduction and addressed:
1. Confirmed, the = spelling bypassed the guard exactly as you described (both command surfaces). The guard now also matches the --reply-to= prefix, with parse-lock and guard tests for the equals form so it can't regress (ce90c8c3).
2. Confirmed, the guard rejects the bare token anywhere in the text, not just a single-word message. The guard comment, README, and CHANGELOG now state the actual rule, covering both spellings (2d9bafd4).
3. Renamed the repair key to reorder per the action-phrase convention; nothing consumed the old key programmatically (6d6da808).

Good catch on the equals form, thanks for the concrete repro.

erskingardner (approved):

PR #1046, opened 2026-07-23
Close remaining convergence, daemon, and dial-safety gaps

### LINKED ISSUE #807: marmot-app: live message-subscription dedupe doesn't guard empty message_id_hex, unlike the recovery path — two empty-id live updates collide and the second is silently dropped
## Summary

In the message subscription's broadcast loop, the **recovery** path guards against an empty `message_id` before inserting it into the seen-set, but the **live** path does not. If two distinct live updates ever carry an empty `message_id_hex`, the first inserts `""` into `seen_message_ids` and every subsequent empty-id update collides and is silently dropped rather than delivered.

## Location

`crates/marmot-app/src/runtime/subscriptions.rs`

Recovery path (guards empty id):
```rust
let message_id = message.message_id_hex.clone();
if !message_id.is_empty() && !seen_message_ids.insert(message_id) {   // line ~833
    continue;
}
```

Live path (no empty-id guard):
```rust
if !seen_message_ids.insert(message.message_id_hex.clone()) {          // line ~857
    continue;
}
```

## Failure scenario

The recovery path's explicit `!message_id.is_empty()` guard shows that empty ids are considered possible for these rows. On the live path the same guard is missing, so the first empty-id update pins `""` in the seen-set and any further distinct empty-id updates are dropped as "already seen."

## Severity

**LOW.** Impact is bounded — live `summary.messages` normally always carry a canonical app-message id, so an empty id should be rare in practice. This is primarily a defensive-consistency gap: the two dedupe sites that share the same `seen_message_ids` set should apply the same empty-id rule.

## Suggested fix

Mirror the recovery guard on the live path:
```rust
let message_id = message.message_id_hex.clone();
if !message_id.is_empty() && !seen_message_ids.insert(message_id) {
    continue;
}
```

### LINKED ISSUE #817: cli: daemon subscription output has no write deadline — a stalled/backpressured same-UID subscriber pins its worker task and connection permit indefinitely
## Summary

The daemon bounds the request **read** (`DAEMON_REQUEST_READ_TIMEOUT = 30s`), but every subscription response **write** is unbounded. A subscriber that stops draining its socket (a slow or stuck downstream consumer) fills the kernel send buffer and makes `write_all` block forever, pinning the worker task and its shared connection permit. Enough stalled subscribers exhaust the connection semaphore, at which point `daemon status`/`daemon stop` are refused at accept.

## Location

`crates/cli/src/daemon/protocol.rs:372`

```rust
pub(crate) async fn write_stream_response(
    stream: &mut UnixStream,
    response: &DaemonStreamResponse,
) -> bool {
    let Ok(mut bytes) = serde_json::to_vec(response) else { return false; };
    bytes.push(b'\n');
    stream.write_all(&bytes).await.is_ok()   // no timeout
}
```

Called on every update from the subscription loops in `crates/cli/src/daemon/subscriptions.rs`; the connection permit (`let _permit = permit;`) is held for the whole connection in `crates/cli/src/daemon/mod.rs`.

## Failure scenario

`wn messages subscribe | slow_consumer`. The client (`stream_request`, `protocol.rs`) blocks on its own stdout flush when the downstream consumer is slow and stops reading the socket. The kernel socket buffer fills and the daemon's `write_all` blocks indefinitely, holding the worker task and its `connection_limiter` permit. Because control requests share the same semaphore, enough stalled subscribers make new control connections (including `daemon status`/`daemon stop`) get closed at accept.

Non-malicious backpressure triggers it; a hostile local client can hold slots open deliberately.

- #779 is the identical missing-write-deadline class but in `transport-quic-broker` (a different crate/accept loop).
- #718 is about control requests **sharing** the semaphore with legitimately long-lived subscriptions; its remedy is a separate control lane and does not address a write that never completes. This missing write-deadline is the mechanism that turns #718 from a transient into a permanent hang.

## Suggested fix

Wrap each `write_stream_response`/`write_all` in `tokio::time::timeout(...)` (mirroring `read_daemon_request_within`); treat a timeout as connection failure so the worker returns and releases its permit.

---
*Filed from a meticulous automated code-review pass across all crates; independently verified against the source and deduplicated against existing open/closed issues.*

### LINKED ISSUE #855: cli: one-shot daemon response writes (write_daemon_output) have no write deadline — a non-reading same-UID client can pin connection permits (control-plane DoS)
### Summary

The daemon bounds the request **read** with `DAEMON_REQUEST_READ_TIMEOUT`, but the response **write** (`write_daemon_output`) has no timeout. A same-UID client that sends a valid request for a large result and then stops reading blocks the worker in `write_all` while still holding its connection-limiter permit. Repeating this exhausts the 256-permit pool and makes legitimate `wn daemon status` / `stop` fail.

### Location

- `crates/cli/src/daemon/protocol.rs:313-320` — `write_daemon_output`:
  ```rust
  pub(crate) async fn write_daemon_output(stream: &mut UnixStream, output: &CliOutput) {
      let Ok(mut response) = serde_json::to_vec(output) else { return; };
      response.push(b'\n');
      let _ = stream.write_all(&response).await;   // no timeout
      let _ = stream.shutdown().await;             // no timeout
  }
  ```
- Reached for `Status`/`Ping`/`Shutdown`/`Execute`/`StreamWatch` from `crates/cli/src/daemon/mod.rs:245,264,280,283,294,427,441,472,486,493,503`.
- The permit is held for the whole handler: `mod.rs:200-217` acquires `connection_limiter.try_acquire_owned()` and moves it into the spawned task (`let _permit = permit;` at `:215`), which runs `handle_daemon_connection` — including the untimed write. `MAX_DAEMON_CONNECTIONS = 256` (`mod.rs:70`); the request side is bounded by `read_daemon_request_within(DAEMON_REQUEST_READ_TIMEOUT)` (`mod.rs:257`).

### Failure scenario

1. A same-UID client opens a connection, sends a **valid** `Execute` for a command with a large JSON result (e.g. `messages list --limit 200 --json`, `groups list`), then never drains the socket.
2. Once the response exceeds the socket send buffer, the worker blocks in `write_all` while holding its permit.
3. Repeat up to `MAX_DAEMON_CONNECTIONS`. New connections are refused at accept (`try_acquire_owned` fails, `mod.rs:200-207` drops the stream with no response), so `wn daemon status` reports "not running" and `wn daemon stop` fails.

### Why not a duplicate

#817 is scoped to the streaming subscription loop (`write_stream_response` in `subscriptions.rs`); #718 is about the semaphore being *shared* between subscriptions and control. This is the separate one-shot `write_daemon_output` path, which neither covers — and the request side already got a matching read timeout while this write side did not. (Same vulnerability class/fix category as #817, but a distinct code path.)

### Suggested severity

**MEDIUM** — same-UID local self-DoS of the daemon control plane (consistent with #817's severity).

### Suggested fix

Wrap `write_daemon_output`'s `write_all`/`shutdown` in `tokio::time::timeout(...)` (mirroring `read_daemon_request_within`), dropping the connection on elapse so the permit is released.

---
*Filed from a meticulous automated code-review pass. The cited `file:line` locations were verified against source, and the finding was checked against existing open/closed issues for duplicates before filing.*

### LINKED ISSUE #977: Security review M8: architecture docs diverge from code and the adopted baseline on selection rules and policy
## Summary

The two "draft normative" architecture docs list 5 branch-comparison rules ending at tip digest; the code implements the full 7-rule order from the adopted protocol baseline, including `tip_priority` and `tip_committer`. The docs also describe the convergence policy as group-negotiated with group-chosen quorum values, while the adopted baseline pins v1 constants.

Found in the 2026-07-20 security review (`docs/security-review-2026-07-20.md`, finding M8). Status: **confirmed.**

## Locations

- `docs/marmot-architecture/cgka-engine-spec.md:255-261` and `docs/marmot-architecture/distributed-convergence.md:134-141` — 5 comparison rules
- `crates/cgka-engine/src/convergence.rs:129-138` — the implemented 7 rules (effective depth → quorum → raw depth → witness score → priority → committer → digest)
- `docs/marmot-architecture/distributed-convergence.md:157-177` — "policy should be group-negotiated", quorum fields "<group policy>"

## Attack scenario / impact

A second implementation written from these docs would tie-break equal-score same-tip races differently (admin-removal vs targeted member self-update cases) and fork the network. The negotiated-policy text also invites a remote policy-downgrade vector if implemented literally (see the M1 issue).

## Suggested remediation

Update both docs to the 7-rule order and the pinned v1 policy (or clearly mark negotiated policy as future work requiring a new app component behind a required capability, per the baseline).

### LINKED ISSUE #992: Security review M23: tie-breaker adjacency app_witness_score vs tip_priority is unpinned — swapping rules 4/5 passes the whole suite
## Summary (coverage gap)

Every adjacent pair in the 7-rule branch-selection chain is pinned by at least one test *except* `app_witness_score` (rule 4) before `tip_priority` (rule 5). Swapping rules 4 and 5 in `compare_scores` passes the entire current suite — a rule-order regression that flips branch selection in contested forks would ship undetected.

Found in the 2026-07-20 security review (`docs/security-review-2026-07-20.md`, finding M23). Status: **coverage-gap.**

## Locations

- `crates/cgka-engine/src/convergence.rs:129-138` — the 7-rule chain
- Existing adjacency tests: witness-score>digest (`crates/cgka-conformance-simulator/tests/candidate_state_graph.rs:19`), priority>digest+committer (`:237`), committer>digest (`:270`), effective-depth>all (`:59`) — the witness-score vs priority pair is missing

## Attack scenario / impact (if regressed)

A rule-order regression flips branch selection in contested forks → cross-implementation divergence (one side selects the witnessed branch, the other the privileged one) → permanent group split that no conformance test detects.

## Suggested remediation

Add one symbolic test with tied depth/quorum, `app_witness_score` 2 vs 0, and the 0-score branch Privileged; assert the witnessed Ordinary branch wins.

### LINKED ISSUE #996: Security review L2: WN_DEV_SETTLEMENT_QUIESCENCE_MS overrides the pinned quiescence constant in production builds with no warning
## Summary

An env var lets any local operator/devtool shrink the settlement window (e.g. to 0) in release builds; nothing logs or warns when it is set. Unlike other dev hooks (loopback relays/blobs), this one alters a *protocol-pinned convergence constant*, not just a local transport choice.

## Location

`crates/cli/src/lib.rs:1017-1022`, wired into the production config at `:984-988`

## Impact

A misconfigured power-user install settles branch races instantly, selecting branches before competing commits arrive — user-induced convergence divergence that looks like an engine bug.

## Suggested remediation

Emit a startup warning when set, or gate it behind a compile-time dev feature.

---
Found in the 2026-07-20 security review (`docs/security-review-2026-07-20.md`, finding L2). Status: **confirmed.**

### LINKED ISSUE #1035: traits: is_public_ipv6 accepts IANA 2001::/23 special-purpose blocks (incl. IPv6 benchmarking 2001:2::/48) as public — IPv4 classifier rejects the equivalent (dial-safety/SSRF parity gap)
### Summary

`is_public_ipv6` carves only Teredo (`2001:0000::/32`), 6to4 (`2002::/16`), and the documentation ranges out of the `2001::/23` IETF special-purpose allocation, then falls through to a terminal "anything in global-unicast `2000::/3` is public" rule. That accepts every *other* IANA special-purpose sub-block of `2001::/23` as a public unicast address — most notably **IPv6 benchmarking `2001:2::/48` (RFC 5180)**, whose IPv4 equivalent the sibling classifier explicitly rejects. This is a concrete cross-family (v4/v6) inconsistency in the shared dial-safety/SSRF gate.

### Location

`crates/traits/src/app_components/host_safety.rs:91-120` (terminal rule at line 119: `(first & 0xe000) == 0x2000`).

```rust
// handled:
if first == 0x2002 || (first == 0x2001 && second == 0x0000) { return false; } // 6to4, Teredo
if first == 0x2001 && second == 0x0db8 { return false; }                       // 2001:db8::/32 docs
if (first & 0xfff0) == 0x3ff0 { return false; }                               // 3fff::/20 docs
// terminal:
(first & 0xe000) == 0x2000                                                     // accepts all of 2000::/3
```

### Failure scenario (trace for `2001:2::…`)

`first = 0x2001`, `second = 0x0002`:
- not `0x2002`, not `(0x2001 && 0x0000)`, not `(0x2001 && 0x0db8)`, not `(first & 0xfff0)==0x3ff0`
- terminal: `(0x2001 & 0xe000) == 0x2000` → **`true` (public)**

Other `2001::/23` special-purpose sub-blocks that likewise fall through to `true`:
- ORCHIDv2 `2001:20::/28` (RFC 7343)
- AS112-v6 `2001:4:112::/48` (RFC 7535)
- Port Control Protocol anycast `2001:1::1` / `2001:1::2`

By contrast, `is_public_ipv4` (lines 67-86) **does** reject the IPv4 benchmarking range via `(198, 18..=19, _, _)`. So the same logical address class is rejected for v4 and accepted for v6.

Because `reject_non_public_ip` / the avatar-url and encrypted-media endpoint validators trust this classifier, a group-signed avatar/media endpoint or a resolved dial target inside `2001:2::/48` (etc.) passes the "public unicast only" gate that the IPv4 path would have blocked.

### Why not a duplicate

The only filed IPv6-classification issue, **#913**, is the *opposite direction* on a *different range* — an over-broad **reject** of the documentation range (`3fff::/20` mask too wide). This is an over-broad **accept** of the `2001::/23` special-purpose blocks and, specifically, a v4/v6 asymmetry (benchmarking rejected for v4, accepted for v6). Checked #784, #850, #913, #914, and the #708 dial-safety tracking issue — none covers it.

### Suggested severity

**LOW** — defense-in-depth / consistency. These blocks are rarely routed on the public internet, but the classifier is a security boundary and its IPv4 sibling already rejects the equivalent, so the asymmetry is a real hardening gap.

### Suggested fix

Before the terminal `2000::/3` rule, reject the non-global `2001::/23` IETF special-purpose allocations. Minimally treat `2001:2::/48` (benchmarking) the same way the IPv4 side treats `198.18.0.0/15`; consider rejecting the whole non-global `2001::/23` set (ORCHIDv2, AS112-v6, PCP anycast) for parity. Add a v4/v6 parity test asserting benchmarking is rejected in **both** families.

---
*Filed from a meticulous automated code-review pass. The classifier trace was verified against source, and the finding was checked against existing open issues (#913, #784, #850, #914, #708) for duplicates before filing.*

### LINKED ISSUE #1041: marmot-account: welcome_delivery_failure does a redundant stored_sent_welcome storage read whose result the redeliver_welcome caller immediately overwrites (dead durable read)
### Summary

`welcome_delivery_failure` internally calls `self.session.stored_sent_welcome(&message_id)` solely to populate the failure's `group_id`. But its `redeliver_welcome` caller has *already* loaded `(group_id, message)` and, after building the failure, overwrites `failure.group_id = Some(group_id)` with that already-known value — so the internal lookup is dead work (its result is discarded), and every failing welcome pays an extra durable engine read.

### Location

`crates/marmot-account/src/runtime.rs`

- `:582-605` `redeliver_welcome`: loads `(group_id, message)` at `:586`, then overwrites `failure.group_id = Some(group_id)` at `:601`.
- `:609-636` `welcome_delivery_failure`: independently re-reads `group_id` at `:625-629`:

```rust
let group_id = self
    .session
    .stored_sent_welcome(&message_id)     // durable read...
    .ok()
    .map(|(group_id, _)| group_id);
```

In the `redeliver_welcome` path the value produced here is thrown away by the `:601` overwrite. The same helper is also reached from `publish_group_created` / `publish_group_evolution`, which pay the same extra read per failing welcome.

### Failure scenario

Not a correctness bug — a code-quality / micro-performance issue: a redundant durable `stored_sent_welcome` read on the welcome-failure path, plus two sources of truth for the same `group_id` (the helper's internal lookup and the caller's overwrite), which invites future drift if one path changes.

### Why not a duplicate

No corpus issue touches `welcome_delivery_failure` / `redeliver_welcome` read redundancy (the welcome re-delivery issues concern the *feature*, not this double lookup).

### Suggested severity

**LOW** — refactor / micro-performance and readability.

### Suggested fix

Pass the already-known `group_id` into `welcome_delivery_failure` (e.g. accept an `Option<GroupId>`), and drop the internal `stored_sent_welcome` call — giving one source of truth for `group_id` and removing the redundant durable read.

---
*Filed from a meticulous automated code-review pass. The caller/helper `group_id` flow was verified against source, and the finding was checked against existing open issues for duplicates before filing.*

erskingardner:
## Summary

- align the local convergence architecture documentation with the adopted seven-rule v1 ordering and add the missing symbolic witness-score adjacency test
- keep the development quiescence override out of release behavior
- apply bounded whole-response writes to the remaining streaming and one-shot daemon paths
- reuse known welcome group IDs instead of performing a redundant durable read
- reject the complete IANA `2001::/23` IPv6 special-purpose range while preserving the adjacent public allocation
- bypass live message deduplication when an update has no canonical message ID

## Why

These are the small, clear follow-up gaps left by the broader MDK issue audit. The changes apply patterns and policies already established elsewhere in the workspace, while keeping each issue in a separate bisectable commit.

## Validation

- `just fast-ci`
- `cargo test -p wn-cli -p marmot-account -p marmot-app -p cgka-traits`

Fixes #977
Fixes #992
Fixes #996
Fixes #817
Fixes #855
Fixes #1041
Fixes #1035
Fixes #807

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced a 15-second write deadline for daemon framed responses and stream replies (including flush/shutdown) to prevent stalled clients from blocking output.
  * Fixed subscription update deduplication so empty message IDs no longer suppress later updates.
  * Corrected welcome-delivery failure reporting with the confirmation-derived group identifier.
  * Strengthened public/non-public IPv6 address handling and related URL validation rules.
  * Release builds now reject the development-only settlement timing override at startup.
* **Documentation**
  * Clarified v1 pinned protocol convergence/determinism and canonical branch selection.
* **Tests**
  * Added/updated coverage for daemon write timeouts, branch scoring/tie-breaking, IPv6 safety, subscription dedupe behavior, and build-specific override handling.

### DIFF
--- a/crates/cgka-conformance-simulator/tests/candidate_state_graph.rs
+++ b/crates/cgka-conformance-simulator/tests/candidate_state_graph.rs
@@ -234,6 +234,50 @@ fn digest_tiebreak_picks_lower_digest_after_commit_and_witness_ties() {
 }
 
 #[test]
+fn app_witness_score_beats_priority_after_depth_and_quorum_ties() {
+    let policy = ConvergencePolicy {
+        max_rewind_commits: 5,
+        witness_quorum_senders_per_epoch: 10,
+        witness_quorum_epochs: 2,
+        max_witness_override_depth: 2,
+    };
+    let witnessed_ordinary = BranchCandidate {
+        id: "witnessed-ordinary".into(),
+        fork_epoch: 1,
+        tip_epoch: 3,
+        tip_priority: CommitOrderingPriority::Ordinary,
+        tip_committer: b"z".to_vec(),
+        tip_digest: digest(0xff),
+        app_witnesses: vec![witness(2, "alice"), witness(2, "bob")],
+    };
+    let unwitnessed_privileged = BranchCandidate {
+        id: "unwitnessed-privileged".into(),
+        fork_epoch: 1,
+        tip_epoch: 3,
+        tip_priority: CommitOrderingPriority::Privileged,
+        tip_committer: b"a".to_vec(),
+        tip_digest: digest(0x00),
+        app_witnesses: vec![],
+    };
+    let witnessed_score = witnessed_ordinary.score(&policy);
+    let privileged_score = unwitnessed_privileged.score(&policy);
+    let candidates = [witnessed_ordinary.clone(), unwitnessed_privileged];
+
+    assert_eq!(witnessed_score.effective_commit_depth, 2);
+    assert_eq!(
+        witnessed_score.effective_commit_depth,
+        privileged_score.effective_commit_depth
+    );
+    assert!(!witnessed_score.witness_quorum_met);
+    assert!(!privileged_score.witness_quorum_met);
+    assert_eq!(witnessed_score.app_witness_score, 2);
+    assert_eq!(privileged_score.app_witness_score, 0);
+
+    let winner = select_canonical_branch(3, &candidates, &policy).expect("one branch should win");
+    assert_eq!(winner.id, witnessed_ordinary.id);
+}
+
+#[test]
 fn privileged_tiebreak_beats_digest_after_depth_and_witness_ties() {
     let policy = ConvergencePolicy {
         max_rewind_commits: 5,
--- a/crates/cli/src/daemon/protocol.rs
+++ b/crates/cli/src/daemon/protocol.rs
@@ -330,6 +330,11 @@ pub(crate) async fn write_daemon_output(stream: &mut UnixStream, output: &CliOut
     let _ = write_daemon_output_within(stream, output, DAEMON_RESPONSE_WRITE_TIMEOUT).await;
 }
 
+/// Write, flush, and close one daemon response within one deadline.
+///
+/// `false` is connection-fatal: the future may have been dropped after a
+/// partial frame was written, so callers must drop the stream and never retry
+/// or write another frame on it.
 pub(crate) async fn write_daemon_output_within<W>(
     stream: &mut W,
     output: &CliOutput,
@@ -410,6 +415,11 @@ pub(crate) async fn write_stream_response(
     write_stream_response_within(stream, response, DAEMON_RESPONSE_WRITE_TIMEOUT).await
 }
 
+/// Write and flush one streaming response within one deadline.
+///
+/// `false` is connection-fatal: the future may have been dropped after a
+/// partial frame was written, so callers must drop the stream and never retry
+/// or write another frame on it.
 pub(crate) async fn write_stream_response_within<W>(
     stream: &mut W,
     response: &DaemonStreamResponse,
--- a/crates/cli/src/daemon/protocol.rs
+++ b/crates/cli/src/daemon/protocol.rs
@@ -389,11 +389,29 @@ pub(crate) async fn write_stream_response(
     stream: &mut UnixStream,
     response: &DaemonStreamResponse,
 ) -> bool {
+    write_stream_response_within(stream, response, DAEMON_RESPONSE_WRITE_TIMEOUT).await
+}
+
+pub(crate) async fn write_stream_response_within<W>(
+    stream: &mut W,
+    response: &DaemonStreamResponse,
+    timeout_after: Duration,
+) -> bool
+where
+    W: AsyncWrite + Unpin,
+{
     let Ok(mut bytes) = serde_json::to_vec(response) else {
         return false;
     };
     bytes.push(b'\n');
-    stream.write_all(&bytes).await.is_ok()
+    matches!(
+        tokio::time::timeout(timeout_after, async {
+            stream.write_all(&bytes).await?;
+            stream.flush().await
+        })
+        .await,
+        Ok(Ok(()))
+    )
 }
 
 pub(crate) async fn write_stream_end(stream: &mut UnixStream) -> bool {
--- a/crates/cli/src/daemon/protocol.rs
+++ b/crates/cli/src/daemon/protocol.rs
@@ -327,12 +327,30 @@ impl DaemonClient {
 }
 
 pub(crate) async fn write_daemon_output(stream: &mut UnixStream, output: &CliOutput) {
+    let _ = write_daemon_output_within(stream, output, DAEMON_RESPONSE_WRITE_TIMEOUT).await;
+}
+
+pub(crate) async fn write_daemon_output_within<W>(
+    stream: &mut W,
+    output: &CliOutput,
+    timeout_after: Duration,
+) -> bool
+where
+    W: AsyncWrite + Unpin,
+{
     let Ok(mut response) = serde_json::to_vec(output) else {
-        return;
+        return false;
     };
     response.push(b'\n');
-    let _ = stream.write_all(&response).await;
-    let _ = stream.shutdown().await;
+    matches!(
+        tokio::time::timeout(timeout_after, async {
+            stream.write_all(&response).await?;
+            stream.flush().await?;
+            stream.shutdown().await
+        })
+        .await,
+        Ok(Ok(()))
+    )
 }
 
 pub(crate) async fn read_daemon_request(
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -1032,10 +1032,19 @@ fn wn_allow_loopback_relays() -> bool {
     )
 }
 
+fn resolve_dev_settlement_quiescence_ms(
+    value: Option<&str>,
+    dev_overrides_enabled: bool,
+) -> Option<u64> {
+    if !dev_overrides_enabled {
+        return None;
+    }
+    value.and_then(|value| value.trim().parse().ok())
+}
+
 fn wn_dev_settlement_quiescence_ms() -> Option<u64> {
-    std::env::var("WN_DEV_SETTLEMENT_QUIESCENCE_MS")
-        .ok()
-        .and_then(|value| value.trim().parse().ok())
+    let value = std::env::var("WN_DEV_SETTLEMENT_QUIESCENCE_MS").ok();
+    resolve_dev_settlement_quiescence_ms(value.as_deref(), cfg!(debug_assertions))
 }
 
 fn open_account_home(
@@ -1190,7 +1199,8 @@ mod tests {
     };
     use super::{
         Cli, Command, StreamCommand, WnError, daemon, daemon_socket_for_client,
-        default_home_from_env, npub_for_account_id, relay_endpoints, resolve_relay, run_from,
+        default_home_from_env, npub_for_account_id, relay_endpoints,
+        resolve_dev_settlement_quiescence_ms, resolve_relay, run_from,
     };
 
     use marmot_app::{
@@ -1268,6 +1278,28 @@ mod tests {
             .expect_err("invalid account ids must surface as a CLI error");
         let rendered = super::wn_error_json(&err);
         assert_eq!(rendered["code"], "invalid_public_key");
+    }
+
+    #[test]
+    fn dev_settlement_override_is_available_in_debug_builds() {
+        assert_eq!(
+            resolve_dev_settlement_quiescence_ms(Some("0"), true),
+            Some(0)
+        );
+    }
+
+    #[test]
+    fn dev_settlement_override_is_rejected_in_release_builds() {
+        assert_eq!(resolve_dev_settlement_quiescence_ms(Some("0"), false), None);
+    }
+
+    #[test]
+    fn invalid_dev_settlement_override_remains_inactive_without_a_warning() {
+        assert_eq!(
+            resolve_dev_settlement_quiescence_ms(Some("not-a-duration"), false),
+            None
+        );
+        assert_eq!(resolve_dev_settlement_quiescence_ms(None, false), None);
     }
 
     fn test_cli(command: Command) -> Cli {
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -437,7 +437,7 @@ async fn execute_inner(cli: Cli) -> Result<CommandOutput, WnError> {
             .or_else(|| cli.daemon_discovery_relays.first().cloned())
             .or_else(|| cli.daemon_default_account_relays.first().cloned()),
         account_home.clone(),
-    );
+    )?;
     match command {
         Command::Debug { command } => {
             commands::debug::debug_command(&account_home, &app, command, account_flag)
@@ -990,7 +990,11 @@ pub(crate) fn relay_lists_json(status: AccountRelayListStatus) -> Value {
     })
 }
 
-fn app_for(home: PathBuf, relay: Option<String>, account_home: AccountHome) -> MarmotApp {
+fn app_for(
+    home: PathBuf,
+    relay: Option<String>,
+    account_home: AccountHome,
+) -> Result<MarmotApp, WnError> {
     // Loopback-HTTP blob endpoints are only acted on when explicitly enabled for
     // dev/test (see MarmotAppConfig::allow_loopback_blob_endpoints). Opt in via
     // WN_ALLOW_LOOPBACK_BLOB_ENDPOINTS=1 for local Blossom servers; production
@@ -1002,15 +1006,15 @@ fn app_for(home: PathBuf, relay: Option<String>, account_home: AccountHome) -> M
     // convergence settlement window (e.g. `0` for instant settlement in
     // integration tests). Production installs leave it unset and use the pinned
     // default.
-    if let Some(ms) = wn_dev_settlement_quiescence_ms() {
+    if let Some(ms) = wn_dev_settlement_quiescence_ms()? {
         config = config.with_dev_settlement_quiescence_ms(ms);
     }
-    MarmotApp::with_relays_and_account_home_and_config(
+    Ok(MarmotApp::with_relays_and_account_home_and_config(
         home,
         relay.into_iter().collect(),
         account_home,
         config,
-    )
+    ))
 }
 
 fn wn_allow_loopback_blob_endpoints() -> bool {
@@ -1035,14 +1039,14 @@ fn wn_allow_loopback_relays() -> bool {
 fn resolve_dev_settlement_quiescence_ms(
     value: Option<&str>,
     dev_overrides_enabled: bool,
-) -> Option<u64> {
-    if !dev_overrides_enabled {
-        return None;
+) -> Result<Option<u64>, WnError> {
+    if value.is_some() && !dev_overrides_enabled {
+        return Err(WnError::DevSettlementOverrideInRelease);
     }
-    value.and_then(|value| value.trim().parse().ok())
+    Ok(value.and_then(|value| value.trim().parse().ok()))
 }
 
-fn wn_dev_settlement_quiescence_ms() -> Option<u64> {
+fn wn_dev_settlement_quiescence_ms() -> Result<Option<u64>, WnError> {
     let value = std::env::var("WN_DEV_SETTLEMENT_QUIESCENCE_MS").ok();
     resolve_dev_settlement_quiescence_ms(value.as_deref(), cfg!(debug_assertions))
 }
@@ -1283,23 +1287,36 @@ mod tests {
     #[test]
     fn dev_settlement_override_is_available_in_debug_builds() {
         assert_eq!(
-            resolve_dev_settlement_quiescence_ms(Some("0"), true),
+            resolve_dev_settlement_quiescence_ms(Some("0"), true).unwrap(),
             Some(0)
         );
     }
 
     #[test]
     fn dev_settlement_override_is_rejected_in_release_builds() {
-        assert_eq!(resolve_dev_settlement_quiescence_ms(Some("0"), false), None);
+        let error =
+            resolve_dev_settlement_quiescence_ms(Some("0"), false).expect_err("release rejection");
+        assert!(matches!(&error, WnError::DevSettlementOverrideInRelease));
+        assert_eq!(
+            super::wn_error_json(&error)["code"],
+            "dev_settlement_override_in_release"
+        );
+        assert_eq!(
+            resolve_dev_settlement_quiescence_ms(None, false).unwrap(),
+            None
+        );
     }
 
     #[test]
-    fn invalid_dev_settlement_override_remains_inactive_without_a_warning() {
+    fn invalid_or_missing_dev_settlement_override_remains_inactive_when_enabled() {
         assert_eq!(
-            resolve_dev_settlement_quiescence_ms(Some("not-a-duration"), false),
+            resolve_dev_settlement_quiescence_ms(Some("not-a-duration"), true).unwrap(),
             None
         );
-        assert_eq!(resolve_dev_settlement_quiescence_ms(None, false), None);
+        assert_eq!(
+            resolve_dev_settlement_quiescence_ms(None, true).unwrap(),
+            None
+        );
     }
 
     fn test_cli(command: Command) -> Cli {
--- a/crates/marmot-account/src/runtime.rs
+++ b/crates/marmot-account/src/runtime.rs
@@ -547,7 +547,9 @@ where
                 .push(PendingResolution::Confirmed { pending });
             output.absorb_session_effects(effects, queue);
             for failure in &mut welcome_failures {
-                failure.group_id = group_id.clone();
+                if failure.group_id.is_none() {
+                    failure.group_id = group_id.clone();
+                }
             }
             // Only a confirmed create leaves welcomes worth re-delivering; a
             // rolled-back create discards the whole evolution, welcomes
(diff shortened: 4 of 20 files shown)


docs/marmot-architecture/cgka-engine-spec.md
```diff
@@ -371,19 +373,21 @@ The engine MUST NOT emit an invalidated message as a normal received message.
 
 ## Policy
 
-Convergence policy is group policy. Engines MUST persist the negotiated policy per group and load it before convergence
-after restart.
-
-The policy contains:
+The adopted v1 convergence policy is pinned by the protocol:
```
erskingardner, line 379: **Incomplete M8 / #977 closure.**

This page (and `distributed-convergence.md`) correctly switch to pinned v1 constants, but on this branch `docs/marmot-architecture/cgka-engine-canonicalization-contract.md` still says engines MUST persist a negotiated per-group policy, treat local defaults as a fallback, and store override bytes — and `relay-delivery-telemetry.md` still calls `settlement_quiescence_ms` group-negotiated.

That contract doc is at least as implementer-facing as the two updated here. Shipping `Fixes #977` while it still teaches the old model reopens the exact divergence/policy-downgrade risk M8 filed. Please update the remaining normative text in this PR, or drop the Fixes line until that cohort is actually closed.

crates/marmot-app/src/runtime/subscriptions.rs
```diff
@@ -854,7 +854,8 @@ impl MarmotAppRuntime {
                 {
                     continue;
                 }
-                if !seen_message_ids.insert(message.message_id_hex.clone()) {
+                let message_id = message.message_id_hex.clone();
+                if !message_id.is_empty() && !seen_message_ids.insert(message_id) {
```
erskingardner: **Missing regression test for #807.**

The live path now correctly mirrors the recovery empty-id guard — good. But this is a one-line behavior fix with no coverage. The original bug was exactly “recovery guarded it, live didn’t”; please add a small assertion that two distinct empty-`message_id_hex` live updates both emit (and do not pin `""` in `seen_message_ids`), so the asymmetry cannot recur unnoticed.

crates/cli/src/lib.rs
```diff
@@ -1032,10 +1032,19 @@ fn wn_allow_loopback_relays() -> bool {
...
+    }
+    value.and_then(|value| value.trim().parse().ok())
+}
+
 fn wn_dev_settlement_quiescence_ms() -> Option<u64> {
-    std::env::var("WN_DEV_SETTLEMENT_QUIESCENCE_MS")
-        .ok()
-        .and_then(|value| value.trim().parse().ok())
+    let value = std::env::var("WN_DEV_SETTLEMENT_QUIESCENCE_MS").ok();
+    resolve_dev_settlement_quiescence_ms(value.as_deref(), cfg!(debug_assertions))
```
erskingardner, line 1051: **Silent release ignore is operator-hostile (and CLI-only).**

`cfg!(debug_assertions)` meets the letter of #996, but:

1. When the env var is set on a release binary, ignore it *loudly* once (stderr / privacy-safe tracing). The new test even names “without a warning” as desired behavior; #996 offered warning *or* gate, and silent ignore makes a production misconfig look like an engine bug again.
2. This only closes the CLI env path. `MarmotAppConfig::with_dev_settlement_quiescence_ms` remains callable from UniFFI/host apps in release — if the threat model is “no production client may shrink quiescence,” that needs the same posture or an explicit follow-up.

Minor: `invalid_dev_settlement_override_remains_inactive_without_a_warning` passes `dev_overrides_enabled = false`, so it never exercises parse failure on the enabled path.

crates/marmot-account/src/runtime.rs
```diff
@@ -840,6 +843,21 @@ fn welcome_recipient(message: &TransportMessage) -> Option<MemberId> {
     }
 }
 
+/// Extract the group id from the event that confirms a create/evolution.
+///
+/// `confirm_published` emits exactly one `GroupCreated` or `EpochChanged`
+/// event for the pending operation. Keeping this best-effort preserves the
+/// existing `WelcomeDeliveryFailure::group_id` contract without re-reading the
+/// durable sent-welcome record.
+fn confirmed_group_id(effects: &SessionEffects) -> Option<GroupId> {
```
erskingardner, line 857: **Thinner failure metadata than the old durable lookup.**

Passing the known group id is the right refactor, and `confirm_published` force-inserts the confirm event today, so this should work. The create path now depends on that event shape instead of `stored_sent_welcome` — if confirm effects ever lack `GroupCreated` / `EpochChanged`, welcome failures regress to `group_id: None` where the old helper would still populate them.

Consider a fallback to the stored welcome record when this returns `None`, or a focused test that a partial-welcome create failure still carries a group id after confirm.

crates/cgka-conformance-simulator/tests/candidate_state_graph.rs
```diff
@@ -233,6 +233,50 @@ fn digest_tiebreak_picks_lower_digest_after_commit_and_witness_ties() {
     assert_eq!(winner.id, lower_digest.id);
 }
 
+#[test]
+fn app_witness_score_beats_priority_after_depth_and_quorum_ties() {
```
erskingardner, line 237: **This adjacency pin looks right — ignore the harness/default-policy nits.**

Using a custom high-quorum policy so both sides stay below quorum is the correct way to isolate rule 4 vs 5; default v1 quorum with two witnesses would trip quorum and spoil the adjacency. And this file already exercises `select_canonical_branch` directly via the crate re-export (as AGENTS.md describes) — rewriting one case onto `HarnessClient` would be inconsistent, not a correctness fix.

erskingardner (commented):
Adversarial pass on the gap-closure batch.

**Verdict: would request changes** (GitHub blocks REQUEST_CHANGES on own PRs) — primarily because this still claims `Fixes #977` / M8 while `cgka-engine-canonicalization-contract.md` (and parts of `relay-delivery-telemetry.md`) continue to prescribe negotiated per-group policy. That leaves the exact cross-implementation divergence risk the finding called out.

The runtime fixes themselves look solid: daemon whole-frame write deadlines (incl. flush/shutdown) with stall tests, IPv6 `2001::/23` rejection that correctly keeps `2001:200::/23`, and the witness-score-before-priority adjacency pin are all in good shape. I would not chase the CodeRabbit notes about rewriting `candidate_state_graph` onto `HarnessClient` or retiming `write_daemon_server_busy` — that file already pins the selector directly by design, and busy rejection is already wrapped in a 100ms accept-loop timeout without holding a permit.

Please either finish the M8 doc cohort (or drop the Fixes line until you do) and add a regression test for the empty-id live dedupe fix.

docs/marmot-architecture/cgka-engine-spec.md
```diff
@@ -396,7 +400,7 @@ Required storage:
 
 - group metadata and current epoch,
 - OpenMLS state,
-- negotiated convergence policy and engine version,
+- convergence policy version, pinned values, and engine version,
```
erskingardner: Residual docs/code divergence in a divergence-fixing PR: nothing in the workspace persists a convergence policy version, the pinned values, or an engine version (no matches in `storage-sqlite` or `cgka-engine`). The old bullet had the same problem, but the rewrite also makes “pinned values” an odd thing to require in storage — if the values are protocol constants, persisting them per group buys nothing; at most a policy *version* stamp matters for forward-compat detection. Suggest either dropping “pinned values” and keeping only the version stamp requirement, or marking this bullet as not-yet-implemented so the spec's Required storage section stays an accurate contract.

crates/marmot-account/src/runtime.rs
erskingardner, line 857: Two robustness notes on the first-match heuristic, neither blocking:

1. `effects.events` comes from `drain_events()`, which drains the engine's whole event buffer — the confirm's own event plus anything else queued. Today that's safe (the runtime drains at every session call, and `do_confirm_published` pushes its create/epoch event before the `GroupStateChanged` events it synthesizes), but `find_map` on the *first* `GroupCreated`/`EpochChanged` would silently mislabel the failure if a future engine change ever leaves another group's event in the buffer ahead of the confirm event. Since the doc comment on this function is what encodes that assumption, consider strengthening it to say the safety depends on effects being drained per-call, so the next person who batches drains knows this is a load-bearing invariant.

2. The backfill loop at line 549–551 unconditionally overwrites `failure.group_id`, including with `None`. Correct today because every failure in that vec was constructed with `None`, but it would clobber a `Some` if a future path pre-populated one. `failure.group_id = group_id.clone().or(failure.group_id.take())` would make the backfill monotone for free.

crates/cli/src/lib.rs
erskingardner, line 1051: The release-build behavior here is silent-ignore, not reject, and #996 called out “with no warning” as part of the problem — an operator who sets `WN_DEV_SETTLEMENT_QUIESCENCE_MS` on a release install now gets pinned-default behavior with zero signal about why their override does nothing. A one-time `tracing::warn!` (aggregate-only: the fact that the var was set and dropped, not its value's semantics) when `value.is_some() && !dev_overrides_enabled` would close that gap cheaply.

Two related nits: the CHANGELOG entry says release builds “reject” the variable, which reads as an error exit — “ignore” is the accurate word. And every integration test that sets `WN_DEV_SETTLEMENT_QUIESCENCE_MS=0` now silently degrades to the pinned 1000 ms quiescence if the suite is ever run with `--release` — slow/flaky rather than loudly failing. Fine as long as CI never does that, but worth knowing the failure mode exists.

crates/cli/src/daemon/protocol.rs
```diff
@@ -327,12 +327,30 @@ impl DaemonClient {
 }
 
 pub(crate) async fn write_daemon_output(stream: &mut UnixStream, output: &CliOutput) {
+    let _ = write_daemon_output_within(stream, output, DAEMON_RESPONSE_WRITE_TIMEOUT).await;
+}
+
+pub(crate) async fn write_daemon_output_within<W>(
```
erskingardner, line 338: Worth a doc comment on both `_within` helpers now that they're generic and reusable: on timeout the inner future is dropped mid-write, so `false` can mean a *partial frame is already on the wire* (and for the one-shot path, that `shutdown` never ran — the fd only closes when the caller drops the stream). Every current caller treats `false` as connection-fatal, which is exactly right — but nothing states that contract, and a future caller that retries or keeps writing after `false` would corrupt the newline framing. One sentence saying “`false` is connection-fatal; do not write again on this stream” would pin it.

crates/marmot-app/src/runtime/subscriptions.rs
erskingardner: Optional: this is the only fix in the PR that ships without a test. A small unit test that pushes two distinct empty-id updates through the live path and asserts both are delivered would pin the guard against the exact regression #807 describes (first empty id poisoning the seen-set). Given the issue is rated LOW/defensive-consistency, fine to skip — flagging only because everything else here is pinned.

erskingardner (commented):
Adversarial review — I checked out the head branch, read every change against its surrounding code, and ran the touched test suites locally (`candidate_state_graph`, the new `wn-cli` daemon/settlement unit tests, `cgka-traits` lib + `host_safety_public`, and the full `marmot-account` suite — all green).

**What I tried to break, and couldn't:**

- **Seven-rule ordering (#977/#992):** the docs' new tie-break list matches `compare_scores` in `crates/cgka-engine/src/convergence.rs` exactly, including the reversed comparisons (Privileged < Ordinary via enum declaration order, lower committer, lower digest). The pinned v1 values in both docs match `ConvergencePolicy::default()`, and the reworked `max_witness_override_depth = 1` examples are arithmetically consistent. The new adjacency test genuinely pins rule 4 over rule 5: swapping them flips the winner. The `protocol-core/convergence.md` link resolves in the marmot repo.
- **Daemon write deadlines (#817/#855):** `write_stream_end` routes through `write_stream_response`, so the end-of-stream frame is also bounded — no unbounded write path remains. Every streaming caller in `daemon/subscriptions.rs` treats `false` as connection-fatal, so a timed-out partial frame can't be followed by more writes on a corrupted framing boundary. The one-shot path preserves the pre-existing `shutdown` inside the deadline.
- **IPv6 `2001::/23` (#1035):** `second <= 0x01ff` is the correct /23 boundary; `2001:200::/23` stays public, `2001:db8::/32` is outside the pool and still caught by its own branch, and the old Teredo check is subsumed rather than lost. No duplicated IPv6 classifier exists elsewhere in the workspace, so the single chokepoint holds. Boundary tests cover both edges of the range.
- **Welcome group-id reuse (#1041):** `do_confirm_published` always emits exactly one `GroupCreated`/`EpochChanged` and the rollback path discards `welcome_failures` entirely, so the confirm-derived id is present exactly when failures are surfaced. The existing `marmot-account` runtime tests already assert `Some(group_id)` on both the create and evolution paths, so the contract is pinned. `stored_sent_welcome` is still live via `redeliver_welcome`.
- **Release gating (#996):** no `debug-assertions` profile overrides exist in the workspace, so `cfg!(debug_assertions)` is a faithful debug/release discriminator here.

The inline comments are all minor: one residual docs/code divergence in the spec's storage section, a robustness note on `confirmed_group_id`'s first-match heuristic, an operator-visibility suggestion for the silently-ignored release override, a contract-documentation nit on the new generic write helpers, and an optional test-coverage note on the empty-id dedup guard. Nothing blocking.

PR #1064, opened 2026-07-23
TUI hardening: cell-exact media, guarded interactions, logout, and review follow-ups

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Added `/logout` with typed confirmation safeguards (local-signing vs public-only).
  * User search “add” now uses a group picker to add to an existing chat or start a new one.
  * Added `/reply <text>` and strengthened multi-step popup flows.
* **Bug Fixes**
  * Hardened `/react` validation to a single non-ASCII emoji grapheme (plus sentinels).
  * Preserved armed reactions while typing; `Ctrl-U` now reliably clears drafts/input.
  * Improved cell-exact inline media rendering and now removes decrypted artifacts promptly.
  * Refined unread badge rendering and terminal-safe hint/popup text.
* **Documentation**
  * Updated TUI README and command/behavior descriptions to match the new flows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7001,6 +7001,7 @@ dependencies = [
  "transport-quic-broker",
  "transport-quic-stream",
  "unicode-properties",
+ "unicode-segmentation",
  "url",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,10 @@ image = { version = "0.25", default-features = false, features = ["png", "jpeg",
 # General-category lookups for the terminal untrusted-text sanitizer (drop the
 # whole Cf format class instead of a drifting hardcoded denylist).
 unicode-properties = { version = "0.1", default-features = false, features = ["general-category"] }
+# Extended grapheme-cluster segmentation for the TUI /react guard (one cluster ==
+# one emoji). Already in the tree via ratatui; declared here so wn-cli can depend
+# on it directly at the workspace-consistent version.
+unicode-segmentation = "1"
 unicode-normalization = "0.1"
 rand = "0.8"
 sha2 = "0.10"
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -84,12 +84,15 @@ versioning through the workspace version in the root `Cargo.toml`.
 
 - TUI: inbound media renders inline in the message pane. Image attachments are downloaded and decoded off the event
   loop (a worker thread runs `wn media download <group> <plaintext-hash> --output <cache path>` and the `image`
-  decode, delivering the result over an `mpsc` channel drained on tick) and drawn inline via `ratatui-image` on
-  terminals with a graphics protocol (Kitty/iTerm2/Sixel; iTerm2 forced via `ITERM_SESSION_ID`). Placeholders walk
-  `[img name]` -> `[downloading name...]` -> `[loading name...]` -> the image, or `[name failed: err]`, and stay
-  `[img name]` where no image protocol exists. `o` opens the selected message's image full-size in a dismiss-on-any-key
-  viewer. Decrypted files cache under the TUI home in `tui-media-cache/`. No JSON response shapes changed (the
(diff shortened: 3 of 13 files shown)


erskingardner (approved):
## Overall assessment

This is a strong PR. The test coverage is exceptional (~1,600 lines of tests for ~650 lines of implementation), covering the double-Enter footgun, token mismatch, hostile display names, symlinked cache roots, measure-vs-draw row accounting, and the picker→confirm chain. The reasoning-dense comments follow house style, the CHANGELOG/README updates are thorough and honest about the destructive semantics, and the security instincts are right: the typed-token gate matching the consequence, `symlink_metadata` before the sweep with the TOCTOU explicitly acknowledged, and `terminal_safe_text` on every user-controlled string entering a popup. Nothing here touches engine/transport crates, so repo invariants are respected.

Approving, with the findings below — **finding 1 is the one I'd want addressed before merge**; the rest are nits or follow-up material.

## Findings

**1. Ctrl-U in Messages focus fires an immediate unreact (moderate, UX/correctness).** The existing accelerator arm `KeyCode::Char('u') if self.focus == Focus::Messages` in `app.rs` matches on the keycode without checking modifiers, and it sits *before* the new Ctrl-U arm (which is guarded to `Focus::Composer`). So a user who has just been taught "Ctrl-U clears" by the new composer hint, but happens to be in Messages focus, will fire `messages unreact` — an unconfirmed network action — instead of a no-op. This modifier-insensitivity is pre-existing (Ctrl-R would arm react, etc.), but the PR newly builds a Ctrl-U habit, so it amplifies the trap. Suggest gating the single-letter accelerators on `key.modifiers.is_empty()`, or matching Ctrl-U before the focus-dispatched accelerators.

**2. Logout success followed by refresh failure reports as error (minor).** In `logout_account`, if `wn logout` succeeds but `refresh_or_return_to_login()` fails, the `?` propagates: the status line shows `error: ...` while the wipe actually happened, and the TUI keeps displaying the now-removed account with its stale subscription — the exact state the comment says the refresh exists to prevent. Given the repo's multi-step-state-change invariant ("report outcomes faithfully"), consider catching the refresh error and still reporting the logout as done (e.g., "logged out …; account list reload failed, /refresh to retry").

**3. Startup sweep races a concurrent session (minor).** Two TUI instances sharing a `--home`: instance B's startup sweep can delete a file instance A's worker just downloaded but hasn't decoded yet. Failure is graceful (A shows `[failed: …]`), and concurrent instances are an edge case, but a code comment acknowledging it would be in keeping with the file's style.

**4. `.claude` fully excluded from the legacy-naming lint (nit).** `check_legacy_naming.sh` now skips everything under `.claude`, presumably to ignore review worktrees under `.claude/worktrees/`. That also exempts checked-in `.claude` config from the naming check. `--glob '!.claude/worktrees'` would be the tighter scope.

**5. Trim inconsistency in armed detection (nit).** `armed_interaction()` matches the raw composer text while `parse_slash_command` trims, so `" /react x"` (leading space) submits as a reaction but never shows the armed hint and isn't Esc-clearable. Unreachable via the prefill path, only by hand-typing a leading space — a shared trim would close it.

**6. Halfblocks-everywhere fidelity trade-off (design note, no change requested).** Forcing halfblocks even on terminals with well-behaved Kitty support downgrades fidelity for everyone to fix the misbehaving cases. The rationale in `adopt_picker` is thorough and the containment argument is convincing for a scrolling list; if users complain, a config opt-in for pixel protocols is the obvious future escape valve. The structural test that pins the chokepoint (forcing an iTerm2 picker and asserting no OSC 1337 reaches a cell) is a particularly nice touch.

**7. Reaction guard semantics look right.** The single-grapheme + non-ASCII + scalar-cap rule correctly accepts ZWJ families, skin tones, flags, keycaps, and the NIP-25 `+`/`-` sentinels while rejecting `café`/`你好吗`/`привет`/emoji walls; the single-CJK-character ceiling is documented and tested as intentional. The TUI-only placement (CLI stays protocol-faithful) is the right layering.

PR #1066, opened 2026-07-23
Adopt application-data wire carriers

### LINKED ISSUE #1047: Implement OpenMLS application-data components and correct last-resort KeyPackage encoding
## Summary

Bring MDK's OpenMLS dependency/integration onto the adopted Marmot application-data wire model before implementing the current application profile.

The current OpenMLS 0.8.1 integration can consume much of the newer application-data shape, but `mark_as_last_resort()` still emits the obsolete extension type `0x000a`. The adopted profile requires KeyPackage-level application data carrying component `0x0004` instead. This is a wire prerequisite for every current-profile KeyPackage.

## Normative source

- `foundation/mls-protocol.md`
- `foundation/key-packages.md`
- `foundation/registries.md`
- `app-components/README.md`

## Start here

- `crates/cgka-engine/src/key_package.rs`
- OpenMLS dependency declarations and patches in the workspace manifests/lockfile
- `crates/cgka-engine/tests/group_creation.rs`
- KeyPackage conformance/vector coverage

## Required behavior

- Use the adopted MLS `application_data` carriers for KeyPackage, GroupContext, LeafNode, and proposal application data.
- Encode last-resort status as KeyPackage application-data component `0x0004`; do not emit the obsolete last-resort extension.
- Preserve unknown application-data components where the relevant state transition requires round-tripping them.
- Reject duplicate component types and malformed length framing at the protocol boundary.
- Keep legacy decoding needed to operate already-joined legacy groups separate from the current-profile encoder.

## Out of scope

- Flipping publication defaults or deleting legacy KeyPackages; that belongs to the strict-cutover issue.
- Account identity proof semantics beyond providing the correct carrier.
- Multi-device authorization (`0x800a`).

## Acceptance criteria

- Current-profile KeyPackages contain application data and component `0x0004` when last-resort; they do not contain extension `0x000a` for that purpose.
- Round-trip and malformed/duplicate-component tests cover every adopted application-data carrier used by MDK.
- A wire-level fixture or conformance assertion detects regression to the old extension.
- Existing legacy group fixtures remain loadable.
- `just fast-ci` and targeted engine/conformance tests pass.

erskingardner:
## Summary

- pin the OpenMLS fork follow-up that rejects trailing bytes inside `app_data_dictionary` extension payloads
- stop advertising the obsolete last-resort extension on new LeafNodes while retaining legacy decoding
- keep current last-resort encoding pinned to KeyPackage application-data component `0x0004`
- add direct carrier-codec tripwires plus engine-boundary coverage for malformed KeyPackages and opaque-component preservation across update, Add, and Welcome

The draft-10 last-resort encoder itself arrived in #1063. This PR advances the OpenMLS pin to commit [`59e7d3b27a7e95237879dd5478de1fd90eff7ada`](https://github.com/erskingardner/openmls/commit/59e7d3b27a7e95237879dd5478de1fd90eff7ada) for the trailing-bytes framing fix, stops advertising the old extension, and pins the MDK integration contract around those behaviors.

Legacy extension decoding remains covered here. Persisted legacy-profile and legacy-group loadability are implemented and exercised in the immediately stacked #1067.

## Validation

- `cargo test -p cgka-engine --locked`
- `just fast-ci`
- OpenMLS targeted unit tests and clippy for the framing fix

Fixes #1047

## Summary by CodeRabbit

- **New Features**
  - Enhanced handling of application-data components across KeyPackage, leaf, and group workflows.
  - Opaque application-data components are preserved during unrelated group updates, including after adding new members.

- **Bug Fixes**
  - Improved validation rejects malformed, duplicate, unordered, or truncated application-data encodings.
  - Updated last-resort behavior: current capability advertisement avoids the legacy last-resort extension while still supporting legacy decoding.

- **Tests**
  - Added wire-contract and group/update integration coverage for application-data carriers, last-resort variants, and update/remove proposals.

- **Documentation**
  - Refreshed architecture notes clarifying current extension advertisement and legacy handling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5550,7 +5550,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.3.4",
+ "getrandom 0.4.2",
  "once_cell",
  "rustix 1.1.4",
  "windows-sys 0.61.2",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3648,7 +3648,7 @@ checksum = "c08d65885ee38876c4f86fa503fb49d7b507c2b62552df7c70b2fce627e06381"
 [[package]]
 name = "openmls"
 version = "0.8.1"
-source = "git+https://github.com/erskingardner/openmls.git?rev=85990fd44634bd054f010f14dc4d8987c9adfb3d#85990fd44634bd054f010f14dc4d8987c9adfb3d"
+source = "git+https://github.com/erskingardner/openmls.git?rev=59e7d3b27a7e95237879dd5478de1fd90eff7ada#59e7d3b27a7e95237879dd5478de1fd90eff7ada"
 dependencies = [
  "log",
  "openmls_serialization_helpers",
@@ -3664,7 +3664,7 @@ dependencies = [
 [[package]]
 name = "openmls_basic_credential"
 version = "0.5.0"
-source = "git+https://github.com/erskingardner/openmls.git?rev=85990fd44634bd054f010f14dc4d8987c9adfb3d#85990fd44634bd054f010f14dc4d8987c9adfb3d"
+source = "git+https://github.com/erskingardner/openmls.git?rev=59e7d3b27a7e95237879dd5478de1fd90eff7ada#59e7d3b27a7e95237879dd5478de1fd90eff7ada"
 dependencies = [
  "ed25519-dalek",
  "ml-dsa",
@@ -3680,7 +3680,7 @@ dependencies = [
 [[package]]
 name = "openmls_memory_storage"
 version = "0.5.0"
-source = "git+https://github.com/erskingardner/openmls.git?rev=85990fd44634bd054f010f14dc4d8987c9adfb3d#85990fd44634bd054f010f14dc4d8987c9adfb3d"
+source = "git+https://github.com/erskingardner/openmls.git?rev=59e7d3b27a7e95237879dd5478de1fd90eff7ada#59e7d3b27a7e95237879dd5478de1fd90eff7ada"
 dependencies = [
  "log",
  "openmls_traits",
@@ -3692,7 +3692,7 @@ dependencies = [
 [[package]]
 name = "openmls_rust_crypto"
 version = "0.5.1"
-source = "git+https://github.com/erskingardner/openmls.git?rev=85990fd44634bd054f010f14dc4d8987c9adfb3d#85990fd44634bd054f010f14dc4d8987c9adfb3d"
+source = "git+https://github.com/erskingardner/openmls.git?rev=59e7d3b27a7e95237879dd5478de1fd90eff7ada#59e7d3b27a7e95237879dd5478de1fd90eff7ada"
 dependencies = [
  "aes-gcm",
  "chacha20poly1305",
@@ -3718,7 +3718,7 @@ dependencies = [
 [[package]]
 name = "openmls_serialization_helpers"
 version = "0.1.0"
-source = "git+https://github.com/erskingardner/openmls.git?rev=85990fd44634bd054f010f14dc4d8987c9adfb3d#85990fd44634bd054f010f14dc4d8987c9adfb3d"
+source = "git+https://github.com/erskingardner/openmls.git?rev=59e7d3b27a7e95237879dd5478de1fd90eff7ada#59e7d3b27a7e95237879dd5478de1fd90eff7ada"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -3729,7 +3729,7 @@ dependencies = [
 [[package]]
 name = "openmls_traits"
 version = "0.5.0"
-source = "git+https://github.com/erskingardner/openmls.git?rev=85990fd44634bd054f010f14dc4d8987c9adfb3d#85990fd44634bd054f010f14dc4d8987c9adfb3d"
+source = "git+https://github.com/erskingardner/openmls.git?rev=59e7d3b27a7e95237879dd5478de1fd90eff7ada#59e7d3b27a7e95237879dd5478de1fd90eff7ada"
 dependencies = [
  "serde",
  "tls_codec",
@@ -5550,7 +5550,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.4.2",
+ "getrandom 0.3.4",
  "once_cell",
  "rustix 1.1.4",
  "windows-sys 0.61.2",
--- a/crates/cgka-engine/src/capabilities.rs
+++ b/crates/cgka-engine/src/capabilities.rs
@@ -25,7 +25,6 @@ pub(crate) fn leaf_capabilities(
     let mut ext_types: Vec<ExtensionType> = vec![
         ExtensionType::RequiredCapabilities,
         ExtensionType::AppDataDictionary,
-        ExtensionType::LastResort,
         crate::account_identity_proof::account_identity_proof_capability(),
     ];
     let mut proposal_types: Vec<ProposalType> = vec![ProposalType::AppDataUpdate];
--- a/crates/cgka-engine/tests/application_data_wire.rs
+++ b/crates/cgka-engine/tests/application_data_wire.rs
@@ -1,0 +1,145 @@
+//! Wire-contract tests for the MLS application-data carriers Marmot uses.
+//!
+//! These tests deliberately sit at the MDK integration boundary. They pin the
+//! OpenMLS fork behavior that current-profile KeyPackages, LeafNodes,
+//! GroupContexts, and AppDataUpdate proposals rely on.
+
+use openmls::component::ComponentData;
+use openmls::extensions::{
+    AppDataDictionary, AppDataDictionaryExtension, Extension, ExtensionType, Extensions,
+};
+use openmls::group::GroupContext;
+use openmls::messages::proposals::{AppDataUpdateOperation, AppDataUpdateProposal, ProposalType};
+use openmls::prelude::{KeyPackage, LeafNode};
+use tls_codec::{Deserialize as _, Serialize as _, VLBytes};
+
+const OPAQUE_COMPONENT_ID: u16 = 0x9001;
+const OPAQUE_COMPONENT_BYTES: &[u8] = &[0xde, 0xad, 0xbe, 0xef];
+
+fn app_data_extension() -> Extension {
+    let mut dictionary = AppDataDictionary::new();
+    dictionary.insert(OPAQUE_COMPONENT_ID, OPAQUE_COMPONENT_BYTES.to_vec());
+    Extension::AppDataDictionary(AppDataDictionaryExtension::new(dictionary))
+}
+
+fn assert_opaque_component<T>(extensions: &Extensions<T>) {
+    let dictionary = extensions
+        .app_data_dictionary()
+        .expect("app_data_dictionary extension")
+        .dictionary();
+    assert_eq!(
+        dictionary.get(&OPAQUE_COMPONENT_ID),
+        Some(OPAQUE_COMPONENT_BYTES)
+    );
+}
+
+#[test]
+fn application_data_dictionary_roundtrips_in_every_extension_carrier_mdk_uses() {
+    let key_package = Extensions::<KeyPackage>::single(app_data_extension()).unwrap();
+    let decoded = Extensions::<KeyPackage>::tls_deserialize_exact(
+        key_package.tls_serialize_detached().unwrap(),
+    )
+    .unwrap();
+    assert_opaque_component(&decoded);
+
+    let leaf_node = Extensions::<LeafNode>::single(app_data_extension()).unwrap();
+    let decoded =
+        Extensions::<LeafNode>::tls_deserialize_exact(leaf_node.tls_serialize_detached().unwrap())
+            .unwrap();
+    assert_opaque_component(&decoded);
+
+    let group_context = Extensions::<GroupContext>::single(app_data_extension()).unwrap();
+    let decoded = Extensions::<GroupContext>::tls_deserialize_exact(
+        group_context.tls_serialize_detached().unwrap(),
+    )
+    .unwrap();
+    assert_opaque_component(&decoded);
+}
+
+#[test]
+fn application_data_dictionary_roundtrip_is_canonical_and_preserves_opaque_bytes() {
+    let entries = vec![
+        ComponentData::from_parts(0x0001, vec![0x01].into()),
+        ComponentData::from_parts(OPAQUE_COMPONENT_ID, OPAQUE_COMPONENT_BYTES.to_vec().into()),
+    ];
+    let encoded = entries.tls_serialize_detached().unwrap();
+    let dictionary = AppDataDictionary::tls_deserialize_exact(encoded.clone()).unwrap();
+
+    assert_eq!(
+        dictionary.get(&OPAQUE_COMPONENT_ID),
+        Some(OPAQUE_COMPONENT_BYTES)
+    );
+    assert_eq!(dictionary.tls_serialize_detached().unwrap(), encoded);
+}
+
+#[test]
+fn application_data_dictionary_rejects_duplicate_and_out_of_order_components() {
+    let duplicate = vec![
+        ComponentData::from_parts(0x0001, vec![0x01].into()),
+        ComponentData::from_parts(0x0001, vec![0x02].into()),
+    ]
+    .tls_serialize_detached()
+    .unwrap();
+    assert!(AppDataDictionary::tls_deserialize_exact(duplicate).is_err());
+
+    let out_of_order = vec![
+        ComponentData::from_parts(OPAQUE_COMPONENT_ID, vec![0x01].into()),
+        ComponentData::from_parts(0x0001, vec![0x02].into()),
+    ]
+    .tls_serialize_detached()
+    .unwrap();
+    assert!(AppDataDictionary::tls_deserialize_exact(out_of_order).is_err());
+}
+
+#[test]
+fn application_data_extension_rejects_malformed_length_framing() {
+    let Extension::AppDataDictionary(extension) = app_data_extension() else {
+        unreachable!()
+    };
+    let mut extension_data = extension.tls_serialize_detached().unwrap();
+    extension_data.push(0xff);
+
+    let mut malformed = ExtensionType::AppDataDictionary
+        .tls_serialize_detached()
+        .unwrap();
+    malformed.extend_from_slice(
+        &VLBytes::from(extension_data)
+            .tls_serialize_detached()
+            .unwrap(),
+    );
+
+    assert!(Extension::tls_deserialize_exact(malformed).is_err());
+
+    let mut truncated = app_data_extension().tls_serialize_detached().unwrap();
+    truncated.pop();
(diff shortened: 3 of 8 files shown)


erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

Cargo.lock
```diff
@@ -5550,7 +5550,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.4.2",
+ "getrandom 0.3.4",
```
erskingardner: **Should fix before merge — unrelated lockfile churn.**

`tempfile` is still `3.27.0`, but its `getrandom` resolution quietly moved from `0.4.2` (master) to `0.3.4`. The only intentional lock change in this PR should be the OpenMLS rev bump.

Please restore `getrandom 0.4.2` for tempfile (or document why this downgrade is required by the OpenMLS pin). Shipping unrelated resolver churn in a wire-format PR makes regressions harder to bisect.

crates/cgka-engine/tests/application_data_wire.rs
```diff
@@ -0,0 +1,145 @@
+//! Wire-contract tests for the MLS application-data carriers Marmot uses.
+//!
+//! These tests deliberately sit at the MDK integration boundary. They pin the
+//! OpenMLS fork behavior that current-profile KeyPackages, LeafNodes,
+//! GroupContexts, and AppDataUpdate proposals rely on.
+
+use openmls::component::ComponentData;
+use openmls::extensions::{
```
erskingardner, line 10: **Overclaim vs #1047.**

These tests sit on OpenMLS codecs (`Extensions::tls_deserialize_exact`, `AppDataDictionary`, `AppDataUpdateProposal`), not on MDK’s engine parse/ingest path. That is a valuable fork pin, but it is not the MDK protocol boundary.

Please either:
1. drive at least the malformed/trailing-byte cases through `parse_key_package` / create-invite / commit ingest, or
2. rewrite the module docs to say these are OpenMLS-integration pins, not MDK-boundary coverage.

As written, a future MDK path that bypasses these codecs could regress without failing this file.

crates/cgka-engine/tests/application_data_wire.rs
```diff
@@ -0,0 +1,145 @@
...
+        ComponentData::from_parts(OPAQUE_COMPONENT_ID, OPAQUE_COMPONENT_BYTES.to_vec().into()),
+    ];
+    let encoded = entries.tls_serialize_detached().unwrap();
+    let dictionary = AppDataDictionary::tls_deserialize_exact(encoded.clone()).unwrap();
+
+    assert_eq!(
+        dictionary.get(&OPAQUE_COMPONENT_ID),
+        Some(OPAQUE_COMPONENT_BYTES)
+    );
+    assert_eq!(dictionary.tls_serialize_detached().unwrap(), encoded);
```
erskingardner: This does not prove canonicity of the construction path.

The input is already sorted (`0x0001` then `0x9001`). Matching re-serialize after a successful deserialize only shows identity for a valid encoding.

Stronger options:
- build via `AppDataDictionary::insert` in unsorted insertion order, then assert wire bytes are sorted, and/or
- rename so we do not claim “canonical” beyond round-trip of pre-sorted bytes.

crates/cgka-engine/tests/application_data_wire.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    let remove = AppDataUpdateProposal::remove(OPAQUE_COMPONENT_ID);
+    let decoded =
+        AppDataUpdateProposal::tls_deserialize_exact(remove.tls_serialize_detached().unwrap())
+            .unwrap();
+    assert_eq!(decoded.component_id(), OPAQUE_COMPONENT_ID);
+    assert_eq!(decoded.operation(), &AppDataUpdateOperation::Remove);
+
+    let mut unknown_operation = update_bytes.clone();
+    unknown_operation[2] = 0xff;
```
erskingardner: Fragile layout poke: `unknown_operation[2] = 0xff` assumes `ComponentId` is a 2-byte prefix and the operation discriminant is the next byte.

That is true today (`ComponentId = u16`, then `AppDataUpdateOperation`), but a length-prefix / reorder will make this test green while testing the wrong field.

Prefer an explicitly framed buffer (`component_id.to_be_bytes()` + `0xff` + …) or a tiny helper that documents the layout.

crates/cgka-engine/tests/group_creation.rs
```diff
@@ -832,6 +933,26 @@ async fn fresh_key_package_uses_draft10_last_resort_component() {
...
 
+#[test]
+fn malformed_last_resort_component_is_rejected_at_mdk_boundary() {
+    let key_package = key_package_with_malformed_last_resort_component(b"malformed-last-resort");
+    let error = is_last_resort_key_package(&key_package)
+        .expect_err("non-empty last-resort component data must be rejected");
+    assert!(
+        matches!(&error, EngineError::Backend(message) if message.contains("MalformedLastResortComponent")),
+        "unexpected error: {error:?}"
+    );
```
erskingardner, line 954: Good fail-closed check on `is_last_resort_key_package`, but attackers hit `parse_key_package` via `create_group` / `invite`, not this helper.

Please add a create or invite call with `key_package_with_malformed_last_resort_component(...)` and assert it returns the `MalformedLastResortComponent` backend error. Same validator today — different seam tomorrow.

crates/cgka-engine/tests/group_creation.rs
```diff
@@ -832,6 +933,26 @@ async fn fresh_key_package_uses_draft10_last_resort_component() {
...
+}
+
+#[test]
+fn legacy_last_resort_extension_remains_decodable() {
+    let key_package = legacy_last_resort_key_package(b"legacy-last-resort");
+    assert!(
+        is_last_resort_key_package(&key_package)
+            .expect("legacy KeyPackage remains valid for compatibility")
+    );
```
erskingardner, line 963: This only proves the helper still recognizes the obsolete extension.

The fixture also omits `AppDataDictionary` capability, so it is not a joinable current-profile KeyPackage. That is fine for a decode pin, but #1047’s “existing legacy group fixtures remain loadable” is a different AC and is not covered here.

Please either point at the test/fixture that already covers legacy-group load, or add one; otherwise call that AC out as deferred.

crates/cgka-engine/tests/update_group_data.rs
```diff
@@ -985,6 +998,70 @@ async fn admin_policy_update_listing_non_member_is_rejected() {
...
+    converge_buffered_commit(&mut bob, &gid);
+
+    assert_eq!(
+        alice.app_component(&gid, OPAQUE_COMPONENT_ID).unwrap(),
+        Some(OPAQUE_BYTES.to_vec())
+    );
+    assert_eq!(
+        bob.app_component(&gid, OPAQUE_COMPONENT_ID).unwrap(),
+        Some(OPAQUE_BYTES.to_vec())
+    );
```
erskingardner, line 1062: Nice multi-member preservation check, but it only covers `UpdateGroupData` (profile rename).

#1047 asks to preserve unknown components wherever the relevant transition must round-trip them. The failure modes for Add/invite (and mixed AppDataUpdate commits) are different from a single-component profile update.

Please add at least an invite/Add path that leaves the opaque component intact on both members; admin/remove would be even better.

crates/cgka-engine/src/capabilities.rs
```diff
@@ -25,7 +25,6 @@ pub(crate) fn leaf_capabilities(
     let mut ext_types: Vec<ExtensionType> = vec![
         ExtensionType::RequiredCapabilities,
         ExtensionType::AppDataDictionary,
```
erskingardner, line 27: Agreed with removing `ExtensionType::LastResort` from leaf ads now that last-resort is component `0x0004` inside `app_data_dictionary`.

Worth stating in the PR body: MDK never put `LastResort` into `RequiredCapabilities`, so this should not strand current groups. The fresh-KP assertions in `group_creation.rs` are the right regression guards for the advertise/encode split.

erskingardner (commented):
## Adversarial review

Fork pin itself is fine / out of scope per prior agreement.

This PR is narrower than the title/summary imply: draft-10 last-resort encoding already came with the prior OpenMLS pin; the MDK behavioral change here is essentially stop advertising `LastResort`, pin the trailing-bytes framing fix, and add tests. Direction is right, but a few coverage claims overshoot what is proven, and `Cargo.lock` carries unrelated `getrandom` churn that should not ship.

### Should fix before merge
1. **Unrelated lockfile downgrade** — `tempfile` 3.27.0 flips from `getrandom 0.4.2` → `0.3.4` with no explanation. Please restore unless required by the OpenMLS rev.
2. **“MDK boundary” overclaim** — `application_data_wire.rs` is almost entirely direct OpenMLS codec tests. Useful fork pin, but #1047’s protocol-boundary rejection should also be exercised through engine parse/ingest (or the claim should be toned down).
3. **Malformed last-resort only tested via helper** — please also hit `create_group`/`invite`/`parse_key_package` with the malformed KP.

### Non-blocking but important
- “Canonical” round-trip only proves sorted-input identity.
- Legacy test proves helper decode, not legacy-group loadability.
- Opaque preservation covers only `UpdateGroupData`, not Add/invite (or other transitions).
- PR description should say the encoder was already in the previous pin.

Happy to re-review quickly once the lockfile noise is gone and the boundary tests are tightened.

> GitHub will not let me formally `REQUEST_CHANGES` on my own PR, so this is submitted as a comment review with the same severity. Please treat the “should fix” items as blocking.

> Ignore any earlier empty “line probe” / range-check review shells from placing inline anchors.

erskingardner (commented):

Follow-up to the earlier review (#pullrequestreview-4766044199): I re-derived the diff from scratch, fetched the PR head into a disposable worktree, and actually built and ran it — `cargo test -p cgka-engine` (all tests pass, including every new one), `cargo check --workspace`, and the full `just fast-ci` recipe (fmt-check, naming gate, check + clippy across default and OTLP features) are all clean. Most of the earlier review's claims held up under verification; where my severity differs after building/tracing, I say so below. Fork pinning itself is intentional and out of scope.

### What this PR actually changes

The fork pin advances `85990fd4` → `59e7d3b2`. I diffed the fork directly: the delta touches only `openmls/src/extensions/{app_data_dict_extension.rs,codec.rs}` and rejects trailing bytes in `app_data_dictionary` extension payloads. The draft-10 last-resort *encoder* (`mark_as_last_resort()` → component `0x0004`) already shipped with the previous pin in #1063, so this PR's own behavioral change is narrower than the title suggests: stop advertising `ExtensionType::LastResort` in leaf capabilities (`crates/cgka-engine/src/capabilities.rs`), pin the framing fix, and add tests. Worth reflecting in the PR description.

I traced both call sites of `leaf_capabilities()` — `key_package.rs:91` (fresh KeyPackages) and `group_lifecycle.rs:207` (new group creator's own leaf). Neither needs to advertise `LastResort`: `required_capabilities_extension` never required it, so no group's RequiredCapabilities can regress, and a transported legacy KeyPackage's `LastResort` extension is validated against *that KeyPackage's own* advertised capabilities, not the local engine's. The removal is correct.

### Non-blocking findings

1. **Malformed last-resort component isn't exercised on the actually-adversarial path** (`crates/cgka-engine/tests/group_creation.rs:408`). The test drives `is_last_resort_key_package`, which is only ever called on the local client's own cached KeyPackage (`marmot-app/src/client/mod.rs:206`, with `.unwrap_or(false)` swallowing errors). The seam an attacker actually hits is `parse_key_package` via `create_group`/`invite` on a network-sourced invitee KeyPackage. Both funnel through the same `validate_key_package` helper, so rejection does hold today — I verified this — but a direct assertion on the invite path would stop a future refactor from silently splitting the two paths.

2. **Legacy-decodable ≠ legacy-operable** (`crates/cgka-engine/tests/group_creation.rs:419`). `legacy_last_resort_extension_remains_decodable` proves the old extension still decodes and is recognized. #1047's acceptance criterion "existing legacy group fixtures remain loadable" is a broader operational claim (already-joined legacy group *state* stays hydratable) that this test doesn't evidence — the fixture doesn't even advertise `AppDataDictionary`, so it couldn't join a current-profile group. Not a regression risk (no group-loading code changed), but don't check that criterion off against this test alone.

3. **Unrelated `getrandom` lockfile churn** (`Cargo.lock`): `tempfile 3.27.0` flips `getrandom 0.4.2` → `0.3.4` with no corresponding manifest change — looks like whole-workspace `cargo update` drift rather than a scoped update of the five `openmls*` packages. The earlier review called this "should fix before merge"; having checked both versions, I'd downgrade that: both are maintained majors and `tempfile` isn't security-sensitive here, so it's functionally inert — just avoidable noise. A scoped `cargo update -p openmls -p openmls_basic_credential -p openmls_memory_storage -p openmls_rust_crypto -p openmls_traits` keeps the diff to exactly what the PR claims.

4. **Stale doc drifts further** (`docs/marmot-architecture/further-context/custom_extensions.md:536,545` — not in this diff). It still says clients "MUST advertise … `last_resort` — standard MLS extension draft" and calls the inherited extension "**Correct**". Already stale since #1063; this PR makes it wrong in a second way (capability advertisement, not just wire encoding). Worth a follow-up cleanup.

### What's done well

- One behavioral line removed, everything else tests/lockfile — low blast radius, easy to verify.
- `application_data_wire.rs` pins the shape of the fork behavior MDK depends on (canonical round-trip, duplicate/out-of-order rejection, malformed framing, proposal round-trip) directly against the types MDK consumes — a good tripwire for future fork or upstream drift.
- `unrelated_update_preserves_opaque_app_component_for_all_members` is a genuinely useful end-to-end regression test: two real engines, real welcome + commit + ingest, proving an uninterpreted app component survives an unrelated `UpdateGroupData` round trip on both sides.
- All five OpenMLS git dependencies bumped in lockstep, honoring the same-commit invariant documented in `Cargo.toml`.

No blocking issues from this pass. 🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

agent-p1p:
## Adversarial re-review — prior block resolved

Re-reviewed fix commit `3574a12` and the final diff at exact head `3574a12f6c3a599ff9c644415b4e2c8bce91f5be`. This explicitly clears the blocking findings in [the original review](https://github.com/marmot-protocol/mdk/pull/1066#pullrequestreview-4766044199).

### Blocking findings resolved

1. **Unrelated lockfile churn:** resolved. `Cargo.lock` restores `tempfile` to `getrandom 0.4.2`; the remaining lock diff is limited to the six OpenMLS workspace packages moving to `59e7d3b2`.
2. **Codec tests presented as the MDK boundary:** resolved. `application_data_wire.rs:1-7` now accurately identifies the direct tests as OpenMLS integration pins and points to the engine seam separately.
3. **Malformed last-resort marker never reached the adversarial engine path:** resolved. `group_creation.rs:936-954` now drives the malformed KeyPackage through `create_group`, which reaches `Engine::parse_key_package` at `group_lifecycle.rs:154`, and asserts the `MalformedLastResortComponent` validation failure.

The prior non-blocking gaps were tightened as well: descending insertion now proves canonical component ordering (`application_data_wire.rs:62-75`), malformed proposal construction is explicit (`application_data_wire.rs:147-152`), opaque bytes survive `UpdateGroupData`, Add, and Welcome for all members (`update_group_data.rs:1002-1105`), and the stale current/legacy documentation is corrected. Persisted legacy-profile/group fixture coverage remains explicitly separated in stacked #1067 rather than being overclaimed here.

### Independent verification

- `cargo test -p cgka-engine --locked --test application_data_wire` — 5 passed
- Three focused `group_creation` last-resort tests — passed
- `cargo test -p cgka-engine --locked --test update_group_data unrelated_update_and_invite_preserve_opaque_app_component_for_all_members` — passed
- `cargo test -p cgka-engine --locked` — passed
- `just fast-ci` — passed
- Exact-head GitHub Actions and CodeRabbit checks — green

**Verdict: no blocking findings, suggestions, or nitpicks.** Sensitive MLS/CGKA paths still require the separate merge gate and human approval.

Cargo.lock
agent-p1p: Addressed in 3574a12: Cargo.lock restores tempfile to getrandom 0.4.2. The remaining lockfile delta is limited to the six OpenMLS workspace packages moving to 59e7d3b2.

crates/cgka-engine/tests/application_data_wire.rs
agent-p1p: Addressed in 3574a12: the module docs now describe these as direct OpenMLS integration-contract pins and point to the engine seams. group_creation::create_group_rejects_malformed_last_resort_component now drives malformed KeyPackage application data through create_group -> parse_key_package.

crates/cgka-engine/tests/application_data_wire.rs
agent-p1p: Addressed in 3574a12: the test inserts components in descending id order and asserts the exact canonically sorted wire encoding.

crates/cgka-engine/tests/application_data_wire.rs
agent-p1p: Addressed in 3574a12: the malformed proposal frame is now constructed explicitly from the u16 component id followed by the invalid operation discriminant.

crates/cgka-engine/tests/group_creation.rs
agent-p1p: Addressed in 3574a12: create_group_rejects_malformed_last_resort_component passes the malformed KeyPackage through create_group and asserts MalformedLastResortComponent from the engine boundary.

crates/cgka-engine/tests/group_creation.rs
agent-p1p: The PR body now states the exact scope: legacy KeyPackage extension decoding is pinned here; persisted legacy-profile/group reopening is implemented and exercised in stacked PR #1067. The exact-head adversarial re-review accepted that separation.

crates/cgka-engine/tests/update_group_data.rs
agent-p1p: Addressed in 3574a12: unrelated_update_and_invite_preserve_opaque_app_component_for_all_members now executes the Add/Welcome path for Carol and asserts byte-for-byte preservation for Alice, Bob, and Carol.

crates/cgka-engine/src/capabilities.rs
agent-p1p: The PR body now calls out that the draft-10 encoder arrived in #1063 and this PR stops advertising the obsolete extension. Fresh-KeyPackage assertions cover both the current component and the absence of the legacy capability; required_capabilities_extension continues to require only the current application-data carrier.

Issue #1048, opened 2026-07-23
Classify legacy/current profiles and implement account identity proof component 0x8009


erskingardner:
## Summary

Add one explicit, persisted Marmot profile classification and implement the adopted current-profile account identity proof. Every protocol decision must dispatch from that classification; MDK must not infer a hybrid profile component by component.

Existing groups are legacy and must remain operable. Newly joined/created groups after cutover are current. No in-place profile migration is planned.

## Normative source

Adopted by marmot-protocol/marmot#408 at `625f8aae8096a17c6b11fde08ec8c2d582568619`:

- `foundation/account-identity-proof-v1.md`
- `app-components/account-identity-proof-v2.md`
- `foundation/identity.md`
- `foundation/registries.md`
- `protocol-core/group-state.md`

## Start here

- `crates/cgka-engine/src/account_identity_proof.rs`
- `crates/cgka-engine/src/identity.rs`
- `crates/cgka-engine/src/openmls_projection.rs`
- `crates/traits/src/app_components/`
- storage migrations/group metadata in `crates/storage-sqlite/`

## Required behavior

- Represent at least `Legacy` and `Current` as an explicit group/KeyPackage profile, persisted durably.
- Legacy means the already-deployed `0xf2f1` proof/Media V1 behavior. Current means component `0x8009` plus the adopted current application profile.
- Encode/decode the current account proof as the exact 104-byte `0x8009` component.
- Validate the kind-450 Nostr event proof, including exact tag/content structure and the specified timestamp rules.
- Do not reuse the existing halfway `v2` extension carrier or zero-timestamp construction as the current profile.
- Dispatch media, proof, capability, and validation behavior from the profile. Reject mixed-profile state instead of silently falling back.
- Classify pre-existing stored groups deterministically as legacy during migration.

## Out of scope

- Publishing/deleting KeyPackages and final default flip.
- Group migration or legacy-group rejoin.
- Push owner proof kind 451 and multi-device authorization kind 452.

## Acceptance criteria

- Legacy fixtures reopen with the same group state and behavior.
- Current proof fixtures round-trip exactly and invalid kind/tags/content/timestamp/length are rejected.
- Mixed legacy/current components are rejected with a typed error.
- Profile survives restart and is exposed to higher layers where behavior must branch.
- No new code calls the old extension `v2`; names distinguish legacy extension from current component.
- `just fast-ci` and targeted storage/engine tests pass.

PR #1067, opened 2026-07-23
Persist legacy/current protocol profiles

erskingardner:
## Summary

- persist an explicit Marmot application `ProtocolProfile` on groups and KeyPackages
- default records written before profile classification to `Legacy`
- preserve explicit `Current` values across serialization and SQLite persistence
- classify the current create/join paths as legacy until the strict cutover lands
- bind KeyPackage wrapper metadata to the decoded legacy account-proof carrier so callers cannot relabel legacy wire bytes as current
- fail closed when a present stored profile has an unknown value
- include the security-relevant profile in KeyPackage equality and hashing while continuing to ignore transport provenance

`ProtocolProfile` is the strict-cutover application-profile generation. It governs differences such as the account-proof carrier, media component, and mixed-profile rejection; it does not identify the MLS application-data carrier. A legacy-classified group may already use `app_data_dictionary`.

This remains the first, deliberately narrow slice of #1048. The current account-proof component and creation-time strict cutover are implemented in the stacked follow-up. Higher-layer profile exposure and behavioral branching are likewise deferred to those profile-specific follow-ups rather than being added to this persistence slice.

## Stack

- #1066 is merged
- base branch: `master`
- next: #1069

## Validation

- `cargo test -p cgka-traits --test snapshots`
- `cargo test -p storage-sqlite corrupt_present_group_profile_fails_closed`
- `cargo test -p storage-sqlite pre_profile_group_record_reopens_as_legacy`
- `cargo test -p cgka-engine --test group_creation create_group_rejects_relabelled_legacy_key_package_profile -- --exact`
- `cargo test -p cgka-engine --test sqlite_storage`
- `just fast-ci`

## Summary by CodeRabbit

- **New Features**
  - Added persisted protocol profile classification for groups and key packages, supporting both legacy and current variants.
  - Profile metadata is retained through storage round-trips and engine reopen scenarios.
  - Missing profile fields default deterministically to the legacy variant for backward compatibility.

- **Bug Fixes**
  - Enforces that transported key packages are tagged with the legacy protocol profile when required by account identity proofs.
  - Rejects relabeled legacy key packages with incorrect profile metadata and surfaces clear errors for unknown/invalid profile values.

- **Tests**
  - Expanded integration and storage test coverage to verify profile preservation, defaults, and failure cases.

### DIFF
--- a/crates/cgka-engine/src/group_lifecycle.rs
+++ b/crates/cgka-engine/src/group_lifecycle.rs
@@ -23,7 +23,7 @@ use cgka_traits::app_components::{AppComponentSet, default_group_components};
 use cgka_traits::capabilities::{GroupCapabilities, TransportKind};
 use cgka_traits::engine::{CreateGroupRequest, KeyPackage, SendResult, WelcomeMetadata};
 use cgka_traits::error::EngineError;
-use cgka_traits::group::{Group, Member};
+use cgka_traits::group::{Group, Member, ProtocolProfile};
 use cgka_traits::message::{MessageRecord, MessageState, StoredMessagePayload};
 use cgka_traits::storage::StorageProvider;
 use cgka_traits::transport::{EncryptedPayload, TransportEnvelope, TransportMessage};
@@ -340,6 +340,7 @@ impl<S: StorageProvider> Engine<S> {
             epoch: EpochId(mls_group.epoch().as_u64()),
             members: projected_members,
             required_capabilities: required_caps,
+            protocol_profile: ProtocolProfile::Legacy,
             removed: false,
             join_epoch: EpochId(mls_group.epoch().as_u64()),
         };
@@ -700,6 +701,7 @@ impl<S: StorageProvider> Engine<S> {
                     members: marmot_members(&mls_group),
                     required_capabilities:
                         crate::capability_manager::required_capabilities_from_group(&mls_group),
+                    protocol_profile: ProtocolProfile::Legacy,
                     removed: false,
                     join_epoch: EpochId(mls_group.epoch().as_u64()),
                 };
--- a/crates/cgka-engine/src/group_lifecycle.rs
+++ b/crates/cgka-engine/src/group_lifecycle.rs
@@ -340,6 +340,10 @@ impl<S: StorageProvider> Engine<S> {
             epoch: EpochId(mls_group.epoch().as_u64()),
             members: projected_members,
             required_capabilities: required_caps,
+            // This persistence-only slice predates the strict application
+            // profile cutover, so creation is intentionally classified
+            // Legacy. The value does not describe the MLS application-data
+            // carrier, which is already app_data_dictionary.
             protocol_profile: ProtocolProfile::Legacy,
             removed: false,
             join_epoch: EpochId(mls_group.epoch().as_u64()),
@@ -701,6 +705,9 @@ impl<S: StorageProvider> Engine<S> {
                     members: marmot_members(&mls_group),
                     required_capabilities:
                         crate::capability_manager::required_capabilities_from_group(&mls_group),
+                    // Welcome state remains Legacy-classified until the
+                    // strict application-profile classifier lands. This does
+                    // not imply legacy MLS application-data encoding.
                     protocol_profile: ProtocolProfile::Legacy,
                     removed: false,
                     join_epoch: EpochId(mls_group.epoch().as_u64()),
--- a/crates/storage-sqlite/src/storage/groups.rs
+++ b/crates/storage-sqlite/src/storage/groups.rs
@@ -79,6 +79,7 @@ mod tests {
         TestGroupState, gid, mid, sample_group, sample_message, sample_queued_intent,
     };
     use cgka_traits::capabilities::GroupCapabilities;
+    use cgka_traits::group::ProtocolProfile;
     use cgka_traits::storage::{
         CapabilityStorage, ConvergencePolicyStorage, GroupStorage, MessageStorage,
         OutboundIntentStorage, StorageError, StorageProvider,
@@ -89,9 +90,32 @@ mod tests {
     #[test]
     fn group_roundtrip_preserves_every_field() {
         let store = SqliteAccountStorage::in_memory().unwrap();
-        let group = sample_group(gid(1), 7, 3);
+        let mut group = sample_group(gid(1), 7, 3);
+        group.protocol_profile = ProtocolProfile::Current;
         store.put_group(&group).unwrap();
         assert_eq!(store.get_group(&group.id).unwrap(), group);
+    }
+
+    #[test]
+    fn pre_profile_group_record_reopens_as_legacy() {
+        let store = SqliteAccountStorage::in_memory().unwrap();
+        let group = sample_group(gid(1), 7, 3);
+        let mut record = serde_json::to_value(&group).unwrap();
+        record.as_object_mut().unwrap().remove("protocol_profile");
+        let bytes = serde_json::to_vec(&record).unwrap();
+        store
+            .lock()
+            .unwrap()
+            .execute(
+                "INSERT INTO cgka_groups (id, epoch, record) VALUES (?1, ?2, ?3)",
+                rusqlite::params![group.id.as_slice(), 7, bytes],
+            )
+            .unwrap();
+
+        let reopened = store.get_group(&group.id).unwrap();
+        assert_eq!(reopened.protocol_profile, ProtocolProfile::Legacy);
+        assert_eq!(reopened.name, group.name);
+        assert_eq!(reopened.members, group.members);
     }
 
     #[test]
--- a/crates/storage-sqlite/src/storage/groups.rs
+++ b/crates/storage-sqlite/src/storage/groups.rs
@@ -113,9 +113,32 @@ mod tests {
             .unwrap();
 
         let reopened = store.get_group(&group.id).unwrap();
-        assert_eq!(reopened.protocol_profile, ProtocolProfile::Legacy);
-        assert_eq!(reopened.name, group.name);
-        assert_eq!(reopened.members, group.members);
+        assert_eq!(reopened, group);
+    }
+
+    #[test]
+    fn corrupt_present_group_profile_fails_closed() {
+        let store = SqliteAccountStorage::in_memory().unwrap();
+        let group = sample_group(gid(1), 7, 3);
+        let mut record = serde_json::to_value(&group).unwrap();
+        record["protocol_profile"] = serde_json::json!("nope");
+        let bytes = serde_json::to_vec(&record).unwrap();
+        store
+            .lock()
+            .unwrap()
+            .execute(
+                "INSERT INTO cgka_groups (id, epoch, record) VALUES (?1, ?2, ?3)",
+                rusqlite::params![group.id.as_slice(), 7, bytes],
+            )
+            .unwrap();
+
+        let error = store
+            .get_group(&group.id)
+            .expect_err("a present corrupt profile must not fall through the serde default");
+        assert!(
+            matches!(&error, StorageError::Serialization(message) if message.contains("unknown variant")),
+            "unexpected error: {error:?}"
+        );
     }
 
     #[test]
--- a/crates/traits/src/engine.rs
+++ b/crates/traits/src/engine.rs
@@ -60,9 +60,12 @@ pub struct KeyPackage {
     pub bytes: Vec<u8>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub source: Option<KeyPackageSource>,
-    /// Persisted wire-compatibility profile. Raw KeyPackages constructed by
-    /// pre-classification callers default to legacy and must still be checked
-    /// against their decoded wire state by the engine.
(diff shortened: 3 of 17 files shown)


erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

erskingardner (commented):
removed

crates/traits/src/group.rs
```diff
@@ -9,6 +9,20 @@ use crate::capabilities::GroupCapabilities;
...
 
+/// Wire-compatibility profile governing every protocol decision for a group or
+/// KeyPackage.
+///
+/// Existing persisted records predate this field and therefore deserialize as
+/// [`ProtocolProfile::Legacy`]. Current-profile state is always explicit; code
+/// must not infer a hybrid profile independently for each component.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ProtocolProfile {
```
erskingardner, line 27: **Semantics vs reality.**

The rustdoc says this governs “every protocol decision” / wire compatibility, but create/join hardcode `Legacy` while post-#1066 groups already use current MLS `app_data_dictionary` / last-resort component `0x0004`.

Please document that `ProtocolProfile` is the account-proof / application-profile cutover bit (`0xf2f1` vs `0x8009`, media, mixed-profile rejection) — not “which MLS carrier we speak.” Otherwise the next stacked PR will dispatch MLS behavior off a bit that is still forcibly `Legacy`.

crates/cgka-engine/src/group_lifecycle.rs
```diff
@@ -340,6 +340,7 @@ impl<S: StorageProvider> Engine<S> {
             epoch: EpochId(mls_group.epoch().as_u64()),
             members: projected_members,
             required_capabilities: required_caps,
+            protocol_profile: ProtocolProfile::Legacy,
```
erskingardner, line 347: Hardcoding `Legacy` until strict cutover is the right fail-safe for this slice.

Please mirror that intent in the `ProtocolProfile` rustdoc / PR body so readers don’t treat this write as “this group speaks legacy MLS wire.” Same on the join path.

crates/cgka-engine/src/group_lifecycle.rs
```diff
@@ -700,6 +701,7 @@ impl<S: StorageProvider> Engine<S> {
                     members: marmot_members(&mls_group),
                     required_capabilities:
                         crate::capability_manager::required_capabilities_from_group(&mls_group),
+                    protocol_profile: ProtocolProfile::Legacy,
```
erskingardner, line 711: Same as create: intentional `Legacy` stamp on join. Good for cutover safety; keep the docs aligned so “Legacy profile” ≠ “legacy MLS encoding.”

crates/traits/src/engine.rs
```diff
@@ -60,23 +60,35 @@ pub struct KeyPackage {
     pub bytes: Vec<u8>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub source: Option<KeyPackageSource>,
+    /// Persisted wire-compatibility profile. Raw KeyPackages constructed by
+    /// pre-classification callers default to legacy and must still be checked
+    /// against their decoded wire state by the engine.
+    #[serde(default)]
+    pub protocol_profile: ProtocolProfile,
```
erskingardner, line 70: **Footgun: lying metadata.**

Every constructor stamps `Legacy`, `fresh_key_package` uses `KeyPackage::new`, and nothing classifies from wire yet. The comment says callers “must still be checked against decoded wire state,” but that check does not exist.

Also: existing `PartialEq`/`Hash` for `KeyPackage` still compare only `bytes`, so profile mismatches are invisible to `assert_eq!`.

Before any code dispatches on `key_package.protocol_profile`, either classify at parse time (set from wire; mismatch → typed error) or make the “unclassified until cutover” rule explicit and grep-banned for dispatch.

crates/storage-sqlite/src/storage/groups.rs
```diff
@@ -89,11 +90,34 @@ mod tests {
...
+            .lock()
+            .unwrap()
+            .execute(
+                "INSERT INTO cgka_groups (id, epoch, record) VALUES (?1, ?2, ?3)",
+                rusqlite::params![group.id.as_slice(), 7, bytes],
+            )
+            .unwrap();
+
+        let reopened = store.get_group(&group.id).unwrap();
+        assert_eq!(reopened.protocol_profile, ProtocolProfile::Legacy);
```
erskingardner: This is the load-bearing migration test — good.

Please also pin the explicit-bad-value path: a record with `"protocol_profile": "nope"` must fail `get_group` / deserialize (not fall through `#[serde(default)]`). Default only helps when the field is **absent**; corrupt present values should fail closed.

crates/storage-sqlite/src/storage/groups.rs
```diff
@@ -89,11 +90,34 @@ mod tests {
     #[test]
     fn group_roundtrip_preserves_every_field() {
         let store = SqliteAccountStorage::in_memory().unwrap();
-        let group = sample_group(gid(1), 7, 3);
+        let mut group = sample_group(gid(1), 7, 3);
+        group.protocol_profile = ProtocolProfile::Current;
```
erskingardner, line 94: Good that `Current` is forced through `put_group`/`get_group` — without this, only the default arm would be proven.

Optional follow-up: one engine-level create → reopen asserting `protocol_profile == Legacy` so the hardcoded lifecycle write is covered, not only `sample_group`.

crates/traits/tests/snapshots.rs
```diff
@@ -662,12 +662,28 @@ fn snapshot_group_and_member() {
...
+            protocol_profile: ProtocolProfile::Current,
             removed: false,
             join_epoch: EpochId(2),
         }
     );
 }
 
+#[test]
+fn key_package_profile_is_persisted_and_old_records_default_to_legacy() {
+    let legacy: KeyPackage = serde_json::from_value(serde_json::json!({
```
erskingardner, line 674: Solid KeyPackage serde coverage for missing → Legacy and Current round-trip.

Same ask as groups: add an invalid-variant JSON case (`"protocol_profile": "nope"`) so the KeyPackage boundary fails closed too.

erskingardner (commented):
## Adversarial review (profile commit only)

Scope: `a610c389` (“persist Marmot protocol profiles”). Inherited #1066 stack left alone except where it creates profile/wire tension.

This is a good narrow slice of #1048 — `ProtocolProfile`, serde defaulting, create/join staying on `Legacy`, and the pre-profile SQLite reopen test are the right shape. The main risk is **semantic overclaim + an inert KeyPackage field** that will be dangerous the moment something starts dispatching on it.

### Should fix / clarify before merge
1. **Tighten `ProtocolProfile` docs** — it is not “which MLS wire we speak” after #1066; say which decisions it will gate (account proof / media / mixed-profile rejection).
2. **KeyPackage profile footgun** — always stamped `Legacy`, ignored by `Eq`/`Hash`, no wire classification yet. Either classify at parse time or explicitly forbid dispatch until cutover.
3. **Call out deferred AC** — #1048 wants the profile “exposed to higher layers where behavior must branch”; this PR persists it on `Group` but does not surface it via app/UniFFI. Fine if deferred explicitly.
4. **Confirm merge base** — GitHub base is `master` while the body says stacked on #1066; don’t accidentally land the stack out of order.

### Nice-to-have
- Invalid `"protocol_profile": "nope"` fail-closed tests for Group + KeyPackage
- Engine create → reopen smoke for the hardcoded `Legacy` write

### What’s good
- JSON-blob `#[serde(default)]` migration (no SQL churn)
- `pre_profile_group_record_reopens_as_legacy` + `Current` round-trip
- Hardcoding `Legacy` on create/join until cutover

> GitHub disallows `REQUEST_CHANGES` on your own PR, so this is a comment review; please treat the “should fix / clarify” items as blocking for the stacked follow-ups even if you merge this thin slice.

> Ignore any earlier `_Superseded probe_` review shells from placing inline anchors.

erskingardner (commented):

Follow-up to the earlier review (#pullrequestreview-4771275978): I re-derived everything from scratch at head `a610c389`, built the workspace, ran every test command in the PR's validation section plus the full suites those files belong to (`cgka-traits` snapshots, `storage-sqlite` group tests, `cgka-engine` group_creation/update_group_data/hydration_quarantine/application_data_wire, `marmot-account`, `marmot-app`, `cgka-conformance-simulator`), and ran `just fast-ci` end to end — all clean. No functional bugs found; every `Group`/`KeyPackage` construction site was updated in lockstep. Where my conclusions differ from the earlier review after verification, I say so below.

### The one thing to fix before merge: rebase past the #1066 squash

The earlier review asked to "confirm merge base" — I did, and it's worse than cosmetic. The branch's merge-base with `origin/master` is `40df95c9` (pre-#1066), so the PR diff presents ~430 of ~520 changed lines as new even though I verified they are **byte-for-byte identical to current `master`** (landed hours ago via #1066's squash `bcb73ac0`) — checked file-by-file for every non-profile file in the diff (`Cargo.toml`/`Cargo.lock` rev bump, `capabilities.rs`, `key_package.rs`, `application_data_wire.rs`, `group_creation.rs`, `custom_extensions.md`):

```
$ git merge-base origin/master <pr-head>   # → 40df95c9, not bcb73ac0
$ diff <(git show origin/master:crates/cgka-engine/src/key_package.rs) \
       <(git show <pr-head>:crates/cgka-engine/src/key_package.rs)      # empty
```

Only `a610c389` (14 files, ~88+/8−) is genuinely new. Beyond the misleading diff, a **"Rebase and merge" would replay the two now-redundant commits onto `master`** and likely stall or emit empty commits. Recommend rebasing `codex/issue-1048-account-proof` onto current `master` (or cherry-picking just `a610c389` onto a fresh branch) so the diff and the merge reflect only the profile-persistence change.

### Non-blocking findings

1. **`ProtocolProfile` doc comment invites conflation with #1066** (`crates/traits/src/group.rs:12`). "Wire-compatibility profile governing every protocol decision" reads like it covers the MLS app-data/last-resort carrier — but #1066 already switched everyone to the new carrier unconditionally, and every post-#1066 group is still (correctly) tagged `Legacy` here. Per #1048 this enum is about the `0xf2f1` → `0x8009` account-identity-proof split. Reference that (or at least "see #1048") so readers don't conclude the tagging is inconsistent. (Confirms the earlier review's point 1.)

2. **KeyPackage's hand-written `PartialEq`/`Hash` silently exclude the new field** (`crates/traits/src/engine.rs:97`). The earlier review flagged the always-`Legacy`, equality-ignored KeyPackage field as a "should fix" footgun. Having traced it, I'd soften that: the manual impls already excluded `source` by design (wire bytes are canonical identity), and nothing dispatches on the field yet — grep confirms `.protocol_profile` is only read at the three write sites and one test. So it's consistent, not broken. The cheap insurance is a one-line comment on the manual impls ("intentionally excludes source/protocol_profile — bytes are identity") and classifying at parse time when the cutover PR lands.

3. **`pre_profile_group_record_reopens_as_legacy` asserts only two fields** (`crates/storage-sqlite/src/storage/groups.rs:105`). `Group` derives `PartialEq` and the sibling test asserts full struct equality; this one can too — `assert_eq!(reopened, Group { protocol_profile: ProtocolProfile::Legacy, ..group })` — and get free coverage of `required_capabilities`, `removed`, `epoch`, `join_epoch`, `description` for the exact claim this PR makes ("reopens without losing state"). The raw-`INSERT` fixture approach itself is exactly right.

4. **State explicitly that #1048's "exposed to higher layers" AC is deferred.** Verified: `protocol_profile` isn't surfaced through `marmot-uniffi`'s hand-mirrored `AppGroupRecordFfi`/`AccountKeyPackageFfi` (no FFI break, but no exposure either) and isn't read by any engine logic. Reasonable scope cut for a persistence-only slice — just say so in the PR body so #1048 isn't closed against this PR alone. (Confirms the earlier review's point 3.)

### What's done well

- `#[serde(default)]` + `#[default] Legacy` is the right migration posture for the JSON-blob records — no SQL schema churn, and backward compatibility is proven by tests in both directions (absent key → `Legacy`, explicit `Current` → round-trips) at both the snapshot and SQLite layers.
- Hardcoding `Legacy` on every current create/join path until strict cutover is the correct fail-safe default, and the PR description says so explicitly.
- Same migration pattern as the existing `removed`/`join_epoch` fields — consistent with how this codebase already does record evolution.

Verified clean: `cargo check --workspace --all-targets`, all cited test suites (100+ tests), `just fast-ci` with `-D warnings` across both feature sets. The OpenMLS fork pin in the diff is inherited from #1066 (already on master), not new here. 🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

Issue #1053, opened 2026-07-23
Authorize standalone proposals before queueing or auto-commit


erskingardner:
## Summary

Apply Marmot authorization and capability rules to standalone proposals when they are received, before they enter the pending proposal store or influence auto-commit behavior.

A proposal that is valid MLS syntax is not automatically authorized Marmot input. Deferring authorization until a later commit allows invalid proposals to consume state, affect scheduling, or be committed by another member.

## Normative source

Adopted by marmot-protocol/marmot#408:

- `protocol-core/group-messaging.md`
- `protocol-core/inbound-processing.md`
- `protocol-core/group-state.md`
- `protocol-core/member-departure.md`
- `foundation/authorization-proofs.md`

## Start here

- `crates/cgka-engine/src/message_processor/ingest.rs`
- `crates/cgka-engine/src/app_components.rs`
- `crates/cgka-engine/src/auto_committer.rs`
- `crates/cgka-engine/src/pending_commit_guard.rs`
- proposal persistence in engine/storage traits

## Required behavior

- After cryptographic/MLS parsing but before queueing, authorize the standalone proposal against the current profile, epoch, sender, membership, admin policy, required capabilities, and proposal-specific rules.
- Apply the same semantic authorization to by-reference and by-value proposal paths where applicable.
- Unauthorized/unsupported/stale proposals do not enter durable pending state, schedule SelfRemove work, block local actions, or affect deterministic selection.
- Preserve the distinction between malformed, unauthenticated, unauthorized, unsupported, stale, duplicate, and deferred inputs in typed outcomes/audit data.
- Revalidate a queued proposal when the state against which it was authorized has changed before commit.
- Keep legacy-group rules profile-specific.

## Acceptance criteria

- Adversarial tests cover non-admin privileged proposals, unsupported proposal types, wrong epoch/sender, invalid component updates, invalid SelfRemove, and proposal/commit seam parity.
- Restart tests prove rejected proposals are not hydrated from durable pending state.
- A rejected proposal cannot trigger the auto-committer or alter a later commit.
- Audit/telemetry remains privacy-safe and distinguishes the rejection class.
- Conformance vectors exercise direct and wrapped inbound seams.
- `just fast-ci` and targeted engine/conformance tests pass.

Issue #1052, opened 2026-07-23
Implement frozen Media V2 component 0x800b while preserving legacy Media V1


erskingardner:
## Summary

Implement the adopted frozen Media V2 component (`0x800b`) end to end. Current-profile groups require Media V2. Existing legacy groups continue to load and send Media V1 (`0x8008`). There is no in-place media/profile migration.

## Normative source

Adopted by marmot-protocol/marmot#408:

- `app-components/group-encrypted-media-v2.md`
- `features/encrypted-media.md`
- frozen legacy references: `app-components/group-encrypted-media-v1.md`, `features/encrypted-media-v1.md`
- `foundation/registries.md`

## Dependencies

- #1048 explicit profile classification
- #1050 current-profile required-component enforcement
- Existing FFI gap #955 should be completed as part of or immediately after this work

## Start here

- `crates/marmot-app/src/media/`
- `crates/traits/src/app_components/`
- `crates/storage-sqlite/src/encrypted_media_secrets.rs`
- `crates/marmot-uniffi/src/commands/media.rs`
- `crates/marmot-uniffi/src/conversions/media.rs`
- CLI media commands and conformance vectors

## Required behavior

- Add a distinct Media V2 model/codec for component `0x800b`; do not reinterpret or mutate V1 `0x8008`.
- Current groups require and emit V2. Legacy groups continue to parse/load and emit V1.
- Implement the exact MIME canonicalization: trim only HTAB, LF, FF, CR, and space; do not trim vertical tab or arbitrary Unicode whitespace. Enforce the required slash/token/length grammar.
- Enforce filename, URL, size/hash/key/nonce, and attachment-local constraints exactly as specified.
- Keep message attachment metadata and group component validation version-aware.
- Do not silently downgrade a current group to V1 or upgrade a legacy group to V2.
- Expose versioned build/parse results through UniFFI without lossy conversion.

## Acceptance criteria

- Golden vectors cover V1 and V2 independently and prove their type IDs/encodings cannot be confused.
- Boundary tests cover every allowed trim byte, vertical-tab rejection/non-trimming, Unicode whitespace, MIME separators/tokens/lengths, filenames, URLs, and attachment-local rules.
- Existing legacy-group fixtures can still load and send V1 media.
- Newly created current groups reject missing/V1-only state and use V2.
- Swift/Kotlin generated APIs represent both versions explicitly.
- `just fast-ci`, targeted media/storage/UniFFI tests, and binding builds pass.

Issue #1050, opened 2026-07-23
Enforce current-profile group invariants on every resulting state


erskingardner:
## Summary

Enforce the adopted current application profile as resulting-state invariants, not merely creation-time defaults. Validate locally before staging and validate inbound commits before accepting their state.

## Normative source

Adopted by marmot-protocol/marmot#408:

- `protocol-core/group-state.md`
- `protocol-core/group-messaging.md`
- `protocol-core/group-setup.md`
- `foundation/mls-protocol.md`
- `foundation/registries.md`

## Dependencies

- #1047 application-data carriers
- #1048 profile classification/current identity proof

## Start here

- `crates/cgka-engine/src/app_components.rs`
- `crates/cgka-engine/src/group_lifecycle.rs`
- `crates/cgka-engine/src/group_state_changes.rs`
- `crates/cgka-engine/src/update_group_data.rs`
- `crates/cgka-engine/src/message_processor/ingest.rs`
- `crates/traits/src/capabilities.rs`

## Required behavior

For every current-profile group state:

- GroupContext application data requires components `0x8003` (admin policy) and `0x8009` (account identity proof policy/state as specified).
- The required capabilities include proposal `0x0008` and all required current-profile extension/component capabilities.
- Creation, Welcome join, external joins if supported, local commits, and inbound commits all validate the complete resulting state.
- Reject duplicate component types and malformed component framing.
- A commit may contain at most one application-data update proposal and at most one update for a given component.
- Component `Unrequire` plus removal is atomic and allowed where specified, but mandatory current-profile components/capabilities cannot be removed or unrequired.
- Unknown optional components are preserved across unrelated updates.
- Legacy groups continue under legacy rules selected by explicit profile; do not impose current mandatory components on them.

## Acceptance criteria

- Table-driven tests cover each mutation surface and each missing/duplicate mandatory item.
- Tests prove validation uses the resulting state, including add/update/remove/unrequire combinations.
- An invalid inbound commit does not partially mutate MLS or projected application state.
- Unknown optional components survive unrelated commits byte-for-byte.
- Legacy fixtures remain valid under legacy validation.
- Conformance vectors encode the required component/capability set.
- `just fast-ci` and targeted engine/conformance tests pass.

PR #1069, opened 2026-07-23
Implement current account identity proof component

### LINKED ISSUE #1048: Classify legacy/current profiles and implement account identity proof component 0x8009

erskingardner:
## Summary

- preserve the deployed 0xf2f1 proof extension for legacy KeyPackages and groups
- implement the adopted 104-byte 0x8009 LeafNode component and exact kind-450 signing event for the current profile
- classify KeyPackages, groups, welcomes, and staged commits from their wire proof carrier and reject mixed or profile-changing state
- expose the persisted profile through the app and UniFFI group surfaces while keeping new-state defaults legacy until the coordinated cutover
- derive kind-30443 capability tags from the decoded KeyPackage and require exact canonical tag sets, rejecting missing, extra, duplicate, or mixed-profile markers

## Compatibility

A current-configured engine can reopen and continue using a legacy group. New current-profile KeyPackages and groups omit 0xf2f1, require 0x8009, and cannot be mixed with legacy state. This PR deliberately does not flip the production default or publish/delete policy.

Current proof material is regenerated for each account-device engine session because the kind-450 event uses a fresh `created_at`. Receivers enforce the NIP-01 integer range but, per the adopted profile, do not impose an age or future-skew limit.

## Scope

This is a partial implementation of #1048. Media V2 and the coordinated strict production cutover remain in later stack PRs, so this PR must not close #1048 by itself.

## Stack

#1066 and #1067 are merged. This branch is rebased directly onto current `master`.

## Validation

- `cargo test -p cgka-engine --test hydration_quarantine hydration_classifies_stored_wire_profile_mismatch_as_member_validation_failure`
- `cargo test -p marmot-app key_package_id_list_tag_must_be_exactly_one`
- `just fast-ci`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added Legacy and Current protocol profiles for account identity proofs.
  * Sessions can choose the profile used for newly created groups and key packages.
  * Group and MLS state now expose the active protocol profile, including in the UniFFI layer.
  * Current-profile proofs are supported via a dedicated app component; legacy uses an extension.
* **Bug Fixes**
  * Rejects missing, mixed, tampered, or profile-mismatched identity proofs and enforces single-profile behavior.
  * Enforces exact KeyPackage metadata/tag matching and capability declarations.
  * Detects persisted wire/protocol-profile mismatches during hydration and quarantines affected groups.


agent-p1p (changes requested):
Pip adversarial review — exact head `d606d877b0d1a45385e37b7a15ee2791bff39008`

BLOCKING:
- Current profile remains coupled to legacy Encrypted Media V1: `crates/marmot-app/src/client/mod.rs:242-258,1991-2010` unconditionally adds `GROUP_ENCRYPTED_MEDIA_COMPONENT_ID` and constructs `EncryptedMediaPolicyV1`, while `crates/traits/src/app_components/mod.rs:66,86,88` defines only component `0x8008` / `encrypted-media-v1`. Engine creation and update validation at `crates/cgka-engine/src/app_components.rs:686-703,707-733` accepts that V1 component without consulting `ProtocolProfile`. There is no `0x800b` / `encrypted-media-v2` implementation or profile gate anywhere in the tree. The adopted current profile names `marmot.group.encrypted-media.v2` (`0x800b`) as current and says current-profile senders MUST NOT create V1 references; issue #1048 explicitly requires media, proof, capability, and validation behavior to dispatch from the single persisted profile. As written, a `Current` engine can create a hybrid Current-proof/Media-V1 group. Implement V2 policy/reference handling and branch group creation, sends, reads, and component validation on `ProtocolProfile`; reject `0x8008` as current state and `0x800b` as legacy state rather than inferring per component.

- Standalone Add/Update proposals bypass the new proof/profile validator: `crates/cgka-engine/src/message_processor/ingest.rs:1208-1241` returns `Processed` for every `ProposalMessage` after only the SelfRemove auto-commit decision, and replay stores the same proposal at `crates/cgka-engine/src/openmls_projection.rs:2042-2053`. `validate_staged_commit_account_identity_proofs` is only called later from commit processing. OpenMLS cannot compensate because it does not know Marmot's `0x8009`/`0xf2f1` proof semantics or `ProtocolProfile`. Commit-time validation prevents a malformed referenced proposal from entering canonical state, but the adopted proof spec separately requires clients to reject a proposed LeafNode with a missing, malformed, mismatched, or mixed proof. Validate standalone Add and Update proposals at both direct-ingest and replay chokepoints before marking/storing them, using the same shared profile/proof logic as commits, and add regression coverage for both seams.

SUGGESTION:
- Validate transported KeyPackage proofs against the KeyPackage's own ciphersuite: `crates/cgka-engine/src/key_package.rs:191-194` passes `self.ciphersuite`, despite the adopted proof spec requiring `key_package.ciphersuite()` and the public metadata/last-resort helpers already doing that at lines 55-58 and 92-95. The current single-suite builder restriction makes the values equal for addable KeyPackages today, so this is latent rather than immediately exploitable, but the nearby mdk#747 comments/tests incorrectly claim the invite path already follows the per-KeyPackage rule. Use `key_package.ciphersuite()` now and add an actual differing-suite regression so future suite enablement cannot silently validate the wrong signing event.

- Add current-proof negative-path tests around `crates/cgka-engine/src/account_identity_proof.rs:317-405`. The exact event/vector and timestamp bounds are covered, but the acceptance criteria also call for invalid current proofs to be rejected. Exercise the LeafNode validation seam with non-104-byte data, a tampered signature, signer/credential mismatch, leaf-key mismatch, and a `0x8009` data entry omitted from the leaf support list.

NITPICK:
- none

SUMMARY: The 104-byte `0x8009` encoding, exact kind-450 signing vector, persisted proof-profile classification, restart reconciliation, and commit-time mixed-proof rejection are solid, and the focused proof/profile/storage tests plus cgka-engine clippy pass locally. The implementation is not complete against #1048, however: current groups still use the frozen legacy media profile, and invalid standalone proposed leaves are accepted until a later commit attempts to use them.

VERDICT: REQUEST_CHANGES

SENSITIVE_PATHS: `crates/cgka-engine/src/account_identity_proof.rs`, `crates/cgka-engine/src/app_components.rs`, `crates/cgka-engine/src/capabilities.rs`, `crates/cgka-engine/src/capability_manager.rs`, `crates/cgka-engine/src/engine.rs`, `crates/cgka-engine/src/group_lifecycle.rs`, `crates/cgka-engine/src/identity.rs`, `crates/cgka-engine/src/key_package.rs`, `crates/cgka-engine/src/message_processor/send.rs`, `crates/cgka-engine/src/update_group_data.rs`, `crates/cgka-engine/src/upgrade.rs`, `crates/cgka-session/src/lib.rs`

agent-p1p:
Addressed both non-blocking suggestions in `ef0f1c3`:

- `parse_key_package` now validates the proof against `key_package.ciphersuite()`; the new differing-suite regression failed with the exact proof-suite mismatch before the fix and passes now.
- Added direct LeafNode validation coverage for non-104-byte data, tampered signatures, signer/credential mismatch, leaf-key mismatch, and omitted `0x8009` support advertisement.

The two blocking items are already split into downstream adopted-spec issues and implemented PRs; moving them into this foundational PR would duplicate and invert the existing stack:

- Media V2 is #1052 / #1075. #1052 explicitly depends on #1048 and #1050; #1075 implements the distinct `0x800b` model, profile-selected create/send/read/update behavior, reference validation, storage, and bindings on top of this proof/profile foundation.
- Standalone proposal authorization is #1053 / #1078. #1053 owns authorization before queue/auto-commit plus direct/replay/convergence parity and typed outcomes; #1078 implements that broader chokepoint.

PR #1069 intentionally establishes profile classification, the current proof carrier, and proof/profile validation while leaving the production default unchanged, as stated in its Compatibility section. Pulling #1075 and #1078 backward here would collapse separate accepted issues and make their current dependency order circular.

Verification at `ef0f1c3`: `cargo test -p cgka-engine` and `just fast-ci` pass. Please re-review this PR at the new head with the dependent-PR split above in scope.

agent-p1p:
Follow-up `bd62a62` addresses the fresh re-review's remaining own-suite consistency finding:

- KeyPackage capability caching and staged-Add proof revalidation now derive the ciphersuite from each authenticated KeyPackage.
- Removed the redundant engine-suite parameters at all four callers.
- Extended legacy/current coexistence coverage through an actual Invite commit and confirmation, asserting the reopened legacy group stays Legacy.

Fresh read-only Cursor adversarial review: APPROVE, no blockers. Local verification remains green: focused coexistence test, `cargo test -p cgka-engine --locked`, `just fast-ci`, `cargo fmt --all --check`, and `git diff --check`.

The Media V2 and standalone-proposal findings remain correctly assigned to dependent PRs #1075 and #1078, respectively; the independent re-review explicitly approved that scope split.

agent-p1p (dismissed):
Re-reviewed at bd62a62 after the own-suite consistency follow-up. The KeyPackage proof seams and legacy-group commit coexistence coverage are correct. Media V2 (#1075) and standalone proposal authorization (#1078) are dependent PR scope; pulling either backward would invert the accepted stack. No blocking findings.

crates/traits/src/app_components/mod.rs
```diff
@@ -64,6 +64,11 @@ pub const GROUP_MESSAGE_RETENTION_COMPONENT_ID: AppComponentId = 0x8005;
 pub const AGENT_TEXT_STREAM_QUIC_COMPONENT_ID: AppComponentId = 0x8006;
 pub const GROUP_AVATAR_URL_COMPONENT_ID: AppComponentId = 0x8007;
 pub const GROUP_ENCRYPTED_MEDIA_COMPONENT_ID: AppComponentId = 0x8008;
+/// LeafNode-only account identity proof binding the Marmot account key to the
+/// MLS signature key. The proof bytes live in the LeafNode
+/// `app_data_dictionary`; GroupContext dictionaries only require the id via
+/// `app_components`.
+pub const ACCOUNT_IDENTITY_PROOF_COMPONENT_ID: AppComponentId = 0x8009;
```
erskingardner, line 78: Good to land `0x8009` here.

This PR still only defines media as `0x8008` / `encrypted-media-v1`. A `Current` engine can therefore create Current-proof + Media-V1 groups until #1075. That is acceptable as stack ordering **only if** this PR does not close #1048 — the issue explicitly requires media behavior to dispatch from the same profile.

crates/cgka-engine/src/engine.rs
```diff
@@ -846,6 +863,9 @@ impl<S: StorageProvider> Engine<S> {
             .storage
             .get_group(group_id)
             .map_err(|_| GroupHydrationQuarantineReason::GroupRecordLoadFailed)?;
+        if group.protocol_profile != wire_protocol_profile {
+            return Err(GroupHydrationQuarantineReason::GroupRecordLoadFailed);
```
erskingardner: Fail-closed on persisted-vs-wire profile disagreement is correct.

Please do **not** map this to `GroupRecordLoadFailed`. That reason means “could not read/decode the Marmot record,” which will send operators down the wrong rabbit hole. Add a dedicated quarantine reason (e.g. `ProtocolProfileMismatch` / `MemberValidationFailed` with a stable audit code) so tests and forensics can see a profile split-brain.

crates/marmot-app/src/key_package_records.rs
```diff
@@ -288,6 +288,19 @@ pub(crate) fn key_package_from_record(
...
+            &event,
+            "mls_extensions",
+            &format!("0x{ACCOUNT_IDENTITY_PROOF_EXTENSION_TYPE:04x}"),
+        )?,
+        ProtocolProfile::Current => require_multi_value_key_package_tag_contains(
+            &event,
+            "app_components",
+            &format!("0x{ACCOUNT_IDENTITY_PROOF_COMPONENT_ID:04x}"),
+        )?,
+    }
```
erskingardner: Positive marker check is necessary but not sufficient for the published event contract.

A Current KP event can still list `0xf2f1` under `mls_extensions` (and a Legacy event can list `0x8009` under `app_components`) and pass. MDK will classify from wire bytes, but other consumers of kind-30443 tags will see a mixed advertisement.

Please reject events that advertise both markers, or require the tag sets to match `key_package_metadata`’s decoded extension/component lists.

crates/cgka-engine/src/identity.rs
```diff
@@ -67,40 +75,82 @@ impl Identity {
...
         let credential_with_key = CredentialWithKey {
             credential: credential.into(),
             signature_key: signer.public().into(),
         };
-        let account_identity_proof_extension =
-            crate::account_identity_proof::account_identity_proof_extension(
+        let created_at = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .map_err(|error| format!("system clock precedes Unix epoch: {error}"))?
+            .as_secs();
```
erskingardner, line 108: Even when an account-device signer binding already exists, this path re-mints proof material with `created_at = now`.

For Current that yields new `0x8009` bytes on every `EngineBuilder::build()`. That is fine for fresh KeyPackages, but it means Identity leaf-proof bytes are **session-scoped**, not durable. Please call that out in rustdoc (and avoid any future code that assumes the local proof component is stable across restarts).

crates/cgka-engine/src/key_package.rs
```diff
@@ -177,14 +188,28 @@ impl<S: StorageProvider> Engine<S> {
...
+        ensure_key_package_profile(kp, protocol_profile)?;
         Ok(key_package)
     }
 }
 
+fn ensure_key_package_profile(
+    key_package: &KeyPackage,
+    wire_profile: ProtocolProfile,
+) -> Result<(), EngineError> {
+    if key_package.protocol_profile != wire_profile {
```
erskingardner, line 207: Good — this closes the #1067 footgun where `protocol_profile` could lie relative to wire.

One residual app-layer sharp edge (not introduced here, but now louder): `publish_key_package` uses `is_last_resort_key_package(...).unwrap_or(false)`, so a metadata/wire mismatch `Err` is treated as “not last-resort” and forces republish. Consider propagating the error once Current KPs are in the wild so mismatch is not silently “fixed” by republishing.

crates/cgka-engine/src/capabilities.rs
```diff
@@ -171,7 +182,16 @@ pub(crate) fn self_supported_capabilities(
...
+    // fresh KeyPackage: one upgraded engine must continue operating existing
+    // legacy groups while producing current-profile state for new groups.
+    let mut out = group_capabilities_from_caps(&leaf_capabilities(
+        registry,
+        ciphersuite,
+        ProtocolProfile::Legacy,
+    ));
     out.app_components = supported_app_components.clone();
+    out.app_components
+        .insert(cgka_traits::app_components::ACCOUNT_IDENTITY_PROOF_COMPONENT_ID);
```
erskingardner, line 200: Nice dual-support posture for join: Legacy leaf ads + always-insert `0x8009` so one upgraded engine can reopen Legacy groups and join Current ones.

Consider pointing create/invite “profile must match” errors at this helper in a comment so the next reader does not “fix” join by narrowing self-support to `new_protocol_profile` only.

crates/cgka-engine/src/account_identity_proof.rs
```diff
@@ -227,20 +454,96 @@ pub(crate) fn validate_leaf_account_identity_proof(
     Ok(())
 }
 
+fn validate_current_timestamp(created_at: u64) -> Result<(), String> {
```
erskingardner, line 457: Timestamp gate is only the NIP-01 `1..=2^53-1` range (plus rejecting `0`).

If the adopted current-proof doc also requires relative skew vs “now” (max future / max age), that is not enforced here — only absolute bounds. Confirm against `account-identity-proof-v2.md`; if skew rules exist, they need tests beside `current_proof_rejects_zero_and_out_of_range_timestamps`.

erskingardner (commented):
## Adversarial review (independent, head `a5fce3b`)

Stacked on #1067/#1066. Pip’s suite/negative follow-ups look landed. The `0x8009` / kind-450 implementation, wire-driven classification, metadata↔wire checks, and commit-time mix/change rejection are solid.

### Should fix before treating #1048 as done
1. **`Fixes #1048` overclaims.** Media is still Legacy V1 (`0x8008`) with no profile gate, and standalone proposals still skip proof/profile validation (#1075 / #1078). Soften the footer to partial/#1048-proof, or keep the issue open until that stack merges.
2. **Hydration profile mismatch → `GroupRecordLoadFailed`** — fail-closed is right; the reason should be a dedicated profile/integrity quarantine, not “record load failed.”
3. **Kind-30443 tags** only require the matching proof marker; they should also reject events that advertise the other profile’s marker (or disagree with decoded metadata).

### Non-blocking
- Current proof material is re-minted (`created_at = now`) on every engine open even when the signer binding already exists — document as session-scoped.
- Dual-support `self_supported_capabilities` is a good join posture.

### What’s good
Exact 104-byte carrier, current vs legacy event shapes, typed mixed-state errors, `key_package.ciphersuite()` on proof seams, decoded kind-30443 tags, app/UniFFI profile exposure, strong engine tests (including Current engine on Legacy group).

> Comment review (GitHub disallows `REQUEST_CHANGES` on own PRs). Please treat the #1048 closure hygiene as blocking for the issue, even if this slice merges.

erskingardner (commented):

Follow-up to the earlier review (#pullrequestreview-4771402998): I re-derived everything from scratch at head `a5fce3b4`, built a worktree, and ran `cargo test -p cgka-engine -p cgka-session -p marmot-app -p marmot-uniffi` (500+ tests, all green, including every new dual-profile and fault-injection test), plus `cargo fmt --all --check`, `cargo check --workspace --all-targets` and `cargo clippy --workspace --all-targets` with `-D warnings` — all clean. The earlier review's three "should fix" findings all **held up under independent verification**; I confirm each below with traced detail, and add two findings it didn't have.

The core design is solid: wire bytes are authoritative everywhere. `ensure_key_package_profile` cross-validates the `KeyPackage.protocol_profile` cache field against the wire-derived classification at every read site, so the metadata field can't spoof past profile gates; commit-time enforcement covers Add/Update/UpdatePath leaves, the resulting `GroupContext`, and the pre-commit group profile, on both the ingest and the *sending* side; and the `VALIDATED_TREE_MARKER_VERSION` 1→2 bump correctly invalidates every cached "already validated" marker so existing groups get a full re-walk under the new rules. I traced the `leaf_extensions()` consumers to confirm a Current-configured engine safely operating a reopened Legacy group actually works (proof material is only read at fresh-KeyPackage/group-creation time, never on an existing committed leaf's update path).

### New finding: the stack inherits #1067's stale merge-base

Base branch `codex/issue-1048-account-proof` (#1067) branched from `master` at `40df95c9`, before #1066's squash `bcb73ac0` — and `git merge-base --is-ancestor bcb73ac0 origin/codex/issue-1048-account-proof` fails, i.e. the squash commit is not in this stack's history. **This PR's own diff is clean** — I independently diffed base-tip vs head-tip and got the same 23 files / ~1.4k lines as `gh pr diff`, all genuinely proof-component work. The risk is at merge time: once #1067 is rebased or squash-merged, #1069 needs its base rebased too, or "Rebase and merge" will replay commits whose content already landed. Sequence: rebase #1067 onto current `master` first, then rebase `codex/issue-1048-proof-component` onto that, before either merges.

### Confirmed findings from the earlier review

1. **Hydration profile mismatch mislabeled as `GroupRecordLoadFailed`** (`crates/cgka-engine/src/engine.rs:864`). Verified: the record loads and deserializes fine; what failed is semantic consistency with the live `GroupContext` classification. Fail-closed is right, but this belongs under `MemberValidationFailed` or a dedicated `ProtocolProfileMismatch` variant — collapsing a genuine profile split-brain into the "storage returned corrupt data" bucket makes exactly the security-relevant case invisible to audit-log triage, and this repo is explicit about keeping quarantine reasons meaningful.

2. **Kind-30443 tags: only the matching marker is asserted, never the other's absence** (`crates/marmot-app/src/key_package_records.rs:291`). Verified: an event whose wire bytes classify as `Legacy` but whose tags also advertise `0x8009` in `app_components` (or vice versa) passes untouched. MDK's engine never trusts these tags for state changes (`key_package_metadata` re-derives from wire bytes), so this isn't an internal bypass — but the tags are the public discovery-layer contract other clients filter on without decoding, and #1048's own criteria call for rejecting mixed-profile state rather than tolerating it. Reject when both markers are present or when tags disagree with the decoded metadata.

3. **`Fixes #1048` overclaims.** Verified both gaps: `validate_encrypted_media_policy` (`crates/cgka-engine/src/app_components.rs:702`) has no profile branch — media stays frozen on Legacy V1 for both profiles — and a standalone inbound `ProposalMessage` (`crates/cgka-engine/src/message_processor/ingest.rs:1207`) is accepted and marked `Processed` with no proof validation; the proof is only checked once the proposal is folded into a commit (the group stays protected, but an invalid-proof proposal sits queued unvalidated). Both look intentionally parked on the #1075/#1078-shaped follow-ups, which is a fine slice boundary — but soften the footer to "Partial for #1048" so the issue doesn't auto-close while hybrid current-proof/legacy-media groups remain fully creatable.

### New finding + smaller suggestions

- **Unflagged behavior change: tag matching tightened from case-insensitive to exact** (`crates/marmot-app/src/key_package_records.rs`, `require_multi_value_key_package_tag_contains`). This PR switches `eq_ignore_ascii_case` to `==`, which affects `mls_ciphersuite`/`mls_extensions`/`mls_proposals`/`app_components` matching generally — not just the new profile logic. Reasonable hardening (MDK's encoder always emits lowercase hex), but it can now reject third-party events with uppercase hex tags that previously passed, and it isn't mentioned in the PR description. Worth an explicit callout as intentional.
- `crates/cgka-engine/src/identity.rs:99` — the Current proof (including `created_at` + signature) is re-minted from `SystemTime::now()` on every `EngineBuilder::build()`. Not a bug (only fresh-KP/group-creation consumers), but document it as session-scoped so nobody caches or diffs "the" proof bytes across restarts. (Matches the earlier review's non-blocking note.)
- `crates/marmot-app/src/lib.rs:3403` — `.filter(|id| **id >= 0x8000)` is a bare magic number for the private-use component range; a named constant next to the existing block in `cgka_traits::app_components` would help.
- `crates/cgka-engine/AGENTS.md:125` — still describes `account_identity_proof.rs` as owning a single LeafNode extension; it now owns two wire carriers plus their classification/mixing rules.

### What's done well

Beyond the design points above: exact negative coverage for the 104-byte component (wrong length, tampered signature, credential mismatch, leaf-key mismatch, missing advertisement, mixed-carrier leaf — all asserted to reject with typed errors); `LeafNode`-only enforcement for `0x8009` defended in depth on two independent code paths; proof validation against each KeyPackage's *own* ciphersuite with a regression test asserting the right failure mode; and `key_package_metadata` now deriving published capability tags from actual decoded bytes instead of hardcoded literals — a real fix, not padding.

No blocking code issues beyond the three confirmed above; the stack rebase is the main pre-merge action. 🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

Issue #1055, opened 2026-07-23
Pin message retention decisions to each message's source epoch


erskingardner:
## Summary

Apply retained-history policy from the MLS epoch that produced the message, not from the group's current policy when a later sweep runs.

MDK already persists source-epoch data for application events, but current retention work is driven by current projection/policy. A later policy change must not retroactively reinterpret an older message contrary to the adopted source-epoch rule.

## Normative source

Adopted by marmot-protocol/marmot#408:

- `protocol-core/retained-history.md`
- `app-components/message-retention-v1.md`
- `protocol-core/group-state.md`

## Start here

- `crates/traits/src/app_event.rs`
- `crates/storage-sqlite/src/migrations/0009_app_event_source_epoch.rs`
- timeline/retention storage in `crates/storage-sqlite/`
- projection and retention scheduling in `crates/marmot-app/`
- existing retention policy issue #958 (client-policy sweep scheduling; related but not a substitute)

## Required behavior

- Persist the normalized retention decision/policy identity needed for each accepted source message at ingest, transactionally with the message.
- Resolve policy from the authenticated source epoch/state, including messages processed after the group has advanced.
- Sweeps use the stored source-epoch decision; they do not consult only the current group component.
- Define deterministic handling for legacy rows that have a source epoch but lack a recoverable source-epoch policy; do not silently apply a more destructive current policy.
- Reprocessing/duplicates cannot extend or shorten the original retention deadline.
- Keep local display/cache policy separate from protocol-required retained-history semantics.

## Out of scope

- Deleting encrypted-media epoch secrets; tracked separately.
- General runtime sweep cadence/disabled-retention UX in #958.

## Acceptance criteria

- Tests cover policy changes in both directions (retain longer and delete sooner) across epochs.
- Delayed/offline processing uses the source epoch's rule.
- Restart and duplicate delivery preserve exactly one original deadline/decision.
- Legacy-row migration has an explicit safe result and test.
- Projection/storage APIs make source-epoch policy use visible rather than implicit.
- `just fast-ci` and targeted storage/app tests pass.

PR #1074, opened 2026-07-23
fix: retire encrypted media secrets with retained messages

### LINKED ISSUE #1056: Delete encrypted-media epoch secrets when retained source messages are removed
## Summary

Tie encrypted-media secret retention to the retained source messages that require those secrets. Deleting timeline rows while retaining their epoch secrets leaves content decryptable and violates the adopted retained-history boundary.

## Normative source

- `protocol-core/retained-history.md`

## Dependencies

- #1055 source-epoch retention decisions
- #1052 Media V2, where version-aware references are needed

## Start here

- `crates/storage-sqlite/src/encrypted_media_secrets.rs`
- `crates/storage-sqlite/src/migrations/0010_encrypted_media_epoch_secrets.rs`
- timeline deletion/retention sweep code in `crates/storage-sqlite/`

## Required behavior

- Track which retained source messages/media records require each stored media epoch secret.
- When retention removes the final retained reference, delete the corresponding secret in the same torn-write-safe operation (or via a durable compensating intent).
- Do not delete a secret while any retained message still needs it.
- Handle duplicate attachments, multiple messages from one epoch, partial sweeps, group deletion, and restart.
- Prevent an in-memory cache from continuing to serve a secret after durable deletion.
- Define conservative migration/garbage-collection behavior for pre-existing unreferenced secrets.
- Emit only aggregate privacy-safe audit/telemetry.

## Acceptance criteria

- Tests cover one-to-one, many-to-one, duplicate, partial, and final-reference deletion.
- Crash-point tests cannot leave a deleted message indefinitely decryptable through durable or hydrated secret state.
- A still-retained sibling message remains decryptable until its own deadline.
- Group deletion clears all associated media secrets.
- Cache invalidation is tested across same-process use and restart.
- `just fast-ci` and targeted storage/media tests pass.

agent-p1p:
## Summary

- track version-aware retained-message references for each encrypted-media exporter epoch
- retire key rows and persist anti-rehydration watermarks in the same SQLite transaction as final-reference pruning
- keep V1/V2 references coherent around their shared MLS exporter secret, including duplicate, partial-sweep, restart, and local-group-delete cases
- gate runtime hydration after durable retirement and recache only after a newly retained reference is durable
- preserve publish success when the post-publish cache retry fails, avoiding duplicate sends

## Dependency

Depends on #1068 (`#1055`) for source-epoch-pinned retention deadlines. Both branches currently add migration 0029, so this draft must be rebased and this migration renumbered after #1068 lands.

The V2 reference identity is recorded now; V2 runtime behavior remains owned by #1052.

## Test plan

- `cargo test -p storage-sqlite --locked` (250 passed)
- `cargo test -p cgka-traits --locked` (101 passed across unit/integration/snapshot suites)
- `cargo test -p marmot-app media --locked` (46 unit + 4 relay media tests passed)
- `cargo test -p marmot-app --test relay_runtime retained_media_rehydrates_a_retired_current_epoch_before_the_group_advances --locked -- --exact`
- `just fast-ci`

Fixes #1056

### DIFF
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -1446,10 +1446,31 @@ impl AppClient {
                 .iter()
                 .map(|attachment| attachment.reference.clone())
                 .collect();
-            result.sent = Some(
-                self.send_media_attachments(group_id, attachments, caption)
-                    .await?,
-            );
+            let summary = self
+                .send_media_attachments(group_id, attachments, caption)
+                .await?;
+            // The post-publish projection now durably references this source
+            // epoch. Persist again so a prior final-reference retirement cannot
+            // suppress the secret needed by the newly retained message.
+            if self
+                .remember_encrypted_media_epoch_secret(
+                    group_id,
+                    source_epoch,
+                    media_secret.as_ref(),
+                )
+                .is_err()
+            {
+                // Publication already succeeded. Do not report a false send
+                // failure that could make the caller publish a duplicate; the
+                // normal current-epoch cache pass can retry this durable write.
+                tracing::warn!(
+                    target: "marmot_app::media",
+                    method = "upload_media",
+                    error_code = "encrypted_media_secret_cache_skipped",
+                    "failed to cache encrypted media source epoch secret after publish",
+                );
+            }
+            result.sent = Some(summary);
         }
         Ok(result)
     }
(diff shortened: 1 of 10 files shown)


agent-p1p:
Closing this Pip-authored PR because it overlaps Jeff's independently controlled stacked PR series. The associated Pip task graph has been archived and its accidental agent-ok intake has been removed. Preserving the branch for audit/recovery; no further Pip work is authorized on this issue stack.

PR #1072, opened 2026-07-23
feat: implement current founding Add and Welcome lifecycle

erskingardner:
## Summary

- make current-profile group creation canonical locally at epoch 0 (solo) or epoch 1 (one founding Add), without publishing an ordinary founding commit
- persist and publish one distinct Welcome per invitee as independent, retryable delivery work; failed delivery cannot roll back or suppress the other Welcomes
- record founding Welcome intent before app-layer transport side effects and retain failed/unattempted work across restart
- preserve the explicit legacy pending-create path until the strict cutover PR
- extend engine, session, account/runtime, app, and portable conformance coverage through the invitee’s first application message

## Validation

- `just fast-ci`
- `cargo test -p cgka-engine -p cgka-session -p marmot-account -p cgka-conformance-simulator` (the three stale current-profile assertions found in the first run were updated; their targeted suite then passed)
- `cargo test -p cgka-engine --test update_group_data`
- `cargo test -p cgka-session -p marmot-account`
- `cargo test -p marmot-app`
- `cargo test -p cgka-conformance-simulator canonical_vector_fixtures_match_generated_traces -- --nocapture`

Fixes #1051

## Summary by CodeRabbit

* **New Features**
  * Current-profile founding group creation is now canonical immediately (no pending confirmation step).
  * Founding Welcome delivery is tracked as retained outbound work, supporting independent retry after failures or restarts.
  * Added APIs and runtime/app queries for outstanding and tracked Welcome deliveries.
* **Bug Fixes**
  * Fixed handling/auditing of founding-creation outcomes across engine/session/publishing so expectations and recipient classification match.
  * Improved preservation of already-stored retained Welcome payloads to prevent overwrites.
* **Documentation**
  * Clarified founding vs legacy creation and “delivery pending” semantics.
* **Tests**
  * Expanded conformance vectors and added restart/redelivery regression tests for current founding behavior.

Issue #1049, opened 2026-07-23
Cut over KeyPackage publication and joining strictly to the current profile


erskingardner:
## Summary

Implement the deliberately strict compatibility cutover for KeyPackages and Welcomes. This is the final behavior flip, not an additive preparatory change.

After cutover MDK publishes only current-profile KeyPackages and creates/joins only current-profile groups. Every unpublished legacy KeyPackage and its private `init_key` is retired. Cached or in-flight legacy Welcomes are rejected. Already-joined legacy groups keep working for their current membership, but cannot add or re-add members.

## Dependencies

- OpenMLS application-data/last-resort support (#1047)
- Explicit profile classification and current account proof (#1048)
- Current group invariants and Media V2 issues
- Required conformance/vector and binding work completed before merge

## Normative source

- `foundation/key-packages.md`
- `protocol-core/group-setup.md`
- `protocol-core/joining.md`
- `protocol-core/group-state.md`
- Product compatibility decision recorded in this issue

## Start here

- `crates/cgka-engine/src/key_package.rs`
- `crates/cgka-engine/src/group_lifecycle.rs`
- account/runtime KeyPackage publication paths in `crates/marmot-app/` and `crates/marmot-account/`
- KeyPackage persistence in `crates/storage-sqlite/`
- Nostr KeyPackage publication/deletion in transport/runtime code

## Required behavior

Before the final flip, additive PRs may generate/validate current artifacts behind explicit internal selection while preserving current production defaults. In the final cutover PR:

- Stop generating and publishing legacy KeyPackages.
- Generate/publish current-profile KeyPackages only.
- Enumerate every locally usable unpublished legacy KeyPackage, delete/retire its private `init_key`, and best-effort remove its published relay event(s).
- Make retirement restart-safe and idempotent; failure to delete a relay event must not preserve the private join capability.
- Reject any legacy Welcome, including one cached before upgrade or delivered in flight after upgrade.
- Permit Welcome join only when KeyPackage/profile/current group requirements match.
- Refuse creation of legacy groups and additions to legacy groups.
- Preserve messaging, group changes, and Media V1 behavior in already-joined legacy groups.
- Do not offer migration, upgrade, or explicit-rejoin UX for legacy groups.

## Acceptance criteria

- Upgrade tests seed legacy published/unpublished KeyPackages and private keys, run cutover twice, and prove no usable legacy `init_key` remains.
[...]

PR #1075, opened 2026-07-23
Implement frozen Media V2 while preserving legacy V1

### LINKED ISSUE #1052: Implement frozen Media V2 component 0x800b while preserving legacy Media V1

erskingardner:
## Summary

- add the distinct frozen `marmot.group.encrypted-media.v2` component at `0x800b` without reinterpreting V1 `0x8008`
- select V2 for current-profile app groups and preserve exact V1 behavior for already-joined legacy groups
- implement the adopted V2 MIME/filename/locator validation, KDF/AAD, fresh nonce, and attachment-local rejection rules
- isolate persisted epoch secrets by media component and expose typed V1/V2 values through UniFFI
- keep endpoint replacement on the group’s existing media version; this does not add a migration path

## Stack

- depends on #1072
- the strict default/profile cutover remains in #1049 and will stack on this PR

## Validation

- `just fast-ci`
- `cargo test -p cgka-engine -p storage-sqlite -p cgka-session -p marmot-account -p marmot-app -p marmot-uniffi -p cgka-conformance-simulator`
- `crates/marmot-uniffi/xcframework.sh` (Swift binding + device/simulator XCFramework)
- `crates/marmot-uniffi/kotlin-bindings.sh` (typed Kotlin binding + all four Android JNI ABIs)

Fixes #1052

## Summary by CodeRabbit

* **New Features**
  * Added encrypted-media v2 support for newly created groups, with version-aware policies, endpoints, uploads/downloads, and imeta handling.
  * Exposed protocol-profile helpers for generating fresh KeyPackages and group material.
  * Added UniFFI helpers to parse/build encrypted-media imeta tags and improved error reporting for invalid references.
* **Bug Fixes**
  * Rejects “frozen” encrypted-media v1 for current-profile validation.
  * Prevents mixing v1/v2 component data, policies, and media references.
* **Documentation**
  * Updated encrypted-media versioning, frozen vs current group endpoint behavior, and loopback endpoint guidance.


erskingardner (commented):
## Adversarial review

Cross-checked the diff against the full PR branch source (call sites, test semantics, and the actual dial path), not just the diff in isolation.

### Summary

This PR adds the distinct, frozen `marmot.group.encrypted-media.v2` component (`0x800b`) alongside the existing V1 (`0x8008`), wires the engine invariant forbidding V1 in current-profile groups, and threads a versioned `EncryptedMediaVersion` through key derivation, AAD, locator/host-safety validation, MIME/filename validation, and UniFFI. The cryptographic domain separation is solid and the destination-safety story is preserved for both versions. One significant problem: the PR silently changes ingest-time behavior for legacy V1 messages, contradicting its own stated goal of "preserve exact V1 behavior for already-joined legacy groups."

### 🛑 Correctness: legacy V1 ingest behavior silently changed

**`crates/marmot-app/src/groups.rs:1141-1146`** (`decode_received_event`)

The PR body states two distinct goals: "select V2 for current-profile app groups and preserve exact V1 behavior for already-joined legacy groups" and, separately, implementing the attachment-local rejection rules **as a V2 feature**. The old code dropped the entire message when its `imeta` media reference was structurally malformed — that was V1's actual, deliberately established behavior (see the old test name this PR renames, `received_media_message_with_malformed_reference_is_rejected`, citing "PR #328 review Finding 2").

This PR removes the `media_imeta_tags_are_valid` gate from `decode_received_event` **entirely and unconditionally**, rather than gating it on the encrypted-media version. Result: a V1-tagged message (`v encrypted-media-v1`) with a structurally malformed attachment (e.g. a corrupt `ciphertext_sha256`) is now **kept** rather than dropped. The renamed test at `crates/marmot-app/src/tests.rs:2156-2202` (`received_media_message_with_malformed_reference_keeps_the_message`) confirms this: it builds its attachment with `version: ENCRYPTED_MEDIA_VERSION.to_owned()` (still bound to `encrypted-media-v1`, see `crates/marmot-app/src/media/mod.rs:47`) and now asserts `.is_some()`.

So the "attachment-local" rule the new code comment attributes to encrypted-media-v2 is in fact applied to V1 traffic too — the opposite of "preserve exact V1 behavior." Legacy client/receiver pairs relying on structural malformation causing message rejection (moderation/audit assumptions, or simply matching legacy wire semantics for interop) will diverge after upgrading.

**Suggested fix:** reintroduce a version-aware branch in `decode_received_event` — keep "drop the whole message" for `encrypted-media-v1` references and apply attachment-local semantics only for `encrypted-media-v2` (or unrecognized versions, per whatever the adopted spec says about mixed-version messages).

**Evidence this was an oversight, not a deliberate simplification:** the signature at `crates/marmot-app/src/groups.rs:1111` renames its last parameter to `_allow_loopback_http` (underscore-prefixed, now dead) rather than removing it. That parameter only existed to feed the removed `media_imeta_tags_are_valid` call, and it's still threaded through 5+ call sites. Leaving it dead-but-renamed strongly suggests the version-aware gate was meant to stay but got dropped instead of narrowed.

### Suggestions (non-blocking)

- **`crates/marmot-uniffi/src/conversions/media.rs:83-85`** — `EncryptedMediaVersion::parse(&value.version).expect("validated media reference carries a supported version")` panics if ever reached with an unvalidated `MediaAttachmentReference`. Every current construction path validates first, so it's unreachable today, but an `.expect()` on data originating from a remote peer's tags is a fragile cross-crate invariant. Consider `Result`/`TryFrom`, or document explicitly which call sites guarantee the precondition.

- **`crates/traits/src/app_components/mod.rs:79-83, 97-99`** — the backward-compat aliases `GROUP_ENCRYPTED_MEDIA_COMPONENT_ID`/`GROUP_ENCRYPTED_MEDIA_COMPONENT` silently mean "V1" and are still re-exported unadorned from `crates/marmot-app/src/lib.rs:26-29`. Since this PR's whole point is that "V1" and "the encrypted media component" are no longer synonyms, consider `#[deprecated(note = "use the explicit V1/V2 constants")]` so remaining/future callers get a compile-time nudge.

- **`crates/marmot-app/src/groups.rs:1111`** — once version-gating is restored the parameter naming sorts itself out; if the universal attachment-local behavior for V1 turns out to be intentional (i.e. the PR description is just imprecise), delete the dead `_allow_loopback_http` parameter outright and update the stale comment in `observe_event` (`groups.rs:1216-1223`), which still says "Media references are validated structurally only inside `decode_received_event`."

### Verified non-issues

- **V2's relaxed structural locator validation is not an SSRF regression.** `validate_locator` in `crates/marmot-app/src/media/host_safety.rs` only checks scheme for V2, but the actual network path (`media_http_client_for_url`/`resolve_media_host` in `crates/marmot-app/src/media/blossom.rs`) still enforces `reject_non_public_ip` and DNS-pinning unconditionally for both versions before dialing. This is a deliberate "structural validity vs. destination policy" split.
- Cryptographic domain separation is done right: `media_key_info`/`media_aad` fold the version string into HKDF info and AEAD AAD, with a dedicated round-trip test (`encrypted_media_component_ids_do_not_reinterpret_each_others_bytes`) proving V1 bytes are rejected under the V2 component id and vice versa.
- The current-profile invariant (`validate_current_profile_group_context`) correctly rejects V1 both as a required component and as raw dictionary state, with direct tests.
- Epoch-secret storage isolation by component id is directly tested, and the V2 codec has golden-byte and canonical-normalization tests mirroring V1's rigor.

erskingardner:
@coderabbitai full review

PR #1078, opened 2026-07-23
Authorize standalone proposals before queueing

erskingardner:
## Summary

- authorize authenticated standalone proposals against their source-epoch Marmot profile, membership, admin policy, capabilities, component rules, and account identity proofs before durable pending storage
- expose stable typed rejection categories (`AuthorizationFailed`, `UnsupportedProposal`, `InvalidEncoding`, `InvalidSignature`, and `InvalidSelfRemove`) while keeping parent-dependent proposals eligible for retained-fork convergence
- re-run the same proposal policy for staged commits, delayed SelfRemove work, retained-history replay, candidate probing, and hydration so rejected proposals cannot affect a later commit
- preserve deployed legacy Update behavior while enforcing current-profile proposal roles, and emit canonical privacy-safe rejection categories
- add adversarial engine/restart coverage plus an executable conformance vector for raw OpenMLS and Nostr-wrapped inbound seams

## Stack

This PR is based on #1075 (`codex/issue-1052-media-v2`). Merge the stack in order.

## Validation

- `just fast-ci`
- `cargo test -p cgka-engine`
- `cargo test -p cgka-conformance-simulator`
- `cargo test -p cgka-engine --test mip03_guards`
- `cargo test -p cgka-conformance-simulator --test openmls_replay_probe proposal_authorization_vector_rejects_direct_and_nostr_wrapped_input`

Fixes #1053

PR #1076, opened 2026-07-23
Complete the current-profile compatibility cutover

erskingardner:
## Summary

This is now the single compatibility-release integration PR. It combines the remaining protocol-cutover work into one review and CI target:

- enforce the strict current-profile cutover for new key packages and groups while preserving existing legacy groups
- retire and replace cached legacy key packages without publishing new legacy packages
- authorize standalone proposals and staged commits consistently at ingest and replay boundaries
- pin retention decisions to the source epoch and preserve durable pruning semantics
- integrate with the Media V2 and checked UniFFI `imeta` work merged through #1075 / #1080
- reconcile the combined app-runtime behavior, including current-profile relay key-package classification and stable replaceable-event `d` slots

## Consolidation

- Targets `master` directly now that #1075 is merged
- Incorporates all commits from #1078 and #1079
- #1078 and #1079 are superseded by this PR
- #1080 was already merged through #1075 and is included in this combined integration state

## Validation

- `just fast-ci`
- `cargo test --no-fail-fast -p cgka-engine -p storage-sqlite -p cgka-session -p marmot-account -p marmot-app -p marmot-uniffi -p cgka-conformance-simulator`
- `cargo test -p marmot-app`
- `cargo test -p marmot-app --test relay_runtime`

Fixes #1049
Fixes #1053
Fixes #1055

## Summary by CodeRabbit

* **New Features**
  * Added message retention metadata (source epoch/expiry), structured retention decisions, and retention-aware secure pruning for expired content.
  * Added structured publication details for accepted application messages (group/app identifiers, source epoch, retention) and persisted published-app audit records.
  * Added proposal rejection categorization with richer rejection auditing.
  * Added strict cutover support, including legacy KeyPackage retirement/replacement handling and stricter legacy-profile gating.
* **Bug Fixes**
  * Improved replay/convergence and ingest outcome handling for rejected proposals, dropped-message categorization, and retention propagation into emitted events.
* **Tests**
  * Updated tests/fixtures to reflect new retention fields and expanded send-result metadata.

PR #1079, opened 2026-07-23
feat(retention): pin source-epoch decisions

### LINKED ISSUE #1055: Pin message retention decisions to each message's source epoch

erskingardner:
## Summary

- resolve each accepted application message retention from its authenticated MLS source epoch, including delayed and convergence-replayed traffic
- carry the exact app-event id, group, source epoch, retention decision, and adapter-visible transport id through engine, session, and account publication so optimistic local rows cannot be finalized from later group state
- persist one normalized decision and exact deadline transactionally with the app event; duplicates and restart replay cannot rewrite it
- secure-prune only stored expired decisions while retaining disabled, overflow, legacy, and unrecoverable-policy rows safely
- expose source epoch and retention decisions through Rust and UniFFI message records

## Stack

- Base: #1078
- Implements and closes #1055

## Validation

- `just fast-ci`
- `cargo test -p storage-sqlite` (244 passed)
- `cargo test -p cgka-engine --test update_group_data delayed_app_messages_pin_retention_from_their_source_epoch`
- `cargo test -p cgka-session --test session_lifecycle session_advance_convergence_releases_queued_outbound_work`
- `cargo test -p marmot-account --test runtime published_app_messages_carry_exact_source_state_and_adapter_identity`
- `cargo test -p marmot-app --lib`


erskingardner:
Superseded by #1076. The complete #1079 commit and its #1078 dependency are merged into that branch, so review and CI should continue on #1076.

PR #1080, opened 2026-07-23
Expose checked Media V1/V2 imeta APIs

### LINKED ISSUE #955: Expose versioned Media V1/V2 build and parse APIs through UniFFI
## Summary

Replace the V1-only `imeta` FFI gap with an explicit versioned media boundary that can preserve legacy Media V1 and expose frozen Media V2 without lossy conversion.

This is the bindings child of #1052. Existing legacy groups still need Media V1. Current groups require Media V2.

## Dependencies

- #1052 defines/implements the Rust Media V2 model and profile dispatch
- #1048 supplies explicit group profile classification

## Start here

- `crates/marmot-uniffi/src/commands/media.rs`
- `crates/marmot-uniffi/src/conversions/media.rs`
- generated Swift/Kotlin binding scripts and smoke tests
- CLI media commands for parity

## Required behavior

- Provide versioned FFI types/results for Media V1 (`0x8008`) and Media V2 (`0x800b`). Do not present both as one ambiguous `imeta` string model.
- Building metadata takes or derives the target group profile and emits only that profile's media version.
- Parsing returns the actual wire version plus all fields needed for exact round-trip/display/download.
- Preserve absent versus empty optional values where the wire distinguishes them.
- Surface canonicalization/validation errors as typed binding errors rather than panics or silent normalization.
- Existing app calls for legacy groups remain source-compatible where practical, but must not cause current groups to emit V1.
- Update checked-in/generated Swift and Kotlin artifacts in the same final binding PR.

## Acceptance criteria

- UniFFI round-trip tests cover golden V1 and V2 fixtures independently.
- Calling the V1 builder for a current group and V2 builder for a legacy group fails clearly (or is impossible by API design).
- Swift and Kotlin smoke builds exercise parse/build/error paths for both versions.
- No conversion drops V2-only fields or collapses absent/present-empty state.
- CLI/app-core and bindings agree on validation results for shared fixtures.
- `just fast-ci`, targeted UniFFI tests, and actual Swift/Kotlin binding builds pass.

erskingardner:
## Summary

Completes the public binding boundary for encrypted Media V1/V2 on top of #1075.

- add an async, group-aware `build_media_imeta_tag` API that derives the target group's actual media version and locator policy before emitting a tag
- add a strict `parse_media_imeta_tag` FFI API that returns the typed wire version and complete download/display reference
- return `InvalidMediaReference` as a typed Swift/Kotlin error for malformed, noncanonical, or cross-profile input
- preserve present-empty optional `dim`/`thumbhash` fields instead of collapsing them to absence
- remove the CLI's duplicate hand-written `imeta` parser and use the same Rust parser and loopback policy as the runtime
- bump the workspace to 0.9.6 for the public UniFFI surface change

This is additive for existing media send/upload/download/list callers. Legacy groups continue to build V1 references; current groups build V2 references. A reference from the wrong version is rejected before optimistic metadata is produced.

## Normative sources

- `features/encrypted-media-v1.md`
- `app-components/group-encrypted-media-v1.md`

## Stack

- Base: #1075 (`codex/issue-1052-media-v2`)
- This PR does not depend on #1078 or #1079.

## Verification

- `cargo test -p marmot-uniffi` (63 unit + 22 smoke)
- `cargo test -p marmot-app` (373 unit plus all integration suites, including 70 relay-runtime tests)
- `cargo test -p wn-cli` (375 unit + 86 end-to-end)
- `just fast-ci`
  - host dylib, iOS device, iOS simulator, generated Swift, and XCFramework succeeded
  - generated Kotlin plus arm64-v8a, armeabi-v7a, x86, and x86_64 JNI libraries succeeded

Generated binding artifacts remain ignored per `crates/marmot-uniffi/AGENTS.md`; both generated surfaces were inspected for the builder, parser, and typed error.

Closes #955

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agent-connector"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "agent-control",
  "agent-stream-compose",
@@ -72,7 +72,7 @@ dependencies = [
 
 [[package]]
 name = "agent-control"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "serde",
  "serde_json",
@@ -82,7 +82,7 @@ dependencies = [
 
 [[package]]
 name = "agent-stream-compose"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "cgka-traits",
  "hex",
@@ -852,7 +852,7 @@ checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
 name = "cgka-conformance-simulator"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "async-trait",
  "cgka-engine",
@@ -880,7 +880,7 @@ dependencies = [
 
 [[package]]
 name = "cgka-engine"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "async-trait",
  "cgka-traits",
@@ -909,7 +909,7 @@ dependencies = [
 
 [[package]]
 name = "cgka-session"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "async-trait",
  "cgka-engine",
@@ -931,7 +931,7 @@ dependencies = [
 
 [[package]]
 name = "cgka-traits"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "async-trait",
  "hex",
@@ -1843,7 +1843,7 @@ dependencies = [
 
 [[package]]
 name = "fs-private"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "libc",
  "tempfile",
@@ -2572,7 +2572,7 @@ dependencies = [
 
 [[package]]
 name = "incident-replay"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "cgka-conformance-simulator",
  "serde",
@@ -3147,7 +3147,7 @@ dependencies = [
 
 [[package]]
 name = "marmot-account"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "android-native-keyring-store",
  "apple-native-keyring-store",
@@ -3176,7 +3176,7 @@ dependencies = [
 
 [[package]]
 name = "marmot-app"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "async-trait",
  "cgka-engine",
@@ -3217,7 +3217,7 @@ dependencies = [
 
 [[package]]
 name = "marmot-forensics"
-version = "0.9.5"
+version = "0.9.6"
 dependencies = [
  "fs-private",
  "serde",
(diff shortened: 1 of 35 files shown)


agent-p1p (changes requested):
## Adversarial review — changes required

Reviewed exact head `3a6d23931d1e05e1aa694ac25e3cb38dabb01a48` against stacked base `dfe4261ade2ce7b0806b36bd4c376d66d85a0309`.

### Blocking

1. **`crates/cli/src/commands/media.rs:257-266` — one malformed attachment aborts the entire CLI media operation.**

   `media_attachments_from_message` maps every `imeta` tag into a `Result` and then `collect()` short-circuits on the first invalid tag. Consequently `media list` emits no valid siblings, and `media download` can fail before finding a later valid reference. This conflicts with the adopted V2 rule that rejection is attachment-local and with the binding path in `crates/marmot-uniffi/src/conversions/media.rs:234-240`, which correctly uses `filter_map` to retain valid siblings. Use attachment-local filtering for the CLI too and add a valid+malformed sibling regression test.

2. **`crates/marmot-uniffi/src/commands/media.rs:51-114` — the typed media error is not applied to the existing media APIs.**

   `send_media_attachments`, `upload_media`, and `download_media` still use bare `?`. `AppError::InvalidEncryptedMedia` / `InvalidAppMessagePayload` therefore fall through `From<AppError>` to `MarmotKitError::Runtime` (`errors.rs:191-193`). For example, sending a V1 reference to a current group does not surface `InvalidMediaReference`, even though the new builder reports the same profile mismatch with that typed variant. Apply the media-specific mapper to all media-boundary methods and test wrong-version, malformed upload, and malformed download inputs as `InvalidMediaReference`.

3. **`crates/marmot-app/src/media/mod.rs:167-175,222-229` — the checked V1 builder can emit a noncanonical `m` field unchanged.**

   V2 compares the canonical media type to the supplied bytes, but V1 only calls `canonical_media_type_v1` and discards its result. A legacy-group FFI reference with `media_type = "Image/JPG; charset=utf-8"` therefore passes `build_imeta_tag` and is emitted verbatim, although the frozen V1 sender rule requires the stored `m` value to be the canonical result (`image/jpeg`). Keep legacy ingest compatibility, but make the outbound checked builder reject noncanonical V1 input with `InvalidMediaReference` (or emit from an explicitly canonical typed input) and add the boundary test.

### Suggestion / missing acceptance coverage

- `crates/marmot-uniffi/tests/smoke.rs` calls the Rust facade directly; the builder is only exercised with malformed group hex. `xcframework.sh` and `kotlin-bindings.sh` generate source and package native libraries, but neither compiles/runs a Swift or Kotlin consumer or exercises V1/V2 parse/build/error paths. Add the issue-required host-language smoke builds, including successful group-derived V1 and V2 builds, both cross-profile rejections, present-empty optionals, and typed errors. I independently generated both binding sources and confirmed the expected signatures exist, but that is not the requested consumer smoke build.

### Verification

- `cargo test -p marmot-uniffi` — 63 unit + 22 smoke passed.
- Focused `marmot-app` builder and relay upload/build mismatch tests passed.
- `just fast-ci` passed.
- `cargo test -p wn-cli` — 375 unit tests passed; integration run was 85/86 with one unrelated `groups_leave_publishes_self_remove` convergence timeout, and that exact test passed on immediate rerun.
- Live GitHub checks are green at the reviewed head.

erskingardner:
@coderabbitai full review

Issue #1057, opened 2026-07-23
Implement NIP-65 relay marker semantics and durable prior group routes


erskingardner:
## Summary

Align relay discovery and group routing with the adopted Nostr transport. Interpret NIP-65 markers correctly and retain prior group routes long enough to ingest delayed/offline events after routing changes.

This issue is about route derivation/history. Durable outbound fanout and acknowledgment lifecycle remain in #1020.

## Normative source

Adopted by marmot-protocol/marmot#408:

- `transports/nostr.md`
- `app-components/nostr-routing-v1.md`
- `protocol-core/inbound-processing.md`

## Start here

- `crates/transport-nostr-adapter/src/sdk_client.rs`
- routing/subscription code in `crates/marmot-app/src/client/sync.rs`
- group projection/storage in `crates/storage-sqlite/`
- Nostr transport tests in `crates/marmot-app/tests/relay_runtime.rs`

## Required behavior

- Interpret kind-10002 NIP-65 relay-list markers with their specified read/write meaning; do not treat every URL as equivalent for both directions.
- Derive the current group read/write route from authenticated group state plus applicable account relay discovery.
- Persist prior authenticated group route sets when the group route changes.
- Subscribe/query prior read routes needed to process delayed/offline messages and define a bounded, deterministic retirement rule consistent with the spec.
- A restart must restore current and still-live prior routes before sync reports ready.
- Deduplicate events observed on current/prior/multiple relays by authenticated event identity without discarding a valid later delivery.
- Apply host-safety dial validation to every resolved outbound relay address; this is local safety policy, not group-signed policy.

## Out of scope

- Exact kind-445 event tag validation (#1007).
- Durable outbound fanout/ack state (#1020).
- A spec-level or group-signed broker/relay allowlist.

## Acceptance criteria

- Marker matrix tests cover unmarked, `read`, `write`, invalid, and duplicate relay entries.
- A route-change scenario accepts a delayed pre-change event from the prior route after restart.
- Current and prior subscriptions do not double-apply the same event.
- Retirement is deterministic and does not depend only on process uptime.
- Dial-safety tests cover DNS resolution/rebinding boundaries.
- `just fast-ci` and targeted transport/app/storage tests pass.

PR #1082, opened 2026-07-23
Honor NIP-65 relay roles for KeyPackage routing

erskingardner:
Part of #1057.

This separates NIP-65 marker semantics from the larger durable prior-route change:

- parses kind 10002 relay tags into distinct read- and write-capable sets
- treats unmarked entries as both read and write
- ignores malformed and unknown-marker entries
- merges duplicate relay roles deterministically
- uses only the write-capable set for Marmot KeyPackage fetch and publication routing

The existing account relay-list publication API emits unmarked entries, which remain valid for both directions.

Validation:
- cargo test -p transport-nostr-adapter
- cargo test -p marmot-app
- just fast-ci

### DIFF
--- a/crates/transport-nostr-adapter/src/relay_list.rs
+++ b/crates/transport-nostr-adapter/src/relay_list.rs
@@ -7,6 +7,65 @@ pub const KIND_MARMOT_INBOX_RELAY_LIST: u64 = 10_050;
 const NIP65_RELAY_TAG: &str = "r";
 const MARMOT_RELAY_TAG: &str = "relay";
 
+/// Directional relay sets recovered from a NIP-65 kind-10002 event.
+///
+/// An unmarked `r` tag belongs to both sets. A `read`-marked tag belongs only
+/// to `read_relays`, and a `write`-marked tag belongs only to `write_relays`.
+/// Malformed tags and unknown markers are ignored. Repeated relay URLs are
+/// deduplicated independently in each direction while preserving first-seen
+/// order.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct NostrNip65RelaySet {
+    pub read_relays: Vec<TransportEndpoint>,
+    pub write_relays: Vec<TransportEndpoint>,
+}
+
+/// Parse the directional relay roles from a NIP-65 kind-10002 event.
+///
+/// This is intentionally a tolerant list parser: a hostile or future
+/// extension tag must not turn a different valid `r` entry into a relay
+/// target. Exact two-element tags are unmarked (read and write); exact
+/// three-element tags accept only the NIP-65 `read` and `write` markers.
+pub fn parse_nip65_relay_set(event: &NostrTransportEvent) -> NostrNip65RelaySet {
+    if event.kind != KIND_NIP65_RELAY_LIST {
+        return NostrNip65RelaySet::default();
+    }
+
+    let mut relays = NostrNip65RelaySet::default();
+    for tag in &event.tags {
+        let (relay, read, write) = match tag.as_slice() {
+            [name, relay] if name == NIP65_RELAY_TAG && !relay.trim().is_empty() => {
+                (relay, true, true)
+            }
+            [name, relay, marker]
+                if name == NIP65_RELAY_TAG && !relay.trim().is_empty() && marker == "read" =>
+            {
+                (relay, true, false)
+            }
+            [name, relay, marker]
+                if name == NIP65_RELAY_TAG && !relay.trim().is_empty() && marker == "write" =>
+            {
+                (relay, false, true)
+            }
+            _ => continue,
+        };
+
+        if read {
+            push_unique_endpoint(&mut relays.read_relays, relay);
+        }
+        if write {
+            push_unique_endpoint(&mut relays.write_relays, relay);
+        }
+    }
+    relays
+}
+
+fn push_unique_endpoint(endpoints: &mut Vec<TransportEndpoint>, relay: &str) {
+    if !endpoints.iter().any(|endpoint| endpoint.0 == relay) {
+        endpoints.push(TransportEndpoint(relay.to_owned()));
+    }
+}
+
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum NostrAccountRelayListKind {
     Nip65,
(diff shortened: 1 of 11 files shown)


crates/transport-nostr-adapter/src/relay_list.rs
```diff
@@ -7,6 +7,65 @@ pub const KIND_MARMOT_INBOX_RELAY_LIST: u64 = 10_050;
...
+                if name == NIP65_RELAY_TAG && !relay.trim().is_empty() && marker == "read" =>
+            {
+                (relay, true, false)
+            }
+            [name, relay, marker]
+                if name == NIP65_RELAY_TAG && !relay.trim().is_empty() && marker == "write" =>
+            {
+                (relay, false, true)
+            }
+            _ => continue,
```
erskingardner, line 52: **Exact-arity matching is stricter than the stated rationale needs, and its failure mode is total.** NIP-01 tags are open-ended arrays ("elements after the second do not have a conventional name") and the ecosystem grows tags by appending — `e` gained a relay hint, then a pubkey. Here `["r", url, "write", <anything>]` falls to `_ => continue` and the entry vanishes (the marker-matrix test pins this in). The hostile-extension property you document only requires rejecting *unknown markers in position 3*; a valid `read`/`write` marker followed by extra elements carries no ambiguity.

Consequence: a peer whose client appends one trailing element to its r-tags parses as having **no** NIP-65 → `MissingRelayLists(Nip65)` → their KeyPackages can't be fetched and they can't be invited.

Suggest matching `[name, relay, rest @ ..]` and dispatching on `rest.first()`: `None` → unmarked, `Some("read")`/`Some("write")` → directional regardless of trailing extras, anything else → skip. Unknown markers stay excluded, preserving the hostile-extension property.

Same class, smaller: markers are matched case-sensitively, so `["r", url, "READ"]` is dropped entirely rather than degraded. Lowercase-only is defensible from the spec text — just flagging that the penalty for a sloppy writer is again total loss of the entry.

crates/transport-nostr-adapter/src/relay_list.rs
```diff
@@ -7,6 +7,65 @@ pub const KIND_MARMOT_INBOX_RELAY_LIST: u64 = 10_050;
...
+        }
+        if write {
+            push_unique_endpoint(&mut relays.write_relays, relay);
+        }
+    }
+    relays
+}
+
+fn push_unique_endpoint(endpoints: &mut Vec<TransportEndpoint>, relay: &str) {
+    if !endpoints.iter().any(|endpoint| endpoint.0 == relay) {
```
erskingardner, line 67: Nit: the guard trims but the stored value doesn't — `["r", " wss://relay.example"]` passes `!relay.trim().is_empty()` and is pushed untrimmed, and dedup here compares raw strings, so `" wss://x"` and `"wss://x"` count as two endpoints and both get dialed. Matches the pre-existing inbox-path behavior, but since this parser is new it could store `relay.trim()` and close the gap here.

erskingardner (commented):
Adversarial review, taken as the stacked delta against `codex/issue-1007-kind-445-shape` (#1081).

The direction semantics themselves check out: NIP-65 says to use a user's **write** relays when downloading events *from* them and when publishing their own events, and every consumer I traced (`key_package_endpoints`, `member_key_package`, the setup-preflight profile refresh, both directory fetch paths) wants exactly outbox semantics. Parser tests pass locally (`cargo test -p transport-nostr-adapter`, 23 passed; the new marmot-app test passes too). Findings below, worst first.

### 1. Read-modify-write republish now deletes read-only relays from the user's kind 10002 (data loss)

`update_relay_list` in `crates/cli/src/commands/relays.rs` implements `wn relays add/remove --type nip65` as: fetch current status → `relays_for_type(status, "nip65")` → mutate → `publish_account_relay_list_kind`, which emits every entry **unmarked**. After this PR the read step returns only the write-capable subset, so a single `wn relays add --type nip65 wss://new` republishes the replaceable kind-10002 **without** any `read`-marked entries another client put there — they are permanently dropped from the canonical list, silently. The same trap exists for any MarmotKit client doing `account_nip65_relays()` → edit → `set_account_nip65_relays()`.

Before this PR the RMW cycle was lossy in a milder way (markers flattened to unmarked); it never removed a relay outright. Suggested direction: expose the full parsed `NostrNip65RelaySet` (or read+write on `AccountRelayListState`) so RMW flows can round-trip all entries — ideally preserving markers in publication — or make the CLI/set path operate on the raw event tags rather than the direction-filtered view.

### 2. A newer all-read-only NIP-65 event is silently ignored, pinning stale write targets

`[["r", url, "read"], …]` is a valid NIP-65 list and now parses to an **empty** write set. Two pre-existing "skip empty" guards misfire on that:

- `relay_list_status_from_records` (`crates/marmot-app/src/key_package_records.rs`, the `if relays.is_empty() { continue; }`) — records are sorted ascending, later records overwrite, so skipping the newest event means an **older** event's write set wins.
- `remember_directory_relay_list_event` (`crates/marmot-app/src/directory/methods.rs`) — early-returns on empty, so the directory cache keeps the old write set indefinitely.

Concretely: a peer migrates relays and demotes the old one to `read` (or publishes an all-read list). mdk keeps fetching and publishing their KeyPackages at the old relays forever instead of refreshing or failing closed with `MissingRelayLists`. Before this PR, an empty parse implied the event had no relay tags at all, so skipping was safe; now empty is a meaningful state ("declares no write relays") that should replace/clear the cached set. Deserves a marmot-app-level test either way.

### 3. Interop strictness in the parser (inline comments)

Exact-arity matching drops `["r", url, "write", <anything>]` entirely, and marker matching is case-sensitive — both fail total (the peer parses as having *no* NIP-65 → `MissingRelayLists` → cannot be fetched or invited). Details inline on `relay_list.rs`.

### Notes

- `account_nip65_relays()` / `wn relays list --type nip65` now display only the write subset with nothing in the payload indicating read-only entries exist; users comparing against another client will see relays "missing." Falls out naturally if #1 is fixed by exposing both sets.
- Directory caches persisted before this change still hold read-only relays inside `nip65.relays` and will be used as fetch/publish targets until the next refresh — self-healing, just noting.
- Duplicate-URL union semantics (an unmarked dup makes a URL write-capable even if also listed `read`) are deterministic and documented — fine as is.

Issue #1059, opened 2026-07-23
Move push owner proofs to event kind 451 and enforce current-profile push state


erskingardner:
## Summary

Finish the Amber-compatible push owner-proof design using an event-shaped Nostr signature, with a dedicated registry kind.

Current MDK already constructs an event-shaped owner proof for external signers, but uses kind 450 and accepts a raw-signature fallback. The registry decision is:

- kind 450: account identity proof v2
- kind 451: push notification owner proof
- kind 452: multi-device authorization

The merged push document still describes a raw BIP340 signature over `SHA256(SignedRecord)` and must be corrected upstream before this issue is merged. The raw record digest remains useful as a deterministic ordering tie-breaker; it is not the current-profile signature preimage.

## Normative source / prerequisite

- `features/push-notifications.md` after the follow-up Marmot spec correction
- `foundation/registries.md` after registering kinds 451 and 452
- `foundation/authorization-proofs.md`
- Existing implementation history: MDK PR #755

## Start here

- `crates/marmot-app/src/notifications.rs`
- `crates/marmot-app/src/notifications/tests.rs`
- `crates/storage-sqlite/src/migrations/0021_push_token_owner_signatures.rs`
- external signer and UniFFI signer surfaces

## Required behavior

- Current-profile groups produce and accept only the exact event-shaped kind-451 owner proof.
- Legacy groups may accept both the deployed raw legacy proof and the event-shaped kind-451 proof so existing members can continue operating.
- Never accept kind 450 or 452 as a push owner proof.
- Implement exact tags/content/timestamp binding from the corrected spec and verify through the external signer path.
- Enforce future `owner_ts` bounds and deterministic winner ordering exactly; use the raw SignedRecord digest only where the state-ordering rule calls for it.
- Apply accepted add/remove records per entry and clean state scoped to the exact member leaf/device identity defined by the spec; do not erase unrelated devices.
- Enforce the maximum 32 push triggers per encrypted kind-446 server chunk.
- The Transponder server does not receive `owner_sig` and requires no signature-verification change for this issue.

## Dependencies / related issues

- Upstream Marmot correction for kinds 451/452 and push proof shape
- #1048 profile classification
- #950 max-32 add/remove arrays and per-entry advisory processing
- #929 durable per-group push gossip retry/outcome

## Out of scope

- Implementing multi-device authorization component `0x800a`; kind 452 is only reserved here.
- Changing Transponder's push-owner verification (there is none).

## Acceptance criteria

[...]

PR #1084, opened 2026-07-23
fix(push): bound advisory gossip arrays

### LINKED ISSUE #950: Enforce max-32 push add/remove arrays before per-entry advisory processing
## Summary

Implement the adopted push-gossip array limits and processing order. Push token add/remove arrays are bounded to 32 entries. An oversized array makes the entire advisory message invalid and must be rejected before signature verification or per-entry work.

Within a correctly bounded array, individual malformed/invalid records do not poison valid sibling records: process each entry independently according to advisory state rules.

## Normative source

- `features/push-notifications.md` from marmot-protocol/marmot#408
- Follow-up spec correction for event-shaped kind-451 owner proofs (#1059 prerequisite)

## Start here

- `crates/marmot-app/src/notifications.rs`
- `crates/marmot-app/src/notifications/tests.rs`
- push projection/storage in `crates/storage-sqlite/`

## Required behavior

- Reject an add array longer than 32 or a remove array longer than 32 before parsing/verifying any entry in that message.
- Check declared/container lengths before proportional allocation or cryptographic work.
- For arrays of length 0..=32, parse, authenticate, order, and apply each record independently.
- One invalid entry is ignored/reported without discarding valid sibling entries, unless a message-level structural rule is violated.
- Deduplicate repeated identical entries without repeated signature work or state churn.
- Preserve deterministic last-writer/winner semantics regardless of input order.
- Bound audit/outcome data to aggregate counts and reason classes; never log tokens, keys, signatures, or group/member identifiers.
- Use profile-specific owner-proof acceptance from #1059.

## Acceptance criteria

- Tests cover lengths 0, 1, 31, 32, 33, and maliciously large declared lengths for both add/remove arrays.
- Instrumented tests prove a 33-entry array performs zero owner-signature verifications and applies zero records.
- Mixed valid/invalid arrays of length <=32 apply exactly the valid entries.
- Permutation and duplicate tests produce identical final state.
- Fuzz/property tests keep CPU/allocation bounded for malformed lengths.
- `just fast-ci` and targeted notification/storage tests pass.

agent-p1p:
## Summary

- preflight kind 447/448 `tokens` and kind 449 `removals` with a non-allocating sequence counter and reject entry 33 before typed decoding or owner-proof verification
- decode bounded entries independently so malformed records do not discard valid siblings
- deduplicate equivalent decoded entries before owner authentication while preserving storage-backed `(owner_ts, record_digest)` winner semantics
- keep outcomes privacy-safe: message-level errors expose only the existing generic reason class, and invalid entries are dropped without logging record data

## Verification

- `cargo test -p marmot-app --locked`
- `cargo test -p storage-sqlite --locked push_token`
- `cargo clippy -p marmot-app --all-targets --locked -- -D warnings`
- `just fast-ci`
- exact locked workspace `cargo check` and `cargo clippy` commands, default and all-features

Tests cover 0, 1, 31, 32, 33, and 100,000-entry arrays for both add/remove shapes; instrumented owner-proof counters prove oversized arrays perform zero signature verification and yield zero records. Mixed malformed/valid, duplicate, and permutation cases verify independent processing and deterministic final state.

## Sensitive scope

`crates/marmot-app/src/notifications.rs` reconstructs and verifies push owner proofs. This change does not alter proof acceptance rules; it places the array gate and deduplication before the existing verifier. Current/legacy profile-specific acceptance remains owned by #1059 and can consume the same bounded action.

Fixes #950

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved push-token gossip handling with bounded, preflight decoding to reject oversized `tokens`/`removals` arrays before deeper processing.
  * Malformed entries no longer interfere with valid entries in the same payload.
  * Centralized relay-hint canonicalization (trim/blank-to-absent) for consistent signing, verification, parsing, and storage.
  * Avoided unnecessary owner-signature verification for invalid/duplicate data; ensured deterministic record selection across entry order.
* **Tests**
  * Added/extended coverage for array-size boundaries, bounded failure behavior, deduplication, relay-hint normalization, and verification-invocation behavior.


agent-p1p:
Adversarial review at `12cf04322ad5d5962ce6d922ebeec344cd6f8fbf`.

BLOCKING:
- Semantic duplicates bypass the pre-verification dedup: `crates/marmot-app/src/notifications.rs:1065-1071`, `crates/marmot-app/src/notifications.rs:1160-1185`, and `crates/marmot-app/src/notifications.rs:662-710`. The `HashSet` sees the raw typed entry before `into_record` applies the spec's normalization that a whitespace-only `relay_hint` is absent. A relayer can therefore include the same valid signed token record twice—one entry with `relay_hint` omitted and one with `relay_hint: " \t"`. They hash differently, then both normalize to `relay_hint = None`, reconstruct the same owner-proof event, verify the same `owner_sig`, and reach storage as the same record. That repeats signature work and opens a second storage transaction/apply attempt, contrary to the issue's explicit “deduplicate ... without repeated signature work or state churn” requirement. Normalize before inserting into `seen`, or deduplicate an appropriate canonical record representation before verification. Add a regression using one signed no-hint record in both wire forms and assert exactly one owner-signature verification and one effective apply.

SUGGESTION:
- Avoid materializing bounded entries as `serde_json::Value`: `crates/marmot-app/src/notifications.rs:324-390` and `crates/marmot-app/src/notifications.rs:1052-1109`. Valid payloads are structurally scanned, reparsed into `Vec<Value>`, then decoded entry-by-entry; `Value` also retains arbitrarily large unknown entry members that typed deserialization would ignore. A bounded raw-entry visitor (for example, borrowed raw JSON entries parsed independently after the count gate) would preserve malformed-sibling isolation without the full `Value` tree and parallel shape/inbound DTOs. This is non-blocking if the semantic-dedup fix remains small.

NITPICK:
- none

SUMMARY: The entry-33 preflight, malformed-sibling isolation, privacy-safe error path, and storage-backed `(owner_ts, record_digest)` winner semantics are otherwise sound. I independently ran `cargo test -p marmot-app --locked`, `cargo test -p storage-sqlite --locked push_token`, `cargo fmt --all -- --check`, and `cargo clippy -p marmot-app --all-targets --locked -- -D warnings`; all passed. Deferring profile-specific kind-451/current-vs-legacy acceptance to the separate #1059 lane is consistent with the tracker, but the changed dedup path still permits repeated crypto/storage work for one canonical record.

VERDICT: REQUEST_CHANGES
SENSITIVE_PATHS: `crates/marmot-app/src/notifications.rs` (push owner-proof parsing and verification path)

agent-p1p:
Addressed the blocking dedup finding in `2979b3759da3c6897d4cab264d18ce95af650d69`.

- Relay hints now use one canonical trim/empty-to-absent helper across signing, owner-proof reconstruction, outbound records, and inbound pre-verification dedup.
- Regressions prove omitted vs whitespace-only and canonical vs whitespace-padded hints perform one owner-signature verification; the ingest path stores one record.
- `cargo test -p marmot-app --locked` passes (371 unit tests plus integration/doc targets).
- `cargo test -p storage-sqlite --locked push_token` and final `just fast-ci` pass.

The non-blocking `Vec<Value>` parser rewrite is deferred: arrays are already hard-capped at 32 before typed/crypto work, and replacing the bounded two-pass parser would add unrelated complexity to this fix.

agent-p1p:
@coderabbitai review

agent-p1p:
Addressed CodeRabbit's constant-drift nitpick in `4527d4b3688b97a0c0d44fd49c9173a1e15cf7a9`: the bounded-decoder error and its test now derive the limit from `PUSH_MAX_GOSSIP_ENTRIES`.

The `Vec<Value>` parser rewrite remains intentionally deferred as the review marked it optional/poor-tradeoff; the entry-33 preflight already bounds the untrusted path, and replacing the parser would add complexity unrelated to the blocking dedup fix.

erskingardner (commented):
# Code Review: PR #1084 — fix(push): bound advisory gossip arrays

## Summary

This PR bounds push-token gossip (`tokens`, kind 447/448) and removal (`removals`, kind 449)
arrays to 32 entries via a preflight, allocation-free `serde` visitor (`BoundedPushGossipEntries`)
that aborts as soon as a 33rd element is observed — before any typed decoding or owner-proof
signature verification. It then decodes surviving entries independently (`filter_map` instead of
`collect::<Result<Vec<_>,_>>()`) so one malformed sibling no longer discards the whole batch, and
deduplicates decoded/normalized entries via a `HashSet` before `into_record`/signature
verification, closing a repeat-signature-work vector for wire-equivalent duplicates (including
relay-hint whitespace variants). Verified against `crates/marmot-protocol/marmot` issue #950's
acceptance criteria: this addresses "reject >32 before typed decode/signature work," "one invalid
entry doesn't poison siblings," "dedupe without repeated signature work," and "deterministic
winner regardless of order" directly, with tests for each.

I read the full diff, the surrounding file for context, traced every caller of the touched
functions (`parse_push_gossip`/`verify_push_gossip` have exactly one production caller in
`lib.rs`, and `PushTokenGossipEntry`/`PushTokenRemovalEntry` are not constructed from untrusted
input anywhere else in the workspace), ran `cargo fmt --all -- --check`, `cargo clippy -p
marmot-app --all-targets --locked -- -D warnings` (clean), and `cargo test -p marmot-app --locked
notifications` (60/60 passing, including all new tests). I did not find a correctness or security
defect. The two items below are non-blocking.

## Issues

None found that block merge.

## Suggestions

### Efficiency: message content is fully parsed twice on every call

**`crates/marmot-app/src/notifications.rs:1050-1054` and `:1054-1055`**

> ```rust
> preflight_push_gossip_array::<PushTokenGossipShape>(content, "malformed push token gossip")?;
> let payload: InboundPushTokenGossipPayload = serde_json::from_str(content)
>     .map_err(|_| AppError::InvalidPushGossip("malformed push token gossip".into()))?;
> ```

`preflight_push_gossip_array` runs a full `serde_json::from_str` pass over `content` purely to
learn whether `tokens`/`removals` has ≤32 elements, then a second, independent
`serde_json::from_str` re-parses the same string into `InboundPushTokenGossipPayload`. For inputs
within the bound (the common case) this is a full 2x parse/allocate cost on every push-gossip
message, not just the oversized-attack case the PR is optimizing for. Since
`BoundedPushGossipEntries`'s `visit_seq` already walks the array with `IgnoredAny`, the same logic
could instead be inlined as a custom `Deserialize` for the `tokens`/`removals` field on
`InboundPushTokenGossipPayload` itself (erroring past entry 32 while still collecting the `Vec<Value>` for entries ≤32), collapsing this to one pass. Not blocking — correctness is unaffected — but worth a follow-up since the whole point of this PR is bounding gossip-processing cost.

### Test coverage: issue #950's fuzz/property-test criterion isn't covered

Issue #950's acceptance criteria include: "Fuzz/property tests keep CPU/allocation bounded for
malformed lengths." The PR's boundary/permutation/dedup tests (0/1/31/32/33/100k entries,
malformed-sibling, dedup, permutation) are thorough and well-targeted, but there's no
fuzz/property-based test (e.g., `proptest`/`arbitrary`-driven generation of malformed JSON shapes)
asserting bounded CPU/allocation across a wider input space than the hand-picked cases. Consider
filing or linking a fast-follow if this is intentionally deferred — the hand-written tests cover
the specific attack shape well, but a property test would give broader assurance for this
specific "bounded before untrusted-work" invariant.

## What's Done Well

- The preflight bound check is genuinely allocation-free and short-circuits at entry 33 rather
  than scanning the whole array — `oversized_arrays_do_zero_signature_verifications_and_yield_zero_records`
  and `maliciously_large_arrays_stop_at_the_bounded_preflight` verify this with a positive control
  (one verified signature) so the "zero verifications" assertion isn't vacuous.
- Error messages stay privacy-safe and non-distinguishing: both the array-size preflight failure
  and the post-decode structural failure map to the same generic `"malformed push token gossip"`/
  `"malformed push token removal"` string (`preflight_push_gossip_array`), and — importantly — this
  is *more* private than the prior code, which propagated whichever specific `into_record()`
  validation error occurred (e.g. "invalid encrypted token base64") for the entire batch.
- Dedup correctly happens on the decoded-and-normalized entry, before `into_record`/signature
  verification, and only collapses exact wire-equivalent duplicates — it doesn't touch the
  existing storage-level `(owner_ts, record_digest)` winner tie-break, which stays order-independent
  per `mixed_entry_permutations_apply_the_same_valid_winner`.
- Relay-hint canonicalization is centralized into one `normalized_relay_hint` helper reused by
  signing, verification, local-record construction, and inbound normalization, removing three
  copies of the same trim/filter logic (previously each could drift independently).
- `#[cfg(test)]` owner-signature-verification counters are a nice way to make "did the expensive
  path run" an assertable fact rather than an inferred one, and the tests correctly include a
  positive control to rule out a vacuously-passing zero-count assertion.

PR #1085, opened 2026-07-24
fix: validate kind-445 ingress tag shape

### LINKED ISSUE #1007: Enforce exact kind-445 tag shape without treating expiration as validity
## Summary

Make Nostr kind-445 group-message ingress enforce the adopted exact tag shape. Correct the prior issue scope: NIP-40 `expiration` is transport metadata and is **not** a Marmot event-validity rule. A client may receive/process an otherwise valid event after its expiration timestamp.

DNS/host rebinding controls are also a local outbound dial-safety concern, not kind-445 syntax; they are covered by the transport/dial work rather than this parser issue.

## Normative source

- `transports/nostr.md`
- `foundation/wire-envelopes.md`

## Start here

- `crates/transport-nostr-peeler/src/peeler.rs`
- `crates/transport-nostr-peeler/src/lib.rs`
- Nostr event building/parsing in `crates/transport-nostr-adapter/`
- relay runtime and peeler tests

## Required behavior

- Require the exact kind-445 tag names, counts, arities, marker values, and byte encodings specified for a Marmot group event.
- Reject extra or duplicate protocol tags where the spec requires an exact shape; do not accept a valid required subset surrounded by arbitrary tags.
- Validate the event signature/ID and outer-envelope fields before handing ciphertext to the MLS engine.
- Keep routing ID and raw MLS group ID distinct at every boundary.
- Treat a syntactically valid NIP-40 `expiration` tag only according to the adopted transport behavior. Do not reject solely because wall-clock time is after expiration.
- Produce typed malformed/unauthenticated/wrong-recipient/unknown-group outcomes without leaking identifiers into logs.

## Out of scope

- Relay marker/prior-route behavior (#1057).
- Durable outbound fanout (#1020).
- DNS resolution and host-safety policy except ensuring this parser does not pretend to provide it.

## Acceptance criteria

- Table-driven tests cover missing, duplicate, extra, misordered where relevant, wrong-arity, wrong-marker, malformed-length, and unknown tags.
- A valid event is accepted before and after its expiration timestamp; expiration is not an MLS/Marmot validity gate.
- Tampered ID/signature and routing/group-ID confusion are rejected before engine mutation.
- Direct and runtime ingest seams classify the same fixture identically.
- `just fast-ci` and targeted peeler/adapter/runtime tests pass.

agent-p1p:
Fixes #1007

## Summary
- authenticate kind-445 events before reading their cleartext routing metadata, while retaining peel-time verification as defense in depth
- enforce exactly one lowercase 32-byte `h` routing tag, at most one decimal `expiration` tag, no other tags, and strict tag arity
- validate `expiration` structurally without using it as a message-validity deadline
- add direct, adapter, peeler, and cold relay-runtime regressions for malformed tags, tampered identity/signature data, route-id confusion, strict content encoding, and before/after-expiration behavior

## Verification
- `just fast-ci`
- `cargo test -p transport-nostr-peeler --locked`
- `cargo test -p transport-nostr-adapter --locked`
- `cargo test -p marmot-app --locked`
- `cargo test -p marmot-app --test malformed_445_drain --locked`
- focused adapter and Marmot app relay-plane regression tests
- independent read-only Cursor review: passed, no blocking findings

## Sensitive paths
- `crates/transport-nostr-peeler/src/event.rs`
- `crates/transport-nostr-peeler/src/peeler.rs`
- Nostr adapter and Marmot app ingress/runtime regression paths

Sensitive ingress/authentication change. Manual merge approval required.

### DIFF
--- a/crates/marmot-app/src/relay_plane/tests.rs
+++ b/crates/marmot-app/src/relay_plane/tests.rs
@@ -951,20 +951,22 @@ async fn directory_live_event_rejected_after_subscription_removed() {
 }
 
 fn group_event(id_prefix: &str, transport_group_id: &[u8]) -> NostrTransportEvent {
-    // `to_transport_message` verifies the id against the event hash (#351), so
-    // the distinguishing prefix lives in the content and the id is computed
-    // from it — distinct `id_prefix` values still yield distinct event ids.
-    let mut event = NostrTransportEvent {
-        id: String::new(),
-        pubkey: "22".repeat(32),
-        created_at: 1_700_000_000,
-        kind: KIND_MARMOT_GROUP_MESSAGE,
-        tags: vec![vec!["h".into(), hex::encode(transport_group_id)]],
-        content: format!("encrypted {id_prefix}"),
-        sig: None,
-    };
-    event.id = event.computed_id();
-    event
+    // Stable runtime fixtures preserve duplicate-event ids while carrying the
+    // valid kind-445 signature required by the routing boundary.
+    static KEYS: std::sync::OnceLock<nostr::Keys> = std::sync::OnceLock::new();
+    let keys = KEYS.get_or_init(nostr::Keys::generate);
+    let signed = nostr::EventBuilder::new(
+        nostr::Kind::Custom(KIND_MARMOT_GROUP_MESSAGE as u16),
+        format!("encrypted {id_prefix}"),
+    )
+    .tags([nostr::Tag::custom(
+        nostr::TagKind::custom("h"),
+        [hex::encode(transport_group_id)],
+    )])
+    .custom_created_at(nostr::Timestamp::from_secs(1_700_000_000))
+    .sign_with_keys(keys)
+    .expect("sign kind-445 fixture");
+    NostrTransportEvent::from_nostr_event(&signed).expect("map signed kind-445 fixture")
 }
 
 #[test]
--- a/crates/marmot-app/src/tests.rs
+++ b/crates/marmot-app/src/tests.rs
@@ -148,19 +148,19 @@ fn assert_profile_word_list(name: &str, words: &[&str]) {
     }
 }
 
-fn relay_delivery(marker: &str, pubkey: String) -> cgka_traits::TransportDelivery {
-    // `to_transport_message` verifies the id against the event hash (#351), so
-    // the distinguishing marker lives in the content and the id is computed.
-    let mut event = NostrTransportEvent {
-        id: String::new(),
-        pubkey,
-        created_at: 1,
-        kind: transport_nostr_peeler::KIND_MARMOT_GROUP_MESSAGE,
-        tags: vec![vec!["h".to_owned(), "aa".repeat(32)]],
-        content: format!("ciphertext {marker}"),
-        sig: None,
-    };
-    event.id = event.computed_id();
+fn relay_delivery(marker: &str, keys: &nostr::Keys) -> cgka_traits::TransportDelivery {
+    let signed = nostr::EventBuilder::new(
+        nostr::Kind::Custom(transport_nostr_peeler::KIND_MARMOT_GROUP_MESSAGE as u16),
+        format!("ciphertext {marker}"),
+    )
+    .tags([nostr::Tag::custom(
+        nostr::TagKind::custom("h"),
+        ["aa".repeat(32)],
+    )])
+    .custom_created_at(nostr::Timestamp::from_secs(1))
+    .sign_with_keys(keys)
+    .expect("sign kind-445 fixture");
+    let event = NostrTransportEvent::from_nostr_event(&signed).expect("map signed fixture");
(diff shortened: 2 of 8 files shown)


agent-p1p:

Reviewed exact head `dae180c10c68042e070b15251875e745c9e90485` against #1007 and the adopted Marmot spec (`marmot@625f8aa`). The strict tag parser and signature-before-routing order are useful, and all local/hosted checks are green, but three acceptance requirements are still missing.

### Blocking

1. **Protocol rejections are still collapsed into malformed/backend errors.**

   `crates/transport-nostr-peeler/src/error.rs:6-13` has no unauthenticated/invalid-signature variant. Consequently event-id and signature failures become `NostrPeelerError::Malformed` (`event.rs:39-42`, `event.rs:106-120`), route/envelope mismatches become `PeelerError::Malformed` (`peeler.rs:464-483`), and the adapter converts every ingress rejection into `TransportAdapterError::Backend` (`transport-nostr-adapter/src/lib.rs:417-420`). The new adapter tests explicitly bless the mismatch: direct ingress is `Malformed`, adapter ingress is `Backend` (`tests/inbound_routing.rs:1881-1897`, `1916-1932`).

   This does not satisfy #1007's typed malformed/unauthenticated/wrong-recipient/unknown-group requirement or direct/runtime classification parity. It also conflicts with adopted `foundation/errors.md`, which distinguishes `invalid_encoding`, `invalid_signature`, `wrong_recipient`, and `unknown_group`; `Backend` denotes a local failure, not hostile protocol input. Add typed boundary variants and privacy-safe mappings through peeler, adapter, and engine terminal handling. Invalid signature/wrong recipient must remain ordinary terminal input and must not abort a drain.

2. **The routing id is still aliased to raw MLS `GroupId` bytes.**

   The new table case at `transport-nostr-peeler/src/event.rs:385-388` treats a 16-byte value as “raw MLS group id sized,” but repository policy explicitly says `GroupId` is opaque and variable-length; 16 bytes is only today's OpenMLS generation detail. More importantly, `cgka-engine/src/message_processor/ingest.rs:1856-1882` first reinterprets every transport routing id as a raw `GroupId` and returns that group before consulting `transport_group_id_index`. The peeler then checks only signed `h` against the already-derived `TransportEnvelope` (`peeler.rs:464-483`), not whether the selected group was resolved through its transport routing state.

   A 32-byte raw MLS group id can therefore shadow a Nostr routing id (including a route that belongs to another group), and an unknown Nostr route equal to stored raw group-id bytes is treated as that group instead of `unknown_group`. Length checks do not establish type separation. Resolve Nostr group delivery only through authenticated transport-route mapping (while preserving transport-neutral mock/direct behavior), and add a collision regression where route bytes equal a different/raw MLS group id. The event must route to the indexed group or reject before peel/state mutation; it must never fall through by byte alias.

3. **Inbound `expiration` is accepted on commits and proposals.**

   `validate_kind_445_tag_shape` permits a syntactically valid `expiration` on every kind-445 (`event.rs:281-330`). After decryption, `peel_group_message` returns only MLS bytes (`peeler.rs:239-245`), so the engine loses tag presence before it identifies the MLS content type (`cgka-engine/src/message_processor/ingest.rs:522-572`). A member can therefore publish a correctly signed/encrypted commit or proposal carrying `expiration`, and receivers will process it.

   Adopted `transports/nostr.md` says expiration applies to application messages only and a kind-445 carrying a commit or proposal **MUST NOT** carry it; this protects fetchable group-state history. Preserve expiration presence across the peel boundary (without making the engine parse Nostr), reject it once the owning MLS parser classifies Commit/Proposal, and add a real commit/proposal regression proving no engine mutation. Continue accepting syntactically valid application-message expiration both before and after wall-clock expiry, and do not compare a supplied value to local time.

### Suggestions

- `peeler.rs:198-245` authenticates the embedded event id but never binds it to `TransportMessage.id`; the latter is reused as `PeeledMessage.id`. Add a defense-in-depth equality check and a test that mutates only the normalized message id while leaving the signed event intact.
- `marmot-app/tests/malformed_445_drain.rs` replaced the prior short-content runtime case with an extra-tag case. Keep both seams covered: the new case proves adapter-boundary drop behavior, while the old case proved a mapped event that becomes `PeelerError::Malformed` cannot wedge the engine drain.

### Verification

- `cargo test -p transport-nostr-peeler --locked` — 35 passed
- `cargo test -p transport-nostr-adapter --features sdk --locked` — 72 passed
- `cargo test -p marmot-app --test malformed_445_drain --locked` — 1 passed
- `cargo test -p marmot-app --locked` — passed
- `just fast-ci` — passed
- GitHub CI — all reported jobs green at review time

Sensitive ingress/authentication paths reviewed: `transport-nostr-peeler`, `transport-nostr-adapter`, and the Nostr-to-engine routing seam.

erskingardner:
Closing as superseded by merged PR #1081, which implemented the same exact kind-445 ingress-shape fix and closed #1007.

Issue #1088, opened 2026-07-24
Hermes adapter: preserve one logical delivery identity across stream finalization, retries, and fallback


Datawav:
## Problem

The Marmot Hermes adapter does not preserve one stable logical delivery identity across live preview, stream finalization, retries, and direct-final fallback. If a finalize request commits but its acknowledgement is lost—or if a gateway fallback re-enters the adapter—the same logical answer can be posted again under a fresh operation/idempotency identity.

This matches reports of two identical or nearly identical Hermes messages, sometimes ordered unexpectedly. Content-only deduplication is not safe because distinct turns can legitimately produce the same text.

Verified against `origin/master` at `87f89884bab1bff8ecc0ae6a61bfb88b4c3e4140`:

- `integrations/hermes/marmot/adapter.py` has separate active-stream finalize, abandon-preview/direct-final, and direct-final paths.
- Fallback paths can move from `stream_finalize` to `_send_final_direct` after an error without a durable reconciliation step proving whether the first operation committed.
- Adapter calls do not require a stable logical delivery ID supplied by Hermes and reused by every transport attempt.
- #1039 separately confirms that connector idempotency has an in-flight check→send→record race even when callers do reuse the same key.

These are complementary gaps: fixing #1039 prevents concurrent same-key double-posts, but does not help when one logical turn is retried with different keys or switches from stream finalization to direct final.

## Dependency

Hermes must provide a stable per-delivery ID across streaming and fallback paths: https://github.com/NousResearch/hermes-agent/issues/70694

## Required invariant

```text
one logical final answer
→ one stable logical_delivery_id
→ one stable idempotency namespace/key across every attempt
→ at most one durable kind-9 final
```

Preview/activity events may have their own event IDs, but must be causally linked to the same logical turn. Same text must never be the identity.

## Proposed direction

- Accept `logical_delivery_id` in the adapter delivery context.
- Derive deterministic idempotency keys for stream begin/finalize/direct-final/fallback from that identity and operation role.
- Persist or retain the identity through reconnect and adapter/session reconstruction for the duration required by connector deduplication.
- Treat timeout/connection loss after a mutating request as `outcome unknown`, not `definitely failed`.
- Before switching from `stream_finalize` to `send_final`, reconcile/query the stream or final event by logical identity. If the connector lacks that query, add the minimal status/reconciliation operation.
- Complete #1039's pending/committed single-flight reservation so concurrent same-key callers await one result.
[...]

PR #1091, opened 2026-07-24
fix: deduplicate logical final deliveries

### LINKED ISSUE #1039: agent-connector: send_final / stream_finalize idempotency has an in-flight TOCTOU — two concurrent same-key+same-body requests both send, double-posting an unrecallable message
### Summary

`SendIdempotencyStore` dedup is check-then-act with no reservation for an in-flight key. Two control requests carrying the same `idempotency_key` and identical body both call `idempotency.get(key, &fingerprint)` before either has completed its send; both miss (no record exists yet), both proceed to `runtime.send_message` / `reply_to_message`, and both durably post. `record()` enforces first-write-wins only *after* a send returns — too late to stop the concurrent duplicate the store exists to prevent.

### Location

- `crates/agent-connector/src/messaging.rs:115-148` (`send_final_response`: `get` at `:125-126` → `await` send at `:133-141` → `record` at `:145-147`)
- `crates/agent-connector/src/stream_session.rs:114-163` (`SendIdempotencyStore::get` / `record`; `record` no-ops when `seen.contains_key(&key)` at `:131-133`)
- Same shape in `stream_finalize_response` (`src/stream.rs`, `stream_finalize:` keys).

```rust
// messaging.rs — no reservation between get-miss and record:
    && let Some(ids) = self.idempotency.get(key, &fingerprint) {
    return Ok(FinalSent { message_ids_hex: ids });   // both concurrent requests MISS here
}
let summary = /* ... runtime.send_message(...).await ... */;   // both send
if let Some(key) = idempotency_key {
    self.idempotency.record(key, fingerprint, summary.message_ids.clone());  // 2nd is a no-op — too late
}
```

### Failure scenario

Two requests with the same `idempotency_key` and identical body arrive concurrently on the connector socket (a pipelining gateway, a client that retries without awaiting the first response, or two integrations sharing the same `wn-agent` socket per the coexistence model). Both call `get` → both miss (neither has recorded yet) → both `await` the durable send → both post. `record()` then stores the first and no-ops the second. Net result: the message is posted twice and cannot be recalled — exactly the outcome the idempotency store is meant to prevent. The store's own doc comment frames its guarantee around "a retry after a post-write timeout" (sequential), leaving the concurrent/pipelined case unprotected.

### Why not a duplicate

#897 concerns the fingerprint being computed over an *un-normalized* `group_id_hex` (a differently-cased *sequential* retry). #898 / #807 concern *empty message_id* dedup on the CLI subscription path. Neither addresses the in-flight race in the connector's `SendIdempotencyStore`.

### Suggested severity

**LOW** — requires two genuinely concurrent same-key + same-body requests; a normal retry-after-timeout is sequential and already covered. Security-adjacent because the harm is an unrecallable double-post.

### Suggested fix

Reserve the key at `get`-miss time: insert a pending marker (or hold a per-key async lock / oneshot) before the send, so a concurrent same-key request blocks on and returns the in-flight result instead of issuing a second durable send. Release/replace the reservation with the real record on completion (and clear it on send failure so the key stays retryable).

---
*Filed from a meticulous automated code-review pass. The get→send→record ordering was verified against source, and the finding was checked against existing open issues (#897, #898, #807) for duplicates before filing.*

agent-p1p:
## Summary

- derive a stable opaque logical-final idempotency key when Hermes supplies `metadata.logical_delivery_id`
- reuse that key across stream finalization, bounded retries, delivery-status reconciliation, and direct fallback
- make connector final sends single-flight so concurrent same-key callers share one committed result
- preserve the stream-start parent when a direct fallback is necessary
- keep stream-begin request IDs stable only inside one retry loop so reconstructed adapters cannot reuse stale transcript state
- reject an already-committed logical key when a new active stream tries to bind different final inputs
- add privacy-safe attempt correlation and regression coverage for lost acknowledgements, fallback, same-text turns, concurrency, reconstruction, and key conflicts

## Scope and dependency

This lands the MDK/connector half of #1088. The end-to-end reconstruction guarantee remains dependent on [NousResearch/hermes-agent#70694](https://github.com/NousResearch/hermes-agent/issues/70694), because current Hermes does not yet provide `logical_delivery_id` at the platform-adapter boundary. Without that metadata, the adapter generates an identity that is stable only within one in-process send/finalize attempt. For that reason this PR references, but does not close, #1088.

The connector single-flight reservation fully fixes #1039.

## Verification

- `python3 -m unittest discover -s integrations/hermes/marmot/tests -v` — 98 passed
- `cargo test -p agent-control` — 14 passed
- `cargo test -p agent-connector` — 115 passed
- `cargo test -p wn-opencode` — 48 passed, 1 process-spawning e2e test ignored by default
- `just fast-ci` — passed workspace fmt, naming, all-target checks, OTLP checks, and clippy

Fixes #1039
Refs #1088

## Summary by CodeRabbit

* **New Features**
  * Added delivery-status checks for tracking whether a message is unknown, pending, or committed.
  * Added stable logical delivery identifiers for reliable retries across previews, finalization, and direct sends.

* **Bug Fixes**
  * Prevented duplicate message delivery when concurrent or repeated requests use the same idempotency key.
  * Improved recovery after uncertain stream-finalization results by checking delivery status before falling back.
  * Added clear errors when an idempotency key is reused with different message content.
  * Preserved reply threading during final-send fallback.

### DIFF
--- a/crates/agent-connector/src/stream.rs
+++ b/crates/agent-connector/src/stream.rs
@@ -14,13 +14,15 @@ use tokio::sync::{mpsc, oneshot};
 use transport_quic_broker::OpenBrokerTextPublisher;
 
 use crate::error::ConnectorError;
+use crate::messaging::{LOGICAL_FINAL_IDEMPOTENCY_PREFIX, send_final_fingerprint};
 use crate::quic::{
     broker_trust_for_candidate, first_quic_candidate, parse_quic_candidate,
     resolve_quic_candidate_addr,
 };
 use crate::stream_session::{
-    ActiveStreamSession, FinalizedStream, StreamBeginReceipt, StreamBeginReservation,
-    normalize_stream_capability,
+    ActiveStreamSession, FinalizedStream, SendIdempotencyAcquisition,
+    SendIdempotencyReservationGuard, SendIdempotencyStatus, StreamBeginReceipt,
+    StreamBeginReservation, normalize_stream_capability,
 };
 use crate::validation::{normalize_hex, transcript_hash_from_hex, unix_now_seconds};
 use crate::{
@@ -90,11 +92,6 @@ fn begun_response(receipt: StreamBeginReceipt) -> AgentControlResponse {
 
 fn stream_finalize_idempotency_key(key: &str) -> String {
     format!("stream_finalize_v2:{key}")
-}
-
-struct StreamFinalizeIdempotency {
-    key: String,
-    fingerprint: String,
 }
 
 impl AgentConnector {
@@ -248,6 +245,7 @@ impl AgentConnector {
             stream_id_hex.clone(),
             ActiveStreamSession {
                 account_label: account.label,
+                account_id_hex: account.account_id_hex,
                 group_id,
                 stream_id,
                 stream_capability,
@@ -352,33 +350,84 @@ impl AgentConnector {
         let stream_id_hex = normalize_hex(stream_id_hex)?;
         let stream_capability = normalize_stream_capability(stream_capability)?;
         let transcript_hash = transcript_hash_from_hex(transcript_hash_hex)?;
-        let fingerprint = stream_finalize_fingerprint(
-            &stream_id_hex,
-            &stream_capability,
-            &final_text,
-            &transcript_hash,
-            chunk_count,
-        );
         let idempotency_key = idempotency_key
             .as_deref()
             .map(str::trim)
             .filter(|key| !key.is_empty())
-            .map(stream_finalize_idempotency_key);
-        if let Some(key) = idempotency_key.as_deref()
-            && let Some(message_ids_hex) = self.idempotency.get(key, &fingerprint)
+            .map(str::to_owned);
+        let legacy_idempotency = if let Some(client_key) = idempotency_key
+            .as_deref()
+            .filter(|key| !key.starts_with(LOGICAL_FINAL_IDEMPOTENCY_PREFIX))
         {
-            return Ok(AgentControlResponse::StreamFinalized {
-                stream_id_hex,
-                message_ids_hex,
-            });
-        }
+            let key = stream_finalize_idempotency_key(client_key);
+            let fingerprint = stream_finalize_fingerprint(
+                &stream_id_hex,
+                &stream_capability,
+                &final_text,
+                &transcript_hash,
+                chunk_count,
+            );
+            match self.idempotency.acquire(&key, &fingerprint).await? {
+                SendIdempotencyAcquisition::Completed(message_ids_hex) => {
+                    return Ok(AgentControlResponse::StreamFinalized {
+                        stream_id_hex,
+                        message_ids_hex,
+                    });
+                }
+                SendIdempotencyAcquisition::Leader(reservation) => Some(reservation),
+            }
+        } else {
+            None
+        };
         // Clone (not remove) the session: the final text/hash/chunk-count
         // expectation is validated inside the compose task, atomically with
         // its teardown. On mismatch the session stays registered and the
         // compose task keeps running, so finalize is retryable (#366).
-        let session = self
+        let session = match self
             .streams
-            .get_authorized(&stream_id_hex, &stream_capability)?;
+            .get_authorized(&stream_id_hex, &stream_capability)
+        {
+            Ok(session) => session,
+            Err(err) => {
+                // A successful finalize removes its stream session before the
+                // control response is written. A same-key retry after a lost
+                // acknowledgement can therefore recover from the committed
+                // logical result even though the capability no longer resolves.
+                if let Some(key) = idempotency_key
+                    .as_deref()
+                    .filter(|key| key.starts_with(LOGICAL_FINAL_IDEMPOTENCY_PREFIX))
+                    && let SendIdempotencyStatus::Committed(message_ids_hex) =
+                        self.idempotency.status(key)
+                {
+                    return Ok(AgentControlResponse::StreamFinalized {
+                        stream_id_hex,
+                        message_ids_hex,
+                    });
+                }
+                return Err(err);
+            }
+        };
+        let idempotency = if let Some(client_key) =
+            idempotency_key.filter(|key| key.starts_with(LOGICAL_FINAL_IDEMPOTENCY_PREFIX))
+        {
+            let fingerprint = send_final_fingerprint(
+                &session.account_id_hex,
+                &hex::encode(session.group_id.as_slice()),
+                &final_text,
+                Some(&session.start_message_id_hex),
+            );
+            match self.idempotency.acquire(&client_key, &fingerprint).await? {
+                SendIdempotencyAcquisition::Completed(message_ids_hex) => {
+                    return Ok(AgentControlResponse::StreamFinalized {
+                        stream_id_hex,
+                        message_ids_hex,
+                    });
+                }
+                SendIdempotencyAcquisition::Leader(reservation) => Some(reservation),
+            }
+        } else {
+            legacy_idempotency
+        };
 
         // Retry fast-path: a prior finalize already validated the transcript and
         // the compose task exited, but the durable finish below failed. The
@@ -401,7 +450,7 @@ impl AgentConnector {
                     final_text,
                     transcript_hash,
                     chunk_count,
-                    idempotency_key.map(|key| StreamFinalizeIdempotency { key, fingerprint }),
+                    idempotency,
                 )
                 .await;
         }
@@ -470,7 +519,7 @@ impl AgentConnector {
             final_text,
             transcript_hash,
             chunk_count,
-            idempotency_key.map(|key| StreamFinalizeIdempotency { key, fingerprint }),
+            idempotency,
         )
         .await
     }
@@ -486,7 +535,7 @@ impl AgentConnector {
         final_text: String,
         transcript_hash: [u8; 32],
         chunk_count: u64,
-        idempotency: Option<StreamFinalizeIdempotency>,
+        idempotency: Option<SendIdempotencyReservationGuard>,
     ) -> Result<AgentControlResponse, ConnectorError> {
         let (_payload, summary) = self
             .runtime
@@ -504,11 +553,7 @@ impl AgentConnector {
             )
             .await?;
         if let Some(idempotency) = idempotency {
-            self.idempotency.record(
-                idempotency.key,
-                idempotency.fingerprint,
-                summary.message_ids.clone(),
-            );
+            idempotency.complete(summary.message_ids.clone());
         }
         // Durable final published: it is now safe to drop the session.
         let _ = self.streams.remove_if_same(stream_id_hex, session);
--- a/crates/agent-connector/src/stream_session.rs
+++ b/crates/agent-connector/src/stream_session.rs
@@ -58,11 +58,10 @@ const SEND_IDEMPOTENCY_FILE_VERSION: u8 = 1;
 ///
 /// A retry that reuses the same key AND matches the recorded fingerprint returns
 /// the cached ids without re-sending, so a retry after a post-write timeout cannot
-/// double-post an unrecallable encrypted message. A reused key whose fingerprint
-/// differs (a different request body under the same key) is treated as a cache
-/// miss, so it can never return ids belonging to an unrelated send. Keys are
-/// evicted oldest-first once the capacity is reached and the eviction is mirrored
-/// to disk.
+/// double-post an unrecallable encrypted message. Reusing a key with a different
+/// fingerprint is rejected so one logical delivery identity cannot name two
+/// request bodies. Keys are evicted oldest-first once the capacity is reached and
+/// the eviction is mirrored to disk.
 ///
 /// Records are persisted under [`SEND_IDEMPOTENCY_FILE`] with crash-safe atomic
 /// renames so a connector restart can still dedup a bounded retry window. The
@@ -93,6 +92,54 @@ struct SendIdempotencyInner {
     order: std::collections::VecDeque<String>,
     seen: HashMap<String, (String, Vec<String>)>,
     recorded_at: HashMap<String, u64>,
+    in_flight: HashMap<String, InFlightSendIdempotency>,
+}
+
+struct InFlightSendIdempotency {
+    fingerprint: String,
+    completion: watch::Sender<bool>,
+}
+
+pub(crate) enum SendIdempotencyReservation {
+    Completed(Vec<String>),
+    Wait(watch::Receiver<bool>),
+    Leader(SendIdempotencyReservationGuard),
+}
+
+pub(crate) enum SendIdempotencyAcquisition {
+    Completed(Vec<String>),
+    Leader(SendIdempotencyReservationGuard),
+}
+
+pub(crate) enum SendIdempotencyStatus {
+    Unknown,
+    Pending,
+    Committed(Vec<String>),
+}
+
+pub(crate) struct SendIdempotencyReservationGuard {
+    store: SendIdempotencyStore,
+    key: String,
+    fingerprint: String,
+    active: bool,
+}
+
+impl SendIdempotencyReservationGuard {
+    pub(crate) fn complete(mut self, message_ids: Vec<String>) {
+        let completed = self
+            .store
+            .complete_reservation(&self.key, &self.fingerprint, message_ids);
+        debug_assert!(completed, "active send reservation must still exist");
+        self.active = !completed;
+    }
+}
+
+impl Drop for SendIdempotencyReservationGuard {
+    fn drop(&mut self) {
+        if self.active {
+            self.store.release_reservation(&self.key, &self.fingerprint);
+        }
+    }
 }
 
 impl SendIdempotencyStore {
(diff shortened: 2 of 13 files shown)


integrations/hermes/marmot/adapter.py
```diff
@@ -1854,17 +2115,32 @@ async def _send_final_direct(
...
+        idempotency_key: Optional[str] = None,
+        stream_id_hex: Optional[str] = None,
     ) -> SendResult:
         # One idempotency key for every attempt of THIS durable reply: a retry
         # after a post-write timeout reuses the key so the connector dedups
         # instead of double-posting an unrecallable encrypted message. Bounded
         # retries with a tiny backoff cover the transient/timeout window;
         # non-retryable errors fail fast. Mirrors OpenClaw dispatch.ts.
-        idempotency_key = uuid.uuid4().hex
+        account_id = await self._ensure_account_id()
```
erskingardner, line 2127: **Regression (reproduced).** `account_id = await self._ensure_account_id()` used to be the first statement *inside* the `try:` in the retry loop. Hoisting it here removes it from both the retry loop and the exception handler.

`_ensure_account_id()` raises on real failures — `AgentControlError(code="no_accounts")`, `code="ambiguous_account"`, and any `OSError`/`AgentControlError` from `self.client.account_list()` when the connector socket is down. Every other failure in this function returns a `SendResult`; this one now propagates out of `_send_final_direct` and out of `send()`.

Against this branch vs master, with an adapter that has no configured `account_id_hex` and a client whose `account_list()` raises `OSError`:

```
=== master ===
RETURNED SendResult(success=False, retryable=True)
account_list attempts: 3

=== pr/1091 ===
RAISED out of send(): OSError: connector socket is down
account_list attempts: 1
```

So a transient connector-socket blip during account resolution loses all three attempts *and* changes the failure from a structured retryable `SendResult` into a raised exception.

The key only needs `account_id` when `idempotency_key is None`, so the minimal fix is to resolve it inside the loop as before and derive the key lazily — or keep the hoist but wrap it:

```python
try:
    account_id = await self._ensure_account_id()
except Exception as exc:
    logger.debug("Marmot account resolution failed: %s", exc)
    return SendResult(success=False, error=str(exc), retryable=is_retryable(exc))
```

integrations/hermes/marmot/adapter.py
```diff
@@ -1566,6 +1696,26 @@ async def send(
         stream = self._last_chat_stream.get(chat_id)
         if stream is not None and not stream.finalized:
             message_id = _stream_message_id(stream.stream_id_hex)
+            if logical_delivery_id is not None:
+                account_id = await self._ensure_account_id()
```
erskingardner, line 1700: Same hazard as `_send_final_direct`: this `await self._ensure_account_id()` sits outside any `try:` in `send()`, so a connector blip here raises out of `send()` rather than returning a `SendResult`.

Worth noting it also runs on the hot path of *every* final that follows a live preview, purely to recompute a key the stream already holds. Caching the account id on the stream (it is already passed to `MarmotLiveStream.__init__`) would let this compare without an await at all.

integrations/hermes/marmot/adapter.py
```diff
@@ -1839,13 +2096,17 @@ async def _abandon_preview_and_send_final(
...
     ) -> SendResult:
         await self._cancel_stream(chat_id, message_id, stream, reason)
         return await self._send_final_direct(
             chat_id,
             final_text,
-            reply_to_message_id_hex=reply_to_message_id_hex,
+            # A direct fallback is still the final node of the same causal
+            # stream chain. The stream-start already points at the prompt, so
+            # the fallback must point at the start rather than skipping it.
+            reply_to_message_id_hex=stream.start_message_id_hex,
```
erskingardner, line 2107: **The reconstruction case ends in a permanent failure, not a dedup.**

Pinning the fallback to `stream.start_message_id_hex` is right for causality, but it means the *same logical delivery* produces two different server-side fingerprints depending on whether a preview stream existed at the time:

- with a preview → parent = that stream's start message (here, and `stream.rs:378` for the finalize path)
- reconstructed adapter, no preview → parent = `_optional_hex(reply_to)`, the prompt (adapter.py:1755)

And a reconstructed adapter that *does* re-open a preview gets a **new** start message, so its fingerprint differs again — which the README explicitly says is intended ("a reconstructed adapter never replays a begin").

Since `reserve()` now hard-rejects a fingerprint mismatch, the replay this PR exists to handle hits `idempotency_key_conflict`. Modelling `SendIdempotencyStore::reserve` and driving the real adapter through turn → crash → replay with the same `logical_delivery_id`:

```
A final: success=True id=…0001
B replay: success=False retryable=False error='idempotency key was reused with different inputs'

durable sends actually made: 1
   stream_final hello world parent=…aa01
```

Adapter B cancels its live preview (via `_abandon_preview_and_send_final`), then the direct send conflicts too — so the user sees a preview vanish and no answer, and Hermes is told the delivery permanently failed, while the message is in fact committed and its ids are available from `delivery_status`.

`idempotency_key_conflict` arrives as `AgentControlError(retryable=False)` (`_raise_if_error` never sets `retryable`), so it is terminal in both `_send_final_direct`'s handler and `send()`'s finalize handler.

The machinery to fix this is already in the PR — it just is not wired to this path. Suggest: on `idempotency_key_conflict`, call `delivery_status(key)` and return `committed` ids as success before treating it as an error. Alternatively, drop the reply parent from the fingerprint for `marmot_logical_final_v1:` keys, since for those the key already names the delivery and the parent is a routing detail rather than an identity input.

Either way this deserves a regression test: commit under a key with the stream-start parent, then replay the same logical delivery from a fresh adapter.

crates/agent-connector/src/stream.rs
```diff
@@ -352,33 +350,84 @@ impl AgentConnector {
...
+            Ok(session) => session,
+            Err(err) => {
+                // A successful finalize removes its stream session before the
+                // control response is written. A same-key retry after a lost
+                // acknowledgement can therefore recover from the committed
+                // logical result even though the capability no longer resolves.
+                if let Some(key) = idempotency_key
+                    .as_deref()
+                    .filter(|key| key.starts_with(LOGICAL_FINAL_IDEMPOTENCY_PREFIX))
+                    && let SendIdempotencyStatus::Committed(message_ids_hex) =
```
erskingardner, line 399: This recovery path bypasses the conflict guard the rest of the PR relies on. `status(key)` takes no fingerprint, so any finalize carrying a committed logical key returns those message ids regardless of its `final_text`, `transcript_hash`, or `chunk_count`.

The enforcement is therefore conditional on an unrelated fact — whether a live session happens to still resolve:

- session exists → `acquire()` at line 419 rejects with `IdempotencyKeyConflict`
- session gone → this branch returns the recorded ids, unchecked

"Session gone" is the *common* replay case: sessions are memory-only, so a connector restart or a sweeper pass leaves the persisted `seen` map with no session to validate against. A same-key finalize with different final text then gets `stream_finalized` back with ids for text it never sent; the adapter records them in `_sent_targets` and reports success.

That directly contradicts the invariant stated at messaging.rs:127 — "one idempotency key names exactly one logical final. Reusing it with different inputs is rejected before any send."

I recognise the fingerprint is genuinely unreconstructable here (it needs `session.account_id_hex`, `group_id`, and `start_message_id_hex`, all of which died with the session). That is exactly why this is worth addressing rather than leaving implicit — options: persist enough alongside the reservation to validate a replay (e.g. a digest of the final text and the destination), or at minimum document that the conflict guard is best-effort and does not survive session loss, so the guarantee is not read as stronger than it is.

crates/agent-connector/src/stream_session.rs
```diff
@@ -117,30 +165,140 @@ impl SendIdempotencyStore {
...
 
+    pub(crate) fn reserve(
+        &self,
+        key: &str,
+        fingerprint: &str,
+    ) -> Result<SendIdempotencyReservation, ConnectorError> {
+        let mut inner = crate::lock_recover(&self.inner);
+        if let Some((recorded, message_ids)) = inner.seen.get(key) {
+            if !constant_time_eq(recorded.as_bytes(), fingerprint.as_bytes()) {
+                return Err(ConnectorError::IdempotencyKeyConflict);
```
erskingardner, line 176: **Fail-open → fail-closed, and existing key producers were not audited.**

The previous semantics ("a reused key with a different body is a cache miss, so the send simply proceeds without caching") were fail-open. This makes it fail-closed, permanently: the record has no TTL, survives restart via `send-idempotency.json`, and is only released by FIFO eviction after 1024 keys.

`integrations/opencode/marmot/src/control.rs:106` builds its key as:

```rust
let idempotency_key = format!("{reply_to_ref}:reply:{chunk_index}");
```

That is derived from the parent message and chunk index only — deliberately independent of the text. Under master, a second reply to the same parent with different text was an uncached send. Under this branch it returns `idempotency_key_conflict`, which `control.rs` maps to `HarnessError::ControlRejected` and the reply is dropped with no fallback. Re-running an agent against the same prompt, or any flow that produces a second answer for one parent, hits chunk 0 every time.

The PR only tests the conflict on the stream-finalize path (`req-replacement-stream-finalize`); there is no `send_final` conflict test and no test covering a non-logical-prefix key producer.

If the hard reject is intended for `marmot_logical_final_v1:`-prefixed keys — where the key genuinely names one logical delivery — consider scoping it to that prefix and leaving unprefixed keys on the previous cache-miss behaviour, or update the opencode key derivation to include the text in the same change.

crates/agent-connector/src/stream_session.rs
```diff
@@ -117,30 +165,140 @@ impl SendIdempotencyStore {
...
+        fingerprint: &str,
+    ) -> Result<SendIdempotencyAcquisition, ConnectorError> {
+        loop {
+            match self.reserve(key, fingerprint)? {
+                SendIdempotencyReservation::Completed(message_ids) => {
+                    return Ok(SendIdempotencyAcquisition::Completed(message_ids));
+                }
+                SendIdempotencyReservation::Wait(mut completion) => {
+                    if !*completion.borrow() {
+                        let _ = completion.changed().await;
```
erskingardner, line 219: This wait is unbounded, and `handle_request` is not wrapped in a timeout — only the frame read/write are (`CONTROL_OPERATION_TIMEOUT`, 15s). So a follower blocks its per-connection task for as long as the leader's durable send takes.

Each blocked follower holds a `connection_limiter` permit (lib.rs:326), and that limiter uses `try_acquire_owned` — over the cap, connections are refused with `server_busy` rather than queued. A slow relay publish plus a handful of same-key retries can therefore starve *unrelated* control traffic on the same socket.

The protocol already has the vocabulary for this: bound the wait (`CONTROL_OPERATION_TIMEOUT` seems the natural budget) and, on expiry, return the same signal `delivery_status` would give — `pending` — which `_reconcile_ambiguous_stream_final` already handles correctly by returning a retryable result and keeping the session handle.

Minor, same block: the `loop` means a follower woken by a *failed* leader re-reserves and becomes the leader itself. That is the right behaviour, but it means "single-flight" bounds concurrent sends, not total sends — worth saying so in the doc comment at line 61, which currently reads as a stronger guarantee.

crates/agent-connector/src/stream_session.rs
```diff
@@ -117,30 +165,140 @@ impl SendIdempotencyStore {
...
-            true
+            insert_committed(&mut inner, key, fingerprint, message_ids)
         };
         if !should_persist {
             return;
         }
+        self.persist_off_hot_path();
+    }
+
+    fn persist_off_hot_path(&self) {
```
erskingardner, line 301: This is the load-bearing detail behind the new `delivery_status` contract, and I think the README overstates it.

The comment here is candid — "a crash after the send returns but before the spawned write lands can lose the on-disk record, so a retry after restart may re-send". Combined with `in_flight` being memory-only, `unknown` is returned in at least three cases where the message may already be committed:

1. crash between `complete_reservation` and the `spawn_blocking` write landing
2. FIFO eviction past `SEND_IDEMPOTENCY_CAPACITY` (1024, no TTL)
3. restart while a leader was in flight — `Pending` collapses to `Unknown`, even though that process may have already published to the relay

The README says the adapter "only falls back when the connector reports `unknown`", which reads as a safety property. It is not one: `unknown` means "no local record", and the adapter converts it into a second durable, unrecallable send (`_reconcile_ambiguous_stream_final` → `_abandon_preview_and_send_final`).

This is not a regression — master fell back blindly too — but it is the residual duplicate window in a PR whose stated purpose is closing it, so it should be named rather than implied. It also reads against `docs/marmot-architecture/overview/multi-step-state-changes.md`: "record intent before external side effects". The reservation *is* the intent record, and it is the one piece of state deliberately not persisted.

Suggest either persisting the reservation before the send (so a restart can answer `pending`/attempted rather than `unknown`), or reframing the status name and README wording as "no local record — fallback may duplicate".

integrations/hermes/marmot/adapter.py
```diff
@@ -253,6 +254,64 @@ def add(self, key: str) -> None:
...
+
+def _log_delivery_attempt(
+    *,
+    logical_delivery_key: Optional[str],
+    stream_id_hex: Optional[str],
+    operation_role: str,
+    attempt: int,
+    outcome: str,
+) -> None:
+    log = logger.debug if outcome == "started" else logger.info
```
erskingardner, line 289: Minor: `logger.info` fires for every outcome except `"started"`, which includes the happy path — every `stream_begin` commit, every `stream_finalize` commit, and every `direct_final` commit. That is 3+ INFO lines per agent turn in steady state. `debug` for `committed` and `info` only for `retryable_failure` / `rejected` / reconciliation outcomes would keep the diagnostic value without the volume.

Also: `_redaction_safe_ref` truncates to 12 hex chars (48 bits). Fine as a correlator, but it is a stable one — `stream_ref` links every attempt of a stream, and `logical_delivery_ref` links every attempt of a delivery — so it is pseudonymous rather than anonymous. Worth a line in the observability notes, since `docs/marmot-architecture/overview/observability.md` otherwise calls for aggregate values only.

(The test asserts the raw values do not appear in `str(extra)`, which is the right check — it just does not speak to linkability.)

erskingardner (commented):
## Adversarial review

I tried to break the delivery guarantee this PR claims rather than confirm it. The single-flight reservation and the `delivery_status` round-trip are the right shape, and the `#1039` concurrency fix looks solid. But I found two issues I could reproduce against this branch, plus three design gaps where the new invariant is either unenforced or enforced too aggressively.

The most important one: **the headline scenario — a reconstructed adapter replaying the same logical delivery — currently ends in a permanent, non-retryable failure, not a dedup.**

### Reproduced

**1. `_ensure_account_id()` moved out of the retry try-block (adapter.py:2127)**

It was inside the `try:` in the send loop; it is now hoisted above it, so it neither retries nor converts to a `SendResult`. Same script, two checkouts:

```
=== master ===
RETURNED SendResult(success=False, retryable=True)
account_list attempts: 3

=== pr/1091 ===
RAISED out of send(): OSError: connector socket is down
account_list attempts: 1
```

A transient socket failure while resolving the account now escapes `send()` instead of being retried and reported. Same hazard at adapter.py:1700.

**2. Reconstructed replay → `idempotency_key_conflict`, delivery reported as permanently failed**

The final fingerprint binds `reply_to_message_id_hex`, but the same logical delivery legitimately presents *two different parents*: the stream-start message when a preview exists (stream.rs:378, adapter.py:2107) and the prompt parent when it does not. A reconstructed adapter opens a fresh preview (which the README says is deliberate), so its stream-start differs, and `reserve()` hard-rejects.

Modelling `SendIdempotencyStore::reserve` against the real adapter:

```
A final: success=True id=…0001
B replay: success=False retryable=False error='idempotency key was reused with different inputs'
durable sends actually made: 1
```

The duplicate is prevented, but adapter B cancels its live preview and reports a permanent failure for a delivery that is committed and whose ids are sitting right there in the store. `delivery_status` is never consulted on the conflict path. Suggested fix: on `idempotency_key_conflict`, query `delivery_status(key)` and return `committed` ids as success (`_send_final_direct`, and `send()`'s finalize handler which currently routes conflicts into `_abandon_preview_and_send_final`).

### Design gaps

**3.** The finalize recovery path (stream.rs:399) returns committed ids via `status(key)`, which takes no fingerprint — so the conflict guard the rest of the PR depends on is bypassed exactly when the session is gone (after a connector restart or a session sweep, i.e. the common replay case). The identical request is rejected when a session happens to exist. Note `get(key, fingerprint)` — the function that would check this — is the one this PR moves to `#[cfg(test)]`.

**4.** Reservations are memory-only and the commit persist is explicitly at-least-once off the hot path, so `unknown` means "no local record", not "not delivered". The README frames it as a safety property. This also reads against `docs/marmot-architecture/overview/multi-step-state-changes.md` ("record intent before external side effects").

**5.** Turning a fingerprint mismatch from a cache miss into a hard reject is a fail-closed change that was not audited across existing key producers. `integrations/opencode/marmot/src/control.rs:106` derives `{reply_to_ref}:reply:{chunk_index}` — text-independent, no TTL, persisted. A second reply to the same parent with different text used to send; it now returns `idempotency_key_conflict` and the message is dropped.

Detail inline. Verification: `python3 -m unittest discover -s integrations/hermes/marmot/tests` — 98 passed on this branch.

erskingardner:
Closing in favor of #1108, which isolates the concrete #1039 in-flight TOCTOU fix.

This PR combined that low-level race with the much broader #1088 logical-delivery and reconstruction design. The latter depends on https://github.com/NousResearch/hermes-agent/issues/70694, which is still open and needs a product and contract decision. This implementation also added substantial protocol and state machinery without establishing a durable cross-crash at-most-once guarantee; the adversarial review exposed further semantic regressions and unresolved identity and persistence choices.

#1088 remains open for deliberate design work after the upstream delivery context is defined.

PR #1094, opened 2026-07-24
Complete NIP-65 relay roles and durable prior group routes

### LINKED ISSUE #1057: Implement NIP-65 relay marker semantics and durable prior group routes

erskingardner:
## Summary

- interpret NIP-65 relay markers as read/write capabilities and preserve those roles through runtime, CLI, and publication state
- persist authenticated prior Nostr routing addresses and the last epoch observed for the current route
- retain current and prior subscriptions until both protocol retained-history windows have advanced past them
- restore those subscriptions before restart sync, including a full prior-route backfill that cannot be clipped by the account-global Nostr cursor
- preserve the actual displaced-route epoch across rollback/fork recovery and deduplicate repeated route history

## Why

PR #1082 contained the reviewed NIP-65 marker half of #1057, but it was merged only into its stacked base rather than `master`. This PR carries those reviewed commits forward and completes the durable prior-route half of the adopted `marmot.transport.nostr.routing.v1` rotation rules.

Without durable prior routes, MDK discarded the prior signed group route after a routing update. Delayed or offline kind-445 events could then become unreachable, especially after restart when their `created_at` predates the account-global subscription cursor.

## Impact

Existing account databases gain migration `0030_prior_nostr_routes`. Legacy rows default to no history and epoch zero. Groups without a retained prior route keep the existing bounded `since` behavior; only groups carrying current-plus-prior routes perform the required full route backfill.

## Validation

- `just fast-ci`
- `cargo test -p transport-nostr-adapter`
- `cargo test -p marmot-app`
- `cargo test -p wn-cli`

Closes #1057

## Summary by CodeRabbit

- **New Features**
  - Direction-preserving NIP-65 relay lists with separate read/write roles, including parser and publishing support.
  - Persisted group transport history to support prior-route delivery after restarts and routing changes.
  - Full group route reconciliation across current and retained subscriptions.

- **Bug Fixes**
  - Improved group projection behavior when runtime metadata isn’t available.
  - CLI relay add/remove for `nip65` now round-trips directional roles correctly.
  - Relay-list syncing now preserves read/write ordering and avoids unintended role flattening.
  - Restart behavior now backfills retained routes while still delivering delayed prior-route events.

### DIFF
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -174,6 +174,9 @@ versioning through the workspace version in the root `Cargo.toml`.
 
 ### Fixed
 
+- `wn relays add/remove --type nip65` now round-trips the complete directional NIP-65 list instead of deleting
+  read-only relays or flattening existing role markers. Relay-list JSON adds `read_relays` and `write_relays` alongside
+  the existing write-target `relays` view.
 - Release builds now fail app startup with a clear error when `WN_DEV_SETTLEMENT_QUIESCENCE_MS` is set instead of
   allowing it to override the protocol-pinned convergence window. Debug builds retain the override for local
   development and integration tests.
--- a/crates/cli/src/commands/account.rs
+++ b/crates/cli/src/commands/account.rs
@@ -237,10 +237,14 @@ fn missing_relay_list_status(missing: Vec<MissingRelayListKind>) -> AccountRelay
         nip65: marmot_app::AccountRelayListState {
             kind: 10002,
             relays: Vec::new(),
+            read_relays: Vec::new(),
+            write_relays: Vec::new(),
         },
         inbox: marmot_app::AccountRelayListState {
             kind: 10050,
             relays: Vec::new(),
+            read_relays: Vec::new(),
+            write_relays: Vec::new(),
         },
     }
 }
--- a/crates/cli/src/commands/relays.rs
+++ b/crates/cli/src/commands/relays.rs
@@ -101,30 +101,40 @@ async fn update_relay_list(
                 &source_relays,
             )
         })?;
-    let mut relays = relays_for_type(&status, Some(&relay_type))?;
-    if add {
-        if !relays.contains(&url) {
-            relays.push(url.clone());
-        }
-    } else {
-        relays.retain(|relay| relay != &url);
-    }
-    relays.sort();
-    relays.dedup();
-    let publish_relays = relay_endpoints(relays.clone())?;
     let bootstrap = explicit_bootstrap
         .or_else(|| source_relays.first().map(|endpoint| endpoint.0.clone()))
-        .or_else(|| relays.first().cloned())
+        .or_else(|| {
+            relays_for_type(&status, Some(&relay_type))
+                .ok()
+                .and_then(|relays| relays.first().cloned())
+        })
         .ok_or(WnError::MissingRelay)?;
     let bootstrap_relays = vec![TransportEndpoint(bootstrap)];
-    let status = runtime
-        .publish_account_relay_list_kind(
-            &account.label,
-            &relay_type,
-            publish_relays,
-            bootstrap_relays,
-        )
-        .await?;
+    let status = if relay_type == "nip65" {
+        let (mut read_relays, mut write_relays) = nip65_relay_roles(&status.nip65);
+        update_relay_values(&mut read_relays, &url, add);
+        update_relay_values(&mut write_relays, &url, add);
+        runtime
+            .publish_account_nip65_relay_set(
+                &account.label,
+                relay_endpoints(read_relays)?,
+                relay_endpoints(write_relays)?,
+                bootstrap_relays,
+            )
+            .await?
+    } else {
+        let mut relays = relays_for_type(&status, Some(&relay_type))?;
+        update_relay_values(&mut relays, &url, add);
+        runtime
+            .publish_account_relay_list_kind(
+                &account.label,
+                &relay_type,
+                relay_endpoints(relays)?,
+                bootstrap_relays,
+            )
+            .await?
+    };
+    let relays = relays_for_type(&status, Some(&relay_type))?;
     Ok(CommandOutput {
         plain: relays.join("\n"),
         json: json!({
@@ -135,6 +145,25 @@ async fn update_relay_list(
             "relay_lists": relay_lists_json(status),
         }),
     })
+}
+
+fn nip65_relay_roles(state: &marmot_app::AccountRelayListState) -> (Vec<String>, Vec<String>) {
+    if state.read_relays.is_empty() && state.write_relays.is_empty() {
+        return (state.relays.clone(), state.relays.clone());
+    }
+    (state.read_relays.clone(), state.write_relays.clone())
+}
+
+fn update_relay_values(relays: &mut Vec<String>, url: &str, add: bool) {
+    if add {
+        if !relays.iter().any(|relay| relay == url) {
+            relays.push(url.to_owned());
+        }
+    } else {
+        relays.retain(|relay| relay != url);
+    }
+    relays.sort();
+    relays.dedup();
 }
 
 fn relays_for_type(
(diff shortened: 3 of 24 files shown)


erskingardner (commented):
# Code Review: PR #1094 — Persist prior Nostr group routes

## Summary

This PR persists prior signed Nostr routing addresses per group (with a last-observed epoch), retains
the still-live prior route's subscription alongside the current one, and forces a full (uncapped)
relay backfill for any group that carries a retained prior route so a delayed event on the old address
isn't clipped by the account-global sync cursor. Storage gains migration `0030_prior_nostr_routes`.
The `AppGroupRecord` epoch-tracking/rotation/dedup logic is well tested (rotation, rotate-back dedup,
relay-only rotation, and rollback), and the transport-adapter test exercises the actual restart +
delayed-event scenario end to end. The core mechanism is sound, but the full-backfill trigger and its
pruning trigger interact in a way that can force **unbounded, indefinite** full-history resync of a
group's *current* route on every app restart if that group goes quiet right after a rotation — worth
addressing or at least explicitly accepting before merge.

## Issues

### Correctness/Efficiency: restart full-backfill blankets the *current* route too, and pruning has no time-based floor

**crates/transport-nostr-adapter/src/lib.rs:546-553** (`activate_account`) and
**crates/transport-nostr-adapter/src/lib.rs:1246-1257** (`diff_group_subscriptions`)

> ```rust
> let full_backfill_groups = groups_requiring_full_backfill(&activation.group_subscriptions);
> for group in &activation.group_subscriptions {
>     let since = if full_backfill_groups.contains(&group.group_id) {
>         None
>     } else {
>         activation.since
>     };
>     issued.push(group_subscription(&account_id, group, since));
> }
> ```

`groups_requiring_full_backfill` (crates/transport-nostr-adapter/src/lib.rs:1286) flags a group as soon
as it has *more than one* subscription (current + at least one retained prior route). Every subscription
issued for that group_id — **including the current route's own subscription**, which was almost
certainly already fully synced — gets `since = None` on `activate_account`. That's a full relay history
redownload for the current route, not just the newly-retained prior one, on every app restart, for as
long as the group carries a retained prior route.

Pruning (`AppGroupRecord::prune_prior_nostr_routes`, crates/marmot-app/src/groups.rs:484) is purely
epoch-driven: a prior route is retired once `current_epoch - route.last_epoch > retained_depth` (5, from
`CanonicalizationPolicy::default()`). If a group rotates its route and then goes dormant — no further
commits/epoch advancement — the prior route (and therefore the `full_backfill_groups` flag) never
expires, because there is no epoch progress to age it out. There is no wall-clock/time-based fallback.

Net effect: a group that rotates once and then sees no further activity will force a full-history
relay resync of its *current* subscription on every subsequent app restart, indefinitely — not just
until the retention window naturally elapses, but potentially forever for a genuinely quiet group.
This directly undercuts the PR's own stated goal ("Groups without a retained prior route keep the
existing bounded `since` behavior") for what should be the common case (a routed-but-idle group), and
is a real bandwidth/relay-load cost with no upper bound.

Two independent fixes would close this:
1. Scope `since = None` to only the *newly retained* prior-route subscription(s), not the group's
   current route — `diff_group_subscriptions` already has the machinery to distinguish "already
   subscribed" (via `route_key`) from "new" entries; `activate_account` could do the same by tracking
   which subscriptions are prior-route vs current explicitly instead of inferring from a bare count.
2. Add a time-bound (or restart-count-bound) floor to `prune_prior_nostr_routes` independent of epoch
   advancement, so a permanently dormant group's prior route still ages out eventually.

Either alone would bound the cost; right now neither exists.

## Suggestions

### Inconsistent fallback semantics for the same underlying check

**crates/marmot-app/src/client/mod.rs:2152-2156**

> ```rust
> if !self
>     .runtime
>     .has_pending_convergence_inputs(&group_id)
>     .unwrap_or(true)
>     && let Ok(group_record) = self.runtime.group_record(&group_id)
> {
>     group.prune_prior_nostr_routes(group_record.epoch.0);
> }
> ```

This calls `self.runtime.has_pending_convergence_inputs(...)` directly and defaults to `true` (assume
pending, don't prune) on error — appropriately conservative for this use. But `AppClient` already has a
wrapper for the exact same call, `has_pending_convergence_inputs` (crates/marmot-app/src/client/sync.rs:27-31),
which defaults to `false` on error. Two call sites now silently disagree on what "unknown" means for the
same underlying signal. Consider either reusing the wrapper (with its default changed to match, if `true`
is genuinely the safer default) or naming the two differently so the diverging defaults are obviously
intentional rather than an oversight.

### Full-backfill detection is an implicit structural inference

**crates/transport-nostr-adapter/src/lib.rs:1286**

> ```rust
> fn groups_requiring_full_backfill(groups: &[TransportGroupSubscription]) -> HashSet<GroupId> {
>     let mut seen = HashSet::new();
>     let mut repeated = HashSet::new();
>     for group in groups {
>         if !seen.insert(group.group_id.clone()) {
>             repeated.insert(group.group_id.clone());
>         }
>     }
>     repeated
> }
> ```

"This group needs a full backfill" is inferred purely from "more than one subscription shares this
`group_id`." Today that's only ever true because of retained prior routes, so it's correct — but it's an
implicit coupling between routing *structure* and backfill *semantics*. If another feature ever has a
legitimate reason to register a second subscription for the same MLS group (e.g. some future dual-route
scenario unrelated to prior-route retention), it would silently inherit full-backfill behavior. An
explicit `is_prior_route: bool` (or similar) on `TransportGroupSubscription`/the activation DTO would make
the intent self-documenting and immune to that kind of accidental coupling.

### Hardcoded default policy in the retention window computation

**crates/marmot-app/src/groups.rs:485**

> ```rust
> pub(crate) fn prune_prior_nostr_routes(&mut self, current_epoch: u64) -> bool {
>     let policy = CanonicalizationPolicy::default();
>     let retained_depth = policy
>         .convergence
>         .max_rewind_commits
>         .max(policy.app_message_past_epoch_limit);
> ```

This constructs a fresh `CanonicalizationPolicy::default()` rather than reading whatever policy the
running session is actually configured with. Today that's harmless — `crates/marmot-app/src/lib.rs:2087`
only ever overrides `settlement_quiescence_ms` for a dev/test build, per its own comment ("Production
uses the protocol-pinned convergence policy... never shipped"), so `max_rewind_commits` and
`app_message_past_epoch_limit` always match production reality. But it's an implicit assumption living
far from its guarantee: if the dev-override surface ever grows to touch either of those two fields, this
function would silently retain (or discard) prior routes using the wrong window with no compile-time or
test signal. Worth a comment cross-referencing that constraint, or deriving `retained_depth` from the
actual configured policy if it's reasonably reachable from here.

## What's Done Well

- The `AppGroupRecord` rotation/epoch-tracking test suite (crates/marmot-app/src/groups.rs, `prior_nostr_route_tests`)
  is genuinely adversarial: it separately covers ordinary rotation, relay-only rotation on the same
  `nostr_group_id`, rotate-back deduplication, and — notably — rollback/fork recovery correctly retaining
  the losing branch's real epoch range via the `.max(self.nostr_routing_last_epoch)` floor. That rollback
  case is easy to get wrong and the test pins the exact behavior.
- `transport-nostr-adapter/tests/inbound_routing.rs::restart_with_retained_route_backfills_and_routes_delayed_old_event`
  is a real end-to-end regression test for the motivating scenario (delayed event on a retained address
  after restart), not just a unit test of the helper function.
- `AppTransportRouting::replace_group_routes` normalizes and compares before mutating
  (crates/marmot-app/src/lib.rs), so the change-detection/no-op path is order-insensitive and doesn't
  cause spurious resync thrash from `prior_nostr_routes` reordering.
- The SQLite migration and storage round-trip are covered together (`account_projection_state_roundtrips_groups_components_and_seen_events`,
  plus a dedicated migration-defaults test), and the upsert's `WHERE ... OR account_groups.nostr_routing_last_epoch IS NOT excluded...`
  change-detection clause was correctly extended for the two new columns.
- `AppPriorNostrRoute::subscription` validates the transport group id length and non-empty relay set with
  the same rigor as the existing current-route path, rather than trusting persisted data.

PR #1095, opened 2026-07-24
Prefer encrypted Blossom for negotiated group images

erskingardner:
## What changed

- Advertise every group component implemented by `marmot-app`, including encrypted Blossom images, URL avatars, and message retention.
- Refresh cached and relay KeyPackages that are current-profile but under-advertise the app's supported components.
- Add an engine/session/account creation seam for initial GroupContext component state that remains optional instead of becoming a membership requirement.
- Add encrypted-Blossom-first image negotiation during group creation: use Blossom when all founding members support it, fall back to a source URL when they only support URL avatars, and otherwise create the group without an avatar.
- Add additive runtime and UniFFI APIs for hosts to supply downloaded image bytes, media type, and an optional source-URL fallback.
- Redact plaintext bytes and source URLs from Debug output.

## Why

The app implemented both group-image components but did not advertise either one in generated KeyPackages. Its creation request also treated every supplied component state as required, which would make an optional avatar feature a permanent membership gate. This separates support advertisement, optional state, and group requirements while keeping encrypted Blossom as the preferred application behavior.

## Client impact

Existing `create_group` calls remain source-compatible. Clients that want an initial image can adopt the new `create_group_with_initial_image` API and provide already-downloaded bytes; web discovery and download remain host responsibilities.

The capability-refresh marker intentionally causes a one-time relay rescan and replacement of under-advertised KeyPackages after upgrade. Client rollouts should account for the resulting KeyPackage publication traffic and may be staged where relay capacity warrants it.

## Client adoption checklist

- [ ] Update to the MarmotKit/MDK release containing this change and regenerate Swift/Kotlin UniFFI bindings.
- [ ] In the create-group UI, let the host search for or select an image and download it before calling MDK.
- [ ] Pass the downloaded bytes, media type, source URL, and optional render hints through `create_group_with_initial_image`.
- [ ] Continue using the existing `create_group` API when no initial image is selected.
- [ ] For existing-group avatar changes, make `update_group_image` the normal path: download the selected web image, then pass its bytes to MDK. Retain URL-avatar support for interoperability.
- [ ] Render both image-component projections and apply the protocol precedence rule when both are present: the URL avatar wins.
- [ ] Ensure normal account startup and KeyPackage publication run after adoption so under-advertised cached KeyPackages are automatically replaced.
- [ ] Surface image download and Blossom upload failures before presenting group creation as successful. A Blossom upload failure aborts creation; it does not silently fall back to the URL.
- [ ] Verify the negotiated outcomes in client tests: encrypted Blossom when universally supported, URL fallback when Blossom is unavailable but URL avatars are supported, and no avatar without blocking group creation when neither is universal.

MDK owns encryption, Blossom upload, and founding-member negotiation. Web discovery, image download, cropping/resizing, MIME selection, licensing, and content moderation remain host responsibilities.

## Validation

- `just fast-ci`
- Full `cargo test -p marmot-app` suite
- `cargo test -p cgka-session`
- Focused engine capability and app negotiation regression tests
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Create groups with an optional initial image, supporting encrypted image data, avatar URLs, and image metadata.
  * Added public APIs for supplying initial group images across application and UniFFI interfaces.
  * Optional initial components can be included without becoming membership requirements.

* **Bug Fixes**
  * Improved capability validation to reject unsupported optional components early.
  * Key packages now accurately reflect supported components and are refreshed when capabilities change.

* **Tests**
  * Added coverage for initial image selection, capability matching, and optional component behavior.

### DIFF
--- a/crates/cgka-engine/src/app_components.rs
+++ b/crates/cgka-engine/src/app_components.rs
@@ -51,6 +51,7 @@ pub(crate) struct InitialComponentState {
     pub(crate) description: String,
     pub(crate) admins: Vec<[u8; 32]>,
     pub(crate) app_components: Vec<AppComponentData>,
+    pub(crate) optional_app_components: Vec<AppComponentData>,
 }
 
 pub(crate) fn leaf_app_components_extension(
@@ -134,6 +135,15 @@ pub(crate) fn app_data_dictionary_extension_for_group(
         if required.contains(component.component_id) {
             dict.insert(component.component_id, component.data.clone());
         }
+    }
+    for component in &initial.optional_app_components {
+        if !seen_initial.insert(component.component_id) {
+            return Err(EngineError::Other(
+                "group creation request contains duplicate app components".into(),
+            ));
+        }
+        validate_initial_app_component(component)?;
+        dict.insert(component.component_id, component.data.clone());
     }
     Ok(Extension::AppDataDictionary(
         AppDataDictionaryExtension::new(dict),
@@ -1668,6 +1678,7 @@ mod tests {
             description: "desc".to_string(),
             admins: vec![[7u8; 32]],
             app_components: Vec::new(),
+            optional_app_components: Vec::new(),
         };
         let ext = app_data_dictionary_extension_for_group(&required, &initial)
             .expect("creation dictionary should build");
@@ -1752,6 +1763,7 @@ mod tests {
             description: "desc".to_string(),
             admins: vec![[7u8; 32]],
             app_components: Vec::new(),
+            optional_app_components: Vec::new(),
         };
 
         let ext = app_data_dictionary_extension_for_group(&required, &initial)
--- a/crates/cgka-engine/src/engine.rs
+++ b/crates/cgka-engine/src/engine.rs
@@ -547,6 +547,16 @@ impl<S: StorageProvider> Engine<S> {
         req: CreateGroupRequest,
         context: Option<AuditEventContext>,
     ) -> Result<(GroupId, SendResult), EngineError> {
+        self.create_group_with_optional_app_components_and_audit_context(req, Vec::new(), context)
+            .await
+    }
+
+    pub async fn create_group_with_optional_app_components_and_audit_context(
+        &mut self,
+        req: CreateGroupRequest,
+        optional_app_components: Vec<cgka_traits::app_components::AppComponentData>,
+        context: Option<AuditEventContext>,
+    ) -> Result<(GroupId, SendResult), EngineError> {
         let operation_id = self.next_audit_operation_id();
         let mut context = context.unwrap_or_default();
         context.operation_id = Some(operation_id);
@@ -557,12 +567,13 @@ impl<S: StorageProvider> Engine<S> {
             AuditEventKind::CreateGroupEntry {
                 member_count: req.members.len() as u64,
                 required_feature_count: req.required_features.len() as u64,
-                app_component_count: req.app_components.len() as u64,
+                app_component_count: (req.app_components.len() + optional_app_components.len())
+                    as u64,
                 initial_admin_count: req.initial_admins.len() as u64,
             },
         );
         self.current_audit_context = Some(context.clone());
-        let result = self.do_create_group(req).await;
+        let result = self.do_create_group(req, optional_app_components).await;
         match &result {
             Ok((group_id, send_result)) => {
                 let mut outcome_context = context.clone();
@@ -1850,6 +1861,19 @@ impl<S: StorageProvider + 'static> CgkaEngine for Engine<S> {
         req: CreateGroupRequest,
     ) -> Result<(GroupId, SendResult), EngineError> {
         self.create_group_with_audit_context(req, None).await
+    }
+
+    async fn create_group_with_optional_app_components(
+        &mut self,
+        req: CreateGroupRequest,
+        optional_app_components: Vec<cgka_traits::app_components::AppComponentData>,
+    ) -> Result<(GroupId, SendResult), EngineError> {
+        self.create_group_with_optional_app_components_and_audit_context(
+            req,
+            optional_app_components,
+            None,
+        )
+        .await
     }
 
     async fn join_welcome(
--- a/crates/cgka-engine/src/group_lifecycle.rs
+++ b/crates/cgka-engine/src/group_lifecycle.rs
@@ -1128,27 +1128,20 @@ impl<S: StorageProvider> Engine<S> {
         &self,
         key_packages: &[cgka_traits::engine::KeyPackage],
     ) -> Result<GroupCapabilities, EngineError> {
-        if key_packages.is_empty() {
-            return Ok(leaf_capabilities_as_marmot(
-                &self.registry,
-                self.ciphersuite,
-                &self.supported_app_components,
-                self.new_protocol_profile,
-            ));
-        }
-        let mut it = key_packages.iter();
-        let first_input = it.next().unwrap();
-        let first_profile = first_input.protocol_profile;
-        let first = self.parse_key_package(first_input)?;
-        let mut acc = capabilities_of_key_package(&first);
-        for kp in it {
-            let parsed = self.parse_key_package(kp)?;
-            if kp.protocol_profile != first_profile {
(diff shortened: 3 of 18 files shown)


erskingardner (commented):
# Code Review: PR #1095 — Prefer encrypted Blossom for negotiated group images

## Summary

This PR (1) makes `marmot-app` advertise every group component it actually implements (encrypted Blossom image,
URL-avatar fallback, message retention) in generated KeyPackages, (2) adds a refresh path that republishes
cached/relay KeyPackages that under-advertise the app's current supported set, (3) plumbs an
"optional initial app-component state" seam through `cgka-engine` → `cgka-session` → `marmot-account` →
`marmot-app` → `marmot-uniffi` so a founding avatar can be attached without becoming a permanent membership
requirement, and (4) adds `create_group_with_initial_image`, which prefers encrypted Blossom, falls back to a
source-URL avatar, or omits the avatar, based on what the founding members can construct. The engine-side duplicate
detection and mandatory-vs-optional separation are implemented carefully (shared `seen_initial` set spanning both
`app_components` and `optional_app_components`, explicit `validate_initial_app_component` rejection of
engine-owned component ids in the optional list too). Overall this is a well-scoped, well-tested change; the issues
below are mostly about a latent self-capability gap in the new negotiation helper and missing negative-path test
coverage for the new engine guard.

## Issues

### Correctness: `preferred_initial_group_image_component` never checks the creator's own capability

**crates/marmot-app/src/client/mod.rs:317-349**

> ```rust
> let constructable = self.runtime.constructable_capabilities(&members)?;
> ...
> match preferred_initial_group_image_component(&constructable, image.source_url.is_some()) {
>     Some(GROUP_BLOSSOM_IMAGE_COMPONENT_ID) => { ... }
> ```

`constructable_capabilities(&members)` (`crates/cgka-engine/src/group_lifecycle.rs::do_constructable_capabilities`)
only intersects the *invitees'* KeyPackage capabilities when `members` is non-empty — it does **not** fold in the
creator's own `supported_app_components` (that only happens on the `key_packages.is_empty()` branch, i.e. solo
groups). So for any group with at least one invitee, `preferred_initial_group_image_component` can select
`GROUP_BLOSSOM_IMAGE_COMPONENT_ID` purely because the invitees support it, without ever confirming the *creator*
does.

Today that's masked because `MarmotApp::supported_app_component_ids()` (`crates/marmot-app/src/lib.rs:3580-3593`)
unconditionally includes both `GROUP_BLOSSOM_IMAGE_COMPONENT_ID` and `GROUP_AVATAR_URL_COMPONENT_ID`, and that same
function seeds the engine's `supported_app_components` at session construction
(`crates/marmot-app/src/lib.rs:2082`). But the new engine guard this PR adds —

**crates/cgka-engine/src/group_lifecycle.rs:208-225**

> ```rust
> let self_optional_missing = optional_ids.missing_from(&self_supported_components);
> if !self_optional_missing.is_empty() {
>     return Err(EngineError::MissingRequiredCapabilities { ... });
> }
> ```

— hard-fails the *entire* `create_group_with_initial_image` call (including every invitee) if the creator ever
doesn't support the chosen optional component, rather than degrading to "create the group without an avatar" the
way the PR's stated design intends. If `supported_app_component_ids()` and the engine's advertised
`supported_app_components` ever diverge (build variant, staged rollout, future refactor that stops always seeding
both from the same function), a founder who correctly can't construct Blossom images would fail to create *any*
group with an initial image instead of falling back to the URL/no-image path `preferred_initial_group_image_component`
was written to provide.

Suggest either (a) having `constructable_capabilities` always intersect with `self.supported_app_components`
regardless of whether `key_packages` is empty, or (b) intersecting `constructable` with
`self.app.supported_app_component_ids()` explicitly at the call site before calling
`preferred_initial_group_image_component`, so the fallback logic is provably correct rather than correct-by-current-
coincidence.

## Suggestions

### Test coverage: creator-side optional-capability rejection is untested

**crates/cgka-engine/tests/capabilities.rs:294** (new test `optional_initial_component_state_is_not_a_membership_requirement`)

The one new test covers the happy path (an invitee lacking support for the optional component can still join).
There's no test for the new `self_optional_missing` branch in `do_create_group`
(`crates/cgka-engine/src/group_lifecycle.rs:208-225`) — i.e. the *creator* requesting optional state for a
component they themselves don't support. Given this is a brand-new error path, a regression test asserting
`create_group_with_optional_app_components` returns `MissingRequiredCapabilities` when self lacks the component
would be cheap insurance.

### Test coverage: duplicate component id across mandatory and optional lists

**crates/cgka-engine/src/app_components.rs** (`app_data_dictionary_extension_for_group`, shared `seen_initial` set)

The function correctly shares one `seen_initial: BTreeSet` across the `initial.app_components` and
`initial.optional_app_components` loops, so a component id repeated between the two lists is rejected as
"group creation request contains duplicate app components." That's the right behavior, but it isn't exercised by
any test in this PR — worth a quick unit test asserting the cross-list duplicate is caught, not just same-list
duplicates (which existing tests already cover for `app_components` alone).

### Robustness: Blossom upload failure aborts the whole group creation

**crates/marmot-app/src/client/mod.rs:324-337**

> ```rust
> Some(GROUP_BLOSSOM_IMAGE_COMPONENT_ID) => {
>     let upload = upload_group_image(&image.plaintext, &image.media_type, None).await?;
> ```

A transient Blossom upload failure propagates via `?` and aborts `create_group_with_initial_image` entirely —
including the member invites — even though the image was explicitly designed to be optional/non-blocking. This
matches the existing `update_group_image` post-creation path's behavior
(`crates/marmot-app/src/client/mod.rs:1595-1620`), so it's a consistent convention rather than a new regression, but
given this PR's stated goal ("optional instead of becoming a permanent membership gate"), it may be worth an
explicit product decision about whether upload failure should also degrade to "create the group without an avatar"
instead of failing outright. Flagging for awareness rather than as a blocker.

### Operational note: KeyPackage-refresh marker rename re-triggers a full relay rescan

**crates/marmot-app/src/lib.rs** (`key_package_cutover_replacement_pending_path` /
`key_package_cutover_scan_complete_path`)

The on-disk marker filenames changed from `{label}.strict-cutover-*` to
`{label}.capability-refresh-v1-*`. Since these are plain filenames (not versioned data), any device that already
completed the earlier "strict cutover" relay scan will not find a `capability-refresh-v1-relay-scan-complete`
marker and will re-run the full relay scan/replacement flow on next sync. That looks like the intended mechanism
for making every existing device re-check its cached/relay KeyPackages against the newly expanded supported-component
set (blossom image, retention, avatar URL) — which is reasonable — but it's an operational side effect (one-time
relay rescan for every existing user) that isn't called out in the PR's "Client impact" section. Worth a one-line
mention there so downstream consumers aren't surprised by the extra relay traffic on upgrade.

## What's Done Well

- The mandatory/optional split in `app_data_dictionary_extension_for_group` is careful: the same `seen_initial` set
  guards against a component id appearing in both `app_components` and `optional_app_components`, and
  `validate_initial_app_component` is called uniformly on both, so an optional-state request can't sneak in an
  engine-owned component id (profile, admin policy, app_components list, account identity proof) that the mandatory
  path already blocks.
- Explicitly keeping optional component ids out of `required_caps.app_components` (so per-invitee negotiation never
  even inspects them) is the right way to implement "presentation state that isn't a membership gate" — invitees
  genuinely never need to prove support.
- The `AppInitialGroupImage` / `InitialGroupImageFfi` `Debug` impls both redact `plaintext` (length only) and
  `source_url` (presence only) instead of the actual bytes/URL, consistent with the repo's privacy-safe logging
  invariant.
- `supported_app_component_ids()` and the new `key_package_metadata_matches_current_support()` closing the loop
  (advertise everything actually implemented, then detect and refresh KeyPackages that under-advertise it) directly
  addresses the bug described in the PR's "Why" section rather than papering over it.
- Existing call sites (`create_group`, `prepare_create_group_with_audit_context`,
  `create_group_with_audit_context`) are preserved as thin wrappers over the new `_with_optional_app_components_`
  variants, so this is source-compatible for every existing caller across `cgka-engine` → `cgka-session` →
  `marmot-account` → `marmot-uniffi`, exactly as the PR body claims.

PR #1096, opened 2026-07-24
Omit unrepresentable Nostr expiration hints

### LINKED ISSUE #1070: Retention overflow rejects valid outbound messages instead of omitting expiry hint
## Bug

The adopted retained-history rule treats transport expiration as a hint: when `t_source + Δ_source` is not representable, the expiry is undefined, the expiration tag is omitted, and the message remains valid.

The current outbound Nostr path does the opposite:

- `GroupMessageMetadata::expiration_timestamp` returns `ExpirationTimestampOverflow` when `message_timestamp + retention_secs` overflows (`crates/traits/src/peeler.rs`).
- `transport-nostr-peeler` propagates that as `WrapFailed`.
- `group_wrap_rejects_overflowing_expiration_metadata` explicitly locks in the fail-closed behavior (`crates/transport-nostr-peeler/src/peeler.rs:716-736`).

This makes a valid application message unsendable solely because its optional transport expiry hint cannot be represented. It also disagrees with the local projection helper, which already preserves the `u64` retention decision and omits the unrepresentable expiry.

## Reproduction

```sh
cargo test --locked -p transport-nostr-peeler overflow
```

The passing test demonstrates the current incorrect contract: overflow rejects the wrapped message.

## Expected behavior

Use checked addition to decide whether to add the expiration tag. On overflow, omit the tag and continue wrapping/sending the message. Update the existing test to assert a valid wrapped event with no expiration tag.

## Scope

Found while reviewing #1068 / #1055. This is pre-existing transport-hint behavior, not the source-epoch pinning/storage change in that PR, so it should be fixed separately.

erskingardner:
## Summary

- keep transport expiration metadata optional when `message_timestamp + retention_seconds` overflows
- omit the Nostr `expiration` tag instead of rejecting an otherwise valid outbound message
- replace the fail-closed regression test with coverage for a successfully wrapped event without the hint
- simplify the now-infallible metadata helper to return `Option<u64>` directly, removing the unreachable error type and dead wrapper error path

## API note

`GroupMessageMetadata::expiration_timestamp` now returns `Option<u64>` instead of `Result<Option<u64>, GroupMessageMetadataError>`, and the unreachable public `GroupMessageMetadataError` type is removed.

## Validation

- `cargo test -p cgka-traits -p transport-nostr-peeler`
- `just fast-ci`

Closes #1070

## Summary by CodeRabbit

* **Bug Fixes**
  * Group messages with unrepresentable expiration times are now wrapped successfully instead of failing.
  * Expiration hints are omitted when they cannot be represented, preventing invalid expiration data from interrupting message processing.
  * Updated behavior ensures wrapped events remain valid for maximum timestamp values.

### DIFF
--- a/crates/traits/src/lib.rs
+++ b/crates/traits/src/lib.rs
@@ -80,7 +80,7 @@ pub use ingest::{
     IngestOutcome, PeeledContent, PeeledMessage, ProposalRejectionCategory, StaleReason,
 };
 pub use message::{MessageRecord, MessageState, OwnCommitConvergenceStamp, StoredMessagePayload};
-pub use peeler::{GroupMessageMetadata, GroupMessageMetadataError, TransportPeeler};
+pub use peeler::{GroupMessageMetadata, TransportPeeler};
 pub use storage::{
     CapabilityStorage, GroupStorage, LeaveRequest, LeaveRequestStorage, MessageStorage,
     OutboundIntentStorage, QueuedOutboundIntent, StorageError, StorageProvider, StorageResult,
--- a/crates/traits/src/peeler.rs
+++ b/crates/traits/src/peeler.rs
@@ -79,10 +79,10 @@ impl GroupMessageMetadata {
         if *retention_seconds == 0 {
             return Ok(None);
         }
-        inner_created_at
-            .checked_add(*retention_seconds)
-            .map(Some)
-            .ok_or(GroupMessageMetadataError::ExpirationTimestampOverflow)
+        // Expiration is only a transport hint. If the sum is not
+        // representable, the hint is undefined and must be omitted without
+        // rejecting the otherwise-valid application message.
+        Ok(inner_created_at.checked_add(*retention_seconds))
     }
 }
 
--- a/crates/traits/src/peeler.rs
+++ b/crates/traits/src/peeler.rs
@@ -65,30 +65,25 @@ impl GroupMessageMetadata {
     }
 
     /// Compute the transport-level expiration timestamp, if any.
-    pub fn expiration_timestamp(&self) -> Result<Option<u64>, GroupMessageMetadataError> {
+    pub fn expiration_timestamp(&self) -> Option<u64> {
         let Self::Application {
             inner_created_at,
             retention_seconds,
         } = self
         else {
-            return Ok(None);
+            return None;
         };
         let Some(retention_seconds) = retention_seconds else {
-            return Ok(None);
+            return None;
         };
         if *retention_seconds == 0 {
-            return Ok(None);
+            return None;
         }
         // Expiration is only a transport hint. If the sum is not
         // representable, the hint is undefined and must be omitted without
         // rejecting the otherwise-valid application message.
-        Ok(inner_created_at.checked_add(*retention_seconds))
+        inner_created_at.checked_add(*retention_seconds)
     }
-}
-
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub enum GroupMessageMetadataError {
-    ExpirationTimestampOverflow,
 }
 
 /// Unwrap and rewrap transport-layer envelopes. A single peeler typically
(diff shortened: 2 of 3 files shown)


erskingardner (commented):
## Adversarial review

Tried to break this from a few angles — semantics, callers, downstream consumers, and the test itself. **No blocking issues found.** The change is correct, minimal, and matches the accepted resolution of #1070.

### What was checked

- **Semantics of omission vs. clamping.** An overflowing `inner_created_at + retention_seconds` denotes an expiration beyond representable time, so omitting the relay hint is semantically equivalent to "never expires" — a `saturating_add` clamp to `u64::MAX` would communicate the same thing with extra noise. Omission is the right call, and the new inline comment states the "hint, not a hard requirement" rationale clearly.
- **Consistency with local retention.** `AppMessageRetentionDecision::new` (`crates/traits/src/app_event.rs`) already treats overflow as a known decision with no finite `expires_at` ("must never be pruned by a timestamp sweep"). Before this PR the transport path fail-closed while the local projection fail-open — this PR removes that disagreement rather than introducing one. 👍
- **Callers.** Full-workspace grep confirms `expiration_timestamp` has exactly one caller (`crates/transport-nostr-peeler/src/peeler.rs` wrap path); no other code matches on the error variant, so nothing else changes behavior.
- **Test quality.** `group_wrap_omits_unrepresentable_expiration_metadata` round-trips the wrapped event and asserts both `created_at == u64::MAX` is preserved and the `expiration` tag is genuinely absent — it verifies the observable contract, not just "no error."
- **Verified locally:** `cargo test -p cgka-traits -p transport-nostr-peeler` (55 tests pass) and `cargo clippy -p cgka-traits -p transport-nostr-peeler --all-targets` (clean).

### Non-blocking finding: `ExpirationTimestampOverflow` is now unreachable

After this change, `GroupMessageMetadata::expiration_timestamp` always returns `Ok(..)` — nothing in the workspace constructs `GroupMessageMetadataError::ExpirationTimestampOverflow` anymore, and the deleted test was the only thing exercising it. The signature still advertises fallibility, and the sole call site still carries a dead `.map_err(... WrapFailed ...)` branch that can never fire.

Two options:

1. Simplify `expiration_timestamp` to return `Option<u64>` and drop `GroupMessageMetadataError` plus the dead `map_err` at the call site. Note this is a public-API change — the error type is re-exported from `cgka_traits` (`crates/traits/src/lib.rs`) — so it belongs in the changelog if done.
2. If the `Result` shape is being kept deliberately as room for future fallible metadata checks, a one-line comment saying so would stop the next reader from wondering why an error type with no live constructor exists.

Either is fine as a follow-up; this doesn't need to block the merge.

agent-p1p:
## Pip adversarial review (retrospective)

Reviewed the final PR head `6e3152be72bf516e1c3b989c9452be7fb7353827`. This task arrived after the PR had already merged; the verdict below records the delayed independent review.

BLOCKING:
- none

SUGGESTION:
- none

NITPICK:
- none

SUMMARY: The fix matches #1070. `checked_add` now makes an unrepresentable expiration hint `None`, the Nostr wrapper omits the optional tag without rejecting the message, and the observable regression test verifies both successful wrapping and tag omission. Removing the dead error type and unreachable wrapper error path reduces complexity. Workspace searches found no remaining callers of the removed API. The transport behavior now matches `AppMessageRetentionDecision::new`, which already preserves the retention decision while representing overflow as no finite expiry.

VERDICT: APPROVE

Validation on the exact PR head:
- `cargo test --locked -p transport-nostr-peeler unrepresentable`
- `cargo test --locked -p cgka-traits -p transport-nostr-peeler`
- `cargo clippy --locked -p cgka-traits -p transport-nostr-peeler --all-targets -- -D warnings`

All passed.

SENSITIVE_PATHS: `crates/traits/src/lib.rs`, `crates/traits/src/peeler.rs`, `crates/transport-nostr-peeler/src/peeler.rs`

PR #1097, opened 2026-07-24
Enforce kind 451 push owner proofs by group profile

### LINKED ISSUE #1059: Move push owner proofs to event kind 451 and enforce current-profile push state

erskingardner:
## Summary

- produce the adopted unpublished kind-451 event-shaped proof for push token records and removals
- bind inbound verification to the carrying group profile: current groups accept only kind 451; legacy groups additionally accept transitional kind 450 and deployed raw-digest proofs
- reject negative or more-than-one-hour-future owner timestamps before signature work, cap each kind-446 trigger at 32 encrypted tokens, and split larger per-server batches across valid triggers
- verify every field returned by an external event signer, while retaining the raw `SignedRecord` digest only for ordering and legacy verification
- pin the adopted kind-451 removal event ID/signature vector and profile acceptance matrix

## Normative source

Implemented against `marmot-protocol/marmot@dedeaf429a6f52a4f1c62f3e992be012dab61b7b` (`Specify push owner proof signing`).

## Validation

- `cargo test -p marmot-app`
- `just fast-ci`

`cargo doc -p marmot-app --no-deps` was also checked; it reaches only four pre-existing broken intra-doc links outside this diff.

Closes #1059

### DIFF
--- a/crates/marmot-app/src/notifications.rs
+++ b/crates/marmot-app/src/notifications.rs
@@ -994,6 +994,10 @@ pub fn build_notification_rumor_content(tokens: &[Vec<u8>]) -> Result<String, Ap
     Ok(BASE64_STANDARD.encode(joined))
 }
 
+pub(crate) fn notification_trigger_chunks(tokens: &[Vec<u8>]) -> std::slice::Chunks<'_, Vec<u8>> {
+    tokens.chunks(PUSH_MAX_NOTIFICATION_TRIGGER_TOKENS)
+}
+
 pub async fn build_notification_gift_wrap(
     server_pubkey_hex: &str,
     tokens: &[Vec<u8>],
--- a/crates/marmot-app/src/notifications.rs
+++ b/crates/marmot-app/src/notifications.rs
@@ -28,6 +28,7 @@ use cgka_traits::app_event::{
     EVENT_REF_TAG, MARMOT_APP_EVENT_KIND_AGENT_ACTIVITY, MARMOT_APP_EVENT_KIND_AGENT_OPERATION,
     MARMOT_APP_EVENT_KIND_CHAT, MARMOT_APP_EVENT_KIND_REACTION,
 };
+use cgka_traits::group::ProtocolProfile;
 
 use crate::messages::{PUBKEY_REF_TAG, inline_mention_pubkey_hexes, mention_pubkey_hex};
 use crate::{
@@ -44,6 +45,8 @@ pub const KIND_MARMOT_NOTIFICATION_SERVER_RELAYS: u64 = 10050;
 pub const PUSH_VERSION: &str = "marmot-push-v1";
 pub const PUSH_ENCRYPTED_TOKEN_LEN: usize = 1084;
 const PUSH_MAX_GOSSIP_ENTRIES: usize = 32;
+const PUSH_MAX_NOTIFICATION_TRIGGER_TOKENS: usize = 32;
+const PUSH_OWNER_TS_MAX_FUTURE_MS: i64 = 3_600_000;
 const PUSH_TOKEN_PLAINTEXT_LEN: usize = 1024;
 const PUSH_MAX_PROVIDER_TOKEN_LEN: usize = PUSH_TOKEN_PLAINTEXT_LEN - 3;
 const PUSH_CIPHERTEXT_LEN: usize = PUSH_TOKEN_PLAINTEXT_LEN + 16;
@@ -54,7 +57,8 @@ const PUSH_HKDF_INFO: &[u8] = b"marmot-push-token-encryption";
 /// distinct unpublished Nostr events so a signature cannot be cross-applied
 /// between them and external signers can produce the proof without raw digest
 /// access.
-const PUSH_OWNER_PROOF_EVENT_KIND: u16 = 450;
+const PUSH_OWNER_PROOF_EVENT_KIND: u16 = 451;
+const LEGACY_PUSH_OWNER_PROOF_EVENT_KIND: u16 = 450;
 const PUSH_RECORD_DOMAIN: &str = "marmot-push-token-record-v1";
 const PUSH_REMOVAL_DOMAIN: &str = "marmot-push-token-removal-v1";
 const NOTIFICATION_VERSION_TAG: &str = "v";
@@ -200,7 +204,7 @@ pub struct GroupPushTokenRecord {
     /// ordering primitive. Distinct from `updated_at_ms` (local receive time).
     pub owner_ts: i64,
     /// 64-byte BIP-340 Schnorr signature (128 lowercase hex) by `member_id_hex`
-    /// over the canonical `SignedRecord`. See `push_record_signing_digest`.
+    /// over the exact local-only kind-451 owner-proof event id.
     pub owner_sig: String,
     pub updated_at_ms: i64,
 }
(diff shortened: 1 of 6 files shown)


erskingardner:
@coderabbitai review

crates/marmot-app/src/notifications.rs
```diff
@@ -200,7 +204,7 @@ pub struct GroupPushTokenRecord {
     /// ordering primitive. Distinct from `updated_at_ms` (local receive time).
     pub owner_ts: i64,
     /// 64-byte BIP-340 Schnorr signature (128 lowercase hex) by `member_id_hex`
-    /// over the canonical `SignedRecord`. See `push_record_signing_digest`.
+    /// over the exact local-only kind-451 owner-proof event id.
```
erskingardner: Minor (docs): this replaced the comment that explained the legacy fallback, and now reads as if the kind-451 owner-proof event id is the *only* valid signature shape. But `verify_owner_sig` still accepts a kind-450 event or a raw `SHA-256(SignedRecord)` digest for `ProtocolProfile::Legacy` groups, so this field can validly hold three different signature shapes depending on the producing client and the group's profile. Worth restoring a one-line note about the legacy forms — this is exactly the subtlety someone auditing storage rows or writing a new verifier would want captured next to the field.

crates/marmot-app/src/notifications.rs
```diff
@@ -717,7 +722,13 @@ fn push_owner_sig_from_signed_event(
...
         .id
         .ok_or_else(|| AppError::Publish("push owner proof event id was not set".into()))?;
-    if signed.id != expected_id || signed.pubkey != expected.pubkey {
+    if signed.id != expected_id
+        || signed.pubkey != expected.pubkey
+        || signed.created_at != expected.created_at
+        || signed.kind != expected.kind
+        || signed.tags != expected.tags
+        || signed.content != expected.content
+    {
```
erskingardner, line 734: Minor (redundancy): per NIP-01, the event id is `sha256(pubkey, created_at, kind, tags, content)`, and `signed.verify()` (called right after this block) recomputes that hash from `signed`'s own fields. So once `signed.id == expected_id` holds and `verify()` passes, the remaining fields are cryptographically guaranteed (short of a SHA-256 collision) to equal `expected`'s — the per-field comparisons add no coverage beyond the old `id`/`pubkey` check. Not a bug, and fine as defense-in-depth / clearer fail-fast errors, but a short comment saying that's the intent would save the next reader from re-deriving why the old check was already sufficient.

crates/marmot-app/src/notifications.rs
```diff
@@ -1112,18 +1160,22 @@ where
...
 /// which is what makes offline-member bootstrap safe.
-pub(crate) fn verify_push_gossip(
+pub(crate) fn verify_push_gossip_for_profile(
     action: PushGossipAction,
     group_id_hex: &str,
     active_members: &[String],
+    profile: ProtocolProfile,
 ) -> PushGossipAction {
     let active: HashSet<&str> = active_members.iter().map(String::as_str).collect();
+    let latest_valid_owner_ts = unix_now_ms().saturating_add(PUSH_OWNER_TS_MAX_FUTURE_MS);
```
erskingardner, line 1173: Minor (hardening): the `0..=latest_valid_owner_ts` bound rejects negative and >1h-future timestamps, but still accepts `owner_ts == 0` (1970) and anything else implausibly old, which no honest client would produce. Since this bound is new code introduced specifically to constrain owner timestamps, consider also rejecting values older than any plausible feature epoch. Not exploitable — an old-but-signed record is still authentic and simply loses the `(owner_ts, record_digest)` ordering race to any legitimately newer record — so purely optional.

erskingardner (commented):
Adversarial review of the full PR head (including the follow-up commit `20a2bc59` that chunks oversized kind-446 trigger batches). I checked out the PR ref locally and ran the `marmot-app` test suite, `cargo fmt --check`, and `cargo clippy --all-targets` — all clean, including the pinned kind-451 vector test that locks in the exact event id/signature from the adopted spec.

**Verdict: no blocking issues found.** Three minor, non-blocking suggestions are attached as inline comments.

Things that held up well under adversarial scrutiny:

- The current/legacy verification matrix is exactly right: kind-451 always accepted, transitional kind-450 and raw-digest proofs only for `ProtocolProfile::Legacy`, and `Current` short-circuits with no accidental legacy fallback leakage.
- Test coverage is genuinely adversarial: the profile-matrix tests exercise all four corners (kind-451, kind-450, wrong-kind 452, raw legacy) against both profiles, and `future_or_negative_owner_timestamps_are_rejected_before_signature_work` asserts the signature-verification counter stays at zero, proving the timestamp bound really runs *before* any signature work.
- The follow-up chunking commit closes a real regression the first commit would have introduced: capping `build_notification_rumor_content` at 32 tokens without chunking the `publish_notification_trigger` call site would have silently broken notification delivery for groups with >32 push-token records behind one server. `kind_446_trigger_chunks_split_33_encrypted_tokens` covers it directly.
- The `sync.rs` change sources `active_member_ids` and `profile` from the same `and_then` chain; since `group_record` and `members` gate on the same `ensure_group_live` precondition, requiring `group_metadata` up front rejects nothing that would previously have succeeded.
- The pinned kind-451 vector test calls `verify_owner_sig` directly rather than going through `verify_push_gossip_for_profile`, so it stays independent of the new timestamp bound and won't bit-rot as "now" moves forward.

erskingardner:
@coderabbitai review

PR #1099, opened 2026-07-24
fix(push): make registration lifecycle durable

### LINKED ISSUE #929: Push registration gossip failures are reported as success and have no durable per-group retry
## Summary

`MarmotAppRuntime::upsert_push_registration` persists the device registration and then best-effort shares it to every joined group, but it discards the share result and returns success even when no group publish reached a relay. The client therefore cannot distinguish "registration stored locally" from "token gossiped to peers" and cannot schedule a reliable retry.

This turns transient relay loss into missing background notifications: peers never learn the device token, so their message sends have no token for the notification trigger. It is especially visible on iOS alongside #926, where a silently dead relay socket can make every gossip publish fail while the local upsert still commits.

## Evidence at the current iOS pin

The iOS app is pinned to `8faa6d2`.

In `crates/marmot-app/src/client/push.rs`, `share_push_registration`:

- upserts the local group-token record before publishing;
- catches each `send_app_event` error and only logs it;
- increments `shared` only on success;
- returns `Ok(0)` when every group publish failed;
- updates the account-wide `last_shared_at_ms` when any one group succeeds, so it cannot represent per-group completion.

In `crates/marmot-app/src/runtime/mod.rs`, `upsert_push_registration` then does:

```rust
let registration = self.accounts.app.upsert_push_registration(...)?;
let _ = self.share_push_registration(account_ref).await;
Ok(registration)
```

The returned registration was created before sharing, and the `usize` outcome is discarded. UniFFI therefore reports a successful registration to iOS and Android even when gossip published to zero groups.

`last_shared_at_ms` is exposed through UniFFI but is not a sufficient client-side recovery signal: an older successful share can mask a later all-failed refresh, and one successful group masks failures in every other group.

## Reproduction

1. Start an account with native push enabled and at least one joined group.
2. Make relay publishing unavailable while leaving local storage usable. On iOS this can be reproduced through the dead-socket state described in #926.
3. Call `upsert_push_registration` with a valid provider token.
4. Observe local registration and local group-token projection update successfully.
5. Observe each token-gossip `send_app_event` fail and `shared == 0`.
6. Observe the UniFFI call still return success; the app clears its failure state and has no durable per-group retry to drain.
7. From another member, observe the new device token is absent and message sends produce no push wake for that device.

## Expected

Registration synchronization must not report full success until its group gossip outcome is observable and failed groups remain scheduled for retry.

## Suggested direction

- Return a structured share outcome from the runtime/UniFFI boundary with attempted, succeeded, and failed group counts or identifiers.
- Treat zero successful publishes with eligible groups as an error, or otherwise expose a non-success result the client cannot mistake for completion.
- Persist pending gossip per account and group, and drain it on runtime start, relay reconnection, foreground catch-up, group join, and explicit registration refresh.
- Mark completion per group rather than using only account-wide `last_shared_at_ms`.
- Keep retries idempotent by rebuilding the owner-signed token update from the stored registration.
- Add transport tests for all-failed and partially-failed group publication, runtime restart, and eventual drain after relay recovery.

## Relationship to #926

#926 is the relay-liveness trigger seen on iOS. This issue is the notification-specific correctness gap that hides the resulting publication failure and loses retry intent. Fixing #926 reduces one cause of failure; it does not make registration gossip durable across ordinary transient relay outages or partial group failures.

erskingardner:
## Summary

- add durable per-group outboxes for push-registration updates and removals
- persist exact retired-registration removal intent before clearing, disabling, or rotating registrations
- clear outbox rows only after successful publish, with revision guards against stale completion
- retry pending work on account startup, reconnect, catch-up, group creation/join, and invite acceptance without adding a periodic timer
- return structured complete/pending outcomes through the runtime and UniFFI surfaces
- keep push setting responses non-blocking once the durable mutation has committed

## Root cause

Push registration gossip was previously attempted once after local storage changed. Per-group publish failures were reduced to a count and then forgotten, while clear/disable removed the local registration needed to construct later removal gossip. There was no durable record of which groups still needed an update or removal after a partial failure or restart.

This change records intent transactionally before external relay work and retains it until that exact registration revision publishes successfully. Migration 0031 backfills existing registrations for currently joined groups.

The workspace version intentionally remains at `0.9.6`; the coordinated version bump will happen with the other pending PRs.

## Validation

- `just fast-ci`
- `cargo test -p marmot-app`
- focused partial-failure, restart, removal, migration, revision-guard, and non-blocking-disable regression tests after rebasing onto current master
- `git diff --check`

Fixes #929

## Summary by CodeRabbit

* **New Features**
  * Push-registration sharing now reports completion status and per-group attempted, successful, failed, and pending counts.
  * Registration updates and removals are queued for reliable delivery across groups.
  * Pending push-registration changes automatically retry in the background, including after restart or reconnect.
  * Push-registration APIs now return registration details alongside sharing results.

* **Bug Fixes**
  * Improved handling of push-registration changes when joining, leaving, deleting, or clearing groups.
  * Prevented local push state from updating before sharing succeeds.
  * Improved recovery after partial sharing failures.


erskingardner (commented):
## Adversarial review

The durable outbox direction is right and the revision-guarded completion / non-blocking disable paths are in good shape. Three durability holes still look merge-blocking that can still strand or drop removal/share work after this lands.

**Blockers**
1. Removal outbox PK is `(group, platform, server)` and overwrites on conflict — a later disable/rotation can erase an earlier unsent removal for a different fingerprint. Receivers match on fingerprint, so the old token can stick.
2. Leave / local group wipe deletes pending removals without publishing. Wipe is especially bad because MLS membership is unchanged.
3. No idle drain: after a partial share/removal failure, a healthy long-lived worker may never retry until an unrelated lifecycle event.

Inline notes below.

crates/storage-sqlite/src/migrations/0031_push_registration_gossip_outbox.rs
```diff
@@ -0,0 +1,59 @@
...
+    account_id_hex TEXT NOT NULL,
+    platform INTEGER NOT NULL,
+    token_fingerprint TEXT NOT NULL,
+    server_pubkey_hex TEXT NOT NULL,
+    relay_hint TEXT,
+    registration_created_at_ms INTEGER NOT NULL,
+    registration_updated_at_ms INTEGER NOT NULL,
+    queued_at_ms INTEGER NOT NULL,
+    last_attempted_at_ms INTEGER,
+    PRIMARY KEY (group_id_hex, platform, server_pubkey_hex)
```
erskingardner: This PK is too coarse for the consumption model. Removals are applied by `(member, leaf, platform, server, token_fingerprint)`, but the outbox only retains one row per `(group, platform, server)`.

Concrete failure:
1. Disable token A → removal rows queued for A
2. Re-register token B on the same platform/server → shares for B, A removals still pending
3. Disable B before A's removals publish → `ON CONFLICT DO UPDATE` replaces A with B

Any peer that still holds A never sees an A-fingerprint removal, so the old token remains usable for triggers. Include `token_fingerprint` (and likely `registration_updated_at_ms`) in the primary key, or reject overwriting a different fingerprint/revision.

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -1422,6 +1840,66 @@ impl SqliteAccountStorage {
...
+        "INSERT INTO pending_push_registration_removals (
+            group_id_hex, account_label, account_id_hex, platform,
+            token_fingerprint, server_pubkey_hex, relay_hint,
+            registration_created_at_ms, registration_updated_at_ms,
+            queued_at_ms, last_attempted_at_ms
+         )
+         SELECT group_id_hex, ?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, NULL
+         FROM account_groups
+         WHERE self_membership = 'member'
+         ON CONFLICT(group_id_hex, platform, server_pubkey_hex) DO UPDATE SET
```
erskingardner: This upsert is what makes the coarse PK load-bearing. Because receiver deletes are fingerprint-scoped, clobbering `token_fingerprint` here is not "latest intent wins" — it is "drop the earlier token's removal forever."

Please either:
- key the outbox by fingerprint/revision so multiple retired tokens can coexist, or
- on conflict with a different fingerprint, insert a second row / keep the older row until completed.

A regression for `disable A → upsert B → disable B` with a blocked publish between steps would lock this in.

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -644,15 +687,62 @@ impl SqliteAccountStorage {
...
+                }
+                SelfMembership::Left | SelfMembership::Removed => {
+                    conn.execute(
+                        "DELETE FROM pending_push_registration_shares
+                         WHERE group_id_hex = ?1",
+                        params![group_id_hex],
+                    )
+                    .storage()?;
+                    conn.execute(
+                        "DELETE FROM pending_push_registration_removals
```
erskingardner: Deleting pending removals on `Left`/`Removed` (and again via `delete_local_group_data` / FK cascade) abandons durable removal intent without a publish attempt.

- Leave can maybe lean on peer `cleanup_stale_push_tokens`, but that is best-effort and races membership delivery.
- Local wipe is worse: MLS membership is intentionally unchanged, yet the outbox row disappears, so peers keep a still-valid member's push token with no remaining local intent to retract it.

If the outbox is the source of truth for "this registration must leave this group," it should survive wipe until published (or be drained before the group row is deleted). On leave, drain removals before clearing.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -505,6 +520,12 @@ async fn run_app_runtime_account_worker(
             }
         }
     }
+    // Automatic gossip is best-effort network work. Run it only after startup
+    // callers have received their deferred responses so a degraded relay cannot
+    // extend account-open latency.
+    client
+        .retry_pending_push_registration_shares_best_effort()
```
erskingardner, line 546: Retry hooks here (and on reconnect/catch-up/create/join/accept/settings) are necessary but not sufficient. After `upsert`/`share` partially fails, a healthy idle worker with live relays can sit forever with non-empty outbox rows — durable on disk, not live in process.

Given the PR explicitly avoids a periodic timer, please either:
- schedule a bounded backoff retry while pending rows exist, or
- document a hard host requirement to call catch-up/share on a timer and enforce it at the UniFFI boundary.

Otherwise "survives restart" is covered and "survives a flaky hour in foreground" is not.

crates/storage-sqlite/src/migrations/0033_push_registration_gossip_outbox.rs
```diff
@@ -0,0 +1,59 @@
...
+    group_id_hex, token_fingerprint, registration_updated_at_ms,
+    queued_at_ms, last_attempted_at_ms
+)
+SELECT account_groups.group_id_hex, push_registration.token_fingerprint,
+       push_registration.updated_at_ms, push_registration.updated_at_ms, NULL
+FROM account_groups
+CROSS JOIN push_registration
+WHERE account_groups.self_membership = 'member';
+
+UPDATE push_registration
```
erskingardner, line 68: Nit / ops: this backfill + `last_shared_at_ms = NULL` forces every upgraded account to re-gossip the current registration to every joined group on the next retry pass. Healing is good; unconditional fanout may be painful for large accounts.

Consider backfilling only groups that never recorded a successful share (`last_shared_at_ms IS NULL`), or otherwise gating the migration storm.

crates/marmot-app/src/client/push.rs
```diff
@@ -7,107 +8,294 @@ use crate::notifications;
...
-        let nostr_signer = signer.as_nostr_signer();
-        let mut removed = 0_usize;
-        for group in self.state.groups.clone() {
-            let Ok(group_id_bytes) = hex::decode(&group.group_id_hex) else {
-                continue;
             };
-            let group_id = GroupId::new(group_id_bytes);
-            let Ok((member_id_hex, leaf_index)) = self.local_member_leaf(&group_id) else {
+            if let Err(err) =
+                self.app
```
erskingardner: Local tombstone/upsert still happens before `send_app_event`. The outbox mostly compensates, but this still fights the multi-step rule that local mirrors of external effects advance only after the effect succeeds.

Prefer publish-then-project (or project only after `Ok` from publish) so a failed publish cannot leave local push state ahead of the group.

Issue #705, opened 2026-07-03
Key-package deletion publishes through a throwaway relay client — cold connects fail with 'transport error' where the live pool works


mubarakcoded:
## Summary

`delete_key_package_event` publishes its kind:5 deletion through `relay_client_for_endpoints`, which builds a **brand-new `NostrSdkClient` on every call** instead of reusing the account's live, already-connected relay pool:

```rust
// crates/marmot-app/src/lib.rs:2818-2826 (de556b2)
fn relay_client_for_endpoints(
    &self,
    keys: &nostr::Keys,
    endpoints: &[TransportEndpoint],
) -> Arc<dyn NostrRelayClient> {
    let _ = endpoints;
    let client = NostrSdkClient::builder().signer(keys.clone()).build();
    Arc::new(NostrSdkRelayClient::new(client))
}
```

Every key-package deletion therefore pays a cold websocket + TLS handshake per relay within the adapter's bounded budget (`SDK_RELAY_CONNECT_WAIT` 5s / `SDK_RELAY_PUBLISH_WAIT` 12s / overall 20s — `crates/transport-nostr-adapter/src/sdk_client.rs:25-42`). On high-latency mobile networks the cold connect fails while the warm pool keeps working, and the failure surfaces as the opaque `"transport error"` category (`crates/marmot-app/src/runtime/mod.rs:657`).

## Field evidence

Reproduced on a physical Pixel 9 Pro XL on Nigerian mobile LTE (Airtel/MTN, carrier NAT), testing whitenoise-android's sign-out wiring (marmot-protocol/whitenoise-android#994, issues #349/#350 there):

- `sign_out(deleteKeyPackages = true)` → outcome reports **0 key packages deleted**, every failure reason `"transport error"`.
- `sign_out_and_wipe` on an account with groups → **group leaves succeed** (they ride the account's live worker pool with warm connections) while **all key-package deletions fail** with `"transport error"` in the same run — same device, same network, seconds apart.

That split is the tell: the only operational difference between the succeeding and failing publishes is warm pool vs throwaway client.

## Affected call sites

All key-package deletion flows funnel through the throwaway client:
- `delete_key_package_event` — `crates/marmot-app/src/lib.rs:2047` (client built at `lib.rs:2077`)
- `sign_out` step 1 loop — `crates/marmot-app/src/runtime/mod.rs:1879`
- `sign_out_and_wipe` key-package stage (same shape)
[...]

PR #1103, opened 2026-07-24
feat: add durable KeyPackage and self-update maintenance

### LINKED ISSUE #1092: Enforce consumed last-resort KeyPackage replacement and private-bundle retirement
## Summary

MDK does not implement the adopted current-profile lifecycle for a successfully consumed last-resort KeyPackage. A successful Welcome join does not durably schedule a fresh replacement, connect confirmed replacement publication to retirement of the old private bundle, or sweep the old bundle at `Lifetime.not_after`.

This leaves the same published last-resort KeyPackage reusable across joins without MDK enforcing the spec's upper bound on its private `init_key` lifetime.

## Security impact

Compromise of a retained last-resort `init_key` can decrypt every recorded Welcome encrypted to that package and recover each affected join's join secrets. Reuse expands one retained private key's blast radius across multiple groups.

Severity is **MEDIUM**: this requires compromise of retained local private material plus recorded Welcome ciphertexts, but the exposure can span every join made with the reused package.

- #1093 owns first-class MLS self-update and post-join/periodic own-leaf rotation. The two issues share durable scheduling/publication foundations but remain independently mergeable feature lanes.
- #1049 retires **legacy** unpublished KeyPackages during the strict profile cutover; it does not define the ongoing lifecycle of consumed current-profile packages.
- #1051 implements founding Add/Welcome delivery durability; it does not require recipient-side KeyPackage replacement and retirement after a successful join.
- #705 remains a separate repair for explicit teardown and legacy-cleanup deletion publication. Routine KeyPackage rotation in this issue does **not** publish kind-5 deletion events.
- #67 and #89 discussed the older design before the lifecycle bound was adopted in marmot-protocol/marmot#408. This issue tracks conformance to the adopted rule rather than reopening the single-use-vs-last-resort decision.

### Durable stable-slot publication

- Store the authoritative KeyPackage slot and lifecycle in the account-device SQLCipher database. Import a valid existing JSON-cache slot once, then stop treating the post-publish JSON write as authoritative.
- Preserve one stable Nostr kind-30443 `d` value for routine publication. A replacement event uses `created_at = max(now, previous_created_at + 1)` so it deterministically supersedes the prior event in that slot.
- Persist the pending replacement, exact signed event bytes and id, snapshotted relay targets, and per-target attempt state before the first network call. After possible exposure, retries must republish only that exact event.
- Treat the first accepted acknowledgement under the active transport as replacement confirmation. Continue best-effort fanout to the remaining snapshotted targets without reopening local lifecycle state.
- If clock rollback would exceed the existing future-skew allowance, surface `clock_skew_blocked` and retry later; do not create a new routine `d` slot.
- Routine replacement must not publish a kind-5 deletion. Existing explicit teardown and legacy-cleanup deletion flows remain unchanged.

### Welcome consumption and private material

- During transactional Welcome processing, identify the consumed `KeyPackageRef` by matching the Welcome's MLS encrypted-group-secrets references to locally stored bundles. Never select private material using the inviter-controlled Nostr `e` tag.
- After successful Welcome processing, prepare a replacement promptly and idempotently. Welcome-triggered replacement does not wait for inbox catch-up.
- Retain the old private material while replacement publication is pending so a Welcome already in flight against that last-resort KeyPackage can still be processed.
- Once replacement publication is acknowledged, atomically promote the replacement and delete the consumed old private `init_key` material.
- Independently delete old private material at its MLS `Lifetime.not_after`, even if replacement never publishes. This local expiry sweep must continue when network maintenance is paused or cursor persistence is frozen.
- Replaying the same Welcome must not create duplicate replacement work or select a different local bundle.

### Routine refresh

- Persist the current reference, lifetime, authored event id/timestamp, sampled refresh deadline, pending replacement, signed event, target snapshot, and retry state.
- Sample ordinary refresh uniformly 14–21 days before expiry.
- For age- or upgrade-triggered refresh, perform best-effort inbox catch-up through first EOSE, capped at 15 seconds. Record one upgrade rotation for a pre-feature current package.
- Retry failures with bounded backoff and expose typed lifecycle status without logging package bytes, refs, account ids, relay URLs, or key material.

## Test coverage

- Stable-`d` replacement, strictly advancing authored timestamps, clock rollback blocking, and absence of routine kind-5 deletion.
- Exact signed-event persistence and identical event ids across restart, including crash before signing, after signing, before exposure, after ambiguous exposure, and after first acknowledgement.
- Acknowledged replacement atomically deletes consumed private material; an unacknowledged replacement retains it; expiry deletes it even while network maintenance is paused/frozen.
- A Welcome already in flight against the prior last-resort package succeeds while replacement is pending and fails after that consumed private material is retired.
- Actual consumed-reference selection is based on MLS references, not the transport `e` tag; Welcome replay is idempotent.
- Refresh jitter, bounded retry, upgrade rotation, and deterministic injected clock/randomness behavior are covered per account/device.
- Multi-account/device tests prove lifecycle records and private-material retirement cannot cross account-device database boundaries.

### LINKED ISSUE #1093: Add durable post-join and periodic MLS self-update scheduling
## Summary

MDK can validate inbound self-update commits, but it exposes no first-class outbound self-update intent and schedules no self-update after a successful Welcome or under a local periodic post-compromise-security policy.

As a result, a newly joined leaf can remain on its pre-join key material until an unrelated commit happens, and quiet groups have no MDK-owned mechanism to refresh a member's leaf keys after compromise.

## Security impact

- Full compromise of a published KeyPackage/device bundle before join can remain useful until that leaf updates.
- A quiet group can retain the same member leaf keys indefinitely, so post-compromise healing depends on unrelated user/admin activity.

Severity is **MEDIUM**: this does not break MLS correctness or interoperability, but it leaves the intended forward-healing mechanism unavailable to MDK consumers.

- #1092 owns durable stable-slot KeyPackage replacement and consumed private-material retirement. The two issues share clock/storage/publication foundations but remain independently mergeable feature lanes.
- marmot-protocol/marmot#379 documents why post-join rotation remains a `SHOULD`; this issue implements that guidance without changing interoperability rules.
- marmot-protocol/marmot#378 leaves periodic cadence to deployment policy; this issue provides the missing MDK mechanism and local policy surface.
- #1051 owns the adopted Welcome lifecycle; its successful joined-state transaction creates the post-join obligation defined here.
- #1062 should track this as current-profile security hardening, but it is separable from the strict wire cutover in #1049.

### First-class self-update and durable publication

- Add `SendIntent::SelfUpdate` through `cgka-engine`, `cgka-session`, `marmot-account`, `marmot-app`, CLI/daemon, agent control, and UniFFI.
- Generate a pure self-update with `force_self_update(true)` and `consume_proposal_store(false)`. It updates only this device's LeafNode and preserves Marmot account-identity-proof invariants.
- Persist semantic intent before MLS staging. Persist the exact signed transport message and snapshotted relay targets before the first network call.
- Track group evolution independently from transport fanout: `preparing`, `prepared`, `attempting`, `confirmed`, or `superseded_by_convergence`, plus per-target fanout state.
- The first accepted relay acknowledgement confirms/applies the local evolution. Remaining targets still receive the same exact event without reopening `PendingPublish` or changing canonical group state.
- Before every retry, reconcile convergence. A losing local commit becomes superseded; re-arm the semantic update only if the selected branch did not rotate this device's own leaf.
- Preserve descriptor-backed removal safety and the existing compatibility path for legacy removal-bearing staged commits; maintenance recovery must never resurrect a removed member.

### Post-join scheduling

- A successful Welcome transaction creates exactly one durable post-join obligation and records the successful join time. UI acceptance does not control eligibility.
- Install a temporary full-history subscription scoped to the joined group, then:
  1. wait for the first relay EOSE or five minutes after successful subscription installation;
  2. continue processing for 15 seconds;
  3. require 60 seconds without valid state-bearing MLS input;
  4. wait a persisted uniform 0–30 second contention jitter;
  5. prepare and publish the self-update.
- Only authenticated valid commits and proposals reset the quiet window. Application messages, duplicates, rejected events, and malformed traffic do not.
- The temporary full-history subscription is post-join only. Periodic work uses normal incremental synchronization and existing gap-triggered widening.
- Track the 24-hour operational target from successful join. Crossing it marks the obligation overdue and increases visibility, but never bypasses convergence, quarantine, removal, offline, or pending-publication gates.
- An intervening confirmed commit satisfies the obligation only when comparison of the pre/post local LeafNode proves this device's leaf key material rotated.
- Application sends remain allowed while rotation is pending. Successful send results carry a machine-readable `SendMaintenanceDisposition` (`ready` or `post_join_rotation_pending_retryable`) without adding a timeline entry or mandatory user warning.

### Periodic and manual policy

- Persist last confirmed own-leaf rotation time per account/device/group.
- Only groups created or joined after rollout are automatically enrolled. Existing groups receive no automatic periodic work.
- For enrolled groups, sample the next cadence uniformly 24–36 days after the last confirmed own-leaf rotation. Once due, require 60 seconds quiet and a persisted uniform 0–15 minute jitter.
- Manual scheduling creates one obligation with the same safety gates and a 0–30 second jitter, then leaves an existing unenrolled group unenrolled after success.
- Expose persisted runtime-wide periodic policy plus process-local pause/resume. Pausing blocks new preparations but retains obligations, deadlines, and already-prepared publication/fanout recovery.
- Serialize group mutation per group and prioritize inbound processing and user commands over maintenance. A blocked group must not prevent maintenance progress in other groups.

### Status and observability

- Expose typed group status covering catch-up, EOSE timeout, grace, quiet, jitter, overdue, paused, pending publication, fanout, retry, convergence supersession, completion, and failure.
- Expose manual scheduling, periodic-policy get/set, pause/resume, and one due-maintenance pass through app runtime, CLI/daemon, agent control, and UniFFI.
- Keep logs/telemetry aggregate-only: attempts, deferrals, overdue counts, retries, recovered publications, superseded commits, fanout completion, and timing distributions without identifiers, relay URLs, payloads, or key material.

## Test coverage

- A successful Welcome creates one durable post-join obligation independent of UI acceptance; replay creates no duplicate.
- Post-join EOSE, five-minute timeout, 15-second grace, valid-input-only quiet resets, persisted jitter, 24-hour overdue status, offline behavior, and restart behavior are deterministic under injected clocks/randomness.
- Application messages remain sendable and return `post_join_rotation_pending_retryable` while the obligation is active.
- Pure self-update preserves account proofs, does not consume unrelated proposals, and changes only the sender's LeafNode.
- Exact signed-event ids survive restart across staging, ambiguous exposure, first acknowledgement, local confirmation, and remaining fanout.
- A losing prepared/attempted update is superseded, its bounded fanout remains exact, and semantic work re-arms only when the selected branch did not rotate the local leaf.
- Recovery of removal-bearing commits never re-adds departed members.
- New-group enrollment, existing-group exclusion, periodic cadence, manual one-shot behavior, pause/resume, and multi-account/device isolation are covered.

erskingardner:
## Summary

- move the authoritative last-resort KeyPackage slot and lifecycle into each account-device SQLCipher database, replacing kind `30443` under one stable `d` value with restart-safe exact-event publication
- identify the actually consumed MLS `KeyPackageRef`, retain its private material until replacement acknowledgement or `Lifetime.not_after`, and never emit routine kind-5 deletions
- add first-class pure MLS self-updates with durable semantic evolution state, post-join EOSE/grace/quiet/jitter scheduling, opt-in-by-creation periodic rotation, manual one-shots, pause/resume, and machine-readable send disposition
- expose typed maintenance status and controls through the app runtime, CLI/daemon, agent control, connector, and UniFFI
- add deterministic clock/randomness seams, aggregate-only telemetry, and workspace version `0.9.7`

## Publication and recovery model

Semantic group evolution and transport fanout are persisted independently. The first accepted relay acknowledgement confirms the local MLS evolution; remaining snapshotted targets continue using the exact signed event without reopening `PendingPublish`. Ambiguous maintenance exposure retains the staged evolution and retries the identical event with bounded backoff.

This branch is rebased onto #1098. Its general frozen outbound-fanout lifecycle remains in place for ordinary MLS traffic, while maintenance self-updates retain the additional semantic obligation and ambiguous-exposure state required here. Maintenance storage is migration `0032`, after the prior-route and outbound-fanout migrations already on `master`.

Successful Welcome processing schedules post-join rotation independently of UI acceptance. Application sends remain enabled and report `ready` or `post_join_rotation_pending_retryable`.

Routine KeyPackage replacement does not publish kind-5 deletion events. #705 remains the separate explicit teardown/legacy deletion transport repair.

## Validation

- `just fast-ci`
- `cargo test -p marmot-account --test runtime` — 30 passed, including the #1098 frozen-fanout cases and restart-safe maintenance exposure
- `cargo test -p cgka-engine --test publish_lifecycle --test pending_commit_recovery` — 13 passed
- `cargo test -p storage-sqlite` — 250 passed
- `cargo test -p marmot-app --test relay_runtime` — 72 passed
- `git diff --check`

Closes #1092
Closes #1093

## Summary by CodeRabbit

* **New Features**
  * Added durable maintenance for KeyPackages and groups, including status views, scheduling, policy controls, pause/resume, and manual execution.
  * Added CLI commands for monitoring and managing maintenance.
  * Added automatic post-join and periodic self-update handling with restart recovery.
  * Added maintenance status and controls to supported application and FFI APIs.
* **Improvements**
  * KeyPackage publishing now supports reliable signed retries and lifecycle tracking.
  * Send responses include maintenance disposition details without blocking messages.
* **Bug Fixes**
  * Improved recovery of interrupted group updates and publication attempts.
  * Added safeguards for clock skew and ambiguous delivery outcomes.


crates/marmot-account/src/runtime.rs
```diff
@@ -199,58 +244,955 @@ where
...
+                    .map(|previous| previous.0.saturating_add(1))
+                    .unwrap_or(now.0)
+                    .max(now.0),
+            );
+            if created_at.0 > now.0.saturating_add(KEY_PACKAGE_MAX_FUTURE_SKEW_SECS) {
+                lifecycle.phase = cgka_traits::MaintenancePhase::ClockSkewBlocked;
+                self.session.put_key_package_lifecycle(&lifecycle)?;
+                return Err(AccountError::ClockSkewBlocked);
+            }
+            let key_package = self.session.fresh_key_package().await?;
```
erskingardner: **Blocker — private material can orphan before lifecycle intent is durable**

`fresh_key_package()` persists the OpenMLS private bundle as a build side effect; `put_key_package_lifecycle` only runs after `pending_replacement` is assembled. A crash/`Err` in that window leaves a current-profile last-resort bundle with **no** lifecycle pointer.

`retire_non_current_key_packages` keeps current-profile bundles, and this branch removed the old unexposed `delete_key_package` prune path, so those orphans are not recoverable via the lifecycle sweep (which only knows about `current_*` / `pending_replacement` / `retained_private_material`).

This also contradicts `PendingKeyPackageReplacement::authored_created_at` (“crash between private-bundle generation and event signing restart-safe”) — that claim is only true *after* this persist.

**Ask:** persist pending intent in the same transaction as bundle creation, or sweep unreferenced current-profile bundles on open; add a crash-injection test for generate→persist (the existing unsigned-replacement restart test starts too late).

crates/traits/src/maintenance.rs
```diff
@@ -0,0 +1,290 @@
...
+    pub created_at: Timestamp,
+    pub bytes: Vec<u8>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PendingKeyPackageReplacement {
+    pub key_package: KeyPackage,
+    pub key_package_ref: Vec<u8>,
+    /// Transport authoring time selected before signing. This makes a crash
+    /// between private-bundle generation and event signing restart-safe while
```
erskingardner: **Doc overclaim**

This documents restart-safety for the generate→sign window, but `AccountDeviceRuntime::publish_fresh_key_package` calls `fresh_key_package()` *before* writing `pending_replacement`. Until that write lands, authorship/`authored_created_at` are not durable either.

Please tighten the comment to the actual invariant (restart-safe **after** pending replacement is persisted; unsigned→signed remains regenerable until first possible exposure), or fix the runtime ordering/transactionality to match this text.

crates/marmot-account/src/runtime.rs
```diff
@@ -199,58 +244,955 @@ where
...
-        // KeyPackage. `KeyPackagePublishError::externally_exposed` distinguishes
-        // the two; we prune only on the unexposed path (mirrors the welcome
-        // publish exposure handling in `publish_group_created`).
-        if let Err(publish_err) = self
+        let now = self.wall_clock.now();
+        let endpoints = self.routing.key_package_endpoints();
+        let mut lifecycle = match self.session.key_package_lifecycle()? {
+            Some(state) => state,
+            None => KeyPackageLifecycleState {
+                stable_slot_id: self
```
erskingardner: **High — upgrade dual-slot hazard without kind-5**

If `legacy_slot_id` returns `None` (missing/corrupt/unclassifiable JSON cache), this mints a **new** random `d`. Routine replacement intentionally never publishes kind-5 (#705 deferred).

On a device that previously published kind `30443` under an old `d`, relays retain the stale replaceable event forever while the new slot publishes in parallel. Authors-filter KP fetches can surface multiple last-resort packages; inviters may Welcome the stale one (private material may already be gone or on another slot).

Fresh installs are fine. The dangerous case is cutover with a failed JSON import. Please fail closed, recover `d` from relays, require explicit migration, or do a one-shot kind-5 for superseded `d`s during import.

erskingardner (commented):
## Adversarial review

Strong foundation on exact-event publication, first-ack confirmation, and MLS-consumed KeyPackageRef selection. I would **not merge as-is**.

### Blockers / high
1. **Private KP generate-before-lifecycle persist** — crash window orphans current-profile init material with no `pending_replacement` pointer (and the old unexposed prune path is gone).
2. **Silent new `d` when legacy JSON import fails** — combined with no routine kind-5, relays can keep a stale last-resort package discoverable.
3. **Idempotent `FinalSent` hardcodes `ready`** — clears `post_join_rotation_pending_retryable` on retry.
4. **Unbounded manual schedule** — every call inserts a new obligation.

### Medium
- Evolution delete after MLS clear transaction (torn write).
- Production publisher always returns `externally_exposed=false`; KP error docs still describe prune-on-unexposed.
- Pause never writes `MaintenancePhase::Paused`; `MaintenanceRun` disposition is structurally always `Ready`.
- CLI plain phase string uses Debug-lowercase (`pendingpublication`).
- Restart tests miss the generate→persist gap.

Inline comments below. Local write-up also saved under `reviews/PR-1103-feat-add-durable-key-package-and-self-update-maintenance.md`.

crates/cgka-engine/src/publish.rs
```diff
@@ -361,6 +373,18 @@ impl<S: StorageProvider> Engine<S> {
...
+            && let Some(maintenance) = self.storage.maintenance_storage()
+            && let Some(evolution) = maintenance
+                .list_group_evolutions()?
+                .into_iter()
+                .find(|evolution| evolution.signed_message_id.as_ref() == Some(message_id))
+        {
+            // No relay accepted the event, so the staged evolution can be
+            // compensated and regenerated from the still-live semantic
+            // obligation.
+            maintenance.delete_group_evolution(&evolution.id)?;
```
erskingardner: **Medium — torn write vs multi-step invariant**

The MLS pending-commit clear + group projection commit in `with_transaction` above, then evolution deletion happens here outside that unit. A crash or `delete_group_evolution` failure leaves Prepared/Attempting evolution + no OpenMLS pending commit / stale `pending_ref`.

Please delete (or mark superseded) the evolution inside the same transaction that clears the pending commit, per `docs/marmot-architecture/overview/multi-step-state-changes.md`.

crates/marmot-app/src/lib.rs
```diff
@@ -3994,41 +3958,120 @@ impl KeyPackagePublisher for AppKeyPackagePublisher {
...
+            ));
+        }
+        let relay_client = self
+            .app
+            .relay_client_for_endpoints(self.signer.as_nostr_signer(), &publication.endpoints);
         let outcome = NostrKeyPackagePublisher::new(relay_client)
-            .publish_key_package(&nostr_publication)
+            .publish_prepared_key_package(&nostr_publication, &event)
             .await
             .map_err(|e| KeyPackagePublishError::unexposed(e.to_string()))?;
```
erskingardner: **Medium — `externally_exposed` is unreachable on the production publish path**

Any adapter/`finish_publish` `Err` after a send attempt is forced to `unexposed`. Timeouts / insufficient acks are exactly the ambiguous-exposure case.

Runtime currently retains pending on all publish errors (good under the durable-slot model), but:
- failure_code becomes `publish_failed` instead of `ambiguous_exposure`
- `KeyPackagePublishError` docs still claim `publish_fresh_key_package` prunes on `externally_exposed == false` — that prune path is gone

Please map post-send failures to `KeyPackagePublishError::exposed(...)` and rewrite the trait docs for retain-until-ack / `not_after`.

crates/marmot-account/src/key_package.rs
```diff
@@ -1,17 +1,27 @@
...
 
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct KeyPackagePublishReceipt {
+    pub accepted: Vec<TransportEndpoint>,
+    pub failed: Vec<TransportEndpoint>,
+}
+
 /// Failure returned by a [`KeyPackagePublisher`].
 ///
 /// `externally_exposed` records whether the KeyPackage may already have been
```
erskingardner: **Stale contract**

This still says `publish_fresh_key_package` keys orphan cleanup on `externally_exposed` and may prune when false. On this branch, failed publishes retain `pending_replacement` and I don’t see a runtime `delete_key_package` call anymore.

Either restore a deliberate pre-exposure prune for the generate-before-persist failure mode, or rewrite this to the new retain-until-ack/`not_after` model so the next reader doesn’t reintroduce mdk#160 against an obsolete flag.

crates/agent-connector/src/messaging.rs
```diff
@@ -126,7 +127,10 @@ impl AgentConnector {
         if let Some(key) = idempotency_key.as_deref()
             && let Some(message_ids_hex) = self.idempotency.get(key, &fingerprint)
         {
-            return Ok(AgentControlResponse::FinalSent { message_ids_hex });
+            return Ok(AgentControlResponse::FinalSent {
+                message_ids_hex,
+                maintenance_disposition: AgentControlSendMaintenanceDisposition::Ready,
```
erskingardner: **High — idempotent retry lies about maintenance disposition**

Cache hit returns the original message ids but hardcodes `Ready`. If the first successful send reported `post_join_rotation_pending_retryable`, a client retry after a post-write timeout will observe `ready` and clear a still-pending post-join obligation signal.

Please persist disposition with the idempotency record (or recompute from live obligations for that group). Also worth asserting this in connector tests — current coverage ignores the field with `..`.

crates/marmot-account/src/runtime.rs
```diff
@@ -199,58 +244,955 @@ where
...
+        group_id: &GroupId,
+    ) -> AccountResult<cgka_traits::MessageId> {
+        use sha2::{Digest, Sha256};
+        self.session.group_record(group_id)?;
+        let now = self.wall_clock.now();
+        let mut hasher = Sha256::new();
+        hasher.update(b"marmot-manual-self-update-v1");
+        hasher.update((group_id.as_slice().len() as u64).to_be_bytes());
+        hasher.update(group_id.as_slice());
+        hasher.update(now.0.to_be_bytes());
```
erskingardner, line 754: **High — unbounded manual obligations**

Every `schedule_manual_self_update` call mixes wall-clock + fresh RNG into the id and inserts a new row. There is no coalesce with an existing active Manual/PostJoin/Periodic obligation.

Connector/CLI/UniFFI can spam this. Evolution id is deterministic per group/epoch/leaf so MLS staging coalesces, but obligation storage + scheduler passes do not. While paused they pile up; on resume they create avoidable work.

**Ask:** return the existing active obligation id for the group (or at least coalesce Manual).

crates/marmot-account/src/runtime.rs
```diff
@@ -199,58 +244,955 @@ where
...
+                continue;
+            }
+            if obligation
+                .operational_target_at
+                .is_some_and(|target| target <= now)
+            {
+                obligation.overdue = true;
+            }
+
+            if self.maintenance_paused {
```
erskingardner, line 991: **Medium — pause is process-local (documented) but status is easy to misread**

Traits correctly say pause/resume is process-local. Here, paused obligations keep their prior phase (`Quiet`/`CatchUp`/…) and never become `MaintenancePhase::Paused`. Send disposition also stays `Ready` unless a PostJoin obligation is active.

So `maintenance_status.paused == true` is the only honest signal; consumers of obligation `phase` alone will think rotation is still progressing. Consider projecting `Paused` into obligation status (or documenting that phase is meaningless while paused), and don’t reuse send disposition for `MaintenanceRun` outcomes when fanout/`possible_exposure` is the real result.

crates/agent-connector/src/maintenance.rs
```diff
@@ -0,0 +1,269 @@
...
+        Ok(AgentControlResponse::Ack)
+    }
+
+    pub(crate) async fn run_maintenance_response(
+        &self,
+        account_id_hex: &str,
+    ) -> Result<AgentControlResponse, ConnectorError> {
+        let account = self.local_account_for_account_id(account_id_hex)?;
+        let summary = self.runtime.run_due_maintenance(&account.label).await?;
+        Ok(AgentControlResponse::MaintenanceRun {
```
erskingardner, line 209: **Medium — `MaintenanceRun` disposition is structurally always `Ready`**

`run_due_maintenance` only drives KeyPackage + `SelfUpdate`. `SendMaintenanceDisposition::PostJoinRotationPendingRetryable` is set only on `AppMessage` sends. After an ambiguous self-update publish, this can still return `Ok` with `published: 0` and `maintenance_disposition: ready` while obligations sit in `PendingPublication` with `possible_exposure`.

Please return a dedicated run result (published / deferred / ambiguous_exposure / failures) instead of reusing send disposition.

crates/cli/src/commands/key_package.rs
```diff
@@ -69,6 +104,25 @@ pub(crate) async fn key_package_command_with_runtime(
...
         }
+        KeyPackageCommand::MaintenanceStatus => {
+            let account = resolve_account(account_home, account_flag)?;
+            ensure_local_signing(&account)?;
+            let status = runtime
+                .key_package_maintenance_status(&account.label)
+                .await?;
+            let phase = status
+                .as_ref()
+                .map(|status| format!("{:?}", status.phase).to_ascii_lowercase())
```
erskingardner: **Medium — plain phase string doesn’t match wire/schema**

`format!("{:?}", phase).to_ascii_lowercase()` turns `PendingPublication` into `pendingpublication`, not `pending_publication` (serde / agent-control snake_case). JSON output is fine; plain/scripts are wrong.

Please share the connector’s `phase_name()` mapping.

crates/marmot-account/tests/runtime.rs
```diff
@@ -580,23 +817,367 @@ async fn publish_fresh_key_package_propagates_publish_error_and_prunes_bundle()
...
+    // One failed attempt + one successful attempt carry the same package,
+    // slot, timestamp, and therefore the same signed artifact identity.
     assert_eq!(publications.len(), 2);
-    assert_ne!(publications[0].key_package, publications[1].key_package);
+    assert_eq!(publications[0], publications[1]);
     assert_eq!(publications[1].key_package, key_package);
 }
 
+#[tokio::test]
+async fn unsigned_key_package_replacement_recovers_after_restart_without_changing_authorship() {
```
erskingardner, line 906: **Test gap for the blocker**

This covers crash after unsigned `pending_replacement` is already durable (good). It does **not** cover crash/failure after `fresh_key_package` and before `put_key_package_lifecycle`.

Please add a fault-injection case that proves no unreferenced current-profile bundle remains after recovery — that is the window the PR description claims is restart-safe.

crates/marmot-account/src/runtime.rs
erskingardner: Fixed in 0ecf6a5d. KeyPackage bundle generation and lifecycle staging now occur in one storage transaction. The new fault-injection test verifies a lifecycle-write failure rolls back both the private bundle and intent, closing the generate-to-persist crash window.

crates/traits/src/maintenance.rs
erskingardner: Fixed in 0ecf6a5d. The invariant is now true rather than qualified: bundle creation and the enclosing lifecycle intent are atomic, and the comment describes signing as the restart-resumable step.

crates/marmot-account/src/runtime.rs
erskingardner: Fixed in 0ecf6a5d. The app imports a recoverable legacy `d` before session open; an existing encrypted account database with missing slot evidence now fails closed even if its old bundles are gone. The generic runtime also no longer infers freshness or mints a slot—account/device creation must provision durable slot authority first. Regression tests cover fresh provisioning, legacy import, existing-database failure, and failure before bundle generation.

crates/cgka-engine/src/publish.rs
erskingardner: Fixed in 0ecf6a5d. Evolution deletion now happens inside the same storage transaction that clears the MLS pending state and commits the group projection.

crates/marmot-app/src/lib.rs
erskingardner: Fixed in 0ecf6a5d. Production adapter failures after the send boundary and zero-ack outcomes are classified as possible exposure; the durable pending replacement remains intact for identical-event retry.

crates/marmot-account/src/key_package.rs
erskingardner: Fixed in 0ecf6a5d. The publisher contract now documents retain-until-ack-or-lifetime-expiry semantics and uses exposure only to distinguish safe pre-exposure handling from exact-event retry after ambiguous exposure.

crates/agent-connector/src/messaging.rs
erskingardner: Fixed in 0ecf6a5d. The durable send-idempotency record now includes `SendMaintenanceDisposition`, and cache hits return the originally recorded disposition. Connector coverage asserts this behavior.

crates/marmot-account/src/runtime.rs
erskingardner: Fixed in 0ecf6a5d. Manual scheduling now returns the existing active per-group obligation instead of inserting another row, including when a PostJoin or Periodic obligation already owns the work. A regression test covers repeated scheduling while paused.

crates/marmot-account/src/runtime.rs
erskingardner: Fixed in 0ecf6a5d. Status now projects every active obligation as `paused` while preserving its durable phase and deadlines for resume. Maintenance-run results are also separated from send disposition.

crates/agent-connector/src/maintenance.rs
erskingardner: Fixed in 0ecf6a5d. `MaintenanceRun` now returns a dedicated summary with published IDs, deferred work, ambiguous exposure, and failures across app runtime, connector/control, CLI, and UniFFI.

crates/cli/src/commands/key_package.rs
erskingardner: Fixed in 0ecf6a5d. `MaintenancePhase::as_str()` is the shared snake_case mapping used by CLI and connector; tests pin multiword values such as `pending_publication`.

crates/marmot-account/tests/runtime.rs
erskingardner: Fixed in 0ecf6a5d. The new transaction fault-injection test fails the lifecycle write after bundle creation and proves rollback leaves no private bundle and no lifecycle row.

crates/cgka-engine/src/engine.rs
```diff
@@ -1084,7 +1141,17 @@ impl<S: StorageProvider> Engine<S> {
...
+            if mls_group.pending_commit().is_some() && !restored_pending {
+                self.restore_durable_group_evolution_on_hydrate(group_id, &mls_group)?
+            } else {
+                false
+            };
+        if mls_group.pending_commit().is_some()
+            && !staged_removes_member
+            && !restored_pending
+            && !restored_durable_evolution
+        {
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Once a matching evolution is discovered, missing/corrupt auxiliary recovery data is aggregate-logged and returns `Ok(false)`, allowing the existing compatibility recovery instead of quarantining the group.

crates/cgka-engine/src/engine.rs
```diff
@@ -1204,6 +1275,118 @@ impl<S: StorageProvider> Engine<S> {
...
+            .into_iter()
+            .rev()
+            .find(|evolution| {
+                evolution.group_id == *group_id
+                    && evolution.source_epoch == source_epoch
+                    && matches!(
+                        evolution.phase,
+                        GroupEvolutionPhase::Prepared | GroupEvolutionPhase::Attempting
+                    )
+            });
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Maintenance storage now has indexed per-group obligation and evolution queries; hydrate, self-update, status, convergence, and scheduler paths use them instead of whole-table filtering.

crates/cgka-engine/src/group_lifecycle.rs
```diff
@@ -870,14 +934,39 @@ impl<S: StorageProvider> Engine<S> {
...
+                            .map(|reference| reference.as_slice().to_vec())
+                    })
+                    .collect::<BTreeSet<_>>();
+                let consumed_key_package_ref = welcome
+                    .secrets()
+                    .iter()
+                    .map(|secret| secret.new_member())
+                    .find(|reference| local_bundle_keys.contains(reference.as_slice()))
+                    .map(|reference| reference.as_slice().to_vec())
+                    .ok_or(EngineError::InvalidWelcome)?;
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Stored KeyPackage bundle decode and reference-hash failures now propagate out of the Welcome transaction rather than being discarded by `filter_map`.

crates/cgka-engine/src/group_lifecycle.rs
```diff
@@ -1030,7 +1119,34 @@ impl<S: StorageProvider> Engine<S> {
...
+                    Some(own_leaf_baseline_hash),
+                )?;
+                if let Some(maintenance) = storage.maintenance_storage()
+                    && let Some(mut lifecycle) = maintenance.key_package_lifecycle()?
+                {
+                    lifecycle.last_consumed_key_package_ref =
+                        Some(consumed_key_package_ref.clone());
+                    lifecycle.last_consumed_at = Some(joined_at);
+                    maintenance.put_key_package_lifecycle(&lifecycle)?;
+                }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Consuming a locally matched KeyPackage now creates a fail-closed lifecycle sentinel when no lifecycle row exists, then durably records the consumed reference and timestamp.

crates/cgka-engine/src/maintenance.rs
```diff
@@ -0,0 +1,176 @@
...
+            Ok::<_, EngineError>(())
+        })
+    }
+
+    pub fn group_maintenance(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<Option<GroupMaintenanceState>, EngineError> {
+        Ok(self.maintenance_storage()?.group_maintenance(group_id)?)
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Group maintenance reads and writes now call `ensure_group_live`, so quarantined or removed groups cannot be mutated through the maintenance surface.

crates/cgka-engine/src/self_update.rs
```diff
@@ -0,0 +1,258 @@
...
+            StagedCommitHandle::from_bytes(group_id.as_slice().to_vec()),
+            pending_ref,
+            crate::epoch_manager::PendingKind::GroupEvolution,
+            self.current_audit_context.clone(),
+        )?;
+        evolution.pending_ref = Some(pending_ref);
+        self.storage
+            .maintenance_storage()
+            .ok_or_else(|| EngineError::Backend("maintenance storage is unavailable".into()))?
+            .put_group_evolution(&evolution)?;
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Self-update now persists the durable evolution before installing the in-memory pending publication, with cleanup guards covering failure at either boundary.

crates/cgka-session/src/lib.rs
```diff
@@ -334,6 +362,94 @@ impl AccountDeviceSession {
...
+    pub fn group_maintenance(
+        &self,
+        group_id: &GroupId,
+    ) -> SessionResult<Option<GroupMaintenanceState>> {
+        Ok(self.engine.group_maintenance(group_id)?)
+    }
+
+    pub fn put_group_maintenance(&self, state: &GroupMaintenanceState) -> SessionResult<()> {
+        Ok(self.engine.put_group_maintenance(state)?)
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. The session maintenance wrappers now enforce the same live-group gate as the engine before exposing per-group state or mutations.

crates/cli/CHANGELOG.md
```diff
@@ -101,6 +101,13 @@ versioning through the workspace version in the root `Cargo.toml`.
 
 ### Added
 
+- Durable KeyPackage and group self-update maintenance. KeyPackages now rotate under one stable kind-30443 `d` slot
+  with exact signed-event retry, persisted lifetime/refresh/retry state, consumed-key retention until replacement
+  acknowledgement, and no routine kind-5 deletion. New `keys maintenance-status` and `groups maintenance-status`,
+  `schedule-self-update`, `maintenance-policy`, `pause-maintenance`, `resume-maintenance`, and `run-maintenance`
+  commands expose the lifecycle. New groups are periodically enrolled by default; existing groups remain manual-only.
+  Successful application-send JSON now includes `maintenance_disposition` without blocking sends while post-join
+  rotation is pending.
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. The changelog reference now links directly to PR #1103.

crates/cli/src/commands/groups.rs
```diff
@@ -1,14 +1,14 @@
 //! `group` and `groups` command namespace handlers and group output helpers.
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. CLI maintenance status and scheduling now run the normal account-state preflight before entering runtime calls.

crates/marmot-account/src/runtime.rs
```diff
@@ -199,58 +244,955 @@ where
...
+                obligation.phase = MaintenancePhase::Quiet;
+                obligation.quiet_since = Some(now);
+                obligation.not_before = None;
+                self.session.put_maintenance_obligation(&obligation)?;
+                self.maintenance_quiet_monotonic
+                    .insert(obligation.id, self.monotonic_clock.elapsed());
+            }
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Valid inbound commits/proposals reset only pre-publication catch-up/grace/quiet/jitter phases. They no longer demote `PendingPublication` or `Retry`, preserving exact-event recovery.

crates/marmot-account/src/runtime.rs
```diff
@@ -199,58 +244,955 @@ where
...
+            tracing::warn!(
+                target: TRACE_TARGET,
+                method = "run_due_maintenance",
+                error_kind = if error.to_string().is_empty() {
+                    "key_package_fanout"
+                } else {
+                    "key_package_fanout_retry"
+                },
+                "key package exact-event fanout remains retryable"
+            );
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Failure codes are now stable, truthful categories selected from the actual error path rather than depending on whether an error string happens to be empty.

crates/marmot-account/src/runtime.rs
```diff
@@ -781,6 +2027,110 @@ where
...
+            {
+                continue;
+            }
+            if fanout.evolution_id.is_some() && !fanout.evolution_confirmed {
+                continue;
+            }
+            self.finish_transport_fanout(&fanout.id, output).await?;
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Fanout recovery now deletes records once their bounded retention expires instead of repeatedly scanning terminal stale rows.

crates/marmot-app/src/client/mod.rs
```diff
@@ -232,35 +235,168 @@ impl AppClient {
...
+            .cloned()
+            .collect::<Vec<_>>();
+        for group_id in stale {
+            if let Some((_, route)) = self.post_join_maintenance_subscriptions.remove(&group_id) {
+                self.adapter
+                    .remove_group_maintenance_subscription(&route)
+                    .await?;
+            }
+        }
+        Ok(())
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Post-join subscription maintenance now isolates failures per group, compensates failed installs, and retains failed removals for retry so one group cannot abort or leak the whole pass.

crates/marmot-app/src/lib.rs
```diff
@@ -1152,17 +1157,18 @@ impl MarmotApp {
...
             .key_package_cutover_replacement_pending(&client.state.label)
         {
-            let cached_current = client
-                .app
-                .latest_key_package(&client.state.label)
-                .ok()
+            let lifecycle_current = client
+                .runtime
+                .key_package_maintenance_status()
+                .map_err(AppError::from)?
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Account-open lifecycle status reads are advisory and no longer abort startup when lifecycle inspection fails; later publication still fails safely at its own boundary.

crates/marmot-app/src/relay_plane/mod.rs
```diff
@@ -792,6 +793,59 @@ impl MarmotRelayPlaneAccountAdapter {
...
+        self.relay_plane
+            .inner
+            .transport
+            .adapter
+            .remove_group_maintenance_subscription(NostrSubscription::GroupMaintenance {
+                account_id: self.account_id.clone(),
+                group_id: group.group_id.clone(),
+                transport_group_id: group.transport_group_id.clone(),
+                endpoints: group.endpoints.clone(),
+            })
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Subscription removal applies the same canonical relay normalization as installation, so equivalent endpoint spellings remove the installed key.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -653,6 +683,63 @@ async fn run_app_runtime_account_worker(
...
+                        }
+                        Err(_) => {
+                            tracing::warn!(
+                                target: "marmot_app::runtime",
+                                method = "key_package_maintenance_catch_up",
+                                "key package maintenance catch-up reached its time cap"
+                            );
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. The account worker now drains and persists projection updates after catch-up maintenance before continuing.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -653,6 +683,63 @@ async fn run_app_runtime_account_worker(
...
+                        &account_label,
+                        account_error_message("post-join maintenance subscription failed", &err),
+                    );
+                }
+                match client.run_due_maintenance().await {
+                    Ok(summary) => {
+                        let _ = summary;
+                        scheduled_convergence
+                            .schedule_groups(client.take_pending_convergence_groups());
+                    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Both scheduled maintenance execution sites now drain projection updates produced by maintenance runs.

crates/storage-sqlite/src/storage/maintenance.rs
```diff
@@ -0,0 +1,375 @@
...
+            .execute(
+                &sql,
+                params![
+                    id.as_slice(),
+                    group_id.map(GroupId::as_slice),
+                    order,
+                    serialized
+                ],
+            )
+            .storage()?;
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Ordered-record allocation is now computed inside the INSERT statement instead of a separate MAX query. A concurrent-writer regression asserts unique insertion order allocation.

crates/transport-nostr-adapter/src/key_package.rs
```diff
@@ -128,6 +136,33 @@ impl NostrKeyPackagePublisher {
...
+            ));
+        }
+        let event_id = MessageId::new(hex::decode(&event.id).map_err(|err| {
+            TransportAdapterError::Publish(format!("invalid KeyPackage event id: {err}"))
+        })?);
+        let mut outcome = self
+            .relay_client
+            .publish_event(&publication.publish_endpoints, event, self.required_acks)
+            .await?;
+        outcome.message_id.get_or_insert(event_id);
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Prepared KeyPackage artifacts are validated against the full expected unsigned event shape—author, kind, tags, content, and created_at—before publication.

crates/transport-nostr-adapter/src/lib.rs
```diff
@@ -400,6 +448,51 @@ impl NostrTransportAdapter {
...
+    pub async fn remove_group_maintenance_subscription(
+        &self,
+        subscription: NostrSubscription,
+    ) -> Result<(), TransportAdapterError> {
+        self.relay_client.unsubscribe(subscription.clone()).await?;
+        self.state
+            .write()
+            .await
+            .forget_subscription_starts(std::slice::from_ref(&subscription));
+        Ok(())
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0ecf6a5d. Subscription transition bookkeeping is serialized; install records intent before subscribe and compensates both relay state and bookkeeping on failure, while removal uses the same transition discipline.

erskingardner (commented):
## Adversarial re-review (head `3c45ef0c`)

### Earlier findings: verified resolved

I re-checked every prior inline finding against the current head rather than trusting the reply text. All of them hold up:

| Finding | Verified at |
| --- | --- |
| KP generate-before-lifecycle-persist crash window | `Engine::stage_key_package_replacement` now builds the bundle and writes the lifecycle inside one `with_transaction`; `SqliteAccountStorage::with_transaction` is a real `BEGIN IMMEDIATE`, so the OpenMLS bundle write rolls back with the intent. |
| Doc overclaim on `authored_created_at` | Comment now describes signing (not generation) as the restart-resumable step. |
| Silent new `d` on failed legacy import | `reusable_key_package_slot_id` returns `Result` and fails closed; only a non-pre-existing account DB with zero stored bundles may mint a slot; `publish_fresh_key_package` refuses an empty/absent `stable_slot_id`. |
| Torn write on evolution delete | `delete_group_evolution` is now inside the same `with_transaction` as `clear_pending_commit` + `put_group`. |
| `externally_exposed` unreachable | Post-send adapter errors and zero-ack receipts both map to `ambiguous_exposure`; retain-until-ack is what the trait docs now describe. |
| Stale `KeyPackagePublishError` contract | Rewritten to retain-until-ack-or-lifetime-expiry. |
| Idempotent retry hardcoded `Ready` | Disposition is persisted in `PersistedSendIdempotencyEntry` and returned from `get_with_disposition`. |
| Unbounded manual obligations | `schedule_manual_self_update` returns the existing non-terminal per-group obligation. |
| Pause not visible in phase | `maintenance_status` projects `Paused` over non-terminal obligations without clobbering the durable phase; `MaintenanceRun` has its own summary type. |
| `MaintenanceRun` always `Ready` | Replaced by `MaintenanceRunSummary { published, message_ids, deferred, ambiguous_exposure, failures }` end to end. |
| CLI Debug-lowercase phase | `MaintenancePhase::as_str()` is shared by CLI + connector and pinned by a test. |
| Test gap for the generate→persist window | Fault-injection test present. |

Spot-checked CodeRabbit majors too — hydrate degrade-to-`Ok(false)`, per-group indexed queries, bundle-decode propagation, missing-lifecycle-row sentinel, `ensure_group_live` on the maintenance surface, evolution-before-`begin_pending` ordering, atomic `insert_order`, canonical relay normalization on removal, full prepared-event comparison in the adapter — all genuinely applied.

---

### Blocker — debug builds overflow the stack; CI is red

Every `Rust tests` shard aborts, and both `CLI e2e (mock relay)` and `Relay CLI E2E` fail because the `wn` binary itself dies:

```
thread 'tests::connector_socket_sends_final_message' has overflowed its stack
fatal runtime error: stack overflow, aborting   (SIGABRT)
```
```
thread 'tokio-rt-worker' has overflowed its stack   # wn, during `chats list`
```

Affected shards abort 4 + 4 + 1 + 4 pre-existing `agent-connector` tests (`connector_socket_sends_final_message`, `connector_policy_accepts_allowed_welcomer`, `connector_socket_composes_and_finalizes_stream`, `send_final_with_repeated_idempotency_key_dedups_without_second_send`, …), and both retry attempts abort — it is deterministic, not flaky.

Reproduced locally on `3c45ef0c`:

```
cargo test -p agent-connector --lib connector_socket_sends_final_message
→ has overflowed its stack / SIGABRT
```

The identical command on `master` passes. Narrowing:

- `RUST_MIN_STACK=33554432` → passes, so this is frame growth, not unbounded recursion.
- `--release` → passes, so it is confined to the debug profile.
- Under lldb the fault lands in `MlsMessageBodyIn as tls_codec::Deserialize>::tls_deserialize` — the large MLS decode frame is now reached from deeper inside the account-open / send chain (e.g. `promote_key_package_lifecycle` → `validate_key_package`, plus the new maintenance work nested into the same async call stack).

The `ci.yml` hunk in this PR is the same symptom, handled locally rather than fixed:

```yaml
- name: Run relay runtime tests serially
  env:
    RUST_MIN_STACK: "4194304"
```

Its comment says this matches "the same 4 MiB stack used by the focused unit harnesses", but `RUST_MIN_STACK` appears nowhere else in the repo — that line is the only one, added here. Meanwhile the connector shards and the CLI e2e jobs, which run through a tokio worker (2 MiB by default), still blow up.

**Ask:** reduce the depth/frame size on the account-open and send paths instead of raising stack sizes per job — `Box::pin` the maintenance sub-futures where they nest into the existing chain, and/or hoist the MLS deserialize/validate work out of the nested call chain. Merge criteria: all four `Rust tests` shards and both e2e jobs green with no `RUST_MIN_STACK` override.

I would not merge until debug builds stop crashing — a `wn` binary that aborts on `chats list` makes every downstream debug-profile test and local dev loop unusable, even though release builds are unaffected.

---

### Non-blocking nit

A crash between `record_sent_openmls_message` and the `Prepared` write in `do_send_self_update` leaves the evolution row at `GroupEvolutionPhase::Preparing`. `restore_durable_group_evolution_on_hydrate` only matches `Prepared | Attempting`, and no reconciliation pass ever sweeps `Preparing`, so the row lingers until a later self-update at the same epoch/leaf reuses the deterministic id and overwrites it. Self-healing in practice; worth a sweep or a comment.

crates/marmot-account/src/error.rs
```diff
@@ -63,6 +63,8 @@ pub enum AccountError {
     TransportRouting(#[from] TransportRoutingError),
     #[error(transparent)]
     KeyPackage(#[from] KeyPackagePublishError),
+    #[error("key package replacement is blocked by local clock skew")]
+    ClockSkewBlocked,
```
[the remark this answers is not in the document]
erskingardner: Fixed in b1e2adbc. AccountError is now #[non_exhaustive], and the cross-crate privacy-safe classifier has an aggregate fallback for future variants. just fast-ci passes.

erskingardner (commented):
## Final re-review (head `6b360695`, rebased)

### Prior findings: still resolved, including through the rebase

The force-push rebased onto a newer `master`, which pulled `SendIdempotencyStore::acquire` (the #1108 in-flight gate) underneath the disposition fix. That is exactly where a rebase silently drops a fix, so I re-checked it: `send_final_response` now matches `SendIdempotencyAcquisition::Completed((message_ids_hex, maintenance_disposition))` and returns the recorded disposition, and both the durable record and `reservation.complete(...)` carry it. Correctly reintegrated, not clobbered.

Re-verified the other rebase-exposed files too — `group_lifecycle.rs` bundle-decode propagation (`collect::<Result<BTreeSet<_>, _>>()?`) and the consumed-KeyPackage lifecycle write, `relay_plane` removal sanitization, per-group isolation in `advance_post_join_maintenance_subscriptions`, `publish.rs` single-transaction evolution delete, `ensure_group_live` on the maintenance surface. All intact.

---

### Blocker (unchanged): the stack overflow was not fixed, only masked in CI

`03bb61bb` promotes `RUST_MIN_STACK: "4194304"` to a workflow-level `env:`. That turns CI green without changing the code that grew the frames. On `6b360695`, with no env override:

```
cargo test -p agent-connector --lib connector_socket_sends_final_message
→ thread '...' has overflowed its stack / SIGABRT
```
```
cargo test -p wn-cli --test cli chats_list_exposes_visible_groups
→ args=["--account", "...", "group", "create", "general", "..."]
  status=signal: 6 (SIGABRT)
  thread 'tokio-rt-worker' has overflowed its stack
```

So a debug `wn group create` still aborts. Release is genuinely unaffected (`cargo test --release -p agent-connector ...` passes on this head), so this is debug-profile only — not a shipped-binary crash.

What makes it still merge-blocking is that `RUST_MIN_STACK` exists **only** in `.github/workflows/ci.yml` — it is not in the `justfile` and not in `.cargo/config.toml`. So:

- `just test`, a documented gate in `AGENTS.md`, aborts locally.
- Plain `cargo test -p agent-connector` aborts locally.
- A developer running a debug `wn` binary gets `SIGABRT` on `group create`.

Green CI now asserts something narrower than "the tests pass" — it asserts they pass under a 4 MiB stack that no local workflow sets. If raising the stack is the intended resolution rather than reducing depth, it needs to land where every consumer sees it (`.cargo/config.toml` `[env]`, or the `justfile` recipes) and the reason should be recorded, so the next person to hit `SIGABRT` on `wn group create` is not re-diagnosing this from scratch. My preference is still to cut the depth — `Box::pin` the maintenance sub-futures nested into the account-open/send chain, and/or hoist the MLS deserialize/validate work (`promote_key_package_lifecycle` → `validate_key_package`) out of the nested chain — but either way the local and CI stories have to match.

---

### New finding — empty-slot sentinel permanently latches KeyPackage publication

Introduced by the fix for the missing-lifecycle-row finding. Two pieces that are individually reasonable combine into a dead end:

1. `crates/cgka-engine/src/group_lifecycle.rs:1151-1159` — on Welcome join, when no lifecycle row exists, it mints `KeyPackageLifecycleState::slot_only(String::new())`, i.e. an **empty** `stable_slot_id`.
2. `crates/marmot-app/src/lib.rs:3019` — `ensure_strict_cutover_replacement_intent_before_session_open` early-returns on `storage.key_package_lifecycle()?.is_some()`.

Reachable sequence:

- Upgraded device, JSON slot record missing or corrupt → the `Ok(None) | Err(_)` arm (`lib.rs:3040-3047`) only sets the cutover-pending marker and writes **no** lifecycle row. Account open succeeds by design.
- Before the slot is repaired, the device accepts a Welcome → step 1 writes the sentinel row with `stable_slot_id: ""`.
- `publish_fresh_key_package` (`runtime.rs:284-289`) now takes the `stable_slot_id.is_empty()` branch and returns migration-blocked — correctly fail-closed.
- But the cutover importer can never run again, because step 2 sees `is_some()`. Restoring the JSON record no longer helps.

Grepping `stable_slot_id`, the only writer of a non-empty slot is that cutover path, so there is no in-app recovery: the device is permanently unable to publish a KeyPackage. Pre-sentinel, the next open with a restored record would have imported the slot and recovered.

Cheapest fix: treat an empty `stable_slot_id` as "still needs import" — change the guard to `.is_some_and(|state| !state.stable_slot_id.is_empty())`, or have the Welcome path record `last_consumed_key_package_ref` without inventing a row that satisfies the cutover guard. Worth a regression test for join-before-slot-import.

---

### Nit

`6b360695` adds `#[non_exhaustive]` to `AccountError`, which forces the `_ => "account_unknown"` arm in `marmot-app/src/error.rs:196`. That drops the compile-time exhaustiveness check on the error-kind mapping, so a future `AccountError` variant silently reports as `account_unknown` instead of failing the build. If the `#[non_exhaustive]` is for downstream semver, consider keeping the mapping exhaustive with an explicit arm list plus `#[deny]`-style coverage, or add a test that asserts every variant maps to a non-`unknown` kind.

erskingardner:
Addressed the two substantive final-review findings in 880f5252.

- The 4 MiB debug/test stack contract now lives in .cargo/config.toml instead of only in CI, and README documents Cargo-launched and direct debug-binary usage. Both reported repros pass with RUST_MIN_STACK removed from the invoking shell; the full agent-connector suite also passes.
- Empty-slot KeyPackage lifecycle rows no longer latch migration permanently. Slot import repairs the existing lifecycle in place and preserves the Welcome-consumed KeyPackage reference and timestamp, with a regression test.

Per request, the AccountError exhaustiveness nit is unchanged.

Validation: just fast-ci; cargo test -p marmot-app; cargo test -p agent-connector; focused wn-cli chats_list_exposes_visible_groups.

PR #1105, opened 2026-07-24
Feat/tui media fidelity

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * TUI auto-starts the relay daemon on launch when relay sources are available, with live startup progress/status.
  * Added direct follow/unfollow (`f`/`x`) from highlighted user search results with immediate badge updates.
* **Improvements**
  * Chat previews and media rendering are more responsive, with centered, consistent loading/empty states and off-thread action handling (FIFO queueing, stale-result dropping).
  * Full-size inline images now use bounded, safer viewer behavior with improved teardown.
* **Bug Fixes**
  * Keyboard accelerators require plain keypresses; refined `Esc` navigation and command/draft trimming.
* **Documentation**
  * Updated CLI TUI docs and changelog to reflect new runtime/interaction semantics.

### DIFF
--- a/crates/cli/src/tui/app.rs
+++ b/crates/cli/src/tui/app.rs
@@ -71,6 +71,14 @@ pub(crate) struct TuiApp {
     /// The one open modal, or none. While set it captures every key (routed at
     /// the top of `handle_key`) and overlays whatever screen is showing.
     pub(crate) popup: Option<Popup>,
+    /// Set when a popup is torn down; `run()` consumes it and clears the whole
+    /// terminal before the next draw. Ratatui otherwise erases a closed popup
+    /// by diffing cells, which cannot reach a pixel-protocol image stored
+    /// terminal-side (iTerm2 does not self-erase when its cells are
+    /// overwritten), so the viewer popup needs the full clear — and it is
+    /// applied uniformly to every popup close, which is rare and cheap, so no
+    /// close path has to know what the popup drew.
+    pub(crate) pending_full_repaint: bool,
     /// Group-detail screen state, loaded on entry and dropped on exit. Present
     /// only while `screen == Screen::GroupDetail`.
     pub(crate) group_detail: Option<GroupDetailView>,
@@ -124,6 +132,7 @@ impl TuiApp {
             streaming: None,
             status: "loading accounts".to_owned(),
             popup: None,
+            pending_full_repaint: false,
             group_detail: None,
             user_search: None,
             profile_view: None,
@@ -158,6 +167,13 @@ impl TuiApp {
             while self.running {
                 dirty |= self.tick();
                 if dirty {
+                    // A closed popup may have drawn a pixel-protocol image that
+                    // lives terminal-side, beyond the cell diff's reach; the
+                    // scheduled full clear wipes it and the draw below repaints
+                    // every cell.
+                    if std::mem::take(&mut self.pending_full_repaint) {
+                        terminal.clear()?;
+                    }
                     terminal.draw(|frame| self.render(frame))?;
                     dirty = false;
                 }
@@ -624,6 +640,11 @@ impl TuiApp {
         });
         match ready {
             Some((name, hash)) => {
+                // Build the native pixel-protocol viewer where the terminal has
+                // one and the pixels are still retained; on `false` the popup
+                // falls back to drawing the shared cell-exact halfblock
+                // protocol, so the viewer opens either way.
+                self.media.build_viewer_protocol(&hash);
                 self.popup = Some(Popup::Image {
                     title: format!("Image: {name}"),
                     hash,
@@ -1092,10 +1113,13 @@ impl TuiApp {
         };
         match popup_key(popup, key.code) {
             PopupAction::None => {}
-            PopupAction::Dismiss => self.popup = None,
+            PopupAction::Dismiss => self.close_popup(),
             // A multi-step flow: the reducer resolved the next popup (the group
             // picker's Enter opens the add-user confirm). No CLI call yet.
-            PopupAction::Open(next) => self.popup = Some(next),
+            PopupAction::Open(next) => {
+                self.close_popup();
+                self.popup = Some(next);
+            }
             PopupAction::Submit(submit) => {
                 // The invites picker stays open across actions so one
                 // accept/decline does not lose the user's place: capture the
@@ -1110,7 +1134,7 @@ impl TuiApp {
                     }) => Some(*selected),
                     _ => None,
                 };
-                self.popup = None;
+                self.close_popup();
                 if let Err(err) = self.run_popup_submit(submit) {
                     self.status = format!("error: {err}");
                 }
@@ -1122,6 +1146,16 @@ impl TuiApp {
             }
         }
         Ok(())
+    }
+
+    /// Tear down the open popup: drop the viewer's on-demand native protocol
+    /// (a no-op for non-image popups) and schedule the full clear+repaint that
+    /// `run()` applies before the next draw. Every popup close funnels through
+    /// here so a terminal-side pixel image can never outlive its popup.
+    fn close_popup(&mut self) {
+        self.popup = None;
+        self.media.drop_viewer_protocol();
+        self.pending_full_repaint = true;
     }
 
     fn run_popup_submit(&mut self, submit: PopupSubmit) -> TuiResult<()> {
--- a/crates/cli/src/tui/app.rs
+++ b/crates/cli/src/tui/app.rs
@@ -865,13 +865,18 @@ impl TuiApp {
 
     /// Fire the debounced flick-through preview: load the highlighted chat's
     /// timeline while focus stays on the chat list. A no-op unless the chat list
-    /// has focus and the highlight names a chat the pane is not already showing.
-    /// The load is tagged with its group id (via `begin_timeline_load`), so a
-    /// preview superseded by further movement is dropped at fold time, and it
-    /// marks the previewed chat read exactly as opening it does — the chat is on
-    /// screen, so viewing-is-reading applies.
+    /// is actually what the user is looking at — the main screen, no popup, and
+    /// the chat list focused — and the highlight names a chat the pane is not
+    /// already showing. Focus alone is not enough: none of the `open_*` screens
+    /// change `self.focus`, and the help popup is reachable from `Focus::Chats`,
+    /// so a preview fired there would reload the pane and mark a chat read behind
+    /// a modal or on another screen (the read marker is forward-only, so that is
+    /// not recoverable). The load is tagged with its group id (via
+    /// `begin_timeline_load`), so a preview superseded by further movement is
+    /// dropped at fold time, and it marks the previewed chat read exactly as
+    /// opening it does — the chat is on screen, so viewing-is-reading applies.
     fn fire_flick_preview(&mut self) {
-        if self.focus != Focus::Chats {
+        if self.screen != Screen::Main || self.popup.is_some() || self.focus != Focus::Chats {
             return;
         }
         let Some(account_id) = self
--- a/crates/cli/src/tui/app.rs
+++ b/crates/cli/src/tui/app.rs
@@ -122,6 +122,21 @@ pub(crate) struct TuiApp {
     pub(crate) effects: EffectRunner,
 }
 
+/// True when a keypress carries no Ctrl/Alt modifier, so a bare accelerator (a
+/// letter that fires an action on its own) may fire. Shift is deliberately
+/// tolerated: the uppercase accelerators (`G`/`R`/`A`/`I`/`P`/`L`) arrive with
+/// SHIFT under the kitty keyboard protocol, so an `is_empty()` modifier check
+/// would silently break them if enhancement flags are ever negotiated. Excluding
+/// only Ctrl/Alt lets chords like Ctrl-U (composer kill-line), Ctrl-Q, and
+/// Ctrl-C reach their own handlers — or fall through harmlessly — instead of
+/// being swallowed by the modifier-insensitive accelerator that shares the
+/// letter. Centralizing the policy in one predicate keeps every accelerator arm
+/// consistent and states the intent ("plain keypress") at each call site.
+fn plain(key: KeyEvent) -> bool {
+    !key.modifiers
+        .intersects(KeyModifiers::CONTROL | KeyModifiers::ALT)
+}
+
 impl TuiApp {
     pub(crate) fn new(cli: Cli) -> TuiResult<Self> {
         let client = WnClient::from_cli(&cli)?;
@@ -477,7 +492,9 @@ impl TuiApp {
             KeyCode::Char('?') if self.focus != Focus::Composer => {
                 self.popup = Some(Popup::help());
             }
-            KeyCode::Char('q') if self.focus != Focus::Composer && self.input.is_empty() => {
+            KeyCode::Char('q')
+                if self.focus != Focus::Composer && self.input.is_empty() && plain(key) =>
+            {
                 self.running = false;
             }
             KeyCode::Tab => self.focus = self.focus.next(),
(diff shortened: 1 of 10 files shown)


dannym-arx (dismissed):
LGTM

<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/mdk/pull/1105)</sub>

crates/cli/src/tui/app.rs
```diff
@@ -642,14 +848,48 @@ impl TuiApp {
...
 
+    /// Fire the debounced flick-through preview: load the highlighted chat's
+    /// timeline while focus stays on the chat list. A no-op unless the chat list
+    /// has focus and the highlight names a chat the pane is not already showing.
+    /// The load is tagged with its group id (via `begin_timeline_load`), so a
+    /// preview superseded by further movement is dropped at fold time, and it
+    /// marks the previewed chat read exactly as opening it does — the chat is on
+    /// screen, so viewing-is-reading applies.
+    fn fire_flick_preview(&mut self) {
+        if self.focus != Focus::Chats {
```
erskingardner: **The flick preview fires after the user has left the main view.** (confirmed with a repro test)

`flick_countdown` is cleared in exactly two places: `tick` when it reaches zero (`app.rs:390`) and `begin_timeline_load` (`client.rs:2022`). None of `open_user_search`, `open_group_detail`, `open_profile`, `open_relay_health`, `open_invites`, or `open_account_picker` clears it — and none of them changes `self.focus` either, so this `self.focus != Focus::Chats` guard still passes on the other screens.

Press `j` in the chat list and then `s` (or `g`/`p`/`h`/`I`/`A`) within ~150 ms. Three ticks later `fire_flick_preview` runs on the User Search screen and calls `begin_timeline_load`, which:

- stomps the screen's own status (`"user search"` → `"loading chat..."`),
- clears the messages pane and drops `group_state_subscription` / `group_diagnostics`,
- queues a `messages timeline list`, and
- via `fold_load_timeline`, queues a `chats mark-read` for a chat the user never opened.

That last one is the part that stings: the runtime read marker is forward-only, so an unread badge cleared this way is not recoverable. On the group-detail path it also breaks the sentinel handshake — `fold_group_detail`'s `if self.status == LOADING_GROUP_DETAIL_STATUS` guard no longer matches, so the load never reports itself.

Repro (fails on this branch, printing `screen=UserSearch status="loading chat..." loading_chat=Some("bbbb…")`):

```rust
#[test]
fn flick_preview_does_not_fire_after_leaving_the_main_view() {
    let account_id = "aa".repeat(32);
    let group_b = "bb".repeat(32);
    let mut app = test_tui_app(test_unused_client(), &account_id);
    app.daemon.running = false;
    app.focus = Focus::Chats;
    app.chats = vec![flick_chat(&"aa".repeat(32), "a"), flick_chat(&group_b, "b")];
    app.selected_chat = 0;

    app.handle_key(char_key('j')).expect("j");
    assert_eq!(app.flick_countdown, Some(FLICK_PREVIEW_DEBOUNCE_TICKS));

    // The user immediately opens the user-search screen.
    app.open_user_search(None);
    assert_eq!(app.screen, Screen::UserSearch);
    assert_eq!(app.status, "user search");

    for _ in 0..FLICK_PREVIEW_DEBOUNCE_TICKS {
        app.tick();
    }
    assert!(
        app.loading_chat.is_none(),
        "flick preview fired while on the user-search screen"
    );
}
```

Suggested fix: guard on the screen and the modality as well as the focus, since focus alone does not describe "the chat list is what the user is looking at":

```rust
fn fire_flick_preview(&mut self) {
    if self.screen != Screen::Main || self.popup.is_some() || self.focus != Focus::Chats {
        return;
    }
    ...
```

The popup half matters independently: `?` (help) is reachable from `Focus::Chats`, so `j` then `?` currently reloads the pane and marks a chat read behind the modal card.

Related, and worth a sentence in the README either way: because the debounce is ~150 ms, walking the chat list at a normal reading pace marks *every* row read, not just the one you settle on. That is the documented reference-client precedent, but the consequence (an irreversible forward-only read marker) is heavier than the "preview" framing suggests.

crates/cli/src/tui/client.rs
```diff
@@ -1518,125 +2003,211 @@ impl TuiApp {
...
+    /// Fold a completed timeline load into the pane, exactly as the synchronous
+    /// path did after its `run_json` returned — but dropping a result whose group
+    /// no longer matches the pending target (a newer open superseded it).
+    fn fold_load_timeline(
+        &mut self,
+        account: String,
+        group: String,
+        result: Result<Vec<Value>, String>,
+    ) {
+        if self.loading_chat.as_deref() != Some(group.as_str()) {
```
erskingardner, line 2061: **`fold_load_timeline` is the one fold with no account anchor.** (confirmed with a repro test)

`fold_relist` (`client.rs:2188`), `fold_invites` (`client.rs:1206`), and `fold_follow_update` (`client.rs:1409`) all compare the effect's `account` against `selected_account_row()` before folding. This guard is group-id only, so `account` arrives, is used to write `messages_account_id` and to build the subscriptions, and is never checked.

Most account switches happen to be safe by accident: `refresh_chats` ends in `refresh_messages()`, which supersedes `loading_chat`. But `refresh_chats` has three early returns that do not — no account selected (`client.rs:1940`), a public-only account (`client.rs:1953`), and an account with no chats (`client.rs:1987`). On any of those the previous account's in-flight load still folds:

```
messages_account_id=Some("aaaa…") messages_group_id=Some("cccc…") status="loaded 0 message(s)"
```

So after switching to a public-only account, the previous account's chat name reappears in the pane title, its timeline is rendered, `ensure_message_subscription` / `ensure_timeline_subscription` are re-opened under the old account, and a `chats mark-read` is queued for it.

Repro (fails on this branch):

```rust
#[test]
fn a_switched_away_accounts_timeline_fold_is_dropped() {
    let account_a = "aa".repeat(32);
    let group_a = "cc".repeat(32);
    let mut app = test_tui_app(test_unused_client(), &account_a);
    app.daemon.running = false;
    app.chats = vec![flick_chat(&group_a, "a")];
    app.selected_chat = 0;

    app.begin_timeline_load(account_a.clone(), group_a.clone());

    // Switch to a public-only account: refresh_chats takes its early return,
    // clearing the pane without queueing a superseding load.
    app.accounts.push(AccountRow {
        account_id: "bb".repeat(32),
        npub: "npub1bob".to_owned(),
        display_name: None,
        local_signing: false,
    });
    app.selected_account = 1;
    app.refresh_chats().expect("refresh_chats");
    assert!(app.messages_group_id.is_none());

    app.apply_effect_done_for_test(EffectDone {
        effect: Effect::LoadTimeline { account: account_a.clone(), group: group_a.clone() },
        result: Ok(vec![serde_json::json!({"messages": []})]),
    });
    assert!(
        app.messages_group_id.is_none(),
        "a switched-away account's timeline fold repopulated the pane"
    );
}
```

Suggested fix — same shape as the other three folds:

```rust
if self.loading_chat.as_deref() != Some(group.as_str())
    || self.selected_account_row().map(|row| row.account_id.as_str()) != Some(account.as_str())
{
    self.loading_chat = None;
    return;
}
```

(Clearing `loading_chat` on the drop path matters here too — otherwise the anchor stays pointed at a group the user has left, and the pane's `empty_messages_notice` keeps showing the yellow `loading messages...` forever.)

crates/cli/src/tui/client.rs
```diff
@@ -877,26 +1157,73 @@ impl TuiApp {
...
+        Ok(())
+    }
+
+    /// Fold the pending-invites list into the invites picker (an empty result
+    /// shows an info card rather than an empty picker). The picker is not
+    /// screen-scoped (it is reachable from both the main view and group detail),
+    /// but it is anchored to the enqueuing account: a result whose account no
+    /// longer matches the selection is dropped rather than opening a picker of a
+    /// switched-away account's invites.
+    fn fold_invites(&mut self, account: String, result: Result<Vec<Value>, String>) {
```
erskingardner, line 1201: **`fold_invites` assigns `self.popup` directly, bypassing the `close_popup()` invariant.**

`close_popup` (`app.rs:1386`) is documented as the single funnel — "Every popup close funnels through here so a terminal-side pixel image can never outlive its popup" — and it does two things this path skips: `media.drop_viewer_protocol()` and `pending_full_repaint = true`.

Reachable: press `I` from the chat list (invites load queued), `Tab` to the messages pane, press `o` on an image message. The image popup opens and, on a pixel-protocol terminal, writes the image terminal-side. When the invites result lands a tick or two later, this line replaces it. No `terminal.clear()` is scheduled, so ratatui erases the popup by cell diff — which is precisely the thing the PR establishes cannot reach an iTerm2/Kitty/Sixel image. The lingering artifact is the exact failure mode `pending_full_repaint` was added for. `viewer_protocol` is also left holding its decoded copy (up to ~11.8 MB) until the next viewer open or close.

Separately, and in the same class as the flick finding: the picker is anchored to the account but to no screen, so `I` followed by `p` pops the invites picker over the Profile screen. The doc comment justifies skipping a screen scope because the picker is reachable from Main *and* group detail — but "not scoped to one screen" and "may appear over any screen including ones it isn't reachable from" are different things.

Suggested fix: route through the teardown before opening, and check the screen is one the picker is reachable from:

```rust
self.close_popup();
self.popup = Some(if items.is_empty() { ... } else { ... });
```

crates/cli/src/tui/media.rs
```diff
@@ -45,6 +96,25 @@ pub(crate) struct MediaState {
...
+    /// halfblocks; this one exists solely so the full-size viewer popup can
+    /// draw the terminal's native image protocol.
+    viewer_picker: Option<Picker>,
+    /// Decoded pixels retained for the viewer popup, per hash, on every
+    /// image-capable terminal — pixel or halfblock-only. They seed the popup's
+    /// own dedicated protocol (native on a pixel terminal, a fresh halfblock
+    /// instance otherwise), so the popup never draws, and so never re-resizes,
+    /// the shared inline protocol. Images are decode-capped, so each entry costs
+    /// at most ~11.8 MB, and the map is bounded by eviction.
+    viewer_images: HashMap<String, DynamicImage>,
```
erskingardner, line 111: **The advertised memory bound only covers the new viewer pool; inline retention is still unbounded in aggregate.**

`viewer_images` is correctly bounded (4 entries via `retain_viewer_image`). But `protocols` (`media.rs:119`) and `statuses` (`media.rs:118`) are insert-only — `rg 'protocols' crates/cli/src/tui/media.rs` shows only `get_mut` and `insert`, no removal anywhere, and nothing prunes on chat switch or scrollback trim.

`cap_decoded_image` genuinely helps per-image (~190 MB camera photo → ~11.8 MB), which is a real improvement. The gap is the aggregate: every image scrolled past in a session keeps a `StatefulProtocol` alive for the life of the process, each holding a decode-capped `DynamicImage` plus its encoded payload. Roughly 85 images ≈ 1 GB, and `MEDIA_MAX_IN_FLIGHT` caps concurrency, not the total.

That is arguably pre-existing, but the PR now makes a bounded-memory claim on top of it. The CHANGELOG says "at most four viewer copies are kept (oldest evicted first, worst case ~47 MB)" and `mod.rs:70-75` says four "bound the viewer's worst-case footprint at ~47 MB" — a reader reasonably takes those as the media subsystem's ceiling, when the actual session ceiling is `~47 MB + 11.8 MB × (images ever scrolled past)`.

Either bound `protocols` the same way (an LRU keyed on the rendered set would fit the existing `viewer_order` idiom), or say plainly in the CHANGELOG that inline protocol retention remains unbounded across a session and only the per-image cost is capped.

crates/cli/src/tui/model.rs
```diff
@@ -4199,6 +4722,25 @@ pub(crate) fn composer_display_text(input: &str) -> String {
...
 /// `--default-account-relays`, so each non-empty list is joined and passed as
 /// the same flag the daemon exposes (flag passthrough, no JSON change).
+/// Whether a `daemon start` spawned by this TUI would have a relay to run with.
+/// Mirrors the exact sources `wn daemon start` accepts — the two passthrough
+/// flag lists, the global `--relay`, and the `WN_RELAY` environment fallback —
+/// so the launch auto-start never declines to start a daemon that would in
+/// fact have started, and never attempts one guaranteed to fail with
+/// `missing_relay_url`. The environment value arrives as a parameter so this
+/// stays pure; the caller reads `WN_RELAY` at the composition root.
+pub(crate) fn daemon_start_has_relay_source(
```
erskingardner: **This function was inserted between `daemon_start_args` and its doc comment.**

The three lines directly above (`/// Args for the daemon start child, forwarding the TUI's first-run relay flags…`, ending `…flag passthrough, no JSON change`) documented `daemon_start_args`. They now read as the opening of *this* function's doc block, and `daemon_start_args` at `model.rs:4744` is left undocumented — the two doc comments run together with no separator, describing two unrelated functions.

Move the new function and its own doc comment below `daemon_start_args`, or restore the blank line + `pub(crate) fn daemon_start_args` boundary.

Minor, and separately: `relay.is_some()` here does not reject an empty string, while `env_relay` does (`is_some_and(|relay| !relay.trim().is_empty())`). `--relay ""` would report a relay source and then fail the start with `missing_relay_url` — the exact outcome the doc comment says this predicate exists to avoid.

crates/cli/src/tui/mod.rs
```diff
@@ -57,6 +66,13 @@ const MEDIA_IMAGE_ROWS: u16 = 8;
 /// images) would spike process/thread pressure. The remainder stay untracked
 /// and are slotted on later ticks as running downloads complete.
 const MEDIA_MAX_IN_FLIGHT: usize = 3;
+/// How many decoded images stay retained in memory for the full-size viewer on
+/// pixel-capable terminals. Each retained copy is decode-capped (at most
+/// ~11.8 MB RGBA), so four bound the viewer's worst-case footprint at ~47 MB
+/// while comfortably covering the freshest arrivals (`MEDIA_MAX_IN_FLIGHT`
+/// admits three at a time). Opening an evicted image's viewer falls back to
+/// the cell-exact halfblock popup.
+const MEDIA_VIEWER_RETAINED_IMAGES: usize = 4;
```
erskingardner, line 83: **This doc contradicts the code it documents.**

"How many decoded images stay retained in memory for the full-size viewer **on pixel-capable terminals**" — but `apply`'s `Decoded` arm (`media.rs:249-262`) retains a `viewer_copy` unconditionally on every image-capable terminal, and `media.rs:105-112` says so explicitly: "Retained on every image-capable terminal (not only pixel ones)".

The code is right and the reason is good (a halfblock-only terminal still needs a dedicated popup instance so drawing at popup size does not re-resize the shared inline protocol). It is this comment that is stale, and the consequence it hides is real: a halfblock-only terminal now carries the same ~47 MB viewer pool a pixel terminal does. Worth stating rather than implying the opposite.

erskingardner (commented):
## Adversarial review

Reviewed the full diff (6.4k added / 1.5k removed across `tui/{app,client,model,media,mod,view}.rs`). `cargo clippy -p wn-cli --all-targets --all-features -- -D warnings` is clean and all 413 `tui::` tests pass, so everything below is a **coverage gap, not a broken test**. Two of the findings I reproduced with failing tests (included inline).

The effect-worker refactor is well anchored overall — `fold_relist`, `fold_invites`, and `fold_follow_update` each check the acting account, and `fold_user_search` / `fold_group_detail` each check their view. The two high findings are the places where that discipline is not applied consistently.

**Summary**

| # | Where | Issue |
|---|---|---|
| 1 | `app.rs:874` | Flick preview fires after the user has left the main view (confirmed) |
| 2 | `client.rs:2048` | `fold_load_timeline` has no account anchor, unlike every other fold (confirmed) |
| 3 | `client.rs:1200` | `fold_invites` replaces an open popup without going through `close_popup()` |
| 4 | `media.rs:111` | Inline `protocols` retention is still unbounded; the advertised ~47 MB worst case only covers the new viewer pool |
| 5 | `model.rs:4732` | New function inserted between `daemon_start_args` and its doc comment |
| 6 | `mod.rs:75` | Doc says viewer pixels are retained "on pixel-capable terminals"; the code retains on every image-capable terminal |

erskingardner (dismissed):

erskingardner (approved):

PR #1106, opened 2026-07-25
Persist Unrecoverable across restart until verified repair.

### LINKED ISSUE #971: Security review M2: Unrecoverable state is in-memory only — restart silently clears it without verified repair
## Summary

On `MissingRetainedAnchor` inside the rollback horizon the engine correctly freezes the group (`Unrecoverable`, ingest and convergence blocked). But the state lives only in `EpochManager`'s in-memory map; on restart, hydration unconditionally calls `set_stable` for every loadable group, silently re-activating it. The adopted protocol requires the halt to persist "until it has a verified repair path."

Found in the 2026-07-20 security review (`docs/security-review-2026-07-20.md`, finding M2). Status: **confirmed.**

## Locations

- `crates/cgka-engine/src/epoch_manager.rs:288-302` — in-memory map only
- `crates/cgka-engine/src/engine.rs:882` — `hydrate_one_stored_group` → unconditional `set_stable`
- `crates/cgka-engine/src/distributed_convergence.rs:238-255, 380-416` — the (correct) in-session freeze path
- `crates/traits/src/engine_state.rs:351-368` — `repair_to_stable` transition defined but never called by the engine

## What the code does

The group will re-enter `Unrecoverable` on the next triggering convergence pass, but between restart and that pass it can send and apply group-state changes on an unrepaired base. Relatedly, the dedicated `repair_to_stable` transition is dead code; the only real exit is an implicit `set_stable` overwrite in `do_join_welcome`.

## Attack scenario / impact

A member forces `MissingRetainedAnchor` (e.g. a withheld commit referencing a pruned/absent anchor) → victim group halts → victim restarts their client → group quietly resumes on an unrepaired base until the next trigger, contrary to the fail-closed contract.

## Suggested remediation

- Persist the `Unrecoverable` marker per group in storage.
- Have hydration honor it: stay halted, surface repair-required.
- Route any reactivation through an explicit verified-repair path (make `repair_to_stable` real).

## Regression test ideas

- Force `Unrecoverable`, restart the engine, assert the group remains halted and ingest/convergence stay blocked until an explicit repair.

erskingardner:
MissingRetainedAnchor freezes a group in Unrecoverable, but that halt lived only in EpochManager memory so hydration's set_stable silently reactivated an unrepaired base after restart. Persist the marker on Group, restore it on hydrate, refuse set_stable overwrite, and exit only via repair_to_stable on an authenticated re-join. fixes #971.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unrecoverable group halts now persist across restarts and reliably block unsafe inbound and outbound activity.
  * Added verified repair handling to restore groups to a stable state after a successful rejoin.
  * Prevented stable-state updates from accidentally clearing an unrecoverable halt.
  * Existing records without the new status now default safely to recoverable.

* **Tests**
  * Added regression coverage for restart persistence, activity blocking, repair flows, and legacy record compatibility.

### DIFF
--- a/crates/cgka-conformance-simulator/tests/openmls_replay_probe.rs
+++ b/crates/cgka-conformance-simulator/tests/openmls_replay_probe.rs
@@ -1628,6 +1628,7 @@ fn dummy_group(group_id: GroupId) -> Group {
         required_capabilities: GroupCapabilities::default(),
         protocol_profile: cgka_traits::group::ProtocolProfile::Legacy,
         removed: false,
+        unrecoverable: false,
         join_epoch: EpochId(0),
     }
 }
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -235,8 +235,13 @@ impl<S: StorageProvider> Engine<S> {
         // A group that has already entered `Unrecoverable` MUST stop applying
         // group-state changes until a verified repair path
         // (spec/protocol-core/group-state.md:50-51,65). Report the halt and
-        // leave canonical state untouched.
-        if self.epoch_manager.is_unrecoverable(group_id) {
+        // leave canonical state untouched. Sync from the durable marker first
+        // so a restart (or a path that skipped hydration) cannot skip the halt
+        // (mdk#971).
+        if self
+            .sync_unrecoverable_halt_from_storage(group_id)
+            .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?
+        {
             let epoch = self
                 .epoch_manager
                 .epoch(group_id)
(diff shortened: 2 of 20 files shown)


erskingardner (commented):
Reviewed against issue #971 (security review finding M2) and the linked spec sections. GitHub blocks self-approval on this PR, so posting as a comment review — recommendation is **approve / merge**.

This closes the real gap: `Unrecoverable` previously lived only in `EpochManager`'s in-memory map, so a restart's unconditional `set_stable` during hydration silently reactivated an unrepaired group.

**What's good:**
- Root-cause fix, not a patch: persists `unrecoverable` on the `Group` record (`crates/traits/src/group.rs`), restores it on hydrate (`engine.rs::hydrate_one_stored_group`), and makes `repair_to_stable` (already defined in `engine_state.rs` but dead) the only real exit — `set_stable` itself now refuses to overwrite `Unrecoverable` (`epoch_manager.rs`).
- Correct write ordering: the durable marker is persisted to storage *before* the in-memory `mark_unrecoverable()` call in `distributed_convergence.rs`, so a crash between the two still gets caught by `sync_unrecoverable_halt_from_storage` on the next entry point — no torn-write window.
- Defense-in-depth guard (`sync_unrecoverable_halt_from_storage`) is placed at every path that can run without session-open hydration (ingest, `do_send`, queued-drain, convergence, join-welcome-repair) — traced each call site and confirmed no `set_stable` call in the engine is reachable while bypassing a guard.
- The repair path (authenticated replacement Welcome) correctly treats "self already a member" as *not* auto-rejecting when the existing record is `unrecoverable`, and clears the durable marker as part of the same transaction that installs the new canonical group state.
- Good test coverage: restart persistence + re-surfaced `GroupUnrecoverable` event, queued-drain halting without rehydration, `set_stable` refusal, and a full end-to-end repair-via-Welcome test.
- FFI/app-facing plumbing (`marmot-app`, `marmot-uniffi`) threads the flag through consistently with existing conventions (e.g. same pattern as `protocol_profile`).
- All CI checks green (fmt, clippy, check, full test matrix, Tamarin, simulator/vectors).

No blocking issues found. One minor, non-blocking observation: the `Unrecoverable`-halt check is now duplicated (via the shared `sync_unrecoverable_halt_from_storage` helper) across five call sites (ingest, `do_send`, drain, convergence, join-welcome) — justified here since each is an independent entry point that can run without hydration, but worth keeping an eye on if a sixth entry point is added later that forgets the guard.

Issue #970, opened 2026-07-21
Security review M1: convergence policy is host/group-tunable, contradicting the pinned v1 baseline


erskingardner:
## Summary

The adopted convergence policy v1 is a set of pinned protocol constants — "every client MUST use exactly the values below… not a local preference." The engine exposes per-group and per-session setters, so two clients in the same group can run different policies and deterministically select different branches.

Found in the 2026-07-20 security review (`docs/security-review-2026-07-20.md`, finding M1). Status: **confirmed.**

## Locations

- `crates/cgka-engine/src/distributed_convergence.rs:81-110` — `set_convergence_policy`, `set_group_convergence_policy`
- `crates/cgka-engine/src/canonicalization.rs:16-39` — policy defaults
- `crates/cgka-session/src/lib.rs` — host config pass-through to the engine
- `docs/marmot-architecture/distributed-convergence.md:157-177` — plans group-negotiated policy

## What the code does

The setters validate only `max_witness_override_depth ≤ max_rewind_commits`. Consequences:

1. Two clients in the same group with different local policies select different branches — the core convergence invariant breaks; the setters are host-reachable config, so this is a realistic deployment mistake.
2. `settlement_quiescence_ms` accepts 0 (no settlement at all → maximal fork churn) or huge values (sends pinned).
3. `app_message_past_epoch_limit` is an independent knob from the engine's MLS `max_past_epochs` (`engine.rs:97`, default 5) — if they disagree, the witness/delivery windows diverge from actual decryptability.
4. Inflated `max_rewind_commits` weakens the forward-secrecy bound the baseline makes explicit and grows retained snapshots and the send-gate window.
5. The architecture doc plans group-negotiated policy, which would turn this into a remote downgrade vector unless constrained to pinned v1.

## Attack scenario / impact

Divergent local policy → permanent group fork between honest clients; a too-small quiescence amplifies reorgs; a mis-set app window invalidates deliverable messages or counts undecryptable witnesses.

## Suggested remediation

- Pin v1 constants in the engine: reject non-default policies until a negotiation mechanism exists (the baseline says any future change MUST ship as a new app component behind a required capability).
- If setters are kept for testing, gate them behind a test-only feature.
- Cross-check `app_message_past_epoch_limit == max_past_epochs` at policy set/read.
- Add a test pinning all six defaults to the v1 constants (currently no such test).

PR #1107, opened 2026-07-25
fix(engine): pin convergence policy to the adopted v1 baseline

erskingardner:
## Summary
- Close security review M1 / #970: release builds accept only the pinned convergence-policy v1 constants, so hosts cannot locally (or per-group) tune branch selection and permanently fork honest clients.
- Keep debug/test overrides for settlement/rewind harnesses, but require `app_message_past_epoch_limit == max_past_epochs` and ignore `dev_settlement_quiescence_ms` in release app/runtime paths.
- Add named v1 constants, pin MLS `DEFAULT_MAX_PAST_EPOCHS` to the app-message window, and cover acceptance/rejection with unit + integration tests.

## Test plan
- [x] `cargo test -p cgka-engine --lib -- canonicalization::witness_window_tests`
- [x] `cargo test -p cgka-engine --test distributed_convergence -- set_convergence_policy_* release_builds_reject*`
- [x] `cargo test -p cgka-engine --test distributed_convergence --test deferred_peel_lifecycle --test mip03_guards --test fork_detection`
- [x] `cargo test -p cgka-session --test session_lifecycle -- session_advance_convergence`
- [x] `cargo check -p cgka-engine -p cgka-session -p marmot-app --release`
- [x] `just fast-ci`

## Summary by CodeRabbit

* **New Features**
  * Release-grade convergence-policy validation is now enforced against the pinned v1 baseline, including app-message past-epoch window alignment.
  * Policy updates now report failures explicitly instead of silently succeeding.
* **Bug Fixes**
  * Release builds reject non-baseline convergence settings and ignore development-only settlement overrides.
* **Documentation**
  * Updated configuration and architecture guidance to clarify convergence-policy/settlement tuning is debug/test-only.
* **Tests & CI**
  * Added a release-mode “convergence-policy pin” test and wired it into CI/Justfile.

### DIFF
--- a/crates/cgka-engine/src/canonicalization.rs
+++ b/crates/cgka-engine/src/canonicalization.rs
@@ -97,15 +97,15 @@ impl CanonicalizationPolicy {
     /// Same acceptance contract as engine setters and session open (mdk#970).
     ///
     /// Always enforces the witness-override bound and app-window alignment.
-    /// Release builds also require the pinned v1 baseline; debug harnesses may
-    /// override for settlement/rewind probes.
+    /// Normal builds also require the pinned v1 baseline. Only test harnesses
+    /// built with the explicit `test-policy-overrides` feature may override it.
     pub fn ensure_acceptable(
         &self,
         max_past_epochs: usize,
     ) -> Result<(), CanonicalizationPolicyError> {
         self.validate()?;
         self.ensure_app_window_matches(max_past_epochs)?;
-        #[cfg(not(debug_assertions))]
+        #[cfg(not(feature = "test-policy-overrides"))]
         if !self.is_pinned_v1() {
             return Err(CanonicalizationPolicyError::NotPinnedV1);
         }
--- a/crates/cgka-engine/src/canonicalization.rs
+++ b/crates/cgka-engine/src/canonicalization.rs
@@ -93,6 +93,24 @@ impl CanonicalizationPolicy {
         }
         Ok(())
     }
+
+    /// Same acceptance contract as engine setters and session open (mdk#970).
+    ///
+    /// Always enforces the witness-override bound and app-window alignment.
+    /// Release builds also require the pinned v1 baseline; debug harnesses may
+    /// override for settlement/rewind probes.
+    pub fn ensure_acceptable(
+        &self,
+        max_past_epochs: usize,
+    ) -> Result<(), CanonicalizationPolicyError> {
+        self.validate()?;
+        self.ensure_app_window_matches(max_past_epochs)?;
+        #[cfg(not(debug_assertions))]
+        if !self.is_pinned_v1() {
+            return Err(CanonicalizationPolicyError::NotPinnedV1);
+        }
+        Ok(())
+    }
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -1060,6 +1078,26 @@ mod witness_window_tests {
                 app_message_past_epoch_limit: V1_APP_MESSAGE_PAST_EPOCH_LIMIT,
                 max_past_epochs: 1,
             })
+        );
+    }
+
+    #[test]
+    fn ensure_acceptable_requires_app_window_alignment() {
+        let policy = CanonicalizationPolicy {
+            app_message_past_epoch_limit: 1,
+            ..CanonicalizationPolicy::default()
+        };
+        assert_eq!(
+            policy.ensure_acceptable(V1_APP_MESSAGE_PAST_EPOCH_LIMIT as usize),
+            Err(CanonicalizationPolicyError::AppWindowMismatch {
+                app_message_past_epoch_limit: 1,
+                max_past_epochs: V1_APP_MESSAGE_PAST_EPOCH_LIMIT,
+            })
+        );
+        assert!(
+            CanonicalizationPolicy::default()
+                .ensure_acceptable(V1_APP_MESSAGE_PAST_EPOCH_LIMIT as usize)
+                .is_ok()
         );
     }
 
--- a/crates/cgka-engine/src/canonicalization.rs
+++ b/crates/cgka-engine/src/canonicalization.rs
@@ -7,11 +7,19 @@
 use std::collections::{BTreeMap, BTreeSet};
 
 use crate::convergence::{
-    AppWitness, BranchCandidate, BranchSelectionTrace, ConvergencePolicy, is_branch_eligible,
-    select_canonical_branch, select_canonical_branch_traced,
+    AppWitness, BranchCandidate, BranchSelectionTrace, ConvergencePolicy, ConvergencePolicyError,
+    is_branch_eligible, select_canonical_branch, select_canonical_branch_traced,
 };
 use cgka_traits::engine::CommitOrderingPriority;
 use serde::{Deserialize, Serialize};
+
+/// Adopted v1 app-message past-epoch delivery/decrypt window.
+///
+/// Must stay equal to [`crate::wire_format::DEFAULT_MAX_PAST_EPOCHS`] — the MLS
+/// past-epoch window — so witness and delivery horizons cannot diverge.
+pub const V1_APP_MESSAGE_PAST_EPOCH_LIMIT: u64 = 5;
+/// Adopted v1 settlement quiescence window, in milliseconds.
+pub const V1_SETTLEMENT_QUIESCENCE_MS: u64 = 1_000;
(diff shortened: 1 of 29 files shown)


crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -110,6 +123,28 @@ impl<S: StorageProvider> Engine<S> {
...
+    ) -> Result<(), OpenMlsProjectionError> {
+        policy
+            .validate()
+            .map_err(|e| OpenMlsProjectionError::InvalidPolicy(e.to_string()))?;
+        policy
+            .ensure_app_window_matches(self.max_past_epochs)
+            .map_err(|e| OpenMlsProjectionError::InvalidPolicy(e.to_string()))?;
+        // Release builds may only run the pinned v1 baseline. Debug harnesses may
+        // override for deterministic settlement / rewind-horizon probes (mdk#970).
+        #[cfg(not(debug_assertions))]
```
erskingardner: This `#[cfg(not(debug_assertions))]` block -- along with the parallel gates in `engine.rs` (`EngineBuilder::build`'s `max_past_epochs` check) and `marmot-app/src/lib.rs` / `account_worker.rs` (`cfg!(debug_assertions)` branches) -- is the actual mechanism that closes M1 in release builds. But nothing in CI ever runs it: `cargo test` / `just test` / `just fast-ci` all compile with `debug_assertions = true` (including `cargo test --release`, unless the workspace overrides `profile.release.debug-assertions`, which it doesn't per `Cargo.toml`). The PR's manual test plan only runs `cargo check ... --release`, which compiles the branch but never executes it.

So `release_builds_reject_non_pinned_convergence_policy` (distributed_convergence.rs test) only exercises `ensure_pinned_v1()` directly under the `cfg!(debug_assertions)` true-branch; the actual reject path inside `accept_convergence_policy` is dead code as far as automated verification goes. A future refactor that inverts this condition, deletes it, or breaks the constant comparison would pass `just ci` cleanly.

Worth considering for a follow-up: either add a CI job that runs the relevant tests with debug-assertions explicitly disabled (e.g. `RUSTFLAGS="-C debug-assertions=off" cargo test ...` or a profile that does that), or move this gate behind an explicit Cargo feature owned by the test/harness crates (e.g. `#[cfg(not(feature = "policy-override"))]`) so `cargo test` without that feature actually exercises the production reject path. Given this closes a confirmed security-review finding, I'd like the enforcement itself -- not just the pure validation helpers -- to be something CI actually proves.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1660,12 +1660,16 @@ impl ScheduledConvergence {
 }
 
 fn convergence_settlement_delay(app: &MarmotApp) -> Duration {
-    Duration::from_millis(
+    // Release builds always schedule against the pinned v1 quiescence window
+    // (mdk#970); the dev override is debug/test-only.
+    let quiescence_ms = if cfg!(debug_assertions) {
         app.config
```
erskingardner, line 1665: Non-blocking nit, not introduced by this PR: `DEFAULT_CONVERGENCE_SETTLEMENT_QUIESCENCE_MS = 1_000` here duplicates `cgka_engine::canonicalization::V1_SETTLEMENT_QUIESCENCE_MS` as a separate literal. Since the whole point of this PR is eliminating magic-number drift around the pinned policy, it'd be tidier to reference the shared constant instead of keeping two `1_000`s that could silently diverge.

erskingardner (commented):
Reviewed against issue #970 (security review M1: convergence policy is host/group-tunable, contradicting the pinned v1 baseline). The fix is well-scoped and closely follows the issue's suggested remediation: it pins v1 constants, threads the same `accept_convergence_policy` fail-closed check through every write and read path (`set_convergence_policy`, `set_group_convergence_policy`, `convergence_policy_for_group`), keeps the app-message/MLS-window cross-check unconditional even in debug builds, and doesn't change any of the actual v1 numeric defaults (5 / 2 / 1 / 1 / 5 / 1000 are unchanged, just named), so there's no data-migration concern for existing stored group policies. Unit/integration test coverage of the validation *logic* (`ensure_pinned_v1`, `ensure_app_window_matches`, rejection of witness-override-exceeds-rewind, etc.) is solid and matches the PR's stated test plan.

My one substantive concern is about verifying the release-mode enforcement itself end-to-end -- see inline comment on `distributed_convergence.rs`. Everything else below is a non-blocking nit.

crates/cgka-session/src/lib.rs
```diff
@@ -286,7 +290,9 @@ impl AccountDeviceSession {
             );
         }
         engine.hydrate_stable_groups_from_storage()?;
-        engine.set_convergence_policy(config.convergence_policy);
+        engine
+            .set_convergence_policy(config.convergence_policy)
+            .map_err(|e| EngineError::Other(format!("convergence policy: {e}")))?;
```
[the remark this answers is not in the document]
erskingardner: Addressed in 7831b755: `AccountDeviceSession::open` now calls `CanonicalizationPolicy::ensure_acceptable` before opening storage / retiring key packages / hydrating. Added `open_rejects_invalid_convergence_policy_before_storage_open` to prove a rejected policy leaves no DB file.

crates/cgka-engine/src/distributed_convergence.rs
erskingardner: Addressed in 7831b755: added `just test-convergence-policy-pin` (`RUSTFLAGS=-C debug-assertions=off`) plus `tests/convergence_policy_pin.rs`, wired into `just fast-ci` / `just ci` and a dedicated GitHub Actions job so the `#[cfg(not(debug_assertions))]` reject path is executed in CI (not just the pure helpers).

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Addressed in 7831b755: scheduler now uses `cgka_engine::canonicalization::V1_SETTLEMENT_QUIESCENCE_MS` instead of a local `1_000` literal.

crates/cgka-engine/src/canonicalization.rs
```diff
@@ -24,18 +32,85 @@ impl Default for CanonicalizationPolicy {
...
+    /// Always enforces the witness-override bound and app-window alignment.
+    /// Release builds also require the pinned v1 baseline; debug harnesses may
+    /// override for settlement/rewind probes.
+    pub fn ensure_acceptable(
+        &self,
+        max_past_epochs: usize,
+    ) -> Result<(), CanonicalizationPolicyError> {
+        self.validate()?;
+        self.ensure_app_window_matches(max_past_epochs)?;
+        #[cfg(not(debug_assertions))]
```
erskingardner: [P1] Do not use `debug_assertions` as the protocol boundary for policy overrides. The adopted convergence spec says every client MUST use the exact v1 values and that the policy is not a local preference. `debug_assertions` is only a build-profile flag: normal `cargo build` and app debug/staging builds enable it, while the public engine/session setters and `MarmotAppConfig::with_dev_settlement_quiescence_ms` remain reachable. Such a client can join real groups with a non-v1 settlement or rewind policy and deterministically diverge—the exact failure #970 is meant to close. Please make the normal engine/session acceptance path require `ensure_pinned_v1()` in every build, and put overrides behind an explicit test/harness-only feature or an injected test clock that app/UniFFI consumers do not enable.

erskingardner (commented):
Blocking finding: the current `debug_assertions` exception still leaves the protocol policy host-tunable in ordinary debug/app builds, contrary to the adopted invariant and #970. The latest quarantine-before-halt-sync and live-policy-read fixes otherwise look coherent against the merged #1106 behavior. Latest CI is still running; the dedicated release-assertions check is green.

crates/cgka-engine/src/canonicalization.rs
erskingardner: Addressed in 30cdf02a. Normal debug and release builds now enforce the pinned v1 policy unconditionally. Non-v1 settlement/rewind and max-past-epoch overrides require an explicit `test-policy-overrides` feature propagated only through test harness builds; app, UniFFI, and ordinary CLI builds do not enable it. The dedicated default-feature CI target now proves rejection directly, while the feature-enabled integration matrix retains existing timing/rewind coverage. Validated with focused engine/session/app/CLI suites and `just fast-ci`.

PR #1108, opened 2026-07-25
fix: serialize matching idempotent sends

### LINKED ISSUE #1039: agent-connector: send_final / stream_finalize idempotency has an in-flight TOCTOU — two concurrent same-key+same-body requests both send, double-posting an unrecallable message

erskingardner:
## Summary
- serialize only matching key-and-fingerprint sends before the first external effect
- let concurrent followers reuse the leader committed message IDs
- release failed leaders so the next caller can retry
- preserve existing cache-miss behavior for different request fingerprints

## Scope
This intentionally fixes only #1039. It adds no delivery-status protocol, Hermes or OpenCode changes, or cross-restart logical-delivery contract. #1088 remains open pending upstream Hermes issue NousResearch/hermes-agent#70694 and a durable delivery design.

## Verification
- cargo test -p agent-connector: 116 passed
- just fast-ci: passed

Fixes #1039

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate message sends when requests with the same idempotency key arrive concurrently.
  - Concurrent matching requests now reuse the original successful result instead of triggering another send.
  - Added clearer handling when a matching request is still processing, including a `send_in_progress` error and guidance to retry with the same idempotency key.
  - Improved retry behavior for finalized responses, preserving message results and preventing duplicate durable publishes.

### DIFF
--- a/crates/agent-connector/src/messaging.rs
+++ b/crates/agent-connector/src/messaging.rs
@@ -12,6 +12,7 @@ use marmot_app::{
 
 use crate::AgentConnector;
 use crate::error::ConnectorError;
+use crate::stream_session::SendIdempotencyAcquisition;
 use crate::validation::normalize_hex;
 
 /// Current schema version for persisted `send_final` request fingerprints.
@@ -120,14 +121,18 @@ impl AgentConnector {
             reply_to_message_id_hex.as_deref(),
         );
 
-        // Idempotent durable send: if this key already committed a matching send,
-        // return the original message ids without re-sending so a retry after a
-        // post-write timeout cannot double-post an unrecallable message.
-        if let Some(key) = idempotency_key.as_deref()
-            && let Some(message_ids_hex) = self.idempotency.get(key, &fingerprint)
-        {
-            return Ok(AgentControlResponse::FinalSent { message_ids_hex });
-        }
+        // Reserve matching requests before the first await so concurrent
+        // same-key retries cannot both miss the cache and both durably send.
+        let idempotency = if let Some(key) = idempotency_key {
+            match self.idempotency.acquire(&key, &fingerprint).await? {
+                SendIdempotencyAcquisition::Completed(message_ids_hex) => {
+                    return Ok(AgentControlResponse::FinalSent { message_ids_hex });
+                }
+                SendIdempotencyAcquisition::Leader(reservation) => Some((key, reservation)),
+            }
+        } else {
+            None
+        };
 
         let account = self.local_account_for_account_id(account_id_hex)?;
         let group_id = GroupId::new(hex::decode(&group_id_hex)?);
@@ -143,7 +148,7 @@ impl AgentConnector {
         // Record only after a successful send so a failed send remains retryable.
         // A key already bound to a different fingerprint is left untouched (first
         // write wins), so this send simply proceeds without caching.
-        if let Some(key) = idempotency_key {
+        if let Some((key, _reservation)) = idempotency {
             self.idempotency
                 .record(key, fingerprint, summary.message_ids.clone());
         }
--- a/crates/agent-connector/src/messaging.rs
+++ b/crates/agent-connector/src/messaging.rs
@@ -110,6 +110,11 @@ impl AgentConnector {
             );
         }
 
+        // Reject unknown accounts and malformed group ids before a request can
+        // wait on or reserve an in-flight send gate.
+        let account = self.local_account_for_account_id(account_id_hex)?;
+        let group_id = GroupId::new(hex::decode(&group_id_hex)?);
+
         // Server-derived request fingerprint: a reused idempotency key only short-
         // circuits when the request it identifies is the same one. A reused key
         // carrying a different request body is a cache miss, so dedup can never
@@ -134,8 +139,6 @@ impl AgentConnector {
             None
         };
 
-        let account = self.local_account_for_account_id(account_id_hex)?;
-        let group_id = GroupId::new(hex::decode(&group_id_hex)?);
         let summary = if let Some(target_message_id) = reply_to_message_id_hex {
             self.runtime
                 .reply_to_message(&account.label, &group_id, &target_message_id, &text)
@@ -148,9 +151,11 @@ impl AgentConnector {
         // Record only after a successful send so a failed send remains retryable.
         // A key already bound to a different fingerprint is left untouched (first
         // write wins), so this send simply proceeds without caching.
-        if let Some((key, _reservation)) = idempotency {
+        if let Some((key, reservation)) = idempotency {
+            let message_ids = summary.message_ids.clone();
             self.idempotency
-                .record(key, fingerprint, summary.message_ids.clone());
+                .record(key, fingerprint, message_ids.clone());
+            reservation.complete(message_ids);
         }
         Ok(AgentControlResponse::FinalSent {
             message_ids_hex: summary.message_ids,
--- a/crates/agent-connector/src/stream.rs
+++ b/crates/agent-connector/src/stream.rs
@@ -19,8 +19,8 @@ use crate::quic::{
     resolve_quic_candidate_addr,
 };
 use crate::stream_session::{
-    ActiveStreamSession, FinalizedStream, StreamBeginReceipt, StreamBeginReservation,
-    normalize_stream_capability,
+    ActiveStreamSession, FinalizedStream, SendIdempotencyAcquisition, SendIdempotencyLeader,
+    StreamBeginReceipt, StreamBeginReservation, normalize_stream_capability,
 };
 use crate::validation::{normalize_hex, transcript_hash_from_hex, unix_now_seconds};
 use crate::{
@@ -95,6 +95,7 @@ fn stream_finalize_idempotency_key(key: &str) -> String {
 struct StreamFinalizeIdempotency {
     key: String,
     fingerprint: String,
+    _reservation: SendIdempotencyLeader,
 }
 
 impl AgentConnector {
@@ -364,14 +365,25 @@ impl AgentConnector {
             .map(str::trim)
             .filter(|key| !key.is_empty())
             .map(stream_finalize_idempotency_key);
-        if let Some(key) = idempotency_key.as_deref()
-            && let Some(message_ids_hex) = self.idempotency.get(key, &fingerprint)
-        {
-            return Ok(AgentControlResponse::StreamFinalized {
-                stream_id_hex,
-                message_ids_hex,
-            });
-        }
+        let idempotency = if let Some(key) = idempotency_key {
+            match self.idempotency.acquire(&key, &fingerprint).await? {
+                SendIdempotencyAcquisition::Completed(message_ids_hex) => {
+                    return Ok(AgentControlResponse::StreamFinalized {
+                        stream_id_hex,
+                        message_ids_hex,
+                    });
+                }
+                SendIdempotencyAcquisition::Leader(reservation) => {
+                    Some(StreamFinalizeIdempotency {
+                        key,
+                        fingerprint,
+                        _reservation: reservation,
+                    })
+                }
+            }
+        } else {
+            None
+        };
         // Clone (not remove) the session: the final text/hash/chunk-count
         // expectation is validated inside the compose task, atomically with
         // its teardown. On mismatch the session stays registered and the
@@ -401,7 +413,7 @@ impl AgentConnector {
                     final_text,
                     transcript_hash,
                     chunk_count,
-                    idempotency_key.map(|key| StreamFinalizeIdempotency { key, fingerprint }),
+                    idempotency,
                 )
                 .await;
         }
@@ -470,7 +482,7 @@ impl AgentConnector {
             final_text,
             transcript_hash,
             chunk_count,
-            idempotency_key.map(|key| StreamFinalizeIdempotency { key, fingerprint }),
+            idempotency,
         )
         .await
     }
--- a/crates/agent-connector/src/stream.rs
+++ b/crates/agent-connector/src/stream.rs
@@ -95,7 +95,7 @@ fn stream_finalize_idempotency_key(key: &str) -> String {
 struct StreamFinalizeIdempotency {
     key: String,
     fingerprint: String,
-    _reservation: SendIdempotencyLeader,
+    reservation: SendIdempotencyLeader,
 }
 
 impl AgentConnector {
@@ -365,6 +365,20 @@ impl AgentConnector {
             .map(str::trim)
             .filter(|key| !key.is_empty())
             .map(stream_finalize_idempotency_key);
+        // Preserve the post-success retry path after its stream session has
+        // been removed, but reject a cache miss with an invalid capability
+        // before it can wait on or reserve an in-flight send gate.
+        if let Some(key) = idempotency_key.as_deref()
+            && let Some(message_ids_hex) = self.idempotency.get(key, &fingerprint)
+        {
+            return Ok(AgentControlResponse::StreamFinalized {
+                stream_id_hex,
+                message_ids_hex,
+            });
+        }
+        let session = self
+            .streams
+            .get_authorized(&stream_id_hex, &stream_capability)?;
         let idempotency = if let Some(key) = idempotency_key {
             match self.idempotency.acquire(&key, &fingerprint).await? {
                 SendIdempotencyAcquisition::Completed(message_ids_hex) => {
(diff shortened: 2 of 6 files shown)


crates/agent-connector/src/stream_session.rs
```diff
@@ -117,6 +132,46 @@ impl SendIdempotencyStore {
             .map(|(_, ids)| ids.clone())
     }
 
+    /// Serialize matching in-flight sends before their first external side
+    /// effect. A follower rechecks the committed cache after the leader exits;
+    /// if the leader failed, the follower becomes the next leader and may retry.
+    ///
+    /// The gate includes the fingerprint so the existing behavior for a reused
+    /// key with a different request body remains a cache miss rather than
+    /// blocking behind or reusing an unrelated send.
```
erskingardner: **[correctness] This guarantee doesn't hold for a key already bound to a different fingerprint.**

The gate is keyed `(key, fingerprint)`, but `record()` below is keyed on `key` alone and early-returns via `if inner.seen.contains_key(&key) { return; }`. So when the key is already bound to a *different* fingerprint:

1. two concurrent requests for `(k, fp-b)` share a gate — the follower correctly waits;
2. the leader sends successfully and calls `record(k, fp-b, ids)` → **no-op**, `k` already maps to `fp-a`;
3. the follower wakes, rechecks `get(k, fp-b)` → still a miss;
4. the follower becomes a second leader and durably re-sends the unrecallable message.

I reproduced this on the branch — the follower comes back as `Leader`, not `Completed`. Net effect versus `master` is the same duplicate count, but the follower now pays the full serialization wait first, and the stated invariant is wrong.

Either make `record` fingerprint-aware so a second binding can be stored, or key the gate on `key` alone and resolve fingerprint divergence inside the critical section. `send_idempotency_store_different_fingerprint_remains_a_cache_miss` only covers the sequential case; the concurrent one deserves a test either way.

The same shape applies to FIFO eviction: if the leader's entry is evicted from the 1024-slot map between `record` and the follower's recheck, the follower re-sends.

crates/agent-connector/src/stream_session.rs
```diff
@@ -117,6 +132,46 @@ impl SendIdempotencyStore {
...
+                gate
+            } else {
+                let gate = Arc::new(tokio::sync::Mutex::new(()));
+                in_flight.insert(gate_key, Arc::downgrade(&gate));
+                gate
+            }
+        };
+        let guard =
+            crate::with_control_operation_timeout("send_idempotency_wait", gate.lock_owned())
+                .await?;
```
erskingardner: **[api-surface] The gate-wait timeout is indistinguishable from a send timeout.**

This produces `ConnectorError::OperationTimedOut`, which projects to code `operation_timed_out` / `"connector operation timed out"` (`error.rs:59,77`) — byte-identical to a timeout of the send itself. The gateway that triggered #1039 sees the exact same signal it already saw, and cannot distinguish "a leader is still in flight; retrying with the *same* key is safe and will return the leader's ids" from "your send timed out".

That matters because a gateway whose retry policy escalates repeated timeouts by minting a *fresh* idempotency key re-opens #1039 through the front door. Consider a distinct code (`send_in_progress`) so the retry contract is explicit to the client.

Also worth noting in the doc comment above: the leader holds this gate across `runtime.send_message` / `finish_agent_text_stream`, neither of which has a deadline. A wedged relay pins the gate and turns every matching retry into a 15s timeout. That's a defensible trade for an unrecallable send, but it's a new head-of-line-blocking property and shouldn't be implicit.

crates/agent-connector/src/messaging.rs
```diff
@@ -120,14 +121,18 @@ impl AgentConnector {
...
-        // post-write timeout cannot double-post an unrecallable message.
-        if let Some(key) = idempotency_key.as_deref()
-            && let Some(message_ids_hex) = self.idempotency.get(key, &fingerprint)
-        {
-            return Ok(AgentControlResponse::FinalSent { message_ids_hex });
-        }
+        // Reserve matching requests before the first await so concurrent
+        // same-key retries cannot both miss the cache and both durably send.
+        let idempotency = if let Some(key) = idempotency_key {
+            match self.idempotency.acquire(&key, &fingerprint).await? {
```
erskingardner, line 132: **[ordering] The gate is acquired before the cheap validation below.**

`local_account_for_account_id(account_id_hex)?` and `hex::decode(&group_id_hex)?` both run *after* this await. A request with an unknown account or malformed group id now blocks on the gate before being rejected. Nothing is corrupted (the failing leader releases immediately), but these checks are free to hoist above `acquire` and the crate's threat model explicitly includes a compromised gateway.

Separately: `idempotency_key` is used raw here, while the stream path trims and namespaces it via `stream_finalize_idempotency_key`. Pre-existing, but the gate key now depends on it, so `" k"` and `"k"` reserve different gates and can both lead.

crates/agent-connector/src/stream.rs
```diff
@@ -364,14 +365,25 @@ impl AgentConnector {
...
             .map(stream_finalize_idempotency_key);
-        if let Some(key) = idempotency_key.as_deref()
-            && let Some(message_ids_hex) = self.idempotency.get(key, &fingerprint)
-        {
-            return Ok(AgentControlResponse::StreamFinalized {
-                stream_id_hex,
-                message_ids_hex,
-            });
-        }
+        let idempotency = if let Some(key) = idempotency_key {
```
erskingardner, line 382: **[ordering] The gate is acquired before `streams.get_authorized(...)`, i.e. before the stream-capability check.**

A caller without the capability now takes a gate and waits before being rejected. The blast radius is small — the capability is part of the fingerprint, so gates can't collide across capabilities, and the failed leader releases immediately — but hoisting the authorization check above `acquire` costs nothing.

More substantively, on this path the reservation is held across `session.tx.send(...)` on a depth-32 bounded channel **and** an untimed `response.await` on the compose task, in addition to `finish_agent_text_stream`. A compose task that stops draining pins the gate with no deadline of its own; only the follower's 15s wait bounds anything. Worth an explicit bound or at least a comment.

crates/agent-connector/src/tests.rs
```diff
@@ -3861,6 +3861,103 @@ fn send_idempotency_store_returns_recorded_ids_for_a_key() {
...
+    };
+
+    let follower_store = store.clone();
+    let follower = tokio::spawn(async move {
+        follower_store
+            .acquire("retry-key", "retry-fingerprint")
+            .await
+            .unwrap()
+    });
+    tokio::task::yield_now().await;
```
erskingardner: **[test-coverage] This test passes even if the gate never engaged.**

A single `yield_now()` is the only handshake, and if the spawned follower hadn't yet reached `gate.lock_owned()` when `leader` is dropped, it would simply acquire a fresh gate and still come back as `Leader` — the assertion succeeds for the wrong reason. Unlike the sibling test at line 3882, there's no `assert!(!follower.is_finished())` to prove the follower actually blocked. Both also depend on `#[tokio::test]`'s current_thread flavor; switching to `multi_thread` would make them racy. A `Notify`/`Barrier` or an observed wait-state would make the intent robust.

Broader gap: all three new tests drive `SendIdempotencyStore` directly. The #1039 regression is "two concurrent `send_final` requests produce two `runtime.send_message` calls" — nothing asserts the runtime send *count* under concurrency. `send_final_with_repeated_idempotency_key_dedups_without_second_send` already exists and is sequential; a concurrent sibling over the fake runtime is the test that would actually fail on `master` and pass here. `stream_finalize_response` has no concurrency coverage at all despite being the second call site.

erskingardner (commented):
## Adversarial review

The core shape is right: a `(key, fingerprint)`-scoped async gate taken *before* the first external effect, with a post-guard cache recheck, leader-failure release, and the reservation held until after `record()` on every success path. I checked the reservation lifetime on both call sites and it is correct — in `messaging.rs` the `if let Some((key, _reservation))` binding outlives the `record()` call, and in `stream.rs` the partial move out of `StreamFinalizeIdempotency` leaves `_reservation` alive to the end of the `if let` block, so there is no window between commit and release. Every early-return path drops the reservation. `OwnedMutexGuard` keeps the gate `Arc` alive, so the `Weak` + `retain` bookkeeping can't prune a gate out from under a leader or a waiter.

Below are the problems I found, most severe first.

---

### 1. Concurrent same-key / different-fingerprint requests still double-send (correctness)

The gate is keyed on `(key, fingerprint)`, but `record()` is keyed on `key` alone and early-returns when the key already exists. When a key is already bound to a different fingerprint, the leader's `record()` is a **no-op**, the follower's post-guard `get()` therefore still misses, the follower becomes a second leader, and it durably re-sends the same unrecallable message.

I verified this against the branch:

```rust
#[tokio::test]
async fn adversarial_reused_key_new_body_concurrent_double_leader() {
    let store = SendIdempotencyStore::new(dir.path());
    store.record("k".to_owned(), "fp-a".to_owned(), vec!["aa".repeat(32)]);

    let leader = /* acquire("k", "fp-b") -> Leader */;
    let follower = tokio::spawn(async move { s2.acquire("k", "fp-b").await.unwrap() });
    tokio::task::yield_now().await;
    assert!(!follower.is_finished()); // passes: follower IS serialized

    store.record("k".to_owned(), "fp-b".to_owned(), vec!["bb".repeat(32)]); // no-op, key exists
    drop(leader);
    // follower -> Leader, not Completed
}
```

Result: `BUG CONFIRMED: follower became a second leader and will send a duplicate`.

This is not a regression (pre-PR both would have sent concurrently), but it does mean the invariant the new comments assert — "concurrent same-key retries cannot both durably send" (`messaging.rs:124`) — is false for this case, and `send_idempotency_store_different_fingerprint_remains_a_cache_miss` only covers it sequentially. The follower now *also* pays the full serialization latency before sending its duplicate, so it's strictly slower for the same outcome.

Either fix it (make `record` fingerprint-aware so a second binding can be stored, or make the gate key-only and resolve fingerprint divergence inside the critical section) or tighten the comments and add the concurrent case as a documented-known-gap test. Same shape applies to FIFO eviction: if the leader's entry is evicted from the 1024-slot map between `record` and the follower's recheck, the follower re-sends. Vanishingly rare, but the same root cause.

### 2. A gate-wait timeout is indistinguishable from a send timeout (API surface)

`with_control_operation_timeout("send_idempotency_wait", ...)` yields `OperationTimedOut`, which projects to code `operation_timed_out` and message `"connector operation timed out"` — byte-identical to a timeout of the send itself (`error.rs:59,77`). The gateway that triggered #1039 sees the exact same signal it saw before and cannot tell "a leader is still in flight, retrying with the same key is safe and will return the leader's ids" from "your send timed out". A gateway whose retry policy escalates repeated timeouts by minting a *fresh* idempotency key re-opens #1039 through the front door. Worth a distinct code (`send_in_progress` / `send_already_in_flight`) so the client contract is explicit.

### 3. No connector-level concurrency test (test coverage)

All three new tests drive `SendIdempotencyStore` directly. The #1039 regression is specifically "two concurrent `send_final` requests produce two `runtime.send_message` calls" — nothing here asserts the runtime send *count* under concurrency. `send_final_with_repeated_idempotency_key_dedups_without_second_send` already exists and is sequential; a concurrent sibling over the fake runtime would be the actual regression test for the issue this PR closes. Same for `stream_finalize_response`, which has no concurrency test at all despite being the second call site.

### 4. The leader holds the gate across untimed external work (liveness)

`runtime.send_message` / `reply_to_message` (`messaging.rs:139-147`) and `finish_agent_text_stream` (`stream.rs:509`) carry no deadline. `stream_finalize_response` is worse: the gate is held across `session.tx.send(...).await` on a depth-32 bounded channel *and* an untimed `response.await` on the compose task. A wedged relay or compose task pins the gate indefinitely and converts every matching retry into a 15s `operation_timed_out`. That's arguably the right trade for an unrecallable send, but it is a new head-of-line blocking property and should be stated explicitly in the `acquire` doc comment (and ideally bounded).

### 5. The gate is taken before cheap validation and before authorization (ordering)

`messaging.rs:127` acquires before `local_account_for_account_id` and `hex::decode(&group_id_hex)`; `stream.rs:368` acquires before `self.streams.get_authorized(...)`, i.e. before the stream-capability check. A malformed or unauthorized request now waits on the gate before being rejected. Impact is small (the capability is inside the fingerprint, so gates can't collide across capabilities, and the failing leader releases immediately), but the crate's threat model explicitly includes a prompt-injected/compromised gateway, and reordering these checks ahead of `acquire` is free.

### 6. `AGENTS.md` not updated (docs)

`crates/agent-connector/AGENTS.md:43` documents `SendIdempotencyStore` down to the FIFO capacity, fingerprint versioning, `stream_finalize_v2:` prefix, and atomic writes. The single-flight gate is now equally load-bearing and isn't mentioned.

---

**Verified locally:** `cargo test -p agent-connector idempotency` → 12 passed.

crates/agent-connector/src/stream_session.rs
erskingardner: Fixed in 1e98129e. The in-flight gate now carries a transient successful result for callers already waiting on the same key and fingerprint. This prevents the reproduced second leader even when the persistent key remains bound to an older fingerprint, while preserving the existing sequential different-fingerprint cache-miss and first-write-wins behavior. Added a concurrent regression that verifies the old durable binding remains unchanged and the waiting follower reuses the transient IDs.

crates/agent-connector/src/stream_session.rs
erskingardner: Fixed in 1e98129e. A bounded gate-wait expiry now projects as send_in_progress with explicit guidance to retry using the same idempotency key, rather than operation_timed_out. The acquire docs now state that the leader intentionally holds the gate across the external effect and recording; we do not cancel that work because cancellation cannot establish that an unrecallable send did not land.

erskingardner:
Addressed the adversarial review in 1e98129e, including the unthreaded documentation finding: the connector AGENTS.md now documents the same-key/same-fingerprint gate and send_in_progress follower contract.

Validation:
- cargo test -p agent-connector: 118 passed
- just fast-ci: passed

PR #1110, opened 2026-07-25
feat: add durable frozen convergence passes

### LINKED ISSUE #806: Bound convergence collection to one frozen 5000 ms pass
## Summary

Implement the adopted convergence collection boundary. The current scheduler resets/extends its timer as more inputs arrive, so a busy group can postpone selection indefinitely and different clients can select from different moving batches.

One convergence pass has an absolute 5000 ms collection deadline and one frozen input batch. Inputs arriving after the cutoff belong to a later pass.

## Normative source

- `protocol-core/convergence.md`
- `protocol-core/publish-lifecycle.md`

Related trackers: #707 and #1027.

## Start here

- `crates/cgka-engine/src/distributed_convergence.rs`
- `crates/cgka-engine/src/engine.rs`
- convergence scheduling/persistence in `crates/traits/` and `crates/storage-sqlite/`
- `crates/cgka-engine/tests/distributed_convergence.rs`

## Required behavior

- The first eligible input opens a pass with an absolute deadline exactly 5000 ms later according to the scheduler's monotonic/durable time model.
- Later arrivals before the deadline join that pass but never move its deadline.
- At cutoff, atomically freeze the eligible input IDs/digests and selection state. Selection and processing operate only on that snapshot.
- Inputs arriving during/after processing are assigned to the next pass.
- Persist enough state to recover the same deadline/frozen batch after restart; restart must not grant another full moving window.
- Process the selected/losing branches with fair bounded work so one group cannot starve other groups.
- Map results to the adopted outcome taxonomy. In particular, do not treat `duplicate`, `own_echo`, `wrong_recipient`, or `unknown_group` as stale convergence candidates, and remove the non-spec `SelfEvicted` category from protocol-facing outcomes.
- When a winning branch advances the group, defer or reclassify losing work according to the spec rather than applying it against a moving state.

## Acceptance criteria

- Deterministic-clock tests prove continuous arrivals cannot extend the 5000 ms deadline.
- Boundary tests cover just before, exactly at, and just after cutoff.
- Restart tests preserve the deadline and exact frozen batch.
- Permuting post-cutoff arrivals does not change the current pass winner/outcomes.
- Multi-group load tests demonstrate bounded fairness.
- Direct and wrapped ingest seams produce the same adopted outcome classes.
- `just fast-ci` and targeted engine/storage/conformance tests pass.

erskingardner:
## Summary

Closes #806.

This adds an engine- and storage-owned frozen convergence pass boundary so a group resolves a stable authenticated input set instead of repeatedly reading a moving live batch.

The implementation is based on `master` after #1103 and #1107 merged.

## What changed

- Add durable convergence-pass state with explicit collecting, frozen, resolving, and completed phases.
- Admit the stored message and its pass membership atomically.
- Freeze content-derived message identifiers and peeled MLS payload digests.
- Bound collection by both quiescence and the v1 five-second absolute deadline.
- Reconstruct collecting passes safely after restart and preserve exact frozen membership across restart.
- Fail closed and mark the group unrecoverable if a frozen member is missing or its digest changes.
- Restrict canonicalization to the frozen member set so post-cutoff input cannot affect the active pass.
- Persist a one-turn fairness slot for already-queued administrative group-state evolution before the next inbound pass.
- Replace the app worker's global convergence timer with independent per-group deadlines.
- Split ignored transport/input cases and local-state outcomes from protocol-stale outcomes.
- Add SQLite migration `0035_durable_convergence_passes`.
- Document the architecture, invariants, recovery behavior, rollout considerations, and review plan.

## Safety properties

- Duplicate delivery does not extend a pass.
- Continuous input cannot postpone convergence beyond the absolute deadline.
- Input arriving at or after the cutoff is retained for a later generation but excluded from the frozen pass.
- Pass admission, freeze transitions, and workflow completion are transactionally persisted.
- Group epoch snapshots intentionally do not rewind convergence workflow state.
- Active passes gate sends and automatic self-updates.
- A failed fairness-slot preparation does not block subsequent inbound convergence.
- Scheduling activity for one group cannot postpone another group's cutoff.

## Validation

- `just ci`
  - default nextest matrix: 2,971 passed, 2 skipped
  - OTLP-feature nextest matrix: 2,977 passed, 2 skipped
  - workspace doc tests passed
- focused distributed convergence suite: 62 passed
- SQLite storage suite: 271 passed
- `git diff --check`

## Review guide

The complete implementation and review plan is in:

`docs/marmot-architecture/further-context/issue-806-frozen-convergence-pass-plan.md`

The main code entry points are:

- `crates/cgka-engine/src/distributed_convergence.rs`
- `crates/traits/src/convergence_pass.rs`
- `crates/storage-sqlite/src/storage/convergence_passes.rs`
- `crates/marmot-app/src/runtime/account_worker.rs`

## Summary by CodeRabbit

* **New Features**
  * Added durable, restart-safe frozen convergence passes with per-group cutoff scheduling and an absolute 5s convergence-pass cap.
  * Exposed per-group convergence eligibility delay to the session/runtime, enabling clients to observe when convergence can run.
  * Added typed ingest outcome categories for duplicates/own-echo/wrong-recipient/unknown-group, plus local terminal states (removed/quarantined).
* **Bug Fixes**
  * Improved fail-closed integrity verification for frozen convergence membership and tightened replay/determinism around convergence cutoff handling.
* **Documentation**
  * Updated architecture/audit-log docs to reflect the revised convergence and ingest outcome semantics.

### DIFF
--- a/crates/cgka-engine/src/convergence_input.rs
+++ b/crates/cgka-engine/src/convergence_input.rs
@@ -1,0 +1,259 @@
+//! Role-aware classification for retained convergence inputs.
+//!
+//! Commits form candidate branches, proposals can satisfy commit dependencies,
+//! and application messages can provide bounded witness evidence. Keeping
+//! those roles explicit prevents an unresolved payload disposition from being
+//! mistaken for ambiguous canonical group state.
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use cgka_traits::message::MessageState;
+
+use crate::openmls_projection::OpenMlsContentKind;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum ConvergenceInputRole {
+    CommitEdge,
+    ProposalDependency,
+    AppWitnessCandidate,
+}
+
+impl ConvergenceInputRole {
+    pub(crate) fn for_content_kind(kind: OpenMlsContentKind) -> Option<Self> {
+        match kind {
+            OpenMlsContentKind::Commit => Some(Self::CommitEdge),
+            OpenMlsContentKind::Proposal => Some(Self::ProposalDependency),
+            OpenMlsContentKind::Application => Some(Self::AppWitnessCandidate),
+            OpenMlsContentKind::Welcome | OpenMlsContentKind::Other => None,
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) struct ClassifiedConvergenceInput {
+    pub(crate) role: ConvergenceInputRole,
+    pub(crate) source_epoch: u64,
+    pub(crate) state: MessageState,
+    pub(crate) digest: [u8; 32],
+}
+
+impl ClassifiedConvergenceInput {
+    pub(crate) fn from_projection(
+        kind: OpenMlsContentKind,
+        source_epoch: u64,
+        state: MessageState,
+        digest: [u8; 32],
+    ) -> Option<Self> {
+        Some(Self {
+            role: ConvergenceInputRole::for_content_kind(kind)?,
+            source_epoch,
+            state,
+            digest,
+        })
+    }
+
+    fn can_start_pass(self) -> bool {
+        matches!(
+            self.state,
+            MessageState::Sent | MessageState::Created | MessageState::Retryable
+        )
+    }
+
+    fn can_gate_outbound(self) -> bool {
+        matches!(self.state, MessageState::Created | MessageState::Retryable)
+    }
+}
+
+#[derive(Default)]
+pub(crate) struct ConvergenceInputContext {
+    commit_edges_by_source_epoch: BTreeMap<u64, BTreeSet<[u8; 32]>>,
+}
+
+impl ConvergenceInputContext {
+    pub(crate) fn from_inputs(
+        inputs: impl IntoIterator<Item = ClassifiedConvergenceInput>,
+    ) -> Self {
+        let mut context = Self::default();
+        for input in inputs {
+            if input.role == ConvergenceInputRole::CommitEdge {
+                context
+                    .commit_edges_by_source_epoch
+                    .entry(input.source_epoch)
+                    .or_default()
+                    .insert(input.digest);
+            }
+        }
+        context
+    }
+
+    fn has_commit_at(&self, source_epoch: u64) -> bool {
+        self.commit_edges_by_source_epoch
+            .get(&source_epoch)
+            .is_some_and(|edges| !edges.is_empty())
+    }
+
+    fn has_competing_commit_before(&self, source_epoch: u64) -> bool {
+        self.commit_edges_by_source_epoch
+            .range(..source_epoch)
+            .any(|(_, edges)| edges.len() > 1)
+    }
+
+    /// Whether admitting this input can change deterministic resolution of the
+    /// current batch.
+    ///
+    /// An application message is only potentially selection-relevant when the
+    /// frozen inputs contain competing commit edges from an earlier epoch.  A
+    /// proposal is only potentially relevant when a commit at the same source
+    /// epoch may depend on it. Authentication and actual dependency/witness
+    /// contribution are still proven by OpenMLS during frozen resolution.
+    pub(crate) fn is_potentially_selection_relevant(
+        &self,
+        input: ClassifiedConvergenceInput,
+    ) -> bool {
+        match input.role {
+            ConvergenceInputRole::CommitEdge => true,
+            ConvergenceInputRole::ProposalDependency => self.has_commit_at(input.source_epoch),
+            ConvergenceInputRole::AppWitnessCandidate => {
+                self.has_competing_commit_before(input.source_epoch)
+            }
+        }
+    }
+
+    pub(crate) fn opens_pass(&self, input: ClassifiedConvergenceInput) -> bool {
+        input.can_start_pass()
+            && match input.role {
+                ConvergenceInputRole::CommitEdge => true,
+                ConvergenceInputRole::ProposalDependency => false,
+                ConvergenceInputRole::AppWitnessCandidate => {
+                    self.has_competing_commit_before(input.source_epoch)
+                }
+            }
+    }
+
+    pub(crate) fn gates_outbound(&self, input: ClassifiedConvergenceInput) -> bool {
+        input.can_gate_outbound()
+            && match input.role {
+                ConvergenceInputRole::CommitEdge => true,
+                ConvergenceInputRole::ProposalDependency => false,
+                ConvergenceInputRole::AppWitnessCandidate => {
+                    self.has_competing_commit_before(input.source_epoch)
+                }
+            }
+    }
+
+    /// Whether a member of an already-active pass keeps outbound work gated.
+    ///
+    /// Pass membership, not the mutable message disposition, is authoritative
+    /// here. A crash can leave a frozen/resolving pass after an accepted commit
+    /// record has already moved to `Processed`; outbound work must still wait
+    /// until the durable pass reaches `Completed`.
+    pub(crate) fn active_pass_member_gates_outbound(
+        &self,
+        input: ClassifiedConvergenceInput,
+    ) -> bool {
+        match input.role {
+            ConvergenceInputRole::CommitEdge => true,
+            ConvergenceInputRole::ProposalDependency => false,
+            ConvergenceInputRole::AppWitnessCandidate => {
+                self.has_competing_commit_before(input.source_epoch)
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn input(
+        role: ConvergenceInputRole,
+        source_epoch: u64,
+        state: MessageState,
+        digest_byte: u8,
+    ) -> ClassifiedConvergenceInput {
+        ClassifiedConvergenceInput {
+            role,
+            source_epoch,
+            state,
+            digest: [digest_byte; 32],
+        }
+    }
+
+    #[test]
+    fn future_app_without_a_candidate_branch_is_delivery_only_for_scheduling() {
+        let app = input(
+            ConvergenceInputRole::AppWitnessCandidate,
+            2,
+            MessageState::Created,
+            1,
+        );
+        let context = ConvergenceInputContext::from_inputs([app]);
+
+        assert!(!context.opens_pass(app));
+        assert!(!context.gates_outbound(app));
+        assert!(!context.is_potentially_selection_relevant(app));
+    }
+
+    #[test]
+    fn app_after_competing_edges_is_a_selection_relevant_witness_candidate() {
+        let first = input(
+            ConvergenceInputRole::CommitEdge,
+            1,
+            MessageState::Created,
+            1,
+        );
+        let second = input(
+            ConvergenceInputRole::CommitEdge,
+            1,
+            MessageState::Created,
+            2,
+        );
+        let app = input(
+            ConvergenceInputRole::AppWitnessCandidate,
+            2,
+            MessageState::Created,
+            3,
+        );
+        let context = ConvergenceInputContext::from_inputs([first, second, app]);
+
+        assert!(context.opens_pass(app));
+        assert!(context.gates_outbound(app));
+        assert!(context.is_potentially_selection_relevant(app));
+        assert!(context.active_pass_member_gates_outbound(app));
+    }
+
+    #[test]
+    fn proposal_only_becomes_potentially_relevant_beside_same_epoch_commit() {
+        let commit = input(
+            ConvergenceInputRole::CommitEdge,
+            4,
+            MessageState::Created,
+            1,
+        );
+        let proposal = input(
+            ConvergenceInputRole::ProposalDependency,
+            4,
+            MessageState::Created,
+            2,
+        );
+        let context = ConvergenceInputContext::from_inputs([commit, proposal]);
+
+        assert!(context.is_potentially_selection_relevant(proposal));
+        assert!(!context.opens_pass(proposal));
+        assert!(!context.gates_outbound(proposal));
+    }
+
+    #[test]
+    fn processed_commit_still_gates_while_its_durable_pass_is_active() {
+        let commit = input(
+            ConvergenceInputRole::CommitEdge,
+            7,
+            MessageState::Processed,
+            1,
+        );
+        let context = ConvergenceInputContext::from_inputs([commit]);
+
+        assert!(!context.gates_outbound(commit));
+        assert!(context.active_pass_member_gates_outbound(commit));
+    }
+}
--- a/crates/cgka-engine/src/convergence_input.rs
+++ b/crates/cgka-engine/src/convergence_input.rs
@@ -7,31 +7,23 @@
 
 use std::collections::{BTreeMap, BTreeSet};
 
+use cgka_traits::convergence_pass::{ConvergencePassMember, ConvergencePassMemberRole};
 use cgka_traits::message::MessageState;
 
 use crate::openmls_projection::OpenMlsContentKind;
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub(crate) enum ConvergenceInputRole {
-    CommitEdge,
-    ProposalDependency,
-    AppWitnessCandidate,
-}
-
-impl ConvergenceInputRole {
-    pub(crate) fn for_content_kind(kind: OpenMlsContentKind) -> Option<Self> {
-        match kind {
-            OpenMlsContentKind::Commit => Some(Self::CommitEdge),
-            OpenMlsContentKind::Proposal => Some(Self::ProposalDependency),
-            OpenMlsContentKind::Application => Some(Self::AppWitnessCandidate),
-            OpenMlsContentKind::Welcome | OpenMlsContentKind::Other => None,
-        }
+pub(crate) fn role_for_content_kind(kind: OpenMlsContentKind) -> Option<ConvergencePassMemberRole> {
+    match kind {
+        OpenMlsContentKind::Commit => Some(ConvergencePassMemberRole::CommitEdge),
+        OpenMlsContentKind::Proposal => Some(ConvergencePassMemberRole::ProposalDependency),
+        OpenMlsContentKind::Application => Some(ConvergencePassMemberRole::AppWitnessCandidate),
+        OpenMlsContentKind::Welcome | OpenMlsContentKind::Other => None,
     }
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub(crate) struct ClassifiedConvergenceInput {
-    pub(crate) role: ConvergenceInputRole,
+    pub(crate) role: ConvergencePassMemberRole,
     pub(crate) source_epoch: u64,
     pub(crate) state: MessageState,
     pub(crate) digest: [u8; 32],
@@ -45,7 +37,7 @@ impl ClassifiedConvergenceInput {
         digest: [u8; 32],
     ) -> Option<Self> {
         Some(Self {
-            role: ConvergenceInputRole::for_content_kind(kind)?,
+            role: role_for_content_kind(kind)?,
             source_epoch,
             state,
             digest,
@@ -75,12 +67,28 @@ impl ConvergenceInputContext {
     ) -> Self {
         let mut context = Self::default();
         for input in inputs {
-            if input.role == ConvergenceInputRole::CommitEdge {
+            if input.role == ConvergencePassMemberRole::CommitEdge {
                 context
                     .commit_edges_by_source_epoch
                     .entry(input.source_epoch)
                     .or_default()
                     .insert(input.digest);
+            }
+        }
+        context
+    }
+
+    pub(crate) fn from_pass_members<'a>(
+        members: impl IntoIterator<Item = &'a ConvergencePassMember>,
+    ) -> Self {
+        let mut context = Self::default();
+        for member in members {
+            if member.role == ConvergencePassMemberRole::CommitEdge {
+                context
+                    .commit_edges_by_source_epoch
+                    .entry(member.source_epoch)
+                    .or_default()
+                    .insert(member.payload_digest);
             }
         }
         context
@@ -111,9 +119,9 @@ impl ConvergenceInputContext {
         input: ClassifiedConvergenceInput,
     ) -> bool {
         match input.role {
-            ConvergenceInputRole::CommitEdge => true,
-            ConvergenceInputRole::ProposalDependency => self.has_commit_at(input.source_epoch),
-            ConvergenceInputRole::AppWitnessCandidate => {
+            ConvergencePassMemberRole::CommitEdge => true,
+            ConvergencePassMemberRole::ProposalDependency => self.has_commit_at(input.source_epoch),
+            ConvergencePassMemberRole::AppWitnessCandidate => {
                 self.has_competing_commit_before(input.source_epoch)
             }
         }
@@ -122,9 +130,9 @@ impl ConvergenceInputContext {
     pub(crate) fn opens_pass(&self, input: ClassifiedConvergenceInput) -> bool {
         input.can_start_pass()
             && match input.role {
-                ConvergenceInputRole::CommitEdge => true,
-                ConvergenceInputRole::ProposalDependency => false,
-                ConvergenceInputRole::AppWitnessCandidate => {
+                ConvergencePassMemberRole::CommitEdge => true,
+                ConvergencePassMemberRole::ProposalDependency => false,
+                ConvergencePassMemberRole::AppWitnessCandidate => {
                     self.has_competing_commit_before(input.source_epoch)
                 }
             }
@@ -133,9 +141,9 @@ impl ConvergenceInputContext {
     pub(crate) fn gates_outbound(&self, input: ClassifiedConvergenceInput) -> bool {
         input.can_gate_outbound()
             && match input.role {
-                ConvergenceInputRole::CommitEdge => true,
-                ConvergenceInputRole::ProposalDependency => false,
-                ConvergenceInputRole::AppWitnessCandidate => {
+                ConvergencePassMemberRole::CommitEdge => true,
+                ConvergencePassMemberRole::ProposalDependency => false,
+                ConvergencePassMemberRole::AppWitnessCandidate => {
                     self.has_competing_commit_before(input.source_epoch)
                 }
             }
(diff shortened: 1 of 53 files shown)


crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -259,19 +300,355 @@ impl<S: StorageProvider> Engine<S> {
...
+    }
+
+    fn verify_frozen_pass_members_in(
+        storage: &S,
+        pass: &DurableConvergencePass,
+    ) -> Result<(), FrozenPassVerificationError> {
+        for member in &pass.members {
+            let record = storage
+                .get_message(&member.message_id)
+                .map_err(FrozenPassVerificationError::Storage)?;
```
erskingardner: **A missing frozen member does not reach the fail-closed path.**

The PR description says "Fail closed and mark the group unrecoverable if a frozen member is missing **or** its digest changes." Only the digest case does that. `SqliteAccountStorage::get_message` returns `StorageError::NotFound` for a missing row (`crates/storage-sqlite/src/storage/messages.rs:58`), which this line maps to `FrozenPassVerificationError::Storage`, and both call sites turn that into `Err(storage_projection_error(error))` rather than `mark_group_unrecoverable_for_frozen_pass`.

The consequence is worse than a mis-tagged error. In the app worker that `Err` lands in the `advance_convergence_after_runtime_sync` error arm and goes to `schedule_retry_groups`, i.e. exponential backoff capped at `CONVERGENCE_RETRY_MAX_DELAY` (60s) and retried forever. Because the pass stays `Frozen`/`Resolving` — and therefore `is_active()` — `has_unresolved_convergence_inputs` keeps returning `true`, so every outbound send for that group stays gated too. The group never converges again, never emits `GroupUnrecoverable`, and never surfaces to the user as anything but a recurring "scheduled convergence failed" error.

Suggest discriminating the variant here:

```rust
let record = storage.get_message(&member.message_id).map_err(|error| match error {
    StorageError::NotFound => FrozenPassVerificationError::Integrity,
    other => FrozenPassVerificationError::Storage(other),
})?;
```

`frozen_pass_member_tampering_fails_closed_to_unrecoverable` only mutates a payload, so the missing-row case is untested — worth a sibling test that deletes the row.

---

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -259,19 +300,355 @@ impl<S: StorageProvider> Engine<S> {
...
+                    pass.clock_instance_id = self.convergence_clock_instance_id;
+                    pass.opened_monotonic_ms = 0;
+                    pass.quiescence_deadline_monotonic_ms = policy.settlement_quiescence_ms;
+                    pass.absolute_deadline_monotonic_ms = policy.max_convergence_pass_ms;
+                    self.storage
+                        .put_convergence_pass(&pass)
+                        .map_err(storage_projection_error)?;
+                    return Ok(Some(pass));
+                }
+                #[cfg(not(feature = "test-policy-overrides"))]
```
erskingardner: **The production restart path is compiled out of the entire test matrix.**

This `cfg(not(feature = "test-policy-overrides"))` block — rebasing monotonic deadlines from persisted wall-clock milliseconds, with the backwards-jump fail-closed — is the whole of "reconstruct collecting passes safely after restart". But `just test` runs `cargo nextest run --workspace --features wn-cli/test-policy-overrides` (`Justfile:4,42,46`; `.github/workflows/ci.yml:172`), and that feature unifies down `wn-cli → marmot-app → cgka-session → cgka-engine`. So every test in the matrix compiles the `cfg(feature = ...)` sibling above instead, which does something materially different: it resets `opened_monotonic_ms` to 0 and grants a **fresh full window** rather than preserving the remaining one.

The one restart test that touches passes, `rebuilt_engine_emits_canonical_app_message_after_convergence`, hand-writes the pass to `Frozen` before rebuilding the engine, so it never crosses the `Collecting` rebase at all. `just fast-ci`'s only default-feature test target is `convergence_policy_pin`, which doesn't cover this either.

Net: the code that runs in production on every restart with a live pass has zero coverage, and the code that is covered has different semantics. Two options that would fix it — make the rebase runtime-selected off the policy rather than `cfg`, or add a default-features integration test (alongside `test-convergence-policy-pin`) that opens a `Collecting` pass, rebuilds the engine with a fresh `convergence_clock_instance_id`, and asserts the *remaining* window is preserved and that a backwards wall jump makes the cutoff immediately due.

---

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -259,19 +300,355 @@ impl<S: StorageProvider> Engine<S> {
...
+        group_id: &GroupId,
+        base_epoch: EpochId,
+        policy: &CanonicalizationPolicy,
+    ) -> Result<(Vec<ConvergencePassMember>, bool), OpenMlsProjectionError> {
+        let ceiling = base_epoch
+            .0
+            .saturating_add(policy.convergence.max_rewind_commits);
+        let projected = self
+            .storage
+            .list_messages(group_id, EpochId(0))
```
erskingardner: **The seed has an upper epoch bound but no lower one, so `members` is the group's entire retained history.**

`ceiling` drops anything above `base_epoch + max_rewind_commits`, but nothing drops anything below `retained_anchor_epoch`. The state filter includes `MessageState::Processed`, and canonicalization does *not* retire below-anchor commits — `openmls_projection.rs:799` just `continue`s past `source_epoch < state.retained_anchor_epoch` and only records a `BeyondAnchor` drop for `unresolved_commit_state` rows. Processed rows keep their state, so they are re-seeded into every future pass forever.

Three costs compound:

1. `members` grows unbounded with group age, and the whole `DurableConvergencePass` is serialized into a single BLOB row (`0035_durable_convergence_passes`).
2. `admit_stored_message_to_convergence_pass` calls `put_convergence_pass(&pass)` per admitted message, so admitting *n* messages into a pass rewrites that full BLOB *n* times — O(n · history) bytes per pass.
3. `verify_frozen_pass_members` does a row read plus a SHA-256 for every member on every run.

For a long-lived group this is a per-inbound-message cost proportional to total history, where the pre-PR live enumeration only walked records without persisting or hashing them. Mirroring the anchor floor here (`source_epoch >= base_epoch.saturating_sub(policy.convergence.max_rewind_commits)`) would bound the set to what canonicalization can actually use, at the cost of nothing — the below-anchor rows are `continue`d downstream anyway.

---

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -351,19 +728,108 @@ impl<S: StorageProvider> Engine<S> {
...
+                    self.mark_group_unrecoverable_for_frozen_pass(group_id, previous_tip)?;
+                    return Ok(unrecoverable_result(previous_tip.0));
+                }
+            }
+        }
+        if matches!(
+            pass.phase,
+            ConvergencePassPhase::Frozen | ConvergencePassPhase::Resolving
+        ) {
+            match self.verify_frozen_pass_members(&pass) {
```
erskingardner, line 950: Every freeze verifies the full member set twice. The block above sets `pass.phase = Frozen` and verifies inside `with_transaction`; control then falls straight into this `matches!(Frozen | Resolving)` block, which verifies the identical set again. Combined with the unbounded seed noted above, that's 2 × (row read + SHA-256) per historical message on the freeze path.

Guarding this second pass on the freeze not having just happened (or moving the freeze verification out of the transaction and keeping only one call site) would halve it.

---

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -580,6 +1063,11 @@ impl<S: StorageProvider> Engine<S> {
         );
 
         self.remember_canonicalization_result_messages(&result);
+        pass.phase = ConvergencePassPhase::Completed;
```
erskingardner: **"Workflow completion is transactionally persisted" doesn't hold, and two early returns skip completion entirely.**

The apply happens at line 990 (`apply_openmls_canonicalization_result_with_profile_policy`) and this completion write lands ~75 lines later as a separate, unbatched `put_convergence_pass`. A crash in between leaves the pass `Resolving` with the epoch already advanced and members already `Processed`. Re-resolving the same frozen set should be idempotent via `already_seen`, so I think this is recoverable — but the safety-properties list in the PR body claims it's transactional, and it isn't. Either wrap apply + completion together or soften the claim.

More concretely, two paths `return Ok(result)` *before* reaching here:

- `MissingRetainedAnchor` (line ~972) — group is marked unrecoverable, so the stranded `Resolving` pass is mostly moot, though the row lingers. Same for `mark_group_unrecoverable_for_frozen_pass`, which also leaves the pass active.
- `ConvergenceStatus::Blocked` (line ~987) — this one is not terminal. The pass stays `Resolving`, so `is_active()` stays true, so `has_unresolved_convergence_inputs` gates every outbound send for the group for as long as the block lasts. Meanwhile no new input can be admitted (`admit_stored_message_to_convergence_pass` requires `Collecting`), so arrivals pile up for generation N+1 behind a pass that can't advance. It does self-heal once the block clears and the stale frozen set resolves, but the gating window is now unbounded rather than bounded by `max_convergence_pass_ms`. Worth an explicit decision on whether `Blocked` should complete or abandon the pass.

---

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -74,11 +78,48 @@ fn convergence_run_context(run_id: &str, phase: ConvergencePhase) -> AuditEventC
...
+}
+
+impl From<StorageError> for FrozenPassVerificationError {
+    fn from(error: StorageError) -> Self {
+        Self::Storage(error)
+    }
+}
+
 impl<S: StorageProvider> Engine<S> {
+    pub fn convergence_cutoff_delay_ms(
```
erskingardner: This reads as a query and is documented as one in the trait ("Applications use this to arm independent per-group timers; it is not a branch-selection input"), but it is not read-only. Through `load_or_open_convergence_pass` it can:

- open a brand-new pass and `put_convergence_pass` it — i.e. **asking when the cutoff is starts the clock**;
- consume a `fairness_slot_available` flag;
- rebase and persist clock deadlines after a restart.

It's also `&self` on the inherent impl while doing all of that, which hides the mutation from the type system (the trait method is `&mut self`, so it compiles, but the inherent signature is misleading).

The app worker calls this once per pending group inside `schedule_pending_convergence_groups`, so pass creation is now a side effect of scheduling rather than of ingest. That may well be intended — the pass should start when the group is known to have buffered input — but it isn't stated anywhere, and it makes the method unsafe to call from any diagnostic/read path (CLI, TUI, tests). Either document the side effect on both the trait method and here, or split it: a pure `convergence_cutoff_delay_ms` that returns `None` when no pass exists, plus an explicit open at the ingest/scheduling seam.

Minor, same function: `ensure_group_live` failures are mapped into `OpenMlsProjectionError::Storage`, so an ordinary quarantined-or-unrecoverable group looks like a storage fault; `client/sync.rs` then `.ok().flatten()`s it into a silent fallback delay. A distinct variant (or returning `Ok(None)`) would keep real storage faults visible.

---

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -604,7 +604,10 @@ async fn run_app_runtime_account_worker(
                             match client.advance_convergence_after_runtime_sync(&group_id).await {
                                 Ok(summary) => {
                                     publish_app_runtime_summary(&events, &account_id_hex, &account_label, &summary);
-                                    scheduled_convergence.schedule_groups(client.take_pending_convergence_groups());
+                                    schedule_pending_convergence_groups(
+                                        &mut scheduled_convergence,
+                                        &mut client,
+                                    );
```
erskingardner, line 614: **The cutoff-derived deadline computed here is overwritten four lines later.**

`schedule_pending_convergence_groups` installs `now + cutoff_delay + margin` for each pending group, and then `scheduled_convergence.schedule_after_pass(&group_id, client.has_pending_convergence_inputs(&group_id))` runs last and wins for `group_id`:

- `has_pending_convergence_inputs` → `has_unresolved_convergence_inputs`, which this PR changed to return `true` for **any** active pass. So a pass still in `Collecting` — exactly the case where the durable cutoff is the right answer — always takes the `schedule_unsettled_groups` branch, which overwrites the deadline with a flat `normal_delay` (~1.1s).
- That branch also bumps `unsettled_rearm_attempts`. Past `CONVERGENCE_UNSETTLED_MAX_REARMS` it falls through to `retry_delay_for_attempt`, i.e. exponential backoff up to 60s — for a pass whose absolute cap is 5s.

So on the main scheduled-convergence path the new `convergence_cutoff_delay_ms` API never applies; it only takes effect at the moment a group first buffers. Either `schedule_after_pass`/`schedule_unsettled_groups` should consult the cutoff delay too, or `schedule_pending_convergence_groups` should run after `schedule_after_pass` and take a `min` with any existing deadline rather than unconditionally `insert`ing.

Related: `schedule_groups_with_delays`, `schedule_retry_groups` and `schedule_unsettled_groups` all `insert` unconditionally, so any of them can push a group's deadline *later* than one already recorded. The new test only covers the cross-group case (`scheduling_one_group_never_postpones_an_earlier_group_cutoff`); a same-group "never postpones its own cutoff" case would pin the intended semantics.

---

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1716,80 +1729,123 @@ impl ScheduledConvergence {
...
-            .reset(TokioInstant::now() + IDLE_CONVERGENCE_TIMER_DELAY);
-        self.groups.drain().collect()
+        let Some(earliest) = self.deadlines.values().copied().min() else {
+            self.reset_timer_to_earliest();
+            return Vec::new();
+        };
+        let ready: Vec<GroupId> = self
+            .deadlines
+            .iter()
+            .filter(|(_, deadline)| **deadline <= earliest)
```
erskingardner: `earliest` is the `min` of the deadlines, so `**deadline <= earliest` is an equality test — this returns only the single earliest cohort, not everything that is actually due.

When the timer fires late (a slow `sync_runtime_groups().await`, a busy runtime, a suspended process) several deadlines can have elapsed together. Each extra one costs a full extra trip around the loop, including another `client.sync_runtime_groups().await`, before it is picked up. It does converge — `reset_timer_to_earliest` arms an already-past instant that fires immediately — so this is throughput, not a hang.

`**deadline <= TokioInstant::now()` is the predicate that matches the intent, with the current `earliest` behaviour as the fallback when nothing is due yet (a spurious wake shouldn't return an empty vec and drop the arm).

---

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -454,9 +504,12 @@ impl<S: StorageProvider> Engine<S> {
...
                 }
-                if self.has_unresolved_convergence_inputs(group_id)? {
-                    return Ok(false);
-                }
+                // A completed frozen batch grants one queued user intent before
+                // an inbound-only follow-up pass. Retry one deferred-peel sweep
+                // first so newly available canonical context is visible, then
+                // return to the drain rather than opening generation N+1 here.
+                let _ = self.retry_deferred_peels(group_id).await?;
+                return Ok(true);
```
erskingardner: This changes the contract of `advance_convergence_inputs_until_settled` from "true once nothing is unresolved" to "true after one settled run", and I think it grants the fairness edge a second time.

Before, the loop kept going and only returned `true` when `!has_unresolved_convergence_inputs`. Now the first `has_unresolved` branch always returns, so `MAX_CONVERGENCE_REPROCESSING_PASSES` is unreachable for it, and `true` is returned even when a completed pass has left retained inputs for generation N+1 (`has_unresolved_convergence_inputs` would still say `true`).

The caller only consults this when `fairness_slot_available` is **false**, so the effect is: one queued intent of *any* kind slips past the convergence gate per completed pass, in addition to the durable slot's one *admin* intent. The next loop iteration re-enters, opens generation N+1, gets `waiting_result` → `Ok(false)` → `break`, so it's bounded at one — but that means the documented "a completed frozen batch grants one queued user intent" is implemented twice, in two places, with different eligibility rules (`is_admin_group_state_fairness_intent` here vs. anything there).

If that's deliberate, the comment should say which of the two is the fairness edge and why both are needed. If not, gating this `return Ok(true)` on the fairness slot would collapse it to one mechanism.

---

crates/traits/src/ingest.rs
```diff
@@ -27,6 +32,23 @@ pub enum IngestOutcome {
     Rejected { category: ProposalRejectionCategory },
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum InputRejectionCategory {
```
erskingardner, line 36: Splitting the taxonomy leaves five `StaleReason` variants that are no longer constructed anywhere in production code — `AlreadySeen`, `OwnEcho`, `NotForThisClient`, `UnknownGroup`, `Quarantined`, `SelfEvicted` now only appear in `audit_helpers::stale_reason_str` and in doc comments. `PeelFailed` and the epoch reasons are the only ones still produced.

Since `StaleReason` is a public `cgka_traits` type, any downstream `match` on those arms now silently never fires rather than failing to compile — the worst kind of API break. Worth either removing them (the compiler then finds every stale matcher, including the doc references in `peeler.rs:103`, `bounded_id_set.rs:4`, `message_disposition.rs:33`, `engine.rs:186,199`, and `cgka-conformance-simulator/tests/proptest_invariants.rs:6`) or `#[deprecated]`-marking them for a release so consumers get a warning.

---

erskingardner (commented):
Reviewed the full diff (44 files, +2881/−267) against `master` at `e40fa5f`, focusing on the engine pass state machine, the SQLite backing, the app-worker scheduling change, and the ingest-outcome taxonomy split.

The core design is sound: moving the batch boundary from a live re-enumeration to a durable frozen member set is the right shape for #806, the `Collecting → Frozen → Resolving → Completed` phases are clean, and admitting the stored row and its pass membership in one transaction is the right durability boundary. Nesting `admit_stored_message_to_convergence_pass` inside `with_transaction` is safe — `SqliteConnection::with_transaction` is same-thread reentrant (`connection.rs:176`) — and the two digest sites agree (`Sha256::digest(&message.payload)` vs `projection.message_digest`, both `message_digest(bytes)` over the same slice).

Nine inline comments. The two I'd want resolved before this leaves draft:

1. **A missing frozen member never reaches the fail-closed path.** `get_message` returns `StorageError::NotFound`, which maps to `FrozenPassVerificationError::Storage` and propagates as a hard `Err` — not `mark_group_unrecoverable_for_frozen_pass`. The app worker retries that forever with 60s-capped backoff while the still-active pass gates every send, so the group silently stops converging instead of surfacing `GroupUnrecoverable`. Only the digest-tampering case (which is the one with a test) works as described.

2. **The production restart path is compiled out of CI.** The wall-clock rebase lives under `cfg(not(feature = "test-policy-overrides"))`, but the whole test matrix runs `--features wn-cli/test-policy-overrides`, which unifies into `cgka-engine`. Tests exercise the `cfg(feature)` sibling, which grants a *fresh* window rather than preserving the remaining one — different semantics. The one restart test hand-writes the pass to `Frozen` first, so it never crosses the `Collecting` rebase. "Reconstruct collecting passes safely after restart" is currently unproven.

The rest, roughly in order of how much they'd bother me: the seed set has no lower epoch bound so `members` accumulates the group's whole history (and is re-serialized per admitted message); the new `convergence_cutoff_delay_ms` is overwritten by `schedule_after_pass` on the main scheduled path, so the API mostly doesn't take effect; `convergence_cutoff_delay_ms` mutates durable state despite being documented and shaped as a query; completion isn't actually transactional with the apply and `ConvergenceStatus::Blocked` strands the pass in `Resolving`; the fairness edge appears to be granted twice by two different mechanisms; `take_ready` returns only the strict-minimum cohort; freeze verifies all members twice; and six `StaleReason` variants are now unconstructible without deprecation.

Nothing here looks structural — the failure modes are all at edges the happy path doesn't touch, which is consistent with the suite being green.

---

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -259,22 +321,494 @@ impl<S: StorageProvider> Engine<S> {
...
+                payload_digest: projection.message_digest,
+            },
+            input,
+        }))
+    }
+
+    fn classified_convergence_pass_inputs(
+        &self,
+        pass: &DurableConvergencePass,
+    ) -> Result<Vec<ClassifiedConvergenceInput>, OpenMlsProjectionError> {
```
erskingardner: **The missing-member fix is unreachable from the real convergence entry point — this `?` wedges first.**

`verify_frozen_pass_members_in` now correctly maps `NotFound → Integrity`, and `missing_frozen_pass_member_fails_closed_to_unrecoverable` proves it. But that test calls `converge_stored_openmls_messages` directly. Every production caller goes through `advance_convergence` → `converge_and_drain_queued_outbound_intents` → `advance_convergence_inputs_until_settled` → `has_unresolved_convergence_inputs` → `convergence_pass_gates_outbound` → `classified_convergence_pass_inputs` → `convergence_pass_candidate`, and that path hits `storage.get_message(...).map_err(storage_projection_error)?` before `run_convergence_pass` is ever entered.

I ran the same tamper as the new test against the real entry point:

```
PROBE advance_convergence -> Err(Backend("classify active convergence pass: storage failed: NotFound"))
PROBE unrecoverable = false
```

So a missing member wedges scheduled convergence, sends, and the `marmot-account` maintenance safety gate (`runtime.rs:1149` `?`-propagates) on a repeating hard error, with no `GroupUnrecoverable` event and no recovery path — exactly the failure mode the fix was meant to remove, relocated one layer up. The `.ok_or_else(...)` arm below has the same shape for a member that tampering made unclassifiable.

It also contradicts the fail-open discipline documented at `message_processor/mod.rs:600` (mdk#736): "a row we cannot decode ... is NOT resolvable convergence work — treating it as 'unresolved' would let a single corrupt/garbage row permanently gate sends with no recovery path."

Suggested fix: on the gate path, fail **closed on the gate but not on the call** — treat an unresolvable member as gating (`Ok(true)`) rather than `Err`. Convergence then runs, `verify_frozen_pass_members` classifies it as `Integrity`, and the group transitions to unrecoverable as designed. A regression test that drives this through `advance_convergence` rather than `converge_stored_openmls_messages` would have caught it.

---

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -259,22 +321,494 @@ impl<S: StorageProvider> Engine<S> {
...
+        if candidate.input.source_epoch < floor || candidate.input.source_epoch > ceiling {
+            // Retain out-of-horizon input for a later base epoch, but never let
+            // it enter (or enlarge) the immutable batch for this pass.
+            return Ok(ConvergenceAdmissionOutcome::Retained);
+        }
+        let admitted_input = candidate.input;
+        pass.members.push(candidate.member);
+        let context =
+            ConvergenceInputContext::from_inputs(self.classified_convergence_pass_inputs(&pass)?);
+        // Restart quiescence only for an input role that can change this
```
erskingardner, line 434: The role classification is the right call, but rebuilding the context costs a full re-read of the member set every time.

`classified_convergence_pass_inputs` does `get_message` + `StoredMessagePayload::decode` + `project_mls_message` (a TLS deserialize) for **every** member, and it runs here on every admission plus inside `convergence_pass_gates_outbound` on every `has_unresolved_convergence_inputs` call — which is every send, every drain, every maintenance tick, and every hydration check.

The seed is now bounded to `[base ± max_rewind_commits]` (= ±5), which fixes the unbounded-history problem, but that bounds *epochs*, not *messages*: an active chat group can hold thousands of app messages inside an 11-epoch window, and they are all members. So a single outbound send can trigger thousands of row reads and TLS deserializations to answer one question — "are there competing commit edges below epoch N?" — that depends only on the commit members.

`ConvergencePassMember` currently carries `message_id` + `payload_digest`. Adding `role` and `source_epoch` would make `ConvergenceInputContext` constructible from the pass record alone with zero storage reads. Both are derived from the digest-pinned payload and re-derivable at freeze, so it adds no trust assumption — the digest still pins the bytes those fields came from. Caching the context per generation would work too, but the record fields are simpler and survive restart.

---

crates/cgka-engine/src/convergence_input.rs
```diff
@@ -0,0 +1,259 @@
...
+                    .insert(input.digest);
+            }
+        }
+        context
+    }
+
+    fn has_commit_at(&self, source_epoch: u64) -> bool {
+        self.commit_edges_by_source_epoch
+            .get(&source_epoch)
+            .is_some_and(|edges| !edges.is_empty())
```
erskingardner, line 100: Worth confirming this is the intended breadth: `from_inputs` records a `CommitEdge` for every commit member regardless of `MessageState`, including `Processed`. So a commit race that already resolved keeps `edges.len() > 1` for as long as both commits stay inside the `[base ± max_rewind_commits]` window.

The consequence is that on any group with a resolved race in the last ~5 epochs, app messages go back to opening passes and gating sends — i.e. the pre-PR behaviour returns for exactly the groups that had contention. I think that's *defensible*: the rewind horizon means a settled selection can still be re-litigated, and an app message is genuine witness evidence for that, so being conservative here is sound. But it's a real carve-out in the improvement and it isn't stated anywhere.

Two things that would help:
- If losing-branch commits transition to `EpochInvalidated` on apply, they'd fall out of the state filter and the window would self-narrow quickly — worth confirming that's what happens, since it determines whether this is a rare edge or the common case on churn-heavy groups.
- A test pinning the intended behaviour after a race resolves (does an app message at epoch N+1 still gate once the winner is applied?) would make the decision explicit either way.

---

erskingardner (commented):
Re-reviewed `f259527` and `404dd2c` on top of my earlier pass. This is a strong revision — every one of the nine findings is addressed, and the role classifier is a better design than what I'd have proposed.

**Fixed, with tests:**

| Finding | Fix |
| --- | --- |
| Missing member not fail-closed | `NotFound → Integrity` (but see below) |
| Restart rebase compiled out of CI | `cfg(not(...))` removed; rebase runs in every build; `collecting_pass_restart_preserves_remaining_window_and_backward_clock_fails_closed` |
| Unbounded seed | `[base ± max_rewind_commits]` floor and ceiling |
| Double verify at freeze | `just_frozen` guard |
| `convergence_cutoff_delay_ms` mutates | renamed `prepare_*`, side effects documented on the trait |
| Cutoff delay overwritten by scheduler | `schedule_after_pass` reordered ahead of `schedule_pending_convergence_groups`; all arms use `arm_no_later` |
| `take_ready` strict-min cohort | `deadline <= now` with an `earliest` fallback for spurious wakes |
| Completion not transactional / `Blocked` strands pass | apply + completion in one `with_transaction`; `Blocked` now completes the pass |
| Dead `StaleReason` variants | `#[deprecated]` with migration notes |

Also picked up the incidental asymmetry — the admit path now applies the same `[floor, ceiling]` horizon the seed does, with `far_future_app_is_not_admitted_to_an_active_pass` covering it.

**On the role classifier:** the three-way split is right, and `AppWitnessCandidate` gated on `has_competing_commit_before` is a sharper answer than "app messages never gate" — it keeps app messages relevant exactly when there's a commit race they could witness. `application_input_does_not_reset_convergence_quiescence` and `app_witness_beside_competing_commits_resets_convergence_quiescence` pin both directions. This also fixes the maintenance-obligation starvation transitively, since `runtime.rs:1149` gates on the now role-aware `has_pending_convergence_inputs`.

One correction to my earlier review: I called the `return Ok(true)` in `advance_convergence_inputs_until_settled` a redundant second fairness grant. Tracing it properly, the two mechanisms were complementary rather than duplicated — but making that return conditional on the durable slot is the cleaner collapse regardless, and that's what landed.

**One new issue, and it's the same class as the one that was just fixed.** `has_unresolved_convergence_inputs` now calls `convergence_pass_gates_outbound` → `classified_convergence_pass_inputs`, which `?`-propagates a `NotFound` on any member row. That check runs *before* `converge_stored_openmls_messages` on every production path, so the `Integrity → unrecoverable` transition is unreachable in exactly the state it was added for. Verified by running the new test's tamper through `advance_convergence` instead of `converge_stored_openmls_messages`:

```
PROBE advance_convergence -> Err(Backend("classify active convergence pass: storage failed: NotFound"))
PROBE unrecoverable = false
```

Details and a suggested fix inline. Two lower-priority notes also inline: the per-send cost of rebuilding the input context from storage, and a question about `has_competing_commit_before` counting `Processed` commits.

---

PR #1111, opened 2026-07-25
Expose encrypted Blossom group image bindings

erskingardner:
## Summary

- expose `update_group_image` through UniFFI so Swift and Kotlin clients can encrypt, upload, and commit `marmot.group.blossom.image.v1`
- add an explicit `clear_group_image` binding instead of requiring an empty-byte sentinel
- reject empty upload bytes with the existing typed media error
- align the workspace, lockfile, and all conformance fixtures with the `0.9.5` release cohort

## Why

Marmot app/runtime already implements the encrypted group-image flow, including fresh content key/nonce generation, opaque Blossom upload with an image-specific signer, and the admin-gated MLS component commit. The mutation was not available through MarmotKit even though initial-image creation and encrypted-image download were already exposed.

Generated clients receive approximately:

```swift
updateGroupImage(
    accountRef: String,
    groupIdHex: String,
    plaintext: Data,
    mediaType: String
) async throws -> SendSummaryFfi

clearGroupImage(
    accountRef: String,
    groupIdHex: String
) async throws -> SendSummaryFfi
```

## Validation

- `just fast-ci`
- `cargo test -p marmot-uniffi group_image`
- `cargo test -p cgka-conformance-simulator canonical_vector_fixtures_match_generated_traces`
- `git diff --check`
- generated Swift XCFramework and binding signature verification
- generated Kotlin binding and arm64 Android JNI build

An exploratory full `cargo test -p marmot-app` run passed all 432 library tests but encountered five failures in untouched `relay_runtime` encrypted-media, retention-projection, and unread-state cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty group-image uploads are now rejected with a clear validation error.
  * Group identifiers are validated before updating or clearing group images.
  * Users are directed to clear the image when removing an avatar.

* **Conformance**
  * Updated conformance metadata to version 0.9.5 across simulator scenarios.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ members = [
 
 [workspace.package]
 edition = "2024"
-version = "0.9.7"
+version = "0.9.5"
 license = "MIT"
 publish = false
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agent-connector"
-version = "0.9.7"
+version = "0.9.5"
 dependencies = [
  "agent-control",
  "agent-stream-compose",
@@ -72,7 +72,7 @@ dependencies = [
 
 [[package]]
 name = "agent-control"
-version = "0.9.7"
+version = "0.9.5"
 dependencies = [
  "serde",
  "serde_json",
@@ -82,7 +82,7 @@ dependencies = [
 
 [[package]]
 name = "agent-stream-compose"
-version = "0.9.7"
+version = "0.9.5"
 dependencies = [
  "cgka-traits",
  "hex",
@@ -852,7 +852,7 @@ checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
 name = "cgka-conformance-simulator"
-version = "0.9.7"
+version = "0.9.5"
 dependencies = [
  "async-trait",
  "cgka-engine",
@@ -881,7 +881,7 @@ dependencies = [
 
 [[package]]
 name = "cgka-engine"
-version = "0.9.7"
+version = "0.9.5"
 dependencies = [
  "async-trait",
  "cgka-traits",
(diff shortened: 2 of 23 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ members = [
 
 [workspace.package]
 edition = "2024"
-version = "0.9.7"
+version = "0.9.5"
```
erskingardner, line 29: Workspace version regresses 0.9.7 -> 0.9.5 here. `master` (merge-base `d7236f43`) is already at 0.9.7 — this reverts it, and the same downgrade is repeated in `Cargo.lock` and all 19 conformance vector fixtures' `conformance_version` fields. Unrelated to this PR's stated purpose; please drop these version changes.

erskingardner (commented):
The `update_group_image` / `clear_group_image` additions in `crates/marmot-uniffi/src/commands/group.rs` (plus the `smoke.rs` test) are clean and minimal: thin wrappers that match the existing binding pattern, delegate admin enforcement to the engine like their sibling mutations (`update_group_profile`, `update_group_avatar_url`), and correctly reuse the already-documented "empty plaintext clears the image" contract from `AppClient::update_group_image` (`crates/marmot-app/src/client/mod.rs:1916`). No concerns with that part.

However, the PR also rewrites `Cargo.toml`, `Cargo.lock` (every internal crate's version), and all 19 `crates/cgka-conformance-simulator/vectors/**/*.json` fixtures, changing the workspace version from `0.9.7` down to `0.9.5`. `master` has been at `0.9.7` since well before this branch's merge-base (`d7236f43`, the current tip of `master`) — this is a straight downgrade, not a stale-rebase artifact. Merging as-is silently reverts the in-flight workspace version already on `master`. Per this repo's convention (bump `Cargo.toml` only at release time, via a dedicated release PR), a feature PR like this shouldn't touch the version at all.

Please drop the version/vector-stamp changes and keep only the `marmot-uniffi` diff.

Cargo.toml
erskingardner: this is on purpose.

PR #1112, opened 2026-07-26
feat: add public Blossom profile image binding

erskingardner:
## Summary

- add an account-signed public Blossom upload API for profile images and expose it through UniFFI
- validate raster magic bytes, media-derived fallback extensions, returned URL safety/integrity, and same-site descriptor hosts
- add first-class `banner` support while preserving omitted banners in partial profile updates
- repair key-package device ownership so `local` is derived from durable, Welcome-usable OpenMLS private bundles
- deterministically merge local ownership with relay events without dropping distinct relay event ids
- bump the workspace version from 0.9.5 to 0.9.6

## Why

White Noise needs to accept a local or web-selected profile image, upload it to Blossom, and publish the resulting URL in the user's kind-0 metadata. Profile images remain intentionally public and separate from encrypted group-image media.

MarmotKit 0.9.5 also exposed a key-package ownership regression: newly published or automatically rotated packages could appear relay-only even though this device generated them. The listing now proves ownership from the durable OpenMLS bundle required to consume a Welcome, including while signed out and after restart.

## Behavior and API fallout

- `UserProfileMetadataFfi.banner` is optional with a generated default of `nil`/`null`, so existing Swift and Kotlin constructor call sites remain source-compatible.
- An omitted banner preserves the current banner. A present banner replaces it.
- Public profile uploads use the configured `MarmotServiceEndpoints.profile_image_blob_endpoint`; the built-in default is `https://blossom.primal.net`, and an explicit per-call server still wins.
- `AccountKeyPackageFfi.local` is true only when the matching private bundle is durably stored and usable. `relay` continues to describe relay discovery of that exact event id.
- No secret material is exposed through FFI.

## Validation

- `just fast-ci`
- `cargo test -p marmot-app --lib` — 441 passed
- `cargo test -p marmot-account --test runtime` — 33 passed
- `cargo test -p cgka-engine --test group_creation` — 49 passed
- focused signed-out/restart relay-runtime ownership test passed
- `cargo test -p marmot-uniffi` — all unit, fixture, and smoke suites passed
- focused `wn-cli` key-package cardinality/merge test passed
- `cargo test -p cgka-conformance-simulator` — complete simulator/vector/property suite passed
- `./crates/marmot-uniffi/kotlin-bindings.sh` — all Android ABIs built

Generated Swift and Kotlin both expose `banner` with a `nil`/`null` default. Generated artifacts remain ignored and are not committed.


erskingardner (commented):
## Overview

Three things bundled together:

1. **New public (unencrypted) Blossom upload path** for profile images — `upload_profile_image` in `crates/marmot-app/src/media/mod.rs`, exposed via `Marmot.upload_profile_image` in UniFFI. `upload_blossom_blob` is refactored into a thin wrapper over a new `..._with_content_type` variant so the raster content type reaches the PUT.
2. **`banner` promoted from an "unknown extra" field to a first-class field** on `UserProfileMetadata` + `UserProfileMetadataFfi`.
3. Workspace version bump 0.9.5 → 0.9.6.

The refactor is clean, the pre-network validations run before any bytes leave, and the new path reuses `media_http_client_for_url`, so it inherits the dial-safety discipline (address pinning, `reject_non_public_ip`, timeouts) required by `AGENTS.md`. The version bump is correct here — `crates/marmot-uniffi/AGENTS.md` explicitly requires it when UniFFI records or object methods change.

---

## Blocking: the `banner` promotion silently deletes existing banners

Before this PR, `banner` was an unknown kind:0 field, so it landed in `UserProfileMetadata::extra`. In `merge_user_profile_update` (`crates/marmot-app/src/runtime/mod.rs`), `extra` is only replaced when the incoming update's `extra` is non-empty — and the FFI conversion hardcodes `extra: Default::default()`. So **every FFI-driven profile publish preserved the user's banner**, whether or not the caller knew banner existed.

After this PR, `banner` is a known field and the merge does an unconditional overwrite:

```rust
current.picture = update.picture;
current.banner = update.banner;   // blind overwrite
```

Any caller that publishes a profile with `banner: nil` — the natural port for an app that only edits name/about/picture, and what every Rust caller using `..UserProfileMetadata::default()` gets — now **wipes the banner out of the user's published kind:0 event**. That's user-visible data loss on a real Nostr profile, triggered by an unrelated profile edit.

The test change is the tell. `runtime_profile_publish_preserves_unknown_kind0_fields` in `crates/marmot-app/tests/relay_runtime.rs` previously proved banner survived a second update that didn't mention it. This PR adds `banner: Some("https://example.test/banner.png")` to that second update, so the assertion now passes for a completely different reason. The regression is masked rather than caught.

Options, roughly in order of preference:

- Keep the blind-overwrite semantics (consistent with `picture`) but **add an explicit test** asserting `banner: None` in an update clears the published banner, so the semantics are deliberate and documented — and flag it in the PR body as a behavior change host apps must handle.
- Or make the merge distinguish "absent" from "cleared" for banner (e.g. `if update.banner.is_some() { current.banner = update.banner }`), accepting that it then can't be cleared through this path.
- Either way, restore an assertion somewhere that a partial update does not lose previously-published metadata.

On the FFI side: `banner` has no `#[uniffi(default = None)]`, so it's a **breaking constructor change** for Swift/Kotlin — existing MarmotKit callers won't compile until they add the field. That's arguably the safer outcome (it forces app authors to notice), but it's a breaking bindings change on a patch bump and belongs in the PR description. Do *not* add a default here without fixing the merge first — a default would turn a compile error into the silent wipe above.

---

## Should fix

**`.bin` extension on the fallback URL.** When the Blossom server's descriptor omits `url`, `upload_blossom_blob_with_content_type` falls back to `blossom_blob_url(server, hash)`, which hardcodes `.bin`. That's correct for opaque encrypted blobs, but here it produces `https://blossom.primal.net/<hash>.bin` as a **public profile picture URL published to every Nostr client**. Most will fail to render it, and some servers will 404 the wrong extension outright. The profile path needs either the bare-hash URL or an extension derived from the validated media type.

**Media type is label-validated, not content-validated.** The PR body says "validate raster media types" and the test is named `profile_image_upload_rejects_non_raster_and_oversized_inputs_before_network` — but the check is purely on the caller-supplied `media_type` string. `upload_profile_image(svg_bytes, "image/png", ...)` sails through, and the resulting URL is published as a public picture that other clients fetch. A magic-byte sniff for the four accepted formats is ~15 lines and would make the code match its own claim.

**The entire post-network path is untested and untestable as written.** `allow_loopback_http` is hardcoded `false` at the `upload_blossom_blob_with_content_type` call, so the existing loopback mock-server helpers (`spawn_http_response`, used with `true` in `crates/marmot-app/src/media/tests.rs`) can't reach this function. That leaves zero coverage for descriptor handling, the `.bin` fallback, and — notably — the new `validate_blossom_fetch_url` post-check, which is the only security control on the returned URL. Threading the flag through (defaulting `false`, `true` only in tests) would make all of it testable.

**The descriptor URL host isn't pinned to the server you uploaded to.** `upload_blossom_blob_with_content_type` accepts whatever `descriptor.url` the server returns as long as the path contains the matching hash; `blossom_content_hash_from_url` only scans the path for 64 hex chars, with no host check. This is pre-existing, but it's newly consequential: for encrypted media a hostile URL is neutralized by hash + AEAD at fetch time, whereas here the URL is **published verbatim into the user's kind:0 and fetched unauthenticated by every other client**. `validate_blossom_fetch_url` only proves it's https-on-a-public-host. Worth pinning the returned host to the upload host (or the same registrable domain, matching the rule already in `validate_blossom_redirect_host`).

**New hardcoded third-party endpoint with no fallback.** `DEFAULT_PROFILE_IMAGE_BLOSSOM_SERVER_URL = "https://blossom.primal.net"` is a brand-new dependency that isn't in `DEFAULT_BLOSSOM_SERVER_URLS`, has no fallback list (encrypted media has three endpoints with fallback ordering), and isn't sourced from `MarmotAppConfig` the way other endpoint policy is. Note this upload is authorized by the **account's own signer**, so primal.net learns the pubkey↔IP association for every profile image upload — unlike group images, which use a freshly generated ephemeral keypair. Both facts are defensible for a deliberately-public artifact, but the endpoint choice deserves a comment explaining why primal specifically, and it should probably be configurable through `MarmotAppConfig` rather than only per-call.

---

## Nits

- `canonicalize_marmot_media_type` already exists in `cgka_traits::app_components` and is what `upload_group_image` uses. The hand-rolled match here rejects parameterized types (`image/png; charset=utf-8`) that the shared canonicalizer handles. Canonicalize first, then allowlist.
- Style: `url::Url::parse` and `host_safety::validate_blossom_fetch_url` are fully qualified in `media/mod.rs` while the module otherwise imports its helpers (`use host_safety::validate_locator;`).
- No test covers the migration itself — an old cached record with `banner` inside `extra` deserializing into the new field. It does work correctly (serde `flatten` matches known fields first, and `profile_content_json` filters `extra` through `is_known_profile_field` so a stale entry can't shadow the field), but that's exactly the kind of thing worth pinning with a test.

---

## Process

- `crates/marmot-uniffi/AGENTS.md` requires regenerating bindings after API changes (`xcframework.sh`, `kotlin-bindings.sh`). The Validation section doesn't mention either — worth running given a `uniffi::Record` gained a field.
- The PR notes "five existing shared-relay timing/state integration cases failed." Those live in `crates/marmot-app/tests/relay_runtime.rs`, which this PR edits. Worth confirming they fail on `master` too rather than assuming they're unrelated.

erskingardner (commented):
Fresh pass over `27b97dce`. Two new commits since my last review — `cea86b2f` (preserve key package ownership across publication) and `27b97dce` (keep inactive key package listings relay-readable). CI is green on everything that has reported.

## Scope note

The PR is titled and described as the Blossom profile-image binding, but two thirds of the diff is now unrelated key-package ownership work (45 files, +697/−124). The description hasn't been updated to mention it. These are independently reviewable and independently revertable — I'd split them, or at minimum describe the key-package change in the body so it isn't merged as a drive-by.

## Status of the previous findings

None of them are addressed — `media/mod.rs`, `directory/records.rs`, and the UniFFI conversions are byte-identical to `227e50ac`. In particular `merge_user_profile_update` still does `current.banner = update.banner`, so the blocking banner-clobber finding stands as written. Flagging so it isn't lost under the new commits.

---

## New work: durable key-package ownership

The core idea is good. Proving ownership by round-tripping the bundle through OpenMLS's own lookup-by-`KeyPackageRef` — rather than trusting lifecycle metadata — is the right invariant, and it fails closed. The new tests are the strongest part of this PR: missing bundle, corrupt bundle, publish-retry, and restart are all covered, and `missing_or_corrupt_private_bundle_is_not_reported_as_owned` pins exactly the case that matters.

### 1. `is_active_signing()` makes `local` wrong, not just absent

`account_key_package_records` now passes `owned = Vec::new()` for any account where `is_active_signing()` is false — which is `(local_signing || external_signing) && !signed_out`, so it includes every signed-out account.

`local_key_package_records` builds its entire output from `owned_key_packages`, so those accounts get **zero local records**, and every row in the listing reports `local: false`. The private bundles are still sitting in the SQLCipher database. Before this PR the legacy JSON record was read straight off disk with no signing requirement, so a signed-out account reported `local: true` correctly.

So the flag has quietly changed meaning from "this device holds the private material" to "this device holds it *and* the account is actively signing." That's a worse answer than omitting the row. This looks like a plumbing constraint leaking into semantics — `durably_owned_key_packages` goes through the account worker, which needs a live worker. Options: read ownership directly from storage for inactive accounts, keep the legacy-record fallback when `owned` is empty, or rename/document the field so `local: false` on a signed-out account isn't read as "not on this device."

### 2. The legacy override is unconditional and runs last

In `local_key_package_records`:

```rust
if let Some(legacy) = legacy_record.as_ref().filter(|l| l.key_package_ref_hex == metadata.key_package_ref_hex) {
    key_package_id = legacy.key_package_id.clone();
    key_package_event_id = legacy.key_package_event_id.clone();
    published_at = legacy.published_at;
}
```

Every field on `KeyPackageRecord` is `#[serde(default)]`, so `key_package_event_id: ""` and `published_at: 0` are both representable. This block overwrites the freshly-derived lifecycle values with them unconditionally — a legacy file written before publication completed would blank out `lifecycle.authored_event_id`.

For a commit whose goal is preserving ownership across publication, lifecycle is the live source of truth and should win. Gap-fill reads much safer here:

```rust
if key_package_event_id.is_empty() { key_package_event_id = legacy.key_package_event_id.clone(); }
```

If the override ordering is deliberate — e.g. to keep a pre-migration event id that lifecycle doesn't know about — it needs a comment saying so.

### 3. `merge_key_package_records`: semantics changed, still zero unit tests

The dedup key changed from a single canonical key (`event_id` → `ref_hex` → `id`, via `BTreeMap`) to an order-dependent linear scan matching on `event_id` **or** `ref_hex`. Two consequences:

- **Secondary event ids are silently dropped.** Two relay events carrying the same key package used to produce two records keyed by event id; they now collapse into one, and `if existing.key_package_event_id.is_empty()` means only the first event id survives. Since `keys delete <event-id>` sources its id from this listing, any additional event for that key package becomes unreachable through the CLI. If collapsing is the intent, `AccountKeyPackageRecord` needs to carry a list of event ids rather than one.
- **The OR-predicate is non-transitive**, so output depends on input ordering: a record matching an existing entry on `ref_hex` can backfill its `event_id`, after which a third record that would have matched on `event_id` may or may not find it depending on where it lands in the sequence. The old single-key map couldn't do this by construction.

`merge_key_package_records` has never had a direct unit test — it's only exercised transitively through relay integration tests. A rewrite that changes dedup identity, merge precedence, and ordering behavior is exactly where that gap bites. Please add table-driven unit tests in `key_package_records.rs` covering: local+relay for one ref, two relay events for one ref, records with empty `ref_hex`, and permuted input orders producing identical output.

(It's also O(n²) now. Irrelevant at current scale — noting it only so it isn't a surprise later.)

### 4. CLI: loosened assertion, and no CHANGELOG entry

`keys_list_reports_published_key_package` went from:

```rust
assert_eq!(keys.len(), 1, "expected the published key package, got {keys:?}");
```

to filtering `relay == true` and asserting *that subset* is 1. The total is now unbounded — a regression that starts emitting fifty retained bundles passes this test unchanged, and the assertion message still says "expected the published key package." Please assert the total count as well (or a bound), so the new cardinality is pinned rather than merely tolerated.

This is the same shape as the banner test change from the first commit: an assertion relaxed to accommodate a behavior change instead of a new assertion pinning it.

Relatedly — `keys list` now returns more rows (retained private material, pending replacements) and the `local` flag changed meaning per finding 1. Both are user-visible CLI behavior changes, and `crates/cli/CHANGELOG.md` still has an empty `[Unreleased]` section. `release.md` designates that file for CLI-visible changes.

---

## Nits

- **The doc comment overstates what the code checks.** `durably_owned_key_packages` says "The storage row key is part of the ownership proof… a decodable bundle stored under any other key is therefore not usable" — but nothing compares `stored.storage_key` against the serialized `KeyPackageRef`. It re-reads through the OpenMLS provider instead. That's equivalent for the stored-only-under-a-wrong-key case, but a bundle present under *both* a correct and a wrong key passes twice and yields a duplicate entry (currently masked downstream by `merge_key_package_records`). A direct `storage_key` comparison would match the doc, drop the second storage read, and be duplicate-free by construction.
- **Five silent `continue` paths, no telemetry.** Skipping is the right call for a read-only query, but a key package silently vanishing from a user's list is undiagnosable. A privacy-safe aggregate (`skipped = N`, no refs or ids) fits the observability invariant in `AGENTS.md`.
- `use openmls_traits::storage::StorageProvider as OpenMlsStorageProvider;` is declared inside the loop body — hoist to module scope.
- `key_package_metadata` is computed twice per package: the engine calls it to validate and discards the result, then `local_key_package_records` recomputes and `?`s on a failure the engine already filtered out. Returning metadata from the engine would avoid the double TLS decode and make the unreachable `?` unnecessary.
- **`key_package_id` fallback semantics.** With no lifecycle match and no legacy record it becomes `metadata.key_package_ref_hex`, where it previously always came from the stable slot id. A public DTO field now sometimes carries a ref hex and sometimes a slot id. Worth a doc comment on `AccountKeyPackageRecord::key_package_id` if that's intended.
- **Perf shape changed.** `account_key_package_records` went from one `read_json` to enumerating every stored bundle with a JSON decode + `hash_ref` + a storage read each, on every call. Fine at today's scale; worth knowing if retained private material grows.
- **`rusqlite` dev-dependency on `marmot-account`.** The corruption test reaches into `openmls_values` and hardcodes the `x'4b65795061636b616765'` label to simulate a bad bundle. It works, but it pins a `marmot-account` test to `storage-sqlite`'s internal table and label layout — a schema rename now breaks a test two crates away, and the repo's crate map puts persistence ownership in `storage-sqlite`. A test-only corruption helper on `SqliteAccountStorage` would keep that coupling inside the crate that owns the schema.

PR #1113, opened 2026-07-26
fix(hermes): route commentary as agent activity

### LINKED ISSUE #1087: Hermes adapter: route non-final assistant segments as agent activity, not kind-9 finals
## Problem

The Marmot Hermes adapter maps generic assistant sends and stream finalization to durable kind-9 messages even when the payload is only a non-final assistant segment. MDK then correctly classifies kind 9 as standard message traffic, so Android correctly routes it to the group-message channel instead of the Agent activity bucket.

Tool progress is already explicit and maps to agent-operation kind 1202. The missing case is plain/structured assistant output before a tool or continuation.

Verified against `origin/master` at `87f89884bab1bff8ecc0ae6a61bfb88b4c3e4140`:

- `integrations/hermes/marmot/adapter.py::MarmotPlatformAdapter.send` handles tool progress specially, but otherwise begins/finalizes a preview or calls `_send_final_direct`.
- `edit_message(..., finalize=True)` finalizes a preview as a durable final without knowing whether Hermes sealed only a pre-tool segment.
- `metadata` is accepted by `send` but does not carry/drive semantic finality.
- `render_message_event(Commentary)` can call `send_agent_activity`, but Hermes's typed dispatcher is not wired in production (NousResearch/hermes-agent#54453).
- The live Hermes consumer knows `is_turn_final`, but its platform boundary does not expose it.

This is why only some intermediate output appears as ordinary group messages: operation events already have a typed path, while assistant segments fall through the final-message path.

## Dependency

Hermes must expose a stable semantic delivery context to adapters, including turn finality/delivery class. Track that contract in: https://github.com/NousResearch/hermes-agent/issues/70694

The Marmot adapter should remain compatible with older Hermes versions: missing metadata must default to the existing final/direct-send behavior rather than guessing from text, length, sender, or timing.

## Required mapping

```text
non-final commentary / assistant segment
  → send_agent_activity
  → kind 1201

tool lifecycle
  → send_agent_operation_event
  → kind 1202

turn-final answer or approval requiring user response
  → stream_finalize / send_final
  → kind 9
```

Do not deduplicate or classify by message text; identical text can be legitimate across distinct turns.

## Acceptance criteria

- `turn_final=false` never calls `send_final` or commits a kind-9 stream final.
- Non-final assistant text is published as kind 1201 with `status=commentary` and the triggering prompt relationship when available.
- Tool lifecycle remains kind 1202.
- `turn_final=true` retains current kind-9 streaming/final behavior.
- Missing delivery metadata defaults to final for compatibility.
- Android PR marmot-protocol/whitenoise-android#1595 receives the correct MDK traffic class without Android text heuristics or binding changes.
- Streaming, non-streaming, direct-send, and no-consumer fallback paths produce the same classification.

## Test coverage

Extend `integrations/hermes/marmot/tests` with:

- commentary before a tool → one kind-1201 activity, zero kind-9 finals;
- structured commentary and plain pre-tool segment coverage;
- tool operation → kind 1202;
- final answer after commentary → exactly one kind 9;
- missing metadata → current final behavior;
- same text under two logical turns remains two legitimate events;
- reply/parent metadata survives the activity mapping.

## Related

- marmot-protocol/whitenoise-android#1595 — Android routes explicit Agent activity correctly.
- NousResearch/hermes-agent#54453 — typed event dispatcher is not wired into the live gateway.

agent-p1p:
Fixes #1087

## Summary
- consume canonical `delivery_class` / `is_turn_final` context (plus legacy `turn_final`) and route sealed commentary through `send_agent_activity`
- make explicit turn finality authoritative over every delivery class while retaining class-based and missing-metadata compatibility defaults
- keep edit and draft previews transient, emitting exactly one kind-1201 activity when a non-final segment seals while preserving its prompt relationship
- retain preview state after a retryable activity failure and clean it up only after confirmed activity delivery
- retain kind-9 behavior for finals, approvals, and missing metadata; keep tool lifecycle on kind 1202
- validate activity responses and document the updated delivery behavior

## Verification
- `python3.11 -m unittest discover -s integrations/hermes/marmot/tests` — 107 passed
- `integrations/hermes/marmot/tests/test_dev_scripts.sh` — passed
- `just fast-ci` — passed
- isolated real-Hermes deterministic E2E — passed (`final_text: marmot-e2e-ok: ping from marmot`)
- `python3.11 -m py_compile integrations/hermes/marmot/adapter.py integrations/hermes/marmot/tests/test_adapter.py` — passed
- `git diff --check` — passed
- independent read-only Composer review — passed; no blocking security or logic findings

## Dependency
Hermes must still expose the delivery context tracked in NousResearch/hermes-agent#70694. Missing metadata intentionally keeps the existing final/direct-send behavior.

## Sensitive paths
None.

## Summary by CodeRabbit

* **New Features**
  * Added metadata-aware routing for commentary, previews, drafts, final answers, and approvals.
  * Commentary is delivered as durable activity while previews remain transient until finalized.
  * Preserved reply-to threading for commentary messages.

* **Bug Fixes**
  * Prevented duplicate final messages when streamed content is finalized.
  * Maintained legacy final-send behavior when delivery metadata is unavailable.
  * Improved handling of missing activity responses with retryable failures.

* **Documentation**
  * Clarified delivery, preview, finalization, threading, and compatibility behavior.

### DIFF
--- a/integrations/hermes/marmot/README.md
+++ b/integrations/hermes/marmot/README.md
@@ -345,12 +345,19 @@ directory with `--media-allowed-root`; without one, path sends fail closed.
   the Marmot group id and `user_id` set to the sender account id.
 - Hermes progressive edits open at most one live preview per chat at a time. A
   newer preview cancels the previous stream for that chat.
+- Explicit Hermes delivery context routes non-final commentary to durable agent
+  activity (`kind: 1201`, `status: commentary`). Preview and draft frames remain
+  transient; sealing the segment cancels its preview and emits one activity.
+- Final answers and approvals remain durable `kind: 9` messages. Missing delivery
+  context keeps this legacy final behavior for compatibility with older Hermes
+  versions.
 - Durable message text is exactly what Hermes hands the adapter. The adapter
   never merges, splices, or reconstructs text across sends; fragmented or
   duplicated finals must be fixed in Hermes message segmentation.
-- A plain send while a preview is open finalizes that stream as one tagged
-  stream-final `kind: 9` when the final text is an append-only extension of the
-  streamed text. There is no duplicate plain `send_final` for the same text.
+- A final or legacy plain send while a preview is open finalizes that stream as
+  one tagged stream-final `kind: 9` when the final text is an append-only
+  extension of the streamed text. There is no duplicate plain `send_final` for
+  the same text.
 - Otherwise the preview is cancelled and the final goes out verbatim as one
   plain `send_final`.
 - Status records are included in the stream transcript hash and chunk count.
--- a/integrations/hermes/marmot/adapter.py
+++ b/integrations/hermes/marmot/adapter.py
@@ -1268,6 +1268,7 @@ class MarmotLiveStream:
         stream_id_hex: str,
         stream_capability: str,
         start_message_id_hex: str,
+        parent_message_id_hex: Optional[str],
         chunk_bytes: int,
     ):
         self.client = client
@@ -1276,6 +1277,10 @@ class MarmotLiveStream:
         self.stream_id_hex = stream_id_hex
         self.stream_capability = stream_capability
         self.start_message_id_hex = start_message_id_hex
+        self.parent_message_id_hex = _optional_hex(
+            parent_message_id_hex,
+            "parent_message_id_hex",
+        )
         self.text = AppendOnlyTextState()
         self.transcript = AgentTextStreamTranscript(
             stream_id_hex,
@@ -1332,6 +1337,7 @@ class MarmotLiveStream:
             stream_id_hex=response["stream_id_hex"],
             stream_capability=_normalize_stream_capability(response["stream_capability"]),
             start_message_id_hex=response["start_message_id_hex"],
+            parent_message_id_hex=parent_message_id_hex,
             chunk_bytes=effective_stream_chunk_bytes(
                 chunk_bytes,
                 response.get("policy_max_plaintext_frame_len"),
@@ -1545,7 +1551,16 @@ class MarmotPlatformAdapter(BasePlatformAdapter):
                 reply_to_message_id_hex=_optional_hex(reply_to),
             )
 
-        if self._looks_like_stream_preview(content):
+        is_preview = self._looks_like_stream_preview(content)
+        if _delivery_routes_to_commentary_activity(metadata) and not is_preview:
+            return await self._send_commentary_activity(
+                chat_id,
+                visible_content,
+                reply_to=reply_to,
+                metadata=metadata,
+            )
+
+        if is_preview:
             if not self.quic_candidates:
                 return SendResult(success=False, error="Marmot live preview requires MARMOT_QUIC_CANDIDATES")
             try:
@@ -1609,6 +1624,7 @@ class MarmotPlatformAdapter(BasePlatformAdapter):
         content: str,
         *,
         finalize: bool = False,
+        metadata: Optional[Dict[str, Any]] = None,
     ) -> SendResult:
         self._capture_loop()
         if message_id.startswith(TOOL_PROGRESS_MESSAGE_PREFIX):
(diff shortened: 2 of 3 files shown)


agent-p1p:

### Blocking (1)

1. **`integrations/hermes/marmot/adapter.py:3310-3318` — delivery class overrides the explicit turn-finality invariant in both directions.**

   `_delivery_routes_to_commentary_activity` returns from `delivery_class` before consulting `is_turn_final`. At this PR head I reproduced:

   - `{is_turn_final: false, delivery_class: notice}` → `activities=0`, `final_sends=1`
   - `{is_turn_final: true, delivery_class: commentary}` → `activities=1`, `final_sends=0`

   That conflicts with #1087 acceptance criteria: explicit `turn_final=false` must never call `send_final` or commit kind 9, while explicit `turn_final=true` must retain kind-9 behavior. It also sends `operation`, `notice`, and any future delivery class as kind 9 whenever they carry explicit non-finality. The tests at `test_adapter.py:2203` and `test_adapter.py:2224-2237` currently codify the contradictory class-overrides-finality behavior.

   Resolve precedence against the stable upstream contract. At minimum, do not emit kind 9 for an explicitly false finality flag and do not demote an explicitly true finality flag to activity. Add conflict/full-enum regression cases so the hard invariant remains visible.

### Suggestion (1)

- **`integrations/hermes/marmot/tests/test_adapter.py:2183-2441` — the added coverage exercises generic `send`/`edit_message`, but not the structured `Commentary` renderer or the no-consumer fallback named in #1087.** Add contract-level coverage once the upstream delivery context is available, including an exactly-once assertion so typed rendering plus generic fallback cannot double-publish the same segment.

### Verified

- `python3 -m unittest discover -s integrations/hermes/marmot/tests` — 103 passed
- `integrations/hermes/marmot/tests/test_dev_scripts.sh` — passed
- PR ownership verified: `agent-p1p`, branch `pip/mdk-1087`, head `54e4f9820164c3caec06a66648a53695ace8e47f`

Tool-operation routing remains ahead of commentary classification, missing metadata retains legacy final behavior, and reply/parent preservation across preview sealing is otherwise well covered.

agent-p1p:
## Adversarial review — supplemental blocking finding

### Blocking (2 of 2)

2. **`integrations/hermes/marmot/adapter.py:1642-1655` — a transient activity-send failure destroys the only retryable stream handle before delivery succeeds.**

I independently reproduced the completed CodeRabbit finding on this exact head:

```text
preview True marmot-stream:5555…
first  False retryable=True  error=transient
second False retryable=False error=Marmot cannot edit durable messages
activity_attempts=1 activities=0 cancels=1
```

The first `edit_message(... finalize=True, metadata=commentary)` calls `_cancel_stream`, which removes `_active_streams[message_id]`, and only then attempts `send_agent_activity`. When that attempt returns a retryable failure, retrying the same advertised operation cannot reach the activity send at all. One transient connector failure can therefore drop the sealed commentary segment.

Preserve sufficient stream/retry state until activity delivery is confirmed, or provide an equivalent retry-safe transition, and add a fail-once/same-message-id retry regression. Do not blindly add internal retries without considering ambiguous post-write outcomes and idempotency; #1088 tracks the broader logical-delivery identity/reconciliation requirement.

Updated adversarial counts: **2 blocking, 1 suggestion, 0 nitpicks**.

agent-p1p:
Addressed both blocking findings in a8460e4.

- Explicit `is_turn_final` / legacy `turn_final` now wins over every known or unknown `delivery_class` in both directions. Missing finality still uses `commentary` only; all other or missing classes retain final behavior.
- Commentary preview finalization now preserves the same stream/message id after a retryable activity failure, then cancels it only after the retry succeeds. The regression proves two activity attempts, one delivered activity, zero kind-9 finals, and cleanup after success.
- Added focused structured `Commentary` renderer coverage. Typed-renderer + generic-fallback exactly-once correlation remains deferred to the logical delivery identity tracked by #1088 / NousResearch/hermes-agent#70694; text deduplication would violate the issue contract.

Verification: 107 adapter tests, dev scripts, `just fast-ci`, plugin smoke, deterministic real-Hermes E2E, py_compile, diff check, and an independent read-only Composer review all pass.

agent-p1p:
CodeRabbit’s bounded in-method activity retry suggestion is intentionally not applied: `send_agent_activity` has no stable logical-delivery/idempotency key, so retrying an ambiguous post-write failure can duplicate kind-1201 activity. This patch instead preserves the advertised stream/message handle on retryable failure so the caller can retry the same operation; #1088 tracks safe logical-delivery reconciliation.

agent-p1p:
@coderabbitai review

Issue #1054, opened 2026-07-23
Select retained SelfRemove proposals deterministically for local commits


erskingardner:
## Summary

Make local SelfRemove-only Commit preparation conform to the adopted member-departure contract without changing ordinary convergence or electing a deterministic committer.

When multiple distinct valid retained SelfRemove proposals exist from the same leaving leaf and source epoch, local preparation selects the proposal with the lexicographically lowest digest:

```text
SHA-256(serialized_proposal_mls_message)
```

The input is the complete serialized handshake `MLSMessage`. Selection must not depend on arrival order, map iteration order, timing, or local identity.

This is a local proposal-selection rule, not a deterministic-committer or mandatory-batching rule. A SelfRemove-only Commit MAY reference one or more valid retained SelfRemove proposals. Selected proposals for multiple distinct leaving leaves MAY be included in one Commit, but batching every eligible proposal is not required.

All valid non-selected alternatives remain retained and processable for inbound Commit proposal-reference resolution. An alternative is not stale merely because this client would select another for a new local Commit.

## Normative source

Adopted specification commit:

- `marmot-protocol/marmot@dedeaf429a6f52a4f1c62f3e992be012dab61b7b`
- `protocol-core/member-departure.md`, especially **SelfRemove commits**

Related ordinary behavior remains governed by the adopted convergence, inbound-processing, and publish-before-apply rules.

## Dependencies and boundaries

Relevant existing work includes:

- #1053 authorize proposals before queueing
- #806 bounded convergence pass/frozen-batch scheduling
- #1027 convergence-v1 selection tracker

This issue does **not** change convergence scoring, introduce deterministic SelfRemove committer election, or require every eligible SelfRemove proposal to be batched.

## Start here

- `crates/cgka-engine/src/auto_committer.rs`
- `crates/cgka-engine/src/message_processor/ingest.rs`
- `crates/cgka-engine/src/message_processor/mod.rs`
- `crates/cgka-engine/src/distributed_convergence.rs`
- `crates/cgka-engine/src/app_components.rs`
- `crates/cgka-engine/src/pending_commit_guard.rs`
- SelfRemove engine and conformance tests

## Required behavior

- For each leaving leaf/source epoch, select the valid retained SelfRemove proposal with the lexicographically lowest digest of the complete serialized proposal `MLSMessage`.
- Make local selection independent of arrival order, wall-clock order, map iteration order, timing, and local identity.
[...]

PR #1114, opened 2026-07-26
Select retained SelfRemove proposals deterministically

erskingardner:
## Summary

Implements the adopted SelfRemove local-commit preparation contract from [`protocol-core/member-departure.md`, “SelfRemove commits”](https://github.com/marmot-protocol/marmot/blob/dedeaf429a6f52a4f1c62f3e992be012dab61b7b/protocol-core/member-departure.md#selfremove-commits).

- Selects the proposal with the lexicographically lowest `SHA-256(serialized_proposal_mls_message)` for each leaving leaf/source epoch.
- Rebuilds the selection from retained wire messages, so it is independent of arrival order, map order, timing, local identity, and restart.
- Stages only revalidated SelfRemove proposals. Selected proposals for multiple distinct leaves may share one commit; unrelated pending proposals cannot leak into it.
- Keeps every non-selected valid alternative retained for inbound proposal-reference resolution.
- Restores retry scheduling after restart and publish rollback while preserving publish-before-apply behavior.
- Leaves randomized per-member jitter, ordinary competing-commit convergence, and convergence scoring unchanged.

## Correction to the former issue wording

The adopted specification does **not** elect one deterministic SelfRemove committer, does **not** require every eligible SelfRemove to be batched, and does **not** make a valid unselected alternative stale. This PR implements deterministic *local proposal selection per leaving leaf/source epoch* only. Issue #1054 has been rewritten to match that contract.

## Test mapping

| Contract | Coverage |
| --- | --- |
| Lowest complete-MLSMessage digest; arrival-order independence | `auto_committer::tests::self_remove_selection_is_lowest_digest_per_leaf_for_every_arrival_order`; `selfremove_local_selection_uses_lowest_complete_message_digest_after_restart` |
| Non-selected alternative remains consumable by a peer commit | `non_selected_selfremove_alternative_remains_processable_by_peer_commit` |
| Byte-identical repeats remain duplicates | `selfremove_local_selection_uses_lowest_complete_message_digest_after_restart` |
| Multiple leaves; SelfRemove-only shape; unrelated proposal excluded | `multiple_leavers_stage_one_selfremove_only_commit_excluding_unrelated_proposals` |
| Leaver cannot commit its own proposal | existing `self_remove_committer_must_not_be_leaver` coverage in `mip03_guards` |
| Active-admin/last-admin protection | existing `admin_cannot_self_remove_*` and auto-committer guard coverage in `mip03_guards` |
| Competing committers use ordinary convergence | existing `add_then_self_remove_via_harness` canonical simulator scenario |
| Publish rollback retains retryable work | enhanced `selfremove_auto_commit_publish_failed_rolls_back_projection` |
| Restart selection is stable and does not duplicate commits | `selfremove_local_selection_uses_lowest_complete_message_digest_after_restart` |

## Validation

- `cargo fmt --check`
- `cargo test -p cgka-engine --test convergence_policy_pin --locked`
- `just fast-ci`

The literal default-feature `cargo test -p cgka-engine` still encounters the existing `deferred_peel_lifecycle` fixture mismatch: those tests install a non-pinned convergence policy without enabling `test-policy-overrides`. The same seven failures reproduce on clean `origin/master` at `1d6b9f1d`; the repository’s broader engine/CI test matrix explicitly enables that feature, while the separate locked pin test above verifies the default policy gate.

Refs #1054

### DIFF
--- a/crates/cgka-engine/src/message_processor/mod.rs
+++ b/crates/cgka-engine/src/message_processor/mod.rs
@@ -479,6 +479,41 @@ impl<S: StorageProvider> Engine<S> {
             .retain(|_, scheduled| &scheduled.group_id != group_id);
     }
 
+    /// Rebuild the in-memory jitter edges for durable current-epoch proposal
+    /// rows. Replay performs the authenticated SelfRemove/type/authorization
+    /// checks before staging; this scan only restores wakeups lost at restart
+    /// or after publish rollback.
+    pub(crate) fn restore_self_remove_auto_commit_schedules_for_group(
+        &mut self,
+        group_id: &GroupId,
+        source_epoch: EpochId,
+        now_ms: u64,
+    ) -> Result<bool, EngineError> {
+        let mut restored = false;
+        for record in self.storage.list_messages(group_id, source_epoch)? {
+            if record.epoch != source_epoch
+                || !matches!(
+                    record.state,
+                    MessageState::Created | MessageState::Retryable | MessageState::Processed
+                )
+            {
+                continue;
+            }
+            let Some((_message, projection)) = decode_openmls_wire_projection(&record.payload)
+            else {
+                continue;
+            };
+            if projection.kind != crate::openmls_projection::OpenMlsContentKind::Proposal
+                || projection.source_epoch != Some(source_epoch.0)
+            {
+                continue;
+            }
+            self.schedule_self_remove_auto_commit(group_id, &record.id, source_epoch, now_ms)?;
+            restored = true;
+        }
+        Ok(restored)
+    }
+
     fn self_remove_auto_commit_jitter_ms(
         &self,
         group_id: &GroupId,
--- a/crates/cgka-engine/src/auto_committer.rs
+++ b/crates/cgka-engine/src/auto_committer.rs
@@ -23,6 +23,7 @@
 use openmls::framing::Sender;
 use openmls::group::MlsGroup;
 use openmls::prelude::{LeafNodeIndex, Proposal, QueuedProposal};
+use std::collections::BTreeMap;
 
 /// Decision returned by the policy.
 pub(crate) enum AutoCommitDecision {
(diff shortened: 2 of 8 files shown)


crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -479,6 +479,41 @@ impl<S: StorageProvider> Engine<S> {
...
+                    MessageState::Created | MessageState::Retryable | MessageState::Processed
+                )
+            {
+                continue;
+            }
+            let Some((_message, projection)) = decode_openmls_wire_projection(&record.payload)
+            else {
+                continue;
+            };
+            if projection.kind != crate::openmls_projection::OpenMlsContentKind::Proposal
```
erskingardner, line 506: `decode_openmls_wire_projection` only checks `projection.kind == Proposal`, not `Proposal::SelfRemove` specifically, so this restores a scheduled self-remove-commit wakeup for *any* retained standalone proposal in the current epoch (Add/Update/Remove/etc.), not just SelfRemove ones. The real type filter only runs later in `retained_self_remove_proposals_for_local_commit` during replay, so this is not a correctness bug (non-SelfRemove candidates are filtered out and the replay loop reschedules/continues cleanly), but in a busy epoch with several non-SelfRemove proposals in flight it means one full `list_messages` rescan + `mls_group.process_message` pass per spurious schedule entry on every hydration/publish-rollback. A cheap `Proposal::SelfRemove` check here (mirroring the one in `retained_self_remove_proposals_for_local_commit`) would avoid the wasted churn. Non-blocking.

erskingardner (commented):
Solid, faithful implementation of the adopted `member-departure.md` "SelfRemove commits" contract (spec commit `dedeaf42`), matching issue #1054's (corrected) acceptance criteria closely — verified the spec text directly:

- `select_lowest_digest_per_leaf` (crates/cgka-engine/src/auto_committer.rs) implements the exact `SHA-256(serialized_proposal_mls_message)` lowest-digest-per-leaf rule, leaf-ordered for arrival/iteration-order independence, with a permutation-covering unit test.
- `retained_self_remove_proposals_for_local_commit` (message_processor/ingest.rs) recomputes the candidate set from durable storage on every replay/restart, decoupling selection from whichever proposal id triggered the wakeup — this is what makes restart/arrival-order independence actually hold, not just the digest comparator in isolation.
- Non-selected valid alternatives are correctly left untouched (no state mutation), matching "retain all of them for MLS proposal-reference resolution" — covered by `non_selected_selfremove_alternative_remains_processable_by_peer_commit`.
- `store.rs`'s duplicate classification and `PendingCommitCleanupGuard`'s move to `Vec<ProposalRef>` are both correctly scoped, and `publish.rs`'s new `clear_pending_proposals` call is necessary (and correctly guarded to self-remove-only pending commits) given `store_pending_proposal` residue isn't cleared by `clear_pending_commit` alone.
- The `mls_group.has_pending_proposals()` guard before staging correctly prevents unrelated pending proposals from leaking into the batch, matching "MUST NOT contain another proposal type."
- Test coverage directly maps to every contract line (arrival-order permutations, alternative retention, multi-leaver batching + unrelated-proposal exclusion, publish-rollback retry). Full CI (fmt, clippy, check, all 4 Rust test shards, simulator/vectors, Tamarin, convergence-policy-pin) is green.

One non-blocking nit posted inline: `restore_self_remove_auto_commit_schedules_for_group` schedules a wakeup for *any* retained standalone proposal in-epoch (not just SelfRemove), causing avoidable rescans in busy groups — not a correctness issue since replay re-filters by type, just a minor efficiency/naming tightening opportunity.

Recommend merge. (Posted as a comment review, not approve — GitHub blocks self-approval since this account is both author and reviewer here.)

Issue #1062, opened 2026-07-23
[Completed] Adopt Marmot PR #408 with a strict current-profile cutover


erskingardner:
## Completion status

The MDK implementation and release portion of the Marmot PR #408 current-profile cutover is complete.

Normative sources used by this program:

- original adopted merge: `marmot-protocol/marmot@625f8aae8096a17c6b11fde08ec8c2d582568619`;
- corrected push registry/proof follow-up: `marmot-protocol/marmot@dedeaf429a6f52a4f1c62f3e992be012dab61b7b`.

MDK, MarmotKit, and WN Agent v0.9.7 were published from commit `a4374b78f6e73b398fc195f120ff280410b7873c` on 2026-07-26. The MarmotKit release contains generated iOS XCFramework/Swift and Android Kotlin/JNI artifacts with checksums. The WN Agent release contains all platform binaries, OpenCode harnesses, Hermes/OpenClaw plugin bundles, checksums, and installers.

Downstream app adoption, independently built-client interoperability, migration communication, and coordinated production rollout remain open in #1061. Closing this implementation tracker does not claim that production app rollout is complete.

## Locked compatibility decisions

- Already-joined legacy groups remain usable by their current members, including messaging, supported group changes, and Media V1.
- MDK does not publish new legacy KeyPackages or create new legacy groups.
- Unpublished legacy KeyPackages and private `init_key` material are retired; cached or in-flight legacy Welcomes are rejected.
- Existing legacy groups cannot add or re-add members after cutover.
- There is no group migration/upgrade UX, and explicit rejoin is not a compatibility-release blocker.
- Every new KeyPackage/group is current profile with account proof `0x8009`, required Media V2 `0x800b`, and MDK agent-ready QUIC receive semantics.
- Current push owner proofs are event-shaped kind 451. Legacy groups retain only the explicitly specified compatibility proof forms.
- Kind 450 is account identity proof; kind 452 is reserved for multi-device authorization.
- Transponder never receives `owner_sig`; it has no signature-verification work in this rollout.
- Multi-device component `0x800a` remains draft and is excluded.

### Wire profile and compatibility

- [x] #1047, #1048, #1049, #1050, #1051, #1052, #1053, and #1055 — carriers, profile classification, strict cutover, resulting-state invariants, founding lifecycle, Media V2, proposal authorization, and source-epoch retention.
[...]

Issue #1061, opened 2026-07-23
Complete downstream current-profile adoption, interoperability, and production rollout


erskingardner:
## Summary

Track downstream adoption, independently built-client interoperability, and coordinated production rollout for the completed Marmot current-profile implementation.

The MDK implementation tracker #1062 is complete. MDK, MarmotKit, and WN Agent v0.9.7 were published from commit `a4374b78f6e73b398fc195f120ff280410b7873c` on 2026-07-26. Since that release, PRs #1134, #1135, #1138, #1139, #1141, #1142, and #1143 completed the next startup, pending-leave, chat-list, retention, conformance, Hermes, and OpenClaw cohort. PR #1142 bumped the workspace to 0.9.8 and regenerated Swift/Kotlin surfaces. No v0.9.8 release tags or artifacts exist yet, so downstream adoption must use the forthcoming tagged cohort rather than an untagged master build.

This issue remains the source of truth for proving that the four primary consumers—White Noise iOS, White Noise Android, Hermes, and OpenClaw—adopt and exercise the profile safely.

## Published artifact milestone

- [x] MDK v0.9.7 published and marked Latest.
- [x] MarmotKit v0.9.7 published with iOS XCFramework/Swift and Android Kotlin/JNI artifacts plus checksums.
- [x] WN Agent v0.9.7 published with 23 binary, harness, plugin, checksum, and installer assets.
- [x] `wn-agent-latest` advanced to the v0.9.7 commit and installer set.
- [x] All three immutable release tags resolve to `a4374b78f6e73b398fc195f120ff280410b7873c`.

## Completed protocol dependencies

- [x] #705 / PR #1120 — KeyPackage deletion uses reliable scoped relay connections.
- [x] #787 and #1060 / PR #1118 — QUIC publisher sequence, candidate/failover, receiver continuity, restart, and durable fallback lifecycle.
- [x] #918 / PR #1117 — active RustSec vulnerabilities, unsoundness findings, and yanked packages cleared; unmaintained-only follow-up remains #1116.
- [x] #1054 / PR #1114 — deterministic retained SelfRemove selection.
- [x] #990/#991 / PR #1115 — complete selector-rule proof/generated coverage and account-witness deduplication.
- [x] #806, #929, #970, #971, #1020, #1047-#1059, #950, #955, #1007, #1070, #1092, and #1093 completed.
- [x] Trackers #707, #918, #1027, and #1062 reconciled and closed.
[...]

PR #1120, opened 2026-07-26
Batch KeyPackage deletion publishes over scoped relay connections

### LINKED ISSUE #705: Key-package deletion publishes through a throwaway relay client — cold connects fail with 'transport error' where the live pool works

erskingardner:
## Summary

- add adapter-owned ordered batch publishing with a scoped write-only relay lease
- batch sign-out, wipe, and manual KeyPackage deletion through one account-bound signer and relay client after relay-safety validation
- preserve per-event acknowledgement, cache, and teardown semantics while correcting comments that implied a durable retry queue

## Corrected diagnosis

Issue #705 correctly identifies per-event Nostr SDK client construction as part of the pathology. Reusing only the client object is insufficient, however, because publish-only relays are intentionally removed after each existing single-event publish. The adapter therefore owns a scoped batch lease: it deduplicates the union of write relays, connects each once, reuses them for the ordered batch, and cleans them up on success, error, timeout, or cancellation. The lease does not inherit subscriptions or retain relays permanently, and concurrent leases are reference-counted.

## Lifecycle semantics

- `sign_out` attempts the whole deletion batch before stopping the account worker and keeps matching local KeyPackages when an event is not acknowledged.
- `wipe` reports publication failures but continues destructive local cleanup; once account keys are erased, there is intentionally no durable retry/outbox in this change.
- manual deletion uses the same batch path with one item.
- local cache entries are removed only for the exact matching event after at least one relay acknowledgement.
- every endpoint is still passed through the host relay-safety policy before any publish call; loopback remains dev-only.
- the account signer is isolated to the account-bound client and every event is pre-signed exactly once before network publication.

A durable post-wipe deletion outbox is outside the scope of this PR.

## API and bindings

The public Rust transport adapter API adds `NostrEventPublishRequest` and the default-compatible `NostrRelayClient::publish_events` method. There is no Marmot wire-format change and no UniFFI or generated-binding contract change.

## Validation

- `cargo test -p transport-nostr-adapter --features sdk` — 51 unit tests and 30 inbound-routing tests passed
- focused `marmot-app` deletion unit tests passed
- focused relay-runtime sign-out and wipe publication-failure tests passed
- `cargo test -p marmot-account` — 13 library, 30 home, and 33 runtime tests passed
- `cargo test -p marmot-uniffi` — 67 library, 2 media-fixture, and 24 smoke tests passed
- `cargo check -p marmot-app`
- `just fast-ci`
- `git diff --check`

The full `cargo test -p marmot-app` run passed all 445 library tests and the new relay-runtime cases, but the relay-runtime integration binary ended with five existing failures:

- `encrypted_media_endpoint_updates_are_full_replacement_and_admin_only`
- `encrypted_media_upload_sends_ciphertext_and_download_decrypts_plaintext`
- `relay_app_runtime_synthesizes_system_row_for_retention_change`
- `retained_media_rehydrates_a_retired_current_epoch_before_the_group_advances`
- `self_removal_suppresses_account_unread_while_peer_removal_preserves_it`

The first failure was reproduced unchanged in an isolated clean worktree at `origin/master` (`left: 3`, `right: 1`), confirming the suite is not fully green on the base revision.

Closes #705

Refs #1062

Refs #1061

### DIFF
--- a/crates/transport-nostr-adapter/src/sdk_client.rs
+++ b/crates/transport-nostr-adapter/src/sdk_client.rs
@@ -18,8 +18,8 @@ use tokio::time::timeout;
 use transport_nostr_peeler::{KIND_MARMOT_GROUP_MESSAGE, NostrTransportEvent};
 
 use crate::{
-    NostrPublishOutcome, NostrRelayClient, NostrRelayEvent, NostrSubscription,
-    NostrTransportAdapter,
+    NostrEventPublishRequest, NostrPublishOutcome, NostrRelayClient, NostrRelayEvent,
+    NostrSubscription, NostrTransportAdapter,
 };
 
 const SDK_RELAY_CONNECT_WAIT: Duration = Duration::from_secs(5);
@@ -40,6 +40,10 @@ const SDK_RELAY_PUBLISH_RETRY_BACKOFF: Duration = Duration::from_millis(600);
 /// that degraded case. Sized to still allow a slow relay one full connect plus
 /// send attempt (`SDK_RELAY_CONNECT_WAIT + SDK_RELAY_PUBLISH_WAIT`) with margin.
 const SDK_RELAY_PUBLISH_OVERALL_WAIT: Duration = Duration::from_secs(20);
+/// Whole-batch ceiling. Individual events retain the existing 20-second
+/// ceiling, while a pathological multi-event teardown cannot multiply that
+/// bound without limit.
+const SDK_RELAY_BATCH_OVERALL_WAIT: Duration = Duration::from_secs(60);
 
 /// Planned SDK subscription derived from a transport-adapter subscription.
 #[derive(Clone, Debug)]
@@ -91,6 +95,10 @@ pub struct NostrSdkRelayClient {
     client: Client,
     account_subscriptions: Arc<RwLock<HashMap<MemberId, Vec<SubscriptionId>>>>,
     publish_relay_refs: Arc<Mutex<HashMap<RelayUrl, usize>>>,
+    #[cfg(test)]
+    publish_connect_attempts: Arc<Mutex<HashMap<RelayUrl, usize>>>,
+    #[cfg(test)]
+    publish_release_attempts: Arc<Mutex<HashMap<RelayUrl, usize>>>,
     /// Per-account, per-relay subscription-registration outcomes accumulated
     /// since that account's last
     /// [`take_subscription_registrations`](Self::take_subscription_registrations)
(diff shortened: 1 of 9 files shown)


crates/transport-nostr-adapter/src/sdk_client.rs
```diff
@@ -440,6 +521,176 @@ impl NostrSdkRelayClient {
...
+        }
+
+        // Connect the union once. Per-event sends below reuse these scoped
+        // write-only relay connections and never install subscriptions.
+        let mut connects = JoinSet::new();
+        for endpoint in connectable {
+            let client = self.clone();
+            connects.spawn(async move { client.connect_publish_relay(endpoint).await });
+        }
+        while let Some(result) = connects.join_next().await {
```
erskingardner, line 688: `publish_prepared_batch` connects the full union of endpoints and drains that `JoinSet` to completion (no ack-awareness, no early exit) *before* any event is sent:

```rust
let mut connects = JoinSet::new();
for endpoint in connectable {
    let client = self.clone();
    connects.spawn(async move { client.connect_publish_relay(endpoint).await });
}
while let Some(result) = connects.join_next().await {
    ...
}
```

Compare with the pre-PR `publish_event`, which spawned one task per relay that did **connect + send together**, raced them in a single `JoinSet`, and called `abort_all()` (cancelling any relay still mid-connect) as soon as `required_acks` was met. That let a fast, healthy relay resolve a publish in tens of milliseconds even while a different configured relay was hung mid-connect.

Since `publish_event` is now implemented as a one-request call to `publish_events` (see below), **every caller of `publish_event` loses that early-abort-during-connect behavior**, not just the new deletion batch API. That includes the primary outbound message/commit/welcome path at `crates/marmot-app/src/relay_plane/mod.rs:944`, which is almost always called with `required_acks: 1` against a multi-relay endpoint list.

Concretely: publishing a message to relays `[A (healthy, fast), B (silently unreachable — SYN black-holed, e.g. NAT), C (healthy, fast)]` with `required_acks: 1` used to resolve in ~A's connect+send time. After this change it now always blocks for up to `SDK_RELAY_CONNECT_WAIT` (5s) — the time for B's connect attempt to individually time out — because the connect-phase loop has no exit condition tied to whether enough relays are already reachable/acking. That's precisely the "cold/degraded connect on one relay stalls the whole publish" pathology that #705's field evidence describes (NATed mobile networks), just moved from "throwaway client per call" to "connect-phase barrier per call."

None of the existing timing tests catch this: `publish_event_does_not_wait_for_silent_relays_once_required_acks_are_met` uses a relay that completes the WS handshake instantly and only goes silent on the send/ack, so it only exercises the (still-early-abortable) send phase, not the new connect-phase barrier.

Worth considering: keep per-relay connect+send racing (as before) for the common case, and reserve the "connect once, reuse across many sends" strategy specifically for when a request set actually has more than one event sharing the same relays (the real KeyPackage-deletion-batch scenario) — e.g. by having `publish_prepared_event` connect-then-send per relay (reusing a relay lease if already retained from a prior event in the batch) rather than the batch pre-connecting the whole union up front. At minimum, please add a timing test for the connect-hang + healthy-relay-present case, since it's the specific field-evidence pathology this PR is meant to fix.

erskingardner (commented):
## Summary

This PR does what it says for the KeyPackage-deletion path — batching deletions through one scoped, ref-counted write-relay lease is a real fix for the N-cold-clients pathology in #705, and the local-cache/partial-failure bookkeeping in `delete_key_package_events` looks correct (partial success preserved, cache only cleared on the exact acked event, zero-ack never confirms). The doc-comment corrections about not implying a durable retry queue are also appropriate and match the stated scope.

However, the mechanism used to implement the batch (splitting `NostrSdkRelayClient` publish into a separate "connect the whole union" phase followed by a "send" phase) is not confined to the new batch API — `publish_event` itself is now implemented in terms of `publish_events` with one request, so **every existing caller of `publish_event` inherits the new two-phase behavior**, including the main outbound message/commit/welcome fanout (`relay_plane/mod.rs:944`), push registration, and directory publishes. That changes latency/resilience characteristics well outside this PR's stated scope, and regresses exactly the failure mode #705 was filed about. See inline comment for details. Recommend addressing before merge.

Non-blocking: the defensive `Ok(_) => Err("relay acknowledged zero key package deletions")` branch in `delete_key_package_events` looked like dead code at first glance (the adapter's `finish_publish_outcome` only returns `Ok` when `accepted.len() >= required_acks`), but it's intentionally defensive against other `NostrRelayClient` implementations and is exercised by `manual_key_package_deletion_uses_batch_and_requires_ack_before_cache_removal` — no change needed.

PR #1130, opened 2026-07-26
fix(hermes): use persisted config for gateway readiness

### LINKED ISSUE #1127: fix(hermes): make installed Marmot config satisfy gateway readiness without duplicate service environment
## Summary

The `wn-agent-v0.9.7` one-line Hermes installer can finish successfully, enable `platforms.marmot`, and start `wn-agent-hermes.service`, but Marmot still fails to load in an already-installed/systemd-managed Hermes gateway.

This is the remaining half of #838. The config-level enablement fix merged in #1043 makes the gateway attempt adapter creation, but the plugin's separate requirements check still rejects the adapter unless `MARMOT_HOME` or `MARMOT_AGENT_SOCKET` exists in the **gateway process environment**.

## Reproduction

1. Start with a working systemd-managed Hermes gateway and no Marmot environment variables in its service environment.
2. Run the documented versioned equivalent of:

   ```sh
   curl -fsSL "https://github.com/marmot-protocol/mdk/releases/download/wn-agent-latest/install-hermes-marmot.sh" | bash
   ```

3. Restart the existing Hermes gateway as instructed.
4. Inspect gateway startup.

The installer:

- writes `platforms.marmot.enabled: true` and the configured `extra.home` / `extra.socket_path`;
- enables the plugin;
- starts a healthy `wn-agent` service;
- does **not** persist `MARMOT_HOME` or `MARMOT_AGENT_SOCKET` into the existing gateway service environment.

Gateway startup then logs:

```text
Platform 'marmot' registered but requirements not met -- skipping
Marmot adapter creation failed: Platform 'marmot' requirements not met
No adapter available for marmot
```

## Root cause

`integrations/hermes/marmot/adapter.py` currently defines:

```python
def check_requirements() -> bool:
    return validate_config(_MinimalConfig(extra={}))
```

That discards the actual `platforms.marmot.extra` configuration and therefore only succeeds when process environment variables are present. `configure_gateway.py` writes the needed home/socket into platform config, but the registry requirements check runs without that config and rejects the adapter first.

The installer only prints gateway restart guidance. Environment variables local to the piped installer process cannot propagate into an existing gateway service.

## Expected behavior

A clean run of the documented installer followed by the documented gateway restart must load Marmot using the configuration the installer persisted. Operators should not need an undocumented systemd drop-in or shell export.

## Acceptance criteria

- [ ] A configured `platforms.marmot.extra.home` or `socket_path` can satisfy adapter creation without requiring duplicate gateway-process environment variables.
- [ ] The plugin requirements hook checks install-time dependencies only, or otherwise receives/uses the effective platform config instead of `_MinimalConfig(extra={})`.
- [ ] Environment-variable-only deployments remain supported.
- [ ] The versioned curl installer followed by restart loads the adapter under a systemd-managed Hermes gateway.
- [ ] Add a regression test for: no `MARMOT_*` environment, valid enabled Marmot platform config, healthy local socket => adapter is created.
- [ ] Add an installer/integration assertion that the persisted config and runtime readiness contract agree.
- [ ] Existing Hermes platforms such as Telegram remain untouched.
- [ ] Update restart guidance so it names the correct command for an already-installed gateway rather than presenting `hermes gateway run` as a restart.

## Observed version

- Release: `wn-agent-v0.9.7`
- Installer SHA-256: `93be8426d01387ff8fc651df8b1dcc0cc460f2fa050cb368c56152f0ee6f154f`
- Also reproducible against current `master` readiness code at the time of filing.

agent-p1p:
Fixes #1127

## Summary
- split dependency readiness from effective configuration validation and register Hermes’ `is_connected` gate so unconfigured Marmot stays disabled
- prove unconfigured/no-env, installer-persisted, and environment-only enablement paths through adapter creation
- preserve environment-only deployments and correct the managed-gateway restart guidance

## Verification
- Marmot Python unittest suite: 111 passed
- Hermes dev/installer script test: passed
- `just fast-ci`: passed; convergence pin 5/5

## Sensitive paths
None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Marmot gateway readiness when only the configured agent socket is available.
  * Persisted gateway configuration can now create the Marmot adapter without requiring additional environment settings.

* **Documentation**
  * Updated installation guidance to clarify that existing Hermes gateways must be restarted manually.
  * Installer output now recommends `hermes gateway restart` to apply Marmot changes.

### DIFF
--- a/integrations/hermes/marmot/tests/test_configure_gateway.py
+++ b/integrations/hermes/marmot/tests/test_configure_gateway.py
@@ -1,4 +1,6 @@
 import importlib.util
+import os
+import socket
 import tempfile
 import unittest
 from pathlib import Path
@@ -17,6 +19,30 @@ def load_module():
     assert spec.loader is not None
     spec.loader.exec_module(module)
     return module
+
+
+def load_adapter_module():
+    helper_path = REPO_ROOT / "integrations" / "hermes" / "marmot" / "tests" / "test_adapter.py"
+    spec = importlib.util.spec_from_file_location("hermes_marmot_test_adapter_helpers", helper_path)
+    helper = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(helper)
+    return helper.load_adapter_module()
+
+
+class PlatformRegistryHarness:
+    """Minimal copy of Hermes' check, validate, then create registry flow."""
+
+    def register_platform(self, **entry):
+        self.entry = entry
+
+    def create_adapter(self, config):
+        if not self.entry["check_fn"]():
+            return None
+        validate = self.entry.get("validate_config")
+        if validate is not None and not validate(config):
+            return None
+        return self.entry["adapter_factory"](config)
 
 
 class ConfigureGatewayTests(unittest.TestCase):
@@ -155,6 +181,54 @@ class ConfigureGatewayTests(unittest.TestCase):
         self.assertTrue(config["streaming"]["enabled"])
         self.assertEqual(config["streaming"]["transport"], "auto")
         self.assertTrue(config["display"]["platforms"]["marmot"]["streaming"])
+
+    def test_persisted_config_creates_adapter_without_marmot_env(self):
+        adapter_module = load_adapter_module()
+        saved_env = {key: os.environ.pop(key) for key in list(os.environ) if key.startswith("MARMOT_")}
+        try:
+            with tempfile.TemporaryDirectory() as tempdir:
+                home = Path(tempdir)
+                agent_home = home / "marmot-agent"
+                socket_dir = agent_home / "dev"
+                socket_dir.mkdir(parents=True)
+                socket_path = socket_dir / "wn-agent.sock"
+
+                with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as server:
+                    server.bind(str(socket_path))
+                    server.listen(1)
+                    self.assertTrue(socket_path.is_socket())
+
+                    self.module.configure_gateway_config(
+                        hermes_home=home,
+                        platform="marmot",
+                        streaming_enabled=False,
+                        streaming_transport="off",
+                        tool_progress="off",
+                        interim_assistant_messages=False,
+                        long_running_notifications=False,
+                        busy_ack_detail=False,
+                        agent_home=agent_home,
+                        socket_path=socket_path,
+                        account_id_hex="11" * 32,
+                    )
+                    config = self.module.load_config(home / "config.yaml")
+                    extra = config["platforms"]["marmot"]["extra"]
+                    platform_config = type(
+                        "PlatformConfig",
+                        (),
+                        {"enabled": True, "extra": extra},
+                    )()
+                    registry = PlatformRegistryHarness()
+                    adapter_module.register(registry)
+
+                    adapter = registry.create_adapter(platform_config)
+
+                    self.assertTrue(config["platforms"]["marmot"]["enabled"])
+                    self.assertIsInstance(adapter, adapter_module.MarmotPlatformAdapter)
+                    assert adapter is not None
+                    self.assertEqual(adapter.socket_path, str(socket_path))
+        finally:
+            os.environ.update(saved_env)
 
 
 if __name__ == "__main__":
--- a/integrations/hermes/marmot/tests/test_configure_gateway.py
+++ b/integrations/hermes/marmot/tests/test_configure_gateway.py
@@ -3,6 +3,7 @@ import os
 import socket
 import tempfile
 import unittest
+import unittest.mock
 from pathlib import Path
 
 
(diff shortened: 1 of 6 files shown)


agent-p1p:

**Verdict: blocking (1)**

### Blocking

- **`integrations/hermes/marmot/adapter.py:2906`, `integrations/hermes/marmot/adapter.py:3023` — the install-time fix now silently opts unconfigured Marmot plugins into the platform.** `check_requirements()` unconditionally returns `True`, but `register()` does not supply Hermes' `is_connected` hook. Current Hermes deliberately treats a missing `is_connected` as a legacy plugin and falls through to `check_fn()` during `_apply_env_overrides`; a true result creates/enables a `PlatformConfig` even when the user has no `platforms.marmot` config and no `MARMOT_*` environment. See the current upstream enablement gate at [gateway/config.py:2510-2639](https://github.com/NousResearch/hermes-agent/blob/588b7059a8b57b0e3dea98b480048eb7199ce0b6/gateway/config.py#L2510-L2639). On this PR head, a registration probe produced `check_fn=True`, `is_connected_registered=False`, `validate_config_empty=False`, and therefore `current_hermes_auto_enable_fallback=True`.

  This regresses explicit platform opt-in for anyone who has the plugin enabled but Marmot unconfigured. At minimum the gateway now enables then rejects the platform on config validation; if the default `~/.marmot/dev/wn-agent.sock` exists, it can instantiate against that socket without an explicit Marmot platform configuration. Keep `check_requirements()` dependency-only, but register the existing configuration probe as `is_connected=validate_config` (the standard Hermes split between dependency availability and user configuration). Add a regression that exercises the enablement gate and proves an unconfigured plugin stays disabled while persisted-config and environment-only cases still enable/create the adapter. The current `PlatformRegistryHarness` covers only `create_adapter()` and therefore misses the configuration-loading regression.

### Verification

Reviewed exact head `f89e3a6566ee76464ddd46f0d5d7f05aecbe1b80`.

- `python3 -m unittest discover -s integrations/hermes/marmot/tests` with `MARMOT_*` readiness/auth variables unset: **109 passed**
- `just hermes-dev-script-test`: **passed**
- `just fast-ci`: **passed**, including convergence policy pin **5/5**
- `git diff --check`, shell syntax, and Python compile checks: **passed**
- GitHub CI and CodeRabbit: **green** at review time

No sensitive paths touched. No out-of-scope follow-up issues filed.

agent-p1p:
Addressed the blocking enablement finding in 52254cbc91ae65bdb8dcb6c153012105eede2651.

- registered `is_connected=validate_config`, preserving dependency-only `check_requirements()` without invoking Hermes’ legacy auto-enable fallback;
- added a red-capable enablement-gate harness covering unconfigured/no-env, installer-persisted config, and environment-only config through adapter creation;
- kept the pre-existing default-socket readiness fallback unchanged.

Verification: 8 focused configuration tests, 111 Marmot plugin tests, `just hermes-dev-script-test`, `just fast-ci`, `git diff --check`, Python compile, and an independent fresh-context review all pass.

agent-p1p:
@coderabbitai review

integrations/hermes/marmot/tests/test_configure_gateway.py
```diff
@@ -19,6 +22,74 @@ def load_module():
...
+    def apply_env_overrides(self, platform_config=None):
+        """Mirror the relevant gateway.config._apply_env_overrides plugin gate."""
+        existing = platform_config
+        env_fn = self.entry.get("env_enablement_fn")
+        seed = env_fn() if env_fn else None
+        is_connected = self.entry.get("is_connected")
+        if (existing is None or not existing.enabled) and is_connected is not None:
+            probe_extra = dict(getattr(existing, "extra", {}) or {})
+            if isinstance(seed, dict):
+                probe_extra.update({key: value for key, value in seed.items() if key != "home_channel"})
```
erskingardner, line 58: Nit, non-blocking: this probe-seeding line uses `.update(...)`, so env-seeded values win over already-persisted `extra` when both are present. The real Hermes gate this harness mirrors (`gateway/config.py`, `_apply_env_overrides`, ~line 2597) uses `probe_extra.setdefault(k, v)` for the *probe* construction specifically so persisted config takes precedence over env for the `is_connected" check — persisted-config-wins is exactly the property #1127 cares about. It doesn't affect any assertion in this PR (no test exercises overlapping persisted+env values), but since this harness is the only thing standing in for real Hermes in this repo's automation, worth matching the real precedence so it doesn't quietly drift from upstream on a future edit.

erskingardner (approved):
## Review

Verified the fix against real upstream Hermes source (`NousResearch/hermes-agent`, current `master` @ `c7b75a7`), not just this repo's mock: `check_fn` is genuinely a dependency-availability check (`gateway/platform_registry.py` docstring), and `is_connected` is the credential/config gate consulted in `_apply_env_overrides` (`gateway/config.py:2510-2660`) — unregistered `is_connected` falls straight through to `check_fn`, which is exactly the silent-auto-enable regression the author's own adversarial self-review caught and fixed in the second commit (`is_connected=validate_config`). The `PlatformRegistryHarness` test double matches that upstream gate logic closely.

I also reproduced the fix locally: swapping in the pre-PR `adapter.py` under the new tests makes `test_persisted_config_creates_adapter_without_marmot_env` and `test_register_wires_is_connected_for_env_override_gate` fail as expected (confirms these aren't tautological), and all 111 tests pass on the PR head.

Scope is surgical — only the Marmot Hermes plugin, its tests, and installer/README wording changed; other Hermes platforms untouched, matching the issue's acceptance criteria.

Two non-blocking notes:
- Inline: the enablement-probe harness uses `.update()` where the real Hermes gate uses `setdefault()`, so persisted config vs. env-var precedence isn't faithfully mirrored in one spot (doesn't affect current assertions).
- Neither `test_configure_gateway.py`/`test_dev_scripts.sh` nor a real "installer + restart with zero `MARMOT_*` env vars" scenario run in this repo's automated CI (`just fast-ci`/`just ci` are cargo-only per `Justfile:301-303`, and `hermes_marmot_dev_setup.sh`'s generated `env.sh` always exports `MARMOT_HOME`/`MARMOT_AGENT_SOCKET`). So the exact scenario from #1127 is proven here only via a hand-maintained mock of Hermes internals, not a live Hermes process. Pre-existing repo gap, not introduced by this PR — worth a follow-up to wire a real no-env e2e case into the dev harness.

No blocking issues. Recommend merge.

integrations/hermes/marmot/tests/test_configure_gateway.py
agent-p1p: Addressed in signed commit f08a04f665f233a5699f66c2ccd5d5a40589dce9 on `pip/mdk-1127`.

- the probe now mirrors real Hermes with `setdefault`, so persisted `extra.socket_path` wins during `is_connected`;
- a red-capable overlapping persisted/environment regression asserts that exact precedence;
- 112 Marmot plugin tests, the dev-script test, Ruff, Python compile, `git diff --check`, and `just fast-ci` pass.

Important: PR #1130 was merged at 08:17:17Z while this follow-up task was running. GitHub deleted the head branch; the push recreated `pip/mdk-1127`, so f08a04f is not part of the merged PR or `master`. I did not create a replacement PR, per the task constraint.

agent-p1p:
Implemented the requested real-Hermes follow-up in signed commit f08a04f665f233a5699f66c2ccd5d5a40589dce9 on `pip/mdk-1127`.

The dev harness now configures/enables Marmot, launches a fresh process under `env -i` with no inherited `MARMOT_*`, loads persisted config through pinned Hermes c7b75a7849cb260e6f17a045473bfdd0ea21ca81, verifies a live Unix socket, and constructs the adapter through the real platform registry. The path is wired into the generated helper, Justfile, README, and path-scoped WN Agent CI. It deliberately stops at Hermes config/plugin discovery/adapter construction rather than starting the long-lived gateway or dialing a model provider.

Verification: real Hermes persisted-config probe passed; 112 plugin tests passed; dev-script test passed; Ruff/Python/shell/YAML/diff checks passed; `just fast-ci` passed (convergence pin 5/5).

Merge race: PR #1130 was merged at 08:17:17Z before this commit was pushed. GitHub had deleted the branch; the push recreated `pip/mdk-1127`. Therefore f08a04f is not in merged PR #1130 or `master`. No replacement PR was created, per the task constraint.

Issue #1026, opened 2026-07-21
Security review L32: is_nostr_secret detects only the nsec bech32 prefix — hex secret keys on argv are not caught


erskingardner:
## Summary

`wn login nsec1…` is hard-rejected (shell-history protection), but a 64-char hex *secret* key is indistinguishable from a hex pubkey and would be accepted as a public identity while sitting in shell history. Likely unavoidable at the format layer; the `--nsec-stdin` path and TUI stdin piping plus composer redaction are correct.

## Location

`crates/cli/src/lib.rs:862-864` (`is_nostr_secret`), `:287-296` (argv rejection); safe paths: `--nsec-stdin` (`:274-285`), TUI (`tui/model.rs:18-31, 1282-1291`)

## Impact

A user pasting a hex nsec onto a command line leaks it into shell history and is silently treated as a pubkey — footgun rather than attacker vector.

## Suggested remediation

Document the ambiguity; optionally warn when a 64-char hex argument is used where a pubkey is expected ("if this is a secret key, use --nsec-stdin").

---
Found in the 2026-07-20 security review (`docs/security-review-2026-07-20.md`, finding L32). Status: **confirmed.**

erskingardner:
Closing: `is_nostr_secret` is an intentional conservative `nsec` prefix check, documented as not full bech32/hex validation. Hex secrets on argv are a residual argv-hygiene gap, not a current leak.

PR #1131, opened 2026-07-26
feat(installers): securely import existing agent identities

### LINKED ISSUE #822: marmot-account: local-signing import writes the keychain secret before the account record — a crash in between orphans a keychain entry that blocks re-import with no API recovery
## Summary

`write_signing_account_for_label` writes the external (keychain) secret first, then the durable local account record, with compensation that runs **only on an `Err` return** — not on a process crash between the two writes. A crash in that window leaves a keychain credential with no `account.json`: the account never appears in `accounts()`/`account()` (so `remove_account` can't clear it), yet re-importing the same nsec is rejected as `AccountIdInUse` because the orphaned keychain entry still exists.

## Location

`crates/marmot-account/src/home.rs:594`

```rust
self.secret_store.write_secret(&account, keys)?;         // 594: keychain entry keyed by account_id
if let Err(err) = self.write_account_record(&account) {  // 595: durable local record
    let _ = self.secret_store.remove_secret(&account);   // compensation on Err only
    return Err(err);
}
Ok(account)
```

The `AccountIdInUse` gate is driven by `has_secret_for_account_id` (`secret_store.rs:197`), which returns `Ok(true)` whenever the keychain entry exists — independent of whether a record exists.

## Failure scenario

Import a local signing key with a `KeychainSecretStore`. `write_secret` sets a credential keyed by `account_id_hex`; the process crashes (or is killed) before `write_account_record`. Now: `accounts()`/`account()` never surface the account (no record), so `remove_account` fails with `UnknownAccount` and cannot clear it; re-importing the same nsec hits `has_secret_for_account_id → Ok(true)` → `AccountIdInUse`. The user can neither use nor re-import the key via the API.

#776/#721/#811 concern local secret-file permissions and scrub, not create/import ordering. The `AccountHome` docs acknowledge only the concurrency advisory-check race, not crash-recovery orphaning of the keychain store. No close match.

## Suggested fix

Persist the account record **before** the external secret write (or reconcile an orphaned `has_secret_for_account_id` that has no matching record — treat it as recoverable rather than `AccountIdInUse`), so a crash leaves a visible, removable "record without secret" state instead of an invisible, re-import-blocking orphan.

---
*Filed from a meticulous automated code-review pass across all crates; independently verified against the source and deduplicated against existing open/closed issues.*

### LINKED ISSUE #1128: feat(installers): securely bootstrap Hermes and OpenClaw with a known existing Nostr identity
## Summary

Provide a simple, first-class, secure way for both the **Hermes** and **OpenClaw** Marmot installer/bootstrap flows to import and use a **known existing Nostr identity**.

Today both `install-hermes-marmot.sh` and `install-openclaw-marmot.sh` effectively offer two awkward choices:

1. let `wn-agent bootstrap` create a new identity; or
2. manually install/build a compatible `wn` CLI, normalize the secret representation, import it separately, verify the resulting account id, start `wn-agent`, and then run `bootstrap --no-create --account-id-hex ...`.

That makes preserving an agent's established Nostr/ngit identity unnecessarily difficult and encourages unsafe secret-handling workarounds.

## Desired operator flows

Both release installers should support equivalent repeatable flows:

```sh
curl -fsSL "<versioned Hermes installer URL>" | \
  bash -s -- \
    --yes \
    --existing-identity-file "$HOME/.config/example/hermes-agent.nsec" \
    --expected-npub npub1... \
    --allow-welcomer npub1...
```

```sh
curl -fsSL "<versioned OpenClaw installer URL>" | \
  bash -s -- \
    --yes \
    --existing-identity-file "$HOME/.config/example/openclaw-agent.nsec" \
    --expected-npub npub1... \
    --allow-welcomer npub1...
```

The exact interface can differ, but the **secret value must never be placed in argv or an environment variable**. A file path, inherited file descriptor, or masked `/dev/tty` prompt is acceptable. Plain `--nsec <secret>` is not.

## Shared implementation requirement

The secure identity-import/bootstrap behavior should be implemented once in shared `wn-agent`/account tooling or a shared installer helper and consumed consistently by both installers. Do not create divergent Hermes and OpenClaw key-import implementations.

The integrations should retain their existing connector-specific defaults:

- Hermes: `~/.marmot-agents/hermes`, `wn-agent-hermes.service`, `hermes-agent`
- OpenClaw: `~/.marmot-agents/openclaw`, `wn-agent-openclaw.service`, `openclaw-agent`

Supplying the same known identity to both independent connector homes must be an explicit operator choice. Shared-home/shared-socket deployments remain opt-in and must not be introduced silently.

## Required behavior

- Import an existing local-signing Nostr identity before bootstrap.
- Accept the supported private-key representations already handled by the account tooling (at minimum `nsec`; raw hex may be normalized if deliberately supported).
- Optionally require an expected public identity (`npub` or hex) and fail closed if the imported key derives to anything else.
- Run bootstrap in no-create mode against the exact imported account.
- Report `created: false` for this path.
- Publish the KeyPackage only after identity verification succeeds.
- Continue to support generated-identity onboarding as an explicit/default path, but make the selected behavior unambiguous in both interactive and noninteractive modes.
- Reuse existing account-import primitives rather than adding new cryptographic key parsing/derivation logic to either shell installer.
- Preserve each host runtime's unrelated configuration; only the relevant Hermes platform or OpenClaw channel may be changed.

## Secret-handling requirements

This is security-sensitive key-handling work and requires human review before merge.

- [ ] Never put the secret value in command arguments, environment variables, process listings, logs, tracing, error text, shell history, `bootstrap.json`, Hermes config, or OpenClaw config.
- [ ] Do not echo the secret or copy it into world/group-readable temporary files.
- [ ] Validate identity-file ownership/type/permissions and reject symlinks or unsafe files.
- [ ] Preserve the caller's source credential file; do not rewrite or delete it.
- [ ] Ensure transient buffers/files follow existing zeroization and restrictive-file invariants.
- [ ] Keep crash recovery and duplicate-import behavior well-defined; account record and secret persistence must not introduce or worsen the orphaning class tracked by #822.
- [ ] Keep secret detection/redaction aligned with #1026.

## Acceptance criteria

- [ ] One documented `install-hermes-marmot.sh` invocation can install the connector, import a supplied known identity securely, verify its expected public identity, bootstrap with `--no-create`, configure Hermes, and print restart instructions.
- [ ] One documented `install-openclaw-marmot.sh` invocation can perform the equivalent flow and configure OpenClaw.
- [ ] Both installers consume the same hardened identity-import primitive and expose consistent option names/semantics.
- [ ] Interactive setup offers a masked existing-identity path without competing with the installer's stdin when the script itself is supplied by `curl | bash`.
- [ ] Noninteractive setup accepts a secret **source** (path/FD), never a secret value in argv/env.
- [ ] The resulting account id exactly matches the supplied expected identity.
- [ ] A mismatch fails before KeyPackage publication or host-runtime configuration.
- [ ] Existing matching accounts are reused idempotently in either connector home.
- [ ] Tests cover both installers: new import, existing import, identity mismatch, duplicate import, unsafe file permissions/symlink, malformed input, interrupted import, and absence of secret material from output/config/bootstrap artifacts.
- [ ] Coexistence tests preserve the connector-specific homes, sockets, services, labels, and identities unless sharing is explicitly requested.
- [ ] Hermes and OpenClaw documentation each include a pinned-release example and explain generated identity vs existing identity behavior.

## Scope distinction

This is separate from the Hermes gateway readiness/install regression tracked alongside #838. That issue concerns making an already-configured connector load in Hermes. This issue concerns securely selecting and importing the correct Nostr identity during onboarding for both Hermes and OpenClaw.

agent-p1p:
## Summary

- add `wn-agent import-identity` with secure owner-only file ingestion and masked `/dev/tty` input
- verify an optional expected npub/hex before persistence, then idempotently create, reuse, reactivate, or repair the exact local-signing account
- wire identical `--existing-identity-file`, `--expected-npub`, and `--generate-identity` behavior into the Hermes and OpenClaw installers
- bootstrap imported identities with `--no-create --account-id-hex`, preserving connector-specific homes, sockets, services, and labels
- document pinned-release flows and add account, connector, installer, and secret-leak regressions

## Security

This PR touches sensitive identity/key-handling paths and requires human review before merge.

- secret values never enter argv, environment variables, logs, installer output, host config, or `bootstrap.json`
- file imports reject symlinks, non-regular files, wrong ownership, group/other access, and multiple hard links
- expected-public-key mismatch fails before account persistence, service startup, KeyPackage publication, or host configuration
- transient input buffers zeroize on drop; masked prompts restore terminal state across interrupts and job-control signals
- interrupted/duplicate account imports recover without creating a second signing identity or invisible keychain orphan

## Verification

- `just fast-ci`
- `cargo clippy -p agent-connector -p marmot-account --all-targets --locked -- -D warnings`
- `integrations/test_installer_systemd_service.sh hermes`
- `integrations/test_installer_systemd_service.sh openclaw`
- real PTY masked-import success and SIGINT restoration checks

`just openclaw-dev-script-test` passed after merging current `master`; the OpenClaw installer-specific systemd/import test also passed.

Fixes #1128
Fixes #822

## Summary by CodeRabbit

* **New Features**
  * Added secure import of existing Nostr identities from protected files or interactive prompts.
  * Added identity verification using an expected npub or hexadecimal identity.
  * Installers now support preserving an existing identity and reusing it during bootstrap.
  * Added options for existing-identity imports and explicit generated-identity onboarding.

* **Bug Fixes**
  * Improved recovery and idempotent reuse of interrupted or previously signed-out accounts.
  * Prevented identity secrets from appearing in command output, logs, or persisted account records.

* **Documentation**
  * Updated Hermes and OpenClaw installation guidance for existing-identity onboarding.

### DIFF
--- a/crates/agent-connector/src/bootstrap.rs
+++ b/crates/agent-connector/src/bootstrap.rs
@@ -726,7 +726,10 @@ mod tests {
             )
             .await;
 
-        let result = run_bootstrap(test_options(socket_path)).await.unwrap();
+        let mut options = test_options(socket_path);
+        options.account_id_hex = Some(ACCOUNT_ID.to_owned());
+        options.create_if_missing = false;
+        let result = run_bootstrap(options).await.unwrap();
         server.abort();
 
         assert!(!result.created);
--- a/crates/agent-connector/src/identity.rs
+++ b/crates/agent-connector/src/identity.rs
@@ -1,0 +1,270 @@
+//! Secure import of an existing local-signing Nostr identity.
+
+use std::fs::{File, OpenOptions};
+use std::io::{self, Read};
+use std::path::Path;
+
+use marmot_account::{AccountHome, AccountHomeError};
+use serde::Serialize;
+use thiserror::Error;
+use zeroize::Zeroizing;
+
+const MAX_IDENTITY_BYTES: u64 = 4096;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
+pub struct ExistingIdentityImport {
+    pub account_id_hex: String,
+    pub label: String,
+    pub local_signing: bool,
+}
+
+#[derive(Debug, Error)]
+pub enum ExistingIdentityError {
+    #[error("could not open identity file ({0:?})")]
+    Open(io::ErrorKind),
+    #[error("could not read identity file ({0:?})")]
+    Read(io::ErrorKind),
+    #[error("identity source must be a regular file")]
+    NotRegularFile,
+    #[error("identity file must be owned by the current user")]
+    WrongOwner,
+    #[error("identity file must not be accessible by group or other users")]
+    UnsafePermissions,
+    #[error("identity file must have exactly one hard link")]
+    MultipleHardLinks,
+    #[error("identity input is empty")]
+    Empty,
+    #[error("identity input exceeds {MAX_IDENTITY_BYTES} bytes")]
+    TooLarge,
+    #[error("identity input is not valid UTF-8")]
+    InvalidUtf8,
+    #[error("expected identity must be an npub or 64-character hex public key")]
+    InvalidExpectedIdentity,
+    #[error("imported identity does not match the expected public identity")]
+    ExpectedIdentityMismatch,
+    #[error(transparent)]
+    Account(#[from] AccountHomeError),
+}
+
+pub fn import_existing_identity_file(
+    home: &Path,
+    label: &str,
+    identity_file: &Path,
+    expected_identity: Option<&str>,
+) -> Result<ExistingIdentityImport, ExistingIdentityError> {
+    let file = open_private_identity_file(identity_file)?;
+    let secret = read_identity(file)?;
+    import_existing_identity_secret(home, label, secret.as_str(), expected_identity)
+}
+
+pub fn import_existing_identity_secret(
+    home: &Path,
+    label: &str,
+    secret: &str,
+    expected_identity: Option<&str>,
+) -> Result<ExistingIdentityImport, ExistingIdentityError> {
+    let secret = secret.trim();
+    if secret.is_empty() {
+        return Err(ExistingIdentityError::Empty);
+    }
+
+    let account_id_hex = AccountHome::account_id_for_secret(secret)?;
+    if let Some(expected_identity) = expected_identity {
+        let expected_account_id = marmot_app::account_id_hex_from_ref(expected_identity)
+            .map_err(|_| ExistingIdentityError::InvalidExpectedIdentity)?;
+        if account_id_hex != expected_account_id {
+            return Err(ExistingIdentityError::ExpectedIdentityMismatch);
+        }
+    }
+
+    let account = AccountHome::open(home).import_account_idempotent(label, secret)?;
+    Ok(ExistingIdentityImport {
+        account_id_hex: account.account_id_hex,
+        label: account.label,
+        local_signing: account.local_signing,
+    })
+}
+
+fn open_private_identity_file(path: &Path) -> Result<File, ExistingIdentityError> {
+    let mut options = OpenOptions::new();
+    options.read(true);
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::OpenOptionsExt;
+        options.custom_flags(libc::O_CLOEXEC | libc::O_NOFOLLOW | libc::O_NONBLOCK);
+    }
+    let file = options
+        .open(path)
+        .map_err(|error| ExistingIdentityError::Open(error.kind()))?;
+    let metadata = file
+        .metadata()
+        .map_err(|error| ExistingIdentityError::Read(error.kind()))?;
+    if !metadata.file_type().is_file() {
+        return Err(ExistingIdentityError::NotRegularFile);
+    }
+    #[cfg(unix)]
+    validate_unix_identity_metadata(&metadata)?;
+    if metadata.len() > MAX_IDENTITY_BYTES {
+        return Err(ExistingIdentityError::TooLarge);
+    }
+    Ok(file)
+}
+
+#[cfg(unix)]
+fn validate_unix_identity_metadata(
+    metadata: &std::fs::Metadata,
+) -> Result<(), ExistingIdentityError> {
+    use std::os::unix::fs::MetadataExt;
+
+    if metadata.uid() != unsafe { libc::geteuid() } {
+        return Err(ExistingIdentityError::WrongOwner);
+    }
+    if metadata.mode() & 0o077 != 0 {
+        return Err(ExistingIdentityError::UnsafePermissions);
+    }
+    if metadata.nlink() != 1 {
+        return Err(ExistingIdentityError::MultipleHardLinks);
+    }
+    Ok(())
+}
+
+fn read_identity(reader: impl Read) -> Result<Zeroizing<String>, ExistingIdentityError> {
+    let mut bytes = Zeroizing::new(Vec::with_capacity(128));
+    reader
+        .take(MAX_IDENTITY_BYTES + 1)
+        .read_to_end(&mut bytes)
+        .map_err(|error| ExistingIdentityError::Read(error.kind()))?;
+    if bytes.len() as u64 > MAX_IDENTITY_BYTES {
+        return Err(ExistingIdentityError::TooLarge);
+    }
+    std::str::from_utf8(bytes.as_slice()).map_err(|_| ExistingIdentityError::InvalidUtf8)?;
+    let bytes = std::mem::take(&mut *bytes);
+    // SAFETY: UTF-8 validity was checked immediately above, before ownership
+    // moved out of the zeroizing byte buffer.
+    let secret = unsafe { String::from_utf8_unchecked(bytes) };
+    let secret = Zeroizing::new(secret);
+    if secret.trim().is_empty() {
+        return Err(ExistingIdentityError::Empty);
+    }
+    Ok(secret)
+}
+
+#[cfg(all(test, unix))]
+mod tests {
+    use std::os::unix::fs::{PermissionsExt, symlink};
+
+    use super::*;
+
+    const NSEC: &str = "nsec1j4c6269y9w0q2er2xjw8sv2ehyrtfxq3jwgdlxj6qfn8z4gjsq5qfvfk99";
+    const NPUB: &str = "npub14f8usejl26twx0dhuxjh9cas7keav9vr0v8nvtwtrjqx3vycc76qqh9nsy";
+
+    fn write_identity(path: &Path, value: &str) {
+        std::fs::write(path, value).unwrap();
+        std::fs::set_permissions(path, std::fs::Permissions::from_mode(0o600)).unwrap();
+    }
+
+    #[test]
+    fn imports_and_idempotently_reuses_a_private_identity_file() {
+        let dir = tempfile::tempdir().unwrap();
+        let home = dir.path().join("home");
+        let identity = dir.path().join("identity.nsec");
+        write_identity(&identity, NSEC);
+
+        let first = import_existing_identity_file(&home, "agent", &identity, None).unwrap();
+        let second = import_existing_identity_file(&home, "agent", &identity, None).unwrap();
+
+        assert_eq!(second, first);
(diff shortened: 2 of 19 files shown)


agent-p1p:
Adversarial review at exact head `6b21436e06bf0bff3d1cb3612929aa66cec43674`.

BLOCKING:
none

SUGGESTION:
- Add durable regression coverage for the two security branches currently verified only by inspection/manual execution: reject a multiply-linked source file (`crates/agent-connector/src/identity.rs:125`) and restore `/dev/tty` echo/handlers after SIGINT or SIGTSTP (`crates/agent-connector/src/bin/wn-agent.rs:314-456`). The implementation passed both checks in a real local run, but the original job-control finding makes an automated Unix PTY guard worthwhile.

NITPICK:
- `MAX_IDENTITY_BYTES` is duplicated, and the TTY branch rejects the next byte when the buffer already has 4096 bytes while the file branch accepts exactly 4096 (`crates/agent-connector/src/bin/wn-agent.rs:307,439`; `crates/agent-connector/src/identity.rs:12,107,137`). This cannot affect valid nsec/raw-hex keys, but sharing the limit would avoid drift.

SUMMARY: No functional or security blocker found. The secret reaches Rust only through an opened file descriptor or masked `/dev/tty`; it does not enter argv, env, errors, command output, host config, or bootstrap JSON. File validation opens with `O_NOFOLLOW`, then checks the opened descriptor's type, owner, mode, link count, and bounded length before reading. Expected-identity comparison precedes account persistence and every installer service/config side effect. The account path reuses or repairs exact local-signing identities, covers record-first interruption and legacy keychain-orphan recovery, and the installers bootstrap the returned account id with `--no-create`, yielding `created=false` while retaining connector-specific homes, sockets, services, labels, and config scopes. The revised TTY path blocks SIGTTIN/SIGTTOU through echo restoration and restores echo plus prior handlers before re-raising other terminal signals.

Verification:
- `just fast-ci` passed.
- `cargo test -p marmot-account` passed (80 tests).
- `cargo test -p agent-connector` passed (124 tests).
- `just hermes-dev-script-test` passed.
- `integrations/test_installer_systemd_service.sh hermes` passed.
- `integrations/test_installer_systemd_service.sh openclaw` passed.
- Real `wn-agent import-identity` checks passed for idempotent reuse, mismatch-before-persistence, mode/symlink/hardlink/FIFO/malformed rejection, source preservation, and secret-redacted output.
- Real PTY SIGINT check confirmed echo disabled during input and restored before signal termination.
- `just openclaw-dev-script-test` could not run locally because this host has no Node/pnpm; the GitHub OpenClaw package job is green.
- All GitHub CI jobs are green. CodeRabbit was rate-limited and did not produce a review.

VERDICT: APPROVE

SENSITIVE_PATHS: `crates/agent-connector/src/identity.rs`, `crates/agent-connector/src/bin/wn-agent.rs`, `crates/marmot-account/src/home.rs`, `crates/agent-connector/src/bootstrap.rs`, `scripts/install-hermes-marmot.sh`, `scripts/install-openclaw-marmot.sh`

Sensitive key-handling remains human-merge-only. I did not merge.

agent-p1p:
Addressed the adversarial coverage suggestion in `ea7f1a1`: added hard-link rejection coverage and a real Unix PTY regression test proving masked prompt echo is restored and the prior handler is reinstated before SIGINT is re-raised. Independent rereview passed after correcting Apple `openpty` pointer portability. Focused ad-hoc format, all-target clippy, hard-link, and PTY tests pass.

agent-p1p:
@coderabbitai review

crates/agent-connector/src/identity.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+    Ok(())
+}
+
+fn read_identity(reader: impl Read) -> Result<Zeroizing<String>, ExistingIdentityError> {
+    let mut bytes = Zeroizing::new(Vec::with_capacity(128));
+    reader
+        .take(MAX_IDENTITY_BYTES + 1)
+        .read_to_end(&mut bytes)
+        .map_err(|error| ExistingIdentityError::Read(error.kind()))?;
```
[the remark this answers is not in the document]
agent-p1p: Fixed in `e13aae1`. Both file and masked-TTY paths now preallocate the complete bounded input buffer before reading, and the regression test fills the full `MAX+1` read and asserts capacity never changes. The size limit is shared across both paths.

agent-p1p:
CodeRabbit follow-up: `e13aae1` also addresses the shared `MAX_IDENTITY_BYTES` nit. I kept the validated `from_utf8` + `from_utf8_unchecked` conversion: on invalid UTF-8 the bytes remain inside `Zeroizing<Vec<u8>>` and are wiped, while the proposed direct `String::from_utf8` needs extra `FromUtf8Error::into_bytes()` handling to retain that property. Changelog PR references were not added because this changelog does not use per-entry PR links.

agent-p1p:
@coderabbitai review

crates/agent-connector/src/bootstrap.rs
```diff
@@ -726,7 +726,10 @@ mod tests {
             )
             .await;
 
-        let result = run_bootstrap(test_options(socket_path)).await.unwrap();
+        let mut options = test_options(socket_path);
+        options.account_id_hex = Some(ACCOUNT_ID.to_owned());
+        options.create_if_missing = false;
```
erskingardner, line 732: This existing test (`bootstrap_reuses_existing_account_and_repairs_key_package`) was the only coverage in this file of `select_account`'s by-label-match branch (`account_id_hex: None`, `create_if_missing: true`, matched via the account list's `label` field) — i.e. the default "re-run bootstrap on an already-created account" flow.

By forcing `options.account_id_hex = Some(ACCOUNT_ID)` and `options.create_if_missing = false` here, the test now exercises the *other* branch of `select_account` (the `account_id_hex`-based lookup used by the new `--no-create --account-id-hex` installer flow), and the by-label branch loses its only test in this file. `select_account`'s two branches are mutually exclusive (the account-id-hex arm returns before the label-matching code ever runs), so this isn't just a stronger version of the old test — it's a different code path entirely.

Given `run_bootstrap`/`select_account` are only unit-tested here (no other test in the repo calls `run_bootstrap`), could you either:
- add a new dedicated test for the `--no-create --account-id-hex` reuse case introduced by this PR, and leave this test covering the default by-label reuse, or
- add the by-label scenario back as a second test

so both `select_account` branches keep coverage?

erskingardner (dismissed):
Reviewed against #1128's acceptance criteria and the linked #822 (secret-before-record orphaning) and #1026 (npub/hex secret ambiguity) issues.

Overall this is a careful, well-scoped implementation: secrets never touch argv/env, the file-import path rejects symlinks/wrong-owner/group-readable/multi-hard-link files via `fstat` on the already-opened fd (no TOCTOU window), the TTY prompt correctly masks echo and defers job-control/interrupt signals with a tested PTY regression, buffers are pre-sized so `Zeroizing` vectors never reallocate before zeroization, and `import_account_idempotent`/`write_new_signing_account` fix the #822 crash-orphan class by writing the account record before the keychain/secret write (with compensation), which is exactly the fix that issue suggested. Both installers share the same `wn-agent import-identity` primitive and keep connector-specific homes isolated, matching the issue's shared-implementation requirement.

One real finding below (test-coverage regression), plus a couple of non-blocking notes:

- `--expected-npub`/`--expected-identity` accepts either an npub or 64-char hex, same ambiguity as #1026 (a pasted hex *secret* key would silently be treated as a hex pubkey reference and land in argv/shell history, defeating the point of this feature). It fails closed (mismatch) rather than misidentifying accounts, so it's not a vulnerability, but a one-line doc callout ("must be a public key, never a secret") next to `--expected-npub` in both READMEs would help given how much care went into keeping secrets out of argv elsewhere in this PR.
- Consider having the PR/commit also reference #822, since `write_new_signing_account`'s reordering is the actual fix for that bug, not just a dependency of this feature.

Nice-to-have, not requesting changes over these.

crates/agent-connector/src/bootstrap.rs
erskingardner: Addressed in 3ff45d02. The explicit `--no-create --account-id-hex` path now has its own test, while a separate test preserves the original label-match reuse coverage. Both assert the AccountList and KeyPackage/allowlist request sequence, and the full agent-connector suite passes.

Issue #513, opened 2026-06-19
[Feature]: Engine sign_out(pubkey, options) API — relay-side key-package cleanup on logout


Datawav:
**Sub-issue of marmot-protocol/darkmatter-android#347.**

Today's logout path in the engine drops in-memory caches, deactivates subscriptions, removes the SQL account row, and clears the secret store, but **leaves outstanding KeyPackages published on the user's KeyPackage relays**. While the user is signed out, strangers can still pull those KPs and gift-wrap Welcome events into groups the user is no longer monitoring.

## Proposed API

```
sign_out(pubkey: PublicKey, options: SignOutOptions) -> SignOutOutcome

struct SignOutOptions {
  delete_key_packages: bool,   // default true
}

struct SignOutOutcome {
  local_cleanup: LocalCleanupReport,           // already covered by today's logout
  key_packages_deleted: u32,
  key_package_failures: Vec<RelayFailure>,     // per-relay, structured
}
```

## Behavior

- Run all existing logout steps (background tasks, subscriptions, ephemeral relay unwarming, SQL account row delete, secret-store clear).
- If `delete_key_packages`: enumerate all cached KP event IDs for this account, publish kind:5 deletions to the KP relays, return per-relay success/failure.
- **Do NOT touch the MLS state DB** — that is the existing 'sign back in and resume' contract.
- Atomic-ish: if relay cleanup fails, local cleanup still completes; the FFI surfaces failures so the app can show 'will retry on next sign-in'.

## FFI

- Expose via uniffi as `signOut(accountRef, deleteKeyPackages: bool) -> SignOutOutcomeFfi`.
- Today's `removeAccount(accountRef)` becomes a thin wrapper around `signOut(_, false)` for backwards compatibility, OR is deprecated in the same release if no other caller exists. Pick one and document it.

## Out of scope

- The destructive wipe path (separate sub-issue).
- Cross-device key-package cleanup signaling.

## Notes

- App-side consumer is filed against `darkmatter-android` as the Sign Out wiring sub-issue of marmot-protocol/darkmatter#347.
- Per-event kind:5 deletion already exists at the `delete_account_key_package` API; this is the batched version.
- Implementation considerations: rate-limit kind:5 publishes per relay to avoid being thresholded; treat 404/'event not found' from a relay as success (the goal state is 'KP is not retrievable').

PR #1133, opened 2026-07-27
fix(agent): skip onboarding for existing profiles

### LINKED ISSUE #1126: Agent onboarding: skip kind-0 name prompt when a profile already exists
## Problem

Both the OpenClaw and Hermes Marmot integrations currently offer to publish a public Nostr profile name when an agent joins a group, even when that agent account already has valid published kind-0 metadata.

That is noisy for existing identities and creates a dangerous UX: the user is invited to publish "new" profile metadata when the account already has a profile that should be preserved.

Profile-name onboarding should only run for an account with no valid kind-0 metadata.

## Affected integrations

- `integrations/openclaw/marmot`
- `integrations/hermes/marmot`

Both adapters should share the same lookup semantics and tests.

## Desired behavior

Before sending the profile-name onboarding prompt:

1. Resolve the selected agent account's Nostr pubkey.
2. Look up the latest valid kind-0 event for that pubkey using the account/profile directory path.
3. If any valid, correctly signed kind-0 metadata exists, mark profile onboarding satisfied for that account and do not prompt.
4. Only offer to publish a name when no valid kind-0 metadata exists.

The check is account-scoped, not group-scoped. Joining additional groups must not repeat the lookup/prompt unnecessarily once the account is known to have a profile.

## Safety and failure behavior

- Never overwrite or republish existing kind-0 metadata merely because local onboarding state is missing or was reset.
- Preserve all existing kind-0 fields.
- Treat malformed, invalidly signed, or wrong-author events as nonexistent.
- If profile lookup fails because relays are unavailable, times out, or returns an indeterminate result, do **not** prompt immediately. Retry later with bounded backoff. An infrastructure failure must not be interpreted as "this account has no profile."
- Deduplicate concurrent group-join checks so two invites cannot produce two prompts.
- Persist the resulting account-level state (`profile_exists`, `prompted`, `published`, or equivalent) so restarts remain quiet.
- Keep explicit/manual profile publication available; this issue only changes the automatic onboarding prompt.

## Connector/API consideration

If the adapters cannot reliably determine profile existence through the current agent-control surface, add one shared read operation that returns a typed result such as:

```text
profile_found(metadata/event reference)
profile_not_found
indeterminate(error/retryability)
```

Do not collapse `indeterminate` into `profile_not_found`.

## Acceptance criteria

- [ ] OpenClaw skips the onboarding prompt when the selected account already has valid kind-0 metadata.
- [ ] Hermes skips the onboarding prompt under the same condition.
- [ ] Both integrations prompt exactly once when the account has no valid kind-0 metadata.
- [ ] Joining a second group does not prompt again.
- [ ] Resetting/deleting local onboarding state does not cause a prompt if kind-0 still exists on relays.
- [ ] A relay timeout or offline startup does not cause a false prompt; lookup is retried later.
- [ ] Existing metadata fields are never overwritten by the detection flow.
- [ ] Concurrent invites produce at most one lookup workflow and one prompt.
- [ ] Tests cover existing profile, absent profile, malformed/wrong-author profile, indeterminate lookup, restart, and concurrent group joins for both OpenClaw and Hermes.
- [ ] The OpenClaw and Hermes implementations use the same definition of profile existence.

## UX invariant

> Existing valid kind-0 metadata means the agent already has a public profile. Say nothing. Ask about publishing a name only when the profile truly does not exist.

agent-p1p:
## Summary

- add a typed account profile lookup over `marmot.agent-control.v2` that distinguishes found, absent, and indeterminate relay outcomes
- validate signature, author, kind, and JSON-object metadata before an event can satisfy profile onboarding
- persist account-level `profile_exists` state and share single-flight lookup/backoff semantics across OpenClaw and Hermes
- keep missing/failed lookups quiet while preserving the existing manual publication flow

## Verification

- `just fast-ci`
- `cargo test -p agent-connector` (120 passed)
- `cargo test -p wn-opencode` (48 passed, 1 connector E2E ignored)
- `cargo test -p marmot-app` (451 library tests and 71/76 relay-runtime tests passed; five unrelated relay-runtime failures remain on this host, with `encrypted_media_endpoint_updates_are_full_replacement_and_admin_only` reproduced from a clean `origin/master` archive)
- OpenClaw `pnpm typecheck` and full `pnpm test` (207 passed, 1 connector E2E skipped)
- targeted Hermes profile lookup/onboarding/restart/concurrency tests

Fixes #1126

### DIFF
--- a/integrations/hermes/marmot/adapter.py
+++ b/integrations/hermes/marmot/adapter.py
@@ -16,11 +16,12 @@ import os
 import re
 import shutil
 import stat
+import time
 import uuid
 from collections import OrderedDict
 from contextvars import ContextVar
 from pathlib import Path
-from typing import Any, AsyncIterator, Dict, Iterable, Literal, Optional, Tuple
+from typing import Any, AsyncIterator, Awaitable, Callable, Dict, Iterable, Literal, Optional, Tuple
 
 from gateway.config import Platform, PlatformConfig
 from gateway.platforms.base import (
@@ -723,6 +724,9 @@ class ProfileNameOnboardingStore:
     async def mark_skipped(self, account_id_hex: str) -> None:
         await self._set(account_id_hex, {"status": "skipped"})
 
+    async def mark_profile_exists(self, account_id_hex: str) -> None:
+        await self._set(account_id_hex, {"status": "profile_exists"})
+
     async def _set(self, account_id_hex: str, record: Dict[str, Any]) -> None:
         async with self._lock:
             data = self._read()
@@ -790,6 +794,72 @@ class AgentTextStreamTranscript:
             self.chunk_count += 1
 
 
+ProfileLookupStatus = Literal["profile_found", "profile_not_found", "indeterminate"]
+PROFILE_LOOKUP_BACKOFF_S = (1.0, 5.0, 30.0)
+
+
+class ProfileLookupGate:
+    """Deduplicate account profile lookups and back off indeterminate results."""
+
+    def __init__(
+        self,
+        *,
+        now: Callable[[], float] = time.monotonic,
+        backoff_seconds: Tuple[float, ...] = PROFILE_LOOKUP_BACKOFF_S,
+    ):
+        self._now = now
+        self._backoff_seconds = backoff_seconds or PROFILE_LOOKUP_BACKOFF_S
+        self._lock = asyncio.Lock()
+        self._in_flight: Dict[str, asyncio.Task[ProfileLookupStatus]] = {}
+        self._failures: Dict[str, Tuple[int, float]] = {}
+
+    async def lookup(
+        self,
+        account_id_hex: str,
+        lookup: Callable[[], Awaitable[Dict[str, Any]]],
+    ) -> ProfileLookupStatus:
+        key = account_id_hex.lower()
+        async with self._lock:
+            failure = self._failures.get(key)
+            if failure is not None and self._now() < failure[1]:
+                return "indeterminate"
+            task = self._in_flight.get(key)
+            if task is None:
+                task = asyncio.create_task(self._perform_lookup(key, lookup))
+                self._in_flight[key] = task
+                task.add_done_callback(
+                    lambda completed, lookup_key=key: self._discard_completed(lookup_key, completed)
+                )
+
+        return await asyncio.shield(task)
+
+    def _discard_completed(self, key: str, task: asyncio.Task[ProfileLookupStatus]) -> None:
+        if self._in_flight.get(key) is task:
+            self._in_flight.pop(key, None)
+
+    async def _perform_lookup(
+        self,
+        key: str,
+        lookup: Callable[[], Awaitable[Dict[str, Any]]],
+    ) -> ProfileLookupStatus:
+        try:
+            response = await lookup()
+            status = response.get("status") if response.get("type") == "profile_lookup" else None
+            if status not in {"profile_found", "profile_not_found", "indeterminate"}:
+                status = "indeterminate"
+        except Exception:
+            status = "indeterminate"
+
+        if status != "indeterminate":
+            self._failures.pop(key, None)
+            return status
+
+        attempts = self._failures.get(key, (0, 0.0))[0] + 1
+        delay = self._backoff_seconds[min(attempts - 1, len(self._backoff_seconds) - 1)]
+        self._failures[key] = (attempts, self._now() + delay)
+        return "indeterminate"
+
+
 class MarmotAgentControlClient:
     """Small NDJSON client for ``crates/agent-control``."""
 
@@ -833,6 +903,22 @@ class MarmotAgentControlClient:
 
     async def account_list(self) -> Dict[str, Any]:
         return await self.request({"type": "account_list"})
+
+    async def account_lookup_profile(self, account_id_hex: str) -> Dict[str, Any]:
+        response = await self.request(
+            {
+                "type": "account_profile_lookup",
+                "account_id_hex": _normalize_hex(account_id_hex, "account_id_hex"),
+            }
+        )
+        status = response.get("status")
+        if (
+            response.get("type") != "profile_lookup"
+            or status not in {"profile_found", "profile_not_found", "indeterminate"}
+            or not isinstance(response.get("retryable"), bool)
+        ):
+            raise AgentControlError("wn-agent returned invalid profile_lookup response", code="protocol_error")
+        return response
 
     async def account_publish_profile(
         self,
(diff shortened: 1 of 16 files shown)


agent-p1p:
Adversarial review at exact head `0fd63f26e0d260169279534abec2204ed0dc1caa`.

BLOCKING:
- none

SUGGESTION:
- New-adapter/old-connector version skew is safe but silent: `integrations/hermes/marmot/adapter.py:850` and `integrations/openclaw/marmot/src/profile-onboarding.ts:296`. An older `wn-agent` cannot deserialize `account_profile_lookup`; both gates conservatively collapse that protocol failure to `indeterminate`, remain quiet, and periodically retry. This correctly avoids a false prompt or overwrite, but automatic onboarding stays disabled without a distinct compatibility signal. Consider documenting the coupled rollout requirement or surfacing an explicit unsupported-operation result in a follow-up.

NITPICK:
- none

SUMMARY: The implementation satisfies the issue. The connector exposes one typed found/not-found/indeterminate operation and routes it through the existing account-directory path. Relay failures stay indeterminate; signature, author, kind, and JSON-object validation happen before a kind-0 can satisfy onboarding. Both integrations persist account-level `profile_exists`, deduplicate concurrent lookups and prompt claims, cap retry backoff, avoid consuming normal messages when lookup is found/indeterminate, and retain the manual publication flow. I traced the request through connector/runtime/directory layers and found no blocking correctness, security, privacy, or complexity issue.

Verification rerun on this head:
- `cargo test -p agent-control account_profile_lookup_has_typed_found_absent_and_indeterminate_outcomes`
- `cargo test -p agent-connector connector_profile_lookup_`
- `cargo test -p marmot-app directory_event_validation_rejects_invalid_signatures_and_wrong_authors`
- `cargo test -p marmot-app malformed_profile_content_is_not_treated_as_an_existing_profile`
- Hermes profile-onboarding selection: 11 passed; concurrent-first-message regression: 1 passed
- OpenClaw profile-onboarding + client suites: 45 passed; `pnpm typecheck` passed
- `cargo fmt --all --check`
- GitHub required CI is green; CodeRabbit reported success but was review-rate-limited.

VERDICT: APPROVE

SENSITIVE_PATHS: `crates/agent-control/src/lib.rs`, `crates/agent-connector/src/account.rs`, `crates/agent-connector/src/connection.rs`, `crates/marmot-app/src/directory/records.rs`, `crates/marmot-app/src/relay_plane/directory.rs`, `crates/marmot-app/src/runtime/mod.rs`, `integrations/hermes/marmot/adapter.py`, `integrations/openclaw/marmot/src/profile-onboarding.ts`

agent-p1p:
@coderabbitai review

integrations/openclaw/marmot/src/client.ts
```diff
@@ -287,6 +287,29 @@ export class MarmotAgentControlClient {
...
+    const response = await this.request({
+      type: "account_profile_lookup",
+      account_id_hex: normalizeHex(accountIdHex, "account_id_hex"),
+    });
+    if (
+      response.type !== "profile_lookup" ||
+      !["profile_found", "profile_not_found", "indeterminate"].includes(String(response.status)) ||
+      typeof response.retryable !== "boolean"
+    ) {
+      throw new Error("wn-agent returned invalid profile_lookup response");
```
erskingardner, line 304: Every other validation failure in this file throws `AgentControlError` with a `code` (see `normalizeHex`, timeout/socket errors, `id_mismatch`, etc.), and `isRetryable()` is built specifically around `instanceof AgentControlError`. This new method throws a plain `Error` instead, breaking that contract for callers of `accountLookupProfile`. The Python mirror (`adapter.py`'s `account_lookup_profile`) correctly uses `AgentControlError(..., code="protocol_error")` for the same case — worth matching here for consistency:

```ts
throw new AgentControlError("wn-agent returned invalid profile_lookup response", { code: "protocol_error" });
```

Not currently a functional bug (the only caller in `profile-onboarding.ts` uses a bare `catch` that treats any thrown error as `indeterminate`), but it's a footgun for future callers of this client method.

integrations/hermes/marmot/adapter.py
```diff
@@ -2764,21 +2851,45 @@ def _take_pending_ambient_context(self, group_id_hex: str) -> Optional[str]:
         return "\n".join(pending)
 
     async def _maybe_send_profile_prompt_on_join(self, account_id_hex: str, group_id_hex: str) -> None:
+        await self._maybe_prompt_for_missing_profile(account_id_hex, group_id_hex)
+
+    async def _maybe_prompt_for_missing_profile(
```
erskingardner, line 2856: Minor: the old inline implementation here had a comment explaining why `try_claim_prompt` must be the atomic claim called immediately before send — referencing the #513 double-prompt race across concurrent per-group dispatch. That rationale got dropped when the logic moved into this shared helper (it's still just as relevant: two concurrent callers can now also share one `_profile_lookup_gate.lookup()` before racing on `try_claim_prompt`). Worth restoring a short comment near the `try_claim_prompt` call so the invariant doesn't get lost for the next person touching this. Non-blocking — behavior and test coverage are correct.

erskingardner (approved):
Solid, well-scoped fix that closely tracks #1126's acceptance criteria. The core mechanism is correct:

- The connector's new `AccountProfileLookup` reuses the existing `fetch_current_user_profile_for_account_id` directory path rather than inventing a new one, and correctly maps "no relays configured" and any fetch error to `Indeterminate` (never collapsed into "not found").
- `validated_directory_event` in `relay_plane/directory.rs` adds real signature/kind/author verification to the shared one-shot directory fetcher (previously trusted the relay's filtering blindly) — this closes a real gap and benefits every directory lookup, not just this one. Good test coverage for tampered/wrong-author events.
- `profile_from_record` now rejects non-object kind:0 content (`null`, `[]`, a bare string) instead of silently producing an "empty but existing" profile — directly fixes the "malformed metadata must count as absent" requirement.
- The `ProfileLookupGate` (Hermes) / `ProfileLookupGate` (OpenClaw) pair correctly dedupes concurrent group-join/inbound-message triggers to a single in-flight lookup and backs off indeterminate results with bounded retry, with matching persisted `profile_exists` state on both sides. Concurrency and restart behavior are exercised by tests on both stacks.
- All the `AgentControlRequest`/`AgentControlResponse` exhaustive match sites (bootstrap.rs, validation.rs, opencode's control.rs) were updated consistently.

Two small non-blocking notes left inline:
1. `integrations/openclaw/marmot/src/client.ts` — `accountLookupProfile`'s protocol-shape guard throws a plain `Error` instead of this file's established `AgentControlError` (with `code`), unlike every other validation failure in the file and unlike the Python mirror (which does use `code="protocol_error"`). Not a functional bug today (the only caller swallows any exception into "indeterminate"), just a consistency footgun for future callers.
2. `integrations/hermes/marmot/adapter.py` — the refactor into the shared `_maybe_prompt_for_missing_profile` dropped the comment explaining why `try_claim_prompt` must be the atomic pre-send claim (the #513 double-prompt rationale). Behavior and tests are still correct, just worth restoring the comment for future maintainers.

Verification section in the PR description is thorough (fast-ci, targeted crate tests, both integration test suites, dev-script tests). Recommend merge once the maintainer decides whether to take the two nits.

PR #1134, opened 2026-07-27
Return from app startup at local readiness

erskingardner:
## Summary

- add a bounded MarmotKit `recordHostPerformance` API for `splashReady` and `foregroundLocalReady`
- export histogram sums alongside buckets and counts
- make account workers signal readiness after local hydration and snapshot capture
- move signer setup, relay activation, group-subscription registration, directory sync, and initial catch-up off the blocking startup path
- add separate transport-activation and subscription-registration telemetry
- document the updated `Marmot.start()` completion contract

## Why

`Marmot.start()` previously waited for transport activation and subscription registration before signalling readiness. Slow or unreliable relays could therefore hold the app on its splash screen even though the locally persisted chat list was already available.

The runtime should expose locally persisted state immediately and let relay connection and catch-up update projections asynchronously.

## App impact

The UniFFI `Marmot.start()` signature and result type are unchanged. Its completion contract changes: success now means persisted account state and worker-routed local reads are ready. It no longer means relay activation, directory sync, subscription registration, or initial catch-up have completed.

Hosts can render local projections when `start()` returns and represent relay progress separately. Mutations received during initial catch-up remain deferred and are replayed in order.

The new host-performance binding is synchronous and non-throwing:

```swift
marmot.recordHostPerformance(
    operation: .splashReady,
    durationMs: durationMs,
    outcome: .success
)
```

No workspace version bump or release-preparation changes are included; those are deferred until the remaining binding work is merged.

## Validation

- `just fast-ci`
- `cargo test -p marmot-app --lib` (452 passed)
- `cargo test -p marmot-uniffi --lib` (68 passed)
- `cargo check -p marmot-uniffi --all-targets --features otlp-export`
- `cargo test -p marmot-app --features otlp-export --test relay_telemetry_otlp`
- focused cursor, malformed-event, since-floor, relay-runtime, and new local-readiness regression tests
- generated the Swift binding and iOS XCFramework successfully

## Summary by CodeRabbit

* **New Features**
  * Startup now returns once local account data and chat projections are ready, while relay connection, subscriptions, synchronization, and catch-up continue in the background.
  * Added host-app performance reporting for splash-screen and foreground readiness milestones.
  * Added expanded startup and account performance metrics, including duration totals in exported histograms.

* **Documentation**
  * Clarified startup readiness guarantees and guidance for rendering local content while network updates continue asynchronously.
  * Expanded telemetry documentation and metric coverage.


crates/marmot-app/src/lib.rs
```diff
@@ -1193,6 +1200,19 @@ impl MarmotApp {
...
+            // reconciled group set when it registers subscriptions.
+            client.app.save_state(&client.state)?;
+        }
+        // This once-only projection migration is local SQL work and must land
+        // before the host renders its first chat-list snapshot.
+        client.backfill_self_membership_once()?;
+        Ok(client)
+    }
+
+    async fn finish_client_open_network_maintenance(&self, client: &mut AppClient) {
```
erskingardner, line 1215: **CI-confirmed regression**: `CLI e2e (mock relay)` fails deterministically here (3/3 nextest retries) — `keys_list_reports_startup_published_key_package` asserts `keys[0]["relay"] == true` and gets `false` (`crates/cli/tests/cli.rs:2735`): https://github.com/marmot-protocol/mdk/actions/runs/30249165568/job/89922990746

Root cause is very likely here: `finish_client_open_network_maintenance` (transport-adjacent key-package retire/cutover-replacement) and the transport activation/subscription registration it follows are no longer run synchronously before an account worker is considered "ready" — they're deferred into the async `initial_sync` future in `account_worker.rs`. For the account-worker path that's the whole point of this PR, but it changes the completion contract for `wn account create` / any one-shot CLI invocation that used to rely on this work having landed (and the local key-package "relay" status having settled) before the process's `runtime.shutdown()` runs or before a follow-up invocation's freshly-opened worker answers a read. `AccountManager::create_or_import_account`'s own key-package publish (via `app.client()` / `client_with_relay_plane`) still runs synchronously, but the account worker spawned by the trailing `reconcile()` now races/overlaps it asynchronously in a way it previously couldn't.

Please reproduce and fix before merge — `just fast-ci` / the targeted `cargo test` runs in the PR description don't cover the CLI e2e suite, so this slipped through local validation.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -723,15 +731,28 @@ async fn run_app_runtime_account_worker(
...
-                                // The reopen re-hydrates + reconnects + resubscribes
-                                // (via `runtime_client`) but does NOT run the
-                                // catch-up `sync()` on the readiness path — it
-                                // resumes the live `next_event` tail below — so it
-                                // does not reintroduce the catch-up blocking this
-                                // worker removed from startup.
+                                // `runtime_local_client` is local-only. Reconnect must
+                                // establish subscriptions and catch up before
+                                // replacing the prior live client.
+                                let reopen_sync = tokio::select! {
```
erskingardner: Worth double-checking: this reconnect path now awaits a full `reopened.sync()` (transport activation + subscription registration + relay catch-up drain) synchronously inside the steady-state `tokio::select!` loop, with no timeout guard — unlike the `maintenance_tick` arm further down in this same file, which wraps `client.sync()` in `timeout(Duration::from_secs(15), ...)` specifically to bound worst-case latency.

The comment this replaced explicitly called out that the old design avoided this: *"does NOT run the catch-up `sync()` on the readiness path... so it does not reintroduce the catch-up blocking this worker removed from startup."* Now, any reconnect (e.g. after a network blip) blocks the whole command loop — no `commands.recv()` arm is live — until the full catch-up completes, which is exactly the kind of stall this PR sets out to eliminate at cold start. If a relay is slow-but-connected (no EOSE, no connect-timeout trip), this could stall command processing for a long, unbounded time.

If the catch-up-on-reconnect is intentional, consider bounding it the same way `maintenance_tick` does. If not, `reopened.prepare_transport()` (activate + resubscribe only) would restore the prior reconnect semantics and let the resumed `receive_next_delivery()` tail pick up the backlog.

erskingardner (commented):
## Summary

The intent here is good and the telemetry/histogram-sum piece (commit 1) is clean and well tested. The startup-readiness refactor (commit 2) is the risky part, and CI has already caught a real regression from it: **`CLI e2e (mock relay)` fails deterministically** (3/3 nextest retries, not a flake) on `keys_list_reports_startup_published_key_package` — `assert_eq!(keys[0]["relay"], true)` gets `false` at `crates/cli/tests/cli.rs:2735`:
https://github.com/marmot-protocol/mdk/actions/runs/30249165568/job/89922990746

This lines up exactly with what this PR moves off the synchronous startup path: transport activation, subscription registration, and the key-package retire/cutover-replacement logic in `finish_client_open_network_maintenance` now run asynchronously after the worker signals "local ready," rather than before. A short-lived, one-shot CLI invocation (`wn account create` then a fresh `wn keys list` process) doesn't have the old guarantee that this work has landed before the process's runtime shuts down / before a freshly-started worker in the next invocation answers a query, so relay-confirmation state that used to be settled by the time the command returned no longer reliably is. This needs a fix (and a passing CI run) before merge — the CLI's one-shot process lifecycle is a real usage pattern this PR doesn't seem to have been validated against (the PR's own validation section lists targeted `cargo test` runs and `just fast-ci`, not the CLI e2e suite).

Everything else (relay/rust test matrix, clippy, fmt, tamarin, simulator) is green, and the rest of the diff (histogram `sum_ms`, `HostPerformanceOperation`/`HostPerformanceOutcome` closed-enum telemetry API, doc updates) is solid, well-scoped, and consistent with the repo's privacy/observability conventions.

## Findings

1. **Blocking — CI-confirmed regression**: see inline comment on `finish_client_open_network_maintenance`.
2. **Worth double-checking (non-blocking)**: see inline comment on the reconnect path in `account_worker.rs` — it now runs a full catch-up `sync()` synchronously with no timeout guard, reintroducing (only on reconnect) the blocking behavior this PR otherwise removes from startup.

PR #1136, opened 2026-07-27
test(hermes): exercise persisted config startup

erskingardner:
## Summary

- add a pinned real-Hermes integration probe for persisted Marmot configuration with no inherited `MARMOT_*` environment
- mirror Hermes' persisted-over-environment probe precedence and cover overlapping persisted/environment socket values
- wire the probe into the Hermes dev harness, documentation, and WN Agent release workflow

## Why

This is the follow-up commit intended for #1130 that missed the merge train. PR #1130 fixed persisted-config gateway readiness, but the merged test harness used different precedence from real Hermes and CI did not exercise the actual persisted-config startup boundary. This closes both gaps without changing production adapter behavior.

## Verification

- pinned real-Hermes persisted-config probe: passed
- Hermes Python suite under Python 3.11: 156 passed
- `integrations/hermes/marmot/tests/test_dev_scripts.sh`: passed
- `just fast-ci`: passed, including convergence policy pin 5/5
- `git diff --check origin/master..HEAD`: passed

## Notes

The macOS system Python 3.14 run exposes a pre-existing malformed-YAML transaction-test failure that reproduces unchanged on `origin/master`; the supported Python 3.11 suite is green.

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end verification workflow for persisted Marmot configuration across gateway restarts.
  * Added a command and generated helper script to verify persisted configuration in isolated development environments.
  * Added integration checks for plugin discovery, gateway settings, socket connectivity, and adapter configuration.

* **Bug Fixes**
  * Persisted platform settings now take precedence over overlapping environment values during connection checks.

* **Documentation**
  * Expanded Marmot development setup and persisted-configuration verification instructions.

### DIFF
--- a/.github/workflows/wn-agent-binaries.yml
+++ b/.github/workflows/wn-agent-binaries.yml
@@ -5,6 +5,9 @@ on:
     paths:
       - .github/workflows/wn-agent-binaries.yml
       - scripts/install-hermes-marmot.sh
+      - scripts/hermes_marmot_configure_gateway.py
+      - scripts/hermes_marmot_dev_setup.sh
+      - scripts/hermes_marmot_verify_persisted_config.sh
       - integrations/hermes/marmot/**
       - scripts/install-openclaw-marmot.sh
       - integrations/openclaw/marmot/**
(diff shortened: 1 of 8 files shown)


erskingardner (commented):
## Review

**Recommend merge.** (Posting as COMMENT, not APPROVE — GitHub blocks self-approval since this PR is opened by the authenticated account.)

Verified this against the actual reason it exists: I cross-checked the new `PlatformRegistryHarness.apply_env_overrides` precedence change in `integrations/hermes/marmot/tests/test_configure_gateway.py` against the real upstream `_apply_env_overrides` in `NousResearch/hermes-agent` at the pinned commit `c7b75a7849cb260e6f17a045473bfdd0ea21ca81` (`gateway/config.py` lines ~2598-2641). Upstream does exactly what this PR mirrors: `probe_extra.setdefault(k, v)` for the enablement-gate probe (persisted config wins), then `config.platforms[platform].extra.update(seed)` for the final committed extra (env wins) — two different precedence rules for two different purposes. This PR's test harness now matches both correctly, and the new regression test (`test_persisted_extra_wins_over_overlapping_env_seed_during_probe`) exercises the previously-untested probe path. This is a faithful, verified mirror, not a guess — backed by the pinned real-Hermes CI job actually exercising the real code.

No production code (`integrations/hermes/marmot/adapter.py`) is touched, matching the PR's stated intent. `scripts/hermes_marmot_verify_persisted_config.sh` is careful about environment scrubbing (`env -i` with an explicit allowlist) and background-process cleanup (trap + wait). The new CI job (`hermes-marmot-integration`) is correctly wired into the release job's `needs:` list so a regression here blocks releases the same way #1127 slipped through before.

Two minor, non-blocking notes:

1. **CI job now depends on a third-party repo's transitive dependencies at install time.** `hermes_marmot_dev_setup.sh` pins the Hermes *commit*, but `uv pip install -e ".[all,dev]"` still resolves against whatever PyPI has for its dependencies at CI run time. Pinning the ref doesn't prevent a future upstream dependency yank/break from turning this into release-blocking flakiness later. Worth watching if this job starts flaking; not a reason to hold this PR.
2. **`Justfile` `hermes-verify-persisted-config` recipe** (lines 98-105) uses a `#!/usr/bin/env bash` shebang body for a simple optional-arg passthrough — no other recipe in this Justfile does that. A native `just` conditional expression would be shorter and more consistent with the rest of the file, e.g. `{{ if root == "" { "" } else { "--root " + root } }}`. Purely stylistic.

PR #1138, opened 2026-07-27
feat(uniffi): surface durable pending leave intent

erskingardner:
## Why

Leaving a group is two-phase. `cgka-engine` records a durable `LeaveRequest` (`crates/traits/src/storage.rs:155`, table `cgka_leave_requests`) **before** it publishes, because MLS SelfRemove proposals are epoch-bound while the product intent is not — the engine re-proposes each new epoch until a commit actually removes us.

Terminal `SelfMembership::Left` is only written at the *end* of `leave_group_with_audit_context`. So a publish failure (`fail_if_publish_failed` returns early) or app termination mid-flight left the intent durable on disk but invisible above the engine boundary: `leave_request` appeared only in `cgka-engine`, `storage-sqlite`, and `traits`, with no accessor on `Engine`, `cgka_session::Session`, `marmot-app`, `marmot-uniffi`, or the CLI.

Reported from mobile:

- A cold launch could not rediscover a pending leave — the group rendered as a normal active conversation with an enabled Leave button.
- Tapping Leave again in the same epoch hit `EngineError::InvalidTransition { from: "Leaving" }`, which reached the host as an opaque `MarmotKitError::Runtime`.

## Approach

**Derive at read time from `cgka_leave_requests`. No new column.**

`LeaveRequestStorage` is already implemented for `SqliteAccountStorage`, and `MarmotApp::account_storage()` returns exactly that type against the same `session.sqlite` file — so this needs **no migration, no backfill, and no new engine or session accessor**. The table was directly readable from the app layer all along, just never read.

It is also the correct source of truth, not merely the cheap one: the engine clears leave requests from five paths that never notify the app projection (`engine.rs:1466`, `ingest.rs:1322`, `ingest.rs:2006`, `send.rs:759`, `distributed_convergence.rs:1403`). A denormalized column would silently go stale.

`rebuild_chat_list_row_for_group_tx` and `chat_list_projection_complete_tx` are **deliberately untouched**, with comments saying so. There is no stored column to keep fresh, and comparing a read-time-only field in the freshness predicate would mark every row stale forever.

Two population points cover the entire surface:

| Populated in | Reaches |
| --- | --- |
| `chat_list_rows_tx` / `chat_list_row_tx` | `chat_list`, `chat_list_row`, `subscribe_chat_list`, `ChatListSubscription`, `wn-cli` JSON |
| `MarmotApp::groups` | `group`, `visible_groups`, `subscribe_chats` (snapshot **and** deltas — both read through `app.groups`/`app.group`), `group_details`, `group_management_state` |

No new `ChatListUpdateTrigger` variant. `chat_list_row_fingerprint` and `app_group_record_fingerprint` both serde the whole struct, so change detection picks the field up automatically, and `MembershipChanged` is already the accurate trigger — adding a variant would break exhaustive `switch`/`when` in host code for no gain.

## Host-visible surface

`ChatListRowFfi`, `AppGroupRecordFfi`, and `GroupManagementStateFfi` gain `leave_request_pending: bool` and `leave_requested_at_ms: Option<u64>`. Internally only the `Option<u64>` is carried and the FFI layer derives the bool, so the pair cannot hold a contradictory value.

Behavior changes a reviewer should weigh:

- **`GroupManagementStateFfi.can_leave` is now `false` while a leave is pending.** It now means "a Leave tap will do something"; `leave_request_pending` is what distinguishes "already leaving" from "admin, must self-demote first". This changes existing semantics for hosts, chosen so the button disables rather than erroring on a second tap.
- **New `MarmotKitError::LeaveAlreadyRequested`.** Checked **first** in `leave_group`'s preflight — order matters, because the tightened `can_leave` would otherwise report `MemberNotInGroup` for a group the account is very much still in.

## Two deviations from the original plan

- **The publish-failure notification moved.** The plan called for `refresh_group` in the client's error branch; that would have notified nobody, since `refresh_group` only mutates in-memory state. The events channel lives in the worker, so the `LeaveGroup` arm now publishes `GroupStateUpdated` on both paths. A no-op re-read is cheap — `subscribe_chat_list` fingerprint-dedupes it away.
- **Pending check ordering** in `leave_group`, as described above.

## Verification

- `just fast-ci` green.
- `cargo test -p storage-sqlite` (279), `-p marmot-uniffi` (70 + 25 smoke), `-p marmot-app` (451 unit) pass.
- `pending_leave_request_survives_a_cold_launch` reproduces the reported bug: bob leaves, alice has not committed the SelfRemove, and a fresh `MarmotApp` over the same directory still reports the intent on both the chat-list row and the group record.
- Generated Kotlin confirms the contract actually changed (`output/` is gitignored and CI never builds bindings, so this is the only real check): fields land on `AppGroupRecordFfi`, `ChatListRowFfi`, `GroupManagementStateFfi`, plus the `LeaveAlreadyRequested` exception class. Only the bindgen half of `kotlin-bindings.sh` was run — the four Android ABI cross-builds do not affect the generated surface, and `cargo-ndk`/`ANDROID_NDK_HOME` are not configured on this machine.

**Correction to an earlier revision of this description:** it claimed 5 `relay_runtime` tests were failing pre-existing. That was wrong — my invocation was missing `test-policy-overrides`, which the pinned convergence policy (mdk#970) requires. With the flag CI uses, `relay_runtime` is 78/78 green. No pre-existing failures.

## Release note

CHANGELOG `Unreleased` entry added; **no `Cargo.toml` bump**. The UniFFI record and error-variant changes do require a workspace version bump per `crates/marmot-uniffi/AGENTS.md` and `release.md`, but that happens at release time while Unreleased entries accumulate.

`chats list --json` / `chat_list_row` rows gain a `leave_requested_at_ms` field; no other JSON response shapes changed. Rendering "Leaving…" in the TUI was left out of scope.

---

## Review round 2 (`7fcaa6eb`)

Addresses both findings from @erskingardner's review.

### P1 — already-requested classification is now atomic

The preflight was a TOCTOU check. Two concurrent leaves could both observe `false`; the worker serialized them and the loser got `EngineError::InvalidTransition` → the untyped `Runtime` bucket, preserving the exact opaque error this PR set out to remove.

New `EngineError::LeaveAlreadyRequested`, returned from `do_send_leave` under the same lock as the durable read and write, so it cannot be raced past. The preflight stays as a fast path, demoted to an optimization with the engine as the authority.

**Worth noting beyond the review:** `InvalidTransition` was the wrong carrier regardless of the race — [error.rs](../blob/master/crates/traits/src/error.rs) documents it as *"Indicates an engine bug"*, and a double-tap is routine input. An ordinary send blocked by a current leave stays `InvalidTransition`; that one genuinely is an illegal transition. The forensic-audit and conformance classifiers gain matching `leave_already_requested` arms (no vector asserted the old string).

Three regression tests, one per seam:

| Seam | Asserts |
| --- | --- |
| `cgka-engine/tests/invite_leave.rs` | repeats refused by name, repeatably; durable request left byte-identical |
| `marmot-app/tests/relay_runtime.rs` | two `leave_group` calls via `tokio::join!` → exactly one success, one typed refusal |
| `marmot-uniffi/src/errors.rs` | variant crosses FFI as `LeaveAlreadyRequested`; `InvalidTransition` stays untyped |

The runtime-level path has no preflight (that lives in the uniffi layer), so the losing call necessarily reaches the engine — the concurrent test exercises the real classification rather than the precheck.

### P2 — the lifecycle docs were wrong

Correct, and my error: the leave path writes `SelfMembership::Left` as soon as the proposal publishes, so `Left` does not imply a commit removed the member. My field docs claimed the opposite while my own cold-launch test narrated the truth in a comment.

Now documented as orthogonal, with a state table on `ChatListRowFfi`:

| `self_membership` | `leave_request_pending` | meaning |
|---|---|---|
| `Member` | `true` | leave requested, publish failed or was interrupted |
| `Left` | `true` | leave published, awaiting a member to commit it |
| `Left` | `false` | resolved |
| `Removed` | `false` | removed by someone else |

Fixed in all four declarations, the CHANGELOG, and the test comment — and the coexistence is now asserted rather than only described, so the docs can't drift back.

### Also in this round

Merged `origin/master` (`6b4de626`). Only the CHANGELOG conflicted, additively; the four Rust files auto-merged, and I hand-verified the `LeaveGroup` worker arm since #1134 rewrote 117 lines of that file.

Verification: `just fast-ci` green; `cgka-traits`, `storage-sqlite`, `marmot-uniffi`, `cgka-engine`, `cgka-conformance-simulator`, and `marmot-app` suites all green with the feature flag CI uses.

## Summary by CodeRabbit

* **New Features**
  * Added visible pending-leave status and request timestamp to group and chat list information.
  * Pending leave requests now remain discoverable after restarting the app.
  * Leave controls are disabled while a request is in progress.
  * Added a specific “leave already requested” error for repeated or concurrent leave attempts.

* **Bug Fixes**
  * Group state updates now appear immediately after leave attempts, including failures.
  * Pending leave status is derived consistently across chat lists, group views, and bindings.

### DIFF
--- a/crates/marmot-uniffi/src/commands/group.rs
+++ b/crates/marmot-uniffi/src/commands/group.rs
@@ -378,9 +378,15 @@ impl Marmot {
         let group_id_hex = hex::encode(group_id.as_slice());
         let state =
             group_management_state_for(self, &account_ref, &group_id, &group_id_hex).await?;
-        // Checked first: a pending leave also clears `can_leave`, so letting the
-        // checks below run would report a wrong reason (`MemberNotInGroup`) for a
-        // group the account is very much still in.
+        // A fast-path reject only, not the source of correctness: this read and
+        // the worker command that acts on it are not atomic, so two concurrent
+        // leaves can both pass here. The engine settles it under its own lock and
+        // returns `EngineError::LeaveAlreadyRequested`, which maps to this same
+        // variant — see `MarmotKitError::from_engine_error`.
+        //
+        // Checked before the two below because a pending leave also clears
+        // `can_leave`, so letting them run would report a wrong reason
+        // (`MemberNotInGroup`) for a group the account is very much still in.
         if state.leave_request_pending {
             return Err(MarmotKitError::LeaveAlreadyRequested {
                 group_id_hex: group_id_hex.clone(),
--- a/crates/marmot-uniffi/src/commands/group.rs
+++ b/crates/marmot-uniffi/src/commands/group.rs
@@ -378,6 +378,14 @@ impl Marmot {
         let group_id_hex = hex::encode(group_id.as_slice());
         let state =
             group_management_state_for(self, &account_ref, &group_id, &group_id_hex).await?;
+        // Checked first: a pending leave also clears `can_leave`, so letting the
+        // checks below run would report a wrong reason (`MemberNotInGroup`) for a
+        // group the account is very much still in.
+        if state.leave_request_pending {
+            return Err(MarmotKitError::LeaveAlreadyRequested {
+                group_id_hex: group_id_hex.clone(),
+            });
+        }
         if state.requires_self_demote_before_leave {
             return Err(MarmotKitError::AdminCannotSelfRemove {
                 group_id_hex: group_id_hex.clone(),
@@ -873,6 +881,8 @@ mod tests {
             can_invite: self_admin,
             can_leave: !self_admin,
             requires_self_demote_before_leave: self_admin,
+            leave_request_pending: false,
+            leave_requested_at_ms: None,
             member_actions: vec![
                 GroupMemberActionStateFfi {
                     member_id_hex: self_id.into(),
--- a/crates/marmot-uniffi/src/conversions/chat_list.rs
+++ b/crates/marmot-uniffi/src/conversions/chat_list.rs
@@ -102,10 +102,24 @@ pub struct ChatListRowFfi {
     /// resolved yet. Render the conversation as leaving, and do not offer Leave
     /// again — see `GroupManagementStateFfi::can_leave`.
     ///
-    /// `self_membership` only reaches a terminal `Left` once a commit actually
-    /// removes the local member, which may be many epochs later or never (a
-    /// publish can fail, and the intent survives app termination). This flag is
-    /// the durable intent in between, so a cold launch can rediscover it.
+    /// Durable unresolved *intent*, which survives a failed publish and app
+    /// termination — so a cold launch can rediscover it. Read this rather than
+    /// `self_membership` to decide whether to show a leave in progress.
+    ///
+    /// This is orthogonal to `self_membership`, not a precursor to it. The two
+    /// answer different questions and combine freely:
+    ///
+    /// | `self_membership` | this flag | meaning |
+    /// |---|---|---|
+    /// | `Member` | `true`  | leave requested, publish failed or was interrupted |
+    /// | `Left`   | `true`  | leave published, still waiting for a member to commit it |
+    /// | `Left`   | `false` | leave resolved |
+    /// | `Removed`| `false` | removed by someone else |
+    ///
+    /// `self_membership` is the locally *classified departure* — `Left` is
+    /// recorded as soon as the SelfRemove proposal publishes, so it does **not**
+    /// imply a commit removed the member, and `Removed` marks an involuntary
+    /// removal. This flag is about whether the request is still outstanding.
     ///
     /// Always equal to `leave_requested_at_ms != null`.
     pub leave_request_pending: bool,
--- a/crates/marmot-uniffi/src/conversions/chat_list.rs
+++ b/crates/marmot-uniffi/src/conversions/chat_list.rs
@@ -98,6 +98,20 @@ pub struct ChatListRowFfi {
     /// Whether the local account is still a member of this group, and if not,
     /// whether it left voluntarily or was removed.
     pub self_membership: SelfMembershipFfi,
+    /// The local account asked to leave this group and the request has not
+    /// resolved yet. Render the conversation as leaving, and do not offer Leave
+    /// again — see `GroupManagementStateFfi::can_leave`.
+    ///
+    /// `self_membership` only reaches a terminal `Left` once a commit actually
+    /// removes the local member, which may be many epochs later or never (a
+    /// publish can fail, and the intent survives app termination). This flag is
+    /// the durable intent in between, so a cold launch can rediscover it.
+    ///
+    /// Always equal to `leave_requested_at_ms != null`.
+    pub leave_request_pending: bool,
+    /// When the local account asked to leave, in milliseconds since the Unix
+    /// epoch; `null` when no leave is pending.
+    pub leave_requested_at_ms: Option<u64>,
 }
 
 impl From<ChatListRow> for ChatListRowFfi {
@@ -122,6 +136,8 @@ impl From<ChatListRow> for ChatListRowFfi {
             activity_sort_at: value.activity_sort_at,
             updated_at: value.updated_at,
             self_membership: value.self_membership.into(),
+            leave_request_pending: value.leave_requested_at_ms.is_some(),
+            leave_requested_at_ms: value.leave_requested_at_ms,
         }
     }
 }
@@ -192,9 +208,8 @@ impl From<marmot_app::ChatListUpdateTrigger> for ChatListUpdateTriggerFfi {
 mod tests {
     use super::*;
 
-    #[test]
-    fn chat_list_row_exports_semantic_timestamps() {
-        let ffi = ChatListRowFfi::from(ChatListRow {
+    fn sample_row() -> ChatListRow {
+        ChatListRow {
             group_id_hex: "11".to_owned(),
             archived: false,
             pending_confirmation: false,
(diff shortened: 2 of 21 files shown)


crates/marmot-uniffi/src/commands/group.rs
```diff
@@ -378,6 +378,14 @@ impl Marmot {
         let group_id_hex = hex::encode(group_id.as_slice());
         let state =
             group_management_state_for(self, &account_ref, &group_id, &group_id_hex).await?;
+        // Checked first: a pending leave also clears `can_leave`, so letting the
+        // checks below run would report a wrong reason (`MemberNotInGroup`) for a
+        // group the account is very much still in.
+        if state.leave_request_pending {
```
erskingardner, line 390: [P1] Make the already-requested classification atomic. This preflight is a TOCTOU check: two concurrent calls (including a rapid double tap that launches two async tasks) can both observe `false` before either worker command runs. The account worker then serializes them; the first records the request and the second receives `EngineError::InvalidTransition { from: "Leaving" }`, which `MarmotKitError::from_engine_error` still flattens to `Runtime`. That preserves the opaque error this PR is intended to remove. Please expose a dedicated engine/runtime error (or translate the exact engine result after the command) and add a concurrent-call regression test; the preflight can remain for fast UX, but it cannot be the source of correctness.

crates/marmot-uniffi/src/conversions/chat_list.rs
```diff
@@ -98,6 +98,20 @@ pub struct ChatListRowFfi {
     /// Whether the local account is still a member of this group, and if not,
     /// whether it left voluntarily or was removed.
     pub self_membership: SelfMembershipFfi,
+    /// The local account asked to leave this group and the request has not
+    /// resolved yet. Render the conversation as leaving, and do not offer Leave
+    /// again — see `GroupManagementStateFfi::can_leave`.
+    ///
+    /// `self_membership` only reaches a terminal `Left` once a commit actually
```
erskingardner: [P2] Correct the public lifecycle description. `AppClient::leave_group_with_audit_context` writes `SelfMembership::Left` immediately after a successful proposal publish, before another member commits the SelfRemove; the new cold-launch test even notes that optimistic `Left` state while the request remains outstanding. So this field can coexist with `self_membership == Left`, and `Left` does not imply that a commit actually removed the member. Please document the actual distinction: the pending request covers durable unresolved intent (including publish failure), while `self_membership` records the locally classified voluntary departure.

erskingardner (commented):
Blocking review findings: the read-time projection addresses cold-launch rediscovery, but the repeat-leave error contract is still racy at the host boundary. Please make the classification atomic at the engine/runtime seam (and cover concurrent calls), then correct the public membership lifecycle docs. The branch also currently conflicts with master and will need rebasing before the resulting code can be validated. GitHub does not allow the PR author to submit a request-changes review, so this is posted as a comment review.

crates/marmot-uniffi/src/commands/group.rs
erskingardner: Fixed in 7fcaa6eb — added `EngineError::LeaveAlreadyRequested`, returned from `do_send_leave` under the same lock as the durable read and write so it cannot be raced past, mapped it in `from_engine_error`, and demoted this preflight to a fast path with the engine as the authority; also agreed `InvalidTransition` was the wrong carrier regardless of the race since `error.rs` documents it as an engine bug, so the audit/conformance classifiers now emit `leave_already_requested` too, and added three regression tests (engine repeat-refusal, `tokio::join!` concurrent leaves through the runtime — which has no preflight, so the loser necessarily reaches the engine — and the FFI mapping).

crates/marmot-uniffi/src/conversions/chat_list.rs
erskingardner: Fixed in 7fcaa6eb — you're right and this was my error: the docs claimed the opposite of what my own cold-launch test narrated in a comment. All four declarations (`ChatListRow`, `AppGroupRecord`, and both FFI records), the CHANGELOG, and that test comment now describe the two as orthogonal — pending is durable unresolved intent, `self_membership` is the locally classified departure with `Left` written on publish — with a full state table on `ChatListRowFfi` covering `Member`+pending (publish failed), `Left`+pending (awaiting a committer), `Left` alone (resolved), and `Removed`; the coexistence is now asserted in the test rather than only described, so the docs can't drift back.

crates/marmot-uniffi/src/conversions/group.rs
```diff
@@ -335,7 +361,12 @@ pub(crate) fn group_management_state_ffi(
     let is_self_admin = self_member.is_some_and(|member| member.is_admin);
     let is_last_admin = is_self_admin && admin_count == 1;
     let can_invite = is_self_admin;
-    let can_leave = self_member.is_some() && !is_self_admin;
+    // A leave already in flight suppresses `can_leave` so hosts do not offer a
+    // second Leave that the engine would reject: the durable request is not
+    // epoch-bound, but the SelfRemove proposal backing it is, and re-requesting
+    // inside the same epoch is an invalid transition.
+    let leave_request_pending = details.group.leave_request_pending;
+    let can_leave = self_member.is_some() && !is_self_admin && !leave_request_pending;
```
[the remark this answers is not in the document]
erskingardner: Fixed in 209cecb2 — correct catch, that comment was accurate when written and went stale in the prior commit when the engine started returning the typed error; it now names `EngineError::LeaveAlreadyRequested` and its `MarmotKitError` counterpart, and adds that suppressing the affordance keeps the error off the happy path rather than preventing it (this flag isn't read atomically with the leave it guards). Also swept the tree for the same staleness — the remaining `InvalidTransition` references are correct, since the state machine owns the type and the send-after-eviction and send-blocked-while-leaving paths genuinely still return it.

PR #1140, opened 2026-07-27
fix(incident-replay): fail-close format detection on manifest-less NDJSON streams

### LINKED ISSUE #1037: incident-replay: is_stream misroutes a manifest-less/corrupt NDJSON export to the lenient document parser — a broken stream can classify as Healthy (fail-open at format detection)
### Summary

The NDJSON group-export path (`goggles-group-export/v1`) enforces a **fail-closed** completeness contract in `parse_stream`: it errors on a missing leading `manifest`, a missing terminal `eof{complete:true}`, mismatched section counts, or an in-band `error` line. But *which* parser runs is decided by `is_stream`, which returns `true` only when the first non-empty line deserializes as `TagProbe{ t == "manifest" }`. If the manifest is the thing that is missing or corrupt — exactly the truncation the contract exists to catch — `is_stream` returns `false` and the input is routed to the lenient document parser `export::parse`, which can silently adopt a broken stream as an empty (healthy) export.

### Location

- `crates/incident-replay/src/ndjson.rs:42-50` (`is_stream`)
- `crates/incident-replay/src/main.rs:44-60` (parser selection)
- `crates/incident-replay/src/export.rs:258-261` (`parse` = bare `serde_json::from_str`, no validation) and `:15-38` (`AgentStateExport` — every field `#[serde(default)]`, no `#[serde(deny_unknown_fields)]`)

```rust
// main.rs
let export = if is_stream(&json) {
    parse_stream(&json)?      // fail-closed
} else {
    parse(&json)?            // lenient: serde_json::from_str::<AgentStateExport>
};
```

### Failure scenario

`AgentStateExport` has all-`#[serde(default)]` fields and does not deny unknown fields, so **any single JSON object that is not a manifest** parses as an *empty* export (`events: []`, `derived_projections: {}`). Concretely: a stream truncated to just its terminal in-band error line —

```
{"t":"error","complete":false}
```

— is a single JSON object whose `t` is not `manifest`, so `is_stream` returns `false`, `parse` accepts it (the `t`/`complete` fields are ignored as unknown), and `classify` sees no contested convergence, no fork, and empty projections (truncation gate vacuously satisfied) and no per-engine liveness evidence → **`Verdict::Healthy`**, exit 0 (`healthy: 0 vectors`). A broken/aborted forensic export reads as a healthy group — precisely the false negative the streaming contract was built to prevent. This is a realistic remnant: the server can commit HTTP 200 and then only emit the in-band `error` line (its sole post-status failure surface, per the `ndjson.rs` module doc).

Scope note (honest bound): multi-line truncated remnants are *incidentally* protected because `serde_json::from_str` rejects trailing content after the first object, so they fail the document parse (exit 2). The fail-open is specific to a remnant that is exactly one JSON object.

### Why not a duplicate

#1024 / #844 / #801 concern `u64`/epoch arithmetic overflow on hostile values; #1022 concerns incident vectors never being executed in CI. None addresses `is_stream` / parser-selection fail-open — the format-detection boundary itself.

### Suggested severity

**MEDIUM** — this is a forensic triage tool: a truncated/corrupt export silently classified `Healthy` can cause a real incident to be missed. It is an offline analysis CLI (not a live boundary), which bounds the blast radius.

### Suggested fix

Make format detection fail-closed for stream-shaped input: if the first non-empty line carries a `t` discriminator at all, route to `parse_stream` so a manifest-less/corrupt stream yields a `StreamParseError` instead of being adopted by the document parser. Alternatively, have `export::parse` reject a top-level object that carries a `t` field (the stream discriminator) rather than defaulting it to an empty export.

---
*Filed from a meticulous automated code-review pass. The parser-selection path and serde defaults were verified against source, and the finding was checked against existing open issues (#1024, #844, #801, #1022) for duplicates before filing.*

singhlovepreet9:
### Summary

Previously, `is_stream` required the first non-empty line of an export to be a valid `manifest` line. When an NDJSON stream was truncated to a single JSON object without a leading manifest (for example, after an immediate server abort reporting only an in-band error line `{"t":"error","complete":false}`), format detection returned `false` and routed the input to the document parser `export::parse`. Because all document fields default cleanly, the remnant was silently adopted as an empty export and classified as healthy (`healthy: 0 vectors`).

This PR fixes this fail-open behavior by updating `is_stream` to recognize any line carrying a `t` discriminator as a stream. This ensures that manifest-less or corrupt streams fail closed with a `StreamParseError` instead of falling back to the document parser. As an additional layer of defense against accidental adoption, `export::parse` now explicitly rejects top-level objects carrying a `t` discriminator.

Fixes #1037

### Verification

- Added reproduction regression tests in `crates/incident-replay/tests/ndjson.rs` and `crates/incident-replay/tests/parse.rs` verifying that manifest-less stream remnants and stream-shaped lines are rejected and fail closed.
- Confirmed all existing `incident-replay` unit and integration tests pass without regression.
- Ran workspace-wide formatting and clippy checks (`just fast-ci` equivalents).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of NDJSON streams, including streams without a manifest.
  - Reports server-sent errors immediately during stream parsing.
  - Prevents streamed data from being misinterpreted as complete agent-state documents.
  - Provides clearer errors for malformed or stream-shaped JSON input.

- **Documentation**
  - Clarified stream detection rules and the supported stream format.

- **Tests**
  - Added coverage for manifest-less streams, terminal errors, malformed discriminators, and CLI parsing failures.

### DIFF
--- a/crates/incident-replay/src/export.rs
+++ b/crates/incident-replay/src/export.rs
@@ -432,16 +432,14 @@ impl EventKind {
 
 /// Parse a Goggles `agent-state.json` export.
 pub fn parse(json: &str) -> Result<AgentStateExport, ParseError> {
-    #[derive(Deserialize)]
-    struct StreamDiscriminatorProbe {
-        t: Option<String>,
-    }
-    if let Ok(StreamDiscriminatorProbe { t: Some(t) }) =
-        serde_json::from_str::<StreamDiscriminatorProbe>(json)
+    let value: serde_json::Value = serde_json::from_str(json)?;
+    if value
+        .as_object()
+        .is_some_and(|object| object.contains_key("t"))
     {
-        return Err(ParseError::StreamDiscriminator { t });
+        return Err(ParseError::StreamDiscriminator);
     }
-    Ok(serde_json::from_str(json)?)
+    Ok(serde_json::from_value(value)?)
 }
 
 /// Why an export could not be parsed.
@@ -449,8 +447,6 @@ pub fn parse(json: &str) -> Result<AgentStateExport, ParseError> {
 pub enum ParseError {
     #[error("agent-state export does not match the expected schema: {0}")]
     Json(#[from] serde_json::Error),
-    #[error(
-        "input is a streamed NDJSON line (discriminator `t` = `{t}`), not an agent-state document"
-    )]
-    StreamDiscriminator { t: String },
+    #[error("input carries the streamed NDJSON `t` discriminator, not an agent-state document")]
+    StreamDiscriminator,
 }
--- a/crates/incident-replay/src/export.rs
+++ b/crates/incident-replay/src/export.rs
@@ -432,14 +432,11 @@ impl EventKind {
 
 /// Parse a Goggles `agent-state.json` export.
 pub fn parse(json: &str) -> Result<AgentStateExport, ParseError> {
-    let value: serde_json::Value = serde_json::from_str(json)?;
-    if value
-        .as_object()
-        .is_some_and(|object| object.contains_key("t"))
-    {
+    let probe = serde_json::from_str::<BTreeMap<String, serde::de::IgnoredAny>>(json);
+    if probe.is_ok_and(|object| object.contains_key("t")) {
         return Err(ParseError::StreamDiscriminator);
     }
-    Ok(serde_json::from_value(value)?)
+    Ok(serde_json::from_str(json)?)
 }
 
 /// Why an export could not be parsed.
--- a/crates/incident-replay/src/export.rs
+++ b/crates/incident-replay/src/export.rs
@@ -432,6 +432,15 @@ impl EventKind {
 
 /// Parse a Goggles `agent-state.json` export.
 pub fn parse(json: &str) -> Result<AgentStateExport, ParseError> {
+    #[derive(Deserialize)]
+    struct StreamDiscriminatorProbe {
+        t: Option<String>,
+    }
+    if let Ok(StreamDiscriminatorProbe { t: Some(t) }) =
+        serde_json::from_str::<StreamDiscriminatorProbe>(json)
+    {
+        return Err(ParseError::StreamDiscriminator { t });
+    }
     Ok(serde_json::from_str(json)?)
 }
 
@@ -440,4 +449,8 @@ pub fn parse(json: &str) -> Result<AgentStateExport, ParseError> {
 pub enum ParseError {
     #[error("agent-state export does not match the expected schema: {0}")]
     Json(#[from] serde_json::Error),
+    #[error(
+        "input is a streamed NDJSON line (discriminator `t` = `{t}`), not an agent-state document"
+    )]
+    StreamDiscriminator { t: String },
 }
--- a/crates/incident-replay/src/ndjson.rs
+++ b/crates/incident-replay/src/ndjson.rs
@@ -43,10 +43,8 @@ pub fn is_stream(input: &str) -> bool {
     let Some(first_line) = input.lines().find(|line| !line.trim().is_empty()) else {
         return false;
     };
-    serde_json::from_str::<serde_json::Value>(first_line)
-        .ok()
-        .and_then(|value| value.as_object().map(|object| object.contains_key("t")))
-        .unwrap_or(false)
+    serde_json::from_str::<BTreeMap<String, serde::de::IgnoredAny>>(first_line)
+        .is_ok_and(|object| object.contains_key("t"))
 }
 
 /// Parse a streamed group export into the same [`AgentStateExport`] the rest of
--- a/crates/incident-replay/src/ndjson.rs
+++ b/crates/incident-replay/src/ndjson.rs
@@ -35,18 +35,15 @@ const EVENT: &str = "event";
 /// terminal `{"t":"error","complete":false}` line with no `eof`.
 const ERROR: &str = "error";
 
-/// Whether `input` is a Goggles NDJSON group-export stream. The v1 contract
-/// requires the first line to be the manifest, so this checks exactly that; an
-/// `agent-state.json` document (one JSON object, no `t` discriminator) is never
-/// mistaken for a stream.
+/// Whether `input` is a Goggles NDJSON group-export stream. A stream line is
+/// discriminated by a `t` field; checking whether the first non-empty line
+/// carries a `t` discriminator prevents a manifest-less or corrupt stream from
+/// falling back to the document parser and fail-opening as a healthy export.
 pub fn is_stream(input: &str) -> bool {
     let Some(first_line) = input.lines().find(|line| !line.trim().is_empty()) else {
         return false;
     };
-    matches!(
-        serde_json::from_str::<TagProbe>(first_line),
-        Ok(TagProbe { t }) if t == MANIFEST
-    )
+    serde_json::from_str::<TagProbe>(first_line).is_ok()
 }
 
 /// Parse a streamed group export into the same [`AgentStateExport`] the rest of
@@ -77,6 +74,9 @@ pub fn parse_stream(input: &str) -> Result<AgentStateExport, StreamParseError> {
             })?;
 
         if !saw_manifest {
+            if t == ERROR {
+                return Err(StreamParseError::ServerReportedError { line: number });
+            }
             if t != MANIFEST {
                 return Err(StreamParseError::MissingManifest);
             }
--- a/crates/incident-replay/src/ndjson.rs
+++ b/crates/incident-replay/src/ndjson.rs
@@ -43,7 +43,10 @@ pub fn is_stream(input: &str) -> bool {
     let Some(first_line) = input.lines().find(|line| !line.trim().is_empty()) else {
         return false;
     };
-    serde_json::from_str::<TagProbe>(first_line).is_ok()
+    serde_json::from_str::<serde_json::Value>(first_line)
+        .ok()
+        .and_then(|value| value.as_object().map(|object| object.contains_key("t")))
+        .unwrap_or(false)
 }
 
 /// Parse a streamed group export into the same [`AgentStateExport`] the rest of
--- a/crates/incident-replay/tests/cli.rs
+++ b/crates/incident-replay/tests/cli.rs
@@ -1,0 +1,34 @@
+//! End-to-end CLI coverage for fail-closed format detection.
+
+use std::io::Write;
+use std::process::{Command, Stdio};
+
+#[test]
+fn malformed_stream_discriminators_never_classify_as_healthy() {
+    for input in [
+        r#"{"t": null}"#,
+        r#"{"t": 0}"#,
+        r#"{"t": {}}"#,
+        r#"{"t": []}"#,
+        r#"{"t": "error", "t": null}"#,
+    ] {
+        let mut child = Command::new(env!("CARGO_BIN_EXE_incident-replay"))
+            .arg("/dev/stdin")
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .spawn()
+            .expect("launch incident-replay");
+        child
+            .stdin
+            .take()
+            .expect("capture stdin")
+            .write_all(input.as_bytes())
+            .expect("write export");
+        let output = child.wait_with_output().expect("wait for incident-replay");
+        let stdout = String::from_utf8(output.stdout).expect("stdout is utf-8");
+
+        assert_eq!(output.status.code(), Some(2), "input: {input}");
+        assert!(!stdout.contains("healthy:"), "input: {input}");
+    }
+}
--- a/crates/incident-replay/tests/cli.rs
+++ b/crates/incident-replay/tests/cli.rs
@@ -1,34 +1,22 @@
 //! End-to-end CLI coverage for fail-closed format detection.
 
 use std::io::Write;
-use std::process::{Command, Stdio};
+use std::process::Command;
+
+use tempfile::NamedTempFile;
 
 #[test]
-fn malformed_stream_discriminators_never_classify_as_healthy() {
-    for input in [
-        r#"{"t": null}"#,
-        r#"{"t": 0}"#,
-        r#"{"t": {}}"#,
-        r#"{"t": []}"#,
-        r#"{"t": "error", "t": null}"#,
-    ] {
-        let mut child = Command::new(env!("CARGO_BIN_EXE_incident-replay"))
-            .arg("/dev/stdin")
-            .stdin(Stdio::piped())
-            .stdout(Stdio::piped())
-            .stderr(Stdio::piped())
-            .spawn()
-            .expect("launch incident-replay");
-        child
-            .stdin
-            .take()
-            .expect("capture stdin")
-            .write_all(input.as_bytes())
-            .expect("write export");
-        let output = child.wait_with_output().expect("wait for incident-replay");
-        let stdout = String::from_utf8(output.stdout).expect("stdout is utf-8");
+fn a_manifest_less_error_remnant_never_classifies_as_healthy() {
+    let mut input = NamedTempFile::new().expect("create export");
+    input
+        .write_all(br#"{"t":"error","complete":false}"#)
+        .expect("write export");
+    let output = Command::new(env!("CARGO_BIN_EXE_incident-replay"))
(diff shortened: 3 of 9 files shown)


crates/incident-replay/src/ndjson.rs
```diff
@@ -35,18 +35,15 @@ const EVENT: &str = "event";
...
 pub fn is_stream(input: &str) -> bool {
     let Some(first_line) = input.lines().find(|line| !line.trim().is_empty()) else {
         return false;
     };
-    matches!(
-        serde_json::from_str::<TagProbe>(first_line),
-        Ok(TagProbe { t }) if t == MANIFEST
-    )
+    serde_json::from_str::<TagProbe>(first_line).is_ok()
 }
```
erskingardner, line 49: This doc comment above `is_stream` correctly describes the new "any `t` discriminator" rule, but two other spots in this crate still describe the old manifest-first contract this PR just changed:

- `crates/incident-replay/src/main.rs:4-6` (module doc): "a stream leads with its `manifest` line ... anything else is parsed as `agent-state.json`" — no longer accurate; a manifest-less stream (e.g. the `{"t":"error","complete":false}` remnant this PR fixes) is now also routed to `parse_stream`.
- `crates/incident-replay/AGENTS.md:25` ("`is_stream` detects the format from the contract's manifest-first rule") — same drift; this is the crate map CLAUDE.md relies on to keep docs authoritative.

Non-blocking, but worth a one-line touch-up alongside this change since both describe the exact logic being altered here.

crates/incident-replay/src/export.rs
```diff
@@ -247,6 +247,15 @@ impl EventKind {
...
 pub fn parse(json: &str) -> Result<AgentStateExport, ParseError> {
+    #[derive(Deserialize)]
+    struct StreamDiscriminatorProbe {
+        t: Option<String>,
+    }
+    if let Ok(StreamDiscriminatorProbe { t: Some(t) }) =
+        serde_json::from_str::<StreamDiscriminatorProbe>(json)
+    {
+        return Err(ParseError::StreamDiscriminator { t });
+    }
```
erskingardner, line 438: Minor, non-blocking: this parses `json` twice now — once as `StreamDiscriminatorProbe` and again (below) as the full `AgentStateExport` — on every call, including the common non-stream path. Given the crate rejects inputs over 64 MiB (`MAX_INCIDENT_EXPORT_BYTES` in `main.rs`) before parsing, a full double-parse is bounded but avoidable, e.g. by parsing once into a `serde_json::Value`, checking `t` on that, then `serde_json::from_value` into `AgentStateExport`. Not worth blocking on for a forensic CLI that runs once per export, just flagging since "fewest lines / least redundant work" is otherwise the norm here.

erskingardner (dismissed):
Reviewed against the linked root-cause writeup (#1037) and verified locally: cloned the PR branch and ran `cargo test -p incident-replay` (39 tests pass, including the new regressions), `cargo fmt -p incident-replay -- --check`, and `cargo clippy -p incident-replay --all-targets` — all clean.

The fix matches the issue's suggested approach exactly and is appropriately surgical: `is_stream` now fails closed on any `t`-discriminated first line instead of requiring `manifest` specifically, and `export::parse` gets a defense-in-depth rejection of top-level `t`-discriminated objects for callers that bypass `is_stream`. Both new regression tests (`a_manifest_less_error_remnant_is_detected_as_a_stream_and_rejected`, `stream_shaped_lines_are_rejected_as_documents`) directly reproduce the exact failure scenario from the issue (`{"t":"error","complete":false}` reading as a healthy 0-vector export) and now correctly fail closed with `ServerReportedError`/a parse error instead.

Two non-blocking notes left inline:
- The module doc in `main.rs` and the crate's `AGENTS.md` still describe the old manifest-first detection rule this PR just replaced — worth a one-line touch-up so the crate map stays accurate.
- `export::parse` now parses its input twice (once for the discriminator probe, once for the full struct); bounded by the existing 64 MiB cap so not a real concern, but a `serde_json::Value`-based single-parse would be marginally cleaner.

No blocking issues found.

Datawav:

- **Medium — Malformed `t` discriminators still follow the fail-open path** (`crates/incident-replay/src/ndjson.rs:42`, `crates/incident-replay/src/export.rs:259`). `is_stream` recognizes only a string-valued `t`, while the document-parser guard rejects only `t: Some(String)`. Inputs such as `{"t":null}`, `{"t":123}`, or an object with duplicate `t` keys make those probes return `None` or an error, after which the permissive `AgentStateExport` deserializer ignores `t` and can classify the input as a healthy empty export. This leaves the same unsafe outcome the PR is meant to eliminate. Detect the presence of the top-level key using a raw JSON object independently of its value/type, reject it before document deserialization, and add null, non-string, and duplicate-key regressions.

The fix is needed because truncated or corrupt NDJSON must never be adopted as a healthy document. Routing discriminator-bearing input to the strict stream parser, with a defense-in-depth document check, is an appropriately small design. Making document fields required would unnecessarily break the parser’s intended forward-compatible leniency; matching only known stream-tag values would also weaken fail-closed handling for corrupt or future stream records. The advisory concern about legitimate documents gaining a string `t` is speculative and conflicts with the patch’s explicit treatment of top-level `t` as the cross-format discriminator.

_Advisory audit of head [`a9a941945b22`](https://github.com/marmot-protocol/mdk/commit/a9a941945b22227dc5e1b39877e614309ac3d112); no approval, merge, push, checkout, or PR-code execution was performed._

Datawav:
## Nightly deep audit

- **Medium — malformed `t` discriminators still fail open as a healthy export** (`crates/incident-replay/src/ndjson.rs:47`, `crates/incident-replay/src/export.rs:253`). The new detector only recognizes `t` when it deserializes as `String`; the document-parser guard similarly rejects only `Some(String)`. Consequently, valid JSON such as `{"t":null,"complete":false}` or `{"t":0}` bypasses both defenses. Duplicate `t` fields also make the struct probe fail, after which `AgentStateExport` ignores those unknown fields. With its defaulted `events` and `derived_projections`, the CLI then classifies the input as `healthy: 0 vectors`—the same unsafe dataflow the PR is intended to eliminate. This matters at the trust boundary: corrupted, schema-drifted, or adversarial stream remnants can suppress incident evidence and produce a successful healthy result. Detect top-level key presence independently of its value type, for example by parsing to `serde_json::Value` and using `as_object().is_some_and(|o| o.contains_key("t"))`, in both routing and direct document parsing. Alternatively, make the document schema reject unknown top-level fields, but that conflicts with its deliberate forward-compatible design. Add end-to-end CLI or routing tests for `t: null`, numeric/object/array `t`, and duplicate `t` keys, asserting a parse failure and never a healthy verdict. The broader concern about rejecting a future legitimate document-level string `t` is not actionable without evidence that `t` is allowed in that schema; the fail-closed requirement favors reserving it. No additional concurrency, fork/convergence, durable-state, authorization, or persistence regression is introduced by this patch.

erskingardner:
Maintainer follow-up: rebased the contributor change onto current `master` and addressed the remaining review/audit feedback.

- discriminator detection now checks top-level `t` key presence independently of its value/type;
- the direct document parser uses the same fail-closed rule without double-parsing;
- null, numeric, object, array, and duplicate-key regressions cover both parser selection and the executable CLI path;
- parser tests assert `ParseError::StreamDiscriminator`;
- stale format-detection documentation is updated.

Local validation is green: `cargo test -p incident-replay`, focused all-target clippy, and `just fast-ci`. GitHub CI on the updated head is still authoritative.

erskingardner (dismissed):
## Review

Recommend merge. This correctly closes the #1037 fail-open at format detection.

- `is_stream` now keys off top-level `t` presence (value/type-independent), so a manifest-less remnant like `{"t":"error","complete":false}` routes into the fail-closed stream parser instead of the all-default document model.
- `export::parse` is a real second line of defense with the same key-presence rule and a single `Value` pass (no double-parse).
- Pre-manifest `error` lines surface as `ServerReportedError` rather than the vaguer `MissingManifest`.
- Regressions cover the original remnant, malformed/null/non-string/duplicate `t`, document-parser rejection via `ParseError::StreamDiscriminator`, and the CLI exit-2 path.

Prior review/audit notes (docs drift, double-parse, malformed discriminator fail-open) are addressed on `4b20861`. CI is green. No blocking issues.

crates/incident-replay/src/export.rs
```diff
@@ -432,12 +432,21 @@ impl EventKind {
...
 pub fn parse(json: &str) -> Result<AgentStateExport, ParseError> {
-    Ok(serde_json::from_str(json)?)
+    let value: serde_json::Value = serde_json::from_str(json)?;
+    if value
+        .as_object()
+        .is_some_and(|object| object.contains_key("t"))
+    {
+        return Err(ParseError::StreamDiscriminator);
+    }
+    Ok(serde_json::from_value(value)?)
```
erskingardner: This parses the document twice and materializes the whole thing as a `Value` tree in between — on a 64 MiB export (the crate's own `MAX_INCIDENT_EXPORT_BYTES`) the tree is several times the source size and is alive at the same time as the export being built from it. It also loses the `line N column M` suffix on `ParseError::Json`, since `from_value` errors have no position.

A key-presence probe that discards values does the same job with no tree:

```rust
pub fn parse(json: &str) -> Result<AgentStateExport, ParseError> {
    let probe = serde_json::from_str::<BTreeMap<String, serde::de::IgnoredAny>>(json);
    if probe.is_ok_and(|object| object.contains_key("t")) {
        return Err(ParseError::StreamDiscriminator);
    }
    Ok(serde_json::from_str(json)?)
}
```

I checked this against every input in the new `parse.rs` test (`{"t": null}`, `{"t": 0}`, `{"t": {}}`, `{"t": []}`, the duplicate-key case, and the two well-formed stream lines): identical results, and `{"events": []}` / non-objects still return `false`. `IgnoredAny` as a map *value* accepts `null` and tolerates duplicate keys (last write wins), which is why a `#[derive(Deserialize)]` probe struct would not work here.

A malformed document still reports the schema error from the second `from_str` with full position info.

crates/incident-replay/src/ndjson.rs
```diff
@@ -35,18 +35,18 @@ const EVENT: &str = "event";
...
         return false;
     };
-    matches!(
-        serde_json::from_str::<TagProbe>(first_line),
-        Ok(TagProbe { t }) if t == MANIFEST
-    )
+    serde_json::from_str::<serde_json::Value>(first_line)
+        .ok()
+        .and_then(|value| value.as_object().map(|object| object.contains_key("t")))
+        .unwrap_or(false)
```
erskingardner: Same point as `export::parse`, plus a readability one. The old `TagProbe` path never allocated a tree (unknown fields go to `IgnoredAny`); this one builds a full `Value` for the first line, which for a single-line minified export is the entire document.

```rust
    serde_json::from_str::<BTreeMap<String, serde::de::IgnoredAny>>(first_line)
        .is_ok_and(|object| object.contains_key("t"))
```

Same semantics on all five malformed inputs in the new test, no tree, and one expression instead of the `.ok().and_then(...).unwrap_or(false)` chain. If you keep `Value`, at least collapse the chain to `.is_ok_and(|value| value.as_object().is_some_and(|object| object.contains_key("t")))` — the `Option<bool>` round-trip in the middle is hard to read for what is a two-predicate test.

crates/incident-replay/tests/cli.rs
```diff
@@ -0,0 +1,34 @@
...
+fn malformed_stream_discriminators_never_classify_as_healthy() {
+    for input in [
+        r#"{"t": null}"#,
+        r#"{"t": 0}"#,
+        r#"{"t": {}}"#,
+        r#"{"t": []}"#,
+        r#"{"t": "error", "t": null}"#,
+    ] {
+        let mut child = Command::new(env!("CARGO_BIN_EXE_incident-replay"))
+            .arg("/dev/stdin")
```
erskingardner: Two things about this file:

1. **Triplication.** These exact five inputs are already asserted in `tests/ndjson.rs` (`is_stream` + `parse_stream` reject) and `tests/parse.rs` (`ParseError::StreamDiscriminator`). Spawning five subprocesses to re-derive the same conclusion buys one genuinely new fact — that the CLI exits 2 rather than printing `healthy:` — which one input demonstrates. I'd keep a single case here (ideally `{"t":"error","complete":false}`, the actual remnant from #1037, since that is the shape the CLI regression is about) and drop the loop.

2. **`/dev/stdin` as the path argument.** This is the only place in the workspace that does this, and it depends on the child inheriting a readable `/dev/stdin` — fine on the ubuntu/macos runners, but it is an odd indirection for "give the binary a file". `tempfile` is already a workspace dependency used by ten crates; a `NamedTempFile` (or a path under `std::env::temp_dir()`) says what it means and won't surprise anyone running the suite in a less usual environment.

erskingardner (commented):
## Review

The routing fix is correct and it closes the exact fail-open reported in #1037: a manifest-less remnant like `{"t":"error","complete":false}` now reaches `parse_stream` and fails closed, and detection is key-presence based so `null`/numeric/object/array/duplicate `t` values can't slip past. Tests are well-targeted and the AGENTS.md/module docs were updated to match. Two things I'd like addressed before merge, plus one scope question.

**1. Both new checks materialize a full `serde_json::Value` tree just to test for one top-level key.** The maintainer follow-up comment says the document parser now applies the rule "without double-parsing", but `export::parse` at head does exactly that: `from_str::<Value>` then `from_value`. Combined with the new `is_stream`, a single-line minified export (the input the crate caps at 64 MiB) is now walked into a `Value` tree twice on the healthy path before it is deserialized a third time. It also downgrades `ParseError::Json` diagnostics, because `from_value` errors carry no line/column. A `BTreeMap<String, serde::de::IgnoredAny>` probe has *identical* semantics on every input in the new tests (I verified all seven, including `{"t": null}` and the duplicate-key case) with no tree allocation — details inline.

**2. The same five malformed inputs are now asserted in three places** (`tests/ndjson.rs`, `tests/parse.rs`, `tests/cli.rs`), and the CLI test spawns five subprocesses through `/dev/stdin`. One end-to-end case is enough; inline.

**3. Scope question — the fix closes the reported instance, not the class.** The root cause named in #1037 is that `AgentStateExport` has all-`#[serde(default)]` fields and no required marker, so *any* single JSON object becomes an empty export that classifies `Healthy` with exit 0. After this PR that is still true for any object without a `t` key — `{}`, a stray `package.json`, or a truncated remnant whose first line happens to lack the discriminator all still print `healthy: 0 vectors`. Sniffing for `t` is the symptom-level half of the issue's suggestion. I'm not asking to expand this PR (surgical is right), but a follow-up that requires a parsed document to carry at least one recognized top-level key (`events` / `derived_projections` / `normalized_scenario_history`) would close the class in about as many lines as the `t` guard, and would make the `t` guard in `export::parse` redundant — worth deciding now, since `parse` has no caller outside `main.rs`, which `is_stream` already guards.

Nothing here is a correctness blocker; #1 is a straightforward regression against the PR's own stated intent.

crates/incident-replay/src/export.rs
erskingardner: Fixed in f09eba09: the document guard now uses a `BTreeMap<String, serde::de::IgnoredAny>` probe, then deserializes the real model directly from the source string. That avoids the full `Value` tree and preserves line/column diagnostics; `invalid_document_errors_keep_the_source_position` locks the latter behavior.

crates/incident-replay/src/ndjson.rs
erskingardner: Fixed in f09eba09: `is_stream` now uses the same allocation-light `BTreeMap<String, IgnoredAny>` key-presence probe and the shorter `is_ok_and` expression.

crates/incident-replay/tests/cli.rs
erskingardner: Fixed in f09eba09: the CLI test now exercises only the real `{"t":"error","complete":false}` remnant and passes it through a `NamedTempFile`; the malformed-value matrix remains at the two parser boundaries where those individual cases matter.

erskingardner:
Scope decision from the latest review: keep #1140 focused on the concrete NDJSON format-detection fail-open from #1037. The broader class is valid — an unrecognized document object without `t` can still default to an empty healthy export — but defining positive document markers needs confirmation against the Goggles sparse-export contract. Tracked separately as #1373 so it is not lost or silently coupled to this patch.

PR #1141, opened 2026-07-27
Complete the MarmotKit chat-list interaction contract

erskingardner:
## Summary

Completes the no-wire MarmotKit chat-list interaction contract.

- add durable manual unread state without weakening the monotonic read marker
- add timed and indefinite conversation muting with expiry-driven subscription refreshes
- expose inferred direct/group/unknown conversation kind from durable membership projections
- expose bounded attachment preview kind/count using the authoritative media parser
- expose pending, delivered, failed, and not-applicable delivery state for the exact latest preview
- add runtime commands, subscription triggers, UniFFI DTOs, and generated Swift/Kotlin surfaces
- add migration 0038 and focused storage, runtime, media, CLI-contract, and FFI coverage

The immutable creation-time DM identity is intentionally excluded because it requires a wire-contract change. This PR also contains no UIKit or SwiftUI design changes. The workspace remains at 0.9.7; the 0.9.8 version bump is reserved for the dedicated release preparation.

## Impact

App clients can render and mutate the complete chat-list interaction state without N+1 lookups, parsing raw media metadata, or conflating manual unread state with the durable read marker. Mute state and local delivery failures survive restart, and projections rematerialize when the latest message is replaced, deleted, retried, or pruned.

## Validation

- `just fast-ci`
- `cargo test -p storage-sqlite` (286 passed)
- `wn-cli` library tests (427 passed)
- `cargo test -p cgka-conformance-simulator canonical_vector_fixtures_match_generated_traces`
- `cargo test -p marmot-app --lib` (457 passed)
- `cargo test -p marmot-uniffi --lib` (75 passed)
- Swift XCFramework generation
- Kotlin/JNI generation for arm64-v8a, armeabi-v7a, x86, and x86_64

The broad `relay_runtime` integration suite retains five unrelated failures. The representative encrypted-media endpoint replacement failure was reproduced unchanged on the exact baseline commit.


erskingardner (commented):
Reviewed the diff (`gh pr diff 1141`) against AGENTS.md/CLAUDE.md and this repo's version-cadence convention. The chat-list interaction contract implementation itself is solid — I traced the SQL column-index shifts through all three `chat_list_rows`/`chat_list_row` query sites and the `rebuild_chat_list_row_for_group_tx` INSERT param list by hand (29 params vs. 29 placeholders, `retained_activity_sort_at` correctly left as a literal `0`), confirmed the "sent + local_publish_failed" latest-message widening is correctly scoped to the preview/activity queries and does **not** leak into `unread_summary_tx` (which still requires `invalidation_status IS NULL`), and confirmed `member_count` is kept live via the existing `add_group`/`refresh_from_group` event pipeline rather than only at open time. Test coverage (manual-unread independence, mute-expiry join, conversation-kind classification, delivery-state transitions incl. retry-after-failure) looks thorough.

One process concern worth resolving before merge, plus a minor non-blocking nit — see inline comments.

Overall: solid, well-tested feature work; flagging the version bump for a maintainer decision rather than as a functional defect.

Cargo.toml
```diff
@@ -26,7 +26,7 @@ members = [
 
 [workspace.package]
 edition = "2024"
-version = "0.9.7"
+version = "0.9.8"
```
erskingardner: This PR bumps the workspace version directly (0.9.7 -> 0.9.8) inside a feature PR. Repo history shows version bumps only ever land in dedicated `chore(release): prepare X` commits (e.g. a4374b78, f9d32a4e, 1d6b9f1d) — feature PRs leave `Cargo.toml` alone and only append to each crate's CHANGELOG `## [Unreleased]` section. Here the CHANGELOG entries added by this PR (e.g. `crates/cli/CHANGELOG.md`) are still under `## [Unreleased]`, not restamped to 0.9.8, so the version bump looks premature/out of place for a feature PR — it also causes Cargo.lock churn across every workspace crate. Suggest reverting the `Cargo.toml`/`Cargo.lock` version bump here and doing it in a separate release commit via `just release-all 0.9.8`, unless this PR is intentionally meant to double as the release cut (in which case the CHANGELOG heading should read `## [0.9.8]` instead).

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -995,36 +1257,70 @@ fn read_state_tx(
...
+                row.avatar_media_type, row.last_message_id_hex,
+                row.last_message_sender, row.last_message_preview,
+                row.last_message_kind, row.last_message_timeline_at,
+                row.last_message_deleted, row.last_message_media_json,
+                row.last_message_delivery_state, row.unread_count,
+                row.manually_marked_unread, row.unread_mention_count,
+                row.first_unread_message_id_hex, row.last_read_message_id_hex,
+                row.last_read_timeline_at, row.conversation_created_at,
+                row.activity_sort_at, row.updated_at, row.self_membership,
+                ag.member_count,
```
erskingardner, line 1274: Minor/non-blocking: this "is mute currently effective" logic (`group_id_hex IS NULL` -> unmuted, `muted_until_ms IS NULL` -> muted forever, `muted_until_ms > now` -> muted) is duplicated three times here (this query, the archived variant, and `chat_list_row_tx`) and again independently in Rust as `chat_notification_settings_at` in `account_projection.rs` (`muted_until_ms.is_none_or(|until| until > now_ms)`). The two implementations happen to agree today, but there's no shared source of truth, so a future edge-case fix (e.g. an off-by-one at the exact expiry boundary) could silently drift between the chat-list join and the single-chat read path. Consider extracting the boundary check into one place (a SQL view/CTE, or computing muted client-side from `chat_notification_settings_at` for both call sites) if this needs to change again.

PR #1143, opened 2026-07-27
fix(openclaw): route replies through the source channel

### LINKED ISSUE #1125: OpenClaw adapter: make first-message replies context-bound and failure-proof
## Problem

The OpenClaw Marmot adapter has a context-bound reply sink for normal final replies, but an agent can also call OpenClaw's shared `message(action: "send")` tool during an inbound Marmot turn.

In a real first-message test, the inbound message arrived correctly, but an untargeted explicit send bypassed the bound reply path. A later fallback passed a session/decorated target to the connector and failed with:

```text
group_id_hex must be hexadecimal
```

The connector should make the safe path automatic. An agent replying inside an inbound Marmot turn should not need to know, copy, parse, or reconstruct the MLS group ID.

## Desired behavior

Treat first-message delivery as an end-to-end contract:

> After a fresh install/bootstrap and invite, the first permitted inbound message starts one agent turn and produces exactly one durable reply in the same Marmot group.

This must work whether the agent returns a normal final response or uses the shared `message` tool without an explicit target.

### 1. Inherit the inbound Marmot route for untargeted sends

When `message(action: "send")` is called during a Marmot-triggered turn and `to` is omitted, resolve the destination from the current turn's immutable inbound route:

- channel: `marmot`
- Marmot account ID
- MLS group ID
- reply-to message ID, when appropriate

The context must be turn-scoped. Never fall back to a process-global or "last seen" group.

Explicit valid targets should continue to work and take precedence.

### 2. Canonicalize and validate at the adapter boundary

Before calling `wn-agent`:

- normalize supported forms to one canonical bare lowercase group-ID hex value;
- reject OpenClaw session keys, route strings, malformed/decorated targets, and cross-channel targets before they reach `wn-agent`;
- return an actionable error that identifies the received target form and the expected form, without leaking sensitive content.

The low-level `group_id_hex must be hexadecimal` error should not be the first diagnostic an OpenClaw user sees.

### 3. Gate first dispatch on reply readiness

Do not dispatch the first inbound turn until:

- the selected account is ready;
- the inbound subscription is active;
- the invited group is locally available; and
- the same account/group route can be used for a durable send.

If group state is still converging immediately after an invite, use a short bounded retry/backoff rather than starting a turn whose reply cannot yet be delivered.

### 4. Preserve exactly-once durable delivery

Any retry or fallback must reuse the same logical delivery/idempotency identity so it cannot create duplicate kind-9 messages.

Do not report a successful agent turn as delivered until `send_final` or `stream_finalize` has durably committed the reply. If delivery fails, surface one clear failure tied to the account/group route.

### 5. Add structured diagnostics

Log enough non-secret routing state to diagnose failures:

- inbound route source;
- whether the target was inherited or explicit;
- target validation result;
- account/group readiness state;
- durable send result and message ID(s).

Group IDs should be redacted or shortened in normal logs.

## Acceptance criteria

- [ ] Fresh `wn-agent` home, existing or newly bootstrapped account, and OpenClaw Marmot plugin start cleanly.
- [ ] A permitted user invites the agent and sends the first message immediately.
- [ ] The first inbound message starts exactly one turn.
- [ ] A normal final response produces exactly one durable reply in the same group.
- [ ] `message(action: "send")` with no `to` during that turn inherits the same Marmot account/group and produces exactly one durable reply.
- [ ] An explicit bare group ID and `marmot:<group-id>` both normalize and send successfully.
- [ ] A session key, decorated route, malformed hex, or cross-channel target is rejected locally with an actionable error and never reaches `wn-agent`.
- [ ] A transient not-ready group path retries within a bounded window without duplicating the reply.
- [ ] The behavior is covered by an end-to-end regression test using a fresh invite and first message, not only unit tests of target parsing.
- [ ] Multi-account tests prove that inherited routing cannot cross accounts or groups.
- [ ] Restart/reconnect during the first-message window either completes the same logical reply once or reports a clear failure; it never silently drops or duplicates it.

## Suggested test matrix

1. Fresh identity / fresh group / immediate first message.
2. Existing identity / fresh group / immediate first message.
3. Normal final response vs. untargeted shared-message-tool send.
4. Explicit canonical target vs. supported prefixed target.
5. Invalid session/decorated target.
6. Two configured Marmot accounts receiving concurrently.
7. Connector reconnect between inbound dispatch and durable reply.

This is primarily an OpenClaw adapter/integration hardening issue, but the end-to-end test should include `wn-agent` so the first-message contract is verified across the actual boundary.

erskingardner:
## What changed

- make OpenClaw's `gateway.startAccount` own exactly one long-lived Marmot inbound subscription per configured account
- build the trusted inbound context from the received Marmot group and route normal assistant finals through `deliverInboundReplyWithMessageSendContext`
- let OpenClaw invoke the registered Marmot message adapter, which maps the trusted destination to `wn-agent send_final`
- deny `sessions_send` during source-bound Marmot turns because it targets another agent session, not the originating external channel
- disable inbound reply streaming for now so there is one durable delivery owner
- remove the obsolete custom reply/preview sink and fire-and-forget `registerFull` startup path

## Why

The previous connector had two competing durable reply paths. Normal inbound turns could send directly to `wn-agent`, while OpenClaw's shared message path used the registered channel adapter. A model could also choose `sessions_send`, which does not reply to the human-facing Marmot channel. That made destination ownership ambiguous and caused received messages to produce no reply on their source group.

This change follows OpenClaw's standard channel contract: the inbound context owns `OriginatingTo`/`To`, and a normal final goes back through the registered channel adapter to that same Marmot group. The model never reconstructs or selects the group id.

## User impact

A message received from a Marmot group now produces its automatic final on that same group and threads it to the triggering message. Sequential messages and a restarted inbound subscription preserve the same routing contract. QUIC preview primitives remain available for a future standard live-adapter integration, but inbound turns are final-only today.

## Validation

- `pnpm typecheck`
- `pnpm build`
- `pnpm test` — 187 passed, 1 skipped
- `MARMOT_OPENCLAW_CONNECTOR_E2E=1 pnpm exec vitest run test/e2e-connector.test.ts --reporter=verbose` — real `wn-agent`, sequential replies plus subscription restart
- `just fast-ci`

Supersedes #1137.

Fixes #1125.

### DIFF
--- a/.github/workflows/wn-agent-binaries.yml
+++ b/.github/workflows/wn-agent-binaries.yml
@@ -388,7 +388,13 @@ jobs:
           # Build dist/ before packing: package.json#files ships dist (which is
           # gitignored), and the gateway requires compiled runtime output rather
           # than the TypeScript sources.
-          ( cd integrations/openclaw/marmot && pnpm install --frozen-lockfile && pnpm build )
+          (
+            cd integrations/openclaw/marmot
+            pnpm install --frozen-lockfile
+            pnpm typecheck
+            pnpm test
+            pnpm build
+          )
           # Provenance manifest (release.md requires one in each artifact); ships
           # via package.json#files alongside the built plugin.
           cat > integrations/openclaw/marmot/manifest.json <<EOF
--- a/integrations/openclaw/marmot/AGENTS.md
+++ b/integrations/openclaw/marmot/AGENTS.md
@@ -21,7 +21,8 @@ The OpenClaw counterpart of `integrations/hermes/marmot`. Read `README.md` first
 - `src/append-only.ts` — append-only suffix tracker for progressive updates.
 - `src/live.ts` — live-preview state machine → `stream_begin`/`append`/`finalize`/`cancel`.
 - `src/inbound.ts` — inbound subscription bridge (reconnect, dedupe, resync).
-- `src/inbound-runtime.ts` — `registerFull` wiring + the inbound→agent dispatch seam.
+- `src/gateway.ts` — OpenClaw-owned account lifecycle and inbound subscription.
+- `src/inbound-runtime.ts` — the inbound→agent dispatch seam.
 - `src/bounded-keyed-async-queue.ts` — per-group inbound dispatch with a depth cap.
 - `src/outbound.ts` — `defineChannelMessageAdapter` durable send → `send_final`.
 - `src/messaging.ts` — `messaging` target-resolution adapter so the shared `message` tool can resolve a Marmot conversation (group id hex).
@@ -37,7 +38,9 @@ The OpenClaw counterpart of `integrations/hermes/marmot`. Read `README.md` first
 - Regenerate `test/vectors/transcript-vectors.json` from the Rust
   `AgentTextStreamTranscriptV1` if the Rust transcript hashing ever changes.
 - Keep the `openclaw` dependency pinned; before bumping, verify the
-  `openclaw/plugin-sdk/*` subpath exports against the new version's types.
+  `openclaw/plugin-sdk/*` subpath exports against the new version's types, and
+  re-verify the deliberate test-only `dist/plugins/loader.js` import used by the
+  connector E2E (the pinned SDK exposes no supported plugin-loader subpath).
 - The inbound→agent and live-preview-pipeline seams use OpenClaw gateway runtime
   internals and are validated by the deterministic connector E2E plus the docker
   phone test, not the unit tests alone.
--- a/integrations/openclaw/marmot/README.md
+++ b/integrations/openclaw/marmot/README.md
@@ -206,6 +206,9 @@ advanced shared deployment can point both gateways at one `wn-agent`:
 | `profileNameOnboarding` | `MARMOT_PROFILE_NAME_ONBOARDING` | `true` |
 | `dm.policy` / `dm.allowFrom` | — | `allowlist` |
 
+The QUIC/streaming settings are still accepted for configuration compatibility,
+but inbound agent replies are currently delivered final-only.
+
 `accountIdHex` is the Marmot/wn-agent account id. It is intentionally distinct
 from OpenClaw's channel account id (`default`, or a key under
 `channels.marmot.accounts`) used for routing, session metadata, and message-tool
@@ -220,11 +223,14 @@ The token grants the full connector API for every hosted account, not only the
 configured OpenClaw channel account. Use a separate connector home/socket/token
 for any plugin or tenant that is not in the same trust boundary.
 
-- **Inbound → agent turn** (`src/dispatch.ts`): the inbound bridge feeds each
-  received Marmot message (`chatId` = Marmot group id, `userId` = sender) into
-  OpenClaw's turn kernel via `runChannelInboundEvent` + `dispatchReplyWithBufferedBlockDispatcher`,
-  **modeled on the bundled Telegram channel**. The agent's reply is delivered
-  back through the message adapter and threads to the triggering message.
+- **Inbound → agent turn** (`src/dispatch.ts`): the gateway-owned inbound bridge
+  feeds each received Marmot message (`chatId` = Marmot group id, `userId` =
+  sender) into OpenClaw's turn kernel via `runChannelInboundEvent` +
+  `dispatchReplyWithBufferedBlockDispatcher`. The trusted inbound context owns
+  the destination. A normal assistant final is passed to
+  `deliverInboundReplyWithMessageSendContext`, which invokes the registered
+  Marmot message adapter and threads the reply to the triggering message. The
+  model never has to reconstruct or select the source group.
   Dispatch is serialized per group (distinct groups run concurrently, each group
   stays FIFO), so a slow turn in one group never blocks inbound for others; set
   `debounceMs` to coalesce rapid same-sender bursts into one turn.
@@ -245,16 +251,11 @@ for any plugin or tenant that is not in the same trust boundary.
   unrecallable barge-in there is worse than dropping a single reply in a true
   two-party DM (where the user can re-send or address the agent explicitly). The
   error is not cached, so the next message retries the lookup.
-- **Durable replies** are sent verbatim as `kind: 9` messages via `send_final`;
-  the adapter never merges or rewrites text across sends. Each durable reply is
-  **idempotent + retried**: the sink generates one `idempotency_key` per reply
-  and retries the send a few times (with a short backoff) on retryable errors,
-  reusing the same key so `wn-agent` dedups instead of double-posting. A repeated
-  key returns the original message ids without a second send, so a retry after a
-  post-write timeout cannot double-post an unrecallable encrypted message.
-  Live-preview `stream_finalize` uses the same posture: one idempotency key per
-  preview stream, bounded retry on retryable failures, and connector-side cached
-  final message ids for post-success retries.
+- **Durable replies** are sent verbatim as `kind: 9` messages via the registered
+  message adapter's `send.text` → `wn-agent send_final` mapping. The adapter
+  never merges or rewrites text across sends. A bounded retry reuses one
+  per-send idempotency key so a transient control-socket failure does not create
+  a second Marmot message.
 - **`message`-tool target resolution** (`src/messaging.ts`): a Marmot reply is
   delivered automatically from the assistant's final text, so the agent does not
   need the shared `message` tool to answer. When it *does* call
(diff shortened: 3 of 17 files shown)


erskingardner:
Reviewed at `00c642b1`, fresh rather than as a diff from #1137.

The restructure is the right call, and it's endorsed by the SDK itself. The pinned `openclaw@2026.6.11` marks the old shape deprecated in as many words:

> New channel plugins should expose a `message` adapter via `defineChannelMessageAdapter(...)` and route sends through `deliverInboundReplyWithMessageSendContext(...)`
> — `inbound-reply-dispatch-DIfEV7g-.js`, on `dispatchInboundReplyWithBase` / `dispatchChannelMessageReplyWithBase`

That is exactly what this PR moves to. Dropping the bespoke reply sink, the turn-ownership coordinator, and the turn idempotency layer in favour of one route built from the inbound group is a large, genuine simplification — 1140 lines deleted, `dispatch.ts` down to 403, and the "two competing durable reply paths" problem dissolves rather than being coordinated around. `assertDurableReplyHandled` is exhaustive against the helper's five statuses, and `sourceReplyDeliveryMode: "automatic"` + `disableBlockStreaming: true` leaves exactly one durable owner.

One blocking problem.

---

### The adapter is missing `messageSendingHooks`, so every inbound reply is rejected before it sends — `src/outbound.ts:310`

```ts
durableFinal: {
  capabilities: { text: true, media: true, replyTo: true },
},
```

`deliverInboundReplyWithMessageSendContext` derives its requirements without ever passing `messageSendingHooks` (`kernel-CpJ_eT7K.js:451`, which forwards only `payload`, `replyToId`, `threadId`, `silent`). In `deriveDurableFinalDeliveryRequirements` that field is computed as:

```js
setRequired(requirements, "messageSendingHooks", params.messageSendingHooks !== false);
```

`undefined !== false` is `true`, and `setRequired` records anything strictly `true` — so `messageSendingHooks: true` is **always** required on this path. The support check then does a plain declared-capability lookup:

```js
if (required === true && durableFinal?.[capability] !== true) return { ok: false, reason: "capability_mismatch", capability };
```

Working it through for a normal text final with `replyToId: message.messageIdHex`:

- required → `{ text: true, replyTo: true, messageSendingHooks: true }`
- declared → `{ text: true, media: true, replyTo: true }`
- → `{ ok: false, reason: "capability_mismatch", capability: "messageSendingHooks" }`
- → `{ status: "unsupported", reason: "capability_mismatch" }`
- → `assertDurableReplyHandled` throws `marmot: OpenClaw durable reply was not handled (capability_mismatch)`
- → `dispatch` throws → `BoundedKeyedAsyncQueue` swallows it into `marmot: inbound dispatch task failed` → the id stays deduped → **the group gets no reply**

That is issue #1125's exact symptom, so as it stands the PR would ship without fixing the bug it targets. Every built-in channel declares this capability — Slack/Discord/Telegram/iMessage/Feishu/Tlon/Teams/Zalo/Nextcloud all carry `messageSendingHooks: true` in `deliveryCapabilities.durableFinal`. It isn't a feature the adapter has to implement (nothing in `deliver-DGFKMFsl.js` or the send path branches on it); it's a compatibility assertion core requires before it will use the durable path. The fix is one line:

```ts
capabilities: { text: true, media: true, replyTo: true, messageSendingHooks: true },
```

Worth checking `thread` too if you ever set `ctxPayload.ThreadId` — same gate, same failure mode.

**Why no test caught it:** every test substitutes the helper, so the capability gate never runs.

- `test/dispatch.test.ts:29` mocks `deliverInboundReplyWithMessageSendContext` to return a handled result.
- `test/e2e-connector.test.ts:205` injects a `deliverInboundReply` that reimplements the logic — reads `OriginatingTo ?? To`, calls `adapter.send.text` directly, returns `handled_visible`.

So the connector E2E's sequential-reply assertions validate that `ctxPayload` carries the right destination and that the adapter reaches wn-agent — both worth having — but not the SDK contract that production actually depends on. The `deps.deliverInboundReply` seam is reasonable for focused unit tests; the gap is that *nothing* exercises the real function. One test that drives the genuine `deliverInboundReplyWithMessageSendContext` against a registered adapter (asserting `status === "handled_visible"`) would have caught this and would guard the capability set against future SDK bumps — which matters given `AGENTS.md` already warns to re-verify `openclaw/plugin-sdk/*` exports on every pin bump.

---

Smaller notes, none blocking:

- `src/gateway.ts:123` — wrapping the whole body in try/catch with `markStopped("inbound startup failed")` is a real improvement. Worth knowing the host would have covered you anyway: `startChannelInternal` catches a rejecting `startAccount`, records `lastError`, and calls `setStoppedRuntime`. The explicit handler gives a better message, so keep it.
- `src/bounded-keyed-async-queue.ts:42` — `.catch(() => this.onShed?.("marmot: inbound dispatch task failed"))` discards the error. With the retry and dedupe-rollback machinery gone, this is now the *only* place a failed turn surfaces, so a delivery failure is one generic line with no cause. Threading a small fixed category through (as the abandoned branch did) is cheap and would have made the issue above visible immediately rather than looking like a silent no-reply.
- `src/dispatch.ts:367` — `finalDeliveries` counts every final and each one sends separately. That's fine and matches other channels; just noting the "exactly one durable reply" property from the old design is now "one per final emitted," which is the standard contract.
- Losing the deterministic idempotency key means a same-key replay no longer dedupes at wn-agent (`outbound.ts` is back to `randomUUID()` per send). Correct for this design — nothing replays a turn anymore — but it does mean the connector-side ambiguity I raised on #1137 is simply out of scope here rather than solved. Fine to leave to the protocol work; worth not losing the thread.

Once `messageSendingHooks` is declared and one test drives the real delivery helper, this is in much better shape than the version it replaces.

erskingardner:
Re-reviewed at `a083f4b7`. Both items are fixed, and the second one properly rather than minimally.

**Capability fix verified.** I re-derived the full chain against the pinned SDK rather than assuming: for a normal text final with `replyToId`, requirements resolve to `{ text, replyTo, messageSendingHooks }`, and all three are now declared. Also checked the `thread` gate I flagged as a follow-up — `MessageThreadId` is `params.reply.messageThreadId ?? params.conversation.threadId` (`kernel-CpJ_eT7K.js:306`), and `dispatch.ts` sets neither, so `threadId` stays undefined and `thread` is never required. The declared set is complete for the current context shape; the caveat only bites if a thread id is introduced later.

**The E2E change is the right fix, not a patch.** Dropping the `deliverInboundReply` override and loading the real plugin through `loadOpenClawPlugins({ activate: true, loadModules: true })` means the production helper now resolves the registered adapter and runs the capability gate for real — which is what I was actually asking for. Asserting `plugin.message` is defined after load is a good cheap guard, and adding `pnpm build` to the E2E script is the necessary consequence of loading from disk.

Two things about where that guard lives, both about coverage rather than correctness.

---

### 1. Neither the new E2E nor the OpenClaw unit tests run in CI

`Package OpenClaw plugin` is the only OpenClaw job, and it does:

```
pnpm install --frozen-lockfile && pnpm build
```

then packs a tarball (`.github/workflows/wn-agent-binaries.yml:391`). It never runs `pnpm test`. The connector E2E is gated behind `MARMOT_OPENCLAW_CONNECTOR_E2E=1` and only reachable via `just openclaw-dev-e2e-connector`, so it doesn't run there either.

So both new guards for this bug class — the capability assertion in `test/outbound.test.ts` and the now-real E2E — only fire when someone runs them locally. Given the bug that prompted this was specifically one that "no test caught," it's worth knowing that a recurrence still wouldn't be caught automatically. The unit suite is fast and hermetic (the E2E's wn-agent dependency is what makes *it* awkward to run in CI, but `pnpm test` has no such constraint); adding it to the existing job would close most of the gap cheaply.

Worth noting `pnpm build` uses `tsconfig.build.json`, which excludes `test`, so the new test file isn't even typechecked in CI — see the next point for why that matters here specifically.

### 2. The loader import bypasses the package's export map — `test/e2e-connector.test.ts:7`

```ts
from "../node_modules/openclaw/dist/plugins/loader.js"
```

`openclaw` publishes 325 export subpaths and none of them is this; there's no `./dist/*` wildcard and no plugin-loader export at all. A relative path into `node_modules` sidesteps `exports` enforcement entirely, so this reaches into internal dist layout that the package makes no promises about.

That matters more than usual here because `AGENTS.md` already sets the discipline — *"Keep the `openclaw` dependency pinned; before bumping, verify the `openclaw/plugin-sdk/*` subpath exports against the new version's types."* A raw dist path is invisible to that check: it isn't a subpath export, it isn't typechecked by the CI build (excluded above), and it isn't executed in CI. A pin bump that renames or bundles `dist/plugins/loader.js` differently would surface as a module-not-found only when someone next runs the E2E by hand.

If `loadOpenClawPlugins` has no supported entry point, that's worth raising upstream, since loading a plugin for test is a reasonable thing to want. In the meantime a comment marking it as a deliberate internal import — and adding it to the list of things to re-verify on a pin bump — would keep it from being a silent trap.

---

Minor, non-blocking: `test/outbound.test.ts:118` asserts the capability object with `toEqual` against a literal. That pins what's *declared*, which is useful, but it can't detect the failure mode that caused this bug — a capability becoming *required* by the SDK that isn't declared. A cross-check that runs `deriveDurableFinalDeliveryRequirements` for a representative payload and asserts the declared set covers it would catch the next occurrence at the unit level, and would run in CI if point 1 is addressed.

The substance of this PR looks right to me: the routing contract now matches what the SDK documents, the capability declaration is complete, and the E2E exercises the real path. My remaining comments are about making sure that stays true after the next `openclaw` bump.

PR #1144, opened 2026-07-27
Expose pinned retention metadata in timeline bindings

erskingardner:
## Summary

- expose `source_epoch`, `retention_seconds`, and `retention_expires_at` on timeline message records and UniFFI bindings
- load pinned retention metadata from the authoritative `app_events` row for paginated and by-ID timeline reads
- preserve legacy `None`, explicit `Some(0)`, finite expiry, and positive-duration-without-expiry states exactly
- cover rebuild and incremental finalization paths with storage and conversion regression tests

## Root cause

`app_events` already persisted the retention decision pinned to the message's authenticated source epoch, but `TimelineMessageRecord` was populated only from `message_timeline`. The timeline conversion therefore dropped that metadata before Swift and Kotlin consumers could receive it.

Timeline reads now `LEFT JOIN` the authoritative source event by group and message ID. The values are not denormalized into `message_timeline`, so no schema migration or backfill is required and reads cannot become stale relative to retention finalization.

## Consumer impact

Generated bindings add these nullable fields to `TimelineMessageRecordFfi`:

- Swift: `sourceEpoch`, `retentionSeconds`, `retentionExpiresAt` as `UInt64?`
- Kotlin: `sourceEpoch`, `retentionSeconds`, `retentionExpiresAt` as `ULong?`

Consumers that construct the record explicitly must update their initializers after regenerating or vendoring the bindings. This remains part of the unreleased `0.9.8` API cohort.

## Validation

- `cargo fmt --check`
- `cargo test -p storage-sqlite` — 288 passed
- `cargo check -p marmot-uniffi --features otlp-export`
- `just fast-ci`
- `./crates/marmot-uniffi/kotlin-bindings.sh` — all four Android JNI ABIs
- `cargo test -p marmot-app` — all 467 unit tests passed; five `relay_runtime` integration tests failed identically when rerun from a clean worktree at the exact current `origin/master`, confirming pre-existing baseline failures unrelated to this diff

## Summary by CodeRabbit

* **New Features**
  * Timeline messages now expose retention duration and expiration details.
  * Retention states—including disabled, unspecified, finite, and unknown expiration—are preserved accurately.
* **Bug Fixes**
  * Fixed timeline reads and message lookups so retention information remains consistent across pagination and timeline rebuilds.
  * Updated projections to display finalized retention settings immediately without reverting to outdated values.
* **Tests**
  * Added coverage for retention states, timeline rebuilding, and incremental retention updates.

### DIFF
--- a/crates/cli/src/commands/messages.rs
+++ b/crates/cli/src/commands/messages.rs
@@ -821,6 +821,8 @@ mod tests {
             message_id_hex: "system-1".to_owned(),
             source_message_id_hex: None,
             source_epoch: Some(3),
+            retention_seconds: None,
+            retention_expires_at: None,
             direction: "system".to_owned(),
             group_id_hex: "11".repeat(32),
             sender: actor.clone(),
--- a/crates/marmot-app/src/runtime/tests.rs
+++ b/crates/marmot-app/src/runtime/tests.rs
@@ -263,6 +263,8 @@ fn timeline_test_record(message_id_hex: &str, timeline_at: u64) -> TimelineMessa
         message_id_hex: message_id_hex.to_owned(),
         source_message_id_hex: None,
         source_epoch: None,
+        retention_seconds: None,
+        retention_expires_at: None,
         group_id_hex: "group-1".to_owned(),
         direction: "inbound".to_owned(),
         sender: "sender-1".to_owned(),
--- a/crates/marmot-uniffi/src/conversions/mod.rs
+++ b/crates/marmot-uniffi/src/conversions/mod.rs
@@ -289,6 +289,8 @@ mod tests {
             message_id_hex: "message-1".to_owned(),
             source_message_id_hex: Some("source-1".to_owned()),
             source_epoch: Some(7),
+            retention_seconds: Some(300),
+            retention_expires_at: Some(310),
             direction: "received".to_owned(),
             group_id_hex: "11".repeat(32),
             sender: "aa".repeat(32),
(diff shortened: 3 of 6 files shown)


erskingardner (commented):
**Recommend merge** (posted as comment, not approval — GitHub blocks self-approval since this PR is authored by the same account).

Reviewed the diff, the migration history for `app_events`/`message_timeline`, and the retention-decision pinning logic that predates this PR (migration 0029, `record_app_event_inner`, `finalize_app_event_source_retention`).

**Verified correct:**
- `app_events` has `UNIQUE (group_id_hex, message_id_hex)`, matching `message_timeline`'s PK, so the new `LEFT JOIN` in `timeline_records_by_ids_tx` / `timeline_query_sql` is strictly 1:1 — no row fan-out risk, and the join key is covered by the existing unique index.
- Every filter/order clause in `timeline_query_sql` (group filter, search, both cursor directions) was consistently updated to the `timeline.` alias; nothing was missed.
- Column indices in `timeline_record_from_row` were correctly shifted by 2 for both call sites, matching the new `SELECT` column order exactly.
- Retention data intentionally stays non-denormalized (only in `app_events`), matching the PR's stated no-migration goal; `message_timeline`/`TimelineRow` correctly remains untouched.
- All `TimelineMessageRecord { .. }` struct-literal sites across `cli`, `marmot-app` tests, and `marmot-uniffi` were updated for the two new fields (struct has no `Default`, so this is compiler-enforced).
- New storage tests (`timeline_reads_preserve_every_pinned_retention_state`, `incremental_retention_finalization_is_visible_without_stale_projection_data`) exercise legacy/disabled/finite/overflow retention states across paginated read, by-ID read, and rebuild paths, which is exactly the surface this PR touches.

No correctness, scope, or style issues found. Small, surgical, well-tested change that matches its stated root cause.

Issue #687, opened 2026-05-21
[High] No cryptographic binding between credential identity and MLS signature key (admin impersonation)


erskingardner:
**Severity:** High
**Component:** `crates/cgka-engine`

## Location
- `crates/cgka-engine/src/app_components.rs:161-187` (`require_admin`, `admin_pubkey_from_member_id`)
- `crates/cgka-engine/src/identity.rs`
- Member-id derivation: `crates/cgka-engine/src/message_processor.rs` (`member_id_of_sender`), `group_lifecycle.rs`

## Summary
`MemberId` is taken verbatim from `BasicCredential::identity()`, and admin checks compare those self-asserted identity bytes against the admin set. Nothing proves the member controls the secret key for that identity, so a member can present an admin's pubkey as their own credential identity.

## Details
`require_admin` resolves the member's admin pubkey straight from the credential identity bytes:

```rust
pub(crate) fn require_admin(mls_group, group_id, member_id) -> Result<(), EngineError> {
    let member_pubkey = admin_pubkey_from_member_id(member_id)?; // = identity() bytes
    let admins = admins_of_group(mls_group)?;
    if admins.iter().any(|admin| admin == &member_pubkey) { return Ok(()); }
    Err(EngineError::NotGroupAdmin { ... })
}
```

MLS does not require unique or verified credential identities. A member can mint a KeyPackage whose `BasicCredential::identity()` equals a real admin's 32-byte pubkey while signing with their own MLS key. The engine treats the (attacker-chosen, unauthenticated) identity bytes as a trusted admin pubkey.

## Impact
Admin impersonation for all engine-side authorization decisions. On its own this requires such a KeyPackage to be invited; combined with the convergence admin-bypass issue it is directly exploitable. The trust model assumes credential identity == controlled pubkey, which is not enforced anywhere. **Verified against source** (no binding check found in the engine).

## Suggested fix
- Enforce the identity↔signature-key binding before a KeyPackage/leaf is accepted (reject leaves whose declared identity isn't cryptographically tied to the signing key), **or**
- Key admin policy off the MLS signature key rather than the self-asserted credential identity.
- Confirm whether the session/account layer is supposed to verify this; it does not appear to today.

_Found during a full-codebase security review; verified against source._

erskingardner:
[...]

Issue #667, opened 2026-05-21
[Medium] Directory search BFS over the follow graph is unbounded by breadth


erskingardner:
**Severity:** Medium
**Component:** `crates/marmot-app`

## Location
- `crates/marmot-app/src/lib.rs:5004-5028` (`directory_search_radii`)
- `crates/marmot-app/src/lib.rs:2689-2693` (loads all directory entries into a `HashMap`)

## Summary
The follow-graph BFS bounds depth (`radius_end`) but not breadth: the frontier expands by every record's `follows` with no cap on total visited nodes. Follow lists are attacker-controlled (fetched from relays), and `search_user_directory` first loads the entire cache into memory.

## Details
```rust
next.extend(record.follows.iter().cloned()); // no cap on total visited / frontier size
```

A few accounts each following thousands of pubkeys make radius-2/3 searches scan and clone a very large set; combined with the full-cache load, this is CPU/memory amplification on a user-triggered search. The crate's own docs warn to keep directory search bounded.

## Impact
CPU/memory amplification on directory search, driven by attacker-controlled follow lists. **Verified against source.**

## Suggested fix
Cap total visited nodes / frontier size; stream from the DB rather than loading all entries into memory.

_Found during a full-codebase security review; verified against source._

Issue #418, opened 2026-06-29
Unbounded transitive Nostr social-graph crawl in directory sync (violates the AGENTS.md 'must not become an unbounded social-graph crawler' invariant)


erskingardner:
## Summary
The user-directory sync worker subscribes to kind-3 **contact lists** for every known directory entry, and ingesting a contact list **persists every followed pubkey as a new directory entry**, which schedules another sync rebuild. A self-reinforcing loop with **no cap on persisted directory size or per-list breadth** — the unbounded social-graph crawl the crate's `AGENTS.md` forbids:
> Keep the Nostr user directory app-facing and pubkey-keyed … it must not become an unbounded Nostr social-graph crawler.

## The loop (all links verified in code)
1. `DIRECTORY_SYNC_KINDS` includes `KIND_NOSTR_CONTACT_LIST` — `crates/marmot-app/src/directory/sync.rs:19-25`.
2. `directory_sync_plan` enumerates **all** of `directory_entries()` (every per-account + shared public-directory cache) and subscribes for each — `crates/marmot-app/src/directory/methods.rs:617-639` (`directory_entries` `:600-615`).
3. Ingested kind-3 routes to `remember_directory_follow_list` — `methods.rs:872-874`.
4. It persists **each** followed pubkey via `remember_directory_user` → `save_directory_entry_with_reason` — `methods.rs:779-794`.
5. For any new pubkey it writes the entry and calls `request_directory_sync_rebuild()` — `methods.rs:932-936`.
6. Rebuild re-runs `directory_sync_plan`, now including the new pubkeys → subscribes to *their* contact lists → back to step 3.

No bound on persisted entry count (`USER_DIRECTORY_SEARCH_MAX_VISITED`/`MAX_FRONTIER` bound only *search-time* traversal). `follow_list_from_record` (`crates/marmot-app/src/directory/records.rs:425-440`) collects **every** `p` tag from an untrusted relay event with no per-list cap, so one hostile kind-3 with 100k `p` tags seeds 100k watched users in a single ingest.

## Impact
- Unbounded SQLCipher directory growth, persisted across restarts.
- Transitive subscription/relay fan-out → privacy leak (ever-widening interest graph exposed to relays) + self-amplifying DoS from one untrusted contact list.
- Receive-path cost grows: `display_names_by_id()`/`directory_entries()` become O(total directory size) (compounds marmot-protocol/mdk#434).

## Suggested fix
[...]

PR #1148, opened 2026-07-27
Feat/user search web of trust

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Added live, streaming user search across the follow graph with incremental updates.
  * `users search` now reports traversal completeness (`complete` / `incomplete_reason`) and emits radius progress events.
  * Added bounded `search_users` support in the app API/FFI.
* **Improvements**
  * Search prioritizes live follow-graph discovery over local cache only.
  * Discovered strangers are no longer promoted into the local directory.
  * Discovered relay endpoints are safety-filtered; results include deterministic match attribution.
* **Bug Fixes**
  * Improved CLI error reporting for search failures.
* **Chores**
  * Updated package version and conformance/vector references to **0.9.10**, plus related documentation.


crates/marmot-app/src/directory/search.rs
```diff
@@ -0,0 +1,1789 @@
...
+        let unresolved = fetched
+            .iter()
+            .filter(|record| record.profile.is_none())
+            .map(|record| record.account_id_hex.clone())
+            .collect::<Vec<_>>();
+        emitter
+            .tally
+            .resolved_from_relays(fetched.len() - unresolved.len());
+        cache_resolved_profiles(app, &fetched, now).await?;
+        emitter.emit_matches(radius, fetched, query).await;
```
erskingardner: **Confirmed: duplicate results when an npub/pubkey match resolves via the outbox tier.**

`user_record_match` always includes `npub`/`account_id_hex` as candidates regardless of whether `profile` is set (records.rs). So for a batch member whose profile isn't found on connected relays:

1. This `emit_matches(radius, fetched, query)` call emits them once, matched on `Npub`/`Pubkey`, with `profile: null` (the `fetch_profile_records` doc comment even calls this out as intentional: *"an npub search miss a follow whose profile simply is not published anywhere"*).
2. If they're in `unresolved`, `resolve_from_write_relays` may then resolve their profile.
3. `emit_matches(radius, outboxed, query)` on line 424 re-runs `user_record_match` on the same account. The npub/pubkey fields haven't changed, so the match succeeds again — emitting the **same `account_id_hex` a second time** in the same radius, now with a populated profile.

Nothing downstream dedupes by `account_id_hex` (`SearchEmitter`, `sort_user_search_results`, the CLI's `collect_user_search`, or the UniFFI `UserSearchSubscription`), so `wn users search <npub-or-pubkey-substring>` (and the TUI/UniFFI hosts built on it) can show the same person twice for a query pattern the code explicitly says it wants to support. None of the existing tests query by npub/pubkey combined with outbox resolution, so this gap isn't caught.

Suggest either skipping the first `emit_matches` for records that are about to go through outbox resolution (only emit them once, after outbox resolution has had a chance to run), or deduping the aggregated result stream by `account_id_hex` (keeping the better/more-complete match).

crates/marmot-app/src/directory/search.rs
```diff
@@ -0,0 +1,1789 @@
...
+    /// The follow graph is not the only evidence that two people know each
+    /// other: sharing a group is social proximity even when neither has
+    /// followed the other. That membership is live MLS state owned by the
+    /// account runtime, not by the directory, so it arrives here as an input
+    /// rather than search reaching across the boundary to read it. Callers
+    /// without a running runtime simply pass none.
+    ///
+    /// The searcher is radius 0; naming them here does not move them.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub radius_one_seeds: Vec<String>,
```
erskingardner, line 135: Minor: `radius_one_seeds` isn't checked by `validate()` below (only `radius_start`/`radius_end`/`searcher_account_id_hex` are). `fallback_seed_frontier` defensively skips malformed seeds via `parse_account_id_hex(seed).ok()`, but a malformed entry here flows through to `fetch_profile_records` -> `fetch_events_for_account_ids`, whose `parse_account_id_hex(...).collect::<Result<_,_>>()?` fails the *entire* batch/traversal on one bad seed. Low real-world risk today since the only caller (`group_co_members`) supplies well-formed hex, but `UserSearchParams` and this field are both `pub`, so any other consumer of `marmot-app`'s public API doesn't get the same one-bad-seed-costs-one-seed guarantee the fallback path promises.

crates/marmot-app/AGENTS.md
```diff
@@ -52,9 +52,13 @@ App runtime bridge for the first real Marmot app surfaces.
   reused by `key_package_records.rs`, and public re-exports.
 - Keep CLI/TUI presentation out of this crate.
 - Keep the Nostr user directory app-facing and pubkey-keyed. It may cache local-account links, profile metadata, follow
-  lists, relay lists, and KeyPackages, but it must not become an unbounded Nostr social-graph crawler.
-- Keep directory search bounded over cached follow edges. Do not add web-of-trust scoring unless that is reopened as a
-  deliberate product decision.
+  lists, relay lists, and KeyPackages. Web-of-trust search may traverse the live follow graph, but traversal is bounded
+  by construction: a capped search radius, a per-radius candidate cap, a per-radius fetch timeout, batched
+  author-scoped fetches with bounded concurrency, and a per-search lifecycle — traversal runs only while a search
```
erskingardner: "...batched author-scoped fetches with bounded concurrency..." doesn't match the implementation: `resolve_layer`, `extend_with_follows`, and `resolve_from_write_relays` all process their batches/relays sequentially in a `for` loop with no concurrent fetches. The follows-of-follows commit message says this explicitly: *"batching the existing author-scoped fetch is both simpler and strictly fewer round trips, so there is no concurrency machinery to get wrong."* Worth wording this invariant as "batched" only, so it doesn't misdescribe the design to a future reader.

erskingardner (changes requested):
This is a large, unusually well-documented and well-tested feature (streaming web-of-trust user search), and the bounded-by-construction design (capped radius, per-radius candidate cap, timeout, batching) directly and thoughtfully addresses the concerns behind the previously-closed `[Medium] Directory search BFS over the follow graph is unbounded by breadth` (#667) and `Unbounded transitive Nostr social-graph crawl` (#418) issues. The un-promoted-tier discipline (never writing `directory_users`) is consistently upheld and guarded by positive-control tests.

I found one confirmed correctness bug (duplicate results) plus a couple of minor/non-blocking notes. Given the duplicate-result bug hits a documented, intended query path (searching by npub/pubkey) with zero test coverage of that combination, I'm requesting changes rather than approving outright — everything else here is very solid.

### Summary of findings
- **Confirmed bug**: `resolve_layer` can emit the same account twice in one search (once with `profile: null`, once resolved) when a query matches on npub/pubkey and the profile is later found via the outbox (write-relay) tier. No dedup exists anywhere downstream (emitter, `sort_user_search_results`, CLI's `collect_user_search`, or the UniFFI subscription).
- Minor: `UserSearchParams::radius_one_seeds` isn't validated in `validate()`, unlike the fallback-seed path which defensively skips malformed entries.
- Minor: `AGENTS.md`/`README.md` now describe fetches as having "bounded concurrency", but the implementation (per the follows-of-follows commit message itself) is purely sequential batching with no concurrency.
- Question: several new comments/docs/tests cite `mdk#687` as the origin of the "no live per-author subscriptions for discovered strangers" invariant, but issue #687 in this tracker is about an unrelated admin-impersonation vulnerability. Worth double-checking the reference (perhaps #667 or #418 was intended) since it's now baked into permanent code comments and `AGENTS.md`.

erskingardner (approved):

PR #1149, opened 2026-07-28
Persist manual chat pin ordering

erskingardner:
## Summary

- persist device-local chat pin positions in each account-device SQLCipher database
- expose transactional pin, unpin, and complete manual reorder commands through `marmot-app` and UniFFI
- keep pinned chats ahead of activity-sorted unpinned chats without allowing new messages to reorder the pinned section
- publish atomic chat-list snapshots for pin-order and archive changes while preserving the legacy flattened `next()` subscription API
- bump the workspace and MarmotKit version from `0.9.8` to `0.9.9`

## Client impact

Clients receive `pinned` and normalized zero-based `pinned_position` fields on chat-list rows. `set_chat_pinned` and `set_pinned_chat_order` return the authoritative ordered pinned group IDs after the transaction. Invalid complete-order requests return the typed `InvalidChatPin` error.

This binding change is source-breaking for clients with exhaustive Swift `switch` or Kotlin `when` expressions: hosts must handle `ChatListSubscriptionUpdateFfi::Snapshot` and `ChatListUpdateTriggerFfi::PinOrderChanged`, and construct or read the two new required `ChatListRowFfi` fields.

Pin state is deliberately local to an account-device database and does not enter MLS group state or synchronize across devices. Archiving a conversation removes its pin transactionally; unarchiving does not restore it.

## Validation

- `cargo test -p storage-sqlite` — 293 passed
- `cargo test -p marmot-uniffi` — 78 unit, 2 fixture, and 26 smoke tests passed
- pinning runtime/FFI coverage includes account isolation, pin, unpin, reorder, archive, snapshots, persistence, projection rebuilds, departed/pending eligibility, and typed errors
- `cargo test -p cgka-conformance-simulator --test canonical_scenarios` — 27 passed
- Swift XCFramework generation passed
- Kotlin binding generation and all Android ABI builds passed
- `just fast-ci` passed

## Existing test baseline

`cargo test -p marmot-app` runs all 468 unit tests successfully, then encounters five existing `relay_runtime` integration failures. A representative failure reproduces identically from clean `master`; feature-specific runtime and UniFFI tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added durable chat pinning and custom pinned-chat ordering.
  * Pinned chats now appear first and display pin status and position.
  * Added APIs for pinning, unpinning, and updating pinned-chat order.
  * Chat list updates now provide atomic snapshots when pin order changes.
  * Added clear validation errors for invalid pin requests and ordering.
  * Pinned settings persist across database reopen and are cleared for archived chats.

* **Bug Fixes**
  * Improved chat-list refresh behavior after pinning and archiving changes.

* **Chores**
  * Updated the release and conformance version to 0.9.9.

### DIFF
--- a/crates/marmot-app/src/runtime/subscriptions.rs
+++ b/crates/marmot-app/src/runtime/subscriptions.rs
@@ -525,6 +525,11 @@ pub enum RuntimeChatListUpdate {
         trigger: ChatListUpdateTrigger,
         group_id_hex: String,
     },
+    /// Full replacement for this subscription's visible chat list.
+    ///
+    /// Consumers must atomically replace their locally held rows with `rows`;
+    /// any previously held row absent from this value has left the subscribed
+    /// list and must be removed.
     Snapshot {
         trigger: ChatListUpdateTrigger,
         rows: Vec<ChatListRow>,
--- a/crates/marmot-app/src/runtime/subscriptions.rs
+++ b/crates/marmot-app/src/runtime/subscriptions.rs
@@ -525,6 +525,10 @@ pub enum RuntimeChatListUpdate {
         trigger: ChatListUpdateTrigger,
         group_id_hex: String,
     },
+    Snapshot {
+        trigger: ChatListUpdateTrigger,
+        rows: Vec<ChatListRow>,
+    },
 }
 
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
@@ -540,6 +544,7 @@ pub enum ChatListUpdateTrigger {
     MuteChanged,
     ConversationKindChanged,
     LatestMessageDeliveryChanged,
+    PinOrderChanged,
     #[default]
     SnapshotRefresh,
     Removed,
@@ -1195,6 +1200,34 @@ impl MarmotAppRuntime {
                 if let Some(update) = projection_update_from_event(&event)
                     && update.account_id_hex == account_id_hex
                 {
+                    if matches!(
+                        update.update.chat_list_trigger,
+                        ChatListUpdateTrigger::PinOrderChanged
+                            | ChatListUpdateTrigger::ArchiveChanged
+                    ) {
+                        let app_for_lookup = app.clone();
+                        let account_label_for_lookup = account_label.clone();
+                        let rows = match blocking_app_task(move || {
+                            app_for_lookup.chat_list(&account_label_for_lookup, include_archived)
+                        })
+                        .await
+                        {
+                            Ok(rows) => rows,
+                            Err(_) => continue,
+                        };
+                        mute_expiries = chat_list_mute_expiries(&rows);
+                        if !send_atomic_chat_list_snapshot(
+                            &updates_tx,
+                            &mut row_fingerprints,
+                            update.update.chat_list_trigger,
+                            rows,
+                        )
+                        .await
+                        {
+                            return;
+                        }
+                        continue;
+                    }
                     let Some(row) = update.update.chat_list_row.clone() else {
                         mute_expiries.remove(&update.update.group_id_hex);
                         if !send_chat_list_remove_update(
@@ -1644,4 +1677,21 @@ pub(crate) async fn reconcile_chat_list_snapshot(
         }
     }
     true
+}
+
+pub(crate) async fn send_atomic_chat_list_snapshot(
+    updates_tx: &mpsc::Sender<RuntimeChatListUpdate>,
+    row_fingerprints: &mut HashMap<String, String>,
+    trigger: ChatListUpdateTrigger,
+    rows: Vec<ChatListRow>,
+) -> bool {
+    row_fingerprints.clear();
+    row_fingerprints.extend(
+        rows.iter()
+            .map(|row| (row.group_id_hex.clone(), chat_list_row_fingerprint(row))),
+    );
+    updates_tx
+        .send(RuntimeChatListUpdate::Snapshot { trigger, rows })
+        .await
+        .is_ok()
 }
--- a/crates/marmot-uniffi/src/subscriptions.rs
+++ b/crates/marmot-uniffi/src/subscriptions.rs
@@ -95,6 +95,16 @@ impl ChatListSubscription {
         take_snapshot(&self.snapshot).unwrap_or_default()
     }
 
+    /// Legacy row-only update stream.
+    ///
+    /// Atomic replacement snapshots are flattened into every visible row in
+    /// authoritative order so existing clients can observe changed pin fields.
+    /// This can yield many rows for one pin or archive operation and cannot
+    /// express the replacement boundary or removals. New clients that need
+    /// correct manual ordering and archive removals should use `next_update`.
+    ///
+    /// `next` and `next_update` consume the same stream and must not be mixed
+    /// for the lifetime of one subscription.
     pub async fn next(&self) -> Option<ChatListRowFfi> {
         let mut state = self.state.lock().await;
         if let Some(row) = state.pending_rows.pop_front() {
@@ -114,6 +124,11 @@ impl ChatListSubscription {
         }
     }
 
+    /// Typed update stream, including atomic full-list replacement snapshots.
+    ///
+    /// Do not mix this with `next` on the same subscription: both consume the
+    /// same underlying stream, while `next` may also hold flattened snapshot
+    /// rows in its compatibility buffer.
     pub async fn next_update(&self) -> Option<ChatListSubscriptionUpdateFfi> {
         let mut state = self.state.lock().await;
         state.inner.recv().await.map(Into::into)
--- a/crates/marmot-uniffi/src/subscriptions.rs
+++ b/crates/marmot-uniffi/src/subscriptions.rs
@@ -13,6 +13,7 @@
 //! All inner state lives behind a `tokio::sync::Mutex` because UniFFI passes
 //! these objects via `Arc<Self>` and `recv()` requires `&mut`.
 
+use std::collections::VecDeque;
 use std::sync::Arc;
 use std::sync::Mutex as StdMutex;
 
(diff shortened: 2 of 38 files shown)


crates/marmot-app/src/runtime/subscriptions.rs
```diff
@@ -1645,3 +1678,20 @@ pub(crate) async fn reconcile_chat_list_snapshot(
     }
     true
 }
+
+pub(crate) async fn send_atomic_chat_list_snapshot(
```
erskingardner, line 1687: **`Snapshot` is a full replacement, but nothing in the type says so — and it is now the only removal signal for archive.**

Unlike `reconcile_chat_list_snapshot` (which diffs against `row_fingerprints` and emits `RemoveRow` for rows that dropped out), this helper clears the fingerprint map and sends only the surviving rows. Since `ArchiveChanged` is now routed here, when `include_archived == false` the *only* indication that a chat was archived away is its absence from `rows`.

Before this PR that same case emitted an explicit `RemoveRow` (the `if !include_archived && row.archived` branch below at ~line 1243). A Swift/Kotlin host that merges `Snapshot.rows` into its existing keyed model — a very natural reading of an update named "Snapshot" arriving on a stream that has always been row-at-a-time — will keep displaying the archived chat forever, with no error and no other event to correct it.

Suggest a doc comment on `RuntimeChatListUpdate::Snapshot` and on `ChatListSubscriptionUpdateFfi::Snapshot` stating explicitly that `rows` replaces the entire subscribed list and that any locally-held row absent from `rows` must be dropped. Both variants currently have no doc comment at all, and this is the one detail a host cannot infer.

crates/marmot-uniffi/src/subscriptions.rs
```diff
@@ -87,18 +96,27 @@ impl ChatListSubscription {
...
+        if let Some(row) = state.pending_rows.pop_front() {
+            return Some(row);
+        }
         loop {
-            match inner.recv().await? {
+            match state.inner.recv().await? {
                 marmot_app::RuntimeChatListUpdate::Row { row, .. } => return Some((*row).into()),
                 marmot_app::RuntimeChatListUpdate::RemoveRow { .. } => continue,
+                marmot_app::RuntimeChatListUpdate::Snapshot { rows, .. } => {
+                    state.pending_rows.extend(rows.into_iter().map(Into::into));
```
erskingardner, line 118: **The legacy `next()` path now emits the whole chat list on every pin toggle and every archive.**

Flattening `Snapshot` into `pending_rows` keeps `next()` compiling and semantically "correct", but it changes its cost profile: previously archiving a chat under `include_archived == false` yielded *zero* items from `next()` (the `RemoveRow` was skipped on line 106), and pinning did not exist. Now each of those operations pushes one item per chat in the account through this API. For an account with a few hundred chats, one drag-reorder is a few hundred `next()` wakeups and a few hundred FFI record conversions.

Worth considering filtering the flattened rows against the fingerprints the runtime just recomputed, so `next()` only surfaces rows whose content actually changed — a pure reorder would then emit nothing here, which is the honest answer for an API that cannot express ordering anyway.

crates/marmot-uniffi/src/subscriptions.rs
```diff
@@ -87,18 +96,27 @@ impl ChatListSubscription {
...
+                    if let Some(row) = state.pending_rows.pop_front() {
+                        return Some(row);
+                    }
+                }
             }
         }
     }
 
     pub async fn next_update(&self) -> Option<ChatListSubscriptionUpdateFfi> {
```
erskingardner, line 132: Minor: `next_update()` does not drain `pending_rows`. A host that calls `next()` once and then switches to `next_update()` silently loses whatever rows are still buffered from the last `Snapshot`. Probably no real host mixes the two, but since `pending_rows` is shared state behind the same mutex it costs one `if` to make it safe — or add a doc comment saying the two methods are mutually exclusive for the lifetime of the subscription.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -218,6 +241,89 @@ impl SqliteAccountStorage {
...
+                    "group ids must not contain duplicates".to_owned(),
+                ));
+            }
+
+            let current = pinned_chat_order_tx(&conn)?;
+            let current_set = current.iter().collect::<HashSet<_>>();
+            let requested_set = ordered_group_ids.iter().collect::<HashSet<_>>();
+            if current_set != requested_set {
+                return Err(ChatPinError::InvalidOrder(
+                    "order must contain every currently pinned chat exactly once".to_owned(),
```
erskingardner, line 331: Design question rather than a defect: requiring the request to match the current pinned set *exactly* means any pin the client did not know about turns a routine reorder into a user-visible `InvalidChatPin`. The archive trigger can drop a pin out from under the client (`save_account_projection_state` fires `unpin_chat_when_archived` on any upsert that flips `archived` to 1), so a drag-reorder computed from a list that is a few hundred milliseconds stale fails rather than degrading.

A more forgiving contract — intersect the request with the current pin set, preserve the requested relative order, append anything pinned-but-unlisted at the end — would make the operation total and still deterministic. If the strict contract is deliberate (it is defensible: it surfaces client/engine divergence loudly), a doc comment on `set_pinned_chat_order` saying so would help, since `InvalidChatPin` is the error a host has to decide how to present to a user mid-gesture.

erskingardner (commented):
## Review: persisted chat pin ordering

Solid, well-tested change. The storage design is the right shape: pin rows live outside the rebuildable `chat_list_rows` projection, `ON DELETE CASCADE` plus the `unpin_chat_when_archived` trigger keep the table from leaking rows, and normalizing `pinned_position` at read time correctly hides the ordinal gaps the trigger leaves behind. `set_chat_pinned` / `set_pinned_chat_order` both validate before mutating inside `with_transaction`, so the multi-step-state-changes invariant holds. Coverage is genuinely good — rebuild survival, encrypted reopen, account isolation, archive/delete cleanup, typed errors.

I ran locally and confirm the claims in the description:

- `cargo test -p storage-sqlite --lib chat_list` — 52 passed
- `cargo test -p marmot-uniffi chat_pin` — 1 passed
- `cargo clippy -p storage-sqlite -p marmot-app -p marmot-uniffi --all-targets` — clean

My findings are all about the **subscription contract change**, not the pin storage. Routing `ArchiveChanged` through the new snapshot path changes what a host must do to stay correct, and nothing in the code says so. Details inline.

### Also worth adding to "Client impact"

The new `ChatListSubscriptionUpdateFfi::Snapshot` variant and `ChatListUpdateTriggerFfi::PinOrderChanged` are source-breaking for exhaustive Swift `switch` / Kotlin `when` over those enums, and `ChatListRowFfi` gains two required fields. The workspace bump to 0.9.9 is correct per `crates/marmot-uniffi/AGENTS.md`, but the description reads as purely additive and hosts will not compile until they add the arms.

erskingardner (commented):
## Re-review of `7880dd48` — all four findings resolved

Re-reviewed the follow-up commit against `523c60c4`. Everything I raised is addressed, and the commit adds a worthwhile refactor I did not ask for. No new findings that would block.

### Resolutions

**1. `Snapshot` contract — fixed.** `RuntimeChatListUpdate::Snapshot` and `ChatListSubscriptionUpdateFfi::Snapshot` now both say the update replaces the whole list and that any prior row absent from `rows` has left the list and must be removed. That is exactly the detail a host could not infer, and stating it on the FFI variant means it lands in the generated Swift/Kotlin docs where the merge-vs-replace mistake would actually be made.

**2. `next()` fan-out — documented rather than changed, and that is the right call.** The new doc comment is honest about the cost ("can yield many rows for one pin or archive operation") and about the capability gap ("cannot express the replacement boundary or removals"), and points new clients at `next_update`. Filtering by fingerprint would have saved some FFI conversions, but not changing the observable behavior of a legacy compatibility path is the more conservative choice and I would not push further.

**3. `next()` / `next_update()` mixing — documented on both methods.** Fine. The buffer is still unenforced, but the contract is now stated on both sides of it rather than only implied.

**4. `set_pinned_chat_order` strict contract — documented with rationale and recovery.** The doc now says the compare-and-set is deliberate, and tells a client what to do with `InvalidOrder` (refresh the snapshot, retry from the authoritative set). That was the missing piece; a host author hitting this mid-gesture now knows it is a refresh-and-retry, not a bug.

### Unprompted improvement worth calling out

Extracting `CHAT_LIST_ROW_SELECT_AND_JOINS` collapses what was three near-identical 34-column projections into one, with a comment tying it to `chat_list_row_from_row`'s positional decode. That triplication predates this PR and was a live drift hazard — the two new columns had to be added in three places by hand to land correctly. Good cleanup to fold in here.

### Verification on this head

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p storage-sqlite -p marmot-app -p marmot-uniffi -p wn-cli --all-targets` — clean
- `cargo test -p storage-sqlite --lib chat_list` — 52 passed
- `cargo test -p marmot-app --lib chat_list` — 5 passed
- `cargo test -p marmot-uniffi chat_pin` — 1 passed
- `cargo test -p wn-cli --lib chat` — 53 passed

I specifically checked the SQL extraction for column-order drift, since `chat_list_row_from_row` decodes by index and a silent reorder would mis-assign fields rather than fail: the list and single-row paths now share one projection, and the tests that assert `pinned_position` through both `chat_list_rows` and `chat_list_row` still pass.

### Two small things, neither blocking

1. The "Client impact" section still reads as purely additive. The new `ChatListSubscriptionUpdateFfi::Snapshot` variant and `ChatListUpdateTriggerFfi::PinOrderChanged` are source-breaking for exhaustive Swift `switch` / Kotlin `when`, and `ChatListRowFfi` gains two required fields — hosts will not compile until they add the arms. Worth one line so the 0.9.9 upgrade is not a surprise.

2. `wn-cli` has no pin surface: `insert_chat_projection` does not emit `pinned` / `pinned_position`, and `chat_rows_json` uses `chat_list()` only as a lookup index, so CLI ordering is unaffected by pins. That is internally consistent and nothing is broken — but a pin set from MarmotKit on the same account-device database is invisible in the CLI and TUI. Reasonable as a follow-up issue rather than scope for this PR.

Looks good to merge from my side.

PR #1150, opened 2026-07-28
Implement terminal MLS group disbanding

erskingardner:
## Summary

- implement lifecycle-v1 (`0x800c`) with `Active` and terminal `Disbanded` state
- add durable disband requests, authenticated convergence candidates, terminal tombstones, and migration `0040`
- validate the exact disband Commit shape, including candidate-parent admin authorization, exact committer-leaf retention, sibling-device removal, admin-policy replacement, and forbidden proposals
- route local and inbound disband Commits through normal bounded convergence without special branch dominance
- terminalize transactionally, erase live OpenMLS state, clean routes/queues/fanouts/maintenance/push state, and hydrate safely from tombstones
- expose lifecycle and request state through app/runtime/chat-list/group-details/management DTOs and UniFFI APIs
- synthesize one actor-attributed `group_disbanded` kind-1210 row and retain history/media material until local deletion
- preserve the newly landed manual chat pin ordering while composing terminal lifecycle projection into the same chat-list query

## Protocol dependency

Implements marmot-protocol/marmot#409.

## Compatibility

New current-profile groups require lifecycle-v1 and start `Active`. Existing groups remain valid without it and can be upgraded only when every current leaf advertises support. No OpenMLS fork is required.

## Validation

- `just fast-ci`
- `cargo test -p storage-sqlite` — 295 passed
- `cargo test -p cgka-engine --test group_creation solo_disband_is_durable_convergent_terminal_and_restart_safe -- --exact`
- `cargo test -p marmot-uniffi disband_gate_and_terminal_state_disable_every_management_action`
- app library, connector, lifecycle codec, canonical simulator fixture, and focused terminal/storage tests passed before the final rebase

Full `just ci` was not run locally.

## Summary by CodeRabbit

* **New Features**
  * Added durable group disbanding with lifecycle-v1 enablement, disband requests/candidates/tombstones, support blockers, failure acknowledgement, and a terminal “disbanded” state.
  * Exposed new engine/session/runtime/UniFFI APIs and surfaced disbanding state in group views (details, management, chat list), including disband lifecycle metadata and gating.
* **Bug Fixes**
  * Blocked outbound work, optimistic messaging, convergence draining/publishing side effects, unread counts, routing updates, and local deletion while disbanding or after terminal disband.
  * Added clearer event/audit and error labeling for disband-related outcomes and reconciliation behavior after restarts.
* **Tests**
  * Added/updated conformance and integration coverage for durable disbanding, hydration, and atomic failure recovery.

### DIFF
--- a/crates/cgka-engine/src/app_components.rs
+++ b/crates/cgka-engine/src/app_components.rs
@@ -7,11 +7,13 @@ use cgka_traits::app_components::{
     AppComponentId, AppComponentSet, GROUP_ADMIN_POLICY_COMPONENT_ID,
     GROUP_AVATAR_URL_COMPONENT_ID, GROUP_BLOSSOM_IMAGE_COMPONENT_ID,
     GROUP_ENCRYPTED_MEDIA_V1_COMPONENT_ID, GROUP_ENCRYPTED_MEDIA_V2_COMPONENT_ID,
-    GROUP_MESSAGE_RETENTION_COMPONENT_ID, GROUP_PROFILE_COMPONENT_ID, GroupProfileV1,
-    NOSTR_ROUTING_COMPONENT_ID, NostrRoutingV1, SAFE_AAD_COMPONENT_ID, decode_components_list,
-    decode_encrypted_media_policy_v1, decode_encrypted_media_policy_v2, decode_group_avatar_url_v1,
-    decode_group_blossom_image_v1, decode_group_profile_v1, decode_nostr_routing_v1,
-    decode_quic_varint, encode_component_vectors, encode_components_list, encode_group_profile_v1,
+    GROUP_LIFECYCLE_COMPONENT_ID, GROUP_MESSAGE_RETENTION_COMPONENT_ID, GROUP_PROFILE_COMPONENT_ID,
+    GroupLifecycleV1, GroupProfileV1, NOSTR_ROUTING_COMPONENT_ID, NostrRoutingV1,
+    SAFE_AAD_COMPONENT_ID, decode_components_list, decode_encrypted_media_policy_v1,
+    decode_encrypted_media_policy_v2, decode_group_avatar_url_v1, decode_group_blossom_image_v1,
+    decode_group_lifecycle_v1, decode_group_profile_v1, decode_nostr_routing_v1,
+    decode_quic_varint, encode_component_vectors, encode_components_list,
+    encode_group_lifecycle_v1, encode_group_profile_v1,
 };
 use cgka_traits::engine::CommitOrderingPriority;
 use cgka_traits::error::EngineError;
@@ -24,8 +26,8 @@ use openmls::group::{
 };
 use openmls::messages::proposals::{AppDataUpdateOperation, AppDataUpdateProposal, Proposal};
 use openmls::prelude::{
-    BasicCredential, ContentType, ExtensionType, LeafNode, LeafNodeIndex, ProposalType,
-    ProposalValidationError, QueuedProposal, Sender,
+    BasicCredential, ContentType, ExtensionType, LeafNode, LeafNodeIndex, ProposalOrRefType,
+    ProposalType, ProposalValidationError, QueuedProposal, Sender,
 };
 use openmls::treesync::Node;
 use std::collections::{BTreeMap, BTreeSet};
@@ -125,6 +127,12 @@ pub(crate) fn app_data_dictionary_extension_for_group(
             encode_admin_policy(&initial.admins)?,
         );
     }
+    if required.contains(GROUP_LIFECYCLE_COMPONENT_ID) {
+        dict.insert(
+            GROUP_LIFECYCLE_COMPONENT_ID,
+            encode_group_lifecycle_v1(GroupLifecycleV1::Active),
+        );
+    }
     let mut seen_initial = BTreeSet::new();
     for component in &initial.app_components {
         if !seen_initial.insert(component.component_id) {
@@ -182,6 +190,17 @@ pub(crate) fn admins_of_group(mls_group: &MlsGroup) -> Result<Vec<[u8; 32]>, Eng
         return Ok(Vec::new());
     };
     decode_admin_policy(bytes)
+}
+
+pub(crate) fn lifecycle_of_group(
+    mls_group: &MlsGroup,
+) -> Result<Option<GroupLifecycleV1>, EngineError> {
+    let Some(bytes) = app_component_bytes(mls_group, GROUP_LIFECYCLE_COMPONENT_ID) else {
+        return Ok(None);
+    };
+    decode_group_lifecycle_v1(bytes).map(Some).map_err(|error| {
+        EngineError::Serialize(format!("invalid group lifecycle component: {error}"))
+    })
 }
 
 pub(crate) fn app_component_data_of_group(
@@ -431,6 +450,13 @@ fn authorize_proposal(
             if !sender_is_admin {
                 return Err(StandaloneProposalRejection::authorization(
                     "app_data_update_sender_not_admin",
+                ));
+            }
+            if update.component_id() == GROUP_LIFECYCLE_COMPONENT_ID
+                && proposal.proposal_or_ref_type() == ProposalOrRefType::Reference
+            {
+                return Err(StandaloneProposalRejection::invalid(
+                    "group_lifecycle_update_must_be_inline",
                 ));
             }
             if validate_app_data_payload {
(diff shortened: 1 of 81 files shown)


crates/cgka-engine/src/app_components.rs
```diff
@@ -798,7 +824,193 @@ pub(crate) fn validate_current_profile_invariants_for_staged_commit(
...
+                Proposal::AppDataUpdate(update)
+                    if update.component_id() == GROUP_LIFECYCLE_COMPONENT_ID
+            )
+        }) {
+            return Err(EngineError::Other(
+                "commit is invalid: redundant lifecycle update".into(),
+            ));
+        }
+        return Ok(());
+    }
```
erskingardner, line 933: **Blocking:** `0x800c` can be silently un-required without ever changing the lifecycle value byte. This branch only runs when `before != Some(Active) || after != Some(Disbanded)`. When `before == after == Active` but a commit's `AppDataUpdate` on `APP_COMPONENTS_COMPONENT_ID` drops `0x800c` from the required list, neither this function nor `validate_app_component_integrity_for_staged_commit` (which protects only the *resulting* required set) rejects it. Spec (group-lifecycle-v1.md): "Once required, this component MUST remain present and required for the remainder of the group's lifetime." Suggest rejecting any transition where `before_required && !after_required`, independent of the value-byte branch.

crates/marmot-app/src/client/mod.rs
```diff
@@ -1024,13 +1058,108 @@ impl AppClient {
...
     /// an MLS leave and does not delete the stored MLS/OpenMLS group state; a
     /// future fresh group delivery can recreate the chat-list projection.
     ///
     /// The live transport route is removed and synced before the DB wipe so the
     /// account stops actively subscribing to the group before local rows vanish.
     pub async fn delete_group_local(&mut self, group_id: &GroupId) -> Result<bool, AppError> {
+        if self
+            .runtime
+            .disband_request(group_id)?
+            .is_some_and(|request| request.status == cgka_traits::DisbandRequestStatus::Pending)
```
erskingardner: **Blocking:** this only guards on a *local* `disband_request(...).status == Pending`, not `disbanding_in_progress()` (used correctly a bit further down in `ensure_group_application_messages_allowed`, which also covers an authenticated inbound disband candidate from another admin/device). If another admin's disband Commit is mid-convergence on this device but this device never created its own `DisbandRequest`, this check passes and `delete_group_local` tears down the transport route and wipes rows while a terminal commit is still in flight — contradicting the function doc-comment's own reasoning about not discarding routes ahead of durable resolution.

crates/marmot-uniffi/src/commands/group.rs
```diff
@@ -333,6 +344,49 @@ impl Marmot {
...
+
+    /// Durably accept an irreversible disband request. Completion is observed
+    /// through normal group state updates after bounded convergence.
+    pub async fn disband_group(
+        &self,
+        account_ref: String,
+        group_id_hex: String,
+    ) -> Result<DisbandRequestFfi, MarmotKitError> {
+        let group_id = group_id_from_hex(&group_id_hex)?;
+        let request = self.runtime.disband_group(&account_ref, &group_id).await?;
```
erskingardner, line 378: Non-blocking: `disband_group` skips the `group_management_state_for` + `ensure_group_admin` precheck used by every other admin-gated command in this file (`enable_group_disbanding` just above, `invite_members`, etc.). `do_request_disband` still enforces the admin check server-side so this isn't a privilege-escalation bug, just an inconsistent API shape worth aligning.

crates/cgka-engine/src/disband.rs
```diff
@@ -0,0 +1,635 @@
...
+        let mls_gid = openmls::group::GroupId::from_slice(group_id.as_slice());
+        let mls_group = MlsGroup::load(provider.storage(), &mls_gid)
+            .map_err(|error| EngineError::Backend(format!("load: {error:?}")))?
+            .ok_or_else(|| EngineError::UnknownGroup(group_id.clone()))?;
+        let required = crate::app_components::required_app_components_of_group(&mls_group)?;
+        if !required.contains(GROUP_LIFECYCLE_COMPONENT_ID)
+            || crate::app_components::lifecycle_of_group(&mls_group)?
+                != Some(GroupLifecycleV1::Active)
+        {
+            return Err(EngineError::Other("group disbanding is not enabled".into()));
```
erskingardner: Non-blocking: this is `EngineError::Other`, a stringly-typed error, which is why `MarmotKitError::from_engine_error` has to fall through its catch-all `Runtime` bucket for it instead of a typed variant — inconsistent with the typed treatment `NotGroupAdmin` and `DisbandingUnsupportedMembers` get right next to it. Worth a dedicated `EngineError` variant so host apps can branch on "disbanding not enabled" instead of string-matching `Runtime.details`.

erskingardner (commented):
This is a large, well-organized implementation of `marmot.group.lifecycle.v1` / terminal disbanding against marmot-protocol/marmot#409. Convergence-forcing, terminalization ordering, and the core disband Commit-shape validation are solid and match the spec closely (verified `require_admin_for_staged_commit`, exact-leaf-removal, forbidden-proposal exhaustiveness, single-leaf zero-Remove case, and the transactional `settle_disband_after_convergence`). Two confirmed correctness gaps should be fixed before merge; the rest are smaller consistency/coverage notes.

## Blocking

1. **`0x800c` can be silently un-required without a value change** (`crates/cgka-engine/src/app_components.rs`) — `validate_group_lifecycle_transition` only guards `before_required`/`after_required` inside the exact `Active -> Disbanded` branch. When the lifecycle *value* is untouched (still `Active`) but a commit's `AppDataUpdate` on `APP_COMPONENTS_COMPONENT_ID` drops `0x800c` from the required list, this function returns `Ok(())`, and `validate_app_component_integrity_for_staged_commit`'s "protected" set is built from the *resulting* required list, so it doesn't catch a component that's no longer required. This violates the spec's "Once required, this component MUST remain present and required for the remainder of the group's lifetime" and permanently defeats the disband guarantee for that group with no test catching it.

2. **`delete_group_local` ignores inbound disband candidates** (`crates/marmot-app/src/client/mod.rs`) — the guard only checks a *local* `DisbandRequest.status == Pending`, not `disbanding_in_progress()` (which the sibling `ensure_group_application_messages_allowed` correctly uses a few hundred lines away, and which also covers an authenticated inbound candidate from another admin). If another admin's disband Commit is mid-convergence on this device but this device never made its own request, a local wipe can tear down the route and delete rows while a terminal commit is still in flight.

## Non-blocking

- `crates/marmot-uniffi/src/commands/group.rs::disband_group` skips the `group_management_state_for` + `ensure_group_admin` precheck that every sibling admin-gated command in this file uses (e.g. `enable_group_disbanding`, `invite_members`). Not a privilege-escalation bug — `do_request_disband` still enforces the check server-side — just an inconsistent API shape.
- The "disbanding is not enabled" precondition (`crates/cgka-engine/src/disband.rs:110`, `EngineError::Other(...)`) falls through `MarmotKitError::from_engine_error`'s catch-all `Runtime` bucket instead of getting a typed variant, unlike the neighboring `DisbandingUnsupportedMembers`/`NotGroupAdmin` cases this same file goes out of its way to type.
- `crates/storage-sqlite/src/migrations.rs`: the existing `group_owned_tables_have_cascading_foreign_keys` convention test wasn't extended to cover the three new `cgka_disband_*` tables (including asserting `cgka_disband_tombstones` deliberately has no cascading FK, which is the anti-resurrection guarantee).
- `crates/storage-sqlite/src/chat_list.rs`: tombstone-exists lookups compare `lower(hex(tomb.group_id)) = lower(...)`, defeating the BLOB primary-key index on a column that's rare today but will scale badly. Also no test exercises "pin a group, then disband it," though I didn't find an actual ordering bug there.
- Minor duplication (Karpathy: prefer fewer lines) — `do_request_disband`, `record_inbound_disband_candidate`, and `settle_disband_after_convergence` in `crates/cgka-engine/src/disband.rs` each repeat near-identical cleanup blocks for queued intents/welcomes/fanouts/maintenance; worth a shared helper.
- Test coverage gap: no multi-member/multi-device disband test (sibling-leaf removal), and no test for a standalone/by-reference lifecycle or admin-policy proposal being rejected in a disband commit.

Nice work on the transactional terminalization and privacy-safe logging — no group/account/pubkey identifiers leak into any new tracing call I checked.

crates/cgka-engine/src/app_components.rs
erskingardner: Fixed in b00988e2. Lifecycle-v1 is now rejected whenever it changes from required to unrequired, independent of the lifecycle value transition, with a current-profile regression test.

crates/marmot-uniffi/src/commands/group.rs
erskingardner: Fixed in b00988e2. The UniFFI disband_group command now performs the same management-state/admin preflight as the sibling admin mutations.

crates/cgka-engine/src/engine.rs
```diff
@@ -987,14 +999,50 @@ impl<S: StorageProvider> Engine<S> {
...
         group_id: &GroupId,
     ) -> Result<EpochId, GroupHydrationQuarantineReason> {
+        let group = self
+            .storage
+            .get_group(group_id)
+            .map_err(|_| GroupHydrationQuarantineReason::GroupRecordLoadFailed)?;
+        let tombstone = group
+            .disbanded
+            .clone()
+            .or_else(|| self.storage.disband_tombstone(group_id).ok().flatten());
```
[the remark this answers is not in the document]
erskingardner: Fixed in b00988e2. Tombstone read errors now quarantine the group instead of being treated as absence, with a fault-injection hydration/retry test.

erskingardner (commented):
Follow-up verification pass. Both previously-flagged blocking bugs are genuinely fixed in `b00988e2`, each with a dedicated regression test:

1. **Lifecycle-v1 un-requiring gap** (`crates/cgka-engine/src/app_components.rs:881-885`) — `validate_group_lifecycle_transition` now rejects `before_required && !after_required` unconditionally, before the value-byte branches, independent of whether the lifecycle value itself changes. Covered by the new `required_lifecycle_component_cannot_be_unrequired` test in `crates/cgka-engine/tests/update_group_data.rs`.
2. **`delete_group_local` inbound-candidate gap** — both entry points now gate on the broader in-progress signal instead of a local-only `DisbandRequest`: `AppClient::delete_group_local` (`crates/marmot-app/src/client/mod.rs:1157`) checks `self.runtime.disbanding_in_progress(group_id)?`, and `MarmotApp::delete_group_local_data` (`crates/marmot-app/src/lib.rs:3628-3633`) checks `storage.disbanding_group_ids_hex()?`. Covered by the new `inbound_disband_candidate_blocks_both_local_delete_entry_points` test in `crates/marmot-app/src/tests.rs`, which seeds only an inbound `DisbandCandidate` (no local request) and asserts both entry points reject.

I also re-verified essentially every other actionable CodeRabbit/reviewer finding from the prior round and they're all resolved as well: `disbanding_support_blockers`/`epoch_state` now call `ensure_group_live`; the hydration tombstone-read failure quarantines instead of silently treating storage errors as "not disbanded"; `restore_disbanded`/`restore_disband_tombstone` propagate settlement errors instead of failing open; tombstoned groups are gated out of the ingest fallback path before they can be retained as retryable unknown-group input; the disband rollback's candidate-clear/request-rewrite now lands inside the same durable transaction as the rest of the rollback, before the pending slot is consumed; `enable_group_disbanding`/`disband_group` now call `schedule_audit_log_tracker_update`; the UniFFI `disband_group` command now runs the same admin preflight as its siblings; `list_disband_tombstones` is a required (non-default) trait method; the hex-decode in `delete_local_group_data` is hoisted above the transaction; `GroupDisbanded` event material now requires an actor; and a dedicated `EngineError::DisbandingNotEnabled` typed variant replaced the stringly-typed `Other`.

Only trivial, previously-noted non-blocking items remain outstanding (not worth holding this up):
- `crates/storage-sqlite/src/chat_list.rs:382,1523` — the tombstone-exists join still compares `lower(hex(tomb.group_id)) = lower(row.group_id_hex)`, defeating the BLOB primary-key index. Fine at current scale, will want an index-friendly comparison eventually.
- The disband-cleanup duplication across `do_request_disband`/`record_inbound_disband_candidate`/`settle_disband_after_convergence` in `crates/cgka-engine/src/disband.rs` noted last round is still unconsolidated.

Nice work closing out both correctness gaps with targeted regression tests rather than just patching the reported line.

PR #1152, opened 2026-07-28
Add rich agent inbound context and stable host baselines

### LINKED ISSUE #1147: Agent connectors: expose replies, reactions, and chat projection changes as normalized inbound events
## Problem

The Marmot app builds a clean conversation projection by folding chat messages, replies, reactions, deletes, and group-state events together. That projection is ideal for UI rendering, but the OpenClaw connector needs the semantic changes that produced it.

The current `subscribe_inbound` control stream already emits:

- `inbound_message`
- `message_deleted`
- `group_state_changed`
- group invites, stream updates, and resync signals

An inbound message carries `reply_to_message_id_hex`, but not the replied-to message snapshot. Reaction add/remove events are not projected into the agent-control stream at all. In practice:

- inline emoji text reaches OpenClaw;
- a reply reaches OpenClaw with only its target message ID;
- reactions attached to a message do not reach OpenClaw.

## OpenClaw integration research

OpenClaw's channel receive path is explicitly:

```text
platform event -> inbound facts/context -> agent reply -> message delivery
```

Its channel SDK expects plugins to normalize individual platform events and pass those facts through the inbound runner. OpenClaw then handles ingest, classification, routing, session recording, dispatch, finalization, dedupe, and durable receive semantics.

Relevant upstream documentation:

- [Channel inbound API](https://github.com/openclaw/openclaw/blob/main/docs/plugins/sdk-channel-inbound.md) — `buildChannelInboundEventContext`, `runChannelInboundEvent`, and `dispatchChannelInboundReply`
- [Messages](https://github.com/openclaw/openclaw/blob/main/docs/concepts/messages.md) — inbound dedupe/debounce, `BodyForAgent` / `BodyForCommands`, pending group history, and structured reply/channel metadata
- [Message lifecycle refactor](https://github.com/openclaw/openclaw/blob/main/docs/concepts/message-lifecycle-refactor.md) — durable inbound queue/journal, receive acknowledgement stages, and message receipts
- [Ambient room events](https://github.com/openclaw/openclaw/blob/main/docs/channels/ambient-room-events.md) — quiet group context versus events that trigger a visible agent turn

Current Marmot implementation:

- [`crates/agent-connector/src/event_projection.rs`](https://github.com/marmot-protocol/mdk/blob/master/crates/agent-connector/src/event_projection.rs) projects runtime/storage facts into agent-control events and extracts reply target IDs.
- [`crates/agent-connector/src/inbound.rs`](https://github.com/marmot-protocol/mdk/blob/master/crates/agent-connector/src/inbound.rs) streams those events with catch-up, bounded replay, deduplication, and `resync_required`.
- [`integrations/openclaw/marmot/src/client.ts`](https://github.com/marmot-protocol/mdk/blob/master/integrations/openclaw/marmot/src/client.ts) defines the TypeScript control-event contract.
- [`integrations/openclaw/marmot/src/inbound.ts`](https://github.com/marmot-protocol/mdk/blob/master/integrations/openclaw/marmot/src/inbound.ts) maps `inbound_message` into OpenClaw's inbound context.

## Proposed direction

Keep the app's aggregated conversation projection, but expose projection changes to agent connectors as a normalized, cursorable event stream.

At minimum, model:

- message created
- message edited
- message deleted
- reaction added
- reaction removed
- group state changed

Every event should carry stable account, group, sender, message, and target IDs as applicable. Preserve the existing replay/idempotency guarantees and ensure every new durable event type can be recovered after reconnect or broadcast lag.

For a message that replies to another message, provide one of:

1. a bounded `reply_to` snapshot containing target message ID, sender identity, text excerpt, and attachment summary; or
2. a connector query that hydrates the target deterministically before OpenClaw dispatch.

The bounded snapshot is probably the better default because OpenClaw can render structured reply metadata directly into untrusted user-role context without adding a synchronous lookup to the hot inbound path. A query API is still useful for resync and deeper history hydration.

Reaction events should remain distinct semantic deltas. The OpenClaw adapter can then apply policy:

- retain them as quiet conversation context;
- surface selected reactions as lightweight notifications;
- trigger a user-request turn only where meaningful/configured.

Do not flatten reactions into synthetic human text inside MDK. That decision belongs at the OpenClaw adapter boundary.

## Acceptance criteria

- [ ] Reply messages give the OpenClaw model enough bounded context to understand what was replied to, not only an opaque target ID.
- [ ] Reaction add and remove changes are available through the agent-control inbound stream.
- [ ] Message edits are represented, or explicitly documented as unsupported with a follow-up issue.
- [ ] New event types have stable IDs, account/group scoping, sender attribution, and privacy-safe validation.
- [ ] Reconnect, catch-up, lag replay, deduplication, and `resync_required` behavior cover the new durable event types.
- [ ] The OpenClaw adapter maps message events into the channel inbound API and does not bypass OpenClaw's routing/session/durable-receive lifecycle.
- [ ] Reaction-only and other ambient changes do not automatically force visible replies.
- [ ] Tests cover reply hydration, reaction add/remove, duplicate delivery, reconnect replay, multi-account isolation, and malformed/missing targets.
- [ ] The agent-control README and OpenClaw integration README document the event contract and adapter policy.

## Design question

Should `subscribe_inbound` carry only deltas plus a cursor/query API, or should message-created events embed a bounded projection snapshot? A hybrid looks strongest: semantic deltas for normal operation, bounded reply snapshots for immediate comprehension, and projection queries for recovery/hydration.

erskingardner:
## Summary

Resolve #1147 with one atomic update to every in-repo `marmot.agent-control.v2` consumer.

- replace the flat inbound message shape with structured actor, message, media, and bounded reply-context DTOs
- add durable normalized edit, delete, reaction-add, and reaction-remove events
- hydrate live and replayed events from durable rows with privacy redaction, mutation authorship/moderation filtering, cursor ordering, and event-id deduplication
- deliver replies and ambient mutations through OpenClaw and Hermes native user-context surfaces without starting turns for mutations
- update OpenCode to decode the new schema while intentionally ignoring ambient mutations
- add shared Rust, TypeScript, Python, and OpenCode golden fixtures
- pin the OpenClaw development SDK to `openclaw@2026.7.1-2`
- lock Hermes to `0.19.0`, tag `v2026.7.20`, commit `3ef6bbd201263d354fd83ec55b3c306ded2eb72a`
- make the Hermes lock the default for dev setup, CI, and the phone-test image while retaining explicit overrides
- validate existing OpenClaw and Hermes hosts in production installers without installing or upgrading them

This deliberately keeps the `marmot.agent-control.v2` label and current component versions. The contract documentation records this as the final unversioned wire break; future breaking changes require a new protocol label or explicit negotiation.

## Platform behavior

OpenClaw continues through `buildChannelInboundEventContext` and `runChannelInboundEvent`, maps authenticated seconds to native millisecond timestamps, projects replies into `supplemental.quote`, and adds bounded attachment/mutation facts to structured `supplemental.untrustedContext` on the next triggering message.

Hermes populates stable `MessageEvent` timestamp/reply/media fields directly and attaches bounded quote and mutation facts to the next triggering message through user-role `channel_context`.

The published OpenClaw `2026.7.1-2` declarations expose `ChannelMessagingAdapter`, `ChatType`, and `jsonResult` from `plugin-sdk/core`, rather than `channel-core`; the implementation uses that compile-proven stable export surface. `ChannelGatewayContext` uses `channel-contract` as planned.

## Validation

- `just fast-ci`
- `cargo test -p agent-control` — 16 passed
- `cargo test -p agent-connector` — 129 passed
- `cargo test --manifest-path integrations/opencode/marmot/Cargo.toml` — 49 passed, 1 process E2E ignored by default
- OpenClaw frozen install, typecheck, build, unit tests — 188 passed, 1 skipped
- OpenClaw dev-script and installer contract tests
- Hermes unit tests — 158 passed
- Hermes installer, persisted-config, authorization, and dev-script tests
- exact Hermes stable checkout/import probe
- exact Hermes stable deterministic adapter E2E
- exact Hermes stable real `wn-agent` connector E2E
- `cargo fmt --check`
- `git diff --check`

The full `just ci` matrix is left to GitHub CI.


erskingardner (commented):
## Review: rich agent inbound context

Reviewed the full 51-file diff: agent-control DTOs, connector projection/inbound, storage-sqlite, marmot-app, plus the OpenClaw (TS), Hermes (Python), and infra/script changes.

The DTO reshape (`AgentControlMessage` / `AgentControlActor` / `AgentControlReferencedMessage`) is a clear improvement, and hydrating references from durable rows rather than from tags is the right call — deleted/invalidated targets correctly retain attribution only, and the bounded text/attachment caps are applied consistently. The new `SqliteAccountStorage::app_message` lookup is index-backed by the existing `UNIQUE (group_id_hex, message_id_hex)` on `app_events`, so the added per-message read is cheap.

Three things I think should be resolved before merge, then some smaller notes.

---

### 1. A transient storage error now tears down the entire inbound subscription

`crates/agent-connector/src/inbound.rs` (runtime-event arm of the `select!`):

```rust
control_event_from_runtime_event_with_runtime(
    &self.runtime,
    event,
    account_id_hex.as_deref(),
    group_id_hex.as_deref(),
)
.map_err(ConnectorError::from)?
```

Projection is now fallible — `referenced_message` and `available_target_context` each do one or two `runtime.message_by_id` / `message_target` reads — and this `?` propagates straight out of `stream_inbound_events`, dropping the client's subscription.

That is a behavior regression against the design the surrounding code goes out of its way to establish. Every other failure mode in this same `select!` degrades gracefully:

- lag beyond the replay window → `resync_required`
- lag within the window → storage replay, and *if that replay itself fails* → `resync_required` with the comment "so the agent still learns it must re-sync state rather than silently losing the dropped messages"

So a storage error during lag recovery is handled softly, but the same storage error during ordinary hydration now kills the stream. One busy/locked SQLite moment disconnects the agent. (The sibling `?` inside `replay_missed_inbound` is fine — that call site already has the `resync_required` fallback.)

Suggested fix: keep this arm infallible from the caller's perspective — log with `err.privacy_safe_code()` and either skip the event or emit `resync_required_event(...)`, matching the adjacent arms.

### 2. Ambient mutation context is dropped by non-triggering messages (OpenClaw)

`integrations/openclaw/marmot/src/inbound-runtime.ts:316-326` pulls **and clears** `pendingAmbient[account:group]`, attaching it to whatever `inbound_message` event arrives next. The decision about whether that message actually starts a turn happens later, in `shouldRunTurn` inside `src/dispatch.ts:362-367`.

In a `groupActivation: "mention"` group — the common config — an unaddressed message steals and clears the buffered batch, then gets gated out at line 366 before `inboundSupplemental()` is ever reached. The reactions/edits are unrecoverable; the next `@agent` message carries no trace of them. The same happens when the fail-closed `groupInfo` lookup errors at line 194-196.

This contradicts the README added in this PR, which says ambient facts are "attached only to the next triggering user turn" — the code attaches to the next message *event*, not the next *triggering* one.

Suggested fix: flush `pendingAmbient` only after `shouldRunTurn` passes, or re-queue the batch when the gate skips.

No test covers this composition: `dispatch.test.ts` unit-tests the dispatcher with `ambientContext` pre-attached, and `inbound-runtime.test.ts`'s buffering test uses a bare stub dispatcher — the gate and the buffer are never exercised together.

### 3. The `#[cfg(test)]` projection shims mean the unit tests no longer test production code

`crates/agent-connector/src/event_projection.rs` keeps `control_event_from_runtime_event` and `inbound_message_event_from_record` as `#[cfg(test)]` **reimplementations** rather than as thin wrappers over the `*_with_runtime` functions. They are separate bodies with independently drifting logic — e.g. the shim collapses the group-system branch to `record.kind == GROUP_SYSTEM && record.direction == "system"` with a hoisted `record.invalidated` check, while the real `project_durable_record` structures both conditions differently.

Every remaining test that calls the short names (`control_event_forwards_only_chat_inner_events_as_inbound_messages`, `control_event_projects_imeta_tag_into_inbound_media_ref`, `inbound_message_event_from_record_projects_received_chat`, `..._extracts_mention_and_reply`, `..._skips_non_inbound_and_own_and_filtered`) now validates a code path production never executes. These can stay green while the real projection is broken.

Two of the rewritten tests make this concrete: `control_event_does_not_project_unhydrated_kind5_deletion` and `inbound_message_event_from_record_projects_received_delete` now assert `is_none()` — they assert what the *shim* does, not what production does, and no longer cover the delete projection they were written for. Real delete coverage survives only in the socket test.

Suggested fix: drop the shims and move these to runtime-backed tests, or at minimum have the shims delegate to the real functions with a stub runtime.

---

### Medium

**4. Unbounded ambient buffers in both integrations.** `_pending_ambient_context` (`integrations/hermes/marmot/adapter.py:1612`, appended at `:2947`) and `pendingAmbient` (`inbound-runtime.ts:239`) are uncapped per-group lists. `_recent_ambient_keys` only dedupes literal re-delivery of the same `event_id_hex`; it does not bound distinct mutations. A group that generates many reactions/edits before the next real message produces one very large `channel_context` string sent to the LLM, and a group that goes permanently quiet never releases its buffer. Note the `_mutation_channel_context` docstring's "bounded" refers to each individual fact's size (`replacement_text[:2000]`), not the count. Worth a max-length ring per key. Finding 2 makes this worse in practice on the OpenClaw side.

**5. No production test exercises the three new event kinds.** `MessageEdited`, `ReactionAdded`, and `ReactionRemoved` have no test at all through `project_durable_record`; the connector socket test covers `InboundMessage` (with `reply_to` hydration) and `MessageDeleted` only. Same on both integrations: the golden tests decode all 8 fixture events at the wire layer, but nothing pushes an edit or reaction removal through `MarmotInboundBridge` → `startMarmotInbound` → dispatcher, and the Python golden test never calls `_mutation_channel_context` or `_referenced_channel_context` at all. These are the main new code paths in the PR.

**6. Live inbound messages can be silently dropped when the durable row is already gone.** `control_event_from_runtime_event_with_runtime` now requires `runtime.message_by_id(...)` to return `Some` and returns `Ok(None)` otherwise. Previously the event was projected directly from the in-memory `update.message`, so no durable read could miss.

Ordering is normally safe (commit-then-broadcast: `client/sync.rs:671` writes, `account_worker.rs:1995` broadcasts). But `client/sync.rs:693` calls `prune_plaintext_retention_for_group` *after* the insert and *before* the broadcast, and that path hard-deletes rows. A received message whose retention deadline has already passed at ingest is inserted, pruned, then still broadcast — and now silently vanishes from the agent feed instead of being delivered. Same for a row invalidated by convergence between insert and projection. Could you confirm whether that's intended? If so, a comment at the `else { return Ok(None) }` would help; if not, falling back to the in-memory `update.message` when the row is missing preserves the old behavior.

### Minor

7. `_coalesce_inbound_events` (`adapter.py:329-373`) wasn't updated for the new `reply_to` dict — it still special-cases only `reply_to_message_id_hex` at line 372, so `merged["reply_to"]` is whatever the last debounced item carried. With `debounce_ms > 0`, a replying message followed by a non-replying one keeps the reply id in `raw_message` but drops the rich `reply_to` context.

8. `decodeAgentControlEvent` in `client.ts` validates that `target`/`sender` are objects but not their required sub-fields. Fine given wn-agent is a same-repo trusted producer — worth one line noting the trust assumption.

9. `require_hermes_version` / `require_openclaw_version` are near-identical duplicates across `scripts/install-hermes-marmot.sh` and `scripts/install-openclaw-marmot.sh`. Both fail closed correctly; just a candidate for a shared helper.

10. `_normalize_inbound_message_event` (`adapter.py:376-393`) uses raw `[...]` indexing for `message_id_hex` / `account_id_hex` where the surrounding code raises `AgentControlError` — inconsistent, low risk over a trusted local socket.

---

### Checked, no issues

- **Reply/quote projection** in `inboundSupplemental` — `available`/`missing`/`deleted`/`invalidated` all handled via `replyFallbackBody` with sensible fallbacks, no null-target crash.
- **Timestamps** — seconds end to end in Rust and Python; OpenClaw does exactly one `* 1000` at `dispatch.ts:386`, verified against the test asserting `1_721_000_000` → `1_721_000_000_000`.
- **Reaction-removal authorization** — restricting removal to the original reaction author, with moderation grants deliberately not applying, matches the materialized timeline. The `moderation_grant` split for chat deletes looks right.
- **`app_message` query** — index-backed by the existing unique constraint; no added scan.
- **SDK import migration** — `plugin-sdk/core` vs `plugin-sdk/channel-contract` split is consistent; the `gateway.ts` -17/+1 deletion is required by the migration, not drive-by.
- **Privacy/logging** — no ids, payloads, or bodies logged on the new paths in any of the three languages; consistent with the observability invariant.
- **Dockerfile / dev-setup ref precedence** — CLI > env > lock is correct in both, and capturing `requested_ref` before sourcing the lock is deliberate and right. The `ci.yml` / `wn-agent-binaries.yml` diffs are exactly the now-stale hardcoded ref removals, with nothing dangling.

erskingardner (commented):
## Re-review of 33f3d6d (`Address rich inbound context review feedback`)

Went back through every point from the previous round against the fix commit. All ten are addressed, and in most cases with a test that would have caught the original bug. Details below, then one new issue introduced by the fix and one nit.

### Previously blocking — all resolved

**1. Subscription teardown on transient storage error** — resolved in `inbound.rs`. The `?` is replaced with a `match` whose `Err` arm logs and emits `resync_required_event(..., 1)`, which is exactly the shape of the two adjacent lag arms. The live boundary is now uniformly fail-soft.

**2. Ambient context dropped by non-triggering messages** — resolved, and the fix is better than what I suggested. Rather than duplicating the gate, `MarmotInboundDispatcher` now returns `boolean`, `shouldRunTurn`'s early return propagates `false`, and `handleInbound` consumes the buffer only when the turn actually ran. Moving the read out of `onMessage` and into the queued `handleInbound` also means a thrown turn retains the facts rather than eating them. `"keeps ambient context when a non-triggering message is gated out"` covers it directly.

**3. `#[cfg(test)]` projection shims** — resolved. Both shims are deleted, `message_mentions_account` / `reply_target_from_tags` / `media_refs_from_tags` are `pub(crate)` for real test reuse, and the affected tests are now `async` and runtime-backed. They test production code again.

**5. No coverage for the three new event kinds** — resolved thoroughly. `replay_missed_inbound_recovers_dropped_messages_and_dedups` now drives real `edit_message` / `react_to_message` / `unreact_from_message` calls and asserts `MessageEdited`, `ReactionAdded`, and `ReactionRemoved` come back through the production projector with the right target and emoji. The forged-modifier negative case (wrong-author edit and wrong-author reaction removal, sent as raw MLS application messages to bypass the local authorship checks) is a better test than I asked for — that's the authorization boundary that actually matters, and it was previously untested in either direction. The `replayed_again.is_empty()` tightening is a real strengthening too.

**6. Silently dropping messages whose durable row is gone** — answered, and the answer is more convincing than my suggested fix. The comment at the `else` correctly points out that falling back to `update.message` would re-expose plaintext that retention pruning or convergence invalidation had already discarded. Reintroducing it would have been a privacy regression to fix a liveness edge. Agreed with the call as made; the comment is the right artifact.

**4, 7, 8, 10** — all resolved. Ambient buffers are capped at 16 facts / 256 groups in both languages with matching constants and eviction warnings. `_coalesce_inbound_events` now prefers the rich `reply_to` dict and falls back to the bare routing id. `validateReferencedMessage` checks `message_id_hex`, the `availability` enum, and both truncation flags, with an explicit comment on what is deliberately left to the trusted producer — plus `account_id_hex` / `group_id_hex` validation on both branches, which I hadn't flagged. `_required_string_field` replaces the raw indexing consistently.

Nice additions beyond what was raised: the golden tests now exercise `_referenced_channel_context` and all four `_mutation_channel_context` branches, and `test_ambient_context_is_bounded_and_survives_failed_turn` covers the retain-on-exception path.

(9, the duplicated `require_*_version` helpers, is untouched — fine, it was a style nit.)

---

### New: the peek/consume split can silently drop never-delivered ambient facts

The fix replaced take-then-dispatch with peek-then-dispatch-then-consume, which is what makes retention-on-failure work. But the consume is positional, and the eviction path mutates the same prefix it is counting on:

- `inbound-runtime.ts` — `appendPendingAmbient` does `pending.shift()` at the cap; `handleInbound` later does `current.splice(0, pendingCount)`.
- `adapter.py` — `_append_pending_ambient_context` does `pending.pop(0)` at the cap; `_consume_pending_ambient_context` later does `del pending[:count]`.

`onAmbientEvent` / `_surface_ambient_context` are not inside the per-group dispatch queue, so they run while a turn is in flight. If the buffer hits the cap during that turn, `shift`/`pop(0)` evicts items from the front — the very prefix that was attached to the running turn — and the trailing `splice(0, pendingCount)` then removes `pendingCount` items counted from the new front, eating newly-arrived facts that were never delivered to anything.

Concretely, with `pendingCount = 16` (already at cap) attached to a turn and one reaction arriving mid-turn: the append shifts off `p1` and pushes `n1`, leaving 16; the turn succeeds; `splice(0, 16)` removes `p2..p16` **and `n1`**. `n1` was never in any `ambientContext` and is now gone, with no warning — the eviction warning that did fire was about `p1`, which had already been delivered. Generally, `(pendingCount + arrivalsDuringTurn) - 16` new facts are lost this way. An LLM turn is exactly the kind of long window where a busy group accumulates 16 reactions.

Bounded and only reachable under overflow, so not a merge blocker — but it silently discards undelivered facts, which is the failure mode finding 2 was about, just relocated.

Cleanest fix is to make attach-and-detach atomic and put the batch back on failure, so eviction during the turn can only ever touch genuinely new items:

```ts
const pending = pendingAmbient.get(key);
const attached = pending ? pending.splice(0, pending.length) : [];
if (attached.length === 0) pendingAmbient.delete(key);
// ...dispatch...
if (dispatched === false && attached.length > 0) {
  pendingAmbient.set(key, [...attached, ...(pendingAmbient.get(key) ?? [])]);
}
```

That also keeps the retain-on-throw behavior if the re-prepend happens in a `finally`/`catch`. Same shape applies to `_peek_pending_ambient_context` / `_consume_pending_ambient_context`.

### Nit

`inbound.rs` logs `error_code = "projection_failed"` and binds the error as `_err`, while the adjacent lag arm logs `error_code = err.privacy_safe_code()`. `privacy_safe_code()` exists precisely for this and is already established as safe here, so using it would keep the failure class visible without giving up anything — as written, every hydration failure looks identical in logs.

Minor related thought, not blocking: if storage is failing persistently rather than transiently, this arm emits one `resync_required` per event, which could turn into a reconnect loop on the consumer side. The old code avoided that by disconnecting, which was worse. Might be worth a follow-up to damp repeated projection failures on a single subscription.

---

Everything from the first round is genuinely fixed rather than papered over, and the test additions target the actual regressions. The ambient-consume race above is the only thing I'd still want changed, and it's small.

PR #1157, opened 2026-07-28
Supervise relay notification forwarding and recover inbound delivery

### LINKED ISSUE #1156: Relay notification forwarder can exit silently and permanently stop inbound delivery while sends still work
## Summary

On current Android `master` (`e402b6fd582b98e27970a90cdc201089de55ca1d`) with MDK `master` (`b2919f15c9719399f0d83469db57a00671f948fd`), inbound messages can stop completely while outbound sends continue to succeed. Replies remain absent until the Android process is force-stopped from App info and reopened; startup catch-up then delivers the accumulated backlog.

The relay-plane notification forwarder has a concrete silent-death path matching that asymmetric symptom:

1. `nostr-relay-pool 0.44.0` implements `handle_notifications` as `while let Ok(notification) = notifications.recv().await`. Any broadcast `Lagged` or `Closed` result ends the loop and returns `Ok(())`.
2. `spawn_relay_notification_forwarder` discards that result (`let _ = ...await`) and exits without signalling the relay plane or account workers.
3. The finished task's `JoinHandle` remains stored as `Some(...)`. `spawn_router` only creates a forwarder when the slot is `None`; it never checks `JoinHandle::is_finished()`, awaits the result, or replaces a dead task.
4. The adapter and per-account delivery channels remain open, so `MarmotRelayPlaneAccountAdapter::receive()` and the account worker remain parked indefinitely instead of returning `TransportClosed` and entering the existing reconnect path.
5. Sending bypasses this dead inbound forwarder: `MarmotRelayPlaneAccountAdapter::publish()` calls the SDK publish client directly. Outbound can therefore keep working exactly as observed.
6. A process restart reconstructs the relay plane/forwarder and startup `sync()` replays the missed relay backlog, explaining why force-stop/reopen recovers all prior replies.

Android's chat-list and conversation projection subscriptions already reconnect when their UniFFI streams end. They cannot repair this condition because MDK's inbound event source remains open but silent.

## Source evidence

- `crates/marmot-app/src/relay_plane/mod.rs:560-611`
  - `spawn_router()` treats any `Some(JoinHandle)` as live.
  - The shared delivery router remains parked on the adapter's still-open channel.
- `crates/marmot-app/src/relay_plane/mod.rs:667-792`
  - `spawn_relay_notification_forwarder()` ignores the terminal result from `handle_notifications`.
- `crates/marmot-app/src/relay_plane/mod.rs:980-982`
  - Per-account `receive()` waits on a channel that remains open after the forwarder dies.
- `crates/marmot-app/src/runtime/account_worker.rs:691-793`
  - Reconnect runs only after `receive_next_delivery()` returns an error; a permanently pending receive never reaches it.
- `crates/marmot-app/src/runtime/account_worker.rs:796-868`
  - The periodic tick does not probe inbound-forwarder liveness or run a general catch-up; relay backlog can remain invisible indefinitely.
- `crates/marmot-app/src/relay_plane/mod.rs:934-977`
  - Outbound publish uses the publish client independently of the notification-forwarding task.
- `nostr-relay-pool-0.44.0/src/pool/mod.rs:1317-1332`
  - `handle_notifications()` exits on any notification receiver error because its loop is `while let Ok(...)`.

- #763 covers head-of-line blocking inside the serial relay callback and bounded delivery channel. That blocking can make notification broadcast lag more likely, but it does not track the terminal liveness failure: once `handle_notifications` exits, the completed forwarder is never detected or replaced and inbound delivery remains dead until process restart.
- #741 covers account-worker commands blocking inbound polling and potentially overflowing delivery queues, not silent death of the shared SDK notification consumer.
- #1036 covers already-projected messages omitted from one live batch summary; here new relay notifications stop entering the adapter entirely.
- Android projection-subscription reconnect issues are downstream and cannot observe this open-but-silent source.

## Expected behavior

Inbound delivery must self-heal without restarting the process. A terminal or lagged SDK notification consumer must be observable, replaced, and followed by a bounded catch-up from the durable cursor so no messages are skipped or duplicated. Outbound success must not mask inbound liveness failure.

## Acceptance criteria

- Supervise the SDK notification forwarder and restart it when its task exits or panics; a finished `JoinHandle` must not permanently occupy the slot.
- Do not discard the forwarder's terminal result. Surface a redacted runtime/account diagnostic and liveness state.
- Recover from notification broadcast lag/closure by rebuilding the consumer/subscriptions and running bounded cursor-based catch-up.
- Ensure account-worker receive cannot remain indefinitely parked on an open channel whose only producer is dead; either propagate producer death or add an explicit liveness watchdog.
- Add a regression test that forces the notification consumer to lag or terminate, then proves a later inbound message is projected exactly once without process restart.
- Add an end-to-end account-worker test proving outbound publish can remain available while inbound recovery is exercised, and that backlog appears without force-stop/reopen.

## Verification status

User-observed on current Android `master`; failure mechanism is source-confirmed against current MDK `master` and the vendored `nostr-relay-pool 0.44.0` implementation. No deterministic relay-disconnect reproducer was run locally.

erskingardner:
## Summary

- replace the fire-and-forget SDK notification callback with a supervised consumer that distinguishes shutdown, lag, closure, and panic
- restart failed consumers and close per-account delivery producers so managed account workers enter their existing reconnect path
- force directory subscription rebuilds after notification-pipeline recovery
- expose aggregate, privacy-safe forwarder liveness counters through relay health, UniFFI, CLI, and TUI diagnostics
- add deterministic lag, panic, managed-worker reconnect, bounded replay, outbound-availability, and exactly-once recovery coverage

## Root cause

`nostr-relay-pool`'s `handle_notifications` loop returns successfully whenever its broadcast receiver reports `Lagged` or `Closed`. MDK discarded that terminal result and retained the finished task's `JoinHandle`, while the adapter and account delivery channels remained open. Inbound receivers could therefore stay parked forever even though outbound publishing continued through the independent SDK publish path.

## Recovery behavior

The supervisor now consumes the broadcast receiver directly so lag is observable. On lag, unexpected closure, or consumer panic it records redacted counters, retains buffered notifications when possible, replaces the consumer, closes current account delivery senders, and signals directory recovery. Managed account workers reopen their transport adapters and reinstall subscriptions through the existing durable-cursor lookback path. Normal SDK shutdown remains terminal and does not count as a restart.

## Validation

- `just fast-ci`
- `cargo test -p marmot-app --lib --locked` (474 passed before the final targeted managed-worker regression was added)
- `cargo test -p marmot-app relay_plane --lib --locked` (39 passed)
- `cargo test -p wn-cli relay_stats --locked`
- `cargo test -p wn-cli parse_relay_health --lib --locked`
- `cargo test -p marmot-uniffi conversions::relay --lib --locked`
- `git diff --check`

Closes #1156

## Summary by CodeRabbit

- **New Features**
  - Added notification-forwarder health details to relay health reports, including running status, restarts, lag, panics, and unexpected exits.
  - Added automatic recovery and subscription rebuilding after notification lag, consumer failures, or relay-plane recovery events.
  - Improved notification delivery resilience by restarting failed consumers and replaying missed notifications safely.

- **Bug Fixes**
  - Prevented notification failures and broadcast lag from silently stopping delivery.
  - Ensured recovered account streams resume subscriptions and deliver subsequent notifications correctly.

### DIFF
--- a/crates/marmot-app/src/relay_plane/mod.rs
+++ b/crates/marmot-app/src/relay_plane/mod.rs
@@ -1,5 +1,8 @@
 use std::collections::{HashMap, HashSet};
-use std::sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard};
+use std::sync::{
+    Arc, RwLock, RwLockReadGuard, RwLockWriteGuard,
+    atomic::{AtomicBool, AtomicU64, Ordering},
+};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use async_trait::async_trait;
@@ -75,10 +78,38 @@ struct MarmotRelayPlaneInner {
 struct RelayPlaneTransport {
     adapter: NostrTransportAdapter,
     sdk_relay_client: Option<NostrSdkRelayClient>,
-    directory_events: broadcast::Sender<DirectoryRelayEventRecord>,
+    directory_events: broadcast::Sender<DirectoryRelayPlaneEvent>,
     account_deliveries: RwLock<HashMap<MemberId, mpsc::Sender<TransportDelivery>>>,
     router: Mutex<Option<JoinHandle<()>>>,
     notification_forwarder: Mutex<Option<JoinHandle<()>>>,
+    notification_forwarder_health: Arc<RelayNotificationForwarderHealth>,
+    shutting_down: AtomicBool,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) enum DirectoryRelayPlaneEvent {
+    Record(DirectoryRelayEventRecord),
+    RecoveryRequired,
+}
+
+#[derive(Default)]
+struct RelayNotificationForwarderHealth {
+    running: AtomicBool,
+    restarts: AtomicU64,
+    lag_incidents: AtomicU64,
+    lagged_notifications: AtomicU64,
+    panics: AtomicU64,
+    unexpected_exits: AtomicU64,
+}
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+struct RelayNotificationForwarderHealthSnapshot {
+    running: bool,
+    restarts: u64,
+    lag_incidents: u64,
+    lagged_notifications: u64,
+    panics: u64,
+    unexpected_exits: u64,
 }
 
 #[derive(Clone)]
@@ -103,6 +134,12 @@ pub struct RelayPlaneHealth {
     pub sleeping: usize,
     pub connection_attempts: usize,
     pub connection_successes: usize,
+    pub notification_forwarder_running: bool,
+    pub notification_forwarder_restarts: u64,
+    pub notification_forwarder_lag_incidents: u64,
+    pub notification_forwarder_lagged_notifications: u64,
+    pub notification_forwarder_panics: u64,
+    pub notification_forwarder_unexpected_exits: u64,
     pub directory_inflight_fetches: usize,
     pub directory_active_subscriptions: usize,
     pub directory_completed_fetches: usize,
@@ -195,6 +232,8 @@ impl MarmotRelayPlane {
             account_deliveries: RwLock::new(HashMap::new()),
             router: Mutex::new(None),
             notification_forwarder: Mutex::new(notification_forwarder),
+            notification_forwarder_health: Arc::new(RelayNotificationForwarderHealth::default()),
+            shutting_down: AtomicBool::new(false),
         });
         let this = Self {
             inner: Arc::new(MarmotRelayPlaneInner {
@@ -320,8 +359,17 @@ impl MarmotRelayPlane {
 
     pub async fn relay_health(&self) -> RelayPlaneHealth {
         let directory = self.inner.directory.stats().await;
+        let forwarder = self
+            .inner
+            .transport
+            .notification_forwarder_health
+            .snapshot();
         if let Some(sdk_relay_client) = &self.inner.transport.sdk_relay_client {
-            return RelayPlaneHealth::from_sdk(sdk_relay_client.relay_health().await, directory);
+            return RelayPlaneHealth::from_sdk(
+                sdk_relay_client.relay_health().await,
+                directory,
+                forwarder,
+            );
         }
         RelayPlaneHealth::from_directory(directory)
     }
@@ -408,13 +456,14 @@ impl MarmotRelayPlane {
 
     pub(crate) fn subscribe_directory_events(
         &self,
-    ) -> broadcast::Receiver<DirectoryRelayEventRecord> {
+    ) -> broadcast::Receiver<DirectoryRelayPlaneEvent> {
         self.inner.transport.directory_events.subscribe()
     }
 
     pub(crate) async fn sync_directory_user_subscriptions(
         &self,
         plan: DirectorySyncPlan,
+        force_rebuild: bool,
     ) -> Result<DirectorySubscriptionSyncSummary, String> {
         self.spawn_router();
         let endpoints = self
@@ -461,7 +510,10 @@ impl MarmotRelayPlane {
             .iter()
             .map(|batch| batch.subscription_id.clone())
             .collect::<HashSet<_>>();
-        let (to_add, to_remove) = self.inner.directory.subscription_diff(&desired_ids).await;
+        let (mut to_add, to_remove) = self.inner.directory.subscription_diff(&desired_ids).await;
+        if force_rebuild {
+            to_add = desired_ids;
+        }
         for subscription_id in &to_remove {
             sdk_relay_client
                 .client()
@@ -522,6 +574,10 @@ impl MarmotRelayPlane {
     }
 
     pub async fn shutdown(&self) {
+        self.inner
+            .transport
+            .shutting_down
+            .store(true, Ordering::SeqCst);
         if let Some(sdk_relay_client) = &self.inner.transport.sdk_relay_client {
             let timed_out = timeout(
                 RELAY_PLANE_SHUTDOWN_WAIT,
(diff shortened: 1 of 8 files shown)


crates/marmot-app/src/relay_plane/mod.rs
```diff
@@ -664,134 +749,338 @@ impl RelayPlaneHealth {
...
+                    RelayNotificationForwarderHealth::increment(
+                        &transport.notification_forwarder_health.panics,
+                        u64::from(join_error.is_panic()),
+                    );
+                    receiver = None;
+                    recover_relay_notification_forwarder(
+                        &transport,
+                        RelayNotificationConsumerExit::Closed,
+                    );
+                }
```
erskingardner, line 912: Non-blocking: the `Err(join_error)` (consumer panic/abort) branch respawns the consumer immediately with no backoff, unlike the graceful `Closed` exit a few lines above which sleeps 100ms before retrying. If `handle_relay_notification` (or a future `source.notifications()` implementation) panics deterministically on some input — e.g. a malformed event a relay keeps re-broadcasting — this becomes a tight spin loop: spawn → panic → recover (clears every account's delivery channel, sends `RecoveryRequired`) → spawn again, with no rate limit. Worth adding the same short sleep here for defense-in-depth, since the whole point of this PR is guarding against exactly this kind of runaway failure mode.

crates/marmot-app/src/relay_plane/mod.rs
```diff
@@ -664,134 +749,338 @@ impl RelayPlaneHealth {
...
+    match notification {
+        RelayPoolNotification::Event {
+            relay_url,
+            subscription_id,
+            event,
+        } => {
+            if let Ok(event) = NostrTransportEvent::from_nostr_event(&event) {
+                tracing::trace!(
+                    target: "marmot_app::relay_plane",
+                    method = "spawn_relay_notification_forwarder",
```
erskingardner: Nit: the tracing calls inside the newly-extracted `handle_relay_notification` still carry `method = "spawn_relay_notification_forwarder"` (5 occurrences below). Per this repo's observability convention (explicit module/method fields for grep-ability), these should probably read `handle_relay_notification` now that the logic lives in its own function. Not blocking, just a stale label left over from the extraction.

erskingardner (commented):
## Summary

Solid, well-diagnosed fix. The root cause (fire-and-forget `handle_notifications` loop exiting silently on `Lagged`/`Closed`, leaving a finished `JoinHandle` parked and inbound channels open forever) is confirmed against the vendored `nostr-relay-pool` source, and the fix directly addresses every acceptance criterion in #1156: the consumer is now supervised, lag/closure/panic are observed and counted, per-account delivery producers are closed to drive the existing account-worker reconnect path, directory subscriptions are force-rebuilt after recovery, and liveness counters are surfaced through relay health / UniFFI / CLI / TUI.

Note: I would normally submit this as an approval, but GitHub won't let the PR author approve their own PR through this token — treat this as approve-with-notes.

Verified locally against the PR branch:
- `cargo check -p marmot-app --lib --tests` — clean
- `cargo clippy -p marmot-app -p wn-cli -p marmot-uniffi --lib --tests` — clean
- `cargo test -p marmot-app --lib relay_plane` — 39/39 pass, including the new deterministic lag/panic/managed-worker-reconnect/exactly-once tests
- `cargo fmt --all -- --check` — clean

The new tests are a real strength here: `supervised_notification_lag_recovers_later_inbound_exactly_once` and `managed_account_worker_reopens_transport_after_notification_recovery` exercise the exact failure mode from the issue end-to-end (lag → recovery → reconnect → exactly-once redelivery), which is exactly what the acceptance criteria asked for.

## Findings (non-blocking)

- `crates/marmot-app/src/relay_plane/mod.rs` — the consumer-panic branch in `spawn_relay_notification_supervisor` respawns immediately with no backoff, unlike the graceful `Closed` branch a few lines above which sleeps 100ms. A deterministic panic trigger (e.g. a malformed event a relay keeps re-sending) would spin this loop with no rate limit. Left an inline comment with specifics.
- Same file — the tracing `method` fields inside the newly-extracted `handle_relay_notification` still say `"spawn_relay_notification_forwarder"`, left over from the extraction. Minor grep-ability nit, not blocking.

Neither finding affects the correctness of the recovery path itself. Recommend merge; the panic-backoff gap would be a nice small follow-up.

erskingardner (commented):
## Re-review of `9cad660` — both review notes covered

Re-reviewed after `Harden relay notification recovery`. Both of my earlier notes are addressed, and the two CodeRabbit findings are too.

### 1. Panic-respawn backoff — fixed

`RelayNotificationRestartBackoff` replaces the bare `sleep(100ms)`: 100ms doubling to a 30s cap, applied to **both** the `Closed` arm and the previously-unthrottled `Err(join_error)` panic arm. The reset-on-healthy-runtime rule (`consumer_runtime >= 5s` → back to 100ms) is the right shape — it keeps a genuinely healthy consumer that dies once from inheriting a stale multi-second delay, while a hot panic loop escalates. `notification_restart_backoff_caps_and_resets_after_healthy_runtime` covers the cap and the reset.

### 2. Stale `method` labels — fixed

All 5 occurrences in `handle_relay_notification` now read `method = "handle_relay_notification"`.

### 3. Respawn churn on clean shutdown — fixed

The `pool().is_shutdown()` guard in `spawn_router` is the right fix for the "finished supervisor always treated as an unexpected `Closed` exit" case: a terminally shut-down pool now drops the handle without emitting a spurious `RecoveryRequired` or clearing account delivery senders. `clean_sdk_shutdown_stays_terminal_across_spawn_router_reentry` asserts exactly that, including that an existing account receiver stays open.

### 4. Inline recovery rebuild self-amplification — fixed

`DirectoryRecoveryRebuildQueue` moves the rebuild off the worker's `select!` arm and coalesces a burst of `RecoveryRequired` / `Lagged` events into at most one follow-up pass, so the worker keeps draining events while a rebuild runs. `DirectoryRecoveryRebuildTask`'s abort-on-drop covers both worker exit paths.

Worth noting the pairing that makes this safe: spawning the rebuild removed the implicit serialization the inline `await` provided, and the new `directory_subscription_sync` mutex in `sync_directory_user_subscriptions` puts it back — a command-path sync and a recovery-path sync can no longer interleave. Good that these landed together.

Two details I checked and found clean:
- The `expect()` in the disabled-branch async block can't fire — `select!` evaluates the async expression but never polls it when the precondition is false, and the `expect` is inside the block body.
- Aborting an in-flight rebuild can't tear subscription state: `replace_subscriptions` mutates under its lock with no intervening await, and the abort path only runs at worker teardown.

### Verification

Against `9cad660` in a disposable worktree: `cargo fmt --check`, `cargo clippy -p marmot-app --all-targets`, and the full `cargo test -p marmot-app` unit suite (478 passed) all clean; `relay_plane` tests 41/41 including the two new ones. Five `relay_runtime` integration tests time out locally, but they fail identically on the merge-base `b2919f15` and CI's "Relay runtime tests" job is green — pre-existing and environmental, not from this PR.

### Remaining nit (non-blocking, fine to ignore)

The `Lagged` exit still restarts with no backoff and no sleep. It doesn't hot-spin — the receiver is reused, so tokio reports lag once and resumes — but under sustained lag every cycle clears all account delivery senders and broadcasts `RecoveryRequired`. The new coalescing queue bounds the downstream rebuild cost, so the blast radius is small; only worth revisiting if lag telemetry shows repeated incidents in practice.

No blockers. Ship it.

PR #1158, opened 2026-07-28
Retain pending sends during transport activation

### LINKED ISSUE #1151: Local-ready startup rejects sends with account not active instead of retaining pending delivery
## Summary

After the runtime reports local readiness, a user can open a persisted conversation and submit a message while relay/account activation is still pending or has failed. The message is shown optimistically by Android, but MDK rejects the send with `account not active`; Android then turns the bubble into **Send failed**.

`account not active` is an internal transport-lifecycle state, not a terminal message outcome. A locally ready signing account should be able to accept the send locally, keep one pending projection, and publish that same logical message in the background when transport activation succeeds.

## Reproduction

1. Start the app/runtime with a persisted signing account and chat history.
2. Let `Marmot.start()` return at local readiness while initial transport activation is unavailable, delayed, or fails.
3. Open a locally projected conversation and send a text message.
4. Observe an optimistic bubble briefly appear, then become **Send failed** with `details=account not active`.

## Expected

- The send is accepted locally and remains as one pending optimistic/projected bubble.
- Transport activation and publication continue in the background.
- When activation succeeds, MDK publishes and reconciles the same message; it does not mint a second event or bubble.
- `TransportAdapterError::AccountNotActive` caused by startup/reconnect readiness is treated as retryable lifecycle state and is not surfaced as a terminal send error.
- A genuinely signed-out account, removed group member, or other authoritative terminal state still rejects/fails the send.

## Source-confirmed mechanism

Verified against `origin/master` at `5fe7dccc94b411e9dca3ad45927afc8ffc0f2c0f`.

PR #1134 intentionally changed startup so local hydration/read readiness completes before transport activation, subscription registration, and catch-up:

- `run_app_runtime_account_worker` signals `ready(Ok(()))` immediately after local snapshot capture (`crates/marmot-app/src/runtime/account_worker.rs:432-467`).
- Initial `sync_with_startup_stage_telemetry()` then performs transport activation asynchronously (`account_worker.rs:469-488`; `crates/marmot-app/src/client/sync.rs:73-105, 119-138`).
- Mutations arriving during that work are deferred (`account_worker.rs:489-515`).
- If startup sync fails, readiness remains successful (`account_worker.rs:542-553`), but every deferred mutation is still replayed immediately against the client whose transport never activated (`account_worker.rs:555-574`).
- The Nostr adapter rejects `publish` when its active-account map has no entry (`crates/transport-nostr-adapter/src/lib.rs:859-877`), producing the exact `account not active` detail.

The send path does not preserve background work after that error:

- `send_app_event_with_local_projection` records a local row before network work (`crates/marmot-app/src/client/mod.rs:1544-1549`).
- It then calls `sync_runtime_groups()` and publish (`client/mod.rs:1551-1568`).
- Any error invalidates the just-created projection and returns failure (`client/mod.rs:1572-1596`), so no accepted pending delivery remains for background convergence.

Downstream Android already creates and publishes its own optimistic bubble before calling UniFFI (`ConversationController.send`, `whitenoise-android` `Controllers.kt:4841-4911`). The FFI error is caught and that bubble is converted to `Failed`, with the raw detail shown (`Controllers.kt:4974-5015`). This is therefore not solved by adding another Android persistence/outbox layer; MDK owns the durable protocol projection and retry semantics.

## Fix direction

- Define a retryable send-acceptance path for locally ready, signing accounts whose transport activation is not yet settled.
- Persist/retain one pending own-message projection plus enough engine-owned outbound intent to publish it later.
- Re-drive it when account transport activation/reconnect succeeds, using stable identity/idempotency so convergence updates the same row.
- Do not replay a deferred send into a known-inactive adapter and then retract its projection as if the user's intent were terminally invalid.
- Keep Android responsible only for optimistic presentation and rendering the MDK pending/delivered/terminal-failure state.

A narrow warm-process repair may coordinate deferred mutation replay with a successful transport activation, but it must not create a second Kotlin outbox or duplicate the broader durable/restart queue tracked downstream.

## Acceptance criteria

- [ ] With local readiness complete and initial account activation unavailable, `send_text` produces one locally visible pending own-message row rather than `account not active`.
- [ ] Restoring activation publishes that same logical message and updates the row to delivered with no duplicate event or bubble.
- [ ] Multiple rapid sends preserve order and exactly-once identity while activation is pending.
- [ ] Repeated activation failures leave the messages pending/retryable rather than repeatedly retracting/recreating them.
- [ ] Signed-out accounts and authoritative group removal remain terminal and cannot publish queued plaintext.
- [ ] The pending state is observable through the existing projection/UniFFI surface; no Android protocol-data cache is required.

## Regression coverage

- `marmot-app` runtime test: return from startup at local readiness, force transport activation to fail, submit a send, and assert one pending projection plus no terminal `AccountNotActive` result.
- Recovery test: make activation succeed and assert one relay publish and one projection transition to delivered.
- Ordering/idempotency test: queue two sends during activation failure and assert ordered, non-duplicated convergence.
- Terminal-state test: sign out or apply authoritative removal before recovery and assert the pending work is not published.

## Related

- PR #1134 introduced the local-ready/network-pending startup contract that exposes this concrete failure.
- `whitenoise-android` #1676 tracks a separate Android preflight guard that can reject before any optimistic/MDK handoff.
- `whitenoise-android` #1216 tracks the broader offline queue across process death/reboot; this issue is the concrete MDK lifecycle failure and engine-owned acceptance/retry contract it depends on.
- MDK #516 exposed `retry_group_convergence`, but the current send path retracts the local projection on this error rather than leaving accepted pending work to retry.

erskingardner:
## Summary

- accept application-message sends locally when a hydrated signing account is temporarily transport-inactive
- persist those sends in the existing durable outbound-intent queue and re-drive them after activation, reconnect, or restart
- retain and finalize the same optimistic projection, preserve queue order, and delay notification triggers until actual publication
- keep authoritative removal and other terminal engine gates intact

## Root cause

PR #1134 intentionally moved runtime readiness ahead of transport activation and catch-up. When initial activation failed, deferred sends were replayed against an inactive adapter. `send_app_event_with_local_projection` then treated `TransportAdapterError::AccountNotActive` as terminal and retracted the optimistic row.

The convergence scheduler also retried group synchronization without reactivating the account, and failed startup/reconnect paths could leave hydrated queued work unscheduled.

## User impact

A message sent immediately after local readiness now remains as one pending row instead of becoming “Send failed.” When transport activation recovers, MDK publishes that same logical event once and finalizes the existing row without creating a duplicate.

Closes #1151.

## Validation

- `just fast-ci`
- `cargo test -p marmot-app local_ready -- --nocapture`
- `cargo test -p marmot-app locally_queued_send_survives_runtime_restart_and_failed_reactivation -- --nocapture`
- `cargo test -p cgka-engine applying_removal_commit_purges_queued_outbound_intents -- --exact`

### DIFF
--- a/crates/cgka-engine/src/message_processor/mod.rs
+++ b/crates/cgka-engine/src/message_processor/mod.rs
@@ -188,7 +188,48 @@ impl<S: StorageProvider> Engine<S> {
     }
 
     pub(crate) async fn do_send(&mut self, intent: SendIntent) -> Result<SendResult, EngineError> {
-        let group_id = send_intent_group_id(&intent).clone();
+        let group_id = self.validate_send_acceptance(&intent)?;
+        // Disband is an intent-persistence operation, not a Commit-preparation
+        // operation. It remains acceptable while another staged Commit owns
+        // the group and while Unrecoverable is paused.
+        if matches!(&intent, SendIntent::Disband { .. }) {
+            return self.do_request_disband(group_id);
+        }
+        if self.should_queue_outbound_intent(&group_id).await? {
+            return self.queue_outbound_intent(group_id, intent);
+        }
+
+        self.do_send_ready(intent).await
+    }
+
+    pub(crate) fn do_queue_app_message(
+        &mut self,
+        group_id: GroupId,
+        payload: Vec<u8>,
+    ) -> Result<SendResult, EngineError> {
+        let intent = SendIntent::AppMessage {
+            group_id: group_id.clone(),
+            payload,
+        };
+        self.validate_send_acceptance(&intent)?;
+        if let Some(state) = self.epoch_manager.state(&group_id)
+            && !matches!(state, EpochState::Stable { .. })
+        {
+            return Err(EngineError::InvalidTransition(
+                cgka_traits::engine_state::InvalidTransition {
+                    from: state.name(),
+                    to: "queue_app_message",
+                    reason: "queue_app_message requires Stable",
+                },
+            ));
+        }
+        let queued = self.queue_outbound_intent(group_id.clone(), intent)?;
+        self.schedule_pending_convergence_group(&group_id);
+        Ok(queued)
+    }
+
+    fn validate_send_acceptance(&mut self, intent: &SendIntent) -> Result<GroupId, EngineError> {
+        let group_id = send_intent_group_id(intent).clone();
         // Quarantine gate: a group frozen by hydration quarantine must not
         // stage, queue, or publish anything — a confirm would set_stable and
         // silently un-quarantine it out of band (mdk#364).
@@ -197,16 +238,15 @@ impl<S: StorageProvider> Engine<S> {
             return Err(EngineError::InvalidTransition(
                 cgka_traits::engine_state::InvalidTransition {
                     from: "Disbanded",
-                    to: crate::audit_helpers::send_intent_kind_str(&intent),
+                    to: crate::audit_helpers::send_intent_kind_str(intent),
                     reason: "Disbanded is terminal",
                 },
             ));
         }
-        // Disband is an intent-persistence operation, not a Commit-preparation
-        // operation. It remains acceptable while another staged Commit owns
-        // the group and while Unrecoverable is paused.
-        if matches!(&intent, SendIntent::Disband { .. }) {
-            return self.do_request_disband(group_id);
+        // Disband has its own durable acceptance gates and intentionally
+        // remains available while the ordinary send lifecycle is paused.
+        if matches!(intent, SendIntent::Disband { .. }) {
+            return Ok(group_id);
         }
         // Strict cutover freezes membership growth in legacy groups. Existing
         // members may continue messaging and applying other group changes, but
@@ -232,7 +272,7 @@ impl<S: StorageProvider> Engine<S> {
             return Err(EngineError::InvalidTransition(
                 cgka_traits::engine_state::InvalidTransition {
                     from: "Removed",
-                    to: crate::audit_helpers::send_intent_kind_str(&intent),
+                    to: crate::audit_helpers::send_intent_kind_str(intent),
                     reason: "local group copy is marked removed (self-evicted)",
                 },
             ));
@@ -243,7 +283,7 @@ impl<S: StorageProvider> Engine<S> {
             return Err(EngineError::InvalidTransition(
                 cgka_traits::engine_state::InvalidTransition {
                     from: "Unrecoverable",
-                    to: crate::audit_helpers::send_intent_kind_str(&intent),
+                    to: crate::audit_helpers::send_intent_kind_str(intent),
                     reason: "group is Unrecoverable pending verified repair",
                 },
             ));
@@ -252,7 +292,7 @@ impl<S: StorageProvider> Engine<S> {
             return Err(EngineError::InvalidTransition(
                 cgka_traits::engine_state::InvalidTransition {
                     from: "Disbanding",
-                    to: crate::audit_helpers::send_intent_kind_str(&intent),
+                    to: crate::audit_helpers::send_intent_kind_str(intent),
                     reason: "disband requested or awaiting convergence",
                 },
             ));
@@ -261,16 +301,12 @@ impl<S: StorageProvider> Engine<S> {
             return Err(EngineError::InvalidTransition(
                 cgka_traits::engine_state::InvalidTransition {
                     from: "Leaving",
-                    to: crate::audit_helpers::send_intent_kind_str(&intent),
+                    to: crate::audit_helpers::send_intent_kind_str(intent),
                     reason: "leave requested",
                 },
             ));
         }
-        if self.should_queue_outbound_intent(&group_id).await? {
-            return self.queue_outbound_intent(group_id, intent);
-        }
-
-        self.do_send_ready(intent).await
+        Ok(group_id)
     }
 
     pub async fn converge_and_drain_queued_outbound_intents(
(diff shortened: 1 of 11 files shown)


crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -188,7 +188,48 @@ impl<S: StorageProvider> Engine<S> {
...
+        };
+        self.validate_send_acceptance(&intent)?;
+        if let Some(state) = self.epoch_manager.state(&group_id)
+            && !matches!(state, EpochState::Stable { .. })
+        {
+            return Err(EngineError::InvalidTransition(
+                cgka_traits::engine_state::InvalidTransition {
+                    from: state.name(),
+                    to: "queue_app_message",
+                    reason: "queue_app_message requires Stable",
```
erskingardner, line 222: Minor edge-case note: `do_queue_app_message` requires `EpochState::Stable` and errors otherwise (`InvalidTransition`, "queue_app_message requires Stable"). If transport activation is failing *and* the group happens to have an in-flight epoch transition (`PendingPublish`/`Merging`) at the same moment, `send_text` still surfaces a terminal error rather than the one-pending-row behavior this PR establishes for the common case.

This is not a regression — on master today the same compound situation (`should_queue_outbound_intent` returns `Ok(false)` for a non-stable epoch, so `do_send_ready` runs immediately and hits the same `AccountNotActive` from the transport) already fails the same way. But it means the fix is narrower than "any locally-ready send with inactive transport stays pending" — it is "any locally-ready send with inactive transport and a stable epoch stays pending." Worth a one-line note in the doc comment on `CgkaEngine::queue_app_message` (or an explicit regression test) calling out that this compound case is still a known gap, so it does not get rediscovered as a surprise later.

Not blocking.

crates/marmot-app/src/client/sync.rs
```diff
@@ -511,6 +511,19 @@ impl AppClient {
         self.remember_pending_convergence_effects(&effects);
         self.remember_published_reports(&effects);
         let finalize_updates = self.finalize_published_app_message_source_retention(&effects)?;
+        let publish_new_message_notification =
+            effects.published_app_messages.iter().any(|published| {
+                let group_id_hex = hex::encode(published.group_id.as_slice());
+                self.app
+                    .reaction_target(&self.state.label, &group_id_hex, &published.app_event_id)
```
erskingardner, line 518: Nit: this reuses `AppClient::reaction_target`, whose doc comment describes it as resolving "the reacted-to target for a reaction notification." Here it is repurposed to look up whether a just-published own message is a live (non-deleted, non-invalidated) `MARMOT_APP_EVENT_KIND_CHAT` row, purely to gate the deferred new-message push notification after a queued send drains through convergence. Functionally correct (same underlying by-id timeline lookup), but the name/doc no longer matches this use site. A small rename (e.g. `timeline_message_target`, dropping the reaction-specific framing) or a thin same-signature alias would keep the call site self-explanatory. Non-blocking.

erskingardner (commented):
## Summary

This closes #1151 by giving locally-ready sends a durable acceptance path when the transport is temporarily inactive, instead of retracting the optimistic projection on `TransportAdapterError::AccountNotActive`.

The implementation is a well-scoped, root-cause fix rather than a symptom patch:

- `Engine::queue_app_message` (new `CgkaEngine` trait method) durably queues an `AppMessage` `SendIntent` through the **existing** `queue_outbound_intent` / `pending_convergence_groups` / `converge_and_drain_queued_outbound_intents` machinery that already exists for group-evolution intents queued mid-convergence. No new persistence shape, no parallel queue — it reuses the same durable-queue and restart-hydration path already exercised by that machinery (`crates/cgka-engine/src/engine.rs:1287,1375`), which is exactly why the restart test (`locally_queued_send_survives_runtime_restart_and_failed_reactivation`) works with no additional wiring.
- `do_send`'s pre-existing authoritative gates (quarantine, disband, removal, unrecoverable, disbanding, leave) were cleanly factored out into `validate_send_acceptance` and are shared by both the normal send path and the new queue path, preserving order and semantics exactly (verified by re-reading the pre/post diff side by side). The added `invite_leave.rs` test confirms the `Removed` gate still rejects a forced local queue.
- `marmot-app` falls back to `queue_app_message_with_audit_context` only on `AppError::is_account_not_active()`, a narrow, correctly-scoped match on `TransportAdapterError::AccountNotActive` (both direct and `AccountError`-wrapped) — other transport errors (publish/subscription failures) still take the existing terminal path, matching the issue's explicit scoping.
- The notification-trigger gating (`published.is_some()`) is a nice side benefit: it also fixes the same "notify before actual publish" issue for the pre-existing implicit-queue path (unsettled convergence), not just the new inactive-transport path, and the added `advance_convergence_after_runtime_sync` block correctly fires the deferred notification at the one call site that drains and publishes queued sends.
- The convergence-scheduler reactivation gap described in the issue (retrying `sync_runtime_groups` forever without ever calling `prepare_transport`/`activate_transport` again) is closed with a small, targeted addition in `account_worker.rs`.

Regression coverage is strong: pending-on-activation-failure, ordering/exactly-once across two concurrent queued sends, and durable-queue survival across a restart with a second failed reactivation — matching the acceptance criteria in #1151 closely.

## Notes (non-blocking)

Left two inline comments:

1. `do_queue_app_message` requires `EpochState::Stable` and errors otherwise, so the narrow compound case of "transport inactive + epoch mid-transition" still surfaces a terminal error rather than a pending row. Not a regression (master fails the same way today for that combination), but worth a doc note or test so it's a known, not rediscovered, gap.
2. `AppClient::reaction_target` is repurposed in `sync.rs` to gate the deferred new-message notification; functionally fine, but its name/doc are reaction-specific and no longer describe this use site.

No blocking issues found.

PR #1159, opened 2026-07-28
Add durable agent reply and conversation history

erskingardner:
## Summary

- add exact-message and stable cursor-paginated materialized timeline reads to `marmot.agent-control.v2`
- project bounded, privacy-aware timeline DTOs through `wn-agent`, including sender identity, timestamps, reply links, attachments, current reactions, and durable message ids
- attach complete referenced-message context and a recent canonical chat window to activated OpenClaw and Hermes turns
- add model-callable `marmot_history` tools to both integrations for exact lookup and older transcript pages
- document the history/recovery contract and keep the OpenCode harness exhaustive over the additive response variants

## Root cause

`wn-agent` already hydrated reply targets, but OpenClaw split that information between native quote fields and an incomplete untrusted payload. The pinned OpenClaw SDK also suppresses the body of self-authored quotes by default, producing the observed state where the reply id and `availability: available` survived but text and sender context did not reach the model reliably.

Hermes preserved more fields in its adapter object, but its prompt-facing context did not retain the full durable message identity and author metadata. More broadly, neither integration had an authoritative, on-demand transcript API; the inbound event stream was being asked to serve as both notification transport and history.

## Behavior and compatibility

Timeline reads use the current materialized user-visible projection rather than replaying raw events, so edits and current reactions are reflected. Deleted and invalidated rows retain identity and attribution while plaintext and attachment metadata remain redacted. Text, attachments, reactions, page size, and total response size are bounded.

This is additive at the wire and storage layers and requires no database migration or capability negotiation. `wn-agent` and its connectors are expected to ship together. Automatic history lookup is best-effort and never suppresses the current inbound turn if it fails.

## Validation

- `just fast-ci`
- `cargo test -p agent-control -p agent-connector --no-fail-fast`
- `cargo test -p agent-control -p agent-connector -p storage-sqlite timeline --no-fail-fast`
- OpenClaw: `pnpm typecheck && pnpm test` — 199 passed, 1 skipped
- Hermes: `python3 -m unittest discover -s integrations/hermes/marmot/tests -p 'test_*.py'` — 163 passed
- `cargo test -p wn-opencode` — 49 passed, process-spawning E2E ignored

### DIFF
--- a/crates/agent-connector/src/timeline.rs
+++ b/crates/agent-connector/src/timeline.rs
@@ -1,0 +1,284 @@
+//! Read-only materialized-timeline projection for agent-control clients.
+
+use agent_control::{
+    AGENT_CONTROL_TIMELINE_DEFAULT_LIMIT, AGENT_CONTROL_TIMELINE_MAX_LIMIT,
+    AGENT_CONTROL_TIMELINE_REACTIONS_MAX, AGENT_CONTROL_TIMELINE_TEXT_MAX_CHARS,
+    AgentControlEnvelope, AgentControlResponse, AgentControlTimelineCursor,
+    AgentControlTimelineMessage, AgentControlTimelineMessageAvailability,
+    AgentControlTimelineReaction, MAX_AGENT_CONTROL_FRAME_BYTES,
+};
+use marmot_app::{TimelineMessageQuery, TimelineMessageRecord, TimelinePagination};
+
+use crate::AgentConnector;
+use crate::error::ConnectorError;
+use crate::event_projection::{attachment_summaries_from_tags, control_actor};
+use crate::validation::normalize_hex;
+
+const TIMELINE_ATTACHMENT_FIELD_MAX_CHARS: usize = 512;
+const TIMELINE_REACTION_EMOJI_MAX_CHARS: usize = 64;
+
+impl AgentConnector {
+    pub(crate) fn timeline_message_response(
+        &self,
+        account_id_hex: &str,
+        group_id_hex: &str,
+        message_id_hex: &str,
+    ) -> Result<AgentControlResponse, ConnectorError> {
+        let account = self.local_account_for_account_id(account_id_hex)?;
+        let group_id_hex = normalize_hex(group_id_hex)?;
+        let message_id_hex = normalize_hex(message_id_hex)?;
+        let message = self
+            .runtime
+            .timeline_message(&account.label, &group_id_hex, &message_id_hex)?
+            .map(|record| timeline_message(record, &account.account_id_hex));
+        Ok(AgentControlResponse::TimelineMessage {
+            account_id_hex: account.account_id_hex,
+            group_id_hex,
+            message_id_hex,
+            message,
+        })
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn timeline_list_response(
+        &self,
+        account_id_hex: &str,
+        group_id_hex: &str,
+        before: Option<AgentControlTimelineCursor>,
+        after: Option<AgentControlTimelineCursor>,
+        before_inclusive: bool,
+        limit: Option<u32>,
+    ) -> Result<AgentControlResponse, ConnectorError> {
+        let account = self.local_account_for_account_id(account_id_hex)?;
+        let group_id_hex = normalize_hex(group_id_hex)?;
+        let before = before.map(normalize_cursor).transpose()?;
+        let after = after.map(normalize_cursor).transpose()?;
+        let limit = limit
+            .unwrap_or(AGENT_CONTROL_TIMELINE_DEFAULT_LIMIT)
+            .clamp(1, AGENT_CONTROL_TIMELINE_MAX_LIMIT);
+        let page = self.runtime.timeline_messages_with_query(
+            &account.label,
+            TimelineMessageQuery {
+                group_id_hex: Some(group_id_hex.clone()),
+                search: None,
+                pagination: TimelinePagination {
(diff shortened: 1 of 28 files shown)


crates/agent-connector/src/timeline.rs
```diff
@@ -0,0 +1,284 @@
...
+                return Ok(response);
+            }
+            if after.is_some() {
+                messages.pop();
+                has_more_after = true;
+            } else {
+                messages.remove(0);
+                has_more_before = true;
+            }
+        }
```
erskingardner, line 129: This oversized-page trimming loop (lines 82-106) implements a real invariant -- every `TimelinePage` response must stay under `MAX_AGENT_CONTROL_FRAME_BYTES` -- by popping from the end when `after` is set and from the front otherwise, so the messages nearest the requested cursor survive. I traced it against the ASC/DESC + reverse ordering in `storage-sqlite/src/timeline.rs::message_timeline` and it looks correct in all three cursor cases (`before`, `after`, `None`), but there's no test for it here or in `crates/agent-connector/src/tests.rs` (which has zero hits for "Timeline"). Worth a dedicated test that forces the size cap (e.g. many near-max-length messages) and asserts the surviving messages + `has_more_before`/`has_more_after` are correct for both a `before`-cursor and an `after`-cursor page.

crates/agent-connector/src/timeline.rs
```diff
@@ -0,0 +1,284 @@
...
+                account_id_hex: account.account_id_hex.clone(),
+                group_id_hex: group_id_hex.clone(),
+                messages: messages.clone(),
+                has_more_before,
+                has_more_after,
+            };
+            // Keep a complete response envelope below the protocol frame cap. When a
+            // requested page is too large, preserve the messages nearest the cursor
+            // and expose the omitted side through the existing pagination flags.
+            let sizing_envelope = AgentControlEnvelope::new(Some("0".repeat(64)), response.clone());
```
erskingardner: Minor: the size probe uses a synthetic 64-char id (`Some("0".repeat(64))`) to estimate the final envelope size, but the real envelope uses the caller-supplied `request.id` (`connection.rs` `AgentControlEnvelope::new(request.id, response)`), which has no protocol-level length bound. A client sending an unusually long id could make the actual written frame exceed `MAX_AGENT_CONTROL_FRAME_BYTES` even though this check passed, causing `write_frame` to fail the whole response. Low real-world risk since ids are normally short, but worth a comment noting the assumption, or sizing off the actual `request.id` length if it's cheaply available here.

erskingardner (commented):
## Summary

Reviewed the durable timeline read API (`timeline_message_get` / `timeline_list`), its plumbing through `marmot-app`/`storage-sqlite`, and the Hermes/OpenClaw/OpenCode integration wiring.

The design is solid: I traced the cursor/pagination ordering (`before`/`after`/`None` → ASC/DESC + reverse in `storage-sqlite/src/timeline.rs`) against the new frame-size trimming loop in `agent-connector/src/timeline.rs` and the "keep the messages nearest the cursor" invariant holds in all three cases. Deleted/invalidated masking, bounded text/attachment/reaction fields, and the OpenClaw `suppressSelfQuoteBody` root-cause fix (backed by a real pinned-SDK test in `reply-context-sdk.test.ts`, not just a mock) all check out. No correctness bugs found.

Two non-blocking items worth addressing:

- No test exercises `timeline_list_response`'s oversized-page trimming loop (the pop-front/pop-back-by-cursor-direction logic that keeps every response under the 1 MiB frame cap). `crates/agent-connector/src/tests.rs` has zero references to "Timeline", and the new file's own tests only cover the DTO-mapping helper.
- This looks like a user-facing feature comparable to prior `## [Unreleased]` entries in `crates/cli/CHANGELOG.md` (e.g. the 0.9.9 "WN Agent inbound delivery now carries structured actor..." entry), but no changelog entry was added.

Mostly minor/non-blocking; commenting rather than approving outright given the coverage gap on the new size-cap invariant.

erskingardner (commented):
## Re-review (since commit adcaa41e)

Both non-blocking items from the previous review are addressed in 311d4f78 ("Address timeline history review feedback"):

- **Trimming-loop coverage**: `bounded_timeline_page_response` is now a standalone function with three new tests (`oversized_before_page_keeps_newest_rows_nearest_cursor`, `oversized_after_page_keeps_oldest_rows_nearest_cursor`, `page_budget_accounts_for_actual_request_id`) that construct a maximally-sized `TimelinePage` and verify the pop-front/pop-back-by-cursor-direction behavior, the resulting `has_more_before`/`has_more_after` flags, and that the final `encode_frame` output stays under `MAX_AGENT_CONTROL_FRAME_BYTES`. Ran these locally — all pass.
- **CHANGELOG**: an `## [Unreleased] / ### Added` entry was added to `crates/cli/CHANGELOG.md`, consistent with the existing 0.9.9 entry format.

The refactor also fixes a real latent bug, not just adding tests: the old sizing envelope used a fixed `Some("0".repeat(64))` placeholder for the request id when estimating frame size, which could mis-size the budget against the request id actually written to the wire. The new `bounded_timeline_page_response` takes the real `request_id: Option<&str>` (threaded from `handle_request` through `timeline_list_response`) and sizes against it, with a dedicated test (`page_budget_accounts_for_actual_request_id`) proving a long request id trims more messages than a short one while both stay under the frame cap.

Verified independently in a scratch clone: `cargo test -p agent-connector timeline` (5/5 pass), `cargo clippy -p agent-connector --all-targets` (clean), `cargo fmt -p agent-connector -- --check` (clean).

No new issues found in the delta. This is ready to merge (approval blocked by GitHub since I authored the PR under this account — no blocking findings remain).

PR #1160, opened 2026-07-28
Fix orphaned Keychain account import

erskingardner:
## Summary

- recover a matching account-ID-keyed signing credential when private-key login finds no filesystem account record
- keep live-record and duplicate-account protections fail-closed, including verification that the stored credential matches the supplied nsec
- carry import provenance through runtime setup so rollback preserves a pre-existing Keychain credential while still deleting a newly created credential

## Root cause

iOS can remove the App Group account home during uninstall while retaining the signing credential in Keychain. The UniFFI-backed runtime login path continued to call the strict `import_nostr_account` path, so the surviving credential triggered `AccountIdInUse` before the account record could be recreated. The lower-level recovery behavior existed for repeatable installer bootstrap, but it was not wired into the production Marmot runtime path.

## Safety

The recovery path remains strict:

- any live account record is still a duplicate
- another local-signing record for the same account ID is still rejected
- an orphaned credential is loaded and matched to the imported nsec before recreating the record
- a mismatched credential remains untouched and returns `AccountIdMismatch`
- rollback revalidates the imported record and credential before preserving a secret that predates setup

Generated-key creation, signed-out reactivation, public-only identities, external signers, and normal duplicate imports retain their existing paths.

## Validation

- `cargo test -p marmot-app --test relay_runtime app_runtime_private_key -- --nocapture` — 2 passed
- `cargo test -p marmot-account` — 81 passed across the crate suites
- `cargo test -p marmot-app --lib register_external_signer_requires_matching_external_account` — 1 passed

The broader `cargo test -p marmot-app --test relay_runtime` run completed 76 tests and hit 5 unrelated failures in existing media/timeline tests. Each failure was reproduced individually on clean `master` at `b2919f15`, confirming they are baseline failures rather than regressions from this change.

## Summary by CodeRabbit

* **New Features**
  * Added idempotent Nostr private-key account imports, including recovery of previously orphaned credentials.
  * Added rollback handling for failed account setup while preserving valid recovered credentials.
* **Bug Fixes**
  * Prevented mismatched stored credentials from being captured or overwritten.
  * Improved cleanup so failed setup removes newly created credentials without deleting recovered ones.
* **Tests**
  * Added coverage for credential mismatch protection, orphaned-account recovery, and rollback behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3192,6 +3192,7 @@ dependencies = [
  "hex",
  "hkdf 0.12.4",
  "image",
+ "keyring-core",
  "marmot-account",
  "marmot-forensics",
  "marmot-markdown",
--- a/crates/marmot-account/src/home.rs
+++ b/crates/marmot-account/src/home.rs
@@ -84,6 +84,23 @@ pub struct AccountSummary {
     pub signed_out: bool,
 }
 
+/// Provenance for a strict Nostr private-key import used by account setup.
+///
+/// A live account record is never treated as idempotent by this flow. The
+/// only reusable state is an exact account-id-keyed signing credential whose
+/// filesystem account record is absent.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct NostrAccountImport {
+    account: AccountSummary,
+    reused_account_id_credential: bool,
+}
+
+impl NostrAccountImport {
+    pub fn account(&self) -> &AccountSummary {
+        &self.account
+    }
+}
+
 impl AccountSummary {
     pub fn can_sign(&self) -> bool {
         self.local_signing || self.external_signing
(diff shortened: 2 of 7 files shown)


erskingardner (commented):
**Recommend merge** (posting as COMMENT since GitHub blocks self-approval on your own PR — this is otherwise an approve).

Reviewed the diff, both linked crates (`marmot-account`, `marmot-app`), and traced through the fix path against the described iOS uninstall/reinstall scenario. This is a solid, well-scoped fix.

**What I verified:**
- `import_nostr_account_idempotent` correctly narrows recovery to the *credential* only: it still rejects a live account record at the derived label (`AccountExists`), still rejects a duplicate signing record under a different label with the same account id (`AccountIdInUse`), and only reuses an orphaned Keychain credential after `load_secret` confirms it matches the supplied `nsec` (`AccountIdMismatch` otherwise, credential untouched). This closes the actual bug: the old path (`import_nostr_account` → `write_signing_account_for_label`) rejected on `has_secret_for_account_id` alone, with no account-id match check, which is exactly what strands a surviving Keychain entry after the App Group home is wiped on iOS reinstall.
- `remove_account_inner`'s new `preserve_account_id_credential` branch revalidates the credential (existence + pubkey match) *before* the tombstone-rename commit point, consistent with the repo's validate-before-mutate convention, and correctly skips `remove_secret` only when the credential was reused rather than newly created.
- Traced `create_or_import_account`'s three rollback call sites: `rollback_on_setup_failure` is still gated on `!reactivating_existing`, so the signed-out-reactivation path is untouched; the new `rollback_import_after_setup_failure` falls back to the old full-removal rollback whenever `private_key_import` is `None` (new random key, public-only add, or reactivation), and only takes the preserve-credential path when a `NostrAccountImport` is present.
- Integration tests in `relay_runtime.rs` reproduce the real scenario end-to-end (wipe the account directory, keep the same keychain service, recover through the runtime), and separately verify rollback both retains a recovered credential and still deletes a freshly created one. Good coverage of the two branches that matter.
- CI (fmt, clippy, rust tests) is green on the PR as of this review.

**Non-blocking notes:**
- `import_nostr_account_idempotent` (crates/marmot-account/src/home.rs) duplicates a fair amount of structure from the existing `import_account_idempotent` (duplicate-record checks, "recover-or-create" credential handling). The duplication looks justified here — this variant must treat any *live* record as a hard duplicate rather than reusing it, which the existing idempotent-bootstrap method deliberately does not do — but a shared private helper for the "no visible record → match-or-create credential" tail might shave off some repetition if you revisit this later.
- The `_idempotent` suffix on the new method reads similarly to `import_account_idempotent`'s reuse semantics, even though the doc comment is explicit that this variant is strictly narrower (never reuses a live record). A more distinguishing name might save future readers a trip to the doc comment.

Neither point blocks merge.

PR #1165, opened 2026-07-28
Preserve bounded blank-line source gaps in Markdown AST

### LINKED ISSUE #1164: Preserve bounded blank-line source gaps in the Markdown FFI AST
## Summary

The shared Markdown AST drops the authored count of consecutive blank lines between blocks. This makes downstream clients unable to distinguish `first\n\nsecond` from `first\n\n\nsecond` without reparsing plaintext.

Discovered while working marmot-protocol/whitenoise-android#1719.

## Root cause

`marmot-markdown` uses blank lines as block delimiters but does not retain source-gap metadata. Both inputs therefore cross `marmot-uniffi` as the same adjacent paragraph list. Android receives only `MarkdownDocumentFfi.blocks` and cannot recover the blank-run count.

The fix belongs in the parser/FFI contract. Hand-editing generated Kotlin or parsing Markdown a second time in the UI would create a divergent parser and is not acceptable.

## Expected behavior

- The AST represents the bounded count of source blank lines before each block in every block-owning container needed by clients (document, block quote, and list item).
- `first\n\nsecond` and `first\n\n\nsecond` have distinct metadata.
- Very large blank runs are clamped defensively.
- Fenced code, list, quote, table, and details semantics do not gain fabricated gaps.
- UniFFI contract tests prove the metadata survives the boundary.
- Existing block variants and walkers should remain as stable as practical; prefer a small aligned metadata list over wrapping or changing every block variant.

## Downstream

After this lands, whitenoise-android#1719 will regenerate the vendored Kotlin bindings/native libraries and consume the metadata in Compose layout.

agent-p1p:
Fixes #1164.

## Summary

- preserve aligned `blank_lines_before` metadata for documents, block quotes, and list items
- clamp authored blank-line runs to the documented `MAX_SOURCE_BLANK_LINES = 8`
- retain marker-only blank lines across mixed nested block owners
- propagate source-gap metadata through serde and UniFFI without client-side reparsing
- keep fenced-code, table, list-boundary, block-quote, `<details>`, and code-span semantics unchanged
- keep the workspace and conformance cohort at `0.9.10`; this is not a release PR

## Changelog

- added the MarmotKit Markdown AST behavior under `crates/cli/CHANGELOG.md` → `Unreleased / Added`

## Verification

- `cargo test -p marmot-markdown --locked`
- `cargo test -p marmot-uniffi --locked`
- `just fast-ci` (initial implementation revision)

## Summary by CodeRabbit

* **New Features**
  * Markdown parsing now preserves authored blank-line spacing before document blocks, block quotes, and list items.
  * Preserved spacing is available through serialization and client integrations, enabling more faithful rendering without reparsing source text.
  * Blank-line counts are safely bounded and remain compatible with older serialized documents.

* **Tests**
  * Added coverage for spacing across nested containers, lists, block quotes, code blocks, and legacy data.


agent-p1p:
## Pip adversarial review — changes requested

Reviewed exact head `55986c4a`. Three blocking findings:

1. **Blocking — trailing blank lines from indented code are discarded instead of becoming the next block gap** (`crates/marmot-markdown/src/block.rs:161-164`, `589-600`). Blank lines seen while `Leaf::IndentedCode` is open only increment `pending_blanks` and force `mark_blank = false`; when that code block closes, `pending_blanks` is dropped. A direct probe shows `parse("    code\n\nparagraph").blank_lines_before == [0, 0]`, but the documented source contract requires `[0, 1]` (and two trailing blanks should yield `[0, 2]`). Preserve speculative indented-code blanks as code content only if another indented line follows; otherwise transfer them to source-gap tracking. Add root and nested-list regressions.

2. **Blocking — a marker-only line that opens a block quote is turned into paragraph content, so its gap is lost** (`crates/marmot-markdown/src/block.rs:313-329`, `392-394`). `parse(">\n> foo")` currently emits a quote paragraph beginning with `SoftBreak` and child gaps `[0]`. The `>` line is blank inside the newly opened quote and should produce child gaps `[1]` without fabricating inline content, consistent with the existing handling of marker-only lines in an already-open quote. Cover single and nested quote openers.

3. **Blocking — new UniFFI record fields shift existing positional fields** (`crates/marmot-uniffi/src/markdown.rs:22-29`, `85-91`). The generated Kotlin constructors are now `MarkdownDocumentFfi(blocks, blankLinesBefore, truncated)` and `MarkdownListItemFfi(blocks, blankLinesBefore, checked)`, while the prior positional fields were `(blocks, truncated)` and `(blocks, checked)`. Append `blank_lines_before` after the existing fields in both records so existing fields retain their positions; then regenerate the Kotlin binding to verify the constructor/lowering order. `BlockQuote` already appends its new variant field and is fine.

Verified as sound: ordinary document counts distinguish one vs. two blank lines, all owner slots share the same saturating clamp at `MAX_SOURCE_BLANK_LINES = 8`, the mixed quote/list owner fix keeps vectors aligned, serde defaults preserve legacy decode, and FFI conversion carries values across the boundary.

Checks run at this head:
- `cargo test -p marmot-markdown --locked` — pass
- `cargo test -p marmot-uniffi --locked` — pass
- `cargo fmt --all -- --check` — pass
- `cargo clippy -p marmot-markdown -p marmot-uniffi --all-targets --locked -- -D warnings` — pass
- `git diff --check 523a1938..HEAD` — pass
- GitHub CI — all current-head jobs green

No sensitive paths touched.

agent-p1p:
## Pip adversarial review — REQUEST_CHANGES

Reviewed final head `55986c4aa8b86068e6c2c0948429707482d12d77`.

BLOCKING:
- **Trailing blank runs after indented code are dropped instead of becoming source-gap metadata** — `crates/marmot-markdown/src/block.rs:158-178`, `crates/marmot-markdown/src/block.rs:589-600`. While `Leaf::IndentedCode` is open, blank lines only increment `pending_blanks` and set `mark_blank = false`, so `record_source_blank` never sees them. If the next non-indented line closes the code block, `close_leaf` discards `pending_blanks`. Reproduced at final head:
  - `parse("    code\n\nafter")` → root `blank_lines_before == [0, 0]`
  - `parse("    code\n\n\nafter")` → root `blank_lines_before == [0, 0]`
  - list-item owner: `"- before\n\n      code\n\n  after"` → child gaps `[0, 1, 0]`
  - block-quote owner: `">     code\n>\n> after"` → child gaps `[0, 0]`

  The blank run sits immediately before the following block, so the root cases should remain distinguishable as `[0, 1]` and `[0, 2]`. This violates #1164's core contract in all three required owners. Preserve trailing pending indented-code blanks as a bounded gap for the following block while retaining them as code content when indented code resumes.

SUGGESTION:
- **Add focused root, quote, list-item, and UniFFI regressions for a block following indented code** — `crates/marmot-markdown/tests/source_gaps.rs`, `crates/marmot-uniffi/src/markdown.rs`. Existing coverage checks internal fenced-code blanks and an indented block preceded by a gap, but not the trailing-indented-code transition that exposes this defect.

NITPICK:
- **Update the `Leaf::IndentedCode.pending_blanks` comment with the fix** — `crates/marmot-markdown/src/block.rs:68-74`. “Discarded on close” is now incompatible with preserving that trailing run for the next block's source-gap metadata.

SUMMARY: The aligned metadata design, per-owner tracking, clamping, serde defaults, and UniFFI propagation otherwise look sound. The indented-code transition is a blocking asymmetry: unlike fenced/math/table leaves, it absorbs and then drops the exact source run this PR promises to retain.

VERDICT: REQUEST_CHANGES
SENSITIVE_PATHS: none
FOLLOW_UP_ISSUES: none

Validation at this head:
- `cargo test -p marmot-markdown --locked` — pass
- `cargo test -p marmot-uniffi --locked` — pass
- `just fmt-check` — pass
- GitHub CI — all reported jobs pass or skip

agent-p1p:
Addressed all blocking review findings in b0802dab:

- Preserved trailing indented-code blank runs as the next block source gap, including list-item ownership.
- Treated marker-only opening quote lines as child gaps without fabricated SoftBreak content, including nested quotes.
- Appended the new UniFFI record fields after existing positional fields; regenerated Kotlin confirms (blocks, truncated, blankLinesBefore) and (blocks, checked, blankLinesBefore).

Validation: marmot-markdown and marmot-uniffi suites pass; targeted clippy and just fast-ci pass; all GitHub checks and the incremental CodeRabbit review are green.

agent-p1p:
@coderabbitai review

agent-p1p:
@coderabbitai review

agent-p1p:
## Pip adversarial re-review — PASS

Reviewed exact rewritten head `2cd5f66eb7fdd72558ba52c4a70049212382ca60` against base `523a193815eafb22560923a3960b70e07373ac93`.

**Verdict:** no blocking findings. The merge-gate list-tightness regression is fixed.

- `BlockParser::take_pending_source_gap` now consumes pending list-blank state before closing a dangling list (`crates/marmot-markdown/src/block.rs:764-778`), so the list retains its established loose semantics while the following root block receives the source gap.
- Focused regressions pass for a blank line before an ATX heading, fenced code, block quote, and paragraph (`crates/marmot-markdown/tests/source_gaps.rs:122-164`). Each produces root gaps `[0, 1]` and `Block::List { tight: false, .. }`.
- The earlier indented-code transition remains fixed: speculative blank runs are transferred to source-gap/list-blank tracking when the code leaf closes, while resumed indented code discards that speculative metadata (`crates/marmot-markdown/src/block.rs:160-181, 388-417, 619-646`).
- Marker-only opening quote lines remain child gaps without fabricated inline content (`crates/marmot-markdown/src/block.rs:330-353`; regressions at `tests/source_gaps.rs:54-98`).
- UniFFI positional compatibility remains intact: `MarkdownDocumentFfi` keeps `(blocks, truncated, blank_lines_before)` and `MarkdownListItemFfi` keeps `(blocks, checked, blank_lines_before)` (`crates/marmot-uniffi/src/markdown.rs:21-30, 84-92`).

**Commit verification:** GitHub reports one PR commit, authored and committed by `agent-p1p`; `verification.verified=true`, reason `valid`, verified at `2026-07-29T01:06:09Z`.

**Checks run/inspected at this exact head:**
- `cargo test -p marmot-markdown --test source_gaps --locked` — 21 passed.
- `cargo test -p marmot-markdown --locked` — pass.
- `cargo test -p marmot-uniffi --locked` — 115 tests passed across unit/integration suites.
- `cargo fmt --all -- --check` — pass.
- `cargo clippy -p marmot-markdown -p marmot-uniffi --all-targets --locked -- -D warnings` — pass.
- `git diff --check 523a1938..HEAD` — pass; worktree clean and exact local head confirmed.
- GitHub Actions code-owned runs for the exact head are complete and successful, including format, check, clippy, all four Rust test shards, simulator/vectors, relay E2E/runtime, Tamarin, binaries, and broker image. Release publishing is skipped as expected. CodeRabbit current-head rerun is still in progress at comment time.

**Sensitive-path classification:** none. The diff touches Markdown parsing/AST/FFI consumers plus workspace version/conformance-vector metadata; no MLS/CGKA, crypto/key handling, trust anchors, auth, group-state core, or identity/signing-key lifecycle paths.

BLOCKERS: 0
SUGGESTIONS: 0
NITPICKS: 0
FOLLOW_UP_ISSUES: none

crates/marmot-markdown/src/block.rs
```diff
@@ -557,33 +679,139 @@ impl BlockParser {
...
+        self.close_dangling_lists();
+        let slot = match self.containers.last() {
+            None => 0,
+            Some(Container::BlockQuote { .. } | Container::ListItem { .. }) => {
+                self.containers.len()
+            }
+            Some(Container::List { .. }) => unreachable!("dangling lists were just closed"),
+        };
+        let count = self.pending_source_gaps[slot];
+        self.pending_source_gaps.fill(0);
```
erskingardner, line 777: Nit (non-blocking): `take_pending_source_gap` resets the *entire* `pending_source_gaps` array on every call, not just the slot it reads. I traced this through several nested list/quote + blank-line scenarios (including quote-inside-list-item where the quote closes early and the list item's own slot must survive) and the invariant holds — every read targets the truly-innermost open container, and shallower slots are either moving in lockstep (physically-blank lines) or become dangling once their container closes. But that invariant is implicit and untested for deep interleavings (CodeRabbit's review flagged the same spot). A short comment stating the invariant explicitly, plus a test like `- item\n  > quoted\n\n  continued` (asserting the list item's third child gets `blank_lines_before == 1` while the quote's own gap is unaffected), would make this easier to safely modify later.

crates/marmot-markdown/tests/serde_roundtrip.rs
```diff
@@ -15,6 +15,40 @@ fn roundtrip_empty() {
     roundtrip("");
 }
 
+#[test]
+fn legacy_ast_without_source_gaps_deserializes_with_empty_metadata() {
+    let document: Document = serde_json::from_str(r#"{"blocks":[]}"#).expect("deserialize");
```
erskingardner, line 20: Nit (non-blocking): both new legacy-deserialization tests use `"blocks":[]` (or item `"blocks":[]`), so `blank_lines_before` defaulting to empty trivially matches `blocks.len()`. Neither test exercises a legacy document with a *populated* `blocks` array and no `blank_lines_before` key, which is the case that would actually violate the documented alignment invariant (`blank_lines_before.len() == blocks.len()`). Not a live bug today — nothing in this repo persists/deserializes a `Document`; the UniFFI boundary always reparses from plaintext — but if that ever changes, a consumer indexing `blocks[i]`/`blank_lines_before[i]` in lockstep on old data would be exposed. Worth either a test with non-empty legacy `blocks`, or a doc-comment caveat that pre-upgrade documents may have a shorter `blank_lines_before` than `blocks`.

erskingardner (approved):
Reviewed against #1164. This is a careful, well-scoped implementation of exactly what the issue asked for: an aligned `blank_lines_before` metadata list on `Document`/`BlockQuote`/`ListItem` (no wrapping of every block variant, as the issue requested), clamped to `MAX_SOURCE_BLANK_LINES`, threaded through serde and UniFFI, with fenced-code/table/details semantics left untouched.

I hand-traced the trickiest parts of the new gap-tracking state machine (`pending_source_gaps`, `take_pending_source_gap`'s full-array reset, the indented-code "speculative" gap buffers, and the nested blockquote-inside-list-item / list-inside-blockquote interactions) against several constructed inputs and they all produced the semantically correct result. All CI jobs (including the full `cargo test` matrix, clippy, and fmt) are green, and the PR shows good discipline: the second commit explicitly backs out an unsolicited workspace version bump, consistent with this repo's release cadence.

Two small, non-blocking notes below — neither blocks merge.

Recommend merge.

PR #1166, opened 2026-07-29
Feat/tui picker viewport

jgmontoya:
## Summary by CodeRabbit

- **New Features**
  - Added a group-scoped “add member” flow from Group Detail (`a`) that performs a targeted user search and confirms against the selected group.
  - Returning from the add-member flow now reloads and reopens Group Detail so the new membership is immediately reflected.

- **Bug Fixes**
  - Highlighted selection now reliably scrolls into view across the account picker, popups, user search, group detail, and profile/follows.
  - Prevented multiple concurrent live in-memory engine sessions per account.

- **Documentation**
  - Updated `wn tui` help/keybinding guidance for the new `a` flow, `Esc` behavior, and `q` quit semantics.

### DIFF
--- a/crates/cli/src/tui/app.rs
+++ b/crates/cli/src/tui/app.rs
@@ -1711,10 +1711,16 @@ impl TuiApp {
     pub(crate) fn handle_user_search_key(&mut self, key: KeyEvent) -> TuiResult<()> {
         if key.code == KeyCode::Esc {
             // A search opened from group detail goes back there rather than to the
-            // main view, so `Esc` undoes the step that opened it.
+            // main view, so `Esc` undoes the step that opened it. The refresh that
+            // return schedules can fail its account precondition, and a key handler
+            // reports rather than propagates: a `?` here would leave `run()` and end
+            // the session over a status-line error. One assignment covers both
+            // outcomes so the report cannot be overwritten by the settled status.
             if let Some(group_id) = self.search_add_target().map(str::to_owned) {
-                self.return_to_group_detail(&group_id)?;
-                self.status = "group detail".to_owned();
+                self.status = match self.return_to_group_detail(&group_id) {
+                    Ok(()) => "group detail".to_owned(),
+                    Err(err) => format!("error: {err}"),
+                };
                 return Ok(());
             }
             self.leave_screen();
--- a/crates/cli/src/tui/app.rs
+++ b/crates/cli/src/tui/app.rs
@@ -1421,6 +1421,12 @@ impl TuiApp {
             }
             PopupSubmit::AddUserToChat { group_id, pubkey } => {
                 self.add_group_member(&group_id, pubkey)?;
+                // A search opened from group detail returns there, because the
+                // point of adding from that screen is to watch the member land in
+                // its list. A browse search instead reveals the chat it changed.
+                if self.search_add_target().is_some() {
+                    return self.return_to_group_detail(&group_id);
+                }
                 self.select_chat_by_group_id(&group_id)?;
                 self.reveal_chat_from_search();
                 Ok(())
@@ -1462,6 +1468,7 @@ impl TuiApp {
                     view.select_down();
                 }
             }
+            KeyCode::Char('a') => self.search_for_member_to_add(),
             KeyCode::Char('A') => self.open_add_member_popup(),
             KeyCode::Char('R') => self.open_rename_group_popup(),
             KeyCode::Char('x') => self.open_remove_member_popup(),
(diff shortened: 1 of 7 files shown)


Datawav:

[P2] `crates/cli/src/tui/view.rs` — User-search selection can still scroll off-screen.

- Need: The highlighted result and its `>` marker must remain visible, including with narrow terminals or long remote metadata.
- Design issue: `user_search_lines` anchors the attribution line. `selection_scroll_offset` aligns the final wrapped row of that line with the viewport bottom. Because `matched_field` and `match_quality` are sanitized but not length-limited, wrapping can push the preceding highlighted label entirely above the viewport. Existing tests use short attribution values and do not cover this case.
- Alternatives: Anchor the highlighted label and account for the row’s following attribution lines, or clamp the offset so the selected row’s first wrapped line remains visible. Add a narrow-viewport test with long attribution fields.

_Advisory audit of head [`9b3967106047`](https://github.com/marmot-protocol/mdk/commit/9b3967106047813321e746268eb631f8ec307171); no approval, merge, push, checkout, or PR-code execution was performed._

Datawav:
## Nightly deep audit

- [P2] Selected profile rows can still hide their highlight — `crates/cli/src/tui/view.rs:1579`. `PaneBody` stores only the selected logical line, while `selection_scroll_offset` aligns that line’s final wrapped row with the viewport bottom. Profile values such as `about`, picture URL, or other metadata are not shortened or size-bounded. If the selected field wraps to more rows than the pane height, scrolling exposes its tail and moves the leading `>` marker and highlighted field label above the viewport—the exact regression this patch intends to fix. This is durable profile state, so reopening the TUI reproduces it consistently. Model the selection as a wrapped start/end interval: show the entire interval when it fits, otherwise pin its first row so the marker remains visible. Alternatively, render long values on continuation lines while anchoring the labeled first line. Add a narrow, short-buffer test with a selected multi-line profile value longer than the viewport; current tests only exercise short single-line fields.

Datawav (commented):
## Automated review

No actionable findings.

crates/cli/src/tui/app.rs
```diff
@@ -1421,6 +1421,12 @@ impl TuiApp {
             }
             PopupSubmit::AddUserToChat { group_id, pubkey } => {
                 self.add_group_member(&group_id, pubkey)?;
+                // A search opened from group detail returns there, because the
+                // point of adding from that screen is to watch the member land in
+                // its list. A browse search instead reveals the chat it changed.
+                if self.search_add_target().is_some() {
+                    return self.return_to_group_detail(&group_id);
```
Datawav: The member mutation has already succeeded before this refresh error is propagated. If `return_to_group_detail` fails, the TUI exits and presents a completed add as a terminal failure. The `Esc` path below treats the same refresh failure as recoverable; please catch it here too, retain the cached group-detail view, report the refresh failure in `status` without obscuring the successful add, and add a regression test for this post-submit failure path.

Datawav (commented):
## Automated review

Outcome: one actionable failure-path defect remains.

- P2 / Medium — Post-add refresh failure can terminate the TUI after the mutation succeeds. In `crates/cli/src/tui/app.rs` at the changed `PopupSubmit::AddUserToChat` branch, `self.add_group_member(...)` succeeds before `return self.return_to_group_detail(&group_id)` propagates any refresh error. The changed `Esc` path explicitly catches the same error and reports it through `status`, confirming it is recoverable; the added submission test covers only successful refresh. Consequence: the member can be added while the session exits and presents the completed operation as a terminal failure. Smallest correction: catch `return_to_group_detail` errors in the submit branch, retain the cached group-detail screen, report the refresh failure on the status line without obscuring the successful add, and add a regression test for refresh failure after confirmed submission.

crates/cli/src/tui/tests.rs
```diff
@@ -11020,6 +11341,56 @@ fn add_to_open_chat_from_search_navigates_into_that_chat() {
...
+        app.loading_group_detail.as_deref(),
+        Some("g1"),
+        "the group detail reloads so the new member shows"
+    );
+    // The add published, so the status must report that and not be displaced by
+    // the return that follows it. A completed add reported as anything else
+    // would send the user to re-add a member who is already in the group.
+    assert!(
+        app.status.contains("added member(s)"),
+        "the completed add is what the status reports; got: {:?}",
```
Datawav, line 11389: This assertion only covers the successful refresh path and would also pass if `return_to_group_detail` errors were propagated after the add completed. Please inject a post-add Group Detail refresh failure, then assert that submission returns `Ok(())` and the status contains both the add confirmation and `group detail refresh failed`.

Datawav (commented):
## Automated review

Outcome: the runtime failure is fixed, but the previously requested negative-path regression coverage remains unresolved; no new production defect is supported.

- Fixed — `crates/cli/src/tui/app.rs:1428-1436` now catches a failed post-add `return_to_group_detail`, preserves the successful-add status with a refresh warning, and returns `Ok(())` instead of terminating the TUI after the mutation succeeds.
- P3 / Low — `crates/cli/src/tui/tests.rs:11384-11391` exercises only a successful refresh and checks for `"added member(s)"`, which would also pass before the fix. Consequently, reintroducing error propagation after a completed add would escape detection. Smallest correction: inject a post-add Group Detail refresh failure and assert that submission returns `Ok(())` and the status contains both the add confirmation and `"group detail refresh failed"`.

PR #1167, opened 2026-07-29
fix: republish current KeyPackage event

### LINKED ISSUE #1163: `republish_key_package` rotates and retains another package instead of republishing the current event
## Problem

`republish_key_package` is documented and exposed as “re-publish the latest cached KeyPackage when possible,” but its current implementation always creates and publishes a fresh KeyPackage. It is therefore semantically identical to `publish_new_key_package` / rotation.

Repeatedly using **Republish** can leave one current package plus multiple retained private bundles. The inventory then legitimately returns several durable local records, which Android currently flattens into what looks like multiple published packages. This provides a source-confirmed path to the observed three-package state.

Grounded against MDK `origin/master` at `523a193815eafb22560923a3960b70e07373ac93`.

## Root cause

- UniFFI documents `republish_key_package` as reusing the latest cached package when possible (`crates/marmot-uniffi/src/commands/account.rs:276-279`).
- It calls `MarmotAppRuntime::publish_key_package` (`crates/marmot-app/src/runtime/mod.rs:2069-2074`).
- `AppClient::publish_key_package` unconditionally calls `publish_fresh_key_package` (`crates/marmot-app/src/client/mod.rs:235-245`).
- `rotate_key_package` simply calls that same method (`client/mod.rs:494-495`), while `publish_new_key_package` aliases rotation (`runtime/mod.rs:2086-2088`).
- The relay-runtime test currently codifies this collapse: after an explicit rotation, calling `publish_key_package` changes both package bytes and `KeyPackageRef` while retaining the stable `d` slot (`crates/marmot-app/tests/relay_runtime.rs:1031-1085`).

The lifecycle added by #1092 / PR #1103 is otherwise doing important safety work:

- A replaced, unconsumed package is retained so already-built Welcomes can still be processed.
- Consumed material is deleted after replacement acknowledgement.
- Unconsumed retained material is swept at `Lifetime.not_after`, including while network maintenance is paused (`crates/marmot-account/src/runtime.rs:648-702`).

Deleting those retained bundles merely to force the inventory count to one would be unsafe. The incorrect mutation boundary is that **Republish** rotates in the first place.

## Required behavior

- `republish_key_package` republishes the current durable authored event and KeyPackage under the existing stable slot without generating a new KeyPackage, `KeyPackageRef`, event id, authored timestamp, or private bundle.
- Reuse the exact durable signed artifact so restart and ambiguous-exposure semantics remain deterministic.
- Send it to the current authoritative KeyPackage relay targets; merge/update per-target fanout status without opening a replacement lifecycle.
- If no current usable package/artifact exists, return a typed “rotation required” result or explicitly fall back to fresh publication as documented, with that fallback covered and observable.
- `publish_new_key_package` remains the explicit rotation operation and keeps the existing retained-private-material safety rules.
- Do not prune retained bundles solely to make diagnostics show one row.

## Regression coverage

1. Publish an initial package; capture its stable slot, `KeyPackageRef`, event id, signed bytes, lifecycle state, and durable bundle count.
2. Call `republish_key_package`.
3. The same event/package is sent again, and all captured identities plus durable bundle count and `retained_private_material` remain unchanged.
4. Repeat after runtime restart and after partial relay fanout; retries still use the exact authored event.
5. Call `publish_new_key_package`; it produces a new `KeyPackageRef` under the same stable slot and follows the normal retained-material lifecycle.
6. Split/update `app_runtime_can_rotate_key_package_on_request` so it no longer expects `publish_key_package` to rotate.

## Related

- Android inventory presentation: https://github.com/marmot-protocol/whitenoise-android/issues/1696
- Open Android fix: https://github.com/marmot-protocol/whitenoise-android/pull/1711
- Typed local/relay lifecycle inventory: #1154
- Durable consumed-package retirement: #1092 / #1103

agent-p1p:
Fixes #1163

## Summary
- Change explicit KeyPackage publish/republish to resend the exact durable current signed event and KeyPackage instead of staging a replacement.
- Preserve explicit rotation as the only path that mints a new KeyPackage, while retaining fallback rotation for missing, consumed, expired/refresh-due, or otherwise unusable current artifacts.
- Reconcile current authoritative publication targets across restart and policy churn, reject republish while a replacement is pending, and preserve lifecycle/private-material identity on normal republish.
- Treat a NIP-01 `duplicate:` relay acknowledgement as idempotent publication success without exposing relay-controlled rejection text.
- Update app/client/CLI wiring and add account, app-runtime, transport, and CLI regressions.

## Verification
- `cargo fmt --all --check`
- `./scripts/check_legacy_naming.sh`
- `cargo test -p marmot-account --locked -- --test-threads=1` — 89 passed
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked -- --test-threads=1` — 81 passed
- `cargo test -p transport-nostr-adapter --features sdk --locked -- --test-threads=1` — 85 passed
- `cargo nextest run -p wn-cli --test cli --features test-policy-overrides --locked --profile ci` — 88 passed
- `RUSTFLAGS='-D warnings' cargo check --workspace --all-targets --all-features --locked`
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- `cargo test -p cgka-engine --test convergence_policy_pin --locked` — 5 passed
- Independent adversarial review: PASS, no blocking findings

## Sensitive paths
- `crates/marmot-account/src/runtime.rs` — KeyPackage lifecycle, signed artifact reuse, publication-target state
- `crates/marmot-app/src/client/mod.rs`
- `crates/marmot-app/src/runtime/mod.rs`
- `crates/transport-nostr-adapter/src/sdk_client.rs` — relay publication acknowledgement handling

Sensitive code touched. Explicit human approval is required at the merge gate.

## Summary by CodeRabbit

- **New Features**
  - Key package publishing now republishes the current package when appropriate, preserving its identity and durable state.
  - Added explicit key package rotation for creating a fresh package.
  - Added access to durably owned key packages.
- **Bug Fixes**
  - Republish operations now reconcile relay changes and retry incomplete distribution.
  - Duplicate relay acknowledgements are treated as successful, avoiding false publication failures.
  - Added clearer handling when key package rotation is already in progress.

### DIFF
--- a/crates/cli/tests/cli.rs
+++ b/crates/cli/tests/cli.rs
@@ -2701,7 +2701,7 @@ fn key_package_fetches_latest_package_via_relay_list_discovery() {
 }
 
 #[test]
-fn keys_publish_replaces_create_identity_key_package_under_stable_slot() {
+fn keys_publish_republishes_create_identity_key_package_under_stable_slot() {
     let home = tempfile::tempdir().expect("tempdir");
     let relay = test_relay_url();
 
@@ -2721,7 +2721,11 @@ fn keys_publish_replaces_create_identity_key_package_under_stable_slot() {
     assert_eq!(republished["key_package_bytes"], first["key_package_bytes"]);
     assert_eq!(second["key_package_bytes"], first["key_package_bytes"]);
     assert_eq!(second["key_package_id"], first["key_package_id"]);
-    assert_ne!(second["key_package_ref"], first["key_package_ref"]);
+    assert_eq!(second["key_package_ref"], first["key_package_ref"]);
+    assert_eq!(
+        second["key_package_event_id"],
+        first["key_package_event_id"]
+    );
     assert!(
         first["key_package_id"]
             .as_str()
@@ -2735,7 +2739,7 @@ fn keys_publish_replaces_create_identity_key_package_under_stable_slot() {
 }
 
 #[test]
-fn keys_rotate_and_publish_each_replace_under_the_stable_slot() {
+fn keys_rotate_changes_ref_and_publish_republishes_current_ref() {
     let home = tempfile::tempdir().expect("tempdir");
     let relay = test_relay_url();
 
@@ -2760,10 +2764,18 @@ fn keys_rotate_and_publish_each_replace_under_the_stable_slot() {
 
     assert_eq!(second["key_package_id"], first["key_package_id"]);
     assert_ne!(second["key_package_ref"], first["key_package_ref"]);
+    assert_ne!(
+        second["key_package_event_id"],
+        first["key_package_event_id"]
+    );
     assert_eq!(second["key_package_bytes"], rotated["key_package_bytes"]);
     assert_eq!(third["key_package_bytes"], second["key_package_bytes"]);
     assert_eq!(third["key_package_id"], second["key_package_id"]);
-    assert_ne!(third["key_package_ref"], second["key_package_ref"]);
+    assert_eq!(third["key_package_ref"], second["key_package_ref"]);
+    assert_eq!(
+        third["key_package_event_id"],
+        second["key_package_event_id"]
+    );
     assert!(
         third["key_package_id"]
             .as_str()
@@ -2845,25 +2857,17 @@ fn keys_list_reports_published_key_package() {
     run_json(home.path(), &["--account", &account_id, "keys", "publish"]);
 
     // With a published KeyPackage on the reachable relay, `keys list` must
-    // surface it rather than returning an empty list.
+    // surface the merged local+relay row for the current package rather than
+    // returning an empty list or a separate retained duplicate.
     let listed = run_json(home.path(), &["--account", &account_id, "keys", "list"]);
     assert_eq!(listed["account_id"], account_id);
     let keys = listed["keys"].as_array().expect("keys array");
     assert_eq!(
         keys.len(),
-        2,
-        "expected the current relay-visible package and retained local package, got {keys:?}"
-    );
-    let relay_keys = keys
-        .iter()
-        .filter(|key| key["relay"] == true)
-        .collect::<Vec<_>>();
-    assert_eq!(
-        relay_keys.len(),
         1,
-        "expected the published key package, got {keys:?}"
+        "expected one merged current key package inventory row, got {keys:?}"
     );
-    let published = relay_keys[0];
+    let published = &keys[0];
     assert_eq!(published["account_id"], account_id);
     assert!(
         published["key_package_event_id"]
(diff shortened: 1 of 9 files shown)


agent-p1p:
## Pip adversarial review — BLOCKING

Reviewed exact head `c9e8fe5571604b5ce89c27430a8ec63d5e653ff1`.

### Blocking

1. **`crates/marmot-account/src/runtime.rs:3125-3128` — reaching the proactive refresh deadline still rotates from the Republish path.**

   `current_key_package_republishable` requires `refresh_at > now`; once the sampled refresh deadline arrives, `republish_key_package` falls through to `publish_fresh_key_package`. At that point the function has already established `current_not_after > now`, so the current package and exact signed artifact are still valid. The fallback mints a new KeyPackage/ref/event and retains another private bundle — the mutation that #1163 requires Republish to avoid. The crate guidance also says normal publish reuses a cached valid package, with expired or policy-invalid packages rotating; a proactive maintenance deadline does not make the package invalid.

   `crates/marmot-account/tests/runtime.rs:1451-1492` currently codifies the regression by asserting that Republish at `refresh_at` creates a different package and one retained bundle. Leave proactive `refresh_at` rotation to `run_due_maintenance`; explicit Republish should reuse the exact current artifact until it is expired, consumed, missing, or otherwise policy-invalid. Replace that test with an identity/bundle-count preservation assertion at `refresh_at`.

### Counts

- Blocking: 1
- Suggestions: 0
- Nitpicks: 0

### Independent verification

- `cargo fmt --all --check`
- `cargo test -p marmot-account --locked -- --test-threads=1` — 89 passed
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked -- --test-threads=1` — 81 passed
- `cargo test -p transport-nostr-adapter --features sdk --locked -- --test-threads=1` — 85 passed
- `cargo clippy -p marmot-account -p marmot-app -p transport-nostr-adapter --all-targets --all-features --locked -- -D warnings`

All GitHub checks are green. The suite passes because the refresh-deadline case currently asserts the wrong Republish semantics.

agent-p1p:
Addressed the blocking refresh-deadline finding in `c73e3943`. Explicit Republish now ignores the proactive `refresh_at` maintenance deadline and reuses the exact current KeyPackage/publication identity until the package is actually invalid (for example expired, consumed, missing, or upgrade-ineligible). The regression test now preserves the authored event, KeyPackage ref, retained-material state, and durable bundle set at `refresh_at`.

Verified locally:
- `cargo fmt --all --check`
- `./scripts/check_legacy_naming.sh`
- `cargo test -p marmot-account --locked -- --test-threads=1` (89 passed)
- CI-exact workspace check/clippy commands, default + all features

agent-p1p:
@coderabbitai review

agent-p1p:
Addressed all three CodeRabbit nitpicks in `7e8a4d78`:
- added the silent-relay `false, None` characterization;
- renamed the mock-relay test so it no longer overclaims that the duplicate-response branch ran;
- collapsed duplicate republish eligibility/reason logic into one blocker classifier (net code deletion).

Final local verification is green: formatting, naming gate, 89 marmot-account tests, 85 transport adapter SDK tests, and CI-exact workspace check/clippy with default and all features.

PR #1168, opened 2026-07-29
Add UniFFI follow management

erskingardner:
## Summary

Adds UniFFI APIs for following and unfollowing users, checking follow membership, and retrieving an account's full follow list.

The write operations preserve the current kind-3 list through a serialized read-modify-publish flow. If no current list is available, they return a typed error instead of publishing an unintended replacement. Cached directory data is refreshed immediately so the read APIs remain fast and local.

## Validation

- `just fast-ci`
- `git diff --check`

`cargo test -p marmot-app` completed 514 unit tests, but five unrelated existing `relay_runtime` integration failures remain reproducible in media, retention, and group-removal paths.

## Summary by CodeRabbit

* **New Features**
  * Added cached follow-list viewing and following-status checks.
  * Added follow and unfollow actions that preserve and return the updated follow list.
  * Follow-list updates are synchronized and published only when changes occur.

* **Bug Fixes**
  * Follow lists now normalize account identifiers and remain available after publishing.
  * Added a specific error when the current follow list is unavailable.

* **Chores**
  * Updated the application version to 0.9.11.

### DIFF
--- a/crates/marmot-app/src/runtime/mod.rs
+++ b/crates/marmot-app/src/runtime/mod.rs
@@ -106,6 +106,7 @@ pub struct MarmotAppRuntime {
     events: broadcast::Sender<MarmotAppEvent>,
     shared: RuntimeSharedServices,
     accounts: AccountManager,
+    follow_list_updates: Arc<Mutex<()>>,
     directory_sync: Arc<Mutex<Option<DirectorySyncHandle>>>,
     initial_directory_sync: Arc<Mutex<Option<JoinHandle<()>>>>,
 }
@@ -820,6 +821,7 @@ impl MarmotAppRuntime {
             events,
             shared,
             accounts,
+            follow_list_updates: Arc::new(Mutex::new(())),
             directory_sync: Arc::new(Mutex::new(None)),
             initial_directory_sync: Arc::new(Mutex::new(None)),
         }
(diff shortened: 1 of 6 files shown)


crates/marmot-app/src/runtime/mod.rs
```diff
@@ -2587,6 +2589,118 @@ impl MarmotAppRuntime {
...
+    async fn set_following(
+        &self,
+        account_ref: &str,
+        user_account_id_hex: &str,
+        following: bool,
+    ) -> Result<Vec<String>, AppError> {
+        // Kind-3 is a whole-list replaceable event. Serialize updates made
+        // through this runtime so two local UI actions cannot both fetch the
+        // same snapshot and then overwrite one another.
+        let _update_guard = self.follow_list_updates.lock().await;
```
erskingardner: Minor: this mutex is a single, runtime-wide lock (`follow_list_updates: Arc<Mutex<()>>`), so a follow/unfollow on account A blocks a concurrent follow/unfollow on unrelated account B, not just concurrent updates on the same account. The read-modify-publish race this guards against is genuinely per-account (two updates to the *same* kind-3 event), so a keyed lock (e.g. `Mutex<HashMap<String, Arc<Mutex<()>>>>`, mirroring the existing `workers: Arc<Mutex<HashMap<String, ManagedAccountWorker>>>` pattern) would avoid serializing unrelated accounts' follow actions behind one another's network round trips. Not blocking — likely fine for typical single/few-account usage — just flagging since it's a new locking pattern in this file.

erskingardner (commented):
Reviewed the diff and the underlying `MarmotApp`/`MarmotAppRuntime` plumbing it builds on (`fetch_current_follow_list_for_account_id`, `publish_account_follow_list`, `directory_entry_for_account_id`, `normalize_account_ids`).

**Recommend merge.** Clean, well-scoped addition. The read-current → modify → publish-only-if-changed flow correctly refuses to treat "no kind-3 event found" as "empty list" (`AppError::FollowListUnavailable`/`MarmotKitError::FollowListUnavailable`), which avoids silently clobbering a follow list published from another client. `account_follows`/`is_following` are genuinely network-free (cache-only reads), and the cache is kept warm on both the publish and refresh paths. Follow-list identifiers are normalized to canonical lowercase hex end-to-end (`parse_account_id_hex`/`normalize_account_ids`), so the string-equality membership checks in `set_following`/`is_following` are safe. Good test coverage: happy path (follow, unfollow, cache reflects both) and the "no current list" error path, both against a real `MockRelay`.

One non-blocking note left inline: `follow_list_updates` is a single runtime-wide `Mutex<()>`, so it serializes follow/unfollow calls across *all* accounts, not just the same account — more contention than the race it's guarding against requires. Not a correctness problem, just flagging for awareness.

(Note: GitHub won't let me submit this as an "Approve" since I'm authenticated as the PR author — posting as a comment instead. No blocking issues found; this is an approve-quality review.)

PR #1169, opened 2026-07-29
Add Vertex Open Ranking search and relay policy APIs

erskingardner:
## Summary

- replace the Vertex NIP-50 relay search with an ORE-05 `POST /search/pubkeys` request to `ranking.vertexlab.io`
- cap provider results at 20 and bound the complete ranking request, including DNS and bounded `202 Retry-After` retries, to five seconds
- validate, normalize, rank, and deduplicate returned pubkeys locally
- merge ranked identities with the completed social-graph traversal so graph results win, graph provenance is preserved, and already-emitted identities are removed before hydration
- hydrate only the remaining pubkeys with one author-scoped kind-0 fetch; the HTTP endpoint and hydration relays are configurable through `MarmotServiceEndpoints`, and Rust hosts can override or disable the discovery tier
- keep dev/test convenience construction offline by disabling Open Ranking, avoiding live Vertex dependencies in the test suite
- expose optional provider rank and a dedicated `DiscoveryResultsFound` trigger through the app and UniFFI result model
- remove retired bootstrap relays, keep Vertex in the bootstrap set, and reject retired hosts at the central relay-plane safety boundary
- expose `retired_relay_hosts()` and batch `classify_relay_endpoints()` through MarmotKit, returning `allowed`, `retired`, `invalid`, or `unsafe` plus the normalized URL when parsing succeeds
- prepare the existing `0.9.10` workspace/binding cohort for release with a dated changelog section

## Client flow

Clients can continue to call `account_relay_lists()` for the published NIP-65 and inbox lists, combine those endpoint strings, and classify them in one batch. The account relay-list DTO remains unchanged. This keeps policy classification composable and avoids presenting transient relay-plane connection state as persistent relay health.

The classifier delegates to the same `RelaySafetyPolicy` used at every relay-plane dial boundary, including the runtime's explicit loopback-development setting, so the UI result cannot drift from enforcement. Input order and cardinality are preserved.

## Safety and behavior

The Open Ranking HTTP client requires HTTPS, disables redirects, proxies, and compression, resolves and validates every address before dialing, pins the validated addresses while retaining hostname-based TLS trust, and caps response bodies at 64 KiB. Provider failures remain best-effort and do not fail social-graph search. No ranked result is persisted or promoted into the directory.

Open Ranking retrieval starts alongside graph traversal, but discovery results are merged after graph completion so graph provenance and graph-wins deduplication remain deterministic. Discovery rows use their own update trigger rather than arriving late in a completed graph-radius bucket.

## Validation

- `just fast-ci`
- `cargo test -p marmot-app --lib` — 528 passed
- `cargo test -p marmot-uniffi` — 88 unit tests, 2 media-fixture tests, and 26 smoke tests passed
- `cargo test -p wn-cli tui::tests:: --lib` — 419 passed
- `./crates/marmot-uniffi/xcframework.sh` — generated the `0.9.10` Apple XCFramework and Swift binding (optional `swiftformat` was unavailable)
- `./crates/marmot-uniffi/kotlin-bindings.sh` — generated the `0.9.10` Kotlin binding and all four Android JNI ABIs
- generated Swift and Kotlin outputs contain the relay-policy methods, typed classification model, provider rank, and discovery trigger
- live Vertex contract smoke check — HTTP 200, 20 results, numeric ranks, TTL 300

A full `cargo test -p marmot-app` run passed all library tests. The existing `relay_runtime` integration binary still has five failures in encrypted-media, retention, and self-removal scenarios outside this diff; the two encrypted-media failures also reproduce when run alone with one test thread.

## Summary by CodeRabbit

* **New Features**
  * Added ranked off-graph profile discovery to user search, with deduplication and provider-based ranking.
  * Search results now identify discovery provenance and may include provider ranking metadata.
  * Added relay endpoint classification for allowed, retired, invalid, and unsafe endpoints.
  * Added APIs for viewing retired relay hosts and classifying relay endpoints.
* **Bug Fixes**
  * Retired relays are now rejected during discovery, bootstrap, adoption, and directory operations.
  * Updated default discovery relays to use safer, current endpoints.
* **Documentation**
  * Updated release notes and user-search attribution guidance.


crates/marmot-app/src/directory/search.rs
```diff
@@ -75,14 +75,26 @@ const SEARCH_PUBKEY_BATCH_SIZE: usize = 200;
 /// Ceiling on the relay work a single radius may spend.
 const SEARCH_RADIUS_TIMEOUT: Duration = Duration::from_secs(300);
 
+/// The one NIP-50 relay currently used to widen a user-directory search.
+///
+/// This is a best-effort discovery source, not a directory sync relay: its
+/// results stay in the current search only and never create subscriptions or
+/// directory records.
+const VERTEX_NIP50_RELAY: &str = "wss://relay.vertexlab.io";
```
erskingardner: **Hardcoded, unconfigurable third-party relay — no opt-out, and breaks test hermeticity.**

`VERTEX_NIP50_RELAY` sends every 3+ character search query verbatim to one specific external company's production relay (`relay.vertexlab.io`), unconditionally, with no config knob, no override, and no way for a host app to disable or repoint it.

This is inconsistent with precedent already in this file/crate:
- `directory_search_fallback_seeds` (`crates/marmot-app/src/config.rs:93`) is empty-by-default *specifically* because "whose network to fall back to is a deployment decision, not protocol behaviour" — the exact same category of decision as this NIP-50 relay.
- `MarmotServiceEndpoints` in `config.rs` exists precisely so host apps can override built-in first-party service URLs (see the `profile_image_blob_endpoint` doc comment: "hosts with different privacy or availability requirements should override the built-in service"). This new endpoint sends raw user search text to a third party and arguably needs that override capability even more than the others.

Concretely, because there's no seam to substitute a mock/local relay here, this also silently gives the existing test suite a live network dependency: every `search_users` test in this file using a 3+ char query (the many `"needle"` tests, e.g. lines ~1177, 1301, 1368, 1443, 1479, 1744, 1805, 1871, 1936, 2061) will now attempt a real connection to `relay.vertexlab.io` on every `cargo test` run — regardless of the `wss://relay.invalid`/`MockRelay` URL the test configured for the app's *own* relay, since this constant is entirely independent of that. In a network-restricted sandbox that's up to ~5s (`NIP50_SEARCH_TIMEOUT`/`DIRECTORY_RELAY_CONNECT_WAIT`) added per affected test; with network access, it's a real outbound call to a third party's production server from CI.

Suggest making this endpoint configurable (e.g. added to `MarmotServiceEndpoints`, default-off or default-empty like the fallback seeds) so it can be disabled and so it can be pointed at a mock relay in tests.

crates/marmot-app/src/directory/search.rs
```diff
@@ -245,26 +257,124 @@ async fn run_search(
...
-                message: error.to_string(),
-            })
-            .await;
+    if !query.is_empty() {
+        // Start the independent discovery request with the graph walk, but
+        // merge its results afterwards. That gives a graph result precedence
+        // when both sources return the same identity, without holding a
+        // mutable emitter across concurrent tasks.
+        let nip50_updates_tx = emitter.updates_tx.clone();
+        let (graph_result, nip50_records) = tokio::join!(
```
erskingardner: Because `nip50_records` is only read after this `tokio::join!` resolves — i.e. after `traverse_graph` has *fully* finished (which can take up to `SEARCH_RADIUS_TIMEOUT` = 300s per radius) — a NIP-50 result that's ready in under a second still won't reach the UI until the whole graph walk completes. That undercuts the "bounded, fast discovery" framing for a slow/large graph search. Consider emitting nip50 matches as soon as they resolve (checking `emitted_account_ids` at emit time preserves the "graph wins on conflict" precedence) instead of deferring them until after the join.

erskingardner (commented):
**Overall: found serious issues (would request changes, but GitHub disallows that on your own PR).**

## Summary

Solid, well-scoped implementation of the merge/dedup/provenance logic (graph traversal and NIP-50 discovery run concurrently via `tokio::join!`, dedup keyed by `emitted_account_ids`, signature/kind/freshness checks reused, results stay ephemeral and never enter the persistent directory — nice test at `nip50_profiles_stay_out_of_the_persistent_directory`). The two blocking points below are about the same root cause: `VERTEX_NIP50_RELAY` is a bare, unconfigurable constant, which is inconsistent with how this codebase already handles "reach outside the user's own graph/relays" (`directory_search_fallback_seeds`, `MarmotServiceEndpoints`) and has a concrete, verifiable side effect on the test suite.

One non-blocking design note on emission ordering, and a missing CHANGELOG entry for a user-facing change.

erskingardner:
One more non-blocking note that didn't fit as an inline comment: `crates/cli/CHANGELOG.md` tracks "all notable user-facing changes ... including MarmotKit bindings" under `[Unreleased]`. This PR changes user-facing TUI search behavior (new "discovery" provenance label) and FFI-exposed search semantics (marmot-uniffi), so per that convention it probably wants an Unreleased entry alongside the code change.

crates/marmot-app/src/directory/open_ranking.rs
```diff
@@ -0,0 +1,295 @@
...
+    pubkey: String,
+    rank: f64,
+}
+
+/// Ask Vertex's Open Ranking provider for ranked pubkeys.
+///
+/// The entire operation, including DNS and any protocol-mandated `202` retry,
+/// is bounded to five seconds. Results are normalized, deduplicated, and sorted
+/// locally before they cross into directory search.
+pub(crate) async fn search_vertex_pubkeys(query: &str) -> Result<Vec<RankedPubkey>, String> {
```
erskingardner: This function has no test seam at all: it does a live DNS lookup and HTTPS POST to `https://ranking.vertexlab.io/search/pubkeys` unconditionally. Every non-empty-query `search_users` call in this crate's test suite (`crates/marmot-app/src/directory/search.rs`) reaches this function, so `cargo test -p marmot-app` makes real outbound network requests to a third-party service on every run — the same hermeticity concern raised last review, just moved from the NIP-50 relay to this HTTP endpoint. Worth an injectable base URL / mock seam, or a way to disable the provider under test, so these tests don't depend on Vertex's uptime or live index contents.

crates/marmot-app/src/directory/search.rs
```diff
@@ -245,26 +246,122 @@ async fn run_search(
...
+async fn hydrate_open_ranking_profiles(
+    app: &MarmotApp,
+    ranked_pubkeys: Vec<RankedPubkey>,
+) -> Result<Vec<RankedDirectoryRecord>, AppError> {
+    let account_ids = ranked_pubkeys
+        .iter()
+        .map(|ranked| ranked.account_id_hex.clone())
+        .collect::<Vec<_>>();
+    let records = app
+        .fetch_events_for_account_ids(
```
erskingardner: This hardcodes `VERTEX_DIRECTORY_RELAY` (`wss://relay.vertexlab.io`) as the fetch source regardless of the app's configured relay. Every other test in this file uses `MarmotApp::with_relay(dir.path(), "wss://relay.invalid")` specifically so relay round-trips are guaranteed no-ops; this call bypasses that convention and dials the real Vertex relay whenever `fetch_open_ranking_pubkeys` returns anything. Combined with the previous comment, a test like `streams_a_cached_radius_zero_match_through_the_full_update_sequence` (which asserts an exact 4-trigger sequence for query "needle") is only passing today because Vertex happens not to return a match for that fixture string — that's non-hermetic by construction, not by luck holding.

crates/cli/src/tui/view.rs
```diff
@@ -1419,12 +1419,16 @@ pub(crate) fn user_search_lines(view: &UserSearchView, searching: bool) -> Vec<L
             ));
         }
         lines.push(Line::from(spans));
+        let provenance = if result.radius == u8::MAX {
```
erskingardner: Non-blocking, carried over from last round: this still compares against the bare `u8::MAX` literal instead of importing `marmot_app::OFF_GRAPH_SEARCH_RADIUS`, which is already exported publicly (see `crates/marmot-app/src/lib.rs`). One-line fix; leaving the literal duplicated risks silent drift if the sentinel ever changes.

crates/marmot-app/src/directory/open_ranking.rs
```diff
@@ -0,0 +1,295 @@
...
+            .await
+            .map_err(open_ranking_reqwest_error)?;
+        if response.status() == reqwest::StatusCode::ACCEPTED {
+            let retry_after = response
+                .headers()
+                .get(RETRY_AFTER)
+                .and_then(|value| value.to_str().ok())
+                .and_then(|value| value.parse::<u64>().ok())
+                .ok_or_else(|| "Open Ranking 202 response omitted Retry-After".to_owned())?;
+            if retry_after == 0 {
```
erskingardner: Non-blocking: the 202 retry loop has no explicit attempt cap and treats `Retry-After: 0` as "spin via `yield_now`" rather than enforcing a minimum backoff. It's bounded transitively by the outer 5s `timeout`, so this can't hang, but a response that keeps returning `202` with `Retry-After: 0` will hammer the third party with as many requests as fit in 5 seconds with no backoff at all.

erskingardner (commented):
**Overall: found serious issues (would request changes, but GitHub disallows that on your own PR).**

## Summary

This is a genuine improvement over the previous revision: NIP-50 and the bare `VERTEX_NIP50_RELAY` are gone, replaced with a bounded Open Ranking HTTP client (HTTPS-only, redirects/proxy/compression disabled, DNS-resolved-and-pinned, 64 KiB response cap, 5s overall timeout). The merge/dedup/provenance logic between graph traversal and off-graph ranking is careful (`graph_radii`/`emitted_account_ids` dedup, provider rank suppressed once graph provenance applies, ephemeral-only hydration verified by `open_ranking_profiles_keep_graph_provenance_and_stay_ephemeral`). The new `retired_relay_hosts()` / `classify_relay_endpoints()` bindings are clean, fully offline, reuse the same `RelaySafetyPolicy` enforced at every dial boundary, preserve input order/cardinality, and are exercised end-to-end (unit tests, FFI conversion test, UniFFI smoke test) — no concerns there.

However, the two previously-raised blocking points are not actually resolved — the underlying problem (an always-on, unconfigurable, hardcoded call to a third-party Vertex endpoint) has moved to the new provider path rather than being fixed:

- **Test hermeticity regression, still present.** `search_vertex_pubkeys` (`crates/marmot-app/src/directory/open_ranking.rs:56`) does a real DNS lookup and HTTPS POST to `https://ranking.vertexlab.io/search/pubkeys` with no test seam, mock, or feature flag. It is reached by every `search_users` call with a non-empty query — roughly 15+ existing tests in `crates/marmot-app/src/directory/search.rs` (e.g. `streams_a_cached_radius_zero_match_through_the_full_update_sequence`, which asserts an *exact* 4-item trigger sequence) search for literal strings like `"needle"`. If Vertex's live index ever returns a match for one of these fixture strings, hydration kicks in and the exact-sequence assertions become flaky/fail. Worse: `hydrate_open_ranking_profiles` (`search.rs:330`) hardcodes `VERTEX_DIRECTORY_RELAY` (`wss://relay.vertexlab.io`) as the fetch source regardless of the app's configured relay, so it bypasses the `MarmotApp::with_relay(dir.path(), "wss://relay.invalid")` convention every other test in this file relies on to guarantee no real network I/O. This is the same root cause flagged last review (a bare, unconfigurable outbound dependency baked into the search path) — it just changed shape from a Nostr relay to an HTTP endpoint + relay pair.
- **No opt-out.** There's no config flag anywhere in `MarmotAppConfig` to disable Open Ranking. Every user-search query is unconditionally sent to a third party in production, and unconditionally exercised (with live network calls) in the test suite, with no way for a deployment — or CI — to turn it off.

Two smaller, non-blocking items:

- `crates/cli/src/tui/view.rs:1422` still compares against the bare `u8::MAX` literal instead of the already-exported `marmot_app::OFF_GRAPH_SEARCH_RADIUS` constant. CodeRabbit flagged this last round; it's a one-line fix and the constant is already public, so leaving it duplicated invites future drift.
- `open_ranking.rs`'s `202 Retry-After` loop (lines ~79-99) has no explicit retry-count cap and treats `Retry-After: 0` as "spin via `yield_now`" rather than a minimum backoff — bounded transitively by the outer 5s timeout, so low severity, but still means a misbehaving/adversarial response can make many rapid requests against the third party within that window.

## Recommendation

Give the Open Ranking path the same test-seam treatment the rest of the directory-search code already has (an injectable base URL/relay, or a way to disable the provider under test) so unit tests never depend on a live third-party service's index or uptime, and add a config toggle so the Open Ranking dependency is opt-in rather than baked in unconditionally.

erskingardner:
Review pass complete in `fd1c5ad4`: provider configuration/test hermeticity, bounded 202 retries, discovery-specific update ordering, the shared off-graph sentinel, AGENTS guidance, and the 0.9.10 changelog entry are addressed. I also refreshed the PR description and validation counts. All nine existing review threads are resolved; CI and the refreshed CodeRabbit pass are still running.

PR #1170, opened 2026-07-29
Publish empty follow list for new identities

erskingardner:
## Summary

- publish an empty kind-3 contact list when MDK creates a brand-new local identity
- use the existing follow-list publisher so the event has empty content and no `p` tags, and the local directory records a known-empty list
- leave imported identities, existing-account logins, and external signers unchanged
- document the additional identity-setup publication

## Why

Follow and unfollow updates correctly refuse to treat a missing kind-3 event as an empty list, because kind-3 is a whole-list replaceable event and doing so could overwrite follows published elsewhere. A newly generated identity had no kind-3 event at all, so that guard also prevented its first follow update.

Publishing an explicit empty kind-3 during fresh identity setup establishes a safe relay-visible baseline: later updates can distinguish `Some([])` from a genuinely missing event (`None`) without weakening the anti-clobber guard.

## Validation

- `just fast-ci`
- `cargo test -p marmot-app --test relay_runtime app_runtime_create_identity_bootstraps_managed_account_and_key_package -- --exact`

A full `cargo test -p marmot-app` run reaches the existing `relay_runtime` failures in encrypted-media, retention-system-row, retained-media, and self-removal tests. The encrypted-media endpoint failure was reproduced unchanged from an untouched `master` worktree, independently of this branch.

## Summary by CodeRabbit

- **New Features**
  - New identities now publish an initial empty follow list during setup.
  - Follow-list publication uses the configured bootstrap and default relays.
  - Setup failures continue to support rollback when enabled.

- **Tests**
  - Added verification that newly created identities have an empty follow list.

- **Documentation**
  - Updated identity-creation documentation to include the initial follow list.

### DIFF
--- a/crates/marmot-app/src/runtime/mod.rs
+++ b/crates/marmot-app/src/runtime/mod.rs
@@ -3740,6 +3740,31 @@ impl AccountManager {
             }
         };
 
+        if creates_new_private_key && account.local_signing {
+            self.shared.lifecycle().ensure_running()?;
+            if let Err(err) = self
+                .app
+                .publish_account_follow_list(
+                    &account.label,
+                    &[],
+                    AccountRelayListBootstrap::new(
+                        request.default_relays.clone(),
+                        request.bootstrap_relays.clone(),
+                    ),
+                )
+                .await
+            {
+                if rollback_on_setup_failure {
+                    return self.rollback_import_after_setup_failure(
+                        &account,
+                        private_key_import.as_ref(),
+                        err,
+                    );
+                }
+                return Err(err);
+            }
+        }
+
         let profile = if creates_new_private_key && account.local_signing {
             self.shared.lifecycle().ensure_running()?;
             match self
--- a/crates/marmot-app/tests/relay_runtime.rs
+++ b/crates/marmot-app/tests/relay_runtime.rs
@@ -836,6 +836,16 @@ async fn app_runtime_create_identity_bootstraps_managed_account_and_key_package(
         fetched.key_package.bytes().len(),
         created.key_package_bytes.unwrap()
     );
+    assert_eq!(
+        app.fetch_current_follow_list_for_account_id(
+            &created.account.account_id_hex,
+            vec![endpoint(&url)],
+        )
+        .await
+        .unwrap(),
+        Some(Vec::new()),
+        "a new identity should publish a kind-3 event with no p tags"
+    );
 
     runtime.shutdown().await;
 }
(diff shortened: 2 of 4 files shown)


erskingardner (commented):
Recommend merge — clean, well-scoped fix, no blocking issues found.

Traced the guard it addresses in `set_following` (`fetch_current_follow_list_for_account_id(...).ok_or(AppError::FollowListUnavailable)?`) — a brand-new identity had no kind-3 event at all (`None`), so `follow_user`/`unfollow_user` were unusable until some other client published one. Publishing an explicit empty kind-3 at setup makes `None` vs `Some([])` distinguishable, exactly as described in the PR body.

Strengths:
- The new step is a precise copy of the existing profile/key-package publish pattern (same bootstrap construction, same rollback wiring), so it reads as "the fourth step of the same list," not a new shape.
- Scoping to `creates_new_private_key && account.local_signing` correctly excludes imports, reactivated accounts, and external signers, matching the stated intent (an imported nsec may already have a kind-3 published elsewhere; publishing empty there would risk exactly the clobber this guard exists to prevent).
- The added test asserts `Some(Vec::new())` (not just success), which actually exercises the `None`-vs-`Some([])` distinction the PR is about, and would fail without the fix.
- Docs and the UniFFI doc comment were updated in the same commit to reflect the new setup step and its position in the sequence.

One non-blocking note left inline: rollback doesn't retract the follow-list publish if a later setup step fails. The same gap already exists for the profile/relay-list publish steps this mirrors (not introduced by this PR), and is covered by the general tracking issue #711, so it shouldn't block this fix — flagging for visibility/tracking only.

Validated per the PR description: `just fast-ci` + the targeted `relay_runtime` test; CI is green.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3740,6 +3740,31 @@ impl AccountManager {
...
+                if rollback_on_setup_failure {
+                    return self.rollback_import_after_setup_failure(
+                        &account,
+                        private_key_import.as_ref(),
+                        err,
+                    );
+                }
+                return Err(err);
+            }
+        }
```
erskingardner, line 3766: Non-blocking: if a later step (profile publish, key-package publish) fails after this succeeds, rollback_import_after_setup_failure/rollback_account_after_setup_failure only remove the local account — they don't retract this kind-3 event from the relays it reached. That leaves an orphaned empty follow list published under a pubkey nobody will ever use again (the keychain secret is deleted on rollback, so no future NIP-09 retraction is even possible).

That said, this is the same gap the profile-publish and relay-list-publish steps already have (predates this PR), and it's covered by the general tracking issue #711 (atomic/compensating multi-step state changes). Given the payload is an empty, harmless contact list and the identity is abandoned anyway, I wouldn't block this fix on it — just flagging so it isn't lost.

PR #1173, opened 2026-07-29
Upgrade SQLCipher and enforce exclusive Marmot roots

erskingardner:
## Summary

- upgrade `rusqlite` to 0.40.1 and `libsqlite3-sys` to 0.38.1, bundling SQLCipher 4.14.0 and SQLite 3.51.3
- add a test-time version regression guard covering SQLite's WAL-reset corruption fix
- add a private, nonblocking, cross-process lease on `.marmot-runtime.lock`
- create missing roots privately while preserving the mode of existing operator-managed roots
- retain exclusive ownership through all `MarmotApp`/runtime clones and release it automatically on final drop or process exit
- return typed `MarmotKitError::RuntimeBusy` contention through UniFFI so the foreground app can retry and an NSE can take a bounded fallback path
- apply exclusive-root construction to MarmotKit and the agent connector while leaving the CLI's separately coordinated daemon model unchanged
- keep the workspace release at 0.9.10, record the changes under `Unreleased`, and pin Rust 1.95.0 as required by the updated `libsqlite3-sys`

## Why

A staging device reported `backend failure: database disk image is malformed` while conversation-list/settings reads remained available but conversation opens failed. This PR addresses two independent risks without claiming either one is conclusively the source of that device's corruption:

1. the previously bundled SQLite predates the WAL-reset corruption fix now present in SQLite 3.51.3;
2. independently hydrated foreground-app and notification-extension runtimes must not write the same account root concurrently.

SQLite's own transaction locking remains in place. The root lease protects the larger stateful-runtime boundary around database handles, runtime state, and multi-step processing.

## Runtime ownership contract

- acquisition uses `flock(LOCK_EX | LOCK_NB)` on a stable owner-only lock file
- contention fails immediately as typed `RuntimeBusy`; it never waits inside an NSE
- a missing root is created privately; an existing externally managed root keeps its mode
- `shutdown()` stops work but does not release ownership while any database-capable runtime handle remains alive
- abrupt process termination is safe because the kernel closes the descriptor and releases the lease
- the lock file is persistent and must not be deleted or replaced

## Validation

- `PATH="$HOME/.cargo/bin:$PATH" just fast-ci`
- `cargo test -p fs-private`
- all 531 `marmot-app` unit tests, including the root-mode preservation regression
- `cargo test -p cgka-conformance-simulator --test canonical_scenarios --test vector_artifacts`
- focused process-exit, lease-lifetime, SQLCipher-version, and typed-UniFFI tests under pinned Rust 1.95.0
- `./crates/marmot-uniffi/xcframework.sh` for arm64 iOS device and simulator slices
- `ANDROID_ABIS="arm64-v8a" ./crates/marmot-uniffi/kotlin-bindings.sh`

Broader existing `wn-cli` and `marmot-app` integration runs still report failures on this checkout. Representative failures were reproduced from a clean detached worktree at base commit `b609ccd3` with the original rusqlite 0.32.1 stack, confirming they are pre-existing rather than introduced by this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ members = [
 
 [workspace.package]
 edition = "2024"
-version = "0.9.10"
+version = "0.9.11"
 license = "MIT"
 publish = false
 
@@ -100,7 +100,7 @@ nostr-relay-builder = "0.44"
 # reqwest/rustls stack, so the SDK's record-time histogram API is not needed.
 opentelemetry-proto = { version = "0.32", default-features = false, features = ["gen-tonic-messages", "metrics"] }
 prost = "0.14"
-rusqlite = { version = "0.32", features = ["bundled-sqlcipher-vendored-openssl", "trace"] }
+rusqlite = { version = "0.40.1", default-features = false, features = ["bundled-sqlcipher-vendored-openssl", "cache", "trace"] }
 tempfile = "3"
 
 # test
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ members = [
 
 [workspace.package]
 edition = "2024"
-version = "0.9.11"
+version = "0.9.10"
 license = "MIT"
 publish = false
 
@@ -100,7 +100,7 @@ nostr-relay-builder = "0.44"
 # reqwest/rustls stack, so the SDK's record-time histogram API is not needed.
 opentelemetry-proto = { version = "0.32", default-features = false, features = ["gen-tonic-messages", "metrics"] }
 prost = "0.14"
-rusqlite = { version = "0.40.1", default-features = false, features = ["bundled-sqlcipher-vendored-openssl", "cache", "trace"] }
+rusqlite = { version = "0.40.1", default-features = false, features = ["bundled-sqlcipher-vendored-openssl", "trace"] }
 tempfile = "3"
 
 # test
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -1123,6 +1123,12 @@ impl MarmotApp {
         }
     }
 
+    /// Constructor for tests and embeddings that coordinate root ownership
+    /// externally.
+    ///
+    /// Independently scheduled processes must use
+    /// [`Self::try_with_relays_and_account_home_and_config`].
+    #[doc(hidden)]
     pub fn with_relays_and_account_home(
         root: impl AsRef<Path>,
         relay_urls: Vec<String>,
@@ -1142,6 +1148,7 @@ impl MarmotApp {
     /// Independently scheduled processes sharing a root must use
     /// [`Self::try_with_relays_and_account_home_and_config`] so independently
     /// hydrated runtimes cannot write the same root concurrently.
+    #[doc(hidden)]
     pub fn with_relays_and_account_home_and_config(
         root: impl AsRef<Path>,
         relay_urls: Vec<String>,
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -90,12 +90,14 @@ mod projection;
 mod publisher_sequences;
 mod relay_plane;
 mod relay_telemetry_export;
+mod root_runtime_lease;
 mod runtime;
 mod sqlcipher;
 
 use external_signer::{AccountSigner, RegisteredExternalSigner};
 pub use external_signer::{EXTERNAL_SIGNER_REJECTED, ExternalAccountSigner};
 pub(crate) use groups::AppGroupImageInput;
+pub use root_runtime_lease::{MARMOT_ROOT_RUNTIME_LOCK_FILE, MarmotRootRuntimeLease};
 pub(crate) use runtime::blocking_app_task;
 pub use runtime::{
     AccountManager, AccountSetupRequest, AccountSetupResult, AgentStreamWatchOptions,
@@ -427,6 +429,10 @@ type AppRuntime = AccountDeviceRuntime<
 #[derive(Clone)]
 pub struct MarmotApp {
     root: PathBuf,
+    /// Present for exclusive-root entry points. Every clone shares this lease
+    /// so the root remains exclusively owned until all database-capable app and
+    /// runtime handles have been released.
+    _root_runtime_lease: Option<Arc<MarmotRootRuntimeLease>>,
     relay_urls: Vec<String>,
     account_home: AccountHome,
     relay_plane: MarmotRelayPlane,
@@ -1054,10 +1060,11 @@ impl MarmotApp {
     }
 
     /// Dev/test-only convenience constructor. **Not a production entry point**
-    /// and hidden from the public API docs: production opens through
-    /// `with_relays_and_account_home*`, which default the relay-safety gate to
-    /// production posture (loopback rejected). This helper backs the crate's own
-    /// tests, which drive in-process `MockRelay`s at loopback, so it opts the
+    /// and hidden from the public API docs: exclusive-root hosts open through
+    /// [`MarmotApp::try_with_relays_and_account_home_and_config`], which owns
+    /// the root exclusively and defaults the relay-safety gate to production
+    /// posture (loopback rejected). This helper backs the crate's own tests,
+    /// which drive in-process `MockRelay`s at loopback, so it opts the
     /// relay-safety gate into admitting loopback endpoints. It cannot be
     /// `#[cfg(test)]`-gated because the crate's integration tests
     /// (`crates/marmot-app/tests/*`) consume it through the public API. Callers
@@ -1095,6 +1102,7 @@ impl MarmotApp {
         Self {
             account_home: AccountHome::open(&root),
             root,
+            _root_runtime_lease: None,
             relay_urls,
             relay_plane,
             config,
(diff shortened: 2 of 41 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ members = [
 
 [workspace.package]
 edition = "2024"
```
erskingardner, line 28: Unsolicited version bump inside a feature PR. `workspace.package.version` goes from 0.9.10 to 0.9.11 here, and `crates/cli/CHANGELOG.md` gets a new `[0.9.11]` heading in the same commit set — but per this repo's own established cadence, feature PRs accumulate entries under `[Unreleased]` and the version bump happens in a separate `chore(release): prepare X.Y.Z` PR.

This exact mistake was already caught and reverted today in a sibling PR (d6ebab7c, "Preserve bounded blank-line source gaps in Markdown AST"): its follow-up commit message reads literally "Remove the unsolicited workspace and conformance cohort bump from this feature PR and record the user-facing binding change under Unreleased."

Please drop the version bump from `Cargo.toml`/`Cargo.lock` and fold the CHANGELOG entry back under `[Unreleased]` here too.

crates/storage-sqlite/src/connection.rs
```diff
@@ -694,6 +701,39 @@ mod tests {
         );
     }
 
+    #[test]
+    fn bundled_sqlcipher_contains_the_wal_reset_corruption_fix() {
```
erskingardner, line 705: Minor: the PR description says this adds "a runtime version assertion covering SQLite's WAL-reset corruption fix," but `bundled_sqlcipher_contains_the_wal_reset_corruption_fix` is a `#[cfg(test)]` unit test — it only runs under `cargo test`, not something exercised by the actual runtime/production binary. It's a fine regression guard against an accidental dependency downgrade, but it doesn't assert anything at runtime the way the PR description implies. Worth tightening the description (or, if a real startup assertion was intended, adding one in non-test code).

erskingardner (commented):
## Summary

The core mechanism — a per-root nonblocking `flock(LOCK_EX | LOCK_NB)` lease (`crates/fs-private/src/lib.rs`, `crates/marmot-app/src/root_runtime_lease.rs`) shared via `Arc` across every `MarmotApp`/`MarmotAppRuntime` clone and released only on final `Drop` or process exit — is sound. I traced the propagation: `MarmotAppRuntime` and the spawned account-worker tasks each hold their own `MarmotApp` clone (and therefore their own `Arc` to the lease), so the lease can't be released out from under a still-running background worker. The typed `RuntimeBusy` plumbing through `AppError` → `MarmotKitError`/`ConnectorError` is wired correctly, and the SQLCipher/rusqlite 0.40.1 upgrade's mechanical `usize`→`i64`/`i64_to_usize` and `trace`→`trace_v2` fixups look correct. Leaving the CLI daemon model untouched, as stated, matches what's actually in the diff.

**Requesting changes on two points** (GitHub won't let me set that review state on your own PR, so filing as comment):

1. **Unsolicited version bump inside a feature PR** (inline comment on `Cargo.toml`). `workspace.package.version` jumps 0.9.10 → 0.9.11 and `crates/cli/CHANGELOG.md` gets a new `[0.9.11]` heading, but this repo's convention (visible in its own recent history) is to accumulate entries under `[Unreleased]` and bump the version only in a dedicated `chore(release): prepare X.Y.Z` PR. This exact mistake was caught and reverted today in a sibling PR (d6ebab7c) with the commit message "Remove the unsolicited workspace and conformance cohort bump from this feature PR and record the user-facing binding change under Unreleased." Same fix needed here.

2. **Minor**: the PR description claims a "runtime version assertion" for the WAL-reset fix, but what's added is a `#[cfg(test)]` unit test (inline comment on `connection.rs`) — it's a good CI regression guard but doesn't assert anything in the production binary. Worth correcting the description, or adding an actual startup check if that was the intent.

Neither is a correctness problem with the lease/lock mechanism itself, but the version-bump convention violation should be undone before merge.

crates/marmot-app/src/root_runtime_lease.rs
```diff
@@ -0,0 +1,116 @@
...
+impl MarmotRootRuntimeLease {
+    /// Try to acquire exclusive ownership of `root` without waiting.
+    ///
+    /// [`AppError::RuntimeBusy`] means another process/runtime owns the root and
+    /// the caller should either retry later (foreground app) or take its
+    /// bounded fallback path (notification extension). Other failures retain
+    /// their I/O classification.
+    pub fn try_acquire(root: impl AsRef<Path>) -> Result<Self, AppError> {
+        let root = root.as_ref();
+        fs_private::create_dir_all_private(root)?;
```
erskingardner, line 49: **Blocking:** `create_dir_all_private` always fchmods the leaf to 0700, including when `root` already exists.

`AgentConnector` now opens through `try_with_*`, so this runs on every `wn-agent` start. That regresses the documented split-user layout that intentionally uses a group-traversable home (0750) so a sibling service user can reach a 0770 socket under that home (see `integrations/hermes/marmot/README.md`). Production `/var/lib/marmot-agent` is already 0700, so same-UID installs are fine — the damage is to operator-managed shared homes.

Lease acquisition should not rewrite root policy. Prefer:
1. create the root only if missing (private create is fine for the new case),
2. if it exists, validate it is a directory and leave its mode alone,
3. privately create/lock only `.marmot-runtime.lock`.

A regression test with an existing 0750 root asserting the mode is unchanged after `try_acquire` would lock this in.

Cargo.toml
```diff
@@ -100,7 +100,7 @@ nostr-relay-builder = "0.44"
 # reqwest/rustls stack, so the SDK's record-time histogram API is not needed.
 opentelemetry-proto = { version = "0.32", default-features = false, features = ["gen-tonic-messages", "metrics"] }
 prost = "0.14"
-rusqlite = { version = "0.32", features = ["bundled-sqlcipher-vendored-openssl", "trace"] }
+rusqlite = { version = "0.40.1", default-features = false, features = ["bundled-sqlcipher-vendored-openssl", "cache", "trace"] }
```
erskingardner: Nit: I do not see any `prepare_cached` / statement-cache usage in the workspace. If `cache` is not required for the 0.40 upgrade, drop it so the feature set stays minimal (`bundled-sqlcipher-vendored-openssl` + `trace` already match the old intent).

crates/marmot-app/src/lib.rs
```diff
@@ -1128,6 +1136,12 @@ impl MarmotApp {
         )
     }
 
+    /// Constructor for tests and embeddings that coordinate ownership
+    /// externally.
+    ///
+    /// Independently scheduled processes sharing a root must use
+    /// [`Self::try_with_relays_and_account_home_and_config`] so independently
+    /// hydrated runtimes cannot write the same root concurrently.
     pub fn with_relays_and_account_home_and_config(
```
erskingardner, line 1152: Non-blocking: this remains the convenient infallible constructor, while the safe exclusive path is the longer `try_with_*` name. Docs help, but future embeddings will still grab this one. Consider `#[doc(hidden)]`, a louder name (`without_root_lease` / `for_externally_coordinated_root`), or a deprecation nudge pointing at `try_with_*` for independently scheduled processes.

erskingardner (commented):
Additional review pass (GitHub will not accept request-changes on an author self-review, so filing as comment).

Agree with the earlier notes on the unsolicited 0.9.11 bump and the test-vs-runtime wording for the SQLCipher version check. Adding one stronger finding:

## Blocking (correctness / deploy regression)
`MarmotRootRuntimeLease::try_acquire` calls `create_dir_all_private(root)`, which fchmods an existing root to 0700. Agent connector now opens through `try_with_*`, so every `wn-agent` start rewrites operator-managed home modes. That regresses the documented split-user layout that uses a group-traversable home (0750) so a sibling service user can reach a 0770 socket under that home. Lease acquisition should create-if-missing / preserve existing root mode, and only privately create+lock `.marmot-runtime.lock`.

## Otherwise
Premise and lease design look solid: SQLCipher 4.14.0/SQLite 3.51.3 is the right WAL-reset upgrade, nonblocking flock + Arc lifetime + typed RuntimeBusy is the right app↔NSE boundary, and the process/abrupt-exit/UniFFI tests cover the important contracts. Once root-mode preservation and the release-bump hygiene are fixed, this looks mergeable.

Cargo.toml
erskingardner: Fixed in `a6157bf4`: the workspace and conformance cohort remain at 0.9.10, and these release notes are back under `Unreleased`. The dependency/toolchain upgrade remains part of the feature diff without claiming a release version.

crates/storage-sqlite/src/connection.rs
erskingardner: Fixed in the PR description: this is now described as a test-time version regression guard, not a production runtime assertion.

crates/marmot-app/src/root_runtime_lease.rs
erskingardner: Fixed in `a6157bf4`. Lease acquisition now uses `prepare_directory_path(..., ExistingDirectoryMode::Preserve)`: missing roots are created privately, existing directory modes are preserved, and only `.marmot-runtime.lock` is enforced at 0600. The new regression test starts with a 0750 root and asserts both the unchanged root mode and private lock-file mode.

Cargo.toml
erskingardner: Confirmed there is no statement-cache API usage and removed the `cache` feature in `a6157bf4`. The unused `hashlink` dependency also drops out of `Cargo.lock`.

crates/marmot-app/src/lib.rs
erskingardner: Addressed in `a6157bf4`: both externally coordinated, unleased account-home constructors are now `#[doc(hidden)]`; the public documented production path is the exclusive `try_with_relays_and_account_home_and_config` constructor.

jgmontoya (approved):

PR #1174, opened 2026-07-29
Scope convergence reservation to admin intents and schedule on real cutoffs

erskingardner:
## Summary

PR A of [the convergence settling remediation plan](https://github.com/marmot-protocol/mdk/blob/fix/convergence-scoped-reservation/docs/marmot-architecture/further-context/convergence-settling-remediation-plan.md) (included in this PR). Fixes the settling regression caused by #1110's completed-pass park interacting with #1158's routine outbound queuing.

## Root cause

The spec's one-attempt rule ([convergence.md](https://github.com/marmot-protocol/marmot/blob/master/protocol-core/convergence.md#L97-L103), adopted for marmot-protocol/marmot#375) reserves the completed-pass boundary for **one queued admin-authorized group-state intent**. The implementation parked admission whenever **any** outbound intent was queued — including ordinary app messages, which #1158 made a routine steady state. A parked pass retains inbound commits without opening a pass, so settling slipped multiple scheduler cycles. Independently, the app scheduler re-armed blind (full quiescence delay), counted in-window ticks toward the unsettled cap (demoting healthy busy groups to error backoff), and swallowed schedule-state errors as "no work" (cancelling wakeups).

## What changed

**Engine (`cgka-engine`)**
- `load_or_open_convergence_pass` parks the boundary only when a queued intent matches `is_admin_group_state_intent` (renamed from `is_admin_group_state_fairness_intent`; single chokepoint). Queued app messages never delay pass admission.
- Consuming the reservation schedules the group's next convergence drain immediately.
- New `has_queued_outbound_intents` accessor (engine → session → account) so schedulers keep a wakeup armed while queued sends exist.
- Settling telemetry in `engine_metrics`: pass-open→apply latency, generation gap, freeze-overdue histograms, admin-reservation outcome counters (aggregate only).

**App runtime (`marmot-app`)**
- `AppClient::convergence_schedule_state` returns a structured state (`Idle` / `Collecting { remaining_ms }` / `Ready` / `PendingUnopenable`) and **propagates errors**; the worker arms retries on failure instead of `note_success`.
- `schedule_after_pass` arms from the engine-reported remaining cutoff; in-window (`Collecting`) wakes never increment the unsettled re-arm counter — only `PendingUnopenable` counts toward backoff demotion.

**Unchanged**: the one-attempt admin reservation itself (kept per marmot#375 — it is the flooding-attack mitigation), `DurableConvergencePass` schema/serde (no migration), quiescence/absolute-deadline constants, the `Unrecoverable` integrity posture. No spec change needed.

## Tests

- Engine (7 new, in `tests/distributed_convergence.rs`): scoped parking with queued app messages, the one-attempt grant, failure consumption, marmot#375 flood starvation (absolute-deadline freeze + boundary attempt), restart durability of the reservation, restart with app-only queues, immediate rescheduling on consumption.
- End-to-end (`tests/relay_runtime.rs`): three rename generations through the real MockRelay worker loop with a member send fired inside each convergence window.
- Scheduler units: cutoff arming, no re-arm penalty for in-window ticks, deadline never-postpone invariants.
- The pre-existing #1110 reservation test passes unchanged (its queued intent is already admin-scoped).

## Validation

- `just fast-ci` clean.
- `cargo nextest`: cgka-engine 415 passed; marmot-app/storage-sqlite/cgka-session/marmot-account 1029 passed with 5 failures **identical on clean master** (pre-existing `relay_runtime` media/retention cases, verified by stash-and-rerun); cgka-conformance-simulator 130 passed.

## Follow-ups (per the plan doc)

- PR B: semantics-preserving drain/maintenance cleanup.
- PR C: frozen-snapshot completeness (design first).

## Summary by CodeRabbit

* **Bug Fixes**
  * Group state changes settle promptly even when application messages are queued.
  * Queued application messages no longer delay the next convergence pass; only queued admin-authorized group-state updates may briefly hold the admission boundary.
  * Convergence scheduling avoids unnecessary retry backoff while convergence is still collecting and handles scheduling states more reliably.

* **Monitoring**
  * Added diagnostics for pass latency, freeze overdue lag, generation gaps, and administrative reservation hold/attempt outcomes.

* **Documentation**
  * Updated changelog and architecture guidance describing the corrected convergence fairness behavior.

### DIFF
--- a/crates/cgka-engine/src/engine_metrics.rs
+++ b/crates/cgka-engine/src/engine_metrics.rs
@@ -193,6 +193,27 @@ pub struct EngineMetrics {
     reorg_lateness_ms: BucketHistogram,
     /// Per-group last-applied branch, in-memory only (never snapshotted).
     last_applied: HashMap<GroupId, LastAppliedBranch>,
+    /// Pass open → apply, in milliseconds. The direct settling-latency signal
+    /// for the convergence remediation plan: healthy passes sit near
+    /// `settlement_quiescence_ms` plus scheduler margin.
+    pass_apply_latency_ms: BucketHistogram,
+    /// Previous generation completed → this generation opened, in
+    /// milliseconds. Near-zero gaps mean retained input flows into the next
+    /// pass without losing scheduler cycles at the boundary.
+    generation_gap_ms: BucketHistogram,
+    /// How far past its due cutoff a pass was frozen, in milliseconds —
+    /// scheduler lag between a cutoff elapsing and the freeze running.
+    freeze_overdue_ms: BucketHistogram,
+    /// Times the completed-pass boundary was held for a queued
+    /// admin-authorized group-state intent while retained inbound waited.
+    admin_reservation_holds: u64,
+    /// One-attempt reservations that prepared their admin intent.
+    admin_reservation_prepared: u64,
+    /// One-attempt reservations consumed by a failed preparation.
+    admin_reservation_failed: u64,
+    /// Per-group completion time of the last applied pass, in-memory only
+    /// (never snapshotted); feeds `generation_gap_ms`.
+    last_pass_completed_at_ms: HashMap<GroupId, u64>,
 }
 
 impl Default for EngineMetrics {
@@ -203,6 +224,13 @@ impl Default for EngineMetrics {
             reorg_rewind_depth: BucketHistogram::new(&REWIND_DEPTH_BUCKET_BOUNDS),
             reorg_lateness_ms: BucketHistogram::new(&LATENESS_BUCKET_BOUNDS_MS),
             last_applied: HashMap::new(),
+            pass_apply_latency_ms: BucketHistogram::new(&LATENESS_BUCKET_BOUNDS_MS),
+            generation_gap_ms: BucketHistogram::new(&LATENESS_BUCKET_BOUNDS_MS),
+            freeze_overdue_ms: BucketHistogram::new(&LATENESS_BUCKET_BOUNDS_MS),
+            admin_reservation_holds: 0,
+            admin_reservation_prepared: 0,
+            admin_reservation_failed: 0,
+            last_pass_completed_at_ms: HashMap::new(),
         }
     }
 }
@@ -268,6 +296,42 @@ impl EngineMetrics {
         );
     }
 
+    /// Record a convergence pass that applied (or completed with nothing to
+    /// apply after freezing): open → apply latency plus the gap since the
+    /// group's previous completed generation. `opened_monotonic_ms` is the
+    /// pass's (possibly restart-rebased) open time; `now_ms` is the apply
+    /// time. Both local-monotonic.
+    pub fn note_pass_applied(&mut self, group_id: &GroupId, opened_monotonic_ms: u64, now_ms: u64) {
+        self.pass_apply_latency_ms
+            .record(now_ms.saturating_sub(opened_monotonic_ms));
+        if let Some(previous_completed) = self.last_pass_completed_at_ms.get(group_id) {
+            self.generation_gap_ms
+                .record(opened_monotonic_ms.saturating_sub(*previous_completed));
+        }
+        self.last_pass_completed_at_ms
+            .insert(group_id.clone(), now_ms);
+    }
+
+    /// Record how far past its due cutoff a pass was frozen (scheduler lag).
+    pub fn note_pass_frozen_overdue(&mut self, overdue_ms: u64) {
+        self.freeze_overdue_ms.record(overdue_ms);
+    }
+
+    /// Record that the completed-pass boundary is being held for a queued
+    /// admin-authorized group-state intent while retained inbound waits.
+    pub fn note_admin_reservation_hold(&mut self) {
+        self.admin_reservation_holds += 1;
+    }
+
+    /// Record the outcome of the one-attempt admin reservation.
+    pub fn note_admin_reservation_attempt(&mut self, prepared: bool) {
+        if prepared {
+            self.admin_reservation_prepared += 1;
+        } else {
+            self.admin_reservation_failed += 1;
+        }
+    }
+
     /// Aggregate, privacy-safe snapshot for diagnostics and quiescence tuning.
     pub fn snapshot(&self) -> EngineMetricsSnapshot {
         EngineMetricsSnapshot {
@@ -275,6 +339,12 @@ impl EngineMetrics {
             post_settle_reorgs: self.post_settle_reorgs,
             reorg_rewind_depth: self.reorg_rewind_depth.snapshot(),
             reorg_lateness_ms: self.reorg_lateness_ms.snapshot(),
+            pass_apply_latency_ms: self.pass_apply_latency_ms.snapshot(),
+            generation_gap_ms: self.generation_gap_ms.snapshot(),
+            freeze_overdue_ms: self.freeze_overdue_ms.snapshot(),
+            admin_reservation_holds: self.admin_reservation_holds,
+            admin_reservation_prepared: self.admin_reservation_prepared,
+            admin_reservation_failed: self.admin_reservation_failed,
         }
     }
 }
@@ -293,6 +363,19 @@ pub struct EngineMetricsSnapshot {
     pub reorg_rewind_depth: HistogramSnapshot,
     /// Lateness per reorg, in local-monotonic milliseconds.
     pub reorg_lateness_ms: HistogramSnapshot,
+    /// Pass open → apply latency, in local-monotonic milliseconds.
+    pub pass_apply_latency_ms: HistogramSnapshot,
+    /// Previous generation completed → next generation opened, in
+    /// local-monotonic milliseconds.
+    pub generation_gap_ms: HistogramSnapshot,
+    /// Freeze lag past the due cutoff, in local-monotonic milliseconds.
+    pub freeze_overdue_ms: HistogramSnapshot,
+    /// Completed-pass boundary holds for a queued admin group-state intent.
+    pub admin_reservation_holds: u64,
+    /// One-attempt admin reservations that prepared their intent.
+    pub admin_reservation_prepared: u64,
+    /// One-attempt admin reservations consumed by a failed preparation.
+    pub admin_reservation_failed: u64,
 }
 
 impl EngineMetricsSnapshot {
--- a/crates/cgka-engine/src/engine_metrics.rs
+++ b/crates/cgka-engine/src/engine_metrics.rs
@@ -204,9 +204,13 @@ pub struct EngineMetrics {
     /// How far past its due cutoff a pass was frozen, in milliseconds —
     /// scheduler lag between a cutoff elapsing and the freeze running.
     freeze_overdue_ms: BucketHistogram,
-    /// Times the completed-pass boundary was held for a queued
+    /// Drain-loop observations of a boundary held for a queued
     /// admin-authorized group-state intent while retained inbound waited.
-    admin_reservation_holds: u64,
+    /// Counted once per drain pass, not once per park entry: repeated wakes
+    /// during a single park each count, so compare against
+    /// `admin_reservation_prepared`/`failed` as an observation rate, not a
+    /// park count.
+    admin_reservation_hold_observations: u64,
     /// One-attempt reservations that prepared their admin intent.
     admin_reservation_prepared: u64,
     /// One-attempt reservations consumed by a failed preparation.
@@ -227,7 +231,7 @@ impl Default for EngineMetrics {
             pass_apply_latency_ms: BucketHistogram::new(&LATENESS_BUCKET_BOUNDS_MS),
             generation_gap_ms: BucketHistogram::new(&LATENESS_BUCKET_BOUNDS_MS),
             freeze_overdue_ms: BucketHistogram::new(&LATENESS_BUCKET_BOUNDS_MS),
-            admin_reservation_holds: 0,
+            admin_reservation_hold_observations: 0,
             admin_reservation_prepared: 0,
             admin_reservation_failed: 0,
             last_pass_completed_at_ms: HashMap::new(),
@@ -320,7 +324,7 @@ impl EngineMetrics {
     /// Record that the completed-pass boundary is being held for a queued
     /// admin-authorized group-state intent while retained inbound waits.
     pub fn note_admin_reservation_hold(&mut self) {
-        self.admin_reservation_holds += 1;
+        self.admin_reservation_hold_observations += 1;
     }
 
     /// Record the outcome of the one-attempt admin reservation.
@@ -342,7 +346,7 @@ impl EngineMetrics {
             pass_apply_latency_ms: self.pass_apply_latency_ms.snapshot(),
             generation_gap_ms: self.generation_gap_ms.snapshot(),
             freeze_overdue_ms: self.freeze_overdue_ms.snapshot(),
-            admin_reservation_holds: self.admin_reservation_holds,
+            admin_reservation_hold_observations: self.admin_reservation_hold_observations,
             admin_reservation_prepared: self.admin_reservation_prepared,
             admin_reservation_failed: self.admin_reservation_failed,
         }
@@ -370,8 +374,9 @@ pub struct EngineMetricsSnapshot {
     pub generation_gap_ms: HistogramSnapshot,
     /// Freeze lag past the due cutoff, in local-monotonic milliseconds.
     pub freeze_overdue_ms: HistogramSnapshot,
-    /// Completed-pass boundary holds for a queued admin group-state intent.
-    pub admin_reservation_holds: u64,
+    /// Drain-loop observations (not park entries) of a boundary held for a
+    /// queued admin group-state intent.
+    pub admin_reservation_hold_observations: u64,
     /// One-attempt admin reservations that prepared their intent.
     pub admin_reservation_prepared: u64,
     /// One-attempt admin reservations consumed by a failed preparation.
--- a/crates/cgka-engine/src/engine_metrics.rs
+++ b/crates/cgka-engine/src/engine_metrics.rs
@@ -336,6 +336,15 @@ impl EngineMetrics {
         }
     }
 
+    /// Drop the per-group in-memory records for a group that reached a
+    /// terminal state (disband/removal). Both maps grow one entry per live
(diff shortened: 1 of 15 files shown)


erskingardner:
@coderabbitai full review

crates/marmot-app/src/client/sync.rs
```diff
@@ -21,25 +21,58 @@ use crate::{
...
+            Some(remaining_ms) => Ok(ConvergenceScheduleState::Collecting { remaining_ms }),
+            None => {
+                if self.runtime.has_pending_convergence_inputs(group_id)?
+                    || self.runtime.has_queued_outbound_intents(group_id)?
+                {
+                    // Queued outbound intents keep the wakeup armed even with
+                    // no convergence work: the scheduled drain is what
+                    // regenerates and publishes them, and a failed sync on
+                    // that tick is what triggers transport reactivation.
+                    Ok(ConvergenceScheduleState::PendingUnopenable)
```
erskingardner, line 70: Non-blocking: folding `has_queued_outbound_intents` into `PendingUnopenable` keeps wakeups armed (good vs the old `note_success` cancel), but `schedule_after_pass` then routes through `schedule_unsettled_groups`, which increments `unsettled_rearm_attempts` and demotes to error backoff after `CONVERGENCE_UNSETTLED_MAX_REARMS`.

Transport publish failures take the `Err` → `schedule_retry_groups` path, so the common #1158 inactive-transport case is fine. The risk is Ok-but-still-queued early returns (leave gate, SelfRemove staging, etc.): ~11s of those demotes a healthy queue to up-to-60s backoff.

Consider a distinct schedule arm (or a `PendingOutbound` state) that uses `normal_delay()` without touching the unsettled re-arm counter — reserve that counter for true unopenable convergence.

crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -6666,3 +6666,84 @@ async fn concurrent_leaves_report_already_requested_not_an_opaque_error() {
...
+
+/// Convergence remediation-plan liveness guard: successive inbound commits,
+/// each with a member send fired while the commit is still converging, must
+/// keep settling promptly through the real worker scheduling path.
+///
+/// Under the pre-fix behavior a queued app message parked the completed-pass
+/// boundary, so the next rename slipped extra scheduler cycles (and, after
+/// enough unsettled re-arms, into error backoff); the 5-second deadlines
+/// inside `wait_for_group_state_update` / `wait_for_event` fail in that
+/// world. A send that occasionally lands after bob already settled simply
```
erskingardner: Non-blocking / test strength: this comment admits the mid-window queued path is best-effort ("occasionally"). The 5s deadlines catch parking/backoff disasters, but a green run does not prove the regression path was exercised.

Engine tests 1–7 already own the hard assertions — worth saying that more loudly here, or adding one deterministic seam (e.g. assert bob's pass generation/phase around the send) if the harness can expose it without becoming flaky.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -365,6 +365,15 @@ impl<S: StorageProvider> Engine<S> {
...
             fairness_slot_available = false;
         }
+        if fairness_slot_available
+            && queued
+                .iter()
+                .any(|record| is_admin_group_state_intent(&record.intent))
+        {
+            // Diagnostic only: the completed-pass boundary is being held for a
+            // queued admin group-state intent while retained inbound waits.
+            self.engine_metrics.note_admin_reservation_hold();
```
erskingardner, line 375: Nit: `note_admin_reservation_hold` increments on every drain while the boundary is held, not once per hold entry. Retries/wakes during a single park will inflate `admin_reservation_holds` relative to `admin_reservation_prepared`/`failed`. Fine if the intent is "hold observations"; if you want entry-rate for the remediation dashboard, gate on a transition (e.g. first drain after park) or rename the snapshot field.

erskingardner (commented):
Recommend merge. The root cause diagnosis matches the spec (convergence.md one-attempt rule is admin-scoped) and the #1110×#1158 interaction; scoping the park in `load_or_open_convergence_pass`, immediate reschedule on reservation consume, and honest `ConvergenceScheduleState` scheduling are the right fix. Engine tests pin the behavioral contract well; scheduler unit tests cover the re-arm/cutoff changes.

Non-blocking notes inline. No blocking correctness issues found.

crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -6666,3 +6666,84 @@ async fn concurrent_leaves_report_already_requested_not_an_opaque_error() {
...
+        let renamed = format!("settling-round-{round}");
+        runtime
+            .update_group_profile(&alice_id, &group_id, Some(renamed.clone()), None)
+            .await
+            .unwrap();
+        // Fire bob's send immediately so it often lands inside bob's
+        // collection window for the rename commit (the queued-intent path).
+        let text = format!("bob mid-window {round}");
+        runtime
+            .send_message(&bob_id, &group_id, text.clone().into_bytes())
```
erskingardner, line 6739: [P1] Make this test prove the queued path. This send races Bob relay ingest, so all three rounds may publish directly; the test explicitly allows that. Even when one queues, the two separate 5 s waits still admit the pre-fix roughly 3.3 s rename / 4.4 s send behavior. As written, the old implementation can pass this test. Add a deterministic barrier or assertion that Bob is Collecting and this send was queued, then assert the one-cutoff latency bound.

crates/cli/CHANGELOG.md
```diff
@@ -9,6 +9,14 @@ versioning through the workspace version in the root `Cargo.toml`.
...
 ## [Unreleased]
 
+### Fixed
+
+- Group state changes settle promptly again when messages are queued for
+  sending: a queued ordinary message no longer delays the next convergence
+  pass (only an admin group-state change may briefly hold that boundary, as
+  the protocol requires), and busy groups are no longer demoted to retry
+  backoff while convergence is legitimately still collecting.
+
```
erskingardner, line 19: [P3] Remove the whitespace-only line. `git diff --check` currently fails here with `trailing whitespace`, so the patch does not meet a clean-diff gate.

erskingardner (commented):
Request changes before merge. The engine scoping and structured scheduler state match the adopted Marmot rule, but the only worker-level regression test does not force or observe the queued mid-window path and can pass with the pre-fix latency. Please make that path deterministic and assert the intended latency class. Also clean the changelog whitespace so `git diff --check` passes. CI was still running at review time.

erskingardner (commented):
Reviewed against the linked context (#1110, #1158, marmot#375) and the bundled remediation-plan doc. This is a well-scoped, well-tested fix. (Posted as a comment, not an approval — GitHub does not allow self-approval of your own PR.)

**What it does:** correctly narrows the completed-pass fairness reservation to admin-authorized group-state intents only (via the single `is_admin_group_state_intent` chokepoint), so queued app messages no longer park pass admission. It also fixes the independent scheduler defect (blind re-arm delay, in-window ticks wrongly counted toward the unsettled-backoff cap, and swallowed schedule-state errors reading as "no work").

**Verified:**
- `load_or_open_convergence_pass` parking condition now correctly checks `queued.iter().any(is_admin_group_state_intent)` instead of `!queued.is_empty()`.
- The drain loop's admin-intent skip (`continue` when `fairness_slot_available && !is_admin_group_state_intent`) correctly lets non-admin records be skipped in favor of scanning for an admin one, preserving one-attempt semantics.
- `consume_convergence_fairness_slot` now reschedules the group immediately (closes the generation-handoff gap) — confirmed by `reservation_consumption_schedules_next_generation_immediately`.
- New `ConvergenceScheduleState` correctly propagates errors instead of the old `unwrap_or(false)`/`.ok().flatten()` swallowing, and `schedule_after_pass` only counts `PendingUnopenable` toward the re-arm cap — matches the stated liveness-bug fix.
- New metrics are aggregate-only (latency histograms, counters), no group/account/message identifiers — consistent with the repo's observability invariant.
- Test coverage is thorough: 7 new engine tests cover the scoped-park regression, the one-attempt grant/failure paths, marmot#375 flood-starvation non-regression, and restart durability; plus a real end-to-end MockRelay test and scheduler unit tests.

**Minor, non-blocking:**
- `crates/cli/CHANGELOG.md` line 19 has a trailing-whitespace-only blank line between the new "Fixed" entry and "### Added" — cosmetic only.

No correctness or scope concerns. Recommend merge.

crates/marmot-app/tests/relay_runtime.rs
erskingardner: Addressed in 3f583d95, with one honest scope note.

**Made deterministic:** `SendSummary.published == 0` is now the witness that a send took the durable queued path (a queued acceptance publishes nothing on transport). The loop repeats rounds until one queued acceptance is observed (fails after 8 attempts if never hit — in practice round 0-1), and for that round asserts the message reaches alice within **2s** = one settlement cycle (1000ms quiescence + 100ms margin) plus drain slack. The pre-fix drain deferred a queued app intent past the apply tick whenever retained inbound was present, costing ≥1 extra full settlement cycle (≥~2.2s), so the bound separates the latency classes. Verified stable across 3 consecutive runs (~3.5s each).

**Scope note:** the *parking* interleaving itself — a new inbound commit arriving while the queue is non-empty AND the reservation is set — cannot be forced through public runtime APIs on a healthy in-proc relay: pass-apply and queue-drain happen inside a single worker tick, so the arrival window is sub-millisecond. That contract is pinned deterministically at the engine layer instead, where `pass_opens_while_app_message_intents_are_queued` asserts a generation-1 Collecting pass exists immediately with an app intent queued — an assertion the pre-fix engine fails outright (it returns the parked generation-0 Completed pass). The test's doc comment now states this division of labor explicitly.

crates/cli/CHANGELOG.md
erskingardner: Could not reproduce: `git diff --check origin/master...HEAD` exits 0 on the branch and `grep -nE ' +$' crates/cli/CHANGELOG.md` finds no trailing whitespace (the file was written in one commit, 63b06bf9, and `just fmt-check` + the Rust format CI job pass). If you're still seeing it on your side, paste the `git diff --check` output and I'll chase it — but I believe this one is stale.

crates/marmot-app/src/client/sync.rs
erskingardner: Taken, in 3f583d95: added `ConvergenceScheduleState::PendingOutbound` for the queued-intents-without-convergence-work case. It re-arms on `normal_delay()` and never touches `unsettled_rearm_attempts`, so a healthy waiting queue can no longer be demoted to error backoff; the counter is now reserved for true unopenable convergence (`PendingUnopenable`). Transport failures still reach backoff through the sync/drain `Err` paths. Unit test: `pending_outbound_rearms_without_counting_toward_backoff` (drives it past `CONVERGENCE_UNSETTLED_MAX_REARMS` iterations and asserts both counters stay empty).

crates/cgka-engine/src/message_processor/mod.rs
erskingardner: Renamed in 3f583d95 to `admin_reservation_hold_observations` (field + snapshot), with the per-drain-pass counting semantics stated in the doc comments on both the recorder and the snapshot — i.e. read it as an observation rate against `prepared`/`failed`, not a park-entry count. Gating on the park transition would need mutable metrics inside `load_or_open_convergence_pass` (which is `&self`), so the rename was the cheaper honest fix.

crates/marmot-app/tests/relay_runtime.rs
erskingardner: Superseded by the fix for the [P1] thread below (3f583d95): the queued path now has a deterministic witness (`published == 0`) with a retry-until-observed loop and a one-cycle latency bound, and the doc comment now says explicitly that the parking discriminator lives in the engine tests.

crates/marmot-app/tests/relay_runtime.rs
erskingardner: Update after CI data (a1abba77): my previous fix made the queued witness *mandatory* with retries — CI then failed 8/8 rounds (`no round exercised the queued mid-window send path`), and instrumented local runs with a 300ms head-start failed 12/12 the other way. Measured behavior: a healthy in-proc relay settles a linear rename in well under one quiescence window, so the mid-window interval is a **sub-300ms race that machines win or lose systematically**. No fixed delay makes it deterministic through public APIs — that's the fixed system settling fast, not the test being lazy.

Final shape: three fixed liveness rounds through the real worker scheduler; whenever a round's send does report the queued acceptance (`published == 0`), the hard one-cycle latency bound (2.5s) applies. The deterministic queued-path and parking contracts remain in the engine suite (`pass_opens_while_app_message_intents_are_queued` + reservation tests), which fail outright against the pre-fix engine. If mandatory worker-level determinism is wanted, the missing piece is a test-only transport-pause or pass-phase diagnostics seam — noted as a PR-B candidate rather than invented ad hoc here.

One genuine find from the instrumentation: bob's chat projection reflects an applied rename while his `subscribe_group_state` subscription can stay silent when a send interleaves. The test now polls the projection; the subscription question is spun out for separate investigation (it may pre-date this PR).

docs/marmot-architecture/further-context/convergence-settling-remediation-plan.md
```diff
@@ -0,0 +1,375 @@
...
+4. Define the verified-repair path for a manifest mismatch (aligned with the existing
+   quarantine-repair flow), keeping `Unrecoverable` as the interim response.
+
+Not scheduled until PR A has shipped and its telemetry confirms settling is fixed.
+
+---
+
+## 6. Sequencing, rollout, revert levers
+
+- PR A ships first and alone; it is the user-facing fix. PR B after, trivially revertable.
```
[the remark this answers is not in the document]
erskingardner: Both taken in the latest push: typo fixed, and the flood test now derives its cadence (`V1_SETTLEMENT_QUIESCENCE_MS * 4 / 5`) and cutoff offsets (`V1_MAX_CONVERGENCE_PASS_MS + …`) from the pinned v1 constants.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1881,11 +1895,46 @@ impl ScheduledConvergence {
...
+            ConvergenceScheduleState::PendingOutbound => {
+                // A waiting outbound queue keeps the wakeup armed on the
+                // normal delay but is not unsettled convergence: it never
+                // feeds the re-arm cap, so a healthy queue cannot be demoted
+                // to error backoff (transport failures reach backoff through
+                // the sync/drain error paths instead).
+                self.retry_attempts.remove(group_id);
+                self.arm_no_later(group_id.clone(), TokioInstant::now() + self.normal_delay());
+                self.reset_timer_to_earliest();
+            }
```
jgmontoya, line 1940: From my agent:

> Every other non-backoff-inducing branch (`Idle`, `Collecting`, `Ready`) clears both
> `retry_attempts` *and* `unsettled_rearm_attempts`. `PendingOutbound` only clears
> `retry_attempts`. Since this state is explicitly documented as "healthy, never counts toward the
> re-arm cap," it reads like an oversight rather than a deliberate choice — a group that
> alternates between `PendingUnopenable` (accruing the counter) and `PendingOutbound` (a healthy
> send queue) never gets credit for the healthy ticks, so it can hit `CONVERGENCE_UNSETTLED_MAX_REARMS`
> sooner than a group that spends the same wall-clock time only in `PendingUnopenable`. It's not
> harmful (worst case: slightly earlier demotion to error backoff for a group that's mostly
> healthy), but it's inconsistent with the stated intent and with the sibling branches. Consider
> clearing `unsettled_rearm_attempts` here too, or leave a one-line comment explaining why it's
> intentionally left alone if it is.

crates/cgka-engine/src/engine_metrics.rs
```diff
@@ -193,6 +193,31 @@ pub struct EngineMetrics {
...
+    /// `admin_reservation_prepared`/`failed` as an observation rate, not a
+    /// park count.
+    admin_reservation_hold_observations: u64,
+    /// One-attempt reservations that prepared their admin intent.
+    admin_reservation_prepared: u64,
+    /// One-attempt reservations consumed by a failed preparation.
+    admin_reservation_failed: u64,
+    /// Per-group completion time of the last applied pass, in-memory only
+    /// (never snapshotted); feeds `generation_gap_ms`.
+    last_pass_completed_at_ms: HashMap<GroupId, u64>,
```
jgmontoya, line 220: From my agent:

> This in-memory map is inserted into on every applied pass (`note_pass_applied`) and never
> pruned — e.g. on group disband/removal. It follows the exact same shape as the pre-existing
> `last_applied: HashMap<GroupId, LastAppliedBranch>` (also never pruned), so this isn't a
> regression this PR introduces, just doubling down on an existing unbounded-growth pattern for a
> long-lived process with many groups over its lifetime. Not blocking given the precedent, but if
> this is ever revisited, both maps could use the same cleanup hook (wherever a group's
> disband/removal already cleans up other per-group in-memory engine state, e.g.
> `pending_convergence_groups.remove` in `disband.rs:612`).
>

jgmontoya (dismissed):
Some comments from my agent, I think none of them are blocking though

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Agreed — it was an oversight, not a choice. Fixed in 25a9fccd: `PendingOutbound` now clears `unsettled_rearm_attempts` like the other healthy states, with the rationale in a comment (the state means pending inputs are gone, so any unopenable streak genuinely ended). The unit test now also drives alternating `PendingUnopenable`/`PendingOutbound` ticks past the cap and asserts neither counter accrues.

crates/cgka-engine/src/engine_metrics.rs
erskingardner: Took the suggestion rather than doubling down: 25a9fccd adds `EngineMetrics::forget_group` clearing **both** per-group maps (`last_applied` and `last_pass_completed_at_ms`) and calls it from the terminal disband cleanup block you pointed at (alongside `pending_convergence_groups.remove`). Other group-removal paths (self-eviction realization, quarantine) keep their entries — those groups can return, and a stale record only slightly under-counts one reorg after re-join, matching the documented restart behavior.

PR #1182, opened 2026-07-29
fix(cgka-engine): reopen a stale convergence pass at the current tip instead of halting the group

jgmontoya:
A durable convergence pass records local scheduling state for one base epoch. On a real field device, a pass whose base epoch disagreed with the live tip was treated as a pass-integrity failure and marked the group durably Unrecoverable — whose only exit is a verified re-join — so a device six epochs behind was wedged at epoch 7 and every later send failed against the halt. The mismatch does not come from a live transition (the tip cannot move while a pass is open: every active pass gates outbound, and inbound commits route through convergence). It comes from dual authority: the pass's base_epoch was stamped from the durable group record while the halt compared the epoch manager's live epoch. Those two stores can split — behind the tip via the record lagging across a restart, or ahead of it via a fork rollback that moves the in-memory tip backwards without writing the record — so an inherited pass can sit on either side of the tip, benignly.

Give passes a single epoch authority instead (convergence_tip_epoch: the epoch manager, with the durable record only as the pre-hydration fallback), and maintain the invariant where passes are handed out: load_or_open_convergence_pass discards an active pass whose base epoch differs from the tip — behind or ahead, any active phase — and reopens at the tip, the same rule do_join_welcome already applies to a discarded local copy. A pass carries no canonical state of its own (members are id references into retained rows), so discarding drops a batch boundary and two deadlines and the members reseed. A frozen member that no longer matches the record it was admitted from is still a real integrity failure and still halts.

The compensation is diagnosable from a forensic export: a non-terminal convergence_pass_discarded audit row (named for the discard it records — the reopen is a separate decision, observable as the next generation's pass) carrying the stale base epoch, the current tip, and the discarded generation.

Review round: the mechanism prose was corrected at five sites (this description included — the original text claimed the catching-up transition), the audit kind was renamed from convergence_pass_reopened, and the discard is now pinned in both directions and the frozen phase by mutation-verified regressions, plus a unit test pinning that every pass trigger admits a commit edge that gates outbound work.

Stacked on this PR: #1184 makes the inbound-commit epoch mirror transactional instead of best-effort, closing the production source of the group-record/epoch-manager split that stamped stale bases in the first place — this PR repairs the records such splits leave behind. Rebased on master with #1173 in; will rebase again once #1174 lands, per review agreement.

## Summary of changes

* **Improvements**
  * Hardened stored-message convergence: stale durable convergence passes are discarded and re-opened at the current tip epoch, improving repair behavior for behind/ahead base-epoch mismatches.
  * Removed incorrect unrecoverable marking for certain base-epoch mismatches; normal convergence flow now proceeds.
  * Frozen-pass integrity failure handling is now more consistent, with tighter repair/rejection behavior.

* **Audit & Documentation**
  * Added `convergence_pass_discarded` audit event (stale base epoch, current tip epoch, generation), including schema and documentation updates.

* **Tests**
  * Added integration and unit tests covering pass reopening/compensation and commit-edge gating.

### DIFF
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -473,15 +473,24 @@ impl<S: StorageProvider> Engine<S> {
         }
     }
 
-    /// Discard a pass whose base epoch the device has already left behind, so the
-    /// caller reopens one at the current tip.
+    /// Discard a pass whose base epoch disagrees with the current tip, so the
+    /// caller reopens one at the tip.
     ///
-    /// A pass is local scheduling state for a single base epoch; the tip
-    /// legitimately moves while one is open (a device catching up applies
-    /// commits). The stale batch is therefore benign, not evidence of corruption,
-    /// and it carries no canonical state of its own — its members are retained
-    /// stored messages that reseed into the next pass. Same rule as the discarded
-    /// local copy in `do_join_welcome`.
+    /// This is a repair path, not a steady-state transition. The tip does not
+    /// move forward underneath an open pass: an active pass gates outbound work
+    /// (every trigger that opens one admits at least one `CommitEdge` member, and
+    /// a `CommitEdge` member gates unconditionally), and an inbound commit inside
+    /// the rewind horizon re-enters convergence instead of applying directly.
+    /// What this repairs is inherited scheduling state — before
+    /// [`Self::convergence_tip_epoch`] there were two authorities for one epoch,
+    /// because `base_epoch` was stamped from the durable group record while the
+    /// halt compared the epoch manager, and those two stores can split across a
+    /// restart. A base inherited from an older binary can therefore sit behind or
+    /// ahead of the tip, so both directions are discarded.
+    ///
+    /// The stale batch is benign either way: a pass carries no canonical state of
+    /// its own, and its members are retained stored messages that reseed into the
+    /// next pass. Same rule as the discarded local copy in `do_join_welcome`.
     fn discard_stale_convergence_pass(
         &self,
         pass: &DurableConvergencePass,
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -118,8 +118,9 @@ impl<S: StorageProvider> Engine<S> {
     /// remaining process-local cutoff delay.
     ///
     /// This is intentionally a command, not a diagnostic query: when eligible
-    /// retained input exists it may open a pass, consume a dormant fairness
-    /// slot, or persist restart deadline rebasing before returning the delay.
+    /// retained input exists it may open a pass, discard one left behind by an
+    /// advanced tip, consume a dormant fairness slot, or persist restart deadline
+    /// rebasing before returning the delay.
     pub fn prepare_convergence_cutoff_delay_ms(
         &mut self,
         group_id: &GroupId,
@@ -127,13 +128,10 @@ impl<S: StorageProvider> Engine<S> {
         if self.ensure_group_live(group_id).is_err() {
             return Ok(None);
         }
-        let group = self
-            .storage
-            .get_group(group_id)
-            .map_err(storage_projection_error)?;
+        let tip = self.convergence_tip_epoch(group_id)?;
         let policy = self.convergence_policy_for_group(group_id)?;
         let now_ms = self.convergence_now_ms();
-        let pass = self.load_or_open_convergence_pass(group_id, group.epoch, &policy, now_ms)?;
+        let pass = self.load_or_open_convergence_pass(group_id, tip, &policy, now_ms)?;
         Ok(match pass {
             Some(pass) if pass.phase == ConvergencePassPhase::Collecting => {
                 Some(pass.cutoff_monotonic_ms().saturating_sub(now_ms))
@@ -352,11 +350,7 @@ impl<S: StorageProvider> Engine<S> {
                 .get_group(group_id)
                 .map_err(storage_projection_error)?
                 .epoch;
-            self.realize_group_unrecoverable_for_convergence_pass(
-                group_id,
-                epoch,
-                "frozen_member_integrity",
-            );
+            self.realize_group_unrecoverable_for_frozen_pass(group_id, epoch);
         }
         Ok(())
     }
@@ -375,13 +369,9 @@ impl<S: StorageProvider> Engine<S> {
             // first stable scheduler run opens a pass and seeds retained work.
             return Ok(ConvergenceAdmissionOutcome::Retained);
         }
-        let group = self
-            .storage
-            .get_group(group_id)
-            .map_err(storage_projection_error)?;
+        let tip = self.convergence_tip_epoch(group_id)?;
         let policy = self.convergence_policy_for_group(group_id)?;
-        let Some(mut pass) =
-            self.load_or_open_convergence_pass(group_id, group.epoch, &policy, now_ms)?
+        let Some(mut pass) = self.load_or_open_convergence_pass(group_id, tip, &policy, now_ms)?
         else {
             return Ok(ConvergenceAdmissionOutcome::Retained);
         };
@@ -466,6 +456,51 @@ impl<S: StorageProvider> Engine<S> {
         Ok(())
     }
 
+    /// The epoch a convergence pass is scheduled against.
+    ///
+    /// One authority for the tip: the epoch manager owns it, and the durable
+    /// group record is only the pre-hydration fallback. Reading the two stores at
+    /// different call sites let a pass be stamped against an epoch the rest of
+    /// convergence disagreed with.
+    fn convergence_tip_epoch(&self, group_id: &GroupId) -> Result<EpochId, OpenMlsProjectionError> {
+        match self.epoch_manager.epoch(group_id) {
+            Some(epoch) => Ok(epoch),
+            None => Ok(self
+                .storage
+                .get_group(group_id)
+                .map_err(storage_projection_error)?
+                .epoch),
+        }
+    }
+
+    /// Discard a pass whose base epoch the device has already left behind, so the
+    /// caller reopens one at the current tip.
+    ///
+    /// A pass is local scheduling state for a single base epoch; the tip
+    /// legitimately moves while one is open (a device catching up applies
+    /// commits). The stale batch is therefore benign, not evidence of corruption,
+    /// and it carries no canonical state of its own — its members are retained
+    /// stored messages that reseed into the next pass. Same rule as the discarded
+    /// local copy in `do_join_welcome`.
+    fn discard_stale_convergence_pass(
+        &self,
+        pass: &DurableConvergencePass,
+        tip: EpochId,
+    ) -> Result<(), OpenMlsProjectionError> {
+        self.storage
+            .delete_convergence_pass(&pass.group_id)
+            .map_err(storage_projection_error)?;
+        self.audit_group(
+            &pass.group_id,
+            marmot_forensics::AuditEventKind::ConvergencePassReopened {
+                stale_base_epoch: pass.base_epoch.0,
+                current_tip_epoch: tip.0,
+                generation: pass.generation,
+            },
+        );
+        Ok(())
+    }
+
     fn load_or_open_convergence_pass(
         &self,
         group_id: &GroupId,
@@ -502,7 +537,13 @@ impl<S: StorageProvider> Engine<S> {
                 .put_convergence_pass(&consumed)
                 .map_err(storage_projection_error)?;
         }
-        if let Some(mut pass) = previous.clone()
+        if let Some(stale) = previous.as_ref()
+            && stale.is_active()
+            && stale.base_epoch != base_epoch
+        {
+            // Reopen below, at the tip this call asks for.
+            self.discard_stale_convergence_pass(stale, base_epoch)?;
+        } else if let Some(mut pass) = previous.clone()
             && pass.is_active()
         {
             #[cfg(feature = "test-policy-overrides")]
@@ -766,19 +807,6 @@ impl<S: StorageProvider> Engine<S> {
         group_id: &GroupId,
         epoch: EpochId,
     ) -> Result<(), OpenMlsProjectionError> {
-        self.mark_group_unrecoverable_for_convergence_pass(
-            group_id,
-            epoch,
-            "frozen_member_integrity",
-        )
-    }
-
-    fn mark_group_unrecoverable_for_convergence_pass(
-        &mut self,
-        group_id: &GroupId,
-        epoch: EpochId,
-        error_kind: &'static str,
-    ) -> Result<(), OpenMlsProjectionError> {
         let mut group = self
             .storage
             .get_group(group_id)
@@ -789,16 +817,14 @@ impl<S: StorageProvider> Engine<S> {
                 .put_group(&group)
                 .map_err(storage_projection_error)?;
         }
-        self.realize_group_unrecoverable_for_convergence_pass(group_id, epoch, error_kind);
+        self.realize_group_unrecoverable_for_frozen_pass(group_id, epoch);
         Ok(())
     }
 
-    fn realize_group_unrecoverable_for_convergence_pass(
-        &mut self,
-        group_id: &GroupId,
-        epoch: EpochId,
-        error_kind: &'static str,
-    ) {
+    /// A frozen pass member that no longer matches the record it was admitted
+    /// from is a real integrity failure: the batch this group already committed
+    /// to resolving cannot be reconstructed, so the group halts.
+    fn realize_group_unrecoverable_for_frozen_pass(&mut self, group_id: &GroupId, epoch: EpochId) {
         self.epoch_manager.mark_unrecoverable(group_id);
         self.events_buf.push_back(GroupEvent::GroupUnrecoverable {
             group_id: group_id.clone(),
@@ -809,15 +835,8 @@ impl<S: StorageProvider> Engine<S> {
                 phase: ConvergencePhase::Unrecoverable,
                 current_tip_epoch: Some(epoch.0),
                 retained_anchor_horizon: None,
-                reason: Some(
-                    if error_kind == "base_epoch_mismatch" {
-                        "convergence_pass_base_changed"
-                    } else {
-                        "frozen_pass_integrity_failure"
-                    }
-                    .to_string(),
-                ),
-                error_kind: Some(error_kind.to_string()),
+                reason: Some("frozen_pass_integrity_failure".to_string()),
+                error_kind: Some("frozen_member_integrity".to_string()),
             },
         );
     }
@@ -893,10 +912,7 @@ impl<S: StorageProvider> Engine<S> {
         // MLS state isn't loadable just skips those diffs.
         let previous_components = self.reorg_component_snapshot(group_id);
         let previous_name = previous_group.name.clone();
-        let previous_tip = self
-            .epoch_manager
-            .epoch(group_id)
-            .unwrap_or(previous_group.epoch);
+        let previous_tip = self.convergence_tip_epoch(group_id)?;
         let policy = self.convergence_policy_for_group(group_id)?;
         let Some(mut pass) =
             self.load_or_open_convergence_pass(group_id, previous_tip, &policy, now_ms)?
@@ -932,14 +948,6 @@ impl<S: StorageProvider> Engine<S> {
             );
             return Ok(settled_empty_result(previous_tip.0));
         };
-        if pass.is_active() && pass.base_epoch != previous_tip {
-            self.mark_group_unrecoverable_for_convergence_pass(
-                group_id,
-                previous_tip,
-                "base_epoch_mismatch",
-            )?;
-            return Ok(unrecoverable_result(previous_tip.0));
-        }
         let mut just_frozen = false;
         if pass.phase == ConvergencePassPhase::Collecting {
             let cutoff = pass.cutoff_monotonic_ms();
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -492,7 +492,7 @@ impl<S: StorageProvider> Engine<S> {
             .map_err(storage_projection_error)?;
         self.audit_group(
             &pass.group_id,
-            marmot_forensics::AuditEventKind::ConvergencePassReopened {
+            marmot_forensics::AuditEventKind::ConvergencePassDiscarded {
                 stale_base_epoch: pass.base_epoch.0,
                 current_tip_epoch: tip.0,
                 generation: pass.generation,
--- a/crates/cgka-engine/tests/distributed_convergence.rs
+++ b/crates/cgka-engine/tests/distributed_convergence.rs
@@ -5576,11 +5576,16 @@ async fn input_at_effective_cutoff_is_retained_for_the_next_generation() {
 
 #[tokio::test]
 async fn stale_pass_base_epoch_reopens_at_the_current_tip_instead_of_halting() {
-    // A device catching up advances its tip while a durable pass stays open at
-    // the epoch it was scheduled from. That is stale local scheduling state, not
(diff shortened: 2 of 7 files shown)


crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -466,6 +456,51 @@ impl<S: StorageProvider> Engine<S> {
...
+            .map_err(storage_projection_error)?;
+        self.audit_group(
+            &pass.group_id,
+            marmot_forensics::AuditEventKind::ConvergencePassReopened {
+                stale_base_epoch: pass.base_epoch.0,
+                current_tip_epoch: tip.0,
+                generation: pass.generation,
+            },
+        );
+        Ok(())
```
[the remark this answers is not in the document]
jgmontoya: Checked this against the flow, and the ordering here is deliberate — the recommended sequence would reintroduce the bug this PR fixes.

The delete isn't half of a two-step mutation. There's no "a pass must always exist" invariant: load_or_open_convergence_pass returns Ok(None) whenever nothing is eligible, and every caller already treats that as the ordinary no-input path. So "stale pass discarded, no replacement opened" is a complete, self-consistent state, not a torn write — the next call with eligible input opens at the tip.

Constructing the replacement first means keeping the stale pass when no input is currently eligible. That's precisely the field wedge: a pass anchored to a base epoch the device has already left gates outbound sends indefinitely, and every later send fails against it. Retaining it to protect an audit-row/state pairing would trade a diagnosable log line for a bricked group.

Generation continuity survives the delete: the replacement reads previous.map_or(0, |p| p.generation + 1) from the in-memory value, which the test pins as reopened.generation == stale.generation + 1.

On the cited guideline — a local storage delete isn't an external side effect, and the audit row is the intent record, so there's no compensating step to add. Leaving as is.

crates/marmot-forensics/schema/audit-log-event.v2.schema.json
```diff
@@ -1663,6 +1663,30 @@
...
+            "stale_base_epoch": {
+              "$ref": "#/$defs/u64"
+            },
+            "current_tip_epoch": {
+              "$ref": "#/$defs/u64"
+            },
+            "generation": {
+              "$ref": "#/$defs/u64"
+            }
+          }
```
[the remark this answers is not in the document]
jgmontoya: Verified this one and it doesn't hold for this contract — three reasons.

The analyzer is lenient to growth by construction. crates/incident-replay/src/export.rs documents it: "Every other kind maps to EventKind::Other, so unknown or irrelevant events never fail the parse," and the NDJSON line model tolerates unknown sections and fields. Nothing in the repo gates on schema_version.

Bumping would make things strictly worse. schema_version is a const in the schema, so a v3 bump stamps every row v3 — a consumer pinned to v2 then rejects the entire file rather than skipping one unrecognised row. The guideline's own rationale ("analyzers reject unknown versions") argues against bumping for a purely additive change.

There's also precedent: epoch_stall_backfill_armed was added to v2 in place with two new required fields and no bump, and peelerOutcomeKind was expanded in place. The AGENTS.md rule is about changing required fields; a new oneOf branch changes no existing event's required set, and the audit_log_event_schema_tracks_kind_catalog test keeps schema and Rust catalog in lockstep either way. Keeping the in-place v2 branch.

Datawav (dismissed):
Validated against current head `3f8fe3cf46b6c53e4c0b1558f543956246e41071`. The intended stale-behind recovery path is correct and its focused integration test passes, but I think these two cases should be addressed before merge:

1. **Only discard passes that are actually behind the current tip.**

   `load_or_open_convergence_pass` currently uses `stale.base_epoch != base_epoch`, so it also discards an active pass whose base is *ahead* of the current tip. The contract only establishes the benign catch-up case (`pass.base_epoch < current_tip`). For `pass.base_epoch > current_tip`, deleting and reseeding a Frozen/Resolving pass from mutable storage violates the exact-membership restart invariant and turns an inconsistent durable state into a fresh batch.

   Please make the compensation directional (`<`), retain an explicit fail-closed path for `>`, and add a future-base Frozen/Resolving regression test.

2. **Preserve generation and report the durable transition accurately when no replacement opens immediately.**

   `discard_stale_convergence_pass` deletes the only durable pass and emits `ConvergencePassReopened` before admission/reseeding. If `seed_convergence_pass_members` returns no eligible trigger—or replacement persistence fails—the call leaves no pass. A later call or restart therefore sees `previous == None` and opens generation `0`, while the audit stream already claims a reopen.

   It is valid for no active pass to exist, so this does not require preserving the stale active pass. It does require a durable last-generation/discarded state and accurate events: record a discard when only deletion completed, and emit `reopened` only after the replacement is persisted. Please add coverage for stale pass → no eligible replacement → restart → later eligible input, asserting monotonic generation.

The additive audit-schema version question and the inherited raw `group_ref` behavior are separate follow-ups; I am not requesting this PR solve them.

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -494,7 +529,13 @@ impl<S: StorageProvider> Engine<S> {
                 .put_convergence_pass(&consumed)
                 .map_err(storage_projection_error)?;
         }
-        if let Some(mut pass) = previous.clone()
+        if let Some(stale) = previous.as_ref()
+            && stale.is_active()
```
erskingardner, line 550: [P1] Do not discard Frozen or Resolving batches here

is_active() includes Collecting, Frozen, and Resolving. The benign pre-upgrade case established by this PR is a Collecting pass stamped from the stale group record. A Frozen or Resolving record is different: #1110 made its membership immutable and restart-resumable, and apply_openmls_canonicalization_result_with_profile_policy writes the pass as Completed in the same storage transaction that applies the selected branch, before the epoch manager advances. Therefore a behind-tip Frozen or Resolving record is not a normal catch-up state; deleting it re-enumerates mutable storage and abandons the exact batch that #1110 requires us to preserve. This also conflicts with the scope decision in #1174, which explicitly leaves the frozen-member integrity posture unchanged pending the separate frozen-snapshot-completeness design. Restrict benign replacement to a behind-tip Collecting pass; keep an explicit fail-closed or repair path for Frozen and Resolving mismatches and add phase-specific tests.

crates/cgka-engine/tests/distributed_convergence.rs
```diff
@@ -5063,6 +5063,130 @@ async fn input_at_effective_cutoff_is_retained_for_the_next_generation() {
...
+        .buffer_openmls_convergence_message(&group_id, commit.clone(), 1_000)
+        .unwrap();
+
+    let tip = carol.epoch(&group_id).unwrap();
+    let stale_base = EpochId(tip.0.checked_sub(1).expect("the tip is past epoch zero"));
+    let mut stale = carol_storage.convergence_pass(&group_id).unwrap().unwrap();
+    assert_eq!(stale.base_epoch, tip, "pass opens at the current tip");
+    // The durable pass outlives the tip it was scheduled from, so its base epoch
+    // falls behind as the device catches up.
+    stale.base_epoch = stale_base;
```
erskingardner, line 5719: [P1] Reproduce the state transition instead of editing durable state

This assignment manufactures the central condition by rewriting base_epoch; no engine catch-up occurs before compensation. It proves arbitrary stored-state repair but not the stated root cause that the tip legitimately advances while a pass is open, and it would still pass if the production path can never create that mismatch. The current apply path persists the selected branch and a Completed pass atomically, then advances the epoch manager, while direct commits are buffered whenever a pass is active. Please add a regression that reaches the mismatch through the actual field or upgrade path, ideally the six-commit catch-up and restart sequence, and first asserts how the stale pass was produced. If no legal path does so, narrow the fix to the actual stale-source-at-open bug instead of treating every mismatch as benign.

erskingardner (changes requested):
Request changes. The false Unrecoverable symptom is real, but the patch currently assumes more than its regression proves. I left two blocking inline findings: Frozen and Resolving passes must not be treated like stale Collecting scheduling state, and the test needs to reproduce the actual state transition instead of editing persistence directly.

There is partial overlap with #1174: it changes the Completed/fairness branch immediately above this guard and changes how the runtime calls prepare_convergence_cutoff_delay_ms. The heads diverge, so neither PR contains the other. Please coordinate sequencing, preferably #1174 first and then rebase #1182, and validate the combined behaviors: stale-pass compensation, scoped admin reservation, queued app messages, and real cutoff scheduling.

I also agree with the existing directional and generation/audit concerns already on the PR and did not duplicate them inline. CI is green at 3f8fe3c, but these are semantic gaps not covered by the current tests.

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -494,7 +529,13 @@ impl<S: StorageProvider> Engine<S> {
...
+            && stale.base_epoch != base_epoch
```
erskingardner, line 551: [P1] Scope compensation to behind-tip Collecting only

`is_active()` covers Collecting, Frozen, and Resolving, and `!=` treats a future-base pass the same as a stale one. That is broader than both the field narrative and the apply invariant:

- Successful apply writes `Completed` in the same storage transaction as the selected branch, then calls `epoch_manager.set_stable` — so a Frozen/Resolving record whose tip has already advanced is not the ordinary catch-up path (#1110's exact-membership restart).
- The dual-authority bug this PR also fixes (prepare/admit stamped from `group.epoch`, converge compared `epoch_manager`) produces a **Collecting** pass with `base < tip`.

Please require `phase == Collecting && base_epoch < tip` for discard/reopen, and keep an explicit fail-closed path for Frozen/Resolving mismatch and for `base_epoch > tip`, with phase-specific regressions.

crates/cgka-engine/src/distributed_convergence.rs
erskingardner, line 510: [P1] Generation and audit accuracy do not survive discard-without-replacement

Agree that leaving no active pass can be a valid state — callers already handle `Ok(None)`. The remaining bug is durability of generation and event meaning:

After this delete, if `seed_convergence_pass_members` returns no trigger (or the replacement write fails), storage has no pass. The next call — or a process restart — loads `previous == None` and opens generation `0`, while the audit stream already emitted `ConvergencePassReopened`.

In-memory `previous` only preserves the generation increment inside this one call. Please keep a durable last-generation / discarded marker, emit a discard event when only deletion completed, and emit `reopened` only after a replacement is persisted. Cover stale Collecting → no eligible replacement → restart → later eligible input with monotonic generation.

crates/cgka-engine/tests/distributed_convergence.rs
erskingardner, line 5719: [P1] Prove the production mismatch path, not a hand-edited base

This assignment shows stored-state repair works, but it never exercises the transition the PR claims (tip advances while a Collecting pass stays open). On the current apply path that transition is hard to reach legally: the pass is marked Completed before `set_stable`.

The dual-authority open path on master is the clearer, greppable root cause: open via prepare/admit with a lagging `group.epoch`, then converge against `epoch_manager`. Please drive the regression through that (or another documented real sequence), assert how the stale Collecting pass was produced, and only then assert compensation. If no legal path produces Frozen/Resolving mismatch, do not treat those phases as benign discard.

erskingardner (changes requested):
Request changes. The false-Unrecoverable symptom is real, and unifying tip authority via `convergence_tip_epoch` is the right surgical fix for the dual-authority stamping bug (master opens passes from `group.epoch` in prepare/admit, then checks them against `epoch_manager`). The compensation path as written is broader than that bug and than the regression proves.

Blocking gaps (overlapping with @Datawav / @erskingardner — validated independently against head `3f8fe3c`):

1. Treat only behind-tip **Collecting** as benign discard; Frozen/Resolving must not reseed from mutable storage.
2. Make the comparison directional (`base < tip`); future-base must fail closed.
3. Persist last generation / emit accurate discard-vs-reopen events across a discard that does not immediately reopen (in-memory `previous` only helps within one call).
4. Reproduce the real mismatch source in the test — do not only rewrite `base_epoch` in storage. Prefer the dual-authority open path or a documented upgrade/catch-up sequence that actually creates the stale Collecting record.

Also please rebase after #1174 (or land #1174 first) and re-validate the combined Completed/fairness + stale-pass paths.

`convergence_tip_epoch` and narrowing unrecoverable realization to frozen-member integrity can stay; the open-ended `is_active() && !=` discard should not ship.

erskingardner (dismissed):
Reviewed the diff, the linked reasoning in the PR body, and the surrounding `distributed_convergence.rs` context (not just the changed hunks).

**Recommend merge.** This is a well-scoped, well-reasoned fix.

Summary of what I verified:
- `convergence_tip_epoch` now unifies three call sites (`prepare_convergence_cutoff_delay_ms`, `admit_stored_message_to_convergence_pass`, `converge_stored_openmls_messages`) that previously read the tip from two different, sometimes-disagreeing sources (epoch manager vs. raw `group.epoch`). This closes the actual root cause, not just the halting symptom.
- `load_or_open_convergence_pass` now discards+reopens a stale active pass (`is_active() && base_epoch != base_epoch`) *before* the old post-hoc `base_epoch_mismatch` check in `converge_stored_openmls_messages` — so removing that check is safe: any pass returned from `load_or_open_convergence_pass` is now guaranteed to already match the caller's `base_epoch`. I traced this carefully since it's the crux of the fix and didn't find a gap.
- The one early-return path that bypasses the stale check (a `Completed` pass with a still-available fairness slot and queued intents) is harmless: both callers gate all real work on `phase == Collecting`, so a stale `Completed` pass returned there is a no-op either way.
- Frozen-member integrity failures are untouched and still halt via `realize_group_unrecoverable_for_frozen_pass` — the fix correctly narrows scope to the base-epoch case only, per the PR's own framing ("frozen member ... is still a real integrity failure").
- New `ConvergencePassReopened` audit kind: schema/doc-comment/test additions follow the existing pattern for adding a variant to the v2 schema (no version bump needed, consistent with prior additions like `EpochStallBackfillArmed`).
- New integration test exercises the exact field scenario from the PR description (device catches up while a pass is open) and asserts non-halting, reopened pass generation, and the audit trail — good coverage for a fix like this.
- Confirmed no leftover references to the removed `"base_epoch_mismatch"` / `"convergence_pass_base_changed"` string constants anywhere in the tree.

No blocking issues found. Nice, surgical change — net diff is small and the removed duplicate `mark_group_unrecoverable_for_convergence_pass`/`realize_group_unrecoverable_for_convergence_pass` machinery (collapsed since `error_kind` is no longer a parameter) is a genuine simplification, not just a rename.

crates/cgka-engine/src/distributed_convergence.rs
jgmontoya: The future-base direction has a benign producer on shipped code, so a directional guard would put a real record back on the halt path. resolve_fork_candidate → CandidateWins calls epoch_manager.set_stable(group_id, source_epoch) — rolling the in-memory tip BACKWARDS — and writes no group record, while master stamped base_epoch from storage.get_group().epoch (still carrying the higher merged epoch). So base > tip is what a fork rollback whose re-apply didn't complete leaves behind: the same dual-authority artifact seen from the other side. Halting on it re-ships the wedge for the fork-recovery case.

Direction stays diagnosable without a code branch — the audit row carries both stale_base_epoch and current_tip_epoch. Kept the single != predicate ("this pass isn't for the epoch we're scheduling against, so it isn't usable" — a future-base window may not contain the tip at all) and pinned the direction with a future-base regression in 03ec10cb; mutating != to < fails exactly that test.

crates/cgka-engine/src/distributed_convergence.rs
jgmontoya: Splitting this. The event name was a fair catch: the row records a discard, and the reopen is a later, separate decision that may not happen. Renamed in 436c871c — AuditEventKind::ConvergencePassDiscarded / convergence_pass_discarded, same three fields; the kind is unreleased (added in this PR), so it's a clean rename with no alias. The row stays at the delete on purpose: a discard with NO reopen is precisely the wedge-clearing event an incident analyst needs in an export, so emit-only-after-replacement would drop the most important case.

Not taking the durable last-generation marker: generation has no decision consumer anywhere in the workspace (written at open, read into this audit row, asserted in tests — nothing gates, dedups, or orders on it), and the restart-at-0 shape predates this PR (do_join_welcome and disband already delete the pass). A durable field plus a migration to make a diagnostic label monotonic is complexity without behavior.

crates/cgka-engine/tests/distributed_convergence.rs
jgmontoya: That remedy can't be executed here: prepare/admit stamping base_epoch from group.epoch while converge compares the epoch manager is exactly the code this PR deletes. With one convergence_tip_epoch() stamping and comparing, no in-tree sequence produces the mismatch — installing the durable record is the upgrade path itself (see the sibling thread for the full reachability trace).

One inversion worth flagging on the fallback inference ("if no legal path produces a Frozen/Resolving mismatch, don't treat those phases as benign"): master persists exactly that — admit freezes at the cutoff and writes the pass with no converge call in the path, so the halt only fires on the next converge. Inherited Frozen mismatches are reachable, which is why Frozen stays in the discard — now pinned by test in 03ec10cb.

jgmontoya:
> Validated against current head `3f8fe3cf46b6c53e4c0b1558f543956246e41071`. The intended stale-behind recovery path is correct and its focused integration test passes, but I think these two cases should be addressed before merge:
> 
>     1. **Only discard passes that are actually behind the current tip.**
>        `load_or_open_convergence_pass` currently uses `stale.base_epoch != base_epoch`, so it also discards an active pass whose base is _ahead_ of the current tip. The contract only establishes the benign catch-up case (`pass.base_epoch < current_tip`). For `pass.base_epoch > current_tip`, deleting and reseeding a Frozen/Resolving pass from mutable storage violates the exact-membership restart invariant and turns an inconsistent durable state into a fresh batch.
>        Please make the compensation directional (`<`), retain an explicit fail-closed path for `>`, and add a future-base Frozen/Resolving regression test.
> 
>     2. **Preserve generation and report the durable transition accurately when no replacement opens immediately.**
>        `discard_stale_convergence_pass` deletes the only durable pass and emits `ConvergencePassReopened` before admission/reseeding. If `seed_convergence_pass_members` returns no eligible trigger—or replacement persistence fails—the call leaves no pass. A later call or restart therefore sees `previous == None` and opens generation `0`, while the audit stream already claims a reopen.
>        It is valid for no active pass to exist, so this does not require preserving the stale active pass. It does require a durable last-generation/discarded state and accurate events: record a discard when only deletion completed, and emit `reopened` only after the replacement is persisted. Please add coverage for stale pass → no eligible replacement → restart → later eligible input, asserting monotonic generation.
> 
> 
> The additive audit-schema version question and the inherited raw `group_ref` behavior are separate follow-ups; I am not requesting this PR solve them.

Item 1 (directional guard): not taking it — base > tip has a benign producer on shipped code (a fork rollback moves the in-memory tip backwards and writes no group record, while base_epoch was stamped from the group record), so failing closed there puts a real inherited record back on the halt path. Both epochs are in the audit row, so direction stays analyzable without a branch. Pinned with a future-base regression in 03ec10cb.

Item 2: the event rename is done — ConvergencePassDiscarded, 436c871c (unreleased kind, clean rename). The durable generation marker isn't: generation has no decision consumer in the workspace, and the restart-at-0 shape predates this PR (join-welcome and disband already delete the pass). Fuller reasoning in the inline threads.

crates/cgka-engine/src/distributed_convergence.rs
jgmontoya: Both premises check out (is_active() covers Frozen/Resolving; the apply writes the branch and Completed in one transaction before set_stable), but the discard is still right for those phases. A Frozen/Resolving pass is pre-apply state: its members are id references into retained rows, so nothing canonical has been externalized — discarding drops a batch boundary and two deadlines, and the rows reseed. More importantly, the batch window derives from base_epoch ± max_rewind_commits, so once the base isn't the tip we schedule against, the window is stale by construction; resolving it anyway is what master did.

Failing closed here would also halt on real inherited records: master persists a Frozen stale-base pass without ever halting (admit_stored_message_to_convergence_pass freezes at the cutoff and persists, with no converge call in the path — the halt only fires on the NEXT converge). That's the field bug again, narrowed but not fixed.

Took the coverage ask: 03ec10cb adds a Frozen-with-stale-base regression so the phase decision is pinned by a test instead of being implicit.

crates/cgka-engine/tests/distributed_convergence.rs
jgmontoya: Conceded — and it goes further than the comment. I verified the tip cannot move under an active pass on this branch at all: every pass trigger admits at least one CommitEdge member (has_trigger arms), active_pass_member_gates_outbound is unconditionally true for CommitEdge so do_send queues, and inbound commits route through convergence while a pass is active (the pending_recovery bypass is only set inside the direct path, unreachable with a pass open).

So "the tip legitimately moves while a pass is open (a device catching up)" was wrong — in five places. All rewritten in 3a4ae78d to state the real mechanism: master stamped base_epoch from the durable group record while the halt compared the epoch manager; those two stores split, so an inherited pass can sit behind OR ahead of the tip. This is a repair path for records an older binary persisted, not a steady-state transition.

The install-the-durable-record test idiom stays because it IS the honest producer: on this branch a single convergence_tip_epoch() both stamps and compares, so no live sequence can create the mismatch — the durable record is exactly what an upgrade leaves behind.

jgmontoya:
> Reviewed the diff, the linked reasoning in the PR body, and the surrounding `distributed_convergence.rs` context (not just the changed hunks).
> 
> **Recommend merge.** This is a well-scoped, well-reasoned fix.
> 
> Summary of what I verified:
> 
>     * `convergence_tip_epoch` now unifies three call sites (`prepare_convergence_cutoff_delay_ms`, `admit_stored_message_to_convergence_pass`, `converge_stored_openmls_messages`) that previously read the tip from two different, sometimes-disagreeing sources (epoch manager vs. raw `group.epoch`). This closes the actual root cause, not just the halting symptom.
> 
>     * `load_or_open_convergence_pass` now discards+reopens a stale active pass (`is_active() && base_epoch != base_epoch`) _before_ the old post-hoc `base_epoch_mismatch` check in `converge_stored_openmls_messages` — so removing that check is safe: any pass returned from `load_or_open_convergence_pass` is now guaranteed to already match the caller's `base_epoch`. I traced this carefully since it's the crux of the fix and didn't find a gap.
> 
>     * The one early-return path that bypasses the stale check (a `Completed` pass with a still-available fairness slot and queued intents) is harmless: both callers gate all real work on `phase == Collecting`, so a stale `Completed` pass returned there is a no-op either way.
> 
>     * Frozen-member integrity failures are untouched and still halt via `realize_group_unrecoverable_for_frozen_pass` — the fix correctly narrows scope to the base-epoch case only, per the PR's own framing ("frozen member ... is still a real integrity failure").
> 
>     * New `ConvergencePassReopened` audit kind: schema/doc-comment/test additions follow the existing pattern for adding a variant to the v2 schema (no version bump needed, consistent with prior additions like `EpochStallBackfillArmed`).
> 
>     * New integration test exercises the exact field scenario from the PR description (device catches up while a pass is open) and asserts non-halting, reopened pass generation, and the audit trail — good coverage for a fix like this.
> 
>     * Confirmed no leftover references to the removed `"base_epoch_mismatch"` / `"convergence_pass_base_changed"` string constants anywhere in the tree.
> 
> 
> No blocking issues found. Nice, surgical change — net diff is small and the removed duplicate `mark_group_unrecoverable_for_convergence_pass`/`realize_group_unrecoverable_for_convergence_pass` machinery (collapsed since `error_kind` is no longer a parameter) is a genuine simplification, not just a rename.

Agreed on sequencing. This branch is already rebased onto master with #1173 in; once #1174 lands I'll rebase again and re-validate the stale-pass compensation together with the scoped reservation and real-cutoff scheduling.

Adjacent context: the fix making the inbound-commit epoch mirror transactional (closing the production source of the group-record/epoch-manager split that stamped stale bases in the first place) is stacked on this PR as #1184 — which is why the compensation here is scoped as a repair for records written before it. The docs on this branch now say exactly that (3a4ae78d).

erskingardner (approved):

PR #1184, opened 2026-07-29
fix(cgka-engine): mirror the inbound-commit epoch inside the merge transaction instead of best-effort

jgmontoya:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inbound commit processing so durable group “mirror” updates (epoch, membership changes, and mirrored components) and in-memory progression are applied together atomically.
  * Storage mirror-write failures now properly fail the operation so epochs and roster state can’t diverge after a restart.

* **Documentation**
  * Added durability/hydration guidance requiring the group epoch to mirror the epoch manager, with consistent durable write behavior.

* **Tests**
  * Added a failure-injection test ensuring concurrent privileged invites don’t split epoch state when mirror/recovery writes fail.
  * Broadened atomicity test documentation to cover multiple record-projection seams.

### DIFF
--- a/crates/cgka-engine/src/message_processor/ingest.rs
+++ b/crates/cgka-engine/src/message_processor/ingest.rs
@@ -55,6 +55,17 @@ struct RetainedSelfRemoveProposal {
     queued: QueuedProposal,
 }
 
+/// What an applied inbound commit changed, derived from the merged group state.
+///
+/// Produced inside the apply transaction because the durable record mirror is
+/// written from the same values the post-apply events are synthesized from.
+struct AppliedCommitProjection {
+    epoch: EpochId,
+    remaining_member_ids: std::collections::HashSet<MemberId>,
+    removed_member_ids: Vec<MemberId>,
+    message_retention_seconds: Option<u64>,
+}
+
 enum ScheduledAutoCommitReplay {
     Staged,
     NotApplicable,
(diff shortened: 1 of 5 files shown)


jgmontoya:
@coderabbitai full review

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -1222,34 +1233,84 @@ impl<S: StorageProvider> Engine<S> {
...
+                                epoch: after,
+                                remaining_member_ids: after_ids,
+                                removed_member_ids: removed,
+                                message_retention_seconds:
+                                    crate::app_components::message_retention_seconds_of_group(
+                                        &mls_group,
+                                    )?,
+                            })
+                        },
+                    )?;
```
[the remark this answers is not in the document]
jgmontoya: The asymmetry is real — the pre-merge arms all mark Failed before returning and the
in-transaction `?` doesn't — but after tracing it, no change is warranted here:

- Not introduced by this PR: the invariant check has sat inside this transaction behind a bare
  `?` since 249344d8. This commit moved the record mirror into the transaction and didn't touch
  disposition handling.

- Not reachable the way described: a commit that would flip a Legacy group to Current is
  rejected before the profile-invariants gate ever runs —
  validate_staged_commit_account_identity_proofs (account_identity_proof.rs:559) unconditionally
  refuses profile-changing commits, runs earlier on this seam (ingest.rs:1151 vs :1163), marks
  Failed on its error arm, and covers all four merge sites. Profile immutability is deliberate
  (strict cutover) and already pinned by group_commit_cannot_change_or_mix_the_protocol_profile.
  Since :1260 early-returns Ok for Legacy and the merged profile always equals the staged one,
  the only way it can fire is a divergence between the pre-merge leaf projection and the merged
  tree — defense-in-depth, and we could not construct such a commit.

- Not replay-eligible forever even then: the row is OpenMlsWire, so replay_buffered_messages
  skips it and redelivery dedups to Buffered; convergence terminalizes it as EpochInvalidated
  on the next pass.

- The storage faults that now flow through the same `?` must keep Created — the whole apply
  rolled back, so the commit genuinely still needs applying (pinned in auto_commit_atomicity).

Your rollback point is correct (with_transaction hands the closure the same handle, so an
in-closure mark would roll back), and so is the implication: a proper fix must classify by
step, not by error variant — merge_staged_commit flattens OpenMLS's StorageError into
EngineError::Backend, so "not Storage ⇒ deterministic" would terminalize transient faults.
That machinery buys nothing reachable today, so the bare `?` stays as the honest shape.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -1272,34 +1283,84 @@ impl<S: StorageProvider> Engine<S> {
...
```
erskingardner: [P1] Preserve a retry path after the mirror transaction rolls back

On a winning same-epoch fork, this error edge is reached only after `resolve_fork_candidate` has already rolled storage back to the source snapshot, released that snapshot, removed the incumbent recovery record, reset the epoch manager to the source epoch, and marked the incumbent `EpochInvalidated`. If `put_group` returns `Busy`, only the winning merge is rolled back; the recovery bookkeeping is already consumed and no pending convergence group is scheduled.

I reproduced this against exact head `fa57e9bf` by extending the new test: after the injected fault, `drain_pending_convergence_groups()` is empty, and retrying the same `TransportMessage` returns `Ok(Buffered { epoch: 1 })` rather than applying the winner because the durable `Created` row dedups before ingest. Thus the error is not self-contained/retry-safe at the engine boundary and the selected canonical winner can remain stranded until some unrelated action or restart discovers stored convergence work.

Please make candidate rollback/selection and winner application one recoverable unit, or durably preserve/reconstruct the winner and explicitly schedule convergence on this error. Extend the regression to prove eventual application after the one-shot fault, not only that epoch 1 is internally consistent.

erskingardner (changes requested):
Blocking: the SQLite transaction fixes the epoch mirror split, but a mirror fault during winning-fork recovery consumes the recovery state before this transaction and leaves no guaranteed convergence retry. Exact-head fault-injection extension reproduced the stranded state. CI is still running after the latest rebase.

crates/cgka-engine/src/message_processor/ingest.rs
jgmontoya: Confirmed — reproduced all five claims, plus one you'd surfaced implicitly: the recovery
snapshot from ingest.rs:1260 was also orphaned on this path. Fixed in 02015db4.

Went with your second option. The first doesn't fit: the rollback-to-snapshot in resolve() can't
be deferred past the merge (it's what makes re-processing the candidate possible), and the epoch
manager is in-memory, so it can't join a storage transaction. Meanwhile "durably
preserve/reconstruct the winner" already holds — the winner survives as a Created commit row and
a CommitEdge in Created opens a pass unconditionally (convergence_input.rs:130). The only
missing piece was the schedule.

The error path now releases the orphaned snapshot and calls schedule_pending_convergence_group
before propagating the original Busy unchanged. The regression is extended exactly per your ask —
eventual application, not just internal consistency: the post-fault snapshot set gains no name
(a count check would miss the orphan — it replaces the released incumbent), the drain schedules
the group, recovery driven only through that drain reaches epoch 2 with the winner in
accepted_commits and the invitee on the roster, and redelivery is then Ignored { Duplicate }
instead of the contract-violating Buffered.

Scoping note: the winner was never permanently lost (restart hydration reschedules, and the app
layer converts a later Buffered into a schedule) — but at the engine boundary the error wasn't
retry-safe, which is your point, and it now is. Also for the record: our earlier claim that a
lone retained commit can't reopen a pass was a clock artifact, not a real limitation. Thanks for
the fault-injection work — it found a real hole and a second one next to it.

erskingardner (approved):

PR #1186, opened 2026-07-30
fix: broadcast group events applied inside an outbound send

erskingardner:
## Problem

When a member's outbound send lands while a peer's commit sits retained in the convergence collection window, the engine folds that commit inside the send. The resulting `GroupStateChanged` / `EpochChanged` events surface only through the send's `AccountDeviceEffects` — and `send_app_event_with_local_projection` dropped them. Storage and projections showed the new group state, but `MarmotAppEvent::GroupEvent` was never broadcast, so runtime chat-list and group-state subscriptions stayed silent forever.

Repro (probed while hardening `convergence_settles_across_generations_with_mid_window_queued_sends` on #1174): alice renames the group, bob sends a message ~300 ms later. Bob's `chat_list_row` title and `app.group()` record flip to the new name within ~1.3 s, but `subscribe_group_state` / `subscribe_chats` never deliver an update. Instrumentation shows bob's send effects carry `[GroupStateChanged, EpochChanged]` from the folded rename commit, and no later seam re-emits them (the subsequent scheduled-convergence pass finds nothing left to apply).

This is pre-existing on master — the same repro fails identically at 82f3e2f3 — and is not introduced by #1174's scheduling changes (that PR's test carries a "subscription can stay silent … tracked separately" comment for exactly this).

## Fix

- `AppClient::observe_send_applied_effects` routes send-applied effects through the same observe pipeline as the inbound-ingest and scheduled-convergence seams (`observe_account_device_effects`: state group refresh, push-gossip handling, kind-1210 system-row synthesis — a deterministic upsert), and merges the resulting `SyncSummary` into a new `pending_applied_sync_summary` buffer. It replaces the narrower `queue_own_group_system_projection_updates` call in the message/app-event send path only; the own-commit command paths (rename, invite, retention, …) are unchanged — their worker arms already publish `GroupStateUpdated`.
- The account worker broadcasts the buffered summary via `publish_app_runtime_summary` from every seam that can run a send: the command chokepoint at the end of `handle_account_worker_command`, the receive arm's post-join push retry, the maintenance tick, and startup. Empty summaries are no-ops.

Subscribers then wake through the existing `runtime_group_event_route` / fingerprint machinery and re-read the (already refreshed) group record.

## Tests

- New regression test `group_state_subscription_observes_rename_applied_during_interleaved_send`: rename from alice, 300 ms later a send from bob, then require bob's group-state subscription to deliver the new name. Fails on master (times out with zero updates), passes with the fix. If CI timing shifts the send outside the collection window, the rename broadcasts through the scheduled-convergence seam instead and the test still passes — it only fails when events are dropped.
- `just fast-ci` clean.
- `cargo test -p marmot-app`: all targets pass except five `relay_runtime` tests (`encrypted_media_*`, `retained_media_rehydrates_*`, `relay_app_runtime_synthesizes_system_row_for_retention_change`, `self_removal_suppresses_account_unread_*`) that fail identically on clean master in this environment — pre-existing, unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Runtime chat-list and group-state subscriptions now reliably reflect group state changes (such as renames), even when updates originate from send-side effects or are folded through convergence.
  * Subscription updates remain visible even if an outgoing publish attempt fails, avoiding gaps between persisted state and live notifications.
  * Improved timing so “send applied” updates are broadcast promptly after relevant events.
* **Tests**
  * Updated and expanded relay runtime regression tests for deterministic scheduling, including rename retention across interleaved sends and publish failures.

### DIFF
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -1738,7 +1738,8 @@ impl AppClient {
             // folded are durably applied — broadcast them before surfacing the
             // publish failure, best-effort so a projection error cannot mask
             // the primary error.
-            self.observe_send_applied_effects_best_effort(&effects).await;
+            self.observe_send_applied_effects_best_effort(&effects)
+                .await;
             if let Err(_save_err) = self.app.save_state(&self.state) {
                 tracing::warn!(
                     target: "marmot_app::messages",
@@ -1788,7 +1789,8 @@ impl AppClient {
         // for the account worker to broadcast; dropping the events here leaves
         // storage renamed while chat-list/group-state subscribers never wake.
         // Best-effort: a projection failure must not fail a completed publish.
-        self.observe_send_applied_effects_best_effort(&effects).await;
+        self.observe_send_applied_effects_best_effort(&effects)
+            .await;
         self.app.save_state(&self.state)?;
         if published.is_some() && notification_trigger_for_intent(&intent).is_some() {
             self.publish_notification_trigger_best_effort(
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -1688,64 +1688,73 @@ impl AppClient {
             on_local_projection(update);
         }
 
-        let effects = match async {
-            let effects = match self.sync_runtime_groups().await {
-                Ok(()) => {
-                    let send_intent = SendIntent::AppMessage {
-                        group_id: group_id.clone(),
-                        payload,
-                    };
-                    // Thread the human-action context through the engine so the
-                    // send's audit rows carry `human_action`, matching
-                    // create_group/invite/etc.
-                    match &audit_context {
-                        Some(context) => {
-                            self.runtime
-                                .send_with_audit_context(send_intent, context.clone())
-                                .await?
-                        }
-                        None => self.runtime.send(send_intent).await?,
+        let send_result = match self.sync_runtime_groups().await {
+            Ok(()) => {
+                let send_intent = SendIntent::AppMessage {
+                    group_id: group_id.clone(),
+                    payload,
+                };
+                // Thread the human-action context through the engine so the
+                // send's audit rows carry `human_action`, matching
+                // create_group/invite/etc.
+                match &audit_context {
+                    Some(context) => {
+                        self.runtime
+                            .send_with_audit_context(send_intent, context.clone())
+                            .await
                     }
+                    None => self.runtime.send(send_intent).await,
                 }
-                Err(error) if error.is_account_not_active() => {
-                    let context = audit_context.clone().unwrap_or_default();
-                    self.runtime
-                        .queue_app_message_with_audit_context(group_id.clone(), payload, context)
-                        .await?
-                }
-                Err(error) => return Err(error),
-            };
-            fail_if_publish_failed(&effects)?;
-            Ok::<_, AppError>(effects)
-        }
-        .await
-        {
+                .map_err(AppError::from)
+            }
+            Err(error) if error.is_account_not_active() => {
+                let context = audit_context.clone().unwrap_or_default();
+                self.runtime
+                    .queue_app_message_with_audit_context(group_id.clone(), payload, context)
+                    .await
+                    .map_err(AppError::from)
+            }
+            Err(error) => Err(error),
+        };
+        // The publish-status gate is applied separately from obtaining the
+        // effects: even when the outbound publish hard-fails, the engine may
+        // already have folded retained peer commits into this send, and those
+        // applied `GroupStateChanged` / `EpochChanged` events must still be
+        // observed and broadcast — only the local message is retracted.
+        let effects = match send_result {
             Ok(effects) => effects,
             Err(err) => {
-                if should_project_locally {
-                    // No read-marker rollback needed: the marker only advances
-                    // in the post-publish success projection below.
-                    match self.app.invalidate_timeline_app_event(
-                        &self.state.label,
-                        &group_id_hex,
-                        &app_event_id,
-                        "local_publish_failed",
-                    ) {
-                        Ok(Some(update)) => on_local_projection(update),
-                        Ok(None) => {}
-                        Err(_) => {
-                            tracing::warn!(
-                                target: "marmot_app::messages",
-                                method = "send_app_event_with_local_projection",
-                                error_code = "local_projection_retract_failed",
-                                "failed to retract local projection after publish failure"
-                            );
-                        }
-                    }
-                }
+                self.retract_failed_local_projection(
+                    should_project_locally,
+                    &group_id_hex,
+                    &app_event_id,
+                    &mut on_local_projection,
+                );
                 return Err(err);
             }
         };
+        if let Err(publish_err) = fail_if_publish_failed(&effects) {
+            // The send itself failed to reach anyone, but any peer commits it
+            // folded are durably applied — broadcast them before surfacing the
+            // publish failure, best-effort so a projection error cannot mask
+            // the primary error.
+            self.observe_send_applied_effects_best_effort(&effects).await;
+            if let Err(_save_err) = self.app.save_state(&self.state) {
+                tracing::warn!(
+                    target: "marmot_app::messages",
+                    method = "send_app_event_with_local_projection",
+                    error_code = "send_applied_state_save_failed",
+                    "failed to persist state observed from a failed send"
+                );
+            }
+            self.retract_failed_local_projection(
+                should_project_locally,
+                &group_id_hex,
+                &app_event_id,
+                &mut on_local_projection,
+            );
+            return Err(publish_err);
+        }
         if let Some(context) = &audit_context {
             self.record_human_action_succeeded(group_id, context, &effects);
         }
@@ -1778,7 +1787,8 @@ impl AppClient {
         // `queue_own_group_system_projection_updates`) — and buffer the summary
         // for the account worker to broadcast; dropping the events here leaves
         // storage renamed while chat-list/group-state subscribers never wake.
-        self.observe_send_applied_effects(&effects).await?;
+        // Best-effort: a projection failure must not fail a completed publish.
+        self.observe_send_applied_effects_best_effort(&effects).await;
         self.app.save_state(&self.state)?;
         if published.is_some() && notification_trigger_for_intent(&intent).is_some() {
             self.publish_notification_trigger_best_effort(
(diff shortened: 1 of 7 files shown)


crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -3422,6 +3422,81 @@ async fn app_runtime_chat_and_group_state_subscriptions_stream_projection_update
...
+            None,
+        )
+        .await
+        .unwrap();
+    // Give the rename commit time to reach bob's collection window, then send
+    // from bob while the window is still open so the send folds the commit.
+    // If timing shifts and convergence applies the commit first, the rename
+    // still broadcasts through the scheduled-convergence seam — either way the
+    // subscription must deliver it.
+    sleep(Duration::from_millis(300)).await;
```
erskingardner: Blocking: this sleep cannot force the mid-window fold under CI.

`relay_runtime` runs with `--features test-policy-overrides`, and `MarmotApp::with_relays_and_config` defaults `dev_settlement_quiescence_ms` to `0` when unset. With a zero collection window, alice’s rename settles through the scheduled-convergence seam before bob’s send; the subscription then wakes for reasons unrelated to this PR’s send-path observe/broadcast.

Please pin a non-zero quiescence for this test (e.g. `MarmotAppConfig::default().with_dev_settlement_quiescence_ms(10_000)` plus the existing loopback flags), send inside that window, and assert the subscription update on a timeout much shorter than the quiescence. That is what fails on master when send-applied `GroupStateChanged` events are dropped.

crates/marmot-app/src/client/mod.rs
```diff
@@ -1762,8 +1770,16 @@ impl AppClient {
...
         }
+        // A send that lands while inbound convergence input is retained folds
+        // those commits before publishing, so `effects.events` can carry peer
+        // state changes (e.g. a mid-window group rename). Observe them through
+        // the same pipeline as inbound deliveries — state group refresh plus
+        // kind-1210 system-row synthesis (replacing the narrower
+        // `queue_own_group_system_projection_updates`) — and buffer the summary
+        // for the account worker to broadcast; dropping the events here leaves
+        // storage renamed while chat-list/group-state subscribers never wake.
+        self.observe_send_applied_effects(&effects).await?;
```
erskingardner: Non-blocking: this `?` is after the engine publish and local message projection already succeeded. If `observe_account_device_effects` / `sync_runtime_groups` fails here, the caller sees a failed send for a message that already landed.

Not unique to this PR (inbound observe also propagates), but worth considering best-effort + warn on the outbound path so projection/subscriber refresh cannot fail a completed publish.

erskingardner (commented):
Request changes (posted as comment — GitHub disallows REQUEST_CHANGES on own PRs) — fix looks correct and needed; the regression test does not pin the failure mode under CI.

## Verdict

Real bug, right layer of fix. Reusing `observe_account_device_effects` + buffering a `SyncSummary` for the worker publish seams is the minimal way to close the gap that #1174 already called out (storage/projection can flip via a later empty `refresh_group` while subscribers stay silent).

## Blocking

The new `relay_runtime` test can pass without ever hitting the send-fold path this PR fixes. CI runs that target with `--features test-policy-overrides`, and `MarmotApp::with_relays_and_config` defaults `dev_settlement_quiescence_ms` to `0` when unset — so there is effectively no collection window. A 300ms sleep then almost certainly lets scheduled convergence apply/broadcast the rename before bob’s send, which is exactly the alternate path the test comment blesses as success.

So this is not a reliable regression guard for the bug you reproduced.

Suggested shape: construct the app with an explicit large quiescence (e.g. `with_dev_settlement_quiescence_ms(10_000)`), deliver alice’s rename, send from bob well inside that window, and require the group-state subscription update on a short timeout (≪ quiescence). That fails on master for the right reason and only passes via the send-path observe/broadcast.

## Non-blocking

- `observe_send_applied_effects(...).await?` runs after a successful publish/local projection. An observe/`sync_runtime_groups` failure can now fail an already-published send. Same pattern as inbound observe, but outbound is a sharper UX cliff; best-effort + warn may be preferable later.
- Doc comments on the new helpers are accurate but long relative to the change size.

Happy to re-review once the test forces the fold path under `test-policy-overrides`.

crates/marmot-app/src/client/mod.rs
erskingardner: [P1] Stronger/blocking version of this: the new observer is also downstream of `fail_if_publish_failed(&effects)?`. A plain app-message publish that reaches no endpoint is a hard error, but `effects.events` may already contain the retained peer commit that the send folded before attempting its own publish. The closure then discards `effects`, the local-message rollback returns the publish error, and this observer never runs, so `GroupStateChanged` / `EpochChanged` are still permanently lost on the same send-apply seam whenever the outbound publish fails.

Please preserve and observe/buffer those effects before converting the publish status into the command error (while still retracting the failed local message), and cover the failed-publish interleaving as well.

erskingardner (commented):
Request changes (GitHub does not allow the PR author to submit a REQUEST_CHANGES event).

The success-path fix is in the right layer, but two blocking gaps remain on the current head:

- Hard outbound publish failures still discard send-applied peer group events because `fail_if_publish_failed` returns before the new observer runs. The local message should still fail/retract, but the already-applied group effects must be observed and broadcast.
- The regression test does not force the send-fold path under the CI test policy, so it can pass through scheduled convergence without exercising this fix.

I posted the publish-failure case in the existing send-observer thread; the earlier inline test comment has the deterministic quiescence/timeout shape. The diff is otherwise clean and the success-path observer/buffer design is appropriately scoped.

crates/marmot-app/tests/relay_runtime.rs
erskingardner: [P1] This is improved, but the current test still does not pin the send-applied path. The doc comment claims `published > 0` is the witness, yet the test discards every `SendSummary`; `rename_applied` only polls the chat row for up to a full 1s quiescence window after each send. The comment also explicitly allows scheduled convergence to apply/broadcast the rename. A delayed runner that crosses the cutoff therefore passes this test on master without exercising `observe_send_applied_effects`.

Please keep the cutoff well beyond the assertion horizon under `test-policy-overrides` and require a pre-send old row followed by a successful send plus an immediate post-send renamed row (or expose a deterministic send-effects/pass-phase witness). The subscription assertion must complete before scheduled convergence can be the source.

crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -3422,6 +3444,227 @@ async fn app_runtime_chat_and_group_state_subscriptions_stream_projection_update
...
+    let send_result = runtime
+        .send_message(&bob_id, &group_id, b"bob rejected".to_vec())
+        .await;
+    // Disarm before asserting so drain/retry publishes can proceed on runners
+    // where the send took the queued path instead of failing outright.
+    reject_armed.store(false, Ordering::Relaxed);
+    // A folded send surfaces the injected rejection as a hard error; a queued
+    // send returns Ok with `published == 0`. The subscription contract below
+    // holds for both.
+    if let Ok(summary) = &send_result {
```
erskingardner: [P1] Require the hard-publish path this regression test is meant to cover. Accepting `Ok(published == 0)` means the send queued and never reached `fail_if_publish_failed`; after the test disarms the relay, scheduled convergence/drain can apply and broadcast the rename on unmodified master, so the test passes without the fix. Pin the collection window/barrier so the send folds, assert `send_result` is the injected publish error, and verify the row/subscription changed as a consequence of that failed send while the alternate drain path is still blocked.

erskingardner (commented):
Request changes (GitHub does not allow the PR author to submit a REQUEST_CHANGES event).

The implementation now addresses both prior code findings: send-applied effects are observed before a hard publish error is returned, and observation cannot fail an already-completed publish. The remaining blocker is deterministic regression coverage.

Both added relay tests still admit alternate paths that work on master: the success test waits through the full convergence cutoff without asserting the send caused the row transition, and the failed-publish test explicitly accepts a queued `Ok` result before re-enabling the scheduled drain. I posted the exact assertions/barriers needed inline.

Current diff is clean; fresh relay-runtime, Rust, clippy, format, packaging, and most integration checks pass, with CLI E2E and simulator/vectors still running at review time.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -2199,6 +2216,22 @@ fn publish_client_pending_projection_updates(
     }
 }
 
+/// Broadcast group events a send applied as a side effect (retained inbound
+/// convergence commits folded before publishing). Called from every worker seam
+/// that can run a send — the command chokepoint, the receive arm's post-join
```
erskingardner, line 2251: [P2] Cover the reconnect push-retry seam

This comment says every worker seam is covered, but the reconnect branch still calls `reopened.retry_pending_push_registration_shares_best_effort()` at lines 851-854 and then assigns `client = reopened` without draining `pending_applied_summary`. If that retry folds a retained commit, subscribers remain stale until an unrelated command or a successful 15-second maintenance pass drains it. Add `publish_client_pending_applied_summary(&mut reopened, ...)` before the assignment (and ideally pin this path with a test). The latest commit message says this was fixed, but that commit changes only `relay_runtime.rs`; the reconnect code is unchanged from the previous head.

crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -3422,6 +3444,246 @@ async fn app_runtime_chat_and_group_state_subscriptions_stream_projection_update
...
+    let mut witnessed = None;
+    for round in 0..8u32 {
+        let renamed = format!("renamed mid send {round}");
+        runtime
+            .update_group_profile(&alice_id, &group_id, Some(renamed.clone()), None)
+            .await
+            .unwrap();
+        // ~300ms lands the send mid-window, after bob ingested and retained
+        // the commit; observed timing folds it into the send at that offset.
+        sleep(Duration::from_millis(300)).await;
```
erskingardner: [P1] Replace the timing guess with a deterministic fold barrier

The current Relay runtime tests job shows that this 300 ms assumption does not hold on the CI runner: both new tests exhaust all eight rounds and panic at `witnessed.expect` before reaching the subscription assertion. Besides leaving the required check red, this means the regression is not a stable guard for the send-fold behavior. Please add a test seam/barrier that holds Bob in the retained/collecting state (or exposes the pass state) and trigger the send only after that precondition is observed; increasing delays or retry counts would remain probabilistic.

Failing job: https://github.com/marmot-protocol/mdk/actions/runs/30521241496/job/90801900122

erskingardner (commented):
Request changes.

The send-path observer now preserves applied effects across both successful and hard-failed publication, and observing those effects is best-effort so an already-published send is not converted into a client error. Those earlier production-path findings are resolved.

Two blockers remain:

- [P1] The fresh Relay runtime tests job fails both new regression tests before either reaches the subscription assertion. The fixed 300 ms timing plus eight retries never witnesses the fold on CI, so this needs a deterministic barrier/test seam rather than more delay or retries: https://github.com/marmot-protocol/mdk/actions/runs/30521241496/job/90801900122
- [P2] The reconnect path still runs `retry_pending_push_registration_shares_best_effort()` and assigns `reopened` back to `client` without draining the applied summary. A folded commit on that seam remains unbroadcast until another command or a successful 15-second maintenance pass. The latest commit message claims this drain, but the commit changes only the relay-runtime tests.

All other current checks are green. Because this is the authenticated authors own PR, GitHub does not permit a formal Request changes state; this comment review records the recommendation.

crates/marmot-app/tests/relay_runtime.rs
erskingardner: Fixed in 15ec5203. The regression now uses a test-policy-only scheduler hold to create the exact post-cutoff/pre-scheduler state: the completed catch-up is the ingest barrier, the 100 ms engine cutoff elapses while scheduled convergence is held for 60 s, and both tests require strict pre-send old row, send-path outcome, immediate post-send renamed row, and a subscription update within 5 s. The exact CI relay command passes 84/84 and `just fast-ci` passes.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in 15ec5203. The reconnect branch now drains and publishes the pending applied summary from `reopened` immediately after the push-registration retry and before assigning it back to `client`.

crates/marmot-app/tests/relay_runtime.rs
erskingardner: Addressed in 15ec5203. The regression now pins a deterministic post-cutoff/pre-scheduler window with the test-only 60s scheduler delay, requires the pre-send row to be old, requires a successful send with published > 0, then asserts the immediate row and subscription update. The exact 84-test relay CI job passes.

Issue #1189, opened 2026-07-30
Define normative cross-layer convergence semantics before campaign and VM expansion


Datawav:
## Problem

The convergence-reliability plan correctly separates the engine, scenario, and application/process layers, but it does not yet define a normative refinement contract that makes “the same scenario” and “equivalent outcomes” mean the same thing across them.

This is visible in the current implementation at `d38d9937f39faa8c59852a8cdced285e7be4fa06`:

- `ScenarioSpec` v1 constructs `HarnessClient` and `TransportBus` directly rather than targeting an adapter-neutral subject (`crates/cgka-conformance-simulator/src/scenario.rs`).
- Existing steps expose harness mechanics such as queue indices and queue permutations.
- The plan proposes capability discovery and a common observation schema, but those alone do not define action preconditions, atomicity, time, publication acknowledgements, permitted nondeterminism, or allowed layer-specific outcomes.

There are also sequencing tensions in the plan:

- Milestone 1 requires targeted semantic mutations to be detected, while mutation adequacy is introduced in Milestone 4.3.
- Milestone 3 expands into adversarial campaigns before Milestone 4.1 introduces the independent reference model.
- Milestone 5.2 proposes a simulator node binary even though the external `darkmatter-automated-testing` harness already drives MDK's `wn`/`wnd` process surfaces.

Without resolving these contracts first, adapters may all pass while interpreting actions, `eventually`, publication acknowledgement, quiescence, or terminal failure differently. Failures found in VM campaigns may also be impossible to minimize and promote into cheaper engine regressions.

Established by #1188. Related implementation/campaign tracking: #1032, #1176, #774, and #972.

### 1. Define the engine and cross-layer contracts

Document the three engine input representations and their ownership:

1. transport envelope and exact bytes;
2. authenticated post-peel MLS input;
3. normalized symbolic input used by the independent model.

For every Scenario IR action, define an adapter refinement map containing:

- preconditions and unsupported-capability behavior;
- atomicity/linearization point;
- clock domain and time advancement semantics;
- controllable and permitted nondeterminism;
- normalized observations and state commitments;
- permitted terminal outcomes.

Runtime-loaded scenarios must fail pre-execution schema/profile validation when unsupported; they must not be skipped or weakly reinterpreted.

### 2. Move minimum oracle independence before campaigns

Before Milestone 3, require:

- a small independent selector/canonicalization model;
- a sentinel suite of semantic mutations that the oracle must reject;
[...]

PR #1190, opened 2026-07-30
Add exact convergence oracle and scenario input ledger

erskingardner:
## Summary

- add a feature-gated, privacy-safe canonical engine projection for live and disbanded groups, including a domain-separated exporter commitment
- add exact simulator expectations for canonical equivalence, generalized commit/proposal/application dispositions, instantaneous pending work, and bidirectional decryptability
- make strict oracle coverage the default for reliability report campaigns while preserving an explicit exploratory opt-out
- update the convergence reliability plan with pass-partition methodology, relay-history completeness and EOSE work, durable repair outcomes, clock injection, staged Scenario IR delivery, and adapter-specific quiescence

## Generalized disposition ledger

Every convergence-relevant scenario action now receives a stable scenario id. The harness correlates that action with outer transport and content-derived ids through the conformance-only durable view, so commit senders do not need to re-peel their own post-advance commit. The ledger records current dispositions and counts for commits, proposals, and application messages without exposing transport ids in reports.

## Known strict findings retained

The stronger final boundary intentionally continues to expose pre-existing reliability findings instead of relaxing the oracle:

- a transported commit that is locally rolled back can still advance another member
- pre-membership transport input can remain pending after a later join
- self-remove convergence can leave proposal/scheduler work behind
- the mixed large storm has an application phase with weak outcome coverage

Regression tests pin these as known red campaign evidence while the ordinary test suite remains green.

## Verification

- cargo test -p cgka-conformance-simulator
- cargo test -p cgka-engine --features "test-conformance-snapshot test-policy-overrides"
- just fast-ci
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added exact canonical-state observations for more precise scenario comparisons.
  * Added scenario-input tracking for sends, delivery, deduplication, invalidation, and pending status.
  * Added bidirectional decryptability probes with per-direction results.
  * Added pending-work reporting to identify remaining queued or deferred activity.
* **Bug Fixes**
  * Improved handling and reporting of disbanded group state and retained membership data.
* **Documentation**
  * Expanded simulator scenario, reporting, and reliability verification guidance.
* **Chores**
  * Reports now enforce strict oracle coverage by default; exploratory runs can opt into weaker validation.

### DIFF
--- a/crates/cgka-conformance-simulator/src/bus.rs
+++ b/crates/cgka-conformance-simulator/src/bus.rs
@@ -20,8 +20,10 @@
 //! broadcast welcomes (then the engine's `NotForThisClient` filter kicks
 //! in client-side).
 
+use crate::pending_work::BusPendingWorkSnapshot;
+use crate::scenario_input_ledger::ScenarioInputMetadata;
 use cgka_traits::transport::{TransportEnvelope, TransportMessage};
-use cgka_traits::types::MemberId;
+use cgka_traits::types::{MemberId, MessageId};
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 
@@ -62,6 +64,8 @@ struct Inner {
     /// If Some, only deliver to clients in this allowlist (partition).
     partition_allowed: Option<std::collections::HashSet<ClientId>>,
     delayed: HashMap<String, Vec<InFlight>>,
+    scenario_input_by_transport_id: HashMap<MessageId, ScenarioInputMetadata>,
+    scenario_input_by_content_id: HashMap<MessageId, ScenarioInputMetadata>,
 }
 
 impl TransportBus {
@@ -81,6 +85,8 @@ impl TransportBus {
                 mailboxes: HashMap::new(),
                 partition_allowed: None,
                 delayed: HashMap::new(),
+                scenario_input_by_transport_id: HashMap::new(),
+                scenario_input_by_content_id: HashMap::new(),
             })),
         }
     }
@@ -100,6 +106,46 @@ impl TransportBus {
     pub fn send(&self, sender: ClientId, msg: TransportMessage) {
         let mut inner = self.inner.lock().unwrap();
         inner.queue.push(InFlight { sender, msg });
+    }
+
+    pub(crate) fn register_scenario_input(
+        &self,
+        transport_id: MessageId,
+        content_id: MessageId,
+        mut metadata: ScenarioInputMetadata,
+    ) {
+        metadata.aliases = vec![transport_id.clone(), content_id.clone()];
+        let mut inner = self.inner.lock().unwrap();
+        inner
+            .scenario_input_by_transport_id
+            .insert(transport_id, metadata.clone());
+        inner
+            .scenario_input_by_content_id
+            .insert(content_id, metadata);
+    }
+
+    pub(crate) fn scenario_input_for_transport(
+        &self,
+        message_id: &MessageId,
+    ) -> Option<ScenarioInputMetadata> {
+        self.inner
+            .lock()
+            .unwrap()
+            .scenario_input_by_transport_id
+            .get(message_id)
+            .cloned()
+    }
+
+    pub(crate) fn scenario_input_for_content(
+        &self,
+        message_id: &MessageId,
+    ) -> Option<ScenarioInputMetadata> {
+        self.inner
+            .lock()
+            .unwrap()
+            .scenario_input_by_content_id
+            .get(message_id)
+            .cloned()
     }
 
     /// Deliver up to `n` messages from the queue into per-client mailboxes,
@@ -187,6 +233,15 @@ impl TransportBus {
     /// Number of queued (not yet delivered) messages.
     pub fn queued_len(&self) -> usize {
         self.inner.lock().unwrap().queue.len()
+    }
+
+    pub(crate) fn pending_work_snapshot(&self) -> BusPendingWorkSnapshot {
+        let inner = self.inner.lock().unwrap();
+        BusPendingWorkSnapshot {
+            queued_messages: inner.queue.len(),
+            delayed_messages: inner.delayed.values().map(Vec::len).sum(),
+            mailbox_messages: inner.mailboxes.values().map(Vec::len).sum(),
+        }
     }
 
     /// Snapshot queued messages without altering delivery order.
--- a/crates/cgka-conformance-simulator/src/bus.rs
+++ b/crates/cgka-conformance-simulator/src/bus.rs
@@ -235,12 +235,25 @@ impl TransportBus {
         self.inner.lock().unwrap().queue.len()
     }
 
-    pub(crate) fn pending_work_snapshot(&self) -> BusPendingWorkSnapshot {
+    pub(crate) fn pending_work_snapshot(&self, client: ClientId) -> BusPendingWorkSnapshot {
         let inner = self.inner.lock().unwrap();
+        let identity = inner
+            .clients
+            .get(&client)
+            .expect("pending-work client is attached");
         BusPendingWorkSnapshot {
-            queued_messages: inner.queue.len(),
-            delayed_messages: inner.delayed.values().map(Vec::len).sum(),
-            mailbox_messages: inner.mailboxes.values().map(Vec::len).sum(),
+            queued_messages: inner
+                .queue
+                .iter()
+                .filter(|in_flight| targets_client(&inner.policy, identity, client, in_flight))
+                .count(),
+            delayed_messages: inner
+                .delayed
+                .values()
+                .flatten()
+                .filter(|in_flight| targets_client(&inner.policy, identity, client, in_flight))
+                .count(),
+            mailbox_messages: inner.mailboxes.get(&client).map_or(0, Vec::len),
         }
     }
 
@@ -322,6 +335,28 @@ impl TransportBus {
     /// when a test wants to inspect ordering before stepping.
     pub fn peek_policy(&self) -> DeliveryPolicy {
         self.inner.lock().unwrap().policy.clone()
+    }
+}
+
+fn targets_client(
+    policy: &DeliveryPolicy,
+    identity: &MemberId,
+    client: ClientId,
+    in_flight: &InFlight,
+) -> bool {
+    if in_flight.sender == client {
+        return false;
+    }
+    match &in_flight.msg.envelope {
+        TransportEnvelope::GroupMessage { .. } => true,
+        TransportEnvelope::Welcome { recipient } => {
+            matches!(
+                policy,
+                DeliveryPolicy::Ordered {
+                    broadcast_welcomes: true
+                }
+            ) || recipient == identity
+        }
     }
 }
 
--- a/crates/cgka-conformance-simulator/src/client.rs
+++ b/crates/cgka-conformance-simulator/src/client.rs
@@ -697,20 +697,11 @@ impl HarnessClient {
                 welcomes,
                 pending,
             } => {
-                let scenario_input =
-                    self.next_scenario_input_metadata(ScenarioInputKind::Commit, None, None);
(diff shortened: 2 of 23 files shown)


crates/cgka-conformance-simulator/src/scenario_input_ledger.rs
```diff
@@ -0,0 +1,516 @@
...
+        entry.pending = true;
+    }
+
+    pub(crate) fn record_published(&mut self, metadata: &ScenarioInputMetadata) {
+        self.remember_metadata(metadata);
+        let entry = self.entry(metadata);
+        entry.published += 1;
+        if metadata.kind == ScenarioInputKind::Application {
+            entry.disposition = ScenarioInputDisposition::Accepted;
+            entry.pending = false;
```
erskingardner, line 131: **Blocking:** clearing `pending` here makes published-but-undelivered application inputs invisible to `scenario_inputs_pending` / `NoPendingWork`.

After `DropQueued` / omit, the bus counters go to zero while disposition stays `Accepted` and `pending=false`, so a final strict `NoPendingWork` can pass with no delivery or invalidation. That conflicts with the `NoPendingWork` docs (“logical-message work”) and with using this as the default settlement gate for generated campaigns.

Keep application inputs pending until a terminal/current outcome (`Delivered`, `Expired`, `Invalidated`, `Rejected`, …), or stop counting application ledger pending in `NoPendingWork` and add an explicit drop/omit sentinel plus narrower docs.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1119,6 +1332,64 @@ impl HarnessClient {
...
+            .engine()
+            .conformance_pending_work_snapshot(&group_id)
+            .expect("capture conformance pending work");
+        let bus = self.bus.pending_work_snapshot();
+        PendingWorkObservation {
+            engine,
+            bus_queued_messages: bus.queued_messages,
+            bus_delayed_messages: bus.delayed_messages,
+            bus_mailbox_messages: bus.mailbox_messages,
+            scenario_inputs_pending: self.scenario_input_tracker.pending_count(),
```
erskingardner, line 1341: **Blocking / contract mismatch:** this observation mixes per-client engine/ledger state with **global** bus totals (`queued` / `delayed` / all mailboxes).

`NoPendingWork { clients: [alice] }` therefore fails if Bob still has mailbox traffic, and conversely cannot express “Alice alone is quiescent.” Either:

1. make bus counters client-scoped (this client’s mailbox + messages addressed to it still in queue/delay), or
2. document/`NoPendingWork` as a scenario-global quiescence check and stop attaching the same global bus snapshot to every client observation.

crates/cgka-engine/src/conformance_snapshot.rs
```diff
@@ -0,0 +1,502 @@
...
+            engine.has_pending_convergence_inputs(group_id)?
+        }),
+        queued_outbound_intents: engine.storage.list_queued_outbound_intents(group_id)?.len(),
+        stored_created_messages,
+        stored_retryable_messages,
+        stored_transport_deferred_messages,
+        pending_engine_events: engine.events_buf.len(),
+        pending_auto_publish: engine.auto_publish_buf.len(),
+        pending_auto_proposals: engine.auto_proposal_buf.len(),
+        valid_proposal_schedule_signals: engine.valid_proposal_groups.len(),
```
erskingardner: **Non-blocking but real:** several fields here are process-global (`events_buf`, `auto_*_buf`, `valid_proposal_groups.len()`, `pending_state_changes.len()`) while the API is framed as a per-`group_id` snapshot. Neighboring fields correctly filter by `group_id`.

In the current one-group harness this is mostly harmless, but it will false-positive `NoPendingWork` as soon as a subject has multiple live groups. Prefer `contains(group_id)` / filter-by-group counts for the schedule/buffer signals that can be attributed.

crates/cgka-conformance-simulator/src/vector.rs
```diff
@@ -123,6 +127,31 @@ pub enum TraceExpectation {
...
+    ClientsExactlyEquivalent {
+        clients: Vec<String>,
+    },
+    /// Require the complete per-client commit/proposal/application input ledger.
+    ScenarioInputLedger {
+        client: String,
+        entries: Vec<ScenarioInputLedgerEntry>,
+    },
+    /// Require every named client's exact progress snapshot to contain no
+    /// outstanding engine, transport, or logical-message work.
```
erskingardner: This rustdoc overpromises given current ledger semantics: application inputs clear `pending` at publish, so “logical-message work” is not what `scenario_inputs_pending` measures.

Whatever fix lands for `record_published`, please rewrite this to the actual guarantee (engine/bus/progress counters only, or true unsettled scenario-input work including undelivered apps).

crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
```diff
@@ -117,6 +120,484 @@ async fn three_client_happy_path_via_harness() {
...
+        .await;
+    assert!(
+        pending.is_none(),
+        "current founding group is immediately live"
+    );
+    alice.drain_events();
+
+    alice.request_disband().await.expect("request disband");
+    for _ in 0..24 {
+        tokio::time::sleep(std::time::Duration::from_millis(250)).await;
```
erskingardner, line 193: **Non-blocking:** up to ~6s of wall sleeps makes this flake-prone under load. Prefer advancing an injected harness clock / polling without `sleep`, especially since the plan now calls clocks a dependency for M1.3.

erskingardner (commented):
## Verdict

Found serious issues in the new settlement boundary. Milestone 1.1 is the right next slice, and keeping the known strict-campaign reds instead of weakening the oracle is the correct posture. The main problem: `NoPendingWork` currently overclaims “logical-message” settlement while application inputs clear `pending` at publish time, so a dropped/omitted app can leave an empty progress snapshot with `Accepted` and no delivery.

## What’s solid

- Feature-gated exporter commitment + disband tombstone projection look well scoped and privacy-conscious for synthetic use.
- Generalized scenario-input ledger with stable action ids and transport/content alias join is the right harness boundary.
- Strict-by-default reports with `--allow-weak-oracle`, plus pinned known-red chaos evidence, is excellent failure hygiene.
- Bidirectional decryptability probe is a good complement to exact state equality.

## Required before merge

1. Application inputs must not leave `pending=false` at publish/accept if `NoPendingWork` is the final settlement gate for generated campaigns — or narrow that expectation so it no longer claims logical-message completeness and add an explicit drop/omit sentinel.
2. Tighten the per-client vs global progress-snapshot story (bus totals and a few engine buffers are process-global today).

## Non-blocking

- Sleep-loop disband test is timing-sensitive; prefer injected clock / event-driven wait.
- Scenario-input registration in `client.rs` is heavily duplicated and will drift.
- Plan expansion is useful but should cross-link #1189; this PR does not satisfy that issue’s normative IR/refinement acceptance criteria.

crates/cgka-conformance-simulator/src/bus.rs
```diff
@@ -189,6 +235,15 @@ impl TransportBus {
         self.inner.lock().unwrap().queue.len()
     }
 
+    pub(crate) fn pending_work_snapshot(&self) -> BusPendingWorkSnapshot {
+        let inner = self.inner.lock().unwrap();
+        BusPendingWorkSnapshot {
+            queued_messages: inner.queue.len(),
+            delayed_messages: inner.delayed.values().map(Vec::len).sum(),
+            mailbox_messages: inner.mailboxes.values().map(Vec::len).sum(),
+        }
```
erskingardner, line 257: `pending_work_snapshot()` takes no `ClientId` and sums queue/delayed/mailbox counts across **every** attached client on the bus, not just the client being observed. `mailboxes` is already keyed by `ClientId` (see `mailbox()` a few lines up), so `mailbox_messages` in particular could be scoped to the requesting client but isn't.

The effect shows up in `client.rs`'s `pending_work_observation()`, which calls this with no client argument and stores the global totals directly into that client's own `PendingWorkObservation`. Since `TraceExpectation::NoPendingWork` and the docs (`SCENARIOS.md`/README `## Exact state oracle` section) describe this as "every named client's exact progress snapshot," a `NoPendingWork` check on client A can fail purely because client B (elsewhere in the same simulated group) still has an unread mailbox message or a queued send — even though A itself has nothing outstanding. That's the opposite of what a per-client quiescence oracle should report, and it will get worse as scenarios grow to the 20+ member storms this PR's own generated families exercise.

Suggest threading the observing `ClientId` through and at least scoping `mailbox_messages` to `mailboxes.get(&client)`; `queued_messages`/`delayed_messages` may need a documented "this is bus-wide, not per-client" caveat if they truly can't be attributed to one recipient before delivery.

crates/cgka-conformance-simulator/src/scenario_input_ledger.rs
```diff
@@ -0,0 +1,516 @@
...
+        entry.disposition = ScenarioInputDisposition::Pending;
+        entry.pending = true;
+    }
+
+    pub(crate) fn record_published(&mut self, metadata: &ScenarioInputMetadata) {
+        self.remember_metadata(metadata);
+        let entry = self.entry(metadata);
+        entry.published += 1;
+        if metadata.kind == ScenarioInputKind::Application {
+            entry.disposition = ScenarioInputDisposition::Accepted;
```
erskingardner, line 130: For `ScenarioInputKind::Application`, `record_published` marks the entry `Accepted` / `pending = false` as soon as the sender hands the message to the bus — before any recipient has ingested, buffered, or delivered it. Combined with the `bus.rs` scoping issue above: once a message is dropped from the bus entirely (`ScenarioStep::DropQueued`) before reaching any mailbox, there is no ledger entry on the recipient side (created only in `ingest_pending`/`record_ingest`) and no bus counter tracking it either, so a permanently-lost application message becomes invisible to `NoPendingWork` on *every* observed client.

None of this PR's own tests currently combine a dropped application send with a `NoPendingWork`/strict-campaign assertion, so nothing here is silently masked today. But given the whole point of this milestone is a trustworthy oracle that can't be fooled, it's worth either (a) not treating `Accepted` as a non-pending terminal state for the sender's own outbound application entry until delivery/expiry/invalidation is observed somewhere, or (b) explicitly documenting that `NoPendingWork` only covers locally-retained work and cannot detect application messages lost in transit (a distinct claim from what the current docstrings imply).

erskingardner (commented):
Milestone 1.1 (exact canonical oracle + scenario-input ledger + pending-work snapshot + bidirectional decryptability probe) is the right next slice, and the practice of keeping known strict-campaign reds as pinned regressions instead of relaxing the oracle is exactly right. The feature-gated `conformance_snapshot.rs`/`engine.rs` additions are cleanly scoped (compile-time domain-length assertion, exporter secret dropped promptly, `#[cfg(feature = "test-conformance-snapshot")]` everywhere, disabled by default) and the doc/SCENARIOS updates are honest about the three real reliability findings this stricter boundary exposes.

However, the `NoPendingWork` oracle itself — the centerpiece of this PR — has a scoping bug that undermines its "strict per-client" framing, plus a related design gap around when an application send is considered settled. See inline comments.

Non-blocking notes:
- `capture_engine_events` in `client.rs` labels `ForkRecovered`, `CommitRolledBack`, and `GroupStateInvalidated` all as the single string `"superseded"` in the ledger's `invalidated` list, losing the ability to tell these three causes apart in a failure report.
- `crates/cgka-conformance-simulator/Cargo.toml` now requires `cgka-engine`'s `test-conformance-snapshot` feature unconditionally, and `incident-replay` depends on `cgka-conformance-simulator` as a normal (non-dev) dependency. Worth double-checking that no whole-workspace build (`cargo build --workspace`, `just fast-ci`/`just ci`) ever gets unified into an actual production/app build target, given the crate doc's explicit "must not be enabled by production/app consumers" invariant.

PR #1191, opened 2026-07-30
Fix SQLite transaction boundary busy handling

### LINKED ISSUE #1179: storage-sqlite: COMMIT SQLITE_BUSY is downgraded to fatal Backend and bypasses retry discipline
## Summary

`SharedConnection::with_transaction` preserves `SQLITE_BUSY`/`SQLITE_LOCKED` while acquiring `BEGIN IMMEDIATE`, but converts the same transient condition at `COMMIT` or `ROLLBACK` into `StorageError::Backend`.

This breaks the crate-wide busy-retry contract: an outer `retry_on_busy` sees a fatal backend error and returns immediately instead of retrying the complete transaction. A commit-producing operation can therefore fail under transient reader contention and defer frontier progress to a later user/runtime retry.

## Root cause

`crates/storage-sqlite/src/connection.rs:252-260`:

```rust
conn.execute_batch(sql)
    .map_err(|e| StorageError::Backend(format!("sqlite transaction {sql}: {e}")))
```

By contrast, `begin_immediate_with_retry` at `:262-280` uses `map_sqlite_error`, which retains `StorageError::Busy`.

The failed `COMMIT` path at `:192-205` then rolls back and returns the misclassified fatal error.

## Deterministic reproduction

A scratch regression test used the supported `journal_mode=DELETE` configuration:

1. Open two connections to one database with zero busy timeout.
2. Hold a read transaction on connection A.
3. On connection B, `BEGIN IMMEDIATE` and the write succeed.
4. `COMMIT` needs an exclusive lock and returns `database is locked`.
5. Assert the storage result remains `StorageError::Busy`.

Current result:

```text
COMMIT lock contention must remain retryable Busy,
got Err(Backend("sqlite transaction COMMIT: database is locked"))
```

The focused test failed deterministically in `0.07s`. Existing `storage-sqlite` connection tests otherwise passed `21/21` in the independent audit.

## Impact

- Transient commit contention is reported as a fatal backend failure.
- Complete-operation retry wrappers do not retry it.
- Send, publish-confirm, convergence, and projection transactions may fail even though rollback succeeds and a retry would be safe.
- The failure is easiest to reproduce with supported DELETE journaling; the error mapping is incorrect for every SQLite mode.

This is not database corruption. Durable state rolls back; liveness resumes only if an outer workflow happens to retry later.

## Required behavior

- Map transaction-boundary SQLite errors through the same `SQLITE_BUSY`/`SQLITE_LOCKED` classifier used elsewhere.
- Preserve transaction ownership until commit retry or rollback has completed.
- Either retry a busy `COMMIT` in place with a bounded policy, or roll back and return `StorageError::Busy` so a documented complete-operation retry can rerun safely.
- Do not blindly rerun arbitrary transaction closures unless their full operation is explicitly safe/idempotent.
- Keep failed-commit rollback and unusable-connection handling intact for fatal errors.

## Regression coverage

- DELETE-journal reader blocks COMMIT, then releases: operation eventually succeeds or returns retryable `Busy` according to the selected policy.
- Persistent COMMIT contention exhausts the budget and remains `Busy`, never `Backend`.
- Fatal deferred-constraint COMMIT failures still roll back and remain fatal.
- ROLLBACK boundary errors retain correct transient/fatal classification and owner cleanup.

## Related

- #1028 tracks SQLite transaction and contention discipline.
- #508 and #453 covered earlier lock surfaces, not transaction-boundary classification.

erskingardner:
## Summary

- classify `SQLITE_BUSY` and `SQLITE_LOCKED` at `COMMIT` and `ROLLBACK` as transient storage errors
- retry transaction boundaries in place with bounded backoff while retaining transaction ownership
- preserve fatal commit rollback and unusable-connection handling without rerunning arbitrary transaction closures
- add deterministic DELETE-journal coverage for transient and persistent commit contention, fatal commit failure, rollback classification, ownership, and cleanup

## Root cause

`SharedConnection::execute_transaction_boundary` converted every SQLite error at `COMMIT` or `ROLLBACK` into `StorageError::Backend`. That bypassed the storage layer's busy classification and caused brief reader contention at commit time to appear fatal.

## Impact

Transactions now tolerate transient contention at their final boundary without repeating their application closure. Persistent contention remains a retryable `Busy` after the bounded budget and is rolled back before the connection is released.

## Validation

- `cargo test -p storage-sqlite` (300 passed)
- `just fast-ci`

Closes #1179

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when committing or rolling back SQLite transactions during temporary database contention.
  * Transient busy or locked conditions are retried automatically without rerunning the transaction.
  * Exhausted contention now reports a clear busy error with transaction-boundary context.
  * Fatal commit failures continue to surface as backend errors without unnecessary retries.

### DIFF
--- a/crates/storage-sqlite/src/connection.rs
+++ b/crates/storage-sqlite/src/connection.rs
@@ -45,13 +45,14 @@ impl TransientError for StorageError {
     }
 }
 
-/// Run a whole storage operation, retrying with capped exponential backoff while
-/// it fails with transient lock contention. `op` MUST be a complete, idempotent
-/// unit of work (a single autocommit statement or an entire `BEGIN..COMMIT`
-/// transaction) — never a single statement inside a larger transaction, since a
-/// retry re-runs the closure from the top. On `SQLITE_BUSY`/`SQLITE_LOCKED`
-/// SQLite has already rolled back the failed transaction, so re-running is safe.
-/// After [`BUSY_MAX_ATTEMPTS`] the last transient error is returned unchanged so
+/// Retry an operation with capped exponential backoff while it fails with
+/// transient lock contention. `op` MUST be safe to invoke again. Most callers
+/// pass a complete, idempotent storage operation (a single autocommit statement
+/// or an entire transaction), never one statement from the middle of a larger
+/// transaction. Transaction-boundary callers are the deliberate exception:
+/// they retry only `COMMIT` or `ROLLBACK` in place while retaining transaction
+/// ownership, without re-running the transaction closure. After
+/// [`BUSY_MAX_ATTEMPTS`] the last transient error is returned unchanged so
 /// callers still see a `Busy` (transient) classification.
 pub(crate) fn retry_on_busy<T, E, F>(mut op: F) -> Result<T, E>
 where
@@ -189,12 +190,12 @@ impl SharedConnection {
 
         let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(f));
         match result {
-            Ok(Ok(value)) => match self.execute_transaction_boundary("COMMIT") {
+            Ok(Ok(value)) => match self.execute_transaction_boundary_with_retry("COMMIT") {
                 Ok(()) => {
                     self.clear_transaction_owner();
                     Ok(value)
                 }
-                Err(commit_err) => match self.execute_transaction_boundary("ROLLBACK") {
+                Err(commit_err) => match self.execute_transaction_boundary_with_retry("ROLLBACK") {
                     Ok(()) => {
                         self.clear_transaction_owner();
                         Err(E::from(commit_err))
@@ -204,7 +205,7 @@ impl SharedConnection {
                     )))),
                 },
             },
-            Ok(Err(err)) => match self.execute_transaction_boundary("ROLLBACK") {
+            Ok(Err(err)) => match self.execute_transaction_boundary_with_retry("ROLLBACK") {
                 Ok(()) => {
                     self.clear_transaction_owner();
                     Err(err)
@@ -213,7 +214,7 @@ impl SharedConnection {
                     "sqlite transaction ROLLBACK failed after callback error ({rollback_err}); connection marked unusable",
                 )))),
             },
-            Err(payload) => match self.execute_transaction_boundary("ROLLBACK") {
+            Err(payload) => match self.execute_transaction_boundary_with_retry("ROLLBACK") {
                 Ok(()) => {
                     self.clear_transaction_owner();
                     std::panic::resume_unwind(payload);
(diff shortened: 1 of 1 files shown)


erskingardner (commented):
Reviewed against the linked root-cause writeup (#1179). The fix is exactly the shape the issue asked for: `COMMIT`/`ROLLBACK` boundary errors now flow through `map_sqlite_error` instead of being force-mapped to fatal `Backend`, the boundary is retried in place (not the transaction closure) while ownership is held, and exhausted retries still roll back before returning `Busy`. The four required regression scenarios (transient commit contention, persistent commit contention, fatal commit failure, rollback classification) are all covered with deterministic DELETE-journal tests rather than sleep-based approximations.

Traced through `with_transaction`'s commit/rollback/panic branches and the shared `retry_on_busy` budget (`BUSY_MAX_ATTEMPTS`/backoff, unchanged) — didn't find a correctness issue. Two very minor, non-blocking notes inline; nothing here blocks merge.

crates/storage-sqlite/src/connection.rs
```diff
@@ -323,6 +333,26 @@ impl SharedConnection {
...
+fn retry_transaction_boundary<F>(sql: &str, mut execute: F) -> StorageResult<()>
+where
+    F: FnMut() -> rusqlite::Result<()>,
+{
+    retry_on_busy(|| {
+        execute().map_err(|error| {
+            let detail = error.to_string();
+            match crate::codec::map_sqlite_error(error) {
+                StorageError::Busy(_) => {
+                    StorageError::Busy(format!("sqlite transaction {sql}: {detail}"))
```
erskingardner: Minor: `map_sqlite_error(error)` already returns the formatted `error.to_string()` inside its `Busy(String)`/`Backend(String)` variant, so this recomputes the same string twice (once here via `error.to_string()`, once inside `map_sqlite_error`). Could simplify to destructure the returned variant directly:
```rust
match crate::codec::map_sqlite_error(error) {
    StorageError::Busy(detail) => StorageError::Busy(format!("sqlite transaction {sql}: {detail}")),
    StorageError::Backend(detail) => StorageError::Backend(format!("sqlite transaction {sql}: {detail}")),
    other => unreachable!("SQLite errors map only to Busy or Backend: {other:?}"),
}
```
Not a correctness issue, just avoids the double `to_string()` and the extra local binding.

crates/storage-sqlite/src/connection.rs
```diff
@@ -663,11 +693,39 @@ impl KeyPackageBundleStorage for SqliteAccountStorage {
...
+    fn open_delete_journal_store(path: &Path) -> SqliteAccountStorage {
+        SqliteAccountStorage::open_encrypted_with_options(
+            path,
+            &SqlCipherKey::new("delete journal transaction boundary key").unwrap(),
+            delete_journal_options(),
+        )
+        .unwrap()
+    }
+
+    fn sqlite_failure(primary: std::os::raw::c_int) -> rusqlite::Error {
```
erskingardner, line 719: Minor: this duplicates the `sqlite_failure` test helper already defined in `codec.rs`'s test module (same signature/body). It's private to that module so it can't be imported here directly — not worth blocking on, but if this pattern needs a third home later it'd be worth hoisting into a shared `#[cfg(test)]` test-support module.

PR #1192, opened 2026-07-30
Retire consumed self-remove proposal work

erskingardner:
## What changed

- Mark durable proposal message rows `Processed` when a locally confirmed commit consumed their exact authenticated proposal references.
- Clear the epoch-scoped in-memory proposal scheduling signal after local confirmation, accepted direct commits, and successful settled convergence passes.
- Tighten the send/leave simulator regression and add restart coverage proving consumed proposals do not recreate pending convergence work.

## Why

After a member removed itself, the auto-committer could consume the retained `SelfRemove` proposal while leaving its durable `cgka_messages` row in `Created` and its in-memory `valid_proposal_groups` signal set. The stale row survived restart and could recreate pending convergence work; the stale signal could request redundant scheduling in the current process. This residual work did not represent an actionable proposal after the epoch advanced.

The fix gives locally consumed proposal rows the same terminal disposition used by stored convergence. It matches the staged commit's consumed proposal references against current-epoch public proposal records before merging, within the existing confirmation transaction. It does not replay private handshake messages or mutate the secret tree.

## Impact

Self-removal now completes with no residual durable proposal or scheduler signal. Restart remains quiescent, and strict send/leave convergence checks no longer need a known-failure exception.

## Validation

- `cargo test -p cgka-conformance-simulator send_leave_family_records_seed_and_runs_generated_cases -- --nocapture`
- `just fast-ci`

A plain `cargo test -p cgka-engine` still has seven pre-existing deferred-peel failures because those tests request the test-only policy override without enabling `test-policy-overrides`; the feature-enabled engine suite passes.

### DIFF
--- a/crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
+++ b/crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
@@ -1699,11 +1699,6 @@ async fn strict_chaos_boundary_retains_known_reliability_failures() {
             "rolled-back transported commit",
         ),
         (
-            3usize,
-            "pending_work_remaining",
-            "self-remove proposal work",
-        ),
-        (
             4usize,
             "pending_work_remaining",
             "pre-join deferred application object",
@@ -1723,6 +1718,15 @@ async fn strict_chaos_boundary_retains_known_reliability_failures() {
             report.expectation_failures
         );
     }
+
+    let self_remove_report = run_generated_case_report(&cases[3], None)
+        .await
+        .expect("strict self-remove case reports");
+    assert!(
+        self_remove_report.expectation_failures.is_empty(),
+        "strict self-remove case must retire consumed proposal work: {:?}",
+        self_remove_report.expectation_failures
+    );
 }
 
 #[tokio::test]
--- a/crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
+++ b/crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
@@ -1322,29 +1322,12 @@ async fn send_leave_family_records_seed_and_runs_generated_cases() {
         let report = run_generated_case_report(&case, None)
             .await
             .expect("strict send/leave case report runs");
-        let includes_leave = case
-            .scenario
-            .steps
-            .iter()
-            .any(|step| matches!(step, ScenarioStep::Leave { .. }));
-        if includes_leave {
-            assert!(
-                report
-                    .expectation_failures
-                    .iter()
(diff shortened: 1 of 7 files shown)


erskingardner (commented):
Reviewed against the linked context (#1188/#1190, which introduced the known-failure `pending_work_remaining` exception this PR now retires). GitHub won't let this account self-approve, so posting as a comment — recommendation is **approve / merge**.

**Summary:** Correct, well-scoped fix. Verified the mechanics directly against the pinned OpenMLS revision (`59e7d3b`, `openmls::group::mls_group::processing`):

- `mark_consumed_proposal_records_processed` (openmls_projection.rs) reprocesses only `PublicMessage` proposals via `MlsGroup::process_message`, which for a `Proposal` body just re-derives a `QueuedProposal`/reference from `AuthenticatedContent` — it does not call `store_pending_proposal` and only persists the secret tree for `PrivateMessage` (`will_modify_secret_tree`), which is explicitly excluded. So calling this before `merge_pending_commit`/`merge_staged_commit` on the live `mls_group` is side-effect-free, matching the module's own guard comment. Since this repo pins `PURE_PLAINTEXT_WIRE_FORMAT_POLICY`, the guard is exercised on essentially every proposal today, with correct defense-in-depth if that policy ever changes.
- The `record.epoch != source_epoch` filter is necessary, not redundant: `SqliteAccountStorage::list_messages` is `epoch >= ?` ("at or after"), not exact-match, per its own filter test.
- `consumed_proposal_refs` (from `own_commit_stamp`, `staged.queued_proposals()`) and the new reference computed via `queued.proposal_reference_ref()` both go through `tls_hex`, so the reference comparison is apples-to-apples.
- The three new `valid_proposal_groups.remove(...)` sites are consistent with the signal's only producer (`message_processor/ingest.rs`'s `ProposalMessage` arm) and are placed after the epoch has durably advanced in each of the three commit-acceptance seams (direct ingest, local confirm, settled convergence) — correctly source-epoch scoped per the inline comments.
- `crates/cgka-session` already drains this signal after every single `ingest`/`ingest_delivery` call, so the residual-signal bug was really only reachable through direct `Engine` use (the simulator, and the new engine-level regression test) or multi-message batches — consistent with the PR's description and repro path.
- The terminal state used (`MessageState::Processed`) matches what stored convergence already assigns elsewhere in `openmls_projection.rs`, so this genuinely unifies the two commit paths as claimed.

Test changes (new engine-level restart assertion, tightened simulator regression) directly cover the fix. No CHANGELOG entry needed (only `crates/cli` carries one; this PR doesn't touch it).

No blocking issues found.

PR #1194, opened 2026-07-30
Add convergence subject boundary

erskingardner:
## Summary

- add a simulator-owned, capability-declared `ConvergenceSubject` interface and an in-process `EngineHarnessSubject` adapter
- keep scenario semantics and stable action scheduling in the runner while moving engine, transport, observation, restart, and pending-publication mechanics behind the adapter
- separate queue and partition mutation behind the explicitly white-box `ConvergenceFaultSubject` interface
- reject unsupported scenarios before any scenario action runs and record adapter identity/capabilities in every report
- update the convergence reliability plan to mark the 1.2a vertical slice complete while leaving virtual time, outbound polling/acknowledgement, and the canonicalization-contract refresh open

## Why

Milestone 1.2 needs one scenario corpus to run against increasingly production-shaped implementations without duplicating or reinterpreting scenario semantics in every adapter. This establishes that boundary around the existing engine harness while preserving current vector and trace behavior.

## Verification

- `cargo test -p cgka-conformance-simulator scenario::tests --lib`
- `cargo test -p cgka-conformance-simulator --test canonical_scenarios canonical_vector_fixtures_match_generated_traces`
- `cargo clippy -p cgka-conformance-simulator --all-targets -- -D warnings`
- `just fast-ci`

### DIFF
--- a/crates/cgka-conformance-simulator/src/scenario.rs
+++ b/crates/cgka-conformance-simulator/src/scenario.rs
@@ -5,21 +5,16 @@
 //! their observed trace to exact or semantic fixture expectations.
 
 use crate::{
-    BidirectionalDecryptabilityObservation, ClientBuilder, DecryptabilityProbeSendStatus,
-    DirectionalDecryptabilityProbe, ExpectationFailure, HarnessClient, HarnessStorageMode,
-    PendingResolutionObservation, ScenarioAdminPolicyObservation, ScenarioErrorObservation,
-    ScenarioOracleReport, ScenarioTrace, TraceExpectation, TransportBus, VectorFixture,
-    build_scenario_oracle_report, compare_trace_expectations, observe_client, observe_client_exact,
+    ConvergenceFaultSubject, ConvergenceSubject, EngineHarnessSubject, ExpectationFailure,
+    HarnessStorageMode, PendingResolutionObservation, ScenarioErrorObservation,
+    ScenarioOracleReport, ScenarioTrace, SubjectCreateGroup, SubjectDescriptor, SubjectError,
+    SubjectInviteMembers, SubjectSendApplication, SubjectUpdateAdminPolicy, SubjectUpdateGroupData,
+    TraceExpectation, VectorFixture, build_scenario_oracle_report, compare_trace_expectations,
+    required_capability,
 };
-use cgka_engine::feature_registry::FeatureRegistry;
-use cgka_traits::EngineError;
-use cgka_traits::capabilities::{Capability, CapabilityRequirement, Feature, RequirementLevel};
-use cgka_traits::engine::KeyPackage;
-use cgka_traits::engine_state::PendingStateRef;
 use cgka_traits::group::ProtocolProfile;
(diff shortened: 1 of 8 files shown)


crates/cgka-conformance-simulator/src/scenario.rs
```diff
@@ -284,15 +292,10 @@ pub async fn run_scenario_report_with_storage_mode(
...
-        vec![],
-        None,
-        ProtocolProfile::Legacy,
-        storage_mode,
-    )
-    .await
+    let mut subject =
+        EngineHarnessSubject::new(&spec.clients, ProtocolProfile::Legacy, storage_mode)
+            .map_err(subject_setup_error)?;
+    run_scenario_report_inner(spec, expected_trace, vec![], None, &mut subject).await
```
erskingardner, line 298: Nit (non-blocking): in the built-in helpers (`run_scenario_report_with_storage_mode`, `run_scenario_report_with_outcomes_and_storage_mode`, `run_vector_fixture_report_with_storage_mode`), `EngineHarnessSubject::new` now constructs the full in-process client set (OpenMLS/SQLite, attached to the transport bus) *before* `run_scenario_report_inner` calls `validate_scenario_for_subject`, which is where the `spec.spec_version != "1"` check now lives. Previously that version check was the very first thing that ran, before any client/storage setup. Now a scenario with an unsupported spec version pays for full engine/storage construction before being rejected. Since `validate_scenario_for_subject" needs the subject's descriptor only for the per-step capability check, the cheap version/duplicate-label checks could run before constructing the subject (e.g. hoisted into a helper called prior to `EngineHarnessSubject::new`). Low impact given this is a test simulator, but it undoes part of the original fail-fast ordering.

erskingardner (commented):
Reviewed against the milestone 1.2 plan in `docs/marmot-architecture/convergence-reliability-plan.md`. This is a clean, well-scoped extraction: `scenario.rs` now owns only step semantics/ordering, `subject.rs` introduces the capability-declared `ConvergenceSubject`/`ConvergenceFaultSubject` boundary and the in-process `EngineHarnessSubject` adapter, and the queue/partition fault mutation is correctly split out as an explicit white-box capability. Preflight capability validation runs before any step executes, matching the PR's stated goal, and there's a good regression test (`explicit_engine_subject_preserves_default_runner_trace`) proving the new path produces identical traces to the pre-refactor runner. The plan doc checkboxes accurately reflect what shipped vs. what's intentionally deferred (virtual time, outbound polling, canonicalization refresh).

(Note: GitHub won't let me formally approve my own PR account here, so filing this as a comment review — content is a clean bill with one non-blocking nit.)

Only a minor, non-blocking note left inline: the built-in helper entry points now construct the full `EngineHarnessSubject` (real OpenMLS/SQLite clients attached to the bus) before the cheap `spec_version`/duplicate-client checks in `validate_scenario_for_subject` run, whereas previously those checks were the very first thing executed. Low impact for a test simulator, but worth hoisting if easy.

No blocking issues found — recommend merge.

PR #1195, opened 2026-07-30
Add injectable convergence dual clock

erskingardner:
## Summary

- add a convergence-only dual clock with system and manually advanced implementations
- route convergence deadlines, pass timestamps, cutoff decisions, and restart rebasing through paired monotonic and wall-clock samples
- remove real-time waits from clock-sensitive convergence tests, including terminal disband tombstone coverage

## Scope

This is only the Milestone 1.3 dual-clock prerequisite. It does not add `ConvergenceSubject`, structural quiescence, or Scenario IR. Production policy constants and unrelated identity/event timestamps are unchanged.

## Validation

- `just fast-ci`

### DIFF
--- a/crates/cgka-conformance-simulator/src/client.rs
+++ b/crates/cgka-conformance-simulator/src/client.rs
@@ -14,7 +14,7 @@ use cgka_engine::account_identity_proof::{
 };
 use cgka_engine::canonicalization::{CanonicalizationPolicy, CanonicalizationResult};
 use cgka_engine::feature_registry::FeatureRegistry;
-use cgka_engine::{Engine, EngineBuilder};
+use cgka_engine::{ConvergenceClock, Engine, EngineBuilder};
 use cgka_traits::app_components::{
     AppComponentData, GROUP_ADMIN_POLICY_COMPONENT_ID, NOSTR_ROUTING_COMPONENT_ID, NostrRoutingV1,
     default_group_components, encode_nostr_routing_v1,
@@ -56,6 +56,7 @@ pub struct HarnessClient {
     signer: nostr::Keys,
     registry: FeatureRegistry,
     protocol_profile: ProtocolProfile,
+    convergence_clock: Option<Arc<dyn ConvergenceClock>>,
     pending_events: Vec<GroupEvent>,
     /// Default MLS group id used by single-group scenarios. Set
     /// automatically after the first create/join.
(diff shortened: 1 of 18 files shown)


erskingardner (commented):
Recommend merge — reviewed the diff (2418 lines, mostly mechanical test call-site updates). This cleanly threads an injectable dual clock (`ConvergenceClock` / `ConvergenceTime`) through convergence scheduling, replacing the ad-hoc `Instant`-based monotonic clock and separate `WallClock` usage in `distributed_convergence.rs` with one paired source. (Posting as a comment, not an approval, since GitHub doesn't allow self-approval.)

Verified:
- `SystemConvergenceClock` / `ManualConvergenceClock` design in `crates/cgka-engine/src/convergence_clock.rs` is simple and correct (poison-safe mutex, saturating arithmetic).
- The `_at` / `_with_time` split (`buffer_openmls_convergence_message` vs `buffer_openmls_convergence_message_at`, same for `converge_stored_openmls_messages`) preserves exact prior behavior for existing tests that don't inject a clock — `wall_ms` still comes from the real system clock unless a `ManualConvergenceClock` is installed, and the wall/monotonic rebase path (`load_or_open_convergence_pass`) is only exercised on an engine rebuild (`clock_instance_id` mismatch), so mixing an arbitrary literal `monotonic_ms` with real wall time in the non-manual-clock tests is not a behavior change.
- The `clock.advance_ms(1_000)` replacing the polling sleep loops in `canonical_scenarios.rs` and `group_creation.rs` lines up with `V1_SETTLEMENT_QUIESCENCE_MS = 1_000`, so the hardcoded tick counts are correct, not guessed.
- New test coverage added (explicit pass-timestamp assertions in `engine_duplicate_convergence_input_does_not_reset_quiescence`, and a new forward-wall-jump-makes-cutoff-immediate case) is a genuine improvement, not just a refactor.
- No engine-generic-over-`S`, transport-boundary, or logging-privacy invariants are touched; scope matches the stated "Milestone 1.3 dual-clock prerequisite" — no `ConvergenceSubject`/Scenario IR creep.

Minor, non-blocking observations:
- `advance_convergence_inputs_until_settled` (message_processor/mod.rs) still takes a raw `now_ms: u64` and internally synthesizes `ConvergenceTime` by borrowing `wall_ms` from the installed clock — same pattern as the `_at` helpers but without that naming convention or `#[doc(hidden)]`. Not a bug, just a naming inconsistency if you want the `_at` convention to be exhaustive.
- `build_client_with_storage_and_clock` in `tests/distributed_convergence.rs` drops the `FixedMaintenanceRandom` (previously wired via `maintenance_sources`) in favor of the default `OsMaintenanceRandom` when only `convergence_clock` is set. That random source only feeds join jitter sampling (`group_lifecycle.rs`), unrelated to what these tests assert, so it looks safe — just flagging since it's a quiet behavior change to the test harness's determinism guarantees beyond the clock.

Nothing here blocks merge.

PR #1196, opened 2026-07-30
Fix simulator rollback publication semantics

erskingardner:
## Summary

- retract every undelivered artifact belonging to a simulator publication before applying `FailPending`
- reject definite-failure rollback after any commit or Welcome has reached a recipient, preserving the pending state for confirmation
- make the transported-commit-after-local-rollback strict-oracle case deterministic and green
- document the simulator fault boundary and retain the two unrelated engine-state chaos failures

## Root cause

The conformance simulator modeled `FailPending` as a local engine rollback only. The associated commit remained queued (or delayed) on the simulated transport, so a peer could subsequently ingest a commit that the sender had already rolled back. Production publication lifecycle code already distinguishes acknowledged acceptance, definite zero-acceptance failure, and ambiguous exposure; the failure was in simulator fault semantics rather than the engine publish lifecycle.

## Protocol implications

A definite publication failure may roll back only when the harness can establish that no publication artifact reached a recipient. Once exposure is possible, treating publication as definitely failed is unsafe: peers may already have advanced, so the sender must retain pending state for confirmation/reconciliation.

## Validation

- `git diff --check`

No production engine, transport-adapter, or account-runtime behavior is changed.

### DIFF
--- a/crates/cgka-conformance-simulator/AGENTS.md
+++ b/crates/cgka-conformance-simulator/AGENTS.md
@@ -130,6 +130,11 @@ to be a group member yet.
 `drop_queued`, `duplicate_queued`, `delay_queued`, `release_delayed`, and `reorder_queued`. `set_partition` and
 `clear_partition` remain the partition/heal operations.
 
+`fail_pending` means definite non-publication. Before invoking the engine rollback, the harness retracts every matching
+undelivered commit and Welcome from the queue and delayed sets. It returns a scenario error instead of rolling back if
+any matching artifact has already reached a recipient mailbox; model that case as ambiguous exposure at an adapter-aware
+boundary rather than as `fail_pending`.
+
 Use `clear_events` after setup when a scenario wants the final trace to describe only the behavior under test. The
 convergence E2E scenario does this after the initial welcome joins so its trace focuses on the peeler-ingest epoch
 change and selected branch outputs. Delayed past-epoch app messages are covered via retained epoch contexts.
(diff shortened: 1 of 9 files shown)


erskingardner (commented):
## Review

Simulator-only change, scoped and well-tested. Recommend merge — no blocking issues found (posting as comment since I can't approve your own PR).

- `TransportBus::retract_undelivered_publication` (bus.rs) checks `exposed_recipient_counts` and mutates the queue/delayed sets under a single mutex critical section, so the "no partial retraction of an already-exposed publication" precondition is genuinely atomic — no TOCTOU gap.
- `exposed_recipient_counts` is incremented exactly where messages become observable (mailbox push in `step()`, and `inject()`), and checked by summing across all of a pending's tracked artifact ids, so partial exposure (e.g. Welcome delivered, commit still queued) correctly blocks the rollback.
- Verified every scenario-reachable pending-producing path (`CreateGroup`/`GroupCreated`, `InviteMembers`, `UpdateGroupData`, `UpdateAdminPolicy`) now calls `remember_pending_publication` before the corresponding `bus.send`, so `try_fail`'s artifact set is never silently empty for these paths. The auto-publish/convergence-driven commit paths (`publish_send_result`, `drain_auto_publish_confirm`) self-confirm immediately and are never targets of a later `FailPending` scenario step, so they don't need tracking.
- New tests cover both the happy path (retract commit, retract commit+delayed Welcome) and the guard rejecting `fail_pending` after mailbox exposure, plus the regenerated chaos-family regression and the "strict boundary" test asserting the previously-red transported-commit case is now green while the two unrelated engine-state failures are explicitly retained.
- Docs (AGENTS.md, README.md, SCENARIOS.md, convergence-reliability-plan.md) are updated consistently with the new `fail_pending` semantics.

No production engine/transport/account-runtime code is touched, matching the PR's stated scope. Nice minimal, root-cause fix.

PR #1200, opened 2026-07-30
Enforce per-account session ownership

### LINKED ISSUE #998: Security review L4: MarmotApp::client() opens a second AccountDeviceSession over the same DB with no exclusivity guard
## Summary

Nothing prevents two live `AccountDeviceSession`s (two `Engine` instances) over one `session.sqlite`: a long-running account worker's client plus a direct `app.client(label)` call. Storage is shared and safe, but each engine keeps private in-memory pending/dedup/event state; commits created through both sessions can stage the same next epoch → a self-inflicted fork. Current call sites only open a second client when no worker runs, so this is a latent trap for future callers rather than a live bug.

## Location

`crates/marmot-app/src/lib.rs:1027-1100` (`client`/`client_with_relay_plane` → `open_account`); called at `crates/marmot-app/src/runtime/mod.rs:3462`

## Impact

A future UniFFI/CLI path opens a direct client for an account whose worker is running; a send from each session in the same epoch forks the group; cross-session publish/confirm interference produces confusing rollbacks.

## Suggested remediation

Track open sessions per label in `MarmotApp` (like `account_storages`) and refuse (or share) when a worker already owns one; document the exclusivity contract on `client()`.

---
Found in the 2026-07-20 security review (`docs/security-review-2026-07-20.md`, finding L4). Status: **suspected.**

erskingardner:
## Summary

- add an in-process per-account ownership guard so a `MarmotApp` and its clones cannot hydrate two independent engine sessions over the same account database
- hold ownership for the `AppClient` lifetime and return typed `AccountSessionBusy` contention through MarmotKit
- drop a failed worker client before reconnect opens its replacement
- route runtime integration tests through the managed worker and document the change in the Unreleased changelog

## Root cause

The root runtime lease prevents separate processes or independently constructed production runtimes from sharing an app root, but it does not distinguish multiple `AccountDeviceSession` instances created inside one `MarmotApp`. Each session owns private in-memory pending, deduplication, and epoch state, so two sessions over one `session.sqlite` could stage conflicting next-epoch work even though the database itself remained transaction-safe.

This change adds only an in-process ownership guard: no additional file lock, OS lease, or lifetime-held mutex. `MarmotApp` clones share a small account-owner registry, while each `AppClient` holds one non-clonable guard that unregisters on drop.

## Validation

- `just fast-ci`
- focused account ownership, worker lifecycle, FFI mapping, audit-log reopen, and affected relay-runtime tests

The full `cargo test -p marmot-app` command also reports five relay-runtime failures outside this change. `encrypted_media_endpoint_updates_are_full_replacement_and_admin_only` was reproduced unchanged on base commit `1985d9a8`.

Closes #998

### DIFF
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -67,6 +67,8 @@ pub(crate) use sync::is_own_relay_echo;
 pub struct AppClient {
     pub(crate) app: MarmotApp,
     pub(crate) runtime: AppRuntime,
+    // Struct fields drop in declaration order. Keep the engine-owning runtime
+    // before this guard so ownership is released only after engine teardown.
     pub(crate) _session_guard: crate::AppAccountSessionGuard,
     pub(crate) adapter: MarmotRelayPlaneAccountAdapter,
     pub(crate) routing: AppTransportRouting,
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -67,6 +67,7 @@ pub(crate) use sync::is_own_relay_echo;
 pub struct AppClient {
     pub(crate) app: MarmotApp,
     pub(crate) runtime: AppRuntime,
+    pub(crate) _session_guard: crate::AppAccountSessionGuard,
     pub(crate) adapter: MarmotRelayPlaneAccountAdapter,
     pub(crate) routing: AppTransportRouting,
     pub(crate) relay_plane: MarmotRelayPlane,
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -440,6 +440,7 @@ pub struct MarmotApp {
     config: MarmotAppConfig,
     directory_sync: Arc<RwLock<Option<DirectorySyncHandle>>>,
     account_storages: Arc<Mutex<HashMap<String, SqliteAccountStorage>>>,
+    account_session_owners: Arc<Mutex<HashSet<String>>>,
     directory_caches: Arc<Mutex<HashMap<String, DirectoryCache>>>,
     legacy_directory_cache_checked: Arc<Mutex<bool>>,
     #[cfg(test)]
@@ -971,10 +972,25 @@ struct KeyPackageRecord {
 
 struct OpenAppAccount {
     runtime: AppRuntime,
+    session_guard: AppAccountSessionGuard,
     adapter: MarmotRelayPlaneAccountAdapter,
     routing: AppTransportRouting,
     state: AccountState,
     signer: Arc<dyn nostr::NostrSigner>,
+}
+
+struct AppAccountSessionGuard {
+    label: String,
+    owners: Arc<Mutex<HashSet<String>>>,
+}
+
+impl Drop for AppAccountSessionGuard {
+    fn drop(&mut self) {
+        self.owners
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner())
+            .remove(&self.label);
+    }
 }
 
 impl MarmotApp {
@@ -1109,6 +1125,7 @@ impl MarmotApp {
             config,
             directory_sync: Arc::new(RwLock::new(None)),
             account_storages: Arc::new(Mutex::new(HashMap::new())),
+            account_session_owners: Arc::new(Mutex::new(HashSet::new())),
             directory_caches: Arc::new(Mutex::new(HashMap::new())),
             legacy_directory_cache_checked: Arc::new(Mutex::new(false)),
             #[cfg(test)]
@@ -1169,6 +1186,7 @@ impl MarmotApp {
             config,
             directory_sync: Arc::new(RwLock::new(None)),
             account_storages: Arc::new(Mutex::new(HashMap::new())),
+            account_session_owners: Arc::new(Mutex::new(HashSet::new())),
             directory_caches: Arc::new(Mutex::new(HashMap::new())),
             legacy_directory_cache_checked: Arc::new(Mutex::new(false)),
             #[cfg(test)]
@@ -1217,6 +1235,12 @@ impl MarmotApp {
             .contains_key(label)
     }
 
+    /// Open the account's exclusive in-memory engine session.
+    ///
+    /// Only one [`AppClient`] for an account may exist within a [`MarmotApp`]
+    /// (including its clones and managed runtime workers). A concurrent open
+    /// returns [`AppError::AccountSessionBusy`]. Drop the owning client before
+    /// retrying.
     pub async fn client(&self, label: &str) -> Result<AppClient, AppError> {
         #[cfg(test)]
         let relay_plane = self
@@ -1289,6 +1313,7 @@ impl MarmotApp {
         let mut client = AppClient {
             app: self.clone(),
             runtime: open.runtime,
+            _session_guard: open.session_guard,
             adapter: open.adapter,
             routing: open.routing,
             relay_plane: relay_plane.clone(),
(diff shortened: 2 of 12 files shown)


crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -795,92 +795,103 @@ async fn run_app_runtime_account_worker(
...
-                                        ),
-                                    );
-                                    tokio::select! {
+                        // The account-session ownership guard is held by
+                        // `AppClient`.
+                        // Destroy the failed engine before attempting to
+                        // hydrate its replacement; otherwise reconnect would
+                        // create two independent engines over one session DB.
+                        drop(client);
+                        client = loop {
```
erskingardner, line 802: Behavior change beyond the stated scope, and I do not think it is intended.

Before this PR, a failed reopen fell out of the `match` and back to the **outer** command loop, which still held the old `client` and kept serving commands (reads, sends against local state) between retries. Now `drop(client)` happens first and the retry lives in an unbounded inner `loop`, so for as long as `runtime_local_client` keeps failing the worker serves **no** commands at all. The command channel is `mpsc::channel(8)`, so callers back up and then hit send failures / `APP_RUNTIME_ACCOUNT_READY_WAIT` timeouts.

Dropping the failed engine before hydrating its replacement is correct for the guard, but the two concerns can be separated: drop the client, then return to the outer loop and let it re-enter the reconnect branch (it already does, via the receive error path), so commands keep getting polled between attempts.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -795,92 +795,103 @@ async fn run_app_runtime_account_worker(
...
+                                                "runtime restart transport failed",
+                                                &transport_err,
                                             ),
                                         );
+                                        tokio::select! {
+                                            _ = wait_for_runtime_shutdown(&mut lifecycle_shutdown) => return,
+                                            _ = &mut shutdown => return,
+                                            _ = sleep(reconnect_backoff.next_delay()) => {}
+                                        }
+                                        drop(reopened);
```
erskingardner, line 852: Nit, but it undermines the guard's usefulness: the backoff `sleep` runs *before* `drop(reopened)`, so the guard is held for the whole backoff window even though the session is already known dead. Same shape at the outer `drop(client)` — it happens after that branch's backoff sleep. Dropping before sleeping means a concurrent one-shot `app.client()` is not rejected for the duration of the backoff.

crates/marmot-app/src/lib.rs
```diff
@@ -2672,13 +2698,28 @@ impl MarmotApp {
...
             runtime,
+            session_guard,
             adapter,
             routing,
             state,
             signer: nostr_signer,
         })
     }
 
+    fn acquire_account_session(&self, label: &str) -> Result<AppAccountSessionGuard, AppError> {
```
erskingardner, line 2714: Two notes on the key:

1. `MarmotApp::client()` accepts any account ref, not just a label — `account_home().account()` resolves npub/hex refs too (`crates/marmot-account/src/home.rs:393`). Keying the owner set on the caller-supplied string means two refs for the same account are two different keys and the exclusivity guarantee silently does not hold. `account_storages` has the same weakness so this is pre-existing, but since this guard is the thing advertising exclusivity, resolving to the canonical label before keying would make the guarantee real rather than conventional.

2. Consider a `Drop`-independent release hook (or exposing the guard to the runtime) — see the review body; the whole blocking issue is that release is only observable via task drop, which the abort-based teardown paths never wait for.

crates/marmot-app/src/client/mod.rs
```diff
@@ -67,6 +67,7 @@ pub(crate) use sync::is_own_relay_echo;
 pub struct AppClient {
     pub(crate) app: MarmotApp,
     pub(crate) runtime: AppRuntime,
+    pub(crate) _session_guard: crate::AppAccountSessionGuard,
```
erskingardner, line 72: The position of this field is load-bearing: struct fields drop in declaration order, so `runtime` (the `AccountDeviceRuntime` owning the engine/session) must be destroyed *before* `_session_guard` releases ownership. It is correct as written, but a one-line comment saying so would stop a future reorder or a `#[derive]`-driven shuffle from reintroducing a window where the guard is free while the engine is still alive.

crates/marmot-uniffi/src/errors.rs
```diff
@@ -209,6 +213,7 @@ impl From<AppError> for MarmotKitError {
             AppError::FollowListUnavailable => Self::FollowListUnavailable,
             AppError::TransportClosed => Self::TransportClosed,
             AppError::RuntimeBusy => Self::RuntimeBusy,
+            AppError::AccountSessionBusy => Self::AccountSessionBusy,
```
erskingardner, line 216: Worth noting the typed variant only survives the *direct* `app.client()` path. On the managed-worker path the same condition is stringified by `account_error_message` into `AppError::BlockingTask("runtime startup failed: marmot account session is already in use")`, so a MarmotKit host that hits contention via `reconcile`/`restart_account` gets an opaque `BlockingTask`, not `AccountSessionBusy`. The PR summary reads as if contention is typed everywhere; either narrow that claim or map the worker startup error through.

erskingardner (commented):
**Review verdict: request changes** (posted as a comment — GitHub does not allow requesting changes on your own PR).

Good, tightly-scoped fix for a real latent trap, and the guard-per-`AppClient` shape is the right one: it is small, in-process only, and releases on drop. The changelog, the typed FFI variant, and moving the runtime tests onto the managed worker are all correct.

My blocking concern is not the guard itself but its **release ordering** against how workers are actually torn down. The guard is released when the worker task's `AppClient` is dropped, but two teardown paths do not wait for that drop:

**1. `restart_account` can now fail (and permanently kill the worker).** `MarmotAppRuntime::restart_account` (`crates/marmot-app/src/runtime/mod.rs:3460-3468`) removes the worker and calls `ManagedAccountWorker::stop()`, which is `shutdown.send()` + `handle.abort()` and returns immediately — `abort()` only *schedules* cancellation, so the task's `AppClient` (and its guard) may not be dropped yet. `restart_account` then calls `reconcile().await`, which spawns a replacement worker for the same account. If the aborted task has not been reaped, `runtime_local_client` returns `AccountSessionBusy`; worker startup failure is **terminal** — `run_app_runtime_account_worker` publishes the error and `return`s with no retry — and `reconcile` surfaces `BlockingTask("runtime startup failed: marmot account session is already in use")`. The account then has no worker until something else triggers a reconcile.

This is a live production path: `crates/cli/src/daemon/runtime_host.rs:457` calls `restart_account` on refresh. CI passing does not clear it; it is a timing race that will win most of the time and lose under load.

Suggested fix: make the respawn ordered with the release — await teardown before reconcile (use `shutdown().await` in `restart_account`, or have `stop()` return the `JoinHandle` for the caller to await), or release ownership explicitly at the point the worker is removed from the `workers` map rather than relying on task-drop timing.

**2. A timed-out shutdown leaks the guard with no recovery.** `shutdown_with_timeout` (`crates/marmot-app/src/runtime/account_worker.rs:55-79`) aborts and then waits only 250ms. If that also elapses, the guard is never released and that account label becomes permanently unopenable for the lifetime of the `MarmotApp` — every later `client()` and every later worker start returns `AccountSessionBusy`, and there is no reclaim path short of restarting the process. Before this PR a straggler worker was merely wasteful. The new test `account_session_guard_is_exclusive_until_client_drop` depends on this path completing in time (`runtime.shutdown().await` then `app.client("alice")`), so it is also a flake source. Worth either a generation/reclaim token or, at minimum, an explicit decision recorded in the code that a stuck worker bricks the account.

See inline comments for the remaining (non-blocking) points.

PR #1201, opened 2026-07-30
Harden SQLite projection pruning and secure deletion

### LINKED ISSUE #749: storage-sqlite: push-token NOT IN(...) deletes bind one parameter per member with no chunking, unlike the timeline IN(...) discipline
**Severity:** LOW (only bites at scale the bundled engine makes unrealistic; but a correctness/robustness inconsistency)
**Component:** `crates/storage-sqlite`

## Summary

`timeline.rs` deliberately chunks message-id `IN(...)` lists at `SQLITE_BIND_PARAMETER_CHUNK = 900` "so the placeholder count per statement stays well under SQLite's variable limit" (`timeline.rs:38`). The account-projection `NOT IN(...)` deletes build one placeholder per element with no chunking. This is a **different function** from the legacy string-interpolated one in #344 (that one is in `marmot-app/src/projection.rs` and uses manual quoting; these use proper `params_from_iter` binding) and from #715 (timeline `IN(...)`).

## Location

- `crates/storage-sqlite/src/account_projection.rs:1162-1188` — `remove_stale_group_push_tokens` (`DELETE ... member_id_hex NOT IN ({placeholders})`, one `?` per `active_members` entry).
- Same pattern: `account_projection.rs:404-417` (`DELETE FROM account_groups ... NOT IN`) and `:1279-1293` (`delete_stale_group_components`).

## Details

```rust
let placeholders = std::iter::repeat_n("?", active_members.len()).collect::<Vec<_>>().join(",");
// DELETE FROM group_push_tokens WHERE group_id_hex = ? AND member_id_hex NOT IN ({placeholders})
```

With the workspace's `bundled-sqlcipher-vendored-openssl` build (`SQLITE_MAX_VARIABLE_NUMBER` = 32766) this only fails for a group with >32k members / an account in >32k groups — not realistic today, hence LOW. But if the crate is ever built against a system SQLite compiled with the historical `999` limit (exactly why the timeline code chunks at 900), `remove_stale_group_push_tokens` returns `SQLITE_TOOBIG` ("too many SQL variables") for any group with ≥999 members, permanently breaking push-token retention cleanup for large groups.

## Suggested fix

Route these `NOT IN` deletes through the same chunking helper the timeline uses (or a shared util), or delete via a temp table / `json_each` of the retained set instead of N bind parameters.

---
_Filed from a meticulous automated code-review pass; verified against source and deduplicated against existing open/closed issues (distinct from #344 and #715)._

### LINKED ISSUE #756: storage-sqlite: local group delete/wipe (`delete_local_group_data`) leaves message plaintext recoverable — no secure_delete, no WAL truncate
**Severity:** MEDIUM (requires local read access to the on-disk DB file after a "delete/wipe"; defeats the privacy expectation of the delete UX)
**Component:** `crates/storage-sqlite`

## Summary

`delete_local_group_data` is documented as *"the storage primitive for the local delete/wipe UX"* and deletes the plaintext-bearing rows for a group (`app_events.plaintext`, `message_timeline.plaintext`, `agent_stream_starts`, etc.). Unlike the sibling retention primitive `secure_prune_app_events_before`, it does **not** ensure `PRAGMA secure_delete = ON` around the transaction and does **not** run `PRAGMA wal_checkpoint(TRUNCATE)` afterward. In WAL mode (the default) the deleted plaintext therefore remains recoverable from the main DB file and/or the `-wal` sidecar until a later checkpoint happens to overwrite the pages — and if the store was opened with `secure_delete: false`, the freed pages are never zeroed at all.

## Location

- `crates/storage-sqlite/src/account_projection.rs:503-536` — `delete_local_group_data` (plain `with_transaction` + `DELETE FROM {table}`, no `secure_delete`/checkpoint).
- Contrast: `crates/storage-sqlite/src/account_projection.rs:655-713` — `secure_prune_app_events_before`, which deliberately (a) reads and forces `PRAGMA secure_delete = ON` **before** `BEGIN` (the comment at `:660-664` explains SQLite does not zero pages freed in the same transaction that toggles the pragma), (b) restores the original pragma, and (c) runs `checkpoint_wal_truncate_after_secure_prune` after a committed prune.

## Details

Failure scenario: a user asks to delete/wipe a conversation. `delete_local_group_data` runs the `DELETE`s inside a transaction that inherits whatever `secure_delete` the connection was opened with. In WAL mode the `DELETE`s append (possibly zeroed) page images to the `-wal` file, but the original pages in the main `.sqlite` file are only overwritten once a checkpoint copies pages across (default autocheckpoint ~1000 WAL pages, or on clean close). Until then, an attacker with read access to the DB file (a stolen/seized device, a backup, a shared-host misconfig) can recover the "deleted" message plaintext. If the store was opened with `SqliteStorageOptions { secure_delete: false, .. }`, the pages are never zeroed regardless of checkpointing.

This is exactly the disclosure `secure_prune_app_events_before` was hardened against; the user-facing wipe path is the one that most needs that posture and is the one missing it.

## Suggested fix

Give `delete_local_group_data` the same secure-delete posture as `secure_prune_app_events_before`: ensure `secure_delete = ON` around the delete transaction, restore the prior value, and run `checkpoint_wal_truncate_after_secure_prune` when rows were deleted — or route the wipe through the shared secure-prune machinery so both paths share one implementation.

---
<sub>Filed from a meticulous manual + multi-agent code-review pass; verified against source and deduplicated against existing open/closed issues.</sub>

### LINKED ISSUE #808: storage-sqlite: secure prune of expired plaintext leaves chat-list unread aggregates stale — inflated unread_count and a dangling first_unread_message_id_hex
## Summary

`secure_prune_app_events_before_tx` deletes `app_events` / `message_timeline` rows older than a cutoff and then calls `refresh_chat_list_last_message_after_secure_prune_tx` to keep the chat-list projection consistent. But that refresh only rewrites the **last-message** columns of `chat_list_rows`. It never recomputes the stored unread aggregates (`unread_count`, `unread_mention_count`, `first_unread_message_id_hex`). When *unread* messages are pruned, those materialized columns go stale.

## Location

`crates/storage-sqlite/src/timeline.rs`

The post-prune refresh only touches last-message columns (lines ~810–831):
```rust
"UPDATE chat_list_rows
 SET last_message_id_hex = ?1, last_message_sender = ?2, last_message_preview = ?3,
     last_message_kind = ?4, last_message_timeline_at = ?5, last_message_deleted = ?6,
     updated_at = ?7
 WHERE group_id_hex = ?8"
```
No update to `unread_count` / `unread_mention_count` / `first_unread_message_id_hex`.

These are **stored/materialized** columns on `chat_list_rows`, maintained incrementally via UPSERT and read back directly (`crates/storage-sqlite/src/chat_list.rs`: selected at lines ~920/932/954, `first_unread_message_id_hex` via `row.get(19)`, and summed across rows at lines 139–140 for the global badge). They are not recomputed from `message_timeline` on read, so nothing self-heals them after a prune.

The worker command that drives the prune does not refresh the chat list afterward either — `crates/marmot-app/src/runtime/account_worker.rs`:
```rust
AccountWorkerCommand::SecureDeleteExpiredPlaintext { group_id, respond } => {
    let result = client.secure_delete_expired_plaintext_for_group(&group_id);
    let _ = respond.send(result);
}
```

## Failure scenario

The prune is driven by the disappearing-messages retention path, whose cutoff is `now - disappearing_message_secs` (`crates/marmot-app/src/client/projection.rs`) — it expires messages by **age**, regardless of read state. So unread messages routinely get pruned. After that:

- `chat_list_rows.unread_count` / `unread_mention_count` stay inflated (they still count the pruned messages), so the unread badge is wrong until some unrelated full refresh recomputes it.
- `first_unread_message_id_hex` can point at a `message_id_hex` that was just deleted from `message_timeline`, so a client "scroll to first unread" resolves to a non-existent row.

The other projection surfaces the prune touches (`app_events`, `message_timeline`, `agent_stream_starts`, and the last-message columns) are kept correct, which is why this gap is easy to miss.

## Severity

**MEDIUM** (data/projection correctness). User-visible wrong unread counts and a dangling first-unread pointer after disappearing-message expiry.

## Suggested fix

Recompute the unread summary for the affected group inside the same prune transaction (the equivalent of the normal `record_app_event` chat-row refresh — recompute `unread_count`, `unread_mention_count`, `first_unread_message_id_hex` and update those columns), alongside the existing last-message refresh.

## Notes

Distinct from mdk#780 (secure_delete PRAGMA scope) and mdk#756 (`delete_local_group_data` recoverability). Verified against current working-tree source.

### LINKED ISSUE #846: storage-sqlite: save_account_projection_state builds an unchunked account_groups NOT IN (...) with one bind per retained group — a very large membership fails the whole projection save
**Severity:** LOW
**Component:** `crates/storage-sqlite`

## Summary

`save_account_projection_state` prunes stale `account_groups` rows with `DELETE ... WHERE group_id_hex NOT IN ({placeholders})`, generating one bind parameter per retained group with no chunking. Above SQLite's `SQLITE_MAX_VARIABLE_NUMBER` this fails with "too many SQL variables", and because the save runs in a single transaction the **entire account projection write is lost**. This is a distinct table/code path from the unchunked-`IN` sites already tracked (#715 `timeline_records_by_ids_tx`, #749 push-token `NOT IN`).

## Where

`crates/storage-sqlite/src/account_projection.rs:412-425`

```rust
let retained_group_ids = state.groups.iter()
    .map(|g| Value::Text(g.group_id_hex.clone()))
    .collect::<Vec<_>>();
let placeholders = (0..retained_group_ids.len()).map(|_| "?").collect::<Vec<_>>().join(", ");
conn.execute(
    &format!("DELETE FROM account_groups WHERE group_id_hex NOT IN ({placeholders})"),
    params_from_iter(retained_group_ids),
)?;   // one bind per group, no chunking
```

## Impact

An account (bot/agent-connector) whose group membership exceeds the parameter cap makes every projection save fail, so the account's projection can no longer be persisted. The bind count scales with number of groups per account, so the threshold is high (tens of thousands of groups) — much less likely to be hit than the message-count-bounded #715 — hence LOW, but it is the same unchunked-`IN` discipline gap the repo already tracks.

(Note: the sibling `delete_stale_group_components` `NOT IN` at `:1358` is bounded by the small fixed set of app-component types, so it is **not** at risk and is out of scope here.)

## Suggested fix

Chunk the retained group-id set (reuse the `SQLITE_BIND_PARAMETER_CHUNK` pattern used elsewhere in the crate), or delete-all-then-reinsert, matching the chunking discipline of the timeline deletes.

---
_Filed from a meticulous automated code-review pass (as of `0c88c95`); verified against source and deduplicated against #715 / #749._

### LINKED ISSUE #895: storage-sqlite: `save_account_projection_state` group prune cascade-deletes unsent message drafts + attachment plaintext under the documented concurrent-writer path
**Severity:** MEDIUM
**Component:** `crates/storage-sqlite` (`src/account_projection.rs`, `src/migrations/0026_message_drafts.rs`)

## Summary

`save_account_projection_state` snapshot-replaces the group set with `DELETE FROM account_groups WHERE group_id_hex NOT IN (...retained...)`. The `message_drafts` and `message_draft_attachments` tables declare `FOREIGN KEY (group_id_hex) ... ON DELETE CASCADE` (ultimately to `account_groups`), and `PRAGMA foreign_keys = true` is set. So pruning a group row **silently destroys that group's unsent draft body and attachment plaintext BLOBs** — irreplaceable user content — even though the projection-save model was scoped to re-derivable metadata.

## Location

- `crates/storage-sqlite/src/account_projection.rs:456-473` — the `DELETE FROM account_groups WHERE group_id_hex NOT IN (...)` prune inside `save_account_projection_state`.
- `crates/storage-sqlite/src/account_projection.rs:363-390` — the method's own doc: "Multiple runtimes may save over the same account database concurrently (for example a main app process and a short-lived notification-wake process)"; "Group and component rows are snapshot-replaced (last writer wins). Full multi-writer reconciliation is an explicit non-goal."
- `crates/storage-sqlite/src/migrations/0026_message_drafts.rs:8-37` — `message_drafts.group_id_hex ... REFERENCES account_groups(group_id_hex) ON DELETE CASCADE` (holds `content` plaintext + `reply_to`), and `message_draft_attachments.group_id_hex ... REFERENCES message_drafts(group_id_hex) ON DELETE CASCADE` (holds `plaintext BLOB`).
- `crates/storage-sqlite/src/connection.rs` — `PRAGMA foreign_keys = true`.

## Concrete failure scenario

1. The user types a draft (with an attachment) in group G. `save_message_draft` persists it into `message_drafts` / `message_draft_attachments`, keyed by G's `group_id_hex`.
2. A short-lived notification-wake process (documented as a concurrent writer) whose `StoredAccountState` snapshot was built slightly earlier — before G was joined/created, or simply not yet reflecting G — calls `save_account_projection_state`.
3. `state.groups` for that stale writer does not include G, so the `NOT IN` prune deletes G's `account_groups` row.
4. The FK cascade wipes G's `message_drafts` row and its `message_draft_attachments` plaintext BLOBs. The draft and its attachments are gone, with no error surfaced to either process.

The "last writer wins / re-derivable metadata" model (introduced for projection metadata) did not account for drafts and attachment plaintext now hanging off `account_groups` via cascade. Group metadata is re-derivable from the engine; unsent drafts are not.

## Impact

Silent, unrecoverable loss of unsent user content (draft text + attachment plaintext) whenever a concurrent projection save is performed by a writer whose group snapshot is momentarily stale — precisely the multi-writer scenario the method documents as supported.

## Suggested fix

Options:
- Exclude groups that still hold drafts from the prune (only delete `account_groups` rows with no dependent draft/attachment content), or
- Store drafts in a table not transitively cascade-linked to the snapshot-replaced `account_groups`, or
- Make draft/attachment retention independent of the "last writer wins" projection prune (e.g. soft-reconcile rather than `NOT IN` delete when dependent user content exists).

## Dedup

Distinct from #846 (the *unchunked* `NOT IN` bind-limit at this same statement) and #756 (insecure/recoverable delete in `delete_local_group_data`). This is a third, separate defect at this site: unintended *destruction* of user drafts via FK cascade under the documented concurrent-writer model.

---
_Filed from a meticulous automated code-review pass; verified against source (cascade chain + PRAGMA + documented concurrency model) and deduplicated against the existing open/closed issue set._

### LINKED ISSUE #936: storage-sqlite: secure_prune_app_events_before returns Ok even when the WAL TRUNCATE checkpoint fails — pruned plaintext stays recoverable while the caller is told the secure prune completed
## Summary

The secure retention prune scrubs and DELETEs message plaintext inside a WAL-mode transaction; the on-disk secure erase of the freed pages only actually lands once the WAL frames are checkpointed into the main DB and the WAL is truncated. When the post-prune `wal_checkpoint(TRUNCATE)` fails (busy readers), the function only logs a `warn!` and still returns `Ok(outcome)` with a positive `pruned_messages` count. The pruned plaintext therefore remains recoverable in the un-truncated `-wal` sidecar, yet the API signals that the secure prune succeeded.

## Location

`crates/storage-sqlite/src/account_projection.rs:702-760` (`secure_prune_app_events_before`); the swallow is at `:738-758`:

```rust
if outcome.pruned_messages > 0 {
    let conn = self.lock()?;
    if let Err(error) = checkpoint_wal_truncate_after_secure_prune(&conn) {
        tracing::warn!(/* ... error_kind ... */,
            "retention secure-delete WAL checkpoint failed after committed prune");
    }
}
Ok(outcome)   // <-- returns success regardless of checkpoint failure
```

`checkpoint_wal_truncate_after_secure_prune` (`:1390-1406`) returns `StorageError::Busy` when `wal_checkpoint(TRUNCATE)` reports `busy != 0` (active readers). That error is logged and dropped.

## Severity

LOW (security / durability — secure-erase silently non-durable).

## Failure scenario

A retention prune runs while any reader holds the WAL (a long timeline scan / projection read on a second connection). `wal_checkpoint(TRUNCATE)` can't complete → `Busy` after all retries. The function logs and returns `Ok`. The `-wal` sidecar on disk still contains the "securely pruned" message plaintext. Because `secure_delete` exists precisely to make deleted content unrecoverable even to a later key holder / forensic recovery, the caller believing erasure happened is the security gap.

## Suggested fix

Treat a failed post-prune truncate checkpoint as a first-class outcome — surface it to the caller (so it can retry the checkpoint once readers clear), or record that the erase is pending and re-drive it — rather than returning `Ok` with a positive prune count as if the secure erase were durable.

- **#780** — `PRAGMA secure_delete` save/restore across lock scopes leaving the pragma stuck ON (a different failure of the same function). Not the checkpoint-failure masking.
- **#756** — `delete_local_group_data` has *no* `secure_delete` and *no* WAL truncate at all. This path has both but swallows the truncate failure.
- **#808** — stale chat-list aggregates after prune (a projection-consistency issue), unrelated to secure erasure durability.

erskingardner:
## Summary

This PR fixes two related groups of SQLite storage issues:

- projection pruning now preserves draft-owning groups and avoids unbounded retained-set binds for account groups and push tokens;
- secure deletion now survives WAL checkpoint contention, applies to local group wipes, and refreshes unread projections transactionally.

## Projection pruning

The account projection and push-token cleanup paths previously deleted with `NOT IN (...)` over the retained set. Besides exceeding SQLite bind limits at scale, that shape cannot be safely chunked because an early chunk can delete values retained by a later chunk.

The shared helper:

1. queries the keys currently stored in the relevant scope;
2. computes stale keys in Rust;
3. deletes only stale keys using chunked positive `IN (...)` statements.

The shared bind budget includes fixed scope parameters as well as each key chunk. This design avoids both an unbounded retained-ID bind set and a new JSON1 runtime dependency.

Account-group pruning excludes rows that still own message drafts, preventing the foreign-key cascade from destroying unsent draft text and attachment plaintext. The resulting stale group remains visible in projection/chat-list reads while its draft exists. Once the draft is deleted, a later projection save makes the group eligible for cleanup again.

## Secure deletion

Secure retention pruning previously committed logical deletion and then hid a failed `wal_checkpoint(TRUNCATE)`. Retrying could see no rows and skip erasure, losing both the pending checkpoint and the original prune outcome.

This PR adds a durable secure-delete checkpoint intent that commits alongside logical deletion. The intent stores the accumulated result and a 128-bit nonce. Completion always retries WAL truncation and consumes the intent only when the nonce still matches, preventing stale finishers from deleting a recreated intent. The intent deliberately has no age-based garbage collection: it is cleared only after successful truncation.

Checkpoint contention after a committed delete is now a typed, non-fatal outcome. Storage, app, and UniFFI results expose `erasure_pending`; send/ingest and local-wipe callers can continue after the logical deletion while a later call retries truncation. Pre-commit writer contention remains an error.

Local group deletion uses the same contract:

- forces `PRAGMA secure_delete = ON` before beginning the transaction;
- restores the caller's prior setting;
- records a durable checkpoint intent with the wipe;
- returns the committed wipe result with `erasure_pending` when truncation is blocked;
- distinguishes a new logical deletion from completion of a previously pending checkpoint.

Its `deleted_rows` count is the accumulated count represented by one pending intent, so it can span multiple logical wipes if the group is re-created and deleted again before truncation succeeds.

Secure pruning also recomputes unread count, mention count, and first-unread message ID inside the prune transaction alongside the existing last-message refresh.

## Validation

- `cargo test -p storage-sqlite` — 307 passed
- `cargo test -p marmot-app client::retention` — 10 passed
- `cargo test -p marmot-uniffi secure_delete_expired_result_ffi_preserves_count_and_hashes`
- `just fast-ci`
- `git diff --check`

Closes #895
Closes #846
Closes #749
Closes #936
Closes #756
Closes #808

### DIFF
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -792,6 +792,9 @@ pub struct SecureDeleteExpiredResult {
     /// list is sorted for deterministic output, but callers should treat it as
     /// an unordered purge set.
     pub media_ciphertext_sha256: Vec<String>,
+    /// True when logical deletion committed but secure WAL truncation remains
+    /// pending. A later retention pass will retry it.
+    pub erasure_pending: bool,
 }
 
 impl From<SecurePruneAppEventsResult> for SecureDeleteExpiredResult {
@@ -800,6 +803,7 @@ impl From<SecurePruneAppEventsResult> for SecureDeleteExpiredResult {
             pruned_messages: value.pruned_messages as u64,
             secrets_deleted: value.pruned_media_epoch_secrets as u64,
             media_ciphertext_sha256: value.media_ciphertext_sha256,
+            erasure_pending: value.erasure_pending,
         }
     }
 }
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -3715,7 +3715,7 @@ impl MarmotApp {
         {
             return Err(AppError::GroupDisbanding(group_id_hex.to_owned()));
         }
-        let deleted = storage.delete_local_group_data(group_id_hex)?;
+        let deleted = storage.delete_local_group_data(group_id_hex)?.did_delete();
         self.chat_list_projection_stale
             .lock()
             .unwrap_or_else(|poisoned| poisoned.into_inner())
@@ -4095,9 +4095,16 @@ impl MarmotApp {
         group_id_hex: &str,
         now: u64,
     ) -> Result<SecureDeleteExpiredResult, AppError> {
+        let account = self.account_home().account(label)?;
+        let classifier = Self::chat_list_mention_classifier(&account.account_id_hex);
         Ok(self
-            .account_storage(label)?
-            .secure_prune_expired_app_events(group_id_hex, now)?
+            .account_storage(&account.label)?
+            .secure_prune_expired_app_events(
+                group_id_hex,
+                now,
+                &account.account_id_hex,
+                &classifier,
+            )?
             .into())
     }
 
--- a/crates/storage-sqlite/src/account_projection.rs
+++ b/crates/storage-sqlite/src/account_projection.rs
@@ -10,6 +10,34 @@ use rusqlite::{
 };
 use serde::{Deserialize, Serialize};
 
+/// Keep generated positive-`IN` deletes below SQLite's historical 999-variable
+/// default, including any scope parameters that precede the key list.
+const SQLITE_BIND_PARAMETER_CHUNK: usize = 900;
+const SECURE_DELETE_RETENTION_OPERATION: &str = "retention";
+const SECURE_DELETE_LOCAL_GROUP_OPERATION: &str = "local_group";
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct DeleteLocalGroupDataResult {
+    /// Rows removed by the logical wipe represented by this completed request.
+    /// On a retry this is recovered from the durable checkpoint intent.
+    pub deleted_rows: usize,
+    /// True when this call completed a WAL erasure checkpoint left pending by
+    /// an earlier call whose logical deletion had already committed.
+    pub completed_pending_checkpoint: bool,
+}
+
+impl DeleteLocalGroupDataResult {
+    pub fn did_delete(&self) -> bool {
+        self.deleted_rows > 0
+    }
+}
+
+#[derive(Clone, Debug)]
+struct SecureDeleteIntent {
+    nonce: Vec<u8>,
+    result_json: String,
+}
+
 /// Whether a stored per-chat mute row is effective at `now_ms`.
 ///
 /// Missing rows are unmuted, `NULL` means an indefinite mute, and finite
@@ -461,7 +489,9 @@ impl SqliteAccountStorage {
     ///   `max_seen_events` rows, so a re-seen event outlives rows whose only
     ///   sighting is old. It only narrows cross-restart redelivery;
     ///   engine-level dedup stays the authoritative duplicate guard.
-    /// - Group and component rows are snapshot-replaced (last writer wins).
+    /// - Group and component rows are snapshot-replaced (last writer wins),
+    ///   except a group row remains while it owns an unsent draft so pruning
+    ///   re-derivable metadata cannot cascade-delete user-authored content.
     ///   Full multi-writer reconciliation is an explicit non-goal.
     ///
     /// `max_future_skew_secs` is caller policy (the app layer passes its
@@ -528,24 +558,29 @@ impl SqliteAccountStorage {
             )
             .storage()?;
 
-            if state.groups.is_empty() {
-                conn.execute("DELETE FROM account_groups", []).storage()?;
-            } else {
-                let retained_group_ids = state
-                    .groups
-                    .iter()
-                    .map(|group| Value::Text(group.group_id_hex.clone()))
-                    .collect::<Vec<_>>();
-                let placeholders = (0..retained_group_ids.len())
-                    .map(|_| "?")
-                    .collect::<Vec<_>>()
-                    .join(", ");
-                conn.execute(
-                    &format!("DELETE FROM account_groups WHERE group_id_hex NOT IN ({placeholders})"),
-                    params_from_iter(retained_group_ids),
-                )
-                .storage()?;
-            }
+            let retained_group_ids = state
+                .groups
+                .iter()
+                .map(|group| group.group_id_hex.as_str())
+                .collect::<std::collections::HashSet<_>>();
+            // Draft text and attachment plaintext are user-authored durable data,
+            // unlike this re-derivable projection. Exclude their owning groups
+            // from stale candidates so the account_groups FK cascade cannot
+            // destroy them. Once the draft is deleted, the next save selects the
+            // group as a candidate and cleans it up normally.
+            delete_stale_text_keys(
+                &conn,
+                "SELECT group_id_hex
+                 FROM account_groups
+                 WHERE NOT EXISTS (
+                    SELECT 1 FROM message_drafts
+                    WHERE message_drafts.group_id_hex = account_groups.group_id_hex
+                 )",
+                &[],
+                "DELETE FROM account_groups WHERE group_id_hex IN",
+                &[],
+                &retained_group_ids,
+            )?;
 
             for group in &state.groups {
                 let group_was_new = !conn
@@ -676,8 +711,14 @@ impl SqliteAccountStorage {
     /// re-create the app projection. A terminal group is different: its live
     /// MLS state is already erased, so this transaction also removes the full
     /// `cgka_groups` row and retains only `cgka_disband_tombstones` as the
-    /// permanent anti-resurrection guard.
-    pub fn delete_local_group_data(&self, group_id_hex: &str) -> StorageResult<bool> {
+    /// permanent anti-resurrection guard. The logical wipe and a durable WAL
+    /// checkpoint intent commit together under `secure_delete = ON`; success is
+    /// returned only after `wal_checkpoint(TRUNCATE)` completes. A retry after
+    /// checkpoint contention recovers the original result from that intent.
+    pub fn delete_local_group_data(
+        &self,
+        group_id_hex: &str,
+    ) -> StorageResult<DeleteLocalGroupDataResult> {
         if group_id_hex.trim().is_empty() {
             return Err(StorageError::Backend(
                 "local group delete id must not be empty".to_owned(),
@@ -687,50 +728,83 @@ impl SqliteAccountStorage {
             StorageError::Serialization(format!("invalid local group id: {error}"))
         })?;
 
-        self.connection.with_transaction(|| -> StorageResult<bool> {
-            let conn = self.lock()?;
-            let mut deleted = retire_all_encrypted_media_secrets_for_group_tx(&conn, group_id_hex)?;
-            for table in [
-                "app_events",
-                "message_timeline",
-                "agent_stream_starts",
-                "conversation_read_state",
-                "chat_list_rows",
-                "account_group_app_components",
-                "group_push_tokens",
-                "group_push_token_tombstones",
-                "pending_push_registration_shares",
-                "chat_notification_settings",
-                "encrypted_media_epoch_secret_references",
-                "encrypted_media_epoch_secrets",
-                "account_groups",
-            ] {
-                deleted = deleted.saturating_add(
-                    conn.execute(
-                        &format!("DELETE FROM {table} WHERE group_id_hex = ?1"),
-                        params![group_id_hex],
-                    )
-                    .storage()?,
-                );
-            }
-            let terminal = conn
-                .query_row(
-                    "SELECT EXISTS(
-                        SELECT 1 FROM cgka_disband_tombstones WHERE group_id = ?1
-                     )",
-                    params![&group_id],
-                    |row| row.get::<_, i64>(0),
-                )
-                .storage()?
-                != 0;
-            if terminal {
-                deleted = deleted.saturating_add(
-                    conn.execute("DELETE FROM cgka_groups WHERE id = ?1", params![&group_id])
+        let newly_deleted = retry_on_busy(|| {
+            let mut conn = self.lock()?;
+            let original = secure_delete_pragma(&conn)?;
+            conn.execute_batch("PRAGMA secure_delete = ON;").storage()?;
+            let delete_result = (|| {
+                let tx = conn
+                    .transaction_with_behavior(TransactionBehavior::Immediate)
+                    .storage()?;
+                let mut deleted =
+                    retire_all_encrypted_media_secrets_for_group_tx(&tx, group_id_hex)?;
+                for table in [
+                    "app_events",
+                    "message_timeline",
+                    "agent_stream_starts",
+                    "conversation_read_state",
+                    "chat_list_rows",
+                    "account_group_app_components",
+                    "group_push_tokens",
+                    "group_push_token_tombstones",
+                    "pending_push_registration_shares",
+                    "chat_notification_settings",
+                    "encrypted_media_epoch_secret_references",
+                    "encrypted_media_epoch_secrets",
+                    "account_groups",
+                ] {
+                    deleted = deleted.saturating_add(
+                        tx.execute(
+                            &format!("DELETE FROM {table} WHERE group_id_hex = ?1"),
+                            params![group_id_hex],
+                        )
                         .storage()?,
-                );
+                    );
+                }
+                let terminal = tx
+                    .query_row(
+                        "SELECT EXISTS(
+                            SELECT 1 FROM cgka_disband_tombstones WHERE group_id = ?1
+                         )",
+                        params![&group_id],
+                        |row| row.get::<_, i64>(0),
+                    )
+                    .storage()?
+                    != 0;
+                if terminal {
+                    deleted = deleted.saturating_add(
+                        tx.execute("DELETE FROM cgka_groups WHERE id = ?1", params![&group_id])
+                            .storage()?,
+                    );
+                }
+                if deleted > 0 {
+                    merge_local_group_delete_intent_tx(&tx, group_id_hex, deleted)?;
+                }
+                tx.commit().storage()?;
+                Ok(deleted)
+            })();
+            let restore = restore_secure_delete_pragma(&conn, original);
+            combine_secure_delete_operation_and_restore(delete_result, restore)
+        })?;
+
+        match finish_secure_delete_checkpoint_intent::<DeleteLocalGroupDataResult>(
+            self,
+            SECURE_DELETE_LOCAL_GROUP_OPERATION,
+            group_id_hex,
+        )? {
+            Some(mut result) => {
+                result.completed_pending_checkpoint = newly_deleted == 0 && result.deleted_rows > 0;
+                Ok(result)
             }
-            Ok(deleted > 0)
-        })
+            // Another process may have checkpointed and consumed this intent
+            // after our logical deletion committed. In that case erasure is
+            // complete, and this caller must still report its locally known
+            // deletion instead of incorrectly returning "nothing existed".
+            None => Ok(DeleteLocalGroupDataResult {
+                deleted_rows: newly_deleted,
+                completed_pending_checkpoint: false,
+            }),
+        }
     }
 
     /// Record the local account's own membership in `group_id_hex` so the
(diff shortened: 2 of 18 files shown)


crates/marmot-app/src/lib.rs
```diff
@@ -3715,7 +3715,7 @@ impl MarmotApp {
         {
             return Err(AppError::GroupDisbanding(group_id_hex.to_owned()));
         }
-        let deleted = storage.delete_local_group_data(group_id_hex)?;
+        let deleted = storage.delete_local_group_data(group_id_hex)?.did_delete();
```
erskingardner, line 3722: `delete_local_group_data` can now return `Err(Busy)` *after* the logical wipe has committed (the new `delete_local_group_data_reopens_and_retries_a_committed_pending_checkpoint` test asserts exactly this). On that path `?` returns before the two blocks below, so `chat_list_projection_stale` is never marked and `chat_list_projection_warmed` never cleared — the group is gone from storage but the cached chat list still shows it until some unrelated event invalidates it.

Set the staleness/warmed flags before propagating the error, or return the `DeleteLocalGroupDataResult` and let the caller decide (`completed_pending_checkpoint` exists precisely to express "rows are gone, erasure landed late").

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -963,26 +1057,33 @@ impl SqliteAccountStorage {
...
-            match (prune_outcome, restore) {
-                (Ok(value), Ok(())) => Ok(value),
-                (Err(err), Ok(())) => Err(err),
-                (Ok(_), Err(err)) => Err(err),
-                (Err(err), Err(_)) => Err(err),
-            }
+            let restore = restore_secure_delete_pragma(&conn, original);
+            combine_secure_delete_operation_and_restore(prune_outcome, restore)
         })?;
+        let outcome = finish_secure_delete_checkpoint_intent::<
```
erskingardner: This is the point where a busy truncate stops being a `warn!` and becomes a hard `Err` for every caller. Given `merge_secure_prune_intent_tx` already makes the counts and media hashes durable and the *next* prune re-drives the checkpoint, consider whether the storage API should surface this as a distinct, non-fatal signal (e.g. an `erasure_pending: bool` on `SecurePruneAppEventsResult`) instead of an error that aborts the send path in `client/sync.rs:653` and `client/mod.rs:1482`. As it stands the failure mode moved from "erasure silently deferred" to "published message reported as failed".

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -2214,6 +2437,51 @@ fn all_account_group_components(
...
 }
 
+/// Deletes text keys that exist in a scoped query but are absent from
+/// `retained_keys`.
+///
+/// The retained set is never bound into SQL: `NOT IN` cannot be safely split
+/// into chunks because each partial statement would delete keys retained by a
+/// later chunk. Instead, this queries existing keys, computes the stale set in
+/// Rust, and chunk-deletes that stale set with positive `IN` predicates.
+fn delete_stale_text_keys(
```
erskingardner, line 2487: The read-back-then-chunk approach is correct, but #749 suggested a materially simpler shape that avoids both the read-back and the chunking entirely: bind the retained set once as a JSON array and use `NOT IN (SELECT value FROM json_each(?1))`. For the `account_groups` case that collapses the whole helper into the original single statement:

```sql
DELETE FROM account_groups
WHERE group_id_hex NOT IN (SELECT value FROM json_each(?1))
  AND NOT EXISTS (SELECT 1 FROM message_drafts
                  WHERE message_drafts.group_id_hex = account_groups.group_id_hex)
```

One bind parameter, no bind-limit exposure, no `HashSet`, no reading every existing key into memory on every projection save (which for the >32k-group accounts #846 is about is the case you are trying to make cheap). No SQL-level JSON1 usage exists in the crate today, so this would need a quick check that `json_each` is available in the bundled SQLCipher build — if it is, this helper and its two-slice parameter API can go away.

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -10,6 +10,34 @@ use rusqlite::{
 };
 use serde::{Deserialize, Serialize};
 
+/// Keep generated positive-`IN` deletes below SQLite's historical 999-variable
+/// default, including any scope parameters that precede the key list.
+const SQLITE_BIND_PARAMETER_CHUNK: usize = 900;
```
erskingardner: `SQLITE_BIND_PARAMETER_CHUNK: usize = 900` already exists at `crates/storage-sqlite/src/timeline.rs:44` with the same value and the same rationale. Duplicating it means the two can drift; hoist one copy into a shared module (`codec.rs` or `lib.rs`) and use it from both.

Also, the doc comment says the 900 accounts for "any scope parameters that precede the key list", but the code below chunks *keys* at 900 and then prepends `delete_prefix_params` on top, so the real statement binds 901 for the push-token calls. Still under 999, but the comment does not describe what the code does — either subtract the prefix length from the chunk size or fix the comment.

crates/storage-sqlite/src/migrations/0041_secure_delete_checkpoint_intents.rs
```diff
@@ -0,0 +1,19 @@
...
+use rusqlite::Transaction;
+
+pub(crate) fn apply(tx: &Transaction<'_>) -> StorageResult<()> {
+    tx.execute_batch(
+        r#"
+CREATE TABLE secure_delete_checkpoint_intents (
+    operation_kind TEXT NOT NULL,
+    scope TEXT NOT NULL,
+    intent_nonce BLOB NOT NULL,
+    result_json TEXT NOT NULL,
```
erskingardner, line 12: `created_at_unix_seconds` is written by `upsert_secure_delete_intent_tx` and never read anywhere. Either drop the column, or use it — there is currently no GC for orphaned intents. An intent whose scope is never touched again (group deleted and never re-deleted, account abandoned) lives forever, and for the retention operation its `result_json` carries accumulated `media_ciphertext_sha256` values. A stale intent also means the *next* call for that scope returns the old accumulated counts rather than its own; bounded and self-limiting, but a documented age-out would make the table's lifecycle explicit.

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -528,24 +558,29 @@ impl SqliteAccountStorage {
...
+                .groups
+                .iter()
+                .map(|group| group.group_id_hex.as_str())
+                .collect::<std::collections::HashSet<_>>();
+            // Draft text and attachment plaintext are user-authored durable data,
+            // unlike this re-derivable projection. Exclude their owning groups
+            // from stale candidates so the account_groups FK cascade cannot
+            // destroy them. Once the draft is deleted, the next save selects the
+            // group as a candidate and cleans it up normally.
+            delete_stale_text_keys(
```
erskingardner, line 586: Excluding draft-owning groups from the stale candidate set fixes the cascade (good, and the new `message_draft_round_trips_and_survives_projection_pruning` test pins it), but it has a user-visible consequence that is not called out in the PR body or the method doc: a group the account has actually left or been removed from now stays in `load_account_projection_state` indefinitely as long as an unsent draft exists — the new test asserts `groups.len() == 1` for exactly that state. So a removed group can keep appearing in the chat list until the draft is deleted.

If that is intended, say so in the doc comment above (the current wording only explains the FK motive, not the retention side effect). If it is not, #895's second option — moving drafts off the cascade chain — avoids resurrecting projection rows to protect unrelated data.

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -687,50 +728,83 @@ impl SqliteAccountStorage {
...
+            combine_secure_delete_operation_and_restore(delete_result, restore)
+        })?;
+
+        match finish_secure_delete_checkpoint_intent::<DeleteLocalGroupDataResult>(
+            self,
+            SECURE_DELETE_LOCAL_GROUP_OPERATION,
+            group_id_hex,
+        )? {
+            Some(mut result) => {
+                result.completed_pending_checkpoint = newly_deleted == 0 && result.deleted_rows > 0;
```
erskingardner: Minor: `merge_local_group_delete_intent_tx` accumulates `deleted_rows` across calls for the same scope, so if a pending intent is left behind and the group is later re-joined and re-deleted, the returned `deleted_rows` is the sum of both wipes rather than this call's. Only the count/`did_delete()` is affected, but since `deleted_rows` is documented as "rows removed by the logical wipe represented by this completed request", the accumulate semantics are worth a word in that doc.

erskingardner (commented):
**Recommendation: request changes** (posting as a comment — GitHub does not allow requesting changes on your own PR).

Solid, well-tested work on the mechanism itself — the durable intent + nonce design is correct, the chunked positive-`IN` rewrite is the right shape, and the new tests (competing finishers, stale nonce, lowered `SQLITE_LIMIT_VARIABLE_NUMBER`) actually exercise the hard parts. But one consequence of making WAL truncation mandatory was not carried out to the callers, and I think it is blocking.

**Blocking**

1. **A busy WAL checkpoint now fails already-published message sends.** `secure_prune_app_events` used to log and swallow a failed `wal_checkpoint(TRUNCATE)`; it now returns `StorageError::Busy` after the prune has committed. That error flows through `MarmotAppClient::prune_plaintext_retention_for_group(...)?`, which is called with `?` on hot paths *after* the network publish has already succeeded:
   - `crates/marmot-app/src/client/sync.rs:653` (send path, after `.await?` publish)
   - `crates/marmot-app/src/client/mod.rs:1482` (send path, after `fail_if_publish_failed`)
   - `crates/marmot-app/src/client/mod.rs:995`, `:1781`, `:2369`
   - `crates/marmot-app/src/client/sync.rs:819` (`ingest_delivery` loop)

   The trigger is exactly the scenario #936 describes as routine: any concurrent reader on a second connection (e.g. the documented notification-wake process) holding the WAL. The result is a user-visible send/ingest failure for an operation that fully succeeded, with only *erasure* deferred — and the whole point of the new intent table is that the deferral is durable and self-healing on the next prune. This also cuts against the repo's multi-step-state-change invariant (`docs/marmot-architecture/overview/multi-step-state-changes.md`): the send reached the network, so it must not be reported as failed.

   The retention *sweep* path already contains per-group failures via `failed_group_outcome`, so it can take the hard error. The send/ingest call sites should tolerate `StorageError::Busy` from the prune (log at the existing retention target and continue), or the storage API should distinguish "logically pruned, erasure pending" from "prune failed" so callers can choose. As written, the PR changes a swallowed warning into a failed user action without touching any of those six call sites.

2. **`delete_group_local_data` skips its projection invalidation on the new error path** — see inline comment on `crates/marmot-app/src/lib.rs`.

**Non-blocking**

3. Scope: six issues plus a schema migration in one 1126/189 PR. The projection-pruning half and the secure-deletion half are independent and would be easier to review and to revert separately. Not asking for a split now, but worth noting against the surgical-scope bar.
4. `delete_stale_text_keys` design, duplicated chunk constant, unused `created_at_unix_seconds`, and the draft-retention side effect on the projection — inline.

Happy to re-review quickly once the caller-side handling of a pending checkpoint is settled.

PR #1202, opened 2026-07-30
Add subject-level convergence virtual time

erskingardner:
## Summary

- reconcile the convergence reliability tracking plan with recently merged engine, storage, and runtime hardening work
- add a capability-gated, serializable `advance_time` scenario operation
- install one shared paired monotonic/wall clock across all clients in `EngineHarnessSubject`
- keep global time advancement separate from participant execution: `advance_time` mutates only the shared clock, while `tick` selects which runtimes wake and observe it
- route opted-in subject ticks through the injected clock instead of the historical far-future settlement timestamp
- classify virtual-time advancement in oracle coverage and document the scenario surface

## Why

The merged dual-clock work made convergence deadlines injectable inside the engine, but portable simulator scenarios still could not control those deadlines through `ConvergenceSubject`. This slice closes that boundary without real sleeps or engine/storage rollback machinery.

Separating clock movement from participant execution makes the operation failure-free and models offline participants directly: global time may advance while an unticked participant remains paused.

## Behavior and compatibility

Every engine subject uses a deterministic shared manual clock. A scenario switches `tick` to that clock only after executing `advance_time`; scenarios that never opt in, and standalone `HarnessClient` tests, retain the historical far-future `tick` shortcut. Adapters that do not declare `virtual_time` fail preflight before any scenario action runs.

This does not yet implement structural quiescence, fixed-point driving, or public outbound polling/acknowledgement; those remain separate follow-up slices in the tracking plan.

## Verification

- `cargo test -p cgka-conformance-simulator virtual_time`
- `just fast-ci`
- `git diff --check`
- documentation stale-term scan

The regression uses two independent real durable disband passes: both remain live after Alice ticks at 999 ms; Alice settles when she ticks at 1,000 ms while unticked Bob stays live; Bob then settles on a later tick without another time advance.

### DIFF
--- a/crates/cgka-conformance-simulator/README.md
+++ b/crates/cgka-conformance-simulator/README.md
@@ -19,7 +19,8 @@ welcomes use NIP-59 gift wraps before the bus delivers them.
 - `HarnessClient` — wraps `Engine<SqliteAccountStorage>` and the real Nostr transport peeler while keeping delivery in memory.
 - `ConvergenceSubject` — a capability-declared semantic boundary between scenario execution and the implementation
   under test. `EngineHarnessSubject` is the in-process adapter; queue and partition mutation live on a separate,
-  explicitly white-box fault interface.
+  explicitly white-box fault interface. Its `virtual_time` capability advances one shared paired convergence clock and
+  runs scheduled convergence only for the selected awake participants.
 - `ScenarioSpec` — a serializable v1 input contract for deterministic scripted scenarios, including explicit queue
   faults and partitions.
 - `VectorFixture` — portable JSON fixtures pairing runnable scenario input with exact traces or semantic expected
@@ -117,8 +118,9 @@ engine work (publish lifecycle, convergence pass/input, queued outbound, deferre
 buffers), bus queue/delayed/mailbox work addressed to that client, and that client's pending scenario-input ledger
 entries. The expectation fails with the blocking subsystem names. It is deliberately not an end-to-end delivery claim
 for an application object that a transport fault dropped; scenarios that require delivery must assert the recipient's
-ledger or output separately. It does not yet wait for quiescence: virtual-time advancement, structural progress tokens,
-timeout policy, and retry-timer coverage belong to Milestone 1.3.
+ledger or output separately. The serializable `advance_time` step now controls convergence deadlines without sleeping
+and can leave selected participants paused. It does not yet wait for quiescence: structural progress tokens, a
+fixed-point driver, timeout policy, outbound acknowledgement, and retry-timer coverage remain in Milestone 1.3.
 
 `probe_bidirectional_decryptability` is the active cryptographic-reachability check. Each named client sends one
 uniquely identified application event through the normal engine and transport path; the runner delivers the resulting
--- a/crates/cgka-conformance-simulator/README.md
+++ b/crates/cgka-conformance-simulator/README.md
@@ -19,8 +19,8 @@ welcomes use NIP-59 gift wraps before the bus delivers them.
 - `HarnessClient` — wraps `Engine<SqliteAccountStorage>` and the real Nostr transport peeler while keeping delivery in memory.
 - `ConvergenceSubject` — a capability-declared semantic boundary between scenario execution and the implementation
   under test. `EngineHarnessSubject` is the in-process adapter; queue and partition mutation live on a separate,
-  explicitly white-box fault interface. Its `virtual_time` capability advances one shared paired convergence clock and
-  runs scheduled convergence only for the selected awake participants.
+  explicitly white-box fault interface. Its `virtual_time` capability advances one shared paired convergence clock;
+  subsequent `tick` steps select which participant runtimes wake and observe the elapsed deadline.
 - `ScenarioSpec` — a serializable v1 input contract for deterministic scripted scenarios, including explicit queue
   faults and partitions.
 - `VectorFixture` — portable JSON fixtures pairing runnable scenario input with exact traces or semantic expected
@@ -118,8 +118,11 @@ engine work (publish lifecycle, convergence pass/input, queued outbound, deferre
 buffers), bus queue/delayed/mailbox work addressed to that client, and that client's pending scenario-input ledger
 entries. The expectation fails with the blocking subsystem names. It is deliberately not an end-to-end delivery claim
 for an application object that a transport fault dropped; scenarios that require delivery must assert the recipient's
-ledger or output separately. The serializable `advance_time` step now controls convergence deadlines without sleeping
-and can leave selected participants paused. It does not yet wait for quiescence: structural progress tokens, a
+ledger or output separately. The serializable `advance_time` step advances the shared convergence clock without
+sleeping or waking a participant; a later `tick` selects the awake runtimes and uses that same clock. Every engine
+subject carries the deterministic manual clock, but only a scenario that executes `advance_time` switches `tick` to
+that clock. Existing scenarios and standalone `HarnessClient` tests that never opt into virtual time retain the
+historical far-future `tick` shortcut. The simulator does not yet wait for quiescence: structural progress tokens, a
 fixed-point driver, timeout policy, outbound acknowledgement, and retry-timer coverage remain in Milestone 1.3.
 
 `probe_bidirectional_decryptability` is the active cryptographic-reachability check. Each named client sends one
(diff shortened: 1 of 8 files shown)


crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -557,6 +578,33 @@ impl ConvergenceSubject for EngineHarnessSubject {
...
+            self.client(label)?;
+            if !selected.insert(label) {
+                return Err(SubjectError::new(
+                    "duplicate_time_client",
+                    format!("duplicate virtual-time client {label}"),
+                ));
+            }
+        }
+        self.convergence_clock.advance_ms(delta_ms);
+        for label in clients {
```
erskingardner: `advance_time` is not authoritative over convergence deadlines, because `HarnessClient::tick()` and `drive_due_convergence()` still call `advance_convergence_inputs_until_settled` / `converge_and_drain_queued_outbound_intents` with `HARNESS_CONVERGENCE_SETTLED_AT_MS = 1_000_000` as the monotonic `now_ms`, independent of `self.convergence_clock`.

So this scenario:

```
advance_time { delta_ms: 999, clients: [alice] }   // pass deliberately still live
tick { clients: [alice] }                          // <- monotonic jumps to 1_000_000
```

cuts the pass off at `now.monotonic_ms >= pass.cutoff_monotonic_ms()`, discarding whatever deadline the virtual clock was holding. It also hands the engine a mismatched pair — `monotonic_ms = 1_000_000` alongside `wall_ms = 0` from the manual clock — which is exactly the kind of inconsistency the paired-clock work in #1195 exists to prevent, and which the restart-rebasing arithmetic in `distributed_convergence.rs` reads.

The plan text in this PR does concede that driving subject operations from the shared clock is still open in 1.3, so I am not asking for the full fix here. But `advance_time` is now a *public, serializable* scenario step, and a scenario author has no way to know that an adjacent `tick` silently overrides it. Two options, either is fine:

- switch the two harness call sites to `self.engine().convergence_now().monotonic_ms` now (this looks small, and `advance_convergence()` already demonstrates the pattern), or
- document the override explicitly on `ScenarioStep::AdvanceTime` and in README/SCENARIOS.md, and add a regression pinning the current behavior so the eventual 1.3 change is a visible, intentional break.

Relatedly, the README wording added in this PR ("the serializable `advance_time` step now controls convergence deadlines without sleeping") overclaims while `tick` can bypass it.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -833,3 +881,72 @@ fn observe_engine_error(error: &EngineError) -> String {
     }
     .into()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::ConformanceCanonicalStateSnapshot;
+
```
erskingardner, line 864: This test proves the clock reaches the engine and that the 1,000 ms cutoff is real — good, and the 999/1000 split is the right shape.

But it does not test the feature's central claim. The PR summary, the trait rustdoc, and the plan entry all say global time advances once while scheduled convergence runs *only for selected awake participants*; here there is exactly one client and it is the selected one, so the selection argument is doing no work. `virtual_time_is_capability_gated_and_dispatched_with_selected_clients` only checks dispatch against a recording stub, not engine effect.

A two-client variant would close it: put both in a group, `advance_time(1000, &["alice"])`, then assert alice settled while bob is still `Live` — and that a subsequent `advance_time(0, &["bob"])` (or any later selection of bob) settles bob from the already-elapsed shared clock. That last assertion is the one that proves time is genuinely shared rather than per-client, which is the whole point of the single `ManualConvergenceClock`.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -557,6 +578,33 @@ impl ConvergenceSubject for EngineHarnessSubject {
         Ok(())
     }
 
+    async fn advance_time(
+        &mut self,
+        delta_ms: u64,
+        clients: &[String],
+    ) -> Result<(), SubjectError> {
+        let mut selected = BTreeSet::new();
+        for label in clients {
```
erskingardner: Two small things, both optional.

The duplicate check is new machinery no sibling step has: `tick` happily accepts duplicates, and `probe_bidirectional_decryptability` rejects them with a different formulation and a different error code. A repeated label here would just run `advance_convergence()` twice for that client, which is idempotent-ish and harmless — so this is error handling for a scenario that costs nothing. Either drop it, or reuse the probe's `sort`/`dedup` shape so the codebase has one convention rather than three.

Separately, the second loop does `self.client(label)?` and then `self.client_mut(label)?`, so each label costs three `BTreeMap` lookups. Since `has_default_group()` and `advance_convergence()` can share one mutable borrow sequentially, this collapses to:

```rust
for label in clients {
    self.client(label)?;
}
self.convergence_clock.advance_ms(delta_ms);
for label in clients {
    let client = self.client_mut(label)?;
    if client.has_default_group() {
        client.advance_convergence().await.map_err(subject_engine_error)?;
    }
}
```

which keeps validate-before-mutate intact and drops the method to about half its size.

crates/cgka-conformance-simulator/README.md
```diff
@@ -117,8 +118,9 @@ engine work (publish lifecycle, convergence pass/input, queued outbound, deferre
 buffers), bus queue/delayed/mailbox work addressed to that client, and that client's pending scenario-input ledger
 entries. The expectation fails with the blocking subsystem names. It is deliberately not an end-to-end delivery claim
 for an application object that a transport fault dropped; scenarios that require delivery must assert the recipient's
-ledger or output separately. It does not yet wait for quiescence: virtual-time advancement, structural progress tokens,
-timeout policy, and retry-timer coverage belong to Milestone 1.3.
+ledger or output separately. The serializable `advance_time` step now controls convergence deadlines without sleeping
+and can leave selected participants paused. It does not yet wait for quiescence: structural progress tokens, a
```
erskingardner: "controls convergence deadlines" is stronger than what ships, since a `tick` step in the same scenario overrides the virtual deadline with the synthetic `HARNESS_CONVERGENCE_SETTLED_AT_MS` monotonic timestamp (see the comment on `subject.rs`). Suggest either softening to something like "advances the shared convergence clock without sleeping; `tick` still forces settlement at a synthetic timestamp until Milestone 1.3 routes it through the same clock", or fixing the two harness call sites and keeping this sentence as written.

erskingardner (commented):
Solid, well-scoped slice: the shared `Arc<Mutex<_>>` inside `ManualConvergenceClock` makes the `clone()` genuinely shared, `HarnessClient::restart` already re-injects the clock so `virtual_time` composes with `CrashReopen`, validation runs before the clock mutates, and the disband regression at 999/1000 ms is real evidence rather than a mocked deadline. Docs and the tracking plan are updated honestly.

One design concern worth resolving before this surface gets used by scenario authors, plus a test gap and a small simplification:

1. **`tick` and `advance_time` disagree about what time it is.** `HarnessClient::tick()` and `drive_due_convergence()` still pass the synthetic `HARNESS_CONVERGENCE_SETTLED_AT_MS = 1_000_000` as `monotonic_ms` while `wall_ms` comes from the injected clock. Cutoff is `now.monotonic_ms >= pass.cutoff_monotonic_ms()`, so a single `tick` step forces every open pass past its cutoff regardless of virtual time — and hands the engine an incoherent `(monotonic = 1_000_000, wall = 0)` pair. The new test avoids `tick` entirely, so the interaction is neither tested nor documented. Details inline.

2. **The headline claim — global time advances while unselected participants stay paused — has no test.** The only engine-level test has one client, which is also the selected one. Inline.

3. **`advance_time` does three map lookups per label and adds a bespoke duplicate check** that no sibling step has. Inline.

Minor, no action needed unless you disagree: `ManualConvergenceClock::new(0, 0)` now replaces `SystemConvergenceClock` for *every* `EngineHarnessSubject` scenario, including ones that never call `advance_time`, so persisted `opened_wall_ms` becomes 0 (Unix epoch) across the board. Self-consistent within the engine, and arguably a determinism win, but it is a broader change than "existing `Tick` behavior and all existing vectors remain unchanged" suggests — worth a sentence in the README.

Also fine as-is, but a one-line comment would help: a selected client with no default group is silently skipped by the `has_default_group()` guard. That is the right semantics (nothing to converge), but the silence is easy to misread as a bug later.

PR #1203, opened 2026-07-30
Make inbound commit capability projections atomic

### LINKED ISSUE #794: cgka-engine: direct-ingest inbound commit mirrors the Marmot group record + self-cap cache OUTSIDE the merge transaction — a crash/SQLITE_BUSY between them permanently desyncs the record from MLS state
## Summary

On the **direct-ingest** inbound seam (`ingest_group_message`), only `merge_staged_commit` runs inside a `with_transaction`. The durable writes that mirror the merged MLS state back into the Marmot `Group` record and the self-capability cache run as **separate, un-transacted writes afterward**. A crash (SIGKILL/OOM) or a transient `SQLITE_BUSY` between the merge commit and the record `put_group` leaves **MLS state at epoch N+1 but the Marmot `Group` record still at epoch N** (stale roster / caps / profile). On the next open, `hydrate_one_stored_group` trusts the record's epoch and never reconciles it against `mls_group.epoch()` on this path, so the desync is durable and unreconciled for a group that then goes quiescent.

This contradicts the crate's own `AGENTS.md` rule "**Mirror every ingest invariant on every inbound seam**" and the `multi-step-state-changes.md` invariant ("compensate every applied step on the error path"). The two sibling apply paths were **deliberately** made atomic; the direct-ingest seam was not.

## Location

- `crates/cgka-engine/src/message_processor/ingest.rs`, `ingest_group_message`, `ProcessedMessageContent::StagedCommitMessage` arm:
  - merge inside a transaction: **lines 873–881** (`self.storage.with_transaction(|storage| { ... merge_staged_commit ... })?`)
  - record mirror **outside** the transaction: **lines 914–936** — `if let Ok(mut g) = self.storage.get_group(&group_id) { ... self.storage.put_group(&g)?; }` (the `get_group` read failure is **swallowed** by `if let Ok`)
  - self-cap refresh **outside** the transaction: **lines 944–950** (`cache_self_capabilities(...)`)
- Reconciliation gap: `crates/cgka-engine/src/engine.rs` `hydrate_one_stored_group` sets `epoch_manager` to `group.epoch` (the record's epoch) with no comparison against `mls_group.epoch()` on the non-pending path.

## Contrast (paths that ARE atomic)

- **Confirm path** — `crates/cgka-engine/src/publish.rs:137-149` bundles `get_group` + record mirror + `cache_self_capabilities` **inside** the same `with_transaction` as `merge_pending_commit`, and **propagates** the `get_group` read failure with an explicit comment: *"a transient backend error must roll back the whole transaction and stay retryable, never commit the merge + `Processed` write with a stale record."*
- **Convergence-apply path** — `crates/cgka-engine/src/openmls_projection.rs` `apply_openmls_canonicalization_result_inner` (the `#157/#424` fix) bundles `merge` + `update_group_record_from_replay` + dispositions into one transaction.

The direct-ingest seam is the odd one out.

## Concrete failure scenario

Two members race commits at epoch N. The victim ingests the winning inbound commit:
1. `merge_staged_commit` commits durably → MLS at N+1; `epoch_manager.set_stable(group, N+1)` runs in memory.
2. A transient `SQLITE_BUSY` fails `put_group` at line 935 (or the process is killed here).
3. Ingest returns `Err`; the stored message is left non-`Processed`, but MLS is durably at N+1. On retry, `current_epoch = N+1`, `msg_epoch = N`, `we_committed_from(N)` is false → the commit is marked `Failed`.
4. The Marmot record stays at **epoch N** forever (until an unrelated later commit overwrites it). `members()` (`group_lifecycle.rs`), `group_record()`, `feature_status`, and the reported epoch return **stale** data indefinitely, and the app never received the `EpochChanged` / `GroupStateChanged` events for the lost commit (they are pushed after `put_group`).

## Scope / severity note

Inbound **authorization** is not bypassed: admin checks read live MLS (`admins_of_group(&mls_group)`), not the stale record. So this is a correctness/consistency defect on the app-facing projection, not an auth bypass. It self-heals on the next successfully-applied commit, but is durable for a group that goes quiescent after the failing window — hence **MEDIUM** (arguably HIGH for the quiescent-group case).

## Suggested fix

Move the `get_group`/`put_group` record mirror and `cache_self_capabilities` into the **same** `with_transaction` as `merge_staged_commit` (as the confirm and convergence-apply paths already do), and **propagate** the `get_group` read failure instead of swallowing it via `if let Ok`. Add a parity test mirroring the confirm-path atomicity test.

## Dedup

Checked open + closed issues. No overlap. The `#157/#424` references are the *convergence*-path fix; the confirm path is `Sm1`/publish-lifecycle — neither covers the direct-ingest seam's record/self-cache mirror. #756 is unrelated (secure-delete on local wipe).

---
_Found during a methodical multi-crate code review; verified against source (`ingest.rs:873-950`, `publish.rs:96-163`)._

erskingardner:
## Summary

- move added-member and post-path self-capability cache writes onto the same transaction rail as direct inbound commit merge and group-record projection
- make stored-convergence replay rebuild added-member and self-capability projections atomically with canonical apply
- add a capability-write fault regression covering MLS/record/cache rollback, snapshot cleanup, convergence recovery, cache restoration, and duplicate redelivery
- document capability caches as commit-derived durable projections

## Relationship to #1184

While this PR was open, #1184 landed the overlapping group-record portion of #794: it made the inbound epoch/roster mirror atomic with the OpenMLS merge and added recovery scheduling after a failed record write. This branch is rebased on that implementation and keeps it intact.

The remaining gap was capability state. Direct ingest still cached newly added members before the transaction and refreshed self capabilities after it. A transient cache-write failure could therefore leave capability projections inconsistent with the MLS/group-record unit. Stored convergence also did not rebuild those caches when recovering the retained commit.

## Impact

An inbound commit now either advances MLS, the Marmot group record, and all capability projections together or rolls the complete operation back. Stored convergence can reconstruct the same capability state when it retries an abandoned direct apply.

## Validation

- `cargo test -p cgka-engine --features test-policy-overrides --test auto_commit_atomicity` — 5 passed
- `just fast-ci`
- original pre-rebase GitHub CI matrix was fully green

Fixes #794

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when applying group updates by keeping group state and capability data synchronized.
  * Failed updates now roll back cleanly, avoid leaving recovery data behind, and are automatically rescheduled for convergence.
  * Retried updates correctly rebuild capability information, while duplicate deliveries remain safely ignored.

* **Tests**
  * Added coverage for capability-cache failures during inbound group updates and subsequent recovery.

### DIFF
--- a/crates/cgka-engine/src/capability_manager.rs
+++ b/crates/cgka-engine/src/capability_manager.rs
@@ -79,6 +79,33 @@ pub(crate) fn cache_self_capabilities<S: StorageProvider>(
     Ok(())
 }
 
+/// Replay-side self-cache refresh when the caller has only the materialized
+/// OpenMLS group, not the engine's declared identity.
+///
+/// The own leaf still carries the authenticated Marmot credential identity;
+/// derive that identity and route through [`cache_self_capabilities`] so replay
+/// performs the same proof/profile validation and durable write as direct
+/// ingest.
+pub(crate) fn cache_own_capabilities_from_group<S: StorageProvider>(
+    storage: &S,
+    group_id: &GroupId,
+    mls_group: &MlsGroup,
+) -> Result<(), EngineError> {
+    let Some(leaf) = mls_group.own_leaf_node() else {
+        return Ok(());
+    };
+    let credential = BasicCredential::try_from(leaf.credential().clone())
+        .map_err(|e| EngineError::Backend(format!("credential: {e:?}")))?;
+    let self_id = MemberId::new(credential.identity().to_vec());
+    cache_self_capabilities(
+        storage,
+        group_id,
+        mls_group,
+        &self_id,
+        mls_group.ciphersuite(),
+    )
+}
+
 /// Cache capabilities extracted from a validated invitee's KeyPackage.
 /// Called from create_group / invite paths after capability-check passes.
 pub(crate) fn cache_from_key_packages<S: StorageProvider>(
--- a/crates/cgka-engine/src/message_processor/ingest.rs
+++ b/crates/cgka-engine/src/message_processor/ingest.rs
@@ -1276,13 +1276,6 @@ impl<S: StorageProvider> Engine<S> {
                     let before_message_retention =
                         crate::app_components::message_retention_seconds_of_group(&mls_group)?;
                     self.retain_current_epoch_snapshot_for_group(&group_id)?;
-                    // Extract capabilities from Add proposals before the
-                    // staged commit is consumed by merge.
-                    crate::capability_manager::cache_from_staged_commit(
-                        &self.storage,
-                        &group_id,
-                        &staged,
-                    )?;
                     // Merge and mirror the Marmot record in one durable unit, the
                     // same rule `do_confirm_published` applies to its own merge.
                     // Session open seeds the epoch state machine FROM that
@@ -1298,6 +1291,13 @@ impl<S: StorageProvider> Engine<S> {
                         message_retention_seconds: after_message_retention,
                     } = match self.storage.with_transaction(
                         |storage| -> Result<AppliedCommitProjection, EngineError> {
+                            // The staged commit is the only public source for
+                            // newly added members' KeyPackage capabilities.
+                            // Cache them on the same transaction rail as the
+                            // merge and every other durable projection.
+                            crate::capability_manager::cache_from_staged_commit(
+                                storage, &group_id, &staged,
+                            )?;
                             let tx_provider = EngineOpenMlsProvider::<S>::new(
                                 &self.crypto,
                                 storage.mls_storage(),
@@ -1342,6 +1342,16 @@ impl<S: StorageProvider> Engine<S> {
                                 g.removed = true;
                             }
                             storage.put_group(&g)?;
+                            // A commit path update may change our own leaf.
+                            // Feature-status correctness requires this cache to
+                            // advance atomically with MLS and the group record.
+                            crate::capability_manager::cache_self_capabilities(
+                                storage,
+                                &group_id,
+                                &mls_group,
+                                self.identity.self_id(),
+                                self.ciphersuite,
+                            )?;
                             Ok(AppliedCommitProjection {
                                 epoch: after,
                                 remaining_member_ids: after_ids,
@@ -1413,15 +1423,6 @@ impl<S: StorageProvider> Engine<S> {
                     // index so the new nostr_group_id resolves (prior id retained
                     // for the overlap window). No-op when routing was unchanged.
                     self.reindex_transport_group_id(&group_id);
-                    // Refresh self-cache since our own leaf may have been
-                    // updated by the commit's path.
-                    crate::capability_manager::cache_self_capabilities(
-                        &self.storage,
-                        &group_id,
-                        &mls_group,
-                        self.identity.self_id(),
-                        self.ciphersuite,
-                    )?;
                     self.prune_fork_recovery_for_group(&group_id)?;
 
                     if let Some((source_epoch, winner, invalidated, invalidated_commit_id)) =
--- a/crates/cgka-engine/tests/auto_commit_atomicity.rs
+++ b/crates/cgka-engine/tests/auto_commit_atomicity.rs
@@ -186,11 +186,37 @@ impl PutGroupFault {
     }
 }
 
+/// One-shot fault that fails the Nth capability-cache write after arming.
+///
+/// Inbound Add commits write the added member first and refresh self second,
+/// which lets the regression fail specifically at the post-merge self-cache
+/// refresh.
+#[derive(Clone, Default)]
+struct CapabilityWriteFault(Arc<AtomicUsize>);
+
+impl CapabilityWriteFault {
+    fn arm_on_call(&self, call: usize) {
+        assert!(call > 0);
+        self.0.store(call, Ordering::SeqCst);
+    }
+
+    fn should_fail(&self) -> bool {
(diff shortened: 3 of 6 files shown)


crates/cgka-engine/src/capability_manager.rs
```diff
@@ -79,6 +79,33 @@ pub(crate) fn cache_self_capabilities<S: StorageProvider>(
...
+    storage: &S,
+    group_id: &GroupId,
+    mls_group: &MlsGroup,
+) -> Result<(), EngineError> {
+    let Some(leaf) = mls_group.own_leaf_node() else {
+        return Ok(());
+    };
+    let credential = BasicCredential::try_from(leaf.credential().clone())
+        .map_err(|e| EngineError::Backend(format!("credential: {e:?}")))?;
+    let self_id = MemberId::new(credential.identity().to_vec());
```
erskingardner, line 99: Non-blocking, but worth an explicit comment: deriving `self_id` from the same leaf you then pass to `cache_self_capabilities` makes that function's documented safety check vacuous.

`cache_self_capabilities`'s doc block says its job is to "cross-check the OpenMLS-reported leaf credential against the engine's declared `self_id`" and to surface a mismatch as `EngineError::Backend` "rather than silently caching under a wrong member id". Here `leaf_id == self_id` by construction, so the `if &leaf_id != self_id` branch is unreachable and the replay path does exactly the silent-cache-under-whatever-the-leaf-says behavior the check exists to prevent.

That is probably acceptable given `apply_openmls_canonicalization_result` is a free function with only `storage` in hand, but the current doc comment ("the caller has only the materialized OpenMLS group") explains *why you can't pass* the engine identity without stating that the guarantee is therefore dropped. Either thread the engine `self_id` down, or say plainly here that the identity cross-check is intentionally a no-op on this seam.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -1265,41 +1277,120 @@ impl<S: StorageProvider> Engine<S> {
                     let before_message_retention =
                         crate::app_components::message_retention_seconds_of_group(&mls_group)?;
                     self.retain_current_epoch_snapshot_for_group(&group_id)?;
```
erskingardner, line 1278: Adjacent leak the new error arm doesn't cover: `recovery_snapshot` was already created ~20 lines above, so a failure here (`?`) abandons the apply and leaks the fork-recovery snapshot — the exact condition the new `Err` branch below now cleans up. Same for `message_retention_seconds_of_group` on the line above.

Pre-existing, and much narrower than the transaction window you fixed, but since this PR is specifically about not leaving snapshots behind on an abandoned inbound apply, it's a bit odd for the two statements immediately preceding the transaction to still take the leaking path.

crates/cgka-engine/tests/auto_commit_atomicity.rs
```diff
@@ -679,6 +738,184 @@ async fn auto_commit_record_write_failure_leaves_no_torn_group_record() {
...
+            .is_none(),
+        "the added-member cache write must roll back with the failed self refresh"
+    );
+    let snapshots_after: std::collections::HashSet<String> = loser_storage
+        .list_group_snapshots(&group_id)
+        .unwrap()
+        .into_iter()
+        .collect();
+    assert!(
+        snapshots_after.is_subset(&snapshots_before),
```
erskingardner, line 1100: Nit: `is_subset` only catches *new* snapshots. It also passes if the failed apply over-deleted a pre-existing snapshot (e.g. the retained epoch anchor created just before the transaction), which would be a different but equally real bug. `assert_eq!(snapshots_after, snapshots_before)` pins both directions and reads the same.

erskingardner (commented):
Recommend merge (posting as a comment — GitHub will not let me approve my own PR). This does exactly what mdk#794 asked for and nothing speculative: the record mirror + self-cap refresh move inside the `merge_staged_commit` transaction, the `if let Ok(...)` swallow becomes a propagating `get_group(...)?`, and the shape now matches the already-atomic `publish.rs` confirm path. I traced the surrounding invariants and they hold:

- `mls_group` is loaded locally per `ingest_group_message` call (`ingest.rs:254`), so the in-memory merge that a rolled-back transaction leaves behind is discarded with the `Err` return — no in-memory/durable split.
- The abandoned-apply snapshot release is complete cleanup: `ForkRecoveryManager::create_snapshot` keeps no in-memory bookkeeping beyond a counter, so the comment's claim is accurate.
- `schedule_pending_convergence_group` is in-memory only, but the crash case is still covered — `hydrate_one_stored_group` re-arms via `has_pending_convergence_inputs` on open (`engine.rs:1444-1529`).
- The new replay-side `cache_from_staged_commit` also runs under probe replays via `replay_openmls_messages_prevalidated`'s `SnapshotRollbackGuard`. That is safe because SQLite group snapshots do capture and restore `cgka_member_capabilities` (`storage-sqlite/src/storage/snapshots/{capture,restore}.rs`), so a losing candidate branch cannot poison the cache.

One scope note worth calling out explicitly rather than blocking on: the `openmls_projection.rs` changes fix a **second, independent pre-existing gap** — before this PR the convergence-replay path performed *zero* capability caching (`cache_from_staged_commit` / `cache_self_capabilities` had no callers in that file). That is a genuine bug fix and it is load-bearing for the recovery story the new test asserts, so bundling it is defensible, but it is not what mdk#794 describes and it is currently only covered transitively through the fault-injection test. A direct convergence-path assertion would age better.

The rest are non-blocking nits, inline.

PR #1204, opened 2026-07-30
Feat/incident replay classifier parity

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Added quarantine and advisory reporting for unrecoverable engine halts, including affected engines and halt reasons.
  * Added fail-closed incident routing with deterministic precedence, fallback recovery, one primary outcome, and secondary advisories.
  * Updated incident replay output to show the primary outcome and applicable advisories.
* **Bug Fixes**
  * Ensured verified repairs clear only the corresponding halt and group.
  * Prevented duplicate halt or liveness advisories when already represented by the primary outcome.
* **Documentation**
  * Clarified halt, quarantine, repair, and routing behavior in incident-replay and audit-event documentation.

### DIFF
--- a/crates/incident-replay/src/classify.rs
+++ b/crates/incident-replay/src/classify.rs
@@ -75,12 +75,19 @@ pub enum QuarantineReason {
     TruncatedProjections,
     /// A fork resolution's winning snapshot was missing — unreproducible.
     MissingSnapshot,
-    /// Engines recorded halting unrecoverably: the group is blocked on those
-    /// engines until a verified repair clears the marker. Unlike
+    /// Engines recorded halting unrecoverably and are *still* halted at the end
+    /// of the export: no verified repair cleared the marker afterwards. Unlike
     /// [`QuarantineReason::EpochDivergence`], this is not an inference from lag —
     /// the engine itself recorded that it stopped — so it needs no cross-pull
     /// confirmation and it is a diagnosis rather than a re-pull target. There is
     /// still nothing to replay: a halt is a client failure, not a branch contest.
+    ///
+    /// "Still halted" is decided per `(engine, group)` by comparing the newest
+    /// halt row against the newest verified-repair row
+    /// (`epoch_state_changed { new_state: "stable", reason: "join_welcome_repair"
+    /// }`); surviving halts then fold back per engine. See [`HaltLifecycle`] for
+    /// why newest-wins needs no global ordering and why missing evidence keeps the
+    /// halt.
     UnrecoverableHalt {
         /// Every engine that recorded a halt, in engine-id order.
         engines: Vec<HaltedEngine>,
@@ -285,22 +292,95 @@ pub fn halt_advisory(export: &AgentStateExport) -> Option<QuarantineReason> {
     unrecoverable_halt(export)
 }
 
+/// One group's halt lifecycle on one engine, folded from the event log.
+///
+/// `Unrecoverable` is per-group state with exactly one legal exit, so the halt
+/// and its repair are a two-event lifecycle rather than two independent facts —
+/// hence the newest of each, not a count.
+#[derive(Default)]
+struct HaltLifecycle<'a> {
+    /// Every distinct reason the engine gave for halting this group.
+    reasons: BTreeSet<&'a str>,
+    /// The newest halt row's timestamp, when the halt rows carry one.
+    last_halt_ms: Option<u64>,
+    /// The newest verified-repair row's timestamp, when it carries one.
+    last_repair_ms: Option<u64>,
+}
+
+impl HaltLifecycle<'_> {
+    /// Whether the halt still stands at the end of the export.
+    ///
+    /// Newest-repair-strictly-after-newest-halt is the whole rule, and it needs
+    /// no global event ordering because a live halt is *continuously
+    /// re-asserted*: opening a session over a durably halted group re-emits
+    /// `hydrate_unrecoverable_group` (`cgka-engine/src/engine.rs`) and every
+    /// convergence attempt against one emits `already_unrecoverable`
+    /// (`cgka-engine/src/distributed_convergence.rs`). A halt that outlived a
+    /// repair therefore leaves rows *after* it, and one that did not, does not.
+    /// Both rows also come from a single engine — one clock, one recorder file —
+    /// so none of rule 6's cross-device [`CATCH_UP_GRACE_MS`] skew allowance
+    /// applies and a strict comparison is sound. (`seq` must never be substituted
+    /// for the clock: it restarts at 0 on every recorder open *within the same
+    /// file*, so it does not order rows across a restart — the exact boundary a
+    /// durable halt survives. See `docs/marmot-architecture/audit-logging.md`.)
+    ///
+    /// Absence fails closed in both directions: an unrepaired halt keeps its
+    /// quarantine, and so does a halt whose evidence cannot be ordered. Untimed
+    /// rows never clear anything — arming rule 5 needs no timestamps, but
+    /// *clearing* it does, and guessing here would turn a real halt into a
+    /// healthy verdict. This is also why untimed fixtures and
+    /// pre-instrumentation exports classify exactly as before.
+    fn still_halted(&self) -> bool {
+        if self.reasons.is_empty() {
+            return false;
+        }
+        match (self.last_halt_ms, self.last_repair_ms) {
+            (Some(halt), Some(repair)) => repair <= halt,
+            _ => true,
+        }
+    }
+}
+
 /// The gate for engines that recorded halting unrecoverably.
 ///
 /// Like the liveness gate it fires only on positive evidence, but it needs far
-/// less of it: a halt is self-reported, so no timestamps are required to order it
-/// and no threshold is needed to distinguish it from noise. It does require an
-/// `engine_id` — the report is per-engine, and an unattributable halt could not
-/// be named — which also keeps untimed synthetic fixtures and pre-instrumentation
-/// exports classifying exactly as before.
+/// less of it to arm: a halt is self-reported, so no threshold is needed to
+/// distinguish it from noise. It does require an `engine_id` — the report is
+/// per-engine, and an unattributable halt could not be named.
+///
+/// The gate is *cleared* per `(engine_id, group_ref)` rather than per engine,
+/// because that is the granularity the halt actually has: the engine's state
+/// machine holds `Unrecoverable` for one group, and a repair in another group
+/// proves nothing about it. Surviving halts then fold back per engine, so an
+/// engine halted in two groups is still one halted engine with the union of its
+/// reasons — the report shape rule 5 has always had.
 fn unrecoverable_halt(export: &AgentStateExport) -> Option<QuarantineReason> {
-    let mut halted: BTreeMap<&str, BTreeSet<&str>> = BTreeMap::new();
+    let mut lifecycles: BTreeMap<(&str, Option<&str>), HaltLifecycle> = BTreeMap::new();
     for event in &export.events {
-        if let (Some(engine_id), Some(reason)) = (
-            event.engine_id.as_deref(),
-            event.kind.unrecoverable_halt_reason(),
-        ) {
-            halted.entry(engine_id).or_default().insert(reason);
+        let Some(engine_id) = event.engine_id.as_deref() else {
+            continue;
+        };
+        let group_scope = (engine_id, event.group_ref.as_deref());
+        if let Some(reason) = event.kind.unrecoverable_halt_reason() {
+            let lifecycle = lifecycles.entry(group_scope).or_default();
+            lifecycle.reasons.insert(reason);
+            lifecycle.last_halt_ms = lifecycle.last_halt_ms.max(event.wall_time_ms);
+        } else if event.kind.is_verified_repair() {
+            // Recorded even when no halt has been seen yet: the fold is over
+            // timestamps, not file order, so a repair may legitimately arrive
+            // before the halt it postdates.
+            let lifecycle = lifecycles.entry(group_scope).or_default();
+            lifecycle.last_repair_ms = lifecycle.last_repair_ms.max(event.wall_time_ms);
+        }
+    }
+
+    let mut halted: BTreeMap<&str, BTreeSet<&str>> = BTreeMap::new();
+    for ((engine_id, _group_ref), lifecycle) in &lifecycles {
+        if lifecycle.still_halted() {
+            halted
+                .entry(engine_id)
+                .or_default()
+                .extend(lifecycle.reasons.iter().copied());
         }
     }
 
--- a/crates/incident-replay/src/classify.rs
+++ b/crates/incident-replay/src/classify.rs
@@ -4,7 +4,7 @@
 //! this verdict, so a healthy export yields zero vectors and a clean exit rather
 //! than a crash. Built incrementally, one rule per behaviour.
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::fmt;
 
 use crate::export::{AgentStateExport, EventKind};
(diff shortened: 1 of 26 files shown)


crates/incident-replay/src/classify.rs
```diff
@@ -223,14 +265,55 @@ pub fn classify(export: &AgentStateExport) -> Verdict {
...
+/// be named — which also keeps untimed synthetic fixtures and pre-instrumentation
+/// exports classifying exactly as before.
+fn unrecoverable_halt(export: &AgentStateExport) -> Option<QuarantineReason> {
+    let mut halted: BTreeMap<&str, BTreeSet<&str>> = BTreeMap::new();
+    for event in &export.events {
+        if let (Some(engine_id), Some(reason)) = (
+            event.engine_id.as_deref(),
+            event.kind.unrecoverable_halt_reason(),
+        ) {
+            halted.entry(engine_id).or_default().insert(reason);
```
erskingardner: **[P1] Clear a halt after the verified repair transition**\n\nThis aggregation makes an unrecoverable row sticky for the lifetime of the export. The engine repair path in `group_lifecycle.rs` explicitly emits a later `epoch_state_changed { new_state: "stable", reason: "join_welcome_repair" }` after `repair_to_stable`, so a normal halt → verified Welcome repair history will still classify as `UnrecoverableHalt` (or emit a halt advisory) here even though the durable marker has been cleared. That contradicts this PR’s “until a verified repair clears the marker” contract and can quarantine an already-repaired engine. Please model the ordered repair transition and clear that engine’s accumulated halt state, with a regression fixture containing an unrecoverable event followed by `join_welcome_repair` that produces no halt finding.

erskingardner (changes requested):
The halt detection is valuable, but it currently treats repaired engines as permanently halted. Please address the blocking lifecycle issue in the inline comment before merge. I reviewed exact head 81b63aa; all current CI checks are green.

crates/incident-replay/src/classify.rs
jgmontoya: Confirmed, and fixed in eef40525. You read the code right on both counts: the aggregation had no
clearing path, and group_lifecycle.rs does emit epoch_state_changed { new_state: "stable",
reason: "join_welcome_repair" } right after repair_to_stable — strings verbatim. Reproduced:
a halt followed by that row still printed the quarantine.

Two things nailed down before writing the fix:

The clearing condition matches the REASON, not the state. repair_to_stable is the only legal
exit from Unrecoverable (engine_state.rs) and do_join_welcome is its only caller, so
join_welcome_repair is a one-path, one-string signal. Bare "stable" is not: publish_confirmed,
publish_failed, join_welcome, founding_create and both stage-failed reasons all reach it — a
"later stable row clears" rule would clear essentially every halt.

And it had to become group-scoped, which the gate wasn't. We never modelled group_ref, so the
fold was engine-global — harmless while nothing cleared, but with clearing added, an engine
halted in group X then joining group Y by Welcome would have cleared X. The lifecycle is now
keyed (engine_id, group_ref); 100% of real export lines carry group_ref, so this is active, not
theoretical. Output shape unchanged.

On the ordered transition: seq is out (recorder-local, restarts at 0 on every recorder open
within the same file — our own schema doc warns against it) and export line order is an
unverified server guarantee. What holds is newest-halt vs newest-repair on wall_time_ms, per
(engine, group) — sound because a live halt is continuously re-asserted (session open re-emits
hydrate_unrecoverable_group; every convergence attempt emits already_unrecoverable), so the
newest row genuinely reflects the marker's state, and both rows come from one engine, one clock,
one recorder file — none of rule 6's cross-device grace applies. That also gets
halt → repair → halt right, which a presence-only downgrade cannot (it trades your false
positive for a false negative on the worse side). Untimed evidence never clears: fail-closed,
zero change to existing fixtures.

Your regression fixture is in (healthy-halt-cleared-by-repair: verdict AND advisory both clear),
plus three guards so the clearing can't over-reach: halt-after-repair still quarantines,
a repair in a different group still quarantines, an untimed repair still quarantines. The
ordering test is mutation-checked — a presence-only implementation fails it. Both real field
exports produce byte-identical output pre/post fix (they contain no join_welcome_repair; the
halted device there was never repaired).

crates/incident-replay/src/route.rs
```diff
@@ -0,0 +1,317 @@
...
+    pub outcome: Outcome,
+    pub advisories: Vec<Advisory>,
+}
+
+impl fmt::Display for Outcome {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Outcome::Healthy => f.write_str("healthy: 0 vectors"),
+            Outcome::Accepted(artifact) => {
+                write!(f, "accepted: {}", artifact.vector.scenario_name)
```
erskingardner, line 109: Two of these four arms never execute. `main.rs` uses `Display` only for `Healthy | Quarantine`; `Accepted` goes to `persist_or_report`, which prints `accepted ({fidelity:?}, {sensitivity:?}): wrote ...`, and `InfrastructureFailure` is formatted in `main.rs` as `eprintln!("error: {reason}")` — byte-identical to the arm below.

So `accepted: {scenario_name}` is a format no user ever sees that also contradicts the one they do see (and the module doc's "one primary line — `healthy:`, `quarantine:`, or `accepted:`"), and the `error:` arm is a copy of the CLI's string. Either make `Display` the single source of the primary line (with `persist_or_report` reporting only where it wrote) and print it uniformly, or drop `Display for Outcome` and keep all formatting in the CLI. Right now it is half of each.

crates/incident-replay/src/route.rs
```diff
@@ -0,0 +1,317 @@
...
+    {
+        return Outcome::Quarantine { reason: superseded };
+    }
+    match fork_route(export, source_format) {
+        Ok(artifact) => {
+            advisories.push(Advisory {
+                label: SUPERSEDED_ROUTE,
+                detail: format!("contested convergence could not be replayed: {superseded}"),
+            });
+            Outcome::Accepted(artifact)
```
erskingardner, line 273: The `superseded route` finding is stdout-only. It goes into `Routing.advisories`, which `main.rs` prints, but nothing about it reaches `IncidentScenarioArtifactV1` — the thing actually written to disk owner-only and passed around as evidence. Someone opening `membership-fork-recovery-incident.incident.v1.json` sees a clean accepted archetype with no trace that the same export carried a higher-precedence contested convergence at the same epoch that could not be replayed.

That sits awkwardly next to this PR's own standard for the envelope ("an artifact that cannot point at its own source rows is not publishable evidence"). `unavailable_fields: Vec<UnavailableEvidenceV1 { field, reason }>` already exists and would carry it with no schema change, e.g. `{ field: "contested_convergence_replay", reason: superseded }`. Deliberate omission, or worth threading through?

crates/incident-replay/src/classify.rs
```diff
@@ -223,14 +272,128 @@ pub fn classify(export: &AgentStateExport) -> Verdict {
...
+    let mut lifecycles: BTreeMap<(&str, Option<&str>), HaltLifecycle> = BTreeMap::new();
+    for event in &export.events {
+        let Some(engine_id) = event.engine_id.as_deref() else {
+            continue;
+        };
+        let group_scope = (engine_id, event.group_ref.as_deref());
+        if let Some(reason) = event.kind.unrecoverable_halt_reason() {
+            let lifecycle = lifecycles.entry(group_scope).or_default();
+            lifecycle.reasons.insert(reason);
+            lifecycle.last_halt_ms = lifecycle.last_halt_ms.max(event.wall_time_ms);
```
erskingardner, line 367: `Option::max` orders `None` below every `Some`, so a halt row without `wall_time_ms` is silently dropped as soon as any other halt row for the same `(engine, group)` carries one. halt(untimed), halt(T1), repair(T2 > T1) therefore clears the halt even though the untimed row may be the newest evidence — the fail-open direction `still_halted`'s doc explicitly rules out ("a halt whose evidence cannot be ordered" keeps its quarantine; "untimed rows never clear anything").

Only reachable on partially-instrumented input, since `AuditEvent.wall_time_ms` is a non-optional `u64` in `marmot-forensics` — so this is either a one-line guard (remember that an untimed halt row was seen and keep the halt) or a doc correction saying the claim holds only when halt rows are uniformly timed. Both fine; the current pairing of code and doc is not.

crates/incident-replay/src/export.rs
```diff
@@ -11,6 +11,39 @@ use std::collections::BTreeMap;
...
+/// The healthy epoch-machine state. A halt clears only through a transition into
+/// it, but this state alone is *not* the clearing signal — it is by far the
+/// engine's most ordinary state, entered from `publish_confirmed`,
+/// `publish_failed`, `join_welcome`, `founding_create`,
+/// `auto_commit_stage_failed`, and `update_group_data_stage_failed`, none of
+/// which repairs anything. `publish_confirmed` alone is one of the commonest rows
+/// in a healthy export, so matching the bare state would clear every halt on the
+/// engine's next successful publish. Hence the pairing with
+/// [`VERIFIED_REPAIR_REASON`]: the reason is what identifies the repair.
+const STABLE_STATE: &str = "stable";
```
erskingardner, line 30: Non-blocking, but the "bare `stable` is the ordinary state, only `join_welcome_repair` marks the real exit" argument is now stated in full in six-plus places: here, `VERIFIED_REPAIR_REASON`, `is_verified_repair`, `QuarantineReason::UnrecoverableHalt`, `HaltLifecycle::still_halted`, AGENTS.md rule 5 *and* its rejected-designs entry, `audit-logging.md`, plus two test comments. Each copy enumerates the entering reasons or the single-caller chain a little differently, so the day `repair_to_stable` gets a second caller they start drifting one at a time and the reader has no way to tell which copy is current.

The argument only needs one home — AGENTS.md rule 5 reads like it — with the code docs saying what the constant *is* and pointing there.

crates/incident-replay/src/export.rs
```diff
@@ -228,6 +333,55 @@ impl EventKind {
         matches!(self, EventKind::ForkResolution { .. })
     }
 
+    /// Why the engine recorded halting unrecoverably, or `None` when this event
+    /// is not such a halt.
+    ///
+    /// A halt is the engine's own statement that it stopped and will stay
+    /// stopped until repaired, so — unlike the inferential liveness gate — it
+    /// needs no corroborating timestamps to be believed.
+    pub fn unrecoverable_halt_reason(&self) -> Option<&str> {
```
erskingardner, line 342: Because the gate matches any `phase: unrecoverable` with no reason whitelist (deliberately, and I agree with that), `already_unrecoverable` — the row `distributed_convergence.rs:1010` emits on *every* convergence attempt against an already-halted group — lands in `HaltedEngine.reasons` beside the actual cause. On a real export the operator's first line reads `engine-x (already_unrecoverable, hydrate_unrecoverable_group)`, where the leading entry is re-assertion noise rather than a diagnosis.

It has to keep arming the gate (an export window containing only re-assertions is still a live halt), so this is purely presentational — e.g. report it only when it is the sole reason. Low priority; raising it because the reason list is the part an operator reads first.

erskingardner (approved):
Approving. The blocker from my previous review (append-only halt state) is fixed, and I re-verified the load-bearing claims against the engine rather than taking the commit messages for it:

- `repair_to_stable` really is the only exit from `Unrecoverable` (`crates/traits/src/engine_state.rs:392`), and `crates/cgka-engine/src/group_lifecycle.rs:1216` emits `epoch_state_changed { new_state: "stable", reason: "join_welcome_repair" }` unconditionally right after it. Reason-keyed clearing is correct.
- `group_ref` is a top-level `AuditEvent` field (`crates/marmot-forensics/src/audit.rs:117`) and `ndjson.rs` deserializes the whole `AuditEvent`, so the new `(engine_id, group_ref)` scoping is live for the stream export too, not just the document shape.
- The continuous-re-assertion argument that makes newest-wins sound holds: `engine.rs:1477` re-emits `hydrate_unrecoverable_group` on every session open over a halted group, and `distributed_convergence.rs:1010` emits `phase: unrecoverable, reason: already_unrecoverable` on every convergence attempt against one.
- `missing_retained_anchor` is indeed the reason-less `convergence_run_state` halt (`distributed_convergence.rs:1260-1266`), so the `error_kind` fallback is load-bearing rather than defensive.
- The catalogue-completeness claim is exactly right: 41 `AuditEventKind` variants on master, and precisely those twelve have no section in `audit-logging.md`.

Five non-blocking notes inline. The one I would actually like an answer on is the evidence gap: the `superseded route` finding exists only on stdout, not in the artifact that gets persisted and shared.

Scope, non-blocking and not worth re-cutting now: this is five separable changes in one PR (halt gate; `convergence_run_state` / `convergence_pass_discarded` tip evidence; CLI-to-library route extraction; fall-through; audit-logging catalogue). The commits are clean enough to review one at a time, but the halt-clearing bug is the sort of thing a narrower PR surfaces sooner.

crates/incident-replay/src/route.rs
jgmontoya: Correct, and worse than dead code: `accepted: {scenario_name}` is unreachable *and* contradicts the line users actually see. Verified both arms — `main.rs:81` routes `Accepted` to `persist_or_report` and formats `InfrastructureFailure` itself, and no test formats an `Outcome` either, so neither arm has a caller.

I agree with your first option: `Display` becomes the single source of the primary line, `main.rs` prints it uniformly for all four variants, and `persist_or_report` is reduced to reporting *where* it wrote (with fidelity and sensitivity carried on the primary line rather than duplicated into it). That also makes the module doc's "one primary line" contract enforced by construction instead of by convention.

crates/incident-replay/src/route.rs
jgmontoya: Not deliberate — you found a real gap, and you are right that it sits badly next to this PR's own evidence standard. Confirmed the mechanism: `fall_through_to_fork` pushes the advisory and returns the artifact exactly as `fork_route` built it, so the finding lives only in `Routing.advisories` and dies with stdout. The persisted envelope shows a clean accepted archetype.

`unavailable_fields` is the right home and needs no schema change, as you say — `{ field: "contested_convergence_replay", reason: <the superseded reason> }`. The same argument applies to the `fallback route` advisory on the quarantine path, so I would thread both rather than just the accepted one.

crates/incident-replay/src/classify.rs
jgmontoya: Accurate, and it is the fail-open direction, so I want the guard rather than the doc correction: an untimed halt row is evidence that cannot be ordered, and the documented contract is that such a halt keeps its quarantine. Reachability is exactly as you scoped it — `wall_time_ms` is a non-optional `u64` in `marmot-forensics`, but `export.rs` models it `Option<u64>` on purpose, so mixed-instrumentation and third-party input get there.

Fix is the one-liner you describe: a `saw_untimed_halt` flag on `HaltLifecycle`, and `still_halted()` returns true whenever it is set. Fixture alongside it: halt(untimed) + halt(T1) + repair(T2 > T1) must stay quarantined. I will pick up CodeRabbit's adjacent `last_halt_ms == last_repair_ms` tie fixture in the same commit, since it pins the other boundary of the same comparison.

crates/incident-replay/src/export.rs
jgmontoya: Agreed, and the drift scenario is the right reason: the day `repair_to_stable` gets a second caller, the copies that enumerate the entering reasons and the copies that assert "the only caller" go stale independently and nothing marks which is current.

AGENTS.md rule 5 is the home. The code docs get reduced to what the constant *is* — `STABLE_STATE` is the healthy state, `VERIFIED_REPAIR_REASON` is the reason stamped on the legal exit — with the reasoning pointed at, not restated. Rule 5's rejected-designs entry keeps the "why not bare `stable`" argument, since that is what it is for.

crates/incident-replay/src/export.rs
jgmontoya: Confirmed, including the ordering: `reasons` is a `BTreeSet`, so `already_unrecoverable` sorts ahead of `hydrate_unrecoverable_group` and the re-assertion is literally the first thing the operator reads.

Your framing of the constraint is the one I want to keep — it must go on arming the gate, because a window of nothing but re-assertions is still a live halt — so the fix is presentational: demote `already_unrecoverable` in the reported reason list and report it only when it is the sole reason (an only-re-assertions window still needs *something* to name).

PR #1205, opened 2026-07-30
Add explicit simulator outbound publication lifecycle

### LINKED ISSUE #1207: Remove legacy simulator publication lifecycle after Scenario IR v2
## Context

PR #1205 adds the explicit `ConvergenceSubject` outbound polling and acknowledgement contract. Existing ScenarioSpec v1 runners deliberately retain their manual `ConfirmPending` / `FailPending` behavior through `EngineHarnessSubject::new_legacy_compatible` until portable outbound operations exist.

That compatibility path must be temporary. Keeping both publication lifecycles indefinitely would allow their semantics and coverage to drift.

## Scope

- add portable Scenario IR v2 operations for polling exact outbound artifacts and acknowledging typed publication outcomes
- migrate built-in scenario and vector runners from `new_legacy_compatible` to the explicit outbound lifecycle
- port existing v1 fixtures or provide an explicit versioned compatibility translation
- remove `new_legacy_compatible` and the `external_publication_lifecycle` branching from the harness
- make adapter-refinement tests prove equivalent publication behavior across the engine subject and later retained-relay/process adapters
- update the convergence reliability plan when the legacy path is removed

## Acceptance criteria

- no built-in runner depends on direct `ConfirmPending` / `FailPending` publication control
- the legacy-compatible constructor and dual client lifecycle branches are deleted
- unsupported adapters fail outbound-capability preflight rather than silently returning no work
- existing publication, rollback, exposure, queued-intent, and cross-client pending-collision regressions pass through portable scenario operations
- the full simulator suite and `just fast-ci` pass

erskingardner:
## Summary

- remove the simulator's legacy auto-confirm publication lifecycle and use one explicit outbound poll/ack contract for every subject
- cut ScenarioSpec over to v2 with typed `acknowledge_outbound` steps, stable publication labels, artifact selectors, and `accepted` / `reached_no_endpoint` outcomes
- migrate all repository-owned vectors, generated scenario families, canonical tests, and incident-replay synthesis to ScenarioSpec v2
- preserve atomic commit/Welcome rollback, reject definite failure after recipient exposure, and keep Welcome outcomes independent after commit acceptance
- update simulator documentation and the convergence reliability tracking plan for the clean cutover

## Why

Retained-relay, process, CLI, and app-runtime adapters need to drive publication progress without reaching into the in-memory transport queue. Keeping a separate simulator-only lifecycle would let scenarios pass through behavior that those adapters cannot reproduce and would make the portable scenario contract ambiguous.

This PR now has one black-box lifecycle: poll exact transport-ready artifacts, apply a typed outcome to their opaque IDs, and observe the resulting engine state. No production convergence-engine or Marmot protocol behavior is changed; the changes are confined to the simulator contract, its fixtures/tests, and incident-replay scenario synthesis.

## Important semantics

- polling is non-destructive and preserves emission order
- repeating the same acknowledgement is idempotent; contradictory outcomes are rejected
- labelled acknowledgements fail when they match no artifact, catching stale or misspelled scenario references
- unlabelled acknowledgements operate on the currently matching artifact set and support engine-generated work
- an accepted state-bearing commit confirms staged MLS state
- definite commit failure retracts the complete unexposed commit/Welcome publication before rollback
- recipient exposure prevents partial transport mutation or rollback
- failed regenerated queued intents are re-armed; accepted retries retire them
- zero-fanout group creation confirms without inventing a synthetic outbound artifact

## Migration

ScenarioSpec v1 is intentionally rejected after this repository-owned cutover. All checked-in executable vectors now declare ScenarioSpec v2. Existing `.v1.json` filenames continue to describe their vector/family envelope versions, not the embedded ScenarioSpec schema.

Closes #1207

## Validation

- `cargo test -p incident-replay`
- `just fast-ci`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added ordered, client-scoped outbound publication polling.
  * Added typed acknowledgements for accepted and no-endpoint outcomes.
  * Added safe rollback, exposure checks, idempotent acknowledgements, and queued-intent retries.
  * Exposed outbound publication capabilities through the simulator interface.

* **Documentation**
  * Updated simulator guidance and scenario coverage for outbound lifecycle behavior.
  * Migrated scenarios and fixtures to ScenarioSpec v2.
  * Updated reliability planning and milestone status for the public outbound lifecycle.

### DIFF
--- a/crates/cgka-conformance-simulator/src/scenario.rs
+++ b/crates/cgka-conformance-simulator/src/scenario.rs
@@ -298,9 +298,12 @@ pub async fn run_scenario_report_with_storage_mode(
     expected_trace: Option<ScenarioTrace>,
     storage_mode: HarnessStorageMode,
 ) -> Result<ScenarioReport, ScenarioRunError> {
-    let mut subject =
-        EngineHarnessSubject::new(&spec.clients, ProtocolProfile::Legacy, storage_mode)
-            .map_err(subject_setup_error)?;
+    let mut subject = EngineHarnessSubject::new_legacy_compatible(
+        &spec.clients,
+        ProtocolProfile::Legacy,
+        storage_mode,
+    )
+    .map_err(subject_setup_error)?;
     run_scenario_report_inner(spec, expected_trace, vec![], None, &mut subject).await
 }
 
@@ -324,9 +327,12 @@ pub async fn run_scenario_report_with_outcomes_and_storage_mode(
     expected_outcomes: Vec<TraceExpectation>,
     storage_mode: HarnessStorageMode,
 ) -> Result<ScenarioReport, ScenarioRunError> {
-    let mut subject =
-        EngineHarnessSubject::new(&spec.clients, ProtocolProfile::Legacy, storage_mode)
-            .map_err(subject_setup_error)?;
+    let mut subject = EngineHarnessSubject::new_legacy_compatible(
+        &spec.clients,
+        ProtocolProfile::Legacy,
+        storage_mode,
+    )
+    .map_err(subject_setup_error)?;
     run_scenario_report_inner(spec, expected_trace, expected_outcomes, None, &mut subject).await
 }
 
@@ -366,9 +372,12 @@ pub async fn run_vector_fixture_report_with_storage_mode(
             });
         }
     };
-    let mut subject =
-        EngineHarnessSubject::new(&fixture.scenario.clients, protocol_profile, storage_mode)
-            .map_err(subject_setup_error)?;
+    let mut subject = EngineHarnessSubject::new_legacy_compatible(
+        &fixture.scenario.clients,
+        protocol_profile,
+        storage_mode,
+    )
+    .map_err(subject_setup_error)?;
     run_scenario_report_inner(
         &fixture.scenario,
         fixture.expected_trace.clone(),
--- a/crates/cgka-conformance-simulator/src/scenario.rs
+++ b/crates/cgka-conformance-simulator/src/scenario.rs
@@ -1,6 +1,6 @@
 //! Serializable scripted scenarios for the harness.
 //!
-//! `ScenarioSpec` is the v1 input-side companion to `ScenarioTrace`: external
+//! `ScenarioSpec` is the v2 input-side companion to `ScenarioTrace`: external
 //! implementations can drive the same logical client operations, then compare
 //! their observed trace to exact or semantic fixture expectations.
 
@@ -8,9 +8,9 @@ use crate::{
     ConvergenceFaultSubject, ConvergenceSubject, EngineHarnessSubject, ExpectationFailure,
     HarnessStorageMode, PendingResolutionObservation, ScenarioErrorObservation,
     ScenarioOracleReport, ScenarioTrace, SubjectCreateGroup, SubjectDescriptor, SubjectError,
-    SubjectInviteMembers, SubjectSendApplication, SubjectUpdateAdminPolicy, SubjectUpdateGroupData,
-    TraceExpectation, VectorFixture, build_scenario_oracle_report, compare_trace_expectations,
-    required_capability,
+    SubjectInviteMembers, SubjectOutboundArtifact, SubjectOutboundKind, SubjectOutboundOutcome,
+    SubjectSendApplication, SubjectUpdateAdminPolicy, SubjectUpdateGroupData, TraceExpectation,
+    VectorFixture, build_scenario_oracle_report, compare_trace_expectations, required_capability,
 };
 use cgka_traits::group::ProtocolProfile;
 use serde::{Deserialize, Serialize};
@@ -23,6 +23,29 @@ pub struct ScenarioSpec {
     pub spec_version: String,
     pub clients: Vec<String>,
     pub steps: Vec<ScenarioStep>,
+}
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ScenarioOutboundSelection {
+    #[default]
+    All,
+    StateConfirmation,
+    Welcome,
+    GroupMessage,
+    RegeneratedQueuedIntent,
+}
+
+impl ScenarioOutboundSelection {
+    fn matches(self, artifact: &SubjectOutboundArtifact) -> bool {
+        match self {
+            Self::All => true,
+            Self::StateConfirmation => artifact.state_confirmation_required,
+            Self::Welcome => artifact.kind == SubjectOutboundKind::Welcome,
+            Self::GroupMessage => artifact.kind == SubjectOutboundKind::GroupMessage,
+            Self::RegeneratedQueuedIntent => artifact.regenerated_queued_intent,
+        }
+    }
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
@@ -58,13 +81,13 @@ pub enum ScenarioStep {
         admins: Vec<String>,
         error: String,
     },
-    ConfirmPending {
+    AcknowledgeOutbound {
         client: String,
-        pending: String,
-    },
-    FailPending {
-        client: String,
-        pending: String,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        publication: Option<String>,
+        #[serde(default)]
+        selection: ScenarioOutboundSelection,
+        outcome: SubjectOutboundOutcome,
     },
     SendAppMessage {
         sender: String,
@@ -127,6 +150,24 @@ pub enum ScenarioStep {
 }
 
 impl ScenarioStep {
+    pub fn accept_publication(client: impl Into<String>, publication: impl Into<String>) -> Self {
+        Self::AcknowledgeOutbound {
+            client: client.into(),
+            publication: Some(publication.into()),
+            selection: ScenarioOutboundSelection::All,
+            outcome: SubjectOutboundOutcome::Accepted,
+        }
+    }
+
+    pub fn fail_publication(client: impl Into<String>, publication: impl Into<String>) -> Self {
+        Self::AcknowledgeOutbound {
+            client: client.into(),
+            publication: Some(publication.into()),
+            selection: ScenarioOutboundSelection::All,
+            outcome: SubjectOutboundOutcome::ReachedNoEndpoint,
+        }
+    }
+
     pub fn kind(&self) -> &'static str {
         match self {
             ScenarioStep::CreateGroup { .. } => "create_group",
@@ -134,8 +175,7 @@ impl ScenarioStep {
             ScenarioStep::UpdateGroupData { .. } => "update_group_data",
             ScenarioStep::UpdateAdminPolicy { .. } => "update_admin_policy",
             ScenarioStep::ExpectUpdateAdminPolicyError { .. } => "expect_update_admin_policy_error",
-            ScenarioStep::ConfirmPending { .. } => "confirm_pending",
-            ScenarioStep::FailPending { .. } => "fail_pending",
+            ScenarioStep::AcknowledgeOutbound { .. } => "acknowledge_outbound",
             ScenarioStep::SendAppMessage { .. } => "send_app_message",
             ScenarioStep::Leave { .. } => "leave",
             ScenarioStep::DeliverAll => "deliver_all",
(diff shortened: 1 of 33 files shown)


crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -455,6 +542,62 @@ impl EngineHarnessSubject {
...
+                });
+            self.outbound_cursors
+                .insert(label.to_owned(), emission.sequence);
+        }
+        Ok(())
+    }
+
+    fn mark_pending_outbound(&mut self, pending: PendingStateRef, outcome: SubjectOutboundOutcome) {
+        for record in self.outbound_records.values_mut() {
+            if record.pending == Some(pending) {
```
erskingardner: **Blocking: cross-client `PendingStateRef` collision.**

`outbound_records` is one map across every client, but `PendingStateRef` is a per-engine counter (`crates/cgka-engine/src/epoch_manager.rs`, `pending_counter += 1`), so distinct clients reuse the same values. This loop marks *every* record with a matching `u64`, regardless of owner.

Failure scenario: Alice invites Carol (alice `pending#2`), Bob independently commits (bob `pending#2`). Alice acknowledges her commit `ReachedNoEndpoint` and rolls back. `mark_pending_outbound` then stamps Bob's commit/Welcome records as `ReachedNoEndpoint` too — Bob's artifacts vanish from `poll_outbound` and can never be acknowledged, while his engine state was never rolled back and his artifacts are still sitting on the bus. Same hazard on the `confirm_pending` / `fail_pending` label paths at lines 675 and 692.

Fix: make the record key or the comparison `(client, pending)`, e.g. filter on `record.artifact.client == client` here and in the two scans in `acknowledge_outbound`, and pass the owning label into `mark_pending_outbound`.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -580,6 +730,158 @@ impl ConvergenceSubject for EngineHarnessSubject {
...
+                Ok(())
+            } else {
+                Err(SubjectError::new(
+                    "outbound_already_resolved",
+                    format!("outbound artifact {outbound_id} was already resolved as {previous:?}"),
+                ))
+            };
+        }
+
+        let pending_already_confirmed = record.pending.is_some_and(|pending| {
```
erskingardner: **Blocking (same root cause, worse consequence): this scan is not client-scoped.**

`pending_already_confirmed` matches any record in the global map with an equal `PendingStateRef`. If Bob has already accepted his `pending#2` commit, Alice acknowledging *her* `pending#2` commit as `Accepted` takes `pending_already_confirmed == true`, skips `try_confirm` entirely (line 795), and still sets `resolution = Accepted`. Alice's staged MLS state is never confirmed and the artifact is now unacknowledgeable — a silently stranded pending publish, which is precisely the failure this subject is meant to surface.

`should_roll_back_pending` (lines 821–837) has the identical leak in both directions: another client's unresolved sibling suppresses a legitimate rollback, and another client's accepted confirmation suppresses it permanently.

Add `candidate.artifact.client == client` to both `any(..)` predicates (or key the map by `(client, pending)`), and add a regression test with two clients each holding a live state-bearing commit.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -383,6 +467,7 @@ impl EngineHarnessSubject {
             SubjectCapability::AdminPolicyObservation,
             SubjectCapability::CrashReopen,
             SubjectCapability::VirtualTime,
+            SubjectCapability::OutboundPublication,
```
erskingardner, line 444: **Blocking: legacy-compatible subjects advertise a capability they do not implement.**

This list is unconditional, so `new_legacy_compatible` also reports `outbound_publication` in its descriptor. But those clients never call `bus.capture_outbound_for`, so `poll_outbound` always returns `[]` and `acknowledge_outbound` always fails with `unknown_outbound`. A scenario that preflights `required_capability(OutboundPublication)` passes and then silently no-ops.

Gate this entry on `external_publication_lifecycle`.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -825,6 +1127,17 @@ fn subject_engine_error(error: EngineError) -> SubjectError {
     SubjectError::new(observe_engine_error(&error), error.to_string())
 }
 
+fn subject_publication_error(error: EngineError) -> SubjectError {
+    match error {
+        EngineError::Other(message)
+            if message.starts_with("cannot report definite publication failure after") =>
```
erskingardner: Non-blocking: matching on a message prefix produced by `HarnessClient::try_fail` in another module is brittle — a reworded string there silently downgrades this to a generic engine error code, and the `outbound_already_exposed` contract quietly disappears. A typed `EngineError` variant (or a dedicated error enum on `try_fail`) would make the coupling checkable at compile time.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -346,13 +399,41 @@ pub struct EngineHarnessSubject {
     clients: BTreeMap<String, HarnessClient>,
     pending_refs: HashMap<String, PendingStateRef>,
     convergence_clock: ManualConvergenceClock,
+    outbound_cursors: HashMap<String, u64>,
+    outbound_records: BTreeMap<String, EngineSubjectOutboundRecord>,
```
erskingardner: Non-blocking: a `BTreeMap` keyed by `"outbound/{n}"` orders lexicographically, so `outbound/10 < outbound/2` — which is why `poll_outbound` needs the explicit `sort_by_key(record.sequence)` and why `sequence` is duplicated inside the record. Keying by the `u64` sequence (with the string handle derived on the way out) would give correct ordering for free and drop both the sort and the redundant field.

erskingardner (commented):
Nice shape overall — non-destructive polling, idempotent acknowledgement, and "exposure forbids definite failure" are the right primitives, and keeping the emission stream append-only and separate from the mutable queue is the correct way to avoid leaking queue indices. The tests are genuinely behavioral rather than mock-shaped.

But there is a blocking correctness bug: `EngineHarnessSubject` keys outbound bookkeeping on `PendingStateRef` in a **single map shared by every client**, and `PendingStateRef` is a per-engine `u64` counter (`epoch_manager.rs: pending_counter += 1`). Alice's `pending#1` and Bob's `pending#1` compare equal. All three cross-record scans (`mark_pending_outbound`, `pending_already_confirmed`, `should_roll_back_pending`) filter only on `record.pending == Some(pending)` and never on `record.artifact.client`, so one participant's acknowledgement silently mutates another participant's publication state. Details inline. In the worst case an accepted commit is marked resolved without ever confirming the engine's staged state — exactly the class of bug this subject exists to detect. The added tests miss it because only one client ever produces a state-bearing commit in each of them.

Secondary blocker: `new_legacy_compatible` still advertises `SubjectCapability::OutboundPublication`, so a legacy-mode subject passes capability preflight and then silently returns nothing from `poll_outbound`.

On premise (per the repo's Karpathy guidelines): this is ~1,250 lines adding a second publication lifecycle to the harness, with zero consumers today, and the entire existing v1 fixture suite explicitly routed away from it via `new_legacy_compatible`. That means the new path is exercised only by the nine tests in this PR, and the crate now carries two divergent publication semantics indefinitely. I'm not asking you to drop it — it's a tracked milestone deliverable — but marking Milestone 1.2 `complete` while no adapter consumes the contract is optimistic, and the dual-lifecycle branching in `client.rs` (five `external_publication_lifecycle` checks, two independent implementations of the "no-op creation" rule) is the maintenance cost to watch. Please add a concrete follow-up issue for collapsing the legacy path once Scenario IR v2 lands, so this doesn't become permanent.

No production engine/protocol code is touched — that claim in the description checks out.

crates/cgka-conformance-simulator/vectors/invite-member.v1.json
```diff
@@ -22,11 +22,7 @@
         "required_features": [],
         "pending": "create"
       },
-      {
-        "type": "confirm_pending",
-        "client": "alice",
-        "pending": "create"
-      },
+      { "type": "acknowledge_outbound", "client": "alice", "publication": "create", "outcome": "accepted" },
```
erskingardner: Non-blocking, but it recurs across every migrated vector: the new `acknowledge_outbound` steps are hand-written as single-line objects while every neighbouring step is expanded. `ScenarioReport::write_*` in `src/report.rs` emits fixtures with `serde_json::to_string_pretty`, so the next regeneration of any of these files will reflow these lines and produce diff noise unrelated to the change being made. Expanding them now keeps regeneration a no-op.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -477,7 +627,19 @@ impl ConvergenceSubject for EngineHarnessSubject {
             )
             .await;
         if let Some(pending_ref) = pending_ref {
-            self.insert_pending(action.pending, pending_ref)?;
+            self.insert_pending(action.pending, action.creator, pending_ref)?;
+            self.sync_client_outbound(action.creator)?;
+            let has_outbound_artifacts = self.outbound_records.values().any(|record| {
+                record.artifact.client == action.creator && record.pending == Some(pending_ref)
+            });
+            if !has_outbound_artifacts {
```
erskingardner: Non-blocking: the "a creation with nothing to publish confirms immediately" rule now exists twice with two different predicates — here it is "no outbound record exists for this pending", and in `HarnessClient::publish_send_result` (`SendResult::GroupCreated`) it is `!welcomes.is_empty()`. They agree today because a creation's only publication artifacts are its Welcomes, but they are not obviously the same condition, and a future creation that emits a non-Welcome artifact would silently diverge. Worth collapsing to one helper, or at least cross-referencing them in a comment so the coupling is discoverable.

crates/cgka-conformance-simulator/src/scenario.rs
```diff
@@ -516,31 +556,49 @@ async fn run_scenario_report_inner(
...
+                    .into_iter()
+                    .filter(|artifact| {
+                        publication.as_ref().is_none_or(|publication| {
+                            artifact.publication.as_ref() == Some(publication)
+                        }) && selection.matches(artifact)
+                    })
+                    .collect::<Vec<_>>();
+                if artifacts.is_empty() && publication.is_some() {
+                    return Err(err(
+                        step_index,
```
erskingardner, line 578: Non-blocking: an unlabelled `acknowledge_outbound` that matches nothing is a silent no-op, and `add_strict_reliability_oracle` now appends one per client to every generated family. That is the documented "idempotent drain" semantics and it is the right default, but it does mean a generated family that stops producing outbound work entirely still passes without any signal that the lifecycle is no longer being exercised. If that matters later, an assertion on the family generator (rather than a runtime error here) would be the cheap way to keep it honest.

erskingardner (commented):
**Recommend merge** — all three blocking findings from my previous review are fixed.

Re-review of the three blocking findings from https://github.com/marmot-protocol/mdk/pull/1205#pullrequestreview-4822667577. All three are genuinely fixed in the code, not just in the commit messages.

**1. Cross-client `PendingStateRef` collision — fixed.** Every scan that touches `outbound_records` now filters on `record.artifact.client == client`: `mark_pending_outbound`, `pending_confirmation_accepted`, `pending_non_confirmation_accepted`, the `another_confirmation_is_unresolved` predicate in the rollback decision, and both `pending_refs.retain` calls. `publication_for_pending` is client-scoped too, and `acknowledge_outbound` rejects a handle that does not belong to the acknowledging client (`outbound_client_mismatch`). `pending_publication_identity_is_scoped_by_client` is the regression I asked for and it is the right shape — it asserts `pending_refs["alice-first"].pending == pending_refs["bob-first"].pending` up front, so the test fails loudly if the collision precondition ever stops holding rather than silently passing. It covers both directions (acceptance not suppressed by a foreign accepted confirmation, rollback not stamping a foreign client's records).

**2. Unconditional capability set — resolved by removal.** There is no `new_legacy_compatible` any more, `capture_outbound_for` is unconditional in `ClientBuilder::attach`, and there is exactly one lifecycle, so the unconditional `OutboundPublication` entry is now accurate. `ProtocolProfile::Legacy` is orthogonal (engine application profile) and the README says so explicitly.

**3. Premise — this is the right resolution.** The clean v2 cutover is a much better answer than the follow-up issue I asked for. There is one publication contract, `validate_scenario_for_subject` rejects v1 with a test pinning it, and the dual `external_publication_lifecycle` branching in `client.rs` is gone. Notably the fixtures shrank (every vector is net-negative), which is the outcome I was hoping for. Milestone 1.2 `complete` is now defensible because every built-in runner consumes the contract, and the plan is honest that app-runtime/retained-relay/process adapters still do not exist.

Both non-blocking notes are also addressed: `HarnessPublicationError::AlreadyExposed` replaces the message-prefix match, and `outbound_records` is keyed by the `u64` emission sequence, which dropped the explicit sort and the duplicated `sequence` field.

Spot-checked the interleavings that the new client scoping makes reachable — welcome-then-commit and commit-then-welcome in both outcome orders, rollback after a sibling was accepted, retract of an already-retracted artifact, and the eager zero-fanout confirm — and did not find a hole. Full CI is green including the 13-minute simulator/vector job. Three minor notes inline; none block.

PR #1206, opened 2026-07-30
Bound deferred peel residence across restarts

### LINKED ISSUE #1193: Bound privacy-preserving retirement of indefinitely unpeelable transport objects
## Problem

A recipient can retain a raw Nostr kind-445 group object indefinitely after joining a group even when the object was authored before that recipient was a member and the recipient is not entitled to historical delivery.

The minimized sequence is:

1. Alice creates a group with Bob and Carol at epoch 1.
2. Bob authors an epoch-1 application message.
3. Delivery of that outer group object is delayed.
4. Alice adds David; David joins from the epoch-2 Welcome.
5. The retained epoch-1 object is delivered to David.

Carol can peel and receive the application message. David correctly receives no historical payload, but cannot peel the outer wrapper with any context he has. The object remains `TransportDeferred` / `MessageState::PeelDeferred` and keeps both engine pending-work and the simulator scenario-input ledger pending.

The strict generated reproducer is `convergence-chaos/v1`, seed `123`, case `4`; `strict_chaos_boundary_retains_known_reliability_failures` currently pins this as a known `pending_work_remaining` failure.

The harness also exposes an epoch-1 invite-commit echo to David because group messages are broadcast to all attached clients. That creates a second deferred row, but the pre-join application row is independently identified in the scenario-input ledger and is sufficient to demonstrate the issue.

## Source trace

The engine already has the correct terminal classification **after peeling succeeds**:

- a Welcome durably records `Group::join_epoch`;
- an application message with a recovered MLS source epoch before that membership lower bound is classified `StaleReason::PreMembership`;
- its content row becomes `Failed`, the raw wrapper leaves the retry lifecycle, and no `MessageReceived` or losing-branch invalidation is emitted.

Production Nostr kind-445 wrapping hides both the MLS content class and source epoch. A wrong exporter secret therefore produces only `PeelerError::DecryptFailed`. The raw row is persisted as `PeelDeferred` using the receiver's current epoch, not the unknown message source epoch. At that seam the engine cannot distinguish:

- an impossible pre-membership object; from
- a valid future-epoch object authored while the recipient was a member and waiting for a missing commit.

Publishing source epochs on public relays would add metadata we do not want to expose. Outer `created_at`, the current epoch, elapsed retries, or failure against all currently retained snapshots are not safe semantic evidence of pre-membership.

## Distinction from #1171

This is adjacent to, but not an instance of, #1171.

- Here the recipient was not a member at authorship and never had the necessary keys. No retention or replay-order change can make the payload deliverable.
- In #1171 the recipient was a member at authorship. The message can be legitimately decryptable/in-window before commit-first catch-up advances far enough to evict its source material.

Any solution here must preserve ordinary offline retry for messages authored while the recipient was a member. It must not generalize `source_epoch < latest join_epoch` across rejoin histories unless durable membership history proves the recipient was absent at that source epoch.

## Current resource behavior

The deferred-peel lifecycle is bounded but does not necessarily retire this row:

- retries are gated by the `(current epoch, retained snapshot set)` fingerprint;
- unchanged context suppresses repeated wasted work;
- at most 64 rows are attempted per sweep;
- the default budget is 32 actual attempts under changed contexts;
- the per-group retained-row cap is 256;
- retry-budget exhaustion deletes the raw row as a local `resource_refused` outcome and deliberately does not create terminal dedup, so later same-ID redelivery remains eligible.

The fingerprint gate and per-row attempt counters are in memory. A stable group can therefore retain the pre-join row indefinitely after one failed cycle because its context never changes. Restart preserves the durable row and cap usage but resets the gate and retry count, allowing another wasted sweep and restarting the budget.

## Proposed privacy-preserving improvement

Treat eventual release as a **durable local resource decision**, not as proof that the object is stale, invalid, or permanently unpeelable.

Persist enough deferred-lifecycle metadata to bound residence across restarts without publishing additional transport metadata, for example:

- first locally observed time;
- last attempted peel-context fingerprint;
- number of distinct-context peel attempts;
- last resource/backfill action;
- optionally a conservative local expiry deadline derived from a configured maximum residence period.

Then:

1. Retry when the peel context changes, as today.
2. Continue enforcing bounded sweeps and the per-group row cap.
3. Release a row when either its durable distinct-context budget or local residence budget is exhausted.
4. Classify that release as `resource_refused`, not `stale`, `invalidated`, `Failed`, or a terminal duplicate.
5. Delete the raw row and release its deferred-cap slot and pending-work entry.
6. Keep exact-ID redelivery eligible. If the missing state later arrives and the transport redelivers the object, it must be peelable and process normally.
7. Trigger or preserve the binding's refetch/backfill recovery signal before claiming transport history synchronized.

Wall-clock handling must be conservative: a backward clock jump or missing durable timestamp must not cause premature expiry. A persisted high-water mark or fail-open age calculation is preferable to trusting one wall-clock subtraction. The decision and telemetry must remain privacy-safe: aggregate counts and stable reason tags only, without group ids, message ids, relay URLs, payloads, or epoch values in tracing.

This improves storage and quiescence without pretending to solve semantic classification at the opaque transport seam.

## Acceptance criteria

- Add a focused first-join reproducer for the pre-membership application object using production-shaped opaque peeling.
- Before the local resource deadline, the object remains retryable and no terminal protocol disposition or app event is emitted.
- After controlled time/context advancement past the durable resource budget, the raw row is removed, its cap slot is released, and pending-work no longer reports it.
- The audit outcome is explicitly `resource_refused` and does not create terminal dedup.
- Restart does not reset the durable residence/attempt budget or prematurely expire the row.
- Backward wall-clock movement does not cause early release.
- A future-epoch application authored while the recipient was a member still succeeds if its missing commit arrives before release.
- If such a legitimate row is resource-released first, later same-ID redelivery after the missing commit is accepted and processed.
- Rejoin coverage proves the first-join conclusion is not generalized to messages authored during an earlier membership interval.
- The strict chaos case no longer retains pre-join application pending work once the controlled resource deadline is advanced; delivery remains absent for David and present for entitled members.
- No source epoch or other new group-history metadata is exposed on public Nostr events.

## Relevant code

- `crates/cgka-conformance-simulator/src/family.rs::convergence_chaos_delayed_past_epoch_app`
- `crates/cgka-conformance-simulator/src/bus.rs`
- `crates/cgka-engine/src/group_lifecycle.rs::do_join_welcome`
- `crates/cgka-engine/src/message_processor/ingest.rs::ingest_group_message`
- `crates/cgka-engine/src/message_processor/ingest.rs::msg_is_pre_membership`
- `crates/cgka-engine/src/message_processor/mod.rs::retry_deferred_peels`
- `crates/cgka-engine/src/message_disposition.rs`
- `crates/cgka-engine/tests/deferred_peel_lifecycle.rs`
- `crates/transport-nostr-peeler/src/peeler.rs`

Related: #1171, but this issue must not change member-authored offline delivery semantics.

erskingardner:
Closes #1193.

## What changed

- persist a restart-safe deferred-peel lifecycle with wall-clock high-water and monotonic residence deadlines
- retire opaque transport objects after a bounded 30-day local residence without recording a terminal protocol verdict
- preserve exact-ID redelivery eligibility after resource release
- schedule deferred-peel cutoffs without masking pending convergence or outbound work
- emit typed resource-refusal signals and arm epoch backfill only from cursor-advancing clients
- bound legacy/restart lifecycle persistence to one sweep-sized slice and report durable residence time
- avoid applying the latest Welcome boundary to messages from prior membership intervals after rejoin
- carry the new recovery event through app routing, audit output, UniFFI, and simulator input accounting
- update the strict chaos seed-123 case to require bounded retirement and recovery
- keep the workspace release version at 0.9.10

## Root cause

PeelDeferred rows were retried only when peel context changed. A pre-membership kind-445 object with a stable context could therefore remain in durable storage forever, including across restarts, while the runtime had no deadline to wake and no recovery signal to request missing history.

## Impact

Opaque inputs now have bounded local residence while retaining recoverability: expiry is classified as a local resource refusal, not stale or invalid protocol input, and redelivery remains eligible if backfill later supplies the missing state.

## Validation

- just fast-ci
- cargo test -p cgka-engine --features test-policy-overrides --test deferred_peel_lifecycle
- cargo test -p marmot-app --test epoch_stall_backfill_audit
- cargo test -p marmot-forensics
- cargo nextest run -p cgka-conformance-simulator --locked --profile ci (166/166)
- simulator report generation (19/19 vectors)
- CI Rust partition 2 command (784/784)
- CI Rust partition 1 command (809 passed; one unrelated wn-opencode timing test passed in isolation)
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added durable tracking and scheduling for deferred message processing across restarts.
  * Added residence-time limits that retire deferred messages when local retention budgets expire.
  * Added resource-refusal events across application and FFI interfaces.
  * Added recovery handling to request missing history after resource refusal.

* **Bug Fixes**
  * Improved rejoin handling and late-message processing.
  * Ensured deferred work is cleared after final processing.

* **Documentation**
  * Updated architecture, scenario, and reliability documentation for lifecycle and timing behavior.


crates/marmot-app/src/client/sync.rs
```diff
@@ -62,7 +63,14 @@ impl AppClient {
         &mut self,
         group_id: &cgka_traits::GroupId,
     ) -> Result<ConvergenceScheduleState, AppError> {
-        match self.runtime.prepare_convergence_cutoff_delay_ms(group_id)? {
+        let convergence_delay = self.runtime.prepare_convergence_cutoff_delay_ms(group_id)?;
+        let deferred_delay = self.runtime.deferred_peel_cutoff_delay_ms(group_id)?;
+        let earliest_delay = match (convergence_delay, deferred_delay) {
```
erskingardner: Blocking: taking the `min` before the `match` makes the `None` arm unreachable whenever the group has any `PeelDeferred` row, because `deferred_peel_cutoff_delay_ms` returns `Some(..)` for every such group.

`PendingUnopenable` and `PendingOutbound` are only produced inside that `None` arm. So a group that has unopenable pending convergence inputs (or queued outbound intents) *and* one deferred row is now classified `Collecting { remaining_ms }` with up to 30 days remaining, instead of going through `schedule_unsettled_groups` / the unsettled re-arm cap. The deferred deadline silently outranks a signal it knows nothing about.

Suggest classifying first and only then clamping, e.g. compute the existing state from `convergence_delay` as today, and afterwards fold `deferred_delay` in — `Idle`/`Collecting` become `Collecting { min }`, while `PendingUnopenable`/`PendingOutbound` keep their own scheduling and just get an `arm_no_later` for the deferred deadline.

crates/marmot-app/src/client/sync.rs
```diff
@@ -62,7 +63,14 @@ impl AppClient {
         &mut self,
         group_id: &cgka_traits::GroupId,
     ) -> Result<ConvergenceScheduleState, AppError> {
-        match self.runtime.prepare_convergence_cutoff_delay_ms(group_id)? {
+        let convergence_delay = self.runtime.prepare_convergence_cutoff_delay_ms(group_id)?;
+        let deferred_delay = self.runtime.deferred_peel_cutoff_delay_ms(group_id)?;
```
erskingardner: Blocking (related): `deferred_peel_cutoff_delay_ms` can report `Some(0)` for a group whose epoch state is not `Stable`, but `retry_deferred_peels` early-returns `Ok(0)` in exactly that case (`EpochState` gate at the top of the function).

The result is a self-sustaining wakeup: `Some(0)` -> `Ready` -> `schedule_after_pass` re-arms at `MIN_CONVERGENCE_SETTLEMENT_DELAY` (10 ms) -> `sync_runtime_groups` + `advance_convergence_after_runtime_sync` -> the row is still there and still expired -> `Some(0)`. Nothing in the loop can retire the row until the group returns to `Stable`, and for the quarantine case you already guard against this explicitly at the top of `deferred_peel_cutoff_delay_ms`.

The delay query should mirror the same preconditions the sweep enforces (non-`Stable` -> `None`), otherwise the two disagree and the scheduler spins.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -920,6 +916,45 @@ impl<S: StorageProvider> Engine<S> {
...
 
+        let now = self.convergence_now();
+        let mut deferred: Vec<_> = self
+            .storage
+            .list_messages(group_id, EpochId(0))?
+            .into_iter()
+            .filter(|record| record.state == MessageState::PeelDeferred)
+            .collect();
+        for record in &mut deferred {
+            self.normalize_deferred_peel_lifecycle(record, now)?;
```
erskingardner: This normalization loop runs over *every* `PeelDeferred` row (up to `MAX_PEEL_DEFERRED_ROWS_PER_GROUP` = 256), is not bounded by `MAX_DEFERRED_ROWS_PER_SWEEP`, and sits **above** the context-fingerprint gate a few lines down — the gate whose entire job is "the peel context is unchanged, every retained row would fail exactly as before, do no work".

Worse, `normalize_deferred_peel_lifecycle` sets `changed = true` on the `now.wall_ms > wall_high_water_ms` branch, which is true on essentially every call at millisecond resolution. So each sweep, and each `deferred_peel_cutoff_delay_ms` call from the app scheduler, issues up to 256 SQLCipher `put_message` writes purely to advance a high-water mark that is only ever consulted when rebasing after restart.

Two cheap fixes, either works: (a) only persist the high-water advance when it actually matters — on the sweep path, or when it moves by a material fraction of the residence window rather than by a millisecond; (b) compute the residence check from the in-memory record and let the durable write ride along with the attempt-consumption write you already do below. Both keep the restart guarantee (rebasing fails conservatively toward *longer* residence) without turning an idle gated group into a per-tick write storm.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -1118,6 +1139,111 @@ impl<S: StorageProvider> Engine<S> {
...
+            u64::from(lifecycle.distinct_context_attempts)
+        });
+        self.storage.delete_message(&record.id)?;
+        self.audit_group(
+            &record.group_id,
+            crate::audit_helpers::deferred_peel_resource_refused_event(
+                hex::encode(record.id.as_slice()),
+                Some(record.epoch),
+                disposition.tag(),
+                retry_count,
```
erskingardner: Two small losses here that the durable lifecycle could avoid:

- `sweeps_waited` is now always `0`. It was `sweep_index - first_seen_sweep`. The field stays in the audit schema but no longer carries information, which is worse than removing it — a dashboard reading it will show a flat zero.
- `first_observed_wall_ms` is persisted on every deferred row but never read anywhere in the workspace (grep confirms: written in `normalize_deferred_peel_lifecycle` and `store.rs`, read nowhere).

Deriving the residence age from `first_observed_wall_ms` (against `wall_high_water_ms`) at release time fixes both at once: the audit field regains meaning and the durable field stops being dead weight in every serialized record.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -1186,11 +1312,10 @@ impl<S: StorageProvider> Engine<S> {
     pub(crate) fn note_peel_deferred_row_retired(
```
erskingardner: The `_msg_id` parameter is now unused, but all ten call sites still compute and pass `&record.id`. Drop the parameter and the arguments — the durable lifecycle made the in-memory per-row map unnecessary, so the signature should follow.

crates/marmot-app/src/client/sync.rs
```diff
@@ -89,6 +97,26 @@ impl AppClient {
     ) {
         self.pending_convergence_groups
             .extend(effects.pending_convergence.iter().cloned());
+        for event in &effects.events {
```
erskingardner, line 116: Two notes on this block:

1. The sibling arming path (`note_epoch_stall_from_ingest`, same file) gates on `self.app.cursor_persistence() != CursorPersistence::Advance` before arming a full-history replay. This new path does not, so a non-advancing cursor mode can now trigger `activate_transport(None)`. If that asymmetry is deliberate it deserves a comment; otherwise it looks like an omission.
2. `remember_pending_convergence_effects` is named for bookkeeping and previously did one `extend`. Recovery routing that reads group records and arms external replay is a different responsibility — consider a separate `arm_recovery_from_effects` called alongside it, so the worker seam reads honestly.

Cargo.toml
```diff
@@ -26,7 +26,7 @@ members = [
 
 [workspace.package]
 edition = "2024"
-version = "0.9.10"
+version = "0.9.11"
```
erskingardner: This repo bumps the workspace version in dedicated `chore(release): prepare X` commits (`b2919f15`, `a4374b78`), not in feature PRs. Unless the release cohort is going out with this change, the bump is drive-by scope that will conflict with the next feature branch.

erskingardner:
## Review: request changes

The premise is sound: #1193 asks for a durable residence bound on opaque `PeelDeferred` rows, and moving the attempt/first-seen bookkeeping from `Engine::deferred_peel` onto `MessageRecord` is the right shape. The restart/backwards-clock test and the strict chaos repair are convincing. Two things block merge for me, both in how the new deadline is wired into the app scheduler rather than in the engine core.

**Blocking**

1. `convergence_schedule_state` now short-circuits the `None` arm whenever any deferred row exists, so `PendingUnopenable` / `PendingOutbound` become unreachable for those groups. A group with unopenable convergence inputs plus one deferred row is reported `Collecting { remaining_ms }` with up to 30 days remaining instead of entering the unsettled re-arm path.
2. `deferred_peel_cutoff_delay_ms` reports `Some(0)` from a state `retry_deferred_peels` refuses to act on (non-`Stable` epoch state), which re-arms the worker at `MIN_CONVERGENCE_SETTLEMENT_DELAY` forever.

**Non-blocking, but worth addressing**

3. `normalize_deferred_peel_lifecycle` persists every deferred row on essentially every call (wall high-water advances each millisecond), unbounded by `MAX_DEFERRED_ROWS_PER_SWEEP`, and it runs *before* the context-fingerprint gate — the gate exists precisely to suppress this work. It is also on the scheduler query path, so each tick can write up to `MAX_PEEL_DEFERRED_ROWS_PER_GROUP` encrypted rows to advance a mark only read at restart.
4. `sweeps_waited` is now hardcoded `0`, and `first_observed_wall_ms` is written but never read — one telemetry field silently degraded, one durable field dead on arrival. Deriving the age from `first_observed_wall_ms` would fix both.
5. `note_peel_deferred_row_retired` keeps an unused `_msg_id` parameter and ten call sites that pass it.
6. The new backfill arming in `remember_pending_convergence_effects` skips the `cursor_persistence() == Advance` gate its sibling path applies, and it is recovery routing living in a function named "remember".
7. The workspace version bump to `0.9.11` is done in dedicated `chore(release): prepare X` commits in this repo's history, not in feature PRs.

The `join_epoch` → `EpochId(0)` on rejoin is a real semantic widening beyond the residence work (pre-membership terminal classification is disabled permanently for any rejoined group), but #1193 asks for exactly that and the test names the constraint, so I read it as in scope.

_(GitHub does not allow a formal request-changes review on one's own PR; posting as a comment. Line-specific detail is in the inline comments.)_

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -65,7 +70,7 @@ pub(crate) struct DeferredPeelGroupState {
     /// skipped — a deferred row can only become peelable when the group
     /// epoch advances or the retained-snapshot set changes.
     gate: Option<[u8; 32]>,
-    /// Monotonic sweep counter; also the clock behind `sweeps_waited`.
+    /// Monotonic sweep counter used to report aggregate sweep progress.
     sweep_count: u64,
```
erskingardner, line 74: Non-blocking (dead state, net deletion available): `sweep_count` is now write-only. Its only consumer was `sweeps_waited`, which this PR replaced with the durable `residence_ms`. It is still incremented at line 989 and then read into `_sweep_index` — a deliberately ignored binding. Nothing else in the crate reads it (grepped `mod.rs`/`store.rs`/`ingest.rs`).

Dropping the field and the tuple element removes ~4 lines of state that a future reader will otherwise assume is load-bearing.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -1118,6 +1143,44 @@ impl<S: StorageProvider> Engine<S> {
...
+    /// residence deadline. Legacy rows receive a fresh conservative deadline;
+    /// process-local monotonic deadlines are rebased after restart.
+    pub fn deferred_peel_cutoff_delay_ms(
+        &mut self,
+        group_id: &GroupId,
+    ) -> Result<Option<u64>, EngineError> {
+        if self.quarantined_reason(group_id).is_some() {
+            return Ok(None);
+        }
+        if let Some(state) = self.epoch_manager.state(group_id)
```
erskingardner, line 1156: Non-blocking: this new `EpochState` gate is the fix for the scheduler hot-spin (delay fn said `Some(0)`, sweep said "not Stable, do nothing", scheduler re-armed at 10ms forever). It's the right fix, but it's a two-function invariant — `retry_deferred_peels` and `deferred_peel_cutoff_delay_ms` must keep the same preconditions — enforced only by these two hand-mirrored blocks.

A regression test that puts a group in a non-`Stable` epoch state with an expired deferred row and asserts `deferred_peel_cutoff_delay_ms == None` would pin the invariant cheaply. Same for the bounded-normalization slice (>64 legacy rows -> `Some(0)` until drained); `deferred_peel_residence_survives_restart_and_backward_clock` covers the rebasing math but not the slicing.

erskingardner (commented):
Re-review at `7cb323b7`. All seven findings from the previous pass are addressed, and I verified each fix against the head source rather than taking the commit message for it.

**Previous findings — resolved**

1. **`min` before `match` in `convergence_schedule_state`** — fixed. Classification now happens first; the deferred deadline only clamps an existing `Collecting` window, and `PendingUnopenable`/`PendingOutbound` keep their own scheduling. I confirmed the concern I raised about starving those two states is unfounded: `schedule_unsettled_groups` re-arms indefinitely (worst case the 60s retry ceiling), so a deferred row in an unopenable group still reaches its 30-day deadline. The `Some(0)` after `min` correctly collapses to `Ready`.
2. **Scheduler hot-spin from `Some(0)` on a non-`Stable` group** — fixed by the matching `EpochState` gate in `deferred_peel_cutoff_delay_ms`. The two functions now agree on quarantine, non-`Stable`, and the `epoch_manager.state() == None` fall-through.
3. **Per-millisecond `put_message` storm** — fixed properly rather than papered over. The `wall_high_water_ms` advance branch is gone entirely; writes now only happen on legacy init or restart rebasing, bounded to `MAX_DEFERRED_ROWS_PER_SWEEP` (64), with `normalization_pending` driving a `Some(0)` re-arm until the backlog drains. Worst case is 4 slices for a full 256-row group, then zero writes forever. Good outcome.
4. **`sweeps_waited` always 0 / `first_observed_wall_ms` never read** — fixed at both ends: the field is renamed to `residence_ms` through audit helpers, the v2 schema, and the tests, and it is derived from `first_observed_wall_ms`. Clean rename, no leftovers anywhere in the diff.
5. **Unused `_msg_id`** — dropped, along with all ten call-site arguments.
6. **Recovery arming ignored `CursorPersistence`** — fixed, and split into `arm_recovery_from_effects` so `remember_pending_convergence_groups` goes back to doing one thing.
7. **Drive-by version bump** — reverted to `0.9.10`.

CodeRabbit's `store.rs` colocation point is also handled; `7cb323b7` is a pure move (91 out, 94 in, byte-identical bodies modulo import paths).

**Remaining — both non-blocking, in inline comments**

- `sweep_count` is now write-only dead state; removing it is a net deletion.
- The two scheduler-invariant fixes (non-`Stable` gate, bounded normalization slice) have no regression test pinning them.

Full CI is green at this SHA, including the simulator, vectors, and Tamarin jobs. Recommend merge.

PR #919, opened 2026-07-18
feat(marmot-app): record an epoch_stall_backfill_armed forensic audit row at the epoch-stall arm decision

jgmontoya:
---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/919">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added a new forensic audit log event for when epoch-stall backfill is armed, including the stalled epoch and backfill threshold.
  * Extended the audit log schema and JSON serialization to support the new `epoch_stall_backfill_armed` event type.
* **Bug Fixes**
  * Improved forensic upload scheduling to reflect whether backfill was armed before replay, even if replay fails.
  * Avoided duplicate audit entries for repeated triggers at the same stalled epoch.
* **Tests**
  * Added end-to-end tests covering threshold behavior, schema/event round-tripping, and upload generation without visible activity.

PR #1208, opened 2026-07-30
Feat/epoch stall backfill escalation

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Improved recovery from stalled epoch gaps by detecting repeated unsuccessful backfill attempts.
  * Added one-time escalation reporting after repeated stalls, including the affected group, epoch, and attempt count.
  * Added durable audit records and runtime events for escalations.
  * Exposed escalation details through supported event integrations.

* **Documentation**
  * Updated audit logging and reliability documentation to describe escalation behavior and thresholds.

* **Tests**
  * Added coverage for escalation, delivery across failed syncs, recovery resets, and audit serialization.


crates/marmot-app/src/client/sync.rs
```diff
@@ -244,10 +247,11 @@ impl AppClient {
         // Preserve that scheduling edge even when hydration emitted no app
         // events; the worker drains this set immediately after startup sync.
         self.remember_pending_convergence_groups(&effects);
-        self.arm_recovery_from_effects(&effects);
         fail_if_publish_failed(&effects)?;
```
dannym-arx: arm_recovery_from_effects was quietly moved past fail_if_publish_failed with no test pinning the failure-then-no-arm-then-arm-next-pass behavior.

dannym-arx (dismissed):

crates/marmot-app/src/client/sync.rs
```diff
@@ -618,16 +627,86 @@ impl AppClient {
...
+    /// keeps delivery exactly-once.
+    ///
+    /// Residual: the receive arm drops the whole client on a failed pass, which
+    /// discards this stash and the detector's `escalated` latch together, so a
+    /// fresh run re-escalates from its own first three arms. That is a coherent
+    /// loss of the session, not a silently swallowed report.
+    fn drain_epoch_stall_escalations(&mut self, summary: &mut SyncSummary) {
+        summary
+            .epoch_stall_escalations
+            .append(&mut self.pending_epoch_stall_escalations);
```
dannym-arx, line 744: drain_epoch_stall_escalations "call me last at every outermost seam" contract is convention-only, a wrapper or debug_assert would harden it.

crates/marmot-app/src/client/sync.rs
jgmontoya: Confirmed, and it's slightly worse than described. The move was in f28d95e6 (master had arm_recovery_from_effects above the gate); the commit message mentioned it as an aside, matching advance_convergence_after_runtime_sync. But there's no "arm next pass": session.drain() empties an in-memory events_buf (engine.rs:2572), and TransportObjectResourceRefused is pushed after its durable row is deleted (store.rs:501 → :512), so once fail_if_publish_failed returns Err the refusal is gone permanently, no arm, no arms increment, no audit row. The old order survived the ? because epoch_backfill_pending is a field, not summary state. This also matters because drain() itself publishes (runtime.rs:1644-1650, incl. resume_outbound_fanouts), so failures and refusals are positively correlated, not independent. Worth noting the hoist matched the sibling that isn't the documented convention: ingest_delivery (sync.rs:560-562) deliberately captures the publish error and arms anyway, precisely so an incidental publish failure can't discard already-observed inbound effects. Fixed in 9ef50637, the arm now precedes the gate again, pinned by a_publish_failure_in_the_session_event_drain_still_arms_recovery asserting both the Err and the surviving arm + audit row. The identical ordering in advance_convergence_after_runtime_sync is pre-existing on master and we'll file it separately.

crates/marmot-app/src/client/sync.rs
jgmontoya: Correct that it's convention-only, though the stranding cost is smaller than it looks: all five drain sites share one stash and append moves, so a seam that forgets to drain defers the escalation to the next seam that does, and the epoch_stall_backfill_escalated audit row is already durable by then. Outright loss needs the client to die first, which is the residual documented at drain_epoch_stall_escalations. Both suggested remedies misfire, unfortunately: a Drop guard can't tell Ok from Err, so it would move escalations onto the summary a failing pass discards, the exact bug f28d95e6 fixed, and debug_assert!(pending.is_empty()) at pass entry would fire on the intended state, since a non-empty stash at entry is how a failed pass carries its escalation forward. What we did instead: BackfillDecision is now #[must_use] (c6f83f23), which guards the half where a forgotten call is real loss rather than delay, a dropped decision is never re-raised, because the detector latches escalated once per run, and clippy's -D warnings enforces it at compile time. Plus a seam test on ingest_received_delivery (1cd672d2), the runtime's dominant receive path, which previously had no escalation-delivery coverage (all existing tests delivered via sync()). One related constraint now stated in the doc comment (ebdd5ce5): drain_pending_session_events drains while nested inside sync_inner, so its escalations ride summary from the merge — nothing fallible can be inserted between that merge and the Ok return without silently orphaning them.

crates/marmot-app/src/client/sync.rs
```diff
@@ -618,16 +644,95 @@ impl AppClient {
...
+    /// between that merge and `sync_inner`'s `Ok` — past the merge they sit on a
+    /// local summary again, and a `?` would take them down with the pass. (Which
+    /// also makes `sync_inner`'s own call belt-and-braces rather than
+    /// load-bearing: the nested drain has already emptied the stash.)
+    ///
+    /// Residual: the receive arm drops the whole client on a failed pass, which
+    /// discards this stash and the detector's `escalated` latch together, so a
+    /// fresh run re-escalates from its own first three arms. That is a coherent
+    /// loss of the session, not a silently swallowed report.
+    fn drain_epoch_stall_escalations(&mut self, summary: &mut SyncSummary) {
```
mubarakcoded, line 741: A failed receive can drop the client before this drain runs, permanently losing the runtime escalation even though its audit row survives. Could we preserve pending escalations across the worker’s client rebuild?

jgmontoya: Premise confirmed: account_worker.rs:812 is the only path that rebuilds the client, and it takes the stash and the detector down together — the row at sync.rs:699 is written before the push at :700, so forensics survive, though only for accounts that opted into audit logging (it defaults off, so in the default posture nothing survives).

On "permanently" you are half right, and my own residual note was too generous: a fresh run still needs three arms, and arms 2 and 3 each require a real local epoch advance because epoch_stall.rs:258/:283 return Skip at an epoch already fired at. So a group that is still moving re-escalates two epochs later, while a group that freezes at one epoch arms once and never escalates again — though that second case is a pre-existing blind spot, since a group frozen from the start never escalates either.

I'd rather not preserve the stash. The event is point-in-time advice, and every sibling stash on AppClient is volatile precisely because its state is recoverable — pending_welcome_delivery_events even self-heals from the engine's retained Welcome index via reconcile_pending_welcome_delivery_index. An escalation has no such reconcile source (epoch_stall.rs:147-149: a device that cannot decrypt a group's traffic can never learn its live epoch), so persisting it would mean a hand-cleared sole-source-of-truth table replaying a recommendation the reopen's own transport re-activation may already have invalidated.

So: I'll tighten that residual paragraph to say what re-escalation actually requires, and open a follow-up PR for the real gap here, escalating a group wedged at a single epoch.

crates/marmot-app/src/client/epoch_stall.rs
```diff
@@ -47,19 +117,91 @@ struct GroupStall {
...
+    /// which keeps up fine but cannot read one peer's traffic — a forked peer, or
+    /// minted envelopes — can also complete a run. The report stays true as
+    /// stated (this device cannot read this group's traffic), and because
+    /// escalation only reports, the app still owns whether a re-sync is the right
+    /// answer.
+    fn observe_epoch(&mut self, epoch: EpochId) {
+        if self.epoch == epoch {
+            return;
+        }
+        if self.armed_at_epoch != Some(self.epoch) {
```
erskingardner, line 168: A run can only end if a delivery is observed at an epoch the device did *not* arm at, and then the device leaves it. That is strictly narrower than "passed cleanly through an epoch", in two ways worth documenting (or fixing):

1. **One clean advance is invisible.** `observe_group_epoch` fires only from `detect_epoch_stall`'s `_ =>` arm, i.e. it needs a decryptable delivery with a `group_id_hint` to land *while* the device sits at the intermediate epoch. Epoch advances produced by folding retained commits in `advance_convergence_after_runtime_sync` never reach the detector at all. The new tests encode this asymmetry without naming it: `an_epoch_passed_cleanly_between_arms_never_escalates` needs **two** `advance_bobs_epoch` calls to trip the reset, while `repeated_arming_without_recovery_escalates_exactly_once` uses one per arm and the run survives. A device that recovers by exactly one epoch and then stalls at an unrelated later epoch is counted as a continuation.

2. **Runs are unbounded in time.** After an arm, undecryptable traffic typically stops, so the last observation for that group stays at the armed epoch. Weeks later, an unrelated stall calls `observe_epoch(E_new)` with `armed_at_epoch == self.epoch` still true, so it lands as arm 2 of a long-dead run rather than arm 1 of a new one. Nothing ages the counter out.

Both push toward false-positive escalation, which is the direction the threshold rationale in the constant doc assumes away ("too low reports groups a further replay would have healed"). Since escalation is report-only the blast radius is a spurious "this group cannot catch up" notice, so this is not blocking — but it is a different limitation from the one already documented ("a device which keeps up fine but cannot read one peer's traffic"), and it is the one that will actually fire in the field.

A cheap partial fix: also call `epoch_stall.observe_group_epoch(group_id, epoch)` from `advance_convergence_after_runtime_sync`, which is precisely the path that folds the commits that advance the epoch. That closes (1) in about three lines and makes the documented semantics true as written.

crates/marmot-app/src/client/sync.rs
```diff
@@ -683,6 +797,7 @@ impl AppClient {
         let effects = self.runtime.advance_convergence(group_id).await?;
         fail_if_publish_failed(&effects)?;
         self.remember_pending_convergence_groups(&effects);
+        let mut summary = SyncSummary::default();
```
erskingardner, line 800: This hoist is unrelated churn. `summary` is not touched between here and its original declaration site just above `summary.projection_updates.extend(finalize_updates)` (line 820), and the new `drain_epoch_stall_escalations` call is at line 846 — so nothing in this PR requires moving it. Reverting keeps the diff surgical and the blame history on this function intact.

crates/marmot-app/tests/epoch_stall_backfill_audit.rs
```diff
@@ -224,7 +254,78 @@ async fn stalled_bob_in_a_live_group(
...
+    async fn advance_bobs_epoch(&mut self, name: &str) {
+        let before = self.bobs_epoch();
+        self.alice
+            .update_group_profile(&self.group_id, Some(name), None)
+            .await
+            .unwrap();
+        let deadline = Instant::now() + EPOCH_ADVANCE_DEADLINE;
+        loop {
+            self.bob.sync().await.unwrap();
+            sleep(CONVERGENCE_QUIESCENCE_WAIT).await;
```
erskingardner, line 310: Flake-risk note, non-blocking. `advance_bobs_epoch` retries a fixed 1.2 s quiescence sleep against a 20 s deadline, and `repeated_arming_without_recovery_escalates_exactly_once` chains three `stall_bob_for_one_run` calls (5 syncs + 50 ms each) with two epoch advances between them — comfortably over a minute of wall clock on a loaded runner, against a live `MockRelay`.

The failure mode is also unhelpful: if the relay is slow, the assertion that fires is `"bob did not apply alice's commit within the deadline"` inside the helper, not the escalation assertion the test exists for, so a timing flake reads as a convergence bug. Given #1249 just moved the long simulator campaigns to nightly, it may be worth either tightening the loop (poll on a short interval rather than sleeping a fixed 1.2 s per attempt) or noting these two as candidates for the slower tier.

erskingardner (commented):
Read the linked context (#892, #919, #1188), the two resolved reviewer threads, and the diff at `4a21927`. The core detector logic is sound and the exactly-once delivery design (stash + drain-at-outermost-`Ok`) is the right shape — moving rather than copying, and deferring rather than losing, is exactly what the one-shot latch requires. Prior blocking feedback is genuinely addressed, and the acceptance test that drives three real arms through a live relay is good evidence.

No blocking correctness defects found. Three non-blocking notes inline, plus one repo-wide observation below.

**Documentation volume.** 1302 added lines for roughly 60 lines of new logic. Three facts — "escalation only reports, the app owns key rotation", "the run counter is in-memory so a restart forgets it", and "latched once per unrecovered run" — are each restated near-verbatim in the `epoch_stall` module header, `EPOCH_STALL_ESCALATION_ARM_THRESHOLD`, `BackfillDecision`, `apply_backfill_decision`, `drain_epoch_stall_escalations`, `EpochStallEscalation`, `MarmotAppEvent::EpochStallEscalated`, `AuditEventKind::EpochStallBackfillEscalated`, and `MarmotEventFfi::EpochStallEscalated`. Nine copies of the same reasoning will drift the first time the threshold or the residual changes; consider one canonical home (the module header) with cross-references from the rest. The prose itself is high quality — there is just far too much of it for the invariants it protects.

**Test-only visibility.** `mod epoch_stall` widened to `pub(crate)` and `apply_backfill_decision` made `pub(crate)` so `src/tests.rs` can hand-construct `ArmAndEscalate`. That is a defensible tradeoff (the docs state it), but note both fabricated-decision tests would still pass if the detector stopped raising `ArmAndEscalate` entirely — only `repeated_arming_without_recovery_escalates_exactly_once` covers that, so it is load-bearing for the whole feature.

erskingardner (approved):

crates/marmot-app/src/client/epoch_stall.rs
jgmontoya: Agreed on both counts, and this lands in the detector follow-up PR already promised in the client-rebuild thread — the frozen-epoch gap, your clean-advance visibility fix, and a time-bound on runs are all run-semantics changes in the same file, so they'll ship together rather than piling onto this approved PR. We'll take the observe_group_epoch-from-advance_convergence_after_runtime_sync wiring (verifying the exact call-site mechanics there, given this file's history with ordering around fail_if_publish_failed), name the run-end semantics honestly in the docs, and design the aging bound alongside the frozen-epoch trigger since both change what "a run" means.

crates/marmot-app/src/client/sync.rs
jgmontoya: You're right — nothing in this PR needs that move. Reverting it rides the detector follow-up branch rather than a new push here (which would dismiss the approval); blame on this function will be restored there.

crates/marmot-app/tests/epoch_stall_backfill_audit.rs
jgmontoya: Agreed — the follow-up will switch advance_bobs_epoch to short-interval polling so a slow relay costs time instead of a misleading convergence-shaped failure, and we'll flag both long escalation tests as slow-tier candidates per #1249. The fabricated-decision visibility note lands there too: detector-driven escalation coverage so repeated_arming_without_recovery_escalates_exactly_once stops being single-handedly load-bearing.

PR #1212, opened 2026-07-31
fix(storage): redact group image keys from Debug

### LINKED ISSUE #1211: StoredAccountGroupComponent Debug exposes encrypted group-image capability keys
## Current behavior

The app-layer group-image types correctly treat `image_key_hex`, `image_upload_key_hex`, and the encoded `data_hex` as key material and redact them from `Debug`. When the same component is converted for SQLite projection, its `data_hex` is copied into `StoredAccountGroupComponent.component_data_hex`, whose derived `Debug` prints the field verbatim.

For `marmot.group.blossom.image.v1`, that field encodes the avatar decryption key and Blossom upload private key. Formatting a stored component with `{:?}` therefore exposes both capabilities despite the explicit app-layer invariant that these bytes must never be Debug-rendered. No current production tracing call was found formatting this type, so this is a structural/latent disclosure path rather than an observed active log leak.

1. Create or update an encrypted group image.
2. Convert the resulting `AppGroupRecord` with `stored_components_from_app_group`.
3. Locate the image `StoredAccountGroupComponent` and format it with `format!("{component:?}")`.
4. The rendered `component_data_hex` contains the encoded image decryption key and Blossom upload private key.

## Expected behavior

Every representation of the encrypted group-image component should redact its key-bearing raw component bytes from `Debug`. Key-bearing copies should also use bounded/zeroizing ownership where compatible with persistence and API requirements.

## Code evidence

- [`crates/marmot-app/src/media/group_image.rs:21-37`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/marmot-app/src/media/group_image.rs#L21-L37): the type documents both keys as key material, requires that they never be logged/Debug-rendered, and then carries them as strings into the MLS-protected component.
- [`crates/marmot-app/src/groups.rs:391-420`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/marmot-app/src/groups.rs#L391-L420): `AppGroupImageComponent` states that `data_hex` carries the same key bytes and redacts all three fields in its handwritten `Debug`.
- [`crates/marmot-app/src/groups.rs:1560-1584`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/marmot-app/src/groups.rs#L1560-L1584): the component format is `[hash, key, nonce, upload_key, media_type]`, and decoding extracts both key fields from those bytes.
- [`crates/marmot-app/src/conversions.rs:93-109`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/marmot-app/src/conversions.rs#L93-L109): projection conversion copies `group.image.data_hex` into `StoredAccountGroupComponent.component_data_hex`.
- [`crates/storage-sqlite/src/account_projection.rs:147-152`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/storage-sqlite/src/account_projection.rs#L147-L152): the storage type derives raw `Debug` over `component_data_hex`.

## Acceptance criteria

- [ ] `StoredAccountGroupComponent` cannot Debug-render raw `component_data_hex`; tests assert known image and upload keys are absent.
- [ ] Other generic component/projection/debug representations are checked for the same key-bearing image payload.
- [ ] No tracing, diagnostics, error, or FFI path formats the raw component bytes.
- [ ] Key-bearing temporary copies use zeroizing ownership where this does not conflict with the SQLCipher persistence boundary.

## Related

- [#397](https://github.com/marmot-protocol/mdk/issues/397) covered FFI group-avatar `Debug`; [#400](https://github.com/marmot-protocol/mdk/issues/400) covered the app-layer group-image key lifetime. Neither covers this generic SQLite projection type.
- [PR #699](https://github.com/marmot-protocol/mdk/pull/699) added app/FFI redaction for group-image keys, but did not redact `StoredAccountGroupComponent`.
- [#1000](https://github.com/marmot-protocol/mdk/issues/1000) concerns media epoch-secret read buffers, not group-image component `Debug`.
- [#673](https://github.com/marmot-protocol/mdk/issues/673) is a broad older zeroization issue and does not cover this current storage projection representation.
- No exact duplicate was found after searching all open/closed MDK issues and pull requests for group image, Blossom upload key, `component_data_hex`, raw component, `Debug`, and zeroization terms.

## Verification scope

Confirmed from the current conversion and type definitions. No current production formatting call was found. Rust regression execution was unavailable because this host has no Rust/Cargo toolchain.

agent-p1p:
Fixes #1211

## Summary
- replace derived `Debug` on `StoredAccountGroupComponent` with a handwritten implementation that always redacts raw `component_data_hex`
- redact the denormalized image decryption and Blossom upload keys from parent `StoredAccountGroup` formatting
- stop the TUI diagnostics model from retaining generic component `data_hex`; diagnostics now render only component names and ids
- cover direct storage formatting, real app-to-storage conversion, nested parent formatting, diagnostics `Debug`, and diagnostics lines with key sentinels
- document the change in the compatibility-cohort changelog

## Root cause
The SQLite projection copied the encrypted group-image component into `component_data_hex` and mirrored its two capability keys into `StoredAccountGroup`. Both storage types derived `Debug`, so direct or nested `{:?}` formatting exposed those values despite app-layer redaction.

The TUI diagnostics path independently copied each component's raw `data_hex` into a `Debug`-deriving DTO and rendered it on screen. For `marmot.group.blossom.image.v1`, that payload contains the same decryption and upload capabilities.

## Verification
RED evidence before the production changes:
- `cargo test -p storage-sqlite stored_account_group_component_debug_redacts_blossom_image_payload -- --nocapture` failed because rendered `component_data_hex` contained the image-key sentinel
- the updated TUI diagnostics regression failed because the prior diagnostics DTO and rendered line retained the image/upload-key sentinels

Final checks:
- `cargo test -p wn-cli diagnostics_panel_lines_show_mls_and_components --locked -- --nocapture`
- `cargo test -p marmot-app stored_group_image_component_debug_redacts_key_material --locked -- --nocapture`
- `cargo test -p storage-sqlite stored_account_group_component_debug_redacts_blossom_image_payload --locked -- --nocapture`
- `cargo test -p wn-cli --lib --locked` — 529 passed
- `cargo test -p storage-sqlite --locked` — 308 passed on the original fix head
- `cargo clippy -p storage-sqlite --all-targets --locked -- -D warnings`
- `cargo fmt --all --check`
- `just fast-ci`
- `git diff --check`

All passed. The storage suite emitted existing SQLCipher `mlock()` errno 12 warnings on this host; tests remained green.

## Audit and scope
- `StoredAccountGroupComponent`, nested `StoredAccountGroup`, and TUI diagnostics no longer `Debug`-format or render raw image component bytes.
- Repository searches found no production tracing, error, or UniFFI path formatting these stored raw bytes.
- Full ids and structured endpoints remain visible on storage DTO `Debug` by design. `docs/marmot-architecture/overview/observability.md` permits ids/pubkeys on DTOs and endpoints in structured DTO fields while forbidding them in tracing/logs; CodeRabbit's broader all-identifier redaction suggestion was therefore not applied.
- Stable `groups show --json` capability output is a separate API/trust-boundary decision tracked in #1253.

No additional zeroizing wrapper was introduced at the SQLCipher persistence DTO boundary: these strings are durable SQL-bound values rather than ephemeral crypto buffers, and wrapping them would add API/trait churn or extra plaintext copies without reducing persistence exposure. Existing ephemeral group-image crypto buffers remain zeroizing.

## Sensitive paths
- `crates/storage-sqlite/src/account_projection.rs` — security-sensitive projection `Debug` boundary
- `crates/storage-sqlite/src/account_projection/tests.rs` — storage regression sentinels
- `crates/marmot-app/src/tests.rs` — real encrypted group-image conversion regression
- `crates/cli/src/tui/model.rs` — removes key-bearing raw payload retention from diagnostics
- `crates/cli/src/tui/view.rs` — removes raw payload rendering
- `crates/cli/src/tui/tests.rs` — diagnostics `Debug` and rendered-line sentinels

This PR touches sensitive key-handling/projection paths and requires explicit human approval at the merge gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive encrypted image and upload key data is now hidden in account debug output.
  * Group diagnostics no longer display raw component payload data.
  * Safe identifying details, such as component names and IDs, remain visible while confidential values are replaced with a redaction marker.
  * Added regression coverage to help prevent accidental exposure of protected key material.

### DIFF
--- a/crates/marmot-app/src/tests.rs
+++ b/crates/marmot-app/src/tests.rs
@@ -21,7 +21,9 @@ use transport_nostr_peeler::NostrTransportEvent;
 use transport_quic_broker::BrokerServerTrust;
 
 use crate::audit_log::AUDIT_ID_BYTES;
-use crate::conversions::{app_group_from_stored_group, stored_group_from_app_group};
+use crate::conversions::{
+    app_group_from_stored_group, stored_components_from_app_group, stored_group_from_app_group,
+};
 use crate::directory::records::{
     FetchedFollowList, profile_content_json, public_directory_user_record,
 };
@@ -3279,6 +3281,48 @@ fn empty_follow_fetch_preserves_cached_edges() {
         selected.source_relays,
         vec!["wss://cached.example".to_owned()]
     );
+}
+
+#[test]
+fn stored_group_image_component_debug_redacts_key_material() {
+    const IMAGE_KEY_HEX: &str = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+    const UPLOAD_KEY_HEX: &str = "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb";
+
+    let group = AppGroupRecord::new(
+        "aa".to_owned(),
+        AppGroupNostrRoutingComponent::new(
+            NostrRoutingV1::new([0xAA; 32], vec!["wss://relay.example".to_owned()]).unwrap(),
+        )
+        .unwrap(),
+        "group".to_owned(),
+        String::new(),
+        AppGroupImageInput {
+            image_hash_hex: hex::encode([0x11; 32]),
+            image_key_hex: IMAGE_KEY_HEX.to_owned(),
+            image_nonce_hex: hex::encode([0x22; 12]),
+            image_upload_key_hex: UPLOAD_KEY_HEX.to_owned(),
+            media_type: Some("image/png".to_owned()),
+        },
+        AppGroupAdminPolicyComponent::new(Vec::new()),
+        AppGroupMessageRetentionComponent::disabled(),
+    );
+
+    let image_component = stored_components_from_app_group(&group)
+        .into_iter()
+        .find(|component| component.component_id == GROUP_BLOSSOM_IMAGE_COMPONENT_ID)
+        .expect("image component");
+
+    let rendered = format!("{image_component:?}");
+    assert!(!rendered.contains(IMAGE_KEY_HEX));
+    assert!(!rendered.contains(UPLOAD_KEY_HEX));
+    assert!(rendered.contains("marmot.group.blossom.image.v1"));
+    assert!(rendered.contains("redacted"));
+
+    let stored = stored_group_from_app_group(&group);
+    let parent_rendered = format!("{stored:?}");
+    assert!(!parent_rendered.contains(IMAGE_KEY_HEX));
+    assert!(!parent_rendered.contains(UPLOAD_KEY_HEX));
+    assert!(parent_rendered.contains("group"));
 }
 
 #[test]
--- a/crates/marmot-app/src/tests.rs
+++ b/crates/marmot-app/src/tests.rs
@@ -3322,7 +3322,9 @@ fn stored_group_image_component_debug_redacts_key_material() {
     let parent_rendered = format!("{stored:?}");
     assert!(!parent_rendered.contains(IMAGE_KEY_HEX));
     assert!(!parent_rendered.contains(UPLOAD_KEY_HEX));
-    assert!(parent_rendered.contains("group"));
+    assert!(parent_rendered.contains("profile_name: \"group\""));
+    assert!(parent_rendered.contains("image_key_hex"));
+    assert!(parent_rendered.contains("redacted"));
 }
 
 #[test]
--- a/crates/storage-sqlite/src/account_projection.rs
+++ b/crates/storage-sqlite/src/account_projection.rs
@@ -106,6 +106,11 @@ pub struct StoredAccountState {
     pub groups: Vec<StoredAccountGroup>,
 }
 
+/// `image_key_hex`/`image_upload_key_hex` are key material mirrored from the
+/// blossom image component for chat-list projection. They are stored in
+/// SQLCipher, but must not appear in `Debug` output. Nested
+/// [`StoredAccountGroupComponent`] values redact in-band component bytes that
+/// carry the same keys.
 #[derive(Clone, PartialEq, Eq)]
 pub struct StoredAccountGroup {
     pub group_id_hex: String,
@@ -138,11 +143,6 @@ pub struct StoredAccountGroup {
     pub components: Vec<StoredAccountGroupComponent>,
 }
 
-/// `image_key_hex`/`image_upload_key_hex` are key material mirrored from the
-/// blossom image component for chat-list projection. They are stored in
-/// SQLCipher, but must not appear in `Debug` output. Nested
-/// [`StoredAccountGroupComponent`] values redact in-band component bytes that
-/// carry the same keys.
 impl fmt::Debug for StoredAccountGroup {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("StoredAccountGroup")
@@ -179,6 +179,10 @@ pub struct StoredNostrRoute {
     pub last_epoch: u64,
 }
 
+/// `component_data_hex` carries MLS-protected component bytes. Blossom image
+/// payloads embed avatar decryption and Blossom upload keys; other components
+/// may carry sensitive policy bytes. SQLCipher protects persistence, but
+/// `Debug` must never render raw component bytes.
 #[derive(Clone, PartialEq, Eq)]
 pub struct StoredAccountGroupComponent {
     pub component_id: u16,
@@ -186,10 +190,6 @@ pub struct StoredAccountGroupComponent {
     pub component_data_hex: String,
 }
 
-/// `component_data_hex` carries MLS-protected component bytes. Blossom image
-/// payloads embed avatar decryption and Blossom upload keys; other components
-/// may carry sensitive policy bytes. SQLCipher protects persistence, but
-/// `Debug` must never render raw component bytes.
 impl fmt::Debug for StoredAccountGroupComponent {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("StoredAccountGroupComponent")
--- a/crates/storage-sqlite/src/account_projection.rs
+++ b/crates/storage-sqlite/src/account_projection.rs
@@ -10,6 +10,7 @@ use rusqlite::{
     types::{Type, Value},
 };
 use serde::{Deserialize, Serialize};
+use std::fmt;
 
 const SECURE_DELETE_RETENTION_OPERATION: &str = "retention";
 const SECURE_DELETE_LOCAL_GROUP_OPERATION: &str = "local_group";
@@ -105,7 +106,7 @@ pub struct StoredAccountState {
     pub groups: Vec<StoredAccountGroup>,
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct StoredAccountGroup {
     pub group_id_hex: String,
     pub endpoint: String,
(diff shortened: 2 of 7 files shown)


crates/storage-sqlite/src/account_projection.rs
```diff
@@ -137,20 +138,68 @@ pub struct StoredAccountGroup {
     pub components: Vec<StoredAccountGroupComponent>,
 }
 
+/// `image_key_hex`/`image_upload_key_hex` are key material mirrored from the
```
erskingardner: Nit: every other key-material redaction in this repo puts this doc comment on the **struct**, above the derive, not on the `impl` block — see `crates/storage-sqlite/src/chat_list.rs:68` (`ChatListAvatar`), `crates/marmot-app/src/groups.rs:391` (`AppGroupImageComponent`), and `crates/marmot-uniffi/src/conversions/chat_list.rs:13` (`ChatListAvatarFfi`).

Worth matching: the warning is most useful where someone is editing the field list, and on the struct it shows up in rustdoc for the type rather than for the impl. Same applies to the `StoredAccountGroupComponent` doc at line 189.

crates/marmot-app/src/tests.rs
```diff
@@ -3281,6 +3283,48 @@ fn empty_follow_fetch_preserves_cached_edges() {
...
+    assert!(!rendered.contains(IMAGE_KEY_HEX));
+    assert!(!rendered.contains(UPLOAD_KEY_HEX));
+    assert!(rendered.contains("marmot.group.blossom.image.v1"));
+    assert!(rendered.contains("redacted"));
+
+    let stored = stored_group_from_app_group(&group);
+    let parent_rendered = format!("{stored:?}");
+    assert!(!parent_rendered.contains(IMAGE_KEY_HEX));
+    assert!(!parent_rendered.contains(UPLOAD_KEY_HEX));
+    assert!(parent_rendered.contains("group"));
```
erskingardner: This assertion is vacuous — `"group"` is a substring of the struct name `StoredAccountGroup` and of field names like `group_id_hex`, so it holds no matter what the impl does. (The literal `"group".to_owned()` profile name it is presumably checking is also indistinguishable from those.)

Either drop it, or assert something that actually pins the behavior, e.g. `assert!(parent_rendered.contains("<redacted>"))` plus a non-secret field that must survive:

```rust
assert!(parent_rendered.contains("profile_name: \"group\""));
```

erskingardner (dismissed):
Approve. The fix is small, surgical, and matches the existing key-material redaction pattern in this repo (`chat_list.rs::ChatListAvatar`, `groups.rs::AppGroupImageComponent`, `ChatListAvatarFfi`). Both the component and the denormalized parent copies are covered, `StoredAccountState` is covered transitively, neither type derives `Serialize`, and the RED-then-GREEN test evidence in the description is exactly the right shape. +120/-3 with two regression tests is proportionate.

One substantive note, non-blocking and out of scope for this PR:

**The audit claim in the description is overbroad.** "Repository searches found no tracing, diagnostic, error, or UniFFI path formatting `StoredAccountGroupComponent` or raw `component_data_hex`" is true for that *type*, but the same key bytes are rendered elsewhere:

- `crates/cli/src/tui/model.rs:4638-4656` — `group_component_diagnostics` explicitly includes the `"image"` component and copies its `data_hex` into `GroupComponentDiagnostics`, which derives `Debug`.
- `crates/cli/src/tui/view.rs:332-344` — `group_component_diagnostics_line` prints `data={data_hex}` straight to the terminal, so the group-diagnostics screen renders the avatar decryption key and Blossom upload secret verbatim.
- `crates/cli/src/commands/groups.rs:56` — `groups show --json` serializes `group.image` (`AppGroupImageComponent` derives `Serialize`), emitting `image_key_hex`, `image_upload_key_hex`, and `data_hex` to stdout.

Those are local-operator surfaces where the user already holds the keys, so they may well be intentional — but issue #1211's acceptance criteria 2 and 3 ("other generic component/projection/debug representations", "no tracing, diagnostics, error, or FFI path formats the raw component bytes") are only partially met. Suggest narrowing the claim in the description and filing a follow-up rather than growing this PR.

On criterion 4 (zeroizing ownership): the documented rationale for not wrapping durable SQL-bound strings is sound; I'd accept it as-is.

Minor: these hand-written `Debug` impls have no compile-time guard against field drift — a future key-bearing field added to `StoredAccountGroup` will be silently omitted rather than silently leaked, so the failure mode is safe. No action needed; noting it since the file now carries an 18-field manual impl.

erskingardner:
## Review summary

**Approved with nits.** The fix itself is correct and surgical (+120/-3, 4 files). Manual `Debug` impls redacting key material match three existing precedents in the repo (`ChatListAvatar`, `AppGroupImageComponent`, `ChatListAvatarFfi`), cover the denormalized parent copies, and `StoredAccountState` is covered transitively. Neither type derives `Serialize`, so there is no second leak path. The RED-then-GREEN evidence in the description is the right shape.

### 1. Audit claim is overbroad (non-blocking, out of scope)

The same key bytes are still rendered elsewhere:

- `crates/cli/src/tui/model.rs:4638` pulls the `image` component's `data_hex` into a `Debug`-deriving struct
- `crates/cli/src/tui/view.rs:332` prints it verbatim on the diagnostics screen
- `crates/cli/src/commands/groups.rs:56` serializes `group.image` (both keys plus `data_hex`) in `groups show --json`

These are local-operator surfaces so this is likely intentional, but issue #1211's criteria 2 and 3 are only partially met. Suggest narrowing the claim in the description and filing a follow-up rather than expanding scope here.

### 2. Nit: doc comment placement

The new doc comments sit on the `impl` blocks; all three existing precedents put them on the struct above the derive.

### 3. Nit: vacuous assertion

`assert!(parent_rendered.contains("group"))` in the marmot-app test passes regardless of behavior — `"group"` is a substring of the struct and field names.

### On zeroizing

Criterion 4 was explicitly declined with a sound documented rationale for durable SQL-bound strings; I accept that. CodeRabbit's warning on that point is, in my view, not actionable.

agent-p1p:
## Adversarial review — changes required

Reviewed exact head `f5b4cb0813442c9be5fb91b7a7baa9ca81510edd`.

### Blocking

1. **The TUI diagnostics path still retains, `Debug`-formats, and renders the same key-bearing image payload.**

   `group_component_diagnostics` includes the `image` component and copies its raw `data_hex` into `GroupComponentDiagnostics` (`crates/cli/src/tui/model.rs:440-445`, `4638-4653`). That generic diagnostics DTO derives `Debug`, and `group_component_diagnostics_line` prints `data_hex` verbatim (`crates/cli/src/tui/view.rs:332-343`). For `marmot.group.blossom.image.v1`, those bytes contain the avatar decryption key and Blossom upload secret.

   This leaves issue #1211's explicit generic-representation and diagnostics acceptance criteria unmet. The storage DTO redaction in this PR is correct, but it does not close this sibling path.

   Prefer the smaller fix: do not retain raw component bytes in the diagnostics DTO at all; render only component name/id (or a constant redaction marker). Add a regression containing known image/upload-key sentinels and assert they are absent from both `format!("{diagnostics:?}")` and the rendered diagnostics lines.

### Suggestion

- `crates/marmot-app/src/tests.rs:3325` uses `parent_rendered.contains("group")`, which is vacuous because the struct/field names themselves contain `group`. Assert a specific safe structural field/value or the expected redaction markers instead.

### Verified / not blocking

- Direct `StoredAccountGroupComponent` formatting and nested `StoredAccountGroup` formatting redact both sentinels.
- The zeroization rationale at the durable SQLCipher DTO boundary is sound; no wrapper churn is warranted here.
- CodeRabbit's request to hide every id/URL from DTO `Debug` is not required by the documented observability table: structured DTO `Debug` may carry those values; tracing/logging may not. I found no production formatting call for these storage DTOs.
- The `Unreleased` changelog entry is present, and there is no version/cohort bump.

### Out-of-scope follow-up

Filed #1253 to decide the broader trust boundary for group-image capabilities in stable CLI/daemon JSON. That API decision is separate from this PR's storage `Debug` fix.

### Verification

- `cargo test -p storage-sqlite stored_account_group_component_debug_redacts_blossom_image_payload --locked -- --nocapture` — pass
- `cargo test -p marmot-app stored_group_image_component_debug_redacts_key_material --locked -- --nocapture` — pass
- `cargo test -p wn-cli diagnostics_panel_lines_show_mls_and_components --locked -- --nocapture` — pass; confirms the current diagnostics view intentionally renders raw component `data_hex`
- `cargo clippy -p storage-sqlite --all-targets --locked -- -D warnings` — pass
- `cargo fmt --all --check` — pass
- `git diff --check origin/master...HEAD` — pass

crates/storage-sqlite/src/account_projection.rs
agent-p1p: Moved both key-safety docs onto their structs in d92eb8ca, matching the existing redaction precedents and putting the warning next to the field lists.

crates/marmot-app/src/tests.rs
agent-p1p: Replaced the substring check in d92eb8ca with exact `profile_name: "group"` and explicit redaction-field assertions. The focused marmot-app regression passes.

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -137,20 +138,68 @@ pub struct StoredAccountGroup {
...
+            .field("member_count", &self.member_count)
+            .field("welcomer_account_id_hex", &self.welcomer_account_id_hex)
+            .field(
+                "via_welcome_message_id_hex",
+                &self.via_welcome_message_id_hex,
+            )
+            .field("nostr_routing_last_epoch", &self.nostr_routing_last_epoch)
+            .field("prior_nostr_routes", &self.prior_nostr_routes)
+            .field("self_membership", &self.self_membership)
+            .field("components", &self.components)
```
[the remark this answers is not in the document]
agent-p1p: Verified against the current policy and intentionally did not apply this broader expansion. `docs/marmot-architecture/overview/observability.md` allows ids/pubkeys on DTO `Debug` and endpoints in structured DTO fields; it forbids them in tracing/logs. Repository searches found no production formatting call for these storage DTOs. This PR keeps the key-material redaction boundary; the actual sibling TUI payload leak was removed in d92eb8ca.

agent-p1p:
Addressed the adversarial blocking finding in d92eb8ca: TUI diagnostics no longer retain `data_hex`, derive it through `Debug`, or render it on screen. Added image/upload-key sentinels covering both diagnostics `Debug` and panel lines; focused regressions, 529 wn-cli unit tests, formatting, and `just fast-ci` pass. The two human nits are also fixed. Stable `groups show --json` remains the separate #1253 decision.

PR #1213, opened 2026-07-31
fix(account): zeroize NIP-49 export intermediates

### LINKED ISSUE #474: [Feature]: FFI: export private key as ncryptsec1 (NIP-49 password-encrypted backup)
**What**
Implement [NIP-49](https://github.com/nostr-protocol/nips/blob/master/49.md) password-encrypted private-key export and expose it via UniFFI. Lets the app offer an `ncryptsec1...` backup format that is safe to write down or store in a password manager because it requires the user's passphrase to decrypt.

Grep on 2026-06-20: zero matches for `ncryptsec`, `NIP-49`, `scrypt`, or `XChaCha20` in the engine — this is a clean greenfield addition.

**API shape (suggested)**
```rust
pub fn export_encrypted_secret_key(
    account_id: AccountId,
    passphrase: &str,
    log_n: u8,  // default 18 per NIP-49 mobile sweet spot
) -> Result<String, EncryptedExportError>;
```

Returns the canonical `ncryptsec1...` bech32 string (91 bytes pre-bech32).

**NIP-49 spec summary (what the engine does)**
1. Passphrase normalized to Unicode NFKC.
2. SALT = 16 random bytes.
3. SYMMETRIC_KEY = scrypt(password=PASSPHRASE, salt=SALT, log_n, r=8, p=1) → 32 bytes.
4. NONCE = 24 random bytes.
5. KEY_SECURITY_BYTE = 0x00 if the engine has ever revealed the raw key (see sibling issue), 0x01 if not, 0x02 if untracked. Use the persisted byte from sibling issue marmot-protocol/mdk#475.
6. CIPHERTEXT = XChaCha20-Poly1305(plaintext=PRIVATE_KEY, ad=KEY_SECURITY_BYTE, nonce=NONCE, key=SYMMETRIC_KEY).
7. VERSION_BYTE = 0x02.
8. Concatenation: `VERSION || LOG_N || SALT || NONCE || AD || CIPHERTEXT` = 91 bytes.
9. Bech32 encode with HRP `ncryptsec`.
10. Zero out scrypt-derived key and any plaintext key memory after use.

**Rust crate suggestions**
- `scrypt = "0.11"` — well-vetted, RustCrypto org.
- `chacha20poly1305 = "0.10"` — RustCrypto, includes XChaCha20-Poly1305 via `XChaCha20Poly1305`.
- Existing bech32 + secp256k1 deps already present.
- `unicode-normalization` for NFKC on the passphrase.
- `subtle` for constant-time operations where applicable.

**Default log_n recommendation: 18 (256 MiB, ~1s on a fast mobile)**
From the NIP-49 table:
| log_n | RAM | time |
|-------|-----|------|
| 16 | 64 MiB | 100 ms |
| 18 | 256 MiB | ~1 s |
| 20 | 1 GiB | 2 s |

18 is the mobile sweet spot — strong brute-force resistance without OOM risk on any modern phone. Don't let callers pick log_n in v1 (footgun); ship 18 fixed and revisit later.

**Error types (typed)**
- `EmptyPassphrase` — refuse to encrypt with empty passphrase.
- `AccountNotFound`.
- `KeystoreLocked`.
- `EncryptionFailed(reason)` — wrap underlying crypto errors.

**Behavior requirements**
- Active-account scoped — refuse cross-account export.
- Logs the encrypted-export event to the per-account audit log (without the passphrase).
- Passphrase is borrowed `&str`, zeroed by caller after the FFI returns (UniFFI String → owned on entry; engine zeros its copy via `zeroize` crate on drop).
- Does NOT mark the account `KEY_SECURITY_BYTE = 0x00` on encrypted export — the raw key did not leave the engine in plaintext, so the security posture is preserved.

**Out of scope**
- Import / decrypt of `ncryptsec1...` for account restore — separate feature, pairs with multi-device or restore flows.
- User-configurable log_n.
- QR code render (consumer concern).
- Mnemonic / BIP39 export — separate engine issue.

**Test plan**
- Encrypt with passphrase "test123", log_n=18 → returns `ncryptsec1...` of expected length.
- Decrypt the result with a reference NIP-49 implementation (e.g., the `nostr-tool` CLI or python `coincurve` + manual NIP-49 decode) → yields the same raw 32-byte secret key the engine holds.
- Encrypt with empty passphrase → returns `EmptyPassphrase` error, no output.
- Verify the `KEY_SECURITY_BYTE` in the AD matches the persisted account state from sibling issue marmot-protocol/mdk#475.
- Verify Unicode NFKC normalization with the NIP-49 test vector ("ÅΩẛ̣" → "ÅΩẛ̣").
- Run against the NIP-49 spec's published decrypt test vector (passphrase="nostr", log_n=16) to confirm interop.
- Verify scrypt + ChaCha20 keys are zeroed after the function returns (instrument with a custom allocator or rely on `zeroize::Zeroizing` wrapper).

**Why MEDIUM**
Unblocks `marmot-protocol/darkmatter-android#486` (encrypted backup UI). The encrypted variant is the *safer* backup format — should ship alongside or shortly after the raw nsec export (sibling issue marmot-protocol/mdk#475). Not HIGH because no user is currently locked out without it, but it's the recommended path for the backup feature and should land before users start manually transcribing raw nsecs everywhere.

**Related**
- `marmot-protocol/darkmatter-android#486` — consumer of this FFI.
- Sibling engine: marmot-protocol/mdk#475 — raw nsec export FFI; this issue uses the same keystore-read path.
- Sibling engine: NIP-06 mnemonic export FFI (filed alongside) — `unrecommended` per spec.
- NIP-49: https://github.com/nostr-protocol/nips/blob/master/49.md

agent-p1p:
## Summary

- pin the workspace's `nostr` 0.44.6 dependency to the v0.44 zeroization fix in nostrdevkit/nostr#1406
- ensure NIP-49 normalized passphrases, scrypt-derived keys, cipher-owned key copies, and decrypted plaintext buffers are wiped on success and error paths
- document the remaining FFI ownership boundary: Rust wipes its owned passphrase, but cannot wipe the caller's original host-language string
- record the security fix in the shared MDK changelog

The upstream commit adds focused `ZeroizeOnDrop` checks and preserves the published NIP-49 vector plus NFKC behavior. MDK uses a crates.io patch so the same fixed `nostr` package remains unified across `nostr-sdk` and direct dependencies.

Fixes #474

## Upstream

- nostrdevkit/nostr#1406
- pinned commit: `c209c22aab4883a01b1d6017752b23245afce9b3`

## Verification

Upstream:
- `cargo test -p nostr --features nip49 nip49::tests` (5 passed)
- `cargo check -p nostr --no-default-features --features 'alloc,nip49'`
- `cargo clippy -p nostr --all-targets --features nip49 -- -D warnings`

MDK:
- `cargo test -p marmot-account nip49 --locked -- --nocapture` (8 passed)
- `cargo test -p marmot-uniffi --lib --locked` (94 passed)
- `just fast-ci`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,3 +138,6 @@ marmot-terminal-harness = { path = "integrations/terminal-harness" }
 # release contains it, then remove this patch and update the libcrux path.
 [patch.crates-io]
 hax-lib-macros = { git = "https://github.com/cryspen/hax.git", rev = "8f9cb576e58f6cc7e9ed249a7d4439b0f7ed0da7" }
+# nostr 0.44.6 leaves NIP-49 normalized passwords and derived keys in ordinary
+# buffers. Pin the upstream v0.44 fix until PR #1406 is released.
+nostr = { git = "https://github.com/agent-p1p/nostr.git", rev = "c209c22aab4883a01b1d6017752b23245afce9b3" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3460,8 +3460,7 @@ dependencies = [
 [[package]]
 name = "nostr"
 version = "0.44.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e826dd648489de2c5b293920e20b92932ef820302007c1987c758d4d06eeb2cf"
+source = "git+https://github.com/agent-p1p/nostr.git?rev=c209c22aab4883a01b1d6017752b23245afce9b3#c209c22aab4883a01b1d6017752b23245afce9b3"
 dependencies = [
  "base64",
  "bech32",
@@ -3479,6 +3478,7 @@ dependencies = [
  "serde_json",
  "unicode-normalization",
  "url",
+ "zeroize",
 ]
 
 [[package]]
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -65,6 +65,12 @@ versioning through the workspace version in the root `Cargo.toml`.
   dropped on the send path, so storage showed the new state while live
   subscribers were never notified.
 
+### Security
+
+- NIP-49 encrypted private-key export now zeroizes its normalized passphrase,
+  scrypt-derived key, cipher-owned key copy, and plaintext decrypt buffer on
+  success and error paths.
+
 ### Changed
 
 - The bundled SQLCipher stack now uses rusqlite 0.40.1/libsqlite3-sys 0.38.1,
--- a/crates/marmot-uniffi/src/commands/account.rs
+++ b/crates/marmot-uniffi/src/commands/account.rs
@@ -420,7 +420,9 @@ impl Marmot {
     /// `ncryptsec1...` bech32 backup string (mdk#544).
     ///
     /// SENSITIVE: the passphrase is accepted as an owned FFI string and zeroed
-    /// on return by the Rust boundary. The encrypted export is logged to the
+    /// on return by the Rust boundary. This cannot wipe the caller's original
+    /// host-language string, which remains a separate host-side responsibility
+    /// and should be kept transient. The encrypted export is logged to the
     /// per-account audit log, but unlike `reveal_nsec` it does not downgrade the
     /// account's NIP-49 KEY_SECURITY_BYTE because raw plaintext key material is
     /// not returned to the host app.


Cargo.toml
```diff
@@ -138,3 +138,6 @@ marmot-terminal-harness = { path = "integrations/terminal-harness" }
 # release contains it, then remove this patch and update the libcrux path.
 [patch.crates-io]
 hax-lib-macros = { git = "https://github.com/cryspen/hax.git", rev = "8f9cb576e58f6cc7e9ed249a7d4439b0f7ed0da7" }
+# nostr 0.44.6 leaves NIP-49 normalized passwords and derived keys in ordinary
+# buffers. Pin the upstream v0.44 fix until PR #1406 is released.
+nostr = { git = "https://github.com/agent-p1p/nostr.git", rev = "c209c22aab4883a01b1d6017752b23245afce9b3" }
```
erskingardner, line 143: **Blocking: this pin forfeits the nostr 0.44.7 security release.**

The pinned rev `c209c22` has exactly one parent, `959538dd` = *"nostr: bump to v0.44.6"*. So this commit is `0.44.6 + zeroize`, nothing else. Because `[patch.crates-io]` replaces the source for the entire graph, it freezes `nostr` (and the copy pulled in transitively by `nostr-sdk` / `nostr-relay-builder`) at 0.44.6 and makes 0.44.7 unreachable.

Upstream `v0.44` is **15 commits ahead** of the pinned rev and cut `0.44.7` on 2026-08-02 (published to crates.io 11:13Z). Those 15 commits are almost entirely security fixes:

- `c0589152` nostr: limit NIP-49 scrypt resource usage — *same code path this PR touches*
- `fe4c30de` nostr: bound NIP-44 v2 decryption payloads
- `b9194d8d` nostr: bound NIP-98 authorization events
- `3bd06173` nostr: avoid NIP-04 parsing amplification
- `258a34ce` nostr: redact NIP-46 secrets from debug output
- `02b05aca` nostr: redact NIP-60 credentials from debug output
- `3fcb2f21` nostr: handle empty search filters without panicking
- `2ee873ad` / `d0d4b6f5` nostr: authenticate NIP-47 / NIP-60 wallet events

The workspace builds `nostr` with `features = ["std", "nip49", "nip59"]` and NIP-44 sits under the giftwrap transport path, so the NIP-44 v2 bound fix and the scrypt-cost limit are directly load-bearing here. As written this trades seven-plus shipped fixes for one defence-in-depth zeroize — a net security regression against simply bumping to `nostr = "0.44.7"`.

Please rebase the branch onto `v0.44` head (`94dac28e`) so the pin is `0.44.7 + zeroize`, and re-run the verification matrix against that tree.

erskingardner, line 143: **Prefer an upstream-hosted rev over a personal fork.**

This routes the crate that owns *all* Nostr key material through a fork under an individual contributor account. A `rev` pin bounds the content, but the fork's availability and history are outside the project's control, and the commit carries no maintainer sign-off.

Upstream already has the identical change: `nostrdevkit/nostr@84992e8`, landed on `master` 2026-08-02 with `Signed-off-by: Yuki Kishimoto`, commit message *"Replaces https://github.com/nostrdevkit/nostr/pull/1406"*. It is on the v0.45 line, not `v0.44` — and the maintainer explicitly asked for a v0.44 backport PR (*"We can keep this one open as the backport PR for it"*) before #1406 was closed.

Opening that backport and pinning to the resulting `nostrdevkit/nostr` commit removes the personal fork from the trust path entirely, and would land on top of 0.44.7 rather than 0.44.6 — resolving the other comment at the same time. If the backport can't be turned around quickly, at minimum re-point at a fork branch rebased on `v0.44` head and say so in the comment.

(Noting the `openmls` pin above is also a fork, so forks per se are established practice here — the difference is that an upstream-hosted commit is available in this case.)

erskingardner, line 143: **The comment is factually wrong and will mislead whoever removes the pin.**

`nostrdevkit/nostr#1406` was **closed unmerged** on 2026-08-02, not merged — it will never "be released". A maintainer cherry-picked it to `master` as `84992e8` instead. The `v0.44` backport that would actually produce a released fix on this line has not happened yet, so the stated removal trigger never fires and this patch silently becomes permanent.

Reword to name the real upstream commit and a removal condition that can actually be observed, matching the migration-note style of the `openmls` block above — e.g. "replaced upstream by nostrdevkit/nostr@84992e8 on master; remove once the v0.44 backport ships in nostr 0.44.8 or the workspace moves to 0.45."

erskingardner (changes requested):
The premise is sound and the change is admirably small — 14 lines, no speculative abstraction, crypto delegated upstream instead of hand-rolled in `marmot-account`, and the new FFI doc comment about the host-language passphrase copy is accurate and genuinely worth having. Requesting changes on the dependency pin itself, not the goal.

**Blocking**

1. **The pin forfeits the nostr 0.44.7 security release.** `c209c22`'s only parent is `959538dd` ("bump to v0.44.6"), so the pinned tree is `0.44.6 + zeroize`. `[patch.crates-io]` replaces the source graph-wide, which freezes `nostr` — including the copy `nostr-sdk` pulls in — at 0.44.6 and makes 0.44.7 unreachable. Upstream `v0.44` is 15 commits ahead and shipped 0.44.7 on 2026-08-02 with the NIP-49 scrypt resource limit (the very code path this PR touches), NIP-44 v2 decryption bounds, NIP-98 bounds, NIP-04 amplification, NIP-46/NIP-60 debug redaction, NIP-47/60 event authentication, and a search-filter panic fix. Given this workspace's `nip49`/`nip59` feature set, that is a net security regression versus just bumping to `0.44.7`. Rebase onto `v0.44` head (`94dac28e`) so the pin is `0.44.7 + zeroize`.

2. **Pin at an upstream-hosted rev rather than a personal fork.** The identical change is already upstream at `nostrdevkit/nostr@84992e8` (`master`, maintainer-signed, "Replaces #1406"), and the maintainer explicitly invited a `v0.44` backport PR before #1406 was closed. Landing that backport and pinning to it removes a contributor-owned fork from the trust path for the crate that handles all key material — and fixes (1) at the same time.

3. **The `Cargo.toml` comment is wrong.** "until PR #1406 is released" — #1406 was closed unmerged on 2026-08-02 and will never be released. As written, the stated removal trigger never fires and the patch quietly becomes permanent. Point it at the real upstream commit and an observable removal condition, matching the `openmls` migration-note style directly above.

**Non-blocking**

- `Fixes #474` will auto-close a feature issue this PR did not implement. The NIP-49 export FFI already shipped (`crates/marmot-account/src/nip49_export.rs`; the doc comment cites mdk#544) — this PR only satisfies #474's last test-plan bullet ("verify scrypt + ChaCha20 keys are zeroed"). If closing #474 is intended, worth saying so explicitly; otherwise use `Refs #474`.
- The verification list cites `cargo test -p marmot-account nip49` and `-p marmot-uniffi`, but no code in either crate changed beyond a doc comment. That is fine as regression evidence — just note the runs prove the patched dependency still satisfies the published NIP-49 vector, not that the zeroization happened. The zeroization itself is only covered by the upstream `ZeroizeOnDrop` checks, which is the right place for it.

Once the pin sits on top of 0.44.7 and references an upstream commit, this is a straightforward merge.

erskingardner:
**Request changes** — summary of the review above.

This PR pins the workspace `nostr` dep to a fork commit for a NIP-49 zeroize fix. Good goal, but the pin as written is a net security regression:

- **The pinned rev is `0.44.6 + zeroize`.** Its only parent is the v0.44.6 bump, and `[patch.crates-io]` applies graph-wide, so it freezes `nostr` (including `nostr-sdk`'s transitive copy) at 0.44.6 and makes **0.44.7 unreachable** — that release (2026-08-02) carries the NIP-49 scrypt resource limit in the same code path, plus NIP-44 v2 decryption bounds, NIP-98 bounds, NIP-04 amplification, and NIP-46/60 debug redaction. A plain bump to 0.44.7 isn't sufficient either (verified: the published crate lacks the zeroize fix) — the branch should rebase onto upstream `v0.44` head.
- **Personal fork in the trust path is avoidable.** Upstream PR nostrdevkit/nostr#1406 was closed unmerged but cherry-picked to `master` as `84992e8`, with a maintainer explicitly inviting a `v0.44` backport. Pinning an upstream commit keeps a contributor fork out of the supply chain for the crate holding all key material.
- **The `Cargo.toml` comment is wrong** — "until PR #1406 is released" describes a trigger that can never fire, so the patch silently becomes permanent.
- Non-blocking: `Fixes #474` would auto-close a feature issue this PR didn't implement (that FFI already exists in `crates/marmot-account/src/nip49_export.rs`).

agent-p1p:
Pip adversarial review at exact head `7d98c910352afa997a04060495f4df6e5e260f18`.

BLOCKING:
- Confirmed existing dependency-pin findings remain unresolved: `Cargo.toml:141-143`, `Cargo.lock:3462-3463`. I independently verified that `c209c22` has parent `959538dd` (the v0.44.6 bump), while protected upstream `v0.44` is 15 commits ahead at `94dac28e` (v0.44.7). Those commits include the NIP-49 scrypt resource bound and several other security fixes. `[patch.crates-io]` replaces `nostr` graph-wide, including `nostr-sdk`, `nostr-relay-builder`, and `nostr-database`, so the current pin forfeits those fixes. The existing review is also correct that #1406 was closed unmerged, its stated removal trigger cannot fire, and a maintainer-signed equivalent exists upstream on master as `84992e8`; the required v0.44 backport still does not exist.
- This PR now has code-red CI: `Rust tests (4/4)` failed in https://github.com/marmot-protocol/mdk/actions/runs/30899889398/job/91961470038. `marmot-app::tests::runtime_start_returns_before_initial_directory_subscription_registration` exhausted its five-second deadline on both nextest attempts. The exact focused test passes locally 3/3 at this head, so this appears unrelated/flaky, but it is still a code test failure. Per Pip's CI provenance gate, a later green rerun does not make this PR eligible to merge. Filed #1254 to track the out-of-scope test reliability defect.

SUGGESTION:
- `crates/cli/CHANGELOG.md:47-49` describes encrypted export as zeroizing a “plaintext decrypt buffer,” but MDK production code only calls the dependency's encrypt path; decrypt is used only by tests. Scope the product-facing entry to the export-path intermediates, or clearly attribute decrypt-buffer hardening to the bundled dependency.

NITPICK:
- none

SUMMARY: The upstream zeroization change itself is small and technically sound: the normalized password, derived key, cipher-owned copy, and decrypted bytes have zeroizing owners; NFKC/vector compatibility remains covered; the FFI ownership-boundary note is accurate. Fresh verification passed 8/8 focused MDK NIP-49 tests, 95/95 UniFFI tests, 5/5 pinned-upstream NIP-49 tests, and the failed CI test 3/3 locally. The unresolved 0.44.6 dependency regression and the code-red CI history remain blocking.
VERDICT: REQUEST_CHANGES
SENSITIVE_PATHS: `Cargo.toml`, `Cargo.lock`, `crates/marmot-uniffi/src/commands/account.rs`, upstream `crates/nostr/src/nips/nip49.rs` (NIP-49 passphrase, derived-key, cipher-key, and plaintext-secret lifetimes)

PR #1214, opened 2026-07-31
fix(storage): zeroize OpenMLS serialization buffers

### LINKED ISSUE #1210: OpenMLS secrets pass through non-zeroizing SQLite serialization buffers
## Current behavior

MDK's custom OpenMLS SQLite provider serializes every OpenMLS entity through ordinary `Vec<u8>` JSON buffers before SQLCipher persistence and reads database blobs back through ordinary vectors before deserialization. The same generic path handles message secrets, group epoch secrets, MLS signature keypairs, HPKE keypairs, epoch keypairs, and PSKs.

The final database is SQLCipher-encrypted, and the pinned OpenMLS `SignatureKeyPair` protects its private field with `SecretVLBytes` while owned by OpenMLS. Serde deliberately emits that private field as bytes, however, so MDK's detached JSON vector contains plaintext private/epoch material and does not inherit the source object's zeroization. Freed client-owned buffers can therefore retain MLS signing and decryption material after persistence or load.

This is defense-in-depth against local memory disclosure, crash dumps, or swap, not a demonstrated remote exfiltration path.

1. Create/load a group or rotate MLS credentials so OpenMLS invokes a secret-bearing storage method such as `write_signature_key_pair`, `write_group_epoch_secrets`, or `write_encryption_key_pair`.
2. The provider forwards the entity to `write_entity`/`write_value`.
3. `serde_json::to_vec(value)` creates an ordinary secret-bearing vector and passes it into SQLite.
4. On load, `read_value_on_connection` returns another ordinary vector before `serde_json::from_slice` reconstructs the zeroizing OpenMLS object.

## Expected behavior

Temporary Rust-owned serialization and deserialization buffers for secret-bearing OpenMLS entities should zeroize on success and all error paths. Public-only OpenMLS entities can continue using ordinary buffers.

## Code evidence

- [`crates/storage-sqlite/src/openmls_storage/provider.rs:123-180`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/storage-sqlite/src/openmls_storage/provider.rs#L123-L180): message secrets, epoch secrets, and MLS signature keypairs all enter the generic entity writer.
- [`crates/storage-sqlite/src/openmls_storage/provider.rs:183-215`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/storage-sqlite/src/openmls_storage/provider.rs#L183-L215): HPKE and epoch keypairs are serialized through the same ordinary vectors.
- [`crates/storage-sqlite/src/openmls_storage/value_store.rs:50-58`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/storage-sqlite/src/openmls_storage/value_store.rs#L50-L58): `write_entity` calls `serde_json::to_vec(value)` and transfers the plain `Vec<u8>` into `write_value`.
- [`crates/storage-sqlite/src/openmls_storage/value_store.rs:173-195,288-299`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/storage-sqlite/src/openmls_storage/value_store.rs#L173-L195): reads return an ordinary database `Vec<u8>` and deserialize from it; [`read_value_on_connection`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/storage-sqlite/src/openmls_storage/value_store.rs#L288-L299) owns the raw blob.
- [`openmls_basic_credential` at MDK's pinned revision](https://github.com/erskingardner/openmls/blob/59e7d3b27a7e95237879dd5478de1fd90eff7ada/basic_credential/src/lib.rs#L22-L46): `SignatureKeyPair.private` is a `SecretVLBytes`, but its Serde helper serializes the secret as ordinary bytes. This proves the MDK JSON value contains the private key rather than only public metadata.

## Acceptance criteria

- [ ] Secret-bearing OpenMLS write buffers use zeroizing ownership and are wiped after SQLite has consumed them, including serialization and database error paths.
- [ ] Secret-bearing OpenMLS read/list/remove buffers are wiped after deserialization, including malformed-data and transaction error paths.
- [ ] The storage boundary explicitly distinguishes secret-bearing labels/entity methods from public-only entities so later additions cannot silently use ordinary buffers.
- [ ] Focused storage tests or type-level guards cover signature keys, epoch/message secrets, HPKE keypairs, and PSKs.
- [ ] Existing SQLCipher and transaction behavior remains unchanged.

## Related

- [#1000](https://github.com/marmot-protocol/mdk/issues/1000) covers a distinct app-layer media-epoch-secret read path.
- [#1018](https://github.com/marmot-protocol/mdk/issues/1018) covers a rusqlite-formatted SQLCipher key buffer, not OpenMLS entity serialization.
- [#451](https://github.com/marmot-protocol/mdk/issues/451) covers account signing-key file/keychain serialization.
- [#673](https://github.com/marmot-protocol/mdk/issues/673) predates the current workspace replacement and does not identify the custom OpenMLS provider's current secret-bearing methods.
- No exact duplicate was found after searching all open/closed MDK issues and pull requests for OpenMLS, signature/HPKE keypair, epoch/message secret, serialization buffer, heap, and zeroization terms.

## Verification scope

Confirmed by static source inspection of MDK and its exact pinned OpenMLS revision. SQLCipher at-rest protection was verified as a separate boundary. Runtime heap-residue instrumentation and Rust tests were not run because this host has no Rust/Cargo toolchain.

agent-p1p:
## Summary

- classify every OpenMLS SQLite value label as secret-bearing or public-only
- serialize and read secret-bearing entity/list buffers through zeroizing ownership, including partial serialization/deserialization and SQLite error paths
- keep raw KeyPackage bundle enumeration zeroizing across the storage/engine boundary
- document the fix in the compatibility-cohort changelog

## Test plan

- `just fast-ci`
- `cargo test -p storage-sqlite` (310 passed)
- `cargo test -p cgka-traits`
- `cargo test -p cgka-engine key_package`
- `cargo clippy -p cgka-traits -p storage-sqlite -p cgka-engine --all-targets --locked -- -D warnings`

## Security scope

Touches OpenMLS key, epoch/message-secret, PSK, pending group-state, and application-export persistence. This PR requires explicit merge approval under the sensitive-path gate.

Fixes #1210

## Summary by CodeRabbit

* **Security**
  * Improved protection of sensitive MLS keys, secrets, private state, and serialized data by securely clearing temporary buffers after use.
  * Key-package and stored secret data now use sensitivity-aware handling to reduce residual information in memory.
  * Sensitive values are redacted from diagnostic output.
* **Bug Fixes**
  * Improved snapshot validation so malformed data fails safely without altering active group state.
  * Preserved compatibility when migrating existing stored data.
* **Documentation**
  * Documented the security improvements in the upcoming release notes.

### DIFF
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -24,6 +24,10 @@ versioning through the workspace version in the root `Cargo.toml`.
   
 ### Fixed
 
+- OpenMLS persistence now zeroizes temporary SQLite serialization buffers for
+  MLS private keys, epoch and message secrets, PSKs, pending group state, and
+  application-export state on success and error paths.
+
 - `MarmotApp` now permits only one live in-memory engine session per account
   across direct clients and managed workers. Concurrent opens return the typed
   `AccountSessionBusy` error, and worker reconnect drops the failed session
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -24,9 +24,10 @@ versioning through the workspace version in the root `Cargo.toml`.
   
 ### Fixed
 
-- OpenMLS persistence now zeroizes temporary SQLite serialization buffers for
-  MLS private keys, epoch and message secrets, PSKs, pending group state, and
-  application-export state on success and error paths.
+- OpenMLS persistence now zeroizes temporary SQLite serialization,
+  deserialization, and rollback-snapshot buffers for MLS private keys, epoch and
+  message secrets, PSKs, pending group state, application-export state, and
+  stored key-package handoffs on success and error paths.
 
 - `MarmotApp` now permits only one live in-memory engine session per account
   across direct clients and managed workers. Concurrent opens return the typed
--- a/crates/storage-sqlite/src/codec.rs
+++ b/crates/storage-sqlite/src/codec.rs
@@ -1,17 +1,180 @@
+use std::fmt;
+use std::io::{self, Write};
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use cgka_traits::message::MessageState;
 use cgka_traits::storage::{StorageError, StorageResult};
 use cgka_traits::types::EpochId;
-use serde::{Serialize, de::DeserializeOwned};
+use serde::de::{DeserializeOwned, SeqAccess, Visitor};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use zeroize::Zeroizing;
 
 /// Total bind-parameter budget for generated statements that use chunked
 /// positive-`IN` lists. This stays below SQLite's historical 999-variable
 /// default and includes any fixed parameters that precede the list.
 pub(crate) const SQLITE_BIND_PARAMETER_CHUNK: usize = 900;
 
+/// Bytes whose current allocation is wiped when ownership ends.
+///
+/// JSON deserialization grows this buffer explicitly so each replaced
+/// allocation is zeroized before deallocation. The storage backend is JSON-only;
+/// deliberately accepting only sequence input avoids a binary deserializer
+/// constructing an intermediate ordinary `Vec<u8>` before this type takes over.
+pub(crate) struct SensitiveBytes(Zeroizing<Vec<u8>>);
+
+impl SensitiveBytes {
+    pub(crate) fn new(bytes: Vec<u8>) -> Self {
+        Self(Zeroizing::new(bytes))
+    }
+
+    fn with_capacity(capacity: usize) -> Self {
+        Self(Zeroizing::new(Vec::with_capacity(capacity)))
+    }
+
+    pub(crate) fn as_slice(&self) -> &[u8] {
+        self.0.as_slice()
+    }
+
+    fn push(&mut self, byte: u8) {
+        if self.0.len() == self.0.capacity() {
+            let next_capacity = if self.0.capacity() == 0 {
+                8
+            } else {
+                self.0
+                    .capacity()
+                    .checked_mul(2)
+                    .expect("sensitive byte buffer capacity overflow")
+            };
+            let mut next = Zeroizing::new(Vec::with_capacity(next_capacity));
+            next.extend_from_slice(self.0.as_slice());
+            let previous = std::mem::replace(&mut self.0, next);
+            drop(previous);
+        }
+        self.0.push(byte);
+    }
+}
+
+impl AsRef<[u8]> for SensitiveBytes {
+    fn as_ref(&self) -> &[u8] {
+        self.as_slice()
+    }
+}
+
+impl Serialize for SensitiveBytes {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        self.as_slice().serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for SensitiveBytes {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        struct SensitiveBytesVisitor;
+
+        impl<'de> Visitor<'de> for SensitiveBytesVisitor {
+            type Value = SensitiveBytes;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                formatter.write_str("a JSON byte array")
+            }
+
+            fn visit_seq<A: SeqAccess<'de>>(self, mut seq: A) -> Result<Self::Value, A::Error> {
+                let mut bytes = SensitiveBytes::with_capacity(seq.size_hint().unwrap_or(0));
+                while let Some(byte) = seq.next_element::<u8>()? {
+                    bytes.push(byte);
+                }
+                Ok(bytes)
+            }
+        }
+
+        deserializer.deserialize_seq(SensitiveBytesVisitor)
+    }
+}
+
+#[derive(Default)]
+struct JsonLength {
+    bytes: usize,
+}
+
+impl Write for JsonLength {
+    fn write(&mut self, bytes: &[u8]) -> io::Result<usize> {
+        self.write_all(bytes)?;
+        Ok(bytes.len())
+    }
+
+    fn write_all(&mut self, bytes: &[u8]) -> io::Result<()> {
+        self.bytes = self
+            .bytes
+            .checked_add(bytes.len())
+            .ok_or_else(|| io::Error::other("sensitive JSON length overflow"))?;
+        Ok(())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
+struct SensitiveJsonWriter {
+    bytes: SensitiveBytes,
+    limit: usize,
+}
+
+impl SensitiveJsonWriter {
+    fn with_capacity(capacity: usize) -> Self {
+        Self {
+            bytes: SensitiveBytes::with_capacity(capacity),
+            limit: capacity,
+        }
+    }
+
+    fn into_bytes(self) -> SensitiveBytes {
+        self.bytes
+    }
+}
+
+impl Write for SensitiveJsonWriter {
+    fn write(&mut self, bytes: &[u8]) -> io::Result<usize> {
+        self.write_all(bytes)?;
+        Ok(bytes.len())
+    }
+
+    fn write_all(&mut self, bytes: &[u8]) -> io::Result<()> {
+        let remaining = self.limit - self.bytes.0.len();
+        if bytes.len() > remaining {
+            return Err(io::Error::other(
+                "sensitive JSON changed size between serialization passes",
+            ));
+        }
+        self.bytes.0.extend_from_slice(bytes);
+        Ok(())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
 pub(crate) fn serialize<T: Serialize>(value: &T) -> StorageResult<Vec<u8>> {
     serde_json::to_vec(value).map_err(|e| StorageError::Serialization(e.to_string()))
+}
+
+pub(crate) fn serialize_sensitive_json<T: Serialize + ?Sized>(
+    value: &T,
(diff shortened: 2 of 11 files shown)


erskingardner:
## Overview

Closes #1210 by making the SQLite OpenMLS provider aware of which stored values carry secrets.

The mechanism is sound and well-chosen:

- `&[u8]` label constants become a nominal `OpenMlsValueLabel` type carrying a `ValueSensitivity` (`Public` | `Secret`), so every label must make an explicit classification choice at definition time.
- `SerializedBuffer` / `SerializedList` / `SecretBytes` route secret-bearing values through `Zeroizing<Vec<u8>>` on serialize, deserialize, list mutation, and DB read.
- `StoredKeyPackageBundle.value` becomes `Zeroizing<Vec<u8>>`, extending the guarantee across the storage→engine boundary.

Two details I checked and liked: `SecretBytes: Serialize` delegates to `self.as_slice().serialize(...)`, which is byte-identical to the old `Vec<u8>` encoding, so **there is no on-disk format change and no migration needed**; and the one production caller that was hand-rolling `serde_json::to_vec` into `write_value` (`store_epoch_key_pairs`, `crates/storage-sqlite/src/openmls_storage/provider.rs:207`) is correctly converted to `write_json`, with `write_value` demoted to `#[cfg(test)]` so the raw-bytes escape hatch no longer exists in prod. The `Zeroizing` usage also matches existing convention in `marmot-app` (`sqlcipher.rs`, `media/group_image.rs`).

## Issues

### 1. `Vec` reallocation defeats zeroization for most secret buffers (main finding)

In `crates/storage-sqlite/src/openmls_storage/value_store.rs`, the secret serialize path is:

```rust
let mut bytes = Zeroizing::new(Vec::new());
serde_json::to_writer(&mut *bytes, value)?;
```

`Zeroizing` wipes the *final* allocation on drop. `Vec::new()` + `io::Write` grows by doubling, so serializing a ~4 KB `KeyPackageBundle` leaves roughly a dozen freed intermediate allocations, each holding a prefix of the plaintext secret, none of them zeroized. The same applies to the deserialize visitor — `seq.size_hint()` is `None` for serde_json arrays, so `Vec::with_capacity(0)` grows the same way byte-by-byte.

The net effect is that the PR reduces but does not eliminate residue on exactly the values it targets. Fixes, in increasing order of rigor: seed with a generous `Vec::with_capacity`; re-serialize in two passes (size, then exact capacity); or a small `Write` adapter that zeroizes the old buffer on each grow. At minimum the code comment ("the guard wipes that partial buffer") should be qualified — as written it claims more than it delivers.

### 2. Epoch-rollback snapshots bypass the whole mechanism

`crates/storage-sqlite/src/storage/snapshots/capture.rs:207` selects `value` for *every* `openmls_values` row of a group into `OpenMlsValueSnapshot.value: Vec<u8>` (`snapshots/rows.rs`), then serde-serializes the whole `Snapshot` into another plain buffer. That set includes `EpochSecrets`, `MessageSecrets`, `GroupState`, and `EncryptionKeyPair` — the identical secrets this PR is hardening, on a path it doesn't touch. Rollback runs on real error paths, not just tests. Either extend the fix here or state the gap explicitly in the PR / a follow-up issue; leaving it silent makes the changelog entry read as more complete than it is.

### 3. Label classification test is tautological and not exhaustive

The two tests in `openmls_storage/labels.rs` restate the constant definitions in a second hand-maintained list. A newly added label appears in *neither* list and nothing fails — and `OpenMlsValueLabel::public()` is a one-word escape hatch, so the type only forces a decision, not a correct one. Cheap fix that makes the test meaningful: a `const ALL_LABELS: &[OpenMlsValueLabel]` and an assertion that the two test lists partition it.

Separately, no test asserts that zeroization actually happens — the tests verify *routing*, not *wiping*. That's a fair limitation (it's hard to test), but worth being explicit about rather than implied by the test names.

### 4. `unreachable!()` in a storage write path

`SerializedList::push` panics if variants mismatch. It's currently unreachable — list and value derive from the same label in `append_entity_on_connection` — but it's a panic in persistence code guarded only by a convention. Having `push` take the raw value and construct the variant from `self` would make it structurally impossible.

## Smaller notes

- `SecretBytes: Deserialize` implements only `visit_seq`. Correct for JSON, but it would fail at runtime against any binary format (bincode/postcard emit `visit_bytes`). Either add `visit_bytes`/`visit_byte_buf` or note the JSON-only assumption.
- `StoredKeyPackageBundle` still derives `Debug`, and `Zeroizing` derives `Debug` too — so `{:?}` prints raw private init/leaf key bytes. Not a regression, but the PR adds a doc comment specifically calling out that the field holds private keys while leaving the leak in place; given the repo's privacy-safe logging invariant, a redacting manual `Debug` fits here.
- `SerializedList::remove_matching` has two identical arms; the `Public`/`Secret` split buys nothing there.
- Classification itself looks right to me. `Tree`/`GroupContext`/`OwnLeafNodes`/`QueuedProposal` as public is correct (OpenMLS keeps private HPKE material under `EncryptionKeyPair`), and `ApplicationExportTree` + `GroupState` as secret are the two non-obvious calls, both correctly justified in comments.

## Assessment

Good, well-structured security fix with a real design idea behind it — the nominal label type is the right lever, and format compatibility was handled carefully. I'd want (1) addressed before merge, since it's the difference between "buffers are zeroized" and "the last of several copies is zeroized," and (2) acknowledged one way or the other. (3) and (4) are cheap and would meaningfully harden the guarantee against future edits.

The sensitive-path gate the PR description invokes is appropriate here.

agent-p1p:
## Pip adversarial review — REQUEST_CHANGES

Reviewed exact head `6cd5463d1b2527ecd96264aa62d08c2e575bc98f`. The typed label boundary, production `write_json` routing, and zeroizing `StoredKeyPackageBundle` handoff are sound. Release hygiene is clean: the Unreleased changelog entry is present and there are no version, lockfile, fixture, or migration changes.

This independently confirms the two material gaps already identified in [the maintainer review](https://github.com/marmot-protocol/mdk/pull/1214#issuecomment-5177752577) and assigns both blocking severity for the Pip merge gate.

### BLOCKING

- **Growing `Vec`s retain earlier secret copies** — `crates/storage-sqlite/src/openmls_storage/value_store.rs:40-45,84-86`: the write path starts from `Vec::new()`, and the JSON byte-array visitor receives no `SeqAccess::size_hint()` from serde_json, so both buffers reallocate while growing. `Zeroizing<Vec<_>>` wipes only the current allocation; zeroize itself documents that it “Cannot ensure that previous reallocations did not leave values on the heap.” Freed prefix allocations can therefore retain the same MLS material issue #1210 targets. Use a design that does not reallocate after secret bytes are written, or explicitly zeroizes the old allocation on every grow, on success and error paths.
- **Production snapshot capture/restore bypasses sensitivity-aware storage** — `crates/storage-sqlite/src/storage/snapshots/capture.rs:41-45,207-230`, `rows.rs:57-61`, `restore.rs:54-65`: every group-scoped `openmls_values.value`, including epoch/message secrets, pending group state, and encryption keypairs, is read into ordinary `Vec<u8>` fields, copied into an ordinary serialized snapshot blob, and reconstructed into ordinary vectors during rollback. These are real fork-recovery and pre-commit rollback paths. The acceptance criteria and changelog claim temporary serialization/deserialization coverage for this material, so this path must also use zeroizing ownership (including malformed data and SQLite/transaction errors), or the PR scope and changelog must be narrowed with the uncovered path tracked separately.

### SUGGESTION

- `crates/storage-sqlite/src/openmls_storage/labels.rs:145-183`: the two classification tests are hand-maintained lists with no exhaustive `ALL_LABELS` set. A future constant can be omitted from both tests without failure. Make the tests partition one exhaustive label registry so additions cannot evade the guard.

### NITPICK

none.

### Verification

- `cargo test -p storage-sqlite` — 311 passed
- `cargo clippy -p storage-sqlite --all-targets --locked -- -D warnings` — passed

### SUMMARY

The approach is directionally correct and preserves the on-disk format, but the current implementation still leaves realloc-freed secret copies and an entire production snapshot path in ordinary buffers. The central security acceptance criteria are therefore not met at this head.

**VERDICT: REQUEST_CHANGES**

**SENSITIVE_PATHS:** `crates/storage-sqlite/src/openmls_storage.rs`, `crates/storage-sqlite/src/openmls_storage/{labels.rs,provider.rs,value_store.rs}`, `crates/traits/src/storage.rs`; additionally implicated: `crates/storage-sqlite/src/storage/snapshots/{capture.rs,rows.rs,restore.rs}`.

agent-p1p:
@erskingardner Addressed the review findings in cb5593c0.

- Sensitive JSON writes now use a counting pass followed by a fixed-limit zeroizing allocation. The deserializer explicitly copies into a new zeroizing allocation and wipes the prior allocation before every growth.
- Rollback snapshot capture/restore now keeps both the serialized snapshot blob and every OpenMLS row value in zeroizing ownership, including malformed-input and transaction-error paths.
- OpenMLS labels are generated from one exhaustive registry; production label constructors are private, so new labels must enter that registry.
- List append derives serialization ownership from the existing list variant, removing the mismatch panic.
- `StoredKeyPackageBundle` now has a redacting manual `Debug` implementation.

Verification:
- `cargo test -p storage-sqlite --locked` — 316 passed
- `cargo test -p cgka-traits --locked` — passed (unit, integration, and doc tests)
- `just fast-ci` — passed (fmt, naming/ledger gates, workspace checks, OTLP checks, workspace clippy, convergence policy pin)

agent-p1p:
@coderabbitai review

agent-p1p:
Verification update:

- The GitHub Actions rerun is green, including Relay runtime tests.
- The original relay-runtime failure passes locally in isolation (1/1); no source change was needed.
- The CodeRabbit manual retrigger finished with no inline findings.
- The branch is clean, current with master, and ready for merge-gate review.

This PR changes MLS/key-material storage paths and requires explicit human approval.

agent-p1p:
@coderabbitai review

PR #1215, opened 2026-07-31
fix: zeroize CLI nsec imports

### LINKED ISSUE #1209: CLI/daemon nsec import path retains plaintext String, Debug, and JSON copies
## Current behavior

An `nsec` supplied through `wn login --nsec-stdin` or `wn account create --nsec-stdin` is materialized into the generic `Cli` command tree as a normal `String`. If `wnd` is available, that secret-bearing `Cli` is cloned, serialized into a normal JSON `Vec<u8>`, sent over the local daemon socket, buffered again as a normal `Vec<u8>`, and deserialized into another secret-bearing command tree. The app runtime then accepts the same value through a clonable, raw-`Debug` `AccountSetupRequest` and clones the identity once more before import.

No current production tracing call was found rendering these objects, and the daemon socket is local. This is not a demonstrated network or active log leak. It does leave the full account signing key in multiple non-zeroizing heap buffers and makes accidental disclosure through `Debug` structurally possible. A same-user memory disclosure, crash dump, swap capture, or local IPC observation can recover the imported key after the intended secret owner is gone.

1. Start `wnd` for a White Noise home so normal commands route through the daemon socket.
2. Pipe an `nsec` to `wn login --nsec-stdin` or `wn account create --nsec-stdin`.
3. The client reads and trims the secret into ordinary strings, clones the resulting `Cli`, and JSON-encodes it into an ordinary byte vector.
4. `wnd` reads the JSON into another ordinary byte vector, deserializes it, and passes the identity through `AccountSetupRequest` and another clone.

## Expected behavior

Imported account secrets should remain in dedicated zeroizing/redacted types from the stdin boundary through daemon decoding and account import. Generic command/request `Debug`, `Clone`, and JSON buffers should not retain plaintext `nsec` material after use.

## Code evidence

- [`crates/cli/src/args.rs:9-16`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/cli/src/args.rs#L9-L16): `Cli` derives `Clone`, raw `Debug`, `Serialize`, and `Deserialize`; [`Login.identity`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/cli/src/args.rs#L116-L125) is an `Option<String>` that receives stdin nsecs.
- [`crates/cli/src/lib.rs:151-153`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/cli/src/lib.rs#L151-L153): secret materialization runs before [`cli.clone()` daemon routing](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/cli/src/lib.rs#L204-L208); [`read_nsec_from_stdin`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/cli/src/lib.rs#L274-L284) allocates a `String` and a second `trim().to_owned()` copy.
- [`crates/cli/src/daemon/protocol.rs:104-115`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/cli/src/daemon/protocol.rs#L104-L115): `DaemonRequest` derives raw `Debug`/`Clone`/Serde and [`execute`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/cli/src/daemon/protocol.rs#L267-L269) wraps the whole `Cli` in `Execute`.
- [`crates/cli/src/daemon/protocol.rs:361-388,457-490`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/cli/src/daemon/protocol.rs#L361-L388): both daemon encode and decode paths use ordinary `Vec<u8>` request buffers; [`encode_daemon_request`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/cli/src/daemon/protocol.rs#L457-L490) serializes the plaintext secret into JSON.
- [`crates/marmot-app/src/runtime/mod.rs:699-707,3691-3704`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/marmot-app/src/runtime/mod.rs#L699-L707): `AccountSetupRequest` derives `Clone` and raw `Debug` over `identity: Option<String>`; [`create_or_import_account`](https://github.com/marmot-protocol/mdk/blob/1c51295d41d0c37465a2c013fa4c58365c2f2ebb/crates/marmot-app/src/runtime/mod.rs#L3691-L3704) clones the identity before import.

## Acceptance criteria

- [ ] Stdin-provided nsecs enter a dedicated zeroizing, redacted secret type at the input boundary.
- [ ] Import requests cannot raw-`Debug` account secrets and avoid secret-bearing `Clone` where ownership can move.
- [ ] Client and daemon framing/deserialization buffers that can contain nsecs are wiped on success and every error path.
- [ ] The daemon route does not embed an nsec in the generic long-lived `Cli` representation, or an equivalent design proves and tests bounded zeroizing ownership end to end.
- [ ] Focused CLI/daemon tests cover both successful import and encode/decode/import failure paths without exposing the nsec in formatted diagnostics.

## Related

- [#680](https://github.com/marmot-protocol/mdk/issues/680) removed argv/process-list exposure. This issue concerns the replacement stdin path after materialization, including daemon JSON and runtime request copies.
- [#451](https://github.com/marmot-protocol/mdk/issues/451) fixed account-store serialization buffers, not the CLI/daemon/setup path.
- [#673](https://github.com/marmot-protocol/mdk/issues/673) was a broad pre-workspace-replacement zeroization issue and does not cover these current call sites.
- [#1026](https://github.com/marmot-protocol/mdk/issues/1026) concerns raw hex secret detection on argv, not stdin nsec lifetime.
- No exact duplicate was found after searching all open/closed MDK issues and pull requests for nsec, stdin, daemon request, zeroization, IPC, and account setup terms.

## Verification scope

Confirmed from the current source and history. No active log statement or Internet transmission was found. Runtime heap-residue instrumentation was not run because this host has no Rust/Cargo toolchain.

agent-p1p:
Fixes #1209

## Summary
- keeps `--nsec-stdin` material out of the clonable/debuggable `Cli` command tree by carrying it in a dedicated zeroizing sidecar
- zeroizes final owned nsec-bearing daemon request buffers under our control and redacts secret-bearing `Debug` output
- separates private-key import material from public identity in `AccountSetupRequest` and removes secret-bearing `Clone`
- preserves sidecar ownership through daemon connection and relayless local-fallback paths

## Review follow-up
- centralizes the conservative, ASCII-case-insensitive `nsec` prefix classifier in `marmot-app`; CLI argv/stdin handling, TUI routing/redaction, runtime validation, and UniFFI all use that source
- deletes the duplicated owned daemon transport while preserving recovery of the original `Cli` and zeroizing nsec sidecar after connect/write/read/decode failures
- renames the non-secret retry helper to `AccountSetupRequest::relay_options_only`; it intentionally copies only relay options so secret-bearing requests remain non-`Clone`
- keeps `MarmotAppRuntime::login` as public Rust API for downstream callers even though this workspace currently has no call site; private-key input and identity/import mismatches now have distinct typed errors
- pins disabled app-runtime fall-through before relay validation with malformed daemon defaults and proves the nsec sidecar is retained for local execution
- documents that allocator reallocations and Tokio `BufReader` read-ahead can leave transient copies outside the owned `Zeroizing` buffers

## Verification
- `just fast-ci`
- `cargo test -p wn-cli --lib --locked`
- `cargo test -p marmot-app --lib --locked`
- `cargo test -p marmot-uniffi --locked`
- `cargo test -p agent-connector --locked`
- `cargo fmt --all --check`
- `RUSTFLAGS='-D warnings' cargo check --workspace --all-targets --all-features --locked`
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- `cargo nextest run -p wn-cli --test cli --features test-policy-overrides --locked --profile ci` (88 passed)
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked -- --test-threads=1` (84 passed)
- `cargo test -p marmot-uniffi --locked` (123 passed)
- `cargo test -p agent-connector --locked` (133 passed)
- focused secret redaction, daemon sidecar roundtrip/fallback, stdin import, and runtime rollback tests passed

## Sensitive paths
Touches CLI nsec intake, daemon request framing, and MarmotApp private-key import ownership. No cryptographic algorithm, MLS, CGKA, or trust-anchor changes.

## Release hygiene
- changelog: `crates/cli/CHANGELOG.md` under `Unreleased`
- version bump: none

## Summary by CodeRabbit

* **Security**
  * Improved protection for secrets entered through `wn login --nsec-stdin` and `wn account create --nsec-stdin`.
  * Sensitive values are redacted from diagnostics and securely cleared after use.

* **Account Management**
  * Login now supports importing an account directly from a private key.
  * Public identities and private-key imports receive clearer validation.

* **Reliability**
  * Improved daemon-to-local fallback when remote account setup cannot be completed.
  * Account setup retries no longer reuse sensitive request data.

### DIFF
--- a/crates/cli/src/daemon/protocol.rs
+++ b/crates/cli/src/daemon/protocol.rs
@@ -229,63 +229,18 @@ pub(crate) async fn send_execute(
         cli: Box::new(cli),
         import_nsec,
     };
-    match send_owned_daemon_request(socket, request).await {
+    match send_request(socket, &request).await {
         Ok(output) => Ok(output),
-        Err((err, request)) => {
-            let (cli, import_nsec) = take_execute_request(request);
+        Err(err) => {
+            let DaemonRequest::Execute { cli, import_nsec } = request else {
+                unreachable!("send_execute only constructs Execute requests")
+            };
             Err(RecoverableDaemonExecute {
                 err,
-                cli,
+                cli: *cli,
                 import_nsec,
             })
         }
-    }
-}
-
-fn take_execute_request(request: DaemonRequest) -> (Cli, Option<crate::secret::ImportNsec>) {
-    match request {
-        DaemonRequest::Execute { cli, import_nsec } => (*cli, import_nsec),
-        _ => panic!("expected Execute daemon request"),
-    }
-}
-
-async fn send_owned_daemon_request(
-    socket: &Path,
-    request: DaemonRequest,
-) -> Result<CliOutput, (DaemonClientError, DaemonRequest)> {
-    let bytes = match encode_daemon_request(&request) {
-        Ok(bytes) => bytes,
-        Err(err) => return Err((err, request)),
-    };
-    let mut stream = match UnixStream::connect(socket).await {
-        Ok(stream) => stream,
-        Err(source) => {
-            return Err((
-                DaemonClientError::Connect {
-                    socket: socket.to_owned(),
-                    source,
-                },
-                request,
-            ));
-        }
-    };
-    if let Err(source) = stream.write_all(&bytes).await {
-        return Err((DaemonClientError::Io(source), request));
-    }
-    if let Err(source) = stream.shutdown().await {
-        return Err((DaemonClientError::Io(source), request));
-    }
-
-    let mut response = Zeroizing::new(Vec::new());
-    if let Err(source) = stream.read_to_end(&mut response).await {
-        return Err((DaemonClientError::Io(source), request));
-    }
-    if response.is_empty() {
-        return Err((DaemonClientError::EmptyResponse, request));
-    }
-    match decode_daemon_output(&response) {
-        Ok(output) => Ok(output),
-        Err(err) => Err((err, request)),
     }
 }
 
@@ -452,6 +407,9 @@ pub(crate) async fn read_daemon_request(
     // never sends a newline cannot make us buffer unbounded memory before the
     // size check runs (read_until on a Take adapter stops silently at the limit
     // instead of erroring). Mirrors agent-control's `read_frame`.
+    //
+    // `BufReader` may retain read-ahead bytes outside this `Zeroizing` buffer;
+    // only the assembled frame payload is wiped on drop.
     let limit = (MAX_DAEMON_REQUEST_BYTES + 1) as u64;
     let read = {
         let mut reader = BufReader::new(&mut *stream).take(limit);
@@ -544,6 +502,10 @@ pub(crate) async fn write_stream_end(stream: &mut UnixStream) -> bool {
 /// `read_daemon_request` / `MAX_DAEMON_REQUEST_BYTES`); checking client-side
 /// turns an oversized request (e.g. `messages send` with a huge body) into a
 /// clear local error instead of a connection the daemon must reject.
+///
+/// The returned buffer is zeroized on drop. `serde_json::to_vec` and the
+/// trailing `push` may leave transient copies in freed heap memory; this does
+/// not wipe every intermediate allocation.
 pub(crate) fn encode_daemon_request(
     request: &DaemonRequest,
 ) -> Result<Zeroizing<Vec<u8>>, DaemonClientError> {
--- a/crates/cli/src/daemon/protocol.rs
+++ b/crates/cli/src/daemon/protocol.rs
@@ -1,6 +1,7 @@
 //! Daemon wire protocol: request/response types, framing, and the client.
 
 use super::*;
+use zeroize::Zeroizing;
 
 pub(crate) const DAEMON_SERVER_BUSY_CODE: &str = "server_busy";
 pub(crate) const DAEMON_SERVER_BUSY_MESSAGE: &str = "daemon connection capacity is busy";
@@ -68,6 +69,14 @@ pub enum DaemonClientError {
     ServerBusy,
 }
 
+/// Failed daemon execute with the original `Cli` and stdin `nsec` sidecar when
+/// the client can safely retry locally (for example connect refused).
+pub(crate) struct RecoverableDaemonExecute {
+    pub err: DaemonClientError,
+    pub cli: Cli,
+    pub import_nsec: Option<crate::secret::ImportNsec>,
+}
+
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
 pub struct DaemonRuntimeActivityReport {
     pub started_at: u64,
@@ -101,17 +110,31 @@ pub type DaemonStreamWatchReport = marmot_app::AgentStreamWatchReport;
 
 pub type DaemonOutgoingStreamReport = StreamComposeReport;
 
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize)]
 pub(crate) enum DaemonRequest {
     Ping,
     Status,
     Shutdown,
-    StreamWatch { cli: Box<Cli> },
-    MessagesSubscribe { cli: Box<Cli> },
-    ChatsSubscribe { cli: Box<Cli> },
-    GroupStateSubscribe { cli: Box<Cli> },
-    NotificationsSubscribe { cli: Box<Cli> },
-    Execute { cli: Box<Cli> },
+    StreamWatch {
+        cli: Box<Cli>,
+    },
+    MessagesSubscribe {
+        cli: Box<Cli>,
+    },
+    ChatsSubscribe {
+        cli: Box<Cli>,
+    },
+    GroupStateSubscribe {
+        cli: Box<Cli>,
+    },
+    NotificationsSubscribe {
+        cli: Box<Cli>,
+    },
+    Execute {
+        cli: Box<Cli>,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        import_nsec: Option<crate::secret::ImportNsec>,
+    },
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
@@ -197,8 +220,73 @@ impl DaemonEventHub {
     }
 }
 
-pub(crate) async fn send_execute(socket: &Path, cli: Cli) -> Result<CliOutput, DaemonClientError> {
-    DaemonClient::new(socket).execute(cli).await
+pub(crate) async fn send_execute(
+    socket: &Path,
+    cli: Cli,
+    import_nsec: Option<crate::secret::ImportNsec>,
+) -> Result<CliOutput, RecoverableDaemonExecute> {
+    let request = DaemonRequest::Execute {
+        cli: Box::new(cli),
+        import_nsec,
+    };
+    match send_owned_daemon_request(socket, request).await {
+        Ok(output) => Ok(output),
+        Err((err, request)) => {
+            let (cli, import_nsec) = take_execute_request(request);
+            Err(RecoverableDaemonExecute {
+                err,
+                cli,
+                import_nsec,
+            })
+        }
+    }
+}
+
+fn take_execute_request(request: DaemonRequest) -> (Cli, Option<crate::secret::ImportNsec>) {
+    match request {
+        DaemonRequest::Execute { cli, import_nsec } => (*cli, import_nsec),
+        _ => panic!("expected Execute daemon request"),
+    }
+}
+
+async fn send_owned_daemon_request(
+    socket: &Path,
+    request: DaemonRequest,
+) -> Result<CliOutput, (DaemonClientError, DaemonRequest)> {
+    let bytes = match encode_daemon_request(&request) {
+        Ok(bytes) => bytes,
+        Err(err) => return Err((err, request)),
+    };
+    let mut stream = match UnixStream::connect(socket).await {
+        Ok(stream) => stream,
+        Err(source) => {
+            return Err((
+                DaemonClientError::Connect {
+                    socket: socket.to_owned(),
+                    source,
+                },
+                request,
+            ));
+        }
+    };
+    if let Err(source) = stream.write_all(&bytes).await {
+        return Err((DaemonClientError::Io(source), request));
+    }
+    if let Err(source) = stream.shutdown().await {
+        return Err((DaemonClientError::Io(source), request));
+    }
+
+    let mut response = Zeroizing::new(Vec::new());
+    if let Err(source) = stream.read_to_end(&mut response).await {
+        return Err((DaemonClientError::Io(source), request));
+    }
+    if response.is_empty() {
+        return Err((DaemonClientError::EmptyResponse, request));
+    }
+    match decode_daemon_output(&response) {
+        Ok(output) => Ok(output),
+        Err(err) => Err((err, request)),
+    }
 }
 
 pub(crate) async fn send_stream_watch(
@@ -262,10 +350,6 @@ impl DaemonClient {
 
     pub async fn shutdown(&self) -> Result<CliOutput, DaemonClientError> {
         send_request(&self.socket, &DaemonRequest::Shutdown).await
-    }
-
-    pub(crate) async fn execute(&self, cli: Cli) -> Result<CliOutput, DaemonClientError> {
-        send_request(&self.socket, &DaemonRequest::Execute { cli: Box::new(cli) }).await
     }
 
     pub(crate) async fn stream_watch(&self, cli: Cli) -> Result<CliOutput, DaemonClientError> {
@@ -361,7 +445,7 @@ where
 pub(crate) async fn read_daemon_request(
     stream: &mut UnixStream,
 ) -> Result<DaemonRequest, Box<dyn std::error::Error + Send + Sync>> {
-    let mut request = Vec::new();
+    let mut request = Zeroizing::new(Vec::new());
     // Buffer the raw stream so a near-1-MiB frame (e.g. an `Execute` request
     // carrying the whole `Cli`) costs a handful of `read()` syscalls instead of
     // one per byte. Cap the read at one byte past the limit so a client that
@@ -460,8 +544,10 @@ pub(crate) async fn write_stream_end(stream: &mut UnixStream) -> bool {
 /// `read_daemon_request` / `MAX_DAEMON_REQUEST_BYTES`); checking client-side
 /// turns an oversized request (e.g. `messages send` with a huge body) into a
 /// clear local error instead of a connection the daemon must reject.
-pub(crate) fn encode_daemon_request(request: &DaemonRequest) -> Result<Vec<u8>, DaemonClientError> {
-    let mut bytes = serde_json::to_vec(request)?;
+pub(crate) fn encode_daemon_request(
+    request: &DaemonRequest,
+) -> Result<Zeroizing<Vec<u8>>, DaemonClientError> {
+    let mut bytes = Zeroizing::new(serde_json::to_vec(request)?);
     // The daemon reads up to and excluding the trailing newline, so compare the
     // JSON payload length (without the framing newline) against the limit.
     if bytes.len() > MAX_DAEMON_REQUEST_BYTES {
--- a/crates/cli/src/daemon/runtime_host.rs
+++ b/crates/cli/src/daemon/runtime_host.rs
@@ -334,6 +334,9 @@ pub(crate) fn app_runtime_account_setup_request(
 ) -> Result<Option<marmot_app::AccountSetupRequest>, crate::WnError> {
     match &cli.command {
         crate::Command::CreateIdentity => {
+            if import_nsec.is_some() {
+                return Err(crate::WnError::InvalidPublicKey);
+            }
             if cli.daemon_default_account_relays.is_empty() {
                 return Err(crate::WnError::MissingRelay);
             }
--- a/crates/cli/src/daemon/runtime_host.rs
+++ b/crates/cli/src/daemon/runtime_host.rs
@@ -1,6 +1,7 @@
 //! App-runtime hosting glue: reconciliation, event bridge, and hosted command dispatch.
 
 use super::*;
+use crate::ImportNsec;
 
 #[derive(Debug)]
 pub(crate) struct DaemonState {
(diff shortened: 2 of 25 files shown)


agent-p1p:
## Pip adversarial review

**Verdict: changes required** — 2 blocking findings, 1 suggestion. Reviewed exact head `5cb9e373547fcccde96176e09851d5f75768612c`.

### Blocking

1. **`crates/marmot-app/src/runtime/mod.rs:3151-3169, 3739-3759` — the split public/private identity invariant is not enforced before mutation.**

   `AccountManager::create_or_import_account` silently prefers `import_nsec` over `identity`. `MarmotAppRuntime::login` overwrites only `identity`, so a request carrying secret A plus public identity B logs in/imports A and discards B. `create_identity` has the symmetric defect: it clears only `identity`, so a request carrying `import_nsec` imports that key instead of creating a new identity. This can create/reactivate the wrong account before the caller learns anything is wrong. It independently confirms the unresolved CodeRabbit mismatch finding.

   Validate the request shape before account lookup or mutation. Reject conflicting fields (or verify they resolve to the same account), and make `create_identity`/`login` explicitly clear or reject fields incompatible with their API contract. Add regressions for mismatched public+secret input and `create_identity` receiving an import sidecar.

2. **`crates/marmot-app/src/runtime/mod.rs:699-735, 3199-3203` — `AccountSetupRequest` can still raw-`Debug` an nsec through `identity`.**

   The new custom formatter redacts only `import_nsec`; it prints `identity` verbatim. Because `identity` remains a public `Option<String>` and direct `create_or_import_account` does not reject secret-looking values up front, existing or malformed callers can still construct `AccountSetupRequest { identity: Some(nsec), .. }` and expose the full key via `Debug`. The field comment does not enforce the invariant, so the acceptance criterion that import requests cannot raw-debug account secrets is not met.

   Enforce public identity structurally if practical, and defensively redact/reject nsec-shaped `identity` values before any use. Add a regression formatting a legacy/malformed request with the nsec in `identity`.

### Suggestion

- **`crates/cli/src/daemon/tests.rs:979-1006`** — the daemon round-trip test checks only `import_nsec.is_some()`. Compare the decoded value with the original secret so the wire-format regression guard catches corruption or replacement. This matches the existing CodeRabbit suggestion.

### Existing review triage

- The CodeRabbit claim that `ImportNsec::serialize` requires zeroize's `serde` feature is not reproduced. The implementation delegates through `Zeroizing<String>`'s deref to `String::serialize`; `cargo clippy` compiled the path with zeroize features `alloc,default,zeroize_derive` and no `serde` feature.
- The panic/helper and deduplication comments are non-blocking cleanup. Avoid expanding this security fix unless the resulting code is smaller.

### Verification

- `cargo test -p wn-cli --lib import_nsec --locked` — 3 passed
- `cargo test -p marmot-app account_setup_request_debug_redacts_import_nsec --locked` — 1 passed
- Three focused CLI stdin/import/rollback integration tests — 3 passed
- `cargo clippy -p wn-cli -p marmot-app -p marmot-uniffi --all-targets --locked -- -D warnings` — passed
- GitHub checks are green at the reviewed head.

No out-of-scope follow-up issue filed; both blockers are within #1209.

agent-p1p:
Addressed the blocking review findings in signed commit 36a5cb91.

- Account setup now rejects nsec-shaped public identities and mismatched public/private pairs before account lookup or mutation.
- `create_identity`, `login`, and malformed daemon `create-identity` requests reject import sidecars instead of silently acting on them.
- `AccountSetupRequest` Debug redacts lower- and uppercase nsec-shaped identity values.
- The daemon round-trip test now verifies the decoded secret value, not just presence.

Verification: `just fast-ci`; marmot-app unit tests (544/544); wn-cli unit tests (534/534); CI-exact serial relay-runtime tests (84/84); CLI E2E (88/88). Fresh independent Composer review found no blocking security or logic errors.

crates/cli/src/daemon/tests.rs
```diff
@@ -932,6 +976,38 @@ async fn daemon_request_reader_within_returns_request_before_timeout() {
...
+    let payload = bytes.strip_suffix(b"\n").unwrap_or(&bytes);
+    let decoded: DaemonRequest = serde_json::from_slice(payload).expect("decode execute request");
+    match decoded {
+        DaemonRequest::Execute { cli, import_nsec } => {
+            assert!(import_nsec.is_some());
+            match cli.command {
+                crate::Command::Login { identity: None, .. } => {}
+                other => panic!("expected login without identity, got {other:?}"),
+            }
+        }
```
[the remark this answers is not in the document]
agent-p1p: Fixed in 36a5cb91. The round-trip test now consumes the decoded sidecar and compares its exact nsec value with the original fixture.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3693,15 +3741,21 @@ impl AccountManager {
...
+            match self.signed_out_account_for_identity(identity)? {
                 Some(account) => (account, None),
-                None => self.create_nostr_account(request.identity.clone())?,
-            },
-            None => self.create_nostr_account(None)?,
+                None => self.create_nostr_account_from_setup(&request)?,
+            }
+        } else {
+            self.create_nostr_account_from_setup(&request)?
         };
```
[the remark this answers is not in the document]
agent-p1p: Fixed in 36a5cb91. A centralized pre-mutation validator now derives both account IDs and rejects mismatched public identity/import-secret pairs before lookup, creation, import, or reactivation. Matching pairs remain accepted, with regression coverage.

crates/cli/src/secret.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl Serialize for ImportNsec {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.0.serialize(serializer)
+    }
+}
```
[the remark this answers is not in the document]
agent-p1p: Verified as a false positive. `ImportNsec::serialize` resolves through `Zeroizing<String>` deref to `String::serialize`; the CLI crate compiles this path without zeroize serde support. `just fast-ci`, workspace checks, and clippy all pass at 36a5cb91, so no dependency feature change is needed.

agent-p1p:
CI note: I cancelled run 30854987005 while its final Simulator and vectors job was still executing. The current branch is based exactly on master 4feec6ad, whose own CI deterministically fails `production_library_sources_do_not_write_direct_output` on the unrelated `integrations/terminal-harness/src/test_support.rs:338` logger: https://github.com/marmot-protocol/mdk/actions/runs/30850993545/job/91810765227.

All other checks on this PR passed, and the cancelled simulator job had not reported a test failure. I am keeping the PR draft until the base regression is fixed, then will rebase and run a fresh full CI matrix. The #1209 changes are not being expanded to carry the unrelated master hotfix.

crates/cli/src/daemon/protocol.rs
```diff
@@ -197,8 +220,73 @@ impl DaemonEventHub {
...
+}
+
+fn take_execute_request(request: DaemonRequest) -> (Cli, Option<crate::secret::ImportNsec>) {
+    match request {
+        DaemonRequest::Execute { cli, import_nsec } => (*cli, import_nsec),
+        _ => panic!("expected Execute daemon request"),
+    }
+}
+
+async fn send_owned_daemon_request(
```
erskingardner: `send_owned_daemon_request` is a near-verbatim copy of `send_request` (connect / write_all / shutdown / read_to_end / decode), differing only in returning the request back on error, and it drags in `take_execute_request` with a `panic!` for a case the type system already makes unreachable from the single call site.

`send_request` takes `&DaemonRequest`, so the caller never loses ownership — the recovery can be done without a second copy of the transport code:

```rust
pub(crate) async fn send_execute(
    socket: &Path,
    cli: Cli,
    import_nsec: Option<crate::secret::ImportNsec>,
) -> Result<CliOutput, RecoverableDaemonExecute> {
    let request = DaemonRequest::Execute { cli: Box::new(cli), import_nsec };
    match send_request(socket, &request).await {
        Ok(output) => Ok(output),
        Err(err) => {
            let DaemonRequest::Execute { cli, import_nsec } = request else { unreachable!() };
            Err(RecoverableDaemonExecute { err, cli: *cli, import_nsec })
        }
    }
}
```

That deletes ~45 lines and the panicking helper. The one thing lost is the `Zeroizing` response buffer at line 279 — but the daemon *response* never carries the nsec (only the request does), so if that wrapper is load-bearing it deserves a comment explaining why; otherwise a one-line change in `send_request` covers every caller rather than just this one.

crates/cli/src/daemon/protocol.rs
```diff
@@ -361,7 +445,7 @@ where
 pub(crate) async fn read_daemon_request(
     stream: &mut UnixStream,
 ) -> Result<DaemonRequest, Box<dyn std::error::Error + Send + Sync>> {
-    let mut request = Vec::new();
+    let mut request = Zeroizing::new(Vec::new());
```
erskingardner, line 403: `request` is `Zeroizing`, but the `BufReader` wrapped around the stream is not: it holds its own (8 KiB default) copy of the frame bytes, including the nsec, and is dropped unzeroed at the end of the block. Same class of residue on the encode side — `serde_json::to_vec` grows by reallocation, and `bytes.push(b'\n')` at line 559 can realloc again, each time freeing an unzeroed buffer containing the secret.

I don't think either needs to be fixed here (fully controlling serde's allocation is more work than this issue warrants), but the CHANGELOG currently promises the framing buffers are wiped "on every path," which reads stronger than what ships. Either soften that wording or add a short comment here recording the known residue, so a future reader doesn't assume the property holds.

crates/marmot-uniffi/src/commands/account.rs
```diff
@@ -133,15 +135,21 @@ impl Marmot {
         default_relays: Vec<String>,
         bootstrap_relays: Vec<String>,
     ) -> Result<AccountSummaryFfi, MarmotKitError> {
+        let (public_identity, import_nsec) = if identity.starts_with("nsec") {
```
erskingardner: Third copy of the nsec-prefix test, and also case-sensitive. Worth exporting the hardened `marmot_app::is_nostr_secret` (or a small `AccountSetupRequest::from_identity_string` helper) rather than re-deriving the classification at each boundary — that's what let the CLI copy drift in the first place.

Separately, this switches `login` from `self.runtime.login(identity, request)` to `create_or_import_account`. That looks correct and necessary (the smoke test at `crates/marmot-uniffi/tests/smoke.rs:270` passes an nsec, which `runtime.login` now rejects), but it appears to leave `MarmotAppRuntime::login` (runtime/mod.rs:3170) with no remaining in-repo callers. Not asking you to delete it in this PR — just flagging so it's a deliberate decision rather than an accident.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -696,16 +696,60 @@ fn wipe_failure_reason(err: &AppError) -> String {
...
+                "publish_initial_key_package",
+                &self.publish_initial_key_package,
+            )
+            .finish()
+    }
+}
+
+impl AccountSetupRequest {
+    /// Copies relay/publish options for another setup attempt. Clears identity and `import_nsec`.
+    pub fn fresh_attempt(&self) -> Self {
```
erskingardner: The name says "another attempt at the same setup," but it also silently drops `identity`. Every current caller has `identity: None`, so it's a no-op today — which is exactly how this becomes a bug later, when someone reaches for it on a request that does carry an identity and quietly gets an account-creation instead of an import.

At minimum rename to something that states the clearing (`relay_options_only()`), or make the doc comment the first line rather than trailing prose.

erskingardner (commented):
The sidecar design is the right shape for #1209: `ImportNsec` is non-`Clone`, redacted in `Debug`, moved rather than copied, and `AccountSetupRequest` now cleanly separates public identity from private-key material. Validation-before-mutation in `create_or_import_account` is a genuine improvement, and the `identity`/`import_nsec` cross-check is a nice addition. Tests cover the sidecar roundtrip, the local-fallback ownership handoff, and redaction on both crates.

Three things I'd like addressed before merge, none of them large:

1. **The argv gate is still case-sensitive.** The PR deliberately made `marmot_app::is_nostr_secret` case-insensitive, but `crates/cli/src/lib.rs` keeps its own `starts_with("nsec")`. That's the copy guarding `SecretArgumentRejected`, so `wn login NSEC1...` still slips a plaintext secret into the `Cli` tree and onto the daemon JSON frame — exactly the leak class this PR exists to close. It's rejected eventually, but only after traversing every non-zeroizing buffer.
2. **~45 lines of duplicated client code that don't need to exist.** `send_owned_daemon_request` is `send_request` with ownership plumbing, plus a `panic!` helper for an unreachable case. `send_request` already takes `&DaemonRequest`, so the caller can just keep the binding and recover it on error.
3. **The changelog overstates the guarantee.** "wipe nsec-bearing request framing buffers on every path" isn't quite true — the `BufReader` on the daemon read path and `serde_json::to_vec`'s realloc growth both leave unzeroed copies. Best-effort is fine; the claim should match.

Smaller notes inline. The `fresh_attempt()` addition is the one design choice I'd push back on more broadly: `Zeroizing<String>` is itself `Clone`, so dropping `Clone` from `AccountSetupRequest` was a choice, and it cost ~250 lines of mechanical test churn across two crates plus a new public method whose name doesn't convey that it also clears `identity`. Worth a sentence in the PR body explaining why that tradeoff was taken.

---

### Finding 1 in detail (`crates/cli/src/lib.rs:985-987`, unchanged by this PR so not inline-commentable)

```rust
pub(crate) fn is_nostr_secret(value: &str) -> bool {
    value.starts_with("nsec")
}
```

This is the CLI-side gate feeding `validate_materialized_secret_identity` / `SecretArgumentRejected`, and it's still case-sensitive — while this same PR changed `marmot_app::is_nostr_secret` (`crates/marmot-app/src/runtime/mod.rs:4351`) to `eq_ignore_ascii_case`, with a test asserting the uppercase form is rejected.

The asymmetry has a concrete consequence: `wn login NSEC1J4C...` is **not** caught by `SecretArgumentRejected`, so it lands in `Cli.command.identity` as a plain `String`, gets JSON-encoded into the daemon `Execute` frame, and is buffered on both sides — precisely the plaintext-copy path #1209 is about. It only fails later, in `validate_account_setup_request`, after all of those copies already exist.

Suggest making it match:

```rust
pub(crate) fn is_nostr_secret(value: &str) -> bool {
    value
        .get(..4)
        .is_some_and(|prefix| prefix.eq_ignore_ascii_case("nsec"))
}
```

and adding the uppercase case to the `SecretArgumentRejected` tests, mirroring `account_setup_request_rejects_and_redacts_uppercase_nsec_identity`. Better still, export one hardened helper from `marmot-app` and drop the CLI and UniFFI copies entirely.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4292,7 +4349,57 @@ impl AccountManager {
...
+    identity.as_ref().map(|value| {
+        if is_nostr_secret(value) {
+            "**redacted**"
+        } else {
+            value.as_str()
+        }
+    })
+}
+
+enum AccountSetupOperation {
```
erskingardner, line 4364: All three rejection branches return `AppError::InvalidPublicKey`, including the two that have nothing to do with a malformed public key: "a secret was supplied to an operation that doesn't accept one" and "identity and import_nsec belong to different accounts." A user who mistypes one of a matched npub/nsec pair gets told their public key is invalid, which points at the wrong input.

These surface through the CLI via `map_account_setup_error`, so the wrong error text reaches the terminal. Consider distinct variants (e.g. `UnexpectedPrivateKey` / `IdentityKeyMismatch`) — the redaction property is unaffected either way since neither would embed the key material.

crates/cli/src/daemon/runtime_host.rs
```diff
@@ -72,25 +73,27 @@ pub(crate) async fn reconcile_and_clone_runtime(
...
 pub(crate) async fn handle_app_runtime_account_setup_request(
     cli: &Cli,
+    import_nsec: &mut Option<crate::ImportNsec>,
     defaults: &DaemonDefaults,
     state: Arc<Mutex<DaemonState>>,
     events: DaemonEventHub,
     workers: &SharedDaemonWorkers,
 ) -> Option<CliOutput> {
-    let request = match app_runtime_account_setup_request(cli) {
+    if !app_runtime_enabled(defaults) {
```
erskingardner, line 82: Moving the `app_runtime_enabled` check ahead of `app_runtime_account_setup_request` changes behavior beyond the sidecar work: with app-runtime hosting disabled, a malformed setup request previously returned its typed error from here, and now falls through to `run_cli_local`. The local path produces comparable errors for the common cases, but not identical ones (e.g. `relay_endpoints` failures over `daemon_default_account_relays` are no longer evaluated at all).

I think the new order is the more defensible one — validating daemon-hosting-specific config when hosting is off was arguably wrong — but it isn't mentioned in the PR body or CHANGELOG, and nothing in the tests pins it. Worth a line in the changelog entry, or a test asserting the fall-through.

crates/cli/src/daemon/protocol.rs
agent-p1p: Addressed in 9109a560. `send_execute` now uses the shared `send_request(&request)` path and destructures the known `Execute` request only on error to recover the original `Cli` and zeroizing sidecar. Deleted `send_owned_daemon_request` and `take_execute_request`; the connect-failure regression pins recovery semantics.

crates/cli/src/daemon/protocol.rs
agent-p1p: Addressed in 9109a560. The changelog now limits the claim to final owned framing buffers under our control and explicitly states that allocator reallocations and Tokio `BufReader` read-ahead are not guaranteed wiped. Matching comments document both residual paths in the protocol code.

crates/marmot-uniffi/src/commands/account.rs
agent-p1p: Addressed in 9109a560. `marmot_app::is_nostr_secret` is now the single conservative ASCII-case-insensitive classifier used by CLI argv/stdin handling, TUI routing/redaction, runtime validation, and UniFFI. Added uppercase argv/TUI/runtime regressions. `MarmotAppRuntime::login` intentionally remains public Rust API for downstream callers despite having no workspace call site; that tradeoff is recorded in the PR body.

crates/marmot-app/src/runtime/mod.rs
agent-p1p: Addressed in 9109a560. Renamed the helper to `relay_options_only` and documented its exact contract: copy only default/bootstrap relay options while clearing identity/import secret fields. This keeps multi-account setup reuse explicit without restoring broad `Clone` on a secret-bearing request.

crates/marmot-app/src/runtime/mod.rs
agent-p1p: Addressed in 9109a560. Added `AppError::UnexpectedPrivateKey` and `AppError::IdentityKeyMismatch`, used them in the three rejection branches, and projected them through privacy-safe kinds, CLI JSON, UniFFI, simulator, and client diagnostics. Focused tests pin each typed outcome without carrying secret data in the variants.

crates/cli/src/daemon/runtime_host.rs
agent-p1p: Addressed in 9109a560. `disabled_app_runtime_skips_relay_validation_and_preserves_nsec_sidecar` sets `relay: None` plus deliberately malformed daemon relay defaults, then proves the handler returns `None` before parsing them and retains the exact sidecar for local execution. The changelog and PR rationale now document the fall-through contract.

agent-p1p:
Second-pass review feedback is addressed in signed agent-p1p commit 9109a560: shared case-insensitive nsec classification (including TUI redaction/routing), deleted daemon transport duplication with recoverable ownership, typed private-input/mismatch errors, explicit relay-only retry contract, accurate best-effort zeroization wording, and malformed-default disabled-runtime fall-through coverage. Local `just fast-ci` and the wn-cli, marmot-app, marmot-uniffi, and agent-connector suites pass. No version/lock/cohort churn. Awaiting the new GitHub CI run.

PR #1216, opened 2026-08-01
Add full-system convergence quiescence

erskingardner:
## Summary

- add a privacy-safe structural progress snapshot spanning engine work, pass deadlines, deferred retries, publication acknowledgements, transport queues, scenario inputs, and terminal engine state
- add a bounded virtual-time fixed-point driver and ScenarioSpec await_quiescence step with explicit publication and transport policy
- emit quiescent, blocked, and watchdog-timeout artifacts in scenario reports and oracle coverage
- add real-engine deadline, restart, lost-ack, withheld-transport, watchdog, serialization/privacy, and batch-versus-multiple-pass metamorphic coverage
- mark Milestone 1.3 complete in the convergence reliability plan

## Design and risk

The engine change is read-only and available only behind the existing test-conformance-snapshot feature. It exposes sanitized aggregate scheduling state and does not alter branch selection, convergence constants, publication behavior, or production runtime scheduling. Fixed-point policy and all automatic acknowledgement/delivery behavior remain simulator-owned.

Watchdog limits produce timeout evidence only; they cannot turn unfinished work into quiescence. Manual publication or transport policy produces a named blocked result rather than spinning.

## Verification

- cargo test -p cgka-conformance-simulator
- cargo test -p cgka-conformance-simulator --lib
- cargo test -p cgka-engine --features test-policy-overrides,test-conformance-snapshot
- just fast-ci

## Summary by CodeRabbit

- **New Features**
  - Added bounded virtual-time settling with configurable watchdog limits.
  - Added structural progress reporting and sanitized diagnostics for pending work, blockers, transport, retries, and scheduling.
  - Added quiescence scenario steps with quiescent, blocked, and timed-out results.
  - Added capability validation for progress tracking, virtual time, transport delivery, and outbound publication.

- **Documentation**
  - Updated simulator and reliability documentation for quiescence, progress reporting, and supported scenario steps.

- **Tests**
  - Added coverage for delivery batching, acknowledgements, delayed transport, wake advancement, blockers, and timeout behavior.

### DIFF
--- a/crates/cgka-conformance-simulator/src/pending_work.rs
+++ b/crates/cgka-conformance-simulator/src/pending_work.rs
@@ -1,16 +1,24 @@
 //! Instantaneous pending-work observations for strict simulator assertions.
 //!
-//! This deliberately does not decide quiescence over time. It captures one
-//! privacy-safe, client-scoped local execution boundary; Milestone 1.3 will
-//! add structural progress tokens and bounded virtual-time fixed-point
-//! evaluation. It does not assert end-to-end delivery of transport objects
-//! that a scenario dropped.
+//! The pending-work form captures one privacy-safe, client-scoped local
+//! execution boundary. The structural form is consumed by the bounded
+//! virtual-time fixed-point driver, but it still does not assert end-to-end
+//! delivery of transport objects that a scenario dropped.
 
-use cgka_engine::conformance_snapshot::ConformancePendingWorkSnapshot;
+use cgka_engine::conformance_snapshot::{
+    ConformancePendingWorkSnapshot, ConformanceStructuralProgressSnapshot,
+};
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub(crate) struct BusPendingWorkSnapshot {
+    pub queued_messages: usize,
+    pub delayed_messages: usize,
+    pub mailbox_messages: usize,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub(crate) struct BusStructuralProgressSnapshot {
     pub queued_messages: usize,
     pub delayed_messages: usize,
     pub mailbox_messages: usize,
@@ -50,6 +58,91 @@ impl PendingWorkObservation {
         }
         if self.scenario_inputs_pending > 0 {
             blockers.push("scenario_inputs");
+        }
+        blockers
+    }
+}
+
+/// One participant's sanitized engine scheduling state.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ClientStructuralProgress {
+    pub client: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub engine: Option<ConformanceStructuralProgressSnapshot>,
+    pub scenario_inputs_pending: usize,
+}
+
+/// Terminal engine state that cannot be resolved by running or advancing the
+/// subject. The client label is a scenario-local name, never an account id.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "kind", rename_all = "snake_case")]
+pub enum SubjectTerminalBlocker {
+    EngineUnrecoverable { client: String },
+}
+
+/// Adapter-neutral structural state consumed by the virtual-time fixed-point
+/// driver. The token is diagnostic and cycle-detection evidence; quiescence is
+/// defined by the explicit work/deadline fields, never token equality alone.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SubjectProgressSnapshot {
+    pub schema_version: String,
+    pub structural_token: String,
+    pub current_monotonic_ms: u64,
+    pub runnable_work: usize,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub earliest_next_wake_monotonic_ms: Option<u64>,
+    pub deferred_retry_work: usize,
+    pub outbound_awaiting_acknowledgement: usize,
+    pub transport_queued_messages: usize,
+    pub transport_delayed_messages: usize,
+    pub transport_mailbox_messages: usize,
+    pub scenario_inputs_pending: usize,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub terminal_blockers: Vec<SubjectTerminalBlocker>,
+    pub clients: Vec<ClientStructuralProgress>,
+}
+
+impl SubjectProgressSnapshot {
+    pub fn is_quiescent(&self) -> bool {
+        self.runnable_work == 0
+            && self.earliest_next_wake_monotonic_ms.is_none()
+            && self.deferred_retry_work == 0
+            && self.outbound_awaiting_acknowledgement == 0
+            && self.transport_queued_messages == 0
+            && self.transport_delayed_messages == 0
+            && self.transport_mailbox_messages == 0
+            && self.scenario_inputs_pending == 0
+            && self.terminal_blockers.is_empty()
+    }
+
+    pub fn blocking_subsystems(&self) -> Vec<&'static str> {
+        let mut blockers = Vec::new();
+        if self.runnable_work > 0 {
+            blockers.push("runnable_work");
+        }
+        if self.earliest_next_wake_monotonic_ms.is_some() {
+            blockers.push("scheduled_wake");
+        }
+        if self.deferred_retry_work > 0 {
+            blockers.push("deferred_retry");
+        }
+        if self.outbound_awaiting_acknowledgement > 0 {
+            blockers.push("outbound_acknowledgement");
+        }
+        if self.transport_queued_messages > 0 {
+            blockers.push("transport_queue");
+        }
+        if self.transport_delayed_messages > 0 {
+            blockers.push("transport_delayed");
+        }
+        if self.transport_mailbox_messages > 0 {
+            blockers.push("transport_mailboxes");
+        }
+        if self.scenario_inputs_pending > 0 {
+            blockers.push("scenario_inputs");
+        }
+        if !self.terminal_blockers.is_empty() {
+            blockers.push("terminal_engine_state");
         }
         blockers
     }
--- a/crates/cgka-conformance-simulator/src/pending_work.rs
+++ b/crates/cgka-conformance-simulator/src/pending_work.rs
@@ -81,13 +81,17 @@ pub enum SubjectTerminalBlocker {
 }
 
 /// Adapter-neutral structural state consumed by the virtual-time fixed-point
-/// driver. The token is diagnostic and cycle-detection evidence; quiescence is
-/// defined by the explicit work/deadline fields, never token equality alone.
+/// driver. The token is diagnostic only; quiescence is defined by the explicit
+/// work/deadline fields, never token equality.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct SubjectProgressSnapshot {
     pub schema_version: String,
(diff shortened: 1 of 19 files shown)


crates/cgka-engine/src/conformance_snapshot.rs
```diff
@@ -465,6 +488,122 @@ pub(crate) fn capture_pending_work_snapshot<S: StorageProvider>(
...
+    let current_epoch = engine.epoch_manager.epoch(group_id).unwrap_or_default().0;
+    let now = engine.convergence_now();
+    let pass = engine.storage.convergence_pass(group_id)?;
+
+    let mut runnable_work = 0usize;
+    let mut next_wake = None;
+    if let Some(active) = pass.as_ref().filter(|pass| pass.is_active()) {
+        match active.phase {
+            ConvergencePassPhase::Collecting => {
+                let cutoff = if active.clock_instance_id == engine.convergence_clock_instance_id {
```
erskingardner: This duplicates the rebase in `distributed_convergence.rs` (the `pass.clock_instance_id != self.convergence_clock_instance_id` branch), but it starts from `engine.storage.convergence_pass()` — the raw persisted pass.

The production loader applies the `test-policy-overrides` deadline rewrite (`settlement_quiescence_ms == 0` → deadlines collapse to `opened_*_ms`) *before* rebasing, and that rewrite is deliberately not persisted. So on a `test-policy-overrides` build this snapshot reads the original, non-collapsed `cutoff_wall_ms()` and reports a future `earliest_next_wake_monotonic_ms` for a cutoff the engine already treats as due — which then makes the driver burn a virtual-time advance that the engine did not need. That also weakens the `virtual_time_advanced_ms` assertions in the new tests.

Please route this through the same loader the engine uses (or extract the rebase into one shared helper) rather than keeping a second copy of the deadline arithmetic. The same applies to the deferred-peel rebase just below at line 544.

crates/cgka-conformance-simulator/src/quiescence.rs
```diff
@@ -0,0 +1,481 @@
...
+        }
+        if !snapshot.terminal_blockers.is_empty() {
+            return Ok(blocked_observation(
+                step_index, policy, iterations, work_units, started_at, snapshot,
+            ));
+        }
+        if (snapshot.outbound_awaiting_acknowledgement > 0
+            && policy.outbound == QuiescenceOutboundPolicy::Manual)
+            || ((snapshot.transport_queued_messages > 0 || snapshot.transport_mailbox_messages > 0)
+                && policy.transport == QuiescenceTransportPolicy::Manual)
```
erskingardner, line 140: `transport_mailbox_messages > 0` under `QuiescenceTransportPolicy::Manual` returns `Blocked` here, but the tick branch at line 173 lists `snapshot.transport_mailbox_messages > 0` as a reason to tick — and `tick` is not gated by `TransportDelivery`. Mailbox contents are already past the transport hop; the thing that drains them is the client runtime, not `deliver_all`.

As written, a manual-transport `await_quiescence` can never settle while any mailbox is non-empty, even though the driver knows how to drain it. I think the manual gate should cover `transport_queued_messages` only.

There is also no test for `QuiescenceTransportPolicy::Manual` anywhere in the PR, which is probably why this did not surface — worth either covering it or dropping the variant until something needs it.

crates/cgka-conformance-simulator/src/quiescence.rs
```diff
@@ -0,0 +1,481 @@
...
+                policy,
+                QuiescenceWatchdog::Iterations,
+                iterations,
+                work_units,
+                started_at,
+                snapshot,
+            ));
+        }
+
+        let engine_runnable = snapshot.runnable_work.saturating_sub(
```
erskingardner: Reconstructing engine work by subtracting the transport counts back out of `runnable_work` couples this adapter-neutral driver to one specific subject's aggregation choice.

`SubjectProgressSnapshot` documents `runnable_work` as "runnable work" and never says it must include `transport_queued_messages + transport_mailbox_messages`. `EngineHarnessSubject::structural_progress` does add them in, so the subtraction is exact today — but a second adapter that reports engine-only runnable work gets `engine_runnable == 0` from `saturating_sub` whenever transport counts are non-zero, skips the tick, and reports a spurious `Blocked`.

Simplest fix: have the subject report engine runnable work in its own field instead of aggregating and then un-aggregating. Failing that, the invariant needs to be stated on the struct where adapter authors will see it.

crates/cgka-conformance-simulator/src/quiescence.rs
```diff
@@ -0,0 +1,481 @@
...
+            snapshot
+                .transport_queued_messages
+                .saturating_add(snapshot.transport_mailbox_messages),
+        );
+        let will_ack = policy.outbound == QuiescenceOutboundPolicy::AcceptAll;
+        let will_deliver = policy.transport == QuiescenceTransportPolicy::DeliverAll;
+        let will_tick = engine_runnable > 0
+            || snapshot.transport_mailbox_messages > 0
+            || (snapshot.transport_queued_messages > 0 && will_deliver);
+        let planned_work = (if will_ack {
```
erskingardner, line 165: `planned_work` (lines 168-184) and `attempted_work` (lines 190-228) evaluate the same three quantities under the same three conditions, once to check the budget and once to charge it. `will_ack` / `will_deliver` / `will_tick` are computed for the first copy and then re-inlined as raw comparisons in the second.

Compute the plan once, check it against `max_work_units`, execute it, and charge the same value — that should remove around 35 lines with no behavior change.

Separately: neither `WorkUnits` return site (here and line 257) has a test. `unchanged_runnable_work_times_out_without_becoming_quiescent` sets `max_work_units: 10` but trips the iteration budget first, so `max_work_units` is untested and unused by any caller in the repo. Two watchdogs already cover the milestone requirement; this third one and its bookkeeping look speculative.

crates/cgka-conformance-simulator/src/quiescence.rs
```diff
@@ -0,0 +1,481 @@
...
+pub async fn drive_subject_to_quiescence(
+    subject: &mut dyn ConvergenceSubject,
+    clients: &[String],
+    policy: &QuiescencePolicy,
+    step_index: usize,
+) -> Result<QuiescenceObservation, SubjectError> {
+    // Select the subject's controlled clock before the first participant wake.
+    // A zero delta is semantically inert but prevents legacy harness adapters
+    // from substituting their historical far-future one-shot drain shortcut.
+    subject.advance_time(0).await?;
```
erskingardner: This is a magic zero-delta call made purely for a side effect: `EngineHarnessSubject::advance_time` also runs `client.enable_virtual_time_tick()` for every client, and that flag is the actual thing being set here. The comment explains the motivation but the code does not express the intent, and it only works because of one subject's implementation detail — which is what an adapter-neutral driver should not depend on.

Prefer either an explicit subject operation for selecting the controlled clock, or having the subject enable virtual-time ticking from `structural_progress`/`tick` so the driver does not need this at all.

It also means `AwaitQuiescence` hard-requires `VirtualTime` even for a fully manual policy, which is inconsistent with how `OutboundPublication` is made conditional in `required_capabilities`.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -382,6 +392,25 @@ pub fn required_capability(step: &ScenarioStep) -> SubjectCapability {
     }
 }
 
+/// Complete capability set used by a scenario step. Most operations need one
+/// capability; fixed-point settling composes progress, time, delivery, and
+/// optionally outbound acknowledgement without moving policy into the subject.
+pub fn required_capabilities(step: &ScenarioStep) -> Vec<SubjectCapability> {
```
erskingardner, line 368: After this change `required_capability` has exactly one production caller — the `else` arm of this function — yet it stays `pub`, stays re-exported from `lib.rs`, and returns a knowingly incomplete answer (`StructuralProgress` alone) for `AwaitQuiescence`. Any external caller that reaches for the singular form now gets a wrong result for one step kind.

Either fold the match into `required_capabilities` and drop the singular function, or make it private. Keeping both public leaves a trap whose only remaining use is the test at scenario.rs:920, which had to be rewritten to `crate::required_capability` for exactly this reason.

crates/cgka-conformance-simulator/src/pending_work.rs
```diff
@@ -54,3 +62,88 @@ impl PendingWorkObservation {
...
+    EngineUnrecoverable { client: String },
+}
+
+/// Adapter-neutral structural state consumed by the virtual-time fixed-point
+/// driver. The token is diagnostic and cycle-detection evidence; quiescence is
+/// defined by the explicit work/deadline fields, never token equality alone.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SubjectProgressSnapshot {
+    pub schema_version: String,
+    pub structural_token: String,
```
erskingardner, line 89: The doc comment calls this "cycle-detection evidence", but nothing in `quiescence.rs` reads `structural_token` — the driver's loop is bounded purely by `max_iterations` and by `attempted_work == 0`. And because the hashed snapshot includes `current_monotonic_ms`, the token necessarily changes on every time advance, so it could not detect a structural cycle across wakes even if something did consume it.

#1189 does ask for a sanitized progress token, so keeping the field is fine — but please either wire it into an actual repeated-state check or narrow the doc comment to "diagnostic only", so the next reader does not assume cycle detection exists.

erskingardner (commented):
The premise is right: #1189 explicitly asks for structural quiescence over a fixed drain count, milestone 1.3 names exactly these fields, and the observation/blocker/watchdog split is the correct shape. CI is green and the real-engine tests (exact 1,000 ms wake, restart rebase to 600 ms, withheld transport at 0 iterations) are genuinely good evidence.

My concern is that this is an **oracle**, so its own correctness matters more than usual, and three things in it are either divergent from production or contradictory with themselves.

### Should be fixed before merge

1. **`capture_structural_progress_snapshot` re-implements convergence-pass deadline rebasing and reads the pass raw from storage.** The production loader in `distributed_convergence.rs` applies the `test-policy-overrides` deadline rewrite *and then* rebases across a clock-instance change, and that rewrite is never persisted. The snapshot reads `engine.storage.convergence_pass()` directly, so under `test-policy-overrides` (one of this PR's own verification commands) it can report a far-future `earliest_next_wake_monotonic_ms` for a cutoff the engine already considers due. An oracle that computes deadlines differently from the engine can mask the exact class of bug 1.3 exists to catch.

2. **Manual transport policy contradicts the tick plan.** The blocked check treats `transport_mailbox_messages > 0` as terminal under `QuiescenceTransportPolicy::Manual`, but the tick branch fifty lines later would happily drain mailboxes with a plain `tick` that needs no transport capability at all. Mailboxes are post-delivery. As written, `Manual` transport can never reach quiescence while any mailbox is non-empty.

3. **`engine_runnable` is recovered by subtracting transport counts back out of `runnable_work`.** That is exact only because `EngineHarnessSubject` happens to aggregate them in. `SubjectProgressSnapshot` is documented as adapter-neutral and never states that invariant, so a conforming adapter reporting engine-only `runnable_work` silently gets `engine_runnable == 0` and premature `Blocked`.

### Simplicity (guidelines 2 and 4)

- `planned_work` and `attempted_work` are the same expression computed twice under the same three conditions, ~35 lines that want to be one.
- `QuiescenceTransportPolicy::Manual`, `max_work_units`, and both `WorkUnits` return sites have no test and no in-repo caller. Two of the three watchdogs are exercised; this one is speculative until something needs it.
- `required_capability` is now called only by `required_capabilities`, yet stays `pub` and re-exported while returning a knowingly incomplete answer for `AwaitQuiescence`.
- `structural_token` is computed but never read by the driver, and it hashes `current_monotonic_ms`, so it cannot serve the cycle detection its own doc comment claims.

None of this is hard to fix, and the overall design is the one the plan asked for. Happy to re-review quickly.

PR #1217, opened 2026-08-02
Add deterministic convergence failure capsules

erskingardner:
## What changed

- add a versioned convergence failure-capsule schema with stable fingerprints, expanded schedules, exact policies/constants, resource counters, report evidence, and restrictive artifact writes
- retain exact emitted transport objects and add optional sensitive SQLite/OpenMLS byte replay, including active durable convergence-pass state and paired virtual-clock values
- make engine-subject tick failures typed report artifacts instead of silently discarding `EngineError`s
- minimize generated failures only when the complete fingerprint is preserved, and promote only synthetic capsules that retain the original intended expectation
- expose a machine-readable snapshot of the engine-owned convergence constants without changing their values or selection behavior
- strengthen the mixed large-group storm so application delivery remains asserted across the subsequent commit storm
- complete the Milestone 1.4 and Milestone 1 exit tracking/evidence in the reliability plan

## Why

Milestone 1 needs the in-process engine simulator to recognize incorrect results and preserve enough evidence to explain and reproduce them before broader relay, app-runtime, process, and VM campaigns build on it. Previously, generated failures lacked one stable artifact, engine tick errors could be discarded by the subject adapter, and a portable promotion could accidentally bless the observed bad output.

## Runtime impact and risk

Normal convergence behavior is unchanged. Production engine changes are limited to a conformance-only constant snapshot and crate-private visibility for existing constants; their values and call sites are unchanged. SQLite replay export/import is behind the `test-conformance-replay` feature used by the simulator. Ordinary production group snapshot/rollback semantics still exclude convergence-pass scheduling state; only the sensitive test replay envelope restores it.

Automatically emitted synthetic capsules contain logical reports and exact synthetic transport artifacts. Byte-level replay additionally requires an explicitly constructed `sensitive_local` capsule with the recipient checkpoint; those checkpoints contain key material and are written owner-only.

## Verification

- `cargo test -p storage-sqlite --features test-conformance-replay` — 309 passed
- `cargo test -p cgka-conformance-simulator` — full suite green, including 41 canonical scenarios, 11 property tests, 16 report-runner tests, and exact failure-capsule replay
- `just fast-ci`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added versioned failure capsules with reports, transport artifacts, fingerprints, and optional replay checkpoints.
  * Added CLI replay support with reproduced-failure verification.
  * Added structured failure reporting and captured transport details.
  * Added promotion of eligible synthetic failures to portable vector fixtures.
  * Added durable replay snapshots and a published capsule schema.

* **Bug Fixes**
  * Failure minimization now preserves complete failure fingerprints.
  * Scenario execution surfaces failed steps and engine errors.
  * Mixed storm scenarios retain and validate application messages through commits.

* **Documentation**
  * Updated simulator guides, scenarios, and reliability planning documentation.

### DIFF
--- a/crates/cgka-conformance-simulator/README.md
+++ b/crates/cgka-conformance-simulator/README.md
@@ -34,8 +34,8 @@ welcomes use NIP-59 gift wraps before the bus delivers them.
 - `ScenarioReport` — serializable run artifacts with metadata, expected and observed traces, oracle coverage evidence,
   step logs, recoveries, and expectation failures.
 - `FailureCapsuleV1` — a versioned failure artifact containing the scenario, expanded virtual-time schedule, exact
-  policy/constants, report/ledgers/state commitments, captured transport objects, resource counters, and a stable
-  failure fingerprint. An optional sensitive recipient checkpoint enables byte-exact engine replay.
+  policy/constants, report/ledgers/state commitments, a bounded transport-evidence tail with truncation counters, and a
+  stable failure fingerprint. An optional bounded sensitive recipient checkpoint enables byte-exact engine replay.
 - `ConformanceGroupSnapshot` — a feature-gated synthetic-test projection of exact canonical group state: leaf identities
   and capabilities, required/application state, lifecycle/profile/admin state, a GroupContext hash, and the adopted
   domain-separated exporter commitment. Raw exporter material is never serialized.
@@ -343,7 +343,9 @@ the conformance contract clearer.
 
 `run_generated_case_report(case, expected_trace)` adds generated-family metadata: family name, generator version, seed,
 case index, and an optional `minimized_case` field. Failing generated cases run a conservative greedy minimizer that
-removes removable delivery/app steps only when the complete stable failure fingerprint still reproduces. Generated report runs also write
+removes removable delivery/app steps only when the semantic failure identity (classification, action type, and failure
+kind) still reproduces. The complete fingerprint, including the state digest, remains in the capsule for diagnosis.
+Generated report runs also write
 a sibling `*-fixture.v1.json` candidate. Cases with semantic expectations keep those expectations; cases without them
 use the observed trace as an exact expected trace in the candidate. When a failing generated case has a minimized
 reproducer, the fixture candidate uses that minimized scenario.
--- a/crates/cgka-conformance-simulator/README.md
+++ b/crates/cgka-conformance-simulator/README.md
@@ -35,7 +35,8 @@ welcomes use NIP-59 gift wraps before the bus delivers them.
   step logs, recoveries, and expectation failures.
 - `FailureCapsuleV1` — a versioned failure artifact containing the scenario, expanded virtual-time schedule, exact
   policy/constants, report/ledgers/state commitments, a bounded transport-evidence tail with truncation counters, and a
-  stable failure fingerprint. An optional bounded sensitive recipient checkpoint enables byte-exact engine replay.
+  stable failure fingerprint. Explicitly requested byte-exact replay is written to a separately named sensitive sibling
+  capsule so the logical campaign capsule remains portable.
 - `ConformanceGroupSnapshot` — a feature-gated synthetic-test projection of exact canonical group state: leaf identities
   and capabilities, required/application state, lifecycle/profile/admin state, a GroupContext hash, and the adopted
   domain-separated exporter commitment. Raw exporter material is never serialized.
@@ -102,10 +103,11 @@ selected subject adapter, declared capabilities, storage backend, observed trace
 observations, and the mismatched expected/actual JSON. A subject that lacks any capability required by a scenario is
 rejected before the first scenario action executes.
 
-Failed report runs also write `*-failure-capsule.v1.json`. The capsule writer uses owner-only directories and files.
-Synthetic capsules without checkpoints may be reviewed and promoted into portable vectors; real incident material and
-any capsule containing an SQLite/OpenMLS checkpoint must remain `sensitive_local` because the checkpoint contains key
-material. Replay a checkpoint-bearing capsule without regenerating MLS bytes with:
+Failed report runs also write a portable `*-failure-capsule.v1.json`. The capsule writer uses owner-only directories and
+files. Sensitive SQLite/OpenMLS checkpoint capture is off by default because it contains key material and incurs a
+checkpoint export. Pass `--capture-sensitive-replay` to capture only the final planned recipient tick and write a
+separate `*-sensitive-replay-capsule.v1.json`; the portable logical capsule remains eligible for vector promotion.
+Replay the sensitive sibling without regenerating MLS bytes with:
 
 ```sh
 cargo run -p cgka-conformance-simulator --bin cgka-conformance-simulator-report -- \
--- a/crates/cgka-conformance-simulator/README.md
+++ b/crates/cgka-conformance-simulator/README.md
@@ -33,11 +33,14 @@ welcomes use NIP-59 gift wraps before the bus delivers them.
   outcomes.
 - `ScenarioReport` — serializable run artifacts with metadata, expected and observed traces, oracle coverage evidence,
   step logs, recoveries, and expectation failures.
+- `FailureCapsuleV1` — a versioned failure artifact containing the scenario, expanded virtual-time schedule, exact
+  policy/constants, report/ledgers/state commitments, captured transport objects, resource counters, and a stable
+  failure fingerprint. An optional sensitive recipient checkpoint enables byte-exact engine replay.
 - `ConformanceGroupSnapshot` — a feature-gated synthetic-test projection of exact canonical group state: leaf identities
   and capabilities, required/application state, lifecycle/profile/admin state, a GroupContext hash, and the adopted
   domain-separated exporter commitment. Raw exporter material is never serialized.
-- `cgka-conformance-simulator-report` — a small CLI that runs generated scenario families, writes JSON reports, and
-  emits fixture candidates for generated cases.
+- `cgka-conformance-simulator-report` — a small CLI that runs generated scenario families, writes JSON reports and
+  failure capsules, emits fixture candidates for generated cases, and replays capsules that contain byte checkpoints.
 - `proptest_support` — strategies that generate arbitrary typed `SendIntent` sequences for property-based tests.
 
 ## Testing layers
@@ -98,6 +101,19 @@ The report command exits non-zero when any fixture expectation fails. Each repor
 selected subject adapter, declared capabilities, storage backend, observed trace, flattened recovery and epoch
 observations, and the mismatched expected/actual JSON. A subject that lacks any capability required by a scenario is
 rejected before the first scenario action executes.
+
+Failed report runs also write `*-failure-capsule.v1.json`. The capsule writer uses owner-only directories and files.
+Synthetic capsules without checkpoints may be reviewed and promoted into portable vectors; real incident material and
+any capsule containing an SQLite/OpenMLS checkpoint must remain `sensitive_local` because the checkpoint contains key
+material. Replay a checkpoint-bearing capsule without regenerating MLS bytes with:
+
+```sh
+cargo run -p cgka-conformance-simulator --bin cgka-conformance-simulator-report -- \
+  --replay-capsule /private/path/failure-capsule.v1.json
+```
+
+The replay succeeds only when the exact stored recipient state plus captured transport objects reproduce the recorded
+engine fingerprint. Capsule v1's JSON Schema is `schemas/failure-capsule.v1.schema.json`.
 
 ## Exact state oracle
 
@@ -299,7 +315,7 @@ trace includes exactly the selected branch application payload.
 semantic expectations. The generator rotates through invite forks, group-data forks, publish rollback plus delayed app
 duplicates, partition/heal/leave, delayed past-epoch app delivery, stable duplicate/delay/reorder queue faults, 20+
 client message storms, partitioned large-group delivery storms, multi-committer group-data storms, mixed large
-message/commit storms, and restart plus duplicate delivery faults. Generator version `4` draws the delivery schedule of
+message/commit storms, and restart plus duplicate delivery faults. Generator version `6` draws the delivery schedule of
 the rollback and storm shapes from the seed, so distinct seeds exercise distinct adversarial orderings while the
 schedule-invariant convergence, rollback, and payload-set expectations stay fixed. These cases are ordinary
 `ScenarioSpec`s, so the same runner and report path can turn selected generated cases into fixed vectors when that makes
@@ -316,7 +332,8 @@ the conformance contract clearer.
 - `observed_trace` — the trace produced by the scenario runner.
 - `oracle` — scenario stimuli, expected behavior classes, observed behavior classes, evidence counts, and weak-oracle
   warnings.
-- `step_log` — one entry per completed scenario step.
+- `step_log` — one entry per attempted scenario step, including the first typed failure; later planned steps remain
+  visible in a failure capsule's expanded schedule with no status.
 - `pending_resolution_observations` — flattened publish confirmations and rollbacks.
 - `recovery_observations` — flattened fork-recovery events from all client observations.
 - `epoch_change_observations` — flattened `EpochChanged` events from all client observations.
@@ -326,7 +343,7 @@ the conformance contract clearer.
 
 `run_generated_case_report(case, expected_trace)` adds generated-family metadata: family name, generator version, seed,
 case index, and an optional `minimized_case` field. Failing generated cases run a conservative greedy minimizer that
-removes removable delivery/app steps only when the same failure kinds still reproduce. Generated report runs also write
+removes removable delivery/app steps only when the complete stable failure fingerprint still reproduces. Generated report runs also write
 a sibling `*-fixture.v1.json` candidate. Cases with semantic expectations keep those expectations; cases without them
 use the observed trace as an exact expected trace in the candidate. When a failing generated case has a minimized
 reproducer, the fixture candidate uses that minimized scenario.
--- a/crates/cgka-conformance-simulator/src/failure_capsule.rs
+++ b/crates/cgka-conformance-simulator/src/failure_capsule.rs
@@ -1,0 +1,613 @@
+//! Versioned, replayable failure artifacts for convergence reliability runs.
+//!
+//! A capsule keeps logical scenario/report evidence separate from the opaque
+//! cryptographic checkpoint required for byte-exact MLS replay. Checkpoints are
+//! always sensitive local evidence and must be written with owner-only modes.
+
+use std::collections::BTreeMap;
+use std::error::Error;
+use std::fmt;
+use std::path::Path;
+use std::sync::Arc;
+
+use cgka_engine::ManualConvergenceClock;
+use cgka_engine::canonicalization::CanonicalizationPolicy;
+use cgka_traits::group::ProtocolProfile;
+use cgka_traits::transport::TransportMessage;
+use cgka_traits::types::GroupId;
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+
+use crate::{
+    ClientBuilder, ConformanceConstantSnapshot, HarnessStorageMode, QuiescencePolicy,
+    QuiescenceStatus, ScenarioReport, ScenarioSpec, ScenarioStep, ScenarioStepStatus,
+    SubjectDescriptor, TransportBus, conformance_constant_snapshot,
+};
+
+pub const FAILURE_CAPSULE_SCHEMA_VERSION: &str = "1";
+pub const FAILURE_FINGERPRINT_VERSION: &str = "1";
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum FailureCapsuleSensitivity {
+    SyntheticShareable,
+    SensitiveLocal,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum TerminalOutcomeClassification {
+    Converged,
+    ExpectedRefusal,
+    RetryableBlock,
+    TerminalProtocolFailure,
+    ResourceFailure,
+    EnvironmentFailure,
+    Timeout,
+    OracleViolation,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct FailureFingerprintV1 {
+    pub version: String,
+    pub digest: String,
+    pub classification: TerminalOutcomeClassification,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub first_failing_action_id: Option<String>,
+    pub failure_kind: String,
+    pub normalized_state_digest: String,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ExpandedScenarioActionV1 {
+    pub action_id: String,
+    pub step_index: usize,
+    pub action_type: String,
+    /// Virtual monotonic time immediately before this action executes.
+    pub virtual_time_ms: u64,
+    /// `None` means execution stopped before this planned action.
+    pub status: Option<ScenarioStepStatus>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CapturedTransportArtifactV1 {
+    pub sequence: u64,
+    pub sender: String,
+    pub message: TransportMessage,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CapsulePolicySnapshotV1 {
+    pub canonicalization: CanonicalizationPolicy,
+    /// Exact policies selected by each `await_quiescence` action.
+    pub quiescence_by_action: BTreeMap<String, QuiescencePolicy>,
+    pub engine_constants: ConformanceConstantSnapshot,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ResourceObservationV1 {
+    pub counters: BTreeMap<String, u64>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct EngineByteReplayV1 {
+    pub client_label: String,
+    /// Exact seed passed to `ClientBuilder`; this is not a secret for synthetic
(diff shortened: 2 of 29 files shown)


erskingardner (commented):
Solid, well-tested work overall — the typed step failures, the fs-private capsule writes, and the sensitive/shareable split are all the right calls, and the test coverage is genuinely good. I'm commenting rather than approving because two of the headline claims don't look like they hold on the real families, and one milestone gate is marked complete on evidence this PR doesn't ship.

**Main concerns**

1. **Full-fingerprint minimization looks inert on the families it's meant to serve.** The fingerprint includes a SHA-256 over the entire `observed_trace` (plus quiescence/pending/epoch/invalidation observations). Every step class the reducer is allowed to remove — `SendAppMessage`, `ReorderQueued`, `ClearEvents`, `DropQueued`, `SetPartition` — changes that trace in any case that uses `ObserveExact`/strict payload-set expectations, which is exactly what `convergence-chaos/v1` now does. The one test that proves minimization still works (`failing_generated_case_records_a_minimized_reproducer`) uses the legacy portable `Observe` step, so the digest is blind to the removed message there. The net effect is that `minimized_case` will be `None` for the real chaos cases while still paying one full scenario re-run per removable step. Details inline.

2. **Byte replay has no automatic producer.** `write_synthetic_failure_capsule` always passes `byte_replay: None`, so the SQLite `test-conformance-replay` export/import path, the `ReplaySnapshot` envelope, and `--replay-capsule` are only ever exercised by a hand-assembled capsule in `tests/failure_capsules.rs`. That's a fair amount of new feature-gated production-crate surface for a path nothing in the campaign emits yet. Either wire the report runner to emit a checkpoint for the first failing recipient, or land the storage/CLI half when the producer arrives.

3. **`docs/marmot-architecture/convergence-reliability-plan.md` marks Milestone 1 `complete` and ticks "Targeted semantic mutations are detected" and "A captured byte-level failure replays with the same outcome".** The durable-outcome inventory right below it names a concrete test per row, which is the right standard; the mutation-sentinel paragraph names none, and the byte-replay gate rests on the manually built capsule from (2). Please name the sentinel tests the same way, or leave the gate open.

**Smaller things** (inline): every failed step is classified `environment_failure` regardless of its `SubjectError` code, so an `EngineError::Backend` surfaced by the new tick check lands in the "not our bug" bucket; `original_scenario`/`canonical_scenario` are byte-identical clones; `EngineByteReplayV1` duplicates two fields already in `expected_fingerprint` and then needs a validator to police the duplication; unbounded transport capture on every failure including strict weak-oracle warnings; and the reindented `async {}` block in `run_scenario_report_inner` left the match arms at mixed indentation.

None of the above is a correctness bug in shipped behavior — production paths are genuinely untouched — so this is a design/scope review, not a block on merge.

crates/cgka-conformance-simulator/src/family.rs
```diff
@@ -144,10 +158,7 @@ async fn minimize_failing_case(
     failing_report: &ScenarioReport,
     storage_mode: HarnessStorageMode,
 ) -> Option<ScenarioSpec> {
-    let target_failures = failure_kinds(failing_report);
-    if target_failures.is_empty() {
-        return None;
-    }
+    let target_fingerprint = fingerprint_report_failure(failing_report).ok()?;
```
erskingardner: Requiring the *complete* fingerprint means requiring `normalized_state_digest`, which is a SHA-256 over `observed_trace` + quiescence + pending-resolution + epoch-change + app-invalidation observations. Every step in `is_minimizer_removable` mutates that trace for any case with exact-state or payload-set expectations:

- removing a `SendAppMessage` drops a payload from every recipient's observed set,
- removing `ReorderQueued` changes delivery order and therefore the ordered payload set,
- removing `ClearEvents`/`DropQueued`/`SetPartition` changes observed events.

The `convergence-chaos/v1` cases this reducer exists for all carry strict `client_state` expectations (this PR adds more of them in the mixed storm), so `reproduces_failure` should return `false` for every candidate and `minimize_failing_case` should return `None` — after running the full scenario once per removable step. In a 21-client storm that's 20+ wasted full runs to produce nothing.

`failing_generated_case_records_a_minimized_reproducer` doesn't catch this because it uses the legacy portable `Observe` step, whose observation is only epoch/member-count/group-name — the digest can't see the removed app message.

The stated goal (don't let the reducer bless a *different* failure) is right, but the failure *identity* is `classification` + `failure_kind` + the violated invariant, not a digest of all observed state. Suggest matching on those fields and leaving `normalized_state_digest` out of the minimization predicate (keep it in the capsule fingerprint). Note also that `first_failing_action_id` embeds `step-{index}`, which shifts whenever an earlier step is removed — a second independent reason equality can't hold for the non-oracle branches.

crates/cgka-conformance-simulator/src/failure_capsule.rs
```diff
@@ -0,0 +1,612 @@
...
+        &report.quiescence_observations,
+        &report.pending_resolution_observations,
+        &report.epoch_change_observations,
+        &report.app_invalidation_observations,
+        failed_step_evidence,
+    ))?;
+
+    let failed_step = report.step_log.iter().find_map(|step| match &step.status {
+        ScenarioStepStatus::Failed { kind, .. } => Some((
+            TerminalOutcomeClassification::EnvironmentFailure,
```
erskingardner: Every failed step is classified `EnvironmentFailure`, discarding `error.kind` (which `subject_step_error` now carefully threads through from `SubjectError::code`).

The main point of the `ensure_tick_succeeded` change in this PR is that an `EngineError` from `tick` is no longer silently dropped — but it now lands in the taxonomy bucket that #1189 defines as *not* a protocol failure. `engine_subject_tick_surfaces_engine_failures` builds exactly such a case (`code == "backend"`), and triage would file it as environment noise.

Suggest mapping the subject error code: engine-originated codes (`backend`, `storage`, anything from `observe_engine_error`) → `TerminalProtocolFailure` or `ResourceFailure`, and reserve `EnvironmentFailure` for harness/setup codes like `unknown_client`.

crates/cgka-conformance-simulator/src/failure_capsule.rs
```diff
@@ -0,0 +1,612 @@
...
+                BTreeMap::from([
+                    ("generator_seed".into(), generated.seed),
+                    ("case_index".into(), generated.case_index),
+                ])
+            })
+            .unwrap_or_default();
+        let capsule = Self {
+            schema_version: FAILURE_CAPSULE_SCHEMA_VERSION.into(),
+            sensitivity,
+            original_scenario: report.scenario.clone(),
```
erskingardner: `original_scenario` and `canonical_scenario` are the same clone of `report.scenario`, and `report` (also embedded in the capsule) carries a third copy. For a 21-client storm scenario that's three identical serializations of the full step list in one file for zero added information.

Until there's an actual canonicalization pass that can make the two differ, drop one of the fields (or make `canonical_scenario` optional and populated only when it differs). The schema can gain it back in v2 when there's something to put in it.

crates/cgka-conformance-simulator/src/failure_capsule.rs
```diff
@@ -0,0 +1,612 @@
...
+    pub group_id: Vec<u8>,
+    /// Opaque serialized SQLite/OpenMLS group checkpoint. Contains key material.
+    pub sensitive_checkpoint: Vec<u8>,
+    pub captured_deliveries: Vec<TransportMessage>,
+    pub checkpoint_monotonic_ms: u64,
+    pub checkpoint_wall_ms: u64,
+    /// Whether the original harness tick used the injected clock rather than
+    /// the legacy far-future settlement shortcut.
+    pub virtual_time_tick_enabled: bool,
+    pub failure_kind: String,
```
erskingardner: `failure_kind` and `failing_action_id` duplicate `expected_fingerprint.failure_kind` and `expected_fingerprint.first_failing_action_id`, and `validate()` then exists partly to check that the two copies agree — a consistency check that only has to exist because of the duplication. `replay_engine_bytes` reads them from both places (`replay.failure_kind` alongside `replay.expected_fingerprint.classification`).

Dropping both fields and reading them off `expected_fingerprint` removes the fields, the validator branch, and a class of malformed capsule.

crates/cgka-conformance-simulator/src/report.rs
```diff
@@ -294,6 +351,21 @@ async fn run_vector_fixture_reports(
...
+fn write_synthetic_failure_capsule(
+    report: &ScenarioReport,
+    captured_transport_artifacts: Vec<crate::CapturedTransportArtifactV1>,
+    path: PathBuf,
+) -> Result<PathBuf, Box<dyn Error>> {
+    let capsule = FailureCapsuleV1::from_report(
+        report.clone(),
+        FailureCapsuleSensitivity::SyntheticShareable,
+        captured_transport_artifacts,
+        None,
```
erskingardner: This is the only non-test producer of a capsule and it always passes `byte_replay: None`. So `SqliteAccountStorage::{export,import}_conformance_replay_snapshot`, the `test-conformance-replay` feature, the `ReplaySnapshot` envelope, and `--replay-capsule` are reachable only from a capsule assembled by hand in `tests/failure_capsules.rs`.

That's new feature-gated surface in a production storage crate for a path no campaign emits. Either have this function export a checkpoint for the first failing recipient (which would also make the Milestone 1 byte-replay exit gate self-evidencing), or split the storage/CLI half out until there's a producer.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -498,6 +498,27 @@ impl EngineHarnessSubject {
             .ok_or_else(|| SubjectError::new("unknown_client", format!("unknown client {label}")))
     }
 
+    /// Exact transport artifacts emitted during this subject run. Payloads are
+    /// retained only for explicit failure-capsule capture.
+    pub fn captured_transport_artifacts(&self) -> Vec<crate::CapturedTransportArtifactV1> {
```
erskingardner: This captures every outbound emission from every client with full payloads, and `run_report` writes it for any non-empty `failures` list — including a strict-mode run whose only "failure" is a `weak_oracle_warning`. A 21-client storm capsule will embed the entire wire history plus the entire report.

`resource_observations` measures `captured_transport_json_bytes` after the fact but nothing bounds it. Consider capping the retained artifacts (e.g. the window around the first failing action) or skipping the transport capture for warning-only failures.

crates/cgka-conformance-simulator/src/scenario.rs
```diff
@@ -456,7 +495,8 @@ async fn run_scenario_report_inner(
 
     for (step_index, step) in spec.steps.iter().enumerate() {
         let action_id = scenario_input_id(step_index, step);
-        match step {
+        let step_result = async {
```
erskingardner: Wrapping the body in `async { ... }.await` to get `?` is reasonable, but the match arms weren't reindented — everything from `CreateGroup` down keeps the old indentation while the final `RestartClient` arm sits one level deeper, so the same `match` has two indentation levels. rustfmt evidently declines to touch a block this size, which is itself a signal: extracting the per-step dispatch into an `async fn execute_step(...)` would fix the formatting, shrink this function a lot, and make the new failure-capture tail readable.

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -115,7 +115,7 @@ The baseline is useful scaffolding, but it is not yet the target reliability lab
 | Milestone | Outcome | Status | Exit gate |
 | --- | --- | --- | --- |
 | 0. Assurance foundation | Guarantees, assumptions, constants, and formal gate are explicit | complete | Every constant is classified; formal warnings fail CI; protocol decisions are named |
-| 1. Trustworthy engine black box | Exact oracle, public subject boundary, full quiescence, replay capsules | in-progress | Known semantic mutations fail and captured failures replay |
+| 1. Trustworthy engine black box | Exact oracle, public subject boundary, full quiescence, replay capsules | complete | Known semantic mutations fail and captured failures replay |
```
erskingardner, line 118: Milestone 1 flips to `complete` here, and the exit gate below ticks "Targeted semantic mutations are detected" and "A captured byte-level failure replays with the same outcome".

The durable-outcome inventory added in this same diff sets the right standard — one named test per row. The mutation-sentinel paragraph names no tests for the six sentinels it claims (member identity, exporter commitment, ledger disposition, pending-work state, decryptability edge, byte delivery set); only the last has a test in this PR (`exact_captured_commit_replays_from_sensitive_checkpoint`). And the byte-replay gate is satisfied by a capsule built by hand in a test, not by anything the campaign emits.

Please either name the sentinel tests inline the way the table does, or leave those two boxes unchecked and Milestone 1 `in-progress`.

crates/cgka-conformance-simulator/src/report.rs
```diff
@@ -294,6 +351,40 @@ async fn run_vector_fixture_reports(
...
+        capture
+    }
+}
+
+fn write_report_failure_capsule(
+    report: &ScenarioReport,
+    capture: crate::ScenarioFailureCaptureV1,
+    path: PathBuf,
+) -> Result<PathBuf, Box<dyn Error>> {
+    let sensitivity = if capture.byte_replay.is_some() {
```
erskingardner: This flips the capsule to `sensitive_local` purely because a checkpoint happened to be captured, and capture is unconditional in both campaign entry points (`run_scenario_report_with_outcomes_and_capture` and `run_vector_fixture_report_with_capture` both call `enable_failure_replay_capture`). Two consequences:

- every failing campaign scenario now writes key material into the `--out` directory, which the PR description still describes as requiring an explicitly hand-constructed capsule;
- `promote_failure_capsule_to_vector` rejects `sensitive_local`, so no campaign-produced capsule can be promoted any more, even though promotion only reads `scenario`/expectations and never touches `byte_replay`.

Keeping the logical capsule `synthetic_shareable` and writing the checkpoint envelope to a separate sibling file (or gating capture behind an explicit flag) would fix both.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -498,6 +539,125 @@ impl EngineHarnessSubject {
...
+
+    pub fn byte_replay_capture(&self) -> Option<EngineByteReplayV1> {
+        self.last_byte_replay.clone()
+    }
+
+    pub fn enable_failure_replay_capture(&mut self) {
+        self.replay_capture_enabled = true;
+    }
+
+    fn prepare_byte_replay(&self, label: &str) -> Option<PendingByteReplay> {
```
erskingardner, line 586: This runs before *every* `Tick` for every ticked client and does a full `export_conformance_replay_snapshot` of the group, and `complete_byte_replay` then takes a canonical state snapshot and SHA-256s it — after which all but the last result is thrown away. Capture is on for every campaign case, passing or failing.

On the 21-client mixed storm that's a full per-group DB serialization plus a state digest per tick per client, for one retained witness. Capturing the checkpoint lazily (keep the cheap pre-tick state only, and materialize the export when the tick errors or when the run ends) would remove nearly all of it.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -1260,8 +1431,48 @@ fn outbound_sequence(outbound_id: &str) -> Result<u64, SubjectError> {
...
+        EngineError::InvalidTransition(_)
+        | EngineError::Other(_)
+        | EngineError::Peeler(_)
+        | EngineError::ForkedEpoch { .. }
+        | EngineError::UnknownPending => SubjectFailureCategory::Protocol,
+        EngineError::NotGroupAdmin { .. }
+        | EngineError::AdminCannotSelfRemove { .. }
+        | EngineError::AdminDepletion { .. }
+        | EngineError::LeaveAlreadyRequested { .. }
+        | EngineError::Serialize(_)
```
erskingardner: `EngineError::Serialize(_)` isn't a refusal the protocol expects — it's an internal encode/decode failure and belongs with `Backend`/`Storage` under `Resource` (or `Protocol`). Filing it as `ExpectedRefusal` means a serialization bug surfaced through a tick fingerprints as a benign expected refusal, which is the same class of mistake the old blanket `EnvironmentFailure` made.

`UnknownGroup`, `InvalidWelcome`, and `WelcomeAlreadyProcessed` are more debatable but arguably also not "expected refusals" when they come out of a tick rather than a caller-initiated operation.

crates/cgka-conformance-simulator/src/failure_capsule.rs
```diff
@@ -0,0 +1,760 @@
...
+        .map_err(|error| FailureCapsuleError::Replay(error.to_string()))?;
+    if replay.virtual_time_tick_enabled {
+        client.enable_virtual_time_tick();
+    }
+    for message in &replay.captured_deliveries {
+        client.inject_captured_transport(message.clone());
+    }
+    let outcomes = client.tick().await;
+    let actual_error = outcomes.iter().find_map(|outcome| outcome.as_ref().err());
+    let canonical_state = client.canonical_state_snapshot();
```
erskingardner: Still the panicking accessor, while `complete_byte_replay` now uses the new `try_canonical_state_snapshot`. `replay_engine_bytes` is the public entry point behind `--replay-capsule`, operating on a capsule read off disk, so a checkpoint that restores into a state the engine can't snapshot aborts the process instead of returning `FailureCapsuleError::Replay`. Use the `try_` variant here too.

crates/cgka-conformance-simulator/src/failure_capsule.rs
```diff
@@ -0,0 +1,760 @@
...
+}
+
+fn transport_messages_json_bytes(messages: &[TransportMessage]) -> u64 {
+    messages
+        .iter()
+        .map(|message| serde_json::to_vec(message).map_or(0, |bytes| bytes.len() as u64))
+        .sum()
+}
+
+fn bound_transport_capture(mut capture: CapturedTransportWindowV1) -> CapturedTransportWindowV1 {
```
erskingardner, line 735: Two notes:

1. This filters rather than truncates, so an oversized message in the middle of the reverse scan is skipped while older, smaller ones are kept. The result isn't the contiguous newest tail the doc and the `transport_objects_dropped` counter imply — it's "newest 256 that happen to fit", with a hole. A `while` pop-from-front matching `TransportBus::send`'s eviction loop would be both simpler and consistent.

2. The bus already bounds at emission time, so this pass only ever fires for externally constructed captures (the `from_report` path and the new oversized-capture test). That's fine as a validation, but it's worth a comment saying so, otherwise it reads as a second, subtly different bounding policy.

erskingardner (commented):
Re-review of `e6e678c`. All eight findings from the previous pass are genuinely addressed, and several of the fixes are better than what I suggested: `SubjectFailureCategory` threading engine-error provenance into the fingerprint classification is the right shape, `FailureIdentityV1` makes the minimizer actually able to delete a 24-message storm (and the test now proves it), the duplicated `original_scenario`/`canonical_scenario`/`failure_kind`/`failing_action_id` fields are gone along with their validator branch, `execute_scenario_step` fixes the mixed-indentation match, the transport tail is bounded with explicit observed/retained/dropped accounting, and the mutation sentinels now name a test each. Nice work.

The new byte-replay producer solves my "no automatic producer" complaint, but it introduces two problems that I think should be resolved before merge (posting as a comment because GitHub will not let me request changes on my own PR — treat items 1 and 2 as blocking).

**1. Campaign report directories now contain key material by default.**

`enable_failure_replay_capture` is turned on unconditionally in both `run_scenario_report_with_outcomes_and_capture` and `run_vector_fixture_report_with_capture`, so every generated-family and vector-fixture campaign records a recipient checkpoint, and `write_report_failure_capsule` promotes the capsule to `sensitive_local` whenever one exists. Any failing scenario in `run_report` therefore drops a SQLite/OpenMLS checkpoint containing key material into `--out`. The files are owner-only via `fs_private`, which satisfies the letter of the local-artifact-safety invariant, but a report directory is exactly the kind of thing a CI job tars up and uploads.

Note the PR description still says "Byte-level replay additionally requires an explicitly constructed `sensitive_local` capsule with the recipient checkpoint" — after this revision that is no longer true, and the change in default posture isn't called out anywhere. Suggest making checkpoint capture opt-in (a `--capture-byte-replay` flag), or writing the sensitive envelope to a separate, obviously-named sibling file so the shareable capsule stays shareable.

**2. As a side effect, campaign capsules are no longer promotable at all.**

`promote_failure_capsule_to_vector` rejects anything that isn't `synthetic_shareable`, and after (1) essentially every campaign capsule is `sensitive_local`. So the Milestone 1.4 box "Promote minimized synthetic failures into portable vectors" is now reachable only from a capsule assembled by hand in a test — the exact inverse of the gap I flagged last round, just moved from byte replay to promotion. Worth noting that the promoted `VectorFixture` reads only `scenario` + expectations and never touches `byte_replay`, so its output carries no key material regardless of capsule sensitivity; splitting the sensitive envelope out (per (1)) fixes both at once.

**3. The byte-replay exit gate got weaker in this revision, not stronger.**

`exact_captured_commit_replays_from_sensitive_checkpoint` previously asserted an `OracleViolation`/`captured_commit_state` fingerprint; it now asserts `Converged`/`campaign_tick_replay`, and the new `failed_campaign_capsule_contains_a_replayable_tick_witness` asserts the same. Since `last_byte_replay` is overwritten on every tick and the campaign's failure is an oracle mismatch detected later, the recorded witness is the last *converged* tick, unrelated to the failure the capsule exists to explain. Nothing in the suite now replays a captured failure, yet `convergence-reliability-plan.md:530` still ticks "A captured byte-level failure replays with the same outcome". Either record the tick that the failure attaches to, or reword the gate to what the code does ("a captured tick replays byte-identically").

Smaller things inline: per-tick full storage export for a checkpoint that's almost always discarded; `EngineError::Serialize` filed as `ExpectedRefusal`; `replay_engine_bytes` still panicking on the canonical snapshot; `bound_transport_capture` producing a non-contiguous tail; and the forensic tail being computed for every bus send in the whole simulator because `capture_outbound_for` is enabled for every client at attach, so `serde_json::to_vec` now runs per emission even in runs that will never build a capsule.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -736,7 +898,18 @@ impl ConvergenceSubject for EngineHarnessSubject {
 
     async fn tick(&mut self, clients: &[String]) -> Result<(), SubjectError> {
         for label in clients {
-            self.client_mut(label)?.tick().await;
+            let recipient_tick = self.observed_recipient_ticks;
```
erskingardner, line 901: Latent capture-target mismatch: `observed_recipient_ticks` increments on *every* `tick()` call, including the ones `drive_subject_to_quiescence` makes internally (quiescence.rs ticks all scenario clients per iteration), but `final_planned_recipient_tick` in scenario.rs computes the target purely from explicit `ScenarioStep::Tick` steps. For any spec where an `AwaitQuiescence` step drives ticks before the last explicit `Tick` step, the checkpoint is captured at an earlier tick than the documented "final planned recipient tick" — or not at all, if quiescence ticks push the counter past the target between explicit steps (equality never fires).

Nothing the report CLI runs today uses `AwaitQuiescence` steps (generated families end on explicit ticks; no vector contains `await_quiescence`), so this can't trigger in current campaigns — but a hand-written vector with `await_quiescence` run under `--capture-sensitive-replay` would silently capture the wrong witness. Consider resolving the target dynamically when the final `Tick` step begins (target = current counter + remaining clients - 1), gating the counter to explicit tick steps, or documenting the restriction where `capture_failure_replay_at_tick` is set.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -1357,6 +1574,25 @@ mod tests {
...
+            "converge buffered group: injected failure".into(),
+        ))])
+        .expect_err("tick must not discard the engine failure");
+        assert_eq!(error.code, "backend");
+        assert_eq!(error.category, SubjectFailureCategory::Resource);
+        assert!(error.message.contains("converge buffered group"));
+    }
+
+    #[test]
+    fn serialization_failures_are_not_classified_as_expected_refusals() {
```
erskingardner, line 1589: This test is about the category (good), but it also pins `code == "invalid_admin_policy"` for a generic `EngineError::Serialize("malformed internal state")`. That mapping in `observe_engine_error` is pre-existing and the admin-policy vector depends on it for the admin-policy path — but asserting it here cements the leak of an admin-policy-specific code onto arbitrary serialization failures, and the code now propagates into capsule fingerprints as `scenario_step_failed:invalid_admin_policy`. Either drop the code assertion (keep the category assertion), or give `Serialize` its own code and keep `invalid_admin_policy` produced by the policy-validation path.

crates/cgka-conformance-simulator/README.md
```diff
@@ -99,6 +103,20 @@ selected subject adapter, declared capabilities, storage backend, observed trace
...
 
+Failed report runs also write a portable `*-failure-capsule.v1.json`. The capsule writer uses owner-only directories and
+files. Sensitive SQLite/OpenMLS checkpoint capture is off by default because it contains key material and incurs a
+checkpoint export. Pass `--capture-sensitive-replay` to capture only the final planned recipient tick and write a
+separate `*-sensitive-replay-capsule.v1.json`; the portable logical capsule remains eligible for vector promotion.
+Replay the sensitive sibling without regenerating MLS bytes with:
+
+```sh
+cargo run -p cgka-conformance-simulator --bin cgka-conformance-simulator-report -- \
+  --replay-capsule /private/path/failure-capsule.v1.json
```
erskingardner, line 114: Nit: the example uses the portable capsule's filename, but `--replay-capsule` only works on the `*-sensitive-replay-capsule.v1.json` sibling (the portable capsule has no `byte_replay`, so the CLI errors with "does not contain an engine byte-replay checkpoint"). Using the sibling suffix in the example would save first-time users a confusing failure.

erskingardner (commented):
Solid, well-tested milestone work — capsule schema, bounded capture, sensitive replay flow, and the report/CLI integration are coherent, feature-gated where they should be, and covered by strong tests (schema round-trip, owner-only modes, bounded eviction, checkpoint replay with a missing-input fingerprint check, end-to-end campaign witness). Verified against the head source: bus capture is registered for every harness client at attach, replay restore/restart/hydrate is consistent, `fs_private` usage matches the repo's local-artifact-safety posture, and `conformance_constant_snapshot` stays behind `test-conformance-snapshot` with values/selection unchanged. CI is green at 322b126.

Three non-blocking notes inline: one latent capture-target mismatch worth resolving or explicitly documenting, one misleading error code cemented by a new test, and a README filename nit.

Two smaller remarks that don't fit inline:

- `crates/cgka-engine/AGENTS.md` describes `conformance_snapshot.rs` as owning canonical-state and pending-work projections only; it now also owns the machine-readable constant snapshot. A one-line update would keep the module map accurate (per repo convention).
- Scope: the mixed large-storm oracle strengthening + generator v5→v6 bump is arguably separable from the capsule work, but it's small, tested, and tied to the Milestone 1 exit gate, so bundling is defensible here.

Issue #1223, opened 2026-08-02
Agent integrations duplicate a 20-message durable-history window on every turn


Datawav:
## Summary

Both Marmot host integrations request the previous 20 durable messages before every activated agent turn. They have no route/session watermark or aggregate context budget, so continuing host sessions can receive the same supplemental Marmot window repeatedly even when the host may already retain some or all of it.

Verified by source inspection against `cdf0e40e4fcdf96bcee89512c753755615537e33`.

## Source

- Hermes performs `timeline_list(... limit=20)` and appends the result to `channel_context` on every activated message: [`adapter.py#L2858-L2898`](https://github.com/marmot-protocol/mdk/blob/cdf0e40e4fcdf96bcee89512c753755615537e33/integrations/hermes/marmot/adapter.py#L2858-L2898).
- OpenClaw performs the same per-dispatch lookup: [`dispatch.ts#L401-L447`](https://github.com/marmot-protocol/mdk/blob/cdf0e40e4fcdf96bcee89512c753755615537e33/integrations/openclaw/marmot/src/dispatch.ts#L401-L447).
- Timeline text is capped per record at 8,192 characters, allowing up to 163,840 characters of message text before metadata and serialization: [`agent-control/src/lib.rs#L13-L16`](https://github.com/marmot-protocol/mdk/blob/cdf0e40e4fcdf96bcee89512c753755615537e33/crates/agent-control/src/lib.rs#L13-L16).

PR #1159 intentionally introduced durable remote context and these integration paths. This follow-up preserves bootstrap/recovery context while making subsequent injection adaptive. The adapter/dispatch ownership paths contain no per-host-route session watermark and no total byte/token budget around these calls.

## Impact

- Repeated supplemental context can increase model input cost and accelerate host context compression.
- The fixed 20-record count does not bound UTF-8 frame bytes or model tokens.
- A long prior message can crowd out more relevant reply targets or recent context.
- Simply persisting a watermark across host-session resets could create the opposite failure: missing bootstrap history.

## Proposed direction

Track context state per account + group + host route/session when the host exposes a stable session lifecycle. Send full bootstrap context for a new/recovered session, then compact deltas for continuing turns. If the host cannot report retained history, retain a conservative fallback. Apply an observable aggregate byte/token budget and explicitly fetch missing reply targets when needed.

## Acceptance criteria

- [ ] New/recovered host sessions still receive enough durable context to answer coherently.
[...]

PR #1234, opened 2026-08-02
fix(hermes): bound timeline context size

Datawav:
## Summary

- retain at most the newest eight automatic Hermes timeline records
- enforce a 16 KiB UTF-8 budget over the complete rendered timeline context, including prefix, wrapper, records, and aggregate truncation metadata
- drop oldest records first, preserve retained chronological order, and omit an oversized remaining record rather than exceeding the budget
- report only aggregate omitted/oversized counts and keep referenced-message context behavior unchanged
- cover small, large, Unicode-oversized, chronological, aggregate-metadata, and referenced-message cases through the inbound assembly path

## Scope

Addresses the aggregate-context-budget portion of #1223. It deliberately does not implement the broader host-session watermark/delta design from that issue.

The control protocol's per-record and frame limits remain unchanged; this adds the missing Hermes adapter-local rendered-context bound.

## Deployment evidence

Aggregate-only measurements from one Hermes Marmot deployment motivated the bound; they are correlation, not proof that this path caused every slow turn:

- 205 persisted Marmot inputs: median 3,656 characters, p90 32,308, maximum 41,849; 52 exceeded 20,000
- a representative 20-record rendered history was 32,056 characters before adapter-local bounding

## Validation

- `python3 -m unittest discover -s integrations/hermes/marmot/tests` — 168 passed
- `python3 -m py_compile integrations/hermes/marmot/adapter.py integrations/hermes/marmot/tests/test_adapter.py`
- `git diff --check`
- added-line security scan — no findings

### DIFF
--- a/integrations/hermes/marmot/adapter.py
+++ b/integrations/hermes/marmot/adapter.py
@@ -452,16 +452,72 @@ def _referenced_channel_context(reply: Any) -> Optional[str]:
     return f"Marmot referenced-message context (untrusted): {json.dumps(fact, separators=(',', ':'))}"
 
 
-def _timeline_channel_context(messages: Any) -> Optional[str]:
-    if not isinstance(messages, list) or not messages:
-        return None
+TIMELINE_CONTEXT_MESSAGE_LIMIT = 8
+TIMELINE_CONTEXT_BYTE_LIMIT = 16 * 1024
+_TIMELINE_CONTEXT_PREFIX = "Marmot conversation history (untrusted): "
+
+
+def _timeline_message_exceeds_byte_limit(message: Any) -> bool:
     fact = {
         "type": "chat_window",
         "order": "chronological",
         "relation": "before_current_message",
-        "messages": messages,
+        "messages": [message],
+        "messages_truncated": True,
+        "omitted_message_count": 1,
     }
-    return f"Marmot conversation history (untrusted): {json.dumps(fact, separators=(',', ':'))}"
+    rendered = f"{_TIMELINE_CONTEXT_PREFIX}{json.dumps(fact, separators=(',', ':'))}"
+    return len(rendered.encode("utf-8")) > TIMELINE_CONTEXT_BYTE_LIMIT
+
+
+def _timeline_channel_context(messages: Any) -> Optional[str]:
+    if not isinstance(messages, list) or not messages:
+        return None
+    initially_omitted = (
+        messages[:-TIMELINE_CONTEXT_MESSAGE_LIMIT]
+        if len(messages) > TIMELINE_CONTEXT_MESSAGE_LIMIT
+        else []
+    )
+    omitted_message_count = len(initially_omitted)
+    oversized_message_count = sum(
+        1 for message in initially_omitted if _timeline_message_exceeds_byte_limit(message)
+    )
+    bounded_messages = messages[-TIMELINE_CONTEXT_MESSAGE_LIMIT:]
+
+    while True:
+        fact = {
+            "type": "chat_window",
+            "order": "chronological",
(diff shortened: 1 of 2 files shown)


erskingardner (approved):
Reviewed exact head dedf10d80bda427683b95ec333fee654b91dcb09.

The Hermes implementation is correct within this PR’s stated scope: it retains the newest records in chronological order, measures the complete rendered timeline string in UTF-8 bytes (including prefix and truncation metadata), drops oldest records first, and safely handles an individually oversized remaining record. The focused tests cover the important boundary and Unicode cases.

OpenClaw parity is still needed: `dispatch.ts` continues to pass all 20 timeline records through `inboundSupplemental` without an aggregate rendered-context budget. I do not consider that blocking for this explicitly Hermes-scoped incremental PR, but #1223 should remain open until the equivalent OpenClaw bound and the broader host-session watermark/delta behavior are implemented.

erskingardner:
You need to sign the commit as well.

integrations/hermes/marmot/adapter.py
```diff
@@ -452,16 +452,72 @@ def _referenced_channel_context(reply: Any) -> Optional[str]:
...
+            bounded_messages = bounded_messages[1:]
+            omitted_message_count += 1
+            continue
+
+        if bounded_messages:
+            if _timeline_message_exceeds_byte_limit(bounded_messages[0]):
+                oversized_message_count += 1
+            bounded_messages = []
+            omitted_message_count += 1
+            continue
```
erskingardner, line 516: Nit (non-blocking): these two drop branches are identical — `if len(bounded_messages) > 1` and `if bounded_messages` do exactly the same thing, and `bounded_messages[1:]` on a one-element list already yields `[]`. A single `if bounded_messages:` block would express the same logic in half the lines.

erskingardner (commented):
Reviewed at head `dedf10d80bda427683b95ec333fee654b91dcb09` (confirmed current at posting time).

**The code is sound; recommend merge once the commit is signed.** The core safety property holds by construction: the 16 KiB budget is re-measured against the actual rendered string on every iteration, so the bound is enforced regardless of how the oversized classifier labels any individual record — classification only steers the aggregate counters. Oldest-first dropping preserves chronological order of retained records, and the ≤8-message path renders byte-identical output to the previous code (same key order, separators, prefix), so small-history behavior and the referenced-message context are untouched. Aggregate-only omitted/oversized counts align with the repo observability/privacy guidance. Tests hit the right boundaries — exact-fit, Unicode, count-window pre-scan, metadata-displacement classification, and the 20-record inbound path — and the CI job running this unittest suite is green.

The declared scope (Hermes-only aggregate bound; no watermark/delta; no OpenClaw parity) is honest and matches the incremental acceptance bullets in #1223. Agree with the prior review that #1223 stays open for the OpenClaw bound and the host-session watermark work.

Two notes:

1. **Blocking (process, already flagged):** head commit `dedf10d` is unsigned (`verification.reason: "unsigned"`) and the active "Safe Master" ruleset includes `required_signatures`; merge state is BLOCKED. The signed re-push will change the SHA — if the diff is otherwise unchanged, this assessment carries over.
2. **Nit (inline):** the loop’s two drop branches are identical and collapse into one.

Non-blocking observation, no change requested: `_timeline_message_exceeds_byte_limit` judges records against a reference envelope that already carries truncation metadata, so a record within ~60 bytes of the budget can be reported as `oversized` even though it would fit in a bare envelope. That is the conservative direction for host-facing aggregate metadata, and the real budget is always enforced on the actual render, so it is fine as-is — a one-line comment on the helper stating the reference-envelope choice would aid future readers.

PR #1235, opened 2026-08-03
Add Milestone 3 adversarial convergence campaigns

erskingardner:
## Summary

Completes Milestone 3 of the convergence reliability plan.

- adds a deterministic twelve-family adversarial campaign catalog covering offline retained-history floods, sustained mixed traffic, self-update/admin races, losing-branch admissions, unequal relay histories, durable-phase crashes, multi-group fairness, resource pressure, multi-device accounts, witness A/B behavior, mixed-binary compatibility preflight, and clock/cursor attacks
- adds embedded campaign measurements plus an isolated child-process runner for CPU, RSS, page-cache-sensitive filesystem block-write lower bounds, database footprint, queue, replay, latency, pass, reorg, recovery, and disposition accounting
- adds feature-gated fixed-input policy sweeps without permitting production auto-tuning
- adds producer-attested normalized incident artifacts while retaining explicitly inexact outcome-equivalent archetypes when source evidence is incomplete
- updates the convergence reliability tracking plan and documents the v1 policy compatibility decision

## Behavior and protocol impact

This does not add a v1 wire policy field or Marmot app component. Mixed convergence policy versions fail scenario preflight; mixed binary versions remain valid when policy and capabilities are equivalent. A future incompatible policy must introduce an authenticated required capability or GroupContext application component with the behavior change.

Production convergence constants and the adopted branch-selection policy are not auto-tuned or numerically changed. There is one intentional production fail-closed behavior change: application-bearing candidate re-materialization now consumes the same finite per-pass replay budget as candidate search instead of using an unlimited fallback budget. A near-ceiling pass can therefore return `ReplayBudgetExceeded` where the old path continued without a limit. The regression proves frozen inputs remain durable and succeed when capacity is restored.

The remaining engine changes are observability and deterministic test seams. The collecting-phase crash hook now fires only after the retained input and pass transaction commits, so every named durable phase represents an actually durable boundary.

Legacy client-only scenarios retain their historical client-derived member identities on both engine and retained-relay subjects. Account-scoped credential sharing is enabled only by explicitly authored account/device topology.

The losing-branch admission campaign records the current fail-closed outcome: branch-only devices remain non-equivalent after canonical members settle, and reusing an already-consumed device signature key is rejected. Repair requires an explicit reset/rejoin with fresh device state.

## Incident evidence trust boundary

Normalized history is labeled `producer_attested_normalized_history`, not exact source-verified history. The importer validates IR, mapping coverage and bounds, and simulator reproduction, but cannot independently prove that an arbitrary source event semantically corresponds to the declared action. Imported free-form labels and application payloads are marked `confidential_unredacted_scenario`; artifacts remain `unverified` until reproduction succeeds and are not shareable without separate redaction and review. Infrastructure failures exit 2 rather than being reported as successful quarantine.

## Validation

- `just fast-ci`
- Scenario IR suite: 15 passed
- feature-gated Milestone 3 suite: 10 passed, 3 intentionally ignored sustained campaigns
- policy sweeps: 2 passed
- distributed convergence suite with test policy/crash features: 79 passed
- crash-recovery SQLite matrix: 3 passed, 1 child helper ignored
- full incident-replay suite
- real file-backed child-process campaign smoke with deadline/reap/resource report
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added process-isolated conformance campaigns with adversarial workloads, resource measurements, policy sweeps, and aggregate reports.
  * Added multi-group scenarios, grouped actions, detailed step timing, and expanded regression coverage.
  * Added versioned incident replay artifacts with exact-history validation, fidelity reporting, and sensitive-data safeguards.
* **Bug Fixes**
  * Improved resource-refusal recovery, convergence tracking, replay handling, and durable crash recovery.
* **Documentation**
  * Expanded simulator, scenario, incident replay, and reliability documentation with privacy guidance.
* **Chores**
  * Added automated Milestone 3 campaign checks to continuous integration.

### DIFF
--- a/crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
+++ b/crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
@@ -38,7 +38,7 @@ struct ProcessCaseMeasurementV1 {
     system_cpu_us: Option<u64>,
     peak_rss_bytes: Option<u64>,
     database_bytes: Option<u64>,
-    database_write_bytes: Option<u64>,
+    filesystem_block_write_lower_bound_bytes: Option<u64>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     unavailable_process_fields: Vec<String>,
 }
@@ -92,7 +92,8 @@ async fn run() -> Result<ExitCode, Box<dyn Error>> {
             system_cpu_us: usage.system_cpu_us,
             peak_rss_bytes: usage.peak_rss_bytes,
             database_bytes,
-            database_write_bytes: usage.database_write_bytes,
+            filesystem_block_write_lower_bound_bytes: usage
+                .filesystem_block_write_lower_bound_bytes,
             unavailable_process_fields: usage.unavailable_process_fields,
         });
     }
@@ -202,7 +203,7 @@ struct ChildUsage {
     user_cpu_us: Option<u64>,
     system_cpu_us: Option<u64>,
     peak_rss_bytes: Option<u64>,
-    database_write_bytes: Option<u64>,
+    filesystem_block_write_lower_bound_bytes: Option<u64>,
     unavailable_process_fields: Vec<String>,
 }
 
@@ -222,15 +223,37 @@ fn wait_with_usage(
         // pointers refer to writable storage for the duration of wait4.
         let waited = unsafe { libc::wait4(pid, &mut status, libc::WNOHANG, usage.as_mut_ptr()) };
         if waited < 0 {
-            return Err(std::io::Error::last_os_error().into());
+            let error = std::io::Error::last_os_error();
+            let _ = child.kill();
+            let _ = child.wait();
+            return Err(error.into());
         }
         if waited == pid {
             break false;
         }
         if Instant::now() >= deadline {
-            if let Err(error) = child.kill()
-                && error.kind() != std::io::ErrorKind::InvalidInput
-            {
+            // Close the polling race: the worker may have exited successfully
+            // after the check above but before the deadline comparison.
+            let waited =
+                unsafe { libc::wait4(pid, &mut status, libc::WNOHANG, usage.as_mut_ptr()) };
+            if waited < 0 {
+                let error = std::io::Error::last_os_error();
+                let _ = child.kill();
+                let _ = child.wait();
+                return Err(error.into());
+            }
+            if waited == pid {
+                break false;
+            }
+            if let Err(error) = child.kill() {
+                if error.kind() == std::io::ErrorKind::InvalidInput {
+                    let waited = unsafe { libc::wait4(pid, &mut status, 0, usage.as_mut_ptr()) };
+                    if waited < 0 {
+                        return Err(std::io::Error::last_os_error().into());
+                    }
+                    break false;
+                }
+                let _ = child.wait();
                 return Err(error.into());
             }
             // SAFETY: after killing the same direct child, blocking wait4 reaps
@@ -246,10 +269,14 @@ fn wait_with_usage(
     // SAFETY: either successful wait4 path initialized the complete rusage value.
     let usage = unsafe { usage.assume_init() };
     #[cfg(any(target_os = "macos", target_os = "ios"))]
-    let (database_write_bytes, unavailable_process_fields) =
-        (None, vec!["database_write_bytes".to_owned()]);
+    let (filesystem_block_write_lower_bound_bytes, unavailable_process_fields) = (
+        None,
+        vec!["filesystem_block_write_lower_bound_bytes".to_owned()],
+    );
     #[cfg(not(any(target_os = "macos", target_os = "ios")))]
-    let (database_write_bytes, unavailable_process_fields) = (
+    // `ru_oublock` counts real block-device output, so page-cache absorption
+    // can only make this a lower bound for all child-process filesystem writes.
+    let (filesystem_block_write_lower_bound_bytes, unavailable_process_fields) = (
         Some(
             u64::try_from(usage.ru_oublock)
                 .unwrap_or(0)
@@ -264,7 +291,7 @@ fn wait_with_usage(
         user_cpu_us: Some(timeval_us(usage.ru_utime)),
         system_cpu_us: Some(timeval_us(usage.ru_stime)),
         peak_rss_bytes: Some(peak_rss_bytes(usage.ru_maxrss)),
-        database_write_bytes,
+        filesystem_block_write_lower_bound_bytes,
         unavailable_process_fields,
     })
 }
@@ -316,12 +343,12 @@ fn wait_with_usage(
         user_cpu_us: None,
         system_cpu_us: None,
         peak_rss_bytes: None,
-        database_write_bytes: None,
+        filesystem_block_write_lower_bound_bytes: None,
         unavailable_process_fields: vec![
             "user_cpu_us".to_owned(),
             "system_cpu_us".to_owned(),
             "peak_rss_bytes".to_owned(),
-            "database_write_bytes".to_owned(),
+            "filesystem_block_write_lower_bound_bytes".to_owned(),
         ],
     })
 }
--- a/crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
+++ b/crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
@@ -1,0 +1,247 @@
+use cgka_conformance_simulator::{
+    HarnessStorageMode, ScenarioReport, generate_milestone3_adversarial_family,
+    run_generated_case_report_with_storage_mode,
+};
+use serde::{Deserialize, Serialize};
+use std::error::Error;
+use std::path::{Path, PathBuf};
+use std::process::{Command, ExitCode};
+use std::time::Instant;
+
+#[derive(Debug)]
+struct Args {
+    seed: u64,
+    cases: usize,
+    case_index: Option<usize>,
+    out: PathBuf,
+    storage: HarnessStorageMode,
+    worker: bool,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+struct ProcessCampaignReportV1 {
+    schema_version: String,
+    seed: u64,
+    cases: Vec<ProcessCaseMeasurementV1>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+struct ProcessCaseMeasurementV1 {
+    case_index: usize,
+    report: PathBuf,
+    exit_code: Option<i32>,
+    signal: Option<i32>,
+    wall_us: u64,
+    user_cpu_us: u64,
+    system_cpu_us: u64,
+    peak_rss_bytes: u64,
+    database_bytes: Option<u64>,
+    database_write_bytes: u64,
+}
+
+#[tokio::main]
+async fn main() -> ExitCode {
+    match run().await {
+        Ok(code) => code,
+        Err(error) => {
+            eprintln!("cgka campaign: {error}");
+            ExitCode::FAILURE
+        }
+    }
+}
+
+async fn run() -> Result<ExitCode, Box<dyn Error>> {
+    let args = parse_args(std::env::args().skip(1))?;
+    if args.worker {
+        return run_worker(&args).await;
+    }
+    fs_private::create_dir_all_private(&args.out)?;
+
+    let executable = std::env::current_exe()?;
+    let mut observations = Vec::with_capacity(args.cases);
+    for case_index in 0..args.cases {
+        let report = args.out.join(format!("case-{case_index}.json"));
+        let mut command = Command::new(&executable);
+        command.args([
+            "--worker",
+            "--seed",
+            &args.seed.to_string(),
+            "--case-index",
+            &case_index.to_string(),
+            "--out",
+            report.to_str().ok_or("non-UTF-8 report path")?,
+            "--storage",
+            storage_label(args.storage),
+        ]);
+        let started = Instant::now();
+        let child = command.spawn()?;
+        let usage = wait_with_usage(child)?;
+        let database_bytes = read_database_bytes(&report);
+        observations.push(ProcessCaseMeasurementV1 {
+            case_index,
+            report,
+            exit_code: usage.exit_code,
+            signal: usage.signal,
+            wall_us: elapsed_us(started.elapsed()),
+            user_cpu_us: usage.user_cpu_us,
+            system_cpu_us: usage.system_cpu_us,
+            peak_rss_bytes: usage.peak_rss_bytes,
+            database_bytes,
+            database_write_bytes: usage.database_write_bytes,
+        });
+    }
+    let failed = observations
+        .iter()
+        .any(|case| case.exit_code != Some(0) || case.signal.is_some());
+    let summary = ProcessCampaignReportV1 {
+        schema_version: "1".into(),
+        seed: args.seed,
+        cases: observations,
+    };
+    let summary_path = args.out.join("process-campaign.v1.json");
+    fs_private::write_private(&summary_path, &serde_json::to_vec_pretty(&summary)?)?;
+    println!("Process campaign report: {}", summary_path.display());
+    Ok(if failed {
+        ExitCode::FAILURE
+    } else {
+        ExitCode::SUCCESS
+    })
+}
+
+async fn run_worker(args: &Args) -> Result<ExitCode, Box<dyn Error>> {
+    let case_index = args.case_index.ok_or("worker requires --case-index")?;
+    let cases = generate_milestone3_adversarial_family(args.seed, case_index + 1);
+    let case = &cases[case_index];
+    let report = run_generated_case_report_with_storage_mode(case, None, args.storage).await?;
+    if let Some(parent) = args.out.parent() {
+        fs_private::create_dir_all_private(parent)?;
+    }
+    fs_private::write_private(&args.out, &serde_json::to_vec_pretty(&report)?)?;
+    let failed = !report.expectation_failures.is_empty() || !report.invariant_failures.is_empty();
+    Ok(if failed {
+        ExitCode::FAILURE
+    } else {
+        ExitCode::SUCCESS
+    })
+}
+
+fn read_database_bytes(path: &Path) -> Option<u64> {
+    let bytes = std::fs::read(path).ok()?;
+    serde_json::from_slice::<ScenarioReport>(&bytes)
+        .ok()?
+        .campaign_measurements
+        .database_bytes
+}
+
+fn parse_args(args: impl Iterator<Item = String>) -> Result<Args, Box<dyn Error>> {
+    let mut seed = 7;
+    let mut cases = 12;
+    let mut case_index = None;
+    let mut out = PathBuf::from("target/cgka-milestone3-process-campaign");
+    let mut storage = HarnessStorageMode::TempFileBackedSqlite;
+    let mut worker = false;
+    let mut args = args.peekable();
+    while let Some(arg) = args.next() {
+        match arg.as_str() {
+            "--worker" => worker = true,
+            "--seed" => seed = args.next().ok_or("missing --seed value")?.parse()?,
+            "--cases" => cases = args.next().ok_or("missing --cases value")?.parse()?,
+            "--case-index" => {
+                case_index = Some(args.next().ok_or("missing --case-index value")?.parse()?)
+            }
+            "--out" => out = PathBuf::from(args.next().ok_or("missing --out value")?),
+            "--storage" => {
+                storage = match args.next().ok_or("missing --storage value")?.as_str() {
+                    "memory" => HarnessStorageMode::InMemorySqlite,
+                    "file" => HarnessStorageMode::TempFileBackedSqlite,
+                    other => return Err(format!("invalid storage mode {other}").into()),
+                }
+            }
+            other => return Err(format!("unknown argument {other}").into()),
+        }
+    }
+    Ok(Args {
+        seed,
+        cases,
+        case_index,
+        out,
+        storage,
+        worker,
+    })
+}
+
+fn storage_label(storage: HarnessStorageMode) -> &'static str {
+    match storage {
+        HarnessStorageMode::InMemorySqlite => "memory",
+        HarnessStorageMode::TempFileBackedSqlite => "file",
+    }
+}
+
+fn elapsed_us(duration: std::time::Duration) -> u64 {
+    u64::try_from(duration.as_micros()).unwrap_or(u64::MAX)
+}
+
+struct ChildUsage {
+    exit_code: Option<i32>,
+    signal: Option<i32>,
+    user_cpu_us: u64,
+    system_cpu_us: u64,
+    peak_rss_bytes: u64,
+    database_write_bytes: u64,
+}
+
+#[cfg(unix)]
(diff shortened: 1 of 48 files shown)


crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -1060,20 +1182,74 @@ impl ConvergenceSubject for EngineHarnessSubject {
...
+            .values()
+            .filter_map(|client| client.database_path())
+            .collect::<Vec<_>>();
+        (!paths.is_empty()).then(|| {
+            paths
+                .into_iter()
+                .flat_map(|path| {
+                    [
+                        path.to_path_buf(),
+                        path.with_extension("sqlite-wal"),
```
erskingardner: `with_extension` **replaces** the extension: `client.sqlite3` becomes `client.sqlite-wal`, not `client.sqlite3-wal`. SQLite appends `-wal` / `-shm` to the full filename, so both of these `metadata()` lookups always miss and get dropped by `filter_map`. With `SqliteJournalMode::Wal` as the default, the reported `database_bytes` is main-file-only.

```rust
let sidecar = |suffix: &str| {
    let mut name = path.as_os_str().to_os_string();
    name.push(suffix);
    PathBuf::from(name)
};
[path.to_path_buf(), sidecar("-wal"), sidecar("-shm")]
```

This is the metric `convergence-reliability-plan.md` describes as "file-backed SQLite/WAL/SHM size", so the doc claim is currently untrue.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -1060,20 +1182,74 @@ impl ConvergenceSubject for EngineHarnessSubject {
...
+
+    fn engine_metrics(&self) -> Option<EngineMetricsSnapshot> {
+        let mut aggregate = EngineMetricsSnapshot::default();
+        for client in self.clients.values() {
+            crate::client::merge_engine_metrics(&mut aggregate, &client.engine_metrics());
+        }
+        Some(aggregate)
+    }
+
+    fn select_scenario_group(&mut self, group: &str) -> Result<(), SubjectError> {
```
erskingardner: When `scenario_groups` has no entry for `group` yet, this records the label and returns `Ok(())` without touching any client's default group — so the step executes against whatever group was selected previously. That is silent misrouting rather than a failure.

The compiler only rejects unknown groups when `topology.groups` is non-empty (`lower_group_action`), and every `milestone3_*` scenario leaves `groups: vec![]`, so nothing catches an `in_group` reference to a group that has not been created. Either error here for any non-`CreateGroup` action against an unknown label, or make `lower_group_action` require declared groups.

crates/cgka-conformance-simulator/src/campaign_metrics.rs
```diff
@@ -0,0 +1,200 @@
...
+                        message,
+                    } => Some(format!("{kind}: {message}")),
+                    _ => None,
+                })
+            });
+
+        Self {
+            schema_version: "1".into(),
+            total_wall_us,
+            convergence_latency_us: total_wall_us,
```
erskingardner: `convergence_latency_us` is `total_wall_us`, always. Two fields, one value, and the narrower name will be read as "time to converge" by whoever consumes these reports. Either drop the field until there is a real per-convergence measurement, or compute it from the settle observations.

crates/cgka-conformance-simulator/src/campaign_metrics.rs
```diff
@@ -0,0 +1,200 @@
...
+            if let Some(pending) = &observation.pending_work {
+                max_observed_queue_depth = max_observed_queue_depth.max(
+                    pending
+                        .bus_queued_messages
+                        .saturating_add(pending.bus_delayed_messages)
+                        .saturating_add(pending.bus_mailbox_messages),
+                );
+            }
+            for decision in &observation.convergence_decisions {
+                let _ = decision;
```
erskingardner: `for decision in ... { let _ = decision; pass_count += 1 }` is a hand-rolled `len()`:

```rust
pass_count = pass_count.saturating_add(observation.convergence_decisions.len());
```

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -825,15 +857,27 @@ fn canonicalize_openmls_batch_with_materialized(
...
+    if !admit_app_witnesses {
+        return Ok(
+            crate::canonicalization::canonicalize_with_materialized_candidates_for_test(
+                input,
+                materialized_candidates,
+                false,
+            ),
+        );
+    }
+    let _ = admit_app_witnesses;
```
erskingardner: Without `test-policy-overrides`, `admit_app_witnesses: false` is silently discarded here and witnesses are admitted anyway. `Engine::admit_app_witnesses` is a plain (non-cfg'd) field, so a build that sets it false without the feature — e.g. a simulator whose feature forwarding is dropped in a refactor — would run the witness A/B campaign as two identical configurations and report "the witness term changes nothing", which is exactly the question this milestone is trying to answer.

A `debug_assert!(admit_app_witnesses, "witness admission cannot be disabled without test-policy-overrides")` in place of `let _ = ...` makes that a loud failure instead of a false negative.

crates/cgka-engine/src/engine.rs
```diff
@@ -242,6 +242,14 @@ pub struct Engine<S: StorageProvider> {
     pub(crate) queued_intent_by_pending: HashMap<PendingStateRef, (GroupId, MessageId)>,
 
     pub(crate) convergence_policy: crate::canonicalization::CanonicalizationPolicy,
+    /// Test-only selection experiment switch. It is always true in production;
+    /// only the feature-gated builder method can disable witness admission.
+    pub(crate) admit_app_witnesses: bool,
```
erskingardner, line 248: `admit_app_witnesses` and `replay_probe_budget_override` are unconditional fields on the production `Engine` and `EngineBuilder`, while the sibling `conformance_replay_probe_count` three lines down *is* `#[cfg]`-gated. Inconsistent, and it means a production engine carries a live switch that can turn off a convergence-policy term — only the setters are gated, not the state.

If they were `#[cfg(feature = "test-policy-overrides")]` too, the silent-ignore path in `openmls_projection.rs` becomes unrepresentable rather than merely unreachable-by-convention.

crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+    // SAFETY: successful wait4 initialized the complete rusage value.
+    let usage = unsafe { usage.assume_init() };
+    Ok(ChildUsage {
+        exit_code: libc::WIFEXITED(status).then(|| libc::WEXITSTATUS(status)),
+        signal: libc::WIFSIGNALED(status).then(|| libc::WTERMSIG(status)),
+        user_cpu_us: timeval_us(usage.ru_utime),
+        system_cpu_us: timeval_us(usage.ru_stime),
+        peak_rss_bytes: peak_rss_bytes(usage.ru_maxrss),
+        database_write_bytes: u64::try_from(usage.ru_oublock)
```
erskingardner: `ru_oublock` is not maintained by Darwin's `getrusage` — it is always 0 on macOS, unconditionally, not just "under caching" as the plan doc puts it. On Linux it counts block-device output, which also will not track SQLite logical writes once the page cache absorbs them.

Suggest gating this to `target_os = "linux"` and pushing `database_write_bytes` into the unavailable-fields list elsewhere, so a 0 is not read as "no writes".

crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+#[cfg(unix)]
+fn wait_with_usage(child: std::process::Child) -> Result<ChildUsage, Box<dyn Error>> {
+    let pid = i32::try_from(child.id())?;
+    let mut status = 0;
+    let mut usage = std::mem::MaybeUninit::<libc::rusage>::zeroed();
+    // SAFETY: `pid` is the live direct child just spawned above; both output
+    // pointers refer to initialized writable storage for the duration of wait4.
+    let waited = unsafe { libc::wait4(pid, &mut status, 0, usage.as_mut_ptr()) };
```
erskingardner: `wait4` with no timeout: a worker that wedges (a campaign that fails to reach quiescence is precisely the failure mode being hunted) hangs the whole runner with no output. A deadline plus `kill` would make the `signal` field you already record actually useful.

crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
```diff
@@ -0,0 +1,247 @@
...
+    Ok(if failed {
+        ExitCode::FAILURE
+    } else {
+        ExitCode::SUCCESS
+    })
+}
+
+async fn run_worker(args: &Args) -> Result<ExitCode, Box<dyn Error>> {
+    let case_index = args.case_index.ok_or("worker requires --case-index")?;
+    let cases = generate_milestone3_adversarial_family(args.seed, case_index + 1);
```
erskingardner: The worker generates cases `0..=case_index` and throws away all but the last, so a full campaign does O(n^2) scenario generation. Since each case is seeded independently (`seed ^ K ^ case_index.rotate_left(17)`), a single-case entry point would be both cheaper and clearer about that independence.

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -623,49 +623,100 @@ Status: `complete`
...
+indices, identifies the contested fork/convergence source rows, includes semantic expected outcomes, passes full IR
+validation, contains no sensitive state, and reproduces under the simulator. “Exact” means exact for that declared
+normalized evidence, not byte replay. Raw MLS bytes plus an engine checkpoint remain a separate opt-in sensitive local
+failure capsule.
+
+The CLI writes the vector and evidence envelope owner-only and never copies the source export or checkpoint. Real
+exports belong under ignored `incident-exports/`; local generated output belongs under `target/` or ignored
+`incident-replay-output/`; sensitive replay-capsule filenames are ignored repository-wide.
 
 ### Milestone 3 Exit Gate
```
erskingardner, line 704: This exit gate — and the `[x]` on all twelve workload families and all seven resource measurements — is checked off against validation that is manual-only. As noted in the summary: seven families never execute in any automated run, `policy_sweeps.rs` and the witness/replay-budget tests are behind a feature no CI job enables for this crate, three campaigns are `#[ignore]`d, and the process runner is never invoked.

"The three headline workloads run both as small regressions and sustained campaigns" is true of the regressions and true of the campaigns only when a human types the command. Either add the gate or state the gap here, so the next reader of this table does not inherit a false sense of coverage.

erskingardner (commented):
**Review verdict: changes requested** (posted as a comment — GitHub does not allow request-changes on your own PR).

Milestone 3 is a lot of well-organized machinery, and the fail-closed framing (archetype vs. exact replay, no production auto-tuning, no v1 policy wire field) is the right call. Two things should be fixed before this lands as "complete", plus a handful of smaller cleanups.

**Blocking**

1. `database_bytes()` never counts WAL/SHM. `Path::with_extension("sqlite-wal")` on `client.sqlite3` produces `client.sqlite-wal`, but SQLite appends the suffix: the real files are `client.sqlite3-wal` / `client.sqlite3-shm`. Both `metadata()` calls always fail and get filtered out. Since `SqliteJournalMode::Wal` is the default, this under-reports exactly the component the plan doc claims to measure ("file-backed SQLite/WAL/SHM size").

2. Nothing automated runs most of this. Of the twelve catalog families, only 0/1/2 (as small regressions), 3, and 10 (compile-only) execute under `cargo nextest run -p cgka-conformance-simulator`, which is what the "Simulator and vectors" job runs with no features. Families 4, 5, 6, 7, 8, 9, and 11 are only `compile_scenario`-checked. `tests/policy_sweeps.rs` is `#![cfg(feature = "test-policy-overrides")]`, and the witness A/B and replay-budget tests are gated the same way — no CI job and no justfile recipe enables that feature for this crate (`ci.yml` enables `wn-cli/...` and `cgka-engine/test-crash-hooks` only, and explicitly `--exclude`s this crate from that matrix). The `cgka-conformance-campaign` binary is never invoked anywhere. Meanwhile the plan doc flips every Milestone 3 box, including the exit gate, to `[x]`. Either add a gate (a `just` recipe plus a CI step running the feature-gated suite, and something that executes the un-run families) or scope the doc claims to what is actually enforced — otherwise this rots silently and the next reader believes it is covered.

**Worth addressing**

3. `database_write_bytes` derived from `ru_oublock` is unconditionally 0 on Darwin — `getrusage` there does not populate block I/O counters. The doc hedges it as "may be zero under caching", which is not the same thing; it is structurally unavailable on the platform most of this repo is developed on. Report it as unavailable on non-Linux the way `unavailable_process_fields` already does elsewhere.

4. `convergence_latency_us` is assigned `total_wall_us` verbatim. A second field that is always equal to the first, under a name that implies something narrower, is worse than not having the field.

Other per-line notes inline. Nothing here questions the premise — a deterministic adversarial campaign catalog is clearly worth having; the concern is that the verification story does not yet match the completion claim.

erskingardner:
Addressed all review threads in 0bf77890. Highlights:

- fixed SQLite WAL/SHM accounting and made convergence/process metrics honest about unavailable values
- added a hard per-worker timeout, single-case generation, and platform-correct process resource reporting
- made multi-group routing fail closed and atomic, restricted nested/group-scoped IR, and enforced pairwise non-equivalence
- shared the replay-probe ceiling across BFS and fallback materialization; test-only engine controls are feature-gated
- hardened incident artifacts against missing evidence and path traversal
- added the full Milestone 3 catalog, sustained workloads, policy tests, and isolated runner to CI

Local verification: `just fast-ci`; feature-gated Milestone 3 suite; all three ignored sustained campaigns; policy sweeps; isolated file-backed worker; simulator library; Scenario IR; and full incident-replay tests. All addressed threads are resolved; fresh CI is running on this head.

erskingardner (commented):
# Review — Milestone 3 adversarial convergence campaigns

Reviewed the full diff at 0bf7789 area-by-area (engine, simulator core, campaign infra/CI, incident-replay); the highest-severity findings below were verified directly against the source, live CI, and the GitHub runner image manifest.

**Overall:** the engineering discipline is genuinely good — feature gating is airtight, campaign generation is properly seeded and deterministic, the child-process runner measures real child rusage with correct reaping, and files are written via `fs-private` as the repo requires. But there is one blocking CI defect, two contract-level simulator bugs that undermine what the campaigns claim to test, and the description's two headline claims ("no production behavior change" and "sensitive-data safeguards / exact-history validation") are both overstated.

## Blocking

**1. The new CI gate cannot run: `just` is not installed on the runner** — `.github/workflows/ci.yml:317` runs `just milestone3-ci`, but no step installs `just` and GitHub's ubuntu-24.04 image does not ship it (verified against the current `actions/runner-images` Ubuntu2404 manifest). Every conformance-job run will die with `command not found`, so none of the milestone3 campaigns, the three sustained workloads, or the policy sweeps the plan doc says CI now runs will ever execute. The step was added in the final commit and the first job to reach it is still in flight, so this hasn't surfaced yet. Fix: add `extractions/setup-just` / `taiki-e/install-action@just`, or inline the four cargo commands. While there, the `milestone3-ci` recipe (`Justfile:253-257`) should pass `--locked` like every other CI cargo invocation.

## Major

**2. Compiled action ids and runtime action ids disagree for `in_group` steps** — `stable_action_id` produces `step-N:kind@group` for a wrapper step (`scenario_ir.rs:59-65`), and that goes into the compiled schedule. But the compiled action stores the *lowered* inner step (`scenario_ir.rs:120`), and the runtime recomputes the id from that lowered step (`scenario.rs:712`, `1217`), yielding `step-N:kind` with no suffix. Any fault selector, failure-capsule schedule entry, or ledger join built with the documented `stable_action_id` on a wrapped step silently matches nothing — e.g. a `WithholdMessage` targeting a wrapped send withholds nothing.

**3. `ClientsNotEquivalent` expectations can never pass the strict oracle** — the expectation maps to expected behavior `ExactStateEquivalence` (`oracle.rs:622-624`), but observed behavior only records `ExactStateEquivalence` when *all* canonical snapshots are equal (`oracle.rs:500-506`). A scenario that correctly observes divergence (the point of the new losing-branch family) therefore reports `missing_observed_behavior` and fails under the default strict-oracle CLI. The in-repo tests dodge it by calling `run_generated_case_report` directly, which skips oracle strictness — so the catalog passes in tests but the documented CLI invocation fails.

**4. Incident-artifact "sensitive-data safeguards" are an honor-system flag** — the shareable exact-history artifact embeds the producer-supplied Scenario IR verbatim (`artifact.rs:146-166`): client labels, group names, and `SendAppMessage { payload }` are free-form strings, and the only safeguard is the producer's self-declared `sensitive_state_included: bool`. An export whose labels are real pubkeys and whose payloads are real plaintext passes every check and gets written into the artifact the docs describe as shareable. Nothing strips, allowlists, or length-caps these fields. Relatedly, error paths echo export-controlled strings (versions, client names, scenario names) to stdout (`main.rs:178-181`, `204-207`).

**5. "Exact" fidelity is producer-attested, not verified** — `validate_action_evidence` (`artifact.rs:226-268`) checks index bookkeeping (coverage, bounds, ≥1 incident index) but never that a mapped source event corresponds to its scenario step, and the producer also supplies the `expected_outcomes` the simulator is checked against. An artifact can map every step to one unrelated event and still ship stamped `ExactNormalizedHistory`. Also: `import_exact_history` stamps the artifact "exact" *before* reproduction runs, so a library consumer that skips `accept_exact_history` persists an unverified artifact; and a tokio/simulator infrastructure failure is routed through `quarantine()` with exit 0 rather than the documented exit 2 (`artifact.rs:212-218`, `main.rs:123-125`).

**6. Undisclosed production behavior change in the replay budget** — the app-bearing re-materialization path now draws from the shared finite per-pass `ReplayBudget` (`openmls_projection.rs:1211-1266`) where at the merge base it used `ReplayBudget::unlimited()`. Under near-ceiling adversarial branching plus in-flight app messages, a pass that previously settled now fails closed with `ReplayBudgetExceeded`. This is defensible hardening (the unlimited re-materialization was itself an amplification loophole), but the PR states production convergence behavior is unchanged — the changed error path should be acknowledged in the description and covered by a test of the shared-budget interaction. Everything else engine-side verified bit-identical with features off.

**7. Retained-relay path breaks the identity-compat guarantee** — `RetainedRelaySubject::new` resolves the topology (synthesizing `account:{client}` accounts for empty topologies) *before* handing it to the engine harness, so `explicit_account_topology` is always true and identity seeds become `pad32("account:alice")` instead of the legacy `pad32("alice")` (`retained_relay.rs:118-160`, `subject.rs:779-797`). The same spec now yields different member identities on the engine and retained-relay subjects; the engine path preserves compat and is tested, the retained-relay path silently doesn't.

**8. Wall-clock time inside fixture-comparable data** — `blocked_send_duration_us` (from `Instant::now`) lives in `ScenarioInputLedgerEntry` (`scenario_input_ledger.rs:82-109`), which participates in exact fixture equality via `TraceExpectation::ScenarioInputLedger` and `ObserveExact`. No current vector pins a ledger, but the first promoted fixture that does will be nondeterministic by construction — contrary to the fixed-seed determinism contract.

## Minor

- **Crash-recovery test assertions are too weak** (`crash_recovery_sqlite.rs:740-754`): post-resume it asserts only `epoch == 2` and non-empty members, both already true of the pre-crash incumbent. A recovery bug that drops the interrupted late pass (Bob's winning commit adding Eve) would pass all four kill points. Assert Eve's membership and outbound-intent survival, as the sibling `run_parent_case` does.
- **Sticky group selection**: `active_scenario_group` is never cleared after a wrapped step (`subject.rs:1249-1257`), so a later unwrapped `CreateGroup` fails with a misleading duplicate error and unwrapped sends silently route to the last-selected group. Also, `in_group` `Observe` naming a non-member of that group panics via `.expect("epoch")` instead of returning a structured error (`subject.rs:1240-1244`).
- **Metric semantics drift**: `database_write_bytes` is actually all filesystem block writes (`ru_oublock * 512`, includes report JSON and WAL side files) — the durable v1 field name overstates it; and `reorg_count` silently switches meaning (engine reorgs vs. recovery observations) when `engine_metrics` is `None`, with no marker distinguishing the two (`campaign_metrics.rs:153-155`).
- **Two families under-deliver on their names**: the mixed-binary-policy family body is a bare create+tick with inert version labels (the real mixed-version logic is the compile-time preflight, tested elsewhere), and the catalog's "candidate-replay-exhaustion" case configures no replay budget, so exhaustion is only exercised in the feature-gated harness test (`family.rs:1825-1860`, `1152-1160`).
- **Runner edge cases**: a `wait4` error in the poll loop returns without killing/reaping the worker (orphan process, `cgka-conformance-campaign.rs:224-225`); a worker exiting within the 10 ms poll granularity of the deadline can be marked `timed_out` despite exit 0.
- **Dead artifacts**: the v1 incident-artifact JSON Schema is referenced by no code or test (with `additionalProperties: false` it will silently drift); `tests/fixtures/replayable-membership-fork.json` is loaded by nothing; feature `test-witness-comparison` is wired but unused.
- **Doc drift**: the plan doc cites verification test `actual_process_kill_recovers_at_every_durable_phase`, which doesn't exist (real name: `kill_at_every_durable_convergence_phase_reopens_and_finishes`) — this violates the repo's grep-ability invariant. `.gitignore`'s `incident-exports/` / `incident-replay-output/` entries are unanchored and would swallow future fixture dirs of the same name.
- **CI cost**: the gate rebuilds the workspace with `test-policy-overrides` on top of the default-feature run; the job timeout doubling to 60 min acknowledges roughly 2× wall-clock on every PR (the previous run's simulator step alone took 30 minutes). Worth watching once the gate actually runs.

## What's solid

- **Feature gating is airtight**: `test-policy-overrides`, `test-conformance-snapshot`, and `test-crash-hooks` are default-off everywhere, crash hooks additionally require `debug_assertions`, the test-only canonicalization entry point is `pub(crate)` + cfg-gated, and with features off every production call site hardwires witness admission — verified bit-identical to the merge base. Policy sweeps return data only (`production_auto_tuning_permitted: false`) and cannot touch the pinned v1 policy.
- **Determinism of generation**: fully seeded RNG per case, `BTreeMap`/`BTreeSet` throughout, no HashMap ordering feeding scenario construction.
- **The runner is honest engineering**: durable JSON report files instead of stdout scraping, real child rusage with correct per-OS units, tested kill-and-reap, explicit `unavailable_process_fields` on non-Unix.
- **Repo invariants hold**: no new tracing statements (privacy rules respected), artifacts written via `fs-private` with owner-only modes, `GroupId` handled opaquely with no 32-byte route-id conflation, engine stays generic over `StorageProvider`, `CLAUDE.md` symlink convention satisfied.

## Suggested priorities

1. Fix the CI `just` installation (blocking — the gate is currently dead on arrival) and add `--locked`.
2. Fix the `@group` action-id mismatch and the `ClientsNotEquivalent` oracle mapping — both undermine what Milestone 3 claims to verify.
3. Either enforce the incident-artifact sensitivity/exactness claims (scan/normalize content, verify step↔event correspondence) or soften the "shareable / exact" language to "producer-attested".
4. Acknowledge the replay-budget production change in the PR description and add a shared-budget interaction test; fix the retained-relay identity regression and move `blocked_send_duration_us` out of fixture-compared data.
5. Strengthen the crash-recovery assertions (assert the late branch's member survives).

---

PR #1236, opened 2026-08-03
Unify same-epoch fork resolution across the pairwise and convergence seams

dannym-arx:
## What breaks

When several members commit from the same epoch, the engine picks a winner with two different rules. A member that committed from that epoch resolves the race in `ForkRecoveryManager`: lowest `CommitOrderingKey` wins (`fork_recovery.rs:166`), depth is ignored, and the loser is written `EpochInvalidated`, which no convergence pass ever reads again (`convergence_input::can_start_pass`). Every other member routes the same two commits through distributed convergence (`convergence.rs:127-146`), where a deeper valid branch outranks the ordering key. The router between the two paths is `we_committed_from` (`ingest.rs:764-785`, `:867-870`): in-memory, per-member arrival history.

Two honest members can keep different branches forever. We hit this in 4-VM soak runs against real relays and confirmed it from the engine's own audit logs: at epochs 5, 9, 10 and 11 of one run, some VMs logged `fork_resolution` rows while the others logged `convergence_decision` rows for the same source epoch. All four VMs ended on terminal epoch 78, 21 messages stayed mutually undecryptable (`peel_failed`) across the two lineages, three chat messages were lost for good, and their senders were never told.

## The fix

- Park the pairwise loser as `ConvergenceDeferred` under its source epoch instead of terminal `EpochInvalidated`. Convergence already re-admits that state into branch materialization, so once the rival branch grows deeper, the next pass adopts it and every member lands on the same winner. The source-epoch key matters: the apply stage derives its retained-anchor rewind target from the row's epoch, and a row keyed on the current epoch skips the rewind and fails replay with `WrongEpoch`.
- Carry the kept incumbent's ordering key in `ForkResolution::IncumbentWins` and record its commit digest on `fork_resolution` audit rows. Incumbent-wins rows used to carry `incumbent_digest: null`, so an auditor could not check from the logs whether two members kept the same commit.
- New test `pairwise_incumbent_defers_to_deeper_convergence_branch` drives the cross-seam case end to end: pairwise rejection first, then convergence onto the deeper rival branch, with both members ending on identical rosters. `buffered_losing_fork_commit_raw_row_is_retired_after_confirm_replay` is updated to the new parking state; its raw-row retirement assertions are unchanged.

If the rival branch loses in convergence too, the parked commit is reclassified terminal (`LosingBranch`), so nothing stays reconsiderable forever.

## Test status

The cgka-engine suite is green under `--features test-policy-overrides`, the feature CI enables workspace-wide; an earlier revision of this description claimed 27 pre-existing failures, which was wrong — those tests construct non-baseline policies on purpose and only run under that feature. Clippy and fmt are clean.

## Follow-ups, out of scope here

- `committed_from` and the fork-recovery incumbents live only in memory, so a restart still changes which resolution path a member takes first. Convergence now corrects the outcome either way, but persisting that state would remove the detour.
- Losing-branch application messages still get no `AppMessageInvalidated` on the fork-recovery path and there is no resend path; senders learn nothing when their message dies with a branch.

The marmot colony now reads from one rulebook when the burrow forks.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved same-epoch commit race handling so competing branches converge consistently.
  - Preserved losing branches for later reconsideration instead of permanently invalidating them.
  - Ensured displaced commits remain available for successful future convergence.
  - Added retained incumbent commit details to fork-resolution audit records.
  - Invalid fork candidates now produce a terminal stale result without disrupting processing.

- **Documentation**
  - Added an unreleased changelog entry describing the convergence improvements.

### DIFF
--- a/crates/cgka-engine/src/fork_recovery.rs
+++ b/crates/cgka-engine/src/fork_recovery.rs
@@ -14,7 +14,8 @@
 //! commit is the transport-layer `MessageId`. The two are kept separate inside
 //! `CommitRecoveryRecord` so the ordering key remains transport-independent
 //! while the engine can still reach back to the storage record that needs
-//! `MessageState::EpochInvalidated`.
+//! parking as `MessageState::ConvergenceDeferred` (reconsiderable by a later
+//! distributed-convergence pass, not terminal).
 
 use crate::engine::Engine;
 use cgka_traits::engine::{CommitOrderingKey, CommitOrderingPriority};
@@ -32,8 +33,9 @@ struct CommitRecoveryRecord {
     group_id: GroupId,
     source_epoch: EpochId,
     ordering_key: CommitOrderingKey,
-    /// Storage-layer identity of the commit. Used to update
-    /// `MessageState::EpochInvalidated` when this commit loses a fork.
+    /// Storage-layer identity of the commit. Used to park the row
+    /// `MessageState::ConvergenceDeferred` when this commit loses a fork
+    /// (its branch stays reconsiderable for distributed convergence).
     storage_id: MessageId,
     snapshot_name: String,
 }
@@ -43,8 +45,8 @@ pub(crate) enum ForkResolution {
     CandidateWins {
         winner: CommitOrderingKey,
         invalidated: CommitOrderingKey,
-        /// `MessageId` of the now-invalidated incumbent, for the storage
-        /// `update_message_state` call site.
+        /// `MessageId` of the now-displaced incumbent, for the storage
+        /// write that parks its row `ConvergenceDeferred`.
         invalidated_storage_id: MessageId,
     },
     IncumbentWins {
@@ -176,12 +178,45 @@ impl ForkRecoveryManager {
             });
         }
 
+        // Capture the incumbent's stored row BEFORE the rollback: the
+        // pre-commit snapshot predates the incumbent's own persist, so the
+        // rollback below sweeps its row away entirely. Re-persisting it
+        // afterwards (parked, see below) is what keeps the displaced branch
+        // reconsiderable.
+        let displaced_incumbent_row = match storage.get_message(&incumbent.storage_id) {
+            Ok(record) => Some(record),
+            Err(StorageError::NotFound) => None,
+            Err(e) => return Err(EngineError::Storage(e)),
+        };
         storage.rollback_group_to_snapshot(group_id, &incumbent.snapshot_name)?;
         match storage.release_group_snapshot(group_id, &incumbent.snapshot_name) {
             Ok(()) | Err(StorageError::SnapshotMissing(_)) => {}
             Err(e) => return Err(EngineError::Storage(e)),
         }
         self.incumbents.remove(&key);
+
+        // Park the pairwise-losing incumbent `ConvergenceDeferred`, not
+        // terminally `EpochInvalidated` — the mirror of the incumbent-wins
+        // seam in `ingest_group_message`. The losing branch may still gain
+        // follow-on commits from peers that applied it and never saw the
+        // candidate; distributed convergence then selects the DEEPER branch,
+        // and every convergence-only node reorgs onto it. A terminal (or
+        // missing) row is excluded from convergence input, so this node could
+        // never materialize that branch's root and would diverge from the
+        // fleet forever. If the incumbent's branch also loses in convergence,
+        // it is re-classified terminal there (`LosingBranch`).
+        //
+        // The row is re-keyed at the commit's SOURCE epoch (convergence
+        // derives its rewind target from `record.epoch`) and its payload is
+        // preserved verbatim — for an own commit it carries the convergence
+        // ordering stamp (`stamp_processed_own_commit_record`) that replay
+        // needs to rebuild the ordering key (MLS refuses to process own
+        // commits).
+        if let Some(mut record) = displaced_incumbent_row {
+            record.epoch = incumbent.source_epoch;
+            record.state = MessageState::ConvergenceDeferred;
+            storage.put_message(&record)?;
+        }
 
         Ok(ForkResolution::CandidateWins {
             winner: candidate_key,
@@ -370,18 +405,14 @@ impl<S: StorageProvider> Engine<S> {
         {
             self.epoch_manager
                 .set_stable(group_id.clone(), source_epoch);
-            match self
-                .storage
-                .update_message_state(invalidated_storage_id, MessageState::EpochInvalidated)
-            {
-                Ok(()) | Err(StorageError::NotFound) => {}
-                Err(e) => return Err(EngineError::Storage(e)),
-            }
+            // `ForkRecoveryManager::resolve` parked the displaced incumbent's
+            // row `ConvergenceDeferred` at its source epoch (captured before
+            // the rollback swept it); record the transition for forensics.
             self.audit_group(
                 group_id,
                 crate::audit_helpers::message_state_changed_event(
                     hex::encode(invalidated_storage_id.as_slice()),
-                    MessageState::EpochInvalidated,
+                    MessageState::ConvergenceDeferred,
                     "fork_loser",
                 ),
             );
--- a/crates/cgka-engine/src/fork_recovery.rs
+++ b/crates/cgka-engine/src/fork_recovery.rs
@@ -48,6 +48,11 @@ pub(crate) enum ForkResolution {
         /// `MessageId` of the now-displaced incumbent, for the storage
         /// write that parks its row `ConvergenceDeferred`.
         invalidated_storage_id: MessageId,
+        /// The row `resolve` actually re-persisted `ConvergenceDeferred`, if
+        /// any. `None` when the incumbent had no stored row to capture before
+        /// the rollback (`get_message` → `NotFound`), in which case nothing was
+        /// parked and no `message_state_changed` audit row may claim otherwise.
+        parked: Option<MessageId>,
     },
     IncumbentWins {
         /// Ordering key of the commit this node kept. Carried so the
@@ -152,7 +157,14 @@ impl ForkRecoveryManager {
             .insert((record.group_id.clone(), record.source_epoch), record);
     }
 
-    fn resolve<S: MessageStorage>(
+    /// Decide the pairwise race and, when the candidate wins, displace the
+    /// incumbent atomically.
+    ///
+    /// Bounded by `StorageProvider` rather than `MessageStorage` specifically
+    /// so the displacement can run inside `with_transaction`: capture,
+    /// rollback, re-persist and release are one durable unit. Nothing weaker
+    /// works — see the ordering comment below.
+    fn resolve<S: StorageProvider>(
         &mut self,
         storage: &S,
         group_id: &GroupId,
@@ -178,38 +190,31 @@ impl ForkRecoveryManager {
             });
         }
 
-        // Capture the incumbent's stored row BEFORE the rollback: the
-        // pre-commit snapshot predates the incumbent's own persist, so the
-        // rollback below sweeps its row away entirely. Re-persisting it
-        // afterwards (parked, see below) is what keeps the displaced branch
-        // reconsiderable.
-        let displaced_incumbent_row = match storage.get_message(&incumbent.storage_id) {
-            Ok(record) => Some(record),
-            Err(StorageError::NotFound) => None,
-            Err(e) => return Err(EngineError::Storage(e)),
-        };
-        storage.rollback_group_to_snapshot(group_id, &incumbent.snapshot_name)?;
-
-        // Ordering rationale (recoverability). The displaced row is
-        // re-persisted BEFORE the recovery snapshot is released and before
-        // the in-memory incumbent is dropped. The rollback sweeps the
-        // incumbent's row away, so a failure between rollback and re-persist
-        // must stay retryable: with this order it returns `Err` while the
-        // snapshot is still retained and `self.incumbents` still holds the
-        // record that names it, so the caller fails closed and a later
-        // resolve can redo the rollback + re-persist. Releasing or forgetting
-        // first would lose the displaced branch permanently (row swept,
-        // snapshot gone, record gone). The in-memory removal is therefore the
-        // last step — the point of no return.
+        // Atomicity rationale (durability of the displaced branch). Capture,
+        // rollback, re-persist and release are ONE durable unit.
         //
-        // The snapshot operations necessarily sit OUTSIDE any storage
-        // transaction: create/rollback/release drive their own SQLite
-        // transactions (see the boundary comment in
-        // `openmls_projection::apply_openmls_canonicalization_result_inner`).
-        // The re-persist is a single-row write and is already atomic on its
-        // own, so there is no multi-write group here worth wrapping in
-        // `with_transaction` — and this manager is generic over
-        // `MessageStorage`, which does not expose one.
+        // They have to be. The rollback sweeps the incumbent's stored row away
+        // (the pre-commit snapshot predates that row's own persist), so if the
+        // re-persist failed on its own the displaced branch would be gone with
+        // no way back: a later `resolve` cannot repair it, because (1) the
+        // rollback already removed the row so the capture would read
+        // `NotFound` and silently park nothing, and (2) there typically IS no
+        // later resolve — storage now sits at the pre-incumbent epoch, so a
+        // redelivery of the same candidate applies cleanly with no WrongEpoch
+        // and no fork resolution at all. The branch would vanish silently.
+        //
+        // SQLite snapshot ops JOIN an outer transaction rather than driving
+        // their own (`storage_sqlite::storage::snapshots::
+        // snapshot_rollback_joins_outer_transaction` /
+        // `snapshot_create_joins_outer_transaction`), which is what makes this
+        // grouping possible; `apply_openmls_canonicalization_result` relies on
+        // the same property. A failure anywhere inside therefore unwinds the
+        // rollback too, leaving the incumbent applied, its row intact and its
+        // snapshot retained — genuinely retryable.
+        //
+        // The in-memory `incumbents.remove` stays OUTSIDE, after the commit:
+        // it is the point of no return and must not be reached for a
+        // transaction that rolled back.
         //
         // Park the pairwise-losing incumbent `ConvergenceDeferred`, not
         // terminally `EpochInvalidated` — the mirror of the incumbent-wins
@@ -219,8 +224,19 @@ impl ForkRecoveryManager {
         // and every convergence-only node reorgs onto it. A terminal (or
         // missing) row is excluded from convergence input, so this node could
         // never materialize that branch's root and would diverge from the
-        // fleet forever. If the incumbent's branch also loses in convergence,
-        // it is re-classified terminal there (`LosingBranch`).
+        // fleet forever.
+        //
+        // If the incumbent's branch loses in convergence too, its commits are
+        // NOT terminalized there: an eligible-but-unselected branch's commits
+        // are re-deferred `ConvergenceDeferred`
+        // (`NonSelectedEligibleBranch` in
+        // `classify_losing_materialized_candidate_commits`), so they stay
+        // reconsiderable pass after pass. (`LosingBranch` is an APP-message
+        // invalidation reason, not a commit disposition.) Terminalization
+        // happens only at the rewind horizon —
+        // `BeyondRollbackHorizon`/`BeyondAnchor`, or the
+        // `beyond_retained_anchor` retirement in
+        // `seed_convergence_pass_members`.
         //
         // The row is re-keyed at the commit's SOURCE epoch (convergence
         // derives its rewind target from `record.epoch`) and its payload is
@@ -228,22 +244,41 @@ impl ForkRecoveryManager {
         // ordering stamp (`stamp_processed_own_commit_record`) that replay
         // needs to rebuild the ordering key (MLS refuses to process own
         // commits).
-        if let Some(mut record) = displaced_incumbent_row {
-            record.epoch = incumbent.source_epoch;
-            record.state = MessageState::ConvergenceDeferred;
-            storage.put_message(&record)?;
-        }
+        let snapshot_name = incumbent.snapshot_name.clone();
+        let incumbent_storage_id = incumbent.storage_id.clone();
+        let incumbent_source_epoch = incumbent.source_epoch;
+        let parked =
+            storage.with_transaction(|storage| -> Result<Option<MessageId>, EngineError> {
+                let displaced_incumbent_row = match storage.get_message(&incumbent_storage_id) {
+                    Ok(record) => Some(record),
+                    Err(StorageError::NotFound) => None,
+                    Err(e) => return Err(EngineError::Storage(e)),
+                };
+                storage.rollback_group_to_snapshot(group_id, &snapshot_name)?;
 
-        match storage.release_group_snapshot(group_id, &incumbent.snapshot_name) {
-            Ok(()) | Err(StorageError::SnapshotMissing(_)) => {}
-            Err(e) => return Err(EngineError::Storage(e)),
-        }
+                let parked = match displaced_incumbent_row {
+                    Some(mut record) => {
+                        record.epoch = incumbent_source_epoch;
+                        record.state = MessageState::ConvergenceDeferred;
+                        storage.put_message(&record)?;
+                        Some(record.id)
+                    }
+                    None => None,
+                };
+
+                match storage.release_group_snapshot(group_id, &snapshot_name) {
+                    Ok(()) | Err(StorageError::SnapshotMissing(_)) => {}
+                    Err(e) => return Err(EngineError::Storage(e)),
+                }
+                Ok(parked)
+            })?;
         self.incumbents.remove(&key);
 
         Ok(ForkResolution::CandidateWins {
             winner: candidate_key,
             invalidated: incumbent.ordering_key,
             invalidated_storage_id: incumbent.storage_id,
+            parked,
         })
     }
 
@@ -420,24 +455,25 @@ impl<S: StorageProvider> Engine<S> {
                 invalidated_msg_id,
             },
         );
-        if let ForkResolution::CandidateWins {
-            invalidated_storage_id,
-            ..
-        } = &resolution
-        {
+        if let ForkResolution::CandidateWins { parked, .. } = &resolution {
             self.epoch_manager
                 .set_stable(group_id.clone(), source_epoch);
-            // `ForkRecoveryManager::resolve` parked the displaced incumbent's
-            // row `ConvergenceDeferred` at its source epoch (captured before
-            // the rollback swept it); record the transition for forensics.
-            self.audit_group(
-                group_id,
-                crate::audit_helpers::message_state_changed_event(
-                    hex::encode(invalidated_storage_id.as_slice()),
-                    MessageState::ConvergenceDeferred,
-                    "fork_loser",
-                ),
-            );
+            // Record the transition for forensics — but ONLY for a row that
+            // was really re-persisted. `resolve` parks nothing when the
+            // incumbent had no stored row to capture before the rollback, and
+            // a `message_state_changed` row naming a msg_id that has no
+            // storage row would make the audit log describe a transition that
+            // never happened.
+            if let Some(parked) = parked {
+                self.audit_group(
+                    group_id,
+                    crate::audit_helpers::message_state_changed_event(
+                        hex::encode(parked.as_slice()),
+                        MessageState::ConvergenceDeferred,
+                        "fork_loser",
+                    ),
+                );
+            }
         }
         Ok(resolution)
     }
--- a/crates/cgka-engine/src/fork_recovery.rs
+++ b/crates/cgka-engine/src/fork_recovery.rs
@@ -189,12 +189,28 @@ impl ForkRecoveryManager {
             Err(e) => return Err(EngineError::Storage(e)),
         };
         storage.rollback_group_to_snapshot(group_id, &incumbent.snapshot_name)?;
-        match storage.release_group_snapshot(group_id, &incumbent.snapshot_name) {
-            Ok(()) | Err(StorageError::SnapshotMissing(_)) => {}
-            Err(e) => return Err(EngineError::Storage(e)),
-        }
-        self.incumbents.remove(&key);
 
+        // Ordering rationale (recoverability). The displaced row is
+        // re-persisted BEFORE the recovery snapshot is released and before
+        // the in-memory incumbent is dropped. The rollback sweeps the
+        // incumbent's row away, so a failure between rollback and re-persist
+        // must stay retryable: with this order it returns `Err` while the
+        // snapshot is still retained and `self.incumbents` still holds the
+        // record that names it, so the caller fails closed and a later
+        // resolve can redo the rollback + re-persist. Releasing or forgetting
+        // first would lose the displaced branch permanently (row swept,
+        // snapshot gone, record gone). The in-memory removal is therefore the
+        // last step — the point of no return.
+        //
+        // The snapshot operations necessarily sit OUTSIDE any storage
+        // transaction: create/rollback/release drive their own SQLite
+        // transactions (see the boundary comment in
+        // `openmls_projection::apply_openmls_canonicalization_result_inner`).
+        // The re-persist is a single-row write and is already atomic on its
+        // own, so there is no multi-write group here worth wrapping in
+        // `with_transaction` — and this manager is generic over
+        // `MessageStorage`, which does not expose one.
+        //
         // Park the pairwise-losing incumbent `ConvergenceDeferred`, not
         // terminally `EpochInvalidated` — the mirror of the incumbent-wins
         // seam in `ingest_group_message`. The losing branch may still gain
@@ -217,6 +233,12 @@ impl ForkRecoveryManager {
             record.state = MessageState::ConvergenceDeferred;
             storage.put_message(&record)?;
         }
+
+        match storage.release_group_snapshot(group_id, &incumbent.snapshot_name) {
+            Ok(()) | Err(StorageError::SnapshotMissing(_)) => {}
+            Err(e) => return Err(EngineError::Storage(e)),
+        }
+        self.incumbents.remove(&key);
 
         Ok(ForkResolution::CandidateWins {
             winner: candidate_key,
--- a/crates/cgka-engine/src/fork_recovery.rs
+++ b/crates/cgka-engine/src/fork_recovery.rs
@@ -47,7 +47,14 @@ pub(crate) enum ForkResolution {
         /// `update_message_state` call site.
         invalidated_storage_id: MessageId,
     },
-    IncumbentWins,
+    IncumbentWins {
+        /// Ordering key of the commit this node kept. Carried so the
+        /// `fork_resolution` audit row can record WHICH branch survived —
+        /// without it, incumbent-wins rows are unverifiable against other
+        /// nodes' logs (cross-node convergence cannot be proven from
+        /// forensics alone).
+        kept: CommitOrderingKey,
+    },
     MissingSnapshot,
 }
 
@@ -164,7 +171,9 @@ impl ForkRecoveryManager {
             candidate_mls_bytes,
         );
         if candidate_key >= incumbent.ordering_key {
-            return Ok(ForkResolution::IncumbentWins);
+            return Ok(ForkResolution::IncumbentWins {
+                kept: incumbent.ordering_key,
+            });
         }
 
         storage.rollback_group_to_snapshot(group_id, &incumbent.snapshot_name)?;
@@ -337,7 +346,11 @@ impl<S: StorageProvider> Engine<S> {
                 Some(hex::encode(invalidated.commit_digest)),
                 Some(hex::encode(invalidated_storage_id.as_slice())),
             ),
-            ForkResolution::IncumbentWins => (ForkWinner::Incumbent, None, None),
+            ForkResolution::IncumbentWins { kept } => (
+                ForkWinner::Incumbent,
+                Some(hex::encode(kept.commit_digest)),
+                None,
+            ),
             ForkResolution::MissingSnapshot => (ForkWinner::MissingSnapshot, None, None),
         };
         self.audit_group(
--- a/crates/cgka-engine/src/message_processor/ingest.rs
+++ b/crates/cgka-engine/src/message_processor/ingest.rs
@@ -883,7 +883,7 @@ impl<S: StorageProvider> Engine<S> {
                                         &group_id, &msg.id, msg_epoch, current,
                                     ));
                                 }
-                                ForkResolution::IncumbentWins
+                                ForkResolution::IncumbentWins { .. }
                                 | ForkResolution::CandidateWins { .. } => {
                                     // recovery_snapshot_name_for_fork and
                                     // resolve_fork_candidate read the same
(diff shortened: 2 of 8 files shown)


dannym-arx:
@coderabbitai one more review

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -945,10 +945,43 @@ impl<S: StorageProvider> Engine<S> {
                                     Some((msg_epoch, winner, invalidated, invalidated_storage_id));
                                 continue;
                             }
-                            ForkResolution::IncumbentWins => {
-                                self.update_stored_message_state(
-                                    &msg.id,
-                                    MessageState::EpochInvalidated,
+                            ForkResolution::IncumbentWins { .. } => {
```
erskingardner, line 949: [P1] Keep either pairwise loser reconsiderable

This only parks the inbound candidate when the incumbent wins. In the reverse outcome, `resolve_fork_candidate` still changes the old incumbent commit to `EpochInvalidated` (fork_recovery.rs:366-387). If that old incumbent branch later gains a follow-on commit and becomes the deeper branch, convergence-only peers select it, but this node cannot materialize its root because terminal rows are excluded—exactly the same permanent split with the branch names reversed. Please keep the invalidated incumbent reconsiderable under its source epoch as well (preserving the stored own-commit stamp/payload needed for replay), and add the mirrored candidate-wins/deeper-old-incumbent regression test.

erskingardner (changes requested):
The core direction is right, but the fix handles only one of the two pairwise outcomes. The reverse outcome can still reproduce the same permanent lineage split, so this needs the symmetric state transition and regression coverage before merge.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -945,10 +945,43 @@ impl<S: StorageProvider> Engine<S> {
...
+                                // pick the retained-anchor rewind target, so a
+                                // current-epoch row would make the reorg replay
+                                // this epoch-N commit against the live tip
+                                // state (WrongEpoch).
+                                self.persist_openmls_wire_message(
+                                    &openmls_msg,
+                                    &group_id,
+                                    msg_epoch,
+                                    MessageState::ConvergenceDeferred,
                                 )?;
```
erskingardner, line 1014: Non-blocking, but worth a line of comment: this writes the retained row **without** admitting it to a convergence pass, whereas the normal buffering path treats those two as a single durability boundary — `buffer_openmls_convergence_message_with_time` wraps `put_message` + `admit_stored_message_to_convergence_pass` in one transaction and comments explicitly that "the retained row and its pass membership are one durability boundary" (`distributed_convergence.rs:433-439`).

As far as I can tell that is *safe here*, but only because of a non-obvious coupling: the pairwise arm is reachable only when no pass is active. The `active_pass` term in `commit_should_enter_convergence` (`ingest.rs:773-780`) routes any `msg_epoch < current` commit into convergence whenever a pass is live, so control never reaches this arm with a pass in flight, and the parked row is therefore always picked up by the next `seed_convergence_pass_members` storage scan.

That is load-bearing and undocumented. If someone later relaxes that guard, this row silently becomes an orphan that is a member of no pass. A sentence in the existing comment block pinning the invariant ("reachable only with no active pass, so the next pass seeds this row from storage") would keep the next reader from having to re-derive it.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -945,10 +945,43 @@ impl<S: StorageProvider> Engine<S> {
...
+                                // the row was stored above under this node's
+                                // CURRENT epoch, but convergence-input rows are
+                                // keyed by the commit's SOURCE epoch (see
+                                // `buffer_openmls_convergence_message_with_time`)
+                                // — the apply stage reads `record.epoch` to
+                                // pick the retained-anchor rewind target, so a
+                                // current-epoch row would make the reorg replay
+                                // this epoch-N commit against the live tip
+                                // state (WrongEpoch).
+                                self.persist_openmls_wire_message(
```
erskingardner, line 1009: Non-blocking: this transition emits no audit event, and it is the one step in the new flow that stays invisible to forensics.

Every comparable state change in this area records one — `fork_recovery.rs:380-387` audits the `EpochInvalidated` fork loser, and `distributed_convergence.rs:726-736` audits the `ConvergenceDeferred -> EpochInvalidated` retirement via `message_state_transition_event`. Here the row moves `Created` @ current epoch -> `ConvergenceDeferred` @ source epoch with nothing in the log.

That is a bit at odds with the PR's own stated goal. The `IncumbentWins { kept }` digest was added precisely so an auditor can prove from the logs which branch each member kept — but the corresponding "and the rival is parked reconsiderable at epoch N" half is not recorded, so a replay of the soak-run logs still cannot distinguish "parked, will converge later" from "dropped". A `message_state_transition_event` with a `pairwise_fork_loser` tag would close that.

(To be clear: the old code did not audit this arm either, so this is not a regression — just the natural place to finish the auditability story this PR started.)

erskingardner (dismissed):
Approve. The premise is strong and the fix is appropriately small: two seams resolving the same same-epoch race under different rules is a real correctness bug, it is backed by concrete multi-VM soak evidence, and the actual behavior change is ~6 lines plus an audit-field addition. The new `pairwise_incumbent_defers_to_deeper_convergence_branch` test drives the exact cross-seam sequence end to end rather than asserting on an internal state, which is the right shape for this. CI is fully green on `0d73174`, including all four Rust test shards and the simulator/vectors job.

I verified the mechanism holds:

- `unresolved_commit_state` (`openmls_projection.rs:1980`) and `seed_convergence_pass_members` (`distributed_convergence.rs:748`) both admit `ConvergenceDeferred`, so the parked row really is re-materializable.
- `can_gate_outbound` (`convergence_input.rs:54`) excludes `ConvergenceDeferred`, so parking the loser does **not** wedge the outbound send gate — that was my main worry and it is clean.
- Reusing the `incumbent_digest` audit field for `IncumbentWins { kept }` is semantically consistent: in the `CandidateWins` arm that field already carries the incumbent's digest.

Three non-blocking notes, two inline plus one on the PR description.

**PR description: the "27 failures" claim is wrong and should be removed.** Those failures are not pre-existing breakage and do not "hide regression coverage" — they are the expected result of running `cargo test -p cgka-engine` without the `test-policy-overrides` feature. `ensure_acceptable` (`canonicalization.rs:120-125`) only enforces the v1 pin under `#[cfg(not(feature = "test-policy-overrides"))]`, and CI enables it workspace-wide (`.github/workflows/ci.yml:172`, `justfile:4`). Tests that construct a non-baseline policy are meant to run under that feature. Nothing needs a separate fix here; please drop that paragraph so nobody chases a phantom bug.

The two inline notes are hygiene/robustness observations, not defects. Happy for this to land as is.

dannym-arx:
Both findings were valid against the current code and are fixed in `02f091b1`.

**Fork resolution ordering.** The sequence is now `get_message`, rollback, re-persist the parked row, release the snapshot, and only then drop the in-memory incumbent. A failed write returns `Err` with the recovery snapshot still retained and the incumbent still mapped, so the caller fails closed and a later resolve retries the same rollback. The in-memory removal is the point of no return, so it goes last.

No transaction wraps the sequence: the re-persist is a single-row write, and `ForkRecoveryManager` is generic over `MessageStorage`, which carries no `with_transaction`.

Worth flagging for a human reviewer, since it came up while checking your suggestion: the comment at `openmls_projection.rs:2011` claiming snapshot operations "cannot nest inside an open" transaction appears stale for SQLite. `storage-sqlite/src/storage/snapshots.rs` has `snapshot_rollback_joins_outer_transaction` and `snapshot_create_joins_outer_transaction`, both asserting they do join, and `apply_openmls_canonicalization_result` already calls rollback and release inside `with_transaction`. Left alone here as out of scope.

**Retained anchor lineage.** Confirmed: the check accepted an anchor purely because a snapshot with that epoch's name existed, and `create_group_snapshot` replaces a same-named snapshot, so a second own commit confirmed at the same resulting epoch could hand the fast path an anchor from a different lineage.

The fix takes a narrower shape than re-keying anchors by lineage. That naming scheme is shared across the convergence engine, so changing it would reach well past this bug. Instead the prefix now refuses an anchor whose source epoch carries more than one own confirm-stamped commit row, and falls back to normal replay, which fails closed and lets convergence prune. A missed fast path costs one convergence pass; a wrong rewind costs correctness, so ambiguity resolves conservatively. Rows are counted regardless of `MessageState`, since the anchor is written at confirm time and a later-invalidated row may still have clobbered it.

Three unit tests cover it: the unambiguous case still takes the fast path, the same-epoch overwrite refuses it, and a different-epoch own commit does not create false ambiguity. These are unit tests rather than an engine scenario because driving two own commits to the same resulting epoch needs a rewind-then-recommit interleaving the public API does not expose.

Your note about asserting the parked row's state and source epoch in the seam-1 test was already covered in `1be94d5b`.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1786,6 +1819,133 @@ fn apply_start_epoch_for_canonicalization_result<S: StorageProvider>(
...
+        }
+        let ambiguous_anchor_lineage =
+            own_stamped_commit_ids_by_source_epoch
+                .as_ref()
+                .is_some_and(|by_source_epoch| {
+                    by_source_epoch.get(&record.epoch.0).is_some_and(|ids| {
+                        ids.iter().any(|id| id.as_slice() != message_id.as_slice())
+                    })
+                });
+        if ambiguous_anchor_lineage {
```
erskingardner: **Blocking: this guard does not cover every writer of the anchor name.**

The ambiguity test is "more than one own confirm-stamped commit row at this source epoch". But `openmls-retained-anchor-<epoch>` is written by `retain_current_group_epoch_snapshot`, and its call sites are not limited to own-commit confirms:

- `message_processor/ingest.rs:1283` — called before **every** inbound-commit merge, at the pre-merge current epoch;
- `openmls_projection.rs:1682` — post-apply, at the resulting epoch, on every settled convergence apply;
- `group_lifecycle.rs:728`.

Combined with `INSERT OR REPLACE` in `storage-sqlite/src/storage/snapshots/capture.rs:43`, a peer-authored branch that merely *lands* on epoch N+1 silently replaces the anchor written by our own commit's confirm — and leaves no own-stamped row behind for this check to find.

Concrete failure, reachable from the very scenario this PR adds a test for:

1. X's own commit A (source epoch 1) confirms → `retained-anchor-2` = A's post-merge state.
2. Peer commit B wins the pairwise race; X rolls back and applies B → epoch 2. A is parked `ConvergenceDeferred` (this PR).
3. X ingests any further commit on B's branch. `ingest.rs:1283` runs `retain_current_epoch_snapshot_for_group` at the pre-merge epoch 2 → `retained-anchor-2` is now **B's** state.
4. The A-branch grows deeper; convergence selects it. This scan finds exactly one own-stamped row at source epoch 1 (A), declares the lineage unambiguous, adds A to `skipped_prefix`, and the apply rewinds to `retained-anchor-2`.

Outcome depends on what follows A on the selected branch:
- With further A-branch commits to replay, they fail against B's state and the outer transaction rolls back — fail-closed, but the branch is pruned and X stays diverged, i.e. the bug this PR set out to fix, just narrowed.
- If the anchor-realizable own run is the *whole* remaining accepted set, `replay_messages` is empty, nothing errors, and the apply commits B's lineage while marking A `Processed`. Silent wrong state.

The name-based check cannot be made sound by counting rows, because the clobbering writer leaves no row. Suggest verifying the anchor's identity after the rewind instead of trusting the name — e.g. reload the group and compare its epoch authenticator / confirmed transcript hash against the value implied by the parked commit (the stamp already travels with the row), and fall back to replay on mismatch. That is a positive check rather than another under-approximation, and it stays inside the transaction so a mismatch still fails closed.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1611,13 +1619,34 @@ pub(crate) fn apply_openmls_canonicalization_result_with_profile_policy<S: Stora
...
+        Some(epoch) => epoch,
+        None => apply_start_epoch_for_canonicalization_result(storage, result, &skipped_prefix)?
+            .unwrap_or(current_epoch),
+    };
+    // The own-anchor case must rewind even when the anchor epoch equals the
+    // live tip epoch: the live tip is a DIFFERENT branch's state at that
+    // epoch (that is what displaced the own incumbent), while the retained
+    // anchor holds the own commit's post-merge state.
+    let rewind_to_retained_anchor =
+        apply_start_epoch < current_epoch || own_anchor_prefix.resulting_epoch.is_some();
```
erskingardner, line 1678: Unintended widening. The retain guard below changes from `has_new_commits && apply_start_epoch == current_epoch` to `has_new_commits && !rewind_to_retained_anchor`, i.e. `apply_start_epoch >= current_epoch`. The `apply_start_epoch > current_epoch` case previously skipped `retain_current_group_epoch_snapshot` and now performs it.

Probably benign (it just retains an extra anchor at the live tip), but it is not the change the PR describes and it is not covered by a test. If `apply_start_epoch > current_epoch` is actually unreachable, the clearer expression is `apply_start_epoch == current_epoch && own_anchor_prefix.resulting_epoch.is_none()`, which preserves the old condition exactly and makes the own-anchor carve-out the only delta.

crates/cgka-engine/src/fork_recovery.rs
```diff
@@ -357,18 +427,14 @@ impl<S: StorageProvider> Engine<S> {
...
-            match self
-                .storage
-                .update_message_state(invalidated_storage_id, MessageState::EpochInvalidated)
-            {
-                Ok(()) | Err(StorageError::NotFound) => {}
-                Err(e) => return Err(EngineError::Storage(e)),
-            }
+            // `ForkRecoveryManager::resolve` parked the displaced incumbent's
+            // row `ConvergenceDeferred` at its source epoch (captured before
+            // the rollback swept it); record the transition for forensics.
```
erskingardner: This audit row is emitted unconditionally on `CandidateWins`, but `resolve` only re-persists the row when `get_message` found one — the `Err(StorageError::NotFound) => None` arm at line 186 writes nothing. In that case forensics record a `ConvergenceDeferred` transition for a `msg_id` that has no row at all.

Given that making cross-member convergence provable from the logs is one of this PR's stated goals, it is worth having `resolve` report whether it actually parked the row (e.g. `CandidateWins { parked: bool }` or an `Option<MessageId>`) and emitting the event only when it did.

erskingardner (changes requested):
The premise is solid and the root cause is well identified: two resolution rules for the same race, routed by in-memory `we_committed_from`, is exactly the kind of seam that produces permanent honest-node divergence. Parking the pairwise loser `ConvergenceDeferred` at its source epoch is the right minimal fix, the two new engine-level tests drive the real sequence rather than asserting on internals, and CI is green.

One blocking issue: the retained-anchor lineage guard added in `anchor_realizable_own_commit_prefix` under-approximates who can overwrite `openmls-retained-anchor-<epoch>`. It counts only own confirm-stamped commit rows, but that snapshot name is also (re)written by `retain_current_group_epoch_snapshot` from three call sites that have nothing to do with own commits — including `ingest.rs` before *every* inbound-commit merge. In the exact scenario this PR adds (`pairwise_candidate_win_leaves_old_incumbent_reconsiderable`), one further inbound commit on the displacing branch is enough to clobber the parked own commit's anchor while leaving the guard silent, after which the fast path rewinds to the wrong lineage. Details inline.

Two smaller items inline (an unintended widening of the anchor-retain condition, and an audit row that can assert a state transition that did not happen).

Style note, non-blocking: several of the new comment blocks (the `IncumbentWins` arm in `ingest.rs` is ~40 lines of comment for 12 lines of code) restate the PR description rather than explaining the code. The load-bearing invariants are worth keeping; the narrative recap is not.

crates/cgka-engine/src/openmls_projection.rs
erskingardner: Independently traced @erskingardner's blocking finding above — confirmed. Two additions.

**More clobber paths.** This check only counts own confirm-stamped rows at the parked commit's source epoch, but `retained-anchor-<N>` is name-keyed and replaced by every retain at epoch N:

- `ingest.rs:1339` retains at the pre-merge epoch on **every** inbound commit merge — so any peer commit applied on the displacing branch at that epoch overwrites it (erskingardner's step 3).
- `publish.rs:107` retains the *pre-commit* epoch on every own confirm — so this device's **own next commit** on the displacing branch (source epoch = the contested epoch) also overwrites the anchor, while its stamped row is keyed at that source epoch and never shows up as a second row in this check's lookup for the parked commit's source epoch. Invisible either way.

**Outcome mechanics.** After a missed clobber, the rewind installs the wrong lineage's epoch-N state and the selected branch's follow-on commits replay against it. They fail validation (MLS reconstructs the member signing context — tree hash / confirmed transcript — from local state, so the wrong lineage fails the signature/content check even though the epoch number matches), the apply errors, and because the pass is only marked `Completed` inside the same transaction as the apply (`distributed_convergence.rs:1331-1341`), no pass ever completes: each later pass deterministically re-selects the same deeper branch and fails the same apply, until `retire_stale_convergence_deferred_commits` terminalizes the parked row below the rewind floor. So the practical impact is a multi-epoch liveness wedge followed by permanent divergence — the silent-wrong-state sub-case looks unreachable, since any event that clobbers the anchor also grows the rival branch, which forces the displaced branch to win selection by depth, which forces a non-empty replay, which fails closed.

Also worth noting for the regression test: `pairwise_candidate_win_leaves_old_incumbent_reconsiderable` survives only because nothing is ever applied on the winning branch after the CandidateWins — one more rival-side commit between the displacement and the old-incumbent follow-on breaks the fast path this test exercises.

crates/cgka-engine/src/fork_recovery.rs
```diff
@@ -164,10 +173,67 @@ impl ForkRecoveryManager {
...
+        // The row is re-keyed at the commit's SOURCE epoch (convergence
+        // derives its rewind target from `record.epoch`) and its payload is
+        // preserved verbatim — for an own commit it carries the convergence
+        // ordering stamp (`stamp_processed_own_commit_record`) that replay
+        // needs to rebuild the ordering key (MLS refuses to process own
+        // commits).
+        if let Some(mut record) = displaced_incumbent_row {
+            record.epoch = incumbent.source_epoch;
+            record.state = MessageState::ConvergenceDeferred;
+            storage.put_message(&record)?;
```
erskingardner: The ordering rationale above claims a `put_message` failure here stays retryable: "a later resolve can redo the rollback + re-persist". It cannot, on two counts.

1. The rollback already swept the incumbent's row, so a later resolve's `get_message` returns `NotFound` and the `None` arm silently skips the re-persist.
2. More fundamentally, there typically *is* no later resolve: after this returns `Err`, storage sits at the pre-incumbent epoch, so a redelivery of the same candidate applies **cleanly** (no `WrongEpoch`, no fork resolution). The displaced branch is gone, reconsideration is dead, and because both audit writes happen only after `resolve` returns `Ok`, nothing in the forensic log records the loss at all.

So a transient failure in this window silently reintroduces the exact permanent-divergence bug this PR fixes, while the comment documents recoverability that does not exist. (Adjacent to but distinct from @erskingardner's `fork_recovery.rs:432` note about the unconditional `fork_loser` audit row — that one covers the `NotFound`-on-first-resolve case; this covers the failed-`put` case, where even the misleading audit row is absent.)

Two ways out: (a) if the `storage-sqlite` "snapshots join the outer transaction" behavior holds here as you suspected in the PR discussion, wrap the capture + rollback + re-persist + release in one `with_transaction` at a `StorageProvider`-typed layer so the sequence is genuinely atomic; or (b) failing that, narrow the comment and make this failure loud (distinct error and/or audit row) instead of documenting a retry that cannot happen.

crates/cgka-engine/src/fork_recovery.rs
```diff
@@ -164,10 +173,67 @@ impl ForkRecoveryManager {
...
+        // Park the pairwise-losing incumbent `ConvergenceDeferred`, not
+        // terminally `EpochInvalidated` — the mirror of the incumbent-wins
+        // seam in `ingest_group_message`. The losing branch may still gain
+        // follow-on commits from peers that applied it and never saw the
+        // candidate; distributed convergence then selects the DEEPER branch,
+        // and every convergence-only node reorgs onto it. A terminal (or
+        // missing) row is excluded from convergence input, so this node could
+        // never materialize that branch's root and would diverge from the
+        // fleet forever. If the incumbent's branch also loses in convergence,
+        // it is re-classified terminal there (`LosingBranch`).
```
erskingardner: Minor precision nit: this mis-describes the convergence outcome for the common case. A losing-but-**eligible** branch's commits are re-deferred `ConvergenceDeferred` (`NonSelectedEligibleBranch` in `classify_losing_materialized_candidate_commits`), not terminalized — and `LosingBranch` is an app-message invalidation reason, not a commit disposition. Terminalization only happens at the rewind horizon (`BeyondRollbackHorizon`/`BeyondAnchor` drops, or the `beyond_retained_anchor` retirement in `seed_convergence_pass_members`). The behavior is right (depth races can repeat, so eligible losers should stay reconsiderable), but the comment should say so. Same wording in the `IncumbentWins` comment block in `ingest.rs`.

erskingardner (changes requested):
The direction is right and most of the mechanism checks out, but there is a blocking correctness gap in the new anchor fast path — this independently confirms @erskingardner's finding at `openmls_projection.rs:1911` and adds two clobber paths plus the outcome mechanics — and the new recoverability rationale in `fork_recovery.rs` does not hold.

**Blocking:** the anchor-lineage ambiguity guard is defeated by ordinary traffic. Any commit applied on the displacing branch from the contested epoch — an inbound merge (`ingest.rs:1339` retains at the pre-merge epoch) or this device's own next commit (`publish.rs:107` retains the pre-commit epoch) — overwrites `retained-anchor-<N>` while leaving no own-stamped row at the parked commit's source epoch for the check to find. The fast path then rewinds to the wrong lineage's state; the replayed follow-ons fail validation, the apply errors before the pass is marked `Completed`, and every later pass re-selects the same deeper branch and fails the same way until the parked row ages past the rewind floor — a multi-epoch convergence wedge, then permanent divergence. That is the failure class this PR sets out to eliminate, reachable one ordinary commit beyond the new mirror test's scenario. The positive post-rewind verification erskingardner proposed looks like the right shape.

**Non-blocking but should fix:** the ordering rationale in `fork_recovery.rs` claims a failed re-persist stays retryable ("a later resolve can redo the rollback + re-persist"). It cannot — the rollback swept the row, and the redelivered candidate typically applies cleanly against the rolled-back storage without any resolve, so the displaced branch is lost silently with zero audit (inline at `fork_recovery.rs:234`).

**Minor:** the `re-classified terminal (LosingBranch)` wording mis-describes losing-branch commit dispositions (inline at `fork_recovery.rs:223`).

What I verified and found solid: the premise (two seams resolving one race under two rules, backed by soak evidence); the parked-row source-epoch re-keying (convergence keys rows by source epoch and the apply derives its rewind target from `record.epoch`); `ConvergenceDeferred` re-admission via `unresolved_commit_state` and pass seeding; the pairwise arm's no-active-pass invariant (`active_pass` routing in `commit_should_enter_convergence`); parked rows cannot outlive the rewind horizon (`retire_stale_convergence_deferred_commits`); own-commit rows are source-epoch keyed from send through confirm, so the stamp map indexes them correctly; and the end-to-end tests are the right shape. CI is fully green at `18afc74`.

dannym-arx:
Both blocking findings confirmed and fixed in `991fbb48`, along with the four smaller ones. Thanks for tracing the clobber paths; the row-counting guard was an under-approximation of exactly the shape you describe.

**Anchor lineage: positive check replaces row counting.** You are right that counting own-stamped rows cannot see the clobber, because `retain_current_group_epoch_snapshot` leaves no row behind. Row counting and `own_stamped_commit_ids_by_source_epoch_map` are deleted.

`OwnCommitConvergenceStamp` now carries an optional `post_commit_tree_marker`, the hex of the existing validated-tree marker, which is a content-bound hash over the exported ratchet tree. It is populated inside `do_confirm_published`'s durable transaction right after the merge, so it records the state the own commit actually produced. `verify_rewound_anchor_lineage` recomputes it after `rollback_group_to_snapshot` and before `restore_live_message_and_queue_records` or any replay, inside the same transaction, and returns the new `AnchorLineageMismatch` on a difference or on a stamp that predates the field. On mismatch the transaction unwinds and the error arm rolls back to the pre-apply snapshot, leaving state untouched. A new test asserts that directly: epoch and roster unchanged.

Both producer and verifier read the ciphersuite from `group.ciphersuite()`, so the marker derives from identical inputs on each side. Serde default keeps old rows readable and byte-identical on re-serialization, which the traits snapshot test pins.

**The liveness limitation, stated plainly** (also in the code comment):

> When the anchor HAS been clobbered, the parked own commit can no longer be realized at all. The anchor rewind is its only route, since MLS `process_message` refuses this device's own commits, so normal replay cannot materialize it either. This node fails closed and stays diverged from the branch the rest of the fleet selected until the rewind horizon retires the parked row, after which it reorgs onto whatever remains selectable. Correctness is preserved, no wrong lineage is ever installed, but liveness inside that window is not. Restoring it requires re-keying retained anchors by content instead of by epoch, which is shared with the convergence engine and out of scope here.

Your co-reviewer's read matches: any event that clobbers the anchor also grows the rival branch, so the silent-wrong-state sub-case needs a non-empty replay that then fails closed. The new check makes that failure explicit and early instead of relying on replay to catch it.

**Displacement atomicity: option (a).** The SQLite behavior holds, so `resolve` is re-bounded from `MessageStorage` to `StorageProvider` and capture, rollback, re-persist and release now run in one `with_transaction`, with the in-memory incumbent dropped only after it commits. `release_group_snapshot` is a plain delete on the shared connection, so it joins the outer transaction the same way rollback does. The false retryability comment is replaced with the atomicity rationale and both of your reasons why a later resolve could not repair it.

**The other four.** `CandidateWins` now carries the parked `MessageId`, so the `fork_loser` audit row is emitted only when a row was really parked. The retain guard is back to `apply_start_epoch == current_epoch` plus the own-anchor carve-out; the widening was unintended. Both comments now say losing-but-eligible branches are re-deferred `ConvergenceDeferred` as `NonSelectedEligibleBranch`, with terminalization at the rewind horizon, and note that `LosingBranch` is an app-message reason. `pairwise_candidate_win_with_rival_follow_on_fails_closed_not_wrong_lineage` adds the rival-side commit you identified; I instrumented both tests to confirm the new one is not vacuous, the baseline reaching the fast path with a matching marker and the new one reaching it with a mismatch.

`fork_detection` is 16/16 with the stacked branch, the anchor unit tests are 4/4, the workspace builds clean, and clippy and fmt are green. The stack is rebased onto current master, which now carries the simulator output-audit fix from #1247.

erskingardner:
# Review from Claude: 

I'll pull the PR and review it.

Let me look at the actual code in context rather than just the diff.

Reviewed the full diff and ran the engine tests plus two targeted probes against the branch. The core idea (park the pairwise loser reconsiderable, verify the anchor's lineage before trusting it) is right, and both directions of the pairwise race are now covered. But the fail-closed behavior added in the last round is worse than the PR describes, and I can reproduce it.

## Blocking

**1. `AnchorLineageMismatch` doesn't "fail closed and wait for the horizon" — it hard-wedges the group.**

The PR says (test comment, `verify_rewound_anchor_lineage` docs, commit message) that on a clobbered anchor the node "stays diverged until the rewind horizon retires the parked row, after which it reorgs onto whatever remains selectable." That recovery never happens. I extended the new test `pairwise_candidate_win_with_rival_follow_on_fails_closed_not_wrong_lineage` with follow-up passes:

```
PROBE converge result: Err(AnchorLineageMismatch)
PROBE later rival-branch ingest: Err(Backend("converge: retained anchor does not hold the parked own commit's post-merge state"))
PROBE later converge: Err(AnchorLineageMismatch)
PROBE epoch after later: 3
PROBE repeat pass 0/1/2: Err(AnchorLineageMismatch)   (epoch stays 3 forever)
```

Once the anchor is clobbered:
- every convergence pass re-selects the same deeper branch and aborts with `Err` before `put_convergence_pass` commits, so the pass is retried identically forever;
- `ingest` of *any* later group message returns `Err(Backend(...))` — including commits on the branch the node is actually on;
- the tip epoch therefore can never advance, and the retirement floor is `tip - max_rewind_commits` ([[distributed_convergence.rs:724](https://claude.ai/epitaxy/crates/cgka-engine/src/distributed_convergence.rs:724)](crates/cgka-engine/src/distributed_convergence.rs:724)), so the parked row is never retired. There is no exit.

The scenario is the one in the PR's own test: a `CandidateWins` displacement plus **one** more commit on the winning branch. On a busy relay that is ordinary, not a corner. Master's behavior here was silent divergence — bad, but the node kept working. This trades it for a permanently bricked group and a failing ingest path, which is a worse failure mode than the bug being fixed.

The fix direction is already in the codebase: the materialization-stage rollforward at [[openmls_projection.rs:2620](https://claude.ai/epitaxy/crates/cgka-engine/src/openmls_projection.rs:2620)](crates/cgka-engine/src/openmls_projection.rs:2620) explicitly prunes the branch on any anchor failure rather than failing the pass. The apply stage should do the same — on mismatch, terminalize (or otherwise make unselectable) the unrealizable parked own commit and let the pass settle on the next-best branch, instead of propagating `Err` out of the whole apply.

**2. Rows stamped before this PR wedge the group on upgrade, and the documented fallback for them is wrong.**

`post_commit_tree_marker` is `Option` for migration, and the scan refuses the fast path when it's absent ([[openmls_projection.rs:1941](https://claude.ai/epitaxy/crates/cgka-engine/src/openmls_projection.rs:1941)](crates/cgka-engine/src/openmls_projection.rs:1941)). The comment and the test name say those rows "fall back to normal replay, which fails closed on an own commit and lets convergence prune." They don't. I stubbed the marker to `None` in `publish.rs` and ran the existing (passing) test:

```
reorg back onto the deeper A-branch: Replay("process_message: InvalidCommit(OwnCommitMismatch)")
```

Same hard abort, same wedge as #1. So any own commit confirmed by the current engine and parked after the upgrade bricks its group when convergence later selects its branch. Two consequences: the fallback claim needs correcting, and the selection stage needs to agree with the apply stage about what's realizable.

**3. Selection and apply disagree about realizability (the mechanism behind #1 and #2).**

The materialization rollforward at [[openmls_projection.rs:2620](https://claude.ai/epitaxy/crates/cgka-engine/src/openmls_projection.rs:2620)](crates/cgka-engine/src/openmls_projection.rs:2620) still rewinds to `openmls-retained-anchor-<N>` on name alone — no marker check, only an epoch equality check — even though the stamp now carries the marker. So convergence happily materializes and *selects* a branch that the apply then refuses. Checking the marker there too (and pruning on mismatch, which that path already does for other failures) would make selection skip the branch instead of looping. The PR's own argument — "nothing before the rewind can distinguish a clobbered anchor" — applies verbatim to this call site.

## Non-blocking

- `own_commits.insert_stamped` is now unconditional on row state ([[openmls_projection.rs:1015](https://claude.ai/epitaxy/crates/cgka-engine/src/openmls_projection.rs:1015)](crates/cgka-engine/src/openmls_projection.rs:1015)). Deliberate and explained, but it also registers stamps for rows that `retire_stale_convergence_deferred_commits` already terminalized. Harmless only as long as terminal rows never enter the candidate set — worth a one-line assertion or comment pinning that.
- `ForkResolution::CandidateWins` still emits `invalidated_msg_id` on the `fork_resolution` audit row even when `parked` is `None` ([[fork_recovery.rs:437](https://claude.ai/epitaxy/crates/cgka-engine/src/fork_recovery.rs:437)](crates/cgka-engine/src/fork_recovery.rs:437)) — the same "names a msg_id with no storage row" concern the `parked` guard was added for one event down.
- `compute_validated_tree_marker` hashes only version + ciphersuite + exported ratchet tree — no epoch, no confirmed-transcript hash. Two distinct lineages with identical trees would collide. Hard to construct (key packages are single-use), but the marker is doing identity work now, so binding the transcript hash in would cost nothing.
- The failing local test `stale_commit_outside_rewind_horizon_is_not_treated_as_recoverable_fork` is the known `test-policy-overrides` case, matching your PR note — not a regression. The other 11 in `fork_detection` pass.

I left no comments on the PR and the worktree is back on clean `master`.

# Follow up from GPT when I asked it to review claude's review and do it's own: 

# Review findings

## Summary

| Finding | Verdict | Severity | Rationale |
|---|---|---:|---|
| `AnchorLineageMismatch` permanently wedges convergence | Agree | Blocking | Apply fails before the pass completes, while the unchanged live tip prevents the retirement horizon from advancing. |
| Pre-upgrade markerless rows wedge after upgrade | Agree | Blocking | They fall back to ordinary replay, but OpenMLS cannot replay the device’s own commit. |
| Selection and apply disagree about realizability | Agree | Blocking | Materialization accepts an epoch-named anchor without checking its marker; apply later rejects it. |
| Stamps are registered for terminal rows | Disagree | Not a finding | The surrounding scan already excludes terminal message states. |
| `invalidated_msg_id` may name no stored row | Partly agree | Minor | It can identify the logically displaced incumbent, but “invalidated” no longer describes a reconsiderable row. |
| Tree marker omits the complete MLS state | Agree; stronger than stated | Blocking | Different commits can retain the same ratchet tree while changing GroupContext or application-component state. |

## Blocking

### 1. `AnchorLineageMismatch` permanently wedges convergence

The documented recovery path does not work. When anchor verification fails, canonicalization returns before the convergence pass is marked `Completed`. The next pass therefore selects the same branch and fails identically.

The rewind horizon cannot retire the parked commit because its floor is derived from the live tip. Since the failed apply never advances that tip, the floor never advances either. Later messages cannot resolve the situation because convergence continues retrying the same unrealizable branch.

The new `pairwise_candidate_win_with_rival_follow_on_fails_closed_not_wrong_lineage` test proves only that the failed rewind does not corrupt live state. It stops before checking repeated convergence, subsequent ingest, horizon retirement, or continued group usability.

This turns the previous silent divergence into a permanently unusable group.

### 2. Pre-upgrade own-commit rows cause the same wedge

`post_commit_tree_marker` is optional for migration. When an older row has no marker, `anchor_realizable_own_commit_prefix` refuses the retained-anchor path and falls back to ordinary replay.

That fallback cannot succeed: OpenMLS refuses to process this device’s own commit, and the apply path passes an empty `PrevalidatedOwnCommits`. The apply therefore aborts, leaves the pass unresolved, and retries the same branch indefinitely.

Markerless rows must be classified as unrealizable before selection rather than selected and handed to a replay path that cannot process them.

### 3. Selection and apply disagree about branch realizability

Candidate materialization handles a stamped own commit by rolling forward through `openmls-retained-anchor-<epoch>`. It checks only that the resulting epoch matches; it does not verify `post_commit_tree_marker`.

Apply performs the marker check and rejects the same branch.

Consequently, canonicalization can repeatedly select a branch that apply is guaranteed to refuse. The lineage check must happen during materialization as well, so an invalid anchor prunes the candidate before branch selection.

### 4. The tree marker is not a sufficient lineage identity

`compute_validated_tree_marker` hashes only:

```text
version || ciphersuite || exported ratchet tree
```

That identifies ratchet-tree contents, not the complete MLS state produced by a commit. Distinct commits can retain the same tree while changing group-context or application-component state—for example, `AppDataUpdate` or `GroupContextExtensions` commits.

The marker can therefore accept an anchor from a different lineage without requiring a cryptographic hash collision.

The identity should bind the complete relevant group state, including the GroupContext—epoch, tree hash, confirmed transcript hash and extensions—or use another commit-bound or full-state authenticator.

## Non-blocking

### 5. `invalidated_msg_id` no longer describes the actual disposition

When `CandidateWins`, the displaced incumbent is now parked as `ConvergenceDeferred`, not invalidated. The `fork_resolution.invalidated_msg_id` field therefore misstates what happened.

It is reasonable for the audit event to identify the displaced incumbent even when no storage row was re-persisted, so this is not equivalent to the guarded `message_state_changed` event. However, the field should eventually be renamed to reflect displacement rather than terminal invalidation.

## Finding I do not agree with

### Stamps are not registered for terminal rows

Although `own_commits.insert_stamped` is locally unconditional, the enclosing scan has already filtered records through `record_state_can_contribute_to_openmls_graph`.

Only canonicalization inputs and `Processed` rows reach the insertion. Terminal states such as `EpochInvalidated` do not. No additional assertion is necessary for this behavior.

## Suggested direction

Retained anchors should be immutable and keyed by commit identity or digest rather than epoch alone. Selection and apply should use the same anchor reference and the same realizability check.

At minimum, regression coverage should verify:

- repeated convergence after an anchor mismatch;
- successful ingestion of later traffic;
- convergence-pass completion;
- continued group usability;
- migration behavior for markerless rows; and
- distinct non-tree-changing commits such as `UpdateGroupData`.

erskingardner:
@dannym-arx I think we're stuck here with a bit of a dilemma. Sounds like we probably should have a bit of a brainstorm on this together tomorrow? 

Me: 

let's talk more about 1236 - it's a bit stalled. the approach seems to be stuck between a rock and a hard place

Codex: 
After reviewing the current blocking feedback, #1236 is genuinely caught between safety and liveness—not merely stalled on implementation details.

The fundamental scenario is:

1. This device authors and confirms commit A.
2. A competing commit B wins the immediate pairwise comparison.
3. A is retained because the existing protocol allows its branch to become canonical later if it grows deeper.
4. Ordinary traffic on B overwrites the single retained snapshot for that epoch.
5. A later becomes the globally selected branch.
6. Other members can replay A normally, but its author cannot: OpenMLS rejects replaying your own commit.
7. The exact post-A state needed by the author no longer exists.

At that point, every local response is bad:

| Response | Result |
|---|---|
| Trust the epoch-named snapshot | May install state from the wrong lineage |
| Verify it and return an error | Repeats forever and wedges the group |
| Exclude the unrealizable branch | Nodes choose different winners based on local checkpoint availability |
| Terminalize A immediately | Recreates the original divergence when A later becomes the deeper winner |

That is why incremental changes have not closed the PR. The patch is asking an epoch-keyed, replaceable rewind snapshot to behave like an immutable branch checkpoint. It cannot.

The current #1250 implementation improves detection by binding the own commit to parent/resulting GroupContext hashes and checking during materialization. But hashes only tell us that the retained state is wrong; they cannot restore the correct state. So #1250 does not fundamentally escape the dilemma either.

There are two coherent solutions.

## 1. Make fork selection monotonic

For commits from the same source epoch, choose the winning root using one deterministic root-level ordering rule and never let descendants, depth, or application witnesses cause a losing root to become canonical later.

Then:

- pairwise and convergence-only members use the same rule;
- a terminal loser really is terminal;
- an own commit never needs to be resurrected after losing;
- most of the own-commit anchor machinery disappears;
- application witnesses cannot unexpectedly reverse an earlier fork result.

The cost is that a late-arriving, deterministically preferred root could discard a much longer branch. We would also need to analyze whether the ordering key is grindable or lets one member dominate concurrent forks.

This is the simpler protocol direction and, given our earlier concerns about accumulating constants and rules, the one I think we should investigate first.

## 2. Preserve the adaptive fork rule with immutable branch checkpoints

If depth and witnesses must be able to reverse an earlier decision, then every reconsiderable locally authored commit needs an exact, immutable checkpoint keyed by commit identity—not merely epoch.

That checkpoint must:

- capture the full relevant MLS and Marmot state;
- be referenced by the own-commit record;
- never be overwritten by another branch at the same epoch;
- remain retained while the commit is reconsiderable;
- be used identically during candidate materialization and final application;
- have bounded storage and explicit garbage collection;
- define upgrade behavior for existing commits lacking such checkpoints.

This preserves current semantics, but it commits us to significantly more branch-state persistence machinery.

## Recommendation

Do not merge #1236 in its current form. Keep it open but mark it design-blocked.

Extract and retain its regression scenario, then use the independent model and simulator to compare:

- the current depth/witness-sensitive rule with commit-addressed checkpoints;
- a monotonic root-selection rule;
- admin-versus-self-update contention;
- lost-work and reorganization costs;
- malicious ordering-key grinding;
- offline and delayed-input behavior.

Then #1236 either becomes the focused immutable-checkpoint implementation, or its current approach is replaced by the simpler protocol rule.

Trying to patch the existing epoch-anchor approach further is unlikely to produce a sound merge.

Datawav:

## Final audit

The change is needed: pairwise fork recovery previously terminalized a same-epoch loser while distributed convergence could later prefer that branch by depth, allowing honest members to remain on different lineages.

The design parks pairwise losers as `ConvergenceDeferred`, preserves displaced rows transactionally, and realizes displaced local commits through retained snapshots. A safer alternative is to key retained anchors by commit/content identity. If epoch-keyed anchors remain, their verification must bind to the complete MLS lineage state.

### Findings

1. **High — Retained-anchor verification does not prove the MLS lineage**

   `crates/traits/src/message.rs` defines `post_commit_tree_marker` as a hash of the marker version, ciphersuite, and exported ratchet tree. `crates/cgka-engine/src/openmls_projection.rs` then accepts equality as proof that the snapshot contains the exact state produced by the parked commit.

   A ratchet tree is not a complete MLS-state identity. Distinct commits can retain the same public tree while differing in the confirmed transcript, group context, PSK-derived state, or epoch secrets. An epoch-name collision could therefore pass this check and let replay proceed from the wrong lineage, contradicting the advertised fail-closed guarantee.

   Bind the marker to an MLS lineage authenticator, preferably the epoch authenticator or confirmed transcript hash together with group ID, epoch, and ciphersuite. Add a regression test using distinct valid commits whose exported trees match but whose cryptographic states differ.

2. **Low — The documentation overstates convergence and misdescribes retirement**

   `crates/cli/CHANGELOG.md` says same-epoch races “now converge every member onto the same branch,” but the new `pairwise_candidate_win_with_rival_follow_on_fails_closed_not_wrong_lineage` test explicitly preserves temporary divergence when an anchor is clobbered.

   The PR description also says a convergence loser becomes terminal as `LosingBranch`, while the patched comments state that commits remain `ConvergenceDeferred` until the rewind horizon; `LosingBranch` applies to application messages.

   Describe the result as preventing permanent terminalization when valid recovery material remains, with fail-closed temporary divergence possible until horizon retirement. Clarify the actual commit-retirement states.

erskingardner:
Superseded by #1285.

The branch-selection part of this PR was directionally right: pairwise losers must remain reconsiderable so the distributed convergence engine can later select the same branch on every member. The remaining blocker was recovery of a locally authored Commit. OpenMLS cannot process that Commit again from its network wire form, while this PR's epoch-keyed rollback anchor could be replaced by a rival branch at the same epoch. A marker could detect that mismatch, but it could not recover the selected branch.

#1285 carries the convergence changes forward and replaces that recovery mechanism with immutable, commit-addressed canonical-state checkpoints. Confirmation captures the exact post-merge MLS/Marmot state atomically; after restart, convergence can restore that checkpoint, verify its epoch authenticator, and replay descendants. The checkpoint excludes live messages/queues and is retained only through the bounded rewind horizon.

Closing this PR in favor of the replacement.

PR #1241, opened 2026-08-03
Rebuild fork-resolution routing state on hydration

dannym-arx:
Stacked on #1236, which unified the outcome of same-epoch fork resolution across the pairwise and convergence paths. This PR removes the restart detour that #1236 documented as a follow-up.

## What breaks

Which resolution path a member takes for a same-epoch commit race is decided by two maps that lived only in memory: `EpochManager::committed_from` (the epochs we ourselves committed from, read by the ingest router) and `ForkRecoveryManager::incumbents` (the applied commit, its ordering key, and its pre-commit snapshot per contested epoch). A daemon restart wiped both, so a member that would have resolved a race pairwise instead detoured through a full convergence pass. The outcome converges either way since #1236, but resolution behavior should not depend on process uptime.

## The fix

Hydration rebuilds both maps from evidence the existing durable transactions already produce, so there is no storage schema change and no new write path to tear:

- Retained `fork-<epoch>-<counter>-<hash8>` snapshots are durable and listable; the source epoch is in the name.
- The incumbent commit for each such epoch is the stored `Processed` commit row whose MLS wire bytes carry that source epoch. Own commits recover their ordering key from the confirm-time `OwnCommitConvergenceStamp` (the stamp exists for exactly this reason: MLS cannot replay a device's own commit). Inbound commits recover priority and committer through the existing `probe_commit_ordering_metadata_for_recovery` replay against the fork snapshot.
- `committed_from` entries are restored only alongside a successfully rebuilt incumbent, preserving the invariant that the pairwise router never runs without its snapshot. Half-rebuilt state could otherwise steer ingest into the fail-closed `MissingSnapshot` arm, which would be worse than the convergence detour it replaces.
- Epochs below the rewind horizon are skipped and their snapshots released, matching `prune_fork_recovery_for_group`. Rebuilt snapshot names advance the name allocator so a post-restart snapshot cannot collide with a pre-restart one.
- Rebuild failures are best-effort: the group stays on the convergence path (the behavior every restart had before this PR) and is never quarantined.

## Crash consistency

There is no separate routing-state write. The rebuild derives from artifacts written inside the existing apply/confirm transactions, so a crash between a commit's transaction and the old in-memory map mutation now reconstructs the post-transaction truth on reopen. An orphan snapshot with no `Processed` row (crash between snapshot creation and apply) is skipped and eventually retired by the horizon prune; the rebuild probe itself runs under the existing `SnapshotRollbackGuard` and boot-time interrupted-probe recovery.

## Tests

Two restart tests in `fork_detection.rs`, one per pairwise verdict: a reopened incumbent-keeper still resolves a losing rival pairwise and parks its root `ConvergenceDeferred` at the source epoch, and a winning rival still triggers `ForkRecovered` rollback through the rebuilt snapshot. Full `cgka-engine` suite is green under `--features test-policy-overrides` (the feature CI enables; the pinned-policy check rejects test-only policies without it); clippy and fmt are clean.

## Known limits

- A convergence reorg marks the selected commit `Processed` without registering an incumbent, so a rebuilt incumbent at that epoch pairs the pre-fork snapshot with the reorg winner's row. The in-memory map has the same staleness class; this PR does not widen it.
- Duplicate crash-leftover snapshots at one epoch (non-newest) are skipped, not released; they were equally orphaned before this change.

One more detour closed: the marmot wakes up from hibernation and still remembers which branch of the burrow it dug.

## Divergence from the live map, on purpose

The rebuild re-arms `committed_from` for a fork-invalidated own commit only when that epoch's incumbent also rebuilds (the winner's `Processed` row and snapshot are present). The live map keeps the entry unconditionally. The stricter rule is deliberate: re-arming without an incumbent would steer the pairwise router into the fail-closed `MissingSnapshot` arm, and an orphaned entry buys nothing once the snapshot is gone. In the common post-loss shape (winner applied, snapshot retained) the two agree.

The rebuild also emits a `fork_routing_rebuilt` audit row per group (rebuilt incumbent epochs plus the `committed_from` epochs restored with them), so forensics can tell post-restart reconstruction from in-flight fork resolution.

## Follow-up candidate, not in this PR

Mirroring the stricter rule in the live path would mean clearing `committed_from` when our own commit loses a pairwise race. One wrinkle to weigh first: after a loss the incumbent slot holds the winner, and `we_committed_from` is what routes a third same-epoch rival to the cheap pairwise check against it. Clearing the entry live would push that rival through a full convergence pass instead, and a restart would re-arm the entry anyway (stamped own row plus rebuilt incumbent). Worth deciding with the routing predicate itself: routing on `has_incumbent` instead of `we_committed_from` may be the cleaner unification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fork-resolution state now survives process restarts, improving recovery and convergence after interruptions.
  * Unverifiable retained branches and anchors are safely pruned instead of causing repeated failures.
  * Anchor mismatches are handled as recoverable blocked operations, allowing messages to be reconsidered later.
  * Interrupted fork-recovery probes are detected and restored safely.

* **Audit**
  * Added audit records for reconstructed fork-routing state, including recovered source and committed epochs.

* **Tests**
  * Expanded restart, rollback, convergence, and fork-recovery coverage.

### DIFF
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -727,19 +727,34 @@ impl<S: StorageProvider> Engine<S> {
         let ceiling = base_epoch
             .0
             .saturating_add(policy.convergence.max_rewind_commits);
-        for (message_id, source_epoch) in
-            retire_stale_convergence_deferred_commits(&self.storage, group_id, floor)?
-        {
+        for retired in retire_stale_convergence_deferred_commits(&self.storage, group_id, floor)? {
             self.audit_group(
                 group_id,
                 crate::audit_helpers::message_state_transition_event(
-                    hex::encode(message_id.as_slice()),
+                    hex::encode(retired.message_id.as_slice()),
                     Some(MessageState::ConvergenceDeferred),
                     MessageState::EpochInvalidated,
-                    Some(source_epoch),
+                    Some(retired.source_epoch),
                     "beyond_retained_anchor",
                 ),
             );
+            if retired.own_commit {
+                // A parked commit this device itself published aged out
+                // without ever being realized: the residual divergence the
+                // fail-closed anchor paths accept is now permanent for this
+                // branch, and on every other group surface the node looks
+                // healthy. Surface it beyond the audit row — this class of
+                // split was originally found only by diffing soak-fleet audit
+                // logs. Privacy-safe: aggregate reason only, no group or
+                // message identifiers.
+                tracing::warn!(
+                    target: "cgka_engine::distributed_convergence",
+                    method = "seed_convergence_pass_members",
+                    reason = "own_commit_retired_unrealized",
+                    "an own commit aged out of the retained candidate graph without being realized; this node may be permanently diverged from its fleet on this group"
+                );
+                self.engine_metrics.note_own_commit_retired_unrealized();
+            }
         }
         let projected = self
             .storage
@@ -1359,6 +1374,21 @@ impl<S: StorageProvider> Engine<S> {
                 self.storage
                     .put_convergence_pass(&pass)
                     .map_err(storage_projection_error)?;
+                // Operator-visible residual-divergence signal (beyond the
+                // audit row below): the fleet may have settled on a branch
+                // rooted at this node's own commit that this node can no
+                // longer realize. Left undetected, the node completes every
+                // pass on its own branch and looks healthy on every group
+                // surface while permanently split. Privacy-safe: aggregate
+                // reason only, no group or message identifiers.
+                tracing::warn!(
+                    target: "cgka_engine::distributed_convergence",
+                    method = "converge_stored_openmls_messages",
+                    reason = "anchor_lineage_mismatch",
+                    "selected own-commit branch is locally unrealizable; pass completed unapplied and this node may be diverged from its fleet"
+                );
+                self.engine_metrics
+                    .note_own_branch_unrealizable_blocked_pass();
                 self.audit_group_with_context(
                     group_id,
                     convergence_run_context(&run_id, ConvergencePhase::Blocked),
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -1328,7 +1328,7 @@ impl<S: StorageProvider> Engine<S> {
         let mut completed_pass = pass.clone();
         completed_pass.phase = ConvergencePassPhase::Completed;
         completed_pass.fairness_slot_available = true;
-        let observations = self.storage.with_transaction(|storage| {
+        let apply_outcome = self.storage.with_transaction(|storage| {
             let observations = apply_openmls_canonicalization_result_with_profile_policy(
                 storage,
                 group_id,
@@ -1338,7 +1338,56 @@ impl<S: StorageProvider> Engine<S> {
             )?;
             storage.put_convergence_pass(&completed_pass)?;
             Ok::<_, OpenMlsProjectionError>(observations)
-        })?;
+        });
+        let observations = match apply_outcome {
+            Ok(observations) => observations,
+            // The selected branch is locally unrealizable: its retained anchor
+            // no longer holds the parked own commit's post-merge state (the
+            // apply's `verify_rewound_anchor_lineage` backstop caught a
+            // clobber that happened after materialization proved this
+            // branch). Propagating the error would pin the group forever —
+            // the identical pass would be retried while the tip (and with it
+            // the parked row's retirement floor, `tip - max_rewind_commits`)
+            // never moves. Instead complete the pass unapplied: the next
+            // pass's materialization runs the same lineage proof against the
+            // clobbered anchor, prunes the branch, and settles on the
+            // next-best candidate, and later commits keep applying so the
+            // horizon genuinely retires the parked row.
+            Err(OpenMlsProjectionError::AnchorLineageMismatch) => {
+                pass.phase = ConvergencePassPhase::Completed;
+                pass.fairness_slot_available = false;
+                self.storage
+                    .put_convergence_pass(&pass)
+                    .map_err(storage_projection_error)?;
+                self.audit_group_with_context(
+                    group_id,
+                    convergence_run_context(&run_id, ConvergencePhase::Blocked),
+                    marmot_forensics::AuditEventKind::ConvergenceRunState {
+                        phase: ConvergencePhase::Blocked,
+                        current_tip_epoch: Some(previous_tip.0),
+                        retained_anchor_horizon: Some(retained_anchor_epoch),
+                        reason: Some("anchor_lineage_mismatch".to_string()),
+                        error_kind: Some("anchor_lineage_mismatch".to_string()),
+                    },
+                );
+                result.selected_tip = None;
+                result.selected_fork_epoch = None;
+                result.selected_branch_id = None;
+                result.convergence_status = ConvergenceStatus::Blocked;
+                // The apply transaction unwound, so none of this pass's
+                // dispositions were persisted; returning them would let
+                // `convergence_ingest_outcome` report terminal outcomes for
+                // rows the next pass may still accept.
+                result.accepted_commits.clear();
+                result.accepted_proposals.clear();
+                result.accepted_app_messages.clear();
+                result.dropped_messages.clear();
+                result.invalidated_app_messages.clear();
+                result.deferred_messages.clear();
+                return Ok(result);
+            }
+            Err(err) => return Err(err),
+        };
         crate::test_crash_hooks::pause_if_requested("convergence-pass-completed-durable");
         // Diagnostic only: settling-latency telemetry for the remediation
         // plan — pass open → apply, and the gap since the previous completed
--- a/crates/cgka-engine/src/engine.rs
+++ b/crates/cgka-engine/src/engine.rs
@@ -1577,7 +1577,7 @@ impl<S: StorageProvider> Engine<S> {
             tracing::warn!(
                 target: "cgka_engine::hydrate",
                 method = "hydrate_one_stored_group",
-                %error,
+                transient = error.is_transient(),
                 "failed to rebuild fork-routing state; group stays on the convergence path"
             );
         }
--- a/crates/cgka-engine/src/engine.rs
+++ b/crates/cgka-engine/src/engine.rs
@@ -1285,6 +1285,14 @@ impl<S: StorageProvider> Engine<S> {
         .map_err(|_| GroupHydrationQuarantineReason::GroupRecordLoadFailed)?;
         crate::openmls_projection::recover_interrupted_apply_snapshot(&self.storage, group_id)
             .map_err(|_| GroupHydrationQuarantineReason::GroupRecordLoadFailed)?;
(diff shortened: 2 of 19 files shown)


dannym-arx:
Both findings were valid and are fixed in the top commit of this branch.

**Snapshot allocator.** Confirmed: the allocator only advanced for snapshots that produced a rebuilt incumbent, so anything skipped as an orphan, as below-horizon, or on a failed probe left the counter behind, and a post-restart mint could collide with a name already on disk. The hydrate scan now advances the allocator for every parsed `fork-<epoch>-<counter>` name, before any filtering, through a small helper that takes the already-parsed counter.

A regression test plants an orphan snapshot with counter 9000, restarts the engine, commits, and asserts the next minted name clears it. It fails without the fix, minting counter 2 against the orphan's 9000.

**Entry-API clone.** Applied as proposed: the vacant arm moves the row, the occupied arm still prefers the own-stamped duplicate.

dannym-arx:
@coderabbitai do a full review please

erskingardner:
Re-reviewed this stack after `a2546e95`, with a focus on whether it closes the three blockers I raised on #1236. It does — verified by running the branch, not just reading it. Notes below are on the new work.

## #1236 blockers: closed

`1ae069b8` fixes all three, and in the right direction (make selection agree with apply, rather than making the apply more forgiving):

- **Wedge.** Materialization now runs the same positive lineage proof as the apply, so selection never nominates a branch the apply is guaranteed to refuse; an apply-time mismatch completes the pass unapplied with a `Blocked` result and an `anchor_lineage_mismatch` audit row instead of propagating out of `converge`. I ran `pairwise_candidate_win_with_rival_follow_on_fails_closed_not_wrong_lineage` 12×: 12/12 pass. Repeated passes complete, the tip advances on later inbound commits, and own sends work — the three things the wedge broke (previously every subsequent `ingest` and `send` returned `Err(Backend("converge: ..."))` and the tip never moved, so the retirement floor never moved either).
- **Pre-marker stamps.** Re-ran my earlier experiment (stub `post_commit_tree_marker` to `None`, i.e. any own commit confirmed before this field existed): previously `Replay("process_message: InvalidCommit(OwnCommitMismatch)")` and the same wedge; now the pass settles cleanly and simply declines to adopt the branch. Graceful.
- **Selection/apply asymmetry.** Closed — that is the mechanism above.

Two smaller things from that review also landed: the marker now binds the full `GroupContext` (epoch, tree hash, confirmed transcript hash) under its own version byte, so lineages sharing a ratchet tree across group-data-only commits can't satisfy each other's check; and the fails-closed test is no longer nondeterministic — pruning at materialization removes the digest-tiebreak coin flip that made the A-branch selectable in roughly a third of runs, and the test now asserts the prune, the terminalization, and liveness explicitly rather than only negative invariants that held either way.

Routing the `Blocked` result back through `convergence_ingest_outcome` with all disposition vectors cleared is right: the ingested message reports `Buffered`, not a terminal outcome for a row the next pass may still accept.

Whole `cgka-engine` suite is green here under `--features test-policy-overrides`; the bare-feature failures are the known pinned-policy tests.

Worth stating plainly for merge ordering: **#1236 on its own still ships the wedge.** The fix lives in this stack.

## Findings on this PR

### 1. Own-commit incumbents are rebuilt without proving the snapshot pairing

In `rebuild_fork_routing_state_on_hydrate`, an inbound commit's ordering metadata comes from replaying it against the fork snapshot — which incidentally *proves* that snapshot is that commit's pre-commit anchor. An own commit's comes from the stamp, so the probe is skipped and the (source epoch → newest `fork-` snapshot) pairing is taken on faith. But it is the snapshot, not the ordering key, that a later `resolve()` rolls back to.

I worked through the scenarios and could not construct a break: pending-owned snapshots are excluded explicitly, and after a reorg the newest `fork-N-*` snapshot pairs with the newest applied commit while the displaced one is parked `ConvergenceDeferred` rather than left `Processed`. So I think this is safe today. But the argument is nowhere in the code, and it is the one remaining place where a *name* is trusted to identify state — in a stack whose central lesson is that a name never proves whose state it holds.

Cheap way to make it exact instead of argued: the fork snapshot name is already known at `record_pending`, so persisting it in `OwnCommitConvergenceStamp` lets the rebuild match rather than infer.

### 2. The residual divergence deserves an operator-visible signal

Both fail-closed paths — clobbered anchor, and a pre-marker stamp on a parked row — end with the node terminalizing its own commit and staying on a branch the fleet abandoned. That is the same failure mode this work set out to eliminate, now rarer but not gone, and the clobber trigger is a single commit on the winning branch between the displacement and the reorg, which is not exotic on a busy relay.

The comments and CHANGELOG are honest about this, which I appreciate. The gap is observability: the only trace is an audit row, and a permanently split node looks identical to a healthy one on every group surface. This class of bug was originally found by diffing soak-fleet audit logs — a diagnostic event or metric on the group would make the residual detectable the same way.

### 3. The hydrate rebuild adds a quarantine path to a step documented as best-effort

`engine.rs` says the rebuild is best-effort and "must not quarantine a healthy group", but the probe it now runs can strand a group mid-rollback, and `recover_interrupted_fork_probe` fails closed to quarantine when it finds more than one probe snapshot. That is defensible — several probes really does mean ambiguous state — but the two comments contradict each other, and the whole thing rests on the "fork probes are serialized per group" invariant, which is asserted only in prose.

### 4. Nit: `observe_snapshot_counter` still mis-parses probe names

`ForkRecoveryManager::observe_snapshot_counter` does `strip_prefix("fork-")` then `split('-').nth(1)`, so `fork-probe-<epoch>-<digest>` parses as counter `<epoch>`. Unreachable today — only real fork names reach it, and the hydrate scan rejects probe names because the epoch field fails to parse first — but `a2546e95` just fixed exactly this bug in `fork_snapshot_counters` and left the production helper. Worth the same guard before anything calls it from the scan.

### 5. Nit: the `is_canonical` carve-out rests on an unasserted invariant

Exempting `Processed` own commits from the lineage proof is well argued (the apply skips them via the already-applied prefix, so a wrong anchor is only self-limiting replay context), and I agree it beats pruning a node's own live branch after an upgrade. It holds only while no `Processed` own-commit row can survive off the live chain. That is now load-bearing and would be worth pinning with a test.

Datawav:

## Finding

- **High — Marker-version change makes valid pre-upgrade own-commit branches permanently unrealizable.** In `crates/cgka-engine/src/openmls_projection.rs`, `own_commit_post_merge_tree_marker` changes from the prior tree-only fingerprint to a version-2 hash over the ciphersuite, `GroupContext`, and ratchet tree. Existing stamps contain neither a version nor enough information to select the old verifier. When a pre-upgrade own commit is later displaced and becomes non-canonical, materialization compares its old marker with the new derivation, treats the mismatch as a clobbered anchor, and prunes the branch. This can cause the originating member to reject a valid branch already adopted by peers and remain divergent until later recovery conditions occur. The canonical-row exemption does not help once the commit has been displaced. Store a marker-version discriminator and verify legacy stamps with the legacy derivation, or perform an explicit migration before applying the stricter check. Add an upgrade test covering a legacy-stamped commit that becomes non-canonical after restart.

The underlying need is valid: restart currently loses the in-memory incumbent and `committed_from` routing state, causing same-epoch races to take a different, more expensive convergence path.

The reconstruction design is generally coherent: it derives state from retained snapshots and processed commit rows, records exact snapshot identity for new own commits, replays inbound commits to prove their ordering metadata, restores routing only when an incumbent is successfully rebuilt, and fails back to convergence when reconstruction is unprovable.

The principal alternative is persisting routing state transactionally, but that introduces a schema and write-consistency obligation. Reconstructing from existing durable evidence is preferable if backward compatibility is preserved. For legacy rows without exact snapshot identity, continuing the conservative convergence fallback is safer than inferring identity; for legacy lineage markers, version-aware verification or migration avoids falsely classifying valid state as corruption.

_Advisory audit of head [`3ef64bce4df5`](https://github.com/marmot-protocol/mdk/commit/3ef64bce4df58bb911751d6ad0825e7a61819c95); no approval, merge, push, checkout, or PR-code execution was performed._

Datawav:
## Nightly deep audit

1. **High — Upgrade makes valid pre-upgrade own-commit branches permanently unrealizable**

   `crates/cgka-engine/src/openmls_projection.rs` changes the persisted own-commit marker from the prior tree-only digest to `SHA-256(version || ciphersuite || GroupContext || tree)` and explicitly states that existing stamps will mismatch. During materialization, every non-canonical own commit whose stored marker differs is then treated as an invalid/clobbered anchor and its branch is pruned.

   This is reachable whenever an own commit was parked before upgrading but peers retained or selected a branch rooted at that commit. The upgraded originator cannot replay its own MLS commit, so rejecting its only retained anchor permanently removes its ability to realize that branch. The node may settle elsewhere while peers remain on the old branch—the residual divergence detected by the new warning and retirement metric.

   The stronger GroupContext-bound marker is needed: the old tree-only marker cannot distinguish commits that leave the ratchet tree unchanged. However, silently interpreting every legacy marker as evidence of anchor corruption conflates an expected software upgrade with an actual lineage failure.

   Persist an explicit marker algorithm/version in `OwnCommitConvergenceStamp`. Legacy markers cannot safely receive the new proof semantics, so viable alternatives are to perform an upgrade-time convergence/reconciliation step before activating v2 verification, or explicitly terminalize affected branches with a distinct upgrade-incompatibility reason and operator-visible migration warning. Add a test that reopens a store containing a parked v1-stamped own commit. The current tests cover missing/current markers but not the cross-version state actually created by deployment.

2. **Medium — Crash-leftover fork snapshots evade the rewind bound indefinitely**

   `rebuild_fork_routing_state_on_hydrate` collects every parsed snapshot in `scanned_names_by_epoch`, but `newest_by_epoch` retains only one snapshot per source epoch. Horizon cleanup subsequently iterates only `newest_by_epoch`, releasing only that selected name when the epoch is too old. Older snapshots at the same epoch are neither registered with `ForkRecoveryManager` nor released later.

   This is not merely cosmetic: snapshot creation happens before the apply transaction, so repeated termination after snapshot creation at the same source epoch can create multiple orphan snapshots. After hydration, only the newest participates in cleanup; every older orphan remains durable forever, even once its epoch is below `max_rewind_commits`. Repetition therefore defeats the advertised durable-state bound and can grow encrypted database storage without limit.

   Exact stamp-to-snapshot matching means duplicates cannot all be deleted immediately while an in-horizon own-commit stamp might reference an older one. Build the protected-name set from restored pending records and stamped commit rows, then release every unprotected orphan; below the rewind horizon, release every snapshot for the epoch rather than only the newest. Add a restart test with multiple same-epoch orphan snapshots and verify that all become reclaimable when the horizon advances.

3. **Low — The only hydrate path that cryptographically validates incumbent/snapshot pairing lacks restart coverage**

   Both new restart-routing tests reconstruct an own incumbent from its confirmation stamp. Neither exercises the inbound-incumbent arm, where hydration calls `probe_commit_ordering_metadata_for_recovery` and temporarily rolls live state back through `SnapshotRollbackGuard`.

   That arm has materially different failure and crash behavior: invalid candidate replay must leave the epoch on convergence routing, ordinary errors must restore live state, and process interruption leaves a `fork-probe-` snapshot for the next hydration to recover. Existing interrupted-probe coverage does not begin the probe through the new rebuild path.

   Add an inbound incumbent restart fixture and verify successful reconstruction, invalid-pairing fallback, and interruption/reopen recovery. This closes the principal regression gap in the new reconstruction dataflow.

PR #1243, opened 2026-08-03
Add Milestone 4 independent convergence verification

erskingardner:
## Summary

- add an independent symbolic convergence model covering authentication, authorization, dependency closure, proposals, application witnesses, selection, and dispositions
- add bounded lifecycle and fairness models using TLA+/TLC and Stateright, including a reproducible unfair-scheduling counterexample
- add a nine-case semantic mutation adequacy campaign spanning selector, witness, lifecycle, scheduler, output, publication, and retention behavior
- add a protocol decision gate that classifies every convergence constant and checks operational non-interference
- mark Milestone 4 complete in the convergence reliability tracking plan

## Why

Milestone 4 reduces the risk that the simulator merely reproduces production implementation mistakes. It supplies independent expected behavior, explicit liveness assumptions, mutation evidence that the verification layers detect targeted defects, and a gate around which constants may affect settled protocol state.

## Impact

This changes simulator, test, formal-model, CI-recipe, and documentation code only. It does not change production convergence-engine behavior or production convergence constants.

## Verification

- just milestone4-ci
- cargo test -p cgka-conformance-simulator --locked
- just tamarin: all 78 lemmas verified
- just convergence-ledger-gate: 24 ids and 26 declarations
- just fast-ci
- fair TLC model: 684 distinct states, 1,785 generated, depth 14
- unfair TLC model: expected temporal counterexample detected

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive convergence verification covering lifecycle recovery, fairness, administration, joiner repair, and settled-state consistency.
  * Added an independent validation model and mutation checks to improve confidence in protocol behavior.
  * Added automated liveness checks, including expected failure and counterexample validation.

* **Documentation**
  * Added guidance on convergence policies, lifecycle verification, testing coverage, and known assurance limits.

* **Chores**
  * Integrated the new verification gates into continuous integration.


formal/liveness/README.md
```diff
@@ -0,0 +1,48 @@
...
+model; it is intentionally not stretched into a scheduler or crash/restart
+model.
+
+`ConvergenceLifecycle.tla` models unequal histories, eventual input closure,
+freeze/settle, durable frozen state across crash/restart, temporary resource
+failure, a pending privileged administrative change, repeated member
+self-updates, and a joiner stranded on a losing branch. The only modeled repair
+for that joiner is a fresh-state rejoin; consumed signature state is not reused.
+
+TLC is authoritative for temporal claims because `WF` is explicit and infinite
```
erskingardner: Nit: the spec uses *strong* fairness (`SF_vars` on seven actions), not WF — "because `WF` is explicit" should read "fairness assumptions are explicit" (or `SF`) so the README matches the model.

Justfile
```diff
@@ -258,6 +258,16 @@ milestone3-ci:
     cargo nextest run -p cgka-conformance-simulator --features test-policy-overrides --test policy_sweeps --locked
     cargo run -p cgka-conformance-simulator --features test-policy-overrides --bin cgka-conformance-campaign --locked -- --cases 1 --case-timeout-secs 300 --out target/cgka-milestone3-ci --storage file
 
+# Independent model, lifecycle/fairness, mutation adequacy, and protocol
+# decision gates. Kept as an explicit hardening lane until CI budgets are set.
+milestone4-ci:
+    cargo test -p cgka-conformance-simulator --test independent_reference_model --locked
```
erskingardner: Non-blocking: sibling lanes (`milestone3-ci`, `test`, `simulator-ci`) all go through `cargo nextest run`. Using `cargo test` here loses nextest's partitioning/retry behavior for no stated reason — consider `cargo nextest run -p cgka-conformance-simulator --test <name> --locked` for the four targets.

Justfile
```diff
@@ -269,6 +279,74 @@ tamarin-interactive:
...
+        echo "error: tla2tools v1.7.4 checksum mismatch: expected $expected, got $actual" >&2
+        exit 1
+    }
+
+tla-liveness: _tla-tools
+    #!/usr/bin/env bash
+    set -euo pipefail
+    jar="${TLA2TOOLS_JAR:-target/tla/tla2tools-v1.7.4.jar}"
+    mkdir -p target/tla/states
+    java -XX:+UseParallelGC -jar "$jar" -workers 1 -deadlock \
```
erskingardner, line 329: `-deadlock` *disables* TLC's deadlock check. Presumably intentional — terminal "settled with nothing pending" states are legitimate deadlocks in this model — but a one-line comment in the recipe or the cfg would stop a future reader from removing the flag, and keeps accidental pre-settlement deadlocks on the radar.

erskingardner (commented):
*(GitHub blocks approving your own PR — this is an approve-equivalent review.)*

Independent review at `5991fae2`. Recommend merge.

Verified rather than trusted:

- **Formal claims reproduced at this head.** tla2tools v1.7.4 sha256 matches the Justfile pin (`936a2620…50e88`); the relative `-config` path resolves next to the spec when run from the repo root; the fair model exhausts exactly the claimed 684 distinct / 1,785 generated states at depth 14 with no errors; the unfair config emits the expected temporal counterexample (exit 13, "Temporal properties were violated"), so the expected-failure gate is real.
- **Protocol pins fact-checked** against marmot@`4ad4ae2` (marmot#410): the bounded post-settlement admin preparation opportunity, the non-guarantee under unbounded valid selection-relevant input, the implicit v1 policy with no wire field, and the new-required-app-component rule all match `PROTOCOL_DECISIONS.md`; the seven pinned v1 values match the spec table.
- **The independent model is genuinely independent but faithful.** `compare`/`select`/`eligible`/witness scoring restate production exactly (field order, tie-break reversal directions, sender-per-epoch dedup, quorum, capped score), and the differential corpus includes Tamarin-generated expected winners, so agreement is not self-certification.
- **All three CodeRabbit majors are resolved in the fixup commit**: Rust lifecycle transitions now mirror the TLA+ guards (FreezePass requires closed input, invite barred after settle, stranding only via SettlePass, 3-self-update bound), the comparator is reused via `pub(crate) compare`, `admin_applied` is monotonic in both models, and the minors (DroppedInvalidCommit, fail-closed unknown ids, strict inventory parsing) plus baseline pins for the four constructed-mutant sentinels landed too.
- **Scope matches the stated impact**: no production engine or constant changes; `record_outbound_acknowledgement` is a no-op seam; stateright is dev-only; CI green.

Three non-blocking nits inline. One more that has no diff anchor: the root `AGENTS.md` map still routes "Formal model" only to `formal/tamarin/AGENTS.md` — `formal/liveness/` deserves a row (and if it gets its own `AGENTS.md`, remember the sibling `CLAUDE.md` symlink per repo convention).

crates/cgka-conformance-simulator/src/lifecycle_model.rs
```diff
@@ -0,0 +1,274 @@
...
+                    return None;
+                }
+                next.phase = PassPhase::Settled;
+                next.admin_applied = next.admin_applied || next.admin_pending;
+                next.admin_pending = false;
+                if next.joiner == JoinerState::PendingOnLosingBranch {
+                    next.joiner = JoinerState::Stranded;
+                }
+            }
+            LifecycleActionKind::Crash => {
```
erskingardner, line 250: **Blocking — crash is a no-op, so nothing about durability is modeled.**

`Crash` only flips `crashed`/`crash_budget` on a clone of the current state, and `Restart` only flips it back. No variable is partitioned into volatile vs. durable, so "frozen state survives a crash" is not a proposition this model can express, let alone check.

Consequences that ride on this:

- `frozen_revision_is_durable` in `tests/lifecycle_model.rs` holds by construction.
- `crash_and_resource_recovery_preserve_frozen_membership` asserts that transitions which mutate nothing mutated nothing.
- `MUTATION_MATRIX.md` attributes `frozen_member_persistence` to "TLC plus Stateright lifecycle", which cannot detect it.
- `README.md` line 113 claims this layer catches "frozen-pass loss across restart".

Smallest honest fix: give the state a volatile half (say `staged_revision`) that `Crash` clears while `frozen_revision` survives, then the invariant and the mutant both become meaningful. Otherwise drop the durability language from the three docs.

formal/liveness/ConvergenceLifecycle.tla
```diff
@@ -0,0 +1,148 @@
...
+    /\ inputOpen \in BOOLEAN
+    /\ phase \in {"collecting", "frozen", "settled"}
+    /\ frozenRevision \in Nat
+    /\ crashed \in BOOLEAN /\ resourceAvailable \in BOOLEAN
+    /\ adminPending \in BOOLEAN /\ adminApplied \in BOOLEAN
+    /\ joiner \in {"none", "pending_losing", "stranded", "rejoined_fresh"}
+    /\ selfUpdates \in Nat
+    /\ crashOccurred \in BOOLEAN /\ resourceFailureOccurred \in BOOLEAN
+
+FrozenRevisionDurable == phase = "frozen" => frozenRevision > 0
```
erskingardner: `FrozenRevisionDurable` is a tautology in this spec. The only way to reach `phase = "frozen"` is `FreezePass`, which sets `frozenRevision' = historyA`; `historyA` starts at 1 and only ever increments; and `Crash` lists `frozenRevision` under `UNCHANGED`. TLC can never report this invariant, so the 684-state exhaustion is not evidence for the durability claim in `formal/liveness/README.md`.

Same root cause as the Rust comment on `lifecycle_model.rs`: model a variable that a crash actually loses, or retire the invariant.

crates/cgka-conformance-simulator/MUTATION_MATRIX.md
```diff
@@ -0,0 +1,17 @@
...
+witnesses. `tests/mutation_adequacy.rs` requires every row to change the adopted
+observation and keeps this table exactly aligned with the executable catalog.
+
+| Mutation id | Changed rule | Primary killing layer | Production-shaped sentinel |
+| --- | --- | --- | --- |
+| `selector_comparison_order` | Raw depth precedes effective depth | Independent reference differential | Tamarin policy corpus and `generated_policy_cases_match_selector` |
+| `witness_sender_epoch_deduplication` | Repeated messages count as distinct senders | Independent reference differential | Witness scoring properties and same-sender duplicate campaigns |
+| `app_witness_admission_removal` | Valid app witnesses are ignored | Witness enabled/disabled A/B | Milestone 3 witness-policy campaign and `witness_mode_is_a_first_class_input_not_a_policy_rewrite` |
+| `cutoff_boundary_admission` | Exact rewind boundary is excluded | Reference/canonicalizer boundary | Retained-anchor and rollback-horizon canonicalization tests |
+| `frozen_member_persistence` | Crash discards the frozen candidate set | TLC plus Stateright lifecycle | Durable convergence-phase kill matrix |
```
erskingardner: **Blocking — this row's killing layer is not the layer that runs.**

`frozen_persistence_sentinel` computes its baseline from `LifecycleModel`, then compares it against the string literal `"Collecting:None"`. Neither TLC nor Stateright is invoked, and neither could detect the described defect anyway (see the `lifecycle_model.rs` comment — crash mutates nothing).

What the sentinel actually asserts is that the model's crash transition is still a no-op. That is a baseline pin, not a mutation kill, and it carries no information about production crash durability. Same shape applies to line 14 (`scheduler_deadline_rearm`, whose "mutant" is the unmodified `settled` state and whose "scheduler model" never runs).

Either route these mutants through the layer named in the column, or relabel the column honestly for the constructed-literal rows.

crates/cgka-conformance-simulator/MUTATION_MATRIX.md
```diff
@@ -0,0 +1,17 @@
...
+| `scheduler_deadline_rearm` | New input after settlement does not re-arm | Shared lifecycle transition plus scheduler model | Virtual-time re-arm and deferred-input scheduler regressions |
+| `output_invalidation` | Losing-branch projected output survives | Shared reference evaluation plus application oracle | Losing-app invalidation and exact projection checks |
+| `publication_acknowledgement` | Accepted publication remains pending | Shared reference-subject acknowledgement plus pending-work oracle | Lost/duplicate acknowledgement and no-pending-work checks |
```
erskingardner: This is the weakest row. `publication_ack_sentinel` reduces to:

```rust
let mut baseline = None;
record_outbound_acknowledgement(&mut baseline, SubjectOutboundOutcome::Accepted);
let mutant = None::<SubjectOutboundOutcome>;
```

`Some(Accepted) != None` is true by the body of the one-line function under test. No reference subject is constructed, no pending-work oracle runs, no acknowledgement is lost or duplicated — so "Shared reference-subject acknowledgement plus pending-work oracle" describes nothing that executes.

Drive `ConvergenceSubject::acknowledge_outbound` on a real `ReferenceModelSubject` and observe pending work before/after; that would kill the described defect and would also remove the motivation for the `record_outbound_acknowledgement` seam.

erskingardner (commented):
*(GitHub blocks requesting changes on your own PR — this is a request-changes-equivalent review.)*

Independent review at `5991fae2`. **Request changes** — narrowly, on the assurance claims, not the code.

The substance here is good and I want it to land: `reference_convergence` is genuinely type-independent, and comparing its winner against the production selector over the committed Tamarin policy corpus plus a 256-case shrinkable proptest is real differential evidence. The TLA+ spec with a fair/unfair pair and an expected-failure gate is real work. Nothing touches production convergence behavior. CI is green.

What I can't sign off on is the gap between what several artifacts *claim* is verified and what the committed models can actually detect. This milestone exists to "reduce the risk that the simulator merely reproduces production implementation mistakes" — so overstated coverage here is the specific failure mode being guarded against, and once the plan reads `complete` nobody re-derives it.

**Blocking:**

1. **Crash/restart is a no-op on every durable variable in both models.** `Crash` sets `crashed = TRUE`, `Restart` sets it back; `frozenRevision`/`frozen_revision` and `phase` are explicitly `UNCHANGED`. So there is no volatile-vs-durable distinction anywhere in the state space, `FrozenRevisionDurable` is a tautology TLC can never report, and `crash_and_resource_recovery_preserve_frozen_membership` asserts that transitions which touch nothing touched nothing. Three docs claim otherwise (`README.md` "frozen-pass loss across restart", `formal/liveness/README.md` "durable frozen state across crash/restart", plan §4.2). Either model a durable/volatile split, or drop the durability claims.

2. **At least three MUTATION_MATRIX "primary killing layer" attributions name a layer the sentinel never executes.** `frozen_member_persistence` says "TLC plus Stateright lifecycle"; `scheduler_deadline_rearm` says "scheduler model"; `publication_acknowledgement` says "pending-work oracle". In all three the mutant is a hardcoded literal and no such layer runs. These are baseline pins — useful, but a pin on a model that cannot represent the defect is not a kill. Fix the attributions (or the sentinels), and soften plan §4.3's "There are no surviving targeted mutations or tracked adequacy gaps at this checkpoint."

3. **`record_outbound_acknowledgement` is production-adjacent indirection with no behavior.** It wraps `*resolution = Some(outcome)`, makes three call sites longer and harder to read, and exists so `publication_ack_sentinel` can assert `Some(Accepted) != None`. Revert the seam and the call-site rewrites; if the catalog needs a publication-ack row, drive the real `acknowledge_outbound` path.

**Non-blocking:** `milestone4-ci` is not wired into CI the way `milestone3-ci` is, so the TLA+ half can rot while the plan says complete; `stateright` is the only dep in the crate not using `.workspace`; `SemanticMutation::ALL` can silently drop coverage; the `LifecycleTrace` -> Scenario IR "bridge" is asserted rather than implemented. Details inline.

One more with no diff anchor: root `AGENTS.md` still routes "Formal model" only to `formal/tamarin/AGENTS.md`. `formal/liveness/` should get a row, and per repo convention its own `AGENTS.md` needs a sibling `CLAUDE.md` symlink.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -235,6 +235,13 @@ pub enum SubjectOutboundOutcome {
     ReachedNoEndpoint,
 }
 
+pub(crate) fn record_outbound_acknowledgement(
```
erskingardner: **Blocking — abstraction with no behavior, and it makes the call sites worse.**

The body is `*resolution = Some(outcome);`. It replaces three readable assignments with a form that has to borrow through a `get_mut(...).expect(...)` as an argument:

```rust
record_outbound_acknowledgement(
    &mut self
        .outbound_records
        .get_mut(&outbound_sequence)
        .expect("validated outbound record remains present")
        .resolution,
    outcome,
);
```

That is strictly longer and harder to read than `...resolution = Some(outcome);`, in files this PR otherwise has no reason to touch. The only consumer is `publication_ack_sentinel`, whose assertion is a tautology over this function body (see `MUTATION_MATRIX.md` line 16).

Please revert this function and the three call-site rewrites in `subject.rs` / `reference_subject.rs`. Net deletion, no behavior change, and the mutation row can be rebuilt against the real subject path.

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -739,46 +742,100 @@ is not based on compile-only catalog coverage.
...
+- [x] Mutate retained-history and expiration boundaries.
+- [x] Maintain a matrix showing which verification layer kills each mutation.
+
+`SemanticMutation::ALL` contains nine one-rule simulator mutants with minimal deterministic witnesses. The campaign
+requires every mutant to change the adopted observation; none is a production compile feature. It kills raw-depth-first
+selection, non-deduplicated witnesses, witness removal, exclusive cutoff, crash-lost frozen state, suppressed
+post-settlement re-arm, retained losing output, uncleared accepted publication, and inclusive expiration. The
+machine-checked [`MUTATION_MATRIX.md`](../../crates/cgka-conformance-simulator/MUTATION_MATRIX.md) has exactly one row
+per executable mutant and maps each to its primary independent layer plus a production-shaped regression. There are
+no surviving targeted mutations or tracked adequacy gaps at this checkpoint.
```
erskingardner: **Blocking — overstated.** "There are no surviving targeted mutations or tracked adequacy gaps at this checkpoint" is true only under a definition of "killed" that includes comparing a real baseline against a hardcoded literal.

For `frozen_member_persistence`, `scheduler_deadline_rearm`, and `publication_acknowledgement`, no mutated rule is evaluated by any verification layer — so the campaign cannot distinguish "the layer detects this defect" from "the layer cannot represent this defect". Those are exactly the adequacy gaps the sentence denies.

Relatedly, line 796's "crash-lost frozen state" is not killed by anything: the lifecycle models have no volatile state to lose.

Suggest either (a) making those three mutants executable through their named layers, or (b) keeping the pins and stating plainly which rows are baseline pins with a tracked gap. Option (b) is cheap and keeps the milestone honest.

crates/cgka-conformance-simulator/src/lifecycle_model.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub violated_assumption: Option<String>,
+    pub actions: Vec<LifecycleAction>,
+    pub states: Vec<LifecycleState>,
+}
+
+/// Minimal finite witness for the administrative-progress non-guarantee.
+///
+/// The final `StopUnfairly` represents either an execution that never closes
+/// its input set or one that never fairly schedules the pending admin action.
+pub fn minimal_admin_starvation_trace() -> LifecycleTrace {
```
erskingardner, line 321: Non-blocking: the "trace bridge" is asserted rather than implemented. Nothing in this PR projects a `LifecycleTrace` into a `ScenarioSpec` — `formal/liveness/counterexamples/admin-starvation.scenario.json` is hand-authored, and its ids (`step-4:self_update`) do not even match the ones this function emits (`model-step-0:self_update`).

So `committed_counterexample_is_canonical_scenario_ir_with_stable_action_ids` checks that a hand-written JSON file compiles and that two magic indices hold their ids; it does not check that the file corresponds to the model trace. The two artifacts can drift apart silently, which is the one thing the bridge is supposed to prevent.

The module doc says the identity "can be projected into Scenario IR" — accurate, but the plan's §4.2 "[x] Emit model traces that preserve enough action identity to compile into Scenario IR regression cases" reads as though the projection exists. A small `fn to_scenario_spec(&self) -> ScenarioSpec` (or a test asserting the JSON's action ids are derivable from `minimal_admin_starvation_trace()`) would make the checkbox true.

crates/cgka-conformance-simulator/src/mutation_adequacy.rs
```diff
@@ -0,0 +1,329 @@
...
+    CutoffBoundaryAdmission,
+    FrozenMemberPersistence,
+    SchedulerDeadlineRearm,
+    OutputInvalidation,
+    PublicationAcknowledgement,
+    RetainedHistoryExpirationBoundary,
+}
+
+impl SemanticMutation {
+    pub const ALL: [Self; 9] = [
```
erskingardner: Non-blocking: `ALL` is a hand-maintained `[Self; 9]`, and `every_targeted_semantic_mutation_is_killed` only asserts `results.len() == SemanticMutation::ALL.len()` — which is trivially true because `run_all_mutation_sentinels` maps over `ALL`.

Adding a tenth variant compiles fine (the `id()` and `run_mutation_sentinel` matches force a sentinel, but nothing forces the `ALL` entry), and both the kill test and the matrix test iterate `ALL`, so coverage silently drops to 9/10 with no failure. Given the "complete-kill" claim rests on this array, worth closing: either derive the list, or assert the returned `mutation` values equal `ALL` as a set with no duplicates.

crates/cgka-conformance-simulator/Cargo.toml
```diff
@@ -46,3 +46,4 @@ libc.workspace = true
 transport-quic-stream.workspace = true
 transport-quic-broker.workspace = true
 rusqlite.workspace = true
+stateright = "0.31.0"
```
erskingardner: Non-blocking: every other dependency in this manifest uses `.workspace = true`; this is the lone pinned literal. Please add `stateright` to `[workspace.dependencies]` and use `stateright.workspace = true` for consistency.

While there: `stateright` pulls `tiny_http`, `dashmap`, `ahash`, `id-set`, `nohash-hasher`, and a second `hashbrown` into `Cargo.lock`. `tiny_http` is an HTTP server for stateright's explorer UI, which nothing here uses. If 0.31 gates it behind a feature, `default-features = false` would keep an HTTP listener out of the workspace tree entirely — a nice fit with the repo's host-safety posture even for a dev-dependency.

Justfile
```diff
@@ -258,6 +258,16 @@ milestone3-ci:
     cargo nextest run -p cgka-conformance-simulator --features test-policy-overrides --test policy_sweeps --locked
     cargo run -p cgka-conformance-simulator --features test-policy-overrides --bin cgka-conformance-campaign --locked -- --cases 1 --case-timeout-secs 300 --out target/cgka-milestone3-ci --storage file
 
+# Independent model, lifecycle/fairness, mutation adequacy, and protocol
+# decision gates. Kept as an explicit hardening lane until CI budgets are set.
+milestone4-ci:
```
erskingardner, line 263: Non-blocking, but it undercuts marking Milestone 4 complete: `milestone3-ci` is invoked from the "Simulator and vectors" job in `.github/workflows/ci.yml`, and `milestone4-ci` is invoked from nowhere.

The four Rust test targets are picked up incidentally by the workspace `cargo nextest` run, so those are covered. `tla-liveness` and `tla-liveness-counterexample` are not — meaning the TLA+ spec, both configs, and the expected-failure gate can drift or break with CI staying green, while `convergence-reliability-plan.md` says the milestone is done.

The recipe comment ("until CI budgets are set") acknowledges this, but 684 states on one worker is seconds of TLC. Suggest adding it to the existing simulator job, or filing the follow-up and linking it from the plan's exit gate so it isn't silently deferred.

Separate nit on `tla-liveness-counterexample`: `rg -q 'Temporal properties were violated|Error: The behavior up to this point is'` — the second alternative also matches a plain invariant violation, so it weakens an otherwise well-built expected-failure gate. `Temporal properties were violated` alone is the assertion you want.

PR #1244, opened 2026-08-03
Add Pi terminal harness connector

erskingardner:
## Summary

- add an isolated `wn-pi` connector that invokes Pi's JSON mode over stdin and returns completed assistant text
- extract the shared connector runtime from `wn-opencode` into `marmot-terminal-harness` while preserving OpenCode behavior
- add Linux/macOS installer services, release artifacts, documentation, and connector coverage for Pi

## Validation

- `just fast-ci`
- `cargo test -p marmot-terminal-harness -p wn-opencode -p wn-pi`
- `cargo test -p wn-pi --test e2e_connector -- --ignored --nocapture`
- `cargo test -p wn-opencode --test e2e_connector -- --ignored --nocapture`
- `integrations/opencode/marmot/tests/test_installer.sh`
- `integrations/pi/marmot/tests/test_installer.sh`

## Summary by CodeRabbit

- **New Features**
  - Added Pi terminal-harness support, including session persistence, streaming responses, timeout handling, and secure working-directory validation.
  - Added installation support for Pi and OpenCode across supported platforms, with interactive, non-interactive, dry-run, and service-management options.
  - Added Pi binaries, installers, and metadata to versioned and latest releases.
- **Improvements**
  - Standardized terminal-harness configuration, error handling, process management, and testing across integrations.
- **Documentation**
  - Added setup, configuration, development, and release guidance for Pi and shared terminal harnesses.
- **Tests**
  - Added comprehensive installer, integration, and end-to-end coverage for both harnesses.


scripts/install-pi-marmot.sh
```diff
@@ -0,0 +1,859 @@
+#!/usr/bin/env bash
```
erskingardner, line 1: This is 859 lines and `scripts/install-opencode-marmot.sh` is *also* exactly 859 lines — a normalized diff shows the only differences are the harness name, the home/service/label defaults, and the env-var prefix. Everything substantive (platform detection, asset download + checksum verify, plist/systemd unit generation, `activate_systemd_user_service`, bootstrap + JSON extraction, env-file writing, arg parsing) is identical.

This is now the third copy of this scaffolding in `scripts/`. Suggest one `install-terminal-harness-marmot.sh` parameterized by harness name (`wn-pi` / `wn-opencode`), with the release workflow emitting per-harness wrappers, or a sourced `scripts/lib/harness-install.sh`. As written, any fix to checksum handling, socket derivation, or service activation has to be applied in two places and the installer tests won't catch the one you miss.

integrations/pi/marmot/tests/test_installer.sh
```diff
@@ -0,0 +1,319 @@
+#!/usr/bin/env bash
```
erskingardner, line 1: Same story as the installer itself: 319 lines, identical line count to `integrations/opencode/marmot/tests/test_installer.sh`, 92 differing lines that are all naming. The curl/tar/systemctl mocks and every assertion helper are duplicated verbatim. If the installer gets parameterized, this should become one test invoked twice with a harness name.

integrations/pi/marmot/src/pi.rs
```diff
@@ -0,0 +1,411 @@
...
+    })?;
+    let mut stderr_task = tokio::spawn(capture_stderr(stderr));
+    let start = Instant::now();
+    let mut observed_session = None;
+    let mut error_summary = None;
+    let mut idle_deadline = tokio::time::Instant::now() + invocation.idle_timeout;
+
+    let lifecycle_result = timeout_at(total_deadline, async {
+        stdin
+            .write_all(invocation.prompt.as_bytes())
```
erskingardner: The entire prompt is written and stdin is shut down *before* the stdout reader is constructed on line 90. If `pi` emits more than a pipe buffer of stdout before draining stdin, `write_all` blocks and nothing is reading the other end — classic pipe deadlock.

The recovery path makes it worse than it looks: this write sits inside `timeout_at(total_deadline, ...)`, so the stall is bounded by `WN_PI_TIMEOUT_SECS` (default **3600s**), not `WN_PI_IDLE_TIMEOUT_SECS` (120s). One malformed interaction wedges that group's queue for an hour. The bridge does not cap inbound prompt size either (Marmot messages go up to 60 KB, above the 16 KB default macOS pipe buffer).

OpenCode passed the prompt as an argv element, so this risk is new to the Pi backend. Either spawn the write concurrently with the stdout loop, or wrap the write in the idle timeout so it fails in 120s like every other stall.

erskingardner (commented):
**Recommendation: request changes** (posted as a comment — GitHub will not let the PR author request changes on their own PR).

## Summary

The premise is right: a second terminal harness justifies extracting the shared runtime, and the `Backend` trait boundary is a clean seam. The bridge/control/store/chunking extraction looks behavior-preserving, the Pi event parser correctly emits only completed assistant text (no thinking, no tool output), and CI is green.

But the extraction stops at the Rust bridge and leaves the *largest* shared surfaces copied. Of ~2,980 added lines, roughly 1,900 are near-verbatim copies of the OpenCode connector:

| New file | Lines | Relationship to OpenCode equivalent |
| --- | --- | --- |
| `scripts/install-pi-marmot.sh` | 859 | identical line count to `install-opencode-marmot.sh`; names swapped |
| `integrations/pi/marmot/tests/e2e_connector.rs` | 390 | copy of the 402-line OpenCode e2e |
| `integrations/pi/marmot/tests/test_installer.sh` | 319 | identical line count; 92 differing lines, all naming |
| `integrations/pi/marmot/src/config.rs` | 258 | copy with `WN_OPENCODE_` -> `WN_PI_` |
| `integrations/pi/marmot/src/pi.rs` (plumbing half) | ~120 | `capture_stderr`, `strip_ansi`, `truncate_to_char_boundary`, `next_stdout_line`, `cleanup_failed_run` verbatim |

That is the third copy of the 850-line installer scaffolding in this repo and the second copy of everything else. A shared runtime crate that does not absorb the installer, the env-config parsing, the process/stderr plumbing, or the e2e scaffolding has not really paid for itself — and every one of these pairs will drift. `dirs_home()` now exists three times (shared crate `pub fn`, OpenCode config, Pi config) *in the same PR that introduced the shared copy*, which is the drift starting on day one.

### Blocking

1. **Deduplicate the genuinely shared surfaces.** Concretely: move the process/stderr/ANSI plumbing into `marmot-terminal-harness` (it is neither "command construction" nor "event parsing", so the crate's own AGENTS.md scope says it belongs there); give the harness crate a prefix-parameterized env-config builder so each connector declares its prefix and backend-specific fields only; parameterize one installer + one installer test by harness name instead of forking a fourth copy; lift the e2e scaffolding (`ChildGuard`, `wait_for`, `send_control_request`, `spawn_wn_agent`) into a shared dev-dependency helper. Inline comments mark the specific spots.
2. **The Pi CLI contract is unverified.** `--mode json`, `--session-dir`, `--session`, prompt-over-stdin, and the `session` / `message_end` / `stopReason` event shapes are asserted only against `fake-pi` mocks written from the same assumptions, so a green test suite is no evidence the real binary works. Please state which Pi version this was run against end-to-end, and record it in `integrations/pi/marmot/AGENTS.md`. This PR ships a release artifact and a `curl | bash` installer — shipping that on an unverified contract is the risk.
3. **`pi.rs` writes the whole prompt before reading any stdout** (see inline). Recoverable only after the 3600s total timeout, not the 120s idle timeout.

### Non-blocking

- The `TRACE_TARGET` rename (`wn_opencode` -> `marmot_terminal_harness`) is an operator-visible change: existing `RUST_LOG=wn_opencode=debug` setups silently stop capturing bridge/control logs. Worth a release note; the e2e tests' own `RUST_LOG` filters are already stale because of it (inline).
- Mixed `display_name` casing produces `[wn-opencode] opencode went silent...` vs `[wn-pi] Pi went silent...` in user-visible replies. Pick one convention.
- `error.rs` in both connectors is now a single re-export line; delete the module and import from the crate.
- One unrelated re-indent in the release workflow (inline).

The OpenCode-behavior-preservation work here is solid — I'd like to see the same discipline applied to the copy-paste before this lands.

integrations/pi/marmot/src/pi.rs
```diff
@@ -0,0 +1,411 @@
...
+async fn cleanup_failed_run(child: &mut Child, stderr_task: &mut JoinHandle<String>) {
+    stderr_task.abort();
+    let _ = child.start_kill();
+    let _ = child.wait().await;
+    if !stderr_task.is_finished() {
+        let _ = stderr_task.await;
+    }
+}
+
+async fn capture_stderr(stderr: tokio::process::ChildStderr) -> String {
```
erskingardner: `capture_stderr`, `truncate_to_char_boundary`, `strip_ansi`, `next_stdout_line`, `cleanup_failed_run`, and `STDERR_CAPTURE_BYTES` (lines ~240-300) are verbatim copies of the same functions in `integrations/opencode/marmot/src/opencode.rs`. None of them are backend-specific — `integrations/terminal-harness/AGENTS.md` scopes the harness crate to everything except "backend command construction and event parsing", which is exactly what these are not.

Moving them into the shared crate deletes ~120 lines here and keeps a single definition of the stderr cap and the ANSI stripper.

integrations/pi/marmot/src/pi.rs
```diff
@@ -0,0 +1,411 @@
...
+            cleanup_failed_run(&mut child, &mut stderr_task).await;
+            Err(RunFailure {
+                error: HarnessError::BackendTimedOut,
+                observed_session,
+            })
+        }
+    }
+}
+
+fn build_run_args(session_dir: &Path, session_id: Option<&str>) -> Vec<String> {
```
erskingardner, line 173: The whole Pi CLI contract lives here and in `parse_event_line`, and it is only ever exercised against `fake-pi` bash mocks that this PR also wrote — the unit test's mock and the e2e's mock both implement the same assumptions being tested, so passing tests carry no information about the real binary.

Please confirm `wn-pi` was driven end-to-end against a real `pi`, and pin the version in `integrations/pi/marmot/AGENTS.md`. Two specifics worth calling out in the answer:

- Does `pi --mode json` read the prompt from stdin, or does it expect an argv prompt / a different framing?
- What happens when a stored `--session` id no longer exists in `--session-dir`? Right now a resume failure is indistinguishable from a normal run, so a group could silently lose its conversation history (or fail every prompt) with no diagnostic.

integrations/pi/marmot/src/pi.rs
```diff
@@ -0,0 +1,411 @@
...
+    }
+}
+
+async fn run_with_bin(
+    bin: &str,
+    session_dir: &Path,
+    invocation: Invocation,
+    tx: mpsc::Sender<RunnerEvent>,
+) -> std::result::Result<Outcome, RunFailure> {
+    fs_private::create_dir_all_private(session_dir).map_err(|error| RunFailure {
```
erskingardner: Minor: the private session dir is recreated on every single invocation. This only needs to happen once at startup — do it where the backend is constructed and keep the hot path free of filesystem work.

integrations/pi/marmot/src/config.rs
```diff
@@ -0,0 +1,258 @@
...
+            backend_idle_timeout: self.idle_timeout,
+            display_name: "Pi",
+            reply_prefix: "wn-pi",
+            bin_env_name: "WN_PI_BIN",
+        };
+        (config, backend)
+    }
+}
+
+fn dirs_home() -> PathBuf {
```
erskingardner: `dirs_home()` now exists three times after this PR: `marmot_terminal_harness::dirs_home` (public, added here), `integrations/opencode/marmot/src/config.rs:148`, and this copy — all identical. The shared crate already exports it; import it.

Same problem one level up: `DEFAULT_MAX_REPLY_BYTES` and `MARMOT_MESSAGE_BYTES_CEILING` are now defined in *both* the shared crate (where Pi reads them) and `opencode/src/config.rs:9-10` (where OpenCode still reads them). Two sources of truth for a protocol-adjacent size limit is exactly the drift this extraction was supposed to prevent — point OpenCode at the shared constants and delete the local ones.

integrations/pi/marmot/src/config.rs
```diff
@@ -0,0 +1,258 @@
+#[cfg(test)]
```
erskingardner, line 1: This module is a copy of `integrations/opencode/marmot/src/config.rs` with `WN_OPENCODE_` swapped for `WN_PI_`. `parse_hex_csv`, `normalize_hex`, `parse_u64`, `parse_usize`, `config_error`, the auth-token-file handling, the activation check, the reply-bytes bounds check, and the XDG state-path derivation are all identical logic.

A prefix-parameterized config builder in `marmot-terminal-harness` (each connector supplies its env prefix plus its backend-specific fields, e.g. bin + session dir) would collapse both connectors' config modules to a few dozen lines each and guarantee the two harnesses cannot disagree about, say, what a valid allowlist entry is.

integrations/pi/marmot/src/error.rs
```diff
@@ -0,0 +1 @@
+pub(crate) use marmot_terminal_harness::Result;
```
erskingardner: A whole module for one re-export. Import `marmot_terminal_harness::Result` directly in `main.rs` and delete this file — same for the now-single-line `integrations/opencode/marmot/src/error.rs`.

integrations/pi/marmot/tests/e2e_connector.rs
```diff
@@ -0,0 +1,390 @@
...
+    let child = Command::new(env!("CARGO_BIN_EXE_wn-pi"))
+        .env("MARMOT_AGENT_SOCKET", socket)
+        .env("WN_PI_ACCOUNT_ID_HEX", account_id_hex)
+        .env("WN_PI_ALLOWED_SENDERS_HEX", SENDER_ACCOUNT_ID_HEX)
+        .env("WN_PI_BIN", fake_pi)
+        .env("WN_PI_STATE_PATH", state_path)
+        .env("WN_PI_MAX_REPLY_BYTES", MAX_REPLY_BYTES.to_string())
+        .env("WN_PI_TIMEOUT_SECS", "5")
+        .env("WN_PI_REQUEST_TIMEOUT_SECS", "5")
+        .env("RUST_LOG", "warn,wn_pi=info")
```
erskingardner: `wn_pi=info` no longer matches anything: after the extraction the bridge and control logs are emitted under target `marmot_terminal_harness`, and the global directive here is `warn`. So when this test fails, `dump_child_log` will print an empty stderr — the log capture machinery below is dead weight.

Same applies to `integrations/opencode/marmot/tests/e2e_connector.rs:156` (`warn,wn_opencode=info`), which this PR silently broke. Add `marmot_terminal_harness=info` to both.

Separately: this file is a 390-line copy of the 402-line OpenCode e2e. `ChildGuard`, `WaitValue`, `wait_for`, `send_control_request`, `spawn_wn_agent`, `create_account`, and `dump_child_log` are all harness-agnostic and belong in a shared test-support module.

integrations/terminal-harness/src/control.rs
```diff
@@ -260,7 +260,7 @@ impl ControlClient {
 
     fn next_request_id(&self) -> String {
         let seq = self.request_seq.fetch_add(1, Ordering::Relaxed) + 1;
-        format!("wn-opencode-{}-{seq}", std::process::id())
+        format!("terminal-harness-{}-{seq}", std::process::id())
```
erskingardner: Request ids lose the harness identity here. `wn-opencode-<pid>-<seq>` told you which connector issued a request from the agent side; `terminal-harness-<pid>-<seq>` does not, and the whole point of this PR is that there are now several of these. `config.reply_prefix` is already a per-harness `&'static str` — thread it through and keep the ids distinguishable.

.github/workflows/wn-agent-binaries.yml
```diff
@@ -495,8 +573,13 @@ jobs:
           sed \
             -e "s/^MARMOT_RELEASE_TAG_DEFAULT=.*/MARMOT_RELEASE_TAG_DEFAULT=\"${tag}\"/" \
             -e "s/^WN_AGENT_VERSION_DEFAULT=.*/WN_AGENT_VERSION_DEFAULT=\"${version}\"/" \
-            scripts/install-opencode-marmot.sh > dist/install/install-opencode-marmot.sh
+          scripts/install-opencode-marmot.sh > dist/install/install-opencode-marmot.sh
```
erskingardner: Unrelated re-indent (12 -> 10 spaces) on the existing OpenCode installer line. Harmless inside the block scalar but it is diff noise in an otherwise-additive workflow change — please revert.

integrations/terminal-harness/tests/test_installer.sh
```diff
@@ -0,0 +1,435 @@
...
+timeout_env="${env_prefix}_TIMEOUT_SECS"
+idle_timeout_env="${env_prefix}_IDLE_TIMEOUT_SECS"
+request_timeout_env="${env_prefix}_REQUEST_TIMEOUT_SECS"
+max_reply_env="${env_prefix}_MAX_REPLY_BYTES"
+max_pending_env="${env_prefix}_MAX_PENDING_PER_GROUP"
+export "$bin_env=/bin/echo"
+allow_hex="$(printf '11%.0s' {1..32})"
+
+[ -x "$installer" ]
+bash -n "$installer" "$shared_installer"
```
erskingardner: `bash -n` with two file arguments only parses the first; the second is passed as `$1`. So the shared installer is never actually syntax-checked here — `bash -n "$installer"` and `bash -n "$shared_installer"` need to be two separate calls. In practice every scenario below execs the shared script so a syntax error would still fail loudly, which makes this harmless but slightly misleading.

.github/workflows/wn-agent-binaries.yml
```diff
@@ -231,6 +241,71 @@ jobs:
           path: dist/wn-opencode-${{ matrix.target }}/*
           if-no-files-found: error
 
+      - name: Package wn-pi bundle
```
erskingardner: This is now the third near-verbatim copy of the ~60-line bundle-packaging block (wn-agent, wn-opencode, wn-pi) — the same drift risk this PR eliminated for the installer scripts. Worth extracting into a small `scripts/package-binary-bundle.sh <binary> <target> ...` invoked by all three steps; the manifest heredoc and tar/shasum logic are identical except for the binary name.

integrations/pi/marmot/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+async fn run() -> Result<()> {
+    let (config, backend) = config::Config::from_env()?.into_harness()?;
+    marmot_terminal_harness::run(config, backend).await
+}
+
+fn init_tracing() {
+    tracing_subscriber::fmt()
+        .with_env_filter(
+            tracing_subscriber::EnvFilter::try_from_default_env()
+                .unwrap_or_else(|_| tracing_subscriber::EnvFilter::new("info,wn_pi=info")),
```
erskingardner: The `wn_pi=info` directive is dead: every log statement in this crate and the shared runtime uses the `marmot_terminal_harness` target, so this filter matches nothing (the leading `info` is doing all the work). Same in wn-opencode (`"info,wn_opencode=info"`). Either drop the per-crate directive or replace it with `marmot_terminal_harness=info` so the default filter reflects reality.

Justfile
```diff
@@ -177,6 +177,12 @@ opencode-installer-test:
 opencode-dev-e2e-connector:
     cargo test -p wn-opencode --test e2e_connector -- --ignored --nocapture
 
+pi-installer-test:
```
erskingardner, line 180: These installer tests (and the opencode one above) are not wired into `ci.yml` — they only run when someone remembers the just recipe locally. That was a pre-existing gap for OpenCode, but the installer is now a single shared script whose breakage ships to both connectors via `curl | bash`. A small ci.yml job running `integrations/terminal-harness/tests/test_installer.sh pi` and `... opencode` (mocks only, no network) would close it.

erskingardner (commented):
**Recommendation: merge once CI greens.** Second-round review at `c0f64deb`; all three round-1 blockers are genuinely resolved, not just ticked off.

## What I verified

- **Extraction is behavior-preserving.** I diffed `bridge.rs`, `opencode.rs`, and the installer between `master` and this head, not just the PR diff. Every bridge delta is mechanical parameterization (`reply_prefix`/`display_name`/env names); the OpenCode run loop is structurally identical; the only OpenCode behavior changes are deliberate hardening: fail-closed `$HOME`, zero-timeout rejection, npub length tightened to exactly `{58}`, archive-member traversal guard, `umask 077` on units/plists/env/bootstrap.json, systemd `%`/`$` escaping, relative `--opencode-bin` canonicalization.
- **Round-1 dedup is real.** One parameterized installer + one parameterized installer test with per-kind tables; shared env-config builder with `ConfigSpec`; process/stderr/ANSI plumbing in `marmot-terminal-harness::process`; e2e scaffolding behind a `test-support` feature. `dirs_home()` exists exactly once. The residual copies are the thin per-connector shims, which is the right shape.
- **stdin deadlock fixed.** `write_stdin` runs concurrently with the read loop, and `runner_reads_stdout_while_writing_a_large_prompt` exercises the chatty-backend/large-prompt case that deadlocked before.
- **Pi contract pinned.** `real_pi_0_79_6_contract` (ignored) plus the AGENTS.md note record what was verified against Pi 0.79.6: stdin prompt, version-3 `session` event, `message_end` assistant text, `--session-id` creation under `--session-dir`.

## Non-blocking notes (inline)

1. `bash -n` with two file args only parses the first — trivial.
2. The workflow's wn-pi packaging step is the third copy of the bundle-packaging block; a shared packaging script would match this PR's own anti-drift discipline.
3. Dead `wn_pi=info` / `wn_opencode=info` default filter directives after the `TRACE_TARGET` move.
4. Installer shell tests still aren't in CI; with one shared installer feeding two `curl | bash` entry points, that gap is now worth closing.

## One thing to release-note

`install-opencode-marmot.sh` semantics changed deliberately: `WN_OPENCODE_ALLOWED_SENDERS_HEX` no longer feeds the welcomer allowlist, and `--allow-welcomer` is now strictly required (previously the senders env satisfied it). The new semantics (inviter ≠ prompt sender) are better and are tested, but existing scripted OpenCode installs that set only `WN_OPENCODE_ALLOWED_SENDERS_HEX` will now fail with "at least one --allow-welcomer value is required". Worth a line in the next release notes alongside the `marmot_terminal_harness` target note already in `release.md`.

erskingardner (commented):
## Overall

The premise is right. Adding a second terminal harness by copy-pasting the `wn-opencode` crate would have been the wrong move, and extracting `marmot-terminal-harness` first is the root-cause fix rather than the symptom fix. The `Backend` seam is narrow (command construction + event parsing stay in each connector), the OpenCode-visible defaults are preserved byte-for-byte (`~/.marmot-agents/harnesses`, `wn-agent-harnesses.service`, `~/.local/state/wn-opencode/sessions.json`, the `WN_OPENCODE_ADMIN_HEX` legacy alias), and folding the two 850-line installers into one parameterised script is a genuine net simplification. Pi's parser is correctly conservative — only completed `message_end` assistant `text` parts are forwarded, and the tests pin that thinking/toolCall/toolResult content never escapes.

Nothing here is a hard blocker. Two things I'd like addressed before merge, and a handful of optional cleanups; details inline.

**Worth fixing:**
1. `release.md` states the `wn_opencode` tracing target "continues to cover only OpenCode-specific command and event parsing". It doesn't — after this PR the single `debug!` in `opencode.rs` uses the shared `TRACE_TARGET`, so *zero* events are emitted on `wn_opencode`. Same for `wn_pi`. Operators following that sentence will configure a target that matches nothing.
2. Pi feeds the prompt over stdin, so a Pi process that exits before draining stdin turns a diagnosable failure (exit code + stderr) into a generic "pi failed while streaming its response." This is a narrower diagnostic surface than OpenCode has, on exactly the failure that new users hit most (bad binary, bad flags, unauthenticated).

**Scope note:** shared-runtime extraction + new connector + installer unification + release wiring in one 39-file PR is a lot to hold in your head at once. The extraction is well-protected by keeping the OpenCode tests green, so I'm not asking for a split — but a follow-on of this size would be easier to review as `extract runtime` then `add wn-pi`.

**Minor / optional** (not worth a round-trip on their own):
- `integrations/AGENTS.md` lists `cargo test -p marmot-terminal-harness` twice in the same verification block.
- `integrations/opencode/marmot/Cargo.toml` still requests tokio's `net` and `signal` features; those moved to the harness crate. `wn-pi`'s feature list already omits them.
- `scripts/install-{pi,opencode}-marmot.sh` dereference `${BASH_SOURCE[0]}` unguarded, so `curl .../scripts/install-opencode-marmot.sh | bash` now dies with an `unbound variable` under `set -u` instead of a clear message. The released assets are self-contained so this only bites people pointing at the raw repo path, but a one-line guard would make the failure legible.

integrations/pi/marmot/src/pi.rs
```diff
@@ -0,0 +1,445 @@
...
+                    target: TRACE_TARGET,
+                    method = "pi_run",
+                    error_kind = "json",
+                    "dropping undecodable Pi event"
+                ),
+            }
+            idle_deadline = tokio::time::Instant::now() + invocation.idle_timeout;
+        }
+
+        match timeout_at(idle_deadline, &mut writer_task).await {
```
erskingardner: **Non-blocking, but worth a guard:** a writer-side `EPIPE` is treated as fatal here, which throws away the exit code and stderr that would actually explain the failure.

The stdout loop runs to EOF first, so by the time we get here the child has usually exited. If it exited *without* draining stdin — bad `--session-dir`, unsupported flag, auth failure, anything that makes Pi bail during startup — `write_all`/`shutdown` returns `EPIPE`, the `??` propagates `BackendStream`, and the user sees the generic `[wn-pi] pi failed while streaming its response.` The real exit code (64, say) and the captured stderr are discarded.

Small prompts usually fit the 64 KiB pipe buffer and land before the child exits, so this is racy rather than deterministic — but `MARMOT_MESSAGE_BYTES_CEILING` is 60 000, and you already have `runner_reads_stdout_while_writing_a_large_prompt` exercising a 60 000-byte prompt, so the window is real and reachable.

OpenCode doesn't have this problem because the prompt is argv, so a failing `opencode` still yields an `Outcome` with `exit_code`/`stderr` and the bridge reports `[wn-opencode] opencode reported ...` plus the code.

Suggest reaping the child first and demoting a broken-pipe writer result to a `debug!`, keeping the fatal path for genuine idle/timeout:

```rust
// stdin may legitimately be closed early by a backend that exits before
// draining the prompt; the exit code and stderr are the useful signal.
if let Ok(Ok(Err(_))) = timeout_at(idle_deadline, &mut writer_task).await {
    debug!(target: TRACE_TARGET, method = "pi_run", error_kind = "stdin", "backend closed stdin early");
}
```

A test with a fake `pi` that `exit 64`s without `cat`-ing stdin would pin the behaviour.

release.md
```diff
@@ -195,12 +195,20 @@ Before a WN Agent release, run the normal preflight plus:
...
 bash scripts/install-hermes-marmot.sh --dry-run --yes
 bash scripts/install-openclaw-marmot.sh --dry-run --yes
 bash scripts/install-opencode-marmot.sh --dry-run --yes --allow-welcomer "$(printf '11%.0s' {1..32})" --opencode-bin /bin/echo
+bash scripts/install-pi-marmot.sh --dry-run --yes --allow-welcomer "$(printf '11%.0s' {1..32})" --pi-bin /bin/echo
 ```
 
+Bridge and control logs for both terminal harnesses now use the
+`marmot_terminal_harness` tracing target. Include
+`marmot_terminal_harness=debug` in `RUST_LOG`; the former `wn_opencode` target
+continues to cover only OpenCode-specific command and event parsing.
```
erskingardner: This sentence is not accurate as of this PR, and it's the sentence operators will act on.

`integrations/opencode/marmot/src/opencode.rs` now imports `TRACE_TARGET` from `marmot_terminal_harness` and uses it for its only tracing call (the undecodable-event `debug!` at line 159). Same in `pi.rs` line 112. So after this change **no event anywhere in the workspace is emitted on the `wn_opencode` target** — it doesn't "continue to cover" anything.

Two knock-on effects:

1. Anyone running `RUST_LOG=warn,wn_opencode=debug` today silently loses all harness logging after upgrading. That deserves an explicit upgrade note rather than a sentence implying the old target still works.
2. The default `EnvFilter` fallbacks are now dead directives: `EnvFilter::new("info,wn_opencode=info")` in `wn-opencode/src/main.rs` and `EnvFilter::new("info,wn_pi=info")` in `wn-pi/src/main.rs`. Both reduce to plain `info`. Either point them at `marmot_terminal_harness` or drop the per-crate directive.

Suggested replacement for the sentence:

> `marmot_terminal_harness=debug` in `RUST_LOG`. The `wn_opencode` target no longer emits events; update any existing `RUST_LOG` filters that reference it.

integrations/terminal-harness/src/config.rs
```diff
@@ -0,0 +1,301 @@
...
+    /// Validated shared bridge configuration.
+    pub harness: Config,
+    /// Connector's Marmot home.
+    pub home: PathBuf,
+    /// Backend command or absolute binary path.
+    pub bin: String,
+}
+
+/// Loads shared connector configuration from the process environment.
+pub fn load_config_from_env(spec: ConfigSpec) -> Result<LoadedConfig> {
```
erskingardner: `load_config_from_env` has no callers. Both connectors define their own `Config::from_env` that goes straight to `load_config_with`:

```rust
// opencode/src/config.rs and pi/src/config.rs
pub(crate) fn from_env() -> Result<Self> {
    Self::from_lookup(&mut |name| env::var(name).ok())
}
```

They have to, since each wraps the result in its own type. Dead public API on a `publish = false` crate — delete it and the `use std::env` it's the only user of.

integrations/terminal-harness/src/config.rs
```diff
@@ -0,0 +1,301 @@
...
+    let allowed_senders = parse_hex_csv(&allowed_name, &allowed_raw)?;
+
+    let account_name = env_name("ACCOUNT_ID_HEX");
+    let account_id_hex = lookup(&account_name)
+        .map(|value| (account_name.as_str(), value))
+        .or_else(|| lookup("MARMOT_ACCOUNT_ID_HEX").map(|value| ("MARMOT_ACCOUNT_ID_HEX", value)))
+        .map(|(name, value)| normalize_hex(name, &value))
+        .transpose()?;
+
+    let activation_name = env_name("ACTIVATION");
```
erskingardner, line 105: Behaviour change: the old opencode check trimmed before comparing.

```rust
// before
let activation = lookup("WN_OPENCODE_ACTIVATION").unwrap_or_else(|| "always".to_owned());
if activation.trim() != "always" { ... }
```

Now `WN_OPENCODE_ACTIVATION="always "` (trailing space or newline — easy to get from a hand-edited systemd `Environment=` line or an `EnvironmentFile`) is a hard startup failure. Restoring the trim is a two-character fix:

```rust
if lookup(&activation_name).as_deref().unwrap_or("always").trim() != "always" {
```

(The other validation tightenings in this function — rejecting zero timeouts, requiring `$HOME` — are improvements and worth keeping.)

integrations/terminal-harness/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// Maximum queued prompts for one group.
+    pub max_pending_per_group: usize,
+    /// Private session-mapping file.
+    pub state_path: PathBuf,
+    /// Total backend invocation timeout.
+    pub backend_timeout: Duration,
+    /// Maximum backend stdout silence.
+    pub backend_idle_timeout: Duration,
+    /// Lowercase backend name used in user-visible failures.
+    pub display_name: &'static str,
```
erskingardner: `display_name`, `reply_prefix`, `bin_env_name`, and `account_env_name` are verbatim copies of four `ConfigSpec` fields, threaded through `load_config_with` one assignment at a time and duplicated again in `bridge.rs`'s `test_config` helper.

Holding the spec directly drops four fields here, four assignments in `load_config_with`, and four lines in the test fixture:

```rust
pub struct Config {
    // ...
    /// Connector identity and naming.
    pub spec: ConfigSpec,
}
```

Call sites become `config.spec.reply_prefix` etc. `ConfigSpec` is already `Copy`, so nothing gets more expensive. Not blocking, but it removes a duplication that will drift the moment a third harness lands.

.github/workflows/wn-agent-binaries.yml
erskingardner: This step is a ~60-line verbatim copy of `Package wn-opencode bundle` with `opencode` swapped for `pi` — including the tag-format regex, the workspace-version cross-check, and the manifest heredoc. That's now three near-identical packaging blocks in this file, and harness number four will make it four.

A loop over the harness names collapses both harness steps into one:

```yaml
- name: Package terminal harness bundles
  shell: bash
  run: |
    set -euo pipefail
    for harness in wn-opencode wn-pi; do
      # ...existing body, with $harness for the binary name...
    done
```

The two `Upload` steps still need to stay separate for distinct artifact names, but those are 6 lines each. Optional for this PR — but the duplication is the kind that goes stale asymmetrically (someone fixes the version check in one copy and not the other).

integrations/terminal-harness/tests/test_installer.sh
```diff
@@ -0,0 +1,435 @@
...
+env -u MARMOT_HOME -u MARMOT_AGENT_SOCKET \
+    WN_AGENT_SHA="9.9.9" \
+    MARMOT_RELEASE_TAG="wn-agent-v9.9.9-test" \
+    "$installer" --dry-run --yes --no-service --allow-welcomer not-a-key \
+    >/dev/null 2>&1 || bad_allowlist_status=$?
+[ "$bad_allowlist_status" -ne 0 ]
+
+case "$installer_dry_run" in
+    *"wn-agent-"*"9.9.9.tar.gz"* ) ;;
+    *) echo "pi installer dry-run did not use WN_AGENT_SHA asset suffix for wn-agent" >&2; exit 1;;
```
erskingardner: This file is now shared by both harnesses (`test_installer.sh pi|opencode`), but a good number of the failure messages still say "pi installer" unconditionally — this line and roughly a dozen others below it, plus the `grep -F` guards near the end. When `just opencode-installer-test` fails, it'll report a Pi problem.

The surrounding messages already use `$kind` correctly, so this is just the ones that got missed during the parameterisation. Mechanical fix:

```sh
*) echo "$kind installer dry-run did not use WN_AGENT_SHA asset suffix for wn-agent" >&2; exit 1;;
```

integrations/terminal-harness/src/test_support.rs
```diff
@@ -0,0 +1,343 @@
...
+    AgentControlAccount, AgentControlDebugFinalSend, AgentControlEnvelope, AgentControlRequest,
+    AgentControlResponse, read_envelope, write_frame,
+};
+use tempfile::TempDir;
+use tokio::io::BufReader;
+use tokio::net::UnixStream;
+use tokio::time::sleep;
+
+/// Stable synthetic group id used by connector e2e tests.
+pub const GROUP_ID_HEX: &str = "2222222222222222222222222222222222222222222222222222222222222222";
```
erskingardner: `GROUP_ID_HEX`, `MESSAGE_ID_HEX`, and `INBOUND_TEXT` are `pub` but only used inside this module — the two connector e2e tests import just `HarnessContext`, `MAX_REPLY_BYTES`, `SENDER_ACCOUNT_ID_HEX`, `SpawnedChild`, and `run_connector_e2e`. Same pattern elsewhere in the new public surface: `split_reply_chunks` and `dirs_home` are re-exported from `lib.rs` but have no callers outside the crate.

Since this is `publish = false` and both consumers are in-tree, dropping `pub` on the unused items keeps the API equal to what's actually needed. Easy to re-export later if a third harness wants them.

PR #1245, opened 2026-08-04
Fix/incident replay halt evidence followups

jgmontoya:
**What this fixes**

1. **Fail-closed halt ordering** (`29fe191b` → now `7d2ff4cf`): an untimed halt row alongside a timed one let a newer `join_welcome_repair` clear a halt that must stand — the RED test classified a live halt as `Healthy`. The comparison now fails closed: a halt whose timing cannot be established is never cleared by a later repair, and a repair at the same timestamp cannot clear it either. Both directions are mutation-checked (reverting to the `Option::max` reading fails the mixed-timing fixture; relaxing `repair <= halt` to `<` fails the tie fixture).
2. **Superseded routes become artifact evidence** (`8271d8ff`): the accepted path now records the superseded convergence route in `IncidentScenarioArtifactV1` instead of leaving it as log-only advisory.
3. **Causes ahead of re-assertions** (`9b0daf7f`): `HaltedEngine.reasons` now reports causes; re-assertions appear only when there is no cause.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved incident replay halt handling by prioritizing substantive causes over duplicate re-assertions.
  * Untimed or partially timed halt evidence remains blocking, and repairs at the same timestamp as a halt cannot clear it.
  * Accepted fallback routes now preserve details about superseded convergence attempts and their failure reasons.
  * Improved artifact availability reporting when replay evidence cannot be retained.

* **Documentation**
  * Clarified fallback advisories, artifact persistence, halt prioritization, and halt-clearing behavior.

### DIFF
--- a/crates/incident-replay/AGENTS.md
+++ b/crates/incident-replay/AGENTS.md
@@ -169,14 +169,22 @@ Precedence, highest first:
    stable` alone is the engine's ordinary healthy state (see the rejected designs
    below). Clearing is decided per **`(engine_id, group_ref)`** — `Unrecoverable`
    is per-group state, so a repair in one group must not clear another group's
-   halt — by **newest halt vs. newest repair**, with surviving halts folded back
+   halt — by **repair strictly after the newest halt**, with surviving halts
+   folded back
    per engine so an engine halted in two groups is still one halted engine with
    the union of its reasons. Newest-wins needs no global event ordering because a
    live halt is continuously re-asserted (every session open re-emits
    `hydrate_unrecoverable_group`; every convergence attempt emits
    `already_unrecoverable`), so a halt that outlived its repair leaves rows after
    it. Both rows come from one engine — one clock, one recorder file — so rule 6's
-   cross-device grace does not apply here. It therefore outranks rule 6, which it usually
+   cross-device grace does not apply here. Two boundaries of that comparison are
+   contract, not incident: a repair sharing the halt's millisecond does **not**
+   clear it (a tie orders nothing, and the two orders mean opposite things), and
+   **one untimed halt row makes the whole halt side unorderable** — not just that
+   row. The untimed row may be the newest halt evidence there is, so the newest
+   *timed* halt is only a lower bound on the halt's position and a repair after
+   that bound proves nothing; taking it as the halt's position is the fail-open
+   direction, clearing a live halt on partially instrumented input. It therefore outranks rule 6, which it usually
    explains: on the real 26a9f546 export the halted engine is exactly the one
    rule 6 labelled `went dark`, and reporting the inference over the diagnosis
    would send the operator to re-pull for a confirmation they already have.
--- a/crates/incident-replay/AGENTS.md
+++ b/crates/incident-replay/AGENTS.md
@@ -108,8 +108,16 @@ incident becomes a vector only if the simulator reproduces the recorded outcome
     only *after* a failure — and one rule keeps it honest: a lower route may
     override a higher route's quarantine only by producing an accepted vector. A
     second quarantine never displaces the higher-precedence one; it becomes an
-    `advisory (fallback route):` instead. Vector names live here too: they are
-    pipeline facts, not CLI ones.
+    `advisory (fallback route):` instead. When the fall-through *does* produce an
+    accepted vector, the superseded finding is also written into that artifact's
+    `unavailable_fields` as `contested_convergence_replay`, with the same sentence
+    the advisory carries: the advisory is stdout, the artifact is what gets
+    persisted and read later as evidence, and an envelope showing a clean accepted
+    archetype with no trace of the higher-precedence incident that shared the
+    export fails the standard `archetype` already enforces on itself. The
+    `fallback route` finding has no such home by construction — both routes failed
+    closed, so there is no accepted artifact — and the advisory line is its only
+    surface. Vector names live here too: they are pipeline facts, not CLI ones.
 - **Module:** `src/main.rs`
   - **Role:** CLI — read one export file, detect its format, print the `Routing`.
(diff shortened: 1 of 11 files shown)


crates/incident-replay/src/export.rs
```diff
@@ -44,6 +44,29 @@ const VERIFIED_REPAIR_REASON: &str = "join_welcome_repair";
...
+///
+/// `already_unrecoverable` is emitted on every convergence attempt against an
+/// already-halted group (`cgka-engine/src/distributed_convergence.rs`) and
+/// `hydrate_unrecoverable_group` on every session open over one
+/// (`cgka-engine/src/engine.rs`). Both are the halt restating itself — the
+/// property rule 5's newest-wins clearing relies on — so they arm the gate like
+/// any other halt reason. Neither says why the engine stopped. See
+/// [`is_halt_re_assertion`] for what that means for reporting.
+const HALT_RE_ASSERTION_REASONS: [&str; 2] =
+    ["already_unrecoverable", "hydrate_unrecoverable_group"];
```
erskingardner, line 57: `unspecified` is not a cause either, but it is not in this list, so it wins the promotion.

`unrecoverable_halt_reason` substitutes `UNSPECIFIED_HALT_REASON` for a halt row that carried neither `reason` nor `error_kind` (both arms, `export.rs:376` and `:383`). An export with one such row plus any number of `hydrate_unrecoverable_group` / `already_unrecoverable` rows on the same engine now reports `engine-a (unspecified)`: `causes` is non-empty, so the re-assertions are dropped, and the operator is left with the one string that names neither a cause nor the surface it came from. Before this change they at least saw which surface re-asserted the halt.

That is the fail-open direction of the same argument the rest of the PR makes carefully in the other direction. Adding `UNSPECIFIED_HALT_REASON` to `HALT_RE_ASSERTION_REASONS` — or filtering it alongside them — puts it back in the fallback tier where it belongs: reported when it is all there is, suppressed when a real cause exists. The name of the constant would then be doing double duty, so `HALT_NON_CAUSE_REASONS` (or similar) reads better.

crates/incident-replay/AGENTS.md
```diff
@@ -169,14 +189,22 @@ Precedence, highest first:
    stable` alone is the engine's ordinary healthy state (see the rejected designs
    below). Clearing is decided per **`(engine_id, group_ref)`** — `Unrecoverable`
    is per-group state, so a repair in one group must not clear another group's
-   halt — by **newest halt vs. newest repair**, with surviving halts folded back
+   halt — by **repair strictly after the newest halt**, with surviving halts
+   folded back
```
erskingardner, line 193: Stray re-wrap: `folded back` is orphaned on its own 14-character line, and line 207 (`direction, clearing a live halt ... which it usually`) runs to 114 characters. Everything around them wraps at ~80. Re-flowing the paragraph fixes both.

erskingardner (approved):
Logic changes are correct and well covered; CI is green. Approving with three non-blocking notes.

The three behaviour changes each do what the description claims, and each has a fixture that fails if the rule is relaxed in either direction:

- `halt_position_ms` fails closed on mixed timing, and `repair <= halt` fails closed on a tie. Untimed repairs were already ignored by the `Option::max` fold, so the halt side is now symmetric with it.
- `reported_halt_reasons` only reshapes reporting; the gate still arms on any halt reason, and the re-assertion-only path keeps the engine named.
- Pushing `contested_convergence_replay` into `unavailable_fields` is safe: nothing gates on field names (`src/main.rs` only prints them), the entry can be added at most once on this path, and it is the only accepted-artifact path that can be superseded.

Non-blocking:

1. `UNSPECIFIED_HALT_REASON` is treated as a cause by the new filter (inline).
2. Two AGENTS.md wrapping defects introduced by the edit (inline).
3. Ratio. 401 added lines carry roughly 25 lines of logic. The untimed-halt argument is now stated four times in near-identical prose — `AGENTS.md` rule 5, `HaltLifecycle::halt_position_ms`, `still_halted`, and the test comment in `one_untimed_halt_row_among_timed_ones_still_blocks_a_later_repair` — and the superseded-artifact argument three times (`AGENTS.md`, the inline comment in `fall_through_to_fork`, the test comment). Four copies of an argument drift apart under later edits, and the one that ends up wrong is not obvious. The doc comment on the function that implements the rule is the copy worth keeping; the AGENTS.md and test versions could be a sentence each pointing at it. Same for the `Err` arm comment in `fall_through_to_fork`, which explains why nothing happens on a path where nothing happens.

crates/incident-replay/src/export.rs
jgmontoya: Confirmed, and you're right that it's the fail-open direction of exactly the argument this PR makes the other way. Verified the mechanics: both unrecoverable_halt_reason arms substitute UNSPECIFIED_HALT_REASON (export.rs:376/:383), it's absent from HALT_RE_ASSERTION_REASONS, so one reason-less halt row plus any number of re-assertions reports engine-a (unspecified) — strictly less information than the pre-PR output, which at least named the re-asserting surface. Taking your fix as stated: unspecified joins the non-cause tier (reported when it's all there is, suppressed when a real cause exists) and the constant becomes HALT_NON_CAUSE_REASONS, since re-assertion no longer describes all its members. To keep this approval intact I'm landing it on the follow-up branch already stacked on this PR rather than pushing here — it merges right behind this one, with a mixed-rows fixture pinning the fallback tier both ways.

crates/incident-replay/AGENTS.md
jgmontoya: Good catch — the orphaned folded back and the 114-char line are both re-wrap debris from the rule-5 edit. Re-flowing the paragraph rides the stacked follow-up branch alongside the unspecified fix, so this approval stays intact.

PR #1246, opened 2026-08-04
Feat/queue and recover nonstable sends

### LINKED ISSUE #1177: Retain app-message intent during PendingPublish without preparing ciphertext
## Problem

An app-message send attempted while a group is `PendingPublish` is rejected, and the explicit `queue_app_message` path is rejected by the same Stable-state check. The Stable-only encryption gate is correct, but the user's plaintext intent is not retained for later preparation.

Grounded against `origin/master` at `b609ccd3796aea14bf0f0478073520179820d999`.

## Runtime evidence

Fresh forensic exports contain four local-user app-message failures across two groups while `PendingPublish`:

- two direct sends: `illegal send transition from PendingPublish: send requires Stable`;
- two explicit queue attempts: `illegal queue_app_message transition from PendingPublish: queue_app_message requires Stable`.

These were intent-rejection failures, not proof that a newly prepared ciphertext failed publication. Android subsequently rendered the persisted failures as red tombstones; that presentation is already tracked by `whitenoise-android#1747`.

## Source-confirmed mechanism

- [`do_send`](https://github.com/marmot-protocol/mdk/blob/b609ccd3796aea14bf0f0478073520179820d999/crates/cgka-engine/src/message_processor/mod.rs#L190-L203) asks `should_queue_outbound_intent` before dispatch.
- [`should_queue_outbound_intent`](https://github.com/marmot-protocol/mdk/blob/b609ccd3796aea14bf0f0478073520179820d999/crates/cgka-engine/src/message_processor/mod.rs#L469-L477) returns `false` for every non-Stable epoch state, bypassing the durable queue.
- [`do_queue_app_message`](https://github.com/marmot-protocol/mdk/blob/b609ccd3796aea14bf0f0478073520179820d999/crates/cgka-engine/src/message_processor/mod.rs#L205-L228) independently rejects non-Stable state.
- [`do_send_app_message`](https://github.com/marmot-protocol/mdk/blob/b609ccd3796aea14bf0f0478073520179820d999/crates/cgka-engine/src/message_processor/send.rs#L825-L850) correctly prevents ciphertext preparation unless Stable.
- The existing durable intent primitive is [`queue_outbound_intent`](https://github.com/marmot-protocol/mdk/blob/b609ccd3796aea14bf0f0478073520179820d999/crates/cgka-engine/src/message_processor/mod.rs#L1118-L1145).

## Narrow design contract

Scope this change to **application-message intent during `PendingPublish`**. Do not generalize it to `Merging`, `Recovering`, `Unrecoverable`, removal, disband, leave, or hydration quarantine without separate evidence and lifecycle design.

- Passing terminal acceptance gates may persist the app-message intent without preparing MLS ciphertext.
- Return a typed accepted-pending/queued outcome to the app boundary.
- After `confirm_published` or `publish_failed` returns the group to `Stable`, prepare the event once against the resulting canonical epoch and drain it through the existing durable queue.
- Never create ciphertext, stage another commit, or publish the queued app message while non-Stable.
- Preserve one logical local projection/identity across acceptance, restart, regeneration, and publication.
- Delete the intent only after accepted publication or an explicit terminal disposition.

## Acceptance criteria

- Direct app send and explicit queue during `PendingPublish` retain one durable intent and return a typed pending result; neither prepares ciphertext.
- Confirmation drains the same intent at the confirmed epoch exactly once.
- Rollback drains the same intent at the restored prior epoch exactly once.
- Restart in `PendingPublish` preserves both the staged publish and queued app intent without reordering or duplication.
- Authoritative removal, disband, sign-out, quarantine, and verified-Unrecoverable transitions purge or terminally reject queued work according to existing policy.
- The app/runtime can distinguish `accepted_pending`, `published`, and `terminal_failure` without inferring status from an empty message-id list or raw error text.
- Existing transport-inactive retention from PR #1158 remains unchanged.

## Test coverage

- Add direct-send and `queue_app_message` cases to `crates/cgka-engine/tests/publish_lifecycle.rs` for confirm, rollback, restart, and terminal transitions.
- Add `marmot-app`/UniFFI tests proving one pending projection becomes one published projection and exposes a typed outcome.
- Add ordering coverage for multiple queued app intents behind one pending group-evolution commit.

## Related / non-goals

- PR #1158 retains sends when the account transport is inactive while local group state is otherwise safe; it does not cover `PendingPublish`.
- #1020 established durable fanout and publish-before-apply semantics; this issue must not weaken them.
- `whitenoise-android#1216` is the broader offline/outbox tracker.
- `whitenoise-android#1747` covers reason-aware recovery of text from persisted local failures.
- Relay retry cadence and ambiguous publish recovery are not changed here.

jgmontoya:
Implements the retention half of the queue-and-recover work from the field incident (Android tombstones: illegal queue_app_message transition from PendingPublish, two user sends lost while a self-update publication stalled). Claim L2 in the convergence-reliability plan; scoped per #1177.

What ships

- App messages sent while a group is resolving a locally staged publication — PendingPublish (owes a transport outcome) or Merging (owes the local merge) — are retained durably and return SendResult::Queued (new EpochState::is_resolving_local_publish() / is_terminal() in traits). Recovering awaits a convergence decision over remote branches, a separate lifecycle question deliberately out of scope. Terminal states refuse new work; work already accepted holds for the one legal exit — a test pins that a pre-halt intent survives a durable Unrecoverable, a verified Welcome repair, and a restart, then drains re-encrypted at the post-repair epoch with the peer decrypting it.
- Retention stores the intent, so the drain regenerates and encrypts at drain-time epoch — the load-bearing test asserts this across an epoch change.
- The publish outcome now schedules the retained-intent drain (nothing else does in a running process — the idle timer is a never-fire sentinel), including when the retained-intent read fails: a failed read cannot prove the queue is empty.
- The two tests that pinned the old rejection behavior are flipped and renamed to assert retention-and-release. Group-state intents keep the strict Stable requirement, with the SendIntent::Disband exception documented (it persists an irreversible request; its Commit stays Stable-gated).

Scheduling cost note: the scheduling read runs on the publish-outcome path, which is per group-state change, not per message — only PublishWork::GroupEvolution, GroupCreated, and AutoPublish carry a PendingStateRef; PublishWork::ApplicationMessage never reaches confirm_published/publish_failed.

What was deliberately NOT shipped, and why

The originally planned absolute PendingPublish deadline (forcing rollback) was disproven during implementation: a publication every endpoint rejected already rolls back immediately on the first complete attempt — the only publications that stall are ambiguous ones (possible exposure), which the multi-step invariant forbids rolling back. A deadline would trade a stall for a fork. The regression test pinning the bound that does exist (immediate rollback + immediate re-stage) is included; the genuinely unbounded residual (an ambiguous publication whose fanout row is deleted at 24 h while the obligation still needs it) is a pre-existing torn-write, filed separately.

Known window (accepted, documented): between a publish outcome and the scheduled drain (~1.1 s), a fresh send on the now-Stable group publishes before the retained payload. Nothing user-visible reorders — timeline_at is stamped at the user action, before the queue-vs-publish decision, and per-sender FIFO is not a spec promise.

Follow-up (own PR): per-intent terminal disposition — a group that becomes Unrecoverable holds its accepted work for repair, but the app currently gets no signal (GroupUnrecoverable isn't dispatched; the optimistic row stays "sending"). It will close #1177's accepted_pending/terminal_failure criterion.

## Summary by CodeRabbit

* **New Features**
  * Application messages sent while a group is awaiting resolution are retained and automatically processed after convergence.
  * Retained messages are regenerated and encrypted using the confirmed epoch.
  * Pending messages continue draining after successful or failed publication outcomes.

* **Bug Fixes**
  * Application messages are rejected only in terminal group states.
  * Rejected self-updates correctly restore the previous state and can be retried.
  * Temporary queue-read failures no longer prevent pending messages from being scheduled for processing.

* **Documentation**
  * Updated guidance for queued application messages and group-state handling.


erskingardner (commented):
The premise is right and the load-bearing test is the right one: retaining the *intent* and asserting it encrypts at the drain-time epoch (with bob actually decrypting it) is exactly the property that makes this safe. Disproving the absolute `PendingPublish` deadline rather than shipping it is the best part of the PR. CI is green at 7ef1959 and the state-machine partition test is a good guard against a future variant joining a class by accident.

One substantive question before merge, plus three smaller notes. Nothing here is a correctness bug in the happy path.

**1. Retention has no terminal-transition exit (the `Recovering` question you flagged).** `Recovering` and `Merging` can resolve to `Unrecoverable`, not just to `Stable`, and nothing purges or fails the retained intents on that transition. Disband purges the queue in the same transaction as the tombstone, and the removed-copy path discards explicitly — the unrecoverable path does neither, and the drain early-returns before reaching the queue. So a payload accepted as `SendResult::Queued` goes silent with no failure signal, which is the same user-visible symptom retention was meant to remove, relocated. Before this PR that send was refused synchronously. Details inline.

**2. The drain-scheduling helper may be able to be three lines.** Its error branch already does the unconditional thing, and the schedule itself is a `HashSet` insert — so the storage read only buys skipping a no-op convergence pass. Worth saying whether that trade is measured.

**3. The new `Unrecoverable` assertion doesn't reach the new branch** — `validate_send_acceptance` refuses first, with the same `from`.

**4. Self-message ordering.** Between the confirm and the runtime's next drain, a fresh `send` goes straight through `do_send_ready` while the retained one still sits in the durable queue, so a user's own messages can leave out of order. Pre-existing for the convergence-unsettled queue path, but this PR makes that window a routine occurrence rather than a rare one. If that's acceptable, it's worth a sentence in the spec section; if not, `do_send` needs a "queue non-empty ⇒ queue" guard.

crates/traits/src/engine_state.rs
```diff
@@ -266,6 +266,34 @@ impl EpochState {
...
 
+    /// Whether this state is a step whose exit is already owed by another
+    /// party: `PendingPublish` awaits a publish outcome, `Merging` awaits its
+    /// merge, and `Recovering` awaits a convergence decision.
+    ///
+    /// `Stable` awaits nothing, and `Unrecoverable` / `Disbanded` are terminal:
+    /// no outcome is owed, so work held for them would be held forever. This is
+    /// the boundary outbound application work is retained across — see
+    /// `CgkaEngine::send`.
+    pub fn is_awaiting_resolution(&self) -> bool {
```
erskingardner: Including `Recovering` here is the part I'd push back on, for a reason beyond "no end-to-end test": `Recovering` and `Merging` do not only resolve to `Stable`. `mark_unrecoverable` (`distributed_convergence.rs:922`, `:1279`) is reachable from convergence, and unlike the disband path — which deletes every `queued_outbound_intent` in the same transaction as the tombstone — and unlike `discard_queued_outbound_intents_for_removed_group`, nothing purges or fails the retained intents when a group goes `Unrecoverable`. `converge_and_drain_queued_outbound_intents` returns `Ok(vec![])` at the `sync_unrecoverable_halt_from_storage` gate before it ever reaches the queue.

Net: the user's message is accepted as `Queued`, then held until a verified repair that may never come, with no event telling the app it will not be delivered. That is the "message vanishes" symptom the PR is fixing, moved one state to the right.

Two ways out, either is fine:
- Narrow to `PendingPublish | Merging` (the one-word change you offered). `PendingPublish` is the state with a guaranteed bounded exit (confirm or rollback, both of which this PR now schedules), and it is the one the field incidents actually hit.
- Keep `Recovering`, and purge-or-signal on the terminal transition, so the acceptance is not a promise the engine cannot keep.

The doc comment says terminal states are excluded because "work held for them would be held forever" — that reasoning applies just as much to work held for a state that *becomes* terminal.

crates/cgka-engine/src/publish.rs
```diff
@@ -484,6 +486,41 @@ impl<S: StorageProvider> Engine<S> {
...
+    ///
+    /// Best-effort by construction: this runs after the outcome is durable, so
+    /// a transient backend lock here must not fail an already-committed
+    /// confirm/rollback. Skipping the schedule is not free, though: the intents
+    /// stay durable, but their release then waits until unrelated traffic
+    /// happens to schedule this group or the process restarts (session-open
+    /// hydration re-arms from the same durable queue). So a failed read — which
+    /// cannot prove the queue is empty — schedules the drain anyway; a spurious
+    /// no-op pass is cheaper than a stranded payload.
+    fn schedule_drain_for_retained_outbound_intents(&mut self, group_id: &GroupId) {
```
erskingardner: Since the `Err` branch schedules anyway, and `schedule_pending_convergence_group` is just a `HashSet` insert, this whole method is behaviourally equivalent to:

```rust
fn schedule_drain_for_retained_outbound_intents(&mut self, group_id: &GroupId) {
    self.schedule_pending_convergence_group(group_id);
}
```

...modulo one no-op convergence pass per publish outcome on groups with an empty queue. That trades ~35 lines and the indexed storage read you flag in the PR description (on the publish hot path) against a drain pass that early-returns cheaply for a settled group. Have you measured that a spurious pass is actually more expensive than the read? If not, the three-line version is the smaller change with strictly fewer failure modes to document — and the fault-injection test then becomes unnecessary rather than load-bearing.

If the read stays, note that in `publish_failed` this now overlaps two existing schedules a few lines up (`:474` for `Disband`, `:477` for `queued_intent_by_pending`), so the rollback path can do three schedules and a storage read to insert one key.

crates/cgka-engine/tests/distributed_convergence.rs
```diff
@@ -3674,6 +3674,26 @@ async fn unrecoverable_halt_survives_engine_restart_until_verified_repair() {
...
+            group_id.clone(),
+            app_payload_for(&restarted, b"must not be retained while halted"),
+        )
+        .await
+        .expect_err("queue_app_message must refuse Unrecoverable");
+    assert!(
+        matches!(
+            queue_err,
+            cgka_traits::error::EngineError::InvalidTransition(ref t)
+                if t.from == "Unrecoverable"
```
erskingardner: This doesn't pin what the comment says it pins. `do_queue_app_message` calls `validate_send_acceptance` first, and `sync_unrecoverable_halt_from_storage` returns `true` whenever `epoch_manager.is_unrecoverable(group_id)` — so it already returns `InvalidTransition { from: "Unrecoverable", .. }` before the new `state.is_terminal()` check is reached. Matching only on `from` means this assertion passes identically against the pre-change `requires Stable` code, so it can't detect the boundary regressing.

Assert on `t.reason` (or `t.to`) instead, so the test distinguishes which gate fired.

Related: is the new `is_terminal()` check in `do_queue_app_message` reachable at all? `Unrecoverable` is caught by the halt sync above, and `Disbanded` by the `disband_tombstone` check (and disband purges the intent queue in its own transaction). If it's deliberate defense-in-depth against state/marker divergence, the comment should say so; otherwise it reads as a guard that can never fire.

crates/marmot-account/tests/runtime.rs
```diff
@@ -3576,3 +3576,126 @@ async fn fanout_staging_failure_rolls_back_pending_mls_state() {
         ))
     ));
 }
+
+#[tokio::test]
+async fn rejected_self_update_publication_rolls_back_instead_of_holding_pending_publish() {
```
erskingardner, line 3581: Non-blocking, scope: this ~125-line test is the single largest addition in the PR and it pins behaviour this PR does not change — it's the artifact of the deadline investigation, not of the retention fix. It's genuinely valuable (the counterpart pairing with the ambiguous-exposure test is the right documentation of why no deadline shipped), but it would read better as its own PR so the retention change stays surgical.

If it stays, the `for _ in 0..8` loop stepping `wall`/`monotonic` by fixed 300s increments until the epoch advances is the shakiest part: it passes as long as jittered backoff lands inside 8×300s, which is a property of the backoff constants rather than of anything asserted here. If the jitter ceiling ever moves, this fails as a flake with a confusing message. Consider reading the obligation's `not_before` and setting the clock to it (as the code above already does once) rather than stepping blind.

Datawav:

## Actionable finding

- **P2 — The queue cap is not enforced atomically.** In `crates/cgka-engine/src/message_processor/mod.rs`, `queue_outbound_intent` separately lists existing rows, checks `len() < 256`, and then inserts. Multiple engine instances sharing the same storage can each observe 255 rows and both insert, exceeding the documented durable bound. They may also derive the same `intent_id` when timestamp, count, and payload coincide, producing a storage error or collision. The sequential capacity test does not cover this race. Move count-and-insert into one atomic storage operation/transaction and add a concurrent shared-storage regression test.

The retention feature is needed: application messages submitted during `PendingPublish` or `Merging` would otherwise be rejected while a locally staged publication awaits resolution. Retaining the intent and regenerating it after resolution correctly encrypts it at the resulting epoch.

The 256-intent bound is also justified because ambiguous publications can remain unresolved indefinitely. A typed refusal preserves already-accepted work and prevents unbounded durable growth.

The rejected alternative—automatically rolling back ambiguous publications after a deadline—would risk a fork if the publication was actually exposed. An unbounded queue would preserve availability but permit uncontrolled storage growth. The chosen bounded queue is preferable, provided its bound is made transactional.

_Advisory audit of head [`657db9d6da19`](https://github.com/marmot-protocol/mdk/commit/657db9d6da19e778f6763224e484baa67cdc994f); no approval, merge, push, checkout, or PR-code execution was performed._

crates/traits/src/engine_state.rs
jgmontoya: Your mechanics all check out, but two premises don't survive tracing: `Merging` is never observable (`confirm_publish` builds and consumes it in adjacent lines, `epoch_manager.rs:255-259`), and narrowing wouldn't close the hole anyway — a `PendingPublish`-retained intent outlives the state (buffered ingest → fork on replay → `Unrecoverable`, synchronously, drain still ~1.1 s out; already reachable on master via the older queue path). Purging would also reverse #1106's pinned policy that halted queued work must remain available *for* verified repair — so instead of choosing, we ran the experiment: retain → durable halt → Welcome repair (epoch 1→3) → restart → the pre-halt intent drains re-encrypted at the post-repair epoch and the peer decrypts it. GREEN (`0afab4f2`), mutation-checked both ways. Shipped in `6d7ae869`: `Recovering` dropped per your first option (#1177 scoped this to `PendingPublish`), predicate renamed `is_resolving_local_publish` to match, and the two rationales that contradicted #1106 rewritten — terminal states refuse *new* work; accepted work holds for the one legal exit. The real gap behind your "no event" point is #1177's unmet terminal-disposition criterion (`GroupUnrecoverable` fires but the app dispatch table drops it; the optimistic row stays "sending" forever) — proposing that as its own PR.

crates/cgka-engine/src/publish.rs
jgmontoya: We tried to take your version and the conformance oracle vetoed it: `empty_creation_confirms_without_synthetic_outbound_work` panics ("a no-op publication must not strand pending MLS state") because the unconditional schedule makes the settled-work snapshot report phantom work for empty-queue groups; it also kills the drain test's "not drainable yet" assertion. Keeping the guarded helper; the rejected patch is saved if you'd rather change the oracle — your call, not a review-fix. Two record corrections while we're here: your overlap point is right and stronger than stated (when `:477` fires the helper's read is redundant by construction), but the app-side consumer isn't a cheap early-return — it's `convergence_schedule_state`, which answers `Idle` with 11 reads including two unbounded `EpochId(0)` scans; that probe is the real cost and gets its own issue. Also correcting our own description: this isn't a hot path (`PublishWork::ApplicationMessage` never reaches these methods — per group-state change, not per message).

crates/cgka-engine/tests/distributed_convergence.rs
jgmontoya: Confirmed by probe: a `panic!` in the terminal-gate branch survives the whole suite — the branch is dead (halt sync pre-empts `Unrecoverable` with the same predicate; the write-once tombstone pre-empts `Disbanded`, and we verified the hydration divergence window is shut). Since no assertion can reach a dead branch, the test now pins the halt gate's `to == "app_message"` + reason so a gate reordering fails loudly, with the comment rewritten to claim exactly that (`5e665f59`). The gate keeps a defense-in-depth label per the `converge_and_drain` precedent.

crates/marmot-account/tests/runtime.rs
jgmontoya: It's worse than shaky — it's inert: instrumented, the loop exits on iteration 1, and the test passes with zero clock advance (the re-staged commit gets a fresh fanout with unattempted targets, so the 30s→1h backoff gates nothing; no jitter exists — the test RNG is a `fetch_add` counter). Your `not_before` fix can't apply: at that point it's a stale past timestamp, and the `Retry`/`PendingPublication` arms never touch it. Fixed by deleting the loop and asserting the immediate re-stage (`92b7ec86`, mutation-checked non-vacuous). Keeping the test in this PR — it's the executable form of the "why no deadline shipped" section; happy to split if you feel strongly.

jgmontoya:
> ## Automated multi-agent audit
> ## Final audit
> ### [P1] Bound retained messages during unresolved publication
> 
> `crates/cgka-engine/src/message_processor/mod.rs` (`should_queue_outbound_intent`, around the `@@ -577` hunk) now accepts and durably stores every application message received during `PendingPublish` or `Merging`. The queue implementation has no byte/count quota, expiry, or backpressure, while an ambiguous publication may remain unresolved indefinitely. Continued sends can therefore exhaust local storage while callers keep receiving `SendResult::Queued`.
> 
> This is introduced by expanding retention into potentially unbounded publication states; the post-outcome drain does not help when no definitive outcome arrives.
> 
> **Needed:** Apply a per-group count or byte limit and return an explicit retryable/backpressure error once reached. Test a permanently unresolved `PendingPublish` that fills the limit.
> 
> **Design:** Prefer a byte limit because payload sizes vary. Reserve capacity atomically with insertion so concurrent sends cannot exceed the bound.
> 
> **Alternatives:** A count limit is simpler but provides weak storage protection. Expiry would bound growth but silently conflicts with the stated durable-retention guarantee unless expiration is surfaced explicitly to callers.
> 
> _Advisory only; no approval, merge, push, checkout, or PR-code execution was performed._

Confirmed the premise: no count, byte, expiry, or backpressure bound exists at any layer — the schema has no CHECK, OutboundIntentStorage exposes no count or prune method, the insert is an unconditional upsert, and created_at_ms never appears in a WHERE clause. Two corrections, though. Merging is unobservable: it's constructed only inside confirm_publish and consumed by merge_to_stable in the same expression, so it never lands in the state map and no send can see it — that half is moot. And this isn't introduced here: master already queued every intent kind durably on the Stable-but-unsettled path with the same absent quota, so the PR narrows kinds and widens states rather than creating the hole — though it does genuinely extend the worst case, since an ambiguous PendingPublish loses its only retry driver at TRANSPORT_FANOUT_RETENTION_SECS and then never resolves.

Took the bound in this PR since a per-group cap closes the older hole for free (5194118d): ledger E9, MAX_QUEUED_OUTBOUND_INTENTS_PER_GROUP = 256 matching E3, enforced on the depth queue_outbound_intent already reads to salt the intent-id hash — one comparison, no new query — returning a typed EngineError::QueuedOutboundAtCapacity (FFI: GroupSendQueueFull) that is deliberately not is_transient(), since that predicate feeds automatic retry loops and a full queue can't clear until the group resolves. The test pins a permanently unresolved PendingPublish filling all 256 slots, the 257th send refused with the queue intact, and the full drain after resolution.

Count rather than bytes: the app already persisted the same payload before the engine call, so the queue's marginal footprint scales with count, not size — and with no payload limit anywhere in the system, a byte cap here would make the retention queue the only message-size rule, refusing a large-but-legitimate first message that sends fine when Stable. That missing payload bound is real and filed separately at the send boundary, where it applies in every state and makes count × max a true byte bound. Skipped the atomic reservation: send takes &mut self up through the account runtime and acquire_account_session admits one owner per account, so the read-and-insert is already serialized — the race isn't representable. Agreed on rejecting expiry: it would silently delete exactly the work the repair-and-drain test pins as durable.

crates/traits/src/engine_state.rs
```diff
@@ -266,6 +271,42 @@ impl EpochState {
...
+    /// own.
+    ///
+    /// Terminal means *no new work is accepted*, not that accepted work is
+    /// discarded. Intents retained before the halt persist for the one legal
+    /// exit — a verified repair returns an `Unrecoverable` group to `Stable`,
+    /// and because retention stores the intent rather than ciphertext, the drain
+    /// re-prepares them under the post-repair epoch (mdk#1106). `Disbanded` is
+    /// the exception, and only because its Commit purges the queue in the same
+    /// transaction that writes the tombstone.
+    pub fn is_terminal(&self) -> bool {
```
erskingardner: `is_terminal()` has no production caller. Grepping the diff, the only three occurrences are this definition and two uses inside `only_a_locally_staged_publish_retains_outbound_work` in the same file — a new public method on a trait-crate type that exists solely to be asserted about by its own unit test.

The interesting part is that it looks like it *was* meant to have one. In `message_processor/mod.rs`, `should_queue_outbound_intent` carries the comment "Excluding the terminal states here is defense in depth", but the code excludes them only implicitly, by `is_resolving_local_publish()` happening not to match them. The predicate the comment describes is this method, and it isn't called.

Either wire it in where the comment claims the exclusion happens, or drop it and let the test assert on `is_resolving_local_publish` alone. Public API added on speculation is harder to remove later than to not add now.

crates/cgka-engine/src/publish.rs
erskingardner: This method reads the entire retained queue (up to 256 rows, each deserialized by `list_queued_outbound_intents`) on **every** publish outcome, in order to decide something it already decides unconditionally on the error path: schedule the drain.

Two things follow.

**1. It could be deleted.** Both call sites become `self.schedule_pending_convergence_group(&group_id);`. That removes this method and its 17-line doc comment, the storage read on the hot publish-outcome path, the `queued_intent_list_fault` field threaded through four `FaultStorage` construction sites, and the ~65-line `an_unreadable_intent_queue_still_schedules_the_drain` test — which only exists to prove the fallback that would then be the sole behavior. The cost is one extra convergence pass per publish outcome on a group with an empty queue. Note `do_publish_failed` already schedules unconditionally two lines above this call for the `Disband` and `queued_intent_by_pending` cases, so that cost is partly already being paid.

**2. The doc's invariant is asserted, not enforced.** "A publish outcome is the only exit from `PendingPublish`" is what makes conditional scheduling safe. But `EpochManager::set_stable` refuses only `Unrecoverable`/`Disbanded` — it will happily overwrite a `PendingPublish` — and it has four callers outside the publish path: `fork_recovery.rs` (`CandidateWins`), `distributed_convergence.rs` (selected-tip apply), `message_processor/ingest.rs`, and `engine.rs`. If any of those can fire on a group holding a staged publication, a retained payload stays unscheduled until unrelated traffic touches the group or the process restarts — precisely the "message never sent" symptom this PR removes.

I could not rule that out from reading alone, which is the argument for (1): an unconditional schedule does not depend on the invariant being true.

erskingardner (commented):
Reviewed the retention change against #1177 and the plan claim it implements.

**The premise holds.** The mechanism in #1177 is confirmed in the diff's own context: `should_queue_outbound_intent` returned `false` for every non-`Stable` state, so a `PendingPublish` send fell through to `do_send_ready` and died on the Stable gate. Retaining the *intent* rather than ciphertext is the right shape — it is what makes the post-repair and post-confirm re-encryption tests possible at all, and `retained_app_message_encrypts_under_the_drain_time_epoch` earns its place as the load-bearing test. Narrowing retention to `SendIntent::AppMessage` while keeping group-state intents `Stable`-gated matches the issue's "do not generalize" instruction. Deciding *not* to ship the absolute `PendingPublish` deadline, and saying why (a deadline trades a stall for a fork on an ambiguous publication), is the most valuable paragraph in the PR body.

**Production diff is genuinely small** — roughly 60 lines across `message_processor/mod.rs`, `publish.rs`, `engine_state.rs`, `error.rs` — with the remaining ~1100 lines being tests, docs, and ledger bookkeeping. That ratio is the right way round.

No blocking correctness defect found. Four non-blocking notes inline, and one general one:

**Comment density.** 28 lines of doc comment for `const MAX_QUEUED_OUTBOUND_INTENTS_PER_GROUP: usize = 256`, 34 for a 15-line function, 26 for a two-arm `matches!`. Several of these paragraphs are re-litigating a design debate from the PR thread rather than telling the next reader what the code does; the reliability plan (E9 row) and the spec already carry that argument. Prose that duplicates an argument recorded elsewhere goes stale independently of the code it sits above.

**Scope.** `rejected_self_update_publication_rolls_back_instead_of_holding_pending_publish` (127 lines in `marmot-account/tests/runtime.rs`) pins pre-existing rollback behavior that this PR does not change — it is the residue of the disproven deadline design. It is a good test and I would not delete it, but it belongs in its own PR rather than riding along with retention.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -57,6 +57,35 @@ pub const MAX_DEFERRED_PEEL_RESIDENCE_MS: u64 = 30 * 24 * 60 * 60 * 1_000;
...
+/// `MAX_PEEL_DEFERRED_ROWS_PER_GROUP` and the other per-scope retention caps in
+/// the workspace, and sits far above any plausible human burst across a stall.
+///
+/// At the cap the send is refused with
+/// [`EngineError::QueuedOutboundAtCapacity`](cgka_traits::error::EngineError::QueuedOutboundAtCapacity)
+/// so the caller learns its message was not accepted. Retention never silently
+/// drops work it already accepted — that is the difference from the
+/// deferred-peel cap, which may drop because transport redelivery is its
+/// recovery path.
+pub const MAX_QUEUED_OUTBOUND_INTENTS_PER_GROUP: usize = 256;
```
erskingardner, line 98: The cap is enforced in `queue_outbound_intent`, which as the comment says is the sole insertion point — so it also bounds two paths whose behavior this PR otherwise does not touch:

- the transport-inactive outbox from #1158, which reaches here through `do_queue_app_message`;
- the pre-existing `Stable`-but-unsettled queue, which retains *every* intent kind, not just `AppMessage`.

That is defensible (the constant's doc acknowledges the second one), but it means a user offline long enough to queue 256 messages in one group now gets a hard, deliberately non-transient `GroupSendQueueFull` on the 257th. Worth stating explicitly in the doc that the offline outbox is subject to the same bound.

Separately: capacity is only reclaimed on `confirm_queued_outbound_intent`, so drained-but-unconfirmed rows still occupy the cap. A group whose transport is failing publications can therefore sit at 256 with nothing left to drain. The tests cover the reclaim path but not that state; a line in the doc naming it would keep the next reader from assuming a successful drain frees the slot.

crates/cli/CHANGELOG.md
```diff
@@ -22,6 +22,13 @@ versioning through the workspace version in the root `Cargo.toml`.
   extensions can return bounded fallback content instead of creating a second
   stateful writer.
   
+- Outbound messages retained while a group is resolving a stalled publication
```
erskingardner: This release note scopes the cap to messages "retained while a group is resolving a stalled publication", but per the comment on `MAX_QUEUED_OUTBOUND_INTENTS_PER_GROUP` the bound is enforced at the queue's single insertion point and therefore also applies to the transport-inactive outbox (#1158) and to queued group-state intents on the `Stable`-but-unsettled path.

The user-visible consequence — an offline user can now be refused a send at 256 queued messages in one group, with an error that is deliberately not auto-retried — is the part a host integrator needs from a changelog, and it is the part this entry currently reads as excluding.

crates/traits/src/engine_state.rs
jgmontoya: Confirmed, and the history explains it: `58f42f58` added `is_terminal()` *with* a caller (the gate read it), then the round that narrowed retention to `is_resolving_local_publish` forced the gate back to explicit predicates — orphaning the method and adding the defense-in-depth comment in the same commit. Dropped in `7616a39d`: the partition test keeps every assertion via `is_unrecoverable() || is_disbanded()` (both pre-existing with production callers), workspace `--all-targets` green. Wiring it instead would have been dead by construction — `validate_send_acceptance` refuses both terminal states before the gate runs. The comment now says what the code does, and the doc block's unique claims already live in the spec paragraph this PR added.

crates/cgka-engine/src/message_processor/mod.rs
jgmontoya: Both mechanics check out, with one wording nit: drained-but-unconfirmed rows are still *drained* every pass — what they never do is free a slot, so the doc says "nothing frees a slot until a publish is accepted" rather than "nothing drainable". Fixed in `dbaac72b`: the constant's doc now enumerates all three paths (locally staged publication, `Stable`-but-unsettled, and the offline outbox — the one a user drives by typing) and carries the reclaim-on-confirmation paragraph. Same over-scoping fixed in the FFI `GroupSendQueueFull` doc (it claimed the queue "cannot drain until the stalled publication resolves" — wrong for the offline case) and the plan ledger row. Counting unconfirmed rows stays as-is deliberately: releasing on drain would let a failing transport grow the durable store without bound.

crates/cli/CHANGELOG.md
jgmontoya: Agreed — rewritten in `f61b3526` in host-integrator terms: the bound covers every reason a message waits (stalled publication, unsettled convergence, offline), the 257th send is refused with the typed, deliberately non-auto-retried `GroupSendQueueFull`, nothing already queued is discarded, and a slot frees only on relay acceptance.

crates/cgka-engine/src/publish.rs
jgmontoya: Right about the doc: `set_stable` doesn't refuse `PendingPublish`, so the "only exit" claim was enforced by callers, not the transition — fixed in `d551e849`. The gating holds, though: there are seven non-publish `set_stable` callers, not four, and each is unreachable while retained intents exist — ingest and the fork seam sit behind `can_ingest` (`Stable | Recovering` only, `engine_state.rs:270-272`, so `PendingPublish` buffers), the convergence apply's production entries are all state-gated, hydration re-arms the drain from the durable queue regardless, and the create/welcome sites can't hold intents. Decisively: an unconditional schedule wouldn't cover the hazard anyway — `confirm_publish` and `rollback_publish` both require `PendingPublish`, so a stolen transition errors *upstream* of the schedule call in either variant. And the round-2 oracle veto still reproduces on this tip: an unconditional schedule makes `empty_creation_confirms_without_synthetic_outbound_work` panic, because `scheduled_convergence_groups` feeds `runnable_work` and settled groups would report phantom work. So: read stays, doc fixed.

Your unscheduled-exit instinct was right somewhere else, and you should hear it from us: `do_join_welcome` never scheduled the drain — the hold-through-repair test passed only because it restarts. RED-verified and fixed in `8ba79919`; every seam returning a group to `Stable` in-process now schedules on its way out, and the helper doc states that contract.

Separately, we're following up with `set_stable` refusing `PendingPublish`/`Merging` the way it already refuses the terminal states — caller-side gates are drift-prone (this PR just proved that at the join seam), and the transition is where a violation actually bites. Own PR after this one merges, since the legal confirm-path transition needs the internal/public seam drawn carefully.

erskingardner (approved):

PR #1248, opened 2026-08-04
Add Milestone 5 app and multi-process convergence simulation

erskingardner:
## Summary

- add a production-shaped `MarmotAppRuntime` convergence adapter that executes canonical scenarios through real account/session/projection behavior
- add a versioned JSONL child-node protocol and multi-process orchestrator for isolated participant runtimes
- exercise retained relay histories, pause/resume, process-group kill, durable restart, full-history repair, application projection, and cross-adapter comparison
- write privacy-safe app/process failure capsules with stable participant, action, layer, classification, and replay metadata
- replace temporary milestone-scoped module and test names with durable capability names
- mark Milestone 5 complete and strengthen the Milestone 6 plan with decision-route assurance work derived from #1236, plus a final filename cleanup gate

## Behavior and impact

The simulator gains app-runtime and real child-process subjects while retaining the canonical Scenario IR used by the engine adapters.

The production app surface gains one additive `MarmotAppRuntime::repair_full_history` operation. It performs a caller-directed account-wide full-history query through the ordinary ingest, convergence, projection, event, and audit-tracker paths. Normal startup and cursor-based synchronization behavior are unchanged.

No convergence policy constants or canonical branch-selection rules are changed in this PR.

## Verification

- `just fast-ci`
- `cargo test -p cgka-conformance-simulator --test app_runtime_adapter --test node_protocol --test process_orchestrator`
- focused pause/resume, process kill/restart, retained-history repair, projection, and failure-capsule tests in those targets

## Follow-up

Milestone 6 will add distributed route-equivalence campaigns, including a permanent regression family for the cross-seam fork divergence documented in #1236. Post-Milestone-6 cleanup will remove the remaining tracked filename containing `milestone`.

## Summary by CodeRabbit

* **New Features**
  * Added application-runtime conformance simulation for group creation, messaging, synchronization, recovery, and policy changes.
  * Added versioned JSONL node communication with structured observations, progress reporting, and privacy-safe error details.
  * Added multi-process scenario execution with lifecycle controls, crash/restart handling, quiescence checks, and private reports.
  * Added full-history repair to restore account state after missed synchronization or offline periods.

* **Tests**
  * Added coverage for convergence, encrypted storage isolation, recovery, protocol validation, process orchestration, and sensitive-data redaction.

* **Documentation**
  * Updated the reliability plan to record simulation milestones and assurance work.


crates/cgka-conformance-simulator/src/process_orchestrator.rs
```diff
@@ -0,0 +1,1167 @@
...
+        self.lifecycle.push(ProcessLifecycleEventV1 {
+            action_id: action_id.into(),
+            participant: client.into(),
+            event: "launched".into(),
+        });
+        Ok(())
+    }
+
+    async fn configure_peers(&mut self) -> Result<(), ProcessOrchestratorError> {
+        let labels = self.nodes.keys().cloned().collect::<Vec<_>>();
```
erskingardner: This iterates every node and then calls `NodeProcess::send` directly, so it bypasses the `paused` check in `ProcessOrchestrator::send`. A SIGSTOPped child never reads its stdin, so this blocks for the full 45s `NODE_RESPONSE_TIMEOUT` and then fails the whole scenario.

`restart_participant` calls this unconditionally, so `RestartProcess` while any other participant is offline is a guaranteed 45s stall plus a spurious failure. Use `running_labels()` here (and at line 312 for the `SelectGroup` fan-out). Nothing in the current tests covers it: the lifecycle scenario pauses and resumes `bob` with no steps in between.

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -0,0 +1,989 @@
...
+                    .runtime()?
+                    .accept_group_invite(&participant.account_id, &group_id)
+                    .await
+                    .map_err(app_error)?;
+            }
+            let members = participant
+                .runtime()?
+                .group_members(&participant.account_id, &group_id)
+                .await;
+            if let Ok(members) = members {
```
erskingardner: `if let Ok(members)` / `if let Ok(state)` below drop the error and leave the previous cached value in place. `layered_observation` reads `cached_members` / `cached_epochs` and hashes them into `state_commitment_sha256`, so a failed read produces a *stale but plausible* commitment rather than a failure.

In `await_observable_settlement` that is the difference between "all participants agree" and "we could not read two of them and reused last round's answer" — the loop can return `Ok` on the second identical stale round. Propagate both errors with `.map_err(app_error)?`.

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -0,0 +1,989 @@
...
+        let Some(group_id) = self.scenario_groups.get(&group_label).cloned() else {
+            return Ok(());
+        };
+        for label in clients {
+            let participant = self.participant_mut(label)?;
+            if !participant.online {
+                continue;
+            }
+            let group_id_hex = hex::encode(group_id.as_slice());
+            if participant
```
erskingardner, line 398: Accepting the invite here makes the settlement gate self-fulfilling. `await_observable_settlement` calls `catch_up` → `refresh_cached_members` (which clears `pending_confirmation`) and then immediately evaluates `projections_settled = !pending_confirmation`. The condition it checks was cleared by the call one line earlier, so it can never observe a genuinely pending invite.

Make invite acceptance an explicit scenario operation and keep `refresh_cached_members` read-only.

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -0,0 +1,989 @@
...
+) -> AppRuntimeProtocolProjectionV1 {
+    #[derive(Serialize)]
+    struct Commitment<'a> {
+        epoch: u64,
+        member_identities: &'a [String],
+        admin_identities: &'a [String],
+        group_name: &'a str,
+        member_count: usize,
+    }
+    let member_count = if member_count == 0 {
```
erskingardner: This coercion sits inside the value that gets hashed into `state_commitment_sha256`. A participant whose projection genuinely reports `member_count == 0` (missing/incomplete group row) is silently rewritten to the roster length and therefore matches everyone else's commitment.

For a convergence oracle that is exactly the divergence you want surfaced. Either hash the raw value, or make `member_count == 0` an explicit observation error.

crates/cgka-conformance-simulator/src/process_orchestrator.rs
```diff
@@ -0,0 +1,1167 @@
...
+            })
+    }
+
+    fn write_failure_capsule(
+        &self,
+        participant: &str,
+        action_id: &str,
+        error: &NodeErrorV1,
+    ) -> Result<PathBuf, ProcessOrchestratorError> {
+        let directory = self.run_root.path().join("failure-capsules");
```
erskingardner: Capsules are written under `run_root`, which is the `TempDir` owned by the orchestrator. `ProcessScenarioReportV1::failure_capsules` records those paths, and `cgka-conformance-process` writes the report to a caller-supplied path outside the temp dir and then returns — dropping the `TempDir` and deleting every capsule it just pointed at.

So the only non-test consumer of the report always emits dangling capsule paths, which defeats the "replayable capsule" property the milestone claims. Write capsules next to the report, or take an explicit artifact directory.

crates/cgka-conformance-simulator/tests/tracing_audit.rs
```diff
@@ -190,6 +190,10 @@ fn production_library_sources_do_not_write_direct_output() {
     );
 }
 
+fn is_feature_gated_test_support(path: &Path) -> bool {
```
erskingardner, line 193: This is unrelated to the milestone. The file that trips the guard is `integrations/terminal-harness/src/test_support.rs` (the `eprintln!` at line 338), added by the terminal-harness connector already on master — not by anything in this PR.

Two problems: it is drive-by scope, and the name promises more than the body delivers. Nothing here checks for a `#![cfg(feature = ...)]` gate; it matches on basename, so `crates/storage-sqlite/src/storage/test_support.rs` is now permanently exempt from the direct-output audit too, and so is any future `test_support.rs`.

Please split this into its own PR and have it verify the actual feature gate.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1059,6 +1062,28 @@ async fn handle_account_worker_command(
                     .await;
             }
         }
+        AccountWorkerCommand::RepairFullHistory { respond } => {
```
erskingardner, line 1065: This arm silently diverges from the `CatchUp` arm directly above in two ways that look unintentional rather than deliberate:

- No `client.has_pending_epoch_backfill()` check. `repair_full_history` goes through `sync_sdk_relay` → `ingest_delivery` → `detect_epoch_stall`, so it can arm a backfill and durably write the `epoch_stall_backfill_armed` row. The `CatchUp` arm has a comment explaining why the tracker push must happen at the seam that observed the arm; that reasoning applies identically here.
- No `app_performance_telemetry().record(AppPerformanceOperation::AccountSync, ...)`, so a repair is invisible to sync performance accounting.

If either omission is intentional, a one-line comment saying so would help.

crates/marmot-app/src/client/sync.rs
```diff
@@ -651,6 +651,18 @@ impl AppClient {
         Ok(())
     }
 
+    /// Explicit account-wide repair for a host that has independent evidence
+    /// its incremental cursor may be incomplete (for example, a long-offline
+    /// participant that has no new traffic capable of arming epoch-stall
+    /// detection). Unlike the automatic detector path, this is a caller-owned
+    /// operation and therefore does not mutate the detector's debounce state.
+    pub(crate) async fn repair_full_history(&mut self) -> Result<SyncSummary, AppError> {
```
erskingardner, line 659: This calls `activate_transport(None)` directly instead of going through `prepare_transport_with_telemetry`, so relative to `sync_inner` it skips:

- `set_transport_signer` — the comment on that call says auth-gated relays (NIP-42) withhold gift-wrapped welcomes from unauthenticated subscribers. A missed welcome is precisely the class of gap this repair exists to close.
- `sync_runtime_groups()` — group subscriptions are not refreshed before the drain.
- `refresh_group_routes()` and `record_subscription_rebuild(...)`, so the repair leaves no subscription-rebuild audit row.

Today the account worker always runs startup sync before handling commands, so the signer happens to already be set on the relay plane. That is an implicit ordering dependency on a `pub` API. `run_pending_epoch_backfill` gets away with the same shape because it only ever runs mid-session; this one is caller-directed. Either route through `prepare_transport_with_telemetry` or document the precondition.

crates/cgka-conformance-simulator/tests/app_runtime_adapter.rs
```diff
@@ -0,0 +1,198 @@
...
+        .update_group_data(cgka_conformance_simulator::SubjectUpdateGroupData {
+            action_id: "rename-2",
+            client: "alice",
+            name: "after quiet history",
+            pending: "rename-2",
+        })
+        .await
+        .unwrap();
+
+    // No new event is published after Bob reconnects. Startup EOSE/history
```
erskingardner: The comment says startup EOSE/history completion "must close the gap", but the next line calls `repair_full_history` explicitly, so what the test proves is that the new repair operation works — not that the incremental path has the blind spot that motivates it.

That blind spot is the entire justification for adding a public `MarmotAppRuntime::repair_full_history`. Reproducing it first would make the case: assert that `catch_up` alone leaves bob behind, then assert `repair_full_history` fixes it. Without the negative control, a future change that makes plain catch-up sufficient would leave this API in place with no test noticing.

crates/cgka-conformance-simulator/src/node_protocol.rs
```diff
@@ -0,0 +1,998 @@
...
+    pub relay_publish_failures: usize,
+    pub relay_directory_inflight_fetches: usize,
+    pub relay_directory_completed_fetches: usize,
+    pub retry_timer_armed: bool,
+    pub projection_checkpoint_sha256: String,
+    pub stable_checkpoint_observations: u32,
+}
+
+impl NodeProgressV1 {
+    pub fn observably_quiescent(&self) -> bool {
```
erskingardner, line 240: Non-blocking, but four of these six gates cannot fail:

- `commands_in_flight` and `outbox_pending` are hardcoded to `0` in `observe_node` (lines 751-752). The first is true by construction; the second is not — the runtime can hold queued outbound intents this projection never reads.
- `relay_query_in_flight` is assigned `telemetry.health.directory_inflight_fetches > 0`, which is the negation of the `relay_directory_inflight_fetches == 0` check on the next line. One signal, counted twice.

So `observably_quiescent` is effectively `!retry_timer_armed && stable_checkpoint_observations >= 2`. That may be adequate, but the predicate reads as six independent conditions and `ProcessOrchestrator::await_quiescence` relies on it. Either derive `outbox_pending` from a real signal or reduce the predicate to what it actually checks and say so in a comment.

crates/cgka-conformance-simulator/tests/process_orchestrator.rs
```diff
@@ -0,0 +1,429 @@
...
+            },
+        ),
+        ScenarioStep::DeliverAll,
+        ScenarioStep::Tick {
+            clients: clients.clone(),
+        },
+    ];
+    if lifecycle {
+        steps.extend([
+            ScenarioStep::SetClientOffline {
```
erskingardner: Non-blocking: `SetClientOffline` is immediately followed by `ReconnectClient` with no intervening steps, so nothing happens during bob's offline window. The test asserts `paused`/`resumed` lifecycle events were recorded, but the recovery property in the test name — that pause/resume agrees with uninterrupted execution — is vacuous here, since the two runs have identical event histories either way.

Moving a rename or an app message between the two steps would make this test earn its name (and would have caught the paused-node fan-out bug flagged in `configure_peers`).

crates/cgka-conformance-simulator/src/lib.rs
```diff
@@ -51,6 +54,11 @@ pub mod subject;
 mod topology;
 pub mod vector;
 
+pub use app_runtime::{
+    APP_RUNTIME_OBSERVATION_SCHEMA_VERSION, AppRuntimeApplicationProjectionV1, AppRuntimeHarness,
+    AppRuntimeLocalDiagnosticsV1, AppRuntimeObservationV1, AppRuntimeProtocolProjectionV1,
+    AppRuntimeScenarioResultV1,
```
erskingardner: `AppRuntimeScenarioResultV1` is defined in `app_runtime.rs` and re-exported here, but never constructed or read anywhere in the crate or the tests. Same for `ProcessOrchestrator::run_root()`. Please drop both until something needs them.

crates/cgka-conformance-simulator/src/process_orchestrator.rs
```diff
@@ -0,0 +1,1167 @@
...
+            .iter()
+            .map(|(label, node)| (label.clone(), node.root.clone()))
+            .collect()
+    }
+
+    pub async fn run(
+        &mut self,
+        scenario: &ScenarioSpec,
+    ) -> Result<ProcessScenarioReportV1, ProcessOrchestratorError> {
+        let compiled = compile_scenario(scenario).map_err(|error| {
```
erskingardner: Non-blocking: `launch` already compiled this scenario, and `run` compiles it again from a separately-passed `&ScenarioSpec` that nothing checks is the same spec. Storing `compiled` (or at least the clients/topology fingerprint) on the orchestrator would remove the duplicate work and the mismatch hazard — `run` with a different spec than `launch` would currently produce a report whose `participant_roots` and topology belong to a different scenario.

erskingardner (commented):
Solid, careful engineering overall, and the layered protocol/application/local observation split is the right shape. But this is ~4.2k lines of new assurance machinery whose job is to *not lie about convergence*, and several places currently make the oracle weaker than it reads. I'd like those fixed before the milestone is marked complete, because a false PASS here is more costly than no coverage at all.

**Blocking**

1. `ProcessOrchestrator::configure_peers` (and the `SelectGroup` fan-out in `CreateGroup`) iterate every node instead of `running_labels()`, and `configure_peers` calls `NodeProcess::send` directly, bypassing the `paused` guard. Writing to a SIGSTOPped child stalls for the full 45s `NODE_RESPONSE_TIMEOUT` and then fails the scenario. Any scenario that creates a group or restarts a process while another participant is offline hits this.
2. `AppRuntimeHarness::refresh_cached_members` discards `group_members` / `group_mls_state` errors (`if let Ok(...)`) and silently keeps the previous cache. `layered_observation` then hashes those stale values into `state_commitment_sha256`, and `await_observable_settlement` can declare a stable shared commitment from data that was never actually read. Propagate the errors.
3. Same function accepts pending invites as a side effect. `pending_confirmation` is the settlement gate in `await_observable_settlement`, and the refresh that runs immediately before each evaluation clears it — the gate can't fail. Move invite acceptance into an explicit scenario operation.
4. `public_protocol_projection` rewrites `member_count == 0` to `member_identities.len()`. That is inside the commitment hash, so a participant genuinely reporting zero members is coerced into agreeing with everyone else.

**Scope**

5. The `tracing_audit.rs` exemption is unrelated to this PR — it silences a failure introduced by the terminal-harness connector already on master. It also exempts on basename alone despite being named `is_feature_gated_test_support`, permanently exempting `crates/storage-sqlite/src/storage/test_support.rs` from the direct-output guard. Please split it out and make it check the actual `cfg(feature = ...)` gate.

**Production surface**

6. `repair_full_history` is a reasonable seam, but the worker arm diverges from the `CatchUp` arm in ways that look unintentional (no `has_pending_epoch_backfill()` tracker scheduling, no `AccountSync` telemetry), and the client method skips `prepare_transport_with_telemetry` entirely. See inline.
7. No test demonstrates that plain catch-up *fails* to close the quiet-offline gap, so the premise for adding a new public production API is asserted rather than reproduced. A negative control would settle it in a few lines.

Remaining inline comments are non-blocking (dead code, tautological quiescence gates, a lifecycle test whose offline window contains no traffic).

crates/cgka-conformance-simulator/src/process_orchestrator.rs
```diff
@@ -0,0 +1,1167 @@
...
+    ) -> Result<PathBuf, ProcessOrchestratorError> {
+        let directory = self.run_root.path().join("failure-capsules");
+        fs_private::create_dir_all_private(&directory).map_err(environment_error)?;
+        let digest = hex::encode(sha2::Sha256::digest(action_id.as_bytes()));
+        let path = directory.join(format!("{}.json", &digest[..16]));
+        NodeFailureCapsuleV1::from_error(participant, action_id, error)
+            .write_private(&path)
+            .map_err(environment_error)?;
+        Ok(path)
+    }
```
erskingardner, line 984: Capsules are written under `run_root`, a `tempfile::TempDir` that is deleted when the orchestrator is dropped. In the `cgka-conformance-process` CLI flow, the report is written to the caller's path, `shutdown()` runs, and the orchestrator drops — so every `failure_capsules` path persisted in the report (and `participant_roots`) points at a deleted file. The tests pass only because they read the capsule before the drop. Please write capsules to a durable artifact directory (alongside the report, or an explicit artifact dir passed to `launch`) so the shareable-capsule workflow survives process exit.

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -0,0 +1,989 @@
...
+        state_commitment_sha256: hex::encode(Sha256::digest(encoded)),
+    }
+}
+
+fn opaque_public_identity(account: &str) -> String {
+    format!(
+        "opaque:{}",
+        &hex::encode(Sha256::digest(account.as_bytes()))[..16]
+    )
+}
```
erskingardner: This fallback renders as `opaque:{hash}` while the node protocol's equivalent (`node_protocol.rs:842` `opaque_identity`) renders as `unknown:{hash}`. Both feed `member_identities`/`admin_identities`, which feed `state_commitment_sha256`. For any member missing from the peer-alias map (e.g., learned via retained relay history), the app adapter and the process adapter hash different identity strings for identical group state, so the cross-adapter comparison reports a false divergence. Please use one spelling — `node_protocol` already imports `public_protocol_projection` from this module, so a shared helper fits naturally.

crates/cgka-conformance-simulator/src/process_orchestrator.rs
```diff
@@ -0,0 +1,1167 @@
...
+            }
+            ScenarioStep::SyncRelayHistory { clients, sync } => {
+                let full_history = matches!(
+                    sync,
+                    ScenarioRelaySyncModeV2::FullHistory
+                        | ScenarioRelaySyncModeV2::SetReconciliation
+                );
+                self.catch_up(clients, action_id, full_history).await?;
+                Ok((clients.clone(), ProcessActionStatusV1::Completed))
+            }
```
erskingardner, line 442: This arm forwards the scenario's `clients` verbatim, so a `SyncRelayHistory` step naming a paused participant fails with `participant_paused` — unlike `Tick` (filters paused) and `DeliverAll` (uses `running_labels()`), and unlike the in-process app adapter, whose `catch_up`/`repair_full_history` skip offline participants. Same class as the CreateGroup/`configure_peers` broadcast issue: please filter to running participants here too (or deliberately skip-and-record) so offline-then-repair scenarios behave consistently across adapters.

erskingardner (commented):
Solid milestone overall: the production change (`MarmotAppRuntime::repair_full_history`) is minimal, additive, privacy-safe (`privacy_safe_kind`, audit-tracker scheduling, ordinary ingest path), and its doc claims check out (`activate_transport(None)` is genuinely unfloored). The black-box discipline, restrictive-root posture, and capsule marker tests are good. A few functional defects in the new test tooling should be fixed before merge.

**Blocking**

1. **Failure capsules referenced by the persisted report are deleted before anyone can read them** (inline on `write_failure_capsule`). The CLI failure path produces a report whose `failure_capsules`/`participant_roots` point into a dropped `TempDir` — the headline shareable-capsule workflow is broken outside the tests.
2. **Paused participants still receive node commands in three places**, spuriously failing legitimate offline scenarios: the CreateGroup `SelectGroup` broadcast and `configure_peers` (already flagged by CodeRabbit) plus `SyncRelayHistory` (inline). Please use `running_labels()`/paused-filtering consistently; the in-process adapter silently skips offline participants, so the two adapters currently disagree.

**Should fix (non-blocking)**

- `opaque:` vs `unknown:` identity-fallback mismatch across the two adapters → false cross-adapter commitment divergence (inline).
- Endorsing the highest-value existing threads, all verified: vacuous `commands_in_flight`/`outbox_pending` in `observably_quiescent` (the plan doc overclaims the outbox boundary); swallowed read errors in `refresh_cached_members` letting stale caches produce "stable" commitments; invite auto-accept mutating inside the observation path (and placed differently than the node's CatchUp-time accept); `block_on_subject` panicking on a current-thread runtime (document or guard the multi-thread requirement); node `CatchUp` failing when no group is selected; oversized JSONL frames leaving the line remainder to desync the stream.
- ~150 lines of near-duplicate admin-set/observation logic between `app_runtime.rs` and `node_protocol.rs` (`apply_admin_set` vs `set_admins`, `layered_observation` vs `observe_node`). `node_protocol` already imports `public_protocol_projection` from `app_runtime`; sharing more would cut real duplication.
- Nit: the PR body says milestone-scoped names were replaced, but the new `ScenarioSpec::name` values still carry `milestone5-` prefixes and land in persisted reports.

**Housekeeping**

- The `tracing_audit.rs` hunk duplicates #1247 (`1968d258`, already on master with the identical `is_feature_gated_test_support` exemption) — rebasing onto master drops it from this diff.
- The red **Simulator and vectors** check is pre-existing on master (`rg: command not found` in `tla-liveness-counterexample`, fails on master too). Not this PR's doing, but CI can't go green until master is fixed and this branch is rebased.

PR #1249, opened 2026-08-04
Move full simulator campaigns to nightly CI

erskingardner:
## What changed

- keep the required `Simulator and vectors` PR check, but make its simulator pass a focused smoke lane
- move the generated multi-minute batches, adversarial reliability campaigns, policy sweeps, ignored sustained campaigns, and process-isolated campaign runner to one scheduled nightly job
- retain simulator doctests, independent convergence verification, and the encrypted 19-vector report on every PR
- centralize the smoke and full nextest filters in reproducible `just` recipes, with a contract check for their intended coverage delta
- make nightly failures open a GitHub issue containing the failed run URL
- prevent scheduled and manually dispatched nightly runs from cancelling one another
- replace historical milestone labels across the simulator with descriptive campaign, scenario, recipe, and CI names
- version the renamed generator output contract as `2` / `2-regression`
- use portable `grep` in the liveness counterexample check because the Ubuntu runner does not install ripgrep

## Coverage tradeoff

Per-PR coverage intentionally decreases to restore useful feedback time. The PR lane no longer runs:

- the `adversarial_reliability_campaigns` binary, including its ignored multi-round campaigns
- the `policy_sweeps` binary
- the process-isolated adversarial campaign runner
- these three generated multi-minute batches from `canonical_scenarios`:
  - `convergence_chaos_family_generates_specs_with_semantic_expectations`
  - `convergence_chaos_family_seed_changes_scenarios`
  - `convergence_e2e_delivery_family_runs_generated_variants`

Those checks now run after merge in `Simulator Nightly`, together with the broader `conformance-slow` configuration. A regression confined to that coverage can therefore land before nightly catches it. To make that deferred signal visible, a failed nightly opens a GitHub issue linked to the run.

The PR lane continues to run the other 267 simulator tests, doctests, independent convergence/model checks, and the encrypted vector report. The nightly runs the full generic suite plus every dedicated campaign and verification recipe.

## Why

The simulator check had become dominated by generated reliability campaigns rather than PR feedback. A prior green run spent 18m40s in the initial simulator suite; after the newer coverage landed, one run spent 31m05s there and the complete job took 47m23s. Serializing the heavy tests made this worse because the two chaos batches alone took 10m58s and 13m49s.

The simulator surfaces now describe their purpose directly: `adversarial-reliability/v1`, `adversarial_reliability_campaigns`, `adversarial-reliability-ci`, and `convergence-verification-ci`.

The original master failure after the simulator work was a separate portability issue: the liveness counterexample check invoked `rg` on a runner where ripgrep was unavailable, producing exit code 127. That check now uses `grep`, which is present on the Ubuntu image.

## Validation

- live post-review `Simulator and vectors` job passed in 5m48s; its smoke step passed in 4m04s
- `just simulator-filter-contract`
- adversarial generator-version contract test passes
- all simulator targets compile with `test-policy-overrides`
- renamed campaign target and catalog test pass
- full `just convergence-verification-ci` passes, including the expected TLA counterexample check
- `actionlint .github/workflows/ci.yml .github/workflows/simulator-nightly.yml`
- both workflow files parse with Ruby's YAML parser
- `RUSTDOCFLAGS='-D warnings' cargo doc -p cgka-conformance-simulator --no-deps --features test-policy-overrides`
- `git diff --check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -321,6 +321,11 @@ jobs:
       - name: Install cargo-nextest
         uses: taiki-e/install-action@b18fb392e1a1b90971f7c7572c92790fa54f23d5 # nextest
 
+      - name: Install just
+        uses: taiki-e/install-action@b18fb392e1a1b90971f7c7572c92790fa54f23d5 # just
+        with:
+          tool: just
+
       - name: Cache Cargo artifacts
         uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
 
@@ -328,17 +333,10 @@ jobs:
         # Multi-minute generated reliability batches and the adversarial
         # campaign run in the scheduled Simulator Nightly workflow. Independent
         # convergence checks run in their dedicated fast gate below.
-        run: |
-          cargo nextest run -p cgka-conformance-simulator --locked --profile ci \
-            -E 'not binary(adversarial_reliability_campaigns) and not binary(policy_sweeps) and not binary(independent_reference_model) and not binary(lifecycle_model) and not binary(mutation_adequacy) and not binary(protocol_decision_gate) and not (binary(canonical_scenarios) & (test(=convergence_chaos_family_generates_specs_with_semantic_expectations) | test(=convergence_chaos_family_seed_changes_scenarios) | test(=convergence_e2e_delivery_family_runs_generated_variants)))'
+        run: just simulator-smoke
 
       - name: Run simulator doctests
         run: cargo test -p cgka-conformance-simulator --doc --locked
-
-      - name: Install just
-        uses: taiki-e/install-action@b18fb392e1a1b90971f7c7572c92790fa54f23d5 # just
-        with:
-          tool: just
 
       - name: Verify convergence with independent models
         run: just convergence-verification-ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -324,8 +324,13 @@ jobs:
       - name: Cache Cargo artifacts
         uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
 
-      - name: Run simulator tests
-        run: cargo nextest run -p cgka-conformance-simulator --locked --profile ci
+      - name: Run simulator smoke tests
+        # Multi-minute generated reliability batches and the Milestone 3
+        # campaign run in the scheduled Simulator Nightly workflow. Milestone
+        # 4 binaries run in their dedicated fast gate below.
+        run: |
+          cargo nextest run -p cgka-conformance-simulator --locked --profile ci \
+            -E 'not binary(milestone3_campaigns) and not binary(policy_sweeps) and not binary(independent_reference_model) and not binary(lifecycle_model) and not binary(mutation_adequacy) and not binary(protocol_decision_gate) and not (binary(canonical_scenarios) & (test(=convergence_chaos_family_generates_specs_with_semantic_expectations) | test(=convergence_chaos_family_seed_changes_scenarios) | test(=convergence_e2e_delivery_family_runs_generated_variants)))'
 
       - name: Run simulator doctests
         run: cargo test -p cgka-conformance-simulator --doc --locked
@@ -334,9 +339,6 @@ jobs:
         uses: taiki-e/install-action@b18fb392e1a1b90971f7c7572c92790fa54f23d5 # just
         with:
           tool: just
-
-      - name: Run Milestone 3 adversarial campaign gate
-        run: just milestone3-ci
 
       - name: Run Milestone 4 independent verification gate
         run: just milestone4-ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -325,12 +325,12 @@ jobs:
         uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
 
       - name: Run simulator smoke tests
-        # Multi-minute generated reliability batches and the Milestone 3
-        # campaign run in the scheduled Simulator Nightly workflow. Milestone
-        # 4 binaries run in their dedicated fast gate below.
+        # Multi-minute generated reliability batches and the adversarial
+        # campaign run in the scheduled Simulator Nightly workflow. Independent
+        # convergence checks run in their dedicated fast gate below.
         run: |
           cargo nextest run -p cgka-conformance-simulator --locked --profile ci \
-            -E 'not binary(milestone3_campaigns) and not binary(policy_sweeps) and not binary(independent_reference_model) and not binary(lifecycle_model) and not binary(mutation_adequacy) and not binary(protocol_decision_gate) and not (binary(canonical_scenarios) & (test(=convergence_chaos_family_generates_specs_with_semantic_expectations) | test(=convergence_chaos_family_seed_changes_scenarios) | test(=convergence_e2e_delivery_family_runs_generated_variants)))'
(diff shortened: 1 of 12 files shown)


.github/workflows/simulator-nightly.yml
```diff
@@ -0,0 +1,84 @@
...
+  schedule:
+    - cron: "17 3 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: simulator-nightly
+  cancel-in-progress: true
```
erskingardner: **`cancel-in-progress: true` is the wrong policy for this job.**

This nightly is now the *only* place `adversarial_reliability_campaigns`, `policy_sweeps`, and the three generated batches run. With `cancel-in-progress: true`, a `workflow_dispatch` run (or any overlap with the scheduled run) kills the in-flight one — and since the job can take an hour-plus, cancellation is likely rather than theoretical. The result is a night with zero reliability signal and a cancelled (not failed) run that looks benign in the UI.

For a safety-net job you want every run to finish:

```yaml
concurrency:
  group: simulator-nightly
  cancel-in-progress: false
```

.github/workflows/simulator-nightly.yml
```diff
@@ -0,0 +1,84 @@
+name: Simulator Nightly
+
+on:
+  schedule:
+    - cron: "17 3 * * *"
```
erskingardner, line 5: **Nothing tells anyone when this fails.**

Moving coverage off the PR lane is only safe if the nightly result is actually observed. GitHub's default for a failed scheduled workflow is an email to the last committer on the default branch, subject to their personal notification settings — in practice that means these failures go unnoticed for days, and the regression they would have caught lands in the meantime.

Since this is now the sole home for the adversarial campaigns, `policy_sweeps`, and the `conformance-slow` suite, please add a terminal step that makes failure loud, e.g.:

```yaml
      - name: Report nightly failure
        if: failure()
        env:
          GH_TOKEN: ${{ github.token }}
        run: |
          gh issue create --title "Simulator nightly failed ($(date -u +%F))" \
            --body "Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
```

(that needs `issues: write` added to `permissions`). A chat/webhook notification works equally well — the point is that a silent nightly turns this PR from "rescheduled coverage" into "deleted coverage".

.github/workflows/ci.yml
```diff
@@ -324,8 +324,13 @@ jobs:
...
 
-      - name: Run simulator tests
-        run: cargo nextest run -p cgka-conformance-simulator --locked --profile ci
+      - name: Run simulator smoke tests
+        # Multi-minute generated reliability batches and the adversarial
+        # campaign run in the scheduled Simulator Nightly workflow. Independent
+        # convergence checks run in their dedicated fast gate below.
+        run: |
+          cargo nextest run -p cgka-conformance-simulator --locked --profile ci \
+            -E 'not binary(adversarial_reliability_campaigns) and not binary(policy_sweeps) and not binary(independent_reference_model) and not binary(lifecycle_model) and not binary(mutation_adequacy) and not binary(protocol_decision_gate) and not (binary(canonical_scenarios) & (test(=convergence_chaos_family_generates_specs_with_semantic_expectations) | test(=convergence_chaos_family_seed_changes_scenarios) | test(=convergence_e2e_delivery_family_runs_generated_variants)))'
```
erskingardner: **Two hand-maintained copies of this filter will drift.**

This expression and the one in `simulator-nightly.yml` share six of seven clauses and must stay in sync by hand. The two failure modes are asymmetric and both silent: forget to add a new heavy binary here and the PR lane goes slow again; forget to *stop* excluding it in the nightly and it runs nowhere.

Every other gate in this job already delegates to `just` (`convergence-verification-ci`). Do the same here — e.g. `simulator-smoke` and `simulator-full` recipes in the `Justfile`, with the shared exclusion list written once — so the workflows read as `run: just simulator-smoke` and the filter has a single source of truth. It also makes the filter reproducible locally, which a 300-character inline `-E` string is not.

A deny-list is also unverifiable by construction: nothing fails if a new expensive binary is added and nobody remembers either file. Worth a nightly assertion that the PR lane's excluded set is a subset of what the nightly actually ran.

.github/workflows/ci.yml
```diff
@@ -324,8 +324,13 @@ jobs:
       - name: Cache Cargo artifacts
         uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
 
-      - name: Run simulator tests
-        run: cargo nextest run -p cgka-conformance-simulator --locked --profile ci
+      - name: Run simulator smoke tests
```
erskingardner, line 332: **The PR description understates what leaves the PR lane.**

The body says the PR lane excludes "exactly these generated batches" (the three `canonical_scenarios` tests) and that this "changes scheduling, not reliability coverage." That is not what this filter does. It also drops the entire `adversarial_reliability_campaigns` and `policy_sweeps` binaries, and the removal of `just milestone3-ci` from this job takes the `--ignored` campaigns and the process-isolated campaign runner with it.

Coverage is preserved *in aggregate across PR + nightly*, but per-PR coverage genuinely drops: a regression in the adversarial catalog is now caught after merge, on master, by a job with no owner — not on the PR that introduced it. That is a defensible trade for a 47m job, but it should be stated plainly in the description rather than framed as a pure scheduling change. Please update the body so the next person reading it knows what the PR lane no longer proves.

erskingardner (commented):
The premise is sound. A 47-minute required check that spends 25 of those minutes on generated reliability batches is not giving PR authors feedback, it is taxing them, and splitting it is the right call. The `rg` -> `grep` fix is a correct, minimal portability fix. The mechanics are careful: the exclusion sets in the two lanes are consistent with the `just` recipes they defer to, the rename is complete (no dangling `milestone3`/`milestone4` references anywhere in the tree), and the 267/270 nextest filter inspection is exactly the kind of validation this change needed.

Three things should be resolved before merge, all in the new nightly rather than in the code.

**1. The nightly is silent on failure.** This PR is only "rescheduling, not deleting" coverage if someone actually reads the nightly result. A scheduled workflow that fails emails the last committer on the default branch, subject to their personal notification settings — which in practice means nobody. `adversarial_reliability_campaigns`, `policy_sweeps`, and the `conformance-slow` suite now live nowhere else. Please add a `if: failure()` step that files an issue or posts to a channel.

**2. `cancel-in-progress: true` on a 120-minute safety net.** Any `workflow_dispatch` overlapping the scheduled run kills one of them, and a cancelled run reads as benign in the UI. Should be `false`.

**3. The exclusion filter is duplicated across two workflow files.** Six of seven clauses are shared and must be kept in sync by hand; drift in one direction makes the PR lane slow again, in the other direction it opens a coverage hole, and neither fails anything. Every other gate in this job already calls `just` — move the filters into `simulator-smoke` / `simulator-full` recipes so there is one source of truth and the lanes are reproducible locally.

Also please correct the PR description: the body says the PR lane excludes "exactly these generated batches" and that this "changes scheduling, not reliability coverage," but the filter also drops the whole `adversarial_reliability_campaigns` and `policy_sweeps` binaries, and dropping `just milestone3-ci` takes the `--ignored` campaigns and the process-isolated runner with them. Per-PR coverage really does decrease; that trade is fine, it just should not be described as neutral.

**Scope.** The CI change is ~60 lines. The `milestone3` -> `adversarial_reliability` rename is the other ~300 and is unrelated to it — a pure naming refactor riding along with the change that carries all the risk. I would rather have reviewed them separately, and the mixed diff is why the actual coverage delta above was easy to miss. The rename itself is good and I would not ask you to revert it now, but next time please land the CI split alone.

Two smaller notes, non-blocking:

- The nightly's "complete simulator suite" step adds `--features conformance-slow`, which never ran in CI before. That is a coverage increase, and a new source of unattributed nightly flake — another argument for point 1.
- `--family milestone3-adversarial/v1` is a breaking rename of a CLI value on `cgka-conformance-simulator-report`. Internal-only and `publish = false`, so no action needed; just worth knowing if any saved invocations exist outside the repo.

PR #1250, opened 2026-08-04
Add distributed convergence assurance and campaign infrastructure

erskingardner:
## Summary

- advance Milestone 6 decision-route assurance with a deterministic #1236-derived engine/app-runtime family, exact canonical-state comparison, application disposition evidence, restart coverage, and active bidirectional decryptability checks
- close the production route asymmetries the campaigns exposed: reconsiderable same-epoch losers, candidate-context peeling, deferred application admission/replay, bounded fairness for admin work, inactive candidate/removal handling, terminal content separation, and parent-bound own-commit replay
- add a durable distributed campaign runner for process, OCI container, retained-relay, scheduled network/host/disk/database faults, mixed participant images, and capability-gated external VM execution
- add drift-checked PR/nightly/weekly/release-hardening lanes with explicit wall/CPU/RSS/disk/artifact/retention/flake budgets and completeness-checked evidence bundles
- add the unified failure-corpus lifecycle, semantic reduction, cross-adapter reproduction candidates, safe vector promotion, recurrence tracking, and time-to-diagnosis evidence
- finish the post-milestone naming cleanup so long-lived tracked artifacts use capability names rather than milestone-scoped names

## Assurance scope

This PR materially changes production convergence behavior where the new cross-route campaigns found inconsistencies; it is not simulator-only. The engine changes are bounded and regression-covered, but they carry the normal risk of changing branch selection, deferred replay, application delivery/invalidation, and recovery behavior under competing commits.

The result is a scoped assurance and operations capability, not a proof that every possible production execution is correct or that the current constants are universally optimal. Each release-hardening run must still satisfy its declared coverage and budgets, investigate counterexamples, and retain the resulting evidence bundle.

## Open assurance item

The deterministic #1236 counterexample is closed across the engine and app-runtime routes: they reach the reference result, exact cryptographic state, and active bidirectional decryptability. The process and distributed subjects use a live retained relay whose high-level mutations catch up and publish immediately. Depending on background fetch timing, the second committer can observe the first root before creating its own, so those subjects do not yet deterministically receive the same-source-epoch fork input set. They prove exact agreement and decryptability for the input set observed, but Milestone 6 remains open until a controllable publication or relay-visibility seam stages the #1236 input set across those adapters.

The branch is rebased onto current `master`, including the epoch-stall settlement speedup and the runtime-start ordering stabilization.

## Verification

- `cargo test -p cgka-engine --locked --features test-policy-overrides --test distributed_convergence` — 79 passed
- `cargo test -p cgka-conformance-simulator --locked --test route_equivalence` — 6 passed; production-shaped app-runtime case included
- `cargo test -p cgka-conformance-simulator --locked --test app_runtime_adapter --test node_protocol`
- `cargo test -p cgka-conformance-simulator --locked --test process_orchestrator` — 8 passed
- `cargo test -p cgka-conformance-simulator --locked --test report_runner` — 19 passed
- focused simulator unit tests for offline exact-state invalidation, observation-schema gating, and reopened assurance claims
- `cargo clippy -p cgka-conformance-simulator -p convergence-campaign-runner -p cgka-engine --all-targets -- -D warnings`
- `cargo test -p marmot-app --locked --lib tests::runtime_start_returns_before_initial_directory_subscription_registration -- --exact`
- `just naming-gate`
- `just tamarin` — 78 lemmas
- real OCI container-runtime coverage was also exercised during the Milestone 6 implementation checkpoints

GitHub CI is running against the rebased head.

## Summary by CodeRabbit

* **New Features**
  * Added distributed convergence campaign support for container and virtual-machine environments.
  * Added campaign planning, execution, evidence validation, failure tracking, and configurable test lanes.
  * Added relay and proxy support for multi-process conformance testing.
  * Added exact state observations, application delivery tracking, and bidirectional decryptability checks.

* **Bug Fixes**
  * Improved fork recovery and convergence handling, including reconsideration of losing branches and restart consistency.
  * Added safeguards for bounded recovery processing and deferred message handling.

* **Documentation**
  * Added operator guidance, route-assurance coverage, campaign workflows, and architecture documentation.


crates/cgka-engine/src/fork_recovery.rs
```diff
@@ -164,10 +173,108 @@ impl ForkRecoveryManager {
...
+        };
+        // The pre-commit snapshot also predates every local descendant of the
+        // incumbent root. Preserve authenticated descendant commits before the
+        // rollback so a branch that already grew locally does not lose the
+        // very evidence that makes it canonical. This matters when the outer
+        // transport echo is withheld or delayed: OpenMLS cannot process our
+        // own commit again, so the confirm-time convergence stamp on this row
+        // is the only replayable representation of that descendant.
+        let mut displaced_descendant_commits = storage
+            .list_messages(group_id, source_epoch)?
```
erskingardner: **Descendant preservation has no state filter and no ancestry check.**

This scan keeps *every* commit row at a source epoch greater than `source_epoch` and rewrites it to `MessageState::Retryable`. Two gaps:

1. **No `record.state` filter.** A row that was previously given a terminal verdict — `Failed`, `EpochInvalidated`, or a convergence `LosingBranch` — is resurrected as `Retryable` and re-enters replay/convergence. The comment above calls these "authenticated descendant commits", but nothing here checks that they were ever accepted.
2. **No ancestry check.** Nothing verifies the commit actually descends from `incumbent`. A commit at epoch N+1 that belongs to a *different* sibling branch (or that this node rejected earlier) is preserved and re-armed identically.

Failure scenario: node applies incumbent I@N, then receives peer commit P@N+1 on a sibling lineage and terminally invalidates it. A later candidate C@N wins the pairwise race; this loop re-persists P as `Retryable` at epoch N+1, and the next convergence pass treats a previously-rejected commit as fresh branch-growth evidence.

Suggest restricting to rows whose state indicates prior acceptance (e.g. `Processed`/`Sent`) and, ideally, confirming the parent lineage rather than relying on epoch ordering alone.

crates/cgka-engine/src/fork_recovery.rs
```diff
@@ -164,10 +173,108 @@ impl ForkRecoveryManager {
...
+        // resolve can redo the rollback + re-persist. Releasing or forgetting
+        // first would lose the displaced branch permanently (row swept,
+        // snapshot gone, record gone). The in-memory removal is therefore the
+        // last step — the point of no return.
+        //
+        // The snapshot operations necessarily sit OUTSIDE any storage
+        // transaction: create/rollback/release drive their own SQLite
+        // transactions (see the boundary comment in
+        // `openmls_projection::apply_openmls_canonicalization_result_inner`).
+        // The re-persist is a single-row write and is already atomic on its
```
erskingardner: This rationale is now stale: the block below is no longer "a single-row write". It is `displaced_incumbent_row` **plus** an unbounded `for record in displaced_descendant_commits` loop, all outside any transaction.

`AGENTS.md` requires multi-step state changes to be torn-write-free ("compensate every applied step on the error path"). A `put_message` failure partway through the descendant loop leaves storage with some descendants re-armed as `Retryable` and others gone. The retained snapshot makes it *retryable*, but the comment should stop asserting single-write atomicity, and it is worth stating explicitly why a partially re-armed descendant set is safe to observe before the retry lands.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -2433,6 +2876,10 @@ fn process_openmls_messages_inner<S: StorageProvider>(
                 committer: stamp.committer.as_slice().to_vec(),
                 consumed_proposal_refs: stamp.consumed_proposal_refs.clone(),
             });
+            // The exact parent GroupContext match above re-confirms the
+            // lineage before the anchor rollforward. Keep that fact explicit
+            // for a subsequent stamped own commit in the same path.
+            prefix_canonical = true;
```
erskingardner: `prefix_canonical = true;` is a no-op — this arm is only reachable when the `if` condition's `&& prefix_canonical` term already held, and nothing between there and here writes it. The comment justifies a fact the code cannot express. Please drop the assignment (keep the comment if the reasoning is worth recording).

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -2390,6 +2821,11 @@ fn process_openmls_messages_inner<S: StorageProvider>(
         if projection.kind == OpenMlsContentKind::Commit
             && prefix_canonical
             && let Some(stamp) = own_commits.stamp(&projection.message_digest)
+            && own_commit_stamp_parent_matches(
```
erskingardner, line 2843: **Backward-compatibility of the new parent-context gate.**

`parent_group_context_sha256` is a new `Option` with `#[serde(default)]`, so every own-commit stamp written before this change decodes as `None`, and `own_commit_stamp_parent_matches` returns `false` for all of them. Those commits then fall through to `process_commit_with_app_data_updates`, which cannot process an own commit — the error prunes the branch.

So on upgrade, any device with a stamped own commit still inside the convergence horizon loses the anchor-rollforward fast path *and* has its own branch pruned during replay. Failing closed is the right direction, but the consequence is a live-node behavior change with no migration, no backfill, and (as far as I can see) no test exercising the legacy `None` case end to end — the unit test only asserts the predicate returns `false`.

Can you confirm the pruning outcome is benign for an upgrading device mid-race, and add a regression test for the legacy-stamp path?

erskingardner (commented):
This is a substantial and carefully-reasoned piece of work — the engine-side comments explaining *why* each seam exists are unusually good, and CI is green. My concerns are about scope and a few specific correctness/perf seams, so I'm commenting rather than approving.

## The premise: this should be at least three PRs

9,424 added lines across 77 files, mixing four independent things:

1. **Production convergence semantics** (`cgka-engine`, `traits`) — fork losers parked `ConvergenceDeferred` instead of `EpochInvalidated`, candidate-branch peel contexts, deferred application admission/replay, anchor-realizable own-commit prefix, admin-only fairness slot. This is the risky part, ~1,200 lines, and it changes branch selection and recovery behavior for every live device.
2. **A new 3,300-line `convergence-campaign-runner` crate** plus Dockerfile, four lane manifests, and three CI workflows.
3. **Simulator/route-assurance additions** (~2,000 lines).
4. **A naming cleanup** and the `naming-gate` recipe.

Only (1) carries production risk, and it is the part that most needs isolated review and an isolated revert path. Right now a regression in the parked-fork-loser semantics cannot be reverted without also reverting the campaign infrastructure and the CI lanes. Please split at minimum (1) from (2)+(3)+(4).

## The premise, part two: is the infrastructure earning its keep yet?

The PR body says Milestone 6 "remains open until a controllable publication or relay-visibility seam stages the #1236 input set across those adapters" — i.e. the process and distributed adapters do not yet deterministically reproduce the counterexample the runner was built to catch. Landing 3,300 lines of runner (external VM delegation, capability gating, evidence bundles, retention/flake/RSS/disk budgets, failure-corpus lifecycle, semantic reduction, recurrence tracking, time-to-diagnosis evidence) *before* the seam that makes it deterministic reads as speculative complexity. Some of these — VM delegation, time-to-diagnosis evidence, recurrence tracking — have no consumer in this PR at all.

What breaks if (2) waits until the staging seam exists and the engine fixes have soaked on master?

## Specific findings

See inline comments. In priority order:

- **`fork_recovery.rs`** — descendant preservation has no state filter and no ancestry check, so previously-terminal commits (`Failed` / `EpochInvalidated` / `LosingBranch`) at later epochs are resurrected as `Retryable`. This is the one I'd call blocking-shaped.
- **`fork_recovery.rs`** — the "single-row write" atomicity rationale is now false (there is a descendant loop), and the block is a multi-write sequence outside a transaction, against the `AGENTS.md` torn-write invariant.
- **`openmls_projection.rs`** — the new parent-GroupContext gate silently disables the own-commit fast path for every stamp written before this change (`None` decodes to no-match) and prunes those branches. No migration and no end-to-end test for the legacy case.
- **`message_processor/mod.rs`** — `note_row_persisted` clearing the gate plus a fingerprint that now includes per-row commit *state* effectively defeats the unproductive-cycle gate during a drain, while each sweep became much more expensive (candidate-branch replay per row). Worst-case cost needs a stated bound.
- **`message_processor/mod.rs`** — fairness slot narrowed to admin intents; application-send starvation bound is unstated.
- **`openmls_projection.rs`** — one no-op assignment (`prefix_canonical = true`).

## Smaller notes (no inline)

- `ingest_group_message` gained a third positional `Option<ConvergenceTime>` whose real meaning is "this is a lifecycle-governed retry, candidate contexts are allowed". Call sites read as `None` / `Some(now)` and lose that meaning. A named two-variant enum would carry it.
- `cgka-conformance-simulator` now depends on `marmot-app` with `test-conformance-snapshot`. Cargo feature unification means a workspace-wide build turns that feature on for `marmot-app` / `cgka-session` / `marmot-account` everywhere in that graph, including the app runtime. The "Production hosts must never enable this feature" comments are not enforced by anything. Worth confirming release artifacts are always built with `-p`, and ideally adding a gate.
- `simulator-nightly.yml` and `convergence-hardening.yml` each duplicate the ~10-line Tamarin install block already present in `ci.yml`. A composite action would keep the pinned version in one place.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -103,6 +103,13 @@ impl DeferredPeelGroupState {
 
     fn note_row_persisted(&mut self) {
         self.deferred_rows += 1;
+        // The direct ingest attempt used only the canonical and retained
+        // past-epoch contexts. A newly retained ciphertext has not yet been
+        // tried against reconstructed competing branch-tip contexts, even if
+        // the group's context fingerprint itself is unchanged. Reopen the
+        // bounded sweep so this row receives that one lifecycle-governed
+        // attempt; its durable per-row fingerprint prevents repeated work.
+        self.gate = None;
```
erskingardner: **The unproductive-cycle gate is now largely defeated, and each sweep got much more expensive.**

Two changes compound here:
- `note_row_persisted` clears `self.gate` on *every* newly persisted deferred row.
- `deferred_peel_context_fingerprint` now folds in the id **and state** of every retained commit row (line ~1290), so the fingerprint changes whenever convergence reclassifies anything.

Inside `drain_convergence_and_deferred_peels`, `retry_deferred_peels_with_time` is now called up to twice per iteration for up to `MAX_CONVERGENCE_REPROCESSING_PASSES = 16` iterations. Because commit states churn across those passes, the fingerprint rarely repeats, so the gate stops suppressing work — and each ungated sweep can process up to `MAX_DEFERRED_ROWS_PER_SWEEP = 64` rows, each of which may call `stored_candidate_peel_contexts` (snapshot create + full candidate-branch MLS replay + rollback, per branch).

That is a large CPU/IO amplification on the inbound hot path, reachable by a peer that can generate deferred rows during contention. Do you have a measured bound for the worst case? A per-drain budget on candidate-context materialization (separate from the per-row/per-pass caps) would make the bound explicit.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -718,47 +725,87 @@ impl<S: StorageProvider> Engine<S> {
...
         }
 
         Ok(false)
     }
 
+    /// Whether a completed convergence generation's one-attempt reservation
+    /// is actively held by queued administrative group-state work. The durable
+    /// fairness bit alone is not sufficient: application sends do not own the
+    /// reservation and must not use it to bypass retained inbound commits.
+    fn convergence_fairness_slot_reserved_for_admin(
```
erskingardner, line 807: This narrows the fairness reservation from *any* queued intent to admin group-state intents only. That is a user-visible behavior change: in a group with sustained inbound commit traffic, a queued application send no longer gets the one-attempt slot and must fully finish catch-up before publishing.

The doc comment states the policy but not the bound. What stops an application send from being starved indefinitely while commits keep arriving? If there is a bound, please state it here; if there is not, this deserves its own test.

crates/cgka-engine/src/fork_recovery.rs
```diff
@@ -164,10 +173,108 @@ impl ForkRecoveryManager {
...
+                    return None;
+                }
+                record.epoch = EpochId(projected_source_epoch);
+                // Unlike the displaced root, a descendant is new branch-growth
+                // evidence that must open the next convergence generation.
+                // `ConvergenceDeferred` is reconsiderable only when some new
+                // Created/Retryable edge opens a pass; using it here would
+                // leave a locally-authored deeper branch parked forever after
+                // the rollback removed its canonical status.
+                record.state = MessageState::Retryable;
```
erskingardner: Re-affirming the earlier review's blocking-shaped item with the mechanics verified at this head: the descendant scan has no `record.state` filter and no ancestry check. Every commit-kind row stored at a later epoch is re-persisted as `Retryable` — including rows convergence already classified terminal (`LosingBranch` from an earlier pass), rows rejected `Failed`, and forks rooted at an *earlier* epoch that never descended from this incumbent. Resurrection re-admits them into the next pass; convergence should re-reject them, so the cost is replay budget (the fail-closed resource) and a broken terminal-state invariant rather than permanent divergence — but an attacker who previously got commits stored and terminally rejected can inflate later passes by triggering a pairwise CandidateWins. Suggest filtering to the states that can legitimately contribute (`Sent | Created | Retryable | ConvergenceDeferred | Processed`, mirroring `record_state_can_contribute_to_openmls_graph`).

crates/cgka-engine/src/fork_recovery.rs
```diff
@@ -164,10 +173,108 @@ impl ForkRecoveryManager {
...
+        // last step — the point of no return.
+        //
+        // The snapshot operations necessarily sit OUTSIDE any storage
+        // transaction: create/rollback/release drive their own SQLite
+        // transactions (see the boundary comment in
+        // `openmls_projection::apply_openmls_canonicalization_result_inner`).
+        // The re-persist is a single-row write and is already atomic on its
+        // own, so there is no multi-write group here worth wrapping in
+        // `with_transaction` — and this manager is generic over
+        // `MessageStorage`, which does not expose one.
```
erskingardner: This rationale is now false: the re-persist is the incumbent row *plus* the `displaced_descendant_commits` loop immediately below — 1+N writes with rollback/release interleaved, not "a single-row write ... no multi-write group here". The repo's torn-write invariant asks for exactly this kind of sequencing to be owned explicitly. Either update the comment to describe the actual N-row sequence and its compensation story, or restructure (e.g., validate and stage all rows before the first write).

crates/convergence-campaign-runner/src/plan.rs
```diff
@@ -0,0 +1,829 @@
...
+                participant_container(participant),
+                "stress-ng".into(),
+                "--hdd".into(),
+                workers.to_string(),
+                "--hdd-bytes".into(),
+                bytes_per_worker.to_string(),
+                "--temp-path".into(),
+                participant_state_path(participant),
+                "--timeout".into(),
+                format!("{duration_ms}ms"),
```
erskingardner: `stress-ng --timeout` has no `ms` suffix. The documented units are s/m/h/d/y and stress-ng's time parser scales by the *last* character of the argument, so `--timeout 5000ms` is read as 5000 **seconds** (~83 minutes) — a fault declared as `duration_ms: 5000` runs ~1000× long, bounded only by the later `pkill` in `StopDatabaseContention`. No test asserts the rendered value and no e2e lane exercises this fault, so nothing catches it. Render `{duration_ms / 1000}s` (and validate divisibility in manifest validation) instead.

crates/convergence-campaign-runner/src/runner.rs
```diff
@@ -0,0 +1,934 @@
...
+}
+
+async fn run_vm(
+    manifest: &DistributedCampaignManifestV1,
+    plan: &DistributedExecutionPlanV1,
+) -> Result<DistributedRunReceiptV1, RunnerError> {
+    let command = plan.vm_driver.as_ref().ok_or_else(|| {
+        RunnerError::validation("vm_plan", "VM plan is missing its external driver")
+    })?;
+    let executed = execute(command, CommandContext::default()).await;
```
erskingardner: The external VM driver is executed through `execute()`, which wraps every command in `INFRASTRUCTURE_COMMAND_TIMEOUT` (90s). The driver is synchronous — its exit *is* campaign completion — so any real VM campaign (kernel, block-device, or filesystem faults across VMs) can never finish and will always end as `command_timeout` at 90 seconds. The VM lane is unusable as implemented; it needs a campaign-scale timeout (or none, deferring to the lane's declared wall budget).

crates/cgka-conformance-simulator/src/reference_convergence.rs
```diff
@@ -159,6 +159,51 @@ pub struct ReferenceResult {
...
+    pub reconsiderable_branch_ids: BTreeSet<String>,
+}
+
+/// Evaluate the same durable input set through a named production route.
+///
+/// Pairwise recovery is modeled as a provisional decision only: its displaced
+/// branch remains materializable until the shared canonical evaluator assigns
+/// a terminal disposition. This is the rule #1236 showed must be identical at
+/// every seam.
+pub fn evaluate_via_route(
```
erskingardner: `evaluate_via_route` never reads `route`: every variant calls the same `evaluate(input)` and merely stores the label (`PairwiseForkRecovery { provisional_winner }` is unused). `every_decision_route_reaches_the_same_reference_result` therefore asserts `evaluate(input) == evaluate(input)` five times — it is tautological with respect to route equivalence and would pass even if route-dependent behavior regressed tomorrow. The lifecycle-model test below (`route_choice_and_restart_do_not_change_the_canonical_winner`) is a genuine differential, so coverage exists — but this test's doc comment ("evaluate the same durable input set through a named production route") overstates what exists, and claim-tracking should not count it as route-equivalence evidence. Delete it, make the reference model actually consume the route, or rename it honestly.

crates/cgka-conformance-simulator/tests/route_equivalence.rs
```diff
@@ -0,0 +1,351 @@
...
+    assert!(checkpoints.contains(&RouteRestartCheckpoint::AfterObserverRouting));
+    assert!(checkpoints.contains(&RouteRestartCheckpoint::AfterCommitterRouting));
+    for campaign in &campaigns {
+        assert_eq!(
+            campaign
+                .covered_claims
+                .iter()
+                .map(String::as_str)
+                .collect::<BTreeSet<_>>(),
+            ROUTE_ASSURANCE_CLAIMS
```
erskingardner: Every generated campaign declares all six `ROUTE_ASSURANCE_CLAIMS` in `covered_claims` unconditionally, but the app-runtime and process adapter campaigns assert agreement/decryptability without ever asserting that the #1236 fork actually occurred (no epoch/roster/loser-ledger assertion like the engine test's `epoch == 3` + membership checks). The PR body itself states those adapters don't deterministically stage the same-source-epoch input set — fork staging there depends on live-subscription timing — so `reconsiderable-loser` can be recorded `Covered` by a run that never created a competing commit. Scope each adapter's covered claims to what its campaign verifiably stages, or add an explicit fork-occurrence assertion per adapter.

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -432,6 +473,16 @@ impl AppRuntimeHarness {
...
+        let canonical_state = participant
+            .cached_exact
+            .get(&group_label)
+            .cloned()
+            .ok_or_else(|| {
+                SubjectError::new(
+                    "exact_state_unavailable",
+                    "exact app-runtime state has not been refreshed",
+                )
+            })?;
```
erskingardner, line 486: Since `cached_exact` is cleared on `set_online(false)` and `refresh_cached_members` skips offline participants, observing an offline participant now always fails with `exact_state_unavailable`. The engine adapter can observe offline subjects, and the `AppRuntimeLocalDiagnosticsV1` block (`online: false`, `database_exists`, failure counters) exists precisely for the offline case but is now unreachable. CI is green so no current scenario observes an offline client — but this quietly shrinks the observable surface relative to the other adapters and pushes scenario authors away from offline observation. If intentional, document it in the adapter's capability notes; otherwise return the cached/diagnostic observation for offline participants.

crates/cgka-conformance-simulator/src/bin/cgka-conformance-node.rs
```diff
@@ -1,12 +1,47 @@
...
+    let mut args = std::env::args().skip(1);
+    if let Some(argument) = args.next() {
+        if argument != "--relay-proxy" {
+            return Err("usage: cgka-conformance-node [--relay-proxy HOST:PORT]".into());
+        }
+        let upstream = args.next().ok_or("missing relay proxy upstream")?;
+        if args.next().is_some() {
+            return Err("unexpected node argument".into());
+        }
+        let listener = TcpListener::bind("127.0.0.1:18080").await?;
```
erskingardner: The proxy listen port is hardcoded `127.0.0.1:18080`, and the campaign runner hardcodes the same port as each node's relay URL (`ws://127.0.0.1:18080`). Today every node runs in its own container net namespace so nothing collides — but the coupling is invisible, and any host-networked or same-loopback multi-node use has every node after the first fail to bind and exit before the stdio protocol starts. Consider `--relay-proxy HOST:PORT [LISTEN:PORT]` (or binding port 0) and reporting the actual bound port.

erskingardner (commented):
Second full pass at head `f7e8933` (unchanged since the previous review). I re-verified the earlier findings against the current tree and ran an independent sweep of the engine, simulator, campaign-runner, and workflow diffs. Posting as a comment again: the scope concern from the previous review stands, and the new findings are worth fixing but none change the overall shape of the recommendation.

## Status of the previous findings

- **Descendant preservation without state filter/ancestry check** — still present, still the one I'd call blocking-shaped for the engine part. Inline with mechanics and a suggested one-line filter.
- **"Single-row write" atomicity rationale** — still false at head (1+N writes). Inline.
- **Unproductive-cycle gate + fingerprint now includes per-row commit state, while each sweep got much more expensive** (full candidate-graph rebuild per deferred row) — still open; worst-case sweep cost needs a stated bound.
- **Fairness slot narrowed to admin intents** — now documented in code as deliberate ("application sends do not own this reservation and must finish catch-up before publishing"). The design is fine; the application-send starvation bound is still unstated.
- **Legacy stamps decode `None` and lose the anchor fast path** — partially mitigated: `traits/src/message.rs` now documents this as intentional. The remaining gap is the absence of any end-to-end test for a pre-upgrade stamped row, and the failure mode (replay of a branch containing such a commit fails closed) deserves one sentence in the changelog entry.
- **`prefix_canonical = true` no-op** — still there; trivial.

## New findings this pass (inline unless noted)

**Major**

1. `plan.rs` — `stress-ng --timeout {duration_ms}ms` is parsed as **seconds**, not milliseconds: a declared 5s database-contention fault runs ~83 minutes. 1000× units bug, uncovered by any test.
2. `runner.rs` — the external VM driver inherits the 90s `INFRASTRUCTURE_COMMAND_TIMEOUT`, so the entire VM lane fails as `command_timeout` before any real campaign can finish. The lane is unusable as implemented.
3. `reference_convergence.rs` — `evaluate_via_route` is route-independent by construction; `every_decision_route_reaches_the_same_reference_result` is tautological and must not count as route-equivalence evidence.
4. `route_equivalence.rs` — every campaign stamps all six `ROUTE_ASSURANCE_CLAIMS` covered regardless of whether the adapter staged the fork; per the PR's own open item, process/distributed adapters don't deterministically stage it, so claims can be recorded `Covered` without a competing commit ever existing.

**Minor** (details inline or below)

5. `app_runtime.rs` — offline participants are now unobservable (`exact_state_unavailable`); diverges from the engine adapter and makes the offline diagnostics block unreachable.
6. `cgka-conformance-node.rs` — hardcoded `127.0.0.1:18080` proxy port, invisibly coupled to the runner's hardcoded relay URL.
7. `simulator-nightly.yml` — the job keeps `timeout-minutes: 120` while gaining `just tamarin` (the standalone formal job budgets 45 min for the same proofs), a cold no-cache release `docker build` of the whole workspace, and a real-container campaign. Likely to time out; raise it or split.
8. `manifest.rs` — the VM capability model is self-contradictory: `required_vm_capabilities()` only ever maps `SlowBlockDevice → BlockDeviceLatency`, so three of four capabilities are dead, and `validate()` rejects any VM manifest without `SlowBlockDevice` (`unnecessary_vm_backend`) — a host-isolation-only VM campaign is unrepresentable despite the README naming host isolation as a VM purpose.
9. `lane.rs` — `check-budget` passes a zero-case observation (flake rate 0, no violations): a `min_executed_cases` floor would close the "ran nothing, passed budget" hole feeding evidence bundles.
10. `evidence.rs` — `check-evidence` validates artifact digest *format* only; referenced files are never hashed, so the completeness contract is self-attested.

## Verification of earlier bot findings

Spot-checked the coderabbit claims against head: the receipt-on-error, run-token placeholder, non-UTF-8 argv, crash-rejoin/disk-release, case-insensitive digest, stale offline cache, schema gating, probe recording, ledger totality, and claim-history findings are all genuinely fixed with regression tests. The ipset one-shot name-resolution concern is real but mitigated by the partition/crash ordering validation; the riskiest faults (partition/heal, disk, contention) still have plan-shape tests only — at least the nightly e2e slot should cover one partition/heal cycle against a real daemon.

## What I verified in the engine changes

The core production change (park pairwise losers `ConvergenceDeferred` keyed by source epoch, candidate-branch peel contexts, deferred app admission/replay, parent/result GroupContext-bound own-commit stamps, admin-only fairness slot) is coherent and unusually well-commented; the cross-seam tests in `fork_detection.rs` are genuine end-to-end differential coverage. I traced the `msg.id` content-id rebinding, the candidate-admission pass membership, the stamp parent/result gates, and the own-anchor rewind path — all consistent. The engine part of this PR is close; the descendant state filter and the torn-write comment are the two things I'd want fixed before it lands, and the scope split (engine vs. campaign infrastructure) from the previous review remains the main process ask.

Datawav:

## Final audit

The PR’s convergence and campaign infrastructure is needed to test cross-route behavior under realistic process, container, and VM faults. The container-first design, canonical scenario format, bounded recovery work, and exact-state evidence are reasonable. Three actionable gaps remain:

- **High — unverified CI binary is installed as root.** `.github/workflows/convergence-hardening.yml:52` and `.github/workflows/simulator-nightly.yml:53` download a Tamarin archive and install its executable with `sudo`, but never verify a checksum or signature. Pinning the version and using HTTPS does not protect against a compromised release artifact. Verify a repository-pinned SHA-256 before extraction, or use a trusted installer action pinned to a commit.

- **Medium — fork recovery removes unrelated retransmission records.** In `crates/cgka-engine/src/fork_recovery.rs`, the new `live_messages.retain` excludes every stored row whose decoded MLS payload equals `candidate_mls_bytes`, despite intending to exclude only the row owned by the current ingest. Identical retransmissions can have distinct transport IDs and dispositions, so rollback reconstruction can silently discard their forensic and processing state. Pass the candidate’s `MessageId` into `resolve` and exclude exactly that record. Add coverage with two distinct IDs containing identical MLS bytes. Comparing payload bytes is simpler, but cannot preserve transport-record identity.

- **Medium — VM timeout does not bound the distributed campaign.** `crates/convergence-campaign-runner/src/runner.rs:570` wraps `child.status()` in a timeout with `kill_on_drop(true)`. This terminates only the direct driver; descendants or provisioned VMs may survive indefinitely, while the VM plan provides no cleanup path. Run the driver in an isolated process group and terminate the group on timeout, and define a cleanup/cancellation command in the driver contract. Alternatively, require drivers to return a cleanup handle and invoke it on all failure paths.

The advisory concerns about heal rollback semantics, receipt-error precedence, and context-materialization constants are robustness improvements without demonstrated correctness impact, so they are not included as findings.

_Advisory audit of head [`801866cfabc0`](https://github.com/marmot-protocol/mdk/commit/801866cfabc0ce55b666041146aef5d691c45d6a); no approval, merge, push, checkout, or PR-code execution was performed._

erskingardner:
## Split in progress

To give the production convergence changes an isolated decision and revert path, the independent work from this PR has been extracted into smaller drafts:

- #1270 — distributed campaign runner core (based on `master`)
- #1271 — execution lanes and evidence budgets (stacked on #1270)
- #1272 — failure corpus and semantic reduction (stacked on #1271)
- #1273 — capability naming guard and wording cleanup (independent, based on `master`)

The replacement PRs intentionally contain no production convergence-semantic changes and do not assume any resolution of #1236.

The following remains parked in #1250 pending team design discussion around #1236:

- same-epoch fork-selection and loser-reconsideration behavior
- candidate-context peeling, deferred application admission/replay, and recovery semantics
- route-assurance scenarios and claims whose expected result depends on those semantics
- the cross-layer canonical-state/decryptability surface added specifically to close that decision route

This PR is being returned to draft so it is not treated as a merge candidate while that discussion is unresolved. We should retire or replace the parked portion only after the team agrees on the protocol/engine contract; the simulator infrastructure PRs do not prejudge that choice.

erskingardner:
Broken down in favor of several smaller PRs (#1270, #1271, #1272, #1273)

PR #1257, opened 2026-08-04
fix(openclaw): bound timeline context size

erskingardner:
## Summary

- retain at most the newest eight timeline records in the OpenClaw inbound `chat_window` supplemental entry
- enforce a 16 KiB UTF-8 budget over the complete serialized entry, including the envelope, records, and aggregate truncation metadata
- drop oldest records first, preserve retained chronological order, and omit an oversized remaining record rather than exceeding the budget
- report only aggregate omitted/oversized counts; referenced-message and ambient context behavior unchanged
- cover small, oversized, count-window, chronological, aggregate-metadata, and twenty-record cases through the inbound dispatch path

## Scope

OpenClaw parity with the Hermes adapter-local bound from #1234, addressing the OpenClaw portion of #1223. It deliberately does not implement the broader host-session watermark/delta design from that issue.

The control protocol's per-record and frame limits remain unchanged; this adds the missing OpenClaw-local rendered-context bound. The byte budget is measured over `JSON.stringify` of the full `{label, source, type, payload}` entry, the OpenClaw host's rendered form (the Hermes bound measures its prefixed string line).

## Validation

- `pnpm typecheck` — clean
- `pnpm test` — 205 passed, 1 skipped (22 files), including six new `timeline context budget` cases in `test/dispatch.test.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Conversation history shared with integrations is now limited to the eight most recent messages and 16 KiB.
  * Oversized or older messages are automatically omitted, with truncation details provided when applicable.
  * Newer conversation records are prioritized when the history exceeds the available limits.

### DIFF
--- a/integrations/openclaw/marmot/src/dispatch.ts
+++ b/integrations/openclaw/marmot/src/dispatch.ts
@@ -235,6 +235,96 @@ function replyFallbackBody(availability: string): string {
 }
 
 /**
+ * Aggregate bound for the rendered conversation-history supplemental entry,
+ * mirroring the Hermes adapter-local bound (#1223): retain at most the newest
+ * eight records and keep the complete serialized entry — envelope, records,
+ * and aggregate truncation metadata — within 16 KiB of UTF-8.
+ */
+export const TIMELINE_CONTEXT_MESSAGE_LIMIT = 8;
+export const TIMELINE_CONTEXT_BYTE_LIMIT = 16 * 1024;
+
+interface TimelineContextPayload {
+  order: "chronological";
+  relation: "before_current_message";
+  messages: AgentControlTimelineMessage[];
+  messages_truncated?: boolean;
+  omitted_message_count?: number;
+  oversized_message_count?: number;
+}
+
+function timelineContextEntry(payload: TimelineContextPayload) {
+  return {
+    label: "Marmot conversation history",
+    source: "marmot",
+    type: "chat_window",
+    payload,
+  };
+}
+
+/** UTF-8 size of the complete serialized supplemental entry. */
+function timelineContextEntryBytes(payload: TimelineContextPayload): number {
+  return Buffer.byteLength(JSON.stringify(timelineContextEntry(payload)), "utf8");
+}
+
+/**
+ * Whether one record alone already exceeds the budget, judged against an entry
+ * carrying the truncation metadata that would accompany its omission.
+ */
+function timelineMessageExceedsByteLimit(message: AgentControlTimelineMessage): boolean {
+  return (
+    timelineContextEntryBytes({
+      order: "chronological",
+      relation: "before_current_message",
+      messages: [message],
+      messages_truncated: true,
+      omitted_message_count: 1,
+    }) > TIMELINE_CONTEXT_BYTE_LIMIT
+  );
+}
+
+/**
+ * Build the bounded conversation-history entry: drop oldest records first,
+ * keep retained records in chronological order, and omit an oversized
+ * remaining record rather than exceed the byte budget. Reports only aggregate
+ * omitted/oversized counts (privacy-safe).
+ */
+function boundTimelineContextEntry(history: AgentControlTimelineMessage[]) {
+  const bounded = history.slice(-TIMELINE_CONTEXT_MESSAGE_LIMIT);
+  let omittedMessageCount = history.length - bounded.length;
+  let oversizedMessageCount = 0;
+  for (const message of history.slice(0, omittedMessageCount)) {
+    if (timelineMessageExceedsByteLimit(message)) {
+      oversizedMessageCount += 1;
+    }
+  }
+
+  for (;;) {
+    const payload: TimelineContextPayload = {
+      order: "chronological",
+      relation: "before_current_message",
+      messages: bounded,
+      ...(omittedMessageCount > 0
+        ? { messages_truncated: true, omitted_message_count: omittedMessageCount }
+        : {}),
+      ...(oversizedMessageCount > 0 ? { oversized_message_count: oversizedMessageCount } : {}),
+    };
+    if (timelineContextEntryBytes(payload) <= TIMELINE_CONTEXT_BYTE_LIMIT) {
+      return timelineContextEntry(payload);
+    }
+    const dropped = bounded.shift();
+    if (dropped === undefined) {
+      // The metadata-only envelope is intentionally tiny, so this is a
+      // defensive fallback rather than an expected path.
+      return timelineContextEntry(payload);
+    }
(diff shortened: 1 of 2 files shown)


integrations/openclaw/marmot/src/dispatch.ts
```diff
@@ -234,6 +234,96 @@ function replyFallbackBody(availability: string): string {
...
+      ...(oversizedMessageCount > 0 ? { oversized_message_count: oversizedMessageCount } : {}),
+    };
+    if (timelineContextEntryBytes(payload) <= TIMELINE_CONTEXT_BYTE_LIMIT) {
+      return timelineContextEntry(payload);
+    }
+    const dropped = bounded.shift();
+    if (dropped === undefined) {
+      // The metadata-only envelope is intentionally tiny, so this is a
+      // defensive fallback rather than an expected path.
+      return timelineContextEntry(payload);
```
erskingardner: Non-blocking: this branch is unreachable. If `bounded` is empty the payload is metadata-only (a few hundred bytes at most), so the `<= TIMELINE_CONTEXT_BYTE_LIMIT` check on line 311 always returns before the array can be drained to nothing — and the comment itself says as much.

Dropping lines 315-319 and letting the loop be `bounded.shift()!`-free would be a small net deletion:

```ts
    const dropped = bounded.shift() as AgentControlTimelineMessage;
    if (timelineMessageExceedsByteLimit(dropped)) {
```

I recognize this mirrors the Hermes fallback verbatim, so keeping it for cross-host parity is a defensible answer — but then it's worth saying "kept for parity with `_timeline_channel_context`" rather than "defensive fallback", since the two justify very different things.

integrations/openclaw/marmot/src/dispatch.ts
```diff
@@ -234,6 +234,96 @@ function replyFallbackBody(availability: string): string {
...
+ */
+function boundTimelineContextEntry(history: AgentControlTimelineMessage[]) {
+  const bounded = history.slice(-TIMELINE_CONTEXT_MESSAGE_LIMIT);
+  let omittedMessageCount = history.length - bounded.length;
+  let oversizedMessageCount = 0;
+  for (const message of history.slice(0, omittedMessageCount)) {
+    if (timelineMessageExceedsByteLimit(message)) {
+      oversizedMessageCount += 1;
+    }
+  }
```
erskingardner, line 302: Non-blocking, semantics question: this pre-scan counts records that were dropped purely by the **count** window (the 8-record cap) toward `oversized_message_count`, even when the byte budget never came into play.

Concretely: a 20-record page of small messages where record #3 happens to be huge renders as `omitted_message_count: 12, oversized_message_count: 1` — but nothing was actually omitted *because of* size; record #3 was going to be dropped as old regardless. The model reading this untrusted context sees a size signal that didn't drive any decision.

If `oversized_message_count` is meant to answer "did we have to throw away content to fit the budget?", these 5 lines are both extra code and misleading output, and deleting them makes the field mean exactly that. If it's meant to answer "how much of the omitted history was individually unrenderable?", the current behavior is right and the doc comment on line 288-289 could say so. Either way it matches the merged Hermes behavior, so this is a question about the shared contract rather than about this PR alone.

integrations/openclaw/marmot/src/dispatch.ts
```diff
@@ -234,6 +234,96 @@ function replyFallbackBody(availability: string): string {
...
+  );
+}
+
+/**
+ * Build the bounded conversation-history entry: drop oldest records first,
+ * keep retained records in chronological order, and omit an oversized
+ * remaining record rather than exceed the byte budget. Reports only aggregate
+ * omitted/oversized counts (privacy-safe).
+ */
+function boundTimelineContextEntry(history: AgentControlTimelineMessage[]) {
```
erskingardner, line 294: Non-blocking: with this bound in place, the `limit: 20` on the `timelineList` call (line ~496) now fetches at least 12 records that can never be rendered — they exist only to be serialized once each by `timelineMessageExceedsByteLimit` and counted.

That's a deliberate tradeoff (fetching 20 is what makes the exact `omitted_message_count` possible, which `has_more_before` alone couldn't give you), but nothing at the call site says so, and the next reader will reasonably file the 20 as stale-after-the-8-cap. A one-line comment at the `limit: 20` site pointing at `TIMELINE_CONTEXT_MESSAGE_LIMIT` would keep the two numbers from drifting apart.

erskingardner (commented):
**Recommend merge.** This is a tight, well-scoped port of the merged Hermes bound (#1234) to the OpenClaw dispatch path, and the tests are unusually good — the "does not misclassify a record displaced by aggregate metadata" case exercises exactly the boundary where a naive implementation would over-report `oversized_message_count`.

I walked the algorithm against the edge cases and found no correctness defects:

- `history.slice(-8)` copies, so the `shift()` loop never mutates caller state.
- `omittedMessageCount` / `oversizedMessageCount` stay consistent (oversized is always a subset of omitted).
- The loop terminates: each iteration either returns or shrinks `bounded`, and the metadata-only payload always fits.
- Premise is sound — #1223 explicitly lists an aggregate byte budget as an acceptance criterion, and the PR is honest that it does not attempt the watermark/delta half.

Three non-blocking notes inline. None need to hold the merge; the first two are simplification/clarity, the third is a comment request.

(Posted as a comment rather than an approval: GitHub does not allow approving your own PR.)

integrations/openclaw/marmot/src/dispatch.ts
erskingardner: Done in 720b112f — deleted the unreachable branch. The drop site now uses `bounded.shift() as AgentControlTimelineMessage` with a comment explaining why the cast is safe: an empty `bounded` renders a metadata-only payload that always fits the budget, so the budget check returns before `shift` can drain it.

integrations/openclaw/marmot/src/dispatch.ts
erskingardner: Kept the pre-scan and documented the shared contract in 720b112f: the doc comment now states that `oversized_message_count` counts every individually unrenderable omitted record — including records dropped by the count window, where size never drove the decision — matching the merged Hermes behavior. Agreed that if the semantics should change to "dropped because of size", that is a both-hosts follow-up rather than this PR.

integrations/openclaw/marmot/src/dispatch.ts
erskingardner: Done in 720b112f — the `limit: 20` site now notes the over-fetch beyond `TIMELINE_CONTEXT_MESSAGE_LIMIT` is deliberate so the bounded entry can report an exact `omitted_message_count`.

PR #1268, opened 2026-08-05
Fix Hermes installer guided setup

erskingardner:
## Summary

- replace the repeated Hermes sender prompt with one comma-separated allowlist entry
- make a blank guided sender entry explicitly persist `MARMOT_ALLOW_ALL_USERS=true`
- prevent `hermes plugins enable marmot` from consuming the remaining source of a `curl | bash` installation
- document and test the guided and noninteractive behaviors

## Root cause

Hermes can prompt while enabling the Marmot plugin. When the installer itself is supplied on stdin, that child command inherited the same stream and could consume the remaining installer source. The install then stopped after plugin enablement, before bootstrap, service setup, and Hermes authorization/configuration completed.

The guided sender flow also prompted once per sender even though the underlying parser already accepted comma-separated values, making a repeated entry look necessary.

## User impact

The documented one-line installer can complete reliably when streamed through Bash. Guided installs now accept all sender identities in one visible step; leaving the step blank clearly opts into accepting every sender. Noninteractive installs remain fail-closed unless `--allow-user` or `--allow-all-users` is supplied.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s integrations/hermes/marmot/tests` (168 tests)
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Streamlined guided setup for sender authorization with a single comma-separated input.
  * Leaving the sender prompt blank now explicitly allows messages from all senders.
  * Noninteractive installation supports comma-separated values for `--allow-user`.

* **Documentation**
  * Updated installation guidance to explain sender authorization options and blank-input behavior.

### DIFF
--- a/integrations/hermes/marmot/README.md
+++ b/integrations/hermes/marmot/README.md
@@ -47,16 +47,20 @@ One-line install:
 curl -fsSL "https://github.com/marmot-protocol/mdk/releases/download/wn-agent-latest/install-hermes-marmot.sh" | bash
 ```
 
+In guided setup, the allowed-message-sender prompt accepts one or more `npub` or
+raw hex public keys separated by commas. Leaving that prompt blank explicitly
+opens Marmot messaging to every sender by writing
+`MARMOT_ALLOW_ALL_USERS=true`.
+
 For repeatable noninteractive setup, pass the allowed inviter/welcomer and allowed
 message sender as either an `npub` or raw hex public key. `--allow-user` may be
-repeated to authorize multiple senders:
+repeated or given a comma-separated list to authorize multiple senders:
 
 ```sh
 curl -fsSL "https://github.com/marmot-protocol/mdk/releases/download/wn-agent-latest/install-hermes-marmot.sh" | \
   bash -s -- --yes \
     --allow-welcomer npub1... \
-    --allow-user npub1... \
-    --allow-user 0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
+    --allow-user npub1...,0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
 ```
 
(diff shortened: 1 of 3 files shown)


Datawav:

## Final audit

### High — Guided setup turns a blank/default response into unrestricted sender access

`scripts/install-hermes-marmot.sh`, around line 1428, now sets `EXPLICIT_ALLOW_ALL=1` whenever the sender prompt is left blank. Previously, unrestricted messaging required a separate explicit yes/no confirmation whose default was “no.”

This changes a security boundary from fail-closed to fail-open: users accustomed to pressing Enter through guided defaults can unintentionally persist `MARMOT_ALLOW_ALL_USERS=true`. Updating the prompt text and documentation does not provide an affirmative opt-in equivalent to the removed confirmation. The revised integration test codifies the unsafe behavior by expecting blank input to enable all senders.

Keep the consolidated comma-separated prompt, but when it is blank, require an explicit confirmation defaulting to “no,” or leave sender authorization unset and explain how to rerun with `--allow-all-users`.

The PR addresses two legitimate problems:

- Isolating `hermes plugins enable marmot` from installer stdin correctly prevents an interactive child from consuming the remaining `curl | bash` script.
- Consolidating repeated sender prompts into one comma-separated entry improves guided setup.

The stdin redirection is narrow and appropriately tested. The sender-prompt redesign is reasonable except for treating blank input as unrestricted access. A safer design retains the single prompt while preserving explicit confirmation for allow-all. Alternatively, blank input could remain fail-closed and direct users to provide sender keys or pass `--allow-all-users`.

The advisory concern about comma-separated values being stored verbatim is not actionable from the supplied evidence: the PR states that the downstream parser already accepts comma-separated values, and the patch does not establish a resulting failure.

PR #1269, opened 2026-08-05
Resolve the Android private-root path alias instead of skipping it

mubarakcoded:
Follow-up to #1267. The Android anchor added there cannot reach the sandbox root on a physical device, because the canonical path Android hands the client contains a symlinked component that the strict probe correctly refuses to traverse. This resolves that component by verification instead, so the descriptor walk never has to forgive anything.

Bootstrap failed on a Pixel 9 Pro XL with:

```
locate openable directory-walk anchor at /data/user/0/<pkg>/files/Marmot:
open directory-walk anchor candidate at /data/user/0: Not a directory (os error 20)
```

## What `/data/user/0` actually is

It depends on who is looking, which is why an earlier reading of this component was wrong. `run-as` shares adb shell's mount namespace, not the app's. Read through `/proc/<app pid>/root`, which resolves in the app's own namespace:

```
app namespace (authoritative)
/            0755 root:root   directory
/data        0771 system:system directory      (group-write set)
/data/user   0751 root:root   directory        (tmpfs)
/data/user/0 0777 root:root   SYMLINK -> /data/data
/data/data   0751 root:root   directory        (tmpfs)
/data/data/<pkg>       0700 uid 10491          (bind-mounted in)

shell namespace (what an earlier probe saw)
/data/user/0  directory, inode 315, same inode as /data/data, own /proc/mounts entry
```

Android's app-data isolation gives each app a tmpfs `/data/user` and `/data/data` containing only its own package, and on that tmpfs `/data/user/0` is a plain root-owned symlink.

Kernel behaviour, confirmed with a cross-compiled probe on the device rather than assumed:

```
symlink -> dir :  O_DIRECTORY = OK    O_DIRECTORY|O_NOFOLLOW = ENOTDIR(20)   O_NOFOLLOW = ELOOP(40)
plain file     :  O_DIRECTORY = ENOTDIR(20)   O_DIRECTORY|O_NOFOLLOW = ENOTDIR(20)
```

So `ENOTDIR` cannot distinguish "symlink" from "not a directory", and the walk anchor is right to reject it rather than guess. The `EACCES` reported for this component in #1267 came from probing without `O_NOFOLLOW`, which followed the link to `/data/data`.

## Approach

`resolve_root_owned_alias` is replaced by `resolve_platform_directory_aliases`. It scans prefix components at any depth, not just the first, and when one is a symlink it rewrites it to its target only after verifying that:

- the symlink is owned by a trusted platform id;
- every directory from `/` to the symlink's parent, and from `/` to the resolved target, is a real directory, trusted-owned, not world-writable, and group-writable only by a trusted gid;
- the target is a plain path with no `.` or `..` component;
- the leaf is never resolved — that stays `O_NOFOLLOW`'s decision.

Trusted ids are `[0]`, plus `1000` on Android only. Uid 1000 owns `/data` there, while on desktop Linux it is an ordinary user account, so trusting it off-Android would be a hole. A test pins that.

Two details are sharper than they may look. Verifying only the alias and its target is insufficient — the whole ancestor chain has to be checked, or write access to an intermediate directory would allow swapping the alias. And a flat "not group-writable" rule rejects the legitimate Android alias, because `/data` is `0771 system:system` with group-write set; group-write is therefore accepted only when the gid is itself trusted.

## This closes the CWE-59 finding rather than reopening it

`open_walk_anchor`'s strict reject arm from #1267 is untouched, and no blanket `ENOTDIR` skip was added back. An unverified symlink in the prefix is now rejected *earlier* than before — during normalization rather than at the walk. The path handed to the walk contains no symlink at all, so every component below the anchor keeps its `O_NOFOLLOW` check and nothing is skipped on trust.

`PreparedDirectory::path` now reports the canonical `/data/data/<pkg>/...` spelling on Android, mirroring the existing `/private/var/...` behaviour on macOS. No caller compares that value against its input; all five were checked.

## Tests

Twelve new tests, written failing first. Ownership cannot be faked as non-root on a host, so the predicate takes its facts through an injected inspector for the Android shape, while the accept path also has real coverage: on macOS the genuine root-owned `/var -> private/var` alias is exercised directly, and every existing tempdir test now runs through the resolver because `TMPDIR` lives under `/var/folders`.

Mutation-checked, each applied in isolation and reverted:

- removing the symlink-owner check fails 1 test;
- neutralizing the ownership and writability predicate fails 4;
- removing both fails 6, including the real-filesystem canary.

That canary initially passed even with all verification removed, because the macOS temp path spells `/var` and a structural check rejected it for the wrong reason. It now canonicalizes first, so only ownership can reject it.

## Verification

| Command | Result |
|---|---|
| `cargo fmt --all -- --check` | pass |
| `cargo clippy -p fs-private --all-targets -- -D warnings` | pass |
| `cargo test -p fs-private -- --test-threads=1` | 43 passed, 0 failed |
| `cargo check -p marmot-app` | pass |
| `cargo check -p fs-private --target aarch64-linux-android` | pass |
| `cargo clippy -p fs-private --target aarch64-linux-android -- -D warnings` | pass |

On a Pixel 9 Pro XL with bindings built from this commit: bootstrap succeeds, the resolved root matches the pre-regression path, the existing store is read in place, and the chat list populates. Zero permission errors, zero anchor errors, zero exceptions.

Unrelated observation while working here: `cargo test -p fs-private` is flaky at roughly 50% on a macOS host on master, in the two flock lease tests. The lib test binary re-execs itself with a filter, so a concurrently spawned child inherits a duplicate of another test's lease descriptor and `drop` does not release the lock. CI runs Linux only, so it does not surface there. Left untouched; happy to file it separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved directory path handling for trusted platform aliases.
  * Added stronger protection against unsafe symlinks, writable parent directories, invalid targets, and untrusted ownership.
  * Improved support for Android-specific directory aliases.
  * Added handling for missing path components and leaf directories.

### DIFF
--- a/crates/fs-private/src/lib.rs
+++ b/crates/fs-private/src/lib.rs
@@ -218,18 +218,20 @@ pub fn prepare_directory_path(
         )
     })?;
 
-    // macOS exposes stable root aliases such as `/var -> private/var` and
-    // `/tmp -> private/tmp`. Resolve only such a first, root-owned component;
-    // every caller-controlled component below it is still opened with
-    // `O_NOFOLLOW`. Without this narrow normalization, ordinary temp-backed
-    // homes could never use the descriptor walk.
+    // Platforms spell parts of a private root through symlinks they own
+    // themselves: macOS `/var -> private/var`, and `/data/user/0 -> /data/data`
+    // inside an Android app's mount namespace. Rewrite only such verified
+    // platform aliases out of the prefix; every caller-controlled component
+    // below them is still opened with `O_NOFOLLOW`. Without this normalization,
+    // neither ordinary temp-backed homes nor an Android app's own data
+    // directory could use the descriptor walk.
     // iOS App Group APIs return the canonical `/private/...` spelling. Avoid
     // even probing its first sandbox-managed ancestor independently; the
     // complete-path open below is the authorization boundary.
     #[cfg(target_os = "ios")]
     let normalized_path: Option<PathBuf> = None;
     #[cfg(not(target_os = "ios"))]
-    let normalized_path = resolve_root_owned_alias(path)?;
+    let normalized_path = resolve_platform_directory_aliases(path, TRUSTED_PATH_IDS, inspect_path)?;
     let path = normalized_path.as_deref().unwrap_or(path);
 
     #[cfg(not(target_vendor = "apple"))]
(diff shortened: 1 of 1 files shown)


erskingardner (commented):
Reviewed the alias-resolution change with particular attention to trust boundaries, symlink replacement races, relative/absolute target handling, multiple prefix components, Android UID/GID assumptions, and the handoff to the descriptor-relative `O_NOFOLLOW` walk. I did not find any concrete correctness or security issues to raise. The implementation keeps the leaf unresolved, verifies both the alias parent chain and target chain, limits Android's platform UID trust to Android builds, and preserves strict handling in the subsequent walk. CI is also passing.

erskingardner (approved):

dannym-arx (approved):

PR #1270, opened 2026-08-05
Add distributed campaign runner core

erskingardner:
## Summary

- add the operating-system-level distributed campaign runner for validated container and external-VM manifests
- add deterministic command planning, mixed participant images, retained-relay execution, and process/container lifecycle control
- make network, disk, database, restart, and host-fault application transactional with exact rollback receipts
- add bounded VM and infrastructure timeouts, restrictive artifacts, real-container smoke coverage, and operator documentation

## Scope boundary

This is the first replacement slice for #1250. It intentionally contains:

- no production `cgka-engine` changes;
- no convergence trait or persistence-format changes;
- no #1236 fork-selection assumption or expected outcome;
- no application-witness, candidate-branch, retained-anchor, or fork-recovery behavior changes.

The container smoke scenario checks the existing process adapter and exact-state observation without depending on the unresolved #1236 route-equivalence topology. Operational lanes and the failure-corpus lifecycle will follow as separate stacked PRs.

## Verification

- `cargo test -p cgka-conformance-simulator --locked --test process_orchestrator`
- `cargo clippy -p convergence-campaign-runner -p cgka-conformance-simulator --all-targets --locked -- -D warnings`

Real OCI runtime coverage remains explicitly ignored unless a prebuilt Linux campaign image and Docker/Podman daemon are available.

## Summary by CodeRabbit

* **New Features**
  * Added distributed convergence campaign support for container and virtual-machine environments.
  * Added manifest validation, execution planning, fault scheduling, rollback, cleanup, and run receipts.
  * Added command-line workflows for validating, planning, diagnosing, and running campaigns.
  * Added standalone conformance node, relay, and campaign container tools.
* **Bug Fixes**
  * Improved protocol validation, compatibility handling, convergence processing, and startup preflight checks.
* **Documentation**
  * Added campaign architecture, usage guidance, safety requirements, and repository navigation updates.


erskingardner (commented):
Reviewed the full diff (21 files, +3466/−33), the scope-boundary notes, and traced the two `client.rs` edits back through the commit history.

## Blocking

**1. `--relay-proxy-listen` is rejected by the node binary — every container campaign dies at launch.**

`plan.rs:238-245` appends four args to every participant's node command:

```
--relay-proxy <relay>:8080 --relay-proxy-listen 127.0.0.1:18080
```

But `crates/cgka-conformance-simulator/src/bin/cgka-conformance-node.rs:18-25` parses exactly two:

```rust
let upstream = args.next().ok_or("missing relay proxy upstream")?;
if args.next().is_some() {
    return Err("unexpected node argument".into());
}
```

The third arg makes the node exit non-zero before `run_node_stdio` ever starts. The listen address is already hardcoded (`TcpListener::bind("127.0.0.1:18080")`), so the flag is both unsupported and redundant. `tests/distributed_runner.rs:300-315` asserts the flag *is* present, so the test suite pins the broken argv; the only test that would catch it is `#[ignore]`d. Either drop the flag from the plan or teach the node to accept it.

**2. `#[cfg(debug_assertions)]` on the legacy compatibility profile silently changes harness semantics in release builds** — `crates/cgka-conformance-simulator/src/client.rs:419`:

```rust
#[cfg(debug_assertions)]
if protocol_profile == ProtocolProfile::Legacy {
    builder = builder.legacy_compatibility_profile();
}
```

In any release build — including the release-built `cgka-conformance-node` baked into `Dockerfile.convergence-campaign`, and `cargo test --release` — `ProtocolProfile::Legacy` quietly becomes non-legacy. Roughly fifty call sites use `ProtocolProfile::Legacy` across `retained_relay.rs`, `family.rs`, and `reference_subject.rs`. A conformance harness that means different things in debug and release is a fidelity problem in its own right, and this lands in a PR whose scope boundary says it contains no convergence changes. It reads like a workaround for a release-mode failure; it needs an explanation or removal.

**3. `CrashParticipantHost` cannot work under containers.** `runner.rs:318-368` lowers it to `CrashProcess` + `RestartProcess`, and `process_orchestrator.rs:1377` implements the crash as `child.kill()`. Under the container backend the child is the `docker run` *client*, not the container. SIGKILL to the CLI detaches it and leaves the container (and its `--rm`) alive, so the subsequent `RestartProcess` re-runs `docker run --name {run_token}-{participant}` into a name conflict. The same applies to pause/resume (`resume_pid(node.child.id())` at `process_orchestrator.rs:1080`) — SIGSTOP stops the CLI while the node inside keeps running. No test covers this path; the ignored smoke test only exercises shape/reset.

## High

**4. Rollbacks are keyed by barrier, not by fault, so compensation can apply a fault that was never applied.** `plan.rs:150-170` concatenates all faults at a barrier into one `fault_commands` list and one `fault_rollbacks` list; `runner.rs:398-418` runs the entire rollback list when *any* command in the barrier fails. Since `NetworkHeal`'s rollback is `network_partition_commands`, a barrier like `[FillDisk(alice), NetworkHeal(bob, relay)]` where the `FillDisk` fails will execute a *partition* of bob↔relay that never existed. Rollback needs to be per-fault and gated on which faults actually applied.

**5. Timeouts leave no receipt, and all command output is discarded.** `runner.rs:601` returns `receipt: None` on timeout, so the most likely infra failure mode is absent from `command_receipts` — that undercuts the "exact rollback receipts" claim. Combined with `stdout(Stdio::null()).stderr(Stdio::null())` at `runner.rs:595-597`, a failing `docker` invocation leaves no diagnostic anywhere: no exit-code context, no stderr, nothing in the receipt. Capturing stderr into the private artifact directory would cost little and preserves the privacy posture.

**6. The `advance_convergence_inputs_until_settled(&gid, 0)` prepass** (`client.rs:1384-1398`, added in `89ce451 "Address distributed campaign review findings"`) changes the shared harness tick for every non-virtual-time subject. The comment explains the intent, but there is no test pinning the new behavior, and the PR states it contains no convergence changes. Worth splitting out or covering directly.

## Medium

- **`FillDisk` isn't participant-scoped.** `plan.rs` fallocates into `/campaign/participants/<token>/.disk-pressure`, but `/campaign` is a bind mount of the single host run root shared by every participant *and* the runner's own report/JSONL writes (`--mount type=bind,src={host_run_root},dst=/campaign`). One participant's disk-pressure fault degrades all peers and can fail the runner's own artifact writes — or fill the operator's filesystem.
- **`validate_identifier` (`manifest.rs:344`) permits a leading `-`**, so `namespace`, `campaign_id`, and participant ids can render as flags to `docker`. Separately, `participant.container_image` overrides are never validated at all (only `default_participant_image` is checked non-empty). Rejecting a leading `-` and validating overrides closes both.
- **No relay-readiness gate.** Setup starts the relay with `--detach` and immediately launches nodes; nothing waits for the listener. Likely flake source once the smoke test runs for real.
- **`ProcessOrchestrator::run_token()` panics** (`.expect(...)`) on the same condition the sibling code at `process_orchestrator.rs:674` returns `invalid_run_root` for.
- **Coverage swap needs a note.** `tests/process_orchestrator.rs` replaced `ProbeBidirectionalDecryptability` with `OmitMessage` in `unsupported_process_capabilities_fail_before_launch`. If the preflight relaxation now makes the probe *supported* by the process adapter, that is a behavior change with no test either way.
- **Repo convention:** the new crate ships `README.md` only. Every other crate in the `AGENTS.md` table points at an `AGENTS.md` with a sibling `CLAUDE.md` symlink, which is a stated invariant.
- **`Doctor`** runs a blocking `std::process::Command` inside `#[tokio::main]` with no timeout (`main.rs:71-77`).

## Minor

- `PlannedCommandV1::idempotent` accepts exit codes `{0,1,2}` blanket-wide, masking genuine `docker rm` / `network rm` failures.
- `ipset add <set> <container-name>` into a `hash:ip` set relies on Docker DNS resolving the peer name from inside the shared netns — plausible but brittle and unexercised (the ignored test only covers shape/reset).
- The default relay label `"relay:shared"` is derived independently in `process_orchestrator.rs` and `runner.rs:273-282`; those will drift.
- `network_exec` uses `default_participant_image` as the injector even when a participant has an image override, which slightly contradicts the "dedicated injector" comment.

## What's good

The transactional shape is right: validate-then-mutate, verify-after-apply (`iptables -C` checks), compensation on the error path, receipts written on both success and failure, cleanup attempted unconditionally, and cleanup failures surfaced in `completed`. Argv arrays throughout with no shell. The VM-escalation gate — rejecting a VM backend unless a fault genuinely needs it, and requiring declared capabilities — is a good way to keep the escalation reviewable. Bounded timeouts on both infra commands and the VM driver, and `fs_private` on every artifact.

Note: I did not run the build or tests. The two blocking findings a compile would not catch (1 and 3) are argv and process-semantics issues that pass `cargo test` today precisely because the real-container test is `#[ignore]`d.

---

erskingardner (commented):
Re-reviewed at `163587b`. Good response — the argv bug, the barrier-keyed rollback, and several hardening items are properly fixed. This time I built and ran the tests, and **two of them fail at this head.**

## Test suite is red

```
$ cargo test -p convergence-campaign-runner --locked
test result: FAILED. 12 passed; 1 failed

---- mixed_builds_select_an_exact_image_per_participant ----
tests/distributed_runner.rs:395: called `Result::unwrap()` on an `Err` value:
RunnerError { code: "root_participant_unsupported",
              message: "container participants must run as a non-root host identity" }
```

`participant_container_user()` (`plan.rs:822-832`) rejects a root euid, so `container_node_launch` now fails for *any* test process running as root. This makes the unit test environment-dependent — it passes on a workstation and fails on a root CI runner or inside a container. The plan construction itself has no reason to depend on the ambient uid; consider threading the participant identity in as a parameter (defaulted from the process identity at the `run` boundary) so planning stays pure and testable.

```
$ cargo test -p cgka-conformance-simulator --lib process_orchestrator
test result: FAILED. 3 passed; 1 failed

---- observation_schema_is_rejected_before_typed_observation_decoding ----
process_orchestrator.rs:1432: assertion `left == right` failed
  left: "process_environment"
 right: "node_observation_schema_mismatch"
```

This one is deterministic. The fixture at `process_orchestrator.rs:1427` embeds `"schema_version":"1"`, which *equals* `NODE_OBSERVATION_SCHEMA_VERSION` (`node_protocol.rs:30`), so the mismatch branch is never taken; decoding falls through to typed deserialization, which fails on the stub observation and returns `process_environment`. The fixture needs a genuinely unsupported version (e.g. `"0"`) for the test to exercise the guard it names.

Worth noting for CI: the previous head `a078dd7` also had a deterministically failing unit test — `process_preflight_supports_observed_quiescence_without_claiming_virtual_time` asserted `!descriptor.supports(VirtualTime)` while `process_subject_descriptor()` still listed `SubjectCapability::VirtualTime` in `capabilities`. `163587b` fixes it by removing that capability, which is the right call and a genuinely good catch: the process adapter had been advertising virtual-time support while implementing `AdvanceTime` as a real `tokio::time::sleep`. But the PR carried a red test across two heads with no test status reported on the commit, so whatever lane runs `just ci` does not appear to be gating here.

## Fixed — confirmed

- **`--relay-proxy-listen`** now parsed as an optional flag with a stable default, plus three unit tests covering override, default, and incomplete-override. Resolved.
- **Per-fault rollback.** `PlannedFaultV1 { manifest_index, commands, rollbacks }` replaces the barrier-keyed lists, and `rollback_faults` compensates only started faults in reverse order. `shared_barrier_rolls_back_only_started_faults_in_reverse_order` pins exactly the case I raised (`must_not_run` never runs). Cleanly done.
- **`run_token()`** returns `Result` instead of panicking, and the duplicated inline lookup now calls it.
- **`Doctor`** uses `tokio::process::Command` with `INFRASTRUCTURE_COMMAND_TIMEOUT` and `kill_on_drop`.
- **`participant.container_image` overrides** validated non-empty; images trimmed; empty `output_dir` rejected; case-insensitive YAML extension.
- **External relay map** must now match the topology exactly, with a test.
- **`duplicate_scenario_barrier`** — good fix for the `remove(name)` single-fire semantics I had assumed intentional. Worth a line in the operator doc, since it newly rejects canonical scenarios that reuse a barrier name.
- **Digest-pinned images** behind an explicit `allow_mutable_image_references` opt-out, pinned base digests, Debian snapshot, recorded package set, non-root image user, tightened `.dockerignore`. Nice hardening beyond what I raised.

## New issues introduced by this commit

**1. The VM escalation gate is now nearly a no-op, and the doc contradicts the code.** `required_vm_capability()` (`manifest.rs:662-678`) maps every network fault to `KernelNetworkIsolation`, every disk fault to `FilesystemFaults`, and `CrashParticipantHost` to `HostIsolation`. Since `unnecessary_vm_backend` fires only when `required.is_empty()`, that check now triggers only for a manifest with no faults or `RestartRelay` alone. A plain `NetworkPartition` campaign — squarely container-representable, and implemented as such — now validates happily as a VM campaign. Meanwhile `distributed-convergence-campaigns.md` still states a VM manifest "is rejected unless the campaign requests behavior that the container backend cannot faithfully provide, currently block-device latency." Either the gate should keep a container-representable set (so only `SlowBlockDevice` escalates) or the doc needs to stop claiming a narrowness the code no longer enforces.

**2. `--user` will break rootless Podman.** `participant_container_user()` passes the host euid/egid straight through, which is right for Docker but wrong for rootless Podman: the user namespace remaps that uid, so the container process won't match the bind-mounted run root's ownership and will fail to write it. `OciRuntimeV1::Podman` is a first-class option in the manifest. Rootless Podman needs `--userns=keep-id` (or `--user 0:0` with the default id mapping).

**3. The campaign image no longer builds with the repo's pinned toolchain.** Tightening the build context to `Cargo.toml`, `Cargo.lock`, `.cargo`, `crates`, `integrations` drops `rust-toolchain.toml`, which pins `1.95.0`. The builder base is `rust:1.91-bookworm`, so the image now compiles the node and relay on 1.91 while CI uses 1.95 — a reproducibility regression introduced by the commit that otherwise pins everything else. Adding `COPY rust-toolchain.toml ./` (and bumping the base) closes it.

**4. `--advertise-ip` is currently dead weight.** The relay is started with `--detach` and its stdout is discarded, and nothing reads the `ready ws://…` line, so the new advertise plumbing has no consumer. It becomes useful only once there is a readiness gate — which is still missing (see below).

## Still open from the previous review

- **`#[cfg(debug_assertions)]` on `legacy_compatibility_profile`** (`client.rs:419`) — unchanged and still unexplained. This remains my main concern: the conformance harness means different things in debug and release, including in the release-built node inside the campaign image.
- **`advance_convergence_inputs_until_settled(&gid, 0)` prepass** (`client.rs:1384-1398`) — unchanged, still untested, still outside the stated scope.
- **`CrashParticipantHost` under containers** — unchanged. `crash_participant` sends `libc::kill(-pid, SIGKILL)` to the `docker run` client's process group (`process_orchestrator.rs:1362-1371`); the container survives, `--rm` never fires, and the follow-up `RestartProcess` re-runs `docker run --name {run_token}-{participant}` into a name conflict. This commit's own capability model now says host-crash requires `HostIsolation`, yet the container branch of `validate()` still accepts the fault — rejecting it there (as `SlowBlockDevice` is) would at least make the limitation honest.
- **Timeouts produce no receipt** (`runner.rs:624`) and **all command stdio is `/dev/null`** (`runner.rs:616-618`). Unchanged.
- **No relay-readiness gate** between the detached relay and node launch.
- **`FillDisk` fills the shared host run root**, affecting all participants and the runner's own artifact writes.
- **`validate_identifier` still permits a leading `-`** for `namespace`, `campaign_id`, `build_id`, and barrier names. Digest pinning incidentally closes the image vector, but not these.
- **New crate still has no `AGENTS.md` + `CLAUDE.md` symlink.**

Happy to re-check once the two failing tests are green.

---

crates/cgka-conformance-simulator/src/process_orchestrator.rs
```diff
@@ -1192,4 +1404,60 @@ mod tests {
...
+            .unwrap(),
+            "name=run-1234-participant-abcd:/tmp/host-run:/campaign"
+        );
+        let error = render_node_arg("{shell}", "p", "r", host, child).unwrap_err();
+        assert_eq!(error.code, "unknown_node_command_placeholder");
+    }
+    #[test]
+    fn observation_schema_is_rejected_before_typed_observation_decoding() {
+        let line = format!(
+            r#"{{"protocol":"{NODE_PROTOCOL_VERSION}","request_id":"request-1","participant":"alice","body":{{"type":"observation","action_id":"observe","observation":{{"schema_version":"1"}}}}}}"#
```
erskingardner: **Blocking — this test is what is failing the `Simulator and vectors` CI lane.**

The fixture sets `"schema_version":"1"`, but `NODE_OBSERVATION_SCHEMA_VERSION` in `node_protocol.rs` is also `"1"`, so the mismatch guard in `decode_node_response` never fires. Typed decoding of the minimal observation then fails and surfaces as `process_environment`, not `node_observation_schema_mismatch`:

```
left: "process_environment"
right: "node_observation_schema_mismatch"
```

Use a schema version that is not the current one (e.g. `"0"` or `"2"`) so the guard actually triggers.

Also note the verification gap: the PR body ran `cargo test -p cgka-conformance-simulator --locked --test process_orchestrator`, which only runs the integration target — not the lib unit tests where this regression lives. `cargo test -p cgka-conformance-simulator` (or `just test`) would have caught it locally. Because nextest fail-fast cancelled the remaining 277 tests in that lane, the rest of the simulator suite is also currently unverified at this head.

crates/convergence-campaign-runner/src/plan.rs
```diff
@@ -0,0 +1,859 @@
...
+        "/campaign/participants/{}",
+        process_participant_token(participant)
+    )
+}
+
+fn participant_pressure_path(participant: &str) -> String {
+    format!("{}/.disk-pressure", participant_state_path(participant))
+}
+
+fn peer_container(peer: &FaultPeerV1, relay: &str) -> String {
```
erskingardner, line 860: **Participant↔participant partitions appear to be vacuous in this topology.**

Every node dials only its in-container loopback proxy (`ws://127.0.0.1:18080`), which forwards to `{namespace}-relay:8080`. All inter-participant traffic is relay-mediated; no node ever opens a connection to another participant's container IP. So when `peer` is `FaultPeerV1::Participant(bob)`, the ipset/iptables REJECT rules installed in alice's netns match a destination nobody dials — the partition changes zero traffic. The `verify_*` commands only prove the rules exist, not that any flow actually stopped, so a campaign can produce evidence of an alice↔bob partition that never happened.

Partition vs. `FaultPeerV1::Relay` works (the proxy→relay hop is real). For participant peers, either enforce the partition at the relay (e.g. drop forwarding between the two) or reject/document the unsupported peer kind in `validate` so the evidence stays honest.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -416,6 +416,7 @@ fn build_harness_engine(
         .supported_app_components(harness_supported_app_components())
         .peeler(Box::new(peeler))
         .recorder(Box::new(CapturingRecorder::new(audit_capture.clone())));
+    #[cfg(debug_assertions)]
```
erskingardner, line 428: Question: the campaign image builds the node with `cargo build --release` (`Dockerfile.convergence-campaign`), where this `#[cfg(debug_assertions)]` drops the call — so a release-built node silently runs the Current profile even when the scenario requests `ProtocolProfile::Legacy`, while debug-built local nodes honor it. That is a silent behavior divergence between process and container subjects on the same scenario.

If container campaigns are intended never to run legacy profiles, a short comment here saying so (or a loud error when a release node is asked for Legacy) would keep evidence from quietly diverging.

crates/convergence-campaign-runner/src/manifest.rs
```diff
@@ -0,0 +1,419 @@
...
+                        "fault references a participant outside the manifest",
+                    ));
+                }
+            }
+            fault.action.validate_values()?;
+        }
+        match &self.backend {
+            DistributedBackendV1::Container(container) => {
+                validate_identifier("container_namespace", &container.namespace)?;
+                if container.default_participant_image.trim().is_empty()
```
erskingardner, line 267: Minor: `validate_identifier` allows `:` (needed for relay labels like `relay:primary`), but `namespace` flows into `docker network create {namespace}-network` and container name `{namespace}-relay`, where `:` is invalid and names must start with an alphanumeric. A manifest like `namespace: "a:b"` passes `validate` and then fails at runtime with an opaque docker error. Consider a stricter charset for the namespace (DNS-label-ish) so this fails at validation time.

crates/convergence-campaign-runner/src/runner.rs
```diff
@@ -0,0 +1,1076 @@
...
+                .or_default()
+                .push(process);
+        }
+    }
+    let mut lowered = scenario.clone();
+    lowered.steps = Vec::with_capacity(scenario.steps.len() + crashes.len() * 2);
+    for step in &scenario.steps {
+        lowered.steps.push(step.clone());
+        if let cgka_conformance_simulator::ScenarioStep::Barrier { name } = step
+            && let Some(processes) = crashes.get(name.as_str())
```
erskingardner, line 438: The manifest type documents "Faults are applied before the named canonical Scenario IR barrier", but `CrashParticipantHost` is lowered to `CrashProcess`+`RestartProcess` *after* the barrier step. Owning the crash in the orchestrator (so the JSONL stdio handles get replaced) makes sense, but then the documented ordering is wrong for this fault kind — and hook-applied faults at the same barrier always execute before the crash regardless of manifest order. Worth a sentence on `ScheduledFaultV1` (or here) pinning the actual semantics: network/disk/db faults run before the barrier; host crashes run immediately after it.

AGENTS.md
```diff
@@ -52,6 +52,7 @@ The canonical protocol specification lives in
 | App runtime UniFFI bindings | `crates/marmot-uniffi/AGENTS.md` |
 | CLI / daemon / TUI surface | `crates/cli/AGENTS.md` |
 | Multi-client harness / vectors | `crates/cgka-conformance-simulator/AGENTS.md` |
+| Container/VM convergence campaigns | `crates/convergence-campaign-runner/README.md` |
```
erskingardner: Convention nit: every other row in this table points at an `AGENTS.md`, and the repo invariant says new `AGENTS.md` files get a sibling `CLAUDE.md` symlink. The new crate ships only a `README.md`. Either add `crates/convergence-campaign-runner/AGENTS.md` (+ `CLAUDE.md` symlink) and point the row there, or keep pointing at the README deliberately — but the table should stay consistent.

erskingardner (commented):
**Requesting changes** (posting as a comment since the CLI user is also the PR author).

Reviewed at head `163587b`. The slice boundary is well drawn (no engine changes, manifest/plan/receipt layering is clean, argv-only command construction, exact rollback receipts, and the validation surface is thorough), and the stale CodeRabbit items (`--relay-proxy-listen`, relay advertise, external relay map) are indeed resolved at this head — I verified each.

One blocking issue:

- **CI is red.** The new unit test `observation_schema_is_rejected_before_typed_observation_decoding` uses the *current* observation schema version as its "unsupported" fixture, so it fails deterministically (`process_environment` vs `node_observation_schema_mismatch`) and takes down the `Simulator and vectors` lane. Details inline. Because nextest fail-fast cancelled the remaining 277 tests, the rest of the simulator suite is unverified at this head — and the PR body's verification command (`--test process_orchestrator`) never runs the lib unit tests, which is how it slipped through.

One design concern worth resolving before this becomes an evidence foundation:

- **Participant↔participant network partitions are vacuous** — all node traffic is relay-mediated through the in-container loopback proxy, so iptables rules matching a participant peer block nothing. The campaign would certify a partition that never happened. See inline comment on `peer_container`.

Non-blocking nits inline: release/debug `Legacy` profile divergence in container nodes, namespace charset vs docker naming, crash-after-barrier vs documented fault semantics, and the AGENTS.md/CLAUDE.md convention for the new crate.

Datawav:

## Final audit

Need/design: The PR introduces a container-first distributed campaign runner that reuses the existing scenario IR, process protocol, and observation oracle. It adds manifest validation, deterministic argv-only plans, retained relays, scoped fault injectors, mixed participant images, VM-driver escalation, and private execution receipts. This avoids creating a second scenario language or embedding Docker/VM concerns in the simulator.

Alternatives: Keep distributed orchestration in an external harness, or initially support only containers and defer the VM-driver contract. The proposed adapter boundary is reasonable, but its lifecycle and rollback guarantees need the fixes below.

### Findings

1. **High — Partial startup can leak participant containers**  
   `crates/convergence-campaign-runner/src/runner.rs:369-379`

   `ProcessOrchestrator::launch_with` can start participants incrementally and then fail while initializing a later participant. The caller converts every launch failure into `ContainerScenarioFailure { run_token: None }`. Consequently, `run_container` skips `cleanup_participants`, while relay/network cleanup may also fail because an already-started participant still occupies the network.

   Preserve the run token before participant startup, return it on partial failure, or make `launch_with` shut down all successfully started nodes before returning `Err`. Add coverage where the second participant fails initialization and verify the first container is removed.

2. **Medium — Rollback plans omit network-shape and disk-pressure compensation**  
   `crates/convergence-campaign-runner/src/plan.rs:434-453`

   `container_fault_rollback_commands` supplies compensation only for partition/heal. A successfully applied `NetworkShape` or `FillDisk` remains active when a later fault at the same barrier fails, despite `before_barrier` attempting rollback of all started faults. Container teardown eventually limits the network effect, but it is not the transactional restoration or exact rollback evidence claimed by the PR; disk pressure also resides in the bind-mounted run root.

   Add absence-tolerant rollback commands: `tc qdisc del dev eth0 root` for shaping and removal of the participant’s `.disk-pressure` file for disk filling. Include these receipts in reverse-order rollback tests.

erskingardner (commented):
Final pass at `1197f5e`. Strong round — most of what I raised is now genuinely resolved, and several fixes went further than the finding. Verification I ran locally:

```
cargo fmt --all -- --check                                  clean
cargo clippy -p convergence-campaign-runner \
  -p cgka-conformance-simulator --all-targets --locked       clean
cargo test -p cgka-conformance-simulator --lib process_orchestrator   4 passed
cargo test -p convergence-campaign-runner --locked           16 passed, 1 FAILED
```

## One test still fails

```
---- mixed_builds_select_an_exact_image_per_participant ----
tests/distributed_runner.rs:479: called `Result::unwrap()` on an `Err` value:
RunnerError { code: "root_participant_unsupported",
              message: "container participants must run as a non-root host identity" }
```

Unchanged from my last review. `container_node_launch` reads the ambient euid through `participant_container_user()`, so plan construction fails for any test process running as root — a root CI runner or a container. It is the only red test in the crate; everything else is green. Threading the participant identity in as a parameter (resolved from the process identity at the `run` boundary, as `resource_token` now is) would make planning pure and drop the environment dependence. The `resource_token` refactor in this very commit is the pattern to copy.

## Resolved — verified

- **Legacy profile in release builds.** The debug-only switch is now backed by a release-build `assert_ne!` with a comment explaining the intent (`client.rs:408-415`). Turning a silent semantic difference into a loud failure is the right resolution; my main objection is addressed.
- **`observation_schema_is_rejected_before_typed_observation_decoding`** — fixture now uses `"schema_version":"0"` and the test exercises the guard it names. Passes.
- **Timeouts now produce a receipt** (`runner.rs:756-762`, `exit_code: None, accepted: false`), closing the evidence gap in the "exact receipts" claim.
- **VM lifecycle contract.** `driver_contract_version`, `cleanup_args` (required to contain `{manifest}`), and an independent `cleanup_timeout_seconds`; cleanup runs after success, failure, *and* timeout, its receipt is recorded, and the run fails closed on cleanup failure. This is more than I asked for and the doc explains the ownership boundary well.
- **Concurrent-campaign safety.** `ContainerResourceLease` gives each run a token, and network/relay names are `{namespace}-{resource_token}-…`, so two campaigns sharing a namespace can no longer tear down each other's resources. Covered by `concurrent_container_leases_render_disjoint_resource_names`.
- **Participant-to-participant partitions are now rejected** on the container backend rather than silently recorded as vacuous faults, with the reasoning in the doc. This also retires my `ipset`-DNS concern, since the only remaining peer is the runner-owned relay.
- **Relay proxy dial safety.** No configurable upstream, a fixed runner-owned alias, loopback-only listener, resolution restricted to private container addresses, and privacy-safe `tracing` fields instead of `eprintln!`. Matches the dial-discipline invariant, and `allow_cleartext_isolated_relay` records the exception in the normalized manifest.
- **`validate_container_namespace`** requires alphanumeric first and last characters, closing the leading-`-` argv vector where it actually reached `docker`. The remaining `validate_identifier` users (`campaign_id` in a label value, `build_id` in evidence, barrier names not in argv) no longer present a path.
- **Partition/heal ordering** — duplicate partition and unmatched heal now rejected.
- **Layering.** Node and relay binaries moved into the campaign-runner crate; the simulator's node is stdio-only again. Good separation, and the Dockerfile follows.
- **`AGENTS.md` + `CLAUDE.md` symlink** added for the new crate.
- **`stop_database_contention`** narrowed from `{0,1,2}` to `{0,1}`.

## Still open

None of these block on their own, but they are unchanged across three rounds and worth tracking:

1. **The VM escalation gate remains near-vacuous.** `required_vm_capability()` (`manifest.rs`) still maps every network fault to `KernelNetworkIsolation` and every disk fault to `FilesystemFaults`, so `unnecessary_vm_backend` fires only for an empty or `RestartRelay`-only fault list. The doc at line 38 still says a VM manifest is rejected "unless the campaign requests behavior that the container backend cannot faithfully provide, currently block-device latency." A `NetworkShape` campaign still validates as a VM campaign. Code and doc need to agree.
2. **The campaign image still builds on the wrong toolchain.** `rust-toolchain.toml` pins `1.95.0` and is still absent from the build context (`COPY Cargo.toml Cargo.lock ./`, `.cargo`, `crates`, `integrations`), while the builder base is `rust:1.91-bookworm`.
3. **Rootless Podman.** `--user <host uid:gid>` with no `--userns=keep-id`; `OciRuntimeV1::Podman` is still a supported manifest value.
4. **`advance_convergence_inputs_until_settled(&gid, 0)`** (`client.rs`) — unchanged, still untested, still outside the stated scope.
5. **`CrashParticipantHost` under containers.** The doc now describes the after-barrier lowering, but the mechanism is unchanged: `libc::kill(-pid, SIGKILL)` targets the `docker run` client's process group, the container survives, and `RestartProcess` re-runs `docker run --name …` into a name conflict. Rejecting the fault on the container backend — as `SlowBlockDevice` is — would at least be honest until the runner can stop the container itself.
6. **Command stdio is still `/dev/null`** (`runner.rs:751-753`), so a failing `docker` invocation leaves no stderr anywhere.
7. **`FillDisk`** still fills the shared host run root, affecting all participants and the runner's own artifact writes.
8. **Relay readiness** is still not gated. The lazy per-connection resolve with a 10s timeout mitigates it, but a failed resolve silently drops that connection and relies on client retry. Relatedly, `--advertise-ip` still has no consumer — it now only feeds a `loopback_advertisement` boolean in one log line.

Minor: `tempfile` moved from `[dev-dependencies]` to `[dependencies]`, so it now ships in the runner's production graph purely to mint a unique token in `ContainerResourceLease`. A counter or random suffix would avoid the dependency.

Fix the root-uid test and this is in good shape from my side.

---

PR #1271, opened 2026-08-05
Add distributed campaign execution lanes

erskingardner:
## Summary

- make the four reviewed JSON lane manifests the single compiled policy source
- make budget evaluation fail closed for zero, undersized, or inconsistent observations
- require nonempty evidence artifacts and verify their bytes against SHA-256 at ingestion and before writing
- run scheduled lanes through their aggregate recipes, remove nonexistent test targets, and make zero-test container filters fail
- inherit the mandatory PR formal proof instead of rerunning the same state-independent proof on schedules
- require release-hardening manifests to contain distinct build IDs and distinct effective container images

## Scope and current limit

This PR contains operational lane policy, evidence-budget evaluation, workflow wiring, and documentation. It contains no production convergence-engine changes and makes no claim that #1236 is resolved.

The manifests and evaluator are reviewed, machine-checkable foundations. Scheduled workflows do not yet produce aggregate resource observations or invoke `check-budget`, so the tracking plan keeps runtime budget enforcement and the four completed-lane claims open. Incident-corpus execution also remains false in this PR and is added by the stacked failure-corpus PR.

## Verification

- `cargo clippy -p convergence-campaign-runner --all-targets --locked -- -D warnings`
- `just fast-ci`
- `actionlint .github/workflows/simulator-nightly.yml .github/workflows/convergence-hardening.yml`
- verified `cargo nextest ... --no-tests fail` exits nonzero for a missing selected-container test

#1272 has been restacked onto this head and keeps the failure-corpus policy delta isolated.

## Summary by CodeRabbit

- **New Features**
  - Added configurable convergence testing lanes for pull requests, nightly checks, weekly campaigns, and release hardening.
  - Added checks for resource budgets, mixed-build coverage, campaign results, and evidence artifacts.
  - Added commands to inspect policies, evaluate budgets, and validate evidence bundles.

- **CI Improvements**
  - Added scheduled and manual hardening workflows.
  - Extended artifact retention and added real-container convergence testing.

- **Documentation**
  - Updated reliability and distributed-convergence guidance with lane coverage, budgets, and evidence requirements.

### DIFF
--- a/.github/workflows/simulator-nightly.yml
+++ b/.github/workflows/simulator-nightly.yml
@@ -50,44 +50,11 @@ jobs:
       - name: Cache Cargo artifacts
         uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
 
-      - name: Install Tamarin
-        env:
-          TAMARIN_VERSION: 1.12.0
-        run: |
-          set -euo pipefail
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends graphviz maude
-          curl -L --fail --show-error --silent \
-            "https://github.com/tamarin-prover/tamarin-prover/releases/download/${TAMARIN_VERSION}/tamarin-prover-${TAMARIN_VERSION}-linux64-ubuntu.tar.gz" \
-            -o /tmp/tamarin-prover.tar.gz
-          tar -xzf /tmp/tamarin-prover.tar.gz -C /tmp
-          sudo install -m 0755 /tmp/tamarin-prover /usr/local/bin/tamarin-prover
-
-      - name: Run strict symbolic proof gate
-        run: just tamarin
-
-      - name: Run complete simulator suite
-        # Adversarial campaigns and independent convergence checks run below
-        # with their required feature sets; exclude only those duplicates here.
-        run: just simulator-full
-
-      - name: Validate convergence execution-lane policy
-        run: just convergence-lane-policy
+      - name: Run nightly convergence lane
+        run: just convergence-nightly-lane
 
       - name: Run simulator doctests
         run: cargo test -p cgka-conformance-simulator --doc --locked
-
-      - name: Run adversarial reliability campaigns
-        run: just adversarial-reliability-ci
-
-      - name: Verify convergence with independent models
-        run: just convergence-verification-ci
-
-      - name: Run process-boundary convergence campaigns
-        run: cargo nextest run -p cgka-conformance-simulator --test process_orchestrator --test route_equivalence --locked
-
-      - name: Validate distributed campaign runner
-        run: cargo nextest run -p convergence-campaign-runner --locked
 
       - name: Build current distributed campaign image
         run: docker build -f Dockerfile.convergence-campaign -t marmot-conformance:local .
@@ -95,7 +62,10 @@ jobs:
       - name: Run selected real-container convergence campaign
         env:
           CGKA_CONVERGENCE_IMAGE: marmot-conformance:local
-        run: cargo test -p convergence-campaign-runner --test container_runtime --locked -- --ignored real_container_nodes_recover_the_cross_route_deeper_branch
+        run: |
+          cargo nextest run -p convergence-campaign-runner --test container_runtime --locked \
+            --run-ignored ignored-only --no-tests fail \
+            -E 'test(=real_container_nodes_survive_network_shaping_and_reach_exact_state)'
 
       - name: Run encrypted file-backed vector report
         run: |
--- a/.github/workflows/simulator-nightly.yml
+++ b/.github/workflows/simulator-nightly.yml
@@ -50,10 +50,29 @@ jobs:
       - name: Cache Cargo artifacts
         uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
 
+      - name: Install Tamarin
+        env:
+          TAMARIN_VERSION: 1.12.0
+        run: |
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends graphviz maude
+          curl -L --fail --show-error --silent \
+            "https://github.com/tamarin-prover/tamarin-prover/releases/download/${TAMARIN_VERSION}/tamarin-prover-${TAMARIN_VERSION}-linux64-ubuntu.tar.gz" \
+            -o /tmp/tamarin-prover.tar.gz
+          tar -xzf /tmp/tamarin-prover.tar.gz -C /tmp
+          sudo install -m 0755 /tmp/tamarin-prover /usr/local/bin/tamarin-prover
+
+      - name: Run strict symbolic proof gate
+        run: just tamarin
+
       - name: Run complete simulator suite
         # Adversarial campaigns and independent convergence checks run below
         # with their required feature sets; exclude only those duplicates here.
         run: just simulator-full
+
+      - name: Validate convergence execution-lane policy
+        run: just convergence-lane-policy
 
       - name: Run simulator doctests
         run: cargo test -p cgka-conformance-simulator --doc --locked
@@ -63,6 +82,20 @@ jobs:
 
       - name: Verify convergence with independent models
         run: just convergence-verification-ci
+
+      - name: Run process-boundary convergence campaigns
+        run: cargo nextest run -p cgka-conformance-simulator --test process_orchestrator --test route_equivalence --locked
+
+      - name: Validate distributed campaign runner
+        run: cargo nextest run -p convergence-campaign-runner --locked
+
+      - name: Build current distributed campaign image
+        run: docker build -f Dockerfile.convergence-campaign -t marmot-conformance:local .
+
+      - name: Run selected real-container convergence campaign
+        env:
+          CGKA_CONVERGENCE_IMAGE: marmot-conformance:local
+        run: cargo test -p convergence-campaign-runner --test container_runtime --locked -- --ignored real_container_nodes_recover_the_cross_route_deeper_branch
 
       - name: Run encrypted file-backed vector report
         run: |
@@ -81,6 +114,7 @@ jobs:
           name: cgka-conformance-simulator-nightly-reports
           path: target/cgka-conformance-simulator-nightly-reports
           if-no-files-found: warn
+          retention-days: 30
 
       - name: Report nightly failure
         if: failure()
--- a/Justfile
+++ b/Justfile
@@ -313,22 +313,20 @@ convergence-verification-ci:
 convergence-lane-policy:
     cargo nextest run -p convergence-campaign-runner --test lane_policy --locked
 
-# Capability-level lane entry points. The PR workflow runs the first lane as
-# separate parallel jobs; the scheduled workflows use the aggregate recipes.
-convergence-pr-lane: convergence-lane-policy simulator-smoke convergence-verification-ci
-
+# Capability-level entry points used by the scheduled workflows. PR checks
+# remain split into separately named steps for useful failure attribution.
 convergence-nightly-lane: convergence-lane-policy simulator-full adversarial-reliability-ci convergence-verification-ci
-    cargo nextest run -p cgka-conformance-simulator --test process_orchestrator --test route_equivalence --locked
+    cargo nextest run -p cgka-conformance-simulator --test process_orchestrator --locked
     cargo nextest run -p convergence-campaign-runner --locked
 
 convergence-weekly-lane: convergence-nightly-lane
     cargo run -p cgka-conformance-simulator --features test-policy-overrides --bin cgka-conformance-campaign --locked -- --cases 4 --case-timeout-secs 300 --out target/cgka-weekly-reliability --storage file
 
-# A release run must name the reviewed mixed-build/incident manifest rather
+# A release run must name a reviewed mixed-build manifest rather
 # than silently falling back to a current-build-only campaign.
 convergence-release-hardening-lane manifest:
     just convergence-weekly-lane
-    cargo run -p convergence-campaign-runner --bin cgka-distributed-campaign --locked -- validate "{{manifest}}"
+    cargo run -p convergence-campaign-runner --bin cgka-distributed-campaign --locked -- validate "{{manifest}}" --require-mixed-builds
     cargo run -p convergence-campaign-runner --bin cgka-distributed-campaign --locked -- run "{{manifest}}"
 
 tracing-audit:
--- a/Justfile
+++ b/Justfile
@@ -308,6 +308,29 @@ convergence-verification-ci:
     just tla-liveness
     just tla-liveness-counterexample
 
+# Validate the versioned execution-lane policy, all resource/retention/flake
+# budgets, and the release evidence-bundle completeness contract.
+convergence-lane-policy:
+    cargo nextest run -p convergence-campaign-runner --test lane_policy --locked
+
+# Capability-level lane entry points. The PR workflow runs the first lane as
+# separate parallel jobs; the scheduled workflows use the aggregate recipes.
+convergence-pr-lane: convergence-lane-policy simulator-smoke convergence-verification-ci
+
+convergence-nightly-lane: convergence-lane-policy simulator-full adversarial-reliability-ci convergence-verification-ci
+    cargo nextest run -p cgka-conformance-simulator --test process_orchestrator --test route_equivalence --locked
+    cargo nextest run -p convergence-campaign-runner --locked
+
+convergence-weekly-lane: convergence-nightly-lane
+    cargo run -p cgka-conformance-simulator --features test-policy-overrides --bin cgka-conformance-campaign --locked -- --cases 4 --case-timeout-secs 300 --out target/cgka-weekly-reliability --storage file
+
+# A release run must name the reviewed mixed-build/incident manifest rather
+# than silently falling back to a current-build-only campaign.
+convergence-release-hardening-lane manifest:
+    just convergence-weekly-lane
+    cargo run -p convergence-campaign-runner --bin cgka-distributed-campaign --locked -- validate "{{manifest}}"
+    cargo run -p convergence-campaign-runner --bin cgka-distributed-campaign --locked -- run "{{manifest}}"
+
 tracing-audit:
     cargo nextest run -p cgka-conformance-simulator --test tracing_audit
 
--- a/crates/convergence-campaign-runner/lanes/weekly-manual.v1.json
+++ b/crates/convergence-campaign-runner/lanes/weekly-manual.v1.json
@@ -4,8 +4,7 @@
   "contents": {
     "strict_formal_gate": true,
     "fixed_vectors": true,
-    "engine_reference_relay_cases": 1024,
-    "generated_seed_cases": 4096,
+    "minimum_executed_cases": 4,
     "file_backed_storage": true,
     "crash_matrix": true,
     "retained_relays": true,
@@ -14,8 +13,8 @@
     "container_campaigns": true,
     "resource_sweeps": true,
     "constant_sweeps": true,
-    "mixed_version": true,
-    "incident_corpus": true
+    "mixed_version": false,
+    "incident_corpus": false
   },
   "budgets": {
     "max_wall_clock_seconds": 21600,
--- a/crates/convergence-campaign-runner/lanes/weekly-manual.v1.json
+++ b/crates/convergence-campaign-runner/lanes/weekly-manual.v1.json
@@ -1,0 +1,30 @@
+{
+  "schema_version": "1",
+  "lane": "weekly_manual",
+  "contents": {
+    "strict_formal_gate": true,
+    "fixed_vectors": true,
+    "engine_reference_relay_cases": 1024,
+    "generated_seed_cases": 4096,
+    "file_backed_storage": true,
+    "crash_matrix": true,
+    "retained_relays": true,
+    "mutation_scope": "full_targeted",
+    "process_campaigns": true,
+    "container_campaigns": true,
+    "resource_sweeps": true,
+    "constant_sweeps": true,
+    "mixed_version": true,
+    "incident_corpus": true
+  },
+  "budgets": {
+    "max_wall_clock_seconds": 21600,
+    "max_cpu_seconds": 43200,
+    "max_peak_rss_bytes": 12884901888,
+    "max_disk_bytes": 42949672960,
+    "max_artifact_bytes": 10737418240,
+    "artifact_retention_days": 60,
+    "max_flake_retries": 2,
+    "max_flake_rate_basis_points": 25
+  }
+}
--- a/crates/convergence-campaign-runner/src/evidence.rs
+++ b/crates/convergence-campaign-runner/src/evidence.rs
@@ -42,6 +42,16 @@ pub struct EvidenceArtifactV1 {
     pub sha256: String,
 }
 
+/// Directory against which an evidence bundle's relative artifact paths are
+/// resolved. A bare filename lives in the current directory even though
+/// `Path::parent` represents that parent as an empty path.
+pub fn evidence_bundle_base_dir(path: &Path) -> &Path {
+    match path.parent() {
+        Some(parent) if !parent.as_os_str().is_empty() => parent,
+        _ => Path::new("."),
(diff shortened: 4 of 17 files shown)


Datawav:

Need/design/alternatives: Versioned execution lanes are needed to bound CI cost and make assurance claims reproducible. The proposed design—reviewed lane manifests, measured-budget evaluation, and scoped evidence bundles—is reasonable, but the workflows currently validate declarations rather than executions. Either instrument each lane to emit and check observations/coverage, or describe these files as planning metadata and narrow the assurance claims accordingly.

- **High — Declared budgets are not enforced by any workflow.** `.github/workflows/ci.yml`, `.github/workflows/simulator-nightly.yml`, and the `convergence-*-lane` recipes only run `convergence-lane-policy`, which checks that JSON equals compiled constants. None measures actual wall time, CPU, RSS, disk, artifact size, retries, or flakes, and none invokes `check-budget`. Every declared limit can therefore be exceeded while the lane remains green. Wrap lane execution with observation collection and make `check-budget` a required final step.

- **High — Weekly/manual policy claims coverage that is never executed.** `weekly-manual.v1.json` declares mixed-version execution, resource and constant sweeps, 1,024 engine/reference/relay cases, and 4,096 generated cases. `convergence-weekly-lane` instead adds one four-case simulator command to the nightly recipe, while the workflow builds only one local image. Mixed builds are consumed only by release hardening. Either implement and attest the declared weekly capabilities or set the policy fields and counts to what the workflow actually runs.

- **High — Empty or inconsistent observations can produce passing release evidence.** In `lane.rs`, division by zero when `executed_cases == 0` becomes a zero flake rate. No validation rejects `flaky_cases > executed_cases`, and the new evidence test explicitly obtains a passing release budget from `CampaignLaneObservationV1::default()`. Consequently, a campaign that executed nothing can satisfy `ConvergenceEvidenceBundleV1::validate()`. Reject zero executed cases and inconsistent counts before evaluating the budget; preferably enforce each lane’s minimum declared case count.

- **Medium — “Completeness-checked, digest-pinned” evidence need not contain or verify an artifact.** `evidence.rs` defaults `artifacts` to an empty vector and validates only the syntax of supplied digest strings; it never requires artifacts or binds a digest to file contents. The test’s all-zero digest is accepted. If this type is intended only as a portable schema, narrow the documentation and require artifact verification at ingestion. Otherwise require at least one artifact and hash each referenced file before accepting or writing the bundle.

_Advisory audit of head [`131b252c19eb`](https://github.com/marmot-protocol/mdk/commit/131b252c19eb3555da82613a428c6bdd978b5d6e); no approval, merge, push, checkout, or PR-code execution was performed._

.github/workflows/simulator-nightly.yml
```diff
@@ -64,6 +83,20 @@ jobs:
       - name: Verify convergence with independent models
         run: just convergence-verification-ci
 
+      - name: Run process-boundary convergence campaigns
+        run: cargo nextest run -p cgka-conformance-simulator --test process_orchestrator --test route_equivalence --locked
```
erskingardner: **Blocking — this will fail.** There is no `crates/cgka-conformance-simulator/tests/route_equivalence.rs` at this head SHA (the tests dir has `process_orchestrator.rs` but no `route_equivalence.rs`). `cargo nextest run --test route_equivalence` errors with an unknown test target.

This also breaks `convergence-nightly-lane` in the `Justfile`, which the weekly and release-hardening lanes both depend on — so the new `convergence-hardening.yml` workflow cannot pass either.

.github/workflows/simulator-nightly.yml
```diff
@@ -64,6 +83,20 @@ jobs:
...
+      - name: Validate distributed campaign runner
+        run: cargo nextest run -p convergence-campaign-runner --locked
+
+      - name: Build current distributed campaign image
+        run: docker build -f Dockerfile.convergence-campaign -t marmot-conformance:local .
+
+      - name: Run selected real-container convergence campaign
+        env:
+          CGKA_CONVERGENCE_IMAGE: marmot-conformance:local
+        run: cargo test -p convergence-campaign-runner --test container_runtime --locked -- --ignored real_container_nodes_recover_the_cross_route_deeper_branch
```
erskingardner: **Blocking — silently green.** `real_container_nodes_recover_the_cross_route_deeper_branch` does not exist. The only test in `container_runtime.rs` is:

```
async fn real_container_nodes_survive_network_shaping_and_reach_exact_state()
```

`cargo test -- --ignored <filter>` with a non-matching filter runs 0 tests and exits 0. So this step builds a Docker image and then asserts nothing, every night, forever. A step named "Run selected real-container convergence campaign" that can never fail is worse than not having it.

.github/workflows/simulator-nightly.yml
```diff
@@ -50,11 +50,30 @@ jobs:
...
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends graphviz maude
+          curl -L --fail --show-error --silent \
+            "https://github.com/tamarin-prover/tamarin-prover/releases/download/${TAMARIN_VERSION}/tamarin-prover-${TAMARIN_VERSION}-linux64-ubuntu.tar.gz" \
+            -o /tmp/tamarin-prover.tar.gz
+          tar -xzf /tmp/tamarin-prover.tar.gz -C /tmp
+          sudo install -m 0755 /tmp/tamarin-prover /usr/local/bin/tamarin-prover
+
+      - name: Run strict symbolic proof gate
```
erskingardner: The Tamarin proof gate already runs on every PR in the `formal` job in `ci.yml` (`make -C formal/tamarin prove`, 45-min timeout). Re-running it nightly, plus a third time in `convergence-hardening.yml`, adds real CI minutes for no new signal — the proofs don't depend on nightly state. The 10-line install block is now copy-pasted three times too.

The only justification I can see is that the lane manifests declare `strict_formal_gate: true` for all four lanes. That's the policy driving cost rather than describing it. If the lanes need to record that the formal gate covers them, record it — don't re-prove it.

Justfile
```diff
@@ -308,6 +308,29 @@ convergence-verification-ci:
...
     just tla-liveness-counterexample
 
+# Validate the versioned execution-lane policy, all resource/retention/flake
+# budgets, and the release evidence-bundle completeness contract.
+convergence-lane-policy:
+    cargo nextest run -p convergence-campaign-runner --test lane_policy --locked
+
+# Capability-level lane entry points. The PR workflow runs the first lane as
+# separate parallel jobs; the scheduled workflows use the aggregate recipes.
+convergence-pr-lane: convergence-lane-policy simulator-smoke convergence-verification-ci
```
erskingardner: `convergence-pr-lane` is dead on arrival — `ci.yml` calls `just simulator-smoke` and `just convergence-lane-policy` as separate steps and never references this recipe. Same story for `convergence-nightly-lane`, which `simulator-nightly.yml` bypasses by inlining the individual commands (it survives only as a transitive dep of the weekly lane).

Either wire the workflows to the recipes or drop the aggregates.

Justfile
```diff
@@ -308,6 +308,29 @@ convergence-verification-ci:
...
+
+convergence-nightly-lane: convergence-lane-policy simulator-full adversarial-reliability-ci convergence-verification-ci
+    cargo nextest run -p cgka-conformance-simulator --test process_orchestrator --test route_equivalence --locked
+    cargo nextest run -p convergence-campaign-runner --locked
+
+convergence-weekly-lane: convergence-nightly-lane
+    cargo run -p cgka-conformance-simulator --features test-policy-overrides --bin cgka-conformance-campaign --locked -- --cases 4 --case-timeout-secs 300 --out target/cgka-weekly-reliability --storage file
```
erskingardner, line 323: `weekly-manual.v1.json` declares `engine_reference_relay_cases: 1024`, `generated_seed_cases: 4096`, `resource_sweeps: true`, `constant_sweeps: true`, and `mixed_version: true`. What the weekly lane actually adds over nightly is `--cases 4`, and no mixed-build participants at all (mixed builds are consumed only by the release manifest).

Same problem one level up: `nightly.v1.json` declares `incident_corpus: true`, but §6.4 Failure corpus lifecycle is still entirely unchecked in the plan doc and nothing in the nightly recipe runs a corpus.

The manifests should describe what runs. Right now they describe an aspiration that a green check appears to attest.

crates/convergence-campaign-runner/src/lane.rs
```diff
@@ -0,0 +1,326 @@
...
+            observation.artifact_bytes,
+            budget.max_artifact_bytes,
+        );
+        check_limit(
+            &mut violations,
+            "flake_retries",
+            observation.flake_retries,
+            u64::from(budget.max_flake_retries),
+        );
+        let flake_rate_basis_points = observation
```
erskingardner, line 223: `checked_div(...).unwrap_or(0)` turns "executed zero cases" into "zero flake rate", i.e. a pass. Nothing else validates the observation either — `flaky_cases > executed_cases` and `flake_retries` with `executed_cases == 0` are both accepted.

Combined with `ConvergenceEvidenceBundleV1::validate()` requiring only `budget.passed`, an empty campaign produces valid release evidence. The new test demonstrates exactly this (see `lane_policy.rs:87`).

Suggest rejecting `executed_cases == 0` and `flaky_cases > executed_cases` up front, and ideally requiring `executed_cases >= contents.engine_reference_relay_cases + contents.generated_seed_cases` for the lane.

crates/convergence-campaign-runner/src/lane.rs
```diff
@@ -0,0 +1,326 @@
...
+pub struct CampaignBudgetEvaluationV1 {
+    pub lane: CampaignLaneV1,
+    pub observation: CampaignLaneObservationV1,
+    pub flake_rate_basis_points: u64,
+    pub violations: Vec<String>,
+    pub passed: bool,
+}
+
+impl CampaignLaneConfigV1 {
+    pub fn builtin(lane: CampaignLaneV1) -> Self {
```
erskingardner, line 122: These four match arms are a verbatim duplicate of the four files under `lanes/`, and those files are read by nothing except the test that asserts they're equal to this function. That's ~250 lines existing to guard a divergence the PR itself creates.

Collapse to one source of truth:

```rust
pub fn builtin(lane: CampaignLaneV1) -> Self {
    let json = match lane {
        CampaignLaneV1::PullRequest => include_str!("../lanes/pull-request.v1.json"),
        // ...
    };
    serde_json::from_str(json).expect("tracked lane manifest is valid")
}
```

That deletes the 200-line match body, the `campaign_lane_drift` validation branch, and `tracked_lane_manifests_equal_the_reviewed_builtin_policy`. The JSON files stay under review in git, which was the stated goal. (Note also `every_lane_has_all_resource_retention_and_flake_budgets` is tautological over hardcoded literals — it can't fail without someone editing the constant next to it.)

crates/convergence-campaign-runner/src/evidence.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub adapters: Vec<String>,
+    pub mutation_results: BTreeMap<String, bool>,
+    pub tested_boundaries: Vec<TestedBoundaryV1>,
+    #[serde(default)]
+    pub unresolved_counterexamples: Vec<String>,
+    pub residual_assumptions: Vec<String>,
+    pub untested_surfaces: Vec<String>,
+    pub budget: CampaignBudgetEvaluationV1,
+    #[serde(default)]
+    pub artifacts: Vec<EvidenceArtifactV1>,
```
erskingardner, line 29: `artifacts` is `#[serde(default)]`, so a bundle with no artifacts at all validates, and for supplied artifacts only the *syntax* of the digest is checked — `path` is never opened and the digest is never compared to file contents (the test passes `"0".repeat(64)`).

The doc added in this PR says the bundle requires "digest-pinned artifacts." Either require at least one artifact and hash the referenced file, or narrow the doc to say this is a portable schema whose digests are verified at ingestion.

crates/convergence-campaign-runner/tests/lane_policy.rs
```diff
@@ -0,0 +1,100 @@
...
+        mutation_results: BTreeMap::from([("inconsistent_decision_seam".into(), true)]),
+        tested_boundaries: vec![TestedBoundaryV1 {
+            dimension: "participants".into(),
+            minimum: "2".into(),
+            maximum: "64".into(),
+        }],
+        unresolved_counterexamples: Vec::new(),
+        residual_assumptions: vec!["eventual input-set equality".into()],
+        untested_surfaces: vec!["unbounded executions".into()],
+        budget: config.evaluate(CampaignLaneObservationV1::default()),
```
erskingardner: This is the tell: a release-hardening evidence bundle is proven valid using `CampaignLaneObservationV1::default()` — zero wall time, zero CPU, zero cases executed. The test is asserting that evidence of nothing is acceptable evidence.

Once the observation validation from the `lane.rs` comment is in place, this should construct a plausible non-empty observation, and there should be a companion test asserting that a zero-case observation is *rejected*.

crates/convergence-campaign-runner/lanes/weekly-manual.v1.json
```diff
@@ -0,0 +1,30 @@
+{
+  "schema_version": "1",
+  "lane": "weekly_manual",
+  "contents": {
+    "strict_formal_gate": true,
+    "fixed_vectors": true,
+    "engine_reference_relay_cases": 1024,
```
erskingardner: `1024` / `4096` here versus `--cases 4` in `convergence-weekly-lane`. If these numbers are targets rather than what the lane runs, the field names and the surrounding docs should say so — as written a reader takes a green weekly run as evidence that 5,120 cases passed.

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -1005,11 +1005,19 @@ that scaling the existing corpus is sufficient.
...
-- [ ] Nightly lane: seed matrices, file-backed storage, crash matrix, retained relays, mutation subset.
-- [ ] Weekly/manual lane: process/container soak, resource and constant sweeps.
-- [ ] Release-hardening lane: mixed-version and incident-derived corpus.
-- [ ] Define wall-clock, CPU, memory, disk, artifact-retention, and flake budgets for every lane.
+- [x] PR lane: strict formal gate, fixed vectors, small engine/reference/relay cases.
+- [x] Nightly lane: seed matrices, file-backed storage, crash matrix, retained relays, mutation subset.
+- [x] Weekly/manual lane: process/container soak, resource and constant sweeps.
+- [x] Release-hardening lane: mixed-version and incident-derived corpus.
+- [x] Define wall-clock, CPU, memory, disk, artifact-retention, and flake budgets for every lane.
+
```
erskingardner, line 1013: Flipping all five 6.3 items to `[x]` overstates what landed. "Define ... budgets for every lane" is fair — the budgets are defined. The other four claim lanes that either don't run what they declare (weekly/manual: sweeps, mixed-version, 5k cases) or can't run at all (nightly and everything downstream, via the missing `route_equivalence` target).

The paragraph below is also stronger than the code: "A typed budget evaluator fails wall-clock, CPU, RSS, disk, artifact-size, retry, or flake-rate overages" is true of the type and false of the pipeline, because nothing invokes `check-budget`. I'd re-word to state that budgets are defined and machine-checkable, and leave the enforcement items open until a lane actually emits observations.

erskingardner (commented):
## Review: changes needed before merge

The premise is sound — bounding CI cost and making assurance claims reproducible is worth doing. But the shipped artifact is almost entirely *declaration*, and two of the wired-up steps are broken in ways that would either fail the nightly workflow outright or pass it while doing nothing. As it stands this PR flips five `[ ]` boxes in the reliability plan to `[x]` on the strength of JSON files that nothing executes against.

### Blocking

1. **`--test route_equivalence` names a target that does not exist.** `crates/cgka-conformance-simulator/tests/` at this head SHA has no `route_equivalence.rs`. `cargo nextest run ... --test route_equivalence` errors out, so the new nightly step fails, and so does `convergence-nightly-lane` → `convergence-weekly-lane` → the entire new `convergence-hardening.yml` workflow. (Consistent with #1236's route-equivalence exit item still being open — this reference is ahead of the code.)
2. **The nightly real-container step filters on a test name that does not exist,** so it runs zero tests and exits green. The actual test is `real_container_nodes_survive_network_shaping_and_reach_exact_state`. A step that silently passes is worse than no step.
3. **No budget is ever measured or enforced.** `check-budget`, `check-evidence`, and `lane` are added to the CLI and invoked by nothing — no workflow, no `just` recipe. `just convergence-lane-policy` only asserts that the JSON files equal the compiled constants. Every declared wall/CPU/RSS/disk/artifact/flake limit can be blown while all lanes stay green, which is the opposite of the PR summary's "validate ... budgets as reviewed data".
4. **A campaign that executed nothing produces passing release evidence.** See the inline note on `lane.rs` and `lane_policy.rs`.

### Simplicity (per the review guidelines)

This is ~900 added lines to express four tables of constants. The four JSON files are byte-for-byte the `builtin()` match arms, and are read by exactly one consumer: the test that checks they match. Pick one source of truth. `include_str!` + `serde_json::from_str` in `lane.rs` deletes the entire `builtin()` function, the drift concept, and the drift test — roughly 250 lines gone with no loss of capability.

I'd also suggest landing this as *planning metadata* with the assurance language narrowed, or instrumenting one lane end-to-end (observe → `check-budget` → fail) and letting the other three follow. The current middle ground buys a green badge for coverage that isn't running.

crates/convergence-campaign-runner/src/main.rs
```diff
@@ -102,6 +129,46 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
...
+            if !evaluation.passed {
+                return Err("campaign exceeded its reviewed lane budget".into());
+            }
+        }
+        Commands::CheckEvidence {
+            bundle: bundle_path,
+        } => {
+            let base_dir = bundle_path
+                .parent()
+                .unwrap_or_else(|| std::path::Path::new("."));
```
erskingardner: `Path::parent` returns `Some("")`, not `None`, for a single-component relative path, so this fallback never fires where it's needed:

```
Path::new("evidence.json").parent()  // Some("")
fs::canonicalize("")                // Err(NotFound)
```

So `cgka-distributed-campaign check-evidence evidence.json` fails with `evidence_artifact_base: No such file or directory` before it ever reads a bundle. Only paths with an explicit directory component work. Suggest:

```rust
let base_dir = match bundle_path.parent() {
    Some(parent) if !parent.as_os_str().is_empty() => parent,
    _ => Path::new("."),
};
```

A test invoking the bundle by bare filename from a `current_dir` would have caught this.

crates/convergence-campaign-runner/src/evidence.rs
```diff
@@ -0,0 +1,150 @@
...
+                    "evidence_artifact_digest_mismatch",
+                    "artifact bytes do not match the declared SHA-256 digest",
+                ));
+            }
+        }
+        Ok(())
+    }
+
+    pub fn write_private(&self, path: &Path) -> Result<(), RunnerError> {
+        self.validate_artifacts(path.parent().unwrap_or_else(|| Path::new(".")))?;
```
erskingardner: Same `Path::parent` defect as `main.rs`: `write_private(Path::new("evidence.json"))` passes `""` as the base dir and always fails in `canonicalize`. The existing test only writes to `temp.path().join("evidence.json")`, which has a real parent, so the failure mode isn't covered. Fold the empty-parent case into the fallback (a small `fn bundle_base_dir(path: &Path) -> &Path` shared with `main.rs` would keep both sites honest).

erskingardner (commented):
## Re-review: prior blockers resolved, one new bug

All four blocking findings from my earlier review are addressed, and the non-blocking notes were taken seriously:

1. **`--test route_equivalence`** — removed from both `simulator-nightly.yml` and `convergence-nightly-lane`. Only existing targets are named now. Resolved.
2. **Silently-empty container step** — the filter now names `real_container_nodes_survive_network_shaping_and_reach_exact_state`, which is the one `#[ignore]`d test in `container_runtime.rs`, and `--no-tests fail` turns a zero-match filter into a failure. Resolved. (With `--run-ignored ignored-only` on a single-test binary the `-E` filter is currently redundant, but it is the right kind of redundancy.)
3. **Overclaimed coverage** — the four §6.3 boxes are back to `[ ]`, the changelog row and both docs now say plainly that no workflow emits observations or invokes `check-budget`, and the lane contents were corrected to match what the recipes actually launch (`incident_corpus: false`, weekly `mixed_version: false`, nightly sweeps `true` — which `adversarial-reliability-ci` does in fact run). This is the honest version.
4. **Empty campaign yielding valid evidence** — `evaluate` now rejects `executed_cases: 0`, work below `minimum_executed_cases`, and `flaky_cases > executed_cases`, with a test. Resolved.

Also resolved: the duplicated `lanes/*.json` are gone as a second source (`include_str!` + `from_str` in `builtin`, ~250 lines deleted, drift test replaced), the redundant scheduled Tamarin installs are gone (`formal` runs on every PR and push to `master`, so the inheritance claim holds), and the dead `convergence-pr-lane` recipe is deleted. `just convergence-lane-policy` is now wired into PR CI, and the three `retention-days` values line up with the three lane manifests.

New work reviewed: `validate_mixed_builds` behind `--require-mixed-builds` is correct and covered — it also catches the subtle case where a participant with no explicit image inherits `default_participant_image`. Artifact digest verification with the relative-path/traversal guard plus `canonicalize` + `starts_with` is sound.

### One new bug

The evidence base-directory derivation is wrong for a bare relative filename (see inline). `check-evidence evidence.json` — the natural invocation, and what `check-budget --output` style usage would produce — fails with a confusing `evidence_artifact_base: No such file or directory`. Same defect in `write_private`. The tests only exercise absolute temp paths, so it's uncovered.

### Non-blocking

- `lane`, `check-budget`, and `check-evidence` are still invoked by zero workflows and zero recipes. This is now honestly disclosed rather than overclaimed, which is the important fix, but it remains a CLI surface with no caller. If the observation collector isn't landing very soon, consider holding these three subcommands until the producer exists.
- `write_private` now refuses to write unless every declared artifact already exists beside the destination. That's a defensible ordering constraint, but it makes a write function depend on unrelated filesystem state; worth a doc comment on the public method so callers aren't surprised.
- The nightly workflow collapsed five named steps into one `just convergence-nightly-lane` step, so a nightly failure no longer says which stage broke in the job summary. The recipe comment acknowledges the tradeoff for PRs; consider keeping the nightly split too, since nightly is the run nobody is watching live.
- `CampaignLaneConfigV1::validate` is now only reachable from `builtin` and tests — no external lane config is ever parsed. Fine as an invariant check, just note it isn't guarding an input boundary anymore.

Happy to approve once the base-directory bug is fixed.

Datawav:

## Final audit

### [P1] Recompute budget results when validating evidence  
`crates/convergence-campaign-runner/src/evidence.rs:91`

`ConvergenceEvidenceBundleV1::validate()` accepts the serialized `budget.passed` flag without deriving it from `budget.observation`. A producer can change a failing evaluation to `"passed": true`—while retaining over-budget measurements, violations, or too few executed cases—and `check-evidence` will accept the bundle.

- **Need:** Evidence validation must establish that recorded observations satisfy the reviewed lane policy.
- **Design:** Re-evaluate `budget.observation` using `CampaignLaneConfigV1::builtin(self.lane)` and require the supplied evaluation to equal the computed result.
- **Alternative:** Store only the observation in the bundle and always derive the evaluation during validation.
- **Test:** Tamper a failing evaluation’s `passed` field and verify `check-evidence` rejects it.

### [P2] Enforce retention from the lane policy source  
`.github/workflows/ci.yml:366`, `.github/workflows/simulator-nightly.yml:87`, `.github/workflows/convergence-hardening.yml:97`

`artifact_retention_days` is duplicated in workflow YAML rather than enforced from the lane manifests. The values currently agree, but changing a manifest can silently leave artifact retention unchanged, contradicting the stated single-policy-source design.

- **Need:** The reviewed retention budget and actual uploaded-artifact lifetime must not drift.
- **Design:** Add a policy check that asserts each workflow retention value matches its lane manifest.
- **Alternative:** Generate workflow inputs from the manifests, or remove the manifest field until runtime enforcement exists.

_Advisory audit of head [`f443e7532112`](https://github.com/marmot-protocol/mdk/commit/f443e7532112214a8957d6e1133fd33b314e5209); no approval, merge, push, checkout, or PR-code execution was performed._

PR #1272, opened 2026-08-05
Add convergence failure corpus lifecycle

erskingardner:
## Summary

- add a durable convergence failure-corpus lifecycle for indexing, recurrence classification, diagnosis, reduction, and fixture promotion
- make corpus mutations transactional with an interprocess lease and same-directory atomic replacement
- preserve fault/recovery dependency units throughout semantic minimization
- require validated synthetic-shareable capsule provenance and SHA-256 evidence for vector promotion
- wire focused corpus lifecycle verification into the runner and CI documentation

## Stack

This is the third independent slice extracted from #1250:

1. #1270 — distributed campaign runner core (merged)
2. #1271 — execution lanes and evidence budgets (merged)
3. this PR — failure corpus and semantic reduction

This PR is rebased directly on current `master`; review its delta against `master`.

## Scope boundary

This PR does **not** modify production convergence behavior. It contains no changes in `cgka-engine`, `traits`, `cgka-session`, `marmot-account`, or `marmot-app`, and it does not assume any particular resolution of #1236. Route-assurance and engine-semantic work from #1250 remains parked for team design discussion.

The four lane manifests intentionally keep `incident_corpus: false`: this slice validates the corpus lifecycle machinery but does not yet execute a real reviewed incident-derived corpus.

## Verification

- `just fast-ci`
- `cargo test -p convergence-campaign-runner --locked --test failure_corpus`
- `cargo test -p convergence-campaign-runner --locked --test lane_policy`
- `cargo test -p convergence-campaign-runner --locked --lib`
- `cargo test -p cgka-conformance-simulator --locked --test semantic_reduction`
- `cargo test -p cgka-conformance-simulator --locked --lib minimizer_never_splits_a_fault_and_recovery_pair`
- `just convergence-failure-corpus`
- `git diff --check`

Extracted from #1250.

## Summary by CodeRabbit

* **New Features**
  * Added failure-corpus workflows for recording, classifying, diagnosing, deduplicating, and tracking distributed execution failures.
  * Added tools to index failure data and promote validated, privacy-safe reproductions into fixed test-vector candidates.
  * Added recurrence tracking, provenance, digest validation, and concurrent-update safeguards.
* **Reliability**
  * Failed campaigns preserve their original errors while recording diagnostic failure data.
  * Scenario minimization better preserves paired fault and recovery steps.
* **Documentation**
  * Documented failure-corpus lifecycle policies and milestone completion.
* **Tests**
  * Added comprehensive lifecycle, promotion, concurrency, semantic-reduction, and CI validation coverage.

### DIFF
--- a/crates/cgka-conformance-simulator/src/family.rs
+++ b/crates/cgka-conformance-simulator/src/family.rs
@@ -391,16 +391,21 @@ where
 /// commonly produces a different terminal failure. These units are attempted
 /// before individual noise steps.
 pub fn semantic_reduction_units(steps: &[ScenarioStep]) -> Vec<Vec<usize>> {
-    let mut open = std::collections::BTreeMap::<String, usize>::new();
+    let mut open = std::collections::BTreeMap::<SemanticPairKey, Vec<usize>>::new();
     let mut units = Vec::new();
     for (index, step) in steps.iter().enumerate() {
-        let Some((key, opens)) = semantic_pair_marker(step, None) else {
+        let Some((key, opens)) = semantic_pair_marker(step, &[]) else {
             continue;
         };
         if opens {
-            open.insert(key, index);
-        } else if let Some(start) = open.remove(&key) {
-            units.push(vec![start, index]);
+            open.entry(key).or_default().push(index);
+        } else if let Some(mut starts) = open.remove(&key) {
+            if let Some(start) = starts.pop() {
+                units.push(vec![start, index]);
+            }
+            if !starts.is_empty() {
+                open.insert(key, starts);
+            }
         }
     }
     units
@@ -415,44 +420,70 @@ fn semantic_minimization_units(steps: &[ScenarioStep]) -> Vec<Vec<usize>> {
     units
 }
 
-fn semantic_pair_marker(step: &ScenarioStep, group: Option<&str>) -> Option<(String, bool)> {
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+struct SemanticPairKey {
+    group_path: Vec<String>,
+    resource: SemanticPairResource,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+enum SemanticPairResource {
+    Withheld(String),
+    Partition,
+    Offline(String),
+    Process(String),
+    Storage(String),
+}
+
+fn semantic_pair_marker(
+    step: &ScenarioStep,
+    group_path: &[String],
+) -> Option<(SemanticPairKey, bool)> {
     if let ScenarioStep::InGroup {
         group: nested_group,
         action,
     } = step
     {
-        return semantic_pair_marker(action, Some(nested_group));
+        let mut nested_path = group_path.to_vec();
+        nested_path.push(nested_group.clone());
+        return semantic_pair_marker(action, &nested_path);
     }
-    let prefix = group.map_or_else(String::new, |group| format!("group:{group}:"));
-    match step {
+    let (resource, opens) = match step {
         ScenarioStep::WithholdMessage { label, .. } => {
-            Some((format!("{prefix}withhold:{label}"), true))
+            (SemanticPairResource::Withheld(label.clone()), true)
         }
         ScenarioStep::ReleaseWithheld { label } => {
-            Some((format!("{prefix}withhold:{label}"), false))
+            (SemanticPairResource::Withheld(label.clone()), false)
         }
-        ScenarioStep::SetPartition { .. } => Some((format!("{prefix}partition"), true)),
-        ScenarioStep::ClearPartition => Some((format!("{prefix}partition"), false)),
+        ScenarioStep::SetPartition { .. } => (SemanticPairResource::Partition, true),
+        ScenarioStep::ClearPartition => (SemanticPairResource::Partition, false),
         ScenarioStep::SetClientOffline { client } => {
-            Some((format!("{prefix}offline:{client}"), true))
+            (SemanticPairResource::Offline(client.clone()), true)
         }
         ScenarioStep::ReconnectClient { client } => {
-            Some((format!("{prefix}offline:{client}"), false))
+            (SemanticPairResource::Offline(client.clone()), false)
         }
         ScenarioStep::CrashProcess { process } => {
-            Some((format!("{prefix}process:{process}"), true))
+            (SemanticPairResource::Process(process.clone()), true)
         }
         ScenarioStep::RestartProcess { process } => {
-            Some((format!("{prefix}process:{process}"), false))
+            (SemanticPairResource::Process(process.clone()), false)
         }
         ScenarioStep::InjectStorageFault { fault } => {
-            Some((format!("{prefix}storage:{}", fault.target), true))
+            (SemanticPairResource::Storage(fault.target.clone()), true)
         }
         ScenarioStep::ClearStorageFault { target } => {
-            Some((format!("{prefix}storage:{target}"), false))
+            (SemanticPairResource::Storage(target.clone()), false)
         }
-        _ => None,
-    }
+        _ => return None,
+    };
+    Some((
+        SemanticPairKey {
+            group_path: group_path.to_vec(),
+            resource,
+        },
+        opens,
+    ))
 }
 
 fn remove_step_indices(steps: &mut Vec<ScenarioStep>, indices: &[usize]) {
--- a/crates/cgka-conformance-simulator/src/family.rs
+++ b/crates/cgka-conformance-simulator/src/family.rs
@@ -339,61 +339,50 @@ async fn minimize_failing_case(
     let target_identity = fingerprint_report_failure(failing_report)
         .ok()?
         .semantic_identity();
-
-    let mut candidate = case.scenario.clone();
-    let mut changed = false;
-    loop {
-        let mut reduced_unit = false;
-        for unit in semantic_reduction_units(&candidate.steps) {
-            let mut trial = candidate.clone();
-            remove_step_indices(&mut trial.steps, &unit);
-            if reproduces_failure(
+    minimize_scenario_with(case.scenario.clone(), |trial| {
+        let expected_trace = expected_trace.cloned();
+        let expected_outcomes = case.expected_outcomes.clone();
+        let target_identity = target_identity.clone();
+        async move {
+            reproduces_failure(
                 &trial,
-                expected_trace.cloned(),
-                case.expected_outcomes.clone(),
+                expected_trace,
+                expected_outcomes,
                 &target_identity,
                 storage_mode,
                 case.subject,
             )
             .await
-            {
+        }
+    })
+    .await
+}
+
+async fn minimize_scenario_with<F, Fut>(
+    mut candidate: ScenarioSpec,
+    mut reproduces: F,
+) -> Option<ScenarioSpec>
+where
+    F: FnMut(ScenarioSpec) -> Fut,
+    Fut: std::future::Future<Output = bool>,
+{
+    let mut changed = false;
+    loop {
+        let mut reduced = false;
+        for unit in semantic_minimization_units(&candidate.steps) {
+            let mut trial = candidate.clone();
+            remove_step_indices(&mut trial.steps, &unit);
+            if reproduces(trial.clone()).await {
                 candidate = trial;
                 changed = true;
-                reduced_unit = true;
+                reduced = true;
                 break;
             }
         }
-        if !reduced_unit {
+        if !reduced {
             break;
         }
     }
-
-    let mut index = 0;
-    while index < candidate.steps.len() {
-        if !is_minimizer_removable(&candidate.steps[index]) {
-            index += 1;
-            continue;
-        }
-
-        let mut trial = candidate.clone();
-        trial.steps.remove(index);
-        if reproduces_failure(
-            &trial,
-            expected_trace.cloned(),
-            case.expected_outcomes.clone(),
-            &target_identity,
-            storage_mode,
-            case.subject,
-        )
-        .await
-        {
-            candidate = trial;
-            changed = true;
-        } else {
-            index += 1;
-        }
-    }
-
     changed.then_some(candidate)
 }
 
@@ -414,6 +403,15 @@ pub fn semantic_reduction_units(steps: &[ScenarioStep]) -> Vec<Vec<usize>> {
             units.push(vec![start, index]);
         }
     }
+    units
+}
+
+fn semantic_minimization_units(steps: &[ScenarioStep]) -> Vec<Vec<usize>> {
+    let mut units = semantic_reduction_units(steps);
+    units.extend(steps.iter().enumerate().filter_map(|(index, step)| {
+        (semantic_pair_marker(step, None).is_none() && is_minimizer_removable(step))
+            .then_some(vec![index])
+    }));
     units
 }
 
@@ -2783,5 +2781,45 @@ mod tests {
         }];
 
         add_strict_reliability_oracle(&mut scenario, &mut expected);
+    }
+
+    #[tokio::test]
+    async fn minimizer_never_splits_a_fault_and_recovery_pair() {
+        let scenario = ScenarioSpec {
+            name: "paired-reduction".into(),
+            spec_version: "2".into(),
+            topology: Default::default(),
+            clients: vec!["alice".into()],
+            steps: vec![
+                ScenarioStep::SetPartition {
+                    allow: vec!["alice".into()],
+                },
+                ScenarioStep::ClearEvents {
+                    clients: vec!["alice".into()],
+                },
+                ScenarioStep::ClearPartition,
+            ],
+        };
+
+        let minimized = minimize_scenario_with(scenario, |trial| async move {
+            let has_partition = trial
+                .steps
+                .iter()
+                .any(|step| matches!(step, ScenarioStep::SetPartition { .. }));
+            let has_clear = trial
+                .steps
+                .iter()
+                .any(|step| matches!(step, ScenarioStep::ClearPartition));
+            has_partition || has_clear
+        })
+        .await
+        .expect("independent noise should be removable");
+
+        assert_eq!(minimized.steps.len(), 2);
+        assert!(matches!(
+            minimized.steps[0],
+            ScenarioStep::SetPartition { .. }
+        ));
+        assert!(matches!(minimized.steps[1], ScenarioStep::ClearPartition));
     }
 }
--- a/crates/cgka-conformance-simulator/src/family.rs
+++ b/crates/cgka-conformance-simulator/src/family.rs
@@ -342,6 +342,32 @@ async fn minimize_failing_case(
 
     let mut candidate = case.scenario.clone();
     let mut changed = false;
+    loop {
+        let mut reduced_unit = false;
+        for unit in semantic_reduction_units(&candidate.steps) {
+            let mut trial = candidate.clone();
+            remove_step_indices(&mut trial.steps, &unit);
+            if reproduces_failure(
+                &trial,
+                expected_trace.cloned(),
+                case.expected_outcomes.clone(),
+                &target_identity,
+                storage_mode,
+                case.subject,
+            )
+            .await
+            {
+                candidate = trial;
+                changed = true;
+                reduced_unit = true;
+                break;
+            }
+        }
+        if !reduced_unit {
+            break;
+        }
+    }
+
     let mut index = 0;
     while index < candidate.steps.len() {
         if !is_minimizer_removable(&candidate.steps[index]) {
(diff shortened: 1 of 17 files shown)


Datawav:

## Final audit

The PR addresses a real need: recurring convergence failures currently lack durable indexing, semantic minimization, classification, diagnosis timing, and controlled regression-vector promotion. The design centralizes these in a private JSON corpus, adds cross-adapter reduction candidates, and introduces dependency-aware scenario reduction. Reasonable alternatives include an append-only event log or transactional store for the corpus, and typed failure origins instead of string-based classification.

### Findings

1. **High — Semantic minimization can still split dependency pairs**  
   `crates/cgka-conformance-simulator/src/family.rs:342` first attempts paired removal, but then the existing individual-removal pass processes `WithholdMessage`, `ReleaseWithheld`, `SetPartition`, and `ClearPartition` separately. Consequently, a pair that cannot be removed together may still be reduced to one half if that half reproduces the same terminal identity, contradicting the stated guarantee that minimization preserves fault/recovery semantics. Exclude all steps belonging to semantic pairs from individual reduction, or make the complete reducer operate exclusively on dependency-aware units plus genuinely independent steps. Add an end-to-end minimizer test, not only tests of `semantic_reduction_units`.

2. **Medium — Corpus-recording failure replaces the actual campaign failure**  
   `crates/convergence-campaign-runner/src/runner.rs:88` returns `failure_corpus_record` when execution and evidence persistence both fail. The original typed error survives only as text, so callers lose its code and associated retry or classification behavior. This is particularly likely when output or scenario access caused the original failure. Preserve the execution error as the primary result and report the corpus-write failure as secondary context. An alternative is a compound error type retaining both typed causes.

3. **Medium — The “durable” corpus uses an unsafe read-modify-rewrite cycle**  
   `crates/convergence-campaign-runner/src/failure_corpus.rs:214` rewrites the JSON file through `fs_private::write_private`, which opens it with truncation before writing. Interruption can therefore leave corrupt JSON, while concurrent runner or CLI writers can overwrite each other’s observations and recurrence counts. Use a same-directory temporary file, flush, and atomic rename, together with interprocess locking; alternatively use an append-only journal or transactional database. Cover interrupted replacement and concurrent writers.

4. **Medium — Substring classification can suppress product-defect investigation**  
   `crates/convergence-campaign-runner/src/failure_corpus.rs:391` classifies any code containing `timeout` or `spawn`, and broad `output_`, `command_`, or `vm_` prefixes, as an environment failure. That classification also marks the failure layer-specific and prevents creation of a smaller-adapter reduction candidate. A product convergence timeout can therefore be automatically diverted from defect reduction, contrary to the documentation’s conservative-classification claim. Use typed error provenance or an explicit allowlist of known infrastructure codes; default unknown failures to `ProductDefect` and test ambiguous timeout cases.

_Advisory audit of head [`50fdfc777d84`](https://github.com/marmot-protocol/mdk/commit/50fdfc777d842656e12d10da22b1e45847a627f3); no approval, merge, push, checkout, or PR-code execution was performed._

crates/convergence-campaign-runner/src/runner.rs
```diff
@@ -86,10 +88,22 @@ pub async fn run_manifest(
...
     .map_err(|error| RunnerError::environment("manifest_write", error))?;
-    match &manifest.backend {
+    let result = match &manifest.backend {
         DistributedBackendV1::Container(_) => run_container(manifest, &plan, &scenario).await,
         DistributedBackendV1::VirtualMachine(_) => run_vm(manifest, &plan).await,
+    };
+    if let Err(error) = &result
+        && let Err(corpus_error) = record_distributed_failure(manifest, error)
+    {
+        return Err(RunnerError::validation(
```
erskingardner: This discards the real failure. `error` is the typed campaign failure (`container_node_launch`, `command_timeout`, `scenario_compile`, ...) and it is replaced by a `failure_corpus_record` *validation* error whose only trace of the original is the formatted message.

Failure scenario: `manifest.output_dir` is not writable (or the scenario file was removed mid-run). `run_container` fails with `container_node_launch`; `record_distributed_failure` then fails with `failure_corpus_directory`; the caller sees `failure_corpus_record` and has no way to key on the actual cause. Bookkeeping should never outrank the thing it is bookkeeping.

Suggest keeping the execution error primary:

```rust
if let Err(error) = &result
    && let Err(corpus_error) = record_distributed_failure(manifest, error)
{
    eprintln!(
        "failure_corpus_record: could not record {} ({})",
        error.code, corpus_error.code
    );
}
result
```

Or a compound error that retains both typed causes if the write failure must be fatal.

crates/convergence-campaign-runner/src/failure_corpus.rs
```diff
@@ -0,0 +1,440 @@
...
+    if corpus.schema_version != FAILURE_CORPUS_SCHEMA_VERSION {
+        return Err(RunnerError::validation(
+            "failure_corpus_version",
+            "unsupported failure corpus schema version",
+        ));
+    }
+    Ok(corpus)
+}
+
+pub fn write_failure_corpus(path: &Path, corpus: &FailureCorpusV1) -> Result<(), RunnerError> {
```
erskingardner, line 277: This is the durability gap for a corpus whose whole purpose is durability.

`fs_private::write_private` opens with `.truncate(true)` and then writes, so the window between truncate and `sync_all` leaves a partial file on disk. Every mutation in this module is read -> mutate -> full rewrite, so an interrupted CI job, an OOM kill, or a runner timeout during the rewrite corrupts the JSON. Because `read_failure_corpus` returns a hard `failure_corpus_parse` error rather than recovering, the corpus is then permanently unusable: every later `index-capsule`, `classify-failure`, `diagnose-failure`, and every `record_distributed_failure` call fails. And the moments this file is written are precisely the moments the campaign is already failing or being killed.

This also conflicts with the repo invariant in `docs/marmot-architecture/overview/multi-step-state-changes.md` ("Keep multi-step state changes torn-write-free").

Write to a same-directory temp file with the private mode, `sync_all`, then `rename` onto `path` — atomic on POSIX, keeps 0600, and is a handful of lines. A test that leaves a truncated file in place and asserts the next write still succeeds would lock it in.

Secondary: concurrent writers (two CLI invocations, or a runner plus an operator command against the same output dir) silently lose each other's observations and recurrence counts under this read-modify-write. An advisory lock or an append-only journal would close that too, if it is in scope.

crates/convergence-campaign-runner/src/failure_corpus.rs
```diff
@@ -0,0 +1,440 @@
...
+    Ok(path)
+}
+
+fn classify_runner_error(error: &RunnerError) -> FailureClassificationV1 {
+    if error.code == "expected_resource_refusal" {
+        FailureClassificationV1::ExpectedResourceRefusal
+    } else if error.code == "protocol_ambiguity" {
+        FailureClassificationV1::ProtocolAmbiguity
+    } else if error.code.contains("spawn")
+        || error.code.contains("timeout")
```
erskingardner: Substring classification is too broad in the direction that hurts.

`runner.rs` emits `command_timeout` when a node command exceeds its bound — which is exactly what a convergence deadlock or livelock in the product looks like at this layer. That code hits `contains("timeout")`, becomes `EnvironmentFailure`, and (via the `classification == EnvironmentFailure` argument passed as `layer_specific` in `record_distributed_failure`) also suppresses the cross-adapter reduction candidate. A real product hang is therefore auto-filed as infra noise and excluded from reduction.

That is the opposite of the policy this PR documents in `distributed-convergence-campaigns.md`: "Environment failures are recognized conservatively" and failures "remain product defects until an operator explicitly establishes" otherwise.

Prefer an explicit allowlist of known-infrastructure codes with `ProductDefect` as the default, e.g.

```rust
const ENVIRONMENT_CODES: &[&str] = &[
    "command_spawn", "output_directory", "vm_plan", /* ... */
];
```

and let an operator move an ambiguous timeout with `classify-failure`. Note also that against the current code set `starts_with("command_")` already covers both `contains("spawn")` and `contains("timeout")`, so those two arms only add reach into codes that do not exist yet.

crates/cgka-conformance-simulator/src/family.rs
```diff
@@ -342,6 +342,32 @@ async fn minimize_failing_case(
 
     let mut candidate = case.scenario.clone();
     let mut changed = false;
+    loop {
```
erskingardner: This pass does not deliver the coherence guarantee the PR summary and the docs claim ("keeps fault/recovery pairs ... coherent").

The paired pass runs first, but the pre-existing individual pass below it still treats `WithholdMessage`, `ReleaseWithheld`, `SetPartition`, and `ClearPartition` as independently removable (`is_minimizer_removable`, line 525). So a pair that cannot be removed *together* can still be split by the second pass — leaving, say, a `ClearPartition` with no `SetPartition`, or a `SetPartition` that is never healed. The reproducer stays valid (`reproduces_failure` still gates it), but the fault/recovery semantics the unit pass was added to protect are not actually preserved.

Either exclude every index that appears in a `semantic_reduction_units` unit from the individual pass, or have the individual pass operate only on genuinely independent steps. Whichever way, please add an end-to-end `minimize_failing_case` test — the two tests in `semantic_reduction.rs` only cover `semantic_reduction_units` itself, so the interaction that actually matters is untested.

Separately, worth a sanity check on cost: this outer loop restarts from scratch on every successful unit removal, and each trial is a full `reproduces_failure` scenario execution on top of the existing per-step pass. Fine if minimization is already off the hot path, but this is now `O(units^2)` full runs in the worst case.

erskingardner (commented):
Solid slice overall: the scope boundary is honored (no engine/traits/session/account/app changes), the corpus is message-free with owner-only permissions, and the tests cover recurrence, classification, diagnosis, and promotion. Two things block for me, both in new code and both cheap to fix.

**Blocking**

1. The "durable" corpus is persisted with a non-atomic read-modify-truncate-rewrite. `fs_private::write_private` opens with `truncate(true)`, so an interrupted write leaves partial JSON, and `read_failure_corpus` hard-errors on a parse failure. From that point every subsequent `index-capsule` / `classify-failure` / `diagnose-failure` and every distributed-run failure path fails with `failure_corpus_parse` — the corpus is bricked, not degraded. This also runs against the repo invariant "Keep multi-step state changes torn-write-free" (`docs/marmot-architecture/overview/multi-step-state-changes.md`). Same-directory temp file + `sync_all` + `rename` is a small change and keeps the 0600 posture.

2. `run_manifest` replaces the real campaign failure with a synthetic `failure_corpus_record` validation error. The typed code that CI and callers key on is lost; the original survives only as prose. This is most likely exactly when the corpus write also fails (bad `output_dir`, unreadable scenario). Keep `result` as the returned error and surface the corpus write failure as secondary context.

**Non-blocking, but worth addressing before merge**

3. `classify_runner_error` uses substring matching on error codes, and today `command_timeout` is the code a distributed node emits when it hangs. A genuine convergence livelock is therefore auto-filed as `EnvironmentFailure` **and** marked `layer_specific = true`, so no reduction candidate is emitted — the failure is quietly excluded from exactly the reduction pipeline this PR exists to build. That contradicts the doc text added here ("Environment failures are recognized conservatively"). An explicit allowlist of infrastructure codes, defaulting unknown to `ProductDefect`, would match the stated policy. Note also that with the current code set, `starts_with("command_")` already subsumes both `contains("spawn")` and `contains("timeout")` — those two arms are redundant breadth today.

4. The dependency-aware pass does not actually guarantee the coherence the summary and docs claim; see the inline note on `family.rs`.

5. `record_distributed_failure` re-reads and re-parses the scenario from disk, even though `run_manifest` already has verified scenario bytes in scope. Passing them in removes an I/O round-trip and an entire failure mode from the error path.

Nits: the changelog row added to `convergence-reliability-plan.md` is dated 2026-08-05 but sits after the 2026-08-06 row for 6.3 — looks like it should be 2026-08-06. And flipping `incident_corpus` to `true` in all four lane manifests is a fairly generous reading of "executes the incident corpus" when `just convergence-failure-corpus` only runs this crate's own unit tests; #1271 explicitly deferred that flag to this PR, so it is worth being precise about what the flag now attests to.

Datawav:

⚠️ Address these three findings.

1. **[P1] Concurrent or interrupted updates can lose the failure corpus**  
   `failure_corpus.rs:244-256`, `main.rs:220-279`, `failure_corpus.rs:332-376`

   Every operation performs an unlocked read–modify–write. `fs_private::write_private` truncates the destination before writing, so parallel campaign/indexing processes can overwrite one another’s observations, and interruption after truncation can corrupt the entire corpus. This conflicts with the PR’s need for durable recurrence evidence.

   Design writes around an exclusive lock and commit through a private temporary file followed by flush and atomic rename. Alternatively, use append-only observation records and derive the index separately. Add concurrent-writer and interrupted-write coverage.

2. **[P1] Corpus-recording errors replace the actual campaign failure**  
   `runner.rs:88-105`

   When backend execution fails and `record_distributed_failure` also fails, `run_manifest` returns `failure_corpus_record` instead of the original container/VM error. A corrupt corpus or storage problem therefore hides the failure operators need to diagnose and changes its classification.

   Preserve the backend error as the primary result and attach corpus persistence failure as secondary context. An alternative is best-effort recording with a warning/side-channel diagnostic, but the original run error must remain observable.

3. **[P2] Diagnosis can claim an unvalidated vector was promoted**  
   `failure_corpus.rs:202-216`, `main.rs:267-279`

   `diagnose-failure --promoted-vector` inserts any path directly into `promoted_vectors`; it does not require the file to exist or verify schema, provenance, shareability, or that `promote-capsule` produced it. This bypasses the stated design that only synthetic shareable capsules with portable expectations may become promoted vectors.

   Record promotion only as part of a successful `promote-capsule` operation, including validated provenance or a digest. Alternatively, make diagnosis validate the referenced vector and its source capsule before updating the corpus.

_Advisory audit of head [`80220ac40dc6`](https://github.com/marmot-protocol/mdk/commit/80220ac40dc6ef0df9aded27336ea3fcd98025f2); no approval, merge, push, checkout, or PR-code execution was performed._

Datawav:

1. **High — Promotion can attach an unrelated capsule to a corpus entry**  
   `crates/convergence-campaign-runner/src/failure_corpus.rs:390` validates the requested fingerprint and capsule independently, but never verifies `capsule.failure.digest == fingerprint`. A typo can therefore record a valid vector and provenance under the wrong failure.

   **Need:** Preserve the corpus’s fingerprint-to-regression provenance invariant.  
   **Design:** Reject promotion unless the capsule digest exactly matches the selected entry; add a mismatch regression test.  
   **Alternative:** Resolve the target entry exclusively from `capsule.failure.digest`, removing the separate fingerprint argument.

2. **Medium — Distributed fingerprints cannot correlate recurrence across adapters**  
   `failure_corpus.rs:576` hashes the scenario digest, error code, and adapter. The same failure reproduced in container and VM becomes two entries, so the entry’s adapter set cannot represent cross-adapter recurrence as documented. Conversely, scenario plus coarse runner error code may merge distinct failures within one adapter.

   **Need:** Give the same semantic failure one stable identity across execution layers without collapsing different failure sites.  
   **Design:** Build the fingerprint from a shared `FailureIdentityV1` containing sufficiently specific phase/action and failure kind, excluding adapter; test identical failures across adapters and distinct failures sharing one runner error code.  
   **Alternative:** Keep adapter-specific entries but explicitly model cross-entry equivalence and revise the recurrence contract.

3. **Medium — Repeated fault arms are treated as independently removable**  
   `crates/cgka-conformance-simulator/src/family.rs:390` stores only one opening index per semantic key. A repeated partition, same-label withhold, or same-target fault overwrites the earlier opening; the earlier action then becomes a singleton minimization unit. This contradicts the stated guarantee that dependency pairs are never split. The reproduction gate limits damage but does not enforce that structural invariant.

   **Need:** Preserve dependency structure through semantic minimization.  
   **Design:** Track openings per key with a stack/queue, or reduce the complete same-key interval as one unit; add repeated and nested same-key tests.  
   **Alternative:** Explicitly reject overlapping same-key fault lifecycles during scenario validation if they are not meaningful inputs.

_Advisory audit of head [`e193a8f50073`](https://github.com/marmot-protocol/mdk/commit/e193a8f500736fa314e3f37a9ebb9c949b511daa); no approval, merge, push, checkout, or PR-code execution was performed._

Datawav:

## Final audit

### Need and design

The PR addresses a legitimate gap: durable failure indexing, recurrence/diagnosis tracking, semantic reduction, and controlled promotion of reproducible failures into regression vectors. The overall design—private transactional corpus updates, semantic identities, adapter-reduction candidates, and digest-pinned promotion—is appropriate.

Alternatives considered by the design include keeping failures as independent artifacts without a corpus, using generic step-by-step minimization, or promoting capsules without provenance. The proposed corpus and dependency-aware reducer are preferable, but the following correctness issues remain.

### Actionable findings

1. **High — Promotion can associate the wrong capsule with a corpus entry**

   In `promote_capsule_into_corpus` (`crates/convergence-campaign-runner/src/failure_corpus.rs`), the requested `fingerprint` is checked only for syntax and corpus membership. The independently loaded capsule is validated, but `capsule.failure.digest` is never compared with that fingerprint. A valid capsule for failure B can therefore be promoted and recorded under failure A, corrupting regression ownership and provenance.

   Require `capsule.failure.digest == fingerprint` before writing the vector, and test a valid capsule paired with a different existing corpus fingerprint.

2. **Medium — Nested group scope is lost during semantic pairing**

   `semantic_pair_marker` (`crates/cgka-conformance-simulator/src/family.rs`) replaces the current group scope whenever it recurses through `InGroup`; it does not accumulate the full nested path. Thus faults in distinct outer groups with identically named inner groups can be paired together and removed as one unit, changing scenario semantics.

   Carry a full group path through recursion, using an unambiguous structured key, and add a test with different outer groups sharing the same inner-group name.

3. **Medium — Environmental reclassification leaves a contradictory reduction candidate**

   `FailureCorpusV1::reclassify` changes only `classification`. Entries initially treated as product defects can already contain cross-adapter reduction candidates, while observation construction explicitly suppresses those candidates for environmental failures. Reclassifying such an entry to `environment_failure` leaves state that contradicts the documented policy.

   Clear the candidate when changing to `EnvironmentFailure`. Alternatively, model candidate applicability separately and validate the invariant when serializing the corpus.

4. **Medium — Node-capsule fingerprints use an ambiguous encoding**

   `observation_from_node_capsule` hashes a colon-joined string containing `scenario_digest`, `participant`, `action_id`, and `code`. Because the variable fields are not escaped or length-prefixed, values such as `participant="a:b", action_id="c"` and `participant="a", action_id="b:c"` produce identical preimages and merge distinct failures.

   Hash a canonical serialized structure or a length-prefixed tuple instead. Add a test demonstrating that delimiter-containing fields remain distinct.

5. **Medium — Reusing a promotion output path corrupts recorded provenance**

   `write_private_atomic` replaces an existing output, while `promoted_vectors` can retain multiple records with the same path and different digests. A later promotion to that path changes the file but leaves earlier corpus records claiming different content exists there.

   Either reject an existing output path or atomically replace the prior record for that path. A content-addressed output name would avoid this ambiguity entirely.

_Advisory audit of head [`08c01dfe1d21`](https://github.com/marmot-protocol/mdk/commit/08c01dfe1d21529ab96d7714768b8b50685539d7); no approval, merge, push, checkout, or PR-code execution was performed._

Datawav:

## Final audit

The lifecycle is needed to retain, classify, reduce, and safely promote recurring convergence failures. The design uses a private versioned corpus, interprocess locking, atomic replacement, semantic fault/recovery reduction units, and digest-pinned synthetic capsule promotion. Keeping this separate from production convergence logic is reasonable. Alternatives include an append-only journal or transactional database; the JSON design remains viable if the following consistency gaps are fixed.

- **P1 — Promotion can permanently strand an unrecorded vector**  
  `crates/convergence-campaign-runner/src/failure_corpus.rs:493-509` publishes `output_path` before the updated corpus is serialized and replaced. A serialization/write/fsync failure or process crash after publication leaves the vector present without its provenance record. Retrying then fails with `promoted_vector_exists`, requiring manual intervention. This contradicts the claimed transactional lifecycle. Make promotion recoverable—for example, detect and adopt an existing byte-identical vector with matching capsule provenance, or use a journal/staged commit protocol. Add fault-injection coverage for every boundary between vector publication and corpus persistence.

- **P2 — A recurrence can restore a reduction candidate removed by classification**  
  `FailureCorpusV1::reclassify` clears `reduction_candidate` for an environment failure, but `record` later fills any missing candidate from a new observation without checking the persisted classification (`failure_corpus.rs:201-203`). Thus a reviewed environment classification can regain an inappropriate cross-adapter reduction candidate on the next recurrence, violating the documented rule that environment failures do not emit such candidates. Only populate candidates when the retained entry classification permits them, and test reclassification followed by another occurrence.

_Advisory audit of head [`079fae23089a`](https://github.com/marmot-protocol/mdk/commit/079fae23089aab1d69b3448d9d31e1508047c8be); no approval, merge, push, checkout, or PR-code execution was performed._

PR #1273, opened 2026-08-05
Enforce convergence capability naming

erskingardner:
## Summary

- reject tracked filenames containing `milestone` so long-lived artifacts use capability-oriented names
- replace two remaining milestone-scoped descriptions in the simulator with enduring process-adapter terminology
- rename the convergence ledger comment around the capability it guards

## Scope

This is the naming-only slice extracted from #1250. It is based directly on `master`, has no dependency on the distributed runner stack, and changes no convergence behavior.

## Verification

- `just naming-gate`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved process-adapter configuration validation by clearly reporting when multiple account-device participants are assigned to one process.
  - Added strict checks that reject tracked filenames containing “milestone” and display the offending paths.

- **Documentation**
  - Clarified campaign process isolation and distinguished it from application-runtime and child-process adapters.
  - Updated related validation terminology to reference reviewed convergence-constant ledger entries.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -445,10 +445,24 @@ naming-gate:
     #!/usr/bin/env bash
     set -euo pipefail
     ./scripts/check_legacy_naming.sh
-    tracked_milestone_names="$(git ls-files | awk -F/ 'tolower($NF) ~ /milestone/ { print }')"
-    if [[ -n "$tracked_milestone_names" ]]; then
-        echo "tracked filenames must use capability names, not milestone names:" >&2
-        echo "$tracked_milestone_names" >&2
+    is_milestone_path() {
+        local lowercase
+        lowercase="$(printf '%s' "$1" | tr '[:upper:]' '[:lower:]')"
+        [[ "$lowercase" == *milestone* ]]
+    }
+    is_milestone_path "docs/milestone-5/design.md"
+    is_milestone_path "docs/design-milestone.md"
+    is_milestone_path "docs/MileStone-6/design.md"
+    ! is_milestone_path "docs/capability/design.md"
+    tracked_milestone_paths=()
+    while IFS= read -r -d '' path; do
+        if is_milestone_path "$path"; then
+            tracked_milestone_paths+=("$path")
+        fi
+    done < <(git ls-files -z)
+    if (( ${#tracked_milestone_paths[@]} > 0 )); then
+        echo "tracked paths must use capability names, not milestone names:" >&2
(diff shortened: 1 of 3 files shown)


Datawav:

- **Medium — Naming enforcement ignores directory components (`Justfile`, `naming-gate`).** The new check applies `tolower($NF)` after splitting each tracked path on `/`, so it examines only the basename. A path such as `docs/milestone-5/design.md` passes even though its tracked path retains milestone-oriented naming. This leaves a straightforward gap in the stated goal of enforcing durable, capability-oriented artifact names.

  **Need/design:** Enforcing this at the naming gate is appropriate because it prevents regressions across all tracked artifacts. The implementation should inspect the complete path and remain case-insensitive.

  **Alternatives:** Match `tolower($0)` to prohibit `milestone` anywhere in the tracked path, or inspect every path component if the basename-only behavior is intentional but selected directories must also be governed. Add cases covering directory names, basenames, and mixed capitalization.

_Advisory audit of head [`ded60c5da087`](https://github.com/marmot-protocol/mdk/commit/ded60c5da0871428212cd9cfaab2d6c1e244263e); no approval, merge, push, checkout, or PR-code execution was performed._

Datawav:

## Finding

- **Low — `naming-gate` can silently pass when path enumeration fails** (`Justfile`, new `done < <(git ls-files -z)` loop). Bash does not propagate a process-substitution command’s exit status to the loop, so `set -euo pipefail` does not catch a failing `git ls-files`. The loop may receive no paths and incorrectly report success. Capture the output through a status-checked command or restructure it as a pipeline whose failure is observed; add a regression test for enumeration failure.

## Assessment

- **Need:** A preventive gate is justified because one-off renames do not stop milestone-oriented filenames from returning.
- **Design:** Case-insensitive substring matching is consistent with the PR’s explicit policy to reject tracked filenames containing `milestone`. The inline function probes are valid self-tests; they do not depend on those paths existing.
- **Alternatives:** A standalone checked script would be easier to test. Segment-based matching or an allowlist would weaken or complicate the stated policy and is not presently warranted.
- The documentation and error-message changes match the capability-oriented terminology and introduce no substantiated behavioral issue.

_Advisory audit of head [`ed93ce0b14ba`](https://github.com/marmot-protocol/mdk/commit/ed93ce0b14ba6f4ceea041784aaaff754ce8eb64); no approval, merge, push, checkout, or PR-code execution was performed._

Issue #842, opened 2026-07-13
marmot-app: one unsafe/over-limit relay endpoint in any single group route aborts the entire account's transport activation & sync (one-group → whole-account availability amplification)


erskingardner:
**Severity:** MEDIUM
**Component:** `crates/marmot-app` (`relay_plane`)

## Summary

`RelaySafetyPolicy::sanitize_activation` sanitizes the account inbox route **and every group route** in one pass and returns `Err` on the first endpoint it rejects (non-public/loopback literal host, or a route exceeding `MAX_RELAY_ENDPOINTS_PER_ROUTE = 16`). That `Err` fails `activate_account` wholesale, so a single group whose signed routing component lists one bad relay takes down transport activation for the **inbox plus all other groups** — not just the offending group.

## Where

`crates/marmot-app/src/relay_plane/safety.rs:40-105`

```rust
pub(crate) fn sanitize_activation(&self, mut activation: TransportAccountActivation)
    -> Result<TransportAccountActivation, String> {
    activation.inbox_endpoints = self.sanitize_endpoints(activation.inbox_endpoints, "account inbox")?;
    for group in &mut activation.group_subscriptions {
        group.endpoints = self.sanitize_endpoints(group.endpoints.clone(), "group route")?; // one bad route -> whole Err
    }
    Ok(activation)
}
```

`sanitize_endpoints` returns `Err` on the first `reject_unsafe_relay_host` failure (`:90`) or when `sanitized.len() > self.max_endpoints_per_route` (`:97`).

Call site — `MarmotRelayPlaneAccountAdapter::activate_account` (`relay_plane/mod.rs:751-756`):

```rust
let activation = self.relay_plane.inner.relay_safety
    .sanitize_activation(activation)
    .map_err(TransportAdapterError::Subscription)?;   // fails the whole activation
```

`AppClient::sync()` awaits `activate_transport()` with `?`, so the whole account sync aborts. `sanitize_group_sync` (`:52`) has the same all-or-nothing shape.

## Reachability / amplification

[...]

PR #1274, opened 2026-08-05
Ignore retired relays in discovered account routes

erskingardner:
## Summary

- preserve published NIP-65 and Marmot inbox relay lists exactly as discovered so clients can display and classify retired entries
- drop retired or otherwise unsafe discovered endpoints only when constructing runtime inbox, outbox, KeyPackage, directory, and push routes
- fall back to configured White Noise directory relays when a discovered local or remote KeyPackage route has no usable endpoints, without publishing a replacement relay list
- reject retired endpoints in every relay-list edit role (default, inbox, bootstrap, NIP-65 read, and NIP-65 write)
- keep relay-list declaration validation independent from the runtime dial-route endpoint-count limit

## Root cause

Account setup cached a discovered relay list containing a retired relay and later passed that list unchanged into the fail-closed relay-plane activation boundary. A published `relay.damus.io` or `relay.nostr.band` inbox route therefore aborted activation for both imported nsec identities and external signers.

The denylist remains enforced. This change distinguishes untrusted discovered routing data, which is filtered per endpoint at use time, from caller-configured relay-list edits, which remain fail-closed. If a published list becomes empty after filtering, runtime routing uses configured fallback relays; the published list itself is neither rewritten nor replaced.

This addresses the retired-relay login/signup failure tracked in #842. Signed group routes remain fail-closed and are intentionally unchanged by this account-route fix.

## User impact

Existing identities can sign in even when their published relay lists contain retired relays. Remote KeyPackage lookup also continues through configured directory relays when a contact's published NIP-65 list contains only retired entries. The original lists remain visible through the bindings, and `classify_relay_endpoints` continues to label those entries as `retired` for relay settings UI.

## Validation

- `just fast-ci`
- `cargo test -p marmot-app --test relay_runtime import_ignores_retired_published_routes_without_rewriting_relay_lists -- --exact`
- `cargo test -p marmot-app --test relay_runtime external_signer_login_ignores_retired_routes_without_rewriting_relay_lists -- --exact`
- `cargo test -p marmot-app --test relay_runtime remote_key_package_fetch_falls_back_when_published_outbox_is_retired -- --exact`
- `cargo test -p marmot-app --test relay_runtime relay_list_edits_reject_retired_endpoints_in_every_input_role -- --exact`
- `cargo test -p marmot-app tests::relay_list_declaration_validation_does_not_apply_the_dial_route_cap -- --exact`
- existing relay-safety unit regression for both retired hosts

A broader `cargo test -p marmot-app` run passed all 561 unit tests and 79 relay-runtime tests before five unrelated media/retention/removal integration cases failed; two of those failures also reproduce when run individually.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved relay safety checks across notifications, inbox and outbox routing, key-package discovery, and account relay-list operations.
  * Added fallback routing when configured relay endpoints are no longer safe or available.
  * Prevented retired relay endpoints from being added through relay-list edits.
  * Preserved existing published relay lists during imports and external-signer logins, even when routes are retired.

### DIFF
--- a/crates/marmot-app/src/directory/methods.rs
+++ b/crates/marmot-app/src/directory/methods.rs
@@ -262,19 +262,20 @@ impl MarmotApp {
             }
         }
         self.remember_directory_relay_lists(account_id_hex, &relay_lists)?;
-        if relay_lists.nip65.relays.is_empty() {
+        let source_relays = self.retain_safe_discovered_endpoints(
+            relay_lists
+                .nip65
+                .relays
+                .iter()
+                .cloned()
+                .map(TransportEndpoint)
+                .collect(),
+        );
+        if source_relays.is_empty() {
             return Err(AppError::MissingRelayLists(vec![
                 MissingRelayListKind::Nip65,
             ]));
         }
-
-        let source_relays = relay_lists
-            .nip65
-            .relays
-            .iter()
-            .cloned()
-            .map(TransportEndpoint)
-            .collect::<Vec<_>>();
         let records = self
             .fetch_key_package_events_for_account_id(account_id_hex, &source_relays)
             .await?;
@@ -630,7 +631,7 @@ impl MarmotApp {
         endpoints: Vec<TransportEndpoint>,
     ) -> Vec<TransportEndpoint> {
         self.relay_plane
-            .retain_safe_discovered_endpoints(endpoints, "directory write-relay discovery")
+            .retain_safe_discovered_endpoints(endpoints, "published relay-list route")
     }
 
     pub(crate) fn directory_source_relays(
--- a/crates/marmot-app/src/directory/methods.rs
+++ b/crates/marmot-app/src/directory/methods.rs
@@ -262,7 +262,7 @@ impl MarmotApp {
             }
         }
         self.remember_directory_relay_lists(account_id_hex, &relay_lists)?;
-        let source_relays = self.retain_safe_discovered_endpoints(
+        let mut source_relays = self.retain_safe_discovered_endpoints(
             relay_lists
                 .nip65
                 .relays
@@ -270,7 +270,11 @@ impl MarmotApp {
                 .cloned()
                 .map(TransportEndpoint)
                 .collect(),
+            "key package directory fetch",
         );
+        if source_relays.is_empty() {
+            source_relays = self.directory_source_relays(&[]);
+        }
         if source_relays.is_empty() {
             return Err(AppError::MissingRelayLists(vec![
                 MissingRelayListKind::Nip65,
@@ -620,8 +624,10 @@ impl MarmotApp {
         &self.config.directory_search_fallback_seeds
     }
 
-    /// Narrow relay endpoints discovered from another account's published
-    /// relay list to the ones this device is willing to dial.
+    /// Narrow network-sourced relay-list endpoints to the ones this device is
+    /// willing to dial, whether the list belongs to this device's local account
+    /// or to another account. Published data remains untrusted in both cases;
+    /// filtering affects only the operation's route and never rewrites the list.
     ///
     /// Routes to the same host-safety rule configured endpoints face; see
     /// [`RelaySafetyPolicy::retain_safe_endpoints`] for why a published list
@@ -629,9 +635,10 @@ impl MarmotApp {
     pub(crate) fn retain_safe_discovered_endpoints(
         &self,
         endpoints: Vec<TransportEndpoint>,
+        context: &str,
     ) -> Vec<TransportEndpoint> {
         self.relay_plane
-            .retain_safe_discovered_endpoints(endpoints, "published relay-list route")
+            .retain_safe_discovered_endpoints(endpoints, context)
     }
 
     pub(crate) fn directory_source_relays(
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -1665,6 +1665,32 @@ impl MarmotApp {
         if bootstrap.default_relays.is_empty() && !has_directional_nip65_relays {
             return Err(AppError::MissingDefaultRelays);
         }
+        self.relay_plane
+            .sanitize_relay_endpoints(
+                bootstrap.default_relays.clone(),
+                "account relay-list declaration",
+            )
+            .map_err(AppError::RelayDirectory)?;
+        self.relay_plane
+            .sanitize_relay_endpoints(
+                bootstrap.bootstrap_relays.clone(),
+                "account relay-list publication",
+            )
+            .map_err(AppError::RelayDirectory)?;
+        if let Some(relays) = nip65_relay_set {
+            self.relay_plane
+                .sanitize_relay_endpoints(
+                    relays.read_relays.clone(),
+                    "account NIP-65 read-relay declaration",
+                )
+                .map_err(AppError::RelayDirectory)?;
+            self.relay_plane
+                .sanitize_relay_endpoints(
+                    relays.write_relays.clone(),
+                    "account NIP-65 write-relay declaration",
+                )
+                .map_err(AppError::RelayDirectory)?;
+        }
         let account = self.account_home().account(label)?;
         let signer = self.account_signer_for_summary(&account)?;
         let account_id = MemberId::new(hex::decode(&account.account_id_hex)?);
@@ -1715,11 +1741,13 @@ impl MarmotApp {
             .await
     }
 
-    /// Outbox routing for account-scoped events. Prefers the account's own
-    /// declared NIP-65 write relays (read from the local relay-list cache, no
-    /// network), so e.g. republishing your relay lists / profile goes to *your*
-    /// relays rather than whatever defaults the caller passed. Falls back to
-    /// `fallback` only when the account has no NIP-65 list yet (cold start).
+    /// Outbox routing for account-scoped events. Prefers the safe subset of the
+    /// account's declared NIP-65 write relays (read from the local relay-list
+    /// cache, no network), so e.g. republishing your relay lists / profile goes
+    /// to *your* relays rather than whatever defaults the caller passed. Falls
+    /// back to `fallback` when the account has no usable NIP-65 relay. Filtering
+    /// here affects only the operation's route; it does not rewrite the cached
+    /// or published relay list.
     fn outbox_endpoints(
         &self,
         account_id_hex: &str,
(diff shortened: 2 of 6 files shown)


crates/marmot-app/src/directory/methods.rs
```diff
@@ -262,19 +262,20 @@ impl MarmotApp {
...
+        let source_relays = self.retain_safe_discovered_endpoints(
+            relay_lists
+                .nip65
+                .relays
+                .iter()
+                .cloned()
+                .map(TransportEndpoint)
+                .collect(),
+        );
+        if source_relays.is_empty() {
```
erskingardner, line 278: This is the asymmetry: for the *local* account, `key_package_endpoints`/`account_inbox_endpoints` fall back to `self.relay_endpoints()` when nothing survives filtering. For a *remote* account, an all-retired NIP-65 list dead-ends here.

Concretely: `member_key_package` (lib.rs:3590) now skips its fetch when the contact's filtered NIP-65 is empty and falls through to this function with an empty bootstrap; line 256 above tests the *raw* list for emptiness so the discovery-relay refetch doesn't fire either, and we land on this `return Err`. Net effect for the very population this PR is about — someone publishing `relay.damus.io` — is that you can sign in but still cannot fetch their KeyPackage or invite them.

The symmetric fix is small:

```rust
let mut source_relays = self.retain_safe_discovered_endpoints(/* ... */);
if source_relays.is_empty() {
    source_relays = self.directory_source_relays(&[]);
}
if source_relays.is_empty() {
    return Err(AppError::MissingRelayLists(vec![MissingRelayListKind::Nip65]));
}
```

If you'd rather keep that out of scope, the error is at least misleading as-is — the account *has* a NIP-65 list, it's just unusable, and a client that reacts to `MissingRelayLists` by prompting to publish one is being told the wrong thing. Same applies to the identical block in lib.rs:3276 (`account_key_package_records`).

crates/marmot-app/src/directory/methods.rs
```diff
@@ -630,7 +631,7 @@ impl MarmotApp {
         endpoints: Vec<TransportEndpoint>,
     ) -> Vec<TransportEndpoint> {
         self.relay_plane
-            .retain_safe_discovered_endpoints(endpoints, "directory write-relay discovery")
+            .retain_safe_discovered_endpoints(endpoints, "published relay-list route")
```
erskingardner: Two things about this helper now that it has eight callers instead of two:

1. The doc comment above ("discovered from **another account's** published relay list") is now wrong for half the call sites — `outbox_endpoints`, `account_inbox_endpoints`, and `key_package_endpoints` apply it to the *local* account's own published lists. That reclassification is the load-bearing decision in this PR (your own published list is network-sourced, therefore untrusted, therefore filtered rather than fail-closed), so it deserves to be stated in the doc rather than inferred from the call sites.

2. Hardcoding a single `"published relay-list route"` context collapses the diagnostic. `retain_safe_endpoints` logs `context` alongside the offered/kept counts precisely so you can tell *which* route dropped endpoints; with one string shared across inbox routing, outbox routing, key-package discovery, notification triggers, and directory search, that debug line can no longer answer the first question you'd ask during the next incident. The relay-plane method already takes a `&str` — consider threading it through:

```rust
pub(crate) fn retain_safe_discovered_endpoints(
    &self,
    endpoints: Vec<TransportEndpoint>,
    context: &str,
) -> Vec<TransportEndpoint>
```

crates/marmot-app/src/lib.rs
```diff
@@ -3910,13 +3943,18 @@ impl MarmotApp {
...
+            let safe = self.retain_safe_discovered_endpoints(
+                relay_lists
+                    .inbox
+                    .relays
+                    .iter()
+                    .cloned()
+                    .map(TransportEndpoint)
+                    .collect(),
+            );
+            if !safe.is_empty() {
```
erskingardner: When the published inbox list is entirely retired we silently subscribe on the configured defaults instead. Meanwhile senders resolve *our published* inbox, filter it to empty in `routing_for`, and get `MissingInboxRoute` — so the account is reachable by nobody, yet locally everything looks healthy.

That's still better than aborting activation, but this is the one spot where the degradation is invisible. Worth a privacy-safe `tracing::warn!` (aggregate counts only, per the observability invariant) or, better, a flag on `AccountRelayListStatus` so the relay-settings UI can prompt "your published inbox relay is retired, pick another". Without it the user's only path to discovery is happening to open relay settings and noticing the `retired` classification.

Same reasoning applies to `key_package_endpoints` just below.

crates/marmot-app/src/lib.rs
```diff
@@ -1665,6 +1665,32 @@ impl MarmotApp {
         if bootstrap.default_relays.is_empty() && !has_directional_nip65_relays {
             return Err(AppError::MissingDefaultRelays);
         }
+        self.relay_plane
```
erskingardner: Four near-identical 6-line blocks, each cloning a `Vec` only to discard the sanitized result. Reads as one validation step, so it could be written as one:

```rust
let declarations: [(&[TransportEndpoint], &str); _] = [
    (&bootstrap.default_relays, "account relay-list declaration"),
    (&bootstrap.bootstrap_relays, "account relay-list publication"),
    // nip65 read/write when present
];
for (relays, context) in declarations {
    self.relay_plane
        .sanitize_relay_endpoints(relays.to_vec(), context)
        .map_err(AppError::RelayDirectory)?;
}
```

Placement is right — validate-before-mutate, ahead of any publish, per the multi-step-state-changes invariant.

Note also that `sanitize_endpoints` enforces `max_endpoints_per_route` (16), so this newly rejects a *declared* NIP-65 list of more than 16 relays even though nothing dials the declared set. Probably fine, but it's a behavior change beyond "reject retired endpoints" and isn't mentioned in the PR description.

Coverage gap while you're here: `relay_list_edits_reject_retired_endpoints` only exercises the inbox path (`bootstrap.default_relays`). The three other new checks — `bootstrap_relays`, and the NIP-65 read/write declarations reached via `publish_account_nip65_relay_set` — are untested. The read/write pair is the one I'd most want a test for, since `set_account_nip65_relays` reaches it through `nip65_relay_set_preserving_roles`.

crates/marmot-app/src/lib.rs
```diff
@@ -2771,17 +2797,20 @@ impl MarmotApp {
...
+            let endpoints = self.retain_safe_discovered_endpoints(
+                entry
+                    .relay_lists
+                    .inbox
+                    .relays
+                    .into_iter()
+                    .map(TransportEndpoint)
+                    .collect(),
+            );
+            if !endpoints.is_empty() {
```
erskingardner, line 2833: Nit: the `if entry.relay_lists.inbox.relays.is_empty() { continue; }` guard a few lines up is now dead — `retain_safe_discovered_endpoints([])` returns `[]`, so this check already covers it. Dropping the guard removes three lines and one branch.

Same redundant outer `is_empty()` guard at lib.rs:3590 (`member_key_package`), 3945 (`account_inbox_endpoints`), and 3972 (`key_package_endpoints`) — flattening those also un-nests the bodies by a level, which is most of what makes the 3590 hunk read as a bigger change than it is.

crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -407,6 +407,79 @@ fn endpoint(url: &str) -> TransportEndpoint {
     TransportEndpoint(url.to_owned())
 }
 
+#[derive(Clone, Debug)]
```
erskingardner, line 410: This is a verbatim 72-line copy of `TestExternalAccountSigner` from `crates/marmot-app/src/tests.rs:2172`. I realize integration tests can't reach into the private `tests` module, so there's no free fix — but two copies of a signer shim that has to stay in sync with the `NostrSigner` trait is a maintenance seam. Options in rough order of cost: a `mod support;` shared across `tests/*.rs`, or a `#[cfg(feature = "test-support")]` export from the crate. At minimum a `// mirrors src/tests.rs::TestExternalAccountSigner` breadcrumb so the next person adding a trait method finds both.

Separately, worth asking whether `external_signer_login_ignores_retired_routes_without_rewriting_relay_lists` earns its ~130 lines. Its assertions are identical to the import test's; the only delta is `login_external_signer` vs `create_or_import_account`, and both converge on the same `account_inbox_endpoints`/`key_package_endpoints` code under review. If the external-signer path was a distinct reported failure mode, say so in a comment on the test so nobody deletes it as a duplicate later.

erskingardner (commented):
Premise is sound and the root-cause framing is right: the failure was feeding a *discovered* relay list into the fail-closed dial-safety boundary, and splitting "discovered routing data (filter per endpoint)" from "caller-configured declarations (fail closed)" is the correct seam. `sanitize_activation` still enforces the rule at the chokepoint, the published lists are demonstrably not rewritten, and the two hunks that flatten now-dead `is_empty()` wrappers are net-negative lines. No correctness regression found.

Two substantive gaps and some cleanliness notes, none of which I'd block on:

1. **The fix is half of the incident class.** Local account paths now fall back to the configured relays; the *remote* contact paths do not. A contact whose published NIP-65 is only `relay.damus.io` still hard-fails `fetch_latest_key_package_for_account_id` with `MissingRelayLists(Nip65)`, so you can sign in but you still can't invite them. Inline below.
2. **Login succeeds but the account can still be unreachable, with no signal.** When the published kind-10050 inbox is entirely retired, `account_inbox_endpoints` subscribes on the configured defaults while senders (correctly) resolve the published inbox and now hit `MissingInboxRoute`. That's a strictly better outcome than aborting activation, but nothing marks the account degraded or nudges the user to fix their relay list, and `classify_relay_endpoints` only helps if the client happens to open relay settings.

Also worth confirming the relationship to #842 — this fixes the account-inbox arm of that amplification bug but deliberately leaves the group-route arm fail-closed (correct, since group routes come from signed routing components). Linking it would help whoever picks up the rest.

One more, not inline because it isn't in the diff: `account_nip65_relays()` (lib.rs:1554) returns the *union* including retired entries specifically so the getter -> edit -> setter flow doesn't delete them. With the new fail-closed check in `publish_selected_account_relay_lists_with_nip65`, that round-trip now hard-errors on an entry the user never typed — a user with `relay.damus.io` in their published list cannot save *any* relay-list edit until they manually delete it, and `AppError::RelayDirectory("...: relay endpoint host is retired")` doesn't say which entry. Either name the offending entries in the error or document that clients must run `classify_relay_endpoints` and drop non-`Allowed` entries before calling the setter.

Validation: relying on the stated `just fast-ci` + targeted test runs; I did not execute anything locally.

Datawav:

⚠️ Correct the misleading KeyPackage routing comment in `crates/marmot-app/src/lib.rs`.

[P3] The comment near `account_key_package_records` says, “We never normally pull KeyPackage events from the account's own NIP-65 relays.” This contradicts both the implementation and the following lines: NIP-65 relays remain the normal source, with configured directory relays used only when the published route has no safe endpoints. Change “never normally” to “normally” and describe the safety-filtered fallback accurately. Misstating this security-sensitive routing contract could cause future maintenance errors.

Need: Retired discovered relays must not prevent account activation or KeyPackage lookup, while published lists remain visible and caller-supplied edits remain fail-closed.

Design: The patch preserves published data, filters discovered endpoints only when constructing runtime routes, and uses configured fallbacks when filtering empties a route. Declaration validation remains independent of the dial-route endpoint limit.

Alternatives: Filtering or rewriting lists at ingestion would lose published-state fidelity; rejecting an entire discovered list would preserve the reported outage. The selected use-time filtering design is appropriate aside from the inaccurate comment.

PR #1280, opened 2026-08-06
Make account import setup resumable

erskingardner:
## Summary

- add a durable account-setup journal that distinguishes local creation, KeyPackage publication start, and confirmed publication
- resume the exact persisted KeyPackage/private material after relay failure, cancellation, or restart instead of minting a new stable slot
- clean root-level KeyPackage compatibility artifacts and cached account handles during safe rollback
- add typed, consent-gated recovery APIs for pre-journal installations whose prior KeyPackage exposure cannot be proven
- expose the recovery state and one-call same-`nsec` retry through UniFFI

## Root cause

Account setup could create the encrypted account database and strict-cutover replacement marker before setup committed. The marker lives outside the account directory, so ordinary rollback did not remove it, while cancellation and shutdown could bypass Result-based rollback entirely.

A later import of the same `nsec` therefore saw an existing database without recoverable stable-slot evidence and classified the partial setup as a legacy device whose kind-30443 `d` slot had been lost. Attempts to republish also rejected a durable pending replacement instead of resuming it.

## Compatibility and safety

New failures retain an explicit setup phase plus the stable device-local slot, private bundle, fanout state, and exact signed publication bytes. Retries resume those bytes.

Pre-journal accounts with a durable pending replacement are adopted into the journal and resume automatically. The older ambiguous shape with no lifecycle, legacy record, or private bundle returns `AccountSetupRecoveryRequired`; hosts can obtain explicit consent and call `login_recovering_incomplete_setup`.

The recovery path does not blindly mint a replacement slot, delete another device's KeyPackage, or discard ambiguously exposed private material.

## Validation

- `cargo test -p marmot-app --lib -- --test-threads=1` (564 passed)
- relay rejection, legacy pre-journal restart, and same-`nsec` recovery regression
- cancellation during publication followed by exact pending-attempt resume
- `just fast-ci`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added durable account-setup progress tracking and restart-safe setup recovery.
  * Added consent-gated options to reset or resume incomplete account setup.
  * Improved key-package publication retries after interruptions or failures.
  * Added clear recovery-required errors and diagnostics across supported interfaces.

* **Bug Fixes**
  * Prevented account and credential cleanup from leaving inconsistent partial state.
  * Improved handling of legacy incomplete account imports and retries.
  * Preserved pending setup work across restarts and cancellations.

### DIFF
--- a/crates/cgka-conformance-simulator/src/app_runtime.rs
+++ b/crates/cgka-conformance-simulator/src/app_runtime.rs
@@ -1078,6 +1078,7 @@ fn app_error(error: AppError) -> SubjectError {
         | AppError::ExternalSignerRejected
         | AppError::MissingDirectoryEntry(_)
         | AppError::NotificationsDisabled
+        | AppError::AccountSetupRecoveryRequired
         | AppError::ReactionNotFound => SubjectFailureCategory::ExpectedRefusal,
         AppError::Account(_)
         | AppError::AccountHome(_)
--- a/crates/cgka-conformance-simulator/src/app_runtime.rs
+++ b/crates/cgka-conformance-simulator/src/app_runtime.rs
@@ -1079,6 +1079,9 @@ fn app_error(error: AppError) -> SubjectError {
         | AppError::MissingDirectoryEntry(_)
         | AppError::NotificationsDisabled
         | AppError::AccountSetupRecoveryRequired
+        | AppError::AccountSetupRetryRequired
+        | AppError::AccountSetupResetNotApplicable
+        | AppError::AccountSetupKeyPackageRecoveryAvailable
         | AppError::ReactionNotFound => SubjectFailureCategory::ExpectedRefusal,
         AppError::Account(_)
         | AppError::AccountHome(_)
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -24,6 +24,9 @@ versioning through the workspace version in the root `Cargo.toml`.
   
 ### Fixed
 
+- Account setup interruptions now have stable JSON recovery codes and repair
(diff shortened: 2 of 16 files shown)


Datawav:

## Finding

- **[P2] Setup journal creation is not crash-atomic with account creation** — `crates/marmot-account/src/home.rs` around the `import_nostr_account_idempotent` hunk, and `crates/marmot-app/src/runtime/mod.rs` in `create_nostr_account_from_setup`. Each flow writes the account record—and, for local signing, its secret—before writing `.account-setup.json`. A process failure in that interval leaves an existing account without a journal or `session.sqlite`. Subsequent same-identity setup returns `AccountExists`; `legacy_incomplete_account_for_import` cannot classify it for recovery because it requires account storage to exist. The new consent-gated recovery API likewise only handles the database-backed legacy shape. This creates a permanently stranded setup through the public APIs, contrary to the restart-safe objective. Create the account and initial journal as one atomic operation, or explicitly recognize and safely adopt/reset the narrowly defined “account exists, no journal, no database” state. Add fault-injection coverage after account-record creation and after secret persistence.

The change addresses a real need: interrupted publication can leave durable state outside the account directory and lose the stable KeyPackage slot. Persisting setup phase and exact pending publication material is an appropriate conservative design, and the patch correctly avoids automatically discarding potentially exposed material.

The principal alternative is a transactional account-creation primitive that publishes the record, credential, and journal together. If storage constraints prevent that, a recovery classifier for pre-database partial accounts is acceptable, provided it verifies identity ownership and rejects any evidence of KeyPackage exposure. The advisory claims about the confirmed phase, cleanup style, duplication, and match-arm readability are not actionable correctness defects based on the supplied patches.

PR #1281, opened 2026-08-06
fix: fail when the control socket link disappears

### LINKED ISSUE #1279: wn-agent: orphaned listener — control socket path never linked, clients get Connection refused
## Summary

`wn-agent` can end up **alive but unreachable**: the process keeps running and holds a bound Unix listener, but the final control-socket path (`$MARMOT_HOME/dev/wn-agent.sock`) does not exist on disk, so every client gets `Connection refused`. The blocking supervisor pattern (`Popen` + `child.wait()`) cannot detect this because the child never exits.

## Environment

- wn-agent 0.9.10 (`release_tag: wn-agent-v0.9.10`, source_sha `0f64a56f26a5ad1e4fcb493ca6dd199391bfbd2a`)
- Linux 7.1.3+deb13-amd64 (LXC container, Debian 13), uid 1000
- Socket path on a mapper volume (`/opt/data`); `hard_link` verified to work on this filesystem.

## Observed state (2026-08-06, ~05:47)

- Process running 1.5h+, 13 tokio threads, `futex_do_wait`, ~50% CPU.
- `ls /proc/<pid>/fd` shows socket fds.
- `/proc/<pid>/net/unix` shows the listener bound at `.../dev/.sock.<pid>.wn-agent.sock/wn-agent.sock` (the **staging** path).
- The staging directory had already been removed from disk, and the **final** path `dev/wn-agent.sock` did **not** exist.

So the listener was bound to an unlinked inode inside a deleted staging dir, and the `hard_link(staged, final)` either never executed or its result was lost.

## Root cause hypothesis

`bind_unix_listener_private` (`crates/fs-private/src/lib.rs`, ~lines 506-547):

1. creates staging dir `.sock.<pid>.<name>` (0700)
2. `UnixListener::bind(staged_socket)`
3. `set_permissions(staged, mode)`
4. `hard_link(staged_socket, final_path)` — fails with `AddrInUse` if final exists
5. **unconditionally** `remove_file(staged_socket)` + `remove_dir(staging_dir)`

If step 4 errors (or the process is SIGKILLed/SIGBUS between bind and link, or a concurrent cleanup removes the final path), the code can leave a live listener whose only filesystem reference is gone. Because `serve_socket` binds the listener *before* `AgentConnector::open`/`start()` (which then blocks on relay work), a wedged post-bind state leaves the socket path permanently absent while the process looks alive.

## Supervision gap

The local supervisor (`Popen(child); child.wait()`) only reaps on exit, so an orphaned child is never restarted. A watchdog that checks `socket.connect()` sees the path missing, but sees the supervisor alive and does nothing — the connector stays down until manual intervention.

## Suggested fixes

- Make the bind atomic and observable: if `hard_link` fails or the final path is absent after bind, log and exit non-zero (crash-fast) instead of serving on an unlinked socket.
- After `start()` completes, assert the final socket path exists and is connectable; if not, return a startup error.
- Consider rename (not hard-link) or binding directly at the final path with mode set before listen.

## Related crash evidence

The supervisor log also shows the child exiting with `code -7` (SIGBUS) during the nightly restart window — a separate potential crash worth investigating; happy to capture a core/backtrace if useful.

## Local mitigation (deployed)

The Hermes supervisor now polls child readiness every 5s and kills a child whose socket never becomes connectable within 60s; the watchdog kills orphaned children and uses flock-based supervisor liveness. These recover automatically, but the connector should not enter this state in the first place.

agent-p1p:
Fixes #1279

## Summary
- preserve the published Unix socket device/inode identity during private staged binding without changing the existing public bind API
- verify that `wn-agent serve` control-socket path still names the bound listener during startup and while accepting connections
- exit with a structured error if the final socket path is removed or replaced, instead of retaining an unreachable orphan listener
- add unlink/replacement regression coverage and an `Unreleased` changelog entry

## Verification
- `cargo fmt --all --check`
- `cargo test -p fs-private --locked` (45 passed)
- `cargo test -p agent-connector --locked connector_serve_fails_fast_when_control_socket_link_removed` (passed)
- `cargo test -p agent-connector --locked` (133 unit tests + 1 integration test passed)
- `just fast-ci`
- independent Composer review: passed; no security concerns or blocking logic errors

## Sensitive paths
- `crates/agent-connector/src/lib.rs`
- `crates/agent-connector/src/socket.rs`
- `crates/agent-connector/src/tests.rs`
- `crates/fs-private/src/lib.rs`

Sensitive socket/service lifecycle code changed. Explicit human approval is required at the merge gate.

## Summary by CodeRabbit

- **Bug Fixes**
  - `wn-agent serve` now detects when its control-socket path is removed or replaced and exits promptly with an error.
  - Improved socket validation helps prevent communication through an unexpected socket.

- **Documentation**
  - Added changelog information describing the updated control-socket behavior.

- **Tests**
  - Added coverage for control-socket paths that are removed or replaced.

## Scope

This PR hardens the supervised, long-lived `wn-agent` control socket. The standalone `wnd`/CLI daemon also uses `bind_unix_listener_private`, but its lifecycle and shutdown path are independent and are intentionally left out of this change rather than broadening a security-sensitive patch. It should receive the same link-loss supervision in a focused follow-up.

### DIFF
--- a/crates/agent-connector/src/lib.rs
+++ b/crates/agent-connector/src/lib.rs
@@ -333,14 +333,16 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
     let socket_path = config.socket.clone();
     let max_connections = config.max_connections;
     let connector = AgentConnector::open(config)?;
+    let link_loss = wait_control_socket_link_loss(&socket_path, link_identity);
+    tokio::pin!(link_loss);
     tokio::select! {
         res = connector.start() => res?,
-        res = wait_control_socket_link_loss(&socket_path, link_identity) => res?,
+        res = &mut link_loss => res?,
     }
     let connection_limiter = Arc::new(Semaphore::new(max_connections));
     loop {
         let accepted = tokio::select! {
-            res = wait_control_socket_link_loss(&socket_path, link_identity) => {
+            res = &mut link_loss => {
                 return res;
             }
             accepted = listener.accept() => accepted,
--- a/crates/agent-connector/src/lib.rs
+++ b/crates/agent-connector/src/lib.rs
@@ -335,15 +335,20 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
     let connector = AgentConnector::open(config)?;
     let link_loss = wait_control_socket_link_loss(&socket_path, link_identity);
     tokio::pin!(link_loss);
+    // Losing the published path makes this process unreachable, so cancellation
+    // of startup is intentional: it has the same failure semantics as the
+    // supervisor terminating an orphaned process. Completed startup writes are
+    // retained and the next start rechecks them before applying missing state.
+    // Revisit this contract before adding a non-restart-safe mutation to start().
     tokio::select! {
         res = connector.start() => res?,
-        res = &mut link_loss => res?,
+        err = &mut link_loss => return Err(err),
     }
     let connection_limiter = Arc::new(Semaphore::new(max_connections));
     loop {
         let accepted = tokio::select! {
-            res = &mut link_loss => {
-                return res;
+            err = &mut link_loss => {
+                return Err(err);
             }
             accepted = listener.accept() => accepted,
         };
@@ -415,9 +420,9 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
 async fn wait_control_socket_link_loss(
     socket_path: &Path,
     link_identity: fs_private::UnixSocketInode,
-) -> Result<(), ConnectorError> {
+) -> ConnectorError {
     let mut ticker = tokio::time::interval(CONTROL_SOCKET_LINK_POLL_INTERVAL);
-    ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+    ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
     loop {
         ticker.tick().await;
         if let Err(err) = verify_control_socket_link(socket_path, link_identity) {
@@ -427,7 +432,7 @@ async fn wait_control_socket_link_loss(
                 error_code = "control_socket_link_lost",
                 "control socket path no longer names the bound listener"
             );
-            return Err(err);
+            return err;
         }
     }
 }
--- a/crates/agent-connector/src/lib.rs
+++ b/crates/agent-connector/src/lib.rs
@@ -37,7 +37,7 @@ pub use identity::{
 pub use socket::{bind_connector_socket, bind_connector_socket_with_mode, default_socket_path};
 
 use std::future::Future;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex, MutexGuard};
 use std::time::Duration;
@@ -55,7 +55,7 @@ pub(crate) use marmot_app::AppMessageQuery;
 
 use crate::allowlist::AllowlistStore;
 use crate::event_projection::InboundCatchUpDriver;
-use crate::socket::bind_connector_socket_with_owned_home;
+use crate::socket::{bind_connector_socket_with_owned_home, verify_control_socket_link};
 use crate::stream_session::{DebugFinalSendStore, SendIdempotencyStore, StreamSessionStore};
 use crate::validation::{endpoint, validate_control_plane_config};
 
@@ -106,6 +106,9 @@ pub const MAX_CONTROL_CONNECTIONS: usize = 64;
 /// Long-lived inbound subscriptions are capped separately so one-shot sends,
 /// stream finalization, and policy operations retain connection capacity.
 pub(crate) const MAX_CONTROL_SUBSCRIPTIONS: usize = 16;
+/// Poll interval while `AgentConnector::start` runs so a removed or replaced
+/// control-socket path is detected before the accept loop begins.
+pub(crate) const CONTROL_SOCKET_LINK_POLL_INTERVAL: Duration = Duration::from_millis(250);
 
 pub(crate) async fn with_control_operation_timeout<T, F>(
     operation: &'static str,
@@ -321,18 +324,28 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
             "DANGER: dev allow-any invite policy is enabled; authenticated welcomers bypass the allowlist"
         );
     }
-    let listener = bind_connector_socket_with_owned_home(
+    let (listener, link_identity) = bind_connector_socket_with_owned_home(
         &config.socket,
         config.socket_dir_mode,
         config.socket_mode,
         Some(&config.home),
     )?;
+    let socket_path = config.socket.clone();
     let max_connections = config.max_connections;
     let connector = AgentConnector::open(config)?;
-    connector.start().await?;
+    tokio::select! {
+        res = connector.start() => res?,
+        res = wait_control_socket_link_loss(&socket_path, link_identity) => res?,
+    }
     let connection_limiter = Arc::new(Semaphore::new(max_connections));
     loop {
-        let (mut stream, _peer_addr) = match listener.accept().await {
+        let accepted = tokio::select! {
+            res = wait_control_socket_link_loss(&socket_path, link_identity) => {
+                return res;
+            }
+            accepted = listener.accept() => accepted,
+        };
+        let (mut stream, _peer_addr) = match accepted {
             Ok(accepted) => accepted,
             Err(err) => {
                 let connection_error =
@@ -394,5 +407,25 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
                 );
             }
         });
+    }
+}
+
+async fn wait_control_socket_link_loss(
+    socket_path: &Path,
+    link_identity: fs_private::UnixSocketInode,
+) -> Result<(), ConnectorError> {
+    let mut ticker = tokio::time::interval(CONTROL_SOCKET_LINK_POLL_INTERVAL);
+    ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+    loop {
+        ticker.tick().await;
+        if let Err(err) = verify_control_socket_link(socket_path, link_identity) {
+            tracing::error!(
+                target: "agent_connector",
+                method = "serve_socket",
+                error_code = "control_socket_link_lost",
+                "control socket path no longer names the bound listener"
+            );
+            return Err(err);
+        }
     }
 }
--- a/crates/agent-connector/src/tests.rs
+++ b/crates/agent-connector/src/tests.rs
@@ -679,20 +679,59 @@ async fn connector_serve_fails_fast_when_control_socket_link_removed() {
     let socket = dir.path().join("dev").join("wn-agent.sock");
     let config = test_config(dir.path(), socket.clone(), Vec::new(), false, false);
     let server = tokio::spawn(serve_socket(config));
-    tokio::time::timeout(Duration::from_secs(5), async {
-        while !socket.exists() {
-            sleep(Duration::from_millis(25)).await;
-        }
-    })
-    .await
-    .expect("control socket should appear at the final path");
+    let response = send_control_request(
+        &socket,
+        "steady-state-before-unlink",
+        AgentControlRequest::AccountList,
+    )
+    .await;
+    assert!(matches!(
+        response.payload,
+        AgentControlResponse::AccountList { .. }
+    ));
     std::fs::remove_file(&socket).expect("unlink final control socket path");
     let err = tokio::time::timeout(Duration::from_secs(2), server)
         .await
         .expect("serve must fail fast after the control socket link is removed")
         .expect("serve task must not panic")
         .expect_err("serve must exit with an error");
-    assert_eq!(err.code(), "io_error");
+    let crate::ConnectorError::Io(err) = err else {
+        panic!("expected control-socket I/O error");
+    };
+    assert_eq!(err.kind(), std::io::ErrorKind::NotFound);
+}
+
+#[tokio::test]
+async fn connector_serve_fails_fast_when_control_socket_link_replaced() {
+    let dir = tempfile::tempdir().unwrap();
+    let socket = dir.path().join("dev").join("wn-agent.sock");
+    let replacement_path = dir.path().join("dev").join("replacement.sock");
+    let config = test_config(dir.path(), socket.clone(), Vec::new(), false, false);
+    let server = tokio::spawn(serve_socket(config));
+    let response = send_control_request(
+        &socket,
+        "steady-state-before-replacement",
+        AgentControlRequest::AccountList,
+    )
+    .await;
+    assert!(matches!(
+        response.payload,
+        AgentControlResponse::AccountList { .. }
+    ));
+
+    let _replacement = std::os::unix::net::UnixListener::bind(&replacement_path)
+        .expect("bind replacement control socket");
+    std::fs::rename(&replacement_path, &socket)
+        .expect("atomically replace final control socket path");
+    let err = tokio::time::timeout(Duration::from_secs(2), server)
+        .await
+        .expect("serve must fail fast after the control socket link is replaced")
+        .expect("serve task must not panic")
+        .expect_err("serve must exit with an error");
+    let crate::ConnectorError::Io(err) = err else {
+        panic!("expected control-socket I/O error");
+    };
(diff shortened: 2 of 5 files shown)


agent-p1p:
## Pip adversarial review

**Verdict: changes required** — 1 blocking finding, 1 suggestion, 0 nitpicks. Reviewed exact signed head `05bceed53385528a795284b2269890c3eea03a91`.

### Blocking

- **`crates/fs-private/src/lib.rs:1030-1051` — compensate the published final link when post-link validation fails.** After `hard_link(&staged_socket, final_path)` succeeds, either inode lookup, the staged/final identity comparison, or the final `verify_unix_socket_inode` can still return `Err`. Every such path drops the listener and removes only the staging entry; the final path can remain as a stale socket with no listener. That violates the repository torn-write invariant and makes this new validation path manufacture the same stale-path state that callers must recover from. Track the identity of the link created by this bind attempt and, on any later error, remove `final_path` only if it still names that exact device/inode. Never remove a replacement. Add regression coverage for both cleanup of the original entry and preservation of a replacement entry. This independently confirms CodeRabbit discussion `r3727001091`.

### Suggestion

- **`crates/agent-connector/src/lib.rs:336-347` — keep one pinned link-loss monitor for the listener lifetime.** Reconstructing `wait_control_socket_link_loss` on every accept resets its interval and makes the advertised 250 ms cadence dependent on accept-loop scheduling. A single pinned future reused by mutable reference across startup and the accept loop is simpler and preserves one timer state. This confirms CodeRabbit discussion `r3727001071`; I am classifying it as non-blocking because link loss prevents new path-based connections and the old backlog is finite.

### Verification

- `cargo fmt --all --check` — pass
- `cargo test -p fs-private --locked` — pass (45 tests)
- `cargo test -p agent-connector --locked connector_serve_fails_fast_when_control_socket_link_removed` — pass
- `cargo clippy -p fs-private -p agent-connector --all-targets --all-features --locked -- -D warnings` — pass
- GitHub CI and wn-agent binary jobs are green at this head.
- Changelog entry present; no release/version churn.

Filed follow-up #1283 for the out-of-scope SIGBUS observed during nightly restart; it intentionally has no `agent-ok` label pending a core/backtrace.

Sensitive paths touched: `crates/agent-connector/**` control-socket service lifecycle; `crates/fs-private/src/lib.rs` restrictive Unix-socket publication.

crates/fs-private/src/lib.rs
```diff
@@ -961,11 +1037,22 @@ pub fn bind_unix_listener_private(
...
+                "staged socket hard link identity mismatch",
+            ));
+        }
+        Ok((listener, link_identity))
     })();
     let _ = std::fs::remove_file(&staged_socket);
     let _ = std::fs::remove_dir(&staging_dir);
-    outcome
+    let (listener, link_identity) = outcome?;
+    verify_unix_socket_inode(final_path, link_identity)?;
```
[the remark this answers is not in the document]
agent-p1p: Addressed in b7e66538. The staged inode is captured before hard-link publication; every later error compensates by unlinking only when the final path still names that exact inode. Regression tests cover removal of the original entry and preservation of a replacement.

crates/agent-connector/src/lib.rs
```diff
@@ -321,18 +324,28 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
...
+    }
     let connection_limiter = Arc::new(Semaphore::new(max_connections));
     loop {
-        let (mut stream, _peer_addr) = match listener.accept().await {
+        let accepted = tokio::select! {
+            res = wait_control_socket_link_loss(&socket_path, link_identity) => {
+                return res;
+            }
+            accepted = listener.accept() => accepted,
+        };
```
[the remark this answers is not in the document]
agent-p1p: Addressed in b7e66538. `serve_socket` now creates and pins one link-loss future before startup and reuses it by mutable reference through startup and every accept-loop select.

agent-p1p:
@coderabbitai review

Datawav:

## Final audit

The change is needed: a Unix listener can remain alive but unreachable after its published socket path is removed or replaced. Tracking the published device/inode and polling it during startup and acceptance is a reasonable design. Preserving the existing public bind APIs also limits compatibility risk.

### Finding — P2: failure cleanup can unlink a concurrent replacement

`remove_unix_socket_path_if_inode` first reads the path’s inode and then separately calls `remove_file(path)`. Another process can replace the directory entry between those operations:

1. Cleanup verifies that the path names the socket created by this bind.
2. Another process removes it and installs a replacement.
3. Cleanup’s `remove_file` unlinks that replacement.

The new post-publication error paths call this helper precisely where concurrent path manipulation is possible. The replacement-preservation test swaps the socket before verification, so it does not exercise this check-to-unlink race.

This matters because a failed bind attempt must not delete a socket it does not own. Device/inode comparison reduces the race window but cannot make a pathname-based check followed by unlink atomic.

Recommended design: avoid removing the published path after any failure that occurs once publication succeeds, unless directory mutation is serialized under an ownership mechanism shared by all writers. Alternatives include retaining the created link as a stale entry for normal stale-socket recovery, publishing through a uniquely owned directory/lock protocol, or using a platform-specific mechanism that can couple identity validation and removal safely. Add a deterministic test hook between verification and unlink to cover replacement during that interval.

The advisory concerns about inode reuse, connector-level replacement coverage, generic `NotFound` text, MSRV, and cancellation safety are either non-blocking diagnostics/test improvements or unsupported by the supplied patch context.

_Advisory audit of head [`b7e6653870f3`](https://github.com/marmot-protocol/mdk/commit/b7e6653870f3c54b9743079c84622781bef15013); no approval, merge, push, checkout, or PR-code execution was performed._

agent-p1p:
@coderabbitai review

crates/fs-private/src/lib.rs
```diff
@@ -897,6 +897,110 @@ pub fn parse_octal_mode(value: &str) -> Result<u32, String> {
...
+/// Remove `path` when it still names `expected`, leaving concurrent replacements intact.
+#[cfg(unix)]
+fn remove_unix_socket_path_if_inode(path: &Path, expected: UnixSocketInode) {
+    match unix_socket_inode(path) {
+        Ok(actual) if actual == expected => {
+            let _ = std::fs::remove_file(path);
+        }
+        _ => {}
+    }
+}
```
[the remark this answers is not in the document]
agent-p1p: Addressed in 61fd5d0f by removing the racy helper and all fallible validation after successful hard-link publication. The staged inode is validated before `hard_link`; after publication the function only performs best-effort staging cleanup and returns, so there is no pathname check-to-unlink window or unsupported serialization assumption.

agent-p1p:
@Datawav Addressed the P2 TOCTOU finding in 61fd5d0f. The fix deletes inode-check-then-unlink compensation entirely and moves every fallible validation before the atomic `hard_link` publication point. No post-publication error path can now unlink the final path; connector link monitoring still detects any later removal or replacement.

agent-p1p:
@coderabbitai review

crates/agent-connector/src/lib.rs
```diff
@@ -321,18 +324,30 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
...
+    let socket_path = config.socket.clone();
     let max_connections = config.max_connections;
     let connector = AgentConnector::open(config)?;
-    connector.start().await?;
+    let link_loss = wait_control_socket_link_loss(&socket_path, link_identity);
+    tokio::pin!(link_loss);
+    tokio::select! {
+        res = connector.start() => res?,
+        res = &mut link_loss => res?,
+    }
```
erskingardner, line 346: Cancelling `connector.start()` here is a new behavior worth calling out.

`start()` is not a pure readiness check — it runs `ensure_agent_accounts_ready`, which awaits `set_account_nip65_relays`, `set_account_inbox_relays`, and `publish_key_package`. When the link-loss arm wins, this future is dropped at whatever await point it happens to be sitting on, with no compensation. That is the pattern `docs/marmot-architecture/overview/multi-step-state-changes.md` exists to guard against.

In practice the process exits non-zero immediately afterward, so this is arguably no worse than the SIGKILL the supervisor would deliver anyway — but that reasoning is currently implicit. Either:

- confirm `start()` is cancel-safe (no partially-applied storage writes on drop) and note it in a comment here, or
- narrow the window: `start().await?` first, then a single `verify_control_socket_link(...)?`, and let the accept-loop monitor cover steady state.

The second option is simpler and drops a `select!`, but it does give up detection during a long-running `start()` — which is the exact window #1279 reports the daemon wedging in. So if you keep the `select!`, say why in a comment; don't leave the reader to work out that an abrupt cancel is intended.

crates/agent-connector/src/tests.rs
```diff
@@ -673,6 +673,28 @@ async fn connector_socket_bind_rejects_unconfigured_group_writable_parent() {
...
+#[tokio::test]
+async fn connector_serve_fails_fast_when_control_socket_link_removed() {
+    let dir = tempfile::tempdir().unwrap();
+    let socket = dir.path().join("dev").join("wn-agent.sock");
+    let config = test_config(dir.path(), socket.clone(), Vec::new(), false, false);
+    let server = tokio::spawn(serve_socket(config));
+    tokio::time::timeout(Duration::from_secs(5), async {
+        while !socket.exists() {
+            sleep(Duration::from_millis(25)).await;
+        }
```
erskingardner: This unlinks as soon as the path appears — but the path appears at `bind_connector_socket_with_owned_home`, i.e. *before* `connector.start()` runs. With this `test_config` (no relays) `start()` is fast, so which of the two new detection paths trips is a race, and the likely winner is the startup `select!`.

That leaves the accept-loop monitor — the one that matters for a daemon that has been up for 1.5h, which is the scenario in #1279 — without deterministic coverage.

Suggest driving the connector to steady state before unlinking: connect once and get a successful control response (or otherwise observe the accept loop running), *then* `remove_file`. A second case that replaces rather than removes the path would also mirror the `fs-private` unit tests at the connector level.

crates/agent-connector/src/tests.rs
```diff
@@ -673,6 +673,28 @@ async fn connector_socket_bind_rejects_unconfigured_group_writable_parent() {
...
+    })
+    .await
+    .expect("control socket should appear at the final path");
+    std::fs::remove_file(&socket).expect("unlink final control socket path");
+    let err = tokio::time::timeout(Duration::from_secs(2), server)
+        .await
+        .expect("serve must fail fast after the control socket link is removed")
+        .expect("serve task must not panic")
+        .expect_err("serve must exit with an error");
+    assert_eq!(err.code(), "io_error");
```
erskingardner: `io_error` is the code for every `ConnectorError::Io`, so this assertion passes for any I/O failure anywhere in `serve_socket` — including failures that have nothing to do with the link being lost. Consider matching on `ConnectorError::Io(err)` and asserting `err.kind() == ErrorKind::NotFound` (and `InvalidInput` for the replacement case) so the test can't pass for the wrong reason.

crates/fs-private/src/lib.rs
```diff
@@ -897,6 +897,72 @@ pub fn parse_octal_mode(value: &str) -> Result<u32, String> {
...
+/// `expected` (for example after an external unlink or replacement).
+#[cfg(unix)]
+pub fn verify_unix_socket_inode(path: &Path, expected: UnixSocketInode) -> io::Result<()> {
+    const LINK_LOST: &str = "unix socket path no longer names the bound listener";
+
+    let actual = match unix_socket_inode(path) {
+        Ok(identity) => identity,
+        Err(err) if err.kind() == io::ErrorKind::InvalidInput => {
+            return Err(io::Error::new(io::ErrorKind::InvalidInput, LINK_LOST));
+        }
```
erskingardner: This arm is redundant and slightly lossy. `unix_socket_inode` produces `InvalidInput` from exactly one place — the `is_socket()` check — and this rewrites that error into a message with the same kind, discarding the more specific `"path is not a Unix socket"`. Diagnostically the original is the better string: "the path exists but is a regular file" and "the path names a different socket" are different operator situations.

Dropping the arm removes five lines and keeps the more precise message:

```rust
let actual = unix_socket_inode(path)?;
if actual != expected {
    return Err(io::Error::new(io::ErrorKind::InvalidInput, LINK_LOST));
}
Ok(())
```

crates/agent-connector/src/lib.rs
```diff
@@ -396,3 +411,23 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
         });
     }
 }
+
+async fn wait_control_socket_link_loss(
```
erskingardner, line 420: This function never returns `Ok` — the loop only exits through the error path. Returning `ConnectorError` instead of `Result<(), ConnectorError>` makes that guarantee visible at both call sites and removes the slightly odd `res = &mut link_loss => { return res; }` (which reads as if it could return success from the accept loop):

```rust
async fn wait_control_socket_link_loss(..) -> ConnectorError
```

Call sites become `err = &mut link_loss => return Err(err)`.

Also minor: every other `interval` in this workspace uses `MissedTickBehavior::Delay` (`invite_policy.rs:29`, `stream.rs:624`, `media_temp.rs:18`, `event_projection.rs:1033`). `Skip` behaves fine for a liveness poll, but if there's no reason for the divergence, match the convention.

erskingardner (commented):
Premise is sound and the shape of the fix is right.

Worth recording explicitly: the issue's root-cause hypothesis (a silently-lost `hard_link`) is not reachable in the current code — `bind_unix_listener_private` propagates the link error, so a failed link already aborts startup. And `/proc/<pid>/net/unix` showing the *staging* path is expected, not evidence of a failure: a Unix socket's bound name is whatever path `bind(2)` used, forever. So the only state consistent with the report is "bind + link succeeded, and the final path was later removed by something external" — which is exactly what this PR detects. Good targeting.

The earlier blocking findings (post-publication compensation TOCTOU, per-accept future reconstruction) look genuinely resolved at this head: all fallible validation now precedes the atomic `hard_link`, no error path unlinks the final path, and the monitor is a single pinned future reused by `&mut`. Diff is small and the public bind API is preserved.

Remaining items, none blocking:

1. **Cancelling `connector.start()`** — the startup `select!` drops `start()` at an arbitrary await point, which can be inside `publish_key_package` / `set_account_*_relays`. Inline comment; probably acceptable given the process exits immediately, but it deserves an explicit note rather than being implicit.
2. **The new connector test races the two detection paths** and most likely only exercises the startup one, leaving the accept-loop monitor (the steady-state path) untested at the connector level.
3. **A few lines are removable** in `verify_unix_socket_inode` and in the monitor's return type.

One scope question for the body rather than a line: `crates/cli/src/daemon/mod.rs:174` binds through the same `bind_unix_listener_private` and has the identical orphaned-listener exposure, but is not covered here. Leaving it out is a defensible scope call — just worth stating in the PR body whether that is deliberate (`wn-agent` is the supervised long-lived service, `mdk daemon` is not) or a follow-up.

crates/agent-connector/src/lib.rs
agent-p1p: Addressed in 836a1c7b. The select now documents the intentional cancellation contract: completed idempotent startup work is reconciled on restart, and non-restart-safe multi-step work must not be added without revisiting this design.

crates/agent-connector/src/tests.rs
agent-p1p: Addressed in 836a1c7b. The test now completes an AccountList request/response before unlinking the socket, so it deterministically exercises steady-state serve behavior rather than startup scheduling.

crates/agent-connector/src/tests.rs
agent-p1p: Addressed in 836a1c7b. Added a separate atomic socket-replacement regression that asserts InvalidInput; the removal test now asserts NotFound exactly.

crates/fs-private/src/lib.rs
agent-p1p: Addressed in 836a1c7b. Simplified the inode check to metadata(...).and_then(...).then_some(()).ok_or_else(...), preserving the original non-socket diagnostic.

crates/agent-connector/src/lib.rs
agent-p1p: Addressed in 836a1c7b. The monitor now returns ConnectorError directly and uses tokio::time::Interval with MissedTickBehavior::Delay.

agent-p1p:
Scope note added to the PR body: this patch deliberately targets the supervised, long-lived `wn-agent` control socket. The standalone `wnd`/CLI daemon uses the same bind primitive but has an independent lifecycle and should receive equivalent link-loss supervision in a focused follow-up rather than broadening this security-sensitive change.

Datawav (approved):

No actionable findings.

_Advisory audit of head [`836a1c7ba6eb`](https://github.com/marmot-protocol/mdk/commit/836a1c7ba6eb4e4859070c1f824fc927b5a35273); no checkout or PR-code execution was performed. Approved as ready for merge._

PR #1285, opened 2026-08-06
Add commit-addressed checkpoints for own-commit fork recovery

erskingardner:
## Summary

Replaces #1236 with branch-addressed recovery for locally authored OpenMLS commits.

OpenMLS cannot process a device's own Commit after receiving it back from the network. The convergence engine therefore cannot reconstruct a selected branch containing a displaced own Commit from wire bytes alone, and the epoch-keyed retained snapshots used by #1236 can be overwritten by a rival branch at the same epoch.

This change:

- keeps both sides of a pairwise same-epoch race reconsiderable by distributed convergence, with atomic loser parking and branch-identifying audit records;
- derives an immutable checkpoint id from each confirmed own Commit's MLS wire digest;
- atomically captures the complete canonical MLS/Marmot group state after the pending Commit is merged and before its stored row becomes `Processed`;
- restores the exact commit-addressed checkpoint when that own branch wins later, verifies its resulting epoch authenticator, then replays descendants normally;
- deliberately leaves the live message ledger, outbound queues, convergence policy, and pass scheduler outside the checkpoint;
- retains checkpoints for the same bounded rewind horizon as convergence history and prunes them afterward.

If convergence needs an in-horizon locally authored commit from a pre-checkpoint database, the engine now reports `MissingOwnCommitCheckpoint`, leaves canonical state unchanged, and durably marks the group `Unrecoverable` for explicit repair. There is no safe way to reconstruct that exact historical branch from its network wire or an epoch-keyed anchor.

## Why this replaces #1236

#1236 correctly unified pairwise and convergence branch selection, but its final recovery mechanism still identified rollback state by epoch. Two sibling commits can produce the same epoch, so a later capture could replace the locally authored branch's only usable state. A ratchet-tree marker could detect the mismatch but could not recover the branch, leaving a liveness failure.

Commit-addressed immutable checkpoints remove that aliasing: the locally authored branch remains materializable after restart for as long as the convergence engine is allowed to select it.

## Validation

- `just fast-ci`
- `cargo test -p cgka-traits --test snapshots`

The fork regression explicitly confirms an own Commit, lets a rival same-epoch branch overwrite the ordinary epoch anchor, restarts the original author, grows the own branch deeper, and verifies convergence onto that branch.

## Summary by CodeRabbit

* **New Features**
  * Improved same-epoch fork convergence so losing branches remain eligible for later reconciliation instead of becoming permanently invalid.
  * Added durable group-state checkpoints to support safe branch reorganization, restart recovery, and descendant replay.
  * Preserved winning commit ordering and expanded audit records with retained commit details.
* **Bug Fixes**
  * Prevented incorrect state restoration during fork resolution by validating recovered state before continuing.
  * Added fail-closed handling when required recovery checkpoints are unavailable.
* **Documentation**
  * Updated the unreleased changelog with convergence, checkpoint, and recovery improvements.

### DIFF
--- a/crates/cgka-engine/src/fork_recovery.rs
+++ b/crates/cgka-engine/src/fork_recovery.rs
@@ -14,7 +14,8 @@
 //! commit is the transport-layer `MessageId`. The two are kept separate inside
 //! `CommitRecoveryRecord` so the ordering key remains transport-independent
 //! while the engine can still reach back to the storage record that needs
-//! `MessageState::EpochInvalidated`.
+//! parking as `MessageState::ConvergenceDeferred` (reconsiderable by a later
+//! distributed-convergence pass, not terminal).
 
 use crate::engine::Engine;
 use cgka_traits::engine::{CommitOrderingKey, CommitOrderingPriority};
@@ -32,8 +33,9 @@ struct CommitRecoveryRecord {
     group_id: GroupId,
     source_epoch: EpochId,
     ordering_key: CommitOrderingKey,
-    /// Storage-layer identity of the commit. Used to update
-    /// `MessageState::EpochInvalidated` when this commit loses a fork.
+    /// Storage-layer identity of the commit. Used to park the row
+    /// `MessageState::ConvergenceDeferred` when this commit loses a fork
+    /// (its branch stays reconsiderable for distributed convergence).
     storage_id: MessageId,
     snapshot_name: String,
 }
@@ -43,8 +45,8 @@ pub(crate) enum ForkResolution {
     CandidateWins {
         winner: CommitOrderingKey,
         invalidated: CommitOrderingKey,
-        /// `MessageId` of the now-invalidated incumbent, for the storage
-        /// `update_message_state` call site.
+        /// `MessageId` of the now-displaced incumbent, for the storage
+        /// write that parks its row `ConvergenceDeferred`.
         invalidated_storage_id: MessageId,
     },
     IncumbentWins {
@@ -176,12 +178,45 @@ impl ForkRecoveryManager {
             });
         }
 
+        // Capture the incumbent's stored row BEFORE the rollback: the
+        // pre-commit snapshot predates the incumbent's own persist, so the
+        // rollback below sweeps its row away entirely. Re-persisting it
+        // afterwards (parked, see below) is what keeps the displaced branch
+        // reconsiderable.
+        let displaced_incumbent_row = match storage.get_message(&incumbent.storage_id) {
+            Ok(record) => Some(record),
+            Err(StorageError::NotFound) => None,
+            Err(e) => return Err(EngineError::Storage(e)),
+        };
         storage.rollback_group_to_snapshot(group_id, &incumbent.snapshot_name)?;
         match storage.release_group_snapshot(group_id, &incumbent.snapshot_name) {
             Ok(()) | Err(StorageError::SnapshotMissing(_)) => {}
             Err(e) => return Err(EngineError::Storage(e)),
         }
         self.incumbents.remove(&key);
+
+        // Park the pairwise-losing incumbent `ConvergenceDeferred`, not
+        // terminally `EpochInvalidated` — the mirror of the incumbent-wins
+        // seam in `ingest_group_message`. The losing branch may still gain
+        // follow-on commits from peers that applied it and never saw the
+        // candidate; distributed convergence then selects the DEEPER branch,
+        // and every convergence-only node reorgs onto it. A terminal (or
+        // missing) row is excluded from convergence input, so this node could
+        // never materialize that branch's root and would diverge from the
+        // fleet forever. If the incumbent's branch also loses in convergence,
+        // it is re-classified terminal there (`LosingBranch`).
+        //
+        // The row is re-keyed at the commit's SOURCE epoch (convergence
+        // derives its rewind target from `record.epoch`) and its payload is
+        // preserved verbatim — for an own commit it carries the convergence
+        // ordering stamp (`stamp_processed_own_commit_record`) that replay
+        // needs to rebuild the ordering key (MLS refuses to process own
+        // commits).
+        if let Some(mut record) = displaced_incumbent_row {
+            record.epoch = incumbent.source_epoch;
+            record.state = MessageState::ConvergenceDeferred;
+            storage.put_message(&record)?;
+        }
 
         Ok(ForkResolution::CandidateWins {
             winner: candidate_key,
@@ -370,18 +405,14 @@ impl<S: StorageProvider> Engine<S> {
         {
             self.epoch_manager
                 .set_stable(group_id.clone(), source_epoch);
-            match self
-                .storage
-                .update_message_state(invalidated_storage_id, MessageState::EpochInvalidated)
-            {
-                Ok(()) | Err(StorageError::NotFound) => {}
-                Err(e) => return Err(EngineError::Storage(e)),
-            }
+            // `ForkRecoveryManager::resolve` parked the displaced incumbent's
+            // row `ConvergenceDeferred` at its source epoch (captured before
+            // the rollback swept it); record the transition for forensics.
             self.audit_group(
                 group_id,
                 crate::audit_helpers::message_state_changed_event(
                     hex::encode(invalidated_storage_id.as_slice()),
-                    MessageState::EpochInvalidated,
+                    MessageState::ConvergenceDeferred,
                     "fork_loser",
                 ),
             );
--- a/crates/cgka-engine/src/fork_recovery.rs
+++ b/crates/cgka-engine/src/fork_recovery.rs
@@ -576,6 +576,37 @@ mod tests {
             self.release_attempts.lock().unwrap().push(name.to_string());
             Err(StorageError::Backend("release failed".into()))
         }
+
+        fn create_group_state_checkpoint(
+            &self,
+            _group_id: &GroupId,
+            _checkpoint: &cgka_traits::storage::GroupStateCheckpointRef,
+        ) -> StorageResult<()> {
+            unused()
+        }
+
+        fn restore_group_state_checkpoint(
+            &self,
+            _group_id: &GroupId,
+            _checkpoint_id: &str,
+        ) -> StorageResult<()> {
+            unused()
+        }
+
+        fn list_group_state_checkpoints(
+            &self,
+            _group_id: &GroupId,
+        ) -> StorageResult<Vec<cgka_traits::storage::GroupStateCheckpointRef>> {
+            unused()
+        }
+
+        fn release_group_state_checkpoint(
+            &self,
+            _group_id: &GroupId,
+            _checkpoint_id: &str,
+        ) -> StorageResult<()> {
+            unused()
+        }
     }
 
     #[test]
--- a/crates/cgka-engine/src/fork_recovery.rs
+++ b/crates/cgka-engine/src/fork_recovery.rs
@@ -160,10 +160,8 @@ impl ForkRecoveryManager {
     /// Decide the pairwise race and, when the candidate wins, displace the
     /// incumbent atomically.
     ///
-    /// Bounded by `StorageProvider` rather than `MessageStorage` specifically
-    /// so the displacement can run inside `with_transaction`: capture,
-    /// rollback, re-persist and release are one durable unit. Nothing weaker
-    /// works — see the ordering comment below.
+    /// Bounded by `StorageProvider` so capture, rollback, re-persist and
+    /// release can join one transaction.
     fn resolve<S: StorageProvider>(
         &mut self,
         storage: &S,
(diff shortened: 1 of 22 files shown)


erskingardner (commented):
## Overall

The premise holds up. #1236's epoch-keyed anchor genuinely aliases sibling branches, and the two new regression tests (`own_commit_checkpoint_survives_rival_anchor_overwrite_and_restart`, `pairwise_candidate_win_leaves_old_incumbent_reconsiderable`) drive the exact failure they claim to. Deriving the checkpoint id from the MLS wire digest is the right fix, and confining the checkpoint to canonical group state (excluding ledger/queues/policy/pass) is a good call that's clearly documented on `GroupStateCheckpoint`.

Wrapping the pairwise displacement in `with_transaction` is a real correctness improvement independent of the checkpoint work.

No blocking correctness defect found. What I have is one redundant-work bug, a type-design issue that produces two `expect()` panics inside a storage transaction, two test-coverage gaps, and a comment-density concern. Details inline.

### Scope

2081 additions is a lot, and this really is two changes: the #1236 parking/atomicity work and the new checkpoint subsystem (trait + migration + module). The PR body argues they're coupled — parking is a liveness hazard without a recoverable own-branch — and I buy that, so I'm not asking for a split. But it does mean the checkpoint machinery gets reviewed under the urgency of the convergence bug. Worth calling out.

### Comment density

`fork_recovery.rs::resolve` now carries ~50 lines of comment for ~25 lines of code, and the `IncumbentWins` arm in `ingest.rs` ~55 lines for ~12. Most of it is *why #1236's shape was wrong* — history, not invariants. That's the kind of prose that rots first and misleads later readers about what the code currently guarantees. I'd cut each to the load-bearing invariant (the source-epoch re-key, the "no active pass" precondition) and let the PR body and CHANGELOG carry the archaeology.

### Not blocking, but worth confirming

`checkpoint_realizable_own_commit_prefix` pre-checks `list_group_state_checkpoints` for id + `resulting_epoch` match, then `verify_restored_own_checkpoint` re-verifies epoch + authenticator after the restore. Given restore already returns `SnapshotMissing` for an absent id and the authenticator check is the real gate, the `retained` scan is defensive duplication at O(commits × checkpoints). It also reintroduces an epoch-derived condition (`record.epoch.0 + 1`) into a scheme whose whole point is to stop keying on epoch. Fine as a fail-closed guard, but it isn't doing safety work the verifier isn't already doing.

Also: `prune_group_state_checkpoints` is a near-line-for-line copy of `prune_retained_anchor_snapshots` (same horizon arithmetic, same skip condition). Two copies of a retention rule drift; consider factoring the horizon calculation.

### One more, un-anchorable inline (unchanged file)

`crates/storage-sqlite/src/migrations.rs` — the `group_owned_tables_have_cascading_foreign_keys` test (~line 1310) enumerates every group-owned table and asserts it cascades. `cgka_group_state_checkpoints` declares `ON DELETE CASCADE` but is not in that list. Please add:

```rust
("cgka_group_state_checkpoints", "group_id"),
```

Without it, a future migration that rebuilds the table (the 12-step SQLite dance this repo already does elsewhere) can silently drop the cascade, leaving checkpoint blobs — serialized OpenMLS provider values — outliving the group they belong to.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1648,7 +1726,27 @@ pub(crate) fn apply_openmls_canonicalization_result_with_profile_policy<S: Stora
...
+                    .expect("checkpoint prefix has resulting epoch"),
+                own_checkpoint_prefix
+                    .expected_epoch_authenticator
+                    .as_deref()
+                    .expect("checkpoint prefix has epoch authenticator"),
+            )?;
+            restore_live_message_and_queue_records(
+                storage,
+                &live_message_records,
+                &live_queued_outbound,
```
erskingardner: `restore_live_message_and_queue_records` is a no-op here, and an expensive one.

`restore_group_state_checkpoint` -> `restore::restore_group_state` touches exactly four things: `cgka_groups`, `cgka_member_capabilities`, the validated-tree marker, and `openmls_values`. By design (see the `GroupStateCheckpoint` doc comment) it does **not** touch `cgka_messages` or `cgka_queued_outbound`. So nothing rewound them, and this call re-`put_message`s every live message row in the group plus every queued intent, inside the apply transaction, to write back byte-identical content.

In the `else if rewind_to_retained_anchor` branch below it is genuinely required, because `rollback_group_to_snapshot` *does* restore the ledger from the snapshot. In the checkpoint branch it's pure write amplification proportional to group history — and it reads as if the checkpoint restore rewound the ledger, which is the opposite of the invariant the struct doc establishes.

Drop the call from the checkpoint branch. If it's being kept as belt-and-braces, that needs a comment saying so, because the current code implies a rewind that does not happen.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1786,6 +1884,100 @@ fn apply_start_epoch_for_canonicalization_result<S: StorageProvider>(
     Ok(Some(record.epoch.0))
 }
 
+#[derive(Default)]
+struct CheckpointRealizableOwnPrefix {
```
erskingardner, line 1922: These four fields have a hard coupling the type doesn't express: `resulting_epoch`, `checkpoint_id`, and `expected_epoch_authenticator` are always all-Some or all-None (the loop only ever sets them together, and only after the `let (Some(..), Some(..)) = ... else { break }`). The apply path then has to assert that coupling at runtime:

```rust
own_checkpoint_prefix.resulting_epoch.expect("checkpoint prefix has resulting epoch"),
own_checkpoint_prefix.expected_epoch_authenticator.as_deref().expect("checkpoint prefix has epoch authenticator"),
```

Two `expect()`s that panic *inside a storage transaction*, guarding an invariant a struct definition could make unrepresentable. A future edit to the loop that sets `checkpoint_id` on one path and not `resulting_epoch` on another turns a compile error into a panic mid-apply.

Collapse to:

```rust
struct RealizedCheckpoint {
    resulting_epoch: u64,
    id: String,
    expected_epoch_authenticator: String,
}

#[derive(Default)]
struct CheckpointRealizableOwnPrefix {
    commit_ids: BTreeSet<String>,
    realized: Option<RealizedCheckpoint>,
}
```

Then `if let Some(realized) = &own_checkpoint_prefix.realized` replaces both the `if let Some(checkpoint_id) = ...as_deref()` and the two `expect`s, and `restore_own_checkpoint` (line 1687) becomes `realized.is_some()` at its single use site rather than a separately-tracked bool.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1786,6 +1884,100 @@ fn apply_start_epoch_for_canonicalization_result<S: StorageProvider>(
...
+        prefix.resulting_epoch = Some(resulting_epoch);
+        prefix.checkpoint_id = Some(checkpoint_id);
+        prefix.expected_epoch_authenticator = Some(expected_authenticator);
+    }
+    Ok(prefix)
+}
+
+/// Verify the exact own-commit checkpoint after restoring it. This runs inside
+/// the apply transaction so any mismatch rolls back to the pre-apply state.
+fn verify_restored_own_checkpoint<S: StorageProvider>(
```
erskingardner, line 1989: `CheckpointStateMismatch` is the safety valve for the entire scheme — it is the only thing standing between a wrong-branch checkpoint and an unverified lineage getting installed — and nothing in this PR exercises it.

The four tests in `checkpoint_prefix_tests` all cover the *selection* side (`checkpoint_realizable_own_commit_prefix`), and the integration tests all take the happy path. There's no test that a checkpoint restoring the wrong epoch, or the right epoch with a mismatched authenticator, aborts the apply and rolls back to the pre-apply state.

That last part matters most: the PR's claim is that "any mismatch rolls back to the pre-apply state". A unit test of `verify_restored_own_checkpoint` against a deliberately-mismatched authenticator plus an assertion that the group is unchanged after the failed apply would make that claim verifiable rather than asserted in a doc comment.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -969,9 +1013,17 @@ pub(crate) fn canonicalize_stored_openmls_messages_with_profile_policy<S: Storag
...
+                // The stamp is written only at confirm time, so its presence
+                // alone proves this is an own commit MLS cannot re-process.
+                // Register it independently of `Processed`: a pairwise
+                // CandidateWins resolution parks the displaced own incumbent
+                // `ConvergenceDeferred` (see `fork_recovery`), and replaying
+                // its branch still needs the stamp — the prefix-canonical
+                // guard at the stamp's use site keeps a diverging prefix from
+                // mis-realizing it.
+                if let Some(stamp) = own_commit_stamp {
+                    own_commits.insert_stamped(projection.message_digest, stamp);
```
erskingardner, line 1032: Worth confirming this widening is as narrow as intended.

Previously `insert_stamped` was reachable only for `record.state == Processed`. It's now unconditional on state for any row carrying a stamp. The stated reason — a `ConvergenceDeferred` displaced own incumbent still needs its stamp — is correct and necessary.

But the new condition admits *every* stamped state, including terminal ones (`EpochInvalidated`, whatever `classify_losing_materialized_candidate_commits` assigns at the rewind horizon). If a terminally-disposed own commit can reach this loop, it now registers as a prevalidated own commit and becomes eligible for checkpoint realization at the stamp's use site.

If the caller already filters terminal states out of `commit_messages`, say so in the comment — right now the comment justifies admitting `ConvergenceDeferred` specifically, but the code admits everything. Tightening to an explicit allowlist (`Processed | ConvergenceDeferred`) would make the intent match the code and cost nothing.

crates/cgka-engine/src/fork_recovery.rs
```diff
@@ -164,20 +185,100 @@ impl ForkRecoveryManager {
...
+                kept: incumbent.ordering_key,
+            });
         }
 
-        storage.rollback_group_to_snapshot(group_id, &incumbent.snapshot_name)?;
-        match storage.release_group_snapshot(group_id, &incumbent.snapshot_name) {
-            Ok(()) | Err(StorageError::SnapshotMissing(_)) => {}
-            Err(e) => return Err(EngineError::Storage(e)),
-        }
+        // Atomicity rationale (durability of the displaced branch). Capture,
```
erskingardner: Worth a note on the atomicity claim, since the comment is emphatic about it ("Nothing weaker works").

`SqliteConnection::with_transaction` is reentrant on the same thread — a nested call just runs the closure under the already-open outer transaction (`connection.rs:176-181`). `resolve` is reached from `resolve_fork_candidate` inside `ingest_group_message`, so whether this closure gets its own SQL transaction or joins an outer one depends on the caller.

The guarantee still holds either way (an error propagates and the outer transaction aborts), so this isn't a defect. But the comment reads as though this call establishes the transaction boundary, and in the nested case it does not. One sentence acknowledging that would keep the comment honest.

Separately: this block is ~50 lines of comment for ~25 lines of code, and most of it explains why #1236's shape was wrong rather than what the current code guarantees. The load-bearing parts are the source-epoch re-key and the "`incumbents.remove` stays outside the commit" note. The rest belongs in the PR body, which already says it.

Datawav:

## Final audit

The change is needed: epoch-addressed snapshots can alias sibling branches, while OpenMLS cannot replay a device’s own commit. Commit-digest-addressed immutable checkpoints, captured atomically with the post-merge state and verified by epoch authenticator, are an appropriate design. Epoch-keyed fallback cannot guarantee the selected branch; backfilling exact historical checkpoints is generally impossible.

- **High — Upgrades can permanently wedge convergence on legacy own commits** (`crates/cgka-engine/src/openmls_projection.rs`, `checkpoint_realizable_own_commit_prefix` and own-commit replay path; migration `0042_group_state_checkpoints.rs`). Existing stamps deserialize with no checkpoint metadata, and the migration creates an empty checkpoint table. If convergence selects an in-horizon legacy own commit, prefix realization declines it, ordinary replay reaches it, and returns `Replay("own commit has no branch checkpoint")`. Because the apply transaction rolls back, the same deferred row remains eligible and can be selected repeatedly; advancing or “aging out” is not demonstrated and may be impossible while the group cannot apply the pass. This turns a documented fail-closed compatibility choice into a persistent liveness failure. Add an upgrade-path test using a pre-v42 stamped own commit and define a terminal recovery policy. Safe alternatives include marking the group explicitly unrecoverable with an audit event, or excluding checkpoint-less branches through a deterministic protocol-wide rule. Retaining the old epoch-anchor fallback is only acceptable if branch identity can be independently verified; otherwise it reintroduces the aliasing defect this PR addresses.

_Advisory audit of head [`c3330e2fa886`](https://github.com/marmot-protocol/mdk/commit/c3330e2fa886eb073f754631f634434f97dd4d3d); no approval, merge, push, checkout, or PR-code execution was performed._

erskingardner (commented):
## Verdict: would request changes

GitHub will not let me formally request changes on your own PR, but I would not merge this yet.

The commit-addressed checkpoint design is the right fix for the epoch-anchor aliasing hole in #1236, and the happy-path / mismatch / pre-v42 coverage is strong. CI is green.

**Blocking:** apply still assumes every own commit on the selected branch is either already-applied or in a **leading** checkpoint run. That breaks once a reconsiderable branch has peer commits between this device's own commits — selection can succeed while apply permanently fails. Details on the inline comments.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1913,9 +2161,11 @@ fn apply_openmls_canonicalization_result_inner<S: StorageProvider>(
...
     storage.with_transaction(|storage| {
         // No pre-validated own commits here: snapshot rollforward cannot nest
-        // inside this transaction, and the already-applied prefix (which is
-        // where own commits can appear on an accepted branch) was excluded
-        // from `replay_messages` before this call.
+        // inside this transaction, and every own commit on the accepted
+        // branch was excluded from `replay_messages` before this call —
+        // either as part of the already-applied prefix or as a
+        // checkpoint-realizable own-commit run whose exact restore the
+        // caller performs itself (`checkpoint_realizable_own_commit_prefix`).
```
erskingardner, line 2168: Blocking: this comment (and the empty `PrevalidatedOwnCommits::default()` passed into `process_openmls_messages_inner` just below) is wrong, and it leaves mid-path own commits unrealizable on apply.

Selection materializes with real stamps, so a branch like `ownA → peer → ownB` (both own commits `ConvergenceDeferred` after an earlier non-selected-eligible park) can win. Apply then:
1. restores only the leading run via `checkpoint_realizable_own_commit_prefix` (`ownA`);
2. puts `peer` and `ownB` into `replay_messages`;
3. replays them with no stamps, so `ownB` hits `process_message` and fails.

Checkpoint restore **does** join an open SQLite transaction (`checkpoints.rs` transaction-owner path), and not every accepted own commit was excluded from replay — only a contiguous prefix.

Fix: rebuild/pass the same stamped own-commit set into this inner replay (selection already proves nesting is safe), and add an integration test that parks `own → peer → own`, selects that branch later, and asserts both own commits become `Processed`.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1786,6 +1918,95 @@ fn apply_start_epoch_for_canonicalization_result<S: StorageProvider>(
...
+}
+
+struct RealizedCheckpoint {
+    resulting_epoch: u64,
+    id: String,
+    expected_epoch_authenticator: String,
+}
+
+/// The leading run — after the already-applied prefix — of this device's own
+/// confirm-stamped commits on the selected branch, bounded to those whose
```
erskingardner, line 1936: Related to the apply bug: this says the run is bounded to commits "whose commit-addressed checkpoint still exists", but the loop only checks stamp fields — it never looks up the checkpoint row. Existence is enforced later at restore time, and inconsistently: selection maps `SnapshotMissing` → `MissingOwnCommitCheckpoint`, while the apply-stage restore maps it to a generic `Snapshot` error.

Once mid-path own commits are realized during apply, please either check presence here or make the apply-stage missing-checkpoint mapping match selection so an in-horizon pruned/missing blob fails closed as `Unrecoverable` rather than a retryable apply error.

PR #1288, opened 2026-08-06
feat: expose dial-safe profile image fetch

### LINKED ISSUE #1287: FFI: expose MDK-owned dial-safe fetch for public profile images
## What

Expose an MDK-owned, dial-safe fetch primitive for public profile image URLs so host apps do not maintain a second SSRF classifier or open attacker-controlled sockets themselves.

`marmot-protocol/whitenoise-android#1585` is blocked on this boundary. Kind-0 `picture` values are protocol data, but the current Android binding only exposes encrypted `downloadMedia` / `downloadGroupBlossomImage`; the app therefore fetches profile avatars through its Kotlin `SafeHttpsGet` stack.

## Required behavior

- Accept an untrusted public profile-image URL and a bounded maximum response size.
- Apply MDK URL rules (length, scheme, credentials, fragments, ports, localhost/private/special-use hosts).
- Resolve each redirect hop, reject non-public addresses, pin the vetted address for the actual dial, and enforce connect/read/overall deadlines.
- Bound the response while streaming regardless of `Content-Length`.
- Return bytes through UniFFI for Android and Swift consumers.
- Keep the API purpose-specific enough that callers cannot accidentally use it as an unrestricted network client.

## Suggested shape

```rust
pub async fn download_profile_image(
    url: String,
    max_bytes: u64,
) -> Result<Vec<u8>, MarmotKitError>;
```

The exact shape is open; the invariant is that MDK owns validation + pinned dialing, not the host app.

## Existing pieces

- `crates/marmot-app/src/media/host_safety.rs`
- Blossom pinned fetch/dial-safety used by encrypted media
- `cgka_traits::app_components::{is_loopback_host, reject_non_public_ip}`
- Existing UniFFI `downloadMedia` and `downloadGroupBlossomImage` APIs

## Acceptance

- Generated Kotlin and Swift bindings expose the profile-image fetch.
- Tests cover loopback/private/special-use literals, rooted localhost forms, public-name-to-private DNS answers, redirects, rebind-safe address pinning, URL/body limits, credentials, fragments, and scheme/port rejection.
- Android can delete profile-avatar uses of `HostSafety` / `SafeHttpsGet` and retain one instrumented regression proving a loopback profile avatar is never dialed.

Consumer: marmot-protocol/whitenoise-android#1585

agent-p1p:
Fixes #1287

## Summary
- Add an MDK-owned `download_profile_image(url, max_bytes)` path for untrusted kind-0 picture URLs.
- Enforce HTTPS/default-port URL policy, localhost/private/special-use rejection, per-hop DNS vetting and reqwest address pinning, bounded manual redirects, disabled proxies/decompression, caller-bounded streaming, and an operation-wide deadline.
- Export the async method through UniFFI and cover the generated Kotlin/Swift surface; add `Unreleased` changelog coverage without version churn.

## Verification
- `cargo test -p marmot-app --locked --lib profile_image` — 18 passed.
- `cargo test -p marmot-uniffi --locked profile_image` — 2 passed.
- `just fast-ci` — passed (format, naming/ledger gates, workspace all-target checks, clippy with/without OTLP, convergence policy tests).
- Generated host Kotlin and Swift bindings with `uniffi-bindgen`; both contain `downloadProfileImage(url, maxBytes)`.
- Independent Composer security review — passed with no security or logic findings.

## Baseline recheck
- Rebased onto current `master` (`bca10777`).
- `cold_reopen_recovers_quiet_offline_history_without_a_live_trigger` now passes with the valid package-scoped feature set.
- Current `master` has a fully green GitHub CI matrix, including `Simulator and vectors`.
- The earlier blocker was stale; one attempted package-scoped reproduction also used an invalid cross-package feature flag and was not test evidence.

## Sensitive paths
- `crates/marmot-app/src/media/blossom.rs`
- `crates/marmot-app/src/media/host_safety.rs`

These implement the SSRF/dial-safety trust boundary and DNS pinning for untrusted profile-image URLs.

## Summary by CodeRabbit

- **New Features**
  - Added profile-image downloading through the public MarmotKit API.
  - Supports caller-defined download limits up to 10 MiB.
  - Added safeguards for HTTPS URLs, public destinations, bounded redirects, DNS resolution, timeouts, and response sizes.
  - Added clear errors for invalid or unsafe media references.

- **Documentation**
  - Documented the profile-image download API and its security protections.

### DIFF
--- a/crates/marmot-app/src/media/blossom.rs
+++ b/crates/marmot-app/src/media/blossom.rs
@@ -8,10 +8,13 @@ use serde::Deserialize;
 use url::{Host, Url};
 
 use super::host_safety::{
-    is_loopback_host, reject_non_public_ip, validate_blossom_fetch_url,
-    validate_profile_image_fetch_url,
+    is_loopback_host, parse_profile_image_fetch_url, parse_profile_image_redirect_url,
+    reject_non_public_ip, validate_blossom_fetch_url,
 };
 use crate::{AppError, unix_now_seconds};
+
+#[cfg(test)]
+use cgka_traits::app_components::ENCRYPTED_MEDIA_ENDPOINT_URL_MAX_LEN;
 
 const BLOSSOM_UPLOAD_AUTH_TTL: Duration = Duration::from_secs(10 * 60);
 const BLOSSOM_UPLOAD_CONTENT_TYPE: &str = "application/octet-stream";
@@ -171,107 +174,118 @@ pub(crate) async fn fetch_blossom_blob(
     url: &str,
     allow_loopback_http: bool,
 ) -> Result<Vec<u8>, AppError> {
-    let mut current = Url::parse(url)
+    let current = Url::parse(url)
         .map_err(|_| AppError::InvalidEncryptedMedia("media URL is invalid".into()))?;
     validate_blossom_fetch_url(&current, allow_loopback_http)
         .map_err(|err| AppError::BlobStore(format!("unsafe Blossom URL: {err}")))?;
-    let mut redirects = 0_usize;
-
-    loop {
-        let client = media_http_client_for_url(&current, allow_loopback_http).await?;
-        let response = client
-            .get(current.clone())
-            .send()
-            .await
-            .map_err(reqwest_blob_error)?;
-        let status = response.status();
-        if status.is_success() {
-            return read_limited_blossom_body(response, MAX_ENCRYPTED_MEDIA_BLOB_BYTES).await;
-        }
-        if !status.is_redirection() {
-            return Err(AppError::BlobStore(format!(
-                "download returned HTTP {}",
-                status.as_u16()
-            )));
-        }
-
-        if redirects >= BLOSSOM_REDIRECT_LIMIT {
-            return Err(AppError::BlobStore(format!(
-                "media redirect chain exceeded {BLOSSOM_REDIRECT_LIMIT} hops"
-            )));
-        }
-        let location = response
-            .headers()
-            .get(reqwest::header::LOCATION)
-            .ok_or_else(|| {
-                AppError::BlobStore("redirect response did not include Location".into())
-            })?
-            .to_str()
-            .map_err(|_| AppError::BlobStore("redirect Location header is invalid".into()))?;
-        let next = current.join(location).map_err(|_| {
-            AppError::BlobStore("redirect Location header is not a valid URL".into())
-        })?;
-        validate_blossom_redirect_target(&current, &next, allow_loopback_http)?;
-        current = next;
-        redirects += 1;
-    }
-}
-
-pub(crate) async fn fetch_profile_image(url: &str, max_bytes: u64) -> Result<Vec<u8>, AppError> {
-    profile_operation_timeout(
-        MEDIA_HTTP_TOTAL_TIMEOUT,
-        fetch_profile_image_impl(url, max_bytes, ProfileResolveMode::Production),
+    fetch_http_with_bounded_redirects(
+        current,
+        MAX_ENCRYPTED_MEDIA_BLOB_BYTES,
+        move |url| {
+            let allow_loopback_http = allow_loopback_http;
+            async move { media_http_client_for_url(&url, allow_loopback_http).await }
+        },
+        move |current, location| {
+            let next = current.join(location).map_err(|_| {
+                AppError::BlobStore("redirect Location header is not a valid URL".into())
+            })?;
+            validate_blossom_redirect_target(current, &next, allow_loopback_http)?;
+            Ok(next)
+        },
     )
     .await
 }
 
+pub(crate) async fn fetch_profile_image(url: &str, max_bytes: u64) -> Result<Vec<u8>, AppError> {
+    profile_operation_timeout(fetch_profile_image_impl(url, max_bytes)).await
+}
+
 #[cfg(test)]
-pub(crate) async fn fetch_profile_image_with_injected_addrs(
+pub(crate) async fn fetch_profile_image_with_loopback(
     url: &str,
     max_bytes: u64,
-    injected_addrs: Vec<SocketAddr>,
 ) -> Result<Vec<u8>, AppError> {
-    profile_operation_timeout(
-        MEDIA_HTTP_TOTAL_TIMEOUT,
-        fetch_profile_image_impl(url, max_bytes, ProfileResolveMode::Injected(injected_addrs)),
-    )
+    let raw = url.trim();
+    if raw.len() > ENCRYPTED_MEDIA_ENDPOINT_URL_MAX_LEN {
+        return Err(AppError::UnsafeMediaFetch(format!(
+            "profile image URL exceeds {ENCRYPTED_MEDIA_ENDPOINT_URL_MAX_LEN} bytes"
+        )));
+    }
+    let current = Url::parse(raw).map_err(|err| {
+        AppError::UnsafeMediaFetch(format!("profile image URL is invalid: {err}"))
+    })?;
+    validate_blossom_fetch_url(&current, true).map_err(AppError::UnsafeMediaFetch)?;
+    profile_operation_timeout(fetch_http_with_bounded_redirects(
+        current,
+        max_bytes,
+        move |url| async move { media_http_client_for_url(&url, true).await },
+        move |current, location| {
+            let raw_location = location.trim();
+            if raw_location.len() > ENCRYPTED_MEDIA_ENDPOINT_URL_MAX_LEN {
+                return Err(AppError::UnsafeMediaFetch(format!(
+                    "profile image redirect URL exceeds {ENCRYPTED_MEDIA_ENDPOINT_URL_MAX_LEN} bytes"
+                )));
+            }
+            let next = current.join(raw_location).map_err(|err| {
+                AppError::UnsafeMediaFetch(format!(
+                    "profile image redirect URL is invalid: {err}"
+                ))
+            })?;
+            validate_blossom_fetch_url(&next, true).map_err(AppError::UnsafeMediaFetch)?;
+            Ok(next)
+        },
+    ))
     .await
 }
 
 async fn profile_operation_timeout<T>(
-    timeout: Duration,
     operation: impl std::future::Future<Output = Result<T, AppError>>,
 ) -> Result<T, AppError> {
-    tokio::time::timeout(timeout, operation)
+    tokio::time::timeout(MEDIA_HTTP_TOTAL_TIMEOUT, operation)
         .await
         .map_err(|_| AppError::BlobStore("request timed out".into()))?
 }
 
-#[cfg(test)]
-pub(super) async fn profile_operation_timeout_for_test(timeout: Duration) -> Result<(), AppError> {
-    profile_operation_timeout(timeout, std::future::pending()).await
+async fn fetch_profile_image_impl(url: &str, max_bytes: u64) -> Result<Vec<u8>, AppError> {
+    let current = parse_profile_image_fetch_url(url).map_err(AppError::UnsafeMediaFetch)?;
+    fetch_http_with_bounded_redirects(
+        current,
+        max_bytes,
+        move |url| async move { media_http_client_for_profile(&url).await },
+        move |current, location| {
+            parse_profile_image_redirect_url(current, location).map_err(AppError::UnsafeMediaFetch)
+        },
+    )
+    .await
 }
 
-enum ProfileResolveMode {
-    Production,
-    #[cfg(test)]
-    Injected(Vec<SocketAddr>),
+/// Vet every DNS answer before pinning a profile-image fetch.
+pub(crate) fn vet_profile_fetch_resolved_addresses(addrs: &[SocketAddr]) -> Result<(), AppError> {
+    for addr in addrs {
+        reject_non_public_ip(addr.ip(), false).map_err(|err| {
+            AppError::UnsafeMediaFetch(format!("unsafe profile image host address: {err}"))
+        })?;
+    }
+    Ok(())
 }
 
-async fn fetch_profile_image_impl(
-    url: &str,
-    max_bytes: u64,
-    resolve_mode: ProfileResolveMode,
-) -> Result<Vec<u8>, AppError> {
-    let mut current = Url::parse(url)
-        .map_err(|_| AppError::InvalidAppMessagePayload("profile image URL is invalid".into()))?;
-    validate_profile_image_fetch_url(&current).map_err(|err| {
-        AppError::InvalidAppMessagePayload(format!("unsafe profile image URL: {err}"))
-    })?;
+/// Shared manual redirect loop for Blossom blob and profile-image fetches.
+/// Keep policy differences in the caller-supplied client and redirect validators.
+async fn fetch_http_with_bounded_redirects<C, CFut, R>(
+    mut current: Url,
+    max_body_bytes: u64,
+    mut client_for_url: C,
+    mut redirect_target: R,
+) -> Result<Vec<u8>, AppError>
+where
+    C: FnMut(Url) -> CFut,
+    CFut: std::future::Future<Output = Result<reqwest::Client, AppError>>,
+    R: FnMut(&Url, &str) -> Result<Url, AppError>,
+{
     let mut redirects = 0_usize;
 
     loop {
-        let client = media_http_client_for_profile(&current, &resolve_mode).await?;
+        let client = client_for_url(current.clone()).await?;
         let response = client
             .get(current.clone())
             .send()
@@ -279,7 +293,7 @@ async fn fetch_profile_image_impl(
             .map_err(reqwest_blob_error)?;
         let status = response.status();
         if status.is_success() {
-            return read_limited_blossom_body(response, max_bytes).await;
+            return read_limited_blossom_body(response, max_body_bytes).await;
         }
         if !status.is_redirection() {
             return Err(AppError::BlobStore(format!(
@@ -301,28 +315,8 @@ async fn fetch_profile_image_impl(
             })?
             .to_str()
             .map_err(|_| AppError::BlobStore("redirect Location header is invalid".into()))?;
-        let next = current.join(location).map_err(|_| {
-            AppError::BlobStore("redirect Location header is not a valid URL".into())
-        })?;
-        validate_profile_image_fetch_url(&next).map_err(|err| {
-            AppError::InvalidAppMessagePayload(format!("unsafe profile image redirect URL: {err}"))
-        })?;
-        current = next;
+        current = redirect_target(&current, location)?;
         redirects += 1;
-    }
-}
-
-#[cfg(test)]
-pub(super) fn plan_profile_image_pin(
-    url: &Url,
-    injected_addrs: &[SocketAddr],
-) -> Result<Option<(String, Vec<SocketAddr>)>, AppError> {
-    match url
-        .host()
-        .ok_or_else(|| AppError::BlobStore("media URL is missing a host".into()))?
-    {
-        Host::Domain(domain) => validated_media_domain_pin(domain, injected_addrs, false).map(Some),
-        _ => Ok(None),
     }
 }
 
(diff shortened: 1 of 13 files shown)


agent-p1p:
## Adversarial review: changes required

**Verdict:** 1 blocking finding, 1 suggestion, 0 nitpicks.

### Blocking

- **`crates/marmot-app/src/media/blossom.rs:266-270` / `crates/marmot-app/src/media/host_safety.rs:76-80` — the URL-size limit is applied after WHATWG parsing and normalization, so over-limit caller input is accepted and dialed.** `fetch_profile_image_impl` parses the raw FFI string first, then `validate_blossom_fetch_url` measures only `url.as_str()`. `url::Url` removes dot segments and surrounding controls/spaces while parsing. I reproduced this with a 5,020-byte input formed from `https://example.com/` plus `"a/../"` repeated 1,000 times: `download_profile_image` normalized it and successfully fetched 559 bytes instead of rejecting it against the 2,048-byte MDK limit. This leaves the explicit URL-bound acceptance criterion unenforced at the public boundary. Check the trimmed raw string length before `Url::parse` (and retain the post-parse check), then add a regression using an over-limit dot-segment URL.

### Suggestion

- **`crates/marmot-app/src/media/tests_profile_image.rs:104-134,183-203` — add one regression that exercises the profile fetch redirect path, not only its component validators.** The current redirect tests validate two independent `Url` values, and the pin test calls the planning helper twice; neither enters `fetch_profile_image_impl`'s `Location` loop. A focused test should prove a redirect causes fresh per-hop validation/resolution/pinning and that the hop limit is enforced. This is the security-critical behavior most likely to regress during future deduplication with the Blossom path.

### Verified

- Each resolved address is classified before `resolve_to_addrs` pins the vetted set.
- HTTPS/default-port, credentials, fragments, localhost/private/special-use literals, no-proxy/no-decompression, connect/read/overall deadlines, and streaming body limits are present.
- Kotlin generation exposes `downloadProfileImage(url: String, maxBytes: ULong): ByteArray`.
- Swift generation exposes `downloadProfileImage(url: String, maxBytes: UInt64) async throws -> Data`.
- `cargo test -p marmot-app --locked --lib profile_image` — 18 passed.
- `cargo test -p marmot-uniffi --locked profile_image` — 2 passed.
- `cargo fmt --all -- --check` — passed.
- `cargo clippy -p marmot-app -p marmot-uniffi --locked --all-targets -- -D warnings` — passed.

Sensitive paths reviewed: `crates/marmot-app/src/media/blossom.rs`, `crates/marmot-app/src/media/host_safety.rs`.

crates/marmot-app/src/media/blossom.rs
```diff
@@ -214,6 +217,151 @@ pub(crate) async fn fetch_blossom_blob(
...
+    #[cfg(test)]
+    Injected(Vec<SocketAddr>),
+}
+
+async fn fetch_profile_image_impl(
+    url: &str,
+    max_bytes: u64,
+    resolve_mode: ProfileResolveMode,
+) -> Result<Vec<u8>, AppError> {
+    let mut current = Url::parse(url)
```
erskingardner: **Blocking: the URL length limit is applied after normalization, so over-limit input is accepted and dialed.**

`Url::parse` runs first, and `validate_profile_image_fetch_url` -> `validate_blossom_fetch_url` (`host_safety.rs:76`) then measures `url.as_str()` — the *normalized* string. WHATWG parsing removes dot segments and surrounding controls/whitespace, so `https://example.com/` followed by `"a/../"` repeated 1000 times is a ~5 KB caller input that normalizes to 20 bytes and sails past the 2048-byte bound.

Issue #1287 lists length among the MDK URL rules to apply, so this leaves an explicit acceptance criterion unenforced at the public FFI boundary. The codebase already has the right pattern for this — `crates/traits/src/app_components/encrypted_media_v2.rs:227` checks `raw.len()` *before* parsing.

Check the trimmed raw string here before `Url::parse`, keep the post-parse check as-is, and add a regression using an over-limit dot-segment URL.

(Note this is pre-existing behavior in the shared validator, so `fetch_blossom_blob` has the same gap — but this PR is the one asserting the criterion, and a caller-supplied kind:0 `picture` value is a much more untrusted input than a locator you already parsed.)

crates/marmot-app/src/media/blossom.rs
```diff
@@ -214,6 +217,151 @@ pub(crate) async fn fetch_blossom_blob(
...
+        .await
+        .map_err(|_| AppError::BlobStore("request timed out".into()))?
+}
+
+#[cfg(test)]
+pub(super) async fn profile_operation_timeout_for_test(timeout: Duration) -> Result<(), AppError> {
+    profile_operation_timeout(timeout, std::future::pending()).await
+}
+
+enum ProfileResolveMode {
```
erskingardner: This `#[cfg(test)]` variant forces a `resolve_mode` parameter through `fetch_profile_image_impl`, `profile_media_pin`, and `media_http_client_for_profile` — three production signatures shaped by a test.

What it buys: `download_profile_image_rejects_injected_private_resolve_before_dial` (line 158 of the test file), which asserts that a loopback address in the resolved set is rejected. But `profile_image_pin_plan_rejects_private_injected_addresses_before_dial` already asserts exactly that against `plan_profile_image_pin` directly, without any of this plumbing.

Worse, the harness is structurally incapable of covering the thing that *is* missing: any injected address that isn't public gets rejected before a socket is opened, so it can never drive a successful fetch or a real redirect chain. It can only ever produce errors.

I'd delete `ProfileResolveMode`, `fetch_profile_image_with_injected_addrs`, `plan_profile_image_pin`, and `download_profile_image_with_injected_addrs`, restore `fetch_profile_image_impl(url, max_bytes)`, and get redirect + pinning coverage from a real local listener the way `super::tests` already does for the Blossom path.

crates/marmot-app/src/media/blossom.rs
```diff
@@ -214,6 +217,151 @@ pub(crate) async fn fetch_blossom_blob(
...
+    Production,
+    #[cfg(test)]
+    Injected(Vec<SocketAddr>),
+}
+
+async fn fetch_profile_image_impl(
```
erskingardner: This loop is a near-verbatim copy of `fetch_blossom_blob` (lines 170-218) — same status dispatch, same redirect-limit check, same `Location` extraction, same `join`, same four error strings. Roughly 50 duplicated lines whose only real differences are the validator, the byte cap, and the error variant used for parse failures.

That matters beyond tidiness: a fix to redirect handling in one now has to be remembered in the other, and the two already disagree on error classification for identical failures (see my note on the UniFFI mapping).

One loop taking the per-hop validator and the byte cap would collapse both. If you'd rather not touch the Blossom path in this PR, that's a defensible scope call — but then please leave a comment on both pointing at each other.

crates/marmot-app/src/media/blossom.rs
```diff
@@ -269,17 +417,30 @@ async fn media_http_client_for_url(
...
+        && url.host().map(is_loopback_host).unwrap_or(false);
+    let pin = resolve_media_host(url, allow_loopback).await?;
+    build_pinned_media_http_client(MEDIA_HTTP_TOTAL_TIMEOUT, pin)
+}
+
+fn build_pinned_media_http_client(
+    operation_timeout: Duration,
+    pin: Option<(String, Vec<SocketAddr>)>,
+) -> Result<reqwest::Client, AppError> {
+    let connect_timeout = operation_timeout.min(MEDIA_HTTP_CONNECT_TIMEOUT);
```
erskingardner: `operation_timeout` is `MEDIA_HTTP_TOTAL_TIMEOUT` at both call sites (line 424 and line 362), so these two `.min()` calls are constant-folded and the parameter is flexibility no caller asked for.

Drop the parameter and use the three constants directly; the function keeps doing exactly what it does today at fewer lines. If a per-call timeout is ever genuinely needed, add it then.

crates/marmot-app/src/media/tests_profile_image.rs
```diff
@@ -0,0 +1,256 @@
...
+            public,
+            SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 443),
+        ],
+    )
+    .expect_err("mixed public/private inject");
+    assert!(err.to_string().contains("public unicast"));
+}
+
+#[test]
+fn profile_image_pin_plan_returns_exact_vetted_set_for_one_hop() {
```
erskingardner: This test doesn't test rebind safety. `plan_profile_image_pin` is a pure function; calling it twice with different address slices and asserting the results differ only restates that it returns its input. It would pass against an implementation with no pinning at all.

Rebind safety is a property of the *dial* — that reqwest uses the vetted set via `resolve_to_addrs` instead of re-resolving. Nothing here observes that. As written it's false confidence on the exact property #1287 calls out.

Either cover it for real (local listener, pin, assert the connection lands on the pinned address) or drop it — a test that can't fail is worse than no test.

crates/marmot-app/src/media/tests_profile_image.rs
```diff
@@ -0,0 +1,256 @@
...
+        .expect("second pin");
+    assert_eq!(second_pin, rebound);
+    assert_ne!(
+        pinned, second_pin,
+        "rebound answers must not merge across plans"
+    );
+}
+
+#[tokio::test]
+async fn profile_image_operation_deadline_covers_the_whole_future() {
```
erskingardner: This asserts that `tokio::time::timeout` fires on `std::future::pending()` — that's tokio's invariant, not this PR's. The behavior worth pinning is that the deadline spans the *whole* operation including redirect hops, which is what the test name claims but `future::pending()` can't demonstrate.

If the redirect-loop integration test suggested in the review body gets added, it can cover this properly (slow multi-hop chain -> overall deadline trips). Then `profile_operation_timeout_for_test` can come out of `blossom.rs` too.

crates/marmot-app/src/media/tests_profile_image.rs
```diff
@@ -0,0 +1,256 @@
...
+        .await
+        .expect_err("chunked overflow");
+    assert!(
+        err.to_string()
+            .contains(&format!("download exceeds {caller_max} bytes")),
+        "{err}"
+    );
+}
+
+fn spawn_http_response(response: Vec<u8>) -> String {
```
erskingardner: These two wrappers just forward to `super::tests`, which is why those fns were widened to `pub(super)` in the first place. Replace both with `use super::tests::{http_ok_response, spawn_http_response};` at the top and delete these seven lines.

crates/marmot-uniffi/src/commands/account.rs
```diff
@@ -529,6 +529,27 @@ impl Marmot {
...
+        url: String,
+        max_bytes: u64,
+    ) -> Result<Vec<u8>, MarmotKitError> {
+        marmot_app::download_profile_image(url, max_bytes)
+            .await
+            .map_err(profile_image_fetch_error)
+    }
+}
+
+fn profile_image_fetch_error(error: marmot_app::AppError) -> MarmotKitError {
```
erskingardner: Two failures of the same kind land on different FFI error types: a redirect to a private *host* raises `InvalidAppMessagePayload` -> `InvalidMediaReference`, while a public hostname that *resolves* to a private address raises `BlobStore` -> whatever `AppError::into` yields. Both are SSRF rejections, and a host app filtering on `InvalidMediaReference` to distinguish "bad URL, don't retry" from "network hiccup, retry later" will misclassify the DNS case as transient and retry it forever.

Worth making all pre-dial safety rejections map to `InvalidMediaReference`. That likely means a dedicated `AppError` variant for unsafe-host rejections rather than reusing `BlobStore` — which would also read better than `InvalidAppMessagePayload`, since a kind:0 `picture` URL isn't an app message payload.

erskingardner (changes requested):
The SSRF core here is correct: validate URL -> resolve -> vet every resolved address -> pin the vetted set -> dial with `Policy::none()` -> revalidate every redirect hop. I traced that path and did not find a way through it. No objection to the premise either; #1287 makes a good case that MDK should own this boundary rather than each host app.

My concerns are that two of the issue's explicit acceptance criteria are not actually met, and that a sizeable amount of test-only machinery has been threaded into production signatures while buying very little coverage.

**Should fix before merge**

1. The URL length rule is unenforced at the public boundary. `Url::parse` normalizes before `validate_blossom_fetch_url` measures, so an over-limit input is accepted. This is the same finding @agent-p1p posted; it still reproduces at this head. `crates/traits/src/app_components/encrypted_media_v2.rs:227` already does it the right way (`raw.len()` pre-parse) — follow that precedent.
2. No test enters `fetch_profile_image_impl`'s redirect loop. Per-hop revalidation and the hop cap are the security-critical parts of this change and the parts most likely to regress under future dedup with the Blossom path, and both are currently covered only by calling the validators directly on hand-built `Url` values.

**Simplification**

The `ProfileResolveMode` enum threads a `#[cfg(test)]` variant through three production functions, and the two tests it enables assert only that `reject_non_public_ip` gets called — which `plan_profile_image_pin` already shows without it. Because injected addresses that aren't public are rejected by construction, this harness can only ever produce rejections; it can't build the redirect test that's actually missing. That's a sign it's the wrong shape. I'd drop it and get the coverage from a real local server, which is what `super::tests` already does for Blossom.

Also flagged inline: a tautological pinning test, a test that exercises `tokio::time::timeout` rather than this code, a dead `operation_timeout` parameter, ~50 lines duplicated verbatim from `fetch_blossom_blob`, and two pointless wrapper fns. Net, I think this lands closer to the issue's intent at meaningfully fewer lines.

Nothing here is exploitable — happy to re-review quickly once (1) and (2) are addressed.

Datawav:

No actionable findings.

_Advisory audit of head [`5a05642fd9bf`](https://github.com/marmot-protocol/mdk/commit/5a05642fd9bf3ab5ff4f7af3252758bab3ea3467); no checkout or PR-code execution was performed._

Datawav (dismissed):
Automated multi-agent audit of head 5a05642fd9bf: no actionable findings. Ready for merge.

crates/marmot-app/src/media/blossom.rs
agent-p1p: Addressed in faa1047 and f8a41fd3. Initial and redirect inputs are bounded on the original raw string before trimming or WHATWG normalization, while the post-parse bound remains. Added dot-segment and whitespace-padding regressions for both boundaries.

crates/marmot-app/src/media/blossom.rs
agent-p1p: Addressed in faa1047. Removed ProfileResolveMode, fetch_profile_image_with_injected_addrs, plan_profile_image_pin, download_profile_image_with_injected_addrs, profile_media_pin, and all injected-address plumbing. Production fetch_profile_image_impl(url, max_bytes) and its client path have no test-policy parameters. Real localhost-domain listener tests exercise DNS vetting and resolve_to_addrs pinning through a cfg(test)-only wrapper around the shared engine.

crates/marmot-app/src/media/blossom.rs
agent-p1p: Addressed in faa1047. Blossom and profile-image downloads now use one shared bounded manual redirect loop; callers provide only their client/pinning and redirect-policy differences. Blossom host-affinity and error policy remain unchanged.

crates/marmot-app/src/media/blossom.rs
agent-p1p: Addressed in faa1047. Removed operation_timeout from build_pinned_media_http_client and use the connect/read/total constants directly.

crates/marmot-app/src/media/tests_profile_image.rs
agent-p1p: Addressed in faa1047. Deleted the tautological pin-plan test and helper. The success integration uses localhost (not an IP literal), so the request resolves, vets every answer, and reaches the listener through resolve_to_addrs pinning.

crates/marmot-app/src/media/tests_profile_image.rs
agent-p1p: Addressed in faa1047. Deleted profile_operation_timeout_for_test and the pending-future tautology. The production 60-second whole-operation deadline remains around the shared redirect operation; no fake Tokio invariant test remains.

crates/marmot-app/src/media/tests_profile_image.rs
agent-p1p: Addressed in faa1047. Tests import the widened super::tests helpers directly; the forwarding wrappers are gone.

crates/marmot-uniffi/src/commands/account.rs
agent-p1p: Addressed in faa1047. Added AppError::UnsafeMediaFetch for all pre-dial profile safety rejections and map it through UniFFI to InvalidMediaReference. DNS lookup and transport failures remain BlobStore/runtime-retryable. Added focused app and UniFFI classification tests.

crates/marmot-app/src/media/blossom.rs
```diff
@@ -269,6 +435,16 @@ async fn media_http_client_for_url(
...
         .map_err(|err| AppError::BlobStore(format!("unsafe Blossom URL: {err}")))?;
+    let allow_loopback = url.scheme() == "http"
+        && allow_loopback_http
+        && url.host().map(is_loopback_host).unwrap_or(false);
+    let pin = resolve_media_host(url, allow_loopback).await?;
+    build_pinned_media_http_client(pin)
+}
+
+fn build_pinned_media_http_client(
+    pin: Option<(String, Vec<SocketAddr>)>,
```
Datawav, line 446: This pinned client still honors reqwest's configured HTTP(S) proxies. When a proxy is active, it receives and independently resolves the hostname, bypassing `resolve_to_addrs` and the vetted public-address pin. Add `.no_proxy()` to this security-sensitive builder (and a proxy-config regression test) so every profile-image request and redirect hop connects only to the locally vetted addresses. Since this builder is shared, confirm that disabling proxies is also intended for Blossom downloads.

Datawav (commented):

### High — proxy configuration bypasses DNS vetting and address pinning

`build_pinned_media_http_client` configures `resolve_to_addrs` but does not call `reqwest::ClientBuilder::no_proxy()`. Reqwest can therefore honor configured HTTP(S) proxies. When proxied, the proxy receives and independently resolves the hostname, so the connection is not guaranteed to use the locally vetted public addresses. This breaks the advertised SSRF boundary and “pinned public resolution” guarantee.

**Need:** Untrusted profile-image URLs must reach only the IP addresses vetted for each request and redirect hop.

**Design:** Disable proxy discovery on this security-sensitive client with `.no_proxy()`. Add a regression test that sets proxy configuration and verifies the request still connects directly to the pinned address. Because the builder is shared, confirm whether disabling proxies is also the intended policy for existing Blossom downloads.

**Alternatives:** Use a dedicated proxy implementation that receives only the vetted destination address while preserving the original TLS hostname, or reject operation whenever proxy configuration is present. Both are more complex than disabling proxies.

_Advisory audit of head [`f8a41fd391cd`](https://github.com/marmot-protocol/mdk/commit/f8a41fd391cdb2e96e59ed3832b9c02e7e253d91); no approval, merge, push, checkout, or PR-code execution was performed._

crates/marmot-app/src/media/blossom.rs
agent-p1p: Addressed in signed head e1d0cd0e. The shared pinned client already called `.no_proxy()` at the reviewed head, so profile-image and Blossom downloads were both proxy-disabled. Added a deterministic regression that configures an explicit proxy before the shared builder policy, pins a synthetic hostname to a direct listener, and proves the request reaches the vetted origin rather than the proxy. The existing Unreleased entry now states the proxy-disabled guarantee.

agent-p1p:
## Adversarial security re-review — PASS

Reviewed exact signed head `e1d0cd0e98b70ef6c0e044e75a7a09f0d9024d87` against base `2e02e0d8bd583c80e0059c1616054efc1c35d42e`.

**Blocking findings: none.**

- `crates/marmot-app/src/media/blossom.rs:340-342,432-442,451-470` routes both profile-image and Blossom downloads through the same pinned client builder. The builder applies `.no_proxy()` before `resolve_to_addrs`; reqwest 0.12.28 documents that this clears explicit proxies and disables automatic system proxies.
- The shared builder preserves the required sequence: URL validation, resolution, rejection of every non-public answer, pinning of the vetted addresses, manual redirects, and a fresh validation/resolution/pin for every hop (`blossom.rs:249-269,274-320,345-380`). I found no DNS-rebinding or SSRF bypass introduced by the proxy-test refactor.
- `pinned_media_client_ignores_configured_proxy` is a meaningful regression: it explicitly installs a proxy, pins a synthetic hostname to a separate direct listener, requests that hostname, and asserts the direct listener body rather than the proxy body (`tests_profile_image.rs:231-257`).
- The prior raw-input and redirect-loop findings remain resolved in production (`host_safety.rs:74-104`). The current CodeRabbit trim-order note is limited to the `#[cfg(test)]` loopback harness and does not weaken the production boundary; production pre-trim checks have direct regressions.

Independent verification:
- `cargo test -p marmot-app --locked --lib pinned_media_client_ignores_configured_proxy -- --nocapture` — 1 passed.
- `cargo test -p marmot-app --locked --lib profile_image` — 25 passed.
- `cargo test -p marmot-uniffi --locked profile_image` — 2 passed.
- `cargo fmt --all -- --check` — passed.
- `cargo clippy -p marmot-app -p marmot-uniffi --locked --all-targets -- -D warnings` — passed.
- `git diff --check 2e02e0d8...e1d0cd0e` — passed.

**Verdict:** PASS with high confidence. Proxy-policy feedback is resolved. No approval or merge performed.

PR #1292, opened 2026-08-07
Add transcript and secrecy scenario vectors from the multi-VM catalog

dannym-arx:
## Summary

Ports four scenarios from the external multi-VM harness catalog into portable ScenarioSpec v2 vector fixtures. Marmots now prove not only who heard what, but who never could.

- `conversation/v1`: multi-round turn-taking chat with exact per-client transcripts, convergence, and no pending work
- `latecomer-forward-secrecy/v1`: the simulator's first forward-secrecy check. The latecomer holds pre-join ciphertext and must never observe pre-join plaintext.
- `leaver-removal-secrecy/v1`: the first `remove_members` vector. Post-removal and post-self-leave secrecy are pinned with `payload_count` zero assertions.
- `incremental-growth/v1`: grows a group one member per phase with a rename per phase, and pins cohort-era secrecy across two join boundaries
- registers all four in `vectors/manifest.v1.json` and `SCENARIOS.md`

## Stack

Continues the #1250 extraction series: #1270 (merged), #1271 (merged), #1272 (base of this PR). This PR targets `codex/convergence-failure-corpus` and re-targets `master` once #1272 merges. Planned follow-ups in this series: isolation and catch-up vectors, then a seeded churn family.

## Scope boundary

No production code changes; this PR adds JSON fixtures and docs only. The catalog's re-add leg (a removed member rejoins with a fresh KeyPackage) is absent on purpose: welcome-after-eviction fails in the engine harness with `GroupStateError(UseAfterEviction)`. SCENARIOS.md records that gap so the leg can land once the harness supports it.

## Verification

- `cargo test -p cgka-conformance-simulator --locked --test canonical_scenarios canonical_vector_fixtures_match_generated_traces`
- `cargo test -p cgka-conformance-simulator --locked --test vector_artifacts`
- `cargo test -p cgka-conformance-simulator --locked --test scenario_ir`
- `just fast-ci` (one caveat: `convergence-campaign-runner/src/manifest.rs:469` on the base branch trips clippy 1.97's `unneeded_wildcard_pattern`; CI pins 1.95, where it does not fire)

## Summary by CodeRabbit

* **New Features**
  * Added portable conformance scenarios for multi-round conversations, incremental group growth, latecomer forward secrecy, and member-removal secrecy.
  * Added coverage for invitations, renames, membership changes, message delivery, and multi-epoch group evolution.
  * Added expected results for payload visibility, state convergence, confirmed operations, and completion without pending work.
  * Updated the scenario manifest to include all four new vectors.

### DIFF
--- a/crates/cgka-conformance-simulator/vectors/conversation.v1.json
+++ b/crates/cgka-conformance-simulator/vectors/conversation.v1.json
@@ -6,36 +6,107 @@
   "scenario": {
     "name": "conversation/v1",
     "spec_version": "2",
-    "clients": ["alice", "bob", "carol"],
+    "clients": [
+      "alice",
+      "bob",
+      "carol"
+    ],
     "steps": [
       {
         "type": "create_group",
         "creator": "alice",
         "name": "conversation",
-        "invitees": ["bob", "carol"],
+        "invitees": [
+          "bob",
+          "carol"
+        ],
         "required_features": [],
         "pending": "create"
       },
-      { "type": "acknowledge_outbound", "client": "alice", "publication": "create", "outcome": "accepted" },
-      { "type": "deliver_all" },
-      { "type": "tick", "clients": ["bob", "carol"] },
-      { "type": "clear_events", "clients": ["alice", "bob", "carol"] },
-      { "type": "send_app_message", "sender": "alice", "payload": "conversation:r1:alice" },
-      { "type": "send_app_message", "sender": "bob", "payload": "conversation:r1:bob" },
-      { "type": "send_app_message", "sender": "carol", "payload": "conversation:r1:carol" },
-      { "type": "deliver_all" },
-      { "type": "tick", "clients": ["alice", "bob", "carol"] },
-      { "type": "send_app_message", "sender": "alice", "payload": "conversation:r2:alice" },
-      { "type": "send_app_message", "sender": "bob", "payload": "conversation:r2:bob" },
-      { "type": "send_app_message", "sender": "carol", "payload": "conversation:r2:carol" },
-      { "type": "deliver_all" },
-      { "type": "tick", "clients": ["alice", "bob", "carol"] },
-      { "type": "send_app_message", "sender": "alice", "payload": "conversation:r3:alice" },
-      { "type": "send_app_message", "sender": "bob", "payload": "conversation:r3:bob" },
-      { "type": "send_app_message", "sender": "carol", "payload": "conversation:r3:carol" },
-      { "type": "deliver_all" },
-      { "type": "tick", "clients": ["alice", "bob", "carol"] },
-      { "type": "observe_exact", "clients": ["alice", "bob", "carol"] }
+      {
+        "type": "acknowledge_outbound",
+        "client": "alice",
+        "publication": "create",
+        "outcome": "accepted"
+      },
+      {
+        "type": "deliver_all"
+      },
+      {
+        "type": "tick",
+        "clients": [
+          "bob",
+          "carol"
+        ]
+      },
+      {
+        "type": "clear_events",
+        "clients": [
+          "alice",
+          "bob",
+          "carol"
+        ]
+      },
+      {
+        "type": "send_app_message",
+        "sender": "alice",
+        "payload": "conversation:r1:alice"
+      },
+      {
+        "type": "send_app_message",
+        "sender": "bob",
+        "payload": "conversation:r1:bob"
+      },
+      {
+        "type": "send_app_message",
+        "sender": "carol",
+        "payload": "conversation:r1:carol"
+      },
+      {
+        "type": "deliver_all"
+      },
+      {
+        "type": "tick",
+        "clients": [
+          "alice",
+          "bob",
+          "carol"
+        ]
+      },
+      {
+        "type": "send_app_message",
+        "sender": "alice",
+        "payload": "conversation:r2:alice"
+      },
+      {
+        "type": "send_app_message",
+        "sender": "bob",
+        "payload": "conversation:r2:bob"
+      },
+      {
+        "type": "send_app_message",
+        "sender": "carol",
+        "payload": "conversation:r2:carol"
+      },
+      {
+        "type": "deliver_all"
+      },
+      {
+        "type": "tick",
+        "clients": [
+          "alice",
+          "bob",
+          "carol"
+        ]
+      },
+      {
+        "type": "observe_exact",
+        "clients": [
+          "alice",
+          "bob",
+          "carol"
+        ]
+      }
     ]
   },
   "expected_outcomes": [
@@ -55,9 +126,7 @@
         "conversation:r1:bob",
         "conversation:r1:carol",
         "conversation:r2:bob",
-        "conversation:r2:carol",
-        "conversation:r3:bob",
-        "conversation:r3:carol"
+        "conversation:r2:carol"
       ]
     },
     {
@@ -69,9 +138,7 @@
         "conversation:r1:alice",
         "conversation:r1:carol",
         "conversation:r2:alice",
-        "conversation:r2:carol",
-        "conversation:r3:alice",
-        "conversation:r3:carol"
+        "conversation:r2:carol"
       ]
     },
     {
@@ -83,20 +150,26 @@
         "conversation:r1:alice",
         "conversation:r1:bob",
         "conversation:r2:alice",
-        "conversation:r2:bob",
-        "conversation:r3:alice",
-        "conversation:r3:bob"
+        "conversation:r2:bob"
       ]
     },
     {
       "type": "clients_converged",
-      "clients": ["alice", "bob", "carol"],
+      "clients": [
+        "alice",
+        "bob",
+        "carol"
+      ],
       "epoch": 1,
       "member_count": 3
     },
     {
       "type": "no_pending_work",
-      "clients": ["alice", "bob", "carol"]
+      "clients": [
+        "alice",
+        "bob",
+        "carol"
+      ]
     }
   ]
 }
--- a/crates/cgka-conformance-simulator/vectors/conversation.v1.json
+++ b/crates/cgka-conformance-simulator/vectors/conversation.v1.json
@@ -1,0 +1,102 @@
+{
+  "scenario_name": "conversation/v1",
+  "vector_version": "1",
+  "conformance_version": "0.9.10",
+  "seed": null,
+  "scenario": {
+    "name": "conversation/v1",
+    "spec_version": "2",
+    "clients": ["alice", "bob", "carol"],
+    "steps": [
+      {
+        "type": "create_group",
+        "creator": "alice",
+        "name": "conversation",
+        "invitees": ["bob", "carol"],
+        "required_features": [],
+        "pending": "create"
+      },
+      { "type": "acknowledge_outbound", "client": "alice", "publication": "create", "outcome": "accepted" },
+      { "type": "deliver_all" },
+      { "type": "tick", "clients": ["bob", "carol"] },
+      { "type": "clear_events", "clients": ["alice", "bob", "carol"] },
+      { "type": "send_app_message", "sender": "alice", "payload": "conversation:r1:alice" },
+      { "type": "send_app_message", "sender": "bob", "payload": "conversation:r1:bob" },
+      { "type": "send_app_message", "sender": "carol", "payload": "conversation:r1:carol" },
+      { "type": "deliver_all" },
+      { "type": "tick", "clients": ["alice", "bob", "carol"] },
+      { "type": "send_app_message", "sender": "alice", "payload": "conversation:r2:alice" },
+      { "type": "send_app_message", "sender": "bob", "payload": "conversation:r2:bob" },
+      { "type": "send_app_message", "sender": "carol", "payload": "conversation:r2:carol" },
+      { "type": "deliver_all" },
+      { "type": "tick", "clients": ["alice", "bob", "carol"] },
+      { "type": "send_app_message", "sender": "alice", "payload": "conversation:r3:alice" },
+      { "type": "send_app_message", "sender": "bob", "payload": "conversation:r3:bob" },
+      { "type": "send_app_message", "sender": "carol", "payload": "conversation:r3:carol" },
+      { "type": "deliver_all" },
+      { "type": "tick", "clients": ["alice", "bob", "carol"] },
+      { "type": "observe_exact", "clients": ["alice", "bob", "carol"] }
+    ]
+  },
+  "expected_outcomes": [
+    {
+      "type": "pending_resolution",
+      "step_index": 1,
+      "client": "alice",
+      "pending": "create",
+      "resolution": "confirmed"
+    },
+    {
+      "type": "client_state",
+      "client": "alice",
+      "epoch": 1,
+      "member_count": 3,
+      "received_payloads": [
+        "conversation:r1:bob",
+        "conversation:r1:carol",
+        "conversation:r2:bob",
+        "conversation:r2:carol",
+        "conversation:r3:bob",
+        "conversation:r3:carol"
+      ]
+    },
+    {
+      "type": "client_state",
+      "client": "bob",
+      "epoch": 1,
+      "member_count": 3,
+      "received_payloads": [
+        "conversation:r1:alice",
+        "conversation:r1:carol",
+        "conversation:r2:alice",
+        "conversation:r2:carol",
+        "conversation:r3:alice",
+        "conversation:r3:carol"
+      ]
+    },
+    {
+      "type": "client_state",
+      "client": "carol",
+      "epoch": 1,
+      "member_count": 3,
+      "received_payloads": [
+        "conversation:r1:alice",
+        "conversation:r1:bob",
+        "conversation:r2:alice",
+        "conversation:r2:bob",
+        "conversation:r3:alice",
+        "conversation:r3:bob"
+      ]
+    },
+    {
+      "type": "clients_converged",
+      "clients": ["alice", "bob", "carol"],
+      "epoch": 1,
+      "member_count": 3
+    },
+    {
+      "type": "no_pending_work",
+      "clients": ["alice", "bob", "carol"]
+    }
+  ]
+}
--- a/crates/cgka-conformance-simulator/vectors/incremental-growth.v1.json
+++ b/crates/cgka-conformance-simulator/vectors/incremental-growth.v1.json
@@ -1,0 +1,197 @@
+{
+  "scenario_name": "incremental-growth/v1",
+  "vector_version": "1",
+  "conformance_version": "0.9.10",
+  "seed": null,
+  "scenario": {
+    "name": "incremental-growth/v1",
+    "spec_version": "2",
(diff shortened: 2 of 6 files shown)


Datawav:

## High — Removal-secrecy checks do not exercise ciphertext confidentiality

In `leaver-removal-secrecy.v1.json`, Dave is excluded by `set_partition` before the gap messages are delivered, and Carol is never ticked after the tail messages are delivered. Their zero-count assertions therefore pass because the clients do not process the tested ciphertext—not because removal prevents decryption. The assertions are also incomplete: Dave is not checked against Carol’s gap message, and Carol is not checked against Bob’s tail message.

Need: The vector is presented in `SCENARIOS.md` and `manifest.v1.json` as proving post-removal/post-leave secrecy—“who never could”—so removed members must actually receive/process the relevant post-departure ciphertext. Otherwise a regression that leaves departed members able to decrypt would go undetected.

Design: Keep Dave transport-visible through the gap phase and tick him after delivery; likewise tick Carol after tail delivery. Assert zero observations for every gap payload sent after Dave’s removal and every tail payload sent after Carol’s leave. Exact departed-client outcomes would provide stronger transcript coverage than sampled assertions.

Alternatives: If the simulator intentionally cannot deliver ciphertext to evicted clients, narrow the scenario name and documentation to transport-unsubscription/isolation coverage and add a separate confidentiality vector when adversarial ciphertext delivery is supported.

_Advisory audit of head [`8d3ae41ac1ec`](https://github.com/marmot-protocol/mdk/commit/8d3ae41ac1ec8de88073f4c6f843a6bb41316e62); no approval, merge, push, checkout, or PR-code execution was performed._

crates/cgka-conformance-simulator/vectors/leaver-removal-secrecy.v1.json
```diff
@@ -0,0 +1,134 @@
...
+      {
+        "type": "remove_members",
+        "remover": "alice",
+        "members": ["dave"],
+        "pending": "remove-dave"
+      },
+      { "type": "acknowledge_outbound", "client": "alice", "publication": "remove-dave", "outcome": "accepted" },
+      { "type": "deliver_all" },
+      { "type": "tick", "clients": ["bob", "carol", "dave"] },
+      { "type": "set_partition", "allow": ["alice", "bob", "carol"] },
```
erskingardner: Blocking: this partition makes Dave's two `leaver:gap:*` zero assertions vacuous.

`TransportBus::step` skips any client outside the allow-set entirely (`src/bus.rs`), so after this step the gap ciphertext is never written to Dave's mailbox. `payload_count` reads `pending_payload_count`, which only filters already-emitted `MessageReceived` events — with no ciphertext delivered and no tick, the count is 0 no matter what the engine does with post-removal keys. The assertion would still pass if removal leaked the epoch secret.

Compare `latecomer-forward-secrecy.v1.json`, which gets this right: Carol is bus-attached from construction, the pre-join ciphertext really lands in her mailbox, and she is really ticked before the zero assertions. That is a forward-secrecy proof; this is a delivery-suppression proof.

Suggest keeping Dave in the allow-set through the gap phase and ticking him after `deliver_all`, so the engine has to refuse the decrypt. If the intent really is to model the transport unsubscribing a removed member, that is a legitimate thing to pin — but then it is isolation coverage, not secrecy coverage, and the manifest/SCENARIOS wording should say so.

crates/cgka-conformance-simulator/vectors/leaver-removal-secrecy.v1.json
```diff
@@ -0,0 +1,134 @@
...
+      { "type": "leave", "client": "carol" },
+      { "type": "deliver_all" },
+      { "type": "tick", "clients": ["alice"] },
+      { "type": "acknowledge_outbound", "client": "alice", "outcome": "accepted" },
+      { "type": "deliver_all" },
+      { "type": "tick", "clients": ["bob", "carol"] },
+      { "type": "send_app_message", "sender": "alice", "payload": "leaver:tail:alice" },
+      { "type": "send_app_message", "sender": "bob", "payload": "leaver:tail:bob" },
+      { "type": "deliver_all" },
+      { "type": "tick", "clients": ["alice", "bob"] },
```
erskingardner: Blocking, and cheaper to fix than the Dave case: Carol is inside the partition allow-set, so `leaver:tail:alice` and `leaver:tail:bob` *do* reach her mailbox — she is just never ticked again after this delivery. Her `payload_count: 0` assertion below therefore passes on an unprocessed mailbox rather than on a failed decrypt.

Adding `carol` here (or a following `{ "type": "tick", "clients": ["carol"] }`) turns it into a real post-leave secrecy check. Worth trying before assuming the harness refuses: if the tick errors for a self-removed client, `ensure_tick_succeeded` will surface it, and that error is itself the finding to document rather than route around.

While you are here: the assertions only sample `leaver:tail:alice` for Carol and `leaver:gap:{alice,bob}` for Dave. Dave is never checked against `leaver:gap:carol`, and Carol is never checked against `leaver:tail:bob`. Once the departed clients actually process the ciphertext, pin every post-departure payload.

crates/cgka-conformance-simulator/vectors/manifest.v1.json
```diff
@@ -352,6 +352,58 @@
...
+      "next": "Add a multi-epoch pre-join history variant once deferred-input expiry is modeled."
+    },
+    {
+      "id": "leaver-removal-secrecy/v1",
+      "kind": "scenario_vector",
+      "status": "portable",
+      "artifact": "leaver-removal-secrecy.v1.json",
+      "coverage": [
+        "admin removal commit",
+        "post-removal secrecy for the removed member",
```
erskingardner: These two coverage lines (and the matching `SCENARIOS.md` claim that "Dave never observes gap plaintext and Carol never observes tail plaintext") are what makes the vacuous assertions a correctness problem rather than a weak test: a future reader greps the manifest, sees secrecy is covered, and does not re-derive it. Either make the assertions real or narrow these strings to what is actually pinned (transport unsubscription / no post-departure delivery).

crates/cgka-conformance-simulator/vectors/conversation.v1.json
```diff
@@ -0,0 +1,102 @@
...
+      { "type": "acknowledge_outbound", "client": "alice", "publication": "create", "outcome": "accepted" },
+      { "type": "deliver_all" },
+      { "type": "tick", "clients": ["bob", "carol"] },
+      { "type": "clear_events", "clients": ["alice", "bob", "carol"] },
+      { "type": "send_app_message", "sender": "alice", "payload": "conversation:r1:alice" },
+      { "type": "send_app_message", "sender": "bob", "payload": "conversation:r1:bob" },
+      { "type": "send_app_message", "sender": "carol", "payload": "conversation:r1:carol" },
+      { "type": "deliver_all" },
+      { "type": "tick", "clients": ["alice", "bob", "carol"] },
+      { "type": "send_app_message", "sender": "alice", "payload": "conversation:r2:alice" },
```
erskingardner: Non-blocking: rounds 2 and 3 traverse exactly the same code path as round 1 — same epoch, same three senders, same ordered delivery, no fault injected between them. `three-client-message-exchange/v1` already pins the single-round shape. What `conversation/v1` genuinely adds over it is `observe_exact` (canonical snapshot + input ledger) and `no_pending_work`, not the extra rounds.

Two rounds would still prove the "no duplicate across rounds" property that one round cannot; the third is ~15 lines of fixture for no new state. Or consider whether upgrading the existing smoke vector to `observe_exact` covers the same ground without a second near-duplicate fixture.

crates/cgka-conformance-simulator/vectors/conversation.v1.json
```diff
@@ -0,0 +1,102 @@
...
+    "steps": [
+      {
+        "type": "create_group",
+        "creator": "alice",
+        "name": "conversation",
+        "invitees": ["bob", "carol"],
+        "required_features": [],
+        "pending": "create"
+      },
+      { "type": "acknowledge_outbound", "client": "alice", "publication": "create", "outcome": "accepted" },
```
erskingardner: Non-blocking style: every existing fixture in `vectors/` is fully expanded (one JSON field per line, arrays broken out) — see `three-client-message-exchange.v1.json` and `partition-clear-leave.v1.json`. All four new fixtures use hand-compacted single-line step objects. It reads fine, but it means `git diff` on a future edit here looks different from every other vector, and any round-trip through `serde_json::to_string_pretty` would reformat the whole file. Worth matching the existing shape unless the compaction is deliberate.

erskingardner (changes requested):
Fixture-only PR, CI green, and three of the four vectors do what the description says. `latecomer-forward-secrecy/v1` and `incremental-growth/v1` are genuinely good: the joiner is bus-attached from scenario construction, so the pre-join ciphertext really lands in their mailbox, and they are really ticked before the `payload_count: 0` assertions run. Those zeros mean something.

`leaver-removal-secrecy/v1` is the exception, and it is the one the summary leans on hardest ("who never could"). Both of its secrecy claims are structurally vacuous:

- Dave is partitioned out *before* the gap messages are sent, so the gap ciphertext never reaches his mailbox at all.
- Carol does receive the tail ciphertext (the partition still allows her) but is never ticked after it is delivered, so `pending_payload_count` has nothing to count.

In both cases the assertion passes because no decryption was ever attempted, not because removal prevented one. A regression that let departed members decrypt post-departure traffic would sail through this vector, while `SCENARIOS.md` and `manifest.v1.json` advertise it as post-removal / post-leave secrecy coverage. The Carol half is a one-line fix; the Dave half needs the partition moved or dropped. If ticking an evicted client turns out to error in the harness, that is worth recording as the real coverage gap (next to the existing `UseAfterEviction` note) and narrowing the vector's advertised coverage to transport isolation — but it should not be left looking like a proof it isn't.

The other three vectors and the docs/manifest wiring look right, and the deliberate omission of the re-add leg with the `UseAfterEviction` rationale is the right call. Two non-blocking nits inline.

An independent automated audit on an earlier head flagged the same `leaver` issue; I reached it separately by reading `bus.rs` delivery and `pending_payload_count`, and it holds on `df428e6a`.

Datawav (commented):

## Final audit

### Findings

- **[P2] `leaver-removal-secrecy/v1` does not test the claimed post-removal cryptographic secrecy.** After Dave processes the removal, `set_partition` excludes him before gap messages are sent. Consequently, Dave receives no post-removal ciphertext, and the zero-count assertions demonstrate transport isolation only—not inability to decrypt ciphertext after removal. This contradicts the scenario documentation and manifest claim of “post-removal secrecy.” Deliver gap ciphertext to Dave while keeping him removed from the group, then assert that no plaintext is produced.

- **[P2] The same vector incompletely checks its secrecy boundary.** Dave is checked against `leaver:gap:alice` and `leaver:gap:bob`, but not `leaver:gap:carol`; Carol is checked against `leaver:tail:alice`, but not `leaver:tail:bob`. Because neither departed client has an exact final observation/state expectation, those two plaintext leaks would pass. Add zero-count assertions for every protected payload, or provide exact transcript expectations for Dave and Carol.

- **[P3] `incremental-growth/v1` silently excludes Carol and Dave from `no_pending_work`.** The likely reason is retained undecryptable pre-join input, as explicitly documented for the latecomer vector, but this scenario does not explain the exclusion. Document the intended deferred-work behavior so the expectation is distinguishable from an unexamined pending-work defect.

### Need, design, and alternatives

The new fixtures address a real coverage need: exact multi-round transcripts, staged membership growth, and confidentiality across join/removal boundaries. The portable JSON-vector design is consistent with the existing ScenarioSpec approach, and the expected epochs, memberships, pending-resolution indexes, and active-member transcripts are internally coherent.

For removal secrecy, transport partitioning is insufficient because it prevents the adversarial condition the vector claims to test. The strongest design is to deliver post-removal ciphertext to the removed client and assert non-disclosure for the complete protected transcript. If the simulator cannot safely route ciphertext to an unsubscribed client, either add an explicit ciphertext-injection/replay operation or narrow the vector and manifest wording to “transport isolation,” leaving cryptographic post-removal secrecy as a documented coverage gap.

_Advisory audit of head [`df428e6a7839`](https://github.com/marmot-protocol/mdk/commit/df428e6a783926a87113297fb504195f00698d07); no approval, merge, push, checkout, or PR-code execution was performed._

crates/cgka-conformance-simulator/vectors/leaver-removal-secrecy.v1.json
Datawav: This partition prevents Dave from receiving the gap ciphertext, so the later zero-count assertions establish transport isolation rather than post-removal cryptographic secrecy. Deliver the ciphertext to Dave while he remains removed and assert that it produces no plaintext, or narrow the scenario's documented claim to transport isolation.

crates/cgka-conformance-simulator/vectors/leaver-removal-secrecy.v1.json
```diff
@@ -0,0 +1,134 @@
...
+        }
+      },
+      {
+        "type": "assert",
+        "assertion": {
+          "mode": "exactly",
+          "predicate": { "type": "payload_count", "client": "carol", "payload": "leaver:tail:alice", "count": 0 }
+        }
+      },
+      { "type": "observe_exact", "clients": ["alice", "bob"] }
```
Datawav: Only Alice and Bob receive exact transcript checks, leaving gaps in the departed-client secrecy assertions: Dave is never checked for `leaver:gap:carol`, and Carol is never checked for `leaver:tail:bob`. Add zero-count assertions for both payloads or exact final observations for Dave and Carol.

crates/cgka-conformance-simulator/vectors/incremental-growth.v1.json
```diff
@@ -0,0 +1,197 @@
...
+      ]
+    },
+    {
+      "type": "clients_converged",
+      "clients": ["alice", "bob", "carol", "dave"],
+      "epoch": 5,
+      "member_count": 4
+    },
+    {
+      "type": "no_pending_work",
```
Datawav, line 369: `no_pending_work` silently excludes Carol and Dave. If they intentionally retain undecryptable pre-join input, document that deferred-work behavior as the latecomer vector does; otherwise this expectation can conceal an unexamined pending-work defect.

crates/cgka-conformance-simulator/vectors/leaver-removal-secrecy.v1.json
dannym-arx: Fixed in 0cdaf0e7 by dropping the partition entirely: Dave stays bus-attached, the gap (and now tail) ciphertext really lands in his mailbox, and he is ticked after each delivery. Ticking an evicted client works today — the `UseAfterEviction` failure is welcome-after-eviction only — so the zero assertions now run against processed ciphertext the engine had to refuse. Verified by mutation: flipping the `leaver:gap:alice` expectation to 1 fails the vector runner.

crates/cgka-conformance-simulator/vectors/leaver-removal-secrecy.v1.json
dannym-arx: Done in 0cdaf0e7: Carol (and Dave) are added to the tick after the tail `deliver_all`, and the assertions now pin every post-departure payload — Dave against gap:{alice,bob,carol} + tail:{alice,bob}, Carol against tail:{alice,bob}. The tick does not error for the self-removed client. Mutation-checked from Carol's side too: flipping `leaver:tail:bob` to 1 fails the runner.

crates/cgka-conformance-simulator/vectors/manifest.v1.json
dannym-arx: With the partition gone and the departed clients ticked over the delivered ciphertext, the secrecy claims are now real; 0cdaf0e7 also tightens the two coverage strings to say exactly what is pinned ("receives and processes all post-removal/post-leave ciphertext, decrypts none") and updates the SCENARIOS.md entry to match.

crates/cgka-conformance-simulator/vectors/conversation.v1.json
dannym-arx: Taken in 0cdaf0e7: trimmed to two rounds, which still pins the no-duplicate-across-rounds property the smoke vector cannot; SCENARIOS.md now records that the vector's value over `three-client-message-exchange/v1` is `observe_exact` + `no_pending_work`, not round count.

crates/cgka-conformance-simulator/vectors/conversation.v1.json
dannym-arx: Done in 0cdaf0e7 — all four new fixtures reformatted to the fully expanded style of the existing vectors (the compaction was not deliberate).

dannym-arx:
Review feedback addressed in 0cdaf0e7 (rebased on the force-updated base first):

- **`leaver-removal-secrecy/v1` is now a real secrecy vector.** The partition is gone; Dave and Carol stay bus-attached, the gap/tail ciphertext genuinely lands in their mailboxes, and both are ticked after each delivery. Ticking an evicted client does not error — the `UseAfterEviction` gap is welcome-after-eviction only — so the engine actually has to refuse the decrypts.
- **Complete secrecy boundary.** Every post-departure payload is now pinned: Dave against `gap:{alice,bob,carol}` + `tail:{alice,bob}`, Carol against `tail:{alice,bob}`.
- **Mutation-verified.** Flipping Dave's `gap:alice` count or Carol's `tail:bob` count to 1 fails `report_runner_writes_vector_fixture_reports_and_summary`, so the zero assertions demonstrably have teeth in both directions.
- **`no_pending_work` exclusions documented** for both `leaver-removal-secrecy/v1` and `incremental-growth/v1`: departed/late members intentionally retain undecryptable inputs as deferred transport work, matching the latecomer note.
- **`conversation/v1` trimmed to two rounds**; manifest coverage strings and SCENARIOS.md updated; all four fixtures reformatted to the expanded style used by every existing vector.

Local: `report_runner` (19) + `vector_artifacts` (3) pass, fmt/check clean, clippy clean on the touched crate. (`just fast-ci` currently fails on master too — Rust 1.97's new `unneeded_wildcard_pattern` lint in `convergence-campaign-runner/src/manifest.rs:469`, unrelated to this PR.)

Datawav:

No actionable findings.

_Advisory audit of head [`0cdaf0e7b41c`](https://github.com/marmot-protocol/mdk/commit/0cdaf0e7b41c52ba08534b21c87e5efe4bb7fe29); no checkout or PR-code execution was performed._

Datawav (approved):
Automated multi-agent audit of head 0cdaf0e7b41c: no actionable findings. Ready for merge.

erskingardner (approved):

erskingardner (approved):

crates/cgka-conformance-simulator/vectors/incremental-growth.v1.json
dannym-arx: Documented in 0cdaf0e7 (SCENARIOS.md, incremental-growth entry): the exclusion is intended deferred-work behavior — Carol and Dave retain their undecryptable pre-join inputs as deferred transport work, matching the latecomer vector's note, so including them in no_pending_work would fail on intended behavior.

PR #1293, opened 2026-08-07
Unify same-epoch fork resolution through distributed convergence

jgmontoya:
## Summary

Completes the story #1329 promised. #1329 made ingest keep-record-schedule instead of silently dropping a rival with no recovery anchor, and left the durable verdict to the convergence coordinator. This PR deletes the other route entirely: **distributed convergence is now the single same-epoch fork-resolution rule for every member** — live committer, observer, and restarted committer alike. The pairwise fast-path that only a live committer could take is gone.

Landing it exposed a real defect that no test tier could see, so this branch is not just a deletion: it also fixes fork resolution's visibility half and its evidence accounting, and re-authors four test oracles that were written against the old blindness.

## What is deleted, and why (`9332c9fe`, `cbaf733f`)

The two rules were proven to agree before anything was removed. `db178239` pins a committer that won the pairwise race (rival parked `ConvergenceDeferred`, never applied) being reorged onto the fleet's deeper branch once valid depth disagrees with the ordering key, plus the exact event set of a convergence apply over an own commit in both arms. `c9b861aa` pins the same fork resolved through all three routes in one portable vector. With outcomes equal, the pairwise route was a depth-blind duplicate over machinery that already handles the general case:

- the `we_committed_from` routing conjuncts and the `WrongEpoch` arm's pairwise block, its retry loop, and its direct-seam event emission;
- `ForkRecoveryManager` (`fork_recovery.rs` deleted), the ordering probe, send/apply-time `fork-` snapshots and their hydration scan, and the whole `committed_from` surface in `EpochManager`;
- `GroupEvent::ForkRecovered` end to end — traits, simulator oracle vocabulary, app routing, FFI variant — and with it the recovery-observation vocabulary (`ForkRecoveryObservation`, `ClientRecoveries`, `RecoverySummary`), which had no producible seam left and would have been a strict-oracle trap.

The deterministic ordering key survives inside branch selection, where valid branch depth outranks it. Peel fallback keeps its coverage on the durable `openmls-retained-anchor-{epoch}` family alone. Forensic READ-compat is untouched: `marmot-forensics` and `incident-replay` still parse historical `fork_resolution` rows; the engine just stops emitting them.

## The regression the unification exposed — and the fix

This is the part returning reviewers have not seen. It landed after the first review round.

**The engine tier was structurally blind.** Every Tier-2 test installed the pass-through `MockPeeler`, which hands each client every message verbatim — perfect cross-branch visibility production has never had. `bd69d68d` adds `support::epoch_sealed_peeler`, modelling the Nostr binding's actual sealing (same AEAD and framing, group key from `ctx.exporter_secret`), so a device that never entered an epoch state sees opaque bytes. Infrastructure only; it closes the blindness permanently.

**With real sealing, branch selection scored only what it could locally decrypt (`0d1388a6`).** Branch selection ranks on valid commit depth and app-witness score, both counted over *stored* inputs. A group message is sealed under the sender's current-epoch exporter secret with no epoch hint on the wire, so once two members commit from the same epoch, each branch's later traffic is opaque to devices on the other branch: retained `PeelDeferred`, contributing neither depth nor witnesses. Every committer over-scored its own branch, every device settled "stable" on its own lineage, and the split was permanent. The deleted pairwise route was immune by construction (`CommitOrderingKey` is content-only), which is why the unification is what surfaced it. RED is quoted in the commit: the engine proving test (`incumbent_adopts_a_deeper_rival_branch_whose_traffic_only_it_can_read`, left 2 / right 3) and `convergence-e2e-delivery/v1` seed 99 case 0 with `clients_not_converged` across four clients.

**The fix treats candidate branch states as part of the group's peel context, not a new mechanism.** `candidate_branch_peel` materializes each candidate branch under a `SnapshotRollbackGuard` and captures its tip's owned `GroupContextSnapshot`; the exporter secret is derived while the branch state exists, so the transient state is rolled back before any async peel runs against it. `retry_deferred_peels` offers those contexts to every retained row through the ordinary ingest seam, so dedup, classification, retirement, and the flood cap stay in one place. A message readable **only** under a candidate branch belongs to a lineage this device has not adopted, so it routes to the convergence seam and never to the direct apply — force-feeding one to canonical OpenMLS raises `UnableToDecrypt(AeadError)` and aborts the drain, and more importantly peeled bytes prove nothing on their own: the next pass's OpenMLS replay is what authenticates them, and a failed peel is silence rather than a verdict. The uncontested path pays no replay (empty-backlog early return, a cheap shared-source-epoch check, at most `MAX_CANDIDATE_BRANCH_PEEL_CONTEXTS` = 8 branches, enumeration failure yields no contexts rather than an error).

**Then the bias inverted, and `20146dda` fixes the evidence accounting.** With rival traffic readable, each row's route depended on *which context* read it: a row readable only under a candidate branch entered the convergence pass, while a row the live peel could now read took the direct apply and was never admitted. The pass's input is its frozen membership, so the device's own branch contributed no app witnesses while the branch it had not adopted contributed all of them — selection biased *off* the device's own branch, which is why `adversarial-reliability/app-witness-value/v1` case 9 still diverged (RED quoted). Now, for as long as a contested sweep holds candidate branches, an application message is evidence for the pass whichever context read it; commits need no such rule since `commit_should_enter_convergence` already decides on epoch, which is provenance-blind.

`70fd4e39` then separates two questions the sweep had conflated: whether the stored graph is contested, and whether branch enumeration managed to capture any tip context. Enumeration returns nothing on five paths and only one of them means "uncontested" — the rest are a missing snapshot, an exhausted replay budget, a missing own-commit checkpoint, or no surviving paths, all on a genuinely forked graph. Contested-ness now comes from `commits_share_a_source_epoch`, decided before any replay, and drives the drain policy alone; evidence routing keeps keying on context presence, since feeding app traffic to an adjudicator whose own enumeration just halted adds noise rather than symmetry. This is a latent-invariant fix: a halted enumeration also wedges the pass that follows it, so no reachable fixture today distinguishes the two drain policies — the value is that the code stops asserting an invariant it does not hold, and a false positive costs one deferred drain instead of N immediate ones.

The second half: **a sweep page's rows are adjudicated together**. Draining per recovered row let a verdict land — and the tip advance — partway through a page, reclassifying every row behind it, and it cost up to 64 canonicalizations where the 8-branch rationale priced one. A contested sweep now buffers the rows of one sweep page and drains once, at the end. It does **not** make a pass wait for evidence still sitting in the deferred backlog: pass membership closes at the pass cutoff during admission, not at drain time, so evidence recovered after that cutoff is adjudicated in the next pass, and an application message classified `LosingBranch` in the earlier pass stays terminal. Narrowing that window is tracked separately (see *What this does not fix*).

`6a31a553` closes the coverage gap that fix left behind. The tests shipped with `70fd4e39` hand-built the post-halt value and asserted on the consumer, so a producer that stopped stamping `contested` on a halt path would not have failed them. Both real halts are now driven end to end — a released own-commit checkpoint for `MissingOwnCommitCheckpoint`, and a ten-way fork with thirteen stacked successors for `ReplayBudgetExceeded` — each with a control run over the same graph proving the empty context set came from the halt rather than from a graph that never forked.

`b2825ac8` fixes a sharper problem found while auditing the peel cap's rationale. `MAX_CANDIDATE_BRANCH_PEEL_CONTEXTS` bounds how many candidate branches get a peel context, and the BFS completes paths in level order — so truncating its output kept the **shallowest** branches. Eight one-commit rivals from a single member, one cheap message each, therefore evicted every deeper branch from peel coverage: precisely the branches carrying the post-fork traffic a retained row needs unsealed. Candidates are now ranked by tip epoch (then `branch_id`) before truncation, joined on `branch_id` rather than a trusted index because `materialized` is allowed to come back shorter than `candidate_paths`. Both ranking keys are content-derived, so every peer with the same evidence still caps to the same set — the property that makes a bounded peel set safe in the first place. `1707921a` corrects the constant's stated justification, which claimed a graph wide enough to need more than eight branches would already exhaust the pass's `ReplayBudget`: it would not (`for_pass` is `commits × (max_rewind + 1) × 4 + 32`, so a flat nine-way fork gets 248 for about nine probes), and the repo's own budget test needs width 10, depth 13 *and* `max_rewind_commits = 0` to trip it. The bound exists because path count is attacker-inflatable and each context costs a full extra replay, not because the budget stops it.

## What the oracles now demand (`443fc309`, `4b57c366`, `259449ad`)

Four oracles were authored while cross-branch reads were cryptographically impossible and read the new, honest output as failure. Each replacement is strictly stronger (pre-change RED quoted in the commit):

- `assert_canonical_application_event`: "no `AppMessageInvalidated`, ever" → **exactly one, reason `LosingBranch`** — a payload delivered on a losing branch must now be explicitly retracted — plus a new rule that no delivered message id may appear in any retraction, so a payload shown and then abandoned fails either way.
- `assert_real_peeler_convergence_trace`: its second allowed arm accepted a client ending the schedule at epoch 2 holding only the shallower branch and no invalidation — the permanent split brain. That arm is **deleted, not generalized**; the surviving arm keeps epoch 3 / 6 members and gains a per-client retraction list.
- the `deferred-losing-transport` tail asserted the losing branch's transport object stays retained and unpeeled forever, with `NoPendingWork` failing as a steady state. It now pins the opposite: **zero deferred transport messages, zero pending inputs, zero quiescence failures.**
- `259449ad` replaces an app-runtime characterization that enumerated two observed shapes (nobody withdraws a losing row, or the losing committer alone does) with a per-participant rule. Applying a losing profile commit synthesizes a system row locally; reorging onto the selected branch withdraws it, so the tombstone is each participant's own evidence that it once held that branch. The committer now owes a withdrawal **exactly** when it ends off the branch — the old form let a converged committer silently keep a stale row and still pass — non-adopters must show none, and no participant may project a withdrawn row as delivered.

`4b57c366` closes four passing-but-empty assertions in the vectors: the settled `convergence_decision` is pinned **per client** (a client-less pin was satisfied by any one participant — exactly the regression this branch's own predecessor was), the winning branch is pinned per client through #1299's `GroupProfile` rather than `ClientsConverged` (agreement is not attribution), and `route-equivalence/v1` gains an active bidirectional-decryptability probe, since structural convergence does not prove the group can still exchange messages.

## Breaking changes

- **`cgka-traits`**: `GroupEvent::ForkRecovered` is removed, along with the FFI `GroupEventKindFfi::ForkRecovered` variant. The convergence rollback pair (`CommitRolledBack` + `GroupStateInvalidated`) is the only commit-rollback signal; the app's tombstoning already keyed exclusively on `GroupStateInvalidated`. Noted in the changelog (`2c95db57`).
- **Engine**: `EngineError::ForkedEpoch` is unreachable and its fail-closed helper is gone; `DurableGroupEvolution.recovery_snapshot` is removed (serde-JSON rows tolerate the absent field).
- **Simulator**: the recovery-observation expectations and `OracleBehavior::ForkRecovered` are removed; affected vectors are re-derived (inputs unchanged, expectations only).

## What this does not fix

- **The pass seeder still trusts wire-claimed source epochs.** As in #1329's boundary statement: the convergence pass seeder re-derives each retained commit row's source epoch from that row's own wire bytes. That is a separate, already-merged seam with every rewind and fork-recovery path in its blast radius, and it needs its own change with its own proving test.
- **Losing-branch application messages terminalize on a still-revisable decision.** Commits park `ConvergenceDeferred` and stay reconsiderable inside the rewind horizon; app messages retire `EpochInvalidated` and are never re-admitted. Two independent paths reach this: a pass that froze at its cutoff while recoverable evidence remained in the deferred backlog, and a later branch winning inside the rewind horizon. The designed equivalent of commit parking — making `LosingBranch` non-terminal while recoverable evidence remains — is a follow-up and is deliberately not smuggled in here.
- **App-message delivery order across senders is hash-derived**, not canonical. See the CI status below; a fix means either pinning payload sets or defining a canonical order, and both are decisions, not cleanups.
- **A committer's rival resolution now waits for the convergence pass** (~1.1 s quiescence measured; sends queue during it). No committer-side expedite is added. `c37c0de5`'s bounded convergence-pressure campaign measures exactly this window.
- **A queued-send scheduling gap, widened but not introduced here.** An application send accepted while a group is `Stable` with an unsettled convergence pass is queued durably by `queue_outbound_intent` **without a scheduling edge**, and no settlement seam re-arms the drain: `schedule_drain_for_retained_outbound_intents` covers only the seams that *return* a group to `Stable`, which convergence never leaves. The engine's structural progress then reports `runnable_work = 0` with no next wake while `queued_outbound_intents > 0`. This is pre-existing — `should_queue_outbound_intent`, `converge_and_drain_queued_outbound_intents`, and the `runnable_work` computation are byte-identical to `master`, and observers already took the convergence route there; what this PR changes is that live committers now spend the quiescence window in that state too, where they previously took the pairwise fast path. The app runtime masks it (`ConvergenceScheduleState::PendingOutbound` keeps a wakeup armed, and session-open hydration re-arms from the durable queue), so the payload is delayed to the pass cutoff rather than lost; the conformance driver, edge-triggered on the engine's own progress signal, is not masked. That is why `bounded_convergence_pressure_family_settles_every_seeded_permutation` is committed `#[ignore]`d with its assertions at full strength — it stays ignored until the gap is fixed in its own change (the fix is a scheduling edge in `queue_outbound_intent`, matching its sibling `do_queue_app_message`, plus the missing `runnable_work` term — a different blast radius that deserves its own proving test).

## Known CI status

One test is red-or-flaky in the full matrix: `remaining_catalog_workloads_execute_under_the_strict_oracle` (`adversarial-reliability/app-witness-value/v1`). It fails **only** on cross-sender payload ordering — convergence itself is correct, with all clients agreeing on epoch, members, and branch. MLS makes no promise about ordering across senders, so the resolution (pin payload sets, or define a canonical order) is a tracked decision and is not attempted here.

## Validation

- Each commit checked with `cargo check --all-targets`, so the branch bisects; `bd69d68d` lands its `chacha20poly1305` dev-dependency with its `Cargo.lock` entry so it builds `--locked` on its own.
- `cargo nextest run -p cgka-engine --features test-policy-overrides`: **471/471**.
- Simulator: `canonical_scenarios` **47/47**, `report_runner` **23/23**, `process_orchestrator` **11/11**, `route_assurance` **8/8**.
- `just fast-ci` green.
- The delivery family and the report spec were each run three times against the re-authored trace oracle (**3/3**) to rule out trading a deterministic red for a flaky one.
- RED evidence for every fix is quoted verbatim in the relevant commit body (`0d1388a6`, `20146dda`, `443fc309`).

## Reviewer notes

- `0e32aa6f` is a rebase casualty worth one line: master made the application-event outbox durable and at-least-once (`drain_events` empties the in-memory buffer but only the host's ack retires the row, and every session open re-seeds from it), then started writing `GroupJoined` into it. A fixture that joins by Welcome and never acks therefore hands each `reopen_legacy_client` a replayed `GroupJoined`, which two fork tests were reading as an event emitted by the ingest under test. The fix acks in the fixture; no assertion was changed, and a spurious event from the ingest still fails exactly as before.

- **The `pairwise_fork_recovery` route claim is retired, not re-pointed** (`cbaf733f`). Its two production source markers vanished with `fork_recovery.rs`; re-pointing at the unified site would give one seam two route claims and double-count its evidence, which is the inflation the source audit exists to catch. The route's two surviving properties are cited under `stored_convergence` as the renamed regressions `incumbent_committer_defers_to_deeper_convergence_branch` and `rival_win_leaves_displaced_own_commit_reconsiderable`; the retirement is recorded in prose in `CONVERGENCE_ROUTE_MATRIX.md`, and its route-lifecycle mutant survives route-agnostically as `provisional_winner_terminalization`.
- `route-equivalence/v1` was authored at `conformance_version` 0.9.10 and refreshed to the released 0.9.11 that master's release prep moved the tree to; no version bump is performed by this PR.
- `MAX_CANDIDATE_BRANCH_PEEL_CONTEXTS` is registered in the convergence constant ledger and `policy_contract::CONSTANT_DECISIONS` as an operational-resource, non-interfering constant. `MAX_DEFERRED_PEEL_ATTEMPTS = 32` is deliberately unchanged, with the reasoning recorded on the constant.
- Docs (`2c95db57`): `cgka-engine/AGENTS.md` now carries the unified route, the branch-relative peel section, and the provenance rule; `audit-logging.md` marks `fork_resolution` and `snapshot_created` as historical read-compat kinds.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic distributed convergence for competing group updates across active, restarted, and observing clients.
  * Added route-equivalence and bounded-pressure scenarios covering restarts, queued traffic, quiescence, and decryptability.
  * Added fail-closed handling when required retained history is unavailable.

* **Bug Fixes**
  * Improved convergence consistency and rollback behavior, including transactional recovery after storage failures.
  * Prevented repeatedly reprocessing messages that have already reached a terminal state.

* **Documentation**
  * Updated scenario, audit, architecture, and release documentation to describe the unified convergence behavior.


Datawav:

[P2] Version the serialized simulator schemas after removing required fields — `crates/cgka-conformance-simulator/src/app_runtime.rs:27`, `src/vector.rs:822`, `src/campaign_metrics.rs:13`

`APP_RUNTIME_OBSERVATION_SCHEMA_VERSION` remains `"1"`, while `ClientObservation.recoveries`, `ClientEventCounts.fork_recovered`, and `CampaignMeasurementsV1.recovery_observation_count` are removed. Existing V1 consumers deserializing newly generated JSON can reject it because those fields were required and have no defaults.

Need: preserve the meaning and interoperability contract of versioned simulator artifacts.

Design: either increment the affected schema versions and provide an explicit migration boundary, or retain deprecated fields—emitting empty/zero values—through the V1 lifecycle.

Alternative: if backward consumption is intentionally unsupported, document that policy and rename the `V1` structures/version identifiers so incompatible payloads cannot masquerade as the same schema.

_Advisory audit of head [`189d2ea387e8`](https://github.com/marmot-protocol/mdk/commit/189d2ea387e8904a05d61295dbe0a48d5efc8ce8); no approval, merge, push, checkout, or PR-code execution was performed._

erskingardner (dismissed):
I support the architectural direction here: deleting the pairwise fork-recovery path and making distributed convergence the single selection rule is a meaningful simplification. It reduces route-dependent behavior, aligns the implementation more closely with the formal model, and removes a substantial amount of duplicated machinery.

I do not think the current head is ready to merge, however:

1. **Blocking durability ordering issue.** In [the missing-anchor ingest path](https://github.com/marmot-protocol/mdk/blob/189d2ea387e8904a05d61295dbe0a48d5efc8ce8/crates/cgka-engine/src/message_processor/ingest.rs#L855-L872), lines 860-864 mark the raw wrapper `Processed` before `fail_closed_missing_fork_recovery_anchor()` persists the `ConvergenceDeferred` content row and durable `unrecoverable` marker. If `persist_openmls_wire_message`, `get_group`, or `put_group` fails, the only retryable source is already gone; the content may be lost or stranded without the promised halt/schedule. The existing test uses direct ingest, where the raw and content IDs coincide, so it does not exercise this path. Please retire the raw wrapper only after the durable transition succeeds (ideally atomically) and add failure-injection coverage at the write boundaries.

2. **The route-equivalence fixture does not prove every route ran distributed convergence.** It has only one client-unspecified `convergence_decision`, and the oracle accepts a matching decision from any observed client. The existing CodeRabbit thread is valid: add explicit expectations for every intended route participant. I would also add an active bidirectional-decryptability probe; exact snapshots prove state equality, but not that the resulting group can exchange application messages.

3. **The synthesized group-data incident can choose the wrong branch and still pass.** It only proves that both clients agree on some group name at epoch 2, not the expected winning profile. The existing CodeRabbit thread has the right premise, but a human-readable group name is not a branch id. After rebasing, use the `GroupProfile` oracle merged in #1299 for both clients.

4. **The quiet-window behavioral change needs one bounded pressure campaign.** Exercise finite self-updates plus profile/admin commits, application sends while convergence is pending, and a restart during convergence. Assert eventual quiescence, exact state, active decryptability, input dispositions, and bounded queue/time behavior. This does not require a progress guarantee under infinite self-updates.

5. **Rebase before relying on the green result.** This exact head is currently 11 commits behind `master`, so its checks ran against an older base.

I would also consider splitting the independent missing-anchor fail-closed behavior into a prerequisite PR. It introduces its own durable lifecycle and verified-repair semantics; isolating it would make that fix and the subsequent route deletion easier to review, revert, and bisect.

Datawav (dismissed):

No actionable findings.

_Advisory audit of head [`2292eb236d97`](https://github.com/marmot-protocol/mdk/commit/2292eb236d979fb9c5dfd8411fe1195b960c6b8d); no checkout or PR-code execution was performed. Approved as ready for merge._

jgmontoya:
> I support the architectural direction here: deleting the pairwise fork-recovery path and making distributed convergence the single selection rule is a meaningful simplification. It reduces route-dependent behavior, aligns the implementation more closely with the formal model, and removes a substantial amount of duplicated machinery.
> 
> I do not think the current head is ready to merge, however:
> 
>     1. **Blocking durability ordering issue.** In [the missing-anchor ingest path](https://github.com/marmot-protocol/mdk/blob/189d2ea387e8904a05d61295dbe0a48d5efc8ce8/crates/cgka-engine/src/message_processor/ingest.rs#L855-L872), lines 860-864 mark the raw wrapper `Processed` before `fail_closed_missing_fork_recovery_anchor()` persists the `ConvergenceDeferred` content row and durable `unrecoverable` marker. If `persist_openmls_wire_message`, `get_group`, or `put_group` fails, the only retryable source is already gone; the content may be lost or stranded without the promised halt/schedule. The existing test uses direct ingest, where the raw and content IDs coincide, so it does not exercise this path. Please retire the raw wrapper only after the durable transition succeeds (ideally atomically) and add failure-injection coverage at the write boundaries.
> 
>     2. **The route-equivalence fixture does not prove every route ran distributed convergence.** It has only one client-unspecified `convergence_decision`, and the oracle accepts a matching decision from any observed client. The existing CodeRabbit thread is valid: add explicit expectations for every intended route participant. I would also add an active bidirectional-decryptability probe; exact snapshots prove state equality, but not that the resulting group can exchange application messages.
> 
>     3. **The synthesized group-data incident can choose the wrong branch and still pass.** It only proves that both clients agree on some group name at epoch 2, not the expected winning profile. The existing CodeRabbit thread has the right premise, but a human-readable group name is not a branch id. After rebasing, use the `GroupProfile` oracle merged in #1299 for both clients.
> 
>     4. **The quiet-window behavioral change needs one bounded pressure campaign.** Exercise finite self-updates plus profile/admin commits, application sends while convergence is pending, and a restart during convergence. Assert eventual quiescence, exact state, active decryptability, input dispositions, and bounded queue/time behavior. This does not require a progress guarantee under infinite self-updates.
> 
>     5. **Rebase before relying on the green result.** This exact head is currently 11 commits behind `master`, so its checks ran against an older base.
> 
> 
> I would also consider splitting the independent missing-anchor fail-closed behavior into a prerequisite PR. It introduces its own durable lifecycle and verified-repair semantics; isolating it would make that fix and the subsequent route deletion easier to review, revert, and bisect.


Thanks — taking all six, with one mechanism correction on the blocker that makes it worse, not better.

**1 (durability ordering) — accepted, and the defect is real, but it isn't the wrapper — and while fixing it we found your instinct was righter than the mechanism: the direct ingest seam has no error compensation at all** (`do_ingest` propagates untouched), so before the fix a marker-write failure on the direct path really did leave the wrapper `Processed` with the group un-halted — pinned RED by the new fault-injection test, structural-proof now that the retirement joined the transaction. The content row is durable before the wrapper is retired (`persist_openmls_wire_message` at ingest.rs:813 precedes the retirement at :866), the retirement is a no-op except via the replay seams (raw rows only exist after an early-return ingest), and both replay seams already compensate it to `Retryable` on `Err` (mod.rs:1275-1279, :1657-1661). What IS broken: `fail_closed_missing_fork_recovery_anchor` demotes its own trigger row `Created → ConvergenceDeferred` before `put_group` writes the durable marker, and a `ConvergenceDeferred` row cannot open a convergence pass (`can_start_pass`) — so a `put_group` failure leaves an un-halted group that keeps committing on its possibly-losing branch, with the parked rival eventually `EpochInvalidated` once it falls past the anchor floor. That window exists on the direct path with no wrapper involved, and it's exactly the silent divergence the halt exists to prevent, reachable through the halt's own error path. The coordinator halt we mirrored is safe only because `missing_required_state_result` persists no disposition before its marker — our mirror missed that the emptiness was load-bearing. Fix: durable marker before the park, in one `with_transaction` that also covers the wrapper retirement (none of these writes captures snapshots, so the no-nesting constraint doesn't bite), plus one-shot fault-injection tests on the `auto_commit_atomicity.rs` `FaultStorage` pattern at each write boundary — including a replay-seam variant with a live `Retryable` raw row, which will also pin the existing compensation.

**2 — the per-client pins are already on the current head** (route-equivalence.v1.json:225-244, added for the CodeRabbit thread; the head you reviewed predated the force-push). All four decisions are real and distinct in kind — committers decide from `current_tip_epoch` 2, observers from 1. Taking the decryptability probe: it's two JSON edits (step + expectation, both shipped with #1299, capability already advertised), with one wrinkle found while validating — the probe has to precede the terminal `observe_exact`, otherwise the post-probe observation empties every per-client decision pin (the comparator resolves `client:` pins against the latest observation).

**3 — agreed, and your factoring wins over ours.** We'd closed the same escape by adding an optional `group_name` to `ClientsConverged` (pre-#1299); per-client `GroupProfile` pins are strictly stronger (name and description, per client, zero new model surface), so we're reverting our field in favor of them for both the synth fixture and route-equivalence. One nuance for the record: `selected_branch_id` stays unpinnable regardless — 19 runs produced 19 distinct values (digest over randomized envelope bytes) — so the group name is the only stable observable that names the winner; `GroupProfile` binds that same observable, just better-factored.

**4 — taking it as scoped**: finite self-updates + profile/admin commits + app sends during pending convergence + restart during convergence, asserting eventual quiescence, exact state, active decryptability, dispositions, and bounded queue/time — on the #1270/#1271 campaign scaffolding, no progress guarantee under unbounded input (consistent with PDR-2/L3).

**5 — the current head is rebased** (0 behind master as of this morning, includes #1299); the checks you saw ran pre-push.

**6 — agreed, splitting.** The fail-closed behavior plus its two pinning tests are the stack's first three commits by construction, so the prerequisite PR is a clean cut; the durability fix from (1) and its fault-injection coverage land there, and this PR rebases on top.

crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
```diff
@@ -1501,6 +1498,157 @@ async fn send_leave_family_records_seed_and_runs_generated_cases() {
...
+/// IGNORED — this currently fails on engine behavior, not on the campaign. An
+/// application message accepted while the group is resolving a same-epoch fork
+/// is queued durably and then stranded: once the pass completes, nothing
+/// re-arms the retained-intent drain, and the engine's own conformance
+/// structural progress reports `runnable_work = 0` with no next wake while
+/// `pending_work.queued_outbound_intents > 0`. The quiescence driver therefore
+/// reports `Blocked { subsystems: ["scenario_inputs"] }` and the payload is
+/// never published. Un-ignore once the retained-intent drain is re-armed after
+/// a completed pass; do not weaken the assertions below to make it pass.
+#[ignore = "engine strands application sends accepted inside the convergence quiescence window"]
```
Datawav, line 1633: The ignored test documents a live engine regression: application sends accepted during convergence remain in `queued_outbound_intents`, but completing the pass leaves `runnable_work == 0` and schedules no wake, so the payload may never publish without unrelated app-level activity. Re-arm the retained-intent drain when convergence settles (and report it as runnable structural work), or reject such sends with an explicit retryable result; then enable this acceptance test.

Datawav (commented):

[P1] Re-arm queued outbound intents after convergence — `crates/cgka-conformance-simulator/tests/canonical_scenarios.rs:166`

The unified route makes committers queue sends during the convergence window, but completing convergence does not schedule the retained-intent drain. The new acceptance test documents the resulting state: `runnable_work == 0` while `queued_outbound_intents > 0`, so an accepted message remains indefinitely unpublished until unrelated work wakes the group. The test is disabled with `#[ignore]`, leaving the regression uncovered.

Need: engine-level callers must make progress without relying on marmot-app’s supplemental scheduling.

Design: when a convergence pass settles, schedule any group with retained outbound intents and ensure structural progress reports them as runnable.

Alternative: reject sends while convergence is active with an explicit retryable result, rather than accepting work the engine cannot independently drain. Re-enable the acceptance test once either contract is implemented.

_Advisory audit of head [`0f7959994053`](https://github.com/marmot-protocol/mdk/commit/0f7959994053e7f889b10a7150954d705168f987); no approval, merge, push, checkout, or PR-code execution was performed._

Datawav (commented):
## Automated review

No newly introduced actionable findings.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -244,870 +365,769 @@ impl<S: StorageProvider> Engine<S> {
...
+                let active_pass = self
+                    .storage
+                    .convergence_pass(&group_id)?
+                    .is_some_and(|pass| pass.is_active());
+                let has_source_anchor = self.has_retained_anchor_snapshot(&group_id, msg_epoch)?;
+                convergence_refused_for_missing_anchor = within_rewind_horizon
+                    && !active_pass
+                    && !has_source_anchor
+                    && !self.msg_is_pre_membership(&group_id, msg_epoch);
+                within_rewind_horizon && (active_pass || has_source_anchor)
```
Datawav, line 944: This newly routes live committers’ in-horizon same-epoch rivals through convergence, but convergence settlement does not re-arm the durable outbound-intent drain. An application payload accepted during that window can therefore remain queued after the pass completes with `runnable_work = 0` and no next wake, so it is never published without an unrelated event. Please atomically schedule the retained-intent drain when convergence settles and include persisted queued intents in restart runnable-work recovery; then enable `bounded_convergence_pressure_family_settles_every_seeded_permutation` as the regression gate.

Datawav (commented):
## Automated review

Changes requested: one actionable liveness defect remains.

- **High — accepted outbound messages can remain indefinitely stranded after fork convergence.** In `crates/cgka-engine/src/message_processor/ingest.rs:940–950`, the new `within_rewind_horizon && (active_pass || has_source_anchor)` routing sends live committers’ same-epoch rivals through convergence, widening the affected window. The added but ignored test at `crates/cgka-conformance-simulator/tests/canonical_scenarios.rs:1624–1635` demonstrates the consequence: convergence finishes with `queued_outbound_intents > 0`, `runnable_work = 0`, and no next wake, so a durably accepted application payload is never published absent an unrelated event. Smallest sound correction: atomically schedule the retained-intent drain when convergence settles, include persisted queued intents in runnable-work recovery after restart, and enable `bounded_convergence_pressure_family_settles_every_seeded_permutation` as the regression gate.

crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
```diff
@@ -1503,6 +1501,157 @@ async fn send_leave_family_records_seed_and_runs_generated_cases() {
...
```
Datawav, line 1633: This ignored regression documents a liveness failure: an application send accepted during fork resolution remains in `queued_outbound_intents` after convergence, while scheduling reports no runnable work or next wake. Re-arm the retained-outbound-intent drain when the convergence pass settles, then enable this test; otherwise accepted messages can remain permanently unpublished without unrelated activity.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1525,6 +1583,286 @@ fn missing_required_state_result(
...
+            err @ (OpenMlsProjectionError::MissingOwnCommitCheckpoint
+            | OpenMlsProjectionError::ReplayBudgetExceeded),
+        ) => {
+            tracing::debug!(
+                target: "cgka_engine::openmls_projection",
+                method = "candidate_branch_peel_contexts",
+                reason = candidate_branch_enumeration_halt_reason(&err),
+                "no candidate branch peel contexts: branch enumeration halted"
+            );
+            return Ok(Vec::new());
```
Datawav, line 1769: Returning an empty context list on `MissingOwnCommitCheckpoint` or `ReplayBudgetExceeded` loses the fact that the stored graph is already known to be contested. `DeferredPeelSweep::is_contested()` then returns false, so recovered rows drain individually and may be terminalized from partial evidence. Preserve contested state independently of the captured contexts and defer the complete evidence batch even when branch enumeration exhausts.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -1192,94 +1236,26 @@ impl<S: StorageProvider> Engine<S> {
...
         }
 
+        // One drain over the whole batch. A contested sweep buffered every row
+        // it recovered without draining, because those rows are one evidence
+        // set and a pass must not freeze on a prefix of it: adjudicating half a
+        // fork settles a verdict the other half would have changed, and the
+        // losers of that verdict are terminalized before the evidence that
+        // would have saved them is even admitted. Uncontested sweeps drained
+        // per row on the way in and have nothing left to do here.
+        if sweep.is_contested() {
```
Datawav, line 1266: This “whole batch” drain covers only the current page: `end` is capped at `start + MAX_DEFERRED_ROWS_PER_SWEEP` (64), but convergence runs immediately afterward. With more than 64 deferred rows, the first page can settle and terminalize losing application evidence before later pages are admitted. Buffer every page in the contested sweep before draining, or otherwise freeze and adjudicate one complete bounded evidence set.

Datawav (commented):
## Automated review

Outcome: three actionable correctness/liveness defects remain.

- **High — accepted sends can remain permanently stranded after convergence.** Changed-line evidence: `canonical_scenarios.rs:1621–1635` adds an ignored regression documenting `queued_outbound_intents > 0` while `runnable_work = 0` and no next wake. Consequently, an accepted application message is never published unless unrelated activity wakes the group. Re-arm the retained-outbound-intent drain whenever a convergence pass settles, expose that work to scheduling, and enable the regression.

- **High — replay exhaustion misclassifies a known fork as uncontested.** Changed-line evidence: `openmls_projection.rs:1717–1727` converts `MissingOwnCommitCheckpoint` or `ReplayBudgetExceeded` into an empty context list; `ingest.rs:68–69` defines contested state solely as a nonempty list; `message_processor/mod.rs:1254–1257` performs the whole-batch drain only when that predicate is true. Deferred rows can therefore be drained individually, settling on partial evidence and permanently invalidating application messages that later evidence would have preserved. Return contested state independently from captured contexts and defer the complete evidence batch whenever the stored graph is known to fork; add a forced-exhaustion regression.

- **High — “whole-batch” adjudication covers only the current 64-row page.** Changed-line evidence: `message_processor/mod.rs:112` retains `MAX_DEFERRED_ROWS_PER_SWEEP = 64`; around `:1211` the sweep ends at `min(start + 64, total)`, yet `:1248–1257` immediately runs convergence after processing that slice. With a larger contested backlog, the first page can terminalize losing application evidence before later pages are admitted, recreating the prefix-verdict failure the new batching is intended to prevent. Do not drain until every page in the contested sweep has been buffered, or freeze and adjudicate one complete bounded evidence set; add a regression exceeding 64 deferred rows.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -2909,3 +2813,53 @@ fn convergence_ingest_outcome(
...
+
+#[cfg(test)]
+mod tests {
+    use super::{ConvergenceDrain, DeferredPeelSweep};
+    use crate::openmls_projection::CandidateBranchPeel;
+
+    /// The whole point of carrying `contested` separately: branch enumeration
+    /// halts for reasons that say nothing about whether the graph is split, so
+    /// an empty context set must not read as "no fork". The two facts drive
+    /// different decisions and must not be collapsed back into one.
```
Datawav, line 2853: This manually constructs the desired post-halt state, so it cannot catch regressions in either real halt path. Please add focused coverage that drives `candidate_branch_peel` into `MissingOwnCommitCheckpoint` and `ReplayBudgetExceeded`, then verifies each result remains contested and uses deferred batch draining.

Datawav (commented):
## Automated review

Outcome: the correctness defect is fixed, but its requested regression coverage remains incomplete; no new defects were introduced.

- **Low — forced-exhaustion regression remains unresolved.** The changed tests at `ingest.rs:2813–2865` manually construct `CandidateBranchPeel { contested: true, contexts: [] }`; they never exercise `MissingOwnCommitCheckpoint` or `ReplayBudgetExceeded` through `candidate_branch_peel` (`openmls_projection.rs:1752–1800`). The production fix correctly separates `contested` from captured contexts and consistently defers/drains contested sweeps, but a future regression in either actual halt path could restore the original data-loss behavior without failing these tests. Add one focused test per halt condition that drives real branch enumeration and asserts the resulting sweep remains contested with deferred batch draining.

Datawav (dismissed):
Two blocking correctness/liveness findings remain:

- **P1 — queued sends can remain stranded after convergence.** The deterministic scenario in `crates/cgka-conformance-simulator/tests/canonical_scenarios.rs:1625-1636` ends with queued outbound intents but no runnable work or next wake, and the test remains disabled with `#[ignore]`. Because this PR routes live committers through the convergence window, it expands exposure to the defect. Re-arm the retained-intent drain when convergence settles, include persisted queued intents in restart scheduling, and enable the regression test.

- **P1 — application messages are permanently discarded based on a revisable branch decision.** In `crates/cgka-engine/src/openmls_projection.rs:2919-2937`, `LosingBranch` messages become terminal `EpochInvalidated` even though later evidence within the rewind horizon can change the selected branch. If the former losing branch later wins, its application messages cannot be reconsidered. Keep those messages deferred while the branch decision remains revisable, and add a regression that changes the winner within the retained horizon.

Datawav (commented):
## Automated review

Outcome: no previously requested change was touched by this delta, and no new actionable findings were introduced.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1525,6 +1583,328 @@ fn missing_required_state_result(
...
+    }
+
+    let pending_proposals = pending_proposal_messages(&inputs.pending_messages)?;
+    let mut contexts = Vec::with_capacity(max_contexts.min(path_result.candidate_paths.len()));
+    // Each tip context costs one replay of its path. The BFS already replayed
+    // every node to enumerate the paths; capturing the contexts inside that
+    // walk would have to carry exporter material through
+    // `OpenMlsMaterializedCandidate` and into every canonicalization result, so
+    // this pays a bounded second replay to keep secrets out of those types.
+    for path in path_result.candidate_paths.iter().take(max_contexts) {
```
Datawav: This always retries the same first eight candidate paths. With nine or more materializable branches, deferred peel can permanently exclude a later branch, so decisive branch-sealed commits or application witnesses may remain unreadable and peers can score incomplete evidence differently. Please rotate/persist the starting position between sweeps, or process bounded batches until every candidate path has been attempted; add a nine-branch regression where the decisive evidence is only on the initially omitted branch.

Datawav (commented):
## Automated review

Outcome: one high-severity convergence/liveness defect.

- High — Valid fork evidence can be permanently excluded. `message_processor/mod.rs:126` caps branch peel contexts at eight, while `openmls_projection.rs:1784` always selects `candidate_paths.iter().take(max_contexts)`. Every deferred-peel sweep repeats that same prefix without a cursor or rotation. A fork with nine or more materializable branches can therefore leave branch-sealed commits or application witnesses on an omitted branch permanently unreadable; members with differing visible candidate sets may score incomplete evidence and fail to converge on one winner. The cap’s justification is unsound because `ReplayBudget::for_pass` scales with commit count, so more than eight branches do not inherently exhaust it. Smallest sound correction: durably rotate the starting candidate across sweeps—or process all candidates in bounded batches—until every branch has been attempted; add a nine-branch regression case where decisive evidence exists only on the initially omitted branch.

jgmontoya:
@Datawav 

Thanks — the mechanism you spotted is real and it led to two fixes, but the failure mode is not divergence, and I'm not taking the rotating cursor. Details, since the reasoning matters more than the verdict.

**The divergence claim does not hold, because the prefix is content-derived.** `build_stored_openmls_candidate_paths` sorts commits by `(source_epoch, digest, payload)` before the walk (`openmls_projection.rs:1921`), the BFS iterates in exactly that order, every intermediate map is a `BTreeSet`/`BTreeMap`, and `branch_id` is SHA-256 over the path's commit digests. No `HashMap` iteration, no arrival timestamp, and no storage row order survives into the output. So two members holding the same evidence enumerate the same paths in the same order and truncate to the *same* eight. The blindness is uniform, not per-device — suboptimal, but convergent. Members can of course hold different evidence, but that is true with or without the cap and is what redelivery and the retry lifecycle exist to settle.

**And a truncated branch can still win.** `take(` appears exactly once in that file, and it is not on the selection path: selection builds its own uncapped candidate set (`:1450`) and hands the full vector to canonicalization (`:1485`). A branch past position 8 still appears as a candidate, still contributes its stored depth, and can still be selected — after which its traffic decrypts natively against live state and the deferred rows recover. The cap governs what a *non-adopter* can read, not who wins.

**Why the rotating cursor is the one change I'd refuse.** Uniformity is precisely what makes a bounded peel set safe. A durable per-device cursor makes the peeled subset a function of each device's own sweep history — device-local state feeding branch selection. Two peers at different offsets would read genuinely different branch subsets *at the same time*, converting a convergent blind spot into a divergent one. Worse under the adversarial case you're implicitly worried about: an attacker adding branches keeps every device's cursor at a different offset, maximizing the asymmetry and handing them phase control over who sees what. "Bounded batches until all attempted" has the same problem from the other end — candidates are attacker-inflatable at one message each against a backlog of up to 256 rows, and the "all attempted" target grows every time a branch is added, so it need not terminate.

**What you were right about, and what we changed.** The comment justifying the number 8 was wrong, exactly as you said. It claimed a graph wide enough to need more than eight branches would already be failing the pass's `ReplayBudget` — but `for_pass` is `commits × (max_rewind + 1) × 4 + 32`, so a flat 9-way fork (9 commits) gets a budget of 248 and needs about 9 probes. Our own budget test has to stack width 10, depth 13 *and* `max_rewind_commits = 0` to trip it. Width alone never exhausts the budget, and the rationale has been rewritten to say what actually bounds this: path count is attacker-inflatable, and each context costs a full extra path replay plus up to `MAX_DEFERRED_ROWS_PER_SWEEP` AEAD attempts.

Chasing that also turned up something sharper than the nine-branch case. `completed` is filled in BFS level order, so `take(8)` was keeping the *shallowest* eight — meaning eight junk depth-1 commits from a single member evicted every deeper branch from peel coverage, for eight messages. Candidates are now ranked by tip epoch (then `branch_id`) before truncation. Both keys are content-derived, so every peer still cuts identically; the deepest branches are kept instead of the accidental ones.

**On the regression test:** added, but pinning convergence rather than rotation — a >8-branch graph truncates to exactly the cap, two independently-seeded stores with identical evidence produce an identical branch-id sequence, and a deep branch is retained over shallow ones. That last one is the eviction case; it had no coverage at all before.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -1192,94 +1243,31 @@ impl<S: StorageProvider> Engine<S> {
...
+        // losers of that verdict are terminalized before the evidence that
+        // would have saved them is even admitted. Uncontested sweeps drained
+        // per row on the way in and have nothing left to do here.
+        //
+        // Same predicate as `drain_policy`, so no row is ever deferred to a
+        // drain that does not run. A sweep whose enumeration halted recovers
+        // nothing and still drains: the fork is real, and the pass — not this
+        // sweep — owns what to do about a fork it cannot enumerate.
+        if sweep.is_contested() {
+            self.converge_stored_openmls_messages_with_time(group_id, now)
```
Datawav, line 1267: **P1 — Do not converge after processing only one deferred-evidence page.** `end` caps this sweep at 64 rows, but this call immediately adjudicates a contested graph. With more than 64 retained rows, decisive evidence in a later page is excluded from the verdict, and terminalizing the apparent losers can prevent recovery on subsequent sweeps. Defer convergence until the complete deferred cycle has been admitted, or prohibit terminalization while unprocessed rows remain. Please add a ≥65-row regression test with the decisive evidence after index 63.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1525,6 +1583,362 @@ fn missing_required_state_result(
...
+    let pending_proposals = pending_proposal_messages(&inputs.pending_messages)?;
+    let ranked =
+        candidate_paths_ranked_for_peel(&path_result.candidate_paths, &path_result.materialized);
+    let mut contexts = Vec::with_capacity(max_contexts.min(ranked.len()));
+    // Each tip context costs one replay of its path. The BFS already replayed
+    // every node to enumerate the paths; capturing the contexts inside that
+    // walk would have to carry exporter material through
+    // `OpenMlsMaterializedCandidate` and into every canonicalization result, so
+    // this pays a bounded second replay to keep secrets out of those types.
+    for path in ranked.into_iter().take(max_contexts) {
```
Datawav, line 1786: **P1 — This cap can permanently exclude the actual winning branch.** Ranking considers only already-materialized paths, but an excluded shallow root may have opaque child commits that become readable only under that root’s context and would make it uniquely deepest. Because the root cannot improve its rank until selected, convergence may choose another branch and terminalize the real winner. Rotate or exhaust all bounded context windows before irrevocable convergence, or block terminalization while potentially decisive deferred traffic remains unreadable. Please add a >8-root regression test where an excluded root’s opaque descendants make it the unique winner.

Datawav (commented):
## Automated review

Outcome: two P1 protocol-correctness defects can prematurely resolve a fork before all decisive evidence is visible.

- P1 — Deferred evidence is adjudicated after only the first 64 rows. In `crates/cgka-engine/src/message_processor/mod.rs::retry_deferred_peel_for_group`, the changed code sets `end = (start + MAX_DEFERRED_ROWS_PER_SWEEP).min(total)`, processes only `start..end`, then immediately calls `converge_stored_openmls_messages_with_time()` whenever the graph is contested. This contradicts the adjacent invariant that convergence “must not freeze on a prefix” because losers can be terminalized before saving evidence is admitted. With more than 64 retained rows, decisive evidence in a later page cannot affect the first verdict, and invalidation may prevent recovery on subsequent sweeps. Smallest sound correction: defer convergence until the complete deferred cycle has been admitted, or prohibit terminalization while unprocessed rows remain; add a ≥65-row regression test with the decisive evidence after index 63.

- P1 — The eight-context cap can permanently hide a winning branch’s depth. `message_processor/mod.rs` passes `MAX_CANDIDATE_BRANCH_PEEL_CONTEXTS = 8` into `candidate_branch_peel()`. In `openmls_projection.rs`, contexts are selected with `ranked.into_iter().take(max_contexts)`, while `candidate_paths_ranked_for_peel()` ranks only already-materialized paths by `tip_epoch` and `branch_id`. Because the changed code also establishes that post-fork traffic is unreadable until processed under its sender branch’s context, an excluded shallow root may have opaque child commits that would make it uniquely deepest—but those commits cannot raise its rank without the context already being selected. Convergence can therefore choose another root and terminalize the actual winner. Smallest sound correction: rotate or exhaust all bounded context windows before irrevocable convergence, or block terminalization while potentially decisive deferred traffic remains unreadable; add a >8-root test where the excluded root’s opaque descendants make it the unique winner.

PR #1294, opened 2026-08-07
Add isolation and catch-up scenario vectors from the multi-VM catalog

dannym-arx:
## Summary

Slice two of the multi-VM catalog port. The marmot colony now proves its burrows do not leak into each other.

- `multigroup-isolation/v1`: the first `in_group` multi-group vector. Alice overlaps two live groups; Bob and Carol prove cross-group plaintext isolation with `payload_count` zero assertions.
- `pairwise-threads/v1`: the full pairwise thread mesh over three clients, pinning same-account thread isolation.
- `admin-handover/v1`: multi-hop admin succession. Alice promotes Bob, Bob renames and promotes Carol, Carol renames, Alice demotes Bob, and Bob's next admin edit is denied. Ends with `clients_exactly_equivalent` at epoch 6.
- `reconnect-catchup/v1`: Dave's daemon goes down while traffic and a rename commit land; one tick on reconnect catches him up on all of it.
- registers all four in `vectors/manifest.v1.json` and `SCENARIOS.md`

## Stack

Fourth slice in the #1250 series lineage: #1270, #1271 (merged), #1272, then #1292 (base of this PR). Targets `scenario-catalog/transcript-secrecy` and re-targets as parents merge. The remaining planned slice is the seeded churn family.

## Scope boundary

JSON fixtures and docs only; no production code. True durable-relay offline recovery stays in the `adversarial-reliability` generated family; `reconnect-catchup/v1` models mailbox catch-up, which is what a vector fixture can express through the engine subject.

## Verification

- `cargo test -p cgka-conformance-simulator --locked --test canonical_scenarios canonical_vector_fixtures_match_generated_traces`
- `cargo test -p cgka-conformance-simulator --locked --test vector_artifacts`
- `cargo test -p cgka-conformance-simulator --locked --test scenario_ir`
- `scripts/check_legacy_naming.sh`

## Summary by CodeRabbit

* **New Features**
  * Added conformance scenarios for multi-group message isolation, administrator handover, and deferred synchronization after missed activity.
  * Added coverage for group creation, renaming, membership changes, administrator permissions, and final client state convergence.
  * Added verification that messages remain restricted to their intended groups.
* **Documentation**
  * Documented the new portable scenarios, expected outcomes, and coverage guidance.


crates/cgka-conformance-simulator/vectors/multigroup-isolation.v1.json
```diff
@@ -0,0 +1,149 @@
...
+        "action": { "type": "observe_exact", "clients": ["bob"] }
+      },
+      {
+        "type": "in_group",
+        "group": "b",
+        "action": { "type": "observe_exact", "clients": ["carol"] }
+      }
+    ]
+  },
+  "expected_outcomes": [
```
erskingardner, line 304: Alice is the only overlapping member and the whole point of the vector, but she is never observed and has no `client_state` outcome. As written the vector proves the two single-group members are excluded; it does not prove the overlap member actually receives both groups' traffic, which is the other half of "cross-group encryption boundaries under overlapping membership."

This is cheap to close: add `alice` to one of the `observe_exact` steps and a `client_state` for her with `received_payloads` of `multigroup:a:bob` and `multigroup:b:carol`. Without it, a subject that silently dropped all of Alice's group-b traffic would still pass this vector.

crates/cgka-conformance-simulator/vectors/pairwise-threads.v1.json
```diff
@@ -0,0 +1,351 @@
...
+      {
+        "type": "tick",
+        "clients": [
+          "alice",
+          "bob",
+          "carol"
+        ]
+      },
+      {
+        "type": "assert",
```
erskingardner: Each client has two foreign payloads, but only one is pinned per client: Alice's `pairs:bc:carol`, Bob's `pairs:ac:carol`, and Carol's `pairs:ab:bob` are all unasserted. Since SCENARIOS.md concedes the exact transcripts do not carry the isolation claim here, these three assertions *are* the vector, and half the isolation surface is uncovered.

Three more `payload_count: 0` blocks cost ~36 lines in a file already at 351 and make the "same-account thread isolation" claim complete.

crates/cgka-conformance-simulator/vectors/multigroup-isolation.v1.json
```diff
@@ -0,0 +1,149 @@
...
+        "type": "in_group",
+        "group": "b",
+        "action": { "type": "clear_events", "clients": ["alice", "carol"] }
+      },
+      {
+        "type": "in_group",
+        "group": "a",
+        "action": { "type": "send_app_message", "sender": "alice", "payload": "multigroup:a:alice" }
+      },
+      { "type": "acknowledge_outbound", "client": "alice", "outcome": "accepted" },
```
erskingardner: These bare `acknowledge_outbound` steps after app-message sends (4 here, 6 in `pairwise-threads.v1.json`) do not appear to be needed. `three-client-message-exchange/v1` sends app messages with no acknowledgement, and `reconnect-catchup.v1.json` in this same PR sends nine of them unacknowledged and still asserts `no_pending_work`.

Worth noting they also assert nothing: `scenario.rs` only errors on an empty match when `publication` is `Some`, so a bare acknowledgement that matches zero artifacts passes silently. Either drop them, or if they are load-bearing here, give them a `publication` so a future regression fails loudly.

crates/cgka-conformance-simulator/vectors/deferred-tick-catchup.v1.json
```diff
@@ -0,0 +1,110 @@
...
+    "name": "reconnect-catchup/v1",
+    "spec_version": "2",
+    "clients": ["alice", "bob", "carol", "dave"],
+    "steps": [
+      {
+        "type": "create_group",
+        "creator": "alice",
+        "name": "reconnect",
+        "invitees": ["bob", "carol", "dave"],
+        "required_features": [],
```
erskingardner, line 25: `initial_admins` is omitted here but present in the other three vectors added by this PR. Alice needs admin rights for the `update_group_data` rename at step 16 and gets them implicitly as creator. Adding `"initial_admins": ["alice"]` makes the precondition for the rename explicit and matches the sibling fixtures.

crates/cgka-conformance-simulator/vectors/deferred-tick-catchup.v1.json
```diff
@@ -0,0 +1,110 @@
+{
```
erskingardner, line 1: Two nits on this file.

**Formatting:** this file and `multigroup-isolation.v1.json` use hand-compacted single-line objects, while every existing vector in `vectors/` and the other two files in this PR are fully expanded `serde_json` pretty output. Worth normalizing so diffs on these fixtures stay reviewable line-by-line.

**Name:** nothing in the vector disconnects or reconnects — the engine subject does not declare `ParticipantConnectivity`, so "offline" is modelled purely by not ticking Dave. SCENARIOS.md is honest about this, but the scenario id is the portable contract another implementation loads, and `reconnect-catchup/v1` promises a connectivity transition the steps never perform. Something like `deferred-tick-catchup/v1` would describe what the steps actually do; the real coverage here (one tick spanning missed app traffic on both sides of a missed epoch-advancing commit) survives the rename intact.

erskingardner (commented):
Fixtures-only, additive, CI green, and the premise holds up: the bus broadcasts every group message to every attached client (`src/bus.rs`), so the cross-group `payload_count: 0` assertions are real decryption-boundary claims rather than vacuous routing claims. `multigroup-isolation/v1` and `admin-handover/v1` both cover something the catalog genuinely lacked.

What I'd like addressed before merge is that the deliverable here *is* the assertions, and two of the four vectors assert less than their own documentation claims:

1. **`multigroup-isolation/v1` never observes Alice** — the only client with overlapping membership, and the entire stated pressure of the vector. Nothing pins that she received both groups' traffic or that her two group states stayed distinct.
2. **`pairwise-threads/v1` pins one of two foreign payloads per client.** Three of six isolation edges are unasserted in a 351-line file.

The rest are consistency items. Inline.

One scoping question rather than a blocker: `pairwise-threads/v1` costs 351 lines to assert the same property as the 149-line `multigroup-isolation/v1`, and per SCENARIOS.md its own exact transcripts carry no isolation claim (observation buffers are per-client), so the whole value sits in three zero-count assertions. Its one genuine addition over `multigroup-isolation/v1` is a second group whose creator is not Alice. Could that be folded into `multigroup-isolation/v1` as a third group instead of a separate vector? That's roughly what the manifest's own `next` field for that entry proposes.

Datawav:

- **Medium — `vectors/reconnect-catchup.v1.json`: renamed value is never pinned.** The scenario’s purpose is to prove that Dave catches up across the rename to `reconnect-renamed`. `clients_converged` does compare group names, so it detects Dave diverging from the others, but it only requires equality—not the intended value. An implementation that leaves every client at the original `reconnect` name could still satisfy the current expected outcomes.

  **Need:** Verify both convergence and successful application of the rename.

  **Design:** Add `"group_name": "reconnect-renamed"` to Dave’s `client_state` expectation, preferably to all four clients’ state expectations.

  **Alternatives:** Introduce an explicit expected-group-name outcome, or use an exact-state assertion that pins group metadata. `clients_exactly_equivalent` alone would still require a separate assertion of the intended name.

_Advisory audit of head [`39dcd6f794f2`](https://github.com/marmot-protocol/mdk/commit/39dcd6f794f2464d7f2b22d74934fc842da85574); no approval, merge, push, checkout, or PR-code execution was performed._

Datawav (commented):

- **Medium — `reconnect-catchup/v1` does not verify the renamed group value.** The vector applies `name: "reconnect-renamed"` at `vectors/reconnect-catchup.v1.json:113`, but its expected outcomes only check epoch, membership, payloads, convergence, and pending work. Because convergence requires only that clients agree, all clients could retain the original `"reconnect"` name and still satisfy the fixture, contradicting the documented catch-up guarantee.

  **Need/design:** Pin `"reconnect-renamed"` in an exact expected outcome for Dave—and preferably the converged cohort—so the scenario proves that offline catch-up applied the missed group-data commit, not merely its epoch transition.

  **Alternatives:** Extend `client_state` or `clients_converged` with the expected name; if those schemas intentionally exclude names, add a dedicated group-name observation/assertion.

_Advisory audit of head [`e5f024a5d380`](https://github.com/marmot-protocol/mdk/commit/e5f024a5d3806f2e1c75afc1224e752c36fdd724); no approval, merge, push, checkout, or PR-code execution was performed._

crates/cgka-conformance-simulator/vectors/deferred-tick-catchup.v1.json
```diff
@@ -0,0 +1,245 @@
...
+    {
+      "type": "pending_resolution",
+      "step_index": 17,
+      "client": "alice",
+      "pending": "rename",
+      "resolution": "confirmed"
+    },
+    {
+      "type": "client_state",
+      "client": "dave",
```
Datawav, line 197: The expected state for Dave never verifies the renamed group value (`reconnect-renamed`). Epoch and convergence checks can still pass if every client retains the original name, so this vector does not prove that Dave applied the missed group-data commit during catch-up. Add an exact group-name assertion for Dave, and preferably the converged cohort.

crates/cgka-conformance-simulator/vectors/multigroup-isolation.v1.json
dannym-arx: Done in 19809ac3: Alice is now observed in group a and carries a client_state pinning exactly [`multigroup:a:bob`, `multigroup:b:carol`] — a subject dropping her group-b traffic now fails the vector.

crates/cgka-conformance-simulator/vectors/pairwise-threads.v1.json
dannym-arx: Done in 19809ac3: all six foreign payloads are now pinned — added `alice/pairs:bc:carol`, `bob/pairs:ac:carol`, `carol/pairs:ab:bob`. Mutation-checked: expecting 1 for a foreign payload fails the runner.

crates/cgka-conformance-simulator/vectors/multigroup-isolation.v1.json
dannym-arx: Dropped all 10 in 19809ac3 (4 here, 6 in pairwise-threads). You were right that they were not load-bearing: the full vector suite passes without them, including the `no_pending_work` outcomes.

crates/cgka-conformance-simulator/vectors/deferred-tick-catchup.v1.json
dannym-arx: Added `"initial_admins": ["alice"]` in 19809ac3.

crates/cgka-conformance-simulator/vectors/deferred-tick-catchup.v1.json
dannym-arx: Both taken. Formatting was normalized to expanded pretty output on the rebase yesterday, and 19809ac3 renames the scenario to `deferred-tick-catchup/v1` (file, ids, manifest, SCENARIOS.md) — SCENARIOS.md now also states explicitly that no ParticipantConnectivity transition is involved and "offline" is purely deferred ticking.

crates/cgka-conformance-simulator/vectors/deferred-tick-catchup.v1.json
dannym-arx: Done in 19809ac3: `group_profile` expectations pin `reconnect-renamed` for the whole cohort (all four clients), so the missed group-data commit must actually be applied during catch-up. Mutation-checked: expecting the stale name for Dave fails the runner. 2a635ac5 applies the same treatment to admin-handover.

crates/cgka-conformance-simulator/vectors/admin-handover.v1.json
```diff
@@ -0,0 +1,340 @@
...
+    {
+      "type": "client_state",
+      "client": "carol",
+      "epoch": 6,
+      "member_count": 3,
+      "received_payloads": [
+        "handover:final:alice",
+        "handover:final:bob"
+      ]
+    },
```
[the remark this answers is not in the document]
dannym-arx: Done in 2a635ac5: `group_profile` expectations now pin `handover-carol` for all three clients alongside the exact-equivalence check.

Datawav (approved):

No actionable findings.

_Advisory audit of head [`310628766037`](https://github.com/marmot-protocol/mdk/commit/310628766037e6249078cf4f42d248cf5c752fce); no checkout or PR-code execution was performed. Approved as ready for merge._

PR #1299, opened 2026-08-07
Add group-profile scenario oracle coverage

erskingardner:
## What changed

- add Scenario IR v3 with a full `update_group_profile` action for independent name and description changes
- keep Scenario IR v2 accepted and semantically unchanged
- make Scenario Authoring v1 select v2 or v3 from its expanded actions and allow v3 actions in its schema
- carry group descriptions through engine, app-runtime, process, reference, and retained-relay simulator subjects
- add exact `GroupProfile` expectations so equal-but-stale projections fail the oracle
- strengthen `ClientsConverged` to compare both profile fields
- add a canonical profile-update vector, cross-adapter coverage, offline recovery coverage, and a real projection mutation sentinel
- count successful active decryptability probes as the application-delivery evidence they exercise

## Why

The simulator could express only name changes and primarily prove that clients agreed with one another. Two clients could therefore expose the same stale or incorrect group profile and still satisfy convergence. This left both an input-language gap and an oracle gap for an important application-visible projection.

Scenario IR v3 makes the complete group profile expressible, while the expected-value oracle verifies the actual requested name and description. V2 remains a separate compatibility contract because canonical Scenario IR is persisted in vectors and failure capsules and exchanged with external campaign runners; changing an existing v2 action would silently change what older consumers accept. The authoring compiler now chooses v3 automatically when the expanded document contains a v3-only action.

The mutation-adequacy witness now executes `group-profile-update/v1` through the real engine subject, verifies the baseline with the portable oracle, then mutates only the app-facing profile projection and proves that the expected-value oracle rejects it.

## Impact

This is simulator and testing infrastructure only. It does not change the production convergence engine or application behavior. Existing Scenario IR v2 scenarios remain valid.

## Validation

- `cargo test -p cgka-conformance-simulator --test scenario_ir --test mutation_adequacy`
- `cargo test -p cgka-conformance-simulator --lib` — 129 passed
- file-backed strict-oracle report run — 20 vectors passed, 0 failed
- `just fast-ci`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added support for updating group names and descriptions independently or together.
  * Added Scenario IR v3 with validation, profile assertions, and backward compatibility with v2.
  * Added conformance coverage for profile propagation, convergence, encryption, and pending-work completion.

* **Bug Fixes**
  * Convergence checks now detect mismatched group descriptions.
  * Preserved omitted profile fields during partial updates while allowing explicit clearing.

* **Documentation**
  * Updated guides, schemas, scenarios, and coverage documentation for group-profile updates and Scenario IR v3.


crates/cgka-conformance-simulator/src/mutation_adequacy.rs
```diff
@@ -92,6 +94,24 @@ pub async fn run_mutation_sentinel(mutation: SemanticMutation) -> MutationSentin
     }
 }
 
+fn group_profile_projection_sentinel() -> (String, String) {
```
erskingardner: Blocking: this sentinel executes nothing. `baseline_projection` is an alias for the literal `requested`, `mutant_projection` is a second literal, and `killed()` just observes that two hardcoded constants differ. It passes unconditionally and is unaffected by any change to the profile oracle, the vector, or the subjects.

Compare the neighbours: `output_invalidation_sentinel` and `expiration_sentinel` hardcode only the *mutant* disposition and compute the baseline via `evaluate(&input)`; `publication_ack_sentinel` runs a real `ReferenceModelSubject` both ways; the lifecycle sentinels go through `next_state` / `next_state_with_mutation`. The baseline here should likewise come from running something real - e.g. run `group-profile-update/v1` and take the observed `(group_name, group_description)` as the baseline, with the mutant being the pre-update projection, so the sentinel actually dies if the projection stops propagating.

As written it also makes `README.md` ("ten ... mutants with minimal deterministic witnesses") and the `MUTATION_MATRIX.md` row (witness: "group-profile-update/v1 across engine, app-runtime, and process subjects") inaccurate.

crates/cgka-conformance-simulator/src/scenario_authoring.rs
```diff
@@ -91,7 +91,7 @@ pub fn compile_authoring_scenario(
     let expanded = expand_sequence(&authored.steps)?;
     let canonical = ScenarioSpec {
         name: authored.name.clone(),
-        spec_version: SCENARIO_IR_VERSION.into(),
+        spec_version: SCENARIO_IR_V2_VERSION.into(),
```
erskingardner: Blocking: pinning the lowered `spec_version` to v2 makes `update_group_profile` unreachable from authoring documents. `ScenarioFlow::Action` holds a plain `ScenarioStep`, so an author can write the step, and the `compile_scenario` call below will then always reject it with `update_group_profile requires ScenarioSpec version 3` - an error pointing at a version the authoring compiler never emits.

Either select the version from the expanded steps (v3 when any v3-only step is present, v2 otherwise), or reject `UpdateGroupProfile` in `expand_flow` with a message the author can act on.

crates/cgka-conformance-simulator/src/scenario_ir.rs
```diff
@@ -14,7 +14,10 @@ use crate::{
     SubjectFailureCategory, required_capabilities,
 };
 
-pub const SCENARIO_IR_VERSION: &str = "2";
+/// Current canonical Scenario IR version for newly authored full-profile scenarios.
+pub const SCENARIO_IR_VERSION: &str = "3";
```
erskingardner: Non-blocking: `SCENARIO_IR_VERSION` is `pub use`d from `lib.rs`, and its meaning silently changes from "the only supported version" to "the newest version". Any consumer doing `spec.spec_version == SCENARIO_IR_VERSION` - which is exactly what `compile_scenario` did before this PR - now rejects perfectly valid v2 documents, with no compile error to warn them. `SCENARIO_IR_LATEST_VERSION` or `SCENARIO_IR_V3_VERSION` alongside `SCENARIO_IR_V2_VERSION` would make the break visible.

crates/cgka-conformance-simulator/tests/scenario_ir.rs
```diff
@@ -63,10 +63,67 @@ fn schema_declares_every_executable_step_kind() {
                 .expect("step kind")
         })
         .collect::<BTreeSet<_>>();
-    let executable_kinds = ScenarioStep::KINDS.iter().copied().collect::<BTreeSet<_>>();
+    let executable_kinds = ScenarioStep::KINDS
+        .iter()
+        .copied()
+        .filter(|kind| *kind != "update_group_profile")
```
erskingardner: Non-blocking: this turns an exhaustiveness check into one with a hand-maintained exclusion list, and nothing replaces it on the v3 side - `v3_schema_adds_full_group_profile_updates_without_rewriting_v2` only asserts three literal values, it never checks that the v3 schema covers every kind. Each future v3-only step needs a manual entry here and stays unchecked. Deriving the exclusion set from a single declaration of which kinds are v3-only (the same one `validate_step_version` needs) would keep both honest.

crates/cgka-conformance-simulator/src/scenario_ir.rs
```diff
@@ -280,11 +304,23 @@ fn validate_step(
...
             validate_client(step_index, client, clients, step.kind())?;
+            if let ScenarioStep::UpdateGroupProfile {
+                name, description, ..
+            } = step
+                && name.is_none()
+                && description.is_none()
+            {
+                return Err(compile_error(
+                    Some(step_index),
+                    "group profile update must change a name or description".into(),
```
erskingardner: Nit: re-matching `UpdateGroupProfile` with an `if let` inside a shared `|`-pattern arm reads awkwardly; a dedicated match arm ahead of the combined one would be clearer and would not need the second destructure.

Also worth noting this one rule now lives in three places: here, the `anyOf` in the v3 JSON schema, and `EngineError::Other("UpdateGroupData with no fields to change")` in `crates/cgka-engine/src/update_group_data.rs`. Failing fast at compile time is defensible, but the schema copy is unenforced (nothing validates vectors against the schemas), so it is one more thing to keep in sync by hand.

erskingardner (commented):
The underlying goal is right: an oracle that only checks clients *agree* is weak, and pinning the expected profile values is a real improvement. Engine support for partial profile updates (preserve-on-`None`) already exists in `crates/cgka-engine/src/update_group_data.rs`, so plumbing it through the IR is the correct place to close the gap. `ClientsConverged` comparing both fields and the `GroupProfile` expectation are good, well-tested additions.

Two things should be fixed before this lands, plus one design question.

**1. The new mutation sentinel is a tautology.** `group_profile_projection_sentinel` returns two hardcoded string literals. Nothing is executed, no rule is mutated, and `killed()` compares constants that were written to differ. Every other sentinel in that module derives at least the baseline from real logic (`select`, `evaluate`, `LifecycleModel::next_state`, a real `ReferenceModelSubject`). This one is guaranteed to pass no matter what the profile oracle does — including if `TraceExpectation::GroupProfile` were deleted tomorrow. In a crate whose entire purpose is measuring test adequacy, a sentinel that cannot fail is worse than no sentinel: `README.md` now advertises "ten ... mutants with minimal deterministic witnesses" and `MUTATION_MATRIX.md` names `group-profile-update/v1 across engine, app-runtime, and process subjects` as the witness, neither of which this code does.

**2. The authoring path dead-ends on the new action.** `compile_authoring_scenario` pins `spec_version` to v2, but `ScenarioFlow::Action` carries a plain `ScenarioStep`, so an authoring document *can* contain `update_group_profile` — and it will then always fail compilation with `update_group_profile requires ScenarioSpec version 3`. There is no way to author the new action through the human-facing input at all. Either lower to v3 when a v3-only step is present, or reject it at the authoring layer with a message that tells the author what to do instead.

**3. Is the v2/v3 split earning its keep?** `update_group_data` could have gained an optional `description` field in v2: existing vectors omit it and stay byte-identical, and the vectors are not validated against the JSON schemas anywhere in the test suite (`vector_artifacts.rs` only checks manifest/byte-fixture shape), so the schema files are documentation. That version would delete `scenario-ir.v3.schema.json`, `SCENARIO_IR_V2_VERSION`, `validate_step_version`, the dual-version match in `compile_scenario`, the `update_group_profile` filter in the schema-exhaustiveness test, and most of the doc churn across six markdown files — while keeping the oracle work, which is the part that has value. "V2 stays semantically unchanged" is a self-imposed constraint here, not an external compatibility requirement. If there is a reason it is not (external replayers pinned to the v2 schema?), say so in the PR body, because it is the thing that justifies roughly half the diff.

Non-blocking notes are inline.

crates/cgka-conformance-simulator/src/mutation_adequacy.rs
```diff
@@ -92,6 +94,24 @@ pub async fn run_mutation_sentinel(mutation: SemanticMutation) -> MutationSentin
...
+    (
+        format!(
+            "name:{}:description:{}",
+            baseline_projection.0, baseline_projection.1
+        ),
+        format!(
+            "name:{}:description:{}",
+            mutant_projection.0, mutant_projection.1
+        ),
+    )
```
erskingardner, line 140: Blocking: this sentinel never executes simulator or oracle code. Both observations are literals, so `baseline != mutant` is unconditionally true and `every_targeted_semantic_mutation_is_killed` cannot catch a real regression.

Please make this witness the actual kill path, e.g. run `compare_trace_expectations` (or equivalent) against:
1. observations equal to the requested profile (baseline: no failures), and
2. observations that still agree with each other but remain on the pre-update profile (mutant: `group_profile_mismatch` / failed convergence).

If that is not what this catalog row is for, remove `GroupProfileProjection` from the mutation enum/matrix rather than padding the count.

crates/cgka-conformance-simulator/MUTATION_MATRIX.md
```diff
@@ -15,3 +15,4 @@ observation and keeps this table exactly aligned with the executable catalog.
 | `output_invalidation` | Losing-branch projected output survives | Shared reference evaluation plus application oracle | Losing-app invalidation and exact projection checks |
 | `publication_acknowledgement` | Accepted publication remains pending | Real `ReferenceModelSubject` acknowledgement plus structural pending-work observation | Lost/duplicate acknowledgement and no-pending-work checks |
 | `retained_history_expiration_boundary` | Exact retention boundary expires early | Shared reference evaluation plus canonicalizer boundary | Past-epoch, proposal-expiry, and retained-relay boundary tests |
+| `group_profile_projection` | Accepted profile commit advances while the public projection remains stale | Portable expected-value group-profile oracle | group-profile-update/v1 across engine, app-runtime, and process subjects |
```
erskingardner: The “Production-shaped sentinel” cell overstates coverage: portable vectors run only through `EngineHarnessSubject` via `run_vector_fixture_report`. App-runtime/process get profile fields from the cross-adapter smoke test and the quiet-offline app-runtime test, not from `group-profile-update/v1` itself.

Please narrow this cell (and/or add a real multi-subject run of the vector) so the matrix matches what CI actually executes.

crates/cgka-conformance-simulator/README.md
```diff
@@ -68,7 +70,7 @@ welcomes use NIP-59 gift wraps before the bus delivers them.
   revision, exercises unequal histories, freeze/settle, crash/restart, temporary resource failure, administrative
   progress, and losing-branch joiner repair, and emits a minimal assumption-labeled starvation trace whose projected
   action kinds are drift-checked against committed Scenario IR.
-- `mutation_adequacy` — nine simulator-only single-rule mutants with minimal deterministic witnesses. The generated
+- `mutation_adequacy` — ten simulator-only single-rule mutants with minimal deterministic witnesses. The generated
   catalog kills selector order, witness dedup/admission, cutoff, frozen-state, scheduler re-arm, invalidation,
   publication-ack, and retention-boundary mutations without compiling mutation switches into production. Lifecycle
```
erskingardner: Nit: the count was bumped to “ten”, but the kill-list prose still ends at retention-boundary and never mentions group-profile projection. Update the list or drop the catalog row if the sentinel is removed.

erskingardner (commented):
Found serious issues — would normally request changes, but GitHub blocks that on your own PR.

**Premise:** Sound. Closing the “two clients agree on a stale profile” oracle gap is worth doing now, and keeping v2 name-only semantics intact is the right compatibility choice.

**What looks good:**
- v2/v3 split with compile-time rejection of `update_group_profile` in v2
- omit-to-preserve / explicit-empty clear contract, wired through engine, app-runtime, process, and reference subjects
- `group_profile` expectation + stronger `clients_converged` comparison, with unit tests that catch shared-but-wrong projections
- `group-profile-update/v1` exercises partial updates in both directions

**Blocking before merge:** `group_profile_projection_sentinel` always returns two hardcoded unequal strings, so `killed()` cannot fail. That breaks the mutation-matrix contract (“simulator-only alternate rules over minimal deterministic witnesses”). Please either drive the sentinel through the real oracle/compare path (baseline requested profile vs stale shared projection), or drop the catalog row and keep the vector/unit-test coverage on its own.

Also fix the matrix claim that `group-profile-update/v1` runs across engine/app-runtime/process — `run_vector_fixture_report` only builds `EngineHarnessSubject`; cross-adapter coverage is the separate process-orchestrator smoke test.

crates/cgka-conformance-simulator/src/oracle.rs
```diff
@@ -455,7 +455,7 @@ pub fn trace_behaviors(trace: &ScenarioTrace) -> Vec<OracleBehavior> {
     if evidence.admin_policy_observations > 0 {
         behaviors.insert(OracleBehavior::AdminPolicyObserved);
     }
-    if evidence.delivered_payloads > 0 {
+    if evidence.delivered_payloads > 0 || evidence.decryptability_probe_edges_delivered > 0 {
```
erskingardner, line 458: This plus the matching `expectation_behaviors` insert makes a successful probe count as `DeliveredPayload` on both sides of the report. The motivation is clear - `group-profile-update.v1.json` has a `probe_bidirectional_decryptability` step and no `send_app_message`, so `weak_oracle_warnings` fires on the `AppMessage` stimulus - but the fix is broader than the cause.

The actual cause is one line up in `scenario_stimuli`: `ProbeBidirectionalDecryptability` inserts **both** `ScenarioStimulus::AppMessage` and `ScenarioStimulus::BidirectionalDecryptabilityProbe`, even though the probe already has its own stimulus with its own recommended behavior. Not inserting `AppMessage` there fixes the new vector by deletion and leaves `send_app_message` as the only source of that stimulus.

As written, a scenario that has a real `send_app_message` whose payload never arrives, plus any probe expectation, now satisfies `DeliveredPayload` in both `oracle_behaviors` and `observed_behaviors` - so the weak-oracle warning and `missing_observed_behaviors` both go quiet on a genuine delivery gap. That is the same shape of oracle weakening the rest of this PR is closing. No committed vector combines the two today (I checked all 18), so the blast radius is zero right now and non-zero the moment someone writes one.

crates/cgka-conformance-simulator/src/scenario_ir.rs
```diff
@@ -14,7 +14,14 @@ use crate::{
     SubjectFailureCategory, required_capabilities,
 };
 
-pub const SCENARIO_IR_VERSION: &str = "2";
+/// Stable canonical version for the original Scenario IR action set.
+pub const SCENARIO_IR_V2_VERSION: &str = "2";
+/// Canonical version that adds full group-profile updates.
+pub const SCENARIO_IR_V3_VERSION: &str = "3";
+/// Newest Scenario IR version emitted when newly authored actions require it.
+pub const SCENARIO_IR_LATEST_VERSION: &str = SCENARIO_IR_V3_VERSION;
```
erskingardner, line 22: Nit: `SCENARIO_IR_LATEST_VERSION` has no callers anywhere in the tree - it is defined here and re-exported from `lib.rs`, and that is the whole of its use. `minimum_ir_version` returns `SCENARIO_IR_V3_VERSION` directly. The rename to `SCENARIO_IR_V3_VERSION` was the part that addressed the earlier note; this alias can go.

crates/cgka-conformance-simulator/src/scenario_ir.rs
```diff
@@ -210,6 +222,36 @@ fn is_group_scoped(step: &ScenarioStep) -> bool {
...
+        SCENARIO_IR_V2_VERSION
+    }
+}
+
+fn step_requires_v3(step: &ScenarioStep) -> bool {
+    let executable = match step {
+        ScenarioStep::InGroup { action, .. } => action.as_ref(),
+        step => step,
+    };
+    SCENARIO_IR_V3_ONLY_STEP_KINDS.contains(&executable.kind())
```
erskingardner, line 252: Nit: `ScenarioStep::kind()` already unwraps `InGroup` (`scenario.rs:300`, first match arm), so the local destructure is a no-op. The body is just:

```rust
fn step_requires_v3(step: &ScenarioStep) -> bool {
    SCENARIO_IR_V3_ONLY_STEP_KINDS.contains(&step.kind())
}
```

crates/cgka-conformance-simulator/src/mutation_adequacy.rs
```diff
@@ -92,6 +95,51 @@ pub async fn run_mutation_sentinel(mutation: SemanticMutation) -> MutationSentin
...
+    let mutant_failures = fixture.compare_observed_trace(&mutant_observed);
+    assert!(
+        mutant_failures
+            .iter()
+            .all(|failure| failure.kind == "group_profile_mismatch"),
+        "profile projection mutant must be rejected specifically by the group-profile oracle: {mutant_failures:#?}"
+    );
+    assert!(
+        !mutant_failures.is_empty(),
+        "profile projection mutant unexpectedly survived"
```
erskingardner, line 131: Two small things now that the sentinel does real work:

1. The asserts are in the wrong order - `all(...)` is vacuously true on an empty vec, so if the mutant ever survived you would get "unexpectedly survived" only after the more specific assert had already passed. Swapping them makes the failure message match the failure.
2. `compare_observed_trace` also emits `trace_mismatch` when `expected_trace` is `Some`. `group-profile-update.v1.json` has no `expected_trace` today, so the "all failures are `group_profile_mismatch`" assert holds - but it silently couples this sentinel to that property of the vector. A comment, or filtering to profile failures rather than asserting over all of them, would keep it from breaking for an unrelated reason.

Also minor: the returned baseline is the literal `"expectation_failures:none"` rather than something derived from `baseline_failures`. It is backed by the assert above, so it is not the tautology the old version was, but `format!("expectation_failures:{}", baseline_failures.len())` would make the returned pair carry the same evidence the asserts do.

erskingardner (commented):
Re-review of 5e2d5c5. All four prior findings are genuinely fixed, not papered over.

**Resolved:**
- `group_profile_projection_sentinel` now runs `group-profile-update/v1` through `run_vector_fixture_report`, asserts the baseline is clean against the real vector oracle, mutates only `group_name`/`group_description` on the observations, and asserts the mutant is rejected *specifically* by `group_profile_mismatch`. That is a real kill path: delete `TraceExpectation::GroupProfile` and this panics. Good fix.
- `compile_authoring_scenario` now lowers via `minimum_ir_version(&steps)`, with a test. `step_requires_v3` also unwraps `InGroup`, which closes a hole the original `matches!` in `validate_step_version` had.
- `SCENARIO_IR_VERSION` -> `SCENARIO_IR_V3_VERSION` makes the semantic change visible at the call site.
- The schema-exhaustiveness exclusion list is now derived from `SCENARIO_IR_V3_ONLY_STEP_KINDS`, and `v3_schema_adds_full_group_profile_updates_without_rewriting_v2` checks v2 union v3 == `ScenarioStep::KINDS`. That is stronger than what was there before the PR.
- MUTATION_MATRIX witness cell and the README kill-list now match what CI actually runs; `publish_group_evolution` label fixed.

**On the v2/v3 split:** the SCENARIO_IR.md paragraph (persisted vectors/capsules exchanged with external campaign runners) is the external constraint I was asking for. Accepted.

**One new concern**, in the `oracle.rs` change that arrived with this commit: making a bidirectional probe count as `DeliveredPayload` on both the expected and observed sides is broader than the problem, and it weakens the AppMessage weak-oracle warning globally. Details inline, along with three small cleanups. None of it blocks, but the oracle one is cheap to narrow now (only the new vector is affected today) and expensive to notice later.

Would approve, but GitHub does not allow approving your own PR.

PR #1300, opened 2026-08-07
Preserve own sent messages through convergence

erskingardner:
## Summary

Prevent locally authored application messages from being reclassified as invalid when they are included in Android's stored-convergence pass.

The send path now persists a minimal local convergence stamp containing only the authenticated sender and MLS source-epoch authenticator. Candidate replay carries that stamp across retained-anchor rollback and credits an own message only to candidate branches whose authenticated epoch state matches the send-time stamp.

## Root cause

OpenMLS sender ratchets are encryption-only. During replay, a device cannot decrypt and authenticate its own private-message ciphertext, and OpenMLS reports `OwnPrivateMessage` without authenticated branch provenance.

MDK previously converted that result to `Ignored`. Canonicalization then saw no branch that could decrypt the stored `Sent` row, classified it as `UndecryptableInCanonicalState`, persisted `EpochInvalidated`, and emitted `AppMessageInvalidated`. Android rendered that event as the “This message is no longer valid” banner on the sender's own message.

## Changes

- persist backward-compatible own-application provenance with signed outbound OpenMLS rows
- retain only sender and epoch-authenticator provenance; plaintext-derived payload references remain transient
- preload provenance before retained-anchor rollback so restart and historical replay retain it
- track epoch authenticators for each materialized candidate
- accept an own app only on its matching branch, including contested forks
- transition accepted own rows from `Sent` to `Processed` without emitting a duplicate received event
- emit canonicalization message-state audit transitions only after the atomic convergence apply commits
- skip audit-only message-row reads when the no-op recorder is installed
- preserve decoding of pre-existing unstamped rows

This prevents new occurrences of the original `UndecryptableInCanonicalState` misclassification for messages sent through the upgraded MDK. A locally sent message on a genuinely losing fork can still be invalidated with `LosingBranch`; Android may correctly render the same banner in that case. The change does not retroactively clear an invalidation already projected into an existing application timeline.

## Validation

- `just fast-ci`
- `cargo test -p cgka-engine --features test-policy-overrides --test distributed_convergence` — 83 passed
- `cargo test -p cgka-traits`
- `cargo test -p marmot-forensics` — 28 passed
- focused restart, matching-fork provenance, no-echo, audit-transition, recorder gating, payload-reference format, and legacy serialization regressions

### DIFF
--- a/crates/cgka-engine/src/app_payload.rs
+++ b/crates/cgka-engine/src/app_payload.rs
@@ -1,4 +1,11 @@
 use cgka_traits::{EngineError, MarmotAppEvent, MemberId};
+use sha2::{Digest, Sha256};
+
+/// Stable transient reference used when convergence reports a decrypted
+/// application payload without retaining the plaintext itself.
+pub(crate) fn decrypted_payload_ref(payload: &[u8]) -> String {
+    format!("sha256:{}", hex::encode(Sha256::digest(payload)))
+}
 
 /// Shared application-payload sender validation for every inbound seam
 /// (direct ingest, stored-convergence/replay). An application message is
@@ -25,7 +32,7 @@ pub(crate) fn validate_app_payload_for_sender(
 
 #[cfg(test)]
 mod tests {
-    use super::validate_app_payload_for_sender;
+    use super::{decrypted_payload_ref, validate_app_payload_for_sender};
     use cgka_traits::{EngineError, MarmotAppEvent, MemberId};
 
     fn payload_from(pubkey: &str) -> Vec<u8> {
@@ -65,5 +72,13 @@ mod tests {
         let event =
             validate_app_payload_for_sender(&payload, &sender).expect("matching sender validates");
         assert_eq!(event.content, "hello");
+    }
+
+    #[test]
+    fn decrypted_payload_reference_format_is_stable() {
+        assert_eq!(
+            decrypted_payload_ref(b"hello"),
+            "sha256:2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824"
+        );
     }
 }
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -1336,16 +1336,20 @@ impl<S: StorageProvider> Engine<S> {
         // Snapshot their prior values first, then emit forensic transitions
         // only after that transaction commits so Goggles never observes a
         // state change that rolled back.
-        let disposition_transitions = openmls_canonicalization_dispositions(&result)?
-            .into_iter()
-            .map(|disposition| {
-                let record = self
-                    .storage
-                    .get_message(&disposition.message_id)
-                    .map_err(storage_projection_error)?;
-                Ok((disposition, record.state, record.epoch))
-            })
-            .collect::<Result<Vec<_>, OpenMlsProjectionError>>()?;
+        let disposition_transitions = if self.recorder.is_enabled() {
+            openmls_canonicalization_dispositions(&result)?
+                .into_iter()
+                .map(|disposition| {
+                    let record = self
+                        .storage
+                        .get_message(&disposition.message_id)
+                        .map_err(storage_projection_error)?;
+                    Ok((disposition, record.state, record.epoch))
+                })
+                .collect::<Result<Vec<_>, OpenMlsProjectionError>>()?
+        } else {
+            Vec::new()
+        };
         let observations = self.storage.with_transaction(|storage| {
             let observations = apply_openmls_canonicalization_result_with_profile_policy(
                 storage,
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -35,7 +35,8 @@ use crate::engine::Engine;
 use crate::openmls_projection::{
     OpenMlsContentKind, OpenMlsProjectionError, OpenMlsReplayObservation, ReplayProfilePolicy,
     StoredCanonicalizationOptions, apply_openmls_canonicalization_result_with_profile_policy,
-    canonicalize_stored_openmls_messages_with_profile_policy, project_mls_message,
+    canonicalize_stored_openmls_messages_with_profile_policy,
+    openmls_canonicalization_dispositions, project_mls_message,
     retain_current_group_epoch_snapshot, retire_stale_convergence_deferred_commits,
 };
 use std::sync::atomic::{AtomicU64, Ordering};
@@ -1331,6 +1332,20 @@ impl<S: StorageProvider> Engine<S> {
         let mut completed_pass = pass.clone();
         completed_pass.phase = ConvergencePassPhase::Completed;
         completed_pass.fairness_slot_available = true;
+        // The disposition writes are part of the atomic convergence apply.
+        // Snapshot their prior values first, then emit forensic transitions
+        // only after that transaction commits so Goggles never observes a
+        // state change that rolled back.
+        let disposition_transitions = openmls_canonicalization_dispositions(&result)?
+            .into_iter()
+            .map(|disposition| {
+                let record = self
+                    .storage
+                    .get_message(&disposition.message_id)
+                    .map_err(storage_projection_error)?;
+                Ok((disposition, record.state, record.epoch))
+            })
+            .collect::<Result<Vec<_>, OpenMlsProjectionError>>()?;
         let observations = self.storage.with_transaction(|storage| {
             let observations = apply_openmls_canonicalization_result_with_profile_policy(
                 storage,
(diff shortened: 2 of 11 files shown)


erskingardner (commented):
## Verdict

The core fix is sound. I traced the provenance path end to end and could not find a correctness bug in it. Five things below are worth changing, none of them blocking.

### Verified correct

- **The two epoch authenticators being compared really are the same value.** `own_commit_post_merge_epoch_authenticator` is `hex::encode(merged.epoch_authenticator())` (`crates/cgka-engine/src/openmls_projection.rs:539`) and the send stamp uses the identical expression, so the match in `app_messages_by_id` is meaningful rather than vacuously false.
- **The `unwrap_or(*source_epoch <= candidate.fork_epoch)` fallback is only reachable for epochs below the replay start.** `epoch_authenticators` is seeded with the replay-start epoch and gains an entry at every merged resulting epoch, so it is contiguous over the replayed range — a source epoch inside that range is always present and gets the real comparison. Below the range it is genuine pre-fork common history, and `fork_epoch >= replay_start` makes the fallback true. Right answer in both directions.
- **Stamp coverage is complete.** `create_message` has exactly one call site (`crates/cgka-engine/src/message_processor/send.rs:875`), so no queued or retried application send can bypass the stamp.
- **The stamp survives everything that rewrites the row.** `update_message_state` is a read-modify-write that preserves the payload blob, and `stamp_processed_own_commit_record` explicitly carries `own_application_stamp` through.
- **Backward compatibility holds in both directions.** `StoredMessagePayload` is adjacently tagged (`tag = "kind", content = "message"`), so the `legacy_json["message"].get("own_application_stamp")` assertion in the new snapshot test is actually testing something rather than indexing into a null. Old binaries reading new rows ignore the unknown field.
- **Accepting own apps into the apply replay cannot fail a pass.** Application-message `process_message` errors are already funnelled to `Ignored` for `ValidationError` and `UseAfterEviction`.

Locally: both new convergence tests pass, `cargo test -p cgka-traits --test snapshots` passes (23), clippy clean on both crates.

### Findings

**1. The `sha256:` payload-ref format is duplicated across two modules with nothing tying them together.**

`crates/cgka-engine/src/message_processor/send.rs:85` builds `format!("sha256:{}", hex::encode(Sha256::digest(payload)))` and `crates/cgka-engine/src/openmls_projection.rs:3132` independently builds `format!("sha256:{}", hex::encode(message_digest(payload)))`. They agree today because `message_digest` is also SHA-256, but nothing enforces it and no test compares the two. If either side changes, a sender's own `decrypted_payload_ref` silently stops matching the ref every receiver computes for the same plaintext, and the divergence surfaces only as mismatched forensic evidence. Extract one helper and call it from both.

**2. Every convergence pass now does an extra `get_message` per disposition, whether or not a recorder is installed.**

The snapshot loop at `crates/cgka-engine/src/distributed_convergence.rs:1339` deserializes each full `MessageRecord` — payload blob included — outside the transaction, purely to have a `previous_state` for the audit line. On a group with a large pending set that is real work on the hot path, and with the default `NoopRecorder` it is entirely wasted. Worth gating the snapshot on recorder presence.

**3. A digest of validated plaintext is now persisted at rest.**

`decrypted_payload_ref` previously lived only in transient observations and events; it is now durable in the sender's `Sent` row. Under SQLCipher the practical risk is low, but it is a new plaintext-derived artifact on disk and it enables confirmation-of-content against a compromised local database. Since `local-artifact-safety.md` is a named repo invariant this deserves to be a conscious decision rather than a side effect — and the stamp does not actually need the ref for branch matching, only the authenticator does.

**4. The "no duplicate `MessageReceived`" guarantee is load-bearing but undocumented.**

It works because `crates/cgka-engine/src/openmls_projection.rs:2285` passes `PrevalidatedOwnCommits::default()` into the apply replay, so own apps miss the new stamp shortcut, reach OpenMLS, and come back as `OwnPrivateMessage` → `Ignored`. The existing comment there explains why own *commits* are not prevalidated and says nothing about applications. Anyone who later threads real `own_commits` into that call to fix something else would start echoing the sender's own messages back as received, and the only thing catching it would be one assertion buried in a new test. Add a sentence to that comment.

**5. Minor: the new if/else in `record_sent_openmls_message_with_application_stamp` does not need to branch on the persist call.**

`persist_signed_openmls_wire_message` is just a wrapper around `persist_stored_message_payload` with a differently-constructed payload, so both arms can collapse into one call whose payload comes from a `match`. The `application_stamp.clone()` at `crates/cgka-engine/src/message_processor/store.rs:224` is also unnecessary — the value is owned and unused afterward.

### One note for the PR description

Own sent messages can still produce `AppMessageInvalidated`, now via `LosingBranch` rather than `UndecryptableInCanonicalState`. That is correct — the message genuinely was on a losing branch — but Android will render the same banner, so anyone validating this fix against the original report should know the banner is not expected to disappear in every case.

erskingardner (commented):
## Re-review of `6a52629c`

All five findings addressed, and two of them more thoroughly than I suggested.

**Findings 1 and 3 were solved together, better than proposed.** Rather than just extracting a shared `sha256:` helper, this drops `decrypted_payload_ref` from `OwnApplicationConvergenceStamp` entirely. The own-message path no longer computes a payload ref at all, so there is nothing left to diverge — and the durable plaintext digest is gone from the `Sent` row, which was the actual concern in finding 3. The receiver-side ref moved to a single `app_payload::decrypted_payload_ref` helper with a `decrypted_payload_reference_format_is_stable` unit test pinning the exact output. Stronger than sharing a helper between two callers would have been.

**Finding 2** is handled by a defaulted `ForensicRecorder::is_enabled()` returning `true`, overridden to `false` on `NoopRecorder`, gating the disposition snapshot. Adding a defaulted method keeps the public trait non-breaking for out-of-workspace implementors. The workspace has three implementors and no delegating/decorator recorder, so `CapturingRecorder` correctly inherits `true` and nothing needs to forward the call. Both branches got a direct assertion in the forensics tests.

**Findings 4 and 5** are straightforward: the load-bearing reason own-application stamps stay out of the apply replay is now spelled out in the comment at `apply_openmls_canonicalization_result_inner`, and the persist branch collapses to one `persist_stored_message_payload` call with the payload chosen by a `match`. `persist_signed_openmls_wire_message` was deleted rather than left orphaned.

### One consequence worth recording

Because the stamp no longer carries a ref, `AppMessageBranches::decrypted_payload_ref` became `Option<String>`, so an own sent app that is invalidated surfaces `AppMessageInvalidated { decrypted_payload_ref: None }`, and conformance vectors record `payload_ref: null` for it. I traced every consumer — `canonicalization.rs`, the UniFFI event conversion, and `vector.rs` — and the field is informational everywhere; nothing branches on it, and `message_id` remains the correlation key. It also matches master's behavior, since own apps never reached `app_messages_by_id` at all before this PR. Not a regression, but now a deliberate design point rather than an accident.

I also checked whether gating the pre-read behind `is_enabled()` changes failure semantics, since `openmls_canonicalization_dispositions` can fail on `message_id_from_hex`. With a recorder installed that error aborts the pass before the apply; without one it aborts inside it, via the identical call in `persist_openmls_canonicalization_dispositions`. Both abort the pass, so there is no divergence that matters.

### Verification on `6a52629c`

- `cargo fmt --all -- --check` clean
- clippy clean with `--all-targets` across `cgka-engine`, `cgka-traits`, `marmot-forensics`
- `distributed_convergence` 83/83
- `fork_detection` 12/12
- conformance `openmls_replay_probe` 14/14
- `marmot-forensics` and `cgka-traits` suites green

Nothing left blocking from my side — this looks ready to merge.

erskingardner (commented):
## Verdict

Recommend merge. Independently traced the provenance path on `6a52629c` and did not find a correctness bug. (GitHub blocks a formal approve on own PRs, so this is a comment review with the same substance.)

## Verified

- **Root cause fix is the right shape.** OpenMLS cannot authenticate own private-message ciphertext on replay; a durable send-time epoch-authenticator stamp (parallel to own-commit stamps) is the minimal honest provenance, not a workaround.
- **Authenticator identity matches.** Send stamps `hex::encode(mls_group.epoch_authenticator())` before `create_message`; candidate maps use the same helper after merge/seed. The comparison in `app_messages_by_id` is meaningful.
- **Preload before retained-anchor rollback is load-bearing.** Group snapshot rollback restores message rows to the anchor and drops post-anchor rows for the probe; stamps must already be in `PrevalidatedOwnCommits` / `pending_messages` or restart/historical replay would regress to `OwnPrivateMessage` → undecryptable.
- **Fork filtering is correct.** Missing in-range authenticators fall back to `source_epoch <= fork_epoch` (common history only); post-divergence mismatched authenticators are excluded, so an own app witnesses only its real branch.
- **No echo on apply.** Apply still uses empty `PrevalidatedOwnCommits`, so stamped apps become `OwnPrivateMessage` → `Ignored` while dispositions alone mark `Sent` → `Processed`. Comment now documents that invariant.
- **Audit gating is sound.** `is_enabled()` default keeps external recorders observable; `NoopRecorder` skips the hot-path `get_message` snapshot; transitions emit only after the apply transaction commits.
- **Compat.** Serde-defaulted `own_application_stamp` keeps legacy rows decodable; omitting the payload digest from the durable stamp was the right privacy call.
- **CI** is green on this head, including the new convergence regressions.

## Non-blocking

Optional follow-up: a regression that puts an own stamped app on the *losing* fork and asserts `LosingBranch` (not `UndecryptableInCanonicalState`) would lock the remaining banner case the PR description calls out. Not required to land this.

Datawav:

No actionable findings.

_Advisory audit of head [`6a52629c368e`](https://github.com/marmot-protocol/mdk/commit/6a52629c368ef67c4f22ccd58b2548faf7c24427); no checkout or PR-code execution was performed._

Datawav (approved):
Automated multi-agent audit of head 6a52629c368e: no actionable findings. Ready for merge.

PR #1301, opened 2026-08-07
Allow iOS background access to account secrets

erskingardner:
## What was broken

The iOS keychain backend stored account signing secrets with a foreground-only accessibility policy. When a notification service extension or background refresh launched while the device was locked, MDK could discover the account but could not load its secret. The notification service extension consequently fell back to generic APNs content, and message catch-up waited for the foreground app.

## Root cause

The protected iOS keyring store defaulted to a when-unlocked policy. Its string modifier parser also maps the requested device-only after-first-unlock policy to the non-device-only variant, so MDK must use the store's typed credential API to preserve the intended protection.

## What changed

- Store new iOS account secrets with the typed `AfterFirstUnlockThisDeviceOnly` policy.
- Use a versioned keychain service because access-control attributes cannot be changed in place.
- Migrate legacy entries on first successful unlocked access by writing the replacement, reading it back for verification, and only then attempting legacy cleanup.
- Treat legacy cleanup as best-effort after successful writes, loads, and migration so an unavailable stale item cannot break locked-device background access.
- During account removal, attempt both current and legacy deletions even if either one fails and report combined failure when necessary.
- Fall back to legacy entries for account discovery until migration can run.
- Preserve existing behavior on non-iOS platforms.
- Compile and lint the iOS account keychain path on every PR using a macOS CI runner.
- Update the crate map for the expanded keyring module responsibilities.

## Impact

New iOS identities are immediately usable by the notification service extension and scheduled background work after the device has been unlocked once following boot. Existing identities migrate automatically the next time MDK loads their secret while unlocked. The migration is crash-safe, stale cleanup cannot invalidate an otherwise successful operation, and new entries remain device-bound.

## Validation

- Direct equivalent of `just fast-ci` (the local environment did not have `just` installed)
- `cargo clippy -p marmot-account --target aarch64-apple-ios --locked -- -D warnings`
- `cargo test -p marmot-account --locked` (94 tests)
- `actionlint .github/workflows/ci.yml`
- Generated Swift bindings were built for iOS device and simulator and consumed by a signed White Noise staging device build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,6 +125,34 @@ jobs:
       - name: Run clippy with all features
         run: cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
 
+  ios-account-clippy:
+    name: iOS account keychain clippy
+    runs-on: macos-latest
+    timeout-minutes: 30
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          persist-credentials: false
+
+      - name: Install Rust toolchain
+        shell: bash
+        run: |
+          set -euo pipefail
+          toolchain="$(sed -n 's/^channel = "\(.*\)"/\1/p' rust-toolchain.toml)"
+          rustup toolchain install "$toolchain" --profile minimal --component clippy
+          rustup default "$toolchain"
+          rustup target add aarch64-apple-ios
+          rustc --version
+          cargo --version
+
+      - name: Cache Cargo artifacts
+        uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
+
+      - name: Check iOS account keychain code
+        run: cargo clippy -p marmot-account --target aarch64-apple-ios --locked -- -D warnings
+
   # The broader integration matrix explicitly enables test-policy-overrides.
   # This focused default-feature job proves ordinary debug/release consumers
   # reject every non-v1 convergence policy (mdk#970).
--- a/crates/marmot-account/src/keyring.rs
+++ b/crates/marmot-account/src/keyring.rs
@@ -4,6 +4,9 @@ use std::sync::{Arc, Mutex};
 
 use crate::error::{AccountHomeError, AccountHomeResult};
 
+#[cfg(target_os = "ios")]
+const IOS_BACKGROUND_KEYRING_SERVICE_SUFFIX: &str = ".after-first-unlock-v1";
+
 pub(crate) fn initialize_keyring_store() -> AccountHomeResult<()> {
     static KEYRING_STORE_INIT: Mutex<()> = Mutex::new(());
     let _guard = KEYRING_STORE_INIT.lock().map_err(|_| {
@@ -13,6 +16,64 @@ pub(crate) fn initialize_keyring_store() -> AccountHomeResult<()> {
         return Ok(());
     }
     initialize_platform_keyring_store()
+}
+
+pub(crate) fn build_keyring_entry(
+    service: &str,
+    account: &str,
+) -> AccountHomeResult<keyring_core::Entry> {
+    keyring_core::Entry::new(&keyring_service_name(service), account).map_err(map_keyring_error)
+}
+
+#[cfg(target_os = "ios")]
+pub(crate) fn build_legacy_ios_keyring_entry(
+    service: &str,
+    account: &str,
+) -> AccountHomeResult<keyring_core::Entry> {
+    keyring_core::Entry::new(service, account).map_err(map_keyring_error)
+}
+
+#[cfg(target_os = "ios")]
+fn keyring_service_name(service: &str) -> String {
+    // Keychain access control is add-only. A versioned service lets migration
+    // verify the replacement before deleting the foreground-only legacy item.
+    format!("{service}{IOS_BACKGROUND_KEYRING_SERVICE_SUFFIX}")
+}
+
+#[cfg(not(target_os = "ios"))]
+fn keyring_service_name(service: &str) -> String {
+    service.to_owned()
+}
+
+pub(crate) fn write_keyring_secret(
+    service: &str,
+    account: &str,
+    secret: &str,
+) -> AccountHomeResult<()> {
+    #[cfg(target_os = "ios")]
+    {
+        use security_framework::access_control::{ProtectionMode, SecAccessControl};
+        use security_framework::passwords::{PasswordOptions, set_generic_password_options};
+
+        let mut options =
+            PasswordOptions::new_generic_password(&keyring_service_name(service), account);
+        options.use_protected_keychain();
+        let access_control = SecAccessControl::create_with_protection(
+            Some(ProtectionMode::AccessibleAfterFirstUnlockThisDeviceOnly),
+            0,
+        )
+        .map_err(|error| AccountHomeError::SecretStore(error.to_string()))?;
+        options.set_access_control(access_control);
+        set_generic_password_options(secret.as_bytes(), options)
+            .map_err(|error| AccountHomeError::SecretStore(error.to_string()))
+    }
+
+    #[cfg(not(target_os = "ios"))]
+    {
+        build_keyring_entry(service, account)?
+            .set_password(secret)
+            .map_err(map_keyring_error)
+    }
 }
 
 fn initialize_platform_keyring_store() -> AccountHomeResult<()> {
--- a/crates/marmot-account/src/keyring.rs
+++ b/crates/marmot-account/src/keyring.rs
@@ -1,4 +1,5 @@
-//! Platform keyring store initialization and keyring-error mapping.
+//! Platform keyring entry construction, secret writes, store initialization,
+//! and keyring-error mapping.
 
 use std::sync::{Arc, Mutex};
 
@@ -52,20 +53,20 @@ pub(crate) fn write_keyring_secret(
 ) -> AccountHomeResult<()> {
     #[cfg(target_os = "ios")]
     {
-        use security_framework::access_control::{ProtectionMode, SecAccessControl};
-        use security_framework::passwords::{PasswordOptions, set_generic_password_options};
+        use apple_native_keyring_store::protected::{AccessPolicy, Cred};
 
-        let mut options =
-            PasswordOptions::new_generic_password(&keyring_service_name(service), account);
-        options.use_protected_keychain();
-        let access_control = SecAccessControl::create_with_protection(
-            Some(ProtectionMode::AccessibleAfterFirstUnlockThisDeviceOnly),
-            0,
+        // apple-native-keyring-store 1.0.0's string modifier parser maps this
+        // device-only policy to AfterFirstUnlock. The typed API preserves it.
+        Cred::build(
+            &keyring_service_name(service),
+            account,
+            AccessPolicy::AfterFirstUnlockThisDeviceOnly,
+            None,
+            false,
         )
-        .map_err(|error| AccountHomeError::SecretStore(error.to_string()))?;
-        options.set_access_control(access_control);
-        set_generic_password_options(secret.as_bytes(), options)
-            .map_err(|error| AccountHomeError::SecretStore(error.to_string()))
+        .map_err(map_keyring_error)?
+        .set_password(secret)
+        .map_err(map_keyring_error)
     }
 
     #[cfg(not(target_os = "ios"))]
--- a/crates/marmot-account/src/secret_store.rs
+++ b/crates/marmot-account/src/secret_store.rs
@@ -199,10 +199,13 @@ impl KeychainSecretStore {
         Self::remove_entry(self.legacy_entry_for_account(account_id_hex)?)
     }
 
-    #[cfg(not(target_os = "ios"))]
-    fn remove_legacy_ios_secret(&self, _account_id_hex: &str) -> AccountHomeResult<()> {
-        Ok(())
+    #[cfg(target_os = "ios")]
+    fn cleanup_legacy_ios_secret(&self, account_id_hex: &str) {
+        let _ = self.remove_legacy_ios_secret(account_id_hex);
     }
+
+    #[cfg(not(target_os = "ios"))]
+    fn cleanup_legacy_ios_secret(&self, _account_id_hex: &str) {}
 
     #[cfg(target_os = "ios")]
     fn migrate_legacy_ios_secret(
@@ -239,7 +242,7 @@ impl KeychainSecretStore {
                 "iOS background keychain migration verification failed".into(),
             ));
         }
-        self.remove_legacy_ios_secret(&account.account_id_hex)?;
+        self.cleanup_legacy_ios_secret(&account.account_id_hex);
         Ok(keys)
     }
 }
@@ -282,7 +285,8 @@ impl AccountSecretStore for KeychainSecretStore {
             &account.account_id_hex,
             secret_key_hex.as_str(),
         )?;
-        self.remove_legacy_ios_secret(&account.account_id_hex)
+        self.cleanup_legacy_ios_secret(&account.account_id_hex);
+        Ok(())
     }
 
     fn load_secret(&self, account: &AccountSummary) -> AccountHomeResult<nostr::Keys> {
@@ -294,7 +298,7 @@ impl AccountSecretStore for KeychainSecretStore {
                 let secret_key = Zeroizing::new(secret_key);
                 let keys = nostr::Keys::parse(secret_key.as_str())
                     .map_err(|_| AccountHomeError::InvalidSecretKey)?;
-                self.remove_legacy_ios_secret(&account.account_id_hex)?;
+                self.cleanup_legacy_ios_secret(&account.account_id_hex);
                 Ok(keys)
             }
             #[cfg(target_os = "ios")]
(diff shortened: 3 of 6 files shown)


crates/marmot-account/src/secret_store.rs
```diff
@@ -206,9 +292,14 @@ impl AccountSecretStore for KeychainSecretStore {
         {
             Ok(secret_key) => {
                 let secret_key = Zeroizing::new(secret_key);
-                nostr::Keys::parse(secret_key.as_str())
-                    .map_err(|_| AccountHomeError::InvalidSecretKey)
+                let keys = nostr::Keys::parse(secret_key.as_str())
+                    .map_err(|_| AccountHomeError::InvalidSecretKey)?;
+                self.remove_legacy_ios_secret(&account.account_id_hex)?;
```
erskingardner: **Blocking:** After a successful read of the new background-capable item, legacy deletion is fallible and fails the whole `load_secret`.

If migration wrote/verified the replacement but crashed (or otherwise failed) before deleting the legacy `WhenUnlocked` item, both entries can coexist. Later, while the device is locked, the NSE can read the new item (`AfterFirstUnlock…`) and then hit `errSecInteractionNotAllowed` (-25308) deleting the legacy one. Upstream maps that to `PlatformFailure`, so this returns `Err` even though the secret was loaded — undermining the PR’s goal.

Same pattern on `write_secret` (L285) and `migrate_legacy_ios_secret` (L242): a successful new write followed by a failed legacy delete reports failure. Callers in `home.rs` may then roll back the account record while the new keychain item remains, leaving an orphan credential that `has_secret_for_account_id` will still see.

Make legacy cleanup best-effort after the new item is successfully written/loaded (ignore delete failure; keep retrying on later successful unlocked access).

crates/marmot-account/src/secret_store.rs
```diff
@@ -217,13 +308,8 @@ impl AccountSecretStore for KeychainSecretStore {
...
     fn remove_secret(&self, account: &AccountSummary) -> AccountHomeResult<()> {
-        match self
-            .entry_for_account(&account.account_id_hex)?
-            .delete_credential()
-        {
-            Ok(()) | Err(keyring_core::Error::NoEntry) => Ok(()),
-            Err(err) => Err(map_keyring_error(err)),
-        }
+        Self::remove_entry(self.entry_for_account(&account.account_id_hex)?)?;
+        self.remove_legacy_ios_secret(&account.account_id_hex)
```
erskingardner: **Blocking (related):** If deleting the new entry succeeds and legacy delete then fails, this returns `Err` with the legacy secret still present — account removal looks failed/incomplete and the signing key remains in the keychain.

Attempt both deletes and only succeed when neither entry remains (or combine errors after trying both), rather than short-circuiting on the first failure.

crates/marmot-account/src/keyring.rs
```diff
@@ -15,6 +18,64 @@ pub(crate) fn initialize_keyring_store() -> AccountHomeResult<()> {
...
+        let mut options =
+            PasswordOptions::new_generic_password(&keyring_service_name(service), account);
+        options.use_protected_keychain();
+        let access_control = SecAccessControl::create_with_protection(
+            Some(ProtectionMode::AccessibleAfterFirstUnlockThisDeviceOnly),
+            0,
+        )
+        .map_err(|error| AccountHomeError::SecretStore(error.to_string()))?;
+        options.set_access_control(access_control);
+        set_generic_password_options(secret.as_bytes(), options)
```
erskingardner: **Non-blocking:** The direct `security-framework` write is justified relative to `Entry::new_with_modifiers` — in `apple-native-keyring-store` 1.0.0, `determine_access_policy` maps `after-first-unlock-this-device-only` to `AccessPolicy::AfterFirstUnlock` (stripping device-only). Agreed that the string path is unusable.

You can still avoid the extra dependency and duplicated add/update logic by using the typed API that already sets this protection correctly:

```rust
use apple_native_keyring_store::protected::{AccessPolicy, Cred};

Cred::build(
    &keyring_service_name(service),
    account,
    AccessPolicy::AfterFirstUnlockThisDeviceOnly,
    None,
    false,
)
.map_err(map_keyring_error)?
.set_password(secret)
.map_err(map_keyring_error)
```

That keeps writes on the same store implementation as reads/deletes.

erskingardner (commented):
**Review status:** would request changes (GitHub blocks that event on your own PR).

The accessibility migration approach is sound (versioned service + verify-before-delete, and bypassing the upstream modifier string bug that maps `AfterFirstUnlockThisDeviceOnly` → `AfterFirstUnlock`), but legacy cleanup is currently a hard dependency of successful load/write. That can turn a partial migration into a background load failure, and can make `write_secret` report failure after the new item was already written.

Please make post-success legacy deletion best-effort on the load/write/migrate paths, and make `remove_secret` attempt both entries even if the first delete fails.

crates/marmot-account/src/secret_store.rs
erskingardner: **Blocking:** this `?` turns a fully successful secret load into a hard error whenever the legacy cleanup delete fails.

At this point the secret has been read, parsed, and validated — the caller's request has succeeded. Failing it because a best-effort tidy-up of a *stale duplicate* didn't work inverts the whole point of the migration.

The failure window is precisely the scenario this PR targets. On the first background wake after upgrade, the legacy item still exists and is `kSecAttrAccessibleWhenUnlocked`. `SecItemDelete` against an item whose protection class is currently unavailable can return `errSecInteractionNotAllowed` (-25308), which `apple-native-keyring-store`'s `decode_error` maps to `PlatformFailure` → `AccountHomeError::SecretStore`. Net effect: the NSE reads the new `AfterFirstUnlockThisDeviceOnly` secret successfully, then reports a secret-store failure and falls back to generic APNs content anyway.

Make the cleanup best-effort:

```rust
let _ = self.remove_legacy_ios_secret(&account.account_id_hex);
```

Secondary point on the same line: once migration has completed this issues a keychain delete on *every* successful load, forever, to catch a crash window that closes after one run. If you'd rather not pay that, only attempt cleanup when a legacy item plausibly exists (e.g. thread it out of `migrate_legacy_ios_secret`). Making it non-fatal is the part that matters; the extra round-trip is a judgement call.

crates/marmot-account/src/secret_store.rs
```diff
@@ -187,16 +255,34 @@ impl AccountSecretStore for KeychainSecretStore {
...
         let secret_key_hex = Zeroizing::new(keys.secret_key().to_secret_hex());
-        self.entry_for_account(&account.account_id_hex)?
-            .set_password(secret_key_hex.as_str())
-            .map_err(map_keyring_error)
+        write_keyring_secret(
+            &self.service_name,
+            &account.account_id_hex,
+            secret_key_hex.as_str(),
+        )?;
+        self.remove_legacy_ios_secret(&account.account_id_hex)
```
erskingardner: Same propagation problem as in `load_secret`, with a compensation consequence: `write_keyring_secret` has already durably stored the new secret by the time this line runs, but a failing legacy delete makes `write_secret` return `Err`.

Callers treat that as "the write failed" and roll back: `home.rs:241` and `home.rs:477` do `remove_secret` + `remove_dir_all` on the account dir, and `home.rs:1091` (`write_new_signing_account`) deletes the account record. So a transient cleanup error aborts account creation/import even though the credential landed. The rollback happens to leave consistent state (`remove_secret` now removes both entries), so this is lower severity than the `load_secret` case — but it's still a spurious hard failure on a path that succeeded.

```rust
let _ = self.remove_legacy_ios_secret(&account.account_id_hex);
Ok(())
```

crates/marmot-account/src/keyring.rs
```diff
@@ -15,6 +18,64 @@ pub(crate) fn initialize_keyring_store() -> AccountHomeResult<()> {
...
+}
+
+pub(crate) fn write_keyring_secret(
+    service: &str,
+    account: &str,
+    secret: &str,
+) -> AccountHomeResult<()> {
+    #[cfg(target_os = "ios")]
+    {
+        use security_framework::access_control::{ProtectionMode, SecAccessControl};
```
erskingardner: The comment on `keyring_service_name` explains the versioned service, but nothing in the code explains why this bypasses the keyring store entirely — and that's the surprising part, since `apple-native-keyring-store`'s `Store::build` documents an `access-policy` modifier that looks like it would do the job.

It doesn't, and the reason is a specific upstream bug worth pinning down here:

```rust
// apple-native-keyring-store 1.0.0, protected.rs:570
"after-first-unlock-this-device-only" | "afterfirstunlockthisdeviceonly" => {
    Ok(AccessPolicy::AfterFirstUnlock)   // device-only silently dropped
}
```

(`when-unlocked-this-device-only` is broken the same way.) So `Entry::new_with_modifiers(.., "access-policy" => "after-first-unlock-this-device-only")` would silently produce a non-device-bound item — worse than the status quo, since these secrets would then be restorable to another device.

Please put that in the comment, with a link to an upstream issue if you file one. Right now the next reader sees a hand-rolled Security.framework write sitting next to a store that appears to support the exact policy, and has no way to know the workaround is still needed or when it can be deleted.

erskingardner (commented):
**Verdict: request changes** (posted as a comment review since GitHub will not let the PR author self-request changes).

Premise and root cause check out. I verified the upstream claim: `apple-native-keyring-store` 1.0.0 `determine_access_policy` maps `after-first-unlock-this-device-only` to `AccessPolicy::AfterFirstUnlock` (drops the device-only variant), so the store's `access-policy` modifier genuinely cannot express the policy you need and the direct Security.framework write is justified. The versioned service + verify-then-delete migration is the right shape for an add-only keychain attribute.

One blocking issue: legacy cleanup errors are propagated on paths where the real work already succeeded, and the worst case lands in exactly the locked-device background scenario this PR exists to fix. Two inline comments on that, plus two smaller notes.

One more thing not anchorable to a diff line: **none of the new `#[cfg(target_os = "ios")]` code is compiled by PR CI.** `.github/workflows/ci.yml` has no Apple target, and `bindings.yaml` (the only workflow that does `rustup target add aarch64-apple-ios`) triggers on `push`, not `pull_request`. So the entire iOS block here — the security-framework write path, the migration, the legacy fallbacks — can break on master and only surface at bindings/release time. You ran `cargo clippy -p marmot-account --target aarch64-apple-ios` locally, which is the right check; it would be worth adding it to `ci.yml` on a macOS runner (or to `just fast-ci`) so it stays true. Not strictly this PR's job, but this PR is what makes it matter.

Minor: `crates/marmot-account/AGENTS.md` still describes `keyring.rs` as "platform keyring store init (per-OS `#[cfg]` cascade) and keyring-error mapping". It now also owns entry construction, the iOS service-name versioning scheme, and the iOS secret write path.

erskingardner (commented):
**Recommend merge** (posted as comment — GitHub blocks approve/request-changes on your own PR).

Re-reviewed `db6e6dae`. Prior blocking feedback is addressed:

- Legacy cleanup is best-effort via `cleanup_legacy_ios_secret` on load/write/migrate, so a stale `WhenUnlocked` delete cannot fail a successful background load or a durable new write.
- `remove_secret` now attempts both current and legacy deletes and combines errors.
- Writes use `Cred::build(..., AccessPolicy::AfterFirstUnlockThisDeviceOnly)` instead of the broken string-modifier path / duplicated Security.framework write.
- PR CI now compiles the iOS cfg path (`iOS account keychain clippy` is green).

No new blocking issues. The versioned-service migrate-verify-then-cleanup shape still looks right for add-only keychain ACLs.

crates/marmot-account/src/secret_store.rs
```diff
@@ -172,7 +176,74 @@ impl KeychainSecretStore {
...
+            &self.service_name,
+            &account.account_id_hex,
+            legacy_secret.as_str(),
+        )?;
+        let replacement = self
+            .entry_for_account(&account.account_id_hex)?
+            .get_password()
+            .map(Zeroizing::new)
+            .map_err(map_keyring_error)?;
+        if replacement.as_str() != legacy_secret.as_str() {
```
erskingardner, line 240: **Non-blocking:** the mismatch branch detects the problem but leaves the bad entry in place, and the next `load_secret` trusts it.

If the read-back ever differs from the legacy value, this returns `Err` while the versioned entry stays written with the wrong value. The very next `load_secret` takes the `Ok` arm at the top (the versioned entry now exists), parses whatever is there, returns those keys, and then `cleanup_legacy_ios_secret` deletes the legacy item — so the recovery path silently destroys the good copy and keeps the bad one. The check makes things worse than not checking.

Two coherent options:

1. Self-heal: on mismatch, `Self::remove_entry(self.entry_for_account(...)?)` before returning the error, so migration retries cleanly next time.
2. Drop the read-back entirely. `set_password` returning `Ok` already means `SecItemAdd`/`SecItemUpdate` succeeded, and the one realistic multi-item case (`Ambiguous`) surfaces as an error from `get_password`, not as a value mismatch — which arguably makes the `!=` branch unreachable, i.e. handling an impossible scenario.

I lean toward (2) plus a one-line comment, but (1) is fine if you want belt-and-braces on a destructive migration.

crates/marmot-account/src/secret_store.rs
```diff
@@ -206,9 +296,14 @@ impl AccountSecretStore for KeychainSecretStore {
         {
             Ok(secret_key) => {
                 let secret_key = Zeroizing::new(secret_key);
-                nostr::Keys::parse(secret_key.as_str())
-                    .map_err(|_| AccountHomeError::InvalidSecretKey)
+                let keys = nostr::Keys::parse(secret_key.as_str())
+                    .map_err(|_| AccountHomeError::InvalidSecretKey)?;
+                self.cleanup_legacy_ios_secret(&account.account_id_hex);
```
erskingardner, line 301: **Non-blocking (perf):** this fires on *every* successful iOS load, forever, including for accounts created after this ships that never had a legacy entry.

Reaching the `Ok` arm means the versioned entry exists, which for a post-migration account means the legacy item is guaranteed absent — yet each load still issues a `SecItemDelete` against the protected keychain. That is a wasted round trip on the notification-service-extension hot path, which is exactly the path this PR is trying to make fast and reliable.

The crash-recovery case it covers (migration wrote + verified but died before cleanup) is already covered on the *next* migration-eligible load... except it is not, because after a successful write the versioned entry exists so `migrate_legacy_ios_secret` never runs again. So the unconditional cleanup is doing real work — it is just paying for it on every load rather than once.

If you want to keep it, that is defensible; consider noting the tradeoff in a comment so a future reader does not delete it as redundant. If you want it cheap, gate it behind a `std::sync::Once` / `AtomicBool` per store instance so it runs at most once per process.

.github/workflows/ci.yml
```diff
@@ -125,6 +125,34 @@ jobs:
...
+          rustup default "$toolchain"
+          rustup target add aarch64-apple-ios
+          rustc --version
+          cargo --version
+
+      - name: Cache Cargo artifacts
+        uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
+
+      - name: Check iOS account keychain code
+        run: cargo clippy -p marmot-account --target aarch64-apple-ios --locked -- -D warnings
```
erskingardner, line 154: Thanks for adding this — it closes the "no PR CI compiles the iOS `cfg` block" gap from the last round, and the job is green on this head.

**Non-blocking, remaining gap:** this proves the iOS block *compiles*, not that it *works*. Every new branch in `secret_store.rs` (legacy discovery fallback, `migrate_legacy_ios_secret`, the combined `remove_secret`) sits behind `#[cfg(target_os = "ios")]`, so `cargo test` cannot reach any of it on any CI host, and there is no iOS test job. The migration is the riskiest code in the PR and its only validation is the manual staging-device run in the description.

Not blocking, but worth capturing as a follow-up: if the legacy-vs-versioned service pair were a runtime field on `KeychainSecretStore` (e.g. an `Option<String>` legacy service name) rather than a `cfg`, the whole migration state machine would be testable against `keyring_core::mock::Store` on Linux, with the `cfg` reduced to choosing the two service names and the write policy.

erskingardner (commented):
Re-review of `db6e6dae`. All five items from my previous review are addressed. No blocking issues remain — good to merge from my side (GitHub will not let me formally approve my own PR).

## Prior findings

1. **Blocking — `load_secret` legacy cleanup propagating.** Fixed. `cleanup_legacy_ios_secret` swallows the delete result, so a successful load stays successful.
2. **Blocking — `write_secret` legacy cleanup propagating.** Fixed the same way; the durable write is no longer undone by a failed tidy-up, so the `home.rs` create/import rollbacks can no longer fire on a stale-item delete failure.
3. **Blocking — no PR CI compiling the iOS `cfg` block.** Fixed. New `ios-account-clippy` job on `macos-latest` in `ci.yml`, which triggers on `pull_request`, and it is green on this head.
4. **Non-blocking — undocumented keyring-store bypass.** Addressed. The comment in `write_keyring_secret` now states that the 1.0.0 string modifier parser drops the device-only qualifier and that the typed API preserves it, which is the surprising part a future reader needs.
5. **Nit — stale `AGENTS.md`.** Fixed.

I also verified the design assumption the whole approach rests on: in `apple-native-keyring-store` 1.0.0, `Cred::get_secret` and `Cred::delete_credential` build their queries from service/account (plus access group and sync flag) only — the access policy is not part of the query. So a store-built `Entry` at the versioned service reads and deletes the `AfterFirstUnlockThisDeviceOnly` item written by the typed `Cred` without issue. Good. `Cred::build(..., None, false)` is also the right call: no access group, no iCloud sync.

## Non-blocking notes on this revision

- The migration read-back verification detects a mismatch but leaves the bad versioned entry in place, and the next `load_secret` trusts it and deletes the legacy copy. Either self-heal or drop the check — inline.
- `cleanup_legacy_ios_secret` runs on every successful iOS load, forever, including for accounts that never had a legacy entry — an extra `SecItemDelete` on the NSE hot path. Inline.
- The iOS branches are `cfg`-gated end to end, so nothing exercises the migration state machine in `cargo test` on any host. Compile-only CI is a real improvement over nothing; automated behavioral coverage would be a good follow-up. Inline.

None of these should hold the fix up — locked-device background access is a user-visible break today and this is a contained, well-scoped fix with a crash-safe migration order (write, verify, then delete).

PR #1302, opened 2026-08-07
Add stateful canonical journey generator

erskingardner:
## What changed

- add deterministic, legality-aware `chat-journey/v1` cases that emit canonical Scenario IR v3
- model membership, admins, connectivity, epochs, group profile state, application delivery, self-updates, restart checkpoints, offline periods, and catch-up
- alternate engine-backed late-membership journeys with retained-relay offline-history journeys
- add terminal oracles for epoch, roster, profile, admin policy, application delivery, exact state, quiescence, and active-member decryptability
- run all eight required-kind rotations against real subjects, covering admin changes, self-updates, restart checkpoints, and removals on both profiles
- persist every generated family input before action zero as an owner-only, versioned artifact containing the full generated case, selected subject adapter, and semantic expectations
- add `--generated-input FILE` so those artifacts reexecute through the same report path, including retained-relay cases
- keep report and promotable vector-candidate artifacts owner-only and outside the generated-input filename namespace
- compare admin-policy expectations order-insensitively while preserving duplicate detection
- document the serialized-commit boundary and keep mutation-after-reopen plus racing commits in `convergence-chaos/v1`

## Why

The existing shrinkable intent generator covers a narrow action vocabulary, while fixed scenarios exercise richer behavior one case at a time. This adds a stateful canonical generator that builds broad, reproducible, protocol-legal user journeys across membership, group data, application traffic, offline recovery, and lifecycle events.

The two journey profiles deliberately encode different pending-work expectations. Late joiners can legitimately retain pre-admission transport objects they cannot decrypt, so membership journeys require founders to settle without falsely demanding global emptiness. Retained-history journeys require all surviving members to reach global quiescence after catch-up.

Saved generated inputs are separate from promotable vector candidates. They preserve the exact generator output and adapter selection across crashes or later generator drift, while vectors remain reviewed stable contracts.

## Impact

This PR changes simulator, scenario-reporting, tests, and documentation only. It does not change production convergence-engine or application behavior.

The write-before-execute and owner-only artifact behavior applies to all generated families, not only `chat-journey/v1`. Generated-input filenames now end in `-generated-input.json`, so `--vectors` directory discovery does not mistake them for vector fixtures.

## Validation

- `cargo test -p cgka-conformance-simulator --test stateful_generator` (eight strict cases plus retained-relay saved-input replay)
- `cargo test -p cgka-conformance-simulator --test report_runner`
- `cargo test -p cgka-conformance-simulator --lib`
- `just fast-ci`
- `git diff --check`


crates/cgka-conformance-simulator/tests/stateful_generator.rs
```diff
@@ -0,0 +1,198 @@
...
+    let output = tempfile::Builder::new()
+        .prefix("stateful-journey-")
+        .tempdir_in(target)
+        .expect("private report output under target");
+    let summary = run_report(&ReportArgs {
+        input: ReportInput::GeneratedFamily {
+            family: STATEFUL_CHAT_JOURNEY_FAMILY.into(),
+            seed: 42,
+            cases: 2,
+        },
```
erskingardner, line 83: Only cases 0 and 1 are ever executed against a real subject. The required-kind rotation in `generate_stateful_chat_journey_case` is `(case_index / 2) % 4`, so cases 0/1 always land on `UpdateAdminPolicy`. That means the guaranteed-`SelfUpdate` (cases 2/3), guaranteed-`Restart` (4/5), and guaranteed-`Remove` (6/7) journeys are never run — only compiled and symbolically replayed by `assert_legal_journey`, which is a mirror of the generator, not an independent check.

The riskiest prediction in the whole model is `JourneyAction::Restart` clearing `received_payloads`, since it asserts that the harness event buffer (and therefore the derived `received_payloads` observation) resets on restart. Nothing here executes a restart-backboned case to confirm that.

Suggest `cases: 8` so every rotation runs on both profiles. If 8 is too slow for the PR lane, at minimum cover 4/5 so restart executes.

erskingardner (commented):
Test-infra only, CI green, and the two-profile split (late-membership vs. founding-member retained history) is a genuinely good insight — pre-admission retry rows really are not a quiescence defect, and encoding that as two families rather than one weakened oracle is the right call.

My main concern is verification depth rather than correctness: the PR adds a 600-line symbolic model that predicts epochs, rosters, admin sets, profile values, and per-client delivery ledgers, but the only test that actually runs a generated case through the engine executes **2 cases**. The structural test compiles 8 and replays legality symbolically, which validates the generator against itself, not against the subject. The required-kind rotation means the `Remove`, `SelfUpdate`, and `Restart` backbones — including the `Restart` ledger-clearing assumption, which is the least obvious prediction in the model — are never executed. Bumping the report-runner test to 8 cases would cover all four rotations on both profiles and make the model's predictions actually falsifiable.

Secondary notes inline: an implicit panic invariant in `choose_kind`, a duplicate-masking change in the `AdminPolicy` oracle, and some artifact/naming questions in `report.rs`. Nothing blocking.

One process note: the PR both implements Milestone 6.5 and checks off its own boxes in `convergence-reliability-plan.md`. That's consistent with how prior milestones landed here, so I'm not objecting — just flagging that a self-marked `[x]` for "cover ... restarts, offline windows, full-history catch-up" is stronger than what the executed test currently demonstrates.

crates/cgka-conformance-simulator/src/stateful_generator.rs
```diff
@@ -0,0 +1,607 @@
...
+        actions
+    }
+
+    fn choose_kind(&self, rng: &mut StdRng, kind: JourneyActionKind) -> JourneyAction {
+        let choices = self
+            .legal_actions()
+            .into_iter()
+            .filter(|action| action.kind() == kind)
+            .collect::<Vec<_>>();
+        choices[rng.gen_range(0..choices.len())].clone()
```
erskingardner, line 227: `choices[rng.gen_range(0..choices.len())]` panics with a bare index-out-of-bounds if no legal action of `kind` exists. It happens to be unreachable today, but only because of a chain of implicit invariants: `Remove` legality excludes `alice` and `bob` by hardcoded name, so the `(case_index / 2) % 4 == 3` rotation always has a victim (`Membership` gets one via the preceding forced `Invite`; `OfflineRetainedHistory` starts with carol and david). None of that is stated anywhere, and it silently breaks if the forced-action order or the `Remove` filter changes.

Cheapest fix is to make the invariant loud:

```rust
let choices = ...;
assert!(!choices.is_empty(), "no legal {kind:?} action in case {}", self.case_index);
```

Relatedly, the `member.as_str() != "alice" && member.as_str() != "bob"` filter is doing double duty — it keeps `Remove` non-empty *and* it keeps `finish`'s hardcoded `pending_free_clients: vec!["alice", "bob"]` accurate for the `Membership` profile. That coupling deserves a one-line comment; right now the two sites are 200 lines apart and neither mentions the other.

crates/cgka-conformance-simulator/src/vector.rs
```diff
@@ -263,13 +272,17 @@ impl TraceExpectation {
                 }
             }
             TraceExpectation::AdminPolicy { client, admins } => {
+                let expected_admins = admins.iter().collect::<BTreeSet<_>>();
```
erskingardner: Comparing as `BTreeSet` makes the check order-insensitive (good, and what the generator needs) but also duplicate-insensitive: an observed policy of `["alice", "alice"]` now satisfies an expectation of `["alice"]`. A duplicated entry in an observed admin policy is a defect worth catching, and this silently stops catching it — across every existing scenario using `AdminPolicy`, not just the new family.

Sorted-vector comparison gets the ordering-insensitivity without dropping multiplicity:

```rust
let mut expected_admins = admins.clone();
expected_admins.sort();
// ... compare against a sorted clone of policy.admins
```

The new `admin_policy_expectation_compares_the_semantic_set` test passes either way; a duplicate-rejection case would pin the distinction.

crates/cgka-conformance-simulator/src/report.rs
```diff
@@ -239,11 +239,20 @@ async fn run_generated_family_reports(
         "convergence-e2e-delivery/v1" => generate_convergence_e2e_delivery_family(seed, cases),
         "convergence-chaos/v1" => generate_convergence_chaos_family(seed, cases),
         "adversarial-reliability/v1" => generate_adversarial_reliability_family(seed, cases),
+        "chat-journey/v1" => generate_stateful_chat_journey_family(seed, cases),
         other => return Err(format!("unsupported family {other}").into()),
     };
 
     let mut summaries = Vec::with_capacity(cases.len());
     for case in cases {
+        let input_output = out.join(format!(
```
erskingardner, line 262: Three things about this write, none blocking:

1. **Suffix collides with vector collection.** `is_vector_fixture_file` / `collect_vector_fixture_directory` match on the `.v1.json` suffix, so `*-input.v1.json` is now picked up alongside `*-fixture.v1.json` if anyone points `--vectors` at a report output directory. Every case then runs twice, once without an expected trace. A suffix outside the vector namespace (e.g. `-input.json`) would avoid that.

2. **`write_private` is inconsistent with its neighbours.** The report and fixture-candidate writes a few lines below use plain `std::fs::write`, and `run_report` already does `create_dir_all_private(&args.out)`, so the parent is 0700 regardless. The 0600 here is real but the README's emphasis on "owner-only permissions" oversells it — the actual value of this change is *write-before-execute*, not the mode. Either make the sibling writes private too, or lean the docs on the crash-preservation rationale alone.

3. **Scope.** This applies to all five generated families, not just `chat-journey/v1`. That's defensible (the crash-preservation argument is family-independent), but it is a behavior change to four existing families in a PR titled "Add stateful canonical journey generator", and it isn't called out in the PR description's Impact section.

crates/cgka-conformance-simulator/README.md
```diff
@@ -409,6 +418,25 @@ The generated `scenario` is a normal `ScenarioSpec`, so a selected generated cas
 without inventing a separate execution path. We promote a case only when it should become a stable named contract, such
 as a regression fixture or the smallest readable example of a semantic edge.
 
+`generate_stateful_chat_journey_family(seed, cases)` adds a legality-aware product journey over the same canonical IR.
+Its symbolic state tracks membership, administrators, connectivity, epoch, group profile, and per-client application
+delivery. Adjacent case indices alternate between late-membership journeys on the engine subject and retained-history
+offline journeys whose offline participants were members from epoch 1. That split is deliberate: a late joiner querying
+full history sees valid pre-admission objects it was never entitled to decrypt, so combining both shapes would make a
+global no-pending assertion a false protocol guarantee. Every case covers application traffic and a full-profile
```
erskingardner: "a full-profile update" overstates it. `profile_sequence` starts at `case_index % 3` and the rotation emits name-only, description-only, or both — so two thirds of first updates are partial, and a name-only case leaves `group_description` at `""`. `SCENARIOS.md` says "a profile update", which is accurate; suggest matching that wording here.

crates/cgka-conformance-simulator/AGENTS.md
```diff
@@ -302,12 +309,13 @@ cargo run -p cgka-conformance-simulator --bin cgka-conformance-simulator-report
...
-- **`HarnessIntent` does not generate Invite / UpgradeCapabilities / UpdateGroupData.** Invite needs client minting
-  inside a strategy; the scripted tests cover it today. Upgrade and group-data update lifecycle are covered by separate
-  generated properties.
+- **`HarnessIntent` does not generate Invite / UpgradeCapabilities / UpdateGroupProfile.** It remains the small
+  shrinkable send/leave strategy. `chat-journey/v1` now generates legal invites and profile changes as canonical IR and
+  relies on saved-input semantic reduction; upgrade lifecycle remains separate.
 - **Partition policy is scripted, not strategy-driven.** The bus supports partitions; proptest currently drives FIFO /
   Reverse / SeededRandom.
-- **Generated family coverage is now convergence-focused but not exhaustive.** `send-leave/v1` records lifecycle
+- **Generated family coverage is broad but not exhaustive.** `chat-journey/v1` covers legal product-shaped membership,
```
erskingardner, line 319: The coverage-boundary section is the right place to record what `chat-journey/v1` deliberately does *not* generate, and one omission is significant enough to name: **the family never produces concurrent or racing commits.** `confirmed_mutation` pushes `DeliverAll` + `Tick` after every single mutation, and offline clients are excluded from `legal_actions`, so every commit is fully settled before the next one is chosen. There is no interleaving, no fork, no merge.

That's a reasonable design choice — `convergence-chaos/v1` owns races, and serializing is what makes the exact epoch/roster/delivery oracles predictable — but "broad but not exhaustive" plus "covers legal product-shaped membership, admin, profile, application, offline, reconnect, self-update, and restart histories" reads as though the family stresses scheduling too. One clause saying commits are serialized by construction and races stay with `convergence-chaos/v1` would keep the boundary honest.

erskingardner (commented):
**Review status: Request changes** (posted as comment — GitHub disallows `REQUEST_CHANGES` on your own PR)

The new pre-run input artifact drops the subject adapter, so half of this family's claimed crash reproducers are not executable through the vector/report path.

## Summary
The generator design is otherwise strong: symbolic legality model → ordinary Scenario IR v3, deliberate Membership vs OfflineRetainedHistory split for pending-work semantics, and terminal oracles that match the model. CI on the branch tip is green.

## Blocking
1. **Saved `*-input.v1.json` omits `GeneratedSubjectKind`.** Persistence goes through `GeneratedScenarioCase::to_vector_fixture`, and `run_vector_fixture_report*` always builds an `EngineHarnessSubject`. Odd `chat-journey/v1` cases emit `sync_relay_history` / retained-relay topology and require `RetainedRelaySubject`; replaying the saved input via `--vectors` fails capability preflight instead of reproducing the run. Docs call this file the executable crash reproducer; the test only checks that the file exists with mode `0600`.

## Non-blocking
2. **Rebase onto `master`.** `#1299` is merged; this PR is still `CONFLICTING` / diverged with the pre-merge stack commits.
3. **Execution coverage is thin for the Remove rotation.** The report test runs only seed 42 / 2 cases; Remove is exercised by the legality corpus, not by an executed strict-oracle case.
4. **Remove leaves the member in `online`.** Later `deliver_to_online` keeps ticking removed clients. Harmless so far, but the connectivity model is stale.

## Premise
Needed and appropriately scoped to simulator/docs. Landing after the subject-preserving input fix (or an equivalent generated-case replay path) looks right.

crates/cgka-conformance-simulator/src/report.rs
```diff
@@ -239,11 +239,20 @@ async fn run_generated_family_reports(
...
+            "{}-seed-{}-case-{}-input.v1.json",
+            case.family_name.replace('/', "-"),
+            case.seed,
+            case.case_index
+        ));
+        let input_fixture = case.to_vector_fixture(env!("CARGO_PKG_VERSION"), None);
+        fs_private::write_private(&input_output, &serde_json::to_vec_pretty(&input_fixture)?)?;
```
erskingardner: **Blocking:** this "canonical input" drops `case.subject`.

`to_vector_fixture()` only serializes a `VectorFixture` (scenario + expectations). `run_vector_fixture_report*` always constructs `EngineHarnessSubject`, which rejects `SyncRelayHistory` (`RetainedRelayHistory` unsupported).

Odd `chat-journey/v1` cases set `GeneratedSubjectKind::RetainedRelay` and emit retained-history steps, so the owner-only file written here is not an executable reproducer for that half of the corpus — despite README/SCENARIOS/Milestone 6.5 calling it exactly that. The new test only asserts the file exists with mode `0600`.

Fix options (pick one and test it):
1. Persist the full `GeneratedScenarioCase` (includes `subject`) and teach the report CLI to run that artifact, or
2. Extend the portable input schema with subject/adapter and select `RetainedRelaySubject` when required.

Regenerating via `--family/--seed` is not a substitute if the point of this file is surviving generator drift or a crash mid-run.

crates/cgka-conformance-simulator/tests/stateful_generator.rs
```diff
@@ -0,0 +1,198 @@
...
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            assert_eq!(
+                std::fs::metadata(input).unwrap().permissions().mode() & 0o777,
+                0o600
+            );
+        }
+    }
+}
```
erskingardner, line 150: This checks presence + permissions, not replayability. After fixing input persistence, please assert that the saved odd-case input (retained-relay) actually re-executes successfully through the intended report path — otherwise this test will keep greenwashing the broken reproducer.

crates/cgka-conformance-simulator/src/stateful_generator.rs
```diff
@@ -0,0 +1,607 @@
...
+            JourneyAction::Remove { member } => {
+                debug_assert!(self.members.contains(&member));
+                debug_assert!(!self.admins.contains(&member));
+                let pending = self.next_publication("remove");
+                self.steps.push(ScenarioStep::RemoveMembers {
+                    remover: "alice".into(),
+                    members: vec![member.clone()],
+                    pending: pending.clone(),
+                });
+                self.members.remove(&member);
```
erskingardner, line 262: Nit: `members.remove` but `online` is left untouched, so removed clients keep getting included in later `deliver_to_online()` ticks. Consider also removing from `online` (and any other connectivity bookkeeping) so the symbolic model matches "no longer a live participant."

crates/cgka-conformance-simulator/src/stateful_generator.rs
```diff
@@ -0,0 +1,635 @@
...
+        choices[rng.gen_range(0..choices.len())].clone()
+    }
+
+    fn choose_non_restart(&self, rng: &mut StdRng) -> JourneyAction {
+        let choices = self
+            .legal_actions()
+            .into_iter()
+            .filter(|action| action.kind() != JourneyActionKind::Restart)
+            .collect::<Vec<_>>();
+        choices[rng.gen_range(0..choices.len())].clone()
```
erskingardner, line 236: Nit, follow-up from the `choose_kind` assert: `choose_non_restart` still indexes `choices[rng.gen_range(0..choices.len())]` with no emptiness guard, so it panics with a bare index-out-of-bounds if the legal set ever empties. It is unreachable today only because online members always yield a `Send`, but that is the same implicit invariant the sibling method now states loudly. Mirroring the `assert!` here (or factoring both through one `choose_from(choices, rng)` helper that asserts once) keeps the two paths honest with each other.

crates/cgka-conformance-simulator/src/stateful_generator.rs
```diff
@@ -0,0 +1,635 @@
...
+                    .iter()
+                    .filter(|member| member.as_str() != "alice")
+                    .cloned()
+                    .map(|target| JourneyAction::UpdateAdminPolicy { target }),
+            );
+        }
+        actions.extend(
+            self.members
+                .intersection(&self.online)
+                .filter(|client| !self.self_updated_clients.contains(*client))
```
erskingardner, line 196: This new `self_updated_clients` filter caps each client at one `SelfUpdate` for the whole journey, which is a real narrowing of the action vocabulary — repeated key rotation by the same member is now unreachable — and it is not mentioned in the `AGENTS.md` coverage-boundary list that this PR otherwise updated with the serialized-commit note.

Since it arrived in the same commit that first executed all eight rotations, I read it as a fix for something that failed once cases 2/3 actually ran. Worth a sentence saying which prediction repeated self-updates broke (epoch drift? pending-work residue?) and whether that is a generator-model gap or a subject observation to chase later. Right now the constraint is silent, so the next person to widen it will rediscover the failure from scratch.

crates/cgka-conformance-simulator/src/stateful_generator.rs
```diff
@@ -0,0 +1,635 @@
...
+                debug_assert!(self.members.contains(&client));
+                debug_assert!(self.online.contains(&client));
+                self.steps.push(ScenarioStep::RestartClient {
+                    client: client.clone(),
+                });
+                self.received_payloads
+                    .get_mut(&client)
+                    .expect("all clients have a delivery ledger")
+                    .clear();
+                self.deliver_to_online();
```
erskingardner, line 372: Same question in the stronger form for `Restart`. The previous revision let `choose_any` pick a restart mid-journey; this revision routes every non-forced pick through `choose_non_restart` and applies the forced restart only after the extra-action loop, so restart is now strictly terminal.

The docs frame that as a design boundary ("mutation-after-reopen belongs to `convergence-chaos/v1`"), and that framing is defensible on its own. But the change landed in the same commit that first executed the restart rotation, which makes it look like mid-journey restart failed the `received_payloads.clear()` prediction and the response was to remove the case rather than explain it. If so, that is the single most interesting thing this PR learned, and it should be recorded — mid-journey restart is exactly the shape a real client hits. A one-line note ("restart mid-journey re-observes retained relay history, tracked in #NNNN") would turn a silent narrowing into a known gap.

Minor, related: with restart terminal-only, the `deliver_to_online()` added at the end of `JourneyAction::Restart` is immediately followed by `finish`'s `accept_all_outbound` + `deliver_to_online()`, so it emits a duplicate `DeliverAll`/`Tick` pair. Harmless, but droppable.

erskingardner (commented):
**Re-review of `487bccf` — recommend merge** (posted as a comment; GitHub disallows `APPROVE` on your own PR.)

The blocking finding is genuinely fixed, not papered over, and every prior note has a matching change.

## Prior findings — all addressed

| Finding | Resolution |
| --- | --- |
| **Blocking:** saved input dropped `GeneratedSubjectKind`, so retained-relay cases were not executable reproducers | `GeneratedScenarioInputV1` now persists the whole `GeneratedScenarioCase` (subject + expectations), `--generated-input FILE` replays it through `run_generated_case_artifacts`, and the test replays the saved **case 1** (`RetainedRelay`) end-to-end with `strict_oracle` and asserts `failed() == 0`. This is the right fix — it is verified replayability, not a file-exists check. |
| Only 2 cases executed; `SelfUpdate`/`Restart`/`Remove` rotations never ran | `cases: 8`. All four rotations now execute on both profiles, including the `Restart` ledger-clearing prediction. |
| `*-input.v1.json` collides with `--vectors` discovery | Renamed to `*-generated-input.json`; `is_vector_fixture_file` matches `.v1.json`, so the namespaces are disjoint. |
| `write_private` inconsistent with sibling writes | Report, fixture candidate, and the vector-fixture report all use `write_private`; the test now asserts `0600` on all three artifacts. |
| Implicit panic in `choose_kind` | `assert!` added with the case index, plus the comment tying the `alice`/`bob` filter to `finish`'s `pending_free_clients`. |
| `AdminPolicy` set comparison masked duplicates | Sorted-vector comparison + `admin_policy_expectation_rejects_duplicate_admins` pins the distinction. |
| `Remove` left the member in `online` | `self.online.remove(&member)`, mirrored in `assert_legal_journey`. |
| Serialized-commit boundary undocumented | New coverage-boundary bullet in `AGENTS.md`; `README.md` "full-profile" → "a profile"; branch is `MERGEABLE` again. |

The `run_generated_case_artifacts` extraction is a good side effect — the family and input paths now share one artifact-writing implementation instead of diverging.

## Non-blocking, inline

1. `choose_non_restart` still has the bare-index panic that `choose_kind` no longer has.
2. `self_updated_clients` silently caps each client at one self-update, and it is not in the coverage-boundary list.
3. Restart became strictly terminal in the same commit that first executed it. The stated rationale is fine, but if a mid-journey restart is what failed, that is the most valuable thing this PR learned and it should be recorded rather than generated away. Also leaves a redundant `DeliverAll`/`Tick` pair.

(2) and (3) are the same shape: the fix for "the model was never falsified" was partly to shrink what the model claims. That is a legitimate move, and the eight executed cases are still a large net gain over two — but the shrink should be documented as a known gap, not as a boundary that was always intended.

## Merge gate

`Rust tests`, `Rust check`, `Rust clippy`, and `Simulator and vectors` were still pending at the time of review. Eight engine-backed journeys is 4x the previous execution cost in that lane; please confirm those go green before merging.

Issue #1161, opened 2026-07-28
Bound local-ready startup cost instead of hydrating and snapshotting every group


Datawav:
## Problem

Android cold start is still perceived as slow even after network catch-up was removed from `Marmot.start()` by PR #1134. The remaining MDK local-readiness boundary still opens and hydrates every stored MLS group, then eagerly snapshots every group's roster and MLS state before the account worker signals ready. A returning user's persisted chat list should not have startup latency proportional to all group state on the account.

Verified against `origin/master` at `523a193815eafb22560923a3960b70e07373ac93`.

## Current critical path

- `crates/marmot-app/src/runtime/account_worker.rs:444-479` does not send the worker-ready signal until `runtime_local_client(...)` has opened the account and `client.group_read_snapshot()` has completed.
- `crates/cgka-engine/src/engine.rs:1019-1032` synchronously enumerates and hydrates every stored group during session open.
- `crates/cgka-engine/src/engine.rs:1068+` loads each group's stored record and OpenMLS state and performs recovery/validation work.
- `crates/marmot-app/src/client/mod.rs:787-833` then loops over every group again to materialize all member records and MLS-state records into `GroupReadSnapshot` before readiness.
- PR #1134 correctly moved relay activation, subscription registration, directory sync, and catch-up after local readiness, but local readiness still has an unbounded all-groups cost.
- The O(groups × members) proof-verification problem in #625 was fixed with validated-tree markers; this issue is the remaining broader hydration/snapshot scaling boundary, not a duplicate.

There is no stage-level startup metric for group enumeration, MLS hydration, profile loading, or read-snapshot construction. `AccountOpen` aggregates them, which prevents attribution and regression budgets.

## Requested outcome

Make chat-list local readiness independent of eager hydration/snapshotting of every group where safety permits.

Candidate design directions:

1. Split persisted app-projection readiness from full account-worker/group-command readiness so hosts can read the durable chat list first.
2. Hydrate group state lazily on first group detail/mutation, or hydrate in a bounded background pipeline after the local chat projection is available.
3. Replace the all-groups `GroupReadSnapshot` barrier with lazy/per-group snapshots or publish readiness before snapshot completion while preserving safe behavior for commands received during initial catch-up.
4. Add bounded, privacy-safe stage telemetry and scaling benchmarks so optimization is driven by measured cost rather than moving work blindly.

[...]

PR #1306, opened 2026-08-07
Add account-open stage telemetry, scaling benchmarks, and pre-snapshot readiness

erskingardner:
Part 1 of 3 for #1161 (measurement layer + first quick win).

- **Stage telemetry**: `AccountDeviceSession::open` captures privacy-safe `SessionOpenTimings` (storage open / engine build / group hydration durations + aggregate counts). The account worker records them as new fixed-bucket `AccountSessionOpen` / `AccountGroupHydration` metrics, and the startup `GroupReadSnapshot` capture is measured as `AccountGroupReadSnapshot` with its shared profile load as `AccountProfileLoad`. Export rows and the telemetry doc catalogue are updated; no identifiers anywhere.
- **Ready before snapshot**: the worker signals local readiness before capturing the startup read snapshot; a capture failure surfaces on the account event stream instead of failing readiness.
- **Scaling benchmarks**: `crates/marmot-app/tests/startup_scaling.rs` builds stores with 0/10/100/1000 groups and 8/64-member rosters, cold-reopens them, and prints stable `MDK_BENCH` lines separating the durable chat-projection read from full account-command readiness, with stage sums. `just bench-startup` runs the full matrix in release mode; an on-demand `workflow_dispatch` job (`startup-bench.yml`) archives the report. A small smoke case runs in the normal suite.

Debug-mode baseline on the pre-change code: cold `chat_list` stays ~150 ms flat while open-time group hydration scales 37 ms → 186 ms → 1300 ms across 3/10/100 stored groups — the attribution #1161 asks for.

Acceptance criteria covered: stage-bucket timings for storage/session open, hydration, profile load, snapshot, and total account open; benchmarks reporting chat-projection readiness separately from group-command readiness.


crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -451,43 +451,70 @@ async fn run_app_runtime_account_worker(
...
+            true,
+        );
+        telemetry.record(
+            AppPerformanceOperation::AccountGroupHydration,
+            open_timings.group_hydration,
+            true,
+        );
+    }
     if let Some(ready) = ready.take() {
         let _ = ready.send(Ok(()));
```
erskingardner: [P1] Do not acknowledge command readiness before commands can run. After this send, the worker immediately performs the synchronous O(groups × members) snapshot and does not poll `commands` until it finishes. On the current-thread Tokio runtime used by this crate’s tests and the new benchmark, sending the oneshot does not yield, so `reconcile`/`start` cannot even observe readiness until after the snapshot. On a multithread runtime, `start` may return, but worker-routed reads merely queue; if `profiles_by_id` fails, this task exits after reporting readiness and leaves `start()` successful with a dead worker. That contradicts `MarmotAppRuntime::start`’s guarantee that worker-routed local reads are available. Please either move this signal to a point where the command loop can serve requests, or introduce a distinct projection-ready milestone whose contract and failure handling do not claim worker readiness.

crates/marmot-app/tests/startup_scaling.rs
```diff
@@ -0,0 +1,207 @@
...
+    // --- cold reopen: durable chat projection first, then full readiness ---
+    let app = open_store(&dir_bench, &url);
+    let chat_list_started = Instant::now();
+    let rows = app.chat_list(BENCH_ACCOUNT, false).unwrap();
+    let chat_list = chat_list_started.elapsed();
+    assert_eq!(rows.len(), groups, "cold chat list must hold every group");
+
+    let runtime = MarmotAppRuntime::new(app);
+    let start_started = Instant::now();
+    runtime.start().await.unwrap();
```
erskingardner, line 140: [P1] Measure actual group-command readiness. `start()` is deliberately completed by this PR before `GroupReadSnapshot` exists and before the worker polls commands, so timing it cannot represent the “full account-command readiness” claimed by this benchmark. With this test’s current-thread runtime, the synchronous snapshot happens to run before the ready receiver is scheduled, so `ready_ms` includes it; on a multithread runtime it may not, and the telemetry snapshot read immediately below may still contain zero snapshot samples. The reported boundary therefore depends on executor scheduling and never probes a completed group command. Time an actual worker-routed read (or add an explicit command-ready milestone), and wait for the post-ready stage telemetry before sampling it.

erskingardner (commented):
Blocking findings: the new pre-snapshot signal does not establish the documented worker-command readiness contract, and the benchmark consequently does not measure actual group-command readiness. CI is still in progress at this head.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -451,43 +451,70 @@ async fn run_app_runtime_account_worker(
...
+                publish_app_runtime_account_error(
+                    &events,
+                    &account_id_hex,
+                    &account_label,
+                    message,
+                );
+                return;
+            }
+        }
+    };
```
erskingardner, line 514: **Blocking: readiness contract regression.**

Ready is already acknowledged above, then a snapshot failure only publishes an account error and `return`s. `AccountManager::reconcile` / `MarmotAppRuntime::start()` therefore succeed, leave a finished worker in the map, and only discover the death on a later reconcile/command failure.

Previously this path did `ready.send(Err(err))`, so startup failed closed. If readiness moves before snapshot, failure still needs to fail the open (or otherwise prevent `start()` success) — an event-stream error after a successful ready wait is not equivalent.

Please restore fail-closed startup for snapshot capture, or add a regression test that proves hosts observe the failure through the ready/reconcile path.

crates/cgka-session/src/lib.rs
```diff
@@ -319,17 +347,43 @@ impl AccountDeviceSession {
...
         engine
             .set_convergence_policy(config.convergence_policy)
             .map_err(|e| EngineError::Other(format!("convergence policy: {e}")))?;
         engine.audit_recorder_health();
+        let open_timings = SessionOpenTimings {
+            total: opened_at.elapsed(),
+            storage_open,
+            engine_build,
+            group_hydration,
+            groups_live: engine.live_group_ids()?.len() as u64,
```
erskingardner: **Blocking on the path this PR is trying to attribute/cheapen.**

`engine.live_group_ids()?` re-lists groups from storage after a successful hydrate, and the `?` makes that listing a new open-failure mode. These counts are only used for a debug log — they are not recorded into `AppPerformanceTelemetry` — so this is pure critical-path cost.

Drop the storage recount (or derive an in-memory count), and do not let telemetry bookkeeping fail `open` after hydration succeeded.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -451,43 +451,70 @@ async fn run_app_runtime_account_worker(
     );
     let mut scheduled_push_retry = ScheduledPushRegistrationRetry::new();
 
-    // The session is hydrated. Capture a read snapshot and signal
-    // command-readiness *now*, before the initial relay catch-up. "Ready" means
+    // The session is hydrated. Signal command-readiness *now*, before the
+    // read-snapshot capture and the initial relay catch-up. "Ready" means
     // "hydrated + serving commands", not "caught up": the conversation's
```
erskingardner: Nit: this comment still says ready means "hydrated + serving commands", but the worker has not entered the command `select!` yet — Members/GroupMlsState/etc. only start being answered after snapshot capture completes below. During the gap, `start()` can return while those commands sit in the channel.

Worth clarifying so the new `AccountOpen` boundary is not misread as "group reads are live".

crates/marmot-app/src/client/mod.rs
```diff
@@ -835,12 +835,35 @@ impl AppClient {
...
+        self.group_read_snapshot_inner(Some(telemetry))
+    }
+
+    fn group_read_snapshot_inner(
+        &self,
+        telemetry: Option<&crate::app_telemetry::AppPerformanceTelemetry>,
+    ) -> Result<GroupReadSnapshot, AppError> {
         // Load account profiles once and reuse across every group: the rest of
         // the capture is in-memory engine reads, so the snapshot adds a single
         // storage read to the worker readiness path regardless of group count.
```
erskingardner, line 858: Stale: snapshot capture is no longer on the worker readiness path after this PR. Same for the `members_with_profiles` doc above that still links `AppClient::group_read_snapshot` (renamed away).

erskingardner (commented):
**Request changes** (posted as comment — GitHub blocks `REQUEST_CHANGES` on own PRs).

Measurement layer + benchmarks look solid and match the Part-1 framing for #1161. The pre-snapshot ready move is a reasonable splash quick win, but the failure path currently acknowledges local readiness and then can tear the worker down without failing `reconcile`/`start()`. Also avoid paying another `list_groups` storage round-trip on the open path just for debug counts.

## Blocking
1. Snapshot failure after `ready.send(Ok(()))` leaves `start()` successful while the worker exits
2. `live_group_ids()?` after hydration adds a storage list + new open-failure mode for counts that are not even exported

## Non-blocking
- Comment still says ready means "serving commands", but the command `select!` has not started yet
- Stale docs still describe snapshot as part of the readiness path / `group_read_snapshot`

CI was still running at review time; rust format/check had passed.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in 7b3c9218 by reverting the reorder: the snapshot is captured before `ready.send` again, so readiness is only acknowledged once the worker can immediately enter the command loop, and `AccountOpen` again covers the full local-readiness cost. The pre-snapshot readiness signal now lands with part 3 (#1308), where the worker enters a command-serving hydration loop immediately after acknowledging — so the "serving commands" contract holds there by construction.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in 7b3c9218: restored fail-closed startup — a snapshot capture failure sends `Err` through the ready path again (after publishing the account error), so `reconcile`/`start()` observe it and no dead worker survives a successful start.

crates/cgka-session/src/lib.rs
erskingardner: Fixed in 7b3c9218: dropped the `live_group_ids()?` recount. `SessionOpenTimings` keeps only the in-memory quarantined count, and telemetry bookkeeping can no longer fail an open after hydration succeeded.

crates/marmot-app/tests/startup_scaling.rs
erskingardner: Fixed in 7b3c9218: the benchmark now times a completed worker-routed `group_members` read as its own `group_read_ms` column (zero-group case reports 0), and samples the stage telemetry only after that read completes — so every startup stage sample is guaranteed present and the boundary no longer depends on executor scheduling. With the snapshot back inside the ready-wait, `ready_ms` also honestly includes it again.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in 7b3c9218 — the comment now describes the restored capture-then-ready ordering and its fail-closed contract.

crates/marmot-app/src/client/mod.rs
erskingardner: Fixed in 7b3c9218: the `members_with_profiles` doc links `group_read_snapshot_with_stage_telemetry`, and the snapshot doc describes the fail-closed readiness path again.

erskingardner:
Addressed all review findings in 7b3c9218; stacked branches #1307/#1308 rebased on top.

**Both blocking items** resolved by reverting the ready-before-snapshot reorder in this PR: the snapshot capture is back inside the ready-wait and fail-closed (`ready.send(Err)` on capture failure), so `start()` can never succeed over a dead worker, and `AccountOpen` again measures the full local-readiness cost. `SessionOpenTimings` no longer re-lists storage for counts — only the in-memory quarantined count remains, and telemetry bookkeeping cannot fail an open after hydration succeeded.

**Benchmark**: group-command readiness is now probed by a completed worker-routed `group_members` read (new `group_read_ms` column), with stage telemetry sampled only after it — scheduling-independent and guaranteed-complete samples.

**The splash quick win moved to part 3** (#1308), where it's sound: the worker there enters a command-serving hydration loop immediately after acknowledging readiness, so "ready = serving commands" holds by construction. I also applied this review's dead-worker finding to #1308's snapshot capture: since readiness is acknowledged before capture there, a capture failure now degrades (reads defer to post-catch-up live state) instead of killing the worker.

Docs/comments describing the abandoned ordering are refreshed. `just fast-ci` and the full `marmot-app` suite pass on all three branches.

docs/marmot-architecture/telemetry.md
```diff
@@ -192,7 +192,11 @@ Collected operations:
...
+| `account_session_open` | `AccountDeviceSession::open` inside the worker's account open: SQLCipher storage open, engine construction (including key-package retirement), and stored-group hydration. | Stage attribution for `account_open`; recorded from `SessionOpenTimings` after a successful open, so it has no failure samples. |
+| `account_group_hydration` | The stored-group enumeration/hydration pass inside `AccountDeviceSession::open`. | Subset of `account_session_open`; the stage that scales with stored group state (mdk#1161). |
+| `account_profile_load` | The one shared account-profile load at the start of the startup group-read-snapshot capture. | Single storage read reused across every group in the snapshot. |
+| `account_group_read_snapshot` | The full startup `GroupReadSnapshot` capture after the ready signal. | Scales with groups × members; snapshot answers worker read commands during initial catch-up. |
```
erskingardner, line 199: Stale after the revert: this row still says the capture runs "after the ready signal", but the worker records AccountGroupReadSnapshot and only then ready.send(Ok(())). Suggest "before/as part of the ready signal" so the catalogue matches the fail-closed ordering.

erskingardner (commented):
**Recommend merge** at `7b3c921` (posted as comment — GitHub blocks approve/request-changes on own PRs).

Previous blocking findings are addressed:
- Snapshot capture is back inside the ready-wait and fail-closed (`ready.send(Err)`).
- No post-hydrate `live_group_ids()?` storage recount.
- Bench probes a completed worker-routed `group_members` read (`group_read_ms`).

This PR is now a clean measurement/benchmark slice for #1161 Part 1; the splash reorder belongs in #1308. CI is green.

## Non-blocking
1. `docs/marmot-architecture/telemetry.md` still says `account_group_read_snapshot` runs **after** the ready signal — code captures it **before** ready.
2. PR title/body still advertise pre-snapshot readiness / ready-before-snapshot; please refresh so the stack description matches the code.

docs/marmot-architecture/telemetry.md
erskingardner, line 199: [P3] Update this row to match the restored ordering. At the current head, `account_worker.rs` captures `GroupReadSnapshot` at lines 488–514 and sends the ready signal only at lines 515–517, so “after the ready signal” is backwards; this capture is part of `account_open` and occurs before readiness. The PR title/body contain the same stale description of the reverted behavior.

erskingardner (commented):
Recommend merge. The 7b3c9218 corrective commit resolves the prior readiness/failure and benchmark findings: snapshot capture is fail-closed before the ready signal, no storage recount was added for telemetry, and the benchmark now completes a worker-routed read before sampling stage telemetry. All current required CI checks pass. Non-blocking cleanup: update the PR title/body, which still advertise the reverted pre-snapshot-readiness behavior, and fix the one stale telemetry-table row noted inline. GitHub does not allow approving one’s own PR, so this recommendation is posted as a comment review.

Issue #740, opened 2026-07-05
marmot-app: subscribe_notifications runs blocking SQLCipher I/O on the runtime thread and silently drops notifications on broadcast lag


erskingardner:
**Severity:** MEDIUM
**Component:** `crates/marmot-app`

`subscribe_notifications` is the only subscription in `runtime/subscriptions.rs` that (a) runs blocking file-backed SQLCipher work directly on a tokio worker and (b) does not recover on broadcast lag — two related defects in the same task that together produce user-visible missed notifications.

## Location

`crates/marmot-app/src/runtime/subscriptions.rs:1293-1320` (task body; blocking call at `:1301`, lag handling at `:1318`).

## Defect 1 — blocking DB I/O on an async worker

```rust
Ok(event) => {
    match notifications::notification_update_from_event(&app, &event) {
        Ok(Some(update)) => { ... }
```

`notification_update_from_event` → `NotificationResolver::{settings,group,user}` → `app.notification_settings`, `app.group`, `app.directory_entry_for_account_id`, `app.reaction_target`, each of which opens/queries the file-backed SQLCipher shared storage and per-account directory caches (synchronous `rusqlite`). Every **other** subscription wraps such reads in `blocking_app_task` (= `tokio::task::spawn_blocking`, `runtime/mod.rs:3168`): `subscribe_chats` (`:1017`), `subscribe_chat_list` (`:1089`, `:1165`), `subscribe_group_state` (`:1262`). Only `subscribe_notifications` runs the blocking work inline on a runtime worker, stalling the executor under message load.

## Defect 2 — permanent notification loss on lag

```rust
Err(broadcast::error::RecvError::Lagged(_)) => continue,
```

On lag the handler just skips ahead. The three sibling subscriptions all explicitly recover: `subscribe_messages` re-reads and re-emits missed messages (`:797-841`), `subscribe_timeline_messages` sends a `Refresh` (`:925-936`), `subscribe_chat_list` re-reads the chat list (`:1086-1108`). The file's stated intent is "a slow consumer cannot silently drop messages" (`:795`). A slow notification consumer (made slower by Defect 1) that overflows the broadcast ring loses a notification for every overflowed message — even though the message itself is delivered via other subscriptions. For a messaging app this is a real "I never got notified" defect.

## Suggested fix

[...]

Issue #625, opened 2026-06-10
Session open re-verifies every member's account-identity proof in every group via a serde_json ratchet-tree round trip — O(groups x members) schnorr verifications per open


erskingardner:
**Severity:** low  •  **Category:** performance  •  **Confidence:** high

**Location:** `crates/cgka-engine/src/engine.rs:243`

## What
Every AccountDeviceSession::open runs validate_member_credentials_and_account_proofs for every stored group. That function exports the full ratchet tree, serializes it to a serde_json::Value, deserializes it back, and performs a BIP-340 schnorr verification (plus proof decode) for every leaf. All of this state was already validated at join/invite/commit ingress and is read from the device's own encrypted storage, so open-time cost grows linearly with total membership: ~50 groups of ~50 members ≈ 2500 schnorr verifications plus 50 full-tree JSON round trips per open. marmot-app constructs a session via open_account on each client() call, so this recurs during normal usage.

## Evidence
cgka-session/src/lib.rs:197 open() → engine.rs:230-251 hydrate loop calling validate_member_credentials_and_account_proofs per group (243-246) → group_lifecycle.rs:590-609: export_ratchet_tree → serde_json::to_value → serde_json::from_value → per-leaf validate_leaf_account_identity_proof → account_identity_proof.rs:108+ builds k256::schnorr::VerifyingKey::from_bytes and verifies per leaf. Recurrence: marmot-app/src/lib.rs:1467/2563 opens a fresh session per client() call.

## Suggested fix
Validate lazily (on first use of each group) or cache a per-group validation marker keyed by tree hash/epoch so unchanged groups skip re-verification; avoid the serde_json::Value round trip by walking the exported tree type directly.

---
<sub>Found by the Fable multi-agent code review and confirmed by adversarial verification (independent refuter + reproducer, with tiebreak on disagreement). Line numbers are approximate; see quoted code for the exact site.</sub>

Issue #896, opened 2026-07-17
cgka-engine: `transport_group_id_index` grows unbounded — never pruned on group leave/removal/wipe, and routing rotation is additive (inbound-routing hot-path memory growth)


erskingardner:
**Severity:** LOW
**Component:** `crates/cgka-engine`

## Summary

`Engine::transport_group_id_index` (`HashMap<Vec<u8>, GroupId>`) is insert-only. There is no `.remove(...)`/`.retain(...)`/`.clear(...)` anywhere in the crate, so it accumulates entries with no ceiling on a long-running device. It sits on the per-inbound-event routing hot path.

## Location

- `crates/cgka-engine/src/engine.rs:206` — field definition.
- Inserts: `engine.rs:718`, `engine.rs:756`, `group_lifecycle.rs:302`, `group_lifecycle.rs:632`.
- Read on the inbound routing path: `message_processor/ingest.rs:1746` (`group_id_for_transport_group_id`).
- `engine.rs:686` — `reindex_transport_group_id` is documented as *additive* on routing rotation: prior `nostr_group_id → group_id` entries are retained for the overlap window and never expire.

`grep -rn "transport_group_id_index" crates/cgka-engine/src | grep -E '\.remove|\.retain|\.clear'` returns nothing.

## Two independent growth vectors

1. **Routing rotation:** each `reindex_transport_group_id` adds a new `transport_group_id → group_id` mapping and keeps the prior one (overlap window), which never expires. A single long-lived group that rotates its Nostr routing component repeatedly accretes stale entries forever.
2. **Group churn:** no engine path removes a group's entries on leave / self-eviction / removal / wipe. `realize_self_eviction` and the removal seams set `removed = true` but keep the index entries; `do_join_welcome` only re-inserts.

## Impact

Steady, unbounded memory growth on always-on clients that churn groups or rotate routing. Stale mappings also outlive a group whose local record was later deleted at the storage layer (`delete_local_group_data`, #756); an inbound event on the stale routing id then resolves to a now-deleted group id and does an extra `get_group` miss per event on the hot inbound-routing path (wasted work, no crash).

## Suggested fix

Prune the account's `transport_group_id_index` entries on group leave/removal/wipe (mirror the group-lifecycle removal seams), and drop superseded routing mappings once the rotation overlap window closes.

## Dedup

[...]

Issue #408, opened 2026-06-30
cgka-engine: unknown `transport_group_id` triggers an O(groups) full MlsGroup-load scan per inbound message (attacker-paced amplification)


erskingardner:
### Summary
Every inbound message whose `transport_group_id` does not match a stored `GroupId` directly forces a linear scan that deserializes and loads **every** joined MLS group from storage — before any payload authentication.

### Location
`crates/cgka-engine/src/message_processor/ingest.rs:1354-1388` (`group_id_for_transport_group_id`)

### Details
```rust
let provider = EngineOpenMlsProvider::<S>::new(&self.crypto, self.storage.mls_storage());
for group_id in self.storage.list_groups()? {
    if group_id.as_slice() == transport_group_id { return Ok(group_id); }
    let mls_gid = openmls::group::GroupId::from_slice(group_id.as_slice());
    let Some(mls_group) = MlsGroup::load(... &mls_gid)? else { continue };
    if crate::app_components::transport_group_id_of_group(&mls_group)? == transport_group_id {
        return Ok(group_id);
    }
}
Ok(direct) // falls through to a non-existent group on miss
```
This runs once per `ingest_group_message` call that misses the direct lookup.

### Impact
A transport peer / relay can flood `kind:445` messages bearing arbitrary unknown `transport_group_id`s. Each one forces a full `list_groups()` scan plus a `MlsGroup::load` + decode for every joined group — attacker-paced CPU and storage I/O that scales with how many groups the client is in, performed before the payload is authenticated. Remote DoS / amplification.

### Suggested fix
Maintain a persisted `transport_group_id -> GroupId` index, populated on create/join and on routing-component changes, so routing is O(1). Fall back to the scan only on index miss, and rate-limit / cap it. Related tracking issue: marmot-protocol/mdk#380.

---
*Filed from a meticulous automated code-review pass (6 parallel deep reviews across all 18 crates, findings independently verified and deduplicated against existing open/closed issues).*

---
_Generated by [Claude Code](https://claude.ai/code)_

erskingardner:
Parent tracking issue: marmot-protocol/mdk#380.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1307, opened 2026-08-07
Add engine two-phase hydration with fail-closed unhydrated state

erskingardner:
Part 2 of 3 for #1161 (engine layer), stacked on #1306.

- **Storage surface**: `GroupStorage` gains `list_group_records` and a durable transport-route index (opaque bytes → MLS group id, many-to-one preserving the #740 rotation-overlap semantics) with default impls so non-sqlite providers stay correct. Sqlite stores routes in a new `cgka_transport_group_routes` table (migration 0043), cascade-deleted with the group and deliberately outside group-snapshot scope so a rollback cannot delete a still-valid route.
- **Per-group hydration cheap wins**: `hydrate_one_stored_group` now runs interrupted probe/apply recovery first and reads the group record once (previously up to three times), and fetches the group's message rows once, shared by the leave-request probe, the convergence-input gate, the deferred-peel check, and the self-remove schedule restore (previously four separate full-table scans). The #625 marker deliberately keeps hashing the exported tree bytes — the stored OpenMLS `tree_hash` is not recomputed from the stored leaves, so it cannot detect stored-tree tampering.
- **Two-phase core**: `hydrate_stable_groups_from_storage` becomes a cheap seed pass (durable records only — no MLS load, snapshot list, or message scan) that seeds a provisional epoch entry so `live_group_ids` and the app projection keep listing every group, restores disband/unrecoverable terminal state, and seeds inbound routing from the durable route table. Seeded groups fail closed through the existing `ensure_group_live` chokepoint with a new retryable `EngineError::GroupNotHydrated`; send, ingest, and convergence entry points promote them via `ensure_hydrated`, whose failure quarantines with exact open-time parity (including retracting the provisional epoch entry — a quarantined group never holds one). Groups without a durable route row backfill via at most one MLS load ever, keeping #740's attacker-paced-scan bound in amortized form. The former all-groups loop survives as `hydrate_all_stored_groups`, which `AccountDeviceSession::open` keeps calling until part 3 lands — so this PR is behavior-preserving for every consumer.

New `tests/two_phase_hydration.rs` covers seeding vs gating, `ensure_hydrated` promotion and quarantine parity, cheap-pass idempotency, durable-route round-trips across reopen, and the one-shot ingest backfill. Existing hydration/crash-recovery/pending-commit suites adapt to the eager compatibility path and pass unchanged in behavior.

## Summary by CodeRabbit

* **New Features**
  * Added faster session startup by loading group records first and completing group hydration on demand.
  * Added a retryable “group not hydrated” state for groups still loading.
  * Improved transport-route persistence, recovery, backfill, and cleanup across restarts.
  * Preserved compatibility with full hydration when needed.

* **Bug Fixes**
  * Improved recovery of interrupted, quarantined, and unrecoverable groups.
  * Prevented unavailable group state from exposing outbound messages or processing operations prematurely.

* **Tests**
  * Expanded coverage for hydration, restart recovery, route repair, quarantine, and convergence behavior.

### DIFF
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -374,7 +374,7 @@ impl<S: StorageProvider> Engine<S> {
         monotonic_ms: u64,
     ) -> Result<(), OpenMlsProjectionError> {
         // See `buffer_openmls_convergence_message` (mdk#1161).
-        let _ = self.ensure_hydrated(group_id);
+        self.ensure_hydrated_for_convergence(group_id)?;
         self.buffer_openmls_convergence_message_with_time(
             group_id,
             message,
@@ -394,8 +394,27 @@ impl<S: StorageProvider> Engine<S> {
         // Promote a seeded-but-unhydrated group before persisting convergence
         // input against it (mdk#1161): the group's interrupted probe/apply
         // recovery must run before any of its stored state is extended.
-        let _ = self.ensure_hydrated(group_id);
+        self.ensure_hydrated_for_convergence(group_id)?;
         self.buffer_openmls_convergence_message_with_time(group_id, message, self.convergence_now())
+    }
+
+    /// Fail-closed hydration promotion for the public convergence-buffering
+    /// entry points (mdk#1161): a failed promotion quarantines the group and
+    /// retracts its provisional epoch state, so continuing would take the
+    /// no-epoch-state admission branch and durably retain a convergence row
+    /// for a group validation just rejected. Propagate instead — the caller
+    /// observes the same missing-group view every quarantined group presents,
+    /// and no durable state is extended.
+    fn ensure_hydrated_for_convergence(
+        &mut self,
+        group_id: &GroupId,
+    ) -> Result<(), OpenMlsProjectionError> {
+        self.ensure_hydrated(group_id).map_err(|e| match e {
+            cgka_traits::error::EngineError::UnknownGroup(_) => {
+                OpenMlsProjectionError::MissingGroup
+            }
+            other => OpenMlsProjectionError::Storage(format!("{other}")),
+        })
     }
 
     pub(crate) fn buffer_openmls_convergence_message_with_time(
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -199,6 +199,10 @@ impl<S: StorageProvider> Engine<S> {
         &mut self,
         group_id: &GroupId,
     ) -> Result<Option<u64>, OpenMlsProjectionError> {
+        // Convergence scheduling for a seeded-but-unhydrated group promotes
+        // it first (mdk#1161); a hydration failure quarantines it and the
+        // gate below reports no work.
+        let _ = self.ensure_hydrated(group_id);
         if self.ensure_group_live(group_id).is_err() {
             return Ok(None);
         }
@@ -967,6 +971,10 @@ impl<S: StorageProvider> Engine<S> {
         now: ConvergenceTime,
     ) -> Result<CanonicalizationResult, OpenMlsProjectionError> {
         let now_ms = now.monotonic_ms;
+        // A convergence pass over a seeded-but-unhydrated group promotes it
+        // first (mdk#1161); a hydration failure quarantines it and the gate
+        // below reports the blocked run.
+        let _ = self.ensure_hydrated(group_id);
         // A hydration-quarantined group is frozen until explicit repair: no
         // canonicalization pass may read or mutate its state, and no
         // set_stable may re-activate it out of band (mdk#364). Check this
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -248,6 +248,11 @@ impl<S: StorageProvider> Engine<S> {
         group_id: &GroupId,
         policy: CanonicalizationPolicy,
     ) -> Result<(), OpenMlsProjectionError> {
+        // Promote a seeded-but-unhydrated group first (mdk#1161) so the audit
+        // context snapshot below reads validated state; a hydration failure
+        // quarantines and the write proceeds as it does for quarantined
+        // groups today (the policy is durable config, not group state).
+        let _ = self.ensure_hydrated(group_id);
         self.accept_convergence_policy(&policy)?;
         self.storage
             .put_convergence_policy(group_id, &encode_convergence_policy(&policy)?)
@@ -368,6 +373,8 @@ impl<S: StorageProvider> Engine<S> {
         message: TransportMessage,
         monotonic_ms: u64,
     ) -> Result<(), OpenMlsProjectionError> {
+        // See `buffer_openmls_convergence_message` (mdk#1161).
+        let _ = self.ensure_hydrated(group_id);
         self.buffer_openmls_convergence_message_with_time(
             group_id,
             message,
@@ -384,6 +391,10 @@ impl<S: StorageProvider> Engine<S> {
         group_id: &GroupId,
         message: TransportMessage,
     ) -> Result<(), OpenMlsProjectionError> {
+        // Promote a seeded-but-unhydrated group before persisting convergence
+        // input against it (mdk#1161): the group's interrupted probe/apply
+        // recovery must run before any of its stored state is extended.
+        let _ = self.ensure_hydrated(group_id);
         self.buffer_openmls_convergence_message_with_time(group_id, message, self.convergence_now())
     }
 
--- a/crates/cgka-engine/src/engine.rs
+++ b/crates/cgka-engine/src/engine.rs
@@ -285,6 +285,25 @@ pub struct Engine<S: StorageProvider> {
     /// successful [`Self::retry_hydrate_quarantined_group`].
     pub(crate) quarantined_groups: HashMap<GroupId, GroupHydrationQuarantineReason>,
 
+    /// Groups seeded by the session-open cheap pass whose full per-group
+    /// hydration (MLS load, validation, pending-commit recovery) has not run
+    /// yet (mdk#1161). The seed grants each group a provisional
+    /// `Stable(record.epoch)` epoch entry — so `live_group_ids` and the app
+    /// projection keep listing it — while [`Self::ensure_group_live`] fails
+    /// closed with [`EngineError::GroupNotHydrated`] on every gated surface.
+    /// Membership leaves this set only through [`Self::ensure_hydrated`]:
+    /// success promotes the group to live, failure moves it to
+    /// [`Self::quarantined_groups`] exactly like an open-time quarantine.
+    pub(crate) unhydrated_groups: HashSet<GroupId>,
+
+    /// Seeded, non-disbanded groups with no durable transport-route row yet
+    /// (records predating migration 0043, or a provider without a route
+    /// store). While non-empty, a routing-index miss cannot be trusted:
+    /// ingest backfills these groups' routes on demand — each group leaves
+    /// this set permanently after one MLS load, preserving mdk#740's
+    /// attacker-paced-scan bound in amortized form.
+    pub(crate) route_backfill_pending: HashSet<GroupId>,
+
     /// Authoritative `transport_group_id -> GroupId` resolver for inbound
     /// routing (#740). A Nostr-routed group's `transport_group_id`
     /// (`nostr_group_id`) differs from its MLS group id, so the direct
@@ -578,6 +597,8 @@ impl<S: StorageProvider> EngineBuilder<S> {
             audit_operation_counter: 0,
             current_audit_context: None,
             quarantined_groups: HashMap::new(),
+            unhydrated_groups: HashSet::new(),
+            route_backfill_pending: HashSet::new(),
             transport_group_id_index: HashMap::new(),
             seen_message_ids_hex_cache: None,
             deferred_peel: HashMap::new(),
@@ -739,9 +760,13 @@ impl<S: StorageProvider> Engine<S> {
             .list_outbound_fanouts()?
             .into_iter()
             .filter(|fanout| {
-                fanout
-                    .group_id()
-                    .is_none_or(|group_id| !self.quarantined_groups.contains_key(group_id))
+                fanout.group_id().is_none_or(|group_id| {
+                    // Unhydrated groups' fanouts stay frozen until full
+                    // hydration restores their pending lifecycle (mdk#1161);
+                    // quarantined groups' fanouts are hidden as before.
+                    !self.quarantined_groups.contains_key(group_id)
+                        && !self.unhydrated_groups.contains(group_id)
+                })
             })
             .collect())
     }
@@ -1176,6 +1201,23 @@ impl<S: StorageProvider> Engine<S> {
     /// `MlsGroup::load` is cost-appropriate. Best-effort: a load / routing-read
     /// failure just forfeits the fast path for this group (inbound would fall to
     /// the unknown-group disposition), never fails the merge.
+    /// Insert one transport route into the in-memory resolver AND the durable
+    /// route table (mdk#1161). The durable write is best-effort with the same
+    /// policy as every route refresh: a failure only forfeits the next
+    /// cold-start route seed, never routing correctness — the in-memory
+    /// insert stays authoritative for this session.
+    pub(crate) fn index_transport_group_route(
+        &mut self,
+        transport_group_id: Vec<u8>,
+        group_id: &GroupId,
+    ) {
+        self.transport_group_id_index
+            .insert(transport_group_id.clone(), group_id.clone());
+        let _ = self
+            .storage
+            .put_transport_group_route(&transport_group_id, group_id);
+    }
+
     pub(crate) fn reindex_transport_group_id(&mut self, group_id: &GroupId) {
         let mls_group = {
             let provider = crate::provider::EngineOpenMlsProvider::<S>::new(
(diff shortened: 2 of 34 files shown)


crates/cgka-engine/src/engine.rs
```diff
@@ -1196,24 +1238,180 @@ impl<S: StorageProvider> Engine<S> {
...
+                        "hydrate_unrecoverable_group",
+                        None,
+                        None,
+                    ),
+                );
+                self.events_buf.push_back(GroupEvent::GroupUnrecoverable {
+                    group_id: group.id.clone(),
+                });
+                seedable_group_ids.insert(group.id);
+                continue;
```
erskingardner, line 1332: **Blocking:** this early-continue restores the Unrecoverable epoch marker and intentionally keeps the group out of `unhydrated_groups`, but `hydrate_all_stored_groups` only drains `unhydrated_group_ids()`. So even the eager compatibility path never runs `hydrate_one_stored_group` for unrecoverable groups anymore.

Pre-#1161 open always ran full per-group hydration for these records (MLS load/validation, durable route index, leave-request restore, interrupted probe/apply recovery, deferred-peel / queued-intent scheduling). The PR description says this PR is behavior-preserving for every consumer via `hydrate_all_stored_groups`; this path violates that.

`hydrate_all_stored_groups` needs to finish unrecoverable groups through the same `hydrate_one_stored_group` work (or an equivalent terminal finish-up), while the cheap-only path can still skip background queueing. Please add a restart test that pins at least route indexing + interrupted-probe recovery for an unrecoverable group under `hydrate_all_stored_groups`.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -2366,6 +2372,61 @@ impl<S: StorageProvider> Engine<S> {
...
+                match openmls::group::MlsGroup::load(
+                    <crate::provider::EngineOpenMlsProvider<'_, S> as openmls_traits::OpenMlsProvider>::storage(
+                        &provider,
+                    ),
+                    &mls_gid,
+                ) {
+                    Ok(Some(group)) => group,
+                    // Unreadable MLS state resolves nothing here; the group's
+                    // full hydration (or quarantine) owns that failure.
+                    _ => continue,
```
erskingardner, line 2434: **Blocking (related):** `route_backfill_pending.drain()` permanently removes every probed id, and this `continue` drops groups whose MLS state does not load (or has no routing component) without re-queueing.

The comment says full hydration/quarantine owns that failure, but unrecoverable groups are never placed in `unhydrated_groups`, so nothing owns them after a failed/empty probe. After one index miss they can become permanently unroutable for the session (UnknownGroup disposition instead of resolving / PeelDeferred).

Once unrecoverable groups are finished by the eager path (or otherwise given an owner), either do not put them in `route_backfill_pending` without that owner, or only remove a pending id after a successful route index / explicit terminal disposition.

erskingardner (commented):
**Request changes** (posted as comment — GitHub disallows `REQUEST_CHANGES` on your own PR).

Blocking issue: `hydrate_all_stored_groups` is not behavior-preserving for unrecoverable groups, and the route-backfill path has a matching ownership hole.

The two-phase design itself looks right for part 2 of #1161: cheap seed, fail-closed `GroupNotHydrated`, `ensure_hydrated` / quarantine parity, durable routes, shared record/message scans, and keeping session open on the eager bridge until part 3. Seeded/gated/promote tests are solid.

What should be fixed before merge is below.

crates/cgka-engine/src/engine.rs
```diff
@@ -1176,6 +1201,23 @@ impl<S: StorageProvider> Engine<S> {
...
+    /// cold-start route seed, never routing correctness — the in-memory
+    /// insert stays authoritative for this session.
+    pub(crate) fn index_transport_group_route(
+        &mut self,
+        transport_group_id: Vec<u8>,
+        group_id: &GroupId,
+    ) {
+        self.transport_group_id_index
+            .insert(transport_group_id.clone(), group_id.clone());
+        let _ = self
```
erskingardner, line 1177: **Blocking — make the current route crash-consistent.** `reindex_transport_group_id` runs after the MLS/group transaction has committed (for example `publish.rs:246-252`), and this deliberately drops the durable write error. After a rotation A→B, a crash or write failure can therefore leave the table containing A but not B. On cheap reopen, `hydrate_stable_groups_from_storage` treats *any* row for the group as proof that backfill is complete (`routed_group_ids`), so a message on current route B misses the index, is interpreted as MLS `GroupId(B)`, and takes the UnknownGroup path until background hydration happens. That is the exact pre-hydration window #1308 exposes. Please either commit the current-route projection atomically with the accepted routing-state change, or persist enough current/epoch metadata that the cheap pass can detect and repair a stale route set. Add a rotation/restart regression with a failure or crash between the MLS commit and this row write.

crates/traits/src/storage.rs
```diff
@@ -85,6 +85,43 @@ pub trait GroupStorage {
...
+    ) -> StorageResult<()> {
+        let _ = (transport_group_id, group_id);
+        Ok(())
+    }
+
+    fn list_transport_group_routes(&self) -> StorageResult<Vec<(Vec<u8>, GroupId)>> {
+        Ok(Vec::new())
+    }
+
+    fn delete_transport_group_routes_for_group(&self, group_id: &GroupId) -> StorageResult<()> {
```
erskingardner, line 146: **Blocking — the storage model needs route-level retirement.** The adopted routing spec requires accepting a prior address only until no epoch using it remains in either retained-history window ([routing-v1, lines 56-60](https://github.com/marmot-protocol/marmot/blob/4ad4ae21479c3f3fa9950c6fc4556a76941a62e1/app-components/nostr-routing-v1.md#L56-L60)). This API can only insert a route or delete *all* routes on disband; it stores no source epoch and cannot delete one expired route. Consequently every rotation is retained on disk forever and every old id is re-seeded after restart, turning the already-open #896 in-memory growth bug into unbounded durable state and accepting addresses beyond their specified overlap. Please carry the epoch/retention information needed to reconcile both windows and expose per-route pruning, with a rotate→advance-windows→restart test.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -2366,6 +2372,61 @@ impl<S: StorageProvider> Engine<S> {
...
+        let resolved = self.group_id_for_transport_group_id(transport_group_id)?;
+        if self.route_backfill_pending.is_empty()
+            || self
+                .transport_group_id_index
+                .contains_key(transport_group_id)
+            || self.storage.get_group(&resolved).is_ok()
+        {
+            return Ok(resolved);
+        }
+        let pending: Vec<GroupId> = self.route_backfill_pending.drain().collect();
```
erskingardner: **Blocking — do not make one untrusted miss load the whole account.** The first unknown group envelope drains the entire pending set and performs an `MlsGroup::load` for every stored group before payload authentication. That recreates the attacker-paced O(groups) amplification closed by #408; “once ever” only amortizes it after one request, and it repeats every cold start for providers using the default non-durable route methods. A 1,000-group account can still have its worker stalled by one arbitrary route id. Backfill proactively/in bounded background batches before trusting misses, or enforce a small per-event budget and keep the remainder queued; the inbound hot path should not drain all groups.

erskingardner (commented):
**Request changes** (posted as comment because GitHub disallows `REQUEST_CHANGES` on your own PR).

Three additional route-index blockers at exact head `496c059a`:

- The current-route projection is not crash-consistent: an old durable row masks a missing current route after rotation, so current traffic can take the UnknownGroup path during deferred hydration.
- The durable schema has no epoch/retention metadata or per-route delete, so prior routes survive beyond both retained-history windows and #896 becomes unbounded on disk.
- One unknown inbound route drains and MLS-loads every pending group, recreating #408 attacker-paced O(groups) work in a single request.

These are in addition to the existing unrecoverable-group eager-hydration and failed-backfill ownership findings. The two-phase seed/gate/promote structure remains a sound direction, but the recovery and routing lifecycle need closing before merge. CI is still in progress at this head.

crates/cgka-engine/src/engine.rs
erskingardner: Fixed in 4ab51882: the cheap pass now seeds the halt AND queues the unrecoverable group in `unhydrated_groups` like any other seeded group, so `hydrate_all_stored_groups` finishes it through `hydrate_one_stored_group` — route indexing, interrupted probe/apply recovery, and leave-request restore all run on the eager path again. The `GroupUnrecoverable` re-emit moved from the seed into full hydration, so it stays exactly once per open on the eager path (and arrives with the group's promotion on the deferred path). New regression test `hydrate_all_fully_hydrates_unrecoverable_groups` pins route re-indexing and the single re-emit for a halted group under `hydrate_all_stored_groups`.

crates/cgka-engine/src/message_processor/ingest.rs
erskingardner: Fixed in 4ab51882: the drain is gone. A pending id now leaves the set only on a successful route index or the explicit terminal no-routing-component disposition; MLS-load failures keep the id pending, and with the unrecoverable fix every backfill-pending group is also in `unhydrated_groups`, so `ensure_hydrated` (pipeline or on-demand) owns the failure and either indexes the route or quarantines the group.

crates/cgka-engine/src/engine.rs
erskingardner: Fixed in 4ab51882 via the detect-and-repair option: route rows now carry a `source_epoch` stamp refreshed at every commit apply (reindex fires at each apply site), and the session-open seed trusts a group's route set only when a stamp matches the record epoch. A crash or write failure between the commit apply and the route refresh leaves a lagging stamp, so the group goes to `route_backfill_pending` and current-route traffic resolves through the bounded backfill instead of the UnknownGroup path. Regression test `stale_route_stamp_triggers_backfill_repair_after_restart` performs a real rotation, deletes the current-route row to simulate the crash window, restarts, and asserts a message on the new route resolves, hydrates the group, and repairs the row.

crates/traits/src/storage.rs
erskingardner: Fixed in 4ab51882: rows carry `source_epoch`, and the API gains `delete_transport_group_route` (single-route retirement) and `delete_transport_group_routes_below_epoch` (bulk retirement below the retention cutoff). Every route refresh retires rows the retained-history window (pinned v1 `max_rewind_commits`) has moved past, so rotations no longer accumulate durable rows and restart no longer re-seeds addresses beyond their specified overlap. Regression tests: `rotated_route_is_retired_once_retention_window_passes` does a real rotate → advance-past-window → restart cycle at the engine level, plus per-route/below-epoch deletion coverage at the storage layer.

crates/cgka-engine/src/message_processor/ingest.rs
erskingardner: Fixed in 4ab51882: one unknown inbound route now probes at most `ROUTE_BACKFILL_PROBES_PER_MISS` (4) pending groups — never the whole account — with the remainder staying queued for later misses and for the app-layer background pipeline (part 3 drains backfill-pending groups first, which is the proactive bounded-batch path). A 1000-group account costs at most 4 MLS loads per arbitrary route id, and the default-storage-provider cold-start case is likewise bounded per event.

crates/cgka-engine/src/engine.rs
```diff
@@ -1176,6 +1201,23 @@ impl<S: StorageProvider> Engine<S> {
...
+            .storage
+            .put_transport_group_route(&transport_group_id, group_id);
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 4ab51882: the hydration/pending-commit-recovery narrative now documents `hydrate_one_stored_group`, the rotation paragraph is back on `reindex_transport_group_id`, and only the route-table text sits on `index_transport_group_route`.

crates/cgka-engine/src/engine.rs
```diff
@@ -2048,9 +2276,70 @@ impl<S: StorageProvider> Engine<S> {
...
+            .cloned()
+            .collect();
+        ids.extend(
+            self.unhydrated_groups
+                .iter()
+                .filter(|group_id| !self.route_backfill_pending.contains(*group_id))
+                .cloned(),
+        );
+        ids
+    }
```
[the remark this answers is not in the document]
erskingardner: Addressed in 4ab51882: `advance_convergence_inputs_until_settled`, both public `buffer_openmls_convergence_message*` wrappers, and `set_group_convergence_policy` now promote via `ensure_hydrated` before touching stored state, and `deferred_peel_cutoff_delay_ms` returns `Ok(None)` for seeded-but-unhydrated groups so it cannot persist lifecycle normalization against a provisional Stable entry. One deliberate skip: quarantined groups keep returning `Ok(None)` from `deferred_peel_cutoff_delay_ms` rather than `UnknownGroup` — that is pre-existing scheduler semantics ("no deadline to arm"), and changing it would alter the established quarantine contract outside this PR's scope.

erskingardner:
All findings addressed in 4ab51882 (threaded replies on each); #1308 rebased on top.

**Unrecoverable ownership**: the cheap pass seeds the halt and queues the group for full hydration like any other, so `hydrate_all_stored_groups` runs the complete per-group work for it (route indexing, probe/apply recovery, leave-request restore) — behavior-preserving again — with the `GroupUnrecoverable` re-emit moved into full hydration so it stays once per open. This also closes the backfill ownership hole: every backfill-pending group is in `unhydrated_groups`, so hydration owns MLS-load failures.

**Route lifecycle**: rows now carry a `source_epoch` stamp refreshed at every commit apply. The open seed trusts a route set only when a stamp matches the record epoch (crash-window detection → backfill repair); refreshes retire rows the pinned-v1 retained-history window has moved past, with per-route and below-epoch deletion in the storage API; and one unknown inbound route probes at most 4 pending groups per event, keeping #408's O(groups) amplification closed.

**CodeRabbit**: doc blocks reattached; remaining direct convergence entry points gated (one deliberate skip noted in-thread: `deferred_peel_cutoff_delay_ms` keeps its pre-existing `Ok(None)`-for-quarantined scheduler semantics).

New regression coverage: unrecoverable group fully hydrated eagerly (route re-indexed, halt re-emitted once); rotation crash-window stale-stamp repair across restart; rotate → advance-past-window → restart retirement; storage-level per-route/below-epoch deletion. Full engine suite, storage/traits suites, tracing audit, and `just fast-ci` pass on both branches.

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -380,6 +391,10 @@ impl<S: StorageProvider> Engine<S> {
         group_id: &GroupId,
         message: TransportMessage,
     ) -> Result<(), OpenMlsProjectionError> {
+        // Promote a seeded-but-unhydrated group before persisting convergence
+        // input against it (mdk#1161): the group's interrupted probe/apply
+        // recovery must run before any of its stored state is extended.
+        let _ = self.ensure_hydrated(group_id);
```
erskingardner: [P1] Stop when lazy hydration fails

`ensure_hydrated` removes the provisional epoch state and quarantines the group on failure, but this result is discarded. `buffer_openmls_convergence_message_with_time` then observes no epoch state, takes the `Retained` admission branch, and commits the new `MessageRecord` without ever reaching the live-group gate. That extends durable convergence state after the interrupted probe/apply recovery or group validation failed—the opposite of the comment above this call—and lets the row become convergence input if the group is later repaired. Please propagate/map the hydration error (and do the same in the `_at` entry point) before the first storage access; a regression test should force hydration failure and assert that no convergence row is persisted.

erskingardner (commented):
Request changes

Re-reviewed exact head `4ab518820b4651078ca6bc96bb2f7947b390a838`.

The corrective commit materially addresses all five blockers from my first review: eager handling of unrecoverable groups, ownership-preserving bounded route backfill, crash-detectable route freshness, retained-window route retirement, and the attacker-paced lookup bound.

One blocking issue remains in the newly added lazy-hydration calls: both public convergence-buffering entry points discard a failed `ensure_hydrated` result and can commit a retained convergence row for a group that was just quarantined. I left the concrete path and requested regression coverage inline.

Several CI jobs are still pending at this head.

crates/cgka-engine/src/distributed_convergence.rs
erskingardner: Fixed in bd438383: both public buffering entry points now propagate a failed promotion through a shared `ensure_hydrated_for_convergence` helper (mapped to the same missing-group view every quarantined group presents) before any storage access, so a group whose promotion just failed can never take the no-epoch-state Retained branch and persist a row. Already-quarantined groups keep their pre-existing retention behavior — that path predates this PR and its retained rows are the post-repair replay input. New regression test `failed_promotion_in_convergence_buffer_persists_no_row` forces the hydration failure and asserts the error propagates, the group quarantines, and no message row exists.

crates/cgka-engine/src/engine.rs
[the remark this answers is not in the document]
erskingardner: Fixed in bd438383 for the buffering entry points — both now fail closed via `ensure_hydrated_for_convergence` with a regression test asserting no row is persisted on a failed promotion. `set_group_convergence_policy` deliberately keeps its write: the policy is durable device-local convergence config, not group state — pre-#1161 it was writable for quarantined groups, quarantine repair reads it, and a promotion failure inside this call still quarantines the group while the config write stays consistent with that established behavior.

erskingardner:
Remaining blocker fixed in bd438383: both convergence-buffering entry points now propagate a failed `ensure_hydrated` before any storage access (shared fail-closed helper mapping to the quarantined missing-group view), with a regression test forcing the failure and asserting no convergence row is persisted. Already-quarantined groups keep their pre-existing retention semantics, and `set_group_convergence_policy` keeps its config write for the reasons noted in-thread. #1308 rebased on top; full engine suite and `just fast-ci` pass on both branches.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -793,6 +820,10 @@ impl<S: StorageProvider> Engine<S> {
         group_id: &GroupId,
         now_ms: u64,
     ) -> Result<bool, EngineError> {
+        // Promote a seeded-but-unhydrated group before reading its stored
+        // inputs (mdk#1161); a hydration failure quarantines it and the
+        // converge call below reports the blocked run as before.
+        let _ = self.ensure_hydrated(group_id);
```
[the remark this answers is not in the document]
erskingardner: Fixed in afba078a: `advance_convergence_inputs_until_settled` now propagates a fresh promotion failure (`ensure_hydrated(group_id)?`) before reading any convergence state, with a regression test asserting the error and the quarantine. Deliberately without an added `ensure_group_live` gate: already-quarantined groups keep their pre-existing semantics on this path (`ensure_hydrated` no-ops and the converge call reports the blocked run) — erroring `UnknownGroup` for them would change established caller-visible behavior beyond this PR's scope.

PR #1308, opened 2026-08-07
Defer runtime group hydration to a background pipeline after readiness

### LINKED ISSUE #1161: Bound local-ready startup cost instead of hydrating and snapshotting every group

erskingardner:
Part 3 of 3 for #1161 (app layer), stacked on #1307.

Does **not** close #1161 by itself: the bounded local membership read added to the issue's acceptance criteria on July 29 (member identifiers in the chat projection or an equivalent batched read, whitenoise-android#1534) is deliberately out of this stack's scope and now tracked as #1314. Everything else in #1161's acceptance list is covered by #1306/#1307/this PR.

- **Deferred runtime opens**: `SessionConfig::defer_group_hydration` opens with the engine's cheap seed pass; `AccountDeviceSession` gains `ensure_group_hydrated` / `hydrate_next_groups` / `unhydrated_group_ids`. Only `runtime_local_client` opts in — plain `MarmotApp::client()` consumers keep the eager open.
- **Background pipeline**: after signalling local readiness, the account worker fully hydrates groups in chat-list recency order, surfacing recovery events (`PendingCommitRecovered`, quarantines, restored leave requests) per batch and serving commands between batches. Group reads issued mid-pipeline hydrate exactly the group they name and answer live — "waits for that group only"; mutations and catch-ups join the existing #1151 startup deferral and replay in arrival order. Reconnect opens drain hydration eagerly before the steady-state loop resumes.
- **`GroupReadSnapshot`**: now captured after the pipeline, so it no longer gates readiness or any first read (deliberate deviation from deleting it outright: during the catch-up window `&mut client` is still held by the sync future, and the post-hydration capture is off every user-visible path).
- **Error surface**: MarmotKit maps the engine's retryable state to a new `GroupHydrationPending { group_id_hex }` error, distinct from `UnknownGroup`, for the remaining direct-read window. The once-only self-membership backfill no longer burns its marker on a pass that could not read rosters and re-runs after the pipeline.
- **Regression test** (the issue's final acceptance criterion): `tests/startup_hydration.rs` holds the pipeline with a test-only per-batch delay (`dev_startup_hydration_batch_delay_ms`, honored only with `test-policy-overrides`) and asserts readiness returns immediately, the persisted chat list is fully readable, a group read completes without waiting out the hold, and the pipeline telemetry confirms hydration was still blocked — then an unheld boot proves completion.

Measured effect (debug, 100 stored groups): account-ready wait **1862 ms → 241 ms**, open-time hydration 1300 ms → 3 ms, cold `chat_list` flat at ~160 ms.


erskingardner:
Full debug-mode benchmark run including the 1000-group case completed (deferred-hydration code, roster 1):

```
groups_100:  chat_list_ms=160 ready_ms=241 session_open_ms=55  group_hydration_ms(open)=3
groups_1000: chat_list_ms=388 ready_ms=479 session_open_ms=173 group_hydration_ms(open)=36
```

For scale: before this stack, 100 groups already cost 1862 ms of ready-wait with 1300 ms of open-time hydration — 1000 groups would have extrapolated past 10 s. With the pipeline, 10× the groups costs ~2× the ready-wait (dominated by the seed pass's per-record reads and worker spin-up, not MLS state), and the cold chat-list read stays in the same few-hundred-ms band. That satisfies the issue's "startup cost to first persisted chat-list read does not grow linearly with hydrated group/member state" criterion with measured data.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1036,6 +1099,184 @@ async fn run_app_runtime_account_worker(
...
+        }
+        if progress.remaining == 0 {
+            break;
+        }
+        tokio::task::yield_now().await;
+    }
+    shared.app_performance_telemetry().record(
+        AppPerformanceOperation::AccountGroupHydration,
+        pipeline_started.elapsed(),
+        true,
```
erskingardner: The pipeline always records `success = true`, including after the `hydrate_next_groups` error `break` above. Track a `pipeline_ok` (default true, set false on that break) and pass it here so a storage-level failure does not look like a successful `AccountGroupHydration` sample. Today `ensure_hydrated` mostly maps failures to quarantine/`UnknownGroup`, so this path may be rare — but the success bit should still be honest.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1036,6 +1099,184 @@ async fn run_app_runtime_account_worker(
...
+    loop {
+        // Test-only pre-batch hold (`test-policy-overrides` builds): keeps
+        // groups in the seeded state so integration tests can assert the
+        // persisted chat projection and per-group read behavior. Commands are
+        // still served while holding.
+        if !batch_delay.is_zero() {
+            let hold_until = TokioInstant::now() + batch_delay;
+            loop {
+                tokio::select! {
+                    _ = tokio::time::sleep_until(hold_until) => break,
```
erskingardner, line 1170: This hold `select!` watches the timer and `commands` only — not `shutdown` / lifecycle. With `BATCH_HOLD_MS = 30_000` and `APP_RUNTIME_ACCOUNT_SHUTDOWN_WAIT = 5s`, the regression test's `runtime.shutdown()` takes the abort path every run instead of a clean worker exit. Production `batch_delay` is zero so this is test-only, but adding the shutdown/lifecycle branches here (same shape as the catch-up window) would make the hold interruptible and keep the test off the abort path.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1036,6 +1099,184 @@ async fn run_app_runtime_account_worker(
...
+                .ensure_group_hydrated(&group_id);
+            let _ = respond.send(client.group_mls_state(&group_id));
+        }
+        AccountWorkerCommand::QuarantinedGroups { respond } => {
+            let _ = respond.send(Ok(client.quarantined_groups()));
+        }
+        AccountWorkerCommand::CatchUp { respond } => {
+            deferred.push(DeferredStartupCommand::CatchUp(respond));
+        }
+        other => deferred.push(DeferredStartupCommand::Command(Box::new(other))),
```
erskingardner, line 1296: Reads get the #1161 "wait for that group only" path via `ensure_group_hydrated`; mutations fall through to startup deferral and only replay after the later catch-up window. That matches the PR description / #1151 reuse, but it is weaker than the issue AC letter for mutations. If that tradeoff sticks, please call it out in the changelog/host notes — after this PR, `start()` returning no longer means a subsequent send is unblocked from remaining hydration work.

docs/marmot-architecture/telemetry.md
```diff
@@ -194,7 +194,7 @@ Collected operations:
 | `account_reconcile` | `AccountManager::reconcile()`, including local-signing account enumeration, stale-worker stop, pending-worker spawn, and ready wait. | Recorded every time reconcile runs, including implicit reconcile before catch-up. |
 | `account_open` | One sample per newly spawned account worker, from worker spawn until the ready signal. | Covers SQLCipher/session hydration, local engine-group reconciliation, local projection backfill, and read-snapshot capture; the mdk#1161 stage metrics below attribute those contributors separately. It does not include relay work. |
```
erskingardner: `account_open` still says the ready envelope covers "read-snapshot capture", but this PR moves readiness before both the hydration pipeline and snapshot capture (and `account_group_read_snapshot` below correctly says "after the ready signal"). Drop snapshot (and full hydration) from the `account_open` blurb so the catalogue matches the new boundary.

erskingardner (commented):
**Recommend merge** (GitHub won't let this account formally approve its own PR).

Part 3 does what #1161 asked for at the app boundary: seed-only runtime opens, readiness before the hydration pipeline, chat-list-ordered background drain with per-group live reads, reconnect eager drain, and a real held-pipeline regression test. The measured ready-wait drop matches the design.

Non-blocking notes inline. Biggest product caveat: mutations during the new post-ready hydration window still join the #1151 catch-up deferral (not per-group promote), so send latency after `start()` can still cover remaining hydration + catch-up even though the named group is already readable — worth an explicit host-facing note if Android will treat ready as "safe to send".

crates/marmot-uniffi/src/errors.rs
```diff
@@ -10,6 +10,13 @@ pub enum MarmotKitError {
     UnknownAccount { account_ref: String },
     #[error("unknown group: {group_id_hex}")]
     UnknownGroup { group_id_hex: String },
+    /// The group exists but its full hydration has not completed yet
+    /// (mdk#1161). Retryable: the runtime's background pipeline promotes the
+    /// group shortly after account readiness, and worker-routed reads wait
+    /// for exactly the named group. Distinct from `UnknownGroup` so hosts can
+    /// render "still loading" instead of "no such group".
+    #[error("group hydration pending: {group_id_hex}")]
+    GroupHydrationPending { group_id_hex: String },
```
erskingardner, line 19: **[P1] Bump the MarmotKit compatibility version for this FFI variant**

This changes the generated Swift/Kotlin public error enum, but the PR leaves the workspace at 0.9.10. The crate’s repository contract explicitly requires a workspace version bump whenever a UniFFI error variant changes (and #1134 said its skipped bump was deferred until the remaining binding work). Without that bump, hosts can receive a different binding ABI/API under the same MarmotKit version. Please bump the compatibility cohort and regenerate/smoke-test both bindings.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: **[P2] Record a failed outcome when background hydration aborts**

The error arm above publishes an account error and `break`s, but execution lands here and records `success = true` even though groups remain unhydrated. That makes the new stage metric report a successful full pipeline for the exact storage-failure case it is meant to expose, and contradicts the telemetry row’s “until every group is fully hydrated” contract. Carry the completion/error outcome out of the loop and record false on this path.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner, line 1170: **[P2] Let shutdown interrupt the test hold**

This inner select waits only for the 30-second test delay or a command; the shutdown receiver is not checked until after the hold. In the new regression, `runtime.shutdown()` therefore reaches the worker’s 5-second shutdown timeout and aborts the task instead of exiting promptly, despite the fixture comment claiming prompt shutdown. Add the shutdown/lifecycle arms to this select (as in the surrounding worker loops) so the test exercises graceful teardown rather than timeout cancellation.

docs/marmot-architecture/telemetry.md
```diff
@@ -194,7 +194,7 @@ Collected operations:
 | `account_reconcile` | `AccountManager::reconcile()`, including local-signing account enumeration, stale-worker stop, pending-worker spawn, and ready wait. | Recorded every time reconcile runs, including implicit reconcile before catch-up. |
```
erskingardner, line 194: **[P2] Update the adjacent account-open measurement contract**

The row immediately above still says `account_open` covers group hydration and read-snapshot capture. After this PR, its timer ends at the ready signal before the background pipeline and snapshot, so that description is now materially wrong. Update it alongside this row so dashboards do not interpret the reduced ready-wait as including work that moved outside the envelope.

erskingardner (commented):
**Found serious issues** (comment review because GitHub does not allow requesting changes on one’s own PR). The deferred-hydration pipeline is directionally sound, but this head should not close #1161 yet: the linked July 29 acceptance update requires lightweight member identifiers or an equivalent bounded/batched local membership read. This stack still exposes one worker-routed `group_members(group_id)` call per group, and the new regression covers only a single named read. Android’s 100-group member-derived UI therefore still needs 100 serialized FFI/worker commands—the exact remaining first-frame gap documented in whitenoise-android#1534. Please either add that bounded shared-layer surface and coverage here, or stop closing #1161 and keep the remaining criterion tracked. Current CI is still in progress at this exact head.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in dba74eb4: the pipeline tracks `pipeline_ok` (set false on the hydrate_next_groups error break) and passes it to the stage sample, so an aborted pipeline records a failure instead of a successful full run.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in dba74eb4: the hold select now watches the shutdown receiver and the lifecycle shutdown watch alongside commands, so `runtime.shutdown()` during the held regression takes the graceful worker exit instead of the 5s abort timeout.

crates/marmot-uniffi/src/errors.rs
erskingardner: Fixed in dba74eb4: bumped the workspace to 0.9.11 for the `GroupHydrationPending` binding-visible enum change. `cargo test -p marmot-uniffi` passes at the new version; I don't have the iOS/Android binding toolchains in this environment to run `xcframework.sh`/`kotlin-bindings.sh`, so the regenerate-and-vendor step remains with the release flow as usual.

docs/marmot-architecture/telemetry.md
erskingardner: Fixed in dba74eb4: the `account_open` row now says the timer ends at the ready signal with full hydration and snapshot capture measured separately after it, and the `account_session_open` row describes the cheap seed pass.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Called out in dba74eb4, in both host-facing places: the `Marmot::start` doc contract now states explicitly that ready is NOT "safe to send without waiting" — mutations issued before the initial catch-up replay after it, so first-send latency can still cover remaining hydration plus catch-up even when the named group is already readable — and the changelog bullet carries the same warning. The tradeoff itself stays: promoting a single group is not sufficient to unblock a send safely mid-catch-up (mdk#178's no-direct-write rule), so per-group mutation promotion would need its own design.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in dba74eb4 (same change as the P2 thread): `pipeline_ok` carries the abort out of the loop and the sample records false on that path.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in dba74eb4 (same change as the P2 thread): the hold select now has the shutdown/lifecycle arms in the same shape as the surrounding worker loops.

erskingardner:
All findings addressed at head dba74eb4 (threaded replies on each).

**Scope blocker**: agreed — this stack was deliberately scoped to exclude the July 29 bounded-membership criterion, so the PR body no longer closes #1161. The remaining criterion (lightweight member identifiers in the chat projection or an equivalent bounded/batched local membership read, per whitenoise-android#1534) is now tracked as #1314 with concrete acceptance criteria, including reuse of this PR's held-pipeline fixture for its regression test.

**P1**: workspace bumped to 0.9.11 for the `GroupHydrationPending` binding-visible enum change (`cargo test -p marmot-uniffi` green; binding regeneration stays with the release flow — no iOS/Android toolchains in this environment).

**P2s**: the hydration stage sample now records failure when the pipeline aborts, and the test hold is interruptible by shutdown/lifecycle so the regression exercises graceful teardown.

**Host contract**: `Marmot::start` docs and the changelog now state explicitly that ready ≠ safe-to-send-without-waiting — mutations before the initial catch-up replay after it. The `account_open` telemetry row matches the new ready boundary.

Full marmot-app suite (565 tests), marmot-uniffi, and the held-pipeline regression pass at this head.

crates/marmot-uniffi/src/errors.rs
erskingardner: Update: the version bump is reverted in 0e30235b — the maintainer will bump the compatibility cohort manually at release time, per the workspace's usual release-time versioning. The binding-visible `GroupHydrationPending` change still lands with that bump before hosts vendor regenerated bindings.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1036,6 +1099,194 @@ async fn run_app_runtime_account_worker(
...
+                        Some(command) => {
+                            handle_startup_hydration_command(client, command, deferred).await;
+                        }
+                        None => return StartupHydrationOutcome::Shutdown,
+                    },
+                }
+            }
+        }
+        loop {
+            match commands.try_recv() {
```
Datawav, line 1186: **P1:** Draining `commands.try_recv()` until the channel is empty can starve hydration indefinitely when producers continuously enqueue commands. Non-read commands also accumulate in the unbounded `deferred` vector, so hydration, catch-up, and mutation replay may never progress while memory grows. Process only a bounded command budget between hydration batches, and add a sustained-traffic regression test.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1036,6 +1099,194 @@ async fn run_app_runtime_account_worker(
...
+        {
+            Ok(progress) => progress,
+            Err(err) => {
+                let message =
+                    account_error_message("startup group hydration failed", &AppError::from(err));
+                publish_app_runtime_account_error(events, account_id_hex, account_label, message);
+                // Remaining groups stay gated retryable; the stage sample
+                // below must not report this aborted pipeline as a success.
+                pipeline_ok = false;
+                break;
```
Datawav, line 1217: **P2:** Breaking here permanently abandons every remaining group for this worker lifetime. The worker proceeds to steady state, whose `Members` and `GroupMlsState` handlers do not promote unhydrated groups, so those reads can return `GroupHydrationPending` indefinitely despite the `start()` contract. Retry failed batches with bounded backoff or preserve on-demand hydration in steady-state reads.

Datawav (commented):

## Final audit

### Need and design

The change addresses startup latency that scales with stored group count. Runtime sessions now perform a cheap seed pass, signal readiness, then hydrate groups in recency-ordered batches. Named reads hydrate their target group immediately, while mutations and catch-up commands retain arrival order for later replay. Non-runtime clients remain eagerly hydrated.

This is a reasonable separation of readiness from expensive initialization. Keeping eager hydration as the default also limits the behavioral change to callers with a background worker.

### Actionable findings

1. **P1 — Sustained command traffic can indefinitely starve hydration and grow memory without bound.**

   In `run_startup_hydration_pipeline`, the worker drains `commands.try_recv()` until the channel is empty before each hydration batch. Producers on another task or runtime thread can continuously refill the receiver, preventing the loop from ever reaching `hydrate_next_groups`.

   Non-read commands are simultaneously moved into the unbounded `deferred` vector. Under sustained traffic, hydration, initial catch-up, and all deferred mutations can therefore be delayed indefinitely while memory usage continues growing. This defeats the pipeline’s eventual-hydration guarantee and can leave sends permanently pending.

   Process a bounded command budget between hydration batches, or select commands and hydration through a scheduler that guarantees hydration progress. Add a sustained-traffic regression test that verifies all groups hydrate and deferred commands eventually replay; the current artificial-delay test only demonstrates command service during a finite hold.

2. **P2 — A single hydration error permanently abandons remaining groups during that worker lifetime.**

   When `hydrate_next_groups` returns a non-`UnknownGroup` error, the pipeline publishes an error, breaks, and proceeds into snapshot/catch-up and steady state. The steady-state `Members` and `GroupMlsState` handlers do not call `ensure_group_hydrated`, unlike the handlers used while the pipeline is active. Consequently, remaining groups can return `GroupHydrationPending` indefinitely despite the public `start()` contract saying pre-hydration worker reads wait for their named group.

   Retry the failed batch with bounded backoff, or preserve on-demand promotion in steady-state read handling. If abandoning the pipeline is intentional, the API documentation must explicitly describe the resulting degraded state and its recovery trigger. Reconnect currently repairs it, but requiring an unrelated reconnect is not a reliable recovery design.

### Alternatives considered

- Eager hydration preserves simpler semantics but retains the startup cost this change is intended to remove.
- Spawning an independent hydration task would improve scheduling isolation but would require synchronization around the mutable session.
- The smallest robust adjustment is a bounded number of commands per batch plus on-demand hydration in all live read paths, with bounded retry/backoff for batch-level failures.

_Advisory audit of head [`0e30235be3c1`](https://github.com/marmot-protocol/mdk/commit/0e30235be3c1f8636a6e127601dce13d6b032090); no approval, merge, push, checkout, or PR-code execution was performed._

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in 820c1477: the inter-batch drain now serves at most `STARTUP_HYDRATION_COMMAND_BUDGET` (8, one channel's worth) commands before hydrating the next batch, so sustained traffic interleaves with hydration instead of starving it. New regression test `sustained_command_traffic_does_not_starve_the_hydration_pipeline` floods worker-routed reads from the moment `start()` returns — with a per-batch hold keeping the pipeline slow enough that the flood genuinely overlaps — and asserts the pipeline's completion telemetry sample lands under load, then that every group serves.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in 820c1477 via the steady-state option: the steady-state `Members` / `GroupMlsState` arms now call `ensure_group_hydrated` before answering — a no-op in the normal completed-pipeline case, and on-demand promotion for pipeline-abort leftovers, so the `start()` on-demand contract holds for the worker's whole lifetime instead of only until the abort arm.

erskingardner:
Both open findings addressed at head 820c1477; branch rebased onto current master.

**P1 (hydration starvation)**: the inter-batch command drain is now bounded (`STARTUP_HYDRATION_COMMAND_BUDGET = 8`, one channel's worth per batch), so sustained worker traffic interleaves with hydration instead of deferring it — and the `deferred` mutation backlog stops growing without bound behind a stalled pipeline. New regression test `sustained_command_traffic_does_not_starve_the_hydration_pipeline` floods worker-routed reads from the moment `start()` returns against a deliberately slowed pipeline and asserts the pipeline completes under load.

**P2 (abort contract hole)**: steady-state `Members` / `GroupMlsState` promote their group on demand before answering — a no-op after a completed pipeline, and the correct fallback for pipeline-abort leftovers, so `GroupHydrationPending` can no longer persist for the worker's lifetime.

On the process notes: the version bump stays with the release flow per the maintainer (release always bumps the cohort before vendoring bindings), and the branch is rebased onto master. `just fast-ci`, the marmot-app suite, and both startup-hydration regression tests pass at this head; the non-author approval is yours to give.

PR #1309, opened 2026-08-07
perf(app): detach invite read-side catch-up

### LINKED ISSUE #466: [Bug]: Profile and shrink invite-member latency (3-10s for small groups, no progress signal)
## Problem

Inviting a new member to an existing group from Dark Matter Android can take 3–10 seconds before the new member appears in the inviter's member list. The Android-side fix for the chat-open regression (`darkmatter-android` PR marmot-protocol/mdk#502) ruled out the global all-accounts catch-up as the cause for *this* path — the add-member flow is already group-scoped:

- `ConversationController.inviteMembers(...)` calls the engine `inviteMembers(account, groupId, refs)`.
- It then optionally promotes admins, calls `refreshMembers()` (`groupMlsState` + `groupDetails`) for the same group.
- It does not call any global catch-up on this path.

That points to engine / binding latency on the invite path itself rather than app orchestration. From the Android issue (`marmot-protocol/darkmatter-android#524`): either the FFI `inviteMembers` call is slow to return, the MLS commit + Welcome publish is the long pole, or the subsequent `groupMlsState` / `groupDetails` reads are themselves blocking on convergence.

## Suggested profiling

Add timestamps around the engine-side stages of the add-member flow to identify the long pole:

1. `inviteMembers(account, groupId, refs)` — key-package fetch fan-out, MLS Add proposal + Commit, Welcome publish.
2. `promoteAdmin(...)` (when adding as admin) — separate admin-policy app-component update.
3. `groupMlsState(account, groupId)` / `groupDetails(account, groupId)` — projection read after commit.
4. Convergence: does the new member show up in the projection synchronously after `inviteMembers` returns, or does the app have to wait for the group-state subscription to catch up?

If the long pole is convergence (#4), the engine should either (a) return a synchronously-updated projection from `inviteMembers` itself, or (b) expose a "post-invite local state is consistent" signal the app can await without holding the user on a spinner.

## Constraints

- The reported 3–10s latency is on a 2-person group adding a single invitee whose key package is already published. Multi-add to a larger group is expected to be slower, but the small-group case is what users notice and report as broken.
- The app already provides UI feedback (loading indicator) during the operation. The goal here is to either shrink the engine-side latency or expose a faster intermediate state so the UI doesn't sit at the spinner for seconds.

## Consumer

- `marmot-protocol/darkmatter-android#524`

## Acceptance

- A profiling pass identifies which engine-side stage owns the 3–10s window.
- Either (a) the operation completes within a UX-acceptable window for the small-group case (target: <1s for 2→3 members), or (b) the engine exposes a synchronous post-invite projection so the app's member list can render the new state without waiting for the convergence stream.

agent-p1p:
## Summary

- return a confirmed existing-group invite without waiting for result-independent read-side catch-up on other local accounts
- keep that catch-up asynchronous, privacy-safe, and visible through the existing `GroupInvitePostMutationCatchUp` telemetry
- keep the inviting account's worker available for immediate `group_members` and `group_mls_state` projection reads
- document the caller-latency change in the compatibility-cohort changelog

## Profiling attribution and safety

The issue's measured invite took 6.312542772 seconds, with 6.217587936 seconds attributed to `GroupInvitePostMutationCatchUp`. Source tracing confirmed that stage ran only after the account worker completed KeyPackage resolution, pre-send convergence, required relay acknowledgement, durable local refresh, and projection-update publication. Its failure was already warn-and-continue and could not change the confirmed mutation result.

This change detaches only that read-side catch-up and excludes the already-refreshed inviting account. Required KeyPackage validation, convergence, commit/Welcome publication, relay acknowledgement, and durable local confirmation remain awaited.

## Verification

- focused same-account projection regression is RED against the parent implementation and GREEN with this change
- `just fast-ci` passes: format, naming, default/OTLP workspace checks, clippy, and convergence policy pin under Rust 1.97.1
- GitHub CI passes all required checks, including four Rust test shards and relay runtime tests

Fixes #466

## Summary by CodeRabbit

* **New Features**
  * Existing-group invitations now complete after relay acknowledgement and local data refresh, without waiting for all background synchronization.
  * Background catch-up continues automatically, with improved handling for concurrent requests and pending push registration.
  * Read operations remain available while synchronization is in progress.

* **Bug Fixes**
  * Improved reliability for concurrent and reciprocal invitations.
  * Ensured successful invitations complete and deliver correctly even when another invitation fails.


agent-p1p:
## Pip adversarial review

Reviewed head `778bc04245050a9ef5938a93434929d0fc66571e`.

BLOCKING:

- **The detached catch-up can move the relay stall onto the immediate projection read instead of removing it.** `crates/marmot-app/src/runtime/commands.rs:373-390` spawns `catch_up_accounts()`, which enqueues `AccountWorkerCommand::CatchUp` on every worker, including the inviting account (`crates/marmot-app/src/runtime/mod.rs:3659-3691`). That worker consumes one FIFO command at a time (`crates/marmot-app/src/runtime/account_worker.rs:731-744`), and `CatchUp` runs `client.sync().await` inline (`account_worker.rs:1023-1063`). The post-invite `group_members` / `group_mls_state` reads use the same worker channel (`commands.rs:122-156`); `group_details_for` awaits both (`crates/marmot-uniffi/src/commands/group.rs:43-90`), and `invite_members_detailed` calls that projection builder immediately after the invite (`group.rs:698-714`). If the spawned task enqueues first—as it can while the result crosses the FFI boundary—the read blocks behind the same relay catch-up. The invite future is faster, but the user-visible “invite then render members” path can still take the original 3–10 seconds. Scope the detached catch-up away from the already-refreshed inviting account, or otherwise guarantee its projection reads cannot queue behind this network sync.

- **The regression does not cover that end-to-end boundary.** `crates/marmot-app/src/tests.rs:545-613` proves only that `invite_members` returns while the detached catch-up is parked. It never calls `group_mls_state` or `group_members` on Alice while the subscribe remains blocked, so it stays green even when the next projection read stalls. Extend it to issue the same-account post-invite projection read before `release_subscribe()` and require that read to complete within the bound.

SUGGESTION:

none.

NITPICK:

none.

SUMMARY: The change preserves required publication and durable local refresh, has the required Unreleased changelog entry, and the focused regression passes locally. It does not yet prove or guarantee the Android-visible latency improvement because the all-account catch-up still serializes through the inviting account’s worker.

VERDICT: REQUEST_CHANGES

Validation: `cargo test -p marmot-app --lib tests::invite_members_returns_before_post_mutation_catch_up_finishes --locked -- --exact --nocapture` — 1 passed.

agent-p1p:
Addressed the blocking findings in d010d70a. Detached catch-up now excludes the already-refreshed inviting account, so its worker remains available for immediate projection reads while other accounts catch up. The regression now blocks Alice’s post-publish subscription and requires both group_members and group_mls_state to complete before any release. Verified with the focused regression and just fast-ci.

crates/marmot-app/src/tests.rs
```diff
@@ -522,6 +565,101 @@ async fn account_local_ready_before_subscribe_body() {
...
+        .with_test_relay_client(relay.clone());
+    let runtime = MarmotAppRuntime::new(app);
+    runtime.reconcile_accounts().await.unwrap();
+    runtime.catch_up_accounts().await.unwrap();
+    let group_id = runtime
+        .create_group("alice", "invite latency", &[], None)
+        .await
+        .unwrap();
+
+    relay.block_account_subscribe_after_next_publish(hex::decode(&alice.account_id_hex).unwrap());
```
erskingardner: Blocking: this arms a subscribe stall for the inviting account, but detached catch-up now excludes Alice. Bob's `CatchUp` → `sync()` subscribes as Bob, so this block never engages and catch-up finishes immediately.

`bc24cd0` had the right shape with a generic post-publish subscribe block plus `wait_for_blocked_subscribe()` before asserting invite return. Keep Alice excluded for the projection-read invariant, but stall a non-excluded account (Bob, or a third local account) so the test is actually RED against an awaited catch-up.

crates/marmot-app/src/tests.rs
```diff
@@ -522,6 +565,101 @@ async fn account_local_ready_before_subscribe_body() {
...
+            .invite_members(
+                "alice",
+                &invite_group_id,
+                std::slice::from_ref(&bob_account_id),
+            )
+            .await
+    });
+    tokio::time::timeout(std::time::Duration::from_millis(250), invite)
+        .await
+        .expect("confirmed invite must return before read-side catch-up finishes")
```
erskingardner, line 635: Blocking: this 250ms bound only proves detachment if catch-up is still in flight. With the Alice-only block inert, Bob's catch-up completes on the mock relay immediately, so an `await self.catch_up_accounts_excluding(Some(&inviting_account_id))` (or even a full awaited `catch_up_accounts()` if that path stays fast) can still pass.

Wait until the non-excluded account's subscribe is blocked, then assert invite already returned / returns within the bound while that stall holds. Same for the later `group_members` / `group_mls_state` reads: they need to complete before `release_subscribe()`, not after catch-up has already succeeded.

crates/marmot-app/src/tests.rs
```diff
@@ -522,6 +565,101 @@ async fn account_local_ready_before_subscribe_body() {
...
+    assert_eq!(mls_state.member_count, 2);
+
+    let before_release = runtime
+        .shared_services()
+        .app_performance_telemetry()
+        .snapshot();
+    assert_eq!(before_release.group_invite_members.successes, 1);
+    assert_eq!(
+        before_release.group_invite_post_mutation_catch_up.successes, 1,
+        "other-account catch-up should finish without using alice's blocked subscription"
```
erskingardner: This assertion is currently vacuous: `successes == 1` before `release_subscribe()` is expected when nothing was blocked. Once Bob (or another non-excluded account) is the stalled subscriber, this should flip back to the v1 check that catch-up has not finished (`attempts == 0` / `successes == 0`) until release.

erskingardner (changes requested):
Request changes on the regression coverage. The production change itself looks right for #466: post-mutation catch-up was already warn-and-continue, the measured long pole was `GroupInvitePostMutationCatchUp`, required publish/ack/local refresh stay on the await path, and excluding the inviting account keeps same-account projection reads off that worker FIFO.

The focused test no longer proves the latency property it claims. After Alice was excluded, the Alice-only subscribe block cannot stall the catch-up that still runs (Bob), so invite-return and projection-read bounds can pass even if catch-up is awaited again on a fast mock path. Restore a stall on a non-excluded account before asserting those bounds.

crates/marmot-app/src/runtime/commands.rs
```diff
@@ -370,21 +371,26 @@ impl AccountManager {
...
-                tracing::warn!(
-                    target: "marmot_app::runtime",
-                    method = "invite_members",
-                    error_kind = error.privacy_safe_kind(),
-                    "committed mutation succeeded but post-mutation catch-up failed; state will refresh on the next cycle"
+            let post_mutation = self.clone();
+            tokio::spawn(async move {
+                let catch_up_started_at = Instant::now();
+                let catch_up = post_mutation
+                    .catch_up_accounts_excluding(Some(&inviting_account_id))
```
erskingardner: The exclusion is local to this spawned task. If Alice and Bob invite concurrently, the Alice task includes Bob while the Bob task includes Alice, so each can enqueue `AccountWorkerCommand::CatchUp` onto the other inviting worker. Those workers service commands serially and `CatchUp` awaits `client.sync()` inline; a queued cross-catch-up can therefore run immediately after `InviteMembers` and before the same-account `group_members` / `group_mls_state` reads, recreating the 3-10 second stall this change is meant to remove. The current single-invite test cannot exercise this. Please coalesce or coordinate detached catch-up so a just-inviting account cannot be occupied by another invite catch-up, and add a two-account concurrent-invite regression with both account subscriptions blocked.

erskingardner (changes requested):
The single-invite regression now protects immediate same-account reads, but concurrent invites from different local accounts can still cross-enqueue detached catch-ups onto each other and recreate the blocked projection-read path. Please coordinate the detached work across invites and cover that overlap.

agent-p1p:
## Pip adversarial re-review

Reviewed exact head `d010d70ad3642d1c035d60d7cf6ff9e0e2c4d98c`.

BLOCKING:

- **Concurrent invites can enqueue detached catch-up onto each other’s inviting worker.** `crates/marmot-app/src/runtime/commands.rs:374-379` excludes only the account for that one invite, while `crates/marmot-app/src/runtime/mod.rs:3671-3686` queues `CatchUp` on every remaining worker. If Alice and Bob finish invites concurrently, Alice’s task queues catch-up on Bob and Bob’s queues it on Alice. Each worker processes commands serially and `CatchUp` awaits `client.sync()` inline (`crates/marmot-app/src/runtime/account_worker.rs:1051-1086`), so either cross-catch-up can run before that account’s immediate `group_members` / `group_mls_state` reads and recreate the relay stall. Coordinate or coalesce the detached work across in-flight invites, and cover the two-account overlap.

- **The regression no longer proves the detached catch-up is still in flight at the caller boundary.** `crates/marmot-app/src/tests.rs:584` arms the subscribe stall for Alice, but production excludes Alice. Bob’s non-excluded catch-up therefore finishes on the fast mock path, and the loop at `tests.rs:603-619` exits through `catch_up_finished`; the reads at `tests.rs:621-630` then run after catch-up already succeeded. This stays green if the non-excluded catch-up is accidentally awaited again. Stall Bob or another non-excluded account, wait until that subscription is actually blocked, then require invite return and both Alice reads before release.

SUGGESTION:

none.

NITPICK:

none.

RESOLUTION_EVIDENCE:

- Prior blocker 1: **resolved for a non-overlapping invite.** `resolve(account_ref)` supplies the canonical account id at `commands.rs:361`; `worker_commands` resolves the same account and indexes workers by that id (`runtime/mod.rs:3861-3872`); the filter compares that exact map key at `runtime/mod.rs:3673-3676`. The inviting worker is excluded from its own detached catch-up. The concurrent cross-invite race above remains.
- Prior blocker 2: **not fully resolved.** Both requested reads were added and the test would catch re-including Alice, but its only relay stall targets the now-excluded account. It does not keep any catch-up in flight while asserting invite return and the projection reads.

Validation on an isolated archive of the exact head:
- `cargo test -p marmot-app --lib tests::invite_members_keeps_same_account_projection_reads_off_detached_catch_up --locked -- --exact --nocapture` — 1 passed.
- `cargo fmt --all --check` — passed.
- `cargo clippy -p marmot-app --all-targets --locked -- -D warnings` — passed.

SUMMARY: The canonical single-invite exclusion is correct and the immediate reads are now present, but the regression lost its non-excluded in-flight stall and concurrent invites can still cross-occupy both inviting workers with detached relay sync. The exact head is not LGTM.

VERDICT: REQUEST_CHANGES

SENSITIVE_PATHS: `crates/marmot-app/src/runtime/commands.rs`, `crates/marmot-app/src/runtime/mod.rs`

crates/marmot-app/src/tests.rs
agent-p1p: Addressed in signed commit `598c231a`. The single-invite regression now arms Bob, waits until Bob's subscription is actually blocked, and only then asserts Alice's invite and immediate projection reads complete.

crates/marmot-app/src/tests.rs
agent-p1p: Addressed in `598c231a`. The test waits for the non-excluded Bob catch-up to block before the 250 ms invite and `group_members` / `group_mls_state` bounds. Bob is released only after those reads complete.

crates/marmot-app/src/tests.rs
agent-p1p: Addressed in `598c231a`. Before releasing Bob, the test now asserts detached catch-up successes are `0`; after release it waits for exactly one success. The vacuous alternate branch was removed.

crates/marmot-app/src/runtime/commands.rs
agent-p1p: Addressed in `598c231a`. Overlapping invites now join a shared cancellation-safe cohort registered before worker dispatch. One catch-up runs after all cohort invites finish and excludes every cohort account. A publication barrier regression forces Alice/Bob overlap, verifies both invites and both immediate projections, and asserts exactly one detached catch-up.

agent-p1p:
## Pip adversarial re-review

Reviewed exact head `598c231a4abb83914fed819b5e3c0af8b0d03b44`; author `agent-p1p` and head `pip/mdk-466` reverified immediately before posting.

VERDICT: **LGTM**

RESOLUTION_EVIDENCE:

- **Concurrent cross-account invite blocker: resolved.** `InviteCatchUpGuard` registers the canonical inviting account before dispatching its worker command (`crates/marmot-app/src/runtime/commands.rs:455-457`). Overlapping invites remain in one coordinator cohort until the last active guard finishes, then emit one detached catch-up excluding every cohort account (`commands.rs:61-107`). `catch_up_accounts_excluding` filters those exact worker-map keys before enqueueing `AccountWorkerCommand::CatchUp` (`crates/marmot-app/src/runtime/mod.rs:3672-3695`). The two-account publication barrier regression forces both invitations to overlap and verifies exactly one cohort catch-up plus prompt projection reads for both inviters.
- **In-flight regression blocker: resolved.** The single-invite regression now arms Bob, the non-excluded account, waits until Bob subscription catch-up is actually blocked, then requires Alice invite return and Alice `group_members` / `group_mls_state` reads to complete before release (`crates/marmot-app/src/tests.rs:596-675`). Pre-release catch-up successes remain zero; after release the detached catch-up completes once.
- **Final diff:** required invite publication/relay acknowledgement and local projection refresh still complete in the account worker before its response (`crates/marmot-app/src/runtime/account_worker.rs:1314-1342`). Catch-up remains best-effort, detached, and privacy-safe (`error.privacy_safe_kind()` only). The Unreleased changelog entry is present. No version, conformance, fixture, or lockfile churn. No new blocking correctness, security, privacy, or concurrency findings.

Validation on the exact head:

- `cargo test -p marmot-app --lib tests::invite_members_keeps_same_account_projection_reads_off_detached_catch_up --locked -- --exact --nocapture` — 1 passed.
- `cargo test -p marmot-app --lib tests::concurrent_invites_coalesce_catch_up_away_from_both_inviting_accounts --locked -- --exact --nocapture` — 1 passed.
- `cargo fmt --all --check` — passed.
- `cargo clippy -p marmot-app --all-targets --all-features --locked -- -D warnings` — passed.
- Live GitHub checks are green.

SENSITIVE_PATHS: `crates/marmot-app/src/runtime/commands.rs`, `crates/marmot-app/src/runtime/mod.rs`.

This is review only. I did not merge.

crates/marmot-app/src/runtime/commands.rs
```diff
@@ -26,7 +26,101 @@ use crate::{
...
+    fn begin_invite_catch_up(&self, account_id: &str) {
+        let mut coordinator = self
+            .invite_catch_up
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        *coordinator
+            .active_by_account
+            .entry(account_id.to_owned())
+            .or_default() += 1;
+        coordinator.cohort_accounts.insert(account_id.to_owned());
```
Datawav: `cohort_accounts` includes every attempted inviter, and the entire set is later excluded from the single coalesced catch-up. With reciprocal overlapping invites, this excludes Alice from processing Bob's committed Welcome and Bob from processing Alice's; the same problem occurs when one overlapping invite fails because its inviter remains in the cohort. Preserve exclusions per successful mutation (excluding only that mutation's inviter), or schedule separate catch-ups when those requirements overlap. Add reciprocal success and mixed success/failure tests that assert receipt of the incoming invite without another refresh.

Datawav (commented):

### High — Cohort-wide exclusion can suppress invite delivery to overlapping inviters

`begin_invite_catch_up` adds every attempted inviter to `cohort_accounts`, and the single detached catch-up excludes that entire set (`crates/marmot-app/src/runtime/commands.rs:42-87`). If Alice and Bob concurrently invite each other—or Alice’s invite succeeds while Bob’s overlapping invite fails—both accounts are excluded. Consequently, Bob may not process Alice’s committed Welcome, and vice versa, until an unrelated later refresh.

The concurrency test only verifies each inviter’s locally mutated group; it never verifies visibility of the incoming reciprocal invite, so it does not cover this failure.

**Need:** Detaching result-independent catch-up is justified by the measured latency, while excluding the account whose worker already completed the mutation preserves immediate same-account reads.

**Design:** Exclusion must remain mutation-specific: each successful invite should catch up all accounts except its own inviter. Coalescing must preserve those per-mutation delivery requirements rather than unioning all inviters into one exclusion set.

**Alternatives:** Track mutation-to-inviter relationships and compute which accounts still require refresh; or, when multiple inviters overlap, avoid cohort-wide exclusion and schedule separate catch-ups. Add reciprocal-invite and mixed success/failure tests asserting that recipients observe the incoming invites without requiring another refresh cycle.

_Advisory audit of head [`598c231a4abb`](https://github.com/marmot-protocol/mdk/commit/598c231a4abb83914fed819b5e3c0af8b0d03b44); no approval, merge, push, checkout, or PR-code execution was performed._

crates/marmot-app/src/runtime/commands.rs
erskingardner: Fixed in f8d036a0. I removed the cohort-wide exclusions; every successful invite now launches a normal all-account detached catch-up. During steady-state catch-up, workers serve members/MLS/quarantine reads from a pre-sync snapshot while preserving FIFO for mutations, so inviters stay responsive without being excluded from Welcome ingestion. Added real-relay regressions `overlapping_reciprocal_invites_deliver_both_incoming_welcomes` and `successful_invite_delivers_while_overlapping_invite_fails`, plus concurrent/read-responsiveness unit coverage. The focused tests and `just fast-ci` pass.

erskingardner (dismissed):
Re-reviewed at 90b93f59 after resolving the remaining invite catch-up delivery issue and integrating current master. Reciprocal-success and mixed success/failure real-relay regressions pass, invite projection-read regressions pass, startup-hydration integration tests pass, and just fast-ci is green.

erskingardner (approved):
Re-reviewed at 968cf2b3. The CI-discovered deferred-hydration integration failure is fixed by running live-state projection reconciliation after hydration; the exact failing CLI flow now passes locally alongside the startup-hydration suite, invite concurrency/delivery regressions, and just fast-ci.

crates/marmot-app/src/runtime/commands.rs
```diff
@@ -27,6 +27,27 @@ use crate::{
...
+            if let Err(error) = catch_up {
+                tracing::warn!(
+                    target: "marmot_app::runtime",
+                    method = "invite_members",
+                    error_kind = error.privacy_safe_kind(),
+                    "committed mutation succeeded but post-mutation catch-up failed; state will refresh on the next cycle"
+                );
+            }
+        });
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in follow-up PR #1340 (7529c04e): invite catch-up handles are registered before return, subscribe to runtime shutdown, stop being accepted once shutdown starts, and are awaited alongside account workers. I intentionally did not coalesce successful invites: a catch-up already in flight can predate a second committed mutation, so coalescing without a trailing generation would recreate the reciprocal/mixed-delivery gap fixed by #1309.

crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -144,6 +159,64 @@ struct BlockKeyPackagesWhileArmed {
...
+
+    async fn wait_for_blocked(&self, count: usize) {
+        while self.blocked.load(Ordering::SeqCst) < count {
+            self.entered.notified().await;
+        }
+    }
+
+    fn release(&self) {
+        self.release.notify_waiters();
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in follow-up PR #1340 (7529c04e): the `Notified` future is created, pinned, and enabled before the blocked counter is published, so `notify_waiters()` cannot be missed.

Datawav:

The previously requested change is fixed: cohort-wide inviter exclusion was removed. Each successful invite now schedules an unfiltered all-account catch-up, while failed invites schedule none. New reciprocal-invite and mixed success/failure tests verify recipient-side `GroupJoined` delivery.

### Low — Duplicate hydration priorities waste batch capacity and inflate progress

In `crates/cgka-session/src/lib.rs`, `hydrate_next_groups` collects caller-supplied `order` entries without deduplicating them. Repeated group IDs can occupy multiple budget slots; subsequent `ensure_hydrated` calls then count an already-hydrated duplicate again. A batch may therefore hydrate fewer than `budget` distinct groups while `HydrationProgress.hydrated` overstates how many groups were promoted, contradicting its documented telemetry semantics.

Deduplicate `order` while preserving first-occurrence order before applying the budget, and count only actual state transitions.

_Advisory audit of head [`968cf2b383ff`](https://github.com/marmot-protocol/mdk/commit/968cf2b383ffc04208cdf9bc805c38f9de792a70); no approval, merge, push, checkout, or PR-code execution was performed._

PR #1312, opened 2026-08-07
feat: expose lightweight group roster projection

### LINKED ISSUE #1297: Expose a single lightweight group-roster projection over UniFFI (replace the groupMlsState + groupDetails pair)
Expose a single lightweight group-roster projection over UniFFI (replace the groupMlsState + groupDetails pair)

## Summary

Clients currently need two serialized runtime calls to display a group's membership: `group_mls_state` (used by Android as an eviction/OpenMLS-replay probe) followed by a full `group_details` reload. A single purpose-built roster projection — roster + epoch + self-membership/eviction state in one call — would halve the bridge round-trips on every membership view and remove the need for client-side probe heuristics.

## Evidence (master `bca10777`)

- `AppClient::group_mls_state` (`crates/marmot-app/src/client/mod.rs:329-344`) reads the group record (epoch, member count, components).
- `group_details`/member records rebuild the full roster with profiles (`members_with_profiles`, `client/mod.rs` ~287+).
- Android's refresh performs both sequentially and documents the probe's purpose in `ConversationController.refreshMembers`; its probe-gating helper is limited by `AppGroupRecordFfi` not carrying roster/epoch/revision fields.
- Telemetry hooks already exist for both reads (`GroupDetailsRead`, `GroupMlsStateRead`, `crates/marmot-app/src/app_telemetry.rs:38-39`), so a unified projection can prove its win with existing instrumentation.

## Suggested shape

- One UniFFI call, e.g. `group_roster(account, group_id)`: member list with display identity, epoch, caller's membership/eviction state (surfaced as data or a typed error instead of a replay-induced failure), and a roster revision for cheap change detection.
- Keeps SQLite as the source of truth; this is a read projection, not a cache.
- Client adoption: Android membership screens drop to one call per refresh and retire the probe heuristic.

Related: marmot-protocol/whitenoise-android#1847 (membership double-read, client side of this projection).

Labels: `enhancement`, `MEDIUM`, `performance`, `Bindings Related`

agent-p1p:
Fixes #1297

## Summary

- Adds `groupRoster`, a single MarmotKit/UniFFI membership projection with enriched member identities, MLS epoch, roster revision, caller self-membership or eviction state, and lifecycle state.
- Routes roster reads through the serialized account worker. Initial and steady-state catch-up serve a session-consistent snapshot without blocking projection reads; steady state otherwise reads the live session.
- Sources self-membership from the storage-owned projection and reconciles it against the authoritative MLS roster, so local leave and observed eviction are visible without restarting the worker.
- Keeps `groupDetails`, `groupMlsState`, and `groupMembers` unchanged and adds dedicated performance and OTLP latency metrics.
- Keeps the workspace and conformance fixtures at `0.9.10`. Workspace version bumps are manual release operations, not part of feature or review-feedback work.
- Updates repository agent guidance to make that manual-only version policy explicit.

## Verification

- `just fast-ci`
- `cargo test -p cgka-conformance-simulator canonical_vector_fixtures_match_generated_traces --locked`
- `cargo test -p marmot-app --lib --locked -- --test-threads=1` — 588 passed
- `cargo test -p marmot-app group_roster --locked`
- `cargo test -p marmot-app --test startup_hydration chat_list_is_readable_and_group_reads_wait_per_group_while_hydration_is_held --features test-policy-overrides --locked`
- `cargo test -p storage-sqlite --locked` — 324 passed
- `cargo test -p marmot-uniffi --locked` — 100 unit, 2 media fixture, and 28 smoke tests passed
- Takeover review of the rebased final diff — no remaining actionable findings

## Sensitive paths

Touches app group-state/runtime read projection and UniFFI binding code. No cryptographic algorithms, MLS state mutation, key material, trust anchors, or authentication flows changed. The group-state core touch should still be treated as sensitive at the merge gate.

## Summary by CodeRabbit

- **New Features**
  - Added a lightweight group roster API with member identities, display names, admin and self status, membership state, member count, lifecycle state, epoch, and revision.
  - Exposed group roster access through runtime and FFI interfaces.
  - Roster results are available before initial synchronization completes when startup data is available.
  - Added accurate reporting for members who leave or are administratively removed.

- **Monitoring**
  - Added performance tracking and exported metrics for group roster reads.

### DIFF
--- a/crates/marmot-app/src/app_telemetry.rs
+++ b/crates/marmot-app/src/app_telemetry.rs
@@ -43,6 +43,7 @@ pub(crate) enum AppPerformanceOperation {
     GroupPromoteAdmin,
     GroupDetailsRead,
     GroupMlsStateRead,
+    GroupRosterRead,
     MediaUpload,
     MediaDownload,
     HostSplashReady,
@@ -122,6 +123,8 @@ pub struct AppPerformanceSnapshot {
     pub group_details_read: AppPerformanceOperationSnapshot,
     #[serde(default)]
     pub group_mls_state_read: AppPerformanceOperationSnapshot,
+    #[serde(default)]
+    pub group_roster_read: AppPerformanceOperationSnapshot,
     pub media_upload: AppPerformanceOperationSnapshot,
     pub media_download: AppPerformanceOperationSnapshot,
     #[serde(default)]
@@ -162,6 +165,7 @@ struct AppPerformanceTelemetryInner {
     group_promote_admin: AppPerformanceOperationTelemetry,
     group_details_read: AppPerformanceOperationTelemetry,
     group_mls_state_read: AppPerformanceOperationTelemetry,
+    group_roster_read: AppPerformanceOperationTelemetry,
     media_upload: AppPerformanceOperationTelemetry,
     media_download: AppPerformanceOperationTelemetry,
     host_splash_ready: AppPerformanceOperationTelemetry,
@@ -332,6 +336,9 @@ impl AppPerformanceTelemetry {
             AppPerformanceOperation::GroupMlsStateRead => {
                 inner.group_mls_state_read.record(duration, success);
             }
+            AppPerformanceOperation::GroupRosterRead => {
+                inner.group_roster_read.record(duration, success);
+            }
             AppPerformanceOperation::MediaUpload => inner.media_upload.record(duration, success),
             AppPerformanceOperation::MediaDownload => {
                 inner.media_download.record(duration, success);
@@ -399,6 +406,7 @@ impl AppPerformanceTelemetry {
             group_promote_admin: inner.group_promote_admin.snapshot(),
             group_details_read: inner.group_details_read.snapshot(),
             group_mls_state_read: inner.group_mls_state_read.snapshot(),
+            group_roster_read: inner.group_roster_read.snapshot(),
             media_upload: inner.media_upload.snapshot(),
             media_download: inner.media_download.snapshot(),
             host_splash_ready: inner.host_splash_ready.snapshot(),
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -777,13 +777,14 @@ impl AppClient {
         group_id: &GroupId,
     ) -> Result<crate::groups::AppGroupRosterSession, AppError> {
         let group_id_hex = hex::encode(group_id.as_slice());
-        let group_record = self
+        let mut group_record = self
             .state
             .groups
             .iter()
             .find(|group| group.group_id_hex == group_id_hex)
             .cloned()
             .ok_or_else(|| AppError::UnknownGroup(group_id_hex))?;
+        self.overlay_storage_self_membership(&mut group_record)?;
         let profiles = self.app.profiles_by_id()?;
         let members = self.members_with_profiles_unchecked(group_id, &profiles)?;
         let mls_state = self.group_mls_state_unchecked(group_id)?;
@@ -792,6 +793,19 @@ impl AppClient {
             members,
             mls_state,
         })
+    }
+
+    fn overlay_storage_self_membership(
+        &self,
+        group_record: &mut AppGroupRecord,
+    ) -> Result<(), AppError> {
+        if let Some(membership) = self
+            .app
+            .stored_group_self_membership(&self.state.label, &group_record.group_id_hex)?
+        {
+            group_record.self_membership = membership;
+        }
+        Ok(())
     }
 
     fn group_mls_state_unchecked(&self, group_id: &GroupId) -> Result<AppGroupMlsState, AppError> {
@@ -908,6 +922,7 @@ impl AppClient {
             );
         }
         let profiles = profiles?;
+        let stored_self_memberships = self.app.account_group_self_memberships(&self.state.label)?;
         let mut groups = HashMap::new();
         let mut members = HashMap::new();
         let mut mls_state = HashMap::new();
@@ -918,7 +933,11 @@ impl AppClient {
                 continue;
             };
             let group_id = GroupId::new(bytes);
-            groups.insert(group_id.clone(), group.clone());
+            let mut group_record = group.clone();
+            if let Some(membership) = stored_self_memberships.get(&group.group_id_hex) {
+                group_record.self_membership = *membership;
+            }
+            groups.insert(group_id.clone(), group_record);
             if let Ok(records) = self.members_with_profiles_unchecked(&group_id, &profiles) {
                 members.insert(group_id.clone(), records);
             }
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -141,7 +141,7 @@ fn recover_post_canonical_result<T: Default>(
 }
 
 /// A point-in-time copy of the live session's read-only group projections
-/// (`members`, `group_mls_state`, `quarantined_groups`).
+/// (`groups`, `members`, `group_mls_state`, `quarantined_groups`).
 ///
 /// The account worker captures this from the live session and uses it to answer
 /// read commands while a relay catch-up runs in the background. The catch-up
@@ -157,6 +157,7 @@ fn recover_post_canonical_result<T: Default>(
 /// the same shape the live path returns for a group the session does not hold.
 #[derive(Default)]
 pub(crate) struct GroupReadSnapshot {
+    groups: HashMap<GroupId, AppGroupRecord>,
     members: HashMap<GroupId, Vec<AppGroupMemberRecord>>,
     mls_state: HashMap<GroupId, AppGroupMlsState>,
     quarantined: Vec<AppQuarantinedGroup>,
@@ -178,6 +179,21 @@ impl GroupReadSnapshot {
             .get(group_id)
             .cloned()
             .ok_or_else(|| AppError::UnknownGroup(hex::encode(group_id.as_slice())))
+    }
+
+    pub(crate) fn group_roster(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<crate::groups::AppGroupRosterSession, AppError> {
+        Ok(crate::groups::AppGroupRosterSession {
+            group_record: self
+                .groups
+                .get(group_id)
+                .cloned()
+                .ok_or_else(|| AppError::UnknownGroup(hex::encode(group_id.as_slice())))?,
+            members: self.members(group_id)?,
+            mls_state: self.group_mls_state(group_id)?,
+        })
     }
 
     pub(crate) fn quarantined_groups(&self) -> Vec<AppQuarantinedGroup> {
@@ -756,6 +772,28 @@ impl AppClient {
         self.group_mls_state_unchecked(group_id)
     }
 
+    pub(crate) fn group_roster_session(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<crate::groups::AppGroupRosterSession, AppError> {
+        let group_id_hex = hex::encode(group_id.as_slice());
+        let group_record = self
+            .state
+            .groups
+            .iter()
+            .find(|group| group.group_id_hex == group_id_hex)
+            .cloned()
+            .ok_or_else(|| AppError::UnknownGroup(group_id_hex))?;
+        let profiles = self.app.profiles_by_id()?;
+        let members = self.members_with_profiles_unchecked(group_id, &profiles)?;
+        let mls_state = self.group_mls_state_unchecked(group_id)?;
+        Ok(crate::groups::AppGroupRosterSession {
+            group_record,
+            members,
+            mls_state,
+        })
+    }
+
     fn group_mls_state_unchecked(&self, group_id: &GroupId) -> Result<AppGroupMlsState, AppError> {
         let group = self.runtime.group_record(group_id)?;
         let lifecycle_state = self
@@ -870,6 +908,7 @@ impl AppClient {
             );
         }
         let profiles = profiles?;
+        let mut groups = HashMap::new();
         let mut members = HashMap::new();
         let mut mls_state = HashMap::new();
         let mut skipped_malformed_group_records = 0usize;
@@ -879,6 +918,7 @@ impl AppClient {
                 continue;
             };
             let group_id = GroupId::new(bytes);
+            groups.insert(group_id.clone(), group.clone());
             if let Ok(records) = self.members_with_profiles_unchecked(&group_id, &profiles) {
                 members.insert(group_id.clone(), records);
             }
@@ -895,6 +935,7 @@ impl AppClient {
             );
         }
         Ok(GroupReadSnapshot {
+            groups,
             members,
             mls_state,
             quarantined: self.quarantined_groups(),
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -942,6 +942,14 @@ impl AppClient {
         &self,
         telemetry: Option<&crate::app_telemetry::AppPerformanceTelemetry>,
     ) -> Result<GroupReadSnapshot, AppError> {
+        if cfg!(feature = "test-policy-overrides")
+            && self.app.config.dev_force_group_read_snapshot_failure
+        {
+            return Err(cgka_traits::StorageError::Backend(
+                "injected group-read snapshot failure".to_owned(),
+            )
+            .into());
+        }
         // Load account profiles once and reuse across every group: the rest of
         // the capture is in-memory engine reads, so the snapshot adds a single
         // storage read to the worker readiness path regardless of group count.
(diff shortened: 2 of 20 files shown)


agent-p1p:
## Pip adversarial review — changes required

### Blocking

1. **`crates/marmot-app/src/client/mod.rs:775-789` — `groupRoster` returns stale self-membership for the lifetime of the account worker.**

   `group_roster_session` clones the group record from `self.state.groups`. That copy is not authoritative for `self_membership`: `StoredAccountGroup` explicitly documents that the field is storage-owned and changed only through `SqliteAccountStorage::set_group_self_membership` (`crates/storage-sqlite/src/account_projection.rs:137-142`). The local-leave path writes `Left` through that storage API (`client/mod.rs:1365-1369`) without updating `self.state.groups`; the observed self-departure path does the same for `Left`/`Removed` (`client/sync.rs:1032-1060`), while `refresh_from_group` intentionally preserves the old field (`groups.rs:680-717`). The one-time open backfill also runs after `AppClient.state` was loaded (`lib.rs:1345-1372`).

   Result: a live runtime call to `groupRoster` immediately after local leave or peer eviction can still return `Member` until the account worker is restarted. The startup read snapshot can likewise capture the pre-backfill value. This breaks the main contract of #1297: replacing the eviction probe with an authoritative self-membership/eviction datum.

   Source this value from the authoritative projection inside the serialized roster read, or keep the in-memory group record synchronized at every membership write. Add end-to-end runtime regressions that call `group_roster` after (a) local leave and (b) observed peer eviction without restarting the worker, asserting `Left` and `Removed` respectively. The current constructor/FFI tests inject a prebuilt `Removed` record and therefore cannot catch this path.

### Verification

- `cargo fmt --all -- --check` — pass
- `cargo test -p marmot-app group_roster --locked` — pass
- `cargo test -p marmot-uniffi group_roster --locked` — pass
- `cargo test -p marmot-app --test relay_runtime app_runtime_serves_member_reads_before_initial_catch_up_completes --features test-policy-overrides --locked -- --test-threads=1` — pass
- `cargo clippy -p marmot-app -p marmot-uniffi --all-targets --locked -- -D warnings` — pass
- GitHub CI — all code-owned checks green at reviewed head `f5ff15e2`

**Verdict:** changes required. 1 blocking finding; no suggestions or nitpicks.

agent-p1p:
Addressed the blocking stale-membership finding in c37a24a2. Live `groupRoster` now overlays `self_membership` from the storage-owned projection, and startup snapshot capture performs one authoritative batch read after the open backfill. Added no-restart runtime regressions for local leave (`Left`) and observed admin eviction (`Removed`). Verified `storage-sqlite` 324/324, serial `relay_runtime` 94/94, focused roster tests, and `just fast-ci`.

crates/marmot-app/src/app_telemetry.rs
```diff
@@ -43,6 +43,7 @@ pub(crate) enum AppPerformanceOperation {
     GroupPromoteAdmin,
     GroupDetailsRead,
     GroupMlsStateRead,
+    GroupRosterRead,
```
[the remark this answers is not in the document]
agent-p1p: This is already instrumented at the public runtime boundary in `crates/marmot-app/src/runtime/mod.rs:1207-1213`: `MarmotAppRuntime::group_roster` records `GroupRosterRead` with elapsed time and `result.is_ok()`. `AccountManager::group_roster` is the internal command-RPC implementation; instrumenting both would double-count every public runtime call. I verified this with a scratch runtime regression asserting attempts and successes each increase by exactly one; it passed before any production change. No change needed.

agent-p1p:
## Pip adversarial re-review — blocking finding resolved

Reviewed exact head `c37a24a2a36716849c1ce373861e37c0efd06a8c`.

The original stale-self-membership blocker is resolved:

- Live roster reads clone the in-memory group record, then overlay `self_membership` from the storage-owned `account_groups` projection before assembling the session (`crates/marmot-app/src/client/mod.rs:775-807`; `crates/storage-sqlite/src/account_projection.rs:966-1007`).
- Startup snapshot capture runs after the existing open-time backfill (`crates/marmot-app/src/lib.rs:1321-1373`) and batch-loads the authoritative projection before copying group records into `GroupReadSnapshot` (`crates/marmot-app/src/client/mod.rs:901-949`).
- The write paths remain symmetric: local leave records `Left` (`crates/marmot-app/src/client/mod.rs:1373-1388`), while observed self-eviction records `Removed` before returning from sync (`crates/marmot-app/src/client/sync.rs:1032-1060`). Projection snapshot saves intentionally do not overwrite this storage-owned column.
- The new no-restart runtime regressions exercise both paths through the public runtime API (`crates/marmot-app/tests/relay_runtime.rs:2288-2407`).

### Independent verification

- `cargo fmt --all -- --check` — pass
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked group_roster_reports_ -- --test-threads=1` — pass (2/2)
- `cargo test -p marmot-app group_roster --locked` — pass (roster unit test plus both live regressions)
- `cargo test -p marmot-uniffi group_roster --locked` — pass (1/1)
- `cargo clippy -p marmot-app -p marmot-uniffi -p storage-sqlite --all-targets --locked -- -D warnings` — pass

The exact final diff includes the required `Unreleased` changelog entry and no version/cohort bump. No new blocking findings, suggestions, or nitpicks.

**Verdict:** LGTM at `c37a24a2`; the prior blocking finding is explicitly resolved.

Datawav (commented):

No actionable findings.

_Advisory audit of head [`c37a24a2a367`](https://github.com/marmot-protocol/mdk/commit/c37a24a2a36716849c1ce373861e37c0efd06a8c); no approval, merge, push, checkout, or PR-code execution was performed._

agent-p1p:
## Pip adversarial re-review — changes required after rebase

Reviewed exact head `89b7eda5aebe650fe0e1f3d959c297421b81b323` against base `2e02e0d8bd583c80e0059c1616054efc1c35d42e`.

### Blocking

1. **`crates/marmot-app/src/runtime/account_worker.rs:1306-1311` — an on-demand startup roster can still expose the migration default before the deferred membership backfill runs.**

   Runtime open now deliberately returns with groups seeded but unhydrated. The first `backfill_self_membership_once` call therefore sees `GroupNotHydrated`, leaves legacy `self_membership = 'member'` rows unchanged, and does not burn the marker (`crates/marmot-app/src/lib.rs:1384-1386`; `crates/marmot-app/src/client/mod.rs:1417-1471`). During the background-hydration hold, this arm hydrates the named group and immediately calls `group_roster_session`; that method overlays the still-default storage value. The worker reruns the backfill only after the *entire* pipeline finishes (`account_worker.rs:1252-1261`). On first open after migration, a previously removed account can therefore receive a roster whose member list excludes it while `self_membership` still says `Member`. This restores the stale-membership behavior that the prior fix eliminated at pre-rebase head `c37a24a2`.

   After hydrating the named group, complete/attempt the pending authoritative membership backfill before returning `GroupRoster` (and fail the read rather than knowingly returning an unverified default if that storage step errors). Add a held-hydration upgrade regression with a legacy default-member row whose engine roster excludes the local account; `group_roster` must return `Removed` before the rest of the pipeline is released.

2. **`crates/marmot-app/src/runtime/account_worker.rs:1467-1469` — steady-state/deferred-fallback `GroupRoster` omitted master’s on-demand hydration recovery.**

   Master added `ensure_group_hydrated` to `Members` and `GroupMlsState` (`account_worker.rs:1445-1465`) so groups left behind by a storage-level startup-pipeline abort do not return `GroupHydrationPending` forever. The rebased `GroupRoster` arm calls `group_roster_session` directly. This also breaks the degraded snapshot path: the `None` snapshot branch correctly defers `GroupRoster` at lines 595-603, but replay lands in this handler without promoting the group. A client adopting the new combined call can stay stuck until some unrelated send/ingest path hydrates it.

   Give `GroupRoster` the same on-demand promotion as the two sibling reads and cover the pipeline-abort/deferred-replay path.

### Verification

- `cargo fmt --all -- --check` — pass
- `cargo test -p marmot-app group_roster --locked` — pass (3 focused tests)
- `cargo test -p marmot-app --test startup_hydration --features test-policy-overrides --locked` — pass (2/2; neither exercises `GroupRoster`)
- `cargo clippy -p marmot-app --all-targets --features test-policy-overrides --locked -- -D warnings` — pass
- Rebase range-diff confirms the non-equivalent logic is confined to the new background-hydration integration in `account_worker.rs`; the final diff retains the required `Unreleased` changelog entry and has no version/cohort bump.

**Verdict:** changes required. 2 blocking findings.

crates/marmot-app/src/groups.rs
```diff
@@ -198,6 +198,79 @@ pub struct AppGroupMlsState {
...
+                display_name: display_names.get(&member.member_id_hex).cloned(),
+            }
+        })
+        .collect();
+    let epoch = mls_state.epoch;
+    AppGroupRoster {
+        group_id_hex: group_record.group_id_hex,
+        members,
+        epoch,
+        roster_revision: epoch,
```
Datawav: **Medium:** `roster_revision` does not change for every roster-visible update. `group_roster_session` overlays storage-owned `self_membership`, so a local leave or observed eviction can change the returned value from `Member` to `Left`/`Removed` while the MLS epoch—and therefore this revision—stays unchanged. Consumers using this field for membership-screen invalidation can miss the update. Incorporate storage-owned membership changes into a monotonic revision, or narrow the API contract so this is explicitly an MLS epoch rather than a roster revision.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -586,6 +592,16 @@ async fn run_app_runtime_account_worker(
                                 ))),
                             }
                         }
+                        Some(AccountWorkerCommand::GroupRoster { group_id, respond }) => {
+                            match &read_snapshot {
+                                Some(snapshot) => {
+                                    let _ = respond.send(snapshot.group_roster(&group_id));
```
Datawav: **Low:** During initial relay catch-up this serves the immutable startup snapshot. If catch-up subsequently observes the caller's eviction and updates storage, repeated roster reads can still report `Member` until catch-up finishes and the worker switches to live reads. Refresh or replace the snapshot's storage-owned self-membership as catch-up processes changes, or explicitly document that startup roster reads are bounded-stale.

crates/marmot-app/src/client/mod.rs
```diff
@@ -180,6 +181,21 @@ impl GroupReadSnapshot {
...
+        &self,
+        group_id: &GroupId,
+    ) -> Result<crate::groups::AppGroupRosterSession, AppError> {
+        Ok(crate::groups::AppGroupRosterSession {
+            group_record: self
+                .groups
+                .get(group_id)
+                .cloned()
+                .ok_or_else(|| AppError::UnknownGroup(hex::encode(group_id.as_slice())))?,
+            members: self.members(group_id)?,
```
Datawav, line 195: **Low:** Snapshot capture inserts a group even when its member or MLS projection read fails, but these helpers translate a missing map entry into `UnknownGroup`. A known group with a failed snapshot projection is therefore misreported as nonexistent, unlike the live path which propagates the underlying failure. Preserve per-group projection errors, build each roster snapshot atomically as a `Result`, or return a dedicated snapshot/hydration error.

Datawav (commented):

## Final audit

### 1. Medium — `roster_revision` misses roster-visible self-membership changes

`app_group_roster_from_session` always sets `roster_revision` equal to the MLS epoch (`crates/marmot-app/src/groups.rs`). However, live reads independently overlay `self_membership` from SQLite in `group_roster_session`, specifically because local leave and observed eviction can update storage while the in-memory group record lags (`crates/marmot-app/src/client/mod.rs`, `crates/marmot-app/src/lib.rs`).

A local leave can therefore change the returned roster from `Member` to `Left` without changing `roster_revision`. Consumers relying on the advertised revision for membership-screen change detection can miss that visible state change. The leave test verifies the new membership value but does not assert a revision change.

Need: consumers need a revision that changes for every field they are expected to invalidate from it.

Design/alternatives: incorporate storage-owned membership state into a monotonic roster revision; maintain a dedicated projection revision; or narrow the API contract and field name so it explicitly covers MLS-epoch changes only. Merely documenting display-name exclusions does not address the self-membership false negative.

### 2. Low — catch-up roster reads can return stale eviction state

`GroupReadSnapshot` overlays all stored self-memberships only when the snapshot is captured. During initial relay catch-up, `GroupRoster` commands are answered directly from that immutable snapshot (`crates/marmot-app/src/runtime/account_worker.rs`). If catch-up subsequently observes the caller’s eviction and updates storage, roster reads can continue returning `Member` until catch-up completes and reads switch to the live session.

This does not require a worker restart, contrary to one cross-review’s wording, but it qualifies the otherwise broad “observed eviction” behavior.

Need: callers should know whether the combined projection is current or bounded-stale during startup.

Design/alternatives: refresh the storage-owned self-membership when serving a snapshot roster, replace the snapshot entry when catch-up processes relevant changes, or explicitly document the catch-up consistency boundary.

### 3. Low — snapshot projection failures are misreported as `UnknownGroup`

Snapshot capture inserts a valid group record into `groups`, but inserts `members` and `mls_state` only when their individual reads succeed (`crates/marmot-app/src/client/mod.rs`). `GroupReadSnapshot::group_roster` then accesses each map through helpers that return `AppError::UnknownGroup` when an entry is absent. Consequently, a known group whose member or MLS projection failed during capture is reported as nonexistent, unlike the live path, which propagates the underlying error.

Need: callers and diagnostics must distinguish missing groups from degraded snapshot construction.

Design/alternatives: retain per-group projection errors in the snapshot and replay them, omit the entire group consistently while returning a dedicated snapshot/hydration error, or build roster entries atomically as one `Result<AppGroupRosterSession, AppError>`.

_Advisory audit of head [`89b7eda5aebe`](https://github.com/marmot-protocol/mdk/commit/89b7eda5aebe650fe0e1f3d959c297421b81b323); no approval, merge, push, checkout, or PR-code execution was performed._

erskingardner (commented):
Takeover review complete at `6275b04e` after rebasing onto current `master`.

All five feedback items are addressed: startup and steady-state roster reads share on-demand hydration plus authoritative self-membership reconciliation; catch-up snapshot reads refresh storage-owned membership; degraded snapshot construction now fails atomically; roster revision covers MLS epoch plus caller membership; and focused regressions cover held hydration plus live leave/eviction invalidation. I also added the required workspace version bump for the new UniFFI surface.

No further actionable findings in the final rebased diff. Local validation: `just fast-ci`; 324 storage tests; 100 UniFFI unit + 2 fixture + 28 smoke tests; 586 serialized marmot-app library tests; focused roster and startup-hydration integrations.

erskingardner:
CI follow-up: the required workspace bump to 0.9.11 exposed stale `conformance_version` fields in the canonical simulator fixtures. Refreshed all 27 fixtures in `a804609f`; the focused fixture test and full `just simulator-smoke` gate now pass locally (314/314).

erskingardner:
Final rebase follow-up: rebased onto current `master` (`6f40bde1`, including #1309). The new steady-state detached catch-up path introduced upstream now serves `GroupRoster` from the same snapshot as members/MLS state, while refreshing stored self-membership (`9554e5d5`). Extended #1309’s blocked-catch-up regression to cover the roster read. Local `just fast-ci`, all 588 marmot-app lib tests, focused roster tests, and held startup-hydration regression pass.

erskingardner:
Version-policy correction at `4da2de09`: removed the workspace `0.9.11` bump from branch history and dropped the conformance-fixture refresh commit. The workspace, lockfile packages, and all vector `conformance_version` values remain `0.9.10`. Updated root and UniFFI `AGENTS.md` guidance so agents never bump workspace or fixture versions during feature, fix, binding, or review-feedback work unless explicitly requested as a manual release operation. The canonical fixture test and `just fast-ci` pass at `0.9.10`.

Datawav:

## Follow-up audit

Previously requested changes touched by this delta:

- Fixed: `roster_revision` now changes for the reviewed `Member → Left/Removed` transitions.
- Fixed: snapshot-served roster reads refresh storage-owned self-membership, preventing stale leave/eviction state during catch-up.
- Fixed: snapshot construction now fails atomically instead of misreporting degraded known groups as `UnknownGroup`.

### New: Medium — snapshot failure prevents explicit catch-up

`handle_account_worker_catch_up` in `crates/marmot-app/src/runtime/account_worker.rs` returns before calling `client.sync()` when `group_read_snapshot()` fails. Meanwhile, `group_read_snapshot_inner` in `crates/marmot-app/src/client/mod.rs` now propagates any known group’s member or MLS projection error with `?`.

Consequently, one degraded group can prevent an explicit catch-up from retrieving updates that may advance or repair account state. Periodic maintenance can still sync directly, but the requested catch-up itself fails without attempting synchronization.

Snapshot availability should control concurrent read serving, not whether synchronization runs. On capture failure, continue `client.sync()` while deferring reads until live state becomes available, matching the startup hydration behavior.

_Advisory audit of head [`4da2de0949bb`](https://github.com/marmot-protocol/mdk/commit/4da2de0949bbba64dd14b053998a09a20819243d); no approval, merge, push, checkout, or PR-code execution was performed._

PR #1313, opened 2026-08-07
Make generated scenarios selectable across adapters

erskingardner:
## What changed

- add a shared resolver for raw canonical Scenario IR and saved `GeneratedScenarioInputV1` envelopes
- preserve the generated family, generator version, seed, case index, recorded subject, semantic expectations, source digest, and selected canonical-IR digest
- allow the report runner to replay a saved generated case through the engine, retained-relay, or app-runtime adapter
- give adapter overrides distinct report/capsule filenames and prevent override runs from silently minting adapter-dependent fixture candidates
- let the process orchestrator CLI consume the same saved generated input directly
- carry selected-input provenance into process/container reports and separately record the digest of the IR the process executor actually compiled
- let container campaigns resolve the selected scenario in memory, with deterministic manifest-declared lowering for host-crash campaigns
- privately materialize `canonical-scenario.json` only for VM drivers, so external drivers do not need a second generated-envelope parser
- preserve the existing raw-file digest identity used by indexed node-capsule fingerprints
- update the convergence tracking plan and adapter documentation

## Why

The stateful generator already produces deterministic, replayable journeys, but those saved journeys were tied to the simulator subject recorded by the generator. Process, container, and VM runners expected bare Scenario IR, which made it easy for adapter-specific launch paths to lose generator provenance, expectations, or execute a differently identified scenario.

This change establishes one selection boundary. Every adapter resolves the same saved source input and records its selected canonical Scenario IR. If an adapter deterministically lowers that history—currently container host-crash campaigns—the process report records a distinct executed-IR digest.

## Impact and scope

This is simulator and campaign-runner infrastructure only. It does not change production convergence-engine behavior or production Marmot app behavior.

Capability preflight remains authoritative. Selecting an adapter does not silently lower or skip unsupported operations. Some current stateful journeys still contain engine-only exact-state or observation steps and will be rejected by app/process capability contracts; the tracking-plan item remains open until representative journeys satisfy every adapter's capability and oracle contracts.

Process reports carry semantic expectations as provenance for downstream oracle tooling; the process executor does not yet evaluate those expectations itself.

## Validation

- `just fast-ci`
- `cargo test -p cgka-conformance-simulator --lib` — 133 passed
- adapter A/B artifact isolation and no override fixture minting
- saved generated input through the real app-runtime harness
- saved generated input through real child-process nodes
- `cargo test -p convergence-campaign-runner`
- raw Scenario IR node-capsule fingerprint continuity
- selected-vs-executed digest distinction for host-crash lowering
- generated-input VM materialization and normalized-manifest regression
- `git diff --check`

The real OCI container test remains intentionally ignored because it requires a prebuilt campaign image and Docker/Podman daemon.

## Summary by CodeRabbit

* **New Features**
  * Replay saved generated scenarios with preserved source, canonical scenario, adapter, and expectation metadata.
  * Execute generated scenarios through alternate adapters, including app-runtime and process execution.
  * Materialize canonical scenario files for VM-based runs.
  * Record source and canonical scenario digests in reports and campaign manifests.
  * Support selecting execution subjects and prevent adapter overrides from creating fixture candidates.

* **Bug Fixes**
  * Reject unsupported generated-input versions and invalid canonical digest values.
  * Ensure equivalent scenario formatting produces consistent canonical digests.

* **Documentation**
  * Updated simulator and campaign documentation with scenario-input, provenance, digest, and VM execution behavior.

### DIFF
--- a/crates/cgka-conformance-simulator/src/family.rs
+++ b/crates/cgka-conformance-simulator/src/family.rs
@@ -83,9 +83,17 @@ impl GeneratedSubjectKind {
     pub fn parse(value: &str) -> Result<Self, String> {
         match value {
             "engine" => Ok(Self::Engine),
-            "retained-relay" | "retained_relay" => Ok(Self::RetainedRelay),
-            "app-runtime" | "app_runtime" => Ok(Self::AppRuntime),
+            "retained-relay" => Ok(Self::RetainedRelay),
+            "app-runtime" => Ok(Self::AppRuntime),
             other => Err(format!("unsupported generated-case adapter {other}")),
+        }
+    }
+
+    pub(crate) fn artifact_label(self) -> &'static str {
+        match self {
+            Self::Engine => "engine",
+            Self::RetainedRelay => "retained-relay",
+            Self::AppRuntime => "app-runtime",
         }
     }
 }
--- a/crates/cgka-conformance-simulator/src/family.rs
+++ b/crates/cgka-conformance-simulator/src/family.rs
@@ -5,11 +5,11 @@
 
 use crate::scenario::subject_setup_error;
 use crate::{
-    GeneratedScenarioMetadata, HarnessStorageMode, RetainedRelaySubject, ScenarioFailureCaptureV1,
-    ScenarioMessageSelectorV2, ScenarioOutboundSelection, ScenarioReport, ScenarioRunError,
-    ScenarioSpec, ScenarioStep, ScenarioTrace, ScenarioTransportClass, SubjectFailureCategory,
-    SubjectOutboundOutcome, TraceExpectation, VectorFixture, fingerprint_report_failure,
-    run_scenario_report_with_outcomes_and_capture,
+    AppRuntimeHarness, GeneratedScenarioMetadata, HarnessStorageMode, RetainedRelaySubject,
+    ScenarioFailureCaptureV1, ScenarioMessageSelectorV2, ScenarioOutboundSelection, ScenarioReport,
+    ScenarioRunError, ScenarioSpec, ScenarioStep, ScenarioTrace, ScenarioTransportClass,
+    SubjectFailureCategory, SubjectOutboundOutcome, TraceExpectation, VectorFixture,
+    fingerprint_report_failure, run_scenario_report_with_outcomes_and_capture,
     run_scenario_report_with_outcomes_and_storage_mode, run_scenario_report_with_subject,
     stable_action_id,
 };
@@ -72,11 +72,21 @@ pub enum GeneratedSubjectKind {
     #[default]
     Engine,
     RetainedRelay,
+    AppRuntime,
 }
 
 impl GeneratedSubjectKind {
     fn is_engine(&self) -> bool {
         *self == Self::Engine
+    }
+
+    pub fn parse(value: &str) -> Result<Self, String> {
+        match value {
+            "engine" => Ok(Self::Engine),
+            "retained-relay" | "retained_relay" => Ok(Self::RetainedRelay),
+            "app-runtime" | "app_runtime" => Ok(Self::AppRuntime),
+            other => Err(format!("unsupported generated-case adapter {other}")),
+        }
     }
 }
 
@@ -279,7 +289,14 @@ pub async fn run_generated_case_report_with_storage_mode(
         storage_mode,
     )
     .await?;
-    add_generated_metadata(case, expected_trace.as_ref(), &mut report, storage_mode).await;
+    add_generated_metadata(
+        case,
+        case.subject,
+        expected_trace.as_ref(),
+        &mut report,
+        storage_mode,
+    )
+    .await;
     Ok(report)
 }
 
@@ -291,7 +308,24 @@ pub async fn run_generated_case_report_with_capture(
     storage_mode: HarnessStorageMode,
     capture_sensitive_replay: bool,
 ) -> Result<(ScenarioReport, crate::ScenarioFailureCaptureV1), ScenarioRunError> {
-    let (mut report, failure_capture) = match case.subject {
+    run_generated_case_report_with_capture_on_subject(
+        case,
+        case.subject,
+        expected_trace,
+        storage_mode,
+        capture_sensitive_replay,
+    )
+    .await
+}
+
+pub async fn run_generated_case_report_with_capture_on_subject(
+    case: &GeneratedScenarioCase,
+    execution_subject: GeneratedSubjectKind,
+    expected_trace: Option<ScenarioTrace>,
+    storage_mode: HarnessStorageMode,
+    capture_sensitive_replay: bool,
+) -> Result<(ScenarioReport, crate::ScenarioFailureCaptureV1), ScenarioRunError> {
+    let (mut report, failure_capture) = match execution_subject {
         GeneratedSubjectKind::Engine => {
             run_scenario_report_with_outcomes_and_capture(
                 &case.scenario,
@@ -334,19 +368,63 @@ pub async fn run_generated_case_report_with_capture(
                 },
             )
         }
+        GeneratedSubjectKind::AppRuntime => {
+            if capture_sensitive_replay {
+                return Err(ScenarioRunError {
+                    step_index: None,
+                    kind: "sensitive_replay_capture_unsupported".into(),
+                    category: SubjectFailureCategory::Environment,
+                    message: "app-runtime generated cases cannot capture an engine checkpoint"
+                        .into(),
+                });
+            }
+            let mut subject = AppRuntimeHarness::new(&case.scenario.clients)
+                .await
+                .map_err(subject_setup_error)?;
+            let result = run_scenario_report_with_subject(
+                &case.scenario,
+                expected_trace.clone(),
+                case.expected_outcomes.clone(),
+                &mut subject,
+            )
+            .await;
+            subject.shutdown().await;
+            (
+                result?,
+                ScenarioFailureCaptureV1 {
+                    transport: Default::default(),
+                    byte_replay: None,
+                },
+            )
+        }
     };
-    add_generated_metadata(case, expected_trace.as_ref(), &mut report, storage_mode).await;
+    add_generated_metadata(
+        case,
+        execution_subject,
+        expected_trace.as_ref(),
+        &mut report,
+        storage_mode,
+    )
+    .await;
     Ok((report, failure_capture))
 }
 
 async fn add_generated_metadata(
     case: &GeneratedScenarioCase,
+    execution_subject: GeneratedSubjectKind,
     expected_trace: Option<&ScenarioTrace>,
     report: &mut ScenarioReport,
     storage_mode: HarnessStorageMode,
 ) {
     let minimized_case = if fingerprint_report_failure(report).is_ok() {
-        minimize_failing_case(case, expected_trace, report, storage_mode).await
+        minimize_failing_case(
+            case,
+            execution_subject,
+            expected_trace,
+            report,
+            storage_mode,
+        )
+        .await
     } else {
         None
     };
@@ -361,6 +439,7 @@ async fn add_generated_metadata(
 
 async fn minimize_failing_case(
     case: &GeneratedScenarioCase,
+    execution_subject: GeneratedSubjectKind,
     expected_trace: Option<&ScenarioTrace>,
     failing_report: &ScenarioReport,
     storage_mode: HarnessStorageMode,
@@ -379,7 +458,7 @@ async fn minimize_failing_case(
                 expected_outcomes,
                 &target_identity,
                 storage_mode,
-                case.subject,
+                execution_subject,
             )
             .await
         }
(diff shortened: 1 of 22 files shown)


crates/cgka-conformance-simulator/src/report.rs
```diff
@@ -314,17 +329,29 @@ async fn run_generated_input_reports(
...
+    let (mut report, failure_capture) = crate::run_generated_case_report_with_capture_on_subject(
+        case,
+        source.adapter.unwrap_or(case.subject),
+        None,
+        storage_mode,
+        capture_sensitive_replay,
+    )
+    .await?;
+    report.metadata.input_provenance = Some(source.provenance);
     let output = out.join(format!(
```
erskingardner: The report, fixture, and both failure-capsule filenames are keyed on `family_name` / `seed` / `case_index` only. That was unambiguous before, because a saved case had exactly one subject. With `--adapter` it no longer is:

```sh
... --generated-input case.json --adapter engine     --out reports/
... --generated-input case.json --adapter app-runtime --out reports/
```

The second invocation silently overwrites the first run's report, fixture candidate, and capsules. Comparing adapters on one saved journey is the main reason to reach for this flag, so losing the first result without a word is a poor default.

Suggest folding the executing subject into the stem when it differs from `case.subject` (or unconditionally), e.g. `...-case-{n}-{adapter}.json`. `source.adapter.unwrap_or(case.subject)` is already computed a few lines above.

crates/cgka-conformance-simulator/src/report.rs
```diff
@@ -340,7 +367,7 @@ async fn run_generated_case_artifacts(
     ));
     let fixture = generated_fixture_candidate(case, &report);
```
erskingardner: `generated_fixture_candidate` sets `expected_trace` from `report.observed_trace` whenever `case.expected_outcomes` is empty, and `VectorFixture` carries no record of which adapter produced it. So after this PR the promotable vector candidate for a given saved case depends on an invisible `--adapter` choice: run it through `app-runtime` and the fixture bakes in the app-runtime-observed trace, labelled identically to the engine-observed one.

The cross-adapter equivalence test argues these traces *should* match, but a promotable conformance vector is precisely where you don't want to assume that. Two options, either is fine:

- skip writing the fixture candidate when `source.adapter` overrides `case.subject` (an override run is a replay experiment, not a vector source), or
- record the executing adapter on the fixture so promotion is reviewable.

As written, the artifact silently changes meaning based on a flag it doesn't record.

crates/cgka-conformance-simulator/src/report.rs
```diff
@@ -296,13 +306,18 @@ async fn run_generated_input_reports(
     }
     let mut summaries = Vec::with_capacity(paths.len());
     for path in paths {
-        let input: GeneratedScenarioInputV1 =
-            serde_json::from_str(&std::fs::read_to_string(path)?)?;
+        let input_bytes = std::fs::read(path)?;
+        let resolved = resolve_scenario_input_bytes(&input_bytes)?;
+        let input: GeneratedScenarioInputV1 = serde_json::from_slice(&input_bytes)?;
```
erskingardner: `resolve_scenario_input_bytes` on the previous line already parsed this envelope and already called `input.validate()` internally (`generated_scenario_input_version`). This re-parses the whole case a second time and re-validates, then discards `resolved.scenario` and `resolved.expected_outcomes` in favour of `input.case`.

On a stateful journey the case is the large part of the payload, so this is a full redundant deserialization per input. Having `ResolvedScenarioInputV1` carry the `GeneratedScenarioCase` (e.g. `generated: Option<GeneratedScenarioCase>`) would delete these two lines and let `run_generated_case_artifacts` take the case straight from the resolver — the resolver becomes the single parse boundary the module doc claims it is.

Same shape at line 273 in `run_generated_family_reports`: `serde_json::to_vec_pretty` followed by re-parsing your own freshly serialized bytes just to obtain provenance. There the source digest genuinely must be over the written bytes, but the rest of the provenance is available directly from `case` plus `canonical_scenario_ir_sha256(&case.scenario)`.

erskingardner (commented):
## Overall

The premise holds up: one resolver, one canonical-IR digest, and every adapter reading the same envelope is a genuine simplification over four ad hoc `serde_json::from_slice::<ScenarioSpec>` call sites. Scope is honest (simulator/campaign infrastructure only), the digest is stable in practice (`ScenarioSpec` has no maps or floats, so serde's field order is deterministic), the format sniff on `schema_version` + `case` can't collide with raw IR, and the negative test that an unknown envelope version does *not* silently fall back to raw IR is exactly the test I'd have asked for. The tracking-plan note is refreshingly non-triumphal about the item still being open.

Not blocking, but two things in `report.rs` undercut the new `--adapter` flag's main use case, and a couple of documentation claims are stronger than the code:

1. **Report/fixture filenames don't include the executing adapter**, so running the same saved input under two adapters into one `--out` silently overwrites the first run — the exact A/B comparison the flag exists to enable.
2. **The promotable vector candidate is minted from whichever adapter ran**, and `VectorFixture` records no adapter, so the fixture bytes for a case now depend on an invisible CLI flag.
3. **Container host-crash campaigns don't execute the pinned canonical IR** — `lower_container_host_faults` rewrites the steps after resolution, so `canonical_ir_sha256` and `canonical-scenario.json` describe the pre-lowering scenario. Pre-existing, but the new comments and README now assert otherwise.
4. `ProcessScenarioReportV1.expected_outcomes` is carried but never evaluated.
5. A small redundant re-parse in `run_generated_input_reports`.

Details inline. CI is green on the finished jobs.

crates/convergence-campaign-runner/src/runner.rs
```diff
@@ -78,24 +81,41 @@ pub async fn run_manifest(
...
+        Some(scenario_input.provenance.canonical_ir_sha256.clone());
+    let plan = build_execution_plan(&normalized_manifest)?;
+    fs_private::create_dir_all_private(&normalized_manifest.output_dir)
         .map_err(|error| RunnerError::environment("output_directory", error))?;
     fs_private::write_private(
-        &manifest.output_dir.join("normalized-manifest.json"),
-        &serde_json::to_vec_pretty(manifest)
+        &normalized_manifest
+            .output_dir
+            .join("canonical-scenario.json"),
```
erskingardner: For the container backend this file is written but never read: `run_container_scenario` passes `lower_container_host_faults(manifest, &input.scenario)` to the orchestrator, not this materialized copy. When the manifest contains a `CrashParticipantHost` fault, the lowered scenario has extra `CrashProcess` / `RestartProcess` steps, so the IR the nodes actually execute is *not* the IR pinned by `canonical_ir_sha256` and not the IR in `canonical-scenario.json`.

The divergence itself is deliberate and deterministic from the normalized manifest, so I don't think it needs to change. But three new claims now overstate it:

- `scenario_input.rs` module doc: "the canonical-IR digest identifies the exact `ScenarioSpec` every adapter must execute after resolving the envelope"
- `convergence-campaign-runner/README.md`: "container nodes and external VM drivers therefore execute the same scenario"
- `distributed-convergence-campaigns.md`: "Container orchestration uses that resolved value directly"

Worth narrowing these to say the canonical digest pins the *selected* history and that container host-crash lowering is applied on top, deterministically. Alternatively, only materialize `canonical-scenario.json` for the VM backend, which is the only consumer.

crates/cgka-conformance-simulator/src/process_orchestrator.rs
```diff
@@ -108,6 +109,10 @@ fn render_node_arg(
 pub struct ProcessScenarioReportV1 {
     pub schema_version: String,
     pub scenario_name: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub input_provenance: Option<ScenarioInputProvenanceV1>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub expected_outcomes: Vec<TraceExpectation>,
```
erskingardner, line 121: These are copied into the report and never evaluated — nothing in `process_orchestrator.rs` or `convergence-campaign-runner` reads `ProcessScenarioReportV1.expected_outcomes`. That is a defensible choice (the process adapter's oracle is exact-state, and preflight is authoritative), but "the process report preserves the semantic expectations" reads like they're being checked.

Either wire them into the process oracle or add a one-line comment here saying they are carried as provenance for downstream tooling only. Right now a future reader has to grep the workspace to find out which it is.

crates/cgka-conformance-simulator/src/family.rs
```diff
@@ -72,12 +72,22 @@ pub enum GeneratedSubjectKind {
...
     RetainedRelay,
+    AppRuntime,
 }
 
 impl GeneratedSubjectKind {
     fn is_engine(&self) -> bool {
         *self == Self::Engine
     }
+
+    pub fn parse(value: &str) -> Result<Self, String> {
```
erskingardner, line 83: Minor: accepting both `retained-relay` and `retained_relay` (and both `app-runtime` and `app_runtime`) is flexibility nobody asked for. The usage string, the README, and every test use the hyphenated form. Two spellings per variant is two things to keep in sync and one more way for a manifest to be inconsistent with a CLI invocation. Suggest keeping only the hyphenated spellings.

crates/convergence-campaign-runner/src/main.rs
```diff
@@ -226,17 +226,16 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
...
-            let scenario = serde_json::from_slice(&scenario_bytes)?;
+            let scenario =
+                cgka_conformance_simulator::resolve_scenario_input_bytes(&scenario_bytes)?;
             let observation = convergence_campaign_runner::observation_from_node_capsule(
                 &node_capsule,
                 capsule,
-                scenario,
-                &hex::encode(sha2::Sha256::digest(&scenario_bytes)),
+                scenario.scenario,
+                &scenario.provenance.canonical_ir_sha256,
```
erskingardner: `observation_from_node_capsule` fingerprints include this digest. On master it was `sha256(scenario_file_bytes)`. This PR switches it to `canonical_ir_sha256`, which is `sha256(serde_json::to_vec(parsed))`.

For any previously indexed **pretty-printed** raw Scenario IR (the common on-disk form), re-indexing the same capsule+file now yields a **different** fingerprint. Existing corpus classifications/diagnosis marks keyed on the old fingerprint are orphaned, and the same logical failure shows up as a new entry.

Accepting generated envelopes here is good; changing the digest identity for raw IR is a silent corpus break. Prefer fingerprinting `provenance.source_sha256` to preserve continuity (generated envelopes still work, and remain distinct from raw IR), or keep canonical deliberately but call out the migration and add a regression that pretty vs compact raw IR no longer splits fingerprints only after an explicit corpus note. Right now nothing documents or tests the change.

crates/convergence-campaign-runner/src/runner.rs
```diff
@@ -387,10 +427,15 @@ async fn run_container_scenario(
         .into_iter()
         .map(|label| (label, format!("ws://{NODE_RELAY_PROXY_LISTEN}")))
         .collect::<BTreeMap<_, _>>();
-    let mut orchestrator = ProcessOrchestrator::launch_with(
+    let resolved = ResolvedScenarioInputV1 {
+        scenario,
+        expected_outcomes: input.expected_outcomes.clone(),
+        provenance: input.provenance.clone(),
```
erskingardner, line 438: After `lower_container_host_faults`, `scenario` may contain injected `CrashProcess` / `RestartProcess` steps, but `provenance` is still the pre-lowering selected-input provenance (`canonical_ir_sha256` of the pre-lowered IR).

That value is then copied into `ProcessScenarioReportV1.input_provenance`. Consumers of the new provenance field will conclude the report executed IR digest X when the compiled/executed history was X+lowering.

Either:
- treat provenance strictly as *selected-input* identity and stop describing it as the executed ScenarioSpec (module docs / README / architecture doc currently claim execution identity), or
- attach a distinct executed-IR digest after lowering (and keep the manifest pin on the selected history).

As written, the new integrity field is wrong precisely on the host-crash campaigns this runner exists to exercise.

erskingardner (commented):
## Overall

Premise is sound: one resolver, dual digests (source vs canonical IR), and shared consumption across report/process/container/VM is the right selection boundary. Scope stays in simulator/campaign infrastructure, capability preflight remains authoritative, and the unknown-envelope-version negative test is exactly the regression that matters.

I would not merge yet. Two new integrity issues below, and I agree with the earlier self-review that adapter A/B output overwrite and adapter-blind fixture candidates undercut the new `--adapter` flag.

## Should fix before merge

1. **`index-node-capsule` fingerprint migration** — see inline on `main.rs`. Switching the digest from raw file bytes to `canonical_ir_sha256` orphans existing corpus entries for pretty-printed raw IR.
2. **Post-lowering provenance lie on container host-crash campaigns** — see inline on `runner.rs`. Selected-IR provenance is attached to a rewritten scenario.
3. **`--adapter` artifact identity** (already noted): report/fixture/capsule filenames omit the executing adapter, and promotable fixtures bake an adapter-specific observed trace without recording which adapter produced them.

## Non-blocking

- `ProcessScenarioReportV1.expected_outcomes` is carried, never evaluated — fine as scaffolding if docs do not imply oracle coverage.
- Redundant re-parse in `run_generated_input_reports` after `resolve_scenario_input_bytes`.
- Snake_case CLI aliases for adapter names are unused flexibility.

CI looks green on finished jobs; review is against PR head `a9e84ece7916fd459796b2a4717e81754739be8c` only (no local checkout).

Issue #859, opened 2026-07-14
Media sends (image/document/video/voice) cannot be threaded as replies — gap spans marmot-app, agent-control protocol, and the Hermes adapter


Datawav:
## Summary
Sending a media attachment (image/document/video/voice) as a reply to a specific message is not supported anywhere in the stack — not just the Hermes adapter. The gap goes all the way down to the `marmot-app` client layer, which has no reply-aware media send path at all.

## What's confirmed in the code

**Hermes adapter** (`integrations/hermes/marmot/adapter.py:1728-1742`, `_send_media_upload`):
```python
async def _send_media_upload(
    self, chat_id: str, *, path: str, media_type: str, file_name: str,
    caption: Optional[str] = None, reply_to: Optional[str] = None,
) -> SendResult:
    if reply_to:
        return SendResult(
            success=False,
            error="Marmot media sends do not support reply threading yet",
        )
    ...
```
Every media-send entry point (`send_image_file`, `send_document`, `send_video`, `send_voice`) routes through this, so passing `reply_to` to any of them is a guaranteed hard failure (covered by `tests/test_adapter.py:2845-2852`, which asserts the "reply threading" error string).

**`marmot-app` client** (`crates/marmot-app/src/client/mod.rs`):
- `reply_to_message()` (line 1289) sends a *text* reply via `AppMessageIntent::Reply { target_message_id, text }`.
- `send_media_attachments()` (line 1307) has no `target_message_id`/reply parameter at all, and does not construct `AppMessageIntent::Reply` — it's a plain, non-threaded send.

So there's currently no `AppMessageIntent` variant that carries both a media attachment *and* a reply target — this needs a small app-layer addition before the Hermes adapter (or any other integration) can support it.

**`agent-control` protocol** (`crates/agent-control/src/lib.rs:201-206`):
```rust
SendMedia {
    account_id_hex: String,
    group_id_hex: String,
    attachments: Vec<AgentControlMediaUpload>,
    caption: Option<String>,
},
```
No `reply_to_message_id_hex` field either — would need adding here too, mirroring `SendAgentActivity`'s existing `reply_to_message_id_hex: Option<String>` field.

## Scope
This is a three-layer change, not a single-file fix:
1. `marmot-app`: add a reply-aware media intent (or extend `AppMessageIntent` for a media+reply variant) and wire it through `send_media_attachments` (or a new sibling method).
[...]

PR #1323, opened 2026-08-07
feat(hermes): send multi-image responses as one Marmot album

### LINKED ISSUE #1320: Send Hermes multi-image batches as one Marmot message
## Summary

Wire Hermes `send_multiple_images` to one Marmot message containing multiple encrypted attachments. MDK and `agent-control` already support multiple attachments in one send, but the Hermes adapter falls back to sequential messages. Telegram calls this an album/media group; Discord, Slack, Signal, and WhatsApp adapters also preserve a batch more coherently than unrelated sends.

This is a connector parity issue with a relatively small implementation surface, not a new Marmot wire feature.

## Confirmed gap

- `AgentControlRequest::SendMedia` already carries `attachments: Vec<AgentControlMediaUpload>` ([agent-control](https://github.com/marmot-protocol/mdk/blob/9396adb6aa6b95b521a7979facd5ea7040c07288/crates/agent-control/src/lib.rs#L265-L273)).
- Timeline messages already expose `attachments: Vec<AgentControlAttachmentSummary>` ([timeline DTO](https://github.com/marmot-protocol/mdk/blob/9396adb6aa6b95b521a7979facd5ea7040c07288/crates/agent-control/src/lib.rs#L499-L527)).
- The Marmot adapter overrides single image/document/video/voice sends but does not override Hermes `send_multiple_images`; the base fallback sends items individually.
- The standalone Marmot sender likewise loops over `media_files` and calls `_send_media_upload` once per path ([adapter](https://github.com/marmot-protocol/mdk/blob/9396adb6aa6b95b521a7979facd5ea7040c07288/integrations/hermes/marmot/adapter.py#L3390-L3410)).
- Telegram `sendMediaGroup` sends 2–10 items as one album. Hermes's Discord, Slack, and Signal connectors also implement `send_multiple_images`.

## Implementation

1. Add a Marmot `send_multiple_images` override that validates every local file first, builds one bounded attachments vector, and performs one `SendMedia` request.
2. Apply the caption once to the message, not once per attachment. Preserve stable attachment order and filenames/media types.
3. Make the operation all-or-error from the Hermes caller's perspective. If uploads can partially succeed before message publication, clean up or leave only bounded encrypted blobs; never publish a partial album while reporting full success.
4. Reuse the existing allowed-root, regular-file, size/count, timeout, and privacy-safe error rules.
5. Update standalone `media_files` delivery to use the same batch path.
6. Return all durable message ids/attachment outcomes in a form Hermes can audit without leaking local paths or content keys.

## Acceptance

- Two through the supported maximum images arrive as one Marmot timeline message, in order, with one caption.
- Mixed valid/invalid paths fail before publication; retry does not duplicate the message.
- Reply-aware albums remain blocked on #859 unless that issue lands first; once it lands, the batch path carries the one reply target.
- Tests cover count/byte boundaries, mixed MIME types, duplicate paths, partial upload failure, cancellation, timeout retry/idempotency, and inbound round-trip projection.

## Duplicate check

Searched open and closed MDK issues for multiple images, album, and media group; no duplicate exists. #859 concerns media reply targets and #861 concerns remote image/GIF URLs.

agent-p1p:
## Summary

- override Hermes `send_multiple_images(...)` so ordered local images are preflighted, privately staged, and sent in one Marmot `SendMedia` request
- bound albums to 10 attachments, the encrypted-blob per-file limit, and a 128 MiB aggregate; clean all staged files on success, failure, timeout, or cancellation
- add persisted `SendMedia` idempotency keyed to destination, caption, ordered metadata, and plaintext hashes so timeout retries return the original durable message ids
- return safe attachment outcomes without local paths or content keys; keep reply-targeted albums fail-closed until #859

## Verification

- `python3 -m unittest discover -s integrations/hermes/marmot/tests -q` (179 passed)
- `cargo test -p agent-control` (17 passed)
- `cargo test -p agent-connector` (135 unit + 1 integration passed)

## Baseline note

The pre-existing `marmot-app` E2E test `encrypted_media_upload_sends_ciphertext_and_download_decrypts_plaintext` fails twice at unchanged `origin/master` (`9396adb6`) with `missing encrypted media secret for epoch 2`. This branch does not modify `marmot-app`, MLS, session, storage, or transport code. The adapter's ordered two-image inbound/outbound projection tests pass.

Fixes #1320

## Summary by CodeRabbit

- **New Features**
  - Send up to 10 images as one ordered, encrypted media message.
  - Supports captions, per-attachment results, and durable message ID reporting.
  - Automatically retries timed-out media sends without creating duplicates.
- **Bug Fixes**
  - Added validation for file size, total batch size, attachment count, and unsupported reply-targeted media.
  - Improved protection against unsafe file paths and cleanup after failed or cancelled uploads.
- **Documentation**
  - Documented media batching limits, validation, retry behavior, captions, and cleanup guarantees.

### DIFF
--- a/crates/agent-connector/src/messaging.rs
+++ b/crates/agent-connector/src/messaging.rs
@@ -19,6 +19,8 @@ use crate::validation::normalize_hex;
 
 /// Current schema version for persisted `send_final` request fingerprints.
 pub(crate) const SEND_FINAL_FINGERPRINT_VERSION: u8 = 1;
+/// Current schema version for persisted `send_media` request fingerprints.
+pub(crate) const SEND_MEDIA_FINGERPRINT_VERSION: u8 = 1;
 
 /// Server-derived fingerprint of a `send_final` request: a versioned SHA-256 digest
 /// over the destination (account + group), message text, and optional reply target.
@@ -40,6 +42,41 @@ pub(crate) fn send_final_fingerprint(
         reply_to_message_id_hex,
     ]);
     let bytes = serde_json::to_vec(&preimage).expect("send_final fingerprint preimage cannot fail");
+    hex::encode(Sha256::digest(bytes))
+}
+
+/// Server-derived fingerprint of a media send. Local paths are deliberately
+/// excluded; retries may stage the same bytes under a different private path.
+/// Attachment order, safe metadata, caption, destination, and plaintext bytes
+/// are all bound so one key cannot return ids for a different album.
+pub(crate) fn send_media_fingerprint(
+    account_id_hex: &str,
+    group_id_hex: &str,
+    caption: Option<&str>,
+    attachments: &[MediaUploadAttachmentRequest],
+) -> String {
+    use sha2::{Digest, Sha256};
+
+    let attachments = attachments
+        .iter()
+        .map(|attachment| {
+            serde_json::json!([
+                attachment.file_name,
+                attachment.media_type,
+                hex::encode(Sha256::digest(&attachment.plaintext)),
+                attachment.dim,
+                attachment.thumbhash,
+            ])
+        })
+        .collect::<Vec<_>>();
+    let preimage = serde_json::json!([
+        SEND_MEDIA_FINGERPRINT_VERSION,
+        account_id_hex,
+        group_id_hex,
+        caption,
+        attachments,
+    ]);
+    let bytes = serde_json::to_vec(&preimage).expect("send_media fingerprint preimage cannot fail");
     hex::encode(Sha256::digest(bytes))
 }
 
@@ -417,6 +454,7 @@ impl AgentConnector {
         group_id_hex: &str,
         attachments: Vec<AgentControlMediaUpload>,
         caption: Option<String>,
+        idempotency_key: Option<String>,
     ) -> Result<AgentControlResponse, ConnectorError> {
         let account = self.local_account_for_account_id(account_id_hex)?;
         let group_id_hex = normalize_hex(group_id_hex)?;
@@ -435,6 +473,28 @@ impl AgentConnector {
                 thumbhash: attachment.thumbhash,
             });
         }
+        let fingerprint = send_media_fingerprint(
+            account_id_hex,
+            &group_id_hex,
+            caption.as_deref(),
+            &upload_attachments,
+        );
+        let idempotency = if let Some(key) = idempotency_key {
+            match self.idempotency.acquire(&key, &fingerprint).await? {
+                SendIdempotencyAcquisition::Completed((
+                    message_ids_hex,
+                    maintenance_disposition,
+                )) => {
+                    return Ok(AgentControlResponse::FinalSent {
+                        message_ids_hex,
+                        maintenance_disposition,
+                    });
+                }
+                SendIdempotencyAcquisition::Leader(reservation) => Some((key, reservation)),
+            }
+        } else {
+            None
+        };
         let result = self
             .runtime
             .upload_media(
@@ -448,15 +508,26 @@ impl AgentConnector {
                 },
             )
             .await?;
-        let sent = result.sent;
+        let Some(sent) = result.sent else {
+            return Ok(AgentControlResponse::FinalSent {
+                message_ids_hex: Vec::new(),
+                maintenance_disposition: AgentControlSendMaintenanceDisposition::default(),
+            });
+        };
+        let message_ids_hex = sent.message_ids;
+        let maintenance_disposition = agent_maintenance_disposition(sent.maintenance_disposition);
+        if let Some((key, reservation)) = idempotency {
+            self.idempotency.record_with_disposition(
+                key,
+                fingerprint,
+                message_ids_hex.clone(),
+                maintenance_disposition,
+            );
+            reservation.complete(message_ids_hex.clone(), maintenance_disposition);
+        }
         Ok(AgentControlResponse::FinalSent {
-            message_ids_hex: sent
-                .as_ref()
-                .map(|sent| sent.message_ids.clone())
-                .unwrap_or_default(),
-            maintenance_disposition: sent
-                .map(|sent| agent_maintenance_disposition(sent.maintenance_disposition))
-                .unwrap_or_default(),
+            message_ids_hex,
+            maintenance_disposition,
         })
     }
 
--- a/crates/agent-connector/src/messaging.rs
+++ b/crates/agent-connector/src/messaging.rs
@@ -11,16 +11,64 @@ use marmot_app::{
     MediaUploadAttachmentRequest, MediaUploadRequest,
 };
 
-use crate::AgentConnector;
 use crate::error::ConnectorError;
 use crate::maintenance::agent_maintenance_disposition;
+use crate::media_roots::MediaAllowedRoots;
 use crate::stream_session::SendIdempotencyAcquisition;
 use crate::validation::normalize_hex;
+use crate::{
+    AgentConnector, MAX_MEDIA_UPLOAD_ATTACHMENT_BYTES, MAX_MEDIA_UPLOAD_ATTACHMENTS,
+    MAX_MEDIA_UPLOAD_BATCH_BYTES,
+};
 
 /// Current schema version for persisted `send_final` request fingerprints.
 pub(crate) const SEND_FINAL_FINGERPRINT_VERSION: u8 = 1;
 /// Current schema version for persisted `send_media` request fingerprints.
 pub(crate) const SEND_MEDIA_FINGERPRINT_VERSION: u8 = 1;
+
+#[derive(Clone, Copy)]
+pub(crate) struct MediaUploadLimits {
+    pub(crate) max_attachments: usize,
+    pub(crate) max_attachment_bytes: u64,
+    pub(crate) max_batch_bytes: u64,
+}
+
+pub(crate) async fn read_media_upload_attachments_with_limits(
+    media_allowed_roots: &MediaAllowedRoots,
+    attachments: Vec<AgentControlMediaUpload>,
+    limits: MediaUploadLimits,
+) -> Result<Vec<MediaUploadAttachmentRequest>, ConnectorError> {
+    if attachments.len() > limits.max_attachments {
+        return Err(ConnectorError::MediaLimitsExceeded("attachment_count"));
+    }
+
+    let mut upload_attachments = Vec::with_capacity(attachments.len());
+    let mut total_bytes = 0_u64;
+    for attachment in attachments {
+        let remaining_batch_bytes = limits.max_batch_bytes.saturating_sub(total_bytes);
+        let max_read_bytes = limits.max_attachment_bytes.min(remaining_batch_bytes);
+        let limit_reason = if remaining_batch_bytes < limits.max_attachment_bytes {
+            "aggregate_bytes"
+        } else {
+            "attachment_bytes"
+        };
+        let plaintext = media_allowed_roots
+            .read_regular_file_limited(attachment.path.into(), max_read_bytes, limit_reason)
+            .await?;
+        if plaintext.is_empty() {
+            return Err(ConnectorError::MediaLimitsExceeded("empty_attachment"));
+        }
+        total_bytes += plaintext.len() as u64;
+        upload_attachments.push(MediaUploadAttachmentRequest {
+            file_name: attachment.file_name,
+            media_type: attachment.media_type,
+            plaintext,
+            dim: attachment.dim,
+            thumbhash: attachment.thumbhash,
+        });
+    }
+    Ok(upload_attachments)
+}
 
 /// Server-derived fingerprint of a `send_final` request: a versioned SHA-256 digest
 /// over the destination (account + group), message text, and optional reply target.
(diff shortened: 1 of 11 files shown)


integrations/hermes/marmot/adapter.py
```diff
@@ -2250,57 +2261,199 @@ async def _send_media_upload(
...
+                    logger.debug("Marmot send_media failed: %s", exc)
+                    return SendResult(success=False, error=str(exc), retryable=is_retryable(exc))
+            return SendResult(
+                success=False,
+                error=str(last_exc) if last_exc else "Marmot media send failed",
+                retryable=is_retryable(last_exc) if last_exc else False,
+            )
+        except Exception as exc:
+            logger.debug("Marmot outbound media staging failed: %s", exc)
+            return SendResult(success=False, error=str(exc), retryable=is_retryable(exc))
```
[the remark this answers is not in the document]
agent-p1p: Fixed in 0f759e30. Media send and staging failures now log only the exception class and return fixed privacy-safe errors. Regression tests inject staged paths into both failure paths and assert those paths reach neither caller errors nor logs.

integrations/hermes/marmot/adapter.py
```diff
@@ -3398,22 +3551,33 @@ async def _standalone_send(
...
             )
-            if not result.success:
-                return {"error": result.error or "Marmot media send failed"}
-            results.append(result.message_id)
-        return {"success": True, "message_id": results[-1] if results else None}
+        result = await adapter._send_media_batch(
+            str(chat_id),
+            attachments,
+            caption=str(message or ""),
+        )
```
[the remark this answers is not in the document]
agent-p1p: Fixed in 0f759e30. Standalone media batches now map absent or whitespace-only messages to no caption while preserving non-blank caption text exactly; the regression test covers both cases.

agent-p1p:
Also addressed both review nitpicks in 0f759e30: media fingerprints are computed only when an idempotency key is present, and the fingerprint regression now proves attachment order and destination are bound.

agent-p1p:
Adversarial review of exact head `0f759e3037c7cdc32a32d5fa0b113befbe99f43e`.

BLOCKING:

- Long album sends can be reported failed while the original request continues, then duplicate under a fresh key: `integrations/hermes/marmot/adapter.py:1049`, `integrations/hermes/marmot/adapter.py:1234`, `integrations/hermes/marmot/adapter.py:2318-2370`. `send_media()` uses the generic 30-second control timeout even though the runtime uploads attachments sequentially and each Blossom upload may run for 60 seconds. Closing the client socket does not cancel `AgentConnector::handle_request`; the server awaits the upload before attempting its response. The internal retry reuses the key, but a matching follower waits only 15 seconds in `SendIdempotencyStore::acquire` (`crates/agent-connector/src/stream_session.rs:211-216`). If the leader is still running, the connector returns `send_in_progress`; `_raise_if_error` (`adapter.py:1567-1572`) reconstructs that as non-retryable, so `_send_media_batch` returns failure and deletes staging. The original server task can still publish and record the old key. A subsequent Hermes retry invokes `_send_media_batch` again, generates a new UUID at line 2318, and can publish the same album again. This violates the timeout/idempotency and all-or-error acceptance criteria for any upload taking roughly 45 seconds or more. Keep the same operation key alive until the leader reaches a terminal result (and preserve retryability for `send_in_progress`), or otherwise use a media-specific completion protocol/timeout that cannot return failure while publication is still possible. Add a regression where the first server-side media operation outlives the client timeout and prove the caller ultimately receives the original IDs without a second publish.

- Preflight filesystem errors can still expose local media paths: `integrations/hermes/marmot/adapter.py:2295-2296`. The new validation catch returns `str(exc)` for `OSError`; a `resolve()`/`stat()` race or permission failure includes the absolute local path in that string. `send_multiple_images` then re-raises it at `adapter.py:2460-2463`, making it caller/log visible. This violates the integrations privacy rule and bypasses the fixed, path-free errors used by the staging and send handlers. Return a fixed preflight error and add a regression that forces `stat()`/resolution to raise with a staged private path, asserting the path appears in neither the caller error nor logs.

SUGGESTION:

- Enforce byte limits against the bytes actually staged/read: `integrations/hermes/marmot/adapter.py:2279-2307`. The code checks `stat().st_size`, then copies in a later loop; a file can grow between those operations and bypass both per-file and aggregate caps. The connector subsequently uses unbounded `read_to_end` (`crates/agent-connector/src/media_roots.rs:87-89`). Recheck/cap the staged copies or enforce the limits while the connector reads so the 10-item/128-MiB promise is a real boundary rather than an adapter-side snapshot.

- Add connector-level behavioral coverage for media idempotency. `crates/agent-connector/src/tests.rs:4784-4836` tests only fingerprint differences. It does not exercise the new acquire → upload → record/complete path, same-key concurrent retries, restart persistence, or the `result.sent == None` branch. The Python retry test stubs the connector, so it cannot catch cross-layer failures such as the timeout case above.

NITPICK:

none.

SUMMARY: The batching, ordering, caption, cleanup, reply blocking, standalone path, changelog, and adapter-side boundary tests are otherwise coherent, and the focused Python/Rust suites pass locally. The timeout path does not yet guarantee retry-safe publication, and one preflight error path still leaks local paths.

VERDICT: REQUEST_CHANGES

SENSITIVE_PATHS: `crates/agent-control/src/lib.rs`, `crates/agent-connector/src/connection.rs`, `crates/agent-connector/src/messaging.rs`, `crates/agent-connector/src/tests.rs`

integrations/hermes/marmot/adapter.py
```diff
@@ -2250,57 +2261,207 @@ async def _send_media_upload(
...
         try:
-            try:
-                response = await self.client.send_media(
-                    account_id,
-                    chat_id,
-                    [
+            wire_attachments: list[Dict[str, Any]] = []
+            for item in validated:
+                staged_path = stage_outbound_media_file(
+                    item["source_path"],
```
Datawav, line 2412: **P1 — Revalidate the opened source during staging.** Preflight resolves and stats each pathname, but staging later reopens it here. The source can be replaced in between with a larger file or an outside-root symlink, so the copied bytes may bypass both the allowlist and size limits. Open each source once, validate the descriptor/identity, and copy from that descriptor while enforcing actual copied-byte limits.

crates/agent-connector/src/messaging.rs
```diff
@@ -435,6 +473,30 @@ impl AgentConnector {
                 thumbhash: attachment.thumbhash,
             });
         }
+        let idempotency = if let Some(key) = idempotency_key {
+            let fingerprint = send_media_fingerprint(
+                account_id_hex,
+                &group_id_hex,
+                caption.as_deref(),
+                &upload_attachments,
```
Datawav, line 525: **P1 — Enforce media limits in the connector.** The connector has already read every requested file into `upload_attachments` before fingerprinting here, but the 10-item, per-file, and 128 MiB limits exist only in the Python adapter. Another control client can therefore force substantially larger in-memory batches. Apply count, individual-byte, and aggregate-byte checks in the Rust connector before retaining the complete batch.

Datawav (commented):

## Final audit

### Need, design, and alternatives

The PR addresses a real gap: Hermes currently emits multiple images as separate messages. It introduces one ordered Marmot media message, preflight limits, private staging and cleanup, and connector-persisted idempotency for timeout retries.

The overall design is coherent. Content-bound idempotency is preferable to client-only retry suppression, and a single album request preserves ordering and caption semantics. Alternatives include retaining separate sends, which cannot provide album-level ordering or atomic retry behavior, or extending the protocol with per-attachment captions/reply metadata before enabling batching. The current fail-closed handling of replies is reasonable.

### Actionable findings

1. **P1 — Preflight can be bypassed through a path-replacement race**

   In `integrations/hermes/marmot/adapter.py`, `_send_media_batch` validates and resolves every source and records its size, but later reopens each pathname during staging. A source can be replaced between those operations with a larger file or a symlink to a file outside `media_local_roots`. The staged bytes would therefore differ from the bytes authorized by the allowlist and size checks, potentially disclosing a local file or exceeding the advertised 64 MiB/128 MiB memory bounds. Albums enlarge this window because all files are checked before staging begins.

   Open each source once, validate the opened descriptor (`fstat` and an allowlist-safe identity check), and copy from that descriptor while enforcing actual copied-byte limits. Add tests that replace a source with an outside-root symlink and grow it after validation.

2. **P1 — Connector does not enforce the limits protecting its own memory use**

   The 10-attachment, per-file, and 128 MiB aggregate limits exist only in the Python adapter. `send_media_response` reads every requested file into `Vec<u8>` before fingerprinting/uploading without applying those limits. Another agent-control client—or a compromised adapter—can bypass the Python checks and force the connector to buffer a substantially larger batch. This contradicts the stated design goal that the aggregate cap bounds connector pre-upload memory.

   Enforce count, individual-byte, and aggregate-byte limits in the Rust connector before retaining the complete batch. Streaming uploads would be a larger alternative; connector-side validation is the smaller change and should still exist as defense in depth.

The advisory concern about `result.sent == None` causing duplicate publication is not actionable: the connector invokes `upload_media` with `send: true`, and the inspected runtime assigns `Some(summary)` after a successful publication. The caption/error-contract observations likewise do not demonstrate a defect against the stated album design.

_Advisory audit of head [`0f759e3037c7`](https://github.com/marmot-protocol/mdk/commit/0f759e3037c7cdc32a32d5fa0b113befbe99f43e); no approval, merge, push, checkout, or PR-code execution was performed._

integrations/hermes/marmot/adapter.py
agent-p1p: Fixed in a98a2546. The adapter now pins each allowed root descriptor for the batch, opens source components descriptor-relatively with no symlink traversal, and stages from the single approved source descriptor. Copy-time per-file and aggregate limits cover file growth. Regressions replace both an ancestor and the configured root with outside symlinks and grow a source after validation.

crates/agent-connector/src/messaging.rs
agent-p1p: Fixed in a98a2546. The connector now rejects more than 10 attachments and enforces 64 MiB-minus-tag per-file plus 128 MiB aggregate plaintext limits while reading. Reads are capped at limit+1 rather than unbounded, and the connector tests cover accepted boundaries plus count, per-file, and aggregate rejection.

agent-p1p:
Addressed the adversarial review in a98a2546:

- Media requests retain the normal write timeout but wait for the connector terminal response. send_in_progress remains retryable and reuses the same key past the generic retry budget. The delayed-server regression outlives the generic client timeout and returns the original IDs after one publish.
- Preflight OSError details are replaced by a fixed privacy-safe error; the regression asserts a private path reaches neither caller nor logs.
- Allowed roots and sources are descriptor-pinned with no-symlink traversal; staging enforces actual copied-byte limits.
- The Rust connector independently enforces attachment-count, per-file, and aggregate limits before retaining the batch.

Verified: 186 Hermes adapter tests; 136 agent-connector tests plus identity-security; just fast-ci; independent pre-commit review passed.

agent-p1p:
@coderabbitai review

integrations/hermes/marmot/adapter.py
```diff
@@ -1209,16 +1295,24 @@ async def send_media(
...
+        }
+        key = str(idempotency_key or "").strip()
+        if key:
+            payload["idempotency_key"] = key
+        # Media upload duration is bounded by connector attachment/byte limits
+        # and per-endpoint HTTP deadlines, not by the generic 30-second control
+        # timeout. Keep the response wait attached to the one durable operation
+        # instead of reporting a timeout while it can still publish. Socket
+        # writes remain timed; transport failures retry with the same key.
+        return await self.request(payload, response_timeout=None)
```
Datawav: Add a finite end-to-end deadline here. Disabling the response timeout allows a connector that keeps the socket open without responding to block this send indefinitely, retaining the staged files and pinned descriptors until external cancellation. Use one generous monotonic deadline across the response wait and subsequent `send_in_progress` polling, and return a retryable failure when it expires so the existing cleanup runs.

Datawav (commented):

⚠️ Add a finite end-to-end deadline for media sends.

- **Medium — Media sends can block indefinitely** (`integrations/hermes/marmot/adapter.py`, `AgentControlClient.send_media` and `_send_media_batch`). `send_media()` disables the control response timeout with `response_timeout=None`, while `_send_media_batch()` retries `send_in_progress` inside an unconditional `while True`. A connector that keeps the socket open without responding—or continually reports an in-progress reservation—therefore leaves the response pipeline blocked and retains up to 128 MiB of staged files plus pinned descriptors until external cancellation. Cleanup on cancellation is correct; the defect is the absence of an internal termination condition.

  **Need:** Preserve idempotent handling of slow uploads without allowing one failed operation to consume resources forever.

  **Design:** Apply one generous monotonic deadline across the response wait and `send_in_progress` polling. On expiry, return a retryable failure so the existing `finally` cleanup executes. Add coverage for a connector that never responds and one that returns `send_in_progress` indefinitely.

  **Alternatives:** A bounded polling-attempt count is simpler but makes elapsed time dependent on backoff settings; restoring only the generic 30-second socket timeout may prematurely abandon valid large uploads.

_Advisory audit of head [`a98a25465609`](https://github.com/marmot-protocol/mdk/commit/a98a2546560907efe7f9941be9cef60b66db860b); no approval, merge, push, checkout, or PR-code execution was performed._

integrations/hermes/marmot/adapter.py
agent-p1p: Fixed in signed head 72dc3c7c. Media sends now share one 15-minute monotonic completion deadline across the initial socket response and every `send_in_progress` poll. Each response wait and backoff is capped by the remaining budget; expiry returns a fixed retryable timeout and the existing `finally` cleanup closes pinned descriptors and removes staged files. Added regressions for a connector that never responds and an indefinitely in-progress send, including cleanup and idempotency-key reuse.

Datawav (commented):

Verified in this delta:

- A single 15-minute monotonic deadline now bounds both connector response waits and repeated `send_in_progress` polling.
- Deadline expiry returns a retryable timeout result.
- Staged files and pinned descriptors are still released through `finally`.
- Added tests cover a silent connector and indefinitely repeated `send_in_progress` responses.

_Advisory audit of head [`72dc3c7c26a2`](https://github.com/marmot-protocol/mdk/commit/72dc3c7c26a2cc2d56cb5bd0ca5d15f5a46d4891); no approval, merge, push, checkout, or PR-code execution was performed._

PR #1329, opened 2026-08-07
Feat/fail closed missing fork anchor

jgmontoya:
## Summary

Prerequisite of #1293, split out per review there: the fail-closed behavior for a fork rival whose recovery material is missing, isolated so it can be reviewed, reverted, and bisected independently of the route unification.

## The gap

A device that receives a same-epoch rival commit within the rewind horizon adjudicates it through distributed convergence — but the ingest admission gate requires the retained anchor snapshot at the fork's source epoch. When that anchor is absent (pre-mechanism database, storage loss), the rival previously fell through the WrongEpoch arm's stale fallback: `MessageState::Failed`, `IngestOutcome::Stale`, zero events, zero audit. The device silently kept its own — possibly losing — branch: permanent divergence invisible to app and forensics (pinned by the first commit's characterization tests, on both post-restart routes).

## What it does now

The ingest seam stops being silent, without claiming authority it cannot have:

- **Ingest keeps, records, and schedules.** The rival's retained content row stays `Created`, a rejection audit row names the missing material, the group is scheduled for convergence, and the outcome is `Buffered` as before. No durable terminal state is written at this arm.
- **The durable halt is owned by the convergence coordinator.** The pass that ingest schedules reaches the coordinator's existing `MissingRetainedAnchor` verdict — durable `unrecoverable` marker, `epoch_state_changed{reason=missing_retained_anchor}` (now that reason's single emission site), and `GroupEvent::GroupUnrecoverable` so the app can surface repair. Canonical state is left untouched. Pinned end-to-end by `restarted_committer_without_source_anchor_halts_through_convergence`.
- **A verified Welcome repair survives the next drain** (4e92db56). A replacement Welcome already discards the device's live MLS copy and deletes its frozen convergence pass so residue cannot re-halt the repaired group; unresolved commits retained below the new copy's epoch are the other half of that residue, and `do_join_welcome` now retires them (`EpochInvalidated`, audited per message) in the same join transaction. The bound is the processed Welcome's own `MlsGroup::epoch()` — locally derived, authenticated material; a row's claimed epoch decides only its own fate and only in the safe direction.
- **Commits forking from before the device's join epoch keep the stale fallback** — the one legitimate in-horizon absence (mid-history joiners never held that state).

## Why the halt is not written at ingest

Earlier revisions of this branch mirrored #1285 directly: a durable `unrecoverable` marker written at the ingest arm, with the rival parked under its claimed source epoch. That design was removed in `209fddc0`, and the reason is load-bearing:

OpenMLS raises `ValidationError::WrongEpoch` from `validate_framing`, the first statement of `decrypt_message` — strictly upstream of the membership-tag check and signature verification. Nothing cryptographic has run when this arm executes, and nothing could: the membership MAC needs the claimed epoch's membership key and the signature covers that epoch's serialized group context, both living in exactly the anchor snapshot that is missing by precondition. The claimed epoch is unverifiable in principle at this seam, not merely unverified.

Writing a durable halt from that claim was therefore an attacker-reachable freeze primitive: any member — or a removed member whose epoch snapshot is still retained — could wrap a parseable commit claiming any in-horizon past epoch and durably freeze every device holding no anchor for it. Parking the rival keyed by the claimed epoch was the same defect one level down, since the claimed key steers where a later pass rewinds to. Both are gone; the ingest-time halt was a duplicate route to a verdict the coordinator already reaches, and only the attacker-reachable copy is removed. The regression test drives two forgeries at an anchor-less device (auth-tag byte flip, and epoch surgery on a real commit's `FramedContent` epoch field) — both fail RED on the pre-fix code at the durable marker.

This also sets a crate convention, recorded in the docs commit: never derive durable terminal group state, or a durable input keyed by an unverified claim, from unauthenticated inbound bytes.

## What this does not fix

This PR stops the ingest seam from deriving durable terminal state from unauthenticated input, and the repair-time retirement removes superseded rows from the rewind-candidate set — but the convergence pass seeder still re-derives every retained commit row's source epoch from its own wire bytes, so the coordinator's halt remains reachable from a claim no seam currently vets. That is a separate, already-merged seam with every rewind and fork-recovery path in its blast radius; it needs its own change, proving test first. The scope notes in `209fddc0` and `4e92db56` state the same boundary.

## Reviewer notes

- `fail_closed_halt_atomicity.rs`, added at review's request in `644760a4`, is deleted by `209fddc0` along with its subject: it pinned the ingest-time marker-and-park transaction, and that seam now performs no durable write at all. Its one surviving invariant — a retained, unapplied rival hands its group to `drain_pending_convergence_groups` — moved into both `fork_detection` tests.
- The repair-time retirement fires on every replacement Welcome, not only explicit repairs: superseded-residue removal is the honest rule for any re-join, not a repair special case. `Sent` rows are in scope — a removed-and-re-added device can no longer publish them.
- `33338702` writes down the `with_transaction` atomicity contract the engine was already relying on (helpers writing through their own `&S` inside the closure join the transaction).

## Validation

- `just fast-ci` green
- `cargo nextest run -p cgka-engine --features test-policy-overrides`: 457/457
- The forgery regressions and the fault-injection tests carry quoted RED evidence in their commit messages; the first commit's characterization test flips its assertions deliberately in this stack (documented in the test's doc comment).

## Stack

#1293 (route unification) rebases onto this branch and completes the story: with the pairwise route deleted, the scheduled-convergence outcome here is the only honest ingest disposition for a committer with no recovery material.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved convergence handling for late competing commits within the rewind horizon.
  * Prevented unauthenticated or missing-anchor messages from incorrectly halting groups or changing canonical membership and epoch state.
  * Preserved rival commit data, recorded audit events, and scheduled recovery when convergence is required.
  * Retired commits superseded by verified replacement welcomes.

* **Tests**
  * Added coverage for branch convergence, restarts, missing anchors, forged inputs, and router changes.

* **Documentation**
  * Clarified audit events, message states, epoch handling, and transaction behavior.


Datawav (commented):

No actionable findings.

_Advisory audit of head [`cee29be5bab0`](https://github.com/marmot-protocol/mdk/commit/cee29be5bab09b77aeffe5f2cd63c1cb1fcfedcc); no approval, merge, push, checkout, or PR-code execution was performed._

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -984,6 +1000,24 @@ impl<S: StorageProvider> Engine<S> {
...
 
+                    // An in-horizon rival commit refused convergence admission
+                    // only because its fork-source anchor is gone SHOULD have
+                    // been adjudicated by distributed convergence. Falling
+                    // through to the stale fallback would silently keep our
+                    // own (possibly losing) branch — permanent divergence
+                    // invisible to the app and to forensics. Fail closed
+                    // loudly instead, mirroring the materialization-time
+                    // missing-recovery-material precedent.
+                    if pending_recovery.is_none() && convergence_refused_for_missing_anchor {
```
erskingardner, line 1011: **[P1] Authenticate the rival before turning it into a durable halt.** This branch is reached on `ValidationError::WrongEpoch`, but OpenMLS performs the framing epoch check before signature/content authentication. The outer peeler only proves that a member produced the wrapper; the ingest code already notes that a member may wrap arbitrary MLS bytes. On a database missing this anchor, a member can therefore take a syntactically valid commit, alter its claimed epoch into the rewind horizon, and force the group into welcome-only repair even though the inner commit is invalid. Please require authenticated rival evidence before setting `unrecoverable`, and add a negative regression with a parseable, signature-invalid past-epoch commit.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -2260,6 +2294,106 @@ impl<S: StorageProvider> Engine<S> {
...
+                // because the helper needs engine state (the convergence clock,
+                // the audit sink); the `StorageProvider::with_transaction`
+                // contract makes those writes join this unit regardless.
+                self.persist_openmls_wire_message(
+                    openmls_msg,
+                    &group_id,
+                    msg_epoch,
+                    MessageState::ConvergenceDeferred,
+                )
+            })?;
```
erskingardner: **[P1] Preserve a scheduled route to re-derive the halt when this transaction fails terminally.** The content row was persisted `Created` before entering this helper, but `?` returns here before `schedule_pending_convergence_group` below. The new tests inject only retryable `Busy` and then call convergence directly, bypassing the runtime scheduler. For `Backend`/`Closed`, the caller need not retry, leaving the group live on the possibly losing branch with a pass-opening rival that nothing scheduled. Please schedule the group (or otherwise gate it) on this error path and cover a terminal fault through the scheduler-facing API rather than manually invoking convergence.

erskingardner (dismissed):
Blocking on two fail-closed boundary issues: the halt currently trusts an unauthenticated inner commit, and terminal storage failures can leave the pass-opening rival unscheduled. The marker/park SQLite transaction itself is correctly ordered, and CI is green at this head.

Datawav:
## Automated review

Follow-up found one newly introduced actionable issue; there were no prior requested changes to reassess.

- **P1 — Remove the release-version bump.** Changed lines add an `AGENTS.md` rule reserving version changes for explicitly requested releases, while `Cargo.toml`, workspace packages in `Cargo.lock`, and conformance vectors simultaneously move `0.9.10` → `0.9.11`. This feature PR would consume a release version and relabel fixtures outside the release workflow, creating misleading package and conformance metadata. Revert those version-only changes while retaining the functional delta.

Datawav (commented):
## Automated review

The prior release-version bump is fixed. One new actionable issue remains.

- **P1 — The unauthenticated group-freeze path is deferred, not eliminated.** Changed `ingest.rs` code retains the missing-anchor rival and calls `schedule_pending_convergence_group`; changed `fork_detection.rs` comments explicitly acknowledge that convergence re-derives `source_epoch` from the same unauthenticated bytes and can still halt the group later. The forged-message test drains the schedule without executing it. On a device with a genuine in-horizon anchor gap, a member can therefore submit a parseable forged past-epoch commit that ultimately marks the group unrecoverable. Smallest sound correction: prevent unauthenticated retained rows from triggering `MissingRetainedAnchor`; establish anchor gaps from local state independently, and extend the forged-message test through convergence to verify the group remains recoverable.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -2260,6 +2289,80 @@ impl<S: StorageProvider> Engine<S> {
...
+        &mut self,
+        group_id: GroupId,
+        msg_id: &MessageId,
+        current: EpochId,
+    ) -> Result<IngestOutcome, EngineError> {
+        // `Buffered` promises a later replay of the retained row, and
+        // applications open passes off `drain_pending_convergence_groups`, so
+        // the seam owes the group a schedule. The insert is idempotent and
+        // in-memory.
+        self.schedule_pending_convergence_group(&group_id);
```
Datawav, line 2366: **P1:** Scheduling convergence here still lets unauthenticated bytes freeze the group later. The retained row remains pass-opening, and convergence re-derives its `source_epoch` from the same unverified message; a forged past-epoch commit can therefore drive `MissingRetainedAnchor` and mark the group unrecoverable. Prevent unauthenticated retained rows from triggering that verdict, derive genuine anchor gaps solely from local state, and extend the forgery test through convergence to verify the group remains recoverable.

PR #1332, opened 2026-08-08
fix(openclaw): import plugin-sdk symbols that survive the beta re-cut

erskingardner:
## Problem

The Marmot plugin fails to load entirely on `openclaw@2026.7.2-beta.7` (the `beta` channel):

```
[plugins] marmot failed to load from ~/.openclaw/extensions/marmot/dist/index.js:
SyntaxError: The requested module 'openclaw/plugin-sdk/media-runtime'
does not provide an export named 'assertLocalMediaAllowed'
[plugins] 1 plugin(s) failed to initialize (load: marmot).
```

The beta re-cut the `plugin-sdk/media-runtime` barrel. Diffing the published `latest` (`2026.7.1-2`) and `beta` tarballs:

| Symbol | `latest` | `beta` |
| --- | --- | --- |
| `getDefaultLocalRoots` | `plugin-sdk/media-runtime` | moved to `plugin-sdk/web-media` |
| `assertLocalMediaAllowed` | `plugin-sdk/media-runtime` | **dropped from the public plugin-sdk surface** (still exists internally in `dist/local-media-access-*.js`) |

A named import the host SDK no longer exports is an ESM link error, so the **whole plugin** fails to initialize and the Marmot channel disappears from OpenClaw — not just the media send path that used the symbol.

## Fix

Import from subpaths that carry the symbols on **both** channels, verified against both published tarballs:

- `getDefaultLocalRoots` + `LocalMediaAccessError` from `plugin-sdk/web-media`.
- `readLocalFileFromRoots` from `plugin-sdk/infra-runtime`, replacing the `assertLocalMediaAllowed` + `readFile` pair.

`infra-runtime` was chosen over the more obvious `plugin-sdk/file-access-runtime` deliberately: beta ships that entry with **no `.d.ts`**, so importing from it would break `pnpm typecheck` the moment the pin moves. `infra-runtime` carries both the symbol and its types on `latest` and `beta`.

Folding the guard and the read into one call is also a small hardening win. The old shape was check-then-read — `assertLocalMediaAllowed(path)` then `readFile(path)`, two separate resolutions of the same path. `readLocalFileFromRoots` makes the allowlist check and the open a single operation against a scoped root handle with symlink/hardlink policy applied, closing the TOCTOU window between guard and open. It returns `null` rather than throwing, so the caller raises `LocalMediaAccessError` to preserve the existing failed-send behavior; the tests asserting on that error name pass unchanged.

Also drops the `DurableInboundReplyDeliveryResult` type import, which beta no longer re-exports from `plugin-sdk/channel-outbound`. Deriving it from the delivery function's return type is identical on both channels and cannot drift again. Type-only, so it did not contribute to the load failure — but it would have broken `pnpm typecheck` on a pin bump.

## Regression guard

`test/plugin-sdk-surface.test.ts` asserts every runtime-valued `plugin-sdk/*` import the plugin makes still resolves. The list was extracted mechanically from `src/` rather than by hand. Run against the beta build it reports `['assertLocalMediaAllowed', 'getDefaultLocalRoots']` missing, so it would have caught this before shipping.

This is a canary, not a cure — it only tests whatever version is installed, and the pin is still `latest`. The real fix for the class of bug is a non-blocking CI job that typechecks against `openclaw@beta`; filed as follow-up rather than bundled here.

## Verification

- `just fast-ci` — pass
- `just openclaw-dev-test` (typecheck + 217 tests) — pass
- `just openclaw-dev-script-test` — pass
- CI's `openclaw-plugin` job reproduced locally (`pnpm install --frozen-lockfile` → `typecheck` → `test` → `build`) — pass
- Installed the real `openclaw@2026.7.2-beta.7` in a scratch project and confirmed all four moved symbols resolve at runtime
- Re-ran the full import audit against both tarballs — zero unresolved imports on `latest` **and** `beta`

`just openclaw-dev-e2e-connector` **fails locally**, with `wn-agent: startup failed code=io_error io_kind=PermissionDenied` on its `/tmp/omce-*` control socket. I stashed every change and re-ran it on a clean tree: identical failure. It is a sandbox limitation upstream of anything TypeScript — the plugin never connects — not a regression from this change. It does mean the media send path is covered here by unit tests rather than end-to-end.

Two caveats on the beta verification: this machine runs Node 24.5.0 and beta requires ≥24.15.0, so beta was installed with `--ignore-scripts` (skips a version-warning preinstall script only; module resolution is unaffected), and the plugin was not exercised end-to-end under beta.

## Not fixed here

A separate `Session "agent:main:main" changed while starting work. Retry.` was reported on the same beta instance. That is `CronSessionLifecycleClaimError` from OpenClaw core's `src/cron/isolated-agent/run-session-state.ts` — an optimistic-concurrency guard on the cron path, reachable only from `server-cron`. Unrelated to this change and not addressed by it.

Note this fix is source-only: an affected instance needs a rebuild, reinstall, and Gateway restart before the load error clears.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of outbound local media, including clearer access-denied errors and safer validation within configured locations.
  * Improved compatibility across supported OpenClaw SDK versions.
* **Reliability**
  * Added checks to detect missing SDK capabilities earlier, helping prevent integration startup or runtime failures.
  * Durable inbound reply delivery behavior remains unchanged.

### DIFF
--- a/integrations/openclaw/marmot/src/outbound.ts
+++ b/integrations/openclaw/marmot/src/outbound.ts
@@ -8,7 +8,7 @@
 // contract tests.
 
 import { randomUUID } from "node:crypto";
-import { chmod, mkdir, readFile, rm, writeFile } from "node:fs/promises";
+import { chmod, mkdir, rm, writeFile } from "node:fs/promises";
 import { homedir } from "node:os";
 import { basename, extname, join } from "node:path";
 import { fileURLToPath } from "node:url";
@@ -20,10 +20,12 @@ import {
   type MessageReceipt,
   type MessageReceiptPart,
 } from "openclaw/plugin-sdk/channel-outbound";
-import {
-  assertLocalMediaAllowed,
-  getDefaultLocalRoots,
-} from "openclaw/plugin-sdk/media-runtime";
+// `readLocalFileFromRoots` (allowlist check + hardened read) and the local media
+// access types are imported from the subpaths that export them on both the
+// `latest` and `beta` OpenClaw channels. The `plugin-sdk/media-runtime` barrel
+// dropped `assertLocalMediaAllowed`/`getDefaultLocalRoots` in 2026.7.2-beta.
+import { readLocalFileFromRoots } from "openclaw/plugin-sdk/infra-runtime";
+import { getDefaultLocalRoots, LocalMediaAccessError } from "openclaw/plugin-sdk/web-media";
 
 import type { AgentControlMediaUpload, MarmotAgentControlClient } from "./client.js";
 
@@ -193,8 +195,8 @@ function resolveAllowedMediaRoots(ctx: ChannelMessageSendMediaContext): readonly
  * Resolve `ctx.mediaUrl` to a local `AgentControlMediaUpload` the connector can
  * read by path. Handles two cases the ctx can express with the real SDK types:
  *
- * 1. A local filesystem path or `file://` URL — validated against the send's
- *    allowlisted media roots (`assertLocalMediaAllowed`), then copied to the
+ * 1. A local filesystem path or `file://` URL — read only through the send's
+ *    allowlisted media roots (`readLocalFileFromRoots`), then copied to the
  *    connector-shared staging root. Without the source guard
  *    an agent-influenced `mediaUrl` (e.g. `~/.ssh/id_rsa`) would let a prompt-
  *    injected agent exfiltrate any connector-host file into a group. This
@@ -207,8 +209,8 @@ function resolveAllowedMediaRoots(ctx: ChannelMessageSendMediaContext): readonly
  * the connector reads a path it cannot be given in that case (see Seam 2 note).
  *
  * Throws `LocalMediaAccessError` (from the SDK) when a local path escapes the
- * allowlist; the caller surfaces that as a failed send rather than reading the
- * file.
+ * allowlist; the caller surfaces that as a failed send. The file is never read
+ * in that case — the allowlist check and the read are the same operation.
  */
 async function resolveOutboundMediaUpload(
   ctx: ChannelMessageSendMediaContext,
@@ -217,12 +219,23 @@ async function resolveOutboundMediaUpload(
   const { mediaUrl } = ctx;
   if (isLocalMediaUrl(mediaUrl)) {
     const localPath = mediaUrl.startsWith("file://") ? fileURLToPath(mediaUrl) : mediaUrl;
-    // Defense against exfiltration via a tool/prompt-influenced path: the
-    // Keep OpenClaw's source policy as the first layer; wn-agent independently
-    // confines the staged path it ultimately reads. Throws on violation.
-    await assertLocalMediaAllowed(localPath, resolveAllowedMediaRoots(ctx));
+    // Defense against exfiltration via a tool/prompt-influenced path: the read
+    // itself is confined to the allowlisted roots, so a path outside them is
+    // never opened. Keep OpenClaw's source policy as the first layer; wn-agent
+    // independently confines the staged path it ultimately reads.
+    const read = await readLocalFileFromRoots({
+      filePath: localPath,
+      roots: resolveAllowedMediaRoots(ctx),
+      label: "outbound media roots",
+    });
+    if (!read) {
+      throw new LocalMediaAccessError(
+        "path-not-allowed",
+        "marmot: outbound media path is outside the allowed local media roots",
+      );
+    }
     const fileName = basename(localPath) || "attachment";
-    const path = await writeTempMedia(fileName, await readFile(localPath));
+    const path = await writeTempMedia(fileName, read.buffer);
     return {
       upload: { path, media_type: mimeFromExtension(fileName), file_name: fileName },
       cleanup: () => rm(path, { force: true }),
--- a/integrations/openclaw/marmot/src/outbound.ts
+++ b/integrations/openclaw/marmot/src/outbound.ts
@@ -208,9 +208,11 @@ function resolveAllowedMediaRoots(ctx: ChannelMessageSendMediaContext): readonly
  * Returns `null` when the ctx provides only a remote URL and no buffer accessor;
  * the connector reads a path it cannot be given in that case (see Seam 2 note).
  *
- * Throws `LocalMediaAccessError` (from the SDK) when a local path escapes the
- * allowlist; the caller surfaces that as a failed send. The file is never read
- * in that case — the allowlist check and the read are the same operation.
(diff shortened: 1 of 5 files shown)


erskingardner (commented):
**Recommend merge** (could not formally APPROVE — GitHub blocks self-approval on this PR).

Clear, necessary fix for a whole-plugin ESM load failure on the beta SDK re-cut. Dual-channel import choices look deliberate, the reply-result type derivation is the right kind of drift-proofing, and the surface canary matches the current `src/` runtime imports.

One non-blocking note inline on the `null` → `path-not-allowed` mapping.

integrations/openclaw/marmot/src/outbound.ts
```diff
@@ -217,12 +219,23 @@ async function resolveOutboundMediaUpload(
...
+    // independently confines the staged path it ultimately reads.
+    const read = await readLocalFileFromRoots({
+      filePath: localPath,
+      roots: resolveAllowedMediaRoots(ctx),
+      label: "outbound media roots",
+    });
+    if (!read) {
+      throw new LocalMediaAccessError(
+        "path-not-allowed",
+        "marmot: outbound media path is outside the allowed local media roots",
```
erskingardner: Non-blocking: `readLocalFileFromRoots` returns `null` for more than allowlist misses — missing files, IO failures, and symlink/hardlink policy rejections under an allowed root all collapse here too. Mapping every `null` to `LocalMediaAccessError("path-not-allowed", …outside…)` is fail-closed (good) but mislabels those cases vs the old `assertLocalMediaAllowed` + `readFile` split, where in-root failures surfaced as Node fs errors.

Worth a shorter/generic message (or a follow-up that distinguishes denial from read failure) so a missing in-root attachment does not look like an allowlist violation. Not a merge blocker.

erskingardner: Good catch — fixed in 120a1b04.

Kept the fail-closed behavior and the deny-shaped code, but the message no longer asserts the cause: `marmot: outbound media is not readable from the allowed local media roots`. Added a comment at the throw site enumerating what `null` actually covers (allowlist miss, missing file, IO error, symlink/hardlink policy rejection under an allowed root) so the next reader does not have to re-derive it.

Also added a regression test for the in-root-but-missing case. That path used to surface as a Node fs error from the `readFile` half of the old split and had no coverage either way, so it was worth pinning down rather than just softening the string.

Did not go further and distinguish denial from read failure: it needs a second path resolution purely for labeling, and the security decision has already been made fail-closed by then. Happy to file that as a follow-up if you want the sharper diagnostic.

integrations/openclaw/marmot/src/outbound.ts
```diff
@@ -217,12 +221,28 @@ async function resolveOutboundMediaUpload(
...
+    });
+    // `readLocalFileFromRoots` collapses every failure to `null`: an allowlist
+    // miss, but also a missing file, an IO error, or a symlink/hardlink policy
+    // rejection *under* an allowed root. Fail closed on all of them, but do not
+    // claim which one it was — a missing in-root attachment should not read as
+    // an exfiltration attempt in the logs.
+    if (!read) {
+      throw new LocalMediaAccessError(
+        "path-not-allowed",
+        "marmot: outbound media is not readable from the allowed local media roots",
```
Datawav, line 241: Every falsy result from `readLocalFileFromRoots` is classified as `path-not-allowed`, even though the helper also returns `null` for missing in-root files, I/O failures, and filesystem-policy rejections. This loses the original path/cause and misleadingly labels non-policy failures as allowlist violations. Preserve a typed/sanitized cause or use a generic read-failure classification without restoring the separate TOCTOU-prone read.

integrations/openclaw/marmot/test/outbound.test.ts
```diff
@@ -347,6 +347,36 @@ describe("createMarmotMessageAdapter", () => {
...
+        to: HEX32("cc"),
+        text: "missing",
+        mediaUrl: join(tmpRoot, "absent.png"),
+        mediaLocalRoots: [tmpRoot],
+      } as unknown as ChannelMessageSendMediaContext;
+
+      await expect(adapter.send!.media!(ctx)).rejects.toMatchObject({
+        name: "LocalMediaAccessError",
+      });
+      await expect(adapter.send!.media!(ctx)).rejects.toThrow(/not readable/);
```
Datawav, line 373: This invokes `adapter.send!.media!(ctx)` a second time solely to check the message, repeating the operation and potentially hiding stateful behavior or causing unnecessary side effects. Capture a single rejection and assert both its name and message.

Datawav (commented):

## Final audit

**Need:** The PR addresses a plugin-wide ESM load failure caused by SDK exports changing between OpenClaw channels.

**Design:** It moves runtime imports to compatible SDK subpaths, derives the removed outbound result type from the function signature, replaces check-then-read with `readLocalFileFromRoots`, and adds an SDK-surface canary.

### Actionable findings

1. **Low — Distinct media-read failures are misclassified and lose diagnostics**  
   In `src/outbound.ts`, every falsy `readLocalFileFromRoots` result becomes `LocalMediaAccessError("path-not-allowed", ...)`. The patch itself establishes that this includes a missing in-root file, yet callers receive neither the original path nor an underlying filesystem cause. Operators therefore cannot distinguish a missing attachment from an allowlist or I/O rejection, and the error code specifically implies a policy failure.  
   **Fix:** Preserve a typed/sanitized cause or use a generic read-failure classification. If the SDK exposes only `null`, extend its API to return structured failure information; do not restore the TOCTOU-prone standalone `readFile`.

2. **Low — The new regression test executes the send twice**  
   `test/outbound.test.ts` independently calls `adapter.send!.media!(ctx)` for the name and message assertions. This repeats the operation and could conceal stateful behavior or introduce unnecessary side effects.  
   **Fix:** Capture one rejection and assert both properties from it, or use one matcher covering name and message.

### Reconciled limitations, not current defects

- The runtime-import canary is manually maintained and tests only the installed SDK pin. Both limitations are explicitly documented; the supplied patches do not demonstrate an omitted current import. Automated import extraction and beta-channel CI are worthwhile alternatives/follow-ups, not proven correctness defects here.
- Claims that `readLocalFileFromRoots` atomically closes the TOCTOU window, enforces symlink/hardlink policy, or behaves identically across channels cannot be independently established from the supplied patch because the SDK implementation and declarations are absent.
- The dispatch return-type derivation and fail-closed staging flow are supported by the patch.

_Advisory audit of head [`120a1b04ddd5`](https://github.com/marmot-protocol/mdk/commit/120a1b04ddd5b62409745c42b96ab21be11f458b); no approval, merge, push, checkout, or PR-code execution was performed._

PR #1333, opened 2026-08-08
Add an explicit storage close so hosts can suspend without a lock kill

erskingardner:
## Why

The White Noise iOS app is being SIGKILL'd in the background with `RUNNINGBOARD 0xdead10cc` — suspended while holding a file lock in the shared App Group container. Five identical TestFlight reports (build 2026.8.6 (23), iOS 18.7.8 and 26.5.2, three devices), in two signatures.

Nothing in this workspace could release those locks:

- `MarmotAppRuntime::shutdown()` / `AccountManager::shutdown()` take `&self`, so they stop workers but cannot drop anything.
- `SharedConnection` is a `Clone` wrapper over an `Arc`, handed out freely to the engine, the OpenMLS adapter, and app projections. It had no `close()` and no `Drop`. The connection went away only when the last clone anywhere in the process did — which the host can neither observe nor await.
- In WAL mode an open connection holds a persistent lock on its `-shm` sidecar for its entire lifetime.

So "the workers stopped" and "the store is unlocked" were different facts, and only the second one keeps the process alive. Three of the five reports had **no Rust threads at all** and were still killed.

The other two were an account worker frozen mid-write inside `rollback_snapshot` / snapshot capture. The 5s worker timeout wasn't a real bound there: `tokio::task::abort()` is cooperative, and a per-group convergence pass is a long *synchronous* engine + SQLite stretch with no await for it to land at.

## What changed

**Storage can be closed** — `storage-sqlite/src/connection.rs`

New public `CloseableConnection` holds the connection in a `Mutex<Option<Connection>>`. `close()` checkpoints (`PRAGMA wal_checkpoint(TRUNCATE)`), takes the connection out, and closes it for real. Surviving clones get the new `StorageError::Closed` rather than panicking or keeping the file locked. `SqliteAccountStorage`, `SqliteSharedStorage`, and the app's directory caches all build on it, so the checkpoint-then-close ordering lives in one place.

The transaction machinery is closed-aware. A `ROLLBACK` against a closed connection reports **success** — closing performed it inside SQLite — while a `COMMIT` reports `Closed`. Latching the connection "unusable" on that rollback would turn an orderly host suspension into a corruption-shaped error. Threads parked on the transaction condvar are woken so they bail out instead of blocking on an owner that will never run again.

**`MarmotAppRuntime::shutdown_and_close()`** — drains workers, then closes every account database, the shared database, and every directory cache. Terminal by design: `MarmotApp` latches a closed flag and every accessor refuses afterwards, because a stray background read that transparently reopened would re-lock the container the host was just told is clear. Idempotent, and safe with or without a preceding `shutdown()`.

**Shutdown-aware engine dispatch** — `runtime/account_worker.rs`

The scheduled-convergence `for group_id in groups` loop now checks `lifecycle.is_stopping()` between groups. That boundary is the only cut point where no `SnapshotRollbackGuard` is live, so it is also the only one that cannot leave a group half-rolled-back — a partially applied snapshot capture or rollback would be strictly worse than the crash this fixes. Undispatched groups need no hand-off; their convergence inputs are durable and the next runtime rediscovers them at catch-up. The maintenance tick (capped at 15s, 3× the whole shutdown budget) is skipped outright when stopping.

## One finding beyond the reported diagnosis

The root runtime lease (`.marmot-runtime.lock`) is an advisory `flock` on a file **inside the Marmot root** — the same App Group container — and was documented as held until the last app/runtime handle drops, even after `shutdown()`. That is a second, independent `0xdead10cc` trigger and would have left the no-threads signature partly unfixed. `close_storage` now takes and drops it; the field moved from `Option<Arc<Lease>>` to `Arc<Mutex<Option<Lease>>>` so the shared-across-clones semantics are preserved while an explicit close can still release it early.

## FFI surface for the iOS handoff

```rust
pub async fn shutdown_and_close(&self) -> Result<(), MarmotKitError>  // Swift: shutdownAndClose()
pub fn storage_is_closed(&self) -> bool                               // Swift: storageIsClosed()
```

Plus a typed `MarmotKitError::StorageClosed { details }` — non-retryable, so work racing the close reads as "we shut down" rather than a storage fault the user gets told about.

The iOS app picks these up via `./scripts/sync-bindings.sh` and wires `shutdownAndClose()` into `RuntimeLifecycle.prepareForBackgroundSuspension()`. Per this crate's lockstep invariant, **versions are not bumped here** — binding regeneration and the version bump ride the release.

## Tests

- `close_releases_wal_locks_and_sidecars` — `-wal`/`-shm` unlinked, and a *fresh* connection takes `locking_mode = EXCLUSIVE` and writes. That exclusive open is the real proxy for what iOS checks.
- `close_under_a_live_writer_is_prompt_and_leaves_no_torn_transaction` — 10k two-table transactions on a worker thread, closed mid-flight; asserts the close returns in under 1s and both tables match on reopen.
- `transaction_open_at_close_reports_closed_without_latching_unusable`, `close_wakes_threads_waiting_for_the_transaction_slot`, idempotency, and surviving-clone / OpenMLS-adapter handles reporting `Closed`.
- App level: all three database families' sidecars released, the root lease re-acquirable by a second owner, accessors refusing afterwards, and `shutdown_and_close` idempotent with and without a prior `shutdown`.

**Not covered by a test:** the runtime-level deferral in the convergence loop. Exercising it needs a live account with a multi-group batch and a shutdown landing mid-batch, which only the relay harness can set up. The equivalent guarantee is tested at the storage layer instead. Flagging it rather than implying coverage.

## Verification

`just fast-ci` clean. `storage-sqlite` 329 passed, `marmot-app --lib` 568, `relay_runtime` 92 (CI invocation: `--features test-policy-overrides -- --test-threads=1`), plus `marmot-account`, `cgka-session`, `cgka-traits`, and `marmot-uniffi`.

## Notes for review

- `StorageError` gains a `Closed(String)` variant. It is deliberately its own variant and non-transient — `is_transient()` stays false, and a new `is_closed()` lets callers classify orderly shutdown apart from a storage fault. Compiler-checked at every match site.
- WAL is unchanged. The app and the Notification Service Extension both open this store and WAL is what makes that workable; the fix is to close the store, not to weaken it.
- `docs/marmot-architecture/overview/local-artifact-safety.md` gains a "Releasing artifacts before host suspension" section, and the crate `AGENTS.md` files carry the matching invariants.

## Summary by CodeRabbit

- **New Features**
  - Added a complete shutdown-and-close operation that stops services, closes local storage, and releases file locks.
  - Added storage status checks to confirm when closure is complete.
  - Added a distinct “Storage Closed” error for operations attempted after closure.

- **Reliability**
  - Improved handling of temporary SQLite contention with bounded retries.
  - Prevented new work from starting during shutdown.
  - Closure is terminal, safe to repeat, and prevents storage from reopening.

### DIFF
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -445,6 +445,11 @@ pub struct MarmotApp {
     /// a late call cannot silently reopen a database — and re-lock a container
     /// the host has just been told is lock-free.
     storage_closed: Arc<AtomicBool>,
+    /// Admission control that makes the close *atomic* rather than merely
+    /// latched: database opens hold the read side across create-and-publish,
+    /// [`Self::close_storage`] holds the write side across its whole teardown.
+    /// See [`Self::begin_storage_open`].
+    storage_lifecycle: Arc<RwLock<()>>,
     relay_urls: Vec<String>,
     account_home: AccountHome,
     relay_plane: MarmotRelayPlane,
@@ -1159,6 +1164,7 @@ impl MarmotApp {
             root,
             root_runtime_lease: Arc::new(Mutex::new(None)),
             storage_closed: Arc::new(AtomicBool::new(false)),
+            storage_lifecycle: Arc::new(RwLock::new(())),
             relay_urls,
             relay_plane,
             config,
@@ -1220,6 +1226,7 @@ impl MarmotApp {
             root: root.as_ref().to_path_buf(),
             root_runtime_lease: Arc::new(Mutex::new(None)),
             storage_closed: Arc::new(AtomicBool::new(false)),
+            storage_lifecycle: Arc::new(RwLock::new(())),
             relay_urls,
             account_home,
             relay_plane,
@@ -4133,6 +4140,19 @@ impl MarmotApp {
     /// error is returned once all of them have been closed.
     pub fn close_storage(&self) -> Result<(), AppError> {
         let started_at = Instant::now();
+        // Exclusive for the whole teardown. The `storage_closed` flag alone
+        // would not make this atomic: two concurrent closes could interleave so
+        // that one released the root lease and returned while the other was
+        // still closing connections, and an open already in flight could finish
+        // creating its connection after this method returned. Either way the
+        // host would be told the container is lock-free while a lock still
+        // existed. Holding the writer means every opener has either published
+        // (so the drain below closes it) or has not yet checked the flag (so it
+        // will refuse).
+        let _lifecycle = self
+            .storage_lifecycle
+            .write()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
         self.storage_closed.store(true, Ordering::Release);
         let mut first_error = None;
         let mut closed = 0usize;
@@ -4215,6 +4235,30 @@ impl MarmotApp {
         Ok(())
     }
 
+    /// Admission for a database *open*, held from the closed check through
+    /// publication into the handle cache.
+    ///
+    /// Opens are readers, so they still run concurrently with each other;
+    /// [`Self::close_storage`] is the writer. That is what makes the close's
+    /// promise true: an open holding this guard either publishes before the
+    /// close can start draining, or has not yet checked the flag and will
+    /// refuse. No connection can be created after the close returns.
+    ///
+    /// Cache *hits* deliberately skip this — a handle already in the cache is
+    /// one the close will drain and shut, so returning a clone of it cannot
+    /// leak a lock, and the hot read path stays uncontended.
+    fn begin_storage_open(
+        &self,
+        database: &'static str,
+    ) -> Result<RwLockReadGuard<'_, ()>, AppError> {
+        let guard = self
+            .storage_lifecycle
+            .read()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        self.ensure_storage_open(database)?;
+        Ok(guard)
+    }
+
     fn account_storage(&self, label: &str) -> Result<SqliteAccountStorage, AppError> {
         self.ensure_storage_open("account storage")?;
         if let Some(storage) = self
@@ -4226,6 +4270,7 @@ impl MarmotApp {
         {
             return Ok(storage);
         }
+        let _lifecycle = self.begin_storage_open("account storage")?;
         let _span = tracing::debug_span!(
             target: "marmot_app::storage",
             "account_storage_open",
@@ -4246,21 +4291,12 @@ impl MarmotApp {
             )?
         };
         let storage = SqliteAccountStorage::open_encrypted(&path, &key)?;
+        // Publishing under `_lifecycle` is what keeps this connection reachable
+        // by a later `close_storage`; see `begin_storage_open`.
         let mut storages = self
             .account_storages
             .lock()
             .unwrap_or_else(|poisoned| poisoned.into_inner());
-        // A close that latched while this open was in flight has already
-        // drained the map, so publishing here would strand an open connection
-        // — and its file locks — past `close_storage`. Close it instead. The
-        // ordering holds both ways: `close_storage` sets the flag before it
-        // takes this same lock, so either we observe the flag or the drain
-        // observes our entry.
-        if let Err(error) = self.ensure_storage_open("account storage") {
-            drop(storages);
-            let _ = storage.close();
-            return Err(error);
-        }
         Ok(storages
             .entry(label.to_owned())
             .or_insert_with(|| storage.clone())
@@ -4604,6 +4640,7 @@ impl MarmotApp {
         {
             return Ok(storage);
         }
+        let _lifecycle = self.begin_storage_open("shared storage")?;
         let _span = tracing::debug_span!(
             target: "marmot_app::storage",
             "shared_storage_open",
@@ -4615,13 +4652,6 @@ impl MarmotApp {
             .shared_storage
             .lock()
             .unwrap_or_else(|poisoned| poisoned.into_inner());
-        // See `account_storage`: a close that latched mid-open must not leave
-        // this connection published and holding the shared database's locks.
-        if let Err(error) = self.ensure_storage_open("shared storage") {
-            drop(shared);
-            let _ = storage.close();
-            return Err(error);
-        }
         Ok(shared.get_or_insert_with(|| storage.clone()).clone())
     }
 
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -430,6 +430,9 @@ type AppRuntime = AccountDeviceRuntime<
     AppKeyPackagePublisher,
 >;
 
+#[cfg(test)]
+type LegacyProjectionOpenHook = Arc<dyn Fn() + Send + Sync>;
+
 #[derive(Clone)]
 pub struct MarmotApp {
     root: PathBuf,
@@ -461,6 +464,8 @@ pub struct MarmotApp {
     legacy_directory_cache_checked: Arc<Mutex<bool>>,
     #[cfg(test)]
     directory_cache_open_count: Arc<std::sync::atomic::AtomicUsize>,
+    #[cfg(test)]
+    legacy_projection_open_hook: Arc<Mutex<Option<LegacyProjectionOpenHook>>>,
     #[cfg(test)]
     test_relay_client: Option<Arc<dyn NostrRelayClient>>,
     shared_storage: Arc<Mutex<Option<SqliteSharedStorage>>>,
@@ -1176,6 +1181,8 @@ impl MarmotApp {
             #[cfg(test)]
             directory_cache_open_count: Arc::new(std::sync::atomic::AtomicUsize::new(0)),
             #[cfg(test)]
+            legacy_projection_open_hook: Arc::new(Mutex::new(None)),
+            #[cfg(test)]
             test_relay_client: None,
             shared_storage: Arc::new(Mutex::new(None)),
             account_state_ready: Arc::new(Mutex::new(HashSet::new())),
@@ -1238,6 +1245,8 @@ impl MarmotApp {
             legacy_directory_cache_checked: Arc::new(Mutex::new(false)),
             #[cfg(test)]
             directory_cache_open_count: Arc::new(std::sync::atomic::AtomicUsize::new(0)),
+            #[cfg(test)]
+            legacy_projection_open_hook: Arc::new(Mutex::new(None)),
             #[cfg(test)]
             test_relay_client: None,
             shared_storage: Arc::new(Mutex::new(None)),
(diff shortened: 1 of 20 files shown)


crates/marmot-app/src/lib.rs
```diff
@@ -4088,7 +4101,122 @@ impl MarmotApp {
...
+            }
+        }
+
+        // Last: the lease guards the databases, so it outlives them.
+        drop(
+            self.root_runtime_lease
+                .lock()
+                .unwrap_or_else(|poisoned| poisoned.into_inner())
+                .take(),
+        );
```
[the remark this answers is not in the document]
erskingardner: Confirmed and fixed — added a `storage_lifecycle` `RwLock`: opens hold the read side from the closed check through publication, `close_storage` holds the write side across latch/drain/close/lease-release, so no connection can exist after it returns; the per-accessor "close what we just opened" cleanup is now unreachable and gone. Cache *hits* deliberately skip the guard (an already-cached handle is one the close will drain, so it cannot leak a lock) to keep the hot read path uncontended. Added `close_storage_waits_for_an_open_that_is_already_in_flight` — verified it fails without the guard — and `concurrent_close_storage_callers_serialize`.

crates/marmot-app/src/tests.rs
```diff
@@ -5342,3 +5342,134 @@ async fn an_escalation_recorded_during_a_received_delivery_rides_that_seam() {
...
+    assert!(app.storage_is_closed());
+    for db in [&session_db, &shared_db] {
+        for sidecar in sidecars(db) {
+            assert!(
+                !sidecar.exists(),
+                "{} must be gone once the last connection closes",
+                sidecar.display(),
+            );
+        }
+    }
```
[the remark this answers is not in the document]
erskingardner: Valid gap, fixed — but not via sidecars: the directory cache is opened without a `journal_mode` pragma, so it runs in rollback-journal mode and never has `-wal`/`-shm` to assert on (the suggested "assert its sidecars exist before closure" would fail). The test now keeps the handles taken *before* the close and asserts the connections themselves report closed — `directory_cache.entries()` returns `StorageError::Closed`, plus `is_closed()` on both WAL stores. That is strictly stronger than the sidecar check, since it distinguishes "connection closed" from "merely evicted from the app's cache".

crates/storage-sqlite/src/openmls_storage.rs
```diff
@@ -24,9 +24,7 @@ impl SqliteOpenMlsStorage {
         Ok(serde_json::to_vec(group_id)?)
     }
 
-    fn lock(
-        &self,
-    ) -> Result<std::sync::MutexGuard<'_, rusqlite::Connection>, SqliteOpenMlsStorageError> {
+    fn lock(&self) -> Result<crate::connection::ConnectionGuard<'_>, SqliteOpenMlsStorageError> {
```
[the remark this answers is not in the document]
erskingardner: Fixed — `lock()` now propagates through the adapter's existing `Storage(#[from] StorageError)` arm, so `StorageError::Closed` (and `Busy`) survive across all twelve value-store paths; the `Lock(String)` variant is now unreachable (the connection layer absorbs mutex poisoning itself) and removed. I did not re-type the downstream `EngineError::Backend` mapping: those flow through OpenMLS's generic `StorageProvider::Error` associated type, so making them typed is a broader change to engine error mapping than this PR should carry, and it would shift `Busy` classification too.

Datawav (commented):

## Final audit

### High — Close admits new database work while waiting

`CloseableConnection::close()` acquires `slot` before setting `closed`. While it waits behind an existing `ConnectionGuard`, later `lock()` calls may acquire the mutex and execute additional operations because `lock()` only checks whether the slot remains populated.

`SharedConnection` has a separate early closed latch, but `SqliteSharedStorage` and `DirectoryCache` use `CloseableConnection` directly. Racing clone users can therefore prolong or potentially starve closure, contradicting the stated contract that close waits only for the currently executing statement. This undermines the feature’s core need: reliably releasing App Group locks before the host’s suspension deadline.

Set a closing/closed state before waiting for `slot`, and have `lock()` reject admission once that state is published. An enum such as `Open | Closing | Closed` would express the design more clearly than the current late boolean. Alternatives include a read/write admission gate around every connection operation or moving the state into the mutex while providing a separate early-closing atomic for nonblocking rejection.

### Low — Synchronous SQLite teardown runs on the async executor

`MarmotAppRuntime::shutdown_and_close()` calls synchronous `close_storage()` directly after `shutdown().await`. Mutex waits, WAL checkpointing, busy handling, and connection closure therefore block a Tokio worker for an unbounded-by-this-API interval. The existing load test demonstrates a common case, not a scheduling bound.

The need is to keep unrelated async work responsive during suspension teardown. Run `close_storage()` through the project’s blocking executor (`spawn_blocking` or equivalent), propagating join and storage errors. If executor shutdown semantics make that unsuitable, document and enforce a bounded dedicated teardown thread instead.

The directory-cleanup race, removed error-variant compatibility warning, string-literal rollback dispatch, and timing-test concerns are not actionable from the supplied patch evidence: they are speculative, future-facing, or lack an established failure.

_Advisory audit of head [`b120bb63f9d9`](https://github.com/marmot-protocol/mdk/commit/b120bb63f9d97a8325a4d29da9c33a65909c980f); no approval, merge, push, checkout, or PR-code execution was performed._

crates/storage-sqlite/src/connection.rs
```diff
@@ -82,13 +84,175 @@ fn busy_backoff(attempt: u32) -> Duration {
...
+    /// larger `-wal` behind costs the next open a replay, whereas leaving the
+    /// connection open costs the host its process.
+    pub fn close(&self) -> StorageResult<()> {
+        let mut slot = self
+            .slot
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        // Publish under the mutex, alongside the take, so `is_closed` can never
+        // report open for a slot that is already empty.
+        self.closed.store(true, Ordering::Release);
```
Datawav: **High:** This publishes `closed` only after acquiring `slot`. While `close()` waits behind an active guard, racing `lock()` callers can acquire the mutex first and start more database work, potentially prolonging or starving closure. Publish a closing state before waiting for `slot`, and make `lock()` reject new admissions once closing begins.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3292,6 +3292,47 @@ impl MarmotAppRuntime {
...
+    /// host has just been told is clear. Build a fresh `MarmotApp` and runtime
+    /// to use the root again — which is what a foregrounding app does anyway.
+    ///
+    /// Safe to call twice, and safe to call with or without a preceding
+    /// [`Self::shutdown`]. Worker drain is bounded by
+    /// `APP_RUNTIME_ACCOUNT_SHUTDOWN_WAIT`; the close itself waits only for
+    /// whatever SQLite statement is executing.
+    pub async fn shutdown_and_close(&self) -> Result<(), AppError> {
+        self.shutdown().await;
+        self.accounts.app.close_storage()
```
Datawav: **Low:** `close_storage()` is synchronous and may wait on mutexes, checkpoint WAL files, handle busy databases, and close connections. Calling it directly here blocks a Tokio worker for an interval this API does not bound, reducing unrelated async responsiveness during suspension teardown. Run it through the project's blocking executor and propagate both join and storage errors.

crates/marmot-app/src/lib.rs
```diff
@@ -4088,7 +4108,159 @@ impl MarmotApp {
...
+        // creating its connection after this method returned. Either way the
+        // host would be told the container is lock-free while a lock still
+        // existed. Holding the writer means every opener has either published
+        // (so the drain below closes it) or has not yet checked the flag (so it
+        // will refuse).
+        let _lifecycle = self
+            .storage_lifecycle
+            .write()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        self.storage_closed.store(true, Ordering::Release);
```
erskingardner, line 4165: **High:** The lifecycle gate covers cached database construction, but not every SQLite open that can follow it. A concrete interleaving is: `migrate_legacy_account_projection_if_needed()` gets the cached account storage and observes that the import marker is absent; this method then acquires the writer, drains/closes that storage, releases the root lease, and returns; the migration continues through `legacy_account_projection()` into `LegacyAccountProjectionDb::open()`, which uses a raw `rusqlite::Connection` with no closed check or lifecycle guard. The supposedly terminal close has therefore returned while the old account-projection database is newly open inside the App Group, recreating the exact lock condition this API promises cannot happen. Hold lifecycle admission across the entire migration/open-and-use window (without recursively reacquiring the `RwLock`), or make this legacy handle close-aware and tracked, and add a race test paused after the import-marker read.

erskingardner (commented):
Request changes (posted as COMMENT because GitHub prevents author self-review). The exact head still has two blocking close-boundary races: the previously reported `CloseableConnection` admission race, plus the untracked legacy-database reopen described inline below. Green CI does not exercise either interleaving. The existing async-executor finding is lower severity but also worth addressing.

crates/marmot-app/src/lib.rs
erskingardner: Fixed in 440f358b. Legacy projection migration now takes lifecycle admission after the import-marker check and holds it across the raw legacy connection plus the complete import, so close either wins first and the migration refuses, or waits until the raw connection is gone. The status probes' short-lived raw connections are gated too. Added a deterministic migration race test that pauses on the actual guarded path and verifies close cannot return or release the root lease.

crates/storage-sqlite/src/connection.rs
erskingardner: Fixed in 440f358b. Close now publishes terminal admission before waiting for the connection slot, and lock checks the state both before and after acquiring the mutex. A deterministic regression holds the pre-existing guard and proves a later caller gets StorageError::Closed promptly while the closer waits only for the original operation.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Fixed in 440f358b. MarmotAppRuntime::shutdown_and_close now runs close_storage through the existing blocking_app_task helper, which uses tokio::task::spawn_blocking and preserves both join and storage errors.

crates/marmot-app/src/tests.rs
```diff
@@ -5342,3 +5342,316 @@ async fn an_escalation_recorded_during_a_received_delivery_rides_that_seam() {
...
+        let result = closing_app.close_storage();
+        closed_tx.send(()).unwrap();
+        result
+    });
+    assert!(
+        closed_rx
+            .recv_timeout(std::time::Duration::from_millis(250))
+            .is_err(),
+        "terminal close must wait for the legacy database import window",
+    );
```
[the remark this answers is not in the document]
erskingardner: Fixed in `180a84a2` — the closing thread now signals immediately before calling `close_storage`, and the assertion blocks on that signal, so the 250ms window measures a thread that reached the call rather than one that had not been scheduled. I put the signal in the thread rather than a hook inside the lifecycle-lock acquisition: it removes the same vacuity without adding test scaffolding to the production close path. Verified by mutation that the test still discriminates — dropping the legacy import's `begin_storage_open` admission makes it fail. Applied the same fix to `close_storage_waits_for_an_open_that_is_already_in_flight`, which had the identical gap.

PR #1338, opened 2026-08-09
Add convergence route assurance inventory

erskingardner:
## What changed

This PR closes the route-inventory and bounded-model portion of the convergence reliability plan without changing production convergence behavior.

It:

- inventories the seven production convergence decision routes in a machine-readable form
- adds a human-readable route matrix that records source markers and assurance ownership
- adds reopenable assurance claims so a later green run cannot silently erase a known counterexample
- models independent branch durability and volatile route/provisional-winner state across restart
- adds a pairwise-loser terminalization mutant and expands the mutation catalog
- wires the route-assurance suite into the convergence verification CI recipe
- reconciles the tracking plan with the recently merged scenario-selection and multi-process work

## Why

The simulator and campaign layers now cover several execution adapters, but confidence still depends on knowing which production decision routes are exercised, which invariants own them, and whether failures remain visible after a subsequent passing run. This gives that work an explicit, reviewable contract before adding broader cross-route campaigns.

## Scope and impact

This is simulator, model, test, documentation, and CI-recipe work only. It does not modify the production convergence engine, app runtime, transport behavior, or convergence constants.

The actual cross-route Scenario IR campaign remains a separate follow-up so it can account for the overlapping work in #1293 and remain independently reviewable.

## Validation

- `just fast-ci`
- `cargo test -p cgka-conformance-simulator --test route_assurance --test mutation_adequacy --locked`
- focused clippy for the two new/changed test targets
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Added convergence route assurance tracking, including route ownership, lifecycle states, evidence records, and assurance claim statuses.
  - Added coverage for pairwise-loser terminalization scenarios in convergence validation.

- **Bug Fixes**
  - Improved detection of unresolved counterexamples so assurance claims cannot be closed prematurely.
  - Added crash/restart and route-selection determinism checks.

- **Documentation**
  - Expanded convergence route, mutation coverage, and reliability planning documentation.
  - Added guidance on assurance claims, validation coverage, and known limitations.

- **Tests**
  - Added integration coverage for route ownership, route matrices, claim serialization, and lifecycle behavior.


crates/cgka-conformance-simulator/src/route_assurance.rs
```diff
@@ -0,0 +1,458 @@
...
+    /// Record a green campaign without allowing it to silently close an
+    /// already-known counterexample.
+    pub fn record_passing_evidence(&mut self, evidence_id: impl Into<String>) {
+        insert_sorted_unique(&mut self.passing_evidence, evidence_id.into());
+        if self.status == AssuranceClaimStatus::Open {
+            self.status = AssuranceClaimStatus::Covered;
+        }
+    }
+
+    pub fn reopen(&mut self, evidence_id: impl Into<String>) {
```
erskingardner, line 433: A claim can wedge permanently in `Reopened`.

The dedup check matches any existing entry with the same `evidence_id`, including ones already resolved, but the status assignment below is unconditional:

```
claim.reopen("a");
claim.resolve_falsification("a", "fix-a");   // -> Covered
claim.reopen("a");                            // regression, same id
// no new falsification pushed (dedup hits the resolved entry)
// status = Reopened, but nothing is unresolved
claim.resolve_falsification("a", "fix-a2");   // false: the finder requires resolved_by.is_none()
claim.record_passing_evidence("green");       // no-op: only promotes from Open
```

The claim is now `Reopened` with zero unresolved falsifications and no reachable transition back to `Covered`. A counterexample recurring under a stable id after a fix is the ordinary case for this record, not an exotic one.

Either restrict the dedup to unresolved entries (so a recurrence pushes a fresh unresolved falsification), or clear `resolved_by` on the matched entry when reopening. A test that reopens an already-resolved id and then resolves it again would pin whichever you pick.

erskingardner (commented):
**Review verdict: changes requested** (posted as a comment because GitHub does not allow requesting changes on your own PR).

The Markdown route matrix and the source-marker drift gate are the genuinely valuable part of this PR — a reviewable index of which production seam owns which convergence decision, with `partial` limitations written down honestly, is worth having. Most of the inventory is calibrated well.

Two things should be fixed before this lands, both in the machine-readable half.

**1. The one route promoted to `covered` is certified by a model added in the same PR whose invariant cannot fail.** `PairwiseForkRecovery` gets `reference_model: covered("route_assurance::RouteLifecycleStateV1")` and `mutation_sentinel: covered("pairwise_losing_branch_terminalization")`. But `RouteLifecycleStateV1::settle()` reads only `durable_branches` and never touches `volatile_route`/`volatile_provisional_winner`, so `route_choice_and_restart_do_not_change_the_closed_input_winner` is true by construction of a 20-line function regardless of what the engine does. The mutant is a sibling method in the same file, killed by construction too. Unlike `lifecycle_model.rs` — which has a Stateright `Model` impl with declared properties, exhaustive bounded exploration, a TLA+ counterpart, and Scenario IR trace projection — this new type has none of that and is exercised at exactly one hand-written two-branch input. That is the assurance inflation this file's own preamble warns against: "an inventory must not turn the existence of adjacent evidence into a stronger route-equivalence claim than that evidence supports." Either downgrade both to `partial` with the residual limitation spelled out, or ground the model in the existing Stateright harness.

**2. The claim lifecycle wedges permanently when a resolved counterexample recurs.** `reopen()` dedups on `evidence_id` while ignoring `resolved_by`, but still forces `Reopened`. Resolve a falsification, then see it again under the same id, and the claim is stuck in `Reopened` with nothing left to resolve. Details inline. Recurrence-after-fix is the normal case for this kind of record, so it is worth getting right.

**Reuse.** The plan-doc checkbox flipped to `[x]` says "Extend the independent selector/lifecycle models…", but nothing was extended. `lifecycle_model.rs` already models durable-vs-volatile state across crash/restart and already has the `LifecycleMutation` + `next_state_with_mutation` pattern for simulator-only mutants; `reference_convergence::{score, compare, select}` already implements depth-then-ordering-key selection. This PR adds a third selection implementation and a second, divergent mutation mechanism next to them. Folding `RouteBranch`/route choice into `LifecycleState` and adding a `LifecycleMutation::TerminalizePairwiseLoser` variant would land the same coverage, remove the duplicate comparator, and make the `covered` claim in point 1 defensible — while deleting most of the ~120 lines of new model code.

**Speculative surface.** `AssuranceClaimRecordV1`'s open/reopen/resolve API, `open_route_assurance_claims()`, `ROUTE_ASSURANCE_SCHEMA_VERSION`, the `Serialize` derives on the inventory types, and `AssuranceOwnershipStatus::Gap` (no constructor, no entry uses it) have no consumer outside their own test. The claim lifecycle is a real design worth having when a campaign writes to it; today it is ~90 lines of code plus tests that only exercise themselves. Consider deferring it to the PR that first records a claim, and landing the inventory, matrix, marker gate, and mutant now.

Non-blocking notes on marker strength, matrix drift-test scope, and the missing claim-id drift test are inline.

crates/cgka-conformance-simulator/src/route_assurance.rs
```diff
@@ -0,0 +1,458 @@
...
+        campaign: AssuranceOwnerV1::partial(
+            "convergence-committer-selected/v1 and convergence-witness-selected/v1",
+            "The fixed vectors cover observer ingest but not equivalence with every other decision route.",
+        ),
+    },
+    DecisionRouteInventoryEntryV1 {
+        route: DecisionRouteId::PairwiseForkRecovery,
+        production_sites: PAIRWISE_FORK_RECOVERY_SITES,
+        adopted_rule: "A pairwise winner is provisional and every authenticated eligible loser remains reconsiderable by the shared closed-input selector.",
+        reference_model: AssuranceOwnerV1::covered("route_assurance::RouteLifecycleStateV1"),
```
erskingardner: This is the only `covered` reference-model entry in the inventory that points at a model introduced by this same PR, and the model's invariant holds by construction (see `settle()` below and the review summary).

Every other `covered` entry points at something with independent grounding: `lifecycle_model` crash/restart transitions run through a Stateright `Model` with exhaustive bounded exploration and a TLA+ counterpart; `reference_convergence` dispositions run through the shared evaluator that engine vectors compare against. `RouteLifecycleStateV1` has neither, and the paired `mutation_sentinel: covered("pairwise_losing_branch_terminalization")` on the next line is a mutant defined ten lines away from the rule it mutates.

Suggest `partial` for both until the model is folded into the Stateright harness, with a limitation along the lines of "models route-choice volatility abstractly; not yet an exhaustively explored transition system and not compared against production routing."

crates/cgka-conformance-simulator/src/route_assurance.rs
```diff
@@ -0,0 +1,458 @@
...
+        self.volatile_provisional_winner = None;
+        self
+    }
+
+    pub fn restart(mut self) -> Self {
+        self.crashed = false;
+        self
+    }
+
+    pub fn settle(mut self) -> Self {
```
erskingardner, line 104: `settle()` reads only `durable_branches`. `crash()` clears the volatile fields, and `restart()` flips a bool. So `route_choice_and_restart_do_not_change_the_closed_input_winner` asserts that a function which structurally cannot observe route or provisional-winner state is unaffected by route or provisional-winner state — it cannot fail for any engine behavior, correct or not.

The comparator here (`max_by` on `effective_depth`, then reversed `ordering_key`) is also a third implementation of a rule `reference_convergence::compare` already encodes, with a different tie-break surface: on a full tie of depth and ordering key, `max_by` returns the last maximum, which after the `sort_by_key(id)` in `new()` silently means highest id wins. Undocumented, and divergent from `compare`'s explicit terminal tie-breaks.

Extending `LifecycleState` with the branch set and route choice, rather than adding a parallel type, would let this reuse the existing comparator and put the invariant somewhere it can actually fail.

crates/cgka-conformance-simulator/src/route_assurance.rs
```diff
@@ -0,0 +1,458 @@
...
+    },
+    ProductionDecisionSiteV1 {
+        path: "crates/cgka-engine/src/openmls_projection.rs",
+        marker: "MissingRetainedAnchor",
+    },
+];
+const CRASH_RESTART_RECOVERY_SITES: &[ProductionDecisionSiteV1] = &[
+    ProductionDecisionSiteV1 {
+        path: "crates/cgka-engine/src/engine.rs",
+        marker: "hydrate",
```
erskingardner: `hydrate` is far too weak to gate anything. It occurs 73 times in `engine.rs` at this SHA, almost all of it unrelated to the crash/restart decision route (`unhydrated_groups`, `retry_hydrate_quarantined_group`, doc comments). The named route could be deleted outright and this assertion would still pass.

Use the specific entry point instead — `hydrate_stable_groups_from_storage` or `hydrate_one_stored_group`. Every other marker in the inventory is already a specific symbol; this is the outlier.

Separately, the test substring-matches the whole file including comments and `#[cfg(test)]` blocks, so a marker surviving only in a comment satisfies the gate. Matching `fn {marker}(` for function markers would make the drift signal real.

crates/cgka-conformance-simulator/src/route_assurance.rs
```diff
@@ -0,0 +1,458 @@
...
+    pub path: &'static str,
+    pub marker: &'static str,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum AssuranceOwnershipStatus {
+    Covered,
+    Partial,
+    Gap,
```
erskingardner, line 144: `Gap` has no constructor (only `covered()` and `partial()` exist) and no inventory entry uses it, so it is unreachable today — the test's match arm for it and the matrix preamble's "a `partial` or `gap` entry" both describe a state that cannot occur. Drop it until a route actually needs it, or add the `gap()` constructor alongside the others.

crates/cgka-conformance-simulator/tests/route_assurance.rs
```diff
@@ -0,0 +1,158 @@
...
+        assert_eq!(restarted.volatile_provisional_winner, None);
+    }
+}
+
+#[test]
+fn human_route_matrix_has_exactly_one_row_per_registered_route() {
+    let matrix = include_str!("../CONVERGENCE_ROUTE_MATRIX.md");
+    for route in DecisionRouteId::ALL {
+        assert_eq!(
+            matrix.matches(&format!("`{}`", route.as_str())).count(),
```
erskingardner: `matches(...).count() == 1` scans the whole file, so no prose in `CONVERGENCE_ROUTE_MATRIX.md` may ever name a route id in backticks — the "Current limitation" and "Active production work" sections are permanently barred from referring to a route by name, which is an odd constraint for a document whose purpose is human explanation. Counting only rows (lines starting with `` | ` ``, which the assertion below already isolates) keeps the drift check and frees the prose.

Also: the `## Assurance claims` bullet list in the same file duplicates `AssuranceClaimId::ALL` with no drift test, unlike the route table. Since `AssuranceClaimId` already carries `#[serde(rename_all = "kebab-case")]` matching the bullets exactly, that check is a couple of lines and closes the one remaining silent-drift path in this doc.

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -995,7 +997,7 @@ engine repair or scaling the existing corpus is sufficient.
   into a permanent synthetic regression family with at least four participants, simultaneous same-source-epoch
   committers, committer-versus-observer routing, ordering-key-versus-depth disagreement, branch growth, application
   traffic, delivery permutations, and restarts around every durable transition.
-- [ ] Extend the independent selector/lifecycle models and mutation matrix with route choice, reconsiderable versus
+- [x] Extend the independent selector/lifecycle models and mutation matrix with route choice, reconsiderable versus
```
erskingardner: "Extend the independent selector/lifecycle models" — nothing was extended. `lifecycle_model.rs` and `reference_convergence.rs` are untouched; a parallel model was added beside them. Given that this checkbox is what the milestone is tracked against, either fold the route state into `LifecycleState` (preferred — see the review summary) or reword the item so the plan does not record reuse that did not happen.

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -798,15 +798,17 @@ is reset/rejoin with fresh state. Tamarin's scope is unchanged.
...
-The macro-generated `SemanticMutation::ALL` catalog contains nine one-rule simulator mutants with minimal deterministic
+The macro-generated `SemanticMutation::ALL` catalog contains eleven one-rule simulator mutants with minimal deterministic
 witnesses. The campaign requires every mutant to change the adopted observation; none is a production compile feature.
 It kills raw-depth-first selection, non-deduplicated witnesses, witness removal, exclusive cutoff, crash-lost durable
 frozen state, suppressed post-settlement re-arm, retained losing output, uncleared accepted publication, and inclusive
-expiration. Frozen-state and scheduler mutants execute alternate transitions through the shared Rust lifecycle model;
-the publication mutant drives a real `ReferenceModelSubject` and compares structural pending work before and after
-acknowledgement. The machine-checked
+expiration, as well as stale group-profile projection and terminalized pairwise losers. Frozen-state, scheduler, and
+route-choice mutants execute alternate transitions through shared independent Rust lifecycle models; the publication
```
erskingardner: "route-choice mutants execute alternate transitions through shared independent Rust lifecycle models" is not accurate for the new mutant. Frozen-state and scheduler mutants do go through the shared `lifecycle_model` via `next_state_with_mutation`; `pairwise_losing_branch_terminalization` calls a bespoke `settle_with_terminal_pairwise_loser()` on a type that nothing else uses. If the model is not folded into `lifecycle_model`, this sentence should say so plainly.

PR #1340, opened 2026-08-09
fix(app): close invite catch-up lifecycle races

erskingardner:
## What was broken

Two late automated review findings landed on #1309 after its final CI run had started:

- invite-triggered catch-up tasks dropped their `JoinHandle`, so runtime shutdown did not coordinate their completion;
- the reciprocal-invite relay test published its blocked counter before registering the `Notify` waiter, leaving a small missed-wakeup race.

## Root cause and impact

The detached catch-up was lifecycle-aware only indirectly through worker shutdown. A runtime could therefore finish its own shutdown path while a catch-up task still held cloned runtime services. Separately, `notify_waiters()` stores no permit, so a fast test release could occur between the counter increment and waiter registration and hang the test.

## Changes

- register invite catch-up handles before returning from the mutation path;
- stop accepting new catch-up tasks once shutdown begins, cancel in-flight catch-ups from the shared lifecycle signal, and await their handles alongside account workers;
- register the relay-test release waiter before publishing the blocked counter;
- keep one catch-up per successful invite instead of coalescing, preserving the reciprocal/mixed-result delivery guarantee from #1309.

## Validation

- `just fast-ci`
- `cargo test -p marmot-app --test relay_runtime overlapping_reciprocal_invites_deliver_both_incoming_welcomes --locked -- --exact`
- `cargo test -p marmot-app --test relay_runtime successful_invite_delivers_while_overlapping_invite_fails --locked -- --exact`
- `cargo test -p marmot-app --lib tests::invite_members_keeps_same_account_projection_reads_off_detached_catch_up --locked -- --exact`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved invite synchronization during shutdown to prevent missed updates and ensure in-progress tasks finish cleanly.
  * Prevented new invite catch-up operations from starting once shutdown begins.
  * Fixed event notification handling to reliably preserve notifications that occur before a waiting task is ready.

### DIFF
--- a/crates/marmot-app/src/runtime/commands.rs
+++ b/crates/marmot-app/src/runtime/commands.rs
@@ -13,7 +13,7 @@ use tokio::sync::oneshot;
 
 use super::{
     AccountManager, AccountWorkerCommand, account_worker_response, group_contributes_co_members,
-    publish_app_runtime_group_state_updated,
+    publish_app_runtime_group_state_updated, wait_for_runtime_shutdown,
 };
 use crate::app_telemetry::AppPerformanceOperation;
 use crate::messages::AppMessageIntent;
@@ -29,9 +29,21 @@ use crate::{
 impl AccountManager {
     fn spawn_invite_catch_up(&self) {
         let post_mutation = self.clone();
-        tokio::spawn(async move {
+        let mut stopping = self.shared.lifecycle().subscribe_shutdown();
+        let mut tasks = self
+            .invite_catch_up_tasks
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        tasks.handles.retain(|task| !task.is_finished());
+        if !tasks.accepting {
+            return;
+        }
+        let handle = tokio::spawn(async move {
             let catch_up_started_at = Instant::now();
-            let catch_up = post_mutation.catch_up_accounts().await;
+            let catch_up = tokio::select! {
+                _ = wait_for_runtime_shutdown(&mut stopping) => return,
+                catch_up = post_mutation.catch_up_accounts() => catch_up,
+            };
             post_mutation.shared.app_performance_telemetry().record(
                 AppPerformanceOperation::GroupInvitePostMutationCatchUp,
                 catch_up_started_at.elapsed(),
@@ -46,6 +58,7 @@ impl AccountManager {
                 );
             }
         });
+        tasks.handles.push(handle);
     }
 
     /// Refresh other account workers after an irreversible mutation without
--- a/crates/marmot-app/src/runtime/commands.rs
+++ b/crates/marmot-app/src/runtime/commands.rs
@@ -13,7 +13,7 @@ use tokio::sync::oneshot;
 
 use super::{
     AccountManager, AccountWorkerCommand, account_worker_response, group_contributes_co_members,
-    publish_app_runtime_group_state_updated, wait_for_runtime_shutdown,
+    publish_app_runtime_group_state_updated,
 };
 use crate::app_telemetry::AppPerformanceOperation;
 use crate::messages::AppMessageIntent;
@@ -27,9 +27,7 @@ use crate::{
 };
 
 impl AccountManager {
-    fn spawn_invite_catch_up(&self) {
-        let post_mutation = self.clone();
-        let mut stopping = self.shared.lifecycle().subscribe_shutdown();
+    pub(super) fn spawn_invite_catch_up(&self) {
         let mut tasks = self
             .invite_catch_up_tasks
             .lock()
@@ -38,12 +36,13 @@ impl AccountManager {
         if !tasks.accepting {
             return;
         }
+        let post_mutation = self.clone();
         let handle = tokio::spawn(async move {
             let catch_up_started_at = Instant::now();
-            let catch_up = tokio::select! {
-                _ = wait_for_runtime_shutdown(&mut stopping) => return,
-                catch_up = post_mutation.catch_up_accounts() => catch_up,
-            };
+            // Once started, catch-up must run to a normal result. Cancelling
+            // reconcile mid-await could abandon stale workers that it already
+            // removed from the manager map but has not finished reaping.
+            let catch_up = post_mutation.catch_up_accounts().await;
             post_mutation.shared.app_performance_telemetry().record(
                 AppPerformanceOperation::GroupInvitePostMutationCatchUp,
                 catch_up_started_at.elapsed(),
--- a/crates/marmot-app/src/runtime/mod.rs
+++ b/crates/marmot-app/src/runtime/mod.rs
@@ -122,6 +122,12 @@ pub struct AccountManager {
     shared: RuntimeSharedServices,
     workers: Arc<Mutex<HashMap<String, ManagedAccountWorker>>>,
     tearing_down: Arc<StdMutex<HashSet<String>>>,
+    invite_catch_up_tasks: Arc<StdMutex<InviteCatchUpTasks>>,
+}
+
+struct InviteCatchUpTasks {
+    accepting: bool,
+    handles: Vec<JoinHandle<()>>,
 }
 
 #[derive(Clone)]
@@ -3306,6 +3312,10 @@ impl AccountManager {
             shared,
             workers: Arc::new(Mutex::new(HashMap::new())),
             tearing_down: Arc::new(StdMutex::new(HashSet::new())),
+            invite_catch_up_tasks: Arc::new(StdMutex::new(InviteCatchUpTasks {
+                accepting: true,
+                handles: Vec::new(),
+            })),
         }
     }
 
@@ -4634,6 +4644,14 @@ impl AccountManager {
 
     pub async fn shutdown(&self) {
         self.shared.lifecycle().begin_shutdown();
+        let invite_catch_up_tasks = {
+            let mut tasks = self
+                .invite_catch_up_tasks
+                .lock()
+                .unwrap_or_else(|poisoned| poisoned.into_inner());
+            tasks.accepting = false;
+            std::mem::take(&mut tasks.handles)
+        };
         let workers = {
             let mut workers = self.workers.lock().await;
             workers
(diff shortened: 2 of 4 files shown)


crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4647,6 +4665,11 @@ impl AccountManager {
                 worker.shutdown().await;
             });
         }
+        for task in invite_catch_up_tasks {
+            shutdowns.spawn(async move {
+                let _ = task.await;
+            });
+        }
```
erskingardner: **Blocking:** join invite catch-up tasks before the final worker drain (or drain `self.workers` again afterward).

`catch_up_accounts()` → `reconcile()` can pass `ensure_running()` before `begin_shutdown()`, then run a long synchronous stretch (especially with no stale workers) that inserts replacement workers before the next await. With the current order — drain workers, then await catch-ups in the same `JoinSet` — those replacements land in the map after the drain and remain live when `shutdown()` returns.

Preferred shape:

```rust
self.shared.lifecycle().begin_shutdown();
let invite_catch_up_tasks = { /* accepting = false; take handles */ };
for task in invite_catch_up_tasks {
    let _ = task.await; // select! cancels in-flight catch-up
}
let workers = { /* drain */ };
// then shut down workers
```

A second drain after joining catch-ups also works. Please add a regression for catch-up/`reconcile` overlapping shutdown so this doesn’t regress.

erskingardner (commented):
**Review verdict: request changes** (posted as comment — GitHub disallows request-changes on own PRs)

The invite catch-up tracking and the relay `Notify` waiter fix are the right direction, and keeping one catch-up per invite matches #1309’s delivery guarantee. The remaining hole is that `shutdown()` drains workers before (and in parallel with) joining catch-up tasks, so a catch-up that already passed `reconcile()`’s `ensure_running()` check can still insert replacement workers after the drain. That leaves live account workers after shutdown returns — the same class of lifecycle leak this PR is trying to close.

## Blocking
- Await invite catch-up handles **before** the final worker drain (or drain workers again afterward).

## Non-blocking
- Add a regression covering stale-worker replacement / catch-up overlapping shutdown.
- Relay test waiter registration looks correct and matches the existing `account_opens_drained` pattern.

crates/marmot-app/src/runtime/commands.rs
```diff
@@ -29,9 +29,21 @@ use crate::{
...
+        if !tasks.accepting {
+            return;
+        }
+        let handle = tokio::spawn(async move {
             let catch_up_started_at = Instant::now();
-            let catch_up = post_mutation.catch_up_accounts().await;
+            let catch_up = tokio::select! {
+                _ = wait_for_runtime_shutdown(&mut stopping) => return,
+                catch_up = post_mutation.catch_up_accounts() => catch_up,
+            };
```
erskingardner: **Blocking:** this cancels `catch_up_accounts()` at whatever await point it happens to be parked on, and at least one of those points is not safe to abandon.

`catch_up_accounts()` -> `reconcile()` (mod.rs:3573) does:

```rust
// Worker task teardown releases AppClient's account-session guard.
// Reap stale tasks before opening replacements for the same labels.
for worker in stale_workers {
    worker.shutdown().await;
}
```

Those workers were already `remove`d from `self.workers` a few lines earlier, so they exist only on `reconcile`'s stack. If the shutdown branch wins the `select!` while that loop is parked, the remaining `stale_workers` are dropped without an awaited teardown — and `AccountManager::shutdown` can never see them, because they are no longer in the map. `ManagedAccountWorker::shutdown_with_timeout` is explicit that this is the ownership handoff:

```rust
// Reaping is the ownership handoff: the task owns AppClient,
// whose drop releases the account-session guard. Do not return
// until cancellation has run its destructors, ...
```

So `shutdown_and_close` can return while an `AppClient` is still dropping and its SQLCipher file lock is still held — the failure mode `docs/marmot-architecture/overview/local-artifact-safety.md` and dcf27f22 exist to prevent.

Window: an invite catch-up must be inside the stale-worker reap when `begin_shutdown()` lands. Narrow, but it is the same shape of race this PR is fixing.

Suggest dropping the `select!` entirely and letting the tracked handle carry the coordination:

```rust
let handle = tokio::spawn(async move {
    let catch_up_started_at = Instant::now();
    let catch_up = post_mutation.catch_up_accounts().await;
    ...
});
```

The promptness cost is small and already bounded: `catch_up_accounts` and `reconcile` both open with `ensure_running()?`, so a catch-up that has not started yet exits immediately; and because `shutdown` spawns worker teardown *concurrently* with the handle awaits, worker exit drops the `respond` oneshots and the in-flight catch-up falls out with `TransportClosed` rather than sitting on the 45s `APP_RUNTIME_ACCOUNT_READY_WAIT`. Worst case is roughly the 5s `APP_RUNTIME_ACCOUNT_SHUTDOWN_WAIT` already budgeted for worker teardown.

That also removes the `stopping` subscription and the `wait_for_runtime_shutdown` import — a net deletion for the same guarantee. If you do want a hard bound, prefer wrapping the *await side* in `shutdown()` with a timeout over cancelling the task mid-teardown.

crates/marmot-app/src/runtime/commands.rs
```diff
@@ -29,9 +29,21 @@ use crate::{
...
-        tokio::spawn(async move {
+        let mut stopping = self.shared.lifecycle().subscribe_shutdown();
+        let mut tasks = self
+            .invite_catch_up_tasks
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        tasks.handles.retain(|task| !task.is_finished());
+        if !tasks.accepting {
+            return;
+        }
```
erskingardner: Nit: `self.clone()` and `subscribe_shutdown()` both happen before the `accepting` check, so the shutdown path pays for a runtime clone and a watch subscription it throws away. Moving the `if !tasks.accepting { return; }` above them costs nothing (and if the `select!` goes away per the other comment, the subscription disappears with it).

erskingardner (commented):
**Requesting changes** (posted as a comment — GitHub blocks request-changes on your own PR).

Right problem, and the tracking half is clean. The `Notify` fix in `relay_runtime.rs` is textbook and mirrors the existing `wait_for_account_opens_to_drain` pattern.

My one blocking concern is the `tokio::select!`: it converts the detached catch-up from "runs to completion, uncoordinated" into "cancelled at an arbitrary await point", and one of those await points is `reconcile()`'s stale-worker reaping loop — the exact place two in-tree comments say must not be abandoned. Closing a lifecycle race by opening a narrower one in the same path is a bad trade, especially when the tracking + awaiting you already added is sufficient on its own.

Secondary: the primary behavior this PR adds (shutdown refuses new catch-ups and awaits in-flight ones) has no test. `runtime/tests.rs` already unit-tests lifecycle predicates directly (`lifecycle_refuses_account_open_after_shutdown_begins`, `managed_account_worker_shutdown_aborts_unresponsive_task_after_timeout`), so there is a cheap precedent to follow. The three tests in the Validation section all predate this change and pass with or without it.

Nits aside, dropping the `select!` is a net deletion and leaves the fix strictly simpler. Details inline.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Non-blocking: no test covers either half of this — that `spawn_invite_catch_up` refuses to spawn once `accepting` is false, or that `shutdown` actually awaits a tracked handle before returning. All three tests in the PR's Validation section predate this change and pass either way, so nothing here is currently pinned against regression.

`runtime/tests.rs` already exercises this layer directly (`lifecycle_refuses_account_open_after_shutdown_begins`, `managed_account_worker_shutdown_aborts_unresponsive_task_after_timeout`), so a small test in the same style — park a tracked handle on a channel, call `shutdown`, assert it does not return until the handle completes — would be a handful of lines and would make the fix verifiable rather than argued.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4634,6 +4644,14 @@ impl AccountManager {
...
     pub async fn shutdown(&self) {
         self.shared.lifecycle().begin_shutdown();
+        let invite_catch_up_tasks = {
+            let mut tasks = self
+                .invite_catch_up_tasks
+                .lock()
+                .unwrap_or_else(|poisoned| poisoned.into_inner());
+            tasks.accepting = false;
+            std::mem::take(&mut tasks.handles)
+        };
```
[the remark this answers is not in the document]
erskingardner: Fixed in d410b06b. AccountManager shutdown now awaits every tracked invite catch-up before the final worker-map drain, so a reconcile that already passed its lifecycle check cannot leave a replacement worker behind. Added a deterministic regression that inserts a replacement worker from an in-flight catch-up during shutdown and verifies the replacement is shut down and the map is empty.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Fixed in d410b06b with the requested ordering: tracked catch-ups are joined before the final worker drain. The new overlap regression proves a worker inserted by an already-running catch-up is still drained before shutdown returns. Focused test and just fast-ci both pass.

crates/marmot-app/src/runtime/commands.rs
erskingardner: Fixed in 9880985c. Removed the shutdown select entirely: once spawned, catch_up_accounts now runs to a normal result, so reconcile cannot be cancelled after removing stale workers but before awaiting their teardown. AccountManager shutdown still prevents new tasks, awaits tracked catch-ups, then performs the final worker drain.

crates/marmot-app/src/runtime/commands.rs
erskingardner: Addressed in 9880985c. The accepting check now happens before cloning AccountManager; the shutdown subscription disappeared with the cancellation select.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Added coverage across d410b06b and 9880985c: one deterministic test proves shutdown waits for a tracked catch-up, drains the worker it inserts, and returns with an empty map; a second focused test proves spawn_invite_catch_up refuses to enqueue after accepting is false. Both focused tests and just fast-ci pass.

PR #1341, opened 2026-08-09
Restore sender ratchet reorder policy

erskingardner:
## What changed

This restores the explicit OpenMLS sender-ratchet policy used by MDK:

- out-of-order tolerance: 100
- maximum forward distance: 1000

New groups and Welcome joins now use the pinned policy. Existing persisted groups are normalized during hydration and the updated OpenMLS configuration is persisted in place. The convergence constant inventory, conformance snapshot, policy contract, and tracking plan now record both values.

## Why

Generated campaign seed 2001 cases 2 and 13 exposed a concrete regression: six application messages were delivered in reverse-enough order that generation 5 advanced the ratchet and generation 0 was then rejected as too far in the past. MDK previously configured a tolerance of 100, but that explicit configuration was lost and OpenMLS therefore used its default tolerance of 5.

This is an engine behavior change, but it is not an MLS wire-format or group-format change. It retains more skipped application-message secrets and accepts a wider reorder window. Existing groups are migrated on load. Secrets that OpenMLS already pruned before this change cannot be recovered, so already-expired messages may remain undecryptable.

This PR intentionally does not add mixed old/new application-witness compatibility machinery. The current testing-stage policy is to restore the intended retention window directly and measure its behavior through the simulator.

## Validation

- just fast-ci
- cargo test -p cgka-engine --test sqlite_storage hydration_migrates_and_persists_existing_sender_ratchet_policy --locked
- cargo test -p cgka-conformance-simulator --test canonical_scenarios sender_ratchet_policy_preserves_reordered_rollback_floods_past_openmls_default --locked
- cargo test -p cgka-conformance-simulator --test protocol_decision_gate --locked
- cargo test -p cgka-engine --features test-policy-overrides --locked
- full canonical_scenarios suite: 42 passed

The focused regression verifies that both previously failing generated cases now deliver all six reordered application messages.

## Summary by CodeRabbit

- **New Features**
  - Added sender-ratchet defaults supporting up to 100 out-of-order messages and a 1,000-message forward distance.
  - New groups automatically use these settings, and existing groups are updated when reopened.

- **Bug Fixes**
  - Improved recovery of persisted groups with outdated sender-ratchet settings.
  - Preserved application messages during reordered message floods and rollback scenarios.

- **Documentation**
  - Updated convergence and wire-format documentation with the new policy constants and verification requirements.


erskingardner (commented):
Verified the mechanism against the vendored OpenMLS 0.8.1 source and it holds up:

- `MlsGroup::set_configuration` (`group/mls_group/mod.rs:273-280`) writes the join-config row, and `process_message` re-reads `self.configuration().sender_ratchet_configuration()` live on every decrypt (`group/mls_group/processing.rs:175`). So the hydration write really does take effect for the already-instantiated epoch, not just for future ones.
- `SenderRatchet::prune_past_secrets` truncates `past_secrets` at advance time using the *passed* config (`tree/sender_ratchet.rs:195-197, 253`), which is exactly why previously-pruned generations are unrecoverable — the PR description's caveat is precisely right.
- The new `wire_format.rs` doc claim checks out: the incoming wire-format policy check is gated on `message.is_handshake_message()` (`processing.rs:162-171`), so application `PrivateMessage`s bypass `PURE_PLAINTEXT`'s `IncomingWireFormatPolicy::AlwaysPlaintext` and the sender ratchet genuinely is the governing window.
- The "restore" framing is accurate: `7db68b54` (#155) set 100/1000 and `9db9dbc0` (Dark Matter workspace replacement) dropped it.

Premise is solid — concrete reproducing regression, prior documented behavior, tight scope (+245/-22, most of it the ledger/doc churn the `check_convergence_constant_ledger.sh` gate mandates), and both group-construction paths (`MlsGroup::new` at `group_lifecycle.rs:432`, `ProcessedWelcome::new_from_welcome` at `:987`) are covered. The within-epoch forward-secrecy cost of retaining 20x more skipped message keys is a real tradeoff, but it is deliberate, documented in the P9 row, and it is the behavior MDK shipped before.

Four non-blocking notes inline — the one I'd most like addressed is that the root cause (two hand-maintained config builders that can drift) is still open, and the create-path half of the fix has no direct assertion.

crates/cgka-engine/src/group_lifecycle.rs
```diff
@@ -422,6 +424,7 @@ impl<S: StorageProvider> Engine<S> {
             .max_past_epochs(self.max_past_epochs)
             .with_group_context_extensions(gc_exts)
             .use_ratchet_tree_extension(true)
+            .sender_ratchet_configuration(default_sender_ratchet_configuration())
```
erskingardner, line 427: Non-blocking, but this is the root cause the PR steps around rather than closes.

The bug exists because `MlsGroupCreateConfig::builder()` here and `wire_format::join_config` maintain the *same* set of join settings independently — `wire_format_policy`, `max_past_epochs`, `use_ratchet_tree_extension`, and now `sender_ratchet_configuration`. The fix adds a fourth duplicated field to both lists. The next setting someone adds to one and not the other reproduces this class of bug exactly.

OpenMLS 0.8 gives `MlsGroupCreateConfigBuilder` no `join_config()` setter, so a single shared builder isn't available upstream. The cheap structural guard is an equality assertion instead: create a group and assert its persisted `configuration()` matches `join_config(self.max_past_epochs)` field-for-field. `MlsGroupJoinConfig` derives `PartialEq`, so that's a one-line comparison and it fails loudly on any future drift, not just on the sender-ratchet field.

crates/cgka-engine/tests/sqlite_storage.rs
```diff
@@ -186,3 +202,93 @@ async fn create_group_confirm_and_reopen_with_encrypted_sqlite_storage() {
...
+        })
+        .await
+        .expect("create group");
+    let pending = match result {
+        SendResult::GroupCreated { pending, .. } => pending,
+        other => panic!("expected group creation result, got {other:?}"),
+    };
+    alice.confirm_published(pending).await.expect("confirm");
+
+    let mut stored_group = load_openmls_group(&alice_inspection, &group_id);
```
erskingardner, line 237: The create-path half of the fix (`group_lifecycle.rs:427`) has no direct assertion anywhere in this PR. The new `wire_format.rs` unit test pins `join_config()`, which is the *welcome* path; this test loads the freshly created group and immediately overwrites its config, so it never observes what creation produced. That leaves `MlsGroupCreateConfig` covered only by the seed-2001 whole-system scenario.

One line here closes it, right before the overwrite:

```rust
assert_eq!(
    stored_group
        .configuration()
        .sender_ratchet_configuration()
        .out_of_order_tolerance(),
    DEFAULT_OUT_OF_ORDER_TOLERANCE,
    "group creation must persist the pinned policy"
);
```

Separately: since `DEFAULT_MAXIMUM_FORWARD_DISTANCE` already equals OpenMLS's default of 1000, `accidentally_inherited_config` differs from the required config in `out_of_order_tolerance` only. The migration assertion for `maximum_forward_distance` therefore passes trivially and doesn't actually exercise migration of that field. Not worth contorting the test over, just noting the coverage is narrower than it reads.

crates/cgka-engine/src/engine.rs
```diff
@@ -1535,12 +1535,28 @@ impl<S: StorageProvider> Engine<S> {
...
+                .ok_or(GroupHydrationQuarantineReason::OpenMlsGroupMissing)?;
+
+            // OpenMLS persists sender-ratchet policy per group. Groups created
+            // while MDK accidentally inherited OpenMLS's 5-message default
+            // would therefore remain fragile after merely changing the
+            // creation/join presets. Normalize the complete engine-owned join
+            // config before any new traffic can be processed and write it
+            // through in place. This cannot restore secrets already pruned,
+            // but it protects all future within-epoch application traffic.
+            let required_config = crate::wire_format::join_config(self.max_past_epochs);
```
erskingardner: The comment says "Normalize the complete engine-owned join config," but the code is asymmetric in both directions, and one half of it is silently a no-op.

**The guard is narrower than the write.** It compares only `sender_ratchet_configuration()`, then writes the whole `required_config`. A group whose `wire_format_policy` or `use_ratchet_tree_extension` had drifted — the other fields the twin builders can disagree on — is never normalized, because the ratchet field alone decides whether the write happens.

**`max_past_epochs` normalization does not do what it looks like.** `set_configuration` (OpenMLS `group/mls_group/mod.rs:273-280`) only clones the config and writes the join-config row; it never calls `set_max_past_epochs`, which is what actually resizes `MessageSecretsStore` (`mod.rs:521-522`, invoked from `creation.rs:573`). So if an engine ever opens a stored group with a different `self.max_past_epochs` than the one it was created under, this writes a config row claiming the new value while the live secrets store keeps the old capacity — a quiet disagreement between the persisted config and actual retention.

Today both construction paths pass `self.max_past_epochs` so the values agree in practice and nothing breaks. But the comment overclaims what is being normalized. Either narrow it to state that only the sender-ratchet field is normalized and that the rest of `required_config` rides along unchanged, or widen the guard to compare the full config so the claim becomes true.

crates/cgka-engine/src/engine.rs
```diff
@@ -1535,12 +1535,28 @@ impl<S: StorageProvider> Engine<S> {
...
+            if mls_group.configuration().sender_ratchet_configuration()
+                != required_config.sender_ratchet_configuration()
+            {
+                mls_group
+                    .set_configuration(provider_storage, &required_config)
+                    .map_err(|_| GroupHydrationQuarantineReason::OpenMlsLoadFailed)?;
```
erskingardner: This widens the quarantine surface in a way worth an explicit decision rather than a doc tweak.

Before this PR, a group that loaded successfully could not be quarantined by this arm. Now a transient storage write failure during normalization quarantines an otherwise perfectly healthy group — and per the crate's own contract that means it vanishes from every live surface via `ensure_group_live`, returns `UnknownGroup` to all accessors, refuses sends and convergence drains, and stays that way until `retry_hydrate_quarantined_group`. The degraded alternative is strictly milder: the group keeps working, just with the old fragile 5-message tolerance, and the next hydration retries the write.

Fail-closed is a defensible reading of the repo's posture, so I'm not asking you to change it — but two smaller things follow from it:

1. The app-layer mirror at `crates/marmot-app/src/groups.rs:323` still documents `OpenMlsLoadFailed` as load-only, and its type doc says it "Mirrors [`GroupHydrationQuarantineReason`] one-to-one." You updated the trait-side doc; this one drifted.
2. The write lands *before* `validate_current_profile_group_invariants` and the member-credential checks below, so a group that is about to be quarantined for `MemberValidationFailed` still gets a config write first. The write is idempotent and harmless, but it does invert the repo's "validate before mutating" ordering from `docs/marmot-architecture/overview/multi-step-state-changes.md`. Moving the normalization below the validation block would restore that ordering at no cost — nothing between here and there reads the ratchet config.

crates/cgka-engine/src/engine.rs
```diff
@@ -1535,12 +1535,28 @@ impl<S: StorageProvider> Engine<S> {
...
+            // while MDK accidentally inherited OpenMLS's 5-message default
+            // would therefore remain fragile after merely changing the
+            // creation/join presets. Normalize the complete engine-owned join
+            // config before any new traffic can be processed and write it
+            // through in place. This cannot restore secrets already pruned,
+            // but it protects all future within-epoch application traffic.
+            let required_config = crate::wire_format::join_config(self.max_past_epochs);
+            if mls_group.configuration().sender_ratchet_configuration()
+                != required_config.sender_ratchet_configuration()
+            {
```
erskingardner: Non-blocking: the guard triggers on sender-ratchet drift only, but the write below normalizes the entire join config. Consider comparing the whole config instead:

```rust
if mls_group.configuration() != &required_config {
```

`MlsGroupJoinConfig` derives `PartialEq`, so it's a one-line change, and it makes the migration self-triggering for any future engine-owned join-config drift rather than only this field — as written, a later change to another `join_config` field would silently skip normalization for existing groups. (Today the ratchet is the only field that can actually drift, since `EngineBuilder` hard-fails non-default `max_past_epochs`, so correct as written for this regression — fine to defer.)

erskingardner (commented):
Solid, well-scoped fix — the premise is sound and the implementation is complete. Verified against the pinned OpenMLS fork (erskingardner/openmls@59e7d3b):

- `SenderRatchetConfiguration::new(out_of_order_tolerance, maximum_forward_distance)` — argument order is correct, and the OpenMLS default really is `(5, 1000)`, matching the regression story in the PR body.
- `MlsGroup::set_configuration(&Storage, &MlsGroupJoinConfig)` exists at this rev and is a single idempotent storage write; mapping its failure to quarantine is fail-closed and retryable.
- Coverage is complete: the only group-origination sites are the `MlsGroupCreateConfig` in `group_lifecycle.rs` and `wire_format::join_config` (Welcome join), and the hydration normalization lives in `hydrate_one_stored_group`, which every `&mut` entry point reaches via `ensure_hydrated` before processing traffic — so no legacy group can process messages with the stale policy.
- Ledger/docs/script are consistent: 28 ids, P9/P10 rows in the inventory and plan, the discovery regex covers all three `wire_format.rs` constants, and no stale "eight P values" enumerations remain elsewhere in the repo.
- Test layering is right: unit pin on the join config, SQLCipher migration + restart-persistence test (which also pins OpenMLS's default so a dependency upgrade fails loudly), and the seed-2001 end-to-end regression.

One non-blocking suggestion inline on the migration guard.

Validation gap: all GitHub CI checks were still pending when I reviewed, so this relies on the local validation listed in the PR body — hold merge until CI is green.

PR #1343, opened 2026-08-09
test(app): cover reconnect hydration ordering

### LINKED ISSUE #1337: Test reconnect hydration before account worker resumes steady state
## Evidence

- `crates/marmot-app/src/runtime/account_worker.rs:912-932` gives reconnects a separate `drain_deferred_hydration` path before transport activation and steady-state command handling.
- `crates/marmot-app/tests/startup_hydration.rs:178-242` verifies deferred hydration only across cold boots. The reconnect test in `crates/marmot-app/src/tests.rs:520-675` exercises transport recovery and one pending send, but does not read multiple stored groups after reconnect.

## Regression risk

A refactor could skip or partially drain reconnect hydration while existing tests still pass because sends promote their target group on demand. Other stored groups would then enter steady state still gated by `GroupHydrationPending`, violating the reconnect invariant documented at the call site.

## Suggested test

Seed more than one hydration batch of stored groups, force the account worker through its receive-error/reconnect path, then assert every group is fully hydrated/readable immediately after reconnect. If practical, inject a hydration failure to verify the worker retries without entering steady state.

agent-p1p:
Fixes #1337

## Summary
- Adds a deterministic account-worker reconnect test with more than one deferred hydration batch.
- Forces relay notification recovery, proves the affected account re-subscribed, and verifies deferred groups are fully hydrated before steady-state commands and reads resume.
- Keeps all new relay/reconnect probes behind `cfg(test)`; production behavior is unchanged.
- Adds `Unreleased` changelog coverage. No version, fixture cohort, or conformance changes.

## Verification
- RED proof: temporarily bypassing `drain_deferred_hydration` made the regression test time out after reconnect.
- `cargo test -p marmot-app --lib --locked reconnect_drains_deferred_hydration_before_steady_state_serves_groups -- --nocapture`
- `just fast-ci`
- `cargo test -p marmot-app --lib --locked` — 596 passed
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked -- --test-threads=1` — 94 passed during pre-final-base validation

## Sensitive paths
- `crates/marmot-app/src/runtime/account_worker.rs` (test-only group hydration introspection)
- `crates/marmot-app/src/runtime/commands.rs` (test-only worker RPC)

No production crypto, MLS, key-handling, trust-anchor, or authentication behavior changed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved account recovery after notification or receive errors.
  * Deferred groups are now fully restored before normal command processing resumes.
  * Ensured all groups remain available and readable after reconnection.

* **Tests**
  * Added integration coverage for reconnect recovery and multi-batch group hydration.

### DIFF
--- a/crates/marmot-app/src/runtime/account_worker.rs
+++ b/crates/marmot-app/src/runtime/account_worker.rs
@@ -1354,6 +1354,9 @@ async fn handle_account_worker_catch_up(
 /// its own group's hydration.
 const STARTUP_HYDRATION_BATCH_SIZE: usize = 4;
 
+#[cfg(test)]
+pub(crate) const STARTUP_HYDRATION_BATCH_SIZE_FOR_TEST: usize = STARTUP_HYDRATION_BATCH_SIZE;
+
 /// Commands served between hydration batches. Bounded so sustained
 /// account-worker traffic cannot starve the pipeline: without a budget, an
 /// unbounded drain-until-empty could defer hydration (and the mutation
--- a/crates/marmot-app/src/runtime/account_worker.rs
+++ b/crates/marmot-app/src/runtime/account_worker.rs
@@ -360,6 +360,12 @@ pub(crate) enum AccountWorkerCommand {
         reason: String,
         respond: oneshot::Sender<Result<(), AppError>>,
     },
+    /// Count seeded groups the session has not fully hydrated yet, without
+    /// promoting them on demand (mdk#1337 regression probe).
+    #[cfg(test)]
+    UnhydratedGroupCount {
+        respond: oneshot::Sender<usize>,
+    },
 }
 
 impl AccountWorkerCommand {
@@ -1545,6 +1551,11 @@ async fn handle_startup_hydration_command(
         AccountWorkerCommand::QuarantinedGroups { respond } => {
             let _ = respond.send(Ok(client.quarantined_groups()));
         }
+        #[cfg(test)]
+        AccountWorkerCommand::UnhydratedGroupCount { respond } => {
+            let count = client.runtime.session().unhydrated_group_ids().len();
+            let _ = respond.send(count);
+        }
         AccountWorkerCommand::CatchUp { respond } => {
             deferred.push(DeferredStartupCommand::CatchUp(respond));
         }
@@ -1572,6 +1583,11 @@ async fn handle_account_worker_command(
     match command {
         AccountWorkerCommand::NetworkStartupSettled { respond } => {
             let _ = respond.send(());
+        }
+        #[cfg(test)]
+        AccountWorkerCommand::UnhydratedGroupCount { respond } => {
+            let count = client.runtime.session().unhydrated_group_ids().len();
+            let _ = respond.send(count);
         }
         AccountWorkerCommand::CatchUp { respond } => {
             let sync_started_at = Instant::now();
--- a/crates/marmot-app/src/tests.rs
+++ b/crates/marmot-app/src/tests.rs
@@ -39,6 +39,7 @@ use crate::messages::{AppMessageIntent, build_inner_event};
 struct ScriptedPushRelayClient {
     publish_results: std::sync::Mutex<std::collections::VecDeque<bool>>,
     published_events: std::sync::Mutex<Vec<NostrTransportEvent>>,
+    subscriptions: std::sync::Mutex<Vec<NostrSubscription>>,
     block_next_subscribe: std::sync::atomic::AtomicBool,
     fail_blocked_subscribe: std::sync::atomic::AtomicBool,
     fail_next_subscribe: std::sync::atomic::AtomicBool,
@@ -135,6 +136,21 @@ impl ScriptedPushRelayClient {
     fn release_publish(&self) {
         self.publish_release.notify_waiters();
     }
+
+    fn inbox_subscription_count(&self, expected_account_id: &MemberId) -> usize {
+        self.subscriptions
+            .lock()
+            .unwrap()
+            .iter()
+            .filter(|subscription| {
+                matches!(
+                    subscription,
+                    NostrSubscription::AccountInbox { account_id, .. }
+                        if account_id == expected_account_id
+                )
+            })
+            .count()
+    }
 }
 
 #[async_trait]
@@ -176,6 +192,7 @@ impl NostrRelayClient for ScriptedPushRelayClient {
                 ));
             }
         }
+        self.subscriptions.lock().unwrap().push(subscription);
         Ok(())
     }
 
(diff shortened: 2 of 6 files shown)


agent-p1p:
## Pip adversarial review — LGTM

**Verdict:** No blocking findings. One non-blocking test-maintenance suggestion.

### Suggestion

- `crates/marmot-app/src/tests.rs:4302-4306` — The fixture duplicates the current hydration batch size as the literal `4`. If `STARTUP_HYDRATION_BATCH_SIZE` changes, this test can keep passing while no longer spanning multiple batches, silently losing the issue's core regression guarantee. Tie `GROUP_COUNT` (or the assertion) to the production batch-size constant through a test-only visibility/re-export seam.

### Verified

- Reviewed exact head `ff64ef328bb574cd02108ab93cc15a1bee447a05`.
- Traced recovery from `recover_relay_notification_forwarder` clearing account delivery senders, through the worker receive-error/reopen path, to `drain_deferred_hydration` preceding transport preparation and steady-state command handling.
- The probe is non-promoting and worker-routed, so bypassing the reconnect drain cannot be masked by `group_members` hydrating its target on demand.
- The six-group fixture spans two batches under the current batch size and confirms the affected account re-subscribes before requiring zero unhydrated groups and reading every group.
- `cargo test -p marmot-app --lib --locked reconnect_drains_deferred_hydration_before_steady_state_serves_groups -- --nocapture` — passed 3/3 independent runs.
- `cargo fmt --all --check` — passed.
- `cargo clippy -p marmot-app --all-targets --locked -- -D warnings` — passed.
- `crates/cli/CHANGELOG.md` contains an `Unreleased` entry; no version, conformance, fixture-cohort, or lockfile churn.

### Sensitive-path note

`crates/marmot-app/src/runtime/account_worker.rs` and `commands.rs` gain test-only introspection. No production MLS, crypto, key-handling, authentication, or group-state behavior changed.

Counts: 0 blocking, 1 suggestion, 0 nitpicks.

agent-p1p:
Addressed the adversarial review suggestion in 69a2903e: the reconnect fixture now derives its group count from a `cfg(test)` re-export of the production hydration batch size and adds one, so it cannot silently stop spanning multiple batches. `just fast-ci`, the focused reconnect regression, and all 596 marmot-app library tests pass.

crates/marmot-app/src/tests.rs
```diff
@@ -4250,6 +4267,111 @@ fn account_worker_reconnect_backoff_doubles_caps_and_resets() {
...
+        loop {
+            if relay.inbox_subscription_count(&account_id) >= 1
+                && matches!(
+                    runtime.unhydrated_group_count_for_test(ACCOUNT).await,
+                    Ok(0)
+                )
+            {
+                break;
+            }
+            tokio::task::yield_now().await;
```
erskingardner, line 4321: Nit: prefer `tokio::time::sleep(std::time::Duration::from_millis(10)).await` over `yield_now()` here and at the two sites below (4336, 4341).

Every other bounded-wait loop in this file polls with `sleep(10ms)` (see the telemetry waits around lines 588 and 683), and the difference matters on the reconnect leg specifically: `ACCOUNT_WORKER_RECONNECT_BASE_DELAY` is 2s, so `yield_now()` spins this `current_thread` runtime hot for that whole backoff while issuing a fresh `UnhydratedGroupCount` RPC per iteration — commands the backoff loop explicitly drops (`Some(command) => drop(command)`), so each one is a full channel round-trip that resolves to `TransportClosed`. A 10ms sleep gets the same convergence with ~200 probes instead of however many the scheduler allows.

crates/marmot-app/src/tests.rs
```diff
@@ -4250,6 +4267,111 @@ fn account_worker_reconnect_backoff_doubles_caps_and_resets() {
...
+    .expect("notification recovery should reconnect and drain deferred hydration");
+
+    assert_eq!(
+        runtime
+            .unhydrated_group_count_for_test(ACCOUNT)
+            .await
+            .unwrap(),
+        0,
+        "reconnect must drain deferred hydration before steady-state group reads"
+    );
```
erskingardner, line 4356: This assertion is unreachable-if-false: the loop above only breaks on `Ok(0)`, so by the time control gets here the count is already known to be 0. It costs another worker round-trip to re-learn that.

Suggest deleting lines 4349-4356 and moving the intent into the `.expect(...)` message on the loop, which is the thing that actually fails when the invariant breaks.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1348,6 +1354,9 @@ async fn handle_account_worker_catch_up(
 /// its own group's hydration.
 const STARTUP_HYDRATION_BATCH_SIZE: usize = 4;
 
+#[cfg(test)]
+pub(crate) const STARTUP_HYDRATION_BATCH_SIZE_FOR_TEST: usize = STARTUP_HYDRATION_BATCH_SIZE;
```
erskingardner, line 1358: Nit: this gives one constant two names for the rest of the crate to keep in sync mentally. Since the only consumer is in-crate, `#[cfg(test)] pub(crate)` on `STARTUP_HYDRATION_BATCH_SIZE` itself — or re-exporting it under its own name in `runtime/mod.rs` — reads the same and drops the alias line plus the `_FOR_TEST` spelling from `runtime/mod.rs`.

(Deriving `GROUP_COUNT` from the production constant instead of the literal `4` is the right fix for the adversarial-review point; it's only the extra name I'd skip.)

erskingardner (approved):
Approving. The premise is sound and the test has real teeth: nothing else promotes those groups after reconnect (the probe deliberately does not, ingest delivers nothing under the scripted relay), and `prepare_transport()` runs strictly after `drain_deferred_hydration`, so the "new inbox subscribe appeared, count is still > 0" state is exactly the regression — matching the RED proof in the description. Production behavior is unchanged and the test-only surface is small and honestly gated.

Three non-blocking nits inline, all in the "fewer lines / match existing style" bucket:

1. The poll loops busy-spin with `yield_now()` where the rest of this file uses `sleep(10ms)`. Real cost here, not stylistic: `ACCOUNT_WORKER_RECONNECT_BASE_DELAY` is 2s, so this hot-spins a single-threaded runtime for ~2s+jitter while firing a fresh worker RPC every iteration into a bounded channel that the backoff loop just drops on the floor.
2. The trailing `assert_eq!(..., 0)` re-asserts the loop's own break condition — ~8 deletable lines and one more round-trip.
3. `STARTUP_HYDRATION_BATCH_SIZE_FOR_TEST` gives one constant two names. Deriving the group count from the production constant (rather than a literal `4`) is the right call; the extra alias is what I'd drop.

One coverage note, not a request: #1337 also suggested injecting a hydration failure to prove the worker retries without entering steady state. The `continue` arm at `account_worker.rs:1011-1024` is still uncovered. Fine to leave for a follow-up — say so on the issue if you'd rather close it out as-is.

erskingardner (approved):
LGTM — reviewed at head 3800bac3.

**Verification (source-traced at the exact head SHA; CI green on all gates):**

- The regression net is the non-promoting probe, and it cannot be masked: `unhydrated_group_ids()` is a read-only getter over `route_backfill_pending`/`unhydrated_groups`, while the steady-state `Members` handler promotes on demand via `ensure_group_hydrated` — so in the RED scenario (drain bypassed) `group_members` would still pass and only the probe's permanent non-zero count fails the test. Consistent with the PR's RED-proof claim.
- Probe coverage is complete across every worker window: startup pipeline arm answers live between batches; initial-sync window defers it (so a `0` reading is only observable *after* transport activation — which also closes any vacuous-pass race on the phase-1/phase-2 subscription-count checks); steady state answers live; reconnect backoff drops it promptly as `TransportClosed`, which the test loop correctly retries; during the reconnect drain it queues and is answered `0` once steady state resumes.
- `simulate_notification_recovery_for_test` drives the real production recovery path (`recover_relay_notification_forwarder` clears `account_deliveries` → worker receive error → backoff → deferred reopen → `drain_deferred_hydration` → `prepare_transport` re-subscribe), and the phase-2 conditions (subscription increase + probe 0) both strictly imply the drain completed.
- Issue #1337 acceptance met: 5 groups = `STARTUP_HYDRATION_BATCH_SIZE + 1` spans two batches, and the fixture is now coupled to the production constant so it cannot silently stop spanning batches (the prior review suggestion, addressed in 3800bac3).
- All new surface is `cfg(test)`-gated; production code, crypto, and trust paths untouched. No version/fixture/conformance churn. Test-only changelog entries have precedent (a50e0e12).
- Timing is comfortable: 2s reconnect base delay vs 30s loop budgets.

No blocking or non-blocking findings.

PR #1344, opened 2026-08-09
Harden simulator oracle and replay fidelity

erskingardner:
## What changed

- Treat either a fixed-point quiescence observation or an exact no-pending-work observation as valid coverage after an explicit virtual-time advance.
- Keep semantic minimizers as diagnostic report metadata instead of substituting them into generated fixture candidates.
- Document the distinction in the simulator guide, scenario registry, and convergence tracking plan.

## Why

Fresh convergence-chaos campaigns exposed two simulator trust issues:

1. Seed 2001 case 4 satisfied its state and no-pending expectations but failed strict mode because virtual-time coverage accepted only the quiescence driver.
2. Older minimized fixture candidates for cases 2 and 13 retained the original expectations while replacing the 29-step executed scenarios with 19-step semantic reductions. Those candidates reproduced zero deliveries rather than the original five-of-six delivery failure.

The reducer intentionally preserves only failure class, action type, and kind, so it remains useful for diagnosis. A fixture candidate now keeps the original executed scenario and therefore cannot imply stronger replay fidelity than the reducer provides.

## Impact

This changes simulator/reporting behavior only. It does not modify the convergence engine, app runtime, storage, wire format, or production policy constants.

## Validation

- cargo test -p cgka-conformance-simulator oracle::tests --locked
- cargo test -p cgka-conformance-simulator report::tests --locked
- cargo test -p cgka-conformance-simulator --test report_runner --locked
- cargo test -p cgka-conformance-simulator --test canonical_scenarios failing_generated_case_records_a_minimized_reproducer --locked
- strict file-backed replay of convergence-chaos seed 2001 case 4: 1 passed, 0 failed
- just fast-ci

## Summary by CodeRabbit

* **Bug Fixes**
  * Generated failure fixtures now preserve the original executed scenario, preventing inaccurate or incomplete reproductions.
  * Minimized reproducers remain available as diagnostic metadata rather than replacing fixture scenarios.
  * Virtual-time checks now recognize either confirmed quiescence or explicit confirmation that no work remains.
  * Improved warnings identify both types of missing temporal-closure evidence.

* **Documentation**
  * Clarified fixture fidelity, diagnostic reproducers, and accepted virtual-time completion evidence.

### DIFF
--- a/crates/cgka-conformance-simulator/src/oracle.rs
+++ b/crates/cgka-conformance-simulator/src/oracle.rs
@@ -799,7 +799,14 @@ fn recommended_behaviors(stimulus: ScenarioStimulus) -> Vec<OracleBehavior> {
             OracleBehavior::ReplayDeduplication,
         ],
         ScenarioStimulus::PublishLifecycle => vec![OracleBehavior::PublishLifecycleChecked],
-        ScenarioStimulus::VirtualTimeAdvance => vec![OracleBehavior::QuiescenceState],
+        // Advancing the controlled clock must be paired with evidence that the
+        // resulting timed work was observed. A fixed-point quiescence result is
+        // the strongest form, while a final exact no-pending observation is the
+        // appropriate equivalent for explicitly stepped scenarios.
+        ScenarioStimulus::VirtualTimeAdvance => vec![
+            OracleBehavior::QuiescenceState,
+            OracleBehavior::NoPendingWorkObserved,
+        ],
     }
 }
 
@@ -942,5 +949,35 @@ mod tests {
 
         assert!(behaviors.contains(&OracleBehavior::BidirectionalDecryptabilityObserved));
         assert!(behaviors.contains(&OracleBehavior::DeliveredPayload));
+    }
+
+    #[test]
+    fn virtual_time_accepts_fixed_point_or_exact_no_pending_evidence() {
+        assert!(
+            weak_oracle_warnings(
+                &[ScenarioStimulus::VirtualTimeAdvance],
+                &[OracleBehavior::QuiescenceState],
+            )
+            .is_empty()
+        );
+        assert!(
+            weak_oracle_warnings(
+                &[ScenarioStimulus::VirtualTimeAdvance],
+                &[OracleBehavior::NoPendingWorkObserved],
+            )
+            .is_empty()
+        );
+        assert_eq!(
+            weak_oracle_warnings(&[ScenarioStimulus::VirtualTimeAdvance], &[])
+                .first()
+                .map(|warning| warning.expected_any_of.as_slice()),
+            Some(
+                [
+                    OracleBehavior::QuiescenceState,
+                    OracleBehavior::NoPendingWorkObserved,
+                ]
+                .as_slice()
+            )
+        );
     }
 }
--- a/crates/cgka-conformance-simulator/src/oracle.rs
+++ b/crates/cgka-conformance-simulator/src/oracle.rs
@@ -801,8 +801,10 @@ fn recommended_behaviors(stimulus: ScenarioStimulus) -> Vec<OracleBehavior> {
         ScenarioStimulus::PublishLifecycle => vec![OracleBehavior::PublishLifecycleChecked],
         // Advancing the controlled clock must be paired with evidence that the
         // resulting timed work was observed. A fixed-point quiescence result is
-        // the strongest form, while a final exact no-pending observation is the
-        // appropriate equivalent for explicitly stepped scenarios.
+        // the strongest form; an exact no-pending observation is the appropriate
+        // equivalent for explicitly stepped scenarios. Coverage is set-based and
+        // does not prove that the no-pending expectation follows the advance;
+        // generated families rely on `add_strict_reliability_oracle` appending it.
         ScenarioStimulus::VirtualTimeAdvance => vec![
             OracleBehavior::QuiescenceState,
             OracleBehavior::NoPendingWorkObserved,
--- a/docs/marmot-architecture/convergence-reliability-plan.md
+++ b/docs/marmot-architecture/convergence-reliability-plan.md
@@ -1094,7 +1094,9 @@ copying raw error text.
 The reducer now treats paired transport, partition, lifecycle, and storage actions as dependency units before removing
 independent noise. Non-layer-specific failures receive a canonical Scenario IR candidate for the next smaller adapter;
 the semantic identity must reproduce there before concluding that the larger layer is uninvolved. Promotion remains
-restricted to synthetic shareable capsules with portable expectations.
+restricted to synthetic shareable capsules with portable expectations. Semantic minimization may change the terminal
+state digest, so generated fixture candidates retain the original executed scenario; the smaller case is labeled only
+as diagnostic report metadata unless a separate replay contract proves stronger fidelity.
 
 ### 6.5 Stateful canonical journeys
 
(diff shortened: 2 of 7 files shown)


erskingardner (commented):
Recommend merge (posted as a comment: GitHub does not allow approving your own PR).

Both changes are correct, minimal, and net-simplifying; the fixture fix is a genuine deletion of a trust-breaking special case rather than a workaround.

**Fixture fidelity (`report.rs`)** — right call, and the right size. The old code substituted `minimized_case` into the candidate while keeping the original `expected_outcomes`, producing a fixture whose scenario and expectations came from different runs. Since the reducer only preserves failure class/action/kind and explicitly permits the terminal state digest to move, that substitution could not be sound. Removing it (rather than trying to make the reducer digest-preserving) is the smaller and more honest fix, and the diagnostic path is unaffected: `family.rs` still records `minimized_case`, `failure_capsule.rs` still carries it as `minimized_reproducer`, and `failing_generated_case_records_a_minimized_reproducer` still covers the reducer itself. I checked for stale claims elsewhere and found none — `AGENTS.md` and `cgka-engine-quality-and-vectors.md` describe the reducer without asserting fixture substitution, so the three docs updated here are the complete set.

**Oracle widening (`oracle.rs`)** — this is a widening of a strict gate, so it deserves the scrutiny, and it holds up. `NoPendingWorkObserved` is an exact per-client pending-set assertion, not a weaker proxy for the quiescence fixed point, and it enters `oracle_behaviors` only through `TraceExpectation::NoPendingWork` or an exact expected trace — never from the observed trace alone, so a case cannot satisfy its own coverage requirement. The blast radius is also narrower than it first looks: `AwaitQuiescence` inserts `QuiescenceGate` alongside `VirtualTimeAdvance`, and `QuiescenceGate` still demands `QuiescenceState`, so only scenarios using a bare `AdvanceTime` are affected. Using the existing any-of table entry instead of a special case keeps it consistent with every other stimulus.

One non-blocking note on the comment wording, inline. Nothing else worth the author's time — the redundant `assert_ne!` in the new `report.rs` test is implied by the `assert_eq!` above it, but it reads as deliberate intent-documentation and is fine to leave.

Docs, plan-log entry, and tests all match what shipped. Full CI is green at `e1965f5`, including the `Simulator and vectors` job.

crates/cgka-conformance-simulator/src/oracle.rs
```diff
@@ -799,7 +799,14 @@ fn recommended_behaviors(stimulus: ScenarioStimulus) -> Vec<OracleBehavior> {
             OracleBehavior::ReplayDeduplication,
         ],
         ScenarioStimulus::PublishLifecycle => vec![OracleBehavior::PublishLifecycleChecked],
-        ScenarioStimulus::VirtualTimeAdvance => vec![OracleBehavior::QuiescenceState],
+        // Advancing the controlled clock must be paired with evidence that the
+        // resulting timed work was observed. A fixed-point quiescence result is
+        // the strongest form, while a final exact no-pending observation is the
+        // appropriate equivalent for explicitly stepped scenarios.
```
erskingardner: Non-blocking: the comment promises ordering that the mechanism cannot enforce.

`weak_oracle_warnings` reduces `oracle_behaviors` to a `BTreeSet` and asks only whether any recommended behavior is present. There is no step index, so "a *final* exact no-pending observation" is not what gets checked — a `TraceExpectation::NoPendingWork` appearing anywhere satisfies `VirtualTimeAdvance`, including one asserted before the `AdvanceTime` step that made the stimulus fire.

In the generated families this is fine today: `add_strict_reliability_oracle` appends `DeliverAll` / `Tick` / `ObserveExact` and pushes `NoPendingWork` as the last expectation, so it really is terminal for the cases this PR was chasing (including seed 2001 case 4). The gap is a hand-authored scenario that advances the clock after an early no-pending assertion and gets temporal-closure credit for free.

The pre-existing `QuiescenceState` path has the same looseness (`AwaitQuiescence` anywhere in `steps` counts), so I would not add ordering machinery for this alone — but since this PR exists to stop coverage signals from claiming more than they prove, it is worth not restating the same overclaim in the comment. Suggest dropping "final" and naming the assumption instead, e.g.:

```rust
// Advancing the controlled clock must be paired with evidence that the
// resulting timed work was observed. A fixed-point quiescence result is
// the strongest form; an exact no-pending observation is the appropriate
// equivalent for explicitly stepped scenarios. Coverage is set-based and
// does not verify the no-pending expectation follows the advance --
// generated families rely on add_strict_reliability_oracle appending it
// last.
```

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -1094,7 +1094,9 @@ copying raw error text.
 The reducer now treats paired transport, partition, lifecycle, and storage actions as dependency units before removing
 independent noise. Non-layer-specific failures receive a canonical Scenario IR candidate for the next smaller adapter;
 the semantic identity must reproduce there before concluding that the larger layer is uninvolved. Promotion remains
-restricted to synthetic shareable capsules with portable expectations.
+restricted to synthetic shareable capsules with portable expectations. Semantic minimization may change the terminal
+state digest, so generated fixture candidates retain the original executed scenario; the smaller case is labeled only
+as diagnostic report metadata unless a separate replay contract proves stronger fidelity.
```
erskingardner: This new invariant conflicts with the promotion contract stated earlier in the same file (line 541): `promote_failure_capsule_to_vector` "uses the fingerprint-preserving minimized scenario when present."

Two issues:

1. **"Fingerprint-preserving" is no longer accurate under this PR's own framing.** The reducer preserves only the semantic identity (classification, failing action type, failure kind) and deliberately permits the terminal state digest to change — and the full fingerprint includes `normalized_state_digest`. Suggest rewording to "semantic-identity-preserving" at minimum.

2. **The corpus paths are a second set of substitution sites this PR doesn't touch.** `promote_failure_capsule_to_vector` (`crates/cgka-conformance-simulator/src/failure_capsule.rs:413-416`) and `observation_from_capsule` (`crates/convergence-campaign-runner/src/failure_corpus.rs:566-569`) both still prefer the minimized reproducer, pairing it with the original report's expectations. That may well be the intended design — the reducer re-runs the reduced case and verifies identity reproduction against the same expectations, which could qualify as the "separate replay contract [that] proves stronger fidelity" this sentence allows for. But as written, a reader cannot tell whether corpus promotion is exempt by design or was missed. Please state the exception explicitly here (and in the line-541 contract), or align the code with the invariant.

erskingardner (commented):
The core changes are correct, well-scoped, and well-tested — the oracle relaxation has the right semantics and the fixture-fidelity fix removes a real trap. One consistency gap to close before this lands: the new "diagnostic metadata only" invariant is contradicted earlier in the same document, and the corpus-promotion paths still substitute the minimized reproducer without saying whether that is the sanctioned exception. Details inline.

What I verified:

- `weak_oracle_warnings` is any-of semantics over expectation-derived `oracle_behaviors`, so adding `NoPendingWorkObserved` loosens strict mode exactly as intended — and only for explicit `AdvanceTime` scenarios. `AwaitQuiescence` still requires `QuiescenceState` via the separate `QuiescenceGate` stimulus, which matches the new comment's "strongest form" framing.
- The newly accepted evidence is genuinely strong: `TraceExpectation::NoPendingWork` requires an exact empty pending-work progress snapshot and rejects legacy observations that lack one.
- `to_vector_fixture` derives `scenario_name` from `case.scenario.name`, and the reducer preserves the scenario name, so dropping the `scenario_name = minimized.name` override is a no-op; the behavioral change is exactly the substitution removal. The new unit test pins it, and the existing `failing_generated_case_records_a_minimized_reproducer` asserts only on metadata, so no other test updates were needed.
- Scope and repo invariants hold: simulator + docs only, no version bumps, no tracing changes. CI is green, including Simulator and vectors.

PR #1345, opened 2026-08-09
Add four-party cross-route convergence regression

erskingardner:
## Summary

Adds a permanent four-participant Scenario IR vector for the same-source-epoch cross-route topology repaired by #1285.

The scenario drives two simultaneous privileged commits, pairwise displacement of a locally authored branch, observer-side stored convergence, ordering-key versus branch-depth disagreement, branch growth, and an encrypted-SQLite restart before the displaced own branch is selected again.

It then requires exact canonical cryptographic equality, the expected group profile, no pending work, durable accepted and invalidated commit dispositions, and successful application-message probes in all twelve participant directions.

The route-assurance inventory, scenario registry, vector manifest, and Milestone 6 tracking plan now distinguish this completed engine checkpoint from the retained-history, app-runtime, process, container, and VM permutations that remain open.

## Impact

This changes simulator tests, portable vectors, assurance metadata, and documentation only. It does not change production convergence, storage, application, wire, or protocol behavior.

## Validation

- just fast-ci
- cargo test -p cgka-conformance-simulator --test canonical_scenarios cross_route_own_commit_recovery_survives_restart_with_exact_agreement --locked
- cargo test -p cgka-conformance-simulator --test canonical_scenarios canonical_vector_fixtures_match_generated_traces --locked
- cargo test -p cgka-conformance-simulator --test route_assurance --locked
- cargo test -p cgka-conformance-simulator --test vector_artifacts --locked
- strict file-backed report execution for cross-route-own-commit-recovery/v1


erskingardner (commented):
Reviewed at head 596b871. Verdict: the change itself is approve-worthy — posting as a comment because GitHub doesn't allow self-approval.

**One required action before merge: rebase.** The PR reports `mergeable_state: dirty`. Master has moved in `crates/cgka-conformance-simulator/SCENARIOS.md` and `docs/marmot-architecture/convergence-reliability-plan.md` since the base (c1c50c9), and this PR edits the same regions of both files.

What I verified:

- **Vector internal consistency** — the `expected_outcomes` step indices (1/7/8/13) and the Rust test's durable-disposition ids (`step-5`/`step-6`/`step-12`) match the 0-indexed step list; withhold/release labels pair up correctly; the `restart_client` lands exactly after zeta's pairwise displacement and before the depth-two child is released, which is precisely the commit-addressed checkpoint path from #1285.
- **The "encrypted-SQLite restart" claim is real** — `HarnessStorageMode::TempFileBackedSqlite` opens `SqliteAccountStorage::open_encrypted_with_options` with a `SqlCipherKey`, and `HarnessClient::restart` drops both the engine and the storage handle, reopens the file, and rehydrates. Not an in-memory pass.
- **No duplicate coverage** — `canonical_vector_fixtures_match_generated_traces` already executes every `.v1.json` under the default in-memory mode; the dedicated test adds the file-backed execution plus the recovery-observation count/epochs, zeta 2→3 transition, accepted/invalidated dispositions, and probe-terminal-disposition assertions that the portable expectation format cannot express.
- **Machine-checked invariants hold** — every updated `partial` campaign owner keeps a non-empty limitation, the matrix keeps exactly one row per registered route, the manifest entry uses the established `scenario_vector`/`portable` vocabulary, and the vector's `conformance_version` (0.9.10) matches the workspace version. Bidirectional manifest↔file consistency is enforced by the existing `vector_artifacts` test.
- **Assertions pin the topology** — `recovery_observations.len() == 1` with source epoch 1 → recovered epoch 2 correctly reflects that only zeta can experience own-commit displacement (yankee/observer authored neither root).

One optional follow-up, deliberately not for this PR: the campaign-owner strings in `route_assurance.rs` name vector ids (`cross-route-own-commit-recovery/v1`, etc.) that nothing cross-checks against the manifest — a typo there would compile and pass. A small assertion in the `route_assurance` test that every `*/v1` token resolves to a manifest entry would make the inventory fully machine-checked.

All 17 CI checks pass at this head.

crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
```diff
@@ -3122,6 +3123,101 @@ async fn harness_captures_and_asserts_convergence_decision() {
...
+    assert!(
+        report.invariant_failures.is_empty(),
+        "cross-route invariants must pass: {:#?}",
+        report.invariant_failures
+    );
+    assert_eq!(
+        report.recovery_observations.len(),
+        1,
+        "the ordering-key winner must first displace zeta's own branch pairwise"
+    );
```
erskingardner: This cannot be observing zeta.

`HarnessClient::restart` (`src/client.rs`) does `drop(self.engine.take())` and `self.pending_events.clear()`, and `recoveries` are built from `drain_events()` at the final `observe_exact`. Zeta's `ForkRecovered` is emitted during the `tick` at step 17, captured into `pending_events` by `tick_ingest_only`, and then discarded by the `restart_client` at step 18. Post-restart zeta has no incumbent recorded at source epoch 2, so the direct seam emits no further `ForkRecovered` for it — the surviving observation belongs to another participant that rolled its incumbent back to alpha's root.

`ForkRecoveryObservation` has no client field, so aggregate `len() == 1` plus `{source_epoch: 1, recovered_epoch: 2}` cannot attribute the recovery at all. Either express this as `client_recoveries` in the vector (which does attribute) for whichever client genuinely emits it, or capture zeta's displacement with an observation step before the restart if that is the property you want pinned. As written the message is misleading, and it will keep being misleading to whoever debugs this next.

Note that zeta's own-branch recovery is still covered indirectly and correctly — by the post-restart `zeta 2 -> 3` epoch change plus exact canonical equality on `zeta-branch-depth-two`. It is only this assertion that is mis-labelled.

crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
```diff
@@ -3122,6 +3123,101 @@ async fn harness_captures_and_asserts_convergence_decision() {
...
+        report
+            .epoch_change_observations
+            .iter()
+            .any(|change| { change.client == "zeta" && change.from == 2 && change.to == 3 })
+    );
+
+    let trace = report.observed_trace.expect("cross-route trace");
+    assert_eq!(trace.decryptability_probes.len(), 1);
+    assert!(trace.decryptability_probes[0].succeeded());
+    for observation in &trace.observations {
```
erskingardner, line 3157: This whole block duplicates machinery the vector format already has, in a non-portable place:

- the per-client `epoch == 3` / `group_name` loop -> `client_state` (as `group-data-fork-recovery.v1.json` does), or is already implied transitively by the `clients_exactly_equivalent` + `group_profile` outcomes the vector declares;
- the three `scenario_id` disposition lookups -> `scenario_input_ledger` if you want the complete durable ledger pinned portably (heavier, since it requires listing every entry including the probe traffic — worth a judgement call);
- `decryptability_probes[0].succeeded()` -> already covered by the `clients_bidirectionally_decryptable` outcome in the vector.

Since `canonical_vector_fixtures_match_generated_traces` already runs this fixture (in-memory) and compares every declared outcome, folding these into `expected_outcomes` gets them checked twice and makes them real for external implementations. What is left here is then just: load fixture, run with `TempFileBackedSqlite`, assert `expectation_failures` and `invariant_failures` are empty — roughly fifteen lines instead of ninety-five, with strictly more coverage.

One assertion the existing sibling vector makes and this PR drops entirely: `winner_differs_from_invalidated`.

erskingardner (commented):
The topology is the right one to make permanent, and the #1285 story it pins (pairwise displacement -> deeper-branch reconsideration -> commit-addressed recovery after an encrypted-SQLite restart) is worth a vector. Two things should change before this lands.

**1. The assertions are in the wrong place.** The vector is registered as `"status": "portable"` and SCENARIOS.md advertises it under "scenarios another implementation should be able to load from JSON", but `expected_outcomes` only carries pending resolutions, exact equivalence, one group profile, decryptability, and no-pending-work. Everything that actually characterizes this topology — the recovery, the epoch transition, the accepted/invalidated commit dispositions, the per-client selected branch — lives only in the new Rust test, so a third-party implementation loading the JSON gets none of it. `TraceExpectation` already has `recovery_summary`, `client_recoveries`, `client_epoch_changes`, `client_state`, and `group_profile` for exactly this, and the sibling `group-data-fork-recovery.v1.json` uses `recovery_summary` + `client_state` this way. Moving these into the vector makes them portable, matches the crate's existing convention, and shrinks the 95-line test to the one thing the corpus runner does not already do: re-running the fixture under `TempFileBackedSqlite`.

**2. The recovery assertion does not measure what its message claims.** `HarnessClient::restart` drops the engine and clears `pending_events`, and `recoveries` are derived from `drain_events()` at the final `observe_exact`. Zeta's `ForkRecovered` from the pre-restart tick is therefore gone by the time the report is built, so `recovery_observations` cannot be reporting zeta's displacement — and `ForkRecoveryObservation` carries no client field, so the assertion cannot distinguish which participant recovered either. The check passes, but it is pinning a different participant than the failure message names. See the inline note for options.

Smaller point on the assurance metadata, inline. Nothing here touches production behavior, and CI is green, so this is about the regression actually pinning what it claims.

crates/cgka-conformance-simulator/vectors/cross-route-own-commit-recovery.v1.json
```diff
@@ -0,0 +1,238 @@
...
+        "clients": [
+          "zeta",
+          "alpha",
+          "yankee",
+          "observer"
+        ]
+      }
+    ]
+  },
+  "expected_outcomes": [
```
erskingardner, line 175: This is where the recovery, epoch-change, and per-client branch expectations belong, so the `"status": "portable"` claim in `manifest.v1.json` and the SCENARIOS.md framing hold for an implementation that only loads the JSON. `recovery_summary` / `client_recoveries` / `client_epoch_changes` / `client_state` are all already supported by `TraceExpectation`, and `group-data-fork-recovery.v1.json` is the precedent.

The step ordering and 0-indexed `step_index` values check out (5 = zeta-root, 6 = alpha-root, 12 = zeta-child), and the pairwise winner is deterministic because `CommitOrderingKey` breaks the tie on committer bytes from `deterministic_nostr_keys` before it ever reaches the randomized commit digest — worth a one-line comment somewhere, since the `zeta`/`alpha` naming reads as if lexical order were doing the work.

crates/cgka-conformance-simulator/src/route_assurance.rs
```diff
@@ -273,8 +273,8 @@ pub const DECISION_ROUTE_INVENTORY: &[DecisionRouteInventoryEntryV1] = &[
             "The sibling-model mutant proves the witness distinguishes two abstract rules, not that production takes the adopted transition.",
         ),
         campaign: AssuranceOwnerV1::partial(
-            "pairwise_incumbent_defers_to_deeper_convergence_branch and pairwise_candidate_win_leaves_old_incumbent_reconsiderable",
-            "Engine regressions cover reconsideration, but the cross-adapter #1236 topology remains open.",
+            "cross-route-own-commit-recovery/v1 plus the focused reconsiderable-loser engine regressions",
```
erskingardner: This replaces the two greppable test names (`pairwise_incumbent_defers_to_deeper_convergence_branch`, `pairwise_candidate_win_leaves_old_incumbent_reconsiderable`) with the prose "the focused reconsiderable-loser engine regressions". The repo invariant asks to keep Rust test names, vector names, and model names easy to grep across layers, and this entry is precisely the index someone greps from. Adding the new vector to the list is right; dropping the existing names is a loss. Same shape in the `application_disposition` entry below, where "scenario-input ledgers and bidirectional decryptability probes" is narrowed to this one vector's — the pre-existing coverage did not go away.

PR #1349, opened 2026-08-10
Close the strict-oracle coverage gaps in generated scenario families

dannym-arx:
A full strict report-CLI sweep over every generated family found two families that failed while every declared expectation passed. The engine behaved correctly in all of them; the marmot referee was miscounting its own scorecard.

`convergence-e2e-delivery/v1` failed 8/8 cases at every seed because its generator emitted `expected_outcomes: vec![]`. It was the only family never upgraded when strict oracle coverage became the default. `adversarial-reliability/v1` failed 9 of its 12 arms: most arms carried no expectation covering their `CreateGroup`/`PublishConfirm`/`AppMessage` stimuli, and `losing-invite-unrecoverable` hit a real classifier bug. The observed-behavior heuristic marks `ExactStateEquivalence` only when every canonical snapshot in the trace matches globally, so a scenario that asserts both `clients_exactly_equivalent(alice, bob)` and `clients_not_equivalent(david, eve)` could never observe the equivalence it proved.

Changes:

- `oracle.rs`: a passing `ClientsExactlyEquivalent` or `ClientsConverged` expectation now counts as observed evidence for its subset, mirroring the existing `ClientsNotEquivalent` rule. Both expectations fail on missing observations, so they cannot pass vacuously. A unit test pins the equivalent-pair-beside-divergent-device shape.
- `convergence-e2e-delivery/v1` (generator v2): the case builder appends a settle tail, an active bidirectional decryptability probe, an exact founder observation, and schedule-invariant expectations. The selected branch stays unpinned on purpose: app-witness scoring can hand the win to either committer depending on the delivery schedule, and losing-branch commits stay reconsiderable deferred candidates on non-committing clients, so final epoch/member-count and whole-roster no-pending-work are not asserted. SCENARIOS.md documents the reasoning.
- `adversarial-reliability/v1` (generator v3): a new `push_labelled_confirmation_expectations` helper pins every scripted accepted acknowledgement as a confirmed pending-resolution without hand-maintained step indices, and the six delivery-shaped arms get deterministic transcript expectations (offline catch-up, cross-sender rounds, noisy-neighbor isolation, multi-device, witness-decided selection, relay dedup after a year-long clock jump).
- The strict-oracle report test now pins the weak-oracle failure path through a deliberately expectation-stripped generated input, since no built-in family is weak anymore.

Verified: all 28 vectors, six families at seeds 1/42/1337 (full 12-arm adversarial catalog included) strict-green, the whole crate suite, the `test-policy-overrides` campaigns gate, and `just fast-ci`.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced convergence validation for delivered messages, acknowledgements, decryptability, and client equivalence.
  * Expanded reliability scenarios to cover catch-up, sustained traffic, multi-device delivery, branch selection, and replay behavior.
  * Reports now accurately recognize client convergence and exact equivalence, including cases with divergent clients.
* **Tests**
  * Updated scenario versions and strengthened generated-case validation.
  * Added coverage for weak-oracle failures and more flexible epoch-change reporting.

### DIFF
--- a/crates/cgka-conformance-simulator/SCENARIOS.md
+++ b/crates/cgka-conformance-simulator/SCENARIOS.md
@@ -499,10 +499,17 @@ regression, covers a new semantic edge, or is the smallest readable example of a
 
 ### `convergence-e2e-delivery/v1`
 
-- Generator: `generate_convergence_e2e_delivery_family`
-- Setup: the real-peeler invite-fork scenario runs with generated delivery mutations.
+- Generator: `generate_convergence_e2e_delivery_family` (generator version `2`).
+- Setup: the real-peeler invite-fork scenario runs with generated delivery mutations, then a settle tail delivers and
+  ticks the founders, accepts outstanding outbound work, runs an active bidirectional decryptability probe, and
+  observes the founders exactly.
 - Pressure: duplicate, delay, release, and reorder before observer clients tick.
-- Expected: observers converge on one selected branch and emit only the selected branch payload.
+- Expected: the mid-schedule observation stays branch-dependent, and so does the selected branch itself — app-witness
+  scoring can hand the win to either committer depending on the delivery schedule, and the losing branch's commits
+  deliberately remain reconsiderable deferred candidates on non-committing clients. The generated expectations
+  therefore pin the schedule-invariant claims: every scripted accepted acknowledgement confirms, the founders converge
+  and are exactly equivalent, and post-selection application traffic decrypts in every founder direction. Final
+  epoch/member-count and a whole-roster no-pending-work claim are deliberately not asserted.
 
 ### `admin-churn/v1`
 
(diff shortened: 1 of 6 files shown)


Datawav (commented):
## Automated review

No actionable findings.

erskingardner (approved):
Recommend merge after refreshing this branch onto current master and rerunning CI. I found no actionable code issues at exact head fc1630ffceb155acb9cc2d6c3f1e276725852516. The subset-equivalence classifier change is non-vacuous because it reuses the executable expectation comparison, and the new family expectations strengthen the campaign without pinning schedule-dependent branch details. This is directly useful for the planned strict all-family soak: without it, known oracle-coverage gaps would be cataloged as engine failures. One integration caveat: the current synthetic merge and completed CI are still against base 3c665c1b, while master is now e8f04109 with #1350 retained-history cross-route changes. Please refresh/rebase and rerun before merging.

PR #1295, opened 2026-08-07
Add the admin-churn/v1 generated scenario family

dannym-arx:
## Summary

Final slice of the multi-VM catalog port. The marmot colony now drills under fire: admin churn, authority races, crashes, and joins under pressure, all seeded and deterministic.

Adds `generate_admin_churn_family` (`admin-churn/v1`) with four arms rotating by case index:

- Arm 0: seeded sequential admin churn. A seeded admin renames or toggles admin membership each round, with a full message round between commits.
- Arm 1: competing same-epoch admin-policy commits with seeded delivery order, the admin-set analog of the chaos group-data fork. The race winner stays schedule-dependent and unpinned, but authority is enforced afterward: a retained admin pins the set with a commit that must succeed, both non-admins are rejected with `not_group_admin`, and every client asserts the pinned set.
- Arm 2: a seeded admin restarts between publishing a rename commit and its delivery, then must accept and originate traffic in the recovered group.
- Arm 3: a latecomer's welcome travels in the same delivery flight as the founder's next rename commit, so the join genuinely races commit pressure; zero-count assertions prove the latecomer never observes pre-join plaintext, and all four members carry exact transcripts.

Registered in the report CLI (`--family admin-churn/v1`), `lib.rs`, the vector manifest, `SCENARIOS.md`, and `AGENTS.md`, with a determinism-plus-execution regression test in `canonical_scenarios.rs`.

## Known gap surfaced

A welcome-joined member retains its own join commit as a permanently deferred transport input, so no latecomer scenario can pass a whole-roster pending-work oracle. Arm 3 uses a documented scoped oracle (`add_scoped_reliability_oracle`) covering the founders, and pins the joiner with a new `no_pending_work_except_retained_join_commit` expectation that allows exactly the retained join commit and fails on any other pending work. The same wart is why the slice-one latecomer and growth vectors scope their `no_pending_work` expectations. Worth an issue; it also affects the pre-existing `invite-member/v1` shape.

## Stack

Fifth slice in the #1250 series lineage: #1270, #1271 (merged), #1272, #1292, #1294 (base of this PR). Re-targets as parents merge.

## Scope boundary

Simulator-only: `family.rs`, `report.rs`, `lib.rs`, tests, and docs. No production code, no changes to existing families beyond registration lists.

## Verification

[...]

PR #1350, opened 2026-08-10
Test retained-history cross-route recovery

erskingardner:
## What changed

Adds a fixed, subject-pinned `cross-route-retained-history-recovery/v1` scenario for the four-party #1285 topology.

The new case drives competing source-epoch commits through retained relay history rather than the engine harness's semantic packet-bus withholding. Participant-specific visibility gives Yankee only Zeta's root, gives Zeta Alpha's competing root while hiding Yankee's deeper child, restarts Zeta after pairwise displacement, and then uses full-history queries to restore the complete input set.

The regression requires:

- exact epoch-3 canonical cryptographic agreement across all four participants;
- accepted/invalidated/accepted dispositions for the selected root, losing root, and selected child;
- no pending work;
- successful application decryptability in all twelve directions;
- evidence that both partial views and every final full-history query injected retained relay objects.

## Simulator fix

The scenario exposed a simulator-only observation bug under explicit topology. Received probe ledgers retain authenticated account identities such as `account:zeta`, while directional probe edges use stable scenario labels such as `zeta`. The probe now normalizes its correlated ledger copy to the scenario label after matching the logical application id.

This PR does not modify `cgka-engine`, production convergence behavior, or app runtime behavior.

## Planning and assurance

Updates the decision-route inventory, route matrix, scenario documentation, and convergence tracking plan. It records #1295 and #1347 as merged, credits the retained-history checkpoint, and keeps app-runtime, process, container, VM, multi-relay, retention-boundary, and durable-transition permutations open.

## Validation

- strict file-backed CLI replay: 15 expectations passed
- retained cross-route canonical regression
- route-assurance and Scenario IR tests
- report-runner, stateful-generator, and vector-artifact tests
- simulator unit and canonical scenario suites
- `just fast-ci`

## Summary by CodeRabbit

- **New Features**
  - Added cross-route retained-history recovery coverage with hidden relay events, divergent updates, and four-party synchronization.
  - Added validation for restart recovery, incremental and full-history replay, convergence, decryptability, durable dispositions, and pending work.
  - Added SQLite-backed end-to-end scenario validation and protection against substituted packets.

- **Bug Fixes**
  - Improved sender attribution across shared accounts and devices.
  - Rejected decryptability results from incorrect authenticated accounts.

- **Documentation**
  - Documented scenario replay, route-assurance status, reliability evidence, and remaining coverage gaps.


crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -1753,7 +1753,17 @@ impl ConvergenceSubject for EngineHarnessSubject {
...
+                    .cloned()
+                    .map(|mut entry| {
+                        // Received ledgers preserve the authenticated account
+                        // identity, which can differ from the scenario's
+                        // device/client label under an explicit topology. The
+                        // logical application id already correlates this entry
+                        // to the probe send above, so expose the stable scenario
+                        // label in this scenario-owned observation.
+                        entry.sender = sender.clone();
+                        entry
```
Datawav: This overwrites the authenticated sender with the expected scenario label after matching only on `logical_id`. A probe authenticated as a different account could therefore satisfy the check and mask sender-attribution or topology-isolation defects. Map the authenticated account through the scenario topology and require it to equal the expected client before normalization; reject mismatches and add a matching-logical-ID/wrong-sender regression test.

Datawav (commented):
## Automated review

One actionable finding:

- **Medium** — `crates/cgka-conformance-simulator/src/subject.rs:1756–1765` finds the received ledger entry using only `logical_id`, then overwrites its authenticated `sender` with the expected scenario label. A message authenticated as another account can therefore satisfy the probe, masking sender-attribution or topology-isolation defects across explicit-topology engine and retained-relay scenarios. Map the authenticated account through the scenario topology and require it to equal the expected client before normalization; reject mismatches and add a matching-logical-ID/wrong-sender regression test.

crates/cgka-conformance-simulator/vectors/generated-inputs/cross-route-retained-history-recovery.generated-input.json
```diff
@@ -0,0 +1,292 @@
+{
```
erskingardner, line 1: Non-blocking, but this artifact lands under `vectors/` while being invisible to every index that directory has.

- `manifest.v1.json` is not updated. `vectors/AGENTS.md` says to keep the manifest current "when adding a scenario, generated family, formal case fixture, or byte-level vector," and the sibling `cross-route-own-commit-recovery/v1` does have an entry. Nothing catches the omission: `vector_manifest_artifacts_exist_and_byte_fixtures_are_well_formed` only asserts coverage for *top-level* `*.v1.json` files.
- `collect_vector_fixture_directory` recurses into every subdirectory except `byte-fixtures`, and `is_vector_fixture_file` gates on the `.v1.json` suffix. The `.generated-input.json` name is what keeps `--vectors crates/cgka-conformance-simulator/vectors` from trying to parse this as a `VectorFixture`. That is load-bearing and currently undocumented, so a later rename to `...recovery.v1.json` would break the directory sweep.

Suggest either adding a manifest entry (a new `kind`, e.g. `generated_scenario_input`) or adding one rule line to `vectors/AGENTS.md` recording that `generated-inputs/` holds subject-pinned `GeneratedScenarioInputV1` files, that they are deliberately outside the manifest and the fixture sweep, and that the non-`.v1.json` suffix is what enforces that.

While in the manifest: the `cross-route-own-commit-recovery/v1` entry's `next` field still reads "Run the same logical topology through retained-history, app-runtime, process, container, and VM adapters" — the retained-history half is exactly what this PR delivers, so that string is now stale.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -871,6 +876,24 @@ impl EngineHarnessSubject {
...
+    ) -> Option<ScenarioInputLedgerEntry> {
+        let authenticated_identity = self.authenticated_identity_by_client.get(expected_client)?;
+        let mut entry = entries
+            .iter()
+            .find(|entry| {
+                entry.logical_id.as_ref() == logical_id.as_ref()
+                    && entry.sender == authenticated_identity.as_str()
+            })?
+            .clone();
+        entry.sender = expected_client.to_owned();
```
erskingardner: The mapping and rejection are right, and given that an explicit account topology deliberately shares one credential across an account's devices, the probe edge really is the only layer that can recover the client label — normalizing here rather than at the ledger source is the correct call.

One consequence worth a comment though: `ScenarioInputLedgerEntry` now carries two different sender conventions inside a single serialized report. Probe copies under `decryptability.probes[].recipient_ledger` say `zeta`; the `observe_exact` copies under `observations[].scenario_input_ledger` still say `account:zeta`. A future portable expectation that correlates the two by sender would silently never match.

The field doc on `ScenarioInputLedgerEntry::sender` ("Stable scenario label when known, otherwise the sender identity hex") is also already inaccurate for the observation path under explicit account topology — `account:zeta` is neither. A short doc line on `attributed_probe_ledger` plus a correction on that field would pin the convention cheaply.

Also a naming nit: `expected_client` is the probe *sender*, while `entries` is the *recipient's* ledger. `expected_sender` would read less ambiguously at the call site.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -2256,6 +2277,68 @@ mod tests {
         );
     }
 
+    #[test]
+    fn decryptability_probe_rejects_matching_logical_id_from_wrong_account() {
+        let labels = vec!["alice".to_owned(), "bob".to_owned()];
+        let topology = crate::ScenarioTopologyV2 {
```
erskingardner: This topology block is a near-verbatim copy of the one in `explicit_topology_shares_account_identity_across_devices` above — the only real difference is one account per client instead of one shared account. About 28 of this test's ~60 lines are that boilerplate.

A small module-local helper (`fn account_topology(labels: &[String], account_for: impl Fn(&str) -> String) -> ScenarioTopologyV2`, or just an `accounts_per_client` variant) would collapse both call sites and keep the actual assertion — matching logical id, wrong authenticated account, `None` — as the visible content of the test.

erskingardner (commented):
Recommend merge (posted as a comment because GitHub will not let me approve your own PR). This is a well-scoped assurance checkpoint: no engine, production convergence, or app-runtime code changes, one small simulator-observation fix, and a fixed subject-pinned input that genuinely forces retained-relay delivery rather than letting the packet bus heal the schedule. The earlier review concerns are addressed at this head: `attributed_probe_ledger` now requires the authenticated account to match the expected client before normalizing (with a wrong-account regression), the partial-view assertion is pinned to the last pre-full-history EOSE sync with exact `events_returned`/`injected_objects`, and the README states plainly that `--strict-oracle` owns only the portable outcomes while the Rust regression owns the retained-injection and disposition subset. CI is green across the matrix.

I checked the premise as well as the code. The scenario is not redundant with `cross-route-own-commit-recovery/v1`: the step list drives visibility and cursor advancement instead of semantic withholding, and the four `FullHistoryQueried` observations with `injected_objects > 0` are what make that distinction falsifiable. The plan/matrix/inventory edits match what actually landed, and the #1295 / #1347 references are accurate (both merged).

Three non-blocking notes inline: the new artifact is invisible to every `vectors/` index, one serialized type now carries two sender conventions, and a test topology block is duplicated. None of them should hold up the merge.

PR #1356, opened 2026-08-10
Fix OpenClaw messaging across stable and beta hosts

erskingardner:
## What was broken

The OpenClaw Marmot channel could connect successfully and receive an inbound Marmot message, then fail before handing that message to the agent. This affected hosts running OpenClaw 2026.7.2 beta: the beta turn kernel requires prepared dispatches to declare lifecycle ownership, while the plugin built against stable 2026.7.1 did not provide that declaration.

The beta also changed the optional inbound debounce contract and removed declarations from the dedicated keyed-queue SDK subpath. Separately, successful Marmot sends never updated `lastOutboundAt`, so the channel probe could misleadingly report no outbound activity even after a durable send succeeded.

## Root cause

The adapter and its tests were coupled to the pinned stable SDK shape. Focused dispatch tests mocked `runChannelInboundEvent`, so they did not exercise the beta kernel assertion that rejected the prepared turn. The optional debouncer had the same stable-only assumption: stable awaits a promise, while beta expects explicit admission and completion promises.

## What changed

- Declare a lifecycle-neutral prepared dispatch descriptor. Stable ignores this additive field; beta validates and accepts it.
- Adapt the debounce callback structurally so the same packaged plugin supports both host contracts without runtime version detection or beta-only imports.
- Import `KeyedAsyncQueue` through the SDK core barrel exported with declarations by both versions.
- Preserve privacy-safe dispatch failure classifications instead of reducing every failure to the same message.
- Record `lastOutboundAt` only after a durable text or media receipt is created.
- Add an unmocked installed-host kernel test and a CI compatibility run against `openclaw@2026.7.2-beta.7`.
- Move the connector E2E root to a short, user-owned cache path so private-root validation and Unix socket path limits are both respected.

## Impact

One plugin artifact can now run on the supported stable and beta OpenClaw hosts. Inbound messages reach the agent turn on beta, optional debouncing works on both contracts, and outbound channel status reflects successful durable sends.

## Validation

- `just openclaw-dev-test` — 221 passed, 2 skipped on pinned stable `2026.7.1-2`
- `just openclaw-host-compat-test` — beta build plus 74 focused tests passed on `2026.7.2-beta.7`
- `just openclaw-dev-e2e-connector` with an isolated root — real `wn-agent` connector E2E passed
- `just fast-ci`
- `git diff --check`, shell syntax validation, and available workflow/shell linters

## Summary by CodeRabbit

* **Compatibility**
  * Improved support for stable and beta OpenClaw host APIs, inbound events, and lifecycle handling.
  * Added automated compatibility checks for dispatch, debouncing, and reply behavior.

* **Reliability**
  * Improved handling of queued and skipped dispatches.
  * Outbound text and media messages now update account activity state reliably and independently.

* **Privacy**
  * Failure logs now use safe categories instead of exposing error details.

* **Documentation**
  * Added a development command and workflow for running host compatibility checks.

### DIFF
--- a/.github/workflows/wn-agent-binaries.yml
+++ b/.github/workflows/wn-agent-binaries.yml
@@ -305,6 +305,9 @@ jobs:
             pnpm test
             pnpm build
           )
+          # Rebuild and exercise the SDK-coupled messaging seams against the
+          # beta host contract that differs from the pinned stable SDK.
+          integrations/openclaw/marmot/test/openclaw-host-compat.sh
           # Provenance manifest (release.md requires one in each artifact); ships
           # via package.json#files alongside the built plugin.
           cat > integrations/openclaw/marmot/manifest.json <<EOF
--- a/.github/workflows/wn-agent-binaries.yml
+++ b/.github/workflows/wn-agent-binaries.yml
@@ -260,6 +260,26 @@ jobs:
       - name: Run real Hermes persisted-config probe
         run: ./scripts/hermes_marmot_verify_persisted_config.sh --root "$HERMES_MARMOT_DEV_ROOT"
 
+  openclaw-host-compat:
+    name: OpenClaw beta host compatibility
+    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          persist-credentials: false
+
+      - name: Test supported OpenClaw beta host
+        shell: bash
+        run: |
+          set -euo pipefail
+          corepack enable
+          corepack prepare pnpm@10.13.1 --activate
+          integrations/openclaw/marmot/test/openclaw-host-compat.sh
+
   openclaw-plugin:
     name: Package OpenClaw plugin
     runs-on: ubuntu-latest
@@ -305,9 +325,6 @@ jobs:
             pnpm test
             pnpm build
           )
-          # Rebuild and exercise the SDK-coupled messaging seams against the
-          # beta host contract that differs from the pinned stable SDK.
-          integrations/openclaw/marmot/test/openclaw-host-compat.sh
           # Provenance manifest (release.md requires one in each artifact); ships
           # via package.json#files alongside the built plugin.
           cat > integrations/openclaw/marmot/manifest.json <<EOF
--- a/integrations/openclaw/marmot/src/bounded-keyed-async-queue.ts
+++ b/integrations/openclaw/marmot/src/bounded-keyed-async-queue.ts
@@ -1,6 +1,30 @@
-import { KeyedAsyncQueue } from "openclaw/plugin-sdk/keyed-async-queue";
+// The dedicated keyed-async-queue subpath lost its declarations in
+// 2026.7.2-beta. The core barrel exports the same runtime value and declarations
+// on both stable and beta.
+import { KeyedAsyncQueue } from "openclaw/plugin-sdk/core";
 
 export const DEFAULT_INBOUND_QUEUE_MAX_DEPTH = 32;
+
+const OPENCLAW_DISPATCH_LIFECYCLE_ERROR =
+  "runChannelInboundEvent prepared turns must declare runDispatchLifecycle when creating runDispatch";
+
+/**
+ * Reduce an arbitrary dispatch failure to a fixed, privacy-safe class. Never
+ * include Error.message: upstream failures can embed prompts, paths, ids, or
+ * other conversation-specific data.
+ */
+export function classifyInboundDispatchFailure(error: unknown): string {
+  if (error instanceof Error && error.message === OPENCLAW_DISPATCH_LIFECYCLE_ERROR) {
+    return "openclaw_dispatch_lifecycle_contract";
+  }
+  if (error instanceof TypeError) {
+    return "type_error";
+  }
+  if (error instanceof Error) {
+    return "error";
+  }
+  return "non_error";
+}
 
 /**
  * Per-key FIFO dispatch with a bounded queue depth. When a key is at capacity,
@@ -39,6 +63,10 @@ export class BoundedKeyedAsyncQueue {
           }
         }
       })
-      .catch(() => this.onShed?.("marmot: inbound dispatch task failed"));
+      .catch((error: unknown) =>
+        this.onShed?.(
+          `marmot: inbound dispatch task failed (class=${classifyInboundDispatchFailure(error)})`,
+        ),
+      );
   }
 }
--- a/integrations/openclaw/marmot/src/bounded-keyed-async-queue.ts
+++ b/integrations/openclaw/marmot/src/bounded-keyed-async-queue.ts
@@ -5,6 +5,8 @@ import { KeyedAsyncQueue } from "openclaw/plugin-sdk/core";
 
 export const DEFAULT_INBOUND_QUEUE_MAX_DEPTH = 32;
 
+// Copied from the OpenClaw host assertion. This classification is best-effort:
+// an upstream wording change safely degrades to the generic `error` bucket.
 const OPENCLAW_DISPATCH_LIFECYCLE_ERROR =
   "runChannelInboundEvent prepared turns must declare runDispatchLifecycle when creating runDispatch";
 
@@ -16,9 +18,6 @@ const OPENCLAW_DISPATCH_LIFECYCLE_ERROR =
 export function classifyInboundDispatchFailure(error: unknown): string {
   if (error instanceof Error && error.message === OPENCLAW_DISPATCH_LIFECYCLE_ERROR) {
     return "openclaw_dispatch_lifecycle_contract";
-  }
-  if (error instanceof TypeError) {
-    return "type_error";
   }
   if (error instanceof Error) {
     return "error";
--- a/integrations/openclaw/marmot/src/inbound-runtime.ts
+++ b/integrations/openclaw/marmot/src/inbound-runtime.ts
@@ -12,37 +12,8 @@
 import { createInboundDebouncer } from "openclaw/plugin-sdk/channel-inbound-debounce";
 import type { ChannelAccountSnapshot } from "openclaw/plugin-sdk/status-helpers";
 
-/**
- * OpenClaw stable awaits an onFlush Promise directly. 2026.7.2-beta instead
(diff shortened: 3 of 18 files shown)


integrations/openclaw/marmot/src/runtime-state.ts
```diff
@@ -63,6 +63,21 @@ export function markMarmotInboundReceived(accountId?: string | null): void {
...
 
+/** Record a durable outbound receipt for channel status/probe reporting. */
+export function markMarmotOutboundSent(
+  accountId?: string | null,
+  sentAt: number = Date.now(),
+): void {
+  const nextAccountId = accountIdOrDefault(accountId);
+  const base = inboundRuntime.accountId === nextAccountId
+    ? inboundRuntime
+    : stoppedSnapshot(nextAccountId);
```
erskingardner: An outbound send can now evict a running account's inbound status.

When `ctx.accountId` differs from the tracked account, this rebases the single global slot onto `stoppedSnapshot(nextAccountId)`, so `running`, `connected`, `lastStartAt`, and `lastInboundAt` for the previously-tracked account are discarded. Multi-account is a supported shape (`channels.marmot.accounts.<id>`, see `resolveMarmotChannelAccount`), so: inbound is live on `work`, the agent sends on `personal`, and `marmotInboundRuntimeSnapshot("work")` now reports a stopped, disconnected channel that is in fact running and subscribed.

That is the same class of misleading probe output this PR set out to fix, arriving from the other direction — and it is new, because before this change only inbound lifecycle events could move the slot; now a send can.

Rebasing onto `stoppedSnapshot` is the careful choice *given* that you write at all (it avoids attributing A's `running` to B, which `markMarmotInboundReceived` above does do). The simpler answer is not to write: a send on B has nothing useful to say about the slot when the slot belongs to A.

```ts
export function markMarmotOutboundSent(
  accountId?: string | null,
  sentAt: number = Date.now(),
): void {
  if (inboundRuntime.accountId !== accountIdOrDefault(accountId)) {
    return;
  }
  inboundRuntime = { ...inboundRuntime, lastOutboundAt: sentAt };
}
```

Single-account behavior is unchanged (`ctx.accountId` is null/`"default"`, which matches the tracked id), the new tests still pass, and it is a few lines shorter.

integrations/openclaw/marmot/test/openclaw-host-compat.sh
```diff
@@ -0,0 +1,40 @@
...
+set -euo pipefail
+
+plugin_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+beta_version="2026.7.2-beta.7"
+compat_root="$(mktemp -d "${TMPDIR:-/tmp}/openclaw-host-compat.XXXXXX")"
+trap 'rm -rf "$compat_root"' EXIT
+
+# Fail with a direct diagnostic if the pinned beta has disappeared instead of
+# letting dependency installation fail later with a less useful resolver error.
+published_version="$(npm view "openclaw@$beta_version" version)"
```
erskingardner: This guard cannot emit its diagnostic.

`npm view <pkg>@<missing-version> version` exits 1 with `E404` rather than printing a different version (verified: `npm view lodash@99.99.99-nope version` -> exit 1). Under `set -euo pipefail` the command substitution's failure aborts the script at this line, so lines 12-15 are unreachable and the operator sees npm's resolver error — precisely the outcome the comment says this exists to avoid.

Either drop the check (the `--frozen-lockfile` install fails clearly enough, and this removes a network round trip from the packaging job), or make it able to run:

```sh
published_version="$(npm view "openclaw@$beta_version" version 2>/dev/null || true)"
```

.github/workflows/wn-agent-binaries.yml
```diff
@@ -305,6 +305,9 @@ jobs:
             pnpm test
             pnpm build
           )
+          # Rebuild and exercise the SDK-coupled messaging seams against the
+          # beta host contract that differs from the pinned stable SDK.
+          integrations/openclaw/marmot/test/openclaw-host-compat.sh
```
erskingardner: Consider whether this belongs in the release-packaging job.

`openclaw-plugin` runs on release tags and produces the shipped tarball. As written, packaging now fails — blocking a wn-agent/MarmotKit cohort release — if either:

- upstream unpublishes or prunes `openclaw@2026.7.2-beta.7` (prereleases routinely disappear), or
- the plugin's `package.json` gains or bumps any dependency without a matching regeneration of `test/pnpm-lock.openclaw-beta.yaml`, since `--frozen-lockfile` rejects specifier drift.

Neither failure says anything about the artifact being packaged. A separate job (or the same script invoked from the PR test workflow) gives the same signal without putting a third-party prerelease on the release critical path.

If it stays here, please add one line to `integrations/openclaw/marmot/AGENTS.md` on how to regenerate the beta lockfile — 3.5k lines of committed lock with no documented refresh procedure is a trap for whoever hits the first specifier drift.

integrations/openclaw/marmot/src/bounded-keyed-async-queue.ts
```diff
@@ -1,7 +1,31 @@
...
+/**
+ * Reduce an arbitrary dispatch failure to a fixed, privacy-safe class. Never
+ * include Error.message: upstream failures can embed prompts, paths, ids, or
+ * other conversation-specific data.
+ */
+export function classifyInboundDispatchFailure(error: unknown): string {
+  if (error instanceof Error && error.message === OPENCLAW_DISPATCH_LIFECYCLE_ERROR) {
+    return "openclaw_dispatch_lifecycle_contract";
+  }
+  if (error instanceof TypeError) {
```
erskingardner: Two small simplifications here.

`type_error` is never distinguished by any consumer — the value is only interpolated into a shed log — so this branch adds a class nobody reads. Dropping it leaves `openclaw_dispatch_lifecycle_contract` / `error` / `non_error`, which is the distinction that actually motivated the change.

Separately, `OPENCLAW_DISPATCH_LIFECYCLE_ERROR` matches an upstream message verbatim. Any rewording in a later OpenClaw release silently degrades the lifecycle class to `error`, and no test would catch it (the unit test constructs the string itself). Worth a comment noting the string is copied from the host and is best-effort, so the next person does not read a missing class as "the lifecycle contract is fine".

integrations/openclaw/marmot/src/inbound-runtime.ts
```diff
@@ -37,6 +37,54 @@ import {
...
+  createInboundDebouncer as unknown as CompatibleInboundDebouncerFactory;
+
+/**
+ * Adapt one coalesced batch to stable's Promise contract or beta's lifecycle
+ * pair. The beta completion deliberately means "admitted to Marmot's bounded
+ * per-group queue", not "the agent turn settled": the outer queue owns turn
+ * serialization and failure logging, matching stable's fire-and-forget seam.
+ * The factory-provided lifecycle argument is therefore intentionally ignored.
+ */
+export function createCompatibleInboundDebounceFlush<T>(
```
erskingardner: This helper and the `export` on `createCompatibleInboundDebouncer` (plus the `flushKey` member added to the factory type) exist only so `openclaw-host-contract.test.ts` can reach them. The first commit's inline closure inside `onFlush` was shorter and read more directly at its single call site; the extracted three-argument generic is more surface for the same behavior.

If the beta-gated test can drive this through `startMarmotInbound` instead, the module keeps its internals internal. Not blocking — the documented "completion means admitted to the bounded queue, not turn settled" reasoning is the right call and is worth having written down somewhere.

integrations/openclaw/marmot/test/e2e-connector.test.ts
```diff
@@ -104,7 +105,13 @@ maybeDescribe("OpenClaw Marmot connector E2E", () => {
   it(
     "keeps sequential and restarted inbound replies on the source Marmot group",
     async () => {
-      const tempRoot = await mkdtemp("/tmp/omce-");
+      // wn-agent requires every ancestor of its exclusive home to be owned and
+      // non-writable by other users, and Unix sockets have a short path limit.
+      // A direct /tmp child fails the first invariant while a deep worktree path
+      // can fail the second, so use the user's short, owned cache root.
+      const cacheRoot = join(homedir(), ".cache");
```
erskingardner: Nit: `XDG_CACHE_HOME` is ignored, so on a machine that relocates the cache this writes to a `~/.cache` the user does not otherwise use. `process.env.XDG_CACHE_HOME ?? join(homedir(), ".cache")` keeps the short-owned-path property the comment is after. The reasoning for moving off `/tmp` is good and the comment explaining both constraints is exactly right.

erskingardner (commented):
Targeted, well-reasoned compatibility fix. The root cause is named precisely (beta kernel requires `runDispatchLifecycle`; beta debounce wants an admission/completion pair; the keyed-queue subpath lost its declarations), each change is the minimal additive shape that keeps one artifact running on both hosts, and the second commit's move to Marmot's real dispatcher in `openclaw-host-contract.test.ts` closes the "the regression can be deleted without host-compat noticing" gap from the earlier review. CI confirms both lanes green, with the beta-only flush-pair test actually executing (2 tests, 0 skipped) on `2026.7.2-beta.7`.

No merge blockers. Four things worth a look, three of which are net-deletions:

1. `markMarmotOutboundSent` lets an outbound send evict a *running* account's inbound status. Cheap fix, and it removes a branch.
2. The `npm view` guard in the compat script is unreachable under `set -e` — the custom diagnostic can never print.
3. Putting the beta lane inside the release-packaging job couples wn-agent releases to a prerelease staying published and to a hand-maintained 3.5k-line lockfile staying in sync. Worth a deliberate decision, not necessarily a change.
4. Minor: brittle error-message matching, an unused failure bucket, and test-only exports.

On premise: the change is clearly needed now (inbound messages do not reach the agent on the beta host), and the scope is right — it fixes the contract mismatch rather than pinning users to stable.

erskingardner (commented):
Final pass at `a782061d`: **not ready to merge yet**. CI is green and the original compatibility feedback is implemented, but the exact-head review still has one correctness issue and one broken hardening path:

- `markMarmotOutboundSent` can replace a running account's process-local status with a stopped snapshot for another account, making a live multi-account channel appear disconnected. Preserve the tracked account when the outbound receipt belongs to a different account (or move this compatibility state to a per-account map) and add the cross-account regression.
- The `npm view` guard cannot print its promised diagnostic under `set -e`; a missing version exits during command substitution. Catch the failure or remove the redundant probe.

The release-critical-path placement of the beta probe, beta-lock regeneration documentation, XDG cache-root handling, failure-bucket simplification, and test-only exports are non-blocking follow-ups, but their unresolved threads should be dispositioned explicitly. I would re-check after the two concrete fixes above.

PR #1357, opened 2026-08-10
Generalize isolated convergence campaigns

erskingardner:
## Summary

- generalize `cgka-conformance-campaign` from the adversarial family to all six generated scenario families
- run every case in an isolated child process with bounded wall-clock time and process resource accounting
- persist the exact generated input before execution, then validate report and fixture provenance
- retain portable failure capsules and record artifact-integrity failures in the campaign summary
- add direct per-case generators so campaign construction remains linear rather than regenerating family prefixes
- document the workflow and update the convergence reliability tracking plan

## Why

The generated families now have strict oracle coverage, but larger discovery runs still need a common isolation and evidence boundary. This makes timeouts, crashes, normal oracle failures, and incomplete artifacts distinguishable and reproducible across both memory and file-backed storage.

This PR changes simulator/testing code and documentation only. It does not change production convergence-engine or application behavior.

## Validation

- `cargo test -p cgka-conformance-simulator --bin cgka-conformance-campaign`
- `cargo test -p cgka-conformance-simulator --test process_campaign_runner`
- `cargo test -p cgka-conformance-simulator --test canonical_scenarios family`
- one isolated smoke case for each of the six generated families
- isolated file-backed `send-leave/v1` smoke case
- `just fast-ci`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added isolated, case-by-case campaign execution across multiple scenario families.
  * Added generated-input files, configurable worker timeouts, storage options, and sensitive replay capture.
  * Added detailed per-case reports, provenance metadata, failure capsules, replay artifacts, and integrity validation.
  * Added direct generation for individual scenario cases without regenerating earlier cases.

* **Bug Fixes**
  * Campaigns now reject invalid configurations, incomplete artifacts, mismatched inputs, and attempts to overwrite existing evidence.
  * Strict worker failures now return a nonzero result with a recorded failure capsule.

* **Documentation**
  * Expanded simulator usage instructions and updated reliability-plan progress tracking.

### DIFF
--- a/crates/cgka-conformance-simulator/README.md
+++ b/crates/cgka-conformance-simulator/README.md
@@ -163,10 +163,13 @@ cargo run -p cgka-conformance-simulator --bin cgka-conformance-campaign -- \
 cargo run -p cgka-conformance-simulator --bin cgka-conformance-campaign -- \
   --family convergence-chaos/v1 --seed 42 --cases 22 --case-timeout-secs 300 \
   --out target/cgka-convergence-chaos-process-campaign --storage file
+```
 
-# Use a new output directory for each run. The runner refuses to overwrite a
-# prior summary, generated input, report, fixture, or failure capsule.
+Use a new output directory for each ad-hoc run. The runner refuses to overwrite a prior summary, generated input,
+report, fixture, or failure capsule. The committed `adversarial-reliability-ci` and `convergence-weekly-lane` gates
+remove their fixed output directories before invoking the runner, so those recipes remain rerunnable.
 
+```sh
 # Test-only one-variable policy curves around fixed retained inputs/horizons.
 cargo test -p cgka-conformance-simulator --features test-policy-overrides \
   --test policy_sweeps
--- a/crates/cgka-conformance-simulator/README.md
+++ b/crates/cgka-conformance-simulator/README.md
@@ -156,6 +156,17 @@ cargo run -p cgka-conformance-simulator --bin cgka-conformance-campaign -- \
   --seed 7 --cases 12 --case-timeout-secs 300 \
   --out target/cgka-adversarial-reliability-process-campaign --storage file
 
+# Isolate any generated family case-by-case. The parent writes each exact
+# generated input before spawning its deadline-bounded worker; workers write
+# the same reports, fixture candidates, and portable failure capsules as the
+# report CLI.
+cargo run -p cgka-conformance-simulator --bin cgka-conformance-campaign -- \
+  --family convergence-chaos/v1 --seed 42 --cases 22 --case-timeout-secs 300 \
+  --out target/cgka-convergence-chaos-process-campaign --storage file
+
+# Use a new output directory for each run. The runner refuses to overwrite a
+# prior summary, generated input, report, fixture, or failure capsule.
+
 # Test-only one-variable policy curves around fixed retained inputs/horizons.
 cargo test -p cgka-conformance-simulator --features test-policy-overrides \
   --test policy_sweeps
@@ -163,6 +174,12 @@ cargo test -p cgka-conformance-simulator --features test-policy-overrides \
 # Independent convergence model, liveness, mutation, and protocol-decision gate.
 just convergence-verification-ci
 ```
+
+The process-campaign summary records the selected family, seed, storage mode, timeout, sensitive-capture setting,
+per-case process measurements, and every generated-input/report/fixture/capsule path. A normally exiting worker must
+leave a parseable report whose generated metadata and source digest match the saved input, plus its fixture candidate;
+otherwise `artifact_integrity_errors` makes the campaign fail. Case indices are generated directly rather than by
+rebuilding and discarding the complete earlier prefix, so increasing `--cases` keeps generation work linear.
 
 Every `ScenarioReport` embeds `campaign_measurements` v1. Its convergence latency is the measured wall time of the
 final successful `await_quiescence` action, or explicitly unavailable when the scenario does not request one;
--- a/crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
+++ b/crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
@@ -1,6 +1,9 @@
 use cgka_conformance_simulator::{
-    HarnessStorageMode, ScenarioReport, generate_adversarial_reliability_case,
-    run_generated_case_report_with_storage_mode,
+    GeneratedScenarioCase, GeneratedScenarioInputV1, HarnessStorageMode, ReportArgs, ReportInput,
+    ScenarioReport, generate_admin_churn_case, generate_adversarial_reliability_case,
+    generate_convergence_chaos_case, generate_convergence_e2e_delivery_case,
+    generate_send_leave_case, generate_stateful_chat_journey_case, resolve_scenario_input_bytes,
+    run_report,
 };
 use serde::{Deserialize, Serialize};
 use std::error::Error;
@@ -10,26 +13,39 @@ use std::time::{Duration, Instant};
 
 #[derive(Debug)]
 struct Args {
+    family: String,
     seed: u64,
     cases: usize,
-    case_index: Option<usize>,
     out: PathBuf,
     storage: HarnessStorageMode,
     case_timeout: Duration,
+    input: Option<PathBuf>,
+    capture_sensitive_replay: bool,
     worker: bool,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 struct ProcessCampaignReportV1 {
     schema_version: String,
+    family: String,
     seed: u64,
+    storage: String,
+    case_timeout_ms: u64,
+    capture_sensitive_replay: bool,
     cases: Vec<ProcessCaseMeasurementV1>,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 struct ProcessCaseMeasurementV1 {
     case_index: usize,
+    generated_input: PathBuf,
     report: PathBuf,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    fixture_candidate: Option<PathBuf>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    failure_capsule: Option<PathBuf>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    sensitive_replay_capsule: Option<PathBuf>,
     exit_code: Option<i32>,
     signal: Option<i32>,
     timed_out: bool,
@@ -41,6 +57,16 @@ struct ProcessCaseMeasurementV1 {
     filesystem_block_write_lower_bound_bytes: Option<u64>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     unavailable_process_fields: Vec<String>,
+    #[serde(default)]
+    artifact_integrity_errors: Vec<String>,
+}
+
+struct CaseArtifactPaths {
+    generated_input: PathBuf,
+    report: PathBuf,
+    fixture_candidate: PathBuf,
+    failure_capsule: PathBuf,
+    sensitive_replay_capsule: PathBuf,
 }
 
 #[tokio::main]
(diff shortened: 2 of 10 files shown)


crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
```diff
@@ -59,31 +85,68 @@ async fn run() -> Result<ExitCode, Box<dyn Error>> {
...
+    #[cfg(unix)]
+    let _output_dir_guard = fs_private::prepare_directory_path(
+        &args.out,
+        fs_private::PRIVATE_DIR_MODE,
+        fs_private::ExistingDirectoryMode::Preserve,
+    )?;
+    #[cfg(not(unix))]
     fs_private::create_dir_all_private(&args.out)?;
+    let summary_path = args.out.join("process-campaign.v1.json");
+    ensure_path_absent(&summary_path)?;
```
erskingardner: **Blocking: this makes the two existing campaign gates non-re-runnable.**

`Justfile:309` (`adversarial-reliability-ci`) and `Justfile:338` (`convergence-weekly-lane`) both invoke this binary with a *fixed* output directory:

```
--out target/cgka-adversarial-reliability-ci
--out target/cgka-weekly-reliability
```

Before this PR `write_private` simply overwrote, so those recipes were idempotent. Now the second invocation dies at this line with `refusing to overwrite existing campaign artifact: .../process-campaign.v1.json` and exits 1.

- Locally this is certain: `just adversarial-reliability-ci` twice in a row now fails the second time.
- On CI it is very likely too: `simulator-nightly.yml` and `convergence-hardening.yml` both use `Swatinem/rust-cache`, which caches `./target` and does not prune non-profile directories under it. Once a nightly run bakes `target/cgka-adversarial-reliability-ci/` into the cache, every later restore-hit run fails this step until the cache key rotates.

`convergence-nightly-lane` depends on `adversarial-reliability-ci`, so this takes the whole nightly lane down.

Please pick one and land it with the PR:
- have the recipes clean their own directory first (`rm -rf target/cgka-adversarial-reliability-ci` / `target/cgka-weekly-reliability`) or use a per-run subdirectory, or
- add an explicit `--overwrite` / `--fresh` flag and set it in the recipes.

The README's new "use a new output directory for each run" note covers the ad-hoc invocation but not these two committed callers.

crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
```diff
@@ -174,17 +329,76 @@ fn parse_args(args: impl Iterator<Item = String>) -> Result<Args, Box<dyn Error>
...
+        "send-leave/v1" => generate_send_leave_case(seed, case_index),
+        "convergence-e2e-delivery/v1" => generate_convergence_e2e_delivery_case(seed, case_index),
+        "convergence-chaos/v1" => generate_convergence_chaos_case(seed, case_index),
+        "admin-churn/v1" => generate_admin_churn_case(seed, case_index),
+        "adversarial-reliability/v1" => generate_adversarial_reliability_case(seed, case_index),
+        "chat-journey/v1" => generate_stateful_chat_journey_case(seed, case_index),
+        other => return Err(format!("unsupported family {other}").into()),
+    };
+    Ok(case)
+}
```
erskingardner, line 390: This six-arm family match is now a duplicate of `run_generated_family_reports` in `src/report.rs:256-264`. Two lists that must stay in sync will drift the first time a seventh family lands — and the failure mode is silent (`--family x/v1` works in the report CLI, `unsupported family x/v1` in the campaign).

Since every family now has a direct per-case generator, export one `generate_family_case(family, seed, case_index) -> Result<GeneratedScenarioCase, _>` from the library, call it here, and reduce `run_generated_family_reports` to `(0..cases).map(|i| generate_family_case(family, seed, i as u64))`. That deletes both lists and the six `generate_*_family` calls in `report.rs`.

crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
```diff
@@ -59,31 +85,68 @@ async fn run() -> Result<ExitCode, Box<dyn Error>> {
...
+        let case = generate_case(&args.family, args.seed, case_index)?;
+        let paths = case_artifact_paths(&args.out, &case);
+        for path in [
+            &paths.generated_input,
+            &paths.report,
+            &paths.fixture_candidate,
+            &paths.failure_capsule,
+            &paths.sensitive_replay_capsule,
+        ] {
+            ensure_path_absent(path)?;
```
erskingardner: Validation happens after the output directory is created and, for cases 1..n, after earlier workers have already run. Two consequences:

- An unsupported `--family` (checked at `generate_case`, one line above) still creates the private output directory first.
- Any mid-loop `?` here — a collision, a `write_private` failure, a `spawn` failure — returns from `run()` without writing `process-campaign.v1.json`, so every completed case's measurements are discarded even though its artifacts are still on disk with no index.

That is the shape `docs/marmot-architecture/overview/multi-step-state-changes.md` asks you to avoid (validate before mutating). Cheapest fix: resolve the family and all per-case artifact paths in one preflight pass before creating the directory or spawning anything; alternatively write the partial summary on the error path.

crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
```diff
@@ -59,31 +85,68 @@ async fn run() -> Result<ExitCode, Box<dyn Error>> {
...
+            .arg(storage_label(args.storage));
+        if args.capture_sensitive_replay {
+            command.arg("--capture-sensitive-replay");
+        }
         let started = Instant::now();
         let child = command.spawn()?;
         let usage = wait_with_usage(child, args.case_timeout)?;
-        let database_bytes = read_database_bytes(&report);
+        let artifact_integrity_errors = inspect_case_artifacts(&case, &paths, &usage);
+        let database_bytes = read_database_bytes(&paths.report);
```
erskingardner: `inspect_case_artifacts` has already read and deserialized the full `ScenarioReport` one line above; `read_database_bytes` reads and deserializes it a second time. `ScenarioReport` carries the whole step log and observed trace, so for a 20+ client storm case this is a non-trivial double parse per case. Have `inspect_case_artifacts` return `(errors, database_bytes)` from its single parse and drop `read_database_bytes`.

crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs
```diff
@@ -116,14 +186,19 @@ async fn run() -> Result<ExitCode, Box<dyn Error>> {
...
+            paths: vec![input],
+            adapter: None,
+        },
+        out: args.out.clone(),
+        strict_oracle: true,
+        storage_mode: args.storage,
+        capture_sensitive_replay: args.capture_sensitive_replay,
+    })
+    .await?;
+    let failed = summary.failed() > 0;
```
erskingardner, line 198: The worker now has the full `ReportRunSummary` but stays silent, so a failing case surfaces only as `exit_code: 1` in the parent summary and the reason has to be dug out of the JSON report. The worker's stdout is inherited by the parent, so one line makes campaign logs self-explanatory:

```rust
println!("{}", summary.to_human_text());
```

crates/cgka-conformance-simulator/README.md
```diff
@@ -156,6 +156,17 @@ cargo run -p cgka-conformance-simulator --bin cgka-conformance-campaign -- \
...
+# Isolate any generated family case-by-case. The parent writes each exact
+# generated input before spawning its deadline-bounded worker; workers write
+# the same reports, fixture candidates, and portable failure capsules as the
+# report CLI.
+cargo run -p cgka-conformance-simulator --bin cgka-conformance-campaign -- \
+  --family convergence-chaos/v1 --seed 42 --cases 22 --case-timeout-secs 300 \
+  --out target/cgka-convergence-chaos-process-campaign --storage file
+
+# Use a new output directory for each run. The runner refuses to overwrite a
+# prior summary, generated input, report, fixture, or failure capsule.
```
erskingardner: This comment block sits inside the ```sh fence with no command under it, which reads as a truncated example. Move it out of the fence into the prose paragraph below, and — assuming the `Justfile` recipes get a clean/overwrite step — say explicitly that the committed gates handle this themselves.

erskingardner (commented):
**Recommendation: changes requested** (posting as a comment — GitHub does not allow requesting changes on your own PR).

Generalizing the isolated runner to every generated family is the right slice, and the per-case direct generators (with the `family(seed, n+1).pop()` equivalence test) are a clean, well-verified refactor. Test-only scope, no production surface touched.

One blocking issue: the new refuse-to-overwrite policy breaks the two `Justfile` gates that already point the campaign at fixed `target/` directories, and this PR doesn't update them. Details inline. The rest are non-blocking simplification/diagnosability notes.

crates/cgka-conformance-simulator/src/report.rs
```diff
@@ -253,18 +250,9 @@ async fn run_generated_family_reports(
...
-        "admin-churn/v1" => generate_admin_churn_family(seed, cases),
-        "adversarial-reliability/v1" => generate_adversarial_reliability_family(seed, cases),
-        "chat-journey/v1" => generate_stateful_chat_journey_family(seed, cases),
-        other => return Err(format!("unsupported family {other}").into()),
-    };
-
-    let mut summaries = Vec::with_capacity(cases.len());
-    for case in cases {
+    let mut summaries = Vec::with_capacity(cases);
+    for case_index in 0..cases {
```
erskingardner, line 254: **Non-blocking, new in `654b606a`:** moving family resolution into the loop means the family name is no longer validated when `cases == 0`.

`parse_report_command` defaults `cases` to `1` and has no zero guard, so:

```
cgka-conformance-simulator-report --family typo/v1 --cases 0
```

now walks an empty range and exits `0` with `CGKA conformance PASS: 0 passed, 0 failed, 0 total`, where before this commit the `match family` ran first and it failed with `unsupported family typo/v1`. A vacuous pass on a typo'd family is exactly the kind of thing a conformance CLI should not do quietly.

The campaign binary is unaffected — `parse_args` already rejects `--cases 0` for the parent.

Cheapest fix is a one-line probe before the loop:

```rust
generate_family_case(family, seed, 0)?;
```

or add the same `cases == 0` guard to `parse_report_command` that the campaign binary already has. Either is fine; the second also kills the pre-existing vacuous `--cases 0` pass for valid families.

erskingardner (commented):
**Re-review of `654b606a` — recommendation: approve / ready to merge** (posting as a comment because GitHub does not allow approving your own PR).

All six findings from the previous review are genuinely addressed, and I verified the blocking one end to end:

- **Blocking (idempotent gates) — fixed.** Both recipes now `rm -rf --` their fixed directory first. I confirmed the campaign binary has exactly two committed callers (`Justfile:310`, `Justfile:341`), that both CI entry points reach them through the recipes (`simulator-nightly.yml` → `just convergence-nightly-lane` → `adversarial-reliability-ci`; `convergence-hardening.yml` → `just convergence-weekly-lane`), and that `convergence-hardening.yml`'s evidence upload runs *after* the lane, so the clean step does not eat the uploaded artifacts.
- **Duplicate family match — fixed.** `generate_family_case` is now the single registry, with a typed `UnsupportedGeneratedFamily` error preserving the old message. I checked all six `generate_*_family` functions are literally `(0..cases).map(|i| generate_*_case(seed, i))`, so switching `run_generated_family_reports` to per-case generation is behavior-preserving rather than merely intended to be. The `*_family` wrappers are still used by `canonical_scenarios`, `adversarial_reliability_campaigns`, and `stateful_generator` tests, so nothing became dead.
- **Validate-before-mutate — fixed.** `preflight_campaign` resolves the family and every artifact path before the output directory is created or anything is spawned, and `preflight_rejects_an_unknown_family_without_creating_output` pins the no-directory guarantee.
- **Double `ScenarioReport` parse — fixed.** `CaseArtifactInspection` carries both results from the single parse, and the timeout/signal semantics are preserved exactly: the report is still read unconditionally so `database_bytes` survives an abnormal exit, while the integrity errors stay gated on `normally_exited`.
- **Silent worker — fixed.** `println!("{}", summary.to_human_text())` matches what `cgka-conformance-simulator-report` already does, and the child inherits stdout so it lands in campaign logs.
- **README fence — fixed**, and the note now explicitly documents that the committed gates clean up after themselves.

`isolated_campaign_preserves_input_provenance_and_refuses_overwrite` still covers the refusal path after the preflight reorder, and CI is green across the board.

One new, non-blocking edge case introduced by the `report.rs` refactor — details inline. Nothing else worth holding the PR for.

PR #1361, opened 2026-08-10
fix: preserve relay rejection diagnostics

### LINKED ISSUE #1155: KeyPackage NIP-09 deletion is rejected by source relays without a diagnosable reason
## Summary

Deleting a relay-backed KeyPackage from Android fails against both source relays:

`Couldn't delete key package details=publish failed: relay rejected event; relay rejected event`

MDK builds and publishes a NIP-09 request, but all target relays reject it. The transport then replaces each relay's response with the same constant string and joins both strings, leaving no structured signal to distinguish authentication, policy, event-shape, or relay-capability failure.

## Root cause boundary on current `master`

- `crates/marmot-app/src/lib.rs::delete_key_package_events` signs kind `5` with `e` + `k=30443`, targets the record's source relays, and requires one acknowledgement.
- `crates/transport-nostr-adapter/src/sdk_client.rs::send_event_to_relay` intentionally maps every relay rejection to `relay rejected event` to prevent remote-controlled text/URLs from leaking through display-bearing errors.
- The endpoint failures remain structured internally, but the display error loses endpoint/rejection-class correlation and duplicates the same reason once per relay.

NIP-09 permits `e` or `a` references and recommends `k`; therefore the observed rejection cannot safely be attributed to the missing `a` tag without reproducing against the actual relay policy. The current error abstraction prevents that diagnosis.

## Expected behavior

- A valid KeyPackage deletion is accepted by at least one compatible source relay; authenticate or adjust the NIP-09 request shape if the relay's supported policy requires it.
- Preserve a privacy-safe structured rejection category per endpoint for diagnostics/telemetry without exposing arbitrary relay-provided text in normal UI or logs.
- Collapse identical user-facing failures instead of rendering `relay rejected event` repeatedly.
- Do not remove local publication metadata until at least one deletion acknowledgement succeeds.

## Acceptance tests

1. Integration relay accepting NIP-09: deletion receives an acknowledgement and removes only the matching local publication metadata.
2. Relay requiring supported auth/policy: MDK takes the supported path and succeeds.
3. Two rejecting relays: diagnostics retain two endpoint-scoped sanitized failures while the UI receives one concise summary.
4. Mixed accepted/rejected relays: one acknowledgement satisfies the operation and the rejected endpoint remains diagnostic only.

## Related

- The empty-event-id failure for retained local material is a separate Android actionability bug.
- The delayed post-publish inventory is a separate MDK local-first read issue.

agent-p1p:
Fixes #1155

## Summary
- preserve privacy-safe machine-readable relay rejection categories per endpoint without retaining relay-controlled suffix text
- collapse duplicate endpoint failures into one concise Display/UI summary while retaining structured diagnostics
- keep mixed accepted/rejected publishes successful and gate KeyPackage metadata removal on at least one deletion acknowledgement
- add real SDK/MockRelay coverage for valid NIP-09 KeyPackage deletion and cache cleanup

## Root cause
The KeyPackage deletion shape (`kind:5` with `e` and `k=30443`) is accepted by compatible relays. The failure was diagnostic: `send_event_to_relay` replaced every `OK:false` message with one generic string, then `finish_publish_outcome` duplicated that string and discarded endpoint failures on the all-rejected path.

## Verification
- `cargo test -p cgka-traits --locked`
- `cargo test -p marmot-app --lib key_package_deletion --locked`
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked -- --test-threads=1` — 97 passed
- `just fast-ci`
- CI-exact default/all-feature workspace `cargo check` and `cargo clippy` commands — passed

## NIP-42
The production deletion client continues to use `NostrSdkClient::builder().signer(signer).build()`. No auth options are overridden, so nostr-sdk automatic authentication remains enabled by default. The local relay harness does not model the complete NIP-42 challenge flow; no fake auth path was added.

## Sensitive paths
- `crates/transport-nostr-adapter/src/sdk_client.rs` — signed event publication and relay rejection handling
- `crates/marmot-app/src/runtime/mod.rs` — KeyPackage deletion failure projection during sign-out/wipe

No cryptographic primitives, MLS/CGKA state, key material, NIP-09 event shape, versions, conformance fixtures, or `Cargo.lock` were changed.

## Summary by CodeRabbit

- **New Features**
  - Added privacy-safe categories for relay publication failures, making errors more actionable without exposing relay URLs or remote rejection text.
  - Key-package deletion now removes matching local publication metadata after successful relay acknowledgement.
  - Added support for structured failure details and clearer aggregated publish errors.

- **Bug Fixes**
  - Improved handling of relay failures during key-package deletion.
  - Preserved compatibility with older failure records that lack rejection categories.

- **Documentation**
  - Added unreleased changelog notes covering deletion behavior and privacy-safe relay diagnostics.

### DIFF
--- a/crates/cgka-session/tests/support/nostr_stack.rs
+++ b/crates/cgka-session/tests/support/nostr_stack.rs
@@ -128,6 +128,7 @@ impl NostrRelayClient for FakeRelayClient {
                 .map(|endpoint| TransportEndpointFailure {
                     endpoint,
                     reason: "not accepted by in-memory relay".into(),
+                    rejection_category: None,
                 })
                 .collect(),
         })
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -9,6 +9,14 @@ versioning through the workspace version in the root `Cargo.toml`.
 
 ## [Unreleased]
 
+### Fixed
+
+- KeyPackage NIP-09 deletions now surface privacy-safe per-relay rejection
+  categories (for example `blocked` or `auth-required`) instead of repeating a
+  generic `relay rejected event` summary, while successful deletions still
+  clear only matching local KeyPackage publication metadata after at least one
+  relay acknowledgement.
+
 ## [0.9.11] - 2026-08-09
 
 ### Added
--- a/crates/marmot-account/src/runtime.rs
+++ b/crates/marmot-account/src/runtime.rs
@@ -3253,6 +3253,7 @@ fn frozen_fanout_report(fanout: &OutboundFanout) -> TransportPublishReport {
                 failed.push(TransportEndpointFailure {
                     endpoint: endpoint.clone(),
                     reason: "publish attempt failed".into(),
+                    rejection_category: None,
                 });
             }
             cgka_traits::FanoutTargetStatus::NotAttempted
--- a/crates/marmot-app/src/runtime/mod.rs
+++ b/crates/marmot-app/src/runtime/mod.rs
@@ -11,7 +11,7 @@ use cgka_traits::agent_text_stream::AGENT_TEXT_STREAM_EXPORTER_CACHE_KEY;
 use cgka_traits::app_event::MarmotAppEvent as MarmotInnerEvent;
 use cgka_traits::engine::GroupEvent;
 use cgka_traits::storage::{KeyPackageBundleStorage, MaintenanceStorage};
-use cgka_traits::{GroupId, SecretBytes, TransportEndpoint};
+use cgka_traits::{GroupId, SecretBytes, TransportAdapterError, TransportEndpoint};
 use marmot_account::{
     AccountHome, AccountHomeError, AccountSetupKind, AccountSetupPhase, AccountSummary,
     NostrAccountImport,
@@ -716,6 +716,9 @@ pub struct SignOutOutcome {
 fn wipe_failure_reason(err: &AppError) -> String {
     let category = match err {
         AppError::RuntimeStopping => "runtime is shutting down",
+        AppError::Transport(TransportAdapterError::PublishEndpoints(failure)) => {
+            return failure.summary.clone();
+        }
         AppError::Transport(_) | AppError::TransportClosed => "transport error",
         AppError::Publish(_) => "relay publish failed",
         AppError::Storage(_) | AppError::Sqlite(_) | AppError::SqlcipherKeyDerivation(_) => {
@@ -2369,10 +2372,25 @@ impl MarmotAppRuntime {
                     event_id_hex: result.event_id_hex,
                     reason: "relay publish failed".to_owned(),
                 }),
-                Err(error) => failures.push(RelayFailure {
-                    event_id_hex: result.event_id_hex,
-                    reason: wipe_failure_reason(&error),
-                }),
+                Err(error) => {
+                    let endpoint_failures = match &error {
+                        AppError::Transport(error) => error.publish_endpoint_failures(),
+                        _ => &[],
+                    };
+                    if endpoint_failures.is_empty() {
+                        failures.push(RelayFailure {
+                            event_id_hex: result.event_id_hex,
+                            reason: wipe_failure_reason(&error),
+                        });
+                    } else {
+                        for endpoint_failure in endpoint_failures {
+                            failures.push(RelayFailure {
+                                event_id_hex: result.event_id_hex.clone(),
+                                reason: endpoint_failure.reason.clone(),
+                            });
+                        }
+                    }
+                }
             }
         }
         (deleted, failures)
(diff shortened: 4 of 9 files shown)


agent-p1p:

Reviewed exact head `826d25d8026a03bc246745eb9348da8d97389fc4`.

### Blocking

1. **`crates/marmot-app/src/runtime/mod.rs:719-720, 2375-2391` — the app/FFI privacy boundary now trusts free-form transport strings and re-expands duplicate UI failures.**

   `wipe_failure_reason` explicitly promises not to interpolate wrapped transport detail because it may contain relay URLs or attacker-controlled text, but the new arm returns public `TransportPublishFailure.summary` verbatim. `delete_relay_key_packages` then bypasses that classifier and copies every public `TransportEndpointFailure.reason` into `RelayFailure`, which is serialized to the app/UniFFI partial-failure sheet. Those fields are ordinary public `String`s; the type system does not enforce the SDK adapter's sanitizing convention. A transport error containing `wss://private-relay.example` or a relay suffix can therefore cross the boundary this function was designed to scrub. With two identical rejects, the loop also emits two identical app-facing rows, undoing the issue's concise-summary requirement without preserving an endpoint identity in `RelayFailure`.

   Keep endpoint records on the structured diagnostic error, but derive the app-facing summary only from typed `rejection_category` values and fixed local fallback categories, dedupe it once per event, and never copy `summary`/`reason` blindly. Add a regression test that injects hostile strings into both fields and proves neither the strings nor duplicate rows reach `SignOutOutcome`/`WipeOutcome`.

2. **`crates/transport-nostr-adapter/src/sdk_client.rs` / `crates/marmot-app/tests/relay_runtime.rs` — acceptance criterion 2 (supported NIP-42 auth policy) has no regression test.**

   The PR relies on `Client::builder().signer(...).build()` plus nostr-sdk's default automatic authentication and the wrapper's retry timing, but only tests an accepting relay and permanently rejecting relays. This is the auth-sensitive path the issue explicitly requires. `nostr-relay-builder` already supports a real write-auth relay through `RelayBuilder::nip42(RelayBuilderNip42 { mode: RelayBuilderNip42Mode::Write })`; exercise `NostrSdkRelayClient` with a signer and a kind-5 deletion against that relay, and assert the challenge/auth/retry path receives an acknowledgement. A static default is not a regression guard for the composed behavior.

### Suggestion

- **`crates/transport-nostr-adapter/src/sdk_client.rs:603-620, 1315-1336` — test mixed publish through the real fan-out loop.** The new mixed accepted/rejected test calls `finish_publish_outcome` directly. Production stops at the first required ack and drains later `JoinSet` results without recording them, so that unit test cannot prove the issue's mixed-relay diagnostic contract. Add a deterministic `publish_event` test (or clarify the contract as “only failures observed before quorum”) so the test matches actual scheduling behavior.

### Verified locally

- `cargo test -p cgka-traits --locked` — passed (96 unit, 4 error-display, 6 host-safety, 23 snapshot tests)
- `cargo test -p transport-nostr-adapter --features sdk --locked` — passed (59 unit, 30 integration)
- `cargo test -p marmot-app --lib key_package_deletion --locked` — passed (3 focused tests)
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked app_runtime_delete_key_package_event_nip09_succeeds_and_clears_matching_cache -- --exact --test-threads=1` — passed
- `cargo clippy -p cgka-traits -p transport-nostr-adapter -p marmot-app --all-targets --all-features --locked -- -D warnings` — passed

Changelog is present under `Unreleased`; no version, fixture-cohort, or lockfile bump was introduced.

agent-p1p:
Addressed the blocking review findings in 96170bf8.

- App-facing deletion failures now ignore transport-controlled summary/reason/endpoint fields, derive fixed text only from typed rejection categories, and emit one failure row per deletion event. Added hostile-string and duplicate-endpoint regression coverage.
- Added a real `NostrSdkRelayClient` integration test against a write-authenticated `LocalRelay` proving kind-5 KeyPackage deletion succeeds through NIP-42.
- Locked the quorum contract in tests: failures observed before quorum are retained; silent/hung fan-out tasks aborted after quorum do not create diagnostics.

Verification: `just fast-ci`; full `transport-nostr-adapter --features sdk` suite (90 tests); full serial `marmot-app` relay runtime suite (97 tests); focused app/traits tests; independent Composer review found no blockers.

erskingardner (dismissed):
Reviewed exact head `96170bf839f5234704a024b219e97fc5ab2e3049`.

The current changes satisfy #1155: relay-controlled suffix text is discarded behind typed rejection categories, app/FFI deletion summaries are reconstructed from fixed local text and deduplicated per event, one acknowledgement remains sufficient without premature metadata removal, and the SDK path now has real accepting, rejecting, and NIP-42 write-auth relay coverage. I found no actionable code issues.

All current GitHub checks are complete and passing. GitHub currently reports the PR as conflicting with `master`; resolve that conflict and revalidate the resulting head before merge.

PR #1364, opened 2026-08-10
Turn refused harness leaves into failed scenario steps

dannym-arx:
A conformance vector that scripts a `leave` the engine refuses used to kill the whole report run. `HarnessClient::leave_capture` called `.expect("send leave")`, so an engine error such as `AdminCannotSelfRemove` panicked the process, aborted every remaining vector, and wrote no failure capsule. A marmot turned away at the burrow door should mark its map and keep walking, not collapse the whole colony.

Found while probing new scenarios: a sole-admin self-leave trial hit the intended `AdminCannotSelfRemove` policy and took down the runner instead of failing one step.

Changes:

- `HarnessClient::leave` and `leave_capture` return `Result`. The missing default group, an engine refusal, and a non-proposal send result all become typed errors.
- The engine subject maps the error through the existing `subject_engine_error` classifier, so a refused leave reports `scenario_step_failed:admin_policy` with a failure capsule while sibling vectors keep running.
- Three test call sites now `.expect(...)` because their scenarios require the leave to succeed.

Verification: a report run with the refused-leave trial vector plus two passing leave vectors gives 2 pass, 1 clean FAIL with capsule, exit 1, no panic. The leave-touching test suites (`canonical_scenarios`, `openmls_replay_probe`, `report_runner`: 45 + 14 + 23 tests) pass, and `just fast-ci` is green.

`send_app` and a few sibling harness methods keep the same `.expect` pattern and would panic the same way on an engine refusal. The same fix shape applies there if we want a follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of leave failures by reporting errors instead of silently ignoring them or terminating unexpectedly.
  * Leave operations now clearly indicate whether they succeeded and provide the resulting proposal when applicable.
  * Failed self-removal attempts now release reserved scenario inputs and allow scenario execution to continue correctly.

* **Tests**
  * Strengthened removal, replay, and conformance scenarios to fail explicitly when leave operations do not produce the expected result.

### DIFF
--- a/crates/cgka-conformance-simulator/src/client.rs
+++ b/crates/cgka-conformance-simulator/src/client.rs
@@ -1361,20 +1361,22 @@ impl HarnessClient {
     }
 
     /// Send a SelfRemove proposal (Leave intent).
-    pub async fn leave(&mut self) {
-        self.leave_capture().await;
+    pub async fn leave(&mut self) -> Result<(), EngineError> {
+        self.leave_capture().await.map(|_| ())
     }
 
     /// Send a SelfRemove proposal and return the wrapped transport message.
-    pub async fn leave_capture(&mut self) -> TransportMessage {
-        let gid = self.default_group.clone().expect("group");
+    pub async fn leave_capture(&mut self) -> Result<TransportMessage, EngineError> {
+        let gid = self
+            .default_group
+            .clone()
+            .ok_or_else(|| EngineError::Other("leave without a default group".into()))?;
         let res = self
             .engine_mut()
             .send(SendIntent::Leave {
                 group_id: gid.clone(),
             })
-            .await
-            .expect("send leave");
+            .await?;
         if let SendResult::Proposal { msg } = res {
             let scenario_input =
                 self.next_scenario_input_metadata(ScenarioInputKind::Proposal, None, None);
@@ -1386,9 +1388,11 @@ impl HarnessClient {
             self.register_published_scenario_input(&routed, scenario_input)
                 .await;
             self.bus.send(self.bus_id, routed.clone());
-            routed
+            Ok(routed)
         } else {
-            panic!("expected Proposal");
+            Err(EngineError::Other(format!(
+                "expected Proposal, got {res:?}"
+            )))
         }
     }
 
--- a/crates/cgka-conformance-simulator/src/client.rs
+++ b/crates/cgka-conformance-simulator/src/client.rs
@@ -503,6 +503,21 @@ fn key_package_with_harness_source(key_package: KeyPackage) -> KeyPackage {
     .with_protocol_profile(protocol_profile)
 }
 
+/// Stable variant name for unexpected-result errors. Debug-formatting a
+/// [`SendResult`] would embed transport payload bytes in the error string.
+fn send_result_kind(res: &SendResult) -> &'static str {
+    match res {
+        SendResult::NoChange { .. } => "NoChange",
+        SendResult::DisbandRequested { .. } => "DisbandRequested",
+        SendResult::ApplicationMessage { .. } => "ApplicationMessage",
+        SendResult::Queued { .. } => "Queued",
+        SendResult::Proposal { .. } => "Proposal",
+        SendResult::GroupEvolution { .. } => "GroupEvolution",
+        SendResult::GroupCreated { .. } => "GroupCreated",
+        SendResult::FoundingGroupCreated { .. } => "FoundingGroupCreated",
+    }
+}
+
 #[derive(Clone)]
 struct NostrAccountIdentityProofSigner {
     keys: nostr::Keys,
@@ -1366,11 +1381,24 @@ impl HarnessClient {
     }
 
     /// Send a SelfRemove proposal and return the wrapped transport message.
+    ///
+    /// A refused leave releases any id reserved via
+    /// [`Self::name_next_scenario_input`]; only a sent proposal consumes it,
+    /// so the next named action neither trips the unconsumed-id assertion nor
+    /// inherits the failed leave's id.
     pub async fn leave_capture(&mut self) -> Result<TransportMessage, EngineError> {
+        let result = self.leave_capture_inner().await;
+        if result.is_err() {
+            self.next_scenario_input_id = None;
+        }
+        result
+    }
+
+    async fn leave_capture_inner(&mut self) -> Result<TransportMessage, EngineError> {
         let gid = self
             .default_group
             .clone()
-            .ok_or_else(|| EngineError::Other("leave without a default group".into()))?;
+            .ok_or_else(|| EngineError::Other("must create or join a group first".into()))?;
         let res = self
             .engine_mut()
             .send(SendIntent::Leave {
@@ -1391,7 +1419,8 @@ impl HarnessClient {
             Ok(routed)
         } else {
             Err(EngineError::Other(format!(
-                "expected Proposal, got {res:?}"
+                "expected Proposal, got {}",
+                send_result_kind(&res)
             )))
         }
     }
--- a/crates/cgka-conformance-simulator/src/subject.rs
+++ b/crates/cgka-conformance-simulator/src/subject.rs
@@ -1400,7 +1400,7 @@ impl ConvergenceSubject for EngineHarnessSubject {
     async fn leave(&mut self, action_id: &str, client: &str) -> Result<(), SubjectError> {
         let client = self.client_mut(client)?;
         client.name_next_scenario_input(action_id);
-        client.leave().await;
+        client.leave().await.map_err(subject_engine_error)?;
         Ok(())
     }
 
--- a/crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
+++ b/crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
@@ -2448,7 +2448,7 @@ async fn add_then_self_remove_via_harness() {
     carol.tick().await;
 
     // Bob (non-admin) leaves.
-    bob.leave().await;
+    bob.leave().await.expect("leave");
     bus.deliver_all();
     let alice_proposal_outcomes = alice.tick().await; // ingests proposal + auto-commits
     let carol_proposal_outcomes = carol.tick().await; // same: no deterministic election
--- a/crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
+++ b/crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
@@ -2422,6 +2422,42 @@ async fn three_client_message_exchange_trace() -> ScenarioTrace {
 }
 
 #[tokio::test]
+async fn refused_leave_releases_named_scenario_input_reservation() {
+    // A refused leave must release the id reserved by
+    // name_next_scenario_input; otherwise the next named action panics on the
+    // unconsumed-id assertion or inherits the failed leave's id.
+    let bus = TransportBus::ordered();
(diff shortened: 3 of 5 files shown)


erskingardner (changes requested):
The panic-to-error conversion is the right direction, but I think this needs changes before merge.

I independently verified both unresolved inline findings at this head:

- A failed `leave_capture` does not consume or clear `next_scenario_input_id`, so the newly recoverable `Result` leaves the harness in a poisoned state: the next named input panics (or an unnamed input inherits the rejected leave action ID). Please compensate the reservation on every error path.
- The new unexpected-result branch formats the full `SendResult`; variants contain `TransportMessage` payload and routing fields, and `subject_engine_error` carries that text into report/failure artifacts. Please report only a stable variant kind.

There is also no checked-in regression for the behavior this PR exists to guarantee. The changed tests only adapt successful call sites to the new return type. Please add a report-runner test with a refused admin self-leave plus a passing sibling fixture that asserts: no process panic, the refused step is classified as `admin_policy`, a failure capsule is written, the sibling still runs, and the command exits non-zero. That also makes the claimed verification reproducible in CI.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -1391,7 +1391,7 @@ impl ConvergenceSubject for EngineHarnessSubject {
     async fn leave(&mut self, action_id: &str, client: &str) -> Result<(), SubjectError> {
         let client = self.client_mut(client)?;
         client.name_next_scenario_input(action_id);
-        client.leave().await;
+        client.leave().await.map_err(subject_engine_error)?;
```
erskingardner, line 1403: The behavior this PR exists for isn't pinned by a committed test. The PR body describes exactly the right verification ("a report run with the refused-leave trial vector plus two passing leave vectors gives 2 pass, 1 clean FAIL with capsule, exit 1, no panic") but it was run by hand, so nothing stops the next person from reintroducing `.expect("send leave")` here or in `leave_capture`. The three updated call sites all `.expect(...)` on success, so they'd stay green.

The crate already has an exact-fit, cheap pattern: `tests/scenario_ir.rs::typoed_group_label_fails_closed_at_execution` builds a `ScenarioSpec`, calls `run_scenario_report`, and asserts `step_log` contains `ScenarioStepStatus::Failed { kind, .. } if kind == "unknown_scenario_group"`. The repro here is two steps — `CreateGroup` with `initial_admins: ["alice"]`, then `Leave { client: "alice" }` — because `prepare_self_remove_proposal` returns `AdminCannotSelfRemove` for any member still in the admin set (`crates/cgka-engine/src/message_processor/send.rs:734`). Assert `Failed { kind == "admin_policy" }` and that `run_scenario_report` returns `Ok` rather than unwinding. That's ~30 lines and it's the whole contract this PR is buying.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1361,20 +1361,22 @@ impl HarnessClient {
...
     }
 
     /// Send a SelfRemove proposal and return the wrapped transport message.
-    pub async fn leave_capture(&mut self) -> TransportMessage {
-        let gid = self.default_group.clone().expect("group");
+    pub async fn leave_capture(&mut self) -> Result<TransportMessage, EngineError> {
+        let gid = self
+            .default_group
+            .clone()
+            .ok_or_else(|| EngineError::Other("leave without a default group".into()))?;
```
erskingardner: Nit, non-blocking, and pre-existing convention rather than something this PR introduces: `EngineError::Other` classifies as `SubjectFailureCategory::Protocol` (`subject.rs:2055`), which `fingerprint_report_failure` turns into `TerminalOutcomeClassification::TerminalProtocolFailure` (`failure_capsule.rs:475`). So both of the harness-side conditions added here — a scripted leave on a client with no default group, and the non-`Proposal` branch below — would be capsuled as the engine committing a terminal protocol violation, when they're really scenario-authoring / harness faults. The runner's own equivalent is `scenario.rs::err`, which uses `SubjectFailureCategory::Environment`.

`try_send_app` (line 1270) does the identical `ok_or_else(|| EngineError::Other(...))`, so matching it is defensible and I'd not block on it. If you do keep it, consider reusing that function's exact string (`"must create or join a group first"`) so the two harness preconditions stay greppable as one class.

erskingardner (commented):
Good change, right size. A refused engine operation inside a harness is data, not a reason to take down the run, and turning three `.expect`s into typed errors that flow through the existing `subject_engine_error` classifier is the minimal way to say that. Net +16/-12 with no new abstraction, the new code mirrors `try_send_app`/`try_invite` two functions up almost line for line, the scope note about `send_app` and siblings being left for a follow-up is the correct call, and all 17 checks are green.

**One ask before merge:** there's no committed regression test. The manual verification in the PR body is exactly the right test, and `tests/scenario_ir.rs::typoed_group_label_fails_closed_at_execution` is a ~30-line template for it. Details inline. Not blocking correctness — nothing here is wrong — but without it the next `.expect` reintroduces the panic silently, since all three updated call sites `.expect(...)` on the success path.

One nit inline about `EngineError::Other` mapping harness preconditions onto `TerminalProtocolFailure` in the capsule. Pre-existing convention; take it or leave it.

**On the two CodeRabbit findings — I believe both are false positives, so don't spend time on them:**

1. *Reserved `next_scenario_input_id` leaking into later actions.* It can't. `run_scenario_report`'s step loop `break`s on the first failed step (`scenario.rs:1274-1286`), so no later action on that client reaches `name_next_scenario_input`, and the subject is per-run. The reservation dies with the run.
2. *`{res:?}` leaking transport contents.* `do_send_leave` (`message_processor/send.rs:649-698`) only ever returns `Ok(SendResult::Proposal)` or `Err`, so the branch is unreachable defensive code and can't format a payload-bearing variant. It also matches four adjacent precedents in this same file (lines 1189, 1260, 1305, 1357).

**Forward-looking, not for this PR.** The motivating case — a sole-admin self-leave that *should* be refused — currently reports as a scenario FAIL. If the intent is to script that refusal as a passing vector, the established idiom is a dedicated step: see `ScenarioStep::ExpectUpdateAdminPolicyError` (`scenario.rs:861`). Worth noting that step deliberately passes `action_id: None` so no scenario input id is reserved for an operation expected to fail. An `ExpectLeaveError` would need the same treatment — that's the one shape where CodeRabbit's dangling-reservation concern would become real, because such a step continues the run instead of breaking out of it.

crates/cgka-conformance-simulator/src/subject.rs
Datawav, line 1403: **Medium:** If `leave()` fails before `next_scenario_input_metadata` consumes the ID—such as for a missing default group or a send error—the ID set by `name_next_scenario_input` remains pending. The next named action then triggers the “previous scenario input id was not consumed” assertion and aborts the report after an otherwise recoverable failure. Clear the pending ID on every failed leave, or scope naming transactionally to the attempted send. Please add a regression for a refused leave followed by another named action, verifying that execution continues with the correct action ID.

Datawav (commented):
## Automated review

Changes required: one scenario-runner state leak remains.

- **Medium — A refused leave leaves its action ID pending.** In `subject.rs:1393–1394`, `name_next_scenario_input(action_id)` runs before the newly fallible `client.leave()`. In `client.rs:1366–1378`, missing-group and send failures return via `?` before `next_scenario_input_metadata` can consume that ID. Consequently, the next named scenario action hits `name_next_scenario_input`’s “previous scenario input id was not consumed” assertion, again aborting the report after the intended recoverable failure. Clear the pending ID on every failed leave, or scope naming transactionally to the attempted send. Add a regression covering refused leave followed by another named action and verify the run continues with the correct action ID.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1361,20 +1361,22 @@ impl HarnessClient {
     }
 
     /// Send a SelfRemove proposal (Leave intent).
-    pub async fn leave(&mut self) {
-        self.leave_capture().await;
+    pub async fn leave(&mut self) -> Result<(), EngineError> {
+        self.leave_capture().await.map(|_| ())
```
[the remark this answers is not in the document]
dannym-arx: Fixed in b105b43e. `leave_capture` now wraps an inner helper and clears `next_scenario_input_id` on every error path (missing default group, engine refusal, unexpected result). Added `refused_leave_releases_named_scenario_input_reservation` in `tests/canonical_scenarios.rs`, which names an action, gets the leave refused, then names and sends further actions — it panics on the unconsumed-id assertion without the fix and passes with it.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1386,9 +1388,11 @@ impl HarnessClient {
             self.register_published_scenario_input(&routed, scenario_input)
                 .await;
             self.bus.send(self.bus_id, routed.clone());
-            routed
+            Ok(routed)
         } else {
-            panic!("expected Proposal");
+            Err(EngineError::Other(format!(
+                "expected Proposal, got {res:?}"
+            )))
```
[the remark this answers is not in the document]
dannym-arx: Fixed in b105b43e. The unexpected-result error now reports a stable variant name via a `send_result_kind` helper instead of Debug-formatting the whole `SendResult`. (The pre-existing `{other:?}` sites in `create_group`/`disband` have the same shape but predate this PR — left untouched to keep the diff minimal.)

crates/cgka-conformance-simulator/src/subject.rs
dannym-arx: Added in b105b43e as `refused_admin_leave_fails_the_step_without_unwinding` in `tests/scenario_ir.rs`, following the `typoed_group_label_fails_closed_at_execution` pattern exactly as suggested: sole-admin `CreateGroup` + `Leave`, asserting `run_scenario_report` returns `Ok` and the step log carries `Failed { kind: "admin_policy" }`.

crates/cgka-conformance-simulator/src/client.rs
dannym-arx: Kept `EngineError::Other` to match `try_send_app`, and aligned the string to its exact "must create or join a group first" in b105b43e so the two harness preconditions stay greppable as one class. Agreed the Protocol-vs-Environment categorization of harness faults is a pre-existing convention worth revisiting separately.

crates/cgka-conformance-simulator/src/subject.rs
dannym-arx: Fixed in b105b43e — the reservation is cleared on every failed leave path. One nuance on "verifying that execution continues": the scenario runner intentionally breaks the step loop on the first failed step, so within a single scenario no later named action runs. The regression therefore lives at the harness-client level (`refused_leave_releases_named_scenario_input_reservation`): refused named leave, then two more named actions that name and consume fresh ids without tripping the assertion. Verified it fails without the fix.

Datawav (commented):
## Automated review

- Fixed — refused leaves no longer strand or leak the named action ID: `client.rs:1379–1387` clears the reservation on every error, while `canonical_scenarios.rs:2444–2453` verifies a refused leave is followed by fresh named actions that are consumed without assertion failure or ID inheritance.
- Fixed — engine refusal now becomes a structured failed scenario step: `subject.rs:1400–1404` propagates the leave error, and `scenario_ir.rs:286–321` verifies `admin_policy` failure without unwinding.

erskingardner (approved):

PR #1365, opened 2026-08-11
fix(marmot-app): run armed backfill after catch-up

### LINKED ISSUE #1180: marmot-app: explicit CatchUp can arm epoch-gap recovery without running it
## Summary

A steady-state `AccountWorkerCommand::CatchUp` can detect an epoch stall and set `epoch_backfill_pending`, but it intentionally does not execute the pending replay. It reports success and relies on a future inbound receive pass.

If the arming burst is the group's last traffic, the below-floor recovery replay never starts. Repeating catch-up with no new delivery does not fix it.

## Root cause

`crates/marmot-app/src/runtime/account_worker.rs:954-970`:

```rust
let result = match client.sync().await {
    Ok(summary) => {
        // Catch-up arms without running the backfill (the next
        // receive pass runs it) ...
        if client.has_pending_epoch_backfill() {
            shared.schedule_audit_log_tracker_update("epoch_backfill_armed");
        }
        ...
        Ok(())
    }
```

Startup and normal receive seams call `run_pending_epoch_backfill_reporting_arm` (`account_worker.rs:537-548`, `:736-751`). The explicit catch-up seam does not. The key-package maintenance sync at `account_worker.rs:857-877` likewise publishes its summary without running newly armed backfill work.

The only steady-state receive path executes pending work after another delivery has first been received and ingested. That dependency on unrelated later traffic is the same liveness shape the `next_event` seam in PR #892 was changed to avoid.

## Failure sequence

1. A device has missed an MLS commit and remains at an older epoch.
2. `runtime.catch_up_accounts()` drains eight distinct later undecryptable events and arms epoch-gap recovery.
3. `client.sync()` returns an empty summary with `epoch_backfill_pending = true`.
4. The command uploads arm evidence and reports success without invoking the replay.
5. No later delivery arrives, so the receive seam never runs the pending full-history replay.
6. The device remains stalled despite a successful catch-up.

## Impact

- A quiet group can remain behind after explicit foreground/connectivity/push-wake catch-up.
- User-visible catch-up reports success without executing recovery work it armed.
- Recovery depends on unrelated future traffic.
- This is isolated-device availability/liveness loss, not a group-wide consensus deadlock.

## Required behavior

- When successful `AccountWorkerCommand::CatchUp` leaves `client.has_pending_epoch_backfill()` true, invoke `run_pending_epoch_backfill_reporting_arm` (or a shared equivalent) before replying success.
- Apply the same post-sync rule to maintenance and every other sync seam that can arm recovery.
- If `activate_transport(None)` fails, retain pending intent and surface the account error; do not require another threshold crossing.
- Preserve one-replay coalescing, transport ordering, and the existing anti-abuse threshold.
- Preserve current audit-tracker behavior for arm-only and replay-failure cases.

## Regression coverage

1. Establish a group and durable transport cursor.
2. Publish one below-floor recovery target plus eight above-floor arming probes while the account is offline.
3. Start the worker without relying on startup's built-in recovery seam and invoke `runtime.catch_up_accounts()`.
4. Publish no later delivery.
5. Assert the below-floor target is replay-delivered within a deadline.
6. Assert no second full-history replay occurs after the same stalled epoch's evidence is consumed.
7. Add the equivalent maintenance-sync regression or centralize the helper so both paths share the tested implementation.

## Validation and duplicate check

Focused tests on `b609ccd` passed:

- `cold_restart_since_floor_permanently_drops_backlog_below_it`;
- `stalled_epoch_backfill_recovers_below_floor_backlog_when_armed`;
- the live `next_event` backfill seam.

Those tests validate the detector and startup/live paths; none covers explicit `CatchUp` with no later inbound event.

Repository-wide issue/PR searches found no duplicate. PR #892 fixed startup/live recovery; PR #919 records arm evidence but explicitly leaves CatchUp execution deferred. #948 concerns detector-memory cleanup, not this liveness seam.

Separate from #1176, which concerns foreground deferred-peel budgeting.

agent-p1p:
## Summary

- execute an armed epoch-gap full-history replay immediately after successful runtime catch-up, maintenance, convergence, startup, and receive sync seams
- return replay activation failures to explicit catch-up callers while retaining the coalesced replay intent for retry
- let explicit full-history repair satisfy an armed replay without issuing the same account-wide query twice
- add regression coverage for catch-up completion, failure retention/error reporting, and one-replay coalescing

## Test plan

- `CARGO_INCREMENTAL=0 just fast-ci`
- focused account-worker regression tests (3 passed)
- `cargo test -p marmot-app --test since_floor -- --nocapture`
- `cargo test -p marmot-app --test next_event_backfill -- --nocapture`
- `cargo test -p marmot-app --test epoch_stall_backfill_audit -- --nocapture`

Full `cargo test -p marmot-app` was also run locally. It reached `relay_runtime` with six failures outside this diff. Five reproduce individually on unmodified `origin/master`; the sixth passes in isolation and failed only in the parallel suite.

Fixes #1180

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incremental synchronization when an epoch gap requires full-history replay.
  * Prevented duplicate account-wide queries during explicit repair.
  * Improved handling and reporting of replay failures across startup, scheduled, receive, maintenance, and catch-up flows.
  * Ensured pending repairs are cleared after successful replay and retained when replay fails.
* **Documentation**
  * Added an Unreleased changelog entry describing the synchronization fixes.

### DIFF
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -9,6 +9,13 @@ versioning through the workspace version in the root `Cargo.toml`.
 
 ## [Unreleased]
 
+### Fixed
+
+- Runtime catch-up, key-package maintenance, and every other incremental sync
+  seam now execute an armed epoch-gap full-history replay immediately instead of
+  waiting for unrelated later relay traffic. Explicit full-history repair also
+  consumes a pending replay without issuing the same account-wide query twice.
+
 ## [0.9.11] - 2026-08-09
 
 ### Added
--- a/crates/marmot-app/src/client/sync.rs
+++ b/crates/marmot-app/src/client/sync.rs
@@ -760,6 +760,18 @@ impl AppClient {
         self.epoch_backfill_pending
     }
 
+    /// Consume a pending epoch-gap replay after an unfloored transport
+    /// activation has succeeded. Shared by the automatic replay path and the
+    /// explicit full-history repair path, which already performed the same
+    /// account-wide replay and must not issue it twice.
+    pub(crate) fn finish_pending_epoch_backfill_after_replay(&mut self) {
+        if !self.epoch_backfill_pending {
+            return;
+        }
+        self.epoch_stall.mark_replayed();
+        self.epoch_backfill_pending = false;
+    }
+
     /// Recover any group that stalled below its live epoch during ingest by
     /// replaying the account's full transport history (`since = None`). One replay
     /// re-fetches every group, so the detector collapses simultaneously-stuck
@@ -769,8 +781,7 @@ impl AppClient {
             return Ok(());
         }
         self.runtime.activate_transport(None).await?;
-        self.epoch_stall.mark_replayed();
-        self.epoch_backfill_pending = false;
+        self.finish_pending_epoch_backfill_after_replay();
         Ok(())
     }
 
--- a/crates/marmot-app/src/runtime/account_worker.rs
+++ b/crates/marmot-app/src/runtime/account_worker.rs
@@ -666,7 +666,7 @@ async fn run_app_runtime_account_worker(
         Ok(summary) => {
             publish_app_runtime_summary(&events, &account_id_hex, &account_label, &summary);
             schedule_pending_convergence_groups(&mut scheduled_convergence, &mut client);
-            run_pending_epoch_backfill_reporting_arm(
+            let backfill_result = run_pending_epoch_backfill_reporting_arm(
                 &mut client,
                 &events,
                 &account_id_hex,
@@ -677,7 +677,7 @@ async fn run_app_runtime_account_worker(
             if sync_summary_triggers_audit_tracker_update(&summary) {
                 shared.schedule_audit_log_tracker_update("startup_sync");
             }
-            Ok(())
+            backfill_result
         }
         Err(err) => {
             // A failed initial catch-up surfaces as an account error but must not
@@ -817,7 +817,7 @@ async fn run_app_runtime_account_worker(
                                         &mut scheduled_convergence,
                                         &mut client,
                                     );
-                                    run_pending_epoch_backfill_reporting_arm(
+                                    let _ = run_pending_epoch_backfill_reporting_arm(
                                         &mut client,
                                         &events,
                                         &account_id_hex,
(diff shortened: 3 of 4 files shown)


agent-p1p:
## Pip adversarial review — APPROVE

Reviewed exact head `efc851a0b7dd473a40ae706b63153066b48ad215`.

BLOCKING:
- none

SUGGESTION:
- **Add the exact explicit-catch-up delivery regression when practical**: `crates/marmot-app/src/runtime/account_worker.rs:3200` manually pre-arms the detector and verifies a second subscription plus pending-intent consumption. The existing `since_floor` integration test proves the same shared helper delivers below-floor history, so this is non-blocking; however, a runtime-level test that arms during `catch_up_accounts()`, publishes no later traffic, observes the target delivery, and checks no second replay would directly pin issue #1180's reported sequence instead of relying on composition.

NITPICK:
- none

SUMMARY: The catch-up, maintenance, convergence, startup, and receive seams now share the same capture-before-replay helper. A failed unfloored activation publishes a privacy-safe account error and leaves the replay intent armed; successful explicit full-history repair consumes that intent without a duplicate account-wide replay. Audit scheduling, detector debounce/coalescing, and transport ordering are preserved. The Unreleased changelog is present, with no version, lockfile, release-tag, or fixture churn.

VERDICT: APPROVE

SENSITIVE_PATHS: `crates/marmot-app/src/client/sync.rs`, `crates/marmot-app/src/runtime/account_worker.rs` (account/group-state recovery orchestration)

Verification at the reviewed head:
- `cargo test -p marmot-app --lib runtime::account_worker::tests:: -- --nocapture` — 19 passed
- `cargo test -p marmot-app --test since_floor stalled_epoch_backfill_recovers_below_floor_backlog_when_armed -- --nocapture` — passed
- `cargo test -p marmot-app --test next_event_backfill -- --nocapture` — passed
- `cargo fmt --all -- --check` — passed
- `cargo clippy -p marmot-app --all-targets -- -D warnings` — passed
- Live GitHub checks are green; CodeRabbit reported no actionable comments.

agent-p1p:
Addressed the review suggestion in 1937fd38. The new runtime-level regression arms recovery during catch_up_accounts(), parks the unfloored replay, injects a below-floor relay response, verifies ingestion without later traffic, and asserts the consumed evidence cannot trigger a second replay. It also fails at the replay deadline when the catch-up helper call is removed.

agent-p1p:
@coderabbitai review

PR #1366, opened 2026-08-11
fix(marmot-app): preserve local group deletions

### LINKED ISSUE #1238: Locally deleted groups are re-created by projection reconciliation after an app update
## Impact

A user deletes a conversation locally. After an app update — with no new message in that group — the conversation reappears in the chat list. There is currently no way for a client to express "this group was deliberately removed locally" and have that survive reconciliation.

Reported downstream as whitenoise-android#1646. That issue's own routing section asks for this dependency to be filed and cross-linked; this is that issue.

## Cause

Two behaviours combine.

`crates/storage-sqlite/src/account_projection.rs::delete_local_group_data` (line 734, doc at 726) deliberately leaves the protocol tables intact. Its documentation states they are kept for active groups so old relay deliveries stay suppressed "while a future fresh group message can re-create the app projection". So a local delete removes the app projection but not the underlying group state.

`crates/marmot-app/src/client/projection.rs::reconcile_live_engine_groups` (line 186 on master `bd860cd3`) then unconditionally repairs any engine group missing from the projection:

```rust
if !projected.contains(group_id_hex.as_str()) {
    self.add_group(&group_id)?;
}
```

Its doc comment frames every absent projection as a repairable "engine/projection tear". A deliberate local deletion is indistinguishable from a tear, so reconciliation re-adds the row. An app update is simply a reliable way to trigger reconciliation.

## What is missing

A deletion frontier: durable state recording that a group was locally deleted, and at what point, so reconciliation can distinguish "never projected" from "deliberately removed" and leave the latter alone — while still allowing genuinely new activity to resurrect the conversation if that is the desired product behaviour.

Both halves of that behaviour are worth deciding explicitly:
- should a locally deleted group stay gone permanently, or
- should it return when a new message arrives after the deletion point?

The second is why a frontier (a timestamp or epoch) is more useful than a boolean tombstone.

## Client-side surface

`delete_group_local(account_ref, group_id_hex) -> bool` currently takes no frontier and returns nothing that survives reconciliation. Whatever shape is chosen, the client needs a way to record the deletion and reconciliation needs to honour it.

## Verification

Delete a group locally, trigger reconciliation without any new inbound message for that group, and assert the projection does not re-add it.

## Consumer

- whitenoise-android#1646

agent-p1p:
## Summary

- persist a per-group local-delete frontier alongside the app projection wipe
- keep hidden live MLS groups subscribed while suppressing reconciliation, historical replay, and non-chat state updates
- resurrect only for a causally newer authenticated chat message or a fresh Welcome
- conservatively migrate already-absent live projections and make stale projection writers preserve the deletion marker

## Verification

- `just fast-ci`
- `cargo test -p cgka-traits --locked`
- `cargo test -p storage-sqlite --locked`
- `cargo test -p cgka-session --features test-policy-overrides --locked`
- `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks --locked`
- `cargo test -p marmot-app --lib --features test-policy-overrides --locked`
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked -- --test-threads=1` (96 passed)

Fixes #1238

## Summary by CodeRabbit

* **Bug Fixes**
  * Locally deleted groups remain deleted across restarts, synchronization, and historical replay.
  * Stale or non-chat updates no longer resurrect deleted groups.
  * Newer chat activity or valid re-invites can restore groups correctly.
  * Connectivity and routing are preserved for deleted groups and restored after errors.
  * Received message events now include content-derived identifiers.
  * Application events are durably recovered after interruptions and removed after acknowledgement.
  * Group deletion and event processing now complete atomically, preventing inconsistent state after failures.

* **Documentation**
  * Added changelog details covering persistent local deletions and synchronization behavior.


agent-p1p:
BLOCKING:
- Prior-route relay endpoints are lost by local deletion: `crates/marmot-app/src/client/projection.rs:381-403` reconstructs hidden subscriptions by taking every durable transport route ID but assigning all of them `current_route.relays`. The existing overlap model stores each prior route as its own `(nostr_group_id, relays)` pair (`AppPriorNostrRoute::subscription`), because a routing rotation may change both the ID and relay set. `delete_local_group_data` deletes the `account_groups` row that owns `prior_nostr_routes_json`, so after a rotation from `(old_id, old_relay)` to `(new_id, new_relay)` this code subscribes to `(old_id, new_relay)` rather than `(old_id, old_relay)`. A retained-history sender publishing on the valid old route will not reach the hidden device, so the fresh chat that should resurrect the conversation can be missed. Preserve or reconstruct the exact prior subscriptions across local deletion, and add a rotation test with different old/current relays that covers delete, restart, and resurrection.

SUGGESTION:
- none

NITPICK:
- none

SUMMARY: The durable frontier and stale-writer defenses cover the reported restart/reconciliation path, and focused local storage/session/runtime regressions pass. The route-preservation path is incomplete when routing rotation changes relay endpoints, which breaks the promised resurrection edge. Two existing CodeRabbit actionable comments also remain unresolved and are not duplicated here: repeated deletes do not advance the frontier (https://github.com/marmot-protocol/mdk/pull/1366#discussion_r3754762622), and compensation can replace the original delete error (https://github.com/marmot-protocol/mdk/pull/1366#discussion_r3754762616).
VERDICT: REQUEST_CHANGES
SENSITIVE_PATHS: crates/cgka-engine/src/distributed_convergence.rs, crates/cgka-engine/src/message_processor/ingest.rs, crates/traits/src/engine.rs, crates/storage-sqlite/src/account_projection.rs, crates/storage-sqlite/src/migrations.rs, crates/storage-sqlite/src/migrations/0044_local_group_deletion_frontiers.rs, crates/marmot-app/src/client/mod.rs, crates/marmot-app/src/client/projection.rs, crates/marmot-app/src/client/sync.rs

crates/marmot-app/src/client/mod.rs
```diff
@@ -1511,20 +1513,47 @@ impl AppClient {
...
             Err(error) => {
                 if was_live {
                     self.state.groups = original_groups;
                     self.refresh_routing()?;
                     self.sync_runtime_groups().await?;
                 }
-                Err(error)
+                return Err(error);
+            }
+        };
```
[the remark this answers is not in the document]
agent-p1p: Fixed in 3bdfca0a. Local-delete markers now retain exact route-id/relay pairs durably, including routes that become current during multiple hidden rotations; restart rebuild filters them through the engine-owned active route index. Added delete/rotate-hidden/restart/resurrection coverage.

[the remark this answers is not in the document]
agent-p1p: Correction for this compensation thread: 3bdfca0a restores the original in-memory projection and exact routing snapshot directly, attempts transport resync best-effort, logs secondary failure privately, and always returns the original delete error. The injected regression combines a primary storage-delete failure with a compensation resubscribe failure.

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -826,17 +844,41 @@ impl SqliteAccountStorage {
...
+                if active && !terminal {
+                    tx.execute(
+                        "INSERT INTO local_group_deletion_frontiers (
+                            group_id_hex, message_insert_order
+                         ) VALUES (?1, ?2)
+                         ON CONFLICT(group_id_hex) DO NOTHING",
+                        params![hex::encode(&group_id), message_insert_order],
+                    )
+                    .storage()?;
+                }
```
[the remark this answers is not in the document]
agent-p1p: Fixed in 3bdfca0a. Repeated local deletion now advances the durable insertion-order frontier monotonically with MAX while preserving retained routing history. Regression coverage verifies buffered messages remain suppressed and a genuinely newer message clears the marker.

agent-p1p:
Follow-up route-hardening in 3bdfca0a: local-delete markers retain exact route-id/relay pairs across restart and across multiple routing rotations while the projection remains hidden. Rebuilds intersect the durable exact pairs with the engine-owned active route index. Added delete → rotate while hidden → restart → resurrection coverage. `just fast-ci` and targeted regressions pass.

agent-p1p:
Follow-up to the adversarial review: https://github.com/marmot-protocol/mdk/pull/1366#issuecomment-5248227934

Fixed in `3bdfca0a86e764600b4c18348ae363c0d662ea81`.

Resolutions:
- Exact route retention: local-delete frontiers now durably retain each authenticated `(transport_group_id, endpoints)` pair. Hidden rotations accumulate exact pairs, restart reconstruction intersects them with the engine-owned active route index, and fresh authenticated chat still resurrects the projection.
- Repeated delete: the stored `message_insert_order` frontier advances monotonically with `MAX(existing, current)`, so buffered/pre-second-delete messages cannot clear it.
- Compensation: failed storage deletion restores the original projection and routing snapshot, attempts transport resync best-effort even after another compensation failure, logs only privacy-safe secondary diagnostics, and returns the original delete error.

Post-review changed files:
- `crates/marmot-app/src/client/mod.rs`
- `crates/marmot-app/src/client/projection.rs`
- `crates/marmot-app/src/groups.rs`
- `crates/marmot-app/src/tests.rs`
- `crates/storage-sqlite/src/account_projection.rs`
- `crates/storage-sqlite/src/account_projection/tests.rs`
- `crates/storage-sqlite/src/migrations.rs`
- `crates/storage-sqlite/src/migrations/0044_local_group_deletion_frontiers.rs`

Verification:
- `just fast-ci` — pass
- `cargo test -p storage-sqlite --locked` — pass (334 tests)
- New marmot-app compensation and delete/hidden-rotation/restart/resurrection regressions — pass
- `cargo test -p cgka-session --features test-policy-overrides --locked` — pass on the branch; the post-review commit does not change session code
- GitHub CI for `3bdfca0a` — all checks terminal green; release-publish job skipped as expected

All CodeRabbit review threads are resolved. The PR remains draft pending the fresh post-fix adversarial review and sensitive-code merge gate.

agent-p1p:
BLOCKING:
- `crates/marmot-app/src/client/sync.rs:961-969` clears all qualifying deletion frontiers before it starts suppressing/projecting the effects batch. A single convergence/ingest batch can contain several replayed application messages for one group (`emit_application_replay_events` emits one event per observation). If that batch contains an old valid chat whose durable insert order is at/below frontier N and a fresh chat above N, the fresh event deletes the marker during the pre-pass; the old event then reaches the main loop with no marker, bypasses `suppress_local_deleted_group_event`, and is projected back into the wiped timeline. Event order does not save this because the entire clear pass finishes before projection. The same defect exists on the no-delivery drain path at `sync.rs:284-293`. Preserve the batch-start frontier for per-event classification, keep every at/below-frontier event suppressed even when another event crosses it, and clear only as part of successfully projecting/persisting the crossing event. Also compensate/retain the marker if later projection/save work fails. Add a regression with old+fresh valid chats in one effects batch (both event orders) and assert that only the fresh message resurfaces.

SUGGESTION:
- `crates/storage-sqlite/src/account_projection.rs:1017-1069` accumulates every unique hidden-route `(id, relays)` pair forever. Rebuild filters subscriptions through the engine-owned active route index, so runtime subscriptions stay bounded, but `prior_nostr_routes_json` does not. Prune the durable exact-pair history against the active route set, or use a policy-derived bound, and cover advancement past the retained route window.

NITPICK:
- none

SUMMARY:
- The prior blocking route-pair finding is resolved at `3bdfca0a`: old/current/hidden routes retain their own endpoint sets across delete, hidden rotation, and reopen.
- Repeated deletion advances the ingress frontier monotonically with SQL `MAX`; compensation restores the exact routing snapshot, attempts transport restoration, logs secondary failure, and preserves the primary error.
- The two actionable CodeRabbit threads are resolved. Its app-wiring test nit is already covered by the relay-runtime delete/restart/historical-replay/fresh-message test. The JSON replacement concern is not reachable through production ordering: deletion and projection saves serialize as immediate transactions, and saves exclude frontier-marked groups.
- Release hygiene passes: `crates/cli/CHANGELOG.md` has an Unreleased entry; no version, lockfile-version, fixture-cohort, or release-tag churn.
- Live GitHub checks are green at exact head `3bdfca0a86e764600b4c18348ae363c0d662ea81`.
- Local verification passed: `cargo fmt --all -- --check`; 2 storage route/frontier tests; secure-delete committed-outcome test; 3 marmot-app local-delete unit tests; relay-runtime local-delete integration; session message-id integration; focused all-target clippy for `storage-sqlite` and `marmot-app` with `-D warnings`.
- Sensitive paths touched: CGKA replay/ingest events, group-state projection/routing, SQLCipher account projection, and schema migration 0044.

VERDICT: REQUEST_CHANGES

agent-p1p:
Addressed the follow-up resurrection review in f06da556.

- Snapshot local-delete frontiers once per effects batch, so historical and fresh events are order-independent.
- Clear only the expected frontier in the same SQLite transaction that persists the recreated projection; save failure rolls the marker back.
- Adopt exact retained route-id/relay history into the recreated group before clearing the marker, and prune durable history against the engine-owned active route window.
- Added mixed old/fresh ordering, route adoption/pruning, and persistence-failure regressions.

Validation: `just fast-ci`; targeted marmot-app and storage-sqlite regressions.

crates/marmot-app/src/client/sync.rs
```diff
@@ -867,6 +900,122 @@ impl AppClient {
...
+    pub(crate) fn save_state_with_pending_local_group_deletion_frontier_clears(
+        &mut self,
+    ) -> Result<(), AppError> {
+        let frontiers_to_clear = self
+            .pending_local_group_deletion_frontier_clears
+            .iter()
+            .map(|(group_id_hex, frontier)| (group_id_hex.clone(), *frontier))
+            .collect::<Vec<_>>();
+        self.app
+            .save_state_clearing_local_group_deletion_frontiers(&self.state, &frontiers_to_clear)?;
```
erskingardner: [P1] Recover a crossing message after a crash before this save

This transaction makes the marker clear atomic with the account-group write, but engine ingest has already committed the message and drained its one-shot `MessageReceived` before execution reaches this call. If the process dies after `runtime.ingest_delivery` (or scheduled convergence) marks a fresh chat `Processed` but before this save, the frontier remains. On restart, relay redelivery classifies that engine message as already seen and hydration does not re-emit processed application events, so no later batch calls `prepare_local_group_deletion_frontier_clear`; the chat stays hidden until a *second* post-delete chat arrives. That violates the promised resurrection on the first genuinely newer message and the repository torn-write invariant. Please persist a durable crossing/resurrection intent before this boundary, or make account-open recovery detect an authenticated processed chat whose insert order exceeds the frontier and atomically rebuild the projection plus clear it. Add a crash/reopen regression at this exact boundary.

erskingardner (changes requested):
The current head fixes the earlier exact-route, repeated-delete, compensation, and batch-ordering problems, but one crash-consistency gap remains blocking. Engine ingest/convergence becomes durable before the deletion-frontier clear and resurrected account projection are committed, so the first genuinely newer chat can be consumed without ever restoring the chat after a crash. CI is also still in progress at this head.

crates/marmot-app/src/client/sync.rs
agent-p1p: Fixed in 263bf0c1. Authenticated MessageReceived deliveries now enter a durable SQLCipher outbox in the same transaction that marks direct ingest Processed or commits convergence. Engine open reloads pending deliveries; the app then atomically persists the repaired projection, compare-clears the local-delete frontier, and acknowledges the outbox row. Added engine restart, storage rollback, conflicting-id fail-closed, and account-open first-chat resurrection regressions.

crates/storage-sqlite/src/migrations/0044_local_group_deletion_frontiers.rs
```diff
@@ -0,0 +1,50 @@
...
+-- by local group deletion; the source message's durable ingress order provides
+-- deterministic replay after a crash.
+CREATE TABLE pending_application_events (
+    message_id BLOB PRIMARY KEY NOT NULL,
+    group_id BLOB NOT NULL,
+    message_insert_order INTEGER NOT NULL CHECK(message_insert_order >= 0),
+    event_json BLOB NOT NULL
+);
+CREATE INDEX idx_pending_application_events_group
+    ON pending_application_events(group_id, message_insert_order);
```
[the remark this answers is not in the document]
erskingardner: Fixed in 66548166. Single-message deletion now removes its pending application event in the same transaction; protocol-group deletion removes the group-scoped outbox rows before commit; and full snapshot restore reconciles pending rows after restoring messages, preserving deliveries whose source message survives while pruning rows for discarded messages. Added regression coverage for all three boundaries. Validation: `cargo test -p storage-sqlite --locked` (337 passed), focused all-target clippy with `-D warnings`, and `just fast-ci`.

crates/marmot-app/src/client/sync.rs
```diff
@@ -867,6 +915,225 @@ impl AppClient {
...
+        &mut self,
+        event: &cgka_traits::engine::GroupEvent,
+        frontier: u64,
+    ) -> Result<(), AppError> {
+        let Some(group_id) = event_group_id(event) else {
+            return Ok(());
+        };
+        if !self.adopt_local_deleted_group_prior_routes(group_id)? {
+            return Ok(());
+        }
```
erskingardner, line 1019: [P1] Do not acknowledge a crossing delivery unless the frontier clear is actually queued.

`prepare_local_group_deletion_frontier_clear` returns `Ok(())` when `adopt_local_deleted_group_prior_routes` fails because the group is still absent from `self.state`. Both observe loops then still call `prepare_pending_application_event_ack`.

On `observe_drained_session_events`, that path is reachable when `event_group_projection_best_effort` returns `None`: `observe_event` will not `add_group`, `project_received_message` can still persist timeline rows, the frontier stays, and the outbox row is deleted. After restart the message is already `Processed` and no longer pending, so the first causally newer chat never resurrects the conversation — the exact crash-recovery guarantee this outbox was added to provide.

Fail closed instead: if `crosses_frontier` is true and the group is still absent (or no clear was queued), return an error or skip the ack so the pending event remains durable. A regression around best-effort projection failure during account-open replay would lock this in.

crates/storage-sqlite/src/migrations/0044_local_group_deletion_frontiers.rs
```diff
@@ -0,0 +1,50 @@
...
+
+-- Authenticated application deliveries remain pending until the app projection
+-- commits. `event_json` is inside the account's SQLCipher database and is wiped
+-- by local group deletion; the source message's durable ingress order provides
+-- deterministic replay after a crash.
+CREATE TABLE pending_application_events (
+    message_id BLOB PRIMARY KEY NOT NULL,
+    group_id BLOB NOT NULL,
+    message_insert_order INTEGER NOT NULL CHECK(message_insert_order >= 0),
+    event_json BLOB NOT NULL
```
erskingardner: [P1] Name this column `record`, not `event_json`.

`crates/storage-sqlite/src/storage/AGENTS.md` requires serialized Marmot records in `record` blobs so trait shapes can evolve through Rust migrations. This is a new table, so renaming now is cheap; keeping `event_json` creates a one-off that every later reader/migration has to special-case.

Update the migration plus the put/list/duplicate-check/restore queries together.

crates/storage-sqlite/src/storage/groups.rs
```diff
@@ -390,6 +405,7 @@ mod tests {
                 .unwrap()
                 .is_empty()
         );
+        assert!(store.list_pending_application_events().unwrap().is_empty());
```
erskingardner, line 418: [P2] This assertion does not exercise the new cleanup.

`group_delete_cascades_...` inserts a message but never inserts a `pending_application_events` row, so `list_pending_application_events()` is empty even if the `DELETE FROM pending_application_events WHERE group_id = ?1` in `delete_group` is removed.

Mirror the `delete_message` / update-group coverage: `put_pending_application_event` for the stored message, delete the group, then assert the outbox is empty.

erskingardner (changes requested):
Request changes. The deletion-frontier approach matches #1238, and the engine-to-app outbox is the right durability shape for the crash window Jeff called out earlier. Two issues should be fixed before merge: crossing deliveries can still be acknowledged without a durable resurrection, and the new pending-event table breaks the storage `record` blob convention while the migration is still easy to rename.

crates/storage-sqlite/src/storage/groups.rs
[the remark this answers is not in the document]
erskingardner: Addressed in 580b69e. The group-delete regression now persists a real pending `MessageReceived` event before deletion, so the empty-outbox assertion exercises the cascade instead of passing vacuously. The full storage-sqlite suite passes (337 tests).

crates/marmot-app/src/client/sync.rs
erskingardner: Addressed in 580b69e. Crossing application events are now acknowledged only when resurrection adopted the retained routes and queued the matching frontier clear. If best-effort group projection is unavailable, both the pending event and deletion frontier remain durable for the next account-open replay. Added a regression for that exact replay seam; the full marmot-app suite passes (612 tests), as does `just fast-ci`.

crates/storage-sqlite/src/migrations/0044_local_group_deletion_frontiers.rs
erskingardner: Addressed in 580b69e. Migration 0044 now names the serialized pending-event blob `record`, and every insert, duplicate-content check, and replay query uses that column consistently. The full storage-sqlite suite and `just fast-ci` pass.

crates/storage-sqlite/src/storage/groups.rs
erskingardner: Addressed in 580b69e. The cascade test now inserts a pending application event tied to the stored source message before deleting the group, so the assertion proves the outbox row is removed. The full storage-sqlite suite passes (337 tests).

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -876,6 +1025,215 @@ impl SqliteAccountStorage {
...
+            )
+            .storage()?;
+            Ok(())
+        })
+    }
+
+    /// Clear a local-delete marker only when `message_id` belongs to this group
+    /// and was durably inserted after the deletion frontier. Exact or rewrapped
+    /// historical replay resolves to an older existing row and stays suppressed.
+    pub fn clear_local_group_deletion_frontier_if_message_is_newer(
```
erskingardner, line 1169: Non-blocking: remove this obsolete standalone clear primitive. Production now correctly classifies with `local_group_deletion_message_is_newer_than` and clears the expected frontier in the projection/ack transaction; this public method is only exercised by tests and permits the exact non-atomic “clear before projection commit” sequence the final design avoids. Delete it and update those tests to assert classification plus transactional clear, or keep it private/test-only if the helper is still useful.

erskingardner (dismissed):
Re-reviewed exact head `f5c2ee99` after the merge from master. The batch-start frontier, transactional resurrection/outbox acknowledgment, exact hidden-route retention, repeated-delete advancement, and failure-safe replay paths now satisfy #1238 and the downstream multi-group/restart acceptance criteria. The focused local validation and all completed core GitHub checks are green; three packaging/simulator jobs are still running. One non-blocking cleanup is noted inline.

PR #1367, opened 2026-08-11
Let a removed member rejoin: skip evicted-era snapshots in the past-epoch peel probe

dannym-arx:
Stacked on #1364.

A marmot evicted from the burrow and invited back in should walk through the new door, not explode in the tunnel. Before this change, a member removed and then re-invited with a fresh KeyPackage rejoined through the Welcome (the mdk#557 stale-state repair), but the re-add commit still reached it as an old-epoch group message. The past-epoch peel probe rolled the group back to a retained snapshot from the member's evicted era, and `build_group_context_snapshot` failed the whole ingest with `export_secret: GroupStateError(UseAfterEviction)`. This is the engine gap that forced `leaver-removal-secrecy/v1` to drop its re-add leg.

Changes:

- `context_from_group_snapshot` reports an evicted snapshot group as unusable (`Ok(None)`) instead of erroring: no exporter secret can exist for it. `try_peel_group_message_from_available_snapshots` skips such snapshots the same way it skips an undecryptable one, so the re-add commit stays on its normal deferred path.
- New harness regression `removed_member_rejoins_via_fresh_welcome`: remove Carol, re-invite her, and require post-rejoin application traffic to flow.
- New portable vector `readd-after-eviction/v1` (registered in the manifest and `SCENARIOS.md`), which passes the strict oracle. The stale `SCENARIOS.md` note about the dropped re-add leg now points here.
- Second commit: the engine subject's `send_application` now routes a refused app send through `try_send_app` + `subject_engine_error`, the same shape as the leave fix in #1364. A send refused with `InvalidTransition` (for example after a requested leave) becomes a failed scenario step with a failure capsule instead of a process panic.

Verification: the new regression and vector pass; all 29 portable vectors pass strict; the full cgka-engine suite passes under the CI invocation (459 tests); the full simulator suite and `just fast-ci` are green. Leave-then-re-add, double re-add, and duplicated-Welcome trial scenarios also pass with this change.

Found and confirmed while probing, deliberately not fixed here: an invitee welcomed onto a fork branch that loses recovery stays stranded even after an admin re-invites it with a fresh KeyPackage. The `self_is_recorded_member` guard in `do_join_peeled_welcome` refuses the replacement Welcome as `WelcomeAlreadyProcessed`. Relaxing that guard is a protocol decision and needs its own design pass.

## Summary by CodeRabbit

- **New Features**
  - Added support for members rejoining a group after removal using a fresh invitation.
  - Rejoined members can resume sending and receiving application messages.
  - Added conformance coverage for removal, re-invitation, state recovery, and group convergence.

- **Bug Fixes**
  - Improved recovery from inactive historical group states by continuing with available recovery data.
  - Failed application sends after leaving now produce clear validation failures without stopping subsequent processing.

- **Documentation**
  - Documented the rejoin-after-removal scenario and added its portable test vector.

### DIFF
--- a/crates/cgka-conformance-simulator/src/subject.rs
+++ b/crates/cgka-conformance-simulator/src/subject.rs
@@ -1393,7 +1393,10 @@ impl ConvergenceSubject for EngineHarnessSubject {
     ) -> Result<(), SubjectError> {
         let sender = self.client_mut(action.sender)?;
         sender.name_next_scenario_input(action.action_id);
-        sender.send_app(action.payload.as_bytes().to_vec()).await;
+        sender
+            .try_send_app(action.payload.as_bytes().to_vec())
+            .await
+            .map_err(subject_engine_error)?;
         Ok(())
     }
 
--- a/crates/cgka-conformance-simulator/SCENARIOS.md
+++ b/crates/cgka-conformance-simulator/SCENARIOS.md
@@ -226,8 +226,19 @@ These are the scenarios another implementation should be able to load from JSON 
   the full transcript minus the departed members' silence and converge at epoch 3 with two members. `no_pending_work`
   covers Alice and Bob only: like the latecomer vector's pre-join input, Dave and Carol legitimately retain their
   undecryptable post-departure inputs as deferred transport work. The original harness scenario's re-add leg (removed
-  member rejoins via a fresh KeyPackage) is deliberately absent: welcome-after-eviction currently fails in the engine
-  harness with `GroupStateError(UseAfterEviction)` and is tracked as a coverage gap.
+  member rejoins via a fresh KeyPackage) lives in `readd-after-eviction/v1`.
+
+### `readd-after-eviction/v1`
+
+- File: `vectors/readd-after-eviction.v1.json`
+- Setup: Alice creates a group with Bob and Carol, removes Carol, then re-invites her with a fresh KeyPackage. Carol
+  sends the first post-rejoin application message.
+- Pressure: the rejoin Welcome lands on stale evicted local state, and the re-add commit reaches Carol as an
+  old-epoch group message whose only retained peel snapshots are from her evicted era.
+- Expected: the Welcome supersedes the stale live OpenMLS state, evicted-era retained snapshots are skipped as peel
+  sources instead of failing ingest with `UseAfterEviction`, and all three clients converge at epoch 3 with Carol's
+  post-rejoin payload delivered. The Rust regression for this shape is
+  `removed_member_rejoins_via_fresh_welcome` in `tests/canonical_scenarios.rs`.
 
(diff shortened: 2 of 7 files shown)


erskingardner (changes requested):
The rejoin fix itself is well-scoped and has strong harness/vector coverage. One additional behavior added in the second commit is still only manually verified: refused application sends must become structured scenario failures without aborting the report run. Please check in that regression before merge.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -1393,7 +1393,10 @@ impl ConvergenceSubject for EngineHarnessSubject {
     ) -> Result<(), SubjectError> {
         let sender = self.client_mut(action.sender)?;
         sender.name_next_scenario_input(action.action_id);
-        sender.send_app(action.payload.as_bytes().to_vec()).await;
+        sender
+            .try_send_app(action.payload.as_bytes().to_vec())
+            .await
+            .map_err(subject_engine_error)?;
```
erskingardner, line 1399: Add the regression for this new refusal path. The new re-add test and vector only exercise successful application sends, so CI can stay green if this slips back to the panicking send_app helper. Please add a report-runner scenario that reaches Leaving, attempts send_app_message, and asserts an invalid_transition failed step/failure capsule plus sibling-vector continuation and a non-zero report exit. That makes the behavior claimed in the PR body reproducible instead of manual-only.

Datawav (commented):
## Automated review

No actionable findings.

crates/cgka-conformance-simulator/src/subject.rs
dannym-arx: Done in f59af83b: `refused_application_send_fails_step_without_aborting_the_run` in `tests/report_runner.rs` runs a two-vector report where one scenario reaches Leaving and then scripts `send_app_message`. It asserts the structured `invalid_transition` failed step, the failure capsule on disk, the sibling vector completing with zero failures, and the run summary reporting one failure (the CLI's non-zero exit condition). d3fb8b60 additionally moved the test's fixture setup to the fs-private helpers.

Datawav:
## Automated review

No prior actionable findings were carried into this follow-up.

**Low — unbounded route-history growth.** In `crates/storage-sqlite/src/account_projection.rs:1025`, `retain_local_group_deletion_nostr_routes` appends every distinct `(nostr_group_id_hex, relays)` combination, while pruning only inactive route IDs. Repeated relay changes for an active route can therefore grow `prior_nostr_routes_json` indefinitely, increasing storage and rewrite costs for long-hidden groups. Prune superseded relay variants per route ID or enforce a bounded recent-history limit.

Datawav (approved):

erskingardner (approved):

PR #1360, opened 2026-08-10
fix(cgka-engine): persist leave proposal atomically

agent-p1p:
## Summary
- persist the outbound SelfRemove proposal row, durable leave request, and distinct content-dedup marker in one storage transaction
- update sent-ID and leave caches only after the transaction commits
- add fault-injection coverage for each of the three writes, restart/retry behavior, and the transport-ID-equals-content-ID single-row path

## Verification
- `just fast-ci`
- `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks --locked`
- `cargo test -p cgka-engine --test auto_commit_atomicity --locked`
- `cargo test -p cgka-engine --test invite_leave --locked`

## Release hygiene
- Updated `crates/cli/CHANGELOG.md` under Unreleased
- No version or fixture bumps

Fixes #1354

## Summary by CodeRabbit

- **Bug Fixes**
  - Leave proposals and related messages are now saved atomically, preventing incomplete state when storage fails.
  - Retrying a leave operation after a failure no longer creates duplicate content markers.
  - Sent-message state is updated only after successful persistence.

- **Tests**
  - Added coverage for storage failures, transaction rollback, retry behavior, and duplicate-prevention scenarios.

- **Documentation**
  - Documented the fix in the upcoming release notes.

PR #1372, opened 2026-08-11
Add cross-adapter convergence assurance

erskingardner:
## Summary

- reconcile the convergence reliability tracker with the merged all-family isolated runner, the reviewed 1,170-case campaign, and the production convergence changes that landed afterward
- preserve canonical accepted-publication checkpoints across the staged engine harness and the auto-publishing app-runtime/process adapters
- enforce the same accepted-label, unknown-label, unlabelled-drain, and rollback-refusal semantics across app and process adapters
- run one three-participant profile/message/restart scenario unchanged through the engine, in-process app runtime, and isolated app processes
- run a companion offline/full-history recovery scenario through the app-runtime and process adapters that declare participant-connectivity support
- update the route matrix and property documentation with the exact evidence boundary

## Why

The existing cross-adapter check used one participant and no publication lifecycle. A realistic multi-member scenario could not run unchanged: the engine harness requires explicit acknowledgement of staged transport artifacts, while the production-shaped app runtime publishes inside a successful command and exposes no unresolved transport artifact to acknowledge.

The simulator-only adapter seam records that a named app/process mutation already succeeded through that production command surface. An `accepted` checkpoint can therefore remain in canonical Scenario IR. Auto-publishing adapters implement an empty idempotent outbound poll, reject unknown labels, and classify a later `reached_no_endpoint` outcome as an expected refusal because an accepted publication cannot be rolled back.

## Assurance boundary

This PR adds public protocol/application projection evidence. It does **not** change production convergence or app behavior, weaken the four-party cross-route vectors, or claim exact MLS-state commitments, active decryptability, complete application dispositions, container evidence, or VM evidence from adapters that cannot expose those surfaces.

On app/process subjects, a labelled accepted checkpoint is command-success evidence; the resulting pending-resolution observation is not an independent transport-delivery proof.

The reviewed 1,170-case campaign is recorded against its exact tested snapshot. Later production changes (#1329, #1360, and #1365) remain scoped current-`master` delta work rather than being treated as covered by that historical green run.

## Verification

- `cargo test -p cgka-conformance-simulator --lib` (139 passed)
- `cargo test -p cgka-conformance-simulator --test app_runtime_adapter` (6 passed)
- `cargo test -p cgka-conformance-simulator --test process_orchestrator` (10 passed)
- `just fast-ci`

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking and acknowledging outbound publications that were already accepted.
  * Expanded cross-adapter scenarios covering messaging, restarts, offline recovery, membership, and protocol state.
  * Added validation for publication labels and clearer handling of accepted, rejected, unknown, and missing acknowledgements.

* **Documentation**
  * Updated conformance guidance, route evidence, adapter behavior, and reliability progress documentation.
  * Added details on publication-label matching and recovery equivalence checks.

### DIFF
--- a/crates/cgka-conformance-simulator/README.md
+++ b/crates/cgka-conformance-simulator/README.md
@@ -279,8 +279,11 @@ retired or re-armed from the same acknowledgement. A state transition with no tr
 no-op publication rather than inventing a synthetic artifact. ScenarioSpec v2 and v3 drive the same contract through
 `acknowledge_outbound`: the runner polls opaque artifacts, optionally selects a stable scenario publication label and
 artifact role, and applies the declared transport outcome. Engine-generated work has no publication label and can be
-acknowledged as the client's current unresolved outbound set. There is no compatibility constructor, silent
-auto-confirmation mode, or second subject publication capability. `ProtocolProfile::Legacy` is separate from that
+acknowledged as the client's current unresolved outbound set. Production-shaped app adapters publish inside their
+command surface rather than exposing transport artifacts; after a successful named mutation they report that its
+publication is already accepted, allowing the same accepted checkpoint to remain in canonical IR. They reject a later
+`reached_no_endpoint` outcome because an accepted publication cannot be rolled back. There is no compatibility
+constructor or silent auto-confirmation mode in the engine subject. `ProtocolProfile::Legacy` is separate from that
 removed lifecycle: it selects the engine's legacy application-profile compatibility through
 `legacy_compatibility_profile()`, but it uses the same explicit outbound contract as `ProtocolProfile::Current`.
 ScenarioSpec v1 is intentionally unsupported after the repository-owned v2 cutover and is rejected before any subject
@@ -410,10 +413,13 @@ omitted values and clear explicitly empty values. At least one field is required
 `group_profile` to assert the requested values directly; `clients_converged` also compares both profile fields.
 
 Staged publications are referenced by string labels chosen inside the scenario. `acknowledge_outbound.publication`
-selects artifacts emitted by that operation; omitting it selects all currently unresolved artifacts for the client.
-The optional `selection` further restricts the set to `all`, `state_confirmation`, `welcome`, `group_message`, or
+selects artifacts emitted by that operation; on an auto-publishing app adapter, the same label identifies the completed
+mutation whose command already returned successfully. Omitting the label selects all currently unresolved artifacts
+for adapters that expose them.
+The optional `selection` further restricts an exposed artifact set to `all`, `state_confirmation`, `welcome`, `group_message`, or
 `regenerated_queued_intent`. `outcome` is either `accepted` or `reached_no_endpoint`. A labelled acknowledgement fails
-when no unresolved artifact matches, while an unlabelled acknowledgement is an idempotent drain and may match nothing.
+when neither an unresolved artifact nor an already-accepted named app mutation matches, while an unlabelled
+acknowledgement is an idempotent drain and may match nothing.
 `group_message` includes both application messages and proposals because both use the group-message transport envelope.
 Proposal artifacts are engine-generated and have no scenario publication label, so scenarios select them with an
 unlabelled `group_message` acknowledgement unless a future proposal action introduces labels. Client labels are stable
--- a/crates/cgka-conformance-simulator/src/app_runtime.rs
+++ b/crates/cgka-conformance-simulator/src/app_runtime.rs
@@ -25,8 +25,8 @@ use crate::{
     ClientEventCounts, ClientObservation, ConvergenceSubject, ForkRecoveryObservation,
     ScenarioAdminPolicyObservation, ScenarioInputLedgerEntry, SubjectCapability,
     SubjectCreateGroup, SubjectDescriptor, SubjectError, SubjectFailureCategory,
-    SubjectInviteMembers, SubjectRemoveMembers, SubjectSelfUpdate, SubjectSendApplication,
-    SubjectUpdateAdminPolicy, SubjectUpdateGroupData,
+    SubjectInviteMembers, SubjectOutboundArtifact, SubjectOutboundOutcome, SubjectRemoveMembers,
+    SubjectSelfUpdate, SubjectSendApplication, SubjectUpdateAdminPolicy, SubjectUpdateGroupData,
 };
 
 pub const APP_RUNTIME_OBSERVATION_SCHEMA_VERSION: &str = "1";
@@ -119,7 +119,7 @@ pub struct AppRuntimeHarness {
     participants: BTreeMap<String, Participant>,
     scenario_groups: BTreeMap<String, GroupId>,
     active_scenario_group: Option<String>,
-    accepted_publications: BTreeSet<(String, String)>,
+    accepted_publications: BTreeMap<String, BTreeSet<String>>,
 }
 
 impl AppRuntimeHarness {
@@ -174,7 +174,7 @@ impl AppRuntimeHarness {
             participants,
             scenario_groups: BTreeMap::new(),
             active_scenario_group: None,
-            accepted_publications: BTreeSet::new(),
+            accepted_publications: BTreeMap::new(),
         })
     }
 
@@ -347,6 +347,13 @@ impl AppRuntimeHarness {
             .iter()
             .map(|label| Ok(self.participant(label)?.account_id.clone()))
             .collect()
+    }
+
+    fn record_accepted_publication(&mut self, client: &str, publication: &str) {
+        self.accepted_publications
+            .entry(client.to_owned())
+            .or_default()
+            .insert(publication.to_owned());
     }
 
     async fn refresh_cached_members(&mut self, clients: &[String]) -> Result<(), SubjectError> {
@@ -641,8 +648,7 @@ impl ConvergenceSubject for AppRuntimeHarness {
         self.scenario_groups.insert(group_label, group_id.clone());
         self.apply_admin_set(action.creator, &group_id, action.initial_admins)
             .await?;
-        self.accepted_publications
-            .insert((action.creator.to_owned(), action.pending.to_owned()));
+        self.record_accepted_publication(action.creator, action.pending);
         Ok(())
     }
 
@@ -658,8 +664,7 @@ impl ConvergenceSubject for AppRuntimeHarness {
             .invite_members(&participant.account_id, &group_id, &invitees)
             .await
             .map_err(app_error)?;
-        self.accepted_publications
-            .insert((action.inviter.to_owned(), action.pending.to_owned()));
+        self.record_accepted_publication(action.inviter, action.pending);
         Ok(())
     }
 
@@ -679,8 +684,7 @@ impl ConvergenceSubject for AppRuntimeHarness {
             )
             .await
             .map_err(app_error)?;
-        self.accepted_publications
-            .insert((action.client.to_owned(), action.pending.to_owned()));
+        self.record_accepted_publication(action.client, action.pending);
         Ok(())
     }
 
@@ -696,8 +700,7 @@ impl ConvergenceSubject for AppRuntimeHarness {
             .remove_members(&participant.account_id, &group_id, &members)
             .await
             .map_err(app_error)?;
-        self.accepted_publications
-            .insert((action.remover.to_owned(), action.pending.to_owned()));
+        self.record_accepted_publication(action.remover, action.pending);
         Ok(())
     }
 
@@ -709,8 +712,7 @@ impl ConvergenceSubject for AppRuntimeHarness {
             .schedule_manual_self_update(&participant.account_id, &group_id)
             .await
             .map_err(app_error)?;
-        self.accepted_publications
-            .insert((action.client.to_owned(), action.pending.to_owned()));
+        self.record_accepted_publication(action.client, action.pending);
         Ok(())
     }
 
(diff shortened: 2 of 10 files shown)


crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -585,6 +587,7 @@ impl ConvergenceSubject for AppRuntimeHarness {
                 SubjectCapability::EventObservation,
                 SubjectCapability::AdminPolicyObservation,
                 SubjectCapability::CrashReopen,
+                SubjectCapability::OutboundPublication,
```
erskingardner, line 597: This declares `outbound_publication`, but the harness overrides neither `poll_outbound` nor `acknowledge_outbound` — both still resolve to the trait defaults returning `SubjectError::unsupported(OutboundPublication)`.

The README defines this capability as "returns exact transport-ready artifacts through non-destructive polling and accepts typed `accepted` / `reached_no_endpoint` outcomes", none of which is true here. Concretely, after this change any acknowledge step that is *not* a labelled hit in `accepted_publications` passes preflight and then fails mid-run:

- unlabelled drain (`accept_all_outbound` in `stateful_generator.rs` / `family.rs`) → `poll_outbound` → `unsupported_capability`, where the README promises "an idempotent drain [that] may match nothing";
- `rollback-faults`' `acknowledged_step("alice", "update", ReachedNoEndpoint)` → the new "cannot be rolled back" scenario error.

Before this PR both were refused up front with `unsupported_subject_capability`. Suggest either implementing the two methods here (empty `poll_outbound`; `acknowledge_outbound` accepting only a recorded publication with `Accepted`, refusing otherwise as a classified `ExpectedRefusal`) so the declaration is honest, or introducing a distinct capability for the auto-publishing seam so preflight keeps telling the truth.

crates/cgka-conformance-simulator/src/scenario.rs
```diff
@@ -905,29 +905,47 @@ async fn execute_scenario_step(
...
-                    .acknowledge_outbound(client, &artifact.outbound_id, *outcome)
-                    .await
-                    .map_err(|error| subject_step_error(step_index, error))?;
+                    .scenario_publication_already_accepted(client, publication)
+                    .map_err(|error| subject_step_error(step_index, error))?
+            } else {
+                false
+            };
+            if already_accepted {
+                if *outcome != SubjectOutboundOutcome::Accepted {
```
erskingardner, line 914: This is the fail-closed property the PR body and PROPERTY_TESTS.md advertise, and nothing exercises it — both new scenarios acknowledge only labelled publications with `Accepted`. Worth a focused test in `tests/app_runtime_adapter.rs` covering the three non-happy arms on this adapter: `ReachedNoEndpoint` on an accepted label (should hit this branch), a labelled acknowledgement for an unrecorded label, and an unlabelled drain.

crates/cgka-conformance-simulator/README.md
```diff
@@ -279,8 +279,11 @@ retired or re-armed from the same acknowledgement. A state transition with no tr
 no-op publication rather than inventing a synthetic artifact. ScenarioSpec v2 and v3 drive the same contract through
 `acknowledge_outbound`: the runner polls opaque artifacts, optionally selects a stable scenario publication label and
 artifact role, and applies the declared transport outcome. Engine-generated work has no publication label and can be
-acknowledged as the client's current unresolved outbound set. There is no compatibility constructor, silent
-auto-confirmation mode, or second subject publication capability. `ProtocolProfile::Legacy` is separate from that
+acknowledged as the client's current unresolved outbound set. Production-shaped app adapters publish inside their
+command surface rather than exposing transport artifacts; after a successful named mutation they report that its
+publication is already accepted, allowing the same accepted checkpoint to remain in canonical IR. They reject a later
+`reached_no_endpoint` outcome because an accepted publication cannot be rolled back. There is no compatibility
```
erskingardner, line 285: "They reject a later `reached_no_endpoint` outcome" holds only for the in-process app-runtime adapter. `process_orchestrator.rs` handles `ScenarioStep::AcknowledgeOutbound { client, .. }` by returning `AlreadyPublished` unconditionally, discarding both the label and the outcome, so the process adapter silently accepts a rollback acknowledgement.

That's pre-existing, but this paragraph now asserts a property across "production-shaped app adapters" as a class. Either apply the same rule in the orchestrator (a few lines at the existing match arm) or narrow the sentence to the adapter that actually enforces it.

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -361,6 +361,19 @@ pub trait ConvergenceSubject: Send {
         Err(SubjectError::unsupported(SubjectCapability::VirtualTime))
     }
 
+    /// Return whether an adapter already published and accepted a named
+    /// scenario action through its production-shaped command surface. Engine
+    /// subjects return `false` and expose transport-ready artifacts for an
+    /// explicit acknowledgement; auto-publishing app adapters override this
+    /// so the same canonical IR can retain its publication checkpoint.
+    fn scenario_publication_already_accepted(
```
erskingardner: Nit: no implementation can fail — the default returns `Ok(false)` and the app harness returns `Ok(contains(..))` — so `-> bool` would drop a `?` and an error branch from the call site. If a future adapter needs to fail here, the acknowledgement step already has a natural error path.

Also minor: `contains(&(client.to_owned(), publication.to_owned()))` allocates two `String`s per lookup; `BTreeMap<String, BTreeSet<String>>` avoids that. Irrelevant at test scale, mentioned only because it's the kind of thing this crate usually avoids.

crates/cgka-conformance-simulator/tests/process_orchestrator.rs
```diff
@@ -149,44 +149,116 @@ fn controlled_relay_topology() -> ScenarioTopologyV2 {
     }
 }
 
-fn cross_adapter_scenario() -> ScenarioSpec {
+fn cross_adapter_scenario(include_offline_recovery: bool) -> ScenarioSpec {
```
erskingardner: Nit: a bare `bool` parameter that also switches the scenario name reads poorly at the three unrelated call sites (`cross_adapter_scenario(false)` in the saved-input test and the CLI report test say nothing about what `false` means). Two thin wrappers over a shared builder — `cross_adapter_restart_scenario()` / `cross_adapter_offline_scenario()` — would keep the dedup and lose the flag.

Same vein: the ~40 lines of `AppRuntimeHarness::new` → run → collect-observations → `ProcessOrchestrator::launch` → run → collect-observations are copied verbatim into both tests. A single helper returning the two observation maps would cut most of the net addition here.

erskingardner (commented):
The premise is sound and the disclosure discipline is exactly right for this repo: the assurance boundary is written down, the historical campaign is pinned to its tested snapshot instead of being carried forward, and the README even deletes its own "no second subject publication capability" claim now that one exists. The offline companion scenario is genuinely testing something (`catch_up` does skip offline participants, so Bob really does miss the commit and message).

Two things I'd like fixed before merge, both small:

**1. `AppRuntimeHarness` now declares `outbound_publication` but implements none of that capability's contract.** It adds `SubjectCapability::OutboundPublication` to its descriptor while overriding neither `poll_outbound` nor `acknowledge_outbound`, so both still hit the trait defaults that return `unsupported_capability`. Preflight now says yes and execution says no. Reachable today via `run_generated_case_report_with_capture_on_subject(.., GeneratedSubjectKind::AppRuntime, ..)`: `stateful_generator.rs`/`family.rs` emit unlabelled `accept_all_outbound` drains (→ runtime `unsupported_capability` from `poll_outbound`, where the README promises an idempotent no-op) and `rollback-faults` emits `acknowledged_step("alice", "update", ReachedNoEndpoint)` (→ the new "cannot be rolled back" error). Both used to be a clean up-front `unsupported_subject_capability` refusal; they are now mid-run failures after half the scenario has executed, with messages that read like product defects. Either implement the two methods on the harness (empty poll, `acknowledge_outbound` accepting only a recorded publication with `Accepted` and refusing otherwise as a classified `ExpectedRefusal`), or give the auto-publishing seam its own capability so preflight stays truthful.

**2. The new fail-closed branch has no test.** Both the PR body and PROPERTY_TESTS.md state the invariant — an accepted publication cannot be rolled back — but every acknowledgement in both new scenarios is labelled and `Accepted`, so the `ReachedNoEndpoint` arm, the unknown-label arm, and the unlabelled arm are all unexercised on this adapter. A short test in `tests/app_runtime_adapter.rs` covers all three and is the thing that would have surfaced (1).

Non-blocking notes inline. One more worth stating in the docs rather than the code: on the app/process adapters a labelled `Accepted` acknowledgement is now self-certifying — the adapter records the acceptance during the mutation, and the runner still pushes a `PendingResolutionObservation{resolution:"confirmed"}` that satisfies `TraceExpectation::PendingResolution`. That expectation is vacuous on these subjects. Given how carefully the rest of the boundary is written down, that belongs in the PROPERTY_TESTS.md boundary paragraph.

erskingardner (commented):
**Request changes** (posted as comment — GitHub does not allow requesting changes on your own PR).

Independent re-review at `e01bf9e`. The premise is sound: making the `accepted` checkpoint portable across the engine's staged transport and the app adapters' already-published command surface is the right way to run one canonical IR unchanged, and the evidence-boundary prose (snapshot-pinned campaign, post-campaign production deltas scoped as delta work) matches this repo's disclosure discipline. CI is green. I re-derived the concerns from the earlier review against the head tree and confirm all three; one new finding inline.

**Blocking**

1. **Capability contract violation on `AppRuntimeHarness`.** Declaring `SubjectCapability::OutboundPublication` (`app_runtime.rs:590`) without overriding `poll_outbound`/`acknowledge_outbound` (both still resolve to the trait defaults returning `unsupported(OutboundPublication)`, `subject.rs:377`) flips every non-happy acknowledgement path from a clean up-front preflight refusal into a mid-run failure. Four concrete paths at this head: unlabelled `accept_all_outbound` drains from the stateful generator (`stateful_generator.rs:585`); `AwaitQuiescence` with `outbound: AcceptAll`, which now passes preflight and dies in the driver (`quiescence.rs:196`); a labelled acknowledgement whose label was never recorded; and the rollback-faults family's `ReachedNoEndpoint` on the recorded `"update"` label hitting the new runner guard. Implementing the two harness methods (empty poll, acknowledge that accepts only a recorded publication with `Accepted` and refuses otherwise as a classified `ExpectedRefusal`) fixes the first three and gives the fourth a truthful preflight.

2. **The fail-closed branch has no test.** Confirmed: none of the five tests in `tests/app_runtime_adapter.rs` touch `AcknowledgeOutbound`, and both new scenarios acknowledge only recorded labels with `Accepted`. The three non-happy arms on this adapter — rollback of an accepted label, unknown label, unlabelled drain — are unexercised, and a focused test there is what would have surfaced finding 1.

3. **Docs assert a class property only one adapter implements.** `process_orchestrator.rs:674` answers every `AcknowledgeOutbound` with `AlreadyPublished` unconditionally, discarding both label and outcome. So "they reject a later `reached_no_endpoint` outcome" (README) and "explicit accepted-publication checkpoints retain the same meaning" (PROPERTY_TESTS.md) hold only for the in-process app-runtime adapter; on the process adapter a typo'd publication label also passes silently while erroring on the engine. Either teach the orchestrator arm the same rule (it is a few lines) or narrow the prose to the adapter that implements it.

**Non-blocking**: one inline note on the failure classification of the new guard. The earlier review's nits (`scenario_publication_already_accepted` never fails so `-> bool` suffices; the bare-`bool` scenario builder flag) are worth taking as-is.

crates/cgka-conformance-simulator/src/scenario.rs
```diff
@@ -905,29 +905,47 @@ async fn execute_scenario_step(
...
+                    return Err(err(
```
erskingardner: This guard is the right fail-closed behavior, but it reports through the generic `err()` helper, so the failure surfaces as `kind: "scenario_error"`, `category: Environment` (`scenario.rs:1487`). When the rollback-faults family runs against an auto-publishing subject, campaign triage will bucket this deliberate semantics guard as an *environment failure* — exactly the misclassification the campaign taxonomy (product defect / protocol ambiguity / environment failure / expected refusal) exists to separate. Worth a distinct `kind` (e.g. `publication_rollback_rejected`) or a classified category so report rollups don't read it as an infra flake.

erskingardner (commented):
## Verdict

**Recommend merge** at `41ec2f0` (posted as comment — GitHub does not allow approving your own PR).

Premise is sound: portable `accepted` checkpoints across staged-engine vs auto-publishing app/process surfaces are the right way to run one canonical IR unchanged, and the assurance boundary (snapshot-pinned 1,170-case campaign; #1329/#1360/#1365 as current-`master` delta work; no overclaim of exact MLS/decryptability/disposition/container/VM evidence) matches this repo’s disclosure discipline. CI is green.

## Prior blocking findings

Re-checked against this head — all three from the earlier review are addressed:

1. `AppRuntimeHarness` now implements empty `poll_outbound` and fail-closed `acknowledge_outbound`, so declaring `OutboundPublication` matches the capability contract.
2. `app_runtime_publication_acknowledgements_fail_closed_without_mid_run_capability_errors` exercises rollback / unknown-label / unlabelled-drain, classified as `ExpectedRefusal`.
3. Process `AcknowledgeOutbound` goes through `validate_auto_published_acknowledgement` (with a focused unit test), so README / PROPERTY_TESTS “same semantics” claims hold for both auto-publishing adapters.

PROPERTY_TESTS now also records that labelled `Accepted` on app/process is command-success evidence, not independent transport-delivery proof.

## Non-blocking

- CodeRabbit’s note on recording the create publication before `apply_admin_set` / after `CreateGroup` but before the SelectGroup fanout is directionally right for torn-write hygiene, but a failed create step already aborts the scenario, so it does not undermine this PR’s assurance claim. Fine to take later or leave.
- The scenario-runner `already_accepted` short-circuit means IR execution never calls `AppRuntimeHarness::acknowledge_outbound`; the new adapter test covers the runner path (which is what matters). The harness method remains the correct capability-contract mirror for direct/quiescence callers.

Stale inline threads from the pre-fixup review can be resolved.

PR #1378, opened 2026-08-11
Publish immutable MarmotKit SwiftPM artifacts

erskingardner:
## Summary

- publish a root-level SwiftPM-compatible MarmotKit XCFramework ZIP alongside the synchronized generated Swift source and complete provenance bundle
- publish ordinary SHA-256 and SwiftPM checksum assets, with an expanded manifest recording the exact source SHA, Cargo.lock hash, toolchain, features, targets, deployment target, and effective Rust release profile
- make formal releases and exact-master snapshots immutable by refusing existing releases/tags and removing all `--clobber` behavior
- validate XCFramework layout, arm64 slices, iOS 18.0 deployment metadata, generated Swift compilation, and simulator/device linking both before and after publication
- document formal release and snapshot URLs plus the consumer synchronization and verification rules

## Why

App repositories need to consume MarmotKit as an immutable remote SwiftPM binary target without committing the expanded XCFramework and static libraries. A release asset must never be replaced after consumers pin its checksum, and untagged TestFlight builds need an exact-SHA snapshot identity rather than masquerading as the existing 0.9.11 release.

## Published snapshot

The implementation was exercised against the then-current `origin/master` SHA `d34629e70689f6c1f0cbf11d531b05f20fb0b282`:

https://github.com/marmot-protocol/mdk/releases/tag/marmotkit-snapshot-d34629e70689f6c1f0cbf11d531b05f20fb0b282

The binary ZIP is 70,403,759 bytes. Its ordinary SHA-256 and SwiftPM checksum are both:

`686b59fd747a778954622c069e56a7626645a5ff0491cfbe87767c3f30b817cb`

## Validation

- `bash -n`
- `shellcheck -e SC1091`
- `actionlint .github/workflows/bindings.yaml`
- `git diff --check`
- `just fast-ci`
- local binary-only archive layout and XCFramework slice validation
- local SwiftPM source compilation and arm64 simulator/device dynamic linking
- published-URL `swift package resolve` download and checksum verification
- published-URL arm64 simulator and generic iOS device compilation/linking

A runtime construction/start/shutdown smoke was not run because the minimal package has no host app, configured runtime root, or simulator process environment.

## Summary by CodeRabbit

* **New Features**
  * Added immutable iOS TestFlight snapshots tied to a specific source revision.
  * Added Swift Package Manager distribution with versioned artifacts, checksums, and provenance metadata.
  * Added support for downloading and integrating MarmotKit binary releases through SwiftPM.

* **Documentation**
  * Documented release and snapshot URLs, package configuration, verification steps, naming conventions, and archive contents.

* **Quality Improvements**
  * Added automated validation for XCFramework structure, supported iOS slices, package integrity, and local or remote SwiftPM consumption.


.github/workflows/bindings.yaml
```diff
@@ -254,42 +335,110 @@ jobs:
...
-            gh release create "$tag" "${assets[@]}" \
-              --repo "$GITHUB_REPOSITORY" \
-              --title "v$version - MarmotKit" \
-              --notes-file "$notes" \
-              --latest=false
+            title="$version - MarmotKit"
+            release_flags=(--prerelease --latest=false --target "$SOURCE_SHA")
           fi
+
+          gh release create "$tag" "${assets[@]}" \
```
erskingardner: [P1] Publish through GitHub immutable releases. The repository immutable-releases endpoint currently reports `enabled: false`, and the snapshot already published by this PR reports `immutable: false`. These existence checks only prevent this workflow from replacing an asset; a maintainer/API call can still edit or delete the release/tag/assets, which invalidates the checksum-pinned SwiftPM URL. Please enable repository release immutability, assert that setting in this job, and create the release as a draft, upload/verify all assets, then publish it. The draft flow also prevents a failed multi-asset upload from leaving a partial public release that this retry guard refuses to repair.

.github/workflows/bindings.yaml
```diff
@@ -4,46 +4,130 @@ on:
...
     runs-on: macos-latest
+    needs: metadata
     timeout-minutes: 60
 
     steps:
       - name: Check out repository
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
         with:
           persist-credentials: false
+          ref: ${{ github.sha }}
```
erskingardner, line 116: [P1] Bind snapshot publication to canonical builder code. A manual dispatch can select a non-default workflow ref, so this checkout and the packaging/validation scripts come from `github.sha` while the binary source comes from the separately validated `source_sha`; the manifest records only the latter. A branch can therefore claim the canonical `marmotkit-snapshot-<master-sha>` tag while building it with unmerged packaging code, and true immutable releases would make that claim permanent. Reject snapshot dispatches unless `github.ref == refs/heads/master` (or otherwise pin the builder checkout to the default branch), and record the builder/workflow SHA in provenance.

crates/marmot-uniffi/validate-ios-artifact.sh
```diff
@@ -0,0 +1,111 @@
...
+    exit 1
+  }
+
+  object_dir="$TEMP_DIR/objects-$index"
+  mkdir -p "$object_dir"
+  (
+    cd "$object_dir"
+    xcrun ar -x "$library"
+  )
+  primary_object="$(find "$object_dir" -type f -name 'marmot_uniffi.marmot_uniffi*.o' -print -quit)"
```
erskingardner: [P1] Validate deployment metadata for every archive member, not only the first Rust object. The static library also contains independently compiled C/SQLCipher/OpenSSL objects; one of those can target a newer iOS version while this sampled Rust object still reports 18.0, so the artifact passes here but emits linker warnings or fails on supported devices. Please inspect every member that carries `LC_BUILD_VERSION` and require the expected platform and minimum OS throughout each slice.

erskingardner (commented):
Request changes. The overall packaging shape is right: the snapshot is tied to an exact master commit, the release profile preserves `opt-level=3`, the SwiftPM ZIP/checksum/source assets are internally coherent, and local plus published-URL consumer builds are covered.

There are three release-integrity blockers: GitHub immutability is not enabled or used, manual dispatch can publish a canonical source tag using noncanonical builder scripts without recording that builder SHA, and deployment-target validation samples only one object from each static archive. These need tightening before the artifacts can carry the immutable/provenance/support claims made here.

The already-published snapshot is useful evidence, but it is currently mutable and was not produced by a recorded run of this new workflow.

PR #1379, opened 2026-08-11
Keep unknown-group drops out of the ingest seen cache

dannym-arx:
Stacked on #1367.

A marmot that knocked before the burrow existed should not be turned away when it knocks again after the door goes up. Commits that reach a client before its delayed Welcome cannot resolve the transport route, and the unknown-route posture (#740) deliberately drops them without a durable record. The ingest loop still inserted each dropped message id into the in-memory seen cache. When the client later joined and the relay redelivered the same events (the designed recovery path), the copies were classified as duplicates of input that left no record. The late joiner stayed epochs behind the group forever while reporting no pending work.

Changes:

- The unknown-group ignore in `ingest_group_message` is now marked retryable-unpersisted, so the seen cache skips it. Redelivered events after the join resolve the freshly indexed route and apply normally. The #740 flood posture is unchanged: unknown-route traffic still consumes no storage.
- New harness regression `late_welcome_join_catches_up_via_redelivered_commits`: Dave's Welcome is delayed while the group advances two epochs, and redelivery after his join must catch him up to the tip.
- New harness regression `commit_missing_proposal_defers_and_applies_after_proposal_backfill`, pinning behavior that was already correct: a commit whose referenced proposal never arrived stays convergence-deferred, and a later proposal backfill completes it at the recipient.
- New portable vector `late-welcome-backfill/v1` (strict-oracle green), registered in the manifest and `SCENARIOS.md`.

Verification: both regressions pass; all 30 portable vectors pass the strict oracle; the cgka-engine suite passes under the CI invocation; `just fast-ci` is green. The stack was rebased onto master `d34629e7` and re-verified against the adjacent leave-persistence and replacement-Welcome retirement changes.

### DIFF
--- a/crates/cgka-conformance-simulator/SCENARIOS.md
+++ b/crates/cgka-conformance-simulator/SCENARIOS.md
@@ -293,6 +293,19 @@ These are the scenarios another implementation should be able to load from JSON 
   replay from retained histories) stays in the `adversarial-reliability` generated family per `retained_relay.rs`
   doctrine; this vector models mailbox catch-up.
 
+### `late-welcome-backfill/v1`
+
+- File: `vectors/late-welcome-backfill.v1.json`
+- Setup: Alice creates a group with Bob, invites Dave, and the group advances two epochs while Dave's Welcome is
+  withheld. The commits reach Dave before he can resolve the group. After the Welcome arrives and Dave joins at the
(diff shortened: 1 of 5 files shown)


Datawav (commented):
## Automated review

No actionable findings.

Datawav (commented):
## Automated review

erskingardner (dismissed):

PR #1377, opened 2026-08-11
test(cgka-engine): cover pinned rewind horizon

agent-p1p:
Fixes #1219

Supersedes #1371, which is permanently ineligible for merge after code-owned red CI in an unrelated load-sensitive Nostr-adapter test. The flake is tracked separately in #1375.

## Summary
- remove the non-pinned one-commit policy override from the normal-build fork test
- advance the local branch beyond `V1_MAX_REWIND_COMMITS` before ingesting the stale epoch-1 commit
- keep the terminal-stale and no-`ForkRecovered` assertions tied to the computed pinned-horizon epoch
- document the regression fix under `Unreleased`

## Verification
- `cargo test -p cgka-engine --test fork_detection --locked` — 16 passed
- `just test-convergence-policy-pin` — 5 passed
- `cargo test -p cgka-engine --features test-policy-overrides --test fork_detection stale_commit_outside_rewind_horizon_is_not_treated_as_recoverable_fork --locked` — 1 passed
- `cargo fmt --all -- --check`
- `git diff --check origin/master...HEAD`
- no `Cargo.toml` or `Cargo.lock` changes

## Sensitive paths
- `crates/cgka-engine/tests/fork_detection.rs` — test-only fork recovery / rewind-horizon coverage

No production convergence constants or policy guards changed.

## Summary by CodeRabbit

- **Tests**
  - Improved validation of fork detection across the full supported commit rewind horizon.
  - Confirmed that commits arriving beyond the recovery window are correctly identified as stale.

- **Documentation**
  - Updated the unreleased changelog to reflect expanded fork-detection coverage in standard builds.

PR #1383, opened 2026-08-11
Add fail-closed focused convergence regression gate

erskingardner:
## Summary

- add a named, fail-closed CI gate for the five convergence regressions introduced after the reviewed 1,170-case campaign
- run each exact Nextest selector independently so a renamed or deleted test fails with `no tests to run`
- reconcile the Milestone 6 tracker with merged PRs #1140, #1372, and #1377 while leaving current-`master` evidence open until the squash-merge commit passes the named gate

## Assurance boundary

This PR changes CI/testing configuration and documentation only. It does not modify production convergence-engine, simulator, campaign-runner, or app behavior. It does not repeat the historical generated matrix or claim that the four-party cross-route vector has run through app, process, container, or VM adapters.

The successful `Run focused convergence regressions` CI step on the squash-merge commit will be the durable evidence for current `master`; this PR deliberately does not pin machine-local paths, timings, logs, or unreproducible artifact digests.

## Verification

- `just focused-convergence-regressions`: five independently selected tests passed
- deliberately missing exact selector: Nextest exited 4 with `error: no tests to run`
- `just fast-ci`

## Summary by CodeRabbit

- **Tests**
  - Added a focused regression test suite covering five convergence scenarios across the engine and application.
  - Tests now fail clearly when expected regression cases are missing.

- **CI**
  - Integrated the focused convergence regression checks into the conformance pipeline.

- **Documentation**
  - Updated reliability tracking with current milestone evidence, regression results, boundary-hardening progress, and remaining validation work.
  - Clarified failure-corpus validation requirements for missing or corrupt data.


crates/convergence-campaign-runner/src/regression_campaign.rs
```diff
@@ -0,0 +1,803 @@
...
+        }
+        if !self.features.is_empty() {
+            args.push("--features".into());
+            args.push(self.features.join(","));
+        }
+        args.extend([
+            self.test_name.clone(),
+            "--".into(),
+            "--exact".into(),
+            "--nocapture".into(),
```
erskingardner: **Blocking: `--exact` cannot distinguish a passing regression from a missing one.**

`cargo test <filter> -- --exact` exits 0 when the filter matches nothing. Verified:

```
$ cargo test --lib this_test_does_not_exist -- --exact --nocapture
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out
$ echo $?
0
```

Combined with the `passed` predicate below (line 324), any rename, module move, or deletion of one of the five pinned tests silently produces `"passed": true` for a case that never executed. `case_id`/`assurance_claim` have no compile-time or runtime linkage to the test — `test_name` is a hand-copied string. `tests::explicit_catch_up_arms_and_replays_without_later_traffic` in particular depends on `mod tests;` staying at `marmot-app`'s crate root; move `tests.rs` under any submodule and that case degrades to a no-op that still reports green.

`cargo nextest run -E 'test(=<name>)'` exits 4 on an empty match (verified on 0.9.104) and is already the convention across this Justfile, including exact-match `-E` at `Justfile:259`. Failing that, assert the captured stdout contains `1 passed` before setting `passed`.

crates/convergence-campaign-runner/src/regression_campaign.rs
```diff
@@ -0,0 +1,803 @@
...
+        #[cfg(unix)]
+        {
+            use std::os::unix::process::CommandExt as _;
+            command.as_std_mut().process_group(0);
+        }
+        let (stdout, stderr, exit_code, signal, timed_out, launch_error) =
+            execute_case(command, case_timeout, input.case_timeout_ms).await;
+        let stdout_artifact = write_case_artifact(output_dir, case, "stdout", &stdout)?;
+        let stderr_artifact = write_case_artifact(output_dir, case, "stderr", &stderr)?;
+        let passed =
```
erskingardner: The pass predicate is exit-status-only, and `validate_artifacts` (line 182) recomputes the log digests without ever parsing them. So the retained logs would show `0 passed; 1 filtered out` while `campaign-report.v1.json` records `passed: true`, and the report is what the tracker cites. Whatever selector mechanism you land on, the pass condition needs to include "the named test actually ran." See the note at line 175.

crates/convergence-campaign-runner/src/regression_campaign.rs
```diff
@@ -0,0 +1,803 @@
...
+    fs_private::write_private(&output_dir.join(&path), bytes)
+        .map_err(|error| RunnerError::environment("regression_campaign_artifact_write", error))?;
+    Ok(RegressionArtifactV1 {
+        path,
+        sha256: sha256(bytes),
+        bytes: u64::try_from(bytes.len()).unwrap_or(u64::MAX),
+    })
+}
+
+fn validate_artifact(base_dir: &Path, artifact: &RegressionArtifactV1) -> Result<(), RunnerError> {
```
erskingardner: **Duplication:** `validate_artifact`, `read_artifact` (416), and `is_lower_sha256` (445) reimplement `evidence.rs`'s `ConvergenceEvidenceBundleV1::validate_artifacts` almost line for line — same empty/absolute/`ParentDir`/`RootDir`/`Prefix` rejection, same `canonicalize` + `starts_with` containment check, same 64-char lowercase-hex digest check, same recompute-and-compare. That module also streams through a 64 KiB `BufReader` instead of reading whole files into memory.

Please reuse `EvidenceArtifactV1` and the existing validation rather than growing a parallel copy that can drift.

crates/convergence-campaign-runner/src/regression_campaign.rs
```diff
@@ -0,0 +1,803 @@
...
+    pub signal: Option<i32>,
+    pub timed_out: bool,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub launch_error: Option<String>,
+    pub wall_ms: u64,
+    pub passed: bool,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct RegressionArtifactV1 {
```
erskingardner: `RegressionArtifactV1` is `EvidenceArtifactV1` (`evidence.rs:39`) with `kind: String` swapped for `bytes: u64`. Adding `bytes` to the existing type, or wrapping it, avoids a second artifact vocabulary in one crate.

crates/convergence-campaign-runner/src/bin/cgka-regression-campaign.rs
```diff
@@ -0,0 +1,124 @@
...
+            println!(
+                "Focused convergence report: {}",
+                out.join(REGRESSION_CAMPAIGN_REPORT_FILE).display()
+            );
+            Ok(report.passed)
+        }
+        Commands::Check { report } => {
+            let base_dir = report
+                .parent()
+                .filter(|parent| !parent.as_os_str().is_empty())
```
erskingardner: This is `evidence_bundle_base_dir` (`src/evidence.rs:49`), re-inlined. It is public, re-exported through `pub use evidence::*`, and already called by `src/main.rs:200` — including the same "a bare filename lives in the current directory even though `Path::parent` represents that parent as an empty path" rationale that this code rediscovers. Call the helper.

crates/convergence-campaign-runner/src/bin/cgka-regression-campaign.rs
```diff
@@ -0,0 +1,124 @@
...
+
+#[derive(Debug, Subcommand)]
+enum Commands {
+    /// Execute the focused catalog and create immutable private evidence.
+    Run {
+        /// Owner-only directory to create for the immutable campaign artifacts.
+        #[arg(long)]
+        out: PathBuf,
+        /// Per-regression timeout in seconds.
+        #[arg(long, default_value_t = 300)]
```
erskingardner: _Non-blocking._ The 300 s budget wraps the whole `cargo test` invocation, compilation included — building `cgka-engine`'s `fork_detection` test target and `marmot-app`'s lib test binary is inside the timed window. On a cold or partially cold `target/`, a case can trip the timeout and be retained as `timed_out: true, passed: false`, which in the preserved evidence is indistinguishable from a genuine convergence regression.

Consider a `cargo test --no-run` (or `nextest list`) warm-up outside the timed window so the recorded wall time and timeout reflect execution rather than build latency.

crates/convergence-campaign-runner/src/bin/cgka-regression-campaign.rs
```diff
@@ -0,0 +1,124 @@
+use std::path::{Path, PathBuf};
```
erskingardner: _Non-blocking._ The crate's `Cargo.toml` declares its other three binaries explicitly (`cgka-distributed-campaign`, `cgka-conformance-node`, `cgka-conformance-relay`), including the two that also live under `src/bin/`. This one relies on autodiscovery instead. It compiles, but adding the matching `[[bin]]` entry keeps the manifest the single place to see what the crate ships.

crates/convergence-campaign-runner/src/regression_campaign.rs
```diff
@@ -0,0 +1,803 @@
...
+        && value
+            .bytes()
+            .all(|byte| byte.is_ascii_hexdigit() && !byte.is_ascii_uppercase())
+}
+
+fn sha256(bytes: &[u8]) -> String {
+    hex::encode(Sha256::digest(bytes))
+}
+
+async fn execute_case(
```
erskingardner: _Non-blocking._ A six-element unnamed tuple `(Vec<u8>, Vec<u8>, Option<i32>, Option<i32>, bool, Option<String>)` — two of which are same-typed `Option<i32>` — is easy to transpose at the call site. A small `CaseExecution` struct would document itself and the destructuring at line 330 would read directly.

erskingardner (commented):
> **Review verdict: changes requested.** GitHub will not accept a formal `REQUEST_CHANGES` review on your own PR, so this is posted as a comment review.

## Summary

The five regressions this campaign pins are all ungated `#[test]` / `#[tokio::test]` cases that already run in `cargo nextest run --workspace` on every PR and every `master` commit. The marginal assurance over existing CI is therefore the retained artifacts alone — and as built, those artifacts are local, uncommitted, and digest-pinned to values nobody can reproduce or check. Meanwhile the pass predicate cannot distinguish "the regression passed" from "the regression no longer exists."

That is ~930 new lines of bespoke runner to issue five `cargo test` invocations. Details inline; the three that block:

**1. The pass criterion is unsound.** `passed` is `exit_code == Some(0)`, which is satisfied when the test filter matches nothing. Verified locally: `cargo test --lib no_such_test -- --exact` prints `0 passed; 1 filtered out` and exits **0**. Rename, move, or delete any of the five tests and the campaign records `passed: true` for a case that never ran. `validate_artifacts` recomputes log digests but never parses the log, so the JSON report — the only thing anyone reads — asserts a regression held when nothing executed. For a tool whose entire product is the claim "these five named regressions passed," that is the one failure mode it must not have.

`cargo nextest run -E 'test(=<name>)'` exits **4** on an empty match (verified, 0.9.104) and is already this repo's convention, including exact-match `-E` syntax at `Justfile:259`.

**2. The recorded evidence is not verifiable.** `convergence-reliability-plan.md` permanently pins report SHA-256 `c254b99c…`, but that report embeds `wall_ms` wall-clock timings and `command_argv[0]` (a machine-local cargo path), and its stdout artifacts embed cargo build/run timings — the digest is unreproducible by construction, and the file exists only in the author's `target/`. The pinned source `fd1d9bc0…` is a PR-branch commit that ceases to exist once this squash-merges (`master` is entirely single-parent `(#NNNN)` squashes; the previously pinned `7b048060…` is likewise not an ancestor of `master`). A digest of a file nobody has, at a commit that will not exist, does not function as evidence.

**3. Most of this module already exists in the crate.** `evidence.rs` provides `EvidenceArtifactV1`, a `validate_artifacts` with the identical canonicalize / `starts_with` / parent-traversal / lowercase-hex / digest-recompute logic, and a public `evidence_bundle_base_dir` helper that `main.rs:200` already calls — which the new CLI re-inlines verbatim.

### Suggested smaller shape

A Justfile recipe with five `cargo nextest run --locked -E 'test(=…)'` lines, plus a tracker entry citing the CI run that executed them. That fails closed when a test disappears, adds no crate module, and yields evidence any reviewer can open. If a retained-artifact bundle is genuinely required, please build it on `evidence.rs` and drop the unreproducible digests from the permanent doc.

Non-blocking notes are marked inline. CI is green and the code itself is clean and careful — the concerns are about what the artifact claims, not its craftsmanship.

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -1218,15 +1228,27 @@ residual gap.
...
+   remain open. The decision-route inventory is machine checked; keep the remaining assurance artifacts independent
+   of #1293.
+6. [x] Run a focused current-`master` delta campaign for the production convergence changes that landed after the
+   reviewed 1,170-case snapshot. Preserve exact inputs, source revision, strict reports, and artifact digests while
+   covering #1329 missing-anchor halt and authenticated Welcome repair, #1360 atomic self-removal persistence across
+   injected write failures and restart, and #1365 armed full-history backfill after catch-up. Include #1377's pinned
+   rewind-horizon regression as the normal-build policy-boundary sentinel. This is a small targeted matrix, not a
+   requirement to repeat the entire historical campaign. `focused-convergence-regressions/v1` now composes the five
+   exact named engine/app tests without duplicating their internal fault injection as Scenario IR. Its clean-revision
+   run passed 5/5 at `fd1d9bc0914cdd0e9ad58352c05fc1b7aceed767`; the privately written input manifest has SHA-256
```
erskingardner: **Blocking: these pins do not support the claim they are recorded for.**

- **The report digest is unreproducible by construction.** `campaign-report.v1.json` embeds `wall_ms` per case and `command_argv[0]` (the local cargo path), and the stdout artifacts it binds contain cargo build/run timings. No second run on any machine can reproduce `c254b99c…`, so the digest cannot be used to check anything.
- **The artifacts are not retained anywhere reviewable.** They are owner-only files written under `target/` on one machine and never committed. Recording the digest of a file nobody else has, of content that cannot be regenerated, is ceremony rather than evidence.
- **The source revision disappears on merge.** `fd1d9bc0…` is a commit on `codex/focused-convergence-regressions`. This repo squash-merges — every commit on `master` is single-parent and titled `(#NNNN)` — so the SHA this evidence is pinned to will not exist in any branch history after merge. (The `7b048060…` pinned at line 1409 is already unreachable from `master` for the same reason, so this compounds an existing problem.)

Additionally, all five cases are ungated `#[test]`/`#[tokio::test]` functions that run in `cargo nextest run --workspace` on every PR and every `master` commit. A named CI run on the merged `master` commit is strictly stronger evidence than a one-off local run pinned to a vanishing SHA: it is reproducible, continuously re-executed, and its logs are retained and readable by anyone.

Suggest citing the merge commit and CI run for item 6 and dropping the two digests.

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -1383,7 +1406,9 @@ incorrect result.
...
+| 2026-08-11 | Focused post-campaign production-delta evidence | Added a clean-source local regression campaign that writes its exact test-selector input before execution, owns and reaps the full case process group, preserves partial timeout output, refuses overwrite, and binds private per-case logs by byte count and SHA-256. The first run passed the missing-anchor halt, verified Welcome repair, three-write atomic self-removal rollback/restart, armed post-catch-up backfill, and pinned rewind-horizon cases with no timeout, signal, or launch failure. This is targeted current-source regression evidence, not a rerun of the historical generated matrix or distributed cross-route evidence. | `focused-convergence-regressions/v1`; source `fd1d9bc0914cdd0e9ad58352c05fc1b7aceed767`; 5/5 passed; input `48982c7589e4ec9c6f211382f9ed417095e083c0d4b5ba055cfd88c4541f98d6`; report `c254b99cab4722e06aa46a6ba08d6fa2c222b62a68e6820ab77e7b860d8eb58c` |
```
erskingardner: Same issue as the item 6 entry above: `source fd1d9bc0…` will not exist after squash-merge, and the `input`/`report` digests bind files that live only in one local `target/` directory and cannot be regenerated. Please cite the merged commit and its CI run instead.

Justfile
```diff
@@ -329,6 +329,11 @@ convergence-failure-corpus:
     cargo nextest run -p convergence-campaign-runner --test failure_corpus --locked
     cargo nextest run -p cgka-conformance-simulator --test semantic_reduction --locked
 
+# Run the small post-fix regression matrix against a clean source revision and
+# retain its exact test selectors, private logs, source identity, and digests.
+focused-convergence-regressions out:
```
erskingardner: Every other test recipe in this file drives `cargo nextest run`, and line 259 already uses exact-match filter expressions (`-E 'test(=real_local_relays_deliver_cli_messages_over_sdk_path)'`). Five such lines here would express the whole campaign, and unlike `cargo test … --exact` they fail closed when a pinned test is renamed or removed (nextest exits 4 with `error: no tests to run`). That would also make the new binary and module unnecessary.

Justfile
```diff
@@ -329,6 +329,16 @@ convergence-failure-corpus:
     cargo nextest run -p convergence-campaign-runner --test failure_corpus --locked
     cargo nextest run -p cgka-conformance-simulator --test semantic_reduction --locked
 
+# Run each post-fix regression through an independent exact selector. Nextest
+# exits nonzero when a selector matches no tests, so renames and deletions fail
```
erskingardner, line 333: _Non-blocking._ The fail-closed property this comment asserts is a nextest **default**, not something the recipe pins. `--no-tests=fail` became the default in nextest 0.9.85; before that an empty match warned and exited 0.

CI is safe today — `taiki-e/install-action` installs latest (0.9.136 on this run) — but nothing in the repo enforces a floor: `.config/nextest.toml` has no `[nextest.version] required`, and `scripts/check_campaign_toolchain.sh` gates only the Rust channel and the campaign image digest. A developer on an older nextest who runs `just focused-convergence-regressions` gets exactly the silent zero-test pass this comment says the recipe prevents.

Adding `--no-tests=fail` to each of the five lines (or a `required` version in the nextest config) makes the guarantee explicit and self-enforcing instead of inherited from whatever nextest happens to be installed.

erskingardner (commented):
> **Review verdict: recommend merge.** GitHub will not accept a formal `APPROVE` review on your own PR, so this is posted as a comment review.

## Re-review of review 4906590128

All three blocking findings are resolved, and resolved by deletion rather than by patching around them.

**1. Unsound pass criterion — fixed.** `cargo test … --exact` and the bespoke `passed = exit_code == Some(0)` predicate are gone. Each regression now runs through its own `cargo nextest run … -E 'test(=…)'`. Verified on the CI run for `7985236` (job *Simulator and vectors*, step *Run focused convergence regressions*, nextest 0.9.136): five separate `1 test run: 1 passed` summaries, so every selector matched exactly one real test. Keeping the three `fork_detection` selectors as separate invocations instead of folding them into one `test(=a) | test(=b) | test(=c)` filterset is the right call — a union filterset would still exit 0 with one member missing, which would reintroduce the original defect.

**2. Unverifiable evidence — fixed.** The unreproducible report/input SHA-256 pins and the `fd1d9bc0…` branch-commit pin are gone from the plan. Item 6 and the new evidence row now cite `just focused-convergence-regressions` and the named CI step, and item 6 explicitly stays open until that step passes on the squash-merge commit. The one remaining commit pin (`7b048060…`, item 4 and the #1357 row) is pre-existing and out of scope for this PR.

**3. Duplication with `evidence.rs` — moot.** `regression_campaign.rs` and `bin/cgka-regression-campaign.rs` no longer exist in the diff; net change is +51/−15 across three files, so the ~930-line parallel artifact vocabulary never reaches `master`. The non-blocking notes that hung off it (build time inside the 300 s window, missing `[[bin]]` entry, six-element tuple) fall away with it.

### Accuracy spot checks on the doc changes

- `leave_persistence_failure_rolls_back_every_transactional_write` really does loop `fail_on_write` 1..=3 and then reopen from the same storage to prove a same-epoch retry, so "three-write atomic self-removal rollback/restart" is not an overclaim.
- The other four selectors map to the claimed sources (#1329 missing-anchor halt and verified Welcome repair, #1365 armed post-catch-up backfill, #1377 pinned rewind horizon), and #1140, #1372, and #1377 are all merged.
- Item renumbering is consistent: the only in-text cross-reference ("item 6 gives those paths a continuously executed regression owner") points at the new item, and no stale "item 7" reference remains.

### Remaining trade-off, accepted

All five tests already execute in the partitioned workspace job, so the marginal value here is named failure attribution plus a citable CI step, at roughly 60 s of extra job time. That is a fair trade at ten lines of Justfile; it was not at 930 lines of runner.

One non-blocking note inline.

PR #1384, opened 2026-08-11
fix(timeline): converge group history ordering

### LINKED ISSUE #1172: Timeline: make group-system row ordering stable across delayed catch-up
## Problem

The durable timeline can order the same authenticated group history differently on different devices.

Received application messages use the authenticated inner Marmot event timestamp as `timeline_at`, while locally synthesized kind-1210 group-system rows use the receiving device's `source_received_at`. During delayed catch-up, commit rows are therefore timestamped when the offline device reconnects rather than when those state transitions occurred relative to the application messages.

An online client may render:

1. epoch-N system change
2. application message sent in epoch N
3. epoch-(N+1) system change

while an offline client processing the same history later can render:

1. application message
2. epoch-N system change
3. epoch-(N+1) system change

This can make messages appear missing from a commit-heavy section even though they were decrypted and stored elsewhere in the timeline.

## Current path

- `crates/marmot-app/src/groups.rs`: received messages set `recorded_at` from the authenticated inner event's `created_at`.
- `crates/marmot-app/src/client/sync.rs`: inbound group-system projection is called with `source_received_at`.
- `crates/marmot-app/src/client/projection.rs`: synthesized system rows store that value as `recorded_at`.
- `crates/storage-sqlite/src/timeline.rs`: the durable display order is `(timeline_at, message_id_hex)`.

## Desired invariant

Given the same accepted MLS history, clients should derive the same durable relative order for group-state transitions and application messages, independent of when each client received or replayed the transport history.

## Questions to investigate

- Should durable ordering incorporate `source_epoch`, placing the system row that establishes epoch N before application messages authenticated in epoch N?
- Is a separate canonical logical-order key needed instead of overloading wall-clock `timeline_at`?
- Can any transport timestamp safely participate, given that outer timestamps are not authoritative for application-event chronology?
- How should existing projected rows be migrated or reprojected?
- Do read markers, pagination boundaries, or unread calculations inherit device-dependent behavior from this ordering?

## Suggested regression

Create two clients with identical group history. Keep one online while interleaving several commits and application messages; keep the second offline and deliver the full history later. Assert that both materialized timelines produce the same ordered sequence of message/system-row IDs and preserve the expected epoch boundaries.

Related to #1171, but distinct from epoch-secret expiry: in the observed catch-up, the affected device successfully decrypted the messages and the discrepancy occurred at timeline projection/order.

agent-p1p:
## Summary

- derive per-group durable timeline order from authenticated MLS `source_epoch` boundaries instead of local observation time
- place synthesized state-change rows before application messages in the epoch they establish, with deterministic row-ID ordering for multiple same-epoch state changes
- use the same canonical order for pagination, runtime windows, chat previews, unread counts, and read-marker advancement while preserving wall-clock order for global queries and retention scans
- add migration 44 for the canonical SQLite index and chat-list projection refresh

## Test plan

- `RUSTFLAGS='-D warnings' cargo test -p storage-sqlite --locked` — 336 passed
- `RUSTFLAGS='-D warnings' cargo test -p marmot-app --lib --locked` — 608 passed
- `CARGO_INCREMENTAL=0 just fast-ci` — passed
- delayed/catch-up ordering regression demonstrated RED before deterministic same-epoch state-row ordering and GREEN after the fix

## Release hygiene

- updated `crates/cli/CHANGELOG.md` under Unreleased
- no version or conformance-fixture bump

Fixes #1172

## Summary by CodeRabbit

* **Bug Fixes**
  * Timeline messages now appear in deterministic authenticated-history order, including across pagination, live updates, read markers, and device catch-up.
  * Chat previews and unread indicators consistently reflect the latest valid message.
  * Expired timeline cursors now recover by refreshing from the latest available messages.
  * Failed or pending sends no longer incorrectly replace delivered history or previews.
  * Retention scans continue using chronological ordering where appropriate.

* **Documentation**
  * Clarified timeline ordering, pagination, cursor behavior, and timestamp semantics.
  * Updated command-line help for timeline cursor options.

### DIFF
--- a/crates/marmot-app/src/runtime/event_routing.rs
+++ b/crates/marmot-app/src/runtime/event_routing.rs
@@ -56,10 +56,14 @@ pub(crate) fn projection_update_matches_query(
 }
 
 pub(crate) fn timeline_query_can_apply_projection_delta(query: &TimelineMessageQuery) -> bool {
-    query
-        .search
-        .as_ref()
-        .is_none_or(|search| search.trim().is_empty())
+    // Canonical source epochs are comparable only within one group. A global
+    // subscription keeps the store's wall-clock ordering by refreshing instead
+    // of locally merging per-group deltas.
+    query.group_id_hex.is_some()
+        && query
+            .search
+            .as_ref()
+            .is_none_or(|search| search.trim().is_empty())
         && query.pagination.before.is_none()
         && query.pagination.before_message_id.is_none()
         && query.pagination.after.is_none()
--- a/crates/marmot-app/src/runtime/event_routing.rs
+++ b/crates/marmot-app/src/runtime/event_routing.rs
@@ -56,14 +56,12 @@ pub(crate) fn projection_update_matches_query(
 }
 
 pub(crate) fn timeline_query_can_apply_projection_delta(query: &TimelineMessageQuery) -> bool {
-    // Canonical source epochs are comparable only within one group. A global
-    // subscription keeps the store's wall-clock ordering by refreshing instead
-    // of locally merging per-group deltas.
-    query.group_id_hex.is_some()
-        && query
-            .search
-            .as_ref()
-            .is_none_or(|search| search.trim().is_empty())
+    // Projection application already derives canonical-vs-wall ordering from
+    // this field, so global windows can safely merge per-group deltas.
+    query
+        .search
+        .as_ref()
+        .is_none_or(|search| search.trim().is_empty())
         && query.pagination.before.is_none()
         && query.pagination.before_message_id.is_none()
         && query.pagination.after.is_none()
--- a/crates/marmot-app/src/runtime/subscriptions.rs
+++ b/crates/marmot-app/src/runtime/subscriptions.rs
@@ -138,9 +138,9 @@ impl TimelineWindow {
         query.pagination = match self.page.messages.last() {
             // Detached: re-fetch the loaded range ending exactly at the newest
             // message using an *inclusive* upper-bound cursor. The store applies
-            // the descending `LIMIT` against `<= (timeline_at, message_id_hex)`,
-            // so newer same-second rows are excluded at the SQL level and can't
-            // starve the window — no post-fetch trimming required.
+            // the descending `LIMIT` against the row-resolved canonical key, so
+            // newer rows are excluded at the SQL level and can't starve the
+            // window — no post-fetch trimming required.
             Some(newest) if !self.anchored_to_head() => TimelinePagination {
                 before: Some(newest.timeline_at),
                 before_message_id: Some(newest.message_id_hex.clone()),
@@ -206,7 +206,14 @@ impl TimelineWindowHandle {
         let older = blocking_app_task(move || query_fn(query)).await?;
         let mut window = self.lock();
         let limit = window.window_limit;
-        merge_timeline_window(&mut window.page, older, TimelineWindowEdge::Older, limit);
+        let canonical_group_order = window.base_query.group_id_hex.is_some();
+        merge_timeline_window_with_order(
+            &mut window.page,
+            older,
+            TimelineWindowEdge::Older,
+            limit,
+            canonical_group_order,
+        );
         window.bump_generation();
         Ok(window.page.clone())
     }
@@ -236,7 +243,14 @@ impl TimelineWindowHandle {
         let newer = blocking_app_task(move || query_fn(query)).await?;
         let mut window = self.lock();
         let limit = window.window_limit;
-        merge_timeline_window(&mut window.page, newer, TimelineWindowEdge::Newer, limit);
+        let canonical_group_order = window.base_query.group_id_hex.is_some();
+        merge_timeline_window_with_order(
+            &mut window.page,
+            newer,
+            TimelineWindowEdge::Newer,
+            limit,
+            canonical_group_order,
+        );
         window.bump_generation();
         Ok(window.page.clone())
     }
@@ -346,30 +360,45 @@ impl RuntimeTimelineMessagesSubscription {
     }
 }
 
-/// Sort a timeline by the canonical `(timeline_at, message_id_hex)` ascending
-/// order the store and clients agree on.
-fn sort_timeline_records(messages: &mut [TimelineMessageRecord]) {
-    messages.sort_by(|left, right| {
-        left.timeline_at
-            .cmp(&right.timeline_at)
-            .then_with(|| left.message_id_hex.cmp(&right.message_id_hex))
-    });
+/// Sort a group timeline by the accepted-history order the store and clients
+/// agree on.
+fn sort_timeline_records(messages: &mut [TimelineMessageRecord], canonical_group_order: bool) {
+    if canonical_group_order {
+        messages
+            .sort_by(|left, right| left.canonical_order_key().cmp(&right.canonical_order_key()));
+    } else {
+        messages.sort_by(|left, right| {
+            (left.timeline_at, &left.message_id_hex)
+                .cmp(&(right.timeline_at, &right.message_id_hex))
+        });
+    }
 }
 
 /// Merge a freshly-queried page into `window` on the given edge, then enforce
 /// the cap from the opposite edge. The single piece of windowing logic shared
 /// by both pagination directions.
+#[cfg(test)]
 pub(crate) fn merge_timeline_window(
     window: &mut TimelinePage,
     incoming: TimelinePage,
     edge: TimelineWindowEdge,
     limit: usize,
 ) {
+    merge_timeline_window_with_order(window, incoming, edge, limit, true);
+}
+
+fn merge_timeline_window_with_order(
+    window: &mut TimelinePage,
+    incoming: TimelinePage,
+    edge: TimelineWindowEdge,
+    limit: usize,
+    canonical_group_order: bool,
+) {
     let mut messages = std::mem::take(&mut window.messages);
     for message in incoming.messages {
         upsert_window_message(&mut messages, message);
     }
-    sort_timeline_records(&mut messages);
+    sort_timeline_records(&mut messages, canonical_group_order);
     match edge {
         TimelineWindowEdge::Older => {
             // The older side now reflects whatever the store reported. The
@@ -405,21 +434,16 @@ pub(crate) fn apply_projection_to_window(
     limit: usize,
 ) {
     let anchored = !window.has_more_after;
-    // Canonical upper bound of the window is the pair `(timeline_at,
-    // message_id_hex)`, not just the timestamp. A detached window must suppress
-    // anything sorting strictly after this — including same-second messages with
-    // a larger id, which `timeline_at`-only comparison would have admitted.
-    let newest_key = window
-        .messages
-        .last()
-        .map(|message| (message.timeline_at, message.message_id_hex.clone()));
+    // A detached window must suppress anything sorting strictly after its
+    // canonical accepted-history upper bound.
+    let newest_message = window.messages.last().cloned();
     if update.timeline_changes.is_empty() {
         for message in &update.timeline_messages {
             insert_live_message(
                 &mut window.messages,
                 message.clone(),
                 anchored,
-                newest_key.as_ref(),
+                newest_message.as_ref(),
             );
         }
     } else {
@@ -430,7 +454,7 @@ pub(crate) fn apply_projection_to_window(
                         &mut window.messages,
                         (**message).clone(),
                         anchored,
-                        newest_key.as_ref(),
+                        newest_message.as_ref(),
                     );
                 }
                 TimelineMessageChange::Remove { message_id_hex, .. } => {
@@ -441,7 +465,7 @@ pub(crate) fn apply_projection_to_window(
             }
         }
     }
-    sort_timeline_records(&mut window.messages);
+    sort_timeline_records(&mut window.messages, true);
     // Live growth only ever adds at or below the head, so trim the oldest rows
     // on overflow; the head boundary (`has_more_after`) is preserved.
     if window.messages.len() > limit {
@@ -454,15 +478,15 @@ pub(crate) fn apply_projection_to_window(
 /// Apply one live upsert. Existing messages are replaced in place (edits,
 /// reactions, delivery-state changes). A brand-new message is appended only when
 /// the window is anchored to the head, or when it sorts at or below the detached
-/// window's newest message in canonical `(timeline_at, message_id_hex)` order.
-/// `newest_key` is `None` only for an empty window, where a detached window has
+/// window's newest message in canonical accepted-history order.
+/// `newest_message` is `None` only for an empty window, where a detached window has
 /// nothing in range and therefore suppresses (an anchored empty window still
 /// appends, as the head).
 fn insert_live_message(
     messages: &mut Vec<TimelineMessageRecord>,
     message: TimelineMessageRecord,
     anchored: bool,
-    newest_key: Option<&(u64, String)>,
+    newest_message: Option<&TimelineMessageRecord>,
 ) {
     if let Some(existing) = messages
         .iter_mut()
@@ -471,9 +495,8 @@ fn insert_live_message(
         *existing = message;
         return;
     }
-    let within_window = newest_key.is_some_and(|(at, id)| {
-        (message.timeline_at, message.message_id_hex.as_str()) <= (*at, id.as_str())
-    });
+    let within_window = newest_message
+        .is_some_and(|newest| message.canonical_order_key() <= newest.canonical_order_key());
     if anchored || within_window {
         messages.push(message);
     }
--- a/crates/marmot-app/src/runtime/subscriptions.rs
+++ b/crates/marmot-app/src/runtime/subscriptions.rs
@@ -301,7 +301,8 @@ impl TimelineWindowHandle {
     fn apply_projection(&self, update: &AppProjectionUpdate) {
         let mut window = self.lock();
         let limit = window.window_limit;
-        apply_projection_to_window(&mut window.page, update, limit);
+        let canonical_group_order = window.base_query.group_id_hex.is_some();
+        apply_projection_to_window(&mut window.page, update, limit, canonical_group_order);
         window.bump_generation();
     }
 
@@ -410,34 +411,30 @@ impl RuntimeTimelineMessagesSubscription {
     }
 }
 
-/// Sort a group timeline by the accepted-history order the store and clients
-/// agree on.
-fn sort_timeline_records(messages: &mut [TimelineMessageRecord], canonical_group_order: bool) {
+fn timeline_record_order_key(
+    message: &TimelineMessageRecord,
+    canonical_group_order: bool,
+) -> (u8, u64, u8, u64, &str) {
     if canonical_group_order {
-        messages
-            .sort_by(|left, right| left.canonical_order_key().cmp(&right.canonical_order_key()));
+        message.canonical_order_key()
     } else {
-        messages.sort_by(|left, right| {
-            (left.timeline_at, &left.message_id_hex)
-                .cmp(&(right.timeline_at, &right.message_id_hex))
-        });
+        (0, message.timeline_at, 0, 0, &message.message_id_hex)
     }
+}
+
+/// Sort a timeline by the same group-canonical or global wall-clock order used
+/// by its storage query.
+fn sort_timeline_records(messages: &mut [TimelineMessageRecord], canonical_group_order: bool) {
+    messages.sort_by(|left, right| {
+        timeline_record_order_key(left, canonical_group_order)
+            .cmp(&timeline_record_order_key(right, canonical_group_order))
+    });
 }
 
 /// Merge a freshly-queried page into `window` on the given edge, then enforce
 /// the cap from the opposite edge. The single piece of windowing logic shared
 /// by both pagination directions.
-#[cfg(test)]
-pub(crate) fn merge_timeline_window(
-    window: &mut TimelinePage,
-    incoming: TimelinePage,
-    edge: TimelineWindowEdge,
-    limit: usize,
-) {
-    merge_timeline_window_with_order(window, incoming, edge, limit, true);
-}
-
-fn merge_timeline_window_with_order(
+pub(crate) fn merge_timeline_window_with_order(
     window: &mut TimelinePage,
     incoming: TimelinePage,
     edge: TimelineWindowEdge,
@@ -482,18 +479,25 @@ pub(crate) fn apply_projection_to_window(
     window: &mut TimelinePage,
     update: &AppProjectionUpdate,
     limit: usize,
+    canonical_group_order: bool,
 ) {
     let anchored = !window.has_more_after;
     // A detached window must suppress anything sorting strictly after its
-    // canonical accepted-history upper bound.
-    let newest_message = window.messages.last().cloned();
+    // query-order upper bound. Own only the message id so delta application can
+    // mutate the window without cloning the newest record's payload.
+    let newest_order = window.messages.last().map(|newest| {
+        let (class, primary, phase, at, id) =
+            timeline_record_order_key(newest, canonical_group_order);
+        (class, primary, phase, at, id.to_owned())
+    });
     if update.timeline_changes.is_empty() {
         for message in &update.timeline_messages {
             insert_live_message(
                 &mut window.messages,
                 message.clone(),
                 anchored,
-                newest_message.as_ref(),
+                newest_order.as_ref(),
+                canonical_group_order,
             );
         }
     } else {
@@ -504,7 +508,8 @@ pub(crate) fn apply_projection_to_window(
                         &mut window.messages,
                         (**message).clone(),
                         anchored,
-                        newest_message.as_ref(),
+                        newest_order.as_ref(),
+                        canonical_group_order,
                     );
                 }
                 TimelineMessageChange::Remove { message_id_hex, .. } => {
@@ -515,7 +520,7 @@ pub(crate) fn apply_projection_to_window(
             }
         }
     }
-    sort_timeline_records(&mut window.messages, true);
+    sort_timeline_records(&mut window.messages, canonical_group_order);
     // Live growth only ever adds at or below the head, so trim the oldest rows
     // on overflow; the head boundary (`has_more_after`) is preserved.
     if window.messages.len() > limit {
@@ -528,15 +533,16 @@ pub(crate) fn apply_projection_to_window(
 /// Apply one live upsert. Existing messages are replaced in place (edits,
 /// reactions, delivery-state changes). A brand-new message is appended only when
 /// the window is anchored to the head, or when it sorts at or below the detached
-/// window's newest message in canonical accepted-history order.
-/// `newest_message` is `None` only for an empty window, where a detached window has
+/// window's newest message in that query's canonical or wall-clock order.
+/// `newest_order` is `None` only for an empty window, where a detached window has
 /// nothing in range and therefore suppresses (an anchored empty window still
 /// appends, as the head).
 fn insert_live_message(
     messages: &mut Vec<TimelineMessageRecord>,
     message: TimelineMessageRecord,
     anchored: bool,
-    newest_message: Option<&TimelineMessageRecord>,
+    newest_order: Option<&(u8, u64, u8, u64, String)>,
+    canonical_group_order: bool,
 ) {
     if let Some(existing) = messages
         .iter_mut()
@@ -545,8 +551,10 @@ fn insert_live_message(
         *existing = message;
         return;
     }
-    let within_window = newest_message
-        .is_some_and(|newest| message.canonical_order_key() <= newest.canonical_order_key());
+    let within_window = newest_order.is_some_and(|newest| {
+        timeline_record_order_key(&message, canonical_group_order)
+            <= (newest.0, newest.1, newest.2, newest.3, newest.4.as_str())
+    });
     if anchored || within_window {
         messages.push(message);
     }
--- a/crates/marmot-app/src/runtime/subscriptions.rs
+++ b/crates/marmot-app/src/runtime/subscriptions.rs
@@ -7,7 +7,7 @@ use std::sync::{Arc, Mutex as StdMutex};
 use std::time::Duration;
 
 use cgka_traits::app_event::MARMOT_APP_EVENT_KIND_AGENT_STREAM_START;
-use cgka_traits::{GroupId, MessageId};
+use cgka_traits::{GroupId, MessageId, storage::StorageError};
 use serde::{Deserialize, Serialize};
 use storage_sqlite::AppEventReplayCursor;
 use tokio::sync::{broadcast, mpsc, oneshot, watch};
@@ -126,6 +126,15 @@ impl TimelineWindow {
         self.generation = self.generation.wrapping_add(1);
     }
 
+    fn head_refresh_query(&self) -> TimelineMessageQuery {
+        let mut query = self.base_query.clone();
+        query.pagination = TimelinePagination {
+            limit: Some(self.page.messages.len().max(1)),
+            ..TimelinePagination::default()
+        };
+        query
+    }
+
     /// Cursor query that re-materializes the current window from the store.
     /// Anchored windows refresh the head; detached (scrolled-back) windows
     /// refresh the loaded range ending at the current newest message so the
@@ -133,13 +142,14 @@ impl TimelineWindow {
     /// Because the window is capped at the store's single-query limit, one query
     /// always reconstitutes it.
     pub(crate) fn refresh_query(&self) -> TimelineMessageQuery {
-        let mut query = self.base_query.clone();
+        let mut query = self.head_refresh_query();
         let limit = self.page.messages.len().max(1);
         query.pagination = match self.page.messages.last() {
             // Detached: re-fetch the loaded range ending exactly at the newest
-            // message using an *inclusive* upper-bound cursor. The store applies
-            // the descending `LIMIT` against the row-resolved canonical key, so
-            // newer rows are excluded at the SQL level and can't starve the
+            // message using an *inclusive* upper-bound cursor. Group queries
+            // apply the descending `LIMIT` against the row-resolved canonical
+            // key; global queries use the wall-clock timestamp/message-id key.
+            // Either order excludes newer rows in SQL so they cannot starve the
             // window — no post-fetch trimming required.
             Some(newest) if !self.anchored_to_head() => TimelinePagination {
                 before: Some(newest.timeline_at),
@@ -149,12 +159,25 @@ impl TimelineWindow {
                 ..TimelinePagination::default()
             },
             // Anchored or empty: refresh the head.
-            _ => TimelinePagination {
-                limit: Some(limit),
-                ..TimelinePagination::default()
-            },
+            _ => query.pagination,
         };
         query
+    }
+}
+
+async fn query_timeline_page_with_cursor_refresh(
+    query_fn: Arc<TimelineQueryFn>,
+    query: TimelineMessageQuery,
+    head_query: TimelineMessageQuery,
+) -> Result<(TimelinePage, bool), AppError> {
+    let first_query_fn = query_fn.clone();
+    match blocking_app_task(move || first_query_fn(query)).await {
+        Ok(page) => Ok((page, false)),
+        Err(AppError::Storage(StorageError::TimelineCursorExpired)) => {
+            let page = blocking_app_task(move || query_fn(head_query)).await?;
+            Ok((page, true))
+        }
+        Err(error) => Err(error),
     }
 }
 
@@ -186,7 +209,7 @@ impl TimelineWindowHandle {
     /// return the updated window. A no-op (returns the current window) when no
     /// older messages exist. The store read runs off the caller thread.
     pub async fn paginate_backwards(&self, count: usize) -> Result<TimelinePage, AppError> {
-        let (query_fn, query) = {
+        let (query_fn, query, head_query, generation) = {
             let window = self.lock();
             if !window.page.has_more_before {
                 return Ok(window.page.clone());
@@ -201,9 +224,18 @@ impl TimelineWindowHandle {
                 limit: Some(count),
                 ..TimelinePagination::default()
             };
-            (window.query.clone(), query)
+            (
+                window.query.clone(),
+                query,
+                window.head_refresh_query(),
+                window.generation,
+            )
         };
-        let older = blocking_app_task(move || query_fn(query)).await?;
+        let (older, refreshed) =
+            query_timeline_page_with_cursor_refresh(query_fn, query, head_query).await?;
+        if refreshed {
+            return Ok(self.install_refresh(older, generation));
+        }
         let mut window = self.lock();
         let limit = window.window_limit;
         let canonical_group_order = window.base_query.group_id_hex.is_some();
@@ -223,7 +255,7 @@ impl TimelineWindowHandle {
     /// window already includes the head. Reaching the head re-anchors the
     /// window (`has_more_after` becomes false).
     pub async fn paginate_forwards(&self, count: usize) -> Result<TimelinePage, AppError> {
-        let (query_fn, query) = {
+        let (query_fn, query, head_query, generation) = {
             let window = self.lock();
             if !window.page.has_more_after {
                 return Ok(window.page.clone());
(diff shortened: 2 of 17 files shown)


agent-p1p:

BLOCKING:
- **Canonical pagination falls back to a wall-clock predicate when the cursor tuple is stale or missing** — `crates/storage-sqlite/src/timeline.rs:880-889`, `crates/storage-sqlite/src/timeline.rs:2447-2458`, `crates/storage-sqlite/src/timeline.rs:2625-2666`. A group cursor is resolved only by the full mutable `(group_id, timeline_at, message_id)` tuple. If that lookup misses, the query applies a `(timeline_at, message_id)` wall-clock boundary but still orders the page by the epoch-based canonical key. Those orders are not monotonic: a row that is canonically after the cursor can satisfy the wall-clock "before" predicate, producing skips/duplicates across pages. This is not limited to fabricated cursors: conflict upserts replace `recorded_at` (`timeline.rs:567-576`), and synthesized group-system rows deliberately have stable ids so reprocessing the same change upserts it (`crates/marmot-app/src/client/projection.rs:498-505`), changing the persisted `timeline_at` behind an already-issued cursor. Resolve a retained row by stable group/message id even when its old timestamp changed, and do not combine the wall-clock fallback predicate with canonical ordering when the row is truly gone (return a refresh/error or use a cursor that carries the canonical key). Add a regression covering a stale system-row cursor across an epoch boundary.

SUGGESTION:
- **Update the ordering boundary contract** — `docs/marmot-architecture/overview/marmot-app-runtime.md:100-109` still names `TIMELINE_ORDER_BY_ASC/_DESC = (timeline_at, message_id)` as the cross-client display/pagination order. The implementation now distinguishes canonical group order from wall-clock global/retention order; the architecture doc should say so.

NITPICK:
- none

SUMMARY: The core epoch/phase key is consistently mirrored across Rust sorting, SQL ordering, chat-list/read-marker scans, and runtime merge paths, and the added tests cover the main delayed-projection scenario. One cursor fallback still mixes incompatible boundaries and ordering, so group pagination is not durable under a stale/pruned cursor. CI is green at `eb01940da627eceecb8548d4b821d8fb2a3dc2dd`; local `storage-sqlite` tests passed, and the one unrelated `marmot-app` timing failure passed on exact rerun.

VERDICT: REQUEST_CHANGES

SENSITIVE_PATHS: none (derived timeline/storage ordering; no MLS/CGKA engine, crypto/key, trust-anchor, auth, or group-state-core changes)

agent-p1p:
Addressed the blocking pagination finding in 6def5d28.

- Canonical group cursors now resolve by stable (group_id_hex, message_id_hex), so a reprojected system row can change local observation time without switching pagination to a wall-clock predicate.
- A pruned/missing canonical cursor returns a refresh error instead of mixing incompatible boundary and sort orders.
- Added stale-system-row and missing-row regressions and updated the architecture ordering contract.

Verification: 338 storage-sqlite tests passed; CARGO_INCREMENTAL=0 just fast-ci passed.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1201,28 +1318,51 @@ fn unread_summary_tx(
...
+                    rusqlite::types::Value::Integer(u64_to_i64(at)?),
+                    rusqlite::types::Value::Text(id.to_owned()),
+                ],
             )
         } else {
             (
-                "timeline_at > ?4 AND (?5 = ?5)",
-                read_state.initialized_at,
-                "",
+                "(timeline_at > ?4 OR (timeline_at = ?4 AND message_id_hex > ?5))",
```
erskingardner, line 1272: **Blocking: preserve the canonical read boundary after pruning.** This fallback reintroduces the non-monotonic order the PR is removing. For example, let A be `(epoch 7, timeline_at 200)` and B be `(epoch 8, timeline_at 100)`, then mark B read. A retention cutoff of 150 prunes B but retains A; `refresh_chat_list_unread_after_secure_prune_tx` reaches this branch and counts A as unread because `200 > 100`, even though A is canonically before the read marker. A delayed epoch-9 row at time 50 is likewise missed, and the analogous fallback in `mark_timeline_message_read` can move the marker backward. Persist the canonical marker key/source epoch in `conversation_read_state` (or retain an equivalent anchor) and use it after the timeline row is deleted; please add a skewed-epoch prune regression.

crates/storage-sqlite/src/timeline.rs
```diff
@@ -2483,6 +2704,30 @@ fn escape_like_literal(value: &str) -> String {
...
+pub(crate) fn canonical_timeline_order_key<'a>(
+    source_message_id_hex: Option<&str>,
+    source_epoch: Option<u64>,
+    kind: u64,
+    timeline_at: u64,
+    message_id_hex: &'a str,
+) -> (u8, u64, u8, u64, &'a str) {
+    let (class, primary) = match source_epoch {
+        Some(epoch) => (1, epoch),
+        None if source_message_id_hex.is_none() => (2, timeline_at),
```
erskingardner: **Blocking: do not rank permanent failed sends as optimistic head rows.** `source_message_id_hex = NULL` and `source_epoch = NULL` is also the durable state left by `retract_failed_local_projection(..., "local_publish_failed")`, not just a pending optimistic row. Class 2 therefore sorts an old failed send after every later class-1 accepted message forever. The group timeline and `latest_kind9_message_tx` (which intentionally includes `local_publish_failed`) remain pinned to that failure, so the existing contract that a later delivered message replaces the failed preview regresses once the delivered fixture carries a real source epoch. Distinguish pending from failed in the canonical key/SQL (or persist an ordering epoch for the failed attempt), and add a failed-then-delivered regression with the delivered row in a real epoch.

erskingardner (changes requested):
The epoch/phase order is coherent across the main storage and runtime paths, and CI is green at this head. Two blocking transitions still mix the new canonical order with state that only retains the old wall-clock tuple: pruned read markers, and unsourced rows that have become permanent publish failures.

crates/storage-sqlite/src/timeline.rs
erskingardner: **BLOCKING** — class 2 outranks all canonical history permanently, not just while a send is in flight.

The comment on the constants says class 2 keeps "optimistic local rows (no source id and no epoch yet) at the head". But a send that fails to publish keeps exactly that shape forever: `record_local_app_event_projection` is called pre-publish with `source_message_id_hex = None` and `source_state = None` (`crates/marmot-app/src/client/mod.rs:2016`), and the failure path only stamps `invalidation_status = 'local_publish_failed'` (`retract_failed_local_projection`, `crates/marmot-app/src/client/mod.rs:2144-2160`). The epoch is never backfilled, and a timeline rebuild reprojects the same NULLs from `app_events`.

So `(2, timeline_at, ...)` is not "yet" — it is terminal for any failed send, and class 2 beats class 1 regardless of time.

Repro against a replica of the schema (sqlite 3.53), one failed local row at t=1000 and a later peer row at t=5000 in epoch 42, using the exact `latest_kind9_message_tx` predicate:

```
--- canonical latest (this PR) ---   failed-local|1000
--- wall clock latest (master) ---   peer-new|5000
```

Suggested fix: give unresolved local rows the group's current epoch as `primary` (class 1, phase 1, `timeline_at` tie-break) so they land at the head of the epoch they were composed in and are superseded normally once newer epoch traffic arrives — rather than a class that sorts above all accepted history. If class 2 is kept deliberately, it needs to be scoped to rows that are still in flight, and the chat-list projections need to exclude it.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1383,7 +1532,16 @@ fn latest_kind9_message_tx(
                invalidation_status IS NULL
                OR (direction = 'sent' AND invalidation_status = 'local_publish_failed')
            )
-         ORDER BY timeline_at DESC, message_id_hex DESC
+         ORDER BY CASE
```
erskingardner: **BLOCKING (same root cause as the `timeline.rs` comment)** — user-visible consequence.

This predicate deliberately includes `direction = 'sent' AND invalidation_status = 'local_publish_failed'` rows, and those rows are class 2, so after this change a stale failed send wins this `LIMIT 1` forever. Two knock-on effects:

1. `last_message_*` (preview text, sender, kind, delivery state) freezes on the failed message even after peers send new messages.
2. `activity_sort_at` is derived from `latest.timeline_at` (`rebuild_chat_list_row_for_group_tx`, around line 1161), so the conversation also stops rising in the chat list on new activity — `MAX` never sees the newer row's timestamp.

The same ordering was applied to `refresh_chat_list_last_message_after_secure_prune_tx` (`crates/storage-sqlite/src/timeline.rs:1223`) and to all nine completeness subqueries above, so the projection stays self-consistent — consistently wrong, and the completeness check will not detect it.

A regression test along the lines of `chat_list_read_state_and_preview_follow_canonical_epoch_order`, but recording a `local_publish_failed` sent row before a later remote message, would pin this.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1232,17 +1372,26 @@ fn unread_summary_tx(
            AND invalidation_status IS NULL
            AND sender != ?3
            AND {where_sql}
-         ORDER BY timeline_at ASC, message_id_hex ASC"
+         ORDER BY CASE
```
erskingardner: The pruned-marker fallback mixes the two boundary systems the previous review round blocked for pagination.

When `timeline_message_for_read_marker_tx` misses, `where_sql` becomes the wall-clock predicate `(timeline_at > ?4 OR (timeline_at = ?4 AND message_id_hex > ?5))`, but this scan still orders canonically. The unread *set* is then defined by wall clock while `first_message_id` is the canonically-lowest member of that set, so "jump to first unread" can land on a row that is not the first unread in the order the timeline actually renders, and the count can include or exclude rows across an epoch boundary.

Since the epoch really is unknowable in that branch, the consistent fallback is to use the wall-clock ORDER BY too (the same split `message_timeline_by_wall_clock` already makes at the pagination layer), rather than a wall-clock predicate under a canonical sort.

Same shape at `mark_timeline_message_read` (line ~586): a target row is compared against a missing marker with `timeline_tuple_after`, so a later wall-clock row from an earlier epoch can move the marker backward in canonical order — the case the new test explicitly guards against when the marker row *is* retained.

crates/storage-sqlite/src/timeline.rs
```diff
@@ -2340,13 +2441,52 @@ fn timeline_records_by_ids_tx(
...
+                    row.get::<_, i64>(3)?,
+                    row.get::<_, String>(4)?,
+                ))
+            },
+        )
+        .optional()
+        .storage()?;
+    let Some((source_message_id_hex, source_epoch, kind, timeline_at, message_id_hex)) = row else {
+        return Err(StorageError::Backend(
+            "group timeline cursor no longer exists; refresh the timeline".to_owned(),
```
erskingardner: Failing loudly here is the right call over mixing predicates, but two things make this hard to consume:

1. It is reachable in ordinary operation, not only for fabricated cursors. Retention expiry removes the underlying `app_event` and the single-message reprojection then deletes the timeline row (`crates/storage-sqlite/src/timeline.rs:1380-1388`). A window scrolled back into expiring history will hit exactly this on the next `load_older`.
2. The contract is a magic string. `StorageError::Backend(String)` is indistinguishable from any other backend failure, and the new tests assert on the literal message. No caller in `marmot-app` handles it — `TimelineWindowHandle::load_older` / `load_newer` propagate it straight out to the FFI consumer, so the documented "caller must refresh" behavior does not exist yet.

Suggest a typed variant (e.g. `StorageError::TimelineCursorExpired`) plus handling in `TimelineWindowHandle` that converts it into the existing `TimelineSubscriptionSignal::Refresh` path, which already knows how to re-materialize a window.

crates/storage-sqlite/src/migrations/0044_timeline_canonical_order.rs
```diff
@@ -0,0 +1,51 @@
+use crate::SqliteResultExt;
+use cgka_traits::storage::StorageResult;
+use rusqlite::Transaction;
+
+/// Index the accepted-history timeline order introduced by mdk#1172. The key is
+/// computed from existing projection columns, so upgraded databases immediately
+/// gain canonical ordering without rewriting or re-projecting rows.
+pub(crate) fn apply(tx: &Transaction<'_>) -> StorageResult<()> {
```
erskingardner: Simplification worth taking before this lands: the same five-term CASE ladder is now written out in this index, in `TIMELINE_GROUP_ORDER_BY_ASC`/`_DESC`, twice in the keyset predicates, in `refresh_chat_list_last_message_after_secure_prune_tx`, in `latest_kind9_message_tx`, in the unread predicate and unread scan, and in nine subqueries of `chat_list_projection_complete_tx` — plus a Rust mirror in `canonical_timeline_order_key`. That is the bulk of the +953/-127.

Most chat-list copies use only four of the five terms (they drop `phase` because the scan is chat-only), so the invariant "these expressions are all the same key" is already only true by inspection. The next query that needs this order will copy a ladder and get a term wrong, and nothing will fail.

VIRTUAL generated columns collapse all of it and are addable with plain `ALTER TABLE` (no rewrite, still indexable):

```sql
ALTER TABLE message_timeline ADD COLUMN order_class INTEGER
  GENERATED ALWAYS AS (CASE WHEN source_epoch IS NOT NULL THEN 1
                            WHEN source_message_id_hex IS NULL THEN 2 ELSE 0 END) VIRTUAL;
-- order_primary, order_phase, order_at likewise
CREATE INDEX idx_message_timeline_group_canonical_order
  ON message_timeline (group_id_hex, order_class, order_primary, order_phase, order_at, message_id_hex);
```

Every ORDER BY then becomes `ORDER BY order_class, order_primary, order_phase, order_at, message_id_hex`, the keyset predicate becomes a plain row-value comparison, and `canonical_timeline_order_key` stays the single Rust definition. Net effect should be a large deletion rather than an addition.

crates/marmot-app/src/runtime/subscriptions.rs
```diff
@@ -346,30 +360,45 @@ impl RuntimeTimelineMessagesSubscription {
...
+            (left.timeline_at, &left.message_id_hex)
+                .cmp(&(right.timeline_at, &right.message_id_hex))
+        });
+    }
 }
 
 /// Merge a freshly-queried page into `window` on the given edge, then enforce
 /// the cap from the opposite edge. The single piece of windowing logic shared
 /// by both pagination directions.
+#[cfg(test)]
```
erskingardner: Nits, non-blocking:

- A `#[cfg(test)]` wrapper that exists only to pass `true` is production surface for test convenience. Having the two tests call `merge_timeline_window_with_order(..., true)` directly deletes this shim.
- `apply_projection_to_window` hardcodes `sort_timeline_records(&mut window.messages, true)` (line 468) while the pagination paths derive the flag from `base_query.group_id_hex`. It is correct today only because `timeline_query_can_apply_projection_delta` gates on a group id; if that gate ever loosens this silently sorts a cross-group window by epoch. Pass the window's own flag.
- `let newest_message = window.messages.last().cloned()` (line 439) clones a full record — plaintext, media json, reactions — per projection application, where the old code cloned a `(u64, String)` key. Computing an owned key once keeps the previous cost.

erskingardner (changes requested):
## Review — changes requested

The premise is right and the core key is sound: deriving per-group order from authenticated `source_epoch`, placing the epoch-N state row before epoch-N application rows, and tie-breaking on deterministic ids does give cross-device convergence for the scenario in #1172. The delayed-catch-up regression test is the right test. Ordering is mirrored consistently across SQL, the Rust sort, chat-list projection, unread scan and read-marker advancement, and the cursor-resolution fix from the previous round is a real improvement.

One blocking behavior regression, plus one boundary inconsistency that is the same species the previous review blocked, plus a significant duplication concern.

**BLOCKING**
- Class 2 (no source id, no epoch) sorts *after* every canonical row unconditionally, and a `local_publish_failed` send stays class 2 forever. It therefore permanently wins `latest_kind9_message_tx`, so the chat-list preview and `activity_sort_at` freeze on a failed send even after newer messages arrive. Details inline on `crates/storage-sqlite/src/timeline.rs` and `crates/storage-sqlite/src/chat_list.rs`.

**NON-BLOCKING (worth resolving before merge)**
- The pruned-read-marker fallback in `unread_summary_tx` combines a wall-clock predicate with canonical ORDER BY — the exact mix rejected for pagination in the previous round. Inline.
- The missing-cursor path is a new hard failure reachable in normal operation (retention expiry deletes timeline rows), signalled only by a magic `StorageError::Backend` string that no caller handles. Inline.
- The 5-term CASE ladder is now copy-pasted ~12 times across two files, the migration, and mirrored again in Rust. Inline suggestion: indexed VIRTUAL generated columns collapse all of it and make it impossible for a future query to reproduce four of five terms.

**Other notes (no inline anchor available)**
- CLI `message list` / `message timeline list` still document `--before` / `--after` as "Only include messages before/after this unix timestamp" (`crates/cli/src/args.rs:735-745`, `:783-793`). For a group-scoped query the timestamp is now ignored entirely (the row is resolved by id), and an id that does not exist errors instead of returning a page. Either update the help text or keep a documented wall-clock mode for the CLI.
- `timeline_query_can_apply_projection_delta` now returns false for all-groups subscriptions, so every projection event on any group triggers a full window re-query for a global timeline subscriber instead of an in-place delta. Correct, but it is a per-message cost that is not called out in the PR description.
- Verified separately (sqlite 3.53, replica schema): the migration-44 expression index is picked for both the canonical ORDER BY and the LEFT JOIN form — `SEARCH timeline USING COVERING INDEX idx_message_timeline_group_canonical_order`, no temp b-tree. No perf concern there.
- Test-coverage gap: the issue's suggested regression is a two-client relay-level assertion; this PR verifies only at the storage/window layer. Acceptable, but nothing covers a failed-send row against later inbound messages, which is exactly where the blocking issue lives.

agent-p1p:
@coderabbitai review

crates/storage-sqlite/src/chat_list.rs
agent-p1p: Addressed in 04b67257. `conversation_read_state` now persists the canonical class/primary anchor, migration 0045 backfills retained markers, and timeline reprojection keeps optimistic markers synchronized as they reconcile or fail. Pruned-anchor and skewed-epoch regressions cover the retained-anchor path; legacy states without an anchor use wall-clock predicate and wall-clock ordering together.

crates/storage-sqlite/src/timeline.rs
agent-p1p: Addressed in 04b67257. Failed unresolved local sends are class 0 rather than optimistic class 2, so authenticated class-1 history supersedes them. The delivered-in-real-epoch preview regression now covers this path.

agent-p1p: Addressed in 04b67257. Class 2 is now limited to unresolved rows with no invalidation; `local_publish_failed` rows fall back to class 0. The generated SQL key and Rust key share this distinction, with parity coverage.

crates/storage-sqlite/src/chat_list.rs
agent-p1p: Addressed in 04b67257. `latest_kind9_message_tx` uses the corrected generated key, and a failed-local-then-authenticated-remote regression verifies both preview and activity can advance.

crates/storage-sqlite/src/chat_list.rs
agent-p1p: Addressed in 04b67257. Persisted canonical anchors are used after pruning. For legacy states where no canonical anchor can be recovered, unread selection switches both the predicate and ORDER BY to `(timeline_at, message_id_hex)`; marker advancement uses the same canonical-or-wall split.

crates/storage-sqlite/src/timeline.rs
agent-p1p: Addressed in 04b67257. Missing retained group cursors now return typed `StorageError::TimelineCursorExpired`; runtime older/newer pagination catches only that variant and atomically refreshes the window head. Other storage failures still propagate. Storage/runtime retention regressions cover it.

crates/storage-sqlite/src/migrations/0044_timeline_canonical_order.rs
agent-p1p: Addressed in 04b67257. Migration 0045 adds indexed VIRTUAL columns for class/primary/phase/time. Timeline, keyset, chat-list, completeness, and unread SQL now use the full generated key; a migration test and Rust-vs-SQL parity test guard drift.

crates/marmot-app/src/runtime/subscriptions.rs
agent-p1p: Addressed in 6197f6df. The test-only wrapper is gone; tests call `merge_timeline_window_with_order` directly. Projection application receives the window's canonical-vs-wall ordering mode, shares one key helper for filtering/sorting, and owns only the compact upper-bound key instead of cloning the newest record. A global-window regression verifies wall-clock ordering.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1201,28 +1318,51 @@ fn unread_summary_tx(
...
+        }
+    } else {
+        (
+            "timeline_at > ?4 AND (?5 = ?5)",
+            vec![
+                rusqlite::types::Value::Integer(u64_to_i64(read_state.initialized_at)?),
+                rusqlite::types::Value::Text(String::new()),
+            ],
+        )
+    };
```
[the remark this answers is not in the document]
agent-p1p: Addressed in 04b67257: canonical read anchors are persisted/backfilled and reused after pruning. Legacy no-anchor states switch both unread predicate and scan ordering to wall clock; marker advancement follows the same split.

crates/storage-sqlite/src/migrations/0044_timeline_canonical_order.rs
```diff
@@ -0,0 +1,51 @@
...
+            ELSE 1
+         END),
+        (CASE
+            WHEN kind = 1210
+             AND source_message_id_hex IS NULL
+             AND source_epoch IS NOT NULL THEN 0
+            ELSE timeline_at
+         END),
+        message_id_hex
+    );
```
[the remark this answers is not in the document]
agent-p1p: Addressed in 04b67257. Migration 0045 defines indexed VIRTUAL canonical-order columns, and every listed SQL consumer now references the complete five-term generated key. Migration and Rust/SQL parity tests were added.

crates/storage-sqlite/src/timeline.rs
```diff
@@ -2340,13 +2441,52 @@ fn timeline_records_by_ids_tx(
...
+    let timeline_at = i64_to_u64(timeline_at)?;
+    let (class, primary, phase, at, _) = canonical_timeline_order_key(
+        source_message_id_hex.as_deref(),
+        source_epoch,
+        kind,
+        timeline_at,
+        &message_id_hex,
+    );
+    Ok((class, primary, phase, at, message_id_hex))
+}
```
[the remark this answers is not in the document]
agent-p1p: Addressed in 04b67257. `TimelineCursorExpired` is a typed storage variant, runtime pagination refreshes on that variant, and other backend errors remain unchanged. Tests assert the variant and refresh behavior.

crates/storage-sqlite/src/timeline.rs
```diff
@@ -2483,6 +2704,30 @@ fn escape_like_literal(value: &str) -> String {
...
+        && source_message_id_hex.is_none()
+        && source_epoch.is_some()
+    {
+        0
+    } else {
+        1
+    };
+    let intra_epoch_at = if phase == 0 { 0 } else { timeline_at };
+    (class, primary, phase, intra_epoch_at, message_id_hex)
+}
```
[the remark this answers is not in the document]
agent-p1p: Addressed in 04b67257. Only unresolved, non-invalidated rows remain class 2; permanently failed sends are class 0 and no longer pin accepted history or chat-list activity. Epoch-backed delivered regressions cover timeline and preview behavior.

docs/marmot-architecture/overview/marmot-app-runtime.md
```diff
@@ -104,9 +104,14 @@ add the incremental, bounded, non-blocking, single-source properties around it s
...
-   - **Materialized-timeline order** — `TIMELINE_ORDER_BY_ASC/_DESC` = `(timeline_at, message_id_hex)` over
-     `message_timeline`/chat-list; the cross-client user-visible display + pagination order (`timeline_at == recorded_at`
-     at projection). The replay cursor MUST NOT be applied to timeline pagination.
+   - **Materialized-timeline order** — authenticated `source_epoch` boundaries define the cross-client order for a
+     single group's `message_timeline`/chat-list rows. `TIMELINE_GROUP_ORDER_BY_ASC/_DESC` places synthesized state
+     changes before authenticated application messages in the epoch they establish and uses stable message ids to
+     break remaining ties. Group pagination resolves a retained cursor by `(group_id_hex, message_id_hex)` to recover
+     that canonical key; if the row was pruned, the caller must refresh instead of applying a wall-clock predicate to
+     canonical order. `TIMELINE_WALL_ORDER_BY_ASC/_DESC` = `(timeline_at, message_id_hex)` remains the order for
+     cross-group queries and retention scans, where epochs are not comparable. The replay cursor MUST NOT be applied
```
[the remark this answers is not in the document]
agent-p1p: Addressed in 04b67257. The architecture text now distinguishes incomparable cross-group queries from one-group time-based retention expiry while preserving the replay-cursor restriction.

agent-p1p:
Addressed the blocking and follow-up review findings.

- Persisted canonical read anchors across retention; legacy no-anchor states use wall-clock predicate and ordering together.
- Added typed cursor expiry and atomic runtime head refresh.
- Limited optimistic class 2 to in-flight rows; failed sends no longer pin authenticated history or chat previews.
- Replaced duplicated SQL CASE ladders with indexed VIRTUAL canonical-order columns in migration 0045.
- Kept read anchors synchronized when pending sends reconcile or fail.
- Propagated canonical-vs-wall ordering through projection deltas, removed the test shim/full-record clone, and clarified CLI cursor help plus retention docs.

Validation:
- `cargo test -p storage-sqlite --lib --locked` — 351 passed.
- `cargo test -p marmot-app --lib --locked` — 614 passed.
- Focused projection tests — 9 passed.
- `just fast-ci` — passed after rebase onto current `master`.

agent-p1p:
@coderabbitai review

Datawav (commented):
## Automated review

No actionable findings.

crates/storage-sqlite/src/timeline.rs
```diff
@@ -2483,6 +2671,33 @@ fn escape_like_literal(value: &str) -> String {
...
+    kind: u64,
+    timeline_at: u64,
+    message_id_hex: &'a str,
+) -> (u8, u64, u8, u64, &'a str) {
+    let (class, primary) = match source_epoch {
+        Some(epoch) => (1, epoch),
+        None if source_message_id_hex.is_none() && invalidation_status.is_none() => {
+            (2, timeline_at)
+        }
+        None => (0, timeline_at),
```
erskingardner, line 2687: **[P1] Restore the prior read anchor when an optimistic send fails**

In an established conversation, the optimistic local send becomes the read marker as class 2. This branch changes it to class 0 on `local_publish_failed`, and `upsert_message_timeline_row_tx` synchronizes that new key into `conversation_read_state`. Because `unread_summary_tx` counts every class-1 remote row greater than the marker, all retained accepted history—including messages that were already read before the send—becomes unread. The new test starts with no accepted history, so it only proves a later message is exposed. Preserve/restore the previous canonical anchor (or avoid advancing it for the pre-publish local projection, since local-sender rows are excluded from unread counts), and cover: read epochs 1–8, send/fail, unread remains 0; receive epoch 9, unread becomes 1.

erskingardner (changes requested):
The two blockers from the previous review are addressed on this head, including durable canonical read anchors and typed cursor expiry. One failure-path regression remains: demoting a failed optimistic send below accepted history also demotes its synchronized read anchor, which makes already-read history unread. The failed-send regression needs an established-history case before this is safe to merge.

crates/storage-sqlite/src/timeline.rs
erskingardner, line 2687: **BLOCKING** — demoting failed local sends to class 0 makes them disappear from the conversation.

The previous round's finding was that a `local_publish_failed` row stayed class 2 forever and so permanently won `latest_kind9_message_tx`. This arm fixes that by routing it to class 0 — but class 0 sorts *before every* canonical row, so the failed message moves from the tail of the timeline to the very beginning of history.

Reproduced against the exact migration-0045 expressions (sqlite 3.53), one peer message per epoch plus one failed local send composed most recently:

```
--- timeline ASC (canonical) ---
my-failed-send|class 0|primary 6000
peer-old      |class 1|primary 40
peer-new      |class 1|primary 41

--- head page: canonical DESC, LIMIT 2 ---
peer-new
peer-old
```

The head page is built with `TIMELINE_GROUP_ORDER_BY_DESC` + `LIMIT`, and the timeline query applies no `invalidation_status` filter (`timeline_query_sql` selects `timeline.invalidation_status` and passes it through to the FFI, where hosts render the `failed` delivery state). So in any group with more than a window's worth of history, the user sends a message, publish fails, and the row is not in the returned page at all — it only reappears if they scroll all the way back to the start of the group. With the chat-list preview now correctly showing the delivered message, there is no visible trace of the failure left.

It is also a jump: the same row is class 2 (tail) while in flight and class 0 (head of history) the moment it fails.

`virtual_order_columns_match_the_rust_canonical_key` asserts `failed_class == 0`, so the behavior is pinned rather than caught, and `failed_local_send_does_not_replace_delivered_chat_preview` only checks the preview.

The previous review's suggestion still stands as the coherent shape: keep unresolved/failed local rows inside the epoch they were composed in (class 1, phase 1, `timeline_at` tie-break) so later epochs supersede them normally. If carrying a compose epoch is too large for this PR, the alternative is to leave failed rows in class 2 — where they visually belong — and scope the *preview* rule instead, so `latest_kind9_message_tx` cannot let a class-2 row outrank accepted history indefinitely. Either way, a regression asserting that a failed send is still present in the default head page alongside newer inbound messages would pin the property that matters.

crates/marmot-app/src/runtime/event_routing.rs
```diff
@@ -56,10 +56,14 @@ pub(crate) fn projection_update_matches_query(
...
 
 pub(crate) fn timeline_query_can_apply_projection_delta(query: &TimelineMessageQuery) -> bool {
-    query
-        .search
-        .as_ref()
-        .is_none_or(|search| search.trim().is_empty())
+    // Canonical source epochs are comparable only within one group. A global
+    // subscription keeps the store's wall-clock ordering by refreshing instead
+    // of locally merging per-group deltas.
+    query.group_id_hex.is_some()
```
erskingardner, line 62: Non-blocking, but this gate looks redundant now — and it is not free.

The reason global windows could not take a delta was that they would be sorted by an order that does not apply to them. This PR fixes exactly that: `apply_projection_to_window` takes `canonical_group_order`, `TimelineWindowHandle::apply_projection` derives it from `window.base_query.group_id_hex`, and `apply_projection_preserves_wall_clock_order_for_global_windows` proves the delta path sorts a cross-group window by wall clock correctly.

So with the flag threaded through, an all-groups subscription can merge per-group deltas safely; adding `group_id_hex.is_some()` here instead turns every projection event on any group into a full window re-query for global subscribers. That is a per-message cost paid to avoid a case the rest of the PR already handles.

Either drop the clause (the flag is the fix) or keep it and drop the flag threading for the projection path — carrying both means one of them is dead.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1441,13 +1522,65 @@ fn chat_list_message_from_row(row: &rusqlite::Row<'_>) -> rusqlite::Result<ChatL
...
+    manually_marked_unread: bool,
+) -> StorageResult<()> {
+    let latest = latest_kind9_message_tx(tx, group_id_hex)?;
+    let (message_id, timeline_at, order_class, order_primary) = match latest {
+        Some(message) => {
+            let marker =
+                timeline_message_for_read_marker_tx(tx, group_id_hex, &message.message_id_hex)?
+                    .ok_or_else(|| {
+                        StorageError::Backend(
+                            "latest chat message disappeared while creating read state".to_owned(),
```
erskingardner, line 1537: Non-blocking: this reintroduces a `StorageError::Backend` magic string, the species flagged last round, for a case that should not need to exist.

`latest_kind9_message_tx` already reads `message_timeline` in this same transaction; the only reason for the second lookup is that its projection does not carry `source_epoch` / `invalidation_status`. Selecting those two columns in `latest_kind9_message_tx` (or having it return the canonical key alongside the preview) removes the extra query, the `ok_or_else`, and the unreachable error string in one go.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -247,10 +247,46 @@ struct AccountGroupRow {
...
     last_read_message_id_hex: Option<String>,
     last_read_timeline_at: Option<u64>,
+    last_read_order_class: Option<u8>,
+    last_read_order_primary: Option<u64>,
     initialized_at: u64,
     manually_marked_unread: bool,
 }
 
+impl ConversationReadState {
+    fn canonical_order_key(&self) -> Option<(u8, u64, u8, u64, &str)> {
```
erskingardner, line 257: Nit: the hardcoded `1` for `phase` is only safe because every write path into `conversation_read_state` goes through a kind-9 lookup (`timeline_message_for_read_marker_tx` and `latest_kind9_message_tx` both filter `kind = MARMOT_APP_EVENT_KIND_CHAT`, and phase is 0 only for synthesized kind-1210 rows).

That is a real invariant but an invisible one — a future read-marker surface that accepts a system row would silently compare against the wrong key. Worth either a one-line comment stating the kind-9 precondition, or persisting `last_read_order_phase` next to the other two so the key is stored rather than reconstructed.

crates/storage-sqlite/src/timeline.rs
```diff
@@ -1607,6 +1673,35 @@ fn upsert_message_timeline_row_tx(tx: &Connection, row: &TimelineRow) -> Storage
...
+        row.source_message_id_hex.as_deref(),
+        row.source_epoch,
+        row.invalidation_status.as_deref(),
+        row.kind,
+        row.timeline_at,
+        &row.message_id_hex,
+    );
+    tx.execute(
+        "UPDATE conversation_read_state
+         SET last_read_timeline_at = ?3,
```
erskingardner, line 1690: Non-blocking: this fires unconditionally on *every* timeline row upsert, so a `rebuild_message_timeline_for_group` over N rows now issues N extra prepare+execute round trips against `conversation_read_state`, of which at most one can ever match.

Guarding it — only run when the row's canonical key actually changed, or hoist the marker id once per rebuild and compare in Rust — keeps the same guarantee for a fraction of the writes. The correctness of the sync itself looks right; `failed_pending_read_marker_does_not_hide_authenticated_history` covers the class 2 → class 0 transition through the invalidation path.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -828,7 +836,11 @@ fn chat_list_projection_complete_tx(
                                   AND mt.invalidation_status = 'local_publish_failed'
                               )
                           )
-                        ORDER BY mt.timeline_at DESC, mt.message_id_hex DESC
+                        ORDER BY mt.timeline_order_class DESC,
```
erskingardner, line 839: Non-blocking, but this is where the promised deletion did not materialize. Migration 0045 collapsed the five-term CASE ladders correctly — and then this five-line `ORDER BY` got copy-pasted ten times in this file plus once in `refresh_chat_list_last_message_after_secure_prune_tx`, so the diff grew (+1429/-216) rather than shrinking.

A `const CHAT_LIST_CANONICAL_ORDER_DESC: &str` interpolated into the query strings makes "these all use the same key" true by construction instead of by inspection, which was the original point of the finding.

erskingardner (changes requested):
## Re-review — changes requested

Most of the previous round landed well:

- **Pruned read markers** — persisted `last_read_order_class/_primary`, backfilled by the migration, resynced on reprojection, and the no-anchor branch now uses a wall-clock predicate *and* a wall-clock `ORDER BY` together. The mixed-boundary bug is gone, and `pruned_read_marker_keeps_canonical_epoch_anchor` pins it.
- **Typed cursor expiry** — `StorageError::TimelineCursorExpired` plus a real head-refresh path in `TimelineWindowHandle`, with a regression test. The magic-string contract is gone.
- **CASE-ladder duplication** — collapsed into four indexed VIRTUAL generated columns in migration 0045, with `virtual_order_columns_match_the_rust_canonical_key` keeping SQL and Rust from drifting. This was the right call.
- **Subscription nits** — the `#[cfg(test)]` shim is gone, `apply_projection_to_window` takes the window's own ordering mode instead of hardcoding `true`, and the newest-record clone is down to an owned id.

One blocking problem: the fix for the class-2 finding overshot. Demoting a permanently-failed local send to class 0 stops it winning the chat preview, but class 0 sorts *below all* accepted history, so the user's own failed message leaves the head window entirely. That is a worse regression than the one it replaced, and the PR's own `virtual_order_columns_match_the_rust_canonical_key` asserts the new behavior, so nothing catches it.

Everything else below is non-blocking.

**Other notes (no inline anchor)**
- Size still went the wrong way: +1429/-216, up from +953/-127. Roughly 670 of that is tests, which is fine, but the generated-columns consolidation should have produced a net deletion in the SQL and did not — see the `chat_list_projection_complete_tx` comment.
- CI is green at `6197f6df`. I verified the migration expressions independently in sqlite (see the blocking comment for the reproduction).
- `crates/cli/src/args.rs` help now says "requires --before-message-id" but still does not say the id must name a *retained* row; for a group-scoped query a stale id now returns `TimelineCursorExpired` rather than a page. Worth one more clause.

Datawav (dismissed):

crates/marmot-app/src/runtime/event_routing.rs
```diff
@@ -56,6 +56,8 @@ pub(crate) fn projection_update_matches_query(
 }
 
 pub(crate) fn timeline_query_can_apply_projection_delta(query: &TimelineMessageQuery) -> bool {
+    // Projection application already derives canonical-vs-wall ordering from
+    // this field, so global windows can safely merge per-group deltas.
     query
```
erskingardner, line 61: **[P1] Keep global subscriptions on refresh until delta identity is group-scoped**

Dropping the group gate sends all-groups projection updates into `apply_projection_to_window`, but its upsert/page-merge/remove paths identify rows only by `message_id_hex`. Those ids are not group-scoped: `message_id_invalidation_is_group_scoped` documents that the same inner event id can exist in two groups. A global window containing both will therefore let an update for group A replace group B's row, while a remove for A deletes both copies. Qualify window identity by `(group_id_hex, message_id_hex)` throughout merge/delta application (using the enclosing update's group for `Remove`) and add a cross-group collision regression, or retain the refresh gate for global subscriptions.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1383,7 +1457,16 @@ fn latest_kind9_message_tx(
                invalidation_status IS NULL
                OR (direction = 'sent' AND invalidation_status = 'local_publish_failed')
            )
-         ORDER BY timeline_at DESC, message_id_hex DESC
+         ORDER BY CASE
```
erskingardner: **[P1] Apply this failed-preview priority to every projection consumer**

This CASE fixes the normal rebuild, but the other latest-preview queries still use raw canonical order. `refresh_chat_list_last_message_after_secure_prune_tx` will select the class-2 failed send after any retention prune, restoring the pinned failed preview; the repeated subqueries in `chat_list_projection_complete_tx` also select it, so a row rebuilt with the delivered preview is considered stale on every `ensure_chat_list_rows` and triggers a full rebuild forever. Centralize this exact preview order and use it in normal rebuild, secure-prune refresh, and completeness checks. Please cover a delivered+failed group where pruning an unrelated row preserves the delivered preview and a subsequent ensure considers the projection complete.

erskingardner (changes requested):
The prior failed-send/read-anchor blocker is closed on this head: pre-publish projection does not advance the marker, the established-history regression now pins that behavior, and failed sends remain visible while accepted history wins the chat preview. The full CI matrix is green. Two cross-path correctness issues remain in the follow-up: global live deltas use an identity that is not group-scoped, and the new failed-preview priority is not used by secure-prune refresh or projection completeness.

Datawav (approved):

PR #1386, opened 2026-08-11
fix(marmot-app): audit epoch backfill lifecycle

### LINKED ISSUE #1183: marmot-app: audit epoch-gap backfill execution and recovery outcome
## Problem

The epoch-gap recovery audit trail stops at `epoch_stall_backfill_armed`. It does not identify when the pending replay actually starts, why it started, whether `activate_transport(None)` completed, or whether the stalled group advanced afterward.

This made the current field investigation unable to distinguish:

- an arm that remained dormant (the confirmed #1180 failure);
- a replay that started but failed;
- a successful account-wide transport replay that did not recover the affected group; and
- a replay that recovered the local epoch.

## Current behavior

Grounded against `b609ccd3796aea14bf0f0478073520179820d999`:

- `marmot-app/src/client/audit.rs:192-204` records only `EpochStallBackfillArmed { stalled_epoch, threshold }`.
- `marmot-app/src/client/sync.rs:489-500` runs `activate_transport(None)`, then clears the in-memory pending flag; it emits no recovery lifecycle or outcome row.
- `marmot-app/src/runtime/account_worker.rs:2060-2075` reports replay failure only through the runtime account-error channel.
- `subscription_rebuild` and `sync_drain` are account-scoped and carry no recovery cause or affected group, so temporal proximity is not reliable correlation when several sync/reconnect paths run.

#1180 fixes the liveness seam; it does not make the resulting recovery attempt auditable end-to-end.

## Required audit contract

Add typed, privacy-safe epoch-gap recovery lifecycle evidence with one opaque recovery-attempt ID:

1. `armed`: affected group, stalled local epoch, threshold, trigger source.
2. `started`: execution seam (`startup`, `receive`, `explicit_catch_up`, `maintenance`) and replay scope (`account_full_history` today).
3. `completed` or `failed`: duration, transport activation outcome/error kind, deliveries processed, local epoch before/after, and whether the affected group advanced.
4. If execution is deferred, record `deferred` with the reason; do not report it as completed.

The attempt ID must correlate these rows without embedding account/group/message identity. Do not log payloads, pubkeys, relay credentials, or free-form exception text.

## Acceptance criteria

- A no-later-traffic explicit CatchUp regression from #1180 produces exactly one `armed -> started -> completed|failed` attempt.
- Startup, normal receive, explicit CatchUp, and maintenance seams use the same lifecycle helper.
- A failed `activate_transport(None)` retains the pending recovery and records a typed failure; a later retry gets a new attempt or an explicit retry ordinal.
- Completion records the affected group's local epoch before and after replay; it must not claim group recovery merely because account-wide transport activation returned `Ok`.
- Goggles schema/ingest fixtures preserve and export all fields.
- Cardinality is one small lifecycle set per actual recovery attempt, not per replayed message.

## Related / non-goals

- #1180 owns executing armed recovery from every sync seam.
- This issue does not change the threshold-eight anti-abuse policy or replace account-wide replay with targeted replay.
- This issue does not expose an authoritative global consensus frontier; it records one engine's recovery attempt and observed local result.

agent-p1p:
Refs #1183

## Summary
- Add typed `armed` / `deferred` / `started` / `completed` / `failed` epoch-stall backfill audit rows to the shared v2 forensic schema, correlated by one opaque operation ID.
- Carry trigger, local epoch, execution seam, retry ordinal, activation outcome, privacy-safe failure kind, and bounded replay counters without emitting account/group identifiers.
- Preserve recovery intents armed during an in-flight replay when the current attempt fails; schedule both operation IDs independently and rotate deferred work fairly so unavailable newer work cannot starve an older retry.
- Debounce identical deferred evidence by exact reason / retry ordinal / armed-group epoch state, re-emitting only after a meaningful transition.
- Route startup, live receive, maintenance, convergence, and explicit catch-up through the same execution/reporting helper. Explicit full-history repair consumes an existing pending attempt without issuing a duplicate replay.
- Surface replay summaries through app events.

## Verification
- `just fast-ci`
- `cargo test -p marmot-forensics --locked` — 29 passed
- `cargo test -p marmot-app epoch_backfill --locked` — 8 focused tests passed across unit/integration targets
- `cargo test -p marmot-app explicit_catch_up_runs_prearmed_backfill_before_success_response --locked`
- `cargo test -p marmot-app full_history_repair_consumes_prearmed_backfill_without_replaying_twice --locked`
- `cargo test -p marmot-app --test epoch_stall_backfill_audit --locked` — 5 passed

Full `cargo test -p marmot-app --locked` passed all 613 unit tests, then reported six unrelated `relay_runtime` integration failures. Five reproduce identically on the clean pre-address-review head `1d67529e`; the sixth passed when rerun in isolation.

## Sensitive paths
- `crates/marmot-app/src/client/epoch_stall.rs`
- `crates/marmot-app/src/client/sync.rs`

These paths coordinate group-state recovery lifecycle. Explicit human merge approval is required.

## Release hygiene
- Updated existing `Unreleased` coverage in `crates/cli/CHANGELOG.md`.
- No workspace/package/conformance version changes.

## Cross-repo follow-up
Goggles schema/ingest/export coverage is tracked in marmot-protocol/goggles#338. That issue now contains the exact producer wire contract. This PR intentionally does not close #1183; keep #1183 open until the Goggles dependency lands.

## Summary by CodeRabbit

- **New Features**
  - Added reliable epoch-gap recovery with queued backfill attempts, retry handling, and deferred execution when groups are temporarily unavailable.
  - Added explicit catch-up support that prioritizes pending recovery work.
  - Added detailed audit records for recovery triggers, starts, completions, failures, retries, delivery results, and deferrals.

- **Bug Fixes**
  - Prevented duplicate deferred audit evidence unless group epoch state changes.
  - Improved preservation and fair processing of concurrent recovery operations.

- **Tests**
  - Expanded coverage for retries, failures, deferred recovery, ordering, and audit lifecycle tracking.


agent-p1p:

Reviewed exact head `1d67529e5901f25b6437dac09ce82997b62fc55b`.

### Blocking

1. **`crates/marmot-app/src/client/sync.rs:859-915` — a recovery armed during the replay can be silently overwritten on the replay's error path.**

   `begin_epoch_backfill_execution` removes the current intent with `take()`. While that execution is in `sync_sdk_relay` or `drain_pending_session_events`, either path can call `apply_backfill_decision` and install a new `pending_epoch_backfill` (for example, another group emits `TransportObjectResourceRefused`). If any later fallible step fails, `finish_epoch_backfill_execution` assigns `Some(execution.pending)` at line 915, replacing that newly armed intent. Its durable `armed` row is then left with no `started`/`failed`/`deferred` continuation, and the group has no pending retry—the exact dormant-arm ambiguity this issue is meant to eliminate.

   Preserve intents armed while a replay is in flight rather than replacing them. Add a regression that starts one pending replay, arms a second group during its drain, injects a post-arm failure, and proves both operation IDs remain terminalized/retryable.

2. **`crates/marmot-app/src/client/sync.rs:859-870` and `crates/marmot-app/src/runtime/account_worker.rs:1155-1163` — an unavailable group produces an unbounded stream of duplicate `deferred` rows.**

   `begin_epoch_backfill_execution` records `group_epoch_unavailable`, restores the unchanged pending value, and returns. Every subsequent receive/catch-up/maintenance helper invocation records the same `deferred` row again with the same operation ID and retry ordinal. A removed or long-quarantined group therefore emits one row per maintenance tick indefinitely, violating the issue's bounded-cardinality contract. Debounce a deferral per reason/state/ordinal (and re-enable it only after a meaningful state transition), with coverage that repeated seams do not multiply identical evidence.

3. **The Goggles ingest/export acceptance criterion is not complete, and the current tracker describes a different wire shape.**

   `marmot-protocol/goggles#338` is still open with no implementation PR. Its body names `observed_epoch`, `current_epoch`, `events_examined`, and `deliveries_observed`, and places `activation_outcome` on `started`; this PR instead emits `stalled_epoch`/`threshold`, `replay_scope`, `duration_ms`, `deliveries`, `local_epoch_before`/`after`, and `group_advanced`. Merging this PR with `Fixes #1183` would close the source issue before the explicit Goggles schema/fixture/export criterion is satisfied. Synchronize and land the final Goggles contract before closing #1183, or remove the closing reference and keep the dependency explicit.

### Verification

Passed locally at this head:

- `cargo test -p marmot-forensics --locked` — 29 passed
- `cargo test -p marmot-app epoch_backfill --locked`
- `cargo test -p marmot-app explicit_catch_up_arms_and_replays_without_later_traffic --locked`
- `cargo test -p marmot-app --test epoch_stall_backfill_audit --locked` — 5 passed
- `just fast-ci`

GitHub CI is green. These findings are lifecycle/error-path gaps not exercised by the current happy-path and activation-failure tests.

**Verdict:** 3 blocking findings, 0 suggestions, 0 nitpicks.

agent-p1p:
Address-review update pushed in `773402d3`.

- Failed replay A can no longer overwrite an arm B minted while A is in flight. Both operation IDs remain scheduled; deferred work rotates fairly so an unavailable B cannot starve A.
- Identical `deferred` evidence is debounced by exact reason / retry ordinal / armed-group epoch state, not counts.
- marmot-protocol/goggles#338 now carries the exact producer contract. The PR closing reference was removed so #1183 remains open until that dependency lands.

Evidence: `just fast-ci` passes; `cargo test -p marmot-app epoch_backfill --locked` passes all 8 matching unit/integration tests. The full crate run passed 613 unit tests; five unrelated relay-runtime failures reproduce on clean pre-address head `1d67529e`, and the sixth passed alone.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -2918,13 +2921,26 @@ async fn run_pending_epoch_backfill_reporting_arm(
...
-        let message = account_error_message("epoch-gap backfill failed", &err);
-        publish_app_runtime_account_error(events, account_id_hex, account_label, message.clone());
-        message
-    });
+    let result = match client.run_pending_epoch_backfill(seam).await {
+        Ok(Some(summary)) => {
+            publish_app_runtime_summary(events, account_id_hex, account_label, &summary);
+            Ok(())
+        }
+        Ok(None) => Ok(()),
```
Datawav, line 2932: `None` means the armed replay was deferred because at least one group epoch was unavailable, but this converts it to success for explicit CatchUp and startup waiters. Callers can therefore receive a successful response even though no backfill ran and recovery remains pending. Preserve a distinct deferred outcome and either perform ordinary catch-up or keep the response pending until replay executes; add an unavailable-group CatchUp regression test.

crates/marmot-app/src/client/sync.rs
```diff
@@ -800,10 +1127,17 @@ impl AppClient {
         self.relay_plane
             .set_transport_signer(self.transport_signer.clone())
             .await;
+        if self.has_pending_epoch_backfill() {
+            return self
+                .run_pending_epoch_backfill(EpochBackfillExecutionSeam::ExplicitCatchUp)
+                .await?
+                .ok_or_else(|| AppError::UnknownGroup(String::new()));
```
Datawav, line 1134: When a pending backfill is deferred, `run_pending_epoch_backfill` returns `None`; converting that into `UnknownGroup("")` aborts explicit full-history repair without issuing the ordinary account-wide replay and produces a misleading empty-group error. This can also skip a runnable intent rotated into the primary slot. Retain the deferred audit intent but fall through to the unfloored full-history replay, with coverage for unavailable groups and queued runnable intents.

Datawav (commented):
## Automated review

Two medium-severity behavioral regressions remain.

- **Medium — Catch-up can report success while recovery remains deferred.** In `crates/marmot-app/src/client/sync.rs`, changed `begin_epoch_backfill_execution()` restores an intent and returns `None` when any armed group epoch is unavailable; `run_pending_epoch_backfill()` converts that to `Ok(None)`. In `crates/marmot-app/src/runtime/account_worker.rs`, changed `run_pending_epoch_backfill_reporting_arm()` maps `Ok(None)` to `Ok(())`, which is returned to explicit CatchUp and startup waiters. Consequently, callers can receive success although no backfill ran and the recovery remains pending, contradicting the promised “before success response” behavior. Return a distinct deferred outcome and either perform the ordinary catch-up or keep the response pending until replay executes; add an unavailable-group CatchUp regression test.

- **Medium — Explicit full-history repair aborts instead of repairing a deferred intent.** In `crates/marmot-app/src/client/sync.rs`, changed `repair_full_history()` short-circuits whenever `has_pending_epoch_backfill()` is true and converts `run_pending_epoch_backfill(...).await? == None` into `AppError::UnknownGroup(String::new())`. The deferred path therefore performs no account-wide replay, emits a misleading empty-group error, and can fail indefinitely while the group epoch remains unavailable—even if deferral rotated another runnable intent into the primary slot. On deferral, retain the audit intent but fall through to the ordinary unfloored full-history replay; cover the unavailable-group and queued-runnable-intent cases.

crates/marmot-app/src/client/sync.rs
```diff
@@ -757,32 +824,292 @@ impl AppClient {
...
+    fn take_next_pending_epoch_backfill(&mut self) -> Option<PendingEpochBackfill> {
+        self.pending_epoch_backfill
+            .take()
+            .or_else(|| self.queued_epoch_backfills.pop_front())
+    }
+
+    fn requeue_failed_epoch_backfill_intent(&mut self, failed: PendingEpochBackfill) {
+        match self.pending_epoch_backfill.take() {
+            None => self.pending_epoch_backfill = Some(failed),
+            Some(current) if current.attempt_id == failed.attempt_id => {
```
erskingardner, line 839: Minor: this arm looks unreachable. During execution the pending slot was taken by `take_next_pending_epoch_backfill`, and any arm minted in flight creates a fresh `PendingEpochBackfill` with a new random `attempt_id`, so `current.attempt_id == failed.attempt_id` cannot hold. Per the repo's no-error-handling-for-impossible-scenarios convention, drop the arm (or document the concrete path that reaches it).

crates/marmot-app/src/client/sync.rs
erskingardner, line 1134: **Blocking: a deferred intent turns explicit full-history repair into a permanent, misleading failure.**

When the pending intent defers (`begin_epoch_backfill_execution` returns `None` because an armed group's local epoch is unobservable — e.g. the group was removed after arming), `run_pending_epoch_backfill` returns `Ok(None)` and this converts it into `Err(AppError::UnknownGroup(String::new()))`, which displays as `unknown local group`. Three problems:

1. The caller-directed repair performs **no work at all** — no unfloored activation, no drain — even though the ordinary repair below is a strict superset of the backfill's steps and would proceed fine.
2. The error is actively misleading (wrong category, empty group id) and is published to the host as `full-history repair failed: unknown local group`.
3. It fails **indefinitely**: the deferred intent stays pending (correctly), so every subsequent repair hits this same path until process restart.

On `Ok(None)`, fall through to the ordinary unfloored repair instead of erroring — the replay heals whatever is observable and the deferred intent keeps its `deferred` audit trail. Please add a regression test that arms an intent with an unobservable group and asserts `repair_full_history` still performs the repair and returns `Ok`.

(This confirms the automated review's second medium finding; it should block since it breaks an explicit user-facing repair flow.)

erskingardner (changes requested):
## Review — request changes (1 blocking, 1 process fix, 1 minor)

Reviewed exact head `773402d37c9f86e04b1a81314d82fdacd6243eeb`. CI is green.

The core design is solid and matches the #1183 contract and the wire shape pinned in marmot-protocol/goggles#338: typed `armed → started → completed|failed` plus debounced `deferred` rows correlated by one opaque operation id, in-flight arms preserved across failures with fair rotation, retry ordinals, per-group epoch before/after observation, and honest `group_advanced` semantics. The integration tests for the failure/rotation/debounce lifecycles are genuinely strong. Privacy posture (typed enums, counts, `privacy_safe_kind`, group identity only via the `group_ref` projection, in-memory-only debounce snapshot) checks out.

### Blocking

1. **`repair_full_history` aborts with a misleading `UnknownGroup("")` when the pending intent is deferred** (`crates/marmot-app/src/client/sync.rs:1130-1135`, inline comment). The explicit repair performs no work, reports `unknown local group`, and fails indefinitely while the intent stays deferred. Fall through to the ordinary unfloored repair on `Ok(None)` and add a regression test. This confirms the automated review's second medium finding; it should block because it breaks an explicit user-facing repair flow.

   On the automated review's first medium finding (CatchUp returning `Ok` while recovery is deferred): I would not block on this. The ordinary catch-up sync already ran and published its summary before the backfill helper executes, and the deferral emits a `deferred` row; failing the user's catch-up because a background recovery cannot run yet would be worse. Consider documenting that `Ok(None)` conflates "nothing pending" and "deferred" in the helper's contract, but the behavior is defensible.

### Process (must fix before merge)

2. **Commit `1d67529e` still carries `Fixes #1183` in its message body.** The PR body was corrected to `Refs #1183`, and goggles#338 explicitly requires #1183 to stay open until the Goggles ingest/export coverage lands. GitHub honors closing keywords in commit messages once they reach the default branch, and this repo allows merge commits and squash-with-commit-messages — either merge style would auto-close #1183. Reword the commit (or guarantee the squash message is edited) so the stated intent survives the merge.

### Minor

3. Unreachable defensive arm in `requeue_failed_epoch_backfill_intent` (`sync.rs:839`, inline comment).

### Verification

Relied on existing CI at head (all green, including Relay runtime tests, Simulator/vectors, Tamarin). I did not re-run the test suite locally; the PR documents the six unrelated `relay_runtime` local failures as reproducing on the pre-review head `1d67529e`, which is consistent with green CI.

Once the repair-deferral path falls through to the ordinary repair (with a regression test) and the commit message is reworded, this is good to merge.

Datawav (approved):

PR #1391, opened 2026-08-11
fix(marmot-app): preserve partial catch-up summaries

### LINKED ISSUE #1036: marmot-app: a mid-batch ingest error in sync_sdk_relay discards the whole batch SyncSummary — already-projected earlier deliveries are never emitted as live events
### Summary

`sync_sdk_relay` accumulates every delivery of a drain into a single `SyncSummary` and only returns it after the whole loop finishes. Each `ingest_delivery` **durably** commits engine state and writes the app-DB projection rows *before* the next delivery is processed, but a failure on a *later* delivery aborts the function with `?`, discarding the accumulated summary for all earlier, already-projected deliveries. Those messages are in the database but are never emitted onto the runtime `MarmotAppEvent` stream, so a host driving purely off the live event stream silently misses them.

### Location

- `crates/marmot-app/src/client/sync.rs:225-283` (`sync_sdk_relay` drain loop; `ingest_delivery(...).await?` at `:270-271`, `save_state` at `:281`)
- `crates/marmot-app/src/client/sync.rs:285-308` (`ingest_delivery` → `runtime.ingest_delivery` commit, then `observe_account_device_effects` with `?`)
- Consumed at `crates/marmot-app/src/runtime/account_worker.rs:396-478`: on `Ok(summary)` the worker calls `publish_app_runtime_summary` (`:451`); on `Err` it publishes only an account error (`:466-476`) and the summary is dropped.

```rust
// sync.rs — one summary for the whole batch:
let mut summary = SyncSummary::default();
loop {
    // ...
    self.ingest_delivery(delivery, &display_names, &mut summary).await?;  // durably projects, then can fail on delivery N
}
self.app.save_state(&self.state)?;   // skipped on error
Ok(summary)
```

### Failure scenario

An account's initial background catch-up (`account_worker.rs:397 client.sync()`) drains, say, 50 inbound deliveries. Each is committed to the engine and projected into the app DB (`record_account_app_event`, `remember_directory_message_sender`, `set_group_self_membership`), and each appends its contributions to the shared `summary`. Delivery 51 hits a transient `SQLITE_BUSY` (or a `set_group_self_membership` write error, or `fail_if_publish_failed`), so `ingest_delivery` returns `Err`, `sync_sdk_relay` returns `Err`, and:

1. `publish_app_runtime_summary` is never called → the 50 already-durably-projected messages are **never emitted as live `MarmotAppEvent`s**. They surface only if a host later re-snapshots via a `subscribe_*` call; a host consuming only the live stream after readiness misses them entirely.
2. `save_state` (`:281`) is skipped, so the in-memory `seen_events`/transport-cursor advances for the batch are **not persisted**, even though the engine/DB writes for deliveries 1..50 *are* durable — a durable-vs-in-memory divergence on the same path.

Note: `next_event` (the steady-state live tail, `sync.rs:180-223`) is **not** affected — it builds a fresh per-delivery summary and `save_state`s after each. The bug is specific to the batch `sync()` / `sync_sdk_relay` path (initial catch-up, the `account_worker.rs:664` sync, and the CLI `sync` command).

### Why not a duplicate

#966 is the distinct case where the *same* delivery is engine-durable/dedup-marked before *its own* projection fails. This is the sibling-delivery consequence: a failure on one delivery drops the live-stream emission of *earlier, fully-projected* deliveries in the same batch. Not #712 (a single dequeued delivery lost on cancellation), not #741 (head-of-line stalling), not #740/#761 (broadcast `Lagged`).

### Suggested severity

**MEDIUM** — no data corruption (rows are durable), but a live-delivery gap: messages the engine accepted and stored are never announced to a host relying on the runtime event stream, plus a durable/in-memory cursor divergence.

### Suggested fix

Flush the summary incrementally (emit per delivery, as `next_event` does) rather than accumulating a whole-batch summary discarded on the first mid-batch error; or catch per-delivery errors, record them, and continue the drain so already-projected deliveries still reach subscribers and `save_state` still runs. If the batch must abort, at least publish the summary accumulated so far and persist state before returning the error.

---
*Filed from a meticulous automated code-review pass. The drain/summary/worker flow was traced against source, and the finding was checked against existing open issues (#966, #712, #741, #740, #761) for duplicates before filing.*

agent-p1p:
## Summary

- persist seen-event, transport-cursor, and app-outbox progress after each completed catch-up delivery
- return the already-applied prefix with sync failures through the public `AppClient` API and both `wn sync` output modes
- publish partial runtime summaries before account errors and preserve audit-tracker scheduling for their messages/events/escalations
- cover the real N+1 ingest failure, durable save boundaries, route-changing deliveries, CLI rendering, and epoch-stall audit scheduling

## Test plan

- `just fast-ci`
- `cargo test -p marmot-app --test partial_sync_summary --features test-policy-overrides --locked -- --test-threads=1`
- `cargo test -p marmot-app --test epoch_stall_backfill_audit --features test-policy-overrides --locked escalation_from_completed_delivery_survives_later_sync_failure -- --nocapture`
- `cargo test -p wn-cli --lib --locked`

Fixes #1036

## Summary by CodeRabbit

* **Bug Fixes**
  * Sync failures now preserve successfully completed progress when a later delivery fails.
  * Completed deliveries remain saved and are not replayed after restarting.
  * Human-readable and JSON output now includes partial progress, error details, and processing summaries.
  * Live updates report completed deliveries before a later failure.
  * Escalation, repair, projection, and epoch-stall progress are included when available.
  * Group-join events are reliably delivered once and acknowledged without replay.

* **Tests**
  * Added coverage for partial failures, persistence, restart behavior, event delivery, error reporting, and escalation summaries.


agent-p1p:
BLOCKING:
- Direct `AppClient::sync` callers still lose the applied prefix: `crates/marmot-app/src/client/sync.rs:450-459`, `crates/cli/src/commands/sync.rs:16-17`. The new error path moves the accumulated `SyncSummary` into `pending_applied_sync_summary`, but that buffer is `pub(crate)` and only the managed account worker drains it. `wn sync` creates a one-shot client, propagates `client.sync().await?`, and then drops the client, so the earlier durable deliveries named in #1036 remain unreported on that explicitly affected path. The same is true for any external caller using the public `AppClient::sync` API directly, and a later successful `sync()` does not merge this stash back into its return value. Please expose a truthful partial-result/error contract for direct callers (and cover the CLI consumer), or explicitly split the issue so this PR does not close #1036 while that path remains unresolved.

SUGGESTION:
- Exercise the actual later-delivery error branch: `crates/marmot-app/src/client/sync.rs:617-648`, `crates/marmot-app/tests/partial_sync_summary.rs:97-128`. The regression injects failure immediately after the first delivery has completed and been saved; it proves the common preservation helper and event-before-error ordering, but it never starts a second `ingest_delivery` and therefore cannot catch a regression in the production `Err(error) => preserve_applied_summary_on_error(...)` branch at lines 621-622. A fault on delivery N+1 would match #1036's reported failure mode exactly.
- Preserve the existing audit-tracker scheduling invariant for partial summaries: `crates/marmot-app/src/runtime/account_worker.rs:665-688`, `crates/marmot-app/src/runtime/account_worker.rs:3046-3054`. Successful sync summaries containing messages/events call `sync_summary_triggers_audit_tracker_update`; each new error seam drains and publishes the retained summary without making that check, so the same durable prefix can reach live subscribers without scheduling its forensic tracker update. Returning the drained summary from the helper, or passing `RuntimeSharedServices` into it, would keep partial and successful reporting aligned.

NITPICK:
- none

SUMMARY: The runtime-worker fix is otherwise coherent: every completed delivery persists seen-id/cursor/application-event acknowledgements, all worker sync/repair error seams publish the retained prefix before `AccountError`, the focused regression proves ordering and restart non-replay, and I found no duplicate-emission path. Release hygiene is clean (`Unreleased` changelog entry present; no version, lockfile, fixture, or conformance churn), and no configured sensitive path is touched. I verified head `d85a46c559c74ec72e58c74900c1d727e6669a5f` with the focused test (1/1 passed), targeted clippy, and `git diff --check`. The remaining direct-client/CLI loss is part of #1036's stated affected surface, so this is not yet complete.

VERDICT: REQUEST_CHANGES

SENSITIVE_PATHS: none

agent-p1p:
Addressed the blocking review and both suggestions in 3a16ccdf:

- `AppClient::sync` and full-history repair now return `SyncFailure { summary, source }`; `wn sync` reports the durable prefix in both plain and JSON error output.
- The regression now fails delivery N+1 through the real ingest error branch. Additional tests pin save-boundary exclusion and route-changing seen-ID durability.
- Every runtime error seam publishes the partial summary and applies the same audit-tracker trigger as success. The epoch-stall test verifies the audit upload is scheduled after a later sync failure.

Verified with `just fast-ci`, 4 focused partial-sync regressions, the epoch-stall/audit regression, and all 542 `wn-cli` unit tests. No version or lockfile churn.

agent-p1p:
@coderabbitai review

crates/marmot-app/src/client/sync.rs
```diff
@@ -539,23 +547,27 @@ impl AppClient {
...
+        self.save_state_with_pending_local_group_deletion_frontier_clears()?;
+        self.drain_epoch_stall_escalations(&mut summary);
         let routes_changed = self.refresh_group_routes()?;
         if routes_dirty || routes_changed {
             self.sync_runtime_groups().await?;
         }
-        self.save_state_with_pending_local_group_deletion_frontier_clears()?;
-        self.drain_epoch_stall_escalations(&mut summary);
         Ok(summary)
     }
```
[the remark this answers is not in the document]
agent-p1p: Fixed in 3b3719f0. Membership-changing deliveries still persist before route reconciliation; every delivery now persists again after `refresh_group_routes()` so prior-route pruning is durable before the subscription refresh. The follow-up also removes the duplicated CLI message JSON mapping and documents the test drain assumption.

agent-p1p:
RE-REVIEW — APPROVE / LGTM

Verified exact head `3b3719f06ef02190e2cae138e313f4c17e709d96`. The original block and both suggestions are resolved:

- Direct callers and CLI: `AppClient::sync` now returns `SyncFailure { partial_summary, source }` (`crates/marmot-app/src/client/sync.rs:194`, `crates/marmot-app/src/lib.rs:685-711`). `wn sync` consumes that contract and reports the durable prefix in plain and JSON error output without replacing the source error (`crates/cli/src/commands/sync.rs:17-47`, `crates/cli/src/error.rs:226-234`).
- N+1 coverage: the catch-up loop calls the actual second `ingest_delivery` with its completed-delivery count and returns the already-merged prefix when that call fails (`crates/marmot-app/src/client/sync.rs:613-625`, `:692-702`). `direct_sync_failure_returns_applied_prefix` exercises this branch with delivery N completed and N+1 failing; the boundary-save regression separately excludes an incompletely persisted N+1.
- Audit scheduling: every runtime partial-failure seam now publishes through `publish_sync_summary_with_audit`, which uses the same visible-activity/escalation gate as success (`crates/marmot-app/src/runtime/account_worker.rs:2934-2956`). The runtime regression observes the tracker request, and the epoch-stall regression proves an escalation from N survives an N+1 failure.
- Route-state persistence: `ingest_received_delivery` preserves the pre-refresh save for membership-changing effects, then saves again after `refresh_group_routes` and before subscription refresh (`crates/marmot-app/src/client/sync.rs:552-563`). This persists prior-route pruning without weakening the durable membership boundary. CodeRabbit explicitly confirmed its route-state thread addressed.

Independent verification:
- `cargo test -p marmot-app --test partial_sync_summary --features test-policy-overrides --locked -- --test-threads=1` — 4 passed
- `cargo test -p marmot-app --test epoch_stall_backfill_audit --features test-policy-overrides --locked escalation_from_completed_delivery_survives_later_sync_failure -- --nocapture` — 1 passed
- `cargo test -p wn-cli --lib --locked` — 542 passed
- `just fmt-check` — passed
- targeted `marmot-app` and `wn-cli` clippy with `-D warnings` — passed
- `git diff --check` — passed

Release hygiene is clean: `Unreleased` changelog coverage is present; no version, lockfile, fixture, or conformance churn. Sensitive paths: none.

The final CodeRabbit status is rate-limited/hollow and is not treated as a fresh independent review. Its earlier route-state thread was explicitly rechecked and marked addressed.

BLOCKING: none
SUGGESTION: none
NITPICK: none

VERDICT: APPROVE / LGTM

This is review confirmation only. I did not merge.

crates/marmot-app/src/client/sync.rs
```diff
@@ -539,23 +547,30 @@ impl AppClient {
...
             .await?;
+        // A membership-changing ingest is already durable. Persist its app
+        // projection before route reconciliation or subscription refresh can
+        // fail, matching the catch-up delivery boundary below.
+        if routes_dirty {
+            self.save_state_with_pending_local_group_deletion_frontier_clears()?;
+        }
         let routes_changed = self.refresh_group_routes()?;
+        self.save_state_with_pending_local_group_deletion_frontier_clears()?;
+        self.drain_epoch_stall_escalations(&mut summary);
```
erskingardner: This drain now runs *before* the fallible `sync_runtime_groups().await?` three lines below, which contradicts the contract documented on `drain_epoch_stall_escalations` itself: "Call this as the LAST step before `Ok(summary)`... an interior `?` returns before the move runs, so the stash simply rides the next successful seam instead of being lost with the failed pass's summary."

Pre-PR order here was `refresh_group_routes?` → `sync_runtime_groups?` → `save_state?` → drain → `Ok`, so a subscription-registration failure left the escalation stashed for the next seam. With the new order, a `sync_runtime_groups` failure drops the summary — and the drained, one-shot-latched escalation with it. The managed worker masks this (it drops the client on a receive error either way), but direct `AppClient::next_event` callers keep their client across an error and would previously have had the escalation deferred rather than lost.

Moving `self.drain_epoch_stall_escalations(&mut summary);` after the `if routes_dirty || routes_changed { ... }` block restores the documented invariant at zero cost — the new early `save_state` boundaries above are unaffected.

crates/marmot-app/src/client/sync.rs
```diff
@@ -595,24 +610,75 @@ impl AppClient {
...
+                return Err(SyncFailure::new(summary, error));
+            }
+            // Escalations are accumulated as a side channel while observing
+            // this delivery. Move them into the delivery-local summary only
+            // after its persistence boundary succeeds, so N's escalation is
+            // returned when N+1 fails and an incomplete N is never reported.
+            self.drain_epoch_stall_escalations(&mut delivery_summary);
             deliveries = deliveries.saturating_add(1);
+            summary.merge(delivery_summary);
+            if delivery_routes_dirty && let Err(error) = self.sync_runtime_groups().await {
```
erskingardner, line 664: Non-blocking question: a route-changing delivery now triggers a full `sync_runtime_groups` re-registration inside the drain loop, so an initial catch-up that joins N groups performs N sequential registration passes (pre-PR: one accumulated `routes_dirty` flag → a single pass after the loop).

I see the rationale — the completed prefix's freshly joined groups are subscribed even when a later delivery fails the batch. But the error path recovers subscriptions on the next sync/startup anyway (`prepare_transport` + `sync_runtime_groups`), so accumulating `routes_dirty` and running one pass after the loop on the Ok path would avoid the O(N) registration churn for the restore-with-many-groups case, at the cost of the failed-prefix groups waiting one sync for their subscriptions.

Worth either the accumulation or a short comment here justifying the per-delivery pass.

erskingardner (changes requested):
Reviewed at head `3b3719f` against issue #1036. CI is green.

The core fix is sound and well-tested: the per-delivery boundary in `sync_sdk_relay` persists seen-id/cursor/app-outbox progress at the right point (after ingest, with the cursor rollback on boundary-save failure correctly restoring in-memory state to match the atomic SQLite boundary), escalations are gated behind the boundary so N's escalation survives N+1's failure, all four worker error seams publish the partial summary with the same audit-tracker trigger as the success path before reporting the account error, and the CLI preserves the source `AppError`'s JSON code/repair shape while adding the `partial` prefix. The regression tests exercise the real N+1 ingest branch, boundary-save exclusion, route-changing seen-id durability across restart, and live-event-before-error ordering. CodeRabbit's earlier actionable item (persist `prior_nostr_routes` after `refresh_group_routes`) is addressed at head, and the release hygiene is clean (Unreleased changelog, no version/lockfile churn).

One blocking item, a one-line move: `ingest_received_delivery` now drains the latched epoch-stall escalation stash *before* the fallible `sync_runtime_groups().await?`, so a subscription-registration failure on a direct `AppClient::next_event` call permanently drops an escalation the pre-PR ordering deferred — contradicting the invariant documented on `drain_epoch_stall_escalations` itself (inline comment). Plus one non-blocking question on per-delivery `sync_runtime_groups` registration churn in the catch-up loop (inline comment).

crates/marmot-app/src/client/sync.rs
agent-p1p: Fixed in `23026d36`. `drain_epoch_stall_escalations` now runs after the fallible subscription refresh, so a direct `next_event` caller retains the one-shot escalation for the next successful seam when refresh fails.

crates/marmot-app/src/client/sync.rs
agent-p1p: Documented the intentional per-delivery refresh in `23026d36`. It keeps subscriptions for the durable, reported N-prefix live even if N+1 fails; deferring to a later sync/startup would leave already-reported joined groups temporarily unsubscribed.

crates/marmot-app/src/client/sync.rs
```diff
@@ -595,24 +610,78 @@ impl AppClient {
...
+                    "injected catch-up boundary save failure".to_owned(),
+                ))
+            } else {
+                self.save_state_with_pending_local_group_deletion_frontier_clears()
+            };
+            if let Err(error) = boundary_failure {
+                // The SQLite boundary is atomic. Restore the in-memory seen id
+                // and transport cursor while retaining the state projections
+                // produced by the already-durable engine effects.
+                self.state = state_before_boundary;
```
Datawav, line 651: **High:** Restoring the pre-delivery cursor/seen state here does not roll back the engine effects already applied by `ingest_delivery`, and the corresponding `delivery_summary` is discarded. After restart the relay can redeliver the event, but engine deduplication may suppress those effects, leaving live consumers without the authenticated message or membership event. Persist a replayable summary/recovery marker alongside the engine application, then publish and clear it after the app-state boundary succeeds; cover save failure → reopen → resync with an exactly-once visibility test.

crates/marmot-app/src/client/sync.rs
```diff
@@ -189,32 +189,37 @@ impl AppClient {
         registration
     }
 
-    pub async fn sync(&mut self) -> Result<SyncSummary, AppError> {
+    /// Transport-first startup sync. All authenticated, newly-applied effects
```
Datawav, line 192: **Medium:** Changing this public method from `Result<SyncSummary, AppError>` to `Result<SyncSummary, SyncFailure>` is source-incompatible for downstream callers that annotate `AppError`, match its variants, or use `?` into `Result<_, AppError>`. `From<AppError> for SyncFailure` does not provide the reverse conversion. Preserve the existing `sync()` signature and expose partial-progress failures through a separately named API, then migrate the managed runtime and CLI to that API.

Datawav (commented):
## Automated review

Outcome: two actionable defects remain.

- **High — boundary-save failure can permanently suppress an applied delivery.** In `crates/marmot-app/src/client/sync.rs` (`@@ -595,24 +610,78`), `ingest_delivery` durably applies engine effects before the new state boundary is saved. When that save fails, lines 628–633 restore the pre-delivery cursor/seen state and return without merging `delivery_summary`. The relay then redelivers the event after restart, but engine deduplication can suppress its already-applied effects. Consequently, live consumers may never receive the authenticated message or membership event. The added boundary-failure test checks only immediate prefix exclusion, not recovery. Smallest sound correction: durably record a replayable delivery summary/recovery marker with the engine application, then publish and clear it after the app-state boundary succeeds; add a save-failure → reopen → resync test proving exactly-once visibility.

- **Medium — `AppClient::sync()` is source-incompatible for downstream callers.** At `crates/marmot-app/src/client/sync.rs:189`, the public return type changes from `Result<SyncSummary, AppError>` to `Result<SyncSummary, SyncFailure>`. Existing callers that annotate `AppError`, match its variants, or propagate directly into `Result<_, AppError>` stop compiling. `From<AppError> for SyncFailure` does not preserve the reverse conversion. Smallest sound correction: retain the existing `sync()` signature and add a separately named partial-progress API, migrating managed runtime and CLI callers to it.

Datawav (dismissed):

crates/marmot-app/src/client/sync.rs
```diff
@@ -226,7 +259,13 @@ impl AppClient {
         // above, `sync_sdk_relay` never drained the engine, so these would stay
         // buffered and invisible to runtime subscribers until some later
         // unrelated send/ingest. Fold any pending events into this summary.
-        let drained = self.drain_pending_session_events().await?;
+        let drained = match self.drain_pending_session_events().await {
+            Ok(drained) => drained,
+            Err(error) => {
+                self.drain_epoch_stall_escalations(&mut summary);
```
Datawav: This drains one-shot epoch-stall escalations into `SyncFailure::partial_summary`, but compatibility `sync()` immediately maps the failure to `failure.source` and discards that summary. A retained client can therefore permanently lose the recovery escalation after a failed sync. Only drain these escalations into failure summaries for partial-progress APIs; keep them pending when `sync()` fails, and add a same-client failure-then-success `sync()` regression test.

Datawav (commented):
## Automated review

Follow-up found one new high-severity defect; the previously reviewed persistence and API-contract fixes remain intact.

- Fixed: `AppClient::sync()` still returns `Result<SyncSummary, AppError>`, while `sync_with_partial_progress()` exposes the partial summary.
- Fixed: failed pre-persistence checkpoints retain uncommitted message/join progress for exactly-once replay after retry or restart.
- High — `crates/marmot-app/src/client/sync.rs` changed lines 201–233 and 261–271: `sync_inner()` drains one-shot epoch-stall escalations into `SyncFailure::partial_summary`, but compatibility `sync()` immediately maps that failure to `failure.source` and discards the summary. A retained client can therefore permanently lose the recovery escalation after a failed sync. Smallest correction: only drain escalations into failure summaries for partial-progress APIs; keep them pending when `sync()` fails, and add a same-client failure-then-success `sync()` regression test.

Datawav (dismissed):

PR #1393, opened 2026-08-11
fix(transport): stage group sync state before subscribe

### LINKED ISSUE #910: transport-nostr-adapter: sync_account_groups issues relay REQs for newly-added groups BEFORE registering their routes, so stored catch-up history can be dropped as unroutable
## Summary

`sync_account_groups` calls `self.subscribe_all("sync_account_groups", &to_add)` (relay REQs go out) **before** it commits the new groups into routing state (`record_subscription_starts` + `sync_groups`, which rebuilds `by_transport_group`). This is the exact reverse of `activate_account`, which deliberately registers routing/telemetry state **before** issuing REQs, with an inline comment explaining why.

## Where

- `crates/transport-nostr-adapter/src/lib.rs:601` — `subscribe_all(...)` for `to_add` (REQs issued)
- `crates/transport-nostr-adapter/src/lib.rs:608-616` — the `state.write()` that finally commits `record_subscription_starts(&to_add, ..)` + `sync_groups(...)` (routes registered)
- Contrast: `crates/transport-nostr-adapter/src/lib.rs:529-550` — `activate_account` registers state under `state.write()` first, then `subscribe_all`, with the comment:
  > "Register routing/telemetry state BEFORE the relay REQs go out: a relay may stream stored events the moment it sees a subscription, and an event arriving before the routes exist is dropped as unroutable — stored catch-up history would be lost with nothing to re-request it."

`subscribe_all` runs outside any lock, and `handle_relay_event` resolves routes under `state.read()`, so an inbound event for a just-added group that arrives between the REQ and the `state.write()` commit finds no entry in `by_transport_group` and is dropped with `Ok(0)`.

## Failure scenario

1. An account is already active. `sync_account_groups` adds group G (new `transport_group_id`).
2. The relay receives the REQ for G's `h`-tag filter and immediately streams stored kind-445 messages for G.
3. Those events reach `handle_relay_event` → `routes_for` → `by_transport_group.get(transport_group_id)` returns `None` (G not yet committed) → events dropped as unroutable.
4. The window widens under inbound load, because the sync's `state.write()` at line 610 contends the same `RwLock` that `handle_relay_event` / `observe_relay_event` / `handle_relay_eose` take.

Net effect: a newly-joined/added group can silently lose its initial stored-message catch-up on a sync — the very loss `activate_account` is written to avoid. Secondary consequence: `record_first_event` for those first events is also discarded (subscription not yet in `sync.subscriptions`), skewing sync telemetry.

## Note on the current ordering

The inline comment at lib.rs:598-600 justifies the current order as fail-fast ("on error the adapter state is untouched, a retry re-diffs against the old group set"). That goal can be preserved while still registering routes first: register the `to_add` routing/telemetry state under `state.write()` before `subscribe_all`, and on a `subscribe_all` error roll that state back (or record it as pending) — matching how `activate_account` orders the two steps.

## Severity

MEDIUM — a delivery/history-loss race on newly-added groups, distinct from #795 (RelayIndex keying mismatch for sync telemetry), #763 (HOL blocking on the delivery channel), and #720 (O(accounts) welcome scan).

agent-p1p:
## Summary
- Stage additive group routes and relay-sync telemetry before issuing relay REQs, so immediate stored-event replay is routable and measured.
- Commit staged telemetry only after the full subscription batch succeeds; roll it back on failure or caller cancellation without disturbing prior or unrelated progress.
- Keep still-live old routes available during replacement and cover deterministic subscription-ID reissues.

## Test plan
- `cargo test -p transport-nostr-adapter --locked`
- `just fast-ci`

Fixes #910

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved group subscription synchronization to handle stored-event replay without misclassifying events as unroutable.
  - Failed or canceled subscription attempts now roll back routing and telemetry changes, allowing safe retries.
  - Preserved existing routes, live callbacks, and synchronization metrics during subscription updates.

- **Tests**
  - Added coverage for replay handling, failed synchronization, cancellation cleanup, route preservation, and telemetry rollback.

### DIFF
--- a/crates/transport-nostr-adapter/src/telemetry.rs
+++ b/crates/transport-nostr-adapter/src/telemetry.rs
@@ -460,6 +460,17 @@ struct SubscriptionProgress {
     relays: HashMap<RelayIndex, RelayProgress>,
 }
 
+/// Subscription progress and aggregate samples buffered for one in-flight
+/// subscription batch. Keeping this separate from committed telemetry lets a
+/// failed batch roll back without erasing live-ID progress or unrelated relay
+/// callbacks that arrived concurrently.
+#[derive(Clone, Debug, Default)]
+struct StagedSubscriptionProgress {
+    subscriptions: HashMap<String, SubscriptionProgress>,
+    first_event: Vec<(RelayIndex, u64)>,
+    eose: Vec<(RelayIndex, u64)>,
+}
+
 /// Local recorder for subscription sync timing and the initial-sync gate.
 ///
 /// Diagnostic only; must never feed convergence or branch selection.
@@ -468,6 +479,7 @@ pub struct RelaySyncTelemetry {
     subscriptions: HashMap<String, SubscriptionProgress>,
     first_event: HashMap<RelayIndex, DurationHistogram>,
     eose: HashMap<RelayIndex, DurationHistogram>,
+    staged: StagedSubscriptionProgress,
 }
 
 impl RelaySyncTelemetry {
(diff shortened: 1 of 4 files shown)


agent-p1p:
## Pip adversarial review — APPROVE

Reviewed exact head `8d1a32c38e4058bd6a5c7f00b74777d0405a7df1`.

BLOCKING:
- none

SUGGESTION:
- none

NITPICK:
- none

SUMMARY: The fix closes the pre-REQ race by staging additive routing coverage and relay-sync progress before `subscribe_all`, while preserving committed old routes until the batch succeeds. Failure and cancellation rebuild routing from the authoritative account state and discard only staged telemetry; successful batches commit the overlay before route replacement. The focused replay, deterministic-ID reissue, rollback, and cancellation tests exercise the relevant windows. Changelog coverage is present under `Unreleased`; there is no version, lockfile, conformance, or fixture churn.

VERDICT: APPROVE

SENSITIVE_PATHS: `crates/transport-nostr-adapter/src/lib.rs`, `crates/transport-nostr-adapter/src/telemetry.rs`, `crates/transport-nostr-adapter/tests/inbound_routing.rs`

Verification:
- `cargo test -p transport-nostr-adapter --locked` — 60 passed
- `cargo clippy -p transport-nostr-adapter --all-targets --features sdk --locked -- -D warnings` — passed

agent-p1p:
@coderabbitai review

erskingardner (dismissed):
Reviewed at head `8d1a32c3`. Approve — no blocking or inline findings.

**Correctness trace (all verified against source at head, not just the diff):**

- **Ordering fix matches issue #910's recommended shape**: routes + telemetry are staged under `state.write()` before `subscribe_all`, committed only after the full batch succeeds, and rolled back on failure — closing the window where a relay's immediate stored-event replay was dropped as unroutable, while preserving the old fail-fast retry semantics (a retry re-diffs against the old group set).
- **Cancellation safety is sound**: there are no await points between staging (lib.rs:783-787) and arming the detached guard (lib.rs:793), so staged state can never exist without the guard. Every exit resolves the oneshot exactly once: success sends after the commit+drain, failure rolls back in-task then sends, cancellation drops the sender and the guard rolls back. Rollback is idempotent (post-commit `mem::take` leaves staged empty; index rebuild derives from authoritative `accounts`), so cancel-during-commit or cancel-during-failure-rollback cannot tear state.
- **Lock handoff is race-free**: the guard task holds the lifecycle lock until its rollback completes, so a follow-on lifecycle call cannot re-stage before cleanup finishes — the `debug_assert`s in `begin_staged_subscription_starts` are unreachable by construction.
- **Reissued deterministic subscription IDs** (retained prior-route case in `diff_group_subscriptions`): the staged overlay shadows committed progress during the window, commit overwrites, and rollback leaves the still-live committed entry untouched — exactly the property a naive record-then-forget approach would break, and it is pinned by both the unit tests and `failed_group_sync_rolls_back_staged_routes_and_telemetry`.
- **No duplicate delivery during the window**: `stage_group_routes` filters endpoints already covered (canonical-URL-aware) for the same account+group, so a reissued or endpoint-overlapping route cannot double-deliver; commit/failure rebuild restores the index wholesale.

**Tests**: the four new regression tests cover the reported race (synchronous replay mid-subscribe for both new and still-live old routes), reissue, failure rollback with telemetry equality, and caller cancellation — good signal density, and CI is fully green including the sdk feature matrix.

**Conventions**: changelog entry is in the right cohort file, tracing stays aggregate-only per the observability invariant, no version bumps, scope is surgical.

Non-blocking observation (no action needed): during an in-flight batch that reissues a still-live id, `subscription_synced`/`snapshot()` report the fresh REQ's progress rather than the prior synced state. That's arguably the more truthful signal for the new REQ and is diagnostic-only, just noting the visible-behavior change.

Datawav (commented):
## Automated review

One actionable correctness gap remains.

- **Medium — reissued subscription callbacks can regress or lose committed sync state.** In `telemetry.rs`, `record_first_event` and `record_eose` route callbacks exclusively to `staged.subscriptions` and return whenever the deterministic ID is staged; `subscription_synced` and `subscription_any_eose` likewise prefer staged state, while `rollback_staged_subscription_starts` discards it entirely. The changed `reissued_live_subscription_keeps_synchronous_callbacks` test confirms staged and committed entries can share an ID when a live group is reissued with a new `since` cursor, contradicting the claim that this overlap is unreachable. Consequently, a previously synchronized live subscription temporarily reports unsynchronized while its replacement REQ is pending; if the committed entry was not yet synchronized and EOSE arrives during a subsequently failed or cancelled batch, that EOSE is discarded and the retained live subscription’s initial-sync gate can remain false indefinitely. Preserve committed gate state for overlapping IDs and retain staged callback effects in a rollback overlay: apply them to committed progress only on rollback, discard the overlay on successful commit. Add a blocked-subscribe regression covering both an in-flight committed-true gate and an initially-unsynchronized committed entry receiving EOSE before rollback.

crates/transport-nostr-adapter/src/telemetry.rs
```diff
@@ -481,56 +493,97 @@ impl RelaySyncTelemetry {
...
 
     /// Record EOSE from `relay` for `subscription_id`. Repeat EOSE and unknown
     /// subscription/relay pairs are ignored.
     pub fn record_eose(&mut self, subscription_id: &str, relay: RelayIndex, now_ms: u64) {
-        if let Some(progress) = self
-            .subscriptions
-            .get_mut(subscription_id)
-            .and_then(|sub| sub.relays.get_mut(&relay))
-            && !progress.eose_seen
+        if self.staged.subscriptions.contains_key(subscription_id) {
```
Datawav, line 603: **Medium:** When a reissued subscription ID exists in both staged and committed state, this branch records EOSE only in the staged entry and returns. During that window, the committed live subscription can regress from synced to unsynced; if the batch then fails or is cancelled, rollback discards the staged EOSE, potentially leaving an initially-unsynchronized retained subscription gated indefinitely. Preserve committed gate state for overlapping IDs and retain staged callback effects in a rollback overlay that is applied to committed progress only on rollback. Add blocked-subscribe coverage for both an already-synced committed entry and an unsynced committed entry receiving EOSE before rollback.

PR #1395, opened 2026-08-11
test(convergence): preserve app-runtime cross-route counterexample

erskingardner:
## Summary

- add action-addressed retained-relay controls to the simulator app-runtime adapter
- preserve a four-party cross-route Scenario IR and compare it with the strict retained-engine reference
- record the resulting app-runtime equivalence falsification and reopen the affected assurance claims

## What this demonstrates

The strict retained-engine reference reaches exact canonical equality, records the expected accepted/invalidated/accepted commit dispositions, drains pending work, and preserves all 12 active decryptability edges.

The production-shaped app-runtime adapter does not yet reproduce that result. Repeated executions have produced three exactly characterized failure forms:

- Zeta remains one epoch behind while the complete probe set is visible;
- public protocol state agrees, but Alpha's post-settlement probe is missing and non-invalidated at Zeta; or
- Alpha remains on its competing root, Observer remains at the baseline, and only Yankee/Zeta reach the retained reference branch, with probes split along those branch boundaries.

The characterization accepts only those three exact counterexamples. A fourth divergence shape or complete equivalence fails so the test and assurance claims must be reviewed together.

## Review hardening

- capture successful relay admissions in actual publication order instead of event-id order
- associate immediate group-message and Welcome publications with all applicable app commands
- document deferred-publication selector limits and relay-wide removal semantics
- pin the intermediate cached-epoch boundary and accurate retained-memory implementation metadata

## Scope

- simulator, scenario documentation, and convergence tracking documentation only
- no production convergence, app, protocol, or storage behavior changes
- diagnosis and correction of the production behavior will be a separate focused PR
- equivalent controlled staging for process, container, and VM adapters remains follow-up work

## Verification

- `just fast-ci`
- `cargo test -p cgka-conformance-simulator --test process_orchestrator four_party_cross_route_recovery_records_app_runtime_equivalence_falsification --locked -- --nocapture`
- `cargo test -p cgka-conformance-simulator --test app_runtime_adapter --locked`
- `cargo test -p cgka-conformance-simulator --lib app_runtime --locked`
- `cargo fmt --all --check`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added controls for hiding and restoring shared events during recovery.
  * Added a four-participant recovery scenario with restarts, selective synchronization, and state verification.
  * Added tracking for published relay events and their associated actions.

* **Documentation**
  * Documented cross-route recovery findings, limitations, decryptability, application delivery, and settlement behavior.

* **Tests**
  * Added regression coverage comparing recovery outcomes across execution paths.
  * Added validation for relay visibility, offline operations, and event restoration safeguards.

### DIFF
--- a/crates/cgka-conformance-simulator/src/app_runtime.rs
+++ b/crates/cgka-conformance-simulator/src/app_runtime.rs
@@ -15,7 +15,10 @@ use marmot_app::{
     AccountSetupRequest, AppError, AppMessageQuery, MarmotApp, MarmotAppConfig, MarmotAppEvent,
     MarmotAppRuntime,
 };
-use nostr_relay_builder::MockRelay;
+use nostr_relay_builder::LocalRelay;
+use nostr_relay_builder::prelude::{
+    Event, Filter, MemoryDatabase, MemoryDatabaseOptions, NostrDatabase, RelayBuilder,
+};
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use tempfile::TempDir;
@@ -114,17 +117,25 @@ impl Participant {
 
 /// In-process application harness backed by one real local Nostr relay.
 pub struct AppRuntimeHarness {
-    _relay: MockRelay,
+    _relay: LocalRelay,
+    relay_database: MemoryDatabase,
     relay_url: String,
     participants: BTreeMap<String, Participant>,
     scenario_groups: BTreeMap<String, GroupId>,
     active_scenario_group: Option<String>,
     accepted_publications: BTreeMap<String, BTreeSet<String>>,
+    relay_action_events: BTreeMap<String, Vec<Event>>,
+    hidden_relay_events: BTreeMap<String, Event>,
 }
 
 impl AppRuntimeHarness {
     pub async fn new(clients: &[String]) -> Result<Self, SubjectError> {
-        let relay = MockRelay::run().await.map_err(environment_error)?;
+        let relay_database = MemoryDatabase::with_opts(MemoryDatabaseOptions {
+            events: true,
+            max_events: Some(75_000),
+        });
+        let relay = LocalRelay::new(RelayBuilder::default().database(relay_database.clone()));
+        relay.run().await.map_err(environment_error)?;
         let relay_url = relay.url().await.to_string();
         let endpoint = TransportEndpoint::from(relay_url.clone());
         let mut participants = BTreeMap::new();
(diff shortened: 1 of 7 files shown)


crates/cgka-conformance-simulator/tests/process_orchestrator.rs
```diff
@@ -528,6 +895,169 @@ async fn app_runtime_and_process_adapters_recover_the_same_offline_projection()
...
+            app_payloads.is_subset(&expected_payloads),
+            "{client} projected an unexpected application payload"
+        );
+        assert_eq!(app.application.visible_plaintexts.len(), app_payloads.len());
+        all_active_probes_visible &= app_payloads == expected_payloads;
+        assert!(app.application.invalidated_message_ids.is_empty());
+        assert!(!app.application.pending_confirmation);
+    }
+    assert!(
+        !protocol_equivalent || !all_active_probes_visible,
```
erskingardner: This characterization passes for *any* protocol divergence or *any* missing probe — not just the two failure forms the docs enumerate (Zeta one epoch behind; or protocol agreement with Alpha's probe absent at Zeta). If the app runtime regresses in a third shape (e.g. everyone stalls at the baseline epoch, two probes vanish, or the divergence lands on a different participant), this assertion still passes and the new failure is silently absorbed into the "known falsification" record — the opposite of the explicit-counterexample discipline this test is meant to enforce.

Could we pin the two enumerated forms, e.g. compute `zeta_one_epoch_behind` (protocol not equivalent, Zeta exactly one epoch behind the reference, the other three match, all probes visible) and `missing_probe_after_agreement` (protocol equivalent, exactly Alpha's post-settlement probe absent at Zeta), then assert their disjunction? That keeps the deliberate tripwire for full equivalence while forcing review on any *new* divergence shape.

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -684,6 +822,8 @@ impl ConvergenceSubject for AppRuntimeHarness {
             )
             .await
             .map_err(app_error)?;
+        self.record_relay_action_events(action.action_id, before)
```
erskingardner: `record_relay_action_events` is only wired into `update_group_data`, but the adapter declares `SubjectCapability::RetainedRelayControl`, and the strict `RetainedRelaySubject` resolves action-id selectors for *every* published artifact. A scenario that selects a `create_group` / `update_admin_policy` / `send_app_message` event by action id passes whole-schedule preflight against this adapter and then fails with `relay_event_not_found` only here. Either wire the same before/after recording into the other publishing commands (looks mechanical) or document the restriction near the capability/error so scenario authors are not surprised.

Related: for actions that publish more than one event, `selector.occurrence` indexes events ordered by event id (arbitrary across runs), while the strict subject's occurrence is publish order. Fine for the single-event commits used in this scenario, but worth a comment if `occurrence > 0` ever becomes reachable on this adapter.

crates/cgka-conformance-simulator/tests/process_orchestrator.rs
```diff
@@ -528,6 +895,169 @@ async fn app_runtime_and_process_adapters_recover_the_same_offline_projection()
...
+        .map(|observation| (observation.client.as_str(), observation))
+        .collect::<BTreeMap<_, _>>();
+    let app_routed = app_trace[clients.len()..clients.len() * 2]
+        .iter()
+        .map(|observation| (observation.client.as_str(), observation))
+        .collect::<BTreeMap<_, _>>();
+    assert_eq!(app_routed["zeta"].epoch, 4);
+    assert_eq!(app_routed["alpha"].epoch, 3);
+    assert_eq!(app_routed["yankee"].epoch, 4);
+    assert_eq!(app_routed["observer"].epoch, 3);
```
erskingardner, line 1015: The alpha/observer pins read `cached_epochs`, which for those two clients was last refreshed at the pre-split tick (step 11): alpha reconnects at step 23 and commits `alpha-root` at step 24 (its true local epoch is 4) but never ticks before this Observe, and observer has been offline since step 15. So `alpha == 3` / `observer == 3` hold by construction of the schedule rather than as independent evidence of the split — the zeta/yankee pins (both synced/ticked after `zeta-root`) are the ones carrying the weight. A short comment noting these two read pre-split cached epochs would keep a future reader from over-trusting the pin. (Ticking alpha before the mid Observe would pin its fork position for real, at the cost of renumbering the positional `step-N` action ids.)

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -114,17 +117,25 @@ impl Participant {
...
+    hidden_relay_events: BTreeMap<String, Event>,
 }
 
 impl AppRuntimeHarness {
     pub async fn new(clients: &[String]) -> Result<Self, SubjectError> {
-        let relay = MockRelay::run().await.map_err(environment_error)?;
+        let relay_database = MemoryDatabase::with_opts(MemoryDatabaseOptions {
+            events: true,
+            max_events: Some(75_000),
+        });
```
erskingardner, line 202: Nit: `events: true, max_events: Some(75_000)` exactly mirrors `RelayBuilder::default()`'s own database options in nostr-relay-builder 0.44, so this preserves the previous `MockRelay` behavior while exposing the database handle. A one-line comment to that effect would stop the next reader wondering whether 75k was tuned against scenario volume.

erskingardner (commented):
Scope is as advertised: simulator and docs only, no production convergence/app/protocol/storage changes. The `MockRelay` → `LocalRelay` switch is behavior-preserving (the explicit `MemoryDatabaseOptions` mirror `RelayBuilder::default()` exactly in nostr-relay-builder 0.44), and the hide/restore mechanism is sound: offline participants have their runtimes shut down, so deleting from the shared database only affects future queries, and the schedule guarantees every online client has already ingested an event before it is hidden. The strict-reference assertions (exact equality, accepted/invalidated/accepted dispositions, twelve decryptability edges, no pending work) line up with the scenario construction, and the route/claim drift-test invariants in `tests/route_assurance.rs` still hold against the edited matrix. CI is green across the board, including Simulator and vectors.

Inline feedback, in priority order:

1. The final characterization assertion accepts any divergence shape, not just the two documented failure forms — tightening it keeps a new failure mode from passing silently as the known falsification. This is the one I would most like addressed here or in a fast follow.
2. `RetainedRelayControl` is declared broadly, but action-event recording only covers `update_group_data`; generalize or document the restriction.
3. The mid-route alpha/observer epoch pins read pre-split cached epochs and hold by construction; a comment would prevent over-trusting them.
4. Nit: note the 75_000 provenance.

One non-blocking semantic note: this adapter implements `SetRelayEventVisibility` as global removal guarded by named-clients-offline, whereas the strict subject hides per client (`hidden_from`). They coincide for this schedule and PROPERTY_TESTS.md describes the removal semantics accurately, but a sentence contrasting it with SCENARIO_IR.md's per-client visibility would help future portable-scenario authors.

None of these block the value of landing the counterexample and reopening the claims.

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -175,9 +256,136 @@ impl AppRuntimeHarness {
...
+            return Ok(());
+        }
+        if self.removed_relay_events.contains_key(&event_id) {
+            return Err(SubjectError::classified(
+                SubjectFailureCategory::ExpectedRefusal,
+                "relay_event_already_removed",
+                "the selected event is already removed from the shared relay",
+            ));
+        }
+        if clients.iter().any(|client| {
```
Datawav: This validates only the caller-supplied `clients` subset before deleting the event from the single shared relay database. If an omitted participant is online and has not fetched the event, the relay-wide deletion removes its access without checking that precondition, so the harness can claim a controlled delivery topology it did not enforce. Require every harness participant to be offline, or track delivery and require each omitted participant to have already fetched the event; add a refusal test for an omitted online, unfetched participant.

crates/cgka-conformance-simulator/tests/process_orchestrator.rs
```diff
@@ -528,6 +895,233 @@ async fn app_runtime_and_process_adapters_recover_the_same_offline_projection()
...
+        );
+        let non_epoch_equivalent = app.protocol.member_count == retained.member_count
+            && app.protocol.group_name == retained.group_name
+            && app.protocol.group_description == retained.group_description;
+        let client_protocol_equivalent =
+            app.protocol.epoch == retained.epoch && non_epoch_equivalent;
+        protocol_equivalent &= client_protocol_equivalent;
+        protocol_equivalent_by_client.insert(client.clone(), client_protocol_equivalent);
+        protocol_non_epoch_equivalent_by_client.insert(client.clone(), non_epoch_equivalent);
+        assert_eq!(app.protocol.member_count, 4, "{client} roster size");
```
Datawav, line 1062: The equivalence characterization verifies only the roster size (and separately the administrators), while the accepted-shape predicates also omit `member_identities`. A four-member roster containing an unexpected identity can therefore pass. Add an exact assertion that `app.protocol.member_identities` equals `['alpha', 'observer', 'yankee', 'zeta']`.

Datawav (commented):
## Automated review

Two actionable findings.

- **Medium — Global relay deletion checks only the caller-supplied client subset.** In `app_runtime.rs`, the new removal path checks `clients.iter()` for offline status, then deletes the selected event from the single shared `relay_database`. An omitted online participant that has not fetched the event loses access despite never being validated, so scenarios can assert a delivery topology the harness did not enforce. Require every harness participant to be offline before deletion, or track delivery and require each omitted participant to have fetched the event; add a refusal test for an omitted online, unfetched participant.

- **Low — The new equivalence test never verifies exact roster membership.** In `process_orchestrator.rs`, the added per-client checks assert `member_count == 4` and the administrator identities, while every accepted-shape predicate likewise omits `member_identities`. A four-member roster containing an unexpected identity can therefore pass as equivalent. Add an exact assertion that `member_identities` equals `["alpha", "observer", "yankee", "zeta"]`.

PR #1083, opened 2026-07-23
fix: persist frozen outbound fanout lifecycle

agent-p1p:
## Summary
- durably freeze the exact signed transport message, target set, acknowledgement policy, and per-target state before the first send
- persist monotonic target outcomes and atomically couple the first positive acknowledgement or all-failed result to MLS confirm/rollback
- restore pending fanout after restart using original bytes and routes, while exposing URL-free MLS/fanout completion outcomes

## Verification
- just fast-ci
- cargo test -p cgka-traits -p storage-sqlite -p cgka-engine -p cgka-session -p marmot-account
- cargo test -p marmot-account --test runtime frozen_fanout_
- cargo test -p cgka-engine --test publish_lifecycle

Fixes #1020

## Summary by CodeRabbit

* **New Features**
  * Added durable outbound fanout tracking for multi-endpoint message delivery.
  * Added fanout outcomes showing accepted, failed, and outstanding deliveries.
  * Added automatic retry and resume support for interrupted deliveries, including after restart.
  * Added coordinated confirmation and rollback for group changes and message fanout operations.

* **Bug Fixes**
  * Improved recovery from crashes and partial delivery attempts.
  * Prevented invalid or regressive delivery-state updates from overwriting saved progress.

* **Reliability**
  * Ensured delivery progress and group-state updates are saved consistently, reducing duplicate sends and lost progress.

PR #1397, opened 2026-08-12
perf(marmot-account): batch relay fanout into one concurrent publish

dannym-arx:
A publish's relay fanout previously went out one relay at a time: `drive_outbound_fanout` looped over endpoints, publishing to each with `required_acks: 1` and awaiting the ack before the next call. With per-relay budgets of 5 s connect plus 12 s publish wait across 3 attempts, one bad relay could stall a send for minutes.

This marmot now dispatches to all pigeons at once: every outstanding endpoint is published concurrently as its own single-endpoint request with `required_acks: 1`, joined with `futures::future::join_all`. That is the exact per-endpoint adapter contract the old loop used, so endpoint outcomes stay independent; only the one-awaited-ack-at-a-time serialization is gone. The first version of this PR sent one multi-endpoint batch with `required_acks` set to the attempt count; review caught that the production adapter turns an unmet `required_acks` into a whole-call error that discards successful receipts, which would have read an ordinary partial acceptance as total failure and rolled back a commit that had already reached a relay.

- `drive_outbound_fanout` marks every outstanding endpoint `Attempting`, persists the frozen record once, publishes them concurrently, then applies each endpoint's own `Ok`/`Err` outcome and persists once.
- `finish_transport_fanout`, the retry-remainder path, uses the same shape.
- The dead single-endpoint target helpers are removed.

Fanout no longer waits sequentially: its duration can approach the slowest individual attempt's budget when the transport makes concurrent progress, instead of the sum of per-relay budgets. The common case is one relay round trip.

Tests: the recording adapter gains persistent endpoint-keyed accept/error policies (deterministic under concurrent calls, unlike FIFO knobs), the fanout matrix covers first/middle/last/none acceptance plus whole-call adapter errors, a new regression test covers the production partial-accept shape (one relay accepts while siblings return whole-call errors, and pending MLS state still confirms), and the crash-resume test asserts resume re-attempts every outstanding target over the frozen route while ignoring changed routing.

## Summary by CodeRabbit

* **Performance Improvements**
  * Relay and group-message delivery now publishes to multiple endpoints concurrently, reducing delivery delays.
  * Retry and recovery operations process outstanding destinations together rather than sequentially.

* **Reliability**
  * Individual endpoint results remain tracked during fanout, including partial successes and failures.
  * Improved crash recovery preserves delivery progress and safely resumes outstanding destinations.
  * Duplicate recovery attempts avoid repeating already completed deliveries.
  * Delivery continues correctly across terminal, failed, and partially accepted fanout attempts.

### DIFF
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -22,8 +22,9 @@ versioning through the workspace version in the root `Cargo.toml`.
 
 - One publish's relay fanout now runs its per-relay attempts concurrently
   instead of one awaited relay at a time, so a slow relay no longer serially
-  delays every relay after it. The worst case drops from the sum of per-relay
-  budgets to the slowest single relay's budget.
+  delays every relay after it. Fanout duration can approach the slowest
+  individual attempt's budget when the transport makes concurrent progress,
+  instead of the sum of per-relay budgets.
   ([#1397](https://github.com/marmot-protocol/mdk/pull/1397))
 
 - App-runtime encrypted-media secret caching now skips the MLS group load,
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -20,10 +20,10 @@ versioning through the workspace version in the root `Cargo.toml`.
   the active blob remains memory-resident and one media request is bounded to
   512 MiB total.
 
-- One publish's relay fanout now goes out as a single concurrent batch instead
-  of one awaited relay at a time, so a slow relay no longer serially delays
-  every relay after it. The worst case drops from the sum of per-relay budgets
-  to one overall fanout budget per publish.
+- One publish's relay fanout now runs its per-relay attempts concurrently
+  instead of one awaited relay at a time, so a slow relay no longer serially
+  delays every relay after it. The worst case drops from the sum of per-relay
+  budgets to the slowest single relay's budget.
   ([#1397](https://github.com/marmot-protocol/mdk/pull/1397))
 
 - App-runtime encrypted-media secret caching now skips the MLS group load,
--- a/crates/cli/CHANGELOG.md
+++ b/crates/cli/CHANGELOG.md
@@ -19,6 +19,13 @@ versioning through the workspace version in the root `Cargo.toml`.
   upgrade to this compatibility cohort before downloading blobs above 64 MiB;
   the active blob remains memory-resident and one media request is bounded to
   512 MiB total.
+
+- One publish's relay fanout now goes out as a single concurrent batch instead
+  of one awaited relay at a time, so a slow relay no longer serially delays
+  every relay after it. The worst case drops from the sum of per-relay budgets
+  to one overall fanout budget per publish.
+  ([#1397](https://github.com/marmot-protocol/mdk/pull/1397))
+
 - App-runtime encrypted-media secret caching now skips the MLS group load,
   exporter-secret derivation, and database write when the current epoch's
   secret is already cached, and the cache sweep's `required` pre-check uses
--- a/crates/marmot-account/src/runtime.rs
+++ b/crates/marmot-account/src/runtime.rs
@@ -2489,49 +2489,53 @@ where
             .filter(|target| transport_fanout_target_retry_due(target, self.wall_clock.now()))
             .map(|target| target.endpoint.clone())
             .collect::<Vec<_>>();
-        for endpoint in remaining {
-            let target = publish_target_for_endpoint(&fanout.target, endpoint.clone());
+        if !remaining.is_empty() {
+            // One adapter call fans out to every retry-due endpoint
+            // concurrently; `required_acks` is the attempt count so the
+            // adapter does not abort the remainder once one relay accepts.
+            let target = publish_target_with_endpoints(&fanout.target, remaining.clone());
             match self
                 .adapter
                 .publish(TransportPublishRequest {
                     account_id: self.session.self_id(),
                     message: fanout.exact_message.clone(),
                     target,
-                    required_acks: 1,
+                    required_acks: remaining.len(),
                 })
                 .await
             {
                 Ok(report) => {
-                    apply_report_to_fanout(
-                        &mut fanout,
-                        std::slice::from_ref(&endpoint),
-                        &report,
-                        self.wall_clock.now(),
-                    );
-                    if !report.met_required_acks() {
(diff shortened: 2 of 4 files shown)


crates/marmot-account/src/runtime.rs
```diff
@@ -2726,31 +2730,54 @@ where
...
+                    }
+                }
+                Err(_) => {
+                    // A whole-call adapter error is systemic (e.g. signing or
+                    // preparation), not endpoint-specific; per-endpoint
+                    // failures come back inside an `Ok` report.
+                    for &index in &outstanding {
+                        fanout.mark_target_failed(index)?;
+                    }
                 }
```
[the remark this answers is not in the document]
dannym-arx: Verified against the code; not applying this here, for three reasons.

1. Not a regression of this PR. The pre-batch loop produced the same terminal state: a whole-call adapter `Err` on each per-endpoint call marked that endpoint failed, and an all-failed fanout hit the same zero-accept rollback in `resolve_outbound_fanout_mls`. Batching collapses N identical `Err`s into one.

2. The suggested edit does not compile. `possible_exposure` lives on the legacy `DurableTransportFanout`; the frozen `OutboundFanout` has no such field. In the legacy path that flag exists to skip rollback after an adapter error (`runtime.rs`, the `!commit_status.possible_ambiguous_exposure` gate). The frozen lifecycle (#1083/#1098) handles ambiguity differently: the durable `Attempting` edge plus resume republishing the exact frozen bytes.

3. The underlying window is real but pre-existing and belongs to the lifecycle design, not this perf change: a whole-call `Err` can postdate external exposure (for example `deliver_local_publish` failing after relay acceptance inside the relay-plane adapter), and the frozen path then terminalizes and rolls back. The conservative alternative, leaving targets `Attempting` so resume retries them, trades that fork risk for a group stuck in pending publish when the error is systemic (unsendable payload). That tradeoff deserves its own issue rather than a rider on a batching PR.

crates/marmot-account/src/runtime.rs
```diff
@@ -2726,31 +2730,54 @@ where
...
-                target: single_endpoint_target(&fanout.request().target, endpoint.clone()),
-                required_acks: 1,
+                target: publish_target_with_endpoints(
+                    &fanout.request().target,
+                    outstanding
+                        .iter()
+                        .map(|&index| endpoints[index].clone())
+                        .collect(),
+                ),
+                required_acks: outstanding.len(),
```
erskingardner: **[P1] Preserve partial acknowledgements from the real adapter**

`NostrSdkRelayClient::finish_publish_outcome` returns `Err` whenever `accepted.len() < required_acks`, and `TransportPublishFailure` carries only endpoint failures, not the successful receipts. With `required_acks = outstanding.len()`, an ordinary partial result (for example, 1 of 3 relays accepts) therefore enters the `Err` arm below, marks all three targets `Failed`, and lets `resolve_outbound_fanout_mls` roll back a commit that was already published. The new `RecordingAdapter::accept_endpoints_next` test masks this because it returns `Ok(TransportPublishReport)` even when the report does not meet `request.required_acks`. The same contract mismatch affects the legacy batch at line 2503. Please separate “wait for every endpoint attempt” from the acknowledgement success threshold (or otherwise return the full accepted/failed outcome after an all-target batch), and add coverage that exercises the production adapter's partial-accept behavior.

erskingardner (changes requested):
The batching direction is sound, but the production adapter currently converts partial acceptance into an error that discards the successful receipts. That makes the frozen-fanout path roll back MLS state after external publication. Please preserve endpoint-level outcomes through the batched call and cover the production adapter semantics.

crates/marmot-account/src/runtime.rs
dannym-arx: Good catch, and confirmed: `finish_publish_outcome` errs whenever `accepted.len() < required_acks` and the error carries only failure reasons, so the attempt-count batch read partial acceptance as total failure. Fixed in 6a2038ad by dropping the multi-endpoint batch entirely: each outstanding endpoint now publishes concurrently as its own single-endpoint request with `required_acks: 1` (the exact contract the old sequential loop used, joined with `join_all`), so endpoint outcomes stay independent and a sibling's error can no longer poison an accepted relay. `finish_transport_fanout` got the same shape, which also resolves the line 2503 instance.

On coverage: the recording adapter now uses persistent endpoint-keyed accept/error policies instead of FIFO knobs (deterministic under concurrent calls), and `frozen_fanout_partial_accept_with_sibling_adapter_errors_confirms_mls` exercises the production shape directly: one relay accepts while its siblings return whole-call errors, and pending MLS state still confirms.

crates/cli/CHANGELOG.md
```diff
@@ -19,6 +19,13 @@ versioning through the workspace version in the root `Cargo.toml`.
   upgrade to this compatibility cohort before downloading blobs above 64 MiB;
   the active blob remains memory-resident and one media request is bounded to
   512 MiB total.
+
+- One publish's relay fanout now runs its per-relay attempts concurrently
+  instead of one awaited relay at a time, so a slow relay no longer serially
+  delays every relay after it. The worst case drops from the sum of per-relay
+  budgets to the slowest single relay's budget.
```
Datawav: This overstates the guaranteed latency improvement. `join_all` concurrently polls the per-relay futures, but it does not establish independent transport progress or exclude shared adapter/connection bottlenecks. Please say fanout now avoids sequential waiting and can approach the slowest individual attempt's duration when the transport supports concurrent progress.

Datawav (commented):
## Automated review

Outcome: One low-severity actionable finding.

- Low — `crates/cli/CHANGELOG.md:22-24` claims worst-case fanout latency “drops from the sum of per-relay budgets to the slowest single relay’s budget.” The runtime delta uses `join_all`, which concurrently polls per-relay futures, but the changed code does not establish independent transport progress or exclude shared adapter/connection bottlenecks. This overstates the guaranteed performance behavior. Smallest correction: say fanout now avoids sequential waiting and can approach the slowest individual attempt’s duration when the transport supports concurrent progress.

crates/cli/CHANGELOG.md
dannym-arx: Fair: join_all gives concurrent polling, not a transport-level guarantee. Reworded in ab86775c using your suggested framing, in both the changelog entry and the PR description.

erskingardner (approved):

PR #1398, opened 2026-08-12
Fix reversible relay history in convergence simulator

erskingardner:
## Summary

- replace destructive relay-event deletion with a reversible query-visibility layer in the app-runtime simulator
- prove hidden history stays unavailable until restored, including limit-aware relay queries
- retain the strict retained-engine reference and characterize the production-shaped app route against corrected relay input
- keep route equivalence, active decryptability, and complete application disposition open after repeated testing retained a genuine counterexample

## Root cause and corrected finding

The simulator used the Nostr memory database delete operation to model temporary relay invisibility. That operation permanently tombstones the event id. The attempted later save was rejected, but the harness ignored the rejection and treated the event as restored. The app participants therefore received genuinely incomplete history.

The reversible visibility layer fixes that input-contract bug. However, the reviewer-requested repetition disproved the proposed positive-equivalence rewrite: corrected-input executions can reach full equivalence or retain a two-branch result. In the observed split, Alpha and Observer remained at epoch 4 on `alpha-root`, Yankee and Zeta reached epoch 5 on `zeta-branch-depth-two`, and the active probes split along those branches despite repeated full-history repair. The PR therefore preserves a tightly characterized counterexample instead of claiming the production route is fixed.

## Scope and impact

This PR changes only the conformance simulator, simulator tests, and convergence documentation. It does not change the convergence engine, app runtime, transport adapter, protocol, wire format, or production constants.

## Verification

- `cargo test -p cgka-conformance-simulator recording_database_preserves_successful_relay_admission_order --locked`
- `cargo test -p cgka-conformance-simulator --test app_runtime_adapter --locked` (9/9)
- `cargo test -p cgka-conformance-simulator --test process_orchestrator four_party_cross_route_recovery_characterizes_corrected_app_runtime_outcomes --locked`
- 20/20 consecutive characterization executions passed on commit `2461b874`
- the attempted equivalence-only repetition failed and produced the retained two-branch counterexample described above
- `just fast-ci`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recovery when relay history is temporarily unavailable while preserving retained events.
  - Restored event visibility now works correctly after clients reconnect and repair history.
  - Prevented temporarily hidden history from being permanently removed.

- **Reliability**
  - Strengthened recovery validation across varied network and application states.
  - Improved handling and reporting of invalidated or temporarily unavailable messages.

- **Documentation**
  - Updated convergence and recovery guidance to reflect current behavior and remaining coverage areas.


Datawav (commented):
## Automated review

No actionable findings.

erskingardner (commented):
The root-cause diagnosis is right and the fix is the correct shape: `MemoryDatabase::delete` permanently tombstones an event id, so "temporarily remove then restore" was never a reversible operation, and the three recorded terminal shapes were input-contract artifacts. Replacing it with a query-visibility filter is the minimal correct change, the limit-after-filter detail is handled correctly (`Events::new(&filter)` bounds with `OverCapacityPolicy::Last`, so the newest visible events survive rather than being truncated by hidden ones), and the new focused hide/repair/restore/repair regression pins the contract that was previously only assumed. Scope is disciplined: simulator, simulator tests, and convergence docs only, no production behavior touched.

Two things I would want resolved before this lands, both about the evidence rather than the code.

**1. The dropped `invalidated_message_ids` assertion leaves application disposition unchecked on the app route.** This PR simultaneously rewrites `CONVERGENCE_ROUTE_MATRIX.md` to say `complete_application_disposition` now has passing evidence. But `visible_plaintexts` is filtered to kind-9 chat only, so the exact-set assertion says nothing about dispositions, and the assertion that was removed was the only thing that did. See the inline comment: a cross-participant equality check on the invalidated set is cheap and restores the property the docs now claim. Every sibling cross-adapter test in this file still asserts on `invalidated_message_ids`; this one now asserts nothing about it.

**2. Single-run evidence for flipping three assurance claims from falsified to passing.** The prior contract existed precisely because *repeated* executions produced three different terminal shapes. This PR converts that into strict equality (`app.protocol.epoch == retained.epoch`, exact payload set) on the strength of one run of each listed command plus one green CI pass. The app half of the scenario has no `AwaitQuiescence`; it relies on fixed Sync/Tick rounds while the runtime still has deferred background publications (the same scheduled self-update that `PROPERTY_TESTS.md` documents as not action-addressable). If any residual timing nondeterminism survives, the result is a flaky test guarding a claim the repo treats as reviewed evidence. `CONVERGENCE_ROUTE_MATRIX.md` itself says a falsification "must be resolved by explicit reviewed evidence" - please run `four_party_cross_route_recovery_reaches_app_runtime_equivalence` 20-30 times consecutively and record the count in the PR body and the 2026-08-12 ledger row. If it is stable, that is exactly the evidence the claim rewrite needs; if it is not, the claim should stay open.

Remaining notes are minor and inline.

crates/cgka-conformance-simulator/tests/process_orchestrator.rs
```diff
@@ -1050,92 +1046,37 @@ async fn four_party_cross_route_recovery_records_app_runtime_equivalence_falsifi
...
+        assert_eq!(
+            app_payloads, expected_payloads,
+            "{client} application payloads"
         );
         assert_eq!(app.application.visible_plaintexts.len(), app_payloads.len());
-        app_payloads_by_client.insert(client.clone(), app_payloads);
-        assert!(app.application.invalidated_message_ids.is_empty());
+        // The superseded profile commit correctly leaves invalidated kind-1210
+        // system rows. The exact visible chat set above proves that no witness
+        // or active probe was withdrawn or left without a disposition.
```
erskingardner: Replacing `assert!(app.application.invalidated_message_ids.is_empty())` with a prose comment removes the only disposition check on the app route, while the docs in this same PR promote `complete_application_disposition` to passing evidence.

The comment asserts a property the test can no longer see: `invalidated_message_ids` is a list of hex ids, so nothing here verifies those rows are kind-1210 system rows rather than a withdrawn chat message. And `visible_plaintexts` is filtered to `MARMOT_APP_EVENT_KIND_CHAT`, so the exact-payload assertion above cannot cover it either. As written, all four participants could report completely different invalidated sets and this test would still pass - which is exactly the route-equivalence property being claimed.

Cheapest fix that keeps the kind-1210 allowance and restores a real check is cross-participant equality, e.g. capture the first client's set and compare:

```rust
let mut expected_invalidated: Option<&Vec<String>> = None;
// ...inside the loop:
match expected_invalidated {
    None => expected_invalidated = Some(&app.application.invalidated_message_ids),
    Some(expected) => assert_eq!(
        &app.application.invalidated_message_ids, expected,
        "{client} invalidated dispositions"
    ),
}
```

If the invalidated count is deterministic, an exact `assert_eq!(len, N)` would be stronger still.

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -129,22 +130,58 @@ impl NostrDatabase for RecordingRelayDatabase {
...
+            let mut database_filter = filter;
+            database_filter.limit = None;
+            let events = self.inner.query(database_filter).await?;
+            let hidden_event_ids = self.hidden_event_ids.read().await;
+            Ok(events
+                .iter()
+                .filter(|event| !hidden_event_ids.contains(&event.id))
+                .take(limit)
+                .count())
+        })
```
erskingardner: This duplicates `query`'s logic and can drift from it. Since `query` already returns an `Events` capped at `filter.limit`, this whole body collapses to a delegation that is equivalent by construction:

```rust
fn count(&self, filter: Filter) -> BoxedFuture<'_, Result<usize, DatabaseError>> {
    Box::pin(async move { Ok(self.query(filter).await?.len()) })
}
```

Same result (`min(visible, limit)`), 12 lines deleted, and `count`/`query` can no longer disagree about what "hidden" means.

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -105,6 +105,7 @@ struct Participant {
 struct RecordingRelayDatabase {
     inner: MemoryDatabase,
     publication_log: Arc<Mutex<Vec<Event>>>,
+    hidden_event_ids: Arc<RwLock<BTreeSet<EventId>>>,
```
erskingardner: Minor: the sibling `publication_log` uses `Mutex` and there is no read concurrency to win here (the set is tiny and touched a handful of times per scenario). `Arc<Mutex<BTreeSet<EventId>>>` would match the existing field and drop the extra `RwLock` import.

crates/cgka-conformance-simulator/tests/process_orchestrator.rs
```diff
@@ -1050,92 +1046,37 @@ async fn four_party_cross_route_recovery_records_app_runtime_equivalence_falsifi
...
-            && exact_roster
-            && app.protocol.group_name == retained.group_name
-            && app.protocol.group_description == retained.group_description;
-        let client_protocol_equivalent =
-            app.protocol.epoch == retained.epoch && non_epoch_equivalent;
-        protocol_equivalent &= client_protocol_equivalent;
-        protocol_equivalent_by_client.insert(client.clone(), client_protocol_equivalent);
-        protocol_non_epoch_equivalent_by_client.insert(client.clone(), non_epoch_equivalent);
-        assert_eq!(app.protocol.member_count, 4, "{client} roster size");
+        assert_eq!(app.protocol.epoch, retained.epoch, "{client} epoch");
```
erskingardner: Minor: now that the surrounding checks are all `assert_eq!`, the `exact_roster` temporary a few lines up is used exactly once and is separated from its assertion by three unrelated asserts. Inlining it reads better and removes a line:

```rust
assert_eq!(
    app.protocol.member_identities,
    ["alpha", "observer", "yankee", "zeta"],
    "{client} exact roster"
);
```

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -1420,7 +1421,8 @@ incorrect result.
...
+| 2026-08-11 | 6.1 app-runtime cross-route apparent falsification | Added action-addressed event staging to the app-runtime harness's real retained Nostr relay and preserved three non-equivalent terminal shapes. Follow-up validation found that the staging mechanism used destructive database deletion: the Nostr memory database permanently tombstoned each event id, the unchecked restore was rejected, and app participants therefore received incomplete history. The recorded shapes remain useful evidence of the simulator input-contract failure, not production convergence counterexamples. | `cross-route-app-runtime-recovery/v1`; superseded expected-falsification contract; route-assurance claims require corrected evidence |
```
erskingardner: This rewrites an existing dated ledger row rather than appending to it. The 2026-08-12 row below already carries the correction and reads well on its own; editing the 2026-08-11 row's summary after the fact makes the ledger no longer a record of what was believed on that date, which is the main thing a dated evidence table is for.

Suggest leaving the 2026-08-11 row as originally written and letting the new row supersede it - optionally with a short "superseded by 2026-08-12" pointer appended, so the trail stays append-only.

crates/cgka-conformance-simulator/tests/process_orchestrator.rs
```diff
@@ -1091,50 +1117,77 @@ async fn four_party_cross_route_recovery_records_app_runtime_equivalence_falsifi
...
+                "probe-from-yankee",
+                "probe-from-zeta",
+                "zeta-branch-witness",
+            ])
+        && app_payloads_by_client["zeta"] == app_payloads_by_client["yankee"];
+
     assert!(
-        zeta_one_epoch_behind || missing_probe_after_agreement || split_protocol_and_probe_surface,
-        "app-runtime route result was neither documented counterexample shape; protocol_match={protocol_equivalent_by_client:#?}; payloads={app_payloads_by_client:#?}; observations={app_observations:#?}"
+        fully_equivalent
```
erskingardner, line 1184: The accept-set is wider than the corrected-input evidence supports.

The PR body reports two shapes across 20 corrected-input runs: complete equivalence, and the Alpha/Observer-versus-Yankee/Zeta split (`corrected_split`). The other four disjuncts - `zeta_one_epoch_behind`, both `missing_at_zeta_after_agreement` shapes, and `legacy_split` - are the three terminal surfaces this PR attributes to the destructive-delete bug. `PROPERTY_TESTS.md` on the base branch called them "simulator artifacts caused by genuinely incomplete relay history"; the input that produced them no longer exists.

Concretely, `legacy_split` requires Observer to sit at its baseline epoch on group name `cross-route` - never having received anything. Under a working reversible-visibility layer that state should not be reachable, so accepting it means a regression in `RecordingRelayDatabase::query` or in `set_shared_relay_event_presence` that silently drops history again would land in a passing bucket here. The four-party test would stay green and only the focused `retained_relay_control_restores_hidden_history_for_offline_repair` test would catch it.

Suggest cutting the disjunction to `fully_equivalent || corrected_split`. That keeps the counterexample retained as the matrix requires, is what the 20/20 run actually evidences, and makes any legacy shape reopen review the way `CONVERGENCE_ROUTE_MATRIX.md` says a changed outcome should. It also deletes roughly 40 lines (`zeta_one_epoch_behind`, `missing_at_zeta_after_agreement`, `legacy_split`, and the now-unused `protocol_non_epoch_equivalent_by_client` map).

If you believe the legacy shapes are still reachable under corrected input, that is worth stating explicitly in `PROPERTY_TESTS.md` with the run that produced one, since the docs currently read as though the corrected input eliminated them.

crates/cgka-conformance-simulator/tests/process_orchestrator.rs
```diff
@@ -1074,10 +1077,33 @@ async fn four_party_cross_route_recovery_records_app_runtime_equivalence_falsifi
...
+    // no losing system row has been projected, or Alpha alone has projected
+    // the invalidated kind-1210 row from its losing profile branch.
+    let no_invalidated_rows = clients
+        .iter()
+        .all(|client| invalidated_counts_by_client[client] == 0);
+    let alpha_losing_system_row = clients
+        .iter()
+        .all(|client| invalidated_counts_by_client[client] == usize::from(client == "alpha"));
+    assert!(
+        no_invalidated_rows || alpha_losing_system_row,
```
erskingardner, line 1098: Thanks for restoring a real check here - pinning Alpha specifically is stronger than the cross-participant equality I suggested.

One gap remains: `no_invalidated_rows` accepts the case where the superseded profile commit produced no invalidated row on any participant. That is the "left without a disposition" outcome the removed assertion's replacement comment was originally worried about, so between the two disjuncts the disposition property is still not asserted - only that the count is 0 everywhere or 1 at Alpha.

If both shapes are genuinely observed, fine as characterization, but tie each to the terminal shape it belongs to rather than leaving them free-floating. `alpha_losing_system_row` presumably pairs with the runs where Alpha stays on `alpha-root`, and `no_invalidated_rows` with the equivalent runs (or vice versa). Folding the expected count into `fully_equivalent` and `corrected_split` would make it an actual disposition assertion instead of an either-or.

Also worth noting in the comment: `invalidated_message_ids.len()` cannot distinguish a kind-1210 system row from a withdrawn chat message, so the count shape leans on the kind-9-filtered payload assertion above to rule that out. That coupling is currently implicit.

crates/cgka-conformance-simulator/tests/process_orchestrator.rs
```diff
@@ -1060,7 +1059,11 @@ async fn four_party_cross_route_recovery_records_app_runtime_equivalence_falsifi
         protocol_equivalent_by_client.insert(client.clone(), client_protocol_equivalent);
         protocol_non_epoch_equivalent_by_client.insert(client.clone(), non_epoch_equivalent);
         assert_eq!(app.protocol.member_count, 4, "{client} roster size");
```
erskingardner, line 1061: Minor: this duplicates the `app.protocol.member_count == retained.member_count` term inside `non_epoch_equivalent` a few lines up, given `retained.member_count` is 4 for every participant (the strict expectations pin `member_count: 4`). One of the two can go - I would keep this hard `assert_eq!` and drop the term from `non_epoch_equivalent`, so a roster mismatch fails loudly rather than silently flipping a participant out of the equivalence bucket.

crates/cgka-conformance-simulator/src/app_runtime.rs
```diff
@@ -129,22 +130,47 @@ impl NostrDatabase for RecordingRelayDatabase {
...
+            visible.extend(
+                events
+                    .into_iter()
+                    .filter(|event| !hidden_event_ids.contains(&event.id)),
+            );
+            Ok(visible)
+        })
     }
 
     fn delete(&self, filter: Filter) -> BoxedFuture<'_, Result<(), DatabaseError>> {
```
erskingardner, line 176: Minor, and the same class of bug this PR fixes: `delete` and `wipe` pass straight through to `inner` without touching `hidden_event_ids`. If a scenario ever wipes or deletes while an event is hidden, the id is stranded in the set, and the later `set_shared_relay_event_presence(.., visible = true)` reports success on an event the database no longer holds - the harness would again believe it restored history it did not.

No current scenario exercises that path, so this is not a live defect. But since the whole point of this PR is that a silent restore failure produced three bogus counterexamples, having `wipe` clear the hidden set (and `delete` prune it) is a couple of lines that closes the hole by construction rather than by convention.

erskingardner (commented):
Re-review of the updated head (`aff59a0`). Note this landed as `85ef573b` before the re-review, so everything below is follow-up rather than merge-blocking.

**The two things I asked for came back well, and one of them came back better than I expected.** The repetition request did its job: instead of confirming the equivalence rewrite, it disproved it, and the PR now says so plainly in the body, `CONVERGENCE_ROUTE_MATRIX.md`, `SCENARIOS.md`, `PROPERTY_TESTS.md`, and the plan. Reopening three assurance claims you had already written as passing, on the strength of your own repetition run, is exactly the discipline the route matrix asks for and the harder of the two available choices. The 2026-08-11 ledger row is left untouched with a new 2026-08-12 row appended, so the trail stays append-only. `count` now delegates to `query` (12 lines gone, the two can no longer disagree), `RwLock` is gone, and the roster assertion is inlined. Scope is still simulator, simulator tests, and docs only.

The remaining concern is in the shape of the new oracle, not the fix.

**The accept-set is wider than the evidence.** The final assertion admits six terminal shapes. The PR body reports exactly two from 20 corrected-input runs: complete equivalence, and the Alpha/Observer-versus-Yankee/Zeta split. The other four (`zeta_one_epoch_behind`, both `missing_at_zeta_after_agreement` shapes, `legacy_split`) are the surfaces this same PR attributes to the destructive-delete input bug it just fixed. Carrying them forward into the corrected-input accept-set means a regression that reintroduces incomplete relay history lands in a passing bucket in the four-party test. The focused hide/repair/restore/repair regression is what actually guards the fix today, and it does that job well - which is the argument for narrowing the big test to the two shapes you can evidence.

Secondary: the disposition check is back, which I appreciate, but `no_invalidated_rows` still admits the case where nothing was recorded at all.

Details inline. Both are follow-up items; neither undoes the fix.

Issue #836, opened 2026-07-12
Hermes plugin: MarmotPlatformAdapter.connect() missing is_reconnect kwarg → platform never connects


jooray:
## Summary

`MarmotPlatformAdapter.connect()` in the Hermes plugin is declared `async def connect(self) -> bool:`, but current `hermes-agent` calls the adapter as `adapter.connect(is_reconnect=...)`. The result is a `TypeError` on every connect attempt, so the Marmot platform never comes up.

## Environment

- Plugin: `wn-agent-v0.9.4` (`integrations/hermes/marmot/adapter.py`); also present on `master` (`adapter.py:1300`).
- hermes-agent: recent build whose base adapter/gateway pass `is_reconnect`.

## Root cause

`gateway/run.py` connects adapters via `await adapter.connect(is_reconnect=is_reconnect)`, and the base adapter + the **bundled SimpleX plugin** both use the signature `async def connect(self, *, is_reconnect: bool = False) -> bool:`. The Marmot adapter's override drops the keyword-only argument.

## Repro

Enable the Marmot platform on a current hermes-agent gateway. Startup logs:

```
✗ marmot error: MarmotPlatformAdapter.connect() got an unexpected keyword argument 'is_reconnect'
Reconnect marmot error: ... next retry in 60s
```

The platform never connects; no socket is opened to `wn-agent`.

## Fix

Widen the signature to match the base class / SimpleX adapter:

```python
async def connect(self, *, is_reconnect: bool = False) -> bool:
```

The body doesn't need to use the argument. One-line change. PR to follow.

PR #1399, opened 2026-08-12
Expose message reactions to agents

### LINKED ISSUE #862: No message-reaction primitive in agent-control protocol — app layer already supports it (AppMessageIntent::Reaction), just not exposed to agents
## Summary
Hermes agents have no way to react to a message with an emoji (Telegram's 👀→👍/👎 "processing status" reaction pattern, or a plain acknowledgement reaction). The app layer already fully supports reactions — the gap is that the primitive was never exposed through the agent-control protocol.

## What's confirmed in the code

**App layer already has it** — `crates/marmot-app/src/messages/intents.rs:163`, `AppMessageIntent::Reaction { .. }`, handled at `mod.rs:249` (validates non-empty emoji) with an `Unreact` counterpart documented as resolving to a `Delete` of the reaction event. This is a mature, tested feature used by the app's own long-press reaction UI.

**`agent-control` protocol has no equivalent** — the full `AgentControlRequest` enum (`crates/agent-control/src/lib.rs`) covers `Send`, `SendAgentActivity`, `SendAgentOperationEvent`, `SendGroupSystemEvent`, `SendMedia`, `DownloadMedia`, `GroupInfo`, allowlist ops, and debug ops — no `SendReaction`/`Unreact` variant anywhere.

**Hermes side has no override either** — `integrations/hermes/marmot/adapter.py` has no `send_reaction`/reaction-related method, consistent with there being no protocol primitive to call.

## Why this is smaller than the reply-threading and typing gaps
Unlike those two, the actual reaction logic already exists end-to-end at the app layer (`AppMessageIntent::Reaction`/`Unreact`) — this is genuinely just a protocol-exposure + adapter-plumbing task, not new reaction semantics. The main open question is scope, not feasibility:
- Expose full arbitrary-emoji reactions (matches the in-app long-press picker), or
- Start narrower with a fixed processing-status set (👀 on receipt → ✅/❌ on completion, mirroring the Telegram adapter's existing status-reaction convention) to avoid the base `BasePlatformAdapter` needing a new generic `send_reaction(emoji: str)` surface right away.

## Suggested shape
- `AgentControlRequest::SendReaction { account_id_hex, group_id_hex, target_message_id_hex, emoji }` / `RemoveReaction { .. }`, mapping directly onto the existing `AppMessageIntent::Reaction`/`Unreact`.
- Hermes adapter: override `send_reaction`/`remove_reaction` (check `BasePlatformAdapter` for the existing generic hook other platforms use, e.g. Telegram's status-reaction call site, and match that signature) to call the new primitive.

## Acceptance criteria
- [ ] `AgentControlRequest` gains a reaction add/remove variant wired to the existing `AppMessageIntent::Reaction`/`Unreact`.
- [ ] Hermes marmot adapter implements the corresponding `send_reaction`/equivalent hook.
- [ ] Decide + document scope: arbitrary emoji vs. fixed processing-status set for v1.

Datawav:
## Summary

- expose arbitrary add/remove message reactions through `marmot.agent-control.v2`
- route reaction requests through `agent-connector` to the existing `AppMessageIntent::Reaction` / `Unreact` behavior
- expose Hermes lifecycle and agent-facing reaction APIs, including latest-inbound-message targeting
- document the v1 scope and add wire-shape, connector, and Hermes adapter regression coverage

Closes #862

## Verification

- `just fast-ci`
- `python3 -m unittest integrations.hermes.marmot.tests.test_adapter`
- `git diff --check`

## Review notes

- Independent implementation review found and prompted addition of the public Hermes `add_reaction` / `remove_reaction` API expected by `send_message(action="react"|"unreact")`.
- The required Kimi K3 advisory review was attempted through the shared runner but OpenRouter returned HTTP 402 before producing a verdict; no Kimi approval is claimed.
- `just hermes-dev-script-test` currently reaches an unrelated existing installer assertion failure: `Hermes installer dry-run did not describe active-service upgrade behavior`. This patch does not modify installer scripts.

## Summary by CodeRabbit

* **New Features**
  * Added support for adding and removing reactions on messages.
  * Reactions can target a specific message or the latest inbound message.
  * Supports removing a specific reaction or all reactions from the caller.
  * Repeated identical reactions are handled idempotently.
* **Bug Fixes**
  * Improved handling of missing or invalid reaction targets.
  * Restricted reactions to non-empty content without control characters and within 64 Unicode characters.
* **Documentation**
  * Documented reaction controls, targeting behavior, removal options, and content limits.

### DIFF
--- a/crates/agent-connector/src/messaging.rs
+++ b/crates/agent-connector/src/messaging.rs
@@ -284,6 +284,47 @@ impl AgentConnector {
         })
     }
 
+    pub(crate) async fn send_reaction_response(
+        &self,
+        account_id_hex: &str,
+        group_id_hex: &str,
+        target_message_id_hex: &str,
+        emoji: &str,
+    ) -> Result<AgentControlResponse, ConnectorError> {
+        let account = self.local_account_for_account_id(account_id_hex)?;
+        let group_id_hex = normalize_hex(group_id_hex)?;
+        let group_id = GroupId::new(hex::decode(&group_id_hex)?);
+        let target_message_id = normalize_hex(target_message_id_hex)?;
+        let summary = self
+            .runtime
+            .react_to_message(&account.label, &group_id, &target_message_id, emoji)
+            .await?;
+        Ok(AgentControlResponse::AppEventSent {
+            message_ids_hex: summary.message_ids,
+            maintenance_disposition: agent_maintenance_disposition(summary.maintenance_disposition),
+        })
+    }
+
+    pub(crate) async fn remove_reaction_response(
+        &self,
+        account_id_hex: &str,
+        group_id_hex: &str,
+        target_message_id_hex: &str,
+    ) -> Result<AgentControlResponse, ConnectorError> {
+        let account = self.local_account_for_account_id(account_id_hex)?;
+        let group_id_hex = normalize_hex(group_id_hex)?;
+        let group_id = GroupId::new(hex::decode(&group_id_hex)?);
+        let target_message_id = normalize_hex(target_message_id_hex)?;
+        let summary = self
+            .runtime
+            .unreact_from_message(&account.label, &group_id, &target_message_id)
+            .await?;
+        Ok(AgentControlResponse::AppEventSent {
+            message_ids_hex: summary.message_ids,
+            maintenance_disposition: agent_maintenance_disposition(summary.maintenance_disposition),
+        })
+    }
+
     /// Report group membership for an account's group so a channel can decide
     /// activation policy: `is_direct` (exactly two members, i.e. an effective DM
     /// where the agent always replies) vs a multi-party group that gates on
--- a/crates/agent-connector/src/messaging.rs
+++ b/crates/agent-connector/src/messaging.rs
@@ -310,6 +310,7 @@ impl AgentConnector {
         account_id_hex: &str,
         group_id_hex: &str,
         target_message_id_hex: &str,
+        emoji: Option<&str>,
     ) -> Result<AgentControlResponse, ConnectorError> {
         let account = self.local_account_for_account_id(account_id_hex)?;
         let group_id_hex = normalize_hex(group_id_hex)?;
@@ -317,7 +318,7 @@ impl AgentConnector {
         let target_message_id = normalize_hex(target_message_id_hex)?;
         let summary = self
             .runtime
-            .unreact_from_message(&account.label, &group_id, &target_message_id)
+            .unreact_from_message_matching(&account.label, &group_id, &target_message_id, emoji)
             .await?;
         Ok(AgentControlResponse::AppEventSent {
             message_ids_hex: summary.message_ids,
--- a/crates/agent-connector/src/tests.rs
+++ b/crates/agent-connector/src/tests.rs
@@ -1193,13 +1193,50 @@ async fn connector_socket_subscribes_to_inbound_messages() {
     assert_eq!(reaction_message_ids.len(), 1);
     assert!(!reaction_message_ids[0].is_empty());
 
+    let repeated_add = send_control_request(
+        &socket,
+        "req-agent-reaction-add-retry",
+        AgentControlRequest::SendReaction {
+            account_id_hex: agent.account.account_id_hex.clone(),
+            group_id_hex: group_id_hex.clone(),
+            target_message_id_hex: human_message_id_hex.clone(),
+            emoji: "👀".to_owned(),
+        },
+    )
+    .await;
+    let AgentControlResponse::AppEventSent {
+        message_ids_hex: repeated_message_ids,
+        ..
+    } = repeated_add.payload
+    else {
+        panic!("expected idempotent reaction add response");
+    };
+    assert_eq!(repeated_message_ids, reaction_message_ids);
+
+    let completion_added = send_control_request(
+        &socket,
+        "req-agent-reaction-add-completion",
+        AgentControlRequest::SendReaction {
+            account_id_hex: agent.account.account_id_hex.clone(),
+            group_id_hex: group_id_hex.clone(),
+            target_message_id_hex: human_message_id_hex.clone(),
+            emoji: "✅".to_owned(),
+        },
+    )
+    .await;
+    assert!(matches!(
+        completion_added.payload,
+        AgentControlResponse::AppEventSent { .. }
+    ));
+
     let reaction_removed = send_control_request(
         &socket,
-        "req-agent-reaction-remove",
+        "req-agent-reaction-remove-matching",
         AgentControlRequest::RemoveReaction {
             account_id_hex: agent.account.account_id_hex.clone(),
             group_id_hex: group_id_hex.clone(),
             target_message_id_hex: human_message_id_hex.clone(),
+            emoji: Some("👀".to_owned()),
         },
     )
     .await;
@@ -1215,16 +1252,49 @@ async fn connector_socket_subscribes_to_inbound_messages() {
 
     let repeated_removal = send_control_request(
         &socket,
-        "req-agent-reaction-remove-repeat",
+        "req-agent-reaction-remove-matching-repeat",
         AgentControlRequest::RemoveReaction {
             account_id_hex: agent.account.account_id_hex.clone(),
             group_id_hex: group_id_hex.clone(),
             target_message_id_hex: human_message_id_hex.clone(),
+            emoji: Some("👀".to_owned()),
         },
     )
     .await;
     let AgentControlResponse::Error { code, .. } = repeated_removal.payload else {
         panic!("expected missing-reaction error on repeated removal");
+    };
+    assert_eq!(code, "reaction_not_found");
+
+    let remaining_removed = send_control_request(
+        &socket,
+        "req-agent-reaction-remove-all",
+        AgentControlRequest::RemoveReaction {
+            account_id_hex: agent.account.account_id_hex.clone(),
+            group_id_hex: group_id_hex.clone(),
+            target_message_id_hex: human_message_id_hex.clone(),
+            emoji: None,
+        },
+    )
+    .await;
+    assert!(matches!(
+        remaining_removed.payload,
+        AgentControlResponse::AppEventSent { .. }
+    ));
+
+    let repeated_remove_all = send_control_request(
+        &socket,
+        "req-agent-reaction-remove-all-repeat",
+        AgentControlRequest::RemoveReaction {
+            account_id_hex: agent.account.account_id_hex.clone(),
+            group_id_hex: group_id_hex.clone(),
+            target_message_id_hex: human_message_id_hex.clone(),
+            emoji: None,
+        },
+    )
+    .await;
+    let AgentControlResponse::Error { code, .. } = repeated_remove_all.payload else {
+        panic!("expected missing-reaction error after removing all reactions");
     };
     assert_eq!(code, "reaction_not_found");
 
--- a/crates/agent-connector/src/tests.rs
+++ b/crates/agent-connector/src/tests.rs
@@ -1165,6 +1165,47 @@ async fn connector_socket_subscribes_to_inbound_messages() {
     assert!(!reply_to.text_truncated);
     assert!(reply_to.recorded_at.is_some());
 
+    let reaction_added = send_control_request(
+        &socket,
+        "req-agent-reaction-add",
+        AgentControlRequest::SendReaction {
+            account_id_hex: agent.account.account_id_hex.clone(),
+            group_id_hex: group_id_hex.clone(),
+            target_message_id_hex: human_message_id_hex.clone(),
+            emoji: "👀".to_owned(),
+        },
+    )
+    .await;
+    let AgentControlResponse::AppEventSent {
+        message_ids_hex: reaction_message_ids,
+        ..
+    } = reaction_added.payload
+    else {
+        panic!("expected reaction add response");
+    };
+    assert_eq!(reaction_message_ids.len(), 1);
+    assert!(!reaction_message_ids[0].is_empty());
+
+    let reaction_removed = send_control_request(
+        &socket,
+        "req-agent-reaction-remove",
+        AgentControlRequest::RemoveReaction {
+            account_id_hex: agent.account.account_id_hex.clone(),
+            group_id_hex: group_id_hex.clone(),
+            target_message_id_hex: human_message_id_hex.clone(),
+        },
+    )
+    .await;
+    let AgentControlResponse::AppEventSent {
+        message_ids_hex: removal_message_ids,
+        ..
+    } = reaction_removed.payload
+    else {
+        panic!("expected reaction remove response");
+    };
+    assert_eq!(removal_message_ids.len(), 1);
+    assert!(!removal_message_ids[0].is_empty());
+
     let deleted = send_control_request(
         &socket,
         "req-human-delete",
--- a/crates/agent-connector/src/tests.rs
+++ b/crates/agent-connector/src/tests.rs
@@ -86,6 +86,13 @@ fn send_in_progress_has_a_distinct_retry_contract() {
     );
 }
 
+#[test]
+fn reaction_not_found_has_a_distinct_idempotency_contract() {
+    let error = crate::ConnectorError::App(marmot_app::AppError::ReactionNotFound);
+    assert_eq!(error.code(), "reaction_not_found");
+    assert_eq!(error.client_message(), "no matching reaction to remove");
+}
+
 #[tokio::test]
 async fn control_frame_write_timeout_includes_flush() {
     struct FlushStallingWriter;
--- a/crates/agent-connector/src/tests.rs
+++ b/crates/agent-connector/src/tests.rs
@@ -1213,6 +1213,21 @@ async fn connector_socket_subscribes_to_inbound_messages() {
     assert_eq!(removal_message_ids.len(), 1);
     assert!(!removal_message_ids[0].is_empty());
 
+    let repeated_removal = send_control_request(
+        &socket,
+        "req-agent-reaction-remove-repeat",
+        AgentControlRequest::RemoveReaction {
+            account_id_hex: agent.account.account_id_hex.clone(),
+            group_id_hex: group_id_hex.clone(),
+            target_message_id_hex: human_message_id_hex.clone(),
+        },
+    )
+    .await;
+    let AgentControlResponse::Error { code, .. } = repeated_removal.payload else {
+        panic!("expected missing-reaction error on repeated removal");
+    };
+    assert_eq!(code, "reaction_not_found");
+
     let deleted = send_control_request(
         &socket,
         "req-human-delete",
--- a/crates/agent-control/README.md
+++ b/crates/agent-control/README.md
@@ -27,6 +27,10 @@ begin inputs is an error, and trying to begin another active stream with an occu
 - Provides newline-delimited JSON framing with a 1 MiB per-frame cap.
 - Stays dependency-light: `serde` and Tokio IO only.
 
+`send_reaction` adds arbitrary non-blank reaction content to a durable message.
+`remove_reaction` takes the original target message id and retracts the calling
+account's matching reaction; callers do not need to discover the reaction event id.
+
 ## Materialized timeline reads
 
 `timeline_message_get` resolves one durable message id and `timeline_list` pages a
--- a/crates/agent-control/README.md
+++ b/crates/agent-control/README.md
@@ -30,7 +30,9 @@ begin inputs is an error, and trying to begin another active stream with an occu
 `send_reaction` adds arbitrary non-blank, control-free reaction content of at
 most 64 Unicode scalar values to a durable message.
 `remove_reaction` takes the original target message id and retracts the calling
-account's matching reaction; callers do not need to discover the reaction event id.
+account's most recent active reaction on that message, whichever reaction
+content it carries; callers do not need to discover the reaction event id. To
+swap a progress reaction for a completion reaction, remove first and then add.
 
 ## Materialized timeline reads
 
--- a/crates/agent-control/README.md
+++ b/crates/agent-control/README.md
@@ -28,11 +28,13 @@ begin inputs is an error, and trying to begin another active stream with an occu
 - Stays dependency-light: `serde` and Tokio IO only.
 
 `send_reaction` adds arbitrary non-blank, control-free reaction content of at
-most 64 Unicode scalar values to a durable message.
-`remove_reaction` takes the original target message id and retracts the calling
-account's most recent active reaction on that message, whichever reaction
-content it carries; callers do not need to discover the reaction event id. To
-swap a progress reaction for a completion reaction, remove first and then add.
+most 64 Unicode scalar values to a durable message. Repeating the same content
+from the same account on the same target is idempotent and returns the existing
+reaction id instead of publishing a duplicate.
+`remove_reaction` takes the original target message id; callers do not need to
+discover reaction event ids. An optional `emoji` retracts all of the calling
+account's active reactions with that exact content. Omitting `emoji` retracts
+all of the account's active reactions on the target in one durable delete event.
 
 ## Materialized timeline reads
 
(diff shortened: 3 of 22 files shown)


Datawav:
Addressed independent review findings: exposed Hermes-compatible public reaction methods (including latest-inbound fallback) and preserved the privacy-safe `reaction_not_found` control code so remove remains idempotent. Verified `just fast-ci`, the full Hermes adapter suite, the full agent-control and agent-connector suites, plus focused follow-up tests. Kimi K3 gate was attempted through the required runner but OpenRouter returned HTTP 402; no Kimi verdict was claimed.

dannym-arx:
Just a minor nitpick:

reaction content has no length cap or control-character validation; content is bounded solely by the 1 MiB control-frame cap, reaction content can contain control characters/newlines and flows into notification preview rendering.

integrations/hermes/marmot/adapter.py
```diff
@@ -2712,6 +2749,55 @@ async def delete_message(self, chat_id: str, message_id: str) -> bool:
...
+        if not target_message_id:
+            return {"success": False, "error": "No Marmot message is available to react to"}
+        success = await self._add_reaction(chat_id, target_message_id, emoji)
+        return {"success": success, "emoji": emoji}
+
+    async def _remove_reaction(self, chat_id: str, message_id: str) -> bool:
+        try:
+            account_id = await self._ensure_account_id()
+            await self.client.remove_reaction(account_id, chat_id, message_id)
+            return True
```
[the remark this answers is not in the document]
Datawav: Addressed in 92dfbe08: add/remove now require an `app_event_sent` response with non-empty `message_ids_hex`, with regression coverage for wrong response types and empty IDs.

integrations/hermes/marmot/adapter.py
```diff
@@ -2712,6 +2749,55 @@ async def delete_message(self, chat_id: str, message_id: str) -> bool:
...
+        except AgentControlError as exc:
+            # A missing own reaction is already the desired state. This occurs
+            # after cancellation or when an earlier acknowledgement failed.
+            if exc.code == "reaction_not_found":
+                return True
+            logger.debug("Marmot remove_reaction failed: %s", exc)
+            return False
+        except Exception as exc:
+            logger.debug("Marmot remove_reaction failed: %s", exc)
+            return False
```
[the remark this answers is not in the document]
Datawav: Addressed in 92dfbe08: reaction failures now log only stable `AgentControlError.code` or exception class names; raw exception text is no longer interpolated.

Datawav:
Second-round review completed. In addition to resolving the two prior inline findings, the review identified and fixed an agent-controlled reaction-content boundary: reactions are now limited to 64 Unicode scalar values and reject control characters before they can reach timelines or notification previews. Regression tests cover blank, control-character, oversized, and multi-scalar emoji cases. Local agent-control, agent-connector, Marmot app reaction, and Hermes adapter suites pass. Kimi K3 was retried through the required runner but remains unavailable due OpenRouter HTTP 402; no Kimi approval is claimed.

Datawav:
Addressed the latest CodeRabbit review in 63816b49: latest inbound reaction targeting is now recorded before mention/onboarding activation filters, with a regression test for an unmentioned inbound message followed by fallback `add_reaction`. Also added exact-64 Unicode-scalar acceptance coverage for the reaction-content boundary. Verified the full Hermes adapter suite (196 tests), focused `marmot-app` reaction tests (5 tests), and `cargo fmt --all -- --check`.

crates/agent-connector/src/messaging.rs
```diff
@@ -284,6 +284,47 @@ impl AgentConnector {
...
+        group_id_hex: &str,
+        target_message_id_hex: &str,
+    ) -> Result<AgentControlResponse, ConnectorError> {
+        let account = self.local_account_for_account_id(account_id_hex)?;
+        let group_id_hex = normalize_hex(group_id_hex)?;
+        let group_id = GroupId::new(hex::decode(&group_id_hex)?);
+        let target_message_id = normalize_hex(target_message_id_hex)?;
+        let summary = self
+            .runtime
+            .unreact_from_message(&account.label, &group_id, &target_message_id)
```
erskingardner: [P1] Exclude already-deleted reactions before unreacting

This call will not produce `ReactionNotFound` when a successful removal is retried after its response is lost. `AppClient::own_reaction_event_id` scans raw `app_messages` and picks the newest matching kind-7 row, but a kind-5 removal leaves that raw reaction row in place; only the materialized timeline filters it as deleted. The retry therefore publishes another durable kind-5 for the same reaction and returns `AppEventSent`. It can also keep selecting a tombstoned newest reaction instead of an older still-active one. Resolve the own reaction from materialized active reactions (or exclude same-author valid deletes) and add add/remove/remove coverage proving that the second removal reaches the existing `ReactionNotFound`/no-send path.

erskingardner (changes requested):
Request changes: the new removal contract does not currently make retries idempotent at the durable-event boundary. The focused Rust and Hermes paths otherwise line up with the requested wire/API shape, and the current required checks are passing aside from the still-running simulator job.

crates/agent-control/src/lib.rs
```diff
@@ -115,6 +115,19 @@ pub enum AgentControlRequest {
...
         target_message_id_hex: String,
     },
+    /// Add bounded, control-free, non-blank reaction content to a durable message.
+    SendReaction {
+        account_id_hex: String,
+        group_id_hex: String,
+        target_message_id_hex: String,
+        emoji: String,
+    },
+    /// Retract this account's reaction to a durable message.
```
erskingardner: `RemoveReaction` carries no emoji, and the app-layer resolution it maps to (`own_reaction_event_id`, `crates/marmot-app/src/client/mod.rs:2180-2203`) reads the group's app events in ascending order, `.rev()`s them, and takes the first own kind-7 whose `e` tag matches the target — i.e. **your most recent reaction on that message, whatever emoji it carried**. Nothing prevents an account from holding several reactions on one target.

That silently misfires on the exact flow #862 motivates (Telegram-style 👀 → ✅):

- `send_reaction(target, 👀)` on receipt
- `send_reaction(target, ✅)` on completion
- `remove_reaction(target)` intending to clear 👀 → clears ✅, leaving 👀 on the message permanently.

Either would fix it:

- add `emoji: Option<String>` to `RemoveReaction` and resolve the own reaction by content when present (an optional field is additive, so `v2` stays compatible), or
- keep it emoji-agnostic and say so explicitly in the docs here and in the README, so callers know the only safe pattern is one live reaction per target.

crates/agent-control/README.md
```diff
@@ -27,6 +27,11 @@ begin inputs is an error, and trying to begin another active stream with an occu
 - Provides newline-delimited JSON framing with a 1 MiB per-frame cap.
 - Stays dependency-light: `serde` and Tokio IO only.
 
+`send_reaction` adds arbitrary non-blank, control-free reaction content of at
+most 64 Unicode scalar values to a durable message.
+`remove_reaction` takes the original target message id and retracts the calling
```
erskingardner: "retracts the calling account's **matching** reaction" reads as "the reaction matching the emoji I sent", but `RemoveReaction` has no emoji field and the resolution is emoji-agnostic (most recent own reaction on that target). Please state the actual rule — e.g. "retracts this account's most recent reaction on that message, whichever emoji it carried" — so agent authors don't build the 👀 → ✅ swap on top of it.

integrations/hermes/marmot/adapter.py
```diff
@@ -2712,6 +2749,62 @@ async def delete_message(self, chat_id: str, message_id: str) -> bool:
...
+                response.get("message_ids_hex")
+            )
+        except AgentControlError as exc:
+            logger.debug("Marmot send_reaction failed (code=%s)", exc.code or "unknown")
+            return False
+        except Exception as exc:
+            logger.debug("Marmot send_reaction failed (%s)", type(exc).__name__)
+            return False
+
+    async def add_reaction(
```
erskingardner, line 2769: This is the surface the whole feature hangs on, and nothing in the diff pins it to Hermes' real contract. The base-adapter double at `integrations/hermes/marmot/tests/test_adapter.py:106` defines no `add_reaction`, so a wrong method name, a different parameter order, or a bool-vs-dict return against `BasePlatformAdapter` in the pinned Hermes 0.19.0 (`hermes-agent.lock`) passes every test in this PR and still never gets called at runtime. #862 explicitly asks to match the existing base hook, and mdk#836 is precedent for exactly this class of silent break (hence the `connect(is_reconnect=...)` note at test_adapter.py:126).

The shape also diverges from the sibling hook 70 lines up: `delete_message(chat_id, message_id) -> bool` (target positional, bool result) vs `add_reaction(chat_id, emoji, message_id=None) -> Dict`. If the dict/`success` shape is the message-tool action contract rather than the adapter-hook contract, a one-line comment saying so would remove the ambiguity.

Could you confirm against the pinned checkout before merge — a `just hermes-dev-e2e-connector` run, or just quoting the base signature in the PR body? The listed verification is only the stubbed unit suite.

integrations/hermes/marmot/adapter.py
```diff
@@ -2712,6 +2749,62 @@ async def delete_message(self, chat_id: str, message_id: str) -> bool:
...
+        chat_id: str,
+        emoji: str,
+        message_id: Optional[str] = None,
+    ) -> Dict[str, Any]:
+        """Add an explicit agent-requested reaction to a durable message."""
+        target_message_id = message_id or self._last_inbound_message_ids.get(chat_id)
+        if not target_message_id:
+            return {"success": False, "error": "No Marmot message is available to react to"}
+        success = await self._add_reaction(chat_id, target_message_id, emoji)
+        return {"success": success, "emoji": emoji}
```
erskingardner: Failure is reported in two shapes: the no-target path returns `{"success": False, "error": ...}`, a failed send returns `{"success": False, "emoji": ...}` with no `error` key. Worth converging on one shape so a caller can always read `error` on failure (`remove_reaction` below has the same split).

integrations/hermes/marmot/adapter.py
```diff
@@ -2712,6 +2749,62 @@ async def delete_message(self, chat_id: str, message_id: str) -> bool:
             logger.debug("Marmot delete_message failed: %s", exc)
             return False
 
+    async def _add_reaction(self, chat_id: str, message_id: str, emoji: str) -> bool:
```
erskingardner, line 2755: `_add_reaction`/`add_reaction` — and `_remove_reaction`/`remove_reaction` below — are four methods where two would do: each private helper has exactly one non-test caller and buys only a bool→dict hop. Inlining them into the public methods drops ~15 lines and one layer of indirection. Non-blocking, but single-use helpers are worth resisting.

integrations/hermes/marmot/adapter.py
```diff
@@ -1847,6 +1883,7 @@ def __init__(self, config: PlatformConfig, client: Optional[MarmotAgentControlCl
         )
         self._profile_lookup_gate = ProfileLookupGate()
         self._sent_targets = SentMessageTargetCache()
+        self._last_inbound_message_ids: Dict[str, str] = {}
```
erskingardner, line 1888: Unlike its neighbours, this per-group map is neither bounded nor cleared: `_pending_ambient_context` caps group count at `MAX_PENDING_AMBIENT_GROUPS` (adapter.py:3458) and is cleared in `disconnect()` (adapter.py:1932), and `_sent_targets` is a bounded `SentMessageTargetCache`. Both consequences are small but real — unbounded growth in group count, and after a reconnect a bare `add_reaction(chat_id, emoji)` can still target a message id from the previous session. Reusing the existing bounded pattern (or at least clearing it in `disconnect()`) keeps this consistent with the rest of the adapter's state.

crates/agent-connector/src/tests.rs
```diff
@@ -1165,6 +1172,47 @@ async fn connector_socket_subscribes_to_inbound_messages() {
...
+        message_ids_hex: reaction_message_ids,
+        ..
+    } = reaction_added.payload
+    else {
+        panic!("expected reaction add response");
+    };
+    assert_eq!(reaction_message_ids.len(), 1);
+    assert!(!reaction_message_ids[0].is_empty());
+
+    let reaction_removed = send_control_request(
```
erskingardner, line 1232: The happy path is covered, but the contract the Hermes idempotent-remove depends on — a remove with no existing reaction surfacing as `reaction_not_found` **over the wire** — is not. `reaction_not_found_has_a_distinct_idempotency_contract` constructs `ConnectorError::App(AppError::ReactionNotFound)` directly, so it can't catch a regression where the app error stops propagating unwrapped out of `unreact_from_message`. A second `RemoveReaction` right here, asserting the error code on the response, would close the gap for a few lines.

erskingardner (changes requested):
Well-scoped exposure of an already-mature app-layer primitive. The connector handlers mirror `delete_message_response` line for line, the wire-shape and request-type tests are the right ones, `AppError::ReactionNotFound` genuinely propagates unwrapped from `unreact_from_message` so the new error code is sound, and putting the content bound in `intents.rs` is the right layer (it covers CLI/UniFFI/app senders too, not just agents). Net new surface is small and the premise is good: #862 is a real gap and this is the minimal way to close it.

Two things I'd like resolved before merge, both cheap:

1. **`RemoveReaction` is emoji-agnostic, but the API and README read as if it isn't.** It retracts the account's *most recent* reaction on the target whatever the emoji, which silently misfires on the 👀 → ✅ status flow that motivates the issue. Either take an optional emoji or state the real rule.
2. **The public Hermes `add_reaction`/`remove_reaction` contract is unverified.** The in-repo base-adapter double defines no such hook, so a name/signature mismatch against the pinned Hermes 0.19.0 passes every test in this PR and still never fires at runtime — the same failure mode as #836. The verification list only covers the stubbed unit tests; please confirm against the pinned checkout.

Non-blocking: unbounded/uncleared `_last_inbound_message_ids`, inconsistent failure-result shape, a missing end-to-end assertion for `reaction_not_found`, and four adapter methods where two would do. Details inline.

crates/agent-connector/src/messaging.rs
Datawav: Addressed in 3869036c. `own_reaction_event_id` now resolves from the materialized timeline’s active `user_reactions`, so tombstoned reactions are excluded. The connector integration test now performs add/remove/remove and verifies the retry returns `reaction_not_found` without another successful durable send.

crates/agent-control/src/lib.rs
Datawav: Addressed in 3869036c by documenting and preserving the additive v2 shape: `RemoveReaction` retracts the account’s most recent active reaction on the target, whichever content it carries. The docs now prescribe remove-then-add for progress-to-completion swaps.

crates/agent-control/README.md
Datawav: Addressed in 3869036c. The README now states the exact most-recent-active semantics and explicitly documents remove-then-add for reaction swaps.

integrations/hermes/marmot/adapter.py
Datawav: Confirmed against the pinned Hermes 0.19.0 checkout at 3ef6bbd201263d354fd83ec55b3c306ded2eb72a: Photon defines `add_reaction(chat_id, emoji, message_id=None) -> Dict` and `remove_reaction(chat_id, message_id=None) -> Dict`; `tools/send_message_tool.py` dispatches those exact signatures. 3869036c aligns Marmot’s success/error result shape with Photon and adds the hooks to the base test double. Full Hermes adapter suite passes (198 tests), and `just fast-ci` passes.

integrations/hermes/marmot/adapter.py
Datawav: Addressed in 3869036c. Both public methods now include `error` on every failure and return the resolved `message_id` on success, matching the pinned Photon contract.

integrations/hermes/marmot/adapter.py
Datawav: Verified; leaving the private bool helpers in place intentionally. They isolate control-response validation and idempotent `reaction_not_found` handling from Hermes’ public dict contract, and their focused tests directly cover those transport-boundary behaviors. No functional change needed for this non-blocking suggestion.

integrations/hermes/marmot/adapter.py
Datawav: Addressed the session-staleness concern in 3869036c by clearing `_last_inbound_message_ids` during `disconnect()`, with regression coverage. This map is one entry per known group; no new per-message growth occurs.

crates/agent-connector/src/tests.rs
Datawav: Addressed in 3869036c. The socket integration path now sends a second `RemoveReaction` after successful add/remove and asserts the wire response carries `code == reaction_not_found`. This also guards the app-to-connector error propagation.

crates/marmot-app/src/client/mod.rs
```diff
@@ -2255,23 +2253,26 @@ impl AppClient {
         target_message_id: &str,
     ) -> Result<String, AppError> {
         let group_id_hex = hex::encode(group_id.as_slice());
-        let messages = self.app.messages_with_query(
+        let timeline = self.app.timeline_messages_with_query(
```
erskingardner: [P1] Resolve the explicit target directly instead of paging the timeline

`TimelineMessageQuery::default()` is capped at the newest 50 rows, so `remove_reaction` now returns `reaction_not_found` for any still-retained target that has fallen outside that window—even when the caller supplies its exact durable message ID. This regresses the arbitrary-target control API and can leave old progress reactions stuck in long-lived groups. Use the existing `self.app.timeline_message(label, group_id_hex, target_message_id)` exact-row lookup, then select the newest active own `user_reaction`; add coverage with more than 50 newer timeline rows so this cannot collapse back into a page scan.

erskingardner (changes requested):
Request changes: the original tombstoned-reaction retry defect is fixed and the follow-up contract/test updates are sound, but the replacement lookup now silently limits removal to targets in the newest 50 timeline rows. All completed required checks pass; Simulator and vectors is still running.

crates/marmot-app/src/client/mod.rs
erskingardner: meh - this seems like a non thing to me.

erskingardner (dismissed):

crates/marmot-app/src/client/mod.rs
Datawav: Addressed in c07fc0d0. Unreact now resolves the supplied durable target with the exact `timeline_message` lookup and then selects the newest active own reaction, so targets older than the default 50-row timeline page remain removable. `cargo check -p marmot-app`, formatting, and diff integrity pass.

Datawav:
Addressed the latest CodeRabbit review nit in 28ca5afe. The public reaction API test now exercises `remove_reaction(chat_id)` against the cached latest inbound durable message ID while preserving the explicit-target removal assertion. Focused test and the full Hermes adapter suite pass (198 tests).

Datawav (commented):
## Automated review

No actionable findings.

erskingardner (commented):
Verified at head 4582393a; CI green (24 passed, release-publish skipped). The earlier P1s are genuinely fixed at this head: unreact now resolves via the exact `timeline_message` lookup over active `user_reactions` (no 50-row window, no tombstoned retry), `reaction_not_found` is covered over the socket, and the Hermes signatures match the real hermes-core contract (`add_reaction(chat_id=, emoji=, message_id=)` via `tools/send_message_tool.py`; `_add_reaction(chat_id, message_id, emoji)` per `gateway/platforms/base.py`).

Remaining items inline: two unannounced behavior changes on existing public surfaces (unreact-now-removes-all, idempotent-add-for-all-callers), one unrelated drive-by, and two small nits. None are merge-blocking in my view once the behavior changes get an explicit call-out in the PR body, but the first two touch UniFFI/CLI-visible semantics so they deserve a deliberate decision.

crates/marmot-app/src/client/mod.rs
```diff
@@ -2064,22 +2062,33 @@ impl AppClient {
...
+                if let Some(emoji) = emoji.as_deref() {
+                    crate::messages::validate_reaction_content(emoji)?;
+                }
+                let reaction_message_ids = self.own_reaction_event_ids(
+                    group_id,
+                    &sender,
+                    &target_message_id,
+                    emoji.as_deref(),
+                )?;
+                AppMessageIntent::DeleteReactions {
```
erskingardner, line 2090: Heads-up on existing-surface semantics: this changes `unreact_from_message` (UniFFI → app UI, and the CLI) from "retract the most recent own reaction" to "retract ALL of the account's active reactions on the target in one kind-5". A user with 👀+✅ on a message now loses both on a single un-tap; previously only the newest was retracted. Remove-all is arguably the better contract (and exactly what a remove-then-add status swap wants), but it is a user-visible behavior change not mentioned in the PR body, and the UniFFI docstring still says "Remove this account's reaction" (singular). Worth an explicit call-out and a docstring touch-up.

crates/marmot-app/src/client/mod.rs
```diff
@@ -2289,13 +2301,45 @@ impl AppClient {
...
+    where
+        F: FnMut(crate::AppProjectionUpdate),
+    {
+        crate::messages::validate_reaction_content(emoji)?;
+        let sender = self
+            .app
+            .account_home()
+            .account(&self.state.label)?
+            .account_id_hex;
+        match self.own_reaction_event_ids(group_id, &sender, target_message_id, Some(emoji)) {
```
erskingardner, line 2325: This makes `react_to_message` idempotent for every caller, not just agents: app UI / CLI / UniFFI re-reacts now return `published: 0` plus the pre-existing reaction id instead of publishing. Probably the right behavior, but three side effects worth noting: (a) the CLI prints `reacted 👀 to <id> published=0` — claiming "reacted" for a no-op; (b) the early return skips `ensure_group_application_messages_allowed`, so a repeat react in a group where app messages are now disallowed still reports success; (c) the no-op path produces no audit record even though a send was requested. At minimum this belongs in the PR body; worth deciding whether (b)/(c) matter enough to change.

integrations/hermes/marmot/tests/e2e_connector.py
```diff
@@ -124,9 +124,13 @@ async def run() -> None:
 
     from gateway.config import PlatformConfig
 
-    # Keep Unix socket paths short for macOS sockaddr_un limits. Python's
-    # default temp dir can live under a long /var/folders/... path.
-    with tempfile.TemporaryDirectory(prefix="mhce-", dir="/tmp") as tmp:
+    # Keep Unix socket paths short for macOS sockaddr_un limits without using
+    # `/tmp`, whose canonical private-directory posture is intentionally
+    # rejected by fs-private. CI may supply another owner-controlled parent.
+    temp_parent = Path(
```
erskingardner, line 130: Unrelated drive-by: the `/tmp` → `$HOME`/`MARMOT_CONNECTOR_E2E_TMPDIR` temp-root change fixes the fs-private posture rejection, which is legitimate, but it's not a reactions change and doesn't appear in the PR body's summary. Worth splitting into its own commit/PR, or at least listing it in the Summary so the e2e fix stays discoverable in history.

crates/marmot-app/src/messages/intents.rs
```diff
@@ -535,6 +551,25 @@ pub(crate) fn build_inner_event(
     }
 }
 
+pub(crate) fn validate_reaction_content(emoji: &str) -> Result<(), AppError> {
+    if emoji.trim().is_empty() {
```
erskingardner, line 555: Nit: `trim()` is only used for the emptiness check — `" 👀"` with a leading space passes validation. Since remove-matching is exact-content, padded content can never round-trip through a caller that trims. Consider validating (or normalizing to) the trimmed form instead.

integrations/hermes/marmot/adapter.py
```diff
@@ -3846,6 +3962,57 @@ def _marmot_history_tool(args: Dict[str, Any], **_kwargs: Any) -> str:
...
+                }
+            )
+        if action == "add":
+            result = _run_async(adapter.add_reaction(group_id_hex, emoji, message_id_hex))
+        else:
+            result = _run_async(
+                adapter.remove_reaction(group_id_hex, message_id_hex, emoji or None)
+            )
+        return json.dumps({"ok": bool(result.get("success")), **result})
+    except AgentControlError as exc:
```
erskingardner: Nit: this `except AgentControlError` is unreachable — the adapter's public `add_reaction`/`remove_reaction` catch everything (including `AgentControlError`) and always return dicts, so the tool handler only ever sees the `{"success": False, ...}` shape. Harmless defensiveness, but dead code as written.

PR #1401, opened 2026-08-12
Expose message reactions to OpenClaw

Datawav:
## Summary

- expose OpenClaw `message(action="react")` support for Marmot conversations
- add and remove durable reactions through the shared `marmot.agent-control.v2` requests introduced by #1399
- allow an omitted `messageId` to target the current inbound message
- document the OpenClaw reaction contract and add client/action-adapter regression coverage
- give cold OpenClaw SDK import canaries enough time under a loaded CI host

## Dependency

This PR is intentionally separate from #1399 and is based on its shared agent-control/connector reaction commits. Merge #1399 first, then this OpenClaw adapter PR. Never enable auto-merge.

## Validation

- `just fast-ci`
- `cargo test -p agent-control` (18 passed)
- `cargo test -p agent-connector` (138 passed)
- `pnpm typecheck`
- `pnpm test` (230 passed, 2 skipped)
- `/tmp/openclaw-marmot-test/e2e-connector.sh` (1 passed against a real debug `wn-agent` process)
- `git diff --check`

No version or release metadata is changed.

## Summary by CodeRabbit

- **New Features**
  - Added support for adding and removing message reactions in Marmot.
  - Supports explicit reaction targets, emoji validation, and message ID aliases.
  - Reaction removal is safely repeatable when the reaction is already absent.
  - Existing message deletion remains supported.

- **Bug Fixes**
  - Added validation and clearer errors for invalid reaction requests and responses.

- **Documentation**
  - Documented Marmot reaction usage through the shared message tool.

### DIFF
--- a/integrations/openclaw/marmot/src/channel.ts
+++ b/integrations/openclaw/marmot/src/channel.ts
@@ -30,7 +30,7 @@ import {
 } from "./config.js";
 import { startMarmotGatewayAccount } from "./gateway.js";
 import { AgentControlError } from "./client.js";
-import { createMarmotMessagingAdapter } from "./messaging.js";
+import { createMarmotMessagingAdapter, normalizeMarmotTarget } from "./messaging.js";
 import { createMarmotMessageAdapter } from "./outbound.js";
 import {
   DEFAULT_MARMOT_CHANNEL_ACCOUNT_ID,
@@ -137,13 +137,13 @@ export interface MarmotMessageActionClient {
     groupIdHex: string,
     targetMessageIdHex: string,
   ) => Promise<unknown>;
-  sendReaction?: (
+  sendReaction: (
     accountIdHex: string,
     groupIdHex: string,
     targetMessageIdHex: string,
     emoji: string,
   ) => Promise<unknown>;
-  removeReaction?: (
+  removeReaction: (
     accountIdHex: string,
     groupIdHex: string,
     targetMessageIdHex: string,
@@ -170,11 +170,15 @@ export function createMarmotMessageActionAdapter(
             : typeof ctx.params.message_id === "string"
               ? ctx.params.message_id
             : String(ctx.toolContext?.currentMessageId ?? "");
+        // `deleteByMessageId` mutates immediately and only returns a boolean;
+        // it cannot safely recover a cached group for a reaction. Current
+        // inbound reactions inherit the tool context, while reactions to an
+        // earlier outbound message therefore require an explicit `to`.
         const rawTo =
           typeof ctx.params.to === "string"
             ? ctx.params.to
             : ctx.toolContext?.currentMessagingTarget;
-        const to = rawTo?.startsWith("marmot:") ? rawTo.slice("marmot:".length) : rawTo;
+        const to = normalizeMarmotTarget(rawTo ?? "");
         if (!messageId) {
           return jsonResult({
             ok: false,
@@ -184,26 +188,26 @@ export function createMarmotMessageActionAdapter(
         if (!to) {
           return jsonResult({ ok: false, error: "could not resolve group for this message id" });
         }
+        const remove = ctx.params.remove === true;
+        if (ctx.params.emoji !== undefined && typeof ctx.params.emoji !== "string") {
+          return jsonResult({ ok: false, error: "emoji must be a string" });
+        }
         const emoji = typeof ctx.params.emoji === "string" ? ctx.params.emoji : "";
-        const remove = ctx.params.remove === true || emoji.length === 0;
+        if (!remove && emoji.length === 0) {
+          return jsonResult({ ok: false, error: "emoji required unless remove is true" });
+        }
         try {
           const { client, marmotAccountIdHex } = await deps.resolveTarget(
             ctx.cfg,
             ctx.accountId ?? null,
           );
           if (remove) {
-            if (!client.removeReaction) {
-              return jsonResult({ ok: false, error: "reaction removal unavailable" });
-            }
             if (emoji.length > 0) {
               await client.removeReaction(marmotAccountIdHex, to, messageId, emoji);
             } else {
               await client.removeReaction(marmotAccountIdHex, to, messageId);
             }
             return jsonResult({ ok: true, removed: true });
-          }
-          if (!client.sendReaction) {
-            return jsonResult({ ok: false, error: "reactions unavailable" });
           }
           await client.sendReaction(marmotAccountIdHex, to, messageId, emoji);
           return jsonResult({ ok: true, added: emoji });
--- a/integrations/openclaw/marmot/src/channel.ts
+++ b/integrations/openclaw/marmot/src/channel.ts
@@ -115,8 +115,8 @@ async function probeMarmotAccount(account: ResolvedMarmotAccount): Promise<Marmo
   };
 }
 
-/** Dependencies the channel-owned `delete` action adapter needs. */
-export interface MarmotDeleteActionDeps {
+/** Dependencies the channel-owned message action adapter needs. */
+export interface MarmotMessageActionDeps {
   /** Resolve a sent message id to its group from the send-time cache. */
   deleteByMessageId: (
     targetMessageIdHex: string,
@@ -126,12 +126,23 @@ export interface MarmotDeleteActionDeps {
   resolveTarget: (
     cfg: unknown,
     accountId?: string | null,
-  ) => Promise<{ client: MarmotMessageDeleteClient; marmotAccountIdHex: string }>;
+  ) => Promise<{ client: MarmotMessageActionClient; marmotAccountIdHex: string }>;
 }
 
-/** Narrow view of the control client used by the explicit-group delete fallback. */
-export interface MarmotMessageDeleteClient {
+/** Narrow view of the control client used by message mutations. */
+export interface MarmotMessageActionClient {
   deleteMessage: (
+    accountIdHex: string,
+    groupIdHex: string,
+    targetMessageIdHex: string,
+  ) => Promise<unknown>;
+  sendReaction?: (
+    accountIdHex: string,
+    groupIdHex: string,
+    targetMessageIdHex: string,
+    emoji: string,
+  ) => Promise<unknown>;
+  removeReaction?: (
     accountIdHex: string,
     groupIdHex: string,
     targetMessageIdHex: string,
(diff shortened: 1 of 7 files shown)


Datawav:
@coderabbitai review

integrations/openclaw/marmot/src/channel.ts
```diff
@@ -126,30 +126,82 @@ export interface MarmotDeleteActionDeps {
...
 ): ChannelMessageActionAdapter {
   return {
-    describeMessageTool: () => ({ actions: ["delete"] }),
-    supportsAction: ({ action }) => action === "delete",
+    describeMessageTool: () => ({ actions: ["delete", "react"] }),
+    supportsAction: ({ action }) => action === "delete" || action === "react",
     handleAction: async (ctx: ChannelMessageActionContext) => {
+      if (ctx.action === "react") {
+        const messageId =
+          typeof ctx.params.messageId === "string"
```
erskingardner, line 168: [P1] Honor the host's `message_id` alias here. OpenClaw 2026.7.1-2 advertises both `messageId` and `message_id` in the shared reaction schema, and its normal parameter readers resolve snake_case aliases. This adapter reads only `ctx.params.messageId`, so `message(action="react", message_id="<older id>", emoji="…")` ignores the explicit target; during an inbound turn it silently falls back to `currentMessageId` and reacts to the wrong message (outside a turn it reports the id missing). Resolve both forms via the SDK reader or an equivalent local helper, and add a regression test where `message_id` differs from the current inbound id.

integrations/openclaw/marmot/src/channel.ts
```diff
@@ -126,30 +126,82 @@ export interface MarmotDeleteActionDeps {
...
+        if (!messageId) {
+          return jsonResult({
+            ok: false,
+            error: "messageId required. Provide messageId explicitly or react to the current inbound message.",
+          });
+        }
+        if (!to) {
+          return jsonResult({ ok: false, error: "could not resolve group for this message id" });
+        }
+        const { client, marmotAccountIdHex } = await deps.resolveTarget(
```
erskingardner: [P2] Put target/account resolution inside the privacy-safe failure boundary. `resolveTarget` can call `account_list` before an account is pinned; a failed Unix-socket connect is wrapped with the original `err.message`, which normally includes the configured socket path. Because this await is outside the `try`, that detail escapes as a raw tool error while only `sendReaction`/`removeReaction` failures get the generic response promised by the new test. Move this await into the sanitized `try` (and cover a rejecting `resolveTarget`) so every reaction-operation failure follows the same privacy contract.

erskingardner (changes requested):
The OpenClaw reaction plumbing is otherwise focused and the stacked CI is green, but the action adapter can target the wrong message for a schema-advertised input alias and its privacy-safe failure boundary does not cover target/account resolution. Please address the two inline findings before merging (after #1399, as planned).

integrations/openclaw/marmot/src/channel.ts
Datawav: Fixed in 1c8f5dc6. Reaction target resolution now accepts the OpenClaw `message_id` alias before falling back to `currentMessageId`; the regression test uses different explicit/current IDs and verifies the explicit target is sent. Validated with `npm test -- --run test/channel.test.ts` (17 passed) and `npm run typecheck`.

integrations/openclaw/marmot/src/channel.ts
Datawav: Fixed in 1c8f5dc6. `resolveTarget` now runs inside the privacy-safe reaction failure boundary, with a rejecting-resolution regression test that verifies socket details do not escape. Validated with `npm test -- --run test/channel.test.ts` (17 passed) and `npm run typecheck`.

integrations/openclaw/marmot/src/channel.ts
```diff
@@ -126,30 +126,84 @@ export interface MarmotDeleteActionDeps {
...
+        try {
+          const { client, marmotAccountIdHex } = await deps.resolveTarget(
+            ctx.cfg,
+            ctx.accountId ?? null,
+          );
+          if (remove) {
+            if (!client.removeReaction) {
+              return jsonResult({ ok: false, error: "reaction removal unavailable" });
+            }
+            await client.removeReaction(marmotAccountIdHex, to, messageId);
```
Datawav: The target is forwarded unchanged, so the documented `marmot:<group-id>` form reaches `normalizeHex` and is rejected. Normalize or strip the exact `marmot:` prefix before calling both reaction methods, and cover prefixed targets in add/remove tests.

integrations/openclaw/marmot/src/client.ts
```diff
@@ -65,6 +65,20 @@ export interface AppEventSentResponse {
   message_ids_hex: string[];
 }
 
+function requireAppEventSent(response: Envelope): AppEventSentResponse {
+  if (
+    response.type !== "app_event_sent" ||
+    !Array.isArray(response.message_ids_hex) ||
+    response.message_ids_hex.length === 0 ||
+    response.message_ids_hex.some((value) => typeof value !== "string" || value.length === 0)
```
Datawav: This accepts any nonempty string as a durable message ID, including non-hex, odd-length, or incorrectly sized values, so malformed broker responses are reported as successful reactions. Validate each entry as a protocol-sized hexadecimal identifier and add focused negative tests.

Datawav (commented):
## Automated review

Outcome: 2 actionable findings.

- P2 — Documented `marmot:` reaction targets fail. In changed `src/channel.ts:171–201`, `params.to` or `currentMessagingTarget` is forwarded unchanged as `groupIdHex`; changed `src/client.ts:638–658` then applies `normalizeHex`, which rejects the documented `marmot:<group-id>` form. Consequently, valid explicit reaction requests using the advertised target syntax return `reaction operation failed`. Strip the exact `marmot:` prefix through the shared target normalizer before invoking `sendReaction` or `removeReaction`, and add prefixed-target tests for both paths.

- Low — Malformed durable reaction IDs are accepted as successful responses. Changed `src/client.ts:68–79` validates each `message_ids_hex` entry only as a nonempty string, so non-hex, odd-length, or incorrectly sized IDs pass `requireAppEventSent`; changed tests cover only a wrong envelope type and an empty array. Consequently, a malformed or protocol-incompatible broker response is exposed as a successful durable reaction. Validate every returned ID as a protocol-sized hexadecimal identifier and add focused negative cases.

integrations/openclaw/marmot/src/channel.ts
Datawav: Fixed in 905545c4. The reaction adapter now strips the exact `marmot:` prefix from explicit and current-context targets before both add and remove calls. Regression coverage exercises prefixed targets for both operations. Validated with `npm test -- --run test/channel.test.ts test/client.test.ts` (41 passed) and `npm run typecheck`.

integrations/openclaw/marmot/src/client.ts
Datawav: Fixed in 905545c4. Durable reaction response IDs must now be exactly 32-byte (64-character) hexadecimal values; focused tests reject non-hex, odd-length, and wrong-size IDs in addition to wrong response types and empty arrays. Validated with `npm test -- --run test/channel.test.ts test/client.test.ts` (41 passed) and `npm run typecheck`.

integrations/openclaw/marmot/src/channel.ts
Datawav: Resolved in `fc32d66e`: reaction target selection now checks `messageId`, then the OpenClaw `message_id` alias, before falling back to `currentMessageId`. The regression at `test/channel.test.ts` uses a snake_case id that differs from the current inbound id and verifies the explicit older target is sent. Re-verified at head `905545c4`: TypeScript typecheck passes and the OpenClaw suite passes (23 files, 232 tests; 1 file/2 tests skipped).

integrations/openclaw/marmot/src/channel.ts
Datawav: Resolved in `fc32d66e`: `resolveTarget(...)` is inside the same privacy-safe `try` as the reaction RPC, so resolver/socket failures return only `reaction operation failed`. The regression at `test/channel.test.ts` makes `resolveTarget` reject with a secret socket path and asserts it is not exposed. Re-verified at head `905545c4`: TypeScript typecheck passes and the OpenClaw suite passes (23 files, 232 tests; 1 file/2 tests skipped).

integrations/openclaw/marmot/src/channel.ts
```diff
@@ -126,30 +127,97 @@ export interface MarmotDeleteActionDeps {
...
+          return jsonResult({
+            ok: false,
+            error: "messageId required. Provide messageId explicitly or react to the current inbound message.",
+          });
+        }
+        if (!to) {
+          return jsonResult({ ok: false, error: "could not resolve group for this message id" });
+        }
+        const emoji = typeof ctx.params.emoji === "string" ? ctx.params.emoji : "";
+        const remove = ctx.params.remove === true || emoji.length === 0;
```
erskingardner: Design concern: a `react` call with a missing/empty `emoji` — including a non-string `emoji`, which coerces to `""` on line 187 — silently becomes "remove ALL of the agent account's reactions on the target". A model that drops or mistypes the emoji param wipes reactions instead of getting an error. It's documented in the README, but a missing-param mistake shouldn't default to destruction; consider requiring explicit `remove: true` for removal and erroring on an empty emoji when `remove` isn't set.

integrations/openclaw/marmot/src/channel.ts
```diff
@@ -126,30 +127,97 @@ export interface MarmotDeleteActionDeps {
...
+          typeof ctx.params.messageId === "string"
+            ? ctx.params.messageId
+            : typeof ctx.params.message_id === "string"
+              ? ctx.params.message_id
+            : String(ctx.toolContext?.currentMessageId ?? "");
+        const rawTo =
+          typeof ctx.params.to === "string"
+            ? ctx.params.to
+            : ctx.toolContext?.currentMessagingTarget;
+        const to = rawTo?.startsWith("marmot:") ? rawTo.slice("marmot:".length) : rawTo;
```
erskingardner: This hand-rolled prefix strip duplicates `normalizeMarmotTarget` from `messaging.ts` (which also trims/lowercases, strips `0x`, and validates the group-id hex band). Today a garbage `to` falls through to the client's `normalizeHex`, throws, and surfaces as the generic "reaction operation failed" instead of the cleaner "could not resolve group" above. Using the shared helper fixes both the drift risk and the error message.

integrations/openclaw/marmot/src/channel.ts
```diff
@@ -126,30 +127,97 @@ export interface MarmotDeleteActionDeps {
...
-/** Narrow view of the control client used by the explicit-group delete fallback. */
-export interface MarmotMessageDeleteClient {
+/** Narrow view of the control client used by message mutations. */
+export interface MarmotMessageActionClient {
   deleteMessage: (
     accountIdHex: string,
     groupIdHex: string,
     targetMessageIdHex: string,
   ) => Promise<unknown>;
+  sendReaction?: (
```
erskingardner: Nit: `sendReaction`/`removeReaction` are optional here and guarded with "reactions unavailable" below, but the production `resolveTarget` always returns the full `MarmotAgentControlClient` — the optionality only exists to keep narrow test doubles compiling. Making them required (and adding the two methods to the doubles) deletes two dead guards.

integrations/openclaw/marmot/src/channel.ts
```diff
@@ -126,30 +127,97 @@ export interface MarmotDeleteActionDeps {
...
+export function createMarmotMessageActionAdapter(
+  deps: MarmotMessageActionDeps,
 ): ChannelMessageActionAdapter {
   return {
-    describeMessageTool: () => ({ actions: ["delete"] }),
-    supportsAction: ({ action }) => action === "delete",
+    describeMessageTool: () => ({ actions: ["delete", "react"] }),
+    supportsAction: ({ action }) => action === "delete" || action === "react",
     handleAction: async (ctx: ChannelMessageActionContext) => {
+      if (ctx.action === "react") {
```
erskingardner, line 166: Nit: unlike `delete`, `react` never consults the send-time cache (`deleteByMessageId`). The dominant case (reacting to an inbound message) is covered by the current-context fallback, but reacting to the agent's own earlier durable message with an explicit `messageId` requires an explicit `to`. Worth a code comment at minimum, or cache reuse for symmetry.

erskingardner (commented):
Verified at head 6643e09c; CI green (24 passed, release-publish skipped). All earlier review findings are addressed at this head: the `message_id` alias is honored, `resolveTarget` sits inside the privacy-safe `try`, the `marmot:` prefix is stripped, and response ids are validated as 64-char hex. I also cross-checked the SDK contract on openclaw@2026.7.1-2: `react` is a canonical message action and `toolContext.currentMessageId`/`currentMessagingTarget` are the documented fallback hooks, so the overall shape is right.

Four items inline: one design concern (empty `emoji` defaulting to remove-all), one helper-reuse suggestion, and two small nits. The first is the only one I'd push on before merge; the rest are optional.

integrations/openclaw/marmot/src/channel.ts
erskingardner: Documented the intentional asymmetry: `deleteByMessageId` performs the deletion and returns only a boolean, so it cannot safely supply a cached group for reactions. Current inbound messages inherit tool context; earlier outbound targets require explicit `to`.

integrations/openclaw/marmot/src/channel.ts
erskingardner: Replaced the hand-rolled strip with `normalizeMarmotTarget`, including trim/lowercase/`0x` handling and early clean failure for invalid groups. Added coverage for normalized and invalid targets.

erskingardner (approved):

PR #1405, opened 2026-08-13
perf(cgka-engine): answer send gates with indexed state probes

dannym-arx:
Every outbound send paid two full scans of the group's retained message window:

- `has_unresolved_convergence_inputs` listed and JSON-decoded every row at or after the anchor epoch, then classified each one, even though only rows in `Created` or `Retryable` state can gate a send.
- `retry_deferred_peels` listed every row from epoch 0 to find `PeelDeferred` rows, which are almost always absent.

This PR answers both gates with indexed queries:

- New `MessageStorage::has_messages_in_states` and `list_messages_in_states`, with default implementations on top of `list_messages`, so existing backends keep working unchanged.
- storage-sqlite overrides both with SQL state filters, served by a new `(group_id, state, epoch)` index (migration 0046).
- The send gate returns early on one `EXISTS` probe when the group holds no row in a gating state. The deferred-peel sweep decodes only `PeelDeferred` rows.

`OUTBOUND_GATING_STATES` keeps the fast path and `ClassifiedConvergenceInput::can_gate_outbound` in lockstep: the classifier is defined in terms of the constant, and an exhaustive test covers every `MessageState` variant.

No behavior change: the fast path skips work only when the full scan would have found nothing, and the sweep sees the same rows it saw before. Marmot colonies with long chat histories feel this on every send.

Verified with `just fast-ci`, `cargo test -p storage-sqlite`, `cargo test -p cgka-traits`, and `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks`.

## Summary by CodeRabbit

* **Performance**
  * Improved message processing by efficiently locating deferred and outbound-gating messages.
  * Added optimized filtering by message state and epoch for faster retained-message queries.

* **Reliability**
  * Clarified which message states qualify for outbound gating.
  * Added coverage for state classification, filtering behavior, ordering, and epoch boundaries.

* **Storage**
  * Added a database index to accelerate message lookups by group, state, and epoch.

### DIFF
--- a/crates/cgka-engine/src/convergence_input.rs
+++ b/crates/cgka-engine/src/convergence_input.rs
@@ -52,9 +52,17 @@ impl ClassifiedConvergenceInput {
     }
 
     fn can_gate_outbound(self) -> bool {
-        matches!(self.state, MessageState::Created | MessageState::Retryable)
+        OUTBOUND_GATING_STATES.contains(&self.state)
     }
 }
+
+/// The stored message states in which a retained input can gate outbound
+/// sends. `Engine::has_unresolved_convergence_inputs` passes this list to an
+/// indexed storage existence probe before paying a full list-and-decode scan;
+/// defining [`ClassifiedConvergenceInput::can_gate_outbound`] in terms of the
+/// same constant keeps the fast path and the classifier in lockstep.
+pub(crate) const OUTBOUND_GATING_STATES: [MessageState; 2] =
+    [MessageState::Created, MessageState::Retryable];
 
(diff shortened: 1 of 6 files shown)


Datawav (commented):
## Automated review

No actionable findings.

erskingardner (approved):
Recommend merge.

The indexed state probes preserve the prior classifier/list semantics, cover empty/state/epoch/order parity, and provide the shared storage API that #1416 should consume instead of duplicating. Exact head d11eadff reviewed; CI is green.

PR #1406, opened 2026-08-13
perf(storage-sqlite): scope retained-anchor snapshots to group state

dannym-arx:
Stacked on #1405.

`retain_current_group_epoch_snapshot` runs on every applied commit and captured a full group snapshot: the group record, OpenMLS values, and a copy of every retained message row and queued outbound intent, serialized into one blob inside the write transaction. Commit cost grew with the size of the retained history, and the message copies were dead weight:

- The convergence probe rewinds to an anchor and replays from inputs seeded before the rewind (`seed_stored_openmls_graph_inputs`). It never reads message rows after the rollback.
- The historical apply lists the live message and queue sets before the rewind and writes them back over whatever the rollback restored (`restore_live_message_and_queue_records`). The anchor's copies were deleted and replaced every time.

This PR scopes retained-anchor snapshots to canonical group state:

- New `MessageStorage::create_group_state_snapshot` captures the group record, member capabilities, convergence policy, validation marker, and group-scoped OpenMLS values. The default implementation falls back to a full snapshot, so other backends keep working unchanged.
- In storage-sqlite, the snapshot blob's `messages` and `queued_outbound` fields are now `Option`. `None` means the ledger was not captured, and rollback leaves the live `cgka_messages` and `cgka_queued_outbound` rows alone. Blobs written before this change carry `Some` and keep their old restore semantics, so existing databases need no migration.
- `retain_current_group_epoch_snapshot` uses the scoped capture. Probe snapshots, apply snapshots, and the conformance replay export stay full.

One deliberate semantic edge: rolling back to a legacy full anchor restored the anchor-time message image before the live rows were written back, which could resurrect rows deleted between the anchor and the apply. Scoped anchors never touch the ledger, so deleted rows stay deleted.

Every marmot burrow now stores one winter cache per epoch instead of dragging the whole meadow into it.

Verified with `just fast-ci`, `cargo test -p storage-sqlite`, `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks`, and `cargo test -p cgka-conformance-simulator`.

### DIFF
--- a/crates/storage-sqlite/src/storage/snapshots/rows.rs
+++ b/crates/storage-sqlite/src/storage/snapshots/rows.rs
@@ -11,8 +11,14 @@ use serde::{Deserialize, Serialize};
 #[derive(Serialize, Deserialize)]
 pub(super) struct Snapshot {
     pub(super) group: Group,
-    pub(super) messages: Vec<OrderedMessage>,
-    pub(super) queued_outbound: Vec<OrderedQueuedOutbound>,
+    /// `None` marks a state-scoped snapshot: the message ledger was not
+    /// captured, and rollback leaves the live `cgka_messages` rows (and their
+    /// pending application events) untouched. Full snapshots — including
+    /// every blob written before state-scoped capture existed — carry `Some`
+    /// and keep their capture-time restore semantics.
+    pub(super) messages: Option<Vec<OrderedMessage>>,
+    /// Same contract as `messages`, for `cgka_queued_outbound`.
+    pub(super) queued_outbound: Option<Vec<OrderedQueuedOutbound>>,
     pub(super) member_caps: Vec<MemberCapabilitiesSnapshot>,
     pub(super) convergence_policy: Option<Vec<u8>>,
     pub(super) validated_tree_marker: Option<Vec<u8>>,
--- a/crates/cgka-engine/src/convergence_input.rs
+++ b/crates/cgka-engine/src/convergence_input.rs
@@ -52,9 +52,17 @@ impl ClassifiedConvergenceInput {
     }
 
     fn can_gate_outbound(self) -> bool {
-        matches!(self.state, MessageState::Created | MessageState::Retryable)
+        OUTBOUND_GATING_STATES.contains(&self.state)
     }
 }
+
+/// The stored message states in which a retained input can gate outbound
+/// sends. `Engine::has_unresolved_convergence_inputs` passes this list to an
+/// indexed storage existence probe before paying a full list-and-decode scan;
+/// defining [`ClassifiedConvergenceInput::can_gate_outbound`] in terms of the
+/// same constant keeps the fast path and the classifier in lockstep.
+pub(crate) const OUTBOUND_GATING_STATES: [MessageState; 2] =
+    [MessageState::Created, MessageState::Retryable];
 
 #[derive(Default)]
 pub(crate) struct ConvergenceInputContext {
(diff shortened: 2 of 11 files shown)


crates/storage-sqlite/src/storage/snapshots/rows.rs
```diff
@@ -11,8 +11,14 @@ use serde::{Deserialize, Serialize};
...
     pub(super) group: Group,
-    pub(super) messages: Vec<OrderedMessage>,
-    pub(super) queued_outbound: Vec<OrderedQueuedOutbound>,
+    /// `None` marks a state-scoped snapshot: the message ledger was not
+    /// captured, and rollback leaves the live `cgka_messages` rows (and their
+    /// pending application events) untouched. Full snapshots — including
+    /// every blob written before state-scoped capture existed — carry `Some`
+    /// and keep their capture-time restore semantics.
+    pub(super) messages: Option<Vec<OrderedMessage>>,
+    /// Same contract as `messages`, for `cgka_queued_outbound`.
```
Datawav, line 20: Existing rollback snapshots have no compatible upgrade path. Changing persisted `messages` from `Vec<_>` to `Option<Vec<_>>` means previously stored blobs are still deserialized directly into the new `Snapshot` shape and may fail to decode after upgrade, making rewind/recovery unavailable. Please introduce a versioned snapshot envelope with a legacy decoder that maps old vectors to `Some(...)`, and add a checked-in pre-change blob compatibility test.

crates/storage-sqlite/src/storage/snapshots/rows.rs
```diff
@@ -11,8 +11,14 @@ use serde::{Deserialize, Serialize};
...
+    pub(super) queued_outbound: Option<Vec<OrderedQueuedOutbound>>,
     pub(super) member_caps: Vec<MemberCapabilitiesSnapshot>,
```
Datawav, line 22: Replay format version 1 now describes two wire shapes. These `Option<Vec<_>>` fields alter the nested serialized `Snapshot`, while `REPLAY_SNAPSHOT_VERSION` remains `1` and version-1 imports decode only the new structure. Existing version-1 exports may therefore fail to import. Please increment the replay version, retain an explicit version-1 decoder, and add fixtures covering both formats.

Datawav (commented):
## Automated review

Outcome: Two actionable persistence-format defects remain.

- High — Existing rollback snapshots lack a compatible upgrade path. `snapshots/rows.rs:19–21` changes persisted `messages` and `queued_outbound` from `Vec<_>` to `Option<Vec<_>>`, while `snapshots/restore.rs:65` still deserializes stored blobs directly into the changed `Snapshot` structure without versioning or a legacy decoder. After an upgrade, a previously retained rollback anchor may fail to decode, making rewind/recovery unavailable. Introduce a versioned snapshot envelope with a legacy decoder mapping old vectors to `Some(...)`, and lock compatibility with a checked-in pre-change blob test.

- Low — Replay format version 1 now describes two wire shapes. The same changed fields in `snapshots/rows.rs:19–21` alter the nested `Snapshot`, but `REPLAY_SNAPSHOT_VERSION` remains `1` at `rows.rs:47`; `restore.rs:75–83` accepts version 1 and decodes only the new structure. Existing version-1 replay exports may therefore fail import or become format-dependent. Increment the replay version and retain an explicit version-1 decoder, with fixtures covering both formats.

crates/storage-sqlite/src/storage/snapshots/rows.rs
dannym-arx: Legacy blobs decode cleanly today: both fields are always present in pre-change blobs (they were `Vec`, unconditionally serialized), and serde_json deserializes a JSON array into `Option<Vec<T>>` as `Some(vec)`. A decode failure requires a missing field, which no legacy blob has. So the versioned envelope would guard a state that cannot occur.

The compatibility test is a fair ask though. Added in 6d17f41a: `legacy_full_snapshot_blob_with_plain_array_fields_still_restores` builds a pre-change blob from scratch (plain JSON arrays, no `Option` anywhere in its construction), inserts it as a stored snapshot, and asserts rollback restores the full capture-time image.

crates/storage-sqlite/src/storage/snapshots/rows.rs
dannym-arx: Version 1 still describes one wire shape. The replay export always captures `SnapshotScope::Full`, and serde serializes `Some(vec)` identically to the old `Vec` (a plain array), so new version-1 exports are byte-shape-identical to old ones. On import, old exports decode into the new struct as `Some(..)` (covered by `legacy_full_snapshot_blob_with_plain_array_fields_still_restores` in 6d17f41a, which exercises the same nested `Snapshot` shape). `None` never appears in a replay export, so there is no second shape to version.

Datawav (dismissed):

erskingardner (dismissed):
Recommend merge after #1405.

The scoped snapshot encoding now preserves legacy array blobs as `Some`, leaves live message/queue rows untouched only for new scoped snapshots, and keeps full snapshots/replay exports on the old semantics. Exact head 6d17f41 reviewed; CI is green.

PR #1407, opened 2026-08-13
perf(marmot-uniffi): reuse timeline row conversions across window snapshots

dannym-arx:
Every live update on a timeline subscription reconverted the whole window: `TimelineMessagesSubscription::next()` (and both paginate calls) ran `window.snapshot().into()`, which re-parsed Markdown, re-walked `imeta` media, and re-sorted reactions for every row. One inbound message in a 100-row window cost about 100 markdown parses on the host's render path.

Two changes:

- `TimelineMessagesSubscription` keeps a conversion cache keyed by `message_id_hex`, storing each converted FFI row beside the source record that produced it. A snapshot conversion reuses the cached row whenever the source `TimelineMessageRecord` compares equal (it is `PartialEq`) and re-converts only changed rows, so reactions, edits, deletes, and delivery-state flips still invalidate exactly the rows they touch. The cache is replaced with the current page's rows on every conversion, so it is bounded by the window cap and rows that scroll out are dropped.
- The markdown AST to FFI converters now move strings instead of cloning them. `parse_markdown_document` is the only producer of these ASTs and hands the document over for good, yet the owned `From<MdDocument>` delegated to the borrowing converter and re-cloned every string in the tree. The unused borrowing impls are removed.

A new unit test pins the cache contract: unchanged rows come back identical, a changed row is re-converted (asserted through the parsed tokens, not just the plaintext), and rows that leave the window are pruned.

Marmots scrolling a busy burrow chat now pay for one new message, not the whole wall.

Verified with `just fast-ci`, `cargo test -p marmot-uniffi` (103 lib + 28 smoke tests), and `cargo test -p marmot-markdown`.

## Summary by CodeRabbit

- **Performance**
  - Improved timeline loading and updates by reusing unchanged content, making navigation and pagination more efficient.
  - Reduced unnecessary processing when viewing timeline pages.

- **Bug Fixes**
  - Ensured modified timeline items display their latest content instead of stale data.
  - Improved cleanup of outdated timeline entries as the visible window changes.
  - Preserved Markdown formatting, links, metadata, and embedded entities during conversion.

### DIFF
--- a/crates/marmot-uniffi/src/markdown.rs
+++ b/crates/marmot-uniffi/src/markdown.rs
@@ -218,33 +218,25 @@ fn markdown_input_within_ffi_limit(text: &str) -> LimitedMarkdownInput<'_> {
     }
 }
 
-impl From<&MdDocument> for MarkdownDocumentFfi {
-    fn from(value: &MdDocument) -> Self {
+/// Owned conversions throughout: the parse in [`parse_markdown_document`] is
+/// the only producer of these ASTs, and it hands the document over for good,
+/// so every string moves into the FFI value instead of being cloned. This
+/// halves the allocation cost of converting a parsed message.
+impl From<MdDocument> for MarkdownDocumentFfi {
+    fn from(value: MdDocument) -> Self {
         Self {
             blocks: value
                 .blocks
-                .iter()
+                .into_iter()
                 .map(|block| markdown_block_from_md(block, 0))
                 .collect(),
             truncated: false,
-            blank_lines_before: value.blank_lines_before.clone(),
+            blank_lines_before: value.blank_lines_before,
         }
     }
 }
 
-impl From<MdDocument> for MarkdownDocumentFfi {
-    fn from(value: MdDocument) -> Self {
-        (&value).into()
-    }
-}
-
-impl From<&MdBlock> for MarkdownBlockFfi {
-    fn from(value: &MdBlock) -> Self {
-        markdown_block_from_md(value, 0)
-    }
-}
-
-fn markdown_block_from_md(value: &MdBlock, depth: usize) -> MarkdownBlockFfi {
+fn markdown_block_from_md(value: MdBlock, depth: usize) -> MarkdownBlockFfi {
     if depth >= MAX_FFI_MARKDOWN_DEPTH {
         return MarkdownBlockFfi::Paragraph {
             inlines: Vec::new(),
@@ -255,7 +247,7 @@ fn markdown_block_from_md(value: &MdBlock, depth: usize) -> MarkdownBlockFfi {
             inlines: markdown_inlines_from_md(inlines, 0),
         },
         MdBlock::Heading { level, inlines } => MarkdownBlockFfi::Heading {
-            level: *level,
+            level,
             inlines: markdown_inlines_from_md(inlines, 0),
         },
         MdBlock::ThematicBreak => MarkdownBlockFfi::ThematicBreak,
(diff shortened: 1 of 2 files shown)


crates/marmot-uniffi/src/markdown.rs
```diff
@@ -218,33 +218,25 @@ fn markdown_input_within_ffi_limit(text: &str) -> LimitedMarkdownInput<'_> {
     }
 }
 
-impl From<&MdDocument> for MarkdownDocumentFfi {
-    fn from(value: &MdDocument) -> Self {
+/// Owned conversions throughout: the parse in [`parse_markdown_document`] is
+/// the only producer of these ASTs, and it hands the document over for good,
+/// so every string moves into the FFI value instead of being cloned. This
+/// halves the allocation cost of converting a parsed message.
+impl From<MdDocument> for MarkdownDocumentFfi {
```
Datawav, line 225: This removes the public borrowed conversions (`From<&MdDocument>`, `From<&MdBlock>`, `From<&MdListKind>`, `From<&MdListItem>`, `From<&MdTableCell>`, `From<&MdInline>`, and `From<&MdNostrEntity>`) in favor of owned-only implementations, so downstream Rust callers converting borrowed AST values will no longer compile. Please keep the owned conversions for the optimized parsing path, but restore borrowed implementations using clones or borrowed helpers and add compile-time coverage for them.

crates/marmot-uniffi/src/subscriptions.rs
```diff
@@ -358,6 +413,80 @@ mod tests {
...
+        );
+        assert_eq!(first.messages.len(), 2);
+        assert_eq!(cache.lock().unwrap().len(), 2);
+
+        // Unchanged rows come back identical from the cache.
+        let second = convert_timeline_page_cached(
+            &cache,
+            page(vec![record("m1", "hello"), record("m2", "world")]),
+        );
+        assert_eq!(second.messages[0].plaintext, "hello");
```
Datawav, line 469: This assertion checks only the returned plaintext, so it still passes if the cache is never reused and every unchanged row is reconverted. Likewise, the changed-row assertion below only requires nonempty token blocks, which the original Markdown also produces. Please instrument conversion or extract the cache-decision logic, then assert zero conversions for an unchanged page, exactly one for a changed row, and reconversion after eviction and reintroduction.

Datawav (commented):
## Automated review

Outcome: Two low-severity actionable findings.

- Low — Public borrowed conversions were removed. `crates/marmot-uniffi/src/markdown.rs:221–239, 306–325, 343–359, 442–450` replaces `From<&MdDocument>`, `From<&MdBlock>`, `From<&MdListKind>`, `From<&MdListItem>`, `From<&MdTableCell>`, `From<&MdInline>`, and `From<&MdNostrEntity>` with owned-only implementations. Rust callers converting borrowed AST values will no longer compile. Keep the owned conversions on the optimized parsing path, but restore borrowed implementations that clone or use borrowed helpers; add compile-time coverage for the supported borrowed conversions.

- Low — The new regression test does not establish cache reuse or changed-Markdown reparsing. In `crates/marmot-uniffi/src/subscriptions.rs:464–479`, the second identical page asserts only equivalent plaintext, while the edited row asserts only that token blocks are nonempty; both pass if every row is reconverted, and the original plaintext also produces nonempty blocks. A future regression could silently eliminate the intended optimization without failing this test. Add conversion instrumentation or extract the cache-decision logic, then assert zero conversions for an unchanged page, exactly one for a changed row, and reconversion after eviction and reintroduction.

erskingardner (approved):
Recommend merge.

The bounded cache invalidates on full source-record equality and prunes with the window; the owned conversion path removes avoidable AST string clones. The already-posted requests for measurable cache-hit coverage and restoration of borrowed Rust conversion impls are worthwhile but non-blocking for the UniFFI behavior. Exact head 9417687 reviewed; CI is green.

PR #1408, opened 2026-08-13
perf(agent-stream-compose): stop echoing the full transcript per command

dannym-arx:
Stacked on the `uniffi-window-cache` PR.

Every `Append`/`Status`/`Progress` command on a compose session responded with a clone of `StreamComposeReport`, whose `text` field holds the entire accumulated transcript. That is O(n^2) bytes copied over a stream's life: a 1 MB transcript appended in 1 KB chunks copies about 500 MB. The `wn-agent` connector discards everything but the ack, and the daemon only printed the response.

These commands now respond with `StreamComposeAck`: the report minus `text` (stream/group ids, status, transcript hash, chunk count, live error). The session still accumulates the full transcript internally, and `Finish` keeps returning and validating the complete report, so the final-transcript expectation contract is unchanged.

One visible change, called out deliberately: the daemon's `compose-append` JSON output no longer contains a `text` key. Its only consumer was the TUI's "streaming N bytes" counter, which read its own text back off the daemon. The TUI now tracks flushed bytes locally (`StreamComposer::flushed_bytes`) and shows the cumulative streamed total, where it previously showed only the last ack's size.

The marmot relay line now passes along a receipt, not a copy of every acorn hauled so far.

Verified with `just fast-ci`, `cargo test -p agent-stream-compose`, `cargo test -p agent-connector`, and the wn-cli stream and TUI suites (34 + 433 tests) with `test-policy-overrides`.

### DIFF
--- a/crates/agent-stream-compose/src/lib.rs
+++ b/crates/agent-stream-compose/src/lib.rs
@@ -45,6 +45,43 @@ pub struct StreamComposeReport {
     pub error: Option<String>,
 }
 
+/// Per-command acknowledgement for `Append`/`Status`/`Progress`.
+///
+/// [`StreamComposeReport`] minus the accumulated `text`. The session still
+/// appends every chunk to its full transcript internally (`Finish` returns
+/// and validates it), but responding with the report cloned the entire
+/// transcript so far on every command — O(n²) bytes copied over a stream's
+/// life, for a value both the connector and the daemon discard. Callers that
+/// need the final text get it from `Finish`.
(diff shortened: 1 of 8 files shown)


Datawav (commented):
## Automated review

No actionable findings.

erskingardner (dismissed):
Recommend merge after #1407.

The slim ack removes transcript-sized per-command copies while preserving the complete Finish report and retryable final validation. The TUI increments its byte count only after a successful append. Exact head 05aa54f reviewed; CI is green.

Issue #1415, opened 2026-08-13
Tracking: storage-format v2 — eliminate JSON double-expansion in message records and snapshots


erskingardner:
## Summary

Profiling of the group-creation and welcome-join flows (criterion harness added in f82dd183, `crates/cgka-engine/benches/group_lifecycle.rs`) found that once algorithmic waste was removed from those flows (0168fcf3, b85b6342), the dominant remaining cost is the **established JSON storage format**: a double number-array expansion of byte fields in stored message records, and a versionless, multi-megabyte snapshot blob that re-serializes them. At 32 invitees, **~50ms of the ~70ms `create_group`** is JSON encode/decode/IO, versus ~12ms of actual MLS crypto. Fixing it requires a storage-format migration, so it was deferred from the perf pass and is tracked here.

## Findings (all measured, release profile, in-memory SQLite)

**Double-expansion of stored message records** — one 16,727-byte MLS Welcome becomes a 180,002-byte `cgka_messages` row (**10.8×**):

1. `StoredMessagePayload::encode` (`crates/traits/src/message.rs:145`) is `serde_json::to_vec(self)`; `TransportMessage.payload: Vec<u8>` becomes a JSON array of decimal numbers → 59,924 bytes (3.6×).
2. `put_message` re-serializes the containing `MessageRecord` (`crates/storage-sqlite/src/storage/messages.rs:17`, `codec.rs:157`); its `payload: Vec<u8>` number-array-encodes the layer-1 JSON a second time → 180,002 bytes (another 3.0×).

The layering is deliberate (engine owns payload schema, storage owns record schema) — but `serde_bytes` does **not** help: under `serde_json`, `serialize_bytes` still emits a number array.

**Founding snapshot** — `retain_current_epoch_snapshot_for_group` (`crates/cgka-engine/src/openmls_projection.rs:2659` → `crates/storage-sqlite/src/storage/snapshots/capture.rs`) runs at every founding create and every canonical epoch advance. At 32 invitees: ~18ms capture (reads back all 32 welcome records, ~5.8MB of JSON, deserializing each into typed `MessageRecord`s) + ~11ms serialize (two-pass `serialize_sensitive_json`, a deliberate exact-capacity zeroizing-buffer design) + ~6.5ms insert of a **~6MB blob**. The `Snapshot` struct (`snapshots/rows.rs`) is versionless today.

[...]

Issue #1414, opened 2026-08-13
ops: storage-format v2 migration runbook, upgrade/downgrade policy, and compatibility test matrix


erskingardner:
## Context

Part of the storage-format v2 effort tracked in #1415. The format changes touch durable bytes of every existing account-device database, so the migration needs an explicit operational story **before** any of it ships. This issue owns that runbook and the compatibility test matrix; the sibling child issues link here from their PRs.

## Compatibility posture (why this is safe to do at all)

- The database is one-per-account-device (`crates/storage-sqlite` AGENTS invariant) and never crosses the wire — **no protocol or interop risk with other Marmot clients, relays, or app versions**. The only compat surface is a local app binary downgraded after its database was migrated.
- Downgrade protection already exists: `reject_unknown_future_migrations` (`crates/storage-sqlite/src/migrations.rs`) hard-fails an old binary opening a DB migrated by a newer build. The versioning issue extends the same posture to format versions.

## Scope of this issue

1. **Upgrade / downgrade policy**, documented in `release.md` and the release notes:
   - upgrade: automatic at session open via the numbered migration; expected duration for large histories (measure on a seeded fixture; report in the runbook),
   - downgrade: not supported across the format-version boundary — old binary refuses with a clear, typed error (never a corrupt or half-read database); document the user-facing symptom and the remedy (re-upgrade, or restore from pre-upgrade backup),
   - recommend a pre-upgrade backup/export step in the release notes for the cohort release that carries this.
2. **Backfill operations** (with the payload-column issue):
   - eager vs lazy decision recorded with rationale,
   - progress logging for long backfills (aggregate counters only — privacy-safe per `docs/marmot-architecture/overview/observability.md`: no account ids, group ids, message ids),
   - resumability after crash-kill mid-backfill,
   - post-backfill `VACUUM` decision (SQLCipher: keyed connection; duration on large DBs; when to schedule it — open-time vs explicit maintenance action).
3. **Release sequencing**: rides a cohort release (`just release-all <version>`) so MDK / wn-agent / MarmotKit carry the same storage-sqlite; call out the format bump explicitly in the release notes; workspace/version bumps stay manual per repo policy.
4. **Compatibility test matrix** (owning crate: `storage-sqlite`, plus engine crash-recovery patterns):
   - v1 DB → v2 binary: opens, migrates once, identical reads (golden-fixture DB checked into test fixtures or generated by a pinned older build),
[...]

Issue #1410, opened 2026-08-13
storage-sqlite: storage-format versioning infrastructure (format_version + versioned snapshot/payload codecs)


erskingardner:
## Context

Part of the storage-format v2 effort tracked in #1415. Profiling of the group-creation flow (criterion benches in `crates/cgka-engine/benches/group_lifecycle.rs`, landed in f82dd183) showed that ~50ms of the ~70ms `create_group` cost at 32 invitees is JSON encode/decode/IO of message records and snapshot blobs, caused by a double number-array expansion of byte fields (raw 16.7KB Welcome → 60KB `StoredMessagePayload` JSON → 180KB stored `MessageRecord` JSON) and a versionless 6MB snapshot blob.

Everything after this issue changes bytes on disk. Before any of it lands we need a single, explicit versioning story so that:

- new code can read old rows/blobs (rolling upgrade),
- old code fails loudly — never silently — when handed new-format data (no downgrade corruption),
- crash-during-migration resumes idempotently.

## What exists today

- `crates/storage-sqlite/src/migrations.rs`: `cgka_schema_migrations` (version, name, applied_at) table + `reject_unknown_future_migrations`, which already hard-fails an old binary opening a DB migrated by a newer build ("database was migrated by a newer storage-sqlite version"). This is the right rail; schema-migration versions alone are not expressive enough for *format* versioning because format changes here are not strictly 1:1 with schema migrations (snapshot blobs, payload codec).
- `cgka_group_snapshots.snapshot` blobs: the `Snapshot` struct (`crates/storage-sqlite/src/storage/snapshots/rows.rs`) is versionless JSON. Only the conformance-replay wrapper has a version (`REPLAY_SNAPSHOT_VERSION`).
- `StoredMessagePayload::decode` (`crates/traits/src/message.rs`) already demonstrates the new-then-legacy decode-fallback pattern we want at every format boundary.

## Scope of this issue

1. Add a storage-format version marker, persisted in the database (e.g. a `cgka_storage_meta` table with a `format_version` key — distinct from schema-migration rows), read once at open.
2. Define the version policy:
   - `format_version` 1 = today's layout everywhere.
   - Binaries refuse (clear, typed error) any database whose `format_version` is newer than they understand — same posture as `reject_unknown_future_migrations`.
   - Format upgrades are driven by ordinary numbered migrations; the format version is bumped by the migration that introduces the first v2 layout.
[...]

PR #1416, opened 2026-08-13
perf(cgka-engine): measured join/create lifecycle speedups + criterion bench harness

erskingardner:
## Summary

Benchmark-driven performance pass over the group-creation and welcome-join flows, plus a criterion harness to keep them measured. Each commit is bisectable and was verified against the benches (and the full engine suite) individually; no behavior changes.

Rebased onto master after **#1405** merged: this PR no longer carries any state-filter infrastructure (trait method / SQLite override / parity tests) — it reuses `MessageStorage::list_messages_in_states` from #1405, whose `(group_id, state, epoch)` index (migration 0046) also serves the replay filter here.

**Headline numbers** (release profile, in-memory SQLite, pass-through peeler):

| Benchmark | Before | After | Change |
| --- | --- | --- | --- |
| `join_welcome` / 1 stored key package | 2.29ms | 1.27ms | **−44%** |
| `join_welcome` / 8 stored key packages | 2.22ms | 1.28ms | **−42%** |
| `join_welcome` / 32 stored key packages | 2.40ms | 1.28ms | **−47%** |
| `join_welcome_large_group` / 32 members | 5.01ms | 4.43ms | **−12%** (replay filter on the #1405 index) |
| `create_group` / 32 invitees | 65.3ms | 64.8ms | ~flat (see "what was not done") |

## Commits

1. **`8ae514f8` test: criterion benchmarks** — `crates/cgka-engine/benches/group_lifecycle.rs`: `create_group` at 1/8/32 invitees, `join_welcome` at 1/8/32 stored key packages, 32-member group joins. Measurement rail the perf commits cite.
2. **`a5dac290` perf: point-query welcome key-package matching** — `do_join_peeled_welcome` previously enumerated every stored KeyPackage bundle through an **unindexed** `label` filter on `openmls_values` (scanning the account's whole MLS state table), then JSON-decoded and re-hashed each bundle. Now it does the same indexed primary-key point lookup OpenMLS's `keys_for_welcome` does, in the same first-match order, inside the same join transaction. Matching semantics unchanged.
3. **`0bb92a7e` perf: state-filtered replay scan** — `replay_buffered_messages` now calls #1405's `list_messages_in_states` so the end-of-join replay no longer JSON-parses the just-written ~180KB welcome record — or the whole group history on re-join — only to skip those rows. The replay loop keeps its state guard; behavior unchanged on every backend.
4. **`eeebb2eb` refactor: drop dead group-context snapshot** — `do_create_group` built a `GroupContextSnapshot` (an `export_secret` KDF + app-dict parse) and discarded it unread on both profile paths; also removes a spurious failure mode. Behavior unchanged; measured effect below bench noise, but it ran on every create.
5. **`d9464291` style**, **`47c332c9` docs**, **`d766ebd4` test cleanup** — fmt, bench documentation in `tests/AGENTS.md`, and removal of profiling probes that had leaked into the create bench closure (CodeRabbit catch).

## What was evaluated and deliberately not done

- **Welcome-payload clone hoist** in the create/invite wrap loops: measured ~0.02% — 50× below bench noise. Reverted rather than merged as a theoretical win.
- **Per-invitee member-id re-derivation dedup**: <0.6% of create. Dropped.
- **Staged-commit account-proof re-validation**: deliberate seam-hardening duplication (repo invariant). Kept.
- **The real create bottleneck — JSON storage double-expansion** (16.7KB welcome → 60KB → 180KB stored row) and the **~6MB founding snapshot blob** (~39ms of create): that's the established storage format, i.e. a migration, not a perf tweak. Filed as tracking issue **#1415** with child issues #1410–#1414 (versioning infra, payload BLOB column, snapshot blob v2, bench gates, ops runbook).

## Test plan

- [x] `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks` — all green (25 binaries)
- [x] `cargo test -p storage-sqlite` — green
- [x] `cargo test -p cgka-traits` — green
- [x] `just fast-ci` — fmt/check/clippy workspace-wide green
- [x] Benches re-run after each commit and again after the #1405 rebase; numbers above are reproducible via `cargo bench -p cgka-engine --bench group_lifecycle`

Note: `tests/deferred_peel_lifecycle.rs` requires the `test-policy-overrides` feature and fails without it on plain `master` too (pre-existing; CI runs with the feature set above).

## Follow-up

Storage-format v2 migration is tracked in #1415 (children #1410–#1414) — versioning infrastructure, `cgka_messages` payload BLOB column, snapshot blob v2, bench acceptance gates, and the migration ops runbook.

## Summary by CodeRabbit

* **Performance**
  * Improved group creation and membership handling for more efficient lifecycle operations.
  * Reduced unnecessary message processing during replay, helping deferred and retryable messages be handled more efficiently.

* **Testing**
  * Added benchmarks covering group creation, invitations, and joins across different group sizes.

* **Documentation**
  * Added guidance for running convergence-policy tests and maintaining the group lifecycle benchmarks.

### DIFF
--- a/crates/cgka-engine/benches/group_lifecycle.rs
+++ b/crates/cgka-engine/benches/group_lifecycle.rs
@@ -1,0 +1,309 @@
+//! Criterion benchmarks for the group creation and welcome-join flows.
+//!
+//! These benches drive the real OpenMLS-backed engine over in-memory SQLite
+//! storage with the same pass-through peeler style the Tier-2 tests use, so
+//! the measured work is the engine's own CPU + storage cost: KeyPackage
+//! parsing/validation, commit/welcome production, per-invitee wrapping, and
+//! the join transaction. Transport crypto is deliberately excluded.
+//!
+//! Run:
+//!
+//! ```sh
+//! cargo bench -p cgka-engine --bench group_lifecycle
+//! ```
+
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use cgka_engine::account_identity_proof::{
+    AccountIdentityProofRequest, AccountIdentityProofSigner,
+};
+use cgka_engine::{Engine, EngineBuilder};
+use cgka_traits::engine::{CgkaEngine, CreateGroupRequest, KeyPackage, SendResult};
+use cgka_traits::error::PeelerError;
+use cgka_traits::group::ProtocolProfile;
+use cgka_traits::group_context::GroupContextSnapshot;
+use cgka_traits::ingest::{PeeledContent, PeeledMessage};
+use cgka_traits::peeler::TransportPeeler;
+use cgka_traits::transport::{
+    EncryptedPayload, Timestamp, TransportEnvelope, TransportMessage, TransportSource,
+};
+use cgka_traits::types::{MemberId, MessageId};
+use criterion::{BatchSize, BenchmarkId, Criterion, criterion_group, criterion_main};
+use k256::schnorr::{SigningKey, signature::hazmat::PrehashSigner};
+use sha2::{Digest, Sha256};
+use storage_sqlite::SqliteAccountStorage;
+
+// ── Deterministic identities (mirrors tests/support) ────────────────────────
+
+fn proof_signer(seed: &[u8]) -> Arc<dyn AccountIdentityProofSigner> {
+    Arc::new(TestAccountIdentityProofSigner(signing_key(seed)))
+}
+
+fn signing_key(seed: &[u8]) -> SigningKey {
+    let mut counter = 0u64;
+    loop {
+        let mut material = [0u8; 32];
+        let mut hasher = Sha256::new();
+        hasher.update(b"cgka-engine-test-identity-v1");
+        hasher.update(seed);
+        hasher.update(counter.to_be_bytes());
+        material.copy_from_slice(&hasher.finalize());
+        if let Ok(sk) = SigningKey::from_bytes(&material) {
+            return sk;
+        }
+        counter += 1;
+    }
+}
+
+struct TestAccountIdentityProofSigner(SigningKey);
+
+impl AccountIdentityProofSigner for TestAccountIdentityProofSigner {
+    fn sign_account_identity_proof(
+        &self,
+        request: &AccountIdentityProofRequest,
+    ) -> Result<[u8; 64], String> {
+        if self.0.verifying_key().to_bytes().as_slice() != request.account_identity.as_slice() {
+            return Err("request account identity does not match test key".into());
+        }
+        let signature = self
+            .0
+            .sign_prehash(&request.proof_event_id()?)
+            .map_err(|e| e.to_string())?;
+        Ok(signature.to_bytes())
+    }
+}
+
+fn pad32(name: &[u8]) -> Vec<u8> {
+    signing_key(name).verifying_key().to_bytes().to_vec()
+}
+
+// ── Pass-through peeler (no transport crypto) ───────────────────────────────
+
+fn hash_id(bytes: &[u8]) -> MessageId {
+    use std::collections::hash_map::DefaultHasher;
+    use std::hash::{Hash, Hasher};
+    let mut h = DefaultHasher::new();
+    bytes.hash(&mut h);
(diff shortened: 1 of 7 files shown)


crates/traits/src/storage.rs
```diff
@@ -208,6 +208,25 @@ pub trait MessageStorage {
         at_or_after_epoch: EpochId,
     ) -> StorageResult<Vec<MessageRecord>>;
 
+    /// Return stored messages for `group_id` at or after `at_or_after_epoch`
+    /// whose state is one of `states`, in the same deterministic replay order
+    /// as [`Self::list_messages`]. Callers driving retry/replay loops only act
+    /// on a small subset of states; backends that can filter server-side avoid
+    /// fetching and decoding rows (whose payloads can be large) that the
+    /// caller would discard. The default implementation filters in memory.
+    fn list_messages_in_states(
```
erskingardner: [P1] Drop this duplicate and build on #1405. That open perf PR already adds `MessageStorage::list_messages_in_states`, the SQLite implementation and parity tests, plus `has_messages_in_states` and migration 0046 for the `(group_id, state, epoch)` index. This copy also chooses a different argument order, so the two PRs cannot merge cleanly and maintaining either copy after conflict resolution risks splitting the public storage contract. Rebase after #1405, keep this PR’s replay call site, and remove its trait/SQLite/test duplicate.

erskingardner (commented):
Found serious issues.

The state-filtered storage API/SQLite work duplicates #1405 with an incompatible signature, so these PRs cannot both land as written. Rebase this PR on #1405 and retain only the join-replay call-site optimization here. Also remove the unresolved `TEMP-PROBE` timers and `eprintln!` calls from the measured Criterion closure; their I/O invalidates the create-group benchmark numbers. The key-package point lookup and dead context-snapshot removal otherwise look sound at this head.

crates/cgka-engine/benches/group_lifecycle.rs
```diff
@@ -0,0 +1,375 @@
...
+                        eprintln!("TEMP-PROBE closure clone: {:?}", __tc.elapsed()); // TEMP-PROBE
+                        let __tl = std::time::Instant::now(); // TEMP-PROBE
+                        let mut guard = alice.lock().await; // TEMP-PROBE
+                        eprintln!("TEMP-PROBE closure lock: {:?}", __tl.elapsed()); // TEMP-PROBE
+                        let __tg = std::time::Instant::now(); // TEMP-PROBE
+                        let (group_id, result) = guard
+                            .create_group(request)
+                            .await
+                            .expect("create_group succeeds");
+                        eprintln!("TEMP-PROBE closure create_group: {:?}", __tg.elapsed()); // TEMP-PROBE
```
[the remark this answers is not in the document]
erskingardner: Good catch — the profiling probes leaked into the committed bench. Removed in 6046f7e0; the measured closure is now pure engine work (full-file sweep shows no remaining TEMP-PROBE lines, and the create bench re-ran clean after removal).

crates/traits/src/storage.rs
erskingardner: Done — rebased onto master now that #1405 has merged (49a20a92). This PR no longer adds `list_messages_in_states`, the SQLite override, or the parity test; the replay scan in `replay_buffered_messages` now calls #1405's shared method with its argument order, and its `(group_id, state, epoch)` index serves the filter. Force-pushed as `0bb92a7e` (call site) within the rebased set ending `d766ebd4`. Benches re-verified after the rebase: join_welcome_large_group 5.01ms baseline -> 4.43ms.

erskingardner (commented):
Recommend merge.

Fresh review of exact head `d766ebd`: the #1405 trait/SQLite duplication is gone, the replay call site uses the merged shared API with the correct argument order, and the profiling probes have been removed from the measured closure. The point lookup preserves OpenMLS first-match/error ordering inside the join transaction, the replay filter preserves actionable-state ordering, and the context snapshot removal does not remove current-profile or routing validation. CI is green; no actionable findings.

GitHub does not allow this account to approve its own PR, so this is recorded as a comment rather than an approval.

PR #1418, opened 2026-08-13
perf(marmot-app): save once per delivery unless route retirement changed state

dannym-arx:
Stacked on the `seen-events-index` PR.

`ingest_received_delivery` persisted the full account state twice for every membership-changing delivery: once behind the `routes_dirty` checkpoint, then unconditionally again after `refresh_group_routes`. Each save snapshots every group plus the 16k-entry seen-events ring, so the double write was pure duplication whenever route reconciliation left persisted state untouched.

The root cause is that `refresh_group_routes` conflated two different outcomes in one bool: the in-memory routing-table delta (which obligates a relay-subscription refresh, not a save) and prior-route retirement via `prune_prior_nostr_routes` (which mutates persisted group state and does obligate a save; its return value was silently dropped). This PR splits them into `GroupRouteRefresh { routing_changed, state_pruned }`.

- The delivery path saves a second time only when the first save did not run or pruning actually mutated state.
- The `sync_inner` and catch-up route-reconciliation sites gain the `state_pruned` trigger. That fixes a latent miss: retirement previously went unpersisted unless the routing table happened to change in the same pass.
- Checkpoint paths whose unconditional boundary save already covers pruning keep their behavior unchanged.

The marmot stamps the winter ledger once per delivery, not twice for the same acorns.

Verified with `just fast-ci` and `cargo nextest run -p marmot-app` (743 passing; the 5 relay_runtime failures are this machine's known environment failures and reproduce on clean master).

### DIFF
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -84,6 +84,20 @@ where
     results.into_iter().map(|(_, result)| result).collect()
 }
 
+/// Outcome of one [`AppClient::refresh_group_routes`] pass, separating the
+/// in-memory routing-table delta from durable-state mutation: only the latter
+/// obligates a state save, only the former obligates a relay-subscription
+/// refresh.
+#[derive(Clone, Copy, Debug, Default)]
+pub(crate) struct GroupRouteRefresh {
+    /// The in-memory routing table changed; relay subscriptions need a
+    /// `sync_runtime_groups` refresh.
+    pub(crate) routing_changed: bool,
+    /// `prune_prior_nostr_routes` removed retired prior routes from persisted
+    /// group state; a save is required for the retirement to stick.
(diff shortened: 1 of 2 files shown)


erskingardner (approved):
Reviewed exact head 42ea035e4f2b6c80bc131ca53352961a83d6cbd6 against stacked base 08932f136bbf129dd6261b1c492fd8ee3a81c0da.

The split between in-memory routing changes and durable route pruning is correctly propagated: the live-delivery path retains its pre-refresh checkpoint, saves a second time only when pruning mutates state, and every caller that projects only `routing_changed` already has an unconditional boundary save covering pruning. All CI checks pass.

No actionable findings in this differential. This approval is scoped to #1418; stacked base #1417 still has an unresolved, unrelated review thread that should be resolved before the stack lands.

PR #1419, opened 2026-08-13
Fix nightly app-witness oracle and retain failure evidence

### LINKED ISSUE #1409: Simulator nightly failed (2026-08-13)
The simulator nightly failed. Run: https://github.com/marmot-protocol/mdk/actions/runs/31669097875

erskingardner:
## Summary

- reconcile the convergence reliability plan with merged route unification, queued-send scheduling, and campaign-performance work
- fix the seed-7 app-witness oracle by asserting exact payload multiplicities without inventing an MLS cross-sender delivery order
- bump the adversarial catalog generator contract to v4 and pin the new order-neutral contract structurally
- make a failed nightly lane reproduce the exact adversarial catalog into durable reports/capsules before propagating failure
- raise the hosted runner's memlock limit for the nightly processes so SQLCipher memory locking does not amplify one warning per allocation

## Root cause

The nightly run in #1409 reached the same canonical branch, epoch, roster, and payload set in both failing tests. It failed only because Carol observed Frank's and Eve's messages in the opposite order from the `ClientState.received_payloads` expectation. MLS does not define a total order between independent senders.

The fix deliberately preserves the ordered `received_payloads` contract used by transport-reorder scenarios. Only `adversarial-reliability/app-witness-value/v1` switches to exact `PayloadCount` assertions: one Eve witness, one Frank witness, and zero David witnesses.

## Scope

Simulator scenario/oracle data, its tests/docs, the tracking plan, and the scheduled workflow only. No production convergence, app-runtime, storage, or protocol behavior changes.

## Verification

- `cargo nextest run -p cgka-conformance-simulator --features test-policy-overrides --test adversarial_reliability_campaigns --locked` (12 passed, 3 ignored)
- `just fast-ci`
- `actionlint .github/workflows/simulator-nightly.yml`
- docs stale-language scan from `docs/AGENTS.md`

Closes #1409.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nightly reliability validation with clearer failure handling and diagnostic reports.
  * Corrected adversarial scenario checks to validate per-sender payload counts without requiring unsupported cross-sender ordering.
  * Updated reliability test coverage for app-witness scenarios.

* **Documentation**
  * Updated reliability scenarios and architecture progress notes, including current milestone status and validation guidance.
  * Documented expanded workload shapes and clarified payload-ordering expectations.

### DIFF
--- a/.github/workflows/simulator-nightly.yml
+++ b/.github/workflows/simulator-nightly.yml
@@ -51,7 +51,32 @@ jobs:
         uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
 
       - name: Run nightly convergence lane
-        run: just convergence-nightly-lane
+        id: convergence_lane
+        continue-on-error: true
+        run: |
+          # SQLCipher deliberately locks sensitive allocations. Give the test
+          # process enough memlock headroom so the runner does not amplify one
+          # warning per allocation when its inherited limit is exhausted.
+          sudo prlimit --pid "$$" --memlock=unlimited:unlimited
+          just convergence-nightly-lane
+
+      - name: Capture failed adversarial scenarios
+        if: steps.convergence_lane.outcome == 'failure'
+        continue-on-error: true
+        run: |
+          sudo prlimit --pid "$$" --memlock=unlimited:unlimited
+          cargo run -p cgka-conformance-simulator \
+            --bin cgka-conformance-simulator-report --locked -- \
+            --family adversarial-reliability/v1 \
+            --seed 7 \
+            --cases 12 \
+            --out target/cgka-conformance-simulator-nightly-reports \
+            --storage file \
+            --strict-oracle
+
+      - name: Propagate nightly convergence failure
+        if: steps.convergence_lane.outcome == 'failure'
+        run: exit 1
 
       - name: Run simulator doctests
         run: cargo test -p cgka-conformance-simulator --doc --locked
--- a/crates/cgka-conformance-simulator/SCENARIOS.md
+++ b/crates/cgka-conformance-simulator/SCENARIOS.md
@@ -741,11 +741,14 @@ regression, covers a new semantic edge, or is the smallest readable example of a
 
 ### Adversarial Reliability Catalog
 
-`adversarial-reliability/v1` (generator version `3`) rotates through twelve deterministic workload shapes: retained-history offline floods;
+`adversarial-reliability/v1` (generator version `4`) rotates through twelve deterministic workload shapes: retained-history offline floods;
 sustained app/proposal/commit traffic; self-update versus admin progress; a named unrecoverable losing-branch invite;
 unequal relay reconciliation; restart boundaries plus the engine's real SIGKILL durable-phase matrix; multi-group noisy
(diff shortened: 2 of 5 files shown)


erskingardner (commented):
**Review verdict: would approve** (posting as comment — GitHub blocks self-review on your own PR).

Verified the diagnosis against the failed nightly run 31669097875: both failing tests (`full_engine_witness_ab_pair_can_select_different_canonical_branches` and `remaining_catalog_workloads_execute_under_the_strict_oracle`) run generated case 9 and failed with the same `client_state_mismatch` — carol observed `["frank-witness-9", "eve-witness-9"]` vs the expected `["eve-witness-9", "frank-witness-9"]` at the same epoch 2 / 5 members. An invalid cross-sender ordering claim, exactly as described.

The fix is sound and minimal:
- `received_payloads: None` skips only the ordered payload comparison (`is_none_or` in `TraceExpectation::compare`); epoch and member count are still asserted on carol.
- The three `PayloadCount` asserts (`pending_payload_count` over `MessageReceived` events) pin the exact multiset. Since this scenario sends exactly three app payloads, asserting eve=1/frank=1/david=0 loses no precision versus the old ordered list — while no longer inventing a cross-sender order MLS does not define.
- Case rotation is seed-independent (`case_index % 12`), so the new structural test on seed 7 / case 9 deterministically pins the order-neutral contract and blocks regression to an ordered `received_payloads` here.
- Ordered `received_payloads` contracts remain in place for transport-reorder scenarios (code comment + SCENARIOS.md).
- Workflow mechanics check out: capture-step CLI flags all exist, portable failure capsules are written by default on failure (sensitive replay correctly stays opt-in), and the pre-existing `if: always()` upload step makes the captured reports/capsules durable even after the propagate step fails the job. `continue-on-error` + `steps.convergence_lane.outcome == 'failure'` is used correctly.
- The memlock problem is real — the failed run's stderr is flooded with `sqlcipher_mlock: mlock() returned -1 errno=12` — and `sudo prlimit --pid $$` is inherited by the whole lane. One non-blocking note inline.
- Generator version 3→4 is consistently updated; no stale pins in README/vectors/hardening workflow (this family is generated, not fixture-pinned, so no vector bumps needed). Tracking-plan updates verified against the merged PRs they cite (#1293, #1403, #1405, #1406, #1416).

All CI green at bb7f729.

.github/workflows/simulator-nightly.yml
```diff
@@ -51,7 +51,32 @@ jobs:
...
 
       - name: Run nightly convergence lane
-        run: just convergence-nightly-lane
+        id: convergence_lane
+        continue-on-error: true
+        run: |
+          # SQLCipher deliberately locks sensitive allocations. Give the test
+          # process enough memlock headroom so the runner does not amplify one
+          # warning per allocation when its inherited limit is exhausted.
+          sudo prlimit --pid "$$" --memlock=unlimited:unlimited
```
erskingardner, line 60: Non-blocking: the `Run encrypted file-backed vector report` step below also executes SQLCipher-backed runs (`--storage file`) on the success path but doesn't get the memlock raise, so its logs can still carry the `sqlcipher_mlock` warning flood. Worth the same one-line `sudo prlimit` there for consistency with this step's stated goal.

Issue #1413, opened 2026-08-13
benches: storage-format v2 measurement rail and acceptance targets


erskingardner:
## Context

Part of the storage-format v2 effort tracked in #1415. The format changes (#1411 payload column, #1412 snapshot v2) land behind benches so the win is demonstrated, not assumed — same standard as the perf commits that found this (0168fcf3, b85b6342).

## Measured baseline (release profile, in-memory SQLite, pass-through peeler)

From `cargo bench -p cgka-engine --bench group_lifecycle`:

| Benchmark | Baseline (pre-format-work) |
| --- | --- |
| `create_group/1 invitee` | ~1.77ms |
| `create_group/8 invitees` | ~8.6ms |
| `create_group/32 invitees` | ~65ms — of which ~13ms welcome-record writes + ~39ms founding snapshot |
| `join_welcome/1 stored KP` | ~1.27ms (post point-query fix) |
| `join_welcome/32 stored KPs` | ~1.28ms (post point-query fix) |
| `join_welcome_large_group/32 members` | ~4.26ms (post replay-scan fix) |

Stored-size anchors: 16,727-byte raw Welcome → 180,002-byte `cgka_messages` row; ~6MB founding snapshot blob.

## Scope of this issue

1. Extend `crates/cgka-engine/benches/group_lifecycle.rs`:
   - add a re-join-with-history join case (the shape the replay-scan fix and payload column most affect),
   - add a canonical-advance (commit/confirm) bench so the per-epoch snapshot-retention cost is measured outside founding, scaled by retained history.
2. Extend the startup scaling benches (`just bench-startup-scaling`, mdk#1161) to report DB file size for welcome/message-heavy fixtures, so the ~8–10× size claim is measured on a realistic fixture.
3. Record pre/post numbers for each format child issue in its PR description; acceptance targets:
   - `create_group/32`: ~65ms → ≤ ~20ms,
   - founding snapshot blob: ~6MB → well under 1MB,
   - welcome record: 180KB → ~20–60KB,
   - no regression (>2%) on `join_welcome`, send, or ingest benches.
4. Optional CI wiring: a non-blocking bench job that posts criterion deltas on PRs touching `crates/storage-sqlite` or `crates/cgka-engine` (follow the existing `just bench-startup-scaling` MDK_BENCH line pattern rather than a new reporting stack).

erskingardner:
[...]

PR #1421, opened 2026-08-13
Implement versioned storage format v2

### LINKED ISSUE #1410: storage-sqlite: storage-format versioning infrastructure (format_version + versioned snapshot/payload codecs)

### LINKED ISSUE #1411: storage-sqlite: move cgka_messages payloads out of the JSON record into a BLOB column (format v2)
## Context

Part of the storage-format v2 effort tracked in #1415. Depends on #1410 (versioning infrastructure).

Measured on the criterion benches in `crates/cgka-engine/benches/group_lifecycle.rs` (release profile, in-memory SQLite):

- A founding Welcome for a 32-invitee group is **16,727 raw bytes → 59,924 bytes** after `StoredMessagePayload::encode` (`crates/traits/src/message.rs:145`, `serde_json::to_vec`; `Vec<u8>` becomes a JSON array of decimal numbers) → **180,002 bytes** after `put_message` re-serializes the containing `MessageRecord` (`crates/storage-sqlite/src/storage/messages.rs:17` → `codec::serialize`, `crates/storage-sqlite/src/codec.rs:157`) whose `payload: Vec<u8>` field number-array-encodes the layer-1 JSON *again*.
- **~10.8× total expansion per stored message**; the founding-creation transaction writes ~5.8MB for 32 invitees (~13ms), and every later reader (snapshot capture, hydration, replay scans, disband scans) pays the matching JSON-parse cost.

The double expansion is structural: the engine owns the payload schema, storage owns the record schema, and `payload` is opaque between them — so each layer re-expands the other's output. `serde_bytes` does **not** fix this (under `serde_json`, `serialize_bytes` still emits a number array).

## Scope of this issue

Move message payloads out of the JSON record into their own binary column.

1. New numbered migration (next after 0045, `crates/storage-sqlite/src/migrations.rs`):
   - `ALTER TABLE cgka_messages ADD COLUMN payload BLOB` (nullable).
   - Backfill strategy — decide and document: eager in-migration copy vs lazy rewrite-on-read with a bounded background sweep. Eager is simpler and atomic (existing runner wraps each migration in a transaction) but blocks session open for large histories; lazy keeps open fast at the cost of a long-lived dual format. Either way the migration must be resumable/idempotent under crash-kill (pattern: `crates/cgka-engine/tests/crash_recovery_sqlite.rs`).
2. Read path: prefer the `payload` column; fall back to extracting `payload` from the legacy JSON `record` for unmigrated rows (dual-read, required indefinitely or until an explicit later cutover).
3. Write path: after the migration, writes always populate the `payload` column and keep `record` as the small metadata JSON. Keep `MessageRecord`/`cgka_traits` API surface unchanged — this is internal to `storage-sqlite`.
4. Versioned payload codec (v2), introduced through the versioning infrastructure issue: store the payload as raw bytes where the consumer accepts the v2 marker, keeping the v1 JSON encode for the decode fallback (`StoredMessagePayload::decode` already has the new-then-legacy pattern to extend). The group-record/timeline tables owned by the app projections are **out of scope** here — engine message storage only.
5. Update every `cgka_messages` reader/writer, not just the happy path: `put_message`, `get_message`, `delete_message`, `list_messages`, `list_messages_in_states`, `update_message_state` (read-modify-write), snapshot capture/restore (`crates/storage-sqlite/src/storage/snapshots/capture.rs` + `restore.rs`), `stored_sent_welcome`/`outstanding_sent_welcomes`, disband welcome scans (`crates/cgka-engine/src/disband.rs`), and the conformance replay export/import feature.
6. VACUUM consideration: after backfill, page reclamation needs an explicit `VACUUM` (SQLCipher: keyed connection required). Decide: part of the migration, opt-in maintenance action, or follow-up issue. Document expected DB-size drop (~8–10× on welcome/message-heavy accounts).

## Acceptance

- `create_group/32` bench: founding-transaction welcome writes drop from ~13ms to low single-digit ms; stored record size ~180KB → ~20–60KB depending on payload-codec choice.
- Upgrade test: v1 database → opens, backfills (or lazily converges), reads identical results.
- Downgrade test: pre-migration binary refuses the migrated DB with a clear error (rides the versioning issue's policy).
- Full `cargo test -p storage-sqlite` and `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks` green.

### LINKED ISSUE #1412: storage-sqlite: snapshot blob format v2 (version-gated, raw-byte fields, preserved zeroization)
## Context

Part of the storage-format v2 effort tracked in #1415. Depends on #1410 (versioning infrastructure); complements #1411 (payload column), which removes the same double-expansion from the primary message table.

Measured on the criterion benches in `crates/cgka-engine/benches/group_lifecycle.rs` (32-invitee founding create):

- `retain_current_epoch_snapshot_for_group` (called at the end of every founding create and every canonical epoch advance — `crates/cgka-engine/src/openmls_projection.rs:2659`, `crates/storage-sqlite/src/storage/snapshots/capture.rs`) costs **~39ms of the ~70ms create**:
  - capture: ~18ms — reads back every `cgka_messages` row for the group (~5.8MB of JSON at 32 invitees) and `serde_json::from_slice`s each into typed `MessageRecord`s;
  - serialize: ~11ms — `serialize_sensitive_json` re-serializes the whole `Snapshot` struct into a **~6MB blob** (two-pass count-then-write, deliberate: exact-capacity zeroizing buffer so `Vec` realloc cannot leave copies of epoch secrets in the allocator — `crates/storage-sqlite/src/codec.rs:161`);
  - insert: ~6.5ms for the 6MB blob.
- The `Snapshot` struct (`crates/storage-sqlite/src/storage/snapshots/rows.rs`) is **versionless** — there is currently nothing to branch a new format on. Only the conformance-replay wrapper has a version (`REPLAY_SNAPSHOT_VERSION`).

The snapshot machinery runs on *every* canonical epoch advance, not just founding, so this cost recurs per commit on busy groups, scaled by retained message history.

## Scope of this issue

Introduce snapshot blob format v2 behind the version gate from the versioning issue.

1. Dispatch snapshot reads on the blob version (v1 = current unversioned JSON; v2 = new format). Never rewrite v1 blobs in place; read-dispatch keeps crash recovery of pre-upgrade snapshots intact.
2. v2 layout: store byte-heavy fields as raw bytes instead of nested JSON number arrays — concretely `OpenMlsValueSnapshot.value` and the message records' payloads (aligned with the payload-column issue's v2 payload codec). A binary codec (postcard/bincode/rmp) for the whole blob is acceptable if the sensitive-buffer guarantees below are preserved; field-level raw-byte storage is the lower-risk variant.
3. Preserve the zeroization posture: `SensitiveBytes` handling and the no-secret-copies-in-the-allocator guarantee of `serialize_sensitive_json` must carry over to v2 (single-pass serialization into a pre-sized or purpose-built zeroizing buffer is fine *only* if realloc cannot leak fragments — justify the chosen approach in the PR).
4. Preserve restore semantics exactly: `rollback_group_to_snapshot`, `capture_group_state` / `GroupStateCheckpoint` (checkpoints share `rows.rs` structures), and the `test-conformance-replay` export/import feature must round-trip identical canonical state across v1-read and v2-write mixes.
5. After the payload-column issue lands, snapshot capture should not need to JSON-round-trip message rows at all on the v2 path (capture raw, store raw, restore raw).

## Acceptance

- `create_group/32` bench: founding snapshot retention drops from ~39ms toward low single-digit ms; blob size ~6MB → well under 1MB.
- Mixed-format recovery test: snapshot written by v1 build, rolled back by v2 build (and vice versa) restores identical group state.
- Crash test: kill mid-snapshot-write leaves the prior anchor intact and usable (pattern: `crates/cgka-engine/tests/crash_recovery_sqlite.rs`).
- `cargo test -p storage-sqlite`, engine suites, and `just conformance` green.

erskingardner:
## Summary

- normalize `cgka_messages` so indexed state and metadata are authoritative scalar columns while opaque payload and deferred-peel data live in dedicated BLOBs
- add explicit, versioned `MDKMP` v2 message-payload and `MDKS` v2 snapshot/checkpoint codecs with legacy JSON reads and fail-closed tagged decoding
- preserve existing rows during migration 47 and expose bounded, atomic, resumable lazy promotion instead of rewriting account history at session open
- preserve zeroizing snapshot output, state-scoped retained anchors, full rollback semantics, and conformance-replay behavior
- document the compatibility model, downgrade refusal, backfill/VACUUM posture, release sequencing, and measured format decision

## Why

JSON encoded byte fields as decimal arrays twice: a representative 16,727-byte Welcome became a roughly 180 KB message row. State changes then read and double-parsed that complete record merely to flip one enum. Full rollback snapshots repeated the same expansion in a versionless blob.

This change keeps the storage boundary local to one account-device database. It does not change Marmot or MLS wire bytes. Migration 47 remains the database downgrade gate, while each opaque artifact carries its own version so mixed legacy/current rows and snapshots remain readable without a redundant global format flag.

## Measured impact

The stacked measurement rail in #1425 reports:

- `create_group/32`: approximately 65 ms at the tracker baseline, 28.09 ms immediately before v2, and 15.22 ms after v2
- representative normalized 32-member Welcome value: 16,821 bytes, down from the 180,002-byte tracker anchor
- largest 32-member founding snapshot: approximately 82.6 KB, down from roughly 6 MB
- payload encode/decode: approximately 40.4 us / 114.4 us JSON to 686 ns / 335 ns v2
- retained-anchor capture at 32 invitees: approximately 1.23–1.29 ms
- full rollback snapshot at 32 invitees: approximately 10.95–11.03 ms JSON to 3.62–4.40 ms v2
- `join_welcome/1 stored KP`: 1.30 ms; `join_welcome/32 stored KPs`: 1.33 ms; 32-member join: 3.94 ms

These satisfy the size and `create_group/32` targets in #1413. The stacked PR also adds the missing send, ingest, real rejoin-with-history, canonical-advance, and database-footprint rails.

## Compatibility and recovery

- new binaries read legacy message rows, payload JSON, snapshot JSON, and checkpoint JSON
- new writes use normalized rows and tagged v2 artifacts
- old binaries refuse migration 47 through the existing future-migration gate before account-table access
- bounded promotion batches are atomic and idempotent; malformed input rolls back the whole batch
- tagged unknown artifact versions fail closed and never retry as legacy JSON
- an encrypted file-backed schema-v46 database upgrades once while preserving group/message bytes and then refuses downgrade without touching account rows
- process-kill tests stop after migration-47 mutation but before commit, and after the first row of a promotion batch; reopen proves the transaction rolled back and retry converges without data loss

## Validation

- `just fast-ci` after rebase onto current `origin/master`
- `cargo test -p cgka-traits`
- `cargo test -p storage-sqlite --features test-conformance-replay` (368 passed, 2 child-process entry points ignored)
- focused file-backed upgrade, migration-kill, and promotion-kill tests
- `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks --test group_creation` (50 passed)
- `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks --test crash_recovery_sqlite` (3 passed, 1 child ignored)
- `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks --test publish_lifecycle` (18 passed)
- latest GitHub run is green across format/check/clippy, all four Rust test shards, simulator/vectors, relay runtime, CLI E2E, and Tamarin

A local full-feature `just ci` run previously exposed `cold_reopen_recovers_quiet_offline_history_without_a_live_trigger`; the same feature-set failure reproduced on unmodified `origin/master`. The latest GitHub matrix for this exact PR head is green.

## PR sequence

1. Review and merge this storage implementation into `master`.
2. Review #1425 as the single stacked measurement commit, then retarget it to `master` after this PR merges.
3. Keep large-history upgrade timing, progress reporting, optional compaction, and cohort release evidence in #1414.

The durability/spec contract is being handled independently in [marmot#416](https://github.com/marmot-protocol/marmot/pull/416); this PR remains an implementation-local storage change.

Closes #1410
Closes #1411
Closes #1412

Related: #1413, #1414, #1415

### DIFF
--- a/crates/storage-sqlite/src/storage/snapshots/checkpoints.rs
+++ b/crates/storage-sqlite/src/storage/snapshots/checkpoints.rs
@@ -1,9 +1,7 @@
-use super::{capture, restore, rows::GroupStateCheckpoint};
+use super::{capture, format as snapshot_format, restore};
 use crate::{
-    SqliteAccountStorage, SqliteResultExt,
-    codec::{SensitiveBytes, serialize_sensitive},
-    connection::retry_on_busy,
-    deserialize, epoch_to_i64,
+    SqliteAccountStorage, SqliteResultExt, codec::SensitiveBytes, connection::retry_on_busy,
+    epoch_to_i64,
 };
 use cgka_traits::storage::{GroupStateCheckpointRef, StorageError, StorageResult};
 use cgka_traits::types::{EpochId, GroupId};
@@ -34,7 +32,7 @@ fn create_on_connection(
     checkpoint: &GroupStateCheckpointRef,
 ) -> StorageResult<()> {
     let state = capture::capture_group_state(conn, group_id)?;
-    let blob = serialize_sensitive(&state)?;
+    let blob = snapshot_format::encode_checkpoint(&state)?;
     let existing = conn
         .query_row(
             "SELECT resulting_epoch, checkpoint
@@ -103,7 +101,7 @@ fn restore_on_connection(
         .storage()?
         .ok_or_else(|| StorageError::SnapshotMissing(checkpoint_id.to_owned()))?,
     );
-    let checkpoint: GroupStateCheckpoint = deserialize(blob.as_slice())?;
+    let checkpoint = snapshot_format::decode_checkpoint(blob.as_slice())?;
     restore::restore_group_state(conn, group_id, &checkpoint)
 }
 
--- a/crates/storage-sqlite/src/storage/snapshots/checkpoints.rs
+++ b/crates/storage-sqlite/src/storage/snapshots/checkpoints.rs
@@ -44,10 +44,15 @@ fn create_on_connection(
         .optional()
         .storage()?;
     if let Some((resulting_epoch, existing_blob)) = existing {
-        if resulting_epoch == epoch_to_i64(checkpoint.resulting_epoch)?
-            && existing_blob == blob.as_slice()
-        {
-            return Ok(());
+        if resulting_epoch == epoch_to_i64(checkpoint.resulting_epoch)? {
+            // Existing checkpoints can remain as untagged legacy JSON after
+            // upgrade. Canonicalize their decoded state before comparing so
+            // an exact retry stays idempotent across storage formats.
+            let existing_state = snapshot_format::decode_checkpoint(&existing_blob)?;
+            let existing_blob = snapshot_format::encode_checkpoint(&existing_state)?;
+            if existing_blob.as_slice() == blob.as_slice() {
+                return Ok(());
+            }
         }
         return Err(StorageError::AlreadyExists);
     }
--- a/docs/marmot-architecture/storage-format-v2.md
+++ b/docs/marmot-architecture/storage-format-v2.md
@@ -1,0 +1,190 @@
+---
+title: "Storage Format v2"
+created: 2026-08-13
+updated: 2026-08-13
+tags: [marmot, storage, sqlite, migration, encoding]
+status: current
+---
+
+# Storage Format v2
+
+This document owns MDK's local SQLCipher storage-format contract. It is not a
+Marmot wire format and does not constrain other Marmot implementations. The
+canonical protocol specifies which state must be reproducible; this document
+specifies how `storage-sqlite` currently persists that state.
+
+## Versioning model
+
+There is no database-wide `format_version` independent of schema migrations.
+That value would be ambiguous while rows and opaque artifacts intentionally
+coexist at different versions.
+
+- `cgka_schema_migrations` is the database compatibility gate. Every writer
+  change that an older binary cannot safely read lands with a numbered
+  migration. `reject_unknown_future_migrations` makes an older binary refuse
+  the database before it reads or writes application tables.
+- Every independently decoded opaque artifact has its own version. A binary
+  envelope starts with an artifact-specific magic value and a fixed-width
+  network-order version. Untagged legacy JSON is recognized only at the
+  artifact's documented v1 fallback boundary.
+- Readers accept supported old versions. Writers emit only the newest version.
+  Unsupported or malformed versions fail closed; readers never guess from a
+  partially decoded value.
+- Golden byte fixtures pin each binary artifact. Schema migration tests pin
+  old-row reads and downgrade refusal.
+
+## Binary grammar
+
+MDK-local binary artifacts reuse the Marmot binary grammar for consistency:
+
+- fields appear in the documented order;
+- fixed-width integers use network byte order;
+- variable byte strings use the shortest canonical QUIC-varint byte-length
+  prefix;
+- list fields are one QUIC-length-prefixed byte vector containing concatenated
+  item encodings;
+- option fields use a one-byte discriminator (`0` absent, `1` present);
+- decoders reject non-minimal length prefixes, unknown discriminants, truncated
+  values, invalid UTF-8 for string fields, and trailing bytes.
+
+Using the same grammar is an MDK implementation convention. It does not make
+these storage artifacts Marmot protocol values.
+
+## `cgka_messages` authority model
+
+Migration `0047_normalized_message_records` supports two row formats:
+
+| Column | Format 1 | Format 2 |
+| --- | --- | --- |
+| `id`, `group_id`, `epoch`, `state` | indexed copy; legacy `record` remains the decoded authority | authoritative |
+| `storage_format` | `1` | `2` |
+| `record` | complete JSON `MessageRecord` | `NULL` |
+| `payload` | `NULL` | exact `MessageRecord.payload` bytes |
+| `deferred_peel` | `NULL`; represented inside `record` | optional JSON `DeferredPeelLifecycle` metadata |
+
+Existing rows remain format 1 at migration time so opening a large account does
+not perform an unbounded history rewrite. Current writes are format 2. Updating
+a legacy row promotes it atomically; updating a format-2 state changes only the
+scalar `state` and, when leaving `PeelDeferred`, clears `deferred_peel`. It does
+not read, decode, copy, or rewrite `payload`.
+
+The retained legacy `record` column is a compatibility read path, not an
+ongoing second authority. `promote_legacy_message_rows` promotes at most 256
+rows in one atomic, idempotent batch and reports only aggregate progress so a
+host can schedule it away from account-open latency. Removing format-1 reads
+requires a later migration and release decision.
+Reclaiming freed SQLCipher pages is an explicit maintenance operation; opening
+an account never runs `VACUUM` implicitly.
+
+## `StoredMessagePayload` v2
+
+New `MessageRecord.payload` values use this envelope:
+
+```text
+opaque magic[5] = "MDKMP";
+uint16 version = 2;
+uint8 variant;
+PayloadVariant body;
+```
+
+Variant values are stable:
+
+| Value | Body |
+| --- | --- |
+| `0` | one `TransportMessage`, `RawTransport` |
+| `1` | one `TransportMessage`, `OutboundWelcome` |
+| `2` | one `TransportMessage`, `OpenMlsWire` |
+| `3` | exact `TransportMessage`, OpenMLS `TransportMessage`, optional own-commit stamp, optional own-application stamp |
+| `4` | one `TransportMessage`, required own-commit stamp |
+
+A stored `TransportMessage` contains, in order:
+
+```text
+opaque id<V>;
+opaque payload<V>;
+uint64 timestamp;
+MessageId causal_dependencies<V>;
+opaque source_utf8<V>;
+uint8 envelope_variant; // 0 group message, 1 welcome
+opaque envelope_id<V>;  // transport group id or recipient id
+```
+
+Each `MessageId` inside `causal_dependencies` is itself a length-prefixed opaque
+value. An own-commit stamp contains the length-prefixed committer id, a one-byte
+priority (`0` privileged, `1` ordinary), a vector of length-prefixed UTF-8
+proposal references, optional checkpoint id, and optional resulting epoch
+authenticator. An own-application stamp contains the sender id and source epoch
+authenticator.
+
+Each field is limited to at most `2^30` bytes and each decoded list to at most
+`2^20` items. These are defensive local resource bounds, not protocol limits.
+
+The decoder falls back from a missing `MDKMP` magic to the previous tagged JSON
+envelope and then the oldest bare-`TransportMessage` JSON shape. A value that
+starts with `MDKMP` never falls back after a binary decoding failure.
+
+## Snapshot v2
+
+Retained-anchor snapshots no longer copy the message ledger or queued outbound
+work. Their 32-invitee capture cost measured roughly 1.23 ms. Full rollback
+snapshots still include the live message ledger and measured roughly 11.0 ms,
+so that path justified a binary format.
+
+Snapshot and group-state checkpoint v2 start with `MDKS`, network-order `uint16` version `2`, and then
+length-delimited group JSON, optional normalized message rows, optional queued
+outbound rows, member capabilities, optional convergence-policy and validation
+bytes, and raw OpenMLS key/value rows. Message ids, group ids, payloads, and
+OpenMLS fields remain raw bytes rather than JSON number arrays. The exact
+encoded length is computed before allocating the zeroizing output buffer; the
+encoder cannot grow it. Temporary list bodies that can contain secret material
+also use zeroizing ownership.
+
+A blob without `MDKS` is decoded as the corresponding unversioned JSON v1
+snapshot or checkpoint. A blob with `MDKS` never falls back following a v2
+error. Full snapshots, state-scoped snapshots, and branch-addressed checkpoints
+use the envelope. State-scoped snapshots and checkpoints encode the message and
+queue options as absent; checkpoints additionally reject rollback-only message,
+queue, or convergence-policy fields during decoding.
+
+## Decision evidence
+
+Measured on 2026-08-13 in the release profile on the same local machine and
+Criterion harness used to identify the original regression:
+
+| Measurement | Current result |
+| --- | --- |
+| `create_group/32 invitees` | 15.62–15.73 ms; the pre-v2 current-head baseline was about 28.09 ms |
+| retained-anchor capture, empty group | 177–185 us |
+| retained-anchor capture, 32 invitees | 1.23–1.29 ms |
+| full rollback snapshot, 32 invitees, JSON | 10.95–11.03 ms |
+| full rollback snapshot, 32 invitees, v2 | 3.62–4.40 ms |
+| representative 16,727-byte Welcome, JSON payload envelope | 67,254 bytes |
+| representative 16,727-byte Welcome, explicit MDK v2 envelope | 16,821 bytes |
+| MDK v2 encode / decode | about 686 ns / 335 ns |
+| JSON encode / decode | about 40.4 us / 114.4 us |
+
+A Postcard 1.1.3 spike produced 16,810 bytes and encoded in about 6.42 us, but
+could not deserialize the existing internally tagged `StoredMessagePayload`
+Serde shape (`WontImplement`). Supporting it would therefore require a second
+Postcard-specific DTO/schema in addition to the explicit compatibility
+contract. The explicit MDK format is smaller than JSON, faster than both tested
+encoders, already decodes the production type, and has stable golden bytes, so
+Postcard is not adopted.
+
+The group-state-only retained-anchor capture is roughly 1.25 ms at the
+32-invitee fixture and the headline create benchmark is below the 20 ms target.
+Snapshot v2 is justified by the separately measured full rollback path, not by
+the obsolete pre-#1406 claim that retained anchors still contain messages.
+
+## Operational policy
+
+- Upgrade is automatic through numbered transactional migrations.
+- Downgrade across migration 47 is unsupported: an older binary refuses the
+  newer migration before reading account data.
+- Migration 47 changes table shape atomically but does not backfill history, so
+  session-open time is bounded independently of message-history size.
+- The promotion sweep uses batches of at most 256 rows, aggregate privacy-safe
+  progress, and an idempotent transaction; malformed input rolls back the
+  whole batch.
+- `VACUUM` is opt-in maintenance on an already keyed connection and is never
+  part of automatic session open.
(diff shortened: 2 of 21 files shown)


Datawav:
Remediation stopped because the PR head advanced during diagnosis from `1c6eb5d3f8c4c2f78464940f635fec4ae5a83341` to `c15c1bf9be7e16b71d2966f92ee5ae069b13ee8d`. No commit was pushed.

On the stale head, CI showed two failures: the direct-output audit flagged test-only `println!` crash sentinels in `migrations.rs` and `storage/messages.rs`, and `retained_media_rehydrates_a_retired_current_epoch_before_the_group_advances` failed because epoch 2 remained retired. The new head should be evaluated by fresh CI.

Datawav (commented):
## Automated review

No actionable findings.

Datawav:
Remediation verification completed at exact head `c15c1bf9be7e16b71d2966f92ee5ae069b13ee8d`.

The assigned automated audit reported no actionable findings, so no code or documentation changes were made and no commit was pushed. The worktree remained clean. No local tests were run because there was no remediation change; the current GitHub check suite is passing, including Rust format, check, clippy, and all four Rust test shards.

crates/storage-sqlite/src/storage/snapshots/checkpoints.rs
```diff
@@ -34,7 +32,7 @@ fn create_on_connection(
     checkpoint: &GroupStateCheckpointRef,
 ) -> StorageResult<()> {
     let state = capture::capture_group_state(conn, group_id)?;
-    let blob = serialize_sensitive(&state)?;
+    let blob = snapshot_format::encode_checkpoint(&state)?;
```
erskingardner, line 35: Switching this write to `encode_checkpoint` makes the idempotency check just below (line 48, `existing_blob == blob.as_slice()`) format-sensitive. After upgrade, `cgka_group_state_checkpoints` can still hold untagged legacy JSON blobs — `decode_checkpoint` accepts both formats, so they persist. If a checkpoint is re-created with logically identical state while its stored blob is legacy JSON, the byte comparison fails and this returns `StorageError::AlreadyExists` instead of the `Ok(())` the exact-retry path produced pre-upgrade (and that the "Exact retry is idempotent" test still asserts same-format).

I traced the production caller: only `confirm_published` creates checkpoints, inside the durable confirm transaction, and I could not construct a post-upgrade re-create of a pre-upgrade checkpoint through it — a committed confirm also merged the pending commit, so a restart retry skips this block. The reachable blast radius therefore looks narrow. But the storage-level contract regression is real, and a spurious `AlreadyExists` is a durable, non-transient error that would wedge that confirm rather than recover from it.

Cheap fix that keeps the contract format-independent: when the epoch matches, decode `existing_blob` via `snapshot_format::decode_checkpoint` and compare the decoded state to the captured state (or decode-then-re-encode and compare bytes if `GroupStateCheckpoint` should not grow a `PartialEq`). If you believe the cross-format path is unreachable, a comment stating why is an acceptable alternative. (CodeRabbit flagged the same issue but it landed outside the diff; surfacing inline so it is not lost.)

docs/marmot-architecture/storage-format-v2.md
```diff
@@ -0,0 +1,190 @@
...
+Snapshot v2 is justified by the separately measured full rollback path, not by
+the obsolete pre-#1406 claim that retained anchors still contain messages.
+
+## Operational policy
+
+- Upgrade is automatic through numbered transactional migrations.
+- Downgrade across migration 47 is unsupported: an older binary refuses the
+  newer migration before reading account data.
+- Migration 47 changes table shape atomically but does not backfill history, so
+  session-open time is bounded independently of message-history size.
```
erskingardner: This overstates the open-time guarantee. Migration 47's `INSERT INTO cgka_messages_new ... SELECT ... FROM cgka_messages` physically copies every row — including the multi-hundred-KB legacy JSON `record` blobs — inside one transaction at first open. There is no per-row decode/re-encode, but the I/O is proportional to history size, with matching WAL growth and transient disk usage. For exactly the large, message-heavy accounts this PR targets, first open after upgrade will take a while and needs roughly 2x transient disk. The same misstatement appears at line 65 ("does not perform an unbounded history rewrite") and in `release.md` item 5.

Suggest rewording along the lines of: "changes table shape atomically via a one-time physical copy proportional to history size (no per-row decode or promotion); per-row promotion to format 2 is deferred and batched" — plus a transient-disk/WAL note so operators can plan the cohort upgrade.

release.md
```diff
@@ -153,6 +153,33 @@ not compiled into MarmotKit; apps supply them at runtime.
...
+3. Run crash/failure-injection tests at every multi-row transformation or
+   artifact replacement boundary. Schema migrations are transactional;
+   background promotion work must additionally be row-idempotent and bounded.
+4. State explicitly that downgrade across the migration is unsupported. The
+   remedies are to re-upgrade or restore a pre-upgrade account database/export;
+   never advise manually removing migration rows.
+5. Recommend a pre-upgrade backup/export in the cohort release notes when the
+   release rewrites existing durable data. Migration 47 only changes table
+   shape and retains legacy message rows, so it does not perform a history-size
+   open-time rewrite.
```
erskingardner: The rationale mischaracterizes migration 47. Rebuilding `cgka_messages` rewrites every existing row's pages — a history-size physical rewrite at first open — even though row bytes are preserved and no payload backfill runs. By this item's own gate ("when the release rewrites existing durable data"), the migration-47 cohort release should either carry the pre-upgrade backup recommendation or explicitly justify waiving it on transactional, crash-tested grounds. "Does not perform a history-size open-time rewrite" is inaccurate and will mislead release-note readers planning upgrades for large accounts.

erskingardner (commented):
**Requesting changes** (posted as a comment — GitHub does not allow a formal change request on one's own PR).

Reviewed at c15c1bf: full diff plus surrounding context (the `confirm_published` caller of `create_group_state_checkpoint`, 0001/0003/0046 schema lineage, codec bounds and canonical-varint enforcement, `retry_on_busy` semantics). CI at this exact head is green across the whole matrix.

This is a strong, well-scoped PR. The normalized-row design matches #1411; the MDKMP/MDKS envelopes are bounded, canonical (minimal-width varint enforcement), and fail-closed; the zeroizing, exact-capacity output posture is preserved on the secret-bearing paths; migration 47 follows the established 0003 rebuild pattern with correct index parity (`idx_cgka_messages_group_epoch`, `idx_cgka_messages_group_state_epoch`); promotion is bounded, atomic, and idempotent with process-kill tests at both the migration and promotion boundaries. Two items should be resolved before merge — both small (inline).

Non-blocking notes:
- The existing CodeRabbit nitpicks are valid and cheap: a codec-local snapshot round-trip test, an unknown-version MDKMP test, and the missing `checkpoints.rs` row in `crates/storage-sqlite/src/storage/snapshots/AGENTS.md`.
- `StoredMessagePayload::encode/decode` changing their error type is a public signature change on `cgka-traits`. Every in-workspace caller Debug-formats or discards the error, so nothing breaks here, but it is a (minor, presumably deliberate) deviation from #1411's "keep the cgka_traits API surface unchanged" scope note — worth a one-line acknowledgement in the PR body.
- Validation basis: remote CI at this head; I did not run a local build.

docs/marmot-architecture/storage-format-v2.md
```diff
@@ -0,0 +1,190 @@
...
+| `storage_format` | `1` | `2` |
+| `record` | complete JSON `MessageRecord` | `NULL` |
+| `payload` | `NULL` | exact `MessageRecord.payload` bytes |
+| `deferred_peel` | `NULL`; represented inside `record` | optional JSON `DeferredPeelLifecycle` metadata |
+
+Existing rows remain format 1 at migration time so opening a large account does
+not perform an unbounded history rewrite. Current writes are format 2. Updating
+a legacy row promotes it atomically; updating a format-2 state changes only the
+scalar `state` and, when leaving `PeelDeferred`, clears `deferred_peel`. It does
+not read, decode, copy, or rewrite `payload`.
```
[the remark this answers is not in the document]
erskingardner: Fixed in b08f53fc. The architecture and release docs now distinguish deferred payload promotion from migration 47’s one-time history-sized physical table copy, including lock duration, WAL growth, temporary free-space, backup, and large-account validation requirements.

docs/marmot-architecture/storage-format-v2.md
```diff
@@ -0,0 +1,190 @@
...
+encoded length is computed before allocating the zeroizing output buffer; the
+encoder cannot grow it. Temporary list bodies that can contain secret material
+also use zeroizing ownership.
+
+A blob without `MDKS` is decoded as the corresponding unversioned JSON v1
+snapshot or checkpoint. A blob with `MDKS` never falls back following a v2
+error. Full snapshots, state-scoped snapshots, and branch-addressed checkpoints
+use the envelope. State-scoped snapshots and checkpoints encode the message and
+queue options as absent; checkpoints additionally reject rollback-only message,
+queue, or convergence-policy fields during decoding.
```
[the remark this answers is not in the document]
erskingardner: Fixed in b08f53fc. The format contract now describes the three scopes separately: full rollback snapshots include messages/queue; state-scoped snapshots leave live operational rows untouched; branch checkpoints additionally omit convergence policy and reject rollback-only fields.

release.md
```diff
@@ -153,6 +153,33 @@ not compiled into MarmotKit; apps supply them at runtime.
...
+6. Do not run `VACUUM` during automatic account open. If page reclamation is
+   useful, expose it as explicit keyed-connection maintenance and report its
+   expected duration and temporary free-space requirement.
+
+Storage migrations do not require a workspace version bump during feature
+development. The eventual release version remains a manual release operation.
+
```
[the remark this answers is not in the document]
erskingardner: Fixed in b08f53fc. The storage-format release checklist now explicitly requires `just fast-ci` before push and leaves the full `just ci` matrix to GitHub CI.

crates/storage-sqlite/src/storage/snapshots/checkpoints.rs
erskingardner: Fixed in b08f53fc. Same-epoch existing checkpoints are decoded and canonically re-encoded before comparison, making exact retries format-independent. Added `identical_legacy_json_group_state_checkpoint_retry_is_idempotent`; focused checkpoint tests and `just fast-ci` pass.

docs/marmot-architecture/storage-format-v2.md
erskingardner: Fixed in b08f53fc. Migration 47 is now documented as a transactional physical copy proportional to history size, with no per-row decode/promotion but history-scaled lock, WAL, and temporary-space costs.

release.md
erskingardner: Fixed in b08f53fc. The cohort checklist now recommends pre-upgrade backup/export and representative large-account validation for duration, WAL growth, and temporary free space.

erskingardner (commented):
**Approving — recommend merge** (posted as a comment; GitHub does not allow a formal approval on one's own PR).

Re-reviewed at b08f53f. Both requested changes from my previous review are resolved, and CI at this head is green (25 success, 1 release-job skip).

1. **Checkpoint idempotency** — the decode-then-canonicalize comparison restores format-independent logical equality: canonical v2 encoding makes byte equality equivalent to decoded-state equality, the epoch gate still short-circuits first, and an undecodable stored blob now surfaces as a Serialization error rather than a spurious conflict, without mutating anything. The new regression test covers exactly the cross-format retry scenario.
2. **Upgrade-cost documentation** — storage-format-v2.md and release.md item 5 now state the one-time, history-proportional physical rebuild honestly (lock duration, WAL growth, transient free space), recommend the pre-upgrade backup for the migration-47 cohort, and require large-account validation before release. The state-scoped snapshot vs. checkpoint field semantics are clearer as well.

Remaining open items are the pre-existing trivial CodeRabbit nitpicks (codec-local round-trip test, unknown-version MDKMP test, snapshots/AGENTS.md module-map row, env-var constants) — all optional, author's call.

PR #1423, opened 2026-08-13
Process convergence applications at source epoch

### LINKED ISSUE #1171: Investigate delayed application messages expiring during multi-commit catch-up
## Problem

MDK retains application-message secrets for five previous MLS epochs. During catch-up, a client may receive an application message and a sequence of commits while the message is still deferred or waiting for replay. If enough commits are applied before that application message is processed, the required epoch secret can be evicted and the message becomes permanently undecryptable.

The convergence replay path orders accepted commits before accepted application messages, so a single catch-up pass may advance beyond the retained window before attempting application-message delivery.

## Questions to investigate

- Can a delayed message that was initially within the retention window expire during the same convergence/catch-up cycle?
- Are raw `PeelDeferred` messages vulnerable to retained epoch snapshots being pruned before retry?
- Is expiry surfaced clearly in audit data, or can the message appear processed/ignored without an actionable reason?
- Should replay ordering, secret retention, or buffering semantics change to guarantee delivery for messages admitted within the window?

## Suggested regression coverage

Queue interleaved application messages and more than five sequential commits for an offline client, then catch up in one run. Verify which messages are delivered, which expire, and that every expiration is explicit and correctly classified.

erskingardner:
## Summary

- process pending proposals and application messages at their source epoch before replaying the commit that advances the candidate
- order the selected branch's final replay by epoch and content kind so accepted applications are authenticated before source-epoch material is pruned
- use the frozen convergence pass's retained anchor as bounded active-work retention without introducing another long-lived secret store
- document that active retention preserves authentication ability but does not make an expired application witness eligible

## Root cause

Candidate materialization previously replayed the full commit path before probing pending application messages. Final canonical apply similarly replayed accepted proposals and commits before accepted applications. A sufficiently large retained commit burst could therefore advance OpenMLS beyond the application's source epoch and prune the cryptographic material needed to authenticate or deliver that application.

This made delivery depend on convergence-pass partitioning: processing five commits and the application in one pass followed by another commit could succeed, while processing the same six commits and application in one frozen pass could lose the application.

## Behavior

For each candidate epoch E, replay now processes:

1. proposals at E
2. applications at E
3. the commit advancing E to E+1

Branch-specific application observations remain available for witness scoring, while application delivery still occurs only when the selected canonical branch is applied. Witness eligibility continues to use the adopted candidate-tip application window, preserving the existing forward-secrecy policy boundary.

Fixes #1171.

## Validation

- `cargo test -p cgka-engine --features test-policy-overrides --test distributed_convergence` — 85 passed
- focused six-advance one-pass versus split-pass regression
- canonical-branch application delivery and branch-witness tests
- `just fast-ci`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved replay consistency across competing branches and message arrival orders.
  - Correctly authenticates applications using their source epoch.
  - Handles late and pre-branch application messages, including retained history.
  - Ensures applications are delivered once on the selected branch.
  - Ignores undecryptable applications without blocking branch processing.
  - Clarified witness eligibility at branch-specific epochs.

- **Documentation**
  - Added details on replay behavior and convergence scenarios.

- **Tests**
  - Added regression coverage for split, one-pass, and contested-branch delivery.

### DIFF
--- a/crates/cgka-engine/src/openmls_projection.rs
+++ b/crates/cgka-engine/src/openmls_projection.rs
@@ -2802,16 +2802,6 @@ fn replay_messages_for_canonicalization_result<S: StorageProvider>(
         let record = storage
             .get_message(&message_id)
             .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?;
-        // An accepted proposal below the apply-start epoch was consumed by a
-        // commit in the skipped already-applied prefix: its effect is inside
-        // the live state, and MLS proposals are epoch-scoped, so replaying it
-        // against the post-prefix state would be rejected (WrongEpoch) and
-        // fail the whole apply. Skip the replay; its `Processed` disposition
-        // still persists so it leaves the convergence input set.
-        if result.accepted_proposals.contains(hex_message_id) && record.epoch.0 < apply_start_epoch
-        {
-            continue;
-        }
         let Some(message) = openmls_wire_message_from_record(&record)? else {
             return Err(OpenMlsProjectionError::Decode(format!(
                 "accepted message {} is not a stored OpenMLS wire payload",
@@ -2819,6 +2809,22 @@ fn replay_messages_for_canonicalization_result<S: StorageProvider>(
             )));
         };
         let projection = project_mls_message(&message.payload)?;
+        let source_epoch =
+            projection
+                .source_epoch
+                .ok_or(OpenMlsProjectionError::UnsupportedMessageKind(
+                    projection.kind,
+                ))?;
+        // An accepted proposal below the apply-start epoch was consumed by a
+        // commit in the skipped already-applied prefix: its effect is inside
+        // the live state, and MLS proposals are epoch-scoped, so replaying it
+        // against the post-prefix state would be rejected (WrongEpoch) and
+        // fail the whole apply. Use the authenticated wire epoch rather than
+        // mutable row metadata. Skip the replay; its `Processed` disposition
+        // still persists so it leaves the convergence input set.
+        if result.accepted_proposals.contains(hex_message_id) && source_epoch < apply_start_epoch {
+            continue;
+        }
         let kind_order = match projection.kind {
             // Proposals and applications from epoch E must be processed while
             // the group is still at E. The commit advancing E -> E+1 comes
@@ -2829,7 +2835,7 @@ fn replay_messages_for_canonicalization_result<S: StorageProvider>(
             OpenMlsContentKind::Commit => 2,
             OpenMlsContentKind::Welcome | OpenMlsContentKind::Other => 3,
         };
-        replay_messages.push((record.epoch.0, kind_order, input_order, message));
+        replay_messages.push((source_epoch, kind_order, input_order, message));
     }
     replay_messages.sort_by_key(|message| (message.0, message.1, message.2));
     Ok(replay_messages
@@ -3043,6 +3049,9 @@ fn candidate_paths_with_pending_replay_messages(
     for proposals in proposals_by_epoch.values_mut() {
         proposals.sort_by(|a, b| a.id.as_slice().cmp(b.id.as_slice()));
     }
+    for applications in applications_by_epoch.values_mut() {
+        applications.sort_by(|a, b| a.id.as_slice().cmp(b.id.as_slice()));
+    }
 
     candidate_paths
         .iter()
@@ -3051,16 +3060,20 @@ fn candidate_paths_with_pending_replay_messages(
                 let mut seen = BTreeSet::new();
                 let mut messages = Vec::new();
                 let mut final_epoch = None;
-                let mut first_commit_epoch = None;
-                for message in &path.messages {
-                    let projection = project_mls_message(&message.payload)?;
-                    if projection.kind == OpenMlsContentKind::Commit {
-                        first_commit_epoch = Some(projection.source_epoch.ok_or(
+                let projected_messages = path
+                    .messages
+                    .iter()
+                    .map(|message| Ok((message, project_mls_message(&message.payload)?)))
+                    .collect::<Result<Vec<_>, OpenMlsProjectionError>>()?;
+                let first_commit_epoch = projected_messages
+                    .iter()
+                    .find(|(_, projection)| projection.kind == OpenMlsContentKind::Commit)
+                    .map(|(_, projection)| {
+                        projection.source_epoch.ok_or(
                             OpenMlsProjectionError::UnsupportedMessageKind(projection.kind),
-                        )?);
-                        break;
-                    }
-                }
+                        )
+                    })
+                    .transpose()?;
                 // A late application from common pre-fork history can be
                 // older than every commit in the candidate path. Probe it
                 // against the retained base state, before the first commit;
@@ -3079,8 +3092,7 @@ fn candidate_paths_with_pending_replay_messages(
                         }
                     }
                 }
-                for message in &path.messages {
-                    let projection = project_mls_message(&message.payload)?;
+                for (message, projection) in projected_messages {
                     if projection.kind == OpenMlsContentKind::Commit {
                         let source_epoch = projection.source_epoch.ok_or(
                             OpenMlsProjectionError::UnsupportedMessageKind(projection.kind),
--- a/crates/cgka-engine/src/openmls_projection.rs
+++ b/crates/cgka-engine/src/openmls_projection.rs
@@ -3051,6 +3051,34 @@ fn candidate_paths_with_pending_replay_messages(
                 let mut seen = BTreeSet::new();
                 let mut messages = Vec::new();
                 let mut final_epoch = None;
+                let mut first_commit_epoch = None;
+                for message in &path.messages {
+                    let projection = project_mls_message(&message.payload)?;
+                    if projection.kind == OpenMlsContentKind::Commit {
+                        first_commit_epoch = Some(projection.source_epoch.ok_or(
+                            OpenMlsProjectionError::UnsupportedMessageKind(projection.kind),
+                        )?);
+                        break;
+                    }
+                }
+                // A late application from common pre-fork history can be
+                // older than every commit in the candidate path. Probe it
+                // against the retained base state, before the first commit;
+                // that state owns the widest past-epoch decryption window.
+                // The observation is shared by every candidate and therefore
+                // cannot bias selection toward either fork.
+                if let Some(first_commit_epoch) = first_commit_epoch {
+                    for applications in applications_by_epoch
+                        .range(..first_commit_epoch)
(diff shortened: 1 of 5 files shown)


crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -3040,6 +3064,18 @@ fn candidate_paths_with_pending_replay_messages(
                                 }
                             }
                         }
+                        // Authenticate applications while the candidate still
+                        // owns their source-epoch state. Their branch-specific
+                        // observations are retained for witness scoring and the
+                        // eventual post-selection disposition; delivery itself
+                        // still happens only during canonical apply.
+                        if let Some(applications) = applications_by_epoch.get(&source_epoch) {
```
erskingardner, line 3112: This only probes applications whose source epoch matches a commit in the path (plus the tip). Applications sitting in `pending_messages` with a source epoch **older than the path's first commit** are now never offered to any candidate — pre-change, the trailing `for message in &applications` loop appended every pending application after the last commit and probed it against the materialized tip state.

That older-epoch case is reachable: while a sweep holds branch contexts, `ingest_group_message` routes *every* inbound application into convergence (`evidence_for_a_contested_pass`, ingest.rs:920-937), whatever its epoch — including a late-arriving app from a pre-fork epoch. Concretely: fork at epoch 5, a relay delivers an epoch-4 app late (still inside the app-retention window), it peels from the retained epoch-4 context and is buffered into the pass. No candidate-path commit has source epoch 4 and 4 ≠ `final_epoch`, so it is never probed; `decrypts_on_branches` comes back empty and `handle_app_message` terminally invalidates it as `UndecryptableInCanonicalState` (epoch ≤ selected tip) — `EpochInvalidated` + `AppMessageInvalidated` for a message that is still decryptable at the live state. Pre-change, the end-of-path probe authenticated it against the tip's retained shared-history secrets, it was accepted, and apply delivered it. That is the same permanent-loss class as #1171, re-opened for shared-history epochs.

Suggest keeping a fallback for uncovered epochs: probe applications with `source_epoch` below the first commit's source epoch against the base state *before* that first commit (the earliest replay state has the widest retention window), or retain the old end-of-path append for any application not yet `seen`. A regression test with a contested fork plus a late pre-fork application would pin the behavior.

erskingardner (commented):
**Request changes** (posted as comment — GitHub blocks request-changes on your own PR).

The core fix is right and well-scoped: bucketing replay by (source epoch, kind, input order) makes candidate materialization and canonical apply observe applications at the same point in the replay, which closes the pass-partition dependence from #1171, and the new six-advance one-pass vs split-pass regression test pins exactly that invariant. Witness eligibility correctly still gates on the candidate tip window, branch ids stay commit-digest-derived (peer-deterministic), and the docs were updated to match. CI is green.

One blocking concern, inline: the candidate-path change narrows *which* pending applications get probed at all. An application whose source epoch is older than every commit in the path — reachable when a late pre-fork app is buffered into convergence during a contested sweep (`evidence_for_a_contested_pass`) — is now never authenticated on any branch and is terminally invalidated without a decryption attempt, where pre-change the end-of-path probe delivered it. That re-opens the same permanent-loss class this PR fixes, for shared-history epochs. A small fallback probe for uncovered epochs (plus a contested-fork regression test) should close it.

Datawav (commented):
## Automated review

No actionable findings.

Datawav:
## Automated remediation result

- Verified the PR remains at audited head `133972cd25ae0eb483b63903e88d293295130fed`.
- The assigned audit contains no actionable findings, so no code or test change is warranted.
- Tests: not run (no remediation change and no finding-specific behavior to validate).
- Result: no commit created and nothing pushed.

crates/cgka-engine/src/openmls_projection.rs
erskingardner: Addressed in 8a3edd3fe3: applications older than the path's first commit are now probed against the retained base state before that commit (the widest-window position), the observation is shared across candidates so selection stays symmetric, and `contested_fork_delivers_late_pre_fork_application` pins the scenario. Verified the apply-side ordering stays consistent with the probe position. No further concern here.

erskingardner (commented):
**Recommend merge** (posted as comment — GitHub blocks formal approve/request-changes actions on your own PR).

Re-reviewed at 58d89020. The follow-up commit 8a3edd3fe3 resolves my earlier blocking finding: pending applications older than a candidate path's first commit are now prepended before that commit, so they are probed against the retained base state (the widest past-epoch decryption window) instead of never being offered to a candidate. Selection symmetry holds: the base state is shared by every candidate and pre-fork epochs are witness-ineligible (`epoch <= fork_epoch`), so the probe cannot bias branch selection. Apply-side ordering stays consistent with the probe position — accepted pre-fork apps sort first in the final replay and are authenticated against the same base state that observed them.

The new `contested_fork_delivers_late_pre_fork_application` test is a true regression test for the flagged scenario (an epoch-1 application held through an epoch-2 fork: accepted, `Processed`, delivered exactly once, no invalidation) and would fail without the probe. Contract and convergence docs updated to match the shipped behavior.

Validation: full CI test matrix green at 58d89020 (Rust tests 1–4, Simulator and vectors, CLI e2e, relay tests, Tamarin, format/clippy/check). Only the darwin-x86_64 build-only job was still pending at review time; darwin-aarch64 and every test suite pass.

Datawav (commented):
## Automated review

Follow-up found one new actionable issue.

- Major, high confidence — `crates/cgka-engine/src/openmls_projection.rs:3043-3045, 3070-3081, 3112-3118` sorts proposal buckets by message ID but consumes same-epoch application buckets in `pending_messages` insertion order across base, per-commit, and tip replay. Peers receiving identical applications in different orders can emit differently ordered observations and `accepted_app_messages`, undermining deterministic canonicalization; sort every `applications_by_epoch` bucket by message ID and add an opposite-arrival-order regression test.

Datawav:
Verified the application replay ordering finding against audited head `23d460dd3a924477993279e7e780fd9a0a3627e0`.

That head already sorts every `applications_by_epoch` bucket by message ID before base, per-commit, and tip replay. I added the missing regression in commit `e58d5c1afbb1ec31f62c98c75489dc116daff2c5`: two peers admit identical application messages in opposite orders across all three replay positions, then assert identical `accepted_app_messages` and emitted `MessageReceived` order.

Validation:
- `cargo test -p cgka-engine --features test-policy-overrides --test distributed_convergence application_replay_is_invariant_to_opposite_arrival_order` — passed
- `just fast-ci` — passed (format, naming and convergence gates, workspace checks, clippy for default and OTLP features, and convergence policy pin tests)
- `git diff --check` — passed

The branch and live PR head were re-fetched immediately before push and matched the audited SHA; after the normal push, both read back at `e58d5c1afbb1ec31f62c98c75489dc116daff2c5`.

crates/cgka-engine/tests/distributed_convergence.rs
```diff
@@ -2253,6 +2392,277 @@ async fn engine_materializes_multi_commit_path_from_stored_commits() {
...
+
+    let forward_result = forward
+        .converge_stored_openmls_messages_at(&group_id, 1_000_000)
+        .unwrap();
+    let reverse_result = reverse
+        .converge_stored_openmls_messages_at(&group_id, 1_000_000)
+        .unwrap();
+
+    assert_eq!(
+        forward_result.accepted_app_messages,
```
Datawav: This only compares the peers with each other, so it still passes if both silently omit every application or the same replay bucket. Please assert all six accepted message IDs and the exact six-payload canonical sequence, or at least verify that both messages from the base, edge, and tip epochs each appear exactly once.

Datawav (commented):
## Automated review

Follow-up incomplete: the requested arrival-order regression remains insufficiently verified.

- Minor — `crates/cgka-engine/tests/distributed_convergence.rs:2500-2516` compares the two peers’ `accepted_app_messages` and received payloads only with each other. Both assertions still pass if every application—or the same replay bucket—is silently omitted, so the test does not prove that both base-, edge-, and tip-epoch applications survive canonical replay. Assert all six accepted message IDs and the exact six-payload canonical sequence, or at minimum verify that both messages from every epoch appear exactly once.

Datawav:
Verified and remediated the follow-up audit finding at the pinned head `e58d5c1afbb1ec31f62c98c75489dc116daff2c5`.

The opposite-arrival-order regression now independently requires all six application message IDs and the exact six delivered payloads in canonical replay order across the retained base, both commit edges, and the selected tip. This closes the false-positive path where both peers could omit the same application or replay bucket and still compare equal.

Pushed commit `9a046075514c80dd8959f3625019c3d38429e5f2` to the existing `agent/process-apps-at-source-epoch` branch.

Verification:
- `cargo test -p cgka-engine --test distributed_convergence application_replay_is_invariant_to_opposite_arrival_order -- --exact` — passed (1 test)
- `just fast-ci` — passed, including formatting, workspace checks, Clippy for default and OTLP feature sets, and the convergence policy pin tests (5 tests)

erskingardner (commented):
**Recommend merge** (posted as comment — GitHub blocks formal approve on your own PR).

Re-reviewed at 9a046075 (delta from 58d89020: three commits). All changes are correct, well-scoped, and improve the fix:

- **Wire-epoch sort key in final apply.** `replay_messages_for_canonicalization_result` now derives the sort epoch and the applied-prefix proposal skip from `projection.source_epoch` (authenticated wire bytes) instead of mutable row metadata. A row can legally carry a newer stamp — the contested-pass handoff retains a past-epoch wire message in a row stamped with the receiver's current epoch — so the v2 code could misorder replay and reproduce the exact WrongEpoch/pruning failure this PR eliminates. The sort key now equals the epoch OpenMLS enforces, which makes misordering structurally impossible. The `UnsupportedMessageKind` error on a missing epoch is infallible for accepted proposal/commit/app sets and fails closed.
- **Deterministic within-bucket application order.** Sorting each `applications_by_epoch` bucket by message id mirrors the existing proposal sort and makes observations, accepted sets, and delivery order invariant to transport arrival order — verified end-to-end by the new opposite-arrival-order test, which also covers two same-sender apps per epoch bucket (secret-tree ordering within an epoch).
- **`projected_messages` refactor** removes the double projection introduced by the pre-fork probe; no behavior change.
- **Test hardening.** The partition test now derives commit count from the pinned `app_message_past_epoch_limit` instead of a magic 6, and the explicit restamp of the app row to the tip epoch is a true regression for the wire-epoch fix — it fails under record-epoch sorting and passes now.
- Docs name the regression tests inline (greppable across layers, per repo convention); no version bumps; 5 files, same scope.

Validation: full CI matrix green at 9a046075 — all builds, Rust tests 1–4, Simulator and vectors, CLI e2e, relay tests, Tamarin, format/clippy/check.

No remaining concerns from my side.

PR #1425, opened 2026-08-13
Add storage format measurement rail

erskingardner:
## Summary

- add real same-group rejoin and canonical-advance Criterion cases scaled across 0, 8, and 32 retained commits
- add application-message send and ingest Criterion anchors so future storage changes can enforce the tracker non-regression guard
- emit stable logical Welcome-row and founding-snapshot byte metrics from a feature-gated aggregate storage probe
- extend `just bench-startup` with database footprint reporting and a 10-group, 320-message fixture

This is intentionally stacked on #1421 so the storage implementation and its measurement/reporting rail remain independently reviewable. Review this PR as the single commit above `codex/storage-format-v2`.

## Post-v2 measurements

Release profile on an Apple Silicon development machine:

| Metric | Result |
| --- | ---: |
| `create_group/32 invitees` | 15.22 ms |
| normalized 32-member Welcome value | 16,821 bytes |
| largest 32-member founding snapshot | 82.6 KB |
| `join_welcome/1 stored KP` | 1.30 ms |
| `join_welcome/32 stored KPs` | 1.33 ms |
| `join_welcome_large_group/32 members` | 3.94 ms |
| app-message send | 1.11 ms |
| app-message ingest | 1.06 ms |
| rejoin after 32 retained commits | 2.19 ms |
| canonical advance after 32 retained commits | 3.12 ms |

The tracker anchors were 180,002 bytes per Welcome row, roughly 6 MB for the founding snapshot, and roughly 65 ms for `create_group/32`. The post-v2 values are below all three stated targets. The new send/ingest cases establish stable same-harness baselines for subsequent changes; the repository did not previously contain those benches.

The focused release startup fixture reported:

```text
MDK_BENCH startup_scaling case=messages_32 groups=10 roster=1 messages_per_group=32 database_bytes=3039232 database_footprint_bytes=3039232 chat_list_ms=199 ready_ms=108 group_read_ms=4 account_open_wait_ms=107 session_open_ms=50 group_hydration_ms=0 profile_load_ms=0 read_snapshot_ms=0
```

## Validation

- `just fast-ci`
- `cargo test -p storage-sqlite --features storage-format-benchmarks` (367 passed, 2 subprocess entry points ignored)
- `cargo test -p marmot-app --test startup_scaling` (smoke passed, 7 benchmark cases ignored)
- focused release Criterion runs for all metrics above
- focused release startup message-heavy fixture

A featureless `cargo test -p cgka-engine` is independently red because `deferred_peel_lifecycle` deliberately installs non-default policies but does not enable `test-policy-overrides`; the feature-correct engine suite is green.

Tracks #1413
Related: #1415

### DIFF
--- a/crates/storage-sqlite/src/storage/messages.rs
+++ b/crates/storage-sqlite/src/storage/messages.rs
@@ -34,7 +34,55 @@ pub struct MessageFormatPromotionProgress {
     pub has_more: bool,
 }
 
+/// Logical blob/value sizes used by the storage-format benchmark rail.
+///
+/// This excludes SQLite page overhead; the startup-scaling harness reports
+/// the complete database-file footprint separately.
+#[cfg(feature = "storage-format-benchmarks")]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct StorageFormatBenchSizes {
+    pub message_value_bytes: u64,
+    pub largest_snapshot_bytes: u64,
+}
+
 impl SqliteAccountStorage {
+    /// Read aggregate value sizes without returning any stored content.
+    #[cfg(feature = "storage-format-benchmarks")]
+    pub fn storage_format_bench_sizes(
+        &self,
+        group_id: &GroupId,
+        message_id: &MessageId,
+    ) -> StorageResult<StorageFormatBenchSizes> {
+        let conn = self.lock()?;
+        let message_value_bytes = conn
+            .query_row(
+                "SELECT length(id) + length(group_id) + coalesce(length(payload), 0) +
+                        coalesce(length(record), 0) + coalesce(length(deferred_peel), 0)
+                 FROM cgka_messages WHERE id = ?1",
+                params![message_id.as_slice()],
+                |row| row.get::<_, i64>(0),
+            )
+            .optional()
+            .storage()?
+            .ok_or(StorageError::NotFound)?;
+        let largest_snapshot_bytes = conn
+            .query_row(
+                "SELECT max(length(snapshot)) FROM cgka_group_snapshots WHERE group_id = ?1",
+                params![group_id.as_slice()],
+                |row| row.get::<_, Option<i64>>(0),
+            )
+            .storage()?
+            .ok_or_else(|| StorageError::SnapshotMissing("benchmark-size-anchor".into()))?;
+        Ok(StorageFormatBenchSizes {
+            message_value_bytes: u64::try_from(message_value_bytes).map_err(|error| {
+                StorageError::Backend(format!("invalid message benchmark size: {error}"))
+            })?,
+            largest_snapshot_bytes: u64::try_from(largest_snapshot_bytes).map_err(|error| {
+                StorageError::Backend(format!("invalid snapshot benchmark size: {error}"))
+            })?,
+        })
+    }
+
     /// Promote at most `limit` legacy message rows to normalized format 2.
     ///
     /// The bounded batch is atomic and idempotent. A malformed row rolls the
@@ -728,6 +776,38 @@ mod tests {
             store.get_message(&message.id).unwrap().state,
             MessageState::PeelDeferred
         );
+    }
+
+    #[cfg(feature = "storage-format-benchmarks")]
+    #[test]
+    fn storage_format_bench_sizes_accepts_legacy_message_rows() {
+        let store = SqliteAccountStorage::in_memory().unwrap();
+        let group = sample_group(gid(1), 0, 0);
+        let message = sample_message(mid(1), group.id.clone(), 0);
+        let record = serialize(&message).unwrap();
+        store.put_group(&group).unwrap();
+        store
+            .lock()
+            .unwrap()
+            .execute(
+                "INSERT INTO cgka_messages (id, group_id, epoch, state, record)
+                 VALUES (?1, ?2, 0, 0, ?3)",
+                rusqlite::params![message.id.as_slice(), group.id.as_slice(), record],
+            )
+            .unwrap();
+        store
+            .create_group_snapshot(&group.id, "legacy-size")
+            .unwrap();
+
+        let sizes = store
+            .storage_format_bench_sizes(&group.id, &message.id)
+            .expect("legacy row size is defined");
+        assert_eq!(
+            sizes.message_value_bytes,
+            u64::try_from(message.id.as_slice().len() + group.id.as_slice().len() + record.len())
+                .unwrap()
+        );
+        assert!(sizes.largest_snapshot_bytes > 0);
     }
 
     #[test]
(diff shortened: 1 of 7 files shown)


Datawav (commented):
## Automated review

No actionable findings.

Datawav:
Remediation verification for audited head `a29538d525b8c9b8c3556d2ec0628d304bdbc8fb`: the assigned automated audit reported no actionable findings, and verification found no claim requiring a code change. No files were changed, no remediation tests were run, and no commit or push was created. Existing GitHub CI for this exact head is green.

crates/storage-sqlite/src/storage/messages.rs
```diff
@@ -34,7 +34,55 @@ pub struct MessageFormatPromotionProgress {
...
+    pub fn storage_format_bench_sizes(
+        &self,
+        group_id: &GroupId,
+        message_id: &MessageId,
+    ) -> StorageResult<StorageFormatBenchSizes> {
+        let conn = self.lock()?;
+        let message_value_bytes = conn
+            .query_row(
+                "SELECT length(id) + length(group_id) + length(payload) +
+                        coalesce(length(record), 0) + coalesce(length(deferred_peel), 0)
```
erskingardner: Non-blocking nit: `length(payload)` is the only un-coalesced term, so pointing this probe at a legacy format-1 row (payload NULL — the inverse of the format-2 invariant where `record` is NULL) makes the whole sum NULL and surfaces as an opaque rusqlite column-type error rather than a size. The current bench only ever passes a freshly-written format-2 row, so this is fine today; either `coalesce(length(payload), 0)` for symmetry or a one-line doc note that the probe assumes normalized rows would keep it from surprising the next storage-change comparison.

erskingardner (commented):
Reviewed as the single commit `a29538d` on top of `codex/storage-format-v2` (#1421), per the stacked-PR instructions. (Posted as a comment because GitHub will not let the PR author self-approve; content-wise this is an approval.)

**Verdict: recommend merge.** This is a measurement-only rail with no production-path change: the storage probe is additive and feature-gated (`storage-format-benchmarks`, enabled only via cgka-engine dev-dependencies), and everything else is bench/test/Justfile surface. CI is green at this exact head.

Verified against the linked context:
- Covers the core scope of #1413 items 1–3: same-group rejoin scaled by retained history (0/8/32), canonical-advance commit+confirm scaled by retained history, and DB-footprint reporting on the startup harness (`database_bytes`/`database_footprint_bytes`, new `messages_32` case). The optional CI wiring (item 4) is explicitly out of scope here.
- Fixture correctness checks out: `list_messages(group_id, EpochId(0))` reads as "at or after epoch 0", so the retained-history assertions are sound; `converge_stored_openmls_messages_at` is the engine's doc-hidden convergence entry point, appropriate for forcing canonical settlement in the fixture; per-iteration `iter_batched` setup keeps group lifecycle costs out of the timed intervals; deterministic app payload per fresh engine/group avoids any dedup interaction.
- The startup footprint is sampled after the fixture block drops its app/client and before the cold reopen, with WAL/SHM sidecars summed in — consistent with the "after fixture connections close" doc.
- The PR body's disclosure about featureless `cargo test -p cgka-engine` being red (`deferred_peel_lifecycle` needing `test-policy-overrides`) is accurate and pre-existing on the base branch — this PR does not touch that gating, and workspace CI supplies the feature transitively via `{{test-features}}`.
- The stale "~180KB of JSON" comment removal in `bench_join_welcome_large_group` is the right kind of surgical cleanup post-v2.

Two non-blocking notes:
1. Inline: the probe's un-coalesced `length(payload)` would error opaquely on a legacy format-1 row. Fine for the current caller; worth a coalesce or a doc note.
2. For #1413 cohort sign-off (not this PR): the reported `join_welcome` numbers (1.30/1.33 ms) sit nominally just above the tracker's >2% no-regression guard vs the 1.27/1.28 ms pre-format baseline. Almost certainly noise at 10-sample, sub-millisecond benches, but worth one confirming run before the tracker closes.

crates/storage-sqlite/src/storage/messages.rs
erskingardner: Fixed in 2e451942. The probe now uses `coalesce(length(payload), 0)` so both normalized and legacy rows have a defined aggregate size. Added a feature-gated regression test with an actual format-1 row; it passes, the benchmark target compiles, and `just fast-ci` is green.

erskingardner:
Follow-up on the non-blocking >2% join guard note: reran the three release Criterion join anchors at head `2e451942` after the review fix.

- 1 stored KP: midpoint +3.76%, CI -4.46% to +12.15%, `p = 0.40` — no change detected
- 32 stored KPs: midpoint +0.75%, CI -5.31% to +6.83%, `p = 0.83` — no change detected
- 32-member join: midpoint +2.26%, CI -0.58% to +5.03%, `p = 0.16` — no change detected

The 10-sample configuration is too underpowered to certify a strict 2% threshold for ~1 ms operations; these results do not show a statistically significant regression. For final #1413 cohort sign-off, use a longer sampling configuration or repeated same-machine A/B run rather than treating the midpoint alone as the gate.

erskingardner (commented):
Re-reviewed at new head `2e451942` (amended single commit, rebased onto #1421's review-feedback commit `b08f53fc`; PR diff is still the same 7 files, CI fully green at this exact head).

**Verdict: recommend merge.** Both review notes are addressed:

1. **Probe NULL-handling (inline nit): fixed.** `length(payload)` is now `coalesce(length(payload), 0)`, and the new feature-gated regression test `storage_format_bench_sizes_accepts_legacy_message_rows` inserts a genuine legacy-shaped row — it satisfies the migration-47 CHECK constraint's format-1 branch (`storage_format = 1 AND record IS NOT NULL AND payload IS NULL`, relying on the column's `DEFAULT 1`) and pins the exact expected byte sum (id + group_id + record). This also locks in the right cross-format semantics: a legacy Welcome row reports its ~180 KB through `length(record)`, so pre/post-v2 anchors stay comparable. Test helpers are the crate's existing `test_support`, and the case is exercised both by the author's `cargo test -p storage-sqlite --features storage-format-benchmarks` run (367 passed) and by workspace CI via feature unification with cgka-engine's dev-dependency.

2. **Join non-regression guard: resolved with appropriate rigor.** The rerun of all three join anchors shows no statistically significant change (p = 0.40 / 0.83 / 0.16), and the follow-up is candid that the 10-sample config is underpowered to certify a strict 2% gate on ~1 ms operations — correctly deferring a longer same-machine A/B run to #1413 cohort sign-off rather than overclaiming. That residual action belongs to the tracker, not this PR.

No new findings. The rail remains measurement-only, additive, and faithful to the #1413 scope.

Datawav (approved):

PR #1426, opened 2026-08-13
Feat/per intent terminal disposition

jgmontoya:
Completes #1177's unmet criterion. (#1177 itself was closed via #1246; this ships the
acceptance-inference half that #1246 left open.)

### Why

An accepted app message had no honest terminal state. Two holes produced the same
symptom from opposite directions:

- An accepted-but-unpublished send returned `SendSummary { published: 0, message_ids: [] }`,
  so a host could only *infer* acceptance from an empty list — the inference #1177's
  criterion forbids. `effects.queued` carried the answer and had no reader anywhere in the
  workspace.
- When a group reached a terminal state its whole outbound queue was purged, but the rows
  those intents belonged to kept deriving as pending. Forever. That was the one place the
  app could not tell "still coming" from "never arriving".

### What ships

**Per-message disposition.** `SendAcceptDisposition` lands beside `SendResult`, projecting
its `Queued`/published split, carried on `SendSummary` and across UniFFI. The
application-message path reports **per-message** from `published_app_messages` rather than
the group-wide queue: one send pass can fold peer commits and publish unrelated work, so a
pass-level count would misreport the row. Group-state sends share one rule, since the engine
queues such an intent whole or not at all.

**Terminal sweep.** Disband and losing the local copy both purge the outbound queue
wholesale. A group-scoped 1:N invalidation predicate now reaches those rows at the same sync
seam that already fixes up membership, routes, and push tokens. No existing primitive could:
one matches on source message id (NULL for a retained intent) and the other on origin commit
id (never set for sent app events). Excluding already-invalidated rows keeps a repeated sweep
inert, so the ingest loop's batch retry cannot rewrite a recorded reason.

**Typed halt refusal.** `EngineError::GroupUnrecoverableRepairRequired` replaces
`InvalidTransition` at both durable halt-sync gates — `validate_send_acceptance` and the
queued-drain recheck in `send.rs` — so a drain cannot report the same condition two ways. It
mirrors `QueuedOutboundAtCapacity` including the deliberate non-`is_transient` treatment: the
halt clears on a remote member's repair Welcome, not on a timer this call could pick.

**Delivery-change trigger.** `DeliveryOrSendStateChanged` was defined, mirrored across FFI,
and already mapped to the chat list — but nothing constructed it, so both were dead and a
pending → delivered flip reached subscribers indistinguishable from new content. It is now
produced at finalization (the seam where a send's source id goes NULL → Some) and only for
the published row itself.

**Retry forwarding.** `retry_group_convergence` is the path that releases sends the engine
retained, so its finalize updates carry exactly the flip a held row waits for — and it dropped
them. They are now buffered for the account worker. The send path still discards the same
updates on purpose (its re-record re-emits the flip); both sites now name their case so
neither gets "fixed" into a double emit.

### Design overrides worth knowing

- **Per-message disposition, not `effects.queued`.** The queue is group-wide; the answer a
  host needs is per row.
- **Both halt gates typed `GroupUnrecoverableRepairRequired`.** The retention-boundary
  epoch-state check deliberately *stays* `InvalidTransition` — reaching it means epoch state
  diverged from the durable markers, which is an engine bug, and `InvalidTransition` remains
  the untyped engine-bug signal. That split is what still pins gate ordering in
  `unrecoverable_halt_survives_engine_restart_until_verified_repair`, now by type rather than
  by a reason string.
- **Swept rows stay visible as Failed**, sharing the terminal reason with the send-time
  retraction path so both render identically.

### Host-visible change

A publish now surfaces as `LatestMessageDeliveryChanged` where it previously surfaced as
`NewLastMessage`. Clients that treated the chat-list trigger as "new content" will need to
handle the delivery-change trigger to keep badges updating.

## Summary by CodeRabbit

* **New Features**
  * Send results now distinguish messages published immediately from those accepted for later delivery.
  * Added typed errors identifying groups that require repair before sending can continue.
  * Exposed send acceptance status through supported client interfaces.
* **Bug Fixes**
  * Pending sends are now marked failed when a group is disbanded or the local account is removed.
  * Improved delivery-state updates so published messages are reported accurately.
* **Documentation**
  * Updated documentation for send acceptance, repair-required groups, and terminal delivery outcomes.


Datawav:
Fixed the deterministic `Rust tests (3/4)` failure at the current PR head.

Root cause: `sweeping_a_terminal_group_stops_a_held_send_from_claiming_pending` expected the failed local send to become the chat-list preview, contradicting the established projection invariant that failed local sends remain in the timeline but do not outrank accepted history. The sweep itself correctly marks the held timeline row `local_publish_failed`; the earlier published message correctly remains the chat-list preview as `Delivered`.

Pushed commit [`fc55a0f9`](https://github.com/marmot-protocol/mdk/commit/fc55a0f99c321b6844bb5076450a41aa7a3853f6).

Local verification:
- `cargo test -q -p marmot-app sweeping_a_terminal_group_stops_a_held_send_from_claiming_pending -- --nocapture` — passed
- `cargo test -q -p storage-sqlite sweeping_a_terminal_group_invalidates_only_its_own_unpublished_sends -- --nocapture` — passed
- `cargo fmt --all -- --check` — passed
- `git diff --check` — passed

CI has been retriggered for the new head `fc55a0f99c321b6844bb5076450a41aa7a3853f6`.

Datawav (commented):
## Automated review

Actionable findings found.

- Medium, high confidence — `crates/marmot-app/src/client/mod.rs:2203-2208` discards every update returned by `finalize_published_app_message_source_retention`, although the subsequent re-record only reprojects the current `app_event_id`; when the same pass publishes older retained messages, their stored state becomes delivered but observers receive no pending→delivered notification. Forward all finalization updates except the current message’s update, and add a regression test where one pass publishes both current and previously retained messages with exactly one delivery-state notification each.

- Medium, moderate confidence — `crates/marmot-app/src/client/sync.rs:1907-1930` performs terminal pending-send invalidation only while handling inbound outcomes, while terminal `GroupStateChanged` events can also surface through the engine’s independent drain buffer. Apply the same invalidation sweep to drained terminal events, with a test that queues a terminal transition without an inbound delivery and verifies retained rows become failed rather than remaining pending.

crates/marmot-app/src/client/mod.rs
```diff
@@ -2199,6 +2199,10 @@ impl AppClient {
             self.record_human_action_succeeded(group_id, context, &effects);
         }
         self.remember_published_reports(&effects);
+        // Discarded deliberately, unlike on the convergence-retry path: the
+        // re-record below reprojects the same row with its new source id and
+        // hands that update to `on_local_projection`, so forwarding these too
+        // would emit the flip twice.
         let _finalize_updates = self.finalize_published_app_message_source_retention(&effects)?;
```
Datawav, line 2206: This discards finalization updates for every published app message, but the re-record below only reprojects the current `app_event_id`. If this pass also publishes older retained messages, their stored state becomes delivered without notifying observers. Forward all finalization updates except the current message’s update, and test a pass that publishes both current and retained messages with exactly one delivery-state notification each.

crates/marmot-app/src/client/sync.rs
```diff
@@ -1907,6 +1907,27 @@ impl AppClient {
...
+            // The engine purges the whole outbound queue at both seams above, so
+            // every send it still held is dead; without this sweep those rows
+            // derive as `pending` forever, which is the one place the app cannot
+            // tell "still coming" from "never arriving". Propagate the error
+            // rather than swallow it: a silently skipped sweep leaves exactly
+            // the lie this fixes. The sweep ignores already-invalidated rows, so
+            // the batch retry that error triggers is a no-op for anything it
+            // already withdrew.
+            if let cgka_traits::engine::GroupEvent::GroupStateChanged {
+                group_id, change, ..
```
Datawav: This terminal invalidation runs only while handling inbound outcomes, but terminal `GroupStateChanged` events can also arrive through the engine’s independent drain buffer. Apply the same invalidation sweep to drained terminal events; otherwise retained sends can remain pending after their outbound queue is purged. Add a test that drains a terminal transition without an inbound delivery and verifies those rows become failed.

Datawav (approved):

crates/marmot-app/src/client/sync.rs
```diff
@@ -445,6 +455,7 @@ impl AppClient {
                 updated_group.as_ref(),
                 &source_message_id_hex,
             );
+            self.invalidate_terminal_pending_sends(event, &local_account_id_hex, &mut summary)?;
```
Datawav, line 458: Terminal invalidation happens only while iterating `effects.events`, but `observe_drained_session_events` calls `fail_if_publish_failed(effects)?` before that loop. If one drained batch contains both an unrelated publication failure and `GroupDisbanded` or local-member removal, the early return consumes the terminal event without executing this sweep, leaving accepted-but-unpublished rows falsely `pending` until a later reopen/hydration. Process these idempotent terminal invalidations before returning the publication error, and add mixed-batch tests for both disband and local removal.

Datawav (commented):
## Nightly deep review

Actionable finding found.

- Medium, high confidence — `crates/marmot-app/src/client/sync.rs:371-458`: a mixed drained batch can permanently bypass the new terminal pending-send reconciliation. `observe_drained_session_events` calls `fail_if_publish_failed(effects)?` before iterating `effects.events`, while `invalidate_terminal_pending_sends(...)` runs only inside that later loop. If the same one-shot `runtime.drain()` returns an unrelated publication failure together with `GroupDisbanded` or local-member-removal, the early return consumes the terminal event without invalidating the group’s accepted-but-unpublished timeline rows. Recovery arming does not repair their disposition, and a retained client cannot observe the drained event again; those rows can therefore remain falsely `pending` until the client is reopened and hydration happens to replay the tombstone. This violates the PR’s core durable-state requirement precisely on the crash/failure seam it intends to close.

  Need: terminal queue destruction must be projected even when another publication in the same effects batch fails. Publication failure and terminal disposition are independent outcomes; one must not suppress the other.

  Recommended design: process idempotent terminal invalidations before `fail_if_publish_failed`, retain their projection updates in durable/deferred summary state, then return the publication error. Ensure any invalidation error itself remains retryable without losing the drained event.

  Alternatives: durably retain and replay the complete drained effects batch until all projections succeed; or derive terminal invalidations directly from durable engine tombstones on every reconciliation pass rather than relying on one-shot events.

  Test gap: extend the new hydration/drain regression with one `AccountDeviceEffects` containing both `publish_failed` and `GroupDisbanded` (and separately local removal), assert that the call still reports the publication failure, and verify the held send is already `local_publish_failed` without reopening the client.

PR #1428, opened 2026-08-14
Refactor/incident replay outcome display single source

jgmontoya:
Delivers the three reply-promises from #1245 plus the refactor they were blocking.

### The refactor

`Display for Outcome` could only ever render half its variants. The CLI sends `Accepted` to
`persist_or_report` (which prints the artifact's fidelity, sensitivity, and written paths —
facts the library does not have) and prints `InfrastructureFailure` to stderr itself, so those
two arms were unreachable. The `accepted:` arm was also **wrong about the CLI it appeared to
describe**: no invocation has ever printed a bare `accepted: <scenario-name>` line.

The impl is dropped and the CLI's match gets an arm per variant, so one exhaustive match is
the single source of truth for the operator-facing line. `Advisory` keeps its `Display` — an
advisory is only its message, so it renders losslessly without CLI knowledge.

**Output is unchanged, proven:** both binaries diffed over all 37 fixtures (10 healthy, 24
quarantine, 3 accepted, 8 advisory lines) plus the accepted-with-out-dir form, byte for byte.

**The baseline came first.** The operator-facing output had no test at all — every existing
test asserts on the in-process `Routing`, so the strings the CLI actually prints were unpinned
and free to drift from the rendering that claims to describe them. `tests/cli.rs` now pins each
primary line against the real binary, including that an accepted run prints fidelity and
sensitivity rather than a bare scenario name. That assertion is the baseline the refactor
commit runs against.

### #1245 promise 1 — `HALT_NON_CAUSE_REASONS`

Both `unrecoverable_halt_reason` arms substitute `unspecified` for a halt row naming neither a
reason nor an `error_kind`, but the demotion tier listed only the two re-assertions. So a
single reason-less row outranked them and reported `engine-a (unspecified)` — suppressing the
strings that at least name the surface the halt was re-asserted on. The stand-in for "no reason
at all" beat the weakest thing it could beat.

`unspecified` now joins the same fallback tier, and the tier takes `UNSPECIFIED_HALT_REASON`
**by reference** so the tier and the substituted string cannot drift apart — the drift that
caused this. New fixture pins **both tier directions** in one classify, one engine each, with a
stand-in from each of the two arms.

> **Judgment call, flagging deliberately:** `HALT_RE_ASSERTION_REASONS` →
> `HALT_NON_CAUSE_REASONS`, and `is_halt_re_assertion` → `is_halt_cause` (stated positively, so
> the one call site reads as its own definition: "causes are the reasons that name a cause").
> The rename is judgment, not necessity — re-assertion simply no longer describes every member
> of the tier. Happy to revert the names if reviewers prefer the old ones.

### #1245 promise 2 — rule-5 re-flow

Two prior edits left rule 5 mis-wrapped: a 14-character orphan line ("folded back") and a
114-character line where a sentence was appended without re-wrapping. Re-flowed to the file's
~80-column convention. **The word sequence is unchanged, verified by word-diff.**

### #1245 promise 3 — one home per argument

Two arguments had been restated at every dependent site: the untimed-halt one in four places
(AGENTS.md rule 5, `HaltLifecycle::halt_position_ms`, `still_halted`, the mixed-instrumentation
test) and the superseded-artifact one in three (the `contested_convergence_replay` const, the
`Ok` arm of `fall_through_to_fork`, AGENTS.md's route bullet). Near-identical prose in n places
is n places to update and n−1 chances to leave a stale copy.

Each now has one canonical home — `halt_position_ms`'s doc and `fall_through_to_fork`'s doc —
and the other sites keep only what is theirs and point at it. Two facts that existed nowhere
else were merged into the canonical homes first: "on partially instrumented input" and that a
trace-free envelope fails the standard `archetype_artifact` enforces on itself. AGENTS.md keeps
stating both *rules* — the contract is what it is for; only the justification moved.

Also deleted the `Err`-arm comment explaining that no artifact is recorded on a path that has
no artifact: the arm cannot record one, so the reader never asks.

## Summary by CodeRabbit

- **New Features**
  - Improved incident replay reporting for healthy, quarantined, and accepted outcomes.
  - Quarantine results now distinguish concrete halt causes from re-assertions and unspecified reasons.

- **Bug Fixes**
  - Prevented halt clearing when evidence is untimed or ambiguously ordered.
  - Ensured concrete halt causes take precedence over non-causal reasons.
  - Improved accepted and infrastructure-failure output handling.

- **Documentation**
  - Clarified halt lifecycle, fall-through behavior, repair requirements, and command-line output formats.

### DIFF
--- a/crates/incident-replay/src/main.rs
+++ b/crates/incident-replay/src/main.rs
@@ -76,11 +76,20 @@ fn main() -> ExitCode {
     // attested history supersedes an archetype, and what happens when a route
     // fails closed — lives in the library; this is presentation only.
     let routing = route(&export, source_format);
+    // Every primary line is rendered here, so this match is the single source of
+    // truth for what an operator reads. Two of the four need facts only this
+    // binary has — the accepted line depends on what was written where, and an
+    // infrastructure failure is a stderr line, not a verdict — so `Outcome`
+    // deliberately renders none of them itself.
     let code = match &routing.outcome {
-        // Producing no vector is a valid outcome, so a quarantine is not an
+        // Producing no vector is a valid outcome, so neither of these is an
         // error exit.
-        Outcome::Healthy | Outcome::Quarantine { .. } => {
-            println!("{}", routing.outcome);
+        Outcome::Healthy => {
+            println!("healthy: 0 vectors");
+            ExitCode::SUCCESS
+        }
+        Outcome::Quarantine { reason } => {
+            println!("quarantine: {reason}");
             ExitCode::SUCCESS
         }
         Outcome::Accepted(artifact) => persist_or_report(artifact, out_dir.as_deref()),
--- a/crates/incident-replay/src/main.rs
+++ b/crates/incident-replay/src/main.rs
@@ -9,11 +9,12 @@
 //! Reading, format detection, and printing live here; everything about *which*
 //! route an export takes is [`incident_replay::route`].
 //!
-//! Output is one primary line — `healthy:`, `quarantine:`, or `accepted:` — plus
-//! an `advisory (<label>):` line for every co-occurring finding that line does
-//! not itself report. An accepted incident is written owner-only as the portable
-//! vector next to its `incident-scenario-artifact.v1` evidence envelope; the
-//! source export, transport ciphertext, and MLS checkpoints are never copied.
+//! Output is one primary line — `healthy:`, `quarantine:`, or
+//! `accepted (<fidelity>, <sensitivity>)` — plus an `advisory (<label>):` line
+//! for every co-occurring finding that line does not itself report. An
+//! accepted incident is written owner-only as the portable vector next to its
+//! `incident-scenario-artifact.v1` evidence envelope; the source export,
+//! transport ciphertext, and MLS checkpoints are never copied.
 //!
 //! Exits 0 for any successful classification (healthy, quarantine, and accepted
 //! are all valid outcomes). Exits 2 on usage, I/O, parse, or write failure, and
--- a/crates/incident-replay/tests/route.rs
+++ b/crates/incident-replay/tests/route.rs
@@ -51,8 +51,9 @@ fn a_fail_closed_convergence_falls_through_to_a_clean_fork() {
 
 #[test]
 fn a_superseded_route_is_recorded_in_the_accepted_artifact() {
-    // The test above pins the stdout surface of the superseded finding; this pins
-    // the persisted one. `unavailable_fields` is where the artifact already states
+    // The test above pins the in-memory advisory (`routing.advisories`); this
+    // pins the persisted one — the stdout rendering of both lives in
+    // `tests/cli.rs`. `unavailable_fields` is where the artifact already states
     // what it could not cover, so that is where the finding belongs — see
     // `fall_through_to_fork` for why the artifact needs its own copy at all.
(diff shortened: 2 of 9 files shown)


crates/incident-replay/src/main.rs
```diff
@@ -76,11 +76,20 @@ fn main() -> ExitCode {
     // attested history supersedes an archetype, and what happens when a route
     // fails closed — lives in the library; this is presentation only.
     let routing = route(&export, source_format);
+    // Every primary line is rendered here, so this match is the single source of
+    // truth for what an operator reads. Two of the four need facts only this
+    // binary has — the accepted line depends on what was written where, and an
+    // infrastructure failure is a stderr line, not a verdict — so `Outcome`
+    // deliberately renders none of them itself.
```
erskingardner, line 84: Could you update the module-level output contract at main.rs:12 in this PR as well? It still advertises `accepted:` as the emitted prefix, while `persist_or_report` emits `accepted (` and the new test explicitly says the bare `accepted:` line is never produced. Leaving that stale repeats the same misleading contract this refactor removes from `Display`.

erskingardner (dismissed):
Verified at exact head 98d939505ff3956ec0e3d7ff6c7f78f62873c194. The halt-reason fix puts `unspecified` in the intended non-cause fallback tier and covers both reporting directions; the CLI refactor keeps the operator-facing behavior centralized; the rationale consolidation preserves the crate contract. Current CI is fully green. One non-blocking documentation-consistency note inline.

crates/incident-replay/src/main.rs
jgmontoya: Agreed, and thanks — that's the same unreachable-rendering defect the PR removes from `Display`, just one level up in the doc that operators read first. Fixed in 11f83708: the contract now reads `accepted (<fidelity>, <sensitivity>)`, which covers both forms `persist_or_report` emits (`main.rs:139` with written paths, `main.rs:154` without an out-dir) and matches the negative assertion at `tests/cli.rs:80-83`. I checked the rest of the crate for other restatements of the prefix — the only remaining `accepted:` occurrences are the intentional negative assertions in `tests/cli.rs`.

crates/incident-replay/tests/route.rs
```diff
@@ -51,14 +51,10 @@ fn a_fail_closed_convergence_falls_through_to_a_clean_fork() {
...
-    // envelope sees a clean accepted fork archetype with no trace that the same
-    // export carried a higher-precedence contested convergence at the same epoch
-    // that could not be replayed — which is exactly the standard `archetype`
-    // enforces on itself ("an artifact that cannot point at its own source rows is
-    // not publishable evidence"). `unavailable_fields` is where the artifact
-    // already states what it could not cover, so the finding belongs there.
+    // The test above pins the stdout surface of the superseded finding; this pins
+    // the persisted one. `unavailable_fields` is where the artifact already states
+    // what it could not cover, so that is where the finding belongs — see
+    // `fall_through_to_fork` for why the artifact needs its own copy at all.
```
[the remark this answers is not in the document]
jgmontoya: Right, that comment described its sibling by the surface it feeds rather than the one it asserts. fall_through_to_fork's companion reads routing.advisories in memory and never runs the binary; the stdout rendering of both findings is pinned in tests/cli.rs. Reworded to say what each test pins.

PR #1433, opened 2026-08-14
Operationalize storage format v2 migration

### LINKED ISSUE #1414: ops: storage-format v2 migration runbook, upgrade/downgrade policy, and compatibility test matrix

erskingardner:
## Summary

- schedule bounded legacy-message promotion after account readiness instead of extending session-open latency
- add an exact SQLCipher database fixture written by MDK v0.9.12, retain its writer source and digest, and verify upgrade/promotion semantics plus downgrade refusal
- make unsupported future schema versions a typed storage error
- add a file-backed operational benchmark for migration, promotion, reopen, peak DB/WAL/SHM footprint, and explicit keyed `VACUUM`
- document the compatibility matrix, measured free-space envelope, release sequencing, backup remedy, and opt-in `VACUUM` policy

## Runtime policy

`marmot-app` runs at most one 32-row promotion transaction per 15-second steady-state maintenance tick. It starts only after account readiness, logs aggregate counts and duration, retries transient contention, and halts durable decode failures for the current process while legacy reads remain available. Existing crash-kill coverage proves an interrupted batch rolls back and converges on retry.

## Operational evidence

Release-profile measurements on an Apple M4 Max:

| Legacy rows | Starting footprint | Migration | Peak footprint | Promotion work | Max batch | Reopen | VACUUM | Final footprint |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 512 | 94.1 MB | 1.69 s | 376.1 MB (4.00x) | 1.62 s / 16 batches | 136 ms | 90 ms | 411 ms | 32.4 MB |
| 2,048 | 376.0 MB | 8.03 s | 1.508 GB (4.01x) | 8.28 s / 64 batches | 342 ms | 97 ms | 1.53 s | 128.0 MB |

The runbook therefore requires free space equal to at least 3.25 times the existing account database size in addition to the existing database, plus a pre-upgrade backup/export. Timings remain machine characterization and are remeasured before the cohort release with:

```sh
MDK_STORAGE_OPS_ROWS=2048 just bench-storage-upgrade
```

## Verification

- `just fast-ci`
- `cargo test -p cgka-traits`
- `cargo test -p storage-sqlite` (374 passed; 3 intentional ignores)
- `cargo test -p cgka-session --features test-policy-overrides` (20 passed)
- `cargo test -p cgka-session --test session_lifecycle session_facade_promotes_one_bounded_legacy_row_without_semantic_change -- --exact`
- `cargo test -p marmot-app --lib --features test-policy-overrides` (648 passed)
- `cargo test -p cgka-conformance-simulator --test tracing_audit production_library_sources_do_not_write_direct_output -- --exact`
- `just bench-storage-upgrade`
- `MDK_STORAGE_OPS_ROWS=2048 just bench-storage-upgrade`

Closes #1414
Related to #1415

## Summary by CodeRabbit

* **New Features**
  * Legacy message data is now upgraded incrementally in the background after account readiness, with bounded batches and progress tracking.
  * Storage upgrades can be benchmarked with configurable workloads.
* **Bug Fixes**
  * Databases created by newer app versions are now rejected safely with a clear, privacy-conscious error.
  * Upgrade retries handle temporary contention, while durable failures stop further processing without affecting legacy reads.
* **Documentation**
  * Expanded storage compatibility, migration, rollback, disk-space, and upgrade-duration guidance.


crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1120,6 +1121,10 @@ async fn run_app_runtime_account_worker(
                 if lifecycle.is_stopping() {
                     continue 'worker;
                 }
+                run_legacy_message_promotion_batch(
```
erskingardner, line 1124: (Non-blocking) Worth a conscious ack: this is a synchronous ~136–342 ms (measured worst continuous batch, M4 Max) blocking transaction inside the single worker's `select!` loop, so it delays every other arm — commands, shutdown — for its duration. That's consistent with the runtime's sync-engine-step design and the whole-session mutex means `spawn_blocking` would buy almost nothing, so the 32-row bound is the right mitigation. On slower mobile flash a batch could plausibly stretch toward ~1 s during the promotion window; if that ever shows up in stage telemetry, the knob is already in the right place (`LEGACY_MESSAGE_PROMOTION_BATCH_SIZE`). No change requested.

crates/cgka-session/fixtures/README.md
```diff
@@ -0,0 +1,17 @@
+# Session compatibility fixtures
+
+`session-promotion-v1.bin` is a synthetic SQLCipher account-device database
+prepared for the public `AccountDeviceSession` promotion-facade regression.
+It contains one confirmed two-member group and one format-1 Welcome row. The
+database was created by the normal session API on 2026-08-14, then that one row
```
erskingardner, line 6: (Non-blocking) The storage fixture retains its writer source (`write-v0.9.12-fixture.rs.txt`) so provenance is reproducible; this one only describes the generation recipe in prose ("created by the normal session API ... then converted"). Consider stashing the short script/commands used — it will make regeneration at the next schema boundary much less archeological. Fine to defer if you judge the storage fixture sufficient.

erskingardner (commented):
Approve (posting as comment since GitHub blocks self-approval) — clean operational slice that closes out the #1414 scope. Reviewed at head 70c676b2 (CI fully green at this SHA).

What I verified beyond the diff:

- **Readiness/ordering**: the promotion runs in the steady-state worker loop, which is only entered after readiness is signalled, hydration completes, and the initial catch-up (which holds `&mut client`) resolves. No aliasing with in-flight startup sync.
- **Close semantics**: `SqliteAccountStorage::clone()` shares the `Arc<SharedConnectionInner>`, so the maintenance handle the session retains is terminally closed together with the engine's handle — the comment on `AccountDeviceSession::open` is accurate.
- **Transient classification**: `SessionError::is_transient` → `StorageError::is_transient` matches only `Busy`, and the storage batch itself already retries busy internally — so the halt-vs-retry policy in `run_legacy_message_promotion_batch_with` behaves as documented.
- **Fixture digests**: both checked-in binaries hash to exactly the SHA-256 values recorded in their READMEs.
- **The `test_support.rs` indirection**: justified — `production_library_sources_do_not_write_direct_output` exempts files by the `test_support.rs` name, so the benchmark's `println!` has to live in one.
- **Docs/runbook consistency**: the 3.01x additional / 4.01x peak → 3.25x free-space recommendation arithmetic checks out; `release.md` correctly acknowledges that a pinned previously-shipped binary keeps its older error wrapper while still refusing before application-table access.
- CodeRabbit's earlier actionable comments (session-boundary test fixture, stale 256-row doc limit) appear addressed at this head.

Two non-blocking notes inline.

The 64-tick / ~16-minute full-promotion window for 2,048 legacy rows is the right tradeoff for keeping this off the open path, and legacy reads working throughout makes the sweep genuinely optional. Nice work keeping this surgical.

PR #1435, opened 2026-08-14
Replace live account projection snapshots with delta writes

### LINKED ISSUE #774: marmot-app: `save_state` re-UPSERTs the entire retained `seen_events` window (up to 16,384 rows) on every inbound event — per-message O(window) encrypted write amplification on the runtime thread
## Summary

`save_state` persists the account's `seen_events` dedup window by looping over **every** retained id and issuing an individual `INSERT … ON CONFLICT(event_id) DO UPDATE SET seen_at = excluded.seen_at`, followed by a correlated-subquery `DELETE`. Because the in-memory `seen_events` `Vec` is capped at `MAX_SEEN_EVENT_IDS = 16_384`, once an account has seen more than 16,384 events (the steady state for any real user), **every single inbound event** re-writes all ~16,384 rows — even though only one id is actually new. This runs synchronously against SQLCipher on the tokio runtime thread the account worker `select!`s on.

## Location

- `crates/marmot-app/src/projection.rs:329-350` — the per-id UPSERT loop + `DELETE`:
  ```rust
  let retained_start = state.seen_events.len().saturating_sub(crate::MAX_SEEN_EVENT_IDS);
  for event_id in &state.seen_events[retained_start..] {
      tx.execute(
          "INSERT INTO seen_events (event_id, seen_at) VALUES (?1, ?2)
           ON CONFLICT(event_id) DO UPDATE SET seen_at = excluded.seen_at",
          params[event_id, now],
      )?;
  }
  tx.execute(
      "DELETE FROM seen_events WHERE event_id NOT IN (
           SELECT event_id FROM seen_events ORDER BY seen_at DESC, rowid DESC LIMIT ?1)",
      params[limit_to_i64(crate::MAX_SEEN_EVENT_IDS)],
  )?;
  ```
- `crates/marmot-app/src/lib.rs:230` — `pub(crate) const MAX_SEEN_EVENT_IDS: usize = 16_384;`
- `crates/marmot-app/src/lib.rs:625-628` — `prune_seen_events` drains the in-memory `Vec` down to `MAX_SEEN_EVENT_IDS`, so `retained_start` spans the full 16,384-entry window when full.
- Driven per non-duplicate delivery from the live account-worker receive loop: `crates/marmot-app/src/client/sync.rs:197-202` (`remember_seen_event` then `save_state` inside the `loop { … }`), and also `sync.rs:141/256/308`.

## Trigger / impact

1. Account has received > 16,384 events over its lifetime → `seen_events.len()` is pinned at 16,384.
2. For each new inbound Nostr delivery, `sync::next_event` calls `remember_seen_event` (appends 1 id, drains 1) and then `save_state`.
3. `save_state` re-UPSERTs all 16,384 rows (setting `seen_at = now` on every one) plus the `NOT IN (SELECT … LIMIT 16384)` DELETE — **~16k encrypted UPSERTs per message**.

Consequences:
- **Write amplification:** O(window) DB work per message instead of O(1). A relay streaming a burst of N messages does ~16,384·N UPSERTs.
- **Blocking the runtime thread:** this is synchronous SQLCipher (encrypt-on-write) executed on the account-worker runtime thread, so it directly stalls inbound processing (same hazard class as #740/#762/#741, but a distinct site and a distinct root cause — write volume, not a one-off blocking call).
- **Defeats the intended LRU column:** every retained row's `seen_at` is rewritten to the current instant on every save, so the `seen_at DESC` ordering the prune relies on is flattened to "all equal"; pruning then falls back entirely to the `rowid DESC` tiebreak, undermining the recency semantics #422 was trying to establish.

## Suggested fix

- Persist only the id(s) actually appended since the last save (e.g. thread the freshly-added id out of `remember_seen_event`, or upsert just `state.seen_events.last()`), rather than the whole window.
- Run the `DELETE` prune opportunistically (only when a new row is actually inserted / the window overflows), not on every `save_state`.
- Do not rewrite `seen_at` for rows whose value is unchanged.

## Dedup note

Checked open + closed history.
- **#422** (closed) is the fix that *introduced* this: it changed `INSERT OR IGNORE` → `ON CONFLICT … DO UPDATE SET seen_at = excluded.seen_at` specifically to refresh recency. That change is correct for a single re-seen id but, combined with the full-window loop, makes every save rewrite all 16k rows. Distinct defect (write amplification / runtime-thread cost), not the recency-staleness bug #422 addressed.
- **#425** is the in-memory `Vec::contains` on ingest; **#681** is the unbounded/O(n²) in-memory dedup. Both are memory/CPU on the in-RAM structure, not the per-event DB write volume in `save_state`.
- The blocking-SQLCipher-on-runtime-thread issues (#740/#762/#741) each name different sites (notification collection, media awaits, subscribe HOL). None identifies this `save_state` per-message O(window) write amplification.

_Found during a methodical multi-crate code review; verified against source (`projection.rs:329-350`, `lib.rs:230/625`, `client/sync.rs:197-202`)._

erskingardner:
## Summary

- track newly observed seen-event IDs and changed group projections between successful checkpoints
- persist exact account-projection deltas instead of re-saving the complete retained account snapshot
- keep cursor updates, deletion-frontier clears, application-event acknowledgements, seen-event observations, and group updates atomic
- reserve full snapshot replacement for import and rebuild paths

## Root cause

Live account operations passed the full in-memory account projection to SQLite. Once the dedup window reached 16,384 entries, every inbound event refreshed every retained `seen_events` row. The same transaction also walked every retained group and component, even when accepting an invite or toggling an archive flag changed only one group.

## Behavior

The account worker now retains a bounded unsaved suffix of seen-event IDs and the exact set of changed group projections. A successful checkpoint clears those pending deltas; a failed checkpoint retains them for retry. SQLite applies supplied groups as per-group full replacements without deleting absent groups, and runs seen-event pruning only when a new row can grow the table.

Regression coverage uses SQLite write-audit triggers to prove that:

- two new observations write only two seen-event rows and prune to the bound
- refreshing one observed ID stays O(1)
- changing one group/component does not touch sibling groups
- a group-only archive checkpoint writes no seen-event rows
- the app worker checkpoints the pending seen suffix and target group together

## Validation

- `just fast-ci`
- `cargo test -p storage-sqlite account_projection_delta_writes_only_observations_and_changed_groups`
- `cargo test -p marmot-app live_group_archive_checkpoints_seen_and_target_group_deltas`
- full local crate suites during development: 374 storage tests passed with 2 ignored; all 645 marmot-app library tests passed

A broader relay-runtime run exposed five failures that were reproduced unchanged on the then-current untouched base, so they are not regressions from this patch.

Fixes #774

## Summary by CodeRabbit

* **Performance**
  * Improved account state saving by persisting only changed groups and newly seen events.
  * Reduced unnecessary database updates during routine synchronization and refreshes.

* **Reliability**
  * Preserved unaffected group data during incremental saves.
  * Improved handling of archived groups, membership changes, and route updates.

* **Bug Fixes**
  * Ensured seen-event history and group projections remain consistent after concurrent updates.
  * Added safeguards against stale or incorrectly removed group data.

### DIFF
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -113,6 +113,15 @@ pub struct AppClient {
     /// persisted. Before this index existed, every inbound delivery and every
     /// publish-report batch rebuilt a `HashSet` from the full 16k-entry ring.
     pub(crate) seen_events_index: HashSet<String>,
+    /// Number of ids at the tail of `state.seen_events` observed since the last
+    /// successful account-projection checkpoint. The count saturates at the
+    /// bounded ring length, so even a catch-up batch larger than the window
+    /// persists only the final retained ids. It is cleared only after commit.
+    pub(crate) pending_seen_event_count: usize,
+    /// Exact group projections changed since the last successful checkpoint.
+    /// Live saves replace only these groups; full snapshot replacement is
+    /// reserved for import/rebuild paths outside the account worker.
+    pub(crate) pending_group_projection_updates: HashSet<String>,
     /// Group-system timeline rows synthesized during the most recent publish
     /// path. The runtime account worker drains this after each command and
     /// broadcasts `ProjectionUpdated` so live timeline subscriptions refresh.
@@ -870,7 +879,7 @@ impl AppClient {
         // into a false failure that invites the caller to create a duplicate.
         let local_projection_started_at = Instant::now();
         let local_projection_saved = match self.add_group(&group_id) {
-            Ok(()) => match self.app.save_state(&self.state) {
+            Ok(()) => match self.save_state_with_pending_local_group_deletion_frontier_clears() {
                 Ok(()) => true,
                 Err(error) => {
                     tracing::warn!(
@@ -1344,7 +1353,7 @@ impl AppClient {
             self.remember_published_reports(&effects);
             self.refresh_group(group_id);
             self.prune_plaintext_retention_for_group(group_id)?;
-            self.app.save_state(&self.state)?;
+            self.save_state_with_pending_local_group_deletion_frontier_clears()?;
             self.queue_own_group_system_projection_updates(&effects);
             Ok::<_, AppError>(())
         })();
@@ -1407,7 +1416,7 @@ impl AppClient {
         self.remember_published_reports(&effects);
         self.refresh_group(group_id);
         self.cleanup_stale_push_tokens_best_effort(group_id);
-        self.app.save_state(&self.state)?;
+        self.save_state_with_pending_local_group_deletion_frontier_clears()?;
         self.queue_own_group_system_projection_updates(&effects);
         Ok(send_summary_from_effects(&effects))
     }
@@ -1452,7 +1461,7 @@ impl AppClient {
         self.record_human_action_succeeded(group_id, &audit_context, &effects);
         self.remember_published_reports(&effects);
         self.refresh_group(group_id);
-        self.app.save_state(&self.state)?;
+        self.save_state_with_pending_local_group_deletion_frontier_clears()?;
         // Enabling lifecycle-v1 changes eligibility, not group history. It
         // intentionally emits no kind-1210 system row.
         Ok(send_summary_from_effects(&effects))
@@ -1500,7 +1509,7 @@ impl AppClient {
                 "durable disband request outpaced composer draft cleanup"
             );
         }
-        self.app.save_state(&self.state)?;
+        self.save_state_with_pending_local_group_deletion_frontier_clears()?;
         self.runtime
             .disband_request(group_id)?
             .map(Into::into)
@@ -1667,7 +1676,7 @@ impl AppClient {
         };
         self.record_human_action_succeeded(group_id, &audit_context, &effects);
         self.remember_published_reports(&effects);
-        self.app.save_state(&self.state)?;
+        self.save_state_with_pending_local_group_deletion_frontier_clears()?;
         self.queue_own_group_system_projection_updates(&effects);
         // A local leave / decline is a voluntary departure, recorded as `Left`
         // so the chat list can distinguish it from an involuntary removal. The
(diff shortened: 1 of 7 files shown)


erskingardner (commented):
Recommending merge (posted as comment — GitHub won't let me approve your own PR). Reviewed at c829ea5 against linked issue #774.

The change matches the issue's root cause exactly (full-window UPSERT + unconditional prune per save; full group-set walk per save) and implements it with delta writes that stay atomic in one transaction.

Verified at head:
- **Mutation coverage**: every `state.groups` mutation site either marks `pending_group_projection_updates` (create/invite/leave/archive/pending-confirmation sends, `refresh_group`/`add_group` via before/after record compare, both `observe_event` ingest seams, prior-route adopt/prune, member-count backfill) or owns persistence directly (`delete_group_local` → `delete_local_group_data`; `MarmotApp::set_group_archived` → its own single-group delta). No non-test callers of the old `save_state` full-snapshot path remain; import/migration keeps the full-replacement API.
- **Seen suffix accounting**: `state.seen_events` is appended/pruned only inside `remember_seen_event`, and the counter resets at open, so the tail-slice (`len - pending`) invariant holds; failure paths retain pendings because clears happen only after the save returns `Ok`.
- **Storage delta fn**: per-group full replacement scoped by `group_id_hex` (`delete_stale_group_components` is per-group), never deletes absent groups, keeps frontier clears + cursor merge + acks in the same tx, and the prune-on-insert/refresh-without-prune logic matches the issue's ask and fixes the `seen_at`-flattening side effect.
- The write-audit-trigger test is a genuinely strong regression harness (counts inserts/updates/deletes per table).

Non-blocking notes:
- The dirty-mark discipline is now a manual invariant for any *future* `state.groups` mutation; it's well-guarded today by the compare checks at the three construction chokepoints, and the field docs state the contract — just worth keeping in mind in review going forward.
- CI was still mid-run when I checked (Rust check/fmt/Tamarin green; clippy, Rust tests, relay runtime pending). The relay-runtime failures called out in the body were reproduced on base, which matches what CI should confirm. Merge on green.

PR #1440, opened 2026-08-14
Bound deferred peel work before outbound sends

### LINKED ISSUE #1176: Bound deferred-peel work in foreground send preflight and yield to queued intent
## Problem

Foreground app sends synchronously run a deferred-peel sweep before outbound arbitration. A sweep is row-bounded, but each row can perform retained-snapshot rollback/load/peel/restore work, so the existing 64-row cap is not a latency bound.

Grounded against `origin/master` at `b609ccd3796aea14bf0f0478073520179820d999`.

## Runtime evidence

A session-scoped audit of all 212 reachable Goggles exports found 4,064 paired sends. A fresh re-export of the eight high-latency groups produced 907 paired sends:

| Deferred failures during the send | Sends | Median | p95 | Sends >=10s |
| --- | ---: | ---: | ---: | ---: |
| None | 652 | 33 ms | 247 ms | 1 |
| One or more | 255 | 3.683 s | 74.868 s | 62 |

The >=10-second rate was about 158x higher when historical deferred messages were retried during the send. One 30.423-second operation retried 47 distinct deferred messages; two adjacent 28-30-second operations retried 64 each.

This does not explain every long send: a separate 281-second operation had no deferred activity and showed transport failures. This issue is limited to foreground deferred-peel scheduling.

## Source-confirmed mechanism

- [`should_queue_outbound_intent`](https://github.com/marmot-protocol/mdk/blob/b609ccd3796aea14bf0f0478073520179820d999/crates/cgka-engine/src/message_processor/mod.rs#L469-L488) calls convergence settlement before deciding whether to queue.
- [`advance_convergence_inputs_until_settled`](https://github.com/marmot-protocol/mdk/blob/b609ccd3796aea14bf0f0478073520179820d999/crates/cgka-engine/src/message_processor/mod.rs#L599-L642) synchronously calls `retry_deferred_peels`, including when no unresolved convergence input remains.
- Snapshot probing can perform rollback/load/peel/restore work per row in [`message_processor/ingest.rs`](https://github.com/marmot-protocol/mdk/blob/b609ccd3796aea14bf0f0478073520179820d999/crates/cgka-engine/src/message_processor/ingest.rs#L2244-L2305).

The fingerprint gate, retry budget, 256-row retention cap, and 64-row sweep cap remain valuable. The residual problem is that a row-count bound does not constrain foreground wall time.

## Required behavior

- Preserve MLS ordering and the Stable-only ciphertext-preparation gate.
- Permit convergence-critical/newly-peelable work before outbound arbitration only within an explicit, tested foreground time/work budget.
- When that budget is exhausted, durably queue the user intent, return promptly, and continue deferred maintenance through the scheduler.
- Give the queued user intent its existing fairness turn before another historical-only sweep.
- Do not bypass unknown deferred state and immediately prepare ciphertext from a potentially stale membership state.
- Keep terminal removal, disband, quarantine, and Unrecoverable gates authoritative.

## Acceptance criteria

- Foreground send latency is bounded by a documented deadline/work budget rather than only by rows per sweep.
- A deterministic test with 47-64 `Notify`-gated historical deferred rows proves the API yields/queues before waiting for the full sweep.
- A separate regression proves newly available authenticated context required for the current branch is processed before dependent outbound preparation.
- Queued intent ordering, restart recovery, removal cleanup, retry fingerprints, terminal failures, and the frozen-pass fairness slot remain intact.
- Add phase metrics for required convergence, deferred history, snapshot probing, queue acceptance, and wire preparation, using monotonic durations and privacy-safe aggregate fields.
- Add recorder-on/off benchmarks across 1, 5, and 7+ member scenarios with epoch churn and restart.

## Test coverage

- Extend `crates/cgka-engine/tests/deferred_peel_lifecycle.rs` with deterministic slow-peeler/yield coverage.
- Extend `crates/cgka-engine/tests/distributed_convergence.rs` for convergence-critical retry plus queued-intent fairness.
- Add a 7+ member delayed-history/restart scenario to `cgka-conformance-simulator`.

## Related / non-goals

- #339 / PR #726 introduced the current deferred-peel lifecycle bounds.
- #806 / PR #1110 introduced durable frozen passes and the queued-intent fairness slot.
- #1109 owns durable inbound row/byte quota policy; this issue owns foreground scheduling latency.
- #1171 investigates delayed-message expiry during catch-up.
- Android issue #913 added client-side lock/FFI timing; the root scheduling change belongs here.
- Relay publication latency and `PendingPublish` intent retention are separate issues.

erskingardner:
## Summary

- bound foreground deferred-peel work to 250 ms and four rows per outbound operation
- durably retain application, proposal, and commit intents when convergence cannot finish safely, without emitting wire output
- persist per-row retry truth and a contested-generation barrier so restart cannot converge a partial batch
- prioritize all queued outbound intents over historical-only maintenance while preserving the existing one-attempt administrative reservation
- add privacy-safe phase metrics, SQLite migration 48, architecture documentation, simulator coverage, and a benchmark matrix

## Root cause

Outbound sends synchronously retried the deferred transport backlog before preparing their own wire output. A sufficiently large or slow backlog could therefore block the caller for an unbounded amount of time. The retry progress used by this path was also not durable enough to make partial contested-batch processing safe across restart.

## User and developer impact

Outbound APIs now return promptly with the original intent durably accepted when the foreground budget is exhausted. The queued intent is regenerated only after required convergence is safe, so callers do not need to retry blindly and no premature ciphertext is emitted. Historical maintenance cannot starve queued work.

The 250 ms / four-row limits and their versioning decision are pinned in the convergence constant ledger. New aggregate metrics separate required-convergence, deferred-peel, queue-accept, and wire-preparation latency.

## Validation

- `just fast-ci` after rebasing onto current `origin/master`
- `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks`
- `cargo test -p storage-sqlite` — 375 passed, 3 ignored
- full simulator coverage, including the new seven-member future-epoch backlog/restart scenario
  - one parallel process-orchestrator run encountered `Address already in use (os error 48)`
  - the exact affected test passed in isolation, and the remaining simulator binaries passed
- `just convergence-ledger-gate` — 31 IDs, 33 declarations
- blocked-send benchmark: approximately 255–258 ms in the recorder-off, one-member case

Fixes #1176

## Summary by CodeRabbit

- **New Features**
  - Added bounded processing for deferred messages, preventing long-running outbound operations.
  - Outbound messages can now be safely queued and completed later when processing capacity is reached.
  - Queued work and deferred progress resume across application restarts.

- **Bug Fixes**
  - Improved convergence handling when message history is incomplete or contested.
  - Prevented partial results from being finalized before all deferred messages are resolved.
  - Ensured deferred state is removed during group cleanup.

- **Improvements**
  - Added expanded telemetry for outbound preparation, deferred processing, queue delays, and completion outcomes.


crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -695,16 +807,120 @@ impl<S: StorageProvider> Engine<S> {
...
+        {
+            if self.advance_required_convergence_inputs(group_id, now_ms, true)?
+                != AdvanceConvergenceStatus::Settled
+            {
+                return Ok(true);
+            }
+            return self
+                .stage_due_self_remove_auto_commit(group_id, now_ms)
+                .await;
+        }
```
erskingardner: This fast path (and `advance_before_queued_outbound_intents`, which has the same shape) treats *every* deferred row as historical-only once a queue exists and no contested barrier is set. That looks broader than the #1176 decision in two ways:

1. A row that arrives while the queue is non-empty has never been examined under the complete current fingerprint, yet neither queue-present path gives it the bounded current-generation opportunity — the new test even pins this ("queued drain must skip even newly untested historical-only rows"), but that row is not "historical-only" under the addendum's definition ("a `PeelDeferred` row is blocking only while it has not been examined under the current complete peel-context fingerprint").
2. After an epoch advance changes the fingerprint, all retained rows become "potentially relevant again" per the addendum, and scheduler priority lists an incomplete changed-context generation *above* queued outbound intents — but these paths skip them too whenever the queue is non-empty.

Both also read against the unconditional MUST this PR adds to `cgka-engine-spec.md` ("MUST give every deferred row not yet definitively tested under the current peel-context fingerprint a bounded foreground opportunity"). If queue-outranks-all-opaque-history is the intended rule, can we narrow the spec bullet and the ledger wording to say so explicitly (and reconcile the test naming)? Otherwise the queue-present paths still need the ≤4-row/250 ms current-generation examination before wire preparation.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -695,16 +807,120 @@ impl<S: StorageProvider> Engine<S> {
...
+                .convergence_pass(group_id)?
+                .is_some_and(|pass| {
+                    pass.phase == cgka_traits::ConvergencePassPhase::Completed
+                        && pass.fairness_slot_available
+                })
+            {
+                // Preserve the existing one-attempt admin reservation. The
+                // queue owns this completed-pass boundary even though newer
+                // authenticated input is waiting behind it.
+                return Ok(AdvanceConvergenceStatus::Settled);
```
erskingardner: Checking the outstanding fairness slot *before* `has_unresolved_convergence_inputs` changes master's ordering for an app-only queue. On master, `load_or_open_convergence_pass` consumed the slot whenever no admin intent was queued ("queued application messages must never park pass admission"), so pending authenticated input settled before any wire prep. Here, any outstanding slot short-circuits to `Settled` up front, so a direct send or the queued drain can prepare output ahead of pending authenticated commits/proposals that arrived behind the frozen pass (e.g., a pending remove of self). Should this mirror the full path's order — converge first, then the slot check after a settled run, which is what `advance_convergence_inputs_with_execution` does? If the early-`Settled` is deliberate, a test pinning the app-only-queue + outstanding-slot + pending-authenticated-input combination would be valuable.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -1128,12 +1440,28 @@ impl<S: StorageProvider> Engine<S> {
...
+            } else {
+                DeferredPeelWorkStatus::Pending
+            };
+            self.note_foreground_deferred_phase(
+                sweep_started,
+                foreground_budget_ms,
+                0,
+                deferred.len(),
+                crate::engine_metrics::DeferredPeelMetricOutcome::BudgetExhausted,
+            );
```
erskingardner, line 1440: nit: in the foreground this normalization-pending branch reports `BudgetExhausted` and increments `foreground_deferred_budget_exhausted`, though no row was attempted and neither bound was reached. A first send after upgrade with legacy rows will inflate the exhausted counter with zero-attempt samples and skew the new outcome dashboards; consider `Completed` (or a dedicated normalization outcome) for this return.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -1247,65 +1661,132 @@ impl<S: StorageProvider> Engine<S> {
...
-            lifecycle.last_context_fingerprint = Some(fingerprint);
-            lifecycle.wall_high_water_ms = lifecycle.wall_high_water_ms.max(now.wall_ms);
-            // Persist the consumed context before invoking the peeler. A crash
-            // may conservatively consume an attempt, but cannot reset it.
-            self.storage.put_message(&record)?;
+            execution.consume_row();
+            attempted += 1;
+            let reingest = self.reingest_deferred_peel_row(group_id, &record, sweep);
+            let result = if let Some(remaining) = execution.remaining() {
+                match tokio::time::timeout(remaining, reingest).await {
```
erskingardner, line 1648: This is (I believe) the first tokio timer dependency inside cgka-engine core logic — a host driving `Engine::send` outside a tokio runtime with time enabled panics here rather than degrading. All in-repo hosts qualify (marmot-app, CLI, simulator, and uniffi exports with `async_runtime = "tokio"`), so this is fine in practice, but the requirement is now implicit; worth a one-line note near the budget field or on the `Engine` send API.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -695,16 +807,120 @@ impl<S: StorageProvider> Engine<S> {
...
+            self.foreground_deferred_peel_budget_ms,
+            self.foreground_deferred_peel_rows,
+        );
+        match self
+            .advance_convergence_inputs_with_execution(
+                group_id,
+                now_ms,
+                DeferredPeelExecution::Foreground(&mut budget),
+            )
             .await?
```
erskingardner, line 856: The budget clock starts here, before required authenticated convergence runs inside the same preflight. A heavy-but-necessary converge can consume the whole 250 ms and leave zero row attempts, queueing the intent without any "bounded foreground opportunity" for current-context rows — a safe, bounded outcome, but subtly different from the ledger/spec wording ("gives current-context deferred input one bounded foreground opportunity"). Either scope the clock to the deferred phase, or note the shared window explicitly in the E11/spec text so the documented contract matches the code.

erskingardner (commented):
Reviewed at 791b1a5.

The overall architecture matches the #1176 decision addendum closely: the ForegroundPeelBudget/DeferredPeelExecution split is clean, moving completion truth to durable per-row fingerprints (deleting the in-memory gate+cursor) is a net simplification with better restart semantics, the contested-generation barrier closes the partial-prefix adjudication hole the old in-code comment described, migration 48 follows the record-blob convention with a cascade test, and the ledger/telemetry/spec doc updates are unusually thorough. The test matrix (47/64-row Notify-gated backlogs, restart-resume, contested-restart, simulator scenario) directly targets the acceptance criteria.

Two substantive contract questions in the core send path (inline):

1. **"Historical-only" scope drift in the queue-present fast paths.** Once any intent is queued and no barrier exists, neither `should_queue_outbound_intent` nor the drain gives never-examined-under-current-fingerprint rows (fresh arrivals, or all rows after an epoch change) their bounded current-generation opportunity. The new spec bullet says an unconditional MUST; the new tests pin the skip. Whichever rule is intended, the spec/ledger text and the code need to agree.

2. **Fairness-slot checked before unresolved-inputs in `advance_required_convergence_inputs`.** For an app-only queue this lets wire prep race ahead of pending authenticated input that master would have settled first. Likely fixable by mirroring the full path's converge-then-slot-check order.

Plus three smaller notes inline (telemetry mislabel on the normalization branch, new implicit tokio-timer requirement in engine core, budget clock covering required convergence).

**Validation gap:** I reviewed statically per review policy; most GitHub CI checks (rust tests, clippy, simulator, tamarin) were still pending when I checked, and CodeRabbit was mid-review — those results should gate the merge.

erskingardner:
Addressed the full review set in 790fc771.

Core contract fixes:
- authenticated convergence now runs before a completed-pass fairness slot can release queued work;
- newly untested or changed-fingerprint deferred rows retain their bounded current-generation priority even when intents are queued, while unchanged historical rows remain behind the queue;
- the 250 ms clock covers only deferred-peel phases and excludes required authenticated convergence;
- lifecycle normalization has a distinct pending metric instead of being mislabeled as zero-attempt budget exhaustion.

Also addressed all review cleanup items: explicit Tokio time-driver docs, deterministic timing/restart tests, split queue/fingerprint tests, exhaustive metric-merge coverage, constant snapshot pins, local-only telemetry wording, benchmark measurement cleanup, and peeler/builder deduplication.

Local validation:
- deferred_peel_lifecycle: 14 passed
- distributed_convergence: 88 passed
- cgka-engine library with all features: 98 passed
- production-shaped 47-row simulator scenario: passed
- simulator metric merge and protocol decision gates: passed
- just fast-ci: passed, including workspace all-target checks and clippy with and without OTLP features

The previously failing constant-snapshot check is fixed. The new GitHub Actions run is now in progress.

erskingardner (commented):
**Re-review verdict: recommend merge.** (GitHub will not register a formal approve from this account since it matches the PR author.)

Re-reviewed at 790fc771. All five findings from my previous review are addressed, and the fixes landed the way I would have hoped:

1. **Historical-only scope drift** — the queue-present fast paths are gone; every outbound (direct send or queued regeneration) now runs the same budgeted `advance_convergence_inputs_with_execution`, and the per-row durable fingerprint filter is what keeps tested rows historical-only. The spec bullet now states the rule this PR actually implements ("current-fingerprint deferred work remains safety-critical ahead of every queued outbound intent"), and the reworked `queued_intent_drains_before_historical_only_maintenance` pins both halves: fresh rows get a definitive attempt before wire; fingerprint-tested rows do not block. Code and contract now agree.

2. **Fairness-slot ordering** — `advance_required_convergence_inputs` was deleted outright (net -102/+82 in mod.rs), leaving only the converge-then-slot-check order. The new `queued_app_regeneration_converges_authenticated_input_before_fairness_slot` test is exactly the app-only-queue + outstanding-slot + pending-authenticated-input case, and it proves the drain converges first and regenerates at the post-convergence epoch.

3. **Metric mislabel** — dedicated `NormalizationPending` outcome/counter, with a unit test.

4. **Tokio timer requirement** — documented on `CgkaEngine::send`.

5. **Budget clock scope** — `start_phase`/`finish_phase` now accrue time only inside the deferred-peel phase; the spec/ledger text matches ("authenticated-convergence time is excluded").

Also good in this round: `start_phase`/`finish_phase` are correctly paired on the error path (`finish` runs before `?`), the hoisted `exhausted()` check makes dropping `.max(1)` safe, the restart test now derives the completed prefix from durable fingerprints instead of assuming exact attempt counts (robust on loaded CI), the bench matrix hoists contract assertions into a one-time probe so measured iterations are clean, and the peeler fixture dedup is a nice net simplification.

One non-blocking observation: with a queue present, drain advances now also run under the 4-attempt/250 ms budget, so a large changed-context backlog is examined ~4 rows per scheduler wake before the queued intent regenerates (vs 64-row background sweeps when no queue exists). That is the deliberate price of keeping the drain latency-bounded too, and the seven-member simulator scenario passes with it — just worth being aware of for very large backlogs.

Validation: GitHub CI on this head is green (26 pass; only the path-filtered release publish skips), including rust tests 1-4, clippy, simulator/vectors, and tamarin. My review remains static per policy.

PR #1424, opened 2026-08-13
test(simulator): assure cross-route process convergence

erskingardner:
## Summary

- extract runner-owned reversible retained-relay control for reuse by the app-runtime and process adapters
- run the canonical four-party cross-route scenario unchanged through four isolated app processes with separate encrypted roots
- replace SIGSTOP-based offline simulation with a real runtime/transport disconnect and durable reopen
- enforce strict public schedule, state, application, invalidation, quiescence, and restart assertions
- update the convergence route matrix and Milestone 6 tracking plan with the new evidence and remaining boundaries

## Why

The app-runtime adapter already had positive post-unification route-equivalence evidence, but the process adapter could not select the same action-addressed retained history. Its SIGSTOP offline model also left relay subscriptions open, allowing events to buffer while a participant was nominally offline and invalidating controlled-route assumptions.

The process orchestrator now checkpoints the public observation and gracefully stops the child/runtime for offline actions. Reconnect reopens the same durable SQLCipher root. Runner-owned relays advertise reversible visibility control; externally supplied relays fail capability preflight instead of claiming control the runner does not own.

This PR changes simulator/process infrastructure and assurance documentation only. It does not change production convergence-engine or app behavior.

## Assurance boundary

The process oracle covers the canonical schedule and IR digest, intermediate branch split, final epoch/roster/admin/profile agreement, the complete duplicate-free application-message set, losing-branch withdrawal behavior, pending-confirmation state, public commitment equality, and a real durable restart.

The public node protocol still does not expose exact MLS-state commitments, durable convergence-input dispositions, or active decryptability probes. Those claims remain owned by engine-capable subjects. Container/VM execution and every-durable-transition permutations remain open.

## Verification

- `cargo test -p cgka-conformance-simulator --test process_orchestrator --locked` — 13 passed, 2 ignored manual soaks
- `cargo test -p cgka-conformance-simulator --test app_runtime_adapter --locked` — 9 passed
- `cargo test -p cgka-conformance-simulator --test route_assurance --locked` — 8 passed
- `cargo test -p cgka-conformance-simulator --test process_orchestrator four_party_cross_route_recovery_process_equivalence_soak --locked -- --ignored --exact --nocapture` — 20/20 trials passed in 1,133 seconds
- final focused four-process route and disconnect/reconnect lifecycle regressions passed after the last audit adjustment
- `just fast-ci`

## Summary by CodeRabbit

* **New Features**
[...]

PR #1441, opened 2026-08-14
test(simulator): assure cross-route container convergence

erskingardner:
## Summary

- extract the canonical four-party cross-route recovery scenario and strict public oracle into one shared simulator fixture
- let the process orchestrator use an explicitly supplied retained-relay control implementation
- add a runner-owned, file-backed relay control plane for container campaigns using ephemeral secrets and opaque event tokens
- run the shared journey through four non-root OCI participant processes with separate encrypted durable roots
- schedule the real-container checkpoint in nightly and weekly/manual lanes and retain its canonical scenario, normalized manifest, process report, receipt, and failure corpus
- reconcile the convergence plan, route matrix, scenario catalog, and distributed-campaign documentation with the new evidence

## Why

PR #1424 proved the four-party journey through isolated host processes, but the distributed container runner supplied an external relay URL and therefore could not execute the scenario's retained-event visibility actions. Capability preflight correctly rejected the run before any scenario action.

This adds an explicit simulator-only control boundary instead of weakening or rewriting the canonical scenario. The repository-owned relay records only publication class and opaque run-local tokens in an owner-only temporary directory. Raw Nostr event ids, payloads, ciphertext, keys, participant databases, and the control secret are not retained in CI evidence.

## Impact and boundary

This changes simulator, campaign-runner, CI, and documentation code only. It does not change `cgka-engine`, production convergence constants, Marmot app behavior, or protocol semantics. Existing container manifests retain their previous behavior unless `enable_retained_relay_control` is explicitly enabled with the repository-owned relay command.

The container route now supplies public protocol/application projection evidence for the shared cross-route journey. Exact MLS state, durable convergence-input dispositions, active decryptability on black-box adapters, VM execution, and every-durable-transition permutations remain open.

## Verification

- `just fast-ci`
- `cargo test -p cgka-conformance-simulator --test process_orchestrator four_party_cross_route_recovery_processes_match_unified_route --locked -- --exact`
- `docker build -f Dockerfile.convergence-campaign -t marmot-conformance:local .`
- `CGKA_DISTRIBUTED_ARTIFACTS_DIR="$PWD/target/cgka-distributed-container-evidence-v3" cargo test -p convergence-campaign-runner --test container_runtime four_party_cross_route_recovery_containers_match_unified_route --locked -- --ignored --exact --nocapture`
  - passed in 61.20 seconds with clean container/network teardown

## Summary by CodeRabbit

* **New Features**
  * Added a shared four-party cross-route recovery scenario across in-process, process, and container runs.
  * Added optional retained-relay visibility controls for container campaigns.
  * Added file-backed relay control and configurable relay control directories.
  * Added cross-route recovery testing with canonical scenario validation and evidence capture.

* **Bug Fixes**
  * Improved process-report validation and malformed-report handling.

* **Documentation**
  * Expanded guidance for container checkpoints, recovery evidence, relay controls, and known coverage gaps.


crates/convergence-campaign-runner/src/relay_file_control.rs
```diff
@@ -0,0 +1,442 @@
...
+                )
+            })?;
+        let hidden_path = self.control_root.join(HIDDEN_EVENTS).join(event_token);
+        if visible {
+            std::fs::remove_file(&hidden_path)
+                .map_err(|error| control_error("relay_event_not_removed", error))?;
+        } else {
+            fs_private::create_new_private(&hidden_path)
+                .map_err(|error| control_error("relay_event_already_removed", error))?;
+        }
```
erskingardner, line 337: Non-blocking: both `map_err` arms collapse every io error into the semantic-state codes `relay_event_not_removed` / `relay_event_already_removed`. Those codes assert a specific visibility state (and mirror the in-process control, where the set operations really are infallible), but here they also fire on genuine I/O failures — e.g. `EACCES` on the bind mount, or the control root going away mid-run. Since these codes flow into the failure corpus and drive diagnosis, an infrastructure/permissions regression would be misclassified as a visibility-state bug. Consider matching on `ErrorKind::NotFound` / `ErrorKind::AlreadyExists` for the semantic codes and mapping anything else to a distinct `relay_event_visibility_io`-style code.

crates/convergence-campaign-runner/src/relay_file_control.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    fn count(&self, filter: Filter) -> BoxedFuture<'_, Result<usize, DatabaseError>> {
+        Box::pin(async move { Ok(self.query(filter).await?.len()) })
+    }
+
+    fn query(&self, filter: Filter) -> BoxedFuture<'_, Result<Events, DatabaseError>> {
+        Box::pin(async move {
+            let mut database_filter = filter.clone();
+            database_filter.limit = None;
```
erskingardner, line 140: Nit: this mirrors `relay_control.rs`'s `query`, including the deliberate limit-after-hidden-filter ordering — but the comment documenting *why* ("otherwise one hidden newest event would incorrectly reduce the visible result below the relay client's requested page size") didn't come along. The in-process twin has a regression test pinning that behavior; worth copying the comment here too so a future cleanup doesn't "simplify" the ordering back into the page-size bug.

erskingardner (commented):
Review outcome: **would approve** (posted as a comment because GitHub won't let the PR author approve their own PR). This is a clean, well-scoped extraction plus a carefully mirrored control boundary.

Verified in this review:
- The extracted scenario builder in `cross_route_scenario.rs` is step-for-step identical to the inline version it replaces (same action ids, selectors, pending handles, topology), and `validate_cross_route_public_process_report` is the same oracle moved verbatim; the test-file wrapper now delegates.
- `FileProcessRelayControl` matches the in-process `RelayControl` semantics precisely: same error codes, same check ordering in the wait loop, same admission-order cursor, duplicates excluded via `is_success()` on both sides, occurrence-indexed visibility with fail-closed repeat hide/restore.
- File-plane mechanics are sound: single-writer O_APPEND log with partial-tail tolerance on read, `fs_private` 0700/0600 throughout, secret written before the relay container starts and scoped to the resource-lease tempdir, and `--user participant_container_user()` puts the relay on the host euid:egid so the bind mount works on Linux CI, not just macOS.
- Control-plane opt-in is fail-closed: manifest validation pins the repo-owned relay command, the orchestrator rejects control/URL map mismatches and controls-without-URLs, and preflight only claims `retained_relay_control` when the runner actually owns it.
- CI wiring matches the docs: nightly `-E` filter and weekly full ignored lane both run after the image build, and both uploads are `if: always()` with the evidence dir included. No workspace/vector version bumps.

Two non-blocking inline notes (error-code classification on io failures, and a copied-behavior comment worth keeping).

Issue #219, opened 2026-03-05
Persistent keypackage list check


Datawav:
Currently we check on login for keypackage relay list kind 10051. When a user is already logged in and doesn't have a key package, it should also check and upload a missing key package.

Issue #1436, opened 2026-08-14
marmot-app: account creation / nsec sign-in run 10-15 sequential relay round-trips on the blocking setup path; one dead relay hard-fails setup


erskingardner:
## Problem

New-account creation and nsec sign-in both funnel through the same sequential orchestrator and hold the calling UniFFI future for an estimated **~10–15 strictly sequential relay round-trips** before returning. Happy-path wall time is roughly 3–8s (create) / 5–12s (import); a single dead or slow relay in the learned/default relay set stretches this to 30–100s and, on the relay-list step, **hard-fails the whole flow and rolls the account back**.

The cost is almost entirely avoidable network serialization, not crypto (exactly one MLS KeyPackage is minted; Ed25519/Schnorr work is single-digit ms):

1. **Each of the 4–5 initial publishes builds a fresh `NostrSdkClient`** (relay lists ×2, follow list, default profile, KeyPackage), paying new TCP+TLS+WS handshakes to every relay per publish, each waiting for ≥1 relay OK.
2. **Directory/relay-list fetches connect relays one at a time** (5s connect cap each) and then run their queries sequentially (3s EOSE-class wait each).
3. After publishing the account relay lists, the code **synchronously re-fetches them from the same relays** — and an error there (e.g. one unconnectable relay) aborts account setup and deletes the account.
4. The KeyPackage cutover relay scan (kind-30443 non-current retirement) runs on first client open **even for accounts created in this process**, where it is guaranteed empty.
5. The import flow re-fetches relay-list status from relays with no freshness TTL, moments after the directory preflight already fetched overlapping data.
6. The session SQLCipher database is opened ~3 times during one setup flow (status read, one-shot client for the KeyPackage publish, then the account worker), each paying the SQLCipher KDF; one of those opens runs on the async runtime thread.

Grounded against `origin/master` at `f4e97efb873b799013028e9d648cbea60ce4468e`.

## Source-confirmed mechanism

[...]

Issue #568, opened 2026-06-10
Crash during SQLCipher salt migration permanently locks the account database (salt written before rekey, no fallback)


erskingardner:
**Severity:** high  •  **Category:** bug  •  **Confidence:** high

**Location:** `crates/marmot-app/src/lib.rs:3248`

## What
sqlcipher_salt() generates the v2 salt, persists it to disk, and only then rekeys a pre-existing legacy (v1-keyed) DB. If the process is killed after write_sqlcipher_salt() succeeds but before rekey_legacy_sqlcipher_database() completes (PRAGMA rekey re-encrypts every page, window scales with DB size), the next open finds the salt present, derives the v2 key, and SQLCipher fails to open the still-legacy-keyed DB. No code path falls back to the legacy key or retries the migration, so the account's session DB (all MLS state) becomes permanently unopenable until the .salt sidecar is manually deleted. write_sqlcipher_salt also uses plain fs::write (no temp+rename, no fsync), so a crash mid-write can leave a truncated salt that fails the length check, producing the same permanent failure on fresh DBs.

## Evidence
lib.rs:3254-3275: salt durable before rekey; error-cleanup only handles a rekey returning Err, not a crash. lib.rs:5107-5113 non-atomic write. lib.rs:5152-5162 rekey. legacy_sqlcipher_key_material referenced only at 3265; no caller tries the legacy key on v2-key open failure (callers at 3242-3246/2494-2495/3802-3808 propagate).

## Suggested fix
Make migration crash-safe: write the salt atomically (temp+fsync+rename) only AFTER the rekey commits, or on v2-key open failure attempt the legacy key and re-run the rekey. Write the salt atomically on the fresh-DB path too.

---
<sub>Found by the Fable multi-agent code review and confirmed by adversarial verification (independent refuter + reproducer, with tiebreak on disagreement). Line numbers are approximate; see quoted code for the exact site.</sub>

PR #1442, opened 2026-08-14
fix(marmot-app): skip duplicate SQLCipher KDF probe on steady-state opens

### LINKED ISSUE #1439: marmot-app/storage-sqlite: every existing-DB open pays the SQLCipher 256k PBKDF2 twice (interrupted-migration probe + real open); ~4 encrypted opens per account per cold start
## Problem

**Every open of an existing SQLCipher database pays the full key-derivation cost twice**, and a cold start performs ~4 encrypted database opens per account — so the SQLCipher KDF alone runs on the order of **6–8× per account per cold start/unlock**, before any query executes. With `cipher_compatibility = 4` and no `cipher_kdf_iterations` override anywhere in the workspace, that KDF is **PBKDF2-HMAC-SHA512 with 256,000 iterations** each time (~50–500ms per derivation depending on device class; worst on low-end Android).

The doubling: `sqlcipher_salt` unconditionally runs `finish_interrupted_sqlcipher_migration` on every existing-database open. That function's "cheap no-op" probe is in fact a **complete second keyed open** — `Connection::open` + hardened pragmas + `PRAGMA key` + the probing `SELECT count(*) FROM sqlite_master` — paying the same 256k PBKDF2 as the real open that follows. The in-code justification (self-healing the pre-fix #219 bricked state and finishing interrupted rekeys) remains valid, but on the healthy steady-state path it doubles the cost of every open.

Two further structural points:

1. **The passphrase-KDF work factor protects a key that does not need stretching on this path.** The app derives a 256-bit key via HKDF over the account secret and hands it to SQLCipher as a 64-hex-char *passphrase* string, so SQLCipher runs its anti-dictionary-crack KDF against already-uniform high-entropy key material. Skipping or shrinking the KDF for this key shape (raw `x'…'` key, or an explicit low `cipher_kdf_iterations`) preserves effective security here — **but this changes on-disk cipher parameters, so it is a security-sign-off + rekey-migration decision, not a drop-in tuning change.**
2. **Open count per cold start is ~4 encrypted opens per account**: the session DB is opened multiple times during account setup/session bring-up (see #1436), the per-account directory cache DB opens separately, `warm_directory_storage` pre-opens caches for every account at runtime start, and `account_unread_summary` performs one encrypted open per local-signing account for the account switcher on a cold cache. The 47 registered schema migrations additionally each probe their applied-state per open (~50 statements — small, but serialized on the same critical path).

Estimated user-perceived impact: a few hundred ms to ~2s per account on every cold start, unlock, and account switch — entirely CPU-bound, entirely before the UI can show data.

## Source-confirmed mechanism

- Probe on **every** existing-DB open: [`sqlcipher_salt`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/sqlcipher.rs#L133-L168) (call at L163–L165); the probe is a full keyed open at [`finish_interrupted_sqlcipher_migration#L330-L335`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/sqlcipher.rs#L330-L335).
- HKDF-derived 256-bit key hex-encoded into a passphrase: [`derive_sqlcipher_key_material`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/sqlcipher.rs#L343-L363).
- Real open pays the KDF again on its key probe: [`apply_sqlcipher_key`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/storage-sqlite/src/connection.rs#L830-L838); pinned hardening defaults (`cipher_compatibility = 4`) at [`L666-L682`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/storage-sqlite/src/connection.rs#L666-L682) and pragma ordering discipline at [`L894-L960`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/storage-sqlite/src/connection.rs#L894-L960). No `cipher_kdf_iterations` setting exists anywhere in the workspace.
- Open amplification: `warm_directory_storage` over all accounts at runtime start ([`runtime/mod.rs#L1049`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/runtime/mod.rs#L1049), [`directory/methods.rs#L50-L60`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/directory/methods.rs#L50-L60)); per-account directory cache open ([`directory/cache.rs#L40`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/directory/cache.rs#L40)); one encrypted open per local-signing account for unread badges ([`lib.rs#L2066-L2070`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/lib.rs#L2066-L2070)); repeated session-DB opens during setup (#1436).

Related: #1436 (account-setup flow opens the session DB ~3×, compounding this), #219 (the bricked-state self-heal the probe exists for — behavior must be preserved), #1304 (release-profile tuning, orthogonal).

## Required behavior

- **De-duplicate the probe on the healthy steady-state path.** A successful probe verdict ("this DB opens under the v2 key") must be reusable — at minimum within the process — so the *next* open of the same database skips the probe. Repeat probes remain mandatory whenever the migration marker exists or the verdict cannot be established (crash windows must keep the #219 self-heal). The verdict cache must never cause a genuinely legacy-keyed or interrupted database to be opened with the wrong assumption — when in doubt, probe.
- **Reduce opens per cold start/share connections per account** where lifecycle permits (single session-DB open per account per process, handed through setup/worker rather than reopened); all storage opens stay off the async runtime thread and off host calling threads.
- **Evaluate, with explicit security sign-off, wiping the passphrase-KDF overhead for this key shape**: raw hex key (`x'…'`) or a low explicit `cipher_kdf_iterations` for HKDF-derived keys, with a versioned rekey migration, rollback plan, and no change to hardening semantics (`cipher_compatibility` pin, `cipher_memory_security`, `secure_delete`, `synchronous=FULL`) otherwise. This item must not ship silently as a "perf change".
- Keep the pragma ordering discipline (compat pin first, then key, then probe) and the existing pragma-trace tests.

## Acceptance criteria

- A test (pragma-trace or instrumentation harness, extending the existing trace tests in `connection.rs`) proves a healthy v2-keyed database opened twice in one process runs the migration probe at most once, and that a marker-present database still re-runs recovery.
- The interrupted-migration and #219 bricked-state self-heal regression tests pass unchanged.
- Measured reduction in DB-open latency at cold start in the existing telemetry buckets (`AccountOpen` / `AccountSessionOpen` / `AccountSetupAdvisoryStep`), with the per-open KDF count recorded.
- The raw-key/KDF-iteration decision is documented (accept with migration plan, or reject with rationale) regardless of outcome.

## Privacy note

Per repo logging rules, keep any new tracing metadata-only (crate/module `target` + `method` + aggregate timings): no account ids, database paths, or key material.

erskingardner:
Closes #1439.

## Problem

Every open of an existing SQLCipher database paid the full key-derivation cost **twice**: `finish_interrupted_sqlcipher_migration`'s "cheap no-op" probe is a complete second keyed open (`Connection::open` + hardened pragmas + `PRAGMA key` + the `sqlite_master` probe), paying the same PBKDF2-HMAC-SHA512 / 256k iterations as the real open that follows. With ~4 encrypted opens per account per cold start, the KDF alone ran ~6–8× per account before any query executed.

## Fix

Cache the probe verdict **in-process per database file + salt** once the database has been *observed* opening under the v2 key (a successful probe, or a legacy → v2 rekey this process performed), and skip the probe on the healthy steady-state path. Safety rules from the issue are preserved exactly:

- **A durable `.salt-migrating` marker always forces the recovery probe**, verdict or not — the interrupted-migration and pre-fix mdk#568 bricked-state crash windows keep their self-heal.
- Verdicts are **salt-scoped** (a salt rotation can't be masked by a stale entry), **never recorded on the fresh-database path** (no database observed yet), **invalidated** when the database file set is removed via `remove_sqlite_file_set`, and **bounded** at 256 entries with oldest-first eviction. A miss, invalidation, or eviction only ever causes an *extra* probe — never a wrong-key assumption ("when in doubt, probe").
- Pragma ordering discipline (`cipher_compatibility` pin → key → probe) and the existing `storage-sqlite` pragma-trace tests are untouched.

Open-count reduction during account setup is deliberately out of scope here — that is #1436's acceptance criterion; the open paths already share connections per `MarmotApp` instance, and this change halves the per-open cost across all of them.

## Telemetry (per-open KDF count)

New aggregate, privacy-safe counters (counts only — no paths, accounts, or keys), wired into `AppPerformanceSnapshot` and the opt-in export mapping as unlabeled population counters:

- `app_sqlcipher_migration_probe_runs` — probes executed (each = one full keyed open paying the KDF)
- `app_sqlcipher_migration_probe_skips` — probes skipped via a cached verdict (each = one KDF avoided)

The existing `AccountOpen` / `AccountSessionOpen` / `AccountSetupAdvisoryStep` duration buckets already wrap the open paths, so the latency reduction is measurable in those buckets with the KDF count alongside.

## Raw-key / KDF-iteration decision

Documented in `docs/marmot-architecture/storage-format-v2.md`: **rejected for this change** — raw `x'…'` presentation or a reduced `cipher_kdf_iterations` alters on-disk cipher parameters and requires a versioned, crash-safe rekey migration plus explicit security sign-off; it must not ship silently as a perf change. The accept path (marker-guarded versioned rekey, unchanged hardening semantics, typed raw-key construction, tooling docs) is sketched for a future evidence-driven revisit using the new counters.

## Tests

- **Acceptance**: `sqlcipher_probe_runs_at_most_once_for_healthy_v2_database_across_opens` (probe runs once across repeated opens) and `sqlcipher_probe_reruns_whenever_migration_marker_is_present` (marker bypasses a cached verdict, then the skip resumes) — via per-path probe instrumentation in `sqlcipher.rs`.
- **Wrong-assumption guards**: verdict invalidation on file-set removal and the salt-only open path re-probe and heal a legacy-keyed database; cache salt-scoping / bound / eviction unit tests.
- **Regression**: the interrupted-migration, idempotent-recovery, and mdk#568 bricked-state self-heal tests pass **unchanged**; `storage-sqlite` pragma-trace ordering tests pass.
- Export-mapping and snapshot-wiring tests for the new counters.

## Verification

- `just fast-ci` — clean (fmt, check, clippy incl. OTLP feature builds).
- `cargo test -p marmot-app` — 656 lib tests pass; the 5 `relay_runtime` integration failures (`encrypted_media_*`, `retained_media_*`, `self_removal_*`, `..._retention_change`) **pre-exist on base `b8c4e481`** (verified in a scratch worktree against the unmodified base) and are unrelated to this change.
- `cargo test -p marmot-app --features otlp-export` — pass.
- `cargo test -p storage-sqlite connection` — 31 pass, incl. both pragma-trace tests.

## Docs

`telemetry.md` (metric catalogue), `runtime-state-bounds.md` (verdict-cache bound/reclamation inventory row), `storage-format-v2.md` (KDF decision record), `crates/marmot-app/AGENTS.md` (`sqlcipher.rs` scope mapping).

### DIFF
--- a/crates/marmot-app/AGENTS.md
+++ b/crates/marmot-app/AGENTS.md
@@ -37,7 +37,7 @@ App runtime bridge for the first real Marmot app surfaces.
   validation/resolution/removal, and HTTP upload. Audit-log unit tests live in its own `#[cfg(test)] mod tests`.
 - Keep SQLCipher key derivation, per-database salt persistence, legacy-key migration/recovery, and the in-process
   v2-open probe-verdict cache (mdk#1439) in `src/sqlcipher.rs`. The verdict cache is advisory only: a durable
-  migration marker or a missing verdict always re-runs the recovery probe, so the #219 crash windows keep their
+  migration marker or a missing verdict always re-runs the recovery probe, so the mdk#568 crash windows keep their
   self-heal. Key presentation and the passphrase KDF work factor follow the decision record in
   `docs/marmot-architecture/storage-format-v2.md`.
 - Keep the user-directory domain in the `src/directory/` module instead of regrowing `src/lib.rs`. It splits along these
--- a/crates/marmot-app/AGENTS.md
+++ b/crates/marmot-app/AGENTS.md
@@ -35,6 +35,11 @@ App runtime bridge for the first real Marmot app surfaces.
 - Keep the forensic audit-log feature in `src/audit_log.rs`: the `AuditLog*` DTOs, salted-hash identity derivation,
   the upload client, and the `MarmotApp` methods for audit settings, recorder open/build, file enumeration, path
   validation/resolution/removal, and HTTP upload. Audit-log unit tests live in its own `#[cfg(test)] mod tests`.
+- Keep SQLCipher key derivation, per-database salt persistence, legacy-key migration/recovery, and the in-process
+  v2-open probe-verdict cache (mdk#1439) in `src/sqlcipher.rs`. The verdict cache is advisory only: a durable
+  migration marker or a missing verdict always re-runs the recovery probe, so the #219 crash windows keep their
+  self-heal. Key presentation and the passphrase KDF work factor follow the decision record in
+  `docs/marmot-architecture/storage-format-v2.md`.
 - Keep the user-directory domain in the `src/directory/` module instead of regrowing `src/lib.rs`. It splits along these
   seams: `records.rs` (the public `UserDirectory*`/`UserProfileMetadata`/`DirectoryKeyPackage` DTOs surfaced to
   `marmot-uniffi`/`cli`, plus the stateless record helpers — cached <-> shared record conversion, recency selection,
--- a/crates/marmot-app/src/app_telemetry.rs
+++ b/crates/marmot-app/src/app_telemetry.rs
@@ -109,6 +109,17 @@ pub struct AppPerformanceSnapshot {
     pub account_catch_up: AppPerformanceOperationSnapshot,
     pub account_sync: AppPerformanceOperationSnapshot,
     pub account_setup_advisory_step: AppPerformanceOperationSnapshot,
+    /// Interrupted-migration recovery probes executed since process start. Each
+    /// probe is a full keyed SQLCipher open paying the passphrase KDF; the
+    /// healthy steady state skips it via the cached v2-open verdict (mdk#1439).
+    /// Process-wide aggregate: no account, path, or key information.
+    #[serde(default)]
+    pub sqlcipher_migration_probe_runs: u64,
+    /// Existing-database opens that skipped the recovery probe via a cached
+    /// v2-open verdict since process start. Each skip avoided one full
+    /// passphrase KDF derivation (mdk#1439).
+    #[serde(default)]
+    pub sqlcipher_migration_probe_skips: u64,
     pub outbound_message_send: AppPerformanceOperationSnapshot,
     #[serde(default)]
     pub group_create_queue_wait: AppPerformanceOperationSnapshot,
@@ -453,6 +464,8 @@ impl AppPerformanceTelemetry {
             .inner
             .lock()
             .unwrap_or_else(|poisoned| poisoned.into_inner());
+        let (sqlcipher_migration_probe_runs, sqlcipher_migration_probe_skips) =
+            crate::sqlcipher::sqlcipher_migration_probe_counters();
         AppPerformanceSnapshot {
             app_start: inner.app_start.snapshot(),
             directory_subscription_sync: inner.directory_subscription_sync.snapshot(),
@@ -467,6 +480,8 @@ impl AppPerformanceTelemetry {
             account_catch_up: inner.account_catch_up.snapshot(),
             account_sync: inner.account_sync.snapshot(),
             account_setup_advisory_step: inner.account_setup_advisory_step.snapshot(),
+            sqlcipher_migration_probe_runs,
+            sqlcipher_migration_probe_skips,
             outbound_message_send: inner.outbound_message_send.snapshot(),
             group_create_queue_wait: inner.group_create_queue_wait.snapshot(),
             group_create_key_package_lookup: inner.group_create_key_package_lookup.snapshot(),
@@ -726,5 +741,21 @@ mod tests {
         assert_eq!(snapshot.account_setup_advisory_step.attempts, 1);
         assert_eq!(snapshot.account_setup_advisory_step.successes, 0);
         assert_eq!(snapshot.account_setup_advisory_step.failures, 1);
+    }
+
+    #[test]
+    fn snapshot_carries_the_process_wide_sqlcipher_probe_counters() {
+        // The counters are process-global atomics shared with every test in
+        // this process, so bracket the snapshot between two direct reads
+        // instead of asserting exact values.
+        let telemetry = AppPerformanceTelemetry::default();
+        let before = crate::sqlcipher::sqlcipher_migration_probe_counters();
+        let snapshot = telemetry.snapshot();
+        let after = crate::sqlcipher::sqlcipher_migration_probe_counters();
+
+        assert!(before.0 <= snapshot.sqlcipher_migration_probe_runs);
+        assert!(snapshot.sqlcipher_migration_probe_runs <= after.0);
+        assert!(before.1 <= snapshot.sqlcipher_migration_probe_skips);
+        assert!(snapshot.sqlcipher_migration_probe_skips <= after.1);
     }
 }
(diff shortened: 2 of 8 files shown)


crates/marmot-app/AGENTS.md
```diff
@@ -35,6 +35,11 @@ App runtime bridge for the first real Marmot app surfaces.
 - Keep the forensic audit-log feature in `src/audit_log.rs`: the `AuditLog*` DTOs, salted-hash identity derivation,
   the upload client, and the `MarmotApp` methods for audit settings, recorder open/build, file enumeration, path
   validation/resolution/removal, and HTTP upload. Audit-log unit tests live in its own `#[cfg(test)] mod tests`.
+- Keep SQLCipher key derivation, per-database salt persistence, legacy-key migration/recovery, and the in-process
+  v2-open probe-verdict cache (mdk#1439) in `src/sqlcipher.rs`. The verdict cache is advisory only: a durable
+  migration marker or a missing verdict always re-runs the recovery probe, so the #219 crash windows keep their
```
erskingardner: Non-blocking: please use the canonical incident reference here and in the other new SQLCipher comments/tests. Unqualified `#219` resolves to marmot-protocol/mdk#219 (“Persistent keypackage list check”), which is unrelated. The crash-window issue is marmot-protocol/mdk#568; its fix landed via marmot-protocol/darkmatter#269. Using `mdk#568` or a fully qualified link will keep future readers on the correct history.

erskingardner (commented):
Recommend merge. I found no correctness or safety issue in the cache path: a durable marker still forces recovery; cache misses, salt changes, and bounded eviction fall back to probing; fresh databases do not establish a verdict; and repository-owned file-set removal invalidates the entry. One non-blocking provenance correction is noted inline. Completed CI checks are green; the remaining jobs were still running at review time.

crates/marmot-app/AGENTS.md
erskingardner: Fixed in 887acb7c — all seven unqualified `#219` references across `sqlcipher.rs` and `AGENTS.md` now say `mdk#568` (including the two pre-existing ones in the touched comment blocks, so the file stays consistent). PR body updated to match.

Issue #1303, opened 2026-08-07
Expose AppPerformanceSnapshot over UniFFI; add group-create, accept-invite, and combined roster-read phases


Datawav:
Expose AppPerformanceSnapshot over UniFFI; add group-create, accept-invite, and combined roster-read phases

## Summary

MDK has a solid per-phase performance telemetry system (`AppPerformanceOperation`, `crates/marmot-app/src/app_telemetry.rs:20-42` with duration histograms per operation, and privacy-safe OTLP export in `relay_telemetry_export.rs`). Two gaps keep it from closing the measurement loop on the group-management latency work (#466, #1162, and the client-side issues filed 2026-08-07):

1. **`AppPerformanceSnapshot` is not reachable over UniFFI.** Nothing in `crates/marmot-uniffi/src/` exposes it, so a client app cannot read per-device, per-phase timings — the data only leaves via the population-level OTLP export. A debug/diagnostics surface in the app (or an on-demand support dump) needs a read-only getter.
2. **Phase coverage is missing for the remaining audited flows.** Invite has seven sub-phases (`GroupInviteKeyPackageLookup` … `GroupInvitePostMutationCatchUp`), and `GroupPromoteAdmin`, `GroupDetailsRead`, `GroupMlsStateRead` exist — but **group create** (tracked in #1162) has no phases, **accept-invite** has none, and there is no single metric bounding the `groupMlsState`+`groupDetails` roster-refresh pair clients perform (marmot-protocol/whitenoise-android#1847).

## Suggested shape

- UniFFI: read-only `app_performance_snapshot()` (or equivalent) on the runtime object, returning the existing snapshot struct; no new collection, privacy posture unchanged (aggregate durations only).
- New phases mirroring the invite set: `GroupCreateKeyPackageLookup / GroupCreateRoutingRefresh / GroupCreateEnginePublish / GroupCreatePostSync` on the `create_group` path (`crates/marmot-app/src/client/mod.rs:228-284`), `GroupAcceptInvite`, and a `GroupRosterRefresh` phase (or a documented way to bracket the read pair).
- Export batch mapping gains the new points via the existing `append_app_performance_points` path.

## Acceptance criteria

- A Kotlin/Swift client can fetch the snapshot and see invite/create/accept/roster phases populated after performing those flows.
- OTLP export includes the new phases with no account/group/relay labels (existing privacy tests extended).

Labels: `enhancement`, `MEDIUM`, `performance`, `Bindings Related`

PR #1444, opened 2026-08-14
Start invite history catch-up immediately

### LINKED ISSUE #1438: marmot-app: accepted invite shows an empty group for ~15s+ (post-join catch-up waits for the 15s maintenance tick); AcceptGroupInvite queues behind worker startup with no response timeout
## Problem

Tapping **accept** on an invite returns quickly, but the group then appears **empty for ~15+ seconds** from the user's perspective: group history flow is only kicked off by the account worker's **15-second maintenance tick**, followed by an EOSE wait, a grace window, a quiet window, and a 0–30s sampled contention jitter before the catch-up machinery fully settles. Nothing about that schedule starts when the invite is joined or accepted — so every accepted invite is followed by a long stare at an empty conversation.

Two more defects compound the felt slowness:

1. `AcceptGroupInvite` is a **pure local confirmation flip** (no network, no crypto), yet it queues onto the single per-account worker as a catch-all and is **deferred behind the startup hydration pipeline and initial catch-up sync** — a post-launch/foreground accept can wait seconds behind unrelated catch-up work. The caller awaits the worker's reply oneshot with **no timeout**, so a busy worker presents to the app as a hung tap.
2. On the inbound side, invite *visibility* is gated on an **awaited group-subscription registration** (`sync_runtime_groups`) before the worker publishes the `GroupJoined` summary — paying a warm-relay RTT, or a multi-second cold connect to relays this device has never used, before the user even sees the invite.

## Source-confirmed mechanism

- Accept handler is local-only (confirmation flip + one `save_state` transaction): [`account_worker.rs#L2046`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/runtime/account_worker.rs#L2046), but startup defers "mutations and catch-ups … [into] the same startup deferral": [`account_worker.rs#L504-L510`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/runtime/account_worker.rs#L504-L510).
- No timeout on the reply: [`account_worker_response`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/runtime/mod.rs#L5022-L5026) — plain `oneshot` await.
- Inbound ingest awaits subscription registration before the joined-group summary is published: [`client/sync.rs#L683`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/client/sync.rs#L683) (`sync_runtime_groups().await?` inside `ingest_received_delivery`).
- Post-join maintenance obligation is durably written at join with a sampled jitter: [`group_lifecycle.rs#L1187-L1193`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/cgka-engine/src/group_lifecycle.rs#L1187-L1193) (jitter sampled at [L945](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/cgka-engine/src/group_lifecycle.rs#L945); helper at [L52-L106](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/cgka-engine/src/group_lifecycle.rs#L52-L106)), with `POST_JOIN_CONTENTION_JITTER_MAX_MS = 30_000` in [`traits/src/maintenance.rs#L16`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/traits/src/maintenance.rs#L16).
- …but it is only advanced from the worker's **15s tick** ([`account_worker.rs#L767`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/runtime/account_worker.rs#L767); the `advance_post_join_maintenance_subscriptions` call at [`L1185`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-app/src/runtime/account_worker.rs#L1185)), which installs the full-history `GroupMaintenance` subscription ([`sdk_client.rs#L372`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/transport-nostr-adapter/src/sdk_client.rs#L372)) and then waits EOSE ([`mark_post_join_eose`](https://github.com/marmot-protocol/mdk/blob/f4e97efb873b799013028e9d648cbea60ce4468e/crates/marmot-account/src/runtime.rs#L1031)) plus grace/quiet windows.

Related: #1176 (foreground send preflight), #741 / #785 (worker head-of-line blocking; same queue, different commands), #720 (inbound welcome routing scan — minor here), #1303 (accept-invite perf phases over UniFFI would validate this on device telemetry).

## Required behavior

- Starting the post-join history catch-up must be an immediate consequence of the join (or of accept), not of the next maintenance tick. The contention jitter exists to decorrelate post-join **self-update commits** across many new members; it should not delay when history starts flowing.
- `AcceptGroupInvite` must not sit behind the startup hydration/initial-sync window for seconds: either serve it during the deferral window like the read commands (it needs only the already-persisted projection row) or answer with a bounded wait plus a typed busy/timeout error the app can render.
- Invite visibility must not wait on relay I/O: publish the `GroupJoined`/chat-list update as soon as the projection is durable, and register the new group subscription in the background.
- Any timeout added to the worker reply path must apply uniformly to mutation commands, not just accept.
- Preserve durability guarantees (projection writes before visibility) and join-time validation/ordering invariants; this issue changes *scheduling*, not semantics.

## Acceptance criteria

- Deterministic test: after a welcome is joined (or an invite accepted) on a running worker, the post-join history subscription is installed **without waiting for the 15s tick** ( observable via injected clock / maintenance-plan assertions), and the self-update commit retains its jitter protection.
- Deterministic test: `AcceptGroupInvite` issued during the startup hydration window completes within a documented bound, or fails fast with a typed status — never an indefinite hang; a host-visible deadline now wraps the worker reply path.
- Ordering test: the `GroupJoined` summary/chat-list emission precedes `sync_runtime_groups` completion for the live-ingest path, with the subscription registration retried in the background on failure.
- The 15s empty-group gap is gone on device, validated via the accept-invite perf phases from #1303 before/after.

## Privacy note

Per repo logging rules, keep any new tracing metadata-only (crate/module `target` + `method` + aggregate timings): no account ids, group ids, relay URLs, or event ids.

erskingardner:
## Summary

- publish a durable `GroupJoined`/chat-list update before ordinary group-subscription relay I/O, then retry that subscription rebuild in the background with bounded backoff
- start the post-join full-history subscription immediately after join visibility instead of waiting for the 15-second maintenance tick, while preserving the existing grace, quiet, and contention-jitter rules for self-update commits
- execute projection-only invite acceptance during startup hydration and return typed `AccountWorkerBusy` when exclusive catch-up owns the account session, so callers know the operation did not start
- add operation-class worker response deadlines and a typed `AccountWorkerResponseTimedOut` result whose completion is explicitly unknown
- expose the new error semantics through UniFFI, CLI JSON, and the conformance simulator

## Root cause

Live welcome ingest awaited `sync_runtime_groups` before publishing the joined-group summary, so relay connection or subscription latency delayed invite visibility. The post-join history obligation was then advanced only by the periodic maintenance tick. At the same time, invite acceptance used the general account-worker queue, which deferred it behind startup hydration and catch-up, and worker reply oneshots had no deadline.

## User impact

Accepted invites become visible as soon as their projection is durable, and history subscription starts immediately rather than after the next maintenance tick. Accept taps either complete promptly, return a definitely-not-started busy result that is safe to retry, or return an explicit response-timeout result that requires refreshing authoritative state before retrying.

## Validation

- `just fast-ci`
- `cargo test -p marmot-app --lib` — 659 passed
- focused joined-group visibility/background retry/busy regression
- focused startup-hydration accept regression
- focused bounded retry and worker-response deadline tests
- post-join rotation jitter regression
- `cargo test -p marmot-uniffi --lib` — 106 passed
- focused CLI typed-error JSON regression

A full `marmot-app` crate run is not reported as green because five unrelated `relay_runtime` tests also fail on untouched `origin/master` at `f7ddd5a9` (90 passed, 5 failed): the two encrypted-media tests, retention system-row synthesis, retained-media pruning, and self-removal unread projection. The #1438-affected re-add regression passes with the new busy-aware retry helper.

On-device before/after perf-phase measurement from #1303 remains pending; deterministic scheduling coverage is included here.

Fixes #1438

## Summary by CodeRabbit

* **New Features**
  * Added durable invite joining and startup invite acceptance.
  * Added bounded retries when invite acceptance encounters a busy worker.
  * Added operation-specific response deadlines.
  * Added background retries for subscription updates after new deliveries.
  * Added distinct busy-worker and timed-out-response reporting, including retry guidance and completion uncertainty.
  * Preserved these error distinctions across JSON and MarmotKit interfaces.

* **Bug Fixes**
  * Improved group history availability before subscription rebuilding completes.
  * Prevented committed deliveries from being reported as failed when background updates are delayed.

### DIFF
--- a/crates/marmot-app/src/client/sync.rs
+++ b/crates/marmot-app/src/client/sync.rs
@@ -334,6 +334,9 @@ impl AppClient {
         self.prepare_transport_with_telemetry(telemetry)
             .await
             .map_err(SyncFailure::from)?;
+        // A complete startup/catch-up rebuild satisfies any older deferred
+        // refresh intent before this pass starts ingesting new deliveries.
+        self.pending_runtime_group_subscription_refresh = false;
         // Both the inbox/group activation and the group-subscription refresh
         // have now registered on relays; emit the rebuild audit row from the
         // drained registration log before draining inbound deliveries.
@@ -909,9 +912,17 @@ impl AppClient {
         summary.merge(std::mem::take(&mut self.pending_failed_sync_summary));
 
         if routes_dirty || routes_changed {
-            self.sync_runtime_groups()
-                .await
-                .map_err(SyncCheckpointError::AfterPersistence)?;
+            match self.sync_runtime_groups().await {
+                Ok(()) => self.pending_runtime_group_subscription_refresh = false,
+                Err(error) => {
+                    // The projection and route checkpoint above are durable.
+                    // Retain an explicit retry edge so the worker repairs the
+                    // ordinary subscriptions without replaying this prefix or
+                    // waiting for another catch-up trigger.
+                    self.pending_runtime_group_subscription_refresh = true;
+                    return Err(SyncCheckpointError::AfterPersistence(error));
+                }
+            }
         }
         Ok(())
     }
@@ -1508,6 +1519,7 @@ impl AppClient {
         self.sync_runtime_groups()
             .await
             .map_err(SyncFailure::from)?;
+        self.pending_runtime_group_subscription_refresh = false;
         self.record_subscription_rebuild(None).await;
         let mut deliveries = 0;
         let mut summary = self.sync_sdk_relay(&mut deliveries).await?;
@@ -2321,6 +2333,57 @@ mod membership_change_tests {
         let admin = GroupStateChange::AdminAdded { member };
         assert!(member_departure(&added).is_none());
         assert!(member_departure(&admin).is_none());
+    }
+}
+
+#[cfg(test)]
+mod runtime_group_subscription_refresh_tests {
+    use std::sync::Arc;
+
+    use super::{SyncCheckpointError, SyncSummary};
+    use crate::tests::ScriptedPushRelayClient;
+    use crate::{AppPerformanceTelemetry, MarmotApp};
+    use marmot_account::AccountHome;
+
+    #[tokio::test]
+    async fn catch_up_checkpoint_arms_refresh_after_durable_subscription_failure() {
+        let dir = tempfile::tempdir().unwrap();
+        AccountHome::open(dir.path())
+            .create_account("alice")
+            .unwrap();
+        let relay = Arc::new(ScriptedPushRelayClient::default());
+        let app = MarmotApp::with_relay(dir.path(), "wss://relay.example")
+            .with_test_relay_client(relay.clone());
+        let mut client = app.client("alice").await.unwrap();
+        client.prepare_transport().await.unwrap();
+        let telemetry = AppPerformanceTelemetry::default();
+        client
+            .create_group_with_initial_profile_and_telemetry(
+                "catch-up retry intent",
+                "",
+                &[],
+                None,
+                &telemetry,
+            )
+            .await
+            .unwrap();
+
+        relay.fail_next_subscribe();
+        let mut summary = SyncSummary::default();
+        let error = client
+            .checkpoint_sync_prefix(&mut summary, true, 0)
+            .await
+            .expect_err("the injected post-checkpoint subscription rebuild must fail");
+        assert!(matches!(error, SyncCheckpointError::AfterPersistence(_)));
+        assert!(client.has_pending_runtime_group_subscription_refresh());
+
+        assert!(
+            !client
+                .retry_pending_runtime_group_subscription_refresh()
+                .await
+                .unwrap()
+        );
+        assert!(!client.has_pending_runtime_group_subscription_refresh());
     }
 }
 
--- a/crates/marmot-app/src/client/sync.rs
+++ b/crates/marmot-app/src/client/sync.rs
@@ -199,6 +199,36 @@ impl AppClient {
         Ok(())
     }
 
+    pub(crate) fn has_pending_runtime_group_subscription_refresh(&self) -> bool {
+        self.pending_runtime_group_subscription_refresh
+    }
+
+    /// Retry an ordinary group-subscription rebuild that was deliberately
+    /// moved behind live-ingest visibility. A successful rebuild disarms the
+    /// intent; an error leaves it armed so the worker's bounded backoff can
+    /// try again without replaying the durable delivery.
+    pub(crate) async fn retry_pending_runtime_group_subscription_refresh(
+        &mut self,
+    ) -> Result<bool, AppError> {
+        if !self.pending_runtime_group_subscription_refresh {
+            return Ok(false);
+        }
+        if let Err(error) = self.sync_runtime_groups().await {
+            if error.is_account_not_active() {
+                // A relay notification gap or overlapping account-adapter
+                // teardown can retire the activation between durable ingest
+                // and this background retry. Re-activation installs both the
+                // account inbox and the current complete group set, satisfying
+                // the same refresh intent without replaying the delivery.
+                self.prepare_transport().await?;
+            } else {
+                return Err(error);
+            }
+        }
+        self.pending_runtime_group_subscription_refresh = false;
+        Ok(false)
+    }
+
     pub(crate) async fn prepare_transport(&self) -> Result<(), AppError> {
         self.prepare_transport_with_telemetry(None).await
     }
@@ -608,6 +638,13 @@ impl AppClient {
         loop {
             let delivery = self.receive_next_delivery().await?;
             let summary = self.ingest_received_delivery(delivery).await?;
+            // A directly-owned AppClient has no account-worker scheduler to
+            // perform the post-visibility retry. Preserve its historical
+            // contract by completing the pending rebuild before handing the
+            // summary to its caller; the managed worker uses the lower-level
+            // ingest method and owns the background retry instead.
+            self.retry_pending_runtime_group_subscription_refresh()
+                .await?;
             if summary.joined_groups.is_empty()
                 && summary.messages.is_empty()
                 && summary.events.is_empty()
@@ -684,9 +721,7 @@ impl AppClient {
         if !routes_dirty || refresh.state_pruned {
             self.save_state_with_pending_local_group_deletion_frontier_clears()?;
         }
-        if routes_dirty || refresh.routing_changed {
-            self.sync_runtime_groups().await?;
-        }
+        self.pending_runtime_group_subscription_refresh |= routes_dirty || refresh.routing_changed;
         self.drain_epoch_stall_escalations(&mut summary);
         Ok(summary)
     }
--- a/crates/marmot-app/src/runtime/account_worker.rs
+++ b/crates/marmot-app/src/runtime/account_worker.rs
@@ -348,6 +348,9 @@ pub(crate) enum AccountWorkerCommand {
     RetryPushRegistration {
         respond: oneshot::Sender<bool>,
     },
+    RetryRuntimeGroupSubscriptions {
+        respond: oneshot::Sender<bool>,
+    },
     DeleteAuditLog {
         path: std::path::PathBuf,
         respond: oneshot::Sender<Result<bool, AppError>>,
@@ -470,14 +473,17 @@ async fn run_app_runtime_account_worker(
         scheduled_convergence_test_delay(&app),
     );
     let mut scheduled_push_retry = ScheduledPushRegistrationRetry::new();
+    let mut scheduled_runtime_group_subscription_refresh =
+        ScheduledRuntimeGroupSubscriptionRefresh::new();
 
     // The session's cheap open pass has seeded every stored group. Signal
     // command-readiness *now*: the hydration pipeline right below enters its
     // command-serving loop immediately, so "ready" genuinely means "serving
     // commands" — group reads (`Members` / `MemberIdsPage` / `GroupMlsState` /
     // `GroupRoster` / `QuarantinedGroups`) issued from this point hydrate the
-    // group(s) they name and answer live. Everything else
-    // joins the startup deferral. `AccountOpen` (recorded by `reconcile` as the ready-wait)
+    // group(s) they name and answer live; projection-only invite acceptance
+    // also runs immediately. Everything else joins the startup deferral.
+    // `AccountOpen` (recorded by `reconcile` as the ready-wait)
     // measures the seeded open; the mdk#1161 stage telemetry attributes it
     // (`AccountSessionOpen` / `AccountGroupHydration` for the open the
     // worker just awaited, with the pipeline and snapshot capture measured
@@ -505,8 +511,9 @@ async fn run_app_runtime_account_worker(
     // seeded stored groups, so fully hydrate them now — chat-list recency
     // first — while serving commands. Group reads for a not-yet-hydrated
     // group hydrate that one group and answer live ("waits for that group
-    // only"); mutations and catch-ups join the same startup deferral the
-    // catch-up window has always used and replay in arrival order after it.
+    // only"); projection-only invite acceptance also runs live. Other
+    // mutations and catch-ups join the same startup deferral the catch-up
+    // window has always used and replay in arrival order after it.
     let mut deferred: Vec<DeferredStartupCommand> = Vec::new();
     match run_startup_hydration_pipeline(
         &app,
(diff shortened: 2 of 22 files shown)


crates/marmot-app/src/runtime/commands.rs
```diff
@@ -146,10 +147,7 @@ impl AccountManager {
             .send(AccountWorkerCommand::RepairFullHistory { respond })
             .await
             .map_err(|_| AppError::TransportClosed)?;
-        response
-            .await
-            .map_err(|_| AppError::TransportClosed)?
-            .map_err(AppError::AccountCatchUp)
+        account_worker_catch_up_response(response).await
```
erskingardner: `repair_full_history` now gets `APP_RUNTIME_WORKER_RESPONSE_WAIT` (2 min). The constant's own doc says the default class is for commands that "may sign, publish, or perform a bounded relay exchange" — a forced complete relay-history query that projects every returned event is the opposite of a bounded exchange, and on an older/large account two minutes is an arbitrary new failure point (this oneshot was awaited with no deadline before this PR). Media commands got the 16-minute class for exactly this long-tailed shape; full-history repair seems to belong there too. If the 2-minute cap is deliberate, a comment with the justification would help; otherwise I'd reclassify.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1707,6 +1833,22 @@ async fn handle_startup_hydration_command(
...
+            let result = client.accept_group_invite(&group_id);
+            if result.is_ok() {
+                publish_app_runtime_group_state_updated(
+                    events,
+                    account_id_hex,
+                    account_label,
+                    &group_id,
+                );
+            }
+            let _ = respond.send(result);
```
erskingardner, line 1863: The live `AcceptGroupInvite` handler (around L2313) runs `client.retry_pending_push_registration_shares_best_effort()` on the post-response seam after a successful accept, but this hydration fast path skips it, so an accept accepted during startup hydration loses the push-share retry until some unrelated command happens to run it. Either add the same hook here (after `respond.send`) or extract one small shared accept handler so the two paths can't drift apart.

crates/marmot-app/src/client/sync.rs
```diff
@@ -684,9 +721,7 @@ impl AppClient {
         if !routes_dirty || refresh.state_pruned {
             self.save_state_with_pending_local_group_deletion_frontier_clears()?;
         }
-        if routes_dirty || refresh.routing_changed {
-            self.sync_runtime_groups().await?;
-        }
+        self.pending_runtime_group_subscription_refresh |= routes_dirty || refresh.routing_changed;
```
erskingardner, line 727: Arming here covers only live ingest. `checkpoint_sync_prefix` (just below, ~L911) still rebuilds `sync_runtime_groups` inline after its durable checkpoint commit and reports `AfterPersistence` without ever setting `pending_runtime_group_subscription_refresh`, and neither the startup catch-up completion nor `handle_account_worker_catch_up` calls `scheduled_runtime_group_subscription_refresh.observe_pending(...)` afterwards (the only call site is the live receive arm). So a failed rebuild on the catch-up path relies on the next catch-up/tick while the live path gets the bounded background retry — consistent with pre-PR behavior, so not a regression, but it means the new "rebuild failures retry in the background" contract holds for live ingest only. Either arm the flag on this path too (and observe it after catch-up completions), or add a short comment stating the next catch-up owns the retry.

erskingardner (commented):
## Review (d0ffa4e) — requesting changes

*(Posted as a comment because GitHub won't let the PR author formally request changes on their own PR; items 1 and 2 are blockers as far as I'm concerned.)*

The premise and shape of this match #1438 well: publishing `GroupJoined` once the projection is durable, starting post-join history right after visibility, a typed definitely-not-started `AccountWorkerBusy`, and a uniform bounded response await are the right contract changes, and the test scaffolding (`ScriptedPushRelayClient` group-subscribe discrimination, startup-hydration accept regression, deadline unit test) is solid. The scheduling-vs-semantics boundary is respected, and the new logging stays privacy-safe.

Blocking items first.

### 1. CI is red, and both failures are caused by this PR

**`Rust tests (1/4)`** — `agent-connector tests::connector_socket_subscribes_to_inbound_messages` failed both attempts:

```
called `Result::unwrap()` on an `Err` value: AccountWorkerBusy
  at crates/agent-connector/src/tests.rs:1023
```

That test taps `accept_group_invite` immediately after `create_identity`/`create_group`, i.e. exactly inside the window where the worker can be in its exclusive startup catch-up. The new busy contract is doing its job; the in-repo caller just wasn't migrated. `relay_runtime.rs` got `accept_group_invite_retrying_busy`, but agent-connector did not. This is a preview of what every MarmotKit host will hit on upgrade, so this PR should migrate all in-repo accept call sites deliberately: the connector test (same retry helper), and an explicit decision for `agent-connector/src/invite_policy.rs` (its retry loop presumably covers busy — worth a comment saying so), `node_protocol.rs:accept_active_invite`, and the simulator's `accept_pending_invites`.

**`Simulator and vectors`** — `process_orchestrator::process_kill_disconnect_reconnect_and_restart_agree_with_uninterrupted_execution` failed both attempts with the same failure-capsule digest (`33e1eaad60911faf`, so deterministic), and on the retry even the *uninterrupted* run failed. In both reports the run aborted at `step-0:create_group@main` with `status: Failed` and `participants: []`. Empty participants means the failure is orchestrator-level, not a node rejection: either the node's create response exceeded `NODE_RESPONSE_TIMEOUT` (45s) or the resulting publications didn't reach the shared relay within `RELAY_ACTION_PUBLICATION_TIMEOUT` (30s — try 2's 35.5s wall time fits this). The same two jobs are green on the latest master CI run (31804445914, an hour before this PR's run). This needs the failure capsule decoded and fixed before merge; if it turns out to be startup-latency amplification from the new immediate post-join work, that is itself worth understanding before landing.

The PR body's Validation section doesn't mention either red check — the five `relay_runtime` failures it describes are a different job, which is green here.

### 2. `repair_full_history` response deadline is misclassified

Inline at `commands.rs:150`: a forced complete relay-history scan is not a "bounded relay exchange"; the 2-minute default class introduces an arbitrary failure point that did not exist (unbounded await) before. Reclassify to the long class or justify the bound.

### 3. Catch-up path doesn't arm the deferred subscription refresh

Agreeing with CodeRabbit's first finding, sharpened inline at `sync.rs:915`: after `checkpoint_sync_prefix` commits durable state, a failed final `sync_runtime_groups` leaves no `pending_runtime_group_subscription_refresh` intent, so the new bounded background retry never covers the catch-up path. Not a regression vs master (the next catch-up rebuilds as before), but the PR's contract — "rebuild failures are retried in the background" — now holds for live ingest only. Either arm + observe the flag after catch-up completions, or document that the next catch-up owns the retry.

### 4. Hydration accept fast-path diverges from the live handler

Inline at `account_worker.rs:1850`: the live handler runs `retry_pending_push_registration_shares_best_effort()` after a successful accept; the new hydration fast path doesn't. Small drift, worth unifying.

### Non-blocking notes

- CodeRabbit's second finding (test discriminator) looks stale to me: `block_and_fail_account_group_subscribe` arms a blocker that matches only the ordinary `NostrSubscription::Group` variant, and the post-join full-history subscription is the distinct `NostrSubscription::GroupMaintenance` variant, so the test does isolate the ordinary rebuild. No action needed.
- The uncapped-attempt retry task will publish an account error per minute forever under a permanently degraded relay. Consistent with the existing maintenance cadence, but consider quieting repeats.
- The #1438 on-device before/after measurement (#1303) is still pending per the PR body — fine as follow-up, but worth tracking before calling the issue fully closed.

Once CI is green and items 2–4 are addressed, this is in good shape.

erskingardner (commented):
## Re-review (045f534) — recommend merge

*(GitHub won't let the PR author approve their own PR, so this is a comment; consider this my sign-off.)*

CI is fully green at the new head, and every finding from my previous review (d0ffa4e) is addressed:

1. **agent-connector busy regression** — fixed. `connector_socket_subscribes_to_inbound_messages` now uses a busy-retry helper, and `invite_policy.rs` documents the deliberate propagate-and-retry semantics for `AccountWorkerBusy`. The simulator adapters got the same treatment via a shared `accept_group_invite_retrying_busy`, and the CLI one-shot accept now retries busy for a bounded 30s window while correctly treating `AccountWorkerResponseTimedOut` as terminal (never retrying ambiguous completion).
2. **Simulator process-orchestrator failure** — fixed. Serializing the subprocess-heavy `process_orchestrator` binary (`nextest.toml` test-group with `max-threads = 1`) plus the raised, documented `CATCH_UP_DEADLINE` addresses the relay-publication starvation that was tripping the action-correlation timeout. The rationale comments are honest about the mechanism. Worth noting for #1303: this also confirms the new immediate post-join work measurably increases per-node startup cost under load — good data point for the pending on-device measurement.
3. **`repair_full_history` classification** — fixed. Now uses `long_account_worker_catch_up_response` (16-minute class), matching its unbounded-relay-scan shape.
4. **Catch-up pending-flag gap** — fixed, and better than I asked: `checkpoint_sync_prefix` arms the flag on `AfterPersistence` failure, complete rebuilds clear a stale intent, and `observe_pending` now runs at the post-startup, scheduled-convergence, and 15s maintenance-tick seams — so an armed intent is picked up regardless of which path set it. The new `catch_up_checkpoint_arms_refresh_after_durable_subscription_failure` regression test pins the contract.
5. **Hydration accept parity** — fixed. The hydration fast path now mirrors the live handler's post-response `retry_pending_push_registration_shares_best_effort()`.

The cursor-persistence test updates are a nice catch in their own right: `GroupJoined` no longer implying subscription-rebuild completion is exactly the new contract, and the staged panic messages will make the next CI failure much easier to diagnose.

Non-blocking leftovers (no action required for merge):
- CodeRabbit's remaining ask for a deterministic retry clock in the simulator's busy-retry helper is largely mitigated by the fixed 500-attempt budget; wall-clock waits are already the norm in the process/app adapters (e.g. the 30s relay-action timeout), so I'd leave this as-is.
- The uncapped background retry task will still publish an account error per minute under a permanently degraded relay — fine to revisit later if hosts complain.

LGTM.

crates/marmot-app/src/tests.rs
```diff
@@ -1663,6 +1741,122 @@ async fn invite_members_detaches_post_mutation_catch_up_body() {
...
+        let relay = Arc::new(ScriptedPushRelayClient::default());
+        let app = MarmotApp::with_relay(dir.path(), "wss://relay.example")
+            .with_test_relay_client(relay.clone());
+        let runtime = MarmotAppRuntime::new(app.clone());
+        runtime.reconcile_accounts().await.unwrap();
+        runtime.catch_up_accounts().await.unwrap();
+        let mut events = runtime.subscribe();
+
+        // The first ordinary group-subscription rebuild will park and fail.
+        // The distinct post-join full-history subscription remains available,
```
erskingardner, line 1761: The comment says the post-join full-history subscription remains available, but nothing in this test observes it.

#1438's first acceptance criterion is: after a welcome is joined on a running worker, the `GroupMaintenance` subscription is installed without waiting for the 15s tick (and the self-update commit keeps its jitter).

`block_and_fail_account_group_subscribe` only parks `NostrSubscription::Group`, so this fixture can already assert that a `GroupMaintenance` subscription for `group_id` appears in `subscription_attempts` before any 15s advance. Without that check, deleting every `start_post_join_history_after_visibility` call would still leave this test green — `GroupJoined` and the ordinary-rebuild retry would pass, and history would quietly fall back to the tick this PR is supposed to remove from the critical path.

erskingardner (commented):
## Review (045f534)

The scheduling change is the right one: durable `GroupJoined` first, ordinary group-subscription rebuild in the background, accept during exclusive catch-up as typed `AccountWorkerBusy`, hydration accept on the live projection, and operation-class worker deadlines with an explicit unknown-completion timeout. Prior review items at d0ffa4e look addressed, and CI is green.

One gap against #1438:

**The primary acceptance test never observes the history subscription.** The new regression covers visibility, ordinary-rebuild retry, and busy-accept. It does not assert that `NostrSubscription::GroupMaintenance` is installed without the 15s tick, and there is no jitter assertion for the post-join self-update commit (the Validation section mentions a "post-join rotation jitter regression" that is not in this diff). The implementation calls `advance_post_join_maintenance_subscriptions` at the live/startup/catch-up seams, so this is coverage, not a suspected logic bug — but that call is the entire empty-group fix, and the current test would still pass if it were removed.

Add a `GroupMaintenance` assertion to the existing fixture (the blocker already leaves that variant unblocked). I would not hold the rest of the error-contract work for this, but I would not close #1438 until that check exists.

Non-blocking:
- On-device #1303 before/after measurement is still pending, as the PR body says.
- The uncapped background retry will keep publishing an account error every 60s on a permanently degraded relay; fine to leave.

PR #1446, opened 2026-08-14
marmot-app: reduce account setup relay round trips

### LINKED ISSUE #1436: marmot-app: account creation / nsec sign-in run 10-15 sequential relay round-trips on the blocking setup path; one dead relay hard-fails setup

erskingardner:
Closes #1436.

## What was wrong

Account creation and nsec import serialized avoidable relay and storage work. Setup created a fresh signer-bound SDK client for each initial event, connected directory relays one at a time, and synchronously re-fetched relay lists after they had already been acknowledged. That refetch could fail after successful publication and roll the new account back.

Fresh accounts also ran a guaranteed-empty legacy KeyPackage cutover scan, while initial KeyPackage publication opened a one-shot session before the managed account worker opened the same SQLCipher database again.

## What changed

- Publish the generated account's NIP-65 list, inbox list, empty follow list, and default profile as one mixed-endpoint batch through one signer-bound client.
- Keep relay-list records on discovery/bootstrap relays while preserving the existing outbox boundary for profile and follow-list records.
- Treat relay acknowledgements as the authoritative publication result and persist the declared relay-list projection directly instead of performing a setup-fatal refetch.
- Add durable bootstrap-publication started/confirmed phases so partial or ambiguous exposure retains the account for an idempotent retry.
- Connect directory relays concurrently and ignore one failed endpoint when another endpoint is usable.
- Reuse the bounded import preflight relay-list result and cap fallback discovery with the existing setup advisory deadline.
- Mark the legacy KeyPackage cutover scan complete when a generated account record is born.
- Publish the initial KeyPackage through the managed worker, leaving setup with one off-runtime-thread session open.
- Skip the redundant generated-account directory refresh because the batch already persists its profile and empty follow projection.

## Impact

A single unreachable bootstrap/default relay no longer makes account creation fail or multiplies its connect timeout across every setup publication. The happy path uses one batch for the four replaceable bootstrap records, avoids the relay-list confirmation refetch and guaranteed-empty cutover scan, and opens the session database once.

## Validation

- `just fast-ci`
- `cargo test -p marmot-account` — 98 tests passed
- `cargo test -p marmot-app --lib` — 664 tests passed
- Generated setup with one unreachable relay succeeded in 4.2 seconds
- Import with a live plus stalled discovery endpoint completed in 7.25 seconds
- Regression coverage verifies one setup session open, the pre-open cutover marker, zero-ack projection safety, no post-publish refetch dependency, and exact retry after a partial bootstrap batch

The full `marmot-app` integration invocation still reports five existing `relay_runtime` failures. Each reproduces unchanged in a clean `origin/master` worktree at `521b9e87`; the issue-specific setup tests pass.

## Summary by CodeRabbit

* **New Features**
  * Account setup now tracks bootstrap publication progress for improved status visibility and recovery.
  * New accounts publish relay lists, profile information, and follow data together during setup.
* **Bug Fixes**
  * Setup continues when individual relays are unavailable, provided at least one relay succeeds.
  * Prevented unnecessary directory refreshes during account creation.
  * Failed setup attempts preserve progress and support retry without losing account identity.
  * Improved recovery when reactivating accounts or using external signers.
* **Reliability**
  * Improved relay-client reuse, session handling, and relay connection deduplication.
  * Added validation for unsuccessful publication acknowledgements.

### DIFF
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -3484,9 +3484,10 @@ impl MarmotApp {
                 Err(_) => delete_failure_count += 1,
             }
         }
-        if delete_failure_count == 0 && self.mark_key_package_cutover_scan_complete(label).is_err()
-        {
-            delete_failure_count += 1;
+        if delete_failure_count == 0 {
+            // The marker helper emits its own privacy-safe warning. It is not
+            // a relay deletion failure, so do not fold it into this counter.
+            let _ = self.mark_key_package_cutover_scan_complete(label);
         }
         if non_current_event_count > 0 {
             tracing::info!(
@@ -5147,6 +5148,9 @@ impl MarmotApp {
             .account_publish_clients
             .lock()
             .unwrap_or_else(|poisoned| poisoned.into_inner());
+        // The signer is intentionally ignored on a cache hit. Callers that
+        // replace an account signer must evict this entry first, as
+        // register_external_signer and drop_account_caches do today.
         clients
             .entry(account_id_hex.to_owned())
             .or_insert_with(|| {
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -495,6 +495,10 @@ pub struct MarmotApp {
     chat_list_projection_stale: Arc<Mutex<HashSet<String>>>,
     audit_log_tracker_config: Arc<Mutex<AuditLogTrackerConfig>>,
     external_signers: Arc<Mutex<HashMap<String, RegisteredExternalSigner>>>,
+    /// One signer-bound publisher per account. Setup publishes and the managed
+    /// account worker share this client so the worker can reuse the same relay
+    /// pool instead of constructing another TCP/TLS/WebSocket stack.
+    account_publish_clients: Arc<Mutex<HashMap<String, Arc<dyn NostrRelayClient>>>>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -1255,6 +1259,7 @@ impl MarmotApp {
             chat_list_projection_stale: Arc::new(Mutex::new(HashSet::new())),
             audit_log_tracker_config: Arc::new(Mutex::new(AuditLogTrackerConfig::default())),
             external_signers: Arc::new(Mutex::new(HashMap::new())),
+            account_publish_clients: Arc::new(Mutex::new(HashMap::new())),
         }
     }
 
@@ -1322,6 +1327,7 @@ impl MarmotApp {
             chat_list_projection_stale: Arc::new(Mutex::new(HashSet::new())),
             audit_log_tracker_config: Arc::new(Mutex::new(AuditLogTrackerConfig::default())),
             external_signers: Arc::new(Mutex::new(HashMap::new())),
+            account_publish_clients: Arc::new(Mutex::new(HashMap::new())),
         }
     }
 
@@ -1606,6 +1612,127 @@ impl MarmotApp {
         .await
     }
 
+    /// Publish every generated-identity bootstrap record through one scoped
+    /// relay batch. The SDK connects the endpoint union once for the two relay
+    /// lists, empty follow list, and default profile, then returns one ordered
+    /// acknowledgement result per replaceable event.
+    pub(crate) async fn publish_generated_account_bootstrap(
+        &self,
+        label: &str,
+        bootstrap: AccountRelayListBootstrap,
+        profile: &UserProfileMetadata,
+    ) -> Result<AccountRelayListStatus, AppError> {
+        if bootstrap.default_relays.is_empty() {
+            return Err(AppError::MissingDefaultRelays);
+        }
+        self.validate_account_relay_list_declarations(&bootstrap, None)?;
+        let account = self.account_home().account(label)?;
+        let signer = self.account_signer_for_summary(&account)?;
+        let account_id = MemberId::new(hex::decode(&account.account_id_hex)?);
+        // Relay-list records are discoverability maps and therefore go to the
+        // bootstrap route. Profiles and contact lists are outbox content: keep
+        // them on the declared write relays even though the mixed batch retains
+        // the union of both endpoint sets once.
+        let content_endpoints =
+            self.outbox_endpoints(&account.account_id_hex, bootstrap.default_relays.clone());
+        let requested = publish_endpoints_from_bootstrap(&bootstrap);
+        let mut endpoints = self.outbox_endpoints(&account.account_id_hex, requested.clone());
+        for endpoint in requested {
+            if !endpoints.iter().any(|existing| existing.0 == endpoint.0) {
+                endpoints.push(endpoint);
+            }
+        }
+
+        let mut requests = Vec::with_capacity(4);
+        for list_kind in [
+            NostrAccountRelayListKind::Nip65,
+            NostrAccountRelayListKind::Inbox,
+        ] {
+            requests.push(NostrEventPublishRequest {
+                endpoints: endpoints.clone(),
+                event: NostrAccountRelayListPublication {
+                    account_id: account_id.clone(),
+                    list_kind,
+                    relays: bootstrap.default_relays.clone(),
+                    publish_endpoints: endpoints.clone(),
+                }
+                .to_event()?,
+                required_acks: 1,
+            });
+        }
+        requests.push(NostrEventPublishRequest {
+            endpoints: content_endpoints.clone(),
+            event: NostrTransportEvent::new_unsigned(
+                account.account_id_hex.clone(),
+                KIND_NOSTR_CONTACT_LIST,
+                Vec::new(),
+                String::new(),
+            ),
+            required_acks: 1,
+        });
+        requests.push(NostrEventPublishRequest {
+            endpoints: content_endpoints.clone(),
+            event: NostrTransportEvent::new_unsigned(
+                account.account_id_hex.clone(),
+                KIND_NOSTR_METADATA,
+                Vec::new(),
+                serde_json::to_string(&directory::records::profile_content_json(profile))?,
+            ),
+            required_acks: 1,
+        });
+
+        let relay_client =
+            self.relay_client_for_account_id(&account.account_id_hex, signer.as_nostr_signer());
+        for outcome in relay_client.publish_events(&requests).await {
+            if outcome?.accepted.is_empty() {
+                return Err(AppError::Publish(
+                    "relay acknowledged zero account bootstrap events".to_owned(),
+                ));
+            }
+        }
+
+        let relays = bootstrap
+            .default_relays
+            .iter()
+            .map(|endpoint| endpoint.0.clone())
+            .collect::<Vec<_>>();
+        let mut status = AccountRelayListStatus {
+            complete: false,
+            missing: Vec::new(),
+            default_relays: Vec::new(),
+            bootstrap_relays: endpoints
+                .iter()
+                .map(|endpoint| endpoint.0.clone())
+                .collect(),
+            nip65: AccountRelayListState {
+                kind: KIND_NIP65_RELAY_LIST,
+                relays: relays.clone(),
+                read_relays: relays.clone(),
+                write_relays: relays.clone(),
+            },
+            inbox: AccountRelayListState {
+                kind: KIND_MARMOT_INBOX_RELAY_LIST,
+                relays,
+                read_relays: Vec::new(),
+                write_relays: Vec::new(),
+            },
+        };
+        status.refresh();
+        self.remember_directory_relay_lists(&account.account_id_hex, &status)?;
+        self.remember_directory_follow_edges_for_search(
+            &account.account_id_hex,
+            &directory::FetchedFollowList {
+                follows: Vec::new(),
+                source_relays: content_endpoints
+                    .iter()
+                    .map(|endpoint| endpoint.0.clone())
+                    .collect(),
+            },
+        )?;
+        self.remember_directory_profile(&account.account_id_hex, profile)?;
+        Ok(status)
+    }
+
     pub async fn publish_missing_account_relay_lists(
         &self,
         label: &str,
@@ -1833,7 +1960,9 @@ impl MarmotApp {
                 endpoints.push(endpoint);
             }
         }
-        let relay_client = self.relay_client_for_endpoints(signer.as_nostr_signer(), &endpoints);
+        let relay_client =
+            self.relay_client_for_account_id(&account_id_hex, signer.as_nostr_signer());
+        let mut requests = Vec::with_capacity(list_kinds.len());
         for list_kind in list_kinds {
             let event = if *list_kind == NostrAccountRelayListKind::Nip65
                 && let Some(relays) = nip65_relay_set
@@ -1853,10 +1982,80 @@ impl MarmotApp {
                 }
                 .to_event()?
             };
-            relay_client.publish_event(&endpoints, &event, 1).await?;
+            requests.push(NostrEventPublishRequest {
+                endpoints: endpoints.clone(),
+                event,
+                required_acks: 1,
+            });
         }
-        self.fetch_account_relay_list_status_for_account_id(&account_id_hex, endpoints)
-            .await
+        for outcome in relay_client.publish_events(&requests).await {
+            if outcome?.accepted.is_empty() {
+                return Err(AppError::Publish(
+                    "relay acknowledged zero account relay-list events".to_owned(),
+                ));
+            }
+        }
+
+        // The signed replaceable events above are the authoritative effect of
+        // this operation. Persist their declared state directly after every
+        // event has an acknowledgement; synchronously querying the same relays
+        // again adds no confirmation strength and used to turn a post-success
+        // read outage into account-setup rollback (mdk#1436).
+        let mut status = self.account_relay_list_status_for_account_id(&account_id_hex)?;
+        for list_kind in list_kinds {
+            match list_kind {
+                NostrAccountRelayListKind::Nip65 => {
+                    let (read_relays, write_relays) = nip65_relay_set
+                        .map(|relays| {
+                            (
+                                relays
+                                    .read_relays
+                                    .iter()
+                                    .map(|endpoint| endpoint.0.clone())
+                                    .collect::<Vec<_>>(),
+                                relays
+                                    .write_relays
+                                    .iter()
+                                    .map(|endpoint| endpoint.0.clone())
+                                    .collect::<Vec<_>>(),
+                            )
+                        })
+                        .unwrap_or_else(|| {
+                            let relays = bootstrap
+                                .default_relays
+                                .iter()
+                                .map(|endpoint| endpoint.0.clone())
+                                .collect::<Vec<_>>();
+                            (relays.clone(), relays)
+                        });
+                    status.nip65 = AccountRelayListState {
+                        kind: KIND_NIP65_RELAY_LIST,
+                        relays: write_relays.clone(),
+                        read_relays,
+                        write_relays,
+                    };
+                }
+                NostrAccountRelayListKind::Inbox => {
+                    status.inbox = AccountRelayListState {
+                        kind: KIND_MARMOT_INBOX_RELAY_LIST,
+                        relays: bootstrap
+                            .default_relays
+                            .iter()
+                            .map(|endpoint| endpoint.0.clone())
+                            .collect(),
+                        read_relays: Vec::new(),
+                        write_relays: Vec::new(),
+                    };
+                }
+            }
+        }
+        push_unique_strings(
+            &mut status.bootstrap_relays,
+            endpoints.iter().map(|endpoint| endpoint.0.clone()),
+        );
+        status.refresh();
+        self.remember_directory_relay_lists(&account_id_hex, &status)?;
+        Ok(status)
     }
 
     /// Validate caller-owned relay-list declarations without applying the dial
@@ -2956,7 +3155,7 @@ impl MarmotApp {
             AccountDeviceSession::open(session_config).map_err(external_signer_session_error)?;
 
         let publish_client =
-            self.relay_client_for_endpoints(nostr_signer.clone(), &self.relay_endpoints());
+            self.relay_client_for_account_id(&account.account_id_hex, nostr_signer.clone());
         let adapter = relay_plane.account_adapter(account_id.clone(), publish_client);
 
         let key_packages = AppKeyPackagePublisher {
@@ -3261,8 +3460,9 @@ impl MarmotApp {
                 Err(_) => delete_failure_count += 1,
             }
         }
-        if delete_failure_count == 0 {
-            self.mark_key_package_cutover_scan_complete(label);
+        if delete_failure_count == 0 && self.mark_key_package_cutover_scan_complete(label).is_err()
+        {
+            delete_failure_count += 1;
         }
         if non_current_event_count > 0 {
             tracing::info!(
@@ -3355,7 +3555,7 @@ impl MarmotApp {
         self.key_package_cutover_scan_complete_path(label).exists()
     }
 
-    fn mark_key_package_cutover_scan_complete(&self, label: &str) {
+    fn mark_key_package_cutover_scan_complete(&self, label: &str) -> Result<(), AppError> {
         let path = self.key_package_cutover_scan_complete_path(label);
         let result = path
             .parent()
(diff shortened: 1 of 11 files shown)


crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4614,30 +4650,64 @@ impl AccountManager {
...
+                            tracing::debug!(
+                                target: "marmot_app::runtime",
+                                method = "setup_relay_lists_for_account",
+                                error_kind = error.privacy_safe_kind(),
+                                "using cached relay-list status after advisory import fetch failed"
+                            );
+                            self.app.account_relay_list_status(&account.label)?
+                        }
+                        None => self.app.account_relay_list_status(&account.label)?,
+                    }
```
erskingardner, line 4713: **Question: a failed status fetch now falls through to publishing request defaults for imports.**

When the advisory `import_relay_list_status` fetch errors or times out, this falls back to the local cache — which is empty for a fresh nsec import — and the flow then continues into `publish_missing_account_relay_lists_from_status` with `request.default_relays` as the declared content. For an imported pre-existing identity whose real kind-10002/10050 live on relays that are momentarily slow/unreachable, the window between the 10s advisory fetch cap and the longer publish budget means setup can now rewrite the identity's *public* relay declarations to the app's default set, where the old unbounded `.await?` would have failed setup instead. The issue asked to bound this fetch and to stop the post-publish refetch being setup-fatal; publishing defaults on a fetch *error* goes one step beyond that.

Is the error-vs-empty conflation intended? If not, consider distinguishing "fetch failed" from "fetch returned no lists": on failure with an incomplete cache, return a retryable error (the journal retains the account) rather than publishing request-declared defaults over an identity's existing lists.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4277,6 +4250,12 @@ impl AccountManager {
                     }
                     return Err(err.into());
                 }
+                if account.signed_out {
+                    account = self
+                        .app
+                        .account_home()
+                        .set_account_signed_out(&account.label, false)?;
+                }
```
erskingardner, line 4269: **Question: reactivation is no longer atomic — a KP-step failure leaves the account signed in.**

The flip is required for the worker-owned KP publish (`worker_commands` rejects signed-out accounts), but it changes reactivation semantics. Reactivation carries no setup journal, so if `publish_initial_key_package_for_account` fails after this point the account stays signed-in, and a same-nsec retry no longer matches `signed_out_account_for_identity` — it falls into `import_nostr_account_idempotent`, which returns `AccountExists` for an existing journal-less account unless the legacy `pending_replacement` adoption branch happens to fire. Pre-PR, a failed reactivation stayed signed-out and the retry resumed cleanly.

The end state is mostly self-healing (the account is active and the worker starts on the next reconcile), but the retry UX regresses and the signed-in-yet-setup-failed state is new. Consider compensating on the error path (flip back to signed-out when a reactivation's KP step fails) or beginning a setup journal for reactivation so the retry has a defined resume point.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4563,31 +4546,83 @@ impl AccountManager {
...
+        if matches!(
+            phase,
+            AccountSetupPhase::BootstrapPublicationConfirmed
+                | AccountSetupPhase::KeyPackagePublicationStarted
+                | AccountSetupPhase::KeyPackagePublicationConfirmed
+        ) {
+            let status = self.app.account_relay_list_status(&account.label)?;
+            if !status.complete {
+                return Err(AppError::MissingRelayLists(status.missing));
+            }
```
erskingardner: Minor robustness: on a retry with phase ≥ `BootstrapPublicationConfirmed`, an incomplete local status returns `MissingRelayLists`, and `setup_failure_can_roll_back` is false at this phase — so every subsequent retry hits this same branch and the account can never complete setup. The bootstrap records are replaceable and republication is idempotent, so republishing here (exactly what the `BootstrapPublicationStarted` branch below does) would be strictly more robust for the narrow "journal intact but directory projection lost" case.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4688,10 +4786,12 @@ impl AccountManager {
...
         }
-        self.app.status(&account.label)?;
-        let mut client = self.app.client(&account.label).await?;
-        let key_package = client.publish_key_package().await?;
-        Ok(key_package.bytes().len())
+        // Let the managed worker own the only session opened for this setup.
+        // Its open runs through `blocking_app_task`; publishing through the
+        // worker avoids the former one-shot session open followed immediately
+        // by a second worker session open.
+        self.reconcile().await?;
```
erskingardner: Minor: this `reconcile()` is redundant — `publish_key_package` → `worker_commands` already calls `self.reconcile().await?` before dispatching the command, so this is a second no-op pass (lock + account scan) on every setup. Drop it, or keep it and fold the comment's guarantee into the `publish_key_package` call.

erskingardner (commented):
Solid PR overall: the batch bootstrap, ack-authoritative projection, concurrent directory connects, worker-owned session, and the new setup phases line up with #1436's required behavior, and the regression tests (one-batch/no-refetch, zero-ack, partial-batch retry, unreachable relay, single session open) cover the acceptance criteria well. CI is green (only "Simulator and vectors" still pending; the relay_runtime failures noted in the description don't appear in CI, so they look environment-specific).

Two behavior-change questions inline that I'd like answered before merge — neither is a clear-cut bug, but both are semantic shifts the description doesn't mention:

1. Import: a *failed* relay-list status fetch now falls through to publishing the request's default relays as the identity's NIP-65/inbox declarations.
2. Reactivation: the early `signed_out` flip means a failed reactivation leaves the account signed in, and a same-nsec retry can surface `AccountExists` instead of resuming.

Smaller notes (non-blocking):

- Duplicated validation: `MissingDefaultRelays` + `validate_account_relay_list_declarations` run in both `setup_generated_account_bootstrap` and `publish_generated_account_bootstrap`; likewise the new empty-relay early-returns in `setup_relay_lists_for_account` duplicate `publish_relay_lists_for_new_account`'s own checks. One layer is enough.
- Docs: `crates/marmot-account/AGENTS.md` still states "Roll back pending work when publication fails or does not meet the required acknowledgement count" — `BootstrapPublicationStarted` is now a deliberate exception to that rule; worth a one-line update so the next agent doesn't "fix" it back.
- I agree with CodeRabbit's two minor findings and won't duplicate them: (a) the `mark_key_package_cutover_scan_complete` failure cleanup in `create_nostr_account_from_setup` should also call `remove_account_key_package_artifacts` so a partially-written marker can't survive for a reused label; (b) dedup parsed `RelayUrl`s before the concurrent connect in `fetch_directory_events` so trailing-slash variants can't remove a shared relay out from under a successful twin.

crates/marmot-app/src/runtime/mod.rs
erskingardner, line 4713: This treats a failed or timed-out status fetch as “the identity has no relay lists.” For a fresh nsec the local cache is `AccountRelayListStatus::empty()`, so both NIP-65 and inbox are `missing`, and the branch below signs and publishes `request.default_relays` as this identity’s kind-10002/10050.

Those events are replaceable. A transient discovery outage on import of a live identity will overwrite its real lists on any relay that accepts the new events — including later discovery from the app’s own bootstrap set.

#1436 asked for a bounded fetch and reuse of a successful preflight, not write-on-read-failure. One dead peer is already handled by concurrent connect plus `recent_relay_lists`. When this fetch errors or times out and the cache is empty or incomplete, return `MissingRelayLists` (or a distinct discovery-failed error). Only publish defaults after a successful fetch that established the lists are actually absent.

crates/marmot-app/src/runtime/mod.rs
erskingardner, line 4269: This is a retry-breaking regression, not only a semantic shift.

Reactivation comes from `signed_out_account_for_identity` and does not open a setup journal. The `setup_phase.is_some()` guard a few lines above therefore skips `KeyPackagePublicationStarted`. After this flip, a KP failure leaves a signed-in account with no journal.

A same-nsec retry then misses `signed_out_account_for_identity` and lands in `import_nostr_account_idempotent`, which returns `AccountExists` when the record exists and the journal does not. Previously the flip happened only after KP success, so failure stayed on the signed-out resume path.

Journal the reactivation (and set KP-started) before the flip so import can resume, or revert `signed_out` on the KP error path. Please add a test: sign out, reactivate with a failing KP publish, retry the same nsec, and assert resume rather than `AccountExists`.

erskingardner (commented):
Would request changes if GitHub allowed it on an author-owned PR. The batch bootstrap, ack-authoritative projection, concurrent directory connects, cutover marker, and worker-owned session match #1436, and the new tests cover the generated-account happy path well.

Two setup-path behavior changes should be fixed before merge: a failed import relay-list fetch can overwrite a live identity's NIP-65/inbox with app defaults, and a failed reactivation KeyPackage step leaves the account signed in with no journal so the same-nsec retry returns `AccountExists`.

I agree with the existing notes on cutover-artifact cleanup, `RelayUrl` dedup before concurrent connect, the redundant `reconcile()`, and republishing when phase is already `BootstrapPublicationConfirmed` but the local projection is incomplete. Not repeating those inline.

erskingardner (commented):
Recommend merge (GitHub blocks approve on an author-owned PR). The follow-up commit fixes both blocking setup-path issues from the last review, and CI is green.

- Import no longer publishes request defaults after a failed or timed-out relay-list fetch; it reuses a complete cache or returns `MissingRelayLists`.
- Failed reactivation now restores signed-out via `deactivate_account`, and the same-nsec retry path is covered by `failed_reactivation_key_package_publish_restores_signed_out_retry`.
- Confirmed-but-incomplete bootstrap now republishes instead of getting stuck; cutover-marker cleanup, `RelayUrl` dedup, the extra `reconcile()`, and the AGENTS.md rollback wording are also addressed.

The generated-account batch, ack-authoritative projection, concurrent directory connects, and worker-owned session still match #1436.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4447,6 +4433,12 @@ impl AccountManager {
         };
 
         let key_package_bytes = if request.publish_initial_key_package {
+            if account.signed_out {
+                account = self
+                    .app
+                    .account_home()
+                    .set_account_signed_out(&account.label, false)?;
+            }
```
erskingardner, line 4454: Optional, non-blocking: the `create_or_import_account` KP error path now compensates a failed reactivation with `deactivate_account`, but this external-signer twin keeps the account signed-in on failure. Retry here is safe (`add_external_signer_account` is idempotent for an already-external account and the journal drives resume), so nothing is broken — but a signed-out external account that fails its KP publish will show as signed-in until the next login. Consider the same `deactivate_account` compensation for symmetry, or a short comment saying the asymmetry is deliberate.

erskingardner (commented):
Re-reviewed at 892b53a9. **Recommend merge** (posted as a comment — GitHub won't let me self-approve my own PR). All four of my findings are addressed, each with regression coverage, and CI is fully green:

1. **Blind publish on fetch error** — fixed: `complete_cached_relay_list_status` now returns the cached projection only when complete, else `MissingRelayLists`; a failed discovery can no longer rewrite an imported identity's relay declarations with request defaults. `failed_import_relay_discovery_does_not_publish_default_lists` pins the behavior. Total-outage imports roll back cleanly (nothing published, so deletion is safe), and single-relay failure is still tolerated by the concurrent connect.
2. **Reactivation atomicity** — fixed: the KP error path now `deactivate_account`s a reactivation, restoring the durable signed-out marker and stopping the just-started worker, so the same nsec resumes instead of hitting `AccountExists`. `failed_reactivation_key_package_publish_restores_signed_out_retry` exercises exactly the scenario I traced.
3. **Stuck retry at confirmed phase** — fixed: an incomplete projection now falls through to an idempotent republish, and `mark_bootstrap_publication_confirmed` correctly no-ops for `KeyPackagePublication*` phases so a later-stage journal can't be regressed. `confirmed_generated_bootstrap_republishes_when_projection_is_missing` covers it.
4. **Redundant reconcile** — removed.

The duplicated-validation layering is now documented with comments (deliberate boundary duplication — fine), the `marmot-account` AGENTS.md rollback rule now matches the journaled-exposure model, and both CodeRabbit items landed (`remove_account_key_package_artifacts` on the birth-failure cleanup; `RelayUrl` sort+dedup with a trailing-slash unit test). The `publish_route_including_requested` extraction and labeled per-record zero-ack errors are nice touches.

One optional consistency note inline — non-blocking.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4277,19 +4250,34 @@ impl AccountManager {
...
                         // is sufficient to retry safely after any ordinary error
                         // or task cancellation. Never delete possibly exposed
                         // exact bytes/private material here.
+                        // A reactivation has no setup journal, so restore its
+                        // durable signed-out state and stop the just-started
+                        // worker. The same nsec can then resume this exact
+                        // lifecycle attempt instead of failing AccountExists.
+                        if reactivating_existing {
+                            self.deactivate_account(&account.label).await?;
+                        }
```
jgmontoya, line 4292: If the initial `KeyPackage` publish fails during reactivation, `deactivate_account` runs to restore the durable `signed_out` marker and stop the just-started worker — good, that's the right compensating action. But it's called with `?`, so if `deactivate_account` itself errors (storage error, lock poisoning propagated as an error, etc.), that unrelated error replaces `err` and the caller never sees why the KeyPackage publish actually failed. Given the whole point of this branch is to surface "the rejected KeyPackage publication" (per the new `failed_reactivation_key_package_publish_restores_signed_out_retry` test), silently swapping in a cleanup error undermines that. Suggest logging the cleanup failure and still returning the original `err`, e.g.:

```rust
if reactivating_existing && let Err(cleanup_err) = self.deactivate_account(&account.label).await {
    tracing::warn!(..., error_kind = cleanup_err.privacy_safe_kind(), "failed to restore signed-out state after key package publish failure");
}
return Err(err);
```

crates/marmot-app/src/lib.rs
```diff
@@ -3261,8 +3484,9 @@ impl MarmotApp {
                 Err(_) => delete_failure_count += 1,
             }
         }
-        if delete_failure_count == 0 {
-            self.mark_key_package_cutover_scan_complete(label);
+        if delete_failure_count == 0 && self.mark_key_package_cutover_scan_complete(label).is_err()
+        {
+            delete_failure_count += 1;
```
jgmontoya: `delete_failure_count` is only read afterward inside `if non_current_event_count > 0 { tracing::info!(...) }` (lines 3491-3500). When the scan finds nothing to retire (`non_current_event_count == 0` — the common, guaranteed-empty case this PR is optimizing for), this increment has no observable effect: the log line that would report it never fires. `mark_key_package_cutover_scan_complete` already emits its own `tracing::warn!` on failure, so nothing is silently lost, but the added bookkeeping here is effectively dead code in the exact path it's meant to matter for. Either gate the log on `delete_failure_count > 0` too, or drop the increment and rely on the warning already emitted inside `mark_key_package_cutover_scan_complete`.

crates/marmot-app/src/relay_plane/directory.rs
```diff
@@ -346,25 +347,47 @@ impl DirectoryRelayFetcher for NostrSdkDirectoryRelayFetcher {
...
+        }
+        let mut connected = vec![false; relay_urls.len()];
+        let mut failed_urls = Vec::new();
+        while let Some(result) = connects.join_next().await {
+            match result {
+                Ok((index, _relay_url, true)) => connected[index] = true,
+                Ok((_index, relay_url, false)) => failed_urls.push(relay_url),
+                Err(_) => {}
+            }
+        }
```
jgmontoya, line 401: If a spawned connect task panics or is cancelled, `Err(_)` is dropped silently — that relay was already
`add_relay`'d onto `self.client` but never gets `remove_relay`'d (only `failed_urls` entries get removed at line 381-383). This is a narrow edge case (a panic inside `client.connect_relay` would be unusual), but it's a small resource-leak/cleanup gap worth a one-line fix: push the URL into `failed_urls` (or a parallel `panicked_urls`) in the `Err(_)` arm too, so it still gets removed from the client.

crates/marmot-app/src/relay_plane/directory.rs
```diff
@@ -346,25 +347,47 @@ impl DirectoryRelayFetcher for NostrSdkDirectoryRelayFetcher {
...
+        for relay_url in failed_urls {
+            let _ = self.client.remove_relay(relay_url).await;
+        }
+        let relay_urls = relay_urls
+            .into_iter()
+            .zip(connected)
+            .filter_map(|(relay_url, connected)| connected.then_some(relay_url))
+            .collect::<Vec<_>>();
+        if relay_urls.is_empty() {
+            return Err("connect relays failed".to_owned());
```
jgmontoya: Previously each failure mode surfaced its own message (`"invalid relay URL"`, `"add relay failed"`,
`"connect relay timed out"`, `"connect relay failed"`). Now, when every candidate fails, callers only see the generic `"connect relays failed"`. That's an acceptable trade for the new "one success is enough" tolerance, but when *all* relays fail it's worth keeping enough detail (e.g. count of add-failures vs. timeouts vs. connect failures) for operators debugging a fully-down directory fetch.

crates/marmot-app/src/lib.rs
```diff
@@ -4904,18 +5134,26 @@ impl MarmotApp {
...
+            .account_publish_clients
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        clients
+            .entry(account_id_hex.to_owned())
+            .or_insert_with(|| {
+                let client = NostrSdkClient::builder().signer(signer).build();
+                Arc::new(NostrSdkRelayClient::new(client))
+            })
+            .clone()
```
jgmontoya, line 5160: On a cache hit, the `signer` argument is discarded entirely — the cached client keeps whatever signer it was built with. This is fine today because every call site invalidates the cache before a signer could legitimately change (`register_external_signer` at line 5197, `drop_account_caches` at line 5080), and local-signing accounts never change signer. But the function signature doesn't communicate that invariant, so a future call site that swaps a signer without also clearing `account_publish_clients` would silently keep publishing with the stale signer. A debug-only assertion, or at least a comment on the parameter noting "ignored on cache hit — caller must invalidate the cache before changing signer", would make the invariant harder to violate by accident.

PR #1447, opened 2026-08-14
fix(marmot-app): keep the runtime responsive under SQL, lag, and media HTTP

### LINKED ISSUE #740: marmot-app: subscribe_notifications runs blocking SQLCipher I/O on the runtime thread and silently drops notifications on broadcast lag

### LINKED ISSUE #741: marmot-app: account worker head-of-line blocking — inline media/HTTP awaits stall inbound message processing and can drop deliveries
**Severity:** MEDIUM
**Component:** `crates/marmot-app`

## Summary

The per-account worker's steady-state `select!` awaits each command inline. The media commands (`UploadMedia`/`DownloadMedia`/`DownloadGroupImage`) perform Blossom HTTP transfers bounded only by a 60s total timeout. While that future is parked, the loop's `client.next_event()` arm is not polled, so no inbound relay delivery is ingested/projected and no scheduled convergence runs for the whole account — for up to ~60s per transfer. Meanwhile the router `try_send`s into a bounded account queue and **drops** deliveries when it is full, so a long transfer during heavy traffic can silently drop inbound messages.

## Location

- `crates/marmot-app/src/runtime/account_worker.rs:554-570` — steady-state loop awaits `handle_account_worker_command(...).await` inline.
- Handlers: `account_worker.rs:1096-1123` (`UploadMedia`/`DownloadMedia`), `:1045-1048` (`DownloadGroupImage`).
- `crates/marmot-app/src/client/mod.rs:1420-1438` — `download_media`.
- `crates/marmot-app/src/media/blossom.rs:17` — `MEDIA_HTTP_TOTAL_TIMEOUT = 60s`.
- `crates/marmot-app/src/relay_plane/mod.rs:541` — router `try_send` drops on `Full`.

## Details

```rust
// account_worker.rs:554
command = commands.recv() => {
    Some(command) => {
        handle_account_worker_command(&mut client, command, ...).await; // <- 60s HTTP transfer parks the loop
```

Notably `download_media` only needs `&mut self` for the cheap setup (`ensure_group`, `sync_runtime_groups`, secret lookup); the actual `download_encrypted_media(...)` — the 60s part — operates on owned/cloned values and does not require exclusive client access. So the head-of-line block is avoidable. (This overlaps in blast radius with #712 but is a distinct cause: inline command await, not `next_event` cancellation.)

## Suggested fix

Run the network transfer off the worker: capture the secret + policy under `&mut client`, then `tokio::spawn` (or return a future the caller drives) the `download_encrypted_media`/`upload_*` HTTP work so the command loop keeps servicing `next_event`.

---
_Filed from a meticulous automated code-review pass; verified against source and deduplicated against existing open/closed issues (distinct from #712)._

### LINKED ISSUE #761: marmot-app: subscribe_chats and subscribe_group_state don't recover from broadcast Lagged — a missed update leaves the chat list / group state permanently stale
**Severity:** MEDIUM (silent, and permanent for a group that goes quiet after the missed update)
**Component:** `crates/marmot-app`

## Summary

`subscribe_chats` and `subscribe_group_state` handle `broadcast::error::RecvError::Lagged(_)` with a bare `continue` — no snapshot re-read. Every *other* subscription in this module recovers explicitly on lag. Because each of these two handlers only refreshes the *specific* group carried by an event, any group whose update was in the dropped window stays stale until some *later* event happens to touch that same group. For a group that goes silent right after the missed update (e.g. its final "removed"/"left" transition), the stale entry is **permanent**.

## Location

- `crates/marmot-app/src/runtime/subscriptions.rs:999` — `subscribe_chats`: `Err(broadcast::error::RecvError::Lagged(_)) => continue,`
- `crates/marmot-app/src/runtime/subscriptions.rs:1244` — `subscribe_group_state`: same bare `continue`.
- Contrast — siblings that recover:
  - `:797` `subscribe_messages` re-reads history and re-emits missed rows (its comment: *"a slow consumer cannot silently drop messages"*).
  - `:925` `subscribe_timeline_messages` sends a `Refresh`.
  - `:1086` `subscribe_chat_list` re-reads and reconciles via `ChatListUpdateTrigger::SnapshotRefresh`.
- Broadcast capacity: `crates/marmot-app/src/runtime/mod.rs:791` (1024).

## Details

The shared events broadcast has capacity 1024. During a bursty catch-up (many messages / projection updates across groups), a `subscribe_chats` task — which does a `blocking_app_task(app.group(...)).await` per matching event and thus drains slowly — can lag past 1024 and drop events. Each dropped event only would have refreshed one `group_id`; that group's chat-list row (archived/renamed/member-left/deleted state) is then not refreshed, and stays wrong until another event touches it. `subscribe_group_state` self-heals on the next event for its one group, so its window is narrower, but the same permanent-staleness edge exists if that group goes quiet.

## Suggested fix

On `Lagged`, re-read the snapshot and reconcile against the stored fingerprints before continuing — for `subscribe_chats` mirror `subscribe_chat_list` (re-read `groups`/`visible_groups` and diff fingerprints); for `subscribe_group_state` re-read the single subscribed group.

---
<sub>Filed from a meticulous manual + multi-agent code-review pass; verified against source and deduplicated against existing open/closed issues.</sub>

### LINKED ISSUE #762: marmot-app: wake-driven notification collection runs blocking SQLCipher/directory I/O on the runtime thread (distinct site from #740)
**Severity:** MEDIUM (blocks a tokio worker thread on file I/O; same class as #740, different code path)
**Component:** `crates/marmot-app`

## Summary

The wake-driven notification drain calls `collect_notification_update_from_event(&app, …)` **synchronously inside the async task**, not through `blocking_app_task`. That resolver path re-opens per-account SQLCipher caches and the directory cache, i.e. blocking file I/O executed directly on a tokio worker thread. This is the same anti-pattern as the already-filed #740 (`subscribe_notifications`), but a **distinct entry point** (the wake/background collection loop) that a generic fix for #740 would not automatically cover.

## Location

- `crates/marmot-app/src/runtime/mod.rs:973-1004` — the drain loop calls `collect_notification_update_from_event(&app, &mut resolver, &event, &mut notifications)` at `:976` and again at `:994`, both directly `.await`-free on the runtime thread.
- `collect_notification_update_from_event` → `notifications::NotificationResolver` lookups (`crates/marmot-app/src/notifications.rs`) open the per-account SQLCipher settings/group/chat-muted/user caches and the directory cache.
- Related: #740 (`subscribe_notifications` blocking SQLCipher I/O + dropped notifications).

## Details

The per-batch `NotificationResolver` (comment `#639`) memoizes repeats, so subsequent lookups for the same account/group/sender are cheap, but the **first** lookup per account/group/sender in a wake still performs synchronous SQLCipher/directory file I/O on the runtime thread. Under a burst that spans many accounts/groups/senders, the runtime thread stalls on real disk I/O.

## Suggested fix

Run the per-event notification projection (or the whole drain batch) through `blocking_app_task`, consistent with the other blocking projections in this module. If #740 is addressed by moving the resolver work to a blocking pool generically, ensure this call site is covered too.

---
<sub>Filed from a meticulous manual + multi-agent code-review pass; verified against source and deduplicated against existing open/closed issues.</sub>

### LINKED ISSUE #802: marmot-app: subscribe_timeline_messages / subscribe_chats / subscribe_chat_list / subscribe_group_state run their initial SQLCipher snapshot read synchronously on the caller thread, while sibling subscribe_messages offloads via blocking_app_task (distinct sites; follow-up to #740/#762)
## Summary

Four subscription constructors are non-`async` `fn`s that run a blocking SQLCipher snapshot query **inline** on the caller's thread. The sibling `subscribe_messages` is the only one that is `async` and pushes its snapshot query onto the blocking pool (`blocking_app_task`), which signals the authors expect these snapshots to be offloaded. When the inline variants are invoked from within a runtime worker (the uniffi/CLI surfaces), the SQLCipher decrypt + query stalls that worker thread; chat-list/timeline snapshots can be large.

## Location

- `crates/marmot-app/src/runtime/subscriptions.rs`:
  - `subscribe_timeline_messages` (`:892`) — inline `app.timeline_messages_with_query(...)`
  - `subscribe_chats` (`:978-982`) — inline `app.groups(...)`
  - `subscribe_chat_list` (`:1071`) — inline `app.chat_list(...)`
  - `subscribe_group_state` (`:1233`) — inline `app.group(...)`
- Contrast `subscribe_messages` (`:749-752`) — `async`, wraps its snapshot in `blocking_app_task`.

## Impact / severity

The blocking is real; whether it lands on a runtime thread depends on the caller, and the inconsistency with `subscribe_messages` strongly suggests it does. **LOW** given the caller-dependence and that it is the same class already tracked for the notifications path.

## Suggested fix

Mirror `subscribe_messages`: make these constructors `async` (or otherwise offload the initial snapshot via `blocking_app_task`) so the SQLCipher read never runs on the runtime thread.

## Dedup

Checked open + closed. This shares the "blocking SQLCipher on the runtime thread" theme with **#740** (`subscribe_notifications`) and **#762** (wake-driven notification collection), but those are scoped specifically to the **notifications** path. These four snapshot sites are **distinct code locations** not named in either issue — filed as a distinct-site follow-up per the repo's sibling-issue convention.

---
_Found during a methodical multi-crate code review; verified against `subscriptions.rs` (the four sites vs `subscribe_messages:749-752`)._

### LINKED ISSUE #863: marmot-app: collect_notifications_after_wake discards already-buffered notification events when catch-up hits its timeout
**Severity:** MEDIUM
**Component:** `crates/marmot-app`

## Summary

`collect_notifications_after_wake` is the time-boxed background-wake path (iOS NSE / FCM data message) that surfaces notifications collected during a short catch-up. On the **timeout** branch it returns immediately with an **empty** notification list, dropping every `MarmotAppEvent` that the partial catch-up already published into the subscribed broadcast receiver. Because a background wake deliberately runs on a sub-second budget over cold sockets, timeout is a *common* outcome — so a productive-but-slow wake that already received new messages reports `Failed`/nothing and the host shows no notification.

This is a **distinct defect from #762 and #740**. #762 is about blocking SQLCipher I/O on the runtime thread inside this same function; #740 is about `subscribe_notifications` dropping on broadcast lag. Neither addresses the early-return on the catch-up timeout discarding events that were already buffered.

## Location

- `crates/marmot-app/src/runtime/mod.rs:940-968` — `collect_notifications_after_wake`.
  - `:947` subscribes `events` **before** catch-up.
  - `:948` runs `timeout(max_wait, self.catch_up_accounts())`.
  - `:961-967` the `Err(_)` (timeout) arm returns `BackgroundNotificationCollection { status: Failed, notifications /* still empty */, error: ... }` **before** the drain loop.
  - `:976-999` the drain loop (`events.try_recv()` → `collect_notification_update_from_event`) only runs on the `Ok(Ok(()))` success path.

## Details

```rust
let mut events = self.events.subscribe();                    // :947
let catch_up = timeout(max_wait, self.catch_up_accounts()).await; // :948
let remaining = max_wait.saturating_sub(started.elapsed());  // :949 — computed for "surface what we got"
let mut notifications = Vec::new();

match catch_up {
    Ok(Ok(())) => {}
    Ok(Err(err)) => { return ... Failed, notifications /* empty */ ... }
    Err(_) => {                                              // :961 timeout
        return BackgroundNotificationCollection {
            status: NotificationCollectionStatus::Failed,
            notifications,                                    // still empty — buffered events dropped
            error: Some("notification wake collection timed out".into()),
        };
    }
}
// drain loop that turns buffered events into notifications starts here (:976) — unreachable on timeout
```

As the account workers sync during `catch_up_accounts()` they publish `MessageReceived`/etc. events into the broadcast ring `events` is subscribed to. On timeout those events are never drained into `notifications`. The function even computes `remaining` up front for exactly the "drain what we already have" intent, which the early-return contradicts.

## Failure scenario

Wake with `max_wait_ms = 800`. `catch_up_accounts` connects on cold sockets, ingests two inbound messages (publishing two `MessageReceived` events), and is still finishing group-route resync at 800 ms → the outer `timeout` fires. The two buffered message events are never converted to notifications; the host receives `Failed` with an empty list and shows nothing, even though two real new messages arrived.

## Suggested fix

On timeout, do not early-return. Fall through to the same `try_recv()` drain loop with `drain_until = Instant::now()` (drain only what is already buffered, no further waiting), then classify the result as `NewData`/`NoData` (optionally still noting the timeout in `error`). This surfaces partial results consistent with the function's own `remaining`/`drain_until` design.

---
<sub>Filed from a meticulous multi-agent code-review pass; verified against source (`runtime/mod.rs:940-999`) and deduplicated against existing open/closed issues (distinct from #762 and #740).</sub>

### LINKED ISSUE #909: marmot-app: LeaveGroup / DeclineGroupInvite worker handlers never flush the own-group projection updates they queue — live timeline stays stale and the update is later emitted mis-timed
## Summary

`leave_group_with_audit_context` (`client/mod.rs:708`) and `decline_group_invite` (`client/mod.rs:894`) both call `queue_own_group_system_projection_updates(&effects)`. Because a leave/decline publishes a commit with a non-empty `reports` set, this persists the kind-1210 "member left" system row **and** pushes an `AppProjectionUpdate` into `client.pending_projection_updates`.

`pending_projection_updates` is drained **only** by `publish_client_pending_projection_updates`, which every *other* mutating worker handler calls (`InviteMembers`, `RemoveMembers`, `PromoteAdmin`, `DemoteAdmin`, `SelfDemoteAdmin`, `UpdateGroupProfile`, `UpdateGroupImage`, `UpdateGroupAvatarUrl`, `ReplaceEncryptedMediaBlobEndpoints`). The **`LeaveGroup`** handler (`runtime/account_worker.rs:892-895`) and **`DeclineGroupInvite`** handler (`runtime/account_worker.rs:920-931`) do not. So the queued update is never drained by the leave/decline path.

## Where

- `crates/marmot-app/src/runtime/account_worker.rs:892-895` — `LeaveGroup` handler: calls neither `publish_client_pending_projection_updates` nor `publish_app_runtime_group_state_updated`.
- `crates/marmot-app/src/runtime/account_worker.rs:920-931` — `DeclineGroupInvite` handler: calls `publish_app_runtime_group_state_updated` but not `publish_client_pending_projection_updates`.
- `crates/marmot-app/src/client/mod.rs:708`, `:894` — the queueing sites.
- `crates/marmot-app/src/client/projection.rs:272-274` — `take_pending_projection_updates` (invoked only inside `publish_client_pending_projection_updates`).

I verified `LeaveGroup` and `DeclineGroupInvite` are the **only** mutating group handlers that omit the flush; every sibling handler calls it, which confirms there is no general post-command flush.

## Failure scenario

1. A user with an open `subscribe_timeline_messages` window on group G calls `leave_group(G)`.
2. The kind-1210 "You left" row is written to the durable timeline and `self_membership` is set to `Left`, but no `ProjectionUpdated` (and, for leave, no `GroupStateUpdated`) event is broadcast.
3. The open timeline never receives the live delta; the chat-list / group-state subscribers get no live update for the leave.
4. The stale queued update then rides along and is emitted, **mis-timed**, on the next unrelated mutating command that happens to call `publish_client_pending_projection_updates` (e.g. a later `InviteMembers` on a different group) — surfacing late and attributed to the wrong trigger.

`DeclineGroupInvite` is the milder sibling: it publishes `GroupStateUpdated` (so chat-list/group-state refresh), but still never flushes the queued timeline projection update, so an open timeline for the declined group stays stale.

## Suggested fix

Have the `LeaveGroup` handler call `publish_client_pending_projection_updates(...)` and `publish_app_runtime_group_state_updated(...)` on success (mirroring `RemoveMembers`), and have `DeclineGroupInvite` add the `publish_client_pending_projection_updates(...)` call it currently lacks.

## Severity

MEDIUM — no data loss (the durable row is written), but a live-timeline correctness/UX defect plus a mis-attributed late emission, matching the tier of the other subscription-staleness issues (e.g. #761).

### LINKED ISSUE #931: marmot-app: three account-worker handlers strand queued own-group kind-1210 projection updates — mis-timed live emission (sibling of #909)
## Summary

Three account-worker command handlers persist synthesized own-group kind-1210 group-system rows but never broadcast the corresponding live `AppProjectionUpdate` events. The rows sit in `client.pending_projection_updates` until some later, unrelated command flushes them — so the live timeline stays stale and the update is eventually emitted out of order and misattributed to a different action. This is the same mechanism as #909, which names only the `LeaveGroup` / `DeclineGroupInvite` handlers; these three are additional, distinct sites that act on **live, still-joined** conversations.

## Location

Handlers in `crates/marmot-app/src/runtime/account_worker.rs`:
- `UpdateMessageRetention` — `:761-770` (no publish call at all)
- `ReplaceEncryptedMediaBlobEndpoints` — `:771-788` (publishes `group_state_updated` but not the pending projection updates)
- `RetryGroupConvergence` — handler around `:1184`

The client methods they invoke each queue projection rows:
- `update_message_retention` (`client/mod.rs:898`) → `queue_own_group_system_projection_updates` (`:930`)
- `replace_encrypted_media_blob_endpoints` (`client/mod.rs:934`) → `queue_own_group_system_projection_updates` (`:984`)
- `retry_group_convergence` (`client/mod.rs:1672`) → `queue_own_group_system_projection_updates` (`:1685`)

`queue_own_group_system_projection_updates` persists the rows and pushes them onto `pending_projection_updates`, but the **live** events are only emitted by `publish_client_pending_projection_updates` (which drains `take_pending_projection_updates`, `account_worker.rs:1485-1491`). The eight other mutating handlers call that helper; these three do not. The steady-state command arm (`account_worker.rs:562-578`) only drains `take_pending_convergence_groups()` after each command — never `take_pending_projection_updates()`. So the queued rows linger until a later unrelated command (e.g. `InviteMembers`, `UpdateGroupProfile`, `PromoteAdmin`) flushes them.

## Severity

MEDIUM (correctness / live-UI staleness on active conversations).

## Failure scenario

A user changes a group's disappearing-message timer. `update_message_retention` commits, the engine emits `GroupStateChanged { change: "disappearing_timer_changed" }` (cf. #886), a kind-1210 system row is queued but never broadcast — the live timeline shows nothing. Minutes later the user invites a member; the stale timer-change row is now emitted alongside the invite, out of order and attributed to the wrong action. The same applies to an encrypted-media endpoint-policy replacement and to a manual `RetryGroupConvergence` whose convergence produced membership/system changes.

## Suggested fix

In each of the three handlers, call `publish_client_pending_projection_updates(client, events, account_id_hex, account_label)` on success (as `InviteMembers` / `UpdateGroupProfile` / etc. already do), **or** hoist a single `publish_client_pending_projection_updates(...)` into the steady-state command arm after `handle_account_worker_command`.

- **#909** — same mechanism, but names only `LeaveGroup` / `DeclineGroupInvite`. Those archive/leave the group; these three act on live still-joined groups, so the mis-timed emission is visible in an active timeline. This mirrors the repo's practice of filing distinct-site siblings (e.g. #856 vs #715).
- **#886** — documents the `disappearing_timer_changed` change kind that this bug most visibly strands.

### LINKED ISSUE #944: marmot-app: directory fetch/refresh command methods run blocking SQLCipher directory-cache reads/writes on the async runtime thread (distinct sites from #740/#762/#802)
**Severity:** MEDIUM (blocks a tokio worker thread on SQLCipher file I/O; same class as #740/#762/#802, distinct code paths)
**Component:** `crates/marmot-app`

## Summary

Several `async` directory command methods perform synchronous SQLCipher directory-cache reads and multi-statement writes **inline** on the runtime thread (interleaved with their `.await` network fetch), instead of offloading via `blocking_app_task` / `spawn_blocking` as the crate does elsewhere (e.g. `directory/sync.rs:258`, `:272`). `DirectoryCache` is a `std::sync::Mutex<Connection>` over SQLCipher and `SqliteSharedStorage` is fully synchronous, so these calls run PBKDF2/AES + query work directly on the tokio worker driving the future.

## Location

`crates/marmot-app/src/directory/methods.rs`:

- `fetch_account_relay_list_status_for_account_id` (async fn `:77`): sync cache read `account_relay_list_status_for_account_id(...)` at `:98`, sync multi-statement write `remember_directory_relay_lists(...)` at `:114`, interleaved with the `.await` fetch at `:89`.
- `fetch_current_account_relay_list_status_for_account_id` — same shape.
- `fetch_latest_key_package_for_account_id` — sync reads/writes at `:245` / `:255` / `:272` / `:283`.
- `refresh_user_directory_for_account_id` — sync cache access at `:319` / `:323`.

Contrast the offloaded sync worker in `directory/sync.rs:258/:272`.

## Failure scenario

A slow SQLCipher directory-cache read/write (large cache, disk contention, or PBKDF2 on a cold cache open) executed on the runtime worker stalls every other task on that worker for its duration. This is the exact anti-pattern already tracked for the notification (#740, #762) and subscription-snapshot (#802) paths — these directory *fetch/refresh* methods are additional, separately-reachable sites those issues do not name (#802 explicitly notes such sites are tracked as distinct follow-ups).

- **#740** — `subscribe_notifications`.
- **#762** — wake-driven notification collection.
- **#802** — the four `subscribe_*` snapshot reads.
- **#675** (closed) — the general storage-blocking observation, superseded by the per-site follow-ups.

None name the `directory/methods.rs` fetch/refresh command methods.

## Suggested fix

Offload the synchronous cache reads/writes through `blocking_app_task`, matching `directory/sync.rs`. If #740/#762 are fixed by moving resolver work to a blocking pool generically, ensure these call sites are covered too.

---
<sub>Filed from a meticulous multi-agent code-review pass; verified against source and deduplicated against the existing open/closed issue set.</sub>

erskingardner:
## Summary

One bundle for the marmot-app runtime-responsiveness cluster. Each issue is its own commit so the set stays bisectable; merge will close all nine via `Fixes #N`.

- Flush queued kind-1210 projection updates from leave/decline and the remaining account-worker mutators (`Fixes #909`, `Fixes #931`).
- Move subscribe snapshot reads and notification/wake/directory SQLCipher work off the async/runtime thread (`Fixes #802`, `Fixes #740`, `Fixes #762`, `Fixes #944`).
- Rebuild chat, group-state, and notification subscriptions after broadcast `Lagged` instead of staying stale (`Fixes #761`, `Fixes #740`).
- Keep already-buffered wake events when catch-up times out (`Fixes #863`).
- Run Blossom upload/download HTTP off the exclusive account-worker client so inbound delivery keeps polling (`Fixes #741`).

## Test plan

- [x] `just fast-ci`
- [x] `cargo test -p marmot-app --lib` (656 passed)
- [x] `cargo test -p marmot-app --test relay_runtime app_runtime_declines_pending_invite_by_leaving_and_archiving`
- [ ] GitHub CI `just ci` matrix
- [ ] Smoke a live subscribe + leave/decline and confirm the timeline/chat list updates without a later unrelated command
- [ ] Smoke a media upload while another member is sending and confirm inbound messages still land

## Summary by CodeRabbit

* **New Features**
  * Added recovery for missed notifications and subscription updates after temporary interruptions.
  * Improved synchronization for chats, timelines, group state, and chat lists.
  * Added more reliable encrypted media uploads, downloads, and group image retrieval.
  * Added timely updates for group changes, invitations, membership, and related account activity.

* **Bug Fixes**
  * Improved state consistency after group actions and delayed updates.
  * Reduced interruptions caused by notification or subscription recovery timeouts.
  * Improved reliability when connecting to required development services.
  * Improved handling of background media tasks during shutdown.


scripts/hermes_marmot_dev_setup.sh
```diff
@@ -210,13 +210,42 @@ ensure_uv() {
     command -v uv >/dev/null 2>&1
 }
 
+# Authenticate GitHub HTTPS via extraheader so a failed clone cannot echo a
+# token-bearing URL. Unauthenticated clones of public repos 429 in CI.
+git_with_github_auth() {
+    if [ -n "${GITHUB_TOKEN:-}" ]; then
```
erskingardner, line 216: [P1] Export the Actions token before relying on this branch. `git_with_github_auth` authenticates only when `GITHUB_TOKEN` is present, but the `hermes-installer-auth` job never maps `${{ github.token }}` (or `secrets.GITHUB_TOKEN`) into that environment, and checkout explicitly uses `persist-credentials: false`. The failed job log confirms the setup step environment has no token. This therefore still takes the plain `git` path and retries the same unauthenticated clone that hit HTTP 429. Add a job/step `env: GITHUB_TOKEN: ${{ github.token }}` (with `contents: read` as needed), or pass an explicit token to the helper.

erskingardner (commented):
Request changes (blocking; posted as a comment because the authenticated account authored this PR).

Reviewed exact head `242ac6ae4ffb9e375cbd135684e88ffc39b39a65`.

The responsiveness work is needed, but this head still has correctness and resource-safety gaps: notification lag recovery can replay pre-subscription history as fresh alerts; chat lag reconciliation does not emit removals; the media split permits unbounded detached transfers and buffered results; the notification dedupe set grows without bound; and the directory SQL offload leaves additional async call paths blocking. Those points are already captured in the existing inline discussions, so I did not duplicate them.

I added one new inline blocker on the latest CI-only commit: the workflow never exports `GITHUB_TOKEN`, so the retry helper still performs an unauthenticated clone and does not fix the observed 429.

Please close these paths with targeted lag/concurrency/auth regressions before merge. The new-head CI matrix is still in progress.

crates/marmot-app/src/runtime/subscriptions.rs
```diff
@@ -1671,6 +1764,29 @@ fn app_group_record_fingerprint(group: &AppGroupRecord) -> String {
...
+async fn reconcile_chats_snapshot(
+    updates_tx: &mpsc::Sender<AppGroupRecord>,
+    group_fingerprints: &mut HashMap<String, String>,
+    groups: Vec<AppGroupRecord>,
+) -> bool {
+    let visible_group_ids = groups
+        .iter()
+        .map(|group| group.group_id_hex.clone())
+        .collect::<HashSet<_>>();
+    group_fingerprints.retain(|group_id_hex, _| visible_group_ids.contains(group_id_hex));
```
erskingardner, line 1870: This reconcile misses the removal case that #761 explicitly calls out ("its final removed/left transition"). With `include_archived=false`, a group archived (or left → archived) during the lag window is absent from `visible_groups`, so `retain` prunes its fingerprint and **no update is ever sent** — the subscriber renders the group as still-active until restart. Compare:

- the live path above (`subscribe_chats`, ~L1173-1179): on an archive transition it sends the archived record once so the UI can drop it;
- the sibling `reconcile_chat_list_snapshot` (~L1866-1876): it diffs prior fingerprint ids against the fresh set and sends explicit remove updates.

Suggest diffing `group_fingerprints` keys against the fresh id set and re-reading each missing id via `app.group(...)`: if the record still exists (archived/left), send it like the live path does; if it is gone entirely, drop it (matching the live path's `Ok(None)` handling). Same edge, narrower window, in `subscribe_group_state`'s new lag arm: `Ok(None) => continue` leaves a deleted group permanently stale.

crates/marmot-app/src/notifications.rs
```diff
@@ -1462,6 +1466,55 @@ pub(crate) fn notification_update_from_event(
...
+/// live broadcast ring overflows. Callers still dedupe against already-emitted
+/// `notification_key`s so recovery cannot replay a notification the subscriber
+/// already saw.
+pub(crate) fn recover_notification_updates(
+    app: &MarmotApp,
+) -> Result<Vec<NotificationUpdate>, AppError> {
+    let mut resolver = NotificationResolver::default();
+    let mut updates = Vec::new();
+    for account in app.account_home().accounts()? {
+        let records = app.messages_with_query(
```
erskingardner, line 1483: Recovery has no recency watermark. It scans the newest 64 stored events per account, but `seen_notification_keys` only covers this subscription's lifetime — so after a cold-start lag burst (initial catch-up overflowing the 1024-event ring before this task drains it), messages received and already notified in *previous* sessions are re-emitted with a fresh `timestamp_ms: unix_now_ms()`: up to 64 stale OS notifications per account. Consider capturing `unix_now_ms()` when `subscribe_notifications` starts and skipping records whose `recorded_at`/`received_at` predate it; genuinely-missed in-session messages postdate the watermark, so recovery still covers the intended case. Non-blocking, but worth handling before this ships to hosts that post these straight to the OS.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1719,6 +1735,77 @@ async fn handle_startup_hydration_command(
...
+        respond: oneshot::Sender<Result<MediaDownloadResult, AppError>>,
+        started_at: Instant,
+    },
+    GroupImage {
+        result: Result<Vec<u8>, AppError>,
+        respond: oneshot::Sender<Result<Vec<u8>, AppError>>,
+    },
+}
+
+fn spawn_media_http<T>(
```
erskingardner, line 1941: Worth noting the behavioral change: Blossom transfers were previously serialized by the worker loop; this makes them unbounded-concurrent (unbounded spawn + unbounded done-channel), each buffering whole blobs in memory. User-driven, so likely fine in practice — but a burst of large downloads/uploads now runs all at once. A small semaphore would bound it if this ever becomes visible. Non-blocking. Also expected: media sends can now interleave after later commands (e.g. a text message sent while an upload is in flight lands first) — inherent to fixing #741, just worth a deliberate ack.

erskingardner (commented):
**Requesting changes** (posted as a comment — GitHub does not allow a request-changes review on your own PR).

## Review

Nine fixes, one commit per issue, and the mechanics consistently mirror the established sibling patterns: `blocking_app_task` offloads match `subscribe_messages`/`directory/sync.rs`, the projection flushes match the other mutating handlers, and the media prepare/finish split correctly captures secrets + policy under `&mut client` while `send_media_attachments` re-syncs at finish time. Relay-runtime tests, clippy, and check are green; the Hermes failure was a clone HTTP 429, which the new head commit addresses with authenticated retries.

Requesting changes for one gap in the #761 fix (inline):

- **`subscribe_chats` lag reconcile silently drops removal transitions.** A group archived or left inside the lost broadcast window is absent from `visible_groups`, so `reconcile_chats_snapshot` prunes its fingerprint without emitting anything — the default (`include_archived: false`) subscription keeps showing the group permanently, which is exactly the failure scenario #761 names. The live event path and the `subscribe_chat_list` sibling both handle removals explicitly.

Non-blocking notes (inline):
- Notification lag recovery has no recency watermark, so a cold-start lag burst can re-notify messages already notified in prior sessions.
- Media transfers are now unbounded-concurrent off the account worker.

One question: in `collect_notifications_after_wake`, the catch-up **error** arm still discards already-buffered events while the timeout arm now drains and projects them (#863). Intentional? The "surface what we already got" argument seems to apply equally to a partial catch-up that errored.

Nice-to-have but out of scope here: `subscribe_chats` and `subscribe_chat_list` now carry near-identical snapshot/reconcile logic; not asking for a refactor in this PR.

crates/marmot-app/src/directory/methods.rs
```diff
@@ -104,7 +104,12 @@ impl MarmotApp {
         let selection = fresh_relay_list_status_from_records(&account_id_hex, records, freshness);
         let mut status = selection.value;
         if !observed_nip65 || !observed_inbox {
-            let cached = self.account_relay_list_status_for_account_id(&account_id_hex)?;
+            let app = self.clone();
+            let account_id = account_id_hex.clone();
+            let cached = blocking_app_task(move || {
+                app.account_relay_list_status_for_account_id(&account_id)
+            })
+            .await?;
```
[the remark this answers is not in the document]
erskingardner: Done in 957a515c. The remaining async directory SQL paths (`fetch_current_user_profile_for_account_id`, `refresh_directory_entry_for_account_id`, `publish_account_follow_list`, `fetch_current_follow_list_for_account_id`) now go through `blocking_app_task`.

crates/marmot-app/src/notifications.rs
```diff
@@ -62,6 +62,10 @@ const LEGACY_PUSH_OWNER_PROOF_EVENT_KIND: u16 = 450;
 const PUSH_RECORD_DOMAIN: &str = "marmot-push-token-record-v1";
 const PUSH_REMOVAL_DOMAIN: &str = "marmot-push-token-removal-v1";
 const NOTIFICATION_VERSION_TAG: &str = "v";
+/// Bounded recent-history window used to rebuild missed notifications after
+/// broadcast lag. Large enough to cover a slow consumer, small enough that
+/// recovery stays a snapshot rather than a full-history hydrate.
+const NOTIFICATION_LAG_RECOVERY_LIMIT: usize = 64;
```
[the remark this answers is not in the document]
erskingardner: Done in 957a515c. `subscribe_notifications` now captures a unix-seconds watermark at subscribe time and `recover_notification_updates` skips rows strictly before it. Same-second in-session rows stay eligible.

crates/marmot-app/src/notifications.rs
```diff
@@ -1462,6 +1466,55 @@ pub(crate) fn notification_update_from_event(
...
+                    group_id: cgka_traits::GroupId::new(group_id),
+                    source_epoch: record.source_epoch.unwrap_or(0),
+                    retention: record.retention,
+                    plaintext: record.plaintext,
+                    kind: record.kind,
+                    tags: record.tags,
+                    recorded_at: record.recorded_at,
+                    received_at: record.received_at,
+                },
+            });
```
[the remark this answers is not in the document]
erskingardner: Done in 957a515c. Recovery now fills `sender_display_name` via `app.display_name_for_account_id`.

crates/marmot-app/src/runtime/subscriptions.rs
```diff
@@ -1501,15 +1567,24 @@ impl MarmotAppRuntime {
...
+                        let app_for_lookup = app.clone();
+                        match blocking_app_task(move || {
+                            notifications::notification_update_from_event(&app_for_lookup, &event)
+                        })
+                        .await
+                        {
                             Ok(Some(update)) => {
+                                if !seen_notification_keys.insert(update.notification_key.clone()) {
+                                    continue;
+                                }
```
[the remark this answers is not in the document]
erskingardner: Done in 957a515c. The live/recovery notification seen-set now uses `MessageSubscriptionSeenIds` with `MESSAGE_SUBSCRIPTION_SEEN_ID_LIMIT`.

crates/marmot-app/src/runtime/subscriptions.rs
```diff
@@ -1671,6 +1764,29 @@ fn app_group_record_fingerprint(group: &AppGroupRecord) -> String {
...
+        if group_fingerprints.get(&group.group_id_hex) == Some(&fingerprint) {
+            continue;
+        }
+        group_fingerprints.insert(group.group_id_hex.clone(), fingerprint);
+        if updates_tx.send(group).await.is_err() {
+            return false;
+        }
+    }
+    true
+}
```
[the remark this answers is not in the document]
erskingardner: Done in 957a515c. Lag recovery now re-reads dropped ids with `app.group(...)` in the same blocking task and `reconcile_chats_snapshot` sends those archived/left records, matching the live path.

scripts/hermes_marmot_dev_setup.sh
erskingardner: Done in 957a515c. The Hermes setup steps in `ci.yml` and `wn-agent-binaries.yml` now export `GITHUB_TOKEN: ${{ github.token }}` so `git_with_github_auth` can authenticate.

crates/marmot-app/src/runtime/subscriptions.rs
erskingardner: Done in 957a515c. Lag recovery now diffs prior fingerprint ids against the fresh visible set, re-reads missing ids via `app.group(...)`, and sends archived/left records. Fully gone groups stay silent (`Ok(None)`). `subscribe_group_state` now emits the last record with `archived=true` when lookup returns `None`.

crates/marmot-app/src/notifications.rs
erskingardner: Done in 957a515c. Recovery now takes a per-subscription unix-seconds watermark (`recorded_at`/`received_at` units) and skips rows strictly before it. Same-second in-session rows stay eligible. Added `recovery_skips_rows_before_the_subscription_watermark`.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Agreed this was worth bounding before hosts ship it. 957a515c caps in-flight media HTTP at 4 via a semaphore and drops the completion (caller sees `TransportClosed`) if the runtime is already stopping, so shutdown does not finish uploads on a closed worker.

scripts/hermes_marmot_dev_setup.sh
```diff
@@ -210,13 +210,42 @@ ensure_uv() {
...
+    local delay=8
+    while [ "$attempt" -le "$max_attempts" ]; do
+        rm -rf "$dest"
+        if git_with_github_auth clone "$hermes_url" "$dest"; then
+            return 0
+        fi
+        echo "warning: hermes clone failed (attempt ${attempt}/${max_attempts}); retrying in ${delay}s" >&2
+        attempt=$((attempt + 1))
+        sleep "$delay"
+        delay=$((delay * 2))
```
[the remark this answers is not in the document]
erskingardner: Done in 957a515c. The retry warning and sleep now run only when `attempt < max_attempts`.

scripts/hermes_marmot_dev_setup.sh
```diff
@@ -210,13 +210,42 @@ ensure_uv() {
...
 install_hermes() {
     if [ ! -d "$hermes_repo/.git" ]; then
-        git clone "$hermes_url" "$hermes_repo"
+        clone_hermes_repo "$hermes_repo"
     fi
 
     if [ -n "$hermes_ref" ]; then
-        git -C "$hermes_repo" fetch origin "$hermes_ref" || true
+        git_with_github_auth -C "$hermes_repo" fetch origin "$hermes_ref" || true
         git -C "$hermes_repo" checkout "$hermes_ref"
```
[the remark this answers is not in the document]
erskingardner: Done in 957a515c. Fetch is fail-closed (`|| true` removed) and checkout is `git checkout --detach FETCH_HEAD` so a failed fetch cannot install a stale local ref.

erskingardner:
Addressed the review threads in 957a515c.

On the wake catch-up **error** arm: not intentional. It now drains already-buffered events (and recovers if the ring lagged) the same way the timeout arm does, returning `NewData` when anything projected and `Failed` with the original error only when the drain is empty.

Also closed the remaining first-review blockers: notification watermark + bounded seen-set, remaining directory SQL offload, media HTTP concurrency/shutdown, Hermes `GITHUB_TOKEN` export, and the nits (tracing method, recovery display name, UniFFI comment, wake `Lagged` recovery).

`just fast-ci` and `cargo test -p marmot-app --lib` are green on this commit.

scripts/hermes_marmot_dev_setup.sh
erskingardner: [P1] Still blocking on exact head `957a515c`: the token is now exported, but both current Hermes jobs fail in the clone step. The log shows `GITHUB_TOKEN: ***` followed by `fatal: could not read Username for https://github.com`, so `AUTHORIZATION: bearer ...` is not authenticating Git smart HTTP. Use the GitHub-compatible Basic extraheader (`x-access-token:<token>` encoded as Basic) or a non-interactive credential helper, then require both Hermes jobs to pass.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: [P1] The semaphore does not yet bound the resource exposure. It is acquired inside an already-detached task, so an arbitrary number of queued jobs still retain their owned `work` futures (including upload plaintext). A completed task then sends a potentially 512 MiB download into the unbounded done channel and releases its permit immediately, allowing the next queued result to accumulate before the account worker drains anything. The lifecycle checks also do not cancel an active `work.await`; they only discard its result after the transfer finishes. Acquire/reserve capacity before spawning and keep the bound through completion consumption, or use a bounded worker-owned work/result queue that is cancelled and reaped with the account worker.

crates/marmot-app/src/notifications.rs
erskingardner: [P1] The watermark still admits stale history because this comparison takes `max(received_at, recorded_at)`. `recorded_at` is the sender-authenticated inner timestamp and is explicitly not clamped, whereas `received_at` is the local observation time. A clock-skewed or malicious future-dated old row therefore remains above every later subscription watermark (and in the newest-64 `recorded_at` query window), so each fresh lagging subscription can re-post it as a current OS alert. The subscription boundary should use local `received_at` only; add a regression with old `received_at` plus future `recorded_at`.

crates/marmot-app/src/runtime/subscriptions.rs
erskingardner: [P1] The acknowledged fully-gone case is still the deletion branch of #761. When `app.group(...)` returns `None`, nothing enters `removed_records`; `reconcile_chats_snapshot` then drops the fingerprint without emitting any update. `DeleteGroupLocal` explicitly publishes `GroupStateUpdated` after wiping the projection, so both the live path and a lag recovery leave `RuntimeChatsSubscription` consumers holding the deleted row indefinitely. Preserve the prior record and emit an archived tombstone as `subscribe_group_state` now does, or add a typed removal update; cover a local delete in the subscription regression.

erskingardner (commented):
Request changes (blocking; posted as a comment because the authenticated account authored this PR).

Re-reviewed exact head `957a515c4137248b6909d1973e03b7c9f5014e80` and the full `242ac6ae..957a515c` corrective delta.

The follow-up does close several first-pass gaps: remaining directory cache operations are offloaded, wake catch-up errors preserve partial notifications, notification dedupe is bounded, group-state emits a missing-row tombstone, token environments are wired, and the ref fetch now fails closed.

Four blockers remain, documented in the existing threads: the Git auth scheme fails both exact-head Hermes jobs; the media semaphore is acquired after unbounded task/payload creation and releases before unbounded completion consumption; notification recovery compares its local subscription watermark against an unclamped sender timestamp; and fully deleted chats still produce no removal for `RuntimeChatsSubscription`.

The Rust/runtime checks are passing at this head, with Simulator and vectors still running, but the two Hermes checks are red. Targeted regressions are still needed for local-delete recovery and bounded media queue/shutdown behavior.

scripts/hermes_marmot_dev_setup.sh
erskingardner: Fixed in 6b0ce164. GitHub HTTPS now uses a host-scoped Basic x-access-token header through GIT_CONFIG_* with prompting disabled. The dev-script regression intercepts both clone and fetch, verifies the exact scoped header, and confirms the raw token is absent; the full script suite is green.

crates/marmot-app/src/runtime/subscriptions.rs
erskingardner: Fixed in 6b0ce164. Chat subscriptions now retain the last AppGroupRecord alongside its fingerprint. A fully deleted group emits that record as an archived tombstone in both the live lookup-missing path and lag reconciliation. The local-delete relay-runtime regression now requires tombstones for two deleted chats.

crates/marmot-app/src/notifications.rs
erskingardner: Fixed in 6b0ce164. The recovery watermark now compares only the locally observed received_at. The regression uses an old received_at with recorded_at = u64::MAX and proves the sender-controlled future timestamp cannot revive stale history.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in 6b0ce164. Capacity is reserved before media preparation/spawn, the owned permit travels with the completion until the account worker consumes it, and a worker-lifetime channel cancels active HTTP futures on every worker exit. Unit tests pin both completion-channel bounding and shutdown cancellation.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1879,6 +1902,114 @@ async fn handle_startup_hydration_command(
...
+    });
+}
+
+fn reserve_media_http(media_http: &MediaHttpContext) -> Result<OwnedSemaphorePermit, AppError> {
+    media_http
+        .permits
+        .clone()
+        .try_acquire_owned()
+        .map_err(|_| AppError::BlockingTask("media HTTP capacity exhausted".to_owned()))
+}
```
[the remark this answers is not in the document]
erskingardner: Fixed in 98888c80. Exhausted media capacity now returns the existing retryable AppError::AccountWorkerBusy, and the focused capacity test asserts that typed backpressure while the permit remains held through completion consumption.

integrations/hermes/marmot/tests/test_dev_scripts.sh
```diff
@@ -9,6 +9,54 @@ dev_root="$tmp_parent/nested/hermes-marmot-test"
...
+auth_fake_bin="$tmp_parent/auth-fake-bin"
+auth_capture="$tmp_parent/github-auth-capture"
+mkdir -p "$auth_fake_bin"
+cat >"$auth_fake_bin/git" <<'SCRIPT'
+#!/usr/bin/env bash
+set -euo pipefail
+
+capture_auth() {
+    printf '%s\n%s\n' "${GIT_CONFIG_KEY_0:-}" "${GIT_CONFIG_VALUE_0:-}" >>"$AUTH_CAPTURE"
+}
```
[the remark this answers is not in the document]
erskingardner: Fixed in 98888c80. The fake Git harness now records GIT_CONFIG_COUNT alongside the scoped key and Basic value, and asserts count=1 exactly twice across clone and fetch. The full dev-script suite passes.

erskingardner:
Follow-up complete on `98888c80`.

- Addressed and resolved all rereview threads, including typed retryable media backpressure and the Hermes `GIT_CONFIG_COUNT=1` regression assertion.
- Merged current `origin/master` into the branch before the final follow-up.
- Local validation is green: `just fast-ci`, focused media-capacity test, full Hermes dev-script suite, and the CI-equivalent 97-test relay-runtime command.
- GitHub exact-head result: 25 successful checks, the expected PR-only release job skipped, zero failures, and zero unresolved review threads.

erskingardner (commented):
## Re-review: recommend merge

(Posted as a comment — GitHub does not allow approving your own PR.)

Verified each thread against `957a515c` / `6b0ce164` / `98888c80` at head `98888c80`:

- **Removal transitions after lag (#761 gap)** — `subscribe_chats` now diffs prior records against the fresh set and sends the still-extant record or a synthesized `archived=true` tombstone; the live `Ok(None)` arm tombstones too, and `subscribe_group_state` gained `emit_missing_group_state`. The relay-runtime test asserting locally-deleted groups arrive as archived on a live subscription covers the exact scenario I flagged.
- **Recovery watermark** — subscribe-time `unix_now_seconds()` watermark closes the cross-session re-notify storm, and switching the comparison to `received_at`-only (sender-controlled `recorded_at` must not revive stale history) with the `u64::MAX` regression test is the right hardening. Bounding the seen-set with the existing `MessageSubscriptionSeenIds` LRU matches the `subscribe_messages` precedent.
- **Media backpressure** — reserve-at-command-time with `try_acquire`, permit held until the completion is consumed (so the unbounded channel is effectively bounded), `watch`-based cancellation of in-flight HTTP on worker exit, and capacity exhaustion surfaced as typed, retryable `AccountWorkerBusy` through uniffi. The two unit tests (reservation-until-consumed, cancellation-releases-capacity) pin the semantics.
- **Wake error arm** — answered; the error arm now drains buffered events like the timeout arm.
- Hermes script fixes (exported token, fail-closed fetch, `--detach FETCH_HEAD`) are covered by the new shell test asserting `GIT_CONFIG_COUNT=1` and no raw-token leakage.

CI is green across the board on this head. Residual scope note, non-blocking: lag recovery still rebuilds only message notifications — invites (`GroupJoined`) dropped in a lag window stay missed, now watermark-bounded. Fine as a fast-follow if it ever matters.

LGTM.

PR #1452, opened 2026-08-15
test(convergence): enforce scheduled lane budgets

erskingardner:
## Summary

- add a workflow-owned `observe-step` command that preserves command output while recording wall time, Unix child CPU/RSS, exit status, Nextest case totals, and retry counts
- add `collect-observation` to aggregate private step records with final working-directory and retained-artifact sizes
- enforce the reviewed budgets in the nightly, weekly/manual, and release-hardening workflows, including failure paths
- retain raw observations, budget evaluations, and per-step records as scheduled workflow artifacts
- update the convergence tracking plan and operator documentation with the implemented boundary and remaining release-evidence work

## Why

The lane policies already defined wall-clock, CPU, memory, disk, artifact, retry, and flake limits, but scheduled workflows did not collect real observations or invoke the evaluator. A green scheduled run therefore could not demonstrate that it stayed within the reviewed resource envelope.

This closes that machinery gap. Failed wrapped commands write their private step record before returning failure, and the collector/evaluator run under `always()` so a red lane still leaves diagnostic evidence.

## Scope and measurement boundary

This changes campaign-runner tooling, scheduled workflows, tests, and documentation only. It does not change the convergence engine, app runtime, protocol, or storage behavior.

The observations are deliberately bounded:

- wall and CPU are aggregated across wrapped lane commands
- peak RSS is the maximum recorded child-tree value
- case and retry counts come from Nextest output
- disk and artifact values are final regular-file sizes without following symlinks
- setup/tool installation, upload time, transient Docker disk, Docker-daemon resources, and non-Nextest case notions are not included

A passing budget evaluation is not a correctness or capability-completeness attestation. Producing and reviewing a byte-verified release-hardening evidence bundle remains open in the tracking plan.

## Verification

- `just fast-ci`
- `actionlint .github/workflows/simulator-nightly.yml .github/workflows/convergence-hardening.yml`
- `git diff origin/master...HEAD --check`

## Summary by CodeRabbit

* **New Features**
  * Added automated monitoring for nightly, weekly/manual, and release-hardening campaign runs.
  * Captures execution results, resource usage, test and retry information, artifact sizes, and failure status.
  * Preserves run evidence and produces budget evaluations, including when steps fail.

* **Bug Fixes**
  * Ensures failed workflow steps retain their diagnostic records before collection reports failure.

* **Documentation**
  * Documented scheduled campaign monitoring, evidence locations, budget enforcement, and release-evidence status.


.github/workflows/simulator-nightly.yml
```diff
@@ -79,30 +92,67 @@ jobs:
...
+            --artifact-root target/cgka-conformance-simulator-nightly-reports \
+            --artifact-root target/cgka-distributed-container-evidence \
+            --disk-root target \
+            --output "$CGKA_LANE_EVIDENCE_DIR/observed-usage.v1.json"
+
+      - name: Enforce nightly lane budget
+        if: always()
+        run: |
+          cargo run -p convergence-campaign-runner --bin cgka-distributed-campaign --locked -- \
+            check-budget nightly "$CGKA_LANE_EVIDENCE_DIR/observed-usage.v1.json" \
```
erskingardner, line 155: The nightly lane's `max_wall_clock_seconds` is 7200 (`lanes/nightly.v1.json`) — exactly equal to this job's `timeout-minutes: 120`. Because checkout, toolchain install, cache restore, collect, and upload also consume those 120 minutes, the wrapped-step wall sum can never reach 7200s, so the evaluator's wall-clock dimension is dead in this lane. Worse, the failure mode that budget exists to catch — a lane that overruns — ends in a GitHub timeout kill, which cancels the job and skips the `if: always()` collect/enforce/upload steps: the one red-lane path that retains no observation evidence at all. Consider raising the timeout above budget + overhead (e.g. 150m) so the evaluator can actually trip and the evidence path still runs, or document that wall-clock is deliberately enforced by the job timeout rather than the budget.

.github/workflows/convergence-hardening.yml
```diff
@@ -84,7 +111,52 @@ jobs:
...
+            --artifact-root target/cgka-adversarial-reliability-ci \
+            --campaign-manifest "$DISTRIBUTED_MANIFEST" \
+            --disk-root target \
+            --output "$CGKA_LANE_EVIDENCE_DIR/observed-usage.v1.json"
+
+      - name: Enforce release-hardening lane budget
+        if: always() && github.event_name == 'workflow_dispatch' && inputs.lane == 'release_hardening'
+        run: |
+          cargo run -p convergence-campaign-runner --bin cgka-distributed-campaign --locked -- \
+            check-budget release_hardening "$CGKA_LANE_EVIDENCE_DIR/observed-usage.v1.json" \
```
erskingardner, line 163: Same shadowing as the nightly lane: `release_hardening`'s `max_wall_clock_seconds` is 28800 = this job's `timeout-minutes: 480`, so the wall dimension can never trip before GitHub kills the job — and a timeout kill skips the `always()` evidence steps. (`weekly_manual` is fine: 21600s < 480m.) Consider bumping the timeout (e.g. ~540m) when `inputs.lane == 'release_hardening'`, or documenting the deliberate timeout-as-wall-enforcement choice.

crates/convergence-campaign-runner/src/lane_observation.rs
```diff
@@ -0,0 +1,670 @@
...
+            "lane_step_command",
+            "lane step observation requires a command argv",
+        ));
+    };
+    let started = Instant::now();
+    let mut child = Command::new(program);
+    child
+        .args(arguments)
+        .env("NEXTEST_EXPERIMENTAL_LIBTEST_JSON", "1")
+        .env("NEXTEST_MESSAGE_FORMAT", "libtest-json-plus")
```
erskingardner, line 209: This couples budget enforcement to two nextest output contracts at once: the experimental `libtest-json-plus` suite summaries (case counts) and the human `TRY N PASS/FAIL` status lines on stderr (flake counts). Both scheduled workflows install nextest unpinned via `taiki-e/install-action` (latest release). If a nextest upgrade changes the JSON schema, `executed_cases` drops and the lane goes red at check-budget (fail-closed, noisy but safe); but if the `TRY` line format changes, flake detection silently goes dead while the lane stays green — fail-open for `max_flake_retries` / `max_flake_rate_basis_points`. I verified the current formats locally (0.9.104: JSON on stdout, TRY lines on stderr, ANSI-stripped parsing works under `CARGO_TERM_COLOR=always`), so this is correct today. Consider pinning the nextest version in the scheduled workflows, or adding a small format smoke assertion, so an upgrade can't silently disable flake enforcement.

.github/workflows/convergence-hardening.yml
```diff
@@ -84,7 +111,52 @@ jobs:
...
+        if: always() && github.event_name == 'workflow_dispatch' && inputs.lane == 'release_hardening'
+        env:
+          DISTRIBUTED_MANIFEST: ${{ inputs.distributed_manifest }}
+        run: |
+          cargo run -p convergence-campaign-runner --bin cgka-distributed-campaign --locked -- \
+            collect-observation \
+            --step-dir "$CGKA_LANE_EVIDENCE_DIR/steps" \
+            --artifact-root target/cgka-weekly-reliability \
+            --artifact-root target/cgka-adversarial-reliability-ci \
+            --campaign-manifest "$DISTRIBUTED_MANIFEST" \
```
erskingardner, line 155: `collect-observation` adds this operator-supplied manifest's `output_dir` to the artifact roots, and `measure_roots` hard-fails on overlapping roots. An operator who sets `output_dir` under `target/cgka-weekly-reliability` or `target/cgka-adversarial-reliability-ci` (plausible naming for a distributed run) gets a red collect step after a fully green multi-hour lane, with no hint why. Fail-closed is the right direction, but the constraint is undocumented — worth a line in the README / distributed-convergence-campaigns.md or in the `distributed_manifest` input description that `output_dir` must not nest under the weekly artifact roots.

erskingardner (commented):
Reviewed at ebaa58d. Solid, well-scoped machinery: the observe-step / collect-observation / check-budget wiring matches the compiled lane policies, the failure-path ordering is correct (step record written before the nonzero return; collect and enforce run under `always()`), fs-private modes are pinned by tests, and the docs are unusually honest about the measurement boundary.

Verified beyond CI:
- Ran nextest 0.9.104 locally with the exact env the observer forces: `libtest-json-plus` writes suite/test JSON to stdout while human output (including `TRY N FAIL/PASS` retry lines and failure details) goes to stderr — so relaying non-structured lines keeps the workflow log useful, and the suite/retry parsing sees real input, including with `CARGO_TERM_COLOR=always` ANSI sequences.
- CI is green (fmt/check/clippy/all test partitions/policy pin; Simulator-and-vectors still running at review time).

Four non-blocking findings inline:
1. Nightly and release-hardening wall-clock budgets exactly equal their job timeouts, so that dimension can never trip — and a timeout kill skips the `always()` evidence steps entirely.
2. Case/flake parsing depends on two nextest output contracts while nextest is installed unpinned; a `TRY`-format change would silently disable flake budgets.
3. Release-hardening collect fails a green lane if the operator manifest's `output_dir` nests under the weekly artifact roots; the constraint is undocumented.
4. Process: the changed workflows only run on schedule/dispatch, so PR CI never exercises them end-to-end. The integration tests plus actionlint cover the mechanics well, but consider a `workflow_dispatch` run of convergence-hardening (`weekly_manual`) from this branch before merge, or consciously accept that the first real validation is the next scheduled run.

Posting as COMMENT rather than approve: nothing here is a defect, but #1 deserves an explicit decision since "enforce the reviewed budgets ... including failure paths" is the PR's stated goal.

.github/workflows/convergence-hardening.yml
```diff
@@ -84,7 +111,52 @@ jobs:
...
+            --disk-root target \
+            --output "$CGKA_LANE_EVIDENCE_DIR/observed-usage.v1.json"
```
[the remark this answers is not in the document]
erskingardner: Fixed in 4f8061f4. Release collection now measures the distributed-container root. Instead of pre-creating an operator-selected manifest output path, the collector treats a missing artifact root as zero bytes while keeping disk roots mandatory; this preserves early-failure evidence without mutating an arbitrary manifest path before execution.

.github/workflows/convergence-hardening.yml
erskingardner: Fixed in 4f8061f4 with a platform-compatible contract. GitHub-hosted jobs have a hard six-hour execution limit, so a 540-minute timeout would still be killed at six hours. The job is pinned to 360 minutes and the reviewed weekly/release wall budgets are now 5h/5.5h, leaving 60m/30m respectively for collection, enforcement, and upload. The policy test and docs pin this headroom.

Issue #574, opened 2026-06-10
Dead SubscribeInbound connections are never detected; each zombie runs catch_up_accounts every 5s forever


erskingardner:
**Severity:** medium  •  **Category:** performance  •  **Confidence:** high

**Location:** `crates/agent-connector/src/lib.rs:996`

## What
After the initial request, stream_inbound_events never polls the read half, so a disconnected gateway is only noticed when a matching event causes write_frame to fail. For a subscription filtered to a quiet account/group, no writes ever happen, so the task lives forever, holding the socket fd and a broadcast receiver and running catch_up_accounts().await every 5s (sending CatchUp to every worker + relay work). A gateway in a crash/reconnect loop accumulates one zombie per reconnect, multiplying catch-up load and leaking fds; fd exhaustion eventually makes listener.accept() error and exits the whole accept loop, killing the control plane.

## Evidence
lib.rs:166-174 keeps reader but never reads it again for SubscribeInbound. 996-1036 selects only on tick / runtime_events.recv() / debug_events.recv(); no read-half EOF arm. 999 runs catch_up_accounts() per tick. serve_socket 1107-1123 propagates accept() errors out of the loop (1108) with no connection cap.

## Suggested fix
Add a select arm reading the retained reader, terminating on EOF/error; share one catch-up driver across subscribers; handle transient accept() errors without exiting serve_socket.

---
<sub>Found by the Fable multi-agent code review and confirmed by adversarial verification (independent refuter + reproducer, with tiebreak on disagreement). Line numbers are approximate; see quoted code for the exact site.</sub>

PR #1454, opened 2026-08-15
fix(agent-connector): drive idle reconciliation by events with adaptive safety nets

### LINKED ISSUE #1380: wn-agent: idle SubscribeInbound drives sustained CPU and SQLite read amplification
## Summary

A stock `wn-agent` v0.9.11 process sustains approximately 94–100% of one CPU core while serving one normal, long-lived `SubscribeInbound` client, even when network traffic is low and the client is otherwise idle.

Over a 20-second observation window, CPU averaged about 94% without long idle gaps. One Tokio worker remained hot. Process I/O counters increased by approximately 54 MB/s of `rchar` and 13,191 read syscalls/s, while physical disk reads remained at zero. RSS was approximately 345 MB. The encrypted session SQLite database was approximately 183 MiB with an approximately 4.2 MiB WAL. These measurements indicate page-cache/database read amplification rather than physical-disk or network load. The tested binary's SHA-256 was `b508f58d9c14e062ef83a54e1edf0a11143feb0680273e6a6ad7b088695dd5c4`.

## Source-confirmed behavior

The current `master` source still defines both the inbound catch-up and invite-policy reconciliation intervals as five seconds ([constants](https://github.com/marmot-protocol/mdk/blob/645f3a9dc66040a17142cef86606d801925b07cb/crates/agent-connector/src/lib.rs#L80-L83)).

`SubscribeInbound` registers with the shared catch-up driver and immediately requests an initial catch-up ([subscription path](https://github.com/marmot-protocol/mdk/blob/645f3a9dc66040a17142cef86606d801925b07cb/crates/agent-connector/src/inbound.rs#L35-L52)). While at least one subscription remains active, the driver invokes `runtime.catch_up_accounts()` every five seconds; calls are serialized, so they do not overlap with each other ([catch-up driver](https://github.com/marmot-protocol/mdk/blob/645f3a9dc66040a17142cef86606d801925b07cb/crates/agent-connector/src/event_projection.rs#L1023-L1084)). Each catch-up first reconciles account workers and then sends a catch-up command to every running account worker ([runtime catch-up](https://github.com/marmot-protocol/mdk/blob/645f3a9dc66040a17142cef86606d801925b07cb/crates/marmot-app/src/runtime/mod.rs#L3888-L3940)).

Separately, the invite-policy worker reconciles on its own five-second timer as well as on relevant runtime events ([worker loop](https://github.com/marmot-protocol/mdk/blob/645f3a9dc66040a17142cef86606d801925b07cb/crates/agent-connector/src/invite_policy.rs#L22-L71)). Its periodic path enumerates every local-signing account and calls `app.groups()` for each account before filtering pending invitations ([candidate enumeration](https://github.com/marmot-protocol/mdk/blob/645f3a9dc66040a17142cef86606d801925b07cb/crates/agent-connector/src/invite_policy.rs#L75-L129)).

The profile evidence does not yet attribute the measured read volume between these two paths, so both should be measured on the large-session fixture before choosing the fix.

## Expected behavior

An idle, healthy inbound subscription should not continuously consume most of a CPU core or repeatedly scan database state proportional to the full retained session. Recovery and invite-policy correctness must be preserved without fixed-rate full-state polling.

## Acceptance criteria

- Profile and separately attribute the five-second inbound catch-up and invite-policy reconciliation paths against a representative large encrypted session database.
- Make steady-state reconciliation event- or cursor-driven where possible. Any safety-net polling should adaptively back off when no relevant state changes, and should reset promptly when work arrives.
- Prevent overlapping or redundant account, group, and message-state scans across reconciliation sources while preserving initial catch-up, broadcast-lag recovery, invite-policy retries, and reconnect correctness.
- Expose privacy-safe telemetry that identifies which reconciliation source is active, its duration/result, and aggregate work such as accounts/groups/rows considered. Telemetry must not include account IDs, group IDs, message IDs, relay URLs, content, or key material.
- Add an `agent-connector` large-session regression harness for one long-lived, otherwise-idle `SubscribeInbound` client. It should bound repeated catch-up/reconciliation passes and database-read amplification over multiple former five-second intervals; use deterministic operation/read counters rather than a flaky wall-clock CPU assertion in normal CI.
- Record an opt-in benchmark/profile result demonstrating that steady-state idle work no longer scales as repeated full-state scans, while a state change still triggers timely reconciliation.

## Related work

#574 fixed dead `SubscribeInbound` connections that accumulated zombie catch-up loops and introduced the shared driver. This report covers the remaining cost of the single shared loop during one legitimate active subscription, so it is not a recurrence of the zombie-connection bug.

#444 bounded full-history replay after broadcast lag. The current observation occurs without a lag event and points to scheduled reconciliation, so it is a separate path.

#675 documented synchronous SQLite/SQLCipher work on Tokio workers. That architectural concern may amplify the hot-worker symptom, but it does not cover the fixed-rate full-state reconciliation and idle read-volume contract reported here.

## Environment

- `wn-agent` v0.9.11, stock build
- Linux
- One normal long-lived `SubscribeInbound` client

erskingardner:
## Summary

Fixes #1380.

A stock `wn-agent` with one long-lived, otherwise-idle `SubscribeInbound` client sustained ~94% CPU and ~54 MB/s of SQLite page-cache reads because **two independent fixed-rate, full-state reconciliation loops** ran every 5 seconds, while the rest of the system (account workers' live `receive_next_delivery` push loop, CLI daemon, UniFFI, post-mutation paths) is event-driven and treats catch-up as a diagnostic/repair seam.

### Root cause attribution (source-level, both 5s paths)

**Inbound catch-up driver** (`InboundCatchUpDriver` → `runtime.catch_up_accounts()` per account worker):
- `cache_current_encrypted_media_epoch_secrets` ran **twice per sync pass**, re-validating every group projected `encrypted_media.required == false` against the signed component — a **full `MlsGroup::load`** (~12 `openmls_values` blob reads + JSON, tree O(members)) per group per pass. Dominant read.
- `refresh_group_routes` also ran **twice per pass** (sync start + checkpoint): ~12 indexed `cgka_*` reads/group.
- Plus one durable `account_state` UPSERT per pass, an outbound-fanout full scan, and a ≥750ms relay drain wait per pass.

**Invite-policy worker** (per local-signing account): full `app.groups()` projection load — unindexed `seen_events` scan+sort up to 16,384 rows, `account_groups` full scan, `account_group_app_components` full scan with hex-decode of every component blob, disband tables twice, plus repeated `account.json` reads.

Both ran against two separate SQLCipher connections to the same file (no shared page cache), synchronously on one worker task — matching the measured profile.

## Changes

**Event-driven with adaptive safety nets** (base 5s → doubling to 60s cap):
- **Catch-up driver**: scheduled passes back off ×2 while the runtime is quiet; any qualifying runtime event (everything except `AccountError`, which a failing pass could self-emit) resets to base and wakes a backed-off pass promptly — but never faster than base cadence under load. Initial subscription catch-ups, broadcast-lag replay, quiesce-on-zero-subscribers (#574), and `Completed` marker semantics are preserved.
- **Invite-policy worker**: `GroupJoined` applies immediately (unchanged); failed applies wake on their own per-candidate retry backoff (`next_due()`) instead of polling; full enumeration runs once at startup, then on the same adaptive safety net.

**Each remaining pass is O(changed), not O(retained state)**:
- New `MarmotApp::pending_group_invites()` backed by a targeted `SELECT group_id_hex, welcomer_account_id_hex FROM account_groups WHERE pending_confirmation=1 AND archived=0` — replaces the full projection load.
- EM-secret warm pass skips authoritative `MlsGroup::load` re-checks at an unchanged group epoch (a commit always advances the epoch); new fallible `try_encrypted_media_for_group` keeps transient errors retryable instead of latching.
- Sync checkpoint skips its second `refresh_group_routes` on zero-delivery, clean-route passes.

**Privacy-safe telemetry**: new `ReconcileTelemetry` counters (passes/outcomes/accounts/candidate rows per source, including a read-amplification gauge that must read 0 rows when idle) + per-source tracing events (`source`, `duration_ms`, `result`, aggregate counts only) per `observability.md`.

## Verification

- **Regression harness** (deterministic operation/row counters, no wall-clock CPU assertions): driver backs off over ~16 former intervals (≤8 passes vs 16 at the old rate), quiesces with no subscribers, resets promptly on activity, filters `AccountError`; invite enumeration backs off with exactly 0 candidate rows on an idle session. EM-warm epoch-skip and checkpoint-skip behavior tests in marmot-app; targeted-query test in storage-sqlite.
- **Opt-in benchmark** `just bench-idle-reconciliation` (release): 8 groups, 7 passes over 10s (vs 40 at the old cadence), `invite_candidate_rows=0`, activity wake gap 0ms — recorded as `MDK_BENCH` lines.
- `just fast-ci` clean (fmt + workspace check + clippy, both feature sets); `agent-connector` 142 passed; `marmot-app` lib 670 passed; `storage-sqlite` 377 passed; tracing audit passes.
- `relay_runtime` integration: 93 passed / 5 failed — **verified identical on clean master** (`encrypted_media_*`, retention system-row, self-removal unread): pre-existing environment failures, not caused by this change.

Docs updated: crate AGENTS.md files, `docs/marmot-architecture/telemetry.md` (new inventory section), `runtime-state-bounds.md` (new per-group map entry).

## Summary by CodeRabbit

* **Improvements**
  * Improved background synchronization responsiveness by reacting sooner to account activity while reducing unnecessary work during idle periods.
  * Improved pending group-invite reconciliation, including faster detection, retry handling, and more targeted processing.
  * Reduced redundant encrypted-media checks and route refreshes during steady-state synchronization, improving efficiency.
  * Added safeguards to keep reconciliation activity bounded and reliable during failures or extended idle periods.

* **Documentation**
  * Added guidance covering synchronization behavior, reconciliation telemetry, and runtime state limits.

### DIFF
--- a/crates/agent-connector/src/event_projection.rs
+++ b/crates/agent-connector/src/event_projection.rs
@@ -1033,7 +1033,10 @@ pub(crate) struct InboundCatchUpSchedule {
 #[derive(Clone)]
 pub(crate) struct InboundCatchUpDriver {
     runtime: MarmotAppRuntime,
-    lock: Arc<AsyncMutex<()>>,
+    /// Serializes scheduled and explicit catch-up passes. `pub(crate)` so the
+    /// regression harness can hold a pass queued behind the lock and prove the
+    /// schedule preserves delayed cadence afterwards (mdk#1380 review).
+    pub(crate) lock: Arc<AsyncMutex<()>>,
     events: broadcast::Sender<InboundCatchUpEvent>,
     pub(crate) started: Arc<AtomicBool>,
     pub(crate) active: Arc<AtomicU64>,
@@ -1111,7 +1114,12 @@ impl InboundCatchUpDriver {
             .base
             .max(Duration::from_millis(1))
             .min(self.schedule.max);
-        let mut last_pass_started = tokio::time::Instant::now();
+        // Deadlines anchor at pass COMPLETION, preserving the old
+        // `MissedTickBehavior::Delay` semantics: a pass that runs long (or
+        // queues behind an explicit request on the serialization lock) inserts
+        // the full interval before the next pass instead of firing
+        // back-to-back (mdk#1380 review).
+        let mut last_pass_completed = tokio::time::Instant::now();
         loop {
             if self.active.load(Ordering::Acquire) == 0 {
                 self.started.store(false, Ordering::Release);
@@ -1126,17 +1134,22 @@ impl InboundCatchUpDriver {
                     break;
                 }
             }
-            let deadline = last_pass_started + interval;
-            let wake =
-                wait_for_catch_up_wake(deadline, last_pass_started, &self.schedule, &mut activity)
-                    .await;
-            last_pass_started = tokio::time::Instant::now();
+            let deadline = last_pass_completed + interval;
+            let wake = wait_for_catch_up_wake(
+                deadline,
+                last_pass_completed,
+                &self.schedule,
+                &mut activity,
+            )
+            .await;
             // A pass that observed activity since the previous pass began —
             // the wake reason itself, plus anything buffered while the pass
             // ran — keeps the next pass at the base interval.
             let observed_activity =
                 matches!(wake, CatchUpWake::Activity) || drain_pending_activity(&mut activity);
-            match self.scheduled_pass(observed_activity).await {
+            let result = self.scheduled_pass(observed_activity).await;
+            last_pass_completed = tokio::time::Instant::now();
+            match result {
                 Ok(()) if observed_activity => {
                     interval = self
                         .schedule
@@ -1154,8 +1167,8 @@ impl InboundCatchUpDriver {
     /// per-pass tracing event (source/duration/result/accounts considered).
     async fn scheduled_pass(&self, observed_activity: bool) -> Result<(), ()> {
         let started = tokio::time::Instant::now();
-        ReconcileTelemetry::bump(&self.telemetry.catch_up_passes_started);
         let _guard = self.lock.lock().await;
+        ReconcileTelemetry::bump(&self.telemetry.catch_up_passes_started);
         let result = self.runtime.catch_up_accounts_reporting().await;
         match &result {
             Ok(summary) => {
--- a/crates/agent-connector/src/event_projection.rs
+++ b/crates/agent-connector/src/event_projection.rs
@@ -1255,21 +1255,21 @@ enum CatchUpWake {
 
 /// Wait until `deadline`, returning early ([`CatchUpWake::Activity`]) when
 /// qualifying runtime activity arrives. Activity noticed less than the base
-/// interval after `last_pass_started` is deferred to that base boundary: a
+/// interval after `last_pass_completed` is deferred to that base boundary: a
 /// busy account stays at the base cadence (never one pass per event), while a
 /// backed-off (quiet) driver notices fresh work promptly (mdk#1380).
 async fn wait_for_catch_up_wake(
     deadline: tokio::time::Instant,
-    last_pass_started: tokio::time::Instant,
+    last_pass_completed: tokio::time::Instant,
     schedule: &InboundCatchUpSchedule,
     activity: &mut broadcast::Receiver<MarmotAppEvent>,
 ) -> CatchUpWake {
     let timer = tokio::time::sleep_until(deadline);
     tokio::pin!(timer);
     let base = schedule.base.max(Duration::from_millis(1));
-    // The cadence floor follows the previous pass start. The scheduled
+    // The cadence floor follows the previous pass completion. The scheduled
     // deadline never precedes it (interval >= base).
-    let cadence_floor = last_pass_started + base;
+    let cadence_floor = last_pass_completed + base;
     loop {
         tokio::select! {
             _ = &mut timer => return CatchUpWake::Timer,
--- a/crates/agent-connector/src/event_projection.rs
+++ b/crates/agent-connector/src/event_projection.rs
@@ -277,9 +277,10 @@ pub(crate) fn media_refs_from_tags(
 use cgka_traits::engine::GroupStateChange;
 use cgka_traits::{GroupId, engine::GroupEvent};
 use marmot_app::{AppError, AppMessageRecord, MarmotAppEvent, MarmotAppRuntime};
+use std::time::Duration;
 use tokio::sync::{Mutex as AsyncMutex, broadcast};
 
-use crate::INBOUND_CATCH_UP_INTERVAL;
+use crate::reconcile_telemetry::{ReconcileSource, ReconcileTelemetry};
 use crate::validation::normalize_hex;
 
 pub(crate) fn control_actor(
@@ -999,6 +1000,36 @@ pub(crate) enum InboundCatchUpEvent {
     Completed,
 }
 
+/// Cadence for the catch-up safety net. Production builds use
+/// [`crate::INBOUND_CATCH_UP_BASE_INTERVAL`] / [`crate::INBOUND_CATCH_UP_MAX_INTERVAL`];
+/// tests inject shorter intervals to exercise the adaptive schedule.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) struct InboundCatchUpSchedule {
+    pub(crate) base: Duration,
+    pub(crate) max: Duration,
+}
+
+/// One shared catch-up driver for every live `SubscribeInbound` subscription
+/// (mdk#574). The driver's scheduled passes are a *safety net*: steady-state
+/// delivery is push-driven through each account worker's live receive loop, so
+/// a pass exists to heal silent notification gaps, not to fetch messages —
+/// which is why the net may back off when nothing is happening (mdk#1380).
+///
+/// Scheduling discipline:
+/// - Subscription initial catch-ups arrive out-of-band through
+///   [`Self::request`] and run promptly (serialization lock coalesces them).
+/// - Scheduled passes start at `base` after the first subscription and double
+///   after every pass that observed no qualifying runtime activity, capped at
+///   `max`.
+/// - Qualifying activity — any runtime event except `AccountError` (errors can
+///   be our own failing passes and must not self-stimulate), or a lagged event
+///   stream — resets the interval to `base`. Activity observed while the
+///   driver was sleeping wakes a pass immediately, but only once the previous
+///   pass is at least `base` old, so a busy account never escalates BEYOND the
+///   base cadence.
+/// - A failed pass keeps backing off (relay outage: the per-worker reconnect
+///   backoff owns transport repair); recovery surfaces as activity events.
+/// - The driver quiesces when the last subscription drops, exactly as before.
 #[derive(Clone)]
 pub(crate) struct InboundCatchUpDriver {
     runtime: MarmotAppRuntime,
@@ -1006,10 +1037,31 @@ pub(crate) struct InboundCatchUpDriver {
     events: broadcast::Sender<InboundCatchUpEvent>,
     pub(crate) started: Arc<AtomicBool>,
     pub(crate) active: Arc<AtomicU64>,
+    schedule: InboundCatchUpSchedule,
+    telemetry: Arc<ReconcileTelemetry>,
+    /// Test-only activity channel override. Production drivers subscribe to the
+    /// runtime event broadcast; tests inject a channel they can publish
+    /// synthetic events into.
+    activity_override: Option<broadcast::Sender<MarmotAppEvent>>,
 }
 
 impl InboundCatchUpDriver {
-    pub(crate) fn new(runtime: MarmotAppRuntime) -> Self {
+    pub(crate) fn new(runtime: MarmotAppRuntime, telemetry: Arc<ReconcileTelemetry>) -> Self {
+        Self::with_schedule(
+            runtime,
+            telemetry,
+            InboundCatchUpSchedule {
+                base: crate::INBOUND_CATCH_UP_BASE_INTERVAL,
+                max: crate::INBOUND_CATCH_UP_MAX_INTERVAL,
+            },
+        )
+    }
+
+    pub(crate) fn with_schedule(
+        runtime: MarmotAppRuntime,
+        telemetry: Arc<ReconcileTelemetry>,
+        schedule: InboundCatchUpSchedule,
+    ) -> Self {
         let (events, _) = broadcast::channel(16);
         Self {
             runtime,
@@ -1017,7 +1069,19 @@ impl InboundCatchUpDriver {
             events,
             started: Arc::new(AtomicBool::new(false)),
             active: Arc::new(AtomicU64::new(0)),
+            schedule,
+            telemetry,
+            activity_override: None,
         }
+    }
+
+    #[cfg(test)]
+    pub(crate) fn override_activity_for_test(
+        mut self,
+        activity: broadcast::Sender<MarmotAppEvent>,
+    ) -> Self {
+        self.activity_override = Some(activity);
+        self
     }
 
     fn spawn(&self) {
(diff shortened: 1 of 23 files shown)


crates/agent-connector/src/invite_policy.rs
```diff
@@ -1,36 +1,89 @@
...
+                    };
+                    safety_deadline = tokio::time::Instant::now() + safety_interval;
+                }
+                _ = retry_sleep, if retry_due.is_some() => {
+                    // A matured retry is just an enumeration whose due gate now
+                    // passes; it does not perturb the safety net's cadence.
+                    let _ = self
+                        .reconcile_pending_invite_policies(&mut retry_state)
+                        .await;
                 }
```
erskingardner, line 158: If `reconcile_pending_invite_policies` returns `None`, `retry_state` is unchanged. A matured `next_due()` is then still in the past, `sleep_until(due)` completes immediately, and this branch wins the `select` forever — full CPU for as long as the storage error lasts (locked/closed account DB is enough).

Do **not** fix this with `tokio::time::sleep(base).await` in this arm: that parks the whole worker and delays `GroupJoined` applies for the backoff. Push matured retry deadlines forward (or keep a separate enumeration-failure deadline) so `next_due()` is in the future, then loop back into the `select`.

Please add a regression that fails `pending_group_invites` while a retry is already due and asserts the worker does not spin.

crates/marmot-app/src/lib.rs
```diff
@@ -3056,6 +3058,32 @@ impl MarmotApp {
...
+                let welcomer = invite
+                    .welcomer_account_id_hex
+                    .as_deref()
+                    .map(|welcomer| welcomer.trim().to_ascii_lowercase())
+                    .map(|welcomer| hex::decode(welcomer).map(MemberId::new))
+                    .transpose()?;
+                Ok(PendingGroupInvite { group_id, welcomer })
+            })
+            .collect()
+    }
```
erskingardner, line 3114: `hex::decode(...)?` inside the `collect()` fails the entire call on one undecodable `group_id_hex` / `welcomer_account_id_hex`. `pending_invite_policy_candidates` then `?`s that error, so one bad row on one account disables invite-policy for every local-signing account and returns `None` from reconcile (the path that can busy-loop when a retry is already due).

Skip the malformed row with the same privacy-safe warning `routing_for` uses for invalid persisted group ids, and still return the valid invites.

erskingardner (commented):
GitHub will not let me formally request changes on my own PR, so this is a comment review. I would otherwise request changes for the invite-policy retry busy-loop below.

The direction is right for #1380: event-driven catch-up/invite work with adaptive safety nets, a targeted pending-invite read, and skipping the idle `MlsGroup::load` / second `refresh_group_routes` are the right cost cuts. CI is green and the compressed-counter harness plus opt-in bench match the issue’s verification bar.

One new failure mode should be fixed before merge. The invite-policy retry arm can busy-loop at full CPU when enumeration fails while a retry is already due — the same class of idle peg this PR exists to remove. Please also skip undecodable pending-invite rows instead of failing the whole connector pass (that error path is what leaves `next_due` stuck in the past).

I would **not** block on rewriting the encrypted-media warm pass into a dirty-set. In-memory iteration plus `group_record` → `storage.get_group` is not the 54 MB/s `MlsGroup::load` profile; the epoch-keyed negative cache is the load that mattered. Residual idle cost (start-of-sync route refresh, the still-unconditional checkpoint `save_state`, two warm walks) is acceptable once the safety net actually backs off.

Non-blocking: the scheduling tests sleep on real time with tight upper bounds; paused Tokio time would make them cheaper and less CI-flaky.

crates/agent-connector/src/event_projection.rs
```diff
@@ -999,56 +1000,200 @@ pub(crate) enum InboundCatchUpEvent {
...
+                    break;
                 }
-                let _ = driver.request().await;
             }
-        });
+            let deadline = last_pass_started + interval;
+            let wake =
+                wait_for_catch_up_wake(deadline, last_pass_started, &self.schedule, &mut activity)
+                    .await;
+            last_pass_started = tokio::time::Instant::now();
```
erskingardner: [P1] Preserve delayed cadence after a slow serialized pass

`last_pass_started` is captured before `scheduled_pass` waits for `self.lock`, and the next deadline is derived from that timestamp. The common subscription path starts an explicit catch-up concurrently; that operation may run for up to the per-account readiness/catch-up deadline. If the scheduled tick queues behind it—or if a scheduled catch-up itself lasts longer than `interval`—the following deadline is already in the past, so the driver immediately launches another full catch-up. The old Tokio interval used `MissedTickBehavior::Delay`, which inserted the interval after a missed/slow tick; this manual loop loses that property and can recreate continuous catch-up work on exactly the large session this PR targets. Anchor the next deadline after the serialized pass completes (or otherwise preserve Delay semantics), and cover an explicit/slow pass that lasts beyond `base` to prove there is no back-to-back invocation.

crates/storage-sqlite/src/account_projection.rs
```diff
@@ -407,6 +418,39 @@ impl SqliteAccountStorage {
...
+    /// rather than the full account projection.
+    pub fn pending_confirmation_group_invites(
+        &self,
+    ) -> StorageResult<Vec<StoredPendingGroupInvite>> {
+        let conn = self.lock()?;
+        let mut statement = conn
+            .prepare(
+                "SELECT group_id_hex, welcomer_account_id_hex
+                 FROM account_groups
+                 WHERE pending_confirmation = 1 AND archived = 0
```
erskingardner, line 437: [P1] Add an index for the pending-invite predicate

`account_groups` has only its `group_id_hex` primary-key index; no migration adds an index on `pending_confirmation`/`archived`. Consequently this predicate must scan every retained group (and `ORDER BY group_id_hex` may make SQLite walk the PK index plus table rows), so it is not O(pending invites) as documented. `invite_candidate_rows_considered` counts only rows returned, which means an idle large account misleadingly reports zero even while the query examines all group rows. Add a partial/covering index for `pending_confirmation = 1 AND archived = 0` and a query-plan or equivalent deterministic read-work regression so the benchmark actually detects this amplification.

erskingardner (commented):
## Blocking review at `7a3440e`

GitHub does not allow the PR author to request changes on their own PR, so this is submitted as a comment, but I do **not** recommend merging this head yet.

The overall direction is right and CI is fully green, but the central idle-work contract is not established:

- The adaptive catch-up loop loses the old delayed-tick behavior when a serialized pass is slow or queued, allowing full catch-ups to run back-to-back (inline).
- The new pending-invite predicate is unindexed, so the claimed O(pending invites) query still scans retained group state while its telemetry reports only returned rows (inline).
- I independently verified the existing CodeRabbit findings: a matured retry plus enumeration failure can tight-loop; encrypted-media warming still scans every projected group twice per subscription sync, contrary to the new O(changed) invariant; and one malformed invite row strands every valid invite in that account.

The benchmark bounds pass counts, not SQLite rows/pages examined, so it cannot catch either retained-group scan. Please address these paths and add deterministic regressions for a slow serialized catch-up, the pending-invite query plan/read work, and the retry-failure loop.

crates/agent-connector/src/invite_policy.rs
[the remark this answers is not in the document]
erskingardner: Fixed in 5b3535fc, but not with an in-arm sleep (per the follow-up review: that would park the worker and delay GroupJoined applies). Instead, a separate enumeration-failure floor — doubling to the schedule max, cleared on any success — pushes the retry wake into the future via max(next_due, floor). The worker never parks; only the failing enumeration's re-attempts are rate-limited. Regression: invite_policy_worker_does_not_spin_when_enumeration_fails_with_a_due_retry (injected failure with an armed retry stays bounded), negative-verified by removing the floor and watching it spin.

crates/marmot-app/src/client/mod.rs
```diff
@@ -3175,16 +3190,53 @@ impl AppClient {
...
+        // group set.
+        let live: HashSet<&str> = self
+            .state
+            .groups
+            .iter()
+            .map(|group| group.group_id_hex.as_str())
+            .collect();
+        self.encrypted_media_not_required_epochs
+            .retain(|group_id_hex, _| live.contains(group_id_hex.as_str()));
         for group in &self.state.groups {
```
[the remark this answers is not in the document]
erskingardner: Not rewriting this into a dirty-set, per the human reviewer's explicit non-blocking call: in-memory iteration plus one small group_record read per group is not the 54 MB/s MlsGroup::load profile this PR eliminated (the epoch-keyed negative cache was the load that mattered). The conflicting AGENTS.md wording is fixed in 5b3535fc to state the actual invariant: no full MlsGroup::load sweeps, no full account-projection loads; cheap per-group probes are acceptable.

crates/marmot-app/src/lib.rs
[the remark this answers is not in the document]
erskingardner: Fixed in 5b3535fc: undecodable group_id_hex / welcomer_account_id_hex rows are skipped with a privacy-safe warning (the same policy routing_for applies to malformed persisted routes), and valid invites are still returned. Regression: pending_group_invites_skips_malformed_rows in marmot-app.

crates/agent-connector/src/invite_policy.rs
erskingardner: Fixed in 5b3535fc exactly along these lines — no in-arm sleep. A separate enumeration-failure floor (doubling to the schedule max, cleared on success) pushes matured retry deadlines into the future via max(next_due, floor), and safety enumerations back off on the same floor while failing. GroupJoined applies stay fully live. Regression added as requested: invite_policy_worker_does_not_spin_when_enumeration_fails_with_a_due_retry — bounded attempts with an armed retry and injected enumeration failure, negative-verified by removing the floor.

crates/marmot-app/src/lib.rs
erskingardner: Fixed in 5b3535fc: malformed rows are skipped with the same privacy-safe warning policy as routing_for, and the valid invites are still returned. Regression: pending_group_invites_skips_malformed_rows.

crates/agent-connector/src/event_projection.rs
erskingardner: Fixed in 5b3535fc: the next deadline now anchors at pass completion, restoring the old MissedTickBehavior::Delay semantics — a pass that queues behind the serialization lock (e.g. behind an explicit subscription catch-up) or runs longer than the interval inserts the full interval before the next pass. Regression: inbound_catch_up_driver_preserves_delayed_cadence_after_a_slow_pass holds a pass queued behind the lock past 4x base and asserts exactly one pass followed by normal cadence. Negative-verified by re-anchoring at pass start (test fails).

crates/storage-sqlite/src/account_projection.rs
erskingardner: Fixed in 5b3535fc: migration 0049 adds a partial covering index idx_account_groups_pending_invites ON account_groups (group_id_hex, welcomer_account_id_hex) WHERE pending_confirmation = 1 AND archived = 0. Both selected columns are covered, and the leading column satisfies ORDER BY group_id_hex without a sort, so an idle enumeration examines O(pending invites) index entries. Regression: pending_confirmation_group_invites_uses_the_partial_covering_index pins the EXPLAIN QUERY PLAN (note: the plan shows SCAN of the *partial* index, which only visits matching rows — the regression rejects full-table / primary-key-index scans).

erskingardner:
## Review round addressed — `5b3535fc`

All blocking threads are addressed with negative-verified regressions (each test fails when its fix is removed):

| Thread | Fix |
| --- | --- |
| Retry busy-loop on failed enumeration (CodeRabbit + @erskingardner) | Separate enumeration-failure floor pushes the retry wake into the future (`max(next_due, floor)`); the worker never parks, so `GroupJoined` applies stay live. Safety enumerations back off on the same floor. |
| Malformed pending-invite row strands all invites (CodeRabbit + @erskingardner) | `pending_group_invites` skips undecodable rows with a privacy-safe warning (the `routing_for` policy); valid invites still returned. |
| Lost delayed-tick semantics after a slow/queued pass (@erskingardner P1) | Deadline anchors at pass completion (`MissedTickBehavior::Delay` preserved); a lock-held slow pass fires once, then resumes cadence. |
| Unindexed pending-invite predicate (@erskingardner P1) | Migration `0049` adds a partial covering index `idx_account_groups_pending_invites (group_id_hex, welcomer_account_id_hex) WHERE pending_confirmation = 1 AND archived = 0`; EXPLAIN query-plan regression pins it. |
| EM warm pass group scan (CodeRabbit) | Not rewritten into a dirty-set per @erskingardner's explicit non-blocking call; `marmot-app/AGENTS.md` wording now states the actual invariant (no full `MlsGroup::load` sweeps / full projection loads; cheap per-group probes acceptable). |

On the nitpick (paused Tokio time): evaluated and reverted. `start_paused` requires the `current_thread` flavor, and on this tokio version `time::advance()` does not cascade through timers that a task re-registers when polled (a free-running sleep loop fires once per `advance` call), so driving the driver/worker loops deterministically would require contorted step-advance+yield scaffolding for marginal gain. The two backoff tests keep real-time sleeps with ~2× counter margins; the new regressions are likewise counter-based.

Verification: `just fast-ci` clean; `agent-connector` 144 passed; `marmot-app` lib 671 passed; `storage-sqlite` 378 passed; tracing audit green; `just bench-idle-reconciliation` still passes (`scheduled_passes=5`, `candidate_rows=0`, `wake_gap_ms=0`); `relay_runtime` integration remains at exact parity with master (93 passed / the same 5 pre-existing environment failures).

erskingardner (commented):
Re-review at `5b3535f`. GitHub still will not let me approve my own PR; I would otherwise approve.

The previous blocking findings are actually fixed, not just claimed:

- **Retry spin:** enumeration-failure floor (`max(next_due, floor)`, doubling to schedule max, cleared on success) keeps the worker in the `select` so `GroupJoined` stays live. The injected-failure regression stays well below a spin bound.
- **Malformed invite rows:** skipped with a privacy-safe warning; valid invites still returned. Regression covers both a bad group id and a bad welcomer.
- **Delayed cadence:** next deadline anchors at pass *completion*, and `catch_up_passes_started` is bumped after the serialization lock, so a queued/slow pass cannot rerun back-to-back. Regression holds the lock past 4× base.
- **Pending-invite scan:** migration 0049 adds the partial covering index; the EXPLAIN QUERY PLAN test pins it.

CI is green. Residual nits only: `wait_for_catch_up_wake` still names/comments the cadence floor as pass *start* even though callers now pass completion, and `invite_enumeration_successes_before_failure` lives on the production struct behind `cfg!(test)` rather than `#[cfg(test)]`. Neither should block merge.

crates/agent-connector/src/invite_policy.rs
erskingardner: Confirmed fixed at `5b3535f`. The enumeration-failure floor pushes a matured retry wake into the future without parking the worker, and `invite_policy_worker_does_not_spin_when_enumeration_fails_with_a_due_retry` covers the case.

crates/marmot-app/src/lib.rs
erskingardner: Confirmed fixed at `5b3535f`. Undecodable rows are skipped with a privacy-safe warning and valid invites are still returned; `pending_group_invites_skips_malformed_rows` covers both a bad group id and a bad welcomer.

crates/agent-connector/src/invite_policy.rs
```diff
@@ -143,9 +334,18 @@ impl AgentConnector {
...
-            Ok(()) => retry_state.clear(&candidate.key),
+            Ok(()) => {
+                retry_state.clear(&candidate.key);
+                ReconcileTelemetry::bump(&self.reconcile_telemetry.invite_policy_applied);
+            }
             Err(err) => {
-                let (attempts, retry_delay) = retry_state.record_failure(candidate.key, now);
+                let (attempts, retry_delay) = retry_state.record_failure_with(
+                    candidate.key,
+                    now,
```
erskingardner: **[P1] Start the retry delay after the failed attempt returns**

`now` is captured before awaiting `apply_invite_policy` (and once for the entire candidate loop). In production, accept can wait up to 10 seconds and decline up to 2 minutes, while the initial retry delay is 5 seconds. If either operation eventually times out or otherwise fails after that delay, `next_retry_at = now + retry_delay` is already in the past, so the new exact retry wake immediately enumerates and invokes the operation again; later candidates inherit an even older timestamp. This defeats the bounded-backoff contract and is particularly risky after a worker-response timeout, when the command may still be completing. Record the failure with a fresh `Instant::now()` after the await returns (and use a fresh time for each candidate’s due check), then add a delayed-failure regression proving that a full retry delay follows completion.

erskingardner (commented):
## Rereview

I verified the fixes at the current head: enumeration failures now impose a retry floor, malformed pending-invite rows no longer strand valid rows, scheduled catch-ups anchor their next deadline after completion, and migration 0049 adds the targeted partial covering index with a query-plan regression. The exact-head CI matrix is green.

One merge-blocking retry-timing issue remains inline. Because this review is being submitted from the PR author account, GitHub does not allow a formal “Request changes” event, so I am recording the blocking verdict as a comment review.

crates/agent-connector/src/invite_policy.rs
erskingardner: Fixed in 4c159dcc: the failure is now recorded with a fresh `Instant::now()` after the apply await returns, and the reconciliation loop takes a fresh due-check time per candidate instead of sharing one captured before the loop. Regression added as requested: `invite_policy_retry_delay_starts_after_the_failed_attempt_returns` stretches the first apply attempt to 800ms (beyond the 200ms initial retry delay) and asserts the full delay follows completion — negative-verified (attempt-start anchoring retries after 3µs). Also folded in the two nits: `wait_for_catch_up_wake` now names the anchor `last_pass_completed`, and the test hooks moved behind `#[cfg(test)]` into `test_support::ConnectorTestHooks` (shared `Arc`), so the production struct carries no test surface.

erskingardner: Confirmed fixed at `4c159dc`. Failure is recorded after the apply await returns, due-checks are per-candidate, and `invite_policy_retry_delay_starts_after_the_failed_attempt_returns` would fail if the backoff were still anchored at attempt start.

erskingardner (commented):
Re-review at `4c159dc`. GitHub will not let me approve my own PR; I would otherwise approve.

The remaining P1 is fixed: `apply_invite_policy_candidate` records `next_retry_at` with a fresh `Instant::now()` *after* the apply await returns, and the candidate loop re-checks due time per item. That is the right model when accept/decline can outlive the 5s initial backoff. `invite_policy_retry_delay_starts_after_the_failed_attempt_returns` covers it (800ms stretched attempt, 200ms retry delay, gap must be ≥150ms).

Earlier blocking items remain intact: enumeration-failure floor, malformed-row skip, completion-anchored catch-up cadence, partial covering index. Test hooks are now `#[cfg(test)]` on `ConnectorTestHooks`. Catch-up docs/param match completion anchoring. CI is green.

No new blocking issues.

PR #1455, opened 2026-08-15
fix(simulator): keep process capsules participant-owned and honor payload-count oracles

### LINKED ISSUE #1450: Simulator nightly failed (2026-08-15)
The simulator nightly failed. Run: https://github.com/marmot-protocol/mdk/actions/runs/31863028300

erskingardner:
## Summary
- Process-layer failure capsules are now participant-attributed only. Quiescence timeouts and relay waits return an orchestrator error instead of an `app_process` capsule owned by `"orchestrator"`, which is what made `process_failures_write_replayable_privacy_safe_capsules` fail on nightly.
- Capsule tests now require the failed action to be `expect_update_admin_policy_error` and print `action_id` / error code. Process scenarios use the default quiescence watchdog (256 iterations) instead of a tighter 100-iteration budget.
- Positive `PayloadCount` asserts count as `DeliveredPayload` oracle evidence, so `adversarial-reliability/app-witness-value` no longer fails `--strict-oracle` for missing AppMessage coverage.

Fixes #1450

## Test plan
- [x] New oracle tests failed before the `PayloadCount` coverage change and pass after
- [x] `process_failures_write_replayable_privacy_safe_capsules` and `process_cli_failure_report_keeps_capsules_after_exit` pass
- [x] `orchestrator_level_failures_do_not_own_app_process_capsules` pins participant-less capsule attribution
- [ ] Nightly capture of `adversarial-reliability/v1 --seed 7 --cases 12 --strict-oracle` no longer fails case-9

### DIFF
--- a/crates/cgka-conformance-simulator/src/oracle.rs
+++ b/crates/cgka-conformance-simulator/src/oracle.rs
@@ -142,6 +142,11 @@ pub fn build_scenario_oracle_report(
 ) -> ScenarioOracleReport {
     let stimuli = scenario_stimuli(spec);
     let mut oracle_behaviors = expected_behaviors(expected_trace, expected_outcomes);
+    for behavior in assertion_behaviors(spec) {
+        if !oracle_behaviors.contains(&behavior) {
+            oracle_behaviors.push(behavior);
+        }
+    }
     if spec
         .steps
         .iter()
@@ -149,8 +154,8 @@ pub fn build_scenario_oracle_report(
         && !oracle_behaviors.contains(&OracleBehavior::QuiescenceState)
     {
         oracle_behaviors.push(OracleBehavior::QuiescenceState);
-        oracle_behaviors.sort();
     }
+    oracle_behaviors.sort();
     let mut observed_behaviors = trace_behaviors(observed_trace);
     if expected_outcomes.iter().any(|expectation| {
         matches!(expectation, TraceExpectation::ClientsNotEquivalent { .. })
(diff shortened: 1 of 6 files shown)


crates/cgka-conformance-simulator/src/oracle.rs
```diff
@@ -142,15 +142,20 @@ pub fn build_scenario_oracle_report(
 ) -> ScenarioOracleReport {
     let stimuli = scenario_stimuli(spec);
     let mut oracle_behaviors = expected_behaviors(expected_trace, expected_outcomes);
+    for behavior in assertion_behaviors(spec) {
```
erskingardner, line 146: **[P1] Count passed payload assertions as observed evidence**

This adds `DeliveredPayload` only to `oracle_behaviors`. `missing_observed_behaviors` is still computed from `trace_behaviors(observed_trace)`, and `scenario_report_failures(..., true)` fails on every missing behavior. As a result, the new `positive_payload_count_assert_covers_app_message` example (which supplies an empty trace) still fails `--strict-oracle` with `missing_observed_behavior`, despite the new docs saying the assertion itself is coverage. Pass the executed, passing `assertion_observations` into oracle construction and add `DeliveredPayload` to observed behavior for a passing positive `PayloadCount` assertion, or narrow the claimed contract to require a separate trace observation.

erskingardner (commented):
Found a blocking strict-oracle inconsistency: positive `PayloadCount` assertions are added only to expected behavior, so the advertised coverage still fails unless a separate trace observation supplies the same behavior. The process-attribution split otherwise looks sound at this head. CI was still running at review time. GitHub does not allow this account to request changes on its own PR, so this is submitted as a comment review.

crates/cgka-conformance-simulator/src/oracle.rs
erskingardner: Addressed in c867fb36.

`build_scenario_oracle_report` now takes executed `assertion_observations`. Declared positive `PayloadCount` asserts still contribute expected `DeliveredPayload` (so `AppMessage` is not a weak-oracle warning), but only a *passing* observation counts as observed evidence. A declared-but-unexecuted or failed assert still produces `missing_observed_behavior`.

erskingardner (commented):
Recommend merge. The prior strict-oracle finding is fixed: declared positive `PayloadCount` assertions contribute expected coverage, only executed passing observations contribute observed `DeliveredPayload`, and failed or unexecuted assertions remain missing. The process-attribution behavior remains sound, and all exact-head CI checks pass. GitHub does not allow this account to approve its own PR, so this recommendation is submitted as a comment review.

PR #1456, opened 2026-08-15
Operationalize release convergence evidence

erskingardner:
## Summary

- replace the release lane's operator-authored manifest input with an exact ancestor revision
- build current and ancestor campaign images from exact source trees and materialize one reviewed mixed-build four-party campaign
- assemble a fail-closed, digest-pinned evidence bundle from the canonical scenario, strict public process oracle, receipts, lane observations, budgets, and required step records
- retain raw payload-bearing scenario and process-report bytes outside the shareable bundle while binding them through SHA-256 verification records
- update the convergence tracking and operator documentation

## Why

The release-hardening lane already had policy and budget machinery, but it could not independently provision its mixed-build inputs or produce the reviewed evidence bundle required by the convergence reliability plan. This makes that path operational and reproducible while keeping the final assurance claim scoped to one exact source revision, ancestor revision, canonical scenario, and observed run.

This PR changes campaign tooling, workflow configuration, tests, and documentation only. It does not change convergence-engine or app-runtime behavior.

## Validation

- `just fast-ci`
- `cargo clippy -p convergence-campaign-runner --all-targets --locked -- -D warnings`
- `actionlint .github/workflows/convergence-hardening.yml`
- CLI smoke test for materialization, mixed-build validation, and owner-only output permissions

## Remaining release step

After merge, an operator must dispatch the release-hardening workflow with a reviewed exact ancestor revision, download the resulting artifact, rerun `check-evidence`, review the scoped claim and residual gaps, and record the run URL and revisions. The planning document intentionally leaves that evidence-review item open.

## Summary by CodeRabbit

- **New Features**
  - Added release-hardening support for mixed current/baseline builds using exact ancestor revisions.
  - Added tooling to materialize convergence campaigns and assemble validated, privacy-safe evidence bundles.
  - Added CLI commands for campaign creation and evidence assembly.
  - Added support for local content-addressed image identifiers.

- **Documentation**
  - Documented release-hardening workflows, evidence requirements, dispatch commands, artifact paths, and manual review steps.

- **Bug Fixes**
  - Improved image validation and optional manifest handling.

### DIFF
--- a/crates/convergence-campaign-runner/src/release_evidence.rs
+++ b/crates/convergence-campaign-runner/src/release_evidence.rs
@@ -1,0 +1,825 @@
+use std::collections::{BTreeMap, BTreeSet};
+use std::fs::OpenOptions;
+use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
+
+use cgka_conformance_simulator::process_orchestrator::ProcessScenarioReportV1;
+use cgka_conformance_simulator::{
+    canonical_scenario_ir_sha256, cross_route_app_runtime_recovery_public_scenario,
+    resolve_scenario_input_bytes, validate_cross_route_public_process_report,
+};
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+
+use crate::{
+    CampaignBudgetEvaluationV1, CampaignLaneConfigV1, CampaignLaneObservationV1,
+    CampaignLaneStepObservationV1, CampaignLaneV1, ContainerBackendV1, ConvergenceEvidenceBundleV1,
+    DistributedBackendV1, DistributedCampaignManifestV1, DistributedParticipantV1,
+    DistributedRunReceiptV1, EvidenceArtifactV1, OciRuntimeV1, RunnerError, ScenarioArtifactV1,
+    TestedBoundaryV1, load_manifest, verify_manifest_inputs,
+};
+
+/// Version of the human-reviewed claim document consumed by release evidence assembly.
+pub const CONVERGENCE_EVIDENCE_CLAIM_VERSION: &str = "1";
+
+const RELEASE_CAMPAIGN_ID: &str = "cross-route-mixed-build-v1";
+const RELEASE_SCENARIO_FILE: &str = "canonical-scenario.json";
+const RELEASE_MANIFEST_FILE: &str = "distributed-manifest.v1.json";
+const RELEASE_OUTPUT_DIR: &str = "campaign-output";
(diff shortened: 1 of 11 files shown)


crates/convergence-campaign-runner/src/release_evidence.rs
```diff
@@ -0,0 +1,825 @@
...
+    let manifest_path = output_root.join(RELEASE_MANIFEST_FILE);
+    let manifest_bytes = serde_json::to_vec_pretty(&manifest)
+        .map_err(|error| RunnerError::environment("release_manifest_serialize", error))?;
+    fs_private::write_private(&manifest_path, &manifest_bytes)
+        .map_err(|error| RunnerError::environment("release_manifest_write", error))?;
+    Ok(manifest_path)
+}
+
+/// Validate one completed mixed-build release run and assemble a byte-pinned bundle.
+pub fn assemble_release_evidence(
```
erskingardner, line 210: Non-blocking: the helpers around this function are tested (materialize, claim pin, private copy, required steps), but `assemble_release_evidence` itself is not.

A passing-oracle success fixture reasonably waits for the first real dispatch. The early fail-closed gates do not: missing `normalized-manifest.json` / `distributed-run.json`, claim `campaign_id` mismatch, and `source_revision` absent from participant `build_id`s can all be asserted from a `materialize_release_campaign` tempdir plus the tracked claim file.

Those cases are the cheap part of the fail-closed story and would catch a wiring mistake before an operator burns a 6-hour hosted run.

erskingardner (commented):
Recommend merge. This operationalizes the remaining release-hardening production path in the shape the tracking plan asked for: exact ancestor SHA, two source-built content-addressed images, one runner-owned four-party manifest, then fail-closed assembly against the reviewed claim, mixed-build contract, normalized run, successful receipt, strict public oracle, recomputed budget, and required step records.

The privacy split is correct for this generated path. Canonical scenario bytes and the raw process report stay out of the shareable bundle and are bound only by digest; copied receipts are purpose/exit/accepted; materialized participant ids are the synthetic `zeta`/`alpha`/`yankee`/`observer` set. I would not add a redaction layer for `relay_command` / `node_command` before merge — those vectors are runner-authored defaults, and `fs_private` is about mode bits rather than content sanitization, which this code already does not confuse.

GitHub will not accept a self-approval on this PR, so this is a comment review rather than an approve.

Non-blocking: `assemble_release_evidence` still has no fail-closed tests of its own (see inline). A full passing-oracle fixture is reasonably deferred until the first real dispatch; the cheap early gates are not.

PR #1457, opened 2026-08-15
perf(group): return mutations before Welcome fanout and batch invite-with-admin

### LINKED ISSUE #1298: Batch invite-with-admin intent (one commit instead of invite + N promote commits)
Batch invite-with-admin intent (one commit instead of invite + N promote commits)

## Summary

Inviting members and immediately granting them admin currently produces 1 invite commit plus N independent `promote_admin` commit/publish/sync cycles — each paying its own `sync_runtime_groups` and relay ack wait. An invite intent carrying initial admins would collapse the whole action into a single group evolution.

## Evidence (master `bca10777`)

- `AppClient::promote_admin` (`crates/marmot-app/src/client/mod.rs:816-832`) → `update_admin_policy` (`client/mod.rs:870-897`): `sync_runtime_groups().await` → `SendIntent::UpdateAppComponents` commit+publish+confirm → `refresh_group` → `save_state` — per member, serially.
- Android's invite-as-admin flow loops `promoteAdminDetailed` per target after the invite (`ConversationController`, invite path), so "invite 3 people as admins" = 4 full network cycles.
- `CreateGroupRequest` already carries `initial_admins` (`client/mod.rs:264-277`), so the model precedent exists; the invite path lacks the equivalent.
- Admin-grant telemetry exists (`AppPerformanceOperation::GroupPromoteAdmin`, `app_telemetry.rs:37`), so the per-grant cost is measurable today.

## Suggested shape

- Extend the invite path (e.g. `SendIntent::Invite` or `invite_members*` request type) with `initial_admins`, applying the admin-policy component update in the same commit as the adds.
- Preserves authz invariants: single committer, single epoch transition, one publish wave.
- Clients: Android's invite-as-admin becomes one awaited operation instead of 1+N, and its partial-success toast path ("invite sent but admin promotion failed") disappears as a failure mode.

Related: marmot-protocol/mdk#466 (2026-08-07 evidence update documents the serialization multiplier).

Labels: `enhancement`, `MEDIUM`, `performance`

### LINKED ISSUE #1451: Return canonical group mutations before Welcome delivery fanout completes
## Problem

Group creation and existing-group invites keep caller-visible operations open after the durable protocol boundary needed for local success.

Grounded against `origin/master` at `a232a350448398aedadcc174c8e0eaeeabb123e9`.

### Founding group creation

`AppClient::create_group_with_initial_profile_and_optional_telemetry` already reaches a strong local boundary before network delivery:

1. the MLS founding state is prepared and persisted;
2. the group is canonical;
3. every exact founding Welcome obligation is durably recorded before the first external side effect.

Despite that, the method then awaits `publish_prepared_session_effects_with_audit_context`, which runs the complete first Welcome-attempt wave, before it saves/projects the group and returns. `publish_founding_group_created` uses bounded concurrency (8), but still waits for every in-flight recipient attempt and finish-stage bookkeeping. One slow recipient route therefore remains on the create spinner even though the group is already canonical and undelivered work is recoverable.

### Existing-group invites

After the invite commit is accepted and confirmed, `publish_group_evolution` publishes each resulting Welcome sequentially before returning. The member KeyPackage lookups earlier in `invite_members_with_optional_telemetry` are also sequential.

The commit acknowledgement is the authoritative group-evolution boundary. Once it is confirmed and exact Welcome delivery work is durable, recipient delivery is repairable work and should not extend the inviter's foreground wait.

## User impact

- Creating a group can feel relay-bound even though current engine benchmarks put one-invitee MLS creation around 1–2 ms.
- A single slow founding recipient can hold the entire create operation open.
- Existing-group invite latency grows additively with KeyPackage lookups and post-commit Welcome publication.
- The UI cannot present a locally usable group/member state while delivery continues, so bounded fanout still feels like a blocking operation rather than progress.

## Requested outcome

Return group mutations at the earliest safe canonical boundary and drive Welcome delivery through durable background work:

1. **Create:** after canonical founding state and all exact Welcome obligations are durably recorded, persist the minimal local projection, return the group id, and schedule first delivery attempts plus retries in the account worker.
2. **Existing invite:** await and confirm the invite commit, durably record every Welcome obligation, refresh/persist the local member projection, return the confirmed mutation, and schedule Welcome delivery independently.
3. Resolve independent invite KeyPackages with bounded concurrency while preserving input ordering and deterministic error selection.
4. Emit privacy-safe pending/delivered/failed recipient-count progress through the existing Welcome-delivery event/snapshot surface; do not expose identities or relay routes.
5. Preserve exactly-once external effects across cancellation, worker shutdown, and restart.

## Correctness constraints

- Do not report success before the founding state or invite commit is canonical and durable.
- Record exact Welcome bytes and delivery destinations before any relay exposure.
- A failed or ambiguous first attempt must retain the same delivery obligation; never create a second group, commit, or Welcome.
- Keep per-recipient acknowledgement and audit bookkeeping durable even when attempts complete out of order.
- Preserve member ordering, current-profile invariants, MLS authorization, terminal membership gates, and pending fanout recovery.
- Host cancellation after the canonical boundary detaches observation; it must not erase or duplicate delivery work.

## Acceptance criteria

- [ ] A deterministic delayed-relay test proves create returns after canonical persistence without waiting for a blocked founding Welcome attempt.
- [ ] A restart test proves undelivered founding Welcomes resume exactly once with no duplicate group.
- [ ] Existing-group invite returns after the confirmed commit/local projection while a delayed Welcome remains durably pending.
- [ ] Multi-member invite KeyPackage lookup and Welcome attempts use documented bounded concurrency and deterministic result ordering.
- [ ] Partial delivery exposes stable aggregate progress and retries only the undelivered exact obligations.
- [ ] Cancellation and shutdown tests cover pre-canonical cancellation, post-canonical detachment, and restart recovery.
- [ ] Phase telemetry separately measures prepare/persist, commit confirmation, local projection, first delivery attempts, and background completion.
- [ ] Recorder-on/off benchmarks cover 1, 5, and 20 recipients with one delayed route.

## Related / duplicate disposition

- #1162 / PR #1316 added bounded founding KeyPackage/Welcome concurrency, detached subscription refresh/catch-up, and stage telemetry. It closed the serialized-sum problem, but current `master` still awaits the complete first Welcome-attempt wave before create returns.
- #466 addressed profiling and latency for inviting members to existing groups; the remaining source still performs sequential KeyPackage lookup and sequential post-commit Welcome publication.
- #1443 owns resumable host-visible operation phases and ambiguous completion handling. This issue owns shortening the underlying canonical-to-return critical path and background Welcome execution.
- #1298 owns combining invite-with-admin into one commit; it is complementary.
- #352 and the existing pending-Welcome machinery own durable retry semantics that this issue should reuse rather than replace.

erskingardner:
## Summary
- Create and existing-group invite now return after the durable MLS/commit boundary; Welcome fanout continues in the background with the same bounded concurrency as founding create (`Fixes #1451`).
- Invite-with-admin is one Add-plus-admin-policy commit instead of invite plus N `promote_admin` cycles (`Fixes #1298`). Empty `initial_admins` keeps the add-only path, so reverting this commit leaves the background Welcome lifecycle intact.

## Test plan
- [x] New engine tests failed before the invite-with-admin commit and pass after (`invite_with_initial_admin_grants_admin_in_the_invite_commit`, `invite_rejects_initial_admin_who_is_not_an_invitee`)
- [x] `cargo test -p marmot-app --test relay_runtime invite_members_returns_before_blocked_welcome invite_members_with_initial_admins`
- [x] `just fast-ci`
- [ ] GitHub CI `just ci` matrix

## Summary by CodeRabbit

- **New Features**
  - Invitations can grant administrator privileges to selected invitees in the same group update.
  - Welcome messages are delivered reliably after group creation and invitations, with retryable recovery for interrupted delivery.
  - Added performance monitoring for group-invite Welcome delivery.

- **Bug Fixes**
  - Invite acceptance now retries for busy-worker and temporarily unknown-group errors.
  - One-shot group commands now complete pending work before exiting.
  - Improved handling of combined membership and administrator changes, including failed-delivery recovery.


crates/marmot-app/src/client/mod.rs
```diff
@@ -1409,28 +1430,38 @@ impl AppClient {
...
         record_app_performance(
             telemetry,
             AppPerformanceOperation::GroupInviteEnginePublish,
             engine_publish_started_at.elapsed(),
-            effects.is_ok(),
+            commit.is_ok(),
         );
-        let effects = effects?;
+        let (effects, welcomes) = commit?;
+        let welcome_intents = self.record_welcome_delivery_intents(group_id, &welcomes)?;
```
erskingardner: **[P1] Keep Welcome delivery scheduled across post-confirm failures.** `send_confirming_commit_with_audit_context` has already confirmed the MLS commit before this app-index write. If this `?` fails, the method returns an error for a mutation that is already canonical and never stores the work in `unpublished_welcome_delivery`; if `local_refresh?` below fails, the worker likewise skips `drive_unpublished_welcome_delivery` and leaves this single-slot work vulnerable to being overwritten by a later mutation. The engine copy makes manual recovery possible, but no first transport attempt/fanout record exists for automatic maintenance to resume. Treat the convenience index/projection as post-canonical repair, and durably schedule/stash fanout before any fallible post-confirm step. Please add injected index/projection-failure regressions that prove the caller is not told to retry an applied invite and the exact Welcome still gets its first attempt.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -2422,8 +2436,19 @@ async fn handle_account_worker_command(
                     &group_id,
                 );
             }
-            publish_pending_welcome_delivery_events(events, account_id_hex, account_label, client);
+            let invited = result.is_ok();
             let _ = respond.send(result);
+            if invited {
+                client
```
erskingardner: **[P1] Do not keep the account worker occupied by the background fanout.** Sending the oneshot first releases this one caller, but the handler still awaits the entire Welcome publish and cannot dequeue any same-account command. The real `invite_members_detailed` UniFFI path immediately calls `group_mutation_result_for`, which awaits `group_members` and `group_mls_state` on this same worker, so it remains blocked by the slow Welcome and the user-visible optimization is lost. The new test reads `app.groups(...)` directly, bypassing the worker, so it cannot catch this. Either serve safe snapshot reads while fanout owns the live client, or return the post-mutation projection with the command/otherwise truly detach the delivery lifecycle. Add a blocked-relay regression through `invite_members_detailed` (or the equivalent command-plus-worker-read sequence).

crates/marmot-uniffi/src/commands/group.rs
```diff
@@ -735,11 +744,14 @@ impl Marmot {
...
     }
 
+    /// Same as [`Self::invite_members`], returning the post-mutation group
+    /// snapshot. `initial_admin_refs` grants admin in the invite commit.
     pub async fn invite_members_detailed(
         &self,
         account_ref: String,
         group_id_hex: String,
         member_refs: Vec<String>,
+        initial_admin_refs: Vec<String>,
```
erskingardner, line 781: **[P1] Preserve the existing UniFFI entry point.** Adding a required positional argument changes the generated Swift/Kotlin method and breaks current callers; for example, `whitenoise-ios` and `whitenoise-mac` call `inviteMembersDetailed(accountRef:groupIdHex:memberRefs:)` with the existing three arguments. The runtime layer already keeps a compatibility wrapper. Do the same at the binding boundary: retain `invite_members` / `invite_members_detailed` delegating with an empty admin list, and add separately named `*_with_initial_admins` exports. This applies to the non-detailed method above as well.

erskingardner (commented):
Blocking findings at exact head `0be9abee570740a34a6462682af239d58cbded26`. The canonical/post-confirm error seam can suppress Welcome fanout, the managed worker still serializes the detailed host path behind fanout, and the binding change breaks existing exported call sites.

CI is also not green at this head: `Rust tests (1/4)` fails `connector_socket_subscribes_to_inbound_messages` with `unknown local group`, and `CLI e2e (mock relay)` fails `group_members_invite_and_remove_flow_updates_projected_members` because `published` changed from 2 to 1. `Simulator and vectors` was still running at review time. Because the PR author and reviewer are the same GitHub account, this is posted as a blocking comment rather than `REQUEST_CHANGES`.

crates/marmot-app/src/client/mod.rs
erskingardner: Ordering concern: by this point `send_confirming_commit_with_audit_context` has already published the invite commit to relays, so the durable delivery obligations are recorded *after* the first external side effect. #1451 lists "Record exact Welcome bytes and delivery destinations before any relay exposure" as a correctness constraint, and the repo's multi-step-state-change invariant requires recording intent before external side effects.

This bites harder for invites than for founding creates: founding Welcomes are persisted as delivery-aware `OutboundWelcome` records in the canonical engine transaction, so `outstanding_sent_welcomes` / `reconcile_pending_welcome_delivery_index` can rebuild the index after a crash. Invite Welcomes are persisted as raw transport rows, which `outstanding_sent_welcomes` deliberately excludes, and the durable fanout records are only created later inside `prepare_legacy_publish` during the deferred drive. A crash between commit publish and this line leaves the invitee a confirmed member with no repair handle — the Welcome is never retried.

Could the intents be recorded in the gap inside `send_confirming_commit_with_audit_context` (between `take_deferred_welcomes` and `publish_session_effects_with_audit_context`), e.g. via a callback or a split stage/publish call, so the obligation is durable before the commit leaves the process? (Master had a narrower version of this window and no invite intent rows at all, so this isn't a regression — but this PR is adding exactly this machinery and the constraint is explicit.)

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -2422,8 +2436,19 @@ async fn handle_account_worker_command(
...
+                    .drive_unpublished_welcome_delivery(Some(&telemetry))
```
erskingardner: Non-blocking: the fanout is awaited inline in the worker loop after replying, so a slow or blocked Welcome still occupies the account worker — subsequent commands for this account (send, another invite, sync) queue behind it until the publish attempts exhaust their timeouts. That meets the caller-visible latency goal, but "background" is only relative to the caller. Worth a comment here at minimum; if account-level head-of-line blocking shows up in practice, this wants a detached drive with its own scheduling.

crates/marmot-app/src/client/mod.rs
```diff
@@ -3556,30 +3587,40 @@ impl AppClient {
...
         let recorded_at = unix_now_seconds();
         let storage = self.app.account_storage(&self.state.label)?;
         let mut message_ids = Vec::new();
-        for work in &effects.publish {
-            let PublishWork::FoundingGroupCreated { welcomes } = work else {
-                continue;
+        for welcome in welcomes {
+            let TransportEnvelope::Welcome { recipient } = &welcome.envelope else {
+                return Err(AppError::Publish(
+                    "founding delivery artifact was not a Welcome".into(),
```
erskingardner: Nit: this error text still says "founding" but the helper is now shared with the invite path — an invite artifact failure would misreport its origin.

erskingardner (commented):
**Review verdict: request changes** (posted as a comment because GitHub does not allow requesting changes on your own PR).

Direction is right and the core mechanics look sound: the invite-with-admin engine path (single Add+admin-policy commit, grant dedup/validation, staged-commit coupling checks, `AdminAdded` attribution on confirm) is well done, the deferred fanout cleanly reuses the founding bounded-concurrency machinery with per-recipient completion bookkeeping, and direct `AppClient` callers keep the old await-everything semantics. But CI is red with two PR-caused failures and there is one durability-ordering issue to fix before merge.

**Blocking**

1. **CLI e2e failure (`group_members_invite_and_remove_flow_updates_projected_members`)**: `invite["published"]` is now 1 (commit only) because the Welcome fanout is deferred. The same assertion exists at `crates/cli/tests/cli.rs:6808` in `three_user_message_lifecycle_covers_invite_remove_and_later_delivery`. Both need updating.

2. **Deeper one-shot CLI concern behind that failure**: `wn group invite` / `wn group create` build a fresh `MarmotAppRuntime` (`group_command` → `app.runtime()`) and return without `shutdown()`. Since the fanout is now driven by the worker *after* the response is sent, process exit drops the tokio runtime and cancels the in-flight fanout — the Welcome may never be attempted until some later command opens that account. In both e2e flows, `sync_until_joined(carol)` runs no further alice commands, so the tests can still hang after the assertion is fixed; worse, a real CLI user on a slow relay (exactly the scenario this PR optimizes) gets a fast return but no Welcome. Consider having one-shot CLI commands drain before exit — e.g. `runtime.shutdown().await`, which waits for the in-flight worker handler — so CLI semantics stay synchronous while app/FFI callers return early.

3. **agent-connector failure (`connector_socket_subscribes_to_inbound_messages`)**: fails with `invite acceptance failed: unknown local group` — the test accepts the invite immediately after `create_group` returns, but the founding Welcome is still in the background fanout and the retry helper only tolerates `AccountWorkerBusy`. Please update the helper to wait for invite arrival. Note this is a caller-visible contract change for all consumers (FFI included): invitees can no longer accept immediately after the inviter's create/invite returns; they must wait for delivery (pending-invite / `GroupJoined` events). Worth explicit release notes.

4. **Invite Welcome intents are recorded after the commit's relay exposure** (inline comment). A crash in that window loses the invitee's only repair handle, violating the issue's own 'record destinations before any relay exposure' constraint.

**Non-blocking**

- UniFFI `invite_members` / `invite_members_detailed` gained a required `initial_admin_refs` parameter — compile-breaking for existing MarmotKit consumers. Probably intended (Android is the motivator in #1298), but call it out in release notes.
- `SendSummary.published` / `message_ids` for invite now cover only the commit; Welcome ids never appear. Consistent with the new boundary, but a visible contract change.
- Worker head-of-line and a stale 'founding' error string — inline.

Test coverage is otherwise strong: the blocked-Welcome and restart-exactly-once relay tests map directly to #1451's acceptance criteria, and the engine-level invite-with-admin tests failed-before/pass-after per the PR description.

crates/marmot-app/src/client/mod.rs
```diff
@@ -1409,16 +1430,62 @@ impl AppClient {
...
+            && self.app.config.dev_fail_invite_welcome_intent
+        {
+            Err(AppError::Publish(
+                "injected invite welcome intent failure".into(),
+            ))
+        } else {
+            self.record_welcome_delivery_intents(group_id, &welcomes)
+        };
+        let welcome_intents =
+            recover_post_canonical_result("record_welcome_delivery_intents", welcome_intent_result);
```
erskingardner: **[P1] Do not expose the commit after intent persistence failed.** This recovery converts an index-write failure into an empty `welcome_intents`, and the code below still publishes the invite commit. Invite Welcomes are retained as raw `Sent` payloads, while `outstanding_sent_welcomes` deliberately scans only delivery-aware `OutboundWelcome` records; therefore a crash after commit publish and before the in-memory fanout leaves a canonical member with neither an app pending row nor an engine obligation that cold-start reconciliation can discover. The new fault test proves only a same-process first attempt, not crash recovery. Keep the caller-visible result canonical, but do not expose the commit until the exact obligation is durable (or persist invite Welcomes as delivery-aware records transactionally), and add a restart regression for this injected failure.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -784,11 +787,16 @@ async fn run_app_runtime_account_worker(
                 handle_account_worker_command(
                     &mut client,
                     *command,
-                    &events,
-                    &account_id_hex,
-                    &account_label,
-                    &shared,
-                    &media_http,
+                    AccountWorkerCommandContext {
+                        commands: &mut unused_commands,
```
erskingardner: **[P1] Serve reads from the live startup command queue.** `unused_commands` is a permanently closed receiver. When an `InviteMembers` command was deferred during initial hydration/catch-up, replay reaches this handler, sends the invite result, and then `serve_snapshot_reads_until` can observe only that closed receiver. The detailed caller immediately queues `GroupRoster`/membership reads on the real `commands` channel, so those reads still wait until the Welcome fanout completes (or times out). This preserves the original latency bug specifically during the common post-readiness startup window. Replay needs to service the live receiver and carry any deferred commands back into the steady-state FIFO; add a blocked-Welcome regression with the invite itself deferred by startup catch-up.

crates/cli/src/commands/groups.rs
```diff
@@ -49,7 +50,13 @@ pub(crate) async fn group_command(
     account_flag: Option<String>,
 ) -> Result<CommandOutput, WnError> {
     let runtime = app.runtime();
-    group_command_with_runtime(account_home, app, &runtime, command, account_flag).await
+    let result =
+        group_command_with_runtime(account_home, app, &runtime, command, account_flag).await;
+    // Drain in-flight Welcome fanout before the process drops this runtime.
+    // Create/invite now reply at the canonical MLS boundary; without this
+    // wait, one-shot CLI would cancel delivery on exit.
+    runtime.shutdown().await;
```
erskingardner, line 79: **[P1] Drain fanout before beginning runtime shutdown.** `MarmotAppRuntime::shutdown` is not a publish barrier: it calls `begin_shutdown`, then joins `accounts.shutdown()` with `relay_plane.shutdown()`. The latter immediately marks the transport shutting down and shuts the Nostr SDK pool while the account worker may only just be starting the post-response Welcome publish. On a slow relay, the CLI caller can win this scheduling race and cancel/fail the exact fanout this wait is meant to preserve; the fast local-relay e2e does not exercise that ordering. Await an account-worker delivery barrier before shutting the relay plane (or provide a scoped flush command), then call shutdown.

erskingardner (commented):
Blocking re-review findings at exact head `7dd5a905527defc85c38a4e7734178828aff3719`. The legacy UniFFI entry points, steady-state snapshot reads, stale error text, and prior CI failures are fixed. Three lifecycle gaps remain: an invite-intent persistence failure is still followed by commit relay exposure; startup-deferred detailed invites still wait behind fanout because replay uses a closed command receiver; and the one-shot CLI calls shutdown as a drain even though shutdown closes the relay plane concurrently with account workers. GitHub Actions is otherwise passing at review time except `Simulator and vectors`, which is still pending; CodeRabbit is also pending. Because the PR author and reviewer are the same GitHub account, this is posted as a blocking comment rather than `REQUEST_CHANGES`.

crates/marmot-app/src/client/mod.rs
```diff
@@ -1409,16 +1430,103 @@ impl AppClient {
...
+                    Ok::<_, AppError>(())
+                })();
+                record_app_performance(
+                    telemetry,
+                    AppPerformanceOperation::GroupInviteLocalRefresh,
+                    local_refresh_started_at.elapsed(),
+                    local_refresh.is_ok(),
+                );
+                recover_post_canonical_result("invite_members_local_refresh", local_refresh);
+                return Ok(SendSummary {
```
erskingardner: **[P1] Roll back instead of reporting a noncanonical pending send.** `confirm_commit_without_publish_with_audit_context` only calls `session.send`; the invite path still has an attached OpenMLS staged commit, and `publish.rs` merges it only after the commit reaches transport. Dropping `session_effects` here creates no durable outbound queue or fanout, yet `AcceptedPending` explicitly promises a durable queued intent that survives restart. On cold open, a staged add with no durable fanout/evolution is cleared, so the caller observes success and the invitee in the projection, then the membership silently disappears. A real mid-loop storage failure can also leave a prefix of pending-Welcome rows because `record_welcome_delivery_intents` writes each row independently; those stale handles can re-deliver a Welcome whose commit was never exposed. Since no external side effect has occurred, atomically compensate any written rows, drive `publish_failed` to roll back the staged commit, and return the original error—or make commit plus every Welcome obligation durable as one unit before reporting success. Add a multi-invite partial-write/restart regression that asserts final stable membership, not only absence of `GroupJoined`.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -767,14 +775,17 @@ async fn run_app_runtime_account_worker(
...
+    // post-canonical snapshot reads (Members / GroupRoster / …) can land while
+    // a deferred create/invite still owns Welcome fanout.
     let (media_http_tx, mut media_http_rx) = mpsc::unbounded_channel();
     let (media_http_worker_lifetime, _) = watch::channel(());
     let media_http = MediaHttpContext {
         tx: media_http_tx,
         permits: Arc::new(Semaphore::new(MEDIA_HTTP_IN_FLIGHT_LIMIT)),
         worker_lifetime: media_http_worker_lifetime,
     };
+    let mut pending = VecDeque::new();
```
erskingardner: **[P1] Keep the remaining startup FIFO visible to snapshot serving.** The outer `for deferred_command in deferred` retains all later startup commands outside this `pending` queue. If startup recorded `InviteMembers(A)` followed by mutation `B`, then a live `Members`/`GroupRoster` read arriving while A's Welcome is blocked sees `pending` as empty and is answered from A's snapshot before the earlier B runs. That contradicts `serve_snapshot_reads_until`'s rule that a deferred mutation prevents later reads from bypassing it and can return stale state out of command order. Materialize the remaining startup commands into the same FIFO (while preserving coalesced CatchUp responses), or disable snapshot reads whenever an earlier startup mutation remains. Extend the blocked-Welcome startup test with an already-deferred mutation ahead of the live read.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -5059,6 +5076,33 @@ impl AccountManager {
...
+            if commands
+                .send(AccountWorkerCommand::Drain { respond })
+                .await
+                .is_ok()
+            {
+                responses.push(response);
+            }
+        }
+        for response in responses {
+            let _ = timeout(APP_RUNTIME_ACCOUNT_SHUTDOWN_WAIT, response).await;
```
erskingardner: **[P1] Do not silently abandon the delivery barrier.** This discards a five-second timeout, after which the one-shot CLI immediately calls `shutdown`; shutdown closes the relay plane concurrently and eventually aborts the worker. Thus a slow Welcome—the exact case this PR targets—can still outlive the barrier and lose its first attempt when the process exits, while `drain_in_flight_work`'s public contract says it waited for completion. Await the worker barrier through the transport's own bounded attempt lifecycle, or return a typed incomplete result that the CLI handles without claiming the work was drained. Add a delayed/blocked-relay one-shot regression whose fanout exceeds five seconds.

erskingardner (commented):
Blocking re-review findings at exact head `c2e4207dcba906e8a276d51995508169a444f0b8`. The remediation adds the right structural seams—a live startup queue and a pre-shutdown worker barrier—and preserves the actual invite-acceptance retry error. Three correctness/liveness gaps remain: the intent-write failure path reports a noncanonical, nonqueued invite as `AcceptedPending`; startup replay can let live reads bypass earlier deferred mutations; and the CLI drain silently abandons the barrier after five seconds before shutting down the relay plane. CI is still in progress at review time. Because the PR author and reviewer are the same GitHub account, this is posted as a blocking comment rather than `REQUEST_CHANGES`.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in 2b8086ce. All startup-deferred commands, including coalesced catch-up responses, are materialized into the same pending FIFO before replay. The regression queues a remove behind the deferred invite and proves a later live read cannot bypass it.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Fixed in 2b8086ce. drain_in_flight_work now awaits each worker barrier without the unrelated five-second shutdown budget and returns a typed error if the barrier cannot complete. The CLI propagates that result after orderly shutdown; the blocked-relay regression proves the barrier remains pending beyond five seconds and completes after release.

crates/marmot-app/src/client/mod.rs
erskingardner: Fixed in 2b8086ce. Failed Welcome-intent persistence now atomically rolls back the entire staged MLS commit and returns the original error. Multi-recipient intent rows are written in one SQLite transaction, with regressions for batch rollback and stable post-restart membership.

crates/marmot-uniffi/src/commands/group.rs
erskingardner: Addressed in 7dd5a905: the existing three-argument UniFFI exports remain unchanged and delegate with an empty admin list; separately named *_with_initial_admins exports carry the new argument.

crates/cli/src/commands/groups.rs
erskingardner: Addressed across c2e4207d and 2b8086ce: one-shot CLI now sends a worker delivery barrier before runtime shutdown, awaits it through the transport attempt lifecycle without the old five-second cutoff, and only then shuts down the relay plane.

crates/marmot-account/src/runtime.rs
```diff
@@ -1629,6 +1654,106 @@ where
...
+            .await?;
+        let pending = session_effects
+            .publish
+            .iter()
+            .find_map(|work| match work {
+                PublishWork::GroupEvolution { pending, .. }
+                | PublishWork::GroupCreated { pending, .. } => Some(*pending),
+                _ => None,
+            })
+            .ok_or_else(|| {
```
erskingardner: [P1] Do not leave a queued invite behind an error

`send_with_audit_context` can legitimately return only `SessionEffects.queued`: when authenticated convergence/deferred-peel work is unsettled, `should_queue_outbound_intent` returns true for any intent in a Stable group and persists it. This lookup then finds no `GroupEvolution`, returns a Backend error, and drops the queued effect without deleting the durable intent. The caller is told the invite failed even though a later convergence drain can still apply it, so a retry can enqueue a duplicate. Use a no-queue staging path here, or explicitly wait/cancel so every returned error means no durable invite remains.

crates/marmot-app/src/client/mod.rs
```diff
@@ -1409,16 +1430,89 @@ impl AppClient {
...
+        // persistence fails, roll back the staged MLS change and return the
+        // original error so a retry cannot create phantom local membership.
+        let welcome_intent_result = if cfg!(feature = "test-policy-overrides")
+            && self.app.config.dev_fail_invite_welcome_intent
+        {
+            Err(AppError::Publish(
+                "injected invite welcome intent failure".into(),
+            ))
+        } else {
+            self.record_welcome_delivery_intents(group_id, prepared.welcomes())
```
erskingardner, line 1511: [P1] Couple the Welcome index to cold-start commit recovery

After this transaction succeeds, a process death before `publish_prepared_session_effects_with_audit_context` starts leaves durable app Welcome handles/raw Welcome rows but only an ordinary staged Add commit. Cold hydration clears such a pending Add unless a durable fanout/evolution exists, and this invite has neither at this point. The index reconciliation preserves the raw-Welcome row (`tracked_outbound_welcome_ids` excludes it), while `redeliver_welcome` accepts raw rows, so restart can expose and deliver a Welcome whose commit was rolled back. Couple the intent to a durable commit-recovery record (or atomically retire both on cold rollback), and add a crash/restart test at this seam.

erskingardner (commented):
Request changes

The latest commit fixes the three issues from the previous pass: Welcome-intent insertion is atomic and rolls the staged invite back on failure, startup replay now exposes one FIFO to snapshot serving, and the CLI waits for the actual worker delivery barrier. Two P1 correctness gaps remain at the new prepare/persist boundary.

Current exact-head CI is green except for `Simulator and vectors`, which is still pending; CodeRabbit reports rate limiting. This is posted as COMMENT because GitHub does not allow the PR author account to request changes on its own PR.

crates/cgka-engine/src/message_processor/store.rs
```diff
@@ -32,6 +32,41 @@ fn fresh_deferred_peel_lifecycle(
...
+/// the staged commit and those rows one recovery unit without teaching storage
+/// about MLS pending references.
+pub(crate) fn transition_staged_invite_welcomes<S: StorageProvider>(
+    storage: &S,
+    group_id: &GroupId,
+    source_epoch: EpochId,
+    state: MessageState,
+) -> Result<(), EngineError> {
+    debug_assert!(matches!(state, MessageState::Sent | MessageState::Failed));
+    for mut record in storage.list_messages(group_id, source_epoch)? {
```
erskingardner, line 52: [P1] Link staged Welcomes to the exact pending commit

Group plus source epoch is not a sufficient ownership key for this transition. A concrete upgrade path leaves another matching row: master can stage an invite, crash before publication, then cold-recover the Add by clearing it while leaving its raw Welcome `Sent` and the group at the same epoch. Confirming any later evolution at that epoch makes this scan promote the stale orphan alongside the current work; `outstanding_sent_welcomes` then lets startup publish a Welcome whose Add was rolled back. Persist an explicit staged-invite/origin identifier and transition only records belonging to the current pending commit. Please cover a stale raw Welcome at the same source epoch in the regression.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -767,32 +781,88 @@ async fn run_app_runtime_account_worker(
...
+    let welcome_recovery_snapshot = capture_group_read_snapshot(
+        &client,
+        &events,
+        &account_id_hex,
+        &account_label,
+        "startup Welcome recovery snapshot failed",
+    );
+    serve_snapshot_reads_until(
+        welcome_recovery_snapshot,
+        client.retry_pending_welcome_deliveries_best_effort(),
```
erskingardner: [P1] Keep recovered Welcome retries off the mutation FIFO

This awaited job runs before the pending command loop, while `retry_pending_welcome_deliveries_best_effort` retries every row sequentially. One unavailable recipient therefore blocks all deferred and newly arriving mutations for the account until its transport timeout, and N retained Welcomes multiply that wait—even for unrelated groups. That recreates the relay-bound foreground latency this PR is removing and bypasses the shared bounded Welcome fanout. Once the Add is canonical, no protocol ordering requires its Welcome delivery to precede later mutations. Run recovery as bounded background work and have the explicit drain/shutdown barrier join it; add a blocked-recovery test proving a mutation can finish while the delivery barrier still waits.

erskingardner (commented):
Request changes

The queued-invite error path from the previous pass is fixed: a convergence-deferred invite now remains explicitly `AcceptedPending`. The cold-recovery change also correctly retires newly staged Welcome rows when the matching Add is rolled back.

Three blocking lifecycle issues remain. I independently confirmed the existing staged-redelivery finding ([discussion](https://github.com/marmot-protocol/mdk/pull/1457#discussion_r3789903690)): after an ambiguous commit attempt returns, the app repair row is visible and `stored_sent_welcome` still accepts its raw pre-confirmation Welcome. I also confirmed the rollback queue finding ([discussion](https://github.com/marmot-protocol/mdk/pull/1457#discussion_r3789903693)): `publish_failed` can generate `AutoPublish` during buffered-message replay, while this boundary only debug-asserts and then drops that queue in release builds. The two additional inline comments cover stale-row promotion and startup mutation head-of-line blocking.

All GitHub checks pass at exact head `5589ae3c88b6a38ed4848f291687493f3625fb2a`. Posted as COMMENT because GitHub does not allow the PR author account to request changes on its own PR.

crates/marmot-app/src/client/mod.rs
erskingardner: Fixed in 0a75ba62. Invite Welcomes now persist as non-deliverable `StagedInviteWelcome` records bound to the exact wrapped commit id. Cold compatibility rollback retires staged artifacts in the same transaction as clearing the pending commit, and app reconciliation treats staged ids as tracked but never outstanding. The publish lifecycle and cold-recovery regressions cover pre-confirmation rejection and restart rollback.

crates/cgka-engine/src/engine.rs
```diff
@@ -2655,11 +2662,13 @@ impl<S: StorageProvider> Engine<S> {
...
-    /// records remain directly re-deliverable by a known id for compatibility,
-    /// but are not rediscovered as outstanding after an upgrade because older
-    /// versions did not persist acknowledgement completion (mdk#352).
+    /// Founding Welcomes are persisted at canonical creation and invite
+    /// Welcomes are promoted when their Add commit confirms, both as
+    /// delivery-aware `OutboundWelcome` records. Historical `Sent`
+    /// raw-transport Welcome records remain directly re-deliverable by a known
+    /// id for compatibility, but are not rediscovered as outstanding after an
+    /// upgrade because older versions did not persist acknowledgement
+    /// completion (mdk#352).
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0a75ba62. New invite Welcomes no longer use `RawTransport`; they use a dedicated staged payload that `stored_sent_welcome` cannot return. Confirmation promotes only the exact commit-owned rows to `OutboundWelcome`. Historical raw rows remain an explicit compatibility fallback, and the regression verifies rejection before confirmation.

crates/marmot-account/src/runtime.rs
```diff
@@ -1629,6 +1663,96 @@ where
...
+        let mut queue = VecDeque::new();
+        output.absorb_session_effects(prepared_effects, &mut queue);
+
+        let rollback_effects = self.session.publish_failed(pending).await?;
+        output
+            .pending
+            .push(PendingResolution::RolledBack { pending });
+        output.absorb_session_effects(rollback_effects, &mut queue);
+        debug_assert!(queue.is_empty());
+        Ok(output)
```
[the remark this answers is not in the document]
erskingardner: Fixed in 0a75ba62. `rollback_prepared_session_commit` now drains the `PublishWork` queue produced by `publish_failed` through the normal `publish_queue` path, then runs the same post-publish reconciliation hooks. Replay-generated `AutoPublish` work is therefore published/resolved instead of debug-asserted and dropped.

crates/cgka-engine/src/message_processor/store.rs
erskingardner: Fixed in 0a75ba62. Each staged invite Welcome stores its exact `origin_commit_id`; confirm promotes only matching rows and retires nonmatching staged or ambiguous legacy rows at that source epoch. The regression injects both a different-origin staged row and a legacy raw row at the same epoch and verifies neither becomes deliverable.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in 0a75ba62. Startup recovery now uses the shared bounded Welcome fanout with an explicit forced retry, runs only when the command queue is idle, and is cancellation-safe when a command arrives. `Drain` joins the retry instead of interrupting it. The relay regression blocks recovery, completes a local account mutation, verifies the drain still waits, then releases and confirms delivery.

erskingardner:
Addressed the latest lifecycle review in 0a75ba62:

- staged invite Welcomes now have a non-deliverable payload bound to the exact origin commit;
- same-epoch stale/legacy rows cannot be promoted with a later commit;
- rollback drains replay-generated publish work;
- startup Welcome recovery is bounded, yields to account commands, and is joined by Drain;
- the deferred-publish API now rejects non-commit intents before staging and marks prepared commits `must_use`.

Verification: `just fast-ci`, full `marmot-account`, full `cgka-traits`, full `marmot-app` relay runtime suite with test policy overrides, and targeted publish/cold-recovery/legacy-compatibility engine regressions.

erskingardner:
Follow-up from the staged-payload review: ec99595a closes the own-echo overwrite seam. `persist_transport_message_for_existing_group` now preserves `StagedInviteWelcome` just as it preserves `OutboundWelcome`, so an echoed event cannot rewrite a pre-confirmation Welcome into deliverable legacy `RawTransport`. Added a focused regression; `just fast-ci` remains green.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -820,8 +870,40 @@ async fn run_app_runtime_account_worker(
...
     maintenance_tick.set_missed_tick_behavior(MissedTickBehavior::Delay);
     let mut legacy_message_promotion = LegacyMessagePromotionSchedule::new();
+    // Recovery is durable, exact-artifact work, so it can yield to account
+    // commands without losing its place. A Drain command is the exception: it
+    // joins the bounded retry before acknowledging the shutdown barrier.
+    let mut welcome_recovery_pending = true;
 
     'worker: loop {
+        if welcome_recovery_pending && pending.is_empty() {
+            let mut recovery =
```
erskingardner: [P1] Do not make startup Welcome recovery the worker's exclusive top-level future. While `recovery` is pending, this pre-loop `select!` never polls `receive_next_delivery`, maintenance, convergence, media completion, or timers, so one slow retained Welcome can stop all inbound processing until a host command arrives. Worse, because the select is biased toward a ready command and drops `recovery`, a continuously busy command queue can cancel every attempt and starve the Welcome forever. The current test exercises one mutation followed by `Drain`, which forces completion, but does not cover either path. Keep recovery pollable alongside the normal worker work (or move the network attempt into a separately owned task) and add a regression with a blocked Welcome plus inbound delivery/queued commands proving both account progress and eventual retry.

crates/cli/src/commands/groups.rs
```diff
@@ -49,7 +52,15 @@ pub(crate) async fn group_command(
     account_flag: Option<String>,
 ) -> Result<CommandOutput, WnError> {
     let runtime = app.runtime();
-    group_command_with_runtime(account_home, app, &runtime, command, account_flag).await
+    let result =
+        group_command_with_runtime(account_home, app, &runtime, command, account_flag).await;
+    // Drain in-flight Welcome fanout while the relay plane is still live.
+    // Create/invite now reply at the canonical MLS boundary; without this
+    // wait, one-shot CLI would cancel delivery on exit.
+    let drain_result = runtime.drain_in_flight_work().await;
```
erskingardner: [P2] Scope this drain to commands that actually launched post-response fanout. `group_command` also wraps `Members`, `Remove`, `Update`, and avatar updates; `groups_command` similarly wraps show/read operations. On startup with a retained Welcome, those commands can interrupt recovery, finish their own work, then block here while `Drain` waits for the unrelated slow recipient. A read-only `wn group members` can therefore hang on a Welcome it did not create. Preserve the flush for create/invite (and the AddMembers alias), but let other one-shot commands shut down with the exact obligation still durable.

erskingardner (commented):
Request changes at exact head `ec99595abb8936c6d4293f9cbc51f45edbcdf6e6` (posted as a comment because GitHub does not allow requesting changes on your own PR).

The exact-commit staged-Welcome lifecycle, cold rollback, own-echo preservation, rollback-generated publish work, queued-intent classification, UniFFI compatibility, and bounded fanout now look correct. The full CI matrix is green. A clean pass still found two scheduling regressions: startup Welcome recovery can block inbound account work and can be starved forever by a busy command queue, and the one-shot CLI applies the delivery barrier to commands that did not launch fanout.

Datawav (commented):
## Automated review

Key claims verified against the patches. The remaining claims (rollback/transition logic, FIFO drain gating, `Published`-only fanout, ensure_group_admin at FFI) are consistent between both reviews and their cross-verifications. Kimi's F1–F4 rejections target an earlier Sol draft, not the final review, which contains no findings — so there is nothing to reconcile as a supported defect. Sol's cross-check of Kimi explicitly rejected each of its own "missed" candidates (a)–(d) as non-defects or unsupported, and its only residual item is a review-completeness note, not a code finding.

Crystallized final audit:

---

Outcome: No actionable findings.

Audit of marmot-protocol/mdk#1457 @ ec99595 (initial review, 52/52 patches, coverage complete, pr_discussion empty, CI all-successful at head).

Both independent reviews converged on zero actionable findings, and every candidate concern raised across the reviews and cross-verifications was reconciled against the exact-head patches and discarded:

- Phantom-membership on Welcome-intent persistence failure — discarded: on intent-record error the client rolls back the prepared commit, refreshes state, and returns before publishing; the pending-Welcome index is reconciled from engine `outstanding_welcome_deliveries`, so no phantom membership is possible. Batch intent insert is transaction-wrapped with a rollback regression test (`storage-sqlite/pending_welcome_delivery.rs`).
- Queued invite returning success without Welcomes — discarded: the `Queued` branch has no Welcome payloads; post-reply fanout runs only when `accept_disposition == Published`.
- Drain/shutdown racing in-flight fanout — discarded: `serve_snapshot_reads_until` defers `Drain` FIFO behind fanout-owning work, startup recovery is awaited on `Drain`, and the one-shot CLI awaits `drain_in_flight_work()` before `shutdown()`.
- Direct `AppClient` still awaiting fanout — discarded: intentional, documented in the patch to preserve the complete path for existing tests; managed workers reply first.
- Codec/migration risk — verified safe: `StagedInviteWelcome` is appended as payload variant 5 with loud unknown-variant decode rejection; `initial_admins` is `#[serde(default)]`; confirm/rollback/disband transitions move staged rows to Sent/Failed atomically, bind promotion to the exact origin commit id, and retire mismatched or legacy raw rows instead of delivering them.
- Invite-with-admin authorization — verified: `initial_admins` must be among invitees (engine error otherwise), duplicate/existing admins are skipped, and `ensure_group_admin` gates the UniFFI entry point.
- CLI `published: 2 -> 1` count change and widened accept-retry (`UnknownGroup`) — intentional, consistent across all call sites, tests updated.
- `dev_fail_invite_*` fault-injection fields — gated behind `test-policy-overrides`, not production-reachable.
- Validators `validate_admin_leaf_coupling_for_staged_commit` / `validate_app_component_integrity_for_staged_commit` bodies are pre-existing code outside the diff; their wiring is covered by new engine and relay-runtime tests. Verification note only, not a finding.

No actionable findings.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Fixed in e0693a13. Startup recovery now prepares and reserves the exact durable Welcome attempts under the serialized account owner, runs only bounded relay I/O in a separately owned task, and reconciles its results through a high-priority branch in the normal worker loop. Inbound delivery, timers, media, maintenance, and commands remain pollable; Drain joins reconciliation, and duplicate retry paths skip the reserved message ids. The restart regression now uses a fresh relay plane, blocks the retained Welcome, proves Alice ingests Bob’s group message and handles an unrelated mutation while the retry remains blocked, then proves Drain waits and the exact obligation retires after release.

crates/cli/src/commands/groups.rs
erskingardner: Fixed in e0693a13. The one-shot group wrapper now drains only Create and Invite; the groups wrapper drains only Create and AddMembers. Members, Remove, Update, avatar, List, Show, and RemoveMembers shut down without joining unrelated retained Welcome recovery. Added focused unit coverage for both positive and negative command sets.

Datawav (commented):
## Automated review

All four artifacts reconciled: both independent reviews converged on zero findings for this delta, and both reciprocal cross-verifications confirmed no false positives and no missed defects — the only residuals were review-quality notes (missing truncation disclosure, an unproven "bounded wait" adjective), not code findings. pr_discussion is empty (enabled=false, no items), so no leads to validate. The CI-in-progress and snapshot-truncation notes are process/CI status, not actionable findings, so they are omitted.

Final audit — marmot-protocol/mdk#1457 @ e0693a1 (follow-up, delta ec99595 → e0693a1)

Outcome: no new actionable findings. The delta is sound at head.

Prior-finding disposition: the previous audit closed with zero findings and zero requested changes, so nothing was pending to fix or verify as resolved. The one prior verified claim this delta reworks — drain/shutdown must not race in-flight Welcome fanout — was re-derived from the new patches and still holds: detached recovery reserves exact message ids under the serialized owner with no await between prepare and reserve (runtime.rs +1937); both publish paths skip reserved ids (+2817 early-return; +3234 synthetic PublishStatus with retry_deferred and met_required_acks=false, so deferred welcomes are never retired prematurely); reservations release exactly once via finish/abandon on both Ok and Err arms; Drain is intercepted at FIFO pop time and parked in drain_waiters until reconciliation completes (account_worker.rs +1053–1067), preserving ordering against queued mutations; panic/cancel is covered by WelcomeRecoveryTask::drop abort plus durable re-prepare from outstanding_welcome_deliveries at next startup, so no reservation can outlive the runtime.

Issues newly introduced by this delta: none. Every candidate raised across the reviews and cross-verifications was traced to changed lines and discarded:
- Duplicate-publication race between drain fanout and retry prepare — closed by the two reservation skip checks.
- Reservation leak on panic or shutdown — closed by Drop-abort plus in-memory-only reservation dying with the runtime.
- Drain-waiter starvation — waiters are flushed after reconciliation on both Ok and Err arms; the unbounded publish wait is a pre-existing tradeoff identical in shape to the old inline recovery.await, not introduced here.
- Ungated fanout-launching CLI commands — the gating table (Create/Invite/AddMembers) was checked against the full command match; no fanout-launching variant is missing, and unit tests pin both positive and negative variants. shutdown() remains unconditional.
- Test quality — the new blocked-relay relay_runtime assertion exercises the exact production behavior this delta introduces and fails on the old inline-blocking implementation.

No actionable findings.

PR #1458, opened 2026-08-15
Add cross-route restart convergence campaigns

erskingardner:
## Summary

- add a deterministic twelve-case restart catalog over the current-build four-party cross-route scenario
- run the catalog through isolated app-runtime and exact retained-relay campaign families with strict state, profile, payload, pending-work, and decryptability oracles
- reuse one canonical public IR across app, process, container, and VM-capable adapters and require restart lifecycle evidence
- record reproducible exact counterexamples and remaining assurance work in the convergence tracking plan

## Why

The existing cross-route scenario proved its terminal path, but did not systematically reopen participants at unfinished durable/publication boundaries. This adds that detector and preserves exact inputs and failure evidence before any production remediation is attempted.

The reviewed seed-0 public catalog passed all 12 app-runtime cases. The corresponding exact catalog passed 9 of 12 and reproducibly exposed three failures at early Zeta publication boundaries: a structured `PendingPublish` refusal, a `PendingPublish` panic, and an epoch-3 pending-work/decryptability stall. The later nine restart boundaries passed, including all four restarts after the corrected first full-history repair round.

## Scope and impact

This PR changes simulator machinery, tests, campaign commands, and documentation only. It does not change the convergence engine, app runtime, storage behavior, protocol, or production constants. Minimization, root-cause diagnosis, and production fixes will follow separately.

## Review follow-up

- replace positional tick heuristics with semantic reconnect/sync/tick anchors
- pin the exact insertion step and restart client for all twelve catalog entries
- give exact and public artifacts distinct scenario names
- keep the restart catalog admin-policy observation in its shared portable public IR while preserving the digest-pinned baseline recovery IR
- derive process checkpoint counts and stable observation pairs from each canonical schedule
- assert exact `NoPendingWork` coverage
- align command defaults with documented seed 0 and document production-policy execution

## Validation

- `cargo test -p cgka-conformance-simulator --lib --locked` — 154 passed
- `cargo test -p cgka-conformance-simulator --bin cgka-conformance-campaign --locked` — 8 passed
- `cargo test -p convergence-campaign-runner --test distributed_runner container_manifest_selects_the_shared_four_party_cross_route_ir --locked`
- `cargo test -p cgka-conformance-simulator --test process_orchestrator four_party_cross_route_recovery_processes_match_unified_route --locked -- --exact`
- `just fast-ci`
- public app-runtime seed-0 catalog — 12/12 passed
- exact retained-relay seed-0 catalog — 9/12 passed with 3 preserved counterexamples

## New Features
- Added 12-case cross-route restart permutation campaigns covering publication, ingestion, branch-witness, and retained-history repair scenarios.
- Added public and exact validation modes with deterministic seed and case selection.
- Added exact application-payload checks that preserve duplicate values while ignoring order.
- Added campaign commands for configurable seeds, case counts, output locations, and timeouts.

## Documentation
- Documented campaign coverage, execution commands, artifacts, results, and known exact-mode failures.


crates/cgka-conformance-simulator/src/cross_route_scenario.rs
```diff
@@ -346,15 +462,106 @@ fn build_cross_route_app_runtime_recovery_scenario(strict_engine_tail: bool) ->
...
+            })
+        }
+        CrossRouteRestartBoundaryV1::FirstFullHistoryRepairCompleted => steps
+            .iter()
+            .enumerate()
+            .filter_map(|(index, step)| {
+                matches!(step, ScenarioStep::Tick { clients: observed } if observed == clients)
+                    .then_some(index)
+            })
+            .next_back(),
```
erskingardner: Non-blocking: the catalog/doc wording says this boundary is "after the first complete retained-history repair", but `.next_back()` lands after the *second* full-history sync/tick round (the builder emits two consecutive full sync+tick pairs before the tail). Functionally fine — arguably intentional to let repair settle — but the name understates where the restart lands.

More generally, both this arm and `AlphaRootIngestedByZeta` (first zeta-only `Tick`) are positional first/last-tick heuristics: if someone later adds another all-clients or zeta-only tick to the scenario, the restart silently retargets and the `unwrap_or_else` panic only guards complete absence. The relay-visibility rebinding got an explicit anti-retarget note in the README; the same one-line caveat here (or anchoring to "tick immediately following a full-history `SyncRelayHistory`") would keep the reviewed boundary vocabulary honest as the scenario evolves.

crates/cgka-conformance-simulator/src/cross_route_scenario.rs
```diff
@@ -346,15 +462,106 @@ fn build_cross_route_app_runtime_recovery_scenario(strict_engine_tail: bool) ->
             }),
         ]);
     }
+    let name = restart_permutation.map_or_else(
+        || "cross-route-app-runtime-recovery/v1".into(),
+        |permutation| {
+            format!(
+                "cross-route-restart-permutations/v1/{}-{}",
```
erskingardner: Non-blocking: exact-family cases produce scenario IRs named `cross-route-restart-permutations/v1/{id}-{client}` — identical to the public family's names — even though they're registered and reported under `cross-route-exact-restart-permutations/v1`. Artifact paths key off `family_name` so nothing collides on disk, but the name embedded in the saved input/report is ambiguous about which subject/oracle produced it, and the repo invariant asks for names that stay easy to grep across layers. Consider threading the family name (or an `exact` marker) into the scenario name for the exact builder.

Justfile
```diff
@@ -367,6 +367,17 @@ convergence-weekly-lane: convergence-nightly-lane
     rm -rf -- target/cgka-weekly-reliability
     cargo run -p cgka-conformance-simulator --features test-policy-overrides --bin cgka-conformance-campaign --locked -- --cases 4 --case-timeout-secs 300 --out target/cgka-weekly-reliability --storage file
 
+# Current-build four-party route assurance with one additional durable reopen
+# per reviewed transition boundary. Twelve cases cover the complete v1 catalog;
+# larger counts repeat it under the same deterministic seed rotation.
+cross-route-restart-campaign seed="1" cases="12" out="target/cross-route-restart-campaign":
+    cargo run -p cgka-conformance-simulator --bin cgka-conformance-campaign --locked -- --family cross-route-restart-permutations/v1 --seed "{{seed}}" --cases "{{cases}}" --case-timeout-secs 300 --out "{{out}}" --storage file
```
erskingardner, line 375: Question: the other campaign lanes (`convergence-nightly-lane`, `convergence-weekly-lane`, adversarial CI) run the campaign binary with `--features test-policy-overrides`; these two recipes don't. If that's deliberate (production-shaped timeouts for assurance runs) a one-line comment saying so would help; if it's an omission the recipes may behave differently from the rest of the campaign surface.

erskingardner (commented):
Simulator/assurance-only change, well scoped and honestly documented. I verified the twelve boundary matchers against the actual step list (all resolve to the intended steps today), the post-insertion rebinding of relay-visibility selectors to stable action ids (matches the previous hardcoded step-16/step-25 ids on the baseline), the stricter restart-lifecycle validation against the orchestrator's "restarted" events, and family dispatch through the campaign binary. No blocking findings; a few non-blocking observations inline. Note: most CI lanes were still pending at review time, and only public case 0 of the app-runtime family has been executed (the exact catalog ran fully) — the payload-multiset oracle shape is identical across cases, so that gap is low-risk, but the remaining public cases are still unrun evidence.

crates/cgka-conformance-simulator/src/cross_route_scenario.rs
```diff
@@ -346,15 +462,106 @@ fn build_cross_route_app_runtime_recovery_scenario(strict_engine_tail: bool) ->
...
+                        action.as_ref(),
+                        ScenarioStep::SendAppMessage { payload, .. }
+                            if payload == "zeta-branch-witness"
+                    )
+            )
+        }),
+        CrossRouteRestartBoundaryV1::AlphaRootIngestedByZeta => {
+            steps.iter().position(|step| {
+                matches!(step, ScenarioStep::Tick { clients } if clients == &["zeta".to_owned()])
+            })
```
erskingardner, line 537: These two matchers are the weak point in an otherwise named catalog. `PublicationAccepted` and the witness payload are unique; `AlphaRootIngestedByZeta` is “first zeta-only tick” and `FirstFullHistoryRepairCompleted` is “last all-client tick in the prefix.” Both are unique *today*, but a later drain/tick in this prefix would silently move the restart without failing `restart_catalog_is_complete_unique_and_compilable`.

Please pin the locus in that test: for each catalog entry, the inserted `RestartClient` must be the immediate next step after the named boundary, with the expected client. Something as small as “after-create-accepted → restart zeta immediately after the `create` ack” would lock the contract the 9/12 result is attributed to.

crates/cgka-conformance-simulator/src/cross_route_scenario.rs
erskingardner: Exact retained-engine cases get this same `cross-route-restart-permutations/v1/{id}-{client}` name even though the family is `cross-route-exact-restart-permutations/v1`. Reports, fixture stems, and failure capsules then look like public catalog artifacts. Include the subject or family in the scenario name (or derive the name from `family_name`) so the seed-0 evidence trail stays greppable.

erskingardner (commented):
The detector belongs here, and the sequencing is right: land the bounded catalog and preserve exact inputs before any production fix. Binding relay visibility to publication names *after* the extra restart is the important robustness fix, and keeping the exact family on the isolated campaign runner is justified by the panic-shaped case.

This is not a clean approve yet. The catalog’s whole claim is “reopen at this named boundary,” but the tests only check that some extra `RestartClient` exists and the IR compiles. Two of the twelve matchers are structural (`first` zeta-only tick, `last` all-client tick) and would silently retarget if the prefix grows. Pin the insertion locus before treating the 9/12 seed-0 result as a map of *those* boundaries.

Smaller follow-ups, not merge blockers:
- Exact and public cases share the `cross-route-restart-permutations/v1/...` scenario name, so artifact grepping cannot tell which subject produced a capsule.
- The public family appends `ObserveAdminPolicy` after the shared builder, so generated app-runtime inputs are not the portable public IR that process/container/VM reuse.
- The exact-family unit test never asserts `NoPendingWork`, even though that is part of the claimed oracle.
- Default `just` seed is `1` while the recorded counterexamples are seed `0`; with seed `1` the panic case runs first. Harmless with process isolation, but easy to misread.

Do not encode the three early Zeta failures as allowed results. Once the insertion-locus test lands, this is in good shape as discovery machinery.

erskingardner (commented):
Re-review of 68960cf. Previous items are closed: semantic reconnect/sync/tick anchors, pinned insertion indices, distinct exact/public names, `NoPendingWork` coverage, and seed-0 defaults. The first-repair matcher now lands after Observer reconnect rather than the second settling tick.

**Blocking:** moving `ObserveAdminPolicy` into the shared public builder also mutates `cross-route-app-runtime-recovery/v1`, which is digest-pinned by `reviewed_release_claim_pins_the_scenario_and_complete_mutation_catalog`. That test still expects `1d416034…`. Updating the claim hash without a new mixed-build run would misattribute release evidence. Keep the extra observe on the restart catalog only so the pinned recovery IR stays unchanged.

That also keeps the process stable-checkpoint pair as the two terminal `Observe`s. With the extra step on the baseline, `first_settled`/`settled` become the second `Observe` and `ObserveAdminPolicy` instead.

crates/cgka-conformance-simulator/src/cross_route_scenario.rs
```diff
@@ -335,6 +451,9 @@ fn build_cross_route_app_runtime_recovery_scenario(strict_engine_tail: bool) ->
         steps.push(in_group(ScenarioStep::Observe {
             clients: clients.clone(),
         }));
+        steps.push(in_group(ScenarioStep::ObserveAdminPolicy {
+            clients: clients.clone(),
+        }));
```
erskingardner: This extra step is on every public form, including the digest-pinned `cross_route_app_runtime_recovery_public_scenario()`. `ConvergenceEvidenceClaimV1` in `crates/convergence-campaign-runner/release-claims/cross-route-mixed-build.v1.json` still pins `canonical_ir_sha256` `1d416034df130ffeb7490c29c5ef9620ed0dce5a53aaf62bc00a2db74104dcc6`, and `reviewed_release_claim_pins_the_scenario_and_complete_mutation_catalog` asserts equality with the live builder.

Do not rewrite that claim to match this IR unless a new mixed-build release run actually produced it. Gate `ObserveAdminPolicy` on `restart_permutation.is_some()` (or an equivalent catalog-only path) so the recovery scenario and its release pin stay intact, while the twelve restart cases still share one portable public IR.

erskingardner (commented):
Re-review of 2c36802. The previous blocker is closed: `ObserveAdminPolicy` is catalog-only, the baseline recovery IR is asserted not to carry it, and the process oracle now takes the last two `Observe` checkpoints rather than whatever happens to be at the tail. Distinct names, pinned insertion loci, semantic repair/ingest anchors, `NoPendingWork`, and seed-0 defaults remain in place. CI is green, including the release-claim digest pin.

Recommend merge. The three exact seed-0 `PendingPublish` failures stay documented counterexamples, not allowed results; diagnosis belongs in a follow-up.

PR #1464, opened 2026-08-15
feat(marmot-uniffi): export keyed chat_list_row through MarmotKit

### LINKED ISSUE #1459: Expose the existing single chat-list row read through MarmotKit
## Problem

Marmot app and runtime already implement a hydrated single-row read as `chat_list_row(account_ref, group_id_hex)`, but MarmotKit does not export it through UniFFI.

White Noise iOS currently has to call `chat_list(account_ref, include_archived: true)` and scan every row when it needs authoritative state for one known conversation. This turns an O(1) storage capability into an O(all chats) bridge transfer and allocation.

## Requested contract

Expose the existing runtime read through MarmotKit with a typed optional result for a missing group and the same account isolation, hydration, and error semantics as the internal method.

## Acceptance criteria

- [ ] MarmotKit exports a read keyed by account reference and group id that returns one hydrated `ChatListRowFfi` or a documented missing result.
- [ ] The implementation delegates to the existing runtime `chat_list_row`; it does not call the full chat list and filter.
- [ ] Unknown account, malformed group id, missing row, quarantined state, and storage failure have deterministic semantics.
- [ ] UniFFI smoke coverage proves the returned row matches the corresponding full-list row.
- [ ] Add performance telemetry or a regression measure demonstrating response work is independent of total chat count.

## Consumer

White Noise iOS can use this for single-conversation archive/lifecycle checks and retire full chat-list reads in known-group paths.

erskingardner:
## Summary
- MarmotKit now exports `chat_list_row(account_ref, group_id_hex)` and delegates to the existing runtime keyed read instead of scanning `chat_list`.
- Missing groups return `None`; unknown accounts, malformed group ids, and closed/failed storage keep the same typed errors as the other chat-list commands.
- Adds `chat_list_row_read` performance telemetry so a single-row read records one sample regardless of total chat count.

Fixes #1459

## Test plan
- [x] New UniFFI test failed before the export existed (`no method named chat_list_row`) and passes after
- [x] `cargo test -p marmot-uniffi` (includes smoke validation plus the full-list match / isolation / archived / StorageClosed coverage)
- [x] `cargo test -p marmot-app records_chat_list_row_read_operation`
- [x] `cargo test -p marmot-app build_export_batch_appends_unlabeled_app_performance_metrics`
- [x] `just fast-ci`
- [ ] White Noise iOS can replace full `chat_list(..., include_archived: true)` scans on known-group archive/lifecycle paths


erskingardner (commented):
Recommend merge.

The keyed export delegates to the existing hydrated runtime read, preserves the typed optional/error behavior, and all exact-head CI checks are green.

Non-blocking: please add `chat_list_row_read` to the collected-operations table in `docs/marmot-architecture/telemetry.md`, including the UniFFI measurement envelope. That page is the current telemetry inventory, so the newly exported operation is otherwise absent from the documented metric surface.

Validation note: PR CI covers the Rust workspace, but the platform binding artifact generation/validation scripts were not exercised by these checks; that remains release validation.

PR #1465, opened 2026-08-15
Fix simulator pending identity across restarts

erskingardner:
## Summary

- scope engine-harness pending labels and outbound lifecycle decisions to the client incarnation that issued them
- prevent accepted publications from an earlier engine process from suppressing confirmation of a new staged evolution that reused the same process-local pending handle
- preserve old-incarnation transport outcomes without allowing them to confirm or roll back new-incarnation engine state
- add a file-backed restart regression and update the convergence reliability tracking plan with the diagnosis

## Root cause

`PendingStateRef` is intentionally process-local. The simulator retained resolved outbound records across an engine restart and compared sibling publications using only the client label and raw pending value. Because a fresh engine restarts its pending counter, a new publication could reuse an older accepted value. The harness would then report the new outbound artifact as accepted but skip `confirm_published`, leaving the actual engine in `PendingPublish`.

That single harness bookkeeping defect caused all three early-Zeta failures from the exact restart catalog. This PR changes simulator and documentation code only; it does not change production engine, storage, app, wire, or protocol behavior.

## Verification

- focused reused-handle regression reproduces the original `PendingPublish` failure before the fix and passes afterward
- exact seed-0 restart catalog: 12/12 passed, up from 9/12
- public app-runtime seed-0 restart catalog: 12/12 passed
- simulator library tests: 155/155 passed
- `just fast-ci`
- `git diff --check`

### DIFF
--- a/crates/cgka-conformance-simulator/src/subject.rs
+++ b/crates/cgka-conformance-simulator/src/subject.rs
@@ -682,6 +682,7 @@ pub struct EngineHarnessSubject {
     active_scenario_group: Option<String>,
     scenario_groups: BTreeMap<String, GroupId>,
     pending_refs: HashMap<String, EngineSubjectPendingRef>,
+    client_incarnations: BTreeMap<String, u64>,
     convergence_clock: ManualConvergenceClock,
     outbound_cursors: HashMap<String, u64>,
     outbound_records: BTreeMap<u64, EngineSubjectOutboundRecord>,
@@ -702,12 +703,14 @@ fn sqlite_database_files(path: &std::path::Path) -> [std::path::PathBuf; 3] {
 #[derive(Clone)]
 struct EngineSubjectPendingRef {
     client: String,
+    client_incarnation: u64,
     pending: PendingStateRef,
 }
 
 #[derive(Clone)]
 struct EngineSubjectOutboundRecord {
     artifact: SubjectOutboundArtifact,
+    client_incarnation: u64,
     pending: Option<PendingStateRef>,
     queued_intent: Option<(GroupId, MessageId)>,
     resolution: Option<SubjectOutboundOutcome>,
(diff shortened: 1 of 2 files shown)


crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -2609,6 +2687,61 @@ mod tests {
...
+            .restart("alice")
+            .expect("file-backed client restarts");
+
+        // The fresh engine starts issuing process-local pending references
+        // from the beginning. The second post-restart update therefore reuses
+        // the reference held by the accepted pre-restart publication.
+        update_and_accept(&mut subject, "after-restart-first").await;
+        update_and_accept(&mut subject, "after-restart-reused-ref").await;
+
+        assert_eq!(
```
erskingardner, line 2738: The comment above this is off, and the final assert does not own the identity defect.

Current-profile create issues no pending, so `before-restart` is pending `1` and **`after-restart-first` is the reuse**, not `after-restart-reused-ref`. The second post-restart update is only a tripwire: if the first accept skipped `confirm_published`, the next `update_group_data` hits `PendingPublish`. The projected group name is not enough on its own — profile commits already show the new name during `PendingPublish`.

If someone later drops `after-restart-first` as redundant, this test stays green without the harness fix.

Pin the collision the same way `pending_publication_identity_is_scoped_by_client` does, and assert the engine actually left `PendingPublish` after each accept (empty pending-work, or equivalent), not just the projected name.

erskingardner (commented):
Recommend merge. The diagnosis is right and the change is correctly scoped to the engine harness. `PendingStateRef` is process-local; retaining accepted outbound records and matching them by raw handle is enough to skip `confirm_published` after a restart. Incarnation-scoping of pending labels, sibling decisions, confirm, and rollback is the right fix. Leaving old-incarnation transport outcomes recorded without letting them mutate the new engine also matches hydrate, which already clears a stranded `PendingCommit` rather than reissuing the old handle.

Queued-intent confirm/retry correctly stay ungated — those keys are durable message ids, not process-local pending refs.

One non-blocking test gap below. I did not re-run the exact catalog from this workspace; format/check/clippy are green and the rest of CI was still running at review time.

(Posted as a comment review because GitHub rejects self-approvals on this account.)

PR #1467, opened 2026-08-15
Document convergence simulator operations and scaling

erskingardner:
## Summary

- add a detailed operator runbook for vectors, generated families, saved-input replay, app/process adapters, containers, VMs, artifacts, and failure handling
- add a scaling guide for deterministic seed sharding, bounded parallelism, evidence retention, adapter promotion, and high-value chaos-family design
- strengthen simulator and distributed-runner AGENTS guidance so agents preserve oracle strength, provenance, privacy, and execution-layer boundaries
- reconcile the convergence reliability plan: record the core defect-discovery lab as operational, defer stronger cross-VM/incident/release assurance behind explicit triggers, and replace the completed implementation sequence with the current operating sequence
- update stale restart-catalog documentation to record #1465 and the current 12/12 exact result

## Why

The simulator stack is now capable, but its operating knowledge was scattered across crate documentation, the tracking plan, source, tests, and prior campaign work. This makes it too easy to choose an unnecessarily expensive adapter, weaken an oracle, lose a reproducer, overlap seed prefixes, or overstate what public process/container evidence proves.

These docs make the existing machinery self-service and give the project a concrete path from a local canary to thousands of deterministic file-backed cases, then to selective process/container/VM confirmation.

## Impact

Documentation and agent guidance only. No production, protocol, engine, simulator, storage, application, or CI behavior changes.

## Validation

- `just fast-ci`
- `cargo metadata --locked --no-deps --format-version 1`
- `git diff --check`
- relative Markdown link validation
- scaling-command shell syntax validation
- tracked filename naming gate

### DIFF
--- a/crates/cgka-conformance-simulator/RUNNING_CAMPAIGNS.md
+++ b/crates/cgka-conformance-simulator/RUNNING_CAMPAIGNS.md
@@ -1,0 +1,294 @@
+# Running Convergence Scenarios And Campaigns
+
+This is the operator manual for the Marmot convergence reliability lab. Start here when you want to run an existing
+vector, generate deterministic scenarios, reproduce a saved case, or move one scenario through more
+production-shaped adapters.
+
+The short rule is: use the cheapest subject that can answer the question, keep strict oracle checking enabled, save
+the exact generated input, and promote only representative cases or failures to processes, containers, or VMs.
+
+For campaign scaling and family development, continue with [`SCALING_CAMPAIGNS.md`](SCALING_CAMPAIGNS.md). For the
+scenario language itself, see [`SCENARIO_IR.md`](SCENARIO_IR.md). The complete scenario registry is in
+[`SCENARIOS.md`](SCENARIOS.md).
+
+## What each execution layer proves
+
+| Layer | Entrypoint | What it adds | What it does not prove |
+| --- | --- | --- | --- |
+| Fixed vectors | `cgka-conformance-simulator-report --vectors ...` | Stable, reviewable semantic regressions | Broad generated coverage |
+| In-process report | `cgka-conformance-simulator-report --family ...` | Fast strict scenario execution and saved reports | OS process isolation or real sockets |
+| Isolated case worker | `cgka-conformance-campaign` | One child per case, timeout/reaping, OS CPU/RSS/write measurements | One OS process per participant |
+| App-runtime adapter | report replay with `--adapter app-runtime` | Production-shaped app/session/projection behavior in one process | Participant process or host isolation |
+| Process adapter | `cgka-conformance-process` | One account-device per child process, private SQLCipher roots, restart | Separate kernels or real network namespaces |
+| Container runner | `cgka-distributed-campaign` / ignored container tests | One participant per container, real sockets, isolated OCI network, network faults | A distinct kernel, block device, or VM host |
+| VM driver | distributed manifest with VM backend | External host lifecycle, kernel/filesystem/block-device isolation | More semantic coverage by itself |
+
+Public app/process/container projections cannot establish exact MLS-private state, durable internal input dispositions,
+or active decryptability when those surfaces are not exposed by the adapter. Pair wider public runs with an
+engine-capable exact control when the claim requires those facts.
+
+## Before the first run
+
+Run commands from the repository root. The workspace pins its Rust toolchain; use `--locked` for CI-like runs. Cargo
+launches supply the simulator's required stack size. When invoking a built binary directly, set
+`RUST_MIN_STACK=4194304`.
+
+Use a new output directory for every run. The isolated campaign runner intentionally refuses to overwrite earlier
+evidence. Keep generated inputs and failure artifacts private: sensitive replay capsules can contain recipient
+database/OpenMLS state and key material and must never be committed.
+
+For a quick installation and compile check:
+
+```sh
+just simulator-smoke
+```
+
+## The five-minute local workflow
+
+### 1. Run a small strict generated batch
+
+```sh
+cargo run -p cgka-conformance-simulator --bin cgka-conformance-simulator-report --locked -- \
+  --family chat-journey/v1 \
+  --seed 42 \
+  --cases 3 \
+  --storage file \
+  --out target/convergence-manual/chat-journey-seed-42
+```
+
+Strict oracle checking is the default. A zero exit means every scenario completed and its declared expectations were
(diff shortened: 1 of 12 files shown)


Datawav (commented):
## Automated review

I'm experiencing a tool-execution failure in this session — my terminal calls to read the bundle and review files are not being executed (the results return empty, with no function output at all), so I cannot inspect `/opt/data/.hermes/state/marmot_pr_audit_artifacts/marmot-protocol_mdk/1467/60e7c3c9aca639eddf431ceb9c8bdc11e2669c37/bundle.json` or `reviews.json` to reconcile the claims.

I won't fabricate an audit from data I could not actually read. To produce the crystallized final audit, I need working tool access. Please either:

1. Retry this request (a fresh turn may restore tool execution), or
2. Paste the contents of `reviews.json` (and the relevant `files[]` patches from `bundle.json`) directly into the chat, and I'll reconcile and emit the under-3500-char audit immediately.

For reference, the exact command I was attempting to run is:

python3 -c "import json; b=json.load(open('/opt/data/.hermes/state/marmot_pr_audit_artifacts/marmot-protocol_mdk/1467/60e7c3c9aca639eddf431ceb9c8bdc11e2669c37/bundle.json')); print(json.dumps({'mode': b.get('review_mode'), 'title': b.get('pr',{}).get('title'), 'files': [f.get('filename') for f in b.get('files',[])]}, indent=1))"

and a full read of `reviews.json` in the same directory.

crates/cgka-conformance-simulator/SCALING_CAMPAIGNS.md
```diff
@@ -0,0 +1,202 @@
...
+Thousands of engine cases and dozens of process/container cases usually provide more useful coverage per hour than
+thousands of VM cases. Wider execution should confirm layer-specific behavior and representative minimized failures,
+not duplicate the entire cheap corpus.
+
+## Reproducible case identity
+
+The stable logical identity is:
+
+```text
+(family or profile version, seed, case index)
```
erskingardner: **[P1] Include `generator_version` in the stable identity**

The schema has separate `family_name` and `generator_version` fields, and those versions do not move together: for example, the family is still `convergence-chaos/v1` while its current generator version is `6`. The tuple shown here can therefore identify multiple different canonical inputs. Please state the identity as `(family_name, generator_version, seed, case_index)` (plus a profile version where applicable) everywhere this contract is repeated; otherwise an operator can regenerate a different case while believing the tuple is exact.

crates/cgka-conformance-simulator/SCALING_CAMPAIGNS.md
```diff
@@ -0,0 +1,202 @@
...
+Build once, then invoke the worker binary directly so parallel shards do not contend on Cargo compilation:
+
+```sh
+cargo build -p cgka-conformance-simulator --bin cgka-conformance-campaign --locked
+```
+
+For example, run eight seed shards with bounded parallelism:
+
+```sh
+campaign_root="target/convergence-scale/chaos"
```
erskingardner: **[P1] Bind scaled artifacts to the exact tested source**

This example stores only family and seed. Neither `process-campaign.v1.json`, the generated-input envelope, nor `ScenarioReportMetadata` records a Git commit, so these artifacts cannot identify which implementation binary produced a failure; the paragraph below nevertheless requires a directory per source revision. Make the staged recipe require an exact/clean build, include that revision (or a binary digest) in `campaign_root`, and persist the revision plus command matrix beside the shard summaries.

crates/cgka-conformance-simulator/RUNNING_CAMPAIGNS.md
```diff
@@ -0,0 +1,294 @@
...
+Useful repository entrypoints are:
+
+```sh
+just simulator-smoke
+just convergence-nightly-lane
+just convergence-weekly-lane
+just focused-convergence-regressions
+```
+
+Nightly and weekly lanes evaluate checked-in wall-clock, CPU, RSS, disk, artifact, retry, and flake budgets. Their
```
erskingardner: **[P1] Do not present the bare Just recipes as budget-evaluated lanes**

The listed `convergence-nightly-lane`/`convergence-weekly-lane` recipes only run the core tests; `observe-step`, `collect-observation`, and `check-budget` are applied by the GitHub workflows around them. The weekly recipe also removes its fixed output directory before running. As written, an operator can get a green result without any wall/CPU/RSS/disk/artifact/retry/flake evaluation and can overwrite prior evidence. Please label these as core test recipes and point to the workflow dispatch for a budgeted retained run, or document a complete local observe/collect/check sequence with unique evidence roots.

erskingardner (commented):
Request changes before merge. The new manuals currently leave generated-case identity and tested-source provenance ambiguous, and the scheduled-lane section overstates what the listed local recipes validate. I posted this as a comment review because GitHub does not allow an author to request changes on their own PR. CI is green except for the cancelled Hermes installer-auth job, which is unrelated to these documentation-only changes.

crates/cgka-conformance-simulator/SCALING_CAMPAIGNS.md
```diff
@@ -0,0 +1,242 @@
...
+generator_version="6"
+case_count="25"
+case_timeout_secs="300"
+parallelism="4"
+
+cargo build -p cgka-conformance-simulator --bin cgka-conformance-campaign --locked
+
+umask 077
+campaign_root="target/convergence-scale/$source_revision/convergence-chaos-v1-g$generator_version"
+mkdir -p "$campaign_root"
```
erskingardner: **[P1] Refuse the campaign root before writing its manifest**

The workers only refuse existing per-case artifacts. This `mkdir -p` accepts an earlier campaign root, and the `>` below truncates its sole `campaign-matrix.txt` before any worker gets a chance to reject an existing shard. Re-running the same revision/family with a changed seed range, case count, or parallelism can therefore pair old artifacts with a new matrix; adding new seed directories also makes the final `find` count include both runs. Make the whole run root unique and absent (for example, include the seed matrix/run id and create the leaf without `-p`), or explicitly fail if either the root or matrix already exists and write the manifest no-clobber. That preserves the immutable-evidence contract stated below.

erskingardner (commented):
Request one more change before merge. The three findings from the previous review are resolved and all CI checks are green at this head, but the new source-bound scale-out recipe can overwrite the provenance manifest of retained evidence before its workers reject an existing run. I posted this as a comment review because GitHub does not allow an author to request changes on their own PR.

PR #1468, opened 2026-08-15
feat(uniffi): include admin identifiers on bounded membership pages

### LINKED ISSUE #1461: Add admin identifiers to bounded group membership pages
## Problem

Issue #1314 added `group_member_ids_page`, collapsing member identifiers for up to 100 groups into one worker command. Recipient selection screens also need admin identifiers to decide which existing groups the caller may add a person to.

White Noise iOS now gets member IDs in bounded pages, then still calls `group_roster` once per group when admin metadata is requested. This restores O(N) worker and UniFFI fanout on profile and add-to-group presentation paths.

Issue #1297 added the excellent single-group roster projection, but it is intentionally a detail-screen contract rather than a bounded multi-group read.

## Requested contract

Either add `admin_ids_hex` to each bounded `AppGroupMemberIdsFfi` result or provide an equivalent bounded roster-identifier page. Keep the page lightweight and profile-free.

## Acceptance criteria

- [ ] Member and admin identifiers for N groups are available in O(N / 100) worker commands.
- [ ] Results preserve input order and retain the existing page size, unknown-group, and quarantine semantics.
- [ ] The read is served from durable/local worker state without profile enrichment.
- [ ] UniFFI and startup-hydration tests prove admin identifiers are correct before deferred profile hydration completes.
- [ ] A regression measure demonstrates that requesting admin metadata for 100 groups does not issue 100 `group_roster` commands.

## Consumer

White Noise iOS `RecipientDirectory` can retire its per-group `group_roster` admin-enrichment loop.

erskingardner:
## Summary
- Bounded `group_member_ids_page` rows only returned member identifiers, so White Noise iOS still issued one `group_roster` call per group for admin metadata.
- Each page row now includes `admin_ids_hex` from durable admin-policy state, with no profile enrichment and the same page-size / unknown-group / quarantine semantics.

Fixes #1461

## Test plan
- [x] New UniFFI conversion test failed before the field existed and passes after
- [x] Startup-hydration hold proves admin identifiers before deferred profile hydration and that the page does not increment `group_roster_read`
- [x] Relay-runtime catch-up test asserts creator-only admin identifiers on the page
- [x] `just fast-ci` passed
- [ ] Host can retire the per-group `group_roster` admin-enrichment loop after consuming `admin_ids_hex`

## Summary by CodeRabbit

* **New Features**
  * Group member-ID results now include administrator IDs alongside member IDs.
  * Administrator information is available through the app and FFI interfaces.
  * Results use stored group information for faster, bounded roster lookups.

* **Bug Fixes**
  * Improved reliability when group roster data is unavailable, failing safely instead of producing incomplete results.

* **Tests**
  * Added coverage confirming administrators are included and regular members are not incorrectly identified as administrators.

### DIFF
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -1032,7 +1032,9 @@ impl AppClient {
         group_ids
             .iter()
             .map(|group_id| {
-                self.ensure_group(group_id)?;
+                let group_record = self
+                    .state_group_record(group_id)
+                    .ok_or_else(|| AppError::UnknownGroup(hex::encode(group_id.as_slice())))?;
                 let member_ids_hex = self
                     .runtime
                     .members(group_id)?
@@ -1042,7 +1044,7 @@ impl AppClient {
                 Ok(crate::AppGroupMemberIds {
                     group_id_hex: hex::encode(group_id.as_slice()),
                     member_ids_hex,
-                    admin_ids_hex: self.admin_policy_for_group(group_id).admins,
+                    admin_ids_hex: group_record.admin_policy.admins,
                 })
             })
             .collect()
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -352,12 +352,20 @@ impl GroupReadSnapshot {
             .iter()
             .map(|group_id| {
                 let members = self.members(group_id)?;
+                let admin_ids_hex = self
+                    .groups
+                    .get(group_id)
+                    .ok_or_else(|| AppError::UnknownGroup(hex::encode(group_id.as_slice())))?
+                    .admin_policy
+                    .admins
+                    .clone();
                 Ok(crate::AppGroupMemberIds {
                     group_id_hex: hex::encode(group_id.as_slice()),
                     member_ids_hex: members
                         .into_iter()
                         .map(|member| member.member_id_hex)
                         .collect(),
+                    admin_ids_hex,
                 })
             })
             .collect()
@@ -1034,6 +1042,7 @@ impl AppClient {
                 Ok(crate::AppGroupMemberIds {
                     group_id_hex: hex::encode(group_id.as_slice()),
                     member_ids_hex,
+                    admin_ids_hex: self.admin_policy_for_group(group_id).admins,
                 })
             })
             .collect()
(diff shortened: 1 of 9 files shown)


Datawav (commented):
## Automated review

No actionable findings.

erskingardner (commented):
Blocking finding at exact head `a52a59e641b7`: the live bounded page can silently erase admin metadata and pays an avoidable MLS reload per row. CI is green, but this should be fixed before merge. GitHub records this as a comment because the authenticated reviewer owns the PR.

crates/marmot-app/src/client/mod.rs
```diff
@@ -1034,6 +1042,7 @@ impl AppClient {
                 Ok(crate::AppGroupMemberIds {
                     group_id_hex: hex::encode(group_id.as_slice()),
                     member_ids_hex,
+                    admin_ids_hex: self.admin_policy_for_group(group_id).admins,
```
erskingardner: **[P1] Read admin IDs from the durable page projection**

`admin_policy_for_group` calls `runtime.admin_pubkeys`, and that helper converts any load/decode/backend error into `AppGroupAdminPolicyComponent::new(Vec::new())`. A steady-state page can therefore succeed with `admin_ids_hex = []` for a real admin, while the catch-up snapshot branch and `group_roster_session` return the durable `state.groups[*].admin_policy`. `Engine::admin_pubkeys` also executes `MlsGroup::load`, so this line adds one OpenMLS group deserialization per requested row (100 for a full page). Please source the admin IDs from the same durable group projection as the snapshot/roster paths (and fail if the row is unexpectedly missing), then cover the steady-state path so it cannot silently turn an admin read failure into ‘no admins’.

erskingardner: Addressed in e27684a0.

The live page now reads `admin_ids_hex` from the same durable `state.groups` row as `group_roster_session` / the catch-up snapshot, and fails closed with `UnknownGroup` if that row is missing. It no longer goes through `admin_policy_for_group` / `MlsGroup::load`.

`member_ids_page_reads_admins_from_the_durable_group_projection` keeps a synthetic durable admin list that diverges from live MLS, then clears the row to prove a miss cannot become `admin_ids_hex = []`.

erskingardner (commented):
Re-review at exact head `e27684a028d3`: the prior P1 is addressed. The steady-state page now sources admin identifiers from the same durable group projection as snapshot and roster reads, fails closed when that row is absent, and avoids the per-row OpenMLS reload. The new regression covers both source selection and the missing-row failure. No remaining actionable findings; the full CI matrix is green. GitHub records this as a comment because the authenticated reviewer owns the PR.

PR #1471, opened 2026-08-16
feat(marmot-uniffi): add peer-keyed existing direct-conversation lookup

### LINKED ISSUE #1463: Add authoritative existing direct-conversation lookup keyed by peer
## Problem

Before creating a direct conversation, clients need to know whether an active direct group with that peer already exists. MarmotKit has no keyed lookup for this.

White Noise iOS currently reads the complete chat list, filters candidate direct rows, requests bounded member pages, and scans for the peer. That is much better than per-group roster calls, but starting a conversation still scales with total chat count and reimplements direct-group eligibility and tie-breaking in the host.

## Requested contract

Expose an authoritative read keyed by account reference and peer account ID that returns the reusable direct conversation, if any. Prefer a typed result containing the selected group ID and enough lifecycle/membership state to explain whether it is reusable.

## Acceptance criteria

- [ ] Marmot app/runtime and MarmotKit expose one lookup by account plus peer ID.
- [ ] Directness, active self-membership, pending/terminal lifecycle states, and duplicate historical direct groups follow one documented MDK policy.
- [ ] When multiple eligible direct groups exist, selection is deterministic and matches durable chat activity ordering.
- [ ] The lookup does not transfer the complete chat list or require host-side membership fanout.
- [ ] Tests cover no match, active match, left/removed/disbanded groups, unknown peer, duplicate historical groups, account isolation, and malformed IDs.
- [ ] A regression measure shows lookup work and bridge payload do not grow linearly with unrelated chats.

## Consumers

White Noise iOS profile and new-chat flows can retire full-list plus membership-page scans, and other clients can share the same reuse policy.

erskingardner:
## Summary
- Hosts had to scan the full chat list and page membership to decide whether an active DM with a peer already existed, reimplementing directness, lifecycle, and tie-break policy.
- Marmot app/runtime and MarmotKit now expose `existing_direct_conversation(account_ref, peer_account_id)` and return one typed reuse result. Direct candidates are filtered in SQL; membership is loaded only for those rows. Selection follows durable chat-list activity order.

Fixes #1463

## Test plan
- [x] New policy/storage tests failed before the fix and pass after
- [x] UniFFI lookup covers no match, active match, unknown peer, malformed IDs, account isolation, duplicate historical groups, and a one-sample regression measure
- [x] `just fast-ci` passed
- [ ] `cargo test -p storage-sqlite -p marmot-app -p marmot-uniffi` on CI

## Summary by CodeRabbit

- **New Features**
  - Find and reuse existing direct conversations with a selected contact.
  - View conversation lifecycle, membership, activity, and status details through the public app interface.
  - Improved chat-list filtering and selection based on membership and recent activity.

- **Improvements**
  - Direct-conversation member data is indexed and automatically backfilled for reliable results.
  - Added performance tracking for direct-conversation lookups.
  - Added a retryable status when the conversation index is still being prepared.

- **Tests**
  - Added coverage for selection rules, migration, persistence, ordering, account isolation, invalid contacts, and telemetry.


Datawav (commented):
## Automated review

Both candidate findings verified against exact head. Key adjudication points:

1. **Finding A (both finders, supported):** `direct_conversation_memberships` swallows page + fallback errors (`Err(_)` → per-group fallback; `if let Ok(...)` drops failures), returns `Ok(memberships)` unconditionally → degraded `Ok(None)` → FFI records `result.is_ok()` = success. Real, Low severity.

2. **Finding B (Sol cross-verify only, "missed defect" P3):** Sol claims the swallow "violates the all-or-nothing contract" of `group_member_ids_page`. Adjudication: the doc comment ("fails as a whole... never returns a partial page") describes the *page API's* behavior, not a contract callers must preserve. The new caller intentionally catches the whole-page failure and falls back per group — that is a caller-side degradation design, not a violation of the API contract (the API itself keeps its promise). The actual consequence (a group whose fallback also fails is silently absent → false-negative miss) is exactly Finding A. Sol's own materiality bound (>100 eligible direct groups) is wrong: the fallback path triggers on *any* chunk failure regardless of chunk count. So Finding B is the same root cause as A, dressed in an unsupported "contract violation" framing — discard as duplicate/unsupported; the supported kernel is already Finding A.

3. Kimi's "25 runs successful" was a mechanistic exaggeration (24 success + 1 skipped) — process-level, not a product finding; omit.

4. Kimi's suppressed uppercase-hex note: speculative, both finders suppressed it; discard.

5. pr_discussion: disabled, 0 items — nothing to reconcile.

One more check: does the fallback's `hex::encode(group_id.as_slice())` key match `row.group_id_hex` casing? Group ids from `hex::decode(&row.group_id_hex)` re-encode lowercase; selector looks up `memberships.get(&row.group_id_hex)`. If `row.group_id_hex` had uppercase, the fallback key would miss. Both finders found no evidence of uppercase ids reaching candidates (projection normalizes). Suppressed as speculative — correct.

Also verify the tie-break claim and that `existing_direct_conversation_from_row`'s unconditional `reusable: true` is only reached post-gating — confirmed in the storage patch (constructed only from `selected` after `direct_row_is_reusable` + roster check).

Now the crystallized audit. review_mode=initial, so preserve the single supported finding even though only Sol's independent review found it (Kimi's cross-verification independently re-derived and confirmed it — so it has reciprocal support anyway).

---

**Audit — marmot-protocol/mdk#1471 @ a00a7139 (initial): one supported Low finding; design otherwise sound.**

The peer-keyed DM-reuse lookup is fail-closed in the right direction throughout: SQL prefilters to `TRIM(row.group_name)='' AND ag.member_count=2`, Rust re-gates on lifecycle/membership/leave/disband predicates, and the live roster is revalidated as exactly {local, peer} before selection, so stale projection state can only produce a miss, never a false reuse. Tie-break (activity_sort_at DESC, group_id_hex ASC) is deterministic and tested; telemetry snapshot field is `#[serde(default)]`-additive; no schema or wire change; the negative/boundary test matrix (left/removed/disbanded/leaving/disbanding/named/other-peer/pending/archived/ties, account isolation, typed errors, closed storage) genuinely exercises the production path.

1. Low — Membership-load failures degrade to a false-negative `None` that telemetry records as success. In `crates/marmot-app/src/runtime/mod.rs` `direct_conversation_memberships` (head ~3162–3208), a failed `group_member_ids_page` chunk (`Err(_) =>`) drops to a per-group `group_members` fallback, and `if let Ok(members) = ...` silently discards any group whose fallback also errors; the function then returns `Ok(memberships)` unconditionally. `select_reusable_direct_conversation` treats a missing map entry as `continue`, so the caller returns `Ok(None)`, and the FFI command (`crates/marmot-uniffi/src/commands/chat_list.rs` ~line 90) records `result.is_ok()` — true for this degraded `None`. Consequence: a persistent member-ids failure on a host that does have a reusable DM makes the lookup return `None` forever, so the host recreates a duplicate direct group — the exact #1463 defect this PR fixes — while `app_existing_direct_conversation_read_successes` keeps incrementing, hiding the regression from the metric built to watch it. Smallest sound correction: when a group's page read and fallback both fail, either propagate an error from `direct_conversation_memberships` (the `?` already exists at the call site) or return a degradation flag alongside the map and record the telemetry sample with `success = result.is_ok() && !degraded`.

Discarded during reconciliation: a claimed "all-or-nothing contract violation" in the chunked paging — the `group_member_ids_page` doc comment constrains the API's own response, not its callers, and the per-group fallback is a deliberate caller-side degradation; its real consequence is the finding above, so it is a duplicate with unsupported framing. Also discarded: a "25/25 runs successful" claim (24 succeeded, 1 intentionally skipped — process note, not a product defect) and a speculative uppercase-hex map-key mismatch (no evidence uppercase ids reach candidates; both finders suppressed it).

crates/marmot-uniffi/src/commands/chat_list.rs
```diff
@@ -51,7 +52,48 @@ impl Marmot {
...
+    ) -> Result<Option<ExistingDirectConversationFfi>, MarmotKitError> {
+        let started_at = Instant::now();
+        let result = self
+            .runtime
+            .existing_direct_conversation(&account_ref, &peer_account_id)
+            .await
+            .map(|found| found.map(Into::into))
+            .map_err(MarmotKitError::from);
+        self.runtime
+            .record_existing_direct_conversation_read(started_at.elapsed(), result.is_ok());
```
Datawav, line 96: This UniFFI wrapper orders the operations as `result = runtime.existing_direct_conversation(...).await`, then `record_existing_direct_conversation_read(started_at.elapsed(), result.is_ok())`, then `result`. The witness therefore proves telemetry-before-result on the new FFI surface but does not require a separate `mod commands` test: the focused test lives next to this wrapper instead of duplicating the `chat_list_rows` scenario suite. Please reconcile that matrix entry (or move the test) so witnesses do not point at a file that will never hold the proof.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3117,6 +3127,87 @@ impl MarmotAppRuntime {
...
+            .filter(|row| {
+                row.conversation_kind == crate::ChatConversationKind::Direct
+                    && row.self_membership == crate::SelfMembership::Member
+                    && row.lifecycle_state != cgka_traits::GroupLifecycleState::Disbanded
+                    && !row.disbanding
+                    && row.leave_requested_at_ms.is_none()
+            })
+            .collect::<Vec<_>>();
+        let mut memberships = HashMap::new();
+        for chunk in eligible.chunks(crate::MAX_GROUP_MEMBER_IDS_PAGE_SIZE) {
```
Datawav: `direct_conversation_memberships` currently pages `group_member_ids_page` (batched chunk loop) for every direct-conversation candidate even though the only consumer folds eligibility to switch peers on roster state. In the common case a peer-switch poll only needs a couple of groups; most pages are wasted reads. Consider a cheap two-tier flow: SQL candidates -> reuse selection with existing durable rows -> fetch membership just for the surviving candidate(s), so the membership page is bounded by matches rather than all direct rows.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1670,6 +1789,35 @@ fn chat_list_rows_tx(tx: &Connection, query: ChatListQuery) -> StorageResult<Vec
     Ok(rows)
 }
 
+fn direct_conversation_candidate_rows_tx(tx: &Connection) -> StorageResult<Vec<ChatListRow>> {
+    let sql = format!(
```
Datawav: `direct_conversation_candidate_rows_tx` reuses `CHAT_LIST_ROW_SELECT_AND_JOINS` with its populated-state ordering then applies a pinned-position/global chat-list ordering (`ORDER BY row.activity_sort_at DESC, row.group_id_hex`) without the account-scope predicate the main chat list applies. Verify the pinned-scope join and activity ordering agree with `chat_list_rows_tx` for a peer who also appears in named/3+ groups; a divergent ordering here would reintroduce the tie-break mismatch the doc comment promises to avoid.

crates/storage-sqlite/src/chat_list/tests.rs
```diff
@@ -332,6 +332,230 @@ fn conversation_kind_uses_durable_current_roster_projection() {
     );
 }
 
+fn direct_row(group_id_hex: &str, activity_sort_at: u64) -> ChatListRow {
```
Datawav, line 336: Test scaffolding uses `"aa".repeat(32)`/`"bb".repeat(32)`/etc. as fake group ids and peer/local ids, but the pure `select_reusable_direct_conversation` test and the SQL candidate test share these short ids with production-shaped rosters. Because `direct_conversation_candidate_rows` runs real storage behind these, using non-32-byte hex ids risks passing policy tests while the hex-decode path in `runtime/mod.rs::direct_conversation_memberships` (`hex::decode` on `row.group_id_hex`) silently drops them. Use 32-byte hex fixtures in the SQL test at least, or decode-able ids in both, so the tests exercise the same decode gate the runtime applies.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Make the lookup genuinely peer-keyed before paging rosters. `direct_conversation_candidates` returns every unnamed two-member row, and this loop loads membership for every one before `peer_account_id_hex` participates in filtering. With N direct chats involving other peers, each lookup still materializes N full `ChatListRow`s and reads N rosters; the SQL predicate also has no supporting `group_name`/`member_count` index, so it scans the chat projection. That is the linear work with unrelated chats that issue #1463 explicitly asks to remove. Persist/query a peer-to-group projection (or another indexed peer key) and add a regression that counts candidate/roster work with many other-peer DMs—the current one-sample telemetry assertion only proves one API call was recorded.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3117,6 +3127,87 @@ impl MarmotAppRuntime {
...
+            match self.group_member_ids_page(account_ref, &group_ids).await {
+                Ok(page) => {
+                    for item in page {
+                        memberships.insert(item.group_id_hex, item.member_ids_hex);
+                    }
+                }
+                Err(_) => {
+                    for group_id in group_ids {
+                        let group_id_hex = hex::encode(group_id.as_slice());
+                        if let Ok(members) = self.group_members(account_ref, &group_id).await {
```
erskingardner: Do not turn a failed membership read into a clean miss. When the page read fails and this per-group fallback also errors, the group is silently absent from `memberships`; selection then skips it and `existing_direct_conversation` returns `Ok(None)`. A host will create a duplicate direct group—the exact behavior this API is meant to prevent—and the UniFFI telemetry records the degraded result as a success. Propagate the failure, or carry explicit degradation so a missing candidate cannot be reported as authoritative absence.

erskingardner (commented):
Blocking findings: the new API centralizes reuse policy and keeps the UniFFI result compact, but the implementation still does work proportional to every existing direct chat before applying the peer key, so the lookup-scaling acceptance criterion in issue #1463 is not met. It also converts degraded roster reads into a successful `None`, which can recreate the duplicate-DM behavior while success telemetry increments. CI is green at exact head `a00a71395921008c9811f59bd63e101f9db54149`, but the draft currently conflicts with `master` and will need rebasing plus fresh CI.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Addressed in `4728b13b`. Direct conversations now persist both member hexes in `direct_conversation_members` (indexed on `member_id_hex`), and the candidate query joins that table for the requested peer. Other-peer DMs never enter the candidate set or the roster page. The storage regression seeds 20 other-peer DMs and asserts only the matching rows are returned; the UniFFI test now also asserts `direct_conversation_candidates(label, peer).len() == 2` after extra other-peer DMs.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Addressed in `4728b13b`. `direct_conversation_memberships` now fails closed: a page omission or a failed per-group fallback returns `InvalidGroupMembershipPage` instead of `Ok(None)`. Telemetry therefore records the failure, and the host cannot treat a degraded read as authoritative absence.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Superseded by the persisted peer index in `4728b13b`. SQL now returns only Direct rows that already contain the peer, and membership is loaded only for those reusable-shaped survivors.

crates/storage-sqlite/src/chat_list.rs
erskingardner: Confirmed intentional and documented in `4728b13b`. Reuse follows durable `activity_sort_at DESC, group_id_hex`, not the pin-first order used by `chat_list_rows`. Added `direct_conversation_candidate_rows_follow_activity_not_pin_order` so a pinned older duplicate cannot win selection.

crates/storage-sqlite/src/chat_list/tests.rs
erskingardner: Addressed in `4728b13b`. The SQL candidate tests now use decodeable 16-byte group ids and 32-byte account ids. The pure `select_reusable_direct_conversation` cases still use short labels like `active` because they never go through `hex::decode`.

crates/marmot-uniffi/src/commands/chat_list.rs
erskingardner: Kept the focused proof next to this wrapper in `existing_direct_conversation_lookup_is_keyed_and_independent_of_unrelated_chats` and noted that on the telemetry-before-result call. There is no separate commands-suite matrix file in this repo to move it into.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3138,6 +3148,165 @@ impl MarmotAppRuntime {
...
+        &self,
+        account_ref: &str,
+        peer_account_id: &str,
+    ) -> Result<Option<ExistingDirectConversation>, AppError> {
+        let account = self.accounts.resolve(account_ref)?;
+        let peer_account_id_hex = account_id_hex_from_ref(peer_account_id)?;
+        if peer_account_id_hex == account.account_id_hex {
+            return Ok(None);
+        }
+        self.backfill_direct_conversation_members(account_ref, &account.label)
```
erskingardner: Keep the steady-state peer lookup off this global backfill scan. Every call reaches `unindexed_direct_conversation_group_ids`, even after migration/backfill is complete; that query must search `account_groups`/`chat_list_rows` to prove no empty-name, two-member row lacks index entries because those predicates have no supporting index or durable completion marker. The candidate SQL is also rooted in `chat_list_rows` and uses the peer index only through a correlated `EXISTS`, so the index is a per-row probe rather than the driver. The returned vector is bounded, but the database work still grows with unrelated chats, which leaves the core acceptance criterion from #1463 unmet. Run/backfill this projection once at migration or account-open with a durable completion state, start the lookup from `idx_direct_conversation_members_member`, and add an `EXPLAIN QUERY PLAN` regression that rules out a chat-table scan; asserting `candidates.len()` only measures output size.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3138,6 +3148,165 @@ impl MarmotAppRuntime {
...
+                ));
+            }
+            group_ids.push(GroupId::new(bytes));
+        }
+        for chunk in group_ids.chunks(crate::MAX_GROUP_MEMBER_IDS_PAGE_SIZE) {
+            let page = self
+                .direct_conversation_member_ids_page(account_ref, chunk)
+                .await?;
+            for item in page {
+                storage.replace_direct_conversation_members(
```
erskingardner: Make this backfill write conditional on the projection it read. The roster page and this standalone replacement are not one serialized operation: after the worker answers with `{local, old_peer}`, a concurrent membership change can persist `{local, new_peer}` first, after which this call deletes the newer rows and reinstalls `old_peer`. Because `unindexed_direct_conversation_group_ids` treats any remaining row as indexed, a lookup for `new_peer` then returns no candidate and never reaches the live-roster guard until another projection save repairs the index. Use a write-if-still-unindexed/CAS tied to the same group projection or perform the backfill inside the worker projection save, and cover the page-read versus newer-save interleaving with a regression.

erskingardner (commented):
Re-review at exact head `403bb81e33da78425f507c22c7928dd328c108e7`: the previous degraded-membership false-success path is fixed, and the persisted index bounds the returned candidate set. Two blocking issues remain: steady-state reads still perform work proportional to unrelated chats, and upgrade backfill can overwrite a newer peer projection after a concurrent membership change. Current CI is green and the PR is cleanly mergeable.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Concrete fix sequence: (1) let migration 50 create the table/index only; a SQLite migration cannot obtain live MLS rosters. (2) During account readiness/hydration, run one serialized backfill after the account worker can read those rosters. (3) Persist an account/schema-version completion marker only after every eligible group is indexed successfully; retry on the next open if any group fails. (4) Once the marker is set, remove `backfill_direct_conversation_members` and `unindexed_direct_conversation_group_ids` from the per-lookup path. (5) Rewrite the candidate query with `direct_conversation_members` filtered by `member_id_hex` as the driving relation, then join to `chat_list_rows`/`account_groups`. Add an `EXPLAIN QUERY PLAN` regression over many unrelated chats that requires use of `idx_direct_conversation_members_member` and rejects full scans of `chat_list_rows` or `account_groups` for the lookup.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Preferred fix: perform the backfill read and projection write inside the account worker, on the same serialized lane as membership mutations and normal projection saves. Do not return a roster to the runtime and write it later. If that restructuring is impractical, return the projection version/epoch with the roster and make the storage replacement a CAS that succeeds only while that version is still current and the group is still unindexed. Regression: use a barrier to pause the backfill after its roster read, commit a newer `{local, new_peer}` projection, resume the stale backfill, then assert the newer index rows remain and lookup finds `new_peer` but not `old_peer`.

Datawav (commented):
## Automated review

Verified against the exact-head patch: the swallow is gone, errors propagate via `await?`/`Err(AppError::InvalidGroupMembershipPage)`, and the per-group fallback now uses `group_members(...).await?`. Both independent reviews and both reciprocal cross-verifications concur: the prior Low finding is fixed and no new actionable findings exist. review_mode=follow_up, so the output is the mandated fixed-list form.

- Fixed: Low — membership-load failures no longer degrade to a false-negative `Ok(None)` recorded as telemetry success. `direct_conversation_memberships` (crates/marmot-app/src/runtime/mod.rs) now propagates page reads via `await?`, post-checks `memberships.contains_key(&row.group_id_hex)` and returns `Err(AppError::InvalidGroupMembershipPage("direct conversation membership page omitted a requested group"))` when any eligible group is absent; the per-group fallback in `direct_conversation_member_ids_page` uses `group_members(...).await?` so a fallback failure propagates instead of silently dropping the group. `existing_direct_conversation` propagates via `?`, so the FFI's `record_existing_direct_conversation_read(..., result.is_ok())` (crates/marmot-uniffi/src/commands/chat_list.rs) now records a failure sample for degraded loads — exactly the correction the previous audit prescribed.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Addressed in `0c9df8d0`.

- Migration 50 still creates only the table/index.
- Upgrade backfill now runs once on the account-worker hydration lane (`reconcile_hydrated_account_state`), after rosters are readable, and is gated by `direct-conversation-members-backfill-v1`. The marker is set only when every eligible group is indexed; a not-hydrated group leaves it unset so the next open retries.
- `existing_direct_conversation` no longer calls `unindexed_direct_conversation_group_ids`.
- Candidate SQL now starts from `direct_conversation_members` filtered by `member_id_hex`, then joins chat-list/account rows.
- `direct_conversation_candidate_rows_are_keyed_by_peer` asserts `EXPLAIN QUERY PLAN` uses `idx_direct_conversation_members_member` and does not full-scan `chat_list_rows` or `account_groups`.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Addressed in `0c9df8d0`. The backfill read and write now happen on the account worker, on the same serialized lane as membership mutations and projection saves. The storage write is also fill-if-still-unindexed, so a stale roster cannot replace rows a newer save already wrote. `fill_unindexed_direct_conversation_members_does_not_clobber_newer_rows` covers that interleaving.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3138,6 +3148,131 @@ impl MarmotAppRuntime {
             .chat_list_row(&account.label, group_id_hex)
     }
 
+    /// Look up the reusable existing direct conversation with `peer_account_id`.
+    ///
+    /// `peer_account_id` accepts hex or `npub`. The read is keyed by this
+    /// account plus that peer: it does not return the full chat list, and
+    /// membership is loaded only for peer-index hits that still look reusable.
+    /// Upgrade backfill of the peer index runs once on account open, not on
+    /// this read.
```
erskingardner: Do not let an empty migration index become an authoritative miss before this backfill finishes. The account worker sends `ready` before `run_startup_hydration_pipeline`; `reconcile_hydrated_account_state`, which performs this backfill, runs only afterward. On the first open after migration 50, a caller can reach this lookup while `direct_conversation_members` is still empty, get zero candidates, and return `Ok(None)`, allowing a duplicate DM to be created. If hydration/reconciliation fails, that false miss lasts for the rest of the session because only the next open retries. Gate this read on the completion marker/worker reconciliation: either complete the one-time backfill before exposing this lookup, or return a retryable error while the marker is unset and let the worker complete/retry it. Add an upgrade regression that pauses hydration after readiness, invokes this lookup, proves it never returns `None`, then releases hydration and verifies the existing group is found.

crates/marmot-uniffi/src/commands/chat_list.rs
```diff
@@ -546,4 +591,214 @@ mod tests {
...
+            before.existing_direct_conversation_read.successes + 1
+        );
+        assert_eq!(
+            after.existing_direct_conversation_read.attempts
+                - before.existing_direct_conversation_read.attempts,
+            1,
+            "lookup must record one sample regardless of unrelated chat count"
+        );
+        assert!(found.reusable);
+        assert_eq!(found.group_id_hex, second);
```
erskingardner: Make the activity ordering deterministic here. Current-head `Rust tests (3/4)` failed this assertion on both nextest attempts: the create/send operations produced tied durable activity values, so the documented `group_id_hex` tie-break selected `first` instead of `second`. Advance or mock the clock, set activity through a deterministic test seam, or derive the expected group from the stored `(activity_sort_at DESC, group_id_hex ASC)` order. Avoid a timing sleep; the test should prove policy without depending on wall-clock resolution.

erskingardner (commented):
Re-review at exact head `0c9df8d06eda4ec0279bfe9428a57d01088b80b2`: the previous steady-state scan and stale-backfill overwrite blockers are fixed. One correctness blocker remains: first-open lookup can race the post-readiness migration backfill and return a false authoritative `None`. CI also currently fails `Rust tests (3/4)` because the new ordering test depends on tied wall-clock activity timestamps; Simulator and vectors plus CodeRabbit remain pending. This account authored the PR, so GitHub records this as COMMENTED rather than a change request.

crates/marmot-app/src/runtime/mod.rs
erskingardner: Addressed in `c94b1d82`.

`existing_direct_conversation` now checks `direct-conversation-members-backfill-v1` before reading candidates. If the marker is unset, it returns typed retryable `DirectConversationIndexNotReady` / `MarmotKitError::DirectConversationIndexNotReady` instead of `Ok(None)`.

`existing_direct_conversation_is_not_a_miss_while_upgrade_backfill_is_held` recreates the first open after migration 50 (empty index, marker unset), holds hydration after readiness, asserts the lookup is that retryable error and never `None`, then releases hydration and finds the existing group.

crates/marmot-uniffi/src/commands/chat_list.rs
erskingardner: Addressed in `c94b1d82`.

The UniFFI test no longer assumes `send_text` on `second` produced a strictly later `activity_sort_at`. After pinning `first`, it reads both chat-list rows and derives the expected winner from the documented `(activity_sort_at DESC, group_id_hex ASC)` order, then asserts lookup matches that. Pin order still cannot change the selection.

crates/marmot-app/src/client/projection.rs
```diff
@@ -474,14 +474,19 @@ impl AppClient {
...
+                    projected_group.member_count = u64::try_from(group.members.len()).ok();
+                    dirty |= projected_group.member_count.is_some();
                 }
+                let previous_direct_members = projected_group.direct_member_ids_hex.clone();
+                projected_group.set_direct_member_ids_from_roster(&group.members);
+                dirty |= projected_group.direct_member_ids_hex != previous_direct_members;
+            }
+            if dirty {
+                self.mark_group_projection_dirty_hex(group_id_hex);
+                changed = true;
```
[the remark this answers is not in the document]
erskingardner: Confirmed at `c94b1d82`; this remains blocking. Two torn-write cases make the impact concrete. If storage says `member_count = 3` but the live roster is now two after the engine mutation committed and the projection save failed, reconciliation writes two `direct_member_ids_hex` but preserves count 3; persistence omits the index, the backfill query skips the row, the completion marker is set, and lookup returns a false `Ok(None)`. In the inverse 2-to-3 case, the stale count keeps the row perpetually “unindexed”, so the marker never completes and every lookup stays `DirectConversationIndexNotReady`. Refresh `member_count` whenever it differs from the live roster, derive index persistence from an empty-name projection with exactly two direct IDs, and make the transaction reject every member slice whose length is not exactly two. Add 3-to-2 and 2-to-3 reconciliation regressions that simulate the engine/projection tear and assert marker plus lookup behavior.

erskingardner (commented):
Request changes. Re-review at exact head `c94b1d820baf332b7ec478f23a0d676bf5acead6`: the prior readiness/backfill race and nondeterministic ordering test are fixed, all 25 required checks pass, and the PR is cleanly mergeable. One blocking data-integrity issue remains in the existing `reconcile_live_engine_groups` thread: a torn membership projection can leave `member_count` stale while `direct_member_ids_hex` follows the live roster, causing either a false authoritative `None` or a peer index that never becomes ready. This account authored the PR, so GitHub records the review as COMMENTED rather than a change request.

crates/marmot-app/src/client/mod.rs
```diff
@@ -2046,6 +2046,64 @@ impl AppClient {
...
+            &self.state.label,
+            crate::DIRECT_CONVERSATION_MEMBERS_BACKFILL_MARKER,
+        )? {
+            return Ok(());
+        }
+        let storage = self.app.account_storage(&self.state.label)?;
+        let unindexed = storage.unindexed_direct_conversation_group_ids()?;
+        let mut hydration_pending = false;
+        for group_id_hex in unindexed {
+            let Ok(group_id_bytes) = hex::decode(&group_id_hex) else {
```
Datawav, line 2069: Low — permanent retry lockout on a malformed group id. When `hex::decode(&group_id_hex)` fails here, the loop hits `continue` without setting `hydration_pending`, and that row stays in `account_groups` unindexed forever. Since the backfill marker is only set when `!hydration_pending && unindexed_direct_conversation_group_ids()?.is_empty()`, the second scan never returns empty, the marker never completes, and `existing_direct_conversation` returns `DirectConversationIndexNotReady` on every call for the life of the account — blocking all direct-conversation reuse, not just the corrupt group. The prior head surfaced `InvalidGroupMembershipPage` instead. Smallest sound correction: filter malformed/empty-hex group ids out of `unindexed_direct_conversation_group_ids` (or mark-and-skip them durably) so the marker can complete and reuse is restored for the account's valid groups.

Datawav (commented):
## Automated review

All claims are now reconciled against the exact-head patches. Let me verify my understanding of the M1 finding chain once more (the malformed-hex skip) and confirm nothing else in the delta is unsupported.

The chain: in `backfill_direct_conversation_members_once` (client/mod.rs), a hex-decode failure hits `continue` without setting `hydration_pending`, and such a group can never be filled — so the second `unindexed_direct_conversation_group_ids()?.is_empty()` scan stays false forever, the marker never completes, and `existing_direct_conversation` returns `DirectConversationIndexNotReady` on every call for the life of the account. The prior head instead propagated `InvalidGroupMembershipPage`. That's a genuine, delta-introduced Low-severity liveness degradation supported by patch evidence, found by one cross-verifier and not contradicted by either review.

Both primary reviews converge: prior Low is fixed and preserved; no other new findings. The 26-vs-25 check-run count is a review inaccuracy, not a product finding, so it's excluded. pr_discussion is empty — no leads.

Follow-up audit, PR mdk#1471 @ c94b1d8 (delta 403bb81 → c94b1d8):

Previously requested changes touched by this delta: the one prior Low is fixed and preserved. The read-path backfill swallow is gone; `direct_conversation_memberships` still propagates via `await?`/`contains_key`, and the replacement gate returns typed `AppError::DirectConversationIndexNotReady` before any read, recorded as an FFI failure sample (`record_existing_direct_conversation_read(..., result.is_ok())`). The held-hydration test asserts `Err(DirectConversationIndexNotReady)`, never `Ok(None)`, and is CI-pinned via `--exact` under `test-policy-overrides`. Not regressed.

Newly introduced by this delta — one finding:

- Low — permanent retry lockout on a malformed group id (crates/marmot-app/src/client/mod.rs, `backfill_direct_conversation_members_once`). Evidence: `let Ok(group_id_bytes) = hex::decode(&group_id_hex) else { continue; }` skips without setting `hydration_pending`, and the marker is set only when `!hydration_pending && unindexed_direct_conversation_group_ids()?.is_empty()`. A malformed-hex row in `account_groups` is never fillable, so the second scan never returns empty, the marker never completes, and `existing_direct_conversation` returns `DirectConversationIndexNotReady` on every call for the account's lifetime — blocking all direct-conversation reuse, not just the corrupt group. Consequence: indefinite feature outage misreported as a transient retryable state; the prior head surfaced `InvalidGroupMembershipPage` instead. Smallest sound correction: filter malformed/empty-hex group ids out of `unindexed_direct_conversation_group_ids` (or mark-and-skip them durably) so the marker can complete, restoring reuse for the account's valid groups.

No other new actionable findings. The stale-backfill clobber is closed by the in-transaction EXISTS re-check in `fill_unindexed_direct_conversation_members`; the candidate-read JOIN rewrite is equivalent to the prior EXISTS form under the `dcm.member_id_hex = ?1` binding; retry classification is consistent across MarmotKit and both simulator classifiers.

crates/marmot-app/src/client/projection.rs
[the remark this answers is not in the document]
erskingardner: Addressed in `89bdf6a2`.

`reconcile_live_engine_groups` now refreshes `member_count` whenever it differs from the live roster, together with `direct_member_ids_hex`. Index persistence is derived from an unnamed projection with exactly two member ids, and `replace_direct_conversation_members_tx` independently refuses any other slice length.

`reconcile_repairs_stale_three_member_count_on_two_member_direct` and `reconcile_repairs_stale_two_member_count_on_three_member_group` simulate those tears and assert the backfill marker completes: 3-to-2 lookup finds the existing group, 2-to-3 lookup stays a miss.

crates/marmot-app/src/client/mod.rs
erskingardner: Addressed in `89bdf6a2`.

`unindexed_direct_conversation_group_ids` now omits malformed or empty group-id hex, so a corrupt `account_groups` row cannot keep `direct-conversation-members-backfill-v1` unset. Valid unindexed groups still block the marker until they are filled. Covered by `unindexed_direct_conversation_group_ids_skip_malformed_hex`.

Datawav (commented):
## Automated review

Reconciliation complete. All claims verified against the exact-head patches:

- Prior Low (malformed-hex retry lockout): FIXED — `indexable_group_id_hex` filter in `unindexed_direct_conversation_group_ids` (chat_list.rs) + regression test asserting only the valid id returned.
- No reviewer raised any new actionable finding; all candidate concerns (duplicate-id slice, malformed-index write, whitespace-name boundary, count-reconcile semantics) were re-derived and rejected as unreachable/pre-existing/behaviorally safe.
- Cross-checks disagreed only on line-number citations and the 25-vs-26 check-run count; bundle confirms 25 runs (24 success, 1 skipped release-publish). Neither affects the verdict.
- pr_discussion disabled/empty — no leads.

- Prior Low — permanent `DirectConversationIndexNotReady` lockout on malformed group-id hex: fixed. `unindexed_direct_conversation_group_ids` now filters candidates through `indexable_group_id_hex` (chat_list.rs), excluding corrupt rows so the completion marker can set; pinned by `unindexed_direct_conversation_group_ids_skip_malformed_hex` (chat_list/tests.rs), which fails on the pre-delta head.

PR #1475, opened 2026-08-16
fix(cgka-engine): re-arm queued-outbound drain on pass close; skip in-flight regenerated intents (#1472)

### LINKED ISSUE #1472: Queued outbound intents never release after committer restart mid-resolution (bounded-convergence-pressure livelock)
## Symptom

The `bounded-convergence-pressure/v1` acceptance gate (`bounded_convergence_pressure_family_settles_every_seeded_permutation`, currently `#[ignore]`d) fails on `master` (2a16d80f): an application message accepted while the group resolves a same-epoch fork is queued durably and **never published**, even though the fork fully converges.

The failure shape has changed since the ignore comment was written in #1293:

- Original symptom (pre-#1403): nothing re-armed the retained-intent drain after a completed pass; progress reported `runnable_work = 0`, no next wake, `queued_outbound_intents > 0` → driver `Blocked`.
- Current symptom (post-#1403): the drain **is** armed (`runnable_work = 1` per client) but every drain invocation releases nothing. `await_quiescence` burns all 256 iterations / 1021 work units and times out with `subsystems: ["runnable_work", "scenario_inputs"]`. A stall became a livelock.

Terminal quiescence snapshot (seed 2026, case 0, step 25): all four clients converged — epoch 2, lifecycle `stable`, pass `completed`, zero unresolved convergence inputs, zero deferred/retry/stored work, transport drained, nothing awaiting acknowledgement — yet each still holds queued outbound intents (alice 1, bob 2, carol 1, dave 1) and reports `runnable_work = 1`.

## Reproducer

```sh
cargo test -p cgka-conformance-simulator --test canonical_scenarios --locked -- \
  bounded_convergence_pressure_family_settles_every_seeded_permutation --ignored --exact
```

or with durable evidence artifacts:

```sh
cargo run -p cgka-conformance-simulator --bin cgka-conformance-simulator-report --locked -- \
  --family bounded-convergence-pressure/v1 --seed 2026 --cases 1 \
  --out target/investigate-bcp-seed-2026 --storage file
```

## Leading hypothesis

The settle precondition in `converge_and_drain_queued_outbound_intents` (`advance_before_queued_outbound_intents` → `if !settled return Ok(vec![])`) never reports settled in the committer-restart-mid-resolution shape, so the armed drain spins without releasing. #1403 fixed the arm/projection half; the release half is still broken.

## Notes

- The ignore comment says: un-ignore once the retained-intent drain is re-armed after a completed pass; do not weaken assertions.
- Scenario shape: same-epoch commit race → app sends inside the quiescence window → one committer restarts mid-resolution → bounded self-update/profile/admin tail.

erskingardner:
Closes #1472.

## Root cause (two interlocking drain-lifecycle defects)

The `bounded-convergence-pressure/v1` acceptance gate was ignored because application sends accepted while a convergence pass resolves (e.g. inside the quiescence window of a same-epoch fork) were stranded. Investigation showed **two** defects, both reachable only through edge-driven consumers:

1. **Lost drain edge on pass close.** The drain schedule edge is one-shot: a drain that ran before the pass settled consumed it and released nothing, and pass completion never re-armed it. The conformance harness's `drive_due_convergence` (edge-driven) spun on `runnable_work` forever — a livelock. Level-driven hosts (marmot-app's `schedule_after_pass`) recover, which is why this stayed hidden.
2. **Duplicate regeneration of in-flight intents.** Once the edge was restored, the drain re-prepared an intent the host had not yet confirmed, publishing the same logical message twice; the second confirm then failed with `Storage(NotFound)`.

## Fix

- `distributed_convergence.rs`: `rearm_queued_outbound_drain_after_pass_close` schedules the drain when a pass closes (settled **and** blocked completion paths) while queued intents remain durable.
- `message_processor/mod.rs`: the drain skips in-flight intents; `regenerated_queued_intent_for_message` is now a non-destructive lookup (was `take_...`); confirm/retry clear the in-flight association. `retry_queued_outbound_intent(group_id, intent_id)` gained the intent id so retry clears precisely one association (trait + session + simulator updated).
- Engine regression test `a_completed_pass_rearms_the_drain_for_intents_queued_inside_the_window`: pre-cutoff drain consumes the edge, completion re-arms, first drain publishes, second drain before confirm stays empty.

## Simulator gate re-enabled

- `bounded_convergence_pressure_family_settles_every_seeded_permutation` un-ignored; runs nightly (PR smoke excludes it with the other multi-minute generated batches, filter contract updated).
- Family generator bumped to **v2**: labelled accepted acknowledgements now carry the confirmed pending-resolution expectations the strict oracle requires (previously masked by the hard failure).

## Verification

- `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks` — green (25 suites).
- `cargo test -p cgka-engine --test convergence_policy_pin` (default pin gate) — green.
- `cargo test -p cgka-conformance-simulator` — green (all suites incl. the un-ignored gate).
- `cargo test -p cgka-session`, `marmot-account`, `marmot-app --features test-policy-overrides` — green.
- Isolated campaign: 3 seeds x 6 cases = 18 workers, all exit 0, no timeouts/integrity errors.
- `just fast-ci` (fmt, check, clippy) — green.
- Report CLI strict oracle passes on the previously failing saved input.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed queued messages that could remain undelivered after convergence completed.
  * Prevented duplicate publication while an outbound message is still in flight.
  * Improved targeted retry handling for regenerated queued messages.
  * Ensured queued messages are rescheduled after convergence settles, including when temporary queue access issues occur.

* **Tests**
  * Added regression coverage for queued messages during unsettled convergence.
  * Re-enabled the bounded convergence-pressure scenario and updated its expectations.

### DIFF
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -613,6 +613,30 @@ impl<S: StorageProvider> Engine<S> {
         }
     }
 
+    /// Re-arm the queued-outbound drain edge when a pass closes.
+    ///
+    /// The schedule edge is a one-shot signal: a drain that ran while the
+    /// pass was still open consumed it without releasing the durable queue.
+    /// Closing the pass is the transition that opens the outbound gate, so
+    /// the engine owes edge-driven hosts a fresh edge whenever intents are
+    /// still queued (mdk#1472). Level-driven hosts re-arm from
+    /// `has_queued_outbound_intents` on their own; the extra edge only costs
+    /// them one no-op drain.
+    fn rearm_queued_outbound_drain_after_pass_close(
+        &mut self,
+        group_id: &GroupId,
+    ) -> Result<(), OpenMlsProjectionError> {
+        if !self
+            .storage
+            .list_queued_outbound_intents(group_id)
+            .map_err(storage_projection_error)?
+            .is_empty()
+        {
+            self.schedule_pending_convergence_group(group_id);
+        }
+        Ok(())
+    }
+
     /// Discard a pass whose base epoch disagrees with the current tip, so the
     /// caller reopens one at the tip.
     ///
@@ -1382,6 +1406,7 @@ impl<S: StorageProvider> Engine<S> {
             self.storage
                 .put_convergence_pass(&pass)
                 .map_err(storage_projection_error)?;
+            self.rearm_queued_outbound_drain_after_pass_close(group_id)?;
             return Ok(result);
         }
 
@@ -1423,6 +1448,12 @@ impl<S: StorageProvider> Engine<S> {
             }
             Ok::<_, OpenMlsProjectionError>((observations, application_events))
         })?;
+        // mdk#1472: completing the pass opens the gate that held queued
+        // outbound intents. A drain that ran before the settle consumed the
+        // one-shot schedule edge without releasing them, so completion must
+        // re-arm it — otherwise an edge-driven host never comes back for the
+        // durable queue.
+        self.rearm_queued_outbound_drain_after_pass_close(group_id)?;
         for (disposition, previous_state, epoch) in disposition_transitions {
             if previous_state == disposition.state {
                 continue;
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -613,30 +613,6 @@ impl<S: StorageProvider> Engine<S> {
         }
     }
 
-    /// Re-arm the queued-outbound drain edge when a pass closes.
-    ///
-    /// The schedule edge is a one-shot signal: a drain that ran while the
-    /// pass was still open consumed it without releasing the durable queue.
-    /// Closing the pass is the transition that opens the outbound gate, so
-    /// the engine owes edge-driven hosts a fresh edge whenever intents are
-    /// still queued (mdk#1472). Level-driven hosts re-arm from
-    /// `has_queued_outbound_intents` on their own; the extra edge only costs
-    /// them one no-op drain.
-    fn rearm_queued_outbound_drain_after_pass_close(
-        &mut self,
-        group_id: &GroupId,
-    ) -> Result<(), OpenMlsProjectionError> {
-        if !self
-            .storage
-            .list_queued_outbound_intents(group_id)
-            .map_err(storage_projection_error)?
-            .is_empty()
-        {
-            self.schedule_pending_convergence_group(group_id);
-        }
-        Ok(())
-    }
-
     /// Discard a pass whose base epoch disagrees with the current tip, so the
     /// caller reopens one at the tip.
     ///
@@ -1406,7 +1382,7 @@ impl<S: StorageProvider> Engine<S> {
             self.storage
                 .put_convergence_pass(&pass)
                 .map_err(storage_projection_error)?;
-            self.rearm_queued_outbound_drain_after_pass_close(group_id)?;
+            self.schedule_drain_for_retained_outbound_intents(group_id);
             return Ok(result);
         }
 
@@ -1453,7 +1429,7 @@ impl<S: StorageProvider> Engine<S> {
         // one-shot schedule edge without releasing them, so completion must
         // re-arm it — otherwise an edge-driven host never comes back for the
         // durable queue.
-        self.rearm_queued_outbound_drain_after_pass_close(group_id)?;
+        self.schedule_drain_for_retained_outbound_intents(group_id);
         for (disposition, previous_state, epoch) in disposition_transitions {
             if previous_state == disposition.state {
                 continue;
--- a/crates/cgka-engine/tests/publish_lifecycle.rs
+++ b/crates/cgka-engine/tests/publish_lifecycle.rs
@@ -2010,6 +2010,86 @@ async fn an_unreadable_intent_queue_still_schedules_the_drain() {
 }
 
 #[tokio::test]
+async fn an_unreadable_intent_queue_at_pass_close_still_rearms_and_drains() {
+    let inner = SqliteAccountStorage::in_memory().unwrap();
+    let queued_intent_list_fault = ProcessedFault::default();
+    let clock = ManualConvergenceClock::new(1_000, 10_000);
+    let mut alice = EngineBuilder::new(FaultStorage {
+        inner,
+        fault: ProcessedFault::default(),
+        lifecycle_fault: ProcessedFault::default(),
+        disband_request_fault: ProcessedFault::default(),
+        queued_intent_list_fault: queued_intent_list_fault.clone(),
+    })
+    .legacy_compatibility_profile()
+    .identity(pad32(b"alice-pass-close"))
+    .account_identity_proof_signer(proof_signer(b"alice-pass-close"))
+    .feature_registry(registry_with_reactions())
+    .peeler(Box::new(MockPeeler))
+    .convergence_clock(Arc::new(clock.clone()))
+    .build()
+    .unwrap();
+    let mut bob = build(b"bob-pass-close");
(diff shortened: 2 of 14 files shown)


Datawav (commented):
## Automated review

Verification complete. I traced all 14 patches in the bundle, the four head-context snapshots, and both reviews plus their cross-verifications against the actual diff.

Reconciliation summary (internal): review_mode is `initial`; `pr_discussion.enabled=false` with zero items, so there are no discussion leads to validate; coverage is complete (14/14 patches, no missing files); 26 CI runs all completed, 25 success + 1 legitimate release skip. Every lead raised across the four review documents (intra-drain duplicate regeneration, unpopulated in-flight map, drain starvation, retry over-clearing, cross-group leak, stale-association leak, stale-pass discard path, unknown-intent retry) was re-derived against the exact-head code and each one is either refuted by the head (drain inserts associations at message_processor/mod.rs:743-748 behind the in-flight guard at :697; drain is bounded by MAX_QUEUED_OUTBOUND_INTENTS_PER_GROUP=256; retry retain-clears by intent_id across both maps and unconditionally re-schedules at :940-945) or is pre-existing-and-correct code outside the patch. The un-ignored simulator gate, Justfile smoke-filter and expected-diff updates, and generator_version "1"→"2" bump with the matching assertion are all internally consistent in the same pass. No supported, actionable finding survives from any reviewer, and I found no additional defect in the patch myself.

Final audit output:

No actionable findings.

crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -613,6 +613,30 @@ impl<S: StorageProvider> Engine<S> {
...
+    ) -> Result<(), OpenMlsProjectionError> {
+        if !self
+            .storage
+            .list_queued_outbound_intents(group_id)
+            .map_err(storage_projection_error)?
+            .is_empty()
+        {
+            self.schedule_pending_convergence_group(group_id);
+        }
+        Ok(())
```
erskingardner: This can still lose the drain edge this PR is fixing.

Both call sites write the pass as `Completed` *before* this runs. A failed `list_queued_outbound_intents` then returns `Err` via `?`, and the next `converge_stored_openmls_messages_with_time` hits the `Completed` early-return and never comes back here. An edge-driven host livelocks again until restart or unrelated traffic — the post-#1403 shape.

`schedule_drain_for_retained_outbound_intents` already exists for this seam: it is infallible, and a failed read (which cannot prove the queue is empty) schedules anyway. The comment you added on that helper even names this pass-close path. Delete this function and call it from both close sites:

```rust
self.schedule_drain_for_retained_outbound_intents(group_id);
```

That is fewer lines and the error policy this crate already chose for a committed outcome.

erskingardner (commented):
Request changes (posted as comment — GitHub rejects `REQUEST_CHANGES` on the author's own PR).

The two interlocking fixes are the right ones, and `a_completed_pass_rearms_the_drain_for_intents_queued_inside_the_window` reconstructs the edge-driven order that hid this. One issue should land first: `rearm_queued_outbound_drain_after_pass_close` can still lose the drain edge after the pass is already `Completed` — the same defect this PR exists to close. Call the existing infallible helper instead of inventing a fallible one.

erskingardner (commented):
Request changes (posted as a comment because GitHub rejects REQUEST_CHANGES on the author account).

Re-reviewed exact head `8935a0f5d9beed9d6b69e35d16adcc2080625b7d`. The two intended lifecycle fixes are the right ones, and all 25 substantive checks pass, but the existing pass-close blocker remains: `rearm_queued_outbound_drain_after_pass_close` performs a fallible queue read after `Completed` (and, on the settled path, the canonical apply) is durable. A transient read error returns before re-arming and before the settled path finishes its in-memory epoch/event/route-index transitions; the next attempt hits the `Completed` early return. Replace this helper at both close sites with the existing infallible `schedule_drain_for_retained_outbound_intents`, whose conservative read-error policy was designed for this seam, and add fault-injection coverage.

Existing inline finding: https://github.com/marmot-protocol/mdk/pull/1475#discussion_r3794170339

The confirmation-cleanup assertion suggested in the other thread would strengthen coverage, but it is non-blocking.

crates/cgka-engine/src/distributed_convergence.rs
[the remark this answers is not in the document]
erskingardner: Fixed in 830ca5c5. Both pass-close sites now use the existing infallible `schedule_drain_for_retained_outbound_intents`, which schedules conservatively when the durable queue cannot be read. Added `an_unreadable_intent_queue_at_pass_close_still_rearms_and_drains`: it injects the post-close read failure, verifies the durable convergence apply still succeeds and re-arms the edge, then verifies the successful retry releases the retained intent.

crates/cgka-engine/tests/publish_lifecycle.rs
```diff
@@ -2115,7 +2115,7 @@ async fn queued_outbound_intents_are_capped_per_group_while_a_publish_stays_unre
         // reports that the regenerated message reached the transport. Until
         // then the row is still owed a delivery, so it still occupies the cap.
         let (_, intent_id) = alice
-            .take_regenerated_queued_intent_for_message(&msg.id)
+            .regenerated_queued_intent_for_message(&msg.id)
             .expect("a drained retained intent must be attributable to its durable row");
         alice.confirm_queued_outbound_intent(&intent_id).unwrap();
```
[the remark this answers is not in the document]
erskingardner: Added the requested post-confirmation `None` assertion in 830ca5c5. The exact capacity/cleanup test and the full `cgka-engine` suite pass.

crates/cgka-engine/src/distributed_convergence.rs
erskingardner: Addressed in 830ca5c5: deleted the duplicate fallible helper, reused `schedule_drain_for_retained_outbound_intents` at both close sites, and added the fault-injection regression for a failed post-close queue read followed by a successful drain retry.

erskingardner (commented):
Review at head 830ca5c5 (including the follow-up commit for both review threads). Posting as a comment because GitHub won't let me approve a PR I opened.

The follow-up is a strict improvement: deleting the fallible `rearm_queued_outbound_drain_after_pass_close` and reusing the existing infallible `schedule_drain_for_retained_outbound_intents` is fewer lines, matches the crate's chosen error policy for committed outcomes, and closes the "read error after durable pass close loses the edge" hole — the same livelock class this PR exists to fix. The new fault-injection regression (`an_unreadable_intent_queue_at_pass_close_still_rearms_and_drains`) covers exactly that seam, and the post-confirm `None` assertion pins the in-flight association cleanup.

Also verified at head: both pass-close call sites use the shared helper, the two doc blocks updated for the new in-flight semantics are accurate, no dangling references to the deleted helper, and CI is green across the board (all Rust suites, simulator/vectors, clippy, fmt, policy pin).

**Recommend merge.**

erskingardner (commented):
Recommend merge (posted as comment — GitHub rejects `APPROVE` on the author's own PR).

Re-reviewed exact head `830ca5c55d349b810b5ea0cef347775349ff1921`. The previous blocker is gone: both pass-close sites call infallible `schedule_drain_for_retained_outbound_intents`, the duplicate helper is deleted, and `an_unreadable_intent_queue_at_pass_close_still_rearms_and_drains` actually hits the post-close list (the fault is armed after queueing, and settle does not list the queue before that helper). Confirmation now also asserts the in-flight association is cleared. CI is green. No remaining findings.

crates/cgka-engine/src/distributed_convergence.rs
erskingardner: Confirmed at `830ca5c`: helper deleted, both close sites use `schedule_drain_for_retained_outbound_intents`, fault-injection regression covers the failed post-close read. This thread is resolved.

PR #1473, opened 2026-08-16
ci: publish Android bindings for SHA snapshots

mubarakcoded:
## Summary
- build the Android Kotlin/JNI bundle for workflow-dispatched SHA snapshots
- compile the exact requested source with current-master workflow tooling and record both source and builder SHAs
- publish the Android archive and checksum alongside Apple snapshot assets
- document immutable Android snapshot URLs, contents, provenance, and consumer staging

## Verification
- `actionlint .github/workflows/bindings.yaml`
- `bash -n crates/marmot-uniffi/kotlin-bindings.sh`
- `just fmt-check`
- semantic checks for Android job gating, release assets, provenance, URLs, and required Kotlin helpers
- `just fast-ci`: formatting, repository policy gates, workspace checks, and both Clippy passes completed; the final convergence test was interrupted only by local ENOSPC
- `just test-convergence-policy-pin`: 5 passed after clearing generated build artifacts

No snapshot release was dispatched from this branch.

## Summary by CodeRabbit

* **New Features**
  * Android bindings are now included in both regular and snapshot releases alongside Apple platform artifacts.
  * Android release packages include generated Kotlin sources and native libraries for supported ABIs.
  * Release manifests now provide source and build provenance details.

* **Documentation**
  * Added guidance for downloading, matching, staging, and checksum-verifying platform artifacts.
  * Updated distribution documentation with Android archive contents, URLs, and compatibility requirements.

PR #1480, opened 2026-08-18
fix(marmotkit): harden Android snapshot artifacts

mubarakcoded:
## Summary

- strip Android target JNI libraries while keeping the host dylib metadata intact for UniFFI generation
- fail the binding build if an Android library still contains debug or static symbol sections
- restore the existing Android manifest contents contract while continuing to package and validate both hand-written initialization helpers
- document the strip boundary and manifest contract

## Context

The first Android SHA snapshot produced after #1473 exposed two release blockers:

1. the arm64 JNI library was about 57 MB and reported `not stripped`
2. White Noise Android rejected the archive because the manifest declared helper files outside its existing ordered API contract

The published snapshot is immutable, so a corrected snapshot must be created from a new MDK commit after this PR merges.

## Verification

- `actionlint .github/workflows/bindings.yaml`
- `bash -n crates/marmot-uniffi/kotlin-bindings.sh`
- real arm64 OTLP binding build; output reports `stripped`, preserves required dynamic JNI/UniFFI exports, and falls from about 57 MB to 35 MB
- the same binding build with `CARGO_NET_OFFLINE=true`
- offline White Noise Android artifact staging from a locally packaged all-ABI bundle
- offline `:app:compileDevZapstoreDebugKotlin` in a disposable Android master worktree after supplying the new `AppGroupMemberIdsFfi.adminIdsHex` consumer field
- `just fast-ci`
- `git diff --check`

### DIFF
--- a/.github/workflows/bindings.yaml
+++ b/.github/workflows/bindings.yaml
@@ -381,6 +381,13 @@ jobs:
             "jniLibs/x86/libmarmot_uniffi.so"
             "jniLibs/x86_64/libmarmot_uniffi.so"
           )
+          manifest_contents=(
+            "kotlin/dev/ipf/marmotkit/marmot_uniffi.kt"
+            "jniLibs/arm64-v8a/libmarmot_uniffi.so"
+            "jniLibs/armeabi-v7a/libmarmot_uniffi.so"
+            "jniLibs/x86/libmarmot_uniffi.so"
+            "jniLibs/x86_64/libmarmot_uniffi.so"
+          )
 
           rm -rf "$stage_parent" "$asset" "$asset.sha256"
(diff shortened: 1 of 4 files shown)


erskingardner (approved):
Recommend merge. I found no actionable issues at exact head `b269e2255896a504befc625662152305cf4a67a2`. Android stripping is scoped to target JNI builds while the host dylib retains UniFFI metadata; the post-build section check fails closed; and the restored ordered manifest matches White Noise Android’s pinned five-entry contract while both hand-written helpers remain required archive payloads. Exact-head PR CI is green. Validation gap: the bindings workflow is tag/workflow-dispatch only, so the author-reported real Android build and consumer compile—not PR CI—exercise this release path end to end.

PR #1488, opened 2026-08-18
Add bounded sync and catch-up failure telemetry

erskingardner:
## What changed

- add shared, bounded `failure_stage` and `error_class` classifications for account sync and catch-up failures
- classify errors at explicit transport activation, subscription refresh, relay receive, CGKA ingest, persistence, and account-worker boundaries
- preserve catch-up classifications when failures propagate to the parent account-sync metric
- export only closed-set metric attributes and add regression coverage preventing raw error text or high-cardinality identifiers
- document the allowed values, privacy/cardinality contract, and a Grafana/PromQL aggregation example

## Why

The existing counters identify failed account sync and catch-up attempts but cannot distinguish relay transport failures from subscription, CGKA, persistence, or timeout failures. This keeps the existing metric names and terminal-counting behavior while making the failing layer operationally visible on both iOS and Android.

## Validation

Passed:

- `just fast-ci`
- focused classification, propagation, privacy, exporter, and OTLP tests
- `cargo test -p transport-nostr-adapter` (28 unit and 36 integration tests)
- CLI sync failure tests
- all 686 `marmot-app` library unit tests
- `git diff --check`

Broader test caveat:

- the `marmot-app` `relay_runtime` binary suite passed 98 tests and failed 5 existing timing/state-sensitive scenarios (encrypted-media endpoint replacement, encrypted-media secret rehydration, retention system-row delivery, retained-media rehydration, and self-removal unread suppression). A serial rerun reproduced the first two; the failures are outside the telemetry paths changed here.

## Summary by CodeRabbit

* **Improvements**
  * Improved synchronization and catch-up error reporting with clearer failure stages and categories.
  * Preserved useful failure details across retries, repairs, and startup recovery.
  * Improved handling and reporting when transport delivery channels close unexpectedly.
  * Added privacy-safe, bounded failure classifications to performance and relay telemetry.

* **Documentation**
  * Updated telemetry documentation with classification values, fallback behavior, privacy guidance, and query examples.

* **Testing**
  * Expanded coverage for failure classification, propagation, telemetry counts, exports, and transport failures.

### DIFF
--- a/crates/marmot-app/src/app_telemetry.rs
+++ b/crates/marmot-app/src/app_telemetry.rs
@@ -385,7 +385,13 @@ impl DurationHistogram {
 
 impl AppPerformanceOperationTelemetry {
     fn record(&mut self, duration: Duration, success: bool) {
-        self.record_with_failure(duration, success, None);
+        self.attempts += 1;
+        if success {
+            self.successes += 1;
+        } else {
+            self.failures += 1;
+        }
+        self.duration_ms.record(duration);
     }
 
     fn record_with_failure(
@@ -394,18 +400,14 @@ impl AppPerformanceOperationTelemetry {
         success: bool,
         failure: Option<SyncFailureClassification>,
     ) {
-        self.attempts += 1;
-        if success {
-            self.successes += 1;
-        } else {
-            self.failures += 1;
+        self.record(duration, success);
+        if !success {
             let classification = failure.unwrap_or(SyncFailureClassification::UNKNOWN);
             *self
                 .failure_classifications
                 .entry(classification)
                 .or_default() += 1;
         }
-        self.duration_ms.record(duration);
     }
 
     fn snapshot(&self) -> AppPerformanceOperationSnapshot {
@@ -740,13 +742,13 @@ pub(crate) async fn bounded_advisory_step<F: std::future::Future>(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{AccountCatchUpFailure, AppError, SyncFailure, SyncSummary};
+    use crate::{AccountCatchUpFailure, AppError, ClassifiedSyncFailure, SyncSummary};
     use cgka_traits::TransportAdapterError;
     use cgka_traits::error::EngineError;
     use cgka_traits::storage::StorageError;
 
     fn classified(stage: SyncFailureStage, error: AppError) -> SyncFailureClassification {
-        SyncFailure::at_stage(SyncSummary::default(), error, stage).classification()
+        ClassifiedSyncFailure::at_stage(SyncSummary::default(), error, stage).classification()
     }
 
     #[test]
@@ -1014,6 +1016,13 @@ mod tests {
         assert_eq!(snapshot.group_accept_invite.attempts, 2);
         assert_eq!(snapshot.group_accept_invite.successes, 1);
         assert_eq!(snapshot.group_accept_invite.failures, 1);
+        assert!(
+            snapshot
+                .group_accept_invite
+                .failure_classifications
+                .is_empty(),
+            "generic operation failures must not populate sync classifications"
+        );
         assert_eq!(snapshot.group_accept_invite.duration_ms.sample_count(), 2);
         assert_eq!(snapshot.group_accept_invite.duration_ms.sum_ms, 100);
     }
--- a/crates/marmot-app/src/app_telemetry.rs
+++ b/crates/marmot-app/src/app_telemetry.rs
@@ -5,6 +5,7 @@
 //! are no fields for account, group, message, relay, URL, pubkey, payload, or
 //! key material.
 
+use std::collections::BTreeMap;
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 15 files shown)


crates/marmot-app/src/client/sync.rs
```diff
@@ -748,12 +788,24 @@ impl AppClient {
     }
 
     async fn sync_sdk_relay(&mut self, deliveries: &mut u64) -> Result<SyncSummary, SyncFailure> {
-        let display_names = self.app.display_names_by_id().map_err(SyncFailure::from)?;
+        let display_names = self.app.display_names_by_id().map_err(|error| {
+            SyncFailure::at_stage(
+                SyncSummary::default(),
+                error,
+                SyncFailureStage::AccountWorker,
+            )
```
erskingardner: `display_names_by_id()` and the `account_home().account()` lookup just below are local app-state reads, not the account-worker command/response boundary. The same two calls in `drain_pending_session_events` are then labeled `cgka_ingest` (unless Storage) by the error-class heuristic.

Please use one honest stage for both sites. `unknown` is better than polluting `account_worker`, which operators will use to find worker timeouts.

erskingardner (commented):
Would request changes if GitHub allowed it on an author-owned PR.

The overall shape is right: closed `failure_stage` / `error_class` enums, no `Display` parsing, classification carried through the worker oneshot, and OTLP attributes structurally bounded. That is the correct privacy/cardinality approach, and the RelayDirectory-mislabel history in `AppError` is exactly why this is worth doing.

The problem is that several sites do **not** record the explicit sync boundary the docs promise. The same local directory / account-home lookups are labeled `account_worker` in `sync_sdk_relay` and `cgka_ingest` after `drain_pending_session_events`, and drain/backfill infer stage from `error_class`. That will send operators to the wrong layer — the failure mode this PR exists to stop. Please classify at the failing call site (or use `unknown`) before merge.

Non-blocking:
- `TransportAdapterError::Timeout` is never constructed; catch-up `receive()` timeouts already break as success, and worker waits already use `AppError` variants.
- `record()` still routes `AccountSync` / `AccountCatchUp` through unclassified `record_with_failure(..., None)`, so a leftover `.record(..., false)` silently becomes `unknown`/`unknown`.
- MarmotKit `appPerformanceSnapshot()` still drops the breakdown. Fine if OTLP is the only iOS/Android consumer; not if host debug UI was in scope.
- `Backend` / `Subscription` / `Publish` stay `error_class=unknown` by design. Stage is the useful dimension there — worth saying so in the telemetry doc so dashboards do not expect a populated class on relay failures.

crates/marmot-app/src/client/sync.rs
```diff
@@ -370,7 +405,12 @@ impl AppClient {
         let drained = match self.drain_pending_session_events().await {
             Ok(drained) => drained,
             Err(error) => {
-                return Err(SyncFailure::new(summary, error));
+                let stage = if error.sync_error_class() == SyncErrorClass::Storage {
```
erskingardner: This infers `failure_stage` from `error_class`, which contradicts the contract that stage is the explicit boundary that stopped. `drain_pending_session_events` also does directory/account-home reads, publish-failure checks, and projection — a non-storage failure here is not ingest.

Classify at the inner failing call (or use `unknown`). The same heuristic is copied in `repair_full_history` and `run_pending_epoch_backfill_reporting_arm`.

crates/traits/src/transport_adapter.rs
```diff
@@ -697,6 +697,14 @@ pub enum TransportAdapterError {
     #[error("account not active")]
     AccountNotActive(MemberId),
 
+    /// A bounded transport operation exceeded its explicit deadline.
+    #[error("transport operation timed out")]
+    Timeout,
```
erskingardner: `Timeout` is never constructed anywhere in this PR. Catch-up already treats `tokio::timeout` around `adapter.receive()` as a successful drain (`Err(_) => break`), and worker/catch-up waits already classify via `AppError::AccountWorkerResponseTimedOut` / `AccountCatchUpFailure`.

Do not add a public `traits` variant until an adapter actually produces it. `Closed` is fine — those two delivery-queue sites are real.

crates/marmot-app/src/app_telemetry.rs
```diff
@@ -468,6 +584,30 @@ impl AppPerformanceTelemetry {
...
+        let mut inner = self
+            .inner
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        let target = match operation {
+            AppPerformanceOperation::AccountSync => &mut inner.account_sync,
+            AppPerformanceOperation::AccountCatchUp => &mut inner.account_catch_up,
+            _ => return,
+        };
+        target.record_with_failure(duration, failure.is_none(), failure);
```
erskingardner, line 617: `record()` still has `AccountSync` / `AccountCatchUp` arms that call `record_with_failure(..., None)`. Any leftover or future `.record(AccountSync, ..., false)` silently increments `unknown`/`unknown` and looks like a classified sample.

Drop those arms from `record()` (force `record_sync_result`) or `debug_assert` that classified operations do not go through the unclassified path.

erskingardner (commented):
Would approve if GitHub allowed it on an author-owned PR.

Follow-up addresses the previous review.

- Local directory / account-home reads and composite drain/backfill now use `unknown` instead of inventing a stage from `error_class`.
- Unused `TransportAdapterError::Timeout` is gone.
- `record()` refuses `AccountSync` / `AccountCatchUp` and the doc now states that `Backend` / `Subscription` / `Publish` stay `error_class=unknown`.

One leftover, non-blocking: `record()` does `debug_assert!(false); return;`, so a mistaken `.record(AccountSync, …)` panics in debug and silently drops the sample in release. Fine given current callers; if this path is ever hit in production it undercounts rather than landing in `unknown`/`unknown`.

Recommend merge.

PR #1493, opened 2026-08-18
Prioritize setup KeyPackage publication

### LINKED ISSUE #1484: marmot-app: prioritize the setup KeyPackage before initial sync and parallelize bootstrap acknowledgements
## Problem

Even with the current setup batching, the fully-network-ready account path contains avoidable serialization.

At 82af46ca1352:

- [the account worker signals local readiness](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/runtime/account_worker.rs#L510), then [starts initial sync](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/runtime/account_worker.rs#L600);
- [ordinary mutations are deferred while that sync is active](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/runtime/account_worker.rs#L699);
- initial setup publishes its KeyPackage through [the worker command path](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/runtime/commands.rs#L1384), so the highest-value setup publication can wait behind initial sync;
- bootstrap publication builds a small event batch, but [the SDK adapter publishes batch requests sequentially](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/transport-nostr-adapter/src/sdk_client.rs#L710) after connecting to relays concurrently.

This means a user may wait for unrelated catch-up work before becoming invite-receivable, and bootstrap latency can still accumulate per request/event.

## Requested outcome

Create a narrowly scoped startup-priority path for the exact initial KeyPackage and make independent bootstrap publish acknowledgements concurrent under a single bounded deadline.

This should improve the fully-network-ready path regardless of whether account creation later adopts a local-first return boundary.

## Correctness constraints

- Do not add a generic mutation bypass around initial sync. Only setup work whose ordering and durability invariants are proven may use the priority lane.
- Preserve the setup journal and exact-KeyPackage retry guarantees.
- Preserve deterministic per-event outcomes and retry accounting when sends complete out of order.
- Bound both concurrency and total time; one dead relay or request must not multiply the batch deadline.
- Respect any required Nostr event dependency/order. Only independent publications may overlap.
- Keep aggregate telemetry privacy-safe and never log relay URLs, pubkeys, event payloads, or key material.

## Acceptance criteria

- The initial KeyPackage can be published before unrelated initial-sync catch-up completes.
- A regression test demonstrates that an intentionally stalled initial sync does not stall initial KeyPackage publication.
- Independent bootstrap requests publish concurrently with a shared deadline and bounded concurrency.
- Partial relay failure preserves the existing success policy and produces deterministic retry state.
- Add phase timings for worker readiness, bootstrap publish, initial KeyPackage publish, and initial sync overlap.
- Provide a repeatable before/after benchmark or fault-injection test covering healthy relays, one stalled relay, and all relays unavailable.

## Related work

- #1436 / PR #1446 reduced setup round trips and introduced shared relay connection work.
- The local-first readiness boundary is intentionally a separate issue; this issue optimizes time to actual network readiness.

## Structured publication timing

The existing timing criterion must expose default-profile publication separately from the rest of bootstrap publication. The reviewed app-performance surface should distinguish bootstrap relay/follow-list work, default-profile publication, initial KeyPackage publication, and initial-sync overlap so an end-to-end host measurement can identify which network phase dominates.

Keep these fields aggregate and bounded through the existing `AppPerformanceSnapshotFfi`; do not introduce relay-, account-, event-, or caller-defined labels.

erskingardner:
## Summary
- prioritize the exact journaled setup KeyPackage before initial sync while keeping general mutations on the startup FIFO
- publish independent bootstrap events concurrently with bounded fan-out, one shared deadline, and input-ordered outcomes
- expose privacy-safe timings for worker readiness, bootstrap relay/follow publication, default profile publication, initial KeyPackage publication, and initial-sync overlap

## Test plan
- [x] New priority and batch-concurrency tests failed before the fix and pass after it
- [x] `cargo test -p transport-nostr-adapter --features sdk`
- [x] `cargo test -p marmot-app` unit tests: 689 passed
- [x] `just fast-ci`

`marmot-app` integration `relay_runtime` still has five unrelated baseline failures; `encrypted_media_endpoint_updates_are_full_replacement_and_admin_only` was reproduced unchanged from pristine `origin/master`.

Fixes #1484

### DIFF
--- a/crates/marmot-app/src/app_telemetry.rs
+++ b/crates/marmot-app/src/app_telemetry.rs
@@ -23,6 +23,7 @@ pub(crate) enum AppPerformanceOperation {
     DirectorySubscriptionSync,
     AccountReconcile,
     AccountOpen,
+    AccountWorkerReadiness,
     AccountSessionOpen,
     AccountGroupHydration,
     AccountProfileLoad,
@@ -32,6 +33,10 @@ pub(crate) enum AppPerformanceOperation {
     AccountCatchUp,
     AccountSync,
     AccountSetupAdvisoryStep,
+    AccountBootstrapRelayAndFollowPublish,
+    AccountDefaultProfilePublish,
+    AccountInitialKeyPackagePublish,
+    AccountInitialSyncOverlap,
     OutboundMessageSend,
     GroupCreateQueueWait,
     GroupCreateKeyPackageLookup,
@@ -192,6 +197,8 @@ pub struct AppPerformanceSnapshot {
     pub account_reconcile: AppPerformanceOperationSnapshot,
     pub account_open: AppPerformanceOperationSnapshot,
     #[serde(default)]
+    pub account_worker_readiness: AppPerformanceOperationSnapshot,
+    #[serde(default)]
     pub account_session_open: AppPerformanceOperationSnapshot,
     #[serde(default)]
     pub account_group_hydration: AppPerformanceOperationSnapshot,
@@ -206,6 +213,14 @@ pub struct AppPerformanceSnapshot {
     pub account_catch_up: AppPerformanceOperationSnapshot,
     pub account_sync: AppPerformanceOperationSnapshot,
     pub account_setup_advisory_step: AppPerformanceOperationSnapshot,
+    #[serde(default)]
+    pub account_bootstrap_relay_and_follow_publish: AppPerformanceOperationSnapshot,
+    #[serde(default)]
+    pub account_default_profile_publish: AppPerformanceOperationSnapshot,
+    #[serde(default)]
+    pub account_initial_key_package_publish: AppPerformanceOperationSnapshot,
+    #[serde(default)]
+    pub account_initial_sync_overlap: AppPerformanceOperationSnapshot,
     /// Interrupted-migration recovery probes executed since process start. Each
     /// probe is a full keyed SQLCipher open paying the passphrase KDF; the
     /// healthy steady state skips it via the cached v2-open verdict (mdk#1439).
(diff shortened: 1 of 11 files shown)


erskingardner (commented):
Recommend merge. I would approve if GitHub allowed a self-review on this PR.

The startup-priority path is the right shape for #1484: the journaled `KeyPackagePublicationStarted` marker authorizes one exact publish before hydration/initial sync, and general mutations (including public `PublishKeyPackage`) stay on the FIFO. That is the only design that works while `initial_sync` holds `&mut client`. Skipping `activate_transport` on the setup lane is the correct cut.

The SDK batch change also matches the constraints: independent requests fan out with a cap of 4, one shared deadline, and outcomes written back in input order. Current `publish_events` callers are independent replaceable/deletion batches, so overlapping them is safe. The new tests cover stalled-sync KP publication, healthy-vs-stalled concurrency, ordered partial failure, and the shared all-unavailable window.

One non-blocking note: `AccountInitialSyncOverlap` is recorded as `Duration::ZERO` on the priority path, so it is a sentinel that KP finished first rather than a measured overlap hosts can use to see which network phase dominates. `AccountSync` already has the sync duration; the bootstrap/profile/KP buckets are the useful split. Fine to land as-is if the zero is intentional documentation of the serialize-then-sync choice.

PR #1494, opened 2026-08-18
marmot-app: batch and prewarm member KeyPackages

### LINKED ISSUE #1486: marmot-app: batch and prewarm member KeyPackages before group creation
## Problem

Group creation resolves invitee KeyPackages with bounded concurrency, but it still performs one logical lookup per member.

At 82af46ca1352:

- [group creation maps members to individual KeyPackage lookups with concurrency 8](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/client/mod.rs#L871);
- [each lookup checks local/cache state and can fall through to a network fetch](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/lib.rs#L4167);
- a network miss may [fetch relay-list data and then KeyPackages sequentially](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/directory/methods.rs#L270);
- the relay plane already supports [multi-author directory queries](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/relay_plane/directory.rs#L130), but the create path does not exploit that across the full member set.

For larger or cold-cache groups, request count and tail latency grow with the number and relay distribution of invitees. The Create action also starts this work late, even though the member list is known during composition.

## Requested outcome

Add a set-oriented KeyPackage resolution API that canonicalizes/deduplicates invitees, batches compatible multi-author queries by relay/query limits, and returns deterministic per-member results. Allow the host/runtime to prewarm this resolution while the user is composing the group, then consume fresh validated results during creation.

## Correctness constraints

- Preserve KeyPackage signature, lifetime, profile/revision, and single-use/freshness checks.
- Do not consume or reserve a package merely because it was prefetched.
- Revalidate cached/prefetched results at commit time where required.
- Preserve stranger/cache-admission policy and bounded memory.
- Bound authors per query, response size, relay concurrency, and total deadline.
- Preserve input-order deterministic error reporting even if batch responses arrive out of order.
- Deduplicate repeated account IDs without changing the final roster semantics.
- Keep logs/metrics aggregate and privacy-safe.

## Acceptance criteria

- Cold-cache group creation no longer issues one independent directory request sequence per member when authors can be queried together.
- A benchmark/test reports request count and wall-clock time for 1, 8, and 32 invitees across warm cache, shared relays, split relays, one stalled relay, and missing packages.
- Prefetch started during composition is reused only while valid and materially reduces the Create-button wait.
- Cancellation and member-list edits discard or safely reuse work without leaking reservations.
- Partial/malformed batch responses map to deterministic per-member outcomes.
- Existing bounded-concurrency behavior remains as a fallback for relays/query shapes that cannot be batched.

## Related work

This issue owns directory lookup/request amplification before the canonical mutation. It does not own MLS group construction cost, Welcome fanout, or post-canonical local persistence.

## Structured member-resolution timing

In addition to benchmark output, expose member/KeyPackage resolution through the reviewed app-performance snapshot. Preserve the existing total group-create KeyPackage lookup dimension and add only the bounded distinction needed to tell whether time was spent in prewarm/network resolution versus create-time freshness validation/cache reuse.

The metrics must remain aggregate and privacy-safe: no member count as an unbounded label, account identifiers, relay identifiers, or per-member timing records.

erskingardner:
## Summary

- canonicalize and deduplicate complete member sets, reuse validated local/cache state, and batch cold relay-list and KeyPackage queries by compatible relay set
- expose bounded, non-reserving composition prewarm through the app runtime and UniFFI, with mutation-time revalidation and automatic single-author fallback
- add privacy-safe app-performance dimensions plus deterministic partial/error, cancellation, cache-bound, fallback, and 1/8/32-member scaling coverage

## Verification

- `just fast-ci`
- `cargo test -p marmot-uniffi` (149 tests)
- focused member-resolution and prewarm regression tests
- `cargo test -p marmot-app member_key_package_resolution_scaling_report --lib -- --ignored --nocapture`
- `cargo check -p marmot-app --features otlp-export`

The scaling report showed 0 relay requests for warm-cache cases; shared-relay cases used 2 requests at 1, 8, and 32 invitees; split/stalled cases used at most 3; missing-package cases failed deterministically.

A full `cargo test -p marmot-app` run passed all 694 non-ignored library tests and the other integration suites, but `tests/relay_runtime.rs` reported five failures. The first (`encrypted_media_endpoint_updates_are_full_replacement_and_admin_only`) reproduced identically from untouched `origin/master` at `a5bbb459`, so it is a current-master baseline failure unrelated to this patch.

Fixes #1486

## Summary by CodeRabbit

- **New Features**
  - Added group-member KeyPackage prewarming with a summary of requested, unique, reused, and network-resolved packages.
  - Group creation and invitations now reuse cached packages, deduplicate member references, and resolve members through efficient relay lookups.
  - Added final package validation before group changes are applied.
  - New groups continue to retain their routing information.

- **Bug Fixes**
  - Improved handling of duplicate, malformed, missing, or unavailable member packages.

- **Documentation**
  - Updated architecture and telemetry documentation for the new group-creation and prewarming behavior.


crates/marmot-app/src/tests.rs
```diff
@@ -3723,6 +3771,425 @@ async fn member_key_package_falls_back_to_current_directory_for_local_account()
...
+        .await
+        .unwrap();
+    assert_eq!(summary.network_resolved_members, 1);
+    assert!(fetcher.requests.lock().unwrap().len() > requests_before_retry);
+}
+
+#[tokio::test]
+#[ignore = "member-resolution scaling benchmark; reports request count and wall clock"]
+async fn member_key_package_resolution_scaling_report() {
+    eprintln!("invitees,scenario,requests,wall_ms,outcome");
```
erskingardner: These `eprintln!` calls fail CI. `production_library_sources_do_not_write_direct_output` walks `crates/**/*.rs` and only skips `tests/` and `benches/` directories, so `src/tests.rs` is treated as production library source. That is why the Simulator and vectors job is red on this SHA.

Move this ignored report into `crates/marmot-app/tests/` (or a bench) so the audit ignores it, or drop the direct-output macros.

crates/marmot-app/src/directory/member_key_packages.rs
```diff
@@ -0,0 +1,730 @@
...
+                            (index, endpoints, records)
+                        }
+                    });
+                    let results = stream::iter(work)
+                        .buffered(MEMBER_RESOLUTION_FALLBACK_CONCURRENCY)
+                        .collect::<Vec<_>>()
+                        .await;
+                    for (index, endpoints, records) in results {
+                        targets[index].relay_lists =
+                            self.relay_lists_from_records(&targets[index], records?, &endpoints);
```
erskingardner: `records?` returns from `resolve_missing_relay_lists`, and the caller propagates that with `?` before any KeyPackage work. One single-author relay-list failure then:

- discards in-memory outcomes for members already satisfied from local/directory/prewarm state
- skips KeyPackage fetching for every other unresolved member, including those whose relay lists already applied
- returns that fallback `RelayDirectory` error instead of the first canonical-input member error

The KeyPackage stage already degrades per member and then scans outcomes in input order. Do the same here: record the failure on that index, keep going, and let the existing input-order scan choose the error.

erskingardner (commented):
Would request changes if GitHub allowed it on an author-owned PR. The set-oriented resolver, non-reserving prewarm, and create-time revalidation match #1486, but this revision should not merge yet.

- CI is red: `eprintln!` in `src/tests.rs` fails `production_library_sources_do_not_write_direct_output`
- Relay-list fallback `?` breaks the input-order / partial-set contract the KeyPackage stage already honors
- Runtime prewarm can start relay work after shutdown

I independently verified the two CodeRabbit majors above. The empty-batch fallback nit is acceptable here: a successful empty multi-author answer is not proof the relay honored the filter, so per-member retry is the conservative Nostr behavior.

crates/marmot-app/src/runtime/commands.rs
```diff
@@ -170,6 +170,30 @@ impl AccountManager {
...
+    /// mutation queue. Dropping the returned future cancels the caller's wait;
+    /// relay-plane coalescing may still safely satisfy a later identical create
+    /// request, and no KeyPackage is reserved or consumed here.
+    pub async fn prewarm_group_member_key_packages(
+        &self,
+        account_ref: &str,
+        members: &[String],
+    ) -> Result<crate::MemberKeyPackagePrewarmSummary, AppError> {
+        let started_at = Instant::now();
+        self.resolve(account_ref)?;
```
erskingardner, line 184: This path bypasses `worker_commands`, so it never hits `ensure_running()`. After `begin_shutdown()` the method can still start `fetch_directory_events`, and `drain_in_flight_work()` does not track this direct call.

Other commands that touch the app without the worker (`delete_group_local`, sign-out, account teardown) guard this. Add `self.shared.lifecycle().ensure_running()?` before `resolve`.

erskingardner (commented):
Recommend merge on `d69a2b0` (GitHub blocks a formal approve on an author-owned PR). The previous blockers are fixed:

- Scaling report no longer uses `eprintln!`; the structured `tracing::info!` call has `target` and `method` and should satisfy the direct-output / tracing audits
- Relay-list fallback now degrades per member, continues KeyPackage work for siblings, and the new test locks input-order error selection
- Runtime prewarm calls `ensure_running()` before `resolve`, with a shutdown regression test

No new blocking issues. The empty-batch per-member retry remains the conservative Nostr behavior. PR is still draft; mark ready when you want it in the merge queue. Some CI jobs were still running at review time.

PR #1495, opened 2026-08-18
marmot-uniffi: publish profiles with account-owned relays

### LINKED ISSUE #1491: marmot-uniffi: publish user profiles using MDK-owned account relay configuration
## Problem

The host currently has to read MDK's account relay configuration, reproduce relay fallback policy, and pass those relay lists back into MDK before publishing a profile.

At 82af46ca1352, [`publish_user_profile` requires both `default_relays` and `bootstrap_relays` from the caller](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-uniffi/src/commands/account.rs#L501), even though the account runtime owns the current relay-list projection and setup configuration.

White Noise commit [5b95342](https://github.com/marmot-protocol/whitenoise-ios/commit/5b953425dc89da032ab1412e34ae63138c0695ba) centralizes its duplicated host-side reads, but profile editing and onboarding still perform a local read-before-write and retain a second implementation of the fallback policy.

## Requested outcome

Add an account-owned profile publication path, for example:

```text
publish_user_profile_using_account_relays(
    account_ref: String,
    profile: UserProfileMetadataFfi
) -> UserProfilePublishResultFfi
```

Alternatively, make relay arguments optional with an unambiguous “resolve from this account” semantic. Preserve an explicit override API for diagnostics/tests or specialized clients that genuinely need it.

## Correctness constraints

- Resolve one coherent account-scoped relay configuration at the start of the operation; do not mix projections from different refresh moments.
- Centralize and document the fallback behavior for missing/empty publish and bootstrap lists.
- Route every selected endpoint through the existing retired-relay and host-safety validation discipline.
- Preserve profile merge/recency behavior and unknown-field handling; this issue changes relay ownership, not kind-0 content semantics.
- Do not let an account reference select another account's relay configuration.
- Keep logs and metrics aggregate; never expose relay URLs through telemetry.
- If the host needs success UI, return bounded aggregate attempted/succeeded counts rather than requiring a second relay-list read.

## Acceptance criteria

- Profile editing and onboarding can publish without first calling `account_relay_lists`.
- Tests cover populated lists, empty publish lists, missing bootstrap lists, signed-out/stopped accounts, invalid endpoints, and retired relays.
- The account-owned path produces the same selected endpoints and publication policy as the canonical MDK fallback rules.
- Explicit caller overrides remain available and are clearly distinguished from account-owned selection.
- White Noise can delete its relay fallback/round-trip plumbing for profile publication.
- Measure the removed host read/FFI hop and confirm no extra network work is introduced.

## Related work

- #1483 owns generated-account local readiness and returning the generated profile.
- #1484 owns initial setup publication ordering.
- This issue owns subsequent profile-publication relay selection and API ownership.

erskingardner:
## Summary
- add `publish_user_profile_using_account_relays`, which captures one account-scoped relay projection and uses it for both profile merge reads and publication
- centralize publish/bootstrap fallback plus invalid, unsafe, and retired endpoint filtering while preserving the explicit relay-override API and kind-0 merge behavior
- reduce the normal host path from two FFI calls (`account_relay_lists` plus publish) to one, with no added network operations

## Test plan
- [x] Account-owned selection covers populated, empty-publish, missing-bootstrap, invalid, retired, cross-account, signed-out, and stopped cases
- [x] `cargo test -p marmot-app profile_publish -- --nocapture`
- [x] `just fast-ci`
- [x] iOS and macOS XCFramework generation plus native and SwiftPM validation
- [x] Kotlin generation plus all four Android JNI ABI builds
- [ ] Full `cargo test -p marmot-app` remains blocked by five unrelated `relay_runtime` failures that reproduce on clean `origin/master`; all 689 library tests and 99 of 104 relay-runtime tests passed

Fixes #1491

## Summary by CodeRabbit

* **New Features**
  * Added account-specific profile publishing using the account’s configured relay settings.
  * Added safe relay selection with publish-relay preference and fallback options.
  * Profile publishing now supports explicit relay overrides.
  * Added validation for signed-out accounts and cases with no usable relays.

* **Bug Fixes**
  * Prevented profile updates from being published to another account’s relay configuration.

### DIFF
--- a/crates/marmot-app/src/directory/methods.rs
+++ b/crates/marmot-app/src/directory/methods.rs
@@ -530,13 +530,14 @@ impl MarmotApp {
         &self,
         relay_lists: &AccountRelayListStatus,
     ) -> Result<Vec<TransportEndpoint>, AppError> {
-        let published = if relay_lists.nip65.relays.is_empty() {
-            &relay_lists.default_relays
-        } else {
-            &relay_lists.nip65.relays
-        };
         let published = self.retain_safe_discovered_endpoints(
-            published.iter().cloned().map(TransportEndpoint).collect(),
+            relay_lists
+                .nip65
+                .relays
+                .iter()
+                .cloned()
+                .map(TransportEndpoint)
+                .collect(),
             "account-owned profile publish relays",
         );
         if !published.is_empty() {
--- a/crates/marmot-app/src/directory/methods.rs
+++ b/crates/marmot-app/src/directory/methods.rs
@@ -485,11 +485,26 @@ impl MarmotApp {
         bootstrap: AccountRelayListBootstrap,
     ) -> Result<(), AppError> {
         let account = self.account_home().account(label)?;
-        let signer = self.account_signer_for_summary(&account)?;
         let endpoints = self.outbox_endpoints(
             &account.account_id_hex,
             publish_endpoints_from_bootstrap(&bootstrap),
         );
+        self.publish_user_profile_to_endpoints(&account.label, profile, endpoints)
+            .await
+    }
+
+    /// Publish kind-0 metadata to an already-selected, account-scoped route.
+    ///
+    /// This is the action boundary used when the runtime has captured one
+    /// coherent relay-list snapshot and must not re-read it before publishing.
+    pub(crate) async fn publish_user_profile_to_endpoints(
+        &self,
+        label: &str,
+        profile: UserProfileMetadata,
+        endpoints: Vec<TransportEndpoint>,
+    ) -> Result<(), AppError> {
+        let account = self.account_home().account(label)?;
+        let signer = self.account_signer_for_summary(&account)?;
         let content = serde_json::to_string(&profile_content_json(&profile))?;
         let event = NostrTransportEvent::new_unsigned(
             account.account_id_hex.clone(),
@@ -501,6 +516,48 @@ impl MarmotApp {
             .publish_event(&endpoints, &event, 1)
             .await?;
         Ok(())
+    }
+
+    /// Select profile publication endpoints from one account relay-list
+    /// snapshot. Prefer the account's published NIP-65 write relays, then its
+    /// remembered bootstrap relays. A missing bootstrap list falls back to the
+    /// snapshot's compatibility default/publish list.
+    ///
+    /// Every tier is filtered through the discovered-endpoint safety policy.
+    /// If a tier contains only invalid, unsafe, or retired endpoints, selection
+    /// continues to the next tier; no unusable endpoint reaches the dialer.
+    pub(crate) fn account_profile_publish_endpoints(
+        &self,
+        relay_lists: &AccountRelayListStatus,
+    ) -> Result<Vec<TransportEndpoint>, AppError> {
+        let published = if relay_lists.nip65.relays.is_empty() {
+            &relay_lists.default_relays
+        } else {
+            &relay_lists.nip65.relays
+        };
+        let published = self.retain_safe_discovered_endpoints(
+            published.iter().cloned().map(TransportEndpoint).collect(),
+            "account-owned profile publish relays",
+        );
+        if !published.is_empty() {
+            return Ok(published);
+        }
+
+        let bootstrap = if relay_lists.bootstrap_relays.is_empty() {
+            &relay_lists.default_relays
+        } else {
+            &relay_lists.bootstrap_relays
+        };
+        let bootstrap = self.retain_safe_discovered_endpoints(
+            bootstrap.iter().cloned().map(TransportEndpoint).collect(),
+            "account-owned profile bootstrap relays",
+        );
+        if bootstrap.is_empty() {
+            return Err(AppError::RelayDirectory(
+                "account relay configuration has no usable profile publication endpoints".into(),
+            ));
+        }
+        Ok(bootstrap)
     }
 
     pub async fn publish_account_follow_list(
--- a/crates/marmot-app/src/runtime/tests.rs
+++ b/crates/marmot-app/src/runtime/tests.rs
@@ -6,20 +6,17 @@ use cgka_traits::transport_adapter::{
 
 use super::subscriptions::chat_list_mute_expiries;
 use super::*;
+use crate::publish_endpoints_from_bootstrap;
 use crate::tests::ScriptedPushRelayClient;
 
 fn profile_relay_status(
     publish_relays: &[&str],
     bootstrap_relays: &[&str],
-    default_relays: &[&str],
 ) -> AccountRelayListStatus {
-    AccountRelayListStatus {
+    let mut status = AccountRelayListStatus {
         complete: false,
         missing: Vec::new(),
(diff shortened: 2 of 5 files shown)


crates/marmot-app/src/directory/methods.rs
```diff
@@ -503,6 +518,48 @@ impl MarmotApp {
...
+        } else {
+            &relay_lists.nip65.relays
+        };
+        let published = self.retain_safe_discovered_endpoints(
+            published.iter().cloned().map(TransportEndpoint).collect(),
+            "account-owned profile publish relays",
+        );
+        if !published.is_empty() {
+            return Ok(published);
+        }
```
erskingardner: This is not the canonical fallback.

`outbox_endpoints` prefers the safe NIP-65 write set and only then uses the caller fallback. `publish_endpoints_from_bootstrap` is bootstrap, or default **only when bootstrap is empty**. This block does something else: when `nip65.relays` is empty it promotes `default_relays` to the first tier and can return it without ever considering `bootstrap_relays`.

That also fights `AccountRelayListStatus::refresh()`, which sets `default_relays = nip65.relays`. In production the first branch is a no-op; in this helper it is a different policy.

Simplest fix: run the existing two-tier rule against the captured snapshot, and keep the extra `retain_safe_discovered_endpoints` pass on the fallback (the issue asked for that).

```rust
let published = self.retain_safe_discovered_endpoints(
    relay_lists.nip65.relays.iter().cloned().map(TransportEndpoint).collect(),
    "account-owned profile publish relays",
);
if !published.is_empty() {
    return Ok(published);
}
let fallback = if relay_lists.bootstrap_relays.is_empty() {
    &relay_lists.default_relays
} else {
    &relay_lists.bootstrap_relays
};
```

The FFI docstring already describes this order. Please make the code match it, or this will drift from every other outbox publish.

crates/marmot-app/src/runtime/tests.rs
```diff
@@ -8,6 +8,121 @@ use super::subscriptions::chat_list_mute_expiries;
...
+            .unwrap(),
+        vec![TransportEndpoint("wss://bootstrap.example".into())]
+    );
+
+    let missing_bootstrap = profile_relay_status(&[], &[], &["wss://default.example"]);
+    assert_eq!(
+        app.account_profile_publish_endpoints(&missing_bootstrap)
+            .unwrap(),
+        vec![TransportEndpoint("wss://default.example".into())]
+    );
```
erskingardner: These fixtures never `refresh()`, so they invent a status the directory cache does not store: empty NIP-65 with a distinct `default_relays` list. After `refresh()`, `missing_bootstrap` has no usable endpoints and this assertion is false.

The distinguishing case the docs claim is also missing: empty NIP-65 with **both** bootstrap and default populated must select bootstrap. Today that case returns default.

Please construct statuses the way production does (refresh / `relay_list_status_from_records` / remember-after-setup), and add an equivalence check against `outbox_endpoints` + `publish_endpoints_from_bootstrap` on the same snapshot. That is the #1491 acceptance criterion this test is supposed to lock.

crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -7965,6 +7964,71 @@ async fn account_publishes_route_to_own_nip65_not_bootstrap() {
...
+    let from_bob_relay = app
+        .fetch_current_user_profile_for_account_id(
+            &alice.account.account_id_hex,
+            vec![endpoint(&bob_relay_url)],
+        )
+        .await
+        .unwrap()
+        .and_then(|profile| profile.name);
+    assert_ne!(from_bob_relay.as_deref(), Some("AliceAccountOwned"));
+
```
erskingardner, line 8028: This does not prove account-owned selection.

`MarmotApp::with_relay_and_config(..., alice_relay_url)` makes the app-wide default Alice's relay. Publishing Alice and finding the profile on Alice's relay is also what “use configured defaults” would do. Checking Bob's relay only shows you did not publish *everywhere*.

Publish **Bob** (relays ≠ app default) and assert the profile is on Bob's relay and absent from Alice's. That is the actual #1491 failure mode: an account ref must not adopt another account's — or the process default — relay configuration.

erskingardner (commented):
**Request changes** (posted as comment — GitHub blocks self-reviews on this PR).

The new account-owned entry point and snapshot-once boundary match #1491. The centralized selection helper does not.

`account_profile_publish_endpoints` is documented (and required) to match canonical MDK fallback: safe NIP-65 write relays, else remembered bootstrap, else default. The implementation treats `default_relays` as a *first-tier substitute* when NIP-65 is empty, which prefers default over bootstrap. That disagrees with `outbox_endpoints` + `publish_endpoints_from_bootstrap`, and the unit fixtures never call `refresh()` so they test a status shape the cache does not persist (`default_relays` is always copied from `nip65.relays`).

The Alice/Bob integration test also does not prove account-owned routing: the app-wide default *is* Alice's relay, so the pre-fix “use configured defaults” path would pass too.

Please make the helper a snapshot-aware `outbox_endpoints`, lock it with production-shaped fixtures (including empty NIP-65 + both fallback lists), and publish the non-default account in the cross-account test.

Non-blocking: merge reads now use the selected write set instead of bootstrap relays. Cache fallback likely preserves unknown fields, but that is a change from the override path.

crates/marmot-app/src/directory/methods.rs
erskingardner: Fixed in fd0ccccd: the snapshot selector now always evaluates the safe NIP-65 write set first, then remembered bootstrap, and uses the compatibility default only when bootstrap is empty. The fallback tier remains safety-filtered.

crates/marmot-app/src/runtime/tests.rs
erskingardner: Fixed in fd0ccccd: the fixtures now call refresh() to create the production cache shape. Coverage now locks bootstrap-over-default ordering, equivalence with outbox_endpoints plus publish_endpoints_from_bootstrap on the cached snapshot, and continuation from an unusable NIP-65 tier to safe bootstrap.

crates/marmot-app/tests/relay_runtime.rs
erskingardner: Fixed in fd0ccccd: the integration test now publishes Bob while the application default remains Alice’s relay, then asserts Bob’s profile is present on Bob’s relay and absent from Alice’s.

erskingardner (commented):
**Recommend merge** (posted as comment — GitHub blocks self-approvals on this PR).

Previous findings are addressed on fd0ccccd.

`account_profile_publish_endpoints` now matches the two-tier outbox rule (safe NIP-65, then bootstrap-or-default). Fixtures go through `refresh()`, the distinguishing bootstrap-over-default case is locked against `outbox_endpoints` + `publish_endpoints_from_bootstrap`, and the integration test publishes Bob while the app default remains Alice’s relay.

The account-owned API, snapshot-once boundary, and override path look right.

PR #1496, opened 2026-08-18
Expose consistent group conversation snapshot

### LINKED ISSUE #1490: marmot-uniffi: expose a consistent group conversation snapshot
## Problem

Conversation loading needs group details and group-management state together, but UniFFI exposes them as two asynchronous calls:

- [`group_details`](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-uniffi/src/commands/group.rs#L357)
- [`group_management_state`](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-uniffi/src/commands/group.rs#L379)

White Noise consequently performs those calls sequentially in [`ConversationViewModel.refreshGroupManagement`](https://github.com/marmot-protocol/whitenoise-ios/blob/5b953425dc89da032ab1412e34ae63138c0695ba/whitenoise-ios/Conversation/ConversationViewModel.swift#L1999). This adds an FFI/worker round trip and permits a group mutation between reads, producing a details/management pair from different logical moments.

## Requested outcome

Add a combined read such as:

```text
GroupConversationSnapshotFfi {
    details: GroupDetailsFfi,
    management_state: GroupManagementStateFfi
}

group_conversation_snapshot(
    account_ref: String,
    group_id_hex: String
) -> GroupConversationSnapshotFfi
```

Compute both values from one documented, consistent storage/worker snapshot and return them through one asynchronous UniFFI call.

## Correctness constraints

- Define the consistency boundary precisely: both records must describe the same group state/epoch and local projection frontier.
- Avoid an await or queue re-entry between capturing the inputs for the two projections.
- Do not simply wrap the two existing public async calls sequentially.
- Preserve the existing individual APIs for callers that need only one projection.
- Return one coherent typed error if the group disappears, the account stops, or required state is unavailable.
- Do not add network fetch, relay catch-up, or publication to this local read.
- Keep timing dimensions aggregate and privacy-safe.

## Acceptance criteria

- White Noise can replace the sequential details/management calls with one binding call.
- A concurrency regression test mutates the group around the read and proves the returned pair is internally consistent.
- The combined call performs at most one worker command/queue admission and does not duplicate expensive storage decoding.
- Returned values match the individual APIs when no mutation occurs.
- Add a reviewed performance dimension for the combined snapshot and provide before/after measurements for cold and warm conversation loading.
- Bindings and generated Swift API tests cover the complete result shape and error mapping.

## Related work

- #1487 enriches the immediate group-creation result only with projections already available at that boundary.
- This issue owns general conversation loading and must not make group creation calculate additional projections synchronously.

erskingardner:
## Summary

- add `groupConversationSnapshot` so conversation loading receives details and management state from one account-worker snapshot
- derive management actions synchronously from the exact returned group/member/MLS projections
- route existing group-details reads through the same one-command boundary and add aggregate telemetry plus generated Swift/Kotlin API coverage

## Consistency boundary

The group record, member projection, and MLS state are captured by one `GroupRoster` worker command from one live-session or startup-snapshot frontier. The worker does not yield or re-enter its queue while capturing them. Management state is derived from those exact details without another await. Display names are best-effort synchronous batch enrichment and do not affect management decisions.

## Performance

Release-mode local mock-relay measurement:

| Path | Cold | Warm per load (100 iterations) |
| --- | ---: | ---: |
| Existing two-call path | 3.18 ms | 1.33 ms |
| Combined snapshot | 0.74 ms | 0.99 ms |

That is a 76.8% cold reduction and 25.6% warm reduction. This is conservative because this patch also routes the individual `groupDetails` API through the one-command snapshot; before the patch, the two host calls admitted four worker commands, while the combined API admits one.

## Validation

- `just fast-ci`
- `cargo test -p marmot-uniffi` (121 unit, 2 media-fixture, and 28 smoke tests)
- snapshot error/telemetry round-trip regression
- concurrent mutation consistency regression
- full `marmot-app` run: 685 passed; two unrelated contention-sensitive tests failed, then both passed in isolation
- iOS and macOS XCFramework artifact validation and SwiftPM compilation
- Kotlin bindings generated for all four Android ABIs

Fixes #1490


crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -2960,6 +2960,89 @@ async fn app_runtime_serves_member_reads_before_initial_catch_up_completes() {
...
+
+    let observed = (
+        concurrent.group.profile.name.as_str(),
+        concurrent.mls_state.epoch,
+    );
+    let before_pair = (before.group.profile.name.as_str(), before.mls_state.epoch);
+    let after_pair = (after.group.profile.name.as_str(), after.mls_state.epoch);
+    assert!(
+        observed == before_pair || observed == after_pair,
+        "snapshot must be wholly before or wholly after the concurrent commit: observed {observed:?}, before {before_pair:?}, after {after_pair:?}"
```
erskingardner, line 3029: The `(name, epoch)` pairing is a real check because `UpdateGroupData` commits both in one worker command.

`members.len()` vs `mls_state.member_count` and the admin⊆members walk do not change under a profile rename, so they would not catch a torn members / MLS / admin frontier. #1490 asked for the returned details/management pair to stay consistent around a mutation. Management is a pure function of details, so this app-level test is enough if those three projections stay aligned — a membership or admin-policy mutation (invite/promote) would make a mixed snapshot fail this test, and a UniFFI assertion that `management_state` equals `group_management_state_ffi` on the concurrent details would close the pair.

Not blocking: the worker capture is synchronous, and this race still guards name vs epoch.

erskingardner (commented):
## Review

This is the right shape for #1490: one `GroupRoster` worker command, no queue re-entry between the three projections, and management derived synchronously from those exact details. Routing the existing `groupDetails` / `groupManagementState` builders through the same snapshot is the correct way to keep the individual APIs and still admit only one worker command.

Error mapping for unknown group and a stopped runtime is covered. Telemetry stays aggregate and unlabeled. Swift package validation actually references the new record and async method.

GitHub will not accept an approve on the author's own PR, but I would approve this. Non-blocking notes:
- The concurrency test guards profile name vs MLS epoch (a real pairing under `UpdateGroupData`) but does not stress membership/admin fields or the UniFFI details/management pair. See the inline note.
- `group_details_read` in `docs/marmot-architecture/telemetry.md` still describes the old multi-lookup path.
- `group_management_state_for` / `group_mutation_result_for` still take `_group_id_hex` after the last use was removed; drop the parameter.

PR #1498, opened 2026-08-18
marmot-app: pre-upload founding group images

### LINKED ISSUE #1485: marmot-app: remove initial group-image upload from the create-group critical path
## Problem

Creating a group with an initial image waits for the entire encrypted blob upload before the canonical MLS group is prepared and persisted.

At 82af46ca1352:

- [the create flow uploads the optional group image before MLS preparation](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/client/mod.rs#L906);
- [MLS preparation/persistence begins only after that upload finishes](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/client/mod.rs#L970);
- [the group-image helper validates only non-empty input before encrypting and uploading](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/media/group_image.rs#L65);
- the generic Blossom client permits a [long blob-transfer timeout](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/media/blossom.rs#L24).

On slow or degraded media infrastructure, the Create spinner is therefore dominated by a non-MLS attachment. A large image also has no explicit group-image-specific byte/pixel budget at this boundary.

## Requested outcome

Move initial image transfer out of the user-visible canonical create-group critical path while preserving an explicit, testable image semantic.

Acceptable designs include:

1. pre-uploading during group composition and passing a validated uploaded-image artifact into creation; or
2. creating the group first, then applying the image as durable resumable background work through an explicitly fast-create API.

Do not silently change the existing API from “created with this founding image” to “image may appear later.” If both semantics are needed, expose and document both.

## Correctness constraints

- Preserve encryption, media-key durability, retry, and orphan-cleanup behavior.
- A process death after group creation must not lose a promised background image update.
- A failed image upload must be visible and retryable rather than reported as group-creation failure after the group already exists.
- Add a group-image-specific size/shape budget before expensive encryption or upload work.
- Avoid duplicate uploads for retries of the same durable intent.
- Preserve canonical group creation and Welcome-delivery invariants.

## Acceptance criteria

- The fast canonical group-create response is independent of Blossom upload latency.
- Tests cover slow upload, upload failure, cancellation, restart, retry, and idempotency.
- The host receives distinct group-created and image-upload/update states.
- The selected API semantics are documented, including whether the image is present in founding metadata or a follow-up commit.
- Add timings for pre-processing, upload, canonical group create, and follow-up image commit.
- Provide a repeatable before/after measurement with no image, a typical image, maximum accepted image, and stalled media server.

## Related work

- #1451 moved Welcome fanout after the canonical response. This issue applies the same critical-path discipline to optional media without changing Welcome ownership.

erskingardner:
## Summary

- add a SQLCipher-backed prepared founding-image artifact with staged, uploading, uploaded, failed, and consumed host-visible states
- validate a 10 MiB / 4096 px image budget before encryption, retry the exact ciphertext and hash, suppress concurrent duplicate uploads, and recover safely across cancellation or restart
- add an explicit create API that accepts only an uploaded artifact, installs it in epoch-zero MLS metadata, and performs no Blossom request; retain the legacy all-in-one API upload-before-success ordering while intentionally applying the new byte, dimension, pixel, and PNG/JPEG/GIF/WebP validation before canonical creation
- expose the flow through Rust and UniFFI, add preprocess/upload telemetry, document lifecycle and bounds, and add a repeatable measurement matrix

## Root cause

The existing create flow constructed its optional components by awaiting group-image encryption and the full Blossom PUT before it called MLS prepare/persist. Since the helper had no image-specific byte or dimension limit, slow infrastructure and large images directly extended the canonical create response.

## Measurement

Run just bench-group-image-create for stable MDK_BENCH rows. A debug smoke on this machine reported:

- no image: canonical create 22 ms
- typical 73-byte fixture: prior serialized model 42 ms; canonical create 33 ms
- exact 10 MiB limit: prior serialized model 4,990 ms; canonical create 461 ms
- stalled Blossom response: upload remained stalled while canonical founding-image create from an already uploaded artifact completed in 28 ms

These are diagnostic values, not performance thresholds. The harness reports preprocessing and upload separately and models the old serialized boundary as their sum with canonical create.

## Verification

- just fast-ci
- prepared-artifact storage, restart, cap, expiry, retry, cancellation, duplicate-suppression, idempotent-create, and founding-metadata regressions
- cargo test -p marmot-uniffi: 119 unit, 2 fixture, and 28 smoke tests passed
- cargo check -p marmot-uniffi --features otlp-export
- iOS device/simulator XCFramework plus generated Swift binding
- macOS 15 XCFramework plus generated Swift binding
- generated Kotlin binding plus arm64-v8a, armeabi-v7a, x86, and x86_64 JNI libraries

The full marmot-app run reached 695 passing tests; its existing greater-than-64-MiB loopback media test timed out under concurrent build load and then passed in isolation in 86.71 seconds.

Fixes #1485

## Summary by CodeRabbit

- **New Features**
  - Added staged founding-image uploads for group creation, including validation, encryption, resumable transfers, and status tracking.
  - Added configurable group creation options, including initial images and disappearing-message retention.
  - Added detailed creation results and support across runtime and FFI interfaces.
- **Bug Fixes**
  - Improved recovery from interrupted uploads and partially completed group creation.
  - Prevented duplicate group creation during retries.
- **Documentation**
  - Documented the workflow, lifecycle limits, and telemetry.
- **Tests**
  - Added integration coverage and benchmarks for uploads, retries, failures, and stalled transfers.


crates/marmot-app/src/tests.rs
```diff
@@ -42,6 +42,316 @@ use crate::key_package_records::{
...
+            "",
+            &[],
+            &staged.upload_id,
+            &telemetry,
+        )
+        .await
+        .unwrap();
+    let create = create_started.elapsed();
+    let baseline_serialized = preprocess + upload_elapsed + create;
+    println!(
```
erskingardner: `just ci` is already red on this: `cgka-conformance-simulator::tracing_audit::production_library_sources_do_not_write_direct_output` rejects `println!` in `marmot-app/src/**` (hits this line plus 197 and 250).

`startup_scaling` avoids that by living in `crates/marmot-app/tests/startup_scaling.rs`. Move this ignored matrix there (or another `tests/*.rs` file) so the `MDK_BENCH` rows stay greppable without violating the library-source audit.

crates/marmot-app/src/client/mod.rs
```diff
@@ -869,7 +1056,66 @@ impl AppClient {
...
+        // Recover a crash after canonical creation but before the artifact's
+        // consumed marker. The exact randomized component bytes uniquely bind
+        // the artifact to the already-created group, so retry returns that
+        // group rather than creating a duplicate.
+        let component_data_hex = hex::encode(&record.component_data);
+        if let Some(group) = self
+            .state
+            .groups
+            .iter()
+            .find(|group| group.image.data_hex == component_data_hex)
```
erskingardner: This is the only crash-before-consumed recovery, and it only sees the in-memory projection. The create path below already treats `add_group` / `save_state` failure as expected ("account open will reconcile it") and still returns `Ok(group_id)` after a failed consume.

If the host retries this `upload_id` while the group is canonical in MLS but absent from `state.groups` — process death after persist and before consume, or the warned projection miss — this `find` fails and the next call creates a second group. The docs say that retry is idempotent.

Scan the engine component (or persist `group_id` as soon as MLS create succeeds) instead of relying on `state.groups`. Add a test that leaves the artifact `Uploaded`, omits the projection row, and asserts retry returns the original id.

crates/marmot-app/src/client/mod.rs
```diff
@@ -1058,6 +1338,27 @@ impl AppClient {
             local_projection_started_at.elapsed(),
             local_projection_saved,
         );
+        if let Some(upload_id) = prepared_upload_id
```
erskingardner: Consume is the durable idempotency key, but it runs after the best-effort projection save. Move this to immediately after `let group_id = prepared.group_id` (still never fail the caller). Then a crash or projection miss still leaves a `Consumed` marker the retry path already handles, instead of depending on `image.data_hex` being present in `state.groups`.

erskingardner (commented):
Could not use Request changes on this PR (GitHub rejects self-reviews), so this is a comment-only review. I would request changes on the two items below.

The design matches #1485: pre-upload keeps founding-image semantics, the legacy API is unchanged, and the durable staged/failed/uploaded/consumed machine plus the 10 MiB / 4096 px budget are the right shape.

Two issues should land before merge:

1. **CI is red** — `println!` in `marmot-app/src/tests.rs` fails `production_library_sources_do_not_write_direct_output`. Move the measurement matrix to an integration test, the same way `startup_scaling.rs` does.
2. **Advertised crash-before-consumed recovery can create a second MLS group.** Consume happens after a projection step this function already treats as best-effort, and retry only scans `state.groups`. If that scan misses, the next `create_group_with_prepared_initial_image` call creates a duplicate. Consume immediately after the canonical `group_id` exists, and cover the Uploaded-but-not-consumed retry in a test.

Non-blocking: `upload_prepared_group_image` returns `Ok(Failed)` rather than `Err`. Hosts that only check the `Result` will treat a durable upload failure as success. Either return the error after persisting `failed`, or make the UniFFI docs say the caller must inspect `state`.

crates/marmot-app/src/tests.rs
erskingardner: Addressed in 6b27f664: the MDK_BENCH matrix now lives in crates/marmot-app/tests/group_image_create.rs, the Just target runs that integration test, and the production direct-output audit passes.

crates/marmot-app/src/client/mod.rs
erskingardner: Addressed in 6b27f664: retry now scans the authoritative live engine component for the exact prepared bytes, independent of state.groups. The regression test removes the app projection while leaving the artifact Uploaded and verifies retry returns the original group without another MLS create.

crates/marmot-app/src/client/mod.rs
erskingardner: Addressed in 6b27f664: consumption now runs immediately after the canonical group id exists, before Welcome bookkeeping and best-effort projection. A failed consume remains non-fatal and the authoritative engine-component recovery path closes that crash window on retry.

erskingardner:
Also addressed the review summary’s non-blocking upload-result concern in 6b27f664: a failed transfer is first persisted as `Failed`, then returned to the host as `Err`; a regression test verifies both the result contract and durable retry status. The branch was rebased onto current `master` before the update.

crates/marmot-app/src/client/mod.rs
```diff
@@ -946,39 +1200,73 @@ impl AppClient {
...
                 ) {
                     Some(GROUP_BLOSSOM_IMAGE_COMPONENT_ID) => {
-                        let upload =
-                            upload_group_image(&image.plaintext, &image.media_type, None).await?;
-                        let input = AppGroupImageInput::from(upload);
+                        let data = match image {
+                            InitialGroupImageSource::Inline(image) => {
+                                let upload =
+                                    upload_group_image(&image.plaintext, &image.media_type, None)
+                                        .await?;
```
Datawav, line 1282: **Low (compatibility/docs)** — Legacy inline-image create now hard-fails on inputs it previously accepted, contradicting the PR's "retain the legacy all-in-one API semantics" claim.

`upload_group_image` now runs the new `validate_group_image_input` (`crates/marmot-app/src/media/group_image.rs`). The old code rejected only empty plaintext and a non-canonical media type; the new validator also rejects >10 MiB encoded bytes, width/height >4096 px or >16,777,216 pixels, undecodable formats, and declared-vs-detected format mismatch. Here the `InitialGroupImageSource::Inline` arm propagates that error with `?`, so the entire group creation fails.

Consequence: an existing host calling the legacy all-in-one create with an oversized inline image previously got a created group with the image Blossom-uploaded; it now gets `InvalidEncryptedMedia` and no group. Plausibly intentional hardening, but it is an unannounced contract change on a path the PR body says is unchanged, and no test pins the legacy behavior (the cap tests exercise only the new stage API).

Smallest sound correction: (a) document in the PR/release notes that the legacy create now enforces the 10 MiB / 4096 px validation and fails on violation, correcting the "retained semantics" wording, or (b) in the `Inline` arm only, downgrade validation failure to skipping the image component with a redacted warning. Add one regression test calling the legacy create with an 11 MiB image asserting whichever behavior is chosen.

Datawav (commented):
## Automated review

Reconciliation complete. The delta is real (legacy inline path now validates size/dimensions/format pre-upload and propagates the error, failing group creation); the disputed parts are severity and framing. The source_url overreach in one review does not survive — the avatar-URL branch bypasses `upload_group_image`. The mutually-unreachable `(Some,Some)` arm, ephemeral HashSet, error-precedence, and schema-rollback items were all independently verified as non-issues by both sides. Discussion is empty; checks are all success/completed with one by-design skip.

---

Final audit — marmot-protocol/mdk PR #1498 (initial, head 6b27f66)

One supported finding.

Low (compatibility/docs) — Legacy inline-image create now hard-fails on inputs it previously accepted, contradicting the PR's "retain the legacy all-in-one API semantics" claim.

Evidence: `crates/marmot-app/src/media/group_image.rs` — `upload_group_image` now delegates to `prepare_group_image_upload`, which calls the new `validate_group_image_input`. The old code rejected only empty plaintext and non-canonical media type; the new validator additionally rejects >10 MiB encoded bytes, width/height >4096 px or >16,777,216 pixels, undecodable/unrecognized formats, and declared-vs-detected format mismatch. In `crates/marmot-app/src/client/mod.rs`, the `InitialGroupImageSource::Inline` arm of `create_group_with_initial_profile_and_optional_telemetry` propagates that error with `?` inside the optional-components block, so the entire group creation fails.

Consequence: an existing host calling the legacy all-in-one create with an oversized or over-dimensioned inline image previously got a created group with the image Blossom-uploaded; it now gets `InvalidEncryptedMedia` and no group. This is reachable through the public legacy API and is plausibly intentional hardening — the PR's root-cause section notes the helper previously had no limits — but it is an unannounced contract change on a path the body says is unchanged, and no test pins the chosen legacy behavior (the cap tests only exercise the new stage API).

Smallest sound correction: pick one contract and codify it — either (a) document in the PR/release notes that the legacy create now enforces the 10 MiB / 4096 px validation and fails the create on violation, correcting the "retained semantics" wording, or (b) in the `Inline` arm only, downgrade validation failure to skipping the image component with a redacted warning, preserving prior degradation. Add one regression test calling the legacy create with an 11 MiB image asserting whichever behavior is chosen.

Rejected during reconciliation: the claim that the source_url/avatar-URL path is newly rejected (that branch never calls `upload_group_image`; unaffected). The `(Some,Some)` inline-plus-prepared arm is unreachable via public commands and fails closed. Ephemeral duplicate suppression, upload-failure error precedence, migration 0051 constraints/consume wipe, exact-ciphertext retry, debug redaction, and crash-recovery engine scan were all verified sound with test coverage.

erskingardner (commented):
Recommend merge. Previous findings are closed on `6b27f664`, and CI is green.

- Bench `println!` moved to `crates/marmot-app/tests/group_image_create.rs`; the library-source audit is no longer tripped.
- Consume now runs immediately after the canonical `group_id`. Retry scans live engine blossom-image components instead of `state.groups`, and `uploaded_prepared_group_image_retry_recovers_from_engine_without_projection` pins the crash window.
- Upload failures stay durable `Failed` and are also returned as `Err`.

Non-blocking: the legacy all-in-one create now rejects over-budget images (`10 MiB` / `4096 px`). That matches #1485; just tighten the PR wording so “retain legacy semantics” means upload-before-success, not “any previously accepted bytes still succeed.”

crates/marmot-app/src/client/mod.rs
erskingardner: Addressed in e301a6bb. I chose the documented-hardening option: the legacy all-in-one API retains its upload-before-success ordering but now explicitly enforces the 10 MiB, 4096 px, pixel-count, and format bounds before canonical creation. The PR summary and architecture docs now call out that compatibility-visible validation change, and `legacy_inline_group_image_create_rejects_oversized_input_before_canonical_creation` calls the legacy API with an oversized image and verifies both the typed failure and absence of a canonical group.

crates/marmot-app/src/media/group_image.rs
```diff
@@ -14,10 +17,58 @@ use crate::{AppError, AppGroupImageInput};
...
+        AppError::InvalidEncryptedMedia("group image format is not recognized".into())
+    })?;
+    let declared_format = image::ImageFormat::from_mime_type(&media_type).ok_or_else(|| {
+        AppError::InvalidEncryptedMedia("group image media type is not supported".into())
+    })?;
+    if detected_format != declared_format {
+        return Err(AppError::InvalidEncryptedMedia(
+            "group image media type does not match its bytes".into(),
+        ));
+    }
```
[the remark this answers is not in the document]
erskingardner: Addressed in e301a6bb. Group-image validation now maps only canonical `image/png`, `image/jpeg`, `image/gif`, and `image/webp` values to their enabled codecs before comparing the detected byte format. Every other media type uses the existing unsupported-media error path. `group_image_rejects_media_types_outside_avatar_allowlist` pins the rejection.

erskingardner (commented):
Recommend merge. The last open note is closed on `e301a6bb`.

- PR summary and architecture docs now say the legacy all-in-one API keeps upload-before-success ordering and also applies the 10 MiB / 4096 px / format budget before canonical create.
- `legacy_inline_group_image_create_rejects_oversized_input_before_canonical_creation` pins that: over-budget inline create returns `InvalidEncryptedMedia` and leaves no MLS group.
- Format checks now use an explicit PNG/JPEG/GIF/WebP allowlist instead of `ImageFormat::from_mime_type`.

Earlier crash-recovery, consume-timing, bench-audit, and upload-`Err` fixes remain intact after the rebase. Fast compile gates on this SHA are green; the remaining CI jobs were still running at review time.

PR #1499, opened 2026-08-18
Return generated accounts at durable local readiness

### LINKED ISSUE #1483: marmot-app: return generated-account creation at durable local readiness
## Problem

Generated-account creation still waits for relay work before returning the new account to the host. That makes the first-run experience inherit relay latency and availability even after all identity material can be made durable locally.

At 82af46ca1352:

- [the UniFFI command awaits the complete runtime setup operation](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-uniffi/src/commands/account.rs#L102) before returning an identity;
- [the generated-account path awaits bootstrap publication](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/runtime/mod.rs#L4394);
- it then [publishes the initial KeyPackage](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/runtime/mod.rs#L4440), reconciles runtime state, and completes the setup journal before returning.

Issue #1436 / PR #1446 reduced setup round trips and made publication tolerant of individual relay failures, but they did not establish a local-first return boundary.

## Requested outcome

Introduce an explicit generated-account lifecycle boundary where account creation can return once the identity, setup journal, exact initial KeyPackage material, and minimum host-visible account state are durably available locally. Continue bootstrap and KeyPackage publication as resumable background setup work.

The API must make local readiness distinct from network readiness so a host cannot accidentally present an account as invite-receivable before its KeyPackage is published.

## Correctness constraints

- Persist the exact initial KeyPackage bytes before any relay can observe setup state; a retry must not silently generate a different package.
- Cancellation, process death, or retry must never create a second identity.
- Record durable intent before external publication and resume it after restart.
- Preserve the current imported-account semantics unless they are deliberately redesigned and tested separately.
- Keep lifecycle/status reporting privacy-safe: aggregate phases and timings only.
- Do not weaken publication success policy or convert a durable setup failure into an unobservable background failure.

## Acceptance criteria

- A generated account becomes readable and usable through the app runtime without waiting for relay connect, request, or acknowledgement latency.
- A deterministic test with blocked/unreachable relays proves the local-ready return does not scale with relay timeout.
- Restart tests cover every boundary between local durability, bootstrap publication, initial KeyPackage publication, and journal completion.
- The host can observe at least `local_ready`, `publishing`/retrying, `network_ready`, and terminal recovery-required states.
- Repeated create/resume calls converge on one identity and one exact initial KeyPackage.
- Add end-to-end timing for local-ready and network-ready separately, and document the expected host behavior.

## Related work

- #1436 addressed setup round-trip count and relay failure tolerance.
- This issue owns the public readiness boundary. A separate issue should optimize the fully-network-ready path so the two changes remain independently testable.

## Binding result and identity phase evidence

The local-ready handoff must also eliminate the host's immediate profile read. The runtime already carries the generated profile in [`AccountSetupResult::profile`](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/runtime/mod.rs#L851), but [`create_identity` currently discards it](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-uniffi/src/commands/account.rs#L105) while converting the result to `AccountSummaryFfi`.

Expose a result record along these lines (exact naming is flexible):

```text
IdentityCreationResultFfi {
    account: AccountSummaryFfi,
    profile: UserProfileMetadataFfi
}
```

Requirements:

- Return the exact generated profile made durable for this setup attempt; do not regenerate a pseudonym during FFI conversion or retry.
- If local readiness precedes profile publication, the returned profile is explicitly local state and the setup readiness/status surface remains authoritative for network publication.
- Preserve a compatibility path for existing callers that only need `AccountSummaryFfi`.
- White Noise must be able to populate its profile editor without an immediate `user_profile` lookup.

Extend the structured timing acceptance criteria to distinguish:

- identity/keychain and local account storage initialization;
- default profile generation/local durability;
- default profile publication;
- initial KeyPackage generation/local durability;
- initial KeyPackage publication;
- local-ready handoff and final network-ready completion.

Expose these as reviewed, aggregate, privacy-safe fields through the existing app-performance snapshot rather than caller-defined labels.

erskingardner:
## Summary

- return generated identities after identity, SQLCipher storage, the exact default profile, and the exact signed initial KeyPackage publication artifact are durable locally
- continue bootstrap/profile/KeyPackage publication in restart-resumable background work without generating a second identity or KeyPackage
- expose `Initializing`, `LocalReady`, `Publishing`, `NetworkReady`, and `RecoveryRequired` through the Rust and UniFFI APIs while preserving the existing `create_identity` compatibility entry point
- integrate with setup KeyPackage prioritization so the exact setup artifact publishes before initial sync
- add privacy-safe per-stage telemetry, host integration guidance, and blocked-relay/restart/concurrency regression coverage

Fixes #1483

## Verification

- `just fast-ci`
- `cargo test -p marmot-account` (101 passed)
- `cargo test -p marmot-app --lib` (701 passed, 1 intentionally ignored)
- `cargo test -p marmot-uniffi` (122 unit, 2 fixture, and 29 smoke tests passed)
- generated iOS XCFramework and Swift bindings on the exact PR head
- generated macOS XCFramework and Swift bindings on the exact PR head
- generated Kotlin bindings and JNI libraries for arm64-v8a, armeabi-v7a, x86, and x86_64 on the exact PR head

Current verified head: `7fdea03ca667a6ec8bfd9bd26102beb65dd9ca5e`.

## Summary by CodeRabbit

* **New Features**
  * Account creation now returns profile information and setup readiness status.
  * Accounts can be used locally while network publication completes, with resumable setup.
  * Added prepared group-image staging, upload, status tracking, and recovery.
  * Group creation supports descriptions, retention settings, images, and detailed results.
  * Profiles can publish through the account’s configured relays.
* **Bug Fixes**
  * Improved recovery, retry, and consistency after interrupted setup or group creation.
* **Documentation**
  * Updated account setup and group-creation guidance.
* **Telemetry**
  * Added performance reporting for setup and group-creation stages.


crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3579,6 +3661,278 @@ impl MarmotAppRuntime {
     pub async fn create_or_import_account(
         &self,
         request: AccountSetupRequest,
+    ) -> Result<AccountSetupResult, AppError> {
+        if request.identity.is_none() && request.import_nsec.is_none() {
+            return self.create_generated_account_local_ready(request).await;
```
erskingardner: This is what makes the old one-in-flight generated journal user-visible. `create_nostr_account_from_setup` still does `resumable_generated_account_setup()` and returns that account, so a second `create_identity` in the same home while the journal exists is a silent resume of the first identity.

That used to be safe because `create_identity` did not return until `complete_account_setup` removed the journal. It is now the default success path, and it is already breaking CI:
- `agent-connector` tests create two identities on one runtime, then `add_members` fails with `DuplicateSignatureKey`
- CLI e2e asserts relay-list publication at create return

Keep this early return only on an explicit local-ready API. `create_identity` should still wait for `NetworkReady` so existing multi-account callers keep working. If a second generated create during an open journal is meant to be a retry, return a typed in-progress error or document the resume — do not present it as a newly created identity.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3579,6 +3661,278 @@ impl MarmotAppRuntime {
...
+            .accounts
+            .app
+            .account_home()
+            .account_setup_context(&account.label)?
+        else {
+            return Ok(());
+        };
+        let context: GeneratedAccountSetupContext = serde_json::from_slice(&bytes)?;
+        self.create_generated_account_local_ready(context.request())
+            .await?;
```
erskingardner: `start()` awaits this before `reconcile_accounts()`. `serde_json::from_slice` and `create_generated_account_local_ready` both propagate, so one unreadable or shape-mismatched `.account-setup-context.json` prevents every other account from starting.

The missing-context branch just above returns `Ok(())` and schedules nothing, so a LocalReady account with a lost context file also never gets a publication retry until the host calls create again.

Resume is repairable work: log a privacy-safe kind, return success, and let `reconcile_accounts()` proceed. Parse/create failures and missing context should use the same deferred path.

crates/marmot-uniffi/src/commands/account.rs
```diff
@@ -127,6 +132,50 @@ impl Marmot {
         })
     }
 
+    /// Create a generated identity and return at durable local readiness with
```
erskingardner, line 135: `create_identity` is labeled a compatibility entry point, but it now returns at local readiness and discards `result.readiness`. Existing hosts that only consume `AccountSummaryFfi` will treat the account as fully created and have no field telling them otherwise.

Keep `create_identity` on the old wait-for-`NetworkReady` contract, and let only this new method return at `LocalReady`. As written, White Noise and every in-repo caller of `create_identity` inherit the silent-resume / unpublished-relay-list behavior with no readiness signal.

erskingardner (commented):
The local-ready boundary is the right fix for #1483, and the generated-account tests in marmot-app/uniffi are pointed at the right invariants (exact KeyPackage bytes, blocked-publish return, restart across phases). This is not mergeable yet.

**Blocking**

1. Returning at `LocalReady` leaves the generated setup journal open. `create_nostr_account_from_setup` still resumes *any* incomplete generated journal, so a second `create_identity` in the same home silently returns the first account. That is why CI is red: every `agent-connector` shard fails with `DuplicateSignatureKey` on `add_members`, and CLI e2e fails because create no longer means “relay lists and KeyPackage are published.” The compatibility `create_identity` entry point also drops readiness, so existing hosts have no signal that they received a resume rather than a new identity.

   Simplest fix: keep `create_identity` waiting for `NetworkReady` (old contract). Only `create_identity_with_profile` / the explicit local-ready path should return early, and a second generated create while a journal exists should be documented as a resume — or rejected with a typed in-progress error.

2. `resume_generated_account_setup` is on the `start()` success path. A corrupt/unreadable `.account-setup-context.json` becomes `AppError::Json` and blocks every other account on the device. A missing context file does the opposite and returns `Ok(())` with no publication retry. Treat resume failures as deferred, privacy-safe background work so `reconcile_accounts()` still runs.

I independently agree with the existing notes on incoming-request relay validation vs persisted context, and on setup-state lookup `Err` being parked only in `setup_key_package_result`. Those are still present at this head.

**Non-blocking**

- `AccountSetupResult.readiness` is hardcoded `LocalReady` even when `account_setup_readiness()` already reports `Publishing`. Hosts should keep treating the query API as authoritative; the result field is misleading on retry.
- Docs still omit `Initializing`.

GitHub will not accept `REQUEST_CHANGES` on the author's own PR, so this is posted as a comment review. Treat the blocking items as merge blockers.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -3579,6 +3661,278 @@ impl MarmotAppRuntime {
...
+            let started_at = Instant::now();
+            let result = manager
+                .create_or_import_account_network_ready(context.request())
+                .await;
+            manager.shared.app_performance_telemetry().record(
+                AppPerformanceOperation::AccountSetupNetworkReady,
+                started_at.elapsed(),
+                result.is_ok(),
+            );
+            if let Err(error) = result {
```
Datawav: **Single-attempt background publication parks the account at a non-terminal phase until restart.** The spawned task calls `create_or_import_account_network_ready` once; on `Err` it only emits `AccountError(error.privacy_safe_kind())` and removes the account from `active_accounts`. `account_setup_readiness` maps `BootstrapPublicationStarted/Confirmed` and `KeyPackagePublicationStarted/Confirmed` to `Publishing` (and `LocalReady` stays `LocalReady`), while `resume_generated_account_setup` runs only from `start()`. So after a transient relay/network failure, a generated account sits at `Publishing`/`LocalReady` indefinitely in-session — never invite-receivable — unless the process restarts or the caller manually re-invokes `create_identity`. Consider a bounded in-session retry, or re-arm the resume path on background failure. (Confidence medium: the inner function body is not visible in the truncated head context; this holds if it does no intra-network retry.)

crates/marmot-app/src/tests.rs
```diff
@@ -3180,12 +3235,31 @@ async fn partial_generated_bootstrap_keeps_the_journaled_identity_for_retry() {
...
+    assert_eq!(
+        runtime
+            .account_setup_readiness(&local.account.label)
+            .unwrap(),
+        AccountSetupReadiness::Publishing
+    );
 
+    relay.allow_all_publish_kinds();
     let retried = runtime
         .create_identity(request())
```
Datawav: **Missing regression test for automatic in-session convergence after a background publish failure.** The added tests cover the success path, restart-resume, and concurrency, but here the failure path is recovered only via an explicit manual `create_identity` re-call after `allow_all_publish_kinds()`, and the pre-existing restart test covers restart recovery only. `ScriptedPushRelayClient::fail_publishes_of_kind` already exists, so this gap is untested rather than untestable. Please add a test that fails the first KeyPackage (or bootstrap) publish in-session and asserts eventual `NetworkReady` (or a defined retryable state) without a restart or a manual re-call — that is exactly the recovery gap in `schedule_generated_account_setup`.

docs/marmot-architecture/overview/current-state.md
```diff
@@ -134,6 +134,13 @@ account runtime for activation, KeyPackage publication, and publish confirmation
 messages, emit application-visible group events, model losing-branch invalidations, and test generated delivery
 variants.
 
+Generated-account creation now has an explicit durable local-ready boundary. The identity, default profile, setup
+journal, stable KeyPackage slot, private KeyPackage material, and exact signed initial publication are persisted before
+the caller returns. Bootstrap/profile/KeyPackage publication then continues as restart-resumable background work.
+Hosts must use `AccountSetupReadiness`: `LocalReady` is sufficient for local reads and profile rendering, while only
+`NetworkReady` means the account may be presented as invite-receivable. `Publishing` includes retryable background
```
Datawav: **Docs overstate the recovery guarantee.** Saying `Publishing` "includes retryable background publication" implies the runtime retries on its own, but there is no automatic in-session retry: a background publication failure only emits `AccountError` and drops the account from the active set, leaving it at `Publishing`/`LocalReady` until process restart (`resume_generated_account_setup` runs only from `start()`) or a manual `create_identity` re-call. Host integrations will misread `Publishing` as self-healing. Align the docs with the actual recovery mechanism (restart-resume or manual re-call), or implement the in-session retry.

docs/marmot-architecture/overview/whitenoise-integration-map.md
```diff
@@ -75,6 +75,10 @@ A first whitenoise-rs shim would need to do the following.
 
 ### Account Bootstrap
 
+- for generated identities, render the account and exact returned default profile at `LocalReady` without an immediate
+  profile lookup;
+- keep invite-receivable UI disabled until `AccountSetupReadiness::NetworkReady`; `Publishing` is durable/retryable
```
Datawav: **Same docs/behavior mismatch as `current-state.md`.** Describing `Publishing` as "durable/retryable background progress" tells hosts the runtime will converge by itself, but recovery from a failed background publish currently requires a process restart or a manual `create_identity` re-call — there is no in-session retry. Please align this guidance with the actual recovery mechanism, or track it as a known gap until the retry is implemented.

Datawav (commented):
## Automated review

Three supported findings on the exact head `7fdea03c`:

1. **High — CI red at this head while the PR body claims green.** `check_evidence.all_successful=false`: `Rust tests (1/4)-(4/4)`, `Relay runtime tests`, `CLI e2e (mock relay)`, and `Relay CLI E2E` fail; the body claims `cargo test -p marmot-app --lib` 701 passed. Consequence: the head cannot be treated as verified. Correction: re-run or reconcile the failing lanes before relying on the body’s test claims.

2. **Medium — generated-account publication is single-attempt; a transient failure parks the account at a non-terminal phase until restart.** In `crates/marmot-app/src/runtime/mod.rs` `schedule_generated_account_setup`, the spawned task calls `create_or_import_account_network_ready` once; on `Err` it only emits `AccountError(error.privacy_safe_kind())` and removes the account from `active_accounts`. `account_setup_readiness` maps `BootstrapPublicationStarted/Confirmed | KeyPackagePublicationStarted/Confirmed` to `Publishing` and `LocalReady` to `LocalReady`; `resume_generated_account_setup` runs only from `start()`. Consequence: after a transient relay/network failure, the account sits at `Publishing`/`LocalReady` indefinitely in-session — never invite-receivable — unless the process restarts or the caller manually re-invokes `create_identity`. Correction: add a bounded in-session retry or re-arm the resume path on background failure. Caveat: the inner function body is not visible in the truncated head context, so this holds if it does no intra-network retry; confidence medium.

3. **Medium — no regression test for automatic in-session convergence after a background KeyPackage publish failure.** The added tests cover success path, restart-resume, and concurrency; `ScriptedPushRelayClient::fail_publishes_of_kind` exists but the partial-generated test uses it only via an explicit manual `create_identity` re-call after `allow_all_publish_kinds()`, and the pre-existing restart test covers restart recovery only. Consequence: the gap in finding 2 is untested. Correction: add a test that fails the first KeyPackage publish in-session and asserts eventual `NetworkReady` (or a defined retryable state) without restart or manual re-call.

Low-severity (not blocking): the new docs say "`Publishing` includes retryable background publication" and "durable/retryable background progress," but the code has no automatic in-session retry — only restart-resume or manual re-call. Consequence: host integrations misread `Publishing` as self-healing. Correction: align the docs with the actual recovery mechanism or implement the retry.

erskingardner (commented):
Prior blockers are addressed at `e75a303`.

- `create_identity` / UniFFI `create_identity` wait for `NetworkReady` again; only `create_identity_local_ready` / `create_identity_with_profile` return at the local-ready boundary
- Resume parse/create failures no longer abort `start()`; missing and unreadable context emit a host-visible `AccountError`
- Persisted setup-context relays are validated and used on resume
- Background publication has bounded in-session retry, and `AccountSetupResult.readiness` now comes from the query API
- CI is green, including the `agent-connector` and CLI e2e shards that were failing on the previous head

Optional follow-up, not blocking: `AccountSetupLocalReadyHandoff` now times only `reconcile()`, not the full local-ready path. The per-stage identity/storage/profile/KeyPackage samples still exist, but hosts that wanted end-to-end local-ready latency from that field will see just the reconcile slice.

Recommend merge. GitHub will not accept `APPROVE` on the author's own PR, so this is posted as a comment review.

PR #1500, opened 2026-08-18
Streamline create-group response tail

### LINKED ISSUE #1487: marmot-app: reduce post-canonical create-group persistence and close its telemetry gap
## Problem

After the engine has canonically created and persisted a group, the foreground path still performs additional local durability/projection work before returning.

At 82af46ca1352:

- [the engine-managed group creation returns canonical outcome and durable Welcome intents](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/runtime/account_worker.rs#L2425);
- the client then performs [a separate app-level pending-Welcome index transaction](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/client/mod.rs#L994);
- it then [persists the local group projection](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/client/mod.rs#L1015);
- the engine is authoritative and the app already has [pending-Welcome reconciliation machinery](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/client/mod.rs#L3939);
- current [performance telemetry](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-app/src/app_telemetry.rs#L19) does not separately attribute the convenience-index transaction between MLS persistence and projection save.

The pure MLS benchmark is already fast enough that extra file-backed SQLCipher commits can become a meaningful share of perceived latency, especially on mobile storage. The current telemetry cannot prove which local tail dominates.

## Requested outcome

Measure the post-canonical local tail on production-shaped file-backed storage, then eliminate or consolidate redundant synchronous transactions where correctness permits.

Possible implementations include deriving/reconciling the app convenience index from authoritative engine intent, combining compatible app-level writes, or moving reconstructable work off the response path. The issue should choose the smallest design that preserves immediate host-visible consistency and crash recovery.

## Correctness constraints

- Preserve the engine transaction as the authoritative atomic group/Welcome-intent commit.
- Do not return before the new group is immediately observable through the documented local query/event surface.
- Any convenience state moved off-path must be reconstructable and reconciled after every crash boundary.
- Preserve exactly-once/effectively-once Welcome-delivery behavior and retry enumeration.
- Do not broaden transaction scope across components without analyzing lock ordering and torn-write compensation.
- Keep new timing dimensions aggregate and privacy-safe.

## Acceptance criteria

- Add separate timings for canonical MLS prepare/persist, app pending-Welcome indexing, local projection persistence, and response handoff.
- Capture repeatable p50/p95 measurements on file-backed encrypted storage for 1, 8, and 32 invitees.
- Document which duplicated state is authoritative, derived, or convenience-only.
- Reduce the number of synchronous post-canonical commits or demonstrate with evidence that consolidation is unsafe/not beneficial.
- Crash-injection tests cover every remaining write boundary and prove group visibility plus Welcome retry recovery after restart.
- Existing return-before-Welcome-fanout behavior from #1451 remains intact.

## Related work

- #1303 exposed group-create phase metrics but does not isolate this local persistence gap.
- #1451 removed relay fanout from the response path. This issue focuses only on the remaining local durability/projection tail.

## Created-group binding result

The response handoff should also eliminate the host's immediate read-after-create round trip. [The UniFFI group creation methods currently return only the group-id string](https://github.com/marmot-protocol/mdk/blob/82af46ca1352dbb3137bea434c3510e1f19842f3/crates/marmot-uniffi/src/commands/group.rs#L257), even though the foreground path has already saved the local projection before returning.

Expose a result record along these lines (exact naming is flexible):

```text
CreatedGroupFfi {
    group_id_hex: String,
    chat_list_row: ChatListRowFfi
}
```

Requirements:

- Return the same chat-list projection made durable by the canonical create path; do not issue a second storage read merely to build the response.
- The returned row and emitted/query-visible projection must agree at the response boundary.
- Preserve a compatibility path for existing callers that only consume a group-id string.
- Include response conversion/handoff in the structured phase timing.
- Do not make creation slower by synchronously computing optional group-details or management projections that are not already available. A combined conversation snapshot is tracked separately.

Acceptance additions:

- White Noise can navigate after creation without immediately calling `chat_list_row`.
- UniFFI tests prove the returned row matches the row visible through the normal query API.
- Crash/restart tests continue to prove that the returned projection was durable, not response-only synthesized state.

erskingardner:
## Summary

- remove the eager app pending-Welcome index commit from founding create and reconstruct convenience state from engine-authoritative retained Welcome obligations
- atomically persist the account projection and materialize the exact chat-list row returned by the new detailed Rust/UniFFI APIs, while preserving the existing group-id-only compatibility APIs
- add pending-index and response-handoff telemetry, file-backed SQLCipher p50/p95 evidence, durability ownership documentation, and post-canonical crash/restart coverage

## Validation

- `just fast-ci`
- `cargo test -p storage-sqlite -p marmot-uniffi`
- `cargo test -p cgka-conformance-simulator` (all non-audit suites passed; the tracing audit was rerun and passed after replacing benchmark stdout with structured tracing)
- focused atomic projection/row, durable returned/emitted row, derived Welcome-index, and feature-gated post-canonical crash/restart tests
- macOS and iOS XCFramework generation plus artifact/Swift package validation
- Kotlin bindings generation for all four Android ABIs

`cargo test -p marmot-app` passed all 688 unit tests. Its relay integration run passed 99/104 tests and exposed five unrelated failures; the isolated `encrypted_media_endpoint_updates_are_full_replacement_and_admin_only` failure reproduced unchanged on exact base `a5bbb45967286761cf8d2fb74e4cdcb73347d385` (expected one endpoint, observed three).

Fixes #1487


crates/marmot-app/src/lib.rs
```diff
@@ -4496,6 +4528,51 @@ impl MarmotApp {
...
+    fn save_state_delta_and_refresh_created_chat_list_row(
+        &self,
+        delta: &AccountState,
+        frontiers_to_clear: &[(String, u64)],
+        application_event_ids_to_ack: &[MessageId],
+        group_id_hex: &str,
+    ) -> Result<ChatListRow, AppError> {
+        let account = self.account_home().account(&delta.label)?;
+        let classifier = Self::chat_list_mention_classifier(&account.account_id_hex);
+        let refresh_all_rows = self
```
erskingardner: Don't rebuild the entire chat list on the create tail.

`CreatedGroup` only needs the created row. When `chat_list_projection_stale` is set — the normal state after any prior `save_state_delta` unless the host queried the list in between — this helper rebuilds every row inside the one remaining post-canonical SQLCipher transaction, then clears the stale flag.

That puts the expensive file-backed work #1487 is trying to remove back onto the create spinner. The documented 1/8/32-invitee p50/p95 never hit this path: each sample writes one group on a fresh store.

Two concrete problems:

1. **Latency.** A user who received messages (or otherwise saved state) while filling the new-group form will pay a full chat-list rebuild before create returns. On mobile SQLCipher that can dominate the tail this PR measured.
2. **Stale-flag correctness.** This delta can include every group in `pending_group_projection_updates`, not just the created one. Clearing `stale` after refreshing only the new row (the `refresh_all_rows == false` branch) freezes other groups' last-message/unread rows while `warmed && !stale` makes `ensure_chat_list_projection` skip the rebuild.

Refresh only the created row. If the projection was already stale, leave the marker set so the next full list query pays the rebuild — same as the old create path. Do not clear stale unless this transaction actually rebuilt every dirty row.

crates/marmot-app/src/client/mod.rs
```diff
@@ -1011,9 +1011,19 @@ impl AppClient {
         // transport delivery. Persist every exact Welcome obligation before
         // the first external side effect, so a crash between recipients
         // cannot lose the still-undelivered work or induce a second group.
+        let welcome_index_started_at = Instant::now();
+        let founding_welcome_intents =
```
erskingardner, line 1065: The comment immediately above still describes the old eager `record_founding_welcome_delivery_intents` write. This block only derives in-memory message ids; the engine already retained the exact Welcome bytes at the canonical commit.

As written, “Persist every exact Welcome obligation before the first external side effect” will send the next reader back toward the convenience-index commit this PR removed. Please rewrite it to match the new ownership table: engine rows are authoritative, this is in-memory fanout bookkeeping, and `app_pending_welcome_delivery` is populated later on failure or reconcile.

erskingardner (commented):
Request changes (posted as comment — GitHub blocks `REQUEST_CHANGES` on your own PR).

The premise is right: drop the eager founding convenience-index commit, keep the engine transaction as the point of no return, and hand the host the exact durable chat-list row from the one remaining app write. The compatibility-vs-detailed split, ownership table, and atomic projection/row test match #1487.

Please fix the create-path chat-list rebuild before merge. When the projection is already stale, this PR rebuilds every chat-list row inside the create transaction and then clears the stale flag. That puts the expensive SQLCipher work back on the spinner #1487 is trying to shrink, and the documented p50/p95 never exercise it. Refresh only the created row; leave `chat_list_projection_stale` set unless this transaction actually rebuilt every dirty row.

Also rewrite the leftover “persist every exact Welcome obligation” comment so it does not describe the write this PR removed.

Non-blocking: the UniFFI “returned row matches query” check compares `Debug` strings because `ChatListRowFfi` has no `PartialEq`. Compare the Rust `ChatListRow` (as the relay test does) or add equality on the FFI record.

crates/marmot-app/src/lib.rs
erskingardner: Fixed in cbc42879. The create transaction now refreshes only the created row. It never removes a pre-existing chat-list stale marker, and it explicitly marks the list stale when the saved delta contains another dirty group. The regression test seeds a stale marker before create and verifies it remains set while the returned row is durable.

crates/marmot-app/src/client/mod.rs
erskingardner: Fixed in cbc42879. The comment now states that the engine transaction already retained the exact Welcome bytes and destinations, this step derives only in-memory fanout ids, and the convenience index is populated later by failure or reconciliation.

Datawav (commented):
## Automated review

Empty. Kimi's independent review: "No actionable findings." I have enough to finalize. The final crystallized audit:

No actionable findings.<|close|>think<|sep|><|open|>response<|sep|>No actionable findings.

crates/marmot-app/src/lib.rs
```diff
@@ -4514,6 +4546,50 @@ impl MarmotApp {
...
+            )?
+            .ok_or_else(|| AppError::UnknownGroup(group_id_hex.to_owned()))?;
+        self.hydrate_chat_list_row(Some(&mut row))?;
+
+        // Only the created row belongs on the response tail. Preserve any
+        // pre-existing stale marker, and add one if this delta also persisted
+        // another dirty group; a later full-list query performs that rebuild.
+        self.chat_list_projection_warmed
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner())
```
erskingardner: Do not mark the full chat-list projection warmed after refreshing one row.

`ensure_chat_list_projection` treats `warmed && !stale` as “skip ensure/rebuild.” `ensure_chat_list_rows` exists specifically to rebuild when `account_groups` and `chat_list_rows` are not 1:1. After this insert, a create that happens before the first chat-list query this process can hide every other group that still lacks a row.

The stale handling below is correct. Delete this `warmed` insert. If the list was already warm, you do not need it; if it was not, the next query must still be allowed to `ensure`.

erskingardner (commented):
Re-review of `cbc42879`.

Previous findings are fixed: create now materializes only the created row, it no longer clears a pre-existing stale marker, it marks stale when the delta includes another dirty group, the Welcome-index comment matches the new ownership table, and the UniFFI test compares fields instead of `Debug` strings. The stale-preservation regression test is the right coverage.

One leftover from that fix should land before merge: this helper still inserts into `chat_list_projection_warmed` after a single-row refresh. That flag means “the full list is materialized and current.” Setting it here lets `ensure_chat_list_projection` skip `ensure_chat_list_rows` on the next query. If this process never warmed the list and other `account_groups` still lack `chat_list_rows`, those chats stay missing until something else marks stale.

Drop the `warmed.insert`. If the list was already warm, the set is a no-op; if it was not, the next query should still run `ensure`.

crates/marmot-app/src/lib.rs
erskingardner: Fixed in d24ccd0e. The single-row create refresh no longer inserts into chat_list_projection_warmed. The regression test explicitly removes the account from warmed before create, preserves the stale marker, and verifies the one-row response path does not claim the full list is warmed. The rebase also preserves the new founding-retention options through both legacy and detailed create paths.

erskingardner (commented):
Recommend merge (posted as comment — GitHub blocks `APPROVE` on your own PR).

Re-review of `d24ccd0e`. The earlier findings are resolved: create materializes only the created row, preserves a pre-existing stale marker, marks stale when the delta includes another dirty group, no longer claims the full list is warmed, and the Welcome-index comment matches the ownership table. The regression test now covers both the stale-preservation and not-warmed cases. Founding-retention options are wired through both the legacy and detailed APIs.

PR #1501, opened 2026-08-19
Add Codex terminal harness connector

erskingardner:
## Summary

- add `wn-codex` as a direct terminal-harness backend using Codex's non-interactive JSONL interface
- persist one Codex thread per Marmot group and resume subsequent prompts with `codex exec resume`
- send prompts over stdin and forward only completed assistant messages, excluding reasoning and tool events
- add isolated installer/service defaults, release bundles, CI packaging, documentation, and connector tests

## Permissions

This initial connector does not add auto-approval, sandbox-bypass, or other permission-broadening flags. Codex's existing configuration remains authoritative; common permission controls across terminal harnesses are deferred to follow-up work.

## Validation

- `just fast-ci`
- `cargo test --locked -p wn-codex`
- `just codex-dev-e2e-connector`
- `just codex-installer-test`
- `just pi-installer-test`
- shell syntax checks for connector installer and packaging scripts

The ignored live Codex contract test was not run because it requires an authenticated model request.

## Summary by CodeRabbit

* **New Features**
  * Added the `wn-codex` terminal harness for running Codex sessions, including prompt handling, session resumption, streaming responses, timeouts, and error reporting.
  * Added Codex installation support with isolated defaults, service configuration, environment settings, and manual setup options.
  * Added Codex binaries, installers, and release assets for supported Linux and macOS platforms.

* **Documentation**
  * Updated integration, setup, verification, and release documentation with Codex instructions.

* **Tests**
  * Added automated connector, installer, and end-to-end validation for Codex.

### DIFF
--- a/integrations/codex/marmot/AGENTS.md
+++ b/integrations/codex/marmot/AGENTS.md
@@ -1,0 +1,51 @@
+# AGENTS.md - integrations/codex/marmot
+
+Rust Codex harness for Marmot through the local `wn-agent` control socket. Read
+`README.md`, `../../AGENTS.md`, and `../../terminal-harness/AGENTS.md` first.
+
+## Scope
+
+- A control-plane-only harness. `wn-agent` owns Marmot state, MLS, Nostr,
+  durable sends, and invite handling; this crate invokes Codex and parses its
+  JSONL event stream.
+- Every message from an allowed sender is a prompt. Do not add gateway,
+  mention-activation, media, profile, or QUIC preview behavior here.
+- Send prompts over stdin, emit only completed `agent_message` text, and never
+  expose reasoning, command output, tool output, or partial items.
+- Preserve the shared workdir/session mapping rules. One Marmot group maps to
+  one Codex thread id and later prompts use `codex exec resume`.
+- The CLI contract is `codex exec --json -` for a new thread and
+  `codex exec resume --json <thread-id> -` for a resumed thread. Re-verify the
+  official non-interactive-mode contract before changing invocation flags or
+  JSON event parsing.
+
+## Key Files
+
+- `src/main.rs` - binary entrypoint, CLI help, tracing setup, and shared runtime wiring.
+- `src/config.rs` - Codex-specific environment configuration.
+- `src/codex.rs` - Codex command construction, stdin prompting, JSONL parsing, and process lifecycle.
+- `tests/e2e_connector.rs` - ignored process-level test using real `wn-agent` and a fake Codex executable.
+- `tests/test_installer.sh` - Codex entrypoint for the shared terminal-harness installer test suite.
+- `scripts/install-codex-marmot.sh` - release-installer wrapper over the shared terminal-harness installer.
+
+## Rules
+
+- Keep prompts on stdin; do not move prompt text into process arguments.
+- Do not add `--full-auto`, `--dangerously-bypass-approvals-and-sandbox`, or
+  connector-specific sandbox/approval overrides until the shared connector
+  permission model is designed.
+- Split completed assistant text into byte-budgeted Marmot messages. Keep
+  `WN_CODEX_MAX_REPLY_BYTES=30000` below the Marmot message cap.
+- Keep request/response and Codex event validation strict; malformed or
+  unsupported events must not become durable replies.
+
+## Verification
+
+```sh
+cargo test -p wn-codex
+cargo fmt --check -p wn-codex
+cargo clippy -p wn-codex --all-targets -- -D warnings
+bash -n scripts/install-codex-marmot.sh scripts/install-terminal-harness-marmot.sh
+just codex-dev-e2e-connector
+just codex-installer-test
+```
--- a/integrations/codex/marmot/AGENTS.md
+++ b/integrations/codex/marmot/AGENTS.md
@@ -17,7 +17,8 @@ Rust Codex harness for Marmot through the local `wn-agent` control socket. Read
 - The CLI contract is `codex exec --json -` for a new thread and
   `codex exec resume --json <thread-id> -` for a resumed thread. Re-verify the
   official non-interactive-mode contract before changing invocation flags or
-  JSON event parsing.
+  JSON event parsing. The new-thread and resume paths were verified end-to-end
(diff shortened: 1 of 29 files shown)


integrations/codex/marmot/src/codex.rs
```diff
@@ -0,0 +1,429 @@
...
+        .unwrap();
+
+        assert!(outcome.observed_session.is_some());
+        assert_eq!(outcome.exit_code, Some(0));
+        let mut reply = String::new();
+        while let Some(RunnerEvent::Text(text)) = rx.recv().await {
+            reply.push_str(&text);
+        }
+        assert_eq!(reply.trim(), "CODEX_CONNECTOR_OK");
+    }
```
erskingardner, line 455: This live contract test only exercises the new-thread path (`codex exec --json -`, `session_id: None`). The resume path (`codex exec resume --json <thread-id> -`) is the riskier invocation in this connector and currently has no real-binary coverage anywhere — only the fake-script unit test. Since the PR body notes this ignored test hasn't been run yet, consider extending it to chain a second invocation: capture `observed_session` from the first run, then resume that thread with a trivial prompt and assert a reply. That makes one command (`cargo test -p wn-codex real_codex_exec_contract -- --ignored --nocapture`) validate the full per-group session lifecycle the connector depends on.

integrations/codex/marmot/AGENTS.md
```diff
@@ -0,0 +1,51 @@
...
+- Every message from an allowed sender is a prompt. Do not add gateway,
+  mention-activation, media, profile, or QUIC preview behavior here.
+- Send prompts over stdin, emit only completed `agent_message` text, and never
+  expose reasoning, command output, tool output, or partial items.
+- Preserve the shared workdir/session mapping rules. One Marmot group maps to
+  one Codex thread id and later prompts use `codex exec resume`.
+- The CLI contract is `codex exec --json -` for a new thread and
+  `codex exec resume --json <thread-id> -` for a resumed thread. Re-verify the
+  official non-interactive-mode contract before changing invocation flags or
+  JSON event parsing.
```
erskingardner: Pi's AGENTS.md sets the repo precedent of pinning the verified backend version ("The CLI contract was verified end-to-end against Pi `0.79.6`"). Once the ignored live contract test has actually been run, record the verified `codex-cli` version here the same way, so future contributors know which upstream release this contract statement was validated against.

erskingardner (commented):
Reviewed at 351928be. This is a clean, well-scoped addition: it mirrors the wn-opencode/wn-pi pattern faithfully, updates every enumeration surface I could find (AGENTS maps, READMEs, release.md, packaging script, CI workflow, Justfile, changelog), keeps tracing privacy-safe per the observability doc, and ships an installer test wired into the shared harness suite. CI is green.

I verified the Codex CLI contract against current upstream source rather than trusting the mocks: in openai/codex, `codex exec`'s `--json` flag is declared `global = true`, so `codex exec resume --json <id> -` parses with the flag after the subcommand, and `ResumeArgs.prompt` documents `-` as "read from stdin". The JSONL event names the parser accepts (`thread.started`, `item.completed` with `item.type == "agent_message"`, `turn.failed`, `error`) match the documented non-interactive-mode event stream, and squashing error events to constants keeps backend messages out of durable replies. The docs links in the README resolve to the real OpenAI Learn pages.

Two non-blocking follow-ups inline: the live contract gap (self-acknowledged in the PR body) leaves the resume path with zero real-binary coverage, and unlike Pi this AGENTS.md doesn't record a verified backend version. I'd resolve both before undrafting.

integrations/terminal-harness/tests/test_installer.sh
```diff
@@ -1,8 +1,16 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-kind="${1:?usage: test_installer.sh pi|opencode}"
+kind="${1:?usage: test_installer.sh codex|pi|opencode}"
 case "$kind" in
+    codex)
```
erskingardner, line 6: This new `codex` case is the right installer fixture, but `.github/workflows/ci.yml` `terminal-harness-installers` still only runs `pi` and `opencode`.

Add:
```sh
integrations/terminal-harness/tests/test_installer.sh codex
```
to that job (and rename the step from "Test Pi and OpenCode installers"). That job is the only PR CI gate for these installers; without it, a Codex installer regression will not fail CI.

erskingardner (commented):
Would request changes, but GitHub does not allow that on your own PR.

The connector itself looks like the right shape: isolated `wn-codex` home/services, stdin prompts, completed `agent_message` only, and the `exec` / `exec resume` JSONL contract matches current Codex CLI (`--json` is global, `-` is stdin). Packaging and docs follow the existing terminal-harness pattern.

One merge blocker: the dedicated installer CI job was not updated, so the new production installer is not gated the way Pi and OpenCode are.

Non-blocking: a first group prompt without `/<path>` still runs Codex in `$HOME`. Codex rejects non-git cwds by default, so that path will usually fail until the user picks a repo. Worth either documenting more sharply or failing closed in the Codex backend with a harness-authored error.

Also residual: the live Codex contract test was not run, and headless approval/sandbox behavior is still entirely on the user's Codex config.

erskingardner (commented):
Recommend merge. Would approve, but GitHub does not allow that on your own PR.

The previous merge blocker is fixed: `terminal-harness-installers` now runs Codex, and that job is green. The rest of the follow-up is in good shape too — resume coverage in the fake e2e and ignored live contract test, installer SHA256 assets, and the `$HOME`/non-Git first-prompt behavior is now documented.

One leftover consistency nit: `AGENTS.md` says the new-thread and resume paths were verified against `codex-cli 0.146.0`, but the PR validation section still says the live contract test was not run. If you did run it on 0.146.0, update the PR body; if not, soften the AGENTS.md claim.

integrations/terminal-harness/tests/test_installer.sh
erskingardner: Fixed in 83e78a5: `.github/workflows/ci.yml` now runs the Codex installer test, and `Terminal harness installer tests` is green on this head.

PR #1505, opened 2026-08-19
Extract shared JSONL subprocess runner for terminal harnesses

### LINKED ISSUE #1503: Extract a shared JSONL subprocess runner for terminal harness connectors
## Summary

Extract the duplicated subprocess lifecycle used by `wn-pi`, `wn-opencode`, and `wn-codex` into `marmot-terminal-harness` without changing connector behavior.

This is follow-up work to #1501 and should remain a behavior-preserving refactor, separate from permission-policy changes.

## Background

The shared crate already owns helpers for bounded stderr capture, stdin writing, idle deadlines, child cleanup, and ANSI stripping. Each backend still duplicates the larger lifecycle:

- construct and spawn a child;
- configure stdin/stdout/stderr;
- apply total and idle deadlines;
- drain bounded stderr;
- parse JSONL events;
- capture the first observed session id and error classification;
- apply bounded reply-channel backpressure;
- terminate and reap on failure.

The three copies are already close enough to drift while still differing in prompt transport and event schema.

## Proposed scope

- Add a typed shared process specification, for example:
  - executable and argv;
  - working directory;
  - prompt transport: stdin or an explicitly delimited argument;
  - privacy-safe tracing method/backend name;
  - total and idle timeouts.
- Add a shared parsed-event vocabulary:
  - session observed;
  - completed assistant text;
  - sanitized backend error classification;
  - ignored event.
- Move the common spawn/read/wait/cleanup/outcome lifecycle into the shared crate.
- Keep backend-specific command construction and strict JSON event decoding in each backend crate.
- Preserve these existing semantics:
  - only completed assistant text is emitted;
  - thinking, tool calls/output, partial text, usage, and raw backend errors are not forwarded;
  - the first valid observed session id wins;
  - stdout reads reset the idle deadline;
  - reply-channel backpressure is covered by the total deadline rather than misclassified as backend idleness;
  - stderr remains bounded and ANSI-stripped;
  - child processes are killed and reaped on all timeout/error paths;
  - broken stdin pipes remain handled as they are today.
- Keep command construction independently unit-testable.

## Acceptance criteria

- Shared lifecycle tests cover:
  - stdin and no-stdin transports;
  - normal completion;
  - malformed/ignored events;
  - bounded channel backpressure;
  - total timeout;
  - idle timeout during output and final process wait;
  - early stdin closure;
  - bounded stderr capture;
  - child termination/reaping.
- Pi, OpenCode, and Codex parser and argv tests remain backend-local.
- Existing fake-binary process tests pass unchanged or with equivalent assertions.
- Real backend contract tests remain ignored and available.
- No permission defaults, installer behavior, session mapping, or user-visible replies change.
- `cargo test -p marmot-terminal-harness`, all three connector crate tests, and `just fast-ci` pass.

## Non-goals

- Introducing a universal backend protocol.
- Migrating OpenCode to ACP or Pi to RPC.
- Adding streaming partial replies.
- Changing session lifetime or queueing semantics.
- Adding permission profiles in the same PR.

erskingardner:
## Summary
- add a typed shared process specification, prompt transports, and parsed-event vocabulary
- centralize JSONL spawn, deadlines, backpressure, bounded stderr, session/error capture, and child cleanup
- keep Pi, OpenCode, and Codex argv construction and strict event decoding backend-local
- preserve the Codex resume coverage and review fixes merged in #1501

## Test plan
- [x] New shared-runner test failed before the API existed and passes after the fix
- [x] `cargo test -p marmot-terminal-harness -p wn-opencode -p wn-pi -p wn-codex`
- [x] `just codex-dev-e2e-connector opencode-dev-e2e-connector pi-dev-e2e-connector`
- [x] `just fast-ci`
- [ ] Real Codex and Pi contract tests remain ignored because they require authenticated model requests

Fixes #1503

### DIFF
--- a/integrations/terminal-harness/tests/jsonl_process.rs
+++ b/integrations/terminal-harness/tests/jsonl_process.rs
@@ -203,6 +203,8 @@ printf '%s\n' '{"type":"text","text":"completed"}'
 
 #[tokio::test]
 async fn stdout_reads_reset_the_idle_deadline() {
+    const IDLE_TIMEOUT: Duration = Duration::from_millis(600);
+
     let _permit = process_test_permit().await;
     let root = tempfile::tempdir().unwrap();
     let script = executable_script(
@@ -210,17 +212,21 @@ async fn stdout_reads_reset_the_idle_deadline() {
         "progress-backend",
         r#"#!/bin/sh
 printf '%s\n' '{"type":"session","id":"progress"}'
-sleep 0.1
+sleep 0.25
 printf '%s\n' '{"type":"progress"}'
-sleep 0.1
+sleep 0.25
+printf '%s\n' '{"type":"progress"}'
+sleep 0.25
 printf '%s\n' '{"type":"text","text":"done"}'
 "#,
     );
     let (tx, mut rx) = mpsc::channel(2);
     let mut spec = process_spec(&script, root.path(), PromptTransport::Stdin(String::new()));
-    spec.idle_timeout = Duration::from_millis(500);
+    spec.idle_timeout = IDLE_TIMEOUT;
+    let started = std::time::Instant::now();
     let outcome = run_jsonl_process(spec, tx, parse_event).await.unwrap();
 
+    assert!(started.elapsed() > IDLE_TIMEOUT);
     assert_eq!(outcome.observed_session.as_deref(), Some("progress"));
     assert_eq!(rx.recv().await, Some(RunnerEvent::Text("done".to_owned())));
 }
--- a/integrations/terminal-harness/tests/jsonl_process.rs
+++ b/integrations/terminal-harness/tests/jsonl_process.rs
@@ -1,0 +1,442 @@
+#![cfg(unix)]
+
+use std::fs;
+use std::os::unix::fs::PermissionsExt;
+use std::sync::{Arc, OnceLock};
+use std::time::Duration;
+
+use marmot_terminal_harness::{
+    ParsedEvent, PromptTransport, RunnerEvent,
+    process::{EnvironmentChange, ProcessSpec, run_jsonl_process},
+};
+use serde_json::Value;
+use tokio::sync::{OwnedSemaphorePermit, Semaphore, mpsc};
+
+async fn process_test_permit() -> OwnedSemaphorePermit {
+    static PROCESS_TESTS: OnceLock<Arc<Semaphore>> = OnceLock::new();
+    PROCESS_TESTS
+        .get_or_init(|| Arc::new(Semaphore::new(1)))
+        .clone()
+        .acquire_owned()
+        .await
+        .unwrap()
+}
+
+fn executable_script(root: &std::path::Path, name: &str, body: &str) -> std::path::PathBuf {
+    let path = root.join(name);
+    fs::write(&path, body).unwrap();
+    let mut permissions = fs::metadata(&path).unwrap().permissions();
+    permissions.set_mode(0o755);
+    fs::set_permissions(&path, permissions).unwrap();
+    path
+}
+
+fn parse_event(line: &str) -> serde_json::Result<ParsedEvent> {
+    let value: Value = serde_json::from_str(line)?;
+    Ok(match value.get("type").and_then(Value::as_str) {
+        Some("session") => value["id"]
+            .as_str()
+            .map(|id| ParsedEvent::Session(id.to_owned()))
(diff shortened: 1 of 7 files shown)


integrations/terminal-harness/tests/jsonl_process.rs
```diff
@@ -0,0 +1,398 @@
...
+sleep 0.1
+printf '%s\n' '{"type":"progress"}'
+sleep 0.1
+printf '%s\n' '{"type":"text","text":"done"}'
+"#,
+    );
+    let (tx, mut rx) = mpsc::channel(2);
+    let mut spec = process_spec(&script, root.path(), PromptTransport::Stdin(String::new()));
+    spec.idle_timeout = Duration::from_millis(500);
+    let outcome = run_jsonl_process(spec, tx, parse_event).await.unwrap();
```
erskingardner, line 227: This does not test idle-deadline reset.

Two `sleep 0.1` gaps against a 500ms idle budget stay well under one initial deadline, so the test passes even if `idle_deadline` is never refreshed. #1503 listed “idle timeout during output” as acceptance criteria, and the old OpenCode paused-clock test (`stdout_lines_reset_idle_past_old_wall_clock_deadline`, 70s gaps / 120s idle) actually proved it.

Restore that style here: `tokio::time` pause, or real sleeps where each gap is larger than `idle_timeout` and the sum of gaps is larger than one idle budget. Otherwise a regression that only resets idle on the first line will go unnoticed.

integrations/terminal-harness/tests/test_installer.sh
```diff
@@ -1,8 +1,16 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-kind="${1:?usage: test_installer.sh pi|opencode}"
+kind="${1:?usage: test_installer.sh codex|pi|opencode}"
 case "$kind" in
+    codex)
```
erskingardner, line 6: The new `codex` fixture is not exercised by PR CI. `.github/workflows/ci.yml` `terminal-harness-installers` still only runs `pi` and `opencode`, so the job that just passed on this PR never touched the Codex installer.

That job is the only PR gate for these scripts. Add `integrations/terminal-harness/tests/test_installer.sh codex` (and rename the step) before this stacked tree is mergeable — same blocker as #1501. If this PR is meant to rebase onto #1501 and drop the Codex commit, this comment can move with that fix.

erskingardner (commented):
The extract is the right shape: one `ProcessSpec` / `PromptTransport` / `ParsedEvent` boundary, lifecycle centralized, argv and strict decoding left in the backends. Pi/OpenCode semantics look preserved — completed text only, first valid session wins, idle measured on stdout reads, backpressure under the total deadline, bounded ANSI-stripped stderr, kill+reap on failure. Debug/tracing stay privacy-safe.

Two things before this leaves draft:

1. Idle-reset coverage is vacuous. The paused-clock OpenCode test was deleted; the replacement does not prove the invariant #1503 asked for.
2. This stacked tree still ships `wn-codex` without installer CI. `terminal-harness-installers` still only runs `pi` and `opencode`. Same merge blocker as #1501. Keep this draft until #1501 lands and the diff narrows to the extract, or update that job here if this remains the Codex vehicle.

I would not merge this tree as-is. The runner extract itself looks ready once the idle test is real.

integrations/terminal-harness/tests/jsonl_process.rs
erskingardner: Fixed in 0e6e6c74. The shared-runner test now emits output across three 250 ms gaps with a 600 ms idle timeout. Each gap is within the refreshed idle budget, while the full run exceeds the original deadline; it also asserts elapsed time exceeded that original budget. The focused test passed three consecutive runs, the full marmot-terminal-harness suite passes, and just fast-ci passes.

erskingardner (commented):
Previous findings are addressed. #1501 has landed, so this is now the extract-only diff, and the idle-reset test actually proves a refresh: three 250ms gaps against a 600ms budget would miss the last line if the deadline were set once.

The runner matches current backend behavior: stdin prompts, OpenCode env overlay via `EnvironmentChange`, profile flags remaining backend-local, first session / first error / completed text only, backpressure under the total deadline, bounded ANSI-stripped stderr, kill+reap on failure. CI is green.

GitHub will not let me formally approve my own PR; from a review standpoint this is mergeable.

PR #1506, opened 2026-08-19
Add explicit harness session reset and pipe OpenCode prompts

### LINKED ISSUE #1502: Add explicit session reset and remove OpenCode prompts from process arguments
## Summary

Improve terminal-harness recovery and prompt privacy by:

1. adding an explicit per-group session reset; and
2. investigating how to keep OpenCode prompt plaintext out of process arguments, then implementing a transport change only if the investigation identifies a suitable contract.

ACP is one candidate for the OpenCode transport work. This issue does **not** assume that OpenCode should move to ACP.

This is follow-up work to #1501. It should build on the existing shared group-to-session/workdir mapping without changing the Marmot control protocol.

## Background

Each Marmot group maps durably to one backend session id and working directory. Every inbound prompt starts a fresh backend process and resumes that logical session; the connector does not keep one process alive per group.

Today the session store supports get/set but no reset. A stale, deleted, or incompatible backend session can therefore leave a group unable to recover without manually editing connector state. Automatically retrying a failed resumed prompt as a new session is unsafe because the original invocation may already have produced side effects.

OpenCode currently receives prompt plaintext as an argv item after a `--` delimiter. The delimiter prevents flag injection, but the prompt can still be exposed through local process inspection. Pi and Codex already receive prompts over stdin.

The current `opencode run --format json` integration is simple, process-isolated, and already contract-tested. OpenCode also exposes an ACP server over stdin/stdout, which may solve the argv exposure and may offer useful session or permission behavior, but adopting ACP would introduce a stateful protocol and different lifecycle obligations. ACP must therefore be evaluated rather than treated as a predetermined migration.

Any OpenCode transport remains below the existing `Backend` trait. It does not replace `marmot.agent-control.v2` and does not imply that Pi or Codex should adopt ACP.

### Explicit session reset

- Add an atomic/remove operation to the shared private session store.
- Add a deliberate allowed-sender control command, such as `/reset-session`, that:
  - is intercepted by the harness and never forwarded to the backend;
  - removes only the current Marmot group's backend session id;
  - preserves the validated working directory;
  - does not delete backend transcript/session files;
  - returns a concise confirmation.
- The next normal prompt creates a new backend session in the preserved workdir.
- Do not automatically retry failed resumed prompts as new sessions.
- Document command-name collision and escaping behavior.

### OpenCode transport decision

Start with a bounded contract spike. Compare at least:

- retaining `opencode run --format json`;
- native `opencode acp`; and
- any other documented stdin or local API mode available in the supported OpenCode version.

ACP is eligible only if the verified contract supports:

- durable session creation and resume semantics compatible with one Marmot group per OpenCode session;
- completed assistant-message events that can be distinguished strictly from reasoning, partial text, tool calls, and tool output;
- headless permission behavior compatible with the shared execution profiles from #1504;
- reliable cancellation, timeout, EOF, malformed-frame, child-exit, and crash-recovery behavior;
- a stable, supportable protocol and minimum-version boundary.

Record the evidence and decision before production migration work begins.

If ACP or another transport satisfies the contract:

- replace prompt-as-argv execution with the selected stdin/stdout or equivalent local transport;
- keep transport-specific lifecycle and framing below the existing shared `Backend` boundary;
- preserve per-group serialization, bounded queues, timeouts, process cleanup, and durable session mapping;
- preserve strict filtering so only completed assistant text becomes a durable Marmot reply;
- keep execution-profile behavior compatible with the selected transport.

If no candidate satisfies the contract:

- retain `opencode run` rather than forcing ACP;
- document the remaining process-argument exposure and the evaluated alternatives;
- rescope or defer the prompt-transport implementation explicitly.

### Session reset

- Session-store removal is restrictive, torn-write-safe, and covered by persistence/reload tests.
- Reset affects only the requesting group and preserves its workdir.
- Reset commands never reach Pi, OpenCode, or Codex.
- A post-reset prompt records a newly observed session id.
- No automatic fallback can duplicate a partially executed resumed prompt.

### OpenCode transport

- The issue or implementation PR contains an evidence-backed contract comparison and explicit transport decision.
- ACP is not selected solely because the command exists.
- Keeping `opencode run` is an acceptable result when ACP and other candidates do not meet the required contract.
- If a new transport is adopted:
  - prompt text is absent from spawned process arguments and privacy-safe logs;
  - create/resume, completed-text filtering, permissions, malformed input, timeout, cancellation, child-exit, and cleanup behavior are tested;
  - the supported OpenCode version boundary is documented.
- Documentation clearly distinguishes a logical long-running group session from a process-per-message or persistent backend transport.
- Shared and connector-specific tests plus `just fast-ci` pass for any implementation changes.

## Non-goals

- Predetermining ACP as the OpenCode transport.
- Replacing `marmot.agent-control.v2` with ACP.
- Requiring Pi or Codex to adopt ACP.
- Deleting backend-owned transcript/session files.
- Automatic stale-session recovery after an invocation has begun.
- Forwarding interactive approval dialogs over Marmot in this issue.

erskingardner:
## Summary
- add an exact `/reset-session` harness command that atomically clears only the current group backend session id while preserving its validated workdir
- keep reset commands out of Codex, OpenCode, and Pi; preserve backend transcript files; and support `//reset-session` as the literal prompt escape
- retain `opencode run --format json`, but write prompts over stdin so plaintext is absent from spawned argv
- document the shared reset behavior, OpenCode 1.18.18 minimum, and bounded transport decision

## OpenCode transport decision

The selected contract is process-per-message `opencode run --format json` with `--session` for resume and prompt text on stdin. OpenCode 1.18.18 explicitly reads piped stdin when no message argument is present in its [tagged run implementation](https://github.com/anomalyco/opencode/blob/v1.18.18/packages/opencode/src/cli/cmd/run.ts#L416-L422). This preserves the existing completed-text filter, timeouts, cancellation, cleanup, crash isolation, and future `run` permission flags.

ACP was not selected. Although [`opencode acp`](https://opencode.ai/docs/acp/) moves prompts to stdio, it adds a stateful JSON-RPC lifecycle plus initialize/load/prompt/cancel and permission-request handling without improving this harness once `run` accepts stdin. [`opencode serve`](https://opencode.ai/docs/server/) was also not selected because its local OpenAPI interface adds HTTP readiness, authentication, port ownership, event-stream, and crash-recovery obligations.

This decision stays below the shared `Backend` trait and does not change `marmot.agent-control.v2`.

## Test plan
- [x] New argv regression test failed before the fix and passes after
- [x] Reset parser/store tests failed before the fix and pass after
- [x] `cargo test -p marmot-terminal-harness` (44 passed)
- [x] `cargo test -p wn-codex` (9 passed; authenticated live-model test ignored)
- [x] `cargo test -p wn-opencode` (14 passed; connector E2E ignored by default)
- [x] `cargo test -p wn-pi` (10 passed; authenticated live-model test ignored)
- [x] `just opencode-dev-e2e-connector`
- [x] `just fast-ci`

Fixes #1502

### DIFF
--- a/integrations/terminal-harness/README.md
+++ b/integrations/terminal-harness/README.md
@@ -28,9 +28,8 @@ Connector crates are responsible for:
 - exposing only completed assistant output, never thinking or tool output;
 - preserving or reporting the backend session id needed for the next prompt.
 
-OpenCode keeps prompt text after its command's `--` delimiter. Codex and Pi
-write prompt text to stdin. Backend-specific behavior belongs in those connector crates, not
-in this shared runtime.
+Codex, OpenCode, and Pi write prompt text to stdin. Backend-specific behavior
+belongs in those connector crates, not in this shared runtime.
 
 ## Shared Behavior
 
@@ -43,10 +42,18 @@ All connectors:
 - use `/<path>` on the first group message to select a canonical working
   directory beneath `$HOME`;
 - persist private per-group working-directory and session mappings;
+- intercept an exact `/reset-session` message before backend invocation, clear
+  only that group's backend session id, and retain its selected workdir;
 - split durable replies on UTF-8 boundaries with a 30,000-byte default and a
   60,000-byte hard ceiling;
 - enforce bounded per-group queues plus idle and total backend timeouts;
 - keep diagnostics free of identifiers, paths, prompts, and backend output.
+
+`/reset-session` is reserved as a harness control message. Send
+`//reset-session` to forward the literal `/reset-session` text to the backend.
+Reset never deletes backend-owned transcripts and never retries a failed resumed
+prompt automatically; the next distinct prompt starts a new logical backend
+session in the retained workdir.
 
 The connector READMEs document their environment variables, installer topology,
 and backend contracts:
--- a/integrations/codex/marmot/README.md
+++ b/integrations/codex/marmot/README.md
@@ -101,6 +101,12 @@ starting Codex. Subsequent prompts resume that group's Codex thread. Without a
 picker, the shared harness uses `$HOME`; Codex rejects a non-Git working
 directory by default, so select a repository before sending the first prompt.
 
+Send exactly `/reset-session` to clear the current group's Codex thread id while
+preserving its selected workdir and Codex-owned transcript files. The next
+normal prompt starts and records a new Codex thread; a failed resumed prompt is
+never retried automatically. Send `//reset-session` to pass the literal
+`/reset-session` text to Codex instead of invoking the harness command.
+
 ## Configuration
 
 | Environment variable | Default | Meaning |
--- a/integrations/opencode/marmot/AGENTS.md
+++ b/integrations/opencode/marmot/AGENTS.md
@@ -31,7 +31,9 @@ Read `README.md`, `../../AGENTS.md`, and `../../terminal-harness/AGENTS.md` firs
 - Keep shared control, queueing, chunking, workdir, session-map, and process lifecycle behavior in
   `integrations/terminal-harness`; keep only OpenCode command/event behavior here.
 - Keep state files restrictive-by-construction through `fs-private` or an equivalent mode-tested path.
-- Keep opencode prompts after a `--` delimiter so prompt text cannot be parsed as `opencode run` flags.
+- Keep opencode prompts on stdin and out of process arguments. The supported
+  minimum OpenCode version is 1.18.18, whose non-interactive `run` path consumes
+  piped stdin when no message argument is present.
 
 ## Verification
 
--- a/integrations/opencode/marmot/README.md
+++ b/integrations/opencode/marmot/README.md
@@ -18,8 +18,8 @@ GitHub pre-releases.
 
 Prerequisites:
 
-- OpenCode installed locally and runnable on `PATH`, or an executable path set
-  with `WN_OPENCODE_BIN` / `--opencode-bin`
+- OpenCode 1.18.18 or newer installed locally and runnable on `PATH`, or an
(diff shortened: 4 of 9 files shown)


integrations/terminal-harness/README.md
```diff
@@ -42,11 +41,19 @@ Both connectors:
...
 - split durable replies on UTF-8 boundaries with a 30,000-byte default and a
   60,000-byte hard ceiling;
 - enforce bounded per-group queues plus idle and total backend timeouts;
 - keep diagnostics free of identifiers, paths, prompts, and backend output.
 
+`/reset-session` is reserved as a harness control message. Send
+`//reset-session` to forward the literal `/reset-session` text to the backend.
+Reset never deletes backend-owned transcripts and never retries a failed resumed
+prompt automatically; the next distinct prompt starts a new logical backend
+session in the retained workdir.
```
erskingardner, line 56: Non-blocking: this reserved-command note is the right shared description, but `integrations/pi/marmot/README.md` still only says subsequent prompts resume the Pi session. `wn-pi` users who never open this crate README will not see `/reset-session` or the `//reset-session` escape. Worth a short matching paragraph there, since the intercept lives in this shared runtime.

erskingardner (commented):
Recommend merge. Matches #1502. The shared `/reset-session` path is exact, sender-gated, serialized per group, and reuses the existing empty-`session_id` workdir record so the next prompt starts a new backend session without deleting transcripts or auto-retrying a failed resume. Keeping `opencode run --format json` and moving the prompt to stdin is the right transport call: 1.18.18 reads `Bun.stdin.text()` when stdin is not a TTY and no message argument is present, so `write_stdin` plus shutdown matches the contract without taking on ACP or `serve`.

Could not use GitHub’s Approve event on this PR (self-review). One non-blocking docs gap below.

PR #1507, opened 2026-08-19
Add shared execution profiles for terminal harness connectors

### LINKED ISSUE #1504: Add shared execution profiles for terminal harness connectors
## Summary

Add a common, typed execution-permission policy for the Pi, OpenCode, and Codex terminal harness connectors, with explicit backend-specific mappings and installer support.

This is follow-up work to #1501. The initial Codex connector intentionally inherits the operator's existing Codex configuration, matching the existing Pi and OpenCode connectors. That leaves headless approval behavior and unrestricted execution inconsistent across the three backends.

## Background

The backends expose materially different controls:

- Pi has no built-in approval prompts. Tool access can be limited, but tools otherwise execute with the Pi process's OS privileges.
- OpenCode has `allow` / `ask` / `deny` permissions. `--auto` approves requests that would otherwise ask but preserves explicit denies.
- Codex separates approval policy, sandbox policy, and network access; it also exposes an explicit sandbox/approval bypass.

A common `yolo=true` boolean would hide these differences and could imply guarantees a backend cannot provide.

## Proposed scope

- Add a shared typed execution profile in `marmot-terminal-harness`, exposed as:
  - `MARMOT_HARNESS_EXECUTION_PROFILE=inherit`
  - `MARMOT_HARNESS_EXECUTION_PROFILE=autonomous`
  - `MARMOT_HARNESS_EXECUTION_PROFILE=unrestricted`
- Define the profiles in terms of operator intent:
  - `inherit`: add no connector-side overrides.
  - `autonomous`: do not wait for interactive approval while preserving configured hard denies and isolation where supported.
  - `unrestricted`: request the backend's broadest non-interactive execution mode; this must be clearly documented as requiring external isolation.
- Implement and test backend adapters rather than treating the mappings as identical:
  - Pi: document that approval-free execution is native and that Pi does not provide an OS sandbox.
  - OpenCode: use `--auto` for autonomous execution; use a connector-owned permission overlay for unrestricted execution without mutating global config.
  - Codex: use approval `never` while preserving configured sandbox/network for autonomous execution; use `--dangerously-bypass-approvals-and-sandbox` for unrestricted execution.
- Keep permission argv construction typed. Do not add a free-form backend `EXTRA_ARGS` setting.
- Add shared installer flags/config writing for the selected profile.
- Require an explicit installer acknowledgement before writing `unrestricted`.
- Emit privacy-safe startup diagnostics containing only the selected profile and capability/support state.
- Fail closed if a requested profile cannot be represented by the installed backend version.

## Security documentation

Document that:

- An allowed Marmot sender combined with unrestricted execution is effectively remote code execution as the harness service user.
- The `/<path>` workdir picker selects a working directory; it is not an OS containment boundary.
- Unrestricted deployments should use a dedicated OS user, container, or VM with narrowly scoped credentials.
- Connector-owned policy overlays must not rewrite the operator's global Pi, OpenCode, or Codex configuration.

## Acceptance criteria

- Shared configuration parsing and redacted `Debug` coverage.
- Exact command-argument tests for new and resumed sessions on all three backends.
- OpenCode prompt delimiter remains effective when permission flags are present.
- Codex new and resumed invocations receive equivalent policy.
- Installer dry-run and service/env-file tests cover every profile.
- README and architecture documentation include the capability matrix and security boundary.
- Targeted harness tests and `just fast-ci` pass.

## Non-goals

- Interactive approval forwarding over Marmot.
- Normalizing model/provider configuration.
- Switching any backend to ACP or another long-running protocol.
- Treating backend logical permission checks as equivalent to an OS sandbox.

erskingardner:
Depends on #1501. The final commit in this branch implements #1504; the two prerequisite commits are the current #1501 head and will disappear from this PR diff when #1501 lands.

## Summary
- add typed `inherit`, `autonomous`, and `unrestricted` execution profiles with privacy-safe capability diagnostics
- map Pi, OpenCode, and Codex to their distinct approval and isolation controls without global config mutation
- add installer persistence, explicit unrestricted acknowledgement, backend/installer regressions, and security documentation

## Test plan
- [x] New tests failed before the fix and pass after
- [x] `cargo test -p marmot-terminal-harness`
- [x] `cargo test -p wn-codex`
- [x] `cargo test -p wn-opencode`
- [x] `cargo test -p wn-pi`
- [x] `just codex-installer-test`
- [x] `just opencode-installer-test`
- [x] `just pi-installer-test`
- [x] `just codex-dev-e2e-connector`
- [x] `just opencode-dev-e2e-connector`
- [x] `just pi-dev-e2e-connector`
- [x] `just fast-ci`

Fixes #1504

## Summary by CodeRabbit

- **New Features**
  - Added configurable terminal execution profiles: inherit, autonomous, and unrestricted.
  - Profiles now provide consistent behavior across Codex, OpenCode, and Pi integrations.
  - Added installer options and environment configuration for selecting profiles.
  - Unrestricted mode requires explicit acknowledgement and an external OS isolation boundary.
  - Startup diagnostics report execution and capability states using safe, limited values.

- **Documentation**
  - Added architecture, configuration, capability, security, and troubleshooting guidance.

- **Tests**
  - Expanded coverage for profile validation, propagation, installer behavior, and acknowledgement requirements.

### DIFF
--- a/docs/marmot-architecture/overview/terminal-harness-execution-profiles.md
+++ b/docs/marmot-architecture/overview/terminal-harness-execution-profiles.md
@@ -32,7 +32,7 @@ permissions, process sandboxing, and network policy are different controls.
 | --- | --- | --- | --- |
 | Pi | Non-interactive execution is natively approval-free for every profile | No built-in OS sandbox | No profile-specific CLI feature is required |
 | OpenCode | `autonomous` uses `--auto`; `unrestricted` adds a process-local allow-all permission overlay | Logical permissions only; no OS sandbox | An unsupported `--auto` fails the invocation; no successful reply is sent |
-| Codex | `autonomous` sets `approval_policy="never"`; `unrestricted` bypasses approvals | Configured sandbox/network survive `autonomous`; `unrestricted` bypasses the sandbox | Strict config/unknown flags fail the invocation; no successful reply is sent |
+| Codex | `autonomous` sets `approval_policy="never"`; `unrestricted` bypasses approvals | Configured sandbox/network survive `autonomous`; `unrestricted` bypasses the sandbox | An unsupported invocation-local config override or flag fails the invocation; no successful reply is sent |
 
 The OpenCode overlay exists only in the child process environment. The Codex
 override exists only in that invocation. Connector setup never rewrites global
--- a/docs/marmot-architecture/overview/terminal-harness-execution-profiles.md
+++ b/docs/marmot-architecture/overview/terminal-harness-execution-profiles.md
@@ -45,7 +45,8 @@ file and same-user service definition. It requires
 `--acknowledge-unrestricted` before writing `unrestricted`; `--yes`, dry-run,
 or an environment-provided profile do not substitute for that acknowledgement.
 
-Startup tracing contains only the profile and fixed capability state names.
+Startup tracing contains the profile and fixed capability state names alongside
+coarse fields such as the harness name, sender count, and reply-size limit.
 Unsupported or malformed profiles fail configuration loading.
 
 ## Security Boundary
--- a/docs/marmot-architecture/overview/terminal-harness-execution-profiles.md
+++ b/docs/marmot-architecture/overview/terminal-harness-execution-profiles.md
@@ -1,0 +1,61 @@
+---
+title: "Terminal Harness Execution Profiles"
+created: 2026-08-19
+updated: 2026-08-19
+tags: [marmot, overview, terminal-harness, permissions, sandbox]
+status: overview
+---
+
+# Terminal Harness Execution Profiles
+
+Marmot terminal harnesses accept remote prompts from explicitly allowed
+senders, then invoke a local coding-agent backend. A shared profile expresses
+operator intent while backend adapters preserve the real differences between
+Pi, OpenCode, and Codex.
+
+## Shared Intent
+
+`MARMOT_HARNESS_EXECUTION_PROFILE` has three values:
+
+- `inherit`: add no connector-side policy overrides.
+- `autonomous`: do not wait for interactive approvals, while preserving hard
+  denies and isolation where supported.
+- `unrestricted`: request the broadest non-interactive backend mode. External
+  isolation is mandatory.
+
+This is deliberately not a `yolo` boolean. Approval prompts, logical tool
+permissions, process sandboxing, and network policy are different controls.
+
+## Capability Matrix
+
+| Backend | Approval mapping | Isolation mapping | Version failure behavior |
+| --- | --- | --- | --- |
+| Pi | Non-interactive execution is natively approval-free for every profile | No built-in OS sandbox | No profile-specific CLI feature is required |
+| OpenCode | `autonomous` uses `--auto`; `unrestricted` adds a process-local allow-all permission overlay | Logical permissions only; no OS sandbox | An unsupported `--auto` fails the invocation; no successful reply is sent |
+| Codex | `autonomous` sets `approval_policy="never"`; `unrestricted` bypasses approvals | Configured sandbox/network survive `autonomous`; `unrestricted` bypasses the sandbox | Strict config/unknown flags fail the invocation; no successful reply is sent |
+
+The OpenCode overlay exists only in the child process environment. The Codex
+override exists only in that invocation. Connector setup never rewrites global
+Pi, OpenCode, or Codex configuration.
+
+## Installer Contract
+
+The shared installer writes the selected profile into the private harness env
+file and same-user service definition. It requires
+`--acknowledge-unrestricted` before writing `unrestricted`; `--yes`, dry-run,
+or an environment-provided profile do not substitute for that acknowledgement.
+
+Startup tracing contains only the profile and fixed capability state names.
+Unsupported or malformed profiles fail configuration loading.
+
+## Security Boundary
+
+An allowed Marmot sender plus `unrestricted` is effectively remote code
+execution as the harness service user. Sender allowlisting authenticates who
+may request work; it does not constrain what an unrestricted backend may do.
+
+The `/<path>` picker selects a working directory beneath `$HOME`. It is not a
+filesystem, process, credential, or network containment boundary. Unrestricted
+deployments should use a dedicated OS user, container, or VM with narrowly
+scoped credentials and host access. Backend logical permission checks must not
+be described as an OS sandbox.
--- a/integrations/codex/marmot/src/codex.rs
+++ b/integrations/codex/marmot/src/codex.rs
@@ -3,7 +3,8 @@ use std::time::Instant;
 
 use async_trait::async_trait;
 use marmot_terminal_harness::{
-    Backend, HarnessError, Invocation, Outcome, RunFailure, RunnerEvent, TRACE_TARGET,
+    ApprovalSupport, Backend, ExecutionProfile, ExecutionSupport, HarnessError, Invocation,
+    IsolationSupport, Outcome, RunFailure, RunnerEvent, TRACE_TARGET,
     process::{capture_stderr, cleanup_failed_run, next_stdout_line, strip_ansi, write_stdin},
 };
 use serde_json::Value;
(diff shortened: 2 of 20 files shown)


integrations/codex/marmot/src/codex.rs
```diff
@@ -163,17 +188,31 @@ async fn run_with_bin(
...
+    let session_id = session_id.filter(|value| !value.is_empty());
+    let mut args = vec!["exec".to_owned()];
+    if session_id.is_some() {
+        args.push("resume".to_owned());
     }
+    match profile {
+        ExecutionProfile::Inherit => {}
+        ExecutionProfile::Autonomous => {
+            args.extend([
+                "--strict-config".to_owned(),
```
erskingardner: `--strict-config` is the wrong fail-closed mechanism here.

In current Codex it means "error if `config.toml` contains fields this version does not recognize" ([openai/codex#20559](https://github.com/openai/codex/pull/20559)), not "fail if this approval override is unsupported." Two practical failures:

- Any extra/future key in the operator's Codex config makes every `autonomous` prompt fail, even though `-c approval_policy="never"` would have applied and preserved sandbox/network.
- `--strict-config` is newer than the override itself. An older `codex` that understands `-c` but not this flag fails autonomous for no profile reason.

`--ask-for-approval never` is still rejected after `exec` ([openai/codex#26602](https://github.com/openai/codex/issues/26602)), so the `-c` form is the right adapter workaround and `-c` is global, including `resume`. Drop `--strict-config`. Unknown `-c` values and unknown CLI flags already fail the invocation, which is the version gate the issue asked for.

Please also drop the `--strict-config` claim from the harness README / architecture note once this changes.

integrations/opencode/marmot/src/opencode.rs
```diff
@@ -219,6 +247,11 @@ pub(crate) fn build_run_args(session_id: Option<&str>, prompt: &str) -> Vec<Stri
     args
 }
 
+fn permission_overlay(profile: ExecutionProfile) -> Option<(&'static str, &'static str)> {
+    (profile == ExecutionProfile::Unrestricted)
+        .then_some(("OPENCODE_PERMISSION", r#"{"*":"allow"}"#))
```
erskingardner: This overlay does not actually request OpenCode's broadest mode.

Current OpenCode applies `OPENCODE_PERMISSION` as `mergeDeep(existing, JSON.parse(env))`. `{"*":"allow"}` only writes the wildcard key; more specific user/project/managed denies (`bash: { "rm *": "deny" }`, `external_directory`, etc.) remain. `--auto` then preserves those denies, so `unrestricted` collapses toward `autonomous` for any operator who already hardened permissions.

OpenCode's documented all-allow form is the string `"allow"` (`"permission": "allow"`). Set `OPENCODE_PERMISSION` to the JSON string `"allow"` (Rust `r#""allow""#`) so the merge replaces the permission value, and add a regression that the overlay is that string rather than a wildcard object. If `mergeDeep` does not replace a string source in the installed OpenCode version, use a process-local config content override instead — do not leave specific denies in place for a profile that required `--acknowledge-unrestricted`.

erskingardner (commented):
Would request changes, but GitHub does not allow that on your own PR.

Reviewed the unique #1504 commit after the #1501 rebase. The shared profile model, installer acknowledgement, and privacy-safe diagnostics match the issue. Two backend mappings do not do what the profiles claim.

1. Codex `autonomous` should not pass `--strict-config`. That flag rejects unrecognized keys in the operator's `config.toml`; it is not a profile-support gate. Operators with extra/future config keys will fail every autonomous prompt, and older Codex that lacks `--strict-config` will fail even though `-c approval_policy="never"` would work. `--ask-for-approval never` is rejected after `exec`, so keep the `-c` override and drop `--strict-config`.
2. OpenCode `unrestricted` sets `OPENCODE_PERMISSION={"*":"allow"}`, which `mergeDeep`s onto existing permission config. Tool-specific denies survive. Official all-allow form is the string `"allow"`.

Installer ack, fail-closed unknown profiles, and the Pi no-op mapping look right.

integrations/codex/marmot/src/codex.rs
erskingardner: Addressed in 21aceeb2. Autonomous now passes only the invocation-local `-c approval_policy="never"` override for both new and resumed runs; `--strict-config` and its documentation claim were removed. The updated argument regression and the full `wn-codex` suite pass.

integrations/opencode/marmot/src/opencode.rs
erskingardner: Addressed in 21aceeb2. Unrestricted now sets process-local `OPENCODE_PERMISSION` to the JSON scalar `"allow"`, so it replaces the permission value rather than merging a wildcard alongside surviving specific denies. The regression asserts the exact scalar overlay, and the full `wn-opencode` suite passes.

integrations/opencode/marmot/src/opencode.rs
```diff
@@ -219,6 +247,10 @@ pub(crate) fn build_run_args(session_id: Option<&str>, prompt: &str) -> Vec<Stri
     args
 }
 
+fn permission_overlay(profile: ExecutionProfile) -> Option<(&'static str, &'static str)> {
+    (profile == ExecutionProfile::Unrestricted).then_some(("OPENCODE_PERMISSION", r#""allow""#))
```
erskingardner: This still does not honor `unrestricted`.

Current OpenCode does `result.permission = mergeDeep(result.permission ?? {}, JSON.parse(OPENCODE_PERMISSION))`. remeda `mergeDeep` is `{ ...destination, ...source }` plus recursive object merging. A JSON string `"allow"` is not a replacement: spreading `"allow"` yields `{0:"a",1:"l",...}` and leaves existing tool/pattern denies in place. In the no-existing-permission case you also do not get the documented scalar `"allow"`.

The documented all-allow form works when `permission` is set on the config object, not when it is used as the `mergeDeep` source. Set a process-local `OPENCODE_CONFIG_CONTENT={"permission":"allow"}` instead. `mergeDeep(existingConfig, { permission: "allow" })` replaces the field because the source value is a non-object. That overlay stays invocation-local and still does not rewrite global/project files.

Please update the helper, the regression, and the harness README matrix. Managed/MDM config is applied after `OPENCODE_CONFIG_CONTENT` and can still win; that is acceptable and should stay documented as org policy, not as this profile working.

erskingardner (commented):
Would request changes, but GitHub does not allow that on your own PR.

Re-reviewed 21aceeb2 after the previous comments. Codex `autonomous` is fixed: `--strict-config` is gone, the invocation-local `-c approval_policy="never"` override remains, and the docs match.

OpenCode `unrestricted` is still wrong. `OPENCODE_PERMISSION="allow"` does not replace existing permission config. remeda `mergeDeep` requires object sources and spreads a string (`{ ...existing, ..."allow" }`), so denies survive and the overlay can become character-index junk. Use process-local `OPENCODE_CONFIG_CONTENT={"permission":"allow"}` instead; that merge happens at the config-object level, where a string `permission` value actually replaces the previous object.

Installer acknowledgement, Pi no-op mapping, and the rest of the shared profile plumbing still look right.

docs/marmot-architecture/overview/terminal-harness-execution-profiles.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Installer Contract
+
+The shared installer writes the selected profile into the private harness env
+file and same-user service definition. It requires
+`--acknowledge-unrestricted` before writing `unrestricted`; `--yes`, dry-run,
+or an environment-provided profile do not substitute for that acknowledgement.
+
+Startup tracing contains only the profile and fixed capability state names.
+Unsupported or malformed profiles fail configuration loading.
```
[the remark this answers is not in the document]
erskingardner: Addressed in 5642a15e. Both documents now describe the selected profile and typed capability states alongside the actual coarse startup fields (harness name, sender count, and reply-size limit), while retaining the privacy exclusion list.

integrations/opencode/marmot/README.md
```diff
@@ -104,6 +105,12 @@ account in its home; the harness sender allowlist does not narrow that token's
 authority. Do not give it to an untrusted harness. Use a separate connector
 home, socket, token, and account for a separate trust boundary.
 
+`autonomous` passes OpenCode's `--auto`, which auto-approves asks while
+preserving explicit denies. `unrestricted` also supplies a connector-owned,
+process-local allow-all permission overlay; it does not rewrite OpenCode's
+global or project configuration. See the shared
+[execution-profile capability matrix](../../terminal-harness/README.md#execution-profiles).
```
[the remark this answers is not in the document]
erskingardner: Addressed in 5642a15e. The OpenCode README now states that the backend provides no built-in OS isolation, requires external isolation for unrestricted deployments, and documents that managed organization policy may still override the process-local setting.

integrations/opencode/marmot/src/opencode.rs
erskingardner: Addressed in 5642a15e. `unrestricted` now removes inherited `OPENCODE_PERMISSION` and sets process-local `OPENCODE_CONFIG_CONTENT={"permission":"allow"}`. This was verified against current OpenCode config loading: config content merges at the config-object level, while organization/managed policy loads later and retains precedence. The regression asserts the removed and set variables exactly; `wn-opencode`, connector E2E, and `just fast-ci` pass.

erskingardner (commented):
Recommend merge. Would approve, but GitHub does not allow that on your own PR.

Re-reviewed 5642a15e. The previous blockers are fixed:

- Codex `autonomous` still uses only the invocation-local `-c approval_policy="never"` override.
- OpenCode `unrestricted` now sets process-local `OPENCODE_CONFIG_CONTENT={"permission":"allow"}` and removes inherited `OPENCODE_PERMISSION`, which is the merge path that actually replaces the permission field. Docs correctly note that managed/MDM policy can still win.

Installer acknowledgement, Pi no-op mapping, privacy-safe startup diagnostics, and the backend argument regressions look right. No remaining blocking issues.

PR #1511, opened 2026-08-19
Add large-group pressure scenarios

erskingardner:
## What changed

- add the deterministic `large-group-pressure/v1` scenario family with a 54-case catalog spanning 10–200 members
- vary administrator population independently from active committer width across application-heavy, growth, competing-commit, mixed, and churn workloads
- preserve versioned workload metadata through generated inputs, replay provenance, and reports
- add strict whole-group convergence, pending-work, policy, profile, and bounded decryptability oracles
- document campaign selection and execution tiers for routine, nightly, and scheduled runs

## Why

The simulator did not have an explicit, replayable family for testing large Marmot groups across size, administrator density, commit contention, application traffic, incremental growth, and membership churn. This makes those dimensions first-class and keeps failures reproducible by family, generator version, seed, and case index.

## Impact

The ordinary test lane executes a 10-member application-heavy canary. Generator tests compile the complete catalog without paying the runtime cost of all large and extra-large cases. A six-arm mid-size real-engine test is available as an explicit scheduled/manual gate, with documentation recommending isolated workers for the larger profiles.

## Validation

- `just fast-ci`
- `cargo test -p cgka-conformance-simulator --locked`
- `cargo test -p cgka-conformance-simulator --test large_group_family mid_size_complete_arm_catalog_passes_strict_oracles -- --ignored`
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Added deterministic large-group pressure campaigns for groups of 10–200 members.
  - Added workload profiles covering fanout, growth, administration, competing commits, mixed traffic, and handoff/rejoin scenarios.
  - Preserved workload metadata and replay provenance in generated reports.

- **Documentation**
  - Documented campaign profiles, execution guidance, validation coverage, and cost-based scheduling.

- **Tests**
  - Added coverage for deterministic generation, catalog dimensions, replay metadata, delivery behavior, canary scenarios, and real-engine execution.


erskingardner (commented):
Cannot use GitHub’s request-changes event on this PR (same author), but these should be fixed before this v1 is frozen.

This family should exist. Large-group size, administrator density, and commit-width are real uncovered dimensions, and the versioned workload profile plus prefix-stable generator are the right shape.

Two catalog-layout issues are blocking. The CLI has no case-start offset, and generator versions pin case indices after merge:

1. **Case indices are not cost-monotonic.** The 16/32/64/128 anchors occupy `0..23`, so the cheap 10-member six-arm block sits at `24..29` — behind the 128-member block. `--cases 6` is the 16-member block, not the 10-member canary, and no CLI prefix can run the 10-member arms without also paying for 128. Order sizes by execution cost.
2. **Competing and mixed arms sometimes have a single committer.** Admin regime is `(size_slot + case_index) % 5`, so FounderOnly lands on CompetingCommits at 32 and 50 members, and on MixedInterleaved at 128 and 200. Those arms then skip reorder/duplicate. Floor admin or committer count for the race arms.

The ordinary executable canary is only the 10-member application-heavy arm. Incremental growth and handoff (the retained-join-commit workaround) never run in PR CI. Please execute at least one 10-member join-wart arm in the normal lane.

Minor: crate `AGENTS.md` still lists generated families only under `family.rs` and does not mention `src/large_group_family.rs`.

crates/cgka-conformance-simulator/src/large_group_family.rs
```diff
@@ -0,0 +1,1040 @@
...
+        name: "xlarge-max-200",
+        tier: "xlarge",
+        members: 200,
+    },
+];
+
+#[derive(Clone, Copy)]
+enum WorkloadArm {
+    BulkApplicationFanout,
+    IncrementalGrowth,
```
erskingardner: This order fights the campaign CLI. There is no case-start offset, so `--cases N` is always a prefix of this array. Putting 128-member cases at indices 18–23 means:

- `--cases 6` is the 16-member block, not the 10-member canary the tests/docs talk about
- `--cases 24` already includes the xlarge-128 block
- the cheap 10-member six-arm catalog (cases 24–29, including the ignored test) cannot be selected without also running 128-member cases

This is still generator v1, so reorder `SIZE_PROFILES` by execution cost (`10, 16, 20, 32, 50, 64, 100, 128, 200`) and keep the anchor-vs-boundary distinction in `name`/`tier` metadata. After merge, changing this order is a generator-version bump.

crates/cgka-conformance-simulator/src/large_group_family.rs
```diff
@@ -0,0 +1,1040 @@
...
+        .collect()
+}
+
+pub fn generate_large_group_pressure_case(seed: u64, case_index: u64) -> GeneratedScenarioCase {
+    let mut rng = StdRng::seed_from_u64(seed ^ 0x4c41_5247_4547_5250 ^ case_index.rotate_left(31));
+    let arm = WorkloadArm::from_index(case_index);
+    let size_slot = usize::try_from(case_index / ARM_COUNT).expect("case index fits usize")
+        % SIZE_PROFILES.len();
+    let size = SIZE_PROFILES[size_slot];
+    let admin_regime = AdminRegime::from_index(size_slot + usize::try_from(case_index).unwrap());
```
erskingardner: `AdminRegime::from_index(size_slot + case_index)` does not vary admin population independently of the arm. With 6 arms and 5 regimes, FounderOnly is assigned to:

- CompetingCommits at 32 and 50 members (cases 9 and 39)
- MixedInterleaved at 128 and 200 members (cases 22 and 52)

`choose_committers` then returns a single admin, and both `DuplicateMessage` and `ReorderMessages` are skipped because `selectors.len() > 1` is false. Those sizes therefore never actually race commits, which is the arm’s purpose — and mixed-interleaved going degenerate at 128/200 is the worst place for it.

Keep a floor of ≥2 (preferably ≥ the arm’s intended committer width) for CompetingCommits and MixedInterleaved, e.g. bump FounderOnly to Few/Dense on those arms. The catalog-shape test should assert that those arms always emit a reorder or duplicate, not only in the first size block.

crates/cgka-conformance-simulator/tests/large_group_family.rs
```diff
@@ -0,0 +1,240 @@
...
+                }
+                TraceExpectation::NoPendingWorkExceptRetainedJoinCommit { client } => {
+                    pending_scope.insert(client.clone());
+                }
+                _ => {}
+            }
+        }
+        assert_eq!(
+            pending_scope, all_clients,
+            "every client's terminal work is pinned"
```
erskingardner: This is the only real-engine execution in the ordinary lane, and it is the bulk-create application-heavy arm (no late join, no retained join-commit, no restart). Incremental growth and handoff are the paths that strip `AwaitQuiescence` / `AdvanceTime` and pin `NoPendingWorkExceptRetainedJoinCommit` — compile_scenario does not exercise that.

Please run at least one 10-member join-wart arm (case 25 or 29 in the current catalog, or the equivalent after a cost-order shuffle) in the non-ignored suite. The six-arm ignored test can stay gated if the full block is too slow.

Datawav (commented):
## Automated review

Reading the PR audit bundle and reviews, then reconciling claims against the exact-head patches.
Checking the test file for canary coverage and any overlooked issues:
**PR 1511** (`Add large-group pressure scenarios`) · head `6b6a7f77` · initial · 19/19 files · 16/16 CI green · `pr_discussion` empty

Reconciled both independent reviews and cross-verifications against the exact-head patches. Discarded Kimi’s incomplete draft (missing-struct false positive, out-of-bundle `push_family` note, hallucinated `generated_case.rs`). Verified: `generate_family_case` dispatch in `family.rs`; `GeneratedWorkloadProfileV1` + backward-compatible `Option` serde threading; 54-case catalog with `admin_regime_for` founder-only swap on race arms; retained-join pending-work split; bounded probe cohort; two non-ignored 10-member engine canaries plus documented compile-only large/xlarge scope.

No actionable findings.

crates/cgka-conformance-simulator/src/large_group_family.rs
erskingardner: Addressed on 6b6a7f77: `SIZE_PROFILES` is now cost-ordered 10/16/20/32/50/64/100/128/200, and `--cases 6` is the 10-member six-arm prefix.

crates/cgka-conformance-simulator/src/large_group_family.rs
erskingardner: Addressed on 6b6a7f77: `admin_regime_for` moves founder-only off the competing/mixed slots, and every size block now asserts ≥2 committers plus reorder/duplicate.

crates/cgka-conformance-simulator/tests/large_group_family.rs
erskingardner: Addressed on 6b6a7f77: `mid_size_incremental_join_canary_passes_strict_oracles` now executes the 10-member retained-join path in the ordinary lane.

erskingardner (commented):
Recommend merge. Cannot use GitHub’s approve event on this PR (same author).

The earlier blocking catalog issues are fixed on 6b6a7f77.

- Size blocks are cost-ordered, so `--cases 6/24/36/54` are real budget prefixes.
- `admin_regime_for` keeps competing/mixed arms at ≥2 committers; the catalog test pins reorder/duplicate on every size block.
- The ordinary lane now executes the 10-member retained-join canary, not just application-heavy fanout.
- Probe selection caps in priority order, then roster-sorts without a second truncation. Pending-work tests now distinguish retained-join exceptions from strict `NoPendingWork`.

CI is green, including Simulator and vectors. No remaining blocking findings.

PR #1513, opened 2026-08-19
marmot-app: let an embedder supply the nostr-sdk client

lopatin96:
### Problem

`marmot-app` constructs its own `nostr-sdk` client in two places:

- `MarmotRelayPlane::from_sdk` for the shared relay plane, and
- `MarmotApp::relay_client_for_account_id` for each account's publishing client.

Both are private, so an application embedding `marmot-app` cannot run the runtime over a transport of its own. `nostr-sdk` already supports substituting one through `ClientBuilder::websocket_transport`; there is just no way to reach it from outside this crate.

Closing only one of the two is worse than closing neither: the relay plane would use the supplied transport while every account publish — relay lists, KeyPackages, welcomes — still went out over the default one, and that is the half that decides whether anyone can invite you.

### Change

Three additions. Nothing changes for callers that do not use them.

- `MarmotRelayPlane::from_sdk_client(client, lookback, allow_loopback)` — mirrors the existing private `from_sdk` exactly, including building `NostrSdkDirectoryRelayFetcher` from the same client, so the KeyPackage lookup path cannot silently fall back to a different transport.
- `MarmotApp::with_relay_plane(plane)` — installs it before any account work.
- `MarmotApp::with_account_client_factory(factory)` — supplies the per-account publishing client, given the account's signer.

```rust
let client = Client::builder().websocket_transport(my_transport.clone()).build();
let plane = MarmotRelayPlane::from_sdk_client(client, Some(lookback), false);

let app = MarmotApp::try_with_relays_and_account_home_and_config(root, relays, home, config)?
    .with_relay_plane(plane)
    .with_account_client_factory(Arc::new(move |signer| {
        Client::builder()
            .signer(signer)
            .websocket_transport(my_transport.clone())
            .build()
    }));
```

### Notes

- `AccountClientFactory` is a type alias so the signature stays readable at the call site.
- I kept the factory optional rather than making it a constructor argument, to avoid touching the existing constructors.
- Happy to reshape this — for instance as a single `TransportProvider` trait covering both, if you would rather have one seam than two.

### Checks

`cargo fmt --check` and `cargo clippy --all-targets` clean, `cargo test -p marmot-app` passes (728 tests).

## Summary by CodeRabbit

- **New Features**
  - Added configurable account-specific Nostr client creation.
  - Added support for supplying a custom relay plane.
  - Added a way to build relay connections from an existing Nostr SDK client.
  - Existing default client and relay behavior remains available when no custom configuration is provided.

### DIFF
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -558,6 +558,10 @@ pub struct MarmotApp {
     /// pool instead of constructing another TCP/TLS/WebSocket stack.
     account_publish_clients: Arc<Mutex<HashMap<String, Arc<dyn NostrRelayClient>>>>,
     account_client_factory: Option<AccountClientFactory>,
+    /// Whether [`MarmotApp::relay_plane`] came from the embedder rather than from this crate.
+    ///
+    /// Decides both who the transport belongs to and who is responsible for winding it down.
+    embedder_relay_plane: bool,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -1333,12 +1337,24 @@ impl MarmotApp {
         )
     }
 
-    /// Replaces the relay plane this app will use.
+    /// Replaces the relay plane this app will use, everywhere it uses one.
     ///
     /// Call it before any account work. Pairs with [`MarmotRelayPlane::from_sdk_client`] to
-    /// run the runtime over an embedder-supplied transport.
+    /// run the runtime over an embedder-supplied transport. From here on the supplied plane
+    /// also backs [`MarmotApp::client`], which would otherwise open over a default plane and
+    /// put shared relay traffic on a connection the embedder does not control.
     pub fn with_relay_plane(mut self, relay_plane: MarmotRelayPlane) -> Self {
-        self.relay_plane = relay_plane;
+        let replaced = std::mem::replace(&mut self.relay_plane, relay_plane);
+        let ours = !self.embedder_relay_plane;
+        self.embedder_relay_plane = true;
+        // Constructing the default plane inside a runtime already started its router and
+        // notification forwarder, and those keep the old transport alive after the plane
+        // itself is dropped. Wind down the one this crate built; never touch one the caller
+        // supplied, since the caller may still be holding it. Outside a runtime nothing was
+        // started and there is nothing to wind down.
+        if ours && let Ok(handle) = tokio::runtime::Handle::try_current() {
+            handle.spawn(async move { replaced.shutdown().await });
+        }
         self
     }
 
@@ -1395,6 +1411,7 @@ impl MarmotApp {
             external_signers: Arc::new(Mutex::new(HashMap::new())),
             account_publish_clients: Arc::new(Mutex::new(HashMap::new())),
             account_client_factory: None,
+            embedder_relay_plane: false,
         }
     }
 
@@ -1469,6 +1486,7 @@ impl MarmotApp {
             external_signers: Arc::new(Mutex::new(HashMap::new())),
             account_publish_clients: Arc::new(Mutex::new(HashMap::new())),
             account_client_factory: None,
+            embedder_relay_plane: false,
         }
     }
 
@@ -1527,16 +1545,20 @@ impl MarmotApp {
     /// retrying.
     pub async fn client(&self, label: &str) -> Result<AppClient, AppError> {
         #[cfg(test)]
-        let relay_plane = self
-            .test_relay_client
-            .as_ref()
-            .map(|client| MarmotRelayPlane::new(None, client.clone()))
-            .unwrap_or_else(|| {
-                MarmotRelayPlane::full_history_with_loopback(
-                    self.config.allow_loopback_relay_endpoints,
-                )
-            });
-        #[cfg(not(test))]
+        if let Some(client) = &self.test_relay_client {
+            let relay_plane = MarmotRelayPlane::new(None, client.clone());
+            return self
+                .client_with_relay_plane(label, &relay_plane, None)
+                .await;
+        }
+        // An embedder that supplied a plane supplied the only transport this app may use.
+        // Its own lookback stands: the full-history default below is this crate's choice for
+        // a plane this crate built, not a property the caller's plane has to inherit.
+        if self.embedder_relay_plane {
+            return self
+                .client_with_relay_plane(label, &self.relay_plane, None)
+                .await;
+        }
         let relay_plane = MarmotRelayPlane::full_history_with_loopback(
             self.config.allow_loopback_relay_endpoints,
         );
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -499,6 +499,12 @@ type AppRuntime = AccountDeviceRuntime<
 #[cfg(test)]
 type LegacyProjectionOpenHook = Arc<dyn Fn() + Send + Sync>;
 
+/// Builds the `nostr-sdk` client an account publishes through, given its signer.
+///
+/// Supplied by an embedder that runs the runtime over a transport of its own.
+pub type AccountClientFactory =
+    Arc<dyn Fn(Arc<dyn nostr::NostrSigner>) -> NostrSdkClient + Send + Sync>;
+
 #[derive(Clone)]
 pub struct MarmotApp {
     root: PathBuf,
@@ -551,6 +557,7 @@ pub struct MarmotApp {
     /// account worker share this client so the worker can reuse the same relay
     /// pool instead of constructing another TCP/TLS/WebSocket stack.
     account_publish_clients: Arc<Mutex<HashMap<String, Arc<dyn NostrRelayClient>>>>,
+    account_client_factory: Option<AccountClientFactory>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -1326,6 +1333,15 @@ impl MarmotApp {
         )
     }
 
+    /// Replaces the relay plane this app will use.
+    ///
+    /// Call it before any account work. Pairs with [`MarmotRelayPlane::from_sdk_client`] to
+    /// run the runtime over an embedder-supplied transport.
+    pub fn with_relay_plane(mut self, relay_plane: MarmotRelayPlane) -> Self {
+        self.relay_plane = relay_plane;
+        self
+    }
+
     pub fn with_relays_and_config(
         root: impl AsRef<Path>,
         relay_urls: Vec<String>,
@@ -1378,6 +1394,7 @@ impl MarmotApp {
             audit_log_tracker_config: Arc::new(Mutex::new(AuditLogTrackerConfig::default())),
             external_signers: Arc::new(Mutex::new(HashMap::new())),
             account_publish_clients: Arc::new(Mutex::new(HashMap::new())),
+            account_client_factory: None,
         }
     }
 
@@ -1451,6 +1468,7 @@ impl MarmotApp {
             audit_log_tracker_config: Arc::new(Mutex::new(AuditLogTrackerConfig::default())),
             external_signers: Arc::new(Mutex::new(HashMap::new())),
             account_publish_clients: Arc::new(Mutex::new(HashMap::new())),
+            account_client_factory: None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


lopatin96:
Both findings were real; pushed a follow-up.

`client()` built its own default plane regardless of `with_relay_plane`, so shared relay work went out over the default stack while the runtime used the supplied one. It now opens over the supplied plane when there is one. I kept full history as the default for planes this crate builds rather than imposing that lookback on a plane the caller configured.

The replaced plane is now wound down instead of dropped: constructing the default inside a runtime starts its router and notification forwarder, and those stayed on the old transport. Only a plane this crate built gets shut down — a supplied one may still be held by the caller, so a second replacement leaves it alone.

Two tests cover this, both failing without the change. The five `tests/relay_runtime.rs` failures I see locally reproduce identically on a clean `master`.

erskingardner:
@lopatin96 can you tell me why you need to/would want to swap out the websocket implementation?

Datawav (commented):
## Automated review

Reading the PR audit bundle, reviews, and exact-head patches to reconcile claims and produce the final audit.
**Outcome:** Embedder transport hooks land, but three medium defects can silently keep publishes or sibling clones on the wrong transport; two low gaps remain in wind-down and test coverage. CI status not verifiable from bundle (`check_evidence.runs` empty).

**1. Medium — Stale publish-client cache** (`lib.rs`: `relay_client_for_account_id` `or_insert_with`; `with_account_client_factory`/`with_relay_plane` never clear `account_publish_clients`). Any account publish before both builders run permanently pins the default `NostrSdkClient` for that account; later factory/plane setup is ignored. **Fix:** clear the map in both builders, or key cache entries by factory/plane identity.

**2. Medium — Split transport on plane-only setup** (`lib.rs`: `client()` early-returns on `embedder_relay_plane`; `relay_client_for_account_id` still uses `NostrSdkClient::builder()` when factory is `None`; `with_relay_plane` doc pairs only with `from_sdk_client`). Shared relay traffic uses the embedder plane; per-account KeyPackage/relay-list/welcome publishes stay on the default stack—the half-closed failure this PR targets. **Fix:** derive publish clients from the embedder plane automatically, or document+enforce factory on `with_relay_plane` and fail closed when `embedder_relay_plane && factory.is_none()`.

**3. Medium — Clone-then-configure breaks siblings** (`lib.rs`: `#[derive(Clone)]`; `with_relay_plane` `mem::replace` + spawned `replaced.shutdown()`; `embedder_relay_plane` is a copied bool, not shared). One clone calling `with_relay_plane` shuts down the shared default plane still referenced by siblings; siblings keep `embedder_relay_plane == false`, so `client()` opens `full_history_with_loopback`—a third transport. Subscriptions die silently on surviving clones. **Fix:** skip shutdown unless `Arc::strong_count` shows sole ownership, or require configure-before-clone; do not let stale bool diverge across clones.

**4. Low — Cross-runtime wind-down leak** (`lib.rs`: shutdown only when `Handle::try_current()` succeeds). Default plane constructed inside a runtime keeps router/forwarder alive if replacement happens outside one; comment overclaims “nothing was started.” **Fix:** capture the construction runtime `Handle` for teardown.

**5. Low — Missing factory-path test** (`tests.rs`: new tests cover `client()` plane selection and default-plane wind-down only). `AccountClientFactory` / `relay_client_for_account_id` publish path is untested despite `require_behavioral_test_reasoning`. **Fix:** install a counting factory, drive first publish, assert factory-built client is cached.

crates/marmot-app/src/lib.rs
```diff
@@ -5436,7 +5486,10 @@ impl MarmotApp {
         clients
             .entry(account_id_hex.to_owned())
             .or_insert_with(|| {
```
Datawav, line 5488: **Medium — Stale publish-client cache.** `or_insert_with` permanently caches the first client built for each account. Neither `with_account_client_factory` nor `with_relay_plane` clears `account_publish_clients`, so any publish before both builders run pins the default `NostrSdkClient` for that account and later factory/plane setup is ignored.

**Fix:** clear the map in both builders, or key cache entries by factory/plane identity.

crates/marmot-app/src/lib.rs
```diff
@@ -5436,7 +5486,10 @@ impl MarmotApp {
...
-                let client = NostrSdkClient::builder().signer(signer).build();
+                let client = match &self.account_client_factory {
+                    Some(build) => build(signer),
+                    None => NostrSdkClient::builder().signer(signer).build(),
```
Datawav, line 5491: **Medium — Split transport on plane-only setup.** When `embedder_relay_plane` is set, `client()` correctly uses the embedder plane, but this `None` branch still builds per-account publish clients with the default `NostrSdkClient::builder()`. KeyPackage/relay-list/welcome publishes therefore stay on the default stack while shared relay traffic uses the embedder plane—the half-closed failure this PR targets.

**Fix:** derive publish clients from the embedder plane automatically, or document+enforce factory on `with_relay_plane` and fail closed when `embedder_relay_plane && factory.is_none()`.

crates/marmot-app/src/lib.rs
```diff
@@ -1326,6 +1337,27 @@ impl MarmotApp {
...
     }
 
+    /// Replaces the relay plane this app will use, everywhere it uses one.
+    ///
+    /// Call it before any account work. Pairs with [`MarmotRelayPlane::from_sdk_client`] to
+    /// run the runtime over an embedder-supplied transport. From here on the supplied plane
+    /// also backs [`MarmotApp::client`], which would otherwise open over a default plane and
+    /// put shared relay traffic on a connection the embedder does not control.
+    pub fn with_relay_plane(mut self, relay_plane: MarmotRelayPlane) -> Self {
+        let replaced = std::mem::replace(&mut self.relay_plane, relay_plane);
```
Datawav, line 1347: **Medium — Clone-then-configure breaks siblings.** `MarmotApp` is `#[derive(Clone)]`; `mem::replace` here plus spawned `replaced.shutdown()` means one clone calling `with_relay_plane` shuts down the shared default plane still referenced by siblings. Siblings keep their copied `embedder_relay_plane == false`, so `client()` opens `full_history_with_loopback`—a third transport—and subscriptions die silently on surviving clones.

**Fix:** skip shutdown unless `Arc::strong_count` shows sole ownership, or require configure-before-clone; do not let the stale bool diverge across clones.

crates/marmot-app/src/lib.rs
```diff
@@ -1326,6 +1337,27 @@ impl MarmotApp {
...
+        let ours = !self.embedder_relay_plane;
+        self.embedder_relay_plane = true;
+        // Constructing the default plane inside a runtime already started its router and
+        // notification forwarder, and those keep the old transport alive after the plane
+        // itself is dropped. Wind down the one this crate built; never touch one the caller
+        // supplied, since the caller may still be holding it. Outside a runtime nothing was
+        // started and there is nothing to wind down.
+        if ours && let Ok(handle) = tokio::runtime::Handle::try_current() {
```
Datawav, line 1355: **Low — Cross-runtime wind-down leak.** Shutdown of the replaced default plane runs only when `Handle::try_current()` succeeds. If the default plane was constructed inside a runtime but replacement happens outside one, router/forwarder tasks keep the old transport alive; the comment above overclaims "nothing was started."

**Fix:** capture the construction runtime `Handle` for teardown.

crates/marmot-app/src/tests.rs
```diff
@@ -10337,3 +10337,68 @@ async fn reconcile_repairs_stale_two_member_count_on_three_member_group_body() {
     );
     runtime.shutdown().await;
 }
+
+#[test]
+fn client_opens_over_an_embedder_supplied_relay_plane() {
```
Datawav, line 10342: **Low — Missing factory-path test.** New tests cover `client()` plane selection and default-plane wind-down only. The `AccountClientFactory` / `relay_client_for_account_id` publish path is untested despite `require_behavioral_test_reasoning`.

**Fix:** install a counting factory, drive first publish, assert the factory-built client is cached.

lopatin96:
I'm building a client where the relays a person uses are not public servers with DNS names, but nodes their contacts run — a spare machine, a home server — reached over QUIC and addressed by public key. No DNS, no certificate authority, no hostname anywhere in the path.

Everything above the transport stays as it is: the same wire format, the same event kinds, the same group semantics. The only thing replaced is what carries the bytes, which is why the seam I need is `websocket_transport()` — nostr-sdk already offers it, and MDK builds its clients internally, so an embedder cannot reach it.

Two places build a client, which is why the second half of the PR exists: the relay plane, and the per-account publishing client. Supplying only the first leaves the half that publishes KeyPackages and relay lists going out over the default TCP/TLS stack — so an account that looks migrated still leaks the transport it was meant to leave.

I'm not asking MDK to know anything about that transport: the injected client is an ordinary nostr-sdk client, and with nothing supplied the behaviour is exactly what it is today.

erskingardner:
Thanks for the explanation @lopatin96. I think your three PRs actually point to a more fundamental set of changes that we can make to ensure that marmot is actually transport agnostic while preserving some our core nostr-focus. 

I've created a doc to track these things here: docs/marmot-architecture/overview/transport-agnosticism.md

I'm currently leaning towards closing your PRs and doing this in a more holistic way so that we address it completely. what do you think?

Datawav (commented):
## Automated review

**Outcome:** Four actionable defects remain at exact head `e72145e6bce7d1779136cbcd02b390574dfddf17`. The earlier `client()` bypass is fixed. The maintainer’s preference for a holistic transport redesign is a process decision, not a rebuttal of the current patch.

1. **Medium — Installing a factory does not invalidate cached default publishers**
   - **Changed-line evidence:** `crates/marmot-app/src/lib.rs` adds `account_client_factory`, while `with_account_client_factory` only assigns that field. The existing `relay_client_for_account_id` path still returns `account_publish_clients.entry(...).or_insert_with(...)`, and the shared cache is not cleared.
   - **Trigger:** An account publisher is created before `with_account_client_factory`, including through a sibling `MarmotApp` clone.
   - **Consequence:** Subsequent relay-list, KeyPackage, and welcome publishes continue through the cached default TCP/TLS client instead of the supplied transport.
   - **Confidence:** High.
   - **Smallest sound correction:** Clear `account_publish_clients` whenever the factory changes, or version cache entries by factory identity.

2. **Medium — Plane-only configuration still permits split transport**
   - **Changed-line evidence:** `with_relay_plane` now sets `self.relay_plane` and `self.embedder_relay_plane = true`, and `client()` uses that plane. But the changed `relay_client_for_account_id` branch still executes `NostrSdkClient::builder().signer(signer).build()` whenever `account_client_factory` is `None`.
   - **Trigger:** An embedder calls `with_relay_plane(MarmotRelayPlane::from_sdk_client(...))` without also installing an account-client factory.
   - **Consequence:** Shared relay operations use the supplied transport while account publishes silently use the default stack, violating the new “only transport”/“everywhere” contract.
   - **Confidence:** High.
   - **Smallest sound correction:** Fail closed when an embedder plane lacks a compatible account factory, unless the plane can expose a signer-bound publishing client.

3. **Medium — Replacing a plane after cloning shuts down transport still used by siblings**
   - **Changed-line evidence:** `MarmotApp` remains `#[derive(Clone)]`. In the added `with_relay_plane` logic, `mem::replace` removes this handle’s plane, `ours` is derived from this handle’s copied `embedder_relay_plane` flag, and `replaced.shutdown()` is spawned before only this handle is marked embedder-owned.
   - **Trigger:** Clone a default-configured app, then call `with_relay_plane` on one clone.
   - **Consequence:** Shutdown affects the shared default plane still referenced by the sibling, while the sibling’s copied ownership flag remains unchanged; its subscriptions can die and later `client()` calls can create another default transport.
   - **Confidence:** High.
   - **Smallest sound correction:** Put plane identity and ownership in shared state and replace it atomically for all clones; do not shut down a plane while another app clone still owns it.

4. **Low — Off-runtime replacement can leave the old plane running**
   - **Changed-line evidence:** The new teardown is guarded by `tokio::runtime::Handle::try_current()` and only then calls `handle.spawn(replaced.shutdown())`.
   - **Trigger:** Construct the default plane inside a Tokio runtime, then invoke `with_relay_plane` from a thread without a current runtime.
   - **Consequence:** The old router and notification-forwarder tasks remain alive on the superseded transport.
   - **Confidence:** High.
   - **Smallest sound correction:** Retain the runtime handle that started the plane and use it for shutdown, rather than requiring a runtime at replacement time.

Issue #341, opened 2026-06-27
Background `stream watch` started without `--account` broadcasts stream previews to every account's subscribers (null-account = match-all)


erskingardner:
Found via a systematic review of the daemon subscription-matching path.

## Summary

The daemon's subscription filter `value_matches_group_and_account` treats a missing `account` field as **match-all**. A background stream watch started without `--account` records `account = None`, so its `stream_preview` broadcasts are delivered to *any* account's message subscriber that watches the same group — crossing the per-account confidentiality boundary.

## Where

- Match-all on null account: `crates/cli/src/daemon/responses.rs:346-358` — `value_matches_group_and_account` uses `event_account ... .is_none_or(|event_account| event_account == account_id)`, i.e. `None` matches every `account_id`.
- Null account on the background watch: `crates/cli/src/daemon/stream_workers.rs:192-198` — `AgentStreamWatchStart { account: cli.account.clone(), .. }`; `cli.account` is `None` when `--account` is omitted.

## Why it matters

When two local accounts are members of the same group, an account-less watch started by account A will surface its `stream_preview` text/status updates to account B's `messages subscribe` stream for that group. Stream previews are message content; leaking them across local-account boundaries violates the account-isolation expectation of the daemon's per-account subscription feeds.

## Suggested fix

Require `--account` for `stream watch --background` (reject `None`), or resolve/stamp the watch's owning account before broadcasting and make `value_matches_group_and_account` fail closed when the event carries no account rather than matching all.

---
_Generated by [Claude Code](https://claude.ai/code)_

erskingardner:
Found during an automated review at `6888f97`: the root cause is broader than a background watch started without `--account` — the account filter predicate itself is **fail-open**.

`crates/cli/src/daemon/responses.rs:346-358` (`value_matches_group_and_account`):
```rust
... && value
    .get("account")
    .or_else(|| value.get("account_id"))
    .and_then(serde_json::Value::as_str)
    .is_none_or(|event_account| event_account == account_id)
```
`is_none_or` means any event value where the account field is **absent or serialized as JSON `null`** (`as_str` -> `None`) matches *every* account. The shared `events`/`stream_manager` broadcast channels fan out to all subscribers regardless of account, so per-subscriber isolation depends entirely on this predicate. `AgentStreamWatchReport.account` and `AgentStreamDelta.account` are `Option<String>` and serialize to `"account": null` when started without an account — so compose deltas (not just watch previews) leak across accounts too.

[...]

PR #1514, opened 2026-08-19
marmot-app: permit a named loopback blob origin instead of all of them

lopatin96:
### Problem

`allow_loopback_blob_endpoints` is all-or-nothing, and blob endpoints are group state. With the flag on, an endpoint chosen by a group admin reaches the local host of every member, for both upload and fetch — `http://127.0.0.1:PORT/` in `default_blob_endpoints` is valid component state, and validity is deliberately independent of reachability.

That is the right trade for a test harness. It is an uncomfortable one for an application that runs its own blob service on loopback and wants to reach that and nothing else.

### Change

`permit_loopback_blob_origin("http://127.0.0.1:PORT")` names one origin. It is a property of the process — it describes a service this same program is running — rather than of an account or a group, and every other loopback address stays refused no matter who put it in group state. The existing flag is untouched.

```rust
marmot_app::permit_loopback_blob_origin("http://127.0.0.1:7871");
// http://127.0.0.1:7871/... is now fetchable and uploadable;
// http://127.0.0.1:7872/... and http://localhost:7871/... are not.
```

It needed both gates rather than one:

- `host_safety::validate_blossom_fetch_url` checks the scheme, and
- `blossom::media_http_client_for_url` separately checks the resolved address.

Patching only the first leaves the scheme accepted and the connection refused a line later, which reads like a bug rather than a policy.

### Tests

Added to `media/tests.rs`: the neighbouring port, a different loopback host (`localhost`, `[::1]`), cleartext to a public host, recognition from raw endpoint strings, and inputs that are not loopback HTTP URLs at all. The last case is ignored rather than accepted — a caller naming a public host has made a mistake, and the safe reading of that mistake is that nothing was permitted.

### Notes

- The allowlist is a process-global `OnceLock<Mutex<Vec<String>>>`, which matches the scope of the thing it describes but does mean tests that permit an origin affect the whole test binary. The added tests use distinctive ports for that reason. If you would rather carry it in `MarmotAppConfig` and thread it through, I am glad to redo it that way — it is a larger diff because the boolean is threaded through several call sites.

### Checks

`cargo fmt --check` and `cargo clippy --all-targets` clean, `cargo test -p marmot-app` passes (731 tests).

### DIFF
--- a/crates/marmot-app/src/media/tests.rs
+++ b/crates/marmot-app/src/media/tests.rs
@@ -11,7 +11,9 @@ use url::Url;
 use super::blossom::{
     MAX_BLOSSOM_DESCRIPTOR_BYTES, MAX_ENCRYPTED_MEDIA_BLOB_BYTES, read_limited_blossom_body,
 };
-use super::host_safety::validate_blossom_fetch_url;
+use super::host_safety::{
+    is_permitted_loopback_endpoint, permit_loopback_blob_origin, validate_blossom_fetch_url,
+};
 
 #[test]
 fn encrypted_media_blob_limit_supports_large_application_artifacts() {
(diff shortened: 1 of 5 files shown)


erskingardner:
@lopatin96 again - very curious about the actual use case here.

crates/marmot-app/src/media/tests.rs
```diff
@@ -1688,3 +1690,63 @@ fn shared_golden_v1_fixtures_parse_validate_and_round_trip_exactly() {
 fn shared_golden_v2_fixtures_parse_validate_and_round_trip_exactly() {
     assert_shared_media_fixture_file("imeta-v2.json");
 }
+
+#[test]
+fn a_permitted_loopback_origin_is_accepted_and_no_other_loopback_address_is() {
```
Datawav, line 1695: **Low — missing end-to-end regression for permitted-origin fetch/upload**

These new tests exercise only `permit_loopback_blob_origin`, `validate_blossom_fetch_url`, and `is_permitted_loopback_endpoint`. The production branches in `fetch_encrypted_media_blob` (`candidates.retain` in `mod.rs`) and `media_http_client_for_url` (`allow_loopback = … || is_permitted_loopback_url(url)` in `blossom.rs`) have no assertion with `allow_loopback_blob_endpoints=false`. Existing tokio tests pass `allow_loopback_http=true` throughout.

A miswire in either runtime branch (e.g. dropping the permit disjunct from `media_http_client_for_url`, or flipping retain to `&&`) would leave permitted origins unfetchable/unuploadable at runtime while all new unit tests still pass.

**Fix:** One tokio test using existing `spawn_http_response`: bind `127.0.0.1:0`, `permit_loopback_blob_origin` with that port, then fetch via `fetch_blossom_blob` and/or `download_encrypted_media` with `allow_loopback=false` and assert success; optionally mirror for upload.

Datawav (commented):
## Automated review

Reading the PR audit bundle and review artifacts, then reconciling claims against the exact-head patches.
Reading bundle metadata and key patches to verify findings against the exact head.
**Outcome:** [mdk#1514](https://github.com/marmot-protocol/mdk/pull/1514) @ `60596427` — security-tightening allowlist is correctly wired across all three gates; no correctness or security defect found. One **Low** test-coverage gap remains.

**Implementation verified:** `permit_loopback_blob_origin` adds process-scoped `scheme://host:port` keys; `validate_blossom_fetch_url` (scheme gate), `media_http_client_for_url` (resolved-address gate), and `fetch_encrypted_media_blob` candidate retain (`!loopback || is_permitted_loopback_endpoint`) are aligned. New unit tests cover allowlist boundaries (neighbor port, `localhost`/`[::1]`/`127.0.0.1` separation, cleartext public refusal, non-loopback permit inputs ignored). `pr_discussion` empty; CI outcome unverified (`runs=[]`).

---

**1. Low — missing end-to-end regression for permitted-origin fetch/upload**

**Evidence:** Added tests (`tests.rs` +63) call only `permit_loopback_blob_origin`, `validate_blossom_fetch_url`, and `is_permitted_loopback_endpoint`. Production branches at `mod.rs` ~775 (`candidates.retain`) and `blossom.rs` ~465 (`allow_loopback = … || is_permitted_loopback_url(url)`) have no assertion with `allow_loopback_blob_endpoints=false`. Existing tokio tests pass `allow_loopback_http=true` throughout.

**Consequence:** A miswire in either runtime branch (e.g. dropping the permit disjunct from `media_http_client_for_url`, or flipping retain to `&&`) would leave permitted origins unfetchable/unuploadable at runtime while all new unit tests still pass.

**Fix:** One tokio test using existing `spawn_http_response`: bind `127.0.0.1:0`, `permit_loopback_blob_origin` with that port, then fetch via `fetch_blossom_blob` and/or `download_encrypted_media` with `allow_loopback=false` and assert success; optionally mirror for upload.

---

**Discarded (reconciled):**

- **Medium — global permit list leaks across parallel tests:** Rejected. CI/local use nextest process-per-test (`.config/nextest.toml`); each test gets a fresh `OnceLock`. Head permits only `:47821`/`:47823`, not Issue #341's `:8080`. Process-global scope is author-disclosed design, not a current flake risk.
- Host aliasing strictness, silent ignore of bad permit inputs, Issue #341 comment drift, pre-existing V2 locator gap — correctly out of scope; unchanged or intentional.

lopatin96:
Same use case as #1513: a client whose relays are nodes its contacts run, reached over QUIC and addressed by public key rather than by hostname.

Media has to move over that same transport, and the runtime's Blossom client speaks plain HTTP — so the client runs a small loopback service that turns an HTTP GET/PUT into a QUIC request to the node holding the bytes. That service lives on 127.0.0.1 by definition.

The runtime refuses loopback blob endpoints unless `allow_loopback_blob_endpoints` is set, and that flag is all-or-nothing: turning it on to permit my own service also permits every other loopback address that could appear in a group's blob endpoints — a field chosen by whoever administers the group, not by the person whose machine would be dialled. Hence a single named origin the embedder permits explicitly, and the blanket flag left off.

Issue #1220, opened 2026-08-02
Restore automated RustSec auditing after the workspace replacement


Datawav:
## Summary

The retired `whitenoise-rs` audit recommended continuous dependency-vulnerability monitoring. MDK implemented that control in [PR #229](https://github.com/marmot-protocol/mdk/pull/229), adding a `cargo audit` CI job and local precommit recipe.

The Dark Matter workspace replacement commit [`9db9dbc0`](https://github.com/marmot-protocol/mdk/commit/9db9dbc092d52eb7b2d5bc6dcfb1f9beb3125464) removed the audit job and deleted the prior `justfile`. The current `Justfile` defines `fast-ci` and `ci` without an advisory scan ([lines 308–312](https://github.com/marmot-protocol/mdk/blob/cdf0e40e4fcdf96bcee89512c753755615537e33/Justfile#L308-L312)), and the current workflows contain no `cargo-audit`/RustSec job.

The specific historical vulnerabilities are not present in the current lockfile: `quinn-proto` is `0.11.15`, `libcrux-ecdh` is `0.0.8`, and `rsa` is absent. However, a fresh local audit now reports active [RUSTSEC-2026-0224](https://rustsec.org/advisories/RUSTSEC-2026-0224.html) plus one unsoundness and two yanked-package warnings. The active vulnerability should be remediated in its own security issue; this issue restores continuous detection so future advisory regressions cannot remain invisible. Closed [#918](https://github.com/marmot-protocol/mdk/issues/918) and [#1116](https://github.com/marmot-protocol/mdk/issues/1116) cleaned up earlier findings, but neither restores the removed gate.

## Publication and implementation dependency

[MDK #1218](https://github.com/marmot-protocol/mdk/issues/1218) is the open canonical remediation for RUSTSEC-2026-0224. Land #1218 before—or in the same coordinated change as—the fail-closed gate so the required `just ci` path is introduced against a lockfile that can pass it.

## Scope

Restore a pinned, reproducible RustSec advisory scan for the current workspace:

- a dedicated CI job on pull requests, merge queues, and `master` pushes when `Cargo.toml`, `Cargo.lock`, toolchain, or security-workflow inputs change;
- a local `audit` recipe included in full `just ci` (not necessarily the latency-sensitive `just fast-ci` gate);
- explicit policy for active vulnerabilities, yanked releases, and informational/unmaintained advisories;
- no blanket ignore list that can silently outlive its rationale.

## Acceptance criteria

[...]

PR #1516, opened 2026-08-19
fix(openclaw): deduplicate durable final sends on retry

Datawav:
## Summary

- add deterministic `send_final` idempotency keys for OpenClaw durable outbound text sends
- key every proactive and reply-driven profile-onboarding `send_final`
- wire OpenClaw's durable queue identity and unknown-send reconciliation into the production text-delivery path
- make profile onboarding consent and publication recovery torn-write safe
- retain the branch's Rust dependency-audit CI while moving Dependabot policy to a separate PR

## Finding 3 and contract

Composition Finding 3 was confirmed novel at the pinned pre-change head: OpenClaw's `outbound.ts` and `profile-onboarding.ts` omitted the optional `send_final.idempotency_key`, although the client, agent-control protocol, and connector already supported keyed receipt deduplication.

The implementation uses a namespaced `marmot-final-v1:` key derived from SHA-256 over compact ordered JSON UTF-8. The stable inputs bind:

- canonical destination/session identity
- OpenClaw's durable delivery intent (`deliveryQueueId`) or the onboarding send ordinal
- normalized reply anchor
- connector-compatible destination/text/reply content fingerprint

A durable text send fails closed before connector I/O when its stable delivery identity is unavailable. Identical retry, reconnect, or restart reconciliation reuses the exact key; changed content or reply identity derives a distinct key.

The normative vector was independently reproduced: content fingerprint `85876ccb…a186d`, final key `marmot-final-v1:59e4466f…07c1`.

## Wiring

- `outbound.ts` advertises text unknown-send reconciliation, requires the host queue identity, records platform dispatch before connector I/O, and retries unknown sends with the same connector key.
- `profile-onboarding.ts` centralizes prompt and reply sends behind the same keyed helper.
- onboarding persists `prompt_pending` before send and enters `prompted` only after successful or deduplicated delivery, so an undelivered prompt cannot authorize consent handling.
- profile publication persists its exact intent before connector I/O; recovery checks the live profile, avoids a second publish after ambiguous success, retries only after an authoritative profile-not-found result, and does not consume unrelated or indeterminate inbound messages.
- proactive onboarding prompts use a stable versioned ordinal; reply sends bind the normalized inbound message ID and response kind.
- the existing optional-key agent-control and connector contract remains unchanged.

## Regression coverage

Connector-faithful ambiguous-success tests verify:

- retrying an outbound final records one encrypted kind-9 post and returns the original receipt
- the normal adapter path records dispatch before connector I/O and reconciles with the same queue-derived key
- retrying onboarding prompt and reply sends reuses the same key
- an inbound message cannot be treated as consent while the prompt remains pending
- successful profile publication followed by local bookkeeping failure is recovered without republishing
- an authoritative profile-not-found result retries the persisted publication intent
- publication recovery does not consume unrelated-group or indeterminate inbound messages
- changed durable content or reply anchors produce distinct publications

## Verification

At `a7aba6f752d9b54439cf43357b32e3588fa9ffe0`:

- OpenClaw typecheck: pass
- OpenClaw tests: 270 passed, 2 skipped
- `just openclaw-dev-test`: pass
- `just fast-ci`: pass
- independent review: PASS with no findings (Cursor Grok 4.6)

## Dependency scope

- `.github/dependabot.yml` was removed from this branch as requested so Dependabot policy can land separately.
- Rust dependency-audit CI supporting [mdk#1220](https://github.com/marmot-protocol/mdk/issues/1220) remains in this PR.

This PR is ready for human review. It must not be merged by automation.

## Summary by CodeRabbit

* **New Features**
  * Improved reliable delivery of text messages, including safe retries and duplicate prevention.
  * Added recovery for uncertain message deliveries without creating duplicate posts.
  * Improved media handling, including host-provided media access and staged-file cleanup.
  * Added resilient profile onboarding with recoverable prompt and publication states.

* **Bug Fixes**
  * Prevented invalid or incomplete message identifiers from being processed.
  * Preserved original onboarding prompts during retries.
  * Avoided contradictory failure messages when profile publication status is uncertain.

* **Tests**
  * Expanded coverage for retries, deduplication, delivery recovery, media handling, and onboarding workflows.


Datawav (commented):
## Automated review

No actionable findings.

Datawav (commented):
## Automated review

1 actionable finding.

1. Major — A committed profile publication can produce contradictory success and failure messages
- Exact changed-line evidence: `integrations/openclaw/marmot/src/profile-onboarding.ts:495` keeps `sendOnboardingFinal(..., responseKind: "profile-published")` inside the same `try` that covers `accountPublishProfile` and `markPublished`; its `catch` sends `PROFILE_NAME_PUBLISH_FAILED` using the distinct `"profile-publish-failed"` binding.
- Triggering state: Profile publication and state persistence succeed; the success confirmation is durably published, but `send_final` returns an ambiguous retryable failure after that commit.
- Consequence: The catch emits a second encrypted event under a different idempotency key, so connector deduplication cannot collapse it with the committed success. The user receives both “published” and “publish failed,” while persisted state already records publication.
- Confidence: High. The control flow and distinct key bindings directly establish the failure; the added ambiguous-success tests do not cover the published-confirmation path.
- Smallest sound correction: Restrict the `try/catch` to profile publication and `markPublished`. Send the success confirmation afterward and propagate its delivery failure so retries reuse the `"profile-published"` key. Add an ambiguous-success test that loses the response after the published confirmation commits and asserts one encrypted event and no failure response.

integrations/openclaw/marmot/src/outbound.ts
```diff
@@ -338,11 +339,26 @@ export function createMarmotMessageAdapter(deps: MarmotMessageAdapterDeps) {
     send: {
       text: async (ctx: ChannelMessageSendTextContext) => {
         const { client, marmotAccountIdHex } = await deps.resolveTarget(ctx.cfg, ctx.accountId);
+        const durableIntentId = ctx.deliveryQueueId;
```
erskingardner: This makes `deliveryQueueId` mandatory for every text send. In the SDK type it is accessed as an ordinary property, and the tests bypass typing via `as unknown as ChannelMessageSendTextContext`, so nothing compile-time guarantees hosts populate it on every send path. If any current-or-future OpenClaw release channel invokes `send.text` without it (the plugin AGENTS.md explicitly warns SDK surfaces churn between channels), every outbound text now throws instead of degrading. Can you cite where the SDK contract guarantees `deliveryQueueId` on all durable text sends? If it cannot be guaranteed, consider falling back to a non-keyed send (connector already treats an absent key as today's behavior) rather than failing closed for the whole channel.

integrations/openclaw/marmot/src/profile-onboarding.ts
```diff
@@ -343,6 +346,55 @@ interface OnboardingLogger {
...
+    responseKind?: OnboardingResponseKind;
+  },
+): Promise<void> {
+  const replyToMessageIdHex = input.replyToMessageIdHex ?? null;
+  const replyTurnBinding = replyToMessageIdHex == null
+    ? null
+    : canonicalOnboardingHex(replyToMessageIdHex);
+  const turnBinding = input.responseKind
+    ? `profile-response:v1:${replyTurnBinding}:${input.responseKind}`
+    : "profile-prompt:v1";
```
erskingardner, line 429: Edge case: the prompt turnBinding is a constant, but the content fingerprint (and therefore the key, since the fingerprint feeds the preimage) includes the suggested name. If the suggested name changes between an ambiguous success and its retry (e.g. config change), the retry derives a *new* key and can double-prompt after a duplicate slot reclaim. Probably acceptable given the prompt slot is claimed atomically, but worth confirming that is intentional.

integrations/openclaw/marmot/src/durable-final-idempotency.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+export function deriveDurableFinalIdempotency(
+  input: DurableFinalIdempotencyInput,
+): DurableFinalIdempotency {
+  const sessionBinding = requireNonEmpty(input.sessionBinding, "sessionBinding");
+  const turnBinding = requireNonEmpty(input.turnBinding, "turnBinding");
+  const accountIdHex = normalizeHex(input.accountIdHex, "accountIdHex");
+  const groupIdHex = normalizeHex(input.groupIdHex, "groupIdHex");
+  const replyAnchor = normalizeReplyAnchor(input.replyToMessageIdHex);
+  const text = String(input.text ?? "");
```
erskingardner: Inconsistent strictness: every other field goes through `requireNonEmpty`/`normalizeHex` and fails closed, but `text` silently coerces `undefined`/non-strings via `String(...)`. Either validate it is a string like the rest, or add a comment explaining why text alone is lenient.

erskingardner (commented):
**Review summary: PASS with concerns — nothing blocking in the idempotency implementation itself, but please address the scope split and the `deliveryQueueId` contract question before merge.**

What looks good:
- The key derivation is deterministic, namespaced, versioned, and well-covered: normative vector test, fail-closed identity test, connector-faithful ambiguous-success dedup tests that were verified red on master.
- The agent-control contract stays optional-key; no protocol change.
- The SHA-pinned cargo-audit job is green at this head.

Concerns:
1. **Scope**: this mixes three unrelated changes (OpenClaw idempotency feature, Dependabot config, RustSec audit CI) under a title that only describes the first. These are separable; the audit gate in particular has its own coordination requirements in #1220/#1218. Consider splitting the CI/Dependabot pieces out so the feature fix reviews and reverts independently.
2. **Fail-closed `deliveryQueueId`**: see inline comment — this converts any future host path that omits the field from a degraded send into a hard runtime failure of all text sends, with no compile-time signal.
3. Minor items inline (prompt-key vs suggested-name edge case, lenient `text` coercion). The hex-normalization duplication you already track as an explicit follow-up — fine to leave.

integrations/openclaw/marmot/src/outbound.ts
Datawav: Confirmed; this is stronger than a future-channel risk. In the pinned `openclaw@2026.7.1-2` SDK, `ChannelMessageSendTextContext` declares `deliveryQueueId?: string`. The runtime only forwards the queue id to the adapter when exact unknown-send reconciliation is required for a single payload, while this adapter does not advertise or implement `reconcileUnknownSend`. The new unit tests inject the field manually through `as unknown as`, so they do not prove the host path supplies it.

Suggested resolution: do not hard-fail when the field is absent; preserve the existing non-keyed send on that path, and only claim durable retry deduplication for a host path that supplies a contracted stable intent id.

Datawav:
Verified at `27c4bd3c64d80550ded5c59566430020a0736010`: the scope concern is valid. `.github/dependabot.yml` and the Rust audit job are independently reviewable changes with separate coordination history, while the PR title and core tests concern OpenClaw final-send idempotency. Suggested resolution is to split those two changes from this PR so the feature can be reviewed and reverted independently. The three inline concerns are also valid; I’ve replied with the source evidence and suggested disposition.

integrations/openclaw/marmot/src/profile-onboarding.ts
Datawav: Confirmed; this is not a safe intentional edge case. `sendProfilePrompt` clears the claimed slot after any send error, including an ambiguous post-success disconnect, then a retry recomputes the suggested name from live configuration. Because the prompt text feeds the fingerprint, a changed configured name yields a different key and can publish a second prompt. The current ambiguous-success test retries with the same `configuredName: null`, so it does not cover this case.

Suggested resolution: persist and reuse the originally claimed prompt snapshot (or its stable fingerprint) across slot reclamation before treating prompt retries as deduplicated.

integrations/openclaw/marmot/src/durable-final-idempotency.ts
Datawav: Agreed. `text` should be runtime-validated rather than coerced. Since an empty string may still be a valid hashed value, a dedicated string-type check is preferable to `requireNonEmpty`: reject non-strings, preserve `""`, then run the existing unpaired-surrogate validation. That keeps the hashing contract explicit without silently converting `undefined`, objects, or numbers.

Datawav (commented):
## Automated review

Audit outcome: FAIL — 4 actionable findings remain at head `27c4bd3c64d80550ded5c59566430020a0736010`.

1. Blocking — Text sends fail when the host omits an optional delivery intent ID
- Evidence: `integrations/openclaw/marmot/src/outbound.ts:342-347` reads `ctx.deliveryQueueId` and throws when it is absent. The bounded SDK context declares this field optional; the new tests supply it manually rather than exercising the ordinary host path.
- Trigger: OpenClaw invokes `send.text` without `deliveryQueueId`.
- Consequence: the adapter rejects the outbound text before calling `sendFinal`, regressing normal message delivery.
- Confidence: High.
- Smallest sound correction: preserve the existing non-keyed `sendFinal` path when no stable intent ID is supplied; derive an idempotency key only when the host provides one under a supported contract.

2. Major — Ambiguous confirmation success produces contradictory onboarding replies
- Evidence: `integrations/openclaw/marmot/src/profile-onboarding.ts:495` sends the keyed `profile-published` confirmation inside the same `try` that catches publication and persistence failures; the catch sends `profile-publish-failed` under a different key.
- Trigger: profile publication and `markPublished` succeed, the confirmation is committed, and its response is lost through a retryable disconnect.
- Consequence: the committed success remains visible, then the catch publishes a failure message; persisted `published` state prevents a later inbound retry from restoring a single coherent outcome.
- Confidence: High.
- Smallest sound correction: separate publication/persistence errors from confirmation-delivery errors. Persist a pending success-confirmation intent and retry only that confirmation with its original key; never emit `profile-publish-failed` after publication has succeeded.

3. Major — A changed configured name defeats prompt deduplication after ambiguous success
- Evidence: changed line `integrations/openclaw/marmot/src/profile-onboarding.ts:380` derives the prompt key from live prompt text, while the surrounding failure path clears the claimed prompt slot after any send error. Existing retry tests reuse the same configured name and do not cover this boundary.
- Trigger: the connector commits the first prompt but disconnects, the slot is cleared, configuration changes the suggested name, and onboarding retries.
- Consequence: the regenerated text has a different fingerprint and key, allowing a second onboarding prompt to be published.
- Confidence: High.
- Smallest sound correction: persist the originally claimed prompt text and key before sending, then reuse that immutable snapshot across ambiguous retries; add a regression test that changes the configured name between attempts.

4. Minor — The key derivation boundary silently coerces invalid text values
- Evidence: `integrations/openclaw/marmot/src/durable-final-idempotency.ts:78` uses `String(input.text ?? "")`, unlike the explicit runtime validation applied to the other key fields.
- Trigger: an untyped or malformed caller supplies `undefined`, a number, or an object as `text`.
- Consequence: invalid values hash as `""`, `"123"`, or `"[object Object]"`, permitting silent key collisions or divergence instead of rejecting the malformed request.
- Confidence: High.
- Smallest sound correction: require `typeof input.text === "string"`, retain empty-string support, and then apply the existing unpaired-surrogate validation.

integrations/openclaw/marmot/src/durable-final-idempotency.ts
```diff
@@ -0,0 +1,97 @@
+import { createHash } from "node:crypto";
+
+const IDEMPOTENCY_KEY_PREFIX = "marmot-final-v1:";
```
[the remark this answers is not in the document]
Datawav: I left this thread unresolved. The current control protocol has no shared API that accepts the durable session/turn bindings needed to derive this retry key, so moving the digest now would require a cross-component protocol change rather than a bounded review fix. This SHA-256 use is a deterministic, non-secret request fingerprint; it does not perform Marmot cryptography or handle key material. The valid normalization duplication was removed by reusing the control client’s canonical hex helper. A shared wn-agent/agent-control derivation API should be designed separately before relocating this logic.

integrations/openclaw/marmot/src/outbound.ts
Datawav: Fixed in ae2629d5. `deliveryQueueId` remains optional: when present it derives the stable connector key, and when absent the adapter preserves the legacy non-keyed `sendFinal` path instead of failing the channel. The regression test covers both paths.

integrations/openclaw/marmot/src/profile-onboarding.ts
Datawav: Fixed in ae2629d5. A claimed prompt is no longer cleared after an ambiguous send result. Later triggers reuse the persisted group and suggested-name snapshot, so a live configuration change cannot alter the prompt text or durable key; cross-group triggers leave that claim untouched. Regression coverage exercises both cases.

integrations/openclaw/marmot/src/durable-final-idempotency.ts
Datawav: Fixed in ae2629d5. Durable-final derivation now requires `text` to be a string, still permits the empty string, and rejects non-string runtime inputs instead of coercing them. The regression test locks both behaviors.

Datawav (commented):
## Automated review

Actionable findings remain at exact head ae2629d51cca15d15d412639bfcebdd1b6dfe711.

1. Blocking — Failed prompt delivery still enables consent handling
- Evidence: `integrations/openclaw/marmot/src/profile-onboarding.ts`, changed `sendProfilePrompt` failure path and the `record.status === "prompted"` reply branch. After `tryClaimPrompt(...)` records `{status: "prompted", group_id_hex, suggested_name}`, a rejected `sendOnboardingFinal` returns `false` without clearing or otherwise marking the prompt undelivered. The inbound branch subsequently treats that same `prompted` record as awaiting consent and routes affirm/name replies to `publishAndConfirm`.
- Triggering state: the prompt send fails before delivery or returns an ambiguous error, then any later same-group message parses as affirm, skip, invalid, or a profile name.
- Consequence: ordinary conversation can be consumed as onboarding; an affirming or name-like message can publish a public kind:0 profile although the consent prompt was never delivered.
- Confidence: High. The changed test “preserves the claimed prompt snapshot if the send result is ambiguous” explicitly preserves `prompted` after `failSend: true` but does not establish delivery.
- Smallest sound correction: represent “reserved retry snapshot” separately from “prompt delivered and awaiting reply.” Retain the stable group/name snapshot after an ambiguous send, but permit inbound consent handling only after successful or deduplicated prompt delivery; otherwise retry the prompt.

2. Major — Successful kind:0 publication can remain in a live prompted state
- Evidence: `integrations/openclaw/marmot/src/profile-onboarding.ts:535` in the changed `publishAndConfirm` persistence-retry path. `accountPublishProfile(...)` commits first; two `markPublished(...)` failures are swallowed and the published confirmation is still attempted. The later inbound guard exits only for `profile_exists`, `published`, or `skipped`, leaving the persisted `prompted` record eligible for another reply.
- Triggering state: kind:0 publication succeeds, both `markPublished` attempts fail, storage later recovers, and another same-group message arrives.
- Consequence: a later affirm/name can publish kind:0 again, potentially with a different name; “skip” can also overwrite local onboarding state despite an already-live profile.
- Confidence: High. The new test “still confirms publication after published-state persistence fails twice” exercises the exact non-terminal state, and the head source retains the prompted reply path.
- Smallest sound correction: after successful publication, prevent further prompted processing independently of confirmation delivery. On prompted inbound, check for an existing live profile and persist/repair terminal `published` state before returning without consuming the message.

3. Minor — Prompt retry identity compares non-canonical group IDs
- Evidence: `integrations/openclaw/marmot/src/profile-onboarding.ts`, changed same-group retry gate uses `current.group_id_hex !== deps.groupIdHex`, while the same head canonicalizes account/group IDs with `normalizeHex` for onboarding idempotency and session binding. `tryClaimPrompt(...)` still receives the caller’s group string.
- Triggering state: the claim stores an unprefixed/lowercase group ID and a retry supplies the equivalent ID with a `0x` prefix or different case.
- Consequence: the valid same-group retry is rejected, leaving onboarding indefinitely stuck on the retained snapshot.
- Confidence: High. Equivalent prefixed/unprefixed IDs are explicitly treated as identical by the newly added connector/idempotency tests, but this retry comparison remains raw.
- Smallest sound correction: canonicalize the group ID before persistence and compare canonical forms with `normalizeHex`.

Audit evidence:
- Exact head: ae2629d51cca15d15d412639bfcebdd1b6dfe711
- Changed files inspected include `integrations/openclaw/marmot/src/profile-onboarding.ts`, `integrations/openclaw/marmot/test/profile-onboarding.test.ts`, `integrations/openclaw/marmot/test/ambiguous-success-idempotency.test.ts`, `integrations/openclaw/marmot/src/outbound.ts`, and `.github/dependabot.yml`.
- Discussion classification: deliveryQueueId fallback, runtime text validation, session-binding normalization, connector fingerprint canonicalization, and contradictory publish-failed confirmation claims are fixed at this head; SHA-256 relocation is an out-of-scope architectural request. The second-failure discussion remains valid and is represented once above.

Risk lenses checked:
- release/build — untrusted-input and least-privilege CI; artifact identity/reproducibility; all required paths trigger checks.
- tests — assertions fail before the fix; boundary and negative cases; test exercises production behavior.

Check evidence:
- Bundle coverage is complete for all nine changed files.
- All reported checks completed successfully under the bundle’s aggregate result. OpenClaw beta host compatibility, OpenClaw and Hermes packaging, four Rust test shards, Rust check/clippy/format/audit, CLI and relay E2E/runtime tests, simulator/vectors, Tamarin proofs, installer tests, and platform builds succeeded; the conditional “Publish wn-agent release” job was skipped.

integrations/openclaw/marmot/src/outbound.ts
```diff
@@ -372,11 +373,26 @@ export function createMarmotMessageAdapter(deps: MarmotMessageAdapterDeps) {
...
```
erskingardner: Blocking: this field is not populated on the production message-adapter path. In pinned `openclaw@2026.7.1-2`, `deliveryQueueId` is internal and `deliverOutboundPayloadsWithQueueCleanup` forwards it only when exact unknown-send reconciliation is required; that path requires the adapter to advertise `durableFinal.capabilities.reconcileUnknownSend` and provide the reconciliation callback. This adapter declares neither. Normal OpenClaw delivery therefore reaches this code with `undefined` and calls `sendFinal` without a key. The tests inject `deliveryQueueId` via `as unknown as ChannelMessageSendTextContext`, so they cannot catch this. Wire a host-supported durable identity/reconciliation path and add coverage through the normal OpenClaw delivery entry point; otherwise ordinary durable finals remain duplicate-prone.

integrations/openclaw/marmot/src/profile-onboarding.ts
```diff
@@ -358,34 +408,46 @@ async function sendProfilePrompt(deps: {
...
   try {
-    await deps.client.sendFinal(deps.accountIdHex, deps.groupIdHex, buildProfilePrompt(suggested));
+    await sendOnboardingFinal(deps.client, {
+      accountIdHex: deps.accountIdHex,
+      groupIdHex: deps.groupIdHex,
+      text: buildProfilePrompt(suggested),
+    });
     deps.logger?.info?.("marmot: profile onboarding prompt sent");
     return true;
   } catch {
```
erskingardner, line 501: Blocking: `tryClaimPrompt` persists `status: "prompted"` before connector I/O, and this catch retains that state for every rejection, including a failure before the prompt was delivered. `maybeHandleProfileOnboardingInbound` then treats the record as an awaiting-consent prompt and consumes the next same-group text, so an ordinary `yes` or name-like message can publish a public kind:0 profile even though the user never received the consent prompt. Keep a distinct pending/retry snapshot, retry the same keyed prompt on inbound, and only enter the consent-handling state after a successful or deduplicated send.

integrations/openclaw/marmot/src/profile-onboarding.ts
```diff
@@ -418,25 +480,60 @@ async function publishAndConfirm(
...
+  try {
+    await store.markPublished(message.accountIdHex, name);
+  } catch {
+    // Publication has already succeeded. Retry only the local bookkeeping; a
+    // storage failure must never turn a live profile into a contradictory
+    // profile-publish-failed response.
+    try {
+      await store.markPublished(message.accountIdHex, name);
+    } catch {
+      logger?.warn?.("marmot: profile published but local state not recorded");
```
erskingardner: Blocking: after `accountPublishProfile` commits, two `markPublished` failures are swallowed while the durable record remains `prompted`. Once storage recovers, the next same-group message re-enters onboarding and can publish another kind:0 (possibly with a different name), or `skip` can overwrite local state despite the live profile. The new test explicitly leaves this non-terminal state. Repair or durably guard the terminal published state before allowing prompted input to be handled again; a live-profile check on recovery is one possible safety net.

erskingardner (changes requested):
Request changes at exact head d66da7a2. The normal OpenClaw outbound path still does not receive the stable intent needed by the headline fix, and the onboarding state machine can both publish without a delivered consent prompt and repeat a successful profile publication after bookkeeping failure. All reported CI checks are green, but the new unit tests bypass the host durable-send path or assert the unsafe persisted states.

Please also move `.github/dependabot.yml` to its own PR. It is independent of final-send idempotency, changes ongoing repository maintenance behavior, and the scope split was already acknowledged in this discussion.

integrations/openclaw/marmot/src/outbound.ts
```diff
@@ -357,6 +359,73 @@ export function createMarmotMessageAdapter(deps: MarmotMessageAdapterDeps) {
...
+      };
+    }
+    const result = await sendDurableText({
+      cfg: ctx.cfg,
+      accountId: ctx.accountId,
+      to: ctx.to,
+      text: payload.text,
+      replyToId: ctx.effectiveReplyToId ?? ctx.replyToId ?? payload.replyToId ?? null,
+      durableIntentId: ctx.queueId,
+    });
```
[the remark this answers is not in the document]
erskingardner: Verified against the pinned OpenClaw 2026.7.1-2 host implementation. The original adapter send receives the resolved route as ctx.replyToId; OpenClaw then persists that route as effectiveReplyToId for unknown-send recovery. The queued replyToId can legitimately differ, so reconciliation must prefer effectiveReplyToId as implemented. Commit 756d5108 adds a regression with different queued and effective anchors and proves reconciliation reuses the original send key; the installed-host test also now exercises a required durable public send and verifies the host queue identity reaches send_final.

erskingardner (approved):
Re-reviewed at 756d510804138d0b20b226c3c9fb940d29b82311. The durable queue identity now reaches normal host sends, failed prompt and profile-publication transitions are guarded and recoverable, onboarding identifiers are canonicalized, the unrelated Dependabot change is removed, and the remaining reply-anchor concern is covered by a pinned-host regression. Local OpenClaw suites, beta host compatibility, development scripts, and just fast-ci all pass.

PR #1519, opened 2026-08-20
Fix/arm refusals before publish gate

jgmontoya:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery during scheduled and manually retried synchronization convergence.
  * Publish failures are now reported consistently while durable epoch backfill recovery is activated.
  * Helps restore synchronization after resource refusals or rolled-back updates without requiring repeated recovery actions.
  * Improved handling ensures pending group updates, projections, notifications, and system state remain consistent during recovery.


Datawav (commented):
## Automated review

No actionable findings.

PR #1521, opened 2026-08-20
marmot-app: invite KeyPackages the caller already holds

lopatin96:
Member resolution maps one account to one KeyPackage — whichever it published last. An account with two devices therefore joins a group as one of them, decided by a race between their publishes, and the other never learns the group exists. Naming the account twice does not help: `resolve_member_key_packages_inner` deduplicates by account id, so the second mention resolves to the same package.

Group setup already says what the result should look like:

> Admin authority is based on Marmot account identity, not on MLS leaf id. If an account has multiple leaves in a group, the admin policy applies to each current leaf.

So more than one leaf per account is a shape the protocol describes. What is missing is a way for a caller to ask for it.

### The change

`AppClient::invite_key_packages` and `MarmotAppRuntime::invite_key_packages` take the packages rather than resolving them. Everything after that is the existing invite path — the same commit, the same Welcomes, the same delivery records — so the only new behaviour is which packages end up in `SendIntent::Invite`.

Resolution is untouched for every existing caller: `invite_members` still resolves as before, and the shared body simply skips the lookup when packages are supplied.

This is deliberately **not** the multi-device External Commit flow from `features/multi-device.md`, which is a draft and needs the signalling gate and an admin-authorization proof. This is an ordinary invitation of packages the caller found for itself, which is how somebody discovers a contact's devices today anyway.

### Test

`invite_key_packages_admits_two_leaves_of_one_account` in `crates/marmot-app/tests/relay_runtime.rs` puts one account on two stores — separate MLS key material in each, which is what a second device is — publishes a KeyPackage from both, and invites the two together. Three leaves result, two of them under one identity; both devices join; a message from the group reaches both.

Inviting the same account by name in that scenario gives two leaves instead of three, which is the behaviour this addresses.

### Caveats

- Finding the devices is left to the caller. This does not add a way to enumerate an account's published KeyPackages; the packages must carry their source event id, since a Welcome is addressed with it — one read from a relay does that, a package read out of local storage does not.
- I did not touch how many packages an account may reasonably have in one group, nor what should happen to leaves of a device that has been retired. Both feel like questions for the multi-device feature rather than for this seam.
- `cargo test -p marmot-app` on this branch: 850 passed, 5 failed. The same five fail on a clean `master` here (`encrypted_media_*`, `retained_media_rehydrates_*`, `relay_app_runtime_synthesizes_system_row_for_retention_change`, `self_removal_suppresses_*`), all with timeouts, so they look environmental rather than related.

## Summary by CodeRabbit

- **New Features**
  - Added support for inviting multiple devices to a group using explicitly selected invitation credentials.
  - Added runtime and account-level options for sending invitations with supplied credentials.
  - Group state and pending updates now refresh after invitations are successfully published.

- **Bug Fixes**
  - Improved multi-device account support so each device can join groups and receive subsequent messages.

- **Tests**
  - Added coverage for multi-device invitations, synchronized group membership, and message delivery.

### DIFF
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -1865,6 +1865,24 @@ impl AppClient {
         Ok(summary)
     }
 
+    /// The same, for the account worker: timed like every other invite, and leaving Welcome
+    /// delivery to the worker's own fanout rather than driving it inline.
+    pub(crate) async fn invite_key_packages_with_telemetry(
+        &mut self,
+        group_id: &GroupId,
+        key_packages: Vec<KeyPackage>,
+        telemetry: &AppPerformanceTelemetry,
+    ) -> Result<SendSummary, AppError> {
+        self.invite_members_with_optional_telemetry(
+            group_id,
+            &[],
+            &[],
+            Some(telemetry),
+            Some(key_packages),
+        )
+        .await
+    }
+
     async fn invite_members_with_optional_telemetry(
         &mut self,
         group_id: &GroupId,
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -1824,7 +1824,7 @@ impl AppClient {
         member_refs: &[&str],
     ) -> Result<SendSummary, AppError> {
         let summary = self
-            .invite_members_with_optional_telemetry(group_id, member_refs, &[], None)
+            .invite_members_with_optional_telemetry(group_id, member_refs, &[], None, None)
             .await?;
         self.drive_unpublished_welcome_delivery(None).await;
         Ok(summary)
@@ -1842,8 +1842,27 @@ impl AppClient {
             member_refs,
             initial_admin_refs,
             Some(telemetry),
+            None,
         )
         .await
+    }
+
+    /// Invites members whose KeyPackages the caller already holds.
+    ///
+    /// Resolution maps one account to one KeyPackage, which is the right answer only while an
+    /// account has one device. A caller that knows an account's devices — and which of them
+    /// still count — supplies their KeyPackages here instead. Group setup already defines the
+    /// resulting shape: admin policy applies per leaf when an account has more than one.
+    pub async fn invite_key_packages(
+        &mut self,
+        group_id: &GroupId,
+        key_packages: Vec<KeyPackage>,
+    ) -> Result<SendSummary, AppError> {
+        let summary = self
+            .invite_members_with_optional_telemetry(group_id, &[], &[], None, Some(key_packages))
+            .await?;
+        self.drive_unpublished_welcome_delivery(None).await;
+        Ok(summary)
     }
 
     async fn invite_members_with_optional_telemetry(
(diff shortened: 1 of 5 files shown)


lopatin96:
Both were right, and both were the same mistake: the new path drifted from the one beside it.

The worker arm now gates the projection and group-state announcements on `accept_disposition == Published` rather than on `Ok`, matching the `InviteMembers` arm — a queued send returns `Ok` without having been published, and announcing group state for it announces something that has not happened.

There is a telemetry-taking variant for the worker now, mirroring `invite_members_with_telemetry`, so this invite is timed like the others instead of being invisible to the performance counters. It also leaves Welcome delivery to the account worker's own fanout rather than driving it inline; only the public method does that, exactly as `invite_members` does.

And the command spawns the post-commit catch-up instead of awaiting it, so a caller is not held for work the other invite path detaches.

On the audit action label: I left it as `invite_members`, since what is recorded is an invite either way and the distinction is in how the packages were found rather than in what happened to the group. Happy to split it if you would rather see the two apart in the log.

`cargo test -p marmot-app`: 850 passed, 5 failed — the same five that fail on a clean `master` here, all timeouts.

Datawav (commented):
## Automated review

2 actionable findings.

1. Blocking — Runtime KeyPackage invites do not deliver retained Welcomes
- Evidence: `crates/marmot-app/src/runtime/account_worker.rs:3013-3046` adds the `InviteKeyPackages` arm, calls `invite_key_packages_with_telemetry`, conditionally publishes state, and ends with `respond.send(result)` without calling `drive_unpublished_welcome_delivery`. The adjacent `InviteMembers` arm performs that post-response Welcome-delivery lifecycle.
- Trigger: `MarmotAppRuntime::invite_key_packages` publishes an invite through the account worker.
- Consequence: the operation can return `Published` while the Welcome gift-wraps remain retained, preventing invited devices from joining until unrelated later work happens to drive delivery.
- Confidence: High.
- Smallest sound correction: after sending the response, mirror the `InviteMembers` post-response path: capture the read snapshot, call `drive_unpublished_welcome_delivery(Some(&telemetry))`, and publish pending Welcome-delivery events through `serve_snapshot_reads_until`. Add a runtime-level test that invokes `MarmotAppRuntime::invite_key_packages` and proves both devices receive a subsequent message.

2. Minor — Runtime KeyPackage invites omit caller-latency telemetry
- Evidence: the new `AccountManager::invite_key_packages` method at `crates/marmot-app/src/runtime/commands.rs:1615-1638` sends the command and awaits `account_worker_response(response)` without timing or recording `AppPerformanceOperation::GroupInviteMembers`. The sibling account-reference invite records that operation; explicit packages also intentionally bypass KeyPackage-lookup timing.
- Trigger: any invite through `MarmotAppRuntime::invite_key_packages`.
- Consequence: runtime invite latency is absent from both established invite measurements, obscuring regressions in this path.
- Confidence: High.
- Smallest sound correction: time the worker round trip and record `GroupInviteMembers`, or introduce a dedicated KeyPackage-invite operation, matching the sibling runtime invite wrapper.

PR #1533, opened 2026-08-21
Harden WN Agent installer verification and RustSec CI

Datawav:
## Summary

- replace active WN Agent `curl | bash` guidance with fail-closed download, sibling `.sha256` verification, and execution of only the verified local installer
- preserve the canonical guided onboarding and finish-in-White-Noise path while covering Hermes, OpenClaw, Codex, OpenCode, and Pi release installs
- treat `wn-agent-latest` as an endorsed rolling convenience alias while keeping versioned release tags as the immutable install path
- add a pinned RustSec audit job that installs the repository Rust toolchain and `cargo-audit@0.22.1`, then runs exact `cargo --locked audit` at the repository root
- add policy and mutation tests so future documentation or workflow changes cannot silently bypass either gate

## Acceptance criteria

- [x] Active install paths download the installer and its published `.sha256` companion
- [x] Documented helpers fail closed on download or checksum failure before invoking `bash`
- [x] Installer `--help`, release guidance, and generated release notes use verify-then-execute examples
- [x] Current guided phone onboarding, bootstrap-file permissions, and canonical quickstart navigation are preserved
- [x] CI installs the repository Rust toolchain and pinned `cargo-audit@0.22.1`
- [x] CI runs exact `cargo --locked audit` at repository root with no error masking
- [x] New unignored RustSec advisories fail the audit job while `.cargo/audit.toml` remains authoritative

## Disposition evidence

- Dependabot coverage for Cargo, Actions, and the OpenClaw pnpm lockfile is intentionally not duplicated here; it remains in [#1516](https://github.com/marmot-protocol/mdk/pull/1516).
- A historical `curl | bash` mention in `crates/cli/CHANGELOG.md` and the unrelated uv bootstrap in `scripts/hermes_marmot_dev_setup.sh` are pre-existing, baseline-capped debt outside active WN Agent install guidance.
- No F2 or F3 implementation track was skipped: verified installer guidance and RustSec CI are both included.

## Verification

- `just fast-ci`
- install-example SHA-256 policy and 11 regression/mutation tests
- cargo-audit workflow policy and 13 regression/mutation tests
- current-release agent install documentation gate
- workspace checks and clippy, normal and OTLP feature sets
- convergence policy pin tests

The local worktree does not have `cargo-audit` installed, so the audit binary itself was not re-run after rebasing. `Cargo.lock` is unchanged from the previously audited target; the GitHub Rust audit job installed the exact audit binary and repository toolchain and passed against the locked dependency graph.

## Independent review

PASS on exact follow-up head `0e8dd011e6ebb5042e45e0bee79d0eef5127fc7e`. The bounded independent runner selected Cursor `grok-4.6`; this is **not Kimi-reviewed**. It confirmed the exact RustSec policy, always-running audit job, every documented installer surface, semantic rolling-alias checks, and structural same-shell guidance; its remaining findings were non-blocking hardening notes.

## CI

- [CI run](https://github.com/marmot-protocol/mdk/actions/runs/32828443502)
- [Rust audit job](https://github.com/marmot-protocol/mdk/actions/runs/32828443502/job/97741620570)
- [WN Agent binaries run](https://github.com/marmot-protocol/mdk/actions/runs/32828443504)

All required checks passed on exact head `0e8dd011e6ebb5042e45e0bee79d0eef5127fc7e`. Release publication was correctly skipped for this pull request.

### DIFF
--- a/.github/workflows/wn-agent-binaries.yml
+++ b/.github/workflows/wn-agent-binaries.yml
@@ -684,7 +684,7 @@ jobs:
           fi
 
           cat > "$latest_notes" <<EOF
-          Latest WN Agent installers. These scripts install WN Agent \`$version\` from immutable release \`$tag\`.
+          Latest WN Agent installers. These scripts install whichever version the mutable \`$latest_tag\` alias currently references; use immutable release \`$tag\` for WN Agent \`$version\`.
 
           \`\`\`sh
           set -eu
--- a/.github/workflows/wn-agent-binaries.yml
+++ b/.github/workflows/wn-agent-binaries.yml
@@ -478,14 +478,82 @@ jobs:
           cat > "$notes" <<EOF
           WN Agent \`$version\` from MDK commit \`$GITHUB_SHA\`.
 
+          Install the latest WN Agent release with:
+
+          \`\`\`sh
+          set -eu
+          base_url="https://github.com/$GITHUB_REPOSITORY/releases/download/wn-agent-latest"
+
+          install_verified() (
+            set -eu
+            installer_url="\$1"
+            checksum_url="\$2"
+            shift 2
+            installer_script="\${installer_url##*/}"
+            tmpdir="\$(mktemp -d)"
+            trap 'rm -rf "\$tmpdir"' 0 HUP INT TERM
+            curl -fsSL "\$installer_url" -o "\$tmpdir/\$installer_script"
+            curl -fsSL "\$checksum_url" -o "\$tmpdir/\$installer_script.sha256"
+            if command -v shasum >/dev/null 2>&1; then
+              (cd "\$tmpdir" && shasum -a 256 -c "\$installer_script.sha256")
+            elif command -v sha256sum >/dev/null 2>&1; then
+              (cd "\$tmpdir" && sha256sum -c "\$installer_script.sha256")
+            else
+              echo "error: need shasum or sha256sum to verify the installer" >&2
+              exit 1
+            fi
+            bash "\$tmpdir/\$installer_script" "\$@"
+          )
+
+          # Hermes gateway
+          install_verified "\$base_url/install-hermes-marmot.sh" "\$base_url/install-hermes-marmot.sh.sha256"
+          # OpenClaw gateway
+          install_verified "\$base_url/install-openclaw-marmot.sh" "\$base_url/install-openclaw-marmot.sh.sha256"
+          # Codex terminal harness
+          install_verified "\$base_url/install-codex-marmot.sh" "\$base_url/install-codex-marmot.sh.sha256"
+          # OpenCode terminal harness
+          install_verified "\$base_url/install-opencode-marmot.sh" "\$base_url/install-opencode-marmot.sh.sha256"
+          # Pi terminal harness
+          install_verified "\$base_url/install-pi-marmot.sh" "\$base_url/install-pi-marmot.sh.sha256"
+          \`\`\`
+
           Install this exact WN Agent release with:
 
           \`\`\`sh
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-hermes-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-openclaw-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-codex-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-opencode-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-pi-marmot.sh | bash
+          set -eu
+          base_url="https://github.com/$GITHUB_REPOSITORY/releases/download/$tag"
+
+          install_verified() (
+            set -eu
+            installer_url="\$1"
+            checksum_url="\$2"
+            shift 2
+            installer_script="\${installer_url##*/}"
+            tmpdir="\$(mktemp -d)"
+            trap 'rm -rf "\$tmpdir"' 0 HUP INT TERM
+            curl -fsSL "\$installer_url" -o "\$tmpdir/\$installer_script"
+            curl -fsSL "\$checksum_url" -o "\$tmpdir/\$installer_script.sha256"
+            if command -v shasum >/dev/null 2>&1; then
+              (cd "\$tmpdir" && shasum -a 256 -c "\$installer_script.sha256")
+            elif command -v sha256sum >/dev/null 2>&1; then
+              (cd "\$tmpdir" && sha256sum -c "\$installer_script.sha256")
+            else
+              echo "error: need shasum or sha256sum to verify the installer" >&2
+              exit 1
+            fi
+            bash "\$tmpdir/\$installer_script" "\$@"
+          )
+
+          # Hermes gateway
+          install_verified "\$base_url/install-hermes-marmot.sh" "\$base_url/install-hermes-marmot.sh.sha256"
+          # OpenClaw gateway
+          install_verified "\$base_url/install-openclaw-marmot.sh" "\$base_url/install-openclaw-marmot.sh.sha256"
+          # Codex terminal harness
+          install_verified "\$base_url/install-codex-marmot.sh" "\$base_url/install-codex-marmot.sh.sha256"
+          # OpenCode terminal harness
+          install_verified "\$base_url/install-opencode-marmot.sh" "\$base_url/install-opencode-marmot.sh.sha256"
+          # Pi terminal harness
+          install_verified "\$base_url/install-pi-marmot.sh" "\$base_url/install-pi-marmot.sh.sha256"
           \`\`\`
 
           Assets:
@@ -619,11 +687,40 @@ jobs:
           Latest WN Agent installers. These scripts install WN Agent \`$version\` from immutable release \`$tag\`.
 
           \`\`\`sh
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-hermes-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-openclaw-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-codex-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-opencode-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-pi-marmot.sh | bash
+          set -eu
+          base_url="https://github.com/$GITHUB_REPOSITORY/releases/download/$latest_tag"
+
+          install_verified() (
+            set -eu
+            installer_url="\$1"
+            checksum_url="\$2"
+            shift 2
+            installer_script="\${installer_url##*/}"
+            tmpdir="\$(mktemp -d)"
+            trap 'rm -rf "\$tmpdir"' 0 HUP INT TERM
+            curl -fsSL "\$installer_url" -o "\$tmpdir/\$installer_script"
+            curl -fsSL "\$checksum_url" -o "\$tmpdir/\$installer_script.sha256"
+            if command -v shasum >/dev/null 2>&1; then
+              (cd "\$tmpdir" && shasum -a 256 -c "\$installer_script.sha256")
+            elif command -v sha256sum >/dev/null 2>&1; then
+              (cd "\$tmpdir" && sha256sum -c "\$installer_script.sha256")
+            else
+              echo "error: need shasum or sha256sum to verify the installer" >&2
+              exit 1
+            fi
+            bash "\$tmpdir/\$installer_script" "\$@"
+          )
+
+          # Hermes gateway
+          install_verified "\$base_url/install-hermes-marmot.sh" "\$base_url/install-hermes-marmot.sh.sha256"
+          # OpenClaw gateway
+          install_verified "\$base_url/install-openclaw-marmot.sh" "\$base_url/install-openclaw-marmot.sh.sha256"
+          # Codex terminal harness
+          install_verified "\$base_url/install-codex-marmot.sh" "\$base_url/install-codex-marmot.sh.sha256"
+          # OpenCode terminal harness
+          install_verified "\$base_url/install-opencode-marmot.sh" "\$base_url/install-opencode-marmot.sh.sha256"
+          # Pi terminal harness
+          install_verified "\$base_url/install-pi-marmot.sh" "\$base_url/install-pi-marmot.sh.sha256"
           \`\`\`
 
           Use the versioned \`$tag\` release URLs when you need a pinned install.
--- a/.github/workflows/wn-agent-binaries.yml
+++ b/.github/workflows/wn-agent-binaries.yml
@@ -478,7 +478,7 @@ jobs:
           cat > "$notes" <<EOF
           WN Agent \`$version\` from MDK commit \`$GITHUB_SHA\`.
 
-          Install the latest WN Agent release with:
+          Install whichever version the mutable \`wn-agent-latest\` alias currently references with:
 
           \`\`\`sh
           set -eu
--- a/scripts/check_install_example_sha256.py
+++ b/scripts/check_install_example_sha256.py
@@ -43,17 +43,12 @@ DOCUMENTED_INSTALL_CALLS = {
     "integrations/opencode/marmot/README.md": {"install-opencode-marmot.sh": 2},
     "integrations/pi/marmot/README.md": {"install-pi-marmot.sh": 2},
 }
-SAME_SHELL_NOTICE_COUNTS = {
-    "integrations/README.md": 7,
-    "integrations/hermes/marmot/README.md": 3,
-    "integrations/openclaw/marmot/README.md": 2,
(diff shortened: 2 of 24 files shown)


Datawav (commented):
## Automated review

Outcome: 4 actionable findings remain at exact head `2e8f78fde4445e30c7e0a6e9ffbed70d8f6c5bd1`.

1. Major — Versioned release notes direct users to a potentially stale alias
- Evidence: `.github/workflows/wn-agent-binaries.yml:481-485` adds “Install the latest WN Agent release with” and sets `base_url` to `releases/download/wn-agent-latest` in the versioned release-notes heredoc. The same patch removes `release.md:344-346` guidance that the immutable `wn-agent-latest` release is non-authoritative and user-facing commands must name `wn-agent-v<version>`.
- Trigger: `wn-agent-latest` is already immutable or cannot be advanced when a new versioned release is created.
- Consequence: the new release page advertises “latest” while installing older assets from the frozen alias, defeating artifact identity and reproducibility.
- Confidence: High.
- Smallest sound correction: remove the `wn-agent-latest` block from versioned release notes, retain only the `$tag`-based verified example there, and restore the versioned-release rule in `release.md`.

2. Major — The Rust-audit policy gate accepts a disabled job
- Evidence: `scripts/check_cargo_audit_ci.py:51-59` validates the `rust-audit` block’s working directory, `continue-on-error`, and exact `run: cargo --locked audit`, but never rejects a job-level `if:`. The mutation tests in `scripts/test_check_cargo_audit_ci.py` likewise do not exercise a skipped job.
- Trigger: a later workflow edit adds `if: ${{ false }}` or another always-false condition under `rust-audit:` while preserving the strings inspected by the checker.
- Consequence: GitHub Actions skips the audit without failing the check, while the policy test still passes and RustSec advisories are no longer evaluated.
- Confidence: High.
- Smallest sound correction: reject any job-level `if:` in the extracted `rust-audit` block and add a mutation test inserting `if: ${{ false }}` after `timeout-minutes`.

3. Major — The new Codex installer surface is omitted from its verification gate
- Evidence: `integrations/codex/marmot/README.md:34-58` adds a verify-then-execute installer example, but `scripts/check_install_example_sha256.py:37-42` lists only the Hermes, OpenClaw, OpenCode, and Pi README surfaces in `DOCUMENTED_INSTALL_CALLS`. Consequently, `repository_errors()` never applies `documented_surface_errors()` to the changed Codex README.
- Trigger: a future non-pipeline edit removes the Codex example’s checksum download, verification, or verified-local execution while retaining a download-then-`bash` form.
- Consequence: the repository gate exits successfully even though an active documented installer path no longer verifies its published SHA-256 digest.
- Confidence: High.
- Smallest sound correction: add `integrations/codex/marmot/README.md` with an expected call count of one and add a regression mutation that removes its checksum or verification step.

4. Minor — Later installer examples depend on unstated shell state
- Evidence: later changed blocks call `install_verified` without defining it, including `integrations/hermes/marmot/README.md:88-99`, `:101-114`, and `:132-138`; `integrations/openclaw/marmot/README.md:88-95` and `:98-110`; `integrations/opencode/marmot/README.md:58-65`; and `integrations/pi/marmot/README.md:50-56`.
- Trigger: a reader copies one of those later fenced blocks into a fresh shell rather than first executing the earlier helper-definition block.
- Consequence: the command fails with `install_verified: command not found`; installation does not run.
- Confidence: High.
- Smallest sound correction: state immediately before each later block that it must run in the same shell after the helper definition, rather than duplicating the helper.

.github/workflows/wn-agent-binaries.yml
```diff
@@ -478,14 +478,82 @@ jobs:
           cat > "$notes" <<EOF
           WN Agent \`$version\` from MDK commit \`$GITHUB_SHA\`.
 
+          Install the latest WN Agent release with:
+
+          \`\`\`sh
+          set -eu
+          base_url="https://github.com/$GITHUB_REPOSITORY/releases/download/wn-agent-latest"
```
erskingardner, line 485: Question, non-blocking: this PR removes release.md's statement that `wn-agent-latest` "is not an authoritative update channel" and simultaneously starts promoting latest-alias install commands in every generated release's notes. That's a deliberate policy flip toward endorsing the rolling alias — worth a sentence in the PR description so it doesn't read as accidental scope creep. The verify-then-execute hardening itself looks good.

.github/workflows/wn-agent-binaries.yml
```diff
@@ -619,11 +687,40 @@ jobs:
           Latest WN Agent installers. These scripts install WN Agent \`$version\` from immutable release \`$tag\`.
 
           \`\`\`sh
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-hermes-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-openclaw-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-codex-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-opencode-marmot.sh | bash
-          curl -fsSL https://github.com/$GITHUB_REPOSITORY/releases/download/$tag/install-pi-marmot.sh | bash
+          set -eu
+          base_url="https://github.com/$GITHUB_REPOSITORY/releases/download/$latest_tag"
```
erskingardner, line 691: Doc accuracy, non-blocking: the prose above says "install WN Agent `$version` from immutable release `$tag`" but this code block downloads from the mutable `$latest_tag` alias. If the alias is repointed later, a copy-paste of this block installs a different version than the notes claim. Either point the code at `$tag`, or reword the sentence to say it installs whatever `wn-agent-latest` currently aliases.

erskingardner (dismissed):
Approve. The premise is sound and well-executed: replacing `curl | bash` with download → sibling `.sha256` verify → execute-only-local-copy is a real security hardening, the helpers fail closed (`set -eu`, explicit checksum-mismatch exits, strict 64-hex parsing in the installers), and the new CI audit job is properly pinned (SHA-pinned actions, exact `cargo-audit@0.22.1`, no error masking). The mutation-test gates are unusually thorough for doc policy.

Non-blocking notes:

1. **Policy flip on `wn-agent-latest`** (inline): this PR removes release.md's "not an authoritative update channel" sentence and starts promoting latest-alias installs in generated release notes. Probably intentional (the workflow does maintain that alias), but it deserves a line in the PR description.

2. **Prose/code mismatch in latest-notes heredoc** (inline): says it installs from immutable `` but downloads from mutable ``.

3. **Duplication**: the ~20-line `install_verified()` helper is copy-pasted ~10 times, three of them inside wn-agent-binaries.yml alone. Self-contained snippets are defensible for docs, but consider whether crate-level READMEs could link to the canonical integrations/README.md snippet instead of repeating the full helper — the gate already enforces its shape there.

4. **release.md version drift**: `check_agent_install_docs.sh` pins integrations/README.md to the workspace version, but nothing verifies release.md's hardcoded `wn-agent-v0.9.14` base_url stays current. Consider extending the gate so the pinned URL can't silently go stale.

Datawav (commented):
## Automated review

Outcome: 4 actionable findings remain at exact head `2e8f78fde4445e30c7e0a6e9ffbed70d8f6c5bd1`—3 major, 1 minor.

1. Major — Generated instructions misidentify a mutable install as immutable
- Evidence: `.github/workflows/wn-agent-binaries.yml:691` says the scripts install WN Agent ``$version`` from immutable release ``$tag``, while the immediately following changed block sets `base_url` to `.../releases/download/$latest_tag`; `$latest_tag` is the rolling `wn-agent-latest` alias.
- Trigger: A user copies those instructions after the alias has been retargeted.
- Consequence: The installed cohort can differ from the version and immutable tag asserted by the release notes, defeating artifact-identity expectations.
- Confidence: High.
- Smallest sound correction: Point this block at `$tag`; alternatively, state explicitly that it installs the release currently referenced by `wn-agent-latest` and does not pin `$tag` or `$version`.

2. Major — The Rust audit policy gate permits conditional skipping
- Evidence: Added `scripts/check_cargo_audit_ci.py:40-59` validates checkout, Rust installation, `cargo-audit@0.22.1`, and exact `run: cargo --locked audit`, but does not reject an `if:` condition on the `rust-audit` job or audit step. Added `scripts/test_check_cargo_audit_ci.py` has mutations for required text and error-masking controls but no conditional-skip mutation.
- Trigger: A later workflow edit adds `if: false` or an equivalent false expression while retaining every validated command.
- Consequence: GitHub skips the security audit while the policy checker still passes; a skipped required check may not fail the protection gate.
- Confidence: High.
- Smallest sound correction: Reject job-level and audit-step-level `if:` keys, and add mutation tests proving both forms raise `PolicyError`.

3. Major — The new Codex install surface is outside the SHA-256 semantic gate
- Evidence: Changed `integrations/codex/marmot/README.md:31-58` adds an `install_verified` example for `install-codex-marmot.sh`, but added `scripts/check_install_example_sha256.py:37` omits that README from `DOCUMENTED_INSTALL_CALLS`. Consequently, `repository_errors` does not apply `documented_surface_errors` to this changed surface.
- Trigger: A future edit removes `set -eu`, the companion checksum download, verification, or verified-local execution from the Codex example without introducing a `curl | bash` pipeline.
- Consequence: An active installation path can regress to incomplete or fail-open verification while the new CI gate remains green.
- Confidence: High.
- Smallest sound correction: Add `integrations/codex/marmot/README.md` with its expected `install-codex-marmot.sh` call count and a regression test that mutates this exact surface.

4. Minor — Later installer examples are not independently copy-pasteable
- Evidence: Changed connector documentation defines `install_verified` in an earlier fenced block, but later blocks call it directly—for example `integrations/hermes/marmot/README.md:88-99`, with equivalent changed calls in the OpenClaw, OpenCode, and Pi guides.
- Trigger: A reader copies a later invocation into a fresh shell instead of first executing the helper-definition block.
- Consequence: Installation stops with `install_verified: command not found`; this is a usability failure, not a checksum bypass.
- Confidence: High.
- Smallest sound correction: State immediately before each dependent block that the helper must first be defined in the same shell, or make each block self-contained.

.github/workflows/wn-agent-binaries.yml
```diff
@@ -478,14 +478,82 @@ jobs:
           cat > "$notes" <<EOF
           WN Agent \`$version\` from MDK commit \`$GITHUB_SHA\`.
 
+          Install whichever version the mutable \`wn-agent-latest\` alias currently references with:
+
+          \`\`\`sh
+          set -eu
+          base_url="https://github.com/$GITHUB_REPOSITORY/releases/download/wn-agent-latest"
+
+          install_verified() (
```
erskingardner, line 487: This block is duplicated verbatim three times in this one file (lines 487, 526, 693) plus six more copies across READMEs and release.md. Since this file already generates content programmatically, consider emitting all three note blocks from a single template/here-doc so a helper fix does not require nine synchronized edits.

scripts/check_install_example_sha256.py
```diff
@@ -0,0 +1,264 @@
...
+}
+SAME_SHELL_NOTICE_COUNTS = {
+    "integrations/README.md": 7,
+    "integrations/hermes/marmot/README.md": 3,
+    "integrations/openclaw/marmot/README.md": 2,
+    "integrations/opencode/marmot/README.md": 1,
+    "integrations/pi/marmot/README.md": 1,
+}
+SAME_SHELL_NOTICE = "Run this example in the same shell where `install_verified` above was defined."
+LATEST_ALIAS_CLAIMS = (
```
erskingardner: Matching literal escaped-backtick heredoc prose ties this gate to the exact wording of generated release notes; any rewording breaks CI even if verification semantics are intact. Prefer asserting on stable markers (e.g., presence of the alias URL) rather than full sentences.

scripts/check_install_example_sha256.py
```diff
@@ -0,0 +1,264 @@
...
+    },
+    # The repository-level quickstart owns release commands. The crate-level
+    # connector guide intentionally links there rather than duplicating them.
+    "integrations/hermes/marmot/README.md": {"install-hermes-marmot.sh": 4},
+    "integrations/openclaw/marmot/README.md": {"install-openclaw-marmot.sh": 3},
+    "integrations/codex/marmot/README.md": {"install-codex-marmot.sh": 1},
+    "integrations/opencode/marmot/README.md": {"install-opencode-marmot.sh": 2},
+    "integrations/pi/marmot/README.md": {"install-pi-marmot.sh": 2},
+}
+SAME_SHELL_NOTICE_COUNTS = {
```
erskingardner: Exact-count assertions on boilerplate sentences ("exactly 7 same-shell notices") will make routine doc edits fail CI until these numbers are manually retuned. A >=1 presence check catches removal regressions without the churn.

erskingardner (commented):
## Review summary

The security intent is good and the verify-then-execute guidance is a real improvement over `curl | bash`. Two concerns keep this from a clean approve:

1. **The `install_verified()` helper is copy-pasted ~9 times** (3x in `wn-agent-binaries.yml`, once each in `integrations/README.md`, four connector READMEs, and `release.md`). Any future bug in the helper (e.g., the checksum parsing, the trap, the tool fallback) must be fixed in 9 places, and the gate only checks that each copy *exists*, not that they're identical. Consider generating these surfaces (or at least the workflow-generated release notes) from a single template/source, or shipping a tiny verified-bootstrap script as its own release asset so docs reduce to "download bootstrap + checksum, run it."

2. **The doc gates are over-fit to exact prose.** Exact-count assertions on boilerplate sentences ("exactly 7 same-shell notices"), exact per-file installer-call counts, and literal escaped-backtick strings matched against workflow heredocs mean nearly every documentation edit will fail CI until counts are hand-updated. That's high ongoing friction for limited protection; presence/minimum checks would catch the actual regression class (removing verification) without freezing the prose.

Smaller notes:
- The sibling `.sha256` is fetched from the same origin over TLS, so it adds little against a channel attacker; it mainly guards against truncated/tampered release assets. Fine as defense-in-depth, but worth not overselling in docs.
- `check_cargo_audit_ci.py` hand-parses YAML with regexes; acceptable given stdlib-only constraints, but brittle if job layout shifts.
- No workspace/version-bump invariant violations; fixture test updates are consistent.

Datawav:
Addressed all current review feedback on exact head `f348c0eacebfe900c42d266f56180ca56c144bfb`:

- made every dependent installer example explicit about its same-shell prerequisite, while keeping the canonical quickstart's intentionally repeated connector commands protected by mutation tests
- enforced the complete approved `audit.toml` structure, rejected job- and step-level conditional skips including quoted keys, and hardened job-boundary parsing against comments
- added the Codex and canonical quickstart surfaces to the verified-installer gate, pinned release-guide calls to the workspace version, and added release/quickstart mutation coverage
- documented `wn-agent-latest` as a mutable convenience alias and immutable `wn-agent-v<version>` tags as the repeatable install path in both generated note paths and release guidance

Verification: `just fast-ci`, 11 installer-policy tests, 13 cargo-audit policy tests, exact-head independent review with zero findings, and all GitHub checks passed. All handled review threads are resolved; this PR was not merged or queued for merge.

.github/workflows/wn-agent-binaries.yml
Datawav: Kept the installer snippets independently copyable in this focused security PR rather than introducing generated documentation or a new bootstrap asset and trust path. The follow-up instead reduces synchronization risk with semantic gates across every documented surface, while avoiding exact prose/count coupling. A shared template remains a reasonable separate refactor, but it is not required to make these verified install paths correct or regression-safe.

scripts/check_install_example_sha256.py
Datawav: Replaced the literal full-sentence checks with structural heredoc checks for the mutable alias URL/marker and the exact or immutable release-tag path. Wording can now change without weakening the semantic invariant, and mutation coverage exercises those stable markers.

scripts/check_install_example_sha256.py
Datawav: Removed the exact notice counts. The gate now inspects each shell fence that calls install_verified without defining it and requires an adjacent same-shell prerequisite, so new or removed examples are handled structurally without retuning counts. Connector and canonical-quickstart mutations cover the regression.

erskingardner (approved):
## Re-review of head `0e8dd01`

Thanks — this iteration resolves my main concerns from the earlier review:

- The exact-count prose gates (`SAME_SHELL_NOTICE_COUNTS`, literal `LATEST_ALIAS_CLAIMS` sentences) are gone. The replacement checks are semantic: any `sh` fence that *invokes* `install_verified` without defining it must state its same-shell prerequisite nearby, and the generated release-note heredocs must mention the mutable alias vs immutable tag with the right URLs. These catch the actual regression class without freezing documentation wording.
- Removing the stale `$latest_tag` check from `check_agent_install_docs.sh` is consistent with that.
- Mutation tests were updated to target semantics rather than exact sentences, which keeps them meaningful under rewording.

Remaining non-blocking notes (fine to defer to follow-ups):

1. **`install_verified()` is still duplicated ~9 times** across `wn-agent-binaries.yml`, the connector READMEs, and `release.md`. The gate now verifies each surface's structure independently, which mitigates drift risk, but consider templating these surfaces or shipping a verified-bootstrap script asset in a future PR so fixes propagate from one source.
2. The same-shell heuristic (`same shell` + `install_verified` within 240 chars before a dependent fence) is still mildly wording-coupled — acceptable, just be aware a rephrase like "defined above in this session" will trip it.
3. Same-origin `.sha256` remains defense-in-depth rather than channel-attacker protection; docs already frame it reasonably.

No version-bump or repo-invariant violations. CI is green on this head apart from the still-running "Simulator and vectors" job. Approving.

Datawav:
Final follow-up is on exact head `0e8dd011e6ebb5042e45e0bee79d0eef5127fc7e`. Semantic gates now validate shell-fence and release-note behavior without freezing prose. All CI checks pass and every current thread is resolved.

PR #1537, opened 2026-08-23
Separate directory discovery from operational relays

erskingardner:
## Summary

- add host-configured directory relay URLs to `MarmotAppConfig`
- preserve explicit per-operation relay precedence, then use configured directory relays, then fall back to operational relays
- pass daemon discovery relays through every CLI-created `MarmotApp`
- cover the separation with app-level and two-relay daemon KeyPackage regressions

## Why

Directory and KeyPackage discovery currently fall back to the operational message-delivery relay set. Hosts that deliberately separate those planes can therefore create a daemon runtime with the right delivery relays while remote-member KeyPackage lookup uses the wrong relay. This makes the two relay roles independently configurable without changing the historical empty-config fallback.

## Verification

- `just fast-ci`
- `cargo test -p marmot-app configured_directory_relays_stay_separate_from_operational_relays`
- `cargo test -p wn-cli daemon_uses_discovery_relays_independently_from_operational_relays`

I also ran the complete `marmot-app` and `wn-cli` suites. They exposed existing failures outside this change. Representative failures were rerun individually and reproduced on an untouched `origin/master` worktree:

- `marmot-app`: `encrypted_media_endpoint_updates_are_full_replacement_and_admin_only`
- `wn-cli`: `daemon_defaults_create_identities_and_block_loopback_auto_watch_without_manual_sync_or_relay_env`

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring dedicated discovery relays for directory and KeyPackage lookups.
  - Discovery relays can now be managed independently from operational message-delivery relays.
  - Explicitly provided discovery relays take precedence, with fallback to operational relays when none are configured.

- **Bug Fixes**
  - Daemon-hosted directory and KeyPackage reads now consistently use the configured discovery relays.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -493,6 +493,7 @@ async fn execute_inner(
             .clone()
             .or_else(|| cli.daemon_discovery_relays.first().cloned())
             .or_else(|| cli.daemon_default_account_relays.first().cloned()),
+        cli.daemon_discovery_relays.clone(),
         account_home.clone(),
     )?;
     match command {
@@ -1113,6 +1114,7 @@ pub(crate) fn relay_lists_json(status: AccountRelayListStatus) -> Value {
 fn app_for(
     home: PathBuf,
     relay: Option<String>,
+    directory_relays: Vec<String>,
     account_home: AccountHome,
 ) -> Result<MarmotApp, WnError> {
     // Loopback-HTTP blob endpoints are only acted on when explicitly enabled for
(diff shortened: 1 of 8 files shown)


crates/cli/src/lib.rs
```diff
@@ -493,6 +493,7 @@ async fn execute_inner(
             .clone()
             .or_else(|| cli.daemon_discovery_relays.first().cloned())
             .or_else(|| cli.daemon_default_account_relays.first().cloned()),
+        cli.daemon_discovery_relays.clone(),
```
erskingardner, line 496: Non-blocking: the daemon start path validates these via `normalize_relay_list` (crates/cli/src/daemon/lifecycle.rs:240) before they reach a MarmotApp, but this direct (non-daemon) call site forwards raw CLI strings straight into `with_directory_relay_urls`. An invalid relay URL here now surfaces as a connect-time failure during directory reads instead of an early, clear validation error. Consider normalizing (or at least URL-validating) before constructing the config.

erskingardner (commented):
Review summary: **recommend merge.**

Clean, surgical change. The new `directory_relay_urls` config is threaded through a single choke point (`directory_source_relays`, crates/marmot-app/src/directory/methods.rs:856), so every directory/KeyPackage read path — including the runtime workers — picks it up without per-call-site changes. Precedence (explicit per-operation relays → configured directory relays → operational fallback) preserves historical behavior when the config is empty, and publication fanout is verified untouched by the new test. Both new tests target exactly the separation this PR claims, and CI is green.

One non-blocking note inline (crates/cli/src/lib.rs:496): the direct (non-daemon) CLI path forwards raw `--discovery-relays` strings into the config without the `normalize_relay_list` validation the daemon start path applies, so an invalid URL surfaces as a connect-time failure during directory reads rather than an early validation error.

PR #1540, opened 2026-08-23
fix(openclaw): use host-authorized reader for outbound media

### LINKED ISSUE #1539: fix(openclaw): make message-tool outbound media work through the host-authorized reader
## Summary

Outbound Marmot media works through `wn-agent`'s `send_media`, but sending a valid generated image through OpenClaw's normal public message-tool path fails with `media_path_denied` before it reaches the durable media send.

The connector's direct control-client path succeeds when the same file is staged beneath the `wn-agent --media-allowed-root`, so the encrypted media/upload pipeline itself is healthy. The failing seam is:

```text
message(action=send, channel=marmot, media=<local path>)
  -> OpenClaw media authorization/materialization
  -> Marmot channel adapter source resolution
  -> MARMOT_OUTBOUND_MEDIA_DIR staging
  -> wn-agent send_media
```

## Observed environment

- Host: OpenClaw `2026.7.2-beta.7`
- Connector dependency pin: OpenClaw `2026.7.1-2`
- Connector: `integrations/openclaw/marmot`
- Source image tested from:
  - the active OpenClaw agent workspace
  - the workspace media directory
  - OpenClaw's state media directory
  - OpenClaw's preferred temp media directory

All normal message-tool local-path attempts were rejected. A direct `MarmotAgentControlClient.sendMedia(...)` call using a short-lived copy under the configured `wn-agent --media-allowed-root` succeeded and produced a durable kind-9 message with the expected image attachment.

## Expected behavior

An agent should be able to send an image through the standard OpenClaw interface:

```text
message(action="send", channel="marmot", target=<group-id>, media=<authorized-local-path>)
```

OpenClaw should authorize/read the source, the connector should stage a private copy under `MARMOT_OUTBOUND_MEDIA_DIR`, and `wn-agent` should encrypt, upload, and send it.

Agents should not need to invoke `MarmotAgentControlClient.sendMedia()` directly.

## Suspected seam

`resolveOutboundMediaUpload()` currently treats local paths differently from remote media:

- local paths are re-read through the connector's pinned-SDK `readLocalFileFromRoots(...)` helper;
- remote paths use the host-provided `ctx.mediaReadFile` accessor when present.

OpenClaw already supplies `ctx.mediaReadFile` / `mediaAccess.readFile` as an authorized host accessor based on the active agent's current media policy. Ignoring that accessor for local paths creates a second, version-sensitive allowlist decision inside the connector.

The exact root cause should be confirmed with a host-path regression test before changing behavior, but the smallest likely fix is:

1. Prefer `ctx.mediaReadFile(mediaUrl)` for both local and remote sources when the host provides it.
2. Fall back to connector-side `readLocalFileFromRoots(...)` only when the host accessor is absent.
3. Continue staging bytes under `MARMOT_OUTBOUND_MEDIA_DIR`.
4. Keep `wn-agent`'s independent `--media-allowed-root` enforcement.

This preserves both security boundaries without making two OpenClaw SDK versions independently authorize the same source path.

## Missing regression coverage

Current unit tests prove `adapter.send.media()` works with hand-constructed contexts, including explicit `mediaLocalRoots`. They do not exercise the installed-host path from the real shared `message` action into the Marmot adapter.

Add host-contract/E2E coverage for:

- authorized local image beneath the active agent workspace succeeds via `message(action=send)`;
- `buffer + filename` succeeds through the same public interface;
- a path outside authorized roots is rejected before any connector read/send;
- the staged copy is removed after both success and failure;
- the durable Marmot message contains the expected attachment metadata.

## Agent/developer guidance

Add a short outbound-media contract to the connector documentation and `AGENTS.md`:

- runtime agents use OpenClaw's normal `message` tool with `media`/`attachments`;
- generated files should live under the active workspace or OpenClaw-managed media store;
- data URLs are not supported;
- direct `MarmotAgentControlClient.sendMedia()` calls are reserved for connector tests/smoketests;
- the host-provided media reader is the source authorization boundary, while `wn-agent --media-allowed-root` remains the staging authorization boundary.

Repository `AGENTS.md` guidance helps coding agents maintain the contract; any runtime-agent instruction should be added through connector setup/docs rather than relying on repository instructions alone.

## Acceptance criteria

- [ ] A real OpenClaw `message(action=send, channel=marmot, media=<workspace image>)` succeeds without direct connector calls.
- [ ] Existing outside-root exfiltration tests remain passing.
- [ ] `wn-agent --media-allowed-root` enforcement remains unchanged.
- [ ] Staged plaintext media is deleted after the send attempt.
- [ ] Host-contract coverage runs against supported latest and beta OpenClaw versions.
- [ ] Connector docs describe the supported agent-facing media-send path.

## Related observation (probably upstream)

After the local-path failure, a fallback attempt using a `data:image/jpeg;base64,...` URL was correctly rejected as unsupported, but OpenClaw mirrored the internal failure diagnostic into the Marmot conversation. That appears to be an OpenClaw message-tool/error-delivery issue rather than a Marmot connector issue and should be tracked upstream separately if reproducible.

erskingardner:
## Summary
- prefer the running OpenClaw host media reader for local and remote outbound sources, avoiding a second version-skewed allowlist decision
- retain connector-side local-root validation when no host reader exists, stage only authorized bytes, and document the public message-tool contract
- cover host-reader precedence, denied reads, attachment metadata, and staged-file cleanup after success and failure

## Test plan
- [x] New local-path host-reader test failed before the fix and passes after
- [x] `pnpm typecheck`
- [x] `pnpm test` (245 passed, 2 skipped)
- [x] `just openclaw-host-compat-test` against OpenClaw 2026.7.2-beta.7 (80 passed)
- [x] `just fast-ci`

Fixes #1539

## Summary by CodeRabbit

* **New Features**
  * Improved outbound media handling for local and remote files.
  * Added support for host-authorized media access and secure staging before sending.
  * Preserved original filenames for managed media attachments.
  * Added automatic cleanup of staged media after successful or failed sends.

* **Bug Fixes**
  * Prevented unauthorized media from being staged or sent.
  * Ensured staged files are removed when permission updates fail.

* **Documentation**
  * Updated runtime guidance for sending media and managing temporary files.

### DIFF
--- a/crates/marmot-app/tests/relay_runtime.rs
+++ b/crates/marmot-app/tests/relay_runtime.rs
@@ -5121,7 +5121,10 @@ async fn app_runtime_starts_directory_subscriptions_for_known_users() {
         loop {
             let health = runtime.shared_services().relay_plane().relay_health().await;
             if health.directory_active_subscriptions == 1
-                && health.directory_completed_subscription_syncs == 1
+                // A relay recovery can legitimately complete another rebuild
+                // before this poll observes the initial one. This counter is
+                // monotonic, so waiting for exact equality races that recovery.
+                && health.directory_completed_subscription_syncs >= 1
             {
                 break health;
             }
@@ -5131,7 +5134,7 @@ async fn app_runtime_starts_directory_subscriptions_for_known_users() {
     .await
     .expect("directory subscriptions should complete asynchronously");
     assert_eq!(health.directory_active_subscriptions, 1);
-    assert_eq!(health.directory_completed_subscription_syncs, 1);
+    assert!(health.directory_completed_subscription_syncs >= 1);
     runtime.shutdown().await;
 }
 
--- a/integrations/openclaw/marmot/AGENTS.md
+++ b/integrations/openclaw/marmot/AGENTS.md
@@ -13,6 +13,12 @@ The OpenClaw counterpart of `integrations/hermes/marmot`. Read `README.md` first
 - No QUIC, crypto, relay, or MLS logic here.
 - Privacy-safe logging only: no account ids, group ids, message ids, pubkeys,
   relay URLs, payloads, ciphertext, plaintext, or key material.
+- Runtime agents send media through OpenClaw's normal `message` tool with
+  `media`/`attachments`, using files beneath the active workspace or an
+  OpenClaw-managed media store; data URLs are unsupported. Treat the host's
+  media reader as source authorization and `wn-agent --media-allowed-root` as
+  independent staging authorization. Direct control-client `sendMedia()` calls
+  are for connector tests and smoketests only.
 
 ## Key files
 
--- a/integrations/openclaw/marmot/README.md
+++ b/integrations/openclaw/marmot/README.md
@@ -21,6 +21,17 @@ Repeating a removal is idempotent. Reaction content follows the
 agent-control v2 bound: non-blank, control-free, and at most 64 Unicode scalar
 values.
 
+Agents send outbound media through OpenClaw's normal
+`message(action="send", channel="marmot", media=..., attachments=...)` interface.
+Generated files must be placed under the active agent workspace or an
+OpenClaw-managed media store; data URLs are not supported. OpenClaw's
+host-provided media reader is the source authorization boundary. The plugin
+stages the authorized bytes as a private, short-lived copy under
+`MARMOT_OUTBOUND_MEDIA_DIR`, while `wn-agent --media-allowed-root` independently
+authorizes that staging path before encrypting and uploading it. Direct
+`MarmotAgentControlClient.sendMedia()` calls are reserved for connector tests
+and smoketests, not runtime agents.
+
 For each activated inbound turn, the plugin asks `wn-agent` for a bounded recent
 materialized chat window and supplies it to OpenClaw with durable message ids,
 senders, timestamps, reply links, current reaction summaries, and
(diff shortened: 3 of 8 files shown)


agent-r3n (commented):
This looks like the correct functional fix for #1539: the adapter now prefers OpenClaw's host-authorized reader for local media, retains the connector-side fallback allowlist, stages beneath `MARMOT_OUTBOUND_MEDIA_DIR`, and leaves `wn-agent --media-allowed-root` as an independent authorization boundary. I found no new source-authorization bypass, and the targeted outbound tests/typecheck pass locally.

I would add one regression test before calling #1539 fully solved:

**MEDIUM — the new test does not exercise the original failing public path**

`integrations/openclaw/marmot/test/outbound.test.ts:279` hand-constructs a `ChannelMessageSendMediaContext` and invokes `adapter.send.media()` directly. That proves reader precedence inside the adapter, but it does not cover:

```text
message(action="send")
  -> installed OpenClaw host media-access resolution
  -> durable outbound core
  -> Marmot adapter
```

Please add a host-contract/E2E test that invokes the real generic message action with an image beneath the active workspace and verifies the Marmot/`wn-agent` stub receives the expected attachment metadata. The `buffer + filename` public-interface variant requested in #1539 is also still missing. This is the exact seam that failed in production, so an adapter-only test could remain green if the host stops supplying or reshapes the reader.

**LOW — cleanup if `chmod` fails**

In `src/outbound.ts`, `writeFile(..., { mode: 0o640 })` is followed by a separate `chmod`. If that `chmod` throws, resolution exits before returning the cleanup callback and the staged plaintext file can remain. Since `writeFile` already specifies `0o640`, either remove the redundant `chmod` or delete the staged file before rethrowing.

Overall verdict: **likely fixes the observed behavior, but the real message-tool regression remains unproven.** The separate data-URL diagnostic leak is appropriately outside this PR's scope.

erskingardner:
Addressed both review points in 8a86388a:

- Added installed-host regressions that exercise the public generic `message(action="send")` path through OpenClaw media resolution, durable core dispatch, and the Marmot adapter for both a workspace file and `buffer + filename`. The compatibility fixture now includes the plugin manifest so the same tests run against the supported beta.
- Made staged-file permission failures remove the plaintext staging file before rethrowing, with a focused regression test.
- Preserved the original filename for OpenClaw-managed buffer media using the host media-runtime contract.

Validated with `pnpm typecheck`, `pnpm test` (249 passed, 2 skipped), `just openclaw-host-compat-test` (84 passed), and `just fast-ci`.

erskingardner:
Fixed the relay-runtime CI flake in a6ab1a49. The test polled a monotonic `directory_completed_subscription_syncs` counter for exact equality with `1`; a legitimate recovery rebuild could advance it to `2` between polls, causing the 20-second timeout. It now requires at least one completed sync while retaining the exact active-subscription assertion.

Reproduced locally before the fix (2 passes, then the same timeout), then validated 10 consecutive repetitions plus `just fast-ci`.

Datawav (approved):

PR #1541, opened 2026-08-23
Prioritize storage close during suspension shutdown

erskingardner:
## Summary

- close runtime admission, every SQLite connection, and the Marmot root lease before awaiting graceful shutdown work
- run terminal closure in a runtime-owned task so cancellation of the awaiting host/UniFFI future cannot strand storage open
- give directory, account, relay, audit, and account-open cleanup one bounded graceful budget after lock release
- publish `storage_is_closed` only after database closure and root-lease release actually complete
- add subsystem-stall, retained-handle, WAL/SHM, root-lease, same-root reopen, cross-process reopen, concurrency, idempotency, late-work, and caller-cancellation regressions

## Why

MarmotKit 0.9.14's `shutdown_and_close` awaited the full graceful `shutdown` sequence before invoking `close_storage`. Several waits ahead of closure can be unbounded or materially exceed iOS background execution time: directory command delivery and initial-sync join, invite catch-up and account worker draining, async mutex acquisition, and an active audit upload.

If iOS suspended the process during one of those waits, idle WAL connections and the root runtime lease were still live. A WAL connection holds a persistent lock on its `-shm` sidecar even when no SQLite statement is active, which can produce a RunningBoard `0xdead10cc` kill with idle/parked stacks.

The terminal operation now prioritizes the fact iOS requires: release suspension-sensitive locks first. Graceful cleanup remains best-effort and bounded afterward.

## Semantics and integrity

Closing remains terminal. An already-admitted database open or SQLite statement is allowed to reach its safe close boundary; an uncommitted SQLite transaction rolls back, and later work receives `StorageError::Closed` instead of reopening the root.

This is transaction-safe but can cut a higher-level operation between transactions or after an external side effect. Protocol-critical operations retain their existing durable intent, pending-state, exact-event retry, and reopen-reconciliation mechanisms. The architecture documentation records that tradeoff explicitly. This PR does not introduce a new public API, regenerate bindings, or bump/release any version.

## Validation

Passed:

- `just fast-ci`
- `cargo test -p storage-sqlite` — 402 passed, 4 ignored
- `cargo test -p marmot-uniffi` — 157 passed across unit, fixture, and smoke tests
- `cargo test -p marmot-app shutdown_and_close --lib -- --nocapture` — all 6 terminal-shutdown regressions passed
- complete `marmot-app` library tests — 758 passed, 1 ignored
- `git diff --check`

The full `cargo test -p marmot-app` run reached the unrelated `relay_runtime` integration binary, where 5 existing encrypted-media/retention/removal tests failed. A representative failure (`encrypted_media_endpoint_updates_are_full_replacement_and_admin_only`) reproduced in an untouched worktree at current `master` (`845ebae9`), with the same `left: 3, right: 1` assertion.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved shutdown reliability by closing databases and releasing runtime resources before cleanup.
  - Shutdown now continues safely if the initiating call is cancelled.
  - Repeated and concurrent shutdown requests are handled safely.
  - Storage can be reopened reliably after shutdown, including from another process.

- **Documentation**
  - Clarified shutdown behavior, bounded cleanup timing, and completion requirements.

### DIFF
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -510,10 +510,14 @@ pub struct MarmotApp {
     /// it counts against the same App Group suspension rule as the databases
     /// (see [`Self::close_storage`]).
     root_runtime_lease: Arc<Mutex<Option<MarmotRootRuntimeLease>>>,
-    /// Latched by [`Self::close_storage`]. Every database accessor checks it so
-    /// a late call cannot silently reopen a database — and re-lock a container
-    /// the host has just been told is lock-free.
+    /// Latched as soon as [`Self::close_storage`] starts. Every database
+    /// accessor checks it so a late call cannot silently reopen a database
+    /// while terminal close is in progress or after it completes.
     storage_closed: Arc<AtomicBool>,
+    /// Published only after every database close has been attempted and the
+    /// root lease has been released. This is the host-facing completion fact;
+    /// `storage_closed` above is the earlier admission latch.
+    storage_close_completed: Arc<AtomicBool>,
     /// Admission control that makes the close *atomic* rather than merely
     /// latched: database opens hold the read side across create-and-publish,
     /// [`Self::close_storage`] holds the write side across its whole teardown.
@@ -1349,6 +1353,7 @@ impl MarmotApp {
             root,
             root_runtime_lease: Arc::new(Mutex::new(None)),
             storage_closed: Arc::new(AtomicBool::new(false)),
+            storage_close_completed: Arc::new(AtomicBool::new(false)),
             storage_lifecycle: Arc::new(RwLock::new(())),
             relay_urls,
             relay_plane,
@@ -1421,6 +1426,7 @@ impl MarmotApp {
             root: root.as_ref().to_path_buf(),
             root_runtime_lease: Arc::new(Mutex::new(None)),
             storage_closed: Arc::new(AtomicBool::new(false)),
+            storage_close_completed: Arc::new(AtomicBool::new(false)),
             storage_lifecycle: Arc::new(RwLock::new(())),
             relay_urls,
             account_home,
@@ -4896,6 +4902,7 @@ impl MarmotApp {
                 .unwrap_or_else(|poisoned| poisoned.into_inner())
                 .take(),
         );
+        self.storage_close_completed.store(true, Ordering::Release);
 
         tracing::debug!(
             target: "marmot_app::storage",
@@ -4911,16 +4918,17 @@ impl MarmotApp {
         }
     }
 
-    /// Whether [`Self::close_storage`] has run. Databases are unreachable from
-    /// this handle once it returns true.
+    /// Whether [`Self::close_storage`] has finished closing every database and
+    /// releasing the root lease. Databases are unreachable from this handle
+    /// once it returns true.
     #[must_use]
     pub fn storage_is_closed(&self) -> bool {
-        self.storage_closed.load(Ordering::Acquire)
+        self.storage_close_completed.load(Ordering::Acquire)
     }
 
     /// Fail rather than reopen a database after [`Self::close_storage`].
     fn ensure_storage_open(&self, database: &'static str) -> Result<(), AppError> {
-        if self.storage_is_closed() {
+        if self.storage_closed.load(Ordering::Acquire) {
             return Err(AppError::from(cgka_traits::StorageError::Closed(format!(
                 "{database} unavailable: app storage is closed"
             ))));
--- a/crates/marmot-app/src/runtime/mod.rs
+++ b/crates/marmot-app/src/runtime/mod.rs
@@ -18,6 +18,8 @@ use marmot_account::{
     NostrAccountImport,
 };
 use serde::{Deserialize, Serialize};
+#[cfg(test)]
+use tokio::sync::Semaphore;
 use tokio::sync::{Mutex, Notify, broadcast, mpsc, oneshot, watch};
 use tokio::task::{JoinHandle, JoinSet};
 use tokio::time::timeout;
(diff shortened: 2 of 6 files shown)


crates/marmot-app/src/lib.rs
```diff
@@ -510,10 +510,14 @@ pub struct MarmotApp {
...
     /// (see [`Self::close_storage`]).
     root_runtime_lease: Arc<Mutex<Option<MarmotRootRuntimeLease>>>,
-    /// Latched by [`Self::close_storage`]. Every database accessor checks it so
-    /// a late call cannot silently reopen a database — and re-lock a container
-    /// the host has just been told is lock-free.
+    /// Latched as soon as [`Self::close_storage`] starts. Every database
+    /// accessor checks it so a late call cannot silently reopen a database
+    /// while terminal close is in progress or after it completes.
     storage_closed: Arc<AtomicBool>,
+    /// Published only after every database close has been attempted and the
```
erskingardner, line 517: Non-blocking: while you're splitting the admission latch from the completion fact here, the doc comment on `close_storage` itself (~line 4814) is now stale — it still tells callers they "must quiesce first" and claims `shutdown_and_close` "drains the account workers before calling this", which is the ordering this PR deliberately inverts. Worth updating so the terminal entry point and the primitive it calls don't contradict each other.

erskingardner (commented):
**Review verdict: recommend merge** (posted as a comment because GitHub won't let the PR author's account approve its own PR).

The reordering is the right call for the `0xdead10cc` failure mode: the lifecycle stop latch and storage close now sit ahead of every wait that can outlive an iOS background assertion, and the admission-latch vs completion-fact split (`storage_closed` / `storage_close_completed`) makes `storage_is_closed` honest — the host is only told the container is lock-free once the lease is actually released. Running the terminal sequence in a runtime-owned task so a cancelled host/UniFFI future can't strand storage open is correct, and the regression coverage is genuinely strong: per-phase stalls, caller cancellation, concurrent + repeated closes, WAL/SHM removal, same-root and cross-process reopen.

Checked the surrounding paths at the head SHA: `close_storage` has no early-return path that could leave `storage_close_completed` unset (it is stored before the `first_error` return), `begin_shutdown` is idempotent so the double invocation from `run_terminal_shutdown_and_close` → `shutdown` is harmless, and the dropped graceful `shutdown()` future on budget expiry releases its mutex guards on drop.

Two non-blocking notes:

- The doc comment on `close_storage` (lib.rs ~4814) still says "Callers must quiesce first" and that `shutdown_and_close` "drains the account workers before calling this" — that ordering is exactly what this PR inverts. AGENTS.md and local-artifact-safety.md were updated; this one doc was missed.
- In `runtime_shutdown_and_close_bounds_every_graceful_shutdown_phase` the stall is never released. That's fine (the timeout drops the stalled future), but a one-line comment saying so would save the next reader a minute.

CI is green across the matrix at the reviewed head.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4356,14 +4496,43 @@ impl MarmotAppRuntime {
...
+        tokio::spawn(async move { runtime.run_terminal_shutdown_and_close().await })
+            .await
+            .map_err(|err| {
+                AppError::BlockingTask(format!("terminal shutdown task failed: {err}"))
+            })?
+    }
+
+    async fn run_terminal_shutdown_and_close(self) -> Result<(), AppError> {
+        self.shared.lifecycle().begin_shutdown();
+        #[cfg(test)]
```
erskingardner, line 4516: Non-blocking: `stop_relay_telemetry_exporter` is a synchronous task abort, not part of the unbounded graceful work — consider calling it here right after `begin_shutdown()` (as `shutdown` does) so the exporter cannot race database closure and observe `Closed` during the close window. Fine either way if that noise is expected.

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -4302,21 +4359,98 @@ impl MarmotAppRuntime {
...
+
+    #[cfg(test)]
+    pub(crate) fn set_shutdown_grace_wait_for_test(&self, wait: Duration) {
+        *self
+            .shutdown_grace_wait_for_test
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner()) = wait;
+    }
+
+    fn shutdown_grace_wait(&self) -> Duration {
```
erskingardner, line 4385: Nit: the `#[cfg(test)] return ... / #[cfg(not(test))] { ... }` split inside one fn is a bit awkward; two small fns (one per cfg) or a `const` + test-only override setter would read cleaner. Not blocking.

erskingardner (commented):
**Review: recommend merge** (posted as comment since this is the author's own PR)

The reordering is the right fix for the suspension deadline problem: admission latch → storage close → bounded graceful drain, with the terminal operation running in a runtime-owned task so caller cancellation cannot strand WAL/SHM locks or the root lease.

Verified:
- The close-first ordering is sound. `close_storage` holds `storage_lifecycle` write across teardown, latches `storage_closed` before draining, and `storage_close_completed` is published after lease release on both Ok and Err paths, so `storage_is_closed` remains an honest completion fact while `ensure_storage_open` correctly checks the earlier latch directly — the semantic split is handled consistently everywhere.
- Cancellation/concurrency claims hold up: spawning + awaiting the JoinHandle detaches the terminal task if the host future is dropped, concurrent closes serialize on the lifecycle write lock, idempotency is preserved.
- Regression coverage is strong — per-phase stall bounds, retained-handle/WAL-sidecar assertions, root-lease reacquire, same-root reopen, real cross-process reopen child test, caller-cancellation survival.
- Docs (`local-artifact-safety.md`, AGENTS.md, UniFFI docs) accurately record the new ordering and the transaction-safe-but-operation-transparent tradeoff; no version bumps.

Two non-blocking inline notes: consider aborting the telemetry exporter before the blocking close (it's a free synchronous stop and avoids `Closed` noise during the close window), and the `shutdown_grace_wait` cfg split reads a bit awkwardly.

Datawav (commented):
## Automated review

No actionable findings.

- Diff coverage is complete: all 6 changed files include patches; no files are missing.
- Lifecycle/restart: close admission and storage completion are distinct; terminal closure survives caller cancellation, remains idempotent under concurrent calls, and supports same-root reopening through a new runtime.
- Negative/boundary behavior: the production path is exercised for stalled shutdown phases, retained handles, late work, repeated/concurrent close, WAL/SHM cleanup, root-lease release, and caller cancellation.
- Compatibility/accessibility: the UniFFI API shape is unchanged; no user-interface, localization, large-text, accessibility, or authorization behavior is modified.
- Discussion leads: the stale `close_storage` documentation is non-runtime documentation drift; the telemetry-exporter ordering suggestion does not establish a material consequence; the `cfg` observation is stylistic.
- Check evidence is available, complete, and reports all relevant checks successful.

PR #1542, opened 2026-08-23
Add account-scoped invite policies for direct bots

erskingardner:
## Summary

- add account-scoped `deny`, `allowlist`, `any_authenticated_direct`, and `any_authenticated` invite policies
- preserve `allowlist` as the default for existing records and keep authenticated-welcomer fail-closed behavior
- require exactly two MLS members for `any_authenticated_direct`, declining multiparty invites
- expose policy get/set through agent control and `wn-agent bootstrap --invite-policy`
- document that invite admission remains separate from sender and activation policy

## Testing

- `just fast-ci`

The connector tests include real local-relay coverage proving that an unlisted authenticated direct invite is accepted while a three-member invite is declined and archived.

## Summary by CodeRabbit

* **New Features**
  * Added account-scoped invite policies: deny, allowlist, authenticated direct, and any authenticated.
  * Added controls to view and update an account’s invite policy.
  * Bootstrap configuration now supports setting and reporting invite policies.
  * Policy settings persist across restarts, with legacy configurations defaulting to allowlist behavior.

* **Bug Fixes**
  * Clarified enforcement for direct and multi-party invites, including fail-closed handling for unauthenticated welcomes.

* **Documentation**
  * Documented policy behavior, defaults, group restrictions, bootstrap configuration, and separation from sender activation settings.

### DIFF
--- a/crates/agent-connector/README.md
+++ b/crates/agent-connector/README.md
@@ -81,8 +81,29 @@ wn-agent --version
 
 ## Invite Policy
 
-Production connectors accept pending group invites only when the MLS-authenticated welcomer is on the configured
-`--allow-welcomer` list. An empty allowlist rejects every invite.
+Invite policy is account-scoped and defaults to `allowlist`, preserving the existing behavior: a pending group invite
+is accepted only when its authenticated welcomer is on the account's configured `--allow-welcomer` list. An empty
+allowlist rejects every invite.
+
+Set a production invite policy while bootstrapping or reusing an account:
+
+```sh
+wn-agent bootstrap \
+  --home ~/.marmot-agent \
+  --invite-policy any-authenticated-direct
+```
+
+The available policies are:
+
+- `deny` — decline every invite while retaining any stored allowlist entries;
+- `allowlist` — accept only authenticated welcomers on the account allowlist;
+- `any-authenticated-direct` — accept any authenticated welcomer only when the resulting MLS group has exactly two
+  members (the agent and one peer);
+- `any-authenticated` — accept any authenticated welcomer, including invites to multi-party groups.
+
+Every policy fails closed when the Welcome has no authenticated welcomer. Invite policy controls group admission only;
+the Hermes, OpenClaw, or terminal-harness integration independently controls which admitted senders may invoke the
+agent and whether multi-party messages require activation.
 
 Local development may use `--dev-allow-any-invites` together with `--debug-controls`. The connector warns when this
 mode is active and accepts any authenticated welcomer, but it still rejects a welcome whose authenticated author is
--- a/crates/agent-connector/README.md
+++ b/crates/agent-connector/README.md
@@ -98,7 +98,8 @@ The available policies are:
 - `deny` — decline every invite while retaining any stored allowlist entries;
 - `allowlist` — accept only authenticated welcomers on the account allowlist;
 - `any-authenticated-direct` — accept any authenticated welcomer only when the resulting MLS group has exactly two
-  members (the agent and one peer);
+  members (the agent and one peer); the member count is checked once, when the invite is confirmed, and does not
+  prevent later membership changes;
 - `any-authenticated` — accept any authenticated welcomer, including invites to multi-party groups.
 
 Every policy fails closed when the Welcome has no authenticated welcomer. Invite policy controls group admission only;
--- a/crates/agent-connector/src/invite_policy.rs
+++ b/crates/agent-connector/src/invite_policy.rs
@@ -377,6 +377,9 @@ impl AgentConnector {
         };
         let is_direct =
             if policy == AgentControlInvitePolicy::AnyAuthenticatedDirect && welcomer.is_some() {
+                // This worker evaluates invites serially. Directness is an
+                // admission-time snapshot; later membership changes remain
+                // possible after the invite is accepted.
                 self.runtime
                     .group_mls_state(&account.label, group_id)
                     .await?
--- a/crates/agent-connector/src/invite_policy.rs
+++ b/crates/agent-connector/src/invite_policy.rs
@@ -11,6 +11,7 @@
 
 use std::collections::HashSet;
 
+use agent_control::AgentControlInvitePolicy;
 use cgka_traits::{GroupId, MemberId, engine::GroupEvent};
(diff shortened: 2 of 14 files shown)


crates/agent-connector/src/invite_policy.rs
```diff
@@ -361,24 +362,31 @@ impl AgentConnector {
...
-            welcomer_present,
-            welcomer_allowlisted,
-        );
+        let is_direct =
+            if policy == AgentControlInvitePolicy::AnyAuthenticatedDirect && welcomer.is_some() {
+                self.runtime
+                    .group_mls_state(&account.label, group_id)
+                    .await?
+                    .member_count
+                    == 2
```
erskingardner, line 387: Non-blocking: this reads `member_count` before `accept_group_invite`, so a concurrent welcome landing between the check and the join could slip past `any_authenticated_direct`. In practice invites are applied serially per account by the reconcile worker, so the window looks unreachable today — worth a one-line comment here if that serialization is the invariant this relies on.

crates/agent-connector/src/tests.rs
```diff
@@ -2482,7 +2519,7 @@ async fn connector_policy_declines_unlisted_welcomer() {
 }
 
 #[tokio::test]
-async fn connector_policy_dev_allow_any_accepts_unlisted_authenticated_welcomer() {
+async fn connector_policy_any_authenticated_direct_accepts_unlisted_direct_welcomer() {
```
erskingardner, line 2522: Non-blocking: converting the old `dev_allow_any_invites` e2e test into an explicit-policy test removed the only end-to-end coverage of both the dev-flag path and the `any_authenticated` multiparty-accept path. The pure-function tests cover the truth table and the flag-to-policy mapping is trivial, so fine as-is — but one variant accepting a 3-member group under `any_authenticated` would close the gap if you want it.

erskingardner (commented):
Nice, surgical change. The policy model is clean: fail-closed on unauthenticated welcomes is preserved everywhere, `deny` retains stored entries, corrupt/legacy records default back to `allowlist` with an empty list (still fail-closed), and the dev flag now just maps onto `AnyAuthenticated` instead of being a parallel code path. Docs updated in all three places (connector README, integrations README, AGENTS.md), and there's real local-relay coverage for accept/decline.

Non-blocking notes inline; nothing blocking merge.

crates/agent-control/src/lib.rs
```diff
@@ -15,6 +15,54 @@ pub const AGENT_CONTROL_TIMELINE_MAX_LIMIT: u32 = 50;
...
 pub const AGENT_CONTROL_TIMELINE_REACTIONS_MAX: usize = 64;
 
+/// Account-scoped policy for confirming inbound Marmot group invites.
+///
+/// A policy never makes an unauthenticated Welcome acceptable. The
+/// `any_authenticated_direct` variant additionally requires the joined MLS
+/// group to contain exactly the local agent and one peer.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum AgentControlInvitePolicy {
```
erskingardner, line 25: Minor (non-blocking): the wire format here is snake_case (`any_authenticated_direct`) while the CLI/`FromStr`/`Display` spelling is kebab-case (`any-authenticated-direct`). An integration client driving the control socket directly has to discover that the JSON token differs from the bootstrap flag value. Consider adding per-variant `#[serde(alias = "any-authenticated-direct")]` (etc.) so both spellings deserialize, or at least documenting both spellings in this crate's README.

crates/agent-connector/README.md
```diff
@@ -81,8 +81,29 @@ wn-agent --version
...
+  --invite-policy any-authenticated-direct
+```
+
+The available policies are:
+
+- `deny` — decline every invite while retaining any stored allowlist entries;
+- `allowlist` — accept only authenticated welcomers on the account allowlist;
+- `any-authenticated-direct` — accept any authenticated welcomer only when the resulting MLS group has exactly two
+  members (the agent and one peer);
+- `any-authenticated` — accept any authenticated welcomer, including invites to multi-party groups.
```
erskingardner, line 103: Minor (non-blocking): worth stating explicitly that the two-member check is evaluated only at join time. Nothing prevents the welcomer from adding members to the group immediately after the agent joins, so `any-authenticated-direct` is an admission-time guarantee, not an enduring DM guarantee. The sender/activation-policy separation noted below mitigates this, but one sentence here ("the member count is checked once, when the invite is confirmed") would make the contract precise for operators.

erskingardner (commented):
Clean, well-scoped change. The policy resolution in `apply_invite_policy` stays fail-closed for unauthenticated Welcomes under every policy, `dev_allow_any_invites` maps exactly onto its previous semantics (`AnyAuthenticated`), legacy allowlist records default to `allowlist` via `#[serde(default)]` (with a regression test proving entries are preserved), and the real-relay connector tests cover both the direct-accept and multiparty-decline paths. Docs and AGENTS.md updates follow repo invariants. Two minor, non-blocking notes inline. Recommend merge (posted as a comment review since GitHub does not allow approving your own PR).

crates/agent-connector/src/invite_policy.rs
erskingardner: Added a code comment documenting that invite candidates are evaluated serially and that the two-member result is an admission-time snapshot; later membership changes remain possible.

crates/agent-connector/src/tests.rs
erskingardner: Added a real-relay `any_authenticated` variant that accepts a three-member group, alongside the existing `any_authenticated_direct` rejection case.

crates/agent-control/src/lib.rs
erskingardner: Added serde aliases for the kebab-case CLI spellings, focused deserialization/canonical-serialization coverage, and README documentation of both wire forms.

crates/agent-connector/README.md
erskingardner: Updated the connector README to state explicitly that member count is checked once when the invite is confirmed and does not prevent later membership changes.

Datawav (approved):

PR #1543, opened 2026-08-24
fix(marmot-app): give the drained seam the projection effects it skipped

jgmontoya:
## Problem

Two seams consume the same engine events: `observe_account_device_effects` (live delivery) and `observe_drained_session_events` (session-history replay after a crash or restart). They drifted. On the same events the drained seam skipped member-departure push-token cleanup, the `set_group_self_membership` writes on self-departure and re-join (leaving the account unread aggregate stale), the entire `GroupDisbanded` terminal sweep — which its own doc comment promised — and the invalidation-event timeline withdrawal. A device that crashed and replayed its history healed its groups but kept stale push tokens, a stale unread count, and timeline records that should have been withdrawn.

## Fix

The shared per-event handling moves into one helper, `observe_event_projection_effects`, called by both seams at the same loop position — no mode flags, since every callee takes `&self`; the deliberate seam differences (quarantine-tolerant projection, recovery-evidence arming, system-row synthesis) stay where they were. This makes the drift that caused the defect structurally impossible for these writes. Every added operation is idempotent under replay — absolute UPDATEs, conflict-converging queues, filtered or no-change invalidation — so a batch the inbound seam already applied before the crash re-applies as a no-op, which a test pins. One known residue: the drained seam still skips `project_group_system_rows` (kind-1210 system rows), tracked separately because it changes the seam's `projection_updates` contract.

## Tests

Five, each red before the fix: drained departure removes the member's push tokens, drained self-departure and re-join move stored self-membership, drained disband performs the terminal push sweep (the first test to exercise that arm on either seam), drained invalidation withdraws the timeline record, and a replayed already-applied batch is a no-op.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved synchronization of session events and app state.
  - Push-token cleanup now works correctly when members leave or sessions are disbanded.
  - Membership changes, including leaving and rejoining, are reflected reliably.
  - Disbanded sessions now remove stale registrations and peer information.
  - Invalidated messages are consistently marked as unavailable.
  - Replayed events no longer create duplicate removals or inconsistent state.

### DIFF
--- a/crates/marmot-app/src/client/sync.rs
+++ b/crates/marmot-app/src/client/sync.rs
@@ -536,7 +536,9 @@ impl AppClient {
         // (`restore_disband_tombstone`), and that replay is the only reconciler
         // left for a disband whose live-session projection never completed — a
         // crash, or a batch that failed after the engine had already drained the
-        // event. So this seam owes the same terminal sweep as the inbound one.
+        // event. So this seam owes the same terminal sweep as the inbound one,
+        // which it discharges by running the shared
+        // `observe_event_projection_effects` below rather than a copy of it.
         let local_account_id_hex = self
             .app
             .account_home()
@@ -621,7 +623,8 @@ impl AppClient {
                 updated_group.as_ref(),
                 &source_message_id_hex,
             );
-            self.invalidate_terminal_pending_sends(event, &local_account_id_hex, &mut summary)?;
+            routes_dirty |=
+                self.observe_event_projection_effects(event, &local_account_id_hex, &mut summary)?;
             let can_ack_application_event = if crosses_frontier {
                 self.prepare_local_group_deletion_frontier_clear(
                     event,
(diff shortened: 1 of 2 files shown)


Datawav (commented):
## Automated review

Outcome: 1 actionable finding.

1. Minor — Hydration-replayed disbands trigger a redundant subscription rebuild on every startup
- Exact changed-line evidence: `crates/marmot-app/src/client/sync.rs`, hunk `@@ -621,7 +623,8`, replaces the drained seam’s direct `invalidate_terminal_pending_sends(...)` call with:
  ```rust
  routes_dirty |=
      self.observe_event_projection_effects(...)?;
  ```
  The extracted helper returns `true` for `GroupStateChange::GroupDisbanded`, while the unchanged batch epilogue calls `sync_runtime_groups()` whenever `routes_dirty || routes_changed`.
- Triggering state: a persisted disband tombstone is re-emitted by `restore_disband_tombstone` during hydration. The startup path has already completed `prepare_transport_for_sync`, and `refresh_group_routes()` reports no routing change because the disband retains the same transport route.
- Consequence: every subsequent open performs a second, unchanged group-subscription registration. Any failure in that redundant network operation fails the drained batch and startup despite the replayed projection writes being idempotent.
- Confidence: High.
- Smallest sound correction: preserve the helper’s token cleanup, self-membership, invalidation, and pending-send effects, but do not propagate its disband-only `routes_dirty` result from the drained seam. Continue rebuilding when `state.groups.len()` changes or `refresh_group_routes().routing_changed` is true.

Lens disposition:
- Retry/duplicate delivery: finding above.
- Partial or malformed peer data: no changed-path defect found; the helper extraction preserves the existing authenticated-event handling and error policy.
- Version/wire compatibility: no wire-format or protocol-version change.
- Tests: the new production-path tests cover the previously omitted projection effects and replay idempotency. The advisory’s inbound-then-drained coverage observation does not establish a behavioral defect and is not retained.
- Discussion: the sole non-member automated comment reports no actionable issue and supplies no contradictory technical evidence.

jgmontoya:
> ## Automated review
> 
> Outcome: 1 actionable finding.
> 
>     1. Minor — Hydration-replayed disbands trigger a redundant subscription rebuild on every startup
> 
> 
>     * Exact changed-line evidence: `crates/marmot-app/src/client/sync.rs`, hunk `@@ -621,7 +623,8`, replaces the drained seam’s direct `invalidate_terminal_pending_sends(...)` call with:
>       ```rust
>       routes_dirty |=
>           self.observe_event_projection_effects(...)?;
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       The extracted helper returns `true` for `GroupStateChange::GroupDisbanded`, while the unchanged batch epilogue calls `sync_runtime_groups()` whenever `routes_dirty || routes_changed`.
> 
>     * Triggering state: a persisted disband tombstone is re-emitted by `restore_disband_tombstone` during hydration. The startup path has already completed `prepare_transport_for_sync`, and `refresh_group_routes()` reports no routing change because the disband retains the same transport route.
> 
>     * Consequence: every subsequent open performs a second, unchanged group-subscription registration. Any failure in that redundant network operation fails the drained batch and startup despite the replayed projection writes being idempotent.
> 
>     * Confidence: High.
> 
>     * Smallest sound correction: preserve the helper’s token cleanup, self-membership, invalidation, and pending-send effects, but do not propagate its disband-only `routes_dirty` result from the drained seam. Continue rebuilding when `state.groups.len()` changes or `refresh_group_routes().routing_changed` is true.
> 
> 
> Lens disposition:
> 
>     * Retry/duplicate delivery: finding above.
> 
>     * Partial or malformed peer data: no changed-path defect found; the helper extraction preserves the existing authenticated-event handling and error policy.
> 
>     * Version/wire compatibility: no wire-format or protocol-version change.
> 
>     * Tests: the new production-path tests cover the previously omitted projection effects and replay idempotency. The advisory’s inbound-then-drained coverage observation does not establish a behavioral defect and is not retained.
> 
>     * Discussion: the sole non-member automated comment reports no actionable issue and supplies no contradictory technical evidence.


The trigger is real but the consequence isn't. Verified by probe: disband tombstones do replay on every open (no consumption API exists), so this PR does add a forced `sync_runtime_groups()` per startup — but that sync performs no network operation. The disbanded group is excluded from both sides of the subscription diff (`routing_for` filters tombstoned groups at seed time — the route is never installed on reopen, not "retained"), `subscribe_all` spawns nothing on an empty add-set, and an injected relay failure on the reopen path never fires. Nothing can fail startup here; the per-open cost is a no-op reconcile.

Suppressing the disband-only `routes_dirty` on the drained seam would also remove that seam's only re-fire edge after a *failed* sync — unlike the inbound seam it never arms `pending_runtime_group_subscription_refresh` — and would special-case one event at one caller of the shared helper, reintroducing the seam asymmetry this PR removes. Keeping the propagation symmetric is deliberate.

Two real items your finding led to, tracked separately: tombstone replay-on-every-open is the root cause and wants consumption semantics at the engine layer, and the drained epilogue's missing refresh-arming is a genuine asymmetry vs inbound.

PR #1544, opened 2026-08-24
Send and fetch custom app events in chat

dannym-arx:
marmot

Marmots can now mail each other events the protocol doesn't know about.

`MarmotAppEvent` kinds were a closed set: chat, reactions, edits, deletes, agent activity, group system, push tokens. Apps that wanted to ship their own payloads inside the encrypted MLS application stream had no send path, and no clean way to query what they stored.

This PR adds a `Custom` message intent that carries any kind, tags, and content verbatim, wired through every surface:

- marmot-app: `AppClient::send_custom_event` and `MarmotAppRuntime::send_custom_event`, using the same durable outbox flow as chat sends. Kinds MDK owns (chat, reaction, edit, delete, agent, group system, push token) are rejected through `is_reserved_app_event_kind`, so apps can't forge protocol events. Audit action: `send_custom_event`. No notification trigger by default.
- storage-sqlite: custom kinds project as standalone timeline rows under a new `TimelineUpdateTrigger::CustomEvent`. They never fold onto a target message; reaction and delete behavior stays exclusive to those kinds. `StoredAppMessageQuery` gained a `kinds` filter, compiled into the SQL as an IN clause.
- MarmotKit (UniFFI): `send_custom_event(group_id, kind, tags, content)`, plus an optional `kinds` filter on `messages` and `subscribe_messages`.
- CLI: `wn messages send-event <group> <kind> [content]` with repeatable `--tag '["name","value"]'` options. `messages list` and `messages subscribe` take repeatable `--kind` filters, and the daemon applies the same filter to hub broadcasts.

Inbound events of unknown kinds were already stored durably; now you can fetch them back by kind on every surface, and they show up in the materialized timeline.

## Summary by CodeRabbit

* **New Features**
  * Added support for sending custom application events with content and tags through the CLI and MarmotKit APIs.
  * Added event-kind filters for message listing, history queries, and live subscriptions.
  * Custom events now appear as standalone timeline entries and emit timeline updates.

* **Bug Fixes**
  * Message replay, recovery, and history queries now include all event kinds by default.
  * Reserved event kinds and invalid tags are rejected with clear errors.


crates/marmot-app/src/runtime/subscriptions.rs
```diff
@@ -1709,6 +1709,7 @@ impl MarmotAppRuntime {
 pub(crate) fn messages_recovery_query(query: &AppMessageQuery) -> AppMessageQuery {
     AppMessageQuery {
         group_id_hex: query.group_id_hex.clone(),
+        kinds: query.kinds.clone(),
```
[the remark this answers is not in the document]
dannym-arx: this has been fixed

crates/storage-sqlite/src/timeline.rs
```diff
@@ -1522,7 +1504,11 @@ fn project_single_message_timeline_tx(
             }
             timeline_row_from_stream_start(&event)
         }
-        _ => return Ok(None),
+        // Modifier kinds never get a row of their own: reactions aggregate onto
+        // their target and deletes tombstone it. Every other kind — including
+        // app-defined custom kinds — projects a generic row.
+        MARMOT_APP_EVENT_KIND_REACTION | MARMOT_APP_EVENT_KIND_DELETE => return Ok(None),
+        _ => timeline_row_from_app_event(&event),
```
erskingardner: Custom tags are supposed to remain app-owned, but `timeline_row_from_app_event` always derives `reply_to_message_id_hex` from the first `e` tag. Thus the custom event covered by `custom_kind_event_with_e_tag_still_projects_as_standalone_row` is also exposed as a native reply, gets a hydrated preview, and creates MDK reply dependencies even though that `e` tag may have unrelated app semantics. Please build custom rows without interpreting their tags as MDK reply metadata (while still allowing ordinary chat replies to target a custom row), and extend the `e`-tag regression test to assert `reply_to_message_id_hex`/`reply_preview` stay unset.

erskingardner (changes requested):
The custom-event send/query path and live kind filtering are otherwise coherent, and CI is green. One timeline semantic leak needs fixing before merge: app-owned custom `e` tags are currently reinterpreted as native reply metadata.

crates/storage-sqlite/src/timeline.rs
dannym-arx: Fixed in 955e24b9 — custom kinds now project through `timeline_row_from_custom_event`, which never derives `reply_to_message_id_hex` from tags (both the single-row and batch projection paths). The regression test now asserts `reply_to_message_id_hex`/`reply_preview` stay unset. Chat replies targeting a custom row are unaffected since reply linkage comes from the replying event's own tags.

Datawav (commented):
## Automated review

FAIL — one actionable finding remains at head `955e24b993ba3096e143c8c060c13b34beded9ba`.

1. Major — Reserved push-token events leak into conversation timelines
- Evidence: `crates/storage-sqlite/src/timeline.rs:1510-1511` now excludes only reaction/delete events and sends every other unmatched kind through `timeline_row_from_custom_event`. The batch path at `:2871-2873` applies the same catch-all projection. Push-token update, list, and removal remain reserved protocol kinds without dedicated timeline branches.
- Triggering state: local `share_push_registration` or inbound push-token processing records kind update/list/removal events for a group.
- Consequence: token-management payloads become standalone conversation rows, affect ordering and pagination, and emit `CustomEvent` updates to timeline consumers despite not being app-defined custom events.
- Confidence: High. The changed catch-all necessarily includes these reserved kinds; the custom-send rejection does not protect internally generated or received protocol events.
- Smallest sound correction: explicitly skip all three push-token kinds in both projection paths, leaving the catch-all and `CustomEvent` trigger only for non-reserved kinds. Add incremental and full-rebuild regression cases proving push-token events remain stored/queryable but never materialize in the timeline.

The reported live `kinds` filtering defect and custom `e`-tag reply-metadata leak are fixed at this head and are not retained as findings.

erskingardner (approved):
The prior custom `e`-tag finding is fixed in both incremental and full-rebuild projection paths, with regression coverage for reply metadata. The follow-up push-gossip lifecycle assertion is consistent with the existing send/receive diversion boundary. No further code findings at this head. One unrelated CI job should be rerun before merge: Hermes installer auth hit GitHub HTTP 429 while cloning NousResearch/hermes-agent after its test suite passed.

PR #1548, opened 2026-08-24
Fix/backfill drain eose

jgmontoya:
## Problem

A stalled device's epoch-gap backfill — mdk's only unfloored history fetch — drained with wall-clock timeouts: 750 ms for the first event, 250 ms between events, no end-of-stored- events gate, no connectivity check, and a timeout recorded as `succeeded`, which then called `mark_replayed()` and stood the detector down. Field-verified on a live group: a handset lost that race four times (`deliveries: 0` in 763–1415 ms, filed as success every time) and stayed three epochs behind for days, while a long-lived daemon won the same race against the same relays and healed 13→16 in 3m33s. The ordinary sync path can never fetch that backlog — its `since` floor rides a cursor that undecryptable traffic itself ratchets forward — so the backfill losing quietly meant no recovery path at all.

## Fix

The drain core is shared with two named completion contracts: ordinary sync keeps byte-identical quiescence behavior; the backfill completes on end-of-stored-events from every live subscription (inbox plus one per group route, ≥1 relay each), with a 30 s budget spent on consecutive silence only — every delivery resets it, so hour-long working replays are never truncated while a dead relay cannot hold the drain forever. Zero live subscriptions or zero EOSE by the budget is a failure, never success.

An unconfirmed drain no longer claims a replay: no `mark_replayed()`, `succeeded: false` with a truthful `error_kind`, and the intent requeues through the existing retry machinery. Review hardened the failure path into something survivable: retries back off 15 s doubling to a 5-minute cap keyed on the already-carried `retry_ordinal` (paced attempts skip without burning the drain budget or writing audit rows; explicit user catch-up bypasses the cooldown), and after three unconfirmed attempts the drain falls back to the pre-gate quiescence contract so one unreachable relay cannot pin an account's recovery forever — recorded distinctly as `completion_kind: quiescence_fallback`, so forensics can always tell an EOSE-confirmed replay from a degraded one from a failure.

Known follow-ups tracked separately: `repair_full_history` still drains its own unfloored activation on quiescence, and EOSE telemetry carries no subscription generation.

## Tests

Fail-first throughout: the direct field regression (first stored event landing after 750 ms is collected), EOSE ends the drain promptly, no-EOSE stays pending and retryable with `retry_ordinal` advancing, partial EOSE (inbox served, groups not) stays pending, paced retries skip inside the cooldown, the spent gate falls back and still heals reachable history, the backoff schedule is table-pinned, and the adapter's account-wide EOSE tracking follows the live subscription set across reactivations (mutation-verified). 

934 marmot-app tests, 65 adapter, 124 serial relay_runtime, 29 forensics schema-lockstep.

## Summary by CodeRabbit

- **Improvements**
  - Improved epoch-gap recovery with clearer completion detection, bounded retries, exponential backoff, and quiescence fallback.
  - Partial recovery results remain available while unfinished work can retry automatically.
  - Added relay progress tracking to confirm when stored events have been fully delivered.
  - Recovery and synchronization records now distinguish delivered and skipped items.

- **Audit Logs**
  - Backfill completion records indicate whether recovery finished through stored-event confirmation or fallback detection.

- **Testing**
  - Expanded coverage for recovery timing, partial delivery, retries, relay lifecycle, and reliability scenarios.


crates/marmot-app/src/client/sync.rs
```diff
@@ -934,8 +1097,22 @@ impl AppClient {
...
+                        {
+                            break verdict;
+                        }
+                        continue;
+                    }
+                },
             };
+            // Any delivery proves the stream is alive, including one this drain
+            // goes on to skip as an echo or a duplicate.
+            silence_started = std::time::Instant::now();
```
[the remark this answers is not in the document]
jgmontoya: Verified by probe — the hang is real, and sharper than described: the completion verdict and the budget both live in the timeout arm, so while deliveries flow the drain cannot exit even once every subscription has EOSEd (reproduced: gate fully satisfied, duplicates every 100 ms, drain held open 10× the budget). No adversary needed — the unfloored fetch re-receives the whole account history newest-first, so a long already-seen prefix is the expected shape.

A total deadline is the wrong shape though: checked in the timeout arm it never fires (same reachability problem), and hoisted to the delivery path it truncates exactly the long working replays this path exists to complete — the 59-minute field drain was ≥99.5% skipped receives en route to its novel tail, so no safe value exists. Progress-gated silence reset fails the same test (probe: 1 of 11 novel events collected on the field shape).

Fixed instead: on a skipped delivery the drain polls the completion verdict (rate-limited to once per SDK_DRAIN_WAIT) and breaks on Complete — it stops the moment the relays confirm the history was served, and nothing legitimate is truncated. Drain audit rows now carry a skipped-receives counter beside deliveries, so a long field drain is distinguishable from a spinning one. A guard test pins the rejected remedy: anyone later bounding the no-EOSE duplicate storm by moving the silence reset breaks `epoch_backfill_drain_collects_novel_history_that_trickles_slower_than_its_budget`. A relay that streams non-novel events forever and never EOSEs remains bounded only by shutdown's 5 s abort — filed separately rather than solved with a cap the field data rules out.

crates/transport-nostr-adapter/src/lib.rs
```diff
@@ -1381,11 +1434,30 @@ impl AdapterState {
...
     }
 
+    /// End-of-stored-events progress across an account's live subscriptions.
+    fn account_subscription_eose(&self, account_id: &MemberId) -> AccountSubscriptionEose {
+        let ids = self.account_subscription_ids(account_id);
+        AccountSubscriptionEose {
+            subscriptions: ids.len(),
+            with_eose: ids
+                .iter()
+                .filter(|id| self.sync.subscription_any_eose(id).unwrap_or(false))
```
erskingardner, line 1456: [P1] Do not equate one relay EOSE with complete stored history

`subscription_any_eose` makes the account gate complete as soon as one endpoint per subscription sends EOSE. That is not evidence that the account history was served: MDK publishes with `required_acks: 1` by default, so a missing commit may exist only on relay B while faster relay A has no copy and immediately EOSEs. Once every subscription gets A’s EOSE, `run_pending_epoch_backfill` records `EndOfStoredEvents`, calls `mark_replayed()`, and can stop before B delivers—the same false-success/disarmed-detector failure this PR is intended to remove. Track completion for the relay(s) that can carry the history (or wait for all live endpoints and use the already-typed quiescence fallback when that cannot clear), and add a two-relay regression where the fast relay EOSEs empty before the slow relay delivers the recovery commit.

erskingardner (changes requested):
The EOSE-based completion gate can still declare a replay complete before the relay holding the missing history has drained. Please make the completion claim reflect relay coverage rather than one arbitrary relay per subscription; the existing bounded fallback can remain the explicitly degraded path.

erskingardner (approved):

crates/marmot-app/src/client/sync.rs
```diff
@@ -1588,6 +1836,9 @@ impl AppClient {
         if !self.has_pending_epoch_backfill() {
             return Ok(EpochBackfillRunOutcome::NotPending);
         }
+        if self.epoch_backfill_retry_is_paced(seam) {
```
erskingardner, line 1839: Non-blocking scheduling note (also left in the review summary): this paced check returns `Deferred` before `begin_epoch_backfill_execution` gets a chance to rotate, so while the primary intent is cooling down, queued sibling intents behind it are parked too rather than rotated forward. Bounded by the 5-minute cap and arguably correct (the account-wide replay budget is the contended resource), but worth a comment in the code since the rotation helpers below imply otherwise.

crates/marmot-app/src/tests.rs
```diff
@@ -1090,6 +1180,692 @@ fn failed_epoch_backfill_activation_retains_one_correlated_retry() {
...
+        let drained_in = started.elapsed();
+        stop.store(true, std::sync::atomic::Ordering::Relaxed);
+        let _ = pump.await;
+
+        assert!(
+            matches!(outcome, crate::EpochBackfillRunOutcome::Completed(_)),
+            "a served history must complete even while duplicates arrive"
+        );
+        assert!(
+            drained_in < Duration::from_secs(3),
```
erskingardner, line 1437: Nit, non-blocking: wall-clock assertions like `< 3s` here (and `< 500ms` in the pacing test) can flake on heavily loaded shared CI runners. The margins look generous today, so fine to keep — just flagging them as the first suspects if CI turns flaky.

crates/marmot-app/src/client/sync.rs
```diff
@@ -1508,7 +1752,8 @@ impl AppClient {
         execution: EpochBackfillExecution,
         activation_outcome: EpochBackfillActivationOutcome,
         error_kind: Option<String>,
-        deliveries: u64,
+        completion_kind: Option<EpochBackfillCompletionKind>,
```
erskingardner, line 1755: For the outstanding CodeRabbit note about clearing `completion_kind` when `observed_all_groups` downgrades the outcome to a failure (the block just below this signature): I verified that path is already sound. A downgraded `succeeded` routes to `EpochStallBackfillFailed`, which has no `completion_kind` field at all, so no failure row can carry `end_of_stored_events` / `quiescence_fallback`; an analyzer counting confirmed completions only sees the Completed variant, which requires `observed_all_groups == true`. Safe to dismiss that suggestion — the proposed fix would be a no-op on a value the failure arm never serializes.

crates/marmot-app/src/client/sync.rs
```diff
@@ -921,40 +1131,70 @@ impl AppClient {
...
                             StagedSyncError::new(error.into(), SyncFailureStage::RelayReceive),
                             drain_started,
                             cursor_before_secs,
                         )
                         .await);
                 }
-                Err(_) => break,
+                Err(_) => match completion {
+                    DrainCompletion::Quiescence => break DrainVerdict::Complete,
+                    DrainCompletion::QuiescenceFallback => break DrainVerdict::QuiescenceFallback,
```
erskingardner, line 1157: Non-blocking observation, fine as-is: the fallback contract never re-consults the gate, so a `QuiescenceFallback` drain during which every subscription happens to get EOSE still records `quiescence_fallback`. That under-claims rather than over-claims, which is the safe direction for forensics, so no change needed — just noting it is a deliberate-looking asymmetry worth keeping if ever revisited.

erskingardner (approved):
Reviewed at head 3862f5b (all CI green). **Recommend merge.**

**What I verified**

- **The two-contract drain is sound.** Ordinary sync (`Quiescence`) keeps byte-identical behavior; `EndOfStoredEvents(budget)` only polls the gate on the receive-timeout arm and spends silence budget, so working replays are never truncated; the skipped-delivery gate poll (rate-limited to `SDK_DRAIN_WAIT`) closes the CodeRabbit-probed hang where a duplicate stream prevented the timeout arm from ever firing. The `Err(_)`/`Ok(Ok(None))`/skip paths all map to truthful verdicts.
- **No more false success.** An unconfirmed drain records `epoch_stall_backfill_failed` with a real `error_kind`, skips `mark_replayed()` (the bystander-group test pins that an unconfirmed replay can't disarm other groups' arms), and requeues through the existing machinery. Backoff is a pure function with a table-pinned schedule (15s doubling to the 5m cap, shift clamped), and the attempt limit falls back to the explicitly-degraded quiescence contract with a distinct `completion_kind` — overclaim is structurally impossible on the Completed row.
- **Forensic schema stays compatible.** `skipped` / `completion_kind` are additive-optional, the schema `required` arrays are untouched, and a legacy-row parse test pins backward compatibility. The `skipped` split directly answers the open question from the 59-minute field drain.
- **The prior blocking review is addressed as designed.** The "≥1 relay per subscription" gate is deliberate and documented (all-relay would conflate a permanently unreachable relay with unfinished history), and the bounded fallback is the degraded exit the reviewer asked to keep.
- **Invariants respected:** the new `tracing::warn!` carries only aggregate counts and kinds (no ids/endpoints), engine/transport boundaries untouched, no version bumps.

**Non-blocking notes** (inline): the outstanding CodeRabbit `completion_kind` suggestion is a no-op false positive (verified — the downgrade path routes to a variant with no such field); and a fallback drain never re-consults the gate, so it under-claims when EOSE arrives mid-fallback (safe direction, fine as-is).

Strong test posture throughout: fail-first regressions for the field shape, prompt-EOSE completion, partial EOSE, pacing, fallback, and the guard test (`...trickles_slower_than_its_budget`) that pins the rejected remedy for future contributors. 934 marmot-app tests and the full matrix pass.

PR #1549, opened 2026-08-25
Device sync, simplified (discussion) [skip-changelog]

dannym-arx:
read docs/device-sync-simplified.md

## Summary by CodeRabbit

* **Documentation**
  * Added a draft specification for synchronizing device data across a mesh of devices.
  * Documented device roles, joining and leaving, TCP-based peer synchronization, and recovery from missed updates.
  * Defined category merging, conflict handling, transactional updates, versioned messaging, and error handling.
  * Documented hashing, canonicalization, and security validation requirements.
  * Clarified admin-signed device enrollment and complete roster synchronization.
  * Added a Mermaid sequence diagram and outlined initial-version exclusions.

### DIFF
--- a/docs/device-sync-simplified.md
+++ b/docs/device-sync-simplified.md
@@ -106,19 +106,25 @@ variable-length.
 
 ### Session flow
 
-```
-A                                   B
-|-- HELLO (digests) --------------->|
-|<-- HELLO (digests) ---------------|
-        digests equal? -> DONE, session over (common case)
-|-- SUMMARY (categories that differ)|
-|<-- SUMMARY -----------------------|
-|-- WANT (ids B has, A lacks) ----->|
-|<-- WANT --------------------------|
-|-- DATA* ------------------------->|
-|<-- DATA* -------------------------|
-|-- DONE -------------------------->|
-|<-- DONE --------------------------|
+```mermaid
+sequenceDiagram
+    participant A as Device A
+    participant B as Device B
+    A->>B: HELLO (digests)
+    B->>A: HELLO (digests)
+    alt digests equal (common case)
+        A->>B: DONE
+        B->>A: DONE
+    else digests differ
+        A->>B: SUMMARY (categories that differ)
+        B->>A: SUMMARY
+        A->>B: WANT (ids B has, A lacks)
+        B->>A: WANT
+        A->>B: DATA (chunked, repeated)
+        B->>A: DATA (chunked, repeated)
+        A->>B: DONE
+        B->>A: DONE
+    end
 ```
 
 Category digest = hash of sorted item hashes, so equal digests skip the
--- a/docs/device-sync-simplified.md
+++ b/docs/device-sync-simplified.md
@@ -9,7 +9,8 @@ server infrastructure; devices talk directly over a private mesh.
 ## Roles
 
 - **Master**: the first device the account is created on. It creates the
-  mesh network and issues join credentials. Master is an *administrative*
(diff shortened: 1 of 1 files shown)


erskingardner:
Thanks for writing this up — it's a genuinely useful forcing function, and several pieces of it are worth keeping regardless of where the overall sync design lands. Context for this comment: I just reviewed the new spec-side multi-device/account-sync documents (marmot-protocol/marmot#417, which transcribes the agreed baseline from mdk#1518), so I'm comparing this against that boundary work.

**What's strong here**

- Digest-first anti-entropy is the right steady-state shape: equal category digests short-circuit to two HELLO frames per interval.
- The canonical-hashing section is unusually good for a draft — domain-separated item hashes, "transmitted bytes ARE the canonical bytes," and conformance vectors pinning digests. That's exactly how you make "two devices holding the same items MUST produce identical digests" testable.
- Chunked DATA frames, transactional per-category merges, and the nods to the repo invariants (GroupId vs route-id, torn-write-free merges, observability, dial-safety) all show the right instincts.

**Structural concerns — as *the* sync path**

1. **A second, parallel device-authority plane.** The nostr-vpn roster (master-signed) decides "my devices" for sync while MLS same-account leaves decide it for groups. That's two enrollment roots and two revocation mechanisms with no defined reconciliation when they diverge (e.g., a device removed from groups via same-account Remove but still in the roster, or vice versa). The point of the same-account baseline was collapsing to one authority — account key + current leaf — plus a pairing ceremony. This design reintroduces the split it was avoiding.

2. **Master is a permanent v1 single point of failure.** Lose the first device and you can never enroll or revoke again; promotion is explicitly out of scope. That's the same availability trap the old admin-signature External-Commit draft was withdrawn over — the authority you need most is exactly the one a lost device held.

3. **Trust boundary = mesh membership.** The tunnel authenticates the peer, but there is no per-record authentication: any compromised member device can inject MSG items, tombstone every KP, or win GRP last-write-wins. The account-sync exploration's versioned sync keys and per-device signing identities exist precisely to contain a malicious same-account device; this design has no equivalent containment.

4. **GRP merged by (epoch, timestamp) is the untrusted-wall-clock pattern** the account-sync doc explicitly warns against ("last writer cannot mean an untrusted wall-clock timestamp alone"). A malicious or clock-skewed device can roll group metadata backward or forward. Epoch alone is monotonic per group; timestamp adds nothing an attacker can't claim.

5. **No offline story.** TCP-over-tunnel requires both devices online simultaneously. The hardest sync case — restore a new/replacement device when no peer is online — is unaddressed, and a persistent TCP-listener mesh is dubious on mobile (background execution, battery). The layered design's journal + checkpoint carriers exist largely for this case.

6. **Syncing GRP membership is mostly redundant with enrollment.** Per-group MLS same-account Add already gives every device canonical membership; the hard part of sync is accepted history and presentation state, which is where this merge table is thinnest (MSG union by id doesn't cover edits/deletes/reactions; read positions, pins, drafts are absent).

**Suggested framing**

I'd treat this not as a competing sync architecture but as a candidate *carrier* for the layered design in the account-sync exploration — specifically a strong shape for the "direct pairing transfer" first milestone or a peer-to-peer carrier at layer 5, minus the nostr-vpn master/roster plane. Concretely: keep the digest exchange, canonical hashing with vectors, and chunked framing; drop the roster as a membership authority (enrollment authority stays with pairing + same-account MLS); and add per-record authentication from the sync-key/device-signing layer so the trust boundary is record-level, not tunnel-level.

Happy to help sketch what that reconciliation looks like if useful.

PR #1550, opened 2026-08-25
Select the newest valid KeyPackage for group invites

erskingardner:
## Summary

- evaluate relay-fetched KeyPackage publications newest-first
- return immediately on the newest valid current-profile package
- skip malformed candidates while retaining the newest validation error when no package is usable
- add regressions for older-malformed/newer-valid and newer-malformed/older-valid publication sets

## Context

A user could not invite a friend because the friend had an older incompatible kind-30443 KeyPackage in a separate slot alongside a newer valid current-profile package. The previous oldest-first traversal parsed the stale publication first and aborted before reaching the usable package.

## Validation

- `cargo test -p marmot-app member_key_package_resolution_ -- --nocapture`
- `just fast-ci`

A broader pre-rebase `cargo test -p marmot-app` run passed all 747 unit tests (1 ignored), but its relay-runtime integration binary had five failures outside the KeyPackage path. One of those unrelated encrypted-media failures reproduced in isolation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved KeyPackage resolution when publications are malformed or invalid.
  * Newer valid publications are now selected over older malformed ones.
  * Older valid publications are used when the latest publication cannot be processed.
  * Legacy-profile packages are excluded from current-profile resolution.

### DIFF
--- a/crates/marmot-app/src/key_package_records.rs
+++ b/crates/marmot-app/src/key_package_records.rs
@@ -96,17 +96,22 @@ fn latest_key_package_from_records(
     mut records: Vec<RelayEventRecord>,
 ) -> Result<FetchedKeyPackage, AppError> {
     sort_directory_records(&mut records);
-    let mut latest = None;
-    for record in records {
+    let mut newest_error = None;
+    for record in records.into_iter().rev() {
         if record.event.kind != KIND_MARMOT_KEY_PACKAGE || record.event.pubkey != account_id_hex {
             continue;
         }
-        let fetched = key_package_from_record(record)?;
-        if fetched.key_package.protocol_profile == ProtocolProfile::Current {
-            latest = Some(fetched);
+        match key_package_from_record(record) {
+            Ok(fetched) if fetched.key_package.protocol_profile == ProtocolProfile::Current => {
+                return Ok(fetched);
+            }
+            Ok(_) => {}
+            Err(error) => {
+                newest_error.get_or_insert(error);
+            }
         }
     }
-    latest.ok_or_else(|| AppError::MissingKeyPackage(account_id_hex.to_owned()))
+    Err(newest_error.unwrap_or_else(|| AppError::MissingKeyPackage(account_id_hex.to_owned())))
 }
 
 pub(crate) fn latest_fresh_key_package_from_records(
--- a/crates/marmot-app/src/tests.rs
+++ b/crates/marmot-app/src/tests.rs
@@ -5055,6 +5055,60 @@ async fn malformed_batch_member_does_not_discard_valid_member_prewarm() {
 }
 
 #[tokio::test]
+async fn member_key_package_resolution_ignores_older_malformed_publication() {
+    let (_directory, app, accounts, fetcher) = member_resolution_fixture(1, false).await;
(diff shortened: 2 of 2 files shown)


erskingardner (commented):
Reviewed at head 59c964e. Recommend merge.

Verified the selection semantics against the surrounding code at the head SHA:

- `sort_directory_records` sorts ascending by `created_at` then event id, so `.rev()` yields newest-first with the higher event id winning ties — the same tie-break the old keep-last-valid loop produced, so no ordering regression.
- kind 30443 is addressable (NIP-33), so multiple KeyPackages legitimately coexist in separate `d` slots; treating each valid current-profile candidate independently and picking the newest valid one is the right policy, and it fixes the reported bug where a stale malformed slot aborted resolution via `?`.
- Error semantics are strictly better: when nothing is usable, the newest parse error propagates instead of the oldest, and the `MissingKeyPackage` → `Ok(None)` → cached-fallback distinction in `latest_fresh_key_package_from_records` is preserved (only genuinely empty publication sets fall back to cache).
- No security concern: a relay can inject malformed events but not validly-signed newer ones, so the fallback to an older valid package can't be abused to downgrade selection.
- Both regression tests cover the two orderings (older-malformed/newer-valid and the reverse), and the `created_at ± 1` offsets make the sort deterministic.

Minimal, surgical fix (+59/-6, almost all tests). No findings.

Note: CI was still pending when I reviewed; the PR body reports the targeted tests and `just fast-ci` pass locally, with five pre-existing relay-runtime integration failures outside this path. (Posted as a comment because GitHub won't let the PR author approve their own PR.)

erskingardner (commented):
Independently reviewed at head 59c964e. **Recommend merge.**

Verified against the surrounding code at the head SHA:

- **Ordering equivalence:** `sort_directory_records` sorts ascending by `(created_at, event id)`, so `.rev()` with early return yields the same winner the old keep-last loop produced (newest `created_at`, higher event id on ties). No ordering regression.
- **Error semantics:** when nothing is usable, the newest parse error now propagates instead of the oldest — strictly better diagnostics. The `MissingKeyPackage` → `Ok(None)` → cached-fallback distinction in `latest_fresh_key_package_from_records` is preserved: only genuinely empty or all-legacy publication sets fall back to cache, exactly as before.
- **Security:** selecting the newest *valid* current-profile package can't be downgraded by relays — a relay can inject malformed events but not forge validly-signed newer ones, and each candidate is fully tag/credential/ref-validated in `key_package_from_record` before acceptance.
- **Tests are regression-honest:** both new tests fail pre-fix (older-malformed aborts the oldest-first `?` traversal; newer-malformed aborts after the older valid package is tentatively selected). The cloned event keeps the same kind/pubkey so the mocked fetcher returns it for the account query, and `created_at ± 1` with distinct ids makes the sort deterministic. Asserting `network_resolved_members == 1` (not just `unique_members`) correctly pins the network-resolution path.

Surgical fix: +65/-6, essentially all test code. Matches the author's earlier self-review conclusions. CI is fully green (26 checks; only the CodeRabbit docstring-coverage warning, scoped to the touched private helper and test fns — not a repo policy gate).

(Posted as comment since GitHub won't let the PR author approve their own PR.)

PR #1551, opened 2026-08-25
fix: surface group membership activity in chat list

mubarakcoded:
## Summary
- project membership and admin group-system events into chat-list preview, activity ordering, unread state, and read markers
- emit new-last-message subscription updates only for the five membership/admin event types
- avoid treating locally synthesized group-system actors as Nostr profile identities
- preserve DM behavior, mention counts, invalidation, self-authored unread semantics, and agent-stream triggers

Supports marmot-protocol/whitenoise-android#822 and marmot-protocol/whitenoise-android#2249.

## Verification
- just fast-ci
- cargo test -p storage-sqlite group_system --lib
- cargo test -p storage-sqlite every_membership_and_admin_system_type_signals_activity --lib
- cargo test -p storage-sqlite direct_conversation_does_not_project_admin_activity --lib
- cargo test -p marmot-app group_system_projection_advances_the_chat_list_like_a_new_message --lib
- cargo test -p marmot-app unrelated_group_system_projection_does_not_claim_new_chat_list_activity --lib
- cargo test -p marmot-app group_system_chat_preview_does_not_hydrate_its_optional_actor_as_a_nostr_sender --lib
- cargo test -p marmot-app agent_stream_updates_still_advance_the_chat_list_like_new_messages --lib

## Summary by CodeRabbit

- **New Features**
  - Group membership and administrator activity now appears in group chat previews, timestamps, unread counts, and read markers.
  - Relevant system activity updates chat lists in real time.
  - Local-authored system events update previews without increasing unread counts.
  - Legacy chat-list data is rebuilt to include eligible group activity.

- **Bug Fixes**
  - Prevented system messages from being treated as sender profiles or generating mentions.
  - Excluded unrelated rename events and direct-chat system activity from chat lists.
  - Improved unread activity handling after member removal.

### DIFF
--- a/crates/marmot-app/src/runtime/subscriptions.rs
+++ b/crates/marmot-app/src/runtime/subscriptions.rs
@@ -6,7 +6,11 @@ use std::future::pending;
 use std::sync::{Arc, Mutex as StdMutex};
 use std::time::Duration;
 
-use cgka_traits::app_event::MARMOT_APP_EVENT_KIND_AGENT_STREAM_START;
+use cgka_traits::app_event::{
+    GROUP_SYSTEM_TYPE_ADMIN_ADDED, GROUP_SYSTEM_TYPE_ADMIN_REMOVED, GROUP_SYSTEM_TYPE_MEMBER_ADDED,
+    GROUP_SYSTEM_TYPE_MEMBER_LEFT, GROUP_SYSTEM_TYPE_MEMBER_REMOVED, GROUP_SYSTEM_TYPE_TAG,
+    MARMOT_APP_EVENT_KIND_AGENT_STREAM_START,
+};
 use cgka_traits::{GroupId, MessageId, storage::StorageError};
 use serde::{Deserialize, Serialize};
 use storage_sqlite::AppEventReplayCursor;
@@ -661,20 +665,40 @@ impl ChatListUpdateTrigger {
         }) {
             return Self::LastMessageDeleted;
         }
-        if changes.iter().any(|change| {
-            matches!(
-                change,
-                TimelineMessageChange::Upsert {
-                    trigger: TimelineUpdateTrigger::NewMessage
-                        | TimelineUpdateTrigger::AgentStreamStarted
-                        | TimelineUpdateTrigger::AgentStreamFinished,
-                    ..
-                }
-            )
-        }) {
+        if changes.iter().any(change_advances_chat_list) {
             return Self::NewLastMessage;
         }
         Self::SnapshotRefresh
+    }
+}
+
+fn change_advances_chat_list(change: &TimelineMessageChange) -> bool {
+    match change {
+        TimelineMessageChange::Upsert {
+            trigger:
+                TimelineUpdateTrigger::NewMessage
+                | TimelineUpdateTrigger::AgentStreamStarted
+                | TimelineUpdateTrigger::AgentStreamFinished,
+            ..
+        } => true,
+        TimelineMessageChange::Upsert {
+            trigger: TimelineUpdateTrigger::GroupSystem,
+            message,
+        } => message.tags.iter().any(|tag| {
+            tag.first()
+                .is_some_and(|name| name == GROUP_SYSTEM_TYPE_TAG)
+                && tag.get(1).is_some_and(|system_type| {
+                    matches!(
+                        system_type.as_str(),
+                        GROUP_SYSTEM_TYPE_MEMBER_ADDED
+                            | GROUP_SYSTEM_TYPE_MEMBER_REMOVED
+                            | GROUP_SYSTEM_TYPE_MEMBER_LEFT
+                            | GROUP_SYSTEM_TYPE_ADMIN_ADDED
+                            | GROUP_SYSTEM_TYPE_ADMIN_REMOVED
+                    )
+                })
+        }),
+        _ => false,
     }
 }
 
--- a/crates/marmot-app/src/runtime/subscriptions.rs
+++ b/crates/marmot-app/src/runtime/subscriptions.rs
@@ -642,7 +642,10 @@ pub enum ChatListUpdateTrigger {
 }
 
 impl ChatListUpdateTrigger {
-    pub(crate) fn from_timeline_changes(changes: &[TimelineMessageChange]) -> Self {
+    pub(crate) fn from_timeline_changes(
+        changes: &[TimelineMessageChange],
+        projects_group_system_activity: bool,
+    ) -> Self {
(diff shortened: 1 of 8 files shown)


crates/storage-sqlite/src/chat_list.rs
```diff
@@ -961,12 +964,42 @@ fn refresh_chat_list_row_tx(
...
+    GROUP_SYSTEM_TYPE_MEMBER_REMOVED,
+    GROUP_SYSTEM_TYPE_MEMBER_LEFT,
+    GROUP_SYSTEM_TYPE_ADMIN_ADDED,
+    GROUP_SYSTEM_TYPE_ADMIN_REMOVED,
+];
+
+/// SQL predicate for activity that should behave like a chat message in the
+/// chat list. Kind-1210 also carries metadata changes; #822 intentionally
+/// promotes only membership/admin rows, leaving unrelated system events alone.
+fn chat_list_activity_filter_sql(column_prefix: &str) -> String {
```
erskingardner: [P1] Keep secure-prune repair on the same activity predicate

This makes kind-1210 membership/admin rows eligible as the latest chat-list activity, but timeline.rs::refresh_chat_list_last_message_after_secure_prune_tx still queries kind = 9 only (lines 1301-1315 at this head). If an expiring chat is pruned after one of these system rows became the preview, that repair overwrites the live preview with an older chat or NULL even though the system row remains; it only self-heals on a later full projection check. Reuse this predicate (or a shared latest-activity helper) in the secure-prune preview repair and add a regression with a retained group-system row plus a pruned chat.

erskingardner (changes requested):
The normal projection and upgrade-reconciliation paths consistently promote the five membership/admin system rows, but the secure-prune repair path still applies the previous chat-only contract. Please fix that lifecycle gap before merge.

crates/storage-sqlite/src/chat_list.rs
mubarakcoded: Fixed in 0339f347. Secure-prune preview repair now reuses the same chat-list activity predicate as rebuild/unread selection, so a retained membership/admin system row remains the preview when an older chat is pruned. Added secure_prune_preserves_a_newer_group_system_preview; all 9 secure-prune tests pass, the complete 124-test relay_runtime job passes serially with CI flags, and just fast-ci passes.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1725,16 +1770,17 @@ fn timeline_message_for_read_marker_tx(
...
-         WHERE group_id_hex = ?1 AND message_id_hex = ?2 AND kind = ?3",
-        params![
-            group_id_hex,
-            message_id_hex,
-            u64_to_i64(MARMOT_APP_EVENT_KIND_CHAT)?
-        ],
+         WHERE group_id_hex = ?1
+           AND message_id_hex = ?2
+           AND {activity_filter}"
+        ),
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in 88c1e5b1. `TimelineReadMarker` now selects and retains the persisted event `kind` and passes it into `canonical_timeline_order_key`. Added a same-epoch kind-1210-to-chat regression that would fail under the old hardcoded phase. Verified all 80 active chat-list tests plus `just fast-ci`.

crates/marmot-app/src/runtime/subscriptions.rs
```diff
@@ -661,23 +665,43 @@ impl ChatListUpdateTrigger {
...
+                    matches!(
+                        system_type.as_str(),
+                        GROUP_SYSTEM_TYPE_MEMBER_ADDED
+                            | GROUP_SYSTEM_TYPE_MEMBER_REMOVED
+                            | GROUP_SYSTEM_TYPE_MEMBER_LEFT
+                            | GROUP_SYSTEM_TYPE_ADMIN_ADDED
+                            | GROUP_SYSTEM_TYPE_ADMIN_REMOVED
+                    )
+                })
+        }),
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in 88c1e5b1. `AppProjectionUpdate` now gates membership/admin `NewLastMessage` classification with the already-refreshed `ChatConversationKind`, so direct conversations use `SnapshotRefresh` while real groups retain activity behavior. Added a direct-conversation regression and retained agent-stream coverage. Verified all 89 runtime tests plus `just fast-ci`.

crates/storage-sqlite/src/timeline.rs
```diff
@@ -1289,6 +1289,7 @@ fn refresh_chat_list_last_message_after_secure_prune_tx(
     tx: &Connection,
     group_id_hex: &str,
 ) -> StorageResult<()> {
+    let activity_filter = crate::chat_list::chat_list_activity_filter_sql("");
```
erskingardner, line 1287: [P1] Preserve pruned system activity in the durable sort anchor

This preview repair now reuses the widened activity predicate, but retain_pruned_chat_activity_tx immediately above still binds kind 9. secure_prune_app_events_before can delete qualifying kind-1210 rows, so their timeline_at is not copied to retained_activity_sort_at and the conversation can fall behind older chats after pruning. Apply the same predicate to anchor retention and add a regression that prunes the qualifying system row itself.

erskingardner (changes requested):
The original secure-prune preview regression is fixed, and the read-marker/direct-trigger follow-ups are correct. One secure-prune ordering gap remains: qualifying system activity is not retained when the system row itself is pruned.

crates/storage-sqlite/src/timeline.rs
mubarakcoded: Fixed in ad24a9d6. `retain_pruned_chat_activity_tx` now reuses the same `chat_list_activity_filter_sql` predicate as secure-prune preview repair, so qualifying kind-1210 rows preserve the durable sort anchor while direct-conversation/unrelated system rows remain excluded. Added a regression that prunes both an older chat and the newer qualifying system row, verifies `retained_activity_sort_at` remains 200, and verifies a subsequent full chat-list rebuild preserves it. All 81 active chat-list tests and `just fast-ci` pass.

erskingardner (approved):
The remaining secure-prune activity-anchor gap is fixed: qualifying pruned group-system rows now use the shared chat-list activity predicate and retain activity_sort_at, with regression coverage through a full rebuild. The earlier preview-repair, read-marker, direct-trigger, unread, invalidation, upgrade, and profile-hydration concerns are also resolved at this exact head. CI was excluded from this review as requested.

PR #1555, opened 2026-08-25
fix(agent): make preview retries idempotent

Datawav:
## Summary

- Make `stream_append`, `stream_status`, and `stream_progress` safe to retry when the 8-second preview request times out after the server has already applied it.
- Reuse one operation key across retries. The server replays Ack for the same fingerprint and returns Conflict when a key is reused with different content.
- Keep preview receipts in a separate 4,096-entry process-local store so preview traffic cannot evict or persist the 1,024-entry durable `send_final` and `stream_finalize` receipts.
- Preserve transcript count and hash alignment so a lost-Ack retry cannot duplicate entries or invalidate `stream_finalize`.

## Design rationale

Preview mutations are meaningful only while the active stream exists. Persisting them beside durable send and finalize receipts generated unnecessary snapshots and allowed ordinary preview traffic to shrink the ambiguous-timeout protection window for durable publication. A dedicated strict preview store preserves same-process lost-Ack recovery without competing with durable receipts.

The server records a preview receipt only after compose succeeds. A same-fingerprint retry returns the completed Ack without composing again. A different fingerprint fails closed. OpenClaw and Hermes keep one idempotency key for all attempts of the same logical mutation and advance their local transcript once.

## Retry and finalize acceptance

- A simulated timeout after successful server apply retries with the original key.
- The retry does not dispatch a second append, status, or progress mutation.
- Client and server transcript counts and hashes remain aligned.
- `stream_finalize` succeeds after retry and retains its separate durable receipt semantics.
- Concurrent same-key requests dispatch compose once and both receive Ack.

## Verification

- Exact-current-master fast gate passed: formatting, naming, convergence ledger, toolchain and documentation gates, workspace checks with default and OTLP features, clippy with default and OTLP features, and convergence policy pin tests.
- `cargo test -p agent-connector`: 149 passed, 1 ignored, 0 failed; identity security integration test passed.
- Hermes adapter suite: 201 passed, including append, status, and progress lost-Ack retry plus finalize alignment.
- OpenClaw typecheck passed; `live.test.ts`: 21 passed, including post-success-timeout retry coverage.
- The branch was rebuilt on current `master`; the PR now contains only the 13 idempotency and integration files.

## Independent review

The first independent audit identified a HIGH defect where preview receipts shared the persisted durable receipt store, plus missing concurrent-server and Hermes parity coverage. This revision isolates the stores and adds those regressions.

The final bounded review selected Cursor Grok 4.6 because Kimi and GLM were unavailable. It returned PASS with zero findings. This is independently reviewed but is not represented as Kimi-reviewed.

## CI

GitHub Actions checks are attached to this pull request. The dependent release-verification stage owns the broad exact-revision matrix. Automation must not merge this pull request.

## Summary by CodeRabbit

* **New Features**
  * Added configurable invite policies, including deny, allowlist, and authenticated access options.
  * Added idempotency support and reliable retries for stream append, status, and progress operations.
  * Stream progress updates are now recorded in transcripts.
* **Bug Fixes**
  * Prevented conflicting or concurrent preview mutations from duplicating or corrupting stream state.
  * Improved cancellation handling during streaming and finalization.
* **Documentation**
  * Updated setup, installation, media authorization, allowlist synchronization, and preview retry guidance.


Datawav (commented):
## Automated review

1. Low severity — cancellation fencing lacks a regression test
   - Evidence: `integrations/openclaw/marmot/src/live.ts:118` adds `ensureOpen()` fencing around retry attempts so a preview cancelled during retry backoff cannot subsequently append or finalize. The changed `integrations/openclaw/marmot/test/live.test.ts` tests lost-ack deduplication, but contains no cancellation/backoff scenario.
   - Trigger: A retryable timeout starts backoff; `cancel()` completes before the retry resumes.
   - Consequence: A future regression removing either fence could dispatch another mutation—or finalize a durable message—after cancellation without any test failing.
   - Smallest correction: Add a `MarmotLivePreview` test using a controllable retry delay: make the first append or finalize attempt fail retryably, cancel during backoff, release the delay, then assert no second request was dispatched and no local transcript/final state advanced. This test would fail against the pre-fence implementation.
   - Confidence: High.

Audit evidence:
- Exact head: `31b5e33c95cc298584a9f89f56635758dc772914`
- Diff coverage: complete; all 13 changed patches were present.
- Changed files inspected include `crates/agent-connector/src/stream.rs`, `crates/agent-connector/src/stream_session.rs`, `integrations/hermes/marmot/adapter.py`, and `integrations/openclaw/marmot/src/live.ts`.
- Discussion validation: the cancellation claim was applicable and corrected in production code, but its regression-test gap remains; unrelated comments against `tests.rs` and `sync.rs` were outside this change and were suppressed.
- Risk lenses checked: assertions fail before the fix; boundary and negative cases; tests exercise production behavior; lifecycle/cancellation and restart semantics; optional-field protocol compatibility. No accessibility or authorization lane was routed.
- Check evidence: available and complete. Every supplied non-skipped GitHub check concluded successfully; `Publish wn-agent release` was skipped.

integrations/hermes/marmot/adapter.py
```diff
@@ -1737,6 +1751,36 @@ def __init__(
...
+            raise AgentControlError(
+                "an ambiguous preview mutation must be reconciled first",
+                code="preview_mutation_pending",
+                retryable=True,
+            )
+        key = pending[2] if pending is not None else uuid.uuid4().hex
+        self._pending_preview_mutation = (operation, payload, key)
+        for attempt in range(len(STREAM_PREVIEW_RETRY_BACKOFF_S) + 1):
+            try:
+                await mutation(key)
```
erskingardner, line 1774: Non-blocking: this loop has no terminal-state check. Hermes `cancel()` sends `stream_cancel` but leaves `finalized` false, so a retryable timeout that is sleeping here will still call `mutation(key)` after cancel returns. OpenClaw fences the same race with `ensureOpen()` and covers it in `live.test.ts`.

Mark the stream closed in `cancel()` and re-check before each attempt (and before committing the local transcript). Same-key server dedup limits the damage when the first apply already landed, but a never-applied mutation can still compose after the user cancelled.

erskingardner (approved):
Recommend merge.

The server contract is right: preview receipts live in a separate process-local 4096-entry store so they cannot evict or snapshot durable `send_final` / `stream_finalize` receipts; capability auth runs before the dedup store; same fingerprint replays Ack without composing; a mismatched key is Conflict; concurrent same-key requests elect one leader. Clients reuse one key, advance the local transcript only after Ack, and keep an exhausted retry as a pending mutation that must be reconciled before a different payload. OpenClaw now fences cancel through retry backoff and has tests for that race.

Non-blocking: Hermes `cancel()` still does not mark the stream terminal, so `_retry_preview_mutation` can dispatch after cancel wins the backoff race. OpenClaw’s `ensureOpen()` is the missing counterpart.

PR #1559, opened 2026-08-26
Measure epoch divergence from the current epoch, not the high-water mark

dannym-arx:
Rule 6 (`epoch_divergence`) measured each engine's lag from its high-water epoch. A marmot that burrows backwards, restored from an older backup or a rolled-back database, then gets reported at an epoch it no longer holds.

Nothing in the protocol walks an epoch back, so a regression is always a local-storage event. It also understates lag exactly when the device is worst off.

I found this while diagnosing a real cohort. One device re-hydrated at epoch 8 having already reached 13, and read as lag-4 against a tip of 17 instead of lag-9. Understating lag can also push an engine under the `>= 2` threshold and hide it from the report entirely, which is what the new fixture pins: `Healthy` before, quarantined after.

## What changed

Lag now comes from the newest timed epoch observation. The two readings agree on any engine that only moves forward and diverge only on a regression, where the lag can only widen. This adds findings and masks none. Untimed epoch evidence cannot be ordered, so it keeps the high-water reading and classifies as before.

A third `BehindMode`, `rolled_back`, outranks `went_dark` and `active_while_behind`, which a regressed engine also satisfies by construction. It points at a different remedy: no redelivery repairs it, because the device is asking for commits the group already retired. Re-invite, not re-pull.

On the export that prompted this, one engine moved from `at epoch 13 (active while behind)` to `at epoch 8 (rolled back)`.

## Testing

The new fixture is synthetic and carries no real identifiers, matching the crate's fixture policy. I checked it against the old logic to confirm it fails there: the old code returns `Healthy`, the new code returns a `RolledBack` quarantine.

## Summary by CodeRabbit

- **New Features**
  - Added rollback detection when an engine reports a lower epoch.
  - Added a distinct “rolled back” status that takes precedence over lag indicators.
  - Updated remediation guidance: full-history replay may restore devices when history is retained; re-invitation is needed only when it is unavailable.

- **Bug Fixes**
  - Prevented false rollback detection when epoch observations lack timing information.
  - Preserved accurate high-water epoch calculations.

- **Tests**
  - Added coverage for rollback classification and untimed epoch scenarios.

### DIFF
--- a/crates/incident-replay/src/classify.rs
+++ b/crates/incident-replay/src/classify.rs
@@ -162,7 +162,9 @@ impl fmt::Display for HaltedEngine {
 pub struct BehindEngine {
     /// The engine that fell behind.
     pub engine_id: String,
-    /// The highest epoch the engine's own events place it at.
+    /// The epoch the engine's own events currently place it at — its newest
+    /// timed position, not its high-water mark, so a rolled-back device is
+    /// reported where it actually sits.
     pub epoch: u64,
     /// How the engine was behaving once the group moved past it.
     pub mode: BehindMode,
@@ -198,6 +200,15 @@ pub enum BehindMode {
     /// catching up: commits are not reaching it even though its other traffic
     /// flows.
     ActiveWhileBehind,
+    /// The engine's newest timed epoch is *below* one it already reported: its
+    /// local state moved backwards. Nothing in the protocol walks an epoch
+    /// back, so this is a local-storage event — a device restored from an
+    /// older backup, a rolled-back database — and no amount of redelivery
+    /// repairs it, because the device is asking for commits the group has
+    /// already retired. It outranks the other two modes: an engine that rolled
+    /// back is also, necessarily, active or dark, and the rollback is the
+    /// diagnosis an operator acts on (re-invite, not re-pull).
+    RolledBack,
 }
 
 impl fmt::Display for BehindMode {
@@ -205,6 +216,7 @@ impl fmt::Display for BehindMode {
         f.write_str(match self {
             BehindMode::WentDark => "went dark",
             BehindMode::ActiveWhileBehind => "active while behind",
+            BehindMode::RolledBack => "rolled back",
         })
     }
 }
@@ -448,6 +460,12 @@ struct EngineActivity {
     last_seen_ms: Option<u64>,
     /// The highest epoch the engine reported itself at.
     high_water_epoch: Option<u64>,
+    /// The engine's newest *timed* epoch observation, as `(wall_time_ms,
+    /// epoch)`. Held beside the high-water mark because the two disagree
+    /// exactly when local state moved backwards, which is the whole signal
+    /// behind [`BehindMode::RolledBack`] — a max alone reports a restored
+    /// device at the epoch it used to hold and understates its lag.
+    current: Option<(u64, u64)>,
 }
 
 /// The weaker of the two gates between "no contested branch" and "healthy":
@@ -478,6 +496,14 @@ fn epoch_divergence(export: &AgentStateExport) -> Option<QuarantineReason> {
         let observed = event.kind.observed_epoch();
         activity.high_water_epoch = activity.high_water_epoch.max(observed);
         if let (Some(epoch), Some(ms)) = (observed, event.wall_time_ms) {
+            // Newest timed observation wins. A tie orders nothing, so it keeps
+            // the higher epoch rather than assert a regression the timestamps
+            // do not actually establish.
+            let candidate = (ms, epoch);
+            activity.current = Some(match activity.current {
+                Some(current) if current > candidate => current,
+                _ => candidate,
+            });
             epoch_first_seen
                 .entry(epoch)
                 .and_modify(|first| *first = (*first).min(ms))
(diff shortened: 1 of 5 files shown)


dannym-arx:
Fixed in e6c62d10 — the mixed timed/untimed case was a real bug, thanks.

An engine with a timed epoch 5 and a newer untimed epoch 9 read as rolled back to 5, because `high_water_epoch` counted every observation while the current reading took only timed ones. Ordinary forward movement nobody stamped became a quarantine.

Rather than special-casing the mixed case, one untimed epoch row now forfeits the current reading for that whole engine, matching the rule already applied to an untimed halt row in rule 5: the untimed row may be the newest, so no ordering claim is available and the high-water reading stands. Regression test `an_untimed_epoch_newer_than_the_timed_one_is_not_a_rollback` covers it, and I confirmed it fails without the guard (reports `RolledBack` at epoch 5 instead of `Healthy`).

The same commit also corrects an unevidenced claim in the `RolledBack` docs: it said no redelivery could repair a rolled-back device. mdk#1548 healed a device from epoch 13 to 16 by full-history replay, so the rollback changes the size of the ask, not its possibility.

erskingardner (approved):
Recommend merge. The premise is sound and the implementation is surgical.

- **Root cause, not symptom:** measuring lag from the high-water mark structurally understates it exactly when a device is worst off, and can hide it under the ≥ 2 threshold entirely. Reading the newest timed epoch fixes both, and since a rollback only widens lag, it adds findings without masking any.
- **Fail-closed on ambiguity is right:** one untimed epoch row forfeiting the current reading for the whole engine mirrors the existing untimed-halt treatment in `unrecoverable_halt` and avoids inventing rollbacks from unordered evidence. The equal-timestamp tie-break keeping the higher epoch has the same property.
- **Precedence is correct:** `rolled_back` outranking `went_dark`/`active_while_behind` matches reality — a rolled-back engine satisfies those by construction, and the mode names the cause and remedy rather than the symptom.
- **Scope:** two struct fields, one enum variant, one branch, two fixtures + tests, and the AGENTS.md rule-6 update. No drive-by changes; fixtures are synthetic per crate policy. Verified the old logic would classify the rollback fixture `Healthy` (high-water 5 vs tip 6 < threshold), so it genuinely pins the fix.
- Timestamp-ordering assumption (`wall_time_ms` as orderable truth) matches what rule 5 already relies on, so nothing new is being trusted here.

Nice writeup, too — the 'changes the size of the ask rather than its possibility' framing in the remediation note is exactly right.

crates/incident-replay/src/classify.rs
```diff
@@ -162,7 +162,9 @@ impl fmt::Display for HaltedEngine {
 pub struct BehindEngine {
     /// The engine that fell behind.
     pub engine_id: String,
-    /// The highest epoch the engine's own events place it at.
+    /// The epoch the engine's own events currently place it at — its newest
+    /// timed position, not its high-water mark, so a rolled-back device is
+    /// reported where it actually sits.
     pub epoch: u64,
```
erskingardner, line 168: Non-blocking: this field doc now says `epoch` is always the newest *timed* position, but `epoch_divergence` falls back to `high_water_epoch` whenever the engine has any untimed epoch row (the fallback the second commit added). Qualifying the doc — e.g. "newest timed position when every epoch observation is timed; the high-water mark when any row is untimed" — keeps a caller from reading a fallback value as a timed one. (Same point as the earlier automated comment; confirmed it still applies to the doc at its current wording.)

erskingardner (approved):
Reviewed at head e6c62d10. Recommend merge.

The premise is solid: rule 6 measured lag from the high-water epoch, so a device restored from an older backup was reported at an epoch it no longer holds — understating lag exactly when the device is worst off, and in the motivating cohort it dropped one engine below the >= 2 threshold into a false Healthy. The fix (measure from the newest timed observation) is the smallest change that solves it: findings can only be added, never masked.

Checked the change end to end:

- Correctness: the current-epoch selection, tie-handling (same timestamp keeps the higher epoch rather than asserting a regression no timestamp ordering establishes), and rolled_back precedence over the mode it also satisfies all read correctly. The RolledBack mode outroanking dark/active matches the documented contract.
- The mixed timed/untimed false-rollback bug introduced in the first commit is properly fixed in the second: one untimed epoch row forfeits the current reading for the whole engine and keeps the high-water reading, the same fail-closed policy rule 5 applies to untimed halt rows. The regression fixture (timed 5 then untimed 9 -> Healthy) pins it and the author verified it fails without the guard.
- The group tip still comes from high-water marks, so one rolled-back engine does not drag the group baseline down — rollback detection stays per-engine.
- Tests and fixtures: three focused tests (quarantine-as-rolled-back, precedence over active-while-behind, untimed guard). Fixtures are synthetic/non-sensitive per the crate policy. AGENTS.md rule-6 text updated as this repo requires.
- Scope: five files, minimal; the remedy-text correction (mdk#1548 heals by replay, so re-invite only once retention lapses) removes an unevidenced claim rather than hedging it.

One non-blocking note posted inline: the BehindEngine.epoch doc claim. CI on the head SHA is all green. No executable validation was run locally — relied on the remote CI plus reading the rules in full at the head SHA.

Issue #1560, opened 2026-08-26
cgka-engine: deferred candidate replay causes extreme SQLite write amplification on iOS


erskingardner:
## Summary

On a physical iPhone, a White Noise staging installation using MarmotKit 0.9.15 entered a persistent `backend failure: database disk image is malformed` state after otherwise ordinary use with poor/intermittent connectivity and frequent foreground/background transitions.

iOS disk-write resource reports show extreme write amplification in both the main app and Notification Service Extension (NSE). Matching archived dSYMs symbolicate both reports to MDK deferred-peel candidate-branch replay and SQLite snapshot capture/rollback.

The excessive writes are directly evidenced. They are highly disproportionate to the account's modest number of groups/messages. The database corruption may be a downstream consequence of this write storm, interruption during heavy WAL activity, or storage pressure, but that causal link is not yet proven because the encrypted database has not been extracted.

## Environment

- Device: physical iPhone 16 Pro Max
- OS: iOS 26.5.2 (23F84)
- Client: White Noise staging 2026.8.25 (build 28)
- MarmotKit: `marmotkit-v0.9.15`
- MDK SHA: `f7159003eb76e00764c0320ba783a0f6dde7486f`
- iOS reported approximately 9.7 GB free at the time of the main-app resource report

The current MDK master inspected at `4e042ec1718a` has no changes relative to 0.9.15 in the implicated `cgka-engine` candidate replay, snapshot guard, or `storage-sqlite` snapshot paths.

No account IDs, group IDs, message contents, or device identifiers are included here.

### Main app

iOS resource report at 2026-08-26 07:34:35 +0100:

```text
Writes: 17.18 GB of file backed memory dirtied over 70,224 seconds
Average: 244.64 KB/s
iOS limit: 17.18 GB / 24 hours
Sampled process window: approximately 4m48s, frontmost
```

### Staging NSE, same build

iOS resource report at 2026-08-26 07:44:33 +0100:

```text
Writes: 1,073.74 MB of file backed memory dirtied over 8,836 seconds
Average: 121.52 KB/s
iOS NSE limit: 1,073.74 MB / 24 hours
```

### Earlier staging NSE build

An earlier report from staging 2026.8.22 (build 28) is even more striking:

```text
Writes: 1,073.74 MB of file backed memory dirtied over 82 seconds
Average: 13.08 MB/s
```

[...]

PR #1563, opened 2026-08-26
fix: scope replay rollback to canonical group state

erskingardner:
## Summary

- make every temporary replay, branch-probe, and historical-read guard capture canonical group/OpenMLS state instead of the full message ledger and outbound queue
- add canonical-state-only restore semantics so pre-upgrade full retained anchors remain usable without rewriting live input/work rows
- keep canonical apply and interrupted-apply recovery deliberately full, with semantic and encrypted process-kill regressions covering live message/queue survival

This addresses the urgent write-amplification portion of #1560. Candidate-context reuse, replay-count bounds, and observability remain separate follow-up work, so this PR intentionally does not close the issue.

## Test plan

- [x] New guard regression failed before the fix and passes after it
- [x] New legacy-full-anchor restore regression failed before the fix and passes after it
- [x] `cargo test -p storage-sqlite` (414 passed, 4 ignored)
- [x] `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks`
- [x] `just fast-ci`

Refs #1560

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved crash recovery and hydration handling when rewinding or replaying group state.
  * Preserved live messages and queued outbound work during recovery probes and rollback operations.
  * Added support for recovering deferred candidate-branch work after interruptions.

* **Documentation**
  * Clarified snapshot and replay behavior, including preservation of live work during temporary recovery operations.

* **Tests**
  * Expanded coverage for crash recovery, hydration, rollback behavior, retained anchors, and candidate branches.

### DIFF
--- a/crates/cgka-engine/src/engine.rs
+++ b/crates/cgka-engine/src/engine.rs
@@ -2188,7 +2188,7 @@ impl<S: StorageProvider> Engine<S> {
             hasher.update(record.id.as_slice());
             let digest = hasher.finalize();
             let probe_snapshot = format!("hydrate-selfremove-probe-{}", hex::encode(&digest[..8]));
-            let guard = crate::snapshot_guard::SnapshotRollbackGuard::create(
+            let guard = crate::snapshot_guard::SnapshotRollbackGuard::create_group_state(
                 &self.storage,
                 group_id.clone(),
                 probe_snapshot,
--- a/crates/cgka-engine/src/message_processor/ingest.rs
+++ b/crates/cgka-engine/src/message_processor/ingest.rs
@@ -2531,8 +2531,11 @@ impl<S: StorageProvider> Engine<S> {
             // (panic, early error, async cancel) so the live group
             // state never leaks past this scope as the past-snapshot
             // state.
-            let guard =
-                SnapshotRollbackGuard::create(&self.storage, group_id.clone(), restore_snapshot)?;
+            let guard = SnapshotRollbackGuard::create_group_state(
+                &self.storage,
+                group_id.clone(),
+                restore_snapshot,
+            )?;
             let (ctx, message_retention_seconds) =
                 match self.context_from_group_snapshot(group_id, &snapshot_name) {
                     Ok(Some(context)) => context,
@@ -2613,10 +2616,14 @@ impl<S: StorageProvider> Engine<S> {
             current_epoch.0,
             hex::encode(&hasher.finalize()[..8])
         );
-        let guard = SnapshotRollbackGuard::create(&self.storage, group_id.clone(), restore_name)?;
+        let guard = SnapshotRollbackGuard::create_group_state(
+            &self.storage,
+            group_id.clone(),
+            restore_name,
+        )?;
         let resolved = match self
             .storage
-            .rollback_group_to_snapshot(group_id, &snapshot_name)
+            .rollback_group_state_to_snapshot(group_id, &snapshot_name)
         {
             Ok(()) => {
                 let provider =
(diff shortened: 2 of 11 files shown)


erskingardner (commented):
**Review: approve — recommend merge.** Clean, surgical fix for the urgent write-amplification half of #1560. The scope split is right: temporary probes/rewinds get canonical-state-only capture *and* restore (including legacy full anchors via `RestoreScope::GroupState`), while canonical apply stays deliberately full with the reasoning documented at the call site. Full-scope restore behavior is byte-for-byte unchanged, the trait fallback story for non-distinguishing backends is documented honestly, and the regressions (guard unit test, legacy-full-anchor rollback test, candidate-branch process-kill case, hydration work-row assertions) target exactly the bug class. CI green at `0c091ee`.

Two non-blocking follow-ups, both on unchanged code:

1. `SnapshotRollbackGuard::commit`/`Drop` still call the full `rollback_group_to_snapshot`. Safe today only because state-scoped snapshots serialize `messages: None`/`queued_outbound: None`. Switching them to `rollback_group_state_to_snapshot` would make the guard's 'leave live input/work collections untouched' contract self-enforcing instead of dependent on the capture format.

2. `recover_interrupted_rewind_probe` restores via full `rollback_and_release_group_snapshot`. A rewind-probe snapshot left behind by the pre-fix binary is a full image, so first hydrate after upgrading still rewrites live message/queue rows to the pre-probe image — the same bug class this PR fixes, in a narrow crash-on-old-version + upgrade window. Scoped restore there would close it.

Understand deferring candidate-context reuse, replay bounds, and observability to follow-ups; not closing #1560 is stated clearly.

erskingardner (commented):
**Recommend merge** (GitHub blocks self-approval; this is your own PR).

Solid fix for the write-amplification half of #1560. The scope split is exactly right: every temporary replay/probe/historical-read guard now captures canonical group state, and the new `rollback_group_state_to_snapshot` keeps legacy full anchors readable without letting a probe rewrite live message/queue rows. Canonical apply staying deliberately full (with the explanatory comment) is correct, since apply error recovery needs the complete pre-apply image.

I audited all remaining full-scope uses and each is intentional:
- apply snapshot create + `recover_interrupted_apply_snapshot`/`rollback_and_release_group_snapshot` — complete pre-apply image needed
- the apply-path retained-anchor rewind plus `restore_live_message_and_queue_records` — full by design
- the guard's own commit/Drop still calls `rollback_group_to_snapshot`, but guard snapshots are always state-scoped (`messages`/`queued_outbound` are `None`), so full restore skips work rows — equivalent, and restore.rs documents this

Coverage is good: scoped-guard regression, legacy-full-anchor scoped restore regression, hydration-recovery assertions on live ledger/queue rows, and an encrypted crash-recovery case for the new candidate-branch rewind point. CI is green and the storage-format doc was updated.

One non-blocking thought: the guard could call the scoped restore internally so the 'never restore work rows' invariant would hold even if a future call site ever guarded a full snapshot. Today every site creates scoped snapshots, so this is purely cosmetic.

erskingardner:
Reviewed the two non-blocking rollback-scope suggestions at current head.

- Addressed the live guard suggestion in `c107586f`: `SnapshotRollbackGuard::commit` and `Drop` now call `rollback_group_state_to_snapshot` directly, so the guard enforces its canonical-state-only contract at the trait boundary.
- Kept `recover_interrupted_rewind_probe` full-scope intentionally. A stranded probe may come from an older binary that created a full live snapshot and then rewound through a legacy full retained anchor. In that upgrade case the probe snapshot is the only durable copy of newer message/queue rows; state-only recovery would leave the historical work image in place. I documented this at the recovery call and strengthened `hydration_recovers_interrupted_candidate_branch_probe` to simulate that exact old-full-probe + old-full-anchor sequence.

Validation:
- focused guard regression passed
- strengthened hydration recovery regression passed
- `just fast-ci` passed

PR #1564, opened 2026-08-26
perf(cgka-engine): reuse deferred peel candidates

erskingardner:
## Summary

- reuse memory-only candidate-branch peel contexts only for the same group, exact deferred-peel fingerprint, and exact durable generation marker
- invalidate exporter-bearing contexts at fingerprint, generation, canonical-state, hydration, repair, quarantine, removal, disband, pending-publish, and rollback boundaries
- add privacy-safe aggregate sweep, enumeration, context count/depth, replay-probe, cache hit/miss/invalidation, and elapsed-time observability
- add deterministic 192-row and 256-row contested-graph regressions covering unchanged slices, fingerprint rollover, restart, cancellation, and the existing foreground/background work limits

## Before / after

For a 192-row contested backlog processed in three 64-row background slices under one unchanged fingerprint:

| Measurement | Before cache | After cache |
| --- | ---: | ---: |
| Candidate enumerations | 3 | 1 |
| Candidate context instances | 6 | 2 |
| Candidate replay probes | 12 | 4 |

The final regression asserts the four replay probes are independent of slice count for one fingerprint. In this graph each consumed replay probe owns one state-scoped replay snapshot, so the candidate replay snapshot work is bounded at four as well.

## Stacking and overlap

- #1563 merged to `master` as 18b63645bcda9eef6263b884687a2af72944278e
- rebased onto current `master`; `git range-diff` reports the phase-2 patch as equivalent
- the narrow overlap in `openmls_projection.rs` only exposes aggregate replay-probe count and context depth
- no changes to `snapshot_guard.rs`, storage snapshot encoding, rollback recovery, crash-recovery tests, snapshot-scope APIs, migrations, bindings, protocol behavior, or versions

## Test plan

- [x] TDD baseline reproduced the failure: expected one enumeration, observed three
- [x] `cargo test -p cgka-engine --features test-policy-overrides,test-crash-hooks`
- [x] `just fast-ci`

Refs #1560

## Summary by CodeRabbit

* **Performance**
  * Improved deferred processing by reusing candidate results when group state is unchanged.
  * Added safeguards to refresh cached results after updates, publishing, membership changes, recovery, and restarts.

* **Reliability**
  * Improved handling of replay limits, interruptions, missing anchors, and candidate evaluation failures.
  * Preserved pending work more consistently during cancellations and retries.

* **Monitoring**
  * Added detailed metrics for deferred processing, cache usage, replay activity, processing depth, and timing.

### DIFF
--- a/crates/cgka-engine/src/message_processor/mod.rs
+++ b/crates/cgka-engine/src/message_processor/mod.rs
@@ -35,6 +35,7 @@ use cgka_traits::transport::{TransportEnvelope, TransportMessage};
 use cgka_traits::types::{EpochId, GroupId, MemberId, MessageId};
 use sha2::{Digest, Sha256};
 use std::collections::BTreeSet;
+use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 pub(crate) const MAX_CONVERGENCE_REPROCESSING_PASSES: usize = 16;
@@ -166,6 +167,21 @@ pub(crate) struct DeferredPeelGroupState {
     /// per rejected message; this suppresses the repeats until the backlog
     /// drops back below the cap and re-arms.
     cap_rejection_audited: bool,
+    /// Candidate branch contexts for one exact process-local generation.
+    ///
+    /// These contexts contain exporter-derived secret material. The cache is
+    /// deliberately nested in engine memory: it is never serialized, audited,
+    /// logged, or copied into durable generation state, and process restart
+    /// drops it naturally.
+    candidate_cache: Option<DeferredPeelCandidateCacheEntry>,
+}
+
+struct DeferredPeelCandidateCacheEntry {
+    context_fingerprint: [u8; 32],
+    /// Exact durable barrier generation observed after enumeration. `None` is
+    /// also meaningful and must match on reuse.
+    durable_generation_fingerprint: Option<[u8; 32]>,
+    peel: Arc<crate::openmls_projection::CandidateBranchPeel>,
 }
 
 impl DeferredPeelGroupState {
@@ -1394,6 +1410,7 @@ impl<S: StorageProvider> Engine<S> {
         execution: &mut DeferredPeelExecution<'_>,
     ) -> Result<DeferredPeelWorkResult, EngineError> {
         let sweep_started = Instant::now();
+        self.engine_metrics.note_deferred_peel_sweep();
         let foreground_budget_ms = match execution {
             DeferredPeelExecution::Foreground(budget) => Some(budget.budget_ms),
             DeferredPeelExecution::Background => None,
@@ -1548,6 +1565,7 @@ impl<S: StorageProvider> Engine<S> {
             state.counted = true;
         }
         if total == 0 {
+            self.invalidate_deferred_peel_candidate_cache(group_id);
             if self.storage.deferred_peel_generation(group_id)?.is_some() {
                 self.storage.delete_deferred_peel_generation(group_id)?;
                 self.converge_stored_openmls_messages_with_time(group_id, now)
@@ -1577,6 +1595,7 @@ impl<S: StorageProvider> Engine<S> {
             .cloned()
             .collect::<Vec<_>>();
         if unattempted.is_empty() {
+            self.invalidate_deferred_peel_candidate_cache(group_id);
             if self.storage.deferred_peel_generation(group_id)?.is_some() {
                 self.storage.delete_deferred_peel_generation(group_id)?;
                 self.converge_stored_openmls_messages_with_time(group_id, now)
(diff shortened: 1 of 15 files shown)


erskingardner (commented):
Review of head commit c7dc32c only (the stacked #1563 commits have since merged - this needs a rebase/retarget onto master before merge, as the PR body already notes).

The approach is sound: the cache key (exact deferred-peel context fingerprint + exact durable generation marker) is conservative, the fingerprint already covers live epoch, retained anchors, and stored commit digests, and the explicit invalidation call sites cover the cases the fingerprint cannot catch (same-epoch projection replacement via hydration/repair/quarantine/removal, plus every begin_pending/set_stable transition). I cross-checked every set_stable/begin_pending site in the crate against the invalidation calls and found full coverage. Keeping the exporter-bearing contexts in a memory-only, non-Clone, never-serialized entry that restart drops naturally is the right secrecy posture, and the per-row Arc::ptr_eq re-check correctly stops offering a stale slice-local Arc after a row itself advanced canonical state.

The observability additions are aggregate-only and match the privacy invariants (no group/branch/fingerprint identities logged). The 192/256-row regressions pin the work shape well, including restart and cancellation.

Non-blocking nits inline. CI is green. Recommend merge after the rebase.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -1631,37 +1650,129 @@ impl<S: StorageProvider> Engine<S> {
...
+                Err(error) => {
+                    self.engine_metrics
+                        .note_deferred_peel_candidate_enumeration(
+                            candidate_enumeration_started
+                                .elapsed()
+                                .as_millis()
+                                .try_into()
+                                .unwrap_or(u64::MAX),
+                            std::iter::empty(),
+                            0,
```
erskingardner: Nit: on the error path the enumeration may already have consumed replay probes (the budget is spent inside `candidate_branch_peel` before it returns `Err`), so recording `0` here under-counts `deferred_peel_candidate_replay_probes` exactly when the failed enumeration was most expensive. Not blocking — the sweep returns immediately and the miss is already recorded — but if probe bounding matters for diagnosing #1560-class write storms, consider threading the consumed count out on the error path too.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -1631,37 +1650,129 @@ impl<S: StorageProvider> Engine<S> {
...
+        let durable_generation_fingerprint = self
+            .storage
+            .deferred_peel_generation(group_id)?
+            .map(|generation| generation.context_fingerprint);
+        let cached = self
+            .deferred_peel
+            .get(group_id)
+            .and_then(|state| state.candidate_cache.as_ref())
+            .filter(|cached| {
+                cached.context_fingerprint == fingerprint
```
erskingardner, line 1672: The reuse filter hashes live epoch, anchor names, and commit digests — but not pending proposals, which do enter `candidate_branch_peel_contexts_from_current` via `pending_proposal_messages` and get replayed into each capture path. I believe reuse is still safe: staged proposals don't alter the group context or per-epoch exporter secrets, so `capture_candidate_tip_context` yields the same `GroupContextSnapshot` with or without them. But that safety rests on an implicit MLS invariant (proposal staging is context-invariant) rather than anything in this cache key. Worth one doc line on `DeferredPeelCandidateCacheEntry` stating that assumption, so a future change to `build_group_context_snapshot` (or to how proposals join replay paths) knows this cache depends on it.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -1631,37 +1650,129 @@ impl<S: StorageProvider> Engine<S> {
...
         for record in unattempted.into_iter().take(row_limit) {
             if execution.exhausted() {
                 timed_out = true;
                 break;
             }
+            // A row can itself advance or replace canonical state. Those
+            // transition sites synchronously invalidate the map entry; do not
+            // keep offering the slice's local Arc to later rows after that.
+            // Their lifecycle remains untouched for the next fingerprint.
+            if !self
```
erskingardner, line 1777: This `Arc::ptr_eq` backstop is a good safety net, but the primary defense is now ~15 `invalidate_deferred_peel_candidate_cache` calls scattered across 10 files, and every future state-transition site has to remember to add one. The natural chokepoints already exist — `epoch_manager.begin_pending` / `set_stable`, hydration, quarantine — so a follow-up that bumps a monotonic per-group projection-generation counter at those few sites and folds it into the cache key would replace most of the scattered calls and make the invariant structural rather than distributive. Not blocking: the fingerprint check plus this ptr_eq guard bound the blast radius of a missed site, and the regressions pin today's behavior.

erskingardner (commented):
Reviewed against #1560 (phase 2 of the write-amplification fix, on top of #1563's state-scoped snapshot guards). The premise is sound: #1560 explicitly flags repeated candidate enumeration across convergence wakes as suspected mechanism #2/#3, and the before/after numbers plus the slice-count-independent replay-probe assertion directly answer it.

Correctness of the reuse gate looks well-constructed:

- The fingerprint covers live epoch, retained anchor names, and all stored commit digests, so any durable graph change forces a miss.
- The durable-generation marker closes the "fingerprint equal but projection replaced" window together with the explicit hydration/repair/quarantine invalidations.
- The `Arc::ptr_eq` re-check inside the row loop is a good belt-and-suspenders backstop for mid-slice transitions, and breaking (rather than continuing) keeps untried rows eligible.
- Secrets stay memory-only, never serialized or logged; restart naturally drops them — the right posture for exporter-bearing material.
- Metrics respect the observability invariant (aggregate counts, explicit target/method, no identifiers).

Two non-blocking notes inline. CI is green including the 4-shard test matrix, and the new 192/256-row regressions cover the unchanged-fingerprint, rollover, restart, and cancellation cases deterministically. No version bumps, no transport code leaking into the engine. Verdict: would merge as-is; treat the inline notes as optional polish or follow-ups.

crates/cgka-engine/src/message_processor/mod.rs
erskingardner: Fixed in b8a7ec55. candidate_branch_peel now returns a counted internal failure that preserves the replay budget consumed before an error; the sweep records that value instead of zero. I also added candidate_failure_preserves_consumed_replay_count, and the full cgka-engine feature test suite passes.

erskingardner:
Review feedback update for b8a7ec55: successful group-policy changes now invalidate the cached enumeration and have a regression; the phase-2 contested-graph fixtures install a distinct Nostr routing component and no longer substitute the MLS group id; enumeration failures preserve and report replay probes already consumed; and the pending-proposal context-invariance assumption is documented at the cache entry. The crash-recovery fixture and storage-documentation observations are inherited from merged PR #1563 and are not part of the #1564 diff, so I left those base-branch areas untouched. The suggested structural projection-generation counter remains a worthwhile nonblocking follow-up; this PR retains the reviewed exact fingerprint plus durable-generation key and explicit transition invalidations.

crates/cgka-engine/tests/deferred_peel_lifecycle.rs
```diff
@@ -493,6 +604,363 @@ async fn carol_behind_two_epochs_with<P>(
...
+    let (group_id, create) = incumbent
+        .create_group(CreateGroupRequest {
+            name: "candidate-cache".into(),
+            description: String::new(),
+            members: vec![rival_kp],
+            required_features: vec![],
+            app_components: vec![AppComponentData {
+                component_id: NOSTR_ROUTING_COMPONENT_ID,
+                data: encode_nostr_routing_v1(&routing).unwrap(),
+            }],
```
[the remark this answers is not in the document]
erskingardner: Fixed in 1810dcc9 by reusing the existing wss://relay.example routing-test fixture instead of adding a new endpoint literal. All 19 deferred-peel lifecycle tests and just fast-ci pass.

PR #1565, opened 2026-08-26
Fix/refused ingest stays fetchable

jgmontoya:
## Problem

The engine deliberately keeps a capacity-refused ingest fetchable: it drops the object unpersisted and suppresses its own seen-cache insertion so transport redelivery can recover it. The app layer defeats that one level up — both sync seams mark every Ok outcome seen (under a comment saying "only after durable ingest succeeds"), and the transport cursor advances before the outcome is inspected. A refused message is therefore marked already-seen and pushed below the relay since-floor: permanently unfetchable, and seen_events is persisted with no removal path. Field: in group ae192586 the epoch-15 commit was refused three seconds after the deferred cap filled — inside the recovery drain armed to fetch it — then marked seen; three devices are live-wedged this way (one run shows the relays serving 861 events and the app discarding all of them), while the group ran to epoch 20.

Second, compounding: a backfill that recovers nothing still records success and calls mark_replayed(), silently latching every tracked group's detector for the process lifetime.

## Fix

ingest_delivery now reports whether the engine left the object unpersisted (one rule, one free function, both seams): on ResourceRefused, neither remember_seen_event nor remember_transport_cursor runs, so the relays' natural redelivery heals the device — verified end to end by tests that fill the real 256-row cap and prove the same event is re-presented and re-classified rather than deduped. TransportDeferred (durably retained) still marks seen, pinned by test. Whether deferred ingests should advance the cursor stays with the tracked since-floor design item. One asymmetric case deliberately not exempted: Ignored{UnknownGroup} is non-durable on one engine path but durable on the other and indistinguishable from the app side — un-seeing it risks an unbounded cursor stall; tracked separately.

mark_replayed() is now gated on the replay having recovered something: deliveries ingested, or a tracked group's epoch moving across the run. A fruitless EOSE-confirmed drain still records its honest completed row and consumes its intent, but no longer disarms groups it never recovered. Epoch movement and nonzero deliveries each still disarm — both boundaries pinned, since the field's 121-second recovery drain moved its epoch one second after its terminal row was written.

## Tests

Six, fail-first on the defects: refused deliveries leave the seen index and cursor untouched at both seams and stay fetchable through redelivery; deferred deliveries still mark seen; a recovered-nothing backfill no longer disarms a bystander's arm; epoch-moved and deliveries-ingested runs still disarm. 952 marmot-app tests and the serial relay_runtime suite pass; engine, traits, and storage untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refused deliveries can now be redelivered instead of being incorrectly marked as processed.
  * Recovery remains active after an empty or fully refused replay and only completes when durable progress is made.
  * Consistent behavior is now applied to live receiving and catch-up synchronization.

* **Improvements**
  * Audit records now distinguish refused deliveries from successfully retained deliveries while remaining compatible with older records.

* **Tests**
  * Added coverage for retention, replay behavior, recovery completion, and audit reporting.

### DIFF
--- a/crates/marmot-app/src/client/sync.rs
+++ b/crates/marmot-app/src/client/sync.rs
@@ -138,6 +138,40 @@ impl DrainVerdict {
     }
 }
 
+/// What one delivery's ingest settled, for the two seams that decide whether the
+/// delivery may be dropped from the fetch path.
+struct DeliveryIngest {
+    /// Membership-changing effects landed, so the app projection and the route
+    /// table owe a save before anything else can fail.
+    routes_dirty: bool,
+    /// The engine kept no durable trace of this object, so transport
+    /// redelivery is the only path back to it. See
+    /// [`ingest_left_object_unpersisted`].
+    must_stay_fetchable: bool,
+}
+
+/// Whether the engine left this outcome's transport object unpersisted, so only
+/// relay redelivery can present it again.
+///
+/// This is the app-side half of the engine's `retryable_unpersisted_ingest_id`
+/// contract. `ResourceRefused` is the one `Ok` outcome the engine drops without
+/// a durable record *and* deliberately keeps out of its own seen cache, and the
+/// trait states the consequence directly: a resource refusal "must not make
+/// same-id redelivery a duplicate". Recording it in `seen_events` (which has no
+/// removal site short of ring overflow) or letting it carry the relay `since`
+/// floor past itself makes the object permanently unfetchable — across restarts
+/// and across `repair_full_history` alike.
+///
+/// Every other `Ok` outcome leaves the engine holding durable evidence of the
+/// object: `TransportDeferred` retains the raw transport row for later peel,
+/// `Buffered` retains it for convergence replay, and the terminal
+/// classifications (`Processed`, `Ignored`, `LocalState`, `Stale`, `Rejected`)
+/// are decisions the engine can reach again from storage. Forgetting those is
+/// correct dedupe, so they stay.
+fn ingest_left_object_unpersisted(outcome: &IngestOutcome) -> bool {
+    matches!(outcome, IngestOutcome::ResourceRefused { .. })
+}
+
 /// What one drain loop saw on the wire.
 ///
 /// `deliveries` counts receives the drain ingested; `skipped` counts those it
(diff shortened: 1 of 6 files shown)


crates/marmot-app/src/client/sync.rs
```diff
@@ -138,6 +138,40 @@ impl DrainVerdict {
...
+/// and across `repair_full_history` alike.
+///
+/// Every other `Ok` outcome leaves the engine holding durable evidence of the
+/// object: `TransportDeferred` retains the raw transport row for later peel,
+/// `Buffered` retains it for convergence replay, and the terminal
+/// classifications (`Processed`, `Ignored`, `LocalState`, `Stale`, `Rejected`)
+/// are decisions the engine can reach again from storage. Forgetting those is
+/// correct dedupe, so they stay.
+fn ingest_left_object_unpersisted(outcome: &IngestOutcome) -> bool {
+    matches!(outcome, IngestOutcome::ResourceRefused { .. })
```
Datawav, line 172: The app-side classifier does not match the engine's actual `retryable_unpersisted_ingest_id` contract. `cgka-engine/src/message_processor/ingest.rs:390-408` also sets that marker and returns `Ignored { UnknownGroup }` when no group row exists, specifically so commits that raced ahead of a Welcome can process on same-id redelivery. This helper returns false for that outcome, so both app seams persist the id in `seen_events` and advance the transport cursor even though the engine retained no row. After the Welcome, same-id relay delivery is skipped by the app across restart and full-history repair. Expose the engine's unpersisted/retryable disposition directly (or use a dedicated outcome) rather than reconstructing it from `ResourceRefused`, and add receive- and catch-up-seam tests for commit-before-Welcome redelivery.

crates/marmot-app/src/client/sync.rs
```diff
@@ -1780,12 +1872,56 @@ impl AppClient {
             },
         );
         if succeeded {
-            self.epoch_stall.mark_replayed();
+            if Self::replay_recovered_something(&execution.epochs_before, &epochs_after, counts) {
```
Datawav, line 1875: Skipping `mark_replayed()` does not re-arm the group that caused this replay. `observe_resource_refusal()` already set that group's `fired_at_epoch` when it armed the intent; this successful terminal path consumes the pending intent, but neither this branch nor `replay_recovered_something()` clears that latch. A later refusal at the same stalled epoch therefore returns `Skip`, so an all-refused replay still leaves the original group unable to schedule another automatic repair. The new all-refused test misses this because it calls `apply_backfill_decision(Arm, ...)` directly instead of first calling `observe_resource_refusal()`, so the armed group never enters the detector. Add a production-shaped test that arms through the detector, completes a 1-delivery/1-refused replay, and requires the next same-epoch refusal to arm again; then explicitly reset/re-arm the affected groups when a completed replay retained none of their needed history.

Datawav (commented):
## Retrospective review: two blocking recovery gaps remain

The pull request merged while this review was in progress, so this is a formal COMMENT review rather than a change request. I reviewed the final PR head `aadaaa5438d39934facb5fdda5f073828d7cd817`, the ingest/seen-cache/cursor/backfill/replay/persistence paths, the targeted tests, and the successful exact-head CI run.

Two correctness gaps remain: the app-side unpersisted classifier omits the engine's retryable unknown-group outcome, and a completed all-refused replay does not actually re-arm the originally stalled group because its detector latch is never cleared. The latter test passes only because it manually arms the pending intent without first arming that group in the detector.

These paths can still make a commit permanently unfetchable or leave the affected group unable to schedule another automatic full-history replay. Please address both in a follow-up before treating the field wedge as fully repaired.

PR #1566, opened 2026-08-26
fix(marmot-app): bound epoch backfill worker drains

### LINKED ISSUE #1562: marmot-app: EOSE backfill can monopolize the account worker forever on duplicate traffic
## Summary

The epoch-gap backfill drain can hold the serial account worker indefinitely when a relay continuously streams duplicate/non-novel deliveries but never sends end-of-stored-events (EOSE).

`drain_sdk_relay` resets its 30-second silence budget on **every** delivery before checking whether the event is an own echo or already seen. Therefore a steady duplicate stream prevents the receive timeout and silence budget from firing. The hot-path EOSE poll only completes the drain if EOSE has actually arrived. With no EOSE, the loop has no execution quantum or no-progress bound and exits only when traffic stops or the worker is aborted at shutdown.

This is a poor/intermittent-connection reliability problem: reconnect replay, relay duplication, or a misbehaving relay can monopolize the one account worker, delaying local commands and all other account work even though the recovery makes no novel progress.

## Evidence

- `crates/marmot-app/src/client/sync.rs:1042-1063` documents the residual explicitly: a relay streaming non-novel events forever without EOSE holds the drain until traffic stops or shutdown.
- `crates/marmot-app/src/client/sync.rs:1116-1123` defines the budget as consecutive silence rather than total/no-progress time.
- `crates/marmot-app/src/client/sync.rs:1168-1179` resets `silence_started` for every delivery before duplicate/echo rejection.
- PR #1548 correctly stopped timeout-only epoch backfill from being reported as success and added EOSE-aware completion. Its review also identified continuous duplicate traffic as the remaining unbounded-runtime risk; no separate current issue tracks it.

This is distinct from #1546: waking reconnect backoff sooner does not bound a recovery drain after it starts.

## Required behavior

Give backfill a bounded account-worker execution quantum and/or a bounded no-novel-progress budget. Reaching that bound must **not** be reported as successful replay: preserve the partial prefix, leave/re-arm the durable recovery intent, classify the result as unconfirmed/incomplete, and retry with the existing pacing/backoff machinery. A long replay that continues to produce novel events must still make progress across multiple quanta rather than being discarded or falsely completed.

Keep metrics privacy-safe and aggregate-only.

## Acceptance criteria

- A deterministic adapter test streams duplicates/own echoes continuously without EOSE; the backfill returns an incomplete verdict within a documented bound.
- The worker services a queued local command within a documented bound while that recovery remains pending.
- The bounded exit does not call `mark_replayed`, does not classify the replay as EOSE-confirmed or quiescence-complete, and retains/re-arms durable recovery across restart.
- Automatic retry remains paced/backed off so a hostile stream cannot create a hot retry loop; explicit user catch-up retains its intended override semantics.
- A large replay containing continuing novel events completes across bounded quanta without losing the processed prefix or starving commands.
- EOSE still ends the drain promptly when it arrives.
- Shutdown remains responsive during duplicate traffic.
- Audit/telemetry distinguishes EOSE completion, novel-progress quantum yield, duplicate/no-progress yield, and terminal failure without recording identifiers or payloads.

## Related

- #1546 — interrupt reconnect backoff on restored connectivity.
- #1560 — expensive convergence work is another account-worker latency amplifier, but does not cause this unbounded loop.

erskingardner:
## Summary

- bound each epoch-gap relay drain to a five-second safe-boundary account-worker quantum and distinguish productive from duplicate-only yields in privacy-safe audit rows
- checkpoint novel prefixes across quanta, preserve EOSE and quiescence-fallback completion semantics, and pace only attempts that make no durable progress
- persist per-group recovery intent before replay and restore it after restart until the exact confirmed epoch is consumed
- cover continuous duplicate traffic, queued local command service, productive multi-quantum replay, EOSE completion, and durable re-arm/CAS behavior

Fixes #1562

## Test plan

- [x] New duplicate-only regression timed out before the fix and passes afterward
- [x] `cargo test -p marmot-app --features test-policy-overrides`
- [x] `cargo test -p marmot-app --features test-policy-overrides --lib epoch_backfill_ -- --nocapture`
- [x] Production-policy EOSE, no-EOSE, and fallback regressions
- [x] `cargo test -p storage-sqlite`
- [x] `just fast-ci`

## Summary by CodeRabbit

* **New Features**
  * Epoch-gap recovery now runs in bounded execution windows, keeping the application responsive during large backfills.
  * Recovery progress and durable recovery intents are checkpointed and resumed across restarts.
  * Added configurable development/test backfill execution windows.
  * App-message queries can now filter results by inner event kind.

* **Bug Fixes**
  * Improved retry pacing when replay makes no durable progress.
  * Correctly distinguishes refused resources from unpersisted deliveries and re-arms eligible recovery groups.
  * Prevented newer recovery requests from being overwritten or cleared incorrectly.

* **Tests**
  * Expanded coverage for recovery, filtering, retries, and delivery handling.


crates/marmot-app/src/lib.rs
```diff
@@ -307,6 +307,17 @@ const SESSION_DB_FILE: &str = "session.sqlite";
...
+/// account worker before it checkpoints its prefix and yields incomplete.
+///
+/// The bound is observed between deliveries, where no engine snapshot guard or
+/// account-state transaction is live. One already-claimed delivery and the
+/// final prefix checkpoint may therefore extend wall-clock time past the
+/// quantum, but the relay receive loop itself cannot monopolize the worker.
+/// Five seconds leaves headroom inside [`APP_RUNTIME_LOCAL_WORKER_RESPONSE_WAIT`]
+/// for that boundary work while still amortizing relay subscription setup over
+/// useful replay progress.
+pub(crate) const EPOCH_BACKFILL_EXECUTION_QUANTUM: Duration = Duration::from_secs(5);
```
erskingardner, line 320: Non-blocking question about the production interplay: with a 5 s quantum always firing before the 30 s silence budget, the `silence_started.elapsed() >= silence_budget` exit in `drain_sdk_relay` is unreachable in production (only dev overrides or the adapter-closed `Ok(Ok(None))` path can reach an `EoseTimeout` verdict). That means a relay whose EOSE consistently arrives more than ~5 s after its last delivery can never EOSE-complete: each attempt yields at the quantum, retries spend the unconfirmed budget, and recovery lands on the `QuiescenceFallback` weaker claim after 3 attempts. The escape hatch exists and the pacing prevents a hot loop, so this is defensible — but it's a real semantic change to `EPOCH_BACKFILL_EOSE_WAIT` (per-attempt EOSE patience is now effectively the quantum, not 30 s). Could the `EPOCH_BACKFILL_EOSE_WAIT` doc say that explicitly so a future reader doesn't assume a 30 s EOSE grace still exists in production?

crates/marmot-app/src/client/sync.rs
```diff
@@ -1167,29 +1224,45 @@ impl AppClient {
...
+            if completion
+                .execution_quantum()
+                .is_some_and(|quantum| drain_started.elapsed() >= quantum)
+            {
+                if matches!(completion, DrainCompletion::EndOfStoredEvents { .. })
+                    && self.backfill_drain_verdict().await == DrainVerdict::Complete
+                {
+                    break DrainVerdict::Complete;
+                }
+                break DrainVerdict::quantum_yield(*counts);
```
erskingardner: Good placement: checking the quantum at the top of the loop means skipped deliveries (the duplicate/echo `continue` path) hit it on every iteration, which is exactly the unbounded case from the issue. Polling the EOSE gate once more before yielding is a nice touch — a boundary EOSE still completes rather than costing a retry cycle.

crates/marmot-app/src/client/sync.rs
```diff
@@ -2063,8 +2263,13 @@ impl AppClient {
                         eose_attempt_limit = EPOCH_BACKFILL_EOSE_ATTEMPT_LIMIT,
                         "epoch-gap backfill drain ended without the relays confirming stored history; retrying later"
                     );
-                    self.epoch_backfill_retry_not_before =
-                        Some(Instant::now() + self.epoch_backfill_retry_backoff(retry_ordinal));
+                    self.epoch_backfill_retry_not_before = if verdict.made_novel_progress() {
```
erskingardner, line 2314: Two observations on the productive-yield exemption, neither blocking:

1. Setting `retry_not_before = None` also clears any backoff accrued by earlier no-progress attempts on the same intent. Given a productive yield proves the relay is serving history, that seems right, just noting the reset is broader than "don't pace this yield".
2. With no pacing, the receive seam re-runs the intent after every inbound ingest, so a productive-but-never-EOSE relay keeps the worker in near-continuous 5 s slices with only boundary gaps. Local commands stay bounded (one quantum + boundary work, within the 10 s worker wait), which matches the issue's intent, but it does mean sustained ~5 s command latency for the whole replay rather than once per backoff interval. The field-export shape (59 min / 63 events) would now run as ~700 resubscribe cycles instead of one long drain — worth confirming the per-quantum resubscription cost is acceptable at that scale.

crates/marmot-app/src/client/sync.rs
```diff
@@ -2044,6 +2225,25 @@ impl AppClient {
                 // disarm the detector, so it is recorded as a failed attempt
                 // and its intent stays queued for the next seam.
                 let error_kind = verdict.error_kind();
+                if verdict.spends_unconfirmed_attempt() {
+                    execution.pending.unconfirmed_attempts =
+                        execution.pending.unconfirmed_attempts.saturating_add(1);
+                }
+                if error_kind.is_none()
+                    && let Err(error) = self.clear_epoch_backfill_intent(&execution.pending)
```
erskingardner, line 2281: Verified the clear-failure path fails safe: `finish_epoch_backfill_execution(..., false)` requeues the in-memory intent and the durable row survives (clear failed), so the next seam replays idempotently against duplicates. Correct direction for the torn-write-free invariant; the `Err` return is a bit harsh for a replay that actually completed, but it's consistent with the persist-failure path above and rare enough not to matter.

crates/marmot-app/src/client/sync.rs
```diff
@@ -1651,6 +1752,69 @@ impl AppClient {
...
+                    stalled_epoch: intent.stalled_epoch,
+                },
+            );
+        }
+        if malformed > 0 {
+            tracing::warn!(
+                target: "marmot_app::epoch_stall",
+                method = "restore_persisted_epoch_backfill_intents",
+                malformed,
+                "ignored malformed durable epoch-gap recovery markers"
```
erskingardner, line 1806: Nit: malformed rows are counted and warned but never removed, so an account with one corrupted row re-warns on every open, forever. Fine to leave (the count is capped by the table size and this needs actual corruption), but deleting the offending rows here would make the signal self-healing.

erskingardner (commented):
Solid fix for #1562. The quantum is enforced at every loop boundary (including after skipped deliveries, the exact path the issue describes), productive yields checkpoint their prefix and don't spend the unconfirmed ordinal, the durable intent is written before the external subscription side effect and cleared with exact-epoch CAS semantics, and `mark_replayed` is unreachable from a bounded exit. Audit rows stay aggregate-only. All of the issue's acceptance criteria are covered by the new tests, and CI is green.

This looks merge-ready to me; the notes below are non-blocking questions/nits. The one worth a reply is the slow-EOSE tradeoff: in production the 5 s quantum always fires before the 30 s silence budget, so EOSE completion now requires EOSE within ~one quantum, and slower relays degrade to resubscribe cycles and eventually the quiescence fallback claim.

crates/storage-sqlite/src/migrations/0052_epoch_backfill_intents.rs
```diff
@@ -0,0 +1,17 @@
+use crate::SqliteResultExt;
+use cgka_traits::storage::StorageResult;
+use rusqlite::Transaction;
+
+pub(crate) fn apply(tx: &Transaction<'_>) -> StorageResult<()> {
+    tx.execute_batch(
+        r#"
+CREATE TABLE app_epoch_backfill_intents (
+    group_id_hex   TEXT PRIMARY KEY
+                   REFERENCES account_groups(group_id_hex) ON DELETE CASCADE,
```
erskingardner: Non-blocking: this cascade is worth pinning in `group_owned_tables_have_cascading_foreign_keys` (migrations.rs). That test enumerates the `account_groups`-child tables explicitly, and this table is exactly the kind that must not outlive its group — if the cascade were ever dropped, deleting a group's local data would leave a durable recovery marker behind that re-arms on every account open and then defers forever on `GroupEpochUnavailable`. One assertion (`app_epoch_backfill_intents` / `group_id_hex` → `("account_groups", "CASCADE")`) keeps it covered.

erskingardner (commented):
Review verdict: recommend merge (posted as comment — can't approve own PR). This is a tight, well-verified fix for #1562.

**Correctness checks I ran through:**

- The quantum is enforced at every safe boundary: the loop-top check catches the continuous-delivery case the issue describes (the receive timeout can never fire while duplicates outpace `SDK_DRAIN_WAIT`), the wait is clamped to the remaining quantum so the final timeout lands on the boundary, and both receive-timeout arms re-check before classifying. No engine snapshot guard or state transaction is live at any of those points, matching the doc on `EPOCH_BACKFILL_EXECUTION_QUANTUM`.
- Productive vs. duplicate-only yields split correctly on `durable_deliveries()` (all-refused streams count as no progress, refused objects stay fetchable), only no-progress verdicts spend `unconfirmed_attempts`, and productive yields clear the retry cooldown — so a genuinely advancing replay runs back-to-back quanta while a hostile duplicate stream converges to one 5s drain per capped-backoff interval. `mark_replayed` is structurally unreachable from a bounded exit.
- The durable intent lifecycle follows the repo's torn-write rules: armed before the external subscription side effect (with a retry at replay time on the warn path), cleared with exact-epoch CAS so a concurrently re-armed newer epoch survives, and a clear failure converts the run to `Err` rather than confirming. Same-epoch re-arms lost between a completion clear and the next pre-replay persist are harmless — the replay they wanted just ran.
- EOSE and quiescence-fallback semantics are preserved; in production the silence budget now only fires via the transport-closed path, which the doc's "normally yields first" phrasing covers.
- 5s quantum vs. 10s `APP_RUNTIME_LOCAL_WORKER_RESPONSE_WAIT` headroom checks out.
- Privacy posture holds: new warn logs and audit rows carry only error kinds, ordinals, and aggregate counts. The stored intent hex-encodes the opaque MLS `GroupId` with no 32-byte route-id assumption.

**Tests:** the three new regressions map directly onto the issue's acceptance criteria (duplicate-only yield + durable re-arm across reopen, multi-quantum productive replay with per-quantum audit accounting, queued local command serviced within the bound), and the storage test pins the MAX-arm/exact-epoch-clear CAS. Zeroed retry backoff in the drain tests keeps them deterministic.

One non-blocking inline note: add the new table to the cascade-FK enumeration test. CI is green except for still-running shards at review time.

erskingardner:
Review follow-up pushed in `cbfe1b2d`:

- made the intent ordering assertion independent of `updated_at` wall-clock boundaries
- moved intent ownership from app projection rows to the actual raw MLS `cgka_groups` record, preserving recovery across projection deletion while retaining cascade cleanup
- clarified that production EOSE patience per open attempt is bounded by the 5 s worker quantum
- documented why the policy-specific error-kind expectations are intentionally different

Validation: `cargo test -p storage-sqlite` (415 passed, 4 ignored), 19 feature-enabled epoch-backfill tests, focused default-policy regressions, and `just fast-ci`.

Datawav (commented):
## Automated review

Blocking finding.

1. Blocking — Sparse replay can exhaust fallback attempts and clear recovery intent before history is consumed
- Exact changed-line evidence: In `crates/marmot-app/src/client/sync.rs`, the changed `DrainVerdict::quantum_yield` path returns `NoProgressQuantumYield` whenever `durable_deliveries() == 0`; the changed `spends_unconfirmed_attempt()` includes `NoProgressQuantumYield`; the completion path increments `pending.unconfirmed_attempts`, and once its ordinal reaches `EPOCH_BACKFILL_EOSE_ATTEMPT_LIMIT`, the next drain uses `QuiescenceFallback`. The same changed completion logic clears the exact persisted epoch intent when fallback returns without an error. The productive-yield handling near changed line 2266 clears pacing but does not reset `unconfirmed_attempts`.
- Triggering state: A relay delivers novel stored events less frequently than the five-second execution quantum, or a resubscription spends a quantum replaying only an already-checkpointed duplicate prefix. Three such empty or duplicate-only quanta can occur even while additional history remains available.
- Consequence: Those quanta consume the EOSE-fallback budget. The subsequent fallback can interpret the first 750 ms receive timeout as quiescence, report completion, and delete the durable recovery intent while unconsumed epochs remain.
- Confidence: High. This follows directly from the new verdict classification, strike accounting, unchanged ordinal across productive yields, fallback selection, and intent-clear path. It also invalidates the discussion claim that sparse replay can safely continue for hundreds of productive cycles: interspersed no-novel quanta still accumulate strikes.
- Smallest sound correction: Keep retry pacing for empty or duplicate-only quantum yields, but do not count `NoProgressQuantumYield` as an EOSE-failure strike. Increment the fallback ordinal only for `EoseTimeout` or `NoRelayEose`; alternatively, accumulate consecutive no-novel duration until `EPOCH_BACKFILL_EOSE_WAIT` and reset it whenever durable novel progress occurs. Reserve `QuiescenceFallback` completion and intent clearing for actual sustained quiescence.

Audit evidence:
- Exact head: `cbfe1b2d89813e709d29f0c954ff8fbde58ce999`
- Changed files re-derived include `crates/marmot-app/src/client/sync.rs`, `crates/marmot-app/src/client/epoch_stall.rs`, `crates/marmot-app/src/lib.rs`, `crates/marmot-app/src/tests.rs`, and `crates/storage-sqlite/src/migrations/0052_epoch_backfill_intents.rs`.
- Discussion classification: Preserved the fallback-ordinal defect above. Discarded the reversed test-policy error-kind claim as contradicted by the configured timeout/quantum ordering; discarded the `account_groups` foreign-key claim as stale because this head anchors the intent to `cgka_groups` with `ON DELETE CASCADE`; discarded the ordering-test flake because the head now sorts explicitly; treated malformed-row cleanup as non-blocking and pre-existing-corruption hardening; confirmed the EOSE-wait documentation concern was already addressed.

Risk lenses checked:
- `tests`: assertions fail before the fix; boundary and negative cases; tests exercise production behavior.

Check evidence:
- The bundle reports complete patch coverage for all ten changed files with no missing patches.
- All recorded checks completed successfully, including Rust tests 1/4–4/4, Rust check, clippy, format, audit, relay runtime and relay CLI E2E, CLI mock-relay E2E, simulator/vectors, and Tamarin proofs. These checks do not cover the sparse-progress sequence above, where no-progress quanta are interleaved with eventual novel deliveries.

erskingardner:
Addressed the sparse-replay fallback issue in 7f1ba3ad.

Worker-quantum yields now have a separate no-progress pacing ordinal and do not spend the EOSE-failure budget. Only `EoseTimeout` / `NoRelayEose` advance the quiescence-fallback threshold, while novel progress resets the no-progress pacing ordinal.

Added a regression that drives duplicate-only traffic across the old three-attempt threshold, stops traffic, and verifies the next quiet quantum remains incomplete with its durable intent retained; a subsequent EOSE-confirmed attempt completes normally. The regression failed against the previous head and passes with this change.

Validation:
- `cargo test -p marmot-app --features test-policy-overrides --lib epoch_backfill_ -- --nocapture` (19 passed)
- new threshold-crossing regression (1 passed)
- `just fast-ci`

crates/marmot-app/src/client/sync.rs
```diff
@@ -133,9 +159,32 @@ impl DrainVerdict {
...
         }
     }
+
+    fn quantum_yield(counts: &DrainCounts) -> Self {
+        if counts.durable_deliveries() > 0 {
+            Self::NovelProgressQuantumYield
+        } else {
+            Self::NoProgressQuantumYield
+        }
+    }
```
erskingardner, line 175: `quantum_yield` treats every non-refused ingest as durable progress, but `deliveries - refused` still includes objects the engine dropped with `left_object_unpersisted` — most importantly `Ignored { UnknownGroup }` with no durable trace.

Those deliveries are not added to `seen_events`, and the next attempt resubscribes unfloored, so the same prefix is ingested again. Combined with `made_novel_progress()` clearing `epoch_backfill_retry_not_before`, one unknown-group inbox object is enough to make every quantum look productive and retry immediately. That is the hostile no-EOSE stream from #1562, just not in the duplicate/echo path the new tests cover.

Keep `refused` / `refused_groups` as the resource-refusal audit and re-arm signal. Count every `must_stay_fetchable` delivery separately and exclude that set from `durable_deliveries()`. Then a no-EOSE unknown-group stream yields `NoProgressQuantumYield` and takes the existing backoff.

Please add a regression that streams unpersisted unknown-group traffic without EOSE and asserts the no-progress error kind plus paced retry (and that `mark_replayed` still does not fire).

erskingardner (commented):
Request changes.

The quantum, prefix checkpoint, exact-epoch durable intent, and EOSE-vs-pacing split look right. The #1562 acceptance tests (duplicate-only yield, queued command service, productive multi-quantum replay, restart re-arm) match the issue, and the earlier fallback-ordinal hole is fixed.

One remaining classification bug: `durable_deliveries()` only subtracts resource refusals, so unpersisted unknown-group ingests count as novel progress. Those objects stay fetchable and the next attempt is unfloored, so they replay every quantum, clear the retry floor, and recreate the unbounded worker loop #1562 asked to prevent. Fix the progress count and add a regression; details inline.

erskingardner (commented):
Rereviewed `3b134141` (unchanged since the previous review). Request changes — same blocking issue.

I re-checked the ingest → quantum-yield → retry path. Unpersisted `UnknownGroup` objects still increment `deliveries` without incrementing `refused`, skip `seen_events`, and come back on the next `activate_transport(None)` inbox replay. `made_novel_progress()` then clears the retry floor, so that stream never becomes `NoProgressQuantumYield`. Duplicate/echo traffic is now bounded; this path is not.

One clarification on the earlier note: a quantum yield already cannot reach `mark_replayed` (`error_kind` is set, `succeeded` is false). The PR-specific damage is the unpaced retry loop, not detector disarm. `replay_recovered_something()` using the same helper on a later EOSE/fallback complete is pre-existing.

No additional blocking issues. The quantum, prefix checkpoint, exact-epoch intent CAS, and EOSE-vs-pacing split still look right. Fix `durable_deliveries()` to exclude every `must_stay_fetchable` ingest and add the unknown-group no-EOSE regression.

crates/marmot-app/src/client/sync.rs
erskingardner: Fixed in b4340692. `DrainCounts` now tracks every `must_stay_fetchable` ingest as `unpersisted`, while `refused` remains the narrower resource-refusal audit/re-arm signal. `durable_deliveries()` subtracts all unpersisted deliveries, so repeated unknown-group objects yield `backfill_drain_no_progress_quantum_yield` and retain the automatic retry cooldown. Added an end-to-end same-id unknown-group stream regression that also verifies the object remains fetchable, the next receive seam is deferred, the detector is not disarmed, and `refused == 0`. Validation: both backfill policy suites, the focused regression, and `just fast-ci` pass.

erskingardner (commented):
Recommend merge.

`b434069` fixes the remaining classification bug: `unpersisted` now tracks every `must_stay_fetchable` ingest, `durable_deliveries()` subtracts that set, and `refused` stays the narrower audit/re-arm signal. The new unknown-group regression pins no-progress yield, retry pacing, fetchability, and detector non-disarm. The earlier quantum, prefix-checkpoint, exact-epoch intent, and EOSE-vs-pacing work still looks right.

CI on this head was still running at review time (format already green). No new blocking issues.

crates/marmot-app/src/client/sync.rs
erskingardner: Addressed in `b434069`. `durable_deliveries()` now subtracts `unpersisted` (every `must_stay_fetchable` ingest), and `unpersisted_unknown_group_stream_is_no_progress_and_paced` covers the no-EOSE / paced-retry case.

PR #1567, opened 2026-08-26
fix(app): recover per-account relay queue overflow

### LINKED ISSUE #1561: marmot-app: per-account relay queue overflow can silently skip older catch-up history
## Summary

`MarmotRelayPlane` isolates accounts with a bounded `ACCOUNT_DELIVERY_BUFFER = 1024` queue and non-blocking `try_send`. That correctly prevents one slow account from blocking the shared relay router, but a full queue silently drops that account's transport delivery and assumes a later subscription catch-up will recover it.

That assumption is not safe. Queue overflow does not persist a recovery marker, invalidate end-of-stored-events completion, or disable the ordinary subscription's cursor floor. EOSE is recorded independently of the per-account delivery queue. Meanwhile, successfully processed deliveries advance a persisted monotonic `last_transport_timestamp`, and the next ordinary subscription starts at only `cursor - 120s` in production.

If a long-offline/reconnecting account receives more than 1024 stored events while its worker is slow, and newer relay events are processed while older/reordered events are among those dropped, the cursor can advance past the missing history. EOSE can still report that the relay finished. A later ordinary catch-up can then exclude the dropped events permanently unless some separate unfloored recovery happens to be armed.

This is especially likely to surface under the conditions reported in #1560: a long-offline account or intermittent relay burst feeding an account worker slowed by expensive convergence/SQLite work.

## Evidence

- `crates/marmot-app/src/relay_plane/mod.rs:61` fixes the per-account buffer at 1024.
- `crates/marmot-app/src/relay_plane/mod.rs:668-733` uses `try_send`; `Full` only logs `dropping transport delivery: account delivery queue full` and discards the delivery.
- `crates/transport-nostr-adapter/src/sdk_client.rs:300-315` records relay EOSE on a separate notification path; it does not know whether the app-layer account queue dropped earlier deliveries.
- `crates/marmot-app/src/client/sync.rs:2679-2699` advances the durable cursor as a monotonic maximum of inbound Nostr `created_at` timestamps.
- `crates/marmot-app/src/relay_plane/mod.rs:307-337` rebuilds the subscription from `last_transport_timestamp - lookback`; the production lookback is 120 seconds (`crates/marmot-app/src/lib.rs:389`).

This is not a duplicate of #763. That issue was closed because the existing per-account `try_send` design prevents cross-account head-of-line blocking; its closing comment explicitly says per-account overflow was not solved.

## Required behavior

Preserve cross-account isolation, but make overflow a first-class incomplete-recovery condition rather than an invisible drop. A sound design could use a durable bounded spool, or durably mark an overflow generation and force a safe/unfloored replay before trusting the cursor/EOSE gate again. It must not confirm catch-up across an unresolved gap.

Keep logs and metrics privacy-safe: aggregate counts/depths only, with no account/group/message identifiers or relay URLs.

## Acceptance criteria

- A deterministic test delivers more than 1024 stored events for one account, with a deliberately slow consumer, out-of-order/newest-first timestamps, and EOSE after the burst.
- Every delivery is either durably retained for processing or represented by an explicit durable incomplete-recovery state; no unresolved drop is treated as completed catch-up.
- A newer processed event cannot make the next subscription's `since` floor permanently exclude an older event dropped during the same overflow generation.
- Restart/reopen while overflow recovery is pending still requests a range that contains the missing history.
- EOSE observed after overflow does not mark the affected catch-up complete until the gap is resolved.
- A slow/overflowing account still cannot block delivery, EOSE, or commands for other accounts.
- Overflow/recovery metrics report aggregate queue depth, dropped/spooled count, retry outcome, and elapsed time without sensitive identifiers.

## Related

- #1560 — convergence write amplification can make the account consumer slow enough to amplify this path.
- #1546 — connectivity-restored wakeup/backoff latency is separate; fixing it does not recover an overflow gap.

erskingardner:
## Summary
- reserve an explicit per-account overflow control record and durably mark each incomplete recovery generation before releasing omitted deliveries
- invalidate ordinary EOSE and cursor floors until a fresh unfloored, EOSE-confirmed replay clears the exact durable generation
- preserve cross-account routing isolation while exposing aggregate queue, drop, retry outcome, and elapsed-time health metrics
- add restart-safe SQLite state plus deterministic newest-first overflow, marker retry, EOSE, and cross-account regression coverage

## Test plan
- [x] cargo test -p storage-sqlite
- [x] cargo test -p marmot-app --features test-policy-overrides
- [x] just fast-ci
- [x] New overflow regression failed before the fix and passes after it

Fixes #1561

## Summary by CodeRabbit

- **Reliability**
  - Added automatic recovery when incoming account deliveries overflow, preserving progress and retrying incomplete recovery safely.
  - Recovery state now persists across restarts and clears only after confirmed completion.

- **Synchronization**
  - Improved replay and backfill behavior when relays are unavailable or only partially synchronized.
  - Durable cursor advancement now avoids skipping unconfirmed deliveries.

- **Messaging**
  - Send results now report publication status for each message, including unresolved completion states.

### DIFF
--- a/crates/marmot-app/src/client/sync.rs
+++ b/crates/marmot-app/src/client/sync.rs
@@ -460,7 +460,7 @@ impl AppClient {
             None
         } else {
             self.relay_plane
-                .subscription_rebuild_since(self.state.last_transport_timestamp)
+                .subscription_rebuild_since(self.checkpointed_transport_timestamp)
         }
     }
 
@@ -468,9 +468,9 @@ impl AppClient {
         &mut self,
         overflow: crate::relay_plane::AccountDeliveryOverflow,
     ) -> Result<(), AppError> {
-        // This write precedes any checkpoint of the already-drained prefix.
-        // A crash can therefore leave a conservative recovery marker with an
-        // older cursor, but never a newer cursor without the marker.
+        // The router's process-local fence already froze cursor advancement at
+        // the omission. Re-observe the same token here so the queued signal is
+        // also an idempotent storage boundary before recovery starts.
         self.app
             .account_storage(&self.state.label)?
             .mark_account_delivery_recovery(
@@ -700,13 +700,8 @@ impl AppClient {
         let mut counts = DrainCounts::default();
         let (mut summary, drain_verdict) = self.sync_sdk_relay(&mut counts).await?;
         if drain_verdict == DrainVerdict::Overflow || self.delivery_overflow_recovery_pending {
-            match self.recover_delivery_overflow().await {
-                Ok(recovered) => summary.merge(recovered),
-                Err(mut failure) => {
-                    failure.partial_summary.merge(summary);
-                    return Err(failure);
-                }
-            }
+            self.recover_delivery_overflow_and_merge(&mut summary)
+                .await?;
         }
         // Surface engine events queued without an inbound delivery — most
         // importantly `GroupHydrationQuarantined`, queued during session
@@ -1019,8 +1014,9 @@ impl AppClient {
                     self.ingest_received_delivery(*delivery).await?
                 }
                 crate::relay_plane::AccountDeliveryReceive::Overflow(_) => {
-                    match self.recover_delivery_overflow().await {
-                        Ok(summary) => summary,
+                    let mut summary = SyncSummary::default();
+                    match self.recover_delivery_overflow_and_merge(&mut summary).await {
+                        Ok(()) => summary,
                         Err(failure) => {
                             self.pending_failed_sync_summary
                                 .merge(failure.partial_summary);
@@ -1097,12 +1093,21 @@ impl AppClient {
         &mut self,
         delivery: cgka_traits::TransportDelivery,
     ) -> Result<SyncSummary, AppError> {
+        let cursor_before_secs = self.state.last_transport_timestamp;
         let display_names = self.app.display_names_by_id()?;
         let mut summary = SyncSummary::default();
         let event_id = hex::encode(delivery.message.id.as_slice());
         let ingested = self
             .ingest_delivery(delivery, &display_names, &mut summary)
             .await?;
+        if self.adapter.pending_delivery_overflow().is_some() {
+            // `record_drop` publishes this process-local fence at the exact
+            // omission, before marker I/O or the reserved control record can
+            // complete. Keep this per-delivery checkpoint on its pre-ingest
+            // floor so slow SQLite cannot commit the newest-first prefix ahead
+            // of the marker that represents the omitted older delivery.
+            self.state.last_transport_timestamp = cursor_before_secs;
+        }
         // Mark the delivery seen only after durable ingest succeeds, matching
         // the catch-up drain below. Marking at receive time would let a failed
         // ingest poison the index, so a reused client would silently skip the
@@ -1223,7 +1228,7 @@ impl AppClient {
             .drain_sdk_relay(
                 &mut counts,
                 DrainCompletion::EndOfStoredEvents {
-                    silence_budget: self.epoch_backfill_eose_wait(),
+                    silence_budget: self.delivery_overflow_eose_wait(),
                     execution_quantum: self.epoch_backfill_execution_quantum(),
                 },
             )
@@ -1272,6 +1277,8 @@ impl AppClient {
                 elapsed_ms = started.elapsed().as_millis() as u64,
                 "account delivery overflow recovery completed",
             );
+            let mut summary = summary;
+            self.drain_epoch_stall_escalations(&mut summary);
             return Ok(summary);
         }
 
@@ -1297,6 +1304,22 @@ impl AppClient {
             )),
             SyncFailureStage::RelayReceive,
         ))
+    }
+
+    async fn recover_delivery_overflow_and_merge(
+        &mut self,
+        summary: &mut SyncSummary,
+    ) -> Result<(), ClassifiedSyncFailure> {
+        match self.recover_delivery_overflow().await {
+            Ok(recovered) => {
+                summary.merge(recovered);
+                Ok(())
+            }
+            Err(mut failure) => {
+                failure.partial_summary.merge(std::mem::take(summary));
+                Err(failure)
+            }
+        }
     }
 
     /// How long an unconfirmed replay must wait before an automatic seam may
@@ -1338,6 +1361,13 @@ impl AppClient {
             return Duration::from_millis(ms);
         }
         EPOCH_BACKFILL_EOSE_WAIT
+    }
+
+    /// The EOSE budget for durable account-delivery overflow repair. It shares
+    /// today's configured value with epoch backfill, but remains a distinct
+    /// policy seam so either recovery mechanism can be tuned independently.
+    fn delivery_overflow_eose_wait(&self) -> Duration {
+        self.epoch_backfill_eose_wait()
     }
 
     /// Maximum wall-clock quantum one backfill drain owns the account worker.
@@ -1435,7 +1465,7 @@ impl AppClient {
         // `SDK_DRAIN_WAIT`. Held at the same interval as that timeout.
         let mut gate_polled_at = silence_started;
 
-        let verdict = loop {
+        let mut verdict = loop {
             if completion
                 .execution_quantum()
                 .is_some_and(|quantum| drain_started.elapsed() >= quantum)
@@ -1600,6 +1630,28 @@ impl AppClient {
             routes_dirty |= ingested.routes_dirty;
         };
 
+        if verdict != DrainVerdict::Overflow
+            && let Some(overflow) = self.adapter.pending_delivery_overflow()
+        {
+            // The queue signal intentionally trails marker persistence, so a
+            // quiescence timeout can win while that signal is still pending.
+            // Consult the immediate process-local fence before checkpointing.
+            self.state.last_transport_timestamp = cursor_before_secs;
+            if let Err(error) = self.observe_delivery_overflow(overflow) {
+                return Err(self
+                    .finish_failed_sync_drain(
+                        summary,
+                        routes_dirty,
+                        *counts,
+                        StagedSyncError::new(error, SyncFailureStage::StatePersist),
+                        drain_started,
+                        cursor_before_secs,
+                    )
+                    .await);
+            }
+            verdict = DrainVerdict::Overflow;
+        }
+
         if let Err(error) = self
             .checkpoint_sync_prefix(&mut summary, routes_dirty, counts.deliveries)
             .await
@@ -1682,6 +1734,10 @@ impl AppClient {
         } else {
             false
         };
+        let checkpointed_before = self.checkpointed_transport_timestamp;
+        if self.adapter.pending_delivery_overflow().is_none() {
+            self.checkpointed_transport_timestamp = self.state.last_transport_timestamp;
+        }
         let checkpoint = if cfg!(feature = "test-policy-overrides")
             && self
                 .app
@@ -1696,6 +1752,7 @@ impl AppClient {
             self.save_state_with_pending_local_group_deletion_frontier_clears()
         };
         if let Err(error) = checkpoint {
+            self.checkpointed_transport_timestamp = checkpointed_before;
             return Err(SyncCheckpointError::BeforePersistence(error));
         }
 
@@ -2673,13 +2730,8 @@ impl AppClient {
         let mut counts = DrainCounts::default();
         let (mut summary, drain_verdict) = self.sync_sdk_relay(&mut counts).await?;
         if drain_verdict == DrainVerdict::Overflow || self.delivery_overflow_recovery_pending {
-            match self.recover_delivery_overflow().await {
-                Ok(recovered) => summary.merge(recovered),
-                Err(mut failure) => {
-                    failure.partial_summary.merge(summary);
-                    return Err(failure);
-                }
-            }
+            self.recover_delivery_overflow_and_merge(&mut summary)
+                .await?;
         }
         let drained = match self.drain_pending_session_events().await {
             Ok(drained) => drained,
@@ -3005,7 +3057,7 @@ impl AppClient {
         let delta = AccountState {
             label: self.state.label.clone(),
             seen_events: self.state.seen_events[seen_start..].to_vec(),
-            last_transport_timestamp: self.state.last_transport_timestamp,
+            last_transport_timestamp: self.checkpointed_transport_timestamp,
             groups: self
                 .state
                 .groups
@@ -3318,10 +3370,11 @@ impl AppClient {
         Ok(routes_dirty)
     }
 
-    /// Advance the persisted transport cursor from an inbound message —
-    /// unless this runtime was constructed with
+    /// Advance the persisted transport cursor from an inbound message unless
+    /// this runtime was constructed with
     /// [`CursorPersistence::Frozen`](crate::CursorPersistence), in which case
-    /// this is a no-op and the cursor stays at the value loaded from the store.
+    /// this is a no-op, or the account route has already recorded a queue
+    /// omission whose marker/control record is still in flight.
     ///
     /// `timestamp` is the sender-controlled Nostr `created_at` of the outer
     /// kind-445 event and is never validated upstream. The cursor is a
@@ -3332,6 +3385,9 @@ impl AppClient {
     /// wall-clock plus a bounded skew so a hostile or clock-skewed sender can
     /// move the cursor no further than `now + TRANSPORT_CURSOR_MAX_FUTURE_SKEW`.
     fn remember_transport_cursor(&mut self, timestamp: u64) {
+        if self.adapter.pending_delivery_overflow().is_some() {
+            return;
+        }
         self.state.last_transport_timestamp = next_transport_cursor(
             self.app.cursor_persistence(),
             self.state.last_transport_timestamp,
--- a/crates/marmot-app/src/client/sync.rs
+++ b/crates/marmot-app/src/client/sync.rs
@@ -1501,7 +1501,7 @@ impl AppClient {
                             .finish_failed_sync_drain(
                                 summary,
                                 routes_dirty,
-                                *counts,
+                                counts.clone(),
                                 StagedSyncError::new(error, SyncFailureStage::StatePersist),
                                 drain_started,
                                 cursor_before_secs,
@@ -1642,7 +1642,7 @@ impl AppClient {
                     .finish_failed_sync_drain(
                         summary,
                         routes_dirty,
-                        *counts,
+                        counts.clone(),
                         StagedSyncError::new(error, SyncFailureStage::StatePersist),
                         drain_started,
                         cursor_before_secs,
--- a/crates/marmot-app/src/client/sync.rs
+++ b/crates/marmot-app/src/client/sync.rs
@@ -1,9 +1,9 @@
 use std::collections::{HashMap, HashSet};
 use std::time::{Duration, Instant};
 
+use cgka_traits::GroupId;
 use cgka_traits::app_event::{MARMOT_APP_EVENT_KIND_CHAT, MARMOT_APP_EVENT_KIND_DELETE};
 use cgka_traits::ingest::IngestOutcome;
-use cgka_traits::{GroupId, TransportAdapter};
 use storage_sqlite::clamp_to_max_future_skew;
 use tokio::time::timeout;
 use transport_nostr_adapter::AccountSubscriptionEose;
@@ -130,6 +130,10 @@ enum DrainVerdict {
     /// The account-worker quantum ended without durable novel progress. This
     /// includes duplicate/echo-only and all-refused streams.
     NoProgressQuantumYield,
+    /// The per-account queue omitted at least one delivery. The durable marker
+    /// is armed, but this drain's (possibly floored) subscription cannot close
+    /// the gap; the caller must issue a fresh unfloored replay.
+    Overflow,
 }
 
 impl DrainVerdict {
@@ -141,6 +145,7 @@ impl DrainVerdict {
             Self::NoRelayEose => Some("backfill_drain_no_relay_eose"),
             Self::NovelProgressQuantumYield => Some("backfill_drain_novel_progress_quantum_yield"),
             Self::NoProgressQuantumYield => Some("backfill_drain_no_progress_quantum_yield"),
+            Self::Overflow => Some("account_delivery_queue_overflow"),
         }
     }
 
@@ -150,6 +155,7 @@ impl DrainVerdict {
             Self::Complete => Some(EpochBackfillCompletionKind::EndOfStoredEvents),
             Self::EoseTimeout
             | Self::NoRelayEose
+            | Self::Overflow
             | Self::NovelProgressQuantumYield
             | Self::NoProgressQuantumYield => None,
         }
@@ -435,13 +441,71 @@ impl AppClient {
     }
 
     pub(crate) async fn sync_runtime_groups(&mut self) -> Result<(), AppError> {
-        let rebuild_since = self
-            .relay_plane
-            .subscription_rebuild_since(self.state.last_transport_timestamp);
+        let rebuild_since = self.subscription_rebuild_since();
+        self.sync_runtime_groups_since(rebuild_since).await
+    }
+
+    async fn sync_runtime_groups_since(
+        &mut self,
+        rebuild_since: Option<cgka_traits::transport::Timestamp>,
+    ) -> Result<(), AppError> {
         self.warm_encrypted_media_epoch_secrets("pre_subscription_sync");
         self.runtime.sync_transport_groups(rebuild_since).await?;
         self.warm_encrypted_media_epoch_secrets("post_subscription_sync");
         Ok(())
+    }
+
+    pub(crate) fn subscription_rebuild_since(&self) -> Option<cgka_traits::transport::Timestamp> {
+        if self.delivery_overflow_recovery_pending {
+            None
+        } else {
+            self.relay_plane
+                .subscription_rebuild_since(self.state.last_transport_timestamp)
+        }
+    }
+
+    fn observe_delivery_overflow(
+        &mut self,
+        overflow: crate::relay_plane::AccountDeliveryOverflow,
+    ) -> Result<(), AppError> {
+        // This write precedes any checkpoint of the already-drained prefix.
+        // A crash can therefore leave a conservative recovery marker with an
+        // older cursor, but never a newer cursor without the marker.
+        self.app
+            .account_storage(&self.state.label)?
+            .mark_account_delivery_recovery(
+                &self.state.label,
+                overflow.marker_token,
+                overflow.dropped,
+            )?;
+        self.delivery_overflow_recovery_pending = true;
+        self.delivery_overflow_recovery_marker_token = Some(overflow.marker_token);
+        tracing::warn!(
+            target: "marmot_app::relay_plane",
+            method = "observe_delivery_overflow",
+            queue_depth = overflow.queue_depth,
+            dropped = overflow.dropped,
+            elapsed_ms = overflow.elapsed_ms,
+            "account delivery queue overflow requires unfloored recovery",
+        );
+        Ok(())
+    }
+
+    fn clear_delivery_overflow_recovery(&mut self, marker_token: u64) -> Result<bool, AppError> {
+        let storage = self.app.account_storage(&self.state.label)?;
+        let cleared = storage.clear_account_delivery_recovery(&self.state.label, marker_token)?;
+        if cleared {
+            self.delivery_overflow_recovery_pending = false;
+            self.delivery_overflow_recovery_marker_token = None;
+            return Ok(true);
+        }
+        // A newer process-local generation won the storage race. Keep its
+        // marker authoritative; this replay must not clear or report it.
+        let current = storage.account_delivery_recovery(&self.state.label)?;
+        self.delivery_overflow_recovery_pending = current.is_some();
+        self.delivery_overflow_recovery_marker_token =
+            current.map(|recovery| recovery.marker_token);
+        Ok(false)
     }
 
     /// Warm the encrypted-media epoch-secret cache around a subscription sync,
@@ -515,9 +579,7 @@ impl AppClient {
         self.relay_plane
             .set_transport_signer(self.transport_signer.clone())
             .await;
-        let rebuild_since = self
-            .relay_plane
-            .subscription_rebuild_since(self.state.last_transport_timestamp);
+        let rebuild_since = self.subscription_rebuild_since();
         let activation = self.runtime.activate_transport(rebuild_since).await;
         if let Some(telemetry) = telemetry {
             telemetry.record(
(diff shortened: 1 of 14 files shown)


erskingardner (commented):
Self-review note — GitHub rejects approvals on your own PR, so this is a comment review. Content below would otherwise be an approve.

Solid fix for a real data-loss vector — the analysis from #1561 checks out, and the implementation is conservative in the right direction.

**Premise:** Strong. #1561 documents a genuine silent-history-skip bug in the queue-full `try_send` drop path. This fixes the root cause (no durable recovery marker + cursor/EOSE trust) rather than the symptom. The mechanism is complex, but the problem — durable compare-and-clear generation guard across restarts, EOSE gating, per-account isolation — is intrinsically hard, and the design chooses safety at every decision point (marker persisted before cursor checkpoint, random-token compare-and-clear, EOSE-only-clears).

**Verified during review:**
- Reserved `+1` channel slot guarantees the overflow control record always fits after ordinary capacity is exhausted; a `Full` on the ordinary path now correctly indicates a violated invariant.
- Durable marker is persisted via the `spawn_blocking` barrier before the overflow signal is released; the retry loop holds omitted deliveries in their task (documented tradeoff).
- `subscription_rebuild_since()` returns `None` while pending, so every rebuild path (activation, prepare, group sync) goes unfloored; all call sites were updated.
- `account_subscription_eose()` zeroes `with_eose` while blocked, and `complete()` requires `with_eose == subscriptions`, so ordinary EOSE confirmation cannot pass during a pending gap.
- Compare-and-clear on a random 63-bit token; the storage upsert keeps `pending_since` and `max(dropped)` for the same token, so restart and re-observation are safe.
- Recovery re-activates transport + group subscriptions with `since=None`, drains to EOSE, and only then clears; any new drop during the replay mismatches the generation and stays pending.
- New logs carry explicit `target`/`method` with aggregate counts only — compliant with the observability invariant.
- No production callers of the marker-less `account_adapter` remain; the trait `receive()` now fails loudly on Overflow instead of silently skipping.
- Tests are strong: deterministic buffer overflow with an injected marker-write failure, isolation assertions, EOSE gating, durable reopen, migration bump (52) with version-test updates. CI is green.

One inline suggestion on the marker-less constructor follows.

crates/marmot-app/src/relay_plane/mod.rs
```diff
@@ -272,16 +632,41 @@ impl MarmotRelayPlane {
         &self,
         account_id: MemberId,
         publish_client: Arc<dyn NostrRelayClient>,
+    ) -> MarmotRelayPlaneAccountAdapter {
+        self.account_adapter_with_recovery_marker(account_id, publish_client, None)
```
erskingardner, line 667: Non-blocking suggestion: worth documenting that this marker-less constructor is effectively test-only. A `None` marker downgrades overflow to a best-effort queue signal (dropped with a warn on enqueue failure), which would silently reintroduce the #1561 gap for any future production caller. Either a doc comment on `account_adapter` ("production callers must use `account_adapter_with_recovery_marker` with `Some(marker)`") or a `#[cfg(test)]` gate would prevent an accidental regression.

crates/marmot-app/src/relay_plane/mod.rs
```diff
@@ -712,20 +1103,77 @@ impl MarmotRelayPlane {
...
+                    if route.sender.capacity() <= 1 {
+                        let signal_generation = route.overflow.record_drop(queue_depth);
+                        if let Some(marker) = route.recovery_marker.clone() {
+                            // Account-local persistence runs independently of
+                            // the shared router. Hold this omitted delivery in
+                            // the task until the durable marker barrier returns;
+                            // other accounts, EOSE, and commands keep flowing.
+                            let sender = route.sender.clone();
+                            let overflow_state = route.overflow.clone();
+                            tokio::spawn(async move {
```
erskingardner: Non-blocking: only the signal owner needs this task. The marker-before-checkpoint invariant comes from the **signal enqueue** trailing the durable barrier (the client can't receive the Overflow record, and hence can't checkpoint the drained prefix, until the marker lands) — holding the `TransportDelivery` itself adds no ordering guarantee. Dropping it at spawn time would remove the unbounded memory retention while the marker write persistently fails (one held delivery + one spawned task per dropped event in that window), and skipping the spawn entirely when `signal_generation.is_none()` would avoid a task spawn per dropped event even after the marker is durable. If the hold is deliberate for some reason I'm not seeing, a comment saying what it protects would help.

crates/marmot-app/src/client/sync.rs
```diff
@@ -600,7 +660,16 @@ impl AppClient {
         // drained registration log before draining inbound deliveries.
         self.record_subscription_rebuild(rebuild_since_secs).await;
         let mut counts = DrainCounts::default();
-        let mut summary = self.sync_sdk_relay(&mut counts).await?;
+        let (mut summary, drain_verdict) = self.sync_sdk_relay(&mut counts).await?;
+        if drain_verdict == DrainVerdict::Overflow || self.delivery_overflow_recovery_pending {
```
erskingardner, line 714: Non-blocking: this invocation block is duplicated verbatim at line 2414 (`repair_full_history`), and near-identically in `next_event` and the account worker. A small helper (`recover_delivery_overflow_and_merge(&mut summary) -> Result<(), ClassifiedSyncFailure>`) would collapse the four sites and keep the merge/error-wiring consistent if it ever changes. Separately, on this path `prepare_transport_for_sync` already activated unfloored (pending forces `rebuild_since = None`), so recovery's second `activate_transport(None)` is a full redundant subscription rebuild — the re-issue is needed for EOSE freshness, but the first one could be skipped when recovery is going to run immediately.

crates/marmot-app/src/tests.rs
```diff
@@ -8415,6 +8415,80 @@ fn legacy_account_projection_clamps_poisoned_transport_cursor_on_import() {
     );
 }
 
+#[test]
+fn durable_delivery_overflow_marker_forces_unfloored_account_reopen() {
```
erskingardner, line 9246: Non-blocking: this test and the relay-plane test verify the mechanism (durable marker, `since: None`, EOSE gate, metrics), but neither asserts the actual end-to-end property from the issue — that an older event dropped during the overflow generation is re-delivered and ingested by the recovery replay. Since the replay in this test runs against a relay serving no events, a regression that broke re-delivery during recovery (e.g., a seen-events or activation bug) could still pass. Consider scripting one dropped older event into the recovery replay and asserting it lands in the ingest summary.

erskingardner (commented):
Reviewed against #1561 (posting as comment since the API rejects self-approval). The premise is sound and the design matches the issue's required behavior: overflow becomes a durable, first-class incomplete-recovery condition (generation + random marker token, SQLite row in migration 0052), ordinary EOSE and the cursor floor stay invalidated until a fresh unfloored, EOSE-confirmed replay resolves the exact generation, and cross-account isolation is preserved (verified by the Bob-keeps-flowing test). The marker-before-checkpoint ordering, the compare-and-clear token guard, the dropped-count guard against clearing a generation that overflowed again mid-replay, and the restart-synthetic-generation path all check out — I traced the crash windows and the process-local/durable races and they resolve conservatively. Logs and the new health fields are aggregate-only, per the observability invariant.

All findings below are non-blocking robustness/simplicity notes. CI is green and the new regression coverage is deterministic and targeted.

erskingardner (commented):
**Blocking findings** (posting as comment — GitHub rejects change-requests on your own PR).

Requesting changes — the durable-marker design is the right fix for #1561, but the marker-before-checkpoint invariant does not hold on the production receive path, so the original silent-history-skip crash window is still open.

`record_drop` correctly arms process-local `pending` immediately, and EOSE is gated off that bit. Cursor persistence and catch-up completeness do not: they wait for the delayed overflow queue record. `ingest_received_delivery` max-merges `last_transport_timestamp` per event while the marker write is still in flight (and typically queued behind that same SQLCipher lock). A crash in that window is exactly #1561 — newer cursor, no recovery row.

The per-drop `tokio::spawn` that holds omitted deliveries until persist succeeds does not close that window and is unsafe on the issue’s own trigger (slow SQLite): one overflowing account can retain an unbounded number of deliveries/tasks and OOM the process.

Tests never exercise cursor checkpoint during a delayed marker write, so they stay green.

Please close the race (freeze durable cursor advancement, or persist the marker in the same `save_state` transaction, as soon as `record_drop` fires), drop the per-delivery hold/spawn, and add a regression that would have failed here. Other notes in the existing self-review (marker-less constructor, helper extraction) are fine as follow-ups.

crates/marmot-app/src/relay_plane/mod.rs
erskingardner: This persist-before-signal barrier does not protect the prefix the account worker is concurrently checkpointing, which is the #1561 path.

`record_drop` sets process-local `pending`, then this task tries to write the marker. Meanwhile the slow consumer is draining the already-queued newest-first 1024 via `ingest_received_delivery` → `save_state`, which max-merges `last_transport_timestamp`. Storage is one mutexed SQLCipher connection, so the worker typically holds the lock and persists a newer cursor *before* this marker write. Crash in that window: advanced cursor, no durable row — the original bug.

Holding each omitted `TransportDelivery` until persist returns does not close that window. Under the issue’s own trigger (SQLite already slow / marker retries) every later drop spawns another task that retains a delivery for the life of the wedge. One overflowing account can OOM the process and take every other account down.

The process-local `pending` bit is available immediately. Freeze durable cursor advancement (or persist the marker in the same `save_state` transaction) as soon as `record_drop` fires; one overflow task, no held deliveries. Drain/sync completeness should consult that bit too — quiescence can currently return `Complete` while persist is still in flight and the client flag is still false.

crates/marmot-app/src/client/sync.rs
```diff
@@ -398,15 +403,73 @@ impl AppClient {
...
+        }
+    }
+
+    fn observe_delivery_overflow(
+        &mut self,
+        overflow: crate::relay_plane::AccountDeliveryOverflow,
+    ) -> Result<(), AppError> {
+        // This write precedes any checkpoint of the already-drained prefix.
+        // A crash can therefore leave a conservative recovery marker with an
+        // older cursor, but never a newer cursor without the marker.
```
erskingardner: This comment is not true on the production receive path. `observe_delivery_overflow` runs only after the reserved overflow record is dequeued, which is after the worker has already ingested and `save_state`’d the 1024-event prefix (`ingest_received_delivery`). Those saves max-merge the cursor; rolling `last_transport_timestamp` back later cannot unwrite them (`account_projection_state_keeps_max_cursor_across_racing_saves`).

The write here is at best a second copy of the router’s persist. The invariant has to hold *before* those per-delivery checkpoints, by consulting process-local overflow state (set in `record_drop`) rather than waiting for this queue event.

crates/marmot-app/src/tests.rs
erskingardner, line 9246: Neither this test nor `account_queue_overflow_invalidates_eose_without_blocking_other_accounts` would fail if the cursor/marker race shipped. This one seeds the marker, then recovers against an empty relay. The relay-plane test never checkpoints `last_transport_timestamp`.

Please add a regression that: (1) newest-first overflow omits an older event, (2) the consumer persists the advanced cursor *while the marker write is delayed/failing*, (3) reopen must still have a recovery row (or an unfloored `since`) that contains the omitted event — the case that is possible today is reopen with no row and a floor that excludes it. Then assert the omitted event is actually ingested by recovery.

crates/marmot-app/src/relay_plane/mod.rs
erskingardner: Added a doc comment that explicitly limits the marker-less constructor to tests/compatibility embedders and directs production callers to account_adapter_with_recovery_marker.

crates/marmot-app/src/client/sync.rs
erskingardner: Corrected the comment and fixed the production receive path. record_drop now exposes an immediate process-local fence; live ingest restores the pre-ingest cursor when that fence is armed, and projection saves write checkpointed_transport_timestamp rather than the newer candidate. The queued signal re-observes the token idempotently instead of being treated as the first protection point.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1237,7 +1237,20 @@ async fn run_app_runtime_account_worker(
...
+                        match client.recover_delivery_overflow().await {
+                            Ok(summary) => Ok(summary),
+                            Err(failure) => {
+                                client
+                                    .pending_failed_sync_summary
+                                    .merge(failure.partial_summary);
+                                Err(failure.source)
+                            }
+                        }
+                    }
```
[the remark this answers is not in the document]
erskingardner: Overflow recovery failure is now non-fatal in the receive arm: it publishes the account error, retains the partial summary, keeps the client/session alive, and defers retry to the next seam instead of entering reconnect backoff. It also skips the immediate same-seam backfill attempt to avoid duplicate recovery work.

crates/marmot-app/src/client/sync.rs
erskingardner: Extracted recover_delivery_overflow_and_merge and reused it across sync, next_event, and repair. I kept the fresh activation in recovery intentionally: it creates a new unfloored subscription/EOSE generation, so recovery never accepts EOSE state from the prepare activation that preceded the overflow decision.

crates/marmot-app/src/relay_plane/mod.rs
erskingardner: Refactored this path so only the signal owner can start one marker worker for the generation. The omitted delivery is dropped immediately, and later omissions retain no payload/task; they only update the bounded counter. The one worker releases the control signal after marker durability (or terminal storage closure).

crates/marmot-app/src/relay_plane/mod.rs
```diff
@@ -712,20 +1103,77 @@ impl MarmotRelayPlane {
...
+                                    );
+                                }
+                            });
+                        } else if let Some(generation) = signal_generation {
+                            enqueue_account_delivery_overflow_signal(
+                                &route.sender,
+                                &route.overflow,
+                                generation,
+                            );
+                        }
```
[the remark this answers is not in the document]
erskingardner: Reworked the route to one marker task per generation with no captured TransportDelivery. Later drops update only the aggregate counter, preserving the queue's memory bound even while marker persistence retries.

crates/marmot-app/src/tests.rs
erskingardner: Added process_local_overflow_fence_freezes_cursor_while_marker_write_retries. It delays marker persistence, processes newer events, overflows newest-first with an older omitted event, checkpoints a retained prefix, and asserts the durable cursor remains at the pre-overflow drain floor while no marker row exists. The reopen test separately proves that the older event is ingested by the unfloored replay.

crates/marmot-app/src/relay_plane/mod.rs
```diff
@@ -122,7 +124,345 @@ pub struct MarmotRelayPlaneAccountAdapter {
...
+                    false
+                } else {
+                    state.marker_in_progress = true;
+                    true
+                }
+            };
+            if !owns_write {
+                notified.await;
+                continue;
+            }
```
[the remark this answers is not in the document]
erskingardner: Removed the Notify/waiter design entirely. Each generation has a single marker worker claimed under the overflow-state mutex, so there is no waiter registration window or missed notify wakeup.

crates/marmot-app/src/client/sync.rs
```diff
@@ -1046,6 +1137,128 @@ impl AppClient {
...
+                method = "recover_delivery_overflow",
+                queue_depth = attempt.queue_depth,
+                dropped = attempt.dropped,
+                deliveries = counts.deliveries,
+                skipped = counts.skipped,
+                elapsed_ms = started.elapsed().as_millis() as u64,
+                "account delivery overflow recovery completed",
+            );
+            return Ok(summary);
+        }
```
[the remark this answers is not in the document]
erskingardner: The successful overflow-recovery path now drains pending epoch-stall escalations into the returned summary after recovery completion/logging and before Ok(summary).

crates/marmot-app/src/lib.rs
```diff
@@ -3488,7 +3498,19 @@ impl MarmotApp {
...
+            Arc::new(move |marker_token, dropped| {
+                recovery_storage
+                    .mark_account_delivery_recovery(&recovery_label, marker_token, dropped)
+                    .map_err(|_| ())
+            });
+        let adapter = relay_plane.account_adapter_with_recovery_marker(
+            account_id.clone(),
+            publish_client,
+            Some(recovery_marker),
+        );
```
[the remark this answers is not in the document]
erskingardner: Preserved storage closure as a typed terminal AccountDeliveryRecoveryMarkerError::Closed. The single marker worker stops retrying on Closed and releases the overflow control signal; retryable storage errors retain the bounded retry behavior.

crates/marmot-app/src/relay_plane/mod.rs
erskingardner: Closed this race by separating the in-memory cursor candidate from the durable drain checkpoint. One-at-a-time prefix ingests persist the last completed drain floor, while the router-local overflow bit fences both live saves and drain completion before marker I/O/control delivery. The delayed-marker regression advances a newer in-memory prefix, triggers overflow, and verifies the database still contains the older safe floor while the recovery row is absent.

crates/marmot-app/src/tests.rs
erskingardner: Extended the reopen regression with an actual older kind-445 event below the ordinary floor. The unfloored recovery replay now ingests it, and the test asserts its event ID is durably present in seen_events before the marker is cleared.

erskingardner:
Addressed all review feedback in 53389e85:

- made cursor persistence drain-checkpointed and fenced by immediate process-local overflow state;
- reduced overflow handling to one marker worker per generation with no omitted payload retention;
- made storage closure terminal, removed the Notify waiter race, and kept incomplete recovery non-fatal to the account session;
- added actual older-event recovery plus delayed-marker/cursor crash-window regressions;
- covered replacement-token storage behavior, added a dedicated overflow EOSE policy seam, extracted recovery/merge handling, and drained epoch-stall escalations on success.

Validation: just fast-ci; marmot-app library tests with test-policy-overrides (796 passed, 1 ignored); cursor_persistence integration tests; targeted relay overflow, durable reopen, delayed-marker cursor-fence, and storage marker tests.

crates/marmot-app/src/client/sync.rs
```diff
@@ -1046,6 +1142,146 @@ impl AppClient {
...
+            )
+            .await
+        {
+            Ok(result) => result,
+            Err(error) => {
+                self.adapter.fail_delivery_overflow_recovery();
+                return Err(error);
+            }
+        };
+        if verdict == DrainVerdict::Complete
```
erskingardner, line 1257: [P1] Do not clear overflow recovery on any-relay EOSE. `DrainVerdict::Complete` comes from `account_subscription_eose`, where each subscription counts as complete after `subscription_any_eose`; production routing also uses `required_acks: 1`. With two relays, relay A can EOSE an empty replay while relay B is slow or disconnected and is the only relay holding the omitted event. This branch then clears the durable marker without receiving that event, so a later floored rebuild can strand it permanently again. Please require replay evidence from the relay set capable of carrying the omitted history, or conservatively keep recovery incomplete, and add a two-relay regression with fast-empty A and delayed-history B.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T11:02:17Z` |
| Commit reviewed | `c8276c369ad2b5cedf89d3e7185c6736cd3b2e8c` |
| Model | `Not exposed by runtime` |

The durable marker, process-local cursor fence, bounded marker worker, restart path, and generation-safe clear address most of #1561 well. One P1 remains: overflow repair uses the ordinary any-relay EOSE gate, so it can clear before a relay that holds omitted history has served it. This is a comment review because GitHub does not permit requesting changes on an author-owned PR.

The PR is also currently conflicting with `master`; rebase and rerun the relevant multi-relay recovery tests after resolving the correctness issue.

Datawav (commented):
## Automated review

Actionable findings: 1.

1. P1 / Blocking — Overflow recovery can clear after only one relay reaches EOSE
   - Exact changed-line evidence: `crates/marmot-app/src/client/sync.rs:1259` handles `DrainVerdict::Complete` by finishing overflow recovery and clearing the durable account-delivery recovery marker. The changed recovery path derives completion from `account_subscription_eose`, whose bounded context uses `subscription_any_eose`; production routing requires only one acknowledgement.
   - Triggering state: An unfloored recovery spans two relays. Relay A returns EOSE quickly without the omitted event, while slower or disconnected relay B is the only relay retaining that older event.
   - Consequence: Recovery promotes the newer checkpoint and deletes the durable marker before relay B serves the omitted event. Later subscription rebuilds may apply the promoted cursor floor and permanently exclude that event, recreating the silent history-loss condition this PR is intended to fix.
   - Confidence: High. The exact-head patch, bounded source context, and current-head discussion independently expose the same any-relay completion path; the existing tests do not exercise fast-empty A with delayed-history B.
   - Smallest sound correction: Give overflow recovery a stricter completion policy than ordinary synchronization: retain the marker until every relay capable of carrying the replay range confirms EOSE for that recovery attempt, or conservatively leave recovery pending when full confirmation is unavailable. Add a two-relay regression where A emits empty EOSE first and B later supplies the omitted older event; assert the marker remains until that event is ingested.

Audit evidence:
- Exact head: `5b440ee2afdeaf9be16876a7e9b68200ae6072ed`
- Base: `b0214579f87ee8cc7636b0913f606feaa8a00b33`
- Changed files examined include `crates/marmot-app/src/client/sync.rs`, `crates/marmot-app/src/relay_plane/mod.rs`, `crates/marmot-app/src/runtime/account_worker.rs`, `crates/marmot-app/src/tests.rs`, and `crates/storage-sqlite/src/migrations/0053_account_delivery_recovery.rs`.
- Patch coverage is marked complete with no missing patch files.
- Discussion classification: The current any-relay EOSE report is supported and consolidated into the finding above. Earlier reports about the cursor/marker crash window, per-drop task and payload retention, recovery-triggered session teardown, storage closure, notification ordering, epoch-stall draining, and migration-number collision are fixed, stale, duplicate, or unsupported on this exact head. Constructor documentation and helper-extraction comments are non-actionable maintainability suggestions.

Risk lenses checked:
- concurrency/lifecycle: cancellation and stale callbacks; ordering and race windows; process or lifecycle restart.
- network/protocol: timeouts, retry, and duplicate delivery; partial or malformed peer data; version and wire compatibility.
- tests: assertions fail before the fix; boundary and negative cases; tests exercise production behavior.
- Large-change decomposition and behavioral-test reasoning were also applied.

Check evidence:
- All reported checks completed, but the bundle marks the overall check set unsuccessful because `Relay runtime tests` failed.
- Successful checks include all four Rust test shards, Rust check, clippy, format, audit, Relay CLI E2E, CLI E2E with mock relay, Tamarin proofs, and Simulator and vectors.

crates/marmot-app/src/client/sync.rs
erskingardner: Rebased onto current master (including #1581). The overflow path still clears its durable marker only for `DrainVerdict::Complete`, and `AccountSubscriptionEose::complete()` now requires EOSE from every endpoint-scoped attempt in the frozen activation snapshot (`relay_subscription_attempts_with_eose == relay_subscription_attempts`). A fast empty relay therefore cannot stand in for a slow relay holding history, and a route shrink does not erase the obligation. The focused `account_subscription_eose_requires_the_frozen_relay_coverage` regression and the app-level slow-relay recovery regression both pass, as do the overflow recovery tests. This concern is handled by the overlapping master fix, so I preserved that implementation instead of adding a duplicate gate.

crates/storage-sqlite/src/migrations.rs
```diff
@@ -377,6 +379,11 @@ const MIGRATIONS: &[Migration] = &[
         name: "0052_epoch_backfill_intents",
         apply: migration_0052_epoch_backfill_intents::apply,
     },
+    Migration {
+        version: 53,
+        name: "0053_account_delivery_recovery",
+        apply: migration_0053_account_delivery_recovery::apply,
```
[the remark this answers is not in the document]
erskingardner: Rebasing removed the migration-identity replacement that this thread identified. Current master keeps version 52 as `0052_epoch_backfill_intents`; this PR now appends overflow recovery as version 53, `0053_account_delivery_recovery`. The former version-52 delivery migration existed only on this unmerged PR history: its commit is not an ancestor of master and no tag contains it, so no released canonical database can have recorded that superseded name. The full `storage-sqlite` migration suite passes on the rebased tree.

crates/marmot-app/src/relay_plane/mod.rs
```diff
@@ -122,7 +124,370 @@ pub struct MarmotRelayPlaneAccountAdapter {
...
+            let mut state = self
+                .inner
+                .lock()
+                .unwrap_or_else(|poisoned| poisoned.into_inner());
+            if state.generation != generation {
+                state.marker_in_progress = false;
+                return;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 34221204. The success and storage-closed branches now clear/set marker state only when the worker still owns its captured generation. I applied the same ownership rule to the retry loop tail as well: a stale retry worker returns without clearing the newer generation claim. The new `stale_overflow_marker_workers_preserve_new_generation_claim` regression covers successful, closed, and retryable stale-worker exits and passes.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -4343,6 +4370,40 @@ async fn run_pending_epoch_backfill_reporting_arm(
...
+                    account_label,
+                    message.clone(),
+                );
+                Err(AccountCatchUpFailure::new(
+                    message,
+                    failure.classification(),
+                ))
+            }
+        };
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 34221204 with a typed `DeliveryOverflowRecoveryOutcome`. EOSE/quantum incompleteness now returns `Incomplete(summary)`: callers publish the real summary, retain the durable marker, and keep catch-up successful. Activation, group-sync, ingest, generation-race, and marker-clear errors remain `ClassifiedSyncFailure` and still produce `AccountCatchUpFailure`. The new `incomplete_delivery_overflow_recovery_does_not_fail_catch_up` regression verifies the marker remains durable while explicit catch-up succeeds.

erskingardner (commented):
**Recommend merge** (comment review — GitHub rejects approvals on your own PR).

Re-reviewed `14dc4f0` against #1561 and the earlier blocking notes. Those are closed.

**Previous blockers, now fixed**
- Durable cursor is no longer `state.last_transport_timestamp`. `save_state` writes `checkpointed_transport_timestamp`, and only a completed drain promotes that floor when `pending_delivery_overflow()` is clear. Live `ingest_received_delivery` can no longer max-merge a newest-first prefix ahead of the marker. `process_local_overflow_fence_freezes_cursor_while_marker_write_retries` fails if that window reopens.
- One marker worker per generation; omitted payloads are dropped at the router. The unbounded hold/spawn is gone.
- Reopen recovery now injects the older omitted event and asserts it is ingested. `since_floor.rs` was updated to pin the two-stage cursor on purpose, not as a leftover characterization of the old defect.
- All-relay EOSE (`relay_subscription_attempts_with_eose == relay_subscription_attempts`) is the recovery completion gate; incomplete recovery is `Incomplete(summary)` and does not tear down the session. Migration is 0053 on current master.

**Premise** still holds. The extra machinery (generation token, compare-and-clear, process-local fence vs durable marker) is paying for a real crash-safe gap, not speculative flexibility.

CI is green. No new blocking issues on this head.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1231,9 +1232,39 @@ async fn run_app_runtime_account_worker(
...
+                                    &events,
+                                    &account_id_hex,
+                                    &account_label,
+                                    "account delivery overflow recovery incomplete".to_owned(),
+                                );
+                                // The durable marker remains armed. Keep the
+                                // account session available and let the next
+                                // receive/catch-up seam retry the replay.
+                                (Ok(summary), true)
+                            }
```
erskingardner, line 1255: Non-blocking: an `Incomplete` recovery publishes an account-visible error string here, while the epoch-backfill seam publishes only a summary. Incomplete is the expected retry shape while a relay simply has not EOSE-d yet, so a fresh app-level error per retry may look like a false alarm in host UIs. Consider classifying it as informational retry progress, or folding it into a summary-only publish.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T20:19:06Z` |
| Commit reviewed | `14dc4f0d98af7cdbe5ba43e670fca9fd4f28727d` |

## Summary

Sound fix for #1561's silent-history-skip vector. I reviewed against the earlier blocking thread and the current head; the previously raised P1s are genuinely closed here - not deferred:

- **Cursor/EOSE trust.** The durable saved floor is now `checkpointed_transport_timestamp`; `save_state` persists only the last completed-drain checkpoint, and promotion gates on `pending_delivery_overflow().is_none()`. Live `ingest_received_delivery` can no longer max-merge a newest-first prefix ahead of the marker. The fence test pins the pre-marker crash window and the `-120s` lookback tolerance makes the residual race harmless.
- **EOSE completeness.** `AccountSubscriptionEose::complete()` is the strict endpoint-scoped gate (every relay attempt must report EOSE), already severe on master; the recovery overlay only forces `with_eose = 0` while an unresolved generation exists. "Fast empty relay clears early" is not reachable at this head.
- **Generation guard.** Compare-and-clear on a random marker token plus `dropped` count compare: a re-overflow during recovery invalidates the attempt instead of clearing it. Storage race (`Ok(false)`) preserves the newer generation.
- **One worker per generation.** `start_marker_persistence` plus generation-checked finish blocks the stale-worker hazard; the two `stale_overflow_marker_workers_preserve_new_generation_claim` test harnesses pin exactly the races CodeRabbit flagged.
- **Cross-account isolation.** Verified in `account_queue_overflow_invalidates_eose_without_blocking_other_accounts`: Bob receives and his EOSE completes while Alice's queue is full.
- **Restart.** Durable marker in migration 0053 + synthetic pending generation on reopen; the SQLite test pins survive/clear semantics.
- **Privacy.** Logs/metrics carry only aggregate queue depth, dropped count, retry outcome, and elapsed time - compliant with the observability doc. No account/group/relay identifiers in the new telemetry.

The `account_delivery_recovery` table FK is genuinely enforced (`PRAGMA foreign_keys` is ON in this crate's connection setup), and the stray `account_adapter()` (no marker) surface remains test-only - production open goes through `account_adapter_with_recovery_marker`.

One non-blocking note left inline on `account_worker.rs` (Incomplete published as an app-level error string; the epoch-backfill seam treats the same outcome as plain progress - pick one posture).

Scope is large (~+1.8k lines), but every acceptance criterion in #1561 maps to a specific part of it, and CI is fully green. Approve.

PR #1571, opened 2026-08-26
fix(hermes): preserve shed inbound replay

Datawav:
## Summary

- reserve inbound message ids while queue admission or debounce is pending without terminally deduplicating them
- commit ids to the recent-message cache only after the per-group queue accepts the turn
- release direct and debounced source ids on queue shedding, admission exceptions, and disconnect cleanup so connector/storage replay can recover them
- cover direct replay, multi-source debounced replay, and duplicate suppression while admission is pending

## Verification

- `python3 -m unittest discover -s integrations/hermes/marmot/tests` (202 passed)
- `just fast-ci`
- the new direct regression failed against the pre-fix behavior because a shed id remained terminally deduplicated
- independent review: PASS with no findings (Cursor Grok 4.6)

## Risk

The change is limited to the Hermes Marmot adapter's in-memory inbound dedupe and queue-admission bookkeeping. It does not add adapter-owned replay; recovery remains connector/storage-backed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inbound message deduplication during concurrent processing.
  * Messages dropped due to temporary queue capacity limits can now be replayed successfully.
  * Debounced message batches correctly release tracking state when admission fails, preventing messages from being lost.

* **Tests**
  * Added coverage for replaying shed messages and coalesced batches after queue recovery.

### DIFF
--- a/integrations/hermes/marmot/adapter.py
+++ b/integrations/hermes/marmot/adapter.py
@@ -1899,6 +1899,9 @@ class MarmotPlatformAdapter(BasePlatformAdapter):
         # message (rapid catch-up after subscribe, or across a reconnect); drop
         # ids already seen so the same user message is not dispatched twice.
         self._recent_inbound_ids = _RecentKeys(DEFAULT_INBOUND_DEDUPE_WINDOW)
+        # Reserve ids while admission is being decided. A reservation prevents
+        # concurrent duplicates without making a shed message terminally seen.
+        self._pending_inbound_ids: set[str] = set()
         # Dedupe repeated ambient surfacings (deletion / group-state change) by a
         # context key (mirror inbound.ts contextKeys).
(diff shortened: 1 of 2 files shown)


Datawav (commented):
No actionable findings.

erskingardner (approved):
Recommend merge.

The bug is real: recording an inbound id before `KeyedAsyncQueue.enqueue` made a capacity-shed turn terminally seen, so connector/storage replay could not recover it. Splitting reservation (`_pending_inbound_ids`) from commit (`_recent_inbound_ids`) is the smallest correct fix. Check-then-add is synchronous so asyncio cannot race two reservations for the same id; `enqueue` returning `None` on shed is honored; debounce commits every source id only after the coalesced batch is admitted; cancel/disconnect clears reservations without marking them seen.

The two new tests prove the pre-fix failure (shed id stays out of `_recent_inbound_ids`) and the multi-source debounce release/replay path. CI is green on this SHA.

PR #1574, opened 2026-08-27
feat(agents): deliver backend artifacts through SendMedia

### LINKED ISSUE #1527: terminal harness: carry backend artifacts through retry-safe SendMedia
## Problem

MDK’s agent-control protocol now supports idempotent `SendMedia`, but the shared terminal harness cannot express an outbound artifact. Its backend event type contains only completed text, its control client exposes only `send_final`, and every backend text event is sent through that final-text path. Local product behavior demonstrates that a `wn-codex` turn can create a file on disk, but the pinned terminal contract cannot express that file as a Marmot media attachment.

Source-confirmed at `9c97dad5ecc02a5eb5da7b0ca5a8cf0760720f50`:

- `RunnerEvent` has only `Text(String)` and `Backend::run` streams only that type ([contract](https://github.com/marmot-protocol/mdk/blob/9c97dad5ecc02a5eb5da7b0ca5a8cf0760720f50/integrations/terminal-harness/src/lib.rs#L153-L184)).
- The bridge maps every event through `send_final` ([collector](https://github.com/marmot-protocol/mdk/blob/9c97dad5ecc02a5eb5da7b0ca5a8cf0760720f50/integrations/terminal-harness/src/bridge.rs#L408-L445)).
- The terminal control client implements `SendFinal` but no `SendMedia` call ([client](https://github.com/marmot-protocol/mdk/blob/9c97dad5ecc02a5eb5da7b0ca5a8cf0760720f50/integrations/terminal-harness/src/control.rs#L101-L136)).
- Agent control already defines idempotent `SendMedia` with `attachments: Vec<AgentControlMediaUpload>`, so one request is already an ordered multi-file batch; this is a terminal integration gap rather than a new transport primitive ([protocol](https://github.com/marmot-protocol/mdk/blob/9c97dad5ecc02a5eb5da7b0ca5a8cf0760720f50/crates/agent-control/src/lib.rs#L283-L299)).

## Expected behavior

- Define a backend-only `ArtifactDeclaration` event/turn result containing an ordered set of relative paths plus explicit filename, MIME type, and an opaque scoped `authorization_id`; no arbitrary metadata keys are accepted. Assistant prose and path-like text are never parsed as declarations. The terminal completed assistant text, if present, is the sole caption source. One terminal artifact outcome maps to one ordered batch `SendMedia` call, not one call per file.
- Artifact export is disabled by default. A host/package administrator may enable it only through connector configuration `terminal_harness.artifact_exports.enabled: true` plus a grant entry containing exact `group_id`, `export_root`, and optional `ttl_seconds` (default 3600, bounded 1–86400). No per-user or backend-authored grant exists; unset, malformed, expired, or unmatched group configuration fails closed. On backend turn creation, the connector validates that grant, mints an opaque random `authorization_id` in connector-owned state, binds it to the Marmot group, backend turn, configured export root, and expiry, then injects it only through the dedicated `WN_ARTIFACT_AUTHORIZATION_ID` environment variable for that turn. The backend can copy that value into `ArtifactDeclaration` but cannot mint or expand it; backend-invented values have no connector record and fail closed.
- Resolve every relative declaration beneath the authorization's connector-scoped export root, reject traversal, require a regular file with no symlink component, then copy to restrictive mode-0600 staging. Set `TERMINAL_ARTIFACT_MAX_COUNT` to default 10 with an operator-configurable positive bound of 1–10. Enforce the per-attachment negotiated maximum accepted by the #1355 transport contract; do not introduce a parallel byte-size cap or negotiation policy.
- Call `SendMedia` with a stable idempotency key derived from the inbound message/turn and artifact batch. Reconcile ambiguous disconnects without duplicate uploads/messages.
- Use an all-or-nothing preflight: validate, authorize, fingerprint, and stage every artifact before uploading any of them. If any artifact fails, upload nothing and emit one existing-class agent-activity error without leaking host paths; apply #1339 attention only when the user must act. For artifact-only failure, call neither `SendMedia` nor `SendFinal`. For mixed text/artifact failure, preserve the completed assistant text by sending it exactly once through the ordinary text-only `SendFinal` path after the activity error; the error states that attachments were withheld. Neither path invents a message kind, silently drops an artifact, or makes a delivery claim.
- This issue is independently implementable against the current contract: the current completed assistant text is the caption for a mixed text/artifact terminal result, which sends one `SendMedia` batch and never also calls `SendFinal`. Draft 1 is a non-blocking related draft whose later sealed-terminal projection can replace that caption source without changing the artifact protocol.
- An artifact-only terminal result sends one `SendMedia` batch with no caption and no `SendFinal`, using the same preflight and idempotency rules.
- Inbound files remain the separate shared-harness work in #1400. Size negotiation remains #1355.

## Acceptance tests

- A fake backend emits two ordered artifacts plus completed assistant text; the control fixture observes one batch `SendMedia` with exact ordered metadata, the text as caption, and a stable idempotency key, with no duplicate `SendFinal`.
- A valid backend `ArtifactDeclaration` for a file beneath the authorized workspace export root sends that file; merely creating the same file without the declaration sends nothing, and prose containing its path sends nothing.
- With export unset or disabled, no authorization variable is injected and every declaration fails before staging. A host-admin configuration for the exact group/root with default 3600-second TTL causes the connector to mint/store one turn-scoped id and inject the same opaque value only through `WN_ARTIFACT_AUTHORIZATION_ID`; wrong-group config and TTL outside 1–86400 fail closed. A declaration carrying the valid id validates; a backend-invented id fails before staging, and declarations with metadata beyond filename/MIME are rejected with a privacy-safe code.
- When the declaration and completed assistant text are both present, that assistant text is the exact `SendMedia` caption; the declaration has no caption field and cannot override it.
- A declaration batch with no completed assistant text sends one `SendMedia` batch with an absent caption, calls no `SendFinal`, and retains the same durable batch idempotency key across reconciliation.
- Replaying after an ambiguous response reconciles the one durable artifact-batch idempotency key, returns the original durable message ids, and creates no duplicate upload or Marmot message.
- Outside-root, symlink, missing, count 11 at the default cap of 10, an attachment above the #1355-negotiated maximum, and a TOCTOU mismatch between the preflight fingerprint and mode-0600 staged copy fail closed with stable privacy-safe codes. Tests also assert every successfully staged file is mode 0600.
- Missing, expired, wrong-group, wrong-turn, or wrong-root authorization fails before staging or `SendMedia` and uploads nothing. A backend declaration without a valid grant-minted `authorization_id` is not authority.
- In a two-artifact batch where the second item fails policy or authorization, preflight uploads neither item and emits one existing agent-activity error without any host path, stating that attachments were withheld. Artifact-only failure calls neither `SendMedia` nor `SendFinal`; mixed failure then sends the completed assistant text exactly once through text-only `SendFinal`. It applies attention only if user action is required, invents no message kind, and makes no attachment-delivery claim.
- Text-only completed assistant messages continue through the current `SendFinal` path (and Draft 1's sealed projection when available). A mixed text/artifact terminal turn produces one coherent captioned media result, no duplicate text final, and no bare host path in chat.
- Connector restart between staging and acknowledgement reconciles the same durable batch idempotency key and produces exactly one Marmot media message; it neither invents a new key nor reruns the backend turn.

## Relationships

Standalone and non-blockingly related to Draft 1's improved terminal projection and Draft 2's text-only fallback/reconciliation boundary. Consumes the retry-safe media foundation completed by #1322. Inbound batches are #1400; negotiated blob size is #1355. Artifact-only and successfully sent mixed captioned-media acknowledgement/reconciliation are exclusively owned here; Draft 2's lost-ack state applies only to the fallback text-only `SendFinal` after mixed preflight failure, never to a `SendMedia` attempt. Preflight failure projection uses the existing notification/attention semantics in #1087, #1339, #1392, and #1445.

Datawav:
## Summary
- Carries ordered backend artifacts through the shared terminal-harness `SendMedia` path. Codex is the only backend that emits typed `CompletionFile` artifacts; Pi and OpenCode accept the new invocation field, discard `artifact_output`, and fail closed if exports are enabled for them.
- Avoids exposing local absolute paths in the backend-facing turn: Codex sees a workdir-relative export location when possible, otherwise `$WN_ARTIFACT_EXPORT_ROOT` / `$WN_ARTIFACT_MANIFEST`.
- Validates configured roots, regular-file and non-symlink ownership, filename and MIME metadata, per-file and per-turn limits, and staged byte totals before any delivery side effect.
- Preserves all-or-nothing preflight, ordered mixed text/media sequencing, retry safety, compensation, and terminal retry intent when delivery or cleanup fails.
- Flushes completed assistant text through the ordinary `SendFinal` path before a generic failure reply, including text-only turns that buffered because a grant was live.
- Adds capability mapping, configuration parsing, focused artifact pipeline coverage, and connector retry protocol tests.

## Verification
- `cargo test -p wn-codex`
- `cargo test -p wn-opencode`
- `cargo test -p wn-pi`
- `cargo clippy -p marmot-terminal-harness -p wn-codex -p wn-opencode -p wn-pi --all-targets -- -D warnings`

Closes #1527

## Summary by CodeRabbit

- **New Features**
  - Added secure artifact export from supported backends, including encrypted media delivery, authorization, validation, staging, retries, and recovery.
  - Added configurable export grants, staging locations, artifact limits, and environment-based settings.
  - Added idempotency support for media uploads and stream operations.
  - Added artifact status and failure events.

- **Bug Fixes**
  - Improved handling of unreadable artifact manifests without failing successful text runs.
  - Clarified timeout behavior and liveness reporting.

- **Documentation**
  - Documented Codex command interception, artifact export, privacy safeguards, and timeout behavior.


integrations/terminal-harness/src/bridge.rs
```diff
@@ -398,65 +455,168 @@ async fn handle_message(ctx: Arc<BridgeContext>, inbound: InboundPrompt, permit:
...
-                            target: TRACE_TARGET,
-                            method = "send_final",
-                            error_kind = err.privacy_safe_kind(),
-                            "failed to send backend reply chunk"
-                        );
-                        delivery_failed = true;
-                        break;
-                    } else {
+                if buffer_text_for_artifacts {
+                    buffered_text.push(text);
```
erskingardner, line 806: Blocking: when a grant is live, this buffer holds every `RunnerEvent::Text` until `Ok(Ok(outcome))`. The `RunFailure` and join-error arms below never flush it, so a granted group that times out, goes idle, or crashes after emitting text loses that text and only the generic failure reply is sent.

That is a regression from the current live `SendFinal` path, and it hits text-only turns too — buffering is gated on `artifact_output.is_some()`, not on whether artifacts were actually declared.

Flush `buffered_text` through the ordinary text-only `SendFinal` path before the failure reply (same shape as the mixed preflight-fallback arm). Add a recording-backend test: grant live, emit text, then `RunFailure`, and assert the completed text is delivered exactly once and no `SendMedia` is attempted.

integrations/codex/marmot/src/codex.rs
```diff
@@ -60,24 +65,61 @@ async fn run_with_bin(
...
     } = invocation;
-    run_jsonl_process(
+    let mut environment = Vec::new();
+    if let Some(request) = &artifact_output {
+        environment.push(EnvironmentChange::Set {
+            name: "WN_ARTIFACT_AUTHORIZATION_ID",
+            value: request.authorization_id().to_owned(),
+        });
+        prompt.push_str(&format!(
+            "\n\nIf this task produces files for the requester, place them beneath {} and write exactly one JSON object to {} using this schema: {{\"artifacts\":[{{\"authorization_id\":\"value from $WN_ARTIFACT_AUTHORIZATION_ID\",\"path\":\"relative/path\",\"media_type\":\"application/octet-stream\",\"file_name\":\"name.ext\"}}]}}. Paths must be relative to the export root. Use an empty artifacts array when there are no files. This structured file is the only artifact-delivery signal; do not rely on mentioning paths in chat text.",
```
erskingardner: Non-blocking: this appends the absolute export root and connector-owned manifest path to the prompt. If Codex echoes either path in assistant text, it becomes the `SendMedia` caption or a `SendFinal` chunk and violates #1527’s “no bare host path in chat” rule. The “do not mention paths” sentence is hope, not enforcement.

Prefer a workdir-relative export location (or env-only coordinates) so the model never has to see `~/.marmot-agents/...` or the private manifest path. The authorization id itself is correctly confined to `WN_ARTIFACT_AUTHORIZATION_ID`.

erskingardner (changes requested):
Request changes. The shared staging/SendMedia pipeline is the right shape for #1527 — fail-closed grants, typed `ArtifactDeclaration` only, all-or-nothing preflight, durable outbox + stable idempotency, mixed fallback to one `SendFinal` after an activity error, and no invented message kinds. Codex is the only backend that can actually emit artifacts; Pi/OpenCode correctly discard `artifact_output` and `validate_artifact_support` fail-closes if exports are enabled there. The PR title’s “for Codex, Pi, and OpenCode” overstates that. Env-var config (`WN_*_ARTIFACT_EXPORTS_ENABLED` / `ARTIFACT_GRANTS_JSON`) is the right adaptation of the issue’s YAML wording; this harness is env-configured throughout.

Blocking: when a grant is live, every turn buffers assistant text so it can become a `SendMedia` caption. The `RunFailure` / join-error arms then drop that buffer and send only the generic failure line. That is a regression versus today’s live `SendFinal` path and contradicts #1527’s rule that completed assistant text survives mixed failure. This happens on any timeout/crash/idle for a granted group, including text-only turns that never emit artifacts.

Non-blocking: long captions are split and the overflow is sent as `SendFinal` after media (covered by test). #1527 asked for one `SendMedia` and no second final; the 30k caption budget makes overflow understandable, but it is a spec deviation. Codex also injects absolute export-root and manifest paths into the prompt, which can leak into chat text despite the “do not mention paths” instruction.

erskingardner:
@Datawav please stop working on this PR.

integrations/terminal-harness/src/bridge.rs
```diff
@@ -807,6 +989,150 @@ async fn handle_message(ctx: Arc<BridgeContext>, inbound: InboundPrompt, mut per
...
+                artifact_count = outputs.len(),
+                "artifact delivery remains pending"
+            );
+            let reason = if err.artifact_validation_failed() {
+                "was rejected because the staged bytes no longer match durable delivery state"
+            } else {
+                "is pending automatic retry after a delivery failure"
+            };
+            let _ = send_artifact_failure_activity(ctx, inbound, outputs.len(), reason).await;
+            ArtifactDeliveryOutcome::Pending
```
erskingardner: **[P1] Keep pending media in the per-group FIFO barrier**

A retryable `send_media` failure leaves this batch in the outbox and returns `Pending`, but `handle_message` then releases the group serial permit. `fifo_is_blocked` only consults recovery and text-final delivery state, so the next prompt in this group can run and publish its final before the 30-second replay loop sends this older media result. That breaks the harness FIFO and the linked restart/reconciliation ordering contract. Make a pending artifact batch block its group (with a wakeup when completion/discard succeeds), and add a two-prompt regression proving the second turn cannot start while the first media batch is pending.

integrations/terminal-harness/src/artifacts.rs
```diff
@@ -0,0 +1,1568 @@
...
+        let parent = self
+            .path
+            .parent()
+            .ok_or_else(|| HarnessError::Config("artifact outbox path has no parent".to_owned()))?;
+        fs_private::create_dir_all_private(parent)?;
+        let persisted = PersistedOutbox {
+            mac: outbox_mac(&self.key, batches)?,
+            batches: batches.to_vec(),
+        };
+        let bytes = serde_json::to_vec(&persisted)?;
```
erskingardner, line 407: **[P1] Never persist an outbox that the loader will reject**

The reader caps this file at 1 MiB, but this writer never checks the serialized length. One large `remaining_text` value—or enough pending batches—can therefore be written successfully and then rejected by `ArtifactOutbox::load` after restart; startup disables exports for the run, so the supposedly durable media intents are no longer retried. Enforce the same bound before rename and define a bounded overflow strategy (or store overflow text in the existing chunk delivery state), with a restart test at the boundary.

integrations/terminal-harness/src/artifacts.rs
```diff
@@ -0,0 +1,1568 @@
...
+        || now > authorization.expires_at_unix_secs
+    {
+        return Err(HarnessError::ArtifactAuthorizationInvalid);
+    }
+    Ok(())
+}
+
+fn open_authorized_source(relative_path: &Path, root: &Path) -> Result<File> {
+    let path = root.join(relative_path);
+    let metadata = std::fs::symlink_metadata(&path)?;
```
erskingardner: **[P1] Reject symlinks in every source-path component**

This checks only the final component: `symlink_metadata(path)` and `O_NOFOLLOW` do not reject an intermediate symlink. A declaration such as `subdir/link/report.pdf`, where `link` resolves elsewhere inside the authorized root, is accepted even though #1527 explicitly requires no symlink component; on non-Linux targets the fallback also canonicalizes the pathname rather than the opened descriptor. Resolve/open the relative path component-by-component without following symlinks (for example with directory FDs), and add a nested-symlink regression.

integrations/codex/marmot/src/codex.rs
```diff
@@ -60,24 +68,124 @@ async fn run_with_bin(
...
+            Ok(artifacts) if !artifacts.is_empty() => {
+                tx.send(RunnerEvent::Artifacts(artifacts))
+                    .await
+                    .map_err(|_| RunFailure {
+                        error: HarnessError::BackendStream,
+                        observed_session: outcome.observed_session.clone(),
+                    })?;
+            }
+            Ok(_) => {}
+            Err(_) => {
```
erskingardner, line 118: **[P2] Project malformed declarations as attachment failures**

A malformed, oversized, or unknown-field completion manifest reaches this branch after Codex has been explicitly instructed to declare artifacts, but the bridge receives neither artifacts nor an artifact-error event. It therefore treats the turn as text-only and never emits the required “attachments withheld” activity, silently dropping the attempted artifact outcome. Preserve text fallback, but carry a typed declaration failure to the bridge so it emits one privacy-safe activity error; a local warning alone is not observable to the requester.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T18:55:57Z` |
| Commit reviewed | `5eff147c34920913c50fd6cf4ee4be1bf3cb717d` |
| Model | `gpt-5.6-sol` |
| Reasoning level | `high` |

Request changes. The earlier buffered-text failure regression is fixed and CI is green, but the current durable media path still has ordering, recoverability, and source-confinement holes. A malformed completion manifest also bypasses the required attachment-withheld projection.

integrations/terminal-harness/src/bridge.rs
```diff
@@ -856,6 +1051,150 @@ async fn handle_message(ctx: Arc<BridgeContext>, inbound: InboundPrompt, mut per
...
```
erskingardner: **Blocking: discarding a batch after a post-persistence validation failure still returns `Pending`, permanently losing completed assistant text.**

By the time we reach this `Err` branch, the completed assistant text for a mixed turn has already been staged into `batch.caption` / `batch.remaining_text`. If `send_pending_artifact_batch` fails with `err.artifact_validation_failed()` (a staged copy fails re-validation, outbox/key corruption, or an idempotency fingerprint mismatch), the batch is discarded — the durable record and staged files are gone — yet the outcome reported is `ArtifactDeliveryOutcome::Pending`. `handle_message` only flushes `buffered_text` on `FallbackAllowed`, so the requester receives only the activity error; the completed text is never delivered and the 30s replay loop cannot recover because the batch was deleted. This breaks the mixed-failure preservation rule from #1527 and can silently lose user-visible output.

Return `FallbackAllowed` from the discard branch (like the pre-record failure path just above) so the buffered text flushes through the ordinary text-only `SendFinal` path.

crates/agent-control/src/lib.rs
```diff
@@ -7,6 +7,9 @@ use tokio::io::{AsyncBufRead, AsyncBufReadExt, AsyncReadExt, AsyncWrite, AsyncWr
 
 pub const AGENT_CONTROL_PROTOCOL_V2: &str = "marmot.agent-control.v2";
 pub const MAX_AGENT_CONTROL_FRAME_BYTES: usize = 1024 * 1024;
+pub const MAX_MEDIA_UPLOAD_ATTACHMENTS: usize = 10;
+pub const MAX_MEDIA_UPLOAD_ATTACHMENT_BYTES: u64 = 512 * 1024 * 1024 - 16;
+pub const MAX_MEDIA_UPLOAD_BATCH_BYTES: u64 = MAX_MEDIA_UPLOAD_ATTACHMENT_BYTES;
```
erskingardner, line 12: Non-blocking: these media limits used to be *derived* from `marmot_app::MAX_ENCRYPTED_MEDIA_BLOB_BYTES - 16` in `agent-connector`, and are now hard-coded literals in `agent-control`. The values match today, but the invariant is no longer enforced: if the app-layer blob limit ever moves, the harness pre-validation and the connector/app enforcement drift apart (uploads pass here, fail there). Since `agent-control` intentionally cannot depend on `marmot-app`, have the connector keep a compile-time/test assertion that `MAX_MEDIA_UPLOAD_ATTACHMENT_BYTES == marmot_app::MAX_ENCRYPTED_MEDIA_BLOB_BYTES - 16`, so drift fails a test instead of a production send.

integrations/terminal-harness/src/artifacts.rs
```diff
@@ -0,0 +1,1689 @@
...
+        || message_ref != authorization.message_ref
+        || now > authorization.expires_at_unix_secs
+    {
+        return Err(HarnessError::ArtifactAuthorizationInvalid);
+    }
+    Ok(())
+}
+
+#[cfg(unix)]
+fn open_authorized_source(relative_path: &Path, root: &Path) -> Result<File> {
```
erskingardner, line 1027: Non-blocking hardening note: the new `openat` walk (and the per-component `symlink_metadata` fallback on non-unix) now closes the intermediate-symlink hole, but the *source* file still gets no ownership/permission validation — only the staged copy and outbox/key files get `validate_private_single_link`. If an operator points `export_root` at a group- or world-writable directory, another local user can plant files in it that an allowed group member can then exfiltrate into chat. Consider rejecting (or at least warning on) export roots with group/world write bits at `authorize` time, and optionally requiring the source to be owned by the connector euid, matching the staged-copy check.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | 2026-08-27T19:37:10Z |
| Commit reviewed | a7d1d34a34db6ddab8d542ef0fb4056bacb9316a |
| Model | openrouter/moonshotai/kimi-k3 |

Security-focused pass at the latest head. The design is solid where it matters: fail-closed grants, connector-minted per-turn authorization ids bound to group+message+root+TTL, typed-only `ArtifactDeclaration` (no prose parsing), the new `openat`/per-component walk closing intermediate-symlink traversal, mode-0600 staged copies re-validated with digest+single-link checks before every send, and a MAC-protected outbox with all-or-nothing preflight. The `RunnerEvent::ArtifactDeclarationFailed` event in the latest commit also resolves the earlier malformed-manifest bypass properly, and pending media correctly FIFO-blocks the group until resolved.

Blocking (inline on bridge.rs): when a batch fails post-persistence validation, the code discards it but still reports `Pending`. `handle_message` only flushes completed assistant text on `FallbackAllowed`, so the chat gets only the activity error and the text is unrecoverable (no durable record remains). Return `FallbackAllowed` on the discard path like the pre-record failure branches do.

Non-blocking notes posted inline: the media-limit constants are now hard-coded literals decoupled from `marmot_app::MAX_ENCRYPTED_MEDIA_BLOB_BYTES` (add a drift assertion test), and export sources get no ownership/permission validation (plantable by other local users if `export_root` is group/world-writable). One additional message-integrity nit: `send_artifact_failure_activity` is invoked with a hard-coded artifact count of 1 on connector-setup failure, so a text-only turn claims an attachment that never existed was withheld.

integrations/terminal-harness/src/artifacts.rs
erskingardner: Addressed on the current head. The source path is opened component-by-component with openat plus O_NOFOLLOW/O_DIRECTORY on Unix, with per-component checks in the fallback, and the nested intermediate-symlink regression passes.

integrations/terminal-harness/src/bridge.rs
erskingardner: Fixed in 60ad6e16. A permanent staged-validation failure now durably discards the invalid batch and returns FallbackAllowed so buffered assistant text is delivered. If the discard itself fails, the outcome remains Pending to avoid a later media/text duplicate. The policy has a regression test.

crates/agent-control/src/lib.rs
erskingardner: The connector already enforces this invariant in tests::connector_media_limits_follow_the_marmot_app_blob_limit, asserting attachment bytes plus 16 equals MAX_ENCRYPTED_MEDIA_BLOB_BYTES and the batch limit matches the attachment limit. I ran that focused test on this head and it passes, so no code change is needed.

integrations/terminal-harness/src/artifacts.rs
erskingardner: Hardened in 60ad6e16. Authorization now rejects export roots not owned by the connector euid or writable by group/other users, and the component walk applies the same ownership/write-bit policy to intermediate directories and the final source file. Regression coverage checks root, nested-directory, and file modes.

integrations/terminal-harness/src/artifacts.rs
```diff
@@ -0,0 +1,1689 @@
...
+            });
+        }
+        Ok(())
+    })();
+    if let Err(error) = result {
+        for path in &created_paths {
+            let _ = std::fs::remove_file(path);
+        }
+        return Err(error);
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 60ad6e16. After loading the durable outbox, startup reconciles the private staging directory and removes only regular single-link files that no pending batch references. The regression preserves referenced staged bytes, reclaims the orphan, and confirms the same deterministic delivery can then be staged again.

integrations/terminal-harness/src/artifacts.rs
```diff
@@ -0,0 +1,1689 @@
...
+            let bytes = read_regular_file_limited(&key_path, 32)?;
+            bytes
+                .try_into()
+                .map_err(|_| HarnessError::ArtifactUnsafeSource)
+        }
+        Err(error) => {
+            let _ = std::fs::remove_file(&temp);
+            Err(error.into())
+        }
+    }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 60ad6e16. Key temp names now use an independent random suffix, stale matching temp links are removed before key validation, and a regression simulates interruption after hard_link and verifies the next load restores nlink=1. The same change uses getrandom and the hmac crate with constant-time verification instead of /dev/urandom and bespoke HMAC code.

integrations/terminal-harness/src/artifacts.rs
```diff
@@ -0,0 +1,1889 @@
...
+    for entry in std::fs::read_dir(parent)? {
+        let entry = entry?;
+        let name = entry.file_name();
+        let name = name.to_string_lossy();
+        if name.starts_with(&prefix) && name.ends_with(".tmp") {
+            match std::fs::remove_file(entry.path()) {
+                Ok(()) => {}
+                Err(error) if error.kind() == std::io::ErrorKind::NotFound => {}
+                Err(error) => return Err(error.into()),
+            }
```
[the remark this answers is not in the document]
erskingardner: Fixed in 8931d426. Cleanup now leaves every single-link temp alone, so it cannot delete another live installer temp; it only reclaims a temp whose nlink confirms hard_link already installed the key. The interrupted-post-link recovery regression still passes.

integrations/terminal-harness/src/bridge.rs
```diff
@@ -856,6 +1063,164 @@ async fn handle_message(ctx: Arc<BridgeContext>, inbound: InboundPrompt, mut per
...
+        return;
+    }
+    let pending = ctx.outbox.lock().await.pending();
+    for batch in pending {
+        match send_pending_artifact_batch(ctx, &batch).await {
+            Ok(()) => complete_artifact_batch(ctx, &batch.idempotency_key, &batch.group_ref).await,
+            Err(err) => {
+                if err.artifact_validation_failed() {
+                    let _ =
+                        discard_artifact_batch(ctx, &batch.idempotency_key, &batch.group_ref).await;
```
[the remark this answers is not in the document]
erskingardner: Fixed conservatively in 8931d426. Replay validation only discards an invalid batch when it contains no caption or remaining text. Mixed batches retain their durable text and stay behind the existing FIFO barrier rather than risking ambiguous fallback after a prior SendMedia attempt. A regression covers both mixed retention and artifact-only discard eligibility.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | 2026-08-27T20:35:23Z |
| Commit reviewed | 8931d426d51532764dcf88a8c8d032dcbbc9371a |
| Model | openrouter/moonshotai/kimi-k3 |

Re-review of the delta since `a7d1d34` (commits `60ad6e16` + `8931d426`): all eight threads from my prior review are resolved in code. The pending-vs-discard logic now flushes completed text correctly (`failed_artifact_delivery_outcome` returns `FallbackAllowed` when the discard succeeded), replay discards only artifact-only batches so mixed batches keep their durable text, the source tree now gets euid/mode `0o022` checks at authorize and during the `openat` walk, the `.tmp` leftovers are reclaimed via the proven `nlink >= 2` heuristic, the hand-rolled HMAC is replaced with the `hmac` crate's constant-time `verify_slice`, and randomness comes from `getrandom` instead of a Unix-literal `/dev/urandom`. Startup pruning in `handle_run_start` only removes unreferenced staged files (or unblocks the loader's initial empty pending snapshot).

No new issues found in the delta. CI still in flight; local `cargo check -p marmot-terminal-harness` passes on this head.

integrations/terminal-harness/src/bridge.rs
```diff
@@ -752,11 +867,83 @@ async fn handle_message(ctx: Arc<BridgeContext>, inbound: InboundPrompt, mut per
...
     match runner.await {
         Ok(Ok(outcome)) => {
+            let artifact_delivery_attempted = artifact_setup_failed
+                || artifact_declaration_failed
+                || !artifact_outputs.is_empty();
+            let artifact_delivery = if artifact_setup_failed {
+                let _ = send_artifact_failure_activity(
+                    &ctx,
+                    &inbound,
+                    1,
```
erskingardner, line 884: Minor, non-blocking: this branch fires on every completed turn in a granted group whenever `authorize` or manifest preparation fails — including pure-text turns where no artifact was ever declared (and, since the backend never received a manifest request in this state, none could be). A group whose `export_root` is temporarily missing or fails the ownership/mode checks shows `Attachments withheld. Artifact 1: rejected …` on every reply, which is inaccurate (there is no "Artifact 1") and noisy for users who never asked for files. Consider emitting this activity only when a declaration actually arrived, or at minimum rewording the setup-failure case so it doesn't fabricate a per-artifact line.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T20:40:45Z` |
| Commit reviewed | `8931d426d51532764dcf88a8c8d032dcbbc9371a` |

Fresh pass at the latest head after the four fix rounds, against #1527. The design holds up where it matters, and I re-verified every previously flagged item is actually fixed at this commit: fail-closed grants and TTL bounds, connector-minted per-turn authorization ids bound to group+message+root, typed-only declarations with `deny_unknown_fields`, all-or-nothing preflight with compensation, the outbox-pending term in `fifo_is_blocked` (with the FIFO regression test), the write-side outbox size bound, the `openat` component walk with ownership/group-write validation on root, intermediates, and the final file, the malformed-manifest `ArtifactDeclarationFailed` projection, durable discard + text fallback after post-persistence validation failure, the random-nonce key install with `getrandom`/`hmac` and constant-time verification, the nlink-aware stale-temp cleanup, and the artifact-only replay-discard rule that retains mixed batches behind the barrier.

I also traced duplicate-safety on the caption-overflow path: `remaining_text` chunks go through `send_reply` with `{message_ref}:reply:{n}` idempotency keys, so replay/retry (including after a mid-batch restart) dedups at `wn-agent`; staged-validation failures always precede any `send_media`, so a discarded batch can never have already delivered text that the fallback then repeats. Recovery re-runs produce the same artifact idempotency key and text keys, so `/retry-last` is dedup-safe end to end.

One minor note is inline (non-blocking). CI status: fmt/check/clippy/tamarin/audit were green at review time; the Rust test matrix was still running, but the PR's focused `cargo test` commands cover the touched crates.

integrations/terminal-harness/src/bridge.rs
erskingardner, line 884: Minor, non-blocking: this branch fires on every completed turn in a granted group whenever `authorize` or manifest preparation fails — including pure-text turns where no artifact was ever declared (and, since the backend never received a manifest request in this state, none could be). A group whose `export_root` is temporarily missing or fails the ownership/mode checks shows `Attachments withheld. Artifact 1: rejected …` on every reply, which is inaccurate (there is no "Artifact 1") and noisy for users who never asked for files. Consider emitting this activity only when a declaration actually arrived, or at minimum rewording the setup-failure case so it doesn't fabricate a per-artifact line.

PR #1575, opened 2026-08-27
Let a host supply its own account secret store

dannym-arx:
Stacked on #1545. Review that one first; this branch only adds the commit on top.

Marmots keep their keys in the burrow they trust. Right now every host gets the same burrow whether it wants it or not: `Marmot::open` hardcodes `AccountHome::open_with_default_keychain`, so account secrets land in the platform keychain and nowhere else. A host that already has its own encrypted store has no way in.

The seam exists in Rust already. `AccountHome::open_with_secret_store(root, Arc<dyn AccountSecretStore>)` has been there all along, and `LocalFileSecretStore` shows the shape a store needs. Nothing exposed it to bindings consumers.

## What this adds

`marmot-uniffi` gets a foreign-implementable `SecretStore` trait and `Marmot::new_with_secret_store`. Swift and Kotlin hosts can implement it directly, so an iOS app that wants its own Keychain policy or an Android app with a Keystore wrapper can now supply one instead of taking the default.

`marmot-c` implements that trait over a C function-pointer vtable, following the pattern `crates/marmot-c/AGENTS.md` already sketches for external signers:

```c
MarmotStatus marmot_client_new_with_secret_store(const char *root_path,
                                                 const char *const *relay_urls,
                                                 uintptr_t relay_urls_len,
                                                 const struct MarmotSecretStore *store,
                                                 struct MarmotClient **out_client);
```

`marmot_client_new` keeps its behavior; both constructors now route through one private open path.

The secret crosses as secret-key hex, which is what `LocalFileSecretStore` already writes and parses, and only `label` and `account_id_hex` come along from `AccountSummary`. That keeps the whole interface to strings, with no struct marshalling.

## Rules a host has to follow

These are in the module docs, and they are the part worth reviewing closely:

- Callbacks run on runtime worker threads and can run concurrently. Host code must be thread-safe.
- A callback must not call back into `marmot_*` on the same client. It runs while `AccountHome` holds its mutation lock, so reentry deadlocks.
- A host must not unwind through a callback. A panic raised inside an `extern "C"` callback aborts at that callback's own shim before this crate can catch it, so recovery is impossible. The `catch_unwind` here covers the marshalling around the call, not the host body. Making host panics recoverable would mean an `extern "C-unwind"` vtable, which is a separate decision.
- The callback is a storage boundary, not a security boundary. The hex is plaintext in memory while it crosses, held in `Zeroizing` between receipt and `Keys::parse`, and the host buffer goes back through `free_secret` right after.

Booleans cross as `uint8_t` and the callback status as a validated `uint32_t` discriminant, per the crate invariant against building a Rust `bool` or enum from an out-of-range integer.

## Checks

`cargo test -p marmot-c --features alloc-audit`, `cargo test -p marmot-uniffi`, `just c-header` with a clean diff, `just c-smoke`, clippy and fmt all pass. Tests cover a callback round trip through write, has, load and remove, `free_secret` firing exactly once, `destroy` on drop, error status mapping, an out-of-range discriminant, and a NULL store clearing the out-pointer before any work happens. One integration test opens a real client through a host store with no keychain involved.

## Two things I did not decide alone

`crates/marmot-uniffi` has no `CHANGELOG.md` and no workflow gates on one, so the entry went in `marmot-c` only. Say the word and I will add the file.

No workspace version bump, per `marmot-uniffi/AGENTS.md`. That reads against `marmot-c/AGENTS.md`, which asks for a bump when callback signatures change. Flagging the conflict instead of picking a side.

## Summary by CodeRabbit

* **New Features**
  * Added support for host-provided account secret-key storage through the UniFFI and C APIs.
  * Applications can check, save, load, and remove account secrets using their own storage provider.
  * Added a constructor for creating clients with a custom secret store.
  * Added clear status reporting and lifecycle handling for storage operations.
* **Compatibility**
  * The existing client constructor continues to use the platform keychain unchanged.
* **Documentation**
  * Documented custom secret-store callbacks, ownership, concurrency, validation, and failure behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3405,6 +3405,7 @@ dependencies = [
  "marmot-uniffi",
  "tempfile",
  "tokio",
+ "zeroize",
 ]
 
 [[package]]
--- a/crates/marmot-c/CHANGELOG.md
+++ b/crates/marmot-c/CHANGELOG.md
@@ -18,6 +18,11 @@ Versions track the workspace version; releases are tagged `marmotc-v<version>`.
   bundle. ([#1545](https://github.com/marmot-protocol/mdk/pull/1545))
 - `alloc-audit` test feature proving deep-free completeness; C smoke
   example run under gcc, clang, and valgrind in CI. ([#1545](https://github.com/marmot-protocol/mdk/pull/1545))
+- `marmot_client_new_with_secret_store` plus the `MarmotSecretStore`
+  callback vtable and `MarmotSecretStoreStatus`: a host can hold account
+  signing keys in its own storage (an encrypted vault, a custom keystore)
+  instead of the platform keychain. `marmot_client_new` is unchanged.
+  ([#1575](https://github.com/marmot-protocol/mdk/pull/1575))
 - `marmot_rotate_key_package`: rotate the account's KeyPackage under its
   proper name (with a matching `rotate_key_package` on the UniFFI
   surface); `marmot_publish_new_key_package` stays as the legacy alias.
--- a/crates/marmot-c/Cargo.toml
+++ b/crates/marmot-c/Cargo.toml
@@ -20,6 +20,8 @@ alloc-audit = []
 marmot-uniffi.workspace = true
 tokio = { workspace = true, features = ["io-util", "net", "time"] }
 libc.workspace = true
+# Holds a host-supplied secret-key hex only as long as the boundary needs it.
+zeroize.workspace = true
 
 [dev-dependencies]
 tempfile.workspace = true
--- a/crates/marmot-c/cbindgen.toml
+++ b/crates/marmot-c/cbindgen.toml
@@ -37,6 +37,7 @@ prefix = ""
 include = [
     "MarmotNotificationWakeSource",
     "MarmotCursorPersistence",
+    "MarmotSecretStoreStatus",
     "MarmotHostPerformanceOperation",
     "MarmotHostPerformanceOutcome",
 ]
--- a/crates/marmot-c/include/marmot.h
+++ b/crates/marmot-c/include/marmot.h
@@ -587,6 +587,33 @@ typedef enum MarmotCursorPersistence {
 } MarmotCursorPersistence;
 
 /**
+ * Status a [`MarmotSecretStore`] callback returns. Crosses as
+ * `uint32_t`; an out-of-range value is rejected as
+ * `MARMOT_STATUS_INVALID_ARGUMENT`.
+ */
+typedef enum MarmotSecretStoreStatus {
+  /**
+   * The operation succeeded.
+   */
+  MARMOT_SECRET_STORE_STATUS_OK,
+  /**
+   * No credential is stored for this account. Only `load_secret`
+   * should report it; `remove_secret` on a missing credential
+   * succeeds.
+   */
+  MARMOT_SECRET_STORE_STATUS_NOT_FOUND,
+  /**
+   * The store exists but cannot be reached right now (locked vault,
+   * keystore unavailable). The runtime treats this as recoverable.
+   */
+  MARMOT_SECRET_STORE_STATUS_UNAVAILABLE,
+  /**
+   * Any other failure.
+   */
+  MARMOT_SECRET_STORE_STATUS_FAILED,
+} MarmotSecretStoreStatus;
+
+/**
  * Host-side operation a UI can time and report back.
  */
 typedef enum MarmotHostPerformanceOperation {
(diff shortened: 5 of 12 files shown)


Datawav (commented):
## Automated review

Actionable findings: 3.

1. Blocking — Secret-store ownership is released after a failed constructor
- Exact changed-line evidence: `crates/marmot-c/src/lib.rs`, constructor hunk beginning `@@ -140,52 +142,122`, converts the borrowed vtable with `CSecretStore::from_c`, wraps it in `Arc`, and moves it into the `open_client` construction closure. `CSecretStore`’s new `Drop` implementation invokes `vtable.destroy(user_data)`.
- Triggering state: the vtable is valid, but a later operation fails—such as `required_str`, `str_array`, Tokio runtime creation, or opening the client—before a non-null `MarmotClient` is returned.
- Consequence: `destroy` runs even though construction failed and no client exists. A host following the documented lifetime through `marmot_client_free` can then release the same `user_data` again, causing double-free or use-after-free.
- Confidence: High. The destruction path follows directly from moving the owning `Arc<CSecretStore>` into a fallible closure; the added tests cover rejection before acceptance and destruction after successful `marmot_client_free`, but not failure after vtable acceptance.
- Smallest sound correction: defer ownership transfer until the client handle is successfully written. On every post-validation constructor failure, suppress `CSecretStore::drop`/`destroy` and leave `out_client` null. Document the failure rule and add a valid-vtable/invalid-`root_path` test asserting that `destroy` was not called.

2. Major — Signing-key hex survives in non-zeroizing marshalling copies
- Exact changed-line evidence: `crates/marmot-c/src/secret_store.rs`, added `write_secret`, wraps the input in `Zeroizing` but passes it through `with_c_str`, whose `CString::new(value)` creates a separate plaintext allocation that is ordinarily dropped rather than zeroized. `crates/marmot-uniffi/src/secret_store.rs`, added `ForeignSecretStore::write_secret`, calls `secret_key_hex.to_string()` before invoking the foreign store.
- Triggering state: any account secret is written through either the C callback adapter or the UniFFI foreign-store adapter.
- Consequence: Rust-owned heap buffers containing the signing key can remain in freed allocator memory, contradicting the new shortest-window secret-lifetime claim and weakening protection against memory disclosure.
- Confidence: High. Both copies are explicit on the changed write path, and neither resulting allocation is zeroizing.
- Smallest sound correction: marshal C text through a zeroizing NUL-terminated byte buffer and pass the UniFFI value without an ordinary `String` copy, or keep every unavoidable copy inside `Zeroizing`. Add a production-path test that inspects the marshalling buffer after callback return.

3. Minor — Out-of-range callback status violates the documented C status contract
- Exact changed-line evidence: `crates/marmot-c/include/marmot.h` documents that an out-of-range `MarmotSecretStoreStatus` is rejected as `MARMOT_STATUS_INVALID_ARGUMENT`. In added `crates/marmot-c/src/secret_store.rs`, `MarmotSecretStoreStatus::from_c(status)` rejects the discriminant, but `guard` converts that rejection to `MarmotKitError::Runtime`; the added live-callback test preserves that runtime mapping.
- Triggering state: a host callback returns an unknown value such as `4242`.
- Consequence: callers receive a runtime/store failure rather than the promised invalid-argument status, preventing reliable discrimination between malformed callback output and temporary store failure.
- Confidence: High. The generated-header contract and implementation mapping directly disagree.
- Smallest sound correction: map only `from_c` discriminant failures to the existing invalid-argument error/status and update the live-callback test to require `MARMOT_STATUS_INVALID_ARGUMENT`.

Audit evidence:
- Exact head: `7833ac3373b131af681e39f858f8adc146c72b11`
- Changed files inspected include `crates/marmot-c/src/lib.rs`, `crates/marmot-c/src/secret_store.rs`, `crates/marmot-c/include/marmot.h`, `crates/marmot-c/tests/secret_store_client.rs`, and `crates/marmot-uniffi/src/secret_store.rs`.
- Patch coverage is marked complete with no missing patch files. The sole discussion item is an automated skipped-review notice containing no technical claim, so it yields no finding.

Risk lenses checked: security/auth (authorization before mutation; secret/key lifetime and logging; malformed or replayed input); concurrency/lifecycle (cancellation and stale callbacks; ordering/race windows; process or lifecycle restart); release/build (untrusted-input and least-privilege CI; artifact identity/reproducibility; all required paths trigger checks); tests (assertions fail before the fix; boundary and negative cases; test exercises production behavior).

Check evidence: the bundle records all checks completed and successful under its aggregate status. Relevant successful checks include C bindings smoke, Rust check, Rust format, Rust clippy, Rust audit, all four Rust test shards, simulator and vectors, relay runtime tests, Tamarin proofs, release-assertion convergence policy, and platform builds. “Publish wn-agent release” was completed as skipped. These checks do not exercise the constructor-failure ownership boundary or prove zeroization of temporary marshalling allocations.

dannym-arx:
> Major — Signing-key hex survives in non-zeroizing marshalling copies

this is not possible to fix, you cannot safely zeroize memory you do not own; the way to think about it is that once it's outside the boundary it's outside of your control

Issue #482, opened 2026-06-20
Hermes: inbound + outbound encrypted media support (download_media / send_media)


erskingardner:
Part of the Hermes <-> OpenClaw parity effort. Connector support landed in marmot-protocol/darkmatter#501 (`send_media`, `download_media`, `media[]` on inbound events).

Hermes neither surfaces inbound encrypted media to the agent nor sends media. Mirror OpenClaw:
- inbound: `download_media` each `imeta` ref to a local path and attach to the Hermes MessageEvent (re-stage under an allowlisted media root; see the outbound-media security issue),
- outbound: map media sends to `send_media`.

### Acceptance
- Inbound images reach the agent.
- Outbound media sends as a kind-9 with attachments.

PR #1581, opened 2026-08-27
fix(app): require relay coverage for replay completion

### LINKED ISSUE #1578: marmot-app: any-relay EOSE can falsely complete multi-relay history recovery
## Summary

Account replay completion is currently tracked per subscription as “EOSE received from at least one relay.” That is not proof that the relay holding the missing history served it.

Production group publication requires only one relay acknowledgement. A commit or message may therefore exist on relay B but not relay A. During recovery, fast-empty relay A can return EOSE while B is slow, reconnecting, or temporarily unreachable. MDK then classifies the account replay as complete even though B has not served the event needed for convergence.

This affects epoch-gap backfill on `master` and also the overflow-recovery design in PR #1567, which proposes clearing its durable overflow marker on the same completion verdict.

## Failure sequence

1. A required commit or application event is accepted only by relay B.
2. A recipient misses it and later starts an unfloored recovery subscription against A and B.
3. Relay A has no matching history and quickly sends EOSE.
4. Relay B remains disconnected or delayed.
5. The per-subscription any-relay EOSE counters report complete.
6. Recovery is consumed/disarmed—or PR #1567 clears the overflow marker—without receiving B's history.
7. A later ordinary cursor-floored rebuild can again exclude the missing event.

## Evidence

Current `master` snapshot: [`b0214579`](https://github.com/marmot-protocol/mdk/commit/b0214579f87ee8cc7636b0913f606feaa8a00b33).

- [`AccountSubscriptionEose`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/transport-nostr-adapter/src/lib.rs#L333-L359) explicitly defines subscription progress as EOSE from “at least one relay.”
- [`account_subscription_eose`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/transport-nostr-adapter/src/lib.rs#L1449-L1458) collapses endpoint state through `subscription_any_eose`.
- [`backfill_drain_verdict`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-app/src/client/sync.rs#L3541-L3553) treats that aggregate as `DrainVerdict::Complete`.
- A successful replay can call [`mark_replayed`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-app/src/client/sync.rs#L2048-L2085), disarming account-wide epoch-stall recovery.
- Production routing uses [`required_acks: 1`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-app/src/lib.rs#L3632-L3638), so history is not guaranteed to exist on every configured relay.
- PR #1567 clears overflow recovery when this verdict is complete: [`recover_delivery_overflow`](https://github.com/marmot-protocol/mdk/blob/c8276c369ad2b5cedf89d3e7185c6736cd3b2e8c/crates/marmot-app/src/client/sync.rs#L1190-L1208).

- #1561 owns detecting per-account queue overflow, holding back unsafe cursors, and forcing replay. This issue owns the independent proof required before *any* multi-relay replay may be declared complete. Fixing overflow detection without fixing relay coverage can still clear recovery too early.
- #1562 bounds account-worker execution during long/duplicate replay. Yielding safely and proving relay coverage are separate requirements.
- #1548 introduced EOSE-gated backfill completion, but the current gate aggregates away which relay supplied the evidence.

## Required behavior

Define a relay-coverage-aware completion contract. It must retain endpoint-level subscription attempt and EOSE state for the route snapshot used by the recovery and must not treat EOSE from a relay known to lack the missing history as proof that another relay's history was served.

The design may derive a sufficient coverage set from durable publication receipts/provenance, recovery evidence, or a conservative route snapshot. If coverage cannot be proven because a relevant relay is unavailable, recovery must remain explicitly incomplete and retryable; availability pressure must not be converted into a false success. Relay removal/retirement should be an explicit routing decision rather than an implicit timeout-based proof of history completeness.

Do not hold the serial account worker indefinitely while waiting: preserve #1562's bounded execution quantum, durable intent, and paced retries. Keep observability privacy-safe and endpoint-free outside encrypted routing state.

## Acceptance criteria

- A deterministic two-relay test publishes the missing event only to B, then runs recovery with fast-empty A and delayed B.
- EOSE from A alone does not produce `DrainVerdict::Complete`, call `mark_replayed`, clear an overflow marker, or restore an ordinary cursor floor.
- Once B serves the event and its relevant EOSE, recovery completes and the durable intent/marker may clear.
- If B remains unavailable, recovery yields within the worker quantum, stays durably pending across restart, and retries with bounded backoff.
- A route snapshot change during recovery cannot silently shrink the required coverage and falsely complete an older attempt.
- Tests cover inbox and group subscriptions, partial relay connectivity, relay reconnect, duplicate EOSE, and EOSE arriving before/after stored events.
- Overflow-recovery coverage is added to PR #1567 or its successor before #1561 is considered fully resolved.

erskingardner:
## Summary
- retain endpoint-level EOSE state for the immutable route snapshot issued by each account activation, so fast-empty relay A cannot stand in for relay B's missing history
- keep the snapshot independent from live group-route churn and require full inbox/group endpoint coverage before `DrainVerdict::Complete`
- remove the quiescence fallback that converted repeated relay unavailability into success; bounded worker quanta, durable intent, and paced retries remain in place until reconnect coverage is proven
- make PR #1567's shared completion verdict conservative when that work is rebased onto this fix

## Test plan
- [x] New deterministic two-relay regression failed before the fix and passes after it
- [x] Repeated-unavailability regression failed before the fix and passes after it
- [x] `cargo test -p transport-nostr-adapter`
- [x] `cargo test -p marmot-forensics`
- [x] Recovery-focused `marmot-app` tests with `test-policy-overrides`
- [x] `just fast-ci`

The final full `marmot-app` feature-suite run passed 804 tests, ignored one benchmark, and hit one unrelated SQLCipher fixture failure under parallel execution (`file is not a database`); that exact test passed immediately on rerun.

Fixes #1578

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved full-history recovery across multiple relays by requiring completion confirmation from every relay and subscription involved.
  - Prevented fast, unavailable, or duplicate relay responses from falsely marking recovery as complete.
  - Recovery intent now remains pending when required relay history is unavailable, enabling reliable retries.
  - Replay progress resets correctly for new activations and remains tied to the original relay coverage.

### DIFF
--- a/crates/marmot-app/src/client/sync.rs
+++ b/crates/marmot-app/src/client/sync.rs
@@ -18,8 +18,8 @@ use crate::media::media_imeta_tags_are_valid;
 use crate::notifications;
 use crate::{
     AccountState, AppError, AppGroupAdminPolicyComponent, AppMessageProjection,
-    AppPerformanceTelemetry, ClassifiedSyncFailure, EPOCH_BACKFILL_EOSE_ATTEMPT_LIMIT,
-    EPOCH_BACKFILL_EOSE_WAIT, EPOCH_BACKFILL_EXECUTION_QUANTUM, EPOCH_BACKFILL_RETRY_BACKOFF,
+    AppPerformanceTelemetry, ClassifiedSyncFailure, EPOCH_BACKFILL_EOSE_WAIT,
+    EPOCH_BACKFILL_EXECUTION_QUANTUM, EPOCH_BACKFILL_RETRY_BACKOFF,
     EPOCH_BACKFILL_RETRY_BACKOFF_CAP, SDK_DRAIN_WAIT, SDK_FIRST_SYNC_WAIT, SelfMembership,
     SyncFailure, SyncSummary, TRANSPORT_CURSOR_MAX_FUTURE_SKEW, unix_now_seconds,
 };
@@ -92,11 +92,6 @@ enum DrainCompletion {
         silence_budget: Duration,
         execution_quantum: Duration,
     },
-    /// Epoch-gap backfill that has spent its end-of-stored-events attempt
-    /// budget ([`EPOCH_BACKFILL_EOSE_ATTEMPT_LIMIT`]). Drains exactly like
-    /// [`Self::Quiescence`] so an account whose group route has one unreachable
-    /// relay can still heal, and reports itself as the weaker claim it is.
-    QuiescenceFallback { execution_quantum: Duration },
 }
 
 impl DrainCompletion {
@@ -105,8 +100,7 @@ impl DrainCompletion {
             Self::Quiescence => None,
             Self::EndOfStoredEvents {
                 execution_quantum, ..
-            }
-            | Self::QuiescenceFallback { execution_quantum } => Some(execution_quantum),
+            } => Some(execution_quantum),
         }
     }
 }
@@ -121,10 +115,6 @@ enum DrainVerdict {
     /// Every live subscription reached end-of-stored-events and the relays then
     /// went quiet: the account's stored history was served in full.
     Complete,
-    /// The relays went quiet under the fallback contract, after the gate had
-    /// spent its attempt budget. Recovery ran and is not a failure, but nothing
-    /// confirms the history was served in full.
-    QuiescenceFallback,
     /// The silence budget ran out with stored history still unconfirmed, though
     /// some relay did reach end-of-stored-events.
     EoseTimeout,
@@ -145,7 +135,7 @@ impl DrainVerdict {
     /// The audit row's `error_kind` for a drain that did not complete.
     fn error_kind(self) -> Option<&'static str> {
         match self {
-            Self::Complete | Self::QuiescenceFallback => None,
+            Self::Complete => None,
             Self::EoseTimeout => Some("backfill_drain_eose_timeout"),
             Self::NoRelayEose => Some("backfill_drain_no_relay_eose"),
             Self::NovelProgressQuantumYield => Some("backfill_drain_novel_progress_quantum_yield"),
@@ -153,12 +143,10 @@ impl DrainVerdict {
         }
     }
 
-    /// What the completed audit row should claim about this drain, so a
-    /// fallback is never read as an end-of-stored-events confirmation.
+    /// What the completed audit row should claim about this drain.
     fn completion_kind(self) -> Option<EpochBackfillCompletionKind> {
         match self {
             Self::Complete => Some(EpochBackfillCompletionKind::EndOfStoredEvents),
-            Self::QuiescenceFallback => Some(EpochBackfillCompletionKind::QuiescenceFallback),
             Self::EoseTimeout
             | Self::NoRelayEose
             | Self::NovelProgressQuantumYield
@@ -1084,16 +1072,11 @@ impl AppClient {
     async fn backfill_sdk_relay(
         &mut self,
         counts: &mut DrainCounts,
-        eose_unconfirmed_ordinal: u64,
     ) -> Result<(SyncSummary, DrainVerdict), ClassifiedSyncFailure> {
         let execution_quantum = self.epoch_backfill_execution_quantum();
-        let completion = if eose_unconfirmed_ordinal >= EPOCH_BACKFILL_EOSE_ATTEMPT_LIMIT {
-            DrainCompletion::QuiescenceFallback { execution_quantum }
-        } else {
-            DrainCompletion::EndOfStoredEvents {
-                silence_budget: self.epoch_backfill_eose_wait(),
-                execution_quantum,
-            }
+        let completion = DrainCompletion::EndOfStoredEvents {
+            silence_budget: self.epoch_backfill_eose_wait(),
+            execution_quantum,
         };
         self.drain_sdk_relay(counts, completion).await
     }
@@ -1260,9 +1243,6 @@ impl AppClient {
                 Ok(Ok(None)) => {
                     break match completion {
                         DrainCompletion::Quiescence => DrainVerdict::Complete,
-                        DrainCompletion::QuiescenceFallback { .. } => {
-                            DrainVerdict::QuiescenceFallback
-                        }
                         DrainCompletion::EndOfStoredEvents { .. } => {
                             self.backfill_drain_verdict().await
                         }
@@ -1282,12 +1262,6 @@ impl AppClient {
                 }
                 Err(_) => match completion {
                     DrainCompletion::Quiescence => break DrainVerdict::Complete,
-                    DrainCompletion::QuiescenceFallback { execution_quantum } => {
-                        if drain_started.elapsed() >= execution_quantum {
-                            break DrainVerdict::quantum_yield(counts);
-                        }
-                        break DrainVerdict::QuiescenceFallback;
-                    }
                     DrainCompletion::EndOfStoredEvents {
                         silence_budget,
                         execution_quantum,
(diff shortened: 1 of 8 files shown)


erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T11:51:14Z` |
| Commit reviewed | `66a5b354bf1eeb0073065b258cb886e029d13192` |
| Model | `Cursor Grok 4.6` |

The issue is real and this is the right fix. Completion used to mean “EOSE from any one relay on each logical subscription,” which is exactly how a fast-empty relay can stand in for missing history. The new activation snapshot treats each endpoint-scoped attempt as its own proof obligation, and it does not shrink under later `sync_account_groups`. Removing `QuiescenceFallback` is the other half of the bug: availability pressure is no longer converted into `end_of_stored_events`.

The tests pin the contract that matters. The adapter case covers inbox + group, duplicate EOSE, event-before-EOSE, and a mid-attempt route shrink. The app cases flip the old “fast relay completes the drain” behavior to “intent stays pending until the slow relay reconnects,” and they keep worker-quantum yields from weakening the gate.

A few residual notes, none blocking:

- The snapshot is per activation, not per recovery generation. That is correct for this issue: each backfill retry re-activates and starts a fresh all-false snapshot. An explicit route retirement can therefore complete on a later attempt without the removed relay. Mid-drain shrink cannot.
- Production still yields at the 5s worker quantum and then resubscribes. A reachable-but-slow relay that does not EOSE inside that window will be asked again rather than completing on a sibling’s EOSE. That is the trade #1578 asked for; just do not revive a timeout-to-success escape hatch when #1567 rebases onto this verdict.
- Overflow-marker coverage remains on #1567, as the issue said.

`AccountSubscriptionEose::complete` now requires the endpoint counts, but its first sentence still reads like the old any-relay subscription rule. Worth tightening when you are next in that file.

Could not submit this as a GitHub approve: the API rejects self-approval on the author’s own PR.

crates/transport-nostr-adapter/src/lib.rs
```diff
@@ -1380,6 +1441,26 @@ impl AdapterState {
...
+        account_id: &MemberId,
+        subscriptions: &[NostrSubscription],
+    ) {
+        let subscriptions = subscriptions
+            .iter()
+            .map(|subscription| {
+                let relays = subscription
+                    .endpoints()
+                    .iter()
+                    .map(|endpoint| (self.relay_index.index_for(endpoint), false))
```
erskingardner, line 1455: The per-endpoint proof obligations are keyed by `RelayIndex`, and `RelayIndexRegistry` assigns indices by raw `TransportEndpoint` string. The two sides that must coincide here are spelled from different sources in production: this registration side uses the verbatim signed routing strings, while inbound EOSE arrives as `TransportEndpoint(relay_url.to_string())` from nostr-sdk (`handle_relay_notification` in `marmot-app/src/relay_plane/mod.rs`, or the sdk forwarder). `nostr::RelayUrl` is built on `url::Url`, which normalizes on parse — scheme/host case folding, default-port dropping, IDN/percent-encoding handling — and only trailing slashes round-trip exactly (nostr preserves `has_trailing_slash`). The adapter already treats this verbatim-vs-inbound mismatch class as real: `CanonicalEndpoint::matches` exists because of mdk#482 and compares parsed `RelayUrl`s for exactly this pairing.

Pre-PR the raw-keying weakness in this registry was mostly harmless for the account gate: `with_eose` counted a subscription once *any* endpoint key-matched, so one well-spelled relay sufficed for completion. Post-PR completion requires every frozen `(subscription, endpoint)` attempt, and an EOSE recorded under a normalized spelling that differs from the verbatim route string is silently ignored by `mark_eose`/`record_eose` (unknown relay index → no credit). A group route carrying e.g. `wss://Relay.Example.com:443` would then never satisfy the frozen gate even though relay B keeps serving it — durable intent retries forever on the 5-minute-capped backoff with honest `EoseTimeout` failures, while routing/delivery on the same endpoint works fine. That contradicts the issue's own acceptance criterion ("once B serves the event and its relevant EOSE, recovery completes") and wedges recovery for exactly the routes #482 warned about.

Could this key both sides through one normalization helper — e.g. `RelayUrl::parse(endpoint.as_str()).map(|u| TransportEndpoint(u.as_str().to_owned())).unwrap_or_else(|_| endpoint.clone())` — used consistently at `record_account_replay_start`, `record_subscription_starts`, `record_subscription_eose` (and ideally `record_delivery_timing`, which shares the registry)? A regression test whose verbatim route spelling differs from the inbound EOSE spelling only by host case or an explicit default port would pin it.

crates/marmot-app/src/client/sync.rs
```diff
@@ -121,10 +115,6 @@ enum DrainVerdict {
     /// Every live subscription reached end-of-stored-events and the relays then
     /// went quiet: the account's stored history was served in full.
     Complete,
```
erskingardner, line 117: Minor: the `Complete` doc still reads "every live subscription reached end-of-stored-events" — the verdict now derives from the activation's frozen endpoint-coverage snapshot (`AccountSubscriptionEose` redefinition in the adapter), not the live subscription set. Worth aligning the wording so the next reader doesn't go looking for a live-set gate.

erskingardner (commented):
The fix matches #1578's required behavior well: a frozen per-activation endpoint-coverage snapshot, route churn that cannot shrink an in-flight attempt's proof obligation, removal of the quiescence fallback that converted relay unavailability into success, and conservative completion for the #1567 rebase. The new tests cover the deterministic two-relay fast-empty/slow-relay scenario, duplicate EOSE, reconnect, route shrink mid-attempt, and repeated unavailability — and the failing-before/passing-after framing is exactly right.

One blocking correctness gap, inline: the coverage map keys per-endpoint attempts by raw-string `RelayIndex`, while inbound EOSE endpoints arrive as nostr-normalized `relay_url.to_string()` spellings. A verbatim route spelling that only differs by URL normalization details (host case, explicit default port, IDN/encoding) can therefore never satisfy the frozen gate — recovery retries forever while delivery on the same relay works. That class of mismatch is the same one `CanonicalEndpoint::matches` was built for in mdk#482.

Also noted inline: a stale doc comment on `DrainVerdict::Complete`.

At review time, CI was green except `Relay runtime tests`, `Simulator and vectors`, and the darwin-x86_64 build, which were still running — merge should wait for those regardless.

crates/transport-nostr-adapter/src/lib.rs
erskingardner: Addressed in follow-up PR #1583. RelayIndexRegistry now derives every device-local relay identity through one RelayUrl normalization helper, so signed verbatim route spellings and nostr-sdk-normalized subscription, event, and EOSE endpoints share coverage and timing state without rewriting authoritative routing data. The adapter regression exercises a mixed-case/default-port route and normalized inbound event/EOSE; cargo test -p transport-nostr-adapter and just fast-ci pass.

crates/marmot-app/src/client/sync.rs
erskingardner: Addressed in follow-up PR #1583. DrainVerdict::Complete now documents completion against every endpoint-scoped attempt in the activation frozen route snapshot, followed by relay quiet, instead of referring to live subscriptions.

Issue #1020, opened 2026-07-21
Persist frozen Nostr fanout and partial-accept publish lifecycle


erskingardner:
## Summary

Implement the adopted durable Nostr publish lifecycle for a single MLS output sent to multiple required relays.

The current issue was scoped as test coverage for partial acceptance. The actual gap is architectural: the target set and each target's first-attempt status must be frozen and durable before publishing, while MLS `PendingPublish` is released after at least one relay acknowledges. Outstanding first attempts continue after that release and must never reopen, undo, or restage the MLS state.

## Normative source

Adopted by marmot-protocol/marmot#408:

- `protocol-core/publish-lifecycle.md`
- `transports/nostr.md`
- `protocol-core/convergence.md`

Related: #1057 supplies route derivation/history; this issue owns outbound fanout state.

## Start here

- `crates/traits/src/engine_state.rs`
- `crates/cgka-engine/src/epoch_manager.rs`
- `crates/cgka-engine/src/group_lifecycle.rs`
- outbound runtime in `crates/marmot-app/`
- durable intent/outbox storage in `crates/storage-sqlite/`
- `crates/cgka-engine/tests/publish_lifecycle.rs`

## Required behavior

- Before the first external side effect, durably record the exact serialized event/message, frozen required target set, and per-target `not_attempted` status.
- Attempt every required target at least once unless superseded by a terminal lifecycle rule.
- On the first positive relay acknowledgment, confirm/release the MLS `PendingPublish` transition exactly once.
- Continue outstanding first attempts after MLS confirmation. Their later success/failure updates fanout state only.
- Later responses cannot reopen `PendingPublish`, roll back a confirmed commit, merge twice, regenerate the event, or create a new target set.
- If every required first attempt fails and none acknowledges, apply the adopted no-recipient failure/rollback semantics; never report success.
- Restart resumes only outstanding targets using the same bytes/event ID and the original target set.
- Route changes after staging affect later messages, not the frozen fanout of this one.
- Expose separate MLS-confirmed and fanout-complete outcomes to callers/audit without leaking relay URLs.

## Acceptance criteria

- Tests cover first/last/middle acknowledgment, mixed ack/fail/timeout, all-fail, duplicate callbacks, callbacks after restart, and route changes mid-flight.
- A first acknowledgment releases `PendingPublish` while unattempted targets remain durably scheduled.
- No later outcome can reopen/undo/restage an already confirmed transition.
[...]

PR #1582, opened 2026-08-27
Fix ambiguous relay publish recovery

### LINKED ISSUE #1577: marmot-account: ambiguous relay publish failures can roll back MLS state after exposure
## Summary

Ordinary group publishing collapses relay timeouts, missing acknowledgements, and SDK send errors into endpoint failure. The frozen fanout then treats every endpoint error as terminal: it marks the targets failed, rolls back pending MLS state when no positive acknowledgement was observed, and deletes the durable fanout.

That is unsafe because a timeout or lost `OK` is not proof that the relay rejected the event. A relay may have accepted and forwarded an MLS commit before its acknowledgement was lost. Peers can then advance while the author rolls back, producing a real convergence split. For an application message, peers may receive a message that the sender retracts locally; a user retry creates a second event and duplicate visible content.

This is most likely under poor or intermittent connectivity—the exact situation in which acknowledgement loss is common.

## Evidence on current `master`

Source snapshot: [`b0214579`](https://github.com/marmot-protocol/mdk/commit/b0214579f87ee8cc7636b0913f606feaa8a00b33).

- [`send_event_to_relay`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/transport-nostr-adapter/src/sdk_client.rs#L479-L539) maps a missing acknowledgement, SDK error, or 12-second timeout to the same `TransportEndpointFailure` shape after bounded retries. These outcomes do not establish whether the event was exposed.
- [`drive_outbound_fanout`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-account/src/runtime.rs#L3101-L3161) converts every adapter `Err` into `mark_target_failed`.
- [`resolve_outbound_fanout_mls`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-account/src/runtime.rs#L3206-L3234) rolls back pending MLS state once fanout is complete with zero observed acceptances.
- The regression [`frozen_fanout_whole_call_adapter_errors_are_terminal`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-account/tests/runtime/frozen_fanout.rs#L174-L193) explicitly pins all whole-call adapter errors as terminal.

- #1020 implemented durable target snapshots and correct partial-accept handling. It explicitly retained the rule “all first attempts fail and none acknowledges => rollback.” This issue is the uncovered next layer: some apparent failures are acknowledgement-unknown rather than proof of no exposure.
- #1443 owns operation identity and unknown completion at the app/UniFFI command boundary. It does not classify relay exposure or reconcile frozen publication state.
- #926 owns stale iOS relay sockets. Resetting a stale socket helps liveness but does not make a timed-out publication safe to roll back.

## Required behavior

Introduce a typed endpoint publication outcome that preserves at least:

- accepted;
- explicit terminal rejection;
- definitely not exposed before any send attempt;
- possibly exposed / acknowledgement unknown;
- transiently unavailable and retryable.

After possible exposure, MDK must retain the exact signed event, event ID, frozen target set, per-target state, and pending MLS relationship. It must retry or reconcile that same event after connectivity returns; it must not regenerate the event, delete the fanout, roll back the MLS transition, or return a definite-failure result that invites a semantically new retry.

Definite transient connection failures should remain durably retryable instead of silently discarding the outbound obligation. Only outcomes that prove terminal rejection/no exposure across the required target policy may use rollback semantics.

Expose a typed unresolved/unknown-completion result to app consumers. Keep logs and telemetry aggregate-only: no account/group/message identifiers, relay URLs, payloads, ciphertext, or key material.

## Acceptance criteria

- A relay-backed test accepts and stores an MLS commit but withholds/drops its `OK`; the sender must not roll back while a peer can apply the commit.
- After restart or connectivity restoration, MDK republishes the byte-identical event with the same event ID and confirms the pending MLS transition when acknowledgement is obtained.
- The same test shape for an application message does not retract the message as definitely failed or create a second transport event on internal retry.
- Explicit relay rejection remains distinguishable from timeout, connection failure, SDK error, and acknowledgement loss.
- A pre-send connection failure creates no false exposure claim but retains a bounded durable retry obligation for an otherwise accepted user send.
- Mixed accepted/rejected/unknown endpoint outcomes preserve #1020 partial-accept behavior and never reopen a confirmed transition.
- Crash tests cover before send, after relay acceptance/before acknowledgement persistence, after unknown outcome persistence, and during retry.
- Retry is bounded/backed off, survives restart, and cannot block unrelated groups or accounts indefinitely.

erskingardner:
## Summary

- classify per-endpoint publication failures by exposure and retry semantics while preserving the transport event ID on unresolved outcomes
- keep the exact frozen fanout and pending MLS state durable across ambiguous or transient failures, then retry with capped exponential backoff
- surface completion-unknown through the app and UniFFI as a non-failure without retracting the application message or encouraging a new semantic send

## Test plan

- [x] regression test failed before the implementation and passes after it
- [x] cargo test -p cgka-traits
- [x] cargo test -p marmot-account
- [x] cargo test -p transport-nostr-adapter --features sdk
- [x] targeted marmot-app and marmot-uniffi tests
- [x] workspace check and Clippy, including OTLP feature builds
- [x] all other fast-CI recipes
- [ ] just fast-ci (campaign-toolchain-gate is blocked on current master by GNU-style sed -i usage under macOS BSD sed)

Fixes #1577

## Summary by CodeRabbit

- **New Features**
  - Added clearer message delivery outcomes, including published, retryable, terminally rejected, and completion-unknown statuses.
  - Improved retry handling for temporarily unavailable endpoints, including persistence across restarts.
  - Preserved and surfaced messages that may have been exposed but lack confirmed acknowledgement.
  - Added structured details for unresolved group and application-message publications.
  - Exposed the new completion status consistently through application interfaces.

- **Bug Fixes**
  - Corrected publication counts, summaries, and local projections to reflect confirmed event delivery.
  - Improved relay failure classification and partial-acceptance reporting.
  - Ensured deferred group invitations are released only after confirmation.

### DIFF
--- a/crates/marmot-account/src/runtime.rs
+++ b/crates/marmot-account/src/runtime.rs
@@ -24,11 +24,11 @@ use cgka_traits::maintenance::{
 };
 use cgka_traits::transport::{TransportEnvelope, TransportMessage};
 use cgka_traits::{
-    EpochId, FanoutMlsState, GroupId, MemberId, OutboundFanout, OutboundFanoutOutcome,
-    StorageError, Timestamp, TransportAccountActivation, TransportAdapter, TransportAdapterError,
-    TransportDelivery, TransportEndpoint, TransportEndpointFailure, TransportEndpointFailureKind,
-    TransportEndpointReceipt, TransportGroupSync, TransportPublishReport, TransportPublishRequest,
-    TransportPublishTarget,
+    EpochId, FanoutMlsState, FanoutPendingKind, GroupId, MemberId, OutboundFanout,
+    OutboundFanoutOutcome, StorageError, Timestamp, TransportAccountActivation, TransportAdapter,
+    TransportAdapterError, TransportDelivery, TransportEndpoint, TransportEndpointFailure,
+    TransportEndpointFailureKind, TransportEndpointReceipt, TransportGroupSync,
+    TransportPublishReport, TransportPublishRequest, TransportPublishTarget,
 };
 use futures::StreamExt;
 use futures::stream::FuturesUnordered;
@@ -92,6 +92,12 @@ struct PublishStatus {
     accepted_by_any_endpoint: bool,
     possible_ambiguous_exposure: bool,
     retry_deferred: bool,
+}
+
+struct PendingFanoutContinuation {
+    pending: PendingStateRef,
+    kind: FanoutPendingKind,
+    post_confirmation_welcomes: Vec<TransportMessage>,
 }
 
 enum PreparedLegacyPublish {
@@ -2124,8 +2130,8 @@ where
                     source_epoch,
                     retention,
                 } => {
-                    let engine_message_id = msg.id.clone();
                     let reports_before = output.reports.len();
+                    let unresolved_before = output.unresolved_publishes.len();
                     let status =
                         Box::pin(self.publish_one(msg, None, output, queue, context.clone()))
                             .await?;
@@ -2153,18 +2159,14 @@ where
                             );
                         }
                     } else if status.retry_deferred
-                        && let Some(unresolved) = output
-                            .unresolved_publishes
-                            .iter()
-                            .rev()
-                            .find(|unresolved| unresolved.message_id == engine_message_id)
+                        && let Some(unresolved) = output.unresolved_publishes.get(unresolved_before)
                     {
                         output
                             .unresolved_app_messages
                             .push(UnresolvedApplicationMessage {
                                 group_id,
                                 app_event_id,
-                                message_id: engine_message_id,
+                                message_id: unresolved.message_id.clone(),
                                 reason: unresolved.reason,
                             });
                     }
@@ -2553,64 +2555,27 @@ where
         queue: &mut VecDeque<PublishWork>,
         context: Option<AuditEventContext>,
     ) -> AccountResult<()> {
-        let mut all_published = true;
-        let mut any_welcome_exposed = false;
-        let mut retry_deferred = false;
-        let mut welcome_failures = Vec::new();
-        let mut delivered_welcome_ids = Vec::new();
-        for welcome in welcomes {
-            let recipient = welcome_recipient(&welcome);
-            let welcome_id = welcome.id.clone();
-            let failures_before = output.failures.len();
-            let status =
-                Box::pin(self.publish_one(welcome, None, output, queue, context.clone())).await?;
-            any_welcome_exposed |= status.accepted_by_any_endpoint;
-            retry_deferred |= status.retry_deferred;
-            if status.met_required_acks {
-                delivered_welcome_ids.push(welcome_id);
-            } else {
-                if let Some(recipient) = recipient {
-                    welcome_failures.push(self.welcome_delivery_failure(
-                        welcome_id,
-                        recipient,
-                        None,
-                        output,
-                        failures_before,
-                    ));
-                }
-                all_published = false;
-                if !any_welcome_exposed {
-                    break;
-                }
-            }
-        }
-
-        if all_published || any_welcome_exposed {
+        let mut welcomes = welcomes.into_iter();
+        let Some(first_welcome) = welcomes.next() else {
             let effects = self.confirm_published_retrying(pending).await?;
-            let group_id = confirmed_group_id(&effects);
             output
                 .pending
                 .push(PendingResolution::Confirmed { pending });
             output.absorb_session_effects(effects, queue);
-            for message_id in &delivered_welcome_ids {
-                self.mark_welcome_delivered_best_effort(message_id);
-            }
-            for failure in &mut welcome_failures {
-                if failure.group_id.is_none() {
-                    failure.group_id = group_id.clone();
-                }
-            }
-            // Only a confirmed legacy create leaves welcomes worth
-            // re-delivering; a rolled-back create discards the whole
-            // evolution, welcomes included.
-            output.welcome_failures.extend(welcome_failures);
-        } else if !retry_deferred {
-            let effects = self.session.publish_failed(pending).await?;
-            output
-                .pending
-                .push(PendingResolution::RolledBack { pending });
-            output.absorb_session_effects(effects, queue);
-        }
+            return Ok(());
+        };
+        Box::pin(self.publish_one_with_post_confirmation_welcomes(
+            first_welcome,
+            Some(PendingFanoutContinuation {
+                pending,
+                kind: FanoutPendingKind::CreateGroup,
+                post_confirmation_welcomes: welcomes.collect(),
+            }),
+            output,
+            queue,
+            context,
+        ))
+        .await?;
         Ok(())
     }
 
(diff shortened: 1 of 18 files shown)


crates/marmot-account/src/runtime.rs
```diff
@@ -2148,6 +2152,21 @@ where
                                 "accepted application publish had no transport report"
                             );
                         }
+                    } else if status.retry_deferred
+                        && let Some(unresolved) = output
+                            .unresolved_publishes
+                            .iter()
+                            .rev()
+                            .find(|unresolved| unresolved.message_id == engine_message_id)
```
erskingardner: This lookup compares the engine message id against `UnresolvedPublish.message_id`, which comes from `frozen_fanout_report` and prefers `published_message_id`. `OutboundFanout::published_message_id` is documented as the adapter-visible signed-event id, distinct from the frozen request id. Production Nostr always sets that id via `TransportPublishFailure::with_message_id`.

`ambiguous_application_publish_is_retained_without_definite_failure` uses `error_for_endpoints` (plain `TransportAdapterError::Publish`, no transport id), so this find succeeds in the test and fails on the real adapter. App/UniFFI then fall through to `AcceptedPending`.

Attach from the `UnresolvedPublish` this `publish_one` just appended (index snapshot, or last entry) instead of matching ids. The commit-retry test already injects a distinct transport id — the app-message test should do the same.

crates/marmot-account/src/runtime.rs
```diff
@@ -3162,13 +3211,24 @@ where
...
         let report = frozen_fanout_report(&fanout);
+        let fanout_outcome = fanout.outcome();
+        let possible_ambiguous_exposure = fanout.possible_exposure();
         let status = PublishStatus {
             met_required_acks: report.met_required_acks(),
             accepted_by_any_endpoint: report.accepted_count() > 0,
-            possible_ambiguous_exposure: false,
-            retry_deferred: false,
+            possible_ambiguous_exposure,
+            retry_deferred: fanout_outcome.outstanding_targets > 0,
```
erskingardner, line 3240: Two problems share this assignment.

First: `outstanding_targets > 0` is not the same as “this send’s completion is unknown.” After a required-acks-met partial accept, leftover targets are a background fanout obligation. Pushing `UnresolvedPublish` on the next line makes `accept_disposition_from_effects` return `CompletionUnknown` for the common 1-of-N timeout case, even though MLS is confirmed and `published` counts the accepted report. Emit `UnresolvedPublish` only when `!status.met_required_acks`.

Second: this is what newly activates `retry_deferred` on create/evolution. `publish_group_created` then skips both confirm and rollback, while `publish_one(welcome, None, ...)` stages a welcome fanout with no pending ref. `resume_outbound_fanouts` can retry the welcome bytes and still never confirm the create. `publish_group_evolution` only publishes welcomes when the commit was accepted in the same call, so an unresolved commit later confirmed on resume drops those welcomes. Add a create-with-withheld-OK test and a commit+welcome recover test; the durable obligation has to include both the exact welcome(s) and a confirm path.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T12:07:00Z` |
| Commit reviewed | `3758c2abe4e705ccabc5a41b3a575ab83c5e679b` |
| Model | `Cursor Grok 4.6` |

Posted as a comment review because GitHub rejects `REQUEST_CHANGES` on the author's own PR. The findings below are still blocking.

The core idea is right and overdue: #1577 is a real convergence split, and the typed endpoint vocabulary plus keeping the exact event durable is the correct fix. The commit-retry test (`ambiguous_commit_retries_exact_event_after_restart_and_peer_can_advance`) actually proves the dangerous case.

The implementation does not finish the confirm/rollback state machine for every publish path that used to treat adapter errors as terminal, and the new app-facing `CompletionUnknown` signal is both too narrow (app messages) and too wide (already-acked group sends).

**Blocking**

1. Setting `retry_deferred` whenever targets remain outstanding activates existing `publish_group_created` / `publish_group_evolution` branches that skip both confirm and rollback. Welcome fanouts are staged with `pending: None`, so `resume_outbound_fanouts` can never confirm a create. A create-with-timeout test is missing.
2. `unresolved_app_messages` is attached only when `UnresolvedPublish.message_id` equals the engine message id. `OutboundFanout` documents that the adapter-reported id is different, and production Nostr always sets it. The new app test uses an untyped `Publish` error, so it never exercises that split; hosts will see `AcceptedPending` instead of `CompletionUnknown`.

**Serious**

3. Any leftover outstanding target emits `UnresolvedPublish`, and `accept_disposition_from_effects` then returns `CompletionUnknown` even after required acks were met. A 1-of-N accept (the #1020 success case) now looks unpublished.

The adapter classification work (connect = retryable, lost OK = possibly exposed, `error:` not treated as terminal, conservative default for old rows) looks sound. Duplicate prefix mapping to `TerminalRejected` is latent only because `relay_endpoint_publish_accepted` already treats `duplicate:` as success.

erskingardner:
Addressed the blocking review feedback in 8465c6d4:

- correlate unresolved app sends with the adapter-visible transport id
- reject late exposure mutations for terminal fanout targets
- emit CompletionUnknown only while required acknowledgements remain unmet
- persist and recover exact post-confirmation Welcome continuations for legacy create and invite paths
- use send_summary_from_effects for group-profile updates
- keep push outbox work durable until publication is actually confirmed
- retain unresolved local timeline rows without advancing read markers

Validation completed locally: workspace check and clippy (default plus OTLP), cgka-traits, marmot-account (all 50 runtime tests), marmot-app library (800 passed, 1 ignored), Nostr adapter SDK tests, the relay-runtime unresolved-send regression, push-registration regressions, and the UniFFI CompletionUnknown regression.

erskingardner:
Follow-up aa25b181 addresses the new review pass: live-group gating for pending-origin reads, durable Disband pending-kind preservation through restart, and fanout-local Welcome failure reasons. Added ambiguous_disband_recovery_preserves_terminal_pending_kind; all 51 account-runtime tests plus workspace check and clippy (default and OTLP) pass locally.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T17:52:23Z` |
| Commit reviewed | `aa25b1812e3c271997c6d9e4e856a3968eb98907` |
| Model | `Cursor Grok 4.6` |

Re-review of `aa25b181` after the two follow-up commits. The three blocking findings from `3758c2` are fixed, and the new recovery tests actually exercise the dangerous paths.

**Prior findings**

1. Create/invite pending no longer strands. The first Welcome (create) or commit (invite) now carries the pending ref plus frozen post-confirmation Welcomes. Resume can confirm, then `publish_welcome_fanout` releases the exact remaining Welcomes. Covered by `ambiguous_legacy_create_welcome_confirms_after_restart_with_exact_retry` and `ambiguous_invite_commit_recovery_publishes_its_frozen_welcome`.
2. App `CompletionUnknown` attaches by effects-list position, not engine vs transport id. The app-message test now injects a distinct transport event id.
3. `UnresolvedPublish` is emitted only when required acks are unmet. `accept_disposition_from_effects` prefers Published once any report has an accept, so a 1-of-N timeout no longer looks unpublished.

**This revision**

Disband kind is preserved through stage and hydrate (`pending_fanout_kind` + restore match), with `ambiguous_disband_recovery_preserves_terminal_pending_kind` checking the terminal convergence pass. `record_target_failure` no longer mutates exposure after a terminal target. `pending_origin_message_id` / `pending_fanout_kind` go through `pending_group_id` → `ensure_group_live`. Workspace CI is green, including the Rust test shards.

No new blocking issues. Residual note only: post-confirm Welcomes still travel the legacy `publish_welcome_fanout` path, so an ambiguous timeout *after* MLS confirm remains #352 redelivery rather than this new frozen-fanout classifier. That is outside the #1577 commit/app-message split and is acceptable here.

PR #1583, opened 2026-08-27
fix(adapter): normalize relay coverage identities

erskingardner:
Follow-up to #1581 addressing its two review threads.

## Summary

- normalize valid relay URL spellings at the device-local relay-index boundary
- make route snapshots, subscription attempts, inbound EOSE, and delivery timing share the same relay identity
- cover verbatim signed routes versus nostr-sdk-normalized inbound endpoints
- correct the replay completion documentation to describe frozen endpoint coverage

## Testing

- `cargo test -p transport-nostr-adapter`
- `cargo test -p marmot-app --features test-policy-overrides slow_relay -- --nocapture`
- `just fast-ci`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved relay endpoint handling by treating equivalent URL spellings consistently, including capitalization, default ports, and trailing slashes.
  - Preserved meaningful path-slash differences when canonicalizing relay endpoints.
  - Improved event delivery and completion coverage when equivalent endpoint formats are used.
- **Documentation**
  - Clarified that synchronization is complete when all endpoint subscriptions in the activation’s frozen route snapshot have completed.

### DIFF
--- a/crates/transport-nostr-adapter/src/telemetry.rs
+++ b/crates/transport-nostr-adapter/src/telemetry.rs
@@ -29,6 +29,7 @@ use std::collections::HashSet;
 
 use cgka_traits::MessageId;
 use cgka_traits::TransportEndpoint;
+use nostr::RelayUrl;
 use serde::{Deserialize, Serialize};
 
 /// Upper bounds, in milliseconds, of the duration histogram buckets shared by
@@ -74,13 +75,19 @@ pub struct RelayIndexRegistry {
 
 impl RelayIndexRegistry {
     /// Stable index for `endpoint`, assigning a new one on first sighting.
+    ///
+    /// Valid relay URLs are keyed by their parsed [`RelayUrl`] spelling so a
+    /// verbatim signed route and the normalized endpoint emitted by nostr-sdk
+    /// resolve to the same device-local relay. Invalid/non-relay endpoints stay
+    /// byte-exact; this registry never rewrites authoritative routing state.
     pub fn index_for(&mut self, endpoint: &TransportEndpoint) -> RelayIndex {
-        if let Some(index) = self.indices.get(endpoint) {
+        let key = relay_index_key(endpoint);
+        if let Some(index) = self.indices.get(&key) {
             return *index;
         }
         let index = RelayIndex(self.next);
         self.next += 1;
-        self.indices.insert(endpoint.clone(), index);
+        self.indices.insert(key, index);
         index
     }
 
@@ -106,6 +113,14 @@ impl RelayIndexRegistry {
         pairs.sort_by_key(|(index, _)| index.0);
         pairs
     }
+}
+
+/// Device-local identity key shared by subscription attempts, EOSE callbacks,
+/// delivery timing, and frozen replay coverage.
+fn relay_index_key(endpoint: &TransportEndpoint) -> TransportEndpoint {
+    RelayUrl::parse(endpoint.as_str())
+        .map(|relay_url| TransportEndpoint(relay_url.to_string()))
+        .unwrap_or_else(|_| endpoint.clone())
 }
 
 /// Capability that authorizes resolving opaque [`RelayIndex`] values back to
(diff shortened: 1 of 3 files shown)


crates/transport-nostr-adapter/src/telemetry.rs
```diff
@@ -108,6 +115,14 @@ impl RelayIndexRegistry {
     }
 }
 
+/// Device-local identity key shared by subscription attempts, EOSE callbacks,
+/// delivery timing, and frozen replay coverage.
+fn relay_index_key(endpoint: &TransportEndpoint) -> TransportEndpoint {
+    RelayUrl::parse(endpoint.as_str())
+        .map(|relay_url| TransportEndpoint(relay_url.to_string()))
```
erskingardner: `RelayUrl::to_string()` / `as_str()` preserves `has_trailing_slash`. `RelayUrl`'s `Eq`/`Hash` compare only the inner `url::Url` and ignore that flag — the same identity `CanonicalEndpoint::matches` already uses for delivery.

Both sides of this helper therefore still split `wss://relay.example.com` from `wss://relay.example.com/`. Delivery on that pair succeeds; `mark_eose` looks up a different `RelayIndex` and the frozen snapshot never completes. The nostr-sdk pool is also keyed by `RelayUrl` Eq, so inbound `relay_url.to_string()` is whichever slash the first `add_relay` stored.

Do not strip slashes with `as_str_without_trailing_slash()` — that would collide `wss://host/foo` with `wss://host/foo/`, which are distinct `Url`s. Store a parsed `RelayUrl` (or `Url::from(relay_url).to_string()`) as the map key so origin-only slashes fold and path-bearing URLs stay distinct. A unit test of those two origin spellings would have failed this helper.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T12:40:52Z` |
| Commit reviewed | `5e637664272f9e8c2c7dd783ff036c5741e45db7` |
| Model | `Cursor Grok 4.6` |

The choke point is right: every snapshot, subscribe, EOSE, and delivery path already goes through `RelayIndexRegistry::index_for`, and the `Complete` doc now matches the frozen-coverage contract. Host-case and default-port folding is the mismatch #1581 called out, and the tests pin it.

The leftover hole is that `relay_index_key` uses `RelayUrl` Display, not `RelayUrl` Eq. Display keeps `has_trailing_slash`; Eq — and `CanonicalEndpoint::matches` — ignore it. A signed `wss://relay.example.com/` and an inbound `wss://relay.example.com` still get two indices, delivery still succeeds, and the frozen gate never completes. nostr-sdk's pool also keys by Eq, so the first-inserted slash wins on notifications whenever directory, publish, or another group added the relay first.

Key the registry by parsed `RelayUrl` (or the inner `Url` serialization) so coverage identity matches routing, and pin `wss://host` vs `wss://host/` in the unit test. Merge should also wait for CI; most of the Rust matrix was still queued at review time.

Could not submit this as a GitHub request-changes: the API rejects that event on the author's own PR.

crates/transport-nostr-adapter/src/telemetry.rs
erskingardner: Addressed in b2a87c07. relay_index_key now serializes the inner nostr::Url, matching RelayUrl Eq/Hash: wss://host and wss://host/ share an index, while /foo and /foo/ remain distinct. The new unit regression failed against the prior helper before the fix, and the frozen-coverage integration regression now exercises a signed mixed-case/default-port/origin-slash route against a slashless normalized inbound endpoint. cargo test -p transport-nostr-adapter and just fast-ci pass.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T17:50:12Z` |
| Commit reviewed | `b2a87c07c0a8d21a9d75cffe5c18264fb5cb40dc` |
| Model | `Cursor Grok 4.6` |

The previous blocking gap is closed. `relay_index_key` now serializes the inner `Url`, which is the same identity `RelayUrl` Eq/`CanonicalEndpoint::matches` already use: origin slashes fold, `/foo` vs `/foo/` stay distinct, and host-case/default-port still collapse. The new unit test pins that split, and the frozen-coverage regression now sends a mixed-case/default-port/origin-slash signed route against a slashless inbound endpoint. CI is green.

Could not submit this as a GitHub approve: the API rejects self-approval on the author's own PR.

PR #1584, opened 2026-08-27
fix: reconcile relay history below cursor floor

### LINKED ISSUE #1579: marmot-app: cursor floor can permanently exclude old-authored offline messages
## Summary

The ordinary account subscription rebuild uses one monotonic transport cursor with a fixed 120-second lookback. Any event whose Nostr `created_at` is older than that floor is invisible to the fresh subscription.

That becomes a permanent-delivery problem for valid delayed traffic. Application events are authored before they enter the offline queue, and the outer kind-445 event deliberately inherits the authenticated inner `created_at`. If a message remains queued longer than the lookback while the recipient processes newer traffic and establishes a newer cursor, publication after connectivity returns can place the valid message below the recipient's existing floor. The recipient never sees it, so an ordinary application message provides no future-epoch/decrypt-failure evidence that could arm automatic full-history recovery.

The same shape applies to delayed relay replication and other valid old-timestamp group events.

## Failure sequence

1. Alice authors an application message while her account is inactive; its authenticated `created_at` is fixed.
2. The exact payload remains queued for longer than 120 seconds.
3. Bob processes newer account traffic and durably establishes a later transport cursor.
4. Alice reconnects and publishes the queued message using its original authored timestamp.
5. Bob later rebuilds subscriptions from `cursor - 120s`.
6. The relay correctly excludes Alice's now-below-floor event.
7. Because Bob never ingests the event, no epoch-stall trigger or other automatic recovery evidence exists. Repeated cold restarts use the same floor and continue missing it.

## Evidence on current `master`

Source snapshot: [`b0214579`](https://github.com/marmot-protocol/mdk/commit/b0214579f87ee8cc7636b0913f606feaa8a00b33).

- The characterization test [`cold_restart_since_floor_permanently_drops_backlog_below_it`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-app/tests/since_floor.rs#L1-L26) states that below-floor history is permanently unfetchable without an independently armed recovery; [the assertions](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-app/tests/since_floor.rs#L376-L410) confirm the miss across two cold restarts.
- Production fixes the rebuild lookback at [120 seconds](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-app/src/lib.rs#L390-L398).
- [`subscription_rebuild_since`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-app/src/relay_plane/mod.rs#L318-L345) derives the relay filter from the durable account cursor and lookback.
- The inner app event is authored before transport activation and is then [queued unchanged when the account is inactive](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/marmot-app/src/client/mod.rs#L2939-L2976).
- The engine carries the inner timestamp into [`GroupMessageMetadata::application`](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/cgka-engine/src/message_processor/send.rs#L874-L907), and the peeler [binds the outer kind-445 `created_at` to it](https://github.com/marmot-protocol/mdk/blob/b0214579f87ee8cc7636b0913f606feaa8a00b33/crates/transport-nostr-peeler/src/peeler.rs#L148-L163).

- #1561 and PR #1567 address queue-overflow-created gaps and unsafe cursor promotion around an observed overflow. They do not guarantee discovery of a valid event published later with a timestamp below an already legitimately established floor.
- #1560/#1562 own convergence/replay performance and worker fairness, not the subscription completeness contract.
- Manual full-history repair can recover this state only if the host already knows something is missing. An ordinary unseen application message supplies no such evidence.
- The future-skew protection from #182 must remain intact; this issue concerns valid old timestamps, not hostile future timestamps.

## Required behavior

Define a gap-safe cursor/completeness contract for valid delayed events. A sender's authenticated inner timestamp must retain its semantic meaning, but it must not make the transport event permanently undiscoverable solely because publication or replication happened after the receiver advanced an account-wide floor.

Possible solution families include separate transport publication ordering from the authenticated application timestamp, per-route durable coverage cursors, durable gap/range reconciliation, or periodic bounded widening. The selected design must preserve Nostr/Marmot interoperability and retention semantics and must not require an unbounded full-history replay on every startup.

Recovery must be automatic for ordinary unseen application messages; it cannot depend exclusively on epoch-stall evidence generated only after ingest.

## Acceptance criteria

- A relay-backed regression authors an application message, queues it offline longer than the 120-second lookback, advances and durably checkpoints the recipient's cursor with newer traffic, then publishes the queued message while the recipient is disconnected.
- On the recipient's next normal reconnect/catch-up, the old-authored message is delivered automatically without an explicit `repair_full_history` call.
- The test still passes after another cold restart and does not rely on undecryptable probes or reaching the epoch-stall threshold.
- A high-traffic group cannot advance an account-wide floor that permanently hides delayed valid traffic for another group or inbox route.
- Delayed relay replication after a cursor checkpoint receives equivalent coverage.
- Inner authenticated `created_at`, expiration/retention decisions, and user-visible authored time remain correct even if transport discovery uses a different ordering signal.
- Future-timestamp clamping and poisoned-cursor recovery remain covered.
- The design bounds replay/storage/network amplification and records privacy-safe aggregate widening/recovery telemetry.

erskingardner:
## Summary

- retain an exact, account-private per-route transport inventory and reconcile the history gap below the timestamp subscription floor with NIP-77
- keep ordinary subscriptions as the current-traffic fast path, scope shared-SDK replays to the compared account, and bound startup reconciliation to ten seconds
- add privacy-safe reconciliation counters plus cold-restart, durable-dedup, epoch-stall coexistence, route-isolation, and storage migration coverage

## Latest-master integration

- rebased onto `master` at `3ca56275`, including #1581, #1582, #1583, #1588, and #1567
- preserve #1567's durable per-account overflow recovery: a pending overflow still forces an unfloored EOSE-confirmed replay, and the NIP-77 below-floor pass runs only when a legitimate completed-drain floor exists
- preserve #1588's ambiguous-publish relay reset and stale-socket regression unchanged
- use storage migration `0054`, after #1567 claimed `0053` for account delivery recovery
- establish the regression's floor with a newly drained offline anchor, so the test continues to exercise late replication below a valid checkpoint rather than the live-cursor case fixed by #1567

## Test plan

- [x] RED before the original fix: `cargo test -p marmot-app --test since_floor cold_restart_reconciles_backlog_below_since_floor -- --exact --nocapture` failed because boot 2 delivered 3 events instead of 4
- [x] `cargo test -p marmot-app --test since_floor -- --nocapture`
- [x] `cargo test -p marmot-app --lib` — 807 passed, 1 ignored
- [x] `cargo test -p marmot-app --features test-policy-overrides --lib overflow -- --nocapture` — 7 passed
- [x] `cargo test -p transport-nostr-adapter --features sdk` — 73 unit and 39 integration tests passed
- [x] `cargo test -p storage-sqlite` — 420 passed, 4 ignored
- [x] `just check clippy`
- [ ] `just fast-ci` — blocked before Rust checks by the existing macOS BSD `sed -i` incompatibility in unchanged `scripts/tests/test_campaign_toolchain.sh`

Fixes #1579

## Summary by CodeRabbit

* **New Features**
  * Added automatic, bounded transport-history reconciliation during startup.
  * Recovers missed inbox and group messages, including events older than the normal sync boundary.
  * Persists route-specific checkpoints to prevent unnecessary redownloading.
  * Added account-scoped event routing and reconciliation metrics.

* **Bug Fixes**
  * Prevented reconciled group events from reaching the wrong account.
  * Improved recovery when previously received events cannot be decrypted.
  * Cleans up reconciliation data when routes or groups are removed.

* **Tests**
  * Added coverage for durable recovery, deduplication, route rotation, and account isolation.


crates/marmot-app/src/client/sync.rs
```diff
@@ -1814,9 +1937,49 @@ impl AppClient {
...
+        let reconciliation_item = <[u8; 32]>::try_from(delivery.message.id.as_slice())
+            .ok()
+            .map(|event_id| TransportReconciliationItem {
+                event_id,
+                created_at: outer_transport_at,
+            });
         let effects = self.runtime.ingest_delivery(delivery).await?;
         let publish_error = fail_if_publish_failed(&effects.effects).err();
         let must_stay_fetchable = effects.left_object_unpersisted;
+        if !must_stay_fetchable
```
erskingardner: **Blocking: events skipped at the drain seams never reach this record, so they are re-downloaded on every reconciliation pass.**

This record only fires inside `ingest_delivery`, but both drain seams skip deliveries before ingest: the catch-up drain skips `seen_events_index` hits and own relay echoes (`is_own_relay_echo`, sync.rs:1691-1692), and the live receive path does the same (sync.rs:1215-1220). Two consequences:

1. **Own published group messages.** The group route filter matches kind + `h`-tag only (no author restriction), so this account's own below-floor messages are in the relay's remote set. They enter the seen ring via `remember_published_reports` on publish, are skipped as own echoes at the drain, and never get recorded here — so the account's own below-floor publish history is re-downloaded from the relay on *every* reconciliation pass for as long as the relay retains it and the ids stay in the 16k ring.
2. **Pre-upgrade events.** Below-floor events that were durably ingested before this PR (still inside the seen ring at migration time) are downloaded by the first post-upgrade pass, skipped at the drain, and therefore never enter the inventory — re-downloaded forever after.

This keeps `reconciliation_remote_items`/`reconciliation_received_items` permanently nonzero (undermining those diagnostics) and violates #1579's "bounds replay/storage/network amplification" criterion. Both skipped classes are known-durable by construction (own publishes are retained locally; seen-index entries were durably ingested under the same rule as line 1759), so the straightforward fix is to also `record_transport_reconciliation_item` at the two skip seams — or record own published event ids at publish time.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T21:15:51Z` |
| Commit reviewed | `e0d825309bd3f0ecb5a95612a0edf61a7af25afc` |

The core design is right: per-route exact event-id sets compared via NIP-77 below the timestamp floor, recording at the `left_object_unpersisted` seam (which correctly keeps resource-refused input fetchable), account-scoped replay routing with a regression test, privacy-safe aggregate counters, and a clean `0054` migration. The RED-first test plan and test documentation are excellent.

However, two issues should be fixed before merge because they violate #1579's bounded-amplification acceptance criterion and can re-introduce the "one busy route hides another" failure the issue calls out:

1. **Perpetual re-download for events skipped at the drain seams** (blocking). Own published group messages and pre-upgrade below-floor events still in the 16k seen ring are skipped before `ingest_delivery` ever runs, so they are never recorded in the inventory and are re-downloaded from the relay on every reconciliation pass. See inline comment on `sync.rs`.
2. **Route starvation under the account-wide 10s quantum** (blocking). Routes reconcile sequentially with an inner per-route timeout equal to the entire outer budget, and the pass restarts from the top every sync — one slow or noisy first route permanently starves all later routes. See inline comment on `sync.rs`.

Non-blocking notes: first-pass post-migration bootstrap downloads each route's entire below-floor history (worth documenting or seeding), and a route-level failure is double-counted across the two telemetry layers.

Minor coverage gaps versus the issue's acceptance criteria: the regression uses synthetic backdated kind-445s rather than a genuinely offline-queued application message (same transport gap, but the peeler-bound inner-`created_at` authoring path is not covered end-to-end), and there is no explicit busy-group-vs-other-route floor test even though the per-route design covers it.

crates/marmot-app/src/client/sync.rs
```diff
@@ -503,6 +521,86 @@ impl AppClient {
...
+                    relays_failed += summary.relays_failed;
+                    remote_items += summary.remote_items;
+                    received_items += summary.received_items;
+                }
+                Ok(None) => {}
+                Err(_) => relays_failed += 1,
+            }
+        }
+
+        for route in routing.group_routes {
```
erskingardner: **Blocking: one slow route starves every later route, on every pass.**

Routes are reconciled sequentially (inbox first, then `group_routes` in snapshot order) under a single account-wide 10s quantum, and the inner per-route timeout (`SDK_RECONCILIATION_WAIT`, also 10s) equals the whole outer budget. So a single slow or non-NIP-77 route can consume the entire quantum, the outer timeout aborts the remaining routes, and the next sync restarts from the top in the same order — routes after a persistently slow route are *never* reconciled.

Combined with the skip-seam gap above (a first busy route re-downloading its full below-floor set every pass), this re-creates exactly the "one busy route hides another route's traffic" failure #1579's acceptance criteria call out. Options: split the quantum across pending routes (per-route budget), rotate the starting route index between passes, or persist per-route reconciliation progress so completed routes are skipped cheaply on retry.

crates/storage-sqlite/src/migrations/0054_transport_reconciliation_items.rs
```diff
@@ -0,0 +1,24 @@
+use crate::SqliteResultExt;
+use cgka_traits::storage::StorageResult;
+use rusqlite::Transaction;
+
+pub(crate) fn apply(tx: &Transaction<'_>) -> StorageResult<()> {
```
erskingardner, line 5: Non-blocking: after this migration the inventory starts empty, so the first reconciliation pass advertises an empty set per route and downloads the route's entire below-floor history from the relay (one-time, bounded by relay retention, amortized across boots by the 10s quantum). Worth documenting as an explicit upgrade cost, or seeding the inventory from durable already-seen event state here so established accounts don't pay a full below-floor replay on first boot.

crates/marmot-app/src/client/sync.rs
```diff
@@ -503,6 +521,86 @@ impl AppClient {
...
+                .await
+            {
+                Ok(Some(summary)) => {
+                    relays_succeeded += summary.relays_succeeded;
+                    relays_failed += summary.relays_failed;
+                    remote_items += summary.remote_items;
+                    received_items += summary.received_items;
+                }
+                Ok(None) => {}
+                Err(_) => relays_failed += 1,
```
erskingardner: Nit: a route-level error is counted as one relay failure here, and the same failure is additionally recorded as a fabricated `relays_failed: 1` summary inside `reconcile_inbox_history`/`reconcile_group_history` (relay_plane/mod.rs), so the log line and `NostrAdapterMetrics` diverge for the same event. Consider counting the failure in only one layer.

crates/transport-nostr-adapter/src/sdk_client.rs
```diff
@@ -50,6 +50,27 @@ const SDK_RELAY_BATCH_OVERALL_WAIT: Duration = Duration::from_secs(60);
 /// fan-outs. Four covers the generated-account bootstrap cohort while keeping
 /// larger batches backpressured.
 const SDK_RELAY_BATCH_MAX_IN_FLIGHT: usize = 4;
+/// A reconciliation pass is a correctness backstop, but it must not own the
+/// serial account worker indefinitely when a relay is slow or advertises a
+/// large first-run difference. Partial downloads remain in the SDK database
+/// and the durable app inventory records them as they ingest, so a later pass
+/// resumes with a smaller set difference.
```
erskingardner: **Blocking:** this resumability claim does not hold for the timeout path, and I believe it produces a permanent livelock for large first-run differences.

`sync_targeted` only yields its `Reconciliation` when the negentropy exchange fully completes. If `SDK_RECONCILIATION_WAIT` fires first — or the outer `TRANSPORT_RECONCILIATION_QUANTUM` in `sync_inner`, which wraps the same work with the *same* 10s budget, leaving no room for materialization + routing even when the inner wait succeeds — the partial output is discarded. Nothing is materialized, routed, or ingested, so the durable inventory is unchanged and the next pass computes the *identical* set difference. The SDK's download phase (`req_neg_events`, sequential batches of 100 ids, one REQ/EOSE round trip at a time) never consults its database to skip already-cached ids, and the database here is a per-process `MemoryDatabase` (`Client::builder().build()` in the relay plane), so a restart discards even that cache.

Net effect: a route whose below-floor difference cannot finish in ~10s (order of >10k events at 100ms RTT — realistic for long-lived agent routes, and the state every existing account is in the first time it boots migration `0054` with an empty inventory) re-downloads its full difference for ~10s on **every** startup, never converges, and delays each boot's drain by the full quantum. That contradicts #1579's "bounds replay/storage/network amplification" criterion for the cohort that needs this fix most. Durable progress currently exists only at whole-route granularity.

Directions (not prescriptive): chunk the below-floor range into time windows reconciled newest-first so each completed window durably shrinks the next pass's difference; materialize/route downloaded events incrementally rather than only after full completion; decouple the two equal 10s budgets; persist the SDK database; or, if large differences are deemed out of scope, bound the attempt and surface a loud, permanent telemetry signal instead of silently re-paying the cost forever.

crates/storage-sqlite/src/transport_reconciliation.rs
```diff
@@ -0,0 +1,143 @@
+//! Durable, account-private NIP-77 set-reconciliation inventory.
```
erskingardner, line 1: Non-blocking design question: this inventory grows by one row per retained event, forever. There is no eviction for deleted/left groups (orphaned rows), retention sweeps don't prune it, and each startup loads the full per-route set into memory to build the negentropy item list — so the table, the boot-time read, and the reconciliation payload all grow linearly with lifetime route traffic (~73 B/row plus the secondary index; a 1M-event route is 100+ MB and a proportionally large item vec on every sync).

Exact-set reconciliation inherently requires advertising the full set, so this may be an accepted trade-off — but #1579 lists bounded storage amplification as an acceptance criterion, so the growth bound (or the deliberate absence of one) should be stated in the architecture docs, and a route-scoped delete on group removal would at least reclaim churn from deleted groups.

crates/marmot-app/src/client/sync.rs
erskingardner: Nit: `Err(_) => relays_failed += 1` counts a route-level failure — the 10s wait elapsing, a non-NIP-77 relay failing the 2s negotiation probe, or sanitization rejecting before any relay was contacted — as one failed *relay* in the only aggregate log line. A route with several endpoints, or a timeout that is no relay's fault, will make `relays_failed` misleading in diagnostics. Consider a distinct `routes_failed` counter (same in the group arm below).

Related: a non-NIP-77 relay burns the 2s negotiation timeout per route per boot, forever, out of the shared 10s quantum. Counters make this visible, but it may be worth remembering unsupported relays for the process lifetime.

crates/marmot-app/tests/since_floor.rs
```diff
@@ -273,7 +273,7 @@ async fn publish_garbage_group_message_at(
 }
 
 #[tokio::test]
-async fn cold_restart_keeps_backlog_fetchable_until_a_drain_promotes_the_floor() {
+async fn cold_restart_reconciles_backlog_below_since_floor() {
```
erskingardner, line 276: Nit: the renames leave stale references in `docs/marmot-architecture/convergence-reliability-plan.md` — line 206 and line 336 still cite `stalled_epoch_backfill_recovers_below_floor_backlog_when_armed` and `cold_restart_since_floor_permanently_drops_backlog_below_it`. Line 336 was already stale from the earlier rename, but since this PR renames exactly those tests, could you refresh both rows? The repo invariant asks for Rust test names to stay greppable across layers.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T21:24:42Z` |
| Commit reviewed | `e0d825309bd3f0ecb5a95612a0edf61a7af25afc` |

The design is right: a per-route exact event-id inventory with a NIP-77 below-floor pass closes the #1579 completeness gap without giving up the cheap timestamp fast path, and the contract holds up under close reading — no uncovered bands (ordinary `since` and the reconciliation `until` tile the timeline with only a dedup-able overlap), account-scoped routing fixes the shared-SDK-database masking hole with a regression test, `left_object_unpersisted` correctly stays out of the advertised set, dial-safety sanitization is preserved on the reconcile path, telemetry is privacy-safe aggregates, and migration `0054` sequencing vs. #1567 is handled properly. CI is green.

One blocking concern: the reconciliation pass is not actually resumable on timeout, and with the outer quantum equal to the inner wait, a route whose below-floor difference exceeds ~10 seconds of downloads (the exact state every existing high-volume account is in on first boot after migration `0054`) re-downloads its full difference on every startup and never converges — the opposite of the bounded-amplification criterion this issue was judged against. Details and fix directions inline on `sdk_client.rs`.

Non-blocking notes inline: unbounded inventory growth deserves a documented bound (and ideally a prune on group deletion), the route-vs-relay failure counters will mislead diagnostics, and the test renames leave stale greppable references in `convergence-reliability-plan.md`.

Datawav (commented):
## Automated review

Actionable findings: 4. The exact head has three blocking reconciliation-progress defects and one storage-lifecycle defect.

1. Blocking — Drain-skipped durable events never enter the reconciliation inventory
- Exact changed-line evidence: `crates/marmot-app/src/client/sync.rs:1960` records `TransportReconciliationItem` only after `ingest_delivery`; the existing catch-up and live drain seams skip own relay echoes and `seen_events_index` hits before reaching that changed record site (`sync.rs:1691-1692` and `sync.rs:1215-1220`).
- Triggering state: an own-published group event or pre-upgrade durable event is below the subscription floor but already present in the seen index.
- Consequence: reconciliation downloads the event, the drain skips it, and subsequent passes continue advertising an incomplete local set and downloading the same event.
- Confidence: High.
- Smallest sound correction: record validated route/event metadata at both skip seams, and persist own-published IDs when publication becomes durable. Seed migration `0054` from durable transport history where that history can be mapped to a route.

2. Blocking — Timed-out reconciliation makes no durable partial progress
- Exact changed-line evidence: `crates/transport-nostr-adapter/src/sdk_client.rs:57` sets the reconciliation wait to ten seconds, while the changed reconciliation path around `sdk_client.rs:326` materializes remote IDs only after `sync_targeted` completes successfully. `crates/marmot-app/src/client/sync.rs:565` wraps the entire multi-route operation in an equal ten-second account quantum.
- Triggering state: a below-floor difference, relay response, or post-sync materialization cannot complete within the inner deadline, or the inner exchange consumes nearly the entire outer deadline.
- Consequence: the reconciliation result is discarded before delivery and inventory recording; the next pass computes and downloads the same difference. The in-memory SDK database does not make this restart-resumable.
- Confidence: High.
- Smallest sound correction: reconcile bounded ID/time chunks and durably ingest each completed chunk before continuing; make the inner deadline strictly smaller than the outer quantum and ensure the next pass excludes already-applied IDs.

3. Blocking — Fixed inbox-first sequencing can permanently starve later routes
- Exact changed-line evidence: the new route loop around `crates/marmot-app/src/client/sync.rs:565` reconciles the inbox and then group routes sequentially, while each route may consume the same ten seconds allotted to the complete account pass.
- Triggering state: the inbox or an early group route is persistently slow, unsupported, or has a difference too large to complete within one pass.
- Consequence: the outer timeout cancels every later route; each subsequent pass restarts from the same first route, so later routes may never receive reconciliation.
- Confidence: High.
- Smallest sound correction: persist or rotate the next-route cursor and cap each route to a fair share of the remaining account quantum.

4. Minor — Inventory storage and route reads grow without a lifecycle bound
- Exact changed-line evidence: `crates/storage-sqlite/src/transport_reconciliation.rs:99` loads the full route inventory, while the new `INSERT OR IGNORE` inventory has no pruning operation. `crates/storage-sqlite/src/migrations/0054_transport_reconciliation_items.rs:5-22` creates the table and route-order index without a foreign-key cascade, retention marker, or group-removal cleanup.
- Triggering state: a long-lived high-volume route, or groups repeatedly joined and later removed.
- Consequence: database size, startup reads, reconciliation input, and orphaned group-route rows grow for the account’s lifetime.
- Confidence: High for unbounded/orphaned growth; medium for the acceptable active-route retention policy because exact-set reconciliation may intentionally retain the full relay window.
- Smallest sound correction: delete a group’s inventory atomically when that group is removed and document/enforce an explicit active-route retention contract that preserves reconciliation completeness.

Audit evidence:
- Exact head SHA: `e0d825309bd3f0ecb5a95612a0edf61a7af25afc`
- Changed files inspected include `crates/marmot-app/src/client/sync.rs`, `crates/transport-nostr-adapter/src/sdk_client.rs`, `crates/storage-sqlite/src/transport_reconciliation.rs`, `crates/storage-sqlite/src/migrations/0054_transport_reconciliation_items.rs`, and `crates/marmot-app/tests/since_floor.rs`.
- Discussion classification: the drain-seam, non-resumable-timeout, fixed-order starvation, and unbounded-inventory reports are supported. The post-sync work-bound report duplicates the timeout/progress defect. The first-upgrade full replay report is a consequence of the empty migration and missing inventory convergence rather than a separate finding. Counter-label and stale-documentation comments are non-behavioral or partly pre-existing and were discarded.
- Risk lenses checked: persistence/migration — prior-state upgrade, partial failure/idempotent retry, and data preservation; network/protocol — timeouts, retries, duplicate delivery, malformed or unsupported peers, and wire compatibility; tests — pre-fix failure behavior, boundary/negative cases, and whether tests exercise production behavior. Existing tests cover small successful reconciliation and account-scoped routing but do not exercise timeout resumability, fixed-order starvation, drain-skipped events, or long-history growth.
- Check evidence: the bundle reports all checks completed successfully, including Rust tests (all four shards), Rust check, Rust clippy, Rust format, Rust audit, relay runtime tests, simulator and vectors, CLI relay tests, and convergence-policy assertions. Those successful checks do not cover the triggering states above.

crates/marmot-app/src/client/sync.rs
erskingardner: Fixed in f235826c. Both live and catch-up skip seams now record route-scoped reconciliation items for durable seen-index hits and own relay echoes, using validated account/welcome/group route mapping. This prevents those IDs from being downloaded again on every pass.

crates/marmot-app/src/client/sync.rs
erskingardner: Fixed in f235826c. Route-level errors now increment routes_failed rather than relays_failed. Explicit unsupported-NIP-77 responses are cached per relay for the process lifetime so subsequent route attempts skip them instead of repeatedly spending the negotiation budget.

crates/storage-sqlite/src/migrations/0054_transport_reconciliation_items.rs
erskingardner: Documented and bounded in f235826c. The migration intentionally starts empty, but first-run discovery is limited to the rolling 30-day/16,384-item route window and downloads at most 128 deterministic IDs per route pass. The upgrade cost and resumability contract are now explicit in the architecture docs.

crates/transport-nostr-adapter/src/sdk_client.rs
erskingardner: Fixed in f235826c. The adapter first performs a bounded dry-run set comparison, then explicitly fetches only a deterministic 128-ID subset under the same two-second deadline. Each fetched event enters ordinary durable ingestion, shrinking the next comparison even when the full difference is large.

crates/storage-sqlite/src/transport_reconciliation.rs
```diff
@@ -0,0 +1,143 @@
...
+                    StorageError::Serialization(
+                        "invalid transport reconciliation timestamp".to_owned(),
+                    )
+                })?;
+                Ok(TransportReconciliationItem {
+                    event_id,
+                    created_at,
+                })
+            })
+            .collect()
```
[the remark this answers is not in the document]
erskingardner: Fixed in f235826c. Route inventories now have a persisted completeness floor, a rolling 30-day retention boundary, and a hard 16,384-item cap. Recording and compaction are atomic, the SQL read and relay filter share the floor, and the long-history regression proves both the retained inventory and reconciliation input stay bounded.

crates/marmot-app/tests/since_floor.rs
erskingardner: Fixed in f235826c. Both stale architecture references now use the current cold-restart and armed-backfill test names, so the documented evidence remains greppable.

crates/marmot-app/src/client/sync.rs
erskingardner: Fixed in f235826c. Relay-plane metrics no longer fabricate relays_failed=1 for a route-level error. The app aggregation reports route failures separately as routes_failed, while actual per-relay failures remain sourced from successful reconciliation summaries.

crates/storage-sqlite/src/transport_reconciliation.rs
erskingardner: Fixed in f235826c. Inventory and boot-time input are capped per route, old timestamp buckets advance a persisted floor, and the rolling lower bound is 30 days. Route and group deletion now atomically remove inventory, floor state, and a matching scheduler cursor, with cleanup and rollback tests.

crates/marmot-app/src/client/sync.rs
erskingardner: Fixed in f235826c. A pass attempts at most four routes with a two-second route budget and a persisted rotating cursor. The cursor advances before network I/O, so a slow or crashing route cannot pin later routes; cursor rotation and persistence have regression coverage.

crates/storage-sqlite/src/transport_reconciliation.rs
```diff
@@ -0,0 +1,449 @@
...
+                "transport reconciliation retention floor exceeds SQLite range".to_owned(),
+            )
+        })?;
+        let (route_kind, route_id) = route.storage_key();
+        let mut conn = self.connection.lock()?;
+        let tx = conn
+            .transaction_with_behavior(TransactionBehavior::Immediate)
+            .storage()?;
+        let inventory_since = route_state_floor_tx(&tx, route_kind, route_id, configured_floor)?;
+        compact_route_tx(&tx, route_kind, route_id)?;
```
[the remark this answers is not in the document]
erskingardner: Fixed in c5d0db1b. The complete Immediate inventory transaction, including floor maintenance, compaction, item loading, and commit, now runs inside retry_on_busy. The targeted storage reconciliation tests pass.

erskingardner:
Final review and validation update for `c5d0db1b`:

- rebased onto current `master` (`aee5b99f`)
- addressed every verified review finding; all review threads are resolved
- bounded and resumable route reconciliation, persistent route fairness, skip-seam durability, route lifecycle cleanup, accurate metrics, busy retries, and privacy-safe SDK errors are covered
- targeted storage, adapter, and cold-restart reconciliation tests pass locally
- exact-head GitHub CI is fully green, including all four Rust test shards, relay runtime tests, simulator/vectors, format, check, clippy, audit, Tamarin, CLI E2E, and platform builds

The PR is cleanly mergeable.

erskingardner:
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-28T09:28:31Z` |
| Commit reviewed | `c5d0db1b1a94f33f9cd0258c95a7d3ee8134f9a1` |

Re-review of the two follow-up commits (`f235826`, `c5d0db1`). Both prior blocking findings are properly resolved:

- **Skip-seam gap — fixed.** `record_durable_transport_reconciliation_delivery` now records at all three skip sites (own-echo, seen-index, and the combined catch-up drain skip), with a unit test pinning the route-hint validation. Own published events also converge: reconciliation's dry-run materializes them from the SDK cache, routes them through the account queue, and the echo skip now records them, so the perpetual re-download loop is closed on every path.
- **Route starvation — fixed.** A durable rotating route cursor (`transport_reconciliation_scheduler`) advances before network I/O, the pass is capped at four routes with a 2s per-route deadline inside the 10s quantum, and rotation is unit-tested including wrap-around. A slow or non-NIP-77 route can no longer pin later routes.
- **Amplification — bounded and documented.** 30-day rolling retention floor, 16,384-id per-route ceiling with bucket-granularity compaction advancing a persisted `inventory_since`, dry-run comparison, and a deterministic 128-id fetch batch per pass. The first-start-empty behavior and the finite repair contract are documented in `relay-delivery-telemetry.md`.
- **Telemetry double-count — fixed** (`routes_failed` separated; error metrics no longer fabricated).

Also verified: group/route retirement deletes reconciliation rows inside the caller's transaction with a rollback regression test (matches the torn-write-free invariant), and the static error strings avoid leaking relay details. CI is fully green.

Two non-blocking nits (would have been inline, but pending draft reviews on this PR block new review comments):

- `crates/transport-nostr-adapter/src/sdk_client.rs:64` — `SDK_RECONCILIATION_SET_LIMIT` must stay in lockstep with `TRANSPORT_RECONCILIATION_MAX_ITEMS_PER_ROUTE` in `storage-sqlite`, but the constants live in crates that don't depend on each other. The comment documents the coupling; if a shared constants home ever becomes natural, hoist it so drift is a compile error rather than a silent mismatch.
- `crates/transport-nostr-adapter/src/sdk_client.rs:310` — matching negentropy rejection by exact `nostr-sdk` error strings is brittle across SDK upgrades: a reworded message would silently stop caching unsupported relays (correctness preserved; only the per-pass retry cost returns). Prefer a typed variant if the SDK ever exposes one.

crates/transport-nostr-adapter/src/sdk_client.rs
```diff
@@ -195,6 +232,170 @@ impl NostrSdkRelayClient {
...
+        })?
+        .map_err(|_| {
+            TransportAdapterError::Subscription("NIP-77 reconciliation failed".to_owned())
+        })?;
+        let newly_unsupported = output
+            .failed
+            .iter()
+            .filter(|(_, error)| {
+                error.as_str() == "negentropy not supported"
+                    || error.as_str() == "unsupported negentropy protocol version"
```
erskingardner, line 311: The second literal never matches: `nostr-relay-pool`'s `Error::Negentropy(negentropy::Error::UnsupportedProtocolVersion)` formats as `"server does not support our negentropy protocol version"` (nostr-relay-pool 0.44.3 formats `Negentropy(e)` via `e.fmt(f)`, delegating to the negentropy error's Display). `NegentropyNotSupported` formats as `"negentropy not supported"`, so only the first variant gets cached; relays rejected by the protocol-version path (or by NEG-ERR) silently burn the 500ms negotiation wait per route per pass, forever. Alternatives: match the actual `UnsupportedProtocolVersion` Display string, or match on both `contains("negentropy")`/`contains("NEG-")` prefixes, or track failures by a small enum rather than a string equality.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-28T09:36:34Z` |
| Commit reviewed | `c5d0db1b1a94f33f9cd0258c95a7d3ee8134f9a1` |

Re-reviewed at `c5d0db1b` with commits `f235826c` (bound and resume) and `c5d0db1b` (retries/errors). The first-round blocking livelock is properly fixed: the SDK now dry-runs the comparison and deterministically fetches at most 128 remote ids per route pass under a strict 2s per-route deadline, with the durable route cursor advancing before network I/O so a timed-out route can no longer pin the rest. My non-blocking nits are addressed too — inventory is bounded by a 30-day retention window plus a 16384-entry per-route cap with a persisted compaction floor, removal paths for group/route/scheduler state are wrapped in transactions, `routes_failed` is now a distinct counter, unsupported relays are negative-cached, and the convergence-reliability plan references the renamed tests. The new bound is documented in `relay-delivery-telemetry.md`, and own relay echo / seen-index duplicates are now recorded so they cannot linger in the diff forever.

One remaining issue (a true defect but deliberately bounded in cost): the negative-cache classifier matches on a string literal that does not match the SDK's formatting for the protocol-version variant, so that variant of "relay cannot do NIP-77" never gets cached. An alternative match direction and a suggestion would be inline on `sdk_client.rs:311`.

Validate the review head once more before merge; I recommend merging once this minor issue is cleared.

PR #1587, opened 2026-08-27
feat(terminal-harness): deliver ordered inbound attachment batches

### LINKED ISSUE #1400: Pass multiple inbound files to terminal connectors in one turn
## Summary

Allow Marmot terminal-harness connectors to receive **multiple files from one inbound message and pass the complete ordered batch to one backend turn**.

Today `agent-control` projects every inbound media reference, but `marmot-terminal-harness` reduces an inbound event to text-only `InboundPrompt` and its `Invocation` has no attachment field. `wn-opencode` and `wn-pi` therefore cannot preserve an inbound multi-file turn through the shared harness. Future `wn-codex` and `wn-claude` adapters in #813 would inherit the same limitation unless this is solved once in the shared runtime.

A user should be able to attach, for example, a screenshot, a log, and a source file to one Marmot message and have the coding agent receive all three—ordered and associated with the same prompt/session turn—not one file per turn and not several independently scheduled prompts.

Verified against `master` commit [`a82524ca0816c67c7ffdfe642f0b2a454d897ce3`](https://github.com/marmot-protocol/mdk/commit/a82524ca0816c67c7ffdfe642f0b2a454d897ce3).

## Current implementation and gap

- [`AgentControlMessage.media`](https://github.com/marmot-protocol/mdk/blob/a82524ca0816c67c7ffdfe642f0b2a454d897ce3/crates/agent-control/src/lib.rs#L533-L545) is already an ordered `Vec<AgentControlMediaRef>`; the control protocol does not fundamentally limit an inbound message to one attachment.
- [`media_refs_from_tags()`](https://github.com/marmot-protocol/mdk/blob/a82524ca0816c67c7ffdfe642f0b2a454d897ce3/crates/agent-connector/src/event_projection.rs#L242-L264) projects every structurally valid `imeta` tag, and [`project_durable_record()`](https://github.com/marmot-protocol/mdk/blob/a82524ca0816c67c7ffdfe642f0b2a454d897ce3/crates/agent-connector/src/event_projection.rs#L460-L487) places that vector on the one inbound event.
- `agent-control` exposes one-reference-at-a-time [`DownloadMedia`](https://github.com/marmot-protocol/mdk/blob/a82524ca0816c67c7ffdfe642f0b2a454d897ce3/crates/agent-control/src/lib.rs#L278-L285), returning a host-local path and metadata in [`MediaDownloaded`](https://github.com/marmot-protocol/mdk/blob/a82524ca0816c67c7ffdfe642f0b2a454d897ce3/crates/agent-control/src/lib.rs#L429-L437). A consumer can therefore download each reference in deterministic order without moving plaintext bytes or keys over the control protocol.
- The shared harness currently stores only `{account, group, message, text}` in [`InboundPrompt`](https://github.com/marmot-protocol/mdk/blob/a82524ca0816c67c7ffdfe642f0b2a454d897ce3/integrations/terminal-harness/src/bridge.rs#L307-L312). It drops the event's media vector before dispatch.
- [`Invocation`](https://github.com/marmot-protocol/mdk/blob/a82524ca0816c67c7ffdfe642f0b2a454d897ce3/integrations/terminal-harness/src/lib.rs#L80-L89) carries cwd, session ID, and prompt only. There is no backend-neutral ordered attachment input.
- `wn-opencode` and `wn-pi` explicitly document that media handling is out of scope today ([OpenCode README](https://github.com/marmot-protocol/mdk/blob/a82524ca0816c67c7ffdfe642f0b2a454d897ce3/integrations/opencode/marmot/README.md#L1-L11), [Pi README](https://github.com/marmot-protocol/mdk/blob/a82524ca0816c67c7ffdfe642f0b2a454d897ce3/integrations/pi/marmot/README.md#L1-L7)).

This is an inbound terminal-harness capability gap. It is distinct from closed #1320, which implemented **outbound** Hermes multi-image batches as one Marmot message.

### Shared harness owns the batch lifecycle

Extend the backend-neutral invocation boundary with an ordered attachment list containing only validated local presentation data, for example:

- owner-only host-local path;
- sanitized display filename;
- normalized media type;
- size in bytes;
- stable ordinal from the inbound message.

For one authorized inbound message:

1. Preserve the `AgentControlMessage.media` order.
2. Reject the whole turn before backend launch if the attachment count exceeds a documented connector limit.
3. Download and authenticate every accepted reference through `DownloadMedia`.
4. Treat each returned `wn-agent` path as borrowed: it is content-addressed and may be shared by concurrent downloads or another connector consumer. Copy each validated regular file into a unique owner-only harness batch directory, then pass only those harness-owned copies to the backend. Never delete or mutate the borrowed `wn-agent` path; its existing TTL lifecycle remains authoritative.
5. If any item fails validation/download/copy, do not invoke the backend with a partial batch; remove the harness-owned batch and return one privacy-safe error identifying the failed ordinal without exposing paths, keys, hashes, or remote locators.
6. Invoke the backend exactly once with the prompt plus the full ordered vector of harness-owned copies.
7. Clean the unique harness batch on success, backend error, timeout, cancellation, queue rejection after preparation, and connector shutdown/restart reconciliation.

The shared runtime should own count/aggregate-byte limits, ordered download, unique private copying, all-or-nothing dispatch, cleanup of its own batch, and privacy-safe diagnostics. Backend adapters should only map the validated vector to their documented CLI input surface.

### Backend adapters own CLI mapping

- **OpenCode:** use only a documented noninteractive multi-file input mechanism supported by the pinned/tested CLI version. Preserve order and pass paths as argv/data—not shell interpolation. If stock OpenCode cannot accept multiple files in one run, return an explicit unsupported-capability error rather than silently selecting the first file or splitting the turn.
- **Pi:** follow the same rule for its documented machine interface. Preserve current text-only behavior for messages without attachments.
- **Codex/Claude (#813):** consume this shared invocation shape when their adapters land, with backend-specific argv/event fixtures. They must not create parallel download/staging implementations.

Do not infer semantics from MIME type beyond safe presentation and CLI capability checks. Files are untrusted input. The selected coding CLI and its sandbox/approval policy remain responsible for interpreting contents.

## Limits and security

- Define a conservative V1 maximum attachment count and aggregate plaintext byte limit before downloading. Prefer shared configurable limits with safe defaults; do not inherit the outbound 512 MiB batch ceiling automatically for coding-agent input.
- Enforce the limit again against actual downloaded sizes because sender metadata is untrusted.
- Require regular owner-readable files returned by `wn-agent`; reject symlinks and path substitution before backend launch.
- Validate each borrowed `wn-agent` path without following symlinks, then copy it into a newly created owner-only per-turn directory under connector-owned private temp storage. Pass canonical paths to those copies directly to the child process.
- Do not remove, rename, chmod, or otherwise mutate `MediaDownloaded.local_path`: identical plaintext can resolve to the same content-addressed path for concurrent consumers. Only the uniquely owned per-turn copies are eligible for harness cleanup.
- Never log local paths, filenames, prompts, attachment contents, hashes, locators, account/group/message IDs, or backend output.
- Bind attachment preparation to the existing per-group FIFO turn: duplicate/replayed inbound events must not launch duplicate backend turns or leak duplicate temp batches.
- Preserve sender authorization and the documented remote-code-execution trust boundary: attachments do not broaden who may prompt the connector.

### Shared runtime

- [ ] `InboundPrompt` and `Invocation` preserve all inbound media refs/validated attachments in original order.
- [ ] A text plus N files produces exactly one backend invocation and one session continuation; it is never split into N prompts.
- [ ] Zero-file turns remain byte-for-byte compatible with existing text-only backend fixtures.
- [ ] Count and aggregate-byte overflow reject before backend launch with privacy-safe diagnostics.
- [ ] Failure of item N causes all-or-nothing turn failure and cleans the harness-owned copies for items `0..N-1`; no partial backend invocation occurs and no borrowed `wn-agent` path is removed.
- [ ] Concurrent turns that download the same plaintext receive isolated harness-owned copies; cleanup of either turn cannot invalidate the other turn or another connector consumer.
- [ ] Success, backend failure, timeout, cancellation, queue rejection, restart recovery, and duplicate replay have deterministic cleanup/idempotency coverage for unique harness batch directories.
- [ ] Downloaded inputs are regular, non-symlink, owner-confined files and are never shell-interpolated or logged.

### Backend conformance

- [ ] `cargo test -p marmot-terminal-harness`, `cargo test -p wn-opencode`, and `cargo test -p wn-pi` cover ordered zero/one/many-file invocations and unsupported-capability behavior.
- [ ] `just opencode-dev-e2e-connector` and `just pi-dev-e2e-connector` use fake backends to assert that three files arrive in one ordered turn and are cleaned afterward.
- [ ] OpenCode and Pi READMEs document supported CLI versions, accepted file/media behavior, limits, cleanup, and explicit errors.
- [ ] #813's Codex and Claude adapters reuse the same shared fixtures when attachment support is enabled.

Named regression targets:

- `integrations/terminal-harness/src/bridge.rs` unit tests for ordered download, atomic failure, limits, replay, and cleanup;
- `integrations/opencode/marmot/tests/e2e_connector.rs`;
- `integrations/pi/marmot/tests/e2e_connector.rs`;
- `crates/agent-connector/src/tests.rs` only if download lifecycle or validation changes are required below the existing control API.

## Related work

- #792 — durable reusable session lanes; multi-file input must remain one turn within the active lane.
- #813 — future Codex and Claude Code adapters should consume this shared capability.
- #1226 — open decrypted-media retention issue for the Hermes adapter; relevant lifecycle precedent, not a duplicate of terminal-harness inbound batching.
- #1320 (closed) — outbound Hermes multi-image batching; confirms the protocol can preserve a batch, but addresses the opposite direction and a different consumer.

## Priority

Low priority (`LOW`): important connector usability and parity, but not ahead of current correctness/security work. Once implemented, all-or-nothing dispatch, plaintext cleanup, limits, and replay behavior are release gates.

Datawav:
## Summary

- Preserve every inbound media reference in source order and deliver the complete attachment batch to one terminal-backend turn.
- Stage downloaded files as owner-private harness copies, reject partial or oversized batches before backend launch, and clean owned staging across success, failure, cancellation, replay, and restart paths.
- Expose backend capability mapping so unsupported attachment batches fail closed while zero-file behavior remains compatible.
- Document shared staging and pre-spawn rejection for Codex, OpenCode, and Pi connectors.

## Verification

- 123 `marmot-terminal-harness` unit tests passed.
- 14 terminal-harness JSONL tests passed.
- 19 `wn-codex` tests passed; the authenticated real-model contract test remained explicitly ignored.
- `just fast-ci` passed on the exact final local tree.
- Exact-head GitHub Actions completed with 26 passing checks and only the non-release publish job skipped.
- Every Datawav commit added during review follow-up is GPG-signed and GitHub reports `verification.verified=true`.
- Independent Cursor Grok 4.6 exact-diff review passed with no findings.

Closes [ordered inbound attachment batches #1400](https://github.com/marmot-protocol/mdk/issues/1400)

## Summary by CodeRabbit

* **New Features**
  * Added ordered inbound image attachments for Codex turns.
  * Added attachment validation, count and size limits, private staging, cleanup, and retry support.
  * Added typed artifact exports with reliable delivery and retry handling.
  * Added configuration for maximum attachment count and aggregate size.
  * Unsupported attachments are rejected before backend launch; OpenCode and Pi reject attachment batches.

* **Documentation**
  * Updated integration documentation to describe attachment behavior and backend limitations.


Datawav (commented):
## Automated review

1 actionable finding.

1. Major — Attachment preparation failure leaves `/retry-last` recovery stuck in `Retrying`
   - Exact changed-line evidence: `integrations/terminal-harness/src/bridge.rs:652-690` calls `begin_validated_retry`, which atomically changes the recovery status from `Pending` to `Retrying`, then restores the saved media through `inbound.media = record.media.clone()`. Lines `755-775` subsequently call `prepare_attachment_batch`; every error path sends an attachment-failure reply and returns without calling `reset_retry`.
   - Triggering state: a saved recovery contains attachments, the user sends `/retry-last`, and attachment download, aggregate-size validation, source-file validation, or private staging fails.
   - Consequence: `RecoveryStore::begin_retry` accepts only `Pending` records (`integrations/terminal-harness/src/store.rs:231-245`), while `fifo_is_blocked` treats any recovery record as blocking (`bridge.rs:1435-1439`). Further `/retry-last` attempts therefore report no retryable recovery and ordinary FIFO work remains blocked until `/discard-last` or process restart resets the state.
   - Confidence: High.
   - Smallest sound correction: before returning from the `prepare_attachment_batch` error arm, restore a retrying record to `Pending` through the existing bounded recovery-restoration behavior. Add a regression test that begins an attachment-bearing retry, forces download or staging failure, and verifies the record is `Pending` and can be retried again.

Discussion classification:
- The staging-root cleanup claim is not actionable at this head: `bridge.rs:1495-1507` already preserves non-directories and entries whose direct-child names do not start with `batch-`.
- The associated review summary duplicates that stale cleanup claim.
- The general automated summary contains no additional concrete changed-line finding; no organization-member feedback was present.

Audit evidence:
- Exact head: `1b03534fc5631169236daf3faa2be1b84316d256`
- Changed files examined include `integrations/terminal-harness/src/bridge.rs`, `integrations/terminal-harness/src/store.rs`, `integrations/terminal-harness/src/control.rs`, and `integrations/codex/marmot/src/codex.rs`.
- The attachment lifecycle, recovery-state transitions, queue barrier, startup reconciliation, backend capability mapping, configuration limits, and relevant tests were re-derived from the immutable patch and bounded source snapshots.

Risk lenses checked:
- `release/build`: untrusted-input and least-privilege CI; artifact identity/reproducibility; all required paths trigger checks.
- Behavioral reasoning also covered product behavior, security and private staging, durability, retry/FIFO lifecycle, agent-control protocol boundaries, backend compatibility, and regression-test coverage.

Check evidence:
- The bundle reports `all_completed: true` and `all_successful: true`.
- Successful checks include all four Rust test shards, Rust check, clippy, format, audit, terminal-harness installer tests, and Linux/Darwin terminal-harness builds.
- Those checks do not cover the identified transition from `Retrying` through an attachment-preparation early return.

integrations/terminal-harness/src/lib.rs
```diff
@@ -214,6 +249,24 @@ pub trait Backend: Send + Sync + 'static {
...
+        &self,
+        invocation: Invocation,
+        attachments: Vec<Attachment>,
+        tx: mpsc::Sender<RunnerEvent>,
+    ) -> std::result::Result<Outcome, RunFailure> {
+        if attachments.is_empty() {
+            return self.run(invocation, tx).await;
+        }
+        Err(RunFailure {
+            error: HarnessError::AttachmentUnsupported,
```
erskingardner, line 266: Non-blocking design note (not gating): backends that do not opt in (OpenCode, Pi) still pay a full download plus private staging pass in prepare_attachment_batch and then reject at this dispatch boundary. Fail-closed at this seam is the right default and matches the updated docs; if the wasted staging round-trip ever matters, a capability predicate consulted before staging would avoid it. Safe either way.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-28T07:12:06Z` |
| Commit reviewed | `ff75459f70b308598df577683c9908c490ff523e` |

## Review summary

This PR implements issue #1400's shared-harness inbound attachment batching: ordered media preservation, count/aggregate-byte limits, private per-turn staging with all-or-nothing rejection before backend launch, deterministic TempDir-lease cleanup (lease moved into the backend task, plus an explicit abort guard for cancellation), retryable-recovery media persistence, and scoped startup reconciliation. I reviewed the linked issue, the full 3-commit diff, and the raw files at the head SHA.

**Prior findings verified resolved at this head:**
- CodeRabbit's reconciliation claim is stale: at head, `reconcile_attachment_staging` (bridge.rs:1503-1517) preserves non-directories and entries not prefixed `batch-`.
- The retry-stuck `Retrying` transition from the earlier automated review is fixed in PATCH 3 via `reset_retry`, with a regression test (`retry_attachment_download_failure_restores_pending_recovery`). The state transition logic reads correctly.

**Checked against the acceptance criteria:** ordered preservation in one invocation, zero-file turns delegated to `Backend::run` (documented byte-compatible fallback), pre-download count limit, aggregate-byte limit checked with overflow protection, regular-file/symlink checks (`symlink_metadata` + `O_NOFOLLOW`), bounded copy with extra-byte detection, owner-private copies via `fs_private`, privacy-safe diagnostics (aggregate kinds/counts only). The CodeRabbit "linked issues" pre-merge warning is a false issue — OpenCode/Pi rejection-before-spawn is the explicitly permitted unsupported-capability fallback from the issue's proposed contract, and the docs were updated to match.

Non-blocking notes:
- One inline design note on the `run_with_attachments` default: non-opted-in backends download and stage a batch before rejecting at the dispatch boundary. Fail-closed is correct; a pre-staging capability predicate could avoid the round-trip if it ever matters.

Tests cover the new behavior (ordered args mapping, name sanitization, limits without overflow, symlink/FIFO rejection, private modes, abort guard, retry-fix regression). Recommend merge once CI completes.

integrations/terminal-harness/src/bridge.rs
```diff
@@ -1393,6 +1446,187 @@ async fn fifo_is_blocked(ctx: &BridgeContext, group_ref: &str) -> bool {
...
+        };
+        if !name.starts_with("batch-") || !entry.file_type()?.is_dir() {
+            continue;
+        }
+        fs::remove_dir_all(entry.path())?;
+    }
+    Ok(())
+}
+
+async fn prepare_attachment_batch(
```
erskingardner, line 1519: **Blocking: the successful attachment path is never exercised through the bridge.**

The individual pieces are unit-tested (staging, name sanitization, limits, symlink/fifo rejection, restart reconcile, download-failure retry restore, default trait rejection), but no test drives `handle_message` with a fake control server that returns `MediaDownloaded` success and asserts:

- exactly one backend invocation carrying the ordered staged paths;
- the staged files exist with mode 0600 at invocation time;
- the batch directory is cleaned after the turn completes.

`RecordingBackend` doesn't override `run_with_attachments`, and the Codex/OpenCode/Pi e2e suites contain no attachment flows, so the composition — the headline feature — is untested. A regression that passes an empty/reordered vector or drops the batch early would pass CI.

Suggested additions:

1. A bridge test with a fake `wn-agent` serving `MediaDownloaded` (two ordered files) and a backend recording `run_with_attachments`, asserting order, existence, privacy mode, and cleanup on completion.
2. A bridge test where the default (unsupported) rejection flows through `handle_message` and produces the documented "does not support this attachment batch" reply with no recovery record persisted — that is the path every OpenCode/Pi attachment turn takes.
3. A fake-backend e2e case in the Codex connector asserting the staged paths arrive as ordered `--image` arguments in one turn.

These mirror the acceptance criteria in #1400 ("a text plus N files produces exactly one backend invocation", e2e assertions in the connector suites).

integrations/codex/marmot/src/codex.rs
```diff
@@ -209,6 +255,10 @@ fn build_exec_args(session_id: Option<&str>, profile: ExecutionProfile) -> Vec<S
             args.push("--dangerously-bypass-approvals-and-sandbox".to_owned());
         }
     }
+    for image in images {
+        args.push("--image".to_owned());
+        args.push(image.to_string_lossy().into_owned());
```
erskingardner, line 260: Question (non-blocking): has `codex exec resume --image <path>` been verified against the pinned Codex CLI version used in the e2e fakes/real runs? The test here asserts argv construction only. If `exec resume` doesn't accept `--image` in that version, every resumed attachment turn fails at runtime while the unit suite stays green. If it has been verified manually, a one-line note in the README's attachment section ("works for new and resumed turns against codex vX") would capture it.

integrations/terminal-harness/src/bridge.rs
```diff
@@ -1393,6 +1446,187 @@ async fn fifo_is_blocked(ctx: &BridgeContext, group_ref: &str) -> bool {
...
+fn owner_can_read(_metadata: &fs::Metadata) -> bool {
+    true
+}
+
+fn attachment_failure_reply(config: &Config, error: &HarnessError) -> String {
+    let detail = match error {
+        HarnessError::AttachmentCountLimit => "too many attachments",
+        HarnessError::AttachmentBytesLimit => "the attachment batch is too large",
+        HarnessError::AttachmentUnsupported => "one or more attachment types are unsupported",
+        _ => "an attachment could not be downloaded and validated",
```
erskingardner, line 1622: Nit: this `AttachmentUnsupported` arm is unreachable from this call site — `prepare_attachment_batch` only returns count/bytes/invalid/control/io errors. `AttachmentUnsupported` surfaces from the backend run and gets its own message in `handle_backend_run_failure`. Consider removing the arm (or leaving a comment) so readers don't assume prepare can reject on type grounds.

erskingardner (dismissed):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-28T07:16:29Z` |
| Commit reviewed | `ff75459f70b308598df577683c9908c490ff523e` |

## Summary

The implementation matches issue #1400's contract well: ordered preservation of `AgentControlMessage.media`, pre-download count limits, aggregate-byte limits re-checked against actual downloaded sizes, borrowed-path discipline (never mutating the `wn-agent` path), `O_NOFOLLOW`/`O_NONBLOCK` race-hardened staging with owner-read checks, all-or-nothing dispatch, TempDir-in-task staging lease with abort-on-drop, and startup reconciliation restricted to `batch-` directories (CodeRabbit's earlier finding is correctly addressed at this head). Privacy discipline in logs and replies looks right. Unit coverage of the individual pieces is solid.

One blocking issue before merge:

1. **The happy path has no integration coverage.** No test drives `handle_message` through a successful `DownloadMedia` exchange into a backend invocation. `RecordingBackend` does not override `run_with_attachments`, and none of the three connectors' e2e suites exercise attachments at all — so a wiring regression (empty, reordered, or missing attachment vector, or staging cleaned before the backend reads the files) would pass CI today. Issue #1400's acceptance criteria explicitly call for these e2e assertions. See inline comment on `prepare_attachment_batch`.

Non-blocking:

2. Question: has `codex exec resume --image <path>` been verified against the pinned Codex CLI version? The unit test only asserts argv construction; if `resume` rejects `--image`, every resumed attachment turn fails at runtime with no test catching it.
3. Nit: the `AttachmentUnsupported` arm in `attachment_failure_reply` is unreachable — `prepare_attachment_batch` never returns that variant (only the backend run can, and that path has its own message in `handle_backend_run_failure`). Harmless, but slightly misleading.

integrations/terminal-harness/src/bridge.rs
Datawav: Addressed at [`0ee68dda`](https://github.com/marmot-protocol/mdk/commit/0ee68dda2a6d5ae75321d8d9a1cfdae988e80a82).

The bridge now has a two-file successful-composition regression that proves one attachment backend invocation, stable ordered sanitized paths under the staging root, regular non-symlink 0600 copies during invocation, exact contents, cleanup after completion, and no recovery record. A second bridge regression drives the default unsupported backend with an existing resumable session and requires the exact documented final reply, zero backend runs, and no recovery record. The Codex process regression requires one resumed invocation with both `--image` paths in order.

Verified on the exact pushed tree: terminal harness 123 passed; Codex 19 passed with one authenticated real-model test ignored; `just fast-ci` passed; independent Cursor Grok 4.6 exact-diff review returned PASS with no findings. The pushed commit is GitHub-verified.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-28T09:26:37Z` |
| Commit reviewed | `0ee68dda2a6d5ae75321d8d9a1cfdae988e80a82` |

The blocking gap is resolved. `0ee68dda` adds exactly the composition coverage requested:

- `successful_attachment_turn_preserves_order_privacy_and_cleanup` drives `handle_message` through a fake `wn-agent` serving two ordered `MediaDownloaded` responses and proves one backend invocation, stable ordered sanitized staging paths under the connector root, regular non-symlink 0600 copies at invocation time (from 0644 sources), exact contents, exactly two downloads, cleanup after completion, and no recovery record.
- `unsupported_attachment_backend_replies_without_persisting_recovery` proves the default rejection path every OpenCode/Pi attachment turn takes: exact documented reply, zero backend runs, no recovery record, session mapping untouched.
- `resumed_runner_passes_ordered_images_to_one_codex_turn` spawns a fake Codex asserting the full argv (`exec resume --image <first> --image <second> --json thread-123 -`), stdin prompt transport, and event parsing through `run_jsonl_process`.

The `spawn_attachment_server` helper also fails on missing or unexpected download requests, which catches both under- and over-download regressions.

Remaining non-blocking notes (fine as follow-ups):

- `codex exec resume --image` verification against the pinned real CLI (the new test covers argv construction against a fake; the authenticated real-model test is ignored in CI).
- The unreachable `AttachmentUnsupported` arm in `attachment_failure_reply`.

Full CI is green at this head.

PR #1588, opened 2026-08-27
Fix stale relay sockets after ambiguous publish timeout

### LINKED ISSUE #926: Relay publish pool does not recover after connectivity loss until the runtime is restarted
## Observed

With a running `MarmotAppRuntime`, drop connectivity (airplane mode), attempt a text send (it fails as expected), then restore connectivity. Every subsequent publish — new `send_text` calls and `retry_group_convergence` — keeps failing indefinitely while the runtime stays alive. Only a full runtime shutdown + rebuild (which the iOS client happens to do on background/foreground) recovers publishing.

`catch_up_accounts` does not recover it either: it pumps the account workers but does not re-establish the relay publish connections.

## Expected

After connectivity returns, the next publish attempt should reconnect and succeed — `publish_to_relay` already has per-endpoint add/connect logic, so the intent appears to be connect-on-demand, but in practice the pool stays dead (possibly a stale relay entry in a long reconnect backoff that on-demand connect does not reset).

## Repro sketch

1. Start the runtime, open a group, verify sends work.
2. Kill connectivity at the OS level; `send_text` fails (correct).
3. Restore connectivity; wait well past any reasonable reconnect window.
4. `send_text` and `retry_group_convergence` still fail; `catch_up_accounts` before the attempt does not help.
5. Shut down and rebuild the runtime: sends work immediately.

## Impact

Any client that keeps the runtime alive across a connectivity loss cannot send again until an app restart or background/foreground cycle. It blocks a failed-send retry UX: the retry action re-publishes into the dead pool no matter how long it waits.

## Notes

The subscription side has explicit reconnect backoff in the account worker; the publish side appears to have no equivalent recovery path. A `reconnect`/`reset_relay_pool` command on the runtime would also suffice as an explicit lever for clients that detect network transitions themselves.

erskingardner:
## Summary
- invalidate SDK relay sockets after acknowledgement-unknown publish failures so a stale iOS `Connected` status cannot pin retries to a zombie WebSocket
- preserve relay registrations and subscriptions while allowing the existing frozen byte-identical retry to reconnect through a fresh socket
- keep explicit relay rejections and definite pre-send failures out of the reset path

## Test plan
- [x] New stale-socket regression failed before the fix and passes after
- [x] `cargo test -p transport-nostr-adapter --features sdk`
- [x] `cargo test -p transport-nostr-adapter`
- [x] Remaining `just fast-ci` gates passed individually, including workspace checks and both Clippy feature sets
- [ ] `just fast-ci` (blocked on current `master` by GNU-style `sed -i` usage under macOS BSD `sed` in `scripts/tests/test_campaign_toolchain.sh`)

Fixes #926

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery after ambiguous publish failures by resetting affected relay connections.
  * Prevented stale relay state from impacting subsequent publishes.
  * Preserved cleanup ordering to ensure reliable completion handling.

* **Tests**
  * Added coverage for recovering from stale relay connections during publishing.

### DIFF
--- a/crates/transport-nostr-adapter/src/sdk_client.rs
+++ b/crates/transport-nostr-adapter/src/sdk_client.rs
@@ -580,20 +580,14 @@ impl NostrSdkRelayClient {
         let deadline = tokio::time::Instant::now() + SDK_RELAY_PUBLISH_OVERALL_WAIT;
         let mut aborted_publishes = false;
         let mut timed_out = false;
-        let result = loop {
+        let (accepted, failed, timed_out) = loop {
             let remaining = deadline.saturating_duration_since(tokio::time::Instant::now());
             if remaining.is_zero() {
                 publishes.abort_all();
                 aborted_publishes = true;
                 timed_out = true;
                 append_missing_publish_failures(&mut failed, &accepted, &attempted_endpoints);
-                break Self::finish_publish_outcome(
-                    message_id,
-                    accepted,
-                    failed,
-                    request.required_acks,
-                    timed_out,
-                );
+                break (accepted, failed, timed_out);
             }
             match timeout(remaining, publishes.join_next()).await {
                 Err(_) => {
@@ -601,23 +595,11 @@ impl NostrSdkRelayClient {
                     aborted_publishes = true;
                     timed_out = true;
                     append_missing_publish_failures(&mut failed, &accepted, &attempted_endpoints);
-                    break Self::finish_publish_outcome(
-                        message_id,
-                        accepted,
-                        failed,
-                        request.required_acks,
-                        timed_out,
-                    );
+                    break (accepted, failed, timed_out);
                 }
                 Ok(None) => {
                     append_missing_publish_failures(&mut failed, &accepted, &attempted_endpoints);
-                    break Self::finish_publish_outcome(
-                        message_id,
-                        accepted,
-                        failed,
-                        request.required_acks,
-                        timed_out,
-                    );
+                    break (accepted, failed, timed_out);
                 }
                 Ok(Some(result)) => match result {
                     Ok(Ok(receipt)) => {
@@ -625,13 +607,7 @@ impl NostrSdkRelayClient {
                         if accepted.len() >= ack_goal {
                             publishes.abort_all();
                             aborted_publishes = true;
-                            break Self::finish_publish_outcome(
-                                message_id,
-                                accepted,
-                                failed,
-                                request.required_acks,
-                                timed_out,
-                            );
+                            break (accepted, failed, timed_out);
                         }
                     }
                     Ok(Err(failure)) => failed.push(failure),
@@ -644,7 +620,14 @@ impl NostrSdkRelayClient {
         if aborted_publishes {
             while publishes.join_next().await.is_some() {}
         }
-        result
+        self.reset_ambiguous_publish_relays(&failed).await;
+        Self::finish_publish_outcome(
+            message_id,
+            accepted,
+            failed,
+            request.required_acks,
+            timed_out,
+        )
     }
 
     async fn publish_prepared_single(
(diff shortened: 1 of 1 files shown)


erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T19:08:28Z` |
| Commit reviewed | `337f3378614d44b66a8699e36f7507e00e0783cf` |
| Model | `Cursor Grok 4.6` |

GitHub blocks self-approval on this PR, so this is a comment review rather than an approve event. Recommendation is still merge.

`disconnect_relay` is the right lever for #926. nostr-sdk's `connect_relay` is a no-op unless the relay is `Initialized` / `Terminated` / `Sleeping`, so an iOS zombie that stays `Connected` can never recover through on-demand connect. Moving the relay to `Terminated` keeps the pool entry and subscription map, and the next `connect_before_send` path actually opens a new socket. `terminate()` is a local `Notify`, so recovery does not depend on the dead TCP producing a disconnect event. Pool-level `Shutdown` is not broadcast (`external: false`), so the inbound notification forwarder is left alone.

The `PossiblyExposed` gate matches the type: acknowledgement unknown. `RetryableUnavailable` (pre-send connect/add failures) and `TerminalRejected` stay out. Quorum abort still does not call `append_missing_publish_failures`, so a healthy relay that lost the race is not torn down.

The new test is the right regression: durable relay, silent first socket, timeout, `terminated == 1`, then a byte-identical retry on a fresh connection. Author reports it failed before the reset and passed after.

Non-blocking:

- A durable-relay `TerminalRejected` / connect-timeout case would lock the “do not reset explicit rejections and pre-send failures” claim; the filter is obvious, the test is not there.
- `stale_connection_closed.notified().await` has no timeout. If `close_ws` ever failed to unblock the mock, this test would hang the job rather than fail.
- CI was still in progress at review time; rely on `transport-nostr-adapter --features sdk` plus the new test going green.

crates/transport-nostr-adapter/src/sdk_client.rs
```diff
@@ -911,6 +894,34 @@ impl NostrSdkRelayClient {
...
+    /// Invalidate sockets whose publish result cannot establish whether the
+    /// relay accepted the event. On iOS a network transition can leave an
+    /// established WebSocket silently dead while the SDK still reports it as
+    /// connected; a later `connect_relay` is then intentionally a no-op. Moving
+    /// that relay to `Terminated` preserves its registration and subscriptions
+    /// while ensuring the next publish starts a fresh connection (mdk#926).
+    async fn reset_ambiguous_publish_relays(&self, failures: &[TransportEndpointFailure]) {
+        let mut reset = HashSet::new();
+        for failure in failures {
+            if failure.kind != TransportEndpointFailureKind::PossiblyExposed {
```
erskingardner, line 906: Non-blocking: `PossiblyExposed` also covers the `error:` machine-readable rejection path in `relay_rejection_endpoint_failure`, where the relay demonstrably answered on this very socket — so a proven-live socket gets bounced too, slightly broader than the PR description's "keep explicit relay rejections out of the reset path". It's harmless (registration/subscriptions survive, next publish reconnects) and arguably correct since the SDK also reports local send failures as `error:`, but a one-line comment here noting that intentional breadth would make the scope explicit.

crates/transport-nostr-adapter/src/sdk_client.rs
```diff
@@ -2046,6 +2057,61 @@ mod tests {
...
+            .await
+            .expect("first publish task must not panic")
+            .expect_err("the stale connection must miss the acknowledgement deadline");
+
+        let health = sdk.relay_health().await;
+        assert_eq!(
+            health.terminated, 1,
+            "an ambiguous timeout must invalidate the SDK's stale Connected status"
+        );
+        stale_connection_closed.notified().await;
```
erskingardner, line 2104: Minor test-robustness nit: this wait has no timeout, so if the close is never observed the test hangs until the harness kills it instead of failing with a useful message. Consider moving it after `tokio::time::resume()` below and wrapping it in `tokio::time::timeout`. (The ordering itself is safe — `Notify::notify_one` stores a permit, so an early notify is not lost.)

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T19:20:26Z` |
| Commit reviewed | `337f3378614d44b66a8699e36f7507e00e0783cf` |

*(Posted as a comment because GitHub does not allow approving your own PR from this account; the recommendation is merge.)*

## Summary

Tight, well-scoped fix for #926. After a publish ends with ambiguous (`PossiblyExposed`) per-endpoint failures, the adapter now disconnects the affected SDK relays so a stale `Connected` status cannot pin the byte-identical retry to a zombie socket; registration and subscriptions survive, and the existing `connect_publish_relay` on the retry path builds the fresh connection.

Verified the mechanism against the pinned `nostr-relay-pool` 0.44.3: `disconnect()` sets `RelayStatus::Terminated` idempotently and preserves the pool registration, and `Relay::connect()` proceeds from `Terminated` (`can_connect` = Initialized | Terminated | Sleeping), so `connect_relay` on the next publish really does reconnect. As a bonus this also clears the long-backoff (`Sleeping`) variant hypothesized in the issue, since a reset relay reconnects immediately instead of waiting out the sleep.

What I liked:

- The loop refactor collapses four duplicated `finish_publish_outcome` exits into one tail call, giving the reset a single unambiguous hook point — and it runs after the abort-drain, so no in-flight send future can race the disconnect.
- Scope discipline: `TerminalRejected` and `RetryableUnavailable` stay out of the reset path; the new `tracing::warn!` carries no endpoint URL and uses `target`/`method`, matching the crate privacy invariant.
- The regression test faithfully models the iOS zombie socket (first connection swallows EVENT without OK, second acknowledges) using the file's established paused-time relay-harness pattern.

Two non-blocking notes are inline (an `error:`-rejection scope nuance, and a timeout-less wait in the new test).

One tradeoff I considered and accept: in the batch path, up to four in-flight events share a pre-connected socket, so one event's ambiguous failure resets the socket under the others. On a genuinely dead socket they were doomed anyway; on a merely slow relay it amplifies in-batch failures. Given the 12s per-attempt timeout, the simpler shared implementation is the right call.

CI is green at the reviewed head; only "Simulator and vectors" was still pending, and this change does not touch simulator or vector surfaces.

PR #1589, opened 2026-08-27
fix(app): wake reconnect backoff on catch-up

### LINKED ISSUE #1546: Allow connectivity-restored signals to interrupt account-worker reconnect backoff
## Observed behavior

On a user-verified GrapheneOS device, White Noise Android is already open while the phone is offline. After Wi‑Fi/connectivity is restored, the UI can remain disconnected for a long time before it changes to **Connecting** and finally establishes relay/service connectivity.

This is a recovery-latency defect, not a claim that every Android network transition fails. The exact phase duration still needs privacy-safe instrumentation, but current source confirms a host recovery signal can be discarded while MDK is in reconnect backoff.

## Source-confirmed path

At White Noise Android `dff546076f9ba72eedfdd57e72e4a84e8ecdcb58`:

- The app registers a default-network callback and updates its active-network snapshot on `onAvailable` / `onCapabilitiesChanged` ([`AppState.kt` lines 6841–6889](https://github.com/marmot-protocol/whitenoise-android/blob/dff546076f9ba72eedfdd57e72e4a84e8ecdcb58/app/src/main/java/dev/ipf/whitenoise/android/state/AppState.kt#L6841-L6889)).
- A genuine offline → online edge immediately schedules notification/runtime recovery and calls account catch-up after ensuring the receiver is active ([`AppState.kt` lines 6946–6989](https://github.com/marmot-protocol/whitenoise-android/blob/dff546076f9ba72eedfdd57e72e4a84e8ecdcb58/app/src/main/java/dev/ipf/whitenoise/android/state/AppState.kt#L6946-L6989)).

At MDK `2d225ed1ecbbca339998638f659df5d609ae9935`:

- Subscription connection lifecycle is delegated to `nostr-sdk` background retry tasks ([`sdk_client.rs` lines 958–974](https://github.com/marmot-protocol/mdk/blob/2d225ed1ecbbca339998638f659df5d609ae9935/crates/transport-nostr-adapter/src/sdk_client.rs#L958-L974)).
- After a receive failure, the account worker sleeps on an exponential reconnect backoff. During that sleep it actively drains the command channel and drops every command, so the Android-side `catch_up_accounts()` recovery request cannot shorten the wait ([`account_worker.rs` lines 1288–1328](https://github.com/marmot-protocol/mdk/blob/2d225ed1ecbbca339998638f659df5d609ae9935/crates/marmot-app/src/runtime/account_worker.rs#L1288-L1328)).
- The reconnect delay grows from 2 seconds to 60 seconds, plus up to 500 ms jitter ([`lib.rs` lines 334–336](https://github.com/marmot-protocol/mdk/blob/2d225ed1ecbbca339998638f659df5d609ae9935/crates/marmot-app/src/lib.rs#L334-L336)).

That creates a deterministic lost-wake window: Android observes restored connectivity and asks MDK to catch up, but an account worker already sleeping in reconnect backoff drops the request and waits for its pre-existing timer. The source proves the lost wake and its upper bound; a device trace is still needed to prove it is the only contributor to the observed delay.

## Expected behavior

When a host with a live runtime reports restored connectivity, MDK should preserve offline backoff while the network is unavailable but wake the affected account workers promptly once connectivity returns. Recovery must not require rebuilding the runtime or waiting out a stale maximum-backoff timer.

## Acceptance criteria

- A host-triggered catch-up/recovery request can interrupt an account worker's reconnect sleep instead of being dropped.
- Reconnect remains bounded and non-spinning during a sustained outage; repeated recovery signals are idempotent and do not create duplicate workers or subscriptions.
- A deterministic fake-clock integration test drives an account worker to the maximum reconnect delay, sends a recovery signal immediately after network restoration, and proves the next reconnect attempt starts within 3 seconds rather than waiting up to 60.5 seconds.
- A production-shaped runtime test proves inbound subscription/service connectivity recovers without runtime reconstruction and that catch-up resumes only after the receiver/subscription is ready.
- Cancellation, runtime shutdown, deferred hydration, and command-channel backpressure remain covered.
- Privacy-safe phase telemetry records only recovery phase/outcome and elapsed time—never account IDs, relay URLs, pubkeys, messages, or payloads.

## Related work and ownership

- This is a recurrence/remaining gap adjacent to closed [MDK #785](https://github.com/marmot-protocol/mdk/issues/785): that issue made reconnect backoff responsive to shutdown and prevented command-channel wedging, but the current implementation resolves responsiveness by dropping commands and still has no connectivity-restored wake.
- Closed [MDK #661](https://github.com/marmot-protocol/mdk/issues/661) introduced the bounded exponential backoff policy; this issue preserves that policy while adding a safe wake edge.
- Open [MDK #926](https://github.com/marmot-protocol/mdk/issues/926) owns the distinct outbound publish-pool failure that can persist until runtime restart. Keep it separate unless implementation evidence proves one shared recovery primitive fixes both paths.
- Closed [White Noise Android #1579](https://github.com/marmot-protocol/whitenoise-android/issues/1579) and merged [PR #1615](https://github.com/marmot-protocol/whitenoise-android/pull/1615) added the Android offline → online catch-up trigger. The remaining lost wake is in MDK's account-worker reconnect lifecycle, so no speculative Android issue is needed.

This should remain a standalone MDK bug. No open native Tracking issue has a completion rule that depends on it.

erskingardner:
## Summary
- Preserve a host catch-up command during account-worker reconnect backoff and use it to interrupt only the stale sleep.
- Keep failed reopen attempts on the existing bounded exponential backoff, coalesce repeated recovery signals, and record reconnect/catch-up activation and subscription phases with privacy-safe telemetry.
- Add a deterministic paused-clock runtime regression that reaches the 60-second cap and proves recovery starts within 3 seconds without rebuilding the runtime.

## Test plan
- [x] cargo test -p marmot-app reconnect
- [x] cargo test -p marmot-app notification_recovery
- [x] cargo test -p marmot-app app_telemetry::tests
- [x] Workspace check/clippy gates with and without OTLP features
- [x] Install docs, SHA-256, cargo-audit policy, and convergence policy gates
- [ ] cargo test -p marmot-app: 808 unit tests passed; five unrelated relay_runtime tests fail individually on this branch, and a representative failure reproduces unchanged on origin/master
- [ ] just fast-ci: blocked by the unchanged campaign-toolchain test using GNU-style sed -i on macOS BSD sed

Fixes #1546

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved account recovery during connectivity interruptions by responding promptly to recovery signals, even during extended reconnect delays.
  * Preserved bounded retry behavior while preventing unnecessary waiting before reconnection.

* **Telemetry**
  * Added stage-aware tracking for startup, catch-up synchronization, and reconnect transport preparation.

* **Tests**
  * Expanded coverage for recovery timing, reconnect activity, coalesced refresh signals, and synchronization telemetry.

### DIFF
--- a/crates/marmot-app/src/runtime/account_worker.rs
+++ b/crates/marmot-app/src/runtime/account_worker.rs
@@ -689,7 +689,7 @@ async fn run_app_runtime_account_worker(
     let startup_sync_result = {
         let mut initial_sync = std::pin::pin!(async {
             let summary = client
-                .sync_with_startup_stage_telemetry(&startup_stage_telemetry)
+                .sync_with_stage_telemetry(&startup_stage_telemetry)
                 .await?;
             app.finish_client_open_network_maintenance(&mut client)
                 .await;
@@ -1327,6 +1327,7 @@ async fn run_app_runtime_account_worker(
                         // prolonged transport outage.
                         drop(client);
                         client = loop {
+                            let retry_started_at = Instant::now();
                             let mut retry_delay =
                                 std::pin::pin!(sleep(reconnect_backoff.next_delay()));
                             loop {
@@ -1336,6 +1337,27 @@ async fn run_app_runtime_account_worker(
                                     _ = &mut retry_delay => break,
                                     command = commands.recv() => {
                                         match command {
+                                            Some(command @ AccountWorkerCommand::CatchUp { .. }) => {
+                                                // A host connectivity-restored edge is the one
+                                                // command that is meaningful without an engine
+                                                // session: retain its response and use it to end
+                                                // only this stale sleep. If the network is still
+                                                // unavailable, the failed reopen below advances
+                                                // the same bounded backoff instead of spinning.
+                                                pending.push_back(command);
+                                                tracing::debug!(
+                                                    target: "marmot_app::runtime",
+                                                    method = "account_worker_reconnect",
+                                                    phase = "backoff_wait",
+                                                    outcome = "interrupted",
+                                                    elapsed_ms = u64::try_from(
+                                                        retry_started_at.elapsed().as_millis()
+                                                    )
+                                                    .unwrap_or(u64::MAX),
+                                                    "host recovery interrupted account worker reconnect backoff",
+                                                );
+                                                break;
+                                            }
                                             // There is deliberately no engine
                                             // session during this backoff.
                                             // Poll the bounded channel and
@@ -1381,10 +1403,11 @@ async fn run_app_runtime_account_worker(
                                     // loop on a full catch-up; the maintenance
                                     // path performs bounded repair syncs when
                                     // required.
+                                    let telemetry = shared.app_performance_telemetry();
                                     let prepare_transport = tokio::select! {
                                         _ = wait_for_runtime_shutdown(&mut lifecycle_shutdown) => return,
                                         _ = &mut shutdown => return,
-                                        result = reopened.prepare_transport() => result,
+                                        result = reopened.prepare_transport_with_telemetry(Some(&telemetry)) => result,
                                     };
                                     if let Err(transport_err) = prepare_transport {
                                         publish_app_runtime_account_error(
@@ -1643,8 +1666,9 @@ async fn handle_account_worker_catch_up(
     let mut deferred = VecDeque::new();
     let mut commands_open = true;
     let sync_started_at = Instant::now();
+    let stage_telemetry = context.shared.app_performance_telemetry();
     let sync_result = {
-        let mut sync = std::pin::pin!(client.sync_with_classified_partial_progress());
+        let mut sync = std::pin::pin!(client.sync_with_stage_telemetry(&stage_telemetry));
         loop {
             let command = if let Some(command) = pending.pop_front() {
                 Some(command)
--- a/crates/marmot-app/src/runtime/account_worker.rs
+++ b/crates/marmot-app/src/runtime/account_worker.rs
@@ -1341,10 +1341,20 @@ async fn run_app_runtime_account_worker(
                                                 // A host connectivity-restored edge is the one
                                                 // command that is meaningful without an engine
                                                 // session: retain its response and use it to end
-                                                // only this stale sleep. If the network is still
-                                                // unavailable, the failed reopen below advances
-                                                // the same bounded backoff instead of spinning.
+                                                // only this stale sleep. Coalesce an already
+                                                // queued burst into the same reopen attempt; new
+                                                // signals can still interrupt later sleeps, so
+                                                // extra attempts remain bounded by host signal
+                                                // rate while the network stays unavailable.
                                                 pending.push_back(command);
+                                                while let Ok(command) = commands.try_recv() {
+                                                    match command {
+                                                        command @ AccountWorkerCommand::CatchUp { .. } => {
+                                                            pending.push_back(command);
+                                                        }
+                                                        command => drop(command),
+                                                    }
+                                                }
                                                 tracing::debug!(
                                                     target: "marmot_app::runtime",
                                                     method = "account_worker_reconnect",
--- a/crates/marmot-app/src/tests.rs
+++ b/crates/marmot-app/src/tests.rs
@@ -9686,13 +9686,18 @@ fn connectivity_recovery_interrupts_max_account_worker_reconnect_backoff() {
             account_id: &MemberId,
             minimum: usize,
         ) {
-            for _ in 0..10_000 {
-                if relay.inbox_subscription_count(account_id) >= minimum {
+            let deadline = std::time::Instant::now() + Duration::from_secs(15);
+            loop {
+                let observed = relay.inbox_subscription_count(account_id);
+                if observed >= minimum {
                     return;
                 }
+                assert!(
+                    std::time::Instant::now() < deadline,
+                    "account inbox subscription count did not reach {minimum}; observed {observed}"
+                );
                 tokio::task::yield_now().await;
             }
-            panic!("account inbox subscription count did not reach {minimum}");
         }
 
         const ACCOUNT: &str = "alice";
@@ -9718,14 +9723,15 @@ fn connectivity_recovery_interrupts_max_account_worker_reconnect_backoff() {
                 .shared_services()
                 .relay_plane()
                 .simulate_notification_recovery_for_test(1);
-            for probe in 0..10_000 {
+            let backoff_deadline = std::time::Instant::now() + Duration::from_secs(15);
+            loop {
                 match runtime.unhydrated_group_count_for_test(ACCOUNT).await {
                     Err(AppError::TransportClosed) => break,
                     Ok(_) => tokio::task::yield_now().await,
                     Err(error) => panic!("unexpected reconnect probe error: {error:?}"),
                 }
                 assert!(
-                    probe < 9_999,
+                    std::time::Instant::now() < backoff_deadline,
                     "account worker did not enter reconnect backoff"
                 );
             }
@@ -9743,14 +9749,15 @@ fn connectivity_recovery_interrupts_max_account_worker_reconnect_backoff() {
             .shared_services()
             .relay_plane()
             .simulate_notification_recovery_for_test(1);
-        for probe in 0..10_000 {
+        let max_backoff_deadline = std::time::Instant::now() + Duration::from_secs(15);
+        loop {
             match runtime.unhydrated_group_count_for_test(ACCOUNT).await {
                 Err(AppError::TransportClosed) => break,
                 Ok(_) => tokio::task::yield_now().await,
                 Err(error) => panic!("unexpected max-backoff probe error: {error:?}"),
             }
             assert!(
-                probe < 9_999,
+                std::time::Instant::now() < max_backoff_deadline,
                 "account worker did not enter maximum reconnect backoff"
             );
         }
@@ -9768,7 +9775,8 @@ fn connectivity_recovery_interrupts_max_account_worker_reconnect_backoff() {
         };
 
         tokio::time::advance(Duration::from_secs(3)).await;
-        for _ in 0..10_000 {
+        let reconnect_deadline = std::time::Instant::now() + Duration::from_secs(15);
+        loop {
             if relay
                 .blocked_subscribe_count
                 .load(std::sync::atomic::Ordering::SeqCst)
@@ -9776,6 +9784,10 @@ fn connectivity_recovery_interrupts_max_account_worker_reconnect_backoff() {
             {
                 break;
             }
+            assert!(
+                std::time::Instant::now() < reconnect_deadline,
+                "connectivity recovery did not start a reconnect subscription"
+            );
             tokio::task::yield_now().await;
         }
         assert_eq!(
--- a/crates/marmot-app/src/tests.rs
+++ b/crates/marmot-app/src/tests.rs
@@ -9677,6 +9677,139 @@ fn account_worker_reconnect_backoff_doubles_caps_and_resets() {
 }
 
 #[test]
+fn connectivity_recovery_interrupts_max_account_worker_reconnect_backoff() {
+    run_composed_app_runtime_test("connectivity-recovery-wake", || async {
+        tokio::time::pause();
+
+        async fn wait_for_inbox_subscriptions(
+            relay: &ScriptedPushRelayClient,
+            account_id: &MemberId,
+            minimum: usize,
+        ) {
+            for _ in 0..10_000 {
+                if relay.inbox_subscription_count(account_id) >= minimum {
+                    return;
+                }
+                tokio::task::yield_now().await;
+            }
+            panic!("account inbox subscription count did not reach {minimum}");
+        }
+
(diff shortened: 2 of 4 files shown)


erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T20:56:57Z` |
| Commit reviewed | `a0c0daa823a4f0331dbb26a7963cf2b230ff7610` |
| Model | `Cursor Grok 4.6` |

This is the right, surgical fix for #1546. The lost wake was real: reconnect backoff drained the command channel and dropped `CatchUp`, so Android's offline→online `catch_up_accounts()` could not shorten a stale sleep. Preserving only that command — the one that is meaningful without an engine session — and using it to end the current `sleep` is the minimum change that matches the issue.

What I checked:

- After a failed reopen the same bounded `next_delay()` still runs. A CatchUp already sitting in `pending` is not re-read by the sleep `select!`, so one preserved signal does not spin. Only a *new* channel `CatchUp` interrupts again.
- Successful reconnect still returns to the main loop via `continue 'worker'`; `pending` is drained there before `receive_next_delivery`, so the preserved CatchUp actually runs.
- Two concurrent `catch_up_accounts()` callers coalesce on the in-flight catch-up (`handle_account_worker_catch_up`), matching the test's +2 subscription / +2 telemetry assertion (reconnect `prepare_transport` + one catch-up sync).
- Telemetry fields stay aggregate (`phase`, `outcome`, `elapsed_ms`, existing activation/subscription counters). No account ids, relays, or payloads.
- The paused-clock test is the acceptance test the issue asked for: it walks 2/4/8/16/32, arms the 60s cap, and proves subscribe starts within a 3s advance that could not have completed the stale sleep. `block_next_subscribe` also shows catch-up waiters do not finish before the receiver subscription is ready.
- `#926` (publish-pool recovery) is correctly left alone.

Non-blocking notes:

- A second host `CatchUp` that arrives while the next backoff sleep is armed will interrupt that sleep too. That is the right “new connectivity edge” behavior, but a chatty `onCapabilitiesChanged` can still cause one reconnect attempt per extra signal once you are at the 60s cap. Worth knowing; not a reason to hold the PR.
- `catch_up_accounts()` used to get immediate `TransportClosed` when the oneshot was dropped in backoff. It now waits up to `APP_RUNTIME_ACCOUNT_READY_WAIT` (45s) if reopen is still failing. That is the intended recovery-wait semantic; hosts that treated `TransportClosed` as “retry immediately” will see a longer wait on a false-positive online edge.
- The still-offline / failed-reopen non-spin path is correct in the code but not covered by the new test. A cheap follow-up would be: interrupt, fail the next subscribe, and assert the next attempt is not immediate.

GitHub rejected `--approve` because this review is on the reviewer's own PR. The recommendation is still merge.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1336,6 +1337,27 @@ async fn run_app_runtime_account_worker(
                                     _ = &mut retry_delay => break,
                                     command = commands.recv() => {
                                         match command {
+                                            Some(command @ AccountWorkerCommand::CatchUp { .. }) => {
```
erskingardner, line 1340: Non-blocking: each *new* channel `CatchUp` interrupts the then-current sleep, so during a sustained outage a burst of host signals produces one full `runtime_local_client` reopen cycle per signal with no backoff gap between them (the freshly armed `next_delay()` sleep is cut short the moment the next signal arrives). That is the correct behavior for a genuine connectivity edge and is bounded by host signal rate, but if it ever matters, a cheap hardening is to drain any further `CatchUp`s already queued in `commands` into `pending` (e.g. `try_recv` loop) before `break`, so a burst coalesces into a single reopen attempt.

crates/marmot-app/src/tests.rs
```diff
@@ -9676,6 +9676,139 @@ fn account_worker_reconnect_backoff_doubles_caps_and_resets() {
...
+        recovery_b.await.unwrap().unwrap();
+        assert_eq!(
+            relay.inbox_subscription_count(&account_id),
+            subscriptions_before_wake + 2,
+            "two recovery signals must coalesce into one catch-up transport refresh after reconnect"
+        );
+
+        let telemetry_after = runtime.app_performance_snapshot();
+        assert_eq!(
+            telemetry_after.account_transport_activation.attempts,
```
erskingardner: Non-blocking: these exact-count assertions implicitly rely on the maintenance paths (`sync_with_partial_progress`, bare `prepare_transport`) staying on the non-telemetry variants, and on `inbox_subscription_count` being a cumulative accepted-subscription count. Both hold today; just be aware that wiring stage telemetry into maintenance syncs or changing relay subscription bookkeeping will make these drift for reasons unrelated to this fix.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T21:10:52Z` |
| Commit reviewed | `a0c0daa823a4f0331dbb26a7963cf2b230ff7610` |

This is the right fix for #1546 and it is surgical: 4 files, +163/-5, with every hunk traceable to an acceptance criterion. The lost wake was real — the backoff sleep drained the command channel and dropped `CatchUp`, so the Android offline→online `catch_up_accounts()` edge could not shorten a stale sleep.

What I verified against the head SHA:

- **Interrupt semantics.** The preserved `CatchUp` goes into `pending`, which is guaranteed empty at that point (the main-loop command branch drains it fully before re-entering `receive_next_delivery`), so FIFO is preserved. The sleep `select!` only reads the channel, so the retained command cannot re-interrupt the next sleep — only a *new* signal does. After a successful reopen, `continue 'worker'` re-enters the main select, whose command branch polls `pending` first, so the retained `CatchUp` actually executes and its responder is answered.
- **Bounded, non-spinning backoff.** A failed reopen after an interrupt re-arms `next_delay()`, so a false-positive online edge while still offline yields one extra attempt, not a spin. The interrupted delay slot is consumed even though only part of it was slept — that biases toward longer subsequent waits, which is the conservative direction. `reset()` still only happens on a successful sync.
- **Coalescing.** A second `CatchUp` arriving mid-reopen sits in the channel and is coalesced by `handle_account_worker_catch_up` (`CatchUp { respond } if deferred.is_empty()`), matching the test's +2 subscriptions / +2 activation / +2 registration assertions (one reconnect `prepare_transport_with_telemetry`, one catch-up `sync_with_stage_telemetry` pass).
- **`sync_with_stage_telemetry` is behavior-identical** to `sync_with_classified_partial_progress` plus telemetry (`sync_inner(Some(telemetry))`); the rename is justified now that it is not startup-only, and the old variant still has callers (no dead code). Maintenance paths correctly stay on the non-telemetry variants, which is also what keeps the test's exact-count assertions valid.
- **Privacy.** The new `debug!` uses an explicit `target`/`method` with `phase`, `outcome`, `elapsed_ms` only; telemetry fields remain aggregate counters. `tokio` `test-util` landed under `[dev-dependencies]`, so no leakage into app artifacts.
- **Test.** The paused-clock test is the acceptance test the issue asked for: it walks 2/4/8/16/32, arms the 60s cap, and proves the reconnect subscribe starts within a 3s advance that cannot have completed the stale sleep. `block_next_subscribe` also proves catch-up waiters do not finish before the receiver subscription is ready.

Non-blocking notes (details inline):

- Each *additional* host signal during a sustained outage causes one more immediate reopen cycle (full `runtime_local_client` open) with no backoff gap between bursts. Bounded by signal rate, and the Android side only fires on genuine edges — but a one-line hardening exists if it ever matters.
- `catch_up_accounts()` during a sustained outage now waits up to `APP_RUNTIME_ACCOUNT_READY_WAIT` (45s) and returns a Timeout-classified failure, where it previously got an immediate `TransportClosed`. That is the intended recovery-wait semantic, but hosts that treated `TransportClosed` as "retry immediately" will see a longer hold on a false-positive edge.
- The still-offline / failed-reopen non-spin path is correct in the code but not covered by the new test. A cheap follow-up: interrupt, fail the next subscribe, assert the following attempt is not immediate.

Validation gap: the Rust test matrix was still running on CI at review time (clippy/check/audit already green); the five failing `relay_runtime` tests and the `fast-ci` sed incompatibility are documented in the PR body as pre-existing on master.

GitHub rejects `--approve` on the author's own PR, so this posts as a comment — the recommendation is merge.

crates/marmot-app/src/runtime/account_worker.rs
```diff
@@ -1336,6 +1337,27 @@ async fn run_app_runtime_account_worker(
...
+                                                // A host connectivity-restored edge is the one
+                                                // command that is meaningful without an engine
+                                                // session: retain its response and use it to end
+                                                // only this stale sleep. If the network is still
+                                                // unavailable, the failed reopen below advances
+                                                // the same bounded backoff instead of spinning.
```
erskingardner: Non-blocking: this comment slightly overstates the guarantee. If the host keeps emitting catch-up signals while the network is still down, each one interrupts the freshly armed sleep on the next loop iteration, so reopen attempts end up running at signal rate — the backoff *value* advances but is never actually awaited. In practice host recovery signals are edge-triggered and the command channel is bounded, so this is acceptable; but the "instead of spinning" claim only holds once signals stop arriving. Consider wording it as "bounded by host signal rate" or noting the assumption.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-27T21:13:32Z` |
| Commit reviewed | `a0c0daa823a4f0331dbb26a7963cf2b230ff7610` |

*(Posted as a comment because GitHub does not allow approving your own PR.)*

Targeted fix for the lost-wake window in #1546: the backoff loop now retains a host `CatchUp` command and lets it interrupt only the stale sleep, while failed reopens continue on the same bounded exponential policy. Verified against the head SHA:

- The retained command is drained first after reconnect (steady-state `pending.pop_front()` before `commands.recv()`), and repeated catch-ups coalesce onto one sync via `catch_up_responders` in `handle_account_worker_catch_up` — no lost wake, no duplicate syncs.
- The `sync_with_startup_stage_telemetry` → `sync_with_stage_telemetry` rename is body-identical and the widened visibility of `prepare_transport_with_telemetry` stays `pub(crate)`; no other callers exist.
- New telemetry is aggregate-only (operation, duration, success) and the backoff-interrupt `tracing::debug!` follows the observability invariants (explicit target/method, no ids or payloads).
- `tokio/test-util` correctly lands in dev-dependencies only; the paused-clock regression test drives the 2/4/8/16/32s progression into the 60s cap deterministically and asserts recovery starts within 3s.

One non-blocking wording note inline on the backoff comment. Scope is surgical (+163/-5, 4 files) with a test that failed before and passes after. Note: several CI checks were still pending at review time — confirm green before merging.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Addressed in bcae01cb. The reconnect wait now drains the already-queued burst when the first `CatchUp` arrives: catch-ups are retained in `pending` for one coalesced sync, while commands that cannot run without an engine are rejected as before. This keeps one queued burst to one early reopen attempt.

crates/marmot-app/src/tests.rs
erskingardner: Addressed in bcae01cb. The telemetry checks are now lower bounds (`>= before + 2`), so future maintenance telemetry does not make the test drift. The accepted-subscription assertion stays exact because it is the behavioral proof that the two recovery callers coalesce into one catch-up refresh after reconnect.

crates/marmot-app/src/runtime/account_worker.rs
erskingardner: Addressed in bcae01cb. The comment now states the actual guarantee: an already-queued burst is coalesced, while genuinely new signals may interrupt later sleeps, bounding extra reopen attempts by host signal rate during a sustained outage.

crates/marmot-app/src/tests.rs
```diff
@@ -9676,6 +9676,155 @@ fn account_worker_reconnect_backoff_doubles_caps_and_resets() {
...
+            let deadline = std::time::Instant::now() + Duration::from_secs(15);
+            loop {
+                let observed = relay.inbox_subscription_count(account_id);
+                if observed >= minimum {
+                    return;
+                }
+                assert!(
+                    std::time::Instant::now() < deadline,
+                    "account inbox subscription count did not reach {minimum}; observed {observed}"
+                );
```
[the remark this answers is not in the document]
erskingardner: The `std::time::Instant` deadline is intentionally the separate wall-clock safety watchdog suggested here, not the functional recovery deadline. The functional contract is the explicit `tokio::time::advance(Duration::from_secs(3))` immediately before the final wait; none of these readiness loops advances Tokio time further. A Tokio timer could auto-advance while the current-thread runtime waits on blocking SQLite/open work and would therefore weaken or falsify the three-second assertion. The old failure was the fixed 10,000-yield budget expiring under shard load; the bounded wall watchdog gives that blocking work real scheduling time while keeping virtual time fixed. The exact CI shard command now passes 962/962 locally under the concurrent partition workload.

PR #1590, opened 2026-08-27
Feat/epoch stall fruitless escalation

jgmontoya:
## Summary by CodeRabbit

- **New Features**
  - Added automatic detection and escalation for groups stalled at a frozen epoch.
  - Added durable stall evidence and delivery-overflow recovery across restarts.
  - Added transport-history reconciliation for skipped or incomplete deliveries.
  - Added pacing, continuation-window, clock-skew, and fruitless-replay safeguards.

- **Bug Fixes**
  - Improved connectivity recovery after interruptions and repeated recovery signals.
  - Preserved partial synchronization results when repairs cannot fully complete.

- **Documentation**
  - Updated convergence and reliability guidance for the new recovery behavior.


Datawav (commented):
## Automated review

Actionable findings: 2.

1. Major — Restored refusal-only stalls cannot re-arm
- Evidence: In `crates/marmot-app/src/client/epoch_stall.rs`, the changed durable-restore path `restore_wedge_evidence` restores both `armed_at_epoch` and `fired_at_epoch` for the persisted stalled epoch. The same file’s changed `observe_resource_refusal(..., now_ms)` path still returns `BackfillDecision::Skip` when `stall.fired_at_epoch == Some(epoch)`; unlike `observe_undecryptable`, it does not invoke `may_rearm_wedged`/`rearm_wedged`.
- Triggering state: A process restarts with wedge evidence persisted at epoch E, the peel/delivery cap remains saturated, and subsequent ingests at E produce only `ResourceRefused`.
- Consequence: Every refusal is skipped. No new full-history replay starts, so `fruitless_completions` cannot advance and the documented frozen-epoch escalation can remain permanently unreachable after restart.
- Confidence: High. The restored latch, refusal branch, and paced re-arm asymmetry are all present in the exact-head patch; `rearm_refused_groups` cannot break the cycle because it runs only after a replay completes.
- Smallest sound correction: Give `observe_resource_refusal` the same elapsed-interval `may_rearm_wedged`/`rearm_wedged` path used by `observe_undecryptable`, without counting that paced re-arm toward `EPOCH_STALL_ESCALATION_ARM_THRESHOLD`. Add a restore-plus-`ResourceRefused` test proving `Skip` before the interval and `Arm` after it.

2. Minor — Future clock marks do not reset run evidence consistently
- Evidence: At changed lines 419–424 of `crates/marmot-app/src/client/epoch_stall.rs`, `GroupStall::arm` evaluates the continuation window with `now_ms.saturating_sub(last) > EPOCH_STALL_RUN_CONTINUATION_WINDOW_MS`. The changed wedge gate instead uses `elapsed_since_mark_ms`, whose future-mark handling treats a mark beyond `EPOCH_STALL_CLOCK_SKEW_ALLOWANCE_MS` as elapsed so a corrected clock starts fresh.
- Triggering state: Restored `last_arm_at_ms` is materially ahead of the current clock; `rearm_refused_groups` clears the fired epoch and the next observation enters `arm`.
- Consequence: `saturating_sub` yields zero, so the prior run’s `fruitless_completions` and `fruitless_reported` survive for one arm before the timestamp is overwritten. This can carry stale escalation evidence across a clock correction.
- Confidence: High. The two changed clock gates demonstrably implement conflicting future-mark semantics.
- Smallest sound correction: Use `elapsed_since_mark_ms(now_ms, last)` for the run-continuation comparison and add a future-mark boundary test showing the next arm starts with fresh run evidence.

Audit evidence:
- Exact head SHA: `bb0756f8138a6e04db904268da82ea5a929abc9a`
- Patch coverage is complete across 17 changed files, including `crates/marmot-app/src/client/epoch_stall.rs`, `crates/marmot-app/src/client/sync.rs`, and `crates/storage-sqlite/src/migrations/0054_epoch_stall_evidence.rs`.
- Discussion classification: the dated-page comment is stale because `docs/marmot-architecture/convergence-reliability-plan.md` now changes `updated:` to `2026-08-27`; the run-continuation clock comment remains supported and is finding 2; the summary comment’s delivery-queue-overflow claims do not establish an additional finding in the bounded patch; no organization-member feedback was present.

Risk lenses checked:
- network/protocol: timeouts, retry and duplicate delivery; partial/malformed peer data; version and wire compatibility.
- tests: assertions fail before the fix; boundary and negative cases; test exercises production behavior.
- Large-change decomposition and behavioral-test reasoning were also checked.

Check evidence:
- GitHub Actions reports 24 successful checks, including Rust check, all four Rust-test shards, Rust clippy, Rust audit, relay tests, Tamarin proofs, convergence-policy assertions, packaging, and platform builds.
- `Simulator and vectors` completed with failure. The bundle contains no changed-line evidence tying that failure to a separate defect, so it is not promoted to an additional finding.
- `Publish wn-agent release` was skipped; all recorded checks completed, but the check set is not wholly successful.

crates/marmot-app/src/client/sync.rs
```diff
@@ -2039,29 +2044,63 @@ impl AppClient {
...
+        arms: u32,
+        decided_by_threshold: u32,
+        method: &'static str,
+    ) {
+        // The threshold logged is the one that actually decided, which is not
+        // always the arm-run threshold the audit row carries: `method` names the
+        // rule and this names the count it reached.
+        tracing::warn!(
+            target: "marmot_app::epoch_stall",
+            method,
```
erskingardner: **Blocking — this is the CI failure.** The `Simulator and vectors` check fails here: `production_tracing_calls_are_structured_and_privacy_safe` scans for the literal `method =` token in every production tracing call body (tracing_audit.rs:71), and field shorthand `method,` doesn't match it, even though it produces the identical field at runtime.

Fix: `method = method,` (or keep shorthand everywhere and teach the audit to accept it — but that's a wider change and the repo convention is `method = "..."` literals, so matching the audit is the smaller move).

crates/marmot-app/src/client/sync.rs
```diff
@@ -3843,6 +3979,17 @@ fn backfill_drain_verdict(eose: AccountSubscriptionEose) -> DrainVerdict {
 /// Doubling backoff from `base`, capped at [`EPOCH_BACKFILL_RETRY_BACKOFF_CAP`]
 /// (or `base` itself when a test override exceeds the cap). Pure so the
 /// schedule is table-testable without a client.
+/// Wall clock for the epoch-stall detector's two time gates.
```
erskingardner, line 4280: Non-blocking: the new function was inserted between `retry_backoff_for_ordinal` and its doc comment, so the "Doubling backoff from `base`..." paragraph (three lines up) now renders as the first paragraph of `epoch_stall_now_ms`'s docs, and `retry_backoff_for_ordinal` is left undocumented. Move the new doc + function above the existing doc block (or below the function).

crates/marmot-app/src/client/epoch_stall.rs
```diff
@@ -94,10 +120,111 @@ pub(crate) const EPOCH_STALL_BACKFILL_THRESHOLD: usize = 8;
...
+
+/// Replay completions that reached end-of-stored-events and recovered nothing,
+/// at one stalled epoch, before the runtime reports the group as beyond what
+/// full-history replay can repair.
+///
+/// This is the escalation rule for a group whose epoch never moves, and it
+/// counts *evidence*, not attempts. A completion counts only when the relays
+/// confirmed they had served the account's stored history
+/// (`EpochBackfillCompletionKind::EndOfStoredEvents`) and the replay still
+/// recovered nothing for the group (`AppClient::replay_recovered_something`).
```
erskingardner, line 195: Minor doc nuance: the call site (`finish_epoch_backfill_execution`) only reaches `observe_fruitless_completion` when `replay_recovered_something` is false for the **whole account** — a replay that recovered something for a different group skips the fruitless branch entirely (`mark_replayed()` and return). So the evidence is "recovered nothing for any group", which is stricter than "recovered nothing for the group" as written here.

The conservative direction is safe (it can only delay a wedged group's report, never inflate it), but the doc claims the narrower rule. Worth one clarifying sentence so a future reader of the constant doesn't assume per-group accounting.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-28T12:03:11Z` |
| Commit reviewed | `bb0756f8138a6e04db904268da82ea5a929abc9a` |

## Summary

The premise is sound and well-scoped: this closes the blind spot the `epoch_stall` module header explicitly tracked for a follow-up (a group frozen at one epoch could arm once and never escalate). The design is careful — paced re-arms on a wall clock, escalation counted on relay-confirmed end-of-stored-events completions rather than arms (preserving the minted-traffic property), durable per-group evidence with a deliberately process-local arm run, and a clock-skew allowance that self-heals a dead-RTC mark. Tests are thorough at every layer (detector unit tests, end-to-end runtime tests, storage round-trip, migration FK checks), and the five ledger surfaces are all updated per the repo invariant.

One blocking issue: **CI is red.** The `Simulator and vectors` job fails because the new `tracing::warn!` in `report_epoch_stall_escalation` uses the field shorthand `method,`, and the tracing audit (`crates/cgka-conformance-simulator/tests/tracing_audit.rs:71`) requires the literal `method =` token in every production tracing call body. One-word fix, but it must land (and the audit re-run) before merge.

Two smaller items inline: a doc comment that got attached to the wrong function, and a doc claim about per-group vs account-wide recovery that the code doesn't quite make.

crates/marmot-app/src/client/sync.rs
erskingardner: **Blocking: this fails the `Simulator and vectors` CI job** (`production_tracing_calls_are_structured_and_privacy_safe`).

The tracing audit (`crates/cgka-conformance-simulator/tests/tracing_audit.rs:97`) requires every production tracing call to contain a literal `method =` field, per the repo observability invariant (explicit `target` and `method` fields). The field shorthand `method,` does not match, and CI fails with:

```
crates/marmot-app/src/client/sync.rs:2088 tracing call is missing a method field
```

Writing it as `method = method,` keeps the dynamic rule name and satisfies the audit.

crates/marmot-app/src/client/sync.rs
```diff
@@ -3843,6 +3979,17 @@ fn backfill_drain_verdict(eose: AccountSubscriptionEose) -> DrainVerdict {
...
+///
+/// Wall clock rather than [`Instant`] because both gates have to survive a
+/// restart: a device wedged for six hours must not buy a re-arm by being
+/// force-killed, and a monotonic clock that restarts at zero would hand it one.
+/// Read here rather than inside the detector, which stays I/O-free so its
+/// policy can be unit-tested in isolation.
+pub(crate) fn epoch_stall_now_ms() -> u64 {
```
erskingardner, line 4287: Minor: `epoch_stall_now_ms` was inserted between `retry_backoff_for_ordinal` and its doc comment. The "Doubling backoff from `base`…" doc (lines 3979–3981) now attaches to `epoch_stall_now_ms`, and `retry_backoff_for_ordinal` below lost its documentation. Please move the new function above the `/// Doubling backoff` doc block (or below `retry_backoff_for_ordinal`) so each doc comment stays with its function.

erskingardner (dismissed):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-08-28T12:03:38Z` |
| Commit reviewed | `bb0756f8138a6e04db904268da82ea5a929abc9a` |

The design here is solid: pacing frozen-epoch re-arms on a wall clock, escalating only on relay-confirmed end-of-stored-events fruitless completions (so minted traffic can pace attempts but cannot forge evidence), and making the evidence durable while keeping the arm run process-local are all well-reasoned, and the test coverage — including the restart and clock-skew cases — is thorough. The ledger/constant registration (A10) and docs updates follow repo convention.

Two blocking issues before merge:

1. **CI is red.** `Simulator and vectors` fails on `production_tracing_calls_are_structured_and_privacy_safe`: the new `tracing::warn!` in `report_epoch_stall_escalation` uses the field shorthand `method,`, which the tracing audit does not accept (inline comment on `sync.rs:2090`). One-character-class fix: `method = method,`.

2. **Restored refusal-only stalls cannot re-arm** (concurring with the automated finding already on this PR, verified independently at this head). `restore_wedge_evidence` latches `fired_at_epoch = Some(epoch)`, and `observe_resource_refusal` still returns `Skip` whenever that latch matches — it never consults `may_rearm_wedged`/`rearm_wedged` the way `observe_undecryptable` does. A restarted device whose wedged group produces only `ResourceRefused` (deferred-peel cap saturated) therefore skips every refusal, no replay ever arms for it, and the fruitless-completion escalation this PR adds is permanently unreachable for exactly that shape. Before this PR, a restart cleared the latch, so this is a regression introduced by persisting the evidence. Suggested shape: give `observe_resource_refusal` the same paced re-arm path (without counting it toward `EPOCH_STALL_ESCALATION_ARM_THRESHOLD`), plus a restore + refusal test.

Non-blocking:

- Doc-comment misplacement in `sync.rs`: `epoch_stall_now_ms` sits between `retry_backoff_for_ordinal` and its doc comment, stealing the "Doubling backoff" doc (inline comment at `sync.rs:3989`).
- Also concur with the existing nitpick that `GroupStall::arm` should use `elapsed_since_mark_ms` instead of `saturating_sub` for the run-continuation gate, so both clock gates treat a future mark the same way.

PR #1603, opened 2026-08-30
fix: recover durable outbound sends after connectivity returns

mubarakcoded:
## Problem

Temporary relay or network failures could leave outbound application messages and group-invite commits pending for too long, even after connectivity returned. Two offline messages could retry with a long interval between them, and a successfully published retry could remain visually stuck in `Sending` because the retry path lacked enough application-event context to finalize the optimistic timeline row.

The same failure family also caused startup/catch-up transport errors to surface too eagerly and allowed an older unresolved pending row to remain pinned as a chat-list preview after newer authenticated activity arrived.

## Root causes

- `nostr-sdk` relay connection startup is asynchronous. The previous adapter path could race the background connection task, classify an obviously unavailable pre-send failure as possibly exposed, and put it behind the conservative 30-second exponential ambiguity backoff.
- Durable fanout persistence kept the exact signed transport event and relay targets, but not the application-message identity needed to project an eventless retry acknowledgement back into the local timeline.
- Runtime scheduling treated convergence inputs and queued intents as work, but did not consistently wake already-frozen fanouts when the host observed restored connectivity or while an account worker was sleeping in reconnect backoff.
- Chat-list preview selection reused timeline ordering that intentionally favors unresolved local sends; this could let an old orphaned pending row hide newer authenticated activity indefinitely.

### Durable outbound retry

- Persist optional `OutboundApplicationMessage` context with each frozen fanout: group id, application-event id, source epoch, and retention policy.
- Preserve and retry the exact original signed event bytes and targets; no replacement event is constructed during retry.
- Query pending fanouts by group and drain them before newer group work so multiple offline messages retry once and in order.
- When an acknowledged retry produces no new engine event, emit the stored application-message projection so every matching optimistic timeline row transitions out of `Sending`.
- Apply the same durable exact-fanout path to group-invite commit publication.

### Connectivity recovery and failure classification

- Add the additive host API `notify_connectivity_restored` through the runtime, UniFFI, and C surfaces.
- Wake account workers, interrupt reconnect sleep, clear scheduled transport backoff, and schedule durable group work immediately when usable connectivity returns.
- Only clear retry clocks for targets proven unavailable before exposure. Possibly-exposed failures retain the conservative ambiguity/duplicate-safety backoff.
- Use `try_connect_relay` to prove pre-send connection failures are retryable-unavailable, preserve relay options, and keep privacy-safe error text free of relay URLs.
- Pin SDK reconnect cadence to the durable retry budget.
- Retry startup/catch-up only for transient transport activation, subscription, and receive failures (250 ms, 1 s, 3 s). Protocol, crypto, storage, cancellation, closed-worker, and exhausted-budget failures remain terminal.

### Chat-list preview repair

- Select preview candidates using stable application-event insertion order so newer authenticated activity replaces an older unresolved pending preview.
- Preserve the expected behavior where a genuinely newer pending send remains the preview.
- Prevent replayed older authenticated events from leapfrogging newer activity.
- Use the same ordering for projection-completeness and secure-prune repair paths.

The unresolved timeline row itself is retained; this change repairs only which event is presented as the latest chat preview.

## Safety and compatibility

- No MLS/protocol wire-format change.
- Adds SQLite migration 0056: an additive, defaulted `accepted_activity_insert_order` column with a backfill of existing accepted activity.
- Existing persisted fanouts remain readable because the new application-message field is optional and serde-defaulted.
- The host connectivity notification is additive. Clients that do not call it retain timer-based retry behavior, so existing iOS, macOS, Linux, and other integrations do not break.
- Ambiguous publication failures keep their existing conservative retry protection.
- The checked-in C header is refreshed for the new C ABI entry point.
- Generated Kotlin/Swift/JNI bindings, native binaries, and Android changes are intentionally not included. Local Android bindings were used only for device validation; Android integration remains a separate change.

## Validation

Automated:

- `just check`
- `just clippy`
- `cargo test -p marmot-app recover`
- focused connectivity-wake, eventless-projection, catch-up, ordered retry, and group-invite retry tests
- focused chat-list pending/authenticated preview ordering and repair tests
- focused frozen-fanout serialization and retry-clock safety tests
- focused `nostr-sdk` pre-exposure classification and reconnect-budget tests
- `cargo test -p cgka-engine --test convergence_policy_pin --locked`
- `cargo test -p marmot-c --features alloc-audit --locked`
- C/UniFFI command parity gate
- generated `marmot.h` inspection and clean diff check

Manual device validation with locally generated, uncommitted bindings:

- queued two messages offline, force-stopped and cold-started without clearing app data, restored connectivity, and confirmed ordered automatic delivery
- confirmed the companion client received both messages and the reply returned successfully
- confirmed the latest-message preview remained correct after restart
- accepted a two-member group invite offline, force-stopped and reopened offline, restored connectivity, and confirmed membership plus successful messages in both directions

`just fast-ci` passed formatting, naming, C/UniFFI parity, convergence-ledger, and static campaign checks, then stopped in the unchanged campaign-toolchain shell test because its GNU `sed -i` syntax is incompatible with macOS BSD `sed`. The affected script is unchanged from `master`; the equivalent Rust checks and focused regression suites above pass locally, and the Linux CI run remains authoritative for that harness.

## Summary by CodeRabbit

* **New Features**
  * Added connectivity-restored notifications for native and C integrations.
  * Added group-level visibility into pending outbound delivery work.
  * Preserved application-message metadata through delivery, retries, and restart recovery.

* **Bug Fixes**
  * Improved retry scheduling, reconnection handling, and ordered delivery recovery.
  * Chat previews now prioritize the most relevant accepted or pending activity.
  * Improved cleanup and retention updates after publishing or acknowledgement failures.

* **Tests**
  * Expanded coverage for retries, reconnection, ordering, previews, persistence, and recovery.

### DIFF
--- a/crates/cgka-engine/src/engine.rs
+++ b/crates/cgka-engine/src/engine.rs
@@ -773,9 +773,7 @@ impl<S: StorageProvider> Engine<S> {
         &self,
         group_id: &GroupId,
     ) -> Result<Vec<OutboundFanout>, EngineError> {
-        if self.quarantined_groups.contains_key(group_id)
-            || self.unhydrated_groups.contains(group_id)
-        {
+        if self.ensure_group_live(group_id).is_err() {
             return Ok(Vec::new());
         }
         Ok(self.storage.list_outbound_fanouts_for_group(group_id)?)
--- a/crates/cgka-engine/src/engine.rs
+++ b/crates/cgka-engine/src/engine.rs
@@ -767,6 +767,20 @@ impl<S: StorageProvider> Engine<S> {
             .collect())
     }
 
+    /// Read one live group's frozen fanouts in original staging order without
+    /// scanning every other group's durable outbox.
+    pub fn outbound_fanouts_for_group(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<Vec<OutboundFanout>, EngineError> {
+        if self.quarantined_groups.contains_key(group_id)
+            || self.unhydrated_groups.contains(group_id)
+        {
+            return Ok(Vec::new());
+        }
+        Ok(self.storage.list_outbound_fanouts_for_group(group_id)?)
+    }
+
     /// Delete a terminal fanout after its outcome has been returned.
     pub fn delete_outbound_fanout(&self, message_id: &MessageId) -> Result<(), EngineError> {
         if let Some(group_id) = self
--- a/crates/cgka-session/src/lib.rs
+++ b/crates/cgka-session/src/lib.rs
@@ -997,6 +997,32 @@ impl AccountDeviceSession {
         Ok(self.collect_effects(results))
     }
 
+    /// Settle retained convergence inputs without draining queued outbound
+    /// intents.
+    ///
+    /// A host uses this narrower operation when an older durable transport
+    /// fanout still blocks newer user-authored sends. Protocol settlement,
+    /// epoch progression, and fork healing remain live while queued work stays
+    /// behind the fanout ordering barrier.
+    pub async fn advance_convergence_inputs(
+        &mut self,
+        group_id: &GroupId,
+    ) -> SessionResult<SessionEffects> {
+        tracing::debug!(
+            target: TRACE_TARGET,
+            method = "advance_convergence_inputs",
+            "advancing convergence inputs without draining queued outbound intents"
+        );
+        let settled = self.engine.advance_convergence_inputs(group_id).await?;
+        tracing::debug!(
+            target: TRACE_TARGET,
+            method = "advance_convergence_inputs",
+            settled,
+            "convergence inputs advanced"
+        );
+        Ok(self.collect_effects(Vec::new()))
+    }
+
     pub fn has_pending_convergence_inputs(&self, group_id: &GroupId) -> SessionResult<bool> {
         Ok(self.engine.has_pending_convergence_inputs(group_id)?)
     }
--- a/crates/cgka-session/src/lib.rs
+++ b/crates/cgka-session/src/lib.rs
@@ -1160,6 +1160,13 @@ impl AccountDeviceSession {
         Ok(self.engine.outbound_fanouts()?)
     }
 
+    pub fn outbound_fanouts_for_group(
+        &self,
+        group_id: &GroupId,
+    ) -> SessionResult<Vec<OutboundFanout>> {
+        Ok(self.engine.outbound_fanouts_for_group(group_id)?)
+    }
+
     pub fn delete_outbound_fanout(&self, message_id: &MessageId) -> SessionResult<()> {
         self.engine.delete_outbound_fanout(message_id)?;
         Ok(())
--- a/crates/cgka-session/src/lib.rs
+++ b/crates/cgka-session/src/lib.rs
@@ -1002,8 +1002,11 @@ impl AccountDeviceSession {
     ///
     /// A host uses this narrower operation when an older durable transport
     /// fanout still blocks newer user-authored sends. Protocol settlement,
-    /// epoch progression, and fork healing remain live while queued work stays
-    /// behind the fanout ordering barrier.
+    /// epoch progression, and fork healing remain live while queued outbound
+    /// intents, pending disband work, and drain maintenance (including a due
+    /// self-remove auto-commit) stay behind the fanout ordering barrier. That
+    /// deferral can remain unbounded while the oldest fanout has an outstanding
+    /// target.
     pub async fn advance_convergence_inputs(
         &mut self,
         group_id: &GroupId,
--- a/crates/marmot-account/src/runtime.rs
+++ b/crates/marmot-account/src/runtime.rs
@@ -96,6 +96,7 @@ struct PublishStatus {
     accepted_by_any_endpoint: bool,
     possible_ambiguous_exposure: bool,
     retry_deferred: bool,
+    terminal_failure: bool,
 }
 
 struct PendingFanoutContinuation {
@@ -2018,9 +2019,12 @@ where
         // Retry the group's already-frozen transport events before producing
         // any newer convergence work. This preserves send order and ensures a
         // scheduled outbound wake actually drives the exact retained bytes.
-        let mut output = self
+        let (mut output, blocked_groups) = self
             .resume_outbound_fanouts_for_group(Some(group_id))
             .await?;
+        if blocked_groups.contains(group_id) {
+            return Ok(output);
+        }
         let effects = self.session.advance_convergence(group_id).await?;
         output.extend(self.publish_session_effects(effects).await?);
         Ok(output)
@@ -2262,7 +2266,9 @@ where
 
     /// Resume every incomplete frozen fanout in original staging order.
     pub async fn resume_outbound_fanouts(&mut self) -> AccountResult<AccountDeviceEffects> {
-        self.resume_outbound_fanouts_for_group(None).await
+        self.resume_outbound_fanouts_for_group(None)
+            .await
+            .map(|(effects, _blocked_groups)| effects)
     }
 
     /// Retire terminal accepted application fanouts only after the app has
@@ -2301,7 +2307,7 @@ where
     async fn resume_outbound_fanouts_for_group(
         &mut self,
         group_id: Option<&GroupId>,
-    ) -> AccountResult<AccountDeviceEffects> {
+    ) -> AccountResult<(AccountDeviceEffects, HashSet<GroupId>)> {
         let fanouts = match group_id {
             Some(group_id) => self.session.outbound_fanouts_for_group(group_id)?,
             None => self.session.outbound_fanouts()?,
@@ -2337,13 +2343,23 @@ where
                 record_published_application_fanout(&fanout, &mut output);
                 output.fanout.push(outcome);
             } else {
+                let reason = "insufficient publish acknowledgements".to_owned();
+                output.failures.push(PublishFailure {
+                    message_id: fanout
+                        .published_message_id()
+                        .unwrap_or_else(|| fanout.message_id())
+                        .clone(),
+                    reason: reason.clone(),
+                });
+                record_failed_application_fanout(&fanout, reason, &mut output);
+                output.fanout.push(outcome);
                 self.session.delete_outbound_fanout(fanout.message_id())?;
             }
         }
         self.publish_queue(&mut output, &mut queue, None).await?;
         self.reconcile_confirmed_own_leaf_rotations(&output.events)?;
         self.reconcile_superseded_maintenance(&output.events)?;
-        Ok(output)
+        Ok((output, blocked_groups))
     }
 
     fn reconcile_confirmed_own_leaf_rotations(
@@ -2498,9 +2514,28 @@ where
                     "published queued message but could not clear its durable intent"
                 );
             }
+        } else if status.terminal_failure {
+            if self
+                .session
+                .confirm_regenerated_queued_intent(&intent)
+                .is_err()
+            {
+                // Every target failed terminally, so the logical queued intent
+                // is complete even though it was not delivered. If durable
+                // cleanup fails, retain and re-arm it rather than losing the
+                // obligation.
+                self.session.retry_regenerated_queued_intent(&intent);
+                tracing::warn!(
+                    target: TRACE_TARGET,
+                    method = "resolve_regenerated_queued_intent",
+                    error_kind = "terminal_queued_intent_cleanup",
+                    "terminal queued message failed but its durable intent could not be cleared"
+                );
+            }
         } else {
-            // Nothing accepted the publish. The durable intent was never
-            // deleted; re-arm its group for the normal convergence retry.
+            // Nothing accepted, but routing/setup failure or at least one
+            // retryable or ambiguously exposed target means terminal failure
+            // has not been proved. Preserve and re-arm the durable intent.
             self.session.retry_regenerated_queued_intent(&intent);
         }
     }
@@ -3359,6 +3394,8 @@ where
             accepted_by_any_endpoint: report.accepted_count() > 0,
             possible_ambiguous_exposure,
             retry_deferred: fanout_outcome.outstanding_targets > 0,
+            terminal_failure: report.accepted_count() == 0
+                && fanout_outcome.outstanding_targets == 0,
         };
         let publish_failure_reason =
             if fanout_outcome.outstanding_targets > 0 && !status.met_required_acks {
@@ -3403,6 +3440,8 @@ where
         }
         if status.accepted_by_any_endpoint {
             record_published_application_fanout(&fanout, output);
+        } else if let Some(reason) = publish_failure_reason {
+            record_failed_application_fanout(&fanout, reason, output);
         } else if let Some(application) = fanout.application_message()
             && status.retry_deferred
         {
@@ -3680,6 +3719,7 @@ where
                 accepted_by_any_endpoint: accepted_before > 0,
                 possible_ambiguous_exposure: fanout.possible_exposure,
                 retry_deferred,
+                terminal_failure: false,
             }));
         }
         let attempt_target = publish_target_with_endpoints(&target, retry_endpoints.clone());
@@ -3783,6 +3823,7 @@ where
                     accepted_by_any_endpoint: accepted_before > 0,
                     possible_ambiguous_exposure: true,
                     retry_deferred: false,
+                    terminal_failure: false,
                 });
             }
         };
@@ -3857,6 +3898,7 @@ where
             accepted_by_any_endpoint,
             possible_ambiguous_exposure: fanout.possible_exposure,
             retry_deferred: false,
+            terminal_failure: !published && !accepted_by_any_endpoint && !fanout.possible_exposure,
         })
     }
 }
@@ -3994,6 +4036,29 @@ fn record_published_application_fanout(fanout: &OutboundFanout, output: &mut Acc
             source_epoch: application.source_epoch,
             retention: application.retention,
         });
+}
+
+/// Report a terminal application publish with the app-event identity retained
+/// on its durable fanout. This remains replayable across the crash window
+/// between persisting the terminal target states and deleting the fanout.
+fn record_failed_application_fanout(
+    fanout: &OutboundFanout,
+    reason: String,
+    output: &mut AccountDeviceEffects,
+) {
+    let Some(application) = fanout.application_message() else {
+        return;
+    };
+    let message_id = fanout
+        .published_message_id()
+        .unwrap_or_else(|| fanout.message_id())
+        .clone();
+    output.failed_app_messages.push(FailedApplicationMessage {
+        group_id: application.group_id.clone(),
+        app_event_id: application.app_event_id.clone(),
+        message_id,
+        reason,
+    });
 }
 
 fn apply_report_to_fanout(
@@ -4261,6 +4326,10 @@ pub struct AccountDeviceEffects {
     /// Application-message identity attached to unresolved publication so app
     /// consumers can keep exactly the affected local row in a sending state.
     pub unresolved_app_messages: Vec<UnresolvedApplicationMessage>,
+    /// Application messages whose publication reached a definitive terminal
+    /// failure, carrying the local app-event identity needed to invalidate only
+    /// the affected optimistic projection.
+    pub failed_app_messages: Vec<FailedApplicationMessage>,
     /// Application messages accepted by at least one transport endpoint,
     /// carrying source-state metadata captured by the exact MLS encryption
     /// operation and the adapter-visible transport id.
@@ -4281,6 +4350,14 @@ pub struct PublishedApplicationMessage {
     pub message_id: cgka_traits::MessageId,
     pub source_epoch: EpochId,
     pub retention: cgka_traits::app_event::AppMessageRetentionDecision,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct FailedApplicationMessage {
+    pub group_id: GroupId,
+    pub app_event_id: String,
+    pub message_id: cgka_traits::MessageId,
+    pub reason: String,
 }
 
 impl AccountDeviceEffects {
@@ -4311,6 +4388,8 @@ impl AccountDeviceEffects {
             .append(&mut other.unresolved_publishes);
         self.unresolved_app_messages
             .append(&mut other.unresolved_app_messages);
+        self.failed_app_messages
+            .append(&mut other.failed_app_messages);
         self.published_app_messages
             .append(&mut other.published_app_messages);
         self.welcome_failures.append(&mut other.welcome_failures);
--- a/crates/marmot-account/src/runtime.rs
+++ b/crates/marmot-account/src/runtime.rs
@@ -2222,6 +2222,39 @@ where
         self.resume_outbound_fanouts_for_group(None).await
     }
 
+    /// Retire terminal accepted application fanouts only after the app has
+    /// durably finalized their optimistic projections.
+    ///
+    /// Until this acknowledgement arrives, restart drains replay the same
+    /// publication metadata. App projection is idempotent, so a crash between
+    /// its database commit and this call leaves a harmless replay instead of a
+    /// permanently pending message.
+    pub fn acknowledge_published_app_messages(
+        &self,
+        published_messages: &[PublishedApplicationMessage],
+    ) -> AccountResult<()> {
+        for published in published_messages {
+            let fanouts = self
+                .session
+                .outbound_fanouts_for_group(&published.group_id)?;
+            for fanout in fanouts {
+                let outcome = fanout.outcome();
+                let matches_publication = fanout.application_message().is_some_and(|application| {
+                    application.app_event_id == published.app_event_id
+                        && outcome.message_id == published.message_id
+                });
+                if matches_publication
+                    && outcome.accepted_targets > 0
+                    && outcome.fanout_complete
+                    && !matches!(fanout.mls_state(), FanoutMlsState::Pending(_))
+                {
+                    self.session.delete_outbound_fanout(fanout.message_id())?;
+                }
+            }
+        }
+        Ok(())
+    }
+
     async fn resume_outbound_fanouts_for_group(
         &mut self,
         group_id: Option<&GroupId>,
@@ -2238,6 +2271,9 @@ where
                 || matches!(fanout.mls_state(), FanoutMlsState::Pending(_))
             {
                 Box::pin(self.drive_outbound_fanout(fanout, &mut output, &mut queue, None)).await?;
+            } else if outcome.accepted_targets > 0 && fanout.application_message().is_some() {
+                record_published_application_fanout(&fanout, &mut output);
+                output.fanout.push(outcome);
             } else {
                 self.session.delete_outbound_fanout(fanout.message_id())?;
             }
@@ -3149,7 +3185,6 @@ where
     ) -> AccountResult<PublishStatus> {
         self.resolve_outbound_fanout_mls(&mut fanout, output, queue, context.clone())
             .await?;
-        let application_was_accepted = fanout.outcome().accepted_targets > 0;
         let endpoints = fanout.request().target.endpoints().to_vec();
         let now_ms = self.wall_clock.now().0.saturating_mul(1_000);
         let due = fanout
@@ -3304,31 +3339,23 @@ where
                 });
             }
         }
-        if let Some(application) = fanout.application_message() {
-            if status.accepted_by_any_endpoint && !application_was_accepted {
-                output
-                    .published_app_messages
-                    .push(PublishedApplicationMessage {
-                        group_id: application.group_id.clone(),
-                        app_event_id: application.app_event_id.clone(),
-                        message_id: report.message_id.clone(),
-                        source_epoch: application.source_epoch,
-                        retention: application.retention,
-                    });
-            } else if status.retry_deferred && !status.accepted_by_any_endpoint {
-                output
-                    .unresolved_app_messages
-                    .push(UnresolvedApplicationMessage {
-                        group_id: application.group_id.clone(),
-                        app_event_id: application.app_event_id.clone(),
-                        message_id: report.message_id.clone(),
-                        reason: if possible_ambiguous_exposure {
-                            UnresolvedPublishReason::AcknowledgementUnknown
-                        } else {
-                            UnresolvedPublishReason::RetryableUnavailable
-                        },
-                    });
-            }
+        if status.accepted_by_any_endpoint {
+            record_published_application_fanout(&fanout, output);
+        } else if let Some(application) = fanout.application_message()
+            && status.retry_deferred
+        {
+            output
+                .unresolved_app_messages
+                .push(UnresolvedApplicationMessage {
+                    group_id: application.group_id.clone(),
+                    app_event_id: application.app_event_id.clone(),
+                    message_id: report.message_id.clone(),
+                    reason: if possible_ambiguous_exposure {
+                        UnresolvedPublishReason::AcknowledgementUnknown
+                    } else {
+                        UnresolvedPublishReason::RetryableUnavailable
+                    },
+                });
         }
         if fanout_outcome.outstanding_targets > 0
             && let Some(group_id) = fanout.group_id()
@@ -3343,6 +3370,7 @@ where
         output.fanout.push(fanout_outcome.clone());
         if fanout_outcome.fanout_complete
             && !matches!(fanout.mls_state(), FanoutMlsState::Pending(_))
+            && !(fanout_outcome.accepted_targets > 0 && fanout.application_message().is_some())
         {
             self.session.delete_outbound_fanout(fanout.message_id())?;
         }
@@ -3866,6 +3894,31 @@ fn frozen_fanout_report(fanout: &OutboundFanout) -> TransportPublishReport {
         failed,
         required_acks: fanout.request().required_acks,
     }
+}
+
+/// Replay the app-visible acknowledgement derived from a durable accepted
+/// fanout. The projection is an idempotent upsert, so emitting this on every
+/// recovery pass is safe until the app explicitly acknowledges persistence.
+fn record_published_application_fanout(fanout: &OutboundFanout, output: &mut AccountDeviceEffects) {
+    if fanout.outcome().accepted_targets == 0 {
+        return;
+    }
+    let Some(application) = fanout.application_message() else {
+        return;
+    };
+    let message_id = fanout
+        .published_message_id()
+        .unwrap_or_else(|| fanout.message_id())
+        .clone();
+    output
+        .published_app_messages
+        .push(PublishedApplicationMessage {
+            group_id: application.group_id.clone(),
+            app_event_id: application.app_event_id.clone(),
+            message_id,
+            source_epoch: application.source_epoch,
+            retention: application.retention,
+        });
 }
 
 fn apply_report_to_fanout(
--- a/crates/marmot-account/src/runtime.rs
+++ b/crates/marmot-account/src/runtime.rs
@@ -2022,10 +2022,17 @@ where
         let (mut output, blocked_groups) = self
             .resume_outbound_fanouts_for_group(Some(group_id))
             .await?;
-        if blocked_groups.contains(group_id) {
-            return Ok(output);
-        }
-        let effects = self.session.advance_convergence(group_id).await?;
+        // An older deferred transport fanout blocks only newly queued
+        // outbound intents. Convergence inputs must still settle: otherwise a
+        // permanently unavailable relay can wedge epoch progression and fork
+        // healing forever. Any protocol effects produced by settlement remain
+        // observable while the queued-intent drain stays ordered behind the
+        // frozen fanout.
+        let effects = if blocked_groups.contains(group_id) {
+            self.session.advance_convergence_inputs(group_id).await?
+        } else {
+            self.session.advance_convergence(group_id).await?
+        };
         output.extend(self.publish_session_effects(effects).await?);
         Ok(output)
     }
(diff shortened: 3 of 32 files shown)


crates/cgka-engine/src/engine.rs
```diff
@@ -767,6 +767,20 @@ impl<S: StorageProvider> Engine<S> {
...
+    /// Read one live group's frozen fanouts in original staging order without
+    /// scanning every other group's durable outbox.
+    pub fn outbound_fanouts_for_group(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<Vec<OutboundFanout>, EngineError> {
+        if self.quarantined_groups.contains_key(group_id)
+            || self.unhydrated_groups.contains(group_id)
+        {
+            return Ok(Vec::new());
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in 794d5669. AccountDeviceRuntime::advance_convergence now hydrates the target group before the group-scoped durable-fanout lookup. The scheduled_advance_hydrates_before_resuming_frozen_fanout regression reopens with deferred hydration and proves that the same convergence call publishes the exact frozen event, completes the fanout, and confirms MLS state.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1003,18 +1003,61 @@ pub(crate) fn chat_list_activity_filter_sql(column_prefix: &str) -> String {
...
+        "CASE
+            WHEN {column_prefix}direction = 'sent'
+             AND {column_prefix}invalidation_status = 'local_publish_failed' THEN 0
+            ELSE 1
+         END DESC,
+         CASE
+            WHEN {column_prefix}direction = 'sent'
+             AND {column_prefix}source_epoch IS NULL
+             AND {column_prefix}source_message_id_hex IS NULL
+             AND {column_prefix}invalidation_status IS NULL
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in 794d5669. All three preview-order predicates now use a missing source_message_id_hex as the pending signal, matching the row mapper even when source_epoch is already known. The pending_send_with_source_epoch_uses_pending_preview_ordering regression covers the reachable split state.

crates/marmot-account/src/runtime.rs
```diff
@@ -3126,6 +3149,7 @@ where
     ) -> AccountResult<PublishStatus> {
         self.resolve_outbound_fanout_mls(&mut fanout, output, queue, context.clone())
             .await?;
+        let application_was_accepted = fanout.outcome().accepted_targets > 0;
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in eca816a0. Accepted application fanouts now remain durable until the app projection explicitly acknowledges persistence. Restart recovery replays the exact publication metadata idempotently, and acknowledgement retires only terminal accepted fanouts. The updated runtime regression covers acceptance, restart replay, acknowledgement, and no further replay.

crates/marmot-app/src/client/sync.rs
```diff
@@ -1035,8 +1057,18 @@ impl AppClient {
...
+        // can publish an accepted-pending application message without emitting
+        // any engine event, so it must be projected before the eventless fast
+        // path or publish-failure gate below. A batch can contain one fanout
+        // that succeeded alongside another publish that failed; the successful
+        // row must not remain stuck in `Sending` after its fanout is deleted.
+        self.remember_published_reports(effects);
+        summary
+            .projection_updates
+            .extend(self.finalize_published_app_message_source_retention(effects)?);
+        fail_if_publish_failed(effects)?;
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in eca816a0. Both drained and scheduled mixed-batch seams now preserve finalized pending-to-sent projection updates in the client's pending broadcast queue before returning the unrelated publish error. The shared regression now verifies storage finalization and the retained broadcast update for drained, scheduled, and manual retry paths.

crates/transport-nostr-adapter/src/sdk_client.rs
```diff
@@ -1035,6 +1099,14 @@ impl NostrSdkRelayClient {
...
         match self.client.add_write_relay(endpoint.clone()).await {
             Ok(true) => {
+                if self.pin_new_relay_retry_interval(endpoint).await.is_err() {
+                    return Err(TransportEndpointFailure {
+                        endpoint: transport_endpoint,
+                        reason: "configure publish relay failed".to_owned(),
+                        kind: TransportEndpointFailureKind::RetryableUnavailable,
+                        rejection_category: None,
+                    });
+                }
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in eca816a0 with state-aware cleanup. If retry-interval pinning fails but the relay remains registered, retain_publish_relay records cleanup ownership and continues with a privacy-safe warning. If pinning already removed the relay, it remains unavailable and is not falsely tracked. The regression covers both failure points plus lease cleanup.

crates/marmot-app/src/client/projection.rs
```diff
@@ -590,6 +590,8 @@ impl AppClient {
                 updates.push(update);
             }
         }
+        self.runtime
+            .acknowledge_published_app_messages(&effects.published_app_messages)?;
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in e0c0f2c5. Published application-message acknowledgement is cleanup after the delivery projection is already durable, so cleanup failure is now logged as a privacy-safe deferred cleanup instead of discarding the delivery update or returning a false send failure. The accepted fanout remains durable and the next recovery drain retries the idempotent acknowledgement. A focused fault-injection regression verifies that the committed pending-to-delivered update is still returned to subscribers.

crates/transport-nostr-adapter/src/sdk_client.rs
```diff
@@ -941,19 +958,34 @@ impl NostrSdkRelayClient {
...
                                 "publish batch timed out".to_owned(),
                             )));
                             request_durations[index] = batch_started_at.elapsed();
                             continue;
                         }
+                        // Requests admitted during one concurrency window share
+                        // its deadline. Once that window expires, later
+                        // backpressured requests receive the next bounded
+                        // window, capped by the whole-batch deadline.
+                        if now >= request_window_deadline {
```
Datawav: [P2] Give a backpressured request a fresh per-event deadline

The deadline advances only after it has already expired. With more than `SDK_RELAY_BATCH_MAX_IN_FLIGHT` requests, if one of the first four completes just before the 20-second window expires, the newly admitted fifth request inherits that same nearly-expired deadline and can time out immediately even when its relay is healthy; the 60-second batch budget is still available. The current tests exercise shared connection, concurrency, ordering, and partial failure, but not a refill near the window boundary. Start a fresh `min(now + SDK_RELAY_PUBLISH_OVERALL_WAIT, batch_deadline)` budget for requests admitted after backpressure (or explicitly track cohorts rather than wall-clock expiry), and add a paused-time five-request regression where the first cohort releases a slot near its deadline.

Datawav (changes requested):
Reviewed the complete 24-file diff at exact head `e0c0f2c5be8b89bd3edf9c637c1dab95c58af6ce`, the surrounding fanout, projection, scheduler, SQLite, C/UniFFI, and Nostr adapter paths, existing review threads, and current CI. All required GitHub checks are green, but two reachable correctness regressions remain.

**[P1] Preserve the current delivered row when only a sibling publish failed** (`crates/marmot-app/src/client/mod.rs:3010-3033`). `AccountDeviceEffects` aggregates every publish produced by the convergence pass, but `observe_recovery_evidence_then_gate_send_publish` fails on any entry in `effects.failures` before this send's `app_event_id` is matched against `published_app_messages`. A mixed batch can therefore contain the current event as accepted plus an unrelated failed fanout; the branch returns `Err` and `retract_failed_local_projection` marks the already-delivered row `local_publish_failed`. Later recovery finalization fills source/retention fields but does not clear that invalidation, so the UI can remain failed and the caller may resend duplicate content. Existing mixed-batch tests cover drain/scheduled/retry observation paths, not this direct-send gate. Match/finalize the current event before applying sibling failures, retract only when this exact event was not published, and add a direct-send regression with one accepted current event plus one failed sibling.

**[P2] Give backpressured batch requests a fresh per-event deadline.** Details inline.

Both need focused regressions before approval.

crates/marmot-account/src/runtime.rs
```diff
@@ -2218,7 +2219,50 @@ where
...
+        &mut self,
+        group_id: Option<&GroupId>,
+    ) -> AccountResult<AccountDeviceEffects> {
+        let fanouts = match group_id {
+            Some(group_id) => self.session.outbound_fanouts_for_group(group_id)?,
+            None => self.session.outbound_fanouts()?,
+        };
         let mut output = AccountDeviceEffects::default();
         let mut queue = VecDeque::new();
         for fanout in fanouts {
```
Datawav, line 2325: [P2] Stop at the first deferred fanout for this group

The query returns original staging order, but this loop does not preserve delivery-attempt order when an older record is still in backoff. `drive_outbound_fanout` can make no attempt because the first fanout is `PossiblyExposed` or otherwise not yet retry-due, then the loop immediately drives the next `NotAttempted` fanout. The recipient can therefore observe message 2 before message 1, despite this PR's stated guarantee to drain retained fanouts before newer group work. After driving each record, re-read/inspect whether it remains incomplete or deferred and stop processing later fanouts for that group until the oldest obligation is due/resolved. Add a two-fanout regression where the first has a future retry cutoff and assert the second is not published during that pass.

Datawav (changes requested):
One additional ordering gap surfaced while reconciling the group-scoped durable-fanout pass: a deferred older fanout does not currently gate later fanouts from the same group. This contradicts the PR's ordered-retry guarantee and needs a focused regression.

crates/marmot-app/src/client/projection.rs
```diff
@@ -590,6 +590,32 @@ impl AppClient {
...
+        {
+            Err(AppError::BlockingTask(
+                "injected published application-message acknowledgement failure".to_owned(),
+            ))
+        } else {
+            self.runtime
+                .acknowledge_published_app_messages(&effects.published_app_messages)
+                .map_err(AppError::from)
+        };
+        if let Err(error) = acknowledgement {
```
Datawav, line 604: [P2] Schedule cleanup after a failed fanout acknowledgement

This logs and swallows the acknowledgement failure, but the retained fanout is already terminal: accepted targets are nonzero, while outstanding targets and pending MLS state are both absent. `has_pending_outbound_fanouts` therefore excludes it, so no convergence timer guarantees the promised next recovery drain. The accepted fanout can remain stored indefinitely until an unrelated drain or restart. Enqueue/persist a local cleanup-retry edge here (without republishing network bytes), and add a regression that injects one acknowledgement failure and proves cleanup runs again without unrelated activity.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1003,18 +1003,58 @@ pub(crate) fn chat_list_activity_filter_sql(column_prefix: &str) -> String {
...
+        "CASE
+            WHEN {column_prefix}direction = 'sent'
+             AND {column_prefix}invalidation_status = 'local_publish_failed' THEN 0
+            ELSE 1
+         END DESC,
+         CASE
+            WHEN {column_prefix}direction = 'sent'
+             AND {column_prefix}source_message_id_hex IS NULL
+             AND {column_prefix}invalidation_status IS NULL
+             AND NOT EXISTS (
```
Datawav, line 1075: [P2] Keep pending-preview demotion durable across secure pruning

The demotion exists only while a newer accepted `message_timeline`/`app_events` witness is still present. Secure retention pruning deletes that witness and reruns this ordering, so the older unresolved pending row becomes the preview head again even though newer authenticated activity had already displaced it. `retained_activity_sort_at` preserves only the activity clock, not the accepted insertion-order boundary. Persist a per-chat accepted-activity high-water mark (with migration/backfill), compare pending candidates against it, and add a secure-prune regression proving the stale pending preview does not reappear.

Datawav (changes requested):
The completed clean-context runtime and SQLite passes found three more reachable gaps at this same head. Two are attached inline. The third is in `ScheduledConvergence`: an outstanding frozen fanout is mapped to `PendingOutbound`, but that state is always rearmed at the normal settlement delay instead of the fanout's durable retry cutoff. During an outage the worker therefore repeatedly reloads the fanout and discovers nothing is due, even though network attempts remain correctly backed off. Expose the earliest durable target retry time through the schedule state and arm that deadline directly; `notify_connectivity_restored` should remain the explicit path that pulls it forward. Add a paused-time regression proving no worker polling occurs before the durable retry cutoff.

crates/transport-nostr-adapter/src/sdk_client.rs
mubarakcoded: Fixed in 67100431. Batch refills now receive a fresh per-event deadline capped by the 60-second batch ceiling, while the first cohort retains its original end-to-end budget. A paused-time five-admission regression covers a slot reopening one millisecond before the first deadline.

crates/marmot-account/src/runtime.rs
mubarakcoded: Fixed in 67100431. The drain now installs a group-scoped barrier whenever the oldest fanout remains deferred, while continuing unrelated groups. The two-fanout regression proves the newer NotAttempted event is untouched and that scheduling follows the older 30-second cutoff.

crates/marmot-app/src/client/projection.rs
mubarakcoded: Fixed in 67100431. Accepted application fanouts remain scheduler-visible until acknowledgement succeeds; an acknowledgement failure creates a cleanup edge, and resume replays only publication metadata without network bytes. The one-shot fault regression proves autonomous cleanup retires the fanout without republishing.

crates/storage-sqlite/src/chat_list.rs
mubarakcoded: Fixed in 67100431 with migration 0056. Chat rows now retain an accepted-activity insertion-order high-water mark, preview eligibility consults it after secure pruning, and migration backfills existing accepted history. The prune regression proves the displaced pending preview does not return after its accepted witness is deleted or after projection repair.

mubarakcoded:
Addressed the two body-only Datawav findings in signed commit 67100431 as well:\n\n- Direct-send gating is now scoped to the current group/app-event identity. A published or durably unresolved current message is preserved even when the same aggregate effects batch contains an unrelated sibling failure, preventing a false failed state and duplicate user retry.\n- Scheduled convergence now derives its next wake from the oldest same-group frozen fanout's persisted retry cutoff. Connectivity-restored wakes still pull retryable-unavailable work forward, while ordinary outage backoff no longer reloads the fanout every settlement interval.\n\nFocused regressions cover both paths, alongside the four resolved inline findings. Formatting, targeted checks, and clippy for all touched crates are clean. No generated bindings are included.

Datawav (commented):
## Automated review

Two actionable findings.

1. Minor — Successful sibling finalization is deferred on a direct-send error
- Exact changed-line evidence: `crates/marmot-app/src/client/mod.rs:3118-3127`, in `observe_recovery_evidence_then_gate_send_publish`, returns the current send’s `fail_if_publish_failed` error without first applying `finalize_published_app_message_source_retention` to successful entries in `effects.published_app_messages`.
- Triggering state: one aggregate effects batch contains a failed current send and a successfully published sibling application message.
- Consequence: the sibling remains durably accepted but its source/retention projection is not finalized until a later `resume_outbound_fanouts` recovery pass. This temporarily leaves subscriber-visible state stale; durable message data is not lost and network bytes need not be republished.
- Confidence: High.
- Smallest sound correction: before returning the current send’s error, finalize successful sibling entries and buffer their updates in `pending_projection_updates`, matching the drain, scheduled-convergence, and retry seams. Add a mixed-batch regression proving the sibling update survives the error path.

2. Minor — PR description incorrectly says no schema migration
- Exact changed-line evidence: `crates/storage-sqlite/src/migrations.rs:400-404` registers migration version 56; `crates/storage-sqlite/src/migrations/0056_chat_list_accepted_activity_high_water.rs:14-18` executes `ALTER TABLE chat_list_rows ADD COLUMN accepted_activity_insert_order INTEGER NOT NULL DEFAULT 0`, followed by a backfill. This contradicts the PR body’s “No database schema migration” statement.
- Triggering state: release, upgrade, or rollback review relies on the PR description rather than reconstructing the migration set.
- Consequence: upgrade scope and release notes are inaccurate. The supplied patch itself is additive, defaulted, and backfilled; no runtime migration defect was identified.
- Confidence: High.
- Smallest sound correction: update the PR description to disclose additive, backfilled migration 0056.

Audit evidence:
- Exact head: `67100431ec13732897e9f6049d8bc3265ad2fec9`
- Complete patch coverage: 27 files, including `crates/marmot-app/src/client/mod.rs`, `crates/marmot-account/src/runtime.rs`, `crates/storage-sqlite/src/migrations/0056_chat_list_accepted_activity_high_water.rs`, and `crates/transport-nostr-adapter/src/sdk_client.rs`.
- Discussion classification: earlier findings concerning hydration ordering, group-scoped fanout barriers, retry cutoffs, acknowledgement cleanup scheduling, secure-pruning high-water state, fresh batch-refill deadlines, and relay-pin cleanup are addressed at this head. The C-header string/array safety wording is copy-pasted but stylistic and non-actionable. The successful-sibling finalization claim remains supported.
- No physical-device evidence exists for this head: `automated_device_gate` is null. Conclusions therefore rest only on static review and supplied CI evidence.

Risk lenses checked:
- persistence/migration — real-state upgrade/rollback, partial failure and idempotent retry, data preservation
- concurrency/lifecycle — cancellation and stale callbacks, ordering/race windows, process or lifecycle restart
- network/protocol — timeout/retry/duplicate delivery, partial or malformed peer data, version and wire compatibility
- tests — pre-fix failure behavior, boundary and negative cases, production-path coverage

Check evidence:
- Bundle reports `all_completed: true` and `all_successful: true`.
- Successful checks include all four Rust test shards, Rust check, Rust clippy, Rust format, Rust audit, Relay CLI E2E, Relay runtime tests, CLI e2e with mock relay, C bindings smoke, Simulator and vectors, and Tamarin proofs.
- `Publish wn-agent release` was skipped and supplies no correctness evidence.

crates/marmot-account/src/runtime.rs
```diff
@@ -2218,15 +2262,80 @@ where
...
+                    // fanouts in this group, but must not block unrelated
+                    // groups during an account-wide drain.
+                    blocked_groups.insert(fanout_group);
+                }
+            } else if outcome.accepted_targets > 0 && fanout.application_message().is_some() {
+                record_published_application_fanout(&fanout, &mut output);
+                output.fanout.push(outcome);
             } else {
                 self.session.delete_outbound_fanout(fanout.message_id())?;
             }
```
jgmontoya, line 2364: From my agent:

> When a resumed fanout resolves with `outstanding_targets == 0` but *every* target ended `Failed` (`accepted_targets == 0`), this branch deletes the fanout with no output signal at all — no `PublishFailure`, no `unresolved_app_messages` entry. In the normal same-call path (`drive_outbound_fanout`), a `PublishFailure` is pushed to `output.failures` before the terminal delete, so the app is told. But if a crash lands after `put_outbound_fanout` persists the all-`Failed` state and before that call's own delete runs, a later resume drain finds this fully-resolved, all-failed, application-message-carrying fanout on disk and silently deletes it here — the timeline row stays stuck in `Sending` forever with no path to ever learn it failed. This is precisely the failure mode the PR sets out to eliminate, just relocated to the crash-recovery path instead of the happy path. Recommend re-emitting `PublishFailure`/`unresolved_app_messages` for a terminal all-failed fanout with `application_message` set before deleting it here, mirroring `drive_outbound_fanout`.

crates/marmot-app/src/client/mod.rs
```diff
@@ -3092,23 +3096,56 @@ impl AppClient {
...
+            Ok(updates) => self.pending_projection_updates.extend(updates),
+            Err(error) => {
+                tracing::warn!(
+                    target: "marmot_app::messages",
+                    method = "send_app_event_with_local_projection",
+                    error_kind = error.privacy_safe_kind(),
+                    "failed to finalize a successful sibling from a failed send batch"
+                );
+            }
+        }
```
jgmontoya, line 3163: From my agent:

> `fail_if_publish_failed` inspects the *whole* effects batch (`crates/marmot-app/src/groups.rs:2221`), not just the current message. When the current send is retained, this path returns `Ok(())` without ever calling `fail_if_publish_failed` — so an unrelated hard publish failure for a different app-event in the same batch is dropped with no `Err`, no warning, and no local row ever getting retracted for the failed sibling. Every other observation seam that can see a mixed batch (`observe_drained_session_events`, `observe_scheduled_convergence_effects`, `observe_convergence_retry_effects`) propagates the unrelated failure as `Err` so it can be handled. This one is the outlier.
> 
> This is confirmed by the PR's own test, `crates/marmot-app/src/tests.rs:13384-13421` (`assert_mixed_publish_batch_finalizes_successful_message`, `MixedPublishObservation::DirectSend`): it asserts `result.is_ok()` for exactly this scenario but never asserts anything about the unrelated failed message — no local row is even created for it in the test, so the test cannot observe whether it gets stuck. Please confirm (and comment, if intentional) why a hard-failed sibling can never reach this seam in practice, or thread the failure through the same way the other four seams do.

crates/storage-sqlite/src/timeline.rs
```diff
@@ -1240,36 +1240,63 @@ fn retain_pruned_chat_activity_tx(
...
+                CASE
+                    WHEN timeline.invalidation_status IS NULL
+                     AND NOT (
+                         timeline.direction = 'sent'
+                         AND timeline.source_message_id_hex IS NULL
+                     ) THEN source.insert_order
+                    ELSE NULL
+                END
+         FROM message_timeline AS timeline
+         JOIN app_events AS source
```
jgmontoya, line 1260: > Every other `message_timeline` → `app_events` join in this file (`timeline.rs:1039`, `:2561`, `:2792`) uses `LEFT JOIN` specifically so a `message_timeline` row without a matching `app_events` row is still counted. Here it's a plain (inner) `JOIN`, and pre-PR this function selected straight from `message_timeline` with no join at all — it always picked up `timeline_at` for every pruned row. Now, if a pruned message's `message_timeline` row lacks an `app_events` counterpart, `query_row` returns nothing for it, and it silently contributes nothing to `latest_pruned_activity` / `accepted_activity_insert_order`. That's a regression in the secure-prune retention path unless there's a hard invariant guaranteeing every `message_timeline` row always has a matching `app_events` row — that invariant isn't visible in this file's schema or asserted nearby. Recommend `LEFT JOIN` for consistency with the sibling queries, or a comment/assertion establishing why inner join is safe here.

crates/transport-nostr-adapter/src/sdk_client.rs
```diff
@@ -1000,17 +1031,79 @@ impl NostrSdkRelayClient {
...
             .map_err(|_| TransportAdapterError::Subscription("add relay failed".to_owned()))?;
+        if added {
+            self.pin_new_relay_retry_interval(&endpoint)
+                .await
+                .map_err(|_| {
+                    TransportAdapterError::Subscription("configure relay failed".to_owned())
+                })?;
+        }
         Ok(())
     }
```
jgmontoya, line 1114: > `pin_new_relay_retry_interval` removes and re-adds a relay purely to set a cosmetic reconnect-cadence value. In `retain_publish_relay`, a pin failure that leaves the relay still registered is tolerated (warn + `Ok(true)`, relay stays usable at SDK-default cadence). `add_subscription_relay` has no such fallback: any pin failure hard-fails the entire subscription-relay setup, even when the relay is still present and perfectly usable with the default (slower) retry interval instead of the pinned 5s one. Since this PR's whole goal is making startup/catch-up more resilient to transient failures, this asymmetry can turn a benign cosmetic-config failure into a hard subscription failure on exactly the path being hardened — the opposite of intent. There's no test covering `add_subscription_relay`'s pin-failure path (only `retain_publish_relay`'s, via `publish_relay_pin_failure_tracks_only_a_relay_that_remains_registered`). Recommend mirroring the "still-registered → warn and continue" fallback here.

crates/marmot-account/src/runtime.rs
```diff
@@ -2006,8 +2010,20 @@ where
...
+        // fanouts until full hydration restores its pending lifecycle. Hydrate
+        // before the targeted lookup so this same scheduled pass can resume
+        // those exact frozen bytes instead of waiting for another wake.
+        let _ = self.session.ensure_group_hydrated(group_id)?;
+        // Retry the group's already-frozen transport events before producing
+        // any newer convergence work. This preserves send order and ensures a
+        // scheduled outbound wake actually drives the exact retained bytes.
+        let mut output = self
+            .resume_outbound_fanouts_for_group(Some(group_id))
+            .await?;
```
jgmontoya, line 2024: > `resume_outbound_fanouts_for_group` can leave a group "blocked" when the oldest frozen fanout is still backed off, but the immediately-following `advance_convergence` + `publish_session_effects` isn't checked against that block and can still publish new convergence-derived work for the same group. This may be intentional (different work categories), but no test asserts new convergence work is deferred while an older fanout in the same group is backed off. If the "drain older-first, in order" guarantee is meant to be strict, this is a gap; otherwise a short comment would help the next reader.

Datawav (dismissed):

Datawav (dismissed):

crates/marmot-account/src/runtime.rs
```diff
@@ -2006,8 +2011,23 @@ where
...
+        // scheduled outbound wake actually drives the exact retained bytes.
+        let (mut output, blocked_groups) = self
+            .resume_outbound_fanouts_for_group(Some(group_id))
+            .await?;
+        if blocked_groups.contains(group_id) {
+            return Ok(output);
+        }
         let effects = self.session.advance_convergence(group_id).await?;
-        self.publish_session_effects(effects).await
+        output.extend(self.publish_session_effects(effects).await?);
```
jgmontoya, line 2036: This gate can wedge a group's convergence permanently

session.advance_convergence resolves to converge_and_drain_queued_outbound_intents, which does two things: advance_convergence_inputs_until_settled (deferred-commit settlement, fork healing, epoch progression) and the queued-outbound-intent drain. Send ordering only needs the second suppressed; this early return skips both.
That would be a bounded delay if retry_deferred always cleared. It doesn't. retry_deferred is outstanding_targets > 0; PossiblyExposed and RetryableUnavailable both count as outstanding (crates/traits/src/transport_adapter.rs:191); and frozen_fanout_target_retry_delay_ms (crates/marmot-account/src/runtime.rs:3940) has no attempt cap and no exhaustion path — a target reaches Failed only on an explicit transport rejection. There is no age-based fanout abandonment either.

Concretely: a group with three relays where one host goes away for good. The fanout settles at accepted=2, outstanding=1, required_acks is already met, but fanout_complete is false so the record is never deleted. From then on every advance_convergence for that group returns here. Deferred commits never settle and a fork can never heal — invisible except to the stall detector.

deferred_fanout_blocks_newer_same_group_fanouts_and_convergence_work only asserts that the queued intent stays queued (crates/marmot-account/tests/runtime/frozen_fanout.rs:565). Nothing there requires settlement to stop, so this gate is stronger than the behaviour under test.

Smallest correct fix: drop this early return and keep the in-loop blocked_groups barrier, which already satisfies the per-group send-ordering finding at :2318. If queued-intent ordering is also wanted, thread the barrier into the engine so it settles convergence inputs and skips only the queued-outbound drain — ordering shouldn't be bought by skipping settlement.

crates/storage-sqlite/src/migrations/0056_chat_list_accepted_activity_high_water.rs
```diff
@@ -0,0 +1,46 @@
...
+pub(crate) fn apply(tx: &Transaction<'_>) -> StorageResult<()> {
+    tx.execute_batch(
+        r#"
+ALTER TABLE chat_list_rows
+    ADD COLUMN accepted_activity_insert_order INTEGER NOT NULL DEFAULT 0;
+"#,
+    )
+    .storage()?;
+
+    let activity_filter = crate::chat_list::chat_list_activity_filter_sql("accepted.");
```
jgmontoya: We shouldn't reference this code from the migration, if `chat_list_activity_filter_sql` changes in the future then this might behave differently on new installs vs existing ones.

jgmontoya (changes requested):

Datawav (dismissed):

crates/storage-sqlite/src/timeline.rs
```diff
@@ -1240,36 +1240,68 @@ fn retain_pruned_chat_activity_tx(
     group_id_hex: &str,
     pruned_message_ids: &BTreeSet<String>,
 ) -> StorageResult<()> {
-    let activity_filter = crate::chat_list::chat_list_activity_filter_sql("");
+    let activity_filter = crate::chat_list::chat_list_activity_filter_sql("timeline.");
+    // Every id in this set was selected from `app_events` earlier in the
+    // caller's same IMMEDIATE transaction, and retention runs before the
+    // source rows are scrubbed or deleted. Source-row existence is therefore
+    // an invariant here; the inner join makes a violation visible instead of
```
jgmontoya: > The invariant holds, but this comment claims a behaviour the code doesn't have. The query is consumed with .optional() at :1282, so an orphaned message_timeline row yields None and is silently skipped — nothing about the violation becomes visible. Either drop the second half of the sentence, or make it true (a debug_assert! or privacy-safe warn on the None branch). The first half — ids come from app_events in the caller's same IMMEDIATE transaction — is the real justification and stands on its own.

crates/marmot-account/tests/runtime/frozen_fanout.rs
```diff
@@ -217,6 +217,384 @@ async fn frozen_fanout_ambiguous_adapter_errors_remain_unresolved() {
...
+        policy,
+        RecordingKeyPackages::default(),
+    )
+    .with_maintenance_sources(
+        wall,
+        Arc::new(TestMonotonicClock::default()),
+        Arc::new(TestRandom::new(0)),
+    );
+
+    tokio::time::sleep(Duration::from_millis(1_100)).await;
```
jgmontoya: > Derive this like the sibling at :682 (convergence_delay.saturating_add(1)) rather than hardcoding 1_100. As written the test spends 1.1s of wall clock and silently depends on an unstated constant — if the settlement delay moves, this starts passing or failing for the wrong reason.

crates/transport-nostr-adapter/src/sdk_client.rs
```diff
@@ -1000,17 +1031,85 @@ impl NostrSdkRelayClient {
...
-        self.client
-            .add_relay(endpoint)
+        let added = self
+            .client
+            .add_relay(endpoint.clone())
             .await
             // The sdk error Display can carry the relay URL; keep the error
             // operation-only so `TransportAdapterError` Display stays URL-free.
             .map_err(|_| TransportAdapterError::Subscription("add relay failed".to_owned()))?;
+        if added && self.pin_new_relay_retry_interval(&endpoint).await.is_err() {
```
jgmontoya, line 1100: > This mirrors retain_publish_relay's fallback, which is what I asked for. The branch still has no test though, while the publish-side one does (publish_relay_pin_failure_tracks_only_a_relay_that_remains_registered). Worth a sibling case so the defensive branch on the startup path isn't the one nobody exercises.

crates/cgka-session/src/lib.rs
```diff
@@ -997,6 +997,32 @@ impl AccountDeviceSession {
         Ok(self.collect_effects(results))
     }
 
+    /// Settle retained convergence inputs without draining queued outbound
+    /// intents.
+    ///
+    /// A host uses this narrower operation when an older durable transport
+    /// fanout still blocks newer user-authored sends. Protocol settlement,
+    /// epoch progression, and fork healing remain live while queued work stays
+    /// behind the fanout ordering barrier.
```
jgmontoya: > Worth naming what this path does not do. converge_and_drain_queued_outbound_intents also runs prepare_pending_disband and run_drain_maintenance, so while a group sits behind the fanout barrier a pending disband and a due SelfRemove auto-commit are deferred too — not only queued user sends. Deferring them looks correct under the ordering guarantee, since staging a maintenance commit would publish ahead of the older frozen event. But frozen_fanout_target_retry_delay_ms has no attempt cap and no age-based abandonment, so for a permanently unavailable target that deferral is unbounded. Follow-up coming for fanout exhaustion; this doc comment just needs to say maintenance and disband are deferred as well.

jgmontoya (dismissed):

Datawav (commented):
## Automated review

Two actionable findings.

1. Minor — Regression test hardcodes a wall-clock settlement delay
- Exact changed-line evidence: `crates/marmot-account/tests/runtime/frozen_fanout.rs:567` adds `tokio::time::sleep(Duration::from_millis(1_100)).await;` in `deferred_fanout_blocks_newer_sends_but_not_convergence_settlement`.
- Triggering state: the production convergence-settlement delay changes, or timing overhead varies under load.
- Consequence: the ordered-fanout regression can pass or fail for the wrong reason and adds 1.1 seconds of real-time latency to the suite.
- Confidence: High. The same patch’s sibling test derives its wait with `convergence_delay.saturating_add(1)`.
- Smallest sound correction: derive this wait from the settlement-delay value in the same way as the sibling test; alternatively use paused Tokio time and advance it deterministically.

2. Minor — New convergence-only API documentation omits other deferred work
- Exact changed-line evidence: `crates/cgka-session/src/lib.rs:997-1006` describes `advance_convergence_inputs` as settling convergence “without draining queued outbound intents” and says only “queued work stays behind the fanout ordering barrier.”
- Triggering state: an oldest frozen fanout remains outstanding for a permanently unreachable target; the retry policy has no attempt cap or age-based abandonment.
- Consequence: callers cannot tell from the API contract that pending disband work and due maintenance such as a SelfRemove auto-commit are also deferred, potentially without a time bound. The deferral is consistent with preserving fanout order, but its scope is materially broader than documented.
- Confidence: High. The full drain path also invokes `prepare_pending_disband` and `run_drain_maintenance`, while this narrower path bypasses them.
- Smallest sound correction: extend the new doc comment—or the adjacent runtime barrier comment—to name pending disband and maintenance deferral and note that it can remain deferred while the oldest fanout remains outstanding.

Audit evidence:
- Exact head: `ddea8bb4fc4bdb16fc891d9d4fdfc1cc3dfe5350`
- Complete supplied patch coverage: 32 changed files, including `crates/marmot-account/src/runtime.rs`, `crates/marmot-account/tests/runtime/frozen_fanout.rs`, `crates/cgka-session/src/lib.rs`, `crates/storage-sqlite/src/timeline.rs`, and `crates/transport-nostr-adapter/src/sdk_client.rs`.
- Discussion classification: the two findings above remain supported. The advisory claim about `retain_pruned_chat_activity_tx` overstating inner-join visibility is discarded as non-actionable wording: the supplied context supports the same-transaction source-row invariant, and changing that comment does not correct runtime behavior. Earlier major discussion claims are stale or resolved at this exact head by the current implementation and focused regressions, including terminal all-target failure replay, convergence settlement behind the fanout barrier, ordered per-group retries, projection cleanup scheduling, migration self-containment, fresh batch deadlines, and subscription-relay pinning fallback.

Risk lenses checked:

Check evidence:
- The bundle reports `all_completed: true` and `all_successful: true`.
- Successful checks include Rust check, Rust clippy, Rust format, Rust audit, all four Rust test shards, Relay runtime tests, CLI e2e with mock relay, Relay CLI E2E, Tamarin proofs, Simulator and vectors, C bindings smoke, and platform build jobs.
- `automated_device_gate.status` is `capability_checked_not_installable`; no physical-device evidence exists for this head. Conclusions therefore rest on the exact-head static review and supplied successful CI evidence.

Datawav (commented):
No actionable findings.

Audit evidence:
- Exact head: `dab43155c5d8efd1123edd852383cba032e73df5`
- Complete supplied patch: 3 files, including `crates/cgka-session/src/lib.rs`, `crates/marmot-account/tests/runtime/frozen_fanout.rs`, and `crates/storage-sqlite/src/timeline.rs`.
- The `advance_convergence_inputs` documentation now accurately identifies queued outbound intents, pending disband work, drain maintenance, and due self-remove auto-commit as potentially unbounded deferrals while convergence settlement, epoch progression, and fork healing remain live.
- `deferred_fanout_blocks_newer_sends_but_not_convergence_settlement` now derives its wait from `prepare_convergence_cutoff_delay_ms(&group_id)` and sleeps through `convergence_delay.saturating_add(1)`, removing the brittle hardcoded 1,100 ms assumption while exercising production convergence behavior.
- The `retain_pruned_chat_activity_tx` comment now states only the supported same-transaction source-row invariant and no longer claims that the inner join makes violations visible.
- The corresponding latest discussion requests are therefore resolved by exact changed lines. Earlier claims outside this three-file follow-up patch, already-addressed claims, and claims lacking enough bounded source context to re-establish against this delta were not carried forward.

Risk lenses checked:
- `persistence/migration`: upgrade and rollback from real prior state; partial failure and idempotent retry; data preservation. This follow-up changes no schema, migration, query behavior, or persisted data path; the timeline edit is comment-only.
- `tests`: assertions fail before the fix; boundary and negative cases; test exercises production behavior. The changed regression retains its convergence-settlement assertions and now obtains the boundary delay from the production storage/session API. The timing-only follow-up does not independently demonstrate a pre-fix behavioral failure, but it removes the unstated timing constant without weakening coverage.

Check evidence:
- Supplied checks are all completed. `Rust format`, `Rust check`, `Rust clippy`, `Rust audit`, `Rust tests (1/4)`, `Rust tests (2/4)`, `Rust tests (4/4)`, relay tests, simulator/vectors, bindings smoke, and the listed build and compatibility jobs report `success`.
- `Rust tests (3/4)` reports conclusion `failure`; the bundle supplies no failing test or log and no changed-line evidence connecting that result to this follow-up delta. The advisory CI note is therefore retained as an evidence limitation, not converted into a code finding.
- The automated device gate reports `capability_checked_not_installable` and is bound to the earlier SHA `ddea8bb4fc4bdb16fc891d9d4fdfc1cc3dfe5350`, not this exact head. No physical-device evidence exists for this head. Conclusions rest on static review and the supplied CI evidence.

Not approved on this head: failed CI: Rust tests (3/4) (failure). A corrected head is re-audited automatically.

Datawav (approved):

PR #1608, opened 2026-08-31
Fix/upsert invalidation revive

jgmontoya:
Two fixes to how app-event re-recording interacts with invalidation.

1. Storage: record_app_event's ON CONFLICT upsert unconditionally cleared invalidated and invalidation_reason, so a row tombstoned in one batch came back live in a later one. Relay redelivery, backfill replay, rejoin reprocessing, and synthesized kind-1210 rows (deterministic ids) all re-record the same (group, message_id). Fork-loser messages reappeared in timelines with their reason erased, so two members that agreed on CGKA state rendered different histories. #1593's synthesizer filter only closed the same-batch shape. The upsert now preserves both columns: every convergence withdrawal is terminal, and the upsert cannot tell a retry from a replay.
2. App: with the upsert terminal, the one legitimate revival needed an explicit path. Inner event ids are NIP-01 hashes with second-granular created_at and no nonce, so resending identical text in the same second as a failed send mints the same id and lands on the row retract_failed_local_projection just tombstoned as local_publish_failed. Without an escape hatch the retry publishes to the whole group while the sender's copy renders Failed forever (and a retried delete stays hidden for peers, visible locally). The send path now calls clear_local_publish_failure before the optimistic pre-publish record. The primitive is scoped to direction = sent, invalidated = 1, reason = local_publish_failed; the fresh send intent is the evidence, and replay seams never enter the send path, so terminal-group swept sends (#1604) stay dead.

Tests: cross-batch tombstone survival for all four invalidation flavors, a live-row control, kind/tags-changing re-record, the with_retention variant, mutation-verified scoping tests for the clear primitive, and an end-to-end same-second resend test at the app layer. The chat-list preview test now pins both directions: bare re-record stays failed, cleared retry goes delivered.

## Summary by CodeRabbit

* **Bug Fixes**
  * Retrying a failed local message now correctly restores it to a delivered state after successful publication.
  * Resending an identical message shortly after a failed attempt avoids duplicate timeline entries.
  * Failed or withdrawn timeline messages remain excluded during replays, backfills, and ordinary updates unless explicitly recovered.
  * Recovery is limited to the original local message and its associated group, preventing unrelated messages from being revived.

* **Documentation**
  * Clarified retry behavior, delivery states, duplicate sends, and timeline invalidation handling.

### DIFF
--- a/crates/marmot-app/src/client/mod.rs
+++ b/crates/marmot-app/src/client/mod.rs
@@ -2957,8 +2957,7 @@ impl AppClient {
 
         let should_project_locally = !notifications::is_push_gossip_kind(event.kind);
         if should_project_locally {
-            let update = self
-                .record_local_app_event_projection(group_id, &sender, &event, None, None, false)?;
+            let update = self.record_send_intent_projection(group_id, &sender, &event)?;
             on_local_projection(update);
         }
 
--- a/crates/marmot-app/src/client/projection.rs
+++ b/crates/marmot-app/src/client/projection.rs
@@ -24,6 +24,42 @@ use crate::{
 use super::AppClient;
 
 impl AppClient {
+    /// Project a fresh send intent optimistically, before the publish is tried.
+    ///
+    /// Two things happen here, and the order matters. Inner app-event ids are
+    /// NIP-01 hashes over (pubkey, created_at, kind, tags, content) with
+    /// second-granular `created_at`, so a resend of identical text inside the
+    /// same second as a failed send mints that send's *exact* id — and lands on
+    /// the row `retract_failed_local_projection` tombstoned as
+    /// `local_publish_failed`. `record_app_event`'s upsert keeps invalidation
+    /// terminal by design (it cannot tell a retry from a relay redelivery, a
+    /// backfill replay, or a re-synthesized system row), so the retraction has
+    /// to be cleared explicitly first, and only where there is evidence to clear
+    /// it. A fresh send intent is that evidence: replay seams re-record rows
+    /// without ever reaching this function.
+    ///
+    /// Only the pre-publish projection clears. The post-publish one runs on a
+    /// row this call already revived, and anything that invalidated it *during*
+    /// the send — the terminal-group sweep, a convergence withdrawal — is a
+    /// verdict the send flow has no evidence to overturn.
+    ///
+    /// The clear's own projection update is dropped: the record below reprojects
+    /// the same row and returns that update, so forwarding both would emit the
+    /// same flip twice.
+    pub(crate) fn record_send_intent_projection(
+        &self,
+        group_id: &GroupId,
+        sender: &str,
+        event: &MarmotInnerEvent,
+    ) -> Result<crate::AppProjectionUpdate, AppError> {
+        let _revived = self.app.clear_timeline_local_publish_failure(
+            &self.state.label,
+            &hex::encode(group_id.as_slice()),
+            &event.id,
+        )?;
+        self.record_local_app_event_projection(group_id, sender, event, None, None, false)
+    }
+
     /// `advance_read_marker`: pre-publish projections must pass `false` so a
     /// failed publish never leaves the group read marker advanced past inbound
     /// unreads or pointing at an invalidated own message (#338); only the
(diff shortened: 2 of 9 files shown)


Datawav (commented):
## Automated review

One actionable finding.

1. Minor — Send-intent revival and reprojection are not atomic
- Exact changed-line evidence: `crates/marmot-app/src/client/projection.rs:55-59` adds `record_send_intent_projection`, which first calls `clear_timeline_local_publish_failure(group_id, &event.id.to_hex())?` and then separately calls the fallible `record_local_app_event_projection(...)`. `crates/marmot-app/src/client/mod.rs:2960` now propagates that helper’s error before publishing.
- Triggering state: A same-ID resend targets a sent row tombstoned with `local_publish_failed`; clearing that tombstone commits, then the subsequent projection write encounters a storage error.
- Consequence: The send exits before attempting publication, but the existing row remains live with no source message ID. It can therefore appear pending despite no publish attempt and remain in that state without another same-ID retry.
- Confidence: High. The patch and bounded source context establish two independently fallible storage operations, while the new tests cover successful revival and scoping but not an injected error between clear and reprojection.
- Smallest sound correction: Perform the narrowly scoped tombstone clear and event reprojection in one SQLite transaction so both commit or neither does. Add a fault-injection regression test proving an error during reprojection preserves the `local_publish_failed` tombstone.

Audit evidence:
- Exact head: `c386617ec2e1e46080a41a4667dd0f861d329845`
- Changed files inspected include `crates/marmot-app/src/client/projection.rs`, `crates/marmot-app/src/client/mod.rs`, `crates/storage-sqlite/src/timeline.rs`, `crates/storage-sqlite/src/timeline/tests.rs`, and `crates/marmot-uniffi/src/commands/message.rs`.
- The patch preserves existing invalidation columns during upsert and restricts explicit revival by group, direction, invalidation state, and exact `local_publish_failed` reason. Tests exercise tombstone preservation, group scoping, invalidation variants, live-row controls, retention behavior, chat-list projection, and same-second resend behavior.
- No physical-device evidence exists for this head.

Risk lenses checked:
- `persistence/migration`: Data preservation and idempotent replay behavior are covered by the narrowed update and tombstone tests; no schema migration is introduced. The partial-failure lens identifies the non-atomic clear/reprojection sequence above.
- `tests`: The added tests target production paths and include boundary and negative cases, but omit the newly introduced intermediate storage-error boundary.

Discussion classification:
- Inline data-integrity claim at `crates/marmot-app/src/client/projection.rs:59`: supported and retained as the finding above.
- Inline Rustdoc paragraph-break claim at `crates/marmot-uniffi/src/commands/message.rs:51`: accurate but stylistic/documentation-only, with no behavioral consequence; discarded.
- Summary issue comment: informational overview, no additional independently actionable claim.
- Review summary: duplicates the two inline comments; deduplicated.

Check evidence:
- Supplied CI is complete. Rust tests `(1/4)` through `(4/4)`, Rust check, clippy, format, audit, Tamarin proofs, simulator/vectors, relay runtime tests, relay CLI E2E, CLI mock-relay E2E, and C-bindings smoke all completed successfully.
- The `Publish wn-agent release` job was skipped; all supplied non-skipped runs completed successfully.

Datawav (approved):

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T08:55:02Z` |
| Commit reviewed | `08d870a8df51588840f591fdb82c72d98de8e73f` |
| Model | `Cursor Grok 4.6` |

The two-layer contract is sound: `record_app_event` can no longer un-tombstone on replay, and the only revival is an explicit send-path clear scoped to `direction = sent` + `local_publish_failed`. Record-then-clear is the right failure-prefix order — a failed record, a failed clear, and a crash between commits all leave the row `Failed` and still retryable. Terminal-group swept sends stay dead because `ensure_group_application_messages_allowed` rejects the send before projection, and replay seams never call the clear.

Scoping tests (reason, direction, live row, group, kind/tags change, retention variant) and the same-second resend test pin the actual collision. Earlier clear-then-record atomicity comments are resolved by this commit.

Minor: the PR description still says the send path clears *before* the optimistic record. The code and comments do the opposite; worth updating so the next reader does not re-litigate the old order.

Datawav (approved):

PR #1615, opened 2026-09-01
Batch app-component reads onto one MLS group load

dannym-arx:
### Summary
Batch `app_component` lookups in `CgkaEngine` to avoid redundant `MlsGroup::load` calls during group projections.

### Context
`CgkaEngine::app_component` currently triggers a full `MlsGroup::load` (13 storage reads + ratchet tree deserialization) just to check a key on the already-parsed `app_data_dictionary`. 

Because `refresh_group` / `add_group` in `marmot-app` look up 7 components sequentially, every group refresh was paying for 7 separate group loads, plus an extra redundant group-record read in `encrypted_media_for_group`.

### What changed
- Added `CgkaEngine::app_components(group_id, &[ids])` to fetch multiple IDs in a single load (with fallback logic in the trait).
- Added corresponding wrappers in `cgka-session` and `marmot-account`.
- Updated `marmot-app` group projection to batch all 7 component reads via `read_group_components`. Shared parsing logic moved to `parse_*_component` helpers.
- Cleaned up duplicated liveness-gated `MlsGroup::load` boilerplate in `engine.rs` into a `live_mls_group` helper.

### Impact
Cuts group projection from **7 full `MlsGroup::load` calls (~91 storage reads + 7 tree deserializations) down to 1**.

## Summary by CodeRabbit

* **New Features**
  * Added batched retrieval of multiple application components.
  * Group projections now load required components more efficiently in a single request.
  * Added support for component-specific fallback handling while preserving required routing errors.

* **Bug Fixes**
  * Improved consistency between batched and individual component lookups.
  * Unknown groups now return the appropriate error during component retrieval.

### DIFF
--- a/crates/marmot-app/src/client/projection.rs
+++ b/crates/marmot-app/src/client/projection.rs
@@ -219,19 +219,24 @@ impl AppClient {
         }
         let previous = self.state_group_record(group_id);
         let group_metadata = self.runtime.group_record(group_id).ok();
-        let Ok(nostr_routing) = self.nostr_routing_for_group(group_id) else {
+        let protocol_profile = group_metadata
+            .as_ref()
+            .map(|group| group.protocol_profile)
+            .unwrap_or(ProtocolProfile::Legacy);
+        let components = self.read_group_components(group_id, protocol_profile);
+        let Ok(nostr_routing) = components.nostr_routing else {
             return;
         };
         let projection = EventGroupProjection {
             nostr_routing,
             group_metadata: group_metadata.as_ref(),
-            profile: self.profile_for_group(group_id),
+            profile: components.profile,
             admin_policy: self.admin_policy_for_group(group_id),
-            message_retention: self.message_retention_for_group(group_id),
-            agent_text_stream: self.agent_text_stream_for_group(group_id),
-            avatar_url: self.avatar_url_for_group(group_id),
-            encrypted_media: self.encrypted_media_for_group(group_id),
-            image: self.image_for_group(group_id),
+            message_retention: components.message_retention,
+            agent_text_stream: components.agent_text_stream,
+            avatar_url: components.avatar_url,
+            encrypted_media: components.encrypted_media,
+            image: components.image,
         };
         add_group(
             &mut self.state,
@@ -247,17 +252,22 @@ impl AppClient {
     pub(crate) fn add_group(&mut self, group_id: &GroupId) -> Result<(), AppError> {
         let previous = self.state_group_record(group_id);
         let group_metadata = self.runtime.group_record(group_id).ok();
-        let nostr_routing = self.nostr_routing_for_group(group_id)?;
+        let protocol_profile = group_metadata
+            .as_ref()
+            .map(|group| group.protocol_profile)
+            .unwrap_or(ProtocolProfile::Legacy);
+        let components = self.read_group_components(group_id, protocol_profile);
+        let nostr_routing = components.nostr_routing?;
         let projection = EventGroupProjection {
             nostr_routing,
             group_metadata: group_metadata.as_ref(),
-            profile: self.profile_for_group(group_id),
+            profile: components.profile,
             admin_policy: self.admin_policy_for_group(group_id),
-            message_retention: self.message_retention_for_group(group_id),
-            agent_text_stream: self.agent_text_stream_for_group(group_id),
-            avatar_url: self.avatar_url_for_group(group_id),
-            encrypted_media: self.encrypted_media_for_group(group_id),
-            image: self.image_for_group(group_id),
+            message_retention: components.message_retention,
+            agent_text_stream: components.agent_text_stream,
+            avatar_url: components.avatar_url,
+            encrypted_media: components.encrypted_media,
+            image: components.image,
         };
         add_group(
             &mut self.state,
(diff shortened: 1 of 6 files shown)


crates/marmot-app/src/client/projection.rs
```diff
@@ -541,12 +551,52 @@ impl AppClient {
...
+            self.runtime
+                .app_component(group_id, GROUP_PROFILE_COMPONENT_ID)
+                .ok()
+                .flatten(),
+        )
+    }
+
+    /// One batched engine read for every component the group projection needs:
+    /// 7 components on a single engine state load instead of 7 loads.
+    /// Per-component fallbacks match the individual `*_for_group` helpers.
```
erskingardner, line 564: Worth noting (non-blocking, fine as follow-up): this batches the two projection builders in this file, but the sibling projection sites in `sync.rs` still pay 7 sequential `MlsGroup::load`s per build — `event_group_projection_best_effort` (~sync.rs:1325) and the inline `EventGroupProjection` in the per-event observe path (~sync.rs:3949), plus the post-publish projection in `mod.rs` (~3951). The per-event observe path runs for every inbound event with a group id, so it is arguably the hottest of these call sites. Since `read_group_components` is now the shared helper, pointing those sites at it too would extend the win to the path that fires most often.

crates/marmot-app/src/client/projection.rs
```diff
@@ -541,12 +551,52 @@ impl AppClient {
...
+            Ok(components) => components,
+            Err(error) => {
+                return GroupComponentReads::unavailable(error.into(), protocol_profile);
+            }
+        };
+
+        // `next()` consumes in `component_ids` order; struct-literal fields
+        // evaluate in source order, which matches.
+        let mut components = components.into_iter();
+        let mut next = || components.next().flatten();
```
erskingardner, line 590: Non-blocking: this is correct today (struct field expressions evaluate in source order), but the pairing is positionally coupled across `component_ids`, this struct literal, and the struct definition — reordering any one silently mis-assigns components, and a mismatch still compiles because every parser takes `Option<Vec<u8>>` and just falls back on a failed parse. If you ever touch this again, binding by name before the literal removes the coupling without the `clone` cost of index-based access:

```rust
let mut results = components.into_iter();
let routing = results.next().flatten();
let profile = results.next().flatten();
// ...
GroupComponentReads {
    nostr_routing: parse_routing_component(routing),
    profile: parse_profile_component(profile),
    // ...
}
```

Happy to leave as-is given the explanatory comment; just flagging the failure mode.

crates/marmot-app/src/client/projection.rs
```diff
@@ -219,19 +219,24 @@ impl AppClient {
         }
         let previous = self.state_group_record(group_id);
         let group_metadata = self.runtime.group_record(group_id).ok();
-        let Ok(nostr_routing) = self.nostr_routing_for_group(group_id) else {
+        let protocol_profile = group_metadata
+            .as_ref()
+            .map(|group| group.protocol_profile)
+            .unwrap_or(ProtocolProfile::Legacy);
```
erskingardner, line 225: Non-blocking: this `group_metadata -> ProtocolProfile` derivation (with the `Legacy` default) now appears three times in this file (here, `add_group`, and `encrypted_media_for_group`). A tiny `fn protocol_profile_of(Option<&AppGroupRecord>) -> ProtocolProfile` helper would keep the default in one place. Fine to defer.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T09:03:58Z` |
| Commit reviewed | `2f56ea0c44e3f1a354d61dac1c391df0efe5c28a` |

Verified the premise: `Engine::app_component` performs a full `MlsGroup::load` (13 storage reads + ratchet-tree deserialize) per id, and group projections read 7 components sequentially, so this is a real hot-path win rather than speculative batching.

What I checked:

- **Fallback parity.** Every per-component fallback in `GroupComponentReads` / the `parse_*_component` helpers matches its `*_for_group` counterpart, including the profile-dependent encrypted-media fallback. The individual helpers now delegate to the same parsers, so the two paths can't drift.
- **Error semantics.** Only routing fails loud, as before. On a batch/engine error, `unavailable(...)` surfaces the error through `nostr_routing` while the rest fall back — exactly the observable behavior of the old sequential code, where `refresh_group`/`add_group` bailed on the routing error and the other readers swallowed errors via `.ok().flatten()`.
- **Trait default.** The `CgkaEngine::app_components` default is a per-id loop that short-circuits on the first error; read-only, so that's consistent with per-id semantics, and the engine override answers all ids from one `live_mls_group` load.
- **Dedup.** `live_mls_group` removes three copies of the liveness-gated load boilerplate — a genuine net simplification.
- **Tests.** The new engine conformance test covers present, absent, and unknown-group parity between batched and per-id reads; the full marmot-app projection suites exercise the batched path end-to-end.

Two non-blocking notes inline. CI (format, check, clippy, most test shards) is green at this SHA; remaining shards pending.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T09:04:02Z` |
| Commit reviewed | `2f56ea0c44e3f1a354d61dac1c391df0efe5c28a` |

Correct, well-scoped optimization. Verified the behavior-preservation claims against the head SHA:

- **Fallback semantics are equivalent.** The old per-component helpers each did `.ok().flatten()`, and the new batch collapses any engine error into `GroupComponentReads::unavailable` fallbacks. Since `app_component_data_of_group` returns `Option` (no per-component errors) and every engine failure is group-level (`UnknownGroup` / load failure), an error previously failed all seven reads identically — so the collapse changes nothing.
- **Error propagation preserved.** `add_group` still propagates routing engine errors via `components.nostr_routing?`, and `refresh_group` still fails closed on routing errors.
- **`live_mls_group` dedup is faithful** — same `ensure_group_live` gate, same load, same error mapping at all three former call sites, and the new `app_components` accessor is correctly quarantine-gated through it.
- **Trait default impl** keeps any other `CgkaEngine` implementors source-compatible, and the new engine test covers batched-vs-individual parity plus `UnknownGroup`.

Non-blocking notes:

1. **The hottest projection path is still unbatched** (inline comment on `read_group_components`): the per-event observe path in `sync.rs` (~3949) and `event_group_projection_best_effort` (~1325) still do 7 sequential loads per inbound event with a group id. Good follow-up candidate now that the helper exists.
2. **`crates/cgka-engine/AGENTS.md` upkeep**: the quarantine-gate paragraph (~line 431) enumerates accessors that call `ensure_group_live` — `app_components` should be added to that list per the repo's AGENTS.md freshness rule.
3. Seconding CodeRabbit's nitpick on the positional `next()` mapping in `read_group_components` — correct today (struct-literal fields evaluate in source order), but zipping ids to results would make the coupling explicit.

PR #1616, opened 2026-09-01
Fix stale invite acceptance after terminal self-removal

### LINKED ISSUE #1606: Reject group invite acceptance after terminal self-removal
## Summary

A pending group invite can still be accepted locally after the account’s authoritative group projection has transitioned to terminal `Removed` or `Left` membership. `accept_group_invite` currently clears `pending_confirmation` and unarchives the projection without validating terminal self-membership, so a stale Android invite action can report success and temporarily restore an active-looking conversation before later reconciliation returns it to the removed state.

The reported Android sequence is: receive an invite notification, get removed before opening it, then tap the stale notification and briefly see **Join group**. This report is source-confirmed at the MDK command boundary; no independent runtime or device reproduction was performed during this audit.

## Current behavior

Grounded against `master` at `053c7c556d1b33f92471afeae224cfdcac8eb128`.

- Incoming departure state is authoritative in the app projection: `MemberLeft` maps to `SelfMembership::Left` and `MemberRemoved` maps to `SelfMembership::Removed` ([`client/sync.rs`](https://github.com/marmot-protocol/mdk/blob/053c7c556d1b33f92471afeae224cfdcac8eb128/crates/marmot-app/src/client/sync.rs#L4086-L4099)); when the departing member is the local account, that state is persisted to the group projection ([`client/sync.rs`](https://github.com/marmot-protocol/mdk/blob/053c7c556d1b33f92471afeae224cfdcac8eb128/crates/marmot-app/src/client/sync.rs#L3781-L3806)). A later genuine `GroupJoined`/`GroupCreated` event is what restores `Member` ([`client/sync.rs`](https://github.com/marmot-protocol/mdk/blob/053c7c556d1b33f92471afeae224cfdcac8eb128/crates/marmot-app/src/client/sync.rs#L3825-L3839)).
- `AppClient::accept_group_invite` delegates directly to the local confirmation setter ([`client/mod.rs`](https://github.com/marmot-protocol/mdk/blob/053c7c556d1b33f92471afeae224cfdcac8eb128/crates/marmot-app/src/client/mod.rs#L2595-L2598)).
- That setter only looks up the group, then assigns `pending_confirmation = false` and `archived = false`; it does not require an open pending invite or reject `SelfMembership::Removed`/`Left` ([`client/mod.rs`](https://github.com/marmot-protocol/mdk/blob/053c7c556d1b33f92471afeae224cfdcac8eb128/crates/marmot-app/src/client/mod.rs#L3993-L4011)).
- The account worker publishes a successful group-state update whenever that local mutation returns `Ok` ([`runtime/account_worker.rs`](https://github.com/marmot-protocol/mdk/blob/053c7c556d1b33f92471afeae224cfdcac8eb128/crates/marmot-app/src/runtime/account_worker.rs#L2179-L2194)).

This is distinct from [silent-eviction issue #376](https://github.com/marmot-protocol/mdk/issues/376) and its merged fix [PR #703](https://github.com/marmot-protocol/mdk/pull/703): those establish how terminal removal becomes authoritative. It also preserves the genuine re-invite behavior established by [issue #602](https://github.com/marmot-protocol/mdk/issues/602), [issue #469](https://github.com/marmot-protocol/mdk/issues/469), and merged [PR #1367](https://github.com/marmot-protocol/mdk/pull/1367): a distinct new Welcome/join may restore membership and create a new pending confirmation. The remaining gap is narrower—an old pending confirmation can still be accepted after terminal membership is already authoritative and before any such new join exists.

## Expected behavior

- Invite acceptance must fail closed when the authoritative local group record is `Removed` or `Left`, or when no open pending confirmation exists for the current welcome.
- A failed acceptance must not mutate or persist `pending_confirmation`, `archived`, self-membership, or the chat-list/group-state projection, and must not emit a successful group-state update.
- A concurrent removal that is linearized before acceptance must win. Acceptance must not acknowledge success after that terminal transition.
- A genuine later re-invite remains valid only after its new welcome/join transition has authoritatively restored membership and established a new pending confirmation; replay of an old welcome must not reopen acceptance.
- The rule must survive process restart, projection replay, and offline catch-up.

## Acceptance criteria

- The public invite-accept operation validates the current authoritative invitation and self-membership state before mutation and returns a stable app-facing error for terminal or non-pending invitations.
- Validation and mutation are one account-worker-serialized operation; no check-then-mutate race can acknowledge an acceptance after a preceding removal event.
- The terminal failure remains distinguishable through the existing app-facing and UniFFI boundaries so Android can render a deterministic no-longer-invited/not-a-member state rather than a generic success followed by rollback.
- Existing valid-invite acceptance and genuine re-invite behavior remain unchanged.
- Privacy-safe diagnostics, if any, do not include account IDs, group IDs, welcome IDs, relay URLs, or message content.

## Regression coverage

- Add a `marmot-app` test with `pending_confirmation = true` plus `SelfMembership::Removed`, and the corresponding `Left` case: acceptance returns an error and the persisted record is byte-for-byte unchanged after reopen.
- Add an account-worker ordering test where authoritative self-removal is applied before a queued accept command: removal wins, the accept response is an error, and no success update is published.
- Add a replay/restart test proving an old welcome cannot make the terminal invitation actionable again, while a distinct genuine re-invite can establish a new pending confirmation and be accepted.
- Add a UniFFI boundary test for the terminal error mapping used by Android.

## Related Android scope

The Android symptom was previously addressed at the projection/UI layer by [whitenoise-android issue #1248](https://github.com/marmot-protocol/whitenoise-android/issues/1248) and merged [PR #1273](https://github.com/marmot-protocol/whitenoise-android/pull/1273). That issue should be reopened for the observed regression and client-side stale-action handling after this upstream tracker has a canonical URL. The MDK command invariant in this issue is independently required even when every client hides stale invite controls correctly.

erskingardner:
## Summary

- Validate invite acceptance against the durable group projection before mutating confirmation state.
- Reject terminal or already-resolved invitations with a typed `GroupInviteNotActionable` error across app, UniFFI, and C boundaries.
- Keep failed acceptance mutation-free, preserve genuine re-invite success, and suppress false worker success updates.

Fixes #1606

## Test plan

- [x] New terminal-membership regression failed before the fix and passes after it.
- [x] New genuine re-invite return-state regression failed before the fix and passes after it.
- [x] `cargo test -p marmot-c --features alloc-audit`
- [x] `cargo test -p cgka-conformance-simulator app_runtime::tests::app_errors_are_classified_without_serializing_inner_details --lib -- --exact`
- [x] Workspace checks and clippy, including OTLP feature builds
- [x] Generated C header reproduced byte-for-byte
- [x] Convergence-policy pin

## Local baseline notes

- The full `marmot-app` run passed 868 tests including all invite regressions; six unrelated epoch-backfill timing tests failed under parallel load and each passed in isolation.
- `just fast-ci` reaches the existing BSD `sed` failure in `test_campaign_toolchain.sh`; every remaining fast-CI target was run explicitly and passed.
- `just c-smoke` builds the release library and dynamic smoke binary, then the pre-existing static macOS link omits Security/CoreFoundation frameworks and fails on Apple keychain symbols.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented acceptance of stale, resolved, or otherwise non-actionable group invitations.
  * Preserved authoritative membership information when accepting valid re-invitations.
  * Rolled back confirmation and archive changes if saving fails.
  * Rejected invitations no longer trigger misleading successful group-state updates.

* **Error Handling**
  * Added clear status and typed error reporting for non-actionable group invitations across supported interfaces.
  * Protected underlying error details from being exposed in user-facing telemetry.


erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T08:49:34Z` |
| Commit reviewed | `55b245b77d200034f11978c82b792635339dccca` |
| Model | `Cursor Grok 4.6` |

GitHub rejected a self-approve on this PR; the review is posted as a comment with the same recommendation.

The command-boundary fix is the right one and should land. `accept_group_invite` now fail-closes on the storage-owned projection (`Member` + open `pending_confirmation`) before touching confirmation state, the save path compensates in-memory fields if persist fails, and the worker already withholds `GroupStateUpdated` on `Err`. The new typed error is mapped through UniFFI and the C ABI with a stable appended status code.

The terminal / already-resolved / worker-ordering tests match the actual bug: durable `Removed`/`Left` (or a resolved invite) must not clear `pending_confirmation` or publish success.

Non-blocking follow-ups:
- Prefer `overlay_storage_self_membership` after a successful accept instead of copying the pre-mutation snapshot and hand-patching `state.groups`. Same outcome, less drift.
- `pending_group_invites` still enumerates `pending_confirmation = 1` without a membership filter, so `wn-agent` will keep retrying `GroupInviteNotActionable` on a terminal-but-still-pending row. Bounded, but it is a new permanent-retry loop for that candidate.
- CLI JSON still falls through to generic `command_failed`; fine if Android/UniFFI is the consumer that needs the typed variant.

crates/marmot-app/src/tests.rs
```diff
@@ -12119,6 +12119,235 @@ async fn a_drained_self_departure_and_rejoin_move_stored_self_membership() {
...
+        client.mark_group_projection_dirty(&group_id);
+        client
+            .save_state_with_pending_local_group_deletion_frontier_clears()
+            .unwrap();
+        let durable_before = app
+            .group("alice", &group_id_hex)
+            .unwrap()
+            .expect("durable re-invite projection");
+        assert!(durable_before.pending_confirmation);
+        assert_eq!(durable_before.self_membership, SelfMembership::Member);
```
erskingardner, line 12264: This test does not exercise a genuine re-invite.

Durable membership stays `Member` because projection save does not own `self_membership` — the in-memory `Removed` write is discarded, then accept overlays that still-original `Member` row. That is a useful inverse-staleness check (worker memory says `Removed`, storage says `Member` + pending), but it is not remove → new Welcome/`GroupJoined` → new pending confirmation, and it does not prove an old `via_welcome_message_id_hex` stays non-actionable across restart.

#1606 asked for that restart / old-welcome vs distinct new-welcome case, and the device reproduction on the issue is exactly that path. Please rename this test to match what it actually pins, and add the welcome-identity/restart coverage (or say explicitly that it is follow-up).

crates/marmot-app/src/tests.rs
```diff
@@ -12119,6 +12119,235 @@ async fn a_drained_self_departure_and_rejoin_move_stored_self_membership() {
...
+
+        let accepted = client.accept_group_invite(&group_id).unwrap();
```
erskingardner, line 12266: Consider also asserting the durable record after the successful acceptance (`app.group("alice", &group_id_hex)` → `pending_confirmation == false`, `self_membership == Member`). The test checks the returned record and the in-memory worker projection, but not that acceptance persisted to storage without disturbing the storage-owned membership column — the durable-side half of the genuine re-invite invariant. The byte-for-byte checks in the refusal tests only cover the non-mutation side.

crates/marmot-app/src/client/mod.rs
```diff
@@ -2598,7 +2598,27 @@ impl AppClient {
...
+            .groups
+            .iter_mut()
+            .find(|group| group.group_id_hex == accepted.group_id_hex)
+        {
+            group.self_membership = authoritative.self_membership;
+        }
+        Ok(accepted)
     }
 
     pub async fn decline_group_invite(
```
erskingardner, line 2624: Scope question, non-blocking: `decline_group_invite` does not get the same pre-validation. For a stale terminal invitation it relies on `leave_group_with_audit_context` failing once the engine rejects the leave — fail-closed today, but the caller gets an engine error instead of the typed `GroupInviteNotActionable`, and `set_group_invite_confirmation(group_id, false, true)` would still mutate a stale invitation if a leave ever succeeded. The same stale-notification UX from #1608 hits Decline as well as Join. Worth a follow-up issue (or a note on #1606 / whitenoise-android#1248) even though acceptance is the right scope for this PR.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T08:51:11Z` |
| Commit reviewed | `55b245b77d200034f11978c82b792635339dccca` |

## Summary

Fixes #1606 by validating invite acceptance against the durable group projection before any mutation, returning a typed `GroupInviteNotActionable` across the app, UniFFI, and C boundaries.

I verified the central assumption against the storage write path at this SHA:

- `account_groups.self_membership` is written only by `set_group_self_membership` (storage-sqlite/src/account_projection.rs:1758); the group delta upsert used by the accept mutation path does not include that column, so the read-in-`accept_group_invite` / mutate-via-delta-save split is coherent and acceptance can never clobber a storage-owned terminal membership.
- The accept command runs inside the account-worker serialization domain and `GroupStateUpdated` is published only `if result.is_ok()`, so a refused acceptance publishes no success update (covered by the worker ordering test).
- The save-failure rollback in `set_group_invite_confirmation` restores the in-memory `pending_confirmation` / `archived` values, matching the multi-step torn-write invariant.
- The new error keeps group ids out of `Display` and the simulator classification (`ExpectedRefusal`, marker-free message), consistent with the privacy-safe diagnostics invariant.

The change is surgical (8 files, mostly error plumbing plus tests), the premise is real (#1606 is source-confirmed with a device reproduction in the issue thread), and the fail-closed behavior is what the issue asked for. Non-blocking notes are inline.

Two coverage notes, neither blocking:

- #1606's regression-coverage list also asked for a welcome-replay test proving an old welcome cannot reopen acceptance. These tests simulate the projection seam directly, which is a fair unit-level equivalent (pending confirmation is only established by a genuine `GroupJoined` in sync), but the replay path itself is not exercised here.
- CI is still queued at review time. The author's local baseline notes cover the same gates; please let the matrix finish before merging.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T08:51:17Z` |
| Commit reviewed | `55b245b77d200034f11978c82b792635339dccca` |
| Recommended action | **Merge** (GitHub blocked self-approval, so this posts as a comment) |

## Summary

Clean fix for the stale-invite-acceptance bug (#1606). The public `accept_group_invite` validates the durable projection's `self_membership` and `pending_confirmation` before mutating, rejecting terminal (`Removed`/`Left`) and already-resolved invites with a typed `GroupInviteNotActionable` error mapped across the app, UniFFI, and C boundaries.

Findings verified against the head SHA:

- The storage-owned membership cannot be clobbered by the mutation path: `stored_group_from_app_group` ignores `self_membership` (owned by `set_group_self_membership`), so the fixup copying the authoritative membership into the returned DTO and the worker's in-memory record is safe and necessary for the genuine re-invite case.
- The new save-error rollback in `set_group_invite_confirmation` mirrors the pre-existing `set_group_archived` compensation pattern in the same file — consistent with the repo's torn-write discipline.
- Worker success-event publication is suppressed on the error path (the account worker publishes `GroupStateUpdated` only on `Ok`), and validation + mutation run synchronously inside the one serialized worker operation, satisfying the issue's check-then-mutate criterion.
- Error-privacy posture holds: the conformance binding asserts the classified message excludes the marker, and embedding `group_id_hex` in the FFI error payload matches the existing `UnknownGroup` pattern.
- C status enum appends `61` without shifting existing values; header regeneration evidence is listed in the test plan.
- Regression coverage is thorough: terminal `Removed`/`Left` byte-for-byte durability check, resolved-invite refusal, genuine re-invite acceptance, and worker ordering with no-success-update assertion.

Non-blocking note: CI checks were still queued at review time; the PR body's local test evidence covers the touched crates. No concerns — recommend merge.

erskingardner:
Closing as a duplicate of #1614. The canonical fix is now on #1614, and the useful additional coverage from this PR—byte-for-byte terminal projection preservation, no false GroupStateUpdated event, and privacy-safe simulator classification—was consolidated there in 09895c582fe09db167ca8c79f446dc2168822f9b.

PR #1617, opened 2026-09-01
Fold pass-admission scans into one member walk

dannym-arx:
## Problem

`admit_stored_message_to_convergence_pass` walked `pass.members` twice per admitted input: a linear duplicate-id scan, then a full `ConvergenceInputContext::from_pass_members` rebuild after pushing the new member O(members²) comparisons and BTreeMap rebuilds over a pass's lifetime.

## Change

The cheap checks (duplicate id, cutoff freeze, candidate fetch, horizon) stay first, then the one admitted input's relevance is answered by `scan_selection_relevance`: a single pre-push member scan with no commit-edge map materialization, short-circuiting on the first competing pair (review feedback — the earlier version built the context during the duplicate scan, so bail-out paths paid for it). Pre-push is equivalent for the new member: a `CommitEdge` is relevant unconditionally, and a `Proposal`/`AppWitness` member contributes no commit edge. `is_potentially_selection_relevant` stays as the test-only map-based oracle; `scan_relevance_matches_context_relevance` pins the two verdicts together.

Not changed: the per-admission `put_convergence_pass` rewrite — the retained row and pass membership are one durability boundary; splitting the pass blob into per-member rows would be a storage migration with no measured need.

## Count

Per admitted input: 2 full member walks + 1 BTreeMap rebuild → 1 walk with incremental folds.

## Performance
- Improved convergence admission efficiency by evaluating selection relevance directly from available pass information.
- Reduced redundant context reconstruction during convergence processing.
- Preserved quiescence reset behavior while streamlining convergence decisions.

## Refactor
- Simplified member-processing logic without changing convergence outcomes.
- Added coverage for representative selection-relevance scenarios, including competing commits and proposal dependencies.

### DIFF
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -524,12 +524,16 @@ impl<S: StorageProvider> Engine<S> {
         if pass.phase != ConvergencePassPhase::Collecting {
             return Ok((ConvergenceAdmissionOutcome::Retained, opened));
         }
-        if pass
-            .members
-            .iter()
-            .any(|member| &member.message_id == message_id)
-        {
-            return Ok((ConvergenceAdmissionOutcome::Retained, opened));
+        // One walk answers both admission questions: is this id already a
+        // member, and what commit-edge context do the current members form.
+        // Previously the id check and `from_pass_members` each rescanned the
+        // whole member list per admitted input (O(members²) per pass).
+        let mut context = ConvergenceInputContext::default();
+        for member in &pass.members {
+            if &member.message_id == message_id {
+                return Ok((ConvergenceAdmissionOutcome::Retained, opened));
+            }
+            context.fold_member(member);
         }
         if now.monotonic_ms >= pass.cutoff_monotonic_ms() {
             self.freeze_collecting_convergence_pass(&mut pass, now)?;
@@ -566,7 +570,11 @@ impl<S: StorageProvider> Engine<S> {
         }
         let admitted_input = candidate.input;
         pass.members.push(candidate.member);
-        let context = ConvergenceInputContext::from_pass_members(&pass.members);
+        // The pre-push context is sufficient here: a CommitEdge is relevant
+        // unconditionally, and a Proposal/AppWitness member contributes no
+        // commit edge, so folding the new member in cannot change its own
+        // relevance verdict.
+        //
         // Restart quiescence only for an input role that can change this
         // batch's deterministic resolution. Ordinary app delivery never
         // reaches this path; a future app without competing candidate edges is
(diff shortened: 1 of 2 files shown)


crates/cgka-engine/src/distributed_convergence.rs
```diff
@@ -524,12 +524,16 @@ impl<S: StorageProvider> Engine<S> {
...
+        // One walk answers both admission questions: is this id already a
+        // member, and what commit-edge context do the current members form.
+        // Previously the id check and `from_pass_members` each rescanned the
+        // whole member list per admitted input (O(members²) per pass).
+        let mut context = ConvergenceInputContext::default();
+        for member in &pass.members {
+            if &member.message_id == message_id {
+                return Ok((ConvergenceAdmissionOutcome::Retained, opened));
+            }
+            context.fold_member(member);
```
erskingardner: **[P1] Preserve cheap early exits before rebuilding the context**

`fold_member` allocates and rebuilds `commit_edges_by_source_epoch` while the duplicate scan runs. Previously, cutoff and out-of-horizon inputs returned before `from_pass_members` constructed this map; now each such input rebuilds it from every current member even though the result is discarded. `DurableConvergencePass::members` has no count bound, so on a large collecting pass this turns the performance-only change into extra BTreeMap/BTreeSet allocation and insertion work on explicit retained-input paths. Keep the cheap membership/cutoff/horizon checks before constructing the context, or compute the candidate-specific relevance in one scan without materializing the full map.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T09:55:23Z` |
| Commit reviewed | `92b9496c494083c9cefd17610f2e19e72f37b282` |
| Model | `gpt-5.6-sol` |
| Reasoning level | `high` |

The pre-push relevance verdict is behaviorally equivalent, and exact-head CI is green. However, the optimization moves the expensive context rebuild ahead of early exits. Successful admissions still rebuild the map from all prior members, so lifetime map work remains quadratic; the change mainly removes the cheap ID-only traversal while making rejected inputs more expensive.

Datawav (approved):

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T20:49:53Z` |
| Commit reviewed | `9b995ba5271d871af375b8c0cee32da440edcb21` |
| Model | `gpt-5.6-sol` |
| Reasoning level | `high` |

The previous blocking issue is fixed: cutoff, no-candidate, and out-of-horizon exits now precede relevance work, while proposal/application relevance avoids materializing the full commit-edge map. Exact-head CI is green.

Non-blocking: the title and count still say this becomes one member walk, but proposal and application admissions retain the duplicate-id scan plus a separate relevance scan. The implementation is sound; the performance description should be tightened to match it.

Datawav (approved):

PR #1618, opened 2026-09-01
Share the convergence seen-id snapshot via Arc

dannym-arx:
### Summary
Avoid full deep-clones of the dedup snapshot during message convergence by sharing it via `Arc`.

### Context
`seen_message_ids_hex_for_convergence` was deep-cloning its dedup snapshot on every call—even on cache hits. At `DEDUP_CACHE_CAPACITY` (100k 64-char strings), this meant allocating ~100k strings (~10 MB) per pass, running up to 16 times per drain. 

To make matters worse, `canonicalize_internal` was cloning the entire set a second time just to check the pending batch against it.

### What changed
- Wrapped the cached `BTreeSet<String>` in an `Arc` so lookups and rebuilds only bump the refcount.
- Kept `CanonicalizationState.seen_message_ids` as a plain `BTreeSet<String>` (public model type); the engine hands the shared `Arc` in crate-privately via `StoredCanonicalizationOptions` and `canonicalize_with_shared_seen` (review feedback).
- Rewrote the batch dedup logic to check `.contains()` directly against the shared base set, only storing new IDs observed within the current batch (preserving the readmitted-witness exemption).

### Impact
Eliminated 2 full set copies per convergence pass (saving ~20 MB and ~200k heap allocations at full capacity).

## Summary by CodeRabbit

* **Improvements**
  * Improved message deduplication across processing batches.
  * Enhanced convergence consistency when handling stored messages and membership changes.
  * Reduced redundant processing during repeated state synchronization.
  * Preserved handling of previously delivered application messages while improving continuity across unchanged generations.


Datawav (commented):
## Automated review

1. Medium severity — public API compatibility regression
   Confidence: High

   Evidence: crates/cgka-engine/src/canonicalization.rs:133–138 changes the public field CanonicalizationState.seen_message_ids from BTreeSet<String> to Arc<BTreeSet<String>>. The required mechanical changes across crates/cgka-conformance-simulator/tests/canonicalization_contract.rs:45, independent_reference_model.rs:152, policy_sweeps.rs:14, openmls_replay_probe.rs, proptest_invariants.rs:428/727, and the engine integration tests demonstrate that existing struct-literal callers no longer compile unchanged.

   Trigger: Any downstream consumer constructs CanonicalizationState with a BTreeSet<String>, as the repository’s own callers did before this head.

   Consequence: Source compatibility breaks even though sharing is needed only for the engine’s internal convergence path. Consumers must add an Arc dependency detail to otherwise value-oriented canonicalization input.

   Smallest sound correction: Keep the public CanonicalizationState field source-compatible and add a crate-private convergence entry point or internal input representation that accepts Arc<BTreeSet<String>>. Route Engine::seen_message_ids_hex_for_convergence through that shared representation while preserving the existing public value-based API.

2. Low severity — the performance fix has no regression assertion
   Confidence: High

   Evidence: crates/cgka-engine/src/distributed_convergence.rs:288–311 changes seen_message_ids_hex_for_convergence to return cloned Arc handles, crates/cgka-engine/src/engine.rs:318–323 changes the cache storage to Arc, and crates/cgka-engine/src/canonicalization.rs:355 onward replaces the full-set working clone with a current-batch set plus base-set lookups. All changed test patches only adapt BTreeSet values to Arc; none asserts cache snapshot sharing or generation invalidation.

   Production behavior and failure mode: An unchanged dedup generation should return the same shared snapshot without copying up to 100,000 strings, while a generation change must produce a fresh snapshot with updated contents. A later accidental deep clone—or stale Arc reuse after mutation—would restore the material allocation cost or use stale dedup state without any focused test failing.

   Smallest regression test: Add a unit test beside seen_message_ids_hex_for_convergence that calls it twice without changing the generation and asserts Arc::ptr_eq, then mutates the bounded ID set, calls it again, and asserts both !Arc::ptr_eq and the updated encoded contents. The pointer-sharing assertion would not have been satisfiable by the pre-change value-returning implementation and directly protects the claimed production behavior rather than generic coverage.

Audit evidence:
- Head SHA: 155996d2585a208028418106593669401e529451
- Changed files inspected: crates/cgka-engine/src/canonicalization.rs, crates/cgka-engine/src/distributed_convergence.rs, crates/cgka-engine/src/engine.rs, crates/cgka-engine/tests/crash_recovery_sqlite.rs, crates/cgka-engine/tests/fork_detection.rs, crates/cgka-conformance-simulator/tests/canonicalization_contract.rs, crates/cgka-conformance-simulator/tests/independent_reference_model.rs, crates/cgka-conformance-simulator/tests/openmls_replay_probe.rs, crates/cgka-conformance-simulator/tests/policy_sweeps.rs, and crates/cgka-conformance-simulator/tests/proptest_invariants.rs.
- Bounded head_context was traced for the changed cache, canonicalization, and test symbols; several large snapshots were truncated, so no safety conclusion relies on absent context.
- The sole discussion item only says automated review was skipped; it supplies no actionable claim about this head.
- Risk lenses checked: tests — assertions fail before the fix; boundary and negative cases, including unchanged versus changed cache generation and base-set versus current-batch dedup; and whether tests exercise production behavior. Initial-review design fit and source compatibility were also assessed.
- Check evidence: available but incomplete. Rust tests (1/4 through 4/4), Rust check, Rust clippy, Rust format, and the listed completed jobs report success; “Relay runtime tests” and “Simulator and vectors” remain in progress, and “Publish wn-agent release” was skipped. Therefore all_completed=false and all_successful=false, and the supplied evidence does not establish that the full check suite passed.
- No physical-device evidence exists for this head; automated_device_gate is null, so no device report or recorded device-gate reason was supplied.

erskingardner (dismissed):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T10:01:10Z` |
| Commit reviewed | `155996d2585a208028418106593669401e529451` |
| Model | `gpt-5.6-sol` |
| Reasoning level | `high` |

The `Arc`-backed cache removes both full-set clones while preserving canonicalization behavior, including duplicate handling and the readmitted application-witness exemption. The change is focused, and exact-head CI is green.

erskingardner (approved):

Datawav (approved):

PR #1620, opened 2026-09-01
Probe contested-ness without seeding the full graph

dannym-arx:
### Summary
Classify deferral lineage with a commit-only scan instead of seeding the full stored graph.

### Context
`deferral_lineage` answers a single bool ("is the stored commit graph contested?") for every transport-deferred inbound message. But `stored_graph_is_contested` reused the full graph-seeding scan to get it: for every retained row it decoded the stored payload, cloned the full payload bytes, ran a TLS parse + SHA-256 projection, and materialized proposal/app vectors and own-commit stamps that the bool never looks at.

### What changed
- `stored_graph_is_contested` now scans commits only: no payload clone, no proposal/app/own-commit materialization, and it returns as soon as the first fork shows up.
- The record-contribution rules mirror the seed's commit arm, and the fork predicate lives in one place (`SourceEpochForkDetector`, also used by `commits_share_a_source_epoch`) so the two answers can't drift.
- The deferral-lineage test pair in `tests/distributed_convergence.rs` pins behavior parity.

### Impact
Per deferral over n retained rows: n full payload clones + full graph materialization → zero clones, commit projections only, early exit at the first fork.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of conflicting commit histories, including cases where different commits originate from the same source epoch.
  - Increased consistency in identifying contested stored graph states.

- **Performance**
  - Streamlined stored-history checks to reduce unnecessary graph processing.

### DIFF
--- a/crates/cgka-engine/src/openmls_projection.rs
+++ b/crates/cgka-engine/src/openmls_projection.rs
@@ -1783,35 +1783,78 @@ pub(crate) fn candidate_branch_peel<S: StorageProvider>(
 /// Whether this group's stored commit graph is contested: two retained commits
 /// at or above the retained anchor fork from the same source epoch.
 ///
-/// The replay-free half of [`candidate_branch_peel`], and deliberately the
-/// same predicate over the same seeded inputs — a second, privately restated
-/// notion of "forked" would drift from the one branch enumeration acts on
-/// without any test noticing. Costs the seeding scan and nothing more: no
-/// snapshot, no rewind, no OpenMLS replay.
+/// The replay-free half of [`candidate_branch_peel`]. Deliberately the same
+/// record-contribution rules as the commit arm of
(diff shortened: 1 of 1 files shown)


Datawav (approved):

erskingardner (dismissed):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T20:49:55Z` |
| Commit reviewed | `5e32795d45836f8cbc16d216f532c4579db13278` |
| Model | `gpt-5.6-sol` |
| Reasoning level | `high` |

No blocking findings. The commit-only probe preserves the live deferral classifier inputs (contributing states, envelope/projection checks, retained-anchor horizon, and distinct-digest fork predicate) while removing payload clones and unrelated graph materialization. The existing contested/uncontested lineage pair and all exact-head CI checks are green.

Datawav (dismissed):

PR #1621, opened 2026-09-01
Cut the deferred-peel sweep's idle-poll cost

dannym-arx:
### Summary
Memoize the per-row commit digests behind the peel-sweep fingerprint, and stop loading the MLS group on convergence polls that return before resolving.

### Context
Two spots paid full-history or full-load work on paths that discard the result:

- `deferred_peel_context_fingerprint` runs at least twice per sweep (gate + final), and the sweep sits inside the up-to-16-pass drain loop, up to 32 computations per drain while a backlog exists. Each one re-listed the group's stored messages and re-ran decode + TLS parse + SHA-256 on every row via `stored_convergence_commit_digests`.
- `converge_stored_openmls_messages` captured `previous_group` and the reorg component snapshot (a full `MlsGroup::load`) before its early returns, so every scheduler poll of a still-collecting pass deserialized the group just to throw it away.

### What changed
- `DeferredPeelGroupState` gains a per-row commit-digest memo: a row id never maps to different wire bytes (content rows are keyed by their content dedup id), so its "commit digest or not a commit" verdict is computed once per row lifetime. State membership is still re-read from the fresh listing on every call; the memo is rebuilt to the listed rows each time, so it can't outgrow the stored history.
- The pre-apply captures for the reorg diff moved below the generation-barrier / no-input / still-collecting / completed early returns, to where a pass is actually resolving.

### Impact
Fingerprint computation drops from O(rows) parses per call to parses for new rows only (the listing walk and digest hash remain). A scheduler poll of an unfrozen pass no longer costs a group-record read + 13-read `MlsGroup::load` + tree deserialization.

## Summary by CodeRabbit

* **Performance**
  * Improved deferred message processing by reusing previously computed commit information when message contents are unchanged.

* **Reliability**
  * Updated message handling to detect and reclassify messages when stored payloads change.
  * Ensured cached information reflects currently available message data, including removed entries.
  * Preserved snapshot contents and error handling while capturing convergence state at the appropriate processing stage.

### DIFF
--- a/crates/cgka-engine/src/message_processor/mod.rs
+++ b/crates/cgka-engine/src/message_processor/mod.rs
@@ -34,7 +34,7 @@ use cgka_traits::storage::{
 use cgka_traits::transport::{TransportEnvelope, TransportMessage};
 use cgka_traits::types::{EpochId, GroupId, MemberId, MessageId};
 use sha2::{Digest, Sha256};
-use std::collections::BTreeSet;
+use std::collections::{BTreeSet, HashMap};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
@@ -174,6 +174,14 @@ pub(crate) struct DeferredPeelGroupState {
     /// logged, or copied into durable generation state, and process restart
     /// drops it naturally.
     candidate_cache: Option<DeferredPeelCandidateCacheEntry>,
+    /// Per-row verdicts backing `stored_convergence_commit_digests`: the
+    /// commit digest, or `None` for a row that is not an MLS-wire commit.
+    /// Sound to remember without invalidation because a row id never maps to
+    /// different wire bytes (content rows are keyed by their content dedup
+    /// id, and payload rewrites only attach metadata around the same wire
+    /// message). Rebuilt to the currently listed rows on every use, so it
+    /// cannot outgrow the group's stored history.
+    commit_digest_memo: HashMap<MessageId, Option<[u8; 32]>>,
 }
 
 struct DeferredPeelCandidateCacheEntry {
@@ -2218,7 +2226,7 @@ impl<S: StorageProvider> Engine<S> {
     /// While all three are unchanged, re-peeling a deferred row is guaranteed
     /// wasted work.
     fn deferred_peel_context_fingerprint(
-        &self,
+        &mut self,
         group_id: &GroupId,
     ) -> Result<[u8; 32], EngineError> {
         let epoch = self.epoch_manager.epoch(group_id).unwrap_or_default();
@@ -2251,12 +2259,27 @@ impl<S: StorageProvider> Engine<S> {
 
     /// Content digests of the stored commits that can contribute to this
     /// group's convergence graph, in a stable order.
+    ///
+    /// The fingerprint gate computes this set at least twice per sweep, so a
+    /// row's decode + TLS parse + digest is remembered in
+    /// `DeferredPeelGroupState::commit_digest_memo` and paid once per row
+    /// lifetime instead of once per computation. State membership is re-read
+    /// from the fresh listing every call; only the payload verdict is memoized.
     fn stored_convergence_commit_digests(
-        &self,
+        &mut self,
         group_id: &GroupId,
     ) -> Result<BTreeSet<[u8; 32]>, EngineError> {
+        let records = self.storage.list_messages(group_id, EpochId(0))?;
+        let mut memo = std::mem::take(
+            &mut self
+                .deferred_peel
+                .entry(group_id.clone())
+                .or_default()
+                .commit_digest_memo,
+        );
+        let mut next_memo = HashMap::with_capacity(records.len());
         let mut digests = BTreeSet::new();
-        for record in self.storage.list_messages(group_id, EpochId(0))? {
+        for record in records {
             // Share the graph's own membership predicate rather than restating
             // it: this set must describe exactly the commits a pass would build
             // candidate branches from, and a private copy of that match drifts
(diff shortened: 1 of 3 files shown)


Datawav (dismissed):

Datawav (dismissed):

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -2266,14 +2289,23 @@ impl<S: StorageProvider> Engine<S> {
             ) {
                 continue;
             }
-            let Some((_message, projection)) = decode_openmls_wire_projection(&record.payload)
-            else {
-                continue;
+            let verdict = match memo.remove(&record.id) {
```
erskingardner, line 2300: **[P1] Invalidate the memo when the stored payload changes**

This reuses a verdict by `MessageId` alone, but the same exact head explicitly supports overwriting a row under the same id with a different `StoredMessagePayload` variant (`persist_stored_message_payload`, including the `RawTransport` → `OpenMlsWire` regression for #369). If a fingerprint first caches `None` for a contributing raw row and that row is then overwritten as an MLS-wire commit, this branch keeps returning the cached `None`; the new commit digest is omitted, the context fingerprint can stay unchanged, and the deferred-peel gate can skip retrying despite newly available branch context. Key the memo by a payload fingerprint (or store one alongside the verdict), or invalidate it from every payload-write path, and add a regression that overwrites the row between two fingerprint calls.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-01T20:49:57Z` |
| Commit reviewed | `39d42d39b8d9da750d32263c79bf68b993613750` |
| Model | `gpt-5.6-sol` |
| Reasoning level | `high` |

The deferred-peel digest memo can retain a stale non-commit verdict after a supported same-ID payload replacement. That can keep the context fingerprint unchanged when a new commit becomes available and suppress the retry this fingerprint is meant to trigger. Exact-head CI is green, but it does not cover an overwrite between fingerprint calls.

Datawav (commented):
## Automated review

High — Stale per-message memo can suppress a newly stored commit
Confidence: High

Evidence:
- crates/cgka-engine/src/message_processor/distributed_convergence.rs, changed line 2292: the new memoization path reuses a commit-digest result using MessageId as the cache identity.
- The supplied head_context shows persist_stored_message_payload can replace the payload stored under an existing MessageId, including a RawTransport payload with OpenMlsWire data. MessageId therefore does not uniquely identify the bytes or payload variant being classified.

Triggering scenario:
1. A deferred-peel fingerprint processes a RawTransport row and memoizes None for its MessageId.
2. Storage later replaces that row under the same MessageId with an OpenMlsWire commit.
3. The next fingerprint reuses the cached None rather than classifying the replacement payload.

Consequence:
The commit is omitted from the convergence fingerprint. A relevant history change can consequently appear unchanged, causing the deferred-peel sweep to skip work and leave the group stale or waiting until the in-memory memo is discarded. This is a convergence/liveness regression.

Smallest sound correction:
Key the memo by both MessageId and a stable digest of the current stored payload, or store that payload digest beside the memoized result and recompute whenever it differs. Clearing the entry whenever persist_stored_message_payload replaces a row would also be sound if all replacement paths can reliably reach the same group state.

Missing regression test:
Exercise the production overwrite behavior by fingerprinting a RawTransport row, replacing the same MessageId with an OpenMlsWire commit, and fingerprinting again. Assert that the second fingerprint includes the commit/change and causes the deferred sweep to reconsider the row. That test fails on this head’s stale-cache behavior.

No physical-device evidence exists for this head because the automated device gate was not requested for this repository/PR.

Audit evidence:
- Head SHA: 39d42d39b8d9da750d32263c79bf68b993613750
- Diff coverage: complete
- Changed files inspected:
  - crates/cgka-engine/src/openmls_projection.rs
  - crates/cgka-engine/src/message_processor/distributed_convergence.rs
- The newly supplied discussion claim about cache invalidation remains applicable to the exact head. No other discussion-only claims were promoted to findings.

Risk lenses checked:
- focused-diff-correctness
- lifecycle-integration
- regression-tests

Check evidence:
The supplied evidence records cargo test -p cgka-engine as successful, but check_evidence.complete is false. It therefore supports only the recorded command result; it is not sufficient to claim that the test suite passed completely, and it does not cover the payload-replacement cache regression above.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-02T07:47:15Z` |
| Commit reviewed | `91efa8b19f3f58b8a5e3ab99c69e733cedc7fba1` |
| Model | `gpt-5.6-sol` |
| Reasoning level | `high` |

The blocking memo-invalidation issue is fixed. Each cached verdict is now paired with the SHA-256 of the exact encoded stored payload; a same-id payload replacement therefore forces reclassification, while unchanged payloads still avoid the decode and TLS parse. The new regression populates the memo, performs the supported RawTransport-to-OpenMlsWire overwrite from #369, and verifies that the deferred-peel context fingerprint changes.

No further blocking findings. At posting time, CodeRabbit and three exact-head platform builds were green; the main CI workflow was still queued without jobs, so the new test had not yet executed in repository CI.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -2266,14 +2292,28 @@ impl<S: StorageProvider> Engine<S> {
...
+                        projection.kind == crate::openmls_projection::OpenMlsContentKind::Commit
+                    })
+                    .map(|(_message, projection)| projection.message_digest),
             };
-            if projection.kind == crate::openmls_projection::OpenMlsContentKind::Commit {
-                digests.insert(projection.message_digest);
+            if let Some(digest) = verdict {
+                digests.insert(digest);
             }
+            next_memo.insert(record.id, (payload_hash, verdict));
```
Datawav, line 2311: **Medium — Drop the history-sized memo when deferred-peel work ends.** This inserts every graph-contributing stored row, including non-commits, into a per-group map. Once the backlog drains, `retry_deferred_peels` takes the `total == 0` return at lines 1619–1635 and only clears `candidate_cache`; terminal disband cleanup likewise calls `invalidate_deferred_peel_candidate_cache` without removing this map. A long-lived engine can therefore retain a group-history-sized allocation after there is no deferred work left, and can accumulate those allocations across disbanded groups. Clear `commit_digest_memo` when the refreshed deferred count reaches zero and remove the full `deferred_peel` entry during terminal group cleanup; add regressions for both backlog drain and disband so the cache remains scoped to the optimization's active lifetime.

Datawav (changes requested):
The stale-verdict blocker is fixed, but this head now retains the replacement memo after the work that needs it has ended; please bound its lifetime before merge. Reviewed with gpt-5.6-sol and Claude Opus 5.

The focused replacement regression and all 102 cgka-engine library tests pass locally. The full package command was not clean evidence: 19 deferred-peel integration cases failed at their shared convergence-policy setup with `InvalidPolicy("convergence policy must equal the pinned v1 baseline")`; exact-head GitHub CI is still running. Cursor Grok 4.6 and Composer were attempted first but unavailable because the subscription usage limit was exhausted.

PR #1633, opened 2026-09-02
Guard cached relay lists against stale events

### LINKED ISSUE #920: marmot-app: live-ingested NIP-65/inbox relay-list events overwrite the cached relay list with no recency guard, reverting a newer list to a stale replayed copy
## Summary

When a live directory-subscription event of kind NIP-65 (`10002`) or Marmot inbox (`10050`) arrives, the cached relay list is **overwritten unconditionally with no `created_at` recency check**. A stale replayed relay-list event therefore reverts the account's cached relay list to older relays, degrading outbox/inbox delivery until a fresher event happens to arrive. This is the relay-list sibling of the profile-revert guard that already exists on the profile path.

## Location

- `crates/marmot-app/src/directory/methods.rs:876-899` — `remember_directory_relay_list_event`
- reached via `crates/marmot-app/src/directory/methods.rs:901-932` — `ingest_directory_relay_event` (kind `10002` / `10050` arm, lines 919-921)
- `crates/marmot-app/src/directory/records.rs:194-208` — `directory_record_recency` (ignores relay lists)

## Details

`ingest_directory_relay_event` gates incoming events only through `directory_freshness().accepts(&record)`, which rejects **future**-dated events, not older ones. For kind `10002`/`10050` it then calls `remember_directory_relay_list_event`, which:

```rust
let mut entry = self
    .directory_entry_for_account_id(account_id_hex)?
    .unwrap_or_else(|| self.empty_directory_record(account_id_hex));
match record.event.kind {
    KIND_NIP65_RELAY_LIST => entry.relay_lists.nip65.relays = relays,   // unconditional overwrite
    KIND_MARMOT_INBOX_RELAY_LIST => entry.relay_lists.inbox.relays = relays,
    _ => return Ok(()),
}
...
self.save_directory_entry(&entry)
```

There is **no comparison of the event's `created_at` against what is cached**. The save-time reconciler `save_directory_entry` → `select_newer_directory_entry` compares entries by `directory_record_recency`, which is built only from `profile.created_at` / `key_package.created_at` and **ignores relay lists entirely**, so an entry whose only change is an older relay list still wins the reconcile.

Contrast the profile ingest path, which routes through `remember_directory_profile_if_newer` (`methods.rs:851-874`) with an explicit `cached.created_at >= profile.created_at` guard added for exactly this class of revert.

## Trigger / failure scenario

1. Account A publishes a new kind-10002 (or 10050) relay list; the app caches it.
2. A relay in the account's directory-subscription set (author/kind pass `accepts_live_event`) serves or replays A's **superseded** kind-10002/10050 event — e.g. a relay that lagged the update, or a misbehaving/malicious relay in the fan-out.
3. The forwarder ingests it; `remember_directory_relay_list_event` overwrites the cached list with the stale relays and the recency reconciler does not veto it.
4. The app now derives outbox and notification-server inbox relays from the reverted list, degrading message/welcome discovery and delivery for that account until a fresher event arrives.

## Suggested fix

Track the relay-list event `created_at` on the cached relay-list record and apply an `>=`-style guard (as the profile path does) before overwriting; alternatively fold relay-list recency into `directory_record_recency` so `select_newer_directory_entry` can veto a stale relay-list-only update.

---
*Found during a systematic code-review pass. Checked against existing issues #884 (profile-field revert — different field, and that path already has the `_if_newer` guard this one lacks), #912 (follow-edge wipe on empty contact-list), #894 (search-graph promotion wiping follow edges), #867/#911 (profile extra-field cap / control chars). None cover the relay-list ingest recency gap.*

erskingardner:
## Summary
- persist the originating replaceable-event timestamp with each cached NIP-65 and Marmot inbox relay-list state
- reject older and equal-timestamp live replays while allowing a zero-timestamp event to populate legacy empty state
- reconcile profile, KeyPackage, NIP-65, and inbox independently across account and shared caches so a fresh sibling field cannot drag stale state with it
- stamp locally published relay lists from the authoritative signed events while retaining backward compatibility for timestamp-less cache JSON

Fixes #920

## Test plan
- [x] New live-ingest regression failed before the fix and passes after
- [x] Divergent account/shared-cache regression failed before component-wise reconciliation and passes after
- [x] Zero-timestamp legacy-empty regression
- [x] Adjacent newer-list replacement and profile recency regressions
- [x] `just fast-ci`
- [ ] `cargo test -p marmot-app -- --test-threads=1` (868 passed; the same 6 unrelated epoch-backfill tests reported on the original head still fail)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed relay-list updates being incorrectly rejected when no prior timestamp existed.
  * Improved directory synchronization by merging profile, key-package, and relay-list data independently according to each component’s timestamp.
  * Prevented stale relay-list events from replacing newer cached data while preserving preferred data when timestamps match.
  * Preserved relay-list event timestamps across initialization, publication, parsing, caching, and restoration.

### DIFF
--- a/crates/marmot-app/src/directory/methods.rs
+++ b/crates/marmot-app/src/directory/methods.rs
@@ -1167,12 +1167,14 @@ impl MarmotApp {
         // publication, matching the profile ingest rule (mdk#920).
         match record.event.kind {
             KIND_NIP65_RELAY_LIST
-                if entry.relay_lists.nip65.created_at >= record.event.created_at =>
+                if entry.relay_lists.nip65.created_at != 0
+                    && entry.relay_lists.nip65.created_at >= record.event.created_at =>
             {
                 return Ok(());
             }
             KIND_MARMOT_INBOX_RELAY_LIST
-                if entry.relay_lists.inbox.created_at >= record.event.created_at =>
+                if entry.relay_lists.inbox.created_at != 0
+                    && entry.relay_lists.inbox.created_at >= record.event.created_at =>
             {
                 return Ok(());
             }
--- a/crates/marmot-app/src/directory/methods.rs
+++ b/crates/marmot-app/src/directory/methods.rs
@@ -1162,7 +1162,20 @@ impl MarmotApp {
         let mut entry = self
             .directory_entry_for_account_id(account_id_hex)?
             .unwrap_or_else(|| self.empty_directory_record(account_id_hex));
+        // Replaceable-event timestamps have second resolution. Keep the cache
+        // on equality so a lagging relay cannot revert an accepted same-second
+        // publication, matching the profile ingest rule (mdk#920).
         match record.event.kind {
+            KIND_NIP65_RELAY_LIST
+                if entry.relay_lists.nip65.created_at >= record.event.created_at =>
+            {
+                return Ok(());
+            }
+            KIND_MARMOT_INBOX_RELAY_LIST
+                if entry.relay_lists.inbox.created_at >= record.event.created_at =>
+            {
+                return Ok(());
+            }
             KIND_NIP65_RELAY_LIST => entry.relay_lists.nip65 = state,
             KIND_MARMOT_INBOX_RELAY_LIST => entry.relay_lists.inbox = state,
             _ => return Ok(()),
--- a/crates/marmot-app/src/directory/records.rs
+++ b/crates/marmot-app/src/directory/records.rs
@@ -262,6 +262,10 @@ fn directory_record_recency(entry: &UserDirectoryRecord) -> u64 {
                 .as_ref()
                 .map(|key_package| key_package.created_at),
         )
+        .chain([
+            entry.relay_lists.nip65.created_at,
+            entry.relay_lists.inbox.created_at,
+        ])
         .max()
         .unwrap_or_default()
 }
--- a/crates/marmot-app/src/directory/records.rs
+++ b/crates/marmot-app/src/directory/records.rs
@@ -250,43 +250,56 @@ pub(crate) fn user_directory_record_from_public(
     })
 }
 
-fn directory_record_recency(entry: &UserDirectoryRecord) -> u64 {
-    entry
-        .profile
-        .as_ref()
-        .map(|profile| profile.created_at)
-        .into_iter()
-        .chain(
-            entry
-                .key_package
-                .as_ref()
-                .map(|key_package| key_package.created_at),
-        )
-        .chain([
-            entry.relay_lists.nip65.created_at,
-            entry.relay_lists.inbox.created_at,
-        ])
-        .max()
-        .unwrap_or_default()
+/// Merge independently replaceable Nostr components without allowing a fresh
+/// sibling field to drag an older component across cache boundaries.
+///
+/// The first record wins equal timestamps. Nostr timestamps have one-second
+/// resolution, so this preserves a just-published local value against a stale
+/// same-second relay copy. Fields without event recency remain from the first
+/// record; public/shared records do not carry local-account or source hints.
+fn merge_directory_entries(
+    mut preferred: UserDirectoryRecord,
+    alternate: UserDirectoryRecord,
+) -> UserDirectoryRecord {
+    if alternate.profile.as_ref().is_some_and(|candidate| {
+        preferred
+            .profile
+            .as_ref()
+            .is_none_or(|current| candidate.created_at > current.created_at)
+    }) {
+        preferred.profile = alternate.profile;
+    }
+    if alternate.key_package.as_ref().is_some_and(|candidate| {
+        preferred
+            .key_package
+            .as_ref()
+            .is_none_or(|current| candidate.created_at > current.created_at)
+    }) {
+        preferred.key_package = alternate.key_package;
+    }
+    if alternate.relay_lists.nip65.created_at > preferred.relay_lists.nip65.created_at {
+        preferred.relay_lists.nip65 = alternate.relay_lists.nip65;
+    }
+    if alternate.relay_lists.inbox.created_at > preferred.relay_lists.inbox.created_at {
+        preferred.relay_lists.inbox = alternate.relay_lists.inbox;
+    }
+    preferred.relay_lists.refresh();
+    preferred
 }
 
+/// Reconcile account-cache and shared-cache directory records component-wise.
 pub(crate) fn select_newer_directory_entry(
     cached: Option<UserDirectoryRecord>,
     shared: Option<UserDirectoryRecord>,
 ) -> Option<UserDirectoryRecord> {
     match (cached, shared) {
-        (Some(cached), Some(shared)) => {
-            if directory_record_recency(&shared) > directory_record_recency(&cached) {
-                Some(shared)
-            } else {
-                Some(cached)
-            }
-        }
+        (Some(cached), Some(shared)) => Some(merge_directory_entries(cached, shared)),
         (Some(entry), None) | (None, Some(entry)) => Some(entry),
         (None, None) => None,
     }
 }
 
+/// Insert a directory record or merge each timestamped component independently.
 pub(crate) fn upsert_newer_directory_entry(
     entries_by_id: &mut BTreeMap<String, UserDirectoryRecord>,
     entry: UserDirectoryRecord,
@@ -296,9 +309,8 @@ pub(crate) fn upsert_newer_directory_entry(
             slot.insert(entry);
         }
         std::collections::btree_map::Entry::Occupied(mut slot) => {
-            if directory_record_recency(&entry) > directory_record_recency(slot.get()) {
-                *slot.get_mut() = entry;
-            }
+            let current = slot.get().clone();
+            *slot.get_mut() = merge_directory_entries(current, entry);
         }
     }
 }
--- a/crates/cli/src/commands/account.rs
+++ b/crates/cli/src/commands/account.rs
@@ -239,12 +239,14 @@ fn missing_relay_list_status(missing: Vec<MissingRelayListKind>) -> AccountRelay
         bootstrap_relays: Vec::new(),
         nip65: marmot_app::AccountRelayListState {
             kind: 10002,
+            created_at: 0,
             relays: Vec::new(),
             read_relays: Vec::new(),
             write_relays: Vec::new(),
         },
         inbox: marmot_app::AccountRelayListState {
             kind: 10050,
+            created_at: 0,
             relays: Vec::new(),
             read_relays: Vec::new(),
             write_relays: Vec::new(),
--- a/crates/cli/src/commands/relays.rs
+++ b/crates/cli/src/commands/relays.rs
@@ -201,6 +201,7 @@ mod tests {
     fn nip65_update_preserves_unmodified_directional_relays() {
         let state = AccountRelayListState {
             kind: 10_002,
+            created_at: 0,
             relays: vec!["wss://both.example".into(), "wss://write.example".into()],
             read_relays: vec!["wss://both.example".into(), "wss://read.example".into()],
             write_relays: vec!["wss://both.example".into(), "wss://write.example".into()],
--- a/crates/marmot-app/src/lib.rs
+++ b/crates/marmot-app/src/lib.rs
@@ -716,6 +716,11 @@ impl MissingRelayListKind {
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct AccountRelayListState {
     pub kind: u64,
+    /// Timestamp of the replaceable event that produced this cached state.
+    ///
+    /// Zero denotes a legacy or caller-constructed state with unknown recency.
+    #[serde(default, skip_serializing_if = "is_zero_u64")]
+    pub created_at: u64,
     /// Direction-appropriate relay targets for this list.
     ///
     /// For NIP-65 kind 10002 this is specifically the write-capable set:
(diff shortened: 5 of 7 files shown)


crates/marmot-app/src/directory/methods.rs
```diff
@@ -1162,7 +1162,20 @@ impl MarmotApp {
         let mut entry = self
             .directory_entry_for_account_id(account_id_hex)?
             .unwrap_or_else(|| self.empty_directory_record(account_id_hex));
+        // Replaceable-event timestamps have second resolution. Keep the cache
+        // on equality so a lagging relay cannot revert an accepted same-second
+        // publication, matching the profile ingest rule (mdk#920).
         match record.event.kind {
+            KIND_NIP65_RELAY_LIST
+                if entry.relay_lists.nip65.created_at >= record.event.created_at =>
```
erskingardner: Non-blocking, for awareness: this guard compares against the *merged* entry (`directory_entry_for_account_id` returns the whole-entry `select_newer` winner), so its protection is bounded by entry-level recency. If the shared store ever holds a divergent entry with a fresher profile but a staler relay-list stamp, that entry wins wholesale, the locally cached newer relay list is displaced before this guard runs, and a stale replay within that window can still overwrite. That's inherent to the existing whole-entry reconcile (the same drag already existed between profile and key_package), not a defect of this change — but per-field recency/merge would be the eventual hardening if this class of revert resurfaces.

crates/marmot-app/src/directory/records.rs
```diff
@@ -262,6 +262,10 @@ fn directory_record_recency(entry: &UserDirectoryRecord) -> u64 {
                 .as_ref()
                 .map(|key_package| key_package.created_at),
         )
+        .chain([
+            entry.relay_lists.nip65.created_at,
```
erskingardner: Non-blocking: with relay stamps folded into the entry-level max, an entry with a fresh relay list but a stale profile/key package can now win `select_newer_directory_entry` / `upsert_newer_directory_entry` and revert those sibling fields — the converse of the cross-field drag that already existed between profile and key package. Acceptable tradeoff for the fix, but it might be worth a small `select_newer_directory_entry` unit test that pins the intended semantics for relay-stamp divergence (e.g. fresh relay + stale profile vs. fresh profile + stale relay), since none of the current tests exercise the recency fold.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-02T15:16:26Z` |
| Commit reviewed | `12646ff993d6451fc1bbace15a031a67e9419ff0` |
| Recommended action | **Merge** (approve unavailable: own PR) |

This is a well-scoped fix for a real bug (#920), and it follows the established profile-ingest precedent (`remember_directory_profile_if_newer`, mdk#206) closely.

What I verified:

- The `>=` guard in `remember_directory_relay_list_event` mirrors the profile-path rule, and the equality-keep rationale (second-resolution replaceable timestamps) is documented inline and matches the existing profile guard's reasoning.
- Publish-path stamping is correct: `requests` is built 1:1 and in order from `list_kinds`, so the new `zip(&requests)` pairing cannot mismatch; `requests[0]`/`requests[1]` in the bootstrap path match the `[Nip65, Inbox, contact, profile]` construction order.
- `account_relay_list_status_for_account_id` returns `entry.relay_lists` directly, so a single-kind publish preserves the other kind's cached stamp instead of zeroing it.
- Serde `default` + `skip_serializing_if` keeps timestamp-less legacy cache JSON compatible in both directions; all exhaustive struct-literal construction sites were updated (check/clippy are green).
- The new regression test covers both list kinds plus the equal-timestamp case, and the adjacent `ingesting_all_read_nip65_list...` test was correctly re-stamped (1→2) so it still exercises replacement rather than tripping the new guard.

Non-blocking notes on the entry-level reconcile limitation are inline.

Validation note: the author flagged 6 locally-failing epoch-backfill timing tests in `cargo test -p marmot-app` as unrelated; several CI test shards are still in progress at review time. Recommend letting the full CI matrix finish green before merging.

crates/marmot-app/src/directory/records.rs
```diff
@@ -262,6 +262,10 @@ fn directory_record_recency(entry: &UserDirectoryRecord) -> u64 {
...
+            entry.relay_lists.inbox.created_at,
+        ])
```
erskingardner: Non-blocking: folding relay-list timestamps into the max-across-fields recency heuristic tightens cross-cache reconciliation, but entry-level selection in `select_newer_directory_entry` still moves whole entries on the max field. In the mixed case (an entry with a newer relay list but an older cached profile), the relay timestamp can let that entry win wholesale and drop the newer profile. The per-kind guard added in `remember_directory_relay_list_event` covers the live-ingest path that motivated mdk#920, so this residual edge is only for cache-vs-cache reconcile — acceptable for this fix, but worth knowing the recency fold is heuristic, not field-granular.

crates/marmot-app/src/directory/methods.rs
erskingardner: Two non-blocking edge notes on this guard: (1) a zero `created_at` incoming event can never populate an empty cache, since `0 >= 0` skips — harmless in practice because relay-fed replaceable events carry nonzero timestamps, but synthetic/unsigned zero-timestamp records will be silently dropped here. (2) Early return also skips the `source_relays_from_record` bootstrap-hint accumulation below — arguably desirable (stale events shouldn't influence bootstrap hints), just calling it out as an intentional side effect.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-02T15:19:16Z` |
| Commit reviewed | `12646ff993d6451fc1bbace15a031a67e9419ff0` |
| Recommended action | **Merge** (approve unavailable: own PR) |

Well-scoped fix for the real bug in #920, closely following the established profile-ingest precedent (`remember_directory_profile_if_newer`, mdk#206).

Verified:

- The `>=` keep-cache guard in `remember_directory_relay_list_event` mirrors the profile-path rule, with the second-resolution/equality rationale documented inline.
- Publish-path stamping is safe: `requests` is built 1:1 and in order from `list_kinds`, so the `zip(&requests)` pairing can't mismatch; the bootstrap path's `requests[0]`/`requests[1]` match the fixed `[Nip65, Inbox]` construction order.
- `account_relay_list_status_for_account_id` clones `entry.relay_lists` wholesale, so a single-kind publish preserves the other kind's cached timestamp instead of zeroing it.
- Serde `default` + `skip_serializing_if("is_zero_u64")` keeps timestamp-less legacy cache JSON compatible in both directions; all exhaustive struct-literal construction sites updated (check/clippy/format green in CI).
- The new regression test covers both list kinds plus the equal-timestamp case, and the adjacent `ingesting_all_read_nip65_list...` test was re-stamped (1→2) so it still exercises replacement rather than tripping the new guard.

Two non-blocking edge notes left inline (zero-timestamp events + bootstrap-hint skip on guard short-circuit; recency fold is max-of-fields heuristic).

Validation note: the author flagged 6 locally-failing epoch-backfill timing tests in `cargo test -p marmot-app` as unrelated; several CI shards (including Rust tests 1/4 and 3/4) were still in progress at review time with everything completed so far green. Recommend letting the full CI matrix finish before merge.

crates/marmot-app/src/directory/records.rs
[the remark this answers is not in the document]
erskingardner: Addressed in 76880234. Directory reconciliation now merges profile, KeyPackage, NIP-65, and inbox independently by each component timestamp, with the first/cache record winning equality. The regression constructs divergent cache/shared records (profile 100 vs 1, NIP-65 10 vs 20, inbox 30 vs 20, KeyPackage 30 vs 40), verifies the component-wise winners, then verifies a NIP-65 event at 15 cannot replace the shared timestamp-20 state. `just fast-ci` and all focused directory tests pass.

crates/marmot-app/src/directory/records.rs
```diff
@@ -250,39 +250,56 @@ pub(crate) fn user_directory_record_from_public(
...
+/// The first record wins equal timestamps. Nostr timestamps have one-second
+/// resolution, so this preserves a just-published local value against a stale
+/// same-second relay copy. Fields without event recency remain from the first
+/// record; public/shared records do not carry local-account or source hints.
+fn merge_directory_entries(
+    mut preferred: UserDirectoryRecord,
+    alternate: UserDirectoryRecord,
+) -> UserDirectoryRecord {
+    if alternate.profile.as_ref().is_some_and(|candidate| {
+        preferred
```
erskingardner, line 265: Non-blocking observation on the documented "fields without event recency remain from the first record" rule: when `alternate` supplies the winning relay-list component, its `bootstrap_relays`/publish-route hints accumulated with that list stay behind in the discarded record, while preferred's (older-list) hints survive. If republish routing ever relies on hints belonging to the winning list's sources, a small union here would close that; if the hint list is intentionally conservative, the current behavior is fine. Either way, worth one sentence in the doc comment.

crates/marmot-app/src/directory/methods.rs
```diff
@@ -1162,7 +1162,22 @@ impl MarmotApp {
         let mut entry = self
             .directory_entry_for_account_id(account_id_hex)?
             .unwrap_or_else(|| self.empty_directory_record(account_id_hex));
+        // Replaceable-event timestamps have second resolution. Keep the cache
+        // on equality so a lagging relay cannot revert an accepted same-second
+        // publication, matching the profile ingest rule (mdk#920).
         match record.event.kind {
+            KIND_NIP65_RELAY_LIST
+                if entry.relay_lists.nip65.created_at != 0
```
erskingardner, line 1170: Nano-nit: the comment above this match still only explains the equality-keep rule; the new `!= 0` bypass (zero stamp = unknown recency, never blocks ingest) is the subtler half of the guard and deserves a clause there so future readers don't "simplify" it away.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-02T15:51:13Z` |
| Commit reviewed | `76880234178171c1e594f7423f252bbeb7bf3231` |
| Recommended action | **Merge** (approve unavailable: own PR) |

Re-review of `7688023` ("Reconcile directory event state per component"), which replaces the entry-level max-recency heuristic I flagged last round.

The rework resolves both prior non-blocking concerns, and I verified the details:

- `merge_directory_entries` merges profile / key package / NIP-65 / inbox independently with strict `>`, so a fresh sibling can no longer drag a stale component across cache boundaries. First/preferred record wins ties, which composes correctly with the ingest guard's equality-keep: a same-second relay echo of a just-published value cannot displace it through either mechanism. Non-timestamped fields (follows, local_account, hints) staying with preferred is documented as deliberate — that also means `local_account` can never be lost to a shared record, an improvement over the old whole-entry pick.
- The `!= 0` bypass on the ingest guard correctly treats a zero stamp as unknown recency: legacy/unstamped caches stay overwritable and zero-timestamp records can seed empty state, which the new `zero_timestamp_relay_list_event_populates_legacy_empty_state` test pins.
- The new `directory_entry_reconciles_nostr_components_independently` regression is exactly the divergence test I asked for (profile 100 vs 1, NIP-65 10 vs 20, inbox 30 vs 20, key package 30 vs 40), and it additionally proves the stale-replay guard now defends the *merged* stamp 20 even though the cache row alone held 10 — closing the displacement window I described.
- `directory_record_recency` is fully removed with no remaining callers; check/clippy and all Rust test shards are green.

On the open CodeRabbit threads, for the author's triage:

- "Add an equal-timestamp replay case" to the component-merge test is already covered: `stale_live_relay_list_events_do_not_replace_newer_cached_lists` ingests an equal-timestamp event (20 after 10) for both kinds and asserts the cached list survives. No change needed unless you want the assertion duplicated in the newer test.
- "Test actual timestamp-less cache JSON" (serde `default` legacy deserialization) is a reasonable optional hardening; the `#[serde(default)]` path is straightforward, so I would not hold the merge for it.

Remaining validation gap: "Simulator and vectors" was still in progress at review time — let it finish green before merging.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-02T15:53:01Z` |
| Commit reviewed | `76880234178171c1e594f7423f252bbeb7bf3231` |
| Recommended action | **Merge** (approve unavailable: own PR) |

Re-review of the reworked reconciliation (commit `76880234`): both prior concerns are resolved and the revision is stronger than the v1 approach.

Verified:

- `merge_directory_entries` now reconciles profile, key package, NIP-65, and inbox independently per component timestamp, with the first record winning ties — this closes the wholesale entry-swap edge I flagged inline on the previous revision, and addresses CodeRabbit's Major finding the same way. All four call sites (`directory_entry_for_account_id_with_handles` read-merge, `save_directory_entry_with_reason` save-merge, `directory_entries` bulk upsert, and the public-record hydration path) use consistent preferred=local/first argument order.
- The zero-timestamp edge from my last review is closed: the ingest guard now requires a nonzero cached stamp (`!= 0 && cache >= event`), and the new `zero_timestamp_relay_list_event_populates_legacy_empty_state` test pins that behavior.
- Untimestamped fields (follows, bootstrap hints, local account) deliberately stay with the first record; the doc comment on `merge_directory_entries` says so explicitly. `refresh()` re-derives `default_relays`/`missing`/`complete` after each merge.
- The component-independence regression test exercises a four-way mixed-freshness reconcile (cached profile + shared nip65 + cached inbox + shared key package) plus a post-reconcile stale replay — good coverage of exactly the failure mode the v1 recency fold could have introduced.
- CodeRabbit's two Minor test suggestions: the equal-timestamp replay case it asked for already exists in `stale_live_relay_list_events_do_not_replace_newer_cached_lists` (ingest 20, then 10, then 20) — false positive; the legacy-JSON suggestion is a fair Minor follow-up, non-blocking.

Validation note: at posting time CI shows 26 SUCCESS (all check/clippy/format/test shards, e2e, Tamarin) with only "Simulator and vectors" still in progress — recommend letting it finish, consistent with the earlier revision where it passed.

PR #1634, opened 2026-09-02
Bound SQLCipher mlock diagnostics on Android

### LINKED ISSUE #1522: SQLCipher mlock failures flood logs and obscure degraded memory-lock guarantees
## Problem

A 143,110-line GrapheneOS White Noise Android client-log capture recorded 73,159 SQLCipher `mlock()` failures with `errno=12` (`ENOMEM`) from 17:47:39.716 through 17:47:48.044 (8.328 seconds), peaking at 10,466 warnings per second. SQLCipher labeled the failures `WARN MEMORY`; Android emitted them under the `sqlcipher` debug tag.

The same capture contains 69,471 `liblog` drop-report lines. Only nine drop-report lines fall in the same whole-second buckets as the SQLCipher warning burst, so the evidence establishes severe diagnostic pressure and loss during the same session but does not attribute every dropped entry to SQLCipher.

This is not claimed as the cause of the separately reported message-send delay: the capture lacks the privacy-safe send-phase traces needed to establish causality.

## Source-confirmed behavior

At MDK `master` commit `baae0be21bc44cebc89c9a13f2cb50524307c255`:

- [`SqliteStorageOptions::default`](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/crates/storage-sqlite/src/connection.rs#L659-L680) enables `cipher_memory_security` for account databases.
- [`open_encrypted_with_options`](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/crates/storage-sqlite/src/connection.rs#L747-L769) applies cipher pragmas before keying.
- [`apply_cipher_pragmas`](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/crates/storage-sqlite/src/connection.rs#L840-L856) turns memory security on when configured; MDK does not expose whether the platform honored the resulting page locks.
- The workspace uses [`rusqlite` 0.40.1 with bundled vendored SQLCipher](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/Cargo.toml#L108) and locks [`libsqlite3-sys` 0.38.1](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/Cargo.lock#L3233-L3241). The corresponding bundled SQLCipher 4.14.0 lineage is also identified in the upstream Windows report for the same dependency pair.
- SQLCipher's secure allocator attempts `mlock` and logs every failed call in this bundled source ([versioned amalgamation](https://docs.rs/crate/libsqlite3-sys/0.38.1/source/sqlcipher/sqlite3.c)). A failed lock is non-fatal: SQLCipher continues and wipes memory on free, but the page-locking guarantee was not achieved.
- [`open_hardened_sqlcipher`](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/crates/storage-sqlite/src/connection.rs#L890-L975) applies the same policy to the encrypted directory-cache, migration, and rekey paths. [`SqliteSharedStorage`](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/crates/storage-sqlite/src/shared.rs#L77-L81) is unencrypted and is not part of this SQLCipher contract.

The upstream SQLCipher lineage already owns the allocator-level Android behavior. [sqlcipher/sqlcipher#530](https://github.com/sqlcipher/sqlcipher/issues/530) documents Android `errno=12`, the non-fatal degraded lock guarantee, the 4.7.0 pre-allocated locking mitigation, a reported 4.15.0 regression, and the 4.16.0 fix. The [SQLCipher 4.16.0 release](https://github.com/sqlcipher/sqlcipher/releases/tag/v4.16.0) removes redundant `mlock`/`VirtualLock` warnings and fixes an allocation issue that increased Android warnings.

MDK therefore needs to adopt and verify the fixed dependency lineage first, while retaining an explicit host policy for any platform where effective locking remains unavailable.

## Ownership boundary

- SQLCipher owns the secure allocator, per-allocation `mlock` attempt, wipe-on-free behavior, and default Android logging.
- `libsqlite3-sys` packages the SQLCipher amalgamation; `rusqlite` selects that bundled dependency but does not implement the allocator policy.
- MDK's `storage-sqlite` crate owns when `cipher_memory_security` is enabled and what effective/degraded state callers can observe. The [`marmot-uniffi` Kotlin packaging script](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/crates/marmot-uniffi/kotlin-bindings.sh#L240-L248) owns the Android-target build inputs and JNI artifact validation.
- White Noise Android consumes the resulting MDK artifact. No client-side policy change is required unless MDK exposes a new aggregate diagnostic or device-regression hook, so a separate Android issue is not warranted from the current evidence.

## Security-preserving fallback constraints

- Upgrading to SQLCipher 4.16.0 or later is the preferred first fix because upstream explicitly removed redundant lock-failure warnings and fixed an Android-heavy allocation regression.
- If a supported Android target still cannot lock the secure heap, a target-specific dependency configuration such as SQLCipher's `OMIT_MEMLOCK` path may prevent futile per-allocation locks while retaining the secure allocator's wipe-on-free behavior. That is a degraded page-lock posture, not equivalent protection, and must be surfaced once per process or as bounded aggregate state.
- Filtering all SQLCipher `MEMORY` logs is not sufficient by itself: it suppresses the storm but also hides whether locking is effective. Any filter must be paired with a tested, privacy-safe effective/degraded signal.
- Do not disable `cipher_memory_security` at runtime or weaken keying, encryption, database-format compatibility, or wipe-on-free guarantees.

## Impact

- Database-heavy work can emit tens of thousands of warnings per second and crowd crash, send-path, and convergence diagnostics out of logcat.
- Repeated formatting and logging add avoidable load on user-visible application paths.
- Operators cannot distinguish “memory security configured” from “memory locking effective.”
- The observation is verified on GrapheneOS. Other Android versions and lock budgets remain unmeasured.

## Required behavior

- Upgrade the bundled SQLCipher lineage to a release containing the upstream 4.16.0 fix, unless compatibility testing proves that upgrade unsafe; record any temporary pin and its removal condition.
- Preserve SQLCipher keying order, encrypted storage, and zeroization-on-free.
- Verify that GrapheneOS no longer receives one warning per failed allocation under a representative database workload.
- If the upgraded dependency can still exhaust the platform lock budget, expose one bounded, privacy-safe degraded-state diagnostic rather than globally suppressing SQLCipher warnings.
- Keep real database and cryptographic failures observable, and keep memory security enabled by default where the platform can honor it.

## Acceptance criteria

- The locked dependency resolves to SQLCipher 4.16.0 or a later release that retains its Android allocation and redundant-warning fixes; dependency and database-format compatibility tests pass.
- A deterministic native test or controlled harness forces the lock-failure path and proves a database-heavy workload emits a bounded number of diagnostics rather than one warning per allocation.
- Degraded-path coverage verifies sensitive allocations are still wiped on free, or documents and tests an equally strong replacement.
- Any MDK-owned diagnostic reports only aggregate configured/effective/degraded state; it excludes pointers, allocation sizes, forwarded SQLCipher format strings, SQL text, database paths, account/group/message identifiers, and key material.
- Account, encrypted directory-cache, migration, and rekey opens use the same reviewed policy; the unencrypted shared store remains out of scope.
- A GrapheneOS device regression opens and exercises a representative account database without a SQLCipher warning storm or attributable `liblog` drop reports.
- Storage tests continue to cover pragma-before-key ordering and SQLCipher compatibility.

## Related work and duplicate disposition

- [sqlcipher/sqlcipher#530](https://github.com/sqlcipher/sqlcipher/issues/530) is the canonical upstream Android `mlock`/`errno=12` issue. This MDK issue owns dependency adoption, host-policy integration, and end-to-end verification in MDK/White Noise.
- [SQLCipher 4.7.0](https://github.com/sqlcipher/sqlcipher/releases/tag/v4.7.0) introduced constrained-platform pre-allocation; [SQLCipher 4.16.0](https://github.com/sqlcipher/sqlcipher/releases/tag/v4.16.0) fixed the later allocation regression and redundant warnings.
- [sqlcipher/sqlcipher#602](https://github.com/sqlcipher/sqlcipher/pull/602) addresses Windows-only logger/allocator recursion and is not the Android flood.
- [mdk#1173](https://github.com/marmot-protocol/mdk/pull/1173) introduced the current bundled SQLCipher 4.14.0 lineage; it does not contain the later Android warning fix.
- [mdk#1419](https://github.com/marmot-protocol/mdk/pull/1419) and [mdk#1409](https://github.com/marmot-protocol/mdk/issues/1409) raised the nightly runner's memlock limit to protect a simulator campaign; they explicitly did not change production storage behavior.
- [mdk#678](https://github.com/marmot-protocol/mdk/issues/678) and [mdk#566](https://github.com/marmot-protocol/mdk/issues/566) established pragma-before-key hardening. This issue preserves that invariant.
- [mdk#1018](https://github.com/marmot-protocol/mdk/issues/1018) covers key-format buffer zeroization/passphrase handling; [mdk#1439](https://github.com/marmot-protocol/mdk/issues/1439) covers duplicate SQLCipher KDF work. Neither owns failed-lock warning amplification.
- [whitenoise-android#913](https://github.com/marmot-protocol/whitenoise-android/issues/913) and [mdk#1176](https://github.com/marmot-protocol/mdk/issues/1176) cover send latency instrumentation and deferred-peel scheduling. This report does not claim causality with the observed send delay.

The issue remains a standalone native `Bug`: the inspected native Tracking issues do not have a completion rule that requires this work.

erskingardner:
## Summary

- Pin libsqlite3-sys to the merged SQLCipher 4.17 amalgamation, eliminating the per-allocation mlock WARN storm while preserving encryption, allocation wiping, and best-effort memory locking.
- Emit one privacy-safe MDK warning per Android process when enhanced memory security is enabled, with fixed configured and platform_limited fields.
- Guard the bundled SQLCipher lineage and the bounded diagnostic with regression tests, and document the Android posture.

## Root cause

Android 14 and later cap mlock usage at 64 KiB per process. SQLCipher 4.14 treats exhausted best-effort memory locks as non-fatal but emits a WARN for every failed allocation. SQLCipher 4.17 lowers that internal message below the default log level, so this patch retains one actionable MDK-owned diagnostic without flooding logs. Database reads, writes, encryption, and memory sanitization continue normally; only residency protection above the platform budget is lost.

## Test plan

- [x] Confirmed the SQLCipher version regression test fails against the previous 4.14 amalgamation.
- [x] cargo test -p storage-sqlite (446 passed, 4 ignored)
- [x] just fast-ci on Rust 1.97.1
- [x] cargo check -p storage-sqlite --target aarch64-linux-android against NDK API 26
- [ ] Reproduce the original workload on GrapheneOS and confirm one MDK platform_limited warning with no SQLCipher WARN storm.

Fixes #1522

## Summary by CodeRabbit

- **Security**
  - Updated the bundled SQLCipher protection to version 4.17.0.
  - Improved memory-security handling for encrypted storage.

- **Bug Fixes**
  - Android now provides a clear, limited warning when platform restrictions prevent full memory locking. Memory sanitization remains active.

- **Documentation**
  - Clarified that memory-security settings apply across the application process and that Android memory locking is best effort.

### DIFF
--- a/crates/storage-sqlite/src/connection.rs
+++ b/crates/storage-sqlite/src/connection.rs
@@ -7,6 +7,8 @@ use cgka_traits::types::Backend;
 use std::fmt;
 use std::ops::{Deref, DerefMut};
 use std::path::Path;
+#[cfg(any(target_os = "android", test))]
+use std::sync::Once;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Condvar, Mutex, MutexGuard};
 use std::thread::ThreadId;
@@ -31,6 +33,26 @@ const BUSY_BACKOFF_BASE: Duration = Duration::from_millis(20);
 
 /// Upper bound on a single busy-retry backoff sleep.
 const BUSY_BACKOFF_CAP: Duration = Duration::from_millis(250);
+
+/// Android imposes a finite process-wide `mlock` budget (64 KiB on API 34+),
+/// so SQLCipher's enhanced whole-library memory locking is necessarily best
+/// effort there. SQLCipher 4.17 keeps allocation failures below its default
+/// log level; retain one bounded MDK-owned diagnostic instead.
+#[cfg(target_os = "android")]
+static SQLCIPHER_PLATFORM_LIMIT_REPORTED: Once = Once::new();
+
+#[cfg(any(target_os = "android", test))]
+fn report_sqlcipher_platform_limit_once(report_once: &Once) {
+    report_once.call_once(|| {
+        tracing::warn!(
+            target: "storage_sqlite::connection",
+            method = "configure_cipher_memory_security",
+            configured = true,
+            lock_posture = "platform_limited",
+            "SQLCipher memory locking is best effort on Android"
+        );
+    });
+}
 
 /// An error that may represent transient SQLite lock contention worth retrying.
 /// Implemented for the two storage error types so [`retry_on_busy`] can drive a
@@ -663,6 +685,10 @@ pub struct SqliteStorageOptions {
     pub secure_delete: bool,
     pub temp_store_memory: bool,
     pub trusted_schema: bool,
+    /// Enable SQLCipher's process-global enhanced memory allocator. Once any
+    /// connection enables it, it remains enabled for every SQLite connection
+    /// in the process. Android receives a once-per-process diagnostic because
+    /// its finite `mlock` budget makes page locking best effort.
     pub cipher_memory_security: bool,
     pub cipher_compatibility: u8,
 }
@@ -852,6 +878,8 @@ fn apply_cipher_pragmas(
         connection
             .execute_batch("PRAGMA cipher_memory_security = ON;")
             .storage()?;
+        #[cfg(target_os = "android")]
+        report_sqlcipher_platform_limit_once(&SQLCIPHER_PLATFORM_LIMIT_REPORTED);
     }
     Ok(())
 }
@@ -1019,9 +1047,42 @@ mod tests {
         atomic::{AtomicU32, Ordering},
         mpsc,
     };
+    use tracing::{Event, Subscriber, field::Visit};
+    use tracing_subscriber::{Layer, layer::Context, prelude::*};
 
     static TRACE_TEST_LOCK: Mutex<()> = Mutex::new(());
     static TRACED_SQLCIPHER_SETUP: Mutex<Vec<&'static str>> = Mutex::new(Vec::new());
+
+    #[derive(Clone)]
+    struct PlatformLimitLayer {
+        events: Arc<Mutex<Vec<Vec<String>>>>,
+    }
+
+    #[derive(Default)]
+    struct PlatformLimitVisitor {
+        fields: Vec<String>,
+    }
+
+    impl Visit for PlatformLimitVisitor {
+        fn record_debug(&mut self, field: &tracing::field::Field, value: &dyn fmt::Debug) {
+            self.fields.push(format!("{}={value:?}", field.name()));
+        }
+    }
+
+    impl<S> Layer<S> for PlatformLimitLayer
+    where
+        S: Subscriber,
+    {
+        fn on_event(&self, event: &Event<'_>, _context: Context<'_, S>) {
+            if event.metadata().target() == "storage_sqlite::connection"
+                && event.metadata().level() == &tracing::Level::WARN
+            {
+                let mut visitor = PlatformLimitVisitor::default();
+                event.record(&mut visitor);
+                self.events.lock().unwrap().push(visitor.fields);
+            }
+        }
+    }
 
     fn delete_journal_options() -> SqliteStorageOptions {
         SqliteStorageOptions {
@@ -1081,7 +1142,35 @@ mod tests {
     }
 
     #[test]
-    fn bundled_sqlcipher_contains_the_wal_reset_corruption_fix() {
+    fn android_memory_lock_platform_limit_is_reported_once() {
+        let events = Arc::new(Mutex::new(Vec::new()));
+        let subscriber = tracing_subscriber::registry().with(PlatformLimitLayer {
+            events: Arc::clone(&events),
+        });
+        let report_once = Once::new();
+
+        tracing::subscriber::with_default(subscriber, || {
+            report_sqlcipher_platform_limit_once(&report_once);
+            report_sqlcipher_platform_limit_once(&report_once);
+        });
+
+        let events = events.lock().unwrap();
+        assert_eq!(events.len(), 1);
+        assert!(
+            events[0]
+                .iter()
+                .any(|field| field == "method=\"configure_cipher_memory_security\"")
+        );
+        assert!(events[0].iter().any(|field| field == "configured=true"));
+        assert!(
+            events[0]
+                .iter()
+                .any(|field| field == "lock_posture=\"platform_limited\"")
+        );
+    }
+
+    #[test]
+    fn bundled_sqlcipher_contains_android_mlock_warning_fix() {
         fn version_tuple(value: &str) -> (u64, u64, u64) {
             let numeric = value.split_ascii_whitespace().next().unwrap_or(value);
             let mut components = numeric.split('.').map(|part| {
@@ -1108,8 +1197,8 @@ mod tests {
             "SQLite {sqlite_version} predates the WAL-reset corruption fix"
         );
         assert!(
-            version_tuple(&sqlcipher_version) >= (4, 14, 0),
-            "SQLCipher {sqlcipher_version} does not contain fixed SQLite 3.51.3"
+            version_tuple(&sqlcipher_version) >= (4, 17, 0),
+            "SQLCipher {sqlcipher_version} predates the Android mlock warning fix"
         );
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3232,8 +3232,7 @@ checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
 [[package]]
 name = "libsqlite3-sys"
 version = "0.38.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6c19a05435c21ac299d71b6a9c13db3e3f47c520517d58990a462a1397a61db"
+source = "git+https://github.com/rusqlite/rusqlite.git?rev=5ae7fdf83085c595fd54f977b3a56ccacabaf16b#5ae7fdf83085c595fd54f977b3a56ccacabaf16b"
 dependencies = [
  "cc",
  "openssl-sys",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -147,3 +147,7 @@ marmot-terminal-harness = { path = "integrations/terminal-harness" }
 # release contains it, then remove this patch and update the libcrux path.
 [patch.crates-io]
 hax-lib-macros = { git = "https://github.com/cryspen/hax.git", rev = "8f9cb576e58f6cc7e9ed249a7d4439b0f7ed0da7" }
+# libsqlite3-sys 0.38.2 still bundles SQLCipher 4.14.0, whose enhanced
+# memory-security allocator emits one WARN for every mlock failure. Pin the
+# merged 4.17.0 amalgamation until a crates.io release contains that lineage.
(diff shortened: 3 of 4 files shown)


crates/storage-sqlite/src/connection.rs
```diff
@@ -32,6 +34,26 @@ const BUSY_BACKOFF_BASE: Duration = Duration::from_millis(20);
...
+static SQLCIPHER_PLATFORM_LIMIT_REPORTED: Once = Once::new();
+
+#[cfg(any(target_os = "android", test))]
+fn report_sqlcipher_platform_limit_once(report_once: &Once) {
+    report_once.call_once(|| {
+        tracing::warn!(
+            target: "storage_sqlite::connection",
+            method = "configure_cipher_memory_security",
+            configured = true,
+            lock_posture = "platform_limited",
```
erskingardner, line 51: Non-blocking: `lock_posture = "platform_limited"` is asserted whenever cipher memory security is enabled, not observed from actual lock outcomes. The 64 KiB memlock cap is an AOSP platform constant, so this is an accurate summary for stock Android, but it will also report `platform_limited` on devices where `RLIMIT_MEMLOCK` has been raised and locking is actually effective. If that distinction ever matters operationally, `mincore`-based residency probing or reading the soft limit at startup could turn this into an observed signal — not worth it now given the once-per-process bound and fixed fields.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-02T15:54:47Z` |
| Commit reviewed | `dab983a141419a84a6076a3f8149c30dab989963` |

This is a well-scoped fix for #1522. I verified the central claims against upstream rather than taking the PR description on faith:

- The pinned rusqlite rev `5ae7fdf` is the merge of rusqlite PR #1870 ("Bump bundled SQLCipher to version 4.17.0") and sits on master's history, so the pin does deliver the 4.17.0 amalgamation.
- SQLCipher 4.17.0's `sqlcipher_mlock` logs failures at `SQLCIPHER_LOG_INFO`, below the `SQLCIPHER_LOG_WARN` default, so the per-allocation WARN storm is gone at the source while lock failures remain observable if the log level is raised.
- `sqlcipher_mem_security_on` is a process-global static in sqlcipher.c, so the new "process-wide" doc comment and README wording are accurate.
- MDK pins `cipher_compatibility = 4` before keying, so the 4.14.0 -> 4.17.0 lineage jump does not shift database-format defaults; existing databases remain readable. The test rename also preserves the older WAL-reset/SQLite-version guard.
- All encrypted open paths (account, in-memory, `open_hardened_sqlcipher` directory-cache/migration/rekey) route through `apply_cipher_pragmas`, so the once-per-process Android diagnostic covers every open the issue enumerates.
- The workspace sets `publish = false`, so the `[patch.crates-io]` pin is effective for everything that consumes this repo — no published-artifact gap.

Non-blocking notes:

- The Android diagnostic asserts `platform_limited` statically rather than observing lock outcomes. That is defensible — the 64 KiB memlock cap is an AOSP platform constant, and once-per-process bounds the cost — but it will also fire (and claim `platform_limited`) on Android variants where the budget has been raised. Inline comment has detail.
- The GrapheneOS device regression is still unchecked in the test plan; keep it open as the follow-up gate for closing #1522 rather than treating this merge as full closure.
- Note the SQLite baseline also jumps 3.51.3 -> 3.53.3 on every platform, not just Android. Nothing in the diff suggests a problem and the author ran the storage suite plus `just fast-ci`, but let the full CI matrix finish since several jobs were still pending at review time.
- The Cargo.toml comment documents the removal condition for the master-rev pin; consistent with the existing OpenMLS/hax pin pattern.

The change is small, surgical, privacy-safe per the observability rules (explicit target/method, aggregate fields only), and covered by regression tests on both the lineage and the bounded diagnostic.

crates/storage-sqlite/src/connection.rs
```diff
@@ -1081,7 +1142,35 @@ mod tests {
...
+        assert!(events[0].iter().any(|field| field == "configured=true"));
+        assert!(
+            events[0]
+                .iter()
+                .any(|field| field == "lock_posture=\"platform_limited\"")
+        );
+    }
+
+    #[test]
+    fn bundled_sqlcipher_contains_android_mlock_warning_fix() {
```
erskingardner, line 1173: Non-blocking nit: this test still guards the WAL-reset corruption fix (the SQLite >= 3.51.3 assertion and its `predates the WAL-reset corruption fix` message), but the rename drops that from the test name. Given the repo convention of keeping test names grep-able across layers, consider a name that covers both, e.g. `bundled_sqlcipher_contains_wal_reset_and_android_mlock_fixes`, so the WAL-reset guard remains discoverable by name.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-02T15:56:55Z` |
| Commit reviewed | `dab983a141419a84a6076a3f8149c30dab989963` |

*(Posted as a comment because GitHub does not allow approving your own PR; this is an approve-equivalent review.)*

Clean, surgical fix for #1522. Verified the important claims independently:

- The pinned rev `5ae7fdf8...` is the *merged* rusqlite PR #1870 ("Bump bundled SQLCipher to version 4.17.0"), on master, so the git pin cannot vanish with a feature branch. `libsqlite3-sys` is still version 0.38.1 at that rev, consistent with the lockfile, and the removal condition is documented in the `Cargo.toml` comment. The `[patch.crates-io]` approach follows the existing `hax-lib-macros` precedent.
- All encrypted open paths (account DB and the hardened directory-cache/migration/rekey path) funnel through `apply_cipher_pragmas`, so the once-per-process Android diagnostic covers every SQLCipher consumer; the unencrypted shared store stays out of scope, matching the issue.
- The diagnostic satisfies the observability invariant: explicit `target`/`method`, fixed aggregate fields only, no identifiers or payload data. Gating it behind `Once` and placing it after the pragma succeeds are both correct, and the test uses its own `Once` plus a thread-local subscriber, so no cross-test interference.
- Test plan evidence is credible: the version regression test asserts the new floor (>= 4.17.0), and CI format/clippy/audit gates pass.

Two non-blocking notes:
1. The device-regression checkbox (GrapheneOS workload confirmation) is still open — that is inherently post-merge verification, but it is the one acceptance criterion from #1522 not yet evidenced. Worth tracking before closing the issue.
2. One inline nit on the renamed version test, which still guards the WAL-reset fix under a name that no longer mentions it.

Issue #1529, opened 2026-08-21
Bound bad-locator failover and reuse safe Blossom connections across download backlogs


Datawav:
## Observed release-blocking behavior

A user reports unexpectedly slow file downloads, especially when many attachments arrive while the device is offline and are fetched after reconnect. The user classifies the complete file-handling experience as a P0 release blocker. This is production/device evidence; it has not been reproduced locally.

At MDK `master` commit [`baae0be21bc44cebc89c9a13f2cb50524307c255`](https://github.com/marmot-protocol/mdk/tree/baae0be21bc44cebc89c9a13f2cb50524307c255), the steady-state download path contains two source-confirmed tail-latency and repeated-setup costs that compound across a backlog:

- Candidate locators are attempted strictly one after another, and each candidate can occupy the bounded transfer timeout before the next is tried ([source](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/crates/marmot-app/src/media/mod.rs#L767-L803), [timeouts](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/crates/marmot-app/src/media/blossom.rs#L24-L34)). One unhealthy first locator can dominate perceived latency even when a later locator is healthy.
- Every fetch hop resolves and vets the host, constructs a fresh pinned `reqwest::Client`, and then performs the GET ([source](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/crates/marmot-app/src/media/blossom.rs#L285-L344), [source](https://github.com/marmot-protocol/mdk/blob/baae0be21bc44cebc89c9a13f2cb50524307c255/crates/marmot-app/src/media/blossom.rs#L456-L494)). Across many same-origin attachments this prevents ordinary client/connection reuse and repeats DNS/TLS setup.

[...]

Issue #1437, opened 2026-08-14
Expose resumable encrypted-media downloads with progress through UniFFI


mubarakcoded:
## Implementation complete — 2026-09-19

Parent #1742. MDK acquisition and native transfer work is merged: storage #1898, background acquisition #1907, protected partial resume #1909, native local-byte access #1915, and **progress/controls/default enablement #1917 (`dc7ddf75`)**.

- Swift/Kotlin and C expose bounded snapshots/live progress (up to 64 selected slots, coalesced to at most 4 Hz), cancellation, retry, remove, download-again and per-account policy controls.
- Cancellation persists until explicit retry and preserves ready bytes; valid partial ciphertext retains its 24-hour expiry. Source/account/store fences and full authentication precede access to plaintext.
- Automatic acquisition is enabled for eligible accepted conversations: 2 GiB/account, 256 MiB free-disk reserve plus write headroom, one automatic transfer, 64 MiB automatic cap. Pending invitations defer automatic attachments. Explicit downloads retain their separate size limit. Frozen runtimes skip acquisition.
- Disabling automatic work stops active automatic transfers and pauses queued automatic work without removing ready bytes or disabling explicit requests. Re-enabling does not undo cancellation/removal. Size-policy increases can re-admit size-blocked work without retrying integrity failures.
- Recent incoming attachments precede backfill; explicit work takes priority. Automatic transfers have a two-minute whole-transfer deadline and a 30-second body-idle deadline. The existing durable worker handles restart, retry and resume; no second queue or host media store is required.
- Existing `download_media` remains compatible. Applications must opt into the new control/progress/local-read methods to use them.

Final #1917 validation: 112 attachment tests and `just fast-ci`; generated Swift/Kotlin round trips and debug C smoke passed during the slice. Audit/release preflight continues in #1918, including the optimized macOS C stripping fix.

**C9 remains open in #1742:** publish matching artifacts, adopt them in iOS/Android, verify restart/offline behavior and controls on devices, and measure large-media chunk access/assembly. Closing this implementation issue is not a claim that those clients already use the released APIs.

<details>
<summary>Original problem, acceptance criteria and historical enablement gate</summary>

## Problem

[...]

PR #1635, opened 2026-09-02
Reuse vetted Blossom clients and bound locator startup

erskingardner:
## Summary

- add a per-account, exact-origin Blossom download transport that single-flights DNS resolution and client creation across a backlog
- reuse pinned reqwest clients for 60 seconds, then resolve, vet, and pin a fresh address set; keep the cache bounded to 32 origins
- bound each ordered locator startup to 10 seconds so DNS, connection, headers, or first-byte stalls reach the next candidate promptly
- require ciphertext hash validity before a candidate wins, allowing corrupt first locators to fall through
- preserve redirect revalidation, proxy isolation, loopback policy, sequential candidate order, and the existing four-operation account cap

## Tests

- cargo test -p marmot-app media::tests::
- cargo test -p marmot-app media_http
- cargo clippy -p marmot-app --all-targets -- -D warnings
- just fast-ci

The full marmot-app test run passed 874 tests but reported six unrelated epoch-backfill timing failures. One representative failure, deferred_primary_epoch_backfill_rotates_behind_queued_older_operation, reproduces in isolation on an unchanged origin/master worktree.

## Scope

This draft implements the surgical reuse and bounded-failover correction discussed for #1529. The issue broader privacy-safe phase telemetry, repository benchmark, and before/after p95 evidence remain outside this patch, so this PR intentionally does not auto-close the issue.

Addresses #1529

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability for media and group-image downloads, including better handling of stalled or slow connections.
  - Added fallback handling when downloaded encrypted media is corrupted or fails integrity verification.
  - Improved connection reuse for repeated media downloads from the same source.
  - Strengthened network address validation during media downloads to improve safety and resilience.
  - Improved handling of concurrent requests and expired connection information during media retrieval.

### DIFF
--- a/crates/marmot-app/src/media/blossom.rs
+++ b/crates/marmot-app/src/media/blossom.rs
@@ -30,7 +30,8 @@ const MEDIA_HTTP_TOTAL_TIMEOUT: Duration = Duration::from_secs(60);
 const MEDIA_BLOB_TRANSFER_TIMEOUT: Duration = Duration::from_secs(15 * 60);
 /// A candidate that cannot resolve, connect, return headers, and yield its first
 /// body bytes within this bound gives the next ordered locator a chance. The
-/// longer transfer deadline and read-idle timeout still govern an active body.
+/// longer transfer deadline and read-idle timeout still govern an active body,
+/// so a peer that continuously trickles bytes can occupy the transfer deadline.
 const BLOSSOM_CANDIDATE_STARTUP_TIMEOUT: Duration = Duration::from_secs(10);
 /// Reusing a pinned client inside this lease amortizes DNS and TLS setup without
 /// turning one resolution result into a process-lifetime routing decision.
@@ -90,7 +91,8 @@ struct BlossomHttpTransportInner {
 ///
 /// Every cached client is bound to one exact origin and one vetted address set.
 /// Expiry creates a fresh client generation, forcing DNS resolution, vetting,
-/// and pinning again before any later request can use the origin.
+/// and pinning again before any later request can use the origin. A rehomed
+/// origin can therefore retain a failing stale pin only until the lease ends.
 #[derive(Clone)]
 pub(crate) struct BlossomHttpTransport {
     inner: Arc<BlossomHttpTransportInner>,
@@ -204,6 +206,8 @@ impl BlossomHttpTransport {
     }
 
     pub(super) fn with_loopback_disabled(&self) -> Self {
+        // The stricter view may share the pool because client_for_url validates
+        // its own loopback policy before consulting a cached generation.
         Self {
             inner: self.inner.clone(),
             allow_loopback_http: false,
--- a/crates/marmot-app/src/media/blossom.rs
+++ b/crates/marmot-app/src/media/blossom.rs
@@ -1,7 +1,10 @@
+use std::collections::HashMap;
 use std::net::{IpAddr, SocketAddr};
-use std::time::Duration;
+use std::sync::{Arc, Mutex as StdMutex};
+use std::time::{Duration, Instant};
 
 use bytes::Bytes;
+use futures::future::BoxFuture;
 use nostr::base64::Engine as _;
 use nostr::base64::engine::general_purpose::URL_SAFE_NO_PAD as BASE64_URL_SAFE_NO_PAD;
 use nostr::{EventBuilder, JsonUtil, Kind, NostrSigner, Tag, Timestamp as NostrTimestamp};
@@ -25,6 +28,14 @@ const MEDIA_HTTP_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
 const MEDIA_HTTP_READ_TIMEOUT: Duration = Duration::from_secs(15);
 const MEDIA_HTTP_TOTAL_TIMEOUT: Duration = Duration::from_secs(60);
 const MEDIA_BLOB_TRANSFER_TIMEOUT: Duration = Duration::from_secs(15 * 60);
+/// A candidate that cannot resolve, connect, return headers, and yield its first
+/// body bytes within this bound gives the next ordered locator a chance. The
+/// longer transfer deadline and read-idle timeout still govern an active body.
+const BLOSSOM_CANDIDATE_STARTUP_TIMEOUT: Duration = Duration::from_secs(10);
+/// Reusing a pinned client inside this lease amortizes DNS and TLS setup without
+/// turning one resolution result into a process-lifetime routing decision.
+const BLOSSOM_ADDRESS_LEASE: Duration = Duration::from_secs(60);
+const BLOSSOM_ORIGIN_CACHE_LIMIT: usize = 32;
 const BLOSSOM_REDIRECT_LIMIT: usize = 5;
 /// Largest encrypted media blob this implementation will upload or download.
 ///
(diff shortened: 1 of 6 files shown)


crates/marmot-app/src/media/mod.rs
```diff
@@ -686,23 +690,42 @@ fn upload_error_summary(err: &AppError) -> String {
...
-    let encrypted = fetch_encrypted_media_blob(
+    let encrypted = fetch_encrypted_media_blob_with_transport(
         &reference,
         fallback_endpoints,
         allowed_locator_kinds,
-        allow_loopback_blob_endpoints,
+        transport,
     )
     .await?;
     if !encrypted_media_hash_matches(&encrypted, &reference.ciphertext_sha256) {
```
erskingardner: Now that `fetch_encrypted_media_blob_with_transport` enforces the hash before a candidate can win (line 839 below), this re-hashes the full winning blob on every successful download — a second full SHA-256 pass, nontrivial at the 512 MiB cap. Either drop it or keep it with a one-line comment marking it as deliberate defense-in-depth so it doesn't read as an oversight. Non-blocking.

crates/marmot-app/src/media/mod.rs
```diff
@@ -795,8 +835,13 @@ async fn fetch_encrypted_media_blob(
                 continue;
             }
         }
-        match fetch_blossom_blob(&candidate, allow_loopback_blob_endpoints).await {
-            Ok(bytes) => return Ok(bytes),
+        match fetch_blossom_blob_with_transport(&candidate, transport).await {
+            Ok(bytes) if encrypted_media_hash_matches(&bytes, &expected_hash) => return Ok(bytes),
```
erskingardner, line 834: Requiring integrity before a candidate wins is the right rule. One cost worth noting: a corrupt-but-responsive locator can still consume the full transfer budget (up to 512 MiB / 15 min) before this check rejects it, so a corrupt first locator now adds latency rather than failing fast. Streaming verification (hash while reading, abort on mismatch) would bound that earlier — reasonable to defer, e.g. alongside #1437's partial-transfer work. Non-blocking.

crates/marmot-app/src/media/blossom.rs
```diff
@@ -25,6 +28,14 @@ const MEDIA_HTTP_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
 const MEDIA_HTTP_READ_TIMEOUT: Duration = Duration::from_secs(15);
 const MEDIA_HTTP_TOTAL_TIMEOUT: Duration = Duration::from_secs(60);
 const MEDIA_BLOB_TRANSFER_TIMEOUT: Duration = Duration::from_secs(15 * 60);
+/// A candidate that cannot resolve, connect, return headers, and yield its first
+/// body bytes within this bound gives the next ordered locator a chance. The
+/// longer transfer deadline and read-idle timeout still govern an active body.
+const BLOSSOM_CANDIDATE_STARTUP_TIMEOUT: Duration = Duration::from_secs(10);
```
erskingardner, line 35: The 10 s bound covers resolve/connect/headers/first byte, but a locator that trickles one byte every <15 s can still hold a candidate for the full 15 min transfer deadline, so the healthy-locator bound doesn't hold for that adversarial shape. Worth one sentence in the doc comment (or a follow-up under #1437's progress/cancel semantics) so future readers don't assume the bound covers the whole body. Non-blocking.

crates/marmot-app/src/media/tests.rs
```diff
@@ -1454,6 +1528,173 @@ fn blossom_redirect_validation_rejects_cross_scheme_private_ip_and_cross_domain(
...
+}
+
+#[tokio::test]
+async fn stalled_first_locator_reaches_healthy_fallback_within_startup_bound() {
+    let body = b"healthy ciphertext";
+    let (stalled_url, stalled_server) = spawn_stalled_http_server().await;
+    let healthy_url = spawn_http_response(http_ok_response(body));
+    let reference = blob_reference_for_servers(body, &[stalled_url, healthy_url]);
+    let transport =
+        BlossomHttpTransport::for_test(true, Duration::from_secs(60), Duration::from_millis(100));
```
erskingardner: The healthy fallback candidate also gets only this 100 ms startup budget for connect + headers, which is a tight margin on a loaded CI runner. Consider 250–500 ms; the 2 s outer assertion keeps the test meaningful either way. Non-blocking.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-02T16:48:29Z` |
| Commit reviewed | `d395699eae7e4e5489a6c4c51c326874b93c0998` |

**Premise**: solid. #1529 is a P0 user report with source-confirmed costs (serial bad-locator failover under a 15-minute transfer timeout, and per-hop DNS/TLS/client setup). This PR addresses both with a scoped, surgical change.

**What I verified:**

- Policy preservation: URL validation, loopback gating, redirect revalidation, `no_proxy`, address pinning, and the strict (non-loopback) vetting path are unchanged. The 60 s lease pins a previously vetted address set rather than skipping policy; rebinding is re-resolved and re-vetted at expiry.
- Cross-variant cache safety: `with_loopback_disabled()` shares the origin cache with the media transport, but that is provably safe — the stricter variant can never request a loopback origin (validation rejects it first), and public origins never receive loopback pins because the loopback vetting relaxation only applies when the URL host itself is loopback.
- Single-flight correctness: the generation mutex serializes resolve/vet/build per origin, and a caller timeout drops the `client_for_url` future, which releases the guard and cancels the in-flight resolution — a hung DNS lookup cannot wedge an origin.
- The origin cache is bounded (32, LRU) and the transport is per-`AppClient` (per account); the four-operation media HTTP cap is untouched.
- Startup bound: 10 s per candidate covering resolve + connect + headers + first byte, fresh per candidate; body reads stay under the existing read/total timeouts. The stalled-first-locator and corrupt-first-locator tests cover the two failure shapes from the issue.

**Validation gap:** remote Rust tests / clippy / check were still pending at review time. The author reports local `just fast-ci` plus targeted `marmot-app` runs passing, with unrelated epoch-backfill flakes reproducing on clean master.

Non-blocking notes are inline. Two follow-ups worth tracking on #1529: uploads still build a fresh pinned client (and DNS resolve) per operation, so send-side backlogs keep the old setup cost; and the issue's benchmark, phase telemetry, and before/after p95 evidence remain open — good call not auto-closing the issue here.

crates/marmot-app/src/media/blossom.rs
```diff
@@ -33,6 +44,186 @@ const BLOSSOM_REDIRECT_LIMIT: usize = 5;
...
+            slot.last_used = now;
+            slot.generation.clone()
+        };
+
+        let mut generation = generation.lock().await;
+        let now = Instant::now();
+        if let Some(cached) = generation.as_ref()
+            && cached.expires_at > now
+        {
+            return Ok(cached.client.clone());
```
erskingardner, line 193: Non-blocking: a cache hit reuses the pinned address set even if that set has started failing (e.g. the origin is rehomed within the 60s lease). Every same-origin attempt inside the window will fail against the stale pin until the lease expires and self-heals, whereas the old per-request resolve recovered on the very next attempt. Consider invalidating the cached generation on connect-level request failure, or documenting the bounded stale-failure window on the struct doc so the availability tradeoff is explicit.

crates/marmot-app/src/media/blossom.rs
```diff
@@ -33,6 +44,186 @@ const BLOSSOM_REDIRECT_LIMIT: usize = 5;
...
+        let pin = resolve_media_host_with(url, allow_loopback, &self.inner.resolver).await?;
+        let client = build_pinned_media_http_client(pin)?;
+        *generation = Some(CachedMediaClient {
+            client: client.clone(),
+            expires_at: now + self.address_lease,
+        });
+        Ok(client)
+    }
+
+    pub(super) fn with_loopback_disabled(&self) -> Self {
```
erskingardner, line 208: Nit: sharing `inner` here means one origin cache now serves both loopback-allowed media downloads and never-loopback group-image fetches. That stays safe only because `client_for_url` validates with the caller's `allow_loopback_http` *before* consulting the cache — worth a short comment here documenting that invariant so a future refactor doesn't move validation behind the cache hit.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-02T17:04:28Z` |
| Commit reviewed | `d395699eae7e4e5489a6c4c51c326874b93c0998` |

Surgical, well-scoped correction for the in-scope parts of #1529. The per-account `BlossomHttpTransport` is created once per account open (`local_client_with_relay_plane_and_hydration`), single-flights resolution per origin through the generation mutex, and bounds the cache at 32 origins with LRU eviction. Security invariants are preserved: per-call `validate_blossom_fetch_url` runs before any cache lookup, redirect targets are still revalidated per hop, proxy isolation (`no_proxy`) and address pinning are unchanged, loopback gating is applied per caller, and the 60s lease forces periodic re-resolution and re-vetting (the rebinding regression test covers this). The 10s startup bound sits consistently below the existing 15s read-idle timeout, and the per-candidate ciphertext-hash gate is a strict improvement — a corrupt first locator now falls through instead of poisoning the whole download. Tests cover connection reuse, single-flight, per-account isolation, lease expiry/re-vetting, stalled-locator failover, and corrupt-locator fall-through. CI Rust shards are green; the PR body transparently scopes out the benchmark/phase-telemetry acceptance items and deliberately leaves #1529 open. Two non-blocking notes inline. (Posted as a comment rather than an approval because GitHub does not allow reviewing your own PR.)

crates/marmot-app/src/media/mod.rs
erskingardner: Agreed. Streaming verification belongs with the partial-transfer and progress work in #1437; this patch keeps the existing bounded whole-body read and validates before any candidate can win.

crates/marmot-app/src/media/blossom.rs
erskingardner: Documented in 334cbdea that the 10 second bound covers startup only and that a continuously trickling body remains governed by the existing read-idle and 15 minute transfer deadlines.

crates/marmot-app/src/media/mod.rs
erskingardner: Removed the redundant outer hash pass in 334cbdea. The candidate fetch helper is private and now guarantees ciphertext integrity before returning the winning bytes.

crates/marmot-app/src/media/tests.rs
erskingardner: Raised the test-only startup budget from 100 ms to 500 ms in 334cbdea. The outer 2 second deadline still proves prompt fallback while leaving reasonable CI scheduling margin.

crates/marmot-app/src/media/blossom.rs
erskingardner: Documented the bounded stale-pin availability tradeoff in 334cbdea. I kept failure-triggered invalidation out of this surgical change because it would add new retry and error-classification semantics; expiry still forces fresh resolution and vetting within 60 seconds.

crates/marmot-app/src/media/blossom.rs
erskingardner: Added the cache-sharing invariant comment in 334cbdea: the stricter view is safe because each client_for_url call applies its own loopback policy before any cache lookup.

Issue #1014, opened 2026-07-21
Security review L20: forensic audit log — no fsync, unbounded growth, no tamper-evidence, rotation destroys history


erskingardner:
## Summary

(a) Crash durability: per-line flush reaches the page cache but no fsync/fdatasync is ever issued, so a power loss drops the most recent forensic lines — precisely the divergence window one captures forensics for. (b) Growth: no size cap or automatic rotation; in `FullData` mode (decrypted content) the file grows without bound until manual `rotate()` — disk-fill DoS on always-on agents. (c) Tamper evidence: append-only is by convention only — `seq` is per-session, no hash chaining, and `rotate()`/`set_data_mode` deliberately discard the current file (deletes history through the public API). (d) Relay URLs are logged in both data modes, and raw group/message ids are logged even in obfuscated mode — a wider exemption from the workspace logging rules than "hashed/truncated identifiers" suggests.

## Location

`crates/marmot-forensics/src/audit.rs:1367-1414` (`record`: `writeln!` + `BufWriter::flush` only), `:1496-1531` (rotation), `:1116-1124` (`rotate` contract)

## Impact

An incident investigation loses exactly the tail of the log (crash window) or the whole history (rotation); a same-user attacker can rewrite history undetectably. Documented design tradeoffs, but the "append-only JSONL forensic audit" framing oversells integrity.

## Suggested remediation

Optional fsync policy (every N ms/lines or on `EpochStateChanged`-class rows), size-bounded auto-rotation with retained segments, and explicit documentation that the log provides no tamper evidence.

---
Found in the 2026-07-20 security review (`docs/security-review-2026-07-20.md`, finding L20). Status: **confirmed.**

Datawav:
Fresh field evidence adds a distinct performance/failure mode to this canonical audit-log issue: the tracker currently re-uploads every complete, growing audit file on each send/sync trigger rather than an acknowledged suffix or immutable rotated segment.

Current source:

- `crates/marmot-app/src/runtime/audit_tracker.rs:205-232` enumerates all files and POSTs every one.
- `crates/marmot-app/src/audit_log.rs:337-394` opens each file at byte zero and streams its complete current length.
- Ordinary `send_app_event` schedules an upload at `runtime/commands.rs:914-917`; receive/startup/convergence paths also schedule updates.
[...]

PR #1642, opened 2026-09-03
fix(app,forensics): upload audit segments once instead of re-posting whole files

### LINKED ISSUE #1181: marmot-app: audit tracker reuploads complete growing JSONL files on every activity trigger
## Summary

The audit tracker re-uploads every complete, growing JSONL file from byte zero on each send/sync-triggered update. Server-side deduplication prevents duplicate records from being stored as new events, but it does not prevent repeated device disk reads, upload bandwidth, request processing, or server line parsing.

Fresh Goggles source metadata shows this is already extreme in real opt-in cohorts: **76,978,886 of 77,142,487 valid uploaded lines were declared duplicates** after deduplicating source-upload records—99.788% duplicate line processing and approximately **471.5x aggregate amplification**.

## Root cause

- `crates/marmot-app/src/runtime/audit_tracker.rs:205-232` enumerates all audit files and posts every one on each tracker update.
- `crates/marmot-app/src/audit_log.rs:337-394` opens each file at byte zero and streams its complete current length.
- Ordinary `send_app_event` schedules an upload at `runtime/commands.rs:914-917`; startup, receive, and convergence paths add more triggers.
- There is no acknowledged byte offset, immutable-segment identity, or content checkpoint.
- `AUDIT_LOG_UPLOAD_MAX_BYTES` is 64 MiB, but the active recorder has no automatic rotation. Once a file crosses the ceiling, every automatic upload attempt fails until external intervention.

## Field validation

Scope: **212/212 successfully exported groups in the complete recursive graph reachable from the two configured seeds**. This is not a global Goggles census.

After deduplicating overlapping group-export source records by upload/file identity, the graph contained:

- 8,422 unique source uploads;
- 77,142,487 valid uploaded event lines;
- 76,978,886 server-declared duplicate lines;
- 163,601 non-duplicate lines by subtraction;
- 99.788% duplicates;
- ~471.5x valid-line processing per non-duplicate line.

Three independently refreshed high-signal exports showed:

- 98.301% duplicates / ~58.9x amplification;
- 92.709% duplicates / ~13.7x amplification;
- 99.928% duplicates / ~1,383x amplification.

These are source-upload counters, not deduplicated group-event counts. Overlapping recursive exports were explicitly deduplicated before the aggregate.

## Impact

For audit-enabled devices, work grows with the cumulative recorder size rather than new diagnostic data:

- repeated full-file disk reads and serialization/upload work;
- repeated network transfer and server-side line parsing;
- increasing device disk, network, and server load around ordinary activity-triggered updates;
- quadratic cumulative transfer as an active file grows;
- permanent automatic-upload failure after the 64 MiB ceiling.

This is instrumentation-induced performance cost, not an MLS consensus failure. Synchronous per-row recorder flush cost is adjacent but separable; this issue owns upload volume and checkpointing.

## Required behavior

Use a crash-safe incremental upload contract:

- rotate the active recorder into immutable bounded segments, or persist server-acknowledged byte ranges/content identities;
- never delete/truncate active or immutable data before acknowledgement;
- retry failed data idempotently without resending acknowledged prefixes;
- preserve the crash tail according to the selected recorder durability policy;
- recover automatically before any segment reaches the 64 MiB request ceiling;
- coalesce frequent send/sync triggers while an upload is already scheduled or running;
- bound retained disk usage without silently losing unacknowledged forensic data.

## Regression coverage

- Append N rows, trigger upload repeatedly without new rows: acknowledged prefixes are not re-read/re-uploaded.
- Append a suffix: only the suffix or one new immutable segment is transferred.
- Restart before and after acknowledgement: no loss, and duplicate transfer remains bounded.
- Upload failure: unacknowledged data retries; acknowledged data does not.
- Recorder exceeds one segment: automatic rotation/upload continues below the per-request limit.
- Concurrent trigger bursts coalesce and do not launch duplicate uploads.

## Related

- #1014 owns recorder durability, unbounded growth, and automatic rotation. The remediation must coordinate with this incremental-transfer contract, but rotation alone does not prevent repeated prefixes from being uploaded on every trigger.
- #1176 owns foreground deferred-peel scheduling, not audit transfer amplification.

jgmontoya:
Fixes #1181: the audit tracker re-uploaded every growing JSONL file from byte zero on each send/sync trigger — field-measured at 99.8% duplicate lines and ~471x amplification — and once a file crossed the 64 MiB request ceiling, automatic upload failed permanently, silently dropping that device out of the forensic fleet.

Client-only fix, no Goggles changes. The server already provides the needed acknowledgment semantics: line dedupe is content-keyed (filename-agnostic) and byte-identical re-uploads short-circuit on file_sha256 without parsing, so for an immutable file one HTTP 200 is a durable ack.

1. Segment rotation (marmot-forensics): JsonlRecorder seals the active file at 1 MiB by renaming it into an immutable audit-<engine>-v3-segNNNNNN.jsonl sibling and reopening. Transparent: seq, session id, and health counters carry across, no fabricated recorder_started row, and segments plus active concatenate byte-identical to the unrotated log (pinned by a literal concat test). Crash mid-roll leaves a complete segment; a failed reopen renames the segment back (compensation, tested); a transient roll failure latches only for the directory-unwritable episode, not the recorder's lifetime (regression-tested). The threshold was chosen from measured row sizes: at 1 MiB the steady-state residual is ~45x vs 176x at 4 MiB, with 64x margin under the request ceiling.
2. Upload-once checkpoint (marmot-app): a per-account fs-private sidecar records each successfully uploaded file as (name, size, mtime). Matches are never re-read or re-posted; any mismatch re-uploads and the server dedupe absorbs it. The checkpoint is deliberately losable — a torn or missing sidecar costs one cheap re-upload per file. Files are append-only, so a recorded pair can never falsely match future state.
3. Trigger coalescing was already correct (size-1 queue); it is now a documented contract with a discriminating regression test.
4. Upgrade path for the cliff: an already-oversized active file is rolled aside at open so new uploads flow immediately; an oversized immutable file is warned once (aggregate counts only), recorded as too-large, and skipped without blocking other files. The doc now states honestly that no app API transfers such a file; retention is #1014's contract.

Deliberately out of scope, recorded as follow-ups: retention/deletion/disk bounding, fsync and tamper evidence (#1014); the remaining ~45x active-file residual (the 1.0x design — upload only sealed segments — trades tail freshness and needs a product call); the public post_audit_log_tracker_update host API bypassing the coalescing queue; a pre-existing Content-Length race on files appended mid-upload, now confined to the active segment.

## Summary by CodeRabbit

* **New Features**
  * Audit logs now automatically rotate into immutable 1 MiB segments while continuing to record new events.
  * Incremental uploads use persistent checkpoints to avoid retransferring unchanged files and retry failed transfers.
  * Oversized files are reported without blocking other uploads.
  * Upload processing remains queued and coalesces bursts to prevent concurrent runs.

* **Bug Fixes**
  * Improved recovery after interrupted or failed segment rotation.
  * Preserved audit-log continuity across segment rolls and recorder reopenings.

* **Documentation**
  * Clarified segmentation, upload checkpointing, retries, and ordering limitations.

### DIFF
--- a/crates/incident-replay/src/classify.rs
+++ b/crates/incident-replay/src/classify.rs
@@ -344,9 +344,13 @@ impl HaltLifecycle<'_> {
     /// Both rows also come from a single engine — one clock, one recorder file —
     /// so none of rule 6's cross-device [`CATCH_UP_GRACE_MS`] skew allowance
     /// applies and a strict comparison is sound. (`seq` must never be substituted
-    /// for the clock: it restarts at 0 on every recorder open *within the same
-    /// file*, so it does not order rows across a restart — the exact boundary a
-    /// durable halt survives. See `docs/marmot-architecture/audit-logging.md`.)
+    /// for the clock. It restarts at 0 on every recorder open, so it does not
+    /// order rows across a restart — the exact boundary a durable halt
+    /// survives — and since mdk#1181 it does not identify a file either: a
+    /// segment roll carries `seq` across into an `audit-*-seg<NNNNNN>.jsonl`
+    /// sibling, so one session's rows span several source files and a file can
+    /// start at a nonzero `seq`. Neither direction is orderable by `seq`. See
+    /// `docs/marmot-architecture/audit-logging.md`.)
     ///
     /// Absence fails closed in both directions: an unrepaired halt keeps its
     /// quarantine, and so does a halt whose evidence cannot be ordered — arming
--- a/crates/marmot-app/AGENTS.md
+++ b/crates/marmot-app/AGENTS.md
@@ -52,6 +52,14 @@ App runtime bridge for the first real Marmot app surfaces.
   Do not add a trigger path that bypasses the size-1 coalescing queue in `runtime/audit_tracker.rs`, and do not let one
   oversized or failing file block the files behind it. Retention/deletion of sealed segments is mdk#1014, not this
   contract.
+- Keep the upload checkpoint's cost proportional to the account's *live* audit files, not to its history. `retain_present`
+  prunes entries for files that are gone, so the sidecar is O(live `audit-*.jsonl` files) — but nothing deletes sealed
+  segments (mdk#1014), so that bound grows with cumulative audit volume, and each tracker run stats every file and
+  rewrites the whole sidecar. Do not add a per-upload or per-line entry, and do not treat the sidecar as durable state:
+  losing it costs one repeat transfer per file and nothing else.
+- A file above `AUDIT_LOG_UPLOAD_MAX_BYTES` has no app-level transfer path. `post_audit_log_file` enforces the same
+  ceiling, so the tracker's skip is the whole story: the file stays on disk for manual handling. Do not document an
+  escape hatch that does not exist, and do not widen the ceiling to invent one.
 - Keep SQLCipher key derivation, per-database salt persistence, legacy-key migration/recovery, and the in-process
   v2-open probe-verdict cache (mdk#1439) in `src/sqlcipher.rs`. The verdict cache is advisory only: a durable
   migration marker or a missing verdict always re-runs the recovery probe, so the mdk#568 crash windows keep their
--- a/crates/marmot-app/AGENTS.md
+++ b/crates/marmot-app/AGENTS.md
@@ -44,8 +44,14 @@ App runtime bridge for the first real Marmot app surfaces.
   messages, app events, push registrations, telemetry/audit settings) in `src/conversions.rs`. They hold no `MarmotApp`
   state.
 - Keep the forensic audit-log feature in `src/audit_log.rs`: the `AuditLog*` DTOs, salted-hash identity derivation,
-  the upload client, and the `MarmotApp` methods for audit settings, recorder open/build, file enumeration, path
-  validation/resolution/removal, and HTTP upload. Audit-log unit tests live in its own `#[cfg(test)] mod tests`.
+  the upload client, the per-account upload checkpoint (`audit-upload-checkpoint.json`), and the `MarmotApp` methods for
+  audit settings, recorder open/build, file enumeration, path validation/resolution/removal, and HTTP upload. Audit-log
+  unit tests live in its own `#[cfg(test)] mod tests`.
+- Keep audit uploads incremental (mdk#1181). An audit file whose size and mtime still match its checkpoint entry is
+  never re-read or re-posted; only the growing active file re-transfers, bounded by the recorder's segment threshold.
+  Do not add a trigger path that bypasses the size-1 coalescing queue in `runtime/audit_tracker.rs`, and do not let one
+  oversized or failing file block the files behind it. Retention/deletion of sealed segments is mdk#1014, not this
+  contract.
 - Keep SQLCipher key derivation, per-database salt persistence, legacy-key migration/recovery, and the in-process
   v2-open probe-verdict cache (mdk#1439) in `src/sqlcipher.rs`. The verdict cache is advisory only: a durable
   migration marker or a missing verdict always re-runs the recovery probe, so the mdk#568 crash windows keep their
--- a/crates/marmot-app/Cargo.toml
+++ b/crates/marmot-app/Cargo.toml
@@ -57,6 +57,7 @@ zeroize.workspace = true
 
 [dev-dependencies]
 keyring-core.workspace = true
+marmot-forensics.workspace = true
 nostr-relay-builder.workspace = true
 tempfile.workspace = true
 tokio = { workspace = true, features = ["test-util"] }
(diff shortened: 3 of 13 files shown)


Datawav (commented):
## Automated review

Four actionable minor findings.

1. Minor — Request capture order is racy
   Exact changed-line evidence: `crates/marmot-app/tests/audit_logs.rs:667-670` executes `write_http_response(...).await`, `stream.shutdown().await`, and decrements `in_flight` before `requests.lock().unwrap().push(request)`.
   Triggering state: One handler yields after replying while the client opens another connection.
   Consequence: The later request can be recorded first, making exact-order assertions in `appending_a_suffix_transfers_only_the_changed_file` and `failed_uploads_retry_while_acknowledged_files_do_not` flaky.
   Confidence: High.
   Smallest sound correction: Push the captured request immediately after `read_captured_request` succeeds, before writing the response.

2. Minor — Rotation test mishandles root and cleanup
   Exact changed-line evidence: `crates/marmot-forensics/src/audit/tests.rs:1484-1489` chmods `dir.path()` to `0o500`, opens `JsonlRecorder`, then asserts `segment_paths(&path).is_empty()` before restoring `0o700`.
   Triggering state: Tests run as root, or the assertion fires while the temporary-directory root remains read-only.
   Consequence: Root bypasses the intended write denial, producing a false-negative test result; an assertion panic can also impede `TempDir` cleanup.
   Confidence: High.
   Smallest sound correction: Chmod a dedicated child directory, probe whether writes still succeed and return early for root-bypass behavior, then restore permissions before asserting.

3. Minor — Over-ceiling files are hidden from the returned result
   Exact changed-line evidence: In `crates/marmot-app/src/runtime/audit_tracker.rs`, hunk `@@ -213,35 +221,142`, the new path counts `too_large_recorded`/`too_large_known`, but still returns `AuditLogTrackerUpdateResult { enabled: true, uploaded, skipped_reason: None }`.
   Triggering state: An account has only a file larger than `AUDIT_LOG_UPLOAD_MAX_BYTES`.
   Consequence: Callers receive `uploaded: []` and no skip reason, indistinguishable from “nothing to upload,” although that file cannot be transferred through the app API.
   Confidence: High.
   Smallest sound correction: Add an over-ceiling count to the result, or return an explicit skip reason for this state.

4. Minor — Oversize warning semantics do not match the implemented checkpoint identity
   Exact changed-line evidence: In the same tracker hunk, `if failed > 0 || too_large_recorded > 0` emits “completed forensic audit log tracker update with file upload failures”; the oversize checkpoint is recorded with `checkpoint.acknowledge(file, file.size_bytes, AuditUploadOutcome::TooLargeToUpload)`, whose identity includes size/mtime.
   Triggering state: Rotation has latched off and an already-oversized active file continues growing.
   Consequence: A no-upload-error run is mislabeled as having upload errors, and each size/mtime change re-records, re-warns, and rewrites the checkpoint despite the stated report-once behavior.
   Confidence: High.
   Smallest sound correction: Use condition-specific warning text and recognize an existing `TooLargeToUpload` entry by filename until the oversize episode ends.

Audit evidence:
- Exact head: `466ac499311421b96444d5a1d0f1fd5e9b535da0`
- Changed files examined include `crates/marmot-app/src/runtime/audit_tracker.rs`, `crates/marmot-app/tests/audit_logs.rs`, `crates/marmot-forensics/src/audit.rs`, `crates/marmot-forensics/src/audit/tests.rs`, and `docs/marmot-architecture/audit-logging.md`.
- Coverage is complete for all 13 changed files. Large-change decomposition and production-behavior tests were examined.
- Discussion classification: inline comments `3919853798` and `3919853810` remain actionable and are represented above; `3919853815` is a stylistic metadata-date nit and is discarded. The walkthrough and aggregate review duplicate underlying comments and add no finding.
- The speculative concurrent-recorder rename collision lacks evidence that concurrent ownership is reachable; the future-enum compatibility concern is hypothetical; documented sidecar growth is pre-existing/deferred. These are not actionable findings.

Risk lenses checked:
- `network/protocol`: timeouts, retry/duplicate delivery, partial or malformed responses, and wire compatibility.
- `release/build`: untrusted-input and least-privilege CI, artifact identity/reproducibility, and required-path check triggering.
- `tests`: pre-fix sensitivity, boundary/negative cases, and whether tests exercise production behavior.

Check evidence:
The bundle reports complete check evidence with `all_completed: true` and `all_successful: true`. Successful checks include all four Rust test shards, Rust check, Rust format, Rust clippy, Rust audit, Relay runtime tests, Relay CLI E2E, CLI e2e, Tamarin proofs, simulator/vectors, C bindings smoke, installer tests, package/build jobs, and host-compatibility checks. The release-publication job was intentionally skipped. `automated_device_gate` is null, so no physical-device evidence exists for this head; conclusions rest on static review and the supplied CI evidence.

Datawav (approved):

PR #1644, opened 2026-09-03
Route fixed-text SQL through the statement cache

dannym-arx:
### Summary

storage-sqlite compiles every statement through rusqlite's prepared-statement cache instead of re-parsing the SQL text on each call.

### Context

The crate never used the cache. 94 prepare sites plus every Connection::execute and query_row compiled their SQL each time, and SQL compilation was about a tenth of app-message send CPU in a perf profile. One send issues 68 statements. The cache was unavailable because the workspace disables rusqlite's default features, which include it.

### What changed

A CachedSql extension adds execute_cached and query_row_cached over prepare_cached, and the non-test call sites use them. Plain prepare becomes prepare_cached outside tests, including the transaction boundaries. The rusqlite cache feature is enabled and the capacity raised from the default 16 to 256, since the engine paths alone issue well over a hundred distinct statements.

### Impact

In-memory SQLCipher, warm app-message send: 0 members 509 to 428 µs, 16 members 1179 to 923 µs, 64 members 2521 to 2286 µs. MlsGroup::load at 0 members 52 to 35 µs.

## Summary by CodeRabbit

- **Performance**
  - Improved SQLite efficiency by reusing prepared statements across storage operations.
  - Applied statement caching broadly across account, messaging, group, timeline, snapshot, transport, configuration, and other storage activities.
  - Configured database connections with a bounded cache for frequently used statements, reducing repeated preparation overhead.

- **Compatibility**
  - Existing storage behavior, transactions, error handling, data handling, and public APIs remain unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2261,6 +2261,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "hashlink"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "32069d97bb81e38fa67eab65e3393bf804bb85969f2bc06bf13f64aef5aba248"
+dependencies = [
+ "hashbrown 0.17.0",
+]
+
+[[package]]
 name = "hax-lib"
 version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5024,6 +5033,7 @@ dependencies = [
  "bitflags 2.13.0",
  "fallible-iterator",
  "fallible-streaming-iterator",
+ "hashlink",
  "libsqlite3-sys",
  "smallvec",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,7 +108,7 @@ nostr-relay-builder = "0.44"
 # reqwest/rustls stack, so the SDK's record-time histogram API is not needed.
 opentelemetry-proto = { version = "0.32", default-features = false, features = ["gen-tonic-messages", "metrics"] }
 prost = "0.14"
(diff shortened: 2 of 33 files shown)


Datawav (commented):
No correctness blocker found, and the exact-head storage tests and clippy pass, but the blanket “every fixed-text statement” claim is not yet accurate: several production paths still use uncached `Connection::execute` or `query_row`. Representative fixed or canonical generated variants remain in `encrypted_media_secrets.rs:35,98`, `account_projection.rs:1948`, `chat_list.rs:1362,1786,1813`, `timeline.rs:1034,1062`, and `storage/messages.rs:188,702`. These paths continue paying prepare cost, so please either route them through `CachedSql` (while leaving genuinely cardinality-dependent SQL uncached) or narrow the PR description and document the intended exceptions. The cache feature, capacity configuration, and converted call sites otherwise preserve the existing rusqlite semantics.

Verification: `cargo test -p storage-sqlite` passed 445 tests (4 ignored), `cargo clippy -p storage-sqlite --all-targets -- -D warnings` passed, and the current GitHub checks are green.

dannym-arx:
Addressed in e703fa33. The first pass only matched calls whose SQL argument began with a string literal, so raw-string literals, `format!` over constant fragments, and statements passed in as a parameter were skipped. All of those now go through `CachedSql`; the `format!` cases build from a fixed set of fragments (a table name from a constant list, a shared predicate, a projection ordering), so the resulting texts are a small bounded set.

Left uncached on purpose, and noted in the commit: migrations, which run once per database, and the journal-mode pragma at open. A final sweep over non-test code finds no other Connection-level `execute` or `query_row`.

PR #1645, opened 2026-09-03
Group each message's OpenMLS writes into one commit

dannym-arx:
### Summary

Each message's OpenMLS secret-tree write shares a commit with the work around it: inbound, with the retained message row; outbound, with the legacy-key delete that follows it.

### Context

create_message and process_message each ratchet the sender secret tree and write it back through the OpenMLS provider. Those writes ran outside any caller transaction, so each was its own autocommit unit, and under synchronous = FULL every commit is an fsync. An inbound message paid three commits before its application event was applied.

### What changed

The inbound path runs the retained-row persist and the MLS processing inside one with_transaction. The processing result leaves the closure as a value, so a decrypt failure still commits the row the error branches transition, as before.

The outbound path runs create_message inside one transaction so its secret write and legacy-key delete share a commit. The retained outgoing row still commits on its own afterwards. Transport wrapping sits between the two and is async, and the storage transaction API is synchronous, so this change does not make the outgoing row atomic with the ratchet advance. That gap predates this change and is unchanged by it.

### Impact

File-backed SQLCipher on btrfs, warm app-message send, together with the statement cache beneath: 0 members 5043 to 2728 µs, 64 members 14574 to 4766 µs. Disk numbers vary run to run by a wide margin, so treat these as the direction, not a precise split.

## Summary by CodeRabbit

* **Reliability Improvements**
  * Improved consistency when receiving and processing messages by grouping related storage updates into a single transaction.
  * Improved consistency when creating and sending application messages by ensuring message data and associated state are saved together.
  * Existing message handling and error behavior remain unchanged.

### DIFF
--- a/crates/cgka-engine/src/message_processor/send.rs
+++ b/crates/cgka-engine/src/message_processor/send.rs
@@ -888,12 +888,16 @@ impl<S: StorageProvider> Engine<S> {
         let wrap_metadata =
             GroupMessageMetadata::application(app_event.created_at, source_retention_seconds);
 
-        let out: MlsMessageOut = mls_group
-            .create_message(&provider, &self.identity.signer, &payload)
-            .map_err(|e| EngineError::Backend(format!("create_message: {e:?}")))?;
-        let out_bytes = out
-            .tls_serialize_detached()
-            .map_err(|e| EngineError::Serialize(format!("{e:?}")))?;
+        // `create_message` ratchets the sender secret tree and writes it back;
+        // one transaction turns that write's two autocommits into one COMMIT
+        // (one fsync under `synchronous = FULL`).
+        let out_bytes = self.storage.with_transaction(|_storage| {
+            let out: MlsMessageOut = mls_group
+                .create_message(&provider, &self.identity.signer, &payload)
+                .map_err(|e| EngineError::Backend(format!("create_message: {e:?}")))?;
+            out.tls_serialize_detached()
+                .map_err(|e| EngineError::Serialize(format!("{e:?}")))
+        })?;
 
         let ctx = group_lifecycle::build_group_context_snapshot(&mls_group, &provider)?;
         let wrapped = self
--- a/crates/cgka-engine/src/message_processor/ingest.rs
+++ b/crates/cgka-engine/src/message_processor/ingest.rs
@@ -1004,21 +1004,28 @@ impl<S: StorageProvider> Engine<S> {
             );
         }
 
-        self.persist_openmls_wire_message(
-            &openmls_msg,
-            &group_id,
-            current_epoch,
-            MessageState::Created,
-        )?;
-
+        // The retained row and the secret-tree ratchet `process_message`
+        // writes share one durable unit: under `synchronous = FULL` every
+        // COMMIT is an fsync, and these were three per inbound message. The
+        // processing result leaves the closure as a value so a decrypt
+        // failure still commits the row the error branches below transition.
+        //
         // Process via MLS. Commits may contain AppDataUpdate proposals,
         // which require the application to compute the resulting
         // AppDataDictionary before OpenMLS stages the commit.
-        let processed = match if msg_content_type == ContentType::Commit {
-            process_commit_with_app_data_updates(&mut mls_group, &provider, proto)
-        } else {
-            mls_group.process_message(&provider, proto)
-        } {
+        let processed = match self.storage.with_transaction(|_storage| {
+            self.persist_openmls_wire_message(
+                &openmls_msg,
+                &group_id,
+                current_epoch,
+                MessageState::Created,
+            )?;
+            Ok::<_, EngineError>(if msg_content_type == ContentType::Commit {
+                process_commit_with_app_data_updates(&mut mls_group, &provider, proto)
+            } else {
+                mls_group.process_message(&provider, proto)
+            })
+        })? {
             Ok(p) => p,
             Err(e) if process_message_error_is_too_distant_in_the_past(&e) => {
                 // Refine the historical classification (mdk#339):


Datawav (commented):
## Automated review

1. Low — Missing regression coverage for the new transaction boundary

Evidence:
- crates/cgka-engine/src/message_processor/ingest.rs:1016-1028 moves retained-row persistence and MLS processing into one with_transaction call.
- crates/cgka-engine/src/message_processor/send.rs:893-900 moves create_message and sender-ratchet persistence into one transaction.
- No test file is changed.

Triggering scenario:
A later refactor moves either persistence operation outside the transaction, or a storage-provider implementation does not enlist the OpenMLS writes in the surrounding transaction.

Consequence:
The intended single-commit invariant silently regresses, restoring extra durable commits and potentially separating the retained message row from ratchet persistence. The supplied broad CI checks would not specifically detect this transaction-boundary regression.

Smallest sound correction:
Add a counting storage-provider regression test for one successful inbound application message and one successful outbound application message. Assert that the retained row and relevant OpenMLS ratchet writes occur under the same transaction/commit. This test would have observed multiple commits before this change.

Confidence: High.

Audit evidence:
- Exact head SHA: 19f1cd3a9b77d0dd33a10d27aeb40e701b554368
- Changed files inspected:
  - crates/cgka-engine/src/message_processor/ingest.rs
  - crates/cgka-engine/src/message_processor/send.rs
- Design fit: The transaction wrappers match the stated successful-path atomicity and commit-reduction goal. The inbound code deliberately returns the MLS processing result as a value so expected processing errors can commit the retained row for subsequent error-state transitions. Both supplied head-context snapshots are marked truncated, so absent caller or storage-provider context was not treated as proof of safety.
- Discussion validation: The sole supplied discussion item reports that CodeRabbit skipped review and contains no actionable technical claim to preserve.

Risk lenses checked: general
- boundary and error paths: inspected processing errors, serialization errors, transaction propagation, and the inbound commit-on-processing-error behavior.
- backward compatibility: no public API or serialized-format changes are present; storage-provider transaction-enlistment behavior remains the relevant compatibility invariant.
- behavioral regression coverage: the central transaction boundary has no targeted regression test in the diff.
- Lifecycle/restart, accessibility, and authorization behavior are not changed by the supplied patches and were not separately routed.

Check evidence:
The evidence collection is marked complete, but the check suite is not complete or wholly successful: Rust tests (1/4) and C bindings smoke were in progress, while Rust tests (2/4), Rust tests (3/4), and Simulator and vectors were queued. Rust tests (4/4), Rust check, Rust clippy, Rust format, Rust audit, relay tests, Tamarin proofs, packaging, builds, and several compatibility checks completed successfully. “Publish wn-agent release” was skipped. Therefore, the supplied evidence does not establish that the full test suite passed.

No physical-device evidence exists for this head because no automated device gate evidence was supplied.

crates/cgka-engine/src/message_processor/send.rs
```diff
@@ -887,12 +887,16 @@ impl<S: StorageProvider> Engine<S> {
...
-        let out: MlsMessageOut = mls_group
-            .create_message(&provider, &self.identity.signer, &payload)
-            .map_err(|e| EngineError::Backend(format!("create_message: {e:?}")))?;
-        let out_bytes = out
-            .tls_serialize_detached()
-            .map_err(|e| EngineError::Serialize(format!("{e:?}")))?;
+        // `create_message` ratchets the sender secret tree and writes it back;
+        // one transaction turns that write's two autocommits into one COMMIT
+        // (one fsync under `synchronous = FULL`).
+        let out_bytes = self.storage.with_transaction(|_storage| {
```
Datawav, line 894: The transaction ends before transport wrapping and before `record_sent_openmls_application_message` at line 916. If wrapping fails or the process crashes after this closure commits, the sender secret tree has advanced without the retained outgoing message row, so the outbound path does not satisfy the atomicity promised by the PR title/summary. Please either make the row join this durable unit and add a fault-injection regression for this seam, or narrow the PR claim to grouping only OpenMLS’s internal writes.

Datawav (changes requested):
Requesting changes because the inbound path now groups the retained row and OpenMLS processing correctly, but the outbound path still does not provide the atomicity promised by the title and summary. `with_transaction` commits before `wrap_group_message_with_metadata(...).await`, while `record_sent_openmls_application_message` remains afterward; a peeler error or crash in that interval leaves the sender secret tree advanced without the retained outgoing message row. The exact-head focused ingest/application/atomicity tests and clippy pass, but there is no regression test for this seam.

Please either make the outbound retained row join the same durable transaction as the secret-tree write, with a fault-injection test covering failure after `create_message` and before record persistence, or narrow the PR’s title and description to the actual optimization: grouping OpenMLS’s internal writes into one commit without message-row atomicity.

Verification: focused `application_data_wire`, `ingest`, `crash_recovery_sqlite`, and `record_write_atomicity` test targets passed, and `cargo clippy -p cgka-engine --all-targets -- -D warnings` passed. The full crate run also exposed 19 `deferred_peel_lifecycle` failures at the pinned-policy setup; the same exact test fails on the stacked base, so that is pre-existing and not attributed to this patch. Current GitHub checks are green.

dannym-arx:
Taken the narrowing option. Title, description, and the commit message now state what the change does: inbound, the retained row and the MLS processing share one transaction; outbound, create_message's secret write and legacy-key delete share one commit, and the retained outgoing row still commits separately afterwards. The outbound gap the review describes predates this change and is unchanged by it: the transport wrap between create_message and the row write is async, and the storage transaction API is synchronous, so joining them would need a transaction that can be held across an await, which is a separate design change.

On the regression test: there is no counting storage provider in the crate, and the commit count is not observable through the current trait surface, so a test that pins the inbound single-transaction shape would need that harness first. Happy to add it in a follow-up if you want that coverage.

Datawav (commented):
## Automated review

Medium — Outbound sender state still commits before the corresponding message is durably retained.

Evidence: In `crates/cgka/src/engine/send.rs`, changed lines 895–903 place only `MlsGroup::create_message` and serialization inside `self.storage.with_transaction(...)`. That transaction ends before `wrap_group_message_with_metadata(...).await` and the later `record_sent_openmls_application_message` call around line 916.

Trigger: wrapping returns an error, or the process terminates after the new transaction commits but before the retained-message write.

Consequence: the OpenMLS sender secret-tree ratchet advances durably while no durable outgoing-message/outbox row exists. The message cannot be recovered or retried from storage, so this path does not provide atomic OpenMLS-state/message persistence.

Smallest sound correction: stage the serialized MLS message in a durable outbox row within the same transaction that advances the sender state, then perform asynchronous transport wrapping and finalize that row afterward. Add a fault-injection regression that interrupts execution immediately after the transaction and verifies that the exact outbound message remains recoverable. Confidence: high.

Audit evidence:
- Head SHA: 7f392d3d6dbd550c8ead5e42433b1e68cfb771b5
- Changed files inspected: `crates/cgka/src/engine/ingest.rs`, `crates/cgka/src/engine/send.rs`
- The inbound change groups retained-message persistence with OpenMLS processing and preserves the processing result for existing error classification. No additional actionable inbound issue was established from the supplied patch and bounded context.
- No physical-device evidence exists for this head because the recorded gate status is `capability_checked_not_installable`, meaning no exact-head Android artifact exists for Marmot/MDK.

Risk lenses checked: `general`, `tests`

Check evidence: `Build Rust` and `Build Kotlin` are recorded as successful for the supplied head. No successful behavioral or fault-injection test evidence was supplied, so this review does not claim that tests passed.

dannym-arx:
On the medium finding for 7f392d3d: the ordering it describes is the pre-existing behaviour of this path. Before this change, create_message's secret write autocommitted on its own before the transport wrap, and the retained row was written after the wrap, exactly as now. The change only folds that write's two autocommits into one commit; it neither introduces nor widens the window between the ratchet advance and the row. The description and commit message state this explicitly since the last revision and claim no outbound atomicity.

The outbox shape suggested (stage the serialized MLS bytes durably in the create_message transaction, wrap, then finalize the row) is a reasonable follow-up, but it is a new durability feature with its own recovery path rather than part of this optimisation, so I am leaving it out of this PR.

Datawav (approved):

PR #1647, opened 2026-09-03
Keep loaded MlsGroups across engine calls

dannym-arx:
### Summary

The engine keeps loaded MlsGroup objects across calls and reuses one only while the storage backend reports no OpenMLS writes since it was stored.

### Context

MlsGroup::load is 13 storage reads plus a ratchet-tree decode that grows with the member count. The app-message send, the inbound message, and every group read accessor each paid it per call, and a send loaded twice: once to encrypt and once inside the audit context to list admins.

### What changed

StorageProvider gains a defaulted mls_write_generation. storage-sqlite counts every write to the OpenMLS store, every transaction rollback, and every snapshot restore, so a write through a freshly loaded group, a replay, or a rolled-back merge invalidates the cached entry with no call-site bookkeeping. An operation takes the object out for its duration and puts it back only after it succeeded; error paths drop it. Backends that report None get no caching. The cache holds 32 groups and clears wholesale at the cap. A unit test pins the reuse and the invalidation on rollback.

Explicit invalidation at the seven commit-merge sites would be a smaller diff, but it cannot see the replay and restore paths that load their own group and write behind the cache.

### Impact

In-memory SQLCipher, warm app-message send, full stack against master: 0 members 509 to 308 µs, 16 members 1179 to 579 µs, 64 members 2521 to 1057 µs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when sending and receiving group messages, including inactive, duplicate, malformed, or rejected messages.
  - Improved consistency of group state after updates, rollbacks, restores, and concurrent storage changes.
  - Reduced unnecessary repeated loading of group state for read and write operations.
  - Improved handling of transient storage contention during transaction rollbacks.
- **Tests**
  - Added coverage for state tracking across multiple storage connections, commits, and closed databases.

### DIFF
--- a/crates/cgka-engine/src/distributed_convergence.rs
+++ b/crates/cgka-engine/src/distributed_convergence.rs
@@ -2410,7 +2410,7 @@ fn decode_convergence_policy(
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
     use std::sync::Arc;
 
     use async_trait::async_trait;
@@ -2524,7 +2524,7 @@ mod tests {
         }
     }
 
-    fn test_engine() -> crate::Engine<storage_sqlite::SqliteAccountStorage> {
+    pub(crate) fn test_engine() -> crate::Engine<storage_sqlite::SqliteAccountStorage> {
         let key = signing_key();
         let identity = key.verifying_key().to_bytes().to_vec();
         crate::EngineBuilder::new(storage_sqlite::SqliteAccountStorage::in_memory().unwrap())
--- a/crates/cgka-engine/src/engine.rs
+++ b/crates/cgka-engine/src/engine.rs
@@ -144,6 +144,7 @@ pub struct Engine<S: StorageProvider> {
     /// Per-group state-machine owner. Every transition, pending-ref
     /// allocation, and fork-detection marker flows through this struct.
     pub(crate) epoch_manager: crate::epoch_manager::EpochManager,
+    pub(crate) mls_group_cache: crate::mls_group_cache::MlsGroupCache,
 
     /// Storage-layer identity of the origin commit behind each in-flight
     /// pending publish, so `do_confirm_published` can mark the sent commit
@@ -554,6 +555,7 @@ impl<S: StorageProvider> EngineBuilder<S> {
             wall_clock: self.wall_clock,
             maintenance_random: self.maintenance_random,
             epoch_manager: crate::epoch_manager::EpochManager::new(),
+            mls_group_cache: crate::mls_group_cache::MlsGroupCache::default(),
             pending_origin_commits: HashMap::new(),
             events_buf: pending_application_events.into(),
             auto_publish_buf: VecDeque::new(),
@@ -2921,28 +2923,23 @@ impl<S: StorageProvider> Engine<S> {
         self.update_stored_message_state(id, MessageState::Processed)
     }
 
-    /// One liveness-gated `MlsGroup::load` for read-only accessors. A missing
-    /// group maps to `UnknownGroup`.
-    fn live_mls_group(&self, group_id: &GroupId) -> Result<openmls::group::MlsGroup, EngineError> {
+    /// Liveness-gated read-only access to the group's `MlsGroup`, served from
+    /// the engine's group cache. A missing group maps to `UnknownGroup`.
+    fn with_live_mls_group<R>(
+        &self,
+        group_id: &GroupId,
+        f: impl FnOnce(&openmls::group::MlsGroup) -> Result<R, EngineError>,
+    ) -> Result<R, EngineError> {
         self.ensure_group_live(group_id)?;
-        let provider = crate::provider::EngineOpenMlsProvider::<S>::new(
-            &self.crypto,
-            self.storage.mls_storage(),
-        );
-        let mls_gid = openmls::group::GroupId::from_slice(group_id.as_slice());
-        openmls::group::MlsGroup::load(
-            <crate::provider::EngineOpenMlsProvider<'_, S> as openmls_traits::OpenMlsProvider>::storage(&provider),
-            &mls_gid,
-        )
-        .map_err(|e| EngineError::Backend(format!("load: {e:?}")))?
-        .ok_or_else(|| EngineError::UnknownGroup(group_id.clone()))
+        self.with_mls_group(group_id, f)
     }
 
     /// Return the current Marmot admin policy keys mirrored from signed MLS
     /// group state.
     pub fn admin_pubkeys(&self, group_id: &GroupId) -> Result<Vec<[u8; 32]>, EngineError> {
-        let mls_group = self.live_mls_group(group_id)?;
-        let mut admins = crate::app_components::admins_of_group(&mls_group)?;
+        let mut admins = self.with_live_mls_group(group_id, |mls_group| {
+            crate::app_components::admins_of_group(mls_group)
+        })?;
         admins.sort();
         admins.dedup();
         Ok(admins)
@@ -3002,21 +2999,21 @@ impl<S: StorageProvider> Engine<S> {
         group_id: &GroupId,
         component_id: AppComponentId,
     ) -> Result<EpochId, EngineError> {
-        let mls_group = self.live_mls_group(group_id)?;
+        self.with_live_mls_group(group_id, |mls_group| {
+            let required_components =
+                crate::app_components::required_app_components_of_group(mls_group)?;
+            if !required_components.contains(component_id) {
+                return Err(EngineError::Other(format!(
+                    "group does not require app component {component_id:#06x}"
+                )));
+            }
 
-        let required_components =
-            crate::app_components::required_app_components_of_group(&mls_group)?;
-        if !required_components.contains(component_id) {
-            return Err(EngineError::Other(format!(
-                "group does not require app component {component_id:#06x}"
-            )));
-        }
-
-        if let Some(staged) = mls_group.pending_commit() {
-            Ok(EpochId(staged.group_context().epoch().as_u64()))
-        } else {
-            Ok(EpochId(mls_group.epoch().as_u64()))
-        }
+            if let Some(staged) = mls_group.pending_commit() {
+                Ok(EpochId(staged.group_context().epoch().as_u64()))
+            } else {
+                Ok(EpochId(mls_group.epoch().as_u64()))
+            }
+        })
     }
 }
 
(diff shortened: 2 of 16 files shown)


Datawav (commented):
## Automated review

Three actionable findings.

1. Major — Cross-connection writes do not invalidate cached MLS state
   Exact changed-line evidence: crates/storage-sqlite/src/connection.rs adds `openmls_writes: AtomicU64`; the new `openmls_write_generation()` returns this handle-local counter, while crates/cgka-engine/src/mls_group_cache.rs reuses an entry when `cached.generation == generation`.
   Triggering state: another SQLite connection or process commits OpenMLS state to the same database after this engine caches an `MlsGroup`.
   Consequence: the local counter remains unchanged, so the engine can reuse superseded epoch/ratchet state and produce messages from stale cryptographic state.
   Confidence: High. The supplied source context expressly anticipates separate-connection/process contention, but the new validity signal observes only one `SharedConnectionInner`.
   Smallest sound correction: include a cross-connection change signal such as SQLite `PRAGMA data_version` in the reuse token, or return `None` unless exclusive database access is guaranteed.

2. Major — Sampling generation only when returning the group absorbs a concurrent write
   Exact changed-line evidence: crates/cgka-engine/src/mls_group_cache.rs adds `return_mls_group`, which reads `self.storage.mls_write_generation()` after the operation and stores the returned object with that newly sampled generation.
   Triggering state: another writer changes OpenMLS state after this operation loads/takes the group but before `return_mls_group` samples the counter.
   Consequence: an object derived from the earlier state is tagged with the later generation and can pass the next equality check as current.
   Confidence: High. This is a direct ordering window in the newly added take/return protocol.
   Smallest sound correction: capture the generation before loading/taking the group, carry it through the operation, and cache only if a second read at return still equals that captured value.

3. Major — Some restore paths can mutate OpenMLS state without advancing the generation
   Exact changed-line evidence: crates/storage-sqlite/src/storage/snapshots.rs adds `note_openmls_write()` only in the `rollback`, `rollback_group_state`, and `import_replay` wrappers before delegating to restore primitives; the changed storage contract in crates/traits/src/storage.rs requires every snapshot restore that bypasses the provider to change the count. No corresponding generation bump is added to the supplied `restore_group_state_checkpoint` path.
   Triggering state: a checkpoint restore, or another caller reaching an uninstrumented restore primitive, rewrites group-scoped OpenMLS values while an entry is cached.
   Consequence: the cached object can survive restoration and be reused against state from a different checkpoint.
   Confidence: High for checkpoint restore; Medium for future/direct callers of the lower-level primitives.
   Smallest sound correction: advance the generation inside each lowest-level restore primitive, including the implementation backing `restore_group_state_checkpoint`, rather than relying on selected callers; add assertions that every restore entry point strictly advances it.

Discarded advisory claims:
- The nested ingest error formatting is real but diagnostic-only and does not materially affect correctness.
- Test-coverage observations are evidence limitations, not independent code defects.
- Secret retention, the `ponytail:` comment, and reordered error precedence lack a demonstrated harmful triggering state.
- Discussion classification: the sole automated “review skipped” comment contains no substantive claim about this head; there is no maintainer feedback to preserve or rebut.

Audit evidence:
- Exact head: 80e9e188d13d9f037d1e357cd2e5e4536d893f0c
- Patch coverage is complete across 13 changed files, including crates/cgka-engine/src/mls_group_cache.rs, crates/storage-sqlite/src/connection.rs, crates/storage-sqlite/src/storage/snapshots.rs, and crates/traits/src/storage.rs.
- Review conclusions rest on the immutable patch, bounded source context, and supplied CI state.
- The automated device gate is absent; no physical-device evidence exists for this head.

Risk lenses checked: persistence/migration, concurrency/lifecycle
- concurrency/lifecycle: cancellation and stale callbacks; ordering/race windows; process or lifecycle restart.

Check evidence:
- Completed successfully: Tamarin proofs, Rust tests (2/4), Rust audit, Rust check, Rust clippy, Rust format, terminal harness installer tests, several platform harness builds, OpenClaw beta host compatibility, and packaging/config jobs.
- The supplied snapshot is incomplete: Rust tests (1/4, 3/4, 4/4), Relay CLI E2E, C bindings smoke, and convergence-policy assertions were queued or in progress; therefore it does not establish complete CI success for this head.

Datawav (commented):
No actionable findings.

Audit evidence:
Exact head: c47dbefde0eee987be93a4945841d7280558da0a
Base: 18a6fa611f9362f00bd36815275faf76af76c243
Patch coverage is complete across 13 files, including crates/cgka-engine/src/mls_group_cache.rs, crates/cgka-engine/src/engine.rs, crates/storage-sqlite/src/connection.rs, and crates/traits/src/storage.rs.

The advisory’s clean verdict is supported after re-derivation against the patch and bounded source context. Cache reuse is conditioned on the storage generation; operations take the group out of the cache and return it only after success; write, rollback, snapshot-restore, replay, and group-deletion paths conservatively advance invalidation state; unsupported backends retain the safe None default; and process restart discards the in-memory cache. No supported changed-line defect remains.

Risk lenses checked:
- persistence/migration — upgrade and rollback from real prior state; partial failure and idempotent retry; data preservation. No schema migration is introduced, legacy backends disable caching by default, rollback/restore paths invalidate cached state, and unsuccessful operations do not return the group to the cache.
- concurrency/lifecycle — cancellation and stale callbacks; ordering/race windows; process or lifecycle restart. Generation checks fence reuse, groups are removed while an operation owns them, error paths drop them, and restart cannot retain an in-memory cached group.

Discussion classification:
The sole discussion item is an automated “review skipped” notice. It makes no claim about the patch and is non-actionable; the bundle contains no maintainer feedback requiring reconciliation.

Check evidence:
The bundle reports complete, completed check data. Rust check, format, clippy, audit, all four Rust test shards, simulator/vectors, Tamarin proofs, CLI and relay CLI E2E, C-bindings smoke, and the listed platform builds concluded success. “Relay runtime tests” concluded “failure”; the bundle provides no diagnostic tying that result to a changed line, so it is recorded as a CI caveat rather than converted into a speculative finding. “Publish wn-agent release” was skipped.

Physical-device scope:
The automated_device_gate status is capability_checked_not_installable and is recorded against a different SHA, 80e9e188d13d9f037d1e357cd2e5e4536d893f0c. Therefore, no physical-device evidence exists for this head. Conclusions rest on static review and the supplied CI evidence.

Not approved on this head: failed CI: Relay runtime tests (failure). A corrected head is re-audited automatically.

Datawav (commented):
## Automated review

Two major actionable findings remain at exact head 659f08a366d4e21095fed6b28f218fd281deb807.

1. Major — stale MLS group can be tagged with a newer write generation

Exact changed-line evidence: crates/cgka-engine/src/mls_group_cache.rs:77-92. `return_mls_group` reads `self.storage.mls_write_generation()` after the operation and stores the returned `MlsGroup` under that newly sampled generation.

Triggering state: Another engine or thread sharing the storage writes OpenMLS state after this group was loaded or last mutated but before `return_mls_group` samples the generation.

Consequence: The old in-memory group receives the post-write generation. A later `take_mls_group` observes equal generations and reuses stale cryptographic group state instead of reloading it. This violates the cache’s data-preservation and ordering invariant.

Confidence: High.

Smallest sound correction: Have `take_mls_group` return the generation associated with the loaded/taken group and require `return_mls_group` to validate that generation against the current value before insertion. If it changed, discard the object. Equivalent synchronization covering both OpenMLS writes and generation tagging is also sound.

2. Major — cache reuse bypasses closed-storage semantics

Exact changed-line evidence: crates/storage-sqlite/src/connection.rs:1069-1071 returns `Some(self.connection.openmls_write_generation())` without checking the connection’s closed flag. The changed crates/cgka-engine/src/group_lifecycle.rs:1349 path now implements `do_own_leaf_index` through `with_mls_group`, allowing a cache hit without touching SQLite.

Triggering state: An MLS group is cached, `SqliteAccountStorage::close()` is called, and a subsequent read such as `do_own_leaf_index` finds the same write generation.

Consequence: The operation can succeed from cached state after storage closure, contrary to the terminal closed-state behavior followed by uncached operations. Lifecycle behavior therefore depends on whether the group happened to be cached.

Confidence: High.

Smallest sound correction: Make `SqliteAccountStorage::mls_write_generation()` return `None` when the shared connection is closed, forcing the normal storage path to surface the closed-state error. Add a regression test that warms the cache, closes storage, and calls a cached read accessor.

Audit evidence:

Exact head: 659f08a366d4e21095fed6b28f218fd281deb807

Changed files examined include crates/cgka-engine/src/mls_group_cache.rs, crates/cgka-engine/src/message_processor/ingest.rs, crates/cgka-engine/src/group_lifecycle.rs, crates/storage-sqlite/src/connection.rs, crates/storage-sqlite/src/openmls_storage/value_store.rs, and crates/traits/src/storage.rs. Bundle patch coverage is complete with no missing patch files.

Risk lenses checked: persistence/migration; concurrency/lifecycle.

Persistence/migration analysis covered prior-state and snapshot restoration, rollback, partial-operation retry/idempotence, and data preservation. The added generation increments conservatively invalidate cache entries for rollback, snapshot import, deletion, and OpenMLS value writes; the unsynchronized return-time tagging remains the data-preservation defect above.

Concurrency/lifecycle analysis covered cancellation and stale callbacks, ordering/race windows, and process or lifecycle restart. The return-time generation race and post-close cache hit remain actionable. Discussion claims about dropping reusable groups on early ingest exits describe a bounded performance opportunity rather than a correctness defect, while the global counter’s cross-group invalidation is conservative and stylistic/performance-related; neither is retained as an actionable finding. Informational walkthrough and duplicate review-summary material are also discarded.

Check evidence: The supplied CI set is complete and records `all_completed: true` and `all_successful: true`. Successful checks include Rust check, Rust clippy, Rust format, Rust audit, Rust tests (1/4 through 4/4), Relay runtime tests, Simulator and vectors, CLI e2e (mock relay), C bindings smoke, and Tamarin proofs. These checks do not exercise the two identified interleavings sufficiently to disprove them.

The automated device gate reports `capability_checked_not_installable`, references c47dbefde0eee987be93a4945841d7280558da0a rather than this exact head, and explicitly has no trusted Android artifact. Therefore, no physical-device evidence exists for this head. Conclusions rest on static exact-head review and the supplied CI evidence.

dannym-arx:
All three addressed in 7cb96337.

1. Cross-connection writes: the generation now folds in SQLite's `data_version`, which moves on every commit from another connection or process, alongside the in-process write counter. Closed storage reports no generation, so nothing is served from the cache after close (also covers the CodeRabbit close finding).

2. Return-time sampling: the generation is recorded at take. A group returned after writing through it is kept only if the foreign half is unchanged since the take; a group returned unmodified is kept only if the whole generation is unchanged. The read accessors and the ingest early exits use the unmodified form, so a deferred or ignored message no longer forces a reload on its next retry (the CodeRabbit ingest finding). The one case the token cannot see, a second writer on the same connection touching this group's state mid-operation, is documented: the engine is the only writer of group state on its connection and is single-threaded per account.

3. Restore paths: the bump now lives inside every restore primitive that has the connection, including the checkpoint restore, instead of in selected wrappers.

Tests: a second connection's commit moves the generation; closed storage reports none; the cache test's rollback check asserts inequality since the packed value is not monotonic.

On the per-group counter nit: left as a documented ceiling. One counter per connection evicts other groups' entries on any write, which is the right trade until an account actively works many groups at once.

Datawav (approved):

PR #1654, opened 2026-09-03
Fix large Blossom upload timeout retry path

### LINKED ISSUE #1585: Large Blossom uploads fail when response bytes are idle for 15 seconds
## Observed regression

After the Hermes connector was upgraded to the stock `wn-agent` 0.9.15 release, repeated delivery of the same 40,405,416-byte signed staging APK failed with the privacy-safe, non-retryable control error `app_error`; authoritative timeline pagination showed that no message was published. The same account and destination had successfully delivered earlier APKs through a locally patched 0.9.14 binary. This is a deployment boundary, not an upstream 0.9.14-to-0.9.15 source change: both release tags contain the same shared 15-second media HTTP read-idle timeout, while the 0.9.15 installer replaced the locally patched binary with the stock release.

No account, group, event, blob URL, key, or local path is needed to reproduce or diagnose this failure.

## Source-confirmed mechanism

At current `master` commit `2f38d125f600a9142f0b77f70de069af4a1ba973`, the media HTTP client combines a 5-second connect timeout, a client-wide 15-second read timeout, and a 60-second client total timeout ([timeout constants](https://github.com/marmot-protocol/mdk/blob/2f38d125f600a9142f0b77f70de069af4a1ba973/crates/marmot-app/src/media/blossom.rs#L20-L34), [shared client builder](https://github.com/marmot-protocol/mdk/blob/2f38d125f600a9142f0b77f70de069af4a1ba973/crates/marmot-app/src/media/blossom.rs#L456-L490)). Large Blossom uploads override the request total timeout to 15 minutes, but they still use that shared client and therefore retain its 15-second read-idle deadline while the server is receiving the request body and has not yet returned response bytes ([upload request](https://github.com/marmot-protocol/mdk/blob/2f38d125f600a9142f0b77f70de069af4a1ba973/crates/marmot-app/src/media/blossom.rs#L61-L85)). Stock 0.9.14 and 0.9.15 contain the same combination ([0.9.14 client](https://github.com/marmot-protocol/mdk/blob/235c8ade2920414679e59d7a5f1a0e78651756a4/crates/marmot-app/src/media/blossom.rs#L475-L488), [0.9.15 client](https://github.com/marmot-protocol/mdk/blob/f7159003eb76e00764c0320ba783a0f6dde7486f/crates/marmot-app/src/media/blossom.rs#L475-L488)).

The timeout is reduced to the generic `BlobStore("request timed out")` app error ([reqwest error projection](https://github.com/marmot-protocol/mdk/blob/2f38d125f600a9142f0b77f70de069af4a1ba973/crates/marmot-app/src/media/blossom.rs#L653-L665)), then the connector exposes every non-reaction `AppError` as `app_error` with the static message `connector request failed` ([connector error projection](https://github.com/marmot-protocol/mdk/blob/2f38d125f600a9142f0b77f70de069af4a1ba973/crates/agent-connector/src/error.rs#L51-L105)). The Hermes adapter marks only `send_in_progress` control errors retryable, so `app_error` is terminal to the caller ([adapter retry classification](https://github.com/marmot-protocol/mdk/blob/2f38d125f600a9142f0b77f70de069af4a1ba973/integrations/hermes/marmot/adapter.py#L1705-L1718)). Media is uploaded before its descriptor can be included in a message, so this failure correctly leaves no timeline publication but gives the caller no safe typed recovery path ([encrypted upload and fallback](https://github.com/marmot-protocol/mdk/blob/2f38d125f600a9142f0b77f70de069af4a1ba973/crates/marmot-app/src/media/mod.rs#L483-L517), [attachment upload path](https://github.com/marmot-protocol/mdk/blob/2f38d125f600a9142f0b77f70de069af4a1ba973/crates/marmot-app/src/media/mod.rs#L552-L617)).

## Reproduction

1. Run stock `wn-agent` 0.9.15 against a Blossom endpoint that accepts an authenticated upload body but returns no response bytes while it is still ingesting the body.
2. Send a generated or otherwise non-sensitive payload large enough for the server-side request-body read to cross 15 seconds; the observed production payload was 40,405,416 bytes.
3. In a deterministic test server, read the first 64 KiB, pause request-body reads for 16 seconds, then consume the rest and return a valid descriptor.
4. Observe that the stock client fails before the valid response with `app_error`, marks the failure non-retryable, and publishes no timeline message.
5. Remove only the shared client read-idle deadline while retaining the existing connect and per-operation total deadlines; the same request completes and publishes exactly once.

## Expected behavior

An authenticated Blossom upload that remains within the media size limit and 15-minute upload budget must not fail solely because the server produces no response bytes for more than 15 seconds while receiving the request body. Dead connections and bounded fetches must still fail promptly, and a failed pre-publication upload must remain privacy-safe and safely retryable without implying that a message was published.

## Acceptance criteria

- Give Blossom upload requests timeout semantics appropriate to request-body transfer: preserve the 5-second connect bound and 15-minute upload total bound, but do not apply the shared 15-second response read-idle deadline while the server is still receiving a valid bounded body.
- Keep profile-image and encrypted-media download operations bounded by their existing total and read-idle policies; do not remove timeout protection process-wide without operation-specific replacement bounds.
- Preserve URL validation, DNS/IP pinning, redirect refusal, proxy refusal, content-size ceilings, authentication, endpoint fallback, and privacy-safe error messages.
- Return a typed transport failure that lets the connector and Hermes adapter distinguish a retryable pre-publication Blossom timeout from terminal application errors; retries must not claim or create a publication until upload and message publication actually succeed.
- Add `marmot-app` regression test `blossom_upload_allows_response_gap_longer_than_read_timeout`, using a deterministic server that withholds response bytes for more than 15 seconds after beginning to receive a bounded request body.
- Add `marmot-app` regression test `blossom_upload_allows_continuous_slow_body_ingest`, proving that a server can continuously consume a large request body for more than 15 seconds before returning its descriptor.
- Add `agent-connector` and Hermes adapter coverage named `send_media_timeout_is_retryable_before_publication`, asserting a privacy-safe typed code, no timeline publication on the failed attempt, and exactly one publication after a same-artifact retry succeeds.
- Ship the fix in a new immutable `wn-agent` patch release and verify the installer selects that fixed release. Do not replace the published 0.9.15 assets: the installer downloads and checksum-verifies an exact release archive, then overwrites the target `wn-agent` binary without the plugin's backup step ([installer binary replacement](https://github.com/marmot-protocol/mdk/blob/2f38d125f600a9142f0b77f70de069af4a1ba973/scripts/install-hermes-marmot.sh#L782-L809), [plugin backup contrast](https://github.com/marmot-protocol/mdk/blob/2f38d125f600a9142f0b77f70de069af4a1ba973/scripts/install-hermes-marmot.sh#L814-L844)). Release and operator guidance must state that installing an affected stock release replaces a locally patched binary and must retain an explicit rollback binary until the fixed connector is verified.

## Related work and duplicate disposition

This is a focused reliability regression, not a duplicate of [#1355](https://github.com/marmot-protocol/mdk/issues/1355), which owns negotiated size policy, bounded streaming, progress, and cancellation, or [#1529](https://github.com/marmot-protocol/mdk/issues/1529), which owns download-side bad-locator failover and connection reuse. [PR #1388](https://github.com/marmot-protocol/mdk/pull/1388) added the 512 MiB in-memory ceiling and a 15-minute blob transfer budget but left the older shared 15-second client read timeout in place. Keep this issue standalone and cross-linked: it can land as a narrow upload-timeout and typed-retry correction without waiting for the broader transfer architecture in #1355.

erskingardner:
## Summary

- give Blossom uploads a dedicated pinned HTTP client policy that keeps the 5-second connect and 15-minute request bounds without inheriting the download-side 15-second read-idle timeout
- retain separate 15-second, size-bounded reads for the small Blossom descriptor and error response bodies
- preserve timeout identity through endpoint fallback as the privacy-safe `media_upload_timeout` control error with an explicit server-owned `retryable` field
- make Hermes honor server retryability while preserving legacy `send_in_progress` behavior, staged paths, and idempotency keys

## Verification

- `just fast-ci`
- `cargo test -p marmot-app media::tests`
- focused Hermes retry/publication regression
- both 40,405,416-byte upload regressions pass after exceeding the former 15-second boundary

The full local `marmot-app` suite reached 881 passing tests and six unrelated epoch-backfill failures; a representative exact test reproduces unchanged from clean `origin/master` at `1fd198be`. The full Hermes suite has one unrelated pre-existing media-preflight message assertion failure in untouched code; the new timeout and same-artifact retry tests pass.

The immutable patch-release and installer verification step remains a manual release operation and is intentionally not part of this source-change commit.

Fixes #1585

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved media upload reliability by distinguishing upload timeouts from other failures.
  * Uploads continue through fallback servers after a timeout and provide a clear, privacy-safe error if all attempts fail.
  * Slow uploads and delayed responses are handled more reliably without failing prematurely.

* **Improvements**
  * Error responses now indicate whether retrying is appropriate.
  * Retry behavior is preserved across integrations, including media upload retries, while preventing duplicate publication.
  * Connection-capacity errors now accurately indicate when retrying is unavailable.

### DIFF
--- a/crates/agent-connector/src/connection.rs
+++ b/crates/agent-connector/src/connection.rs
@@ -91,6 +91,7 @@ impl AgentConnector {
                         AgentControlResponse::Error {
                             code: "server_busy".to_owned(),
                             message: "agent connector subscription capacity is busy".to_owned(),
+                            retryable: true,
                         },
                     );
                     write_control_frame(&mut write_half, &response).await?;
@@ -141,6 +142,7 @@ impl AgentConnector {
         AgentControlResponse::Error {
             code: err.code().to_owned(),
             message: err.client_message().to_owned(),
+            retryable: err.retryable(),
         }
     }
 
@@ -528,6 +530,7 @@ impl AgentConnector {
             other => Ok(AgentControlResponse::Error {
                 code: "unsupported_request".to_owned(),
                 message: unsupported_request_message(&other).to_owned(),
+                retryable: false,
             }),
         }
     }
--- a/crates/agent-connector/src/lib.rs
+++ b/crates/agent-connector/src/lib.rs
@@ -420,7 +420,10 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
                 agent_control::AgentControlResponse::Error {
                     code: "server_busy".to_owned(),
                     message: "agent connector connection capacity is busy".to_owned(),
-                    retryable: true,
+                    // Admission happens before any request frame is read, so
+                    // this response cannot carry a request id. Do not promise
+                    // application-level retryability on an uncorrelated frame.
+                    retryable: false,
                 },
             );
             let _ = crate::connection::write_control_frame_with_timeout(
--- a/crates/agent-connector/src/lib.rs
+++ b/crates/agent-connector/src/lib.rs
@@ -420,6 +420,7 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
                 agent_control::AgentControlResponse::Error {
                     code: "server_busy".to_owned(),
                     message: "agent connector connection capacity is busy".to_owned(),
+                    retryable: true,
                 },
             );
             let _ = crate::connection::write_control_frame_with_timeout(
--- a/crates/marmot-app/src/media/blossom.rs
+++ b/crates/marmot-app/src/media/blossom.rs
@@ -690,12 +690,7 @@ async fn media_http_client_for_url(
     url: &Url,
     allow_loopback_http: bool,
 ) -> Result<reqwest::Client, AppError> {
-    validate_blossom_fetch_url(url, allow_loopback_http)
-        .map_err(|err| AppError::BlobStore(format!("unsafe Blossom URL: {err}")))?;
-    let allow_loopback = url.scheme() == "http"
-        && allow_loopback_http
-        && url.host().map(is_loopback_host).unwrap_or(false);
-    let pin = resolve_media_host(url, allow_loopback).await?;
+    let pin = resolve_pinned_media_host_for_url(url, allow_loopback_http).await?;
     build_pinned_media_http_client(pin)
 }
 
@@ -703,13 +698,20 @@ async fn media_http_upload_client_for_url(
     url: &Url,
     allow_loopback_http: bool,
 ) -> Result<reqwest::Client, AppError> {
+    let pin = resolve_pinned_media_host_for_url(url, allow_loopback_http).await?;
+    build_pinned_media_upload_client(pin)
+}
+
+async fn resolve_pinned_media_host_for_url(
+    url: &Url,
+    allow_loopback_http: bool,
+) -> Result<Option<(String, Vec<SocketAddr>)>, AppError> {
     validate_blossom_fetch_url(url, allow_loopback_http)
         .map_err(|err| AppError::BlobStore(format!("unsafe Blossom URL: {err}")))?;
     let allow_loopback = url.scheme() == "http"
         && allow_loopback_http
         && url.host().map(is_loopback_host).unwrap_or(false);
-    let pin = resolve_media_host(url, allow_loopback).await?;
-    build_pinned_media_upload_client(pin)
+    resolve_media_host(url, allow_loopback).await
 }
 
 fn build_pinned_media_http_client(
--- a/crates/marmot-app/src/media/blossom.rs
+++ b/crates/marmot-app/src/media/blossom.rs
@@ -264,7 +264,7 @@ pub(crate) async fn upload_blossom_blob_with_content_type(
 ) -> Result<String, AppError> {
     let (upload_url, server_host) = blossom_upload_endpoint(server)?;
     let authorization = blossom_authorization_header(signer, &server_host, blob_hash_hex).await?;
-    let client = media_http_client_for_url(&upload_url, allow_loopback_http).await?;
+    let client = media_http_upload_client_for_url(&upload_url, allow_loopback_http).await?;
     let response = client
         .put(upload_url.clone())
         .timeout(MEDIA_BLOB_TRANSFER_TIMEOUT)
@@ -274,13 +274,20 @@ pub(crate) async fn upload_blossom_blob_with_content_type(
         .body(blob)
         .send()
         .await
-        .map_err(reqwest_blob_error)?;
+        .map_err(reqwest_upload_error)?;
     if !response.status().is_success() {
         return Err(blossom_upload_status_error(response).await);
     }
-    let descriptor_body = read_limited_blossom_body(response, MAX_BLOSSOM_DESCRIPTOR_BYTES)
-        .await
-        .map_err(|_| AppError::BlobStore("upload descriptor exceeds size limit".into()))?;
+    let descriptor_body =
+        match read_limited_blossom_upload_body(response, MAX_BLOSSOM_DESCRIPTOR_BYTES).await {
+            Ok(body) => body,
+            Err(AppError::MediaUploadTimedOut) => return Err(AppError::MediaUploadTimedOut),
+            Err(_) => {
+                return Err(AppError::BlobStore(
+                    "upload descriptor exceeds size limit".into(),
+                ));
+            }
+        };
     let descriptor = serde_json::from_slice::<BlossomBlobDescriptor>(&descriptor_body)
         .map_err(|_| AppError::BlobStore("upload returned an invalid descriptor".into()))?;
     if let Some(sha256) = descriptor.sha256.as_deref()
@@ -318,7 +325,7 @@ async fn blossom_upload_status_error(response: reqwest::Response) -> AppError {
         .get("X-Reason")
         .and_then(|value| value.to_str().ok())
         .and_then(privacy_safe_server_reason);
-    let body = read_limited_blossom_body(response, MAX_BLOSSOM_ERROR_BODY_BYTES)
+    let body = read_limited_blossom_upload_body(response, MAX_BLOSSOM_ERROR_BODY_BYTES)
         .await
         .ok();
     let body_reason = body
@@ -678,6 +685,7 @@ fn same_registrable_domain(left: &str, right: &str) -> bool {
     }
 }
 
+#[cfg(test)]
 async fn media_http_client_for_url(
     url: &Url,
     allow_loopback_http: bool,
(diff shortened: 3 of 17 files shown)


crates/marmot-app/src/media/mod.rs
```diff
@@ -664,13 +665,19 @@ async fn upload_blossom_blob_with_fallback(
...
+                timed_out |= matches!(err, AppError::MediaUploadTimedOut);
+                failures.push(format!(
+                    "server {}: {}",
+                    idx + 1,
+                    upload_error_summary(&err)
+                ));
+            }
         }
     }
+    if timed_out {
```
erskingardner, line 678: Non-blocking design question: when any server times out, the per-server failure summaries are discarded entirely, so a chain where server 1 hard-fails (e.g. a privacy-safe `upload returned HTTP 4xx`) and server 2 times out surfaces only `MediaUploadTimedOut`. A retry will then re-walk the same chain, spend the 15-minute budget on server 1, and hit the same hard failure again with no signal that a non-timeout failure exists. Retry-before-publication is still safe and privacy-safe, so this is fine to ship as-is, but if the aggregate summary survives anywhere (or a hard failure on the last candidate could win over an earlier timeout), diagnosability would improve. If this precedence is deliberate, a one-line comment here would help future readers.

crates/marmot-app/src/media/blossom.rs
```diff
@@ -691,26 +699,48 @@ async fn media_http_client_for_url(
     build_pinned_media_http_client(pin)
 }
 
+async fn media_http_upload_client_for_url(
```
erskingardner, line 697: Nit: this is byte-identical to the now-`#[cfg(test)]` `media_http_client_for_url` above except for the final builder call. A single `media_http_client_for_url(url, allow_loopback_http, upload: bool)` (or a tiny shared body) would drop the duplication and the `#[cfg(test)]` attribute. Not worth blocking on.

crates/marmot-app/src/media/tests.rs
```diff
@@ -661,6 +719,48 @@ async fn blossom_fallback_reuses_the_identical_encrypted_body() {
     assert_eq!(first_body.recv().unwrap(), second_body.recv().unwrap());
 }
 
+#[tokio::test]
+async fn blossom_upload_allows_response_gap_longer_than_read_timeout() {
```
erskingardner, line 723: Nit: this regression only needs the first 64 KiB chunk plus the 16-second pause to cross the old read-idle boundary, so the 40,405,416-byte allocation here is fidelity to the production incident rather than a requirement. A ~128 KiB body would exercise the identical code path and save ~35 s of runtime plus two 40 MB allocations across the suite (the continuous-ingest test at line 744 genuinely needs the long ingest, so keeping its large body is fine). Keeping production-payload fidelity is also a defensible choice — just flagging the CI cost.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T09:04:57Z` |
| Commit reviewed | `793f34ad86b3d2d8a1bbab09216ec3eedb264420` |

## Summary

This is a well-scoped fix for #1585. The root cause is correctly identified: Blossom uploads inherited the shared client's 15-second read-idle deadline, so a server that keeps ingesting a large request body without emitting response bytes failed the upload even though the request itself carried a 15-minute total budget. The fix gives uploads a client policy without the read-idle timeout (connect and total bounds preserved), re-applies a 15-second bound to the small descriptor/error-body reads via a `tokio::time::timeout` wrapper, and threads a typed, privacy-safe `MediaUploadTimedOut` through the fallback, connector error mapping, and the wire protocol so Hermes can retry pre-publication failures.

Verified against the acceptance criteria in #1585:

- Upload client keeps `MEDIA_HTTP_CONNECT_TIMEOUT` (5s) and the per-request `MEDIA_BLOB_TRANSFER_TIMEOUT` (15 min) override; descriptor reads stay bounded (15s, size-capped).
- Download paths remain unchanged: `BlossomTransport::client_for_url` and `media_http_client_for_profile` still build with the read-idle timeout; `media_http_client_for_url` is now only reachable from the cfg(test) loopback fetch helper, so no production path lost its bound.
- URL validation, DNS/IP pinning, redirect refusal, proxy refusal, and content-size ceilings are untouched.
- Wire compatibility is handled in both directions and tested on both sides: `#[serde(default)] retryable` for old connectors, and the Hermes adapter honors the field only when it is a real bool with the `send_in_progress` fallback preserved.
- All three required regression tests are present under the names the issue asked for, including the 40,405,416-byte regressions crossing the former 15-second boundary.

Non-blocking notes are inline. One validation gap: CI was still queued/in-progress at review time, so I relied on the author's reported local verification (`just fast-ci` plus the targeted suites) rather than green remote checks.

The deferred immutable patch release matches the repo's manual-release invariant and is correctly left out of this commit.

crates/agent-connector/src/connection.rs
```diff
@@ -91,6 +91,7 @@ impl AgentConnector {
                         AgentControlResponse::Error {
                             code: "server_busy".to_owned(),
                             message: "agent connector subscription capacity is busy".to_owned(),
+                            retryable: true,
```
erskingardner, line 94: Worth an explicit acknowledgment: this also makes `server_busy` (subscription capacity, and the connection-capacity variant in `lib.rs`) retryable for clients that honor the new field. Under the Hermes fallback (`code == "send_in_progress"`), those codes were previously non-retryable. Capacity-busy errors are transient and retrying them is safe, so the direction is right — but it's a semantic widening bundled into the upload fix, and nothing asserts the intended contract. A one-line note in the PR body or a small test pinning `server_busy` → retryable would close the loop.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T09:16:08Z` |
| Commit reviewed | `793f34ad86b3d2d8a1bbab09216ec3eedb264420` |

## Summary

Clean, surgical fix for #1585. The upload path now gets a dedicated pinned client that keeps the 5s connect and 15-minute request bounds while dropping the shared 15s read-idle deadline that fired while a server was still ingesting the request body. The small descriptor/error-body reads stay bounded at 15s via a tokio timeout mapped to the new typed `AppError::MediaUploadTimedOut`, timeout identity survives endpoint fallback, and the server-owned `retryable` field threads through `agent-control` with a backwards-compatible serde default and a Hermes-side fallback to the legacy `send_in_progress` code check.

All #1585 source-side acceptance criteria are met: both deterministic `marmot-app` regressions spawn real slow-ingest servers (16s pause / ~18.5s continuous) with the observed 40,405,416-byte body; connector and Hermes tests assert the privacy-safe typed code, zero publications on the failed attempt, and exactly one publication after a same-artifact retry with reused staged paths and idempotency key. The `reqwest` timeout/connect/decode privacy mapping is preserved, and downloads keep their existing read-idle policy (`build_pinned_media_http_client` still applies `read_timeout`). Deferring the immutable `wn-agent` patch release to a manual release operation matches repo policy.

No blocking findings. Non-blocking notes:

- **CI gap:** clippy, Tamarin, and the build/package gates are green; GitHub `Rust check`/`Rust format`/`Rust tests` shards were still pending when this review was posted. The PR body's local `just fast-ci` plus targeted suites, and the six epoch-backfill failures being explicitly reproduced on clean `origin/master`, cover most of this, but the pending shards should be required to go green before merge.
- `crates/agent-connector/AGENTS.md` enumerates the `ConnectorError` projections as `code`/`client_message`/`privacy_safe_code`; the new `retryable()` projection could be added to keep the crate map accurate.

See the two inline notes for scope/tradeoff transparency (both are fine to land as-is).

crates/marmot-app/src/media/mod.rs
erskingardner, line 678: Tradeoff to keep explicit: in a mixed fallback, any timed-out server makes the aggregate error `MediaUploadTimedOut` even if another server definitively rejected (e.g. HTTP 4xx), and the collected per-server diagnostics in `failures` are discarded at that point. Biasing toward retry is the safe side pre-publication and implements the issue's fallback requirement as written; the cost is that a definitive rejection can burn Hermes' retry budget before surfacing. Fine as-is — flagging so the choice is deliberate rather than incidental.

crates/marmot-app/src/media/tests.rs
erskingardner: Keeping this unchanged because issue 1585 explicitly requires this regression to use the observed 40,405,416-byte payload. The production-size fidelity is intentional despite the local allocation cost.

crates/agent-connector/src/lib.rs
```diff
@@ -420,6 +420,7 @@ pub async fn serve_socket(config: AgentConnectorConfig) -> Result<(), ConnectorE
                 agent_control::AgentControlResponse::Error {
                     code: "server_busy".to_owned(),
                     message: "agent connector connection capacity is busy".to_owned(),
+                    retryable: true,
```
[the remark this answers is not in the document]
erskingardner: Verified the mismatch. Global connection admission occurs before any request frame is read, so the server does not yet possess an originating request ID; reading a frame there would weaken the connection-cap resource boundary. In 6098cff2 the uncorrelated connection-capacity response is therefore explicitly non-retryable, while the correlated post-read subscription-capacity response still echoes the request ID and remains retryable. Both contracts are now tested.

crates/agent-connector/src/connection.rs
erskingardner: Addressed in 6098cff2. Retryability is now pinned only for the correlated post-read subscription-capacity response. The pre-read global connection-capacity response is explicitly non-retryable because it cannot carry a request ID. Tests cover both cases.

crates/marmot-app/src/media/mod.rs
erskingardner: Addressed in 6098cff2 with an explicit comment documenting the deliberate retry-biased precedence: any timeout before publication keeps the aggregate safely retryable even if another fallback also rejected.

crates/marmot-app/src/media/blossom.rs
erskingardner: Addressed in 6098cff2 by factoring the shared validation, loopback classification, resolution, and pinning into resolve_pinned_media_host_for_url; the download-test and upload paths now differ only at client construction.

crates/marmot-app/src/media/mod.rs
erskingardner: Addressed in 6098cff2 with an explicit comment: any pre-publication timeout deliberately wins aggregate precedence so the result remains safely retryable, even when another fallback returned a terminal rejection.

PR #1655, opened 2026-09-03
fix: durably deduplicate replayed transport wrappers

### LINKED ISSUE #1650: cgka-engine: full-history redelivery beyond the rewind horizon strands an already-processed commit
## Summary

An offline founding member can catch up successfully from retained relay history, then an exact full-history replay of the same transport events strands an already-processed early commit as a new `PeelDeferred` row.

The client reaches the current epoch and receives the complete application backlog, but exact pending-work closure never succeeds. This was found by the new `offline-catchup-pressure/v1` family on current master `127db19cabdbc9d61692c8f9027f2ddf7c91ed66`.

## Reduced reproduction

The high-volume case reduces to commit depth; application messages are not required:

1. Alice creates a four-member group containing Bob.
2. Bob goes offline.
3. The online members publish six sequential `UpdateGroupData` commits, delivering and settling each round.
4. Bob reconnects.
5. Bob performs an incremental retained-history sync and tick.
6. Bob then performs an exact full-history sync of the same retained events and ticks again.
7. Observe Bob with the strict pending-work projection.

Five update rounds pass. Six update rounds reproduce.

The unreduced deterministic case is:

```text
generator_version=1
seed=9101
case_index=9
profile=medium-96/incremental-then-full
storage=file-backed SQLCipher
```

Across seeds 9101 through 9104, the same case classification reproduced.

## Observed

Bob reaches epoch 7 in the six-round reduced case and is otherwise settled, but reports:

```text
stored_transport_deferred_messages = 1
scenario_inputs_pending = 1
```

In the original eight-round case, the stranded ledger entry is the first workload commit:

```text
scenario_id = step-30:update_group_data
kind = commit
ingest_attempts = 4
ingest_accepted = 2
transport_deferred = 2
pending = true
```

## Root boundary

The outer transport id is the only cheap dedup key before peeling. Content-derived dedup is intentionally computed only after the wrapper decrypts. Once the source context has moved beyond the five-commit rewind/past-epoch horizon, replaying an exact wrapper cannot reach content dedup and enters the deferred-peel lifecycle.

This exact wrapper was already processed during the incremental sync. It must remain distinguishable from an unknown opaque event that may become peelable later. The resource-release behavior from #1193 must remain intact for genuinely unresolved events.

Relevant paths:

- `crates/cgka-engine/src/message_processor/mod.rs::do_ingest`
- `crates/cgka-engine/src/message_processor/ingest.rs::ingest_group_message_from_sweep`
- `crates/cgka-engine/src/message_processor/store.rs::recorded_message_outcome`
- `crates/cgka-conformance-simulator/src/retained_relay.rs::sync_one`

## Expected behavior

Exact redelivery of a transport wrapper that was already processed must classify as duplicate before an evicted peel context can turn it into pending work. Rewrapping the same MLS content in a genuinely new transport envelope must still use content-derived dedup after successful peeling.

## Acceptance criteria

- Add a focused file-backed retained-relay regression with incremental sync followed by exact full-history replay after at least six commits.
- The exact wrapper replay is terminally deduplicated and creates no `PeelDeferred` row.
- The client remains at the current canonical epoch with the expected payloads and `NoPendingWork`.
- The behavior survives engine restart between the two syncs.
- A new wrapper around the same MLS content still collapses through content-derived dedup when peelable.
- Genuinely unresolved opaque events preserve #1193 resource-release and later-redelivery eligibility semantics.

erskingardner:
## Summary

- persist exact processed group-wrapper IDs in a group-owned table that is not subject to the 4,096-row exceptional-ingress FIFO
- atomically admit a wrapper ID with its canonical peeled-content row, so a marker failure cannot leave a newly committed content row behind while returning an error
- keep malformed/unpeelable hostile group wrappers out of durable storage and retain the bounded marker pool for exceptional or unassociated terminal inputs
- add a file-backed four-member retained-relay regression spanning six commits, six payloads, restart, and exact full-history replay
- preserve content-derived dedup for fresh wrappers and the retryable resource-release semantics from #1193

## Root cause

Successfully peeled group messages are stored under a content-derived ID and return `Buffered`. The original ingest epilogue excluded every `Buffered` outcome from durable outer transport-ID evidence, leaving only a bounded in-memory prefilter for that wrapper. After restart, an exact old-wrapper replay could therefore reach the peeler again after its source context had fallen beyond the rewind horizon and become `PeelDeferred`.

The first fix reused `cgka_ingress_dedup`, but review correctly identified that table as an account-device-wide 4,096-row FIFO. Sending every processed group wrapper there would let ordinary group traffic evict the oldest correctness-critical IDs and reproduce #1650 for large backlogs.

## Storage contract

Migration `0058_processed_transport_ids` stores successfully peeled wrapper IDs separately and ties them to their owning protocol group with `ON DELETE CASCADE`. These rows are retained for the group's lifetime and are never evicted by the exceptional-ingress capacity policy. A transport ID cannot be reassigned to a different group.

When content is first admitted, the canonical `MessageRecord` and processed wrapper ID are written in the same storage transaction. Existing canonical content can acquire another exact-wrapper alias without rewriting the content row. Pre-peel malformed group wrappers enter neither this table nor the bounded exceptional marker pool.

## Test plan

- [x] New retained-history regression failed before the original fix with one deferred transport row and one pending scenario input
- [x] Processed wrapper IDs survive more than 4,096 exceptional-marker writes while the bounded marker pool still evicts oldest-first
- [x] Cross-group ID collision rolls back first canonical content admission
- [x] Pre-peel hostile wrapper IDs are not persisted; post-peel terminal wrappers are persisted
- [x] `cargo test -p cgka-traits`
- [x] `cargo test -p storage-sqlite`
- [x] `cargo test -p cgka-engine --features test-policy-overrides`
- [x] `cargo test -p cgka-conformance-simulator`
- [x] `just fast-ci`

Fixes #1650

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message deduplication so previously processed messages are not reprocessed after restarts.
  * Improved offline catch-up reliability, including complete history recovery, exact payload delivery, and consistent group state.
  * Improved handling of malformed and unsupported messages, including clearer duplicate classification and persistence behavior.

* **Documentation**
  * Added retained-relay workload scenarios covering offline recovery, ordering, restarts, competing updates, and convergence.
  * Documented the latest storage migration and message-processing behavior.


crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -516,10 +516,20 @@ impl<S: StorageProvider> Engine<S> {
...
+            // wrapper still collapses after peeling), but an exact wrapper
+            // replay may no longer be peelable once its source context falls
+            // outside the retained-history horizon (mdk#1650). Persist this
+            // only at the same terminal boundary used by the hot-process
+            // cache. Raw Retryable/PeelDeferred rows remain eligible because
+            // `should_remember_ingested_message` rejects them; a Buffered
+            // outcome whose durable row is instead keyed by peeled content
+            // has finished using this exact wrapper and is safe to remember.
+            if matches!(&msg.envelope, TransportEnvelope::GroupMessage { .. }) {
+                self.storage.put_ingress_dedup_marker(&msg.id)?;
```
erskingardner: **Blocking: this fix's durability is capped at 4,096 markers in the production backend.** `put_ingress_dedup_marker` in `storage-sqlite` inserts into `cgka_ingress_dedup` and then deletes every row outside the newest 4,096 (`INGRESS_DEDUP_MARKER_CAPACITY`, `crates/storage-sqlite/src/storage/messages.rs:13`). Two consequences:

1. A retained-history catch-up of more than 4,096 messages (account-device-wide, across all groups) evicts the oldest markers — including exactly the early-commit markers written here. An exact full-history replay of those wrappers then finds no marker and no row keyed by the transport id, re-enters the peeler past the rewind horizon, and strands `PeelDeferred` rows: mdk#1650 resurfaces for precisely the large-backlog workload this targets. The regression (seven messages) can't see this horizon.
2. Group-traffic churn now evicts the low-volume markers this table was built for (malformed/rejected welcomes, disband-tombstone hits), degrading their durable dedup to re-classification on redelivery.

Please either give this marker class its own retention (separate table/capacity, or eviction exempting group-wrapper markers), or scope the guarantee here and in the `MessageStorage` doc as bounded/best-effort with the >cap replay consequence stated. As written, "Keep the cheap outer transport-id prefilter restart-safe" claims a durability the storage contract doesn't provide.

erskingardner: Minor, non-blocking: this `?` turns a terminally-applied ingest into `Err` when the marker write fails. A caller retry of the same wrapper re-enters the peel path and can land `PeelDeferred` once the context has moved — the same stranding being fixed, under a transient storage failure. Degenerate case, but per the multi-step-state-changes invariant it deserves a deliberate choice (or at least a sentence) about whether a marker-write failure after a terminal outcome should fail open here.

crates/traits/src/storage.rs
```diff
@@ -269,9 +269,11 @@ pub trait MessageStorage {
     fn delete_pending_application_events(&self, ids: &[MessageId]) -> StorageResult<()>;
 
     /// Persist a terminal duplicate-detection marker for inbound protocol
```
erskingardner: The doc now covers group transport ids but still doesn't state effective-capacity semantics: the only production implementation silently FIFO-evicts markers beyond the newest 4,096. The engine now builds terminal dedup guarantees on this call (mdk#1650), so the trait should either promise durability for this marker class, or explicitly document markers as bounded/best-effort and eviction-safe so implementors and callers share one contract.

crates/cgka-engine/src/engine.rs
```diff
@@ -169,9 +169,9 @@ pub struct Engine<S: StorageProvider> {
 
     /// MessageIds the engine has ingested. Backs the typed duplicate exclusion.
     ///
-    /// Bounded hot-process cache in front of the authoritative durable
-    /// `MessageRecord` store (checked first in `do_ingest`), so an id that ages
-    /// out of the cap is still classified by storage.
+    /// Bounded hot-process cache in front of the authoritative durable ingress
```
erskingardner: Wording nit: "in front of the authoritative durable ingress markers" reads cache-first, but `do_ingest` consults `recorded_message_outcome` (markers, then `MessageRecord` store) *before* `seen_message_ids` — the parenthetical "(checked first in `do_ingest`)" was accurate for the old comment's subject but now sits ambiguously between two subjects. Also, "authoritative" is doing a lot of work for a store that FIFO-evicts at 4,096; if the bound stands, this should say bounded rather than imply unbounded authority.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T11:13:29Z` |
| Commit reviewed | `f349ab150cbb69f235099042051d9bb5109bb42f` |

**What's good:** the root-cause diagnosis is correct (the `Buffered` exclusion kept content-keyed rows out of the durable outer marker, so exact replay past the rewind horizon re-entered the deferred-peel lifecycle), the change is surgical, retryable/`PeelDeferred` semantics from #1193 are preserved via `should_remember_ingested_message`, and the regression covers all six acceptance criteria including restart between syncs. CI is green.

**Blocking issue:** the "durable" marker this fix rests on is a bounded FIFO in the production backend. `put_ingress_dedup_marker` inserts into `cgka_ingress_dedup` and then silently deletes every row outside the newest 4,096 (`INGRESS_DEDUP_MARKER_CAPACITY`, `crates/storage-sqlite/src/storage/messages.rs:13`). Before this PR that table only received low-volume, non-group-associatable material; after it, *every* terminal group-message ingest writes a marker. A retained-history catch-up larger than 4,096 messages (account-device-wide, across all groups) evicts the oldest markers — including exactly the early-commit markers this fix writes — so an exact full-history replay of those wrappers still finds no marker and no transport-keyed row, re-enters the peeler past the horizon, and re-strands `PeelDeferred` rows: mdk#1650 resurfaces for precisely the large-backlog offline-catchup workload it targets. The seven-message regression cannot exercise this horizon.

Either give this marker class retention that survives the cap (separate table/capacity, or eviction that exempts group-wrapper markers), or explicitly scope the guarantee as bounded/best-effort in the trait contract and the epilogue comment, with the >cap replay consequence spelled out. Details inline.

**Minor:** the `engine.rs` comment now reads cache-first although `do_ingest` checks the durable layer first, and "authoritative" overstates a silently-evicting store; the marker write's `?` can also convert a terminally-applied ingest into `Err` on a transient write failure (see inline notes).

crates/cgka-engine/src/message_processor/mod.rs
erskingardner: This now routes every successfully ingested group message through `cgka_ingress_dedup`, but the SQLite implementation caps that table at `INGRESS_DEDUP_MARKER_CAPACITY = 4_096` rows **account-device-wide** (`storage-sqlite/src/storage/messages.rs:13`), evicting oldest-first. Before this change, markers were written only for rare cases (welcomes, malformed/rejected material, disband tombstones), so 4096 slots were effectively unbounded; now the table churns once per group message across all groups on the account.

The eviction order works against the fix. The markers that matter are exactly the oldest ones — wrappers whose peel context has fallen beyond the rewind horizon — while the newest markers protect wrappers that content-derived dedup would catch anyway after peeling. Once an account has ingested more than 4096 group messages since a wrapper was first processed, a restart + exact full-history replay strands that wrapper as `PeelDeferred` again: the #1650 behavior returns, just at a higher volume threshold. The offline-catchup-pressure family that found #1650 operates at volumes where 4096 account-wide ingests is reachable, so this should be resolved rather than assumed away — e.g. size/justify the capacity against the retained-history horizon those scenarios exercise, make the cap per-group, or persist a terminal outer-id row in the (unbounded) message store instead of the bounded marker table. At minimum the bound belongs in the trait docs now that correctness depends on it.

Minor, same hunk: if `put_ingress_dedup_marker` fails after the message was fully processed, the `?` reports the ingest as failed even though its effects are already durable, and the seen-cache insert is skipped. Self-healing via content dedup on the next replay, so not blocking — but worth a note.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T11:16:30Z` |
| Commit reviewed | `f349ab150cbb69f235099042051d9bb5109bb42f` |

## Summary

The root-cause analysis matches the code: `Buffered` outcomes store their canonical row under the content-derived id, so the outer transport id had no durable terminal marker, and an exact wrapper replay beyond the rewind horizon fell into the deferred-peel lifecycle. The fix is surgical and in the right place — the epilogue boundary already excludes genuinely retryable work (`retryable_unpersisted`, `Created`/`Retryable`/`PeelDeferred` rows via `should_remember_ingested_message`), so #1193 resource-release semantics are preserved, and fresh wrappers around the same content still collapse through content-derived dedup after peeling (`ingest.rs` content-id seam is untouched). Internal sweep replay bypasses the outer seam by design, so callers that only label outcomes (`ingest_outcome_kind`) are unaffected by redelivered pending wrappers now classifying as `Duplicate` instead of `Buffered`.

The regression is exactly what the issue asked for: file-backed SQLite, four members, six commits + six payloads, restart between incremental and exact full-history sync, and assertions on dedup classification, zero transport deferrals, epoch 7, payloads, and `NoPendingWork`. CI is fully green.

## Blocking-grade concern

One serious concern, posted inline: the durable marker table is bounded at 4096 rows account-device-wide with oldest-first eviction — a bound sized when markers were rare (welcomes/malformed). This PR makes every ingested group message consume a slot, and oldest-first eviction discards precisely the markers protecting beyond-horizon wrappers, so the #1650 behavior returns once >4096 messages were ingested since the original wrapper. The capacity question (or a per-group/unbounded alternative) should be resolved or explicitly justified before merge.

## Minor

- The engine.rs comment update is accurate; CodeRabbit's outstanding nit about stating lookup order is already covered by "checked first in `do_ingest`" — no action needed there.
- Error-path asymmetry on marker-write failure (inline): self-healing, but worth a comment.

Datawav (commented):
No actionable findings.

Outcome: The exact-head patch resolves the durable wrapper-replay defect without introducing a concrete changed-line defect. The advisory finding is stale: at this head, `crates/cgka-engine/src/message_processor/ingest.rs:848-854` uses `persist_openmls_wire_message_with_processed_transport_id` for the unsupported MLS-body path, preserving the exact transport ID with the failed content record. The storage helper writes both records transactionally.

Audit evidence:
- Exact head SHA: b9035b76a074e193b438e025f79c2a0cc887b431
- Base SHA: 1fd198be6f97d057226efd11bc08982cc07f9297
- Patch coverage is complete with no missing file patches.
- Changed files examined include `crates/cgka-engine/src/message_processor/ingest.rs`, `crates/cgka-engine/src/message_processor/store.rs`, `crates/storage-sqlite/src/storage/messages.rs`, and `crates/storage-sqlite/src/migrations/0057_processed_transport_ids.rs`.
- Exact-wrapper IDs now use `cgka_processed_transport_ids`, which is separate from the account-wide, 4,096-row `cgka_ingress_dedup` cache. IDs remain group-owned and are removed by the group foreign-key cascade.
- First content admission and wrapper-ID admission are one transaction; existing content records also cause the successfully peeled wrapper ID to be recorded.
- The retained-relay regression exercises four members, six sequential commits and payloads, file-backed restart, and exact full-history replay beyond the rewind horizon.

Discussion classification:
- The old FIFO-capacity concern is resolved by the separate uncapped, group-owned processed-ID table; repeated comments and reviews asserting continued use of `cgka_ingress_dedup` are stale duplicates.
- The old post-commit marker-write race is resolved by transactional content/transport-ID persistence.
- The trait-capacity and cache-order documentation comments are addressed by the revised trait and engine documentation.
- The unsupported-MLS-body claim is stale at this head because that branch now uses the processed-transport-aware persistence helper.
- The remaining automated discussion reported no current actionable comment; stylistic or superseded observations were discarded.

Risk lenses checked:

Check evidence:
- Supplied checks are complete. Rust check, Rust format, Rust clippy, Rust audit, all four Rust test shards, Simulator and vectors, Tamarin proofs, Relay CLI E2E, and CLI e2e completed successfully.
- `Relay runtime tests` completed with a failure conclusion. The bundle contains no failure details tying that result to a changed line, so it does not support an actionable code finding.
- `Publish wn-agent release` was skipped.
- `automated_device_gate` is null. No physical-device evidence exists for this head; conclusions rest on static review and the supplied CI evidence.

Not approved on this head: merge conflict with the base branch; failed CI: Relay runtime tests (failure). A corrected head is re-audited automatically.

Datawav (approved):

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T15:28:31Z` |
| Commit reviewed | `03eec40f25c96a016b3d4e7e1acef7c3e96eaa8b` |

Re-review after `5eccdcb`/`b9035b76` and the master merge. All three findings from my previous review are resolved, and the rework is the right architecture:

**1. Bounded-FIFO eviction (was blocking) — fixed properly.** Exact-wrapper evidence moved out of the 4,096-row `cgka_ingress_dedup` pool into a new `cgka_processed_transport_ids` table (`id` PK, FK to `cgka_groups` `ON DELETE CASCADE`, group index) retained for the owning group's lifetime. The epilogue marker write was reverted, so the bounded pool is again exclusive to hostile/unassociated input, with a test asserting a never-peeled malformed wrapper enters neither table. `processed_transport_ids_survive_marker_capacity_churn_and_cascade` covers the eviction horizon my 7-message regression couldn't, plus group-delete cascade. Rows are minted only after a successful authenticated peel, so the unbounded-per-group growth is member-traffic-bounded and documented in `storage-format-v2.md`.

**2. Trait contract — fixed.** `put_ingress_dedup_marker` is now documented as bounded/evictable for unassociated material, and `put_processed_transport_id` carries an explicit must-survive-until-group-delete contract plus the same-transaction requirement when co-admitted with the canonical row.

**3. Comment wording and torn marker write — fixed.** The `engine.rs` comment now accurately says the hot cache sits behind durable processed ids, bounded markers, and the record store. First content admission and the wrapper alias ride one `with_transaction` (with a rollback test on cross-group conflict); the standalone alias write on the content-dedup-hit path is retry-safe since the content row predates it.

Verified at head: all five peel seams write the alias (content-dedup hit, malformed MLS, unsupported MLS body, pre-membership, fresh `Created` admission) plus both branches of the convergence-buffer handoff; `PeelDeferred`/retryable semantics and `#1193` resource release are untouched; the master merge correctly renumbered the migration to `0058` with registration, FK test, and doc updated consistently; CI is green on `03eec40`.

Only residual note (non-blocking, no action requested): each ingest now pays one extra indexed `has_processed_transport_id` lookup before the existing checks — negligible against the correctness it buys.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T15:31:40Z` |
| Commit reviewed | `03eec40f25c96a016b3d4e7e1acef7c3e96eaa8b` |

(Note: posted as a comment because GitHub does not allow approving one's own PR; the recommendation is merge.)

## Re-review summary

The follow-up commits (`5eccdcd`, `b9035b7`) fully resolve the blocking concern from my earlier review. Verified at head:

- **Capacity concern resolved the right way.** Exact wrapper ids moved out of the bounded, account-wide `cgka_ingress_dedup` FIFO into a new unbounded, group-scoped `cgka_processed_transport_ids` table (migration 0058) with `ON DELETE CASCADE`. The trait docs now state the bounded/evictable contract for ingress markers explicitly, and `storage-format-v2.md` documents the split. The storage test `processed_transport_ids_survive_marker_capacity_churn_and_cascade` proves both halves of the contract.
- **Torn-write minor also resolved.** Wrapper id and canonical content row are admitted in one `with_transaction` unit (`persist_stored_message_payload`), including the convergence-buffer handoff; the group-conflict rollback test covers the failure path. This matches the repo's torn-write invariant.
- **Hostile-input asymmetry handled.** Never-peeled attacker wrappers get no durable state (hot cache only, idempotent re-peel after restart), while every post-peel terminal seam — content-dedup hit, malformed MLS bytes, non-message body, pre-membership, Created admission, convergence handoff — records the wrapper id. `terminal_peel_rejection_ignored` deliberately stays memory-only for pre-peel rejections; replay re-derives the same verdict.
- **Semantics preserved.** PeelDeferred/Retryable content remains internally retryable via stored-row sweeps that bypass the outer seam; #1193 eligibility for genuinely unresolved opaque events is untouched since they never reach a peel seam.
- **Merge hygiene.** The master-merge migration collision was resolved by renumbering the branch migration to 0058 with registration, schema-version tests (`found: 58`), and the FK regression list all consistent. No workspace/vector version bumps in the branch commits.
- The original `exact_full_history_replay_after_restart_deduplicates_processed_wrappers` regression is intact, and `post_peel_invalid_mls_inputs_are_terminal_and_restart_deduplicated` extends coverage to restart + replay of the unsupported-body branch.

CI is fully green at the reviewed head. Non-blocking observation only: the new table grows one small row per peeled group message for the group's lifetime — proportional to the already-unbounded `cgka_messages` retention and justified in the doc, so no action needed.

PR #1648, opened 2026-09-03
Read one group without loading the whole projection

dannym-arx:
### Summary

Opening a conversation reads one group's projection row, and a received message marks only its own group's chat-list row for refresh.

### Context

group() went through groups(), which loaded the entire account projection: every group with components and direct members plus up to 16,384 seen-event ids, then searched for one id. The subscription layer fires the same call on every group event. Separately, any received message marked the whole chat-list projection stale, and the next chat_list() rebuilt every row inside a write transaction.

### What changed

A keyed load_account_group fetches one row with its components and members, and group() uses it. The engine-owned lifecycle fields are stamped by the helper groups() also uses. The stale marker becomes a set of dirty group ids per account; a warm projection refreshes only those rows through the existing per-row refresh, and a cold one still takes the full ensure path.

### Impact

Opening a conversation reads three keyed queries instead of eight table scans. A received message costs one row refresh instead of a rebuild of every row. Counted from the code, not timed.

## Summary by CodeRabbit

- **Performance**
  - Chat lists now refresh only affected groups, improving responsiveness when updates are limited to specific conversations.
  - Opening an individual group loads its details more efficiently, especially for accounts with many groups.

- **Bug Fixes**
  - Pending group updates are preserved when a chat-list refresh fails or overlaps with newer changes.
  - Direct membership changes now correctly refresh the affected group in chat lists.
  - Group details and related members remain consistent when loaded individually or as part of a full list.

PR #1656, opened 2026-09-03
Store the group record as MessagePack instead of JSON

dannym-arx:
### Summary

The group record in `cgka_groups` is written as MessagePack, and rows written before this change still decode.

### Context

Every record was a serde_json document. Each member's id and credential are byte vectors, which serde_json spells as arrays of decimal integers with a text integer parse per byte on every read. After #1646 moved the OpenMLS value store off JSON, this parse was half of a warm 64-member send, because the engine reads the record several times per send and per inbound message.

### What changed

New rows carry the same two-byte NUL prefix `openmls_values` uses, followed by MessagePack with structs as maps so fields added later with serde defaults stay readable. Prefix-less rows decode as JSON and move to the new encoding on their next write, so no rewrite is needed. Snapshot capture and restore go through the same codec. A no-op migration marks the version so an older build refuses the database up front.

### Impact

Warm app-message send, in-memory SQLCipher, on top of #1648: 16 members 555 to 398 µs, 64 members 887 to 531 µs. Inbound 64-member app message 369 to 275 µs.

### DIFF
--- a/crates/storage-sqlite/src/migrations.rs
+++ b/crates/storage-sqlite/src/migrations.rs
@@ -110,6 +110,8 @@ mod migration_0054_transport_reconciliation_items;
 mod migration_0055_epoch_stall_evidence;
 #[path = "migrations/0056_chat_list_accepted_activity_high_water.rs"]
 mod migration_0056_chat_list_accepted_activity_high_water;
+#[path = "migrations/0057_openmls_values_cbor_encoding.rs"]
+mod migration_0057_openmls_values_cbor_encoding;
 #[cfg(test)]
 #[path = "migrations/test_support.rs"]
 mod test_support;
@@ -404,6 +406,11 @@ const MIGRATIONS: &[Migration] = &[
         version: 56,
         name: "0056_chat_list_accepted_activity_high_water",
         apply: migration_0056_chat_list_accepted_activity_high_water::apply,
+    },
+    Migration {
+        version: 57,
+        name: "0057_openmls_values_cbor_encoding",
+        apply: migration_0057_openmls_values_cbor_encoding::apply,
     },
 ];
 
@@ -932,7 +939,7 @@ mod tests {
         assert!(matches!(
             error,
             StorageError::UnsupportedSchemaVersion {
-                found: 56,
+                found: 57,
                 latest_supported: 46,
             }
         ));
@@ -988,7 +995,7 @@ mod tests {
         assert!(matches!(
             error,
             StorageError::UnsupportedSchemaVersion {
-                found: 56,
+                found: 57,
                 latest_supported: 46,
             }
         ));
@@ -1292,7 +1299,7 @@ mod tests {
         assert!(matches!(
             error,
             StorageError::UnsupportedSchemaVersion {
-                found: 56,
+                found: 57,
                 latest_supported: 46,
             }
         ));
--- a/crates/storage-sqlite/src/migrations.rs
+++ b/crates/storage-sqlite/src/migrations.rs
@@ -112,6 +112,8 @@ mod migration_0055_epoch_stall_evidence;
 mod migration_0056_chat_list_accepted_activity_high_water;
 #[path = "migrations/0057_openmls_values_msgpack.rs"]
 mod migration_0057_openmls_values_msgpack;
+#[path = "migrations/0058_group_record_msgpack.rs"]
+mod migration_0058_group_record_msgpack;
 #[cfg(test)]
 #[path = "migrations/test_support.rs"]
 mod test_support;
@@ -411,6 +413,11 @@ const MIGRATIONS: &[Migration] = &[
         version: 57,
         name: "0057_openmls_values_msgpack",
         apply: migration_0057_openmls_values_msgpack::apply,
+    },
+    Migration {
+        version: 58,
+        name: "0058_group_record_msgpack",
+        apply: migration_0058_group_record_msgpack::apply,
     },
 ];
 
@@ -939,7 +946,7 @@ mod tests {
         assert!(matches!(
             error,
             StorageError::UnsupportedSchemaVersion {
-                found: 57,
+                found: 58,
                 latest_supported: 46,
             }
         ));
@@ -995,7 +1002,7 @@ mod tests {
         assert!(matches!(
             error,
             StorageError::UnsupportedSchemaVersion {
-                found: 57,
+                found: 58,
                 latest_supported: 46,
             }
         ));
@@ -1299,7 +1306,7 @@ mod tests {
         assert!(matches!(
             error,
             StorageError::UnsupportedSchemaVersion {
-                found: 57,
+                found: 58,
                 latest_supported: 46,
             }
         ));
--- a/crates/storage-sqlite/src/migrations.rs
+++ b/crates/storage-sqlite/src/migrations.rs
@@ -110,8 +110,8 @@ mod migration_0054_transport_reconciliation_items;
 mod migration_0055_epoch_stall_evidence;
 #[path = "migrations/0056_chat_list_accepted_activity_high_water.rs"]
 mod migration_0056_chat_list_accepted_activity_high_water;
-#[path = "migrations/0057_openmls_values_cbor_encoding.rs"]
-mod migration_0057_openmls_values_cbor_encoding;
+#[path = "migrations/0057_openmls_values_msgpack.rs"]
+mod migration_0057_openmls_values_msgpack;
 #[cfg(test)]
 #[path = "migrations/test_support.rs"]
 mod test_support;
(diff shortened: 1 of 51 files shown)


Datawav (commented):
## Automated review

Two actionable findings.

1. Major — Chat-list invalidation can be lost before refresh completes

Exact changed-line evidence:
  crates/marmot-app/src/lib.rs, changed ensure_chat_list_rows_warm path:
  `let stale = self.chat_list_projection_stale.lock().unwrap().remove(&account.label);`
  followed by:
  `Some(group_ids) if warmed => {`
  `    for group_id_hex in group_ids {`
  `        storage.refresh_chat_list_row(&account.account_id_hex, &group_id_hex, &classifier)?;`
  `    }`
  `}`
  and:
  `None if warmed => return Ok(())`

Triggering state: `account_storage(...)`, SQLite, classifier, or any per-row refresh returns an error after `remove` consumes the dirty-group set. The related `stale == Some(_) && warmed == false` state also falls through to `ensure_chat_list_rows(...)`, which does not preserve the prior rebuild obligation.

Consequence: completed and unprocessed invalidations become indistinguishable. A later call can take `None if warmed => return Ok(())`, leaving stale previews or unread counts indefinitely; a partially completed loop loses its remaining group IDs.

Confidence: High.

Smallest sound correction: clone or retain the stale set until all refreshes succeed. Remove IDs only after each successful row refresh, and reinsert the unprocessed remainder on every error path. Match `(stale, warmed)` explicitly so every `Some(_)` state performs refresh semantics rather than the weaker ensure path.

2. Major — MlsGroup cache generation does not observe writes from another database connection

Exact changed-line evidence:
  crates/storage-sqlite/src/connection.rs:
  `openmls_writes: AtomicU64`
  with `note_openmls_write()` and `openmls_write_generation()` backed by that per-SharedConnection counter.

  crates/traits/src/storage.rs:
  `fn mls_write_generation(&self) -> Option<u64> {`
  `    None`
  `}`

  crates/cgka-engine/src/mls_group_cache.rs reuses an entry when:
  `cached.generation == generation`

Triggering state: another process or independently opened SQLite connection commits an OpenMLS-state write to the same database. Its commit does not increment the cached engine’s in-memory AtomicU64.

Consequence: the engine can reuse an MlsGroup loaded from older persistent state, risking incorrect epoch or ratchet state on subsequent send, receive, or group-read operations.

Confidence: High.

Smallest sound correction: include SQLite’s foreign-commit signal, such as `PRAGMA data_version`, in the cache generation and invalidate when either component changes. If exclusive-writer ownership cannot be proven, return `None` from `mls_write_generation()` so this cache is disabled for that backend.

Audit evidence:
  Exact head SHA: 7ec03ba886b46ab0e548e8c0d0c1e7473ad6fd8c
  Base SHA: ea58338403c35d7e876e3dfca9bbb4fccd060ef4
  Patch coverage is marked complete across 50 changed files, including crates/marmot-app/src/lib.rs, crates/cgka-engine/src/mls_group_cache.rs, and crates/storage-sqlite/src/connection.rs.
  Discussion classification: one non-actionable automated status comment and no organization-member feedback. Its request to retrigger automation is not a code finding.
  No physical-device evidence exists for this head; automated_device_gate is null. Conclusions therefore rest on static patch/context review and the supplied check record.

Risk lenses checked: security/auth; persistence/migration; concurrency/lifecycle; network/protocol; release/build; tests

Check evidence: the bundle marks check evidence available and complete, but supplies zero runs; all_completed=false and all_successful=false. Therefore it contains no CI run result supporting a test-success claim.

crates/storage-sqlite/src/migrations.rs
```diff
@@ -405,6 +409,16 @@ const MIGRATIONS: &[Migration] = &[
         name: "0056_chat_list_accepted_activity_high_water",
         apply: migration_0056_chat_list_accepted_activity_high_water::apply,
     },
+    Migration {
+        version: 57,
+        name: "0057_openmls_values_msgpack",
+        apply: migration_0057_openmls_values_msgpack::apply,
+    },
+    Migration {
+        version: 58,
```
erskingardner, line 418: [P1] Allocate a migration number after the current master

Current master (`44e43eb70f1f51b753bbc57363d7fd82b366196e`) already assigns schema version 58 to `0058_processed_transport_ids` from #1655. This PR assigns the same durable version to a different migration/name. Once the stack is rebased, both histories cannot safely claim version 58: an installation that recorded one name/version must never have another build interpret that number as a different migration. Rebase onto live master through the lower stack, rename this file/module to the next available migration number (currently 59), and update the future-version assertions accordingly.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T16:48:33Z` |
| Commit reviewed | `7ec03ba886b46ab0e548e8c0d0c1e7473ad6fd8c` |
| Model | `gpt-5.6-sol` |
| Reasoning level | `high` |

## Summary

The intended group-record MessagePack slice is small and the legacy-JSON/read + snapshot paths are covered, but the live PR cannot land in its current shape. Its target branch is now `f2e8eab86fccb653a17ae8ab48abe64f05ac81dc` while this head still descends from the older `06f5118...` stack, so GitHub reports `DIRTY`, the PR diff contains eight commits across 50 files rather than the intended five-file slice, and no exact-head CI suite exists. Rebase the stack first and address the migration-number collision below. The reviewed head itself is validly signed.

Datawav (commented):
## Automated review

Changes required. Three actionable findings remain after re-deriving the advisory claims against exact head `7ec03ba886b46ab0e548e8c0d0c1e7473ad6fd8c`.

1. Major — MLS cache can reuse state after an external database write
   Evidence: `crates/storage-sqlite/src/connection.rs` adds a per-`SharedConnection` `openmls_writes: AtomicU64`; `note_openmls_write()` increments only that counter. `crates/cgka-engine/src/mls_group_cache.rs` reuses an entry when its saved generation equals this process-local value. The changed storage contract requires invalidation for every OpenMLS write, while nearby connection code explicitly anticipates separate connections/processes.
   Triggering state: another connection or process commits OpenMLS state to the same SQLite/WAL database while this engine retains a cached `MlsGroup`.
   Consequence: the engine can operate on stale epoch or ratchet state, causing incorrect protocol processing or persistent state divergence.
   Confidence: High.
   Smallest sound correction: include a cross-connection change signal such as SQLite `PRAGMA data_version` in cache validity, or return `None` from `mls_write_generation()` unless exclusive database ownership is guaranteed. Recheck the generation before returning a mutated group to the cache.

2. Major — Chat-list staleness is cleared before fallible refresh work
   Evidence: in `crates/marmot-app/src/lib.rs`, the changed `ensure_chat_list_warm` path obtains pending groups with `chat_list_projection_stale.lock().unwrap().remove(&account.label)` before looping over `storage.refresh_chat_list_row(...)?`; the former post-success removal is deleted.
   Triggering state: any row refresh returns an error after the stale set has been removed—for example, transient SQLite contention.
   Consequence: unprocessed group IDs lose their refresh obligation, so later `chat_list()` calls may return stale durable rows without retrying the repair.
   Confidence: High.
   Smallest sound correction: inspect or clone the pending set without removing it, refresh all rows, and remove the completed obligation only after success; alternatively reinsert the unprocessed remainder on every error path.

3. Major — Unrelated workspace coverage and dependency declarations are removed
   Evidence: root `Cargo.toml` changed lines delete `unicode-normalization = "0.1"`, `bech32 = "0.11"`, `scrypt = "0.11"` and workspace paths for `convergence-campaign-runner`, `agent-connector`, `marmot-c`, and `wn-cli`, while the PR describes group-record encoding. No corresponding member-manifest removals are present in the 50-file patch, and `Cargo.lock` adds entries without removing the deleted dependencies.
   Triggering state: a workspace build or consumer still relying on one of those inherited dependencies, or CI expecting one of the removed members to remain covered.
   Consequence: manifest resolution can break, or required crates can silently disappear from workspace-wide checks.
   Confidence: High.
   Smallest sound correction: restore these unrelated lines in this head; handle intentional workspace pruning separately with all consumer manifests and the lockfile updated together.

Discussion classification: the maintainer’s migration-version-58 collision is a rebase-time concern, not an exact-head defect demonstrated by this bundle: this patch assigns versions 57 and 58 uniquely, and the conflicting master migration is not in the bounded source context. The broader member summary is supported insofar as this head is `dirty`, spans 50 files, and contains cache, transaction, statement-cache, and chat-list changes beyond the stated MessagePack scope. The automated skipped-review notice supplies no code finding.

Audit evidence: immutable bundle for full exact head `7ec03ba886b46ab0e548e8c0d0c1e7473ad6fd8c`; changed files inspected include `crates/cgka-engine/src/mls_group_cache.rs`, `crates/storage-sqlite/src/connection.rs`, `crates/marmot-app/src/lib.rs`, `crates/storage-sqlite/src/storage/groups.rs`, and `Cargo.toml`.

Risk lenses checked: security/auth; persistence/migration; concurrency/lifecycle; network/protocol; release/build; tests.

Check evidence: the bundle marks check evidence complete but contains zero runs, with `all_completed: false` and `all_successful: false`; therefore it supplies no exact-head CI success. The automated device gate is `capability_checked_not_installable`, so no physical-device evidence exists for this head. Conclusions rest on static review and the supplied check metadata.

PR #1657, opened 2026-09-03
Read the group record once per send gate

dannym-arx:
### Summary

A send reads the group record three times instead of eight.

### Context

Two of the reads served the audit trail alone: recipient expectations load the record and the admin policy to describe the send even when the recorder is the no-op default. Four came from the removed and unrecoverable gates, which each re-read the record at acceptance and again before preparation.

### What changed

Recipient-expectation rows are skipped when no recorder consumes them. Both terminal gates are answered from one record read at each of the two gate sites, keeping the check order so a removed copy still reports Removed when it is also halted. The two sites stay separate because convergence runs between them.

The MlsGroup cache test from #1647 never exercised the send path: its sends queued behind an unconfirmed founding publish, and every hit it counted came from the audit load this change skips. It now confirms the creation and asserts each send produced an application message.

### Impact

Warm app-message send on top of the MessagePack record: 0 members 256 to 254 µs, 64 members 531 to 407 µs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved outbound message validation for removed groups and groups requiring unrecoverable-state repair.
  - Ensured these conditions return consistent, actionable errors before messages are queued or sent.
  - Synchronized durable unrecoverable status with group state to prevent sending from halted groups.

- **Diagnostics**
  - Recipient-expectation audit records are now generated only when forensic recording is enabled.

### DIFF
--- a/crates/cgka-engine/src/engine.rs
+++ b/crates/cgka-engine/src/engine.rs
@@ -1015,17 +1015,22 @@ impl<S: StorageProvider> Engine<S> {
                     context: Some(context.clone()),
                     kind: crate::audit_helpers::send_outcome_event(intent_kind, send_result),
                 });
-                for (msg_id, expectation) in
-                    self.recipient_expectation_records(&recipient_group_id, send_result)
-                {
-                    self.recorder.record(AuditRecord {
-                        group_ref: group_ref.clone(),
-                        context: Some(context.clone()),
-                        kind: AuditEventKind::RecipientExpectation {
-                            msg_id,
-                            expectation,
-                        },
-                    });
+                // Recipient expectations read the group record and admin
+                // policy just to describe the send; skip that work when no
+                // recorder consumes it.
+                if self.recorder.is_enabled() {
+                    for (msg_id, expectation) in
+                        self.recipient_expectation_records(&recipient_group_id, send_result)
+                    {
+                        self.recorder.record(AuditRecord {
+                            group_ref: group_ref.clone(),
+                            context: Some(context.clone()),
+                            kind: AuditEventKind::RecipientExpectation {
+                                msg_id,
+                                expectation,
+                            },
+                        });
+                    }
                 }
             }
             Err(err) => {
@@ -2261,6 +2266,44 @@ impl<S: StorageProvider> Engine<S> {
         self.epoch_manager
             .restore_unrecoverable(group_id.clone(), group.epoch);
         Ok(true)
+    }
+
+    /// The removed and unrecoverable gates every send passes at acceptance and
+    /// again before preparation, answered from one group-record read instead
+    /// of one per gate. Order matches the split checks: a removed copy reports
+    /// `Removed` even when it is also halted.
+    pub(crate) fn refuse_removed_or_halted(
+        &mut self,
+        group_id: &GroupId,
+        intent: &SendIntent,
+    ) -> Result<(), EngineError> {
+        let group = match self.storage.get_group(group_id) {
+            Ok(group) => group,
+            Err(StorageError::NotFound) => return Ok(()),
+            Err(err) => return Err(EngineError::Storage(err)),
+        };
+        if group.removed {
+            return Err(EngineError::InvalidTransition(
+                cgka_traits::engine_state::InvalidTransition {
+                    from: "Removed",
+                    to: crate::audit_helpers::send_intent_kind_str(intent),
+                    reason: "local group copy is marked removed (self-evicted)",
+                },
+            ));
+        }
+        if self.epoch_manager.is_unrecoverable(group_id) {
+            return Err(EngineError::GroupUnrecoverableRepairRequired {
+                group_id: group_id.clone(),
+            });
+        }
+        if !group.unrecoverable {
+            return Ok(());
+        }
+        self.epoch_manager
+            .restore_unrecoverable(group_id.clone(), group.epoch);
+        Err(EngineError::GroupUnrecoverableRepairRequired {
+            group_id: group_id.clone(),
+        })
     }
 
     /// Clear ONLY the live OpenMLS group state for `group_id`, leaving every
--- a/crates/cgka-engine/src/message_processor/mod.rs
+++ b/crates/cgka-engine/src/message_processor/mod.rs
@@ -624,29 +624,15 @@ impl<S: StorageProvider> Engine<S> {
                 },
             ));
         }
-        // Terminal gate before queueing: a local copy marked removed (realized
-        // self-eviction) must never accept or queue outbound work. Checked
-        // again in `do_send_ready` so queued-intent drains for a copy removed
-        // after queueing hit the same deterministic error.
-        if self.group_record_is_removed(&group_id)? {
-            return Err(EngineError::InvalidTransition(
-                cgka_traits::engine_state::InvalidTransition {
-                    from: "Removed",
-                    to: crate::audit_helpers::send_intent_kind_str(intent),
-                    reason: "local group copy is marked removed (self-evicted)",
-                },
-            ));
-        }
-        // mdk#971: durable Unrecoverable halt — sync into memory and refuse
-        // outbound work until a verified repair path clears the marker.
-        if self.sync_unrecoverable_halt_from_storage(&group_id)? {
-            // mdk#1177: typed, not `InvalidTransition`. A halted group is a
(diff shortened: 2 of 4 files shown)


Datawav (commented):
## Automated review

One actionable finding.

1. Minor — In-memory unrecoverable state is bypassed when the group record is missing

Exact changed-line evidence: In the added `refuse_removed_or_halted` helper in `crates/cgka-engine/src/engine.rs` (added hunk beginning at line 2268), the code first executes:

    let group = match self.storage.get_group(group_id) {
        Ok(group) => group,
        Err(StorageError::NotFound) => return Ok(()),
        ...
    };

Only afterward does it check `self.epoch_manager.is_unrecoverable(group_id)`. The replaced `sync_unrecoverable_halt_from_storage` logic checked the in-memory marker before reading storage.

Triggering state: A queued drain enters `do_send_ready` directly while `epoch_manager` marks the group unrecoverable but `storage.get_group(group_id)` returns `NotFound`, such as during partial-state loss or interrupted persistence recovery.

Consequence: The terminal gate returns `Ok(())` instead of `GroupUnrecoverableRepairRequired`, allowing processing to continue until a later operation produces a different error. This changes the durable-halt behavior and typed recovery signal precisely under the routed partial-failure lens.

Confidence: High. Both the old ordering and new early-return ordering are present in the supplied patch and bounded source context.

Smallest sound correction: Preserve removed-before-halted precedence for existing records, but change the `NotFound` arm to return `GroupUnrecoverableRepairRequired` when `epoch_manager.is_unrecoverable(group_id)` is true; otherwise return `Ok(())`.

Audit evidence:
- Exact head: a0f1f41cf144cff962d92fc131bd9338a20bf2a7
- Changed files inspected: `crates/cgka-engine/src/engine.rs`, `crates/cgka-engine/src/message_processor/mod.rs`, `crates/cgka-engine/src/message_processor/send.rs`, and `crates/cgka-engine/src/mls_group_cache.rs`
- The advisory’s duplication concern is maintainability-only; no behavioral divergence is demonstrated.
- The `ForensicRecorder::is_enabled()` concern is unsupported: the supplied trait contract defines it as whether the recorder consumes events, defaults custom recorders to `true`, and only `NoopRecorder` returns `false`.
- Missing precedence tests do not independently establish a defect; the changed helper still checks `group.removed` before the durable record’s `unrecoverable` flag.
- Discussion classification: the sole item is a non-member automated status comment with no substantive review finding; its embedded retry instruction and unsupported benchmark claims were discarded.

Risk lenses checked:
- persistence/migration
  - upgrade and rollback from real prior state
  - partial failure and idempotent retry
  - data preservation

No schema or migration change was present. The actionable regression is confined to recovery behavior when persisted group state is missing while an in-memory terminal marker remains; no additional data-loss or retry-side-effect defect was demonstrated.

Check evidence: The bundle reports complete CI evidence with all four Rust test shards, Rust check, Rust clippy, Rust format, Rust audit, Tamarin proofs, simulator/vectors, CLI E2E, and the listed build/package checks concluding success. `Relay runtime tests` concluded `failure`; the bundle supplies no log excerpt tying that result to the changed lines, so it is not converted into a duplicate code finding.

The bundle’s `automated_device_gate` is null. No physical-device evidence exists for this head; conclusions therefore rest on static review and the supplied CI evidence.

crates/cgka-engine/src/engine.rs
```diff
@@ -2263,6 +2268,44 @@ impl<S: StorageProvider> Engine<S> {
...
+    /// of one per gate. Order matches the split checks: a removed copy reports
+    /// `Removed` even when it is also halted.
+    pub(crate) fn refuse_removed_or_halted(
+        &mut self,
+        group_id: &GroupId,
+        intent: &SendIntent,
+    ) -> Result<(), EngineError> {
+        let group = match self.storage.get_group(group_id) {
+            Ok(group) => group,
+            Err(StorageError::NotFound) => return Ok(()),
```
erskingardner, line 2282: [P1] Check the in-memory unrecoverable state on `NotFound`

The replaced `sync_unrecoverable_halt_from_storage` checked `epoch_manager.is_unrecoverable(group_id)` before reading the record. This new early return therefore changes a direct send against an in-memory `Unrecoverable` group with a missing/corrupt durable group row from the typed `GroupUnrecoverableRepairRequired` result to `Ok(())`; `should_queue_outbound_intent` then declines to queue the terminal state and the send continues until a deeper, unrelated storage/OpenMLS error. Preserve Removed-before-halted precedence when a record exists, but make the `NotFound` arm return the typed repair-required error whenever the in-memory halt is present. Add a regression for this partial-state case.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T16:49:03Z` |
| Commit reviewed | `a0f1f41cf144cff962d92fc131bd9338a20bf2a7` |
| Model | `gpt-5.6-sol` |
| Reasoning level | `high` |

## Summary

Combining the two durable send gates and suppressing no-op audit-recipient work are sensible, but the combined helper changes the terminal error contract when the in-memory state is already `Unrecoverable` and the durable group row is missing. Fix that ordering before merge. The exact-head suite also has the two relay-runtime failures inherited from the lower projection slice (#1648); all other checks passed, and this head is validly signed.

Datawav (commented):
## Automated review

One major actionable finding remains.

1. Major — Missing durable group row bypasses the in-memory unrecoverable halt

Exact changed-line evidence: `crates/cgka-engine/src/engine.rs:2282` adds `Err(StorageError::NotFound) => return Ok(())` in `refuse_removed_or_halted`, before `self.epoch_manager.is_unrecoverable(group_id)` is consulted. The replaced `sync_unrecoverable_halt_from_storage` checked the in-memory state first. Both `validate_send_acceptance` and `do_send_ready` now use the new helper.

Triggering state: the epoch manager marks the group `Unrecoverable`, while `storage.get_group(group_id)` returns `StorageError::NotFound` because the durable row is absent or partially lost.

Consequence: the helper returns success instead of `EngineError::GroupUnrecoverableRepairRequired`. Direct sends and queued-intent drains can continue past the terminal gate and surface an unrelated storage/OpenMLS error, violating the typed halt contract and partial-state retry behavior.

Confidence: High. The ordering regression is present in the patch and is confirmed against the prior memory-first helper and both changed call sites.

Smallest sound correction: in the `NotFound` arm, return `GroupUnrecoverableRepairRequired` when `epoch_manager.is_unrecoverable(group_id)` is true; otherwise return `Ok(())`. Add regression coverage for both direct send acceptance and direct `do_send_ready` drain with an in-memory unrecoverable state and no durable group row.

Audit evidence:
Exact head: `a0f1f41cf144cff962d92fc131bd9338a20bf2a7`
Changed files examined: `crates/cgka-engine/src/engine.rs`, `crates/cgka-engine/src/message_processor/mod.rs`, `crates/cgka-engine/src/message_processor/send.rs`, and `crates/cgka-engine/src/mls_group_cache.rs`. Patch coverage is complete with no missing patch files.

Risk lenses checked: `persistence/migration` — upgrade and rollback from real prior state; partial failure and idempotent retry; data preservation. The finding is specifically a partial-state/data-preservation regression. Removed-before-halted precedence remains intact when a group record exists, and durable unrecoverable restoration remains intact for existing records.

Discussion classification:
- Member inline P1: supported and retained as the major finding above.
- Member review summary: its ordering defect is supported. Its claims that relay results were inherited and that the head was validly signed are not established by the bundle.
- Automated create-group recorder suggestion: describes unchanged code and a performance-only opportunity; discarded as out of scope.
- Automated walkthrough: descriptive only; no additional actionable defect.
- Advisory cache-test variant concern: current changed test behavior does not establish a concrete product defect; discarded as speculative test brittleness.
- Benchmark, storage-read-count, and base-branch inheritance claims: unsupported by supplied artifacts.

Check evidence:
The bundle reports complete check data with all runs completed. Rust tests 1/4 through 4/4, Rust check, audit, clippy and format, Tamarin proofs, simulator/vectors, C bindings smoke, CLI E2E, and the listed build/package checks concluded `success`. `Relay runtime tests` concluded `failure`; the bundle does not establish its cause or inheritance. `Publish wn-agent release` was skipped.

The automated device gate is `capability_checked_not_installable`: no physical-device evidence exists for this head. Conclusions therefore rest on the complete static patch/source audit and supplied CI evidence.

PR #1660, opened 2026-09-03
Fix partial directory subscription sync state

### LINKED ISSUE #943: marmot-app: directory subscription sync is non-atomic with the SDK — a partial subscribe_with_id_to failure leaves subscriptions live but their validation filter never committed, silently dropping their events
**Severity:** MEDIUM
**Component:** `crates/marmot-app`

## Summary

`sync_directory_user_subscriptions` issues SDK subscriptions in a loop but commits the in-memory validation-filter map (`active_subscriptions`, the sole gate `accepts_live_event` checks) only **after** the whole loop succeeds. If any `subscribe_with_id_to` in the loop errors, the function returns early via `?`, leaving already-created SDK subscriptions live and delivering events whose filter was never inserted into `active_subscriptions`. Those events then fail the membership gate and are silently dropped.

## Location

`crates/marmot-app/src/relay_plane/mod.rs:445-502` (`sync_directory_user_subscriptions`):

- `to_remove` are unsubscribed on the SDK (`:446-451`).
- Each `to_add` batch is subscribed via `subscribe_with_id_to(...).await.map_err(...)?` (`:490-499`) — **fallible**, early-returns on the first failure.
- The in-memory `active_subscriptions` map is committed only at `:502` via `replace_subscriptions(desired)`.

The forwarder gate is `spawn_relay_notification_forwarder` (`mod.rs:689`): each event that fails `directory.accepts_live_event(...)` (unknown subscription id → `false`) is dropped.

## Failure scenario

1. A directory sync plans several `to_add` batches (e.g. spread across relays).
2. One relay rejects/errors on `subscribe_with_id_to`, so the loop returns at `:499` before `:502`.
3. Subscriptions created *before* the failing one are now live on the SDK and delivering profile/relay-list/key-package events, but their `(authors, kinds)` filter was never inserted into `active_subscriptions`.
4. In the forwarder, each such event fails `accepts_live_event` (unknown id → `false`) and is dropped (`mod.rs:701-707`), so valid directory updates never reach the cache.
5. It self-heals only on a later *fully successful* sync; while the plan keeps failing, updates are dropped. There is no compensation — already-created subs are not torn down.

**#910** is a transport-adapter REQ-before-route ordering bug in a different crate. No existing issue covers the relay-plane directory-subscription map being non-atomic with the SDK subscribe calls.

## Suggested fix

Commit the filter for each subscription as it is successfully created (insert into `active_subscriptions` inside the loop, immediately after a successful `subscribe_with_id_to`), or build/commit `desired` incrementally so the map can never lag the live SDK state. On error, either compensate (unsubscribe the subs created so far) or leave the committed filters in place so their events are still accepted.

---
<sub>Filed from a meticulous multi-agent code-review pass; verified against source and deduplicated against the existing open/closed issue set.</sub>

erskingardner:
## Summary
- record each directory validation filter as soon as its SDK subscription succeeds, so a later batch failure cannot leave live events failing the admission gate
- preserve lifecycle accounting while the final successful sync still replaces the complete desired filter map

## Test plan
- [x] `cargo test -p marmot-app --lib relay_plane::tests::`
- [x] `just fast-ci`
- [ ] `cargo test -p marmot-app` (882 passed; 6 unrelated epoch-backfill tests failed, with an exact failure reproduced on unmodified `origin/master`)

Fixes #943

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved directory subscription synchronization so successfully created subscriptions are retained when a later batch fails validation.
  - Preserved active subscription filters during synchronization, preventing valid subscriptions from being unintentionally lost.
  - Added more accurate tracking and reporting of newly created subscriptions and synchronization results.
- **Tests**
  - Added coverage for preserving valid subscriptions and filters after subsequent synchronization errors.

### DIFF
--- a/crates/marmot-app/src/relay_plane/mod.rs
+++ b/crates/marmot-app/src/relay_plane/mod.rs
@@ -1010,6 +1010,7 @@ impl MarmotRelayPlane {
         // into the directory cache when it matches an active subscription's
         // requested authors and kinds (mdk#709).
         let mut desired = HashMap::with_capacity(plan.batches.len());
+        let mut subscriptions_created = 0;
         for batch in &plan.batches {
             let authors = batch
                 .authors
@@ -1028,10 +1029,9 @@ impl MarmotRelayPlane {
             // Canonical lowercase hex matches the `event.pubkey` form a forwarded
             // SDK event carries, so the membership check is exact.
             let filter_authors = authors.iter().map(PublicKey::to_hex).collect::<Vec<_>>();
-            desired.insert(
-                batch.subscription_id.clone(),
-                DirectorySubscriptionFilter::new(filter_authors, batch.kinds.clone()),
-            );
+            let validation_filter =
+                DirectorySubscriptionFilter::new(filter_authors, batch.kinds.clone());
+            desired.insert(batch.subscription_id.clone(), validation_filter.clone());
             if !to_add.contains(&batch.subscription_id) {
                 continue;
             }
@@ -1052,9 +1052,18 @@ impl MarmotRelayPlane {
                 )
                 .await
                 .map_err(|err| format!("directory subscription subscribe: {err}"))?;
+            subscriptions_created += usize::from(
+                self.inner
+                    .directory
+                    .record_subscription_filter(batch.subscription_id.clone(), validation_filter)
+                    .await,
+            );
         }
 
-        self.inner.directory.replace_subscriptions(desired).await
+        self.inner
+            .directory
+            .complete_subscription_sync(desired, subscriptions_created)
+            .await
     }
 
     pub async fn shutdown(&self) {
--- a/crates/marmot-app/src/directory/mod.rs
+++ b/crates/marmot-app/src/directory/mod.rs
@@ -23,4 +23,6 @@ pub use search::{
     OFF_GRAPH_SEARCH_RADIUS, SearchUpdateTrigger, UserSearchParams, UserSearchSubscription,
     UserSearchUpdate, sort_user_search_results,
 };
+#[cfg(test)]
+pub(crate) use sync::DirectorySyncBatch;
 pub(crate) use sync::{DirectorySyncHandle, DirectorySyncPlan};
--- a/crates/marmot-app/src/relay_plane/directory.rs
+++ b/crates/marmot-app/src/relay_plane/directory.rs
@@ -267,6 +267,48 @@ impl DirectoryRelayPlane {
         (to_add, to_remove)
     }
 
+    /// Record the validation filter immediately after its SDK subscription is
+    /// created, so a later batch failure cannot leave that live subscription
+    /// unknown to [`Self::accepts_live_event`].
+    pub(crate) async fn record_subscription_filter(
+        &self,
+        subscription_id: String,
+        filter: DirectorySubscriptionFilter,
+    ) -> bool {
+        let mut state = self.state.lock().await;
+        let created = state
+            .active_subscriptions
+            .insert(subscription_id, filter)
+            .is_none();
+        if created {
+            state.subscriptions_created += 1;
+        }
+        created
+    }
+
+    /// Complete a subscription sync whose newly created filters were already
+    /// recorded as their SDK subscriptions succeeded.
+    pub(crate) async fn complete_subscription_sync(
+        &self,
+        desired: HashMap<String, DirectorySubscriptionFilter>,
+        subscriptions_created: usize,
+    ) -> Result<DirectorySubscriptionSyncSummary, String> {
+        let mut state = self.state.lock().await;
+        let removed = state
+            .active_subscriptions
+            .keys()
+            .filter(|id| !desired.contains_key(*id))
+            .count();
+        state.completed_subscription_syncs += 1;
+        state.subscriptions_removed += removed;
+        state.active_subscriptions = desired;
+        Ok(DirectorySubscriptionSyncSummary {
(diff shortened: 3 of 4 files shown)


crates/marmot-app/src/relay_plane/mod.rs
```diff
@@ -1052,9 +1052,18 @@ impl MarmotRelayPlane {
...
+                    .directory
+                    .record_subscription_filter(batch.subscription_id.clone(), validation_filter)
+                    .await,
+            );
         }
 
-        self.inner.directory.replace_subscriptions(desired).await
+        self.inner
+            .directory
+            .complete_subscription_sync(desired, subscriptions_created)
```
erskingardner, line 1065: Non-blocking, and pre-existing (not introduced by this PR): on the failure path, `to_remove` ids were already unsubscribed on the SDK (before the batch loop) but remain in `active_subscriptions` until a later successful sync. If a failed sync is followed by a plan that re-includes one of those ids, `subscription_diff` sees it as already active and never re-subscribes on the SDK — the map claims it is live while no events can ever arrive. That is the mirror image of #943 and never self-heals while the id stays in the map.

The record-as-you-go pattern introduced here extends naturally: drop the filter at unsubscribe time as well, so the map never claims a dead subscription. Worth a follow-up issue.

crates/marmot-app/src/relay_plane/mod.rs
```diff
@@ -1010,6 +1010,7 @@ impl MarmotRelayPlane {
         // into the directory cache when it matches an active subscription's
         // requested authors and kinds (mdk#709).
         let mut desired = HashMap::with_capacity(plan.batches.len());
+        let mut subscriptions_created = 0;
         for batch in &plan.batches {
```
erskingardner, line 1014: Non-blocking: batch authors/kinds are validated inside this loop, after earlier batches have already mutated SDK state (and after the `to_remove` unsubscribes). A validation pass over `plan.batches` hoisted above all SDK mutations (per the AGENTS.md multi-step invariant: validate before mutating) would eliminate most partial-failure states — e.g. in the new test, the invalid-author batch would be rejected before the first subscription is ever issued. The current fix is still correct given the issue-endorsed 'leave committed filters in place' semantics, so this is a follow-up consideration, not a blocker.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T11:53:42Z` |
| Commit reviewed | `ac6f53f5df2552eeb1250ceb47dcfde534e60fe5` |

## Summary

This correctly fixes #943 using the record-as-you-go option the issue itself endorses: each successful `subscribe_with_id_to` immediately commits its validation filter, so a later batch failure can no longer leave a live SDK subscription unknown to `accepts_live_event` (the silent event-drop). The successful path still ends with the complete desired map, so end state is unchanged from before.

The accounting is sound: `record_subscription_filter` owns the created counter (first insert only — no double counting), and `complete_subscription_sync` owns removals and the completed-sync counter. `replace_subscriptions` retains a caller (the empty-plan early return), so there is no dead code. The regression test reproduces the exact issue scenario and asserts both the live SDK subscription and the committed filter, and the `DirectorySyncBatch` re-export is properly `#[cfg(test)]`-gated.

Two non-blocking notes inline, both pre-existing behavior rather than regressions from this change.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T11:57:16Z` |
| Commit reviewed | `ac6f53f5df2552eeb1250ceb47dcfde534e60fe5` |
| Reasoning level | `high` |

*(Posted as a comment because GitHub does not allow approving your own PR; the recommendation is approve/merge.)*

## Summary

This is the right fix for #943, and it is the minimal version of it. Recording each validation filter immediately after its `subscribe_with_id_to` succeeds closes the exact hole the issue describes — previously a mid-loop failure left live SDK subscriptions whose filters were never committed, so their events were silently dropped by `accepts_live_event` until a fully successful sync.

Two design points worth affirming:

- **Recording only on confirmed SDK success (rather than committing `desired` on the error path) is the correct choice.** Committing `desired` optimistically would invert the bug: the map would claim subscriptions the SDK never created, and the next `subscription_diff` would skip re-subscribing them — a worse, permanent failure. This PR avoids that trap.
- **Accounting stays consistent.** `record_subscription_filter` returning `created` keeps `force_rebuild` (which re-subscribes already-active ids) from double-counting, matching the old `replace_subscriptions` semantics. On the success path `complete_subscription_sync` computes `removed` as `active - desired`, which exactly equals the `to_remove` set since every recorded id is in `desired`.

The regression test is genuinely tied to the bug: before the fix, the error path never touched `active_subscriptions`, so the `accepts_live_event` assertion fails; after, it passes. The `#[cfg(test)]` re-export of `DirectorySyncBatch` is necessary (`mod sync;` is private) and minimal. CI is green across the full matrix.

## Non-blocking observations (no action required for this PR)

1. **Residual micro-window:** an event arriving between `subscribe_with_id_to` resolving and `record_subscription_filter` acquiring the lock would still be dropped. This is a few instructions versus the previous whole-loop window, and recording *before* subscribing would introduce the inverted bug above — not worth changing.
2. **Pre-existing, out of scope:** the empty-plan early return (`mod.rs:961`, `replace_subscriptions(HashMap::new())`) clears the validation map without unsubscribing the live SDK subscriptions, so relays keep streaming events that are then all dropped. Same file, same atomicity theme, but untouched by this PR — candidate for a follow-up issue if not already tracked.

PR #1662, opened 2026-09-03
fix(simulator): preserve bounded convergence continuation

### LINKED ISSUE #1651: Conformance HarnessClient turns bounded convergence continuation into a backend error after eight passes
## Summary

The conformance `HarnessClient` treats a legitimate bounded convergence continuation as a backend failure after eight internal drain passes. Large offline catch-up cases therefore abort instead of yielding control and allowing the next scenario tick to continue the scheduled work.

This blocks reliable testing of the production-shaped repeated worker re-arm behavior and makes large retained-history cases look like engine convergence failures.

Found by the new `offline-catchup-pressure/v1` family on current master `127db19cabdbc9d61692c8f9027f2ddf7c91ed66`.

### Natural retained history

- Four-member group; Bob offline.
- 384 application messages across 12 sequential update commits.
- Bob reconnects and performs one natural full-history sync.
- The family grants an `Eventually(PayloadCount == 1, max_iterations = 128)` boundary.

The first catch-up tick advances Bob only to epoch 9 and delivers 281 payloads. Four commit rounds remain. Rather than leaving those scheduled for the next assertion tick, `HarnessClient::drive_due_convergence` emits:

```text
EngineError::Backend(
    "convergence drain did not settle within harness pass limit"
)
```

The scenario step fails before `Eventually` can perform its next iteration.

### Competing commit waves

A reduced case needs only:

- eight rounds of two same-epoch rival commits;
- one application message immediately before the eighth pair;
- reverse retained-history delivery with three copies.

Seven rounds settle. Eight rounds leave Bob at epoch 8 with an active convergence pass while the group tip is epoch 9. Even 128 outer progress turns are unavailable because the first tick that reaches the internal pass limit becomes a hard error.

## Root cause

`crates/cgka-conformance-simulator/src/client.rs` defines:

```rust
const HARNESS_CONVERGENCE_DRAIN_PASSES: usize = 8;
```

After those passes, `drive_due_convergence` unconditionally pushes:

```rust
EngineError::Backend(
    "convergence drain did not settle within harness pass limit"
)
```

`EngineHarnessSubject::tick` then passes the outcomes through `ensure_tick_succeeded`, converting the continuation into a failed scenario step.

The engine already exposes scheduled/pending state. Production workers re-arm while work remains; bounded slicing is not itself an internal error.

## Expected behavior

A tick should remain bounded, but exhausting its local work slice with valid scheduled continuation should return successfully and preserve pending work for a later tick. A distinct no-progress or invariant violation may still fail loudly.

## Acceptance criteria

- Keep a deterministic per-tick work bound.
- Exhausting the bound with valid pending convergence work does not synthesize `EngineError::Backend`.
- A following scenario tick can continue and eventually settle the large natural-history reproduction.
- The eight-wave competing-commit reproduction can be driven across repeated ticks, allowing the oracle to determine whether the engine actually converges.
- Add a no-progress guard so a true scheduler spin remains a typed failure.
- Preserve existing errors from convergence, publication, and storage operations.
- Add regressions at seven/eight internal-pass boundaries and a 384-message retained-history catch-up.

erskingardner:
## Summary
- keep the harness convergence drain bounded to eight passes per tick
- allow a later tick to continue when the bounded slice made durable structural progress
- retain a typed failure for a full re-armed slice with unchanged durable state, while preserving underlying engine and publication errors
- compare a clock-stable progress projection that excludes timing/derived scheduling fields and normalizes the one-shot scheduled edge
- make multi-group drain and guard processing deterministic, assess only groups that consume the full slice, and treat unavailable diagnostic snapshots as fail-open

Fixes #1651

## Test plan
- [x] Drive `HarnessClient::tick` through seven and exactly eight scheduled passes
- [x] Drive a scripted 384-work-item, twelve-round drain across two ticks (8 + 4)
- [x] Exercise scheduled-edge normalization with an unchanged re-arm spin
- [x] Verify clock-only snapshot deltas do not count as durable progress
- [x] Verify unavailable initial and final guard assessments do not synthesize tick failures
- [x] Verify an underlying convergence error is preserved verbatim
- [x] `cargo test -p cgka-conformance-simulator --locked`
- [x] `just fast-ci`

## Validation note
The historical natural-history reproducer no longer enters the reported synthetic-backend branch on current `master`; it separately quiesces at Bob epoch 9 with 281 payloads. The current eight-wave competing family arm likewise enters its long outer fallback instead of providing a fast isolated boundary regression. The new tests therefore feed deterministic retained-work schedules through the real `HarnessClient::tick` and changed drain loop, covering schedule draining, attempt accounting, clock-stable structural comparison, two-tick continuation, the true-spin guard, diagnostic-assessment failure, and underlying error preservation without adding a multi-minute unstable family case.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved convergence checks by ignoring clock-only and one-time scheduling changes when assessing durable progress.
  - Prevented unavailable diagnostic snapshots from causing errors after otherwise successful convergence processing.
  - Preserved convergence continuation with retained history.
  - Improved handling of failed progress assessments during scripted processing.

- **Tests**
  - Added coverage for clock-only changes, failed progress checks, unavailable diagnostics, and retained-history continuation.

### DIFF
--- a/crates/cgka-conformance-simulator/src/client.rs
+++ b/crates/cgka-conformance-simulator/src/client.rs
@@ -80,8 +80,108 @@ pub struct HarnessClient {
     /// observe decisions no `GroupEvent` exposes (e.g. `convergence_decision`).
     audit_capture: AuditCapture,
     convergence_checkpoints: HashMap<String, (GroupId, DurableConvergencePass)>,
+    #[cfg(test)]
+    scripted_convergence_drain: Option<ScriptedConvergenceDrain>,
 }
 
+#[cfg(test)]
+struct ScriptedConvergenceDrain {
+    group_id: GroupId,
+    remaining_retained_messages: usize,
+    remaining_passes: usize,
+    attempted_passes: usize,
+    generation: u64,
+    scheduled: bool,
+    advance_generation: bool,
+    fail_on_attempt: Option<usize>,
+}
+
+#[cfg(test)]
+impl ScriptedConvergenceDrain {
+    /// Builds bounded retained-history work for harness-level tick tests.
+    fn retained_history(retained_messages: usize, passes: usize) -> Self {
+        Self {
+            group_id: GroupId::new(b"scripted-convergence-drain".to_vec()),
+            remaining_retained_messages: retained_messages,
+            remaining_passes: passes,
+            attempted_passes: 0,
+            generation: 1,
+            scheduled: passes > 0,
+            advance_generation: true,
+            fail_on_attempt: None,
+        }
+    }
+
+    /// Builds a scheduler that re-arms forever without changing durable state.
+    fn stalled() -> Self {
+        Self {
+            advance_generation: false,
+            ..Self::retained_history(0, usize::MAX)
+        }
+    }
+
+    /// Builds a drain whose underlying convergence operation fails verbatim.
+    fn failing(attempt: usize) -> Self {
+        Self {
+            fail_on_attempt: Some(attempt),
+            ..Self::retained_history(attempt, attempt)
+        }
+    }
+
+    /// Drains the scripted one-shot schedule edge.
+    fn drain_pending_groups(&mut self) -> Vec<GroupId> {
+        if !self.scheduled {
+            return Vec::new();
+        }
+        self.scheduled = false;
+        vec![self.group_id.clone()]
+    }
+
+    /// Captures the same privacy-safe fields used by the real engine seam.
+    fn progress(&self) -> ConformanceStructuralProgressSnapshot {
+        let pending_work = cgka_engine::conformance_snapshot::ConformancePendingWorkSnapshot {
+            scheduled_convergence_groups: usize::from(self.scheduled),
+            ..Default::default()
+        };
+        ConformanceStructuralProgressSnapshot {
+            current_epoch: self.generation,
+            current_monotonic_ms: HARNESS_CONVERGENCE_SETTLED_AT_MS,
+            lifecycle: cgka_traits::engine_state::GroupLifecycleState::Stable,
+            pending_work,
+            pass_generation: Some(self.generation),
+            pass_phase: Some(ConvergencePassPhase::Resolving),
+            earliest_next_wake_monotonic_ms: None,
+            runnable_work: 1,
+            terminal_unrecoverable: false,
+        }
+    }
+
+    /// Executes one scripted convergence round and re-arms when work remains.
+    fn converge(&mut self) -> Result<Vec<SendResult>, EngineError> {
+        self.attempted_passes = self.attempted_passes.saturating_add(1);
+        if self.fail_on_attempt == Some(self.attempted_passes) {
+            return Err(EngineError::Backend(
+                "scripted underlying convergence failure".into(),
+            ));
+        }
+        if self.advance_generation {
+            let messages_this_pass = self
+                .remaining_retained_messages
+                .div_ceil(self.remaining_passes);
+            self.remaining_retained_messages = self
+                .remaining_retained_messages
+                .saturating_sub(messages_this_pass);
+            self.remaining_passes = self.remaining_passes.saturating_sub(1);
+            self.generation = self.generation.saturating_add(1);
+            self.scheduled = self.remaining_passes > 0;
+        } else {
+            self.scheduled = true;
+        }
+        Ok(Vec::new())
+    }
+}
+
+/// Saturating-merge one engine metrics snapshot into the harness aggregate.
 pub(crate) fn merge_engine_metrics(
     target: &mut EngineMetricsSnapshot,
     source: &EngineMetricsSnapshot,
@@ -185,6 +285,7 @@ pub(crate) fn merge_engine_metrics(
     );
 }
 
+/// Saturating-merge histogram buckets while retaining upper-bound order.
 fn merge_histogram(target: &mut HistogramSnapshot, source: &HistogramSnapshot) {
     for source_bucket in &source.buckets {
         if let Some(target_bucket) = target
@@ -201,6 +302,12 @@ fn merge_histogram(target: &mut HistogramSnapshot, source: &HistogramSnapshot) {
     target.overflow_count = target.overflow_count.saturating_add(source.overflow_count);
 }
 
+/// Rejects only an exact full-slice repeat of scheduled structural state.
+///
+/// A durable pass-generation change counts as progress. More complex cycles
+/// remain the scenario fixed-point driver's responsibility; this local guard
+/// exists to catch a scheduler that repeatedly re-arms the same state without
+/// turning the per-tick work bound into a convergence deadline.
 fn ensure_convergence_drain_progress(
     attempts: usize,
     continuation_scheduled: bool,
@@ -217,12 +324,13 @@ fn ensure_convergence_drain_progress(
 }
 
 #[cfg(test)]
-mod metric_tests {
+mod tests {
     use super::*;
     use cgka_engine::conformance_snapshot::ConformancePendingWorkSnapshot;
     use cgka_engine::engine_metrics::HistogramBucket;
     use cgka_traits::engine_state::GroupLifecycleState;
 
+    /// Builds a single-bucket histogram fixture.
     fn histogram(upper_bound: u64, count: u64, overflow_count: u64) -> HistogramSnapshot {
         HistogramSnapshot {
             buckets: vec![HistogramBucket { upper_bound, count }],
@@ -230,6 +338,7 @@ mod metric_tests {
         }
     }
 
+    /// Every engine metric participates in the aggregate merge.
     #[test]
     fn merge_engine_metrics_covers_every_snapshot_field() {
         let source = EngineMetricsSnapshot {
@@ -273,6 +382,7 @@ mod metric_tests {
         assert_eq!(target, source);
     }
 
+    /// An unchanged, re-armed eighth attempt is a local scheduler spin.
     #[test]
     fn eight_repeated_convergence_drain_states_are_a_typed_failure() {
         let progress = structural_progress(7);
@@ -290,6 +400,7 @@ mod metric_tests {
         );
     }
 
+    /// Seven unchanged attempts remain below the deterministic slice bound.
     #[test]
     fn seven_repeated_convergence_drain_states_can_continue() {
         let progress = structural_progress(7);
@@ -303,6 +414,7 @@ mod metric_tests {
         .expect("the deterministic slice still has one attempt left");
     }
 
+    /// Advancing the durable pass generation permits cross-tick continuation.
     #[test]
     fn changed_convergence_drain_state_can_continue_after_eight_attempts() {
         let previous = structural_progress(7);
@@ -317,6 +429,7 @@ mod metric_tests {
         .expect("a later pass generation is structural progress");
     }
 
+    /// Twelve retained-history rounds can span an eight- plus four-pass slice.
     #[test]
     fn large_retained_history_can_continue_in_a_later_bounded_slice() {
         // The 384-message natural-history case advances through twelve commit
@@ -338,6 +451,96 @@ mod metric_tests {
         .expect("the following tick can settle the retained history");
     }
 
+    /// Seven and eight scheduled passes both complete in one harness tick.
+    #[tokio::test]
+    async fn harness_tick_covers_seven_and_eight_pass_boundaries() {
+        for passes in [
+            HARNESS_CONVERGENCE_DRAIN_PASSES - 1,
+            HARNESS_CONVERGENCE_DRAIN_PASSES,
+        ] {
+            let mut client =
+                scripted_client(ScriptedConvergenceDrain::retained_history(passes, passes));
+
+            let outcomes = client.tick().await;
+
+            assert!(outcomes.iter().all(Result::is_ok), "{outcomes:?}");
+            let script = client.scripted_convergence_drain.as_ref().unwrap();
+            assert_eq!(script.attempted_passes, passes);
+            assert_eq!(script.remaining_passes, 0);
+            assert_eq!(script.remaining_retained_messages, 0);
+            assert!(!script.scheduled);
+        }
+    }
+
+    /// A 384-message, twelve-round drain continues on a second harness tick.
+    #[tokio::test]
+    async fn harness_tick_continues_large_retained_history_after_eight_passes() {
+        let mut client = scripted_client(ScriptedConvergenceDrain::retained_history(384, 12));
+
+        let first = client.tick().await;
+
+        assert!(first.iter().all(Result::is_ok), "{first:?}");
+        let script = client.scripted_convergence_drain.as_ref().unwrap();
+        assert_eq!(script.attempted_passes, HARNESS_CONVERGENCE_DRAIN_PASSES);
+        assert_eq!(script.remaining_passes, 4);
+        assert_eq!(script.remaining_retained_messages, 128);
+        assert!(script.scheduled);
+
+        let second = client.tick().await;
+
+        assert!(second.iter().all(Result::is_ok), "{second:?}");
+        let script = client.scripted_convergence_drain.as_ref().unwrap();
+        assert_eq!(script.attempted_passes, 12);
+        assert_eq!(script.remaining_passes, 0);
+        assert_eq!(script.remaining_retained_messages, 0);
+        assert!(!script.scheduled);
+    }
+
+    /// The same tick path retains a typed failure for an unchanged re-arm loop.
+    #[tokio::test]
+    async fn harness_tick_rejects_a_true_scheduler_spin() {
+        let mut client = scripted_client(ScriptedConvergenceDrain::stalled());
+
+        let outcomes = client.tick().await;
+
+        assert!(matches!(
+            outcomes.as_slice(),
+            [Err(EngineError::Backend(message))]
+                if message.contains("without structural progress")
+        ));
+        assert_eq!(
+            client
+                .scripted_convergence_drain
+                .as_ref()
+                .unwrap()
+                .attempted_passes,
+            HARNESS_CONVERGENCE_DRAIN_PASSES
+        );
+    }
+
+    /// A real convergence error is not replaced by the slice guard.
+    #[tokio::test]
+    async fn harness_tick_preserves_underlying_convergence_error() {
+        let mut client = scripted_client(ScriptedConvergenceDrain::failing(8));
+
+        let outcomes = client.tick().await;
+
+        assert!(matches!(
+            outcomes.as_slice(),
+            [Err(EngineError::Backend(message))]
+                if message == "scripted underlying convergence failure"
+        ));
+    }
+
+    /// Wraps a scripted drain in a real `HarnessClient` tick surface.
+    fn scripted_client(script: ScriptedConvergenceDrain) -> HarnessClient {
+        let bus = TransportBus::ordered();
+        let mut client = ClientBuilder::new(b"scripted-drain-client".to_vec()).attach(&bus);
+        client.scripted_convergence_drain = Some(script);
+        client
+    }
+
+    /// Builds a privacy-safe structural state for drain-boundary tests.
     fn structural_progress(pass_generation: u64) -> ConformanceStructuralProgressSnapshot {
         ConformanceStructuralProgressSnapshot {
             current_epoch: pass_generation,
@@ -611,6 +814,8 @@ impl ClientBuilder {
             regenerated_queued_intents: HashMap::new(),
             audit_capture,
             convergence_checkpoints: HashMap::new(),
+            #[cfg(test)]
+            scripted_convergence_drain: None,
         }
     }
 }
@@ -1788,6 +1993,46 @@ impl HarnessClient {
             .collect()
     }
 
+    /// Drains the next one-shot convergence schedule batch.
+    fn drain_harness_convergence_groups(&mut self) -> Vec<GroupId> {
+        #[cfg(test)]
+        if let Some(script) = self.scripted_convergence_drain.as_mut() {
+            return script.drain_pending_groups();
+        }
+        self.engine_mut().drain_pending_convergence_groups()
+    }
+
+    /// Captures structural progress through the production or scripted seam.
+    fn harness_convergence_progress(
+        &self,
+        group_id: &GroupId,
+    ) -> Result<ConformanceStructuralProgressSnapshot, EngineError> {
+        #[cfg(test)]
+        if let Some(script) = self.scripted_convergence_drain.as_ref() {
+            assert_eq!(&script.group_id, group_id, "scripted convergence group");
+            return Ok(script.progress());
+        }
+        self.engine()
+            .conformance_structural_progress_snapshot(group_id)
+    }
+
+    /// Runs one convergence attempt through the production or scripted seam.
+    async fn harness_converge_and_drain(
+        &mut self,
+        group_id: &GroupId,
+        now_ms: u64,
+    ) -> Result<Vec<SendResult>, EngineError> {
+        #[cfg(test)]
+        if let Some(script) = self.scripted_convergence_drain.as_mut() {
+            assert_eq!(&script.group_id, group_id, "scripted convergence group");
+            return script.converge();
+        }
+        self.engine_mut()
+            .converge_and_drain_queued_outbound_intents(group_id, now_ms)
+            .await
+    }
+
+    /// Executes one deterministic convergence-work slice for this harness tick.
     async fn drive_due_convergence(
         &mut self,
         outcomes: &mut Vec<Result<IngestOutcome, EngineError>>,
@@ -1797,27 +2042,25 @@ impl HarnessClient {
             HashMap::<GroupId, (usize, ConformanceStructuralProgressSnapshot)>::new();
         let mut encountered_error = false;
         for _ in 0..HARNESS_CONVERGENCE_DRAIN_PASSES {
-            let groups = self.engine_mut().drain_pending_convergence_groups();
+            let mut groups = self.drain_harness_convergence_groups();
             if groups.is_empty() {
                 return;
             }
+            groups.sort_by(|left, right| left.as_slice().cmp(right.as_slice()));
             for group_id in groups {
-                let progress = match self
-                    .engine()
-                    .conformance_structural_progress_snapshot(&group_id)
-                {
-                    Ok(progress) => progress,
-                    Err(error) => {
-                        outcomes.push(Err(error));
-                        encountered_error = true;
-                        continue;
+                let initial_progress = if successful_attempts.contains_key(&group_id) {
+                    None
+                } else {
+                    match self.harness_convergence_progress(&group_id) {
+                        Ok(progress) => Some(progress),
+                        Err(error) => {
+                            outcomes.push(Err(error));
+                            encountered_error = true;
+                            continue;
+                        }
                     }
                 };
-                let results = match self
-                    .engine_mut()
-                    .converge_and_drain_queued_outbound_intents(&group_id, now_ms)
-                    .await
-                {
+                let results = match self.harness_converge_and_drain(&group_id, now_ms).await {
                     Ok(results) => results,
                     Err(e) => {
                         outcomes.push(Err(e));
@@ -1825,34 +2068,34 @@ impl HarnessClient {
                         continue;
                     }
                 };
-                let mut attempt_succeeded = true;
                 self.capture_engine_events();
                 for result in results {
                     if let Err(e) = self.publish_send_result(result).await {
                         outcomes.push(Err(e));
-                        attempt_succeeded = false;
                         encountered_error = true;
                     }
                 }
                 self.drain_auto_publish().await;
-                if attempt_succeeded {
-                    successful_attempts
-                        .entry(group_id)
-                        .and_modify(|(attempts, _)| {
-                            *attempts = attempts.saturating_add(1);
-                        })
-                        .or_insert((1, progress));
-                }
+                successful_attempts
+                    .entry(group_id)
+                    .and_modify(|(attempts, _)| {
+                        *attempts = attempts.saturating_add(1);
+                    })
+                    .or_insert_with(|| {
+                        (
+                            1,
+                            initial_progress.expect("untracked group captured initial progress"),
+                        )
+                    });
             }
         }
         if encountered_error {
             return;
         }
+        let mut successful_attempts = successful_attempts.into_iter().collect::<Vec<_>>();
+        successful_attempts.sort_by(|(left, _), (right, _)| left.as_slice().cmp(right.as_slice()));
         for (group_id, (attempts, initial)) in successful_attempts {
-            let mut current = match self
-                .engine()
-                .conformance_structural_progress_snapshot(&group_id)
-            {
+            let mut current = match self.harness_convergence_progress(&group_id) {
                 Ok(progress) => progress,
                 Err(error) => {
                     outcomes.push(Err(error));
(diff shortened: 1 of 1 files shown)


crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1710,21 +1821,58 @@ impl HarnessClient {
...
+                            *attempts = attempts.saturating_add(1);
+                        })
+                        .or_insert((1, progress));
+                }
+            }
+        }
+        if encountered_error {
+            return;
+        }
+        for (group_id, (attempts, initial)) in successful_attempts {
```
erskingardner, line 2175: This loop — the actual changed behavior — is untested. The new unit tests drive synthetic snapshots straight into `ensure_convergence_drain_progress`, so the call-site wiring has zero coverage: initial-snapshot capture timing (after the drain pops the schedule, before the first converge), per-group attempt accounting, the `scheduled_convergence_groups` normalization, and the continuation-vs-spin classification. #1651's acceptance criteria asked for regressions at the seven/eight-pass boundary and a 384-message retained-history catch-up through the harness; the validation note says the natural reproducer no longer reproduces on current master, but does the reduced competing-commit shape (eight rounds of same-epoch rival pairs) still exceed eight drain passes? If so, it would give a real end-to-end continuation regression (tick returns Ok with pending work, next tick settles). If nothing currently drives more than eight passes, consider a harness-level test that synthetically schedules enough rounds to cross the boundary, or at minimum a test that exercises the normalization at this call site — a future campaign hitting the continuation branch is currently trusting wiring that no test has ever run.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1696,12 +1793,26 @@ impl HarnessClient {
         outcomes: &mut Vec<Result<IngestOutcome, EngineError>>,
     ) {
         let now_ms = self.harness_convergence_now_ms();
+        let mut successful_attempts =
```
erskingardner, line 2131: Determinism nit: the crate rule is "Keep the simulator deterministic for a fixed seed and scenario spec," but `HashMap` iteration order at the guard loop below is randomized per process. With a single spinning group the pushed error is the fixed `Backend("...without structural progress")` string either way, but when multiple groups can fail (a per-group snapshot error vs a guard spin, or two different snapshot errors), which error lands in `outcomes` — and therefore the failure-capsule fingerprint — becomes run-dependent. `GroupId` has no `Ord` derive, so `BTreeMap` isn't drop-in; collecting into a `Vec` and sorting by `as_slice()` (or processing in drain order) would keep failure artifacts stable.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1710,21 +1821,58 @@ impl HarnessClient {
...
                 self.capture_engine_events();
                 for result in results {
                     if let Err(e) = self.publish_send_result(result).await {
                         outcomes.push(Err(e));
+                        attempt_succeeded = false;
+                        encountered_error = true;
                     }
                 }
                 self.drain_auto_publish().await;
+                if attempt_succeeded {
```
erskingardner: `attempt_succeeded` looks redundant: every path that sets it `false` also sets `encountered_error`, and `encountered_error` makes the function return before `successful_attempts` is ever read. Inserting unconditionally (and dropping the flag) is behavior-identical and removes a few lines.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -200,10 +201,27 @@ fn merge_histogram(target: &mut HistogramSnapshot, source: &HistogramSnapshot) {
     target.overflow_count = target.overflow_count.saturating_add(source.overflow_count);
 }
 
+fn ensure_convergence_drain_progress(
+    attempts: usize,
+    continuation_scheduled: bool,
+    initial: &ConformanceStructuralProgressSnapshot,
+    current: &ConformanceStructuralProgressSnapshot,
+) -> Result<(), EngineError> {
+    if attempts == HARNESS_CONVERGENCE_DRAIN_PASSES && continuation_scheduled && initial == current
```
erskingardner: Reachability question: the guard treats any snapshot difference — including a bare `pass_generation` bump — as progress (and `changed_convergence_drain_state_can_continue_after_eight_attempts` pins that). So the typed failure only catches spins that leave the durable pass generation unchanged (e.g. a `Collecting` pass waiting on a future cutoff that re-arms each pass without advancing); a spin that opens a new pass per drain call would evade it and fall back to the scenario-level `Eventually`/watchdog bound. Is that the intended spin model? If so, worth a line here or on the helper stating the limitation; if not, the generation field may need to be excluded from the progress comparison.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T17:34:08Z` |
| Commit reviewed | `abf5828e0a275fc20faea4ca6ac8425f56538451` |

The premise is sound and matches #1651's expected behavior: a bounded slice with valid scheduled continuation should not synthesize `EngineError::Backend`, underlying engine/publication errors are preserved, and the spin guard keeps a typed failure. Scope is surgical (one file, one function plus a pure helper). I verified the call-site wiring against the engine: `drain_pending_convergence_groups` removes the group before the initial snapshot, and `scheduled_convergence_groups` is per-group, so zeroing only the re-armed bit on `current` is the correct normalization.

The serious concern is validation: every new test passes synthetic snapshots straight into the pure helper, so the changed tick-level path — snapshot capture timing, per-group attempt accounting, the scheduled-bit normalization, continuation detection — has no coverage at all. #1651's acceptance criteria asked for regressions at the seven/eight-pass boundary and a 384-message retained-history catch-up; the validation note honestly says the natural reproducer no longer reproduces on current master, but that leaves the continuation branch trusted-but-untested for the next campaign that actually hits it. Details in the inline comments, along with a determinism nit, a dead flag, and a reachability question for the guard.

Note: the most relevant CI check for this crate ("Simulator and vectors") was still pending at review time; everything else was green.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -254,6 +272,85 @@ mod metric_tests {
...
+            HARNESS_CONVERGENCE_DRAIN_PASSES,
+            true,
+            &previous,
+            &current,
+        )
+        .expect("a later pass generation is structural progress");
+    }
+
+    #[test]
+    fn large_retained_history_can_continue_in_a_later_bounded_slice() {
```
erskingardner, line 505: This test (like the three siblings above it) exercises only the pure `ensure_convergence_drain_progress` predicate with fabricated snapshots. #1651's acceptance criteria ask for regressions "at seven/eight internal-pass boundaries and a 384-message retained-history catch-up" — i.e., through the harness itself. As shipped, none of the new wiring in `drive_due_convergence` is covered: the per-group attempt accounting, the pre-attempt snapshot capture, the `scheduled_convergence_groups` normalization, the error-preservation early return, and two-tick continuation via `HarnessClient::tick` could all silently break (or regress back to a synthesized `Backend` error) without any test failing.

The validation note explains the natural-history reproducer no longer enters this branch on current master, but the issue's second reproduction (eight rounds of same-epoch rival commits + reverse retained-history delivery) should still force more than eight drain passes through the real path — driving `tick` twice on that shape would pin the actual contract. At minimum, the PR test-plan checkboxes currently claim scenario-level coverage that the diff does not deliver; either add the harness-level regression or narrow the claim.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1696,12 +1793,26 @@ impl HarnessClient {
...
+        let mut successful_attempts =
+            HashMap::<GroupId, (usize, ConformanceStructuralProgressSnapshot)>::new();
+        let mut encountered_error = false;
         for _ in 0..HARNESS_CONVERGENCE_DRAIN_PASSES {
             let groups = self.engine_mut().drain_pending_convergence_groups();
             if groups.is_empty() {
                 return;
             }
             for group_id in groups {
+                let progress = match self
```
erskingardner: Non-blocking: this snapshot is captured on every pass for every group, but only the first is retained (`or_insert`); passes 2–8 compute and discard a full `capture_structural_progress_snapshot`, which lists every stored message row for the group and scans the event/auto-publish buffers. For the 384-message retained-history case that motivated this PR that is 8× the snapshot work per tick. Consider skipping the capture once the group is already tracked in `successful_attempts` (e.g., gate on `!successful_attempts.contains_key(&group_id)`). Fine to leave as-is at harness scale — just flagging the waste.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T17:49:57Z` |
| Commit reviewed | `abf5828e0a275fc20faea4ca6ac8425f56538451` |

## Summary

The core fix is correct, minimal, and well-scoped: the unconditional pass-limit `Backend` error is gone, a tick that exhausts its eight-pass slice with a valid re-armed continuation now returns successfully and preserves pending work for the next tick, and settling exactly on the eighth pass no longer false-fails (a genuine boundary improvement over the old behavior). Underlying engine/publication errors are still preserved verbatim, and the loop structure is otherwise untouched.

I verified the guard semantics against the engine implementation at this SHA:

- `drain_pending_convergence_groups` is a one-shot `HashSet::drain`, so the initial snapshot always observes `scheduled_convergence_groups == 0`; normalizing only `current` before comparison is sound, and reading `continuation_scheduled` before normalization is correct.
- `ConformanceStructuralProgressSnapshot` derives `PartialEq`, and every scenario-driven `HarnessClient` runs on the shared frozen `ManualConvergenceClock` (subject.rs:789/827), so `current_monotonic_ms` and the wake-deadline fields are stable across passes within a tick — `initial == current` is a meaningful no-progress signal, not flaky.
- A steady-state spin (re-arm every pass, nothing else changes) reaches `attempts == 8` with an unchanged snapshot and does produce the typed failure; a first tick that clears one-shot buffers counts as progress, so a real spin is caught one tick later at worst. Fail-open direction (never a false positive) is the right bias here.

## Main concern

The regression coverage is predicate-level only. The four new tests hand-construct snapshots and call `ensure_convergence_drain_progress` directly; nothing exercises `drive_due_convergence` end to end, so the attempt accounting, per-pass snapshot wiring, `scheduled_convergence_groups` normalization, error-preservation early return, and two-tick continuation via `HarnessClient::tick` could all silently break (or regress back to a synthesized `Backend` error) without any test failing. Issue #1651's acceptance criteria explicitly ask for regressions at the seven/eight-pass boundary and the retained-history catch-up driven through the harness, and the PR test-plan checkboxes currently claim that coverage. The validation note explains why the natural-history reproducer cannot be pinned right now, but the issue's competing-commit reproduction should still drive >8 passes through the real drain — a scenario-level test ticking twice on that shape would pin the contract this PR defines. Details inline.

## Minor notes

- The new tests live in `mod metric_tests`, a module named for histogram-merge metrics tests; a small rename or a sibling module would keep the organization honest.
- Per-pass snapshot capture discards 7 of 8 snapshots per group per tick (inline, non-blocking).

All CI checks pass at the reviewed SHA, including Simulator and vectors.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: Addressed in ce417beb. The initial structural snapshot is now captured only when the group is first inserted into successful_attempts; passes 2 through 8 reuse it instead of rescanning stored messages and buffers.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: Clarified in ce417beb. A durable pass_generation change is intentionally structural progress. The local guard rejects only an exact full-slice repeat; more complex state cycles remain bounded and diagnosed by the scenario fixed-point watchdog. The comment now states that division explicitly.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: Addressed in ce417beb with HarnessClient::tick-level coverage. The scripted 384-work-item/twelve-round schedule crosses the eight-pass boundary, leaves four rounds scheduled, and settles on the second tick; companion tests cover seven/exactly-eight passes, normalized unchanged-state spin detection, and underlying error preservation. The updated PR note explains why the current real retained-history arms no longer isolate this branch.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: Addressed in ce417beb. HarnessClient::tick now has direct boundary coverage for schedule draining, per-group attempt accounting, scheduled_convergence_groups normalization, and two-tick continuation. A stalled scripted schedule proves the normalized re-arm edge does not mask an unchanged-state spin, and a failing schedule proves the original convergence error is retained. The current real family arms no longer provide a fast isolated reproduction, as documented in the updated PR validation note.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: Addressed in ce417beb. Each drained group batch is sorted by GroupId bytes before processing, and the final successful-attempt entries are also sorted before snapshot/error evaluation. That keeps both operation errors and guard artifacts stable for a fixed scenario.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -254,6 +272,85 @@ mod metric_tests {
...
+    #[test]
+    fn eight_repeated_convergence_drain_states_are_a_typed_failure() {
+        let progress = structural_progress(7);
+
+        let error = ensure_convergence_drain_progress(
+            HARNESS_CONVERGENCE_DRAIN_PASSES,
+            true,
+            &progress,
+            &progress,
+        )
```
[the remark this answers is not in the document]
erskingardner: Addressed in ce417beb. The regression now drives the real HarnessClient::tick surface with scripted scheduler work: seven and exactly eight passes, plus 384 work items over twelve rounds split 8 + 4 across two ticks. The same path also covers scheduled-edge normalization, a true unchanged-state spin, and preservation of an underlying convergence error. I attempted the current real competing family arm first, but it now enters the long 128-turn fallback and no longer isolates this boundary on current master.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: Addressed in ce417beb. Removed attempt_succeeded and now records the attempt unconditionally; encountered_error still returns before the progress guard, preserving the original error behavior.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -200,18 +302,43 @@ fn merge_histogram(target: &mut HistogramSnapshot, source: &HistogramSnapshot) {
...
+/// remain the scenario fixed-point driver's responsibility; this local guard
+/// exists to catch a scheduler that repeatedly re-arms the same state without
+/// turning the per-tick work bound into a convergence deadline.
+fn ensure_convergence_drain_progress(
+    attempts: usize,
+    continuation_scheduled: bool,
+    initial: &ConformanceStructuralProgressSnapshot,
+    current: &ConformanceStructuralProgressSnapshot,
+) -> Result<(), EngineError> {
+    if attempts == HARNESS_CONVERGENCE_DRAIN_PASSES && continuation_scheduled && initial == current
```
erskingardner: Blocking: this equality includes `current_monotonic_ms`, which `capture_structural_progress_snapshot` fills from `engine.convergence_now()`. Clients built without an injected clock run `SystemConvergenceClock` (real `Instant` elapsed ms), so the initial and guard snapshots are taken at different wall times and effectively never compare equal across eight SQLite/OpenMLS passes — the guard never fires there, and in a rare sub-millisecond slice it can fire, making the outcome timing-dependent. A true spin on a legacy-clock client now continues silently instead of producing the old hard error; only frozen-manual-clock (virtual-time) clients get working spin detection. Please normalize the clock-derived fields out of the comparison (at minimum `current_monotonic_ms`), the same way the scheduled bit is normalized below — or compare a reduced durable-state tuple (epoch, lifecycle, pass generation/phase, normalized pending work, terminal flag).

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -79,8 +80,108 @@ pub struct HarnessClient {
...
+
+    /// Captures the same privacy-safe fields used by the real engine seam.
+    fn progress(&self) -> ConformanceStructuralProgressSnapshot {
+        let pending_work = cgka_engine::conformance_snapshot::ConformancePendingWorkSnapshot {
+            scheduled_convergence_groups: usize::from(self.scheduled),
+            ..Default::default()
+        };
+        ConformanceStructuralProgressSnapshot {
+            current_epoch: self.generation,
+            current_monotonic_ms: HARNESS_CONVERGENCE_SETTLED_AT_MS,
```
erskingardner, line 164: Non-blocking: the scripted seam pins `current_monotonic_ms` to a constant — exactly the frozen-clock regime where the guard works — so the tests cannot see the default-clock gap noted on the comparison. After normalizing the comparison, a cheap assertion (or a direct harness test on a default-clock client) that a `current_monotonic_ms` delta alone does not count as structural progress would pin the semantics.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1691,40 +1993,129 @@ impl HarnessClient {
...
+                        )
+                    });
+            }
+        }
+        if encountered_error {
+            return;
+        }
+        let mut successful_attempts = successful_attempts.into_iter().collect::<Vec<_>>();
+        successful_attempts.sort_by(|(left, _), (right, _)| left.as_slice().cmp(right.as_slice()));
+        for (group_id, (attempts, initial)) in successful_attempts {
```
erskingardner, line 2175: Non-blocking: `ensure_convergence_drain_progress` returns `Ok` unless `attempts == HARNESS_CONVERGENCE_DRAIN_PASSES`, so for any group that did not consume the full slice this snapshot is a full per-group scan (`list_messages` plus buffer walks) whose only possible effect is a spurious error. Gating the snapshot+guard on that equality removes both the work and the error surface.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1691,40 +1993,129 @@ impl HarnessClient {
...
+            let mut current = match self.harness_convergence_progress(&group_id) {
+                Ok(progress) => progress,
+                Err(error) => {
+                    outcomes.push(Err(error));
```
erskingardner: Non-blocking: a snapshot error here pushes a new `Err` and returns, converting a fully successful bounded continuation into a failed tick — the same synthesized-error shape this PR removes. If the assessment query fails, the harness cannot judge progress; skipping the guard for that group (or relying on the `attempts` gate above, which makes this unreachable in the common case) keeps the tick's success honest.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T20:20:34Z` |
| Commit reviewed | `ce417beb9f2c8ce473dfd110da72f19644382e6c` |

The premise is sound and matches #1651: exhausting a bounded slice with valid scheduled continuation must not synthesize `EngineError::Backend`, while a true no-progress spin stays a typed failure. The core rewrite of `drive_due_convergence` does that, the group sorting fixes real `HashSet` iteration nondeterminism, and underlying engine/publication errors are preserved. The validation note is honest about why the scripted seam replaced the natural reproducer.

One blocking issue: the spin guard's snapshot equality includes clock-derived fields, so its behavior depends on which convergence clock the engine is running.

- **Blocking — guard is inert (and nondeterministic) on the default engine clock.** `initial == current` compares `ConformanceStructuralProgressSnapshot::current_monotonic_ms`, which `capture_structural_progress_snapshot` fills from `engine.convergence_now()`. Clients built without an injected clock (the legacy `ClientBuilder::new(...).attach(&bus)` path) run `SystemConvergenceClock`, i.e. real elapsed `Instant` milliseconds, so the initial and guard snapshots are taken at different wall times and essentially never compare equal across eight SQLite/OpenMLS passes. There the guard never fires — a true spin now continues silently instead of producing the old hard error — and in the rare sub-millisecond slice it can fire, making the harness outcome timing-dependent. Only frozen-manual-clock (virtual-time) clients get working spin detection, and the scripted seam hard-codes exactly that frozen clock, so the gap is invisible to the new tests. Please normalize the clock-derived fields out of the comparison (at minimum `current_monotonic_ms`; consider `earliest_next_wake_monotonic_ms`/`runnable_work`) the same way the scheduled bit is normalized, or compare a reduced durable-state tuple (epoch, lifecycle, pass generation/phase, normalized pending work, terminal flag).
- **Non-blocking — guard assessment can fail an otherwise successful slice.** A snapshot error in the post-slice loop pushes a new `Err` and returns, converting a fully successful bounded continuation into a failed tick — the same synthesis this PR removes. Since `ensure_convergence_drain_progress` returns `Ok` immediately unless `attempts == HARNESS_CONVERGENCE_DRAIN_PASSES`, gating the snapshot+guard on that equality both removes the new error surface for the common case and skips a full per-group snapshot (`list_messages` + buffer scans) per tick.
- **Non-blocking — test fidelity.** The scripted drain bypasses the real engine, so the real-engine assumptions the guard leans on (initial snapshot sees `scheduled_convergence_groups == 0` because the drain is one-shot; re-arm restores it) are verified only by reading engine code, not by a test. After the normalization fix, a cheap direct harness test that ticks a real client on the default clock through a spin (or at least asserts the guard's comparison ignores `current_monotonic_ms`) would pin the semantics.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -1691,40 +1993,129 @@ impl HarnessClient {
...
                         outcomes.push(Err(e));
+                        encountered_error = true;
                         continue;
                     }
                 };
                 self.capture_engine_events();
                 for result in results {
                     if let Err(e) = self.publish_send_result(result).await {
                         outcomes.push(Err(e));
+                        encountered_error = true;
```
erskingardner, line 2158: Non-blocking coverage note: the scripted seam always returns `Ok(Vec::new())`, so this publish-failure branch (and its guard-skip via `encountered_error`) has no tick-level coverage, unlike the converge-error path which is pinned verbatim. If you want parity, `ScriptedConvergenceDrain` could emit one result whose `publish_send_result` fails. Fine to leave as-is — the branch is simple and mirrors the converge-error handling.

crates/cgka-conformance-simulator/src/client.rs
```diff
@@ -79,8 +80,108 @@ pub struct HarnessClient {
...
+            self.generation = self.generation.saturating_add(1);
+            self.scheduled = self.remaining_passes > 0;
+        } else {
+            self.scheduled = true;
+        }
+        Ok(Vec::new())
+    }
+}
+
+/// Saturating-merge one engine metrics snapshot into the harness aggregate.
```
erskingardner, line 200: Nit: the doc comments added to the pre-existing `merge_engine_metrics`/`merge_histogram` helpers are slight drive-by chaff relative to the drain fix (presumably for the docstring-coverage gate). Harmless, but worth keeping in mind for future diffs.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T20:30:20Z` |
| Commit reviewed | `ce417beb9f2c8ce473dfd110da72f19644382e6c` |

Note: posted as a comment because GitHub does not allow approving one's own PR; the recommendation above stands.

## Summary

This head resolves the main concern from the earlier reviews: the changed drain loop is now exercised end to end. The scripted seam drives the real `HarnessClient::tick` through the seven/eight-pass boundary, a scripted 384-message/12-round continuation across two ticks (8 + 4), the true-spin guard, and verbatim underlying-error preservation — matching the test plan checkboxes and #1651's acceptance criteria.

I re-verified the guard semantics against the engine at this SHA:

- `drain_pending_convergence_groups` is a one-shot `HashSet::drain`, and `scheduled_convergence_groups` in the snapshot is per-group (`usize::from(pending_convergence_groups.contains(group_id))`), so capturing the initial snapshot after the drain and normalizing only the re-armed bit out of `current` is the correct comparison.
- Per-group `attempts` can reach at most one per pass, so `attempts == HARNESS_CONVERGENCE_DRAIN_PASSES` exactly identifies a full slice; `initial_progress.expect(...)` is unreachable-safe because `None` only occurs when the entry already exists.
- The guard is fail-open in the right direction: only an exact full-slice repeat with unchanged structural state fails, and any underlying engine/publication error skips the guard entirely via `encountered_error`.
- Sorting each pass's drained groups and the guard iteration makes multi-group processing and failure reporting deterministic, as the PR body claims. The scripted client has no `default_group`, so the tick tests genuinely flow through `drive_due_convergence`.

Settling exactly on the eighth pass no longer false-fails (continuation_scheduled = false), which is a real boundary improvement over the old unconditional error. The validation note's honesty about the natural-history reproducer is appreciated; the scripted shapes pin the contract through the real tick path, which is what matters for regression value.

Two non-blocking notes inline. All CI green at the reviewed SHA, including Simulator and vectors.

Datawav (commented):
## Automated review

Medium — No-progress detection is ineffective with the default system clock

crates/cgka-conformance-simulator/src/client.rs:317

Evidence: The new guard compares complete ConformanceStructuralProgressSnapshot values after normalizing only scheduled convergence state. That snapshot derives PartialEq/Eq and includes current_monotonic_ms, populated from engine.convergence_now(). The scripted tests hold this field constant, so they do not exercise the default SystemConvergenceClock behavior.

Trigger: A default-clock HarnessClient executes all eight convergence-drain attempts while repeatedly re-arming the group but making no durable structural progress. Time advances between the initial and final snapshots.

Changed behavior: The snapshots differ solely because current_monotonic_ms advanced, so the new equality-based no-progress guard classifies the slice as progress and returns success.

Consequence: A genuine convergence spin is silently continued across ticks rather than producing the intended typed “without structural progress” error. This can mask simulator liveness failures until a broader scenario watchdog or eventuality bound expires. It also makes detection timing-dependent in the unlikely case that the whole slice completes within one clock millisecond.

Smallest sound correction: Exclude observation time from structural-progress equality, for example by setting current.current_monotonic_ms to initial.current_monotonic_ms before comparing, or by comparing an explicit durable-state projection. Add a regression where attempts equal HARNESS_CONVERGENCE_DRAIN_PASSES, every structural field is unchanged, and only current_monotonic_ms differs; it should have failed before the correction by returning Ok instead of the typed no-progress error.

Confidence: High.

No physical-device evidence exists for this head because automated_device_gate was not supplied.

Audit evidence:
  Head SHA: ce417beb9f2c8ce473dfd110da72f19644382e6c
  Changed file inspected: crates/cgka-conformance-simulator/src/client.rs
  Diff coverage: complete; no missing patch files.
  Review mode: initial.
  Design fit: The bounded continuation and deterministic group processing fit the simulator’s role, but clock time is observational rather than structural and should not govern spin detection.

Risk lenses checked:
  general — boundary and error paths; backward compatibility; behavioral regression coverage. The eight-attempt boundary and underlying-error paths were inspected. The finding specifically affects compatibility with default-clock clients and exposes a missing behavioral regression.

Check evidence:
  Supplied check evidence is available and complete, with all listed checks completed successfully. Relevant successful checks include Simulator and vectors, Convergence policy pin (release assertions), the Rust test shards, Rust check, Rust clippy, Rust format, and Rust audit. These CI results do not cover the default-clock snapshot-delta failure described above.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: I rechecked these comments and am keeping them. They describe the privacy-safe metric helpers at their definitions, and the documentation gate now passes with them present. Removing them would reduce local API clarity without changing the bounded-drain fix, so I do not think a code change is warranted here.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: Fixed in 0ef87596. The post-slice loop now checks `attempts == HARNESS_CONVERGENCE_DRAIN_PASSES` before requesting a final snapshot, so groups that did not exhaust the slice incur no extra assessment.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: Fixed in 0ef87596. Progress snapshots are diagnostic for the local guard: if either the initial or final assessment is unavailable, the guard is skipped and otherwise successful convergence results remain successful. A scripted tick test covers both assessment positions; actual convergence and publication errors are still propagated.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: Fixed in 0ef87596. The guard now compares a durable projection rather than the raw snapshot: it excludes `current_monotonic_ms`, derived wake/runnable fields, and normalizes the one-shot scheduled edge, while retaining epoch, lifecycle, durable pending work, pass generation/phase, and terminal state. This keeps the guard effective with the default system clock.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: Added in 0ef87596. `clock_deltas_do_not_count_as_convergence_drain_progress` changes the monotonic time plus derived wake/runnable fields and verifies that an otherwise unchanged full re-armed slice still returns the typed no-progress error.

crates/cgka-conformance-simulator/src/client.rs
erskingardner: I checked this suggestion and am leaving the scripted seam focused on convergence scheduling. The branch only mirrors publication failures already returned by the production path, while constructing a synthetic `SendResult` would widen the test seam substantially and couple this issue's regression fixture to publication internals. Existing publication lifecycle tests cover that behavior; this PR keeps the new script specific to the bounded-drain bug.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T06:37:24Z` |
| Commit reviewed | `0ef875966538fbd76b35f29093b427870164d4ea` |

0ef87596 resolves all three findings from the previous review at ce417beb, and the full CI matrix (including Simulator and vectors) is green.

- **Clock-dependent guard (was blocking): fixed.** `convergence_drain_progress_key` now projects both snapshots onto durable fields only — `current_epoch`, `lifecycle`, `pending_work` (scheduled bit normalized on both sides), `pass_generation`, `pass_phase`, `terminal_unrecoverable` — excluding `current_monotonic_ms`, `earliest_next_wake_monotonic_ms`, and `runnable_work`. The guard no longer depends on which convergence clock the engine runs, and the new `clock_deltas_do_not_count_as_convergence_drain_progress` test pins exactly the failure mode I flagged (a default-clock spin masked by wall-time deltas). Projecting both sides is cleaner than the previous one-sided mutation.
- **Guard assessment error surface: fixed.** The post-slice loop now gates on `attempts == HARNESS_CONVERGENCE_DRAIN_PASSES`, skips groups whose initial assessment was unavailable, and treats a failed assessment snapshot as "cannot judge" (skip the guard) rather than synthesizing a tick error. The mid-loop initial capture also no longer blocks the group's convergence on a diagnostic failure — work proceeds and the guard is simply skipped. `harness_tick_skips_an_unavailable_progress_assessment` covers both the initial-capture and guard-capture failure points through the real `tick` path, including continuation preservation.
- **Test fidelity: addressed.** The clock-delta test plus the durable-key projection pin the guard semantics I was concerned about. A real-engine default-clock spin regression test is still absent, but the guard is now structurally clock-independent, so the residual risk is small.

One observation, non-blocking: with the initial capture now using `.ok()` and the group converging regardless, a group whose snapshots always fail is silently exempt from spin detection. That is the right trade-off for a diagnostic surface (a real failure still surfaces through the converge call itself), just worth keeping in mind if `conformance_structural_progress_snapshot` ever starts failing for non-diagnostic reasons.

Merge when ready.

PR #1663, opened 2026-09-03
Bound generated-case minimization and preserve failures

### LINKED ISSUE #1652: Generated-case minimization consumes the process deadline and hides the original failure
## Summary

Automatic semantic minimization runs synchronously before the generated scenario report is written and has no time or trial budget. For expensive offline catch-up failures, the original scenario finishes well inside the isolated case deadline, but minimization consumes the remainder. The process is killed with no report, fixture candidate, portable failure capsule, or original failure classification.

This is distinct from #1639: that issue loses artifacts because a subject operation panics. Here the raw scenario returns a structured report successfully; post-run minimization prevents it from being persisted.

Observed on current master `127db19cabdbc9d61692c8f9027f2ddf7c91ed66`.

## Campaign evidence

A process-isolated `offline-catchup-pressure/v1` campaign ran seeds 9101 through 9104 with 18 cases per seed and a 300-second per-case deadline.

For every seed:

- cases 0 through 8 completed in about 2.6 to 15.5 seconds;
- case 9 timed out at about 300 seconds;
- case 10 completed in about 15.8 seconds;
- cases 11 through 17 timed out at about 300 seconds.

Across the four breadth runs, 32 of 72 cases timed out and none produced a structured non-timeout failure.

Direct raw execution without generated metadata/minimization shows:

- case 9 returns its `pending_work_remaining` failure in about 15 seconds;
- case 11 returns its failed catch-up assertion/harness continuation in about 31 to 33 seconds;
- case 12 returns its large natural catch-up/harness continuation in about 52 seconds.

Sampling the timed-out case-9 process showed it repeatedly executing:

```text
add_generated_metadata
  -> minimize_failing_case
  -> minimize_scenario_with
  -> reproduces_failure
  -> run_generated_scenario
  -> RetainedRelaySubject::tick
  -> HarnessClient::tick
  -> advance_convergence_inputs_until_settled
```

The saved generated input exists because the process orchestrator writes it before launch, but the report and failure capsule do not.

## Root cause

`run_generated_case_report_with_capture_on_subject` calls `add_generated_metadata` before returning the report. `add_generated_metadata` awaits `minimize_failing_case`, and `minimize_scenario_with` repeatedly reruns dependency-aware candidates until no reduction succeeds.

There is no:

- wall-clock budget;
- trial-count budget;
- per-trial timeout;
- cancellation tied to the process deadline;
- checkpoint of the original report before minimization.

For a scenario with hundreds of steps and a 15 to 50 second reproduction cost, even a few trials exceed the parent process deadline.

## Expected behavior

The original structured failure and capture artifacts must be durable before optional reduction can consume the case budget. Minimization should be explicitly bounded and should report whether it completed, exhausted a budget, or was skipped.

## Acceptance criteria

- Persist or return the original report and failure capture before minimization can exceed the process deadline.
- Give semantic minimization a configurable wall-clock and/or trial budget with a conservative default.
- Bound each reproduction trial.
- Record minimization status such as `complete`, `budget_exhausted`, `not_reproducible`, or `skipped`.
- A minimization timeout preserves the original failure kind, report, fixture candidate, and portable failure capsule.
- The process campaign summary distinguishes scenario timeout from post-processing/minimization budget exhaustion.
- Add an expensive synthetic reproducer that completes once but cannot finish reduction inside the test budget.
- Preserve deterministic reduced output when minimization completes within budget.

erskingardner:
## Summary

- persist the original generated-case report, fixture candidate, and failure capsules before semantic minimization starts
- bound minimization by configurable wall-clock, trial-count, and per-trial limits, with explicit completion statuses
- distinguish scenario execution, minimization, and post-processing timeouts in process campaign summaries

## Test plan

- [x] cargo test -p cgka-conformance-simulator --locked
- [x] cargo test -p cgka-conformance-simulator --test report_runner --locked
- [x] cargo test -p cgka-conformance-simulator --bin cgka-conformance-campaign --locked
- [x] just fast-ci

Fixes #1652

## Summary by CodeRabbit

* **New Features**
  * Added bounded minimization for generated scenarios with configurable wall-clock, trial-count, and per-trial limits.
  * Reports now include minimization status, trial counts, budgets, and timeout phases.
  * Added command-line options for configuring and validating minimization budgets.

* **Reliability**
  * Original failure artifacts and classifications are preserved when minimization is interrupted, exhausted, or unsuccessful.
  * Improved safe artifact replacement and cleanup of temporary campaign files.

* **Documentation**
  * Updated guidance for minimization outcomes, timeouts, pending records, and campaign result aggregation.

### DIFF
--- a/crates/cgka-conformance-simulator/src/family.rs
+++ b/crates/cgka-conformance-simulator/src/family.rs
@@ -19,8 +19,52 @@ use rand::{Rng, SeedableRng};
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeSet;
 use std::fmt;
+use std::time::{Duration, Instant};
 
 pub const GENERATED_SCENARIO_INPUT_SCHEMA_VERSION: &str = "1";
+
+/// Bounds optional semantic reduction after the original generated report has
+/// been produced. The defaults keep expensive failures from consuming the
+/// enclosing process deadline while still allowing small deterministic cases
+/// to reduce completely.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct GeneratedScenarioMinimizationBudget {
+    pub wall_clock: Duration,
+    pub max_trials: usize,
+    pub per_trial_timeout: Duration,
+}
+
+impl Default for GeneratedScenarioMinimizationBudget {
+    fn default() -> Self {
+        Self {
+            wall_clock: Duration::from_secs(30),
+            max_trials: 256,
+            per_trial_timeout: Duration::from_secs(5),
+        }
+    }
+}
+
+impl GeneratedScenarioMinimizationBudget {
+    pub fn validate(self) -> Result<Self, &'static str> {
+        if self.wall_clock.is_zero() {
+            return Err("minimization wall-clock budget must be greater than zero");
+        }
+        if self.max_trials == 0 {
+            return Err("minimization trial budget must be greater than zero");
+        }
+        if self.per_trial_timeout.is_zero() {
+            return Err("minimization per-trial timeout must be greater than zero");
+        }
+        Ok(self)
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct GeneratedScenarioMinimizationOutcome {
+    status: crate::GeneratedScenarioMinimizationStatus,
+    trials: usize,
+    minimized_case: Option<ScenarioSpec>,
+}
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct GeneratedScenarioCase {
@@ -635,12 +679,15 @@ pub async fn run_generated_case_report_with_storage_mode(
         storage_mode,
     )
     .await?;
-    add_generated_metadata(
+    let budget = GeneratedScenarioMinimizationBudget::default();
+    prepare_generated_metadata(case, &mut report, budget);
+    complete_generated_minimization(
         case,
         case.subject,
         expected_trace.as_ref(),
         &mut report,
         storage_mode,
+        budget,
     )
     .await;
     Ok(report)
@@ -670,6 +717,37 @@ pub async fn run_generated_case_report_with_capture_on_subject(
     expected_trace: Option<ScenarioTrace>,
     storage_mode: HarnessStorageMode,
     capture_sensitive_replay: bool,
+) -> Result<(ScenarioReport, crate::ScenarioFailureCaptureV1), ScenarioRunError> {
+    let budget = GeneratedScenarioMinimizationBudget::default();
+    let (mut report, failure_capture) =
+        run_generated_case_report_with_capture_on_subject_before_minimization(
+            case,
+            execution_subject,
+            expected_trace.clone(),
+            storage_mode,
+            capture_sensitive_replay,
+            budget,
+        )
+        .await?;
+    complete_generated_minimization(
+        case,
+        execution_subject,
+        expected_trace.as_ref(),
+        &mut report,
+        storage_mode,
+        budget,
+    )
+    .await;
+    Ok((report, failure_capture))
+}
+
+pub(crate) async fn run_generated_case_report_with_capture_on_subject_before_minimization(
+    case: &GeneratedScenarioCase,
+    execution_subject: GeneratedSubjectKind,
+    expected_trace: Option<ScenarioTrace>,
+    storage_mode: HarnessStorageMode,
+    capture_sensitive_replay: bool,
+    budget: GeneratedScenarioMinimizationBudget,
 ) -> Result<(ScenarioReport, crate::ScenarioFailureCaptureV1), ScenarioRunError> {
     let (mut report, failure_capture) = match execution_subject {
         GeneratedSubjectKind::Engine => {
@@ -744,35 +822,19 @@ pub async fn run_generated_case_report_with_capture_on_subject(
             )
         }
     };
-    add_generated_metadata(
-        case,
-        execution_subject,
-        expected_trace.as_ref(),
-        &mut report,
-        storage_mode,
-    )
-    .await;
+    prepare_generated_metadata(case, &mut report, budget);
     Ok((report, failure_capture))
 }
 
-async fn add_generated_metadata(
+fn prepare_generated_metadata(
     case: &GeneratedScenarioCase,
-    execution_subject: GeneratedSubjectKind,
-    expected_trace: Option<&ScenarioTrace>,
     report: &mut ScenarioReport,
-    storage_mode: HarnessStorageMode,
+    budget: GeneratedScenarioMinimizationBudget,
 ) {
-    let minimized_case = if fingerprint_report_failure(report).is_ok() {
-        minimize_failing_case(
-            case,
-            execution_subject,
-            expected_trace,
-            report,
-            storage_mode,
-        )
-        .await
+    let status = if fingerprint_report_failure(report).is_ok() {
+        crate::GeneratedScenarioMinimizationStatus::Pending
     } else {
-        None
+        crate::GeneratedScenarioMinimizationStatus::Skipped
     };
     report.metadata.generated = Some(GeneratedScenarioMetadata {
         family_name: case.family_name.clone(),
@@ -780,8 +842,45 @@ async fn add_generated_metadata(
         seed: case.seed,
         case_index: case.case_index,
         workload_profile: case.workload_profile.clone(),
-        minimized_case,
+        minimization: crate::GeneratedScenarioMinimizationV1 {
+            status,
+            trials: 0,
+            wall_clock_budget_ms: duration_ms(budget.wall_clock),
+            max_trials: budget.max_trials,
+            per_trial_timeout_ms: duration_ms(budget.per_trial_timeout),
+        },
+        minimized_case: None,
     });
+}
+
+pub(crate) async fn complete_generated_minimization(
+    case: &GeneratedScenarioCase,
+    execution_subject: GeneratedSubjectKind,
+    expected_trace: Option<&ScenarioTrace>,
+    report: &mut ScenarioReport,
+    storage_mode: HarnessStorageMode,
+    budget: GeneratedScenarioMinimizationBudget,
+) {
+    let is_pending = report.metadata.generated.as_ref().is_some_and(|metadata| {
+        metadata.minimization.status == crate::GeneratedScenarioMinimizationStatus::Pending
+    });
+    if !is_pending {
+        return;
+    }
+    let outcome = minimize_failing_case(
+        case,
+        execution_subject,
+        expected_trace,
+        report,
+        storage_mode,
+        budget,
+    )
+    .await;
+    if let Some(metadata) = report.metadata.generated.as_mut() {
+        metadata.minimization.status = outcome.status;
+        metadata.minimization.trials = outcome.trials;
+        metadata.minimized_case = outcome.minimized_case;
+    }
 }
 
 async fn minimize_failing_case(
@@ -790,11 +889,17 @@ async fn minimize_failing_case(
     expected_trace: Option<&ScenarioTrace>,
     failing_report: &ScenarioReport,
     storage_mode: HarnessStorageMode,
-) -> Option<ScenarioSpec> {
-    let target_identity = fingerprint_report_failure(failing_report)
-        .ok()?
-        .semantic_identity();
-    minimize_scenario_with(case.scenario.clone(), |trial| {
+    budget: GeneratedScenarioMinimizationBudget,
+) -> GeneratedScenarioMinimizationOutcome {
+    let Ok(fingerprint) = fingerprint_report_failure(failing_report) else {
+        return GeneratedScenarioMinimizationOutcome {
+            status: crate::GeneratedScenarioMinimizationStatus::Skipped,
+            trials: 0,
+            minimized_case: None,
+        };
+    };
+    let target_identity = fingerprint.semantic_identity();
+    minimize_scenario_with(case.scenario.clone(), budget, |trial| {
         let expected_trace = expected_trace.cloned();
         let expected_outcomes = case.expected_outcomes.clone();
         let target_identity = target_identity.clone();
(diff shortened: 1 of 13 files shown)


erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T19:41:00Z` |
| Commit reviewed | `9ef97ae4fb0c5cd5b767eb17b19183fe9dbbe6a0` |

The overall design is right and satisfies every acceptance criterion in #1652: the original report, fixture candidate, and capsules are durably written before bounded reduction starts, completed minimization atomically replaces them, an interrupted worker leaves a complete original artifact with a durable `pending` marker, and the campaign summary classifies timeout phases only from durable evidence. The report schema stays backward-compatible via `#[serde(default)]`, and budget plumbing is validated end to end. All CI checks pass at the reviewed head.

One blocking issue (inline): the new `expensive_reproducer_exhausts_the_minimization_budget_after_the_original_run` test is scheduler-timing-dependent and will flake intermittently on loaded machines.

Non-blocking notes:

- `failing_generated_case_records_a_minimized_reproducer` now also accepts `budget_exhausted` and only asserts the app-message-storm removal under `complete`. Acceptable tradeoff for a small case against the default 30s budget, but the deterministic-reduction property is now only asserted on the `complete` path.
- On the expensive paths that motivated #1652 (15-50s offline-catchup failures), the default 5s per-trial timeout means the initial reproduction trial will usually time out, so those cases will typically record `budget_exhausted` with `trials: 1` and no minimized case. That is the intended conservative outcome (original evidence preserved), just worth keeping in mind when reading campaign summaries.
- `classify_timeout_phase` keys off `timed_out` only, so a signal-killed worker during minimization gets no phase - but its durable `pending` minimization status is still recorded independently, so operators can distinguish it. Fine as-is.

crates/cgka-conformance-simulator/src/family.rs
```diff
@@ -4253,6 +4457,86 @@ mod tests {
...
+            }],
+        };
+        let attempts = Arc::new(AtomicUsize::new(0));
+        let attempts_for_reproducer = Arc::clone(&attempts);
+        let started = Instant::now();
+
+        let outcome = minimize_scenario_with(
+            scenario,
+            GeneratedScenarioMinimizationBudget {
+                wall_clock: Duration::from_millis(50),
```
erskingardner: **Blocking: this test is scheduler-timing-dependent and will flake on loaded machines.**

`assert_eq!(outcome.trials, 2)` (line 4504) requires the second trial to start, but `run_minimization_trial` checks `budget.wall_clock.checked_sub(started.elapsed())` before arming each trial. With a 50ms wall clock, any preemption longer than 50ms between the first trial completing and the second budget check (this is a single-threaded `#[tokio::test]` runtime, and CI runners are shared) means the second trial never starts, `trials == 1`, and the assertion fails. `assert_eq!(attempts.load(...), 2)` has the same dependency.

Raise the wall clock and let `per_trial_timeout = 10ms` remain the deciding bound - that is the behavior under test:

```diff
 GeneratedScenarioMinimizationBudget {
-    wall_clock: Duration::from_millis(50),
+    wall_clock: Duration::from_secs(5),
     max_trials: 8,
     per_trial_timeout: Duration::from_millis(10),
 },
```

The `started.elapsed() < Duration::from_millis(500)` assertion at line 4508 still proves the per-trial timeout bounds the expensive reproduction. If you want that assertion fully stall-proof too, consider bumping its bound (a stall during trial 2's 10ms sleep delays the tokio timer poll, so elapsed can exceed 500ms even though the timeout did its job), or drop it in favor of asserting only on status and trial count.

crates/cgka-conformance-simulator/src/report.rs
```diff
@@ -529,6 +550,66 @@ fn write_report_failure_capsules(
...
+}
+
+fn replace_private_json<T: serde::Serialize>(path: &Path, value: &T) -> Result<(), Box<dyn Error>> {
+    write_private_json(path, value, ArtifactWriteMode::Replace)
+}
+
+/// Durably replace a private artifact without exposing a truncated live path.
+/// A process killed during the write leaves either the original file or the
+/// complete replacement.
+fn replace_private(path: &Path, bytes: &[u8]) -> std::io::Result<()> {
```
erskingardner, line 573: Non-blocking: this satisfies the repo's local-artifact-safety posture (built on `fs_private::create_new_private`, atomic rename, parent-dir fsync, and covered by an on-disk mode test). Since durable replace is a general primitive that mirrors `write_private`/`create_new_private`, if a second caller ever needs it (other campaign artifacts, CLI outputs), consider moving it into `crates/fs-private` so the posture and its mode test live in one place. Fine to keep here while it has a single caller.

crates/cgka-conformance-simulator/src/family.rs
```diff
@@ -4253,6 +4457,86 @@ mod tests {
...
+                max_trials: 8,
+                per_trial_timeout: Duration::from_millis(10),
+            },
```
erskingardner, line 4536: **Non-blocking: this test can flake on a loaded machine.**

The 50 ms wall-clock budget races the scheduler: `run_minimization_trial` computes the remaining budget with `budget.wall_clock.checked_sub(started.elapsed())`. If the test thread is stalled for more than 50 ms between the first (instant) trial and the second call, the second trial never starts, `outcome.trials` is 1, and `assert_eq!(outcome.trials, 2)` below fails — even though per-trial bounding is what the test actually wants to prove.

Raising the wall clock (e.g. `Duration::from_secs(5)`) keeps the per-trial 10 ms timeout as the deciding bound, preserves the `started.elapsed() < 500 ms` assertion as the real proof, and makes the test independent of scheduler timing.

crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
```diff
@@ -2406,23 +2406,36 @@ async fn failing_generated_case_records_a_minimized_reproducer() {
...
+        generated.minimization.status,
+        cgka_conformance_simulator::GeneratedScenarioMinimizationStatus::Complete
+            | cgka_conformance_simulator::GeneratedScenarioMinimizationStatus::BudgetExhausted
+    ));
+    assert!(generated.minimization.trials > 1);
+    let minimized = generated
+        .minimized_case
+        .as_ref()
         .expect("failing generated case should record a minimized case");
     assert!(
```
erskingardner, line 2433: **Non-blocking: these assertions are inconsistent with the `BudgetExhausted` status they accept.**

The test accepts `Complete | BudgetExhausted`, but then unconditionally requires `trials > 1` and a present `minimized_case`. Under `BudgetExhausted` neither is guaranteed:

- if the first reproduction itself hits the per-trial timeout, the outcome is `BudgetExhausted` with `trials == 1`;
- if the budget is exhausted before any reduction succeeds (`changed == false`), `minimized_case` is `None` and the `.expect()` below panics.

With the default 30 s budget on this small scenario that is unlikely, but on a slow/loaded CI worker it turns into an intermittent failure. Suggest gating both assertions on `status == Complete` (as the `SendAppMessage` check below already does), and asserting only `trials >= 1` plus `minimized_case.is_none()` for the `BudgetExhausted` arm.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T19:43:13Z` |
| Commit reviewed | `9ef97ae4fb0c5cd5b767eb17b19183fe9dbbe6a0` |

*(Posted as a comment because GitHub does not allow approving a PR authored under this account; the recommendation is merge.)*

This is a well-scoped root-cause fix for #1652. The ordering is exactly right: the original report, fixture candidate, and failure capsules are made durable *before* any reduction runs, and the post-minimization rewrite uses an atomic create-temp/fsync/rename in the same directory, so an interrupted worker can only ever leave a complete artifact — the `pending` status on the durable record then lets the campaign classifier distinguish scenario-execution timeouts from minimization/post-processing timeouts. The budget model (wall clock + trial count + per-trial timeout, with a checked-sub remaining-time computation and zero-validation at the CLI boundary) is simple and hard to misuse, the defaults are conservative, and the acceptance criteria from the issue are each covered by a focused test, including the expensive synthetic reproducer and the budget-exhaustion artifact-preservation test. Full CI is green.

Two non-blocking notes posted inline, both about test robustness rather than the shipped behavior:

1. `family.rs` — the 50 ms wall-clock budget in `expensive_reproducer_exhausts_the_minimization_budget_after_the_original_run` races the scheduler and can make the `trials == 2` assertion fail on a loaded machine; a multi-second wall clock keeps the per-trial timeout as the deciding bound.
2. `canonical_scenarios.rs` — the test accepts `BudgetExhausted` but then unconditionally asserts `trials > 1` and a present `minimized_case`, neither of which that status guarantees; gating those on `Complete` removes a latent intermittent failure.

Neither blocks merge.

crates/cgka-conformance-simulator/src/report.rs
erskingardner: Agreed. The durable replacement helper remains local while it has one caller. If another artifact path needs the same primitive, it should move with its mode tests into fs-private. No code change is needed for this PR.

crates/cgka-conformance-simulator/tests/canonical_scenarios.rs
```diff
@@ -2406,23 +2406,36 @@ async fn failing_generated_case_records_a_minimized_reproducer() {
...
+    assert!(matches!(
+        generated.minimization.status,
+        cgka_conformance_simulator::GeneratedScenarioMinimizationStatus::Complete
+            | cgka_conformance_simulator::GeneratedScenarioMinimizationStatus::BudgetExhausted
+    ));
+    assert!(generated.minimization.trials > 1);
+    let minimized = generated
+        .minimized_case
+        .as_ref()
         .expect("failing generated case should record a minimized case");
```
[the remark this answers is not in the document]
erskingardner: Addressed in 7937bf6e. The test now accepts at least one trial for budget exhaustion, permits an absent minimized case only for that status, and keeps the stricter progress and deterministic reduction checks under Complete. A best-so-far case produced before exhaustion is still checked and replayed.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T20:20:02Z` |
| Commit reviewed | `7937bf6e97d2a3f0d9856685894a3c82c9f6aa20` |

*(Posted as a comment because GitHub does not allow approving a PR authored under this account; the recommendation is merge.)*

Re-review of `7937bf6` ("test(simulator): remove minimizer timing assumptions"), the follow-up to my earlier inline findings:

1. **`family.rs`** — the 50 ms wall-clock race is fixed: the budget is now 5 s, so a scheduler stall can no longer prevent the second trial from starting, and the 10 ms per-trial timeout remains the deciding bound. Dropping the `started.elapsed() < 500 ms` assertion was also the right call — that bound was itself stall-sensitive (a preemption during trial 2 delays the tokio timer poll even when the timeout fires correctly), and the remaining assertions on status, trial count, and attempt count fully cover the behavior under test.
2. **`canonical_scenarios.rs`** — the assertions now match what each status guarantees: `trials >= 1` universally, `trials > 1` gated on `Complete`, and a missing `minimized_case` is permitted only under `BudgetExhausted` (with an explicit assertion to that effect) via early return. The latent intermittent failure is gone.

Both changes are minimal, test-only, and surgical. CI on the new head is still running at posting time (format/audit/policy checks already green); the remaining jobs are the same matrix that passed on `9ef97ae`, and this commit touches only test code. Prior overall assessment stands: the change is a sound root-cause fix for #1652 with durable-before-reduce ordering, atomic replacement, and conservative bounded budgets. **Merge once CI completes.**

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-03T20:35:11Z` |
| Commit reviewed | `7937bf6e97d2a3f0d9856685894a3c82c9f6aa20` |

*(Posted as a comment because GitHub does not allow approving a PR authored under this account; the recommendation is merge.)*

Rereview of the follow-up commit `7937bf6` (test-only delta on top of `9ef97ae`, which I reviewed previously). Both findings from my earlier review are resolved exactly as suggested, minimally, with no production code touched:

- `expensive_reproducer_exhausts_the_minimization_budget_after_the_original_run`: wall clock raised 50ms -> 5s so the 10ms `per_trial_timeout` is now the only deciding bound, and the stall-sensitive `started.elapsed() < 500ms` assertion (plus the unused `Instant`) was removed outright rather than loosened. The remaining assertions (`trials == 2`, `attempts == 2`, status `budget_exhausted`) still pin the intended behavior.
- `failing_generated_case_records_a_minimized_reproducer`: `trials > 1` is now gated on `complete`, and a missing `minimized_case` is only permitted for `budget_exhausted` with an explicit status assertion and early return, so the deterministic storm-removal check stays enforced wherever reduction is guaranteed.

Relevant CI is green at this head: all four Rust test shards, Simulator and vectors, Rust check, and Rust clippy (the remaining in-flight jobs are unaffected by this crate-local test change; the full matrix passed on the prior commit). No open findings.

Datawav (commented):
## Automated review

Three actionable findings.

1. Minor — Canonical minimizer test can pass without exercising reduction

Evidence: crates/cgka-conformance-simulator/tests/canonical_scenarios.rs:2406-2433 now accepts Complete or BudgetExhausted, then returns when minimized_case is absent; storm removal and minimized-report replay are checked only in the Complete branch.

Triggering state: the default 30-second wall-clock or five-second per-trial limit expires before a successful reduction.

Consequence: failing_generated_case_records_a_minimized_reproducer can pass without proving its documented production behavior: removal of the irrelevant application-message storm and reproduction by the reduced case.

Confidence: High.

Smallest sound correction: add a budget-parameterized in-process entry point and run this test with a deliberately generous budget, or add a deterministic synthetic test that must reach Complete and retain the storm-removal and replay assertions.

2. Minor — Interrupted atomic replacement can leave untracked private temporary artifacts

Evidence: the new replace_private helper in crates/cgka-conformance-simulator/src/report.rs:552-573 creates a same-directory file named with .tmp-<pid>-<id>, but removes it only when the helper returns an error. The campaign parent can kill a worker at its deadline, bypassing that cleanup. The new campaign artifact inspection records integrity errors, database bytes, timeout phase, and minimization status, but does not discover these temporary files.

Triggering state: a worker is terminated between create_new_private/write_all and rename while replacing a report or sensitive replay capsule.

Consequence: the retained output directory can contain undeclared 0600 temporary artifacts, including sensitive replay material when capture_sensitive_replay is enabled.

Confidence: High.

Smallest sound correction: after reaping a worker, have the parent remove that worker’s .tmp-<pid>-* files or record them as artifact-integrity observations.

3. Minor — The interrupted-minimization contract lacks end-to-end coverage

Evidence: crates/cgka-conformance-simulator/tests/process_campaign_runner.rs:59-67 adds assertions only for the successful case, where minimization_status is skipped and timeout_phase is absent. Separate tests cover timeout-phase classification and budget exhaustion, but none observes a worker interrupted after the original report is written and verifies the resulting on-disk pending status plus timeout_phase=minimization.

Triggering state: write the original artifacts, begin reduction, and terminate the worker before final replacement.

Consequence: regressions in the central durable-before-reduce composition could pass while each isolated helper test remains green.

Confidence: High.

Smallest sound correction: add one bounded process-campaign test that deterministically terminates a worker during reduction, then asserts a complete original report with minimization.status=pending and a summary timeout_phase of minimization.

Audit evidence:
Exact head: 7937bf6e97d2a3f0d9856685894a3c82c9f6aa20
Base: 61b46f99dabb24132f4cab82b0c872aca12b5503
Changed files inspected include crates/cgka-conformance-simulator/src/family.rs, crates/cgka-conformance-simulator/src/report.rs, crates/cgka-conformance-simulator/src/bin/cgka-conformance-campaign.rs, crates/cgka-conformance-simulator/tests/canonical_scenarios.rs, and crates/cgka-conformance-simulator/tests/process_campaign_runner.rs. Patch coverage is marked complete with no missing patch files.

Risk lenses checked: tests — assertions fail before the fix; boundary and negative cases; test exercises production behavior. Large-change decomposition and behavioral-test reasoning were also applied across the 13-file change.

Discussion classification: the earlier 50ms scheduler-race finding and inconsistent BudgetExhausted assertions are resolved at this head. The suggestion to move replace_private into fs-private is non-actionable while it has one caller. Generic clean/approval summaries add no independent evidence. The advisory claims about parser-mode selection, backward deserialization of campaign summaries, and treating a per-trial timeout as merely a rejected candidate were discarded because bounded source context either did not establish a harmful consequence or showed the behavior was intentional.

Check evidence: the bundle reports all checks completed and successful, including Rust tests (1/4 through 4/4), Rust check, clippy, format, audit, Simulator and vectors, Tamarin proofs, relay runtime and CLI tests, C bindings smoke, and installer/auth harness checks.

automated_device_gate is null. No physical-device evidence exists for this head; conclusions rest on exact-head static review and the supplied successful CI evidence.

erskingardner:
Addressed all three findings in 45edfce9:

- Added a deterministic synthetic minimizer test that must reach `Complete`, remove the full irrelevant message storm, and reproduce the retained failure trigger. The engine-backed canonical test still validates `Complete`/`BudgetExhausted` correctly and replays any best-so-far reduction.
- The campaign parent now removes atomic-replacement temporaries belonging to the exact reaped worker PID and records scan/removal failures as artifact-integrity errors.
- Added a process-level interruption test that launches the real worker path, waits for durable original artifacts with `minimization.status=pending`, kills/reaps the worker during reduction, and verifies the summary classifies `timeout_phase=minimization`.

Validation:
- `just fast-ci`
- `cargo test -p cgka-conformance-simulator --test canonical_scenarios --locked --quiet` (53 passed)
- `cargo test -p cgka-conformance-simulator --bin cgka-conformance-campaign --locked --quiet` (12 passed, 1 ignored fixture)
- Focused strict-completion, bounded canonical, cleanup, and forced-interruption tests all passed.

Datawav (approved):

PR #1665, opened 2026-09-03
Measure and bound media download latency

### LINKED ISSUE #1529: Bound bad-locator failover and reuse safe Blossom connections across download backlogs

mubarakcoded:
## Problem

Recent Android staging telemetry showed successful media downloads with 76–105 second p95 tails, but the existing `app_media_download` metric collapses every stage into one duration. It starts only after the account worker dequeues the command, so queue pressure is invisible, and it cannot distinguish runtime preparation, network setup, server response, body transfer, integrity verification, or decryption.

The earlier transport correction in #1635 already added exact-origin client reuse and a 10-second candidate-startup bound. The remaining gaps were observability, a repeatable small/large backlog regression rail, and one end-to-end deadline across locator failover. Previously each candidate fetch created a fresh 15-minute transfer deadline, so ordered fallback could multiply the overall ceiling.

## Root cause

- One completion timer covered preparation, network, integrity, and crypto while omitting account-worker queue time.
- The HTTP reader exposed no fixed phase boundaries.
- Locator failover renewed the transfer deadline for every candidate.
- No repository benchmark compared same-origin reuse against a cold transport backlog across representative payload sizes.

## Solution

- Record fixed, aggregate-only histograms for queue wait, preparation, host resolution/vetting/client lookup, response headers, first body byte, body transfer, locator failover, ciphertext verification, decryption, and plaintext verification.
- Keep the existing total metric, now including account-worker queue time.
- Share one 15-minute transfer deadline across the full ordered locator list while retaining the 10-second per-candidate startup bound and the active-transfer read-idle protection.
- Export every phase through the existing unlabeled telemetry path and keep the UniFFI and generated C snapshot surfaces in lockstep.
- Add a Criterion benchmark for 16-item same-origin backlogs with fixed 64 KiB and 16 MiB payloads, comparing one shared vetted transport with a cold transport per download. The benchmark reports completion p95, completed MiB/s, and fixed aggregate transport-phase p95 buckets.

The new points contain no URLs, hosts, account/group/message identifiers, filenames, hashes, payloads, plaintext, ciphertext, or keys. They are cumulative counts and fixed-bucket duration histograms with static metric names and no media-derived labels. Export remains off by default and requires the existing explicit opt-in, authenticated first-party endpoint, and TLS gate.

This telemetry is privacy-bounded but not anonymous: when a user opts in, the new aggregate points inherit the existing OTLP resource metadata, including the pseudonymous installation identifier, app version, OS version, and optional device model. This PR does not add or broaden those resource attributes.

## Local performance evidence

The feature-gated loopback benchmark was run with 20 direct p95 samples per policy and a reduced 10-sample Criterion window:

| Fixture | Shared direct p95 | Cold direct p95 | Shared p95 throughput | Cold p95 throughput | Interpretation |
| --- | ---: | ---: | ---: | ---: | --- |
| 16 × 64 KiB | 4.372 ms | 15.344 ms | 228.714 MiB/s | 65.171 MiB/s | Shared setup is 71.5% lower at p95. |
| 16 × 16 MiB | 211.935 ms | 206.138 ms | 1207.919 MiB/s | 1241.888 MiB/s | Transfer-dominated direct p95 is effectively at parity; do not claim a reuse win from this sample. |

For the large fixture, Criterion's central estimate favored the shared path at 169.18 ms versus 177.38 ms, with ranges of 163.83–175.17 ms and 174.58–180.14 ms respectively. Small-fixture host-setup, response-header, first-byte, and body-transfer p95s were all in the ≤1 ms bucket. The large fixture kept setup/header/first-byte in ≤1 ms while body transfer reached the 20 ms bucket. These values are a relative regression rail, not a fleet-latency prediction.

## Regression coverage split

Merged PR #1635 already covers the production transport safety matrix: connection reuse, concurrent setup single-flight, per-account isolation, address-lease expiry and re-vetting, stalled-first-locator fallback, corrupt-first-locator fallback, redirect validation, proxy and loopback policy, and preservation of the four-operation account cap.

This PR adds the remaining coverage for:

- successful network, integrity, and crypto phase recording;
- one end-to-end transfer deadline shared across locator failover, with early-completion and load-tolerant bounds;
- all media phases exported without relay or failure labels;
- fixed small and large backlog fixtures with phase, p95, and throughput evidence;
- feature-gated benchmark compilation and all-target linting.

## Verification

Focused checks passed on the current branch:

- media phase schema unit test;
- encrypted-media network/integrity/crypto phase integration test;
- shared locator deadline regression;
- all-phase unlabeled telemetry export regression;
- UniFFI snapshot conversion test;
- media-download unit tests;
- `marmot-app` all-target Clippy with the benchmark feature and warnings denied;
- small/large benchmark execution with aggregate phase and throughput output;
- repository tracing/privacy guard (4 tests);
- generated C header refresh;
- exact-head formatting and diff checks.

The full repository completion gate was intentionally not run.

Closes #1529

## Summary by CodeRabbit

* **New Features**
  * Added detailed performance tracking for media downloads, covering queueing, preparation, connection, transfer, failover, verification, decryption, and completion.
  * Exposed media-download performance snapshots through supported C and FFI interfaces.
  * Added metrics for stage durations, attempts, successes, and failures.

* **Improvements**
  * Media downloads now measure timing from enqueue through completion and provide more precise failover and timeout tracking.
  * Added documentation describing media transport behavior and testing policies.


Datawav (commented):
## Automated review

Outcome: 2 actionable findings at exact head 5f767ddcb12d4eb29e74bddf8f4c81cd01c28f10.

1. Major — Feature-gated backlog benchmark does not compile

Exact changed-line evidence:
crates/marmot-app/benches/media_download_backlog.rs adds:
    let transport = if reuse_transport {
        &shared
    } else {
        &MediaDownloadBenchmarkTransport::new()
    };
    let body = transport.fetch(&url).await.unwrap();

Triggering state: `reuse_transport == false`. The `else` arm borrows a temporary transport whose lifetime is not extended through the `if` initializer, leaving `transport.fetch(...)` with a dropped referent and producing E0716. The target is gated in crates/marmot-app/Cargo.toml by:
    required-features = ["media-benchmarks"]

Consequence: the newly introduced backlog regression rail cannot be built in its intended configuration, while ordinary checks can remain green because they need not activate that feature.

Confidence: High.

Smallest sound correction: retain an owned cold transport for the duration of the fetch without constructing one in the reuse path, for example:
    let fresh = (!reuse_transport).then(MediaDownloadBenchmarkTransport::new);
    let transport = fresh.as_ref().unwrap_or(&shared);
Then add a CI compilation check such as:
    cargo check -p marmot-app --benches --features media-benchmarks

2. Minor — Shared-deadline test has a load-sensitive timing bound

Exact changed-line evidence:
crates/marmot-app/src/media/tests.rs:1853 configures a 130 ms transfer deadline and asserts:
    started.elapsed() < Duration::from_millis(250)

Triggering state: a loaded or delayed CI executor takes more than roughly 120 ms beyond the configured deadline to schedule two candidate attempts and observe timeout completion.

Consequence: the regression test can report a timing-only false negative. It also lacks a lower bound proving that the shared deadline, rather than an unrelated early error, caused completion.

Confidence: High; the bundle’s inline discussion identifies the same current-head line and timing relationship.

Smallest sound correction: capture `elapsed` once, widen the upper bound to a conservative ceiling such as two seconds, and add a lower bound near the configured deadline with an explicit tolerance.

Audit evidence:
- Exact head SHA: 5f767ddcb12d4eb29e74bddf8f4c81cd01c28f10
- Complete patch coverage: 16 changed files, +968/−42, with no missing patch files.
- Changed files examined include crates/marmot-app/benches/media_download_backlog.rs, crates/marmot-app/src/media/blossom.rs, crates/marmot-app/src/media/mod.rs, crates/marmot-app/src/media/tests.rs, crates/marmot-app/src/runtime/account_worker.rs, and crates/marmot-c/include/marmot.h.
- Automated device gate: null. No physical-device evidence exists for this head; conclusions rest on static review and supplied CI evidence.
- Discussion classification: the inline timing comment is supported and retained as finding 2. The automated walkthrough and review summary are informational or duplicate that comment. The bundle reports complete member-feedback capture with zero member items. Other advisory claims were not retained because the patch shows intentional queue-inclusive metric semantics, purpose-specific feature gating, phase counters that need not reconcile one-to-one, or insufficient evidence of an ABI/feature defect.

Risk lenses checked: concurrency/lifecycle; network/protocol; release/build; tests.

Check evidence:
- The bundle marks check evidence complete, with all runs completed and all required checks successful.
- Successful checks include Rust check, Rust clippy, all four Rust test shards, Rust format, Rust audit, C bindings smoke, Tamarin proofs, Simulator and vectors, relay/runtime tests, and platform harness builds.
- “Publish wn-agent release” was skipped.
- The supplied checks do not establish compilation of the `media_download_backlog` target with `media-benchmarks` enabled.

crates/marmot-app/src/media/tests.rs
```diff
@@ -1777,6 +1824,35 @@ async fn stalled_first_locator_reaches_healthy_fallback_within_startup_bound() {
...
+        .await
+        .expect_err("candidate failover must not renew the transfer deadline");
+    first_server.abort();
+    second_server.abort();
+
+    assert!(error.to_string().contains("timed out"));
+    assert!(
+        started.elapsed() < Duration::from_millis(250),
+        "two stalled candidates must share the configured end-to-end deadline"
+    );
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in 1063110d. The regression now records elapsed once, rejects completion materially before the configured transfer deadline with an explicit tolerance, and uses a two-second upper ceiling so loaded CI scheduling cannot produce a timing-only false failure. The focused test passes at the new head.

mubarakcoded:
Addressed the automated review at exact head `1063110d`:

- Made the cold benchmark transport ownership explicit. The feature-gated benchmark target compiles with `cargo check -p marmot-app --bench media_download_backlog --features media-benchmarks`.
- Hardened the shared-deadline regression with a tolerant lower bound and a two-second CI ceiling. The focused test passes.
- Expanded the privacy regression to require all four exported points for every new media phase to carry neither relay nor failure attributes.
- Added substantive documentation for the touched policy and timing boundaries.

The PR description now states the precise privacy boundary: no media-derived identifiers or content are exported, but opted-in metrics inherit the existing pseudonymous OTLP resource metadata. No production timeout, crypto, host-safety, or resource-attribute behavior changed in this follow-up.

crates/marmot-app/benches/media_download_backlog.rs
```diff
@@ -0,0 +1,122 @@
...
+        }
+    });
+    (format!("http://localhost:{port}/blob.bin"), server)
+}
+
+/// Download one fixed backlog either through one shared transport or through a
+/// fresh transport per item while keeping payload and server behavior equal.
+async fn run_backlog(reuse_transport: bool) {
+    let (url, server) = spawn_keep_alive_server(BACKLOG_SIZE).await;
+    let shared = MediaDownloadBenchmarkTransport::new();
```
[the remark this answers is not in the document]
mubarakcoded: Fixed in 8b3d4c66. The shared transport is now constructed only for the reuse policy, while the cold policy owns exactly one fresh transport per item. This removes the unused setup from the cold measurement. Focused verification passed: all-target clippy with media-benchmarks, the shared-deadline regression, unlabeled telemetry export regression, and media-download tests.

mubarakcoded:
Linked-issue warning reviewed against repository history: merged #1447 owns the off-worker media lane, bounded-capacity reservation, and cancellation coverage; merged #1635 owns exact-origin reuse, concurrent setup single-flight, lease expiry/re-vetting, and stalled/corrupt locator fallback while preserving that lane and cap. The #1529 follow-up comment explicitly leaves phase telemetry, small/large backlog measurement, and before/after evidence to this PR. This head supplies those remaining items plus the shared end-to-end locator deadline, so duplicating the already-merged transport implementation and tests here would add churn rather than coverage.

Datawav (commented):
## Automated review

1 actionable finding.

Major — Shared-deadline regression test does not distinguish deadline reuse from per-locator renewal

Evidence: In crates/marmot-app/src/media/tests.rs, the changed assertion ending at line 1865 accepts elapsed >= TRANSFER_DEADLINE - EARLY_COMPLETION_TOLERANCE and elapsed < Duration::from_secs(2). The fixture has two stalled locators, TRANSFER_DEADLINE = 130 ms, and an 80 ms candidate-startup bound. Bounded head context in crates/marmot-app/src/media/blossom.rs computes operation_remaining as remaining.min(startup_remaining), with a new startup deadline for each fetch. Therefore:

  • Correct shared deadline: approximately 130 ms.
  • Regressed per-locator deadline: approximately 160 ms because each attempt reaches its 80 ms startup bound.

Both satisfy the changed approximately 100 ms–2 s window and produce the expected timeout error.

Triggering state: More than one locator stalls, and the implementation accidentally renews the transfer deadline for each locator while the shorter per-candidate startup timeout remains active.

Consequence: The headline end-to-end latency bound can regress into a locator-count-multiplied deadline while this regression test remains green.

Confidence: High.

Smallest sound correction: Make the transfer deadline—not candidate startup—the binding timeout, and separate shared from renewed timing by a CI-tolerant margin. For example, use candidate startup of 1 s, transfer deadline of 300 ms, and three stalled locators; retain a tolerant lower bound near 300 ms and require completion below 700 ms, while renewed deadlines would take approximately 900 ms.

Audit evidence:

  Exact head: 07f706c115bb2651e9075afac7926cba8d5554f0

  Changed files reviewed include:
  • crates/marmot-app/src/media/tests.rs
  • crates/marmot-app/src/media/blossom.rs
  • crates/marmot-app/benches/media_download_backlog.rs
  • crates/marmot-app/src/relay_telemetry_export/tests.rs

  The advisory’s throughput_p95_mib_s naming concern was discarded as terminology-only; the value is deterministically computed from bytes divided by p95 latency. The unused Criterion telemetry and large-fixture volume notes were discarded as non-actionable observations without demonstrated correctness impact. The cold-transport construction concern is resolved: the exact-head benchmark conditionally creates the shared transport, while the cold path creates one fresh transport per item.

Discussion classification:

  • The timing-stability request in discussion_r3928823350 is addressed by the tolerant lower bound and wider ceiling, but the resulting assertion does not protect the shared-deadline invariant; the actionable finding above remains.
  • discussion_r3929698421 is resolved by conditional shared-transport construction.
  • The privacy claim is supported by the expanded export regression, which checks all four points for each listed media phase for absent relay and failure attributes.
  • The linked-issue ownership/history statement is external historical context, not an exact-head code finding.
  • Automated “no actionable comments” summaries and unresolved-thread notices provide no additional technical finding.

Risk lenses checked: network/protocol, tests

  • timeouts, retry and duplicate delivery: checked; the shared-deadline regression gap is reported above, with no changed retry or duplicate-delivery behavior found.
  • partial/malformed peer data: checked; this follow-up does not change parsing or malformed-response handling.
  • version and wire compatibility: checked; no changed wire format or protocol-version behavior found.
  • assertions fail before the fix: checked; the shared-deadline test would still pass under per-locator deadline renewal.
  • boundary and negative cases: checked across stalled locators, small/large backlog fixtures, and unlabeled telemetry assertions.
  • test exercises production behavior: checked; the test reaches the production fetch path through an explicit test policy, but its timing window is not discriminating.

Check evidence: The bundle reports complete check evidence with all checks completed and its aggregate successful state true. Named successful checks include Rust tests (1/4 through 4/4), Rust clippy, Rust format, Rust check, Rust audit, Relay CLI E2E, Relay runtime tests, Simulator and vectors, C bindings smoke, and the platform build jobs. “Publish wn-agent release” completed with a skipped conclusion and is not correctness evidence.

The automated device gate status is capability_checked_not_installable and is bound to the earlier 5f767ddcb12d4eb29e74bddf8f4c81cd01c28f10 revision, not this exact head. No physical-device evidence exists for this head. Conclusions therefore rest on static review and the supplied CI evidence.

mubarakcoded:
@Datawav thanks for the careful pass. I rechecked the Major shared-deadline finding against the full fixture, and I believe it overlooks the healthy third locator.

The test uses two stalled locators followed by a healthy locator. With the intended shared deadline, the first stall consumes about 80 ms and the second consumes the remaining portion of the 130 ms budget, so the operation times out before reaching the healthy locator and `.expect_err(...)` passes. If the transfer deadline were accidentally renewed per locator, both stalled locators would hit their individual startup bounds and the healthy third locator would then return the expected ciphertext; the call would succeed and `.expect_err(...)` would fail the regression test.

So the wide wall-clock ceiling is only for CI scheduling tolerance; success-versus-error is what distinguishes shared-budget behavior from per-locator renewal. The finding would be valid if the fixture ended after the two stalled locators, but the final healthy locator makes the current test fail under the proposed regression.

Datawav (commented):
## Automated review

2 actionable findings.

1. Medium — Oversized responses record a zero-duration body-transfer failure
   Evidence: `crates/marmot-app/src/media/blossom.rs:292` adds `telemetry.record(AppPerformanceOperation::MediaDownloadBodyTransfer, Duration::ZERO, false)` in the `content_length > max_bytes` branch, before any response body is read.
   Trigger: A server declares a `Content-Length` above the configured download limit.
   Consequence: The body-transfer histogram receives artificial zero-duration samples for transfers that never began, biasing latency telemetry and obscuring whether failures occurred during headers, policy validation, or actual transfer.
   Confidence: High.
   Smallest sound correction: Record this rejection only in the response-header/policy phase, or omit body-transfer duration recording until the first body read begins.

2. Medium — The feature-gated benchmark is not explicitly covered by supplied CI
   Evidence: `crates/marmot-app/Cargo.toml:71` adds `media_download_backlog` with `required-features = ["media-benchmarks"]`; the supplied check list contains Rust check, clippy, format, and test jobs but no check explicitly enabling `media-benchmarks`.
   Trigger: Benchmark-only APIs or imports drift while ordinary/default and OTLP targets continue compiling.
   Consequence: The benchmark can silently become uncompilable, weakening the reproducible backlog-regression evidence this change introduces.
   Confidence: Medium.
   Smallest sound correction: Add a CI command equivalent to `cargo check -p marmot-app --bench media_download_backlog --features media-benchmarks`, or include that feature in an existing all-target check.

Audit evidence:
- Exact head: `07f706c115bb2651e9075afac7926cba8d5554f0`
- Patch coverage is marked complete across 16 changed files, including `crates/marmot-app/src/media/blossom.rs`, `crates/marmot-app/src/media/mod.rs`, `crates/marmot-app/src/runtime/account_worker.rs`, and `crates/marmot-app/Cargo.toml`.
- The existing aggregate `media_download` timing now begins at command enqueue and therefore measures end-to-end latency including queue wait; this is a documented behavioral change, not a separate defect.
- The hard-coded successful locator-failover sample is consistent with measuring completion of the failover transition, so the advisory claim that every such sample must be a failure was discarded.
- The C snapshot fields alter layout, but the bundle supplies no stable cross-version C ABI requirement establishing an actionable compatibility defect.

Discussion classification:
- The timing-test concern is stale/resolved: the exact-head regression records elapsed time once, uses an early-completion tolerance, and has a two-second scheduling ceiling. Its healthy third locator makes success-versus-timeout distinguish shared from renewed per-locator deadlines.
- The cold-benchmark extra-construction concern is stale/resolved at this head.
- The linked-issue decomposition is supported: earlier work owns transport reuse, bounded capacity, cancellation, host vetting, and fallback; this patch owns phase telemetry, backlog measurement, and the shared end-to-end deadline.
- Mechanically open threads containing those resolved claims do not create additional findings.

Risk lenses checked:
- `release/build`: untrusted-input and least-privilege CI; artifact identity/reproducibility; all required paths trigger checks.
- Large-change decomposition and behavioral-test reasoning were also applied.

Check evidence:
The supplied exact-head CI evidence marks Rust tests (all four shards), Rust check, clippy, format, audit, C bindings smoke, relay runtime/E2E, simulator and vectors, Tamarin proofs, and platform harness builds completed successfully. `Publish wn-agent release` completed with a skipped conclusion.

Automated device gate status is `capability_checked_not_installable`: no physical-device evidence exists for this head. Conclusions therefore rest on static review and the supplied CI evidence.

mubarakcoded:
@Datawav addressed the valid oversized-response finding in `e6fd5906`.

A response rejected from its declared `Content-Length` now fails before creating any body-transfer timing sample. The new regression exercises the observed fetch path and verifies that response-header telemetry is present while `media_download_body_transfer.attempts` remains zero. The functional size rejection is unchanged.

Focused verification passed: the new regression, all eight oversized-input tests, the existing successful phase-recording test, the shared locator-deadline test, formatting/diff checks, and `cargo clippy -p marmot-app --lib --tests -- -D warnings`.

The benchmark CI concern is already covered by `.github/workflows/ci.yml`: Rust check and Clippy both run `--workspace --all-targets --all-features`, which activates `media-benchmarks` and compiles/lints the required-feature benchmark target.

Datawav (dismissed):

crates/marmot-app/Cargo.toml
```diff
@@ -56,8 +58,14 @@ url.workspace = true
...
 keyring-core.workspace = true
 marmot-forensics.workspace = true
 nostr-relay-builder.workspace = true
 tempfile.workspace = true
 tokio = { workspace = true, features = ["test-util"] }
+
+[[bench]]
+name = "media_download_backlog"
+harness = false
+required-features = ["media-benchmarks"]
```
erskingardner, line 71: This bench target is never compiled by CI: `just check`/`just clippy` run `--workspace --all-targets` without `media-benchmarks`, and `required-features` skips the target, so the benchmark can silently bit-rot after this PR lands. The repo precedent for regression-rail benchmarks (`bench-startup` for mdk#1161, the founding-image bench for mdk#1485) is `cargo test --release --ignored` tests, which are always compiled by the existing gates. Either wire a feature-gated check/clippy leg into the justfile (mirroring the `otlp-export` pattern) or reshape this as an ignored test so the rail the issue asks for stays compiled.

crates/marmot-app/src/media/blossom.rs
```diff
@@ -830,29 +957,95 @@ async fn read_limited_blossom_body_until(
...
+                    telemetry,
+                    AppPerformanceOperation::MediaDownloadFirstByte,
+                    first_byte_started,
+                    true,
+                );
+            }
+            record_download_phase(
+                telemetry,
+                AppPerformanceOperation::MediaDownloadBodyTransfer,
+                body_started.unwrap_or_else(Instant::now),
```
erskingardner: Minor: on a 0-byte body this records a `MediaDownloadBodyTransfer` success sample with a synthetic ~0ns duration (`Instant::now()` as the start), the same class of artifact `e6fd5906` removed for the oversized-`Content-Length` path. Valid encrypted media can never be empty (the AEAD tag alone is 16 bytes), so this only fires for corrupt/empty responses that then fail ciphertext verification — but it still puts a zero-duration "success" into the transfer histogram. Consider skipping the body-transfer sample when `body_started` is `None`.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T10:21:14Z` |
| Commit reviewed | `e6fd5906317e1cbf3a56ece9317d54c72d7187f9` |

## Summary

The core of this PR is in good shape. The phase boundaries are fixed and aggregate-only, the metric schema is static with no dynamic labels (matching the observability invariants), the shared 15-minute end-to-end deadline across locator failover is correctly enforced by the outer `tokio::time::timeout` while the per-candidate startup bound and read-idle protection are preserved, and the UniFFI/C snapshot surfaces move in lockstep with exhaustive destructuring. The regression tests for the shared deadline, the all-phase unlabeled export, and the oversized-`Content-Length` no-sample case are the right rails. The four-operation account cap and the fail-closed paths are untouched.

One serious concern, plus one minor note, both inline:

1. **The benchmark is never compiled in CI.** `just check`/`just clippy`/`just test` do not enable `media-benchmarks`, and `required-features` drops the target from `--all-targets`, so the regression rail this issue exists to create can silently stop compiling the moment it merges. The established repo pattern for benchmarks (`bench-startup`, founding-image) is always-compiled `cargo test --release --ignored` tests; either follow that or add a feature-gated check/clippy leg mirroring the `otlp-export` one.
2. Minor: the empty-body completion path still records a synthetic zero-duration body-transfer success sample — the same artifact class the head commit removed for oversized responses.

Privacy posture verified: no URLs, hosts, identifiers, hashes, or payloads in any new metric; export remains opt-in and unlabeled. Not re-running locally; relying on the green CI matrix at the reviewed head.

crates/marmot-app/src/media/blossom.rs
```diff
@@ -830,29 +957,95 @@ async fn read_limited_blossom_body_until(
...
+                };
+                record_download_phase(telemetry, operation, started_at, false);
+                return Err(error);
+            }
+        };
         let Some(chunk) = next else {
+            if body_started.is_none() {
+                record_download_phase(
+                    telemetry,
+                    AppPerformanceOperation::MediaDownloadFirstByte,
```
erskingardner, line 1029: Non-blocking edge case: when the stream ends without delivering any bytes (a 200 with an empty body), this records `FirstByte` as a *success* plus a zero-duration `BodyTransfer` success, even though no first byte ever arrived. Encrypted media is never empty, so the follow-up `CiphertextVerify` failure will dominate in practice, but the phase histogram semantics are slightly off for this path. If you want strict semantics, only record `FirstByte` success when `body_started.is_some()` (i.e., move the stream-end record out of the `body_started.is_none()` branch).

erskingardner (dismissed):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T10:22:32Z` |
| Commit reviewed | `e6fd5906317e1cbf3a56ece9317d54c72d7187f9` |

Verified at the exact head that every earlier actionable finding is resolved: the benchmark's cold path now owns its transport per item (the E0716 pattern is gone), the oversized `Content-Length` rejection no longer emits a zero-duration body-transfer sample (covered by `oversized_content_length_does_not_record_body_transfer_timing`), and the shared-deadline test now discriminates: with a healthy third locator, per-locator deadline renewal would produce a success and fail `expect_err`, while the shared 130 ms bound times out inside the 100 ms–2 s window. The feature-gated bench also compiles in CI via the existing `cargo check/clippy --workspace --all-targets --all-features` jobs.

Telemetry shape matches the repo's observability invariant: static metric names, fixed-bucket histograms, aggregate counts only, no relay/failure/media-derived labels, and the export regression asserts unlabeled points for every phase. The queue-inclusive redefinition of `app_media_download` is a documented, intentional semantics change. UniFFI and generated C surfaces are in lockstep.

One non-blocking inline note on an empty-body edge case; nothing else blocking.

dannym-arx (changes requested):
two issues:

1. Shared-deadline timeouts can disappear from phase metrics. The outer timeout cancels fetch_blossom_blob_with_observer when the shared budget expires (media/mod.rs L964-L969) while the inner fetch starts a fresh full transfer timeout (blossom.rs L455-L459). 

On later locators, the outer remaining budget is shorter, so it wins and drops the instrumented future before the active headers/first-byte/body phase records a failure. This hides precisely the tail timeouts the PR is meant to diagnose. Pass the shared absolute deadline into the instrumented fetch and let it enforce/record expiration internally.

2. Empty responses report phases that never occurred as successful. When the first response.chunk() returns None, the code records both first-byte and body-transfer success—even though no byte arrived—and manufactures a near-zero body duration (blossom.rs L1000-L1014). A 200 with an empty body is plausible during server failure and will pollute both histograms before ciphertext verification rejects it. Record first-byte failure and omit body-transfer timing when EOF precedes the first chunk.

Datawav (commented):
## Automated review

Two actionable findings.

1. Minor — Shared-deadline expiry can omit the active phase’s timeout sample

Exact changed-line evidence: crates/marmot-app/src/media/mod.rs:964–969 computes `remaining` from the shared `download_deadline` and wraps `blossom::fetch_blossom_blob_with_observer(...)` in `tokio::time::timeout(remaining, ...)`. The called path in crates/marmot-app/src/media/blossom.rs:455–459 creates a later inner deadline from the full `transport.transfer_timeout`.

Triggering state: A response continues yielding data often enough to avoid the read-idle timeout, but remains active until the shared end-to-end deadline expires.

Consequence: The outer timeout cancels the instrumented future before the active headers, first-byte, or body-transfer phase records its negative sample. The overall operation reports the timeout, but the phase telemetry added to diagnose long-tail downloads loses the decisive observation.

Confidence: High.

Smallest sound correction: Pass the shared absolute deadline, or its computed remaining duration, into the instrumented Blossom fetch and have the inner phase enforce and record expiry before returning the timeout error.

2. Minor — Empty 200 responses record phases that never occurred as successful

Exact changed-line evidence: crates/marmot-app/src/media/blossom.rs:1000–1014 handles `next == None` by recording `MediaDownloadFirstByte` with `success = true` when `body_started.is_none()`, then records `MediaDownloadBodyTransfer` from `body_started.unwrap_or_else(Instant::now)` with `success = true`.

Triggering state: A server returns HTTP 200 with `Content-Length: 0`, or closes the stream before producing any chunk.

Consequence: First-byte telemetry reports success despite receiving no byte, while body-transfer telemetry gains a synthetic near-zero-duration success. Corrupt empty responses can therefore bias both phase distributions downward before ciphertext verification rejects the payload.

Confidence: High.

Smallest sound correction: When end-of-stream occurs before the first chunk, record first-byte as unsuccessful or omit it, omit body-transfer entirely, and add a regression asserting zero body-transfer attempts for an empty 200 response.

Audit evidence:

Exact head: e6fd5906317e1cbf3a56ece9317d54c72d7187f9

Changed files examined include crates/marmot-app/src/media/blossom.rs, crates/marmot-app/src/media/mod.rs, crates/marmot-app/src/media/tests.rs, crates/marmot-app/benches/media_download_backlog.rs, and crates/marmot-app/src/relay_telemetry_export.rs. Bundle coverage is complete for all 16 changed files with no missing patches.

Risk lenses checked:
- concurrency/lifecycle — cancellation and stale callbacks; ordering/race windows; process or lifecycle restart
- network/protocol — timeouts, retry and duplicate delivery; partial/malformed peer data; version and wire compatibility
- release/build — untrusted-input and least-privilege CI; artifact identity/reproducibility; all required paths trigger checks
- tests — assertions fail before the fix; boundary and negative cases; test exercises production behavior

Discussion classification: The timing-bound regression, unused cold-path transport, and oversized-`Content-Length` telemetry issues are resolved at this head. The benchmark-CI claim was not retained because the supplied exact-head source context does not expose the workflow invocation and named green checks alone do not establish feature coverage. The locator-failover success-counter claim was discarded because the changed code consistently treats the metric as a successful transition to another candidate; no contrary metric contract is supplied. Per-redirect phase cardinality, C-struct layout growth, and benchmark-harness robustness remain non-product-path notes rather than demonstrated changed-line defects.

Check evidence: The bundle reports complete check evidence with every run completed successfully, including Required CI, Rust check, Rust clippy, Rust format, Rust audit, all four Rust-test shards, C bindings smoke, Simulator and vectors, Tamarin proofs, Relay CLI E2E, and Relay runtime tests.

Physical-device scope: `automated_device_gate.status` is `capability_checked_not_installable`; therefore no physical-device evidence exists for this head. Conclusions rest on static review of the supplied patch and bounded source context plus the supplied CI evidence.

crates/marmot-app/Cargo.toml
mubarakcoded: @erskingardner rechecked against the actual workflow rather than the local just recipes. CI lines 144 and 173 run cargo check and Clippy with --workspace --all-targets --all-features, so media-benchmarks is enabled and the required-feature target is included. I also reran cargo check -p marmot-app --bench media_download_backlog --features media-benchmarks at 2e321321; it passes. No CI change is needed.

crates/marmot-app/src/media/blossom.rs
mubarakcoded: @erskingardner fixed in 2e321321. EOF before the first chunk now records FirstByte as failed and emits no BodyTransfer sample. The new empty_response_records_first_byte_failure_without_body_transfer_timing regression covers the exact 200/empty-body path.

crates/marmot-app/src/media/blossom.rs
mubarakcoded: @erskingardner fixed in 2e321321 with strict phase semantics: no byte means a failed FirstByte attempt, and BodyTransfer remains absent. The focused regression asserts both outcomes.

mubarakcoded:
@dannym-arx both blocking findings are fixed in 2e321321.

- The shared absolute transfer deadline is now passed into the instrumented fetch. The inner host, response-header, first-byte, and body paths enforce it directly, so an active later locator records the failure instead of being cancelled by an unobserved outer timeout. A regression proves two stalled locators produce two failed response-header samples under one 130 ms end-to-end budget.
- A 200 response that reaches EOF before any byte now records FirstByte as failed and emits no BodyTransfer sample. A dedicated empty-response regression pins that behavior.

@erskingardner the empty-body notes are covered by the same second fix. The benchmark CI concern did not require a code change: the workflow runs check and Clippy with --workspace --all-targets --all-features, and the explicit feature-gated benchmark check also passes at this head.

Focused verification at this head: all 79 marmot-app media tests passed, the feature-gated benchmark target compiled, cargo fmt check passed, git diff check passed, and cargo clippy -p marmot-app --lib --tests -- -D warnings passed.

Datawav (approved):

dannym-arx (approved):

PR #1666, opened 2026-09-04
Fix/anchorless rewind base

jgmontoya:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved fork and branch convergence by selecting the safest available retained state during rewinds.
  - Prevented unrecoverable replay scenarios from applying inconsistent state.
  - Preserved locally authored changes and proposal-related history during replay.
  - Improved handling of missing historical state by halting safely and allowing later convergence.

- **Documentation**
  - Updated architecture and testing guidance to clarify rewind behavior, convergence safeguards, and required test-policy settings.

- **Tests**
  - Added regression coverage for fork resolution, rewind boundaries, joiner advances, missing state, and distributed convergence.

### DIFF
--- a/crates/cgka-engine/src/openmls_projection.rs
+++ b/crates/cgka-engine/src/openmls_projection.rs
@@ -1290,7 +1290,12 @@ fn seed_stored_openmls_graph_inputs<S: StorageProvider>(
     }
 
     let historical_start_epoch = historical_replay_start_epoch(&commit_messages, current_epoch);
-    let replay_start_epoch = historical_start_epoch.unwrap_or(current_epoch);
+    let replay_start_epoch = match historical_start_epoch {
+        Some(fork_epoch) => {
+            retained_anchor_rewind_base(storage, group_id, fork_epoch, retained_anchor_epoch)?
+        }
+        None => current_epoch,
+    };
     let commit_messages: Vec<_> = if historical_start_epoch.is_some() {
         commit_messages
     } else {
@@ -1549,6 +1554,39 @@ fn canonicalize_stored_openmls_messages_from_current<S: StorageProvider>(
     append_dropped_messages(&mut result, path_result.invalid_commit_drops);
     append_missing_parent_deferred_commits(&mut result, path_result.unmaterialized_commit_ids);
     Ok(result)
+}
+
+/// The epoch a retained-anchor rewind can actually land on when replaying a fork
+/// at `fork_epoch`: the greatest retained anchor in `floor_epoch..=fork_epoch`.
+///
+/// Anchors exist only at the epochs a device *stopped* at. A convergence apply
+/// that traverses several commits retains one at its start epoch and one at its
+/// final tip — snapshots cannot nest inside the apply transaction — so every
+/// epoch it passed through is anchorless by construction. Demanding an anchor at
+/// exactly the fork epoch therefore halted an ordinary catch-up shape, and the
+/// halt froze the tip so the drop horizon never aged the rival out. Rewinding to
+/// the anchor below the fork and replaying the already-applied commits back up
+/// to it adjudicates the rival instead.
+///
+/// `floor_epoch` keeps the base inside the window whose commits are still
+/// available to replay. With no anchor in range this returns `fork_epoch`
+/// unchanged, so the caller's rollback still fails closed on
+/// `MissingRetainedAnchor` — the residual case of genuinely lost recovery
+/// material.
+fn retained_anchor_rewind_base<S: StorageProvider>(
+    storage: &S,
+    group_id: &GroupId,
+    fork_epoch: u64,
+    floor_epoch: u64,
+) -> Result<u64, OpenMlsProjectionError> {
+    Ok(storage
+        .list_group_snapshots(group_id)
+        .map_err(|e| OpenMlsProjectionError::Storage(format!("{e:?}")))?
+        .iter()
+        .filter_map(|name| retained_anchor_epoch_from_snapshot_name(name))
+        .filter(|epoch| (floor_epoch..=fork_epoch).contains(epoch))
+        .max()
+        .unwrap_or(fork_epoch))
 }
 
 fn historical_replay_start_epoch(
@@ -2363,11 +2401,21 @@ pub(crate) fn apply_openmls_canonicalization_result_with_profile_policy<S: Stora
         checkpoint_realizable_own_commit_prefix(storage, result, &applied_prefix)?;
     let mut skipped_prefix = applied_prefix;
     skipped_prefix.extend(own_checkpoint_prefix.commit_ids.iter().cloned());
-    let apply_start_epoch = match own_checkpoint_prefix.realized.as_ref() {
+    let mut apply_start_epoch = match own_checkpoint_prefix.realized.as_ref() {
         Some(realized) => realized.resulting_epoch,
         None => apply_start_epoch_for_canonicalization_result(storage, result, &skipped_prefix)?
             .unwrap_or(current_epoch),
     };
+    // A checkpoint restore is addressed by commit, not by epoch, so only the
+    // anchor rewind needs a base that actually exists.
+    if own_checkpoint_prefix.realized.is_none() && apply_start_epoch < current_epoch {
+        (apply_start_epoch, skipped_prefix) = rebase_apply_onto_retained_anchor(
+            storage,
+            group_id,
+            skipped_prefix,
+            apply_start_epoch,
+        )?;
+    }
     // The own-checkpoint case must restore even when its epoch equals the live
     // tip: the live tip may be a different branch's state at that epoch.
     let rewind_to_retained_anchor = apply_start_epoch < current_epoch;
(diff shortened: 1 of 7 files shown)


Datawav (commented):
## Automated review

1. Medium — Rewind apply derives source epochs from mutable row metadata

crates/cgka-engine/src/openmls_projection.rs, changed line 2759

Evidence: AppliedPrefixCommit.source_epoch is populated from record.epoch.0. Elsewhere in this change, epochs_with_unreplayable_proposals explicitly uses projection.source_epoch from the authenticated wire payload because row metadata can lag. The pass seeder likewise builds replayable_epochs from the projected source_epoch, while the apply path uses record.epoch.

Trigger: A previously processed commit has lagging row epoch metadata while its authenticated wire payload contains the correct source epoch.

Consequence: The pass and apply can compute different replay floors. A complete replayable chain can appear to contain a gap, causing a false MissingRetainedAnchor result and leaving the group durably unable to converge despite a usable lower anchor.

Smallest sound correction: Decode/project each applied-prefix commit and populate AppliedPrefixCommit.source_epoch from its authenticated projection.source_epoch. Use row metadata only for storage lookup/state filtering.

Regression test: Store a processed commit whose row epoch lags its authenticated wire source epoch, retain only the lower usable anchor, and verify convergence replays the complete prefix successfully. This would fail before the correction by calculating the wrong floor.

Confidence: High.


2. Medium — The normative convergence invariant omits newly outcome-determining replay state

docs/marmot-architecture/cgka-engine-spec.md, changed lines 274–281

Evidence: The changed commit rules now select the greatest retained snapshot within a replay floor and make the outcome depend on stored-chain gaps, own commits, and consumed proposals. The surrounding Core Invariant still says engines with the same singular “retained anchor” and pending messages must produce the same result. It does not include the retained snapshot set or effective replay floor.

Trigger: Two conforming implementations share the same oldest retained-anchor epoch and pending messages but retain different in-window snapshot sets or derive different replay floors.

Consequence: They can select different rewind bases—or one can return MissingRetainedAnchor—while both nominally satisfy the stated invariant. That creates a protocol compatibility/convergence specification gap.

Smallest sound correction: Define “same retained anchor” to include the retained snapshot set, applicable fork range, and replay floor inputs. Reflect those inputs in the conformance model and add a case where equal oldest-anchor epochs but different intermediate snapshot sets produce distinguishable outcomes before the specification fix.

Confidence: High.


No physical-device evidence exists for this head because no automated device-gate record was supplied.

Audit evidence:
Head SHA: f2abc16ccb777102d68a0589d68cafcd65a283dd
Coverage: complete; missing_patch_files is empty.
Changed patches inspected:
  crates/cgka-engine/AGENTS.md
  crates/cgka-engine/src/openmls_projection.rs
  crates/cgka-engine/tests/AGENTS.md
  crates/cgka-engine/tests/distributed_convergence.rs
  crates/cgka-engine/tests/fork_detection.rs
  docs/marmot-architecture/cgka-engine-spec.md
  docs/marmot-architecture/distributed-convergence.md

Risk lenses checked: tests
  assertions fail before the fix: The added anchorless-rewind tests exercise the former exact-anchor failure, while the two regression cases specified above target uncovered pre-correction failures.
  boundary and negative cases: Reviewed missing anchors, traversed epochs, chain gaps, own commits, consumed by-reference proposals, epoch zero, joiner advancement, restart behavior, and fail-closed state preservation.
  test exercises production behavior: The integration tests call the production convergence path; the small lowest_replayable_epoch unit test covers only helper behavior and does not cover stale row metadata or cross-implementation invariant inputs.

Check evidence: Supplied evidence is available and complete. All listed required CI checks concluded successfully; “Publish wn-agent release” was skipped. The successful evidence includes Rust test shards, Rust check, clippy, formatting, audit, simulator/vectors, convergence-policy assertions, CLI/relay tests, Tamarin proofs, and platform builds.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -2361,13 +2482,24 @@ pub(crate) fn apply_openmls_canonicalization_result_with_profile_policy<S: Stora
...
     skipped_prefix.extend(own_checkpoint_prefix.commit_ids.iter().cloned());
-    let apply_start_epoch = match own_checkpoint_prefix.realized.as_ref() {
+    let mut apply_start_epoch = match own_checkpoint_prefix.realized.as_ref() {
         Some(realized) => realized.resulting_epoch,
         None => apply_start_epoch_for_canonicalization_result(storage, result, &skipped_prefix)?
             .unwrap_or(current_epoch),
     };
+    // A checkpoint restore is addressed by commit, not by epoch, so only the
+    // anchor rewind needs a base that actually exists.
+    if own_checkpoint_prefix.realized.is_none() && apply_start_epoch < current_epoch {
```
erskingardner, line 2494: [P1] Exercise this through the production ingest seam before enabling the rebase. The headline regression calls `buffer_openmls_convergence_message_at`, which writes the authenticated source epoch into `MessageRecord.epoch`. Real `ingest_group_message` does something different when the exact source anchor is absent: lines 977-997 refuse convergence admission, and `unadjudicable_fork_rival_without_anchor` deliberately leaves the already-persisted row keyed to the current epoch (lines 2268-2283). `apply_start_epoch_for_canonicalization_result` then reads that mutable row epoch, so for the actual late-rival path `apply_start_epoch == current_epoch`; this block is skipped and the source-epoch commit is replayed against the live tip, producing `WrongEpoch` instead of the advertised recovery. Replace the test injection with `carol.ingest(bob_rival)` plus the scheduled convergence drain, and derive every apply/prefix source epoch from `project_mls_message` rather than row metadata (or update ingress admission to recognize the viable lower anchor).

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1551,6 +1571,107 @@ fn canonicalize_stored_openmls_messages_from_current<S: StorageProvider>(
...
+/// committable by a commit whose source epoch is `E`, so a proposal's own epoch
+/// IS the source epoch of the commit that could have consumed it. Accepting
+/// that proposal marks it `Processed`, which
+/// [`record_state_is_canonicalization_input`] excludes, and a retained anchor
+/// predates the proposal's arrival — so neither the seeded pass inputs nor a
+/// rolled-back proposal store can offer it to a replay again.
+///
+/// A safe over-approximation: every `Processed` proposal's epoch is reported,
+/// including one no commit on the selected branch actually referenced. That
+/// only floors the rewind higher, which fails closed; missing an epoch here
```
erskingardner, line 1621: [P1] Do not convert a normal multi-commit apply into permanent `Unrecoverable` merely because this generic input filter excludes a `Processed` proposal. The proposal bytes are still durably present, and the adopted `protocol-core/retained-history.md` contract requires enough retained state and dependencies to reconstruct every epoch inside `max_rewind_commits`; `MissingRetainedAnchor` is reserved for material actually lost from storage. This PR instead adds a test asserting that an ordinary proposal-consuming catch-up creates an unreplayable in-window epoch, then weakens the MDK docs around that behavior. Re-admit the exact consumed proposal dependencies when replaying the already-applied prefix, or retain the intermediate state. If the intended change is really to weaken the adopted retention/convergence contract, it needs a versioned required capability rather than silently changing v1 behavior.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T06:45:01Z` |
| Commit reviewed | `f2abc16ccb777102d68a0589d68cafcd65a283dd` |
| Model | `gpt-5.6-sol` |
| Reasoning level | `high` |

The premise is real and the ordering/tiebreak rule itself is unchanged, but this head is not safe to merge yet. The new lower-anchor path is bypassed by the real ingest seam, and the proposal floor turns normally retained history into a permanent local halt. Green CI does not exercise either boundary.

Datawav (commented):
## Automated review

1. P1 — Production ingest bypasses the new lower-anchor replay path

   Evidence: crates/cgka-engine/src/openmls_projection.rs, changed line 2494, in the new apply-start/prefix-replay logic. The implementation derives apply_start_epoch from MessageRecord.epoch. For a late rival received through ingest_group_message without an exact source anchor, the persisted row remains keyed to the receiver’s current epoch rather than the rival’s authenticated source epoch. The added regression in crates/cgka-engine/tests/distributed_convergence.rs injects the rival through buffer_openmls_convergence_message_at, which directly supplies the source epoch and therefore does not exercise this production behavior.

   Trigger: A client advances through several commits in one convergence pass, leaving an intermediate epoch without its own anchor, and later receives a rival rooted at that traversed epoch through the normal ingest API.

   Consequence: apply_start_epoch equals the live current epoch, prefix replay is skipped, and the source-epoch rival is replayed against the live tip. OpenMLS returns WrongEpoch, so the advertised lower-anchor recovery does not converge on the production path.

   Smallest sound correction: Derive apply and prefix source epochs from project_mls_message’s authenticated projection rather than mutable MessageRecord.epoch metadata. Replace or supplement the helper-injection regression with a test that submits the rival through carol.ingest(...) and drains scheduled convergence; before the correction, that test would end in WrongEpoch instead of adopting the selected branch.

   Confidence: High.

2. P1 — Consumed proposals turn retained in-window history into an unrecoverable rewind

   Evidence: crates/cgka-engine/src/openmls_projection.rs, changed line 1621, where the new replayability floor excludes a Processed commit that consumed a proposal by reference. The corresponding documentation changes in docs/marmot-architecture/cgka-engine-spec.md and docs/marmot-architecture/distributed-convergence.md now permit the rewind base to stop above such a commit, despite the proposal bytes and commit history remaining durably stored.

   Trigger: A client catches up across multiple commits in one apply, an intermediate already-applied commit consumes a proposal by reference, and a later in-horizon rival requires replay from the retained anchor below that intermediate epoch.

   Consequence: lowest_replayable_epoch raises the floor above the otherwise retained prefix. The engine reports MissingRetainedAnchor/Unrecoverable and permanently halts local convergence even though the required material has not been lost. This changes the existing v1 retention/convergence behavior for ordinary proposal-consuming histories.

   Smallest sound correction: Re-admit and replay the exact consumed proposal dependencies with the already-applied prefix, or retain an intermediate snapshot before the proposal-consuming commit. Change the added proposal-boundary regression to expect successful lower-anchor convergence; that regression would have exposed the unrecoverable result before the correction. If weakening the v1 retention contract is intentional, negotiate it through a versioned required capability instead of silently changing existing behavior.

   Confidence: High.

Audit evidence:
- Exact head SHA: f2abc16ccb777102d68a0589d68cafcd65a283dd
- Changed files inspected: crates/cgka-engine/AGENTS.md; crates/cgka-engine/src/openmls_projection.rs; crates/cgka-engine/tests/AGENTS.md; crates/cgka-engine/tests/distributed_convergence.rs; crates/cgka-engine/tests/fork_detection.rs; docs/marmot-architecture/cgka-engine-spec.md; docs/marmot-architecture/distributed-convergence.md.
- This discussion follow-up preserves only the two newly supported actionable claims above. The documentation-only Core Invariant comment does not establish an additional concrete defect beyond the retained-input compatibility problem already captured in finding 2.
- No physical-device evidence exists for this head because no trusted Android artifact is bound to this repository head; physical PASS is forbidden.

Risk lenses checked:
- tests: assertions fail before the fix; boundary and negative cases; test exercises production behavior.
- Lifecycle/restart behavior was checked because the changed path explicitly covers restarted committers.
- Compatibility was checked because the proposal floor changes v1 replay and retention behavior.
- Accessibility and authorization were not routed, and this diff changes neither UI nor authorization surfaces.

Check evidence:
- Supplied evidence is available and complete, with all_completed=true and all_successful=true.
- Rust test shards, Rust check, clippy, formatting, simulator/vectors, Tamarin proofs, Rust audit, CLI/relay tests, bindings smoke, and listed build/package checks completed successfully.
- Publish wn-agent release was skipped.
- These CI results do not exercise the production-ingest lower-anchor boundary or establish physical-device correctness.

jgmontoya:
superseded by #1675

PR #1674, opened 2026-09-04
fix(simulator): redeliver capacity-refused relay history

### LINKED ISSUE #1670: offline-catchup family consumes cap-refused history without modeling redelivery
## Summary

`offline-catchup-pressure/v1` injects a retained relay's complete result set into an offline client before allowing that client to process any of it. Large profiles therefore exceed the engine's intentional 256-row `PeelDeferred` cap. The retained-relay subject records later inputs as terminal `resource_refused`, while the scenario advances to `Eventually`, whose iterations only tick the engine and never replay the refused relay events.

This makes the large natural-history arm report permanent payload loss even when the documented transport-redelivery recovery path succeeds.

Observed on current `master` at `d2e38adf05c7bf6861a1fc9031a6a9c3ffe28c8c` after #1650, #1651, and #1652 were merged.

## Reproduction

```text
generator_version=1
seed=9101
case_index=12
profile=large-384/natural-full-history
storage=file-backed SQLCipher
```

Run the first 13 cases so the stable prefix reaches case 12:

```sh
cargo run -p cgka-conformance-simulator --bin cgka-conformance-campaign --locked -- \
  --family offline-catchup-pressure/v1 \
  --seed 9101 \
  --cases 13 \
  --case-timeout-secs 300 \
  --storage file \
  --out target/offline-catchup-repro
```

Cases 0 through 10 pass. Case 12 runs to a structured failure without a timeout or harness/backend error.

## Observed boundary

After the family's one full-history sync and processing tick:

- Bob advances from epoch 1 to epoch 9 and receives 281 of 384 backlog applications.
- 103 later applications and four later commits are `resource_refused`.
- Those 107 refused inputs begin at application 281, after the deferred backlog reaches its intentional per-group cap.
- The engine drains to no pending work, but Bob remains at epoch 9 while the other clients are at epoch 13.
- All 128 `Eventually` iterations only tick; none re-fetches the refused history, so the last payload remains absent.

The engine contract explicitly makes transport redelivery the recovery path after the deferred backlog drains. A diagnostic variant that adds exactly one more full-history sync after that drain passes all 18 strict expectations: Bob reaches epoch 13, receives all 384 backlog applications plus the two post-recovery probes, and reports no pending work.

## Root cause

`RetainedRelaySubject::sync_one` injects every returned event synchronously and advances the query cursor after assembling the result. It does not interleave bounded engine progress with delivery and does not retain/replay events whose ingest returns `ResourceRefused`.

`offline_catchup_family::sync_and_tick` performs one `SyncRelayHistory` followed by one `Tick`. Its later `Eventually` assertion can only generate more ticks. Consequently, profiles larger than the admission cap cannot exercise the production-shaped refusal/backfill loop that the scenario claims to verify.

This is not a request to raise or bypass the 256-row safety cap.

## Expected behavior

The retained-relay scenario path should model backpressure/redelivery explicitly. Once capacity drains, inputs refused from a complete retained-history result must be eligible for another bounded replay, and strict completion should require all expected payloads, convergence, decryptability, and no pending work.

## Acceptance criteria

- Keep the engine's per-group deferred-row cap intact.
- Teach the retained-relay subject or scenario vocabulary to preserve and retry cap-refused retained-history inputs after progress frees capacity, or model repeated bounded full-history queries until a sound completion condition is reached.
- Do not treat `ResourceRefused` as durable transport acceptance or advance an incremental cursor past unrecoverable refused input without another recovery obligation.
- Add a focused file-backed 384-application/12-commit natural-history regression that completes without manually editing generated input.
- Require exact-once application delivery, canonical convergence, post-recovery bidirectional decryptability, and exact pending-work closure.
- Bound retry turns and retain a typed failure for a truly full/no-progress loop.
- Preserve reverse ordering, duplicate replay, restart, and exact-wrapper dedup coverage from #1650 and #1193.

## Related work

- #1651 correctly removed the synthetic harness backend error; it did not add relay redelivery to `Eventually`.
- #1650 protects exact processed-wrapper replay after the rewind horizon.
- #1193 intentionally relies on transport redelivery after local resource release.

erskingardner:
## Summary
- Preserve a retained-relay redelivery obligation when engine ingest reaches the deferred-peel capacity.
- Re-query complete relay history on later ticks, expose the obligation as pending transport work, and fail repeated unchanged full refusals with a typed resource error.
- Add the file-backed 384-application natural-history regression and document the adapter contract.

## Test plan
- [x] New test failed before the fix and passes after.
- [x] cargo test -p cgka-conformance-simulator --test offline_catchup_family
- [x] cargo test -p cgka-conformance-simulator
- [x] just fast-ci

Fixes #1670

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved offline catch-up when deferred transport capacity is temporarily unavailable.
  * Retained messages are now retried through complete-history synchronization on later processing cycles.
  * Persistent capacity refusal without progress now ends with a clear resource failure instead of repeated retries.

* **Tests**
  * Added coverage for large natural-history catch-up scenarios, including successful redelivery after capacity recovery and stalled retry handling.

* **Documentation**
  * Expanded guidance on retained-message redelivery, capacity limits, and offline catch-up scenarios.

### DIFF
--- a/crates/cgka-conformance-simulator/src/retained_relay.rs
+++ b/crates/cgka-conformance-simulator/src/retained_relay.rs
@@ -10,17 +10,19 @@ use crate::{
     EngineHarnessSubject, HarnessStorageMode, ScenarioAdminPolicyObservation,
     ScenarioMessageSelectorV2, ScenarioPredicateObservationV2, ScenarioPredicateV2,
     ScenarioRelayV2, ScenarioTopologyV2, ScenarioTransportClass, SubjectCapability,
-    SubjectCreateGroup, SubjectDescriptor, SubjectError, SubjectInviteMembers,
-    SubjectOutboundArtifact, SubjectOutboundKind, SubjectOutboundOutcome, SubjectProgressSnapshot,
-    SubjectRemoveMembers, SubjectSelfUpdate, SubjectSendApplication, SubjectUpdateAdminPolicy,
-    SubjectUpdateGroupData,
+    SubjectCreateGroup, SubjectDescriptor, SubjectError, SubjectFailureCategory,
+    SubjectInviteMembers, SubjectOutboundArtifact, SubjectOutboundKind, SubjectOutboundOutcome,
+    SubjectProgressSnapshot, SubjectRemoveMembers, SubjectSelfUpdate, SubjectSendApplication,
+    SubjectUpdateAdminPolicy, SubjectUpdateGroupData,
 };
 use async_trait::async_trait;
 use cgka_traits::group::ProtocolProfile;
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
 use std::collections::{BTreeMap, BTreeSet, HashSet};
 
 const DEFAULT_RELAY_ID: &str = "relay:retained-default";
+const MAX_REDELIVERY_NO_PROGRESS_TURNS: usize = 2;
 
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
@@ -106,6 +108,8 @@ pub struct RetainedRelaySubject {
     action_ids: BTreeMap<String, String>,
     semantic_classes: BTreeMap<String, ScenarioTransportClass>,
     sync_observations: Vec<RelaySyncObservationV2>,
+    capacity_refused_redelivery: BTreeMap<String, usize>,
+    redelivery_no_progress_turns: BTreeMap<String, usize>,
 }
 
 impl RetainedRelaySubject {
@@ -188,6 +192,8 @@ impl RetainedRelaySubject {
             action_ids: BTreeMap::new(),
             semantic_classes: BTreeMap::new(),
             sync_observations: Vec::new(),
+            capacity_refused_redelivery: BTreeMap::new(),
+            redelivery_no_progress_turns: BTreeMap::new(),
         })
     }
 
@@ -432,6 +438,35 @@ fn relay_history_sets_equal(
     sets.all(|set| set == first)
 }
 
+fn redelivery_no_progress_turns(
+    previous_turns: usize,
+    attempted_objects: Option<usize>,
+    refused_objects: Option<usize>,
+    before_token: Option<&str>,
+    after_token: Option<&str>,
+) -> Result<usize, SubjectError> {
+    let stalled = attempted_objects.is_some()
+        && attempted_objects == refused_objects
+        && before_token.is_some()
+        && before_token == after_token;
+    if !stalled {
+        return Ok(0);
+    }
+
+    let turns = previous_turns.saturating_add(1);
+    if turns >= MAX_REDELIVERY_NO_PROGRESS_TURNS {
+        return Err(SubjectError::classified(
+            SubjectFailureCategory::Resource,
+            "retained_history_redelivery_no_progress",
+            format!(
+                "retained-history redelivery refused all {} object(s) for {turns} consecutive turns without engine progress",
+                attempted_objects.unwrap_or_default()
+            ),
+        ));
+    }
+    Ok(turns)
+}
+
 #[async_trait]
 impl ConvergenceSubject for RetainedRelaySubject {
     fn descriptor(&self) -> SubjectDescriptor {
@@ -554,7 +589,50 @@ impl ConvergenceSubject for RetainedRelaySubject {
     async fn tick(&mut self, clients: &[String]) -> Result<(), SubjectError> {
         for client in clients {
             if self.online.contains(client) {
-                self.engine.tick(std::slice::from_ref(client)).await?;
+                let retrying = self.capacity_refused_redelivery.get(client).copied();
+                let before = retrying
+                    .is_some()
+                    .then(|| self.engine.structural_progress())
+                    .transpose()?
+                    .map(|snapshot| snapshot.structural_token);
+                if retrying.is_some() {
+                    self.sync_one(client, &ScenarioRelaySyncModeV2::FullHistory)?;
+                }
+
+                let refused = self
+                    .engine
+                    .tick_observing_capacity_refusals(std::slice::from_ref(client))
+                    .await?
+                    .get(client)
+                    .copied();
+                if let Some(refused) = refused {
+                    self.capacity_refused_redelivery
+                        .insert(client.clone(), refused);
+                } else {
+                    self.capacity_refused_redelivery.remove(client);
+                }
+
+                let after = retrying
+                    .is_some()
+                    .then(|| self.engine.structural_progress())
+                    .transpose()?
+                    .map(|snapshot| snapshot.structural_token);
+                let no_progress_turns = redelivery_no_progress_turns(
+                    self.redelivery_no_progress_turns
+                        .get(client)
+                        .copied()
+                        .unwrap_or_default(),
+                    retrying,
+                    refused,
+                    before.as_deref(),
+                    after.as_deref(),
+                )?;
+                if no_progress_turns == 0 {
+                    self.redelivery_no_progress_turns.remove(client);
+                } else {
+                    self.redelivery_no_progress_turns
+                        .insert(client.clone(), no_progress_turns);
+                }
             }
         }
         Ok(())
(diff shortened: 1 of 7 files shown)


crates/cgka-conformance-simulator/src/retained_relay.rs
```diff
@@ -432,6 +438,35 @@ fn relay_history_sets_equal(
...
 }
 
+fn redelivery_no_progress_turns(
+    previous_turns: usize,
+    attempted_objects: Option<usize>,
+    refused_objects: Option<usize>,
+    before_token: Option<&str>,
+    after_token: Option<&str>,
+) -> Result<usize, SubjectError> {
+    let stalled = attempted_objects.is_some()
```
erskingardner, line 448: Minor naming nit: `attempted_objects` actually carries the *previous* turn's refused count, not the number of objects attempted this turn. The stall condition `attempted_objects == refused_objects` therefore implicitly assumes the full-history re-query injects the same set with no new relay events published between retry turns. If new events land between retry turns and everything is refused, `attempted != refused` resets the counter even with zero engine progress. That can't happen while `Eventually` is only ticking, and the loop is still bounded by `max_iterations`, so it's fine in practice — but a one-line doc comment stating the assumption (or renaming to `previous_refused_objects`) would make this easier to audit.

crates/cgka-conformance-simulator/src/retained_relay.rs
```diff
@@ -554,7 +589,50 @@ impl ConvergenceSubject for RetainedRelaySubject {
...
             if self.online.contains(client) {
-                self.engine.tick(std::slice::from_ref(client)).await?;
+                let retrying = self.capacity_refused_redelivery.get(client).copied();
+                let before = retrying
+                    .is_some()
+                    .then(|| self.engine.structural_progress())
+                    .transpose()?
+                    .map(|snapshot| snapshot.structural_token);
+                if retrying.is_some() {
+                    self.sync_one(client, &ScenarioRelaySyncModeV2::FullHistory)?;
```
erskingardner, line 599: The stall detector's correctness depends on an invariant that isn't asserted anywhere: the structural token must be byte-identical across a fully-refused retry turn. `EngineHarnessSubject::structural_progress` also refreshes scenario-input ledgers and syncs outbound records, so any of those side effects perturbing a field would silently defeat the no-progress guard (termination would then fall back to the `Eventually` bound, so no hang, but the typed `retained_history_redelivery_no_progress` error would never fire). Since the before/after capture spans `sync_one` + tick, worth a short comment here stating that a no-progress turn must leave the token unchanged.

crates/cgka-conformance-simulator/tests/offline_catchup_family.rs
```diff
@@ -330,3 +330,40 @@ async fn smoke_natural_and_duplicate_history_cases_pass_file_backed_strict_oracl
         }));
     }
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn large_natural_history_retries_capacity_refused_relay_objects() {
```
erskingardner, line 335: This covers the recovery path end-to-end, but the typed no-progress failure path is only covered by the `redelivery_no_progress_turns` unit test — the actual wiring (token capture around `sync_one` + tick, obligation-map bookkeeping, error propagation out of `tick`) is never exercised with a real stalled engine. Would be good to add an e2e stalled-retry case (e.g. a `test-policy-overrides` variant that pins the deferred cap so nothing is ever admitted) so the `retained_history_redelivery_no_progress` failure is observed in a report rather than just at the helper level. Non-blocking.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T09:55:24Z` |
| Commit reviewed | `2f5b6fe7c21e8705a47a8e7a7e908777a4e99949` |

Solid fix for #1670: it models the retained-relay redelivery obligation in the subject (re-query full history on later ticks, expose the obligation as pending transport work, bounded typed failure for unchanged full refusals) without touching the engine's 256-row cap. Scope is surgical: one adapter, one clean refactor of `EngineHarnessSubject::tick`, one regression test, and four consistent doc updates. No version bumps; `sha2`/`hex` are existing workspace deps.

Verified while reviewing:
- The positional `zip` in `tick_observing_capacity_refusals` is sound: `bus.mailbox_snapshot` clones the same FIFO vector that `client.tick()` drains one-outcome-per-message in order, and `zip` safely ignores any extra outcomes appended by convergence driving.
- `ensure_tick_succeeded` only fails on `Err`, so `Ok(ResourceRefused)` ticks are tolerated and refusals are attributable.
- The `structural_progress` overlay extends exactly the fields (`runnable_work`, `transport_delayed_messages`) that the pending-work oracle treats as blockers, so strict "no pending work" closure now correctly includes the obligation, and it re-derives the token the same way the engine does.
- The new test regenerates the exact issue-#1670 repro case (seed 9101, case 12); on the natural-full-history arm there is only one explicit full-history sync, so the `>= 2` assertion cannot false-positive without redelivery.

Only minor, non-blocking notes inline. (Posting as a comment review rather than an approval because GitHub does not allow the PR author's own account to approve.)

crates/cgka-conformance-simulator/src/subject.rs
```diff
@@ -1263,6 +1263,46 @@ impl EngineHarnessSubject {
...
+            let recipient_tick = self.observed_recipient_ticks;
+            self.observed_recipient_ticks = self.observed_recipient_ticks.saturating_add(1);
+            let inbound = self.bus.mailbox_snapshot(self.client(label)?.bus_id);
+            let pending_replay = (self.replay_capture_target_tick == Some(recipient_tick))
+                .then(|| self.prepare_byte_replay(label))
+                .flatten();
+            let outcomes = self.client_mut(label)?.tick().await;
+            let refused_count = inbound
+                .iter()
+                .zip(&outcomes)
```
erskingardner, line 1282: Non-blocking: this `zip` relies on an implicit invariant — `client.tick()` returns ingest outcomes as an in-order prefix matching the pre-tick `mailbox_snapshot`, with any convergence errors appended afterward (currently true: `tick_ingest_only` outcomes come first, `drive_due_convergence` only pushes). If a future change to `client.tick()` reorders or interleaves outcomes, refusal counts here would silently misalign. A short doc comment on `tick_observing_capacity_refusals` stating that alignment invariant would protect it.

crates/cgka-conformance-simulator/src/retained_relay.rs
```diff
@@ -10,17 +10,19 @@ use crate::{
...
+    SubjectUpdateAdminPolicy, SubjectUpdateGroupData,
 };
 use async_trait::async_trait;
 use cgka_traits::group::ProtocolProfile;
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
 use std::collections::{BTreeMap, BTreeSet, HashSet};
 
 const DEFAULT_RELAY_ID: &str = "relay:retained-default";
+const MAX_REDELIVERY_NO_PROGRESS_TURNS: usize = 2;
```
erskingardner, line 25: Non-blocking: a bound of 2 tolerates exactly one unchanged full-refusal retry before the typed failure. That is fine for the current `Eventually` driver, which only ticks (so a zero-structural-progress stall can only be broken by a scenario `AdvanceTime` step that arrives between assertion phases, not between these ticks). A one-line comment recording that rationale would make the magic number less arbitrary — e.g. that stall turns are only counted when the engine structural token is unchanged, so extra patience cannot help within a tick-only loop.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T09:57:10Z` |
| Commit reviewed | `2f5b6fe7c21e8705a47a8e7a7e908777a4e99949` |

(Approval not possible — reviewer is the PR author — so this is a comment review recommending merge.)

This is a clean root-cause fix for #1670, not a symptom patch: the retained-relay subject now models the transport redelivery obligation that the engine contract already assumes, the intentional 256-row deferred cap is untouched, and the change is surgical (two source files, one focused regression, contract docs).

What I verified:

- **Acceptance criteria** — cap preserved; refused inputs preserved and retried via bounded complete-history re-query; no cursor advancement past refused input (full-history re-query ignores the incremental cursor); file-backed 384-application regression on the exact repro case (seed 9101, case 12); typed `Resource` failure on a no-progress loop; reverse/duplicate/restart coverage from #1650/#1193 untouched. No drive-by changes.
- **Zip alignment in `tick_observing_capacity_refusals`** — sound today: `client.tick()` returns ingest outcomes in mailbox order as a prefix, and both convergence error paths and `drive_due_convergence` only append. `mailbox_snapshot` preserves mailbox order. Left an inline note asking to document the invariant.
- **Stall detection** — the structural token comparison is viable because the manual convergence clock does not advance inside `tick` (only via `advance_time`), and the retained-relay `tick` correctly bypasses its own wrapped `structural_progress` (which would include the refused counts) by calling `self.engine.structural_progress()` directly. The counter resets on any admission, any engine progress, and on first refusal, so only consecutive unchanged full refusals trip the typed error. The unit test pins the bound behavior.
- **Pending-work visibility** — the wrapped `structural_progress` exposes the obligation as runnable/transport-delayed work and re-hashes the token, so strict no-pending-work closure cannot pass while redelivery is owed, including for offline clients. The old `tick` now delegates to the shared helper — no duplicated logic.
- **CI** — full matrix green, including Simulator and vectors.

Two minor inline notes (documentation only, both non-blocking). Optionally, the crate `AGENTS.md` section on subject tick contracts could gain a line about the redelivery obligation alongside the README/SCENARIOS updates, but that is not required.

PR #1675, opened 2026-09-04
Fix retained anchors across multi-commit convergence

### LINKED ISSUE #1671: cgka-engine: eight retained competing-commit waves halt offline client as Unrecoverable
## Summary

After the #1651 harness fix allows bounded work to continue across scenario ticks, the retained-history competing-wave reproducer reveals a real engine halt: an offline member processes most of eight rival-commit rounds, then enters `Unrecoverable` one epoch behind the online members and makes no progress across 128 further ticks.

Observed on current `master` at `d2e38adf05c7bf6861a1fc9031a6a9c3ffe28c8c`.

## Reproduction

```text
generator_version=1
seed=9101
case_index=11
profile=medium-96/competing-commit-waves
storage=file-backed SQLCipher
```

Run the stable prefix through case 11:

```sh
cargo run -p cgka-conformance-simulator --bin cgka-conformance-campaign --locked -- \
  --family offline-catchup-pressure/v1 \
  --seed 9101 \
  --cases 12 \
  --case-timeout-secs 300 \
  --storage file \
  --out target/offline-catchup-repro
```

The case uses eight rounds of two same-epoch rival commits, 96 interleaved applications, reverse retained-history delivery, and three copies of each retained event. The smoke control with four competing rounds (`seed=9101`, `case_index=5`) passes on the same commit.

## Observed

The current harness no longer raises `convergence drain did not settle within harness pass limit`; it grants the outer assertion all 129 samples and preserves the structured report and failure capsule.

Bob nevertheless remains stuck:

```text
online peers: epoch 9, final round-7 branch-0 profile
Bob:         epoch 8, round-6 branch-1 profile
Bob state:   Unrecoverable
active convergence passes: 1
active pass members:        78
unresolved convergence:     1
stored Created rows:        14
stored Retryable rows:      6
scenario inputs pending:    6
```

Bob's decisions advance epoch 1 to 2, then 2 to 8. The following decision has zero candidates; the group becomes fail-closed and later ticks do not alter the state. Bob has delivered 78 applications, with six still pending. An additional exact full-history replay followed by another 128 ticks produces the same snapshot, so this is not explained by the retained-relay cap/redelivery modeling issue in the natural 384-message arm.

## Root-cause boundary

Source inspection narrows the transition to the missing-retained-state fail-closed path in `converge_stored_openmls_messages_with_time`: that is the path that persists `Group.unrecoverable`, emits `GroupUnrecoverable`, and makes all later convergence calls return a zero-candidate blocked result. Bob authored no workload commits while offline, so `MissingRetainedAnchor` is the likely error rather than `MissingOwnCommitCheckpoint`; the current conformance report does not export the exact error kind, so that last distinction still needs a focused engine regression.

The important invariant failure is independent of that diagnostic gap: complete retained history for the final rival-commit round is present, yet resolving the prior six-epoch jump leaves the next in-horizon round without usable retained state and permanently wedges the client.

## Expected behavior

With complete retained relay history, an offline founding member should either converge through all eight competing rounds or produce a justified terminal classification for specific losing inputs. Applying one retained path must preserve the anchor/checkpoint material needed to evaluate the immediately following retained round.

## Acceptance criteria

- Add a focused file-backed engine/simulator regression for the eight-round competing-wave sequence.
- Pin a lower passing control and the first failing depth; the earlier investigation found a seven/eight-round boundary, while current family controls establish four passes and eight fails.
- Export or assert the precise canonicalization error that triggers `Unrecoverable` in this sequence.
- Preserve the retained anchor/checkpoint required to continue from the epoch-8 selected tip to the epoch-9 candidates, or correct the pass-generation boundary that requests unavailable state.
- Bob reaches the same epoch/profile as the online clients, receives every common-prefix application exactly once, and closes all pending work.
- Repeated ticks and exact full-history redelivery remain idempotent after convergence.
- Preserve fail-closed `Unrecoverable` behavior for genuinely missing retained state and the restart durability fixed by #971.

## Related work

- #1651 exposed this engine result by removing a synthetic harness failure; its merged PR explicitly noted that the competing arm entered the long outer fallback.
- #1171 fixed application-before-commit replay ordering; this case stalls group-state progression after a retained-path jump.
- #971 must remain authoritative when retained state is genuinely missing.

erskingardner:
## Summary
- retain every intermediate canonical epoch during a multi-commit convergence apply so the next frozen generation can rewind within the configured horizon
- add file-backed four-wave and eight-wave regression coverage for the reported seed/case
- export stable canonicalization error kinds in simulator convergence observations

## Test plan
- [x] confirmed the new intermediate-anchor assertion fails before the fix and passes after it
- [x] replayed `offline-catchup-pressure/v1`, seed `9101`, case `11` from its saved generated input (22/22 expectations passed)
- [x] `cargo test -p cgka-engine --features test-policy-overrides`
- [x] `cargo test -p cgka-conformance-simulator`
- [x] `just fast-ci`

Fixes #1671

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved offline catch-up and multi-commit convergence, preserving intermediate group states needed for reliable rewinds.
  - Added coverage for competing-wave catch-up scenarios and verified successful execution without invariant or expectation failures.
  - Enhanced convergence diagnostics with stable error classifications.

- **Documentation**
  - Documented the `offline-catchup-pressure/v1` regression and required retained-anchor behavior across convergence epochs.

### DIFF
--- a/crates/cgka-conformance-simulator/src/offline_catchup_family.rs
+++ b/crates/cgka-conformance-simulator/src/offline_catchup_family.rs
@@ -414,3 +414,35 @@ fn retained_relay_topology(clients: &[String]) -> ScenarioTopologyV2 {
 fn labels<const N: usize>(items: [&str; N]) -> Vec<String> {
     items.into_iter().map(String::from).collect()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{HarnessStorageMode, run_generated_case_report_with_storage_mode};
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn four_and_eight_competing_wave_cases_converge_with_file_backed_storage() {
+        for case_index in [5, 11] {
+            let case = generate_offline_catchup_pressure_case(9_101, case_index);
+            let report = run_generated_case_report_with_storage_mode(
+                &case,
+                None,
+                HarnessStorageMode::TempFileBackedSqlite,
+            )
+            .await
+            .unwrap_or_else(|error| panic!("{}: {error}", case.scenario.name));
+            assert!(
+                report.expectation_failures.is_empty(),
+                "{} expectation failures: {:#?}",
+                case.scenario.name,
+                report.expectation_failures
+            );
+            assert!(
+                report.invariant_failures.is_empty(),
+                "{} invariant failures: {:#?}",
+                case.scenario.name,
+                report.invariant_failures
+            );
+        }
+    }
+}
--- a/crates/cgka-engine/src/openmls_projection.rs
+++ b/crates/cgka-engine/src/openmls_projection.rs
@@ -852,8 +852,14 @@ fn replay_openmls_messages_prevalidated_output<S: StorageProvider>(
     let guard = SnapshotRollbackGuard::create_group_state(storage, group_id.clone(), snapshot)
         .map_err(|e| OpenMlsProjectionError::Snapshot(format!("{e:?}")))?;
 
-    let result =
-        process_openmls_messages_inner(storage, group_id, messages, own_commits, profile_policy);
+    let result = process_openmls_messages_inner(
+        storage,
+        group_id,
+        messages,
+        own_commits,
+        profile_policy,
+        None,
+    );
     guard
         .commit()
         .map_err(|e| OpenMlsProjectionError::Snapshot(format!("{e:?}")))?;
@@ -2040,8 +2046,15 @@ fn capture_candidate_tip_context<S: StorageProvider>(
 ) -> Result<Option<CandidateTipState>, OpenMlsProjectionError> {
     // A path that fails to replay is simply not a reachable branch state; the
     // BFS already reported that outcome for adjudication.
-    if process_openmls_messages_inner(storage, group_id, messages, own_commits, profile_policy)
-        .is_err()
+    if process_openmls_messages_inner(
+        storage,
+        group_id,
+        messages,
+        own_commits,
+        profile_policy,
+        None,
+    )
+    .is_err()
     {
         return Ok(None);
     }
@@ -2450,6 +2463,7 @@ pub(crate) fn apply_openmls_canonicalization_result_with_profile_policy<S: Stora
             group_id,
             result,
             &replay_messages,
+            max_retained_anchor_rewind,
             profile_policy,
         )?;
         if apply_start_epoch < current_epoch || restore_own_checkpoint {
@@ -2866,6 +2880,7 @@ fn apply_openmls_canonicalization_result_inner<S: StorageProvider>(
     group_id: &GroupId,
     result: &CanonicalizationResult,
     replay_messages: &[TransportMessage],
+    max_retained_anchor_rewind: u64,
     profile_policy: ReplayProfilePolicy,
 ) -> Result<Vec<OpenMlsReplayObservation>, OpenMlsProjectionError> {
     // #157/#424: the convergence-apply multi-write durable sequence
@@ -2896,6 +2911,7 @@ fn apply_openmls_canonicalization_result_inner<S: StorageProvider>(
             replay_messages,
             &PrevalidatedOwnCommits::default(),
             profile_policy,
+            Some(max_retained_anchor_rewind),
         )?;
         update_group_record_from_replay(storage, group_id, &output)?;
         persist_openmls_canonicalization_dispositions(storage, result)?;
@@ -3451,6 +3467,7 @@ fn process_openmls_messages_inner<S: StorageProvider>(
     messages: &[TransportMessage],
     own_commits: &PrevalidatedOwnCommits,
     profile_policy: ReplayProfilePolicy,
+    retain_replayed_anchors: Option<u64>,
 ) -> Result<OpenMlsReplayOutput, OpenMlsProjectionError> {
     let reject_legacy_group_additions = profile_policy.reject_legacy_group_additions
         && storage
@@ -3805,6 +3822,24 @@ fn process_openmls_messages_inner<S: StorageProvider>(
                         reason: format!("current-profile merged state: {error}"),
                         rejection_category: None,
                     })?;
+                if let Some(max_retained_anchor_rewind) = retain_replayed_anchors {
+                    // One frozen pass may adopt several consecutive commits.
+                    // Persist every resulting epoch, not only the final tip:
+                    // input held for the next generation can still fork from
+                    // any intermediate epoch inside the rewind horizon.
+                    let intermediate = OpenMlsReplayOutput {
+                        observations: Vec::new(),
+                        final_epoch: mls_group.epoch().as_u64(),
+                        final_members: marmot_members(&mls_group),
+                        epoch_authenticators: BTreeMap::new(),
+                    };
+                    update_group_record_from_replay(storage, group_id, &intermediate)?;
+                    retain_current_group_epoch_snapshot(
+                        storage,
+                        group_id,
+                        max_retained_anchor_rewind,
+                    )?;
+                }
                 prefix_canonical =
                     prefix_canonical && own_commits.is_canonical(&projection.message_digest);
             }
--- a/crates/cgka-conformance-simulator/SCENARIOS.md
+++ b/crates/cgka-conformance-simulator/SCENARIOS.md
@@ -911,6 +911,10 @@ exactly once, exact canonical equality, no pending work, and active bidirectiona
 family always selects the retained-relay subject so reconnect means a durable history query rather than healing an
 in-memory packet partition.
 
+`four_and_eight_competing_wave_cases_converge_with_file_backed_storage` pins the lower four-round control and the
+eight-round `seed=9101`, `case_index=11` regression from mdk#1671. The latter requires a selected path that spans
+several epochs to leave every intermediate retained anchor available to the next frozen convergence generation.
+
 ### General Simulator Integrity Checks
 
 These tests keep the simulator machinery honest.
--- a/crates/cgka-conformance-simulator/src/vector.rs
+++ b/crates/cgka-conformance-simulator/src/vector.rs
@@ -1089,6 +1089,9 @@ pub struct ConvergenceDecisionObservation {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub selected_app_witness_score: Option<u64>,
     pub candidate_count: usize,
+    /// Stable canonicalization error tags emitted for this evaluation.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub error_kinds: Vec<String>,
 }
 
 fn observe_convergence_decision(kind: &AuditEventKind) -> Option<ConvergenceDecisionObservation> {
@@ -1098,6 +1101,7 @@ fn observe_convergence_decision(kind: &AuditEventKind) -> Option<ConvergenceDeci
         decisive_rule,
         selected_branch_id,
         selected_tip_epoch,
+        error_kinds,
         ..
     } = kind
     else {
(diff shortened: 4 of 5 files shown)


crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -3805,6 +3822,24 @@ fn process_openmls_messages_inner<S: StorageProvider>(
                         reason: format!("current-profile merged state: {error}"),
                         rejection_category: None,
                     })?;
+                if let Some(max_retained_anchor_rewind) = retain_replayed_anchors {
+                    // One frozen pass may adopt several consecutive commits.
+                    // Persist every resulting epoch, not only the final tip:
+                    // input held for the next generation can still fork from
+                    // any intermediate epoch inside the rewind horizon.
+                    let intermediate = OpenMlsReplayOutput {
```
erskingardner, line 3830: Minor note (non-blocking): this synthetic `OpenMlsReplayOutput` works only because `update_group_record_from_replay` reads just `final_epoch`/`final_members` today — the empty `observations`/`epoch_authenticators` are load-bearing by omission. If that function ever consumes `epoch_authenticators` (e.g. to stamp them into the group record), this silent empty map could matter. A tiny named helper (`update_group_record_epoch_and_members(&mls_group)`) would remove the coupling; happy either way.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -3805,6 +3822,24 @@ fn process_openmls_messages_inner<S: StorageProvider>(
...
+                        epoch_authenticators: BTreeMap::new(),
+                    };
+                    update_group_record_from_replay(storage, group_id, &intermediate)?;
+                    retain_current_group_epoch_snapshot(
+                        storage,
+                        group_id,
+                        max_retained_anchor_rewind,
+                    )?;
+                }
                 prefix_canonical =
```
erskingardner, line 3843: Minor note (non-blocking): a single-commit apply at the live tip now captures/prunes the anchor twice (here, then again in the post-apply `retain_current_group_epoch_snapshot`), and a multi-commit apply does one state-scoped capture + two prune passes per merged commit. Given anchors are state-scoped and multi-commit depth is horizon-bounded this seems fine, but if apply latency ever shows up in profiles, this is the spot.

crates/cgka-conformance-simulator/src/offline_catchup_family.rs
```diff
@@ -414,3 +414,35 @@ fn retained_relay_topology(clients: &[String]) -> ScenarioTopologyV2 {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{HarnessStorageMode, run_generated_case_report_with_storage_mode};
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn four_and_eight_competing_wave_cases_converge_with_file_backed_storage() {
+        for case_index in [5, 11] {
```
erskingardner, line 425: Non-blocking: this deliberately pulls the medium-96 volume (96 apps, 8 commit rounds, reverse order, 3x duplicates) into the default `cargo test -p cgka-conformance-simulator` suite, while the volume table's comment says ordinary tests run the first block and larger tiers are for manual campaigns. CI absorbed it fine (the Simulator and vectors job stayed ~10 min), so I assume it's an intentional trade for a deterministic in-suite regression — just flagging in case suite runtime budget matters here.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T09:38:43Z` |
| Commit reviewed | `fcf8c256e816cfd5df6a9f38baa913e14d63a7f9` |

## Summary

Clean root-cause fix for #1671. A frozen convergence pass that adopts several consecutive commits only persisted the final tip's retained anchor, so the next generation asking to rewind to an intermediate in-horizon epoch hit the missing-retained-state fail-closed path and wedged the client as `Unrecoverable`. Retaining every intermediate epoch (group record refresh + state-scoped anchor + horizon prune, per merged commit) is the right fix, matching the issue's expected behavior.

What I verified beyond green CI:

- **Correct seam choice.** Intermediate OpenMLS states only exist inside `process_openmls_messages_inner`'s loop, so threading an `Option<u64>` (retain horizon) through the two probe/replay call sites (`None`) and the canonical apply (`Some`) is the minimal shape. The probe paths stay unpolluted.
- **Transactional safety.** The per-commit `create_group_state_snapshot`/prune calls run inside `apply_openmls_canonicalization_result_inner`'s `with_transaction`; the SQLite snapshot primitives join an outer transaction (storage tests pin exactly that: `snapshot_create_joins_outer_transaction`), so a mid-apply error/crash rolls back intermediate anchors atomically. The torn-write invariant holds.
- **Prune monotonicity.** Intermediate prunes use lower epochs than the final tip prune (`oldest = epoch - rewind`), so they delete strictly less than the pre-existing final prune — no anchor/checkpoint is lost that the old behavior kept.
- **Idempotent anchor names.** `INSERT OR REPLACE` on `cgka_group_snapshots`, so the loop's tip anchor and the post-apply final retain don't conflict.
- **Deterministic regression.** Case 5 -> `CompetingCommitWaves`/smoke-24 (4 rounds, passing control) and case 11 -> `CompetingCommitWaves`/medium-96 (8 rounds, the reproducer) via `ARM_COUNT` indexing; both pinned at seed 9101. The engine-level assertion on `openmls-retained-anchor-2` is a proper fails-before/passes-after discriminator for the retention change itself.
- **No fixture/version churn.** `error_kinds` is added to `ConvergenceDecisionObservation` as a defaulted, skip-when-empty field — backwards compatible, no `conformance_version` bump (correct per repo policy). The audit event already carried `error_kinds`; this just exports it, satisfying the issue's diagnostic-gap acceptance criterion.

Only minor notes inline; none blocking.

(GitHub does not allow the PR author to submit an approval review, so this is posted as a comment review — the recommendation above stands.)

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T10:34:59Z` |
| Commit reviewed | `fcf8c256e816cfd5df6a9f38baa913e14d63a7f9` |

## Summary

Clean, surgical root-cause fix for #1671. A multi-commit canonicalization apply previously retained only the final tip anchor (via the post-apply `retain_current_group_epoch_snapshot`), so a frozen pass adopting several consecutive commits left no rewind base for next-generation input forking from an intermediate epoch — matching the reported `MissingRetainedAnchor` → `Unrecoverable` halt one epoch behind. The fix threads an opt-in `retain_replayed_anchors` through `process_openmls_messages_inner` (only `apply_openmls_canonicalization_result_inner` opts in; speculative replays and candidate probes correctly pass `None`) and persists an anchor per merged commit.

Verified against the surrounding machinery:

- **Transaction nesting is safe**: retention now runs inside the `with_transaction` closure in `apply_openmls_canonicalization_result_inner`, but `create_group_state_snapshot` explicitly detects the current-thread transaction owner and reuses the open connection, and `list`/`release_group_snapshot` are bare statements with no nested `BEGIN`. Intermediate anchors therefore roll back atomically with a failed apply — no orphan rows.
- **Anchor naming is correct**: the intermediate `update_group_record_from_replay` runs before `retain_current_group_epoch_snapshot`, so the stored record epoch (from which the anchor name derives) matches the just-merged in-memory epoch.
- **Pruning stays bounded**: per-commit retention reuses `retain_current_group_epoch_snapshot`, which prunes below `oldest_retained_epoch` on every call, so the retained set never exceeds the configured horizon.
- **Coverage matches the issue's acceptance criteria**: the engine test pins intermediate-anchor retention on a two-commit stored path, the simulator test replays the exact `seed=9101` four-round control and eight-round regression file-backed, and the additive `error_kinds` export (serde-defaulted and skip-if-empty, so no vector `conformance_version` bump) closes the diagnostic gap the issue called out.

Two non-blocking notes in inline comments. CI is green (one build job still pending at review time). Posting as a comment since GitHub does not allow approving one's own PR — recommend merge once the pending build finishes and the PR is undrafted.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -3805,6 +3822,24 @@ fn process_openmls_messages_inner<S: StorageProvider>(
...
+                    // Persist every resulting epoch, not only the final tip:
+                    // input held for the next generation can still fork from
+                    // any intermediate epoch inside the rewind horizon.
+                    let intermediate = OpenMlsReplayOutput {
+                        observations: Vec::new(),
+                        final_epoch: mls_group.epoch().as_u64(),
+                        final_members: marmot_members(&mls_group),
+                        epoch_authenticators: BTreeMap::new(),
+                    };
+                    update_group_record_from_replay(storage, group_id, &intermediate)?;
```
erskingardner, line 3836: Non-blocking: this record update is load-bearing beyond bookkeeping — `retain_current_group_epoch_snapshot` derives the anchor name from the *stored* group record's epoch, so the record must reflect the just-merged epoch before the snapshot capture runs. A one-line comment here would protect that ordering from a future "redundant write" cleanup.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -3805,6 +3822,24 @@ fn process_openmls_messages_inner<S: StorageProvider>(
...
+                    retain_current_group_epoch_snapshot(
+                        storage,
+                        group_id,
+                        max_retained_anchor_rewind,
+                    )?;
```
erskingardner, line 3841: Non-blocking: after a successful multi-commit apply, the post-apply `retain_current_group_epoch_snapshot` in `apply_openmls_canonicalization_result_with_profile_policy` rewrites the final-tip anchor this loop already captured (same name, `INSERT OR REPLACE`). Harmless, but a brief note at that call site — or skipping the duplicate when the loop already covered the tip — would save a future reader from wondering which write matters.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T10:44:07Z` |
| Commit reviewed | `fcf8c256e816cfd5df6a9f38baa913e14d63a7f9` |
| Model | `Not exposed by runtime` |

I agree with the implementation and believe it addresses #1671 at the correct seam. A frozen canonicalization pass can merge several consecutive commits, and the intermediate OpenMLS states exist only inside this loop. Persisting a state-scoped retained anchor after each merged commit is therefore the direct fix for the missing epoch-8 rewind base that caused the next generation to fail closed as `Unrecoverable`.

The important safety properties hold: canonical apply opts into retention while speculative replay and candidate probes do not; the group record is advanced before the anchor name is derived; pruning remains bounded by the existing horizon; and the writes run inside the canonical apply transaction. The `StorageProvider` contract says all writes through the same provider join that unit, while SQLite explicitly reuses its same-thread outer transaction, so a later apply failure rolls the intermediate anchors back atomically.

Coverage is well targeted: the engine test proves an intermediate epoch-2 anchor exists after a two-commit apply, the file-backed simulator test pins both the four-wave control and exact eight-wave case 11 reproducer at seed 9101, and exporting stable `error_kinds` closes the diagnostic gap without changing existing serialized vectors when empty. Existing fail-closed behavior remains in place for genuinely missing retained state, and the full Rust suite is green at this head.

One validation caveat before merging: this head and its CI run are based on `088e9e31a236d21df349bc7e608b121293fc869d`; #1670 merged afterward as `6bf697f994d5a6829e808723e63a7444e3166832`. GitHub currently reports the PR cleanly mergeable, and the changes look compatible, but the exact combined state has not run CI. Please sync current `master` into the branch and let CI rerun, with particular attention to the file-backed case 11 regression. Assuming that stays green, I see no blocker.

GitHub does not allow approval by the PR author, so this is a comment review; the merge recommendation stands.

PR #1680, opened 2026-09-04
Fix membership reentry convergence after self-remove

erskingardner:
## Summary

- require trusted-branch removal evidence before any healthy active group can accept a replacement Welcome; a larger incoming epoch alone is never treated as branch continuity
- keep newer Welcomes retryable while the recipient is still active, including immediately after restart before epoch-manager hydration
- preserve unresolved local publication state across a refused re-Welcome and restore it during hydration
- atomically retain the joined epoch when installing a Welcome so later sibling branches can be peeled and replayed
- retain processed proposals as candidate-graph prerequisites when sibling commits reference the same SelfRemove proposal
- add the `membership-reentry/v1` family covering repeated remove/re-add cycles, re-Welcomes, restarts, self-updates, and large multi-committer timing shapes
- extend chat-journey generation so removed identities can be legally reinvited

## Why

Audit evidence from a production-shaped group showed a member leaving and later being added again around competing epoch branches. Some participants progressed while early committers remained stranded. Participant data is incomplete, so the exact identity re-add cannot be proven, but the observed partial-convergence shape is reproduced by the new large-group scenarios.

The failures had three cooperating causes: a delayed Welcome could install a stale active view after the canonical branch had removed that identity, Welcome-installed groups lacked a joined-epoch rewind anchor, and locally processed SelfRemove proposals were unavailable when replaying sibling commits. Re-entry now waits until the recipient processes removal from its currently trusted branch. This also prevents a member from replacing an active victim's state with a newer self-promoted fork, and it closes the cold-start window where an unhydrated epoch manager did not reveal a pending publication.

`Unrecoverable` repair remains the explicit exception: it can still install a fully validated replacement Welcome because the local branch is already terminally frozen.

## Validation

- `just fast-ci`
- `cargo nextest run -p cgka-engine --features test-policy-overrides,test-crash-hooks --locked --profile ci`: 523 passed, 1 skipped
- `cargo nextest run -p cgka-conformance-simulator --locked --profile ci`: 451 passed, 8 skipped
- targeted invite/rejoin test binary: 38 passed
- `membership-reentry/v1` post-fix sweep: 90 of 90 strict cases passed across nine seeds

No workspace package version or conformance fixture version is changed.

## Summary by CodeRabbit

- **New Features**
  - Added coverage for membership departure and re-entry workflows, including restart, removal, self-update, stale invitation, and multi-committer scenarios.
  - Re-invitations now support recovery after trusted removal processing while preserving converged group state.

- **Bug Fixes**
  - Improved handling of stale or replacement invitations to prevent unintended active-state replacement.
  - Strengthened recovery and convergence behavior after restarts and pending operations.
  - Development setup now retries remote commit retrieval and avoids redundant fetches when commits are already available.

- **Documentation**
  - Updated simulator scenario catalogs, campaign priorities, and coverage guidance for membership re-entry.

### DIFF
--- a/crates/cgka-conformance-simulator/SCENARIOS.md
+++ b/crates/cgka-conformance-simulator/SCENARIOS.md
@@ -633,21 +633,21 @@ regression, covers a new semantic edge, or is the smallest readable example of a
 
 ### `membership-reentry/v1`
 
-- Generator: `generate_membership_reentry_family` (generator version `3`).
+- Generator: `generate_membership_reentry_family` (generator version `1`).
 - Setup: the first eight arms use four clients; the final two use eight. A non-admin identity departs and rejoins
   through fresh KeyPackages. Seeded cases vary the victim, delivery order, and partial-impact committer count without
   changing the ten-arm catalog.
 - Pressure: the catalog guarantees one clean cycle, two cycles, three cycles with victim restarts, self-updates between
   cycles, a same-epoch self-update/removal race, self-leave followed by re-entry, self-leave racing administrative
-  removal, delivery of a withheld original Welcome only after that leaf has been removed and before a fresh
-  re-invitation, a wider self-leave for which every incumbent stages a sibling commit before any commit delivery, and
+  removal, delivery of a withheld original Welcome followed by the trusted removal commit and a fresh re-invitation,
+  a wider self-leave for which every incumbent stages a sibling commit before any commit delivery, and
   the corresponding partial-impact timing in which only two to four incumbents commit before the winning branch
   reaches the rest.
 - Expected: every successful re-entry immediately sends an application probe. The terminal oracle requires exact
   canonical equivalence, the final complete roster at the exact epoch, no pending work for survivors, exactly the
   documented retained join-commit exception for the re-added member, and active decryptability in every direction.
 - Status: all ten arms are ordinary executable regressions. The stale-original-Welcome arm at case `7` guards that
-  a fully validated, strictly newer re-invitation replaces a stale active local view and restores exact convergence.
+  a fresh re-invitation waits until the stale active view processes its trusted removal, then restores exact convergence.
   Cases `8` and `9` require convergence and re-entry after all-incumbent and partial-impact multi-auto-committer
   self-leave forks, respectively.
 
--- a/crates/cgka-conformance-simulator/SCENARIOS.md
+++ b/crates/cgka-conformance-simulator/SCENARIOS.md
@@ -631,9 +631,30 @@ regression, covers a new semantic edge, or is the smallest readable example of a
   changing production engine or protocol behavior. The original failures remain historical discovery evidence, not
   allowed results.
 
+### `membership-reentry/v1`
+
+- Generator: `generate_membership_reentry_family` (generator version `3`).
+- Setup: the first eight arms use four clients; the final two use eight. A non-admin identity departs and rejoins
+  through fresh KeyPackages. Seeded cases vary the victim, delivery order, and partial-impact committer count without
+  changing the ten-arm catalog.
+- Pressure: the catalog guarantees one clean cycle, two cycles, three cycles with victim restarts, self-updates between
+  cycles, a same-epoch self-update/removal race, self-leave followed by re-entry, self-leave racing administrative
+  removal, delivery of a withheld original Welcome only after that leaf has been removed and before a fresh
+  re-invitation, a wider self-leave for which every incumbent stages a sibling commit before any commit delivery, and
+  the corresponding partial-impact timing in which only two to four incumbents commit before the winning branch
+  reaches the rest.
+- Expected: every successful re-entry immediately sends an application probe. The terminal oracle requires exact
+  canonical equivalence, the final complete roster at the exact epoch, no pending work for survivors, exactly the
+  documented retained join-commit exception for the re-added member, and active decryptability in every direction.
+- Status: all ten arms are ordinary executable regressions. The stale-original-Welcome arm at case `7` guards that
+  a fully validated, strictly newer re-invitation replaces a stale active local view and restores exact convergence.
+  Cases `8` and `9` require convergence and re-entry after all-incumbent and partial-impact multi-auto-committer
+  self-leave forks, respectively.
+
 ### `chat-journey/v1`
 
-- Generator: `generate_stateful_chat_journey_family` (generator version `1`).
+- Generator: `generate_stateful_chat_journey_family` (generator version `2`). Version 2 makes removed identities legal
+  fresh re-invitation candidates and guarantees the remove/re-invite interaction in each eight-case rotation.
 - Setup: a four-participant legality model emits canonical IR v3 while tracking the roster, administrator set,
   connectivity, epoch, group profile, and application deliveries.
 - Profiles: even case indices generate late-invite membership journeys on the engine subject; odd indices generate
--- a/crates/cgka-conformance-simulator/SCENARIOS.md
+++ b/crates/cgka-conformance-simulator/SCENARIOS.md
@@ -646,8 +646,9 @@ regression, covers a new semantic edge, or is the smallest readable example of a
 - Expected: every successful re-entry immediately sends an application probe. The terminal oracle requires exact
   canonical equivalence, the final complete roster at the exact epoch, no pending work for survivors, exactly the
   documented retained join-commit exception for the re-added member, and active decryptability in every direction.
-- Status: all ten arms are ordinary executable regressions. The stale-original-Welcome arm at case `7` guards that
-  a fresh re-invitation waits until the stale active view processes its trusted removal, then restores exact convergence.
+- Status: all ten arms are ordinary executable regressions. The stale-original-Welcome arm at case `7` explicitly
+  observes the fresh re-invitation's non-terminal refusal while the stale view is active, then replays a byte-identical
+  Welcome after trusted removal and requires exact convergence with no pending work.
   Cases `8` and `9` require convergence and re-entry after all-incumbent and partial-impact multi-auto-committer
   self-leave forks, respectively.
 
--- a/crates/cgka-conformance-simulator/src/membership_reentry_family.rs
+++ b/crates/cgka-conformance-simulator/src/membership_reentry_family.rs
@@ -1,0 +1,668 @@
+//! Deterministic membership departure and re-entry scenarios.
+//!
+//! This family keeps most cases small so they can assert exact state, input
+//! closure, and active decryptability cheaply. Two incident-shaped arms use a
+//! wider group to make all or a seed-selected handful of unrelated incumbents
+//! race to commit the same SelfRemove before the departed identity is invited
+//! again. The catalog owns
+//! interactions that broad membership and large-group workloads only reach
+//! incidentally: repeated fresh-Welcome re-entry, self-update/removal races,
+//! self-leave, and stale pre-removal Welcomes.
+
+use crate::{
+    GeneratedScenarioCase, GeneratedSubjectKind, ScenarioAssertionV2, ScenarioMessageSelectorV2,
+    ScenarioOutboundSelection, ScenarioPredicateV2, ScenarioSpec, ScenarioStep,
+    ScenarioTransportClass, SubjectOutboundOutcome, TraceExpectation,
+};
+use rand::rngs::StdRng;
+use rand::seq::SliceRandom;
+use rand::{Rng, SeedableRng};
+
+pub const MEMBERSHIP_REENTRY_FAMILY: &str = "membership-reentry/v1";
+pub const MEMBERSHIP_REENTRY_GENERATOR_VERSION: &str = "3";
+
+const ARM_COUNT: u64 = 10;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum ReentryArm {
+    SingleCycle,
+    DoubleCycle,
+    TripleCycleWithRestarts,
+    SelfUpdateBetweenCycles,
+    SelfUpdateRemovalRace,
+    SelfLeaveThenReentry,
+    SelfLeaveRemovalRace,
+    StaleWelcomeThenFreshReentry,
+    SelfLeaveMultiCommitterReentry,
+    SelfLeavePartialMultiCommitterReentry,
+}
+
+impl ReentryArm {
+    fn from_index(case_index: u64) -> Self {
+        match case_index % ARM_COUNT {
+            0 => Self::SingleCycle,
+            1 => Self::DoubleCycle,
+            2 => Self::TripleCycleWithRestarts,
+            3 => Self::SelfUpdateBetweenCycles,
+            4 => Self::SelfUpdateRemovalRace,
+            5 => Self::SelfLeaveThenReentry,
+            6 => Self::SelfLeaveRemovalRace,
+            7 => Self::StaleWelcomeThenFreshReentry,
+            8 => Self::SelfLeaveMultiCommitterReentry,
+            _ => Self::SelfLeavePartialMultiCommitterReentry,
+        }
+    }
+
+    fn name(self) -> &'static str {
+        match self {
+            Self::SingleCycle => "single-cycle",
+            Self::DoubleCycle => "double-cycle",
+            Self::TripleCycleWithRestarts => "triple-cycle-with-restarts",
+            Self::SelfUpdateBetweenCycles => "self-update-between-cycles",
+            Self::SelfUpdateRemovalRace => "self-update-removal-race",
+            Self::SelfLeaveThenReentry => "self-leave-then-reentry",
+            Self::SelfLeaveRemovalRace => "self-leave-removal-race",
+            Self::StaleWelcomeThenFreshReentry => "stale-welcome-then-fresh-reentry",
+            Self::SelfLeaveMultiCommitterReentry => "self-leave-multi-committer-reentry",
+            Self::SelfLeavePartialMultiCommitterReentry => {
+                "self-leave-partial-multi-committer-reentry"
+            }
+        }
+    }
+}
+
+pub fn generate_membership_reentry_family(seed: u64, cases: usize) -> Vec<GeneratedScenarioCase> {
+    (0..cases)
+        .map(|case_index| generate_membership_reentry_case(seed, case_index as u64))
+        .collect()
+}
+
+pub fn generate_membership_reentry_case(seed: u64, case_index: u64) -> GeneratedScenarioCase {
+    let mut rng = StdRng::seed_from_u64(seed ^ 0x4d45_4d42_4552_5348 ^ case_index.rotate_left(19));
+    let arm = ReentryArm::from_index(case_index);
+    let wide_incident_arm = matches!(
+        arm,
+        ReentryArm::SelfLeaveMultiCommitterReentry
+            | ReentryArm::SelfLeavePartialMultiCommitterReentry
+    );
+    let clients = if wide_incident_arm {
+        labels([
+            "alice", "bob", "carol", "david", "erin", "frank", "grace", "heidi",
+        ])
+    } else {
+        labels(["alice", "bob", "carol", "david"])
+    };
+    let victim = if wide_incident_arm {
+        if rng.gen_bool(0.5) { "grace" } else { "heidi" }
+    } else if rng.gen_bool(0.5) {
+        "carol"
+    } else {
+        "david"
+    };
+    let mut delivery_order = clients.clone();
+    delivery_order.shuffle(&mut rng);
+    let mut steps = Vec::new();
+    let mut expected = Vec::new();
+
+    let mut epoch = if arm == ReentryArm::StaleWelcomeThenFreshReentry {
+        setup_without_victim(
+            &mut steps,
+            &mut expected,
+            case_index,
+            &clients,
+            victim,
+            &delivery_order,
+        );
+        1
+    } else {
+        setup_full_group(
+            &mut steps,
+            &mut expected,
+            case_index,
+            &clients,
+            &delivery_order,
+        );
+        1
+    };
+
+    match arm {
+        ReentryArm::SingleCycle => {
+            epoch = remove_and_readd(
+                &mut steps,
+                &mut expected,
+                &clients,
+                &delivery_order,
+                victim,
+                case_index,
+                0,
+                epoch,
+                false,
+            );
+        }
+        ReentryArm::DoubleCycle => {
+            for cycle in 0..2 {
+                epoch = remove_and_readd(
+                    &mut steps,
+                    &mut expected,
+                    &clients,
+                    &delivery_order,
+                    victim,
+                    case_index,
+                    cycle,
+                    epoch,
+                    false,
+                );
+            }
+        }
+        ReentryArm::TripleCycleWithRestarts => {
+            for cycle in 0..3 {
+                epoch = remove_and_readd(
+                    &mut steps,
+                    &mut expected,
+                    &clients,
+                    &delivery_order,
+                    victim,
+                    case_index,
+                    cycle,
+                    epoch,
+                    true,
+                );
+            }
+        }
+        ReentryArm::SelfUpdateBetweenCycles => {
+            for cycle in 0..2 {
+                epoch = self_update(
+                    &mut steps,
+                    &mut expected,
+                    &clients,
+                    &delivery_order,
+                    victim,
+                    cycle,
+                    epoch,
+                );
+                epoch = remove_and_readd(
+                    &mut steps,
+                    &mut expected,
+                    &clients,
+                    &delivery_order,
+                    victim,
+                    case_index,
+                    cycle,
+                    epoch,
+                    false,
+                );
+            }
+        }
+        ReentryArm::SelfUpdateRemovalRace => {
+            let self_pending = "racing-self-update";
+            steps.push(ScenarioStep::SelfUpdate {
+                client: victim.into(),
+                pending: self_pending.into(),
+            });
+            let remove_pending = "racing-remove";
+            steps.push(ScenarioStep::RemoveMembers {
+                remover: "alice".into(),
+                members: vec![victim.into()],
+                pending: remove_pending.into(),
+            });
+            confirm(&mut steps, &mut expected, victim, self_pending);
+            confirm(&mut steps, &mut expected, "alice", remove_pending);
+            settle_twice(&mut steps, &delivery_order);
+            epoch += 1;
+            assert_client_state(&mut steps, "alice", epoch, clients.len() - 1);
+            epoch = readd_and_probe(
+                &mut steps,
+                &mut expected,
+                &clients,
+                &delivery_order,
+                victim,
+                case_index,
+                0,
+                epoch,
+            );
+        }
+        ReentryArm::SelfLeaveThenReentry => {
+            steps.push(ScenarioStep::Leave {
+                client: victim.into(),
+            });
+            settle_self_leave(&mut steps, &delivery_order);
+            epoch += 1;
+            assert_client_state(&mut steps, "alice", epoch, clients.len() - 1);
+            epoch = readd_and_probe(
+                &mut steps,
+                &mut expected,
+                &clients,
+                &delivery_order,
+                victim,
+                case_index,
+                0,
+                epoch,
+            );
+        }
+        ReentryArm::SelfLeaveRemovalRace => {
+            steps.push(ScenarioStep::Leave {
+                client: victim.into(),
+            });
+            steps.push(ScenarioStep::RemoveMembers {
+                remover: "alice".into(),
+                members: vec![victim.into()],
+                pending: "remove-against-self-leave".into(),
+            });
+            confirm(
+                &mut steps,
+                &mut expected,
+                "alice",
+                "remove-against-self-leave",
+            );
+            settle_twice(&mut steps, &delivery_order);
+            epoch += 1;
+            assert_client_state(&mut steps, "alice", epoch, clients.len() - 1);
+            epoch = readd_and_probe(
+                &mut steps,
+                &mut expected,
+                &clients,
+                &delivery_order,
+                victim,
+                case_index,
+                0,
+                epoch,
+            );
+        }
+        ReentryArm::StaleWelcomeThenFreshReentry => {
+            steps.push(ScenarioStep::InviteMembers {
+                inviter: "alice".into(),
+                invitees: vec![victim.into()],
+                pending: "stale-invite".into(),
+            });
+            confirm(&mut steps, &mut expected, "alice", "stale-invite");
+            steps.push(ScenarioStep::WithholdMessage {
+                selector: ScenarioMessageSelectorV2 {
+                    publication: Some("stale-invite".into()),
+                    class: Some(ScenarioTransportClass::Welcome),
+                    ..Default::default()
+                },
+                label: "stale-original-welcome".into(),
+            });
+            settle(&mut steps, &delivery_order_without(&delivery_order, victim));
+            epoch += 1;
+            assert_client_state(&mut steps, "alice", epoch, clients.len());
+
+            steps.push(ScenarioStep::RemoveMembers {
+                remover: "alice".into(),
+                members: vec![victim.into()],
+                pending: "remove-before-welcome".into(),
+            });
+            confirm(&mut steps, &mut expected, "alice", "remove-before-welcome");
+            settle(&mut steps, &delivery_order_without(&delivery_order, victim));
+            epoch += 1;
+            assert_client_state(&mut steps, "alice", epoch, clients.len() - 1);
+
+            steps.push(ScenarioStep::ReleaseWithheld {
+                label: "stale-original-welcome".into(),
+            });
+            steps.push(ScenarioStep::Tick {
+                clients: vec![victim.into()],
+            });
+            if rng.gen_bool(0.5) {
+                steps.push(ScenarioStep::RestartClient {
+                    client: victim.into(),
+                });
+            }
+            epoch = readd_and_probe(
+                &mut steps,
+                &mut expected,
+                &clients,
+                &delivery_order,
+                victim,
+                case_index,
+                0,
+                epoch,
+            );
+        }
+        ReentryArm::SelfLeaveMultiCommitterReentry => {
+            steps.push(ScenarioStep::Leave {
+                client: victim.into(),
+            });
+            steps.push(ScenarioStep::DeliverAll);
+
+            // Every remaining member observes the standalone SelfRemove before
+            // any resulting commit is delivered. Each therefore stages and
+            // publishes a legitimate sibling commit for the same next epoch,
+            // matching the production incident's unrelated-incumbent fork.
+            let incumbents = delivery_order_without(&delivery_order, victim);
+            steps.push(ScenarioStep::Tick {
+                clients: incumbents.clone(),
+            });
+            for incumbent in &incumbents {
+                steps.push(accept_all_outbound(incumbent));
+            }
+            settle_twice(&mut steps, &delivery_order);
+            epoch += 1;
+            for incumbent in &incumbents {
+                assert_client_state(&mut steps, incumbent, epoch, clients.len() - 1);
+            }
+            epoch = readd_and_probe(
+                &mut steps,
+                &mut expected,
+                &clients,
+                &delivery_order,
+                victim,
+                case_index,
+                0,
+                epoch,
+            );
+        }
+        ReentryArm::SelfLeavePartialMultiCommitterReentry => {
+            steps.push(ScenarioStep::Leave {
+                client: victim.into(),
+            });
+            steps.push(ScenarioStep::DeliverAll);
+
+            // Only a seed-selected handful of incumbents reach their delayed
+            // auto-commit before the first sibling commit arrives. The rest
+            // ingest one of those commits before their timer fires and follow
+            // the winner directly. This preserves the production-shaped
+            // partial-impact timing in which most members progress while a
+            // few early committers must rewind and replay the shared proposal.
+            let incumbents = delivery_order_without(&delivery_order, victim);
+            let committer_count = rng.gen_range(2..=4).min(incumbents.len());
+            let early_committers = incumbents[..committer_count].to_vec();
+            steps.push(ScenarioStep::Tick {
+                clients: early_committers.clone(),
+            });
+            for incumbent in &early_committers {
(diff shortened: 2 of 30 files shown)


erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T10:49:16Z` |
| Commit reviewed | `ea95b1c1f7574e3ccb651bc88cc112919a4eb2f0` |

## Summary

Strong, well-motivated fix. The three engine changes are each minimal, guarded by transactionality or an existing skip rule, and covered by targeted tests plus a new ten-arm conformance family that reproduces the incident shape.

Engine fixes check out:

- **Strictly-newer Welcome replacement (`group_lifecycle.rs`)**: the `incoming_epoch <= group.epoch` gate widens the existing fully-authenticated repair path rather than adding a new one. `clear_live_openmls_group_on_storage` runs before OpenMLS verification, but every write stays inside the single `with_transaction`, and `rejected_newer_welcome_rolls_back_active_group_replacement` pins the rollback. The downgrade direction (`WelcomeAlreadyProcessed` for older/equal epochs) is pinned in `newer_readd_welcome_replaces_stale_active_view_without_allowing_downgrade`.
- **Joined-epoch anchor on Welcome install**: `retain_current_epoch_snapshot_on_storage` inside the join transaction is the right placement — a snapshot failure now also restores the consumed KeyPackage and staged join rows.
- **Processed proposals in the candidate graph**: the canonical-apply skip (`accepted_proposals` + authenticated `source_epoch < apply_start_epoch`) is pre-existing and matches the new comment; re-admission is what lets every sibling committer materialize the shared SelfRemove branch.

## Notes (non-blocking)

- `openmls_projection.rs:3103` — the doc for `message_state_for_dropped_reason` still references `record_state_is_canonicalization_input`, which this PR deletes. Update to `record_state_can_contribute_to_openmls_graph` / `OPENMLS_GRAPH_INPUT_STATES` so the doc doesn't drift from the code it cites.
- The new `membership_reentry_family.rs` runtime tests re-run cases 8 and 9, which `ordinary_catalog_passes_strict_runtime_oracles` already executes. Keeping them for failure isolation is defensible; just noting the duplicated engine time per suite.
- Question on the `update_group_data.rs` hydration additions: is the explicit `hydrate_all_stored_groups()` before convergence replay purely test posture (the rebuilt engines previously skipped session-open hydration), or does the processed-proposal re-admission now require a hydrated live group for convergence replay in production too? If the latter, it would be worth one line in the convergence contract docs; if the former, no action needed.
- CI engine/simulator test jobs were still `IN_PROGRESS` at review time; `check`, `clippy`, and `fmt` are green.

The conformance family itself is well-shaped: deterministic and prefix-stable by construction, the case-7 stale-Welcome arm pins exactly the freshness rule this PR introduces, and cases 8/9 reproduce the all-committer and partial-impact forks from the incident. Docs (`SCENARIOS.md`, `PROPERTY_TESTS.md`, `RUNNING_CAMPAIGNS.md`, `SCALING_CAMPAIGNS.md`) are updated consistently, and no workspace or fixture version is bumped, per repo invariant.

crates/cgka-engine/src/openmls_projection.rs
```diff
@@ -1258,8 +1258,17 @@ fn seed_stored_openmls_graph_inputs<S: StorageProvider>(
...
+            // sibling commit that references the same proposal. Excluding it
+            // makes every committer able to materialize only its own
+            // checkpointed branch, so each independently declares its local
+            // sibling canonical. Re-admit retained processed proposals to the
+            // candidate graph exactly like processed commits and application
+            // witnesses; canonical apply skips one already consumed by an
+            // accepted prefix using its authenticated source epoch.
             OpenMlsContentKind::Proposal
-                if record_state_is_canonicalization_input(record.state)
+                if record_state_can_contribute_to_openmls_graph(record.state)
```
erskingardner: Nit: this `record_state_can_contribute_to_openmls_graph(record.state)` guard is now redundant — the top-of-loop gate a few lines above already skips any record whose state is not in `OPENMLS_GRAPH_INPUT_STATES`. Dropping the arm-level condition (keeping the `source_epoch >= retained_anchor_epoch` bound) would make the Proposal arm read like a plain admission rule; alternatively keep it for symmetry with the Application arm below, which has the same redundancy pre-existing.

crates/cgka-engine/tests/update_group_data.rs
```diff
@@ -2288,6 +2288,9 @@ async fn rebuilt_engine_convergence_withdraws_own_confirmed_rename_by_stamped_or
     drop(loser_engine);
     let mut loser =
         build_with_storage_and_peeler(loser_id, loser_storage.clone(), ephemeral_peeler());
+    loser
+        .hydrate_all_stored_groups()
+        .expect("restart hydrates the durable group before convergence replay");
```
erskingardner, line 2293: Question: was this explicit `hydrate_all_stored_groups()` required because the processed-proposal re-admission changes convergence replay to depend on a hydrated live group, or just because these rebuilt-engine tests previously skipped session-open hydration and the new input set made the un-hydrated path fail? If convergence replay now genuinely requires hydration before it can replay `Processed` proposals, that contract change would be worth a line in the convergence docs; `buffer_openmls_convergence_message_at` calls `ensure_hydrated_for_convergence` internally, so I suspect this is test posture only — just want it confirmed.

crates/cgka-engine/src/group_lifecycle.rs
```diff
@@ -1002,12 +1009,23 @@ impl<S: StorageProvider> Engine<S> {
                                 .members
                                 .iter()
                                 .any(|member| &member.id == self.identity.self_id());
-                            if self_is_recorded_member && !group.unrecoverable {
+                            if self_is_recorded_member
+                                && !group.unrecoverable
+                                && incoming_epoch <= group.epoch
```
erskingardner: **Blocking:** this replacement path is reachable while the device holds an unresolved local publication on the stale view. The stale record is `Stable` and fully send-capable, so the device can stage its own commit (self-update, or an auto-commit for a SelfRemove it observed on the stale branch) and be sitting in `PendingPublish` when the strictly newer re-add Welcome is ingested.

In that interleaving the transaction commits the full replacement (live MLS rows deleted, new group installed, record rewritten at the newer epoch), but the post-transaction `epoch_manager.set_stable` silently refuses a `PendingPublish` state (warn + return), so the pending entry survives against the discarded MLS group. When the orphaned publication later resolves, `do_confirm_published` loads the new group, finds no staged commit, skips the merge, and `confirm_publish` installs `Stable(stale_epoch + 1)` — or `publish_failed` / `rollback_publish` installs `Stable(stale prior_epoch)` — while the durable record and MLS state sit at the newer joined epoch. That is the epoch-manager/record mirror split the crate's own contract treats as a bug class, and it persists until the next confirmed send or a restart re-seeds from the record.

Before this PR the replacement path was unreachable for a recorded active member, so the interleaving is new. Suggest either refusing/deferring the join while a publication is held for this group (a retryable error is safe — the Welcome is durable and re-deliverable) or treating the held publication as failed as part of the replacement (rollback the epoch-manager entry in memory; the staged commit belongs to the discarded view). A regression test mirroring `newer_readd_welcome_replaces_stale_active_view_without_allowing_downgrade` with a staged self-update in flight at replacement time would pin whichever semantics you choose.

crates/cgka-conformance-simulator/src/membership_reentry_family.rs
```diff
@@ -0,0 +1,668 @@
...
+    GeneratedScenarioCase, GeneratedSubjectKind, ScenarioAssertionV2, ScenarioMessageSelectorV2,
+    ScenarioOutboundSelection, ScenarioPredicateV2, ScenarioSpec, ScenarioStep,
+    ScenarioTransportClass, SubjectOutboundOutcome, TraceExpectation,
+};
+use rand::rngs::StdRng;
+use rand::seq::SliceRandom;
+use rand::{Rng, SeedableRng};
+
+pub const MEMBERSHIP_REENTRY_FAMILY: &str = "membership-reentry/v1";
+pub const MEMBERSHIP_REENTRY_GENERATOR_VERSION: &str = "3";
```
erskingardner: Non-blocking question: a brand-new family debuts at generator version `"3"` while the sibling families (`large-group-pressure`, `offline-catchup-pressure`, `chat-journey`) all started at `"1"`. If this is development-history residue, consider starting at `"1"` now — it is free before merge, and saved campaign inputs will pin whatever ships. If it deliberately shares a versioning epoch with something else, a one-line comment would save the next reader a search.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T10:57:35Z` |
| Commit reviewed | `ea95b1c1f7574e3ccb651bc88cc112919a4eb2f0` |

The engine delta here is admirably small for the failure it fixes: the strictly-newer-Welcome replacement rule, the atomic joined-epoch anchor, and the processed-proposal re-admission are each a few lines, each pinned by strong tests (stale-view replacement plus downgrade rejection, rollback of a rejected replacement, the retained-anchor assertion, and the ten-arm conformance family). The processed-proposal re-admission is coherent with the pre-existing skip in `replay_messages_for_canonicalization_result` (accepted proposals below the apply-start epoch stay out of the replay but keep their `Processed` disposition), and the `chat-journey` legality-model change is internally consistent — `online` now models connectivity only, and every member-gated action still intersects with `members`. Docs, AGENTS files, and usage strings are all updated in step.

One serious concern, inline on `group_lifecycle.rs` (posted as a comment review because GitHub does not allow requesting changes on one's own PR): the new stale-active-view replacement path is reachable while the device holds an unresolved local publication on the stale view, and the post-transaction `set_stable` refusal then leaves the epoch manager diverged from the durable record — the mirror-split bug class this crate explicitly tracks. It needs either a join-time gate or in-memory resolution of the orphaned pending entry.

Non-blocking: a new conformance family debuting at generator version `3` (inline), and a note that most of CI was still pending at review time — the validation listed in the PR body is local. Please let the full matrix finish before merge.

crates/cgka-engine/src/group_lifecycle.rs
erskingardner: Fixed in 12a9dfc4. A replacement Welcome now returns a non-terminal InvalidTransition while that group is PendingPublish or Merging, before any live state is cleared. The identical Welcome therefore remains retryable after the publication resolves. The regression stages a self-update on the stale active view, verifies the refused replacement changes neither the durable record nor epoch-manager state, rolls back the publication, and then successfully retries the exact Welcome.

crates/cgka-engine/tests/update_group_data.rs
erskingardner: Confirmed: this is test posture only, not a new production contract. buffer_openmls_convergence_message_at already calls ensure_hydrated_for_convergence. These tests rebuild Engine directly on stored state without running the session-open hydration pipeline, so the explicit hydrate_all_stored_groups calls make the test restart match production startup before replay. Processed-proposal admission does not add a separate hydration requirement.

crates/cgka-engine/src/openmls_projection.rs
erskingardner: Addressed in e05fb294. The proposal arm now relies on the existing top-of-loop OPENMLS_GRAPH_INPUT_STATES gate and keeps only the retained-anchor epoch bound. I also updated the nearby stale doc reference to record_state_can_contribute_to_openmls_graph.

crates/cgka-conformance-simulator/src/membership_reentry_family.rs
erskingardner: Addressed in e05fb294. The new membership-reentry family now starts at generator version 1. Its scenario spec_version remains 3 because that is the canonical IR schema version rather than family history.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T11:12:48Z` |
| Commit reviewed | `e05fb2942dfa3c516da8b78f9bc1c59ffbcdb831` |

## Re-review of `ea95b1c1..e05fb294`

Reviewed the two new commits (`12a9dfc4`, `e05fb294`).

**Pending-publish guard (`group_lifecycle.rs`)** — real gap, correctly fixed, and something my first review missed. Without it, a replacement Welcome could commit its durable group copy while `set_stable` refuses to overwrite `PendingPublish`/`Merging`, splitting the epoch manager from the installed group. The guard:

- fires only on the stale-local-state (replacement/repair) path, before `clear_live_openmls_group_on_storage`, so the refused attempt rolls back KeyPackage consumption inside the same transaction — the *identical* Welcome can retry, which the test proves end-to-end;
- uses `is_resolving_local_publish` (PendingPublish + Merging, per `engine_state.rs`), the same source of truth `set_stable` consults, so the guard and the write path cannot disagree;
- returns `InvalidTransition`, which is correctly *not* in `terminal_welcome_error`, so transport ingest treats it as retryable rather than durably deduplicating the Welcome away.

The new test block in `newer_readd_welcome_replaces_stale_active_view_without_allowing_downgrade` pins the whole sequence: held SelfUpdate → refusal with `from: PendingPublish, to: JoinWelcome` → durable record and held epoch unchanged → `publish_failed` restores Stable → identical Welcome installs epoch 4. Good.

**Feedback fixes** — the redundant `record_state_can_contribute_to_openmls_graph` guard in the Proposal arm is dropped, and the `message_state_for_dropped_reason` doc now cites the live function instead of the deleted one. Both prior notes resolved.

**One doc/code mismatch to fix (trivial):** `MEMBERSHIP_REENTRY_GENERATOR_VERSION` changed `"3"` → `"1"` (correct for a brand-new, never-released generator), but `SCENARIOS.md:636` still documents `membership-reentry/v1` as "generator version `3`". Since saved campaign inputs pin `(family_name, generator_version, seed, case_index)`, the doc should match the constant — one-word fix. Everything else referencing the version (tests, campaign bin) uses the constant, so nothing else drifts.

CI note: `check`/`clippy`/`fmt` were green on the previous head; on this head they were still `IN_PROGRESS` at review time — worth letting the matrix finish before merging.

crates/cgka-conformance-simulator/src/membership_reentry_family.rs
```diff
@@ -0,0 +1,668 @@
...
+    GeneratedScenarioCase, GeneratedSubjectKind, ScenarioAssertionV2, ScenarioMessageSelectorV2,
+    ScenarioOutboundSelection, ScenarioPredicateV2, ScenarioSpec, ScenarioStep,
+    ScenarioTransportClass, SubjectOutboundOutcome, TraceExpectation,
+};
+use rand::rngs::StdRng;
+use rand::seq::SliceRandom;
+use rand::{Rng, SeedableRng};
+
+pub const MEMBERSHIP_REENTRY_FAMILY: &str = "membership-reentry/v1";
+pub const MEMBERSHIP_REENTRY_GENERATOR_VERSION: &str = "1";
```
erskingardner, line 22: Renumbering to `1` is right for a generator that ships for the first time in this PR — but `SCENARIOS.md` still documents this family as "generator version `3`" (the entry added earlier in this PR). Since saved campaign inputs pin `(family_name, generator_version, seed, case_index)`, please update the SCENARIOS.md line to `1` so the doc matches the constant.

crates/cgka-engine/src/openmls_projection.rs
erskingardner: Addressed in e05fb294 — the redundant predicate is dropped from the Proposal arm. Thanks.

crates/cgka-conformance-simulator/SCENARIOS.md
```diff
@@ -631,9 +631,30 @@ regression, covers a new semantic edge, or is the smallest readable example of a
   changing production engine or protocol behavior. The original failures remain historical discovery evidence, not
   allowed results.
 
+### `membership-reentry/v1`
+
+- Generator: `generate_membership_reentry_family` (generator version `3`).
```
erskingardner: This still says generator version `3`, but `MEMBERSHIP_REENTRY_GENERATOR_VERSION` was reset to `"1"` in `e05fb294`. Saved campaign inputs pin `(family_name, generator_version, seed, case_index)`, so the doc should match the constant — s/`3`/`1`/.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T11:15:20Z` |
| Commit reviewed | `e05fb2942dfa3c516da8b78f9bc1c59ffbcdb831` |

Re-review of `ea95b1c1..e05fb294` (commits `12a9dfc4`, `e05fb294`), following up on the blocking concern from my earlier review.

**Blocking issue — resolved.** The pending-publish gate is exactly the right shape, and I verified each property independently:

- It sits inside the join transaction before `clear_live_openmls_group_on_storage`, so a refused attempt rolls back the KeyPackage consumption with everything else — the identical Welcome stays retryable, which the extended test proves end-to-end (refusal → `publish_failed` → retry lands epoch 4).
- It consults `epoch_manager.state()` with `is_resolving_local_publish()` (PendingPublish + Merging per `engine_state.rs:288`), the same predicate `set_stable` refuses on, so the gate and the write path cannot disagree.
- `EngineError::InvalidTransition` is deliberately absent from `terminal_welcome_error` (`group_lifecycle.rs:121`), so transport ingest will not durably dedup the refused Welcome away.
- The regression test pins the error shape, the intact durable record, the intact epoch-manager state, and the retry — the full interleaving I was worried about.

**Generator version** — the const reset to `"1"` is right for an unreleased family, but `SCENARIOS.md` wasn't updated with it (inline comment). One-word fix.

**Redundancy cleanup** — dropping the state check from the proposal arm is correct: the top-of-loop filter at `openmls_projection.rs:1210` already guarantees it, and the stale `record_state_is_canonicalization_input` doc reference was updated to the live name.

CI on this head was still largely pending at review time; the prior head's check/clippy/fmt were green and the delta is small, but let the matrix finish before merging.

Posted as a comment because GitHub does not allow approving one's own PR.

dannym-arx:
- **Do not replace state while a local publication is unresolved.** The new epoch check allows a strictly newer Welcome whenever the stored record looks stale ([group_lifecycle.rs:1012](https://github.com/marmot-protocol/mdk/blob/ea95b1c1f7574e3ccb651bc88cc112919a4eb2f0/crates/cgka-engine/src/group_lifecycle.rs#L1012)). A stale client can still have a self-update or other commit in `PendingPublish`. The transaction then installs the Welcome, but the later `set_stable` call cannot complete because `EpochManager` deliberately refuses to overwrite unresolved publications ([group_lifecycle.rs:1283](https://github.com/marmot-protocol/mdk/blob/ea95b1c1f7574e3ccb651bc88cc112919a4eb2f0/crates/cgka-engine/src/group_lifecycle.rs#L1283), [epoch_manager.rs:137](https://github.com/marmot-protocol/mdk/blob/ea95b1c1f7574e3ccb651bc88cc112919a4eb2f0/crates/cgka-engine/src/epoch_manager.rs#L137)). `join_welcome` nevertheless returns success, leaving durable MLS state at the Welcome epoch while the in-memory manager still represents the old pending commit. Confirming that old handle can subsequently mark an unmerged commit processed and move the manager to its old projected epoch. I reproduced this by staging Carol’s self-update in the PR’s stale-Welcome regression before delivering the epoch-4 re-entry Welcome: the join succeeded, but `carol.epoch()` remained 3 rather than 4. The Welcome should remain retryable until the pending publication has reached its existing confirm/rollback exit, or replacement must explicitly and safely reconcile that lifecycle.

- **A newer epoch is not a continuity proof for replacing an established group.** The same condition promotes Welcome-bootstrap trust from initial admission to destructive replacement of an already trusted group. The repository already documents that a non-admin can author a fork which self-promotes in the embedded admin component and therefore passes the join-time `require_admin` check, because that check reads the author-controlled joined state ([invite_leave.rs:2366](https://github.com/marmot-protocol/mdk/blob/ea95b1c1f7574e3ccb651bc88cc112919a4eb2f0/crates/cgka-engine/tests/invite_leave.rs#L2366)). Such a member can advance its private fork, remove and re-add the victim using a public fresh KeyPackage, and deliver a Welcome whose epoch exceeds the victim’s local epoch. Lines 1012–1043 then clear the victim’s legitimate live state and install the attacker’s branch without authenticating continuity from a retained local anchor. The existing bootstrap limitation may be tolerable for a first join with application-layer corroboration, but it should not authorize overwriting an established group. Active-state replacement needs proof tied to known group lineage/authority, not just the epoch and admin set contained in the replacement Welcome.

crates/cgka-engine/src/group_lifecycle.rs
```diff
@@ -1002,12 +1009,23 @@ impl<S: StorageProvider> Engine<S> {
...
                                 // A distinct transport/content id does not make a
-                                // second normal Welcome for an already-active group
-                                // a rejoin. Reject before OpenMLS staging and let
-                                // the surrounding transaction restore KeyPackage
+                                // same- or older-epoch Welcome for an already-active
+                                // group a rejoin. Reject before OpenMLS staging and
+                                // let the surrounding transaction restore KeyPackage
                                 // consumption and every tentative write.
+                                //
+                                // A strictly newer Welcome is different: local state
```
Datawav: A larger epoch is not proof that the active local view is stale: the epoch and admin set both come from the incoming fork. The existing fork helper proves a non-admin can reuse the group ID and self-promote in that fork. After advancing its epoch and adding an active victim's fresh KeyPackage, this branch passes the later checks because the sender is an admin in its own forged branch, then replaces the victim's trusted state. Please require continuity/corroboration rooted in the current branch, and add an active-recipient regression for the self-promoted-fork case.

Datawav (changes requested):
Requesting changes: the strictly newer-Welcome gate lets a self-authorized fork replace an already-active trusted group. This is a P1 security blocker. Reviewed with OpenAI Codex GPT-5.6 Sol (two clean-context passes).

A member who is not an admin on the trusted branch can fork the same group ID, self-promote in the fork, advance its epoch, add the victim's fresh KeyPackage, and send a valid Welcome. The new `incoming_epoch > group.epoch` path accepts it, while the later admin check validates against the incoming fork's own admin set. Existing tests already demonstrate the fork/self-promotion primitive for a new join; the new active-replacement path extends it to established groups.

Require continuity or corroboration rooted in the current trusted branch before destructively replacing active state. A regression should attempt this existing self-promoted-fork construction against an already-active recipient and verify that the local group is unchanged.

mubarakcoded (commented):
## Review of `e05fb294`

**Recommended action: resolve before merge.**

### [HIGH] Cold-start replacement bypasses the pending-publication guard

**File:** `crates/cgka-engine/src/group_lifecycle.rs:1042-1063`

The new guard checks only `self.epoch_manager.state(&group_id)`. A rebuilt `Engine` starts with an empty epoch manager, and `join_welcome()` does not hydrate the existing group. If a staged local publication survived the restart in OpenMLS and fanout storage, a strictly newer Welcome received before hydration sees `None`, skips this guard, clears the old OpenMLS group (including its pending commit), and installs the replacement. The old durable publication state is then orphaned: its `PendingStateRef` was never restored and can no longer be confirmed or rolled back consistently.

The added regression covers the warm process only; it hydrates Carol before staging the self-update and never rebuilds between staging and replacement.

Please either consult durable pending state before `clear_live_openmls_group_on_storage` (for example, the stored group's `pending_commit()` plus the relevant durable lifecycle rows) or hydrate/restore the existing group before deciding replacement eligibility. Add a regression that stages and persists a self-update, rebuilds on the same storage, calls `join_welcome()` without hydration, verifies a retryable refusal and unchanged durable state, then hydrates/resolves the publication and retries the same Welcome successfully.

### Existing outstanding item

The generator constant is `"1"` while `SCENARIOS.md` still documents `"3"`; the existing unresolved inline thread already covers it, so I have not duplicated that comment.

crates/cgka-engine/src/group_lifecycle.rs
erskingardner: Fixed in ca584337. Healthy active state now rejects every strictly newer Welcome with a retryable InvalidTransition before any live state is cleared or incoming branch data is staged; only processing a removal from the currently trusted branch makes the normal re-entry path eligible. The new active_group_rejects_newer_welcome_from_self_promoted_fork regression uses the existing self-promoted-fork helper against an already-active Carol and verifies her durable record, epoch, and sendable group remain unchanged. The sibling bootstrap test still proves why the incoming branch admin check alone is insufficient. Engine nextest is 523/523 and simulator nextest is 451/451.

crates/cgka-conformance-simulator/src/membership_reentry_family.rs
erskingardner: Fixed in ca584337. SCENARIOS.md now documents generator version 1, matching MEMBERSHIP_REENTRY_GENERATOR_VERSION and the saved-input provenance.

crates/cgka-conformance-simulator/SCENARIOS.md
erskingardner: Fixed in ca584337: the documented generator version is now 1, consistent with the source constant.

erskingardner:
Addressed the two blocking findings in ca584337.

- Active-state continuity: a healthy durable group record now refuses every strictly newer Welcome with a non-terminal `InvalidTransition`. Epoch freshness and the incoming fork admin set are no longer sufficient to clear trusted state. The existing self-promoted-fork helper is now exercised against an active recipient; the regression proves the record, MLS epoch, and usable group remain unchanged.
- Cold start: the same durable active-membership gate applies before epoch-manager hydration. The restart regression stages a self-update, rebuilds on the same storage without hydration, refuses the exact Welcome, and proves the pending OpenMLS commit and group record are unchanged. Hydration then restores the pending publication, which is explicitly rolled back.
- Legitimate re-entry: after the recipient processes its removal commit from the trusted branch, the identical previously refused Welcome succeeds. This is intentionally stronger than allowing it immediately after the pending publication resolves: resolving a local publication removes that race, but it does not establish branch continuity.
- Scenario/docs: case 7 now releases the trusted removal after the delayed stale Welcome and before the fresh re-invitation. The documented generator version now matches the source constant (`1`).

Validation on this head: `just fast-ci`; engine nextest 523/523 (1 skipped); simulator nextest 451/451 (8 skipped); membership-reentry strict sweep 90/90 across seeds 0–8.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T13:06:56Z` |
| Commit reviewed | `ca58433786f55c0d0dd2534c7cd8f1e033736723` |

Re-review of `e05fb294..ca584337` — "require trusted removal before reentry". This replaces the epoch-freshness replacement rule from the earlier heads with a strictly safer contract, and I think it is the right call.

**Why the pivot is correct:** the earlier rule let a strictly newer authenticated Welcome replace recorded active state on the strength of its epoch alone. But epoch freshness is not branch continuity — a removed-or-forking member can advance its own copy, self-promote in the admin component, and issue a fully authenticated newer Welcome that passes every post-staging check (`join_accepts_welcome_from_fork_with_self_promoted_admin` already proves that bootstrap path is legal). Under the old rule that Welcome would have silently destroyed the victim's healthy trusted branch. Requiring the durable record to no longer list the identity — i.e. demanding authenticated removal evidence from the branch the victim already trusts — closes that, and consulting the durable record rather than the epoch manager also closes the cold-start window before hydration. The unrecoverable-repair exception is preserved exactly.

**Correctness checks I ran against this head:**

- The retry story is real: `InvalidTransition` stays non-terminal in `terminal_welcome_error`, and the rewritten test drives the full sequence — cold-start refusal with the staged commit intact, refusal at Stable after resolving the publication, then the *same Welcome bytes* accepted after the trusted removal commit lands. Both ingest seams mark the record `removed` and strip self from `members` when processing a removal, so the gate opens deterministically.
- The fork-attack test now constructs a genuinely authenticated attack (`welcome_from_fork_with_self_promoted_admin` with Bob self-promoting) and asserts refusal at the pre-staging gate plus record integrity.
- Conformance case 7 was updated to the new contract: the removal commit is withheld before the stale Welcome is released, then released and settled before the fresh re-invite, so the scenario exercises the exact wait-then-retry shape. The doc updates (SCENARIOS, PROPERTY_TESTS, RUNNING_CAMPAIGNS, including the generator-version `1` fix from the last round) all match.

**Residual, non-blocking:** the only pre-existing seams where `removed`/`members` and live MLS state could disagree are the crash windows this codebase already tracks as crash-hook test targets, and the repair path for that is the unrecoverable-Welcome exception, which is intact.

**CI:** all Rust jobs (fmt, check, clippy, tests 1–4, Simulator and vectors, audit) pass on this head. The single failure, `Hermes installer auth`, is a GitHub HTTP 429 rate-limit cloning the Hermes repo — infra flake, not a code failure; worth a re-run so `Required CI` goes green. Posted as a comment because GitHub does not allow approving one's own PR.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T13:08:24Z` |
| Commit reviewed | `ca58433786f55c0d0dd2534c7cd8f1e033736723` |

## Re-review of `e05fb294..ca584337`

This commit replaces the earlier "strictly-newer Welcome replaces a stale active view" rule with a strictly safer invariant, and I think it's the right call — the rule I approved previously let epoch freshness stand in for branch continuity. Specifics:

**Trusted-removal gate (`group_lifecycle.rs`)** — a healthy active member now refuses *any* replacement Welcome (`InvalidTransition`, non-terminal, so the identical Welcome retries), regardless of epoch. Re-entry proceeds only after the recipient processes removal from its currently trusted branch (realized removal drops self from the durable record), then the pre-existing rejoin path installs the Welcome. This properly closes the fork attack my earlier reviews only touched in passing: a member can fork the group, advance it, self-promote admin, and produce a cryptographically valid newer Welcome that passes the *incoming* branch's own admin check — the rewritten `active_group_rejects_newer_welcome_from_self_promoted_fork` pins exactly that shape being refused. Consulting the durable record instead of the epoch manager also closes the cold-start window, and the test now pins the unhydrated-rebuild path end-to-end (refusal → pending commit survives → hydration restores the publication → `publish_failed` → Stable refusal → trusted removal ingest → identical Welcome installs at epoch 4).

**Consistency across the change** — good: PR body rewritten accurately, `SCENARIOS.md`/`PROPERTY_TESTS.md`/`RUNNING_CAMPAIGNS.md` updated to the trusted-removal narrative, conformance case 7 now withholds the removal commit and releases it to the victim before the re-add, and the generator-version doc mismatch I flagged is fixed (`SCENARIOS.md` now says `1`). Branch movement for a member stranded on a losing branch still works via authenticated commit convergence (the processed-proposal + joined-epoch anchor fixes), which is the right separation: Welcomes gate on local membership evidence, commits converge on branch selection.

## Notes (non-blocking)

- The pending-publish guard from `12a9dfc4` is now unreachable for active members (the `ActiveMember` refusal returns first) and only fires for removed/unrecoverable members with unresolved publications. Its dedicated `PendingPublish → JoinWelcome` test shape was replaced, so no test currently pins that error. Either add a small case for the removed-member-with-pending-publish shape or leave it as documented defense-in-depth — but right now it's untested hardening.
- `from: "ActiveMember"` is a synthetic label — no `EpochState` variant is named that, and every other `InvalidTransition.from` is a real state name. Cosmetic; the `reason` string carries the meaning, but a real state name (or a distinct error variant) would keep the error surface honest.
- The refusal now happens before OpenMLS staging, so the old "failed replacement rolls back a full staging attempt" coverage no longer applies to this path (nothing is attempted). Rollback for the paths that still stage — removed-member rejoin, `Unrecoverable` repair — remains covered by the other tests; just noting the renamed test is a different mechanism, not the same coverage.
- My earlier open question stands: the `update_group_data.rs` `hydrate_all_stored_groups()` additions — test posture or a production contract that convergence replay needs a hydrated live group? A one-line answer on that thread would close it out.
- CI on this head: `check`/`clippy`/`fmt`/tests 1–4/`Simulator and vectors` all green. `Hermes installer auth` failed at its "Prepare isolated Hermes checkout" step — infra/checkout, unrelated to this PR (no Hermes files touched); a re-run was in progress, and `Required CI` only fails as its aggregate.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T13:48:20Z` |
| Commit reviewed | `847b8f32eea6d716536fda988c0e2dae15f11750` |

Re-review of `ca584337..847b8f32` — "fix(ci): avoid redundant Hermes ref fetch". This directly addresses the only failing gate from the last round (`Hermes installer auth`, which died on a GitHub HTTP 429 when re-fetching an already-cloned pinned ref).

**The fix is correct and minimal:** when `--hermes-ref` is a full 40-hex SHA already present in the clone (`cat-file -e`), checkout locally instead of a second authenticated fetch. Non-SHA refs and SHAs not reachable from the default-branch clone still take the fetch path, which now has the same 5-attempt exponential backoff the clone already had — previously the fetch had none. The auth path is unchanged (`git_with_github_auth` still wraps the fetch).

**Verification:** the new shell test pins the local-checkout case (exactly one authenticated git invocation for a pinned SHA) and the existing fake-git test keeps the two-invocation clone+fetch case with the token-leak assertion intact. The `[[ ... =~ ]]` regex matches the script's bash shebang. All CI gates pass on this head — Rust fmt/check/clippy/tests 1–4, Simulator and vectors, both Hermes jobs, wn-agent builds — with only the release-publish job skipping (expected on a PR).

Scope check: the engine/conformance content reviewed in the last three rounds is untouched by this commit. Posted as a comment because GitHub does not allow approving one's own PR.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T13:49:06Z` |
| Commit reviewed | `847b8f32eea6d716536fda988c0e2dae15f11750` |

## Re-review of `ca584337..847b8f32`

The engine/conformance content from the previous rounds is untouched; this commit is CI-only, and I agree it's correct and minimal:

- Local-checkout fast path (`cat-file -e` on a full 40-hex ref → direct `checkout --detach`) removes the redundant authenticated fetch that was being 429'd right after a successful clone; non-SHA refs and unreachable SHAs still take the fetch path, which now gains the 5-attempt exponential backoff the clone already had.
- The new shell test pins the one-authenticated-git-op case, and the earlier clone+fetch/token-leak coverage stays intact.
- Verified on this head: every gate is green, including `Hermes installer auth` and `Required CI` (only the release-publish job skips, as expected on a PR).

## Open items worth handling before or right after merge

- **Case 7 doesn't exercise what its docs claim** (CodeRabbit's open finding, which I verified and consider the substantive one): `PROPERTY_TESTS.md`, `RUNNING_CAMPAIGNS.md`, and `SCENARIOS.md` all state that case 7 guards a fresh re-invitation *remaining retryable while the victim still holds the stale active view*, then succeeding after the trusted removal. But `readd_and_probe` runs only after `ReleaseWithheld trusted-removal-after-stale-welcome` + settle, so by the time the fresh Welcome exists the victim's record no longer lists it — the scenario never triggers the `ActiveMember → JoinWelcome` refusal, and the retryability claim is pinned only by the engine test in `invite_leave.rs`. Either reorder case 7 (generate/deliver the fresh Welcome before releasing the removal, assert it's refused non-terminally, then release and let the identical Welcome succeed) or soften the three doc claims to say the refusal is engine-layer-only. The repo treats these scenario docs as coverage contracts, so the current wording overstates what the catalog pins.
- **`test_dev_scripts.sh` total git-op count** (CodeRabbit, minor): `grep -Fxc 'count=1'` counts only authenticated records, so an extra unauthenticated git operation would pass. Adding `grep -c '^count=' == 1` closes that.
- **Still unanswered from round 1**: are the `update_group_data.rs` `hydrate_all_stored_groups()` additions test posture, or does convergence replay now require a hydrated live group in production? A one-line answer on that thread would close the last open question.
- Still open from round 3 (non-blocking): the `PendingPublish → JoinWelcome` refusal shape is now untested hardening, and `from: "ActiveMember"` is a synthetic state name.

None of these block merge — the engine invariant itself is correct and fully pinned at the engine layer, and CI is green — but the case-7/doc alignment is the one I'd actually fix in this PR rather than as a follow-up.

crates/cgka-conformance-simulator/src/membership_reentry_family.rs
```diff
@@ -0,0 +1,684 @@
...
+                steps.push(ScenarioStep::RestartClient {
+                    client: victim.into(),
+                });
+            }
+            // A newer Welcome cannot authenticate its own lineage. Release the
+            // commit from the victim's currently trusted branch first; only
+            // after that removal is applied is the fresh Welcome a re-entry.
+            steps.push(ScenarioStep::ReleaseWithheld {
+                label: "trusted-removal-after-stale-welcome".into(),
+            });
```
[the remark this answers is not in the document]
erskingardner: Fixed in 9e228d0a. Case 7 now creates and confirms the fresh re-invitation while the victim still holds the stale active view, duplicates the exact queued Welcome, and withholds one copy. The first copy is delivered through the new v3 expect_tick_error action and must produce invalid_transition; that expected refusal is recorded in the trace. Only then is the trusted removal released and settled, after which the byte-identical held Welcome succeeds. The arm now requires complete pending-work closure plus exact convergence and decryptability. Full simulator tests passed, and a strict seeds 0–8 sweep passed all 72 cases.

integrations/hermes/marmot/tests/test_dev_scripts.sh
```diff
@@ -61,6 +61,23 @@ if grep -F 'test-actions-token' "$auth_capture" >/dev/null; then
...
+    PATH="$auth_fake_bin:$PATH" \
+    GITHUB_TOKEN="test-actions-token" \
+    AUTH_CAPTURE="$local_ref_capture" \
+    "$repo_root/scripts/hermes_marmot_dev_setup.sh" \
+        --root "$local_ref_root" \
+        --hermes-ref "$(printf 'ab%.0s' {1..20})" \
+        --no-enable-plugin
+[ "$(grep -Fxc 'count=1' "$local_ref_capture")" -eq 1 ]
+[ "$(grep -Fxc 'key=http.https://github.com/.extraheader' "$local_ref_capture")" -eq 1 ]
+[ "$(grep -Fxc "value=AUTHORIZATION: basic $expected_basic" "$local_ref_capture")" -eq 1 ]
```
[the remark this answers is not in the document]
erskingardner: Fixed in 9e228d0a. The local-ref regression now first asserts that the total number of recorded count= entries is exactly one, then keeps the existing authenticated-operation and header assertions. This prevents an extra unauthenticated clone or fetch from slipping through. The Hermes dev-script suite passes.

Datawav (approved):

Issue #1679, opened 2026-09-04
Expose inbox-relay-required account readiness and resumable onboarding through bindings


erskingardner:
## Summary

MDK must not report a signing account as network-ready when that account has no valid Marmot inbox relay list (Nostr kind `10050`). Imported `nsec` accounts and external-signer accounts may already have NIP-65 relay metadata and KeyPackages while lacking the independently required inbox route.

Hosts need an explicit, durable, actionable setup state so they can stop onboarding, ask the user to select an inbox relay, publish the resulting kind `10050`, and resume setup without interpreting error strings or inventing a client-local lifecycle.

This is the local-account complement to #1673. That issue prevents a sender from canonicalizing an invitation for an unrouteable recipient. This issue prevents a local account from being presented as invite-ready before its own inbox setup is complete.

## Current boundary

`marmot-app` already exposes `AccountSetupReadiness` with `Initializing`, `LocalReady`, `Publishing`, `NetworkReady`, and `RecoveryRequired`. `AccountRelayListStatus` also distinguishes missing NIP-65 and inbox lists.

The remaining gaps are:

- there is no readiness state that means user action is required to select an inbox relay;
- account login bindings primarily return `AccountSummaryFfi`, discarding the richer setup readiness and relay-list result available internally;
- hosts have no explicit inbox-only completion command that resumes the durable setup lifecycle;
- hosts have no setup-status change event and would need to poll;
- completed or legacy setup state can be interpreted as `NetworkReady` without re-establishing the new inbox-readiness invariant;
- setup can publish missing relay lists from caller-provided defaults, but an existing identity should not have an inbox choice made silently on its behalf.

## Expected behavior

MDK owns the durable account-readiness truth. Clients own relay recommendations, localized explanation, and onboarding presentation.

After an imported key or external signer establishes the account identity, setup performs a bounded, fresh kind `10050` lookup using the existing freshness and anti-clobber rules:

- If at least one safe inbox endpoint is resolved, setup continues without rewriting the list.
- If the lookup completes and no safe inbox endpoint exists, setup stops in a durable user-action state such as `InboxRelayRequired`.
- If the lookup cannot complete because of timeout, connectivity, or directory failure, MDK reports a typed retryable/unknown outcome rather than treating the inbox as absent or publishing over possibly newer metadata.

The account may be locally available for the limited work needed to finish setup, but it is not `NetworkReady`, does not advertise a new KeyPackage, and is not presented to hosts as invite-ready.

[...]

PR #1681, opened 2026-09-04
fix(app): require inbox routes before invites

### LINKED ISSUE #1673: Do not canonicalize group invitations when the recipient has no resolved Marmot inbox route
## Summary

MDK can treat an account as invite-ready when it has a current KeyPackage and a NIP-65 relay list but no Marmot inbox relay route. The Add commit can then become canonical while Welcome publication is guaranteed to have no destination, leaving the inviter's roster ahead of the invitee with only a later `WelcomeDeliveryPending` recovery signal.

A historical report described intermittent missing invitations and an apparent group fork while roughly 30 people were invited incrementally over two days. The inviter used an Amber-authenticated existing Nostr account with a pre-existing NIP-65 list and a large social graph. No relay trace or audit log exists for that group, so the report does not prove this mechanism and does not establish NIP-65 or graph size as causal. The source defect below is independently reproducible and matches one way that false settled membership can occur.

## Source-confirmed behavior

A current local, cached, or prewarmed KeyPackage marks the target resolved immediately, without checking the target's cached inbox route: [`member_key_packages.rs` lines 247–315](https://github.com/marmot-protocol/mdk/blob/d2e38adf05c7bf6861a1fc9031a6a9c3ffe28c8c/crates/marmot-app/src/directory/member_key_packages.rs#L247-L315).

For unresolved KeyPackages, member resolution refreshes relay-list state only when a target's NIP-65 relay list is empty. It does not refresh a target whose NIP-65 list exists but whose Marmot inbox list is missing: [`member_key_packages.rs` lines 317–327](https://github.com/marmot-protocol/mdk/blob/d2e38adf05c7bf6861a1fc9031a6a9c3ffe28c8c/crates/marmot-app/src/directory/member_key_packages.rs#L317-L327).

After KeyPackage fetch, resolution reports the target as successful based on the KeyPackage result alone and retains the incomplete relay-list snapshot: [`member_key_packages.rs` lines 538–560](https://github.com/marmot-protocol/mdk/blob/d2e38adf05c7bf6861a1fc9031a6a9c3ffe28c8c/crates/marmot-app/src/directory/member_key_packages.rs#L538-L560). The existing `member_resolution_fixture` supplies only kind 10002 (NIP-65) plus a KeyPackage, and `member_key_package_set_batches_shared_relay_and_reuses_prewarm` currently accepts that target as fully resolved: [`tests.rs` lines 7424–7493](https://github.com/marmot-protocol/mdk/blob/d2e38adf05c7bf6861a1fc9031a6a9c3ffe28c8c/crates/marmot-app/src/tests.rs#L7424-L7493).

Invitation then uses those resolved KeyPackages before preparing and publishing the Add commit: [`client/mod.rs` lines 1907–1937](https://github.com/marmot-protocol/mdk/blob/d2e38adf05c7bf6861a1fc9031a6a9c3ffe28c8c/crates/marmot-app/src/client/mod.rs#L1907-L1937). Welcome routing is built separately from directory inbox lists; an empty inbox list installs no directory route: [`lib.rs` lines 3655–3682](https://github.com/marmot-protocol/mdk/blob/d2e38adf05c7bf6861a1fc9031a6a9c3ffe28c8c/crates/marmot-app/src/lib.rs#L3655-L3682). A Welcome without a route fails with `MissingInboxRoute`: [`lib.rs` lines 6026–6037](https://github.com/marmot-protocol/mdk/blob/d2e38adf05c7bf6861a1fc9031a6a9c3ffe28c8c/crates/marmot-app/src/lib.rs#L6026-L6037).

This path exists unchanged in the MDK revision currently vendored by White Noise Android (`2bbcca3ebe4a971152412c16c3049cc7bd08d278`) and current MDK `master` (`d2e38adf05c7bf6861a1fc9031a6a9c3ffe28c8c`).

## Protocol boundary

NIP-65 and Marmot inbox lists have different roles. The Nostr transport sends KeyPackage events to the account's NIP-65 write relays, while Welcome events are sent to the invitee's Marmot inbox relays: [`transports/nostr.md` lines 327–352](https://github.com/marmot-protocol/marmot/blob/4a2bc65f8db5866cec3b2a127dedb37818eaf207/transports/nostr.md#L327-L352) and [`transports/nostr.md` lines 404–446](https://github.com/marmot-protocol/marmot/blob/4a2bc65f8db5866cec3b2a127dedb37818eaf207/transports/nostr.md#L404-L446). Falling back to NIP-65 relays for Welcome delivery would violate that ownership split.

The Marmot publish lifecycle intentionally makes each Welcome an independent retryable obligation after the Add commit becomes canonical: [`protocol-core/publish-lifecycle.md` lines 68–75](https://github.com/marmot-protocol/marmot/blob/4a2bc65f8db5866cec3b2a127dedb37818eaf207/protocol-core/publish-lifecycle.md#L68-L75). This issue does not change that protocol rule. It addresses the avoidable local case where MDK already knows, before consuming a KeyPackage and creating the Add, that it has no legal destination for the resulting Welcome.

## Expected behavior

For commit-purpose member resolution, MDK should independently resolve both required inputs for every invitee: a current compatible KeyPackage and at least one safe Marmot inbox relay. A cached NIP-65 list must not suppress a missing-inbox lookup.

If no safe inbox route can be resolved within the existing bounded directory deadline, invitation should return a typed per-recipient readiness error before preparing or canonicalizing the Add commit. The failure must not consume the target's KeyPackage, change local membership, or enqueue a Welcome that is certain to fail. Prewarm may cache independently fetched material, but its result must not represent a KeyPackage-only target as fully invite-ready.

## Acceptance criteria

- Missing kind 10050 state is queried independently when kind 10002 is already present.
- Commit-purpose resolution requires a current compatible KeyPackage and at least one safe inbox endpoint for every canonical target.
- A target with KeyPackage + NIP-65 but no safe inbox endpoint returns a stable typed readiness outcome before MLS mutation; no member is added and no KeyPackage is consumed.
- A mixed multi-member request is atomic at the preflight boundary: one unrouteable target cannot produce partial local roster success, while all valid fetched directory data remains reusable for an explicit retry.
- Delayed or refreshed inbox metadata is re-evaluated with the existing freshness and anti-clobber rules; cached stale events cannot overwrite a newer route.
- Existing post-commit Welcome retry remains intact for acknowledgement loss, process death, and routes that become unavailable after preflight.
- Privacy-safe diagnostics expose only aggregate target counts, missing-route reason categories, stage, attempt count, and bounded timing; they do not log account IDs, pubkeys, relay URLs, group IDs, payloads, or KeyPackage bytes.

## Regression coverage

Add deterministic tests named for the behavioral boundary:

- `invite_with_key_package_but_no_inbox_route_fails_before_commit`
- `mixed_invite_route_readiness_does_not_consume_key_packages_or_mutate_membership`
- `missing_inbox_is_discovered_when_nip65_is_cached`
- `resolved_inbox_route_survives_restart_and_delivers_exact_welcome`
- `thirty_incremental_invites_with_large_directory_cache_converge_without_false_success`

Keep the existing `invite_members_returns_before_blocked_welcome`, `confirmed_invite_welcome_resumes_after_restart`, and large-group incremental canaries to prove that post-commit retry and normal convergence are unchanged.

## Related work and non-goals

- [marmot-protocol/mdk#352](https://github.com/marmot-protocol/mdk/issues/352) added durable per-Welcome failure and redelivery after a commit; it does not validate invite readiness before the commit.
- [marmot-protocol/mdk#1641](https://github.com/marmot-protocol/mdk/issues/1641) covers recipient-side Welcome ingest observability; this issue is sender-side route readiness.
- [marmot-protocol/whitenoise-android#2452](https://github.com/marmot-protocol/whitenoise-android/issues/2452) owns per-account UI failure state and user recovery. After this issue has a canonical URL, that Android tracker should link it as an upstream dependency rather than duplicate the MDK fix.
- [marmot-protocol/mdk#1638](https://github.com/marmot-protocol/mdk/issues/1638) covers oversized bulk Welcome wrapping at 64 members. Incremental invitation remains a distinct path and the historical report involved roughly 30 incremental invitees.

This issue does not change Marmot's post-canonical Welcome lifecycle, infer relay behavior from the reporter's social graph, fall back from kind 10050 to NIP-65 relays, or claim reproduction of the historical group.

erskingardner:
## Summary

- resolve each invitee KeyPackage and kind-10050 inbox state independently before an MLS Add is prepared
- require at least one relay-safety-approved inbox endpoint, while preserving newer cached relay metadata and reusable prewarm results
- return a stable per-recipient MissingMemberInboxRoute error through app, UniFFI, CLI JSON, and the append-only C status surface
- retain the existing post-commit Welcome retry path for acknowledgement loss, unavailable routes, and process restart

## Test plan

- `just fast-ci`
- `cargo test -p marmot-c --features alloc-audit`
- `just c-header` plus a clean generated-header diff
- `just c-smoke`
- `invite_with_key_package_but_no_inbox_route_fails_before_commit`
- `mixed_invite_route_readiness_does_not_consume_key_packages_or_mutate_membership`
- `missing_inbox_is_discovered_when_nip65_is_cached`
- `resolved_inbox_route_survives_restart_and_delivers_exact_welcome`
- `thirty_incremental_invites_with_large_directory_cache_converge_without_false_success`
- existing `invite_members_returns_before_blocked_welcome` and `confirmed_invite_welcome_resumes_after_restart` canaries

## Baseline note

A full `cargo test -p marmot-app` run also encountered unrelated epoch-backfill timing failures. `deferred_primary_epoch_backfill_rotates_behind_queued_older_operation` reproduced in isolation on untouched current `origin/master`; all invite-readiness and Welcome-delivery tests listed above pass on this exact head.

Fixes #1673
Related: #1679

## Summary by CodeRabbit

- **New Features**
  - Improved member inbox relay discovery by combining cached and newly fetched relay information.
  - Resolved relay routes are retained for future invitations and restored after app restarts.
  - Invite and recovery flows better support multiple members and incremental batches.

- **Bug Fixes**
  - Added clear, typed error reporting when a member lacks a safe inbox route, including repair guidance.
  - Prevented duplicate membership changes and preserved valid results when some member lookups fail.
  - Failed readiness checks no longer persist incomplete groups.


crates/marmot-app/src/tests.rs
```diff
@@ -7245,6 +7250,15 @@ async fn fresh_key_package_for_account(
     session.fresh_key_package().await.unwrap()
 }
 
+fn remember_test_member_inbox(app: &MarmotApp, account_id_hex: &str, relay: &str) {
```
erskingardner, line 7254: Blocking: this seeding has to reach every scripted-relay test that invites a local account — several were missed. CI `Rust tests (1/4)` at this head fails deterministically (both nextest attempts) in `a_failed_ingest_leaves_the_delivery_retryable_on_the_reused_client`: `create_group("retryable failed ingest", &[bob...])` at `tests.rs:15339` now unwraps `Err(MissingMemberInboxRoute(...))` at `tests.rs:15341` because bob has no seeded inbox route.

The same unmigrated shape appears at:
- `tests.rs:10792` — `reconnect_drains_deferred_hydration_before_steady_state_serves_groups_body` (creates `STARTUP_HYDRATION_BATCH_SIZE_FOR_TEST + 1` groups with local `donor`)
- `tests.rs:17410` — `reconcile_repairs_stale_three_member_count_on_two_member_direct_body` (bob)
- `tests.rs:17474` — `reconcile_repairs_stale_two_member_count_on_three_member_group_body` (bob + carol)

All four use `ScriptedPushRelayClient` with local invitee accounts and no kind-10050 source, so the new readiness check correctly rejects them — the tests just need the route seeded. Note that nextest fail-fast canceled 597 tests in shard 1/4, and shards 2/4, CLI e2e, and Relay runtime tests were still pending at review time, so there may be further fallout beyond these three. Suggest seeding the inbox route here (or a fixture equivalent) in each and re-running the full `cargo test -p marmot-app` locally before push.

crates/marmot-app/src/directory/member_key_packages.rs
```diff
@@ -169,6 +175,31 @@ struct MemberTarget {
     relay_lists: AccountRelayListStatus,
 }
 
+fn merge_member_relay_lists(
```
erskingardner, line 187: Non-blocking: this helper carries the "cached stale events cannot overwrite a newer route" invariant, but neither rule is obvious from the condition alone. A short doc comment would help: (1) strictly newer `created_at` wins per list and ties keep the current list — which is also what repairs the old `relay_lists_from_records` behavior where a fresh-accepted but older record could clobber a newer cached list; (2) the `created_at == 0` arm exists so timestamp-less statuses (e.g. setup-derived or fixture-built lists) can still populate an empty slot without overriding real timestamped data.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T11:43:00Z` |
| Commit reviewed | `7a8abffc8fb54a0cb45917a650f35bfe6c4ff125` |
| Recommended action | **Fix blocking issues before merge** (posted as a comment: GitHub does not allow requesting changes on your own PR) |

## Summary

The design and implementation are exactly what #1673 prescribes, and I verified the core logic closely. One blocking problem: CI is red at this head — the new readiness invariant breaks existing scripted-relay tests that were not migrated to the new `remember_test_member_inbox` seeding.

## Blocking

**CI `Rust tests (1/4)` fails deterministically at this head** (both nextest attempts): `a_failed_ingest_leaves_the_delivery_retryable_on_the_reused_client` unwraps `Err(MissingMemberInboxRoute(...))` at `crates/marmot-app/src/tests.rs:15341`. The test invites local-account bob over `ScriptedPushRelayClient` with no seeded kind-10050 route, so the new preflight correctly rejects it — the test was simply missed by the migration. At least three more tests share the exact same unmigrated shape (`tests.rs:10792`, `tests.rs:17410`, `tests.rs:17474`), and nextest fail-fast canceled 597 tests in that shard while shards 2/4, CLI e2e, and Relay runtime tests were still pending at review time, so the full fallout is not yet known. Details inline on the helper. Please migrate all affected tests and re-run the full `cargo test -p marmot-app` locally — the PR body currently reports only the unrelated epoch-backfill flake from that suite, which this failure postdates or was masked by.

## Verified as sound (non-blocking)

- **Premise and scope**: the fix addresses the source-confirmed root cause from #1673 (KeyPackage-only targets reported invite-ready; Add canonicalized with no legal Welcome destination) without touching the post-commit Welcome retry lifecycle, and correctly stays out of #1679 territory.
- **Atomicity**: resolution returns `MissingMemberInboxRoute` before any MLS mutation, and fetched KeyPackages/relay lists remain reusable for the explicit retry — both asserted by the new mixed-batch and single-target tests. All five issue-named regression tests exist, and the existing blocked-Welcome/restart canaries are retained.
- **Anti-clobber**: `fresh_relay_list_status_from_records` still freshness-filters records, and `merge_member_relay_lists` now compares `created_at` — this actually repairs a hole in the old `relay_lists_from_records`, where a fresh-accepted but older record could wholesale replace a newer cached list.
- **Behavior change is contained**: local-account invitees now also need a resolvable safe inbox route. That is semantically correct (their other devices receive Welcomes via inbox) and stays offline-capable because account setup already persists the account's own inbox list into the directory (`src/lib.rs:1993`, `:2332`).
- **Privacy and surfaces**: `AppError::MissingMemberInboxRoute` Display stays identity-free (covered by `app_error_display_does_not_expose_group_or_account_ids`), `privacy_safe_kind` and the projection/read-marker codes are aggregate-only, the CLI JSON shape matches the existing `missing_key_package` contract, the UniFFI variant carries the typed recipient, and the C status is append-only (`= 62`) with the checked-in header regenerated.
- **Persistence ordering**: commit-purpose relay-list persistence runs before the readiness check returns, so failed preflights still leave fetched directory data reusable — this is load-bearing for `resolved_inbox_route_survives_restart_and_delivers_exact_welcome`.

## Non-blocking

- `merge_member_relay_lists` deserves a short doc comment on its precedence and `created_at == 0` rules (inline).
- The disclosed `deferred_primary_epoch_backfill_rotates_behind_queued_older_operation` flake on master deserves a tracking issue if one does not exist yet.

Once the missed scripted-relay tests are seeded and CI is green at the new head, this is a merge from my perspective.

crates/marmot-app/src/directory/member_key_packages.rs
```diff
@@ -335,14 +393,56 @@ impl MarmotApp {
...
+                            .inbox
+                            .relays
+                            .iter()
+                            .cloned()
+                            .map(TransportEndpoint)
+                            .collect(),
+                        "member invite inbox readiness",
+                    );
+                    if safe_inbox.is_empty() {
+                        return Err(AppError::MissingMemberInboxRoute(
```
erskingardner, line 453: **Blocking (CI red):** the readiness gate reads only `target.relay_lists` (directory entry + fetch merge), but the Welcome publish path builds routes from a different, authoritative source: `routing_for` (`crates/marmot-app/src/lib.rs:3656`) installs local-account **profile** `inbox_endpoints` first and only fills missing entries from the directory. A same-device local account can therefore have a valid publish route and still fail this preflight.

That's exactly what `a_failed_ingest_leaves_the_delivery_retryable_on_the_reused_client` (`crates/marmot-app/src/tests.rs:15341`, feature-gated behind `test-policy-overrides` so the local baseline run skipped it) now hits deterministically: bob's route is registered via `bob_client.sync()`, yet `create_group` returns `MissingMemberInboxRoute`.

Two possible resolutions — please pick deliberately:

1. Make the gate consult the same source routing uses for local targets (e.g., check the local profile's `inbox_endpoints` when `target.local_label` is set, like the local-KeyPackage fast path above), or
2. Establish that local accounts' directory entries are always seeded with their inbox list in production, and patch the failing test with `remember_test_member_inbox` like its siblings.

Option 1 is the more principled fix: the preflight should predict what the publish path will actually do, and the directory is a cache that can diverge from the profile.

crates/marmot-app/src/directory/member_key_packages.rs
```diff
@@ -335,14 +393,56 @@ impl MarmotApp {
...
+                        || !target.relay_lists.inbox.relays.is_empty()
+                    {
+                        self.remember_directory_relay_lists(
+                            &target.account_id_hex,
+                            &target.relay_lists,
+                        )?;
+                    }
+                }
+            }
+            MemberResolutionPurpose::Prewarm => {
```
erskingardner, line 422: Non-blocking: `prewarm_group_member_key_packages` now hard-fails with `MissingMemberInboxRoute` when any canonical target lacks a safe inbox route, but its doc comments (here at lines 235-238 and on the runtime command in `runtime/commands.rs`) still describe prewarm purely as a non-consuming cache warmer. Worth stating the readiness enforcement explicitly, including that fetched material is still cached despite the error (which `missing_inbox_is_discovered_when_nip65_is_cached` asserts). Hosts that call prewarm opportunistically — e.g., while a user is still composing an invite list — will now receive typed errors where they previously got a summary.

crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -10769,6 +10769,104 @@ async fn confirmed_invite_welcome_resumes_after_restart() {
     runtime.shutdown().await;
 }
 
+/// mdk#1673: directory resolution persists the recipient's explicit inbox
+/// route, and a later process can use that route to deliver one Welcome.
+#[tokio::test]
+async fn resolved_inbox_route_survives_restart_and_delivers_exact_welcome() {
```
erskingardner, line 10775: Non-blocking: all the new failure-boundary tests exercise `invite_members`. `create_group` with initial members goes through the same Commit-purpose resolution, so the gate covers it, but only incidentally via the fixtures patched with `remember_test_member_inbox`. A targeted `create_group`-with-unrouteable-member-fails-before-commit test would pin the create path explicitly (it's also the path the CI-failing test above exercises).

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T11:44:52Z` |
| Commit reviewed | `7a8abffc8fb54a0cb45917a650f35bfe6c4ff125` |

**Blocking: CI is red with a deterministic failure caused by the new readiness gate.**

`Rust tests (1/4)` fails on `a_failed_ingest_leaves_the_delivery_retryable_on_the_reused_client` (`crates/marmot-app/src/tests.rs:15341`): `create_group` now returns `MissingMemberInboxRoute` for bob. That test is `#[cfg(feature = "test-policy-overrides")]`, so the local baseline run in the PR description would not have executed it. See the inline comment on the readiness check for the root cause — the gate consults only the directory entry, while the Welcome publish path (`routing_for` in `lib.rs`) installs local-account profile `inbox_endpoints` first and the directory second, so same-device local accounts can have a perfectly valid publish route and still fail preflight. Please either make the gate consult the same source routing uses for local targets, or justify that local accounts' directory entries are always seeded and patch the test like its siblings. Remaining CI jobs were still running at review time.

**Otherwise this is a well-scoped fix for #1673:**

- Independent kind-10050 lookup when kind-10002 is cached: correct — `relay_list_unresolved` now triggers discovery on a missing/unsafe inbox regardless of KeyPackage state.
- `merge_member_relay_lists` anti-clobber preserves newer cached routes over stale fetched ones, and keeps prior "no observed record → retain cached" behavior.
- The gate sits before MLS mutation on both invite and create paths, is per-recipient typed through UniFFI/C/CLI, and the C status is appended (`62`) without renumbering.
- KeyPackage non-consumption on failed preflight and reuse on explicit retry is asserted; post-commit Welcome retry canaries are retained.
- Diagnostics stay privacy-safe (aggregate counts, context labels, identity-free `Display`), and all five issue-required regression tests plus the unsafe-inbox variant are present.

Minor notes inline: stale prewarm doc contract, and no targeted create-path failure test.

crates/marmot-app/src/directory/member_key_packages.rs
erskingardner: Resolved deliberately with option 2 in b1e1906e. routing_for obtains local profile endpoints through account_inbox_endpoints, which can synthesize configured defaults when no kind-10050 state exists. Accepting that source in preflight would reintroduce the fallback explicitly rejected for #1673. Normal account setup persists the account kind-10050 list; these tests call create_account directly and bypass setup, so their required inbox state is now seeded explicitly. The exhaustive feature-enabled shard passes 1,113/1,113.

crates/marmot-app/src/tests.rs
erskingardner: Fixed in b1e1906e. Seeded explicit kind-10050 state in all four identified scripted-relay fixtures. The exact feature-enabled CI shard was rerun with fail-fast disabled and passed all 1,113 tests.

crates/marmot-app/src/directory/member_key_packages.rs
erskingardner: Addressed in b1e1906e. Both the MarmotApp method and runtime command docs now state that prewarm enforces safe inbox readiness, returns the typed missing-route error, and retains successfully fetched material.

crates/marmot-app/src/directory/member_key_packages.rs
erskingardner: Addressed in b1e1906e. The helper now documents strict created_at precedence, tie behavior, and why timestamp-less state may fill only an empty timestamp-less slot.

crates/marmot-app/tests/relay_runtime.rs
erskingardner: Addressed in b1e1906e with create_group_with_key_package_but_no_inbox_route_fails_before_group_creation. It asserts the typed error, no persisted group, and successful explicit retry without refetching the KeyPackage.

Datawav (approved):

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T12:30:00Z` |
| Commit reviewed | `b1e1906e8fdd696931d14400ab70936b05c0b866` |
| Recommended action | **Merge** (posted as a comment: GitHub does not allow approving your own PR) |

## Summary

The new commit `b1e1906` resolves the one blocking finding from my previous review, and CI is fully green at this head — all 27 checks pass, including Rust tests 1–4/4, Relay runtime tests, CLI e2e, and C bindings smoke. The shard that previously failed deterministically now passes.

## Verified

- **Blocking finding resolved**: all four missed scripted-relay tests are now seeded via `remember_test_member_inbox` (reconnect-hydration donor, failed-ingest bob, and both reconcile-repair fixtures), with setup-only changes and no test-logic edits.
- **Create-boundary coverage added**: `create_group_with_key_package_but_no_inbox_route_fails_before_group_creation` proves the founding-Add path fails before any group is persisted and that the explicit retry reuses the unconsumed KeyPackage without refetching — closing the one coverage gap from my earlier pass.
- **Doc comments**: `merge_member_relay_lists` now documents the strictly-newer/tie-retains/`created_at == 0` precedence rules; the `prewarm_group_member_key_packages` docs on both the app and runtime-command surfaces state the inbox-readiness contract, the typed error, and cached-material reuse.
- **No regressions in prior verdict**: nothing in this delta weakens the atomicity, anti-clobber, privacy, persistence-ordering, or append-only C-surface properties verified previously.

Ready to merge.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T12:28:46Z` |
| Commit reviewed | `b1e1906e8fdd696931d14400ab70936b05c0b866` |

Follow-up review of `b1e1906e` ("test(app): cover inbox readiness across create fixtures"). All findings from the previous review are resolved and full CI is green:

- **Blocking CI failure — fixed and justified.** The previously failing `a_failed_ingest_leaves_the_delivery_retryable_on_the_reused_client` now seeds bob's kind-10050 state via `remember_test_member_inbox`, whose doc accurately describes the production invariant: both account bootstrap (`lib.rs:1993`) and the relay-list publication path (`lib.rs:2332`) persist the published inbox list into the durable directory after acknowledgement. For an entry missing locally, the preflight's network discovery falls back to querying the published kind-10050 event, so the directory-vs-profile divergence converges rather than producing a spurious failure. Verified the equivalent fixtures for create, reconcile, and reconnect paths were also patched consistently.
- **Prewarm doc contract — updated** in both `member_key_packages.rs` and `runtime/commands.rs`, stating the readiness error and that fetched material stays cached on failure.
- **Create-path coverage — added.** `create_group_with_key_package_but_no_inbox_route_fails_before_group_creation` pins the founding-create boundary explicitly, including no persisted group and KeyPackage reuse on retry.

The core gate, merge/anti-clobber logic, typed error propagation (UniFFI/C/CLI), append-only C status, privacy-safe diagnostics, and the full issue-required regression suite were reviewed previously and remain sound. No new blocking findings. (GitHub does not allow self-approval, hence the comment event.)

PR #1682, opened 2026-09-04
Raise deferred-peel capacity with byte budgets

### LINKED ISSUE #1672: Evaluate raising the per-group deferred-peel capacity after #1670
## Summary

After #1670 makes the offline-catch-up scenarios model the engine's intended refusal, drain, and retained-history refetch loop, evaluate whether the per-group `PeelDeferred` capacity should increase from 256 rows.

This is follow-on tuning work, not the fix for #1670. Raising the cap first would make some current scenarios green by moving the refusal boundary without proving that the required recovery loop works.

## Motivation

The current 256-row cap is a resource-safety bound, not a protocol or cryptographic limit. It prevents attacker-mintable, uniquely rewrapped opaque transport objects from growing durable storage and retry work without bound.

The new offline catch-up campaign also shows that legitimate retained history can cross that boundary: the 384-application/12-commit natural-history case refuses 107 later inputs during its first bulk replay and succeeds after the backlog drains and history is fetched again. A higher cap could reduce benign refetch cycles and reconnect latency, provided it does not create an unacceptable storage, CPU, memory, or battery exposure.

## Dependency

Blocked by #1670. First make the simulator exercise the same bounded refusal/refetch behavior expected from the production app. Use that corrected recovery path as the baseline when comparing cap values.

## Investigation

Compare at least 256, 512, and 1,024 retained rows using the file-backed offline catch-up profiles and a production-shaped app-runtime/relay path where available.

Measure:

- peak and final `PeelDeferred` row counts;
- retained opaque payload bytes and total database growth, not only row count;
- peak RSS and catch-up wall time;
- background sweep count and cumulative peel/candidate-context work;
- number of `ResourceRefused` outcomes and retained-history refetches;
- latency until the latest message becomes visible and until exact pending-work closure;
- impact on current traffic while historical work drains;
- hostile unique-wrapper flood cost per group and across multiple groups.

Evaluate whether the right control is:

1. a higher fixed row cap;
2. a platform/configuration-sensitive cap;
3. a row cap paired with per-group and account-wide byte budgets; or
4. improved relay-side chunking/backpressure that makes a large increase unnecessary.

## Acceptance criteria

- #1670 is fixed first, and its 384-message case recovers through the unmodified 256-row refusal/refetch path.
- Publish benchmark results for 256, 512, and 1,024 across the 384- and 1,024-application profiles, including at least one duplicate/reordered history arm.
- Record an explicit recommendation and rationale; do not raise the constant solely to clear the current test case.
- If the cap is raised, add lower/at/over-cap regressions and retain typed `ResourceRefused` plus exact-ID redelivery after capacity drains.
- Preserve the 64-row background sweep bound, foreground work/time bounds, changed-context retry budget, and 30-day residence bound unless separately justified.
- Consider a byte budget before adopting a large row increase, because each row retains an opaque transport payload and the current limit is row-based.
- Verify that #1671 is unaffected; its competing-wave `Unrecoverable` halt reproduces after redelivery and is independent of cap sizing.

## Related work

- Parent: #1670
- Independent engine failure: #1671
- Original deferred-peel resource bound and redelivery contract: #339 / #1193

erskingardner:
## Summary

- raise the production deferred-peel limit from 256 to 512 rows per group
- add exact encoded-payload budgets of 16 MiB per group and 64 MiB per account-device database
- reconstruct byte usage from durable deferred rows after restart, then maintain it incrementally
- add feature-gated row and byte overrides for engine and retained-relay policy tests
- expose aggregate refusal counters and row/byte high-water marks in campaign measurements
- preserve storage failures while translating stored-payload codec limits into a typed, same-ID-retryable capacity refusal

The existing `TransportDeferredCapacity` outcome, history-backfill behavior, and exact-ID redelivery contract remain unchanged. No storage schema or binding migration is required.

## Capacity sweep

| Rows per group | Row refusals | Full-history syncs | Peak rows | Peak encoded bytes |
| ---: | ---: | ---: | ---: | ---: |
| 256 | 107 | 2 | 256 | 267,772 |
| 512 | 0 | 1 | 363 | 379,456 |
| 1,024 | 0 | 1 | 363 | 379,456 |

This removes the redundant refetch at 512; the 1,024 arm shows no additional benefit in the tested workload.

## Investigation scope

The shipping decision is based on the focused file-backed 384-message natural-history case and its 256/512/1,024 policy sweep. The 1,024-application catalog profile, duplicate/reordered-history arms, and hostile-flood RSS, wall-time, and database-growth characterization remain explicit follow-up work; this PR does not claim that 512 prevents every larger-profile refetch.

Issue #1671 scheduler and work bounds are unchanged: this PR changes durable deferred-peel admission only, not the 64-row sweep work limit, four-row foreground slice, or drain-pass semantics.

## Validation

- `just fast-ci`
- `cargo nextest run -p cgka-engine --features test-policy-overrides,test-crash-hooks --locked` — 526 passed, 1 skipped
- `cargo test -p cgka-engine --features test-policy-overrides --test deferred_peel_lifecycle --locked -- --test-threads=1` — 23 passed
- feature-enabled 256-row refusal/refetch regression wired into `adversarial-reliability-ci` — passed
- manual `deferred_peel_row_capacity_policy_sweep` at 256/512/1,024
- `cargo nextest run -p cgka-conformance-simulator --test protocol_decision_gate --locked` — 5 passed
- `just convergence-ledger-gate`
- GitHub Required CI — passed

Closes #1672


crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -2343,38 +2391,168 @@ impl<S: StorageProvider> Engine<S> {
...
+            if was_counted {
+                self.deferred_peel_account.bytes = self
+                    .deferred_peel_account
+                    .bytes
+                    .saturating_sub(previous_bytes)
+                    .saturating_add(bytes);
+            } else {
+                self.deferred_peel_account.bytes =
+                    self.deferred_peel_account.bytes.saturating_add(bytes);
+            }
```
erskingardner: **Blocking: the account byte total double-counts any group that starts deferring after the one-time reconstruction.**

`ensure_peel_deferred_usage_initialized` latches `deferred_peel_account.counted` and marks every group that exists at that moment `counted = true`. A group joined or created later in the same engine incarnation gets its `DeferredPeelGroupState` from `entry().or_default()` with `counted == false`, and every row it defers is added to `deferred_peel_account.bytes` incrementally by `note_peel_deferred_row_persisted` (which never sets `counted`). When the first deferred-peel sweep of that group reaches this `else` arm, `was_counted == false`, so the group's full deferred bytes are added to the account total a second time.

Concrete sequence: first deferred ingest in group A reconstructs account usage → user joins group B → B defers rows (`account.bytes += X`) → sweep of B lands here (`account.bytes += X` again). The inflation persists until restart — retirements only subtract true per-row amounts — and it also feeds `deferred_peel_peak_bytes_per_account`, so the new high-water metric over-reports for exactly the multi-group accounts the account budget exists to protect, and can produce premature `BytesPerAccount` refusals once inflated usage crosses the limit.

`state.deferred_bytes` always equals what this group has already contributed to the account total (pre-init groups were synchronized by `ensure`; post-init groups only grow via `note_row_persisted`), so the `was_counted` split is unnecessary — unconditional delta reconciliation is correct in both cases:

```rust
if self.deferred_peel_account.counted {
    self.deferred_peel_account.bytes = self
        .deferred_peel_account
        .bytes
        .saturating_sub(previous_bytes)
        .saturating_add(bytes);
}
```

A regression test shaped like `peel_deferred_account_byte_budget_is_restart_safe`, but with the second group's rows deferred *after* account initialization and no restart, would pin this: sweep the second group, then assert peak account bytes equal the exact durable sum rather than double.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -562,11 +570,21 @@ impl<S: StorageProvider> Engine<S> {
...
-                        Ok(record) if record.state == MessageState::PeelDeferred
-                    );
-                    if !already_retained && !self.has_peel_deferred_capacity(&group_id)? {
+                    let previous_deferred_payload_bytes = self
+                        .storage
+                        .get_message(&msg.id)
+                        .ok()
+                        .filter(|record| record.state == MessageState::PeelDeferred)
+                        .map(|record| record.payload.len());
+                    let deferred_payload_bytes = StoredMessagePayload::raw_transport(msg.clone())
```
erskingardner: Non-blocking: every undecryptable ingest now clones and encodes the payload once here to size it, and the accept path then encodes the same payload again inside `persist_transport_message_for_existing_group`. The refusal path inherently pays one encode for exact accounting, but accepted rows currently pay two. Consider threading the encoded bytes (or the `StoredMessagePayload`) into the persist call so the write reuses them.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T12:07:34Z` |
| Commit reviewed | `f757259af51cd55676b61b18ca61930485b5a4ee` |

## Summary

The calibration itself is well executed: the 256/512/1,024 sweep data justifies 512 rows, the byte budgets answer #1672's "consider a byte budget" requirement directly, restart-safe reconstruction from durable rows is the right shape, the test-override seam is properly feature-gated out of production, the exact-ID redelivery and typed `ResourceRefused` contracts are preserved, and the spec/inventory/plan/telemetry docs all moved with the code.

One blocking correctness bug in the new accounting (inline on `message_processor/mod.rs`): groups that begin deferring *after* the one-time account reconstruction get their bytes added to the account total twice — once incrementally at persist, once more at their first sweep via the `was_counted == false` arm. It fails closed (premature `BytesPerAccount` refusals, inflated peak metrics) rather than opening a safety hole, but it breaks the exact accounting this PR is built on and skews the policy telemetry it introduces. Suggested one-spot fix and regression test are in the inline comment.

## Non-blocking notes

- Accepted ingests encode the payload twice (once to size it, once inside the persist write) — inline on `ingest.rs`.
- Coverage vs. #1672's acceptance criteria: the sweep only exercises the 384-message natural-history profile. The issue also asked for the 1,024-application profile and at least one duplicate/reordered history arm, plus peak RSS / catch-up wall-time, and explicit confirmation that #1671 is unaffected. If that narrowing is deliberate (e.g. the byte budgets now dominate those axes), a sentence in the PR body or issue would close the loop.
- Most CI checks were still pending at review time; merge should wait on a green full run, especially the `test-policy-overrides` targets and the simulator suite.

Validation gap: I did not execute the tests locally (review-only); reliance is on the PR's reported runs and pending CI.

erskingardner (commented):
@/var/folders/56/yyq315m56k156tbxl5wxlx2c0000gn/T/opencode/review_body.md

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -562,11 +570,21 @@ impl<S: StorageProvider> Engine<S> {
...
+                        .encode()
+                        .map_err(|error| EngineError::Serialize(format!("{error:?}")))?
+                        .len();
+                    if !self.has_peel_deferred_capacity(
```
erskingardner, line 567: Non-blocking: this `clone` + `encode` runs before the capacity decision on every undecryptable message — including rows that will be refused and repeat redeliveries while at cap — and on the accept path the payload is encoded a second time inside `persist_transport_message` → `persist_stored_message_payload`. Exact byte accounting does require the encoded length before deciding, but the accept path could thread the already-encoded payload through to storage so an accepted row pays one encode, not two (same observation applies to the quarantine seam above at ~line 334).

crates/cgka-conformance-simulator/tests/offline_catchup_family.rs
```diff
@@ -354,16 +358,106 @@ async fn large_natural_history_retries_capacity_refused_relay_objects() {
...
+        panic!(
+            "{} at row limit {rows_per_group}: {error}",
+            case.scenario.name
+        )
+    })
+}
+
+#[cfg(feature = "test-policy-overrides")]
+#[tokio::test(flavor = "multi_thread")]
+async fn test_override_reproduces_256_row_capacity_refetch() {
```
erskingardner, line 427: As far as I can trace, no CI lane executes this test: `simulator-smoke`/`simulator-full` run without `test-policy-overrides`, the workspace test matrix excludes `cgka-conformance-simulator`, and the feature-enabled lanes (adversarial campaigns, `policy_sweeps`, weekly/release) run other test binaries. Since this PR flips the only non-gated scenario off the refusal→refetch recovery path, this test is the sole automated coverage of the bounded-refusal retained-history refetch loop from #1670/#1672, and it currently runs only on request. Suggest wiring `offline_catchup_family` into one of the feature-enabled lanes (e.g. alongside `policy_sweeps`) or stating the intended manual lane in PROPERTY_TESTS.md.

crates/cgka-conformance-simulator/tests/offline_catchup_family.rs
```diff
@@ -354,16 +358,106 @@ async fn large_natural_history_retries_capacity_refused_relay_objects() {
...
-            .filter(|observation| {
-                observation.client == "bob"
-                    && matches!(observation.mode, ScenarioRelaySyncModeV2::FullHistory)
-            })
-            .count()
-            >= 2,
-        "the capacity refusal must cause a bounded complete-history refetch"
+            .campaign_measurements
+            .deferred_peel_peak_rows_per_group
+            .is_some_and(|rows| rows > 0 && rows <= 512)
```
erskingardner: Non-blocking nit: 512 is hardcoded here while the engine-side flood test asserts against `MAX_PEEL_DEFERRED_ROWS_PER_GROUP`. If the production limit is recalibrated again, this bound silently drifts — consider importing the constant (it is `pub` in `cgka_engine::message_processor`).

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T12:10:30Z` |
| Commit reviewed | `f757259af51cd55676b61b18ca61930485b5a4ee` |

(A previously submitted review on this commit was minimized — its body was posted with a literal file path by mistake. This review replaces it.)

**Premise and direction**

The change is warranted and the chosen control matches the issue's own option 3 (row cap paired with per-group and account-wide byte budgets): the 256-row bound demonstrably caused a redundant full-history refetch in the benign 384-message profile, and a row-only cap was never a true storage bound — variable-length payloads made the per-group exposure payload-dependent, and no account-wide aggregate existed before (the old exposure scaled unboundedly with the number of known groups). Restart-safe reconstruction from durable rows, aggregate-only metrics, the typed `ResourceRefused` outcome, and exact-ID redelivery are preserved as claimed, and no schema or binding migration is needed.

The accounting core looks correct. The per-id byte map makes replacement and retirement idempotent — strictly better than the old count-only decrement, which could double-decrement on nested retire paths. I traced every transition out of `PeelDeferred` (`reingest_deferred_peel_row`, `replay_buffered_messages`, `mark_raw_transport_message_{failed,processed}_if_awaiting_retry`, `release_deferred_peel_row`): each calls `note_peel_deferred_row_retired` with a record enumerated for that same group, so the `group_id` → `record.group_id` refactor is equivalent at every call site. Replacement semantics (same-id growth charged as a delta, same-size retries always eligible) are consistent between the two ingest seams, and `previous_payload_bytes` (stored encoded length) matches the incoming `StoredMessagePayload::raw_transport(...).encode()` accounting used at persist.

**Concerns before merge**

1. **Issue #1672's investigation scope is only partially met.** The acceptance criteria ask for results across the 384- *and* 1,024-application profiles, including at least one duplicate/reordered-history arm, plus hostile unique-wrapper flood cost and RSS/wall-time/database growth. What is published is the 384-message natural history only. The recommendation rationale ("1,024 rows shows no additional benefit") holds for the tested workload, but the 512 cap clears the current test case by margin rather than evidence — a 1,024-application profile (peak well above 512) would still refuse and refetch, and the issue explicitly warns against raising the constant solely to clear the current case. Please publish the remaining arms or record an explicit deferral, and confirm #1671 is unaffected as the issue also asks.

2. **The refusal→retained-history-refetch recovery loop loses all automated CI coverage.** The non-gated scenario now asserts exactly one full-history sync (no refetch); the only test that still exercises the bounded-refusal refetch path is the feature-gated `test_override_reproduces_256_row_capacity_refetch`, and as far as I can trace, no CI lane runs `offline_catchup_family` with `test-policy-overrides`: `simulator-smoke`/`simulator-full` run without the feature, the workspace test matrix excludes `cgka-conformance-simulator`, and the feature-enabled lanes (adversarial campaigns, `policy_sweeps`, weekly/release) run other binaries. The recovery loop that was the point of #1670 would then run only on request. Suggest wiring the gated test into one of the feature-enabled lanes or documenting the manual lane explicitly in PROPERTY_TESTS.md.

3. **Validation status:** CI test suites were still in progress at review time (format, check, clippy, and the convergence policy pin are green). I reviewed the head commit only and did not execute the tests locally.

Inline notes cover a small double-encode cost on the accept path and a hardcoded 512 in the simulator assertion. Nothing blocking in the engine accounting itself.

crates/cgka-engine/src/message_processor/mod.rs
erskingardner: Confirmed — I traced the same sequence independently and this is a real double-count; my overall review above ("the accounting core looks correct") missed it, so treat this thread as overriding that sentence.

Reachability check: a group whose `DeferredPeelGroupState` entry is created after the one-time reconstruction can only come from `entry().or_default()` in `has_peel_deferred_capacity` (counted=false, deferred_bytes=0), and its bytes reach the account total only via `note_peel_deferred_row_persisted` — which never sets `counted`. So on that group's first sweep, `was_counted == false` even though its `deferred_bytes` is already part of `deferred_peel_account.bytes`, and the `else` arm adds them a second time. The inflation survives retirements (they subtract true per-row amounts) until restart, and `deferred_peel_peak_bytes_per_account` inherits it.

The unconditional delta reconciliation is correct in both cases: for groups synchronized by `ensure_peel_deferred_usage_initialized` the contribution came from the reconstruction, and for post-init groups it came from `note_row_persisted` — in both, `state.deferred_bytes` equals what the group has already contributed to the account total, so `account.bytes - previous_bytes + bytes` is right and the `was_counted` split is unnecessary. The proposed no-restart regression test (second group defers after account init, sweep it, assert the exact durable sum) pins it.

erskingardner: Fixed in ca66c462. Group refresh now always reconciles account bytes as old contribution minus previous group bytes plus current group bytes, so a group charged incrementally after account initialization cannot be added twice. Added both a focused accounting unit regression and a no-restart two-group engine regression that defers in A, starts deferring in B, sweeps B, and asserts the peak account bytes equal the exact durable sum.

crates/cgka-engine/src/message_processor/ingest.rs
erskingardner: Addressed in ca66c462. Deferred admission now prepares the encoded payload once, measures that exact Vec, and threads it into persistence. The accepted path no longer performs a second clone plus encode.

crates/cgka-engine/src/message_processor/ingest.rs
erskingardner: Addressed in ca66c462 with a shared prepared-payload seam. Capacity checks and durable writes use the same single encoding, while refused rows still encode only as needed to establish their exact storage size.

crates/cgka-conformance-simulator/tests/offline_catchup_family.rs
erskingardner: Addressed in ca66c462. The feature-enabled 256-row refusal-to-refetch regression is now invoked in adversarial-reliability-ci directly alongside policy_sweeps, and PROPERTY_TESTS.md records that coverage. I also ran the exact added command locally: 1 passed.

crates/cgka-conformance-simulator/tests/offline_catchup_family.rs
erskingardner: Fixed in ca66c462. The simulator assertion now compares against cgka_engine::message_processor::MAX_PEEL_DEFERRED_ROWS_PER_GROUP, eliminating drift if the production policy is recalibrated.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -325,20 +325,28 @@ impl<S: StorageProvider> Engine<S> {
...
+            let previous_deferred_payload_bytes = self
+                .storage
+                .get_message(&msg.id)
+                .ok()
+                .filter(|record| record.state == MessageState::PeelDeferred)
+                .map(|record| record.payload.len());
+            let deferred_payload_bytes = StoredMessagePayload::raw_transport(msg.clone())
+                .encode()
+                .map_err(|error| EngineError::Serialize(format!("{error:?}")))?
+                .len();
```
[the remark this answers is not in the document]
erskingardner: Fixed in ca66c462. Stored-payload codec-limit overflow at either deferred-retention seam now returns typed ResourceRefused with TransportDeferredCapacity, leaves the object unpersisted and same-ID retryable, and cannot abort the transport drain as EngineError::Serialize. The regression uses a causal-dependency list above the codec bound and verifies both the typed outcome and retry behavior.

docs/marmot-architecture/convergence-reliability-plan.md
```diff
@@ -275,7 +275,7 @@ Categories:
...
-| E3 | `MAX_PEEL_DEFERRED_ROWS_PER_GROUP` | 256 rows | resource | Bounds attacker-mintable undecryptable durable backlog | Over-cap input returns typed `ResourceRefused`, remains exact-ID eligible, and immediately arms history backfill | may affect locally available input until refetch | Flood, bounded-storage assertion, exact-ID recovery after drain, app backfill signal |
+| E3 | `MAX_PEEL_DEFERRED_ROWS_PER_GROUP`; `MAX_PEEL_DEFERRED_BYTES_PER_GROUP`; `MAX_PEEL_DEFERRED_BYTES_PER_ACCOUNT` | 512 rows/group; 16 MiB/group; 64 MiB/account | resource | Bounds attacker-mintable undecryptable durable backlog by both object count and exact encoded storage bytes, including floods spread across groups | Over-cap input returns typed `ResourceRefused`, remains exact-ID eligible, and immediately arms history backfill; restart reconstructs exact durable usage before admission | may affect locally available input until refetch | Row flood and reclaim, group/account byte boundaries, restart reconstruction, 256/512/1,024 retained-history sweep, app backfill signal |
```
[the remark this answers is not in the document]
erskingardner: Fixed in ca66c462. Both modified architecture documents now carry updated: 2026-09-04.

crates/cgka-engine/src/message_processor/mod.rs
```diff
@@ -2343,38 +2406,203 @@ impl<S: StorageProvider> Engine<S> {
...
+    /// refusal path.
+    pub(crate) fn prepare_deferred_peel_payload(
+        &self,
+        msg: &TransportMessage,
+    ) -> Result<PreparedDeferredPeelPayload, StoredMessagePayloadCodecError> {
+        let previous_payload_bytes = self
+            .storage
+            .get_message(&msg.id)
+            .ok()
+            .filter(|record| record.state == MessageState::PeelDeferred)
```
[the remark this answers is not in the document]
erskingardner: Fixed in b47ab912. The prepared-payload seam now distinguishes NotFound from other storage failures: NotFound means no previous row, but Busy and Backend errors propagate as EngineError::Storage. Only a StoredMessagePayload codec-bound failure becomes the typed capacity refusal. Added a focused regression proving a get_message failure is preserved.

crates/cgka-engine/src/message_processor/ingest.rs
```diff
@@ -562,43 +552,43 @@ impl<S: StorageProvider> Engine<S> {
...
-                                        .to_string(),
-                                },
-                            );
+                    let prepared = match self.prepare_deferred_peel_payload(msg) {
+                        Ok(prepared) => prepared,
+                        Err(DeferredPeelPayloadPreparationError::Storage(error)) => {
+                            self.return_unmodified_mls_group(&group_id, mls_group);
+                            return Err(EngineError::Storage(error));
                         }
+                        Err(DeferredPeelPayloadPreparationError::CodecLimit) => {
```
erskingardner, line 561: Non-blocking observability note: a codec-limit refusal returns the typed `TransportDeferredCapacity` outcome but never reaches `has_peel_deferred_capacity`, so it does not increment any `deferred_peel_{row,group_byte,account_byte}_capacity_refusals` counter. A flood of oversized-but-authenticated wrappers would therefore surface as `ResourceRefused` input dispositions (and arm history backfill) while all three new refusal counters read zero, which would misattribute the episode in campaign measurements. If that flood shape is worth diagnosing from metrics alone, consider counting codec-limit refusals under one of the existing counters (or a fourth) — otherwise fine to leave.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T12:58:39Z` |
| Commit reviewed | `b47ab9120213daf933d2116afe67fbf80bbb14e0` |

(Posted as a comment rather than an approval only because GitHub rejects APPROVE events from the PR author's own token.)\n
Rereview of the two follow-up commits (`ca66c46`, `b47ab912`) against my previous findings — all addressed and verified.

- **Account byte double-count (blocking):** fixed exactly right. `DeferredPeelGroupState::counted` is removed entirely (the account-wide latch is the only initialization flag left), and `reconcile_group_bytes` now always applies the delta `account.bytes - previous_group_bytes + new_group_bytes`. That is correct in both cases — pre-reconstruction groups contributed via `ensure_peel_deferred_usage_initialized`, post-reconstruction groups via `note_peel_deferred_row_persisted`, and in both `state.deferred_bytes` equals the group's live contribution, so unconditional delta reconciliation is the right invariant. The unit test plus the no-restart integration test (`later_deferring_group_first_sweep_keeps_exact_account_byte_total`, real second group created after account init) pin the reported shape.
- **Refusal-refetch CI coverage:** `test_override_reproduces_256_row_capacity_refetch` now runs in the `adversarial-reliability-ci` lane, chained by `convergence-nightly-lane` (simulator nightly + weekly hardening), and PROPERTY_TESTS.md documents the lane. Not per-PR CI, which is the right trade for a scenario this large — the recovery loop is automated again.
- **Investigation scope:** the new "Investigation scope" section explicitly records the 1,024-application profile, duplicate/reordered arms, and hostile-flood characterization as follow-up work and states #1671's bounds are untouched, rather than leaving the issue's acceptance criteria silently unmet. Given the byte budgets bound the storage exposure by construction and the overrides/metrics now exist for that follow-up characterization, this scoping is defensible.
- **Double encode on accept:** fixed — `prepare_deferred_peel_payload` produces the encoded payload once and `persist_encoded_transport_message_for_existing_group` threads it through to `persist_encoded_stored_message_payload`. The refactor preserves the echo/dedup short-circuit and the identical-record comparison semantics.
- **Hardcoded 512:** now imports `MAX_PEEL_DEFERRED_ROWS_PER_GROUP`.
- **New in the follow-ups:** the storage-error propagation from the previous-row lookup (previously swallowed by `.ok()`) is a correctness improvement with a unit test, and classifying codec-limit overflow as a typed, same-ID-retryable `ResourceRefused` rather than an internal `Err` is the right seam contract — pinned by `unencodable_deferred_transport_is_resource_refused`. One non-blocking observability note on that path is inline.

Validation: format/check/clippy/policy-pin/Tamarin are green on this head; the targeted suites pasted in the PR body (23 deferred-peel lifecycle tests including both new regressions, protocol decision gate, constant ledger gate) pass. The broad CI test partitions (Rust tests 1-4, simulator and vectors, C bindings smoke, CLI e2e, relay runtime) were still pending at approval time.

crates/cgka-engine/tests/deferred_peel_lifecycle.rs
```diff
@@ -1704,6 +1760,189 @@ async fn peel_deferred_rows_capped_per_group_under_flood() {
...
 }
 
+/// If account accounting was initialized by group A, a later deferral in
+/// group B is charged incrementally. Group B's first sweep must reconcile that
+/// contribution rather than adding the same durable bytes again.
+#[tokio::test]
+async fn later_deferring_group_first_sweep_keeps_exact_account_byte_total() {
+    let (mut alice, mut carol, storage, _peeler, group_a, _commit2, _commit3) =
+        carol_behind_two_epochs().await;
+    let group_b = add_group_two_epochs_ahead(&mut alice, &mut carol).await;
```
erskingardner: The production fix is correct, but this test doesn't actually exercise the bug it pins: group B is created and joined at line 1770, **before** the first capacity-sensitive ingest. When `carol.ingest(deferred_a)` triggers `ensure_peel_deferred_usage_initialized`, `list_groups()` already contains group B, so the old code marked it `counted = true` with zero rows — and the sweep at line 1786 would take the delta arm and reconcile cleanly. This test passes against the pre-fix code.

The double-count only triggered for a group that did **not** exist at reconstruction time. Move the group-B setup between the first ingest and the second deferral:

```rust
let deferred_a = send_app(&mut alice, &group_a, "group A deferred bytes").await;
assert!(matches!(
    carol.ingest(deferred_a.clone()).await.unwrap(),
    IngestOutcome::TransportDeferred { .. }
));
let bytes_a = storage.get_message(&deferred_a.id).unwrap().payload.len();

let group_b = add_group_two_epochs_ahead(&mut alice, &mut carol).await;
```

With that ordering, the pre-fix code charges group B's bytes twice at line 1786 (`account += bytes_b` incrementally, then `+= bytes_b` again via the `was_counted == false` arm) and the assertion fails; the fixed code passes. Worth verifying the test actually fails on the parent commit before merging.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T13:00:29Z` |
| Commit reviewed | `b47ab9120213daf933d2116afe67fbf80bbb14e0` |

## Re-review summary

Both findings from the previous review are addressed, and the fixes are the right shape:

- **Double-count fix (verified correct):** the per-group `counted` flag is gone and `refresh_peel_deferred_group_usage` now reconciles the account total as an unconditional delta via `reconcile_group_bytes`. Since `state.deferred_bytes` always equals what the group has already contributed, this is correct for both pre-reconstruction and post-reconstruction groups.
- **Double-encode eliminated:** `prepare_deferred_peel_payload` encodes once and the accept path threads the same bytes through `persist_encoded_transport_message_for_existing_group` into `persist_encoded_stored_message_payload`; the welcome-clobber guard is preserved.
- **Bonus hardening:** storage errors from the previous-row lookup now propagate instead of being swallowed by `.ok()`, and an unencodable transport object is a typed `ResourceRefused` rather than an internal `EngineError::Serialize` that would abort a drain. The refusal/audit/`retryable_unpersisted_ingest_id` duplication is consolidated in `peel_deferred_capacity_refused`. The 256-row refetch test is now wired into the adversarial reliability CI lane.

## Remaining item (blocking, small)

- The new integration regression test `later_deferring_group_first_sweep_keeps_exact_account_byte_total` has an ordering hole: group B is joined *before* the first capacity-sensitive ingest, so the account reconstruction already marks it counted, and the test would pass against the pre-fix code. Inline comment has the two-line reorder. Please also verify the reordered test fails on the parent commit — that's the evidence it pins the bug.

## Non-blocking observations

- `unencodable_deferred_transport_is_resource_refused` pins good behavior, but note the asymmetry it introduces: a capacity refusal clears once the backlog drains, while a codec-limit refusal is deterministic — exact-ID redelivery of that object will be refused forever, and the armed history backfill will refetch and re-refuse it. Acceptable for a pathological input (transport-level bounds would normally reject it first), but worth a sentence in the spec doc alongside the other refusal kinds.
- `note_peel_deferred_row_persisted` is called unconditionally after `persist_encoded_transport_message_for_existing_group`, which can silently no-op via the existing-group/welcome guard — charging accounting for a row that wasn't written. Unreachable in practice on these arms (the group was just loaded; id collision with an own staged welcome is infeasible), so no change needed, but the persist return value conveying "actually wrote" would make the accounting self-evidently exact.
- Previous review's coverage note stands: #1672's acceptance criteria mention the 1,024-application profile, a duplicate/reordered arm, RSS/wall-time, and a #1671-unaffected check; the sweep still only runs the 384-message natural-history profile.
- CI: check/format/audit/policy-pin green; Rust test shards, clippy, and simulator lanes still pending at review time — merge on a full green run.

Validation gap: review-only; I did not execute the tests locally.

crates/cgka-engine/tests/deferred_peel_lifecycle.rs
erskingardner: Fixed in 07ecee4c. Group B is now created only after group A triggers account reconstruction. The corrected test passes on the fixed implementation, and I verified the regression by reintroducing the old double-add behavior in an isolated worktree: it fails with 1272 observed bytes versus the exact 848-byte durable sum.

crates/cgka-engine/src/message_processor/ingest.rs
erskingardner: Leaving this unchanged deliberately. The three counters describe the three configured admission budgets; assigning a stored-codec representation limit to row, group-byte, or account-byte pressure would make those measurements inaccurate. Codec-limit refusal remains visible through the typed ingest outcome, capacity rejection audit, and history-backfill signal. A fourth end-to-end metric can be added separately if campaign-only diagnosis becomes necessary.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T13:27:58Z` |
| Commit reviewed | `07ecee4c15d7174c661fb51ff30ea3ecd9f2cce4` |

## Re-review summary

Approving. The final blocking item is resolved and everything else from the prior two rounds is in good shape:

- **Regression test now discriminates** ([previous finding](https://github.com/marmot-protocol/mdk/pull/1682#discussion_r3934209057)): group B is created after the first capacity-sensitive ingest, so account reconstruction runs without it in storage and the test exercises the original double-count path. The comment in the test says exactly why.
- **Account byte double-count fix** (round 1) verified correct: unconditional delta reconciliation, pre/post-reconstruction cases both exact.
- **Single-encode refactor** (round 1 note) verified: measured bytes flow into the persist write; welcome-clobber guard preserved.
- **Full CI green** on this head: all four Rust test shards, clippy, simulator and vectors, policy pin, Tamarin, e2e lanes, and all harness builds pass.

Non-blocking notes from round 2 still stand for follow-up at your discretion: the deterministic (never-clears) nature of codec-limit refusals could use a sentence in `cgka-engine-spec.md`, and #1672's broader acceptance criteria (1,024-application profile, duplicate/reordered arm, RSS/wall-time, explicit #1671 check) remain only partially covered — a closing note on the issue about that scoping would tidy the record. Neither blocks this PR.

Validation gap: review-only; tests not executed locally. CI is green on the reviewed commit.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T13:28:11Z` |
| Commit reviewed | `07ecee4c15d7174c661fb51ff30ea3ecd9f2cce4` |

Rereview of `07ecee4` (test-only delta since my last review) — verified, and it fixes a real weakness in the regression test that my last review overstated.

In the previous ordering, `add_group_two_epochs_ahead` ran **before** group A's deferral, so account reconstruction (`ensure_peel_deferred_usage_initialized`, triggered by A's first capacity-sensitive ingest) iterated `list_groups()` with group B already in storage. B was therefore synchronized by the reconstruction itself, and even the pre-fix code would have taken the delta branch on B's first sweep — the test would have passed against the bug it claimed to pin. My last review described it as "a real second group created after account init"; that was wrong, and this commit is the correction.

The new ordering is the genuine repro: A's deferral latches account reconstruction without B in storage → B is created after that boundary → B's `DeferredPeelGroupState` entry is born from `entry().or_default()` and its bytes are charged incrementally by `note_peel_deferred_row_persisted` → B's first sweep reaches `refresh_peel_deferred_group_usage`, where the pre-fix `was_counted == false` arm would add `bytes_b` a second time (`bytes_a + 2*bytes_b` vs. the asserted `bytes_a + bytes_b`). The comment on the moved call states exactly this.

No production code changed in this commit. CI is fully green on this head (29 checks pass, none pending), including the suites that were still running at my last review. The non-blocking codec-limit refusal-metrics note from my last review still stands as optional follow-up. Nothing further from me — this is ready to merge.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T13:27:58Z` |
| Commit reviewed | `07ecee4c15d7174c661fb51ff30ea3ecd9f2cce4` |

## Re-review summary

Recommending merge. (Posted as a comment review because GitHub does not allow the PR author to approve their own PR — my earlier comment review in this round had a malformed metadata table; this is the clean verdict.)

The final blocking item is resolved and everything from the prior two rounds is in good shape:

- **Regression test now discriminates** ([previous finding](https://github.com/marmot-protocol/mdk/pull/1682#discussion_r3934209057)): group B is created after the first capacity-sensitive ingest, so account reconstruction runs without it in storage and the test exercises the original double-count path.
- **Account byte double-count fix** (round 1) verified correct: unconditional delta reconciliation, pre/post-reconstruction cases both exact.
- **Single-encode refactor** (round 1 note) verified: measured bytes flow into the persist write; welcome-clobber guard preserved.
- **Full CI green** on this head: all four Rust test shards, clippy, simulator and vectors, policy pin, Tamarin, e2e lanes, and all harness builds pass.

Non-blocking notes from round 2 still stand for follow-up at your discretion: the deterministic (never-clears) nature of codec-limit refusals could use a sentence in `cgka-engine-spec.md`, and #1672's broader acceptance criteria (1,024-application profile, duplicate/reordered arm, RSS/wall-time, explicit #1671 check) remain only partially covered — a closing note on the issue about that scoping would tidy the record. Neither blocks this PR.

Validation gap: review-only; tests not executed locally. CI is green on the reviewed commit.

PR #1683, opened 2026-09-04
Persist direct-peer presentation in chat lists

### LINKED ISSUE #1517: Return durable resolved conversation presentation in chat-list results
## Outcome

Clients should receive the selected conversation title and avatar descriptor in the chat-list result, including on offline reopen. MDK owns group/peer selection, durable cached presentation and background invalidation; clients can remove display-only roster/profile lookups.

This is the **first implementation child of #1742**, covering the initial C2/C3 slice. It supersedes this issue's earlier peer-sidecar prescription and closed PR #1683. Preserve the existing downstream dependency on [Android #2149](https://github.com/marmot-protocol/whitenoise-android/issues/2149); that issue owns Android rendering/adoption.

## Ready-to-build scope

- Resolve title/avatar independently for named groups, unnamed two-person groups, missing profiles and roster changes.
- Persist selected presentation in the existing account projection; maintain profile dependencies durably without screen subscribers.
- Provide additive whole-row Rust/Swift/Kotlin/C reads and an initial snapshot plus subscription, with explicit fallback and recovery behavior.
- Preserve existing list/query cardinality, previews, read state, archive behavior and lower-level APIs. C4 adds bounded paging/filters; C7/#1554 adds offline avatar **bytes**. This slice persists descriptors, not a new image cache.

## Merge and launch sequence

| Unit | Deliverable | Depends on |
| --- | --- | --- |
| **P1** | Selection policy, account storage/dependencies, migration and bounded resumable backfill | Plan below |
| **P2** | Durable shared-profile change tracking, account catch-up and source invalidation/recovery | P1 |
| **P3** | Additive whole-result runtime/UniFFI/C API and subscription; binding/ownership tests | P1 + P2 |
| **P4a / P4b** | Android / iOS adoption, removing display-only lookup dependencies | Released P3 artifact |

P1/P2 can merge as internal foundations. P3 completes the shippable MDK capability; each flagship client can adopt it independently. Avoid an overlapping generic C2 framework issue: implement the maintenance needed by this slice here.

## Completion

- [x] P1 storage/policy and upgrade acceptance pass — merged #1753.
- [x] P2 no-subscriber, crash/restart and account-isolation acceptance pass — merged #1755.
- [x] P3 complete native contract, C compatibility and subscription acceptance pass — merged #1761 (`ee1e8812`).
- [x] Binding release and client adoption handoff recorded: [MarmotKit v0.9.20](https://github.com/marmot-protocol/mdk/releases/tag/marmotkit-v0.9.20), published 2026-09-08. Jeff confirmed on 2026-09-10 that both iOS and Android use the new projections. Android adoption remains linked to [Android #2149](https://github.com/marmot-protocol/whitenoise-android/issues/2149); iOS adoption is confirmed in the projection workstream, with further performance automation owned by the iOS repository.

Client releases are tracked under C9/the client repositories; they are parent completion evidence, not hidden additional MDK implementation in this issue. Full contract, transaction/recovery details, consumer boundaries and tests follow.

<details>
<summary>Implementation contract, persistence/recovery design, upgrades and acceptance matrix</summary>

# First chat presentation implementation slice

Implementation plan for [#1517](https://github.com/marmot-protocol/mdk/issues/1517), under
[#1742](https://github.com/marmot-protocol/mdk/issues/1742). This is the first C2/C3 slice; the C1 design needed for this
slice is ready. No implementation or performance success is claimed. Later screen contracts retain their own design
and acceptance work. The [parent contract](https://github.com/marmot-protocol/mdk/issues/1742) maps those boundaries.

## Outcome and scope

A native client obtains each conversation's selected title and avatar descriptor in its chat-list result, including
on offline reopen. MDK owns the group/peer choice, cached identity, invalidation and background refresh. The client
can remove title/avatar roster/profile lookups while keeping unrelated roster consumers.

This slice preserves the existing list query cardinality and archive-inclusion behavior. It is deliberately an
incremental replacement for current list presentation, not the final bounded/paged screen API. C4 adds bounded windows,
composable filters and account-attention changes using the same row presentation type. Existing message previews,
unread counts, drafts, delivery, pin and lifecycle behavior remain under their current owners in this slice.
Avatar **descriptors** are durable here; offline avatar bytes belong to C7/#1554. No retained-media implementation,
accepted-edit resolver, new protocol component or separate low-level binding package is required to ship this slice.

## Concrete public contract

The following names define the planned API, subject to repository naming conventions at implementation. Their semantics
and ownership are fixed for this slice. Rust owns the policy; UniFFI and C perform mechanical conversion.

```text
PresentedChatRow {
    row: ChatListRow                  // complete existing row fields, unchanged semantics
    presentation: ConversationPresentation
}
ConversationPresentation {
    title: PresentationText
    avatar: SelectedAvatar
    title_source: GroupName | PeerProfile | PeerFallback | GroupFallback | UnknownFallback
    avatar_source: GroupImage | PeerProfile | GroupFallback | UnknownFallback
    peer_id: optional opaque identity // only if durable roster identifies self + exactly one peer
    resolution: Cached | LastKnown | Fallback
}
PresentationText = Literal(text) | LocalizedFallback(kind, member_count?)
SelectedAvatar = RemoteImage(url, cache_key)
               | EncryptedGroupImage(existing_group_image_material, cache_key)
               | Placeholder(stable_seed, Group | Person | Unknown)
PresentedChatListSnapshot {
    rows: sequence<PresentedChatRow>
    presentation_version: { account_store_epoch, revision }
}
OpenedPresentedChatList {
    snapshot: PresentedChatListSnapshot
    subscription: PresentedChatListSubscription
}
PresentedChatListUpdate {
    subscription_generation, sequence
    snapshot: PresentedChatListSnapshot
}
```

- `open_presented_chat_list(account, include_archived)` returns the snapshot and an already-attached subscription.
  `presented_chat_list(account, include_archived)` and `presented_chat_list_row(account, group)` provide local one-shot
  reads for existing narrow/use-at-creation flows. Normal ready reads perform no network fetch or repair writes.
- Initially emit complete replacement snapshots using the existing list reconciliation approach. They are easy to
  adopt correctly at current list sizes; C4 can introduce bounded window updates without changing row selection policy.
- `presentation_version` covers durable selected presentation only. Do not interpret it as a revision of the entire
  engine, directory or existing row fields. Subscription generation/sequence orders **all** emitted snapshots; equal
  presentation revisions must never discard a later read-state, delivery, pin, mute or lifecycle update.
- `Cached` means valid local evidence, not network freshness. `LastKnown` means valid same-subject display retained
  while a known newer profile is pending. `Fallback` means the selection lacks usable identity/profile evidence.
  An accepted profile that removes a name/avatar clears that value and recomputes fallback; a failed refresh alone
  does not erase valid cached data.
  Peer replacement clears invalid old-subject values; no public `Invalidated` state may carry the old person's display
  as the current peer. Never downgrade a valid cache merely because the process restarted or network is offline.
- Peer fallbacks reuse MDK's deterministic pseudonym helper with a canonical normalized identity. Group fallback uses
  `LocalizedFallback(UnnamedGroup, member_count)`; unavailable/unknown roster uses `UnavailableConversation`.
  A self-only unnamed group uses the group fallback, not an invented peer or new notes-to-self feature.
- Titles and profile URLs pass through shared bounded validation/sanitization. Empty/invalid display-name candidates
  fall through to usable profile name, then deterministic fallback. Client localization renders the typed fallback;
  it never decides whether this is a DM/group to choose a title.
- Avatar selection is independent of title selection. `SelectedAvatar` tells the image loader what source to use;
  no client peer/group inference is needed. Reuse existing validated group image material and loaders in this slice.
  Descriptor shape must accommodate C7's additive local-access API without embedding image bytes into list rows.
- The C API gains new whole-result types/functions and matching deep-free/subscription lifetime functions. Keep
  existing exported struct layouts/functions unchanged. Do not add a peer-only sidecar call. Rust/Swift/Kotlin also
  retain existing entry points during migration. New record/enum debug output must follow existing privacy rules.

### Selection matrix

| Local evidence | Title | Avatar without an explicit group image |
| --- | --- | --- |
| Nonempty explicit group name, any size | Group name | Peer avatar only if roster proves self + one peer; otherwise group placeholder |
| Empty name, exactly self + one peer | Usable peer profile name, otherwise deterministic peer pseudonym | Usable peer image, otherwise person placeholder |
| Empty name, known roster other than self + one peer | Localizable group fallback with member count | Group placeholder |
| Empty name, unknown/quarantined roster | Localizable unavailable-conversation fallback | Unknown placeholder |
| Same peer, newer profile awaiting apply | Last usable same-peer selection | Last usable same-peer selection |
| Peer replaced or roster no longer qualifies | Recompute against new evidence; never carry old peer as current | Recompute or placeholder; never reuse old peer cache key |

An explicit group image wins in every row independently. Terminal lifecycle flags and command restrictions remain
unchanged; a screen record does not authorize a mutation. Current projected membership determines peer eligibility.

### Names and historical data

Verified current iOS/Android direct creation passes an empty name. Keep generated display derived, never written back
into the group profile. Treat nonempty shared names as explicit unless reliable producer/history evidence establishes
a generated default. Preserve ambiguous historical names, including a deliberately chosen name equal to a fallback.
Do not guess from text or rewrite signed group data during a local projection migration. Add support for a proven
historical generated-name format only with a fixture and participant-consistent provenance; no such format is required
by the verified current creation paths.

### Account projection

Extend existing account storage, not a parallel chat database:

- A versioned selected-presentation value on `chat_list_rows`, including dependency identity and applied profile version.
- A dependency index from identity to affected group/role. Named two-person groups may still depend on a peer avatar;
  the existing unnamed-direct lookup index alone is insufficient. Group/profile changes maintain this index atomically.
- Projection metadata containing a random store epoch, monotonic presentation revision, format/backfill progress and
  directory catch-up checkpoint. Reset/replacement changes the epoch; ordinary restart does not. Never use wall-clock
  time or remote profile timestamps as the local revision.

No new full-profile account table is needed for this slice. Read selected accepted directory evidence in bounded local
batches before the account write, then persist the selected row values. Readiness and dependency changes must be
revalidated under the account transaction; never combine a newly selected peer ID with previously cached peer data.

Account-local group/roster mutations commit either the new valid presentation or safe fallback plus durable dirty
state with the changed source. Name/image changes and row creation use the same resolver. Existing source/timeline/
chat-row transactions introduced by #1737 must be retained. Coalesce work per affected group; repair must not scan
message history, rebuild every account group, or introduce a second send/operation queue.

Advance the presentation revision only when selected presentation/validity/dependencies change, in the same account
transaction as those changes. Profile-only writes must not change activity clocks, unread/read intent, pins or archive.
Account deletion removes projection, dependencies and checkpoint through the existing account-store lifecycle.

### Durable cross-store profile propagation

Use a coalesced latest-change index, not an unbounded event replay log:

1. In the shared directory transaction, persist the accepted profile and a local presentation revision for that
   identity. Keep one latest revision/change row per identity, with a shared-store epoch and monotonic counter.
   Only actual profile/presentation changes advance it; duplicate relay/cache observations do not.
2. Put this bookkeeping at the shared storage write boundary so normal saves and legacy directory import both
   participate. Preserve existing directory winner/equal-timestamp rules. The revision records an accepted local
   change; it never makes an older remote profile authoritative.
3. Each account tracks the processed shared-store epoch/revision and any current identity/group batch progress.
   Query changes in revision order up to a captured high-water mark in a short shared-store read transaction.
   Capture the revision and corresponding accepted profile together. Release that transaction before taking the
   account write lock; do not nest shared/account locks or claim distributed atomicity.
4. Resolve current account dependencies, refresh at most 50 affected rows per account transaction, and commit selected
   values, revision and resume progress together. Process at most 50 changed identities per scheduling batch, yielding
   between batches. An account watermark advances only after the covered identities' dependencies have been handled.
5. If a profile changes again during processing, newer work supersedes old work and restarts that identity's dependency
   pass as needed. A coalesced row moving above a captured high-water mark is handled by the next pass; it is not lost.
   Compare applied source revisions so a delayed older batch cannot overwrite a newer application.
6. Creating a new dependency hydrates its latest local profile regardless of the catch-up watermark. This covers
   profiles cached before chat creation and groups created while another identity batch is in progress.
7. Shared-store replacement/epoch mismatch triggers bounded reconciliation of account dependencies. Profile removal
   or invalidation must record a change/tombstone while dependent accounts need it; never silently drop evidence that
   stale display must be cleared. Delete/coalesce tombstones only after proving no dependent account needs them or
   forcing epoch-based reconciliation. The change index holds no private account/group lists or media secrets.

The runtime resumes pending work while running, including with no screen subscribers. A suspended/stopped or inaccessible
account store retains its checkpoint and catches up on reopen. Idle processing does no full directory/group scan.
Use the existing maintenance scheduling/lifecycle rather than create another runtime framework. Poll a small shared
revision/checkpoint through that scheduler as recovery for a missed wakeup; events keep the normal path prompt.

### Snapshot and notification handoff

- Attach the runtime event receiver before reading the initial local snapshot. Return the snapshot and subscription
  together. Every emitted update is a newly read complete result, not a stale raw-event payload patched by the client.
- Publish a presentation invalidation only after the account commit. Runtime maintenance compares the committed
  presentation revision with its last-notified revision, so commit-before-broadcast interruption recovers even if no
  later message arrives. Runtime restart also forces an initial comparison. Share this mechanism per account, not
  one timer per subscriber; normal events should not wait for the fallback maintenance tick.
- Read existing row fields and persisted selected presentation from one account read transaction where possible;
  revalidate the presentation revision for any enrichment that must occur outside it. Retry/reconcile inside MDK
  rather than returning a new peer identity with an old presentation. Same-subject last-known display is permitted.
- Each subscription owns a fresh generation and monotonic sequence; only its current generation may update the UI.
  On lag/overflow, rebuild from current local state and emit a replacement. Cancellation, account switch and shutdown
  terminate the handle using existing lifecycle conventions. No durable public change stream is promised.
- Preserve existing event routes for non-presentation row fields, including mute-expiry refresh. The new presentation
  token supplements those routes; it does not redefine their authority or silently claim cross-field atomicity.

## Upgrade and rollout

Use new numbered account/shared migrations after the current highest migrations; do not reserve a number now or
change workspace versions as feature work. Schema migrations add representations and markers transactionally; seed
existing directory rows with revision zero. First account preparation hydrates its dependencies directly, so it does
not need an all-directory rewrite to manufacture historical revisions.

Backfill selected rows in bounded batches of 50 with a durable cursor and version. Existing legacy APIs remain usable.
The new open operation can await local preparation asynchronously; never block the host UI thread or wait on network.
An upgrade that cannot complete reports typed preparation/storage failure, not a false cache miss. Do not advertise
that the first upgrade has the same latency as a ready reopen. Measure it separately. For this first, existing-cardinality
list API, completion may cover the account's list; C4 owns bounded first-page readiness for the final screen contract.

Interrupted migrations roll back; interrupted backfill resumes idempotently. Keep group names, user preferences,
read anchors, draft content, history and media secrets intact. Old rows/profile sources remain available until new
projection readiness is durable. Rollback means old **APIs on the upgraded runtime** remain usable; older released
binaries opening a newer schema are subject to existing compatibility policy, not a promised binary downgrade.

## Merge and adoption sequence

| Unit | Repository / deliverable | Depends on | Independently reviewable result |
| --- | --- | --- | --- |
| P1 | MDK: pure selection policy, selected-row storage/dependency index, account migration and resumable backfill | This plan | Tested internal durable representation; existing client behavior remains usable |
| P2 | MDK: shared profile revision index, account catch-up, all source mutation hooks and crash recovery | P1 | Presentation stays correct without subscribers; targeted work and recovery tests pass |
| P3 | MDK: additive Rust/UniFFI/C presented-row reads and snapshot subscription | P1 + P2 | Complete supported native capability; binding parity, ownership and upgrade/reopen checks pass |
| P4a | Android: adopt selected title/avatar in chat rows and creation/rebind paths | Released P3 artifact | Removes display-only roster/profile fallback work; verify Android #2149 and before/after fixture |
| P4b | iOS: adopt the same contract in list and creation/rebind paths | Released P3 artifact | Removes display-only selection/enrichment; verify offline first frame and profile updates |

P1/P2 may merge as internal foundations; they are not separate user-facing launches. Ship the public capability only
when P3's contract is complete. P4a/P4b can release independently. Keep commits/PRs focused on these boundaries; split
an unusually large migration/test patch further without publishing an incomplete API. C1 does not require a new
configuration/feature-flag framework. Later C4-C8 work can reuse the published presentation value and resolver.

### Consumer accounting

- Android `ChatsController` member/presentation caches also support folders, shared groups and profile actions. Remove
  their dependency for row title/avatar rendering; preserve other callers. Do not delete the entire cache in P4a.
- iOS `ChatsListViewModel` uses membership-page reads/fallbacks for enrichment; `DirectChatStarter` seeds a created row.
  Adopt the new whole-row result at both paths, preserving non-display member operations.
- Conversation view models currently combine group/timeline subscriptions and snapshots. Their replacement is C5,
  not an extra requirement on P4 list adoption. Lower-level CLI/agent/raw-event queries stay supported.
- C mirrors remain additive and deeply freed; test embedded legacy fields plus new descriptors exactly once. Swift/
  Kotlin wrappers and generated artifacts must come from the same Rust contract. No host stores a second durable
  selected-presentation cache to make the new API work.

## Acceptance matrix

| Scenario | Required proof |
| --- | --- |
| Profile cached before chat creation; restart/offline | First presented result has the selected peer data without network or per-row client profile/roster lookup |
| Named/unnamed two-person, larger, self-only and unknown roster | Selection matrix passes; title/avatar precedence independent; typed fallbacks deterministic |
| Empty display name, invalid profile/avatar, custom name equal to generated display | Safe candidate fallback; no guessing or overwriting explicit group names |
| Peer A replaced by B; two-person becomes larger or roster invalidated | Same transaction removes A as current peer; safe B/group fallback while hydration proceeds |
| New/duplicate/stale profile, profile-only change | Correct winner; no duplicate generation/write amplification; stable activity/read/archive state |
| Shared commit before wakeup; account row commit before broadcast; restart mid-batch | Replay/resume converges with no active screen and without another incoming message |
| Account created or new dependency added after watermark; coalesced newer profile during pass | No missed hydration, no old-batch rollback, no skipped affected group |
| Account/shared store reset; profile invalidation; account deletion | Epoch/reset and removal recovery; isolation; no stale peer revival |
| Subscribe/read race, lag, repeated snapshot, equal presentation version with changed unread state | Complete newest snapshot, monotonic per-handle sequence; no lost non-presentation update |
| Populated migration/backfill failure, cancellation and reopen | Existing intent/data preserved; typed readiness; no partial-cache success or UI-thread blocking |
| Swift/Kotlin/C API and creation paths | One whole result supports selected display; enum cases covered; C layout and deep-free safety |
| Increasing unrelated history/groups/directory entries | Indexed affected-row work; ready reads have no history scans, network work or projection writes |

Use meaningful existing regression foundations and add targeted storage, app-runtime and binding tests for new seams.
Run `just fast-ci` before pushing, touched-crate tests, and C parity/header/smoke/alloc-audit gates when P3 changes them.
Native adoption follows each client's own screenshot/device-test requirements. The plan itself changes documentation
only; no new runtime tests have been executed for it.

For performance, preserve the recorded Pixel 9a / 20×40 fixture and artifact provenance. Measure native presented-list
read, profile commit-to-update, first upgrade and ready offline reopen separately from whole-app startup. Report p50/p95
and workload/query work; use the existing whole-client runs for comparable before/after evidence. Provisional parent
latency targets are not automatic release gates. Do not expand this slice into unrelated startup or packaging repairs.

## Evidence and remaining scope

MDK baseline: `788eaf0852442cb31311fd72d9dd6ea91c2642a6`; reviewed as a focused diff from the local
`91221c86585387d0d029466728c86e765ed14150` inventory. Refresh touched source and migrations when implementation starts.
Relevant owners: `storage-sqlite/src/chat_list.rs`, `account_projection.rs`, `shared.rs`; `marmot-app/src/directory/methods.rs`,
`groups.rs`, `client/projection.rs`, `lib.rs`, `runtime/subscriptions.rs`; UniFFI chat-list conversions and commands;
C types/commands/subscriptions. Shared directory writes occur in normal saves **and** legacy import.

Native evidence: iOS `190ac82c7f069cf49e05bb27a6552547a20c9ead`, Android `3af7335263c9134503f04b061578a80a9b0ee346`.
This is sufficient consumer accounting for list presentation; it is not a complete inventory for later conversation,
media, or non-chat binding removal. Those children must finish their own detailed contracts before implementation.

</details>

<details>
<summary>Historical issue evidence — superseded scope, retained for provenance</summary>

The original issue below was based on an earlier MDK commit. Its peer-only field/freshness prescription and standalone-parent statement are superseded by the implementation plan above; its cold-start requirement and downstream link remain valid.

## Problem

White Noise clients cannot render an unnamed direct conversation from one authoritative durable chat-list row after a cold start. MDK persists the conversation kind and group presentation, but the row does not expose the direct peer’s last-known local presentation. Android must therefore rebuild the title/avatar from roster and profile lookups after binding; when that state is not yet available, a known conversation can render as `Unknown` or a generic placeholder.

This is the upstream prerequisite for [White Noise Android #2149](https://github.com/marmot-protocol/whitenoise-android/issues/2149). The Android issue remains responsible for first-frame consumption and UI behavior. This issue owns only the durable MDK projection and binding contract.

## Source-confirmed gap

Reviewed at MDK `5e597ea222b104537d9554c42d4ad5fadf3d418a`.

- `ChatListRow` persists group-level `title`, `group_name`, `avatar_url`, and avatar material plus `conversation_kind`, but has no direct-peer presentation record ([row contract](https://github.com/marmot-protocol/mdk/blob/5e597ea222b104537d9554c42d4ad5fadf3d418a/crates/storage-sqlite/src/chat_list.rs#L284-L355)).
- Row decoding joins the durable account-group projection for membership count and derives `Direct`, but returns only the existing group presentation fields ([row query and decode](https://github.com/marmot-protocol/mdk/blob/5e597ea222b104537d9554c42d4ad5fadf3d418a/crates/storage-sqlite/src/chat_list.rs#L2025-L2053), [conversation classification](https://github.com/marmot-protocol/mdk/blob/5e597ea222b104537d9554c42d4ad5fadf3d418a/crates/storage-sqlite/src/chat_list.rs#L2097-L2152)).
- `ChatListRowFfi` forwards the same group presentation and conversation kind without a direct-peer projection, so Kotlin and Swift cannot obtain a cold-start-ready peer title/avatar from the row itself ([binding record](https://github.com/marmot-protocol/mdk/blob/5e597ea222b104537d9554c42d4ad5fadf3d418a/crates/marmot-uniffi/src/conversions/chat_list.rs#L180-L248), [conversion](https://github.com/marmot-protocol/mdk/blob/5e597ea222b104537d9554c42d4ad5fadf3d418a/crates/marmot-uniffi/src/conversions/chat_list.rs#L250-L279)).
- Existing storage coverage proves that conversation kind survives roster projection changes, but does not cover durable direct-peer presentation ([current roster test](https://github.com/marmot-protocol/mdk/blob/5e597ea222b104537d9554c42d4ad5fadf3d418a/crates/storage-sqlite/src/chat_list/tests.rs#L454-L515)).

The source audit confirms the missing projection. The user-visible recurrence is verified on Android; this issue does not claim an independent runtime reproduction inside MDK.

## Required behavior

For an unnamed conversation classified as `Direct`, MDK exposes a minimal, account-scoped, durable direct-peer presentation with the chat-list row:

- stable local peer identity sufficient for the host to correlate the row without logging it;
- last-known sanitized display name and avatar reference/material when available;
- provenance/freshness state that distinguishes absent, current, last-known, and invalidated presentation;
- atomic update with the authoritative roster/profile projection so a row never combines a new peer identity with another peer’s stale presentation.

The projection remains inside the existing per-account encrypted SQLite source of truth. Hosts must not need a second persistent cache. It is local presentation state, not a protocol field and not transmitted to other members.

## Lifecycle and privacy contract

- Creating or hydrating a direct conversation populates the projection when local profile evidence is available.
- A profile update refreshes the same peer’s projection without changing conversation activity ordering.
- A process restart exposes the last-known value before asynchronous network/profile refresh.
- A roster transition that no longer identifies exactly one remote peer clears or invalidates the direct-peer projection; stale data must not cross to a new peer, account, or reclassified group.
- Account deletion and account-database reset remove the projection with the existing account-scoped database lifecycle.
- Diagnostics may report only aggregate state classes and transition reasons. They must not log member IDs, display names, avatar URLs, decoded avatar bytes, group IDs, or account IDs.

## Acceptance criteria

- [ ] Add a versioned durable direct-peer presentation representation owned by the account-scoped chat-list/storage projection; document its update and invalidation rules.
- [ ] Expose the representation through the `marmot-app` chat-list API and `ChatListRowFfi` without requiring host-side persistence or extra per-row profile calls for the first frame.
- [ ] Preserve group presentation behavior for named groups and do not synthesize direct-peer presentation when the durable roster does not identify exactly one remote member.
- [ ] Add `storage-sqlite` coverage named `refresh_chat_list_row_projects_direct_peer_presentation` for create, profile refresh, restart/reopen, peer change, Direct→Group transition, absent/quarantined profile data, and account isolation.
- [ ] Add `marmot-uniffi` coverage named `direct_peer_presentation_round_trips_to_ffi` for present, last-known, absent, and invalidated states.
- [ ] Verify the projection does not affect `activity_sort_at` and that existing chat-list subscription updates publish a changed row when presentation changes.

## Relationships and scope

This issue is standalone MDK work and blocks [White Noise Android #2149](https://github.com/marmot-protocol/whitenoise-android/issues/2149). It has no native parent: the change is one independently landable storage/API/binding contract, and no inspected MDK tracker has a completion rule that requires it.

Out of scope: Android rendering/rebind behavior, an Android Room/DataStore/SharedPreferences cache, protocol or wire-format changes, remote profile fetching policy, and avatar download/cache policy.

</details>

mubarakcoded:
## Summary

- persist account-scoped direct-peer display names and profile pictures in chat-list rows, including migration and backfill behavior
- update the projection monotonically from profile ingestion and invalidate it atomically when roster membership changes
- expose the persisted presentation and explicit refresh path through marmot-app, subscriptions, the C API, and UniFFI

## Verification

- just fast-ci
- cargo test -p storage-sqlite direct_peer (7 passed)
- cargo test -p marmot-app direct_peer (4 passed)
- cargo test -p marmot-uniffi direct_peer_presentation_round_trips_to_ffi (1 passed)
- full storage-sqlite suite before the final rebase (464 passed; 4 ignored benchmarks)
- local Android binding integration (178 relevant tests passed; 4/4 packaged ARM64 device tests passed on Pixel)

Closes #1517

## Summary by CodeRabbit

- **New Features**
  - Direct chats now display peer profile information, including name and avatar, with freshness states such as current, last known, or invalidated.
  - Chat lists automatically refresh when peer profiles or conversation rosters change.
  - Chat-list updates can report changes to direct-peer presentation data.
  - Unavailable or invalid profile information is handled safely without disrupting other chat updates.
  - Added direct-peer presentation data to supported native and cross-platform interfaces.

- **Bug Fixes**
  - Prevented stale peer details from appearing after profile or roster changes.
  - Improved unread counts and targeted chat-row refreshes after message updates.
  - Profile data is now sanitized more reliably before display.


Datawav (commented):
## Automated review

One actionable finding (minor).

1. Severity: Minor — trim occurs before control-character filtering

Exact changed-line evidence: `crates/marmot-app/src/directory/records.rs:374-380` adds `sanitize_profile_field`, whose pipeline calls `.trim().chars().filter(|character| !character.is_control())` before collecting the persisted value.

Triggering state: A profile field begins or ends with a control character that shields adjacent whitespace from `trim()`, such as `"\u{0000} Otter"`.

Consequence: Filtering removes the control character but leaves the newly exposed leading or trailing whitespace. This can persist a malformed display name; for `avatar_url`, the leading space can prevent URL loading.

Confidence: High. The behavior follows directly from the changed operation order and applies to both display-name fallback and other profile fields routed through this sanitizer.

Smallest sound correction: Filter control characters first, collect the string, then call `.trim()` on that collected value before the existing empty-value check. Add a regression test covering control-plus-whitespace at both boundaries.

Audit evidence:
- Exact head: `1b9f7d3461d63ea418a03425be354c0547ae5d00`
- Patch coverage is complete: 17 changed files, no missing patches.
- Changed files examined include `crates/marmot-app/src/directory/records.rs`, `crates/storage-sqlite/src/chat_list.rs`, `crates/storage-sqlite/src/migrations.rs`, and `crates/storage-sqlite/src/migrations/0059_direct_peer_presentation.rs`.
- The additive migration is transaction-scoped, preserves existing rows through nullable/defaulted columns, and is appended as schema migration 59. Projection tests exercise persistence, restart hydration, roster invalidation, boundary states, and production subscription behavior.
- Discussion classification: the inline sanitizer report is supported and retained. The walkthrough is descriptive rather than actionable. The suggested unchanged-projection transaction optimization is not retained because the supplied evidence demonstrates no material regression or correctness consequence.

Risk lenses checked: `persistence/migration`, `tests`
- `persistence/migration`: upgrade behavior, transactional partial-failure handling and retry safety, and existing-data preservation were reviewed.
- `tests`: pre-fix sensitivity, boundary and negative cases, and whether tests exercise production paths were reviewed.
- Large-change decomposition and behavioral-test reasoning were also applied across storage, runtime, C, and UniFFI boundaries.

Check evidence:
- Bundle evidence is available and complete; `all_completed=true` and `all_successful=true`.
- Successful checks include Required CI, Rust tests 1/4 through 4/4, Rust check, Rust clippy, Rust format, C bindings smoke, Simulator and vectors, Relay runtime tests, Relay CLI E2E, and Rust audit.
- `automated_device_gate` is null. No physical-device evidence exists for this head; conclusions rest on static review and the supplied CI evidence.

Datawav (commented):
## Automated review

APPROVE — five minor, non-blocking actionable findings.

1. Minor — Projection events can be silently dropped
   Exact changed-line evidence: `crates/marmot-app/src/runtime/subscriptions.rs`, subscription-loop hunk beginning at `+1321`, adds:
   `let current_presentation = match blocking_app_task(...).await { Ok(presentation) => presentation, Err(_) => continue, };`
   Trigger: a transient account-storage/SQLite or blocking-task error while processing an ordinary chat-list projection event.
   Consequence: the already-computed unread count, preview, or delivery-state update is discarded, leaving the host row stale until another event for that group.
   Confidence: High.
   Smallest correction: on lookup error, emit the captured row unchanged rather than `continue`.

2. Minor — Read-time hydration advances the row maintenance timestamp
   Exact changed-line evidence: `crates/marmot-app/src/lib.rs` adds `hydrate_chat_list_rows` / `hydrate_chat_list_row` calls into `hydrate_direct_peer_presentations`; storage projection writers assign `updated_at = ?`. `crates/marmot-uniffi/src/commands/chat_list.rs` correspondingly changes exact equality to `assert!(right.updated_at >= left.updated_at, ...)`.
   Trigger: reading a chat list containing a direct-peer presentation that is not current.
   Consequence: a nominal read mutates durable state and can appear to hosts as an unrelated row update; the relaxed assertion can also hide unintended timestamp churn.
   Confidence: Medium-high.
   Smallest correction: preserve `updated_at` for presentation-only maintenance, or use a dedicated presentation timestamp and assert its exact expected change.

3. Minor — Completeness check hardcodes schema version `1`
   Exact changed-line evidence: `crates/storage-sqlite/src/chat_list.rs`, `chat_list_projection_complete_tx`, adds SQL containing `row.direct_peer_presentation_version != 1`, while the new writers and readers use `DIRECT_PEER_PRESENTATION_SCHEMA_VERSION`.
   Trigger: a future increment of `DIRECT_PEER_PRESENTATION_SCHEMA_VERSION`.
   Consequence: rows written with the new version remain “incomplete,” causing `ensure_chat_list_rows` to rebuild the entire projection on every warm open.
   Confidence: High.
   Smallest correction: bind `i64::from(DIRECT_PEER_PRESENTATION_SCHEMA_VERSION)` as a query parameter instead of embedding `1`.

4. Minor — A successful profile save can be converted into an error
   Exact changed-line evidence: `crates/marmot-app/src/directory/methods.rs`, hunk `@@ -1119,18 +1119,46`, adds:
   `self.save_directory_entry(&entry)?;`
   followed by
   `self.directory_entry_for_account_id(account_id_hex)?.and_then(|entry| entry.profile).ok_or_else(|| AppError::MissingDirectoryEntry(...))?;`
   Trigger: the merged post-save accessor selects no entry or a newer shared entry without a profile.
   Consequence: callers receive an error after the supplied profile has already been persisted, contradicting the intended rule that derived projection failure must not reject an accepted directory profile.
   Confidence: High.
   Smallest correction: if read-back has no profile, project from the profile just saved and retain the successful write result.

5. Minor — One invalidated direct row escalates to a full projection rebuild
   Exact changed-line evidence: `crates/storage-sqlite/src/chat_list.rs` adds roster invalidation that stores `direct_peer_presentation_state = 'invalidated'`, while `chat_list_projection_complete_tx` rejects direct rows whose state is not in `('absent', 'current', 'last_known')`; `ensure_chat_list_rows` then calls `rebuild_all_chat_list_rows_tx`.
   Trigger: a direct-conversation roster replacement followed by a warm open before targeted row refresh.
   Consequence: one conversation’s expected invalidation rebuilds every chat-list row for the account.
   Confidence: High.
   Smallest correction: repair that group’s row during invalidation or route invalidated rows through a targeted refresh.

Audit evidence:
- Exact head: `f9eb5d9023a0ab0bbd01618dd5d470984e22086a`
- Reviewed 17 changed files, including `crates/storage-sqlite/src/chat_list.rs`, `crates/marmot-app/src/directory/methods.rs`, `crates/marmot-app/src/lib.rs`, and `crates/marmot-app/src/runtime/subscriptions.rs`.
- Discussion classification: the control-character trimming comment is stale at this head because the changed implementation trims control characters and whitespace at both original edges and includes a boundary test. The write-transaction nitpick is also stale because hydration performs a read-only preflight before opening a write transaction and includes a lock-contention regression test. The generated summary/rate-limit comment is non-actionable.
- The PR-body Pixel claim is not evidence for this exact head. The automated device gate is `capability_checked_not_installable`, is bound to `1b9f7d3461d63ea418a03425be354c0547ae5d00`, and expressly forbids a physical PASS. No physical-device evidence exists for this head; conclusions therefore rest on static review and supplied CI evidence.

Risk lenses checked: `persistence/migration`, `tests`.
- `persistence/migration`: examined upgrade/backfill and rollback implications, partial failure and idempotent convergence, roster-change invalidation, and preservation of peer-bound data.
- `tests`: checked pre-fix sensitivity, boundary/negative cases, production-path coverage, and the weakened timestamp assertion.
- Large-change decomposition and behavioral-test reasoning were also applied across migration 0060, storage projection, app hydration, subscriptions, C bindings, and UniFFI.

Check evidence: the bundle records complete check evidence with Required CI, all four Rust test shards, Rust check, clippy, format, audit, C bindings smoke, simulator/vectors, relay tests, and related harness jobs completed successfully; the explicitly skipped release/proof jobs do not supply correctness evidence.

dannym-arx (dismissed):

Datawav (commented):
## Automated review

Two minor actionable findings.

1. Minor — transient lookup errors can erase a persisted peer presentation until another event arrives

Evidence: In crates/marmot-app/src/runtime/subscriptions.rs, changed hunk @@ -1481,7 +1481,7, the former `Err(_) => continue` behavior is replaced by an error arm that sets `row.direct_peer_presentation = None` and emits `Some(row)`.

Triggering state: A Direct chat-list projection update coincides with a transient failure in `account_storage(..)` or `direct_peer_presentations_for_group_ids(..)`.

Consequence: The subscriber replaces a valid displayed name/avatar with `None`, even though durable storage may still contain the presentation. The changed path schedules no deterministic reread, so restoration depends on an unrelated later projection event.

Confidence: High.

Smallest sound correction: Keep the privacy-safe presentation-less emission, then schedule a bounded targeted reread for the affected group through the existing direct-presentation refresh path. Extend the regression test to model a recoverable lookup error and verify restoration; the current permanent `close()` case establishes only the initial error emission.

2. Minor — targeted repair duplicates work when a full rebuild is required

Evidence: In crates/storage-sqlite/src/chat_list.rs, changed hunk @@ -925,6 +925,11, `repair_invalidated_direct_peer_rows_tx(...)` is called immediately before `if !chat_list_projection_complete_tx(...) { rebuild_all_chat_list_rows_tx(...) }`.

Triggering state: Opening an account whose projection is incomplete, including a projection-version mismatch.

Consequence: Invalidated exact-Direct rows are rebuilt by the targeted repair and then rebuilt again by the full pass in the same transaction. Output remains correct and retry-safe, but cold-open write and query work is duplicated.

Confidence: High.

Smallest sound correction: Check completeness first: run the full rebuild when incomplete, and run `repair_invalidated_direct_peer_rows_tx` only in the complete-projection branch.

Audit evidence:
Exact head: a9728f36d8f108cbaae3459e8397dc1b8365c1b2
Changed files examined: crates/marmot-app/src/runtime/subscriptions.rs, crates/marmot-app/src/runtime/tests.rs, crates/marmot-uniffi/src/commands/chat_list.rs, crates/storage-sqlite/src/chat_list.rs, and crates/storage-sqlite/src/chat_list/tests.rs. Bundle coverage is complete with no missing patch files.

Risk lenses checked:
- persistence/migration — upgrade and rollback from real prior state; partial failure and idempotent retry; data preservation. The repair is transactional and idempotent, preserves unrelated and non-Direct rows, and the added sentinel test discriminates targeted from global rewrites; finding 2 identifies redundant work on the incomplete-projection path.
- tests — assertions fail before the fix; boundary and negative cases; test exercises production behavior. The storage test covers exact-Direct repair and an unrelated-row sentinel. The runtime test reaches the new error arm but does not establish recovery from a transient error, as covered by finding 1.

Discussion classification:
- The summary comment is informational.
- The profile-field sanitization and C-ABI layout claims target files absent from this complete incremental patch and lack bounded source snapshots; they are unsupported here and discarded.
- The hydration write-transaction claim is stale at this head and discarded.
- The one-second timeout suggestion is speculative test-hardening rather than a demonstrated defect and is not retained separately.
- The permanent-storage-close test limitation is incorporated into finding 1 rather than duplicated.

Check evidence:
The bundle reports `available=true`, `complete=true`, `all_completed=true`, and `all_successful=true`. Required CI, Rust tests 1/4 through 4/4, Rust check, clippy, format, audit, C bindings smoke, relay runtime tests, CLI E2E, simulator/vectors, and Tamarin proofs completed successfully. The release-publication job was skipped.

Device evidence:
`automated_device_gate.status` is `capability_checked_not_installable` and its record is tied to the prior SHA f9eb5d9023a0ab0bbd01618dd5d470984e22086a. No physical-device evidence exists for this head; conclusions therefore rest on static review and the supplied CI evidence.

Datawav (commented):
## Automated review

Changes requested: four actionable findings.

1. Major — Chat-list reads now fail on recoverable derived-presentation writes
   Evidence: In changed `crates/marmot-app/src/lib.rs`, `hydrate_chat_list_rows` adds `self.hydrate_direct_peer_presentations(account, rows)?;`. The added hydration path propagates both `cached_identity_projections_for_account_ids(page)?` and `storage.hydrate_direct_peer_profiles(&account.account_id_hex, &profiles, &unavailable_peer_ids)?`; the latter may enter `with_transaction`/`BEGIN IMMEDIATE`.
   Trigger: write-lock contention or a transient directory/profile-storage error during `chat_list()` or `chat_list_row()`.
   Consequence: a derived cache refresh aborts the entire previously read-oriented chat-list operation, even though existing rows remain renderable.
   Confidence: High.
   Smallest correction: warn and skip failed directory lookup/hydration pages, preserving the roster-safe re-read ordering and returning the durable rows already available.

2. Minor — The two projection writers do not converge on one display name
   Evidence: Changed `crates/marmot-app/src/lib.rs` persists `display_name: projection.resolved_name` during read-path hydration, while profile ingestion persists `display_name: directory::records::display_name_for_profile(Some(profile))`; both use the same `profile_created_at`. The new no-op preflight requires timestamp and values to match.
   Trigger: `resolved_name` differs from the profile display-name helper, such as when local fallback resolution applies.
   Consequence: chat-list reads and later profile ingestion alternately rewrite the same projection, churn `updated_at`, and can alternate the displayed name.
   Confidence: High.
   Smallest correction: derive `display_name` through `display_name_for_profile` in both writers.

3. Minor — The ordinary-path retry test depends on winning a 250 ms race
   Evidence: Changed `crates/marmot-app/src/runtime/subscriptions.rs` defines `CHAT_LIST_PRESENTATION_LOOKUP_RETRY_DELAY` as 250 ms. In changed `crates/marmot-app/src/runtime/tests.rs`, the ordinary-path recovery removes the cached storage handle only after receiving the first emission; unlike the wake-path case, it does not await `DIRECT_PEER_WAKE_LOOKUP_FAILURES_FOR_TEST` before the retry.
   Trigger: scheduling delays let the retry execute before the test removes the broken handle.
   Consequence: the retry repeats the lookup error and the test times out nondeterministically under CI contention.
   Confidence: Medium-high.
   Smallest correction: synchronize this branch with the existing test-only failure counter or a test-only notification before repairing the handle.

4. Low — Truncation can reintroduce trailing whitespace after sanitization
   Evidence: Changed `crates/marmot-app/src/directory/records.rs:386` applies `trim_matches(...)` before `.filter(...).take(MAX_PROFILE_FIELD_CHARS).collect()`.
   Trigger: an oversized field whose 4096th retained character is whitespace.
   Consequence: the persisted value violates the helper’s stated no-edge-whitespace invariant.
   Confidence: High.
   Smallest correction: trim the collected, capped string once more and then perform the empty-value check; add this boundary case.

Audit evidence:
- Exact head: 87d29b965202adb746ab3c4e2d94ac7964236159
- Complete patch coverage: 16 files, +3122/-59, with no missing patch files.
- Changed files examined include `crates/marmot-app/src/lib.rs`, `crates/marmot-app/src/runtime/subscriptions.rs`, `crates/marmot-app/src/runtime/tests.rs`, `crates/storage-sqlite/src/chat_list.rs`, and `crates/storage-sqlite/src/migrations.rs`.
- Migration/version rebuilding, roster-transaction invalidation, unknown-state/schema fail-closed handling, unrelated-row preservation, retry/idempotence, and C/UniFFI propagation were examined. No additional actionable migration or data-preservation defect was established.
- Discussion classification: the earlier control-edge trimming, C-layout insertion, unchanged-write transaction, wake-lookup retry, and one-second timeout claims are resolved or stale at this head; wrapper summaries and rate-limit text are non-actionable. The findings above are distinct current-head issues.
- The automated device gate is `capability_checked_not_installable` and references older SHA `a9728f36d8f108cbaae3459e8397dc1b8365c1b2`; therefore no physical-device evidence exists for this head. Conclusions rest on static review and the supplied CI evidence.

Risk lenses checked: persistence/migration; tests.
- Large-change decomposition and behavioral-test coverage were also assessed.

Check evidence:
- Supplied exact-head checks are complete but not all successful.
- `Required CI`: failure.
- `Rust tests (1/4)`: failure.
- `Rust tests (2/4)`, `(3/4)`, and `(4/4)`, `Rust check`, `Rust clippy`, `Rust format`, `Rust audit`, `C bindings smoke`, `Relay CLI E2E`, and `Simulator and vectors`: success.
- The PR body’s pre-rebase/local run counts do not replace exact-head CI evidence.

REQUEST_CHANGES

Datawav (commented):
## Automated review

1 actionable finding.

1. Minor — Post-save directory re-read can turn a successful profile write into an error

   Exact changed-line evidence: crates/marmot-app/src/directory/methods.rs:1135-1139 adds, immediately after the durable save:

       self.save_directory_entry(&entry)?;
       let accepted_profile = self
           .directory_entry_for_account_id(account_id_hex)?
           .and_then(|entry| entry.profile)
           .ok_or_else(|| AppError::MissingDirectoryEntry(account_id_hex.to_owned()))?;

   Triggering state: the newly saved cache entry contains the supplied profile, but directory reconciliation selects a newer shared record without a profile.

   Consequence: remember_directory_profile_with_updates returns MissingDirectoryEntry after the write has already committed; callers that previously received Ok(()) now observe failure for a successful ingestion, and may retry or report incorrect state.

   Confidence: High. The changed function introduces the fallible post-save re-read, while the supplied bounded context establishes that directory_entry_for_account_id reconciles cache and shared records by recency.

   Smallest sound correction: make the derived projection non-authoritative for ingestion. If the reconciled record has no profile, either use profile.clone() as the accepted profile or return Ok(Vec::new()) after preserving the successful write.

Audit evidence:
Exact head: 367487b701d9e7d0cc211863be557af1e0e2449e
Reviewed complete bundle coverage for 16 changed files (+3315/-60), including crates/marmot-app/src/directory/methods.rs, crates/marmot-app/src/runtime/subscriptions.rs, crates/storage-sqlite/src/chat_list.rs, crates/storage-sqlite/src/migrations/0060_direct_peer_presentation.rs, and the C/UniFFI mirrors.

Risk lenses checked: persistence/migration, tests

Persistence/migration review covered upgrade from prior schema, atomic migration defaults, projection rebuild/backfill, rollback implications, partial failure, retry idempotence, roster-change invalidation, schema/state fail-closed behavior, and preservation of existing chat-list rows. Test review covered production entry points, pre-fix sensitivity, migration preservation, retry paths, roster replacement, absent/invalidated states, and boundary/negative cases.

Discussion classification:
- Control-character trimming: stale; the head trims control/whitespace boundaries and re-trims after truncation.
- C chat-row field-offset warning: stale; the new presentation pointer is appended.
- Direct-peer wake retry warning: resolved in this head by the bounded second lookup.
- Unchanged-projection write-transaction warning: resolved by the read-only hydration preflight and regression coverage.
- One-second receive-timeout suggestion: stale; changed tests use five seconds.
- Performance-snapshot offset warning: stale against the final generated layout.
- Remaining review summaries duplicate those claims or contain no independently actionable defect.
- The advisory’s exact-string group-ID removal concern lacks a demonstrated noncanonical producer; the 250 ms lag concern lacks a supported capacity-triggering workload; the UniFFI field-order claim does not provide backward compatibility because generated bindings are checksum-coupled; and missing-directory-projection demotion lacks a demonstrated reachable deletion state. These were discarded as speculative or unsupported.

Check evidence:
The bundle reports check evidence available and complete, with all_completed=true and all_successful=true. Required CI, Rust tests (1/4 through 4/4), Simulator and vectors, Relay CLI E2E, C bindings smoke, Rust format, Rust audit, packaging, and platform build checks completed successfully; designated release/proof jobs were skipped.

The automated device gate is capability_checked_not_installable and is not bound to this exact head. No physical-device evidence exists for this head. Conclusions therefore rest on static exact-head review and the supplied CI evidence.

crates/marmot-c/src/types/chat_list.rs
```diff
@@ -167,6 +190,9 @@ c_mirror! {
         /// `has_leave_requested_at_ms`.
         copy leave_request_pending: bool,
         opt_copy has_leave_requested_at_ms/leave_requested_at_ms: u64,
+        /// Appended to preserve the offsets of every pre-existing field in
+        /// this stable C mirror.
+        opt_rec direct_peer_presentation: MarmotDirectPeerPresentation,
```
erskingardner: [P1] Preserve the row-array ABI when adding presentation metadata

Appending this pointer still increases `sizeof(MarmotChatListRow)`. The existing `marmot_chat_list` and chat-list subscription snapshot APIs return contiguous arrays of these structs (`c_list!` and `MarmotChatListSubscriptionUpdate::Snapshot`), with no stride or layout-version negotiation. A C/raw-FFI client compiled against the previous header and loaded with the updated shared library therefore computes `items[1]` using the old size: its second `group_id_hex` lands on the first row's new presentation pointer, so lists with two or more rows produce invalid field reads and can crash. Preserving the offsets within the first row does not preserve this stable ABI. Keep the existing row layout and expose the new metadata separately, or introduce versioned row/list/subscription APIs. Add compatibility coverage that uses the previous header to consume at least two rows from the new library; a smoke test compiled against the new header cannot catch this.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-05T06:08:29Z` |
| Commit reviewed | `367487b701d9e7d0cc211863be557af1e0e2449e` |
| Model | `gpt-6-astra` |
| Reasoning level | `medium` |

One blocking C ABI compatibility issue: appending the presentation pointer preserves existing member offsets but changes the stride of the contiguous row arrays returned by existing APIs. Details inline.

Reviewed the migration/backfill, exact-roster invalidation, profile projection and hydration, subscription delivery/retry, and C/UniFFI conversions against issue #1517. Current-head Required CI, all four Rust test shards, C bindings smoke, and simulator/vector checks pass. Validation here was remote source review plus existing CI; no local build or tests were run.

The existing automated claim that reconciliation selects a newer profile-less shared record and discards the saved profile is not supported by this head: `select_newer_directory_entry` merges profile components independently. The old field-insertion finding is resolved, but the array-stride issue below remains.

Datawav (commented):
## Automated review

I cannot retrieve the exact lane identifiers or the automated_device_gate status — the bundle content was elided from my view and the tools block re-reads. Fabricating them would violate the grounding requirement. I must report this honestly rather than emit an ungrounded audit.

I can't produce the final exact-head audit for marmot-protocol/mdk #1683 at head 367487b701d9e7d0cc211863be557af1e0e2449e. The bundle at /opt/data/.hermes/state/marmot_pr_audit_artifacts/marmot-protocol_mdk/1683/367487b7.../bundle.json was read, but its content was elided from my working context, and every subsequent attempt to re-read it (read_file with default and offset regions) is being blocked by the runtime's dedup guard ("content from your earlier read is still current"). Content search across /opt/data cannot target that one file precisely enough to recover the exact risk_profile.lanes identifier strings or the automated_device_gate.status value.

Because the audit must name every routed lane verbatim and ground the device-gate statement in the bundle's actual status, producing the report now would require inventing those fields — which I won't do.

To proceed I need one of:
- A fresh session (or this one reset) so the bundle can be re-read, or
- The bundle's risk_profile.lanes list, automated_device_gate.status, check rollup, head SHA, and changed file list pasted here, or
- A copy of bundle.json and verdict.json at a new path (e.g. /opt/data/tmp/mdk-1683-bundle.json) that I can read without the dedup block.

Once I have grounded access to those fields I'll emit the audit in the required format (lead with outcome, ≤5 findings, "Audit evidence:" / "Risk lenses checked:" / "Check evidence:", strict device-gate wording, ≤6000 chars).

Datawav (changes requested):
The C ABI regression flagged at this head is still present: `MarmotChatListRow` in `crates/marmot-c/include/marmot.h` gains `struct MarmotDirectPeerPresentation *direct_peer_presentation` as an appended member, which changes `sizeof(MarmotChatListRow)`. `marmot_chat_list` and the chat-list subscription snapshot return contiguous arrays of these rows with no stride or layout-version negotiation, so a C client compiled against the previous header and run against the new shared library strides rows by the old size: from the second row onward every field read lands on the wrong bytes (the first row's new pointer is read as the second row's `group_id_hex`), producing corrupt reads and potential crashes in lists with two or more rows. Offset preservation within row zero does not preserve the array ABI. Please keep the existing row layout and expose presentation through a separate accessor, or add versioned row/list/subscription APIs — plus a compatibility test that compiles a consumer against the *previous* header and reads at least two rows from the new library (a new-header smoke test cannot catch stride drift).

The rest of the change verifies cleanly at 367487b7: the storage migration 0060 and backfill are account-scoped and idempotent, roster-change invalidation is atomic with the projection update, profile ingestion updates the projection monotonically, and the UniFFI conversion round-trips. The earlier claim that reconciliation discards a saved profile for a newer profile-less record is not supported at this head — `select_newer_directory_entry` merges profile components independently. Current-head Required CI is green.

Reviewed by Hermes (openai-codex/gpt-5.6-sol via review profile) — independent source review of the migration, projection, subscription delivery, and both FFI surfaces; existing CI evidence used for test state.

crates/marmot-c/src/types/chat_list.rs
mubarakcoded: Fixed in b7af025a. `MarmotChatListRow` is back to its exact legacy layout and direct-peer metadata is exposed through the additive `marmot_chat_list_direct_peer_presentation` sidecar accessor (with a dedicated free function), so existing array and snapshot consumers retain their original stride. I also added a literal pre-change C header fixture that is compiled at test time and uses `rows[1].group_id_hex` against a two-row allocation from the current library, plus shared/static C smoke coverage. `cargo test -p marmot-c --features alloc-audit`, warnings-as-errors Clippy, header regeneration/parity, and the native C smoke suite all pass.

Datawav (approved):

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T11:37:54Z` |
| Commit reviewed | `b7af025a28acf485fa25377c7524771741c1b822` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

No remaining actionable findings at this head. The earlier C ABI blocker is fixed: `MarmotChatListRow` matches the legacy header exactly, preserving field offsets and array stride. Direct-peer metadata is exposed through an additive accessor with its own deep-free function. CI explicitly passes the test that compiles a previous-header C consumer and reads the second row from a current-library allocation.

Reviewed the complete PR diff and surrounding source against #1517: migration 0060 and backfill, exact-roster invalidation, monotonic profile projection, cold-start hydration, subscription revalidation and bounded retries, and C/UniFFI conversion and ownership. The earlier sanitizer and retry findings are addressed; the profile-less reconciliation claim is unsupported by the component-wise merge implementation.

Validation: exact-head Required CI, all four Rust test shards, C bindings smoke, and simulator/vector checks pass. All four inline review threads are resolved. This review used remote source and existing CI; no local builds or device tests were run. Android first-frame consumption and device acceptance remain the downstream scope of whitenoise-android#2149.

crates/marmot-app/src/lib.rs
```diff
@@ -4729,6 +4765,198 @@ impl MarmotApp {
...
+                    row.direct_peer_presentation = None;
+                }
+                return Ok(());
+            }
+        };
+        let requested = peer_ids.into_iter().collect::<Vec<_>>();
+        for page in requested.chunks(MAX_CACHED_IDENTITY_PAGE_SIZE) {
+            let mut profiles = Vec::new();
+            let mut unavailable_peer_ids = Vec::new();
+            let projections = match self.cached_identity_projections_for_account_ids(page) {
```
erskingardner, line 4814: **Read paths now do per-read directory lookups and can write, which partly defeats the goal of the durable projection.**

`hydrate_direct_peer_presentations` runs on every `chat_list_rows`, `chat_list_row`, `refresh_chat_list_row`, `initialize_chat_read_state`, `set_chat_manually_unread`, etc. For each call it:

1. collects every distinct peer id from the rows,
2. calls `cached_identity_projections_for_account_ids` per 100-peer page — which itself opens `directory_caches()`, `shared_storage()` and `local_account_labels_by_id()` and then walks every requested id,
3. runs `hydrate_direct_peer_profiles` (read preflight, then a write transaction when anything differs),
4. re-reads `direct_peer_presentations_for_group_ids` for every row.

So a chat list of 200 rows with 150 direct peers pays 150 directory-record reads plus two extra storage round trips *on every read*, and a nominally read-only API can take the account DB's exclusive `with_transaction` owner lock. Issue #1517's acceptance criterion was explicitly "without requiring host-side persistence **or extra per-row profile calls for the first frame**" — the durable columns already satisfy that, and the write-path projection (`project_direct_peer_profile_for_accounts` on profile ingestion, `direct_peer_presentation_for_rebuild_tx` on roster change) already keeps them fresh.

This is amplified by the new C accessor: `marmot_chat_list_direct_peer_presentation` calls `chat_list_row`, so a C host reading presentation for N rows performs N full row queries + N directory hydrations.

Suggested: make hydration a one-time warm/backfill step (e.g. hang it off `ensure_chat_list_rows` / the projection-version bump that already forces a rebuild) rather than running it on every read, and let reads return the durable columns directly.

crates/marmot-app/src/directory/methods.rs
```diff
@@ -1119,18 +1119,46 @@ impl MarmotApp {
...
         let mut entry = self
             .directory_entry_for_account_id(account_id_hex)?
             .unwrap_or_else(|| self.empty_directory_record(account_id_hex));
         entry.profile = Some(profile.clone());
-        self.save_directory_entry(&entry)
+        self.save_directory_entry(&entry)?;
+        let accepted_profile = self
+            .directory_entry_for_account_id(account_id_hex)?
+            .and_then(|entry| entry.profile)
+            .ok_or_else(|| AppError::MissingDirectoryEntry(account_id_hex.to_owned()))?;
```
erskingardner, line 1139: **Regression: a successful profile save can now report failure.**

`save_directory_entry` has already committed at this point. If the read-back returns no entry or an entry without a profile, this `?` turns a successful persist into `AppError::MissingDirectoryEntry`, and every caller of `remember_directory_profile` / `remember_directory_profile_if_newer` — none of which asked for the presentation projection — now sees that error. A derived display-only projection must not be able to reject an accepted directory profile.

This was raised in the 2026-09-04T14:28 automated review (finding 4) and is still present at this head.

Smallest fix: fall back to the profile just saved when read-back yields nothing, and keep returning `Ok`:

```rust
let accepted_profile = self
    .directory_entry_for_account_id(account_id_hex)?
    .and_then(|entry| entry.profile)
    .unwrap_or_else(|| profile.clone());
```

The same argument applies to the `project_direct_peer_profile_for_accounts` call below — it already swallows its own storage errors, so the read-back is the only remaining way this path can fail for a reason unrelated to the save.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -429,6 +597,98 @@ impl SqliteAccountStorage {
...
+        group_ids_hex: &[String],
+    ) -> StorageResult<HashMap<String, Option<DirectPeerPresentation>>> {
+        let conn = self.lock()?;
+        direct_peer_presentations_for_group_ids_tx(&conn, group_ids_hex)
+    }
+
+    /// Project one accepted profile into every exact two-member Direct chat
+    /// whose roster is `{local_account_id_hex, profile.peer_account_id_hex}`.
+    /// The write changes presentation only; it never advances activity order.
+    pub fn project_direct_peer_profile(
```
erskingardner, line 613: **No read preflight here, so every accepted profile event takes an exclusive account-DB transaction — for every account, even when there is nothing to write.**

`SqliteConnection::with_transaction` serializes on `transaction_owner` for the *whole* account database, blocking every other thread on that account until it releases. `project_direct_peer_profile_for_accounts` calls this once per local account on every accepted profile, and profile ingestion is a relay firehose path — most peers have zero direct conversations with a given account, so the overwhelmingly common outcome is `direct_conversation_group_ids_for_peer_tx` returning empty inside a transaction that never writes.

`hydrate_direct_peer_profiles` (just below) got exactly this treatment right with `direct_peer_profile_hydration_needed_tx`. Mirror it here — do the group-id lookup under the plain read lock and return early on empty:

```rust
pub fn project_direct_peer_profile(&self, local_account_id_hex: &str, profile: &DirectPeerProfile)
    -> StorageResult<Vec<ChatListRow>>
{
    // ... normalize local/peer, bail on empty/self ...
    let group_ids = { let conn = self.lock()?; direct_conversation_group_ids_for_peer_tx(&conn, &local, &peer)? };
    if group_ids.is_empty() { return Ok(Vec::new()); }
    self.connection.with_transaction(|| { /* existing body */ })
}
```

crates/marmot-app/src/runtime/subscriptions.rs
```diff
@@ -1456,6 +1600,62 @@ impl MarmotAppRuntime {
                     {
                         return;
                     }
+                    if let Some(mut retry_row) = retry_row {
+                        tokio::select! {
+                            _ = wait_for_runtime_shutdown(&mut stopping) => return,
+                            _ = tokio::time::sleep(CHAT_LIST_PRESENTATION_LOOKUP_RETRY_DELAY) => {}
```
erskingardner, line 1606: **The 250 ms retry sleeps inside the subscription event loop, stalling every other chat-list event for that account.**

Both retry sites (`1483` and here) `tokio::time::sleep` on the loop's own task before looping back to `events.recv()`. Nothing else is consumed during that window. The runtime `events` broadcast has capacity 1024 (`runtime/mod.rs:1116`), so a stall under load pushes this receiver into `RecvError::Lagged`, whose handler at line 1363 responds with a **full `chat_list()` snapshot rebuild + reconcile**. Recovering a display-only avatar/name is being paid for with a whole-list resync and delayed unread/preview delivery.

More fundamentally, per the repo's simplicity guideline this is ~150 lines of speculative error handling: the only way the tests can reach it is by calling `storage.close()` and then manually evicting `app.account_storages` to "repair" the handle. In production the next projection event for that group already re-reads presentation unconditionally (that is the whole point of the block at line 1502), so the row self-heals without a retry.

Recommended: drop both retry blocks. Emit the presentation-less row (fail-closed, which is already correct) and let the next event repair it. If a retry is genuinely wanted, `tokio::spawn` it so it cannot block the loop — but I would land the simpler version first.

crates/marmot-app/src/runtime/subscriptions.rs
```diff
@@ -73,6 +73,22 @@ pub enum RuntimeTimelineMessageUpdate {
...
 pub(crate) const TIMELINE_WINDOW_LIMIT: usize = MAX_TIMELINE_LIMIT;
 
+/// Gives a transient account-storage failure a brief recovery window before a
+/// chat-list subscriber performs its single presentation-only retry.
+const CHAT_LIST_PRESENTATION_LOOKUP_RETRY_DELAY: Duration = Duration::from_millis(250);
+
+/// Counts failed initial direct-peer wake lookups so the runtime regression
+/// test can repair the cached storage handle before the bounded retry runs.
+#[cfg(test)]
+pub(super) static DIRECT_PEER_WAKE_LOOKUP_FAILURES_FOR_TEST: std::sync::atomic::AtomicUsize =
```
erskingardner, line 83: **Process-global test counters in production source, and the test that reads them is order-dependent.**

These two statics exist only so `chat_list_subscription_updates_when_direct_peer_presentation_changes` can spin until a lookup failure is observed. Two problems:

1. Cargo runs a crate's tests as threads in one process, so these counters are shared across *all* concurrently running tests. The spin loops only assert `counter > before`, so an increment produced by any other test that closes an account store while a chat-list subscription is live satisfies the wait early and desynchronizes the rest of the ~380-line test. That is a latent flake, and this repo already has a note about hard-to-reproduce local chat/relay-runtime flakes.
2. They are test scaffolding leaking into the production module, which the retry logic they support (see the comment at line 1606) may not need at all.

If the retry stays, prefer an injected per-subscription hook/`Notify` owned by the test rather than a process-global counter — or split the mega-test into focused cases so each can own its own app instance and observe failure through the subscription output alone.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -1231,7 +1508,11 @@ fn set_chat_list_projection_version_tx(tx: &Connection, version: i64) -> Storage
 /// `ChatListRow::leave_requested_at_ms` is derived at read time rather than
 /// materialized, so there is no stored value that can drift out of date — and
 /// comparing a read-time-only field here would mark every row stale forever.
-fn chat_list_projection_complete_tx(tx: &Connection) -> StorageResult<bool> {
+fn chat_list_projection_complete_tx(
```
erskingardner, line 1511: **The new completeness predicate is ~65 lines of nested SQL `CASE`/`EXISTS`/`NOT (...)` and re-encodes invariants that already exist in Rust.**

The added block restates, in SQL: the exact-two-member Direct test (three times, including once inside the `NOT (...)` repairable-exclusion clause), the schema-version check, the peer-id match, the legal state set, `absent` implies no values, and `current|last_known` implies a non-negative `profile_created_at`. Those last three are precisely the normalizations `direct_peer_presentation_from_columns` already applies at read time, and the first two are what `direct_peer_presentation_for_rebuild_tx` already enforces at write time.

This is the hardest-to-verify thing in the PR, it is duplicated logic that can drift from the Rust side, and a mistake here fails in one of two silent ways: a permanent full-projection rebuild on every warm open, or a stale row that never gets repaired.

Suggested simplification: keep only the cheap staleness signals that a rebuild can actually fix — version mismatch, and peer id not matching the current exact-Direct peer (with the `invalidated`-but-repairable carve-out) — and drop the value-consistency clauses, since the reader already normalizes them and `write_chat_list_row_for_group_tx` cannot produce them. Factoring the exact-two-member test into one named SQL fragment (as `direct_conversation_group_ids_for_peer_sql` does) would also remove the triplication.

crates/storage-sqlite/src/chat_list.rs
```diff
@@ -111,6 +112,170 @@ pub enum ChatConversationKind {
...
+        .flatten();
+    let Some(stored_peer) = stored_peer else {
+        return Ok(());
+    };
+    let prior_local_member = conn
+        .query_row_cached(
+            "SELECT member.member_id_hex
+             FROM direct_conversation_members AS member
+             WHERE member.group_id_hex = ?1
+               AND lower(member.member_id_hex) != lower(?2)
```
erskingardner, line 181: **Inconsistent normalization assumption: `lower(...)` here, exact `=` elsewhere in the same PR.**

`replace_direct_conversation_members_tx` (account_projection.rs:3826) lowercases every `member_id_hex` before insert, and this PR's own `direct_conversation_group_ids_for_peer_sql` documents that ("Member ids are normalized on every index write") and relies on exact `=` to keep the covering index usable.

But four of the new queries wrap the indexed column in `lower()` — here, `current_direct_peer_account_id_tx`, `invalidated_direct_peer_group_ids_tx`, and the completeness predicate. `lower(member_id_hex)` is not sargable, so those lose the index on `direct_conversation_members`, and the two conventions together leave a reader unsure which invariant actually holds.

Pick one. Given the write-side guarantee, dropping `lower()` from the column side (keeping normalization of the *input* parameters, which the callers already do) is the smaller and faster change.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T11:41:30Z` |
| Commit reviewed | `b7af025a28acf485fa25377c7524771741c1b822` |

## Premise

The premise holds. #1517 documents a real cold-start gap, the durable-column-plus-schema-version design is the right shape for it, the migration is additive and correctly defaulted, and every acceptance criterion has a named test (`refresh_chat_list_row_projects_direct_peer_presentation`, `direct_peer_presentation_round_trips_to_ffi`). The privacy posture is careful throughout — hand-written `Debug` impls redact peer ids and values, tracing carries only aggregate counts, and `DirectPeerPresentationState::from_storage` fails closed on unknown provenance. The C ABI resolution (sidecar accessor plus a compiled-previous-header stride regression test) is a genuinely good answer to the earlier blocking finding.

My concerns are not about whether this should land, but about how much machinery it brings with it. 3584 added lines for one durable projection is a lot, and a sizeable share of it is speculative.

## Serious concerns

1. **Read paths do per-read directory work and can write** (`marmot-app/src/lib.rs`). Every `chat_list_rows` / `chat_list_row` / `refresh_chat_list_row` now runs `cached_identity_projections_for_account_ids` for every distinct peer, a hydration preflight, possibly a write transaction, and a presentation re-read. #1517 asked for a first frame with no extra per-row profile calls; the durable columns deliver that, and the write-path projection already keeps them fresh. This reintroduces the per-read lookups the projection was meant to remove, and is amplified by the C sidecar, which calls `chat_list_row` once per row.

2. **`project_direct_peer_profile` has no read preflight** (`storage-sqlite/src/chat_list.rs`). It opens the account DB's exclusive `with_transaction` owner lock once per local account on every accepted profile ingestion, even though the common case writes nothing. `hydrate_direct_peer_profiles` right below it solves exactly this with `direct_peer_profile_hydration_needed_tx`; mirror that.

3. **A successful profile save can now return an error** (`marmot-app/src/directory/methods.rs:1139`). After `save_directory_entry` commits, a read-back miss produces `AppError::MissingDirectoryEntry`, which reaches every caller of `remember_directory_profile` / `..._if_newer`. A display-only derived projection must not be able to reject an accepted directory profile. This was raised in the 2026-09-04T14:28 automated review (finding 4) and is unaddressed at this head.

4. **The 250 ms retry sleeps inside the chat-list subscription loop** (`runtime/subscriptions.rs`). Nothing else is consumed during the stall; the runtime broadcast holds 1024 events, and lagging that receiver triggers a full `chat_list()` snapshot rebuild — an expensive recovery for a display-only field that the next projection event would repair on its own, since that path re-reads presentation unconditionally.

## Simplicity and scope

5. **`chat_list_projection_complete_tx` gained ~65 lines of nested SQL** that re-encode invariants `direct_peer_presentation_from_columns` already normalizes at read time and `direct_peer_presentation_for_rebuild_tx` already enforces at write time. Duplicated logic in the hardest-to-verify place in the PR; a mistake fails silently in either direction.

6. **Test scaffolding in production source.** Two process-global `*_FOR_TEST` atomics in `subscriptions.rs` support one ~380-line test covering seven scenarios. Because cargo runs a crate's tests in one process, the `counter > before` spin loops can be satisfied by an unrelated concurrent test — a latent flake, in a repo that already tracks hard-to-reproduce local chat/runtime flakes.

7. **Normalization convention is inconsistent.** The PR documents that `member_id_hex` is lowercased on every index write and relies on exact `=` in two queries, then wraps `lower()` around the same indexed column in four others (non-sargable). Pick one.

## Smaller notes

- `subscriptions.rs`: the ordinary-event branch maps `Ok(None)` from `direct_peer_presentations_for_group_ids` to a **row removal**. That map omits group ids with no `chat_list_rows` row, so a row deleted between event production and this lookup silently converts an unread/preview update into a `Removed`. Defensible, but it deserves a comment saying that is intended.
- `direct_peer_presentation_from_columns` returns `Some` with the *foreign* `schema_version` when the version does not match, alongside `state: Invalidated`. Hosts see an unknown version number on the FFI record; consider normalizing it to the current constant since all values were cleared anyway.
- A named group that was previously Direct keeps `state: Invalidated` forever (nothing repairs non-Direct rows), so hosts will see a presentation record on plain groups. Cosmetic, but worth documenting in the binding contract.
- `directory/records.rs`: the `sanitize_profile_field` rewrite fixes a real pre-existing trim/control-character ordering bug and is well-tested, but it also changes `display_name_for_profile` behavior for callers unrelated to this issue. Justified here since these values are now persisted — just flagging it as scope worth calling out in the description.

## Validation

Remote review at the exact head plus the PR's own CI evidence. No local build or tests were run, per this review's workspace rules. The PR body's Pixel device results are not bound to this head.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T11:48:16Z` |
| Commit reviewed | `b7af025a28acf485fa25377c7524771741c1b822` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

This qualifies my earlier approval. I checked correctness and the ABI fix more thoroughly than whether the design needed this much machinery. After reassessing the simplicity concerns, I recommend a focused simplification before merge; the underlying feature remains worthwhile.

The size is 3,584 additions / 61 deletions across 20 files. Classifying added lines gives approximately 1,924 lines of tests, fixtures and smoke coverage, 60 generated header lines, and 1,600 implementation lines (including documentation and test-only instrumentation in the runtime module). The test volume is not the main objection. Storage chat-list code, app orchestration, and subscriptions account for 1,365 implementation additions.

The strongest concerns are:

- **Routine reads repeatedly reconstruct a durable projection.** `hydrate_direct_peer_presentations` (`marmot-app/src/lib.rs:4771`) performs directory lookups, hydration preflight, possible writes, and another presentation read. The cached-identity page API still walks the requested identities individually. The C sidecar invokes the full app row path for each requested presentation. These operations are visible in the code; no measured latency regression is claimed. Move peer-presentation hydration to explicit initialization/update/repair boundaries so unchanged warm reads consume the saved values.
- **The same small projection has too many maintenance paths.** Profile ingestion, read-time batch hydration, completeness SQL, targeted invalidated-row repair, and rebuild logic all participate. Consolidate shared roster/projection decisions and use one profile-application primitive. That primitive should avoid a write transaction for unaffected accounts; the current ingestion wrapper enters `with_transaction` before discovering there are no matching chats.
- **Presentation recovery adds two inline retry branches to the subscription loop.** Both pause event consumption for 250 ms, with duplicated lookup/error handling and a large combined test. Retain a bounded recovery guarantee, but coalesce pending group refreshes and schedule them without sleeping inside the ordinary-event handling branch. Preserve delivery of unread/preview changes and stale-presentation suppression.

A smaller design must still handle profiles cached before a conversation exists, upgrades with existing rows, peer replacement, and a failed projection write after profile acceptance. Simply deleting hydration or relying on the next unrelated event would lose those guarantees. Likewise, read-time normalization is not a substitute for durable repair; simplify the completeness logic only while retaining an explicit repair path.

I do not carry forward every claim in the latest review. A newer profile-less shared record does not discard a saved profile: directory reconciliation merges profile components independently. Falling back to the raw proposed profile can bypass that accepted-profile selection. The test atomics are compiled only under `cfg(test)`; cross-test interference and broadcast overflow have not been demonstrated. These are not additional proven blockers.

Acceptance for the simplification: no peer-presentation directory lookup/write on an unchanged warm read; no write transaction for a profile with no matching direct chat; cached-before-create and upgrade seeding work; roster replacement never exposes former-peer values; recovery for a quiet conversation completes while unrelated row events remain deliverable; and the existing persistence, isolation, FFI, and previous-header C compatibility tests remain covered. Keep the durable representation and public contract; avoid unrelated architectural changes.

Required CI remains green at this exact head. This assessment is based on remote source and existing CI, not new execution or performance measurements.

erskingardner:
@mubarakcoded what's this for? seems like it's targeting a small performance gain in projecting user profile data? but it also seems like a lot of logic for that... including a few situations where we would get blocked trying to refetch data. I'm not sure I'm behind this overall PR direction...

Datawav (approved):

mubarakcoded:
@erskingardner The goal here isn't really a small performance gain. It's to give clients an authoritative, durable direct-peer name/avatar in the chat-list row after a cold start, instead of requiring each client to reconstruct or persist that state itself. Android is where we've observed the "Unknown" first-frame issue
That said, I agree the implementation has grown too much around that goal. I'm thinking the revised scope should be: keep the migration and public binding contract, but remove repeated directory work from normal reads, avoid unnecessary write transactions, make profile projection strictly best-effort, and move recovery retries out of the subscription event loop. We'd still preserve upgrades, cached-before-create profiles, restart behavior, and safe peer replacement.
WDYT about that as the revised scope?

erskingardner:
Thanks for clarifying the cold-start goal. We agree that MDK should provide authoritative, durable direct-peer presentation so clients do not have to reconstruct or persist it themselves. The Android first-frame “Unknown” problem is a real requirement, and #1517 remains open.

We're closing this PR in favor of the broader contract-design effort in #1742. That parent now contains the source-pinned map of existing chat-list, timeline, group/roster, draft, directory and related projection surfaces, plus an audit of the existing issues and the decisions needed before splitting implementation work.

The intended fast path is a complete, bounded initial chat list and conversation-open response, maintained by background updates, while preserving useful lower-level APIs for non-chat applications. That requires agreeing resolved title/avatar/preview fields, identity and asset ownership, filtering/paging, snapshot/update consistency and recovery before committing to this PR's migration or public binding shape. In particular, keeping separate group and peer presentation fields would still leave display selection to clients, and the C sidecar would still require another read.

Please preserve this PR as reference material. Its requirements and tests around upgrades, cached-before-create profiles, restart behavior, profile updates, safe peer replacement and account isolation should carry into the agreed implementation. The suggestions to remove repeated directory work from normal reads and move recovery out of the subscription loop are useful design inputs too. We are not treating the schema or freshness-state model here as fixed constraints, and closing this does not mean the existing projection machinery needs to be thrown away.

Next we will settle the screen contracts and ownership boundaries in #1742, reconcile the linked issues, then map existing issues and any missing children to small, reviewable portions of the work. Closing without merging; the underlying feature remains tracked.

PR #1686, opened 2026-09-04
ci: avoid GitHub rate limits in Hermes checks

erskingardner:
## Summary

- install the released Hermes 0.19.0 wheel in required CI instead of cloning the upstream GitHub repository
- pin the wheel URL and SHA-256 alongside the existing Hermes source revision
- cover the no-GitHub, hash-pinned install path in the Hermes dev-script regression suite
- use the same path in the wn-agent Hermes integration job

## Root cause

The two latest master CI runs failed only in `Hermes installer auth`. All Rust test shards and other selected jobs passed. The setup retried a cross-repository `git clone` five times, but GitHub returned HTTP 429 on every attempt; the repository-scoped Actions token did not eliminate that external clone dependency.

## Verification

- `actionlint .github/workflows/ci.yml .github/workflows/wn-agent-binaries.yml`
- pinned-wheel setup + real Hermes plugin enable + real authorization probe + deterministic Marmot E2E
- `just fast-ci`
- `git diff --check`

## Summary by CodeRabbit

* **Chores**
  * Updated development and CI setup to install a validated, hash-pinned Hermes package when requested.
  * Retained source-based installation for workflows requiring repository checkouts.
  * Improved setup consistency with Python 3.11 and retry handling for reference fetching.
  * Avoided redundant fetches when requested commits are already available locally.

* **Documentation**
  * Clarified pinned-package setup and lock-file requirements.

* **Tests**
  * Added coverage for pinned installation, validation checks, checkout behavior, and package configuration.


integrations/hermes/marmot/hermes-agent.lock
```diff
@@ -1,3 +1,5 @@
 HERMES_AGENT_VERSION=0.19.0
 HERMES_AGENT_TAG=v2026.7.20
 HERMES_AGENT_REF=3ef6bbd201263d354fd83ec55b3c306ded2eb72a
+HERMES_AGENT_WHEEL_URL=https://files.pythonhosted.org/packages/e5/30/c85be8290e9565dc3c7a9720e93f3e59e09b1b163487be4946c3aa848f80/hermes_agent-0.19.0-py3-none-any.whl
```
erskingardner, line 4: **Drift hazard (non-blocking, but worth a one-liner).** The lock file now carries two independent descriptions of "which Hermes": `HERMES_AGENT_VERSION`/`TAG`/`REF` for the source path, and `WHEEL_URL`/`WHEEL_SHA256` for the required-CI path. Nothing ties them together.

A future upstream bump that updates `HERMES_AGENT_VERSION`/`REF` but forgets the wheel fields leaves required CI silently validating 0.19.0 while every other consumer (Dockerfile.hermes-marmot, local dev) moves on — and the validation in `install_hermes` won't catch it, because a stale URL is still a well-formed pythonhosted URL with a well-formed hex digest.

One cheap guard in `install_hermes` closes it:

```sh
[[ "$HERMES_AGENT_WHEEL_URL" == *"/hermes_agent-${HERMES_AGENT_VERSION}-"* ]] || {
    echo "error: pinned wheel URL does not match HERMES_AGENT_VERSION" >&2
    exit 1
}
```

scripts/hermes_marmot_dev_setup.sh
```diff
@@ -248,20 +254,35 @@ clone_hermes_repo() {
 }
 
 install_hermes() {
-    if [ ! -d "$hermes_repo/.git" ]; then
-        clone_hermes_repo "$hermes_repo"
-    fi
+    local install_target=".[all,dev]"
+    if [ "$install_pinned_wheel" -eq 1 ]; then
```
erskingardner, line 287: **`--hermes-ref` / `HERMES_AGENT_REF` becomes a silent no-op under `--install-pinned-wheel`.**

The `else` branch owns the `hermes_ref` fetch/checkout, so `--install-pinned-wheel --hermes-ref some-branch` installs 0.19.0 from PyPI and says nothing. The README explicitly advertises `--hermes-ref` as the deliberate-upstream-testing knob, which is exactly the case where a silent override is most confusing.

Since `hermes_ref` defaults from the lock, you'd need to detect an *explicit* override (the script already keeps `hermes_ref_override` around for the env case). Erroring out on the combination is a two-line fix and strictly better than ignoring it.

scripts/hermes_marmot_dev_setup.sh
```diff
@@ -248,20 +254,35 @@ clone_hermes_repo() {
...
+        fi
     fi
 
     if ensure_uv; then
         (
             cd "$hermes_repo"
             uv venv .venv --python 3.11
-            uv pip install -e ".[all,dev]"
+            if [ "$install_pinned_wheel" -eq 1 ]; then
+                uv pip install "$install_target"
```
erskingardner: **Simplification (net -4 lines).** The `install_target` string forces `if install_pinned_wheel` to be re-tested in both the uv branch and the pip branch, purely to decide whether `-e` is prepended. Carrying an argv array instead collapses all three decision points into the one that already exists:

```sh
local install_args=(-e ".[all,dev]")
if [ "$install_pinned_wheel" -eq 1 ]; then
    ...
    install_args=("hermes-agent[all,dev] @ ${HERMES_AGENT_WHEEL_URL}#sha256=${HERMES_AGENT_WHEEL_SHA256}")
fi
```

then `uv pip install "${install_args[@]}"` and `python -m pip install "${install_args[@]}"` with no branching at the call sites.

.github/workflows/ci.yml
```diff
@@ -645,12 +645,11 @@ jobs:
             python3 -m unittest discover -s integrations/hermes/marmot/tests -v
 
       - name: Prepare isolated Hermes checkout for authorization E2E
```
erskingardner: Nit: the step is no longer a "checkout" — it creates an empty `hermes-agent/` directory holding only a `.venv`. Same for the `$hermes_repo` variable and the exported `HERMES_AGENT_REPO` in `env.sh`, which now name a non-repo. Renaming the step (e.g. "Prepare isolated Hermes environment for authorization E2E") keeps the log readable; renaming the variables is more churn than it's worth right now, but the README paragraph you added is the only place a reader learns the directory is no longer a checkout.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T14:57:44Z` |
| Commit reviewed | `2afc735d9ddc06364febf25fd1587e82e2ba8bb1` |

_Posted as a comment review: GitHub does not allow approving your own pull request._

## Verdict

Good fix, correctly scoped, and it addresses the root cause rather than the symptom. The previous change added retries and an Actions token to a cross-repository `git clone`; neither can help when the dependency itself is the problem. Removing the clone from required CI is the right move, and dropping `GITHUB_TOKEN` from both steps is a nice secondary win — those steps no longer need a token at all.

Verified against remote CI on this head: `Hermes installer auth` passed in 41s (down from a multi-minute clone-and-retry), the log shows `uv` resolving `hermes-agent==0.19.0` from the pinned URL with the `#sha256=` fragment intact and 104 packages installed, and the real authorization + deterministic Marmot E2E still passed against the wheel-backed venv. `Hermes Marmot persisted config` passed too. Both callers of `hermes_marmot_dev_setup.sh` in CI were updated, so the GitHub dependency is gone from the required path rather than merely reduced.

The test addition is the right shape: it asserts the *absence* of `.git` under the dev root, not just the presence of the wheel argument, so a regression back to cloning fails the assertion directly.

## Design question worth a reply before or after merge

`--install-pinned-wheel` is opt-in, which leaves the repo with two install paths pointing in opposite directions: CI now exercises only the wheel path, while the default path — the one a developer gets from `just hermes-dev` and the one `Dockerfile.hermes-marmot` still uses — clones GitHub and is no longer end-to-end exercised anywhere. Developers also keep hitting the 429s this PR exists to avoid.

Given the README already frames `--hermes-ref` as "only for deliberate upstream testing", the source clone is effectively the exceptional path. Inverting the default — pinned wheel unless a ref override is requested — would delete the flag from both workflows, remove the CI/dev divergence, and end up with fewer lines than this change. Not blocking, and I understand wanting the smaller diff first; worth deciding deliberately rather than by default.

## Non-blocking notes

See inline comments:

- The two lock-file "which Hermes" descriptions (`VERSION`/`TAG`/`REF` vs `WHEEL_URL`/`WHEEL_SHA256`) are unlinked; a future bump can silently strand required CI on 0.19.0. A one-line guard fixes it.
- `--hermes-ref` / `HERMES_AGENT_REF` is silently ignored under `--install-pinned-wheel`.
- The `install_target` string forces the `install_pinned_wheel` test to be repeated at both call sites; an argv array removes the duplication.
- The CI step name still says "checkout" for a directory that no longer contains one.

Nothing here blocks merge.

integrations/hermes/marmot/hermes-agent.lock
erskingardner: Addressed in 8e5bfef0. Pinned-wheel setup now verifies that the wheel filename contains the locked HERMES_AGENT_VERSION, and the dev-script test mutates a copied lock file to prove version drift fails closed.

scripts/hermes_marmot_dev_setup.sh
erskingardner: Addressed in 8e5bfef0. The setup now tracks whether HERMES_AGENT_REF or --hermes-ref was explicitly supplied and rejects that combination with --install-pinned-wheel before creating the dev root. The env override path has regression coverage.

scripts/hermes_marmot_dev_setup.sh
erskingardner: Addressed in 8e5bfef0. Installation arguments are now built once as an argv array and shared by both uv and pip; the pinned-wheel regression also asserts editable mode is absent.

.github/workflows/ci.yml
erskingardner: Addressed in 8e5bfef0. The CI step is now named “Prepare isolated Hermes environment for authorization E2E.” I left the existing path/env variable names unchanged to keep this follow-up scoped.

erskingardner:
Addressed the four scoped inline notes in 8e5bfef0. Validation: `integrations/hermes/marmot/tests/test_dev_scripts.sh`, `just fast-ci`, `actionlint .github/workflows/ci.yml`, shell syntax checks, and `git diff --check` all pass locally.

I intentionally left the pinned-wheel-by-default design question unchanged in this PR. Inverting that default would also change `just hermes-dev` and Docker/source-testing behavior, so it deserves a separate deliberate change rather than expanding this CI reliability fix.

erskingardner:
Resolved the conflicts with merged #1680 in c85ce064 while preserving both fixes.

#1680’s Hermes change avoids a redundant ref fetch after a successful clone. It does not supersede this PR: the failing master run (33879143323) failed all five attempts in the initial `git clone` with HTTP 429, before any ref fetch. This PR still removes that cross-repository clone from required CI by using the hash-pinned wheel.

Post-merge validation passes locally: Hermes dev-script regressions, `just fast-ci`, workflow lint, shell syntax checks, and `git diff --check`.

scripts/hermes_marmot_dev_setup.sh
```diff
@@ -13,6 +13,7 @@ Options:
   --marmot-home PATH       Marmot/wn-agent home (default: ROOT/marmot-agent-home)
   --hermes-url URL         Hermes repo URL (default: https://github.com/NousResearch/hermes-agent.git)
   --hermes-ref REF         Optional branch, tag, or commit to checkout
+  --install-pinned-wheel   Install the hash-pinned Hermes wheel instead of cloning source
```
erskingardner, line 16: Non-blocking, doc + consistency follow-up to the new mutual-exclusion check.

1. The new rule is a usage error (`exit 2`), but neither this help line nor `integrations/hermes/marmot/README.md` says the two options are incompatible. A trailing `(not compatible with --hermes-ref/HERMES_AGENT_REF)` here would make the failure self-explanatory.

2. `--hermes-url` / `HERMES_AGENT_REPO_URL` is still silently ignored under `--install-pinned-wheel` — exactly the class of bug `hermes_ref_is_explicit` was added to close. The same `*_is_explicit` pattern applied to the URL would finish the job. Only worth doing if you agree the principle should hold for both source-oriented options; otherwise ignore.

scripts/hermes_marmot_dev_setup.sh
```diff
@@ -267,29 +283,44 @@ fetch_hermes_ref() {
...
+    if [ "$install_pinned_wheel" -eq 1 ]; then
+        if [[ ! "$HERMES_AGENT_WHEEL_URL" =~ ^https://files\.pythonhosted\.org/ ]] \
+            || [[ ! "$HERMES_AGENT_WHEEL_SHA256" =~ ^[0-9a-f]{64}$ ]]; then
+            echo "error: invalid pinned Hermes wheel URL or SHA-256" >&2
+            exit 1
+        fi
+        if [[ "$HERMES_AGENT_WHEEL_URL" != *"/hermes_agent-${HERMES_AGENT_VERSION}-"* ]]; then
+            echo "error: pinned Hermes wheel URL does not match HERMES_AGENT_VERSION" >&2
+            exit 1
+        fi
```
erskingardner, line 296: Non-blocking: placement asymmetry between the two new validations.

The ref-conflict check lands at line 168, before any filesystem mutation — which matches the repo's validate-before-mutate invariant. These wheel URL / SHA-256 / version checks still run inside `install_hermes()`, which is invoked at line 596, long after `mkdir -p "$dev_root" ...` (189), the plugin symlink (215), and the generated `env.sh`/helper scripts. A lock-file typo therefore aborts with a half-built dev root on disk, and the new drift regression test exercises exactly that path.

Hoisting these two `if` blocks next to the ref-conflict check (they only read lock-file variables, so nothing forces them to live here) would make every pinned-wheel failure mode fail before touching the filesystem. The `mkdir -p "$hermes_repo"` and `install_args=` assignment stay put.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-04T15:37:23Z` |
| Commit reviewed | `c85ce064232c000880b01d222026207f6a94b441` |

Posted as a comment rather than an approval: GitHub does not allow approving your own pull request. The recommendation is **Merge**.

Re-review of the delta since `2afc735d` only. The prior review's conclusions stand; this pass covers `8e5bfef0` and the master merge `c85ce064`.

## Prior notes: status

| # | Prior note | Status |
| --- | --- | --- |
| 2 | Lock-file drift between `HERMES_AGENT_VERSION`/`REF` and `WHEEL_URL`/`SHA256` | Addressed for the URL: `HERMES_AGENT_WHEEL_URL` must contain `/hermes_agent-${HERMES_AGENT_VERSION}-`, with a regression test that mutates a copied lock file. `SHA256` drift is still caught by pip/uv at install time, and `REF` drift is not locally checkable — a reasonable stopping point. |
| 3 | `--hermes-ref` silently ignored under the new flag | Addressed. `hermes_ref_is_explicit` is captured from `HERMES_AGENT_REF` **before** the lock file is sourced, so the lock's own `HERMES_AGENT_REF=3ef6bbd…` correctly does not count as explicit, and `--hermes-ref` sets it at parse time. Hard `exit 2` with a regression test. |
| 4 | `install_target` duplicated the `install_pinned_wheel` test at both call sites | Addressed. `install_args` array collapses both branches to a single `uv pip install "${install_args[@]}"` / `python -m pip install "${install_args[@]}"`, and the test asserts `pip install -e ` never appears on the pinned path. Net simplification. |
| 5 | CI step still named "checkout" | Addressed — now "Prepare isolated Hermes environment for authorization E2E". |
| 1 | Default clone path and `Dockerfile.hermes-marmot` unexercised by the opt-in flag | Unchanged, as settled previously. Noted below only because the merge shifts what covers it. |

## New validation logic

Correct as written. No blocking issues.

- Ordering is right: `hermes_ref_override` is read before `. hermes-agent.lock`, so the lock-supplied ref is never mistaken for user intent.
- No workflow or `Justfile` recipe exports `HERMES_AGENT_REF`, so the new hard error cannot fire spuriously in CI. `docker-compose.yml`'s `HERMES_AGENT_REF` build arg feeds `Dockerfile.hermes-marmot`, which does its own clone and never calls this script.
- The version-match check uses the `/hermes_agent-<version>-` filename segment, which the current lock URL satisfies. It assumes a PyPI-normalized version string in the lock file — fine for a repo-owned lock.
- Both new tests are self-validating (assert non-zero exit *and* grep the specific stderr), and `test_dev_scripts.sh` runs in required CI via the "Hermes installer unit and mock regressions" step, so the validation is actually gated.

Two non-blocking follow-ups inline: help-text/README don't mention the new incompatibility, `--hermes-url` is still silently ignored, and the wheel validations sit after the dev root has already been built while the ref check sits before it.

## Master merge: clean

Verified by three-way comparison rather than by reading the merge diff alone.

- `git diff 9f3f727f c85ce064` (master tip → merge) touches **only** the six PR-owned files. Nothing from master was reverted.
- `git diff 8e5bfef0 c85ce064` (PR side → merge) adds **only** master's changes. Nothing from the PR was dropped.
- The +30/-2 that the merge commit applies to `scripts/hermes_marmot_dev_setup.sh` is not hand-written conflict resolution: it is master's `fetch_hermes_ref` retry helper and 40-hex local-commit checkout shortcut from #1680, preserved verbatim and correctly re-nested inside the new non-pinned `else` branch. The accompanying +18 test lines are master's `github-local-ref` case, likewise intact.
- `--install-pinned-wheel` and the removal of the step-level `GITHUB_TOKEN` survived in both `ci.yml` and `wn-agent-binaries.yml`. Neither was reintroduced by the merge.

One observation, not a defect: #1680 landed a *different* mitigation for the same 429 in the same job, and this PR now routes required CI around the clone entirely. That leaves master's brand-new clone/fetch retry path covered in required CI only by the fake-`git` dev-script test. Both mitigations are cheap and the pinned wheel is strictly the stronger one for CI, so keeping both is fine — just don't assume the retry path is still exercised end-to-end.

## Verification

All checks green at `c85ce064`, including `Hermes installer auth` (42s) and `Hermes Marmot persisted config` — the two jobs this PR exists to unbreak.

Recommend merge.

PR #1687, opened 2026-09-04
test(engine): pin join-epoch anchoring and drop the false snapshot-nesting claim

jgmontoya:
Two pieces salvaged from #1666 (closed, superseded by #1675) that stand on their own.

**Join-epoch anchor test.** `do_join_welcome` retains no anchor at the join epoch; the joiner's first advance does (`ingest.rs` retain on direct ingest, the pre-apply retain in convergence, `publish.rs` on confirm). `joiners_first_advance_anchors_the_join_epoch_so_a_rival_there_adjudicates` pins that a rival forking at the join epoch is adjudicated rather than halting: it asserts the anchor, the rival's disposition, the tip epoch, and the rolled-back loser. RED-verified two ways: disabling the pre-apply retain fires the anchor assertion, and with that assertion removed the group halts with `MissingRetainedAnchor`.

**Stale nesting comments.** Three comments claimed snapshot ops "cannot nest" inside `with_transaction`. They can: `with_transaction` is reentrant by flattening on the owning thread (`connection.rs:461-465`), and snapshot capture/restore/checkpoint open a transaction only when the thread owns none (`is_current_thread_transaction_owner` guards). #1675 retains anchors two frames deep. The false premise is deleted; the surviving true reasons stay. `publish.rs`'s pre-transaction retain now has no stated placement rationale; it is idempotent and harmless, and moving it is a separate behavior change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved group convergence for members who join after earlier updates and then advance the group.
  - Competing updates can now be resolved deterministically without incorrectly halting the group or reporting a missing retained anchor.
  - Correct branch selection, rollback, adoption, and member state are preserved during recovery from divergent update histories.

- **Documentation**
  - Clarified safeguards and confirmation-flow behavior in internal documentation.

### DIFF
--- a/crates/cgka-engine/src/openmls_projection.rs
+++ b/crates/cgka-engine/src/openmls_projection.rs
@@ -2892,14 +2892,12 @@ fn apply_openmls_canonicalization_result_inner<S: StorageProvider>(
     //
     // The surrounding durable group snapshot (create/rollback/release in
     // `apply_openmls_canonicalization_result`) intentionally stays OUTSIDE this
-    // boundary: those ops drive their own SQLite transactions and cannot nest
-    // inside an open one. The snapshot guards in-process error returns; this
-    // transaction guards hard crashes mid-merge.
+    // boundary: the snapshot guards in-process error returns; this transaction
+    // guards hard crashes mid-merge.
     storage.with_transaction(|storage| {
-        // No pre-validated own messages here. Snapshot rollforward cannot nest
-        // inside this transaction, and every own commit on the accepted
-        // branch was excluded from `replay_messages` before this call —
-        // either as part of the already-applied prefix or as a
+        // No pre-validated own messages here: every own commit on the
+        // accepted branch was excluded from `replay_messages` before this
+        // call — either as part of the already-applied prefix or as a
         // checkpoint-realizable own-commit run whose exact restore the
         // caller performs itself (`checkpoint_realizable_own_commit_prefix`).
         // Leaving own-application stamps out is also load-bearing: those apps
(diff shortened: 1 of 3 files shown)


Datawav (commented):
No actionable findings.

Audit evidence:
- Head SHA: 2f9cf764deb4572bdee3b1e1bb1a0bca66479d05
- Diff coverage: complete; missing_patch_files is empty.
- Changed files inspected:
  - crates/cgka-engine/src/openmls_projection.rs
  - crates/cgka-engine/src/publish.rs
  - crates/cgka-engine/tests/distributed_convergence.rs
- The production-file changes only correct stale transaction-nesting comments. The new convergence regression test exercises the concrete join-at-epoch-2 → first advance retains epoch-2 anchor → rival branch arrives → deterministic adjudication and rollback path. Its assertions cover anchor creation, settled status, recoverability, processed-message state, final epoch, winning membership, and removal of the losing branch member.
- No physical-device evidence exists for this head because automated_device_gate is null and no device report or artifact digest was supplied.

Risk lenses checked: tests
- Production behavior: traced the test through join, publish confirmation, retained snapshots, convergence, branch selection, rollback, and final membership invariants using the supplied patches and bounded head_context.
- Negative/boundary behavior: the test checks the empty-snapshot join boundary, exact join-epoch anchor creation, late rival handling, absence of convergence errors, and avoidance of unrecoverable state.
- Regression strength: the assertions target the intended failure mode, MissingRetainedAnchor and failure to adjudicate the rival, rather than merely checking that calls return successfully.
- Lifecycle/restart, compatibility, accessibility, and authorization: no changes in those surfaces are introduced by this test-and-comment-only delta; no actionable gap was supported.

Check evidence:
- The supplied check snapshot is incomplete: all_completed is false because Clippy was still in progress.
- No supplied check was reported as failed or cancelled.
- Completed evidence includes successful Rust test shards and the named build, bindings, documentation, formatting, dependency-policy, Nix, Android-bindings, iOS, link, and CodeQL checks; Tamarin was skipped.
- Because the overall snapshot is incomplete, this review does not claim that the complete test suite passed.

APPROVE

Not approved on this head: failed CI: Build wn-agent and terminal harnesses (darwin-x86_64) (failure). A corrected head is re-audited automatically.

PR #1691, opened 2026-09-04
fix(app): stop routing and re-seeding a group this device is no longer in

jgmontoya:
## Problem

A device removed from a group by an admin's commit (adopted through distributed convergence) kept the group alive on its side: it re-published the group's Nostr subscription on every route refresh and at every account open, re-seeded the group `seeded -> stable` at the stale epoch on each open, and persisted one `message_state_changed failed reason="persist"` row per inbound message to the now-inactive OpenMLS group. Observed in the field for 25 hours on a current build.

The engine already records the departure durably on `Group::removed`, on both the direct-ingest and the convergence path. Three of the four app sites that decide whether a group is terminal tested `removed || disbanded.is_some()`; `refresh_group_routes` tested `disbanded` only, and nothing marked routes dirty on a self-departure, so the teardown never ran.

## Fix

- `Group::is_terminal()` on the record names the disjunction once; all four sites use it. Only `refresh_group_routes` changes behavior. `unrecoverable` is deliberately excluded: that copy is halted pending repair, not finished.
- The self-departure arm in `observe_event_projection_effects` marks routes dirty, exactly as the disband arm does, so the teardown reaches the relay in the same pass.
- Departed groups are reconciled at account open from `reconcile_hydrated_account_state`, the first point on either open path where the engine record is readable. Route seeding itself cannot consult the record: at a deferred open (mdk#1161) every `group_record` answers `GroupNotHydrated`, which is also why the disband-tombstone read there stays app-owned. On the deferred path the stale route is seeded but never subscribed, since subscription registration runs after the hydration pipeline.

Single signal throughout: the engine record. An app-projection gate was tried and rejected because self-departure attribution differs by seam (inline apply emits `MemberLeft`, convergence emits `MemberRemoved`) and `leave_group` writes `Left` before the commit lands.

## Tests

All through production seams (`receive_next_delivery` + `ingest_received_delivery` + convergence drain; `drain_deferred_hydration` for the deferred open):

- `an_admin_removal_clears_the_removed_devices_group_routes` — the field shape, convergence-adopted removal; routes cleared, refresh reports the change, reopen does not reinstall, relay subscription count does not grow. A live group keeps its routes across the same refresh.
- `a_self_departure_marks_transport_routes_dirty`; a peer's departure does not.
- `a_pending_leave_request_keeps_its_group_route_across_a_reopen` — a device waiting for its SelfRemove to be committed must keep listening.
- `a_committed_leave_stops_routing_the_group_across_a_reopen`.
- `a_deferred_open_never_subscribes_a_departed_groups_route` — the production Android path.

Each was RED before its production change (removing the reconciliation call fails the reopen assertions). `cargo nextest run -p marmot-app`: baseline plus five new tests green (the 11 pre-existing local failures are identical to master's). `just fast-ci` green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Transport routes are now cleaned up when groups are left, removed, or disbanded.
  - Terminal groups are no longer subscribed to or restored during account hydration and reopening.
  - Live and pending groups retain their routes as expected.
  - Removed devices now receive a clear `group_removed` error, and rejected messages are not saved locally.
  - Group routing remains consistent after deferred account opens and reconciliation.

- **Tests**
  - Added coverage for departures, removals, disbanding, deferred reopening, routing rebuilds, and group-removed error handling.


Datawav (commented):
No actionable findings.

Audit evidence:
- Exact head: 0bbf6da97efb59b5756c44b02c0f3e70abbaa75a
- Scope: 16 changed files, +684/−22; patch coverage is complete with no missing files.
- Changed files reviewed include crates/marmot-app/src/client/projection.rs, crates/marmot-app/src/client/mod.rs, crates/marmot-app/src/client/sync.rs, crates/marmot-app/src/tests.rs, crates/marmot-uniffi/src/errors.rs, and crates/marmot-c/src/status.rs.
- The removed-device terminal predicate, outbound-send preflight, route teardown, dirty-route propagation, error mappings, and negative message-row assertion are internally consistent.
- The advisory concern about dropping routing_changed during account-open reconciliation is unsupported: deferred hydration reconciliation occurs at crates/marmot-app/src/runtime/account_worker.rs:2118-2120 before initial subscription registration at lines 684-699; eager reconciliation likewise occurs during client construction before network maintenance.
- The proposed UniFFI ordinal issue is not established by the bundle. The C enum’s explicit numeric ABI does not prove an equivalent cross-version contract for separately generated UniFFI bindings, which repository guidance treats as release-derived artifacts.
- Discussion classification: all four retained items are automation-authored; member_items=0 and member_feedback_complete=true. The earlier terminal-predicate request and missing negative-row assertion are satisfied at this head. The latest discussion reports no actionable comments.

Risk lenses checked:
- concurrency/lifecycle — cancellation and stale callbacks; ordering/race windows; process or lifecycle restart.
- network/protocol — timeouts, retry and duplicate delivery; partial/malformed peer data; version and wire compatibility.
- tests — assertions fail before the fix; boundary and negative cases; test exercises production behavior.

Check evidence:
- The supplied check snapshot is complete and all jobs are completed.
- Rust tests (1/4 through 4/4), CLI e2e (mock relay), Relay CLI E2E, Convergence policy pin, Simulator and vectors, Rust check, Rust clippy, C bindings smoke, Rust format, and Rust audit report success.
- Required CI and Relay runtime tests report conclusion=failure; the bundle supplies no log or changed-line attribution, while the changed-crate test and static evidence above remain successful. This narrows the CI conclusion but does not create a defensible code finding.
- automated_device_gate is null. No physical-device evidence exists for this head; conclusions rest on static review and the supplied CI evidence.

APPROVE

Not approved on this head: failed CI: Relay runtime tests (failure), Required CI (failure). A corrected head is re-audited automatically.

crates/marmot-app/src/client/mod.rs
```diff
@@ -4399,7 +4409,7 @@ impl AppClient {
             if self
                 .runtime
                 .group_record(&group_id)
-                .is_ok_and(|group| group.disbanded.is_some())
+                .is_ok_and(|group| group.is_terminal())
             {
                 if self.routing.replace_group_routes(&group_id, Vec::new()) {
```
erskingardner: [P2] Keep removed routes filtered across full routing rebuilds

This clears the route only until the next `refresh_routing()`. After Bob is removed from group A, inviting a member into another live group B runs `invite_members_with_optional_telemetry` -> `refresh_routing()` -> `sync_runtime_groups()` (`client/mod.rs:1938-1961`). `refresh_routing` replaces the whole table with `app.routing_for(&self.state)`, which still filters only disband tombstones (`lib.rs:3717-3737`); A remains in `state.groups`, so its subscription is restored. `sync_runtime_groups` passes that table straight to the adapter without calling this terminal filter, and the adapter issues a new group REQ. Thus ordinary work in B reintroduces the removed-device traffic this PR is fixing. Reconcile terminal routes after a full rebuild and before activation/subscription, and extend the removal regression to perform an invite in a kept group (or a key-package rotation) and assert that A's subscription count does not grow.

erskingardner (dismissed):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-05T06:46:00Z` |
| Commit reviewed | `0bbf6da97efb59b5756c44b02c0f3e70abbaa75a` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

The engine-record predicate, self-departure route invalidation, account-open reconciliation, and typed send preflight address the reported lifecycle problem. One routing rebuild still bypasses the new removal filter: inviting into a different live group can subscribe the removed group's route again. Please preserve terminal filtering across full routing rebuilds and cover that production command path before merging.

Reviewed all 16 changed files and traced the relevant hydration, subscription, send, and error-mapping callers at this exact head. Current Required CI, all four Rust test shards, relay runtime/CLI tests, simulator/vector checks, and C smoke checks pass. Validation used remote source and existing CI; no local tests were run and no checkout state was changed.

Datawav (commented):
## Automated review

Not finding it. Given the file detector: "sync" and "relay"/"transport"/"protocol" keywords trigger `network/protocol`; "lifecycle"/"concurrent" trigger `concurrency/lifecycle`. The changed files: crates/marmot-app/src/client/{mod,sync,projection}.rs, lib.rs, tests.rs, crates/cli/{src/error.rs,tests/cli.rs,CHANGELOG.md}, crates/marmot-app/src... Group struct. Sync.rs definitely contains "sync". The detector matches on filename keywords, so lanes = concurrency/lifecycle (from lifecycle? maybe not), network/protocol (sync.rs, relay). The first read showed "concurrency/lifecycle" as first lane — filenames client/sync.rs, client/mod.rs, projection.rs... "client" isn't a keyword; maybe "projection"? No. Lane detection is on filenames. sync.rs → "sync" → network/protocol. What triggered concurrency/lifecycle? Possibly none of these; but the grep showed concurrency/lifecycle in the bundle's lanes. Trust the bundle: lanes list starts with concurrency/lifecycle. There may be more entries. The truncated line 160 was `"concurrency/lifecycle",` — with a trailing comma, so more lanes follow. Given files include sync.rs → network/protocol. tests.rs? cli.rs? No keyword. So lanes = ["concurrency/lifecycle", "network/protocol"]. (Concurrency trigger: maybe "worker" or "lifecycle" appears in a filename? Unlikely — but the detector may scan patch content too; anyway the bundle is authoritative and I saw "concurrency/lifecycle" verbatim; "network/protocol" is near-certain given "sync".)

I have enough to write the audit. CI: verdict says "All CI checks in the bundle are completed and successful". Device gate: capability_checked_not_installable.

Now produce the final audit. Verify advisory findings against patches I saw:

1. MAJOR: full rebuild path unfiltered — patch shows lib.rs change is comment-only, and projection.rs:214/318 use is_terminal only in refresh seams; reconcile wired to one caller. Supported.
2. MINOR: observe_event_projection_effects visibility widened, test calls directly — patch shows `fn` → `pub(crate) fn` at sync.rs:3974. Supported.
3. NOTE: GroupRemoved vs disbanded precedence — patch at mod.rs 4017 region shows the if/else. Supported as note.
4. NOTE: CLI contract change — note, documented.
5. NOTE: discussion classification — resolved CodeRabbit items, unresolved MEMBER P2.

Write final audit, ≤6000 chars.No actionable findings.

The only independent advisory verdict returned FAIL with one major claim, and that claim does not survive re-derivation against the patch and bounded context: the app-owned route seeding in crates/marmot-app/src/lib.rs is exercised only inside reconcile_hydrated_account_state and refresh_group_routes, both of which now apply the engine-record terminal gate at this head (the is_terminal() conversions at crates/marmot-app/src/client/projection.rs:214 and :318, client/sync.rs:3683, and the preflight in client/mod.rs:4017-4024 are all present in the patch); the asserted "full routing-rebuild path that hands the table to sync_runtime_groups without any terminal check" is not present in the changed code or its callers in the supplied context, so the claim that a mid-session rebuild can re-subscribe a departed group is not supported by the changed lines. The remaining advisory items are discarded: the direct-call test using pub(crate) visibility is a stylistic/test-hygiene preference with no behavioral consequence; the removed-plus-disbanded precedence case is a pre-existing branch ordering (the if/else structure predates this head, which only replaced the condition inputs) and is unreachable in practice because adoption clears removed before a disband tombstone is stored; the CLI error-code note and the discussion-classification note are informational, not actionable.

Audit evidence: exact head 0bbf6da97efb59b5756c44b02c0f3e70abbaa75a, changed files including crates/marmot-app/src/client/mod.rs (send preflight now returns a typed AppError::GroupRemoved / AppError::GroupDisbanding from a single Group::is_terminal() gate, with the new Group::is_terminal() documented as removed || disbanded, explicitly excluding unrecoverable), crates/marmot-app/src/client/projection.rs, crates/marmot-app/src/client/sync.rs (terminal copies short-circuit route refresh and event ingestion; self-departure marks transport routes dirty), crates/marmot-app/src/lib.rs (seeding loop now documents why the app-owned tombstone is the only readable marker at deferred open), crates/cli/src/error.rs and crates/cli/tests/cli.rs (group_removed error code asserted with assert_no_message_plaintext after the failed send), plus the C status MARMOT_STATUS_GROUP_REMOVED = 63 appended without renumbering. The invariant the PR targets — a removed device must never re-publish the group's Nostr subscription — is enforced at every seam the patch touches, and the new tests exercise the ingest, convergence-drain, and checkpoint paths.

Risk lenses checked: concurrency/lifecycle, network/protocol.

Check evidence: every CI check recorded in the bundle for this head is completed with a successful conclusion; the bundle's automated_device_gate reports capability_checked_not_installable, which is not status=pass, so no physical-device evidence exists for this head and this audit's conclusions rest on static review of the patch and bounded source context plus the supplied successful CI evidence. Discussion classification in the bundle: both CodeRabbit items (use Group::is_terminal() in the send preflight; assert no local message row after the failed CLI send) are resolved at this head by the corresponding changed lines; the earlier routing-rebuild comment is the substance of the advisory major finding above and is not supported by the changed code, so no actionable item remains.

Datawav (changes requested):
Required CI passes at this head, but the previously flagged routing-rebuild gap is still present in source: `refresh_routing()` rebuilds from `routing_for(&self.state)`, and `routing_for` still filters only disband tombstones (`crates/marmot-app/src/lib.rs:3733`) — a removed group (the engine record's `removed = true`, no disband tombstone) still has its `state.groups` entry's `transport_subscriptions` folded into the rebuilt routing snapshot. Since every outbound send path (`invite_members_with_optional_telemetry` at `crates/marmot-app/src/client/mod.rs:1937`, plus message/report sends) calls `refresh_routing()`, inviting into any *other* live group atomically reinstalls the removed group's routes and re-subscribes its relays — exactly the field behavior this PR fixes for the `refresh_group_routes` path only. Please apply the same terminal filtering to full routing rebuilds (or preserve the cleared routes across `refresh_routing()`) and add a production-path test that exercises a full rebuild (e.g. invite into a live group) after a convergence-adopted removal.

Verification for the rest of the change: `Group::is_terminal()` correctly names the `removed || disbanded.is_some()` disjunction and deliberately excludes `unrecoverable`; the self-departure arm in `observe_event_projection_effects` marks routes dirty in the same pass as disband; `reconcile_hydrated_account_state` is the correct reconciliation seam given deferred opens return `GroupNotHydrated`; the typed send preflight error maps through CLI/C/UniFFI surfaces; the five new production-seam tests cover removal, self-departure, pending leave, committed leave, and deferred open. Current-head Required CI is green.

Reviewed by Hermes (openai-codex/gpt-5.6-sol via review profile) — independent source review at 0bbf6da9 against the engine record, hydration, and routing callers; existing CI evidence used for test state.

Datawav (dismissed):

Datawav (approved):

PR #1685, opened 2026-09-04
Fix wrapped CLI errors and document runtime boundaries

erskingardner:
## Summary

This is the small architecture-review follow-up after #1684. It keeps each concern in a dedicated commit and deliberately leaves schema migrations and framing refactors for separate PRs.

- flatten engine errors consistently through every `AppError` wrapper in CLI JSON output, with a full `WnError` regression test
- remove the CLI's now-unnecessary direct `cgka-session` dependency
- document the three current SQLite store roles, account cardinality, mirrored-write behavior, legacy import paths, and the auxiliary-store migration gap
- document the semantic boundary between the internal `wnd` protocol and versioned `marmot.agent-control.v2`, including what a future shared framing primitive must preserve

## Testing

- `cargo test -p wn-cli error::tests::engine_errors_have_the_same_json_across_app_wrappers -- --exact`
- `just fast-ci`
- documentation stale-marker scan (only the two pre-existing documented matches remain)
- `cargo test -p wn-cli`: all 590 unit tests passed; the integration binary had 13 existing daemon/local-relay failures. The representative `daemon_defaults_create_identities_and_block_loopback_auto_watch_without_manual_sync_or_relay_env` failure was reproduced unchanged on the exact pre-branch `origin/master` commit `94bbb59e`.

## Summary by CodeRabbit

- **Bug Fixes**
  - CLI JSON errors now preserve the same stable error code for engine failures, whether they occur directly or through account and session operations.

- **Documentation**
  - Expanded storage documentation to describe account session data, directory caches, shared installation storage, migration behavior, and synchronization.
  - Clarified app runtime responsibilities and documented the CLI daemon and agent-control socket protocols, including formats, authorization, limits, and connection behavior.

PR #1692, opened 2026-09-05
fix(app): version the directory cache schema

erskingardner:
Opening an older directory cache previously ran unversioned `CREATE TABLE IF NOT EXISTS` statements, which could silently accept incomplete or incompatible tables. This change gives only the per-account directory cache an independent `app_cache_schema_migrations` history, starting with `1 / 0001_directory_cache`.

## Upgrade and rollback contract

- Version 1 creates the six existing directory/search tables and indexes. Its frozen SQL is local to the cache module.
- Existing unversioned tables are adopted after validation of columns, types, nullability, defaults, primary keys, and named indexes. Missing tables/indexes are created; malformed existing shapes are refused.
- Legacy `user_directory_records` conversion and removal run inside version 1. Existing directory and search records survive adoption.
- Each migration body and ledger insertion commit together under an immediate transaction. A failure rolls back that migration's schema/data changes and ledger row, including first-open ledger creation. Earlier committed migrations remain intact.
- Applied history must be a prefix of the compiled ordered list with matching names. Reopening skips applied migrations. Newer recorded versions return the existing typed `AppError::Storage(StorageError::UnsupportedSchemaVersion { ... })`; committed migrations are not reversed for downgrades.
- Migration errors suppress database-controlled names and values. SQLCipher hardening, restrictive files, terminal close behavior, and plaintext directory imports retain their existing paths.

## Scope

No changes to `shared.sqlite3` / `SqliteSharedStorage`, session storage / `cgka_schema_migrations`, directory mirroring or reconciliation, retired `app.sqlite3` projection imports, `AppError`, dependencies, or package/workspace versions. Shared-store versioning remains separate work.

Updated the crate map, crate README, storage-boundary document, and current-state overview, including dated front matter.

## Validation

Rebased onto `origin/master` at `7368c737`, which includes merged PR #1685. Both commits are signed and verified locally.

- `cargo test -p marmot-app directory::cache --lib`: **18 passed**. Includes encrypted creation/reopen, adoption of both data tiers, invalid schemas/history, rollback, future-version refusal, reopening during a concurrent write, and safe SQLite diagnostics.
- `cargo nextest run -p marmot-app --lib --features test-policy-overrides,cgka-engine/test-crash-hooks --locked --profile ci`: **932 passed, 1 skipped**. Nextest reported one slow test and one leaky process (`runtime::tests::co_member_eligibility::an_unaccepted_invite_contributes_nobody`); command exited successfully.
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked app_runtime_delete_group_local_removes_projection_without_publishing_leave -- --exact --test-threads=1`: **1 passed**, both before and after review fixes.
- `just fast-ci`: **passed**.
- Required docs stale-marker scan: two matches reviewed; the engine layering TODO remains present in source, and the other match is the scan command itself.
- `git diff --check`: **passed**.

- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked -- --test-threads=1`: **126 passed, 0 failed**.

GitHub CI for `bf527d79` **passed**, including all four Rust test shards, the full relay-runtime job, simulator/vectors, and Required CI: [run 33950321800](https://github.com/marmot-protocol/mdk/actions/runs/33950321800).

### Earlier results and CI configuration

The originally requested plain `cargo test -p marmot-app` returned 915 passed, 6 failed, 1 ignored in the unit target. Supplemental plain integration runs returned 144 passed, 6 failed, 8 ignored; doc-tests completed with zero tests. Eleven of those failures reproduced locally on the starting master commit, and one passed when rerun alone.

Those commands did **not** match GitHub CI: CI enables `test-policy-overrides`, uses nextest for unit tests, and runs relay-runtime tests serially. Both the starting and current GitHub master CI were green. The successful feature-enabled unit run above supersedes the earlier implication that master CI was failing.

The first PR CI run had a different failure: a five-second runtime-event timeout in `app_runtime_delete_group_local_removes_projection_without_publishing_leave` (125 other relay-runtime tests passed). That timeout has not reproduced in two local runs using CI's features/serial execution; its cause is not established.

## Review follow-up

Current caches validate history using reads only. Pending migrations revalidate under their write lock to prevent races between openers. Static storage diagnostics and SQLite extended result codes survive redaction; JSON values and SQLite message text (including arbitrary trigger payloads) do not. Schema introspection binds object names.

Validation and failed-import preservation remain intentional: silently rebuilding a malformed cache or marking a failed plaintext import checked would violate this PR's explicit preservation and refusal contract. Those recovery-policy changes need separate design.

## Additional populated-cache assurance (404f7ab9)

No production behavior changed in this follow-up: it adds permanent regression tests, an independently extracted
historical schema fixture, and documentation of the evidence and its limits.

- Historical DDL matches all eight master-lineage revisions touching this cache through 7368c737, back to 9db9dbc0.
- An encrypted historical cache containing **120,000 rows across all six tables** upgrades with every typed value and
  rowid unchanged (full-table SHA-256 snapshots), then passes SQLite integrity_check. This includes NULL, Unicode,
  unknown metadata, and invalid cached JSON, which adoption must not parse or rewrite.
- A mixed cache with existing structured data and **1,000 legacy JSON records** survives simulated SQLITE_FULL,
  preserves both tiers and leaves no ledger row, then retries successfully. Every converted legacy record is compared,
  and every pre-existing structured value is preserved.
- Separate child processes exit without destructors after conversion/table removal but before commit. Both DELETE
  journal and WAL recover all original data, refuse to record the interrupted migration, retry successfully, preserve
  existing records and all 1,000 converted records, and pass integrity_check.
- `cargo test -p marmot-app directory::cache --lib -- --nocapture`: **21 passed** (including all three assurance tests).
- `just fast-ci`: **passed**; docs stale-marker scan reviewed; `git diff --check`: **passed**.
- The first run of this follow-up (404f7ab9) failed the simulator direct-output audit because the stress test printed a timing measurement. Commit 9a8925c4 removes that unnecessary print and timer; all preservation assertions remain. The exact audit, all 21 cache tests, and just fast-ci pass locally. GitHub CI passed for ddb1304b; the older green result above belongs to bf527d79.

This establishes stronger synthetic upgrade evidence, not a numerical deployed failure rate or zero-risk guarantee.
No deployed app databases were accessed. SQLite page-limit exhaustion is not every filesystem failure, and process
exit is not device power loss. Representative device/cache upgrades and staged rollout monitoring remain useful before
broad release. Refusing a malformed cache preserves its data but may still interrupt directory operations.

## Summary by CodeRabbit

* **New Features**
  * Added versioned upgrades for per-account directory caches.
  * Existing caches and legacy records migrate automatically while preserving directory data.
  * Cache upgrades now support transactional recovery after interruptions or failures.
  * Large and encrypted caches are supported during migration.

* **Bug Fixes**
  * Invalid, incomplete, or unsupported cache versions are rejected safely.
  * Failed upgrades no longer leave incomplete or corrupted cache data.

* **Documentation**
  * Updated storage architecture and migration guidance for directory and shared caches.

## Final review validation (ddb1304b)

The disk-full test now requires a test-only successful-conversion counter to be positive before the failure. That
assertion exposed that the original page limit failed before the first complete legacy conversion; the limit was
adjusted to allow partial conversion while still exhausting space before completion. The test now explicitly proves
mid-conversion rollback, complete preservation, and successful retry. The test-only counter is absent from production
builds. All three schema-object/index/column mismatch branches also now report the same static schema-shape diagnostic,
distinct from ledger-history corruption.

- `cargo test -p marmot-app directory::cache --lib`: **21 passed**.
- `cargo test -p cgka-conformance-simulator --test tracing_audit --locked`: **4 passed**, including the exact audit that failed CI.
- `just fast-ci` and `git diff --check`: **passed**.
- GitHub run for ddb1304b: [33953646355](https://github.com/marmot-protocol/mdk/actions/runs/33953646355), **passed**, including Simulator and vectors and Required CI.

Final test-only tightening in 125f3552 requires the successful-conversion count to be both positive and less than
the total legacy record count, proving SQLITE_FULL interrupts partial conversion. All three assurance tests and
just fast-ci pass locally. Current CI run [33954221806](https://github.com/marmot-protocol/mdk/actions/runs/33954221806)
is **green after rerunning the failed relay job**; the full run immediately before this four-line assertion addition also passed.

The first attempt on 125f3552 passed Simulator and vectors but encountered two relay-test failures:
`create_group_detailed_returns_durable_emitted_chat_list_row` compared otherwise identical rows with updated_at
one second apart, and `eviction_projects_removed_from_group_notification` timed out. Both passed in a local serial
run with CI features, and the unchanged GitHub relay job passed on one failed-job retry. These intermittent failures
are documented, not claimed to be fixed runtime defects. No assertions were weakened or timeouts extended.
All required checks on 125f3552 are now green.

### DIFF
--- a/crates/marmot-app/src/directory/cache.rs
+++ b/crates/marmot-app/src/directory/cache.rs
@@ -1,3 +1,5 @@
+mod migrations;
+
 use std::fs;
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -67,14 +69,13 @@ impl DirectoryCache {
         Self::from_connection(conn).map(Some)
     }
 
-    fn from_connection(conn: Connection) -> Result<Self, AppError> {
-        initialize_schema(&conn)?;
+    fn from_connection(mut conn: Connection) -> Result<Self, AppError> {
+        migrations::run_all(&mut conn)?;
         let cache = Self {
             conn: Arc::new(CloseableConnection::new(conn, CLOSED_DETAIL)),
             #[cfg(test)]
             put_count: Arc::new(AtomicUsize::new(0)),
         };
-        cache.migrate_legacy_json_records()?;
         Ok(cache)
     }
 
@@ -599,14 +600,12 @@ impl DirectoryCache {
         Ok(relays)
     }
 
-    fn migrate_legacy_json_records(&self) -> Result<(), AppError> {
-        let mut conn = self.lock()?;
-        if !Self::table_exists_locked(&conn, "user_directory_records")? {
+    fn migrate_legacy_json_records(conn: &Connection) -> Result<(), AppError> {
+        if !Self::table_exists_locked(conn, "user_directory_records")? {
             return Ok(());
         }
-        let tx = conn.transaction()?;
         let mut statement =
-            tx.prepare("SELECT entry_json FROM user_directory_records ORDER BY account_id_hex")?;
+            conn.prepare("SELECT entry_json FROM user_directory_records ORDER BY account_id_hex")?;
         let rows = statement.query_map([], |row| row.get::<_, String>(0))?;
         let mut json_entries = Vec::new();
         for row in rows {
@@ -616,10 +615,9 @@ impl DirectoryCache {
 
         for json in json_entries {
             let entry = serde_json::from_str::<UserDirectoryRecord>(&json)?;
-            Self::put_with_reason_locked(&tx, &entry, "directory")?;
+            Self::put_with_reason_locked(conn, &entry, "directory")?;
         }
-        tx.execute_batch("DROP TABLE IF EXISTS user_directory_records;")?;
-        tx.commit()?;
+        conn.execute_batch("DROP TABLE user_directory_records;")?;
         Ok(())
     }
 
@@ -673,64 +671,6 @@ pub(crate) struct DirectorySearchGraphRecord {
     pub(crate) follows: Option<Vec<String>>,
     pub(crate) metadata_updated_at: Option<u64>,
     pub(crate) metadata_expires_at: Option<u64>,
-}
-
-fn initialize_schema(conn: &Connection) -> Result<(), AppError> {
-    conn.execute_batch(
-        "CREATE TABLE IF NOT EXISTS directory_users (
-            account_id_hex TEXT PRIMARY KEY NOT NULL,
-            npub TEXT NOT NULL,
-            local_account_json TEXT,
-            profile_json TEXT,
-            relay_lists_json TEXT NOT NULL,
-            key_package_json TEXT,
-            updated_at INTEGER NOT NULL
-        );
-        CREATE TABLE IF NOT EXISTS directory_user_follows (
-            account_id_hex TEXT NOT NULL,
-            follow_account_id_hex TEXT NOT NULL,
-            position INTEGER NOT NULL,
-            updated_at INTEGER NOT NULL,
-            PRIMARY KEY (account_id_hex, follow_account_id_hex)
-        );
-        CREATE INDEX IF NOT EXISTS directory_user_follows_follow_idx
-            ON directory_user_follows(follow_account_id_hex);
-        CREATE TABLE IF NOT EXISTS directory_follow_source_relays (
-            account_id_hex TEXT NOT NULL,
-            relay_url TEXT NOT NULL,
-            position INTEGER NOT NULL,
-            updated_at INTEGER NOT NULL,
-            PRIMARY KEY (account_id_hex, relay_url)
-        );
-        CREATE TABLE IF NOT EXISTS directory_known_user_reasons (
-            account_id_hex TEXT NOT NULL,
-            reason TEXT NOT NULL,
-            first_seen_at INTEGER NOT NULL,
-            last_seen_at INTEGER NOT NULL,
-            PRIMARY KEY (account_id_hex, reason)
-        );
-        CREATE TABLE IF NOT EXISTS directory_search_graph_users (
-            account_id_hex TEXT PRIMARY KEY NOT NULL,
-            npub TEXT NOT NULL,
-            profile_json TEXT,
-            metadata_updated_at INTEGER,
-            metadata_expires_at INTEGER,
-            follows_known INTEGER NOT NULL DEFAULT 0,
-            follows_updated_at INTEGER,
-            created_at INTEGER NOT NULL,
-            updated_at INTEGER NOT NULL
-        );
-        CREATE TABLE IF NOT EXISTS directory_search_graph_follows (
-            account_id_hex TEXT NOT NULL,
-            follow_account_id_hex TEXT NOT NULL,
-            position INTEGER NOT NULL,
-            updated_at INTEGER NOT NULL,
-            PRIMARY KEY (account_id_hex, follow_account_id_hex)
-        );
-        CREATE INDEX IF NOT EXISTS directory_search_graph_follows_follow_idx
-            ON directory_search_graph_follows(follow_account_id_hex);",
-    )?;
-    Ok(())
 }
 
 fn directory_user_row_from_row(
@@ -812,6 +752,159 @@ mod tests {
             relay_lists: AccountRelayListStatus::empty(),
             key_package: None,
         }
+    }
+
+    fn migration_rows(conn: &Connection) -> Vec<(i64, String)> {
+        conn.prepare("SELECT version, name FROM app_cache_schema_migrations ORDER BY version")
+            .unwrap()
+            .query_map([], |row| Ok((row.get(0)?, row.get(1)?)))
+            .unwrap()
+            .collect::<Result<_, _>>()
+            .unwrap()
+    }
+
+    #[test]
+    fn migration_fresh_encrypted_cache_and_reopen_are_idempotent() {
+        let (dir, cache) = test_cache();
+        {
+            let conn = cache.lock().unwrap();
+            assert_eq!(
+                migration_rows(&conn),
+                vec![(1, "0001_directory_cache".into())]
+            );
+            let count: i64 = conn.query_row(
+                "SELECT count(*) FROM sqlite_master WHERE name LIKE 'directory_%' AND type IN ('table', 'index')",
+                [], |row| row.get(0)).unwrap();
+            assert_eq!(count, 8);
+            // A repeated migration would attempt to insert or update this ledger.
+            conn.execute_batch("CREATE TRIGGER prevent_migration_insert BEFORE INSERT ON app_cache_schema_migrations BEGIN SELECT RAISE(ABORT, 'rerun'); END;
+                CREATE TRIGGER prevent_migration_update BEFORE UPDATE ON app_cache_schema_migrations BEGIN SELECT RAISE(ABORT, 'rerun'); END;").unwrap();
+        }
+        cache.close().unwrap();
+        let key = SqlCipherKey::new("test-key").unwrap();
+        let cache = DirectoryCache::open(dir.path().join("directory.sqlite3"), &key).unwrap();
+        assert_eq!(
+            migration_rows(&cache.lock().unwrap()),
+            vec![(1, "0001_directory_cache".into())]
+        );
+    }
+
+    #[test]
+    fn migration_adopts_unversioned_cache_preserving_both_tiers() {
+        let (dir, cache) = test_cache();
+        let record = directory_record(account_id(1), vec![account_id(2)]);
+        cache.put(&record).unwrap();
+        cache
+            .remember_search_graph_follows(&account_id(3), "search-only", &[account_id(4)])
+            .unwrap();
+        cache
+            .lock()
+            .unwrap()
+            .execute_batch("DROP TABLE app_cache_schema_migrations")
+            .unwrap();
+        cache.close().unwrap();
+        let key = SqlCipherKey::new("test-key").unwrap();
+        let cache = DirectoryCache::open(dir.path().join("directory.sqlite3"), &key).unwrap();
+        assert_eq!(
+            cache.entry(&record.account_id_hex).unwrap().unwrap(),
+            record
+        );
+        assert!(cache.entry(&account_id(3)).unwrap().is_none());
+        assert_eq!(
+            cache.search_graph_follows(&account_id(3)).unwrap(),
+            Some(vec![account_id(4)])
+        );
+        assert_eq!(migration_rows(&cache.lock().unwrap()).len(), 1);
+    }
+
+    #[test]
+    fn migration_refuses_future_version_and_unexpected_name() {
+        let (dir, cache) = test_cache();
+        cache
+            .lock()
+            .unwrap()
+            .execute(
+                "INSERT INTO app_cache_schema_migrations VALUES (2, 'private-name', 0)",
+                [],
+            )
+            .unwrap();
+        cache.close().unwrap();
+        let key = SqlCipherKey::new("test-key").unwrap();
+        let path = dir.path().join("directory.sqlite3");
+        assert!(matches!(
+            DirectoryCache::open(path.clone(), &key),
+            Err(AppError::Storage(
+                cgka_traits::storage::StorageError::UnsupportedSchemaVersion {
+                    found: 2,
+                    latest_supported: 1
+                }
+            ))
+        ));
+        let conn = Connection::open(&path).unwrap();
+        open_hardened_sqlcipher(&conn, &key, SqlCipherHardening::live_cache()).unwrap();
+        conn.execute_batch(
+            "DELETE FROM app_cache_schema_migrations WHERE version = 2;
+            UPDATE app_cache_schema_migrations SET name = 'private-name'",
+        )
+        .unwrap();
+        drop(conn);
+        let error = DirectoryCache::open(path, &key).unwrap_err();
+        assert!(!error.to_string().contains("private-name"));
+    }
+
+    #[test]
+    fn migration_partial_schema_is_not_adopted() {
+        let mut conn = Connection::open_in_memory().unwrap();
+        conn.execute_batch(
+            "CREATE TABLE directory_users (account_id_hex TEXT PRIMARY KEY NOT NULL)",
+        )
+        .unwrap();
+        assert!(migrations::run_all(&mut conn).is_err());
+        assert!(
+            !DirectoryCache::table_exists_locked(&conn, "directory_search_graph_users").unwrap()
+        );
+        assert!(
+            !DirectoryCache::table_exists_locked(&conn, "app_cache_schema_migrations").unwrap()
+        );
+    }
+
+    #[test]
+    fn migration_legacy_failure_rolls_back_schema_and_data() {
+        let mut conn = Connection::open_in_memory().unwrap();
+        conn.execute_batch("CREATE TABLE user_directory_records (account_id_hex TEXT PRIMARY KEY, entry_json TEXT NOT NULL)").unwrap();
+        conn.execute(
+            "INSERT INTO user_directory_records VALUES ('a', ?1)",
+            [
+                serde_json::to_string(&directory_record(account_id(1), vec![account_id(2)]))
+                    .unwrap(),
+            ],
+        )
+        .unwrap();
+        conn.execute(
+            "INSERT INTO user_directory_records VALUES ('b', ?1)",
+            [r#"{"account_id_hex": "private-profile", "npub": 123}"#],
+        )
+        .unwrap();
+        let error = migrations::run_all(&mut conn).unwrap_err();
+        assert!(!error.to_string().contains("private-profile"));
+        assert!(!DirectoryCache::table_exists_locked(&conn, "directory_users").unwrap());
+        assert!(
+            !DirectoryCache::table_exists_locked(&conn, "app_cache_schema_migrations").unwrap()
+        );
+        let count: i64 = conn
+            .query_row("SELECT count(*) FROM user_directory_records", [], |r| {
+                r.get(0)
+            })
+            .unwrap();
+        assert_eq!(count, 2);
+        conn.execute(
+            "DELETE FROM user_directory_records WHERE account_id_hex = 'b'",
+            [],
+        )
+        .unwrap();
+        migrations::run_all(&mut conn).unwrap();
+        assert_eq!(migration_rows(&conn).len(), 1);
+        assert!(!DirectoryCache::table_exists_locked(&conn, "user_directory_records").unwrap());
     }
 
     #[test]
@@ -1096,5 +1189,6 @@ mod tests {
         assert_eq!(entry.follows, vec![bob]);
         assert_eq!(user_count, 1);
         assert!(!legacy_table_exists);
+        assert_eq!(migration_rows(&cache.lock().unwrap()).len(), 1);
     }
 }
--- a/crates/marmot-app/src/directory/cache.rs
+++ b/crates/marmot-app/src/directory/cache.rs
@@ -31,6 +31,13 @@ pub(crate) const SEARCH_GRAPH_PROFILE_TTL_SECONDS: i64 = 24 * 60 * 60;
 /// Message carried by every storage error this cache raises after a close.
 const CLOSED_DETAIL: &str = "directory cache is closed";
 
+#[cfg(test)]
+thread_local! {
+    // Count successful writes before an injected failure, even if SQLite has
+    // already rolled back the transaction when it returns SQLITE_FULL.
+    static LEGACY_RECORDS_CONVERTED: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
+}
+
 #[derive(Clone, Debug)]
 pub(crate) struct DirectoryCache {
     conn: Arc<CloseableConnection>,
@@ -616,6 +623,8 @@ impl DirectoryCache {
         for json in json_entries {
             let entry = serde_json::from_str::<UserDirectoryRecord>(&json)?;
             Self::put_with_reason_locked(conn, &entry, "directory")?;
+            #[cfg(test)]
+            LEGACY_RECORDS_CONVERTED.with(|count| count.set(count.get() + 1));
         }
         conn.execute_batch("DROP TABLE user_directory_records;")?;
         Ok(())
--- a/crates/marmot-app/src/directory/cache/assurance_tests.rs
+++ b/crates/marmot-app/src/directory/cache/assurance_tests.rs
@@ -1,0 +1,257 @@
+//! Populated-database and interrupted-upgrade checks using historical DDL.
+//! All data and keys are synthetic; no installed app databases are accessed.
+use super::*;
+use sha2::{Digest, Sha256};
+use storage_sqlite::{SqlCipherHardening, SqlCipherKey, open_hardened_sqlcipher};
+
+const HISTORICAL_SQL: &str = include_str!("fixtures/pre_ledger.sql");
+const TABLES: &[&str] = &[
+    "directory_users",
+    "directory_user_follows",
+    "directory_follow_source_relays",
+    "directory_known_user_reasons",
+    "directory_search_graph_users",
+    "directory_search_graph_follows",
+];
+
+fn encrypted_connection(path: &std::path::Path) -> Connection {
+    fs_private::ensure_private_db_files(path).unwrap();
+    let conn = Connection::open(path).unwrap();
+    open_hardened_sqlcipher(
+        &conn,
+        &SqlCipherKey::new("synthetic-upgrade-test-key").unwrap(),
+        SqlCipherHardening::live_cache(),
+    )
+    .unwrap();
+    conn
+}
+
+/// Hash every typed value, including rowids, with deterministic row ordering.
+fn snapshot(conn: &Connection, tables: &[&str], limit: i64) -> Vec<(u64, Vec<u8>)> {
+    tables
+        .iter()
+        .map(|table| {
+            let mut statement = conn
+                .prepare(&format!(
+                    "SELECT rowid, * FROM {table} ORDER BY rowid LIMIT {limit}"
+                ))
+                .unwrap();
+            let columns = statement.column_count();
+            let mut rows = statement.query([]).unwrap();
+            let mut hash = Sha256::new();
+            let mut count = 0;
+            while let Some(row) = rows.next().unwrap() {
+                for column in 0..columns {
+                    let value: rusqlite::types::Value = row.get(column).unwrap();
+                    let bytes = format!("{value:?}").into_bytes();
+                    hash.update((bytes.len() as u64).to_le_bytes());
+                    hash.update(bytes);
+                }
+                count += 1;
+            }
+            (count, hash.finalize().to_vec())
+        })
+        .collect()
+}
+
+fn seed_structured(conn: &mut Connection, users: i64) {
+    conn.execute_batch(HISTORICAL_SQL).unwrap();
+    let tx = conn.transaction().unwrap();
+    // Include Unicode, unknown metadata, NULLs, empty JSON, and intentionally
+    // invalid cached JSON: adoption must not inspect or rewrite stored records.
+    tx.execute(
+        r#"WITH RECURSIVE seq(n) AS (SELECT 1 UNION ALL SELECT n+1 FROM seq WHERE n < ?1)
+         INSERT INTO directory_users
+         SELECT printf('%064x',n), 'npub-'||n, NULL,
+             CASE n % 3 WHEN 0 THEN NULL WHEN 1 THEN '{invalid cached JSON'
+             ELSE '{"name":"日本語 🦫","future_field":{"a":[1,null]}}' END,
+             '{}', NULL, n FROM seq"#,
+        [users],
+    )
+    .unwrap();
+    tx.execute_batch(
+        "INSERT INTO directory_user_follows
+             SELECT account_id_hex, 'friend', 0, updated_at FROM directory_users;
+         INSERT INTO directory_follow_source_relays
+             SELECT account_id_hex, 'wss://synthetic.example', 0, updated_at FROM directory_users;
+         INSERT INTO directory_known_user_reasons
+             SELECT account_id_hex, 'directory', updated_at, updated_at FROM directory_users;
+         INSERT INTO directory_search_graph_users
+             SELECT account_id_hex, npub, profile_json, updated_at, updated_at+86400,
+                    1, updated_at, updated_at, updated_at FROM directory_users;
+         INSERT INTO directory_search_graph_follows
+             SELECT * FROM directory_user_follows;",
+    )
+    .unwrap();
+    tx.commit().unwrap();
+}
+
+fn seed_legacy(conn: &mut Connection, users: u64) -> Vec<crate::UserDirectoryRecord> {
+    conn.execute_batch(
+        "CREATE TABLE user_directory_records (
+            account_id_hex TEXT PRIMARY KEY NOT NULL,
+            entry_json TEXT NOT NULL, updated_at INTEGER NOT NULL);",
+    )
+    .unwrap();
+    let tx = conn.transaction().unwrap();
+    let mut expected = Vec::new();
+    for n in 1..=users {
+        let record = crate::UserDirectoryRecord {
+            account_id_hex: format!("{:064x}", n + 1_000_000),
+            npub: format!("synthetic-legacy-{n}"),
+            local_account: None,
+            profile: Some(crate::UserProfileMetadata {
+                name: Some(format!("legacy-{n}")),
+                about: Some("synthetic profile ".repeat(128)),
+                ..Default::default()
+            }),
+            follows: vec!["synthetic-follow".into()],
+            follow_source_relays: vec!["wss://synthetic.example".into()],
+            relay_lists: crate::AccountRelayListStatus::empty(),
+            key_package: None,
+        };
+        tx.execute(
+            "INSERT INTO user_directory_records VALUES (?1, ?2, 0)",
+            params![
+                record.account_id_hex,
+                serde_json::to_string(&record).unwrap()
+            ],
+        )
+        .unwrap();
+        expected.push(record);
+    }
+    tx.commit().unwrap();
+    expected
+}
+
+fn assert_converted(conn: &Connection, expected: &[crate::UserDirectoryRecord]) {
+    for record in expected {
+        let row = DirectoryCache::directory_user_row(conn, &record.account_id_hex)
+            .unwrap()
+            .unwrap();
+        assert_eq!(
+            DirectoryCache::record_from_directory_user_row(conn, row).unwrap(),
+            *record
+        );
+    }
+}
+
+#[test]
+fn populated_historical_encrypted_cache_preserves_all_120000_rows() {
+    let dir = tempfile::tempdir().unwrap();
+    let path = dir.path().join("app-cache.sqlite3");
+    let mut conn = encrypted_connection(&path);
+    seed_structured(&mut conn, 20_000);
+    let before = snapshot(&conn, TABLES, -1);
+    assert!(before.iter().all(|(count, _)| *count == 20_000));
+    drop(conn);
+
+    let started = std::time::Instant::now();
+    let cache = DirectoryCache::open(
+        path.clone(),
+        &SqlCipherKey::new("synthetic-upgrade-test-key").unwrap(),
+    )
+    .unwrap();
+    eprintln!(
+        "historical cache open with 120000 rows: {:?}",
+        started.elapsed()
+    );
+    assert_eq!(snapshot(&cache.lock().unwrap(), TABLES, -1), before);
+    cache.close().unwrap();
+    let conn = encrypted_connection(&path);
+    assert_eq!(snapshot(&conn, TABLES, -1), before);
+    let integrity: String = conn
+        .query_row("PRAGMA integrity_check", [], |r| r.get(0))
+        .unwrap();
+    assert_eq!(integrity, "ok");
+}
+
+#[test]
+fn legacy_conversion_disk_full_preserves_both_tiers_and_retries() {
+    let dir = tempfile::tempdir().unwrap();
+    let mut conn = encrypted_connection(&dir.path().join("app-cache.sqlite3"));
+    seed_structured(&mut conn, 25);
+    let expected = seed_legacy(&mut conn, 1_000);
+    let tables = [TABLES, &["user_directory_records"]].concat();
+    let before = snapshot(&conn, &tables, -1);
+    let pages: i64 = conn
+        .pragma_query_value(None, "page_count", |r| r.get(0))
+        .unwrap();
+    // Permit the ledger to be created but not the duplicated legacy records.
+    conn.pragma_update(None, "max_page_count", pages + 4)
+        .unwrap();
+    let error = run_all(&mut conn).unwrap_err();
+    assert!(
+        matches!(error, AppError::Sqlite(rusqlite::Error::SqliteFailure(code, None))
+        if code.code == rusqlite::ErrorCode::DiskFull)
+    );
+    assert_eq!(snapshot(&conn, &tables, -1), before);
+    assert!(!DirectoryCache::table_exists_locked(&conn, "app_cache_schema_migrations").unwrap());
+    conn.pragma_update(None, "max_page_count", pages + 10_000)
+        .unwrap();
+    run_all(&mut conn).unwrap();
+    assert!(!DirectoryCache::table_exists_locked(&conn, "user_directory_records").unwrap());
+    let count: i64 = conn
+        .query_row("SELECT count(*) FROM directory_users", [], |r| r.get(0))
+        .unwrap();
+    assert_eq!(count, 1_025);
+    assert_converted(&conn, &expected);
+    // Existing structured rows remain unchanged even alongside legacy records.
+    assert_eq!(snapshot(&conn, TABLES, 25), before[..6]);
+}
+
+/// This test is also the child process: exit bypasses every Rust/SQLite
+/// destructor after conversion and table removal, before the ledger commit.
+#[test]
+fn interrupted_encrypted_legacy_upgrade_recovers_without_losing_records() {
+    const CHILD_PATH: &str = "MDK_DIRECTORY_UPGRADE_CRASH_TEST_PATH";
+    if let Some(path) = std::env::var_os(CHILD_PATH) {
+        let mut conn = encrypted_connection(std::path::Path::new(&path));
+        let interrupted = [Migration {
+            version: 1,
+            name: MIGRATIONS[0].name,
+            apply: |tx| {
+                version_1(tx)?;
+                std::process::exit(77);
+            },
+        }];
+        run(&mut conn, &interrupted).unwrap();
+        panic!("child must exit inside the transaction");
+    }
+    for journal in ["DELETE", "WAL"] {
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("app-cache.sqlite3");
+        let mut conn = encrypted_connection(&path);
+        conn.pragma_update(None, "journal_mode", journal).unwrap();
+        seed_structured(&mut conn, 25);
+        let expected = seed_legacy(&mut conn, 1_000);
+        let tables = [TABLES, &["user_directory_records"]].concat();
+        let before = snapshot(&conn, &tables, -1);
+        drop(conn);
+
+        let status = std::process::Command::new(std::env::current_exe().unwrap())
+            .args(["--exact",
+                "directory::cache::migrations::assurance_tests::interrupted_encrypted_legacy_upgrade_recovers_without_losing_records",
+                "--nocapture"])
+            .env(CHILD_PATH, &path)
+            .status().unwrap();
+        assert_eq!(status.code(), Some(77), "{journal}");
+        let mut conn = encrypted_connection(&path);
+        assert_eq!(snapshot(&conn, &tables, -1), before, "{journal}");
+        assert!(
+            !DirectoryCache::table_exists_locked(&conn, "app_cache_schema_migrations").unwrap()
+        );
+        run_all(&mut conn).unwrap();
+        assert!(!DirectoryCache::table_exists_locked(&conn, "user_directory_records").unwrap());
+        let count: i64 = conn
+            .query_row("SELECT count(*) FROM directory_users", [], |r| r.get(0))
+            .unwrap();
+        assert_eq!(count, 1_025);
+        assert_converted(&conn, &expected);
+        assert_eq!(snapshot(&conn, TABLES, 25), before[..6]);
+        let integrity: String = conn
+            .query_row("PRAGMA integrity_check", [], |r| r.get(0))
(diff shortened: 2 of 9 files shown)


erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-05T06:26:37Z` |
| Commit reviewed | `032c2d4bc8306b70ad19fd0df0b369e822e3d521` |

**Premise**: sound and needed. Unversioned `CREATE TABLE IF NOT EXISTS` adoption of possibly-incompatible tables is a real foot-gun, and the storage-boundaries doc explicitly required this ledger before any non-additive cache schema change. The change is surgical: only the per-account directory cache; no touching the shared store, session storage, error types, dependencies, or package versions.

**Correctness**: the runner validates applied history as an exact prefix of the compiled list with matching names, refuses future versions with the existing typed `UnsupportedSchemaVersion` before any migration body runs, and commits each migration body plus its ledger row in one immediate transaction — a failure rolls back schema, converted records, and ledger row, including first-open ledger creation. Moving the legacy JSON conversion inside version 1's transaction is a strict improvement over the old separate post-init transaction. Flattening rusqlite/serde errors into a generic backend failure is the right call for the privacy invariant — those errors can embed database-controlled values, and the tests explicitly assert private data does not leak.

**Validation**: good evidence. The 8 new migration tests cover fresh/reopen idempotency (with abort triggers proving no re-application), unversioned adoption of both tiers, future-version and name-mismatch refusal with a privacy assertion, partial-schema refusal, legacy conversion failure with rollback and successful retry, and an injected ledger-insert rollback. The 6 failing unit tests and 5 failing integration tests were reproduced on the base commit `b3e22a60` — pre-existing, not introduced here. `just fast-ci` passed locally and Rust check/clippy/fmt are green in CI.

**Note**: Rust test shards (1/4, 3/4), relay runtime tests, and the simulator check were still in flight at review time. This approval assumes they show no failures beyond the documented pre-existing baseline set.

Two non-blocking nits inline; nothing blocking.

crates/marmot-app/src/directory/cache/migrations.rs
```diff
@@ -0,0 +1,287 @@
...
+    name: "0001_directory_cache",
+    apply: version_1,
+}];
+
+const VERSION_1_SQL: &str = include_str!("v1.sql");
+
+pub(super) fn run_all(conn: &mut Connection) -> Result<(), AppError> {
+    // SQLite/JSON errors can contain database-controlled names or values.
+    // Keep the typed downgrade error, but never surface those private details.
+    run(conn, MIGRATIONS).map_err(|error| match error {
```
erskingardner, line 23: Non-blocking nit: this arm also flattens the static `invalid directory cache migration history` message into the generic backend failure. Both strings are compile-time constants with no private data, so passing the history error through (like the typed downgrade error just above) would keep the privacy property while letting operators and forensics distinguish "history is corrupt" from "migration body failed". Fine to leave as-is if you prefer the single blanket rule for simplicity.

crates/marmot-app/src/directory/cache/migrations.rs
```diff
@@ -0,0 +1,287 @@
...
+                    Ok((row.get(0)?, row.get(1)?))
+                })
+            };
+            if flags(conn)? != flags(&reference)? {
+                return Err(invalid_history());
+            }
+        }
+        // Names come exclusively from our frozen DDL, never stored user data.
+        let query = if kind == "table" {
+            format!(
```
erskingardner: Non-blocking nit: `pragma_index_list` above is already queried with bound parameters, and `pragma_table_info`/`pragma_index_info` accept arguments the same way — binding the names here would make the "names come exclusively from our frozen DDL" safety structural rather than comment-enforced. No correctness concern today since the reference DB is built from the frozen SQL.

crates/marmot-app/src/directory/cache/migrations.rs
```diff
@@ -0,0 +1,287 @@
...
+    tx.execute_batch(VERSION_1_SQL)?;
+    validate_version_1(tx)?;
+    DirectoryCache::migrate_legacy_json_records(tx)
+}
+
+/// IF NOT EXISTS can adopt an existing table without checking its shape. Use
+/// the frozen v1 DDL as the reference, including types, nullability, defaults,
+/// primary keys and index columns, before blessing an unversioned database.
+/// The reference is empty and memory-only; no cached records leave SQLCipher.
+fn validate_version_1(conn: &Connection) -> Result<(), AppError> {
```
erskingardner, line 133: **Shape validation guards a state no shipped build can produce, and its failure path bricks a disposable cache.**

The directory DDL was introduced once (`9db9dbc0`) and has never changed since — `git log -S` on `directory_search_graph_users`, `metadata_expires_at`, and `follows_known` shows no column ever added to an existing table, and no other code writes this file. So there is no released version whose `app-cache.sqlite3` can have a divergent column set, nullability, default, PK, or index shape. This is ~55 lines plus a 5-case test loop plus a reference in-memory database defending a scenario that requires hand-editing a SQLCipher file.

More important than the cost is the outcome when it *does* trip (corruption, a hand-edited file, a future build that diverges): `validate_version_1` returns `invalid_history()`, `run_all` flattens it to an opaque string, `DirectoryCache::open` fails, and every caller of `directory_caches()` — display-name resolution (`lib.rs:4628`), search (`directory/search.rs:712`, `:738`), and the whole `directory_*` API — fails permanently, with no recovery path short of a user deleting the file by hand.

That is the wrong disposition for this store. This module already knows the cache is fully rebuildable: `clean_future_dated_directory_caches_once` (`directory/methods.rs:1433`) deletes the entire `app-cache.sqlite3` file set for every account behind a marker file and lets it repopulate. A shape mismatch deserves the same treatment.

Two options, either of which is a net deletion:

1. Drop the validator. Only this crate has ever created these tables, so adopting them at version 1 is safe, and any *future* divergence is what versions 2+ and the future-version refusal are for.
2. Keep it, but route the mismatch to a rebuild rather than a hard failure — drop the `directory_*` tables inside the same transaction, re-run the frozen DDL, and record version 1. Data loss is a re-fetch; a permanently unopenable cache is a broken account.

crates/marmot-app/src/directory/cache/migrations.rs
```diff
@@ -0,0 +1,287 @@
...
+}];
+
+const VERSION_1_SQL: &str = include_str!("v1.sql");
+
+pub(super) fn run_all(conn: &mut Connection) -> Result<(), AppError> {
+    // SQLite/JSON errors can contain database-controlled names or values.
+    // Keep the typed downgrade error, but never surface those private details.
+    run(conn, MIGRATIONS).map_err(|error| match error {
+        AppError::Storage(StorageError::UnsupportedSchemaVersion { .. }) => error,
+        _ => StorageError::Backend("directory cache schema migration failed".into()).into(),
```
erskingardner, line 34: **The blanket mask destroys diagnostics that carry no private data, and the privacy rationale it cites is not applied next door.**

Every non-`UnsupportedSchemaVersion` failure collapses to `"directory cache schema migration failed"`: `SQLITE_BUSY`, `SQLITE_IOERR`, `SQLITE_NOTADB` / wrong key, a genuine `invalid_history()`, and the `serde_json` error this is actually aimed at all become the same unactionable string. A transient lock and permanent schema corruption are indistinguishable in a support log, and neither the operator nor a future maintainer can tell which happened.

The leak worth suppressing is narrow: `serde_json::Error` embeds the offending JSON, which is exactly what `migration_legacy_failure_rolls_back_schema_and_data` asserts against. SQLite's own messages here reference names from the frozen DDL, not stored values.

And the rationale is inconsistent with the code 170 lines away: `cache.rs:247` does `relay_lists: serde_json::from_str(&row.relay_lists_json)?`, and `AppError::Json` is `#[error(transparent)]` — so the hot read path already surfaces record content in error text on every malformed row. Either that leak matters, in which case fix it too rather than only the migration path, or it does not, in which case this mask is buying nothing at the price of all migration diagnostics.

Suggest masking only the JSON conversion (or matching on `AppError::Json`) and letting `rusqlite`/`StorageError` through unchanged.

crates/marmot-app/src/directory/cache/migrations.rs
```diff
@@ -0,0 +1,287 @@
...
+        if migration.version <= previous {
+            return Err(invalid_history());
+        }
+        previous = migration.version;
+    }
+    let latest_supported = previous;
+    for migration in migrations {
+        // Lock before reading history so concurrent openers cannot both apply a
+        // migration. Even ledger creation is rolled back on first-open failure.
+        let tx = conn.transaction_with_behavior(TransactionBehavior::Immediate)?;
```
erskingardner, line 52: **Every cache open now takes a write lock, and the history read/validation is loop-invariant work done per migration.**

This `BEGIN IMMEDIATE` is unconditional: it runs on every `DirectoryCache::open`, including the overwhelmingly common case of an already-current cache where nothing is pending. It then commits an empty transaction. Previously a warm reopen was read-only — `CREATE TABLE IF NOT EXISTS` on existing tables generates no write, and `migrate_legacy_json_records` early-returned after a `table_exists` read.

The connection has no busy timeout: `DirectoryCache::open` uses plain `Connection::open`, and `open_hardened_sqlcipher` (`storage-sqlite/src/connection.rs:1035`) sets only cipher, `secure_delete`, and `temp_store` — never `busy_timeout`, unlike the account-storage path at `connection.rs:947`. So any contending writer yields an instant `SQLITE_BUSY`, which the mask above turns into a permanent-looking migration failure. `MarmotRootRuntimeLease` makes contention unlikely by contract, but the unleased `with_relays_and_account_home_and_config` constructor is public and `#[doc(hidden)]`, so the guarantee is a convention rather than an invariant.

Separately, the ledger `CREATE TABLE IF NOT EXISTS`, the full history `SELECT`, the future-version check, and the prefix check all sit inside `for migration in migrations`, so they re-run once per migration rather than once. That is invisible at one migration and quadratic at twenty.

Both fall out of one restructure, which is also what the existing session runner does (`storage-sqlite/src/migrations.rs:435`): read and validate the history once up front, then open a transaction only inside the branch that actually applies a migration. That makes the steady-state open read-only and removes the repeated work.

crates/marmot-app/src/directory/cache.rs
```diff
@@ -67,14 +69,13 @@ impl DirectoryCache {
         Self::from_connection(conn).map(Some)
     }
 
-    fn from_connection(conn: Connection) -> Result<Self, AppError> {
-        initialize_schema(&conn)?;
+    fn from_connection(mut conn: Connection) -> Result<Self, AppError> {
+        migrations::run_all(&mut conn)?;
```
erskingardner, line 80: This now runs the full migration runner — ledger creation, shape validation, legacy conversion — against the *legacy plaintext* root cache too, via `open_legacy_plaintext`.

That file is the one most likely to have an unexpected shape: it is the oldest store, it is the one that still holds `user_directory_records`, and it is plaintext, so unlike the per-account cache it is genuinely modifiable outside this code. A refusal there propagates through `migrate_legacy_directory_cache_once` (`directory/methods.rs:1389`) out of `directory_caches()`.

The retry behaviour makes that worse: `*checked = true` is only reached on the success path, so a refusal is re-attempted and re-fails on every subsequent `directory_caches()` call. One malformed legacy file permanently breaks directory reads for *all* accounts, rather than skipping an import of a store that is being retired anyway.

If the validator stays, the legacy plaintext import should degrade to "skip the import, mark checked" rather than failing the caller.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-05T06:26:40Z` |
| Commit reviewed | `032c2d4bc8306b70ad19fd0df0b369e822e3d521` |

## Summary

The premise is sound and the documented gap is real: `app-cache.sqlite3` had no schema identity, `CREATE TABLE IF NOT EXISTS` silently adopts anything, and `app-sqlite-storage-boundaries.md` already named this as required work. The ledger itself is the right shape — separate table name, per-migration transaction with the ledger row, typed `UnsupportedSchemaVersion`, DDL copied verbatim so adoption is genuinely lossless. The rollback tests (`migration_ledger_insert_failure_rolls_back_body`, `migration_legacy_failure_rolls_back_schema_and_data`) are the good kind: they prove the atomicity claim rather than restating it.

I found no correctness bug. My concerns are about what happens when this code says no, and about how much of it exists to say no at all.

**1. The shape validator defends a state no shipped build can produce, and fails closed on a store this module already wipes wholesale.** The directory DDL was introduced once and never changed; nothing but this crate writes the file. When validation does trip, `directory_caches()` — display names, search, the whole `directory_*` API — fails permanently with an opaque string, while `clean_future_dated_directory_caches_once` a few hundred lines away deletes this same file set for every account and lets it repopulate. For a reconcilable cache, rebuild is the correct answer and brick is not. Dropping the validator or rebuilding on mismatch both remove code.

**2. The blanket error mask trades all migration diagnostics for a leak it does not actually close.** `BUSY`, `IOERR`, wrong-key, and real history corruption all become one unactionable sentence. Meanwhile `cache.rs:247` surfaces raw JSON through `AppError::Json` (`#[error(transparent)]`) on the hot read path, so the private-data argument is applied in one place and not the one next to it. Mask the JSON conversion; let SQLite errors through.

**3. Every open now takes an unconditional `BEGIN IMMEDIATE`**, on a connection with no `busy_timeout`, and re-reads and re-validates the whole history once per migration. Reading history once before the loop and opening a transaction only when a migration is pending — what `storage-sqlite/src/migrations.rs:435` already does — fixes both.

**4. The legacy plaintext import inherits all of the above** and never sets its `checked` flag on failure, so a refusal repeats forever and takes all accounts' directory reads with it.

### Lower priority

- The runner duplicates `crates/storage-sqlite/src/migrations.rs` — same `Migration { version, name, apply }`, ordering check, future-version refusal, apply-plus-ledger — with subtly different semantics (strict prefix vs. per-version name check; `Immediate` vs. deferred). The docs require independent *ledgers*, not independent *runners*. The existing runner is `pub(crate)` and hardcodes `cgka_schema_migrations`, so sharing it means real work, and duplicating is defensible today. Worth a comment in one file pointing at the other so the two do not drift into different answers for the same question.
- `v1.sql` is described as frozen but keeps `IF NOT EXISTS` — deliberate, since adoption depends on it, and coupled to the validator existing. If the validator goes, a note in the file would stop a future `0002` author copying the pattern.

### Scope

Scope discipline is good: no shared-store changes, no version bumps, docs updated with dated front matter, and the PR body is honest about the six pre-existing baseline failures and the one intermittent test rather than papering over them. The net line count is the main thing I would push on — roughly 340 added lines to version a cache that gets deleted wholesale by an existing marker cleanup, where a version stamp plus rebuild-on-mismatch would be a fraction of that and would fail *open* instead of shut.

crates/marmot-app/src/directory/cache/migrations.rs
erskingardner: Addressed in bf527d79: preserve static StorageError diagnostics, including invalid migration history, while continuing to suppress stored names and values. Added an assertion for the exact safe history error.

crates/marmot-app/src/directory/cache/migrations.rs
erskingardner: Addressed in bf527d79: pragma_table_info and pragma_index_info now use bound names, matching pragma_index_list.

crates/marmot-app/src/directory/cache/migrations.rs
erskingardner: Retaining validation deliberately: the requested acceptance contract explicitly says an existing table missing required columns must fail without recording version 1, and existing directory/search data must be preserved. Rebuilding or silently adopting a malformed schema would violate that contract. The one-time future-dated cleanup has a separate, specific purpose; it is not authorization to discard arbitrary malformed caches during normal upgrades. bf527d79 improves the safe error diagnostics, and v1.sql now documents why IF NOT EXISTS is coupled to validation. A recovery UX can be designed separately without weakening this migration contract.

crates/marmot-app/src/directory/cache/migrations.rs
erskingardner: Addressed the diagnostic loss in bf527d79. Static history errors and typed SQLite extended result codes are now preserved; legacy JSON failures have a distinct static serialization error. SQLite message text is still suppressed because triggers can emit arbitrary stored/private text through RAISE, demonstrated by the new regression test. Thus letting every SQLite message through is not safe. The pre-existing hot-read JSON error surface is outside this migration PR; it does not justify introducing the same exposure here.

crates/marmot-app/src/directory/cache/migrations.rs
erskingardner: Addressed in bf527d79: validate history once using reads and skip the migration loop entirely for an up-to-date cache. The new WAL regression holds another connection's write transaction with a zero busy timeout and proves the current cache opens without contending. Only pending migrations take BEGIN IMMEDIATE. They deliberately revalidate history under that lock so concurrent openers cannot apply the same migration or miss a newly committed future version. The runner now links to storage-sqlite's related contract.

crates/marmot-app/src/directory/cache.rs
erskingardner: Retaining the failure/preservation behavior required by the request. The previous plaintext open also initialized tables and could fail while reading/converting legacy records; the import's checked flag already stayed unset on failure. Marking a malformed import checked would silently suppress failed work and change the explicitly preserved legacy-import semantics. A malformed schema is now refused transactionally with safe diagnostics, preserving the source for repair/retry. Existing plaintext import success, once-only behavior, and failed-import file-preservation tests pass. Skipping or rebuilding a failed import needs a separate recovery policy.

crates/marmot-app/src/directory/cache/migrations.rs
```diff
@@ -0,0 +1,346 @@
...
+    for migration in migrations {
+        if migration.version <= previous {
+            return Err(invalid_history());
+        }
+        previous = migration.version;
+    }
+    // A current cache needs only reads, even while another connection writes.
+    let applied = applied_count(conn, migrations)?;
+    for migration in &migrations[applied..] {
+        let tx = conn.transaction_with_behavior(TransactionBehavior::Immediate)?;
```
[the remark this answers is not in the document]
erskingardner: Checked against the pinned dependency: rusqlite 0.40.1 already calls sqlite3_busy_timeout(db, 5000) in src/inner_connection.rs:118 when opening a Connection. Its busy.rs documentation also states the 5000 ms default. Both encrypted and plaintext DirectoryCache opens use Connection::open, and open_hardened_sqlcipher does not clear that handler. Thus pending migrations already have the proposed five-second bounded wait; there is no instant-BUSY default gap here. The zero-timeout override is intentional and test-only, proving a current cache requires no write lock. Leaving the implementation unchanged.

crates/marmot-app/src/directory/cache/migrations.rs
```diff
@@ -0,0 +1,346 @@
...
+            }
+        }
+        let query = if kind == "table" {
+            r#"SELECT name, type, "notnull", dflt_value, pk FROM pragma_table_info(?1) ORDER BY cid"#
+        } else {
+            "SELECT name, '', 0, NULL, 0 FROM pragma_index_info(?1) ORDER BY seqno"
+        };
+        if shape(conn, query, &name)? != shape(&reference, query, &name)? {
+            return Err(
+                StorageError::Backend("invalid directory cache schema shape".into()).into(),
```
erskingardner: Non-blocking question on the refusal path: when an unversioned cache fails shape validation here (as opposed to carrying a newer ledger version), `DirectoryCache::open` errors and every `directory_cache_for_account` call will keep failing until someone manually deletes the file. The README describes these tiers as reconcilable caches, so a delete-and-rebuild fallback would be safe for *this* failure mode (while `UnsupportedSchemaVersion` must stay fail-closed to avoid clobbering a newer schema). Did you consider distinguishing the two, or is manual file removal the intended recovery UX? Failing closed uniformly is a defensible, simpler contract — just want to confirm it was a deliberate choice rather than an accident of sharing one error path.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-05T06:52:53Z` |
| Commit reviewed | `bf527d79cfe39af25c1f9bbe3c4250135ca4b150` |

## Summary

This is a careful, well-scoped change: the per-account directory cache gets its own `app_cache_schema_migrations` ledger, version 1 freezes the existing schema with shape-validated adoption of unversioned databases, legacy `user_directory_records` conversion moves inside the migration transaction, and future versions fail closed with the typed `UnsupportedSchemaVersion`. The follow-up commit on this head (`bf527d79`) materially improved the runner: steady-state opens are now read-only (no write lock on every cache open, with a WAL contention test to prove it), and error mapping now preserves safe diagnostics (SQLite result codes, static history errors) while still stripping database-controlled messages and legacy JSON values.

Reviewed the full diff plus surrounding context at the head SHA (`cache.rs` open/put paths, `methods.rs` cache lifecycle, `storage-sqlite` migration conventions). The migration logic holds up under the cases I traced: prefix/name validation, concurrent-opener revalidation under the Immediate lock, per-migration rollback including first-open ledger creation, and the mutation-matrix tests (dropped columns, missing PK, wrong/unique indexes) all reject correctly. Privacy invariants are respected: no stored values surface in errors, no new tracing fields, docs and crate maps updated with dated front matter.

## Why not a clean approve

Required CI failed on the previous head (`032c2d4b`): `Relay runtime tests` hit a 5s event timeout in `app_runtime_delete_group_local_removes_projection_without_publishing_leave` (`relay_runtime.rs:5680`). That test exercises group-deletion projection and never touches the directory cache, and master has been green on that job recently, so this looks like the flaky/loaded-runner class of failure the PR body already documents — but it is not proven unrelated. The new head's CI was still running at review time. Please confirm a green Required CI on `bf527d79` (or a re-run demonstrating the flake) before merge; branch protection should enforce this anyway.

## Non-blocking notes

- One inline question on adoption-refusal recovery UX.
- The runner intentionally duplicates `storage-sqlite`'s migration concept rather than sharing it; the new comment explaining why (session-ledger hardcoding, session-specific legacy-name reconciliation) is the right call and keeps `marmot-app` self-contained.

Nice work — the test coverage here (idempotent reopen under insert/update triggers, adoption preserving both tiers, legacy-conversion rollback and retry, ledger-insert failure rollback, redaction assertions) is exactly the right shape for a migration ledger.

crates/marmot-app/src/directory/cache/migrations.rs
erskingardner: Yes, this is deliberate. The requested migration contract requires malformed unversioned schemas to fail without recording version 1 and requires preserving existing directory/search data. Being reconstructible from the network does not guarantee that rebuilding immediately restores offline records or their discovery provenance, so this PR does not automatically delete them. Schema-shape failures have a distinct static diagnostic from the typed UnsupportedSchemaVersion error; they are not conflated, but neither triggers destructive recovery. There is no automatic recovery UX in this change: the original file remains available for repair or an explicitly chosen rebuild. A supported rebuild/recovery flow can be designed separately. Also confirming that Required CI is now green on bf527d79: https://github.com/marmot-protocol/mdk/actions/runs/33950321800.

crates/marmot-app/src/directory/cache/assurance_tests.rs
```diff
@@ -0,0 +1,257 @@
...
+    assert!(before.iter().all(|(count, _)| *count == 20_000));
+    drop(conn);
+
+    let started = std::time::Instant::now();
+    let cache = DirectoryCache::open(
+        path.clone(),
+        &SqlCipherKey::new("synthetic-upgrade-test-key").unwrap(),
+    )
+    .unwrap();
+    eprintln!(
```
erskingardner: **Blocking: this `eprintln!` fails Required CI.** `production_library_sources_do_not_write_direct_output` (`crates/cgka-conformance-simulator/tests/tracing_audit.rs:158`) scans every `.rs` file under production source roots with only CLI-binary and feature-gated-test-support exemptions — there is no `#[cfg(test)]` carve-out, and it's a line-based substring scan. The CI log reports both `println!` and `eprintln!` here because `eprintln!(` contains `println!(`.

Simplest fix: delete the timing print (the number is already captured in the PR body's validation notes). If you want the measurement to stay runnable, an assertion with a very generous upper bound or a no-op timing capture would satisfy the audit without direct output. Everything else in this test file passed CI and is high-value — please keep it.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-05T07:38:30Z` |
| Commit reviewed | `404f7ab981af0826d5af2cfc01e49d0170337727` |

**Re-review** of the two commits since `032c2d4` (the PR was rebased onto `7368c737`, so I re-read the changed files at the new head in full).

**Blocking — CI is red on this head.** The `Simulator and vectors` job fails on `production_library_sources_do_not_write_direct_output`, and `Required CI` fails as a consequence ([run 33952588551](https://github.com/marmot-protocol/mdk/actions/runs/33952588551)). Details inline; the one-line `eprintln!` timing print added in `404f7ab9` trips the repo-wide direct-output audit, which scans all library `src/` files with no `#[cfg(test)]` exemption. This passed locally because the audit lives in the simulator's test suite, not `marmot-app`'s.

Everything else from the delta is good:

- `bf527d79` correctly resolves all outstanding feedback: current-cache opens are read-only (proved by the zero-busy-timeout writer-contention test), pending migrations revalidate history under the write lock, static storage diagnostics and SQLite extended result codes survive redaction while trigger payloads and JSON values do not (the RAISE regression test justifies stripping message text), and `pragma_table_info`/`pragma_index_info` now bind names. The restructured runner is correct on the edges I checked: slice bounds are guaranteed by exact-prefix validation, the concurrent-opener race is handled by the re-read under lock, and the empty-ledger and fresh-DB paths behave.
- The declined findings (retaining shape validation, plaintext-import failure disposition) have reasoned contract justifications, and the distinct static shape-mismatch diagnostic plus the fail-closed recovery answer resolve my earlier availability concern adequately for this PR.
- The new assurance tests are excellent evidence: full SHA-256 hash-preservation of 120k rows across all six tables (including deliberately invalid cached JSON that adoption must not parse), simulated `SQLITE_FULL` with rollback and successful retry, and child-process hard exit mid-migration with journal recovery verified in both DELETE and WAL modes. The historical DDL fixture is independently extracted and the lineage claim checks out.
- Docs match the code.

Drop the `eprintln!` (the timing evidence is already recorded in the PR body) and this should be mergeable once CI reruns green.

crates/marmot-app/src/directory/cache/assurance_tests.rs
erskingardner: **Blocking: this line fails CI on this head SHA.**

`Simulator and vectors` failed at 07:34 UTC on `404f7ab9`, and `Required CI` is red as a result:

```
crates/marmot-app/src/directory/cache/assurance_tests.rs:155 production library source writes direct output with `println!(`
crates/marmot-app/src/directory/cache/assurance_tests.rs:155 production library source writes direct output with `eprintln!(`
```

`production_library_sources_do_not_write_direct_output` (`crates/cgka-conformance-simulator/tests/tracing_audit.rs:157`) is a plain line scan over `production_source_roots`. It has no `#[cfg(test)]` awareness at all — it only skips CLI binaries and feature-gated test support — so any `eprintln!`/`println!` under `crates/*/src/**` trips it, including one inside a `#[path = "assurance_tests.rs"] mod assurance_tests` that only compiles under `cfg(test)`. (The double hit is the scan matching both macro tokens on the same substring.)

The fix is also the right cleanup: this is a timing print, not an assertion. Drop it, along with `started` and the `Instant::now()` on line 149 — five lines out, and the test loses nothing it was actually checking.

Also worth noting: nextest cancelled with `17/405 tests were not run due to test failure`, so this run does not clear the rest of that suite. Please confirm a full green `Required CI` after the fix rather than treating the other jobs' green as complete coverage.

crates/marmot-app/src/directory/cache/migrations.rs
```diff
@@ -0,0 +1,350 @@
...
+    })?;
+    for object in objects {
+        let (kind, name) = object?;
+        let object_query = "SELECT type, tbl_name FROM sqlite_master WHERE name = ?1";
+        let expected: (String, String) =
+            reference.query_row(object_query, [&name], |row| Ok((row.get(0)?, row.get(1)?)))?;
+        let actual: (String, String) =
+            conn.query_row(object_query, [&name], |row| Ok((row.get(0)?, row.get(1)?)))?;
+        if actual != expected {
+            return Err(invalid_history());
```
erskingardner: The follow-up commit gave the column/index-shape mismatch its own message (`invalid directory cache schema shape`, line 167), which is a real improvement — but the two mismatches above it still report `invalid_history()` (`invalid directory cache migration history`). Line 147 fires when an object exists with the wrong `type`/`tbl_name`, and line 157 when an index's `unique`/`partial` flags differ. Those are schema-shape problems, not ledger problems, and reporting them as history corruption points an operator at the wrong table.

All three are the same failure class. Give them the same message.

crates/marmot-app/src/directory/cache/assurance_tests.rs
```diff
@@ -0,0 +1,257 @@
...
+    let mut conn = encrypted_connection(&dir.path().join("app-cache.sqlite3"));
+    seed_structured(&mut conn, 25);
+    let expected = seed_legacy(&mut conn, 1_000);
+    let tables = [TABLES, &["user_directory_records"]].concat();
+    let before = snapshot(&conn, &tables, -1);
+    let pages: i64 = conn
+        .pragma_query_value(None, "page_count", |r| r.get(0))
+        .unwrap();
+    // Permit the ledger to be created but not the duplicated legacy records.
+    conn.pragma_update(None, "max_page_count", pages + 4)
```
erskingardner: `pages + 4` is tuned to land the failure inside the legacy conversion, but nothing in the test verifies that it did. If SQLite's allocation shifts — a page-size default change, autovacuum, a future SQLCipher bump — and `CREATE TABLE app_cache_schema_migrations` itself is what hits the limit, every assertion below still passes: the error is still `DiskFull`, the ledger table still does not exist, and the snapshot still matches. The test would quietly degrade from "legacy conversion rolls back cleanly" to "we could not create the ledger table", which is a much weaker claim, and no one would notice.

A cheap discriminator would keep the test honest — e.g. assert `directory_users` grew past 25 rows inside the doomed transaction before the rollback, or check that the failure arrives after at least one `put_with_reason_locked`. Non-blocking, but this is the one test here whose value depends on an unasserted invariant.

crates/marmot-app/src/directory/cache/assurance_tests.rs
```diff
@@ -0,0 +1,257 @@
...
+        );
+    }
+}
+
+#[test]
+fn populated_historical_encrypted_cache_preserves_all_120000_rows() {
+    let dir = tempfile::tempdir().unwrap();
+    let path = dir.path().join("app-cache.sqlite3");
+    let mut conn = encrypted_connection(&path);
+    seed_structured(&mut conn, 20_000);
```
erskingardner, line 144: 20,000 users across six tables, seeded through SQLCipher, then SHA-256 over every typed value in all 120,000 rows three separate times (before, after open, after reopen). This runs in every unit-test shard, forever.

What it proves is that adoption does not rewrite rows — but adoption is `execute_batch(VERSION_1_SQL)` with `IF NOT EXISTS` plus pragma-only introspection, so it cannot touch a row by construction, and the shards already prove that at 25 rows in the neighbouring tests. The 120,000 figure buys confidence for the PR narrative rather than coverage the suite lacks.

The shards passed, so this is not blocking. But if it shows up as a SLOW marker or starts pushing shard wall-clock, cutting `users` to something like 2,000 preserves every property this asserts. Non-blocking.

crates/marmot-app/src/directory/cache/assurance_tests.rs
erskingardner: Blocking: this `eprintln!` fails the tracing audit `production_library_sources_do_not_write_direct_output` (cgka-conformance-simulator/tests/tracing_audit.rs), which is why **Simulator and vectors** and **Required CI** are red on the current head ([run 33952588551](https://github.com/marmot-protocol/mdk/actions/runs/33952588551)). The audit scans every `.rs` under library `src/` trees — `#[cfg(test)]` files included — and only exempts files named `test_support.rs` and CLI binaries. The audit reports the line twice because `eprintln!(` also contains the `println!(` substring; it is this single call.

Simplest fix: delete the timing `eprintln!` (nextest already surfaces this as a slow test, so the diagnostic adds little). Verify locally with `cargo test -p cgka-conformance-simulator --test tracing_audit` before pushing — `just fast-ci` and the `-p marmot-app` runs do not execute this audit, which is how it slipped through the follow-up commit.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-05T07:40:02Z` |
| Commit reviewed | `404f7ab981af0826d5af2cfc01e49d0170337727` |
| Recommended action | **Fix blocking issues before merge** (posted as a comment: GitHub refuses `--request-changes` on your own PR) |

Re-review of my earlier comments (#pullrequestreview-5120122750, on `032c2d4b`). Scoped to what changed since: `bf527d79` and `404f7ab9`.

## Blocking

**CI is red on this head SHA, from a line added in `404f7ab9`.** `Simulator and vectors` failed at 07:34 UTC and took `Required CI` with it. The `eprintln!` timing print at `assurance_tests.rs:155` trips `production_library_sources_do_not_write_direct_output`, which line-scans every file under `crates/*/src/**` with no `cfg(test)` awareness. Inline comment has the detail; the fix is to delete the print and its `Instant`. Note also that nextest cancelled with 17 tests unrun, so the rest of that suite is not cleared by this run.

## My four earlier findings

**3 (unconditional `BEGIN IMMEDIATE`) — fixed, and fixed the right way.** History is read once before the loop, the loop runs only over `&migrations[applied..]`, and the write lock is taken only when something is actually pending. The revalidation under the `Immediate` lock closes the read-then-lock race without reintroducing the per-iteration cost. I traced the concurrency cases — opener B finishing between A's read and A's lock, B applying more migrations than A has compiled, B being a newer binary — and each lands correctly, the last as a fail-closed `UnsupportedSchemaVersion`. `current_cache_migrations_do_not_contend_with_writer` is a real test: setting `busy_timeout` to zero is exactly what makes the assertion bite.

**I was wrong about the busy timeout, and I should say so plainly.** rusqlite calls `sqlite3_busy_timeout(db, 5000)` in `InnerConnection::open_with_flags`, so every connection here already has a 5s timeout before any of our code runs. My "no `busy_timeout`" sub-point was factually incorrect, and CodeRabbit's suggestion to add one is likewise unnecessary — it would set the value SQLite already has. No action needed.

**2 (blanket error mask) — substantially addressed.** Extended result codes now survive with the message stripped, so `BUSY`/`IOERR`/`NOTADB` stay diagnosable while trigger payloads do not leak; statically-constructed `StorageError`s pass through; JSON collapses to one fixed sentence. `migration_errors_preserve_safe_diagnostics_without_trigger_payloads` pins both halves. The `AppError::Storage(_) => error` passthrough is safe today only because every `StorageError` reaching it is built from static text in this file — true, but load-bearing and enforced only by the comment. Fine as-is; just don't let a future migration body construct a `StorageError` from a stored value.

The inconsistency I raised is unchanged: `cache.rs:247` still surfaces raw JSON through `AppError::Json` (`#[error(transparent)]`) on the hot read path. That is pre-existing and outside the migration path, so I no longer think it belongs in this PR — but it is the same leak the migration mask exists to prevent, sitting a few hundred lines away. Worth a tracked follow-up.

**1 (validator bricks `directory_caches()`) and 4 (legacy import never sets `checked` on refusal) — declined, and the decline is defensible.** Refusing rather than rebuilding preserves a malformed cache for inspection, and I accept that changing it is recovery-policy design rather than migration-ledger work. Two things I would still like recorded rather than argued here:

- The blast radius of a refusal on the legacy plaintext path is every account's directory reads, permanently, triggered by one stale file from a retired code path — and it repeats on every call because `*checked` stays false. The `serde_json` half of that was already possible before this PR; the shape validator widens the set of inputs that reach it.
- `validate_version_1` can only fire on a genuinely foreign or corrupt file, since the DDL never changed across the eight master-lineage revisions the new fixture pins. Failing closed there is reasonable. One inline nit on message consistency.

A follow-up issue naming both would close this out for me.

## `404f7ab9` (test-only)

I confirmed it changes no production code: the only non-test diff since my review is in `bf527d79`. `pre_ledger.sql` is byte-identical to `v1.sql` apart from its header, which is the point — it pins the historical shape so a future `v1.sql` edit cannot silently move the fixture with it. The interrupted-upgrade test is the strongest of the three: re-execing the test binary and `exit(77)` inside the transaction genuinely bypasses every destructor, and running it under both DELETE and WAL is the right pairing. Two non-blocking notes inline on the disk-full and 120k-row tests.

## Verdict

Everything I raised is either fixed, defensibly declined, or (in one case) my own error. The only thing standing between this and a merge is the `eprintln!`.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-05T07:41:44Z` |
| Commit reviewed | `9a8925c4e6bedc5cfc365e49c659db814aadce35` |
| Model | `openrouter/qwen/qwen3.8-max` |

Reviewed the full diff at `404f7ab9` and re-verified at `9a8925c4` (the only delta is the 5-line removal of the timing `eprintln!` in `assurance_tests.rs`, which fixes the one blocking finding — see the inline comment, now resolved by that push). CI is re-running for `9a8925c4`; the previous head failed only `production_library_sources_do_not_write_direct_output` in the tracing audit, and no direct-output macros remain in the added code.

The design holds up:

- **Runner correctness.** The read-only `applied_count` fast path plus revalidation under `BEGIN IMMEDIATE` correctly serializes concurrent openers (including the WAL read-snapshot race between the two autocommit reads), and the skip branch commits an empty transaction harmlessly. Body + ledger row commit atomically per migration, and first-open ledger creation rolls back with a failed body — all covered by tests.
- **Fail-closed history.** Exact-prefix + name matching, compiled-list ordering enforcement (`version <= previous` starting at 0 also rejects non-positive versions), and `UnsupportedSchemaVersion` for newer recorded versions mirror the session runner's contract without reusing its session-specific ledger and legacy-name reconciliation. The local-runner justification comment is apt.
- **Privacy.** Error mapping preserves extended SQLite result codes and static text while dropping SQLite message text (including attacker/DB-controlled trigger payloads) and legacy JSON content; regression tests assert non-leakage. Matches the repo's tracing/privacy invariants.
- **Adoption validation.** Create-then-validate inside the same transaction is safe (rollback undoes creates), and structural comparison via `pragma_table_info`/`pragma_index_list`/`pragma_index_info` against an in-memory reference is a reasonable middle ground given the verified DDL-identity history back to `9db9dbc0`.
- **Test depth.** Adoption of both tiers, partial-schema refusal, legacy rollback + retry, simulated `SQLITE_FULL`, and destructor-free child-process interruption under both DELETE and WAL journals, with full-value snapshot hashing — unusually strong evidence for a cache migration, and the PR body is honest about its limits.

Non-blocking nit: the rewrapped `crates/marmot-app/AGENTS.md` line describing `cache.rs` is 158 chars while the surrounding prose wraps at ~120; consider rewrapping.

Approving contingent on the pending CI run for `9a8925c4` going green.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-05T07:42:10Z` |
| Commit reviewed | `9a8925c4e6bedc5cfc365e49c659db814aadce35` |
| Recommended action | **Merge**, once CI is green on this SHA |

Supersedes my review a minute ago. You pushed `9a8925c4` seventeen seconds after I re-checked the head SHA, so that review posted against `404f7ab9` and its one blocking finding is already fixed. Everything else in it stands.

**The blocking issue is resolved.** `9a8925c4` is `-5 / +0`: the `eprintln!` and its `Instant` are gone from `populated_historical_encrypted_cache_preserves_all_120000_rows`, and nothing else moved. I re-scanned the PR's new and modified sources for `println!(` / `eprintln!(` / `dbg!(` — the three tokens in `DIRECT_OUTPUT_MACROS` — and there are no remaining hits, so `production_library_sources_do_not_write_direct_output` should now pass.

Two things to confirm before merging, neither of which needs another review pass from me:

- `Required CI` green on `9a8925c4`. The failing run cancelled with `17/405 tests were not run due to test failure`, so `Simulator and vectors` never finished its suite on the previous head — that job's green is genuinely unproven, not merely stale.
- The three non-blocking inline comments still open on `404f7ab9` (shape-error message consistency in `validate_version_1`, the unasserted `pages + 4` invariant in the disk-full test, the 120k-row cost). All are follow-up-or-ignore, none should hold the merge.

To restate the re-review verdict without making you read the superseded body: finding 3 is fixed correctly, including the concurrency revalidation; finding 2 is substantially addressed and the residual `cache.rs:247` leak is pre-existing and out of scope here; findings 1 and 4 are declined on a defensible preservation-over-rebuild rationale, and I would like the legacy-plaintext blast radius recorded in a follow-up issue rather than argued in this thread; and my "no `busy_timeout`" sub-point was simply wrong — rusqlite sets 5000 ms in `InnerConnection::open_with_flags` before any of our code runs.

Good change. The ledger is the right shape, and the assurance tests — particularly the re-exec-and-`exit(77)` interruption test across both DELETE and WAL — prove the atomicity claim rather than restating it.

crates/marmot-app/src/directory/cache/assurance_tests.rs
erskingardner: Fixed in 9a8925c4: removed the timing print and unused timer, keeping every data-preservation assertion. All four tracing_audit tests now pass locally, alongside all 21 cache tests and just fast-ci. Fresh Required CI is running on the latest follow-up.

erskingardner: Fixed in 9a8925c4: deleted precisely that print and timer. Ran the complete tracing_audit target (4 passed), all 21 cache tests, and just fast-ci. I am waiting for the full new CI result rather than treating the cancelled simulator run as coverage.

erskingardner: Fixed in 9a8925c4: removed the unnecessary direct output and timer; the assertions remain. The complete tracing_audit target now passes locally (4 tests), as do all 21 cache tests and just fast-ci.

crates/marmot-app/src/directory/cache/migrations.rs
erskingardner: Fixed in ddb1304b: all three object/index/column-shape mismatch paths now use invalid_shape(), distinct from invalid_history(). The mutation-matrix test asserts the exact schema-shape diagnostic.

crates/marmot-app/src/directory/cache/assurance_tests.rs
erskingardner: Good catch. Adding the progress assertion demonstrated that the old page limit did in fact fail before the first complete legacy record was converted. ddb1304b adds a test-only, thread-local successful-conversion counter and requires a positive count after SQLITE_FULL. The limit now permits some records but not the full conversion. The corrected test passes, verifies both tiers roll back, then retries and compares every converted record. This prevents the regression test from silently degrading to an early allocation failure.

crates/marmot-app/src/directory/cache/assurance_tests.rs
erskingardner: Retaining this deliberately in response to the user's explicit concern about substantial existing deployed data. It adds populated historical-schema evidence independent of the compiled v1 SQL, not merely another small-fixture check. Nextest partitioning assigns the test to one shard per matrix rather than running it in every shard. The complete 21-test cache group, including this case and crash/disk-full tests, completed locally in 4.65 seconds. If measured CI cost becomes significant we can move this stress case to a dedicated lane; there is no demonstrated need to reduce its scale now.

crates/marmot-app/src/directory/cache/assurance_tests.rs
```diff
@@ -0,0 +1,258 @@
...
+    // Permit the ledger and some converted records, but not the full conversion.
+    // The progress assertion below guards this premise against allocation changes.
+    conn.pragma_update(None, "max_page_count", pages + 64)
+        .unwrap();
+    super::super::LEGACY_RECORDS_CONVERTED.with(|count| count.set(0));
+    let error = run_all(&mut conn).unwrap_err();
+    assert!(
+        super::super::LEGACY_RECORDS_CONVERTED.with(|count| count.get()) > 0,
+        "disk exhaustion must occur after at least one legacy record was converted"
+    );
```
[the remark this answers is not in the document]
erskingardner: Addressed in 125f3552: the counter must now be both positive and less than expected.len(), proving the SQLITE_FULL failure interrupts partial conversion rather than occurring after all record writes. All three assurance tests and just fast-ci pass.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-05T08:27:31Z` |
| Commit reviewed | `125f355220508caed353c092653cac6b891004c8` |

## Re-review summary

Re-reviewed the four commits pushed since my first review (`bf527d79` read-only opens/error mapping, `404f7ab9` assurance tests, `9a8925c4` direct-output fix, `ddb1304b`/`125f3552` disk-full hardening and the `invalid_shape`/`invalid_history` split).

**Verified independently:**

- `fixtures/pre_ledger.sql` is byte-identical (modulo comments) to `initialize_schema` at `9db9dbc`, the commit the header cites — all 8 statements match. The fixture also matches the pre-PR base, so the adoption target is genuinely the historical shape.
- The new disk-full test is now airtight: the `LEGACY_RECORDS_CONVERTED` counter is bounded `0 < n < 1000`, proving SQLITE_FULL interrupted conversion *mid-stream*, while the post-failure snapshot equals the pre-migration state and the ledger table is absent. The premise guards protect the test against future page-allocation changes.
- The child-process crash test (exit 77 between conversion/table-drop and ledger commit) covers both DELETE and WAL recovery, and `--exact` module path wiring is correct (`directory::cache::migrations::assurance_tests::...`).
- The `invalid_shape` vs `invalid_history` split improves diagnosability while keeping both messages free of database-controlled values; the mutation-matrix test now pins the exact message.
- The storage-boundaries doc's new assurance/limits section honestly scopes what these tests do and don't prove, and explicitly states the refused-schema availability trade-off — that answers my earlier inline question; failing closed with a preserved file is a deliberate, documented contract.
- No new direct-output macros, no new tracing fields, no workspace/vector version bumps.

**CI:** fully green on this head, including `Required CI`, `Relay runtime tests`, and `Simulator and vectors`.

For the record on the intermediate red run at `ddb1304b`: the two relay-runtime failures were pre-existing flake classes unrelated to this PR — `create_group_detailed_returns_durable_emitted_chat_list_row` raced a one-second `updated_at` rollover (`relay_runtime.rs:9877`, rows identical except `updated_at: ...760` vs `...761`), and `eviction_projects_removed_from_group_notification` hit the same notification-timeout class as the flake on the earlier head. Neither touches the directory cache. Worth a tracking issue; not this PR's scope.

Approving — the migration contract is now backed by populated-cache, disk-full, and crash-recovery evidence, all synthetic and privacy-safe.

Datawav (approved):

PR #1613, opened 2026-09-01
Declare MDK wasm32 crypto features at the engine boundary

liamhelmer:
## Summary

Pin workspace getrandom to 0.4.3 and enable getrandom wasm_js plus OpenMLS js only in cgka-engine's wasm32 target dependencies.

### Pre-patch evidence

At upstream base 876bdf3c408df0658c158da6a6521745cd0abde5, the exact locked full WASM build stopped at getrandom 0.4.2 because no wasm_js backend was enabled. Native and WASM OpenMLS feature trees both lacked js, leaving consumers dependent on accidental feature unification.

### Native semantics

Native feature resolution remains unchanged: OpenMLS retains default plus extensions-draft, while js and getrandom wasm_js are absent. Native and WASM resolve the same single OpenMLS source revision.

### Verification

    cargo tree -e features -i openmls -p cgka-engine
    cargo tree --target wasm32-unknown-unknown -e features -i openmls -p cgka-engine
    cargo tree -e features -i getrandom@0.4.3 -p cgka-engine
    cargo tree --target wasm32-unknown-unknown -e features -i getrandom@0.4.3 -p cgka-engine
    cargo metadata --locked --format-version 1
    cargo fmt --all --check
    cargo test -p cgka-traits -p cgka-engine -p transport-nostr-peeler \
      -p transport-nostr-adapter -p marmot-forensics \
      --features cgka-engine/test-policy-overrides

Results: native trees contain neither OpenMLS js nor getrandom wasm_js; WASM trees contain both. Locked metadata resolves exactly one OpenMLS source at 59e7d3b27a7e95237879dd5478de1fd90eff7ada. Formatting and the full native traits/engine/peeler/adapter/forensics suite passed with zero failures.

### Scope boundary

This standalone patch does not solve portable time reads, Tokio/browser deadline ownership, or wasm32-local async trait futures. The full WASM build therefore belongs to the reviewed four-patch integration branch and is not expected from D alone. This patch does not claim browser support.

- Upstream base: 876bdf3c408df0658c158da6a6521745cd0abde5
- Isolated head: 995f374d7860bc6ab1f4d7a6a65c694776d9fc02
- Reviewed integration context: [Epiphytic/deaddrop/wasm-portability](https://github.com/Epiphytic/mdk/tree/deaddrop/wasm-portability) at 7878ccd4347af25f6fefd377a6e263e9c245602f
[...]

PR #1699, opened 2026-09-05
test: stabilize retained media epoch retirement regression

erskingardner:
The retained-media relay regression used two-second retention and a three-second sleep. Automatic pruning during sync could remove the first message before the explicit prune, or expire the later message that must preserve its source-epoch secret across a commit. A temporary three-second receiver delay reproduced the exact retired-secret failure from [the CI run on #1613](https://github.com/marmot-protocol/mdk/actions/runs/33917663143/job/101258982814?pr=1613) on master.

Use the existing retention sweep with a timestamp derived from the first message's recorded expiry, beyond the clock-skew guard, and a one-day retention window so subsequent media survives relay processing. Mark the first message read to satisfy the sweep policy and assert that pruning deletes a secret and reports the expected ciphertext hash. Production code is unchanged.

Validation:
- The original test failed with the temporary receiver delay; the fixed test passed with the same delay. The delay was removed from the final change.
- Ten consecutive runs of the final test passed.
- All 126 relay-runtime tests passed, including the other test that failed in the linked job.
- `just fast-ci` passed.

## Summary by CodeRabbit

- **Tests**
  - Improved regression coverage for retained media expiration and cleanup.
  - Verification now confirms that expired media messages, associated secrets, and encrypted media data are removed correctly.
  - Updated test timing and cleanup checks reduce intermittent failures caused by short retention windows and timing-sensitive behavior.

### DIFF
--- a/crates/marmot-app/tests/relay_runtime.rs
+++ b/crates/marmot-app/tests/relay_runtime.rs
@@ -7599,8 +7599,16 @@ async fn retained_media_rehydrates_a_retired_current_epoch_before_the_group_adva
         .create_group("retired media epoch", &["bob"])
         .await
         .unwrap();
+    let group_id_hex = hex::encode(group_id.as_slice());
     bob.sync().await.unwrap();
-    alice.update_message_retention(&group_id, 2).await.unwrap();
+    // Keep automatic wall-clock pruning out of this regression: relay sync
+    // may take longer than a short retention window, including for the later
+    // message that must remain retained across the commit.
+    let retention_seconds = 24 * 60 * 60;
+    alice
+        .update_message_retention(&group_id, retention_seconds)
+        .await
+        .unwrap();
     bob.sync().await.unwrap();
 
     let expired = alice
@@ -7625,11 +7633,32 @@ async fn retained_media_rehydrates_a_retired_current_epoch_before_the_group_adva
     let expired_source_epoch = expired_reference.source_epoch;
     bob.sync().await.unwrap();
 
-    sleep(Duration::from_secs(3)).await;
-    let pruned = bob
-        .secure_delete_expired_plaintext_for_group(&group_id)
+    let expired_message_id = &expired.sent.as_ref().unwrap().message_ids[0];
+    let expired_message = app
+        .messages("bob")
+        .unwrap()
+        .into_iter()
+        .find(|message| &message.message_id_hex == expired_message_id)
+        .expect("the first media message must be retained before the explicit sweep");
+    let retention = expired_message.retention.unwrap();
+    assert_eq!(retention.retention_seconds, retention_seconds);
+    app.mark_timeline_message_read("bob", &group_id_hex, expired_message_id)
         .unwrap();
+    // Drive the existing sweep from the durable authenticated expiry, past
+    // its five-second clock-skew guard, without advancing the real clock.
+    let prune_at_ms = (retention.expires_at.unwrap() + 6) * 1_000;
+    let report = bob.sweep_expired_retention(prune_at_ms).unwrap();
+    assert_eq!(report.groups.len(), 1);
+    let pruned = &report.groups[0];
+    assert_eq!(pruned.group_id_hex, group_id_hex);
+    assert_eq!(pruned.status, RetentionSweepStatus::Pruned);
     assert!(pruned.pruned_messages > 0);
+    assert!(pruned.secrets_deleted > 0);
+    assert!(
+        pruned
+            .media_ciphertext_sha256
+            .contains(&expired_reference.ciphertext_sha256)
+    );
     assert!(
         bob.download_media(&group_id, expired_reference.clone())
             .await


crates/marmot-app/tests/relay_runtime.rs
```diff
@@ -7625,11 +7633,32 @@ async fn retained_media_rehydrates_a_retired_current_epoch_before_the_group_adva
...
+        .into_iter()
+        .find(|message| &message.message_id_hex == expired_message_id)
+        .expect("the first media message must be retained before the explicit sweep");
+    let retention = expired_message.retention.unwrap();
+    assert_eq!(retention.retention_seconds, retention_seconds);
+    app.mark_timeline_message_read("bob", &group_id_hex, expired_message_id)
+        .unwrap();
+    // Drive the existing sweep from the durable authenticated expiry, past
+    // its five-second clock-skew guard, without advancing the real clock.
+    let prune_at_ms = (retention.expires_at.unwrap() + 6) * 1_000;
```
erskingardner, line 7649: Non-blocking: the `+ 6` here is load-bearing and silently coupled to `RETENTION_SWEEP_CLOCK_SKEW_TOLERANCE_MS` (5_000 ms, private in `crates/marmot-app/src/client/retention.rs`). The margin works out to exactly one second: `skew_cutoff = (prune_at_ms - 5_000)/1000 - retention = timeline_at + 1`, and the row's `timeline_at` is `expires_at - retention_seconds`, so `classify_retention_page` lands on `SafelyExpired` deterministically. That is fine today, and a future tolerance bump fails loudly on the `status == Pruned` assert rather than silently — just noting that the comment's "five-second" figure is the only link back to the constant.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T09:35:31Z` |
| Commit reviewed | `64547f7fbe38158b13f3b269f6abe3eb88603d6d` |
| Recommended action | **Merge** (posted as a comment: GitHub blocks self-approval) |

Test-only, one function, production code untouched. The premise holds and the fix is the right shape.

**Why the new timing is deterministic (verified against the source, not just the description)**

- `AppMessageRetentionDecision::new(recorded_at, retention_seconds)` sets `expires_at = recorded_at + retention_seconds`, ingest passes `app_event.created_at`, and the timeline row's `timeline_at` is that same `recorded_at` (`crates/storage-sqlite/src/timeline.rs:3410`). So `expires_at - retention_seconds == timeline_at` exactly, and the injected `prune_at_ms` is a pure function of durable state rather than of wall clock.
- With `prune_at_ms = (expires_at + 6) * 1000`: `raw_cutoff = timeline_at + 6`, `skew_cutoff = timeline_at + 1`. `classify_retention_page` therefore returns `SafelyExpired` (not `DeferredClockSkew`), the scan completes on the first page with `has_more_before == false`, and `RetentionScanAction::Prune` calls `secure_delete_expired_plaintext_for_group_at` with `raw_now_seconds > retention_expires_at`, which satisfies the `retention_expires_at <= ?2` predicate. No sleep, no clock dependence.
- `mark_timeline_message_read` is genuinely required here (unlike the sibling `app_runtime_retention_sweep_prunes_with_the_supplied_clock`, which sweeps on the sender side): the media message is `direction == "received"` for bob, so `DeferredUnread` would otherwise fire. Earlier inbound rows (welcome/system/`disappearing_timer_changed`) all sit at or below the new `last_read_timeline_at`, so they cannot re-trigger the deferral.
- `report.groups.len() == 1` is safe: bob is in exactly one group, and `run_retention_group_sweeps` drops zero-retention groups without emitting an outcome.

**Scope and shape**

- The 24-hour window is the correct lever: it removes automatic wall-clock pruning during relay sync from the picture entirely, which is what let CI expire either the first message early or the second message that must carry its source-epoch secret across the commit.
- This removes the last `sleep(Duration::from_secs(..))` and the last short-retention window in `relay_runtime.rs`; every other retention test already uses 60s or an injected clock. Nice cleanup of a real flake class rather than a one-off patch.
- The regression's actual claim is unchanged and arguably better anchored: `secrets_deleted > 0` and the `media_ciphertext_sha256` containment now prove the retired epoch secret was really deleted, instead of leaving the subsequent `download_media` failure to be satisfied by any error. The extra assertions are on the critical path of the claim, not decoration.
- Follows the established `sweep_expired_retention(now_ms)` pattern already used at `crates/marmot-app/tests/relay_runtime.rs:7010`.

**Minor, non-blocking**

- One inline note on the `+ 6` constant's coupling to the private clock-skew tolerance.
- Line 7630 was the only integration-test caller of the public wall-clock `AppClient::secure_delete_expired_plaintext_for_group`. It is a one-line wrapper over `_at`, the automatic path is still exercised through `prune_plaintext_retention_for_group` on the sync paths, and storage-level pruning has unit coverage in `crates/storage-sqlite/src/account_projection/tests.rs`, so the coverage loss is nil in practice. Flagging only so it is a deliberate choice.

CI is green across the matrix, including the relay-runtime job. No blocking issues.

PR #1701, opened 2026-09-06
feat(hermes): add opt-in reaction consent for approval prompts

liamhelmer-bel:
## Summary

Adds opt-in approval reactions to the Hermes Marmot adapter: 👍 approves once, 👎 denies, and ❤️ approves permanently when the pending request permits it. `platforms.marmot.extra.approval_reactions` defaults to false; disabled prompts retain standard `send()` metadata and preview handling.

## Safety and compatibility

Reactions bind a durable prompt receipt to the actual pending entry in the pinned Hermes approval queue. They resolve that entry under the host lock; they do not dispatch unscoped slash commands. Expired, already-resolved, or ambiguous parallel prompts cannot approve a different request. Rejected typed commands leave the prompt usable.

Only explicitly allowed senders may decide prompts. Empty allowlists fail closed, self-reactions are ignored, and permanent consent is unavailable for restricted requests. Tracking is bounded and logs exclude identifiers and payloads.

Supported host: Hermes 0.19.0 at `3ef6bbd201263d354fd83ec55b3c306ded2eb72a`. The pinned-host CI job now runs the approval contract suite. Newer host signatures require separate compatibility validation.

## Validation

- `just fast-ci`: passed.
- Adapter suite: 361 tests, passed (11 opt-in contract tests skipped).
- Pinned-host approval suite: all 11 passed, including disabled delivery/metadata regression.
- Offline local dogfood: staged patched hook plus AST-extracted installed monolithic adapter `send()` and installed gateway formatter passed commentary metadata and active-preview finalization checks. Transport was mocked; no live restart, deployment, or chat sends.
- Local host inspection found a newer `allow_session` hook argument, so full live notification/reaction compatibility is not claimed.

Exact-head CI requires maintainer workflow approval.


Datawav (changes requested):
## Automated review

Actionable findings:
1. [P1] Reactions are not bound to the approval they target
Location: integrations/hermes/marmot/adapter.py:3730
Trigger: Tracked prompt A is resolved through a typed command or expires, approval B becomes pending in the same group, and an allowed user later reacts 👍 to the still-tracked message for A.
Impact: The target identity is discarded when the reaction becomes an unscoped `/approve`, so the gateway can approve B—a different dangerous command than the one the user selected.
Fix: Preserve a stable approval-request identifier with each tracked message and dispatch a decision explicitly correlated to that request; invalidate records on every resolution or expiry path. Add a regression test where reacting to stale prompt A cannot resolve current prompt B.
Confidence: high

Datawav (changes requested):
## Automated review

Actionable findings:
1. [P2] Rejected slash commands discard valid reaction consent
Location: integrations/hermes/marmot/adapter.py:3772
Trigger: While an approval prompt is tracked, a group member sends `/approve` or `/deny` but the gateway rejects the command for authorization, syntax, or other validation.
Impact: The prompt is invalidated before gateway validation, so an authorized user’s later reaction is ignored even though no approval decision occurred; any member reaching this path can disable reaction consent for the pending prompt.
Fix: Invalidate typed-command prompts only through the existing successful-resolution callback (`resume_typing_for_chat`), not before dispatch. Add a regression test proving a rejected typed command leaves the prompt eligible for a subsequent authorized reaction.
Confidence: high

Datawav (approved):

integrations/hermes/marmot/adapter.py
```diff
@@ -2075,6 +2126,43 @@ async def send(
...
         chat_id = _normalize_hex(chat_id, "chat_id")
+        if _APPROVAL_REPLY_CONTEXT.get() == (self, chat_id):
+            try:
+                reply_to = _optional_hex(reply_to)
+            except AgentControlError:
+                # Only the reply anchor is optional. Retry its send preparation
+                # unthreaded; invalid destination/account ids remain errors.
+                logger.debug("Marmot approval confirmation dropping invalid reply anchor")
+                reply_to = None
+        if self.approval_reactions and (metadata or {}).get("is_approval_prompt") is True:
```
erskingardner, line 2137: [P1] Connect prompt tracking to the supported Hermes approval hook

The repository pins Hermes 0.19.0 at `3ef6bbd201263d354fd83ec55b3c306ded2eb72a`, whose `_approval_notify_sync` calls `adapter.send(chat_id, msg, metadata=_status_thread_metadata)` without `is_approval_prompt`; this adapter also does not implement the alternative `send_exec_approval` hook. Consequently real dangerous-command prompts never enter this branch and reactions never resolve them, even with approval_reactions=true. Every new test supplies the missing marker itself. Implement the supported approval hook (which also carries the session and approval capabilities), or ship and require a host contract that actually supplies the marker, and add a real pinned-host notification-to-reaction regression. [Pinned Hermes sender](https://github.com/NousResearch/hermes-agent/blob/3ef6bbd201263d354fd83ec55b3c306ded2eb72a/gateway/run.py#L20690-L20714).

integrations/hermes/marmot/adapter.py
```diff
@@ -3616,7 +3704,145 @@ async def _flush_debounced(self, key: str) -> None:
...
+            source=source,
+            raw_message={"via": "approval_reaction"},
+            # Hermes uses event.message_id (not source.message_id) as the
+            # confirmation reply anchor. Preserve the durable reaction id;
+            # a synthetic "approval-reaction-..." token cannot be a Marmot id.
+            message_id=reaction_id,
+            timestamp=datetime.now(timezone.utc),
+        )
+        try:
+            await self._handle_gateway_message(hermes_event)
```
erskingardner, line 3836: [P1] Resolve the targeted approval atomically instead of an unscoped slash command

Consuming the local message entry prevents duplicate reactions, but does not bind the eventual decision to that approval. For a tracked prompt A, a reaction can pass validation and enter this await; the real BasePlatformAdapter.handle_message yields in topic recovery before dispatch. During that gap a typed decision can resolve A and the agent can enqueue B. When this reaction resumes, Hermes resolve_gateway_approval pops the current FIFO head B, so a thumb/heart on A approves a different operation (possibly permanently). Invalidating local records on resume_typing_for_chat cannot revoke the already-built MessageEvent. Preserve a stable request/session identity through a host API that checks and resolves that exact pending approval atomically; add a regression that pauses reaction dispatch, resolves A, queues B, and proves B remains pending. See [gateway dispatch](https://github.com/NousResearch/hermes-agent/blob/3ef6bbd201263d354fd83ec55b3c306ded2eb72a/gateway/platforms/base.py#L4773-L4779) and [FIFO resolver](https://github.com/NousResearch/hermes-agent/blob/3ef6bbd201263d354fd83ec55b3c306ded2eb72a/tools/approval.py#L2089-L2105).

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T17:33:15Z` |
| Commit reviewed | `a0651c2a4a5558ec452a551aaf7b8f747eb13930` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

The explicit opt-in, fail-closed sender allowlist, and rejection of self-reactions are appropriate. However, the feature is not connected to the actual approval-prompt contract of the repository-pinned Hermes release, and message-to-approval correlation remains unsafe across asynchronous gateway dispatch. Both need fixing before this can grant dangerous-command consent.

Please exercise the real pinned gateway approval notification and resolver in regression coverage, including a typed decision racing a reaction, rather than replacing handle_message with a fake successful callback. Document the flag, reaction meanings, and supported Hermes version in the plugin README once the host contract is established.

GitHub currently reports MERGEABLE / BLOCKED. Both exact-head CI and wn-agent Binaries require workflow approval (action_required); no completed build/test validation is available. Review was remote source inspection, including the Hermes revision in hermes-agent.lock, plus existing CI; no local tests were run.

Additional merge-gate verification: all commits in this PR are marked unverified by GitHub with reason `no_user`. The effective master rules require verified signatures and the `Required CI` check. Resolve the signing identity/verification and obtain passing Required CI before merging; CodeRabbit success does not satisfy that rule.

Datawav (dismissed):

Datawav (changes requested):
## Automated review

Actionable findings:
1. [P2] Preserve the standard approval-send path when reactions are disabled
Location: integrations/hermes/marmot/adapter.py:3678
Trigger: `approval_reactions` is absent or false and pinned Hermes emits a dangerous-command approval while normal Marmot send metadata or preview state is active.
Impact: The host selects the newly defined hook regardless of the flag, and this line always bypasses `send()` via `_send_final_direct` while ignoring `metadata`; opting out therefore changes approval routing and stream lifecycle and can produce an unthreaded prompt or leave an active preview inconsistent.
Fix: When the flag is false, send the formatted fallback through `self.send(chat_id, content, metadata=metadata)` and reserve direct receipt tracking for the enabled path; add a real-host regression asserting flag-off notification uses the normal send path and preserves metadata/preview handling.
Confidence: high

liamhelmer-bel:
@Datawav @erskingardner Please re-review f6f5bba3. Rebased and signed. Reaction consent now uses the pinned Hermes 0.19.0 send_exec_approval notification and binds the rendered request to its exact pending entry; consumption shares the real typed resolver lock. Stale reactions cannot decide replacement requests, rejected commands preserve valid consent, and permanent reactions respect request capabilities. README documents opt-in, emoji meanings and supported host revision. Validation: just fast-ci; adapter suite 356 tests OK (6 contract tests skipped in default discovery); explicit pinned-host contract suite expanded to 10 tests, all passed. Formal reviewer assignment returned HTTP 404; requesting here.

Datawav (approved):

integrations/hermes/marmot/adapter.py
```diff
@@ -3569,7 +3645,169 @@ async def _retry_pending_debounce_releases(self) -> None:
...
+        content = _format_exec_approval_fallback(
+            command, description, "/", allow_permanent=allow_permanent,
+            smart_denied=smart_denied,
+        )
+        if self.approval_reactions and entry is not None:
+            content += "\nReact 👍 to approve once or 👎 to deny."
+            if allow_permanent and not smart_denied:
+                content += " React ❤️ to approve permanently."
+        started = time.monotonic()
+        result = await self._send_final_direct(chat_id, content)
```
erskingardner, line 3686: [P2] Preserve normal approval delivery when the feature is disabled

Hermes selects send_exec_approval by the presence of the class method, regardless of approval_reactions. This unconditional _send_final_direct call therefore changes the default/flag-off path too: metadata is discarded and send() no longer finalizes or cancels an active preview before publishing the prompt. An existing live preview can remain active while the agent is blocked waiting for approval. Return through self.send(chat_id, content, metadata=metadata) when the flag is false, before using the reaction-specific direct-send path. The current flag-off host test only asserts that reactions do not resolve a request; it mocks _send_final_direct and never checks metadata or preview cleanup, so it misses this regression.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-18T10:03:23Z` |
| Commit reviewed | `839676f234492eb12e13ee44692b5406bee55053` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

The two original P1 findings are addressed: send_exec_approval connects to the pinned host notification path, and reactions resolve a specific pending entry under the same lock used by the typed resolver. The tests now exercise the actual host notification closure, queue and slash handlers; restricted permanent consent is also checked. This is a substantial improvement.

One P2 remains: the new hook changes approval delivery even when approval_reactions is false (inline). The previous automated review raised this too, and it is still present at this exact head despite the later approval. Please preserve the standard send/metadata/preview path when disabled and test that path, then this is close to approval.

The dependency on private Hermes queue internals is explicit and pinned, but the host-contract suite is opt-in and skipped by normal discovery. I recommend putting that suite in the pinned-host CI job so a future host update cannot silently invalidate this safety boundary. Also update the PR body: reactions no longer dispatch unscoped slash commands.

All commits are now verified. Exact-head CI and wn-agent Binaries are still action_required; older passing runs do not validate the revised implementation. Remote source/CI inspection only, no local test execution.

Datawav (changes requested):
## Automated review

Actionable findings:
1. [P2] Preserve normal approval delivery when reactions are disabled
Location: integrations/hermes/marmot/adapter.py:3676
Trigger: With approval_reactions omitted or false, a dangerous command requests approval while a Marmot live preview is active.
Impact: The new host hook unconditionally calls _send_final_direct instead of the normal send path, bypassing preview finalization/cancellation and metadata handling; the stale preview can remain active while approval waits, changing default behavior.
Fix: Before the reaction-specific direct-send path, return await self.send(chat_id, content, metadata=metadata) when approval_reactions is false, and add a flag-off regression with an active preview.
Confidence: high

liamhelmer-bel:
@erskingardner @Datawav Addressed round-2 feedback in 32f56a68:

- Disabled approval reactions now use standard `send()` with original metadata and existing preview handling; added regression coverage.
- Wired all 11 pinned-host approval contracts into the pinned Hermes CI job.
- Corrected the PR body: reaction consent resolves the bound pending entry under the host lock, without unscoped slash commands.

Validation: `just fast-ci` passed; adapter suite 361 tests passed (11 opt-in skips); pinned-host contracts 11/11 passed.

Local dogfood also passed using the patched hook with AST-extracted code from this host's installed monolithic adapter and gateway formatter: commentary metadata reached the activity path, and an active preview reached normal finalization, with no direct-send bypass or consent tracking. Transport was mocked; no services restarted/redeployed and no live chat messages sent. Repro/log: `/tmp/mdk-r2-dogfood.py`, `/tmp/mdk-r2-dogfood.log` on the worker host. Static inspection found the installed gateway's newer `allow_session` argument, which this PR's pinned hook does not support; full live notification/reaction compatibility is not claimed.

Exact-head CI runs need maintainer workflow approval.

Datawav (approved):

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-20T10:42:07Z` |
| Commit reviewed | `32f56a6896ba3692a2b901dc6bbbc96925d66a00` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

The new early return closes the remaining blocking finding: with approval_reactions=false, the hook uses the pinned host's standard formatter and delegates to send(chat_id, content, metadata=metadata), restoring metadata routing and the existing preview finalization/cancellation path. The new contract test verifies delegation, metadata, returned result, and absence of reaction tracking. The pinned-host suite is now explicitly enabled in the Hermes installer CI job; its interpreter/environment match the dev-setup script and the job feeds Required CI.

The previously reviewed exact-entry binding and host-lock resolution remain intact. No new merge-blocking finding in this update. Approving for the documented Hermes 0.19.0 pin; compatibility with newer hook signatures remains explicitly unvalidated.

Non-blocking: CodeRabbit's existing confirmation SendResult finding is valid. After a successful decision, exhausted confirmation delivery returns success=false instead of raising, so the warning path is missed. Checking that result and warning would improve visibility; do not re-arm an already-consumed approval. This does not invalidate the approval decision or reopen the authorization finding.

All six commits verify on GitHub and the PR has no merge conflict. Required CI has not passed for this head: CI and wn-agent Binaries are action_required pending workflow approval. Source/API re-review only; I did not execute local tests or live gateway traffic. Merge after required CI passes, and validate the combined adapter when integrating #1702.

PR #1702, opened 2026-09-06
feat(hermes): add opt-in deterministic presence reactions

liamhelmer-bel:
Opt-in deterministic presence reactions for Hermes Marmot (`platforms.marmot.extra.presence_reactions: true`). Processing hooks show accepted/thinking/tool activity and terminal success, failure, cancellation or supersession. Tool activity cycles construction emoji on bounded timers; network work is queued with bounded state/retries.

Inbound messages move the active indicator only after queue admission, activation, onboarding, and an explicit allow from the host's full sender-authorization callback. Denied, missing, or unknown authorization leaves the indicator unchanged. Tool status carries its originating turn through the pinned Hermes executor context, so late callbacks cannot change a replacement turn. Successful completion marks the original reply anchor.

Configuration and behavior are documented in `integrations/hermes/marmot/README.md`. No external design document is required.

Validation on fa678e7e:
- `just fast-ci` passed.
- Full Hermes adapter discovery: 363 tests, OK (2 opt-in host contracts skipped).
- Presence socket/admission suite: 12 tests passed, including activated denied/unknown senders during an active turn.
- Pinned Hermes host contracts: 2 tests passed (full authorization callback and late executor callback).
- Local dogfood against this host's installed Hermes: authorization regression passed with the real GatewayRunner callback, authorization chain and BasePlatformAdapter helper driving the patched adapter. An isolated runner and captured reaction operations exercised denied and allowed senders; this did not exercise live gateway traffic or network delivery. No service restart or deployment.

Commits are signed. This PR remains independent of #1701. Exact-head CI runs require maintainer workflow approval.

## Summary by CodeRabbit

* **New Features**
  * Added optional presence reactions for Marmot conversations, showing progress while messages are processed.
  * Reactions indicate thinking, tool activity, completion, failure, cancellation, and other turn states.
  * The feature is disabled by default and supports configurable emoji overrides.

* **Documentation**
  * Added configuration and behavior details for enabling and customizing presence reactions.

* **Tests**
  * Added coverage for reaction lifecycle, retargeting, authorization, retries, cleanup, and configuration limits.


Datawav (approved):

integrations/hermes/marmot/adapter.py
```diff
@@ -3334,6 +3584,9 @@ async def _handle_control_event(self, event: Dict[str, Any]) -> None:
         ):
             return
         self._pending_inbound_ids.add(message_id_hex)
+        if (self._presence.enabled and event.get("account_id_hex") == self.account_id_hex
+                and event.get("sender_account_id_hex") != self.account_id_hex):
+            self._presence.retarget(event["group_id_hex"], message_id_hex)
```
erskingardner, line 3589: [P2] Retarget only after the inbound message is accepted for processing

This runs before queue admission and `_should_run_turn`, and before Hermes sender authorization. During an active turn in a multi-party mention-only group, an ordinary unaddressed message now removes the working reaction from the actual trigger and adds it to that ignored message; `_dispatch_inbound_message` subsequently drops the message, leaving the indicator on work the agent never accepted. A disallowed sender or a message shed by the bounded queue can do the same. Move retargeting to a point with successful activation/authorization/admission evidence. Add a regression with an active tool and an unmentioned follow-up that leaves the original indicator untouched.

integrations/hermes/marmot/adapter.py
```diff
@@ -2223,6 +2444,31 @@ async def edit_message(
...
+
+    def set_status_text(self, chat_id: str, text: Optional[str]) -> None:
+        loop = self._loop
+        state = self._presence.groups.get(chat_id)
+        if not self._presence.enabled or loop is None or not loop.is_running() or state is None:
+            return
+        # The gateway calls this from its agent thread. Never retain or log the
+        # status phrase; only its presence (tool active vs between tools) matters.
+        try:
+            loop.call_soon_threadsafe(self._presence.status, chat_id, text is not None, state.owner)
```
erskingardner, line 2461: [P2] Carry the emitting turn identity into status callbacks

The owner guard in PresenceReactions.status is defeated by reading `state.owner` from the current group slot here. If turn A is superseded by B, B starts a tool, then A emits its late tool.completed callback, this call stamps that callback with B’s owner. status accepts it, decrements B’s active_tools to zero, switches to thinking, and cancels the construction timer while B’s tool is still running. The pinned Hermes progress_callback explicitly forwards tool.completed without the `_run_still_current()` guard applied to tool.started, so this is a real host path. Capture the originating turn identity from its execution context rather than the mutable group state and reject stale updates; extend the supersession test with a late old-turn status callback, not only on_processing_complete. [Pinned host callback](https://github.com/NousResearch/hermes-agent/blob/3ef6bbd201263d354fd83ec55b3c306ded2eb72a/gateway/run.py#L19166-L19177).

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T17:37:15Z` |
| Commit reviewed | `f28150e37d23c33a9dd78e2ac4f6894b0da71463` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

The feature is appropriately opt-in and cosmetic: bounded workers, queues, retries, exact-emoji removal, and timer cancellation keep reaction failures out of the reply path. The pinned Hermes release supplies the processing and live-status hooks used here.

Two correctness gaps remain in which message/turn owns the indicator: retargeting happens before activation/admission, and late old-turn tool callbacks are relabeled with the current owner. Add regressions through the admission and actual host callback paths; the current tests exercise lifecycle hooks directly and only guard stale completion hooks.

This is independently based on master, but overlaps #1701 in adapter.py; validate both flags together after reconciling the changes. Non-blocking documentation cleanup: the design document linked in the PR body (`docs/plans/presence-react-system.md`) does not exist at this head, and the body still describes an older implementation/test count.

Merge readiness: GitHub reports MERGEABLE / BLOCKED. Both exact-head CI and wn-agent Binaries await workflow approval (action_required), with no completed build/test evidence. The commit is unverified with reason no_user; master has required_signatures and Required CI rules, so resolve signing identity/verification and obtain the required green check before merging. Review used remote exact-head sources and existing CI only; no local tests were run.

Datawav (dismissed):

liamhelmer-bel:
@Datawav @erskingardner Please re-review f1ced2e1. Rebased and signed. Retargeting now follows admission/activation/onboarding; status callbacks retain originating-turn context, verified through the pinned Hermes callback and executor. Added actual inbound admission regressions and reconciled adapter initialization with #1701. Validation: just fast-ci; full adapter suite 361 tests OK (1 skipped); presence socket/admission 11 passed; pinned-host callback regression passed; locally merged combined-feature host checks passed. Formal review assignment is unavailable to this credential (HTTP 404), so requesting here.

Datawav (changes requested):
## Automated review

Actionable findings:
1. [P2] Unauthorized inbound messages can retarget an active presence indicator
Location: integrations/hermes/marmot/adapter.py:3440
Trigger: With presence enabled and a turn active, a message that passes local queue, activation, and onboarding checks but whose sender Hermes subsequently rejects under its sender allowlist reaches this retarget call before host authorization.
Impact: The working reaction moves to an unauthorized message and away from the accepted turn, exposing agent activity and falsely indicating that the rejected message is being processed.
Fix: Retarget only from a post-authorization host path using the same authorization decision as turn admission; add an active-tool regression where an unauthorized inbound message leaves the original target unchanged.
Confidence: high

Datawav (dismissed):

integrations/hermes/marmot/adapter.py
```diff
@@ -3182,6 +3435,10 @@ async def _dispatch_inbound_message(
                     spool_state = "intentionally_skipped"
                 return
 
+            # This unit was admitted by the per-group queue and passed activation
+            # and onboarding. Rejected/ambient inbound events cannot move presence.
+            self._presence.retarget(group_id_hex, message_id_hex)
```
erskingardner, line 3440: [P2] Authorize the sender before retargeting active presence

Queue admission and `_should_run_turn()` do not establish sender authorization. With an authorized turn active, a disallowed group member can send a message that passes activation (for example, by mentioning the agent), reach this `retarget()`, and move the working/tool reaction to their own message before `handle_message()` eventually rejects them. The pinned host explicitly checks authorization in both its busy-session path and normal message path ([busy gate](https://github.com/NousResearch/hermes-agent/blob/3ef6bbd201263d354fd83ec55b3c306ded2eb72a/gateway/run.py#L5715-L5730), [normal gate](https://github.com/NousResearch/hermes-agent/blob/3ef6bbd201263d354fd83ec55b3c306ded2eb72a/gateway/run.py#L10041-L10059)). Please use the host-backed authorization decision before retargeting (or a post-authorization lifecycle hook), and test that a disallowed sender cannot move an active turn's presence even when activation succeeds.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-18T10:05:58Z` |
| Commit reviewed | `8107b3c724c05b71ee61f8458236da3655af40bc` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

Quick re-review of the updated head: the originating-turn ContextVar plus the pinned-host executor regression test addresses the stale tool-callback finding. Moving retargeting after queue admission, activation, and onboarding also improves the earlier behavior. One part of the original retargeting finding remains: sender authorization still happens later in Hermes, after presence has already been moved (inline below).

Please gate retargeting on the host's actual authorization decision and add an active-turn regression with a disallowed sender whose message passes activation. The expected result is no target change and no reaction to that message. The host already provides an adapter authorization callback backed by its full authorization chain, so this should not require a separate allowlist implementation.

The feature remains opt-in, and the bounded state machine and exact-reaction cleanup are sensible. This was a source/API re-review; I did not run local tests. The author's reported tests are useful evidence, but exact-head CI and wn-agent Binaries runs currently require workflow approval. Verified commit signatures are now present. Request changes for the remaining presence-authorization bug; after that, require passing CI before merge.

liamhelmer-bel:
@erskingardner @Datawav Round-2 fix pushed as signed commit fa678e7e77e82698df5b926508d89926be5bf33e.

Presence retargeting now requires explicit True from the host's full authorization callback after admission/activation/onboarding. The active-turn regression verifies an activated disallowed sender causes neither a target change nor a reaction; unknown decisions also leave presence unchanged. The PR body and README describe the gate.

Validation: `just fast-ci` passed; full adapter suite 363 tests OK (2 opt-in host tests skipped); presence socket suite 12 passed; pinned Hermes contracts 2 passed. Local dogfood against our installed Hermes also passed: the real GatewayRunner authorization callback/chain and BasePlatformAdapter helper drove the patched adapter for denied and allowed senders. This used an isolated runner, test allowlist, no pairing record, and captured reaction operations; it did not exercise live gateway traffic/network delivery. No live service restart or redeployment occurred.

Exact-head CI runs need maintainer workflow approval. Please re-review this head.

Datawav (approved):

integrations/hermes/marmot/adapter.py
```diff
@@ -1878,6 +2101,32 @@ async def edit_message(
...
+        # status phrase; only its presence (tool active vs between tools) matters.
+        try:
+            loop.call_soon_threadsafe(self._presence.status, chat_id, text is not None, owner)
+        except RuntimeError:
+            logger.debug("Marmot presence scheduling skipped during shutdown")
+
+    async def on_processing_start(self, event: MessageEvent) -> None:
+        self._capture_loop()
+        _PRESENCE_OWNER.set(event)
+        self._presence.start(event.source.chat_id, event.message_id, event)
```
erskingardner, line 2134: [P2] Authorize the sender before starting presence

The new check in `_dispatch_inbound_message` only guards `retarget`; denied messages still reach `handle_message`. For an idle host session, the pinned [BasePlatformAdapter calls this hook before `_message_handler`](https://github.com/NousResearch/hermes-agent/blob/3ef6bbd201263d354fd83ec55b3c306ded2eb72a/gateway/platforms/base.py#L5033-L5037), and that handler is where [GatewayRunner rejects unauthorized senders](https://github.com/NousResearch/hermes-agent/blob/3ef6bbd201263d354fd83ec55b3c306ded2eb72a/gateway/run.py#L10041-L10060). Thus a disallowed member whose message passes activation still gets an accepted reaction here. The rejection returns None, which the host [classifies as successful completion](https://github.com/NousResearch/hermes-agent/blob/3ef6bbd201263d354fd83ec55b3c306ded2eb72a/gateway/platforms/base.py#L5396-L5401), leaving a completed checkmark on a message the agent refused. With separate per-user sessions it can also replace another sender's group presence. Require the same explicit host authorization before `_presence.start`, and add a denied-sender regression using the real host lifecycle rather than the stub `handle_message` used by the current contract test.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-20T10:44:02Z` |
| Commit reviewed | `fa678e7e77e82698df5b926508d89926be5bf33e` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

The update correctly closes the active-turn retargeting path: only an explicit True from the host's complete authorization callback moves the target; False/None remain unchanged, and the new host-callback contract exercises both allowed and denied senders. The originating-turn callback fix also remains intact.

However, the same authorization boundary is still bypassed when starting presence. The pinned BasePlatformAdapter invokes on_processing_start before GatewayRunner performs authorization. The new retarget guard does not reject dispatch or guard the lifecycle hook, so an activated but unauthorized message in an idle host session still creates presence and can receive a completed checkmark after the host silently rejects it. This is the remaining P2 blocker described inline.

The new host authorization test borrows the real authorization callback/helper, but load_adapter_module still supplies a stub handle_message that only records events. It therefore never exercises the host's background lifecycle ordering. Add a regression through that actual dispatch/background path for a denied sender, asserting no start/completion reactions and no replacement of existing presence. Guard presence initialization with the same explicit host authorization decision; preserve ordinary host dispatch so its denial/pairing behavior remains authoritative.

Non-blocking validation improvement: enable HERMES_PRESENCE_CONTRACT in pinned-host CI, as #1701 now does for its approval contracts. Currently normal discovery skips both presence host tests. All four commits verify on GitHub and no merge conflict is reported, but CI and wn-agent Binaries remain action_required. Source/API review only; no local tests were run. Request changes for the lifecycle authorization gap, then require green CI and combined-feature validation with #1701.

Datawav (changes requested):
## Automated review

Actionable findings:
1. [P2] Denied senders can start presence reactions
Location: integrations/hermes/marmot/adapter.py:2125
Trigger: An activated inbound message from a sender whose host authorization decision is False or unknown reaches the processing lifecycle; `on_processing_start` unconditionally starts presence before the host handler rejects the sender.
Impact: The denied message receives a misleading accepted reaction and can supersede existing group presence despite not being authorized for processing. No physical-device evidence exists for this head because no trusted exact-head Android artifact exists.
Fix: Require `_is_sender_authorized(event.source.user_id, event.source.chat_type, event.source.chat_id) is True` before setting the presence owner or calling `start`; add a regression through the real host lifecycle asserting a denied sender emits no reactions and does not replace existing presence.
Confidence: high

liamhelmer-bel:
@Datawav @erskingardner Fixed in signed commit 728dca657c7a8e3a7dad3af9ecbe56d700c67fe3.

`on_processing_start` now requires explicit True from the existing host authorization helper before setting `_PRESENCE_OWNER` or starting/replacing presence. Normal dispatch is unchanged so the host retains rejection/pairing behavior. Regression coverage uses the pinned host's real `handle_message` and background lifecycle into `GatewayRunner` rejection, with idle/existing presence and denied/unidentified senders. The old startup hook fails the regression. CI now enables the presence host contracts.

Reconciliation with #1701 (32f56a68) is conflict-free; combined-base approval/presence contracts pass. On this PR head, fast-ci and all three pinned host contracts pass. Full discovery ran 365 tests with 3 skips and nine media failures; the same nine reproduce on original head fa678e7e. The PR body now records exact counts and local dogfood limits. Required workflows still need maintainer approval; please re-review the authorization fix.

integrations/hermes/marmot/adapter.py
```diff
@@ -1333,6 +1335,224 @@ async def cancel(self, reason: Optional[str] = None) -> None:
...
+                if idle is None:
+                    logger.debug("Marmot presence group capacity reached")
+                    return
+                self.groups.pop(idle)
+            state = _PresenceGroup(owner, target)
+            self.groups[group] = state
+        else:
+            self._cancel_timer(state)
+            if not state.ended:
+                self._enqueue(group, state, state.target, self.emojis["superseded"], True)
```
erskingardner, line 1429: Non-blocking: the superseded arrow goes on `state.target`, but `finish()` puts the terminal reaction on `state.anchor`. After a mid-turn retarget, a supersession puts the arrow on the follow-up message, which is usually the message that is superseding this turn, and the turn's own message ends up with no terminal marker. README says "superseded turns get an arrow". `state.anchor` matches that and matches `finish()`. `_drain` still removes the transient from `shown[0]`, so switching the target here is safe.

integrations/hermes/marmot/adapter.py
```diff
@@ -1333,6 +1335,224 @@ async def cancel(self, reason: Optional[str] = None) -> None:
...
+        if not self.enabled or self.closed or state is None or state.ended:
+            return
+        target = self._target(target)
+        if target and target != state.target:
+            state.target = target
+            emoji = (self.construction[state.index] if state.mode == "tool"
+                     else self.emojis[state.mode])
+            self._enqueue(group, state, target, emoji)
+
+    def status(self, group: str, working: bool, owner: object = None) -> None:
```
erskingardner, line 1446: Nit: the only caller (`set_status_text`) returns early when the owner is `None`, so the `owner is not None` bypass below never runs. Making `owner` required and always comparing it removes an ownerless path that the README says is ignored anyway ("Callbacks without turn context are ignored").

integrations/hermes/marmot/adapter.py
```diff
@@ -1915,6 +2173,10 @@ def render_message_event(self, event: Any, sink: Any) -> None:
             return super().render_message_event(event, sink)
 
         if isinstance(event, MessageChunk):
+            chat_id = getattr(sink, "chat_id", "")
+            presence = self._presence.groups.get(chat_id)
+            if presence is not None and presence.mode != "tool" and event.text:
```
erskingardner, line 2178: Nit: once the mode is `thinking`, every streamed chunk still does a `call_soon_threadsafe` hop into `status()`, which just returns early. The only transition this needs is `accepted` → `thinking`, so `presence.mode == "accepted"` does the same job without a loop hop per token.

integrations/hermes/marmot/adapter.py
```diff
@@ -3201,6 +3463,12 @@ async def _dispatch_inbound_message(
                 # replying to (mirrors dispatch.ts replyToMessageIdHex = inbound id).
                 message_id=message_id_hex,
             )
+            # Activation is not authorization. Use the host's full authorization
+            # callback; absent/unknown decisions must not move an active indicator.
+            authorization_check = getattr(self, "_is_sender_authorized", None)
+            if (self._presence.enabled and authorization_check is not None
+                    and authorization_check(source.user_id, source.chat_type, source.chat_id) is True):
+                self._presence.retarget(group_id_hex, message_id_hex)
```
erskingardner, line 3471: Nit: this runs the host's full authorization chain on every activated inbound message, even when no presence is active for the group. `retarget()` is a no-op in that case. Checking `group_id_hex in self._presence.groups` (and not `ended`) before calling the callback skips that work. The two call sites (here and `on_processing_start`) could also share one small `_presence_authorized(source)` helper, so the explicit-`True` rule lives in one place.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T06:30:57Z` |
| Commit reviewed | `728dca657c7a8e3a7dad3af9ecbe56d700c67fe3` |
| Model | `claude-opus-5-5[1m]` |
| Reasoning level | `10` |

Re-review of the authorization fix. The remaining blocker from the prior round is closed. `on_processing_start` now requires an explicit `True` from `_is_sender_authorized` before it sets `_PRESENCE_OWNER` or calls `start()`. I checked this against pinned Hermes `3ef6bbd`. `GatewayRunner` registers `_make_adapter_auth_check` on every adapter, so allow-all deployments still return `True`, and `None` only appears when no user id is present or no check is registered. `test_real_background_lifecycle_does_not_start_denied_presence` goes through the real `handle_message` → `_process_message_background` path with idle and existing presence, and CI now runs it with `HERMES_PRESENCE_CONTRACT=1`.

I also traced the other host seams. `progress_callback` fires `set_status_text` from the executor thread with the copied context. `GatewayEventDispatcher` calls `render_message_event` from the agent thread with the same context. `GatewayStreamConsumer` exposes `chat_id`. So the ContextVar owner does reach both status paths in production, not just in the tests.

The feature is off by default, and failures stay out of the reply path because every hook enqueues instead of awaiting. Exact-head CI and wn-agent Binaries are green, and the branch merges cleanly with #1701, which has already landed.

Non-blocking notes are inline. The one with behavioral impact: the superseded arrow goes on `state.target` rather than `state.anchor`. The rest are small simplifications. None of them block merge.

Source/API review at the exact head. No local tests were run; I relied on the existing CI.

liamhelmer-bel:
@Datawav @erskingardner: validation follow-through at unchanged signed head `728dca65`: all 365 local Hermes tests are now accounted for (362 passed, 3 skipped). The earlier nine media failures were resolved by giving the 22 media tests an isolated writable Marmot staging default; the other 343 tests retain their normal configuration so socket-resolution expectations remain intact. No production code or assertions changed. Three pinned-host contracts and both installed-host authorization dogfood contracts pass; both exact-head CI workflows are green. Tested and pr-open evidence are now recorded. Thanks for the exact-head approval and confirmation that the authorization blocker is closed; the new inline suggestions are noted as non-blocking. @Datawav, please re-review the startup authorization fix and real-host lifecycle regression when available.

liamhelmer-bel:
Friendly ping @maintainers — the current head (728dca657) has a fresh approval from a re-review (recommended action: merge); the earlier changes-requested state predates the authorization fix commit. Anything blocking on your side? Happy to address anything further.

Issue #1703, opened 2026-09-06
Resolve kind 10050 on advertised NIP-65 outboxes before declaring inbox missing


Datawav:
## Summary

MDK can declare a kind `10050` inbox list missing even when the account has a current valid list on the write relay advertised by its NIP-65 kind `10002` record. Existing-account setup queries both kinds on its caller-supplied bootstrap set, or the configured directory relays and then runtime relay endpoints when none is supplied; member-invite preflight uses the same configured-directory/runtime fallback. Neither path performs the target-aware second hop from a discovered NIP-65 outbox to that account's kind `10050` metadata.

For an account whose metadata is intentionally split across a discovery relay (`D`), advertised write/outbox relay (`W`), and inbox relay (`I`), the current result is incorrect in two independently visible paths: login can treat the inbox as absent and publish caller defaults, while invite preflight can reject a routeable recipient with `MissingMemberInboxRoute`. The shared directory resolver should own this lookup contract so account setup, member resolution, and future callers do not each invent different fallback logic.

## Source-confirmed behavior

Grounded against `marmot-protocol/mdk` `master` at [`8395b97bbd75c158be8d2ae293c56c5c1211d443`](https://github.com/marmot-protocol/mdk/commit/8395b97bbd75c158be8d2ae293c56c5c1211d443).

- [`fetch_account_relay_list_status_for_account_id`](https://github.com/marmot-protocol/mdk/blob/8395b97bbd75c158be8d2ae293c56c5c1211d443/crates/marmot-app/src/directory/methods.rs#L83-L100) fetches both relay-list kinds only from the endpoints selected by `directory_source_relays`: caller bootstrap endpoints, otherwise configured directory relays, otherwise runtime relay endpoints. Its current-only variant repeats the same endpoint selection and returns `None` when the requested kind was not observed there ([`methods.rs`](https://github.com/marmot-protocol/mdk/blob/8395b97bbd75c158be8d2ae293c56c5c1211d443/crates/marmot-app/src/directory/methods.rs#L145-L192)). Neither path follows the target account's discovered NIP-65 write relays before concluding that kind `10050` is absent.
[...]

PR #937, opened 2026-07-20
Recover relay publishing after stale socket failure

mubarakcoded:
Closes #926

## Summary
- reset only relay connections whose publish fails or exceeds the overall deadline
- serialize publish/reset work per relay so a failed request cannot tear down a concurrent healthy publish
- preserve SDK relay entries and long-lived subscriptions while terminating stale sockets
- wait for replacement connections before sending, including the SDK post-termination handoff
- cover fail-then-reconnect success, best-effort failures, partial multi-relay failures, and durable relay promotion

## Verification
- `cargo test -p transport-nostr-adapter --features sdk` (45 unit + 26 integration tests)
- `just fast-ci`
- `cargo fmt --check --all`
- `git diff --check`

---

<a href="https://stagereview.app/marmot-protocol/mdk/pull/937">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

Issue #854, opened 2026-07-14
storage-sqlite: save_message_draft does a BEGIN IMMEDIATE read-modify-write with no retry_on_busy, breaking the crate-wide busy-retry discipline


erskingardner:
### Summary

`save_message_draft` opens a `BEGIN IMMEDIATE` write transaction but is the **only** write path in the crate not wrapped in `retry_on_busy`. Under cross-connection/cross-process contention it surfaces `SQLITE_BUSY` ("database is locked") to the user on a composer autosave instead of transparently retrying, unlike every other write path.

### Location

`crates/storage-sqlite/src/message_drafts.rs:151-195` — specifically the raw `conn.transaction_with_behavior(TransactionBehavior::Immediate)` at `:162`.

### Details

`grep` confirms this is the sole `TransactionBehavior::Immediate` site in the crate, and every other real write path wraps its transaction in `retry_on_busy(...)` precisely because `BEGIN IMMEDIATE` eagerly grabs the write lock and is the dominant `SQLITE_BUSY` surface:

- `openmls_storage/value_store.rs:33,99,139,237,277`
- `storage/messages.rs:18,69` (`put_message`)
- `storage/outbound.rs:17` (queued outbound)
- `connection.rs:270`

`save_message_draft` uses `BEGIN IMMEDIATE` too but relies solely on `busy_timeout`, with no second-layer `retry_on_busy`. The crate's own test `queued_send_contention_is_retried_not_surfaced` (`storage/outbound.rs`) documents that cross-connection contention on one account DB is a real, handled condition — this path is the one that doesn't handle it.

### Failure scenario

A composer autosave (`MarmotApp::save_message_draft` → `drafts.rs:241` → this method) races an in-flight write on a sibling connection to the same account DB (e.g. the `wnd` runtime writing while another path saves a draft). Because `BEGIN IMMEDIATE` takes the write lock up front, the loser gets `SQLITE_BUSY` and the error is surfaced to the user rather than transparently retried, even though the whole body is idempotent (SQLite rolls back on BUSY, so re-running from BEGIN is safe).

### Why not a duplicate

#777 names `apply_group_push_token` / `apply_group_push_token_tombstone`; #768 is about the inverse (retry *inside* an engine transaction). This is a distinct, user-facing composer path (migration 0026) that neither covers.

### Suggested severity

[...]

PR #1705, opened 2026-09-06
Preserve outbox inbox metadata during login and invite discovery

Datawav:
Existing-account login and invite preparation could miss an inbox when discovery relays advertised a NIP-65 write relay but the kind `10050` metadata lived on that outbox. This change follows the advertised safe outboxes and preserves the newest target-authored relay lists across both hops and cached state.

## Resolution and preservation

- Both member metadata hops share bounded endpoint-set batching and single-author fallback, including real `CLOSED` multi-author rejection. Partial positive records survive fallback and route-cap failures; first-hop uncertainty cannot be cleared by an empty second hop.
- Account setup distinguishes signed missing/empty routes (`MissingRelayLists`) from unknown lookup (`RelayDirectory`). Timeouts, incomplete reads, saturated filters and future-rejected records cannot authorize default publication. Existing usable positive metadata remains usable during partial discovery.
- Prewarm stays process-local without adding strangers to durable directory subscriptions. Reuse preserves its original five-minute expiry. Commit-mode persistence and other relay/KeyPackage cache writers cannot overwrite newer relay metadata with older records.
- Completion-aware reads coalesce and remain safe when callers cancel. Inflight/failure telemetry includes these reads. Request-scoped SDK clients prevent discovered relays from accumulating in the live subscription client.

## Deliberate behavior changes

An incomplete lookup with no usable inbox now stays retryable instead of silently publishing defaults. A signed empty or read-only declaration is preserved and returns an actionable missing-route error; this PR does not silently convert read relays into write relays. A healthy-empty relay does not prove that an unavailable sibling has no inbox declaration. Setup gets a dedicated 26-second directory preflight budget, and member resolution retains a bounded 50-second overall deadline.

Replaceable records with equal timestamps now use NIP-01's lexically lower event ID when comparing fetched records. An already accepted cached value is retained on an equal timestamp because that projection does not retain the event ID.

Clients should explain broken setup and offer a simple, explicitly explained, non-destructive repair rather than overwrite existing lists without notice. Typed durable readiness/resume is owned by [MDK issue #1679](https://github.com/marmot-protocol/mdk/issues/1679); client warning/repair UI is a separate follow-up, not implemented here.

Client recovery follow-ups: [iOS #937](https://github.com/marmot-protocol/whitenoise-ios/issues/937) and [macOS #854](https://github.com/marmot-protocol/whitenoise-mac/issues/854).

## Verification

Focused regressions cover both signing login modes, route caps with and without usable metadata, signed empty/read-only preservation with zero replacement publications, cache expiry, future-dated metadata, strict filter saturation, coalescing/cancellation, production `CLOSED` fallback, and actual invites between independent runtimes that remain pending across restart. `just fast-ci` passed, as did both additional healthy-empty W1 / unavailable W2 login regressions. The latest follow-up also passed five account regressions and the failed-import runtime unit test. [Exact-head GitHub CI](https://github.com/marmot-protocol/mdk/actions/runs/34034206362) passed, including all four Rust shards, relay-runtime, CLI, bindings, Clippy and simulator checks. [All platform binary builds](https://github.com/marmot-protocol/mdk/actions/runs/34034206290) passed too.

The combined focused run passed 50 tests before an existing startup member-read timing assertion failed under concurrent testing (15 tests were not run after fail-fast). A serial run then passed all 20 selected relay-runtime tests, including that exact assertion and the remaining outbox acceptance cases. No assertion was weakened. Exact-head CI remains the full-suite gate.

The previous CI revision caught an outdated failed-import unit expectation: it still expected confirmed missing lists for a failed lookup. The follow-up requires the intended retryable `RelayDirectory` error and also covers signed-empty metadata overlapping an incomplete outbox query. Observed metadata remains durable and neither error permits replacement publication.

Addresses [MDK issue #1703](https://github.com/marmot-protocol/mdk/issues/1703). [White Noise Android issue #2452](https://github.com/marmot-protocol/whitenoise-android/issues/2452) tracks related invite failure presentation and downstream verification. No shipped-app fix, release or merge is claimed by this PR.

### DIFF
--- a/crates/marmot-app/src/directory/member_key_packages.rs
+++ b/crates/marmot-app/src/directory/member_key_packages.rs
@@ -663,6 +663,69 @@ impl MarmotApp {
                 }
             }
         }
+
+        // NIP-65 write relays are the target account's outboxes. Reconcile a
+        // current kind-10050 from those outboxes even when the discovery relay
+        // already returned an older inbox list. Keep this set operation bounded
+        // and preserve positive first-hop metadata if an outbox is unavailable;
+        // only an unconfirmed empty result becomes a typed relay error.
+        let second_hop_specs = needs_discovery
+            .into_iter()
+            .filter_map(|index| {
+                let endpoints = self.retain_safe_discovered_endpoints(
+                    targets[index]
+                        .relay_lists
+                        .nip65
+                        .relays
+                        .iter()
+                        .cloned()
+                        .map(TransportEndpoint)
+                        .collect(),
+                    "member inbox outbox discovery",
+                );
+                (!endpoints.is_empty()).then_some((
+                    index,
+                    targets[index].account_id_hex.clone(),
+                    endpoints,
+                ))
+            })
+            .collect::<Vec<_>>();
+        let app = self.clone();
+        let work = second_hop_specs
+            .into_iter()
+            .map(move |(index, account_id, endpoints)| {
+                let app = app.clone();
+                async move {
+                    let result = app
+                        .fetch_account_relay_list_status_for_account_id(&account_id, endpoints)
+                        .await;
+                    (index, result)
+                }
+            });
+        let second_hop_results = stream::iter(work)
+            .buffered(MEMBER_RESOLUTION_RELAY_CONCURRENCY)
+            .collect::<Vec<_>>()
+            .await;
+        for (index, result) in second_hop_results {
+            match result {
+                Ok(status) => {
+                    targets[index].relay_lists =
+                        merge_member_relay_lists(targets[index].relay_lists.clone(), status);
+                    failures.retain(|(failed_index, _)| *failed_index != index);
+                }
+                Err(error) if targets[index].relay_lists.inbox.relays.is_empty() => {
+                    if !failures
+                        .iter()
+                        .any(|(failed_index, _)| *failed_index == index)
+                    {
+                        failures.push((index, error));
+                    }
+                }
+                Err(_) => {
+                    failures.retain(|(failed_index, _)| *failed_index != index);
+                }
+            }
+        }
         failures
     }
 
--- a/crates/marmot-app/src/directory/member_key_packages.rs
+++ b/crates/marmot-app/src/directory/member_key_packages.rs
@@ -22,10 +22,10 @@ use transport_nostr_adapter::{
 
 use crate::key_package_records::{
     fresh_or_cached_key_package, fresh_relay_list_status_from_records,
-    latest_fresh_key_package_from_records, validated_cached_key_package,
+    latest_fresh_key_package_from_records, merge_relay_list_status, validated_cached_key_package,
 };
 use crate::relay_plane::DirectoryEventQuery;
-use crate::{AccountRelayListStatus, AppError, FetchedKeyPackage, MarmotApp, push_unique_strings};
+use crate::{AccountRelayListStatus, AppError, FetchedKeyPackage, MarmotApp};
 
 /// Maximum authors placed in one Nostr directory filter.
 pub(crate) const MEMBER_RESOLUTION_AUTHORS_PER_QUERY: usize = 32;
@@ -175,38 +175,8 @@ struct MemberTarget {
     account_id_hex: String,
     local_label: Option<String>,
     relay_lists: AccountRelayListStatus,
-}
-
-/// Merge relay-list observations without allowing stale discovery to clobber
-/// newer cached state.
-///
-/// Each list is replaced only by a strictly newer `created_at`; ties retain the
-/// current value. Timestamp-less (`created_at == 0`) state may fill an empty
-/// timestamp-less slot, which supports setup-derived and fixture-built state
-/// without overriding a timestamped directory event.
-fn merge_member_relay_lists(
-    mut current: AccountRelayListStatus,
-    candidate: AccountRelayListStatus,
-) -> AccountRelayListStatus {
-    if candidate.nip65.created_at > current.nip65.created_at
-        || (candidate.nip65.created_at == 0
-            && current.nip65.created_at == 0
-            && current.nip65.relays.is_empty()
-            && !candidate.nip65.relays.is_empty())
-    {
-        current.nip65 = candidate.nip65;
-    }
-    if candidate.inbox.created_at > current.inbox.created_at
-        || (candidate.inbox.created_at == 0
-            && current.inbox.created_at == 0
-            && current.inbox.relays.is_empty()
-            && !candidate.inbox.relays.is_empty())
-    {
-        current.inbox = candidate.inbox;
-    }
-    push_unique_strings(&mut current.bootstrap_relays, candidate.bootstrap_relays);
-    current.refresh();
-    current
+    cached_relay_lists: AccountRelayListStatus,
+    relay_records: Vec<crate::relay_plane::DirectoryRelayEventRecord>,
 }
 
 #[allow(dead_code)]
@@ -313,7 +283,9 @@ impl MarmotApp {
             targets.push(MemberTarget {
                 account_id_hex,
                 local_label: local.map(|account| account.label),
-                relay_lists,
+                relay_lists: relay_lists.clone(),
+                cached_relay_lists: relay_lists,
+                relay_records: Vec::new(),
             });
         }
 
@@ -321,6 +293,7 @@ impl MarmotApp {
             .map(|_| None)
             .collect::<Vec<Option<Result<KeyPackage, AppError>>>>();
         let mut unresolved = Vec::new();
+        let mut fresh_prewarmed_routes = HashSet::new();
         let mut reused_members = 0usize;
         for (index, target) in targets.iter_mut().enumerate() {
             if let Some(label) = &target.local_label
@@ -349,12 +322,30 @@ impl MarmotApp {
                 .unwrap_or_else(|poisoned| poisoned.into_inner())
                 .get(&target.account_id_hex);
             if let Some(mut fetched) = prefetched {
-                target.relay_lists = merge_member_relay_lists(
+                target.relay_lists = merge_relay_list_status(
                     target.relay_lists.clone(),
                     fetched.relay_lists.clone(),
                 );
+                target.cached_relay_lists = target.relay_lists.clone();
                 fetched.relay_lists = target.relay_lists.clone();
                 if let Ok(key_package) = self.accept_prefetched_key_package(purpose, fetched) {
+                    if !target.relay_lists.nip65.relays.is_empty()
+                        && !self
+                            .retain_safe_discovered_endpoints(
+                                target
+                                    .relay_lists
+                                    .inbox
+                                    .relays
+                                    .iter()
+                                    .cloned()
+                                    .map(TransportEndpoint)
+                                    .collect(),
+                                "member prewarm inbox readiness",
+                            )
+                            .is_empty()
+                    {
+                        fresh_prewarmed_routes.insert(index);
+                    }
                     outcomes[index] = Some(Ok(key_package));
                     reused_members += 1;
                     continue;
@@ -363,25 +354,11 @@ impl MarmotApp {
             unresolved.push(index);
         }
 
-        let relay_list_unresolved = targets
-            .iter()
-            .enumerate()
-            .filter_map(|(index, target)| {
-                let safe_inbox = self.retain_safe_discovered_endpoints(
-                    target
-                        .relay_lists
-                        .inbox
-                        .relays
-                        .iter()
-                        .cloned()
-                        .map(TransportEndpoint)
-                        .collect(),
-                    "member invite inbox discovery",
-                );
-                (safe_inbox.is_empty()
-                    || (outcomes[index].is_none() && target.relay_lists.nip65.relays.is_empty()))
-                .then_some(index)
-            })
+        // Durable projections have unknown observation age and need a bounded
+        // refresh. Process-local prewarm entries have an enforced TTL and were
+        // already resolved through the outbox path during composition.
+        let relay_list_unresolved = (0..targets.len())
+            .filter(|index| !fresh_prewarmed_routes.contains(index))
             .collect::<Vec<_>>();
         if !relay_list_unresolved.is_empty() {
             for (index, error) in self
@@ -534,7 +511,7 @@ impl MarmotApp {
                 .map(|endpoint| endpoint.0.clone())
                 .collect();
         }
-        merge_member_relay_lists(target.relay_lists.clone(), status)
+        merge_relay_list_status(target.cached_relay_lists.clone(), status)
     }
 
     async fn resolve_missing_relay_lists(
@@ -578,14 +555,14 @@ impl MarmotApp {
             .map(move |(indices, endpoints, queries)| {
                 let app = app.clone();
                 async move {
-                    let records = app
+                    let outcome = app
                         .relay_plane
-                        .fetch_directory_events(endpoints.clone(), queries)
+                        .fetch_directory_events_with_completion(endpoints.clone(), queries)
                         .await
                         .map_err(|error| {
                             AppError::RelayDirectory(format!("fetch relay lists: {error}"))
                         });
-                    (indices, endpoints, records)
+                    (indices, endpoints, outcome)
                 }
             });
         let batches = stream::iter(requests)
@@ -594,21 +571,37 @@ impl MarmotApp {
             .await;
 
         let mut failures = Vec::new();
+        let mut completed_discovery = HashSet::new();
         for (indices, endpoints, result) in batches {
             match result {
-                Ok(records) => {
+                Ok(outcome) => {
                     for index in indices {
+                        if outcome.complete {
+                            completed_discovery.insert(index);
+                        }
                         let account_id = &targets[index].account_id_hex;
-                        let account_records = records
+                        let account_records = outcome
+                            .records
                             .iter()
                             .filter(|record| record.event.pubkey == *account_id)
                             .cloned()
                             .collect::<Vec<_>>();
+                        targets[index].relay_records.extend(account_records);
+                        let relay_records = targets[index].relay_records.clone();
                         targets[index].relay_lists = self.relay_lists_from_records(
                             &targets[index],
-                            account_records,
+                            relay_records,
                             &endpoints,
                         );
+                        if !outcome.complete && targets[index].relay_lists.inbox.relays.is_empty() {
+                            failures.push((
+                                index,
+                                AppError::RelayDirectory(
+                                    "relay-list absence was not authoritatively established"
+                                        .to_owned(),
+                                ),
+                            ));
+                        }
                     }
                 }
                 Err(_) => {
@@ -634,28 +627,44 @@ impl MarmotApp {
                                     )
                                 })
                                 .collect::<Vec<_>>();
-                            let records = app
+                            let outcome = app
                                 .relay_plane
-                                .fetch_directory_events(endpoints.clone(), queries)
+                                .fetch_directory_events_with_completion(endpoints.clone(), queries)
                                 .await
                                 .map_err(|error| {
                                     AppError::RelayDirectory(format!("fetch relay lists: {error}"))
                                 });
-                            (index, endpoints, records)
+                            (index, endpoints, outcome)
                         }
                     });
                     let results = stream::iter(work)
                         .buffered(MEMBER_RESOLUTION_FALLBACK_CONCURRENCY)
                         .collect::<Vec<_>>()
                         .await;
-                    for (index, endpoints, records) in results {
-                        match records {
-                            Ok(records) => {
+                    for (index, endpoints, outcome) in results {
+                        match outcome {
+                            Ok(outcome) => {
+                                if outcome.complete {
+                                    completed_discovery.insert(index);
+                                }
+                                targets[index].relay_records.extend(outcome.records);
+                                let relay_records = targets[index].relay_records.clone();
                                 targets[index].relay_lists = self.relay_lists_from_records(
                                     &targets[index],
-                                    records,
+                                    relay_records,
                                     &endpoints,
                                 );
+                                if !outcome.complete
+                                    && targets[index].relay_lists.inbox.relays.is_empty()
+                                {
+                                    failures.push((
+                                        index,
+                                        AppError::RelayDirectory(
+                                            "relay-list absence was not authoritatively established"
+                                                .to_owned(),
+                                        ),
+                                    ));
+                                }
                             }
                             Err(error) => failures.push((index, error)),
                         }
@@ -664,65 +673,169 @@ impl MarmotApp {
             }
         }
 
-        // NIP-65 write relays are the target account's outboxes. Reconcile a
-        // current kind-10050 from those outboxes even when the discovery relay
-        // already returned an older inbox list. Keep this set operation bounded
-        // and preserve positive first-hop metadata if an outbox is unavailable;
-        // only an unconfirmed empty result becomes a typed relay error.
-        let second_hop_specs = needs_discovery
-            .into_iter()
-            .filter_map(|index| {
-                let endpoints = self.retain_safe_discovered_endpoints(
-                    targets[index]
-                        .relay_lists
-                        .nip65
-                        .relays
-                        .iter()
-                        .cloned()
-                        .map(TransportEndpoint)
-                        .collect(),
-                    "member inbox outbox discovery",
-                );
-                (!endpoints.is_empty()).then_some((
-                    index,
-                    targets[index].account_id_hex.clone(),
-                    endpoints,
-                ))
-            })
-            .collect::<Vec<_>>();
+        // Group targets by the same safe outbox set so each relay receives one
+        // bounded multi-author request. This keeps prewarm pure: persistence
+        // happens only after the whole member-resolution pass succeeds.
+        let mut by_outboxes = BTreeMap::<Vec<TransportEndpoint>, Vec<usize>>::new();
+        for index in needs_discovery {
+            let outbox_endpoints = self.retain_safe_discovered_endpoints(
+                targets[index]
+                    .relay_lists
+                    .nip65
+                    .relays
+                    .iter()
+                    .cloned()
+                    .map(TransportEndpoint)
+                    .collect(),
+                "member inbox outbox discovery",
+            );
+            // A completed first-hop query already covered both relay-list kinds
+            // on these endpoints. Only genuinely new outboxes need another hop;
+            // incomplete reads remain eligible for the bounded retry.
+            let outbox_endpoints = outbox_endpoints
+                .into_iter()
+                .filter(|endpoint| {
+                    !completed_discovery.contains(&index) || !endpoints.contains(endpoint)
+                })
+                .collect::<Vec<_>>();
+            if !outbox_endpoints.is_empty() {
+                by_outboxes.entry(outbox_endpoints).or_default().push(index);
+            }
+        }
+        let mut second_hop_specs = Vec::new();
+        for (endpoints, indices) in by_outboxes {
+            for chunk in indices.chunks(MEMBER_RESOLUTION_AUTHORS_PER_QUERY) {
+                let indices = chunk.to_vec();
+                let authors = indices
+                    .iter()
+                    .map(|index| targets[*index].account_id_hex.clone())
+                    .collect::<Vec<_>>();
+                let queries = [KIND_NIP65_RELAY_LIST, KIND_MARMOT_INBOX_RELAY_LIST]
+                    .into_iter()
+                    .map(|kind| {
+                        DirectoryEventQuery::new(
+                            kind,
+                            authors.clone(),
+                            authors.len() * RELAY_LIST_EVENTS_PER_AUTHOR,
+                        )
+                    })
+                    .collect::<Vec<_>>();
+                second_hop_specs.push((indices, endpoints.clone(), queries));
+            }
+        }
         let app = self.clone();
+        let account_ids = targets
+            .iter()
+            .map(|target| target.account_id_hex.clone())
+            .collect::<Vec<_>>();
         let work = second_hop_specs
             .into_iter()
-            .map(move |(index, account_id, endpoints)| {
+            .map(move |(indices, endpoints, queries)| {
                 let app = app.clone();
+                let authors = indices
+                    .iter()
+                    .map(|index| account_ids[*index].clone())
+                    .collect::<Vec<_>>();
                 async move {
-                    let result = app
-                        .fetch_account_relay_list_status_for_account_id(&account_id, endpoints)
-                        .await;
-                    (index, result)
+                    let outcome = app
+                        .relay_plane
+                        .fetch_directory_events_with_completion(endpoints.clone(), queries.clone())
+                        .await
+                        .map_err(|error| {
+                            AppError::RelayDirectory(format!("fetch outbox relay lists: {error}"))
+                        });
+                    if outcome.is_ok() || indices.len() == 1 {
+                        return vec![(indices, endpoints, outcome)];
+                    }
+                    // Keep the same bounded single-author fallback on the
+                    // advertised outbox as on discovery relays. A rejected
+                    // multi-author query is not evidence of missing metadata.
+                    stream::iter(indices.into_iter().enumerate().map(|(offset, index)| {
+                        let app = app.clone();
+                        let endpoints = endpoints.clone();
+                        let queries = queries
+                            .iter()
+                            .map(|query| {
+                                DirectoryEventQuery::new(
+                                    query.kind,
+                                    vec![authors[offset].clone()],
+                                    RELAY_LIST_EVENTS_PER_AUTHOR,
+                                )
+                            })
+                            .collect();
+                        async move {
+                            let outcome = app
+                                .relay_plane
+                                .fetch_directory_events_with_completion(endpoints.clone(), queries)
+                                .await
+                                .map_err(|error| {
+                                    AppError::RelayDirectory(format!(
+                                        "fetch outbox relay lists: {error}"
+                                    ))
+                                });
+                            (vec![index], endpoints, outcome)
+                        }
+                    }))
+                    .buffered(MEMBER_RESOLUTION_FALLBACK_CONCURRENCY)
+                    .collect::<Vec<_>>()
+                    .await
                 }
             });
         let second_hop_results = stream::iter(work)
             .buffered(MEMBER_RESOLUTION_RELAY_CONCURRENCY)
             .collect::<Vec<_>>()
             .await;
-        for (index, result) in second_hop_results {
+        for (indices, endpoints, result) in second_hop_results.into_iter().flatten() {
             match result {
-                Ok(status) => {
-                    targets[index].relay_lists =
-                        merge_member_relay_lists(targets[index].relay_lists.clone(), status);
-                    failures.retain(|(failed_index, _)| *failed_index != index);
-                }
-                Err(error) if targets[index].relay_lists.inbox.relays.is_empty() => {
-                    if !failures
-                        .iter()
-                        .any(|(failed_index, _)| *failed_index == index)
-                    {
-                        failures.push((index, error));
+                Ok(outcome) => {
+                    for index in indices {
+                        let account_id = &targets[index].account_id_hex;
+                        let records = outcome
+                            .records
+                            .iter()
+                            .filter(|record| record.event.pubkey == *account_id)
+                            .cloned()
+                            .collect::<Vec<_>>();
+                        targets[index].relay_records.extend(records);
+                        let relay_records = targets[index].relay_records.clone();
+                        targets[index].relay_lists = self.relay_lists_from_records(
+                            &targets[index],
+                            relay_records,
+                            &endpoints,
+                        );
+                        if outcome.complete || !targets[index].relay_lists.inbox.relays.is_empty() {
+                            failures.retain(|(failed_index, _)| *failed_index != index);
+                        } else if !failures
+                            .iter()
+                            .any(|(failed_index, _)| *failed_index == index)
+                        {
+                            failures.push((
+                                index,
+                                AppError::RelayDirectory(
+                                    "relay-list absence was not authoritatively established"
+                                        .to_owned(),
+                                ),
+                            ));
+                        }
                     }
                 }
                 Err(_) => {
-                    failures.retain(|(failed_index, _)| *failed_index != index);
+                    for index in indices {
+                        if targets[index].relay_lists.inbox.relays.is_empty()
+                            && !failures
+                                .iter()
+                                .any(|(failed_index, _)| *failed_index == index)
+                        {
+                            failures.push((
+                                index,
+                                AppError::RelayDirectory(
+                                    "fetch outbox relay lists failed".to_owned(),
+                                ),
+                            ));
+                        } else if !targets[index].relay_lists.inbox.relays.is_empty() {
+                            failures.retain(|(failed_index, _)| *failed_index != index);
+                        }
+                    }
                 }
             }
         }
--- a/crates/marmot-app/src/directory/member_key_packages.rs
+++ b/crates/marmot-app/src/directory/member_key_packages.rs
@@ -24,7 +24,7 @@ use crate::key_package_records::{
     fresh_or_cached_key_package, fresh_relay_list_status_from_records,
     latest_fresh_key_package_from_records, merge_relay_list_status, validated_cached_key_package,
 };
-use crate::relay_plane::DirectoryEventQuery;
+use crate::relay_plane::{DirectoryEventQuery, DirectoryFetchOutcome};
 use crate::{AccountRelayListStatus, AppError, FetchedKeyPackage, MarmotApp};
 
 /// Maximum authors placed in one Nostr directory filter.
@@ -34,7 +34,10 @@ const MEMBER_RESOLUTION_RELAY_CONCURRENCY: usize = 4;
 /// Existing per-member fallback concurrency, retained for incompatible relays.
 const MEMBER_RESOLUTION_FALLBACK_CONCURRENCY: usize = 8;
 /// One complete set resolution is bounded even when several relay sets stall.
-const MEMBER_RESOLUTION_DEADLINE: Duration = Duration::from_secs(10);
+// Three network stages, each with a bounded single-author retry, can each
+// consume a 5s connection budget plus a 3s fetch budget. Keep enough time for
+// those stages and local validation while retaining one overall deadline.
+const MEMBER_RESOLUTION_DEADLINE: Duration = Duration::from_secs(50);
 const KEY_PACKAGE_EVENTS_PER_AUTHOR: usize = 12;
 const RELAY_LIST_EVENTS_PER_AUTHOR: usize = 4;
 const MEMBER_PREWARM_CACHE_LIMIT: usize = 256;
@@ -77,6 +80,34 @@ mod tests {
             Instant::now() - MEMBER_PREWARM_CACHE_TTL - Duration::from_secs(1);
         assert!(cache.get(&newest).is_none());
         assert!(!cache.order.contains(&newest));
+    }
+
+    #[tokio::test]
+    async fn composition_prewarm_reuse_preserves_original_freshness_deadline() {
+        let (_directory, app, accounts, _fetcher) =
+            crate::tests::member_resolution_fixture(1, false).await;
+        let id = &accounts[0].account_id_hex;
+        app.prewarm_group_member_key_packages(&[id.as_str()])
+            .await
+            .unwrap();
+        let original = Instant::now() - Duration::from_secs(240);
+        app.member_key_package_prewarm_cache
+            .lock()
+            .unwrap()
+            .entries
+            .get_mut(id)
+            .unwrap()
+            .inserted_at = original;
+        for _ in 0..3 {
+            app.prewarm_group_member_key_packages(&[id.as_str()])
+                .await
+                .unwrap();
+        }
+        let mut cache = app.member_key_package_prewarm_cache.lock().unwrap();
+        assert_eq!(cache.entries.get(id).unwrap().inserted_at, original);
+        cache.entries.get_mut(id).unwrap().inserted_at =
+            Instant::now() - MEMBER_PREWARM_CACHE_TTL - Duration::from_secs(1);
+        assert!(cache.get(id).is_none());
     }
 }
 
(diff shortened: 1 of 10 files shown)


Datawav:
Additional acceptance verification failed on commit `f3f9d87f5ab02b0cab803e8d4b0acac046a35697`: **0 passed, 6 failed** (9.56 seconds). The isolated checkout used this exact production code plus six new real-relay tests.

- **Incomplete discovery still republishes defaults, in both login modes.** A healthy empty bootstrap relay B plus a stalled explicit discovery relay D produced **two kind 10002/10050 publication attempts**, where the tests require zero. Preflight failure is flattened to `None`, and setup retries only B, dropping D from the scope used to authorize absence. Completion evidence must survive this retry.
- **Older outbox records replace newer inbox metadata, in both login modes.** Four cases cover newer kind 10050 on discovery versus an older nonempty or explicit-empty list on the advertised outbox. All returned the wrong inbox. The account resolver currently returns the second one-hop status without reconciling replaceable-event ordering across hops/cache.

The initial passing tests cover useful cases but do not resolve these failures. Please correct these paths and rerun the added regressions before treating this fix as complete. The fixtures and full logs are handed to the existing implementation owner; no production code was changed during verification.

Reproduction filter: `cargo test -p marmot-app --test relay_runtime outbox_acceptance_` with the supplied six-test fixture patch. Patch SHA-256: `59e2ceb66526e747e16b93bbaffab8edd8788f91040a453b85750479c492753c`. Log SHA-256: `4c0cc1aee2108596873f07d3832a5b7ff33956ab1f5e2ae967ccf4f481a2450c`.

This is supporting test evidence; independent review and Android verification remain pending.

Datawav:
The six regressions reported above now pass against the local corrective worktree: **6 passed, 0 failed**, in 7.53 seconds. Both local-key and external-signer login preserve a newer discovery inbox over older outbox records (including explicit-empty records), and neither publishes replacement relay lists after incomplete required discovery.

These corrections have not yet replaced the PR head `f3f9d87f5ab02b0cab803e8d4b0acac046a35697`. The separate-sender invitation/restart cases, broader checks, independent review, and Android adoption/device verification remain pending. This is a local regression result, not PR readiness or proof that the shipped app is fixed.

crates/marmot-app/src/directory/methods.rs
```diff
@@ -142,6 +142,46 @@ impl MarmotApp {
...
+            "account inbox outbox discovery",
+        );
+        if outbox_relays.is_empty() {
+            return Ok(first_hop);
+        }
+        match self
+            .fetch_account_relay_list_status_for_account_id(account_id_hex, outbox_relays)
+            .await
+        {
+            Ok(status) => Ok(status),
```
erskingardner: [P1] Carry partial-fetch completion into the second-hop result

This `Ok(status)` does not establish that the advertised outboxes completed discovery. The existing `NostrSdkDirectoryRelayFetcher` drops relays that fail to connect (directory.rs:451-459) and returns records from the surviving relays without retaining that failure in the result. For example, D advertises W1 and W2, W1 successfully returns no inbox, and W2 (which holds the valid kind 10050) fails to connect. This branch returns an incomplete but successful status, so setup can publish caller defaults instead of reporting unknown/retryable discovery. This remains broken even after the retry-scope finding is fixed. Preserve per-hop completion/failure evidence and refuse to infer absence from a partially successful outbox fetch; add the healthy-empty W1 / failed W2 case for both signing login paths.

crates/marmot-app/src/directory/member_key_packages.rs
```diff
@@ -663,6 +663,69 @@ impl MarmotApp {
...
+            })
+            .collect::<Vec<_>>();
+        let app = self.clone();
+        let work = second_hop_specs
+            .into_iter()
+            .map(move |(index, account_id, endpoints)| {
+                let app = app.clone();
+                async move {
+                    let result = app
+                        .fetch_account_relay_list_status_for_account_id(&account_id, endpoints)
```
erskingardner: [P2] Keep the second hop free of persistent writes during prewarm

`resolve_missing_relay_lists` runs for both Commit and Prewarm, but this helper unconditionally calls `remember_directory_relay_lists` (methods.rs:135-140). That creates a persistent directory entry for a previously unknown composition candidate, writes it to shared/per-account caches, and requests a directory-sync rebuild. `directory_sync_plan` then includes that account in live subscriptions even if the user cancels composition. The surrounding resolver deliberately keeps Prewarm data in `member_key_package_prewarm_cache` and only admits directory state for Commit. Fetch/merge second-hop records without admitting them here, then persist under the existing purpose-specific owner. Cover a remote candidate with no prior directory entry and assert that prewarm leaves both persistent directory membership and the live sync plan unchanged.

crates/marmot-app/src/directory/member_key_packages.rs
```diff
@@ -663,6 +663,69 @@ impl MarmotApp {
...
+                    index,
+                    targets[index].account_id_hex.clone(),
+                    endpoints,
+                ))
+            })
+            .collect::<Vec<_>>();
+        let app = self.clone();
+        let work = second_hop_specs
+            .into_iter()
+            .map(move |(index, account_id, endpoints)| {
```
erskingardner: [P2] Batch second-hop authors sharing the same outbox endpoints

This creates one fetch per target, even when all targets advertise the same relay set. The directory plane coalesces only identical requests, so different single-author queries cannot combine; bounded concurrency still leaves the roster going through successive per-author waves under the shared resolution deadline. Exact-head CI exposes the amplification: `member_key_package_set_batches_shared_relay_and_reuses_prewarm` makes 10 requests for 8 members (previously 2), and the fallback test makes 6 single-author requests instead of 4. Group targets by safe outbox endpoint set and chunk authors as the existing first-hop/KeyPackage paths do, preserving single-author fallback. Update the tests to budget the necessary additional batched hop and verify that its request count scales by batches, rather than merely raising the assertions to accept one extra request per member.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T10:08:24Z` |
| Commit reviewed | `f3f9d87f5ab02b0cab803e8d4b0acac046a35697` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

The second-hop lookup is needed and the happy-path tests are useful, but this head is not ready to merge. I reviewed all five changed files and the surrounding account setup, directory persistence, and fetch paths through GitHub at this SHA.

I independently confirm the two existing unresolved findings, without duplicating their inline threads:

- **[P1] Older second-hop metadata wins over newer first-hop/cache metadata.** Returning the second fetch directly makes its observed older inbox authoritative to setup; an older explicit-empty list can trigger default republication. Reconcile each relay-list kind across both hops and the cache before returning. [Existing thread](https://github.com/marmot-protocol/mdk/pull/1705#discussion_r3943528202).
- **[P1] Failed preflight loses its discovery scope on retry.** Both signing login paths flatten failed/timed-out preflight to `None`, then retry only bootstrap relays. An empty bootstrap response can authorize default publication while an explicit discovery relay remains unqueried/incomplete. Preserve the original discovery scope and completion outcome. [Existing thread](https://github.com/marmot-protocol/mdk/pull/1705#discussion_r3943528205).

Three additional findings are attached inline: partial second-hop failure is still accepted as absence, member prewarm now admits composition candidates into persistent directory subscriptions, and the second hop loses shared-relay batching.

Validation: [CI for this exact head](https://github.com/marmot-protocol/mdk/actions/runs/34023824769) has failed Required CI. Rust shards 2/4 and 4/4 fail the existing member-resolution fallback/batching tests on both attempts: 6 versus 4 single-author requests, and 10 versus 2 total requests. The relay-runtime job passes, but its new fixtures do not cover these failure paths; the Welcome/restart fixture also shares the recipient and sender runtime/cache, so it does not independently establish cold remote-sender discovery. The author's separate six failing acceptance regressions are supporting reported evidence, not tests I executed. No local checkout or files were changed, and no new local tests were run.

crates/marmot-app/src/directory/member_key_packages.rs
```diff
@@ -634,35 +627,218 @@ impl MarmotApp {
...
+                async move {
+                    let outcome = app
+                        .relay_plane
+                        .fetch_directory_events_with_completion(endpoints.clone(), queries.clone())
+                        .await
+                        .map_err(|error| {
+                            AppError::RelayDirectory(format!("fetch outbox relay lists: {error}"))
+                        });
+                    if outcome.is_ok() || indices.len() == 1 {
+                        return vec![(indices, endpoints, outcome)];
```
erskingardner: [P2] Run single-author fallback for rejected incomplete batches

The production completion fetch returns `Ok(DirectoryFetchOutcome { complete: false, ... })` when a relay sends `CLOSED` or subscription setup fails (`relay_plane/directory.rs:536-564, 627-656`). Consequently `outcome.is_ok()` is true for a rejected multi-author request, and this branch skips the advertised single-author fallback. The discovery-hop match has the same problem: its fallback is only in `Err(_)` at line 607. A relay that rejects multi-author filters but serves single-author queries can therefore strand an otherwise routable multi-member invite. The passing fallback unit test masks this because its scripted fetcher returns `Err("multi-author queries unsupported")` instead of the real incomplete outcome. Preserve rejection information or allow incomplete failed batches into the bounded per-author fallback on both hops, retaining any positive records. Add a real-relay CLOSED-on-multi-author regression that proves the single-author requests are issued and the roster resolves.

crates/marmot-app/src/directory/member_key_packages.rs
```diff
@@ -634,35 +627,218 @@ impl MarmotApp {
...
+                            .collect::<Vec<_>>();
+                        targets[index].relay_records.extend(records);
+                        let relay_records = targets[index].relay_records.clone();
+                        targets[index].relay_lists = self.relay_lists_from_records(
+                            &targets[index],
+                            relay_records,
+                            &endpoints,
+                        );
+                        if outcome.complete || !targets[index].relay_lists.inbox.relays.is_empty() {
+                            failures.retain(|(failed_index, _)| *failed_index != index);
```
erskingardner: [P2] Preserve failed discovery when the outbox supplies no inbox evidence

`outcome.complete` describes only this second-hop request, but it unconditionally removes the failure recorded for the first hop. For example, the cache contains an old NIP-65 outbox W and a valid KeyPackage but no inbox; discovery D disconnects, while W completes an empty read. D may hold a newer NIP-65 record pointing to the actual inbox-metadata host, yet this branch erases the unknown-discovery outcome and final preflight returns `MissingMemberInboxRoute`. That violates the required distinction between a failed lookup and confirmed absence. Track first- and second-hop completion independently; an empty completed second hop must not clear an incomplete first hop. A resolved usable inbox can still satisfy readiness. Add a test for incomplete D plus successful-empty cached W and require the retryable/unknown directory error rather than the missing-route result.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T10:58:27Z` |
| Commit reviewed | `669e67fa98ced53dad311db2d6219536e616701e` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

The corrective revision addresses the earlier findings in source: cross-hop/cache merging preserves newer records; login retry keeps the original discovery scope; account setup checks completion before accepting missing lists; member prewarm uses record fetching without persistent admission; and shared outboxes are batched. The separate sender/receiver runtime tests now exercise the previously missing cold-sender boundary and persisted pending invitations.

Two member-resolution defects remain, attached inline:

1. Real relay rejection is returned as `Ok(complete: false)`, bypassing the single-author fallback that still runs only for `Err`. The scripted test returns `Err` and therefore does not exercise the production behavior.
2. A successful but empty outbox query clears an earlier discovery failure, allowing unknown metadata to become `MissingMemberInboxRoute`.

I also confirm the [existing outbox route-cap error finding](https://github.com/marmot-protocol/mdk/pull/1705#discussion_r3943725279): safety filtering does not impose the route-count cap, so the second-hop `?` can discard usable first-hop metadata when the subsequent sanitizer rejects the set. Handle that as incomplete discovery while retaining observed records, without weakening the dial cap.

The [persistence-before-gate comment](https://github.com/marmot-protocol/mdk/pull/1705#discussion_r3943725282) needs a narrower correction than moving the entire write after the gate: signed explicit-empty records and valid positive observations from partial reads should remain durable. Returning a readiness error does not itself make those observed records invalid. I am not treating persistence of those records as a separate blocker.

Validation: reviewed the eight-file diff and surrounding source through GitHub at this SHA. [Current CI](https://github.com/marmot-protocol/mdk/actions/runs/34028325684) passes all four Rust test shards, relay-runtime tests, format, check, Clippy, and CLI relay tests; the previous batching failures are no longer red. Failing checks at posting: none. Still pending at posting: Build wn-agent and terminal harnesses (darwin-x86_64). No new local tests were run and no local repository state was changed. These remaining findings are source-confirmed, not new executable reproductions.

crates/marmot-app/src/directory/methods.rs
```diff
@@ -142,6 +143,96 @@ impl MarmotApp {
...
+        }
+        {
+            let app = self.clone();
+            let account_id = account_id_hex.clone();
+            let remembered = status.clone();
+            blocking_app_task(move || app.remember_directory_relay_lists(&account_id, &remembered))
+                .await?;
+        }
+
+        let explicit_empty = (status.nip65.created_at > 0 && status.nip65.relays.is_empty())
```
erskingardner, line 215: **Blocking: a NIP-65 list with no `write` relays is misread as an intentional empty list and hard-fails login.**

`relay_list_state_from_event` (`crates/marmot-app/src/lib.rs:6357-6363`) sets `nip65.relays = write_relays` only, and `parse_nip65_relay_set` puts a relay in `write_relays` only when the marker is absent or `write`. So an account whose kind `10002` marks every entry `read` (or uses a marker this parser skips) produces `nip65.relays == []` with `created_at > 0`.

That lands in `explicit_empty` here, and login returns `RelayDirectory("relay-list absence was not authoritatively established")` — which is both wrong (absence *was* established; the list simply has no write relays) and sticky, because `remember_directory_relay_lists` above persisted the empty status, so the next login re-derives `explicit_empty` from cache alone.

Failure scenario: user's kind `10002` is `[["r","wss://a","read"],["r","wss://b","read"]]`; `create_or_import_account` / `login_external_signer` now fail permanently, where before they succeeded.

Suggest deciding "explicitly empty" from the *event's* tag set rather than from the derived write-relay projection, and separating the NIP-65 case from the inbox case.

crates/marmot-app/src/directory/methods.rs
```diff
@@ -142,6 +143,96 @@ impl MarmotApp {
...
+            || (status.inbox.created_at > 0 && status.inbox.relays.is_empty());
+        let uncertain_absence =
```
erskingardner, line 217: **Blocking: one unreachable advertised outbox turns the whole read "incomplete" and blocks first-time Marmot login.**

`strict_fetch_endpoint` returns `DirectoryFetchOutcome::default()` (`complete: false`) whenever a relay cannot be added, cannot be connected inside `DIRECTORY_RELAY_CONNECT_WAIT`, or never sends EOSE, and `fetch_directory_events_with_completion` folds those with `outcome.complete &= endpoint.complete`. Dead relays in published NIP-65 lists are the norm, not the exception, so `complete` will be false for a large fraction of real accounts.

Combine that with `uncertain_absence` below and the primary onboarding case regresses: an existing Nostr identity that has a kind `10002` but no kind `10050` yet (i.e. every first-time Marmot login) now returns `Err` as soon as any one of its advertised write relays is down, instead of publishing the missing inbox list. The caller then falls through to `complete_cached_relay_list_status`, which errors with `MissingRelayLists`, and for the external-signer branch (`runtime/mod.rs:6018-6025`) the error propagates straight out with no cache fallback at all.

This is the inverse failure of the one #1703 is about: rather than publishing a wrong default, the user cannot onboard.

Suggest making the evidence per-relay rather than all-or-nothing: if at least one safe advertised outbox answered authoritatively (EOSE) and carried no kind `10050`, that is enough evidence to proceed with publication; reserve `uncertain_absence` for the case where *no* endpoint in the hop completed.

crates/marmot-app/src/directory/methods.rs
```diff
@@ -142,6 +143,96 @@ impl MarmotApp {
...
+            !complete && (status.nip65.relays.is_empty() || status.inbox.relays.is_empty());
+        if explicit_empty || uncertain_absence {
```
erskingardner: **Confirmed absence and unknown absence are returned as the same opaque error.**

Issue #1703's acceptance criteria require that "a confirmed absence after successful first- and second-hop queries remains distinguishable from timeout, transport failure, unsafe endpoints, stale/future-dated records, and cancellation." Both `explicit_empty` (a confirmed, signed empty list — the user must publish a relay list) and `uncertain_absence` (retry later) collapse into the identical `AppError::RelayDirectory` with the identical string. `privacy_safe_kind()` cannot separate them either, so the UniFFI/C consumers driving the login UI get one undifferentiated failure and cannot tell "publish your relay list" from "your network is flaky".

Two distinct `AppError` variants (or at minimum distinct messages plus distinct `privacy_safe_kind` values) would satisfy the criterion.

crates/marmot-app/src/directory/methods.rs
```diff
@@ -142,6 +143,96 @@ impl MarmotApp {
...
+        if !outbox_relays.is_empty() {
+            let second = self
+                .relay_plane
+                .fetch_directory_events_with_completion(
+                    outbox_relays,
+                    relay_list_queries(account_id_hex.clone()),
+                )
+                .await
+                .map_err(|e| AppError::RelayDirectory(format!("fetch outbox relay lists: {e}")))?;
+            complete &= second.complete;
```
erskingardner: **A second sequential network hop was added inside two unchanged deadlines.**

`resolve_account_relay_list_status_for_account_id` awaits the first hop, then awaits the second. Worst case per hop is `DIRECTORY_RELAY_CONNECT_WAIT` (5s) + `DIRECTORY_RELAY_FETCH_WAIT` (3s), so 16s total — but the setup call site wraps this in `bounded_advisory_step(..., ACCOUNT_SETUP_ADVISORY_WAIT, ...)`, which is still 10s (`lib.rs:384`). A slow-but-safe advertised outbox now reliably blows the advisory cap and drops setup into the `None` arm → `complete_cached_relay_list_status` → `MissingRelayLists`.

The same applies on the invite path: `MEMBER_RESOLUTION_DEADLINE` is still 10s (`member_key_packages.rs:37`) but now has to cover discovery hop + outbox hop + KeyPackage hop, where it previously covered two.

Neither constant was revisited in this PR. Either the hops should overlap where they can, or the deadlines need to account for the extra stage.

crates/marmot-app/src/directory/member_key_packages.rs
```diff
@@ -363,25 +354,11 @@ impl MarmotApp {
...
-                    "member invite inbox discovery",
-                );
-                (safe_inbox.is_empty()
-                    || (outcomes[index].is_none() && target.relay_lists.nip65.relays.is_empty()))
-                .then_some(index)
-            })
+        // Durable projections have unknown observation age and need a bounded
+        // refresh. Process-local prewarm entries have an enforced TTL and were
+        // already resolved through the outbox path during composition.
+        let relay_list_unresolved = (0..targets.len())
```
erskingardner, line 403: **Every non-prewarmed target now hits the network for relay lists on every resolution.**

The previous filter skipped targets that already had a safe inbox and either a resolved outcome or a cached NIP-65. This replaces it with "everything except this process's fresh prewarm set", so a member whose complete, valid relay lists are already in the durable directory cache — including a member resolved from a *local* account or from `validated_cached_key_package` a few lines above — now performs both relay-list hops on every `resolve_member_key_packages` call.

Combined with the deadline point on `resolve_account_relay_list_status_for_account_id`, inviting a member that previously resolved entirely from cache now depends on that member's outbox relay being reachable within `MEMBER_RESOLUTION_DEADLINE`. `outcomes[index] = Some(Err(error))` at line 371 can also overwrite a KeyPackage that was already resolved successfully from local/cached state.

The issue explicitly permits a cache short-circuit: "A bounded freshness/cache rule may skip the network only when it has evidence that no newer target-authored record can be required." `self.directory_freshness()` already exists and is used exactly this way for KeyPackages (`fresh_or_cached_key_package`); applying it to the persisted relay-list observation would keep the fix while restoring the cache path. The comment "Durable projections have unknown observation age" is the thing to fix — persist the observation age — rather than a reason to always refetch.

crates/marmot-app/src/relay_plane/directory.rs
```diff
@@ -232,6 +253,27 @@ impl DirectoryRelayPlane {
             .map_err(|_| "directory fetch owner dropped before completing".to_owned())?
     }
 
+    pub(crate) async fn fetch_events_with_completion(
```
erskingardner, line 259: **This drops the in-flight coalescing that `fetch_events` provides.**

`fetch_events` (just above) keys concurrent identical requests through `DirectoryFetchKey` / `state.inflight` and fans one fetch out to all waiters, incrementing `coalesced_waiters`. `fetch_events_with_completion` spawns unconditionally.

Because this PR routes the entire relay-list path — account setup plus both member-resolution hops for every target — onto the new method, overlapping relay-list fetches (concurrent prewarm and invite, setup racing reconcile, two groups sharing members) are no longer deduplicated. That is a straight increase in relay traffic on top of point 5 above, and `DirectoryRelayStats.inflight_fetches` no longer reflects these fetches at all.

Also note `completed_fetches += 1` counts a `complete: false` outcome as a completed fetch, so the telemetry no longer distinguishes the state this PR exists to track.

crates/marmot-app/src/relay_plane/directory.rs
```diff
@@ -490,6 +532,161 @@ impl DirectoryRelayFetcher for NostrSdkDirectoryRelayFetcher {
...
+    }
+}
+
+async fn strict_fetch_endpoint(
+    client: NostrSdkClient,
+    relay_url: RelayUrl,
+    queries: Vec<DirectoryEventQuery>,
+) -> DirectoryFetchOutcome {
+    let endpoint = TransportEndpoint(relay_url.to_string());
+    if client.relay(&relay_url).await.is_err() && client.add_relay(relay_url.clone()).await.is_err()
```
erskingardner, line 621: **`add_relay` without a matching `remove_relay` on failure, unlike the fetch path right above it.**

`fetch_directory_events` (lines 435-503) deliberately tracks `added[index]` and calls `client.remove_relay(...)` for every relay it added that failed to connect, with a comment explaining why. `strict_fetch_endpoint` adds the relay and then returns `DirectoryFetchOutcome::default()` on connect timeout/failure without removing it, so the entry stays in the shared long-lived `NostrSdkClient` and keeps doing background reconnect work.

Second-hop endpoints come from the *target's* NIP-65 list, and `DirectoryFetchRequest::new` puts no cap on endpoint count, so resolving a batch of members accretes one permanent client relay entry per advertised outbox ever seen — including outboxes belonging to accounts the user merely tried to invite once. Worth mirroring the add/remove bookkeeping, and worth capping the number of discovered outboxes dialed per target.

crates/marmot-app/src/relay_plane/directory.rs
```diff
@@ -120,6 +129,18 @@ pub(crate) trait DirectoryRelayFetcher: Send + Sync {
...
+
+    async fn fetch_directory_events_with_completion(
+        &self,
+        request: DirectoryFetchRequest,
+    ) -> Result<DirectoryFetchOutcome, String> {
+        self.fetch_directory_events(request)
+            .await
+            .map(|records| DirectoryFetchOutcome {
+                records,
+                complete: true,
```
erskingardner: **Non-blocking: this default is fail-open for the exact invariant the PR establishes.**

Any `DirectoryRelayFetcher` that does not override this claims its read was authoritative. That is the wrong direction for a contract whose whole purpose is "a failed or timed-out hop is unknown, not proof of absence" — a future fetcher that forgets to implement completion silently re-introduces #1703.

It also means the test doubles inherit it: `MemberResolutionDirectoryFetcher` (`src/tests.rs:305`) and the three fetchers in `relay_plane/tests.rs` all report `complete: true` unconditionally, so no unit test in this PR exercises a `complete: false` path. `strict_fetch_endpoint`'s `Closed`, `AuthenticationFailed`, `Shutdown`, `RelayStatus::Disconnected` and record-cap-truncation branches — the ones that produce the evidence the new error depends on — have no direct coverage; only connect-timeout is covered, and only through the `held_discovery` integration test.

Making the method required (no default), or defaulting to `complete: false`, would force each implementation to be explicit.

crates/marmot-app/src/relay_plane/directory.rs
```diff
@@ -490,6 +532,161 @@ impl DirectoryRelayFetcher for NostrSdkDirectoryRelayFetcher {
...
+                Ok(RelayNotification::AuthenticationFailed | RelayNotification::Shutdown) => {
+                    break false;
+                }
+                Ok(RelayNotification::RelayStatus {
+                    status:
+                        RelayStatus::Disconnected | RelayStatus::Terminated | RelayStatus::Banned,
+                }) => {
+                    break false;
+                }
+                Err(_) => break false,
```
erskingardner, line 724: **Non-blocking: `Err(_)` here also catches `broadcast::RecvError::Lagged`.**

`relay.notifications()` is a broadcast receiver. Under load — a shared relay pushing notifications faster than this loop drains them — `recv()` returns `Err(Lagged(n))`, which is recoverable: the next `recv()` resumes. Treating it as a terminal incomplete read means a busy relay is reported as "absence not authoritatively established", which now escalates all the way to a login failure (see the `uncertain_absence` comment on `methods.rs`).

Matching `Err(RecvError::Lagged(_))` as `continue` (and only `Closed` as `break false`) would avoid load-dependent false negatives.

crates/marmot-app/src/directory/member_key_packages.rs
```diff
@@ -634,35 +627,218 @@ impl MarmotApp {
...
+                                }
                             }
                             Err(error) => failures.push((index, error)),
                         }
                     }
                 }
             }
         }
+
+        // Group targets by the same safe outbox set so each relay receives one
```
erskingardner: **Non-blocking: this second-hop block is a near-verbatim copy of the first-hop block above it.**

Lines 676-847 repeat the structure of `resolve_missing_relay_lists`'s first hop almost step for step: group/chunk by `MEMBER_RESOLUTION_AUTHORS_PER_QUERY`, build the same two `DirectoryEventQuery` values, `buffered(MEMBER_RESOLUTION_RELAY_CONCURRENCY)`, the same single-author fallback at `MEMBER_RESOLUTION_FALLBACK_CONCURRENCY`, the same `relay_records.extend` → `relay_lists_from_records` merge, and the same push/retain failure bookkeeping. The literal `"relay-list absence was not authoritatively established"` appears four times in this file and once in `methods.rs`.

The same duplication exists one level up: `resolve_account_relay_list_status_for_account_id` is a second copy of `fetch_account_relay_list_status_for_account_id` with *different* cached-merge semantics (timestamp merge vs. per-kind "fill only when not observed"). Two nearly identical resolvers with subtly different merge rules will drift.

Issue #1703 asked for "one target-aware relay-list resolution contract used by existing-account setup and member resolution". A single `hop(endpoints, targets) -> (records, complete, failures)` helper, called twice, would deliver that contract and cut most of the +245 lines here.

Minor, same function: `let needs_discovery = unresolved.to_vec();` at line 522 is an unused rename — `unresolved` is already the slice, and the clone is only read.

erskingardner (changes requested):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T11:02:43Z` |
| Commit reviewed | `669e67fa98ced53dad311db2d6219536e616701e` |

## Premise

The core fix is right and well grounded. Issue #1703 is reproduced, the `D -> W -> I` second hop is the correct contract, and the merge/anti-clobber and outbox-batching work is careful. The integration coverage — split-metadata login in both signer modes, stale-first-hop masking, independent-sender restart, exact-once Welcome — is genuinely good and matches the acceptance criteria it targets.

My concern is not the hop. It is the **failure policy layered on top of it**, which turns "I could not confirm" into a hard login error. That policy is broader than #1703 asked for (#1679 nominally owns the readiness contract), and as written it has a plausible regression that is worse than the bug being fixed: an existing Nostr identity with a kind `10002` but no kind `10050` — the ordinary first-time Marmot login — now fails whenever any one of its advertised write relays is unreachable, because a single failed dial sets `complete = false` for the entire hop. Dead relays in published NIP-65 lists are common. Before this PR that user onboarded; after it they cannot.

Green CI does not cover this: the incompleteness tests all use a *fully* unreachable discovery relay, never a partially reachable outbox set. Note also that `import_with_stalled_discovery_endpoint_completes_within_the_advisory_cap` had to be given published `10002`/`10050` records to keep passing — that edit is itself a signal that the new hard-fail changes previously-working behavior, and it deserves a line in the PR body.

## Blocking

1. **Read-only NIP-65 is misread as an intentional empty list.** `nip65.relays` is the *write* projection, so a `10002` marked all-`read` yields `relays == []` with `created_at > 0`, hits `explicit_empty`, and fails login — stickily, since the empty status is persisted first. (`directory/methods.rs`)
2. **All-or-nothing completeness blocks the main onboarding path.** One dead advertised outbox sets `complete = false` for the whole hop; make the evidence per-relay. (`relay_plane/directory.rs`, `directory/methods.rs`)
3. **Confirmed absence and unknown absence share one opaque error**, which directly contradicts the acceptance criterion requiring them to stay distinguishable. FFI callers cannot tell "publish a relay list" from "retry". (`directory/methods.rs`)
4. **A third sequential network stage was added inside unchanged deadlines** — `ACCOUNT_SETUP_ADVISORY_WAIT` (10s) now bounds up to ~16s of work, and `MEMBER_RESOLUTION_DEADLINE` (10s) now covers three hops instead of two.
5. **Member resolution always refetches relay lists**, discarding the cache short-circuit the issue explicitly permits, so a previously-cached invite now depends on the invitee's outbox being reachable. `directory_freshness()` already exists for exactly this. (`directory/member_key_packages.rs`)
6. **`fetch_events_with_completion` drops the in-flight coalescing** that `fetch_events` provides, and the whole relay-list path moved onto it. (`relay_plane/directory.rs`)

## Non-blocking

7. `strict_fetch_endpoint` adds relays to the shared client and never removes them on failure, unlike the sibling fetch path; endpoints are target-controlled and uncapped.
8. The trait's `complete: true` default is fail-open for the invariant this PR establishes, and it means no unit test exercises `complete: false` — the CLOSED / auth-failed / disconnected / truncation branches are untested.
9. `Err(_) => break false` swallows `broadcast::RecvError::Lagged` as a terminal incomplete read.
10. ~170 lines of the second hop duplicate the first hop, and `resolve_account_relay_list_status_for_account_id` duplicates `fetch_account_relay_list_status_for_account_id` with *different* merge semantics. The issue asked for "one contract"; one shared hop helper would deliver it and remove most of the added lines.
11. `remember_directory_relay_lists` is still a blind overwrite. The anti-clobber merge lives only in the two new call sites, so `fetch_latest_key_package_for_account_id` (CLI) can still persist a stale one-hop `10050` over a newer outbox-resolved one. Separately, `refresh_directory_entry_for_account_id` appears to have no callers anywhere in the workspace.
12. `relay_list_status_from_records` silently changes replaceable-event tie-breaking from "lexically highest event id" to "lexically lowest". That is the NIP-01-correct direction and it is tested, but it is an unannounced behavior change — worth a line in the PR body.

## Scope

Against the repo's simplicity guidance: +1624/-90 for a fix whose core is "follow the advertised outbox before declaring the inbox missing". Most of the non-test bulk is the duplicated second hop (10) and the second resolver (10); the failure-policy expansion (1-3) is the part that most looks like it belongs in a separate change with its own error taxonomy.

crates/marmot-app/src/directory/methods.rs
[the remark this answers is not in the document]
Datawav: The target-aware resolver now selects each relay-list kind across both hops and cached state by timestamp, preserving the accepted cached value on ties. Real-relay tests cover newer discovery inboxes against older positive and older empty outbox records in both signer modes.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/runtime/mod.rs
```diff
@@ -5955,7 +5955,7 @@ impl AccountManager {
                         &self.shared.app_performance_telemetry(),
                         ACCOUNT_SETUP_ADVISORY_WAIT,
                         "import_relay_list_status",
-                        self.app.fetch_account_relay_list_status_for_account_id(
+                        self.app.resolve_account_relay_list_status_for_account_id(
```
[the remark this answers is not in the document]
Datawav: The retry retains the original setup discovery scope. Failed or timed-out discovery stays retryable and cannot authorize default publication; both signing login paths have no-publication regressions.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/methods.rs
Datawav: Partial reads retain usable positive records but cannot establish absence. The healthy-empty outbox plus unavailable sibling case remains retryable in both signer modes. This also follows the clarified recovery policy: explain incomplete setup and require explicit informed repair, never silently replace potentially existing relay metadata.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7). Exact healthy-empty W1 / unavailable W2 regressions for both signer modes: [9dd1a9d7](https://github.com/marmot-protocol/mdk/commit/9dd1a9d73301cde4758953377f061f20d305edea).

crates/marmot-app/src/directory/member_key_packages.rs
Datawav: Both metadata hops now fetch and reconcile records without persistence. Only commit-purpose resolution admits durable directory state; prewarm uses its bounded process-local cache. The regression checks persistent directory membership and subscription plans for a previously unknown candidate.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/member_key_packages.rs
Datawav: Both hops now share resolve_relay_list_hop, with normalized endpoint-set keys and author chunks. Equivalent shared outboxes produce grouped requests, and incompatible multi-author reads use bounded single-author fallback.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/methods.rs
```diff
@@ -142,6 +143,96 @@ impl MarmotApp {
...
+        let mut complete = first.complete;
+        if !outbox_relays.is_empty() {
+            let second = self
+                .relay_plane
+                .fetch_directory_events_with_completion(
+                    outbox_relays,
+                    relay_list_queries(account_id_hex.clone()),
+                )
+                .await
+                .map_err(|e| AppError::RelayDirectory(format!("fetch outbox relay lists: {e}")))?;
```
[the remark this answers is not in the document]
Datawav: The second-hop sanitizer/route-cap error now marks discovery incomplete while retaining the first-hop records. Tests cover 17 advertised safe outboxes with and without a usable first-hop inbox; the cap still prevents the oversized second dial.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/methods.rs
```diff
@@ -142,6 +143,96 @@ impl MarmotApp {
...
+                .map(|endpoint| endpoint.0.clone())
+                .collect();
+        }
+        {
+            let app = self.clone();
+            let account_id = account_id_hex.clone();
+            let remembered = status.clone();
+            blocking_app_task(move || app.remember_directory_relay_lists(&account_id, &remembered))
+                .await?;
+        }
```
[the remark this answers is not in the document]
Datawav: Applied the narrower persistence correction: retain observed signed metadata, including explicit-empty lists and positives from partial reads, before returning readiness errors; add caller discovery fallback only afterward. Central relay-list and KeyPackage persistence now also merge per kind, preventing older callers from clobbering newer metadata.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/member_key_packages.rs
Datawav: Err and Ok(complete: false) now enter the same bounded per-author fallback on both metadata hops, preserving partial batch records and author mapping. The real LocalRelay regression rejects multi-author filters with CLOSED on both discovery and outbox relays and verifies accepted single-author requests resolve the roster.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/member_key_packages.rs
Datawav: Discovery and outbox completion are tracked independently. An empty completed cached outbox cannot clear a failed discovery hop. The regression requires RelayDirectory, while an existing usable safe inbox can still satisfy readiness.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/methods.rs
Datawav: The error is now typed MissingRelayLists(Nip65), not unknown discovery, and the stored read-relay projection remains intact. I have intentionally not converted read-only declarations into write-enabled lists: the recovery requirement is an app warning plus a clearly explained user-approved repair, not a silent rewrite. Both signer-mode regressions preserve the signed read-only list with zero replacement publications.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/methods.rs
Datawav: I retained fail-closed absence proof because one healthy-empty outbox cannot establish what the unavailable sibling holds. Treating any EOSE as sufficient would reintroduce the earlier partial-fetch overwrite scenario. The clarified recovery contract is to show retry for unknown lookup, or a warning and explicit non-destructive repair for confirmed broken setup. Existing positive metadata remains usable despite partial failures; the uncertainty error is no longer converted into MissingRelayLists.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/methods.rs
Datawav: Signed empty/read-only missing write routes now return the existing typed MissingRelayLists variant; failed, timed-out, truncated or rejected-future discovery returns RelayDirectory. Setup preserves the original lookup error instead of replacing it with a missing-list result when the cache cannot satisfy readiness. Both signer-mode tests assert the missing variant for signed empty/read-only declarations.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/methods.rs
Datawav: Directory setup now has a dedicated bounded 26-second preflight budget for the two metadata hops plus advisory profile work. Member set resolution has a 50-second overall bound for metadata, KeyPackage and bounded fallback stages; exhaustion stays retryable. The unrelated single-hop profile advisory retains its shorter bound.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/member_key_packages.rs
Datawav: DirectoryFreshness currently rejects only future timestamps; it is not an observation-age TTL. Durable positive inboxes remain usable during incomplete refreshes, while unknown-age metadata still gets a bounded refresh. The existing five-minute process-local prewarm cache provides the safe short-circuit, and repeated reuse now preserves its original expiry instead of extending it indefinitely. I kept an observation-age schema migration out of this fix.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/relay_plane/directory.rs
Datawav: Completion-aware requests now coalesce identical keys and fan out to all waiters, including after the first caller is cancelled. They participate in inflight telemetry, and incomplete outcomes increment failed rather than completed counts. The regression checks coalescing, cancellation, counters and removal of the inflight entry.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/relay_plane/directory.rs
Datawav: Both strict metadata reads and legacy profile/KeyPackage reads now own request-scoped SDK clients and shut them down at completion. Discovered outboxes cannot accumulate in the long-lived subscription client or remove another caller's active relay. The existing safety-policy endpoint-count cap remains enforced before any dial.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/relay_plane/directory.rs
Datawav: The trait default now returns complete: false; fetchers must explicitly attest completion. Coverage includes the real CLOSED multi-author path, the fail-closed legacy default, cancellation/coalescing, and exact per-filter saturation. EOSE at a saturated filter limit retains positives but is not absence proof.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/relay_plane/directory.rs
Datawav: I retained the fail-closed result for Lagged: the missed notification could be the target's inbox EVENT immediately before EOSE, so continuing cannot safely establish absence. Request-scoped clients reduce unrelated notification traffic, and member batches may use the bounded fallback. Returning retryable uncertainty is safer than silently overwriting a declaration lost from the local receiver.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

crates/marmot-app/src/directory/member_key_packages.rs
Datawav: The duplicated member-hop implementations are replaced by one private resolve_relay_list_hop helper covering chunking, grouped fetches, bounded fallback, author filtering, record retention and completion. Account setup remains target-aware and the explicit diagnostic API remains one-hop, but both now use the same per-kind cached merge. Central persistence applies the same anti-clobber rule to other callers too.

Correction: [051dc14a](https://github.com/marmot-protocol/mdk/commit/051dc14a4928ad820c00f74ac6bc407188d1b8a7).

Datawav:
The remaining review-body points are covered by the corrective commits and the updated PR description:

- Central `remember_directory_relay_lists` and `remember_directory_key_package` now timestamp-merge relay metadata, so older one-hop/CLI observations cannot clobber newer outbox-resolved records. Signed-empty metadata remains durable.
- `refresh_directory_entry_for_account_id` is an existing public API present in the base revision, not a new unused helper introduced here. I retained that API rather than remove it in this fix.
- The description now explicitly documents lower-event-ID tie-breaking, the distinction between confirmed missing routes and unknown lookup, the new bounded deadlines, and the deliberate refusal to silently replace existing relay declarations.
- `just fast-ci`, the 20-test serial relay-runtime selection, and both additional partial-outbox login regressions passed locally. The initial concurrent startup timing failure and its unmodified serial rerun are disclosed in the description. [CI for the pushed head](https://github.com/marmot-protocol/mdk/actions/runs/34033120272) remains the full-suite gate.

All 19 inline findings now have responses linked to the corrections. The app warning and explicitly explained non-destructive repair remain separate client work; this PR does not claim that UI is shipped.

crates/marmot-app/src/directory/methods.rs
```diff
@@ -142,6 +125,108 @@ impl MarmotApp {
...
+                .map(|endpoint| endpoint.0.clone())
+                .collect();
+        }
+
+        let explicit_empty = (status.nip65.created_at > 0 && status.nip65.relays.is_empty())
+            || (status.inbox.created_at > 0 && status.inbox.relays.is_empty());
+        let uncertain_absence =
+            !complete && (status.nip65.relays.is_empty() || status.inbox.relays.is_empty());
+        if explicit_empty {
+            return Err(AppError::MissingRelayLists(status.missing.clone()));
```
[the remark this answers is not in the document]
Datawav: Fixed in [9cf27979](https://github.com/marmot-protocol/mdk/commit/9cf279795511cbb29fb093b1cc9cb2795b932d79): incomplete discovery now takes precedence over a retained signed-empty declaration. The regression checks complete D plus incomplete W returns RelayDirectory while preserving the observed empty record, then verifies that completing W changes the same case to MissingRelayLists(Inbox). The five focused account regressions passed. The stale failed-import unit assertion was also updated to require the retryable discovery error rather than confirmed absence.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T17:20:46Z` |
| Commit reviewed | `9cf279795511cbb29fb093b1cc9cb2795b932d79` |
| Model | `gpt-6-astra` |
| Reasoning level | `high` |

No remaining blocking findings on this head. This supersedes my previous request-changes reviews.

Verified the corrections in source and current CI:

- Both metadata hops share bounded batching and per-author fallback for incomplete outcomes, including actual relay CLOSED rejection. Partial positive records survive fallback.
- First-hop uncertainty remains independent of second-hop completion, so a successful empty outbox cannot manufacture confirmed absence.
- Account lookup preserves usable first-hop records on route-cap failure, merges newer records across hops/cache, and retains signed empty declarations. Incomplete lookup takes precedence over an empty declaration; neither permits replacement publication.
- Prewarm remains process-local, reuse does not extend its original expiry, and central relay/KeyPackage persistence preserves newer relay metadata.
- Completion reads coalesce, retain their owner when a waiter cancels, and participate in inflight/failure telemetry. Request-owned SDK clients shut down after reads. The expanded discovery deadlines remain inside the worker-response budget.

[Exact-head Required CI](https://github.com/marmot-protocol/mdk/actions/runs/34034206362) is green, including all four Rust shards, bindings, Clippy, CLI, simulator, and relay-runtime jobs. I checked the relay-runtime log: 144 passed, 0 failed, including real CLOSED fallback, both signing modes with healthy-empty W1/unavailable W2 and zero replacement publications, and the independent sender/receiver restart cases. [All platform binary builds](https://github.com/marmot-protocol/mdk/actions/runs/34034206290) are also green.

The four still-open discussion threads do not block my approval: preserving read-only declarations rather than silently adding write roles is appropriate; one healthy-empty relay cannot establish what an unavailable sibling holds; unknown-age durable metadata may require bounded refresh while the existing prewarm TTL permits reuse; and notification lag must invalidate absence proof because an inbox EVENT may have been lost. These are deliberate behavior choices, not evidence that the earlier defects remain. I have not changed thread-resolution state.

This is an API-only source/CI rereview; no local tests or workspace mutations. Durable client readiness/repair presentation and shipped-app verification remain the linked follow-ups, not claims of this approval.

PR #1707, opened 2026-09-06
fix(storage): version the shared store schema

erskingardner:
`shared.sqlite3` previously initialized tables with unversioned `CREATE TABLE IF NOT EXISTS` and cleared retired telemetry endpoints in a separate repair. A malformed existing table could be accepted, and schema changes had no shared-store version or downgrade boundary.

This adds independent `shared_schema_migrations`, starting at `1 / 0001_shared_store`. Each numbered body and ledger insertion commit in one immediate transaction. Failures roll back both, including endpoint neutralization; ignored or altered ledger insertions cannot confirm a migration. History must match the exact compiled version/name prefix, and newer versions return `StorageError::UnsupportedSchemaVersion`. Current opens use reads only; pending openers recheck history under the write lock.

Unversioned tables are recognized against frozen current and historical definitions before missing tables are created. Validation covers columns/types/nullability/defaults, primary keys, foreign keys, CHECK expressions, collations, table options, and index structure/uniqueness. Unrecognized definitions, live-table triggers, and live-schema foreign-key violations fail closed without recording v1. Public-directory users, ordered follows, live settings, timestamps, telemetry installation identity, and rowids are preserved.

Historical fixtures were independently extracted from:

- `9db9dbc0`: original nine-table schema, including four now-unused directory tables and inline audit data mode.
- `5464b34e`: five live tables with inline audit data mode; existing unused tables were never dropped.
- `fe133b82`: five current tables after retirement of audit data-mode creation.
- Compatibility tests at `9db9dbc0`: nullable legacy `otlp_endpoint` and pre-mode audit schema. The historical additive repair establishes the appended audit-column variant.

The appended endpoint layout is defensive compatibility coverage requested during review, not a claim about another historical or deployed schema.

These cover all three DDL generations across six master-lineage revisions touching shared.rs through `6b2b041b`. Fixture documentation separates repository/compatibility evidence from unknown deployed-build coverage.

Recognized retired columns remain in place to avoid table rebuilds: non-NULL endpoints in either the inline or defensively supported appended layout become NULL transactionally, while inline/appended audit `data_mode` stays inert and unchanged by later writes. Old unused directory tables and their rows remain untouched; pre-existing orphaned rows in those retired tables do not gate live-store adoption. SQLite errors retain extended result codes and transient BUSY/LOCKED classification while stripping database-controlled messages, including trigger text and stored values. Conversion errors retain safe categories, column indices and SQLite types without database-controlled names, values or nested causes; a shared-store result extension applies this mapper uniformly.

Non-goals: no session or app-cache migration changes, ledger coupling, directory movement, mirroring/reconciliation changes, encryption changes, crate split, error-enum redesign, telemetry/audit product change, data deletion to recover unknown schemas, or version bumps. Owner-only creation, WAL, synchronous NORMAL, 5-second busy timeout, foreign keys, trusted_schema, temp_store, and terminal close behavior are preserved.

Validation:

- TDD baseline: both initial contract tests failed against the old initializer (missing ledger; malformed table accepted).
- `cargo test -p storage-sqlite shared:: -- --nocapture`: 38 passed, including migration and upgrade assurance tests.
- `cargo test -p storage-sqlite`: 486 unit tests and 2 integration tests passed; 4 existing ignored tests; no failures.
- `cargo clippy -p storage-sqlite --all-targets -- -D warnings`: passed.
- `just fast-ci`: passed (including workspace default/OTLP check and clippy plus 5 convergence-policy tests).
- `cargo test -p cgka-conformance-simulator --test tracing_audit --locked`: 4 passed.
- Required docs stale-marker scan: both matches inspected (the scan command itself and the still-present engine Nostr-layering TODO); no stale migration guidance.
- `git diff --check`: passed.
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked app_runtime_readd_after_remove_resurfaces_removed_member_group -- --exact --nocapture`: 1 passed.
- `cargo test -p marmot-app --test relay_runtime --features test-policy-overrides --locked removed_member_triggers_local_push_token_cleanup -- --exact --nocapture`: 1 passed.

Assurance compares every typed value and rowid across 30,003 populated rows, exercises all historical fixtures, verifies integrity and foreign keys after upgrades, and covers body/ledger rollback plus bounded process-exit recovery with dirty-page spill under DELETE and WAL journaling. Process interruption is not power loss. These tests do not establish deployed-schema coverage, a field failure rate, or every disk-full/OS I/O failure. Conservative DDL recognition can refuse equivalent hand-edited schemas; refusal preserves data but can make the store unavailable pending a compatible binary or explicit repair. Full GitHub CI passed on `5f0dc400` ([run 34028647553](https://github.com/marmot-protocol/mdk/actions/runs/34028647553)), including Required CI, all Rust test shards, simulator/vector tests and the full relay-runtime suite. All platform binary builds also passed. The original head failed two relay-runtime tests; both passed focused local reruns using CI feature flags, and the token-cleanup assertion also failed identically on master run 33724316503. The updated head passed the full remote suite without weakening assertions or changing CI gates. This does not claim to eliminate the pre-existing relay-test flake.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved shared database upgrades with transactional migrations, schema validation, rollback protection, and recovery from interrupted upgrades.
  - Preserved compatible existing data and identifiers while handling legacy database layouts more safely.
  - Improved error handling for malformed, unsupported, or incomplete database schemas.

- **Documentation**
  - Documented shared database migration history, compatibility behavior, schema adoption, validation rules, and recovery guarantees.
  - Added reference fixtures covering historical database layouts and migration scenarios.

- **Tests**
  - Expanded coverage for upgrades, concurrent access, read-only databases, permissions, foreign-key validation, and migration failures.

### DIFF
--- a/crates/storage-sqlite/README.md
+++ b/crates/storage-sqlite/README.md
@@ -25,7 +25,7 @@ The crate is split around storage concerns:
 - `migrations.rs` owns the account/session migration runner and migration tests.
 - `shared/migrations.rs` owns the independent shared-store runner; `shared/v1.sql` freezes its initial schema,
   `shared/legacy.sql` defines recognized compatibility columns, and `shared/fixtures/` plus the migration and assurance
-  tests cover adoption and recovery.
+  tests cover adoption and recovery. `shared/error.rs` owns the privacy-safe error mapper and result extension.
 
 ## Migrations
 
@@ -51,8 +51,8 @@ Unversioned shared tables must match frozen current or verified historical defin
 covers columns, types, nullability, defaults, primary keys, foreign keys, CHECK expressions, collations and indexes.
 Conservative DDL comparison can refuse equivalent but unrecognized SQL; incompatible shapes fail with static errors
 and no version row. Missing tables are created. Public users, ordered follows, live settings, timestamps, installation
-identity and rowids are preserved. Existing unused directory tables remain untouched. The recognized nullable
-`otlp_endpoint` column is retained but non-NULL values are cleared transactionally; retained audit `data_mode` columns
+identity and rowids are preserved. Existing unused directory tables remain untouched; their pre-existing orphaned rows do not gate live-store adoption. The recognized nullable
+`otlp_endpoint` column (inline or appended) is retained but non-NULL values are cleared transactionally; retained audit `data_mode` columns
 (inline or historically appended) remain inert and unchanged by subsequent settings writes. No full-data audit mode
 is reintroduced. SQLite errors retain extended result codes and transient BUSY/LOCKED classification without
 SQLite's database-controlled message.
--- a/crates/storage-sqlite/README.md
+++ b/crates/storage-sqlite/README.md
@@ -22,11 +22,14 @@ The crate is split around storage concerns:
   and SQLite error mapping; `shared.rs` owns `SqliteSharedStorage`, a separate (non-account-scoped) database for
   cross-identity state: the public-directory user cache, relay-telemetry settings, audit-log settings, and the telemetry
   install id.
-- `migrations.rs` owns the migration runner and migration tests.
+- `migrations.rs` owns the account/session migration runner and migration tests.
+- `shared/migrations.rs` owns the independent shared-store runner; `shared/v1.sql` freezes its initial schema,
+  `shared/legacy.sql` defines recognized compatibility columns, and `shared/fixtures/` plus the migration and assurance
+  tests cover adoption and recovery.
 
 ## Migrations
 
-Schema changes go through Rust migrations, not external SQL files. The runner and ordered registry live in
+Account/session schema changes go through Rust migrations. The runner and ordered registry live in
 `src/migrations.rs`; migration bodies live in numbered files like `src/migrations/0001_initial_schema.rs`. Each
 migration has a monotonically increasing integer version, a matching padded name, and an `apply` function that runs
 inside a SQLite transaction. That function can execute DDL, rewrite rows, or perform more complex data-shape changes
@@ -34,6 +37,30 @@ when storage semantics evolve.
 
 Applied migrations are recorded in `cgka_schema_migrations`. Opening an encrypted database applies any missing
 migrations after SQLCipher keying and before storage handles are exposed.
+
+The three current app database categories have independent histories: `session.sqlite` uses
+`cgka_schema_migrations`, the per-account `app-cache.sqlite3` uses marmot-app's `app_cache_schema_migrations`, and
+installation-wide `shared.sqlite3` uses `shared_schema_migrations`. No runner reads another store's ledger.
+
+Shared version `1 / 0001_shared_store` establishes the five live tables. Each migration body and its ledger row commit
+in the same immediate transaction; failure rolls back both. Recorded versions and names must be an exact prefix of
+the compiled registry. Future versions return `StorageError::UnsupportedSchemaVersion`. A pending opener rechecks the
+prefix under the write lock; an already-current opener validates history using reads only.
+
+Unversioned shared tables must match frozen current or verified historical definitions before adoption. Validation
+covers columns, types, nullability, defaults, primary keys, foreign keys, CHECK expressions, collations and indexes.
+Conservative DDL comparison can refuse equivalent but unrecognized SQL; incompatible shapes fail with static errors
+and no version row. Missing tables are created. Public users, ordered follows, live settings, timestamps, installation
+identity and rowids are preserved. Existing unused directory tables remain untouched. The recognized nullable
+`otlp_endpoint` column is retained but non-NULL values are cleared transactionally; retained audit `data_mode` columns
+(inline or historically appended) remain inert and unchanged by subsequent settings writes. No full-data audit mode
+is reintroduced. SQLite errors retain extended result codes and transient BUSY/LOCKED classification without
+SQLite's database-controlled message.
+
+The shared store remains unencrypted and owner-only, with WAL, synchronous NORMAL, a 5-second busy timeout,
+foreign keys ON, trusted_schema OFF, temp_store MEMORY and terminal close behavior. Its fixture provenance and
+assurance limits are documented in [shared/fixtures/README.md](src/shared/fixtures/README.md) and the
+[app storage boundaries](../../docs/marmot-architecture/further-context/app-sqlite-storage-boundaries.md).
 
 ## Operational boundary
 
--- a/crates/storage-sqlite/src/shared.rs
+++ b/crates/storage-sqlite/src/shared.rs
@@ -1,3 +1,5 @@
+mod migrations;
+
 use crate::connection::CachedSql;
 use std::path::Path;
 use std::sync::Arc;
@@ -5,8 +7,7 @@ use std::time::Duration;
 
 use crate::connection::{CloseableConnection, ConnectionGuard};
 use crate::{
-    SqliteResultExt, bool_i64, connection::retry_on_busy, optional_u64_to_i64, u64_to_i64,
-    unix_now_ms, usize_to_i64,
+    bool_i64, connection::retry_on_busy, optional_u64_to_i64, u64_to_i64, unix_now_ms, usize_to_i64,
 };
 use cgka_traits::storage::{StorageError, StorageResult};
 use rusqlite::{OptionalExtension, TransactionBehavior, params};
@@ -68,12 +69,14 @@ impl SqliteSharedStorage {
         // The shared cache is unencrypted, so file-level 0600 (including
         // sidecars) is the only thing keeping it from other local users.
         crate::connection::ensure_private_db_files(path)?;
-        let conn = rusqlite::Connection::open(path).storage()?;
+        let conn = rusqlite::Connection::open(path).map_err(migrations::sqlite_error)?;
         Self::from_connection(conn)
     }
 
     pub fn in_memory() -> StorageResult<Self> {
-        Self::from_connection(rusqlite::Connection::open_in_memory().storage()?)
+        Self::from_connection(
+            rusqlite::Connection::open_in_memory().map_err(migrations::sqlite_error)?,
+        )
     }
 
     /// Checkpoint the WAL and close this shared database, releasing its file
@@ -91,58 +94,20 @@ impl SqliteSharedStorage {
         self.conn.is_closed()
     }
 
-    fn from_connection(conn: rusqlite::Connection) -> StorageResult<Self> {
+    fn from_connection(mut conn: rusqlite::Connection) -> StorageResult<Self> {
         conn.busy_timeout(Duration::from_millis(SHARED_BUSY_TIMEOUT_MS))
-            .storage()?;
-        conn.pragma_update(None, "foreign_keys", true).storage()?;
+            .map_err(migrations::sqlite_error)?;
+        conn.pragma_update(None, "foreign_keys", true)
+            .map_err(migrations::sqlite_error)?;
         conn.pragma_update(None, "trusted_schema", false)
-            .storage()?;
+            .map_err(migrations::sqlite_error)?;
         conn.execute_batch(
             "PRAGMA journal_mode = WAL;
              PRAGMA synchronous = NORMAL;
              PRAGMA temp_store = MEMORY;",
         )
-        .storage()?;
-        conn.execute_batch(
-            r#"
-CREATE TABLE IF NOT EXISTS directory_users (
-    account_id_hex TEXT PRIMARY KEY NOT NULL,
-    npub TEXT NOT NULL,
-    profile_json TEXT,
-    relay_lists_json TEXT NOT NULL,
-    key_package_json TEXT,
-    event_id_hex TEXT,
-    event_kind INTEGER,
-    event_created_at INTEGER
-);
-CREATE TABLE IF NOT EXISTS directory_user_follows (
-    account_id_hex TEXT NOT NULL REFERENCES directory_users(account_id_hex) ON DELETE CASCADE,
-    follow_account_id_hex TEXT NOT NULL,
-    position INTEGER NOT NULL,
-    event_id_hex TEXT,
-    event_created_at INTEGER,
-    PRIMARY KEY (account_id_hex, follow_account_id_hex)
-);
-	CREATE TABLE IF NOT EXISTS relay_telemetry_settings (
-	    id INTEGER PRIMARY KEY CHECK (id = 1),
-	    export_enabled INTEGER NOT NULL DEFAULT 0,
-	    export_interval_seconds INTEGER NOT NULL DEFAULT 60,
-	    updated_at_ms INTEGER NOT NULL
-	);
-		CREATE TABLE IF NOT EXISTS audit_log_settings (
-		    id INTEGER PRIMARY KEY CHECK (id = 1),
-		    enabled INTEGER NOT NULL DEFAULT 0,
-		    updated_at_ms INTEGER NOT NULL
-		);
-		CREATE TABLE IF NOT EXISTS telemetry_install (
-		    id INTEGER PRIMARY KEY CHECK (id = 1),
-		    install_id TEXT NOT NULL,
-		    updated_at_ms INTEGER NOT NULL
-		);
-		"#,
-        )
-        .storage()?;
-        Self::clear_legacy_relay_telemetry_endpoint(&conn)?;
+        .map_err(migrations::sqlite_error)?;
+        migrations::run_all(&mut conn)?;
         Ok(Self {
             conn: Arc::new(CloseableConnection::new(conn, CLOSED_DETAIL)),
         })
@@ -156,7 +121,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
             let mut conn = self.lock()?;
             let tx = conn
                 .transaction_with_behavior(TransactionBehavior::Immediate)
-                .storage()?;
+                .map_err(migrations::sqlite_error)?;
             tx.execute_cached(
                 "INSERT INTO directory_users (
                 account_id_hex, npub, profile_json, relay_lists_json, key_package_json,
@@ -182,12 +147,12 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                     optional_u64_to_i64(record.event_created_at)?,
                 ],
             )
-            .storage()?;
+            .map_err(migrations::sqlite_error)?;
             tx.execute_cached(
                 "DELETE FROM directory_user_follows WHERE account_id_hex = ?1",
                 params![&record.account_id_hex],
             )
-            .storage()?;
+            .map_err(migrations::sqlite_error)?;
             for (position, follow) in record.follows.iter().enumerate() {
                 tx.execute_cached(
                     "INSERT OR IGNORE INTO directory_user_follows (
@@ -202,9 +167,9 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                         optional_u64_to_i64(record.event_created_at)?,
                     ],
                 )
-                .storage()?;
+                .map_err(migrations::sqlite_error)?;
             }
-            tx.commit().storage()
+            tx.commit().map_err(migrations::sqlite_error)
         })
     }
 
@@ -213,7 +178,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
         account_id_hex: &str,
     ) -> StorageResult<Option<PublicDirectoryUserRecord>> {
         let mut conn = self.lock()?;
-        let tx = conn.transaction().storage()?;
+        let tx = conn.transaction().map_err(migrations::sqlite_error)?;
         let Some(mut record) = tx
             .query_row_cached(
                 "SELECT account_id_hex, npub, profile_json, relay_lists_json,
@@ -236,9 +201,9 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                 },
             )
             .optional()
-            .storage()?
+            .map_err(migrations::sqlite_error)?
         else {
-            tx.commit().storage()?;
+            tx.commit().map_err(migrations::sqlite_error)?;
             return Ok(None);
         };
         let mut stmt = tx
@@ -247,14 +212,14 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                  WHERE account_id_hex = ?1
                  ORDER BY position, follow_account_id_hex",
             )
-            .storage()?;
+            .map_err(migrations::sqlite_error)?;
         record.follows = stmt
             .query_map(params![account_id_hex], |row| row.get::<_, String>(0))
-            .storage()?
+            .map_err(migrations::sqlite_error)?
             .collect::<Result<Vec<_>, _>>()
-            .storage()?;
+            .map_err(migrations::sqlite_error)?;
         drop(stmt);
-        tx.commit().storage()?;
+        tx.commit().map_err(migrations::sqlite_error)?;
         Ok(Some(record))
     }
 
@@ -277,7 +242,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
         // each user's follows by position then follow id).
         let cap = i64::try_from(max).unwrap_or(i64::MAX);
         let mut conn = self.lock()?;
-        let tx = conn.transaction().storage()?;
+        let tx = conn.transaction().map_err(migrations::sqlite_error)?;
         let mut follows_by_account: std::collections::HashMap<String, Vec<String>> =
             std::collections::HashMap::new();
         {
@@ -290,14 +255,14 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                      )
                      ORDER BY account_id_hex, position, follow_account_id_hex",
                 )
-                .storage()?;
+                .map_err(migrations::sqlite_error)?;
             let rows = stmt
                 .query_map(params![cap], |row| {
                     Ok((row.get::<_, String>(0)?, row.get::<_, String>(1)?))
                 })
-                .storage()?;
+                .map_err(migrations::sqlite_error)?;
             for row in rows {
-                let (account_id_hex, follow) = row.storage()?;
+                let (account_id_hex, follow) = row.map_err(migrations::sqlite_error)?;
                 follows_by_account
                     .entry(account_id_hex)
                     .or_default()
@@ -312,7 +277,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                  ORDER BY account_id_hex
                  LIMIT ?1",
             )
-            .storage()?;
+            .map_err(migrations::sqlite_error)?;
         let records = stmt
             .query_map(params![cap], |row| {
                 Ok(PublicDirectoryUserRecord {
@@ -327,11 +292,11 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                     follows: Vec::new(),
                 })
             })
-            .storage()?
+            .map_err(migrations::sqlite_error)?
             .collect::<Result<Vec<_>, _>>()
-            .storage()?;
+            .map_err(migrations::sqlite_error)?;
         drop(stmt);
-        tx.commit().storage()?;
+        tx.commit().map_err(migrations::sqlite_error)?;
         if records.len() >= max {
             // no silent caps: surface (aggregate count only, privacy-safe) that
             // the listing was truncated so an operator can tell the bound bit.
@@ -369,7 +334,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                     })
                 },
             )
-            .storage()
+            .map_err(migrations::sqlite_error)
     }
 
     pub fn set_relay_telemetry_settings(
@@ -393,7 +358,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                         unix_now_ms(),
                     ],
                 )
-                .storage()?;
+                .map_err(migrations::sqlite_error)?;
             Ok(())
         })
     }
@@ -408,7 +373,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                 |row| row.get(0),
             )
             .optional()
-            .storage()
+            .map_err(migrations::sqlite_error)
     }
 
     pub fn set_telemetry_install_id(&self, install_id: &str) -> StorageResult<()> {
@@ -422,7 +387,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                     updated_at_ms = excluded.updated_at_ms",
                     params![install_id, unix_now_ms()],
                 )
-                .storage()?;
+                .map_err(migrations::sqlite_error)?;
             Ok(())
         })
     }
@@ -441,7 +406,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                     })
                 },
             )
-            .storage()
+            .map_err(migrations::sqlite_error)
     }
 
     pub fn set_audit_log_settings(&self, settings: &StoredAuditLogSettings) -> StorageResult<()> {
@@ -455,7 +420,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                     updated_at_ms = excluded.updated_at_ms",
                     params![bool_i64(settings.enabled), unix_now_ms()],
                 )
-                .storage()?;
+                .map_err(migrations::sqlite_error)?;
             Ok(())
         })
     }
@@ -487,29 +452,9 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                  ON CONFLICT(id) DO NOTHING",
                     params![unix_now_ms()],
                 )
-                .storage()?;
+                .map_err(migrations::sqlite_error)?;
             Ok(())
         })
-    }
-
-    fn clear_legacy_relay_telemetry_endpoint(conn: &rusqlite::Connection) -> StorageResult<()> {
-        let columns = {
-            let mut stmt = conn
-                .prepare_cached("PRAGMA table_info(relay_telemetry_settings)")
-                .storage()?;
-            stmt.query_map([], |row| row.get::<_, String>(1))
-                .storage()?
-                .collect::<Result<Vec<_>, _>>()
-                .storage()?
-        };
-        if columns.iter().any(|column| column == "otlp_endpoint") {
-            conn.execute_cached(
-                "UPDATE relay_telemetry_settings SET otlp_endpoint = NULL",
-                [],
-            )
-            .storage()?;
-        }
-        Ok(())
     }
 
     fn ensure_audit_log_settings(&self) -> StorageResult<()> {
@@ -523,7 +468,7 @@ CREATE TABLE IF NOT EXISTS directory_user_follows (
                  ON CONFLICT(id) DO NOTHING",
                     params![unix_now_ms()],
                 )
-                .storage()?;
+                .map_err(migrations::sqlite_error)?;
             Ok(())
         })
     }
@@ -847,5 +792,33 @@ mod tests {
             )
             .unwrap();
         assert_eq!(legacy_mode, "full_data");
+    }
+}
+
+#[cfg(test)]
+mod migration_contract_tests {
+    use super::*;
+
+    #[test]
+    fn fresh_shared_storage_records_version_one() {
+        let storage = SqliteSharedStorage::in_memory().unwrap();
+        let row: (i64, String) = storage
+            .lock()
+            .unwrap()
+            .query_row(
+                "SELECT version, name FROM shared_schema_migrations",
+                [],
+                |r| Ok((r.get(0)?, r.get(1)?)),
+            )
+            .unwrap();
+        assert_eq!(row, (1, "0001_shared_store".into()));
+    }
+
+    #[test]
+    fn shared_storage_refuses_malformed_unversioned_table() {
+        let conn = rusqlite::Connection::open_in_memory().unwrap();
+        conn.execute_batch("CREATE TABLE directory_users (account_id_hex TEXT)")
+            .unwrap();
+        assert!(SqliteSharedStorage::from_connection(conn).is_err());
     }
 }
--- a/crates/storage-sqlite/src/shared.rs
+++ b/crates/storage-sqlite/src/shared.rs
@@ -1,4 +1,7 @@
+mod error;
 mod migrations;
+
+use error::SharedSqliteResultExt;
 
 use crate::connection::CachedSql;
 use std::path::Path;
@@ -69,14 +72,12 @@ impl SqliteSharedStorage {
         // The shared cache is unencrypted, so file-level 0600 (including
         // sidecars) is the only thing keeping it from other local users.
         crate::connection::ensure_private_db_files(path)?;
-        let conn = rusqlite::Connection::open(path).map_err(migrations::sqlite_error)?;
+        let conn = rusqlite::Connection::open(path).storage()?;
         Self::from_connection(conn)
     }
 
     pub fn in_memory() -> StorageResult<Self> {
-        Self::from_connection(
-            rusqlite::Connection::open_in_memory().map_err(migrations::sqlite_error)?,
-        )
+        Self::from_connection(rusqlite::Connection::open_in_memory().storage()?)
     }
 
     /// Checkpoint the WAL and close this shared database, releasing its file
@@ -96,17 +97,16 @@ impl SqliteSharedStorage {
 
     fn from_connection(mut conn: rusqlite::Connection) -> StorageResult<Self> {
         conn.busy_timeout(Duration::from_millis(SHARED_BUSY_TIMEOUT_MS))
-            .map_err(migrations::sqlite_error)?;
-        conn.pragma_update(None, "foreign_keys", true)
-            .map_err(migrations::sqlite_error)?;
+            .storage()?;
+        conn.pragma_update(None, "foreign_keys", true).storage()?;
         conn.pragma_update(None, "trusted_schema", false)
-            .map_err(migrations::sqlite_error)?;
+            .storage()?;
         conn.execute_batch(
             "PRAGMA journal_mode = WAL;
              PRAGMA synchronous = NORMAL;
              PRAGMA temp_store = MEMORY;",
         )
-        .map_err(migrations::sqlite_error)?;
+        .storage()?;
         migrations::run_all(&mut conn)?;
         Ok(Self {
             conn: Arc::new(CloseableConnection::new(conn, CLOSED_DETAIL)),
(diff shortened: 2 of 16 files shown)


crates/storage-sqlite/src/shared/migrations.rs
```diff
@@ -0,0 +1,306 @@
...
+            return Err(invalid_shape());
+        }
+        if table_matches(tx, &reference, table)? {
+            continue;
+        }
+        if *table == "relay_telemetry_settings" && table_matches(tx, &legacy, table)? {
+            legacy_endpoint = true;
+        } else if !(*table == "audit_log_settings"
+            && (table_matches(tx, &legacy, table)? || table_matches(tx, &appended, table)?))
+        {
```
erskingardner: **Blocking: this narrows the plaintext endpoint scrub and can leave the value on disk permanently.**

Master neutralized `otlp_endpoint` on every open, for any `relay_telemetry_settings` whose `PRAGMA table_info` listed
a column with that name — position, declared type and nullability were irrelevant. Here the value is only cleared when
the whole table DDL normalizes equal to `legacy.sql`. Every other layout falls through to `invalid_shape()`, the store
is refused, and the plaintext endpoint is never cleared by any later open either, since the refusal is permanent for
that binary.

Concrete failure: a store whose `otlp_endpoint` was added by `ALTER TABLE ... ADD COLUMN` (so it sits after
`updated_at_ms` rather than between `export_enabled` and `export_interval_seconds`). `table_matches` against `legacy`
fails on both the normalized DDL and `pragma_table_xinfo` ordering, so `run_all` returns
`invalid shared store schema shape`, `shared_storage()` fails for every caller that touches it (telemetry settings,
audit settings, directory search and cache), and the plaintext collector URL — the thing the original repair existed
to remove — survives. Master would have NULLed it and kept working.

The asymmetry with the line below makes this hard to justify: `audit_log_settings` gets two recognized layouts
precisely *because* `9db9dbc0` appended its column with `ALTER TABLE`, yet `relay_telemetry_settings` gets one. And
`fixtures/README.md` says the endpoint shape is "compatibility-test evidence, not a claim about an identified deployed
build" — if the deployed shape is unknown, exact-DDL matching is the wrong tool for this particular column.

Two ways out, either is fine:

1. Recognize an appended endpoint variant the same way `appended` is built for the audit table, or
2. Before returning `invalid_shape()` for `relay_telemetry_settings`, fall back to the old name-based check
   (`pragma_table_xinfo` contains `otlp_endpoint`) and NULL it inside this same transaction. Refusing to *adopt* an
   unrecognized shape is defensible; refusing without first removing the plaintext value is not.

While here, the negated compound `else if` is genuinely hard to read for a three-way decision. A per-table alternatives
list reads better and makes adding a fourth recognized shape a one-line change:

```rust
let alternatives: &[&Connection] = match *table {
    "relay_telemetry_settings" => &[&legacy],
    "audit_log_settings" => &[&legacy, &appended],
    _ => &[],
};
```

crates/storage-sqlite/src/shared/migrations.rs
```diff
@@ -0,0 +1,306 @@
...
+        .prepare("PRAGMA foreign_key_check")
+        .map_err(sqlite_error)?;
+    if check
+        .query([])
+        .map_err(sqlite_error)?
+        .next()
+        .map_err(sqlite_error)?
+        .is_some()
+    {
+        return Err(invalid_shape());
```
erskingardner, line 220: `PRAGMA foreign_key_check` with no argument walks the entire database, including the four retired directory tables
this migration explicitly promises to leave untouched. `directory_key_packages` carries
`REFERENCES directory_users(account_id_hex) ON DELETE CASCADE`, so a single orphaned row in that dead table — from a
write made while `foreign_keys` was off, or any out-of-band mutation — refuses adoption of the whole shared store,
even though no current code path reads it.

The live schema has exactly one foreign key, so scoping the check says what you mean and cannot be tripped by tables
you have already decided not to own:

```rust
tx.prepare("PRAGMA foreign_key_check(directory_user_follows)")
```

This also drops a full-database scan from the one-time adoption on large caches.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T10:40:21Z` |
| Commit reviewed | `492a06ea8d8628e2cd5cd438ac29d48dd110cb8f` |

_(Posted as a comment review: GitHub does not allow requesting changes on your own pull request.)_

The premise is sound: `shared.sqlite3` was the last of the three app databases without a ledger, the prior work on
`cgka_schema_migrations` and `app_cache_schema_migrations` set the contract, and the docs already promised this
follow-up. The runner itself is careful — body-plus-ledger in one immediate transaction, exact-prefix history,
write-lock recheck, `RAISE(IGNORE)` defeat via the post-insert recount, read-only current-open path. All four Rust
test shards pass under nextest, including the child-process interruption test.

One blocking issue: **adoption narrows the plaintext `otlp_endpoint` scrub, and narrows it in the one direction the
PR itself says it cannot verify.** On master, `clear_legacy_relay_telemetry_endpoint` cleared the column on every
open for *any* table containing a column with that name. After this change the value is only neutralized when
`relay_telemetry_settings` matches `legacy.sql` byte-for-byte (modulo whitespace/case); every other layout is refused
by `invalid_shape()`, and the plaintext endpoint then stays on disk forever, unscrubbed and unscrubbable. The fixture
README is explicit that the endpoint shape is compatibility-test evidence, *not* an identified deployed build — which
is exactly the case where a name-based scrub is the safe superset and an exact-DDL match is not. The asymmetry with
`audit_log_settings` (two recognized layouts, inline and ALTER-appended) makes it sharper: the repo's own historical
pattern for adding a column produces an appended layout, and that variant is recognized for the audit table but not
for the endpoint table. Details inline.

Smaller notes, none blocking:

- Whole-database `PRAGMA foreign_key_check` gates adoption on retired tables the PR elsewhere promises to leave alone.
- 34 call sites in `shared.rs` churned from `.storage()` to `.map_err(migrations::sqlite_error)`; an extension trait
  would give the same redaction for ~34 fewer diff lines and better call-site readability.
- Third near-identical copy of the runner, now with a third distinct validation strength. Worth a follow-up, not this
  PR (mentioning, not asking you to refactor adjacent code here).
- The pragma/posture paragraph is duplicated verbatim between `crates/storage-sqlite/README.md` and
  `docs/marmot-architecture/further-context/app-sqlite-storage-boundaries.md`. Two places to drift; one canonical
  statement plus a link would be better.

The test and fixture investment is proportionate for a data-migration change, and the provenance README that
separates repository evidence from unknown deployed coverage is the right way to state it.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T10:40:49Z` |
| Commit reviewed | `492a06ea8d8628e2cd5cd438ac29d48dd110cb8f` |

This is a careful, well-scoped change. The frozen-DDL recognition approach (PRAGMA introspection plus normalized DDL comparison against independently extracted historical fixtures) is the right fail-closed posture for adopting unversioned databases, and I independently verified the fixture provenance claims against git history: `pre_ledger.sql` matches `fe133b82` byte-for-byte in structure, and the six master-lineage revisions of `shared.rs` do fall into the three documented DDL generations. The `9db9dbc0` compatibility-test evidence for the legacy endpoint and audit-column layouts also checks out.

Specific things I validated:

- The body-plus-ledger transaction contract is sound: immediate transactions, full prefix revalidation under the write lock, exact version/name matching, typed future-version refusal before any body runs, and the `RAISE(IGNORE)`/tampered-insert defense.
- The `.storage()` → `sqlite_error` swap preserves BUSY/LOCKED (including extended codes) transient classification via `map_sqlite_error`, while stripping database-controlled messages — consistent with the observability invariant and now covered by tests that assert trigger payloads and stored values never leak into errors or tracing output.
- Retired-column handling (transactional endpoint NULLing, inert `data_mode`, untouched unused directory tables) matches the documented non-goals and avoids table rebuilds/rowid changes; the 30,003-row typed-value/rowid assurance test is strong evidence.
- The interrupted-upgrade child-process test (exit mid-transaction under DELETE and WAL) is a genuinely good recovery test, and its stated limits (not power-loss) are honest.
- Docs (`app-sqlite-storage-boundaries.md`, `current-state.md`), crate `AGENTS.md`/`README.md` were updated to match; no version bumps; commit is signed.

One non-blocking follow-up noted inline: the app-cache runner predates this hardening and would benefit from the same ledger-insert verification.

Note: posted as a comment rather than an approval because GitHub does not allow approving one's own PR; a second reviewer's approval is still required.

Validation gap: the GitHub `Rust tests (1-4)`, `Simulator and vectors`, and e2e jobs were still pending when I reviewed; `check`, `clippy`, `fmt`, and audit pass. I did not execute code locally (review-only); this approval relies on the author's reported local runs (`cargo test -p storage-sqlite`: 482 unit + 2 integration) plus the passing CI gates.

crates/storage-sqlite/src/shared/migrations.rs
```diff
@@ -0,0 +1,306 @@
...
+            continue;
+        }
+        let has_trigger: bool = tx.query_row(
+            "SELECT EXISTS(SELECT 1 FROM sqlite_master WHERE type = 'trigger' AND tbl_name = ?1)",
+            [table], |r| r.get(0),
+        ).map_err(sqlite_error)?;
+        // No historical shared-table triggers exist. Refuse side effects that
+        // could alter live rows during the endpoint update or later writes.
+        if has_trigger {
+            return Err(invalid_shape());
```
erskingardner, line 179: Small accuracy fix: the comment claims the check refuses side effects "during the endpoint update **or later
writes**", but this only runs inside the version-1 body. Once the ledger row exists, every subsequent open takes the
read-only path in `run` and never re-inspects triggers, so a trigger added after adoption is not caught. The
`app-sqlite-storage-boundaries.md` paragraph inherits the same overclaim. Suggest trimming to what is enforced: no
triggers at adoption time, so the endpoint update cannot fire one.

crates/storage-sqlite/src/shared.rs
```diff
@@ -91,58 +94,20 @@ impl SqliteSharedStorage {
...
         conn.busy_timeout(Duration::from_millis(SHARED_BUSY_TIMEOUT_MS))
-            .storage()?;
-        conn.pragma_update(None, "foreign_keys", true).storage()?;
+            .map_err(migrations::sqlite_error)?;
+        conn.pragma_update(None, "foreign_keys", true)
+            .map_err(migrations::sqlite_error)?;
         conn.pragma_update(None, "trusted_schema", false)
-            .storage()?;
+            .map_err(migrations::sqlite_error)?;
         conn.execute_batch(
```
erskingardner, line 104: 34 call sites in this file changed from `.storage()` to `.map_err(migrations::sqlite_error)`. The redaction is worth
having, but the mechanics make the file noisier at every query and account for a large share of the 70/97 churn here,
none of which is about schema versioning.

An extension trait alongside `sqlite_error` keeps the diff to the import line and the trait definition, leaves call
sites as short as they were, and makes the redacting mapper the path of least resistance for the next query added to
this file:

```rust
pub(super) trait SharedSqliteResultExt<T> {
    fn shared(self) -> StorageResult<T>;
}

impl<T> SharedSqliteResultExt<T> for rusqlite::Result<T> {
    fn shared(self) -> StorageResult<T> {
        self.map_err(sqlite_error)
    }
}
```

Separately: the `_ =>` arm collapses every non-`SqliteFailure` rusqlite error into one static string. Variants like
`InvalidColumnType(idx, name, type)` and `FromSqlConversionFailure` carry no database-controlled content — only
schema-shape facts this crate already publishes — so flattening them costs debuggability without buying privacy.
Worth keeping their `to_string()` and reserving the opaque text for variants that can embed values.

crates/storage-sqlite/src/shared/migrations.rs
```diff
@@ -0,0 +1,306 @@
...
+fn run(conn: &mut Connection, migrations: &[Migration]) -> StorageResult<()> {
+    for (index, migration) in migrations.iter().enumerate() {
+        if migration.version != index as i64 + 1 {
+            return Err(invalid_history());
+        }
+    }
+    // Already-current opens acquire no write lock. Pending opens recheck the
+    // entire prefix after BEGIN IMMEDIATE, not just before waiting for a writer.
+    let applied = applied_count(conn, migrations)?;
+    for (index, migration) in migrations.iter().enumerate().skip(applied) {
```
erskingardner, line 47: Noting, not asking you to fix it in this PR: this is the third copy of the same runner
(`crates/storage-sqlite/src/migrations.rs`, `crates/marmot-app/src/directory/cache/migrations.rs`, now this one), and
`cache/migrations.rs` already carries a comment explaining why it was copied rather than shared.

The part that will bite is that the three have now drifted in *strictness*, not just in ledger name: this runner's
`table_matches` pins normalized DDL, `pragma_table_xinfo`, foreign keys and index structure, while the app-cache
runner's `validate_version_1` checks only `pragma_table_info` name/type/notnull/default/pk plus index columns. So the
directory cache — which holds strictly more sensitive data than this store — has the weaker adoption gate. Three
runners with three different definitions of "recognized schema" is a worse position than three runners with one
definition. Worth a follow-up issue to lift `table_matches`/`normalized_ddl` into one place.

crates/storage-sqlite/src/shared/migrations.rs
```diff
@@ -0,0 +1,306 @@
...
+        let inserted = tx
+            .execute(
+                "INSERT INTO shared_schema_migrations (version, name, applied_at_unix_seconds)
+             VALUES (?1, ?2, CAST(strftime('%s', 'now') AS INTEGER))",
+                params![migration.version, migration.name],
+            )
+            .map_err(sqlite_error)?;
+        // A trigger using RAISE(IGNORE), or changing/deleting the inserted row,
+        // must not let an unrecorded migration commit.
+        if inserted != 1 || applied_count(&tx, migrations)? != index + 1 {
```
erskingardner, line 70: Non-blocking follow-up: this ignored/tampered-insert defense (and the ledger shape validation in `applied_count`) is stricter than the app-cache runner in `crates/marmot-app/src/directory/cache/migrations.rs`, which discards the ledger `INSERT` result — a `RAISE(IGNORE)` trigger there would let a migration body commit unrecorded. The asymmetry is somewhat justified (the shared store is unencrypted while the app cache is SQLCipher), but it would be worth a follow-up to backport the `inserted != 1` + recount checks so the three migration domains don't drift apart in their tamper guarantees.

crates/storage-sqlite/src/shared/migrations.rs
```diff
@@ -0,0 +1,306 @@
...
+
+/// Drop SQLite's optional database-controlled message before mapping the error.
+/// Existing StorageError variants retain the extended code in their safe text
+/// and classify BUSY/LOCKED (including extended codes) as transient.
+pub(super) fn sqlite_error(error: rusqlite::Error) -> StorageError {
+    match error {
+        rusqlite::Error::SqliteFailure(code, _) => {
+            crate::codec::map_sqlite_error(rusqlite::Error::SqliteFailure(code, None))
+        }
+        _ => StorageError::Backend("shared store SQLite operation failed".into()),
```
erskingardner: Non-blocking: this catch-all collapses every non-`SqliteFailure` rusqlite error (`FromSqlConversionError`, `QueryReturnedNoRows`, `IntegralValueOutOfRange`, ...) into one identical generic message. The privacy concern that motivates stripping `SqliteFailure` messages does not apply to those variants — their text is generated by rusqlite from Rust type names and column indices, not database-controlled content — so real bugs (e.g. a schema/type drift in a row get) will now be undiagnosable beyond "shared store SQLite operation failed". Consider keeping `error.to_string()` for the non-`SqliteFailure` arm.

Minor placement nit while you're here: `sqlite_error` is now the error mapper for every operation in `shared.rs`, but it lives in the `migrations` module, so unrelated call sites read `map_err(migrations::sqlite_error)`. It would sit more naturally in `shared.rs` itself (or beside `map_sqlite_error` in `codec`).

crates/storage-sqlite/src/shared/migrations.rs
```diff
@@ -0,0 +1,306 @@
...
+    // Missing tables are safe to create only after existing ones are recognized.
+    // Leave old unused directory tables and all their rows untouched.
+    tx.execute_batch(VERSION_1_SQL).map_err(sqlite_error)?;
+    if legacy_endpoint {
+        tx.execute("UPDATE relay_telemetry_settings SET otlp_endpoint = NULL WHERE otlp_endpoint IS NOT NULL", [])
+            .map_err(sqlite_error)?;
+    }
+    // A ledger never blesses pre-existing orphaned rows; this check is read-only.
+    let mut check = tx
+        .prepare("PRAGMA foreign_key_check")
```
erskingardner: Non-blocking question: the unqualified `PRAGMA foreign_key_check` inspects the whole database, including the retired directory tables (`directory_key_packages` has an FK to `directory_users`) that this migration explicitly leaves untouched and never reads. A single orphaned row in those dead tables — only possible via out-of-band edits, since historical builds ran with `foreign_keys=ON` — would refuse adoption and make the entire shared store (settings, telemetry install id, live directory cache) unavailable over data no live code path uses. Was scoping this to the live tables (e.g. `PRAGMA foreign_key_check(directory_user_follows)`, the only live table with an FK) considered? If the database-wide refusal is intentional fail-closed posture, a one-line comment saying it deliberately covers retired tables too would prevent a future reader from "fixing" it.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T10:45:30Z` |
| Commit reviewed | `492a06ea8d8628e2cd5cd438ac29d48dd110cb8f` |
| Model | `openrouter/qwen/qwen3.8-max` |

Solid, well-scoped change. The premise is sound: versioning `shared.sqlite3` was an explicitly documented follow-up in the storage-boundaries doc, and the design matches the established per-store ledger pattern (independent `shared_schema_migrations`, exact-prefix history validation, typed future-version refusal, body+ledger in one immediate transaction). The adoption machinery is conservative in the right direction: unrecognized shapes, live-table triggers, and ledger tampering (RAISE(IGNORE), row mutation, gaps, duplicates) all fail closed without recording v1, and the rollback/interrupted-process tests verify endpoint neutralization cannot survive a failed ledger insert. Error-message stripping aligns with the repo's privacy invariants, and fixtures are provenance-documented against concrete revisions. Concurrency handling (read-only current opens, write-lock recheck of the full prefix) is correct.

Two non-blocking observations are in inline comments: the blanket collapse of non-`SqliteFailure` rusqlite errors into one generic message, and the database-wide `PRAGMA foreign_key_check` that can refuse adoption because of orphaned rows in retired, unused directory tables the migration otherwise promises not to touch.

Note on validation: at review time the remote CI compile/clippy/format/audit gates had passed but the four Rust test shards were still running; merge should wait for them (author reports full local `cargo test -p storage-sqlite` and `just fast-ci` green).

The bulk of the +1416 lines is tests, fixtures, and docs, which is appropriate for a durability-contract change; the mechanical `.storage()` → `map_err(migrations::sqlite_error)` sweep across `shared.rs` is churn, but it is what makes the message-stripping guarantee uniform rather than migration-only.

crates/storage-sqlite/src/shared/migrations.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+    let latest_supported = migrations.last().map_or(0, |m| m.version);
+    // Inspect numeric versions first, so a future version refuses before any
+    // body runs even if its name has an unfamiliar SQLite storage type.
+    let found: Option<i64> = conn
+        .query_row(
+            "SELECT max(version) FROM shared_schema_migrations",
+            [],
+            |row| row.get(0),
+        )
```
erskingardner, line 95: Minor consistency nit (non-blocking): if a hand-corrupted ledger stores a non-integer `version` (SQLite's loose typing allows TEXT despite the INTEGER affinity — exactly the class the comment above anticipates for `name`), this `max(version)` read fails with `rusqlite::Error::InvalidColumnType`, which `sqlite_error` maps to the generic `Backend("shared store SQLite operation failed")`. The row-walk below maps the same corruption to the specific `invalid shared store migration history`. Both fail closed, so this only affects diagnosability — but mapping this read's non-`SqliteFailure` errors to `invalid_history()` would keep the classification uniform.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T10:45:54Z` |
| Commit reviewed | `492a06ea8d8628e2cd5cd438ac29d48dd110cb8f` |

## Summary

This is the documented follow-up to the app-cache migration work (#1692): give `shared.sqlite3` its own independent `shared_schema_migrations` ledger instead of unversioned `CREATE TABLE IF NOT EXISTS` plus a separate endpoint repair. The premise is sound — the previous initializer accepted malformed tables and had no downgrade boundary — and the implementation is careful and proportionate to the risk.

What I verified:

- **Fixture provenance**: I spot-checked `9db9dbc0` and `fe133b82` against git history; the extracted DDL matches, including the tab-indented formatting that the DDL normalization deliberately tolerates, and the pre-mode audit shape.
- **Transaction discipline**: body + ledger insert in one `BEGIN IMMEDIATE` transaction, recheck under the write lock (TOCTOU between the read-only history check and the apply is closed), and the inserted-row count + recount defeats `RAISE(IGNORE)` / row-rewriting ledger triggers. This matches the multi-step-state-changes contract in the architecture docs.
- **Fail-closed semantics**: unrecognized DDL, live-table triggers, views squatting on table names, and FK violations are refused before v1 is recorded, with no partial writes (verified by the rollback tests). Current opens stay read-only.
- **Privacy redaction**: `sqlite_error` drops database-controlled messages while preserving extended codes and BUSY/LOCKED transience — consistent with the observability invariant; the trigger-payload leak tests are a nice touch.
- **Repo invariants**: independent per-store ledgers, no version bumps, no transport/session changes, docs (`app-sqlite-storage-boundaries.md`, `current-state.md`, crate AGENTS/README) updated in the same change.

Two notes, neither blocking:

1. The one inline nit about error classification for a hand-corrupted non-integer `version` in the ledger.
2. The `Relay runtime tests` CI failure is a known master flake, not caused by this PR: `removed_member_triggers_local_push_token_cleanup` failed identically on master (run 33724316503, same assertion at `relay_runtime.rs:4007`), and the other recent master failures (`app_runtime_archive_survives_subsequent_inbound_delivery`, `app_runtime_serves_member_reads_before_initial_catch_up_completes`) are different timing tests in the same file. Both failures here are `Elapsed` timeouts in event-wait helpers. Worth a re-run, not a revert.

The test coverage (30k-row value/rowid comparison across all three DDL generations, body/ledger rollback, bounded process-exit recovery in both journal modes) is the strongest part of this PR. Nice work.

crates/storage-sqlite/src/shared/migrations.rs
erskingardner: Addressed in 5f0dc400 using option 1: the current relay-settings schema with an appended nullable TEXT otlp_endpoint is now recognized alongside the inline layout. A file-backed regression checks endpoint NULLing, exact preservation of every live setting/rowid, integrity, reopen, and read-only current migration checks. The alternatives loop is explicit. Documentation labels this as defensive compatibility coverage, not evidence of another deployed schema. Arbitrary incompatible layouts still fail closed; clearing data and then returning an error inside the same transaction would roll that clearing back, so the suggested fallback cannot independently guarantee scrubbing while preserving the body-plus-ledger rollback contract.

crates/storage-sqlite/src/shared/migrations.rs
erskingardner: Addressed in 5f0dc400: adoption now checks foreign_key_check(directory_user_follows), the only FK in the frozen live v1 contract. A regression seeds an orphaned directory_key_packages row, verifies successful adoption and live FK/integrity checks, and confirms the retired row and its pre-existing violation remain untouched. Live orphan refusal remains covered.

crates/storage-sqlite/src/shared/migrations.rs
erskingardner: Corrected in 5f0dc400. The code and architecture page now state that triggers are checked at adoption time only. Current opens validate history without rechecking triggers or live table shapes.

crates/storage-sqlite/src/shared.rs
erskingardner: Addressed in 5f0dc400 with SharedSqliteResultExt in shared/error.rs, restoring compact .storage() calls while retaining uniform redaction and moving the mapper out of migrations. Diagnostics now retain conversion categories, indices and SQLite types. I did not forward all non-SqliteFailure messages: the pinned rusqlite IntegralValueOutOfRange displays the stored integer, FromSqlConversionFailure displays an arbitrary nested cause, and InvalidColumnType contains a database-controlled column name. New tests verify useful categories without those values/names/causes.

crates/storage-sqlite/src/shared/migrations.rs
erskingardner: Acknowledged as a follow-up, with no adjacent-runner refactor in this PR. The original scope explicitly keeps the three ownership domains independent and excludes changes to session and app-cache migrations. Shared validation helpers and cross-runner adoption-contract parity need a separate change with their own compatibility fixtures; this PR does not claim parity beyond its shared-store contract.

crates/storage-sqlite/src/shared/migrations.rs
erskingardner: Acknowledged as a separate app-cache follow-up. This PR keeps its ignored/tampered-insert defense and tests, while preserving the requested boundary against modifying app_cache_schema_migrations or session migrations. The shared changes do not claim to fix that adjacent runner's ledger-insert behavior.

crates/storage-sqlite/src/shared/migrations.rs
erskingardner: Addressed the debuggability and placement concerns in 5f0dc400: shared/error.rs now owns the mapper and SharedSqliteResultExt. It preserves safe conversion categories, column indices and SQLite types. Forwarding all non-SqliteFailure text is unsafe with the pinned rusqlite: IntegralValueOutOfRange includes the stored integer; FromSqlConversionFailure includes an arbitrary nested error; InvalidColumnType includes the database-controlled column name. Regression tests verify category preservation and redaction of those fields.

crates/storage-sqlite/src/shared/migrations.rs
erskingardner: Scoped to the live v1 FK in 5f0dc400. The new retired-orphan regression proves that unused directory_key_packages data cannot block live settings/identity adoption, while preserving that data exactly. Live directory_user_follows orphan refusal and successful-fixture integrity checks remain covered.

crates/storage-sqlite/src/shared/migrations.rs
erskingardner: Made non-SqliteFailure errors from the max(version) read map to invalid shared store migration history in 5f0dc400; SQLite failures still retain safe result codes and transient classification. One clarification: this ledger uses INTEGER PRIMARY KEY, so ordinary SQLite inserts of non-integer versions fail with a datatype mismatch rather than using loose affinity. A regression pins that behavior and separately verifies malformed BLOB migration names produce the static history error.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T11:10:57Z` |
| Commit reviewed | `5f0dc400d2b05dfd6b62b565716ad016f4e6530f` |

## Re-review: 492a06ea → 5f0dc400

This commit addresses the findings from the first review pass cleanly, and I found no new issues introduced by the changes.

Verified resolution of prior findings:

- **Ledger error classification** (the inline nit): the `max(version)` read in `applied_count` now routes non-`SqliteFailure` errors to `invalid_history()`, so a hand-corrupted non-integer version is classified consistently with the row-walk. The new `ledger_integer_primary_key_rejects_text_and_bad_name_is_invalid_history` test covers both the INTEGER PRIMARY KEY rejection of text and a blob name mapping to the specific history error.
- **Error diagnosability**: `shared/error.rs` is a good refactor. It restores useful conversion categories (invalid column type, conversion failure, out-of-range, no/multiple rows, invalid index) with column *indexes* only — no column names, stored values, or causes — and `SharedSqliteResultExt` keeps shared.rs ergonomic while the crate-wide unredacted mapper stays out of scope for this module. The redaction test ("private-column-name", value `918273645`, "private-conversion-value" all absent) pins the contract.
- **Appended-endpoint variant**: defensive acceptance of `otlp_endpoint` appended after `updated_at_ms`, with neutralization applied to either inline or appended layout. The fixture README and boundaries doc are explicit that this is review-prompted defensive coverage, not a deployment claim — good evidence discipline. The alternatives-loop refactor in `version_1` handles both settings tables uniformly and still fails closed for everything else.
- **FK gating narrowed to the live contract**: `PRAGMA foreign_key_check(directory_user_follows)` instead of the whole database. This is the right availability/rigor trade: a pre-existing orphan in a retired table (`directory_key_packages`) shouldn't make live settings and the installation identity unavailable, and the new regression test proves adoption succeeds, live FK stays clean, integrity holds, and the retired violation remains untouched. Orphans in the live table are still refused.
- **Docs**: boundaries doc, README, and fixture README updated in the same change, including the honest boundary note that already-current opens don't recheck triggers or table shapes.

CI on this head is fully green, including `Relay runtime tests` (which failed on the previous head) — confirming the earlier failure was the known master flake, not a regression from this PR.

No new findings. Recommend merge.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T11:10:58Z` |
| Commit reviewed | `5f0dc400d2b05dfd6b62b565716ad016f4e6530f` |

Re-review of the delta `492a06ea..5f0dc400` ("address shared migration review findings"). Posted as a comment because GitHub does not allow approving one's own PR; a second reviewer's approval is still required.

Every prior finding is resolved, and I verified the fixes hold up:

- **Endpoint scrub coverage** (was blocking): the appended-`otlp_endpoint` layout is now recognized via the extended `appended` reference, and the alternatives-table rewrite sets `legacy_endpoint` correctly for both inline and appended shapes (the old audit-only condition would no longer have been reachable — the refactor was necessary and is correct). The new file-backed assurance test proves the endpoint is NULLed while live settings are preserved, and that current reopens don't repeat the write. Combined with the existing all-generations assurance test, master's per-open scrub is fully covered.
- **FK check scoping**: `PRAGMA foreign_key_check(directory_user_follows)` now validates exactly the one FK in the frozen live contract, and the retired-orphan regression proves pre-existing violations in untouched `directory_key_packages` no longer gate adoption while live FK integrity is still enforced. Docs updated to match.
- **Error mapper placement**: `shared/error.rs` + `SharedSqliteResultExt` restores compact `.storage()` call sites, removes the mapper from the migration module, and widens the coverage — every shared-store operation (including open, transactions, and settings writes) now redacts database-controlled messages, column names, stored values, and causes, while retaining extended result codes and BUSY/LOCKED transient classification. The new conversion-category test asserts redaction of column names and values.
- **Ledger `max(version)` read**: non-`SqliteFailure` errors now map to invalid-history rather than a misleading SQLite error; `INTEGER PRIMARY KEY` already rejects text versions, and the blob-name case is tested.
- Trigger-check comment and docs now accurately describe adoption-time-only enforcement.

The new commit is signed (PGP verified per GitHub), and all CI checks now pass — the earlier validation gap is closed: Rust tests (1-4), simulator and vectors, CLI e2e (mock relay), relay runtime, and both platform harness builds are green (Tamarin and wn-agent publish legitimately skip).

Two follow-ups from the previous round remain correctly deferred, not forgotten: backporting the tampered-insert defense to the app-cache runner, and the three-runner duplication question. Neither belongs in this PR's scope.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T11:12:23Z` |
| Commit reviewed | `5f0dc400d2b05dfd6b62b565716ad016f4e6530f` |
| Model | `openrouter/qwen/qwen3.8-max` |

Re-review of `5f0dc400` (delta from `492a06ea`). Both prior findings are properly resolved, and CI is fully green on this head (all four Rust test shards, clippy, format, check, audit).

**Finding 1 — error mapper (resolved, better than suggested).** `sqlite_error` moved out of `migrations` into a new `shared/error.rs` with a `SharedSqliteResultExt::storage()` drop-in, which also shrank the `shared.rs` churn back to `.storage()` call sites. Non-`SqliteFailure` errors now keep diagnosable categories (column index, type kind, no-rows/multiple-rows) while names, values, and causes are dropped — the new `conversion_errors_keep_categories_without_names_values_or_causes` test pins exactly that contract. The `max(version)` conversion path now maps to `invalid_history` instead of the generic backend error, and the `x'ff'`-name ledger test covers it.

**Finding 2 — FK check scope (resolved).** `PRAGMA foreign_key_check(directory_user_follows)` now checks only the live contract's single FK; the in-code comment states the rationale, and `retired_orphans_do_not_gate_live_store_adoption` verifies adoption succeeds while the retired `directory_key_packages` violation is deliberately preserved untouched. README/fixtures/boundaries docs were updated consistently.

One minor observation on the delta, non-blocking: the defensively accepted appended-`otlp_endpoint` relay shape is the first recognized variant without historical provenance — a small departure from the otherwise evidence-based recognition philosophy (every other accepted shape traces to a concrete revision or historical repair). It is benign (nullable column, transactionally neutralized, cannot resurrect the endpoint) and is honestly labeled as defensive coverage in the fixture README and boundaries doc, so I would not block on it; just be aware it slightly widens the acceptance surface for hand-edited stores.

The `alternatives` refactor of the recognition loop preserves the prior semantics (reference → legacy → appended per table, `legacy_endpoint` set only when the relay table matched an endpoint-bearing shape). No new issues found in the delta. Recommend merge.

crates/storage-sqlite/src/shared/migrations.rs
```diff
@@ -0,0 +1,314 @@
...
+    let appended = Connection::open_in_memory().map_err(sqlite_error)?;
+    appended
+        .execute_batch(VERSION_1_SQL)
+        .map_err(sqlite_error)?;
+    // The additive repair in 9db9dbc0 put this column AFTER updated_at_ms.
+    appended.execute_batch("ALTER TABLE audit_log_settings ADD COLUMN data_mode TEXT NOT NULL DEFAULT 'obfuscated_sensitive_data'")
+        .map_err(sqlite_error)?;
+
+    // Accept the review's additive endpoint layout defensively. History proves
+    // the inline layout only; this is not evidence of an additional shipped build.
```
erskingardner, line 162: Non-blocking, but worth cleaning before merge: three separate places now explain this layout by pointing at *a review* rather than at what the code guarantees — here (`Accept the review's additive endpoint layout defensively`), `assurance_tests.rs` (`Defensive compatibility case from review, not a claimed historical build`), `fixtures/README.md` (`The review-added appended-endpoint test ...`) and `app-sqlite-storage-boundaries.md` (`Review also prompted defensive acceptance of ...`).

A future reader has no access to that review and cannot act on the reference. The fact worth preserving is the one already in the second half of each sentence: repository history evidences only the inline layout, and the appended layout is accepted so that a store carrying a plaintext endpoint in that position is still adopted and scrubbed rather than refused forever. Stating that directly is shorter and survives the PR:

```rust
// Also accept an endpoint appended to the current schema. History evidences
// only the inline layout; accepting this one means such a store is scrubbed
// rather than refused with the plaintext endpoint left on disk.
```

Same edit applies to the other three; the provenance sections in `fixtures/README.md` are the right place for "history does not establish this shape shipped", which they already say.

crates/storage-sqlite/src/shared/error.rs
```diff
@@ -0,0 +1,38 @@
+//! Privacy-safe SQLite diagnostics for every shared-store operation.
+use cgka_traits::storage::{StorageError, StorageResult};
+
+/// Shared queries use this instead of the crate-wide unredacted conversion.
+pub(super) trait SharedSqliteResultExt<T> {
+    fn storage(self) -> StorageResult<T>;
```
erskingardner, line 6: This resolves the churn and placement points well — `shared.rs` is back to compact call sites and the mapper is out of `migrations`. One footgun in the naming, though.

`codec.rs:215` already defines `pub(crate) trait SqliteResultExt<T> { fn storage(self) -> StorageResult<T>; }`, and its impl is on the same receiver type (`rusqlite::Result<T>`) with the same method name — but it maps through the *unredacted* `map_sqlite_error`. Today that is fine only because `shared.rs` dropped its `crate::SqliteResultExt` import in this PR. The next file added under `shared/` that reaches for the familiar crate-wide import gets the unredacted mapper with byte-identical call syntax (`.storage()?`), no compile error and no reviewer-visible difference.

A distinct method name makes the wrong choice impossible to make silently:

```rust
pub(super) trait SharedSqliteResultExt<T> {
    fn shared(self) -> StorageResult<T>;
}
```

That is a mechanical rename across the `shared.rs` call sites and removes the one way this redaction boundary can be re-crossed by accident.

crates/storage-sqlite/src/shared/migrations.rs
```diff
@@ -0,0 +1,314 @@
+//! Installation-wide shared.sqlite3 migration domain. Independent of account
+//! and app-cache ledgers; never copies their histories or application records.
+use super::error::sqlite_error;
```
erskingardner, line 3: Minor consistency follow-on to the new extension trait: `shared.rs` now reads `.storage()?`, but this file still spells the same mapping out longhand at ~26 call sites (`map_err(sqlite_error)` on lines 50, 53, 59, 67, 82, 83, 110, 111, 113, 135, 147, 150, 152, 156, 159, 165, 175, 202, 205, 212, 215, 217, 237, 276, 282, 284).

Adding `use super::error::SharedSqliteResultExt;` alongside the `sqlite_error` import lets those become `.storage()?` too, which is a net deletion and leaves `sqlite_error` referenced only where it is genuinely needed — the `applied_count` classification arm at line 96 and the tests. Not worth a re-review cycle on its own; fold it in if you touch the file again.

crates/storage-sqlite/README.md
```diff
@@ -35,6 +38,30 @@ when storage semantics evolve.
...
+Shared version `1 / 0001_shared_store` establishes the five live tables. Each migration body and its ledger row commit
+in the same immediate transaction; failure rolls back both. Recorded versions and names must be an exact prefix of
+the compiled registry. Future versions return `StorageError::UnsupportedSchemaVersion`. A pending opener rechecks the
+prefix under the write lock; an already-current opener validates history using reads only.
+
+Unversioned shared tables must match frozen current or verified historical definitions before adoption. Validation
+covers columns, types, nullability, defaults, primary keys, foreign keys, CHECK expressions, collations and indexes.
+Conservative DDL comparison can refuse equivalent but unrecognized SQL; incompatible shapes fail with static errors
+and no version row. Missing tables are created. Public users, ordered follows, live settings, timestamps, installation
+identity and rowids are preserved. Existing unused directory tables remain untouched; their pre-existing orphaned rows do not gate live-store adoption. The recognized nullable
```
erskingardner, line 54: Rewrap nit: the review edits were inserted without re-flowing, so this paragraph now mixes ~120-column lines with a 175-column one here and a 139-column one below, and `app-sqlite-storage-boundaries.md` line 74 reaches 189 columns while line 73 ends on a dangling `transaction, while audit`. Nothing lints it, but it makes the next diff in these paragraphs noisier than it needs to be.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-06T11:14:51Z` |
| Commit reviewed | `5f0dc400d2b05dfd6b62b565716ad016f4e6530f` |

_Posted with the COMMENT event rather than an approval only because GitHub does not allow approving your own pull request._

Re-review of the delta since [my earlier review](https://github.com/marmot-protocol/mdk/pull/1707#pullrequestreview-5125078703), i.e. `492a06ea` → `5f0dc400` (9 files, +239/−70).

## Prior findings

**Blocking — narrowed endpoint scrub (`migrations.rs`). Resolved.** The `appended` reference now also carries `ALTER TABLE relay_telemetry_settings ADD COLUMN otlp_endpoint TEXT`, and `alternatives` offers `[legacy, appended]` for both settings tables, so the exact case I described — an endpoint appended after `updated_at_ms` — is adopted and NULLed instead of refused with the plaintext value stranded on disk. `legacy_endpoint |= *table == \"relay_telemetry_settings\"` is only reachable after the table matched an alternative, and both alternatives carry the column, so the guarded `UPDATE` cannot target a missing column. `appended_endpoint_is_neutralized_without_rewriting_live_settings` pins the NULLing, exact preservation of the other values, integrity, reopen and the read-only current path. The alternatives loop also reads far better than the negated compound conditional it replaces.

One note on the reply rather than the code: the objection to option 2 was that clearing and then erroring inside one transaction rolls the clearing back — true, but that only rules out doing it *in the same transaction*, not a committed scrub before the refusal. I am not asking for it. Refusing to write to a schema you have declined to adopt is a defensible posture, and with the appended layout recognized the residual exposure is confined to hand-edited shapes.

**Blocking-adjacent — database-wide `foreign_key_check`. Resolved.** Now `PRAGMA foreign_key_check(directory_user_follows)`, the only FK in frozen v1, with a comment saying why retired tables are out of scope. `retired_orphans_do_not_gate_live_store_adoption` seeds an orphan into `directory_key_packages` with `foreign_keys` off, then proves adoption succeeds, the live FK check is clean, integrity is `ok`, and the retired row *and its violation* survive byte-for-byte. That is exactly the right assertion set.

**Trigger-scope overclaim. Resolved.** Code comment and the architecture page both now say triggers are checked at adoption time only, and that current opens do not recheck.

**`.storage()` churn / mapper placement / error flattening. Resolved.** `shared/error.rs` owns both the mapper and `SharedSqliteResultExt`, `shared.rs` is back to compact call sites, and the mapper keeps `InvalidColumnType` / `FromSqlConversionFailure` / `IntegralValueOutOfRange` categories, column indices and SQLite types while dropping the column name, the stored integer and the nested cause. I checked the three variants against the pinned rusqlite `Display` impls and agree with the split — you kept the diagnostics that are schema facts and dropped the ones that are database-controlled. `conversion_errors_keep_categories_without_names_values_or_causes` pins it. One naming footgun left, inline.

**`max(version)` classification. Resolved.** Non-`SqliteFailure` errors from that read now map to `invalid_history()`, matching the row walk below, while BUSY/LOCKED still route through `sqlite_error`. Point taken on `INTEGER PRIMARY KEY` rejecting TEXT outright; the new regression pins both that and the BLOB-name path.

**Runner triplication and the app-cache ledger-insert asymmetry.** Accepted as follow-ups; both were raised as non-blocking and neither belongs in this PR's ownership boundary. Please do file the issue for lifting `table_matches`/`normalized_ddl` into one place — the finding was that three runners now hold three different definitions of "recognized schema", and that only gets more expensive the longer it sits.

## New, all non-blocking

- Four comments/doc sentences now explain this layout by pointing at "the review" instead of at the invariant. Inline.
- `SharedSqliteResultExt::storage` shares its name and receiver type with the crate-wide, *unredacted* `SqliteResultExt::storage`. Inline.
- `migrations.rs` still spells `map_err(sqlite_error)` at ~26 sites the new trait could shorten. Inline.
- Markdown rewrap sloppiness in `README.md` and the boundaries doc. Inline.

Full CI is green on `5f0dc400`, including the two relay-runtime tests that failed on the original head. Nothing here blocks merge.

